PR #8549, opened 2019-01-02
Drop SDL2/ prefix on includes

janisozaur:
pkg-config, cmake, all demos and tests define the path with SDL2/ already included


janisozaur:
@IntelOrca please update the MSVC project files

IntelOrca:
Not near computer at the moment. The includes are handled in `openrct2.common.props` under `<IncludePath>` so that probably needs updating.

PR #8550, opened 2019-01-02
Fix implicit fallthroughs

janisozaur:


src/openrct2/network/Network.cpp
```diff
@@ -30,7 +30,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "18"
+#define NETWORK_STREAM_VERSION "19"
```
ZehMatt: Does this really need a new network version? I don't see any logic changes.

src/openrct2/Cheats.cpp
```diff
@@ -876,10 +876,7 @@ const char* cheats_get_cheat_string(int cheat, int edx, int edi)
             {
                 return language_get_string(STR_CHEAT_MOWED_GRASS);
             }
-            else if (edx == 1)
-            {
-                return language_get_string(STR_CHEAT_CLEAR_GRASS);
-            }
+            return language_get_string(STR_CHEAT_CLEAR_GRASS);
```
janisozaur: Logic change here

src/openrct2/ride/coaster/AirPoweredVerticalCoaster.cpp
```diff
@@ -670,8 +670,8 @@ static void air_powered_vertical_rc_track_vertical_slope_up(
 
                 paint_util_set_segment_support_height(session, SEGMENTS_ALL, 0xFFFF, 0);
                 paint_util_set_general_support_height(session, height + supportHeights[trackSequence], 0x20);
-                break;
             }
+            break;
```
Broxzier: And here as well.

PR #8558, opened 2019-01-05
Peep debug tab

richard-fine:
Introduce a 'Debug tools' tab to the Guest window, when debugging tools are enabled:

image

For now it just shows a few bits of pathfinding data for that peep, but the idea is that this could be a home for all per-peep-related debugging info and tools (e.g. viewing/editing flags, forcing the peep into particular states, etc).

When debugging tools are not enabled, the tab is hidden.


src/openrct2-ui/windows/Guest.cpp
```diff
@@ -2222,3 +2332,94 @@ void window_guest_inventory_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+
+    rct_peep* peep = GET_PEEP(w->number);
+
+    // cx
+    int32_t x = w->x + window_guest_debug_widgets[WIDX_PAGE_BACKGROUND].left + 4;
+    // dx
+    int32_t y = w->y + window_guest_debug_widgets[WIDX_PAGE_BACKGROUND].top + 4;
+
+    char buffer[4096];
+    snprintf(buffer, 4096, "Position: %i, %i, %i", peep->x, peep->y, peep->z);
```
richard-fine: Should probably use sizeof(buffer) rather than hardcoding 4096 here.

Gymnasiast:
I have to say: I like the idea!
I'll try giving a more in-depth review later.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -49,32 +50,36 @@ enum WINDOW_GUEST_WIDGET_IDX {
...
     WIDX_TRACK,
 
-    WIDX_RIDE_SCROLL = 10
+    WIDX_RIDE_SCROLL = 11
 };
 
 validate_global_widx(WC_PEEP, WIDX_ACTION_LBL);
 validate_global_widx(WC_PEEP, WIDX_PICKUP);
 
+static const int32_t TabWidth = 30;
```
Gymnasiast: This can be a `constexpr` as well.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -486,7 +563,8 @@ rct_window* window_guest_open(rct_peep* peep)
     window = window_bring_to_front_by_number(WC_PEEP, peep->sprite_index);
     if (window == nullptr)
     {
-        window = window_create_auto_pos(192, 157, &window_guest_overview_events, WC_PEEP, WF_RESIZABLE);
+        window = window_create_auto_pos(
```
Gymnasiast: It's probably easier to read if you put the width into a temporary variable.

Gymnasiast:
Apart from the minor line notes, there is major thing: you're completely bypassing the localisation system. This needs to be changed. You also need to set the correct label and text colours. There are many places where you can take a peek, but I think the ride statistics window is probably the tab most similar to what you need here.

If needed, I can answer questions or help with this.

src/openrct2-ui/windows/Guest.cpp
richard-fine: Fixed in 918772be2bc2b6633cd6664efca7d3e58aeff3c3

richard-fine:
@Gymnasiast I skipped it because I figured debug tools maybe didn't need localizing. You're sure it's necessary?

richard-fine:
Also, if we are going to introduce a bunch of localized string IDs for this, then maybe it's worth having more of a discussion first about what info we actually want to display. I threw the current values into the window as a proof of concept as much as anything else.

Gymnasiast:
> You're sure it's necessary?

Yes. Especially since you cannot set stuff like colours properly if you bypass it.

> Also, if we are going to introduce a bunch of localized string IDs for this, then maybe it's worth having more of a discussion first about what info we actually want to display. I threw the current values into the window as a proof of concept as much as anything else.

It's not very difficult to change it afterwards. I think the current information is fine, but perhaps @duncanspumpkin (who did a lot of stuff with pathfinding) has more suggestions.

src/openrct2-ui/windows/Guest.cpp
richard-fine: b4ea407dfbb7259ddf76b894d36cb8e64a169b23

richard-fine:
Alright. I've fixed the minor issues, and I'll wait to see if @duncanspumpkin or anyone else has feedback on the info we actually want to see before I start allocating new string IDs.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -575,7 +660,7 @@ void window_guest_overview_resize(rct_window* w)
 
     widget_invalidate(w, WIDX_MARQUEE);
 
-    window_set_resize(w, 192, 159, 500, 450);
+    window_set_resize(w, 192 + gConfigGeneral.debugging_tools ? TabWidth : 0, 159, 500, 450);
```
janisozaur:
```
../src/openrct2-ui/windows/Guest.cpp:663:63: error: operator '?:' has lower precedence than '+'; '+' will be evaluated first [-Werror,-Wparentheses]
    window_set_resize(w, 192 + gConfigGeneral.debugging_tools ? TabWidth : 0, 159, 500, 450);
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
../src/openrct2-ui/windows/Guest.cpp:663:63: note: place parentheses around the '+' expression to silence this warning
    window_set_resize(w, 192 + gConfigGeneral.debugging_tools ? TabWidth : 0, 159, 500, 450);
                                                              ^
                         (                                   )
../src/openrct2-ui/windows/Guest.cpp:663:63: note: place parentheses around the '?:' expression to evaluate it first
    window_set_resize(w, 192 + gConfigGeneral.debugging_tools ? TabWidth : 0, 159, 500, 450);
                                                              ^
                               (                                            )
```
```suggestion
    window_set_resize(w, 192 + (gConfigGeneral.debugging_tools ? TabWidth : 0), 159, 500, 450);
```

janisozaur:
ping @zaxcav

zaxcav:
I haven't done any active development on OpenRCT2 for well over a year, though I have been observing the amazing work all the active devs have been doing during this time. Great work all.
The A* pathfinding prototype is still waiting for me to get back to it. Maybe later this year, if it's still relevant.
If someone else is interested in looking at some aspects of the pathfinding I'm happy to answer questions if I can.

The current question, I believe, is what information might be useful to the peep debug tab?
The basic details that are available outside the search function are interesting but extremely minimal and honestly not very illuminating for debugging pathfinding issues. The search function itself generates a _massive_ amount of data and then throws all this away at the end, returning only the chosen direction. It's a pity really, because the full chosen path (or even only the first n steps) would be really useful to have available, for multiple reasons not just debugging.

What would be useful to have in the peep debug tab would be the capability to set the peep's pathfinding data, as is suggested in the OP. That would make testing the pathfinding much easier - set up a peep as needed rather than waiting for an agreeable peep to pass by.

ZehMatt:
I've had a look into it, while useful information its lacking very basic things. One example would be perhaps a checkbox on the debug Tab to highlight path finding affected tiles with different colors or any other indicator sitting on the tile. And as zaxcav suggested there is no interactivity, would be cool to have a picker tool that allows us to set certain coordinates on those fields, better yet being able to modify all properties the pathfinding model contains.

AaronVanGeffen:
> I've had a look into it, while useful information its lacking very basic things. One example would be perhaps a checkbox on the debug Tab to highlight path finding affected tiles with different colors or any other indicator sitting on the tile. And as zaxcav suggested there is no interactivity, would be cool to have a picker tool that allows us to set certain coordinates on those fields, better yet being able to modify all properties the pathfinding model contains.

These are some very good points, but shouldn't block the initial merge, IMO.

ZehMatt:
I'm not against it, however I feel like this is way to raw at this point, if it could maybe already provide a basic function set that would at least provide a better example where and how to put things if anyone else decides to further work on the path finding/AI, that would be a better start. Also I don't really like the font but thats probably because the localization was ignored which supports text attributes to make it more lightweight.

Gymnasiast:
I agree with @AaronVanGeffen that we shouldn't block the initial merge. After all, the improvements @ZehMatt mentions can be added later on. I do think it would be good to put these into an issue, so we don't forget about them.

AaronVanGeffen:
@richard-fine Are you still interested in working on this? Let us know if you need any help.

richard-fine:
Yeah, I've not been doing OpenRCT stuff lately but I'm still interested - was waiting to see if @duncanspumpkin would show up with any thoughts on what info to show, as suggested. At some point I'll get back to it and route the text via the localisation system.

AaronVanGeffen:
At the time he was mentioned, @duncanspumpkin was still on a road trip. While he was no doubt mentioned with the best of intentions, it is a bit unfortunate that you were under the impression that you had to wait for him.

He has since returned though, so while he might have some thoughts now, don't let that stop you. :-)

duncanspumpkin:
I'll review it tomorrow. Sorry didn't see it in my backlog. :)

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1951,7 +2062,7 @@ void window_guest_thoughts_resize(rct_window* w)
         window_invalidate(w);
     }
 
-    window_set_resize(w, 192, 159, 500, 450);
+    window_set_resize(w, 192 + (gConfigGeneral.debugging_tools ? TabWidth : 0), 159, 500, 450);
```
duncanspumpkin: You may as well make a separate function call for this to save the 7 repetitions of it.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -2222,3 +2335,97 @@ void window_guest_inventory_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+        w->widgets = window_guest_page_widgets[w->page];
+        window_init_scroll_widgets(w);
+    }
+
+    w->pressed_widgets |= 1ULL << (w->page + WIDX_TAB_1);
+
+    rct_peep* peep = GET_PEEP(w->number);
+
+    set_format_arg(0, rct_string_id, peep->name_string_idx);
+    set_format_arg(2, uint32_t, peep->id);
```
duncanspumpkin: For anyone wondering what this code does it sets the global args to the peep name so that the peep name is correctly written at the top of the window.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -2222,3 +2335,97 @@ void window_guest_inventory_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+    int32_t x = w->x + window_guest_debug_widgets[WIDX_PAGE_BACKGROUND].left + 4;
+    // dx
+    int32_t y = w->y + window_guest_debug_widgets[WIDX_PAGE_BACKGROUND].top + 4;
+
+    char buffer[512];
+    snprintf(buffer, sizeof(buffer), "Position: %i, %i, %i", peep->x, peep->y, peep->z);
+    gfx_draw_string(dpi, buffer, COLOUR_BLACK, x, y);
+    y += LIST_ROW_HEIGHT;
+
+    snprintf(buffer, sizeof(buffer), "Next: %u, %u, %u", peep->next_x, peep->next_y, peep->next_z);
```
duncanspumpkin: Are these actually useful? next_x and x are usually the same. I can't quite remember the exact design of them.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -2175,3 +2039,76 @@ void window_guest_inventory_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+    window_guest_thoughts_tab_paint(w, dpi);
+    window_guest_inventory_tab_paint(w, dpi);
+    window_guest_debug_tab_paint(w, dpi);
+
+    auto peep = GET_PEEP(w->number);
+    auto x = w->x + window_guest_debug_widgets[WIDX_PAGE_BACKGROUND].left + 4;
+    auto y = w->y + window_guest_debug_widgets[WIDX_PAGE_BACKGROUND].top + 4;
+
+    char buffer[512]{};
+    snprintf(buffer, sizeof(buffer), "%i, %i, %i", peep->x, peep->y, peep->z);
```
Gymnasiast: As I mentioned before, I really want these to be proper localised strings. We did that with all other debugging tools as well.

IntelOrca: ok, done

PR #8559, opened 2019-01-05
Include stdexcept instead of exception

IntelOrca:
Fix build when using Windows 10 SDK 10.0.17763.0.

### DIFF
--- a/src/openrct2/network/Http.cpp
+++ b/src/openrct2/network/Http.cpp
@@ -10,8 +10,8 @@
 #include "Http.h"
 
 #include <cstring>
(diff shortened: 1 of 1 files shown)


src/openrct2/network/Http.cpp
```diff
@@ -10,9 +10,9 @@
 #include "Http.h"
 
 #include <cstring>
-#include <exception>
 #include <memory>
 #include <thread>
+#include <stdexcept>
```
Gymnasiast: Should have been above `<thread>`.

PR #8560, opened 2019-01-05
Begin cleanup of direction handling

richard-fine:
Begin cleaning up code that deals with directions, by:

* Introducing a formal typedef for a direction (instead of just uint8_t)
* Introducing a direction_reverse() helper function, replacing the less-clear "^ 2" trick.

There's a lot of followup to do with changing code to actually use the Direction type, but trying to do it all in a single PR is a recipe for endless merge conflicts, so we will need to do it a bit at a time.

### DIFF
--- a/src/openrct2/world/Location.hpp
+++ b/src/openrct2/world/Location.hpp
@@ -144,7 +144,7 @@ typedef uint8_t Direction;
  * Given a direction, return the direction that points the other way,
  * on the same axis.
  */
-inline Direction direction_reverse(const Direction& dir)
+constexpr Direction direction_reverse(Direction dir)
 {
     return dir ^ 2;
 }
--- a/src/openrct2/world/Location.hpp
+++ b/src/openrct2/world/Location.hpp
@@ -129,7 +129,25 @@ struct TileCoordsXYZ
     int32_t x = 0, y = 0, z = 0;
 };
 
+/**
+ * Cardinal directions are represented by the Direction type. It has four
+ * possible values:
+ * 0 is X-decreasing
+ * 1 is Y-increasing
+ * 2 is X-increasing
+ * 3 is Y-decreasing
+ * Direction is not used to model up/down, or diagonal directions.
+ */
 typedef uint8_t Direction;
+
+/**
+ * Given a direction, return the direction that points the other way,
+ * on the same axis.
+ */
+inline Direction direction_reverse(const Direction& dir)
+{
+    return dir ^ 2;
+}
 
 struct CoordsXYZD
 {
--- a/src/openrct2/world/Location.hpp
+++ b/src/openrct2/world/Location.hpp
@@ -129,10 +129,12 @@ struct TileCoordsXYZ
     int32_t x = 0, y = 0, z = 0;
 };
 
+typedef uint8_t Direction;
+
 struct CoordsXYZD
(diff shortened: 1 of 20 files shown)


src/openrct2/world/Location.hpp
```diff
@@ -129,10 +129,30 @@ struct TileCoordsXYZ
...
+ * 3 is Y-decreasing
+ * Direction is not used to model up/down, or diagonal directions.
+ */
+typedef uint8_t Direction;
+
+/**
+ * Given a direction, return the direction that points the other way,
+ * on the same axis.
+ */
+inline Direction direction_reverse(const Direction& dir)
```
ZehMatt: Why would you choose reference over value here when the type is an integer?

Gymnasiast: I think this can be done as a constexpr (as an equivalent of a define). (cc @IntelOrca )

richard-fine: @ZehMatt I don't think it'll make much difference, but I can change it to be pass-by-value if we have any target platforms where the compiler might not handle it as optimally.

@Gymnasiast I don't see how - the output is a function of the input, which is dynamic at runtime?

richard-fine:
@IntelOrca given the naming standard you're going for in #8561, would you prefer that this PR also use `direction_t` as a name instead of `Direction` ?

IntelOrca:
@richard-fine It might help show that it isn't a class or struct, however since you are suggesting it should be at some point maybe it should be left as you have it.

src/openrct2/world/Location.hpp
ZehMatt: It is recommended to pass small data like that by value otherwise you might end up having code dereferencing memory first, I'm not even sure if the compiler is allowed much margin when it comes to types.

richard-fine: I have high confidence in the compiler's ability to do the right thing here, both w.r.t. how it accesses the parameter and w.r.t. optimising the call, but OK, I'll change it just to be safe. We may end up changing it back later if Direction becomes a struct, but we can cross that bridge when it comes to it.

ZehMatt:
```
inline constexpr Direction direction_reverse1(const Direction& dir)
{
    return dir ^ 2;
}
=
unsigned char direction_reverse1(unsigned char const &) PROC           ; direction_reverse1, COMDAT
        movzx   eax, BYTE PTR [rcx]
        xor     al, 2
        ret     0
unsigned char direction_reverse1(unsigned char const &) ENDP           ; direction_reverse1
```
and
```
inline Direction direction_reverse2(const Direction dir)
{
    return dir ^ 2;
}
=
unsigned char direction_reverse2(unsigned char) PROC                        ; direction_reverse2, COMDAT
        xor     cl, 2
        movzx   eax, cl
        ret     0
unsigned char direction_reverse2(unsigned char) ENDP
```

I would be generally careful when it comes to assumptions, I don't believe the compiler can ever go against the type when it comes to references.

IntelOrca: Whilst this is true, when it is inlined (which it always should be) the reference goes away.

Gymnasiast:
LGTM. If I have time, I'll test and merge it tonight.

PR #8561, opened 2019-01-05
Create new ride_id_t typedef

IntelOrca:
In order to increase the maximum number of rides, we need to ensure we can change everywhere it is stored as a `uint8_t` into something bigger.

### DIFF
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -275,7 +275,7 @@ static uint8_t footpath_element_next_in_direction(TileCoordsXYZ loc, TileElement
  * This is the recursive portion of footpath_element_destination_in_direction().
  */
 static uint8_t footpath_element_dest_in_dir(
-    TileCoordsXYZ loc, [[maybe_unused]] TileElement* inputTileElement, uint8_t chosenDirection, uint8_t* outRideIndex,
+    TileCoordsXYZ loc, [[maybe_unused]] TileElement* inputTileElement, uint8_t chosenDirection, ride_id_t* outRideIndex,
     int32_t level)
 {
     TileElement* tileElement;
@@ -301,7 +301,7 @@ static uint8_t footpath_element_dest_in_dir(
             {
                 if (loc.z != tileElement->base_height)
                     continue;
-                int32_t rideIndex = tileElement->AsTrack()->GetRideIndex();
+                ride_id_t rideIndex = tileElement->AsTrack()->GetRideIndex();
                 Ride* ride = get_ride(rideIndex);
                 if (ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_IS_SHOP))
                 {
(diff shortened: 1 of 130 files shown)


Gymnasiast:
This perhaps also a nice opportunity to ensure that MAX_RIDES and RIDE_ID_NULL are properly used everywhere.

ZehMatt:
Another nitpick would be to get rid of NetworkRideId_t in NetworkTypes.h as its just used for serialisation purposes, DataSerialiserTraits can use ride_id_t instead.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -646,7 +646,7 @@ static void peep_pathfind_heuristic_search(
         if (tileElement->flags & TILE_ELEMENT_FLAG_GHOST)
             continue;
 
-        uint8_t rideIndex = 0xFF;
+        ride_id_t rideIndex = 0xFF;
```
Gymnasiast:
```suggestion
        ride_id_t rideIndex = RIDE_ID_NULL;
```

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1826,7 +1826,7 @@ static bool track_design_place_preview(rct_track_td6* td6, money32* cost, uint8_
     }
 }
 
-static money32 place_track_design(int16_t x, int16_t y, int16_t z, uint8_t flags, uint8_t* outRideIndex)
+static money32 place_track_design(int16_t x, int16_t y, int16_t z, uint8_t flags, ride_id_t* outRideIndex)
 {
     *outRideIndex = 255;
```
Gymnasiast:
```suggestion
    *outRideIndex = RIDE_ID_NULL;
```

src/openrct2/world/Footpath.cpp
```diff
@@ -1581,13 +1581,13 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
 
     if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH && tileElement->AsPath()->IsQueue())
     {
-        int32_t rideIndex = -1;
+        ride_id_t rideIndex = RIDE_ID_NULL;
         uint8_t entranceIndex = 255;
         for (size_t i = 0; i < neighbourList.count; i++)
         {
             if (neighbourList.items[i].ride_index != 255)
```
Gymnasiast:
```suggestion
            if (neighbourList.items[i].ride_index != RIDE_ID_NULL)
```

src/openrct2/world/Footpath.cpp
```diff
@@ -1744,7 +1744,7 @@ void footpath_queue_chain_reset()
  *
  *  rct2: 0x006A76E9
  */
-void footpath_queue_chain_push(uint8_t rideIndex)
+void footpath_queue_chain_push(ride_id_t rideIndex)
 {
     if (rideIndex != 255)
```
Gymnasiast:
```suggestion
    if (rideIndex != RIDE_ID_NULL)
```

ZehMatt:
Any reason you left out the network types?

ZehMatt:
Oh well, the idea was to simply replace NetworkRideId_t with your new type on all instances, that should have done the job, but I can do it later this day maybe, no worries.

PR #8594, opened 2019-01-15
Fix #8591: Game loop does not run at a consistent tick rate of 40 Hz

### LINKED ISSUE #8591: Framerate delimiter speeds up guests
**OS:** Windows 10
**Version:** Both 0.2.1 and current dev build
**Commit/Build:** Both 8ac731e and 51fc2fe

Without the framerate delimiter, guests move as normal (e.g as expected from RCT1) at any framerate.

Turning on the framerate delimiter causes all guests/staff to move at speeds proportional to the framerate. In specific, lower framerates increase the guest/staff speed, suggesting that their speed increases as v = v0 * 40/FPS, or similar. This is especially noticeable at FPS of about 6-10, in savefiles with many guests.

**Steps to reproduce:**
1. Enter a crowded savefile and find a spot where FPS is about 6-10
2. Toggle framerate delimiter in the options menu (first tab)
3. Observe guest speeds and repeat step 2 to see the difference

**Screenshots / Video:**
I could record if necessary, but the combination of low FPS in-game and high enough FPS to get a good recording may be very difficult to achieve on my computer set-up.

ZehMatt:
Been meaning to refactor this part for some time anyway, just never knew it would actually influence the speed as is. I've also commented some code so whoever gets over it understands why there is a secondary update loop in GameState::Update. The game should no longer speed up or slow down at any frame rate.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -53,7 +53,7 @@
 - Fix: [#8469] Crash modifying colour on hacked rides.
 - Fix: [#8508] Underground roto-drop is not going up.
 - Fix: [#8555] Multiplayer window text limits are not computed properly.
-- Fix: [#8591] Run game loop at consistent tick rate of 40hz.
+- Fix: [#8591] Game loop does not run at a consistent tick rate of 40 Hz.
 - Improved: [#2940] Allow mouse-dragging to set patrol area (Singleplayer only).
 - Improved: [#7730] Draw extreme vertical and lateral Gs red in the ride window's graph tab.
 - Improved: [#7930] Automatically create folders for custom content.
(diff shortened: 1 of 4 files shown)


Alicecomma:
Something of interest is that this PR doesn't have the problem where pressing the LMB/RMB at low framerate tanks the game speed considerably (while keeping the same FPS counter value) - as can be seen here (51fc2fe): [Imgur link to gif](https://imgur.com/06hB7kc) (_The circle around the black cursor indicates a mouse click_).

That's pretty amazing because the original RCT2 has the same issue but this PR doesn't have it. Either this is really good or somehow something which was (necessarily) taking a lot of overhead was dropped.. somehow.

distribution/changelog.txt
```diff
@@ -53,6 +53,7 @@
 - Fix: [#8469] Crash modifying colour on hacked rides.
 - Fix: [#8508] Underground roto-drop is not going up.
 - Fix: [#8555] Multiplayer window text limits are not computed properly.
+- Fix: [#8591] Run game loop at consistent tick rate of 40hz.
```
Gymnasiast: This should list the problem that was fixed, rather than the solution.
```suggestion
- Fix: [#8591] Game loop does not run at a consistent tick rate of 40 Hz.
```

PR #8595, opened 2019-01-15
Simplify some code in Vehicle.cpp

janisozaur:
This reduces amount of code that depends on registers struct and removes one goto

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -8239,7 +8239,10 @@ loc_6DAEB9:
 
         if (!vehicle_update_track_motion_forwards_get_new_track(vehicle, trackType, ride, rideEntry))
         {
-            goto loc_6DB94A;
+            _vehicleMotionTrackFlags |= VEHICLE_UPDATE_MOTION_TRACK_FLAG_5;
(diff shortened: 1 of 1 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -9071,10 +9061,8 @@ static int32_t vehicle_update_track_motion_mini_golf(rct_vehicle* vehicle, int32
 
 loc_6DC9BC:
     _vehicleMotionTrackFlags |= VEHICLE_UPDATE_MOTION_TRACK_FLAG_5;
-    regs.eax = vehicle->remaining_distance + 1;
-    _vehicleVelocityF64E0C -= regs.eax;
-    vehicle->remaining_distance -= regs.eax;
-    regs.ebx = vehicle->vehicle_sprite_type;
```
duncanspumpkin: ebx is still used in 6dcd2b

janisozaur: https://github.com/OpenRCT2/OpenRCT2/blob/57471edc47b9fbb8461a60f28c4ad59fb02a316c/src/openrct2/ride/Vehicle.cpp#L9216-L9223

Doesn't look like it is

PR #8599, opened 2019-01-17
Add missing nullptr check in rct_peep::UpdateRideMazePathfinding

janisozaur:
Fixes https://www.nedesigns.com/park/3359/h2h7-r4-diamondheights/

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -4779,7 +4779,12 @@ void rct_peep::UpdateRideMazePathfinding()
 
     } while (!(tileElement++)->IsLastForTile());
 
-    uint16_t mazeEntry = tileElement->AsTrack()->GetMazeEntry();
+    auto trackMazeEntry = tileElement->AsTrack();
+    if (trackMazeEntry == nullptr)
+    {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -4779,7 +4779,12 @@ void rct_peep::UpdateRideMazePathfinding()
 
     } while (!(tileElement++)->IsLastForTile());
 
-    uint16_t mazeEntry = tileElement->AsTrack()->GetMazeEntry();
+    auto trackMazeEntry = tileElement->AsTrack();
+    if (trackMazeEntry == nullptr)
+    {
+        return;
+    }
```
IntelOrca: I suggest doing this from line 4774.
```
auto trackMazeEntry = map_get_track_element_at(actionX, actionY, stationHeight);
if (trackMazeEntry == nullptr)
{
    return;
}
```

IntelOrca:
In Map.cpp:
`TileElement* map_get_track_element_at(int32_t x, int32_t y, int32_t z)` should become:
`TrackElement* map_get_track_element_at(int32_t x, int32_t y, int32_t z)`

And then a nullptr check around the do loop inside that helper.

duncanspumpkin:
Will refactor post release

PR #8600, opened 2019-01-18
[feedback] Add RCT2 obj. dir. to ObjectRepository search path

tomlankhorst:
I followed the installation instructions for MacOS but I ended up with an error about missing objects. After some debugging, I found that the `ObjectRepository` did not look for objects in the original RCT2 `ObjDir` folder.

**Do I correctly assume that objects should be searched in the RCT2 game folder?**

Issue

So, I added it to the repository, solving the issue. Still, it leaves me with some missing textures so something is still not alright on Mac OS <develop>.

Texture

### DIFF
--- a/src/openrct2/object/ObjectRepository.cpp
+++ b/src/openrct2/object/ObjectRepository.cpp
@@ -83,6 +83,7 @@ public:
               std::vector<std::string>{
                   env.GetDirectoryPath(DIRBASE::OPENRCT2, DIRID::OBJECT),
                   env.GetDirectoryPath(DIRBASE::USER, DIRID::OBJECT),
+                  env.GetDirectoryPath(DIRBASE::RCT2, DIRID::OBJECT),
               })
         , _objectRepository(objectRepository)
     {


AaronVanGeffen:
Just a quick note, but this looks to me like you're not using OpenRCT2's JSON objects, which in turn refer to RCT2's.

tomlankhorst:
@AaronVanGeffen thanks for your remark. How should these OpenRCT2 JSON objects be generated, I suppose this is part of the build process?

Gymnasiast:
I believe it's in the `make install` phase.
The original .DAT objects should be in RCT2's `ObjData` dir, while the JSON objects should be in OpenRCT2's `object` dir.

tomlankhorst:
@Gymnasiast thanks, It seems there are some steps missing in the installation guide. 

Most importantly: the object JSONs are downloaded on **installation** so that means that `make install` is an essential step. Then, I had to symlink `(install-prefix)/share/openrct2` to `(install-prefix)/bin/data`... now everything works. 

So, how _should_ the installation process look like? I'll try to update the documentation accordingly.

tomlankhorst:
See https://github.com/OpenRCT2/OpenRCT2/wiki/Building-OpenRCT2-on-macOS-using-CMake

janisozaur:
Is macOS really so special it needs its own page and can't reuse our cmake guide we already have?

IntelOrca:
@janisozaur I would say it is much clearer if it only lists macOS related content. The Linux one is full of distro specific information which you have to wade through.

tomlankhorst:
@janisozaur Well, I won't protest if we merge the Mac OS page with the Linux one and rename it 'Building OpenRCT2 with CMake'. AFAIK there's no way of proposing PRs on Wiki pages so shall I just take the opportunity to do so?

AaronVanGeffen:
Personally, I prefer to keep it by platform, as @IntelOrca suggested.

tomlankhorst:
Sure, np!

janisozaur:
Personally I'd rather see just one cmake page, the distro-specific stuff concerns the packages and just that, from that point of view macOS/brew is just another package manager. But it looks like I'm out-voted on this, so I suppose it'll stay the way it is now.

tomlankhorst:
It would be great to have CMake generate the MSVC project as well... I don't have access to a Windows machine currently but [generating](https://cmake.org/cmake/help/v3.13/generator/Visual%20Studio%2015%202017.html) for MSVC is a possibility: 

```
cmake -G "Visual Studio 15 2017" -A x64
```

janisozaur:
@tomlankhorst perhaps you would be interested in switching our Travis ci job to cmake, now that there's nobody using xcode project and it is a constant pain to maintain.

tomlankhorst:
Sure, does an approach that uses CMake and the clang toolchain right in the Travis 'osx' container suit all needs? I suppose the 'packaging' of the app was currently done by XCode, right?

PR #8604, opened 2019-01-18
Name some path drawing variables

Gymnasiast:


src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -794,7 +798,7 @@ void path_paint(paint_session* session, uint16_t height, const TileElement* tile
 {
     session->InteractionType = VIEWPORT_INTERACTION_ITEM_FOOTPATH;
 
-    bool word_F3F038 = false;
+    bool hasSupports = false;
```
ZehMatt: While in this function context this seems correct, path_paint_box_support uses the parameter name 'hasFences' which seems to be incorrect. I've tried to toggle it to see what happens, this is the result: https://share.epic-domain.com/2019-01-18_20-14-371f4d0f6aeb305ac47.webm

Maybe a more fitting name could be 'usePathRailings', not exactly sure what to call it.

ZehMatt: Yeah pretty much.

Gymnasiast: I'm working on stuff like this in another branch, I'll fix it in there.

PR #8606, opened 2019-01-18
Fix #8588: Guest list scrolling breaks above ~2000 guests

### LINKED ISSUE #8588: guest list
-->
**OS:** Windows 10
**Version:** 0.2.1
**Commit/Build:** 8ac731e

Hello! 
I seem to be encountering a problem when viewing the guests in the Guest Window. When I scroll down the lowest I can go, I cannot see all of my guests. Even when I expand the window, the lowest it will go is to the "R" names. I took a screenshot to show you. I get the same problem when I select individual guest's thoughts as well. 
Any thoughts on what I can do to fix this? 

Thanks so much!
  Fill in the placeholders below. Delete any headings and placeholders that you do not fill in.
volcania 2019-01-14 04-36-04

**Steps to reproduce:**
1. 
2. 

**Dump file**

**Screenshots / Video:**

**Save game:**

tomlankhorst:
Fix #8588

### DIFF
--- a/src/openrct2-ui/windows/GuestList.cpp
+++ b/src/openrct2-ui/windows/GuestList.cpp
@@ -88,6 +88,8 @@ static rct_widget window_guest_list_widgets[] = {
 };
 
 static constexpr const uint8_t SUMMARISED_GUEST_ROW_HEIGHT = SCROLLABLE_ROW_HEIGHT + 11;
+static constexpr const auto GUESTS_PER_PAGE = 2000;
+static constexpr const auto GUEST_PAGE_HEIGHT = GUESTS_PER_PAGE * SCROLLABLE_ROW_HEIGHT;
 
 static void window_guest_list_mouseup(rct_window *w, rct_widgetindex widgetIndex);
 static void window_guest_list_resize(rct_window *w);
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Sounds like an unfortunate side-effect of #6522. This has been in for a while…!

Gymnasiast:
I have tried this myself, even lowering the guests-per-page to 2000, but haven't created a PR because it wouldn't solve the problem. Perhaps you have, but I'd like to test that.

tomlankhorst:
@Gymnasiast I've revised to scroll window height limit too, which was incorrectly set to 10*lines instead of 12*lines.

src/openrct2-ui/windows/GuestList.cpp
```diff
@@ -490,7 +492,7 @@ static void window_guest_list_scrollgetsize(rct_window* w, int32_t scrollIndex,
             }
             w->var_492 = numGuests;
             y = numGuests * SCROLLABLE_ROW_HEIGHT;
-            _window_guest_list_num_pages = (int32_t)std::ceil((float)numGuests / 3173);
+            _window_guest_list_num_pages = 1 + (numGuests - 1) / GUESTS_PER_PAGE;
```
Gymnasiast:
```suggestion
            _window_guest_list_num_pages = ((numGuests - 1) / GUESTS_PER_PAGE) + 1;
```

AaronVanGeffen: Personally, I think the original is clearer. There is, after all, always one page.

Gymnasiast: Meh, not sure if I agree, but it's too minor to argue about. I'll merge it as-is.

tomlankhorst:
@Gymnasiast Thanks for reviewing. I'll take care of the messages ;).

src/openrct2-ui/windows/GuestList.cpp
tomlankhorst: I would agree with reversing the order but I don't think that the redundant parentheses add much to clarity, or was there a different motivation?

Issue #8598, opened 2019-01-17
'Screenshot failed' due to park name character?


Richard-L:
A "screenshot failed" prompt appears when trying to take screenshots on this park: https://www.nedesigns.com/park/1698/60/

The park has an asterisk in its name. Perhaps is that the problem?

janisozaur:
![60*60 2019-01-17 22-12-57](https://user-images.githubusercontent.com/550290/51349284-2e496a00-1aa5-11e9-9dfd-4e5fa35f770b.png)

![60*60 2019-01-17 22-13-01](https://user-images.githubusercontent.com/550290/51349296-343f4b00-1aa5-11e9-8fb1-0b9ea834c93a.png)

Works just fine for me.

Gymnasiast:
@janisozaur That's not terribly helpful -- Linux allows asterisks in file names, and Windows does not.

ocalhoun6:
Yes, it's probably because the park name gets incorporated into the filename for the screenshot, and Windows doesn't allow * in file names.

Seems like it shouldn't be too hard to add a bit of code into the screenshot process to strip special characters from the file name before saving.

tomlankhorst:
Currently, there's some Windows specific [filename](https://github.com/OpenRCT2/OpenRCT2/blob/92412d4bab085fe5adbdac80f67af4a50d2fab98/src/openrct2/interface/Screenshot.cpp#L118) processing. Could extend that for more general sanitation and share that functionality with other parts of the code that interface with the filesystem?

bdupree5:
Hello, me and my team would like to work on this issue, we will continue to update you as we continue our work.

janisozaur:
There's PR already open against this issue: https://github.com/OpenRCT2/OpenRCT2/pull/8611

bdupree5:
Thank you for letting me know.

PR #8611, opened 2019-01-19
Check screenshot filename

tomlankhorst:
Fix for #8598

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -25,6 +25,7 @@
 - Fix: [#8219] Faulty folder recreation in "save" folder.
 - Fix: [#8507] Incorrect change in vehicle rolling direction.
 - Fix: [#8537] Imported RCT1 rides/shops are all numbered 1.
+- Fix: [#8598] Taking screenshots fails with some park names.
 - Fix: [#8649] Setting date does not work in multiplayer.
 - Fix: [#8873] Potential crash when placing footpaths.
 - Fix: [#8882] Submarine Ride does not count as indoors (original bug).
@@ -115,8 +116,7 @@
 - Fix: [#8585] Part of track missing on air powered vertical coaster.
 - Fix: [#8588] Guest list scrolling breaks above ~2000 guests.
 - Fix: [#8591] Game loop does not run at a consistent tick rate of 40 Hz.
-- Fix: [#8647] Marketing campaigns check for entry fees below £1 (original bug). 
-- Fix: [#8598] Sanitize screenshot parknames.
+- Fix: [#8647] Marketing campaigns check for entry fees below £1 (original bug).
 - Fix: [#8653] Crash when peeps attempt to enter a ride with no vehicles.
 - Fix: [#8720] Desync due to boats colliding with ghost pieces.
 - Fix: [#8736] Incomplete warning when all ride slots are full.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -116,6 +116,7 @@
 - Fix: [#8588] Guest list scrolling breaks above ~2000 guests.
 - Fix: [#8591] Game loop does not run at a consistent tick rate of 40 Hz.
 - Fix: [#8647] Marketing campaigns check for entry fees below £1 (original bug). 
+- Fix: [#8598] Sanitize screenshot parknames.
 - Fix: [#8653] Crash when peeps attempt to enter a ride with no vehicles.
 - Fix: [#8720] Desync due to boats colliding with ghost pieces.
 - Fix: [#8736] Incomplete warning when all ride slots are full.
--- a/src/openrct2/interface/Screenshot.cpp
+++ b/src/openrct2/interface/Screenshot.cpp
@@ -17,6 +17,7 @@
 #include "../audio/audio.h"
 #include "../core/Console.hpp"
 #include "../core/Imaging.h"
+#include "../core/Optional.hpp"
 #include "../drawing/Drawing.h"
 #include "../drawing/X8DrawingEngine.h"
 #include "../localisation/Localisation.h"
@@ -28,10 +29,13 @@
 #include "../world/Surface.h"
 #include "Viewport.h"
 
+#include <cctype>
 #include <chrono>
 #include <cstdlib>
 #include <memory>
+#include <string>
 
+using namespace std::literals::string_literals;
 using namespace OpenRCT2;
 using namespace OpenRCT2::Drawing;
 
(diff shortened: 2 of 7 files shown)


src/openrct2/interface/Screenshot.cpp
```diff
@@ -143,6 +129,12 @@ static int32_t screenshot_get_next_path(char* path, size_t size)
         fileNameCh, leftBytes, "%s %d-%02d-%02d %02d-%02d-%02d.png", park_name, currentDate.year, currentDate.month,
         currentDate.day, currentTime.hour, currentTime.minute, currentTime.second);
 
+    if (!filename_valid_characters(fileNameCh))
```
ZehMatt: This isn't exactly a fix if you just disallow the saving of screenshots due to the park name, this replaces one error with another. I think it would be better to simply sanitize the filename that will work across all platforms The screenshot file name does not need the exact park name in it as long it can be easily identified still.

ZehMatt: Please also have a look at https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

IntelOrca: The original code sanitised colons which you have now removed. It should instead replace all invalid characters with a valid character like underscore or hyphen. Also note that `filename_valid_characters` is not a platform function so each OS would end up sanitising the same characters regardless of their support.

tomlankhorst: Thanks for your comments. Taking them into account, I am thinking of adding platform functions for determining the valid characters. Maybe analogous to [Microsoft's](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getinvalidpathchars?view=netframework-4.7.2). 

Any resources on the utf8 string classes by the way ?

janisozaur: You mean this: https://github.com/OpenRCT2/OpenRCT2/blob/7de115263e32a3b1daf1aa5bbc81479aa58e278d/src/openrct2/common.h#L27-L28 ?

IntelOrca: @tomlankhorst You should use `std::string_view` where possible.

tomlankhorst:
That's funny. ;)

I will look into using string views!

tomlankhorst:
Oops... XCode 9 has <optional> in experimental.

tomlankhorst:
> I would really like to be able to use `std::optional`, is there a minimum xcode version that has it in experimental?

It's in 9 as experimental. But that needs some playing around with the `std::experimental`  namespace. In 10 it's stable AFAIK.

AaronVanGeffen:
Perhaps we can apply a technique similar to [m-ou-se/mstd](https://github.com/m-ou-se/mstd) here.

tomlankhorst:
> Perhaps we can apply a technique similar to [m-ou-se/mstd](https://github.com/m-ou-se/mstd) here.

Instead of `std::optional`?

AaronVanGeffen:
> Instead of `std::optional`?

No, use a wrapper around it, like [optional.hpp](https://github.com/m-ou-se/mstd/blob/6b48845d88c41bb698c53d8c07f4ccf0c2f41286/include/mstd/optional.hpp).

tomlankhorst:
@AaronVanGeffen , ah, well, that's what the current solution boils down too as well :)
https://github.com/OpenRCT2/OpenRCT2/blob/c5927e3cdec7b2dd04f04e241d8b8daa9e4e711e/src/openrct2/core/Optional.hpp

AaronVanGeffen:
Ah, nice work! I hadn't realised a solution had already made it in. (About 9 hours ago, apparently, so I don't feel too bad about this. 😉)

AaronVanGeffen:
@ZehMatt Are you satisfied with this PR in its current state?

ZehMatt:
Very 👍 , errors are never cool

src/openrct2/interface/Screenshot.cpp
```diff
@@ -93,98 +100,110 @@ static void screenshot_get_rendered_palette(rct_palette* palette)
...
 {
-    char screenshotPath[MAX_PATH];
+    char buffer[128];
+    format_string(buffer, sizeof(buffer), gParkName, &gParkNameArgs);
+    std::string name(buffer);
+
+    // Only allow alpha numeric or whitelisted characters
+    auto isDirtyChar = [](auto& c) {
+        constexpr auto& valid = validScreenshotCharacters;
+        return !(std::isalnum(c) || std::find(valid.begin(), valid.end(), c) != valid.end());
```
AaronVanGeffen: What about CJK characters? If the park is running in e.g. Korean or Japanese, the park name won't be alphanumeric.

AaronVanGeffen: I created a scenario named テスト, resulting in filenames like `  2019-02-07 19-05-57.png` — note the two leading spaces. To me, this isn't really a desireable outcome.

tomlankhorst: I'll try again with the[ localized version of `isalnum` ](https://en.cppreference.com/w/cpp/locale/isalpha)

tomlankhorst:
Do we by any chance support Windows systems with FAT partitions? 
... because their filenames are limited to strings of characters in [code page 437](https://en.wikipedia.org/wiki/Code_page_437).

ZehMatt:
Its possible if you run OpenRCT2 on a USB stick formatted with FAT32, its rather unlikely but you aren't restricted to NTFS.

Patrik356b:
How about exfat?

tomlankhorst:
[About the same](https://en.wikipedia.org/wiki/ExFAT) as NTFS, it seems:
UTF-16 minus `: / \ < > * `.

Op ma 11 feb. 2019 om 04:58 schreef TheExileFox <notifications@github.com>:

ZehMatt:
Friendly ping, whats the current status?

tomlankhorst:
If we want to have a single solution for all combinations of OSses, locales and filesystems, we'll be pretty much doomed to ASCII or to a complex solution that figures out which code page is used or whether UTF-8/16 is used, which characters are allowed on the FS, how these characters should be replaced, etc. 

So, I'm struggling a bit to find the appropriate balance here. 

Maybe we should make a platform function that determines whether the FS supports UTF. 
Then, another function that determines prohibited characters in filenames. 
Then one for directories. 
Then prohibited names. 

I would love to have your opinion on this.

IntelOrca:
1. Stick with UTF-8 strings.
2. Use a platform function, `sanitise_filename`.

Then for now just do a necessary implementation for Windows which replaces the invalid chars with underscore or something.

tomlankhorst:
Current implementation:

- Introduce `platform_sanitise_filename`, replacing forbidden characters, not taking reserved filenames into account. 
- The idea is that typical characters that the user supplies are on the codepage available to their Windows version, if not, they should use a modern FS...

src/openrct2/interface/Screenshot.cpp
```diff
@@ -93,98 +97,103 @@ static void screenshot_get_rendered_palette(rct_palette* palette)
...
-    {
-        log_error("Unable to generate a screenshot filename.");
-        return -1;
-    }
-    const size_t leftBytes = size - strlen(path);
+    log_error("Unable to save screenshots in OpenRCT2 screenshot directory.\n");
+    return opt::nullopt;
+}
+
+static std::tuple<rct2_date, rct2_time> screenshot_get_date_time()
```
ZehMatt: With two values std::pair is recommended.

src/openrct2/interface/Screenshot.cpp
```diff
@@ -93,98 +97,103 @@ static void screenshot_get_rendered_palette(rct_palette* palette)
...
-static int32_t screenshot_get_next_path(char* path, size_t size)
+static std::string screenshot_get_park_name()
 {
-    char screenshotPath[MAX_PATH];
+    char buffer[512];
+    format_string(buffer, sizeof(buffer), gParkName, &gParkNameArgs);
+    return buffer;
+}
 
+static opt::optional<std::string> screenshot_get_directory()
```
ZehMatt: I don't think the result should be optional, it should always return the full directory path even if it doesn't exist, the caller should make sure afterwards that the directory exists, this makes the code complex.

tomlankhorst:
> There are quite a few unrelated changes in this PR like the optional part which I'm not really a fan of to be honest. Would be nice if someone else could also have a look.

The good news is that <optional> is supported in the latest xcode stdlib!

src/openrct2/platform/Shared.cpp
```diff
@@ -198,6 +199,24 @@ uint8_t platform_get_currency_value(const char* currCode)
     return CURRENCY_POUNDS;
 }
 
+#ifndef _WIN32
+std::string platform_sanitise_filename(const std::string& path)
+{
+    auto sanitised = path;
+
+    std::vector<std::string::value_type> prohibited = { '/' };
```
ZehMatt: Feels off to have a single element vector with dynamic memory allocation, too bad theres no std::make_array (yet), could at least make this const.

src/openrct2/platform/Windows.cpp
```diff
@@ -253,6 +256,24 @@ std::string platform_get_rct2_steam_dir()
     return "Rollercoaster Tycoon 2";
 }
 
+std::string platform_sanitise_filename(const std::string& path)
+{
+    auto sanitised = path;
+
+    std::vector<std::string::value_type> prohibited = { '<', '>', '*', '\\', ':', '|', '?', '"', '/' };
```
ZehMatt: Should be at least const.

AaronVanGeffen:
Please clean up the commit history for this PR before merging (or just squash-merge it).

ZehMatt:
Seems to be working, I get a screenshot now named 'テストテスト 2019-05-11 13-19-58.png', so after the requested changes this can be merged.

distribution/changelog.txt
```diff
@@ -114,6 +114,7 @@
 - Fix: [#8588] Guest list scrolling breaks above ~2000 guests.
 - Fix: [#8591] Game loop does not run at a consistent tick rate of 40 Hz.
 - Fix: [#8647] Marketing campaigns check for entry fees below £1 (original bug). 
+- Fix: [#8598] Sanitize screenshot parknames.
```
Gymnasiast:
```suggestion
- Fix: [#8598] Sanitise screenshot parknames.
```

tomlankhorst:
Thanks guys :)

PR #8615, opened 2019-01-20
Cleanup/path drawing

Gymnasiast:
This is a followup to #8541, #8604 and a discussion via DMs with @IntelOrca .

It further continues the split of footpaths and supports/railings, moving the decision whether to draw a path over the supports to the tile element (which currently still uses the old method to check it, but that should be addressed in the new file format).


src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -739,7 +740,7 @@ static void sub_6A3F61(
         // Redundant zoom-level check removed
 
         sub_6A4101(
-            session, tile_element, height, connectedEdges, word_F3F038, footpathEntry, footpathEntry->image | imageFlags,
+            session, tile_element, height, connectedEdges, word_F3F038, railingEntry, railingEntry->image | imageFlags,
```
Gymnasiast:
```suggestion
            session, tile_element, height, connectedEdges, word_F3F038, railingEntry, footpathEntry->image | imageFlags,
```
Signature should be changed accordingly.

src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -557,23 +558,23 @@ static void sub_6A4101(
                 sub_98197C(
                     session, 77 + base_image_id, 28, 0, 1, 28, 7, height, 28, 4,
                     height + 2); // bound_box_offset_y seems to be a bug
-                if (!(dword_F3EF80 & 0x10))
+                if (!(drawnEdges & 0x10))
```
IntelOrca: These literals should be changed to named flags.

PR #8618, opened 2019-01-21
Refactor footpath_update_path_wide_flags

ZehMatt:
Minor cleanup.

### DIFF
--- a/src/openrct2/world/Footpath.h
+++ b/src/openrct2/world/Footpath.h
@@ -139,6 +139,18 @@ enum
     FOOTPATH_CORNER_3 = (1 << 3),
 };
 
+enum
+{
+    FOOTPATH_CONNECTION_S = (1 << 0),
+    FOOTPATH_CONNECTION_NE = (1 << 1),
+    FOOTPATH_CONNECTION_W = (1 << 2),
+    FOOTPATH_CONNECTION_SE = (1 << 3),
+    FOOTPATH_CONNECTION_N = (1 << 4),
+    FOOTPATH_CONNECTION_SW = (1 << 5),
+    FOOTPATH_CONNECTION_E = (1 << 6),
+    FOOTPATH_CONNECTION_NW = (1 << 7),
+};
+
 extern uint8_t gFootpathProvisionalFlags;
 extern LocationXYZ16 gFootpathProvisionalPosition;
 extern uint8_t gFootpathProvisionalType;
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Footpath.h
```diff
@@ -139,6 +139,18 @@ enum
     FOOTPATH_CORNER_3 = (1 << 3),
 };
 
+enum
+{
+    FOOTPATH_CONNECTION_SE_SW = (1 << 0),
```
Broxzier: Since these flags are used for path elements directly north/east/south/west of the paths (and not SE _and_ SW for example), I'd rather see them named like:

```suggestion
    FOOTPATH_CONNECTION_S = (1 << 0),
```

src/openrct2/world/Footpath.h
```diff
@@ -139,6 +139,18 @@ enum
...
+    FOOTPATH_CONNECTION_NE = (1 << 1),
+    FOOTPATH_CONNECTION_SW_NW = (1 << 2),
```
Broxzier:
```suggestion
    FOOTPATH_CONNECTION_W = (1 << 2),
```

src/openrct2/world/Footpath.h
```diff
@@ -139,6 +139,18 @@ enum
...
+    FOOTPATH_CONNECTION_SE = (1 << 3),
+    FOOTPATH_CONNECTION_NE_NW = (1 << 4),
```
Broxzier:
```suggestion
    FOOTPATH_CONNECTION_N = (1 << 4),
```

src/openrct2/world/Footpath.h
```diff
@@ -139,6 +139,18 @@ enum
...
+    FOOTPATH_CONNECTION_SW = (1 << 5),
+    FOOTPATH_CONNECTION_NE_SE = (1 << 6),
```
Broxzier:
```suggestion
    FOOTPATH_CONNECTION_E = (1 << 6),
```

src/openrct2/world/Footpath.h
Broxzier: Diagonal could be seen differently depending on your perspective. Just to clarify:
image

Gymnasiast: That's not how *any* of the other code works, or tools like the Object Editor for that matter. Orthogonal directions are NE/SE/SW/NW, diagonal directions are N/E/S/W.

Broxzier: It is how footpath's edges and corners work, park fences, and probably more. I have not seen it the other way around, at least not for tile elements.

Issue #8597, opened 2019-01-16
Multiplayer server using all available memory


Naemuti:
**OS:** Ubuntu
**Version:** 18.04.1 LTS
**Commit/Build:** 51fc2fe

Starting an OpenRCT2 server running headless causes system to run out of memory after about 10 minutes using a brand new save with nothing built, no guests, and no clients connected. The server is running on an AWS EC2 t2.micro instance with 1 GB of RAM. This happens with both openrct2 and openrct2-cli

- [ ] Reproducible in RCT2 (vanilla)?
- [x] Specific to multiplayer?

**Steps to reproduce:**
1. Start the OpenRCT2 service  hosting a server in headless mode
2. Wait about 10 minutes, the process will steadily consume more memory until almost none is left

**Dump file**

**Screenshots / Video:**

![image](https://user-images.githubusercontent.com/648473/51282034-1a710b80-19b2-11e9-8c41-c9a25192109e.png)
![image](https://user-images.githubusercontent.com/648473/51282489-3b862c00-19b3-11e9-85f3-c6ee4636d606.png)

**Save game:**
[save.zip](https://github.com/OpenRCT2/OpenRCT2/files/2766208/save.zip)

tomlankhorst:
I wasn't able to reproduce this. Could you maybe run your server through [Valgrind](http://valgrind.org/docs/manual/quick-start.html) and send the resulting logs? 

```bash
apt-get install valgrind
valgrind --leak-check=yes ./openrct host ............
```

IntelOrca:
@Naemuti thanks, there definitely memory leaks there that need fixing, e.g. the `json_dumps` in `Server_Send_GAMEINFO`

tomlankhorst:
@IntelOrca but the most problematic leak's here right? 
```
==11907== 110,705,728 bytes in 1,729,777 blocks are definitely lost in loss record 12 of 12
==11907==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11907==    by 0x3092B0: InteractiveConsole::Execute(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) (in /usr/bin/openrct2)
==11907==    by 0x30FF64: StdInOutConsole::ProcessEvalQueue() (in /usr/bin/openrct2)
==11907==    by 0x2BCF63: OpenRCT2::Context::Launch() (in /usr/bin/openrct2)
[...]

PR #8619, opened 2019-01-21
Free allocated memory when 'argc' == 0

tomlankhorst:
Fixes leak in #8597, InteractiveConsole allocates memory for arguments,
but frees it only if the argument count is greater than zero.


IntelOrca:
This should have a changelog entry but since there are still more memory leaks to fix before #8597 can be closed, we can do that later.

PR #8620, opened 2019-01-21
Use vector<string> in InteractiveConsole

tomlankhorst:
Taking #8619 a step further

Issue #8597 shows Valgrind result indicating memory leak in
InteractiveConsole. This is fixed in #8619 by ensuring that free() is
called appropriately.
This commit takes that a step further by removing all manual memory
management in favour of using vector<string>.

- argc is now gone in favour of argv.empty()/argv.size()
- argv_t is a vector of strings of utf8's
- console_command_func's signature is changed accordingly

### DIFF
--- a/src/openrct2/interface/InteractiveConsole.cpp
+++ b/src/openrct2/interface/InteractiveConsole.cpp
@@ -59,7 +59,7 @@
 
 using arguments_t = std::vector<std::string>;
 
-static constexpr const utf8* ClimateNames[] = {
+static constexpr const char* ClimateNames[] = {
     "cool_and_wet",
     "warm",
     "hot_and_dry",
@@ -67,9 +67,6 @@ static constexpr const utf8* ClimateNames[] = {
 };
 
 static void console_write_all_commands(InteractiveConsole& console);
-static int32_t console_parse_int(const utf8* src, bool* valid);
-static double console_parse_double(const utf8* src, bool* valid);
-
 static int32_t cc_variables(InteractiveConsole& console, const arguments_t& argv);
 static int32_t cc_windows(InteractiveConsole& console, const arguments_t& argv);
 static int32_t cc_help(InteractiveConsole& console, const arguments_t& argv);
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This is definitely the better approach. Nice job, I like it! :+1:

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -51,11 +51,14 @@
 #include <exception>
 #include <string>
 #include <thread>
+#include <vector>
 
 #ifndef NO_TTF
 #    include "../drawing/TTF.h"
 #endif
 
+using argv_t = std::vector<std::basic_string<utf8>>;
```
ZehMatt: std::basic_string<utf8> is currently identical to std::string. I'm not sure why we have a typedef utf8 which is essentially char right now. std::string should be able to carry UTF8 generally without a special storage type. So you might as well just use the plain std::string type.

tomlankhorst: So, don't we want to get rid of the utf8 typedef at all then?

IntelOrca: We do, eventually everything will be `std::string` or `std::string_view`.

tomlankhorst: Good to know!

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1714,40 +1708,28 @@ void InteractiveConsole::Execute(const std::string& s)
...
-    if (argc == 0)
+    if (argv.empty())
         return;
 
     bool validCommand = false;
+
     for (const auto& c : console_command_table)
     {
-        if (strcmp(argv[0], c.command) == 0)
+        if (argv[0] == std::string(c.command))
```
ZehMatt: std::string construction is not required here, it can compare against const char*

ZehMatt: std::string construction is not required here, it can compare against const char*

tomlankhorst: That's very true, I didn't catch that.

ZehMatt:
Nice cleanup, good job.

tomlankhorst:
Well, it's the `-Wmissing-declarations` flag. Will add some unuseful prototypes then ;).

PR #8374, opened 2018-12-06
Add support to record and replay game commands/actions.

ZehMatt:
This is something I wanted to do for a longer time now. Basically this records all game actions and game commands and replays it back.

Small preview: https://share.epic-domain.com/2018-12-06_06-30-281437ce2a04a2d1077.webm

PR #7446, opened 2018-04-25
Turn off codecov comments on PRs

janisozaur:
See https://docs.codecov.io/docs/pull-request-comments
and https://github.com/codecov/support/wiki/Codecov-Yaml

PR #6975, opened 2018-01-07
handle some nullptrs gracefully

janisozaur:
As noted in https://github.com/OpenRCT2/OpenRCT2/pull/6973: no-PIE builds can trigger null dereferences. ~~This addresses some of them.~~

This now addresses all of them and to prove, it now also contains the commit from #6973

PR #6973, opened 2018-01-07
Set build type to release on MinGW job

janisozaur:
MinGW is currently our no-PIE build and building it in release mode may
step on some nullptr dereferences. The same can be encountered when
disabling PIE for other modes.

PR #8621, opened 2019-01-21
Name entries in Travis' build matrix

tomlankhorst:
Names the entries in the build matrix, easier identification of CI issues.

Current: , new: 

Note that I also removed the job type check in `after_success`:

```yaml
- if [[ "$TRAVIS_EVENT_TYPE" == "cron" ]] ; then exit 0 ; fi	
```
because there's already an ` if: type != cron` in each entry.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ matrix:
           services:
             - docker
         - os: osx
-          name: macOS Clang Release
+          name: macOS Clang Xcode
           if: type != cron
           osx_image: xcode9.3
           env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,9 +70,9 @@ matrix:
           if: type != cron
           env: OPENRCT2_CMAKE_OPTS="-G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=OpenRCT2 -DCMAKE_BUILD_TYPE=Debug" TARGET=ubuntu_amd64
         - os: linux
-          name: Windows i686-w64 MinGW RelWithDebInfo
+          name: Windows i686-w64 MinGW
           if: type != cron
-          env: OPENRCT2_CMAKE_OPTS="-G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DFORCE32=on -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo" TARGET=windows
+          env: OPENRCT2_CMAKE_OPTS="-G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DFORCE32=on -DBUILD_SHARED_LIBS=ON" TARGET=windows
         - os: linux
           name: Docker64 GCC Release
           if: type != cron
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ matrix:
           if: type != cron
           env: OPENRCT2_CMAKE_OPTS="-G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DFORCE32=on -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo" TARGET=windows
         - os: linux
-          name: Docker64 Release
+          name: Docker64 GCC Release
           if: type != cron
           env: TARGET=docker64 OPENRCT2_CMAKE_OPTS="-G Ninja -DDISABLE_TTF=ON -DCMAKE_BUILD_TYPE=release -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=OpenRCT2"
           services:
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
               else curl --progress-bar --upload-file openrct2-linux.tar.gz https://transfer.sh/openrct2-linux-x86_64.tar.gz -o link && cat link;
               fi
         - os: linux
-          name: Ubuntu i686 GCC Debug
+          name: Ubuntu i686 GCC Debug m32 no-build-shared-libs
           if: type != cron
           services:
             - docker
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This is a very welcome change.

I've noticed you've changed the release type for a few jobs. Were those changes intentional? If so, could you elaborate on them a little?

.travis.yml
```diff
@@ -68,36 +66,40 @@ matrix:
...
             - docker
         - os: linux
+          name: Docker64 GCC Debug disable-opengl
           if: type != cron
-          env: OPENRCT2_CMAKE_OPTS="-G Ninja -DDISABLE_OPENGL=ON -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=OpenRCT2" TARGET=docker64
+          env: OPENRCT2_CMAKE_OPTS="-G Ninja -DDISABLE_OPENGL=ON -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=OpenRCT2 -DCMAKE_BUILD_TYPE=debug" TARGET=docker64
           services:
             - docker
         - os: osx
+          name: macOS Clang Release
```
AaronVanGeffen: Could you clarify that this is building through Xcode as well? (At least, it still is at the moment.)

tomlankhorst:
@AaronVanGeffen I wanted to be explicit about the build type to avoid confusion. But seemingly, there's more involved as the CI behaves differently now and fixing this is not the scope of this PR. So, I reverted changing the build types and just added the names (and mention Xcode).

My reasoning was that all uploaded assets should be built as `RelWithDebInfo` to have a performant app with useful stack traces, while internal tests and coverage should be built as `Debug` to allow easy fixing of problems.

tomlankhorst:
Btw, why was https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/482428042 failing? Didn't effectively change anything. Log overflow?

ZehMatt:
Name tag is highly welcome, appreciated.

janisozaur:
For mingw's warnings see https://github.com/OpenRCT2/OpenRCT2/pull/6975 and https://github.com/OpenRCT2/OpenRCT2/pull/6973.

I'd rather not change the default build type, there's no need for it in most builds (only in coverage one) and it only pollutes the cache and slows down compilation. It may not be that much, but it's still more work than needed of the compiler and related I/O.

> Btw, why was https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/482428042 failing? Didn't effectively change anything. Log overflow?

You can expand the log in question and right at the bottom you will find:
```
$ if [[ "z$OPENRCT2_ORG_TOKEN" != "z" && "$TRAVIS_PULL_REQUEST" == "false" && ("${TRAVIS_BRANCH}" =~ ^(develop|push/) || "z${TRAVIS_TAG}" != "z") ]] ; then curl -o - -v --form "key=$OPENRCT2_ORG_TOKEN" --form "fileName=OpenRCT2-${OPENRCT2_VERSION}${FILENAME_PART}-linux-i686.tar.gz" --form "version=${OPENRCT2_VERSION}" --form "gitHash=$TRAVIS_COMMIT" --form "gitBranch=$PUSH_BRANCH" --form "flavourId=4" --form "file=@openrct2-linux.tar.gz" "https://openrct2.org/altapi/?command=push-build"; else curl --progress-bar --upload-file openrct2-linux.tar.gz https://transfer.sh/openrct2-linux-i686.tar.gz -o link && cat link; fi


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

That's because of https://transfer.sh instability, a workaround to that would be to add a timeout to curl. In majority of cases we don't really need them but it's good to have the ability to provide such one-off builds to users so we can verify it fixes the problem.

The coverage is collected from PRs as well, that's why we can have reports such as https://github.com/OpenRCT2/OpenRCT2/pull/8374#issuecomment-447539719 and https://github.com/OpenRCT2/OpenRCT2/pull/7444#issuecomment-383903071 (disabled in #7446).

.travis.yml
```diff
@@ -68,36 +66,40 @@ matrix:
...
           if: type != cron
           env: OPENRCT2_CMAKE_OPTS="-G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DFORCE32=on -DBUILD_SHARED_LIBS=ON" TARGET=windows
         - os: linux
+          name: Docker64 GCC Release
           if: type != cron
           env: TARGET=docker64 OPENRCT2_CMAKE_OPTS="-G Ninja -DDISABLE_TTF=ON -DCMAKE_BUILD_TYPE=release -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=OpenRCT2"
           services:
             - docker
         - os: linux
+          name: Docker64 GCC Debug coverage
```
janisozaur: You mention `disable-opengl` in other job, why omit network information then?

.travis.yml
```diff
@@ -68,36 +66,40 @@ matrix:
...
         - os: linux
+          name: Ubuntu amd64 Clang
           if: type != cron
           env: OPENRCT2_CMAKE_OPTS="-G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DBUILD_SHARED_LIBS=ON -DCMAKE_INSTALL_PREFIX=OpenRCT2" TARGET=ubuntu_amd64
         - os: linux
+          name: Windows i686-w64 MinGW
           if: type != cron
           env: OPENRCT2_CMAKE_OPTS="-G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DFORCE32=on -DBUILD_SHARED_LIBS=ON" TARGET=windows
         - os: linux
+          name: Docker64 GCC Release
```
janisozaur: TTF?

janisozaur:
There's on job that builds in `release` build type, that's because with higher optimizations GCC can see through more code and it could lead to users running into compilation errors (https://github.com/OpenRCT2/OpenRCT2/pull/4497, https://github.com/OpenRCT2/OpenRCT2/pull/4409#discussion_r79278121)

.travis.yml
tomlankhorst: Added

.travis.yml
tomlankhorst: Added

PR #8622, opened 2019-01-21
Include ICU include dirs in testpaint

tomlankhorst:
`testpaint` uses ICU headers, so add the ICU_INCLUDE_DIR to the target.


AaronVanGeffen:
This surprises me a little, as the original game assets don't require UCI to draw. I suppose it's due to the banner code importing the TrueType path..? Fair catch, either way.

tomlankhorst:
@AaronVanGeffen But it's in `openrct2/core/String.hpp`, right?

janisozaur:
Looks good, but wanted to confirm UCI is not a typo?

tomlankhorst:
@janisozaur It's ICU, certainly!

PR #8458, opened 2018-12-15
Add sprite sorting benchmark

janisozaur:
See #8442

PR #8321, opened 2018-11-30
Add "simulate" command.

ZehMatt:
Before I forget about it again. 
This a basic command to simulate the game for a certain amount of ticks resulting the same network hash we use to identify desyncs. I used this to compare game logic across multiple builds and perhaps someone could actually put that to the CI to see if a PR will actually change the game logic and warn us about it.

PR #8624, opened 2019-01-22
[WIP] Leverage `constexpr` for compile-time evaluation and improved performance

tomlankhorst:
I've run `callgrind` on `develop` and listed the called methods from most frequently called to least. 
Much time is spent on getter methods of tiles, so I started there. 

These are some early experiments that I wanted to run on CI ;).

### DIFF
--- a/src/openrct2/rct12/RCT12.cpp
+++ b/src/openrct2/rct12/RCT12.cpp
@@ -16,21 +16,6 @@
 #include "../world/TileElement.h"
 #include "../world/Wall.h"
 
-uint8_t RCT12TileElementBase::GetType() const
-{
-    return this->type & TILE_ELEMENT_TYPE_MASK;
-}
-
-uint8_t RCT12TileElementBase::GetDirection() const
-{
-    return this->type & TILE_ELEMENT_DIRECTION_MASK;
-}
-
-bool RCT12TileElementBase::IsLastForTile() const
-{
-    return (this->flags & TILE_ELEMENT_FLAG_LAST_TILE) != 0;
-}
-
 uint8_t RCT12SurfaceElement::GetSlope() const
 {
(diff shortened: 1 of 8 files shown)


ZehMatt:
I think half of the things marked constexpr will never be compile time since it has to still read dynamic memory in most cases. While there is no downside given that you still get a runtime version if it can't be evaluated compile time I'm not sure its actually worth doing it.

ZehMatt:
I think the only thing you could benefit from is inline optimizations which you managed by moving all the getters/setters into the header which is good, been meaning to do that for a while, so you could perhaps still go on and focus less on constexpr as you probably won't be that lucky with those, but as for interlining methods, go for it.

janisozaur:
I'm not in favour of that. This is LTO's job, not developer's.

ZehMatt:
> I'm not in favour of that. This is LTO's job, not developer's.

Sorry but I strongly disagree here. Its the typical "Leave it to the optimization pass" which usually brings bad code out there only because there is too much hope in it. Also LTO is not a standard thing either, optimization is usually per compilation unit, whole program optimization only works if you have LTCG. 

I mean strictly speaking why even bother introducing inline keyword or have it inlined if implemented in a header if you could always leave that to the optimizer?

tomlankhorst:
With some _back of the envelope_ calculations, Valgrind showed that putting the simple getters in headers and mark them implicitly as `inline` through `constexpr` results in each game `update` call requiring about 10% fewer CPU cycles when building `RelWithDebInfo` on GCC. Sure, this metric is hardly trustworthy due to between-run differences in the state but I think there's still quite a bit to gain here.

janisozaur:
https://github.com/OpenRCT2/OpenRCT2/pull/8458 adds benchmark (read: both the library and code using it). We already have `simulate` command via https://github.com/OpenRCT2/OpenRCT2/pull/8321 it sounds like another benchmark is needed.

PR #8614, opened 2019-01-20
Fix #8584: Ducks spawning function doesnt check tiles 0..63 (original bug)

deurklink:
Originally, the function that checks for water on tiles to spawn ducks only generated coordinates from 64 to 191. This meant that in small maps like Leafy Lake or Forest Frontiers, ducks will normally not spawn, or only in a corner of the map if all land is owned.

I tested this in vanilla RCT2 on Crazy Castle, and it seems ducks only stick to a small corner of the map with coordinates bigger than 63,63 as well, so it seems to be an original bug (or intended behavior).

This pull modifies the function to consider every tile coordinate from 0 to 255. I tested this on several maps with water inside the park area, and they now all spawn a good amount of ducks every year.

PR #8625, opened 2019-01-22
Put boundaries on target coordinates for ducks 

deurklink:
Should fix the crashes caused by #8614

-Lowest tile coordinate is 1
-Highest tile coordinate is 254
-Create duck is called with 2 tiles of randomness.
-Create duck function itself also adds a random offset of maximum 30 (a little bit less than 1 tile)

If we limit the generated coordinates to be between 4 and 251, we should always get a valid tile.

### DIFF
--- a/src/openrct2/scenario/Scenario.cpp
+++ b/src/openrct2/scenario/Scenario.cpp
@@ -432,10 +432,10 @@ static int32_t scenario_create_ducks()
     // To spawn ducks safely, the generated coordinates should not be in the outer 3 tiles of the map.
 
     r = scenario_rand();
-    x = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE +
-        ((r >> 16) % (MAXIMUM_MAP_SIZE_PRACTICAL - (DUCK_TARGET_TILES_AWAY_FROM_EDGE * 2)));
-    y = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE +
-        ((r & 0xFFFF) % (MAXIMUM_MAP_SIZE_PRACTICAL - (DUCK_TARGET_TILES_AWAY_FROM_EDGE * 2)));
+    x = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE
+        + ((r >> 16) % (MAXIMUM_MAP_SIZE_PRACTICAL - (DUCK_TARGET_TILES_AWAY_FROM_EDGE * 2)));
+    y = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE
+        + ((r & 0xFFFF) % (MAXIMUM_MAP_SIZE_PRACTICAL - (DUCK_TARGET_TILES_AWAY_FROM_EDGE * 2)));
     x = x * 32;
     y = y * 32;
 
--- a/src/openrct2/scenario/Scenario.cpp
+++ b/src/openrct2/scenario/Scenario.cpp
@@ -427,10 +427,15 @@ static int32_t scenario_create_ducks()
     int32_t i, j, r, c, x, y, waterZ, centreWaterZ, x2, y2;
 
     // Originally, this function generated coordinates from 64 to 191.
-    // It now generates coordinates from 0 to 255, so smaller maps may also spawn ducks.
+    // The duck create function is called with 2 tiles of randomness to the target coordinate.
+    // The duck create function itself adds approximately 1 tile of randomness.
+    // To spawn ducks safely, the generated coordinates should not be in the outer 3 tiles of the map.
+
     r = scenario_rand();
-    x = ((r >> 16) & 0xFFFF) % MAXIMUM_MAP_SIZE_TECHNICAL;
-    y = (r & 0xFFFF) % MAXIMUM_MAP_SIZE_TECHNICAL;
+    x = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE +
+        ((r >> 16) % (MAXIMUM_MAP_SIZE_PRACTICAL - (DUCK_TARGET_TILES_AWAY_FROM_EDGE * 2)));
+    y = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE +
+        ((r & 0xFFFF) % (MAXIMUM_MAP_SIZE_PRACTICAL - (DUCK_TARGET_TILES_AWAY_FROM_EDGE * 2)));
     x = x * 32;
     y = y * 32;
 
--- a/src/openrct2/scenario/Scenario.h
+++ b/src/openrct2/scenario/Scenario.h
@@ -363,6 +363,9 @@ enum
 #define AUTOSAVE_PAUSE 0
 #define DEFAULT_NUM_AUTOSAVES_TO_KEEP 10
 
+// Defines how many tiles from the edge duck target coordinates should be generated
+#define DUCK_TARGET_TILES_AWAY_FROM_EDGE 3
+
 extern const rct_string_id ScenarioCategoryStringIds[SCENARIO_CATEGORY_COUNT];
 
 extern uint32_t gScenarioTicks;
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -30,7 +30,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "23"
+#define NETWORK_STREAM_VERSION "24"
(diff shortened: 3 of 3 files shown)


deurklink:
Here's a test map you can use:
[ducktest.zip](https://github.com/OpenRCT2/OpenRCT2/files/2784778/ducktest.zip)

image

janisozaur:
1. I'd rather see the numbers named, possibly tied with relevant values.
2. Bump network version
3. It doesn't fix the crash that's already present in https://github.com/OpenRCT2/OpenRCT2/pull/8614#issuecomment-456534186

deurklink:
By the way, feel free to improve it, I won't be back until late tomorrow

src/openrct2/scenario/Scenario.cpp
```diff
@@ -427,10 +427,12 @@ static int32_t scenario_create_ducks()
...
     // Originally, this function generated coordinates from 64 to 191.
-    // It now generates coordinates from 0 to 255, so smaller maps may also spawn ducks.
+    // The duck create function is called with 2 tiles of randomness to the target coordinate.
+    // The duck create function itself adds approximately 1 tile of randomness.
+    // To spawn ducks safely, the generated coordinates should thus be from 4 to 251.
     r = scenario_rand();
-    x = ((r >> 16) & 0xFFFF) % MAXIMUM_MAP_SIZE_TECHNICAL;
-    y = (r & 0xFFFF) % MAXIMUM_MAP_SIZE_TECHNICAL;
+    x = 4 + (((r >> 16) & 0xFFFF) % (MAXIMUM_MAP_SIZE_PRACTICAL - 6));
+    y = 4 + ((r & 0xFFFF) % (MAXIMUM_MAP_SIZE_PRACTICAL - 6));
```
ZehMatt: You can also use scenario_rand_max, just call it twice instead of using a single value.

deurklink: Since the game runs this function up to 100 times to try and create coordinates to spawn ducks at, it's probably less computationally intensive to only run the scenario_rand function once.

Gymnasiast: Why do we bitmask the value twice? Seems unnecessary to me.

ZehMatt: I highly doubt scenario_rand_max will cause a bottleneck and its easier to read.

deurklink:
> 1. I'd rather see the numbers named, possibly tied with relevant values.
> 2. Bump network version
> 3. It doesn't fix the crash that's already present in [#8614 (comment)](https://github.com/OpenRCT2/OpenRCT2/pull/8614#issuecomment-456534186)

I know this doesn't fix the crash that's already present in that file, but it will prevent it from happening in the future.

src/openrct2/scenario/Scenario.h
```diff
@@ -363,6 +363,9 @@ enum
 #define AUTOSAVE_PAUSE 0
 #define DEFAULT_NUM_AUTOSAVES_TO_KEEP 10
 
+// Defines how many tiles from the edge duck target coordinates should be generated
+#define DUCK_TARGET_TILES_AWAY_FROM_EDGE 3
```
ZehMatt: What is the reason you divided the original number by two?

src/openrct2/scenario/Scenario.cpp
```diff
@@ -427,10 +427,15 @@ static int32_t scenario_create_ducks()
...
     // Originally, this function generated coordinates from 64 to 191.
-    // It now generates coordinates from 0 to 255, so smaller maps may also spawn ducks.
+    // The duck create function is called with 2 tiles of randomness to the target coordinate.
+    // The duck create function itself adds approximately 1 tile of randomness.
+    // To spawn ducks safely, the generated coordinates should not be in the outer 3 tiles of the map.
+
     r = scenario_rand();
-    x = ((r >> 16) & 0xFFFF) % MAXIMUM_MAP_SIZE_TECHNICAL;
-    y = (r & 0xFFFF) % MAXIMUM_MAP_SIZE_TECHNICAL;
+    x = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE
```
ZehMatt: That expression got even more complex, either use scenario_rand_max or create a function that returns two ranged numbers from the initial random value. It's really hard to grasp whats happening here on a first look.

deurklink: A random 32 bit int r is generated.
For the x coordinates, it right shifts the number by 16 bits, so the leftmost 16 bits of r are used. 
For the y coordinate, the number is masked with 0xFFFF so the rightmost 16 bits of r are used.

src/openrct2/scenario/Scenario.h
ZehMatt: Currently the range you select goes like following:
```
start = 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE
end = MAXIMUM_MAP_SIZE_PRACTICAL - (DUCK_TARGET_TILES_AWAY_FROM_EDGE * 2)
```
So why do you remove twice as much from the end but not from the beginning?

deurklink: Because I add an offset of 1 + DUCK_TARGET_TILES_AWAY_FROM_EDGE at the beginning to make the generated coordinates start at the correct number. If I didn't remove DUCK_TARGET_TILES_AWAY_FROM_EDGE twice at the end, the generated coordinates would have a maximum of 254.

ZehMatt:
After a little of thinking I believe the best solution towards this problem is to introduce a new function which could look like this:
```
// Returns random coordinates between MINIMUM_MAP_SIZE_PRACTICAL and MAXIMUM_MAP_SIZE_PRACTICAL
CoordsXY map_get_random_location();

// Returns random coordinates between min and max clamped to MINIMUM_MAP_SIZE_PRACTICAL /MAXIMUM_MAP_SIZE_PRACTICAL
CoordsXY map_get_random_location(const CoordsXY min, const CoordsXY max);
```
Which ensures that all the parameters are within bounds, creates a meaningful name and other code that requires random coordinates can later use it.

deurklink:
Guys, could we please keep in mind this is sort of an emergency fix to stop the game for crashing (since janizosaur did his best to emphasize the number of crash reports he got). 

I like the suggestion from @ZehMatt and it is something I would like to take my time for to make nice, since i only have limited time to work on this every day. But currently the game is crashing for a lot of people who build an a map of the maximum size, so it could be wise to merge this is at is now so I can work on a cleaner version.

Gymnasiast:
If we merge this now, it will likely be forgotten about.

src/openrct2/scenario/Scenario.cpp
Gymnasiast: It is pointless to bitmask first and then modulo. Basically, you're modulo'ing it twice.

deurklink:
> 
> 
> If we merge this now, it will likely be forgotten about.

No, because I will keep working on this 👍

Gymnasiast:
I want it to be changed in the following way:
- First, do a modulo with MAXIMUM_MAP_SIZE_TECHNICAL.
- Then (in a separate instruction!) do a `clamp` between DUCK_TARGET_TILES_AWAY_FROM_EDGE and `MAXIMUM_MAP_SIZE_TECHNICAL - DUCK_TARGET_TILES_AWAY_FROM_EDGE - 1`.

deurklink:
> 
> 
> I want it to be changed in the following way:
> 
>     * First, do a modulo with MAXIMUM_MAP_SIZE_TECHNICAL.
> 
>     * Then (in a separate instruction!) do a `clamp` between DUCK_TARGET_TILES_AWAY_FROM_EDGE and `MAXIMUM_MAP_SIZE_TECHNICAL - DUCK_TARGET_TILES_AWAY_FROM_EDGE - 1`.

Doesn't a clamp mean that the coordinates will be more likely to be generated on the outside of the map? (since all outliers will be clamped to the same number)

Gymnasiast:
It will, but I don't think that it's a problem.

deurklink:
> 
> 
> Which this function never did, though. Especially since it takes the maximum size of the map without taking the actual map size into account.

I guess a good reason to not take the actual map size into account is so smaller maps aren't flooded with the same amount of ducks a larger map would get. Why do you want a modulo with MAP_SIZE_TECHNICAL? Because it is a power of 2?

Gymnasiast:
This is getting tedious. I'm going to revert your commit and I'll take a look myself at some point.

deurklink:
Please, do not take my questions as stubbornness or unwillingness to change code. If someone suggests code, I just want to know the reasoning behind it. Our end goal is the same, to have readable code that functions well.

Gymnasiast:
No hard feelings. Since you and I disagree on what is the best solution, I want to know more about the code and the stuff around it, so I can at least give a better answer. I'll revert the original commit so the crashes go away and then I can take the time to properly investigate.

PR #8631, opened 2019-01-24
WIP: Use Direction struct for direction handling.

ZehMatt:


src/openrct2/world/Location.hpp
```diff
@@ -158,15 +158,173 @@ struct TileCoordsXYZ
...
+        direction = static_cast<decltype(direction)>(direction * times);
+        Normalise();
+        return *this;
+    }
+
+    Direction& Normalise()
+    {
+        direction = static_cast<decltype(direction)>(direction & 3);
+        return *this;
+    }
```
IntelOrca: I don't think this should mutate, rather it should return a new one. Also why is this necessary? Surely it should be impossible to construct a Direction with an invalid number?

src/openrct2/interface/Window.cpp
```diff
@@ -952,7 +952,7 @@ static void call_event_viewport_rotate_on_all_windows()
  *      1: clockwise
  *      -1: anti-clockwise
  */
-void window_rotate_camera(rct_window* w, int32_t direction)
+void window_rotate_camera(rct_window* w, Direction direction)
```
jensj12: I don’t like the change here. The argument is a rotation, not one of the four directions. Having to call this with Direction::Invalid (or any of the other names) just looks wrong to me.

src/openrct2/world/Location.hpp
ZehMatt: Somewhat early to review the code, a lot of things are currently broken and I need to make sure that everything works first before I get to the cleanup.

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -1187,12 +1190,12 @@ static void window_footpath_set_enabled_and_pressed_widgets()
...
         // Enable / disable directional widgets
         direction = gFootpathConstructValidDirections;
-        if (direction != 255)
+        if (direction.IsValid())
         {
             disabledWidgets |= (1 << WIDX_DIRECTION_NW) | (1 << WIDX_DIRECTION_NE) | (1 << WIDX_DIRECTION_SW)
                 | (1 << WIDX_DIRECTION_SE);
 
-            direction = (direction + currentRotation) & 3;
+            direction = direction + currentRotation;
```
marijnvdwerf: missing normalise?

ZehMatt:
Gonna try this again with a better overview next time.

duncanspumpkin:
Just remember directions are not all <4. Some of the directions are also for tracks that go diagonal <8

ZehMatt:
Theres even more than that, the maze also uses a value higher than 4 for something entirely different.

PR #8632, opened 2019-01-24
Refactor scenario_rand() to use c++11 <random>

tomlankhorst:
Currently implements a simple `linear_congruential_engine` but could be changed later to another one.  Chose this engine because its very small state size (single int) so that the state fits in existing save files, scenarios etc.

### DIFF
--- a/src/openrct2/GameState.cpp
+++ b/src/openrct2/GameState.cpp
@@ -47,6 +47,7 @@ void GameState::InitAll(int32_t mapSize)
 {
     gInMapInitCode = true;
 
+    scenario_init();
     map_init(mapSize);
     _park->Initialise();
     finance_init();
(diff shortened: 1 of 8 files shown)


ZehMatt:
Thats not exactly what I meant in the chat. The idea was to implement the current algorithm into a STL compatible class, meaning implementing some of the prng engine details such as min/max/operator()  so it can be used with uniform_int_distribution/uniform_real_distribution and so on. Its better to keep the current algorithm as some tests rely on it currently.

janisozaur:
I'm with @ZehMatt on that, I imagined he meant stl-like wrapper for the shaft existing PRNG.

On semi-related note, I'm impressed by the quality of the tests we now have, kudos @ZehMatt and @richard-fine

tomlankhorst:
Thanks for your comments. I'm only trying out some stuff and I also feel like the RNG behaviour should indeed be retained.

ZehMatt:
If you are talking about https://github.com/OpenRCT2/OpenLoco/blob/master/src/openloco/utility/prng.hpp then yes except its missing a few functions to properly function with something like https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution which should be used for ranged values.

PR #8633, opened 2019-01-24
Templatise `rorN` and `rolN` with `N` in {8, 16, 32, 64}

tomlankhorst:
Reduce code duplication. 
Allows arbitrary unsigned integral type rotations. 

Useful for the upcoming templatized `ror-xor-ror` PRNG. 

E.g.
```cpp
template<typename _Type>
function magic(_Type x) {
  return rol<_Type>(x,5);
}
```


Gymnasiast:
I'm afraid I have to agree with Aaron here. Especially as the current patterns are recognised by most compilers as being `rol`/`ror` instructions.

tomlankhorst:
> In fact, it's now over three times longer than it was...
It is, but that's mainly because I defined function aliases for the existing rolN/rorN. 

> Especially as the current patterns are recognised by most compilers as being rol/ror instructions.
That's not changed very much. The two version produce two times:
```
        mov     eax, edi
        mov     ecx, edi
        ror     eax, cl
```
`GCC 7.1  -O1 -std=c++17`

AaronVanGeffen:
> It is, but that's mainly because I defined function aliases for the existing rolN/rorN.

Fair point. Your code also allows us to drop the integer and infer the amount of bits to rotate based on the type in question, of course. (That is, assuming we don't apply `rot8` to a `uint16_t`...)

IntelOrca:
I like this change.

Gymnasiast:
Well, in that case, I have no objections.

IntelOrca:
@AaronVanGeffen 

> (That is, assuming we don't apply rot8 to a uint16_t...)

Well you would not be able to if you remove the ror8 altogether as `ror(some uint16)` would give you ror16 automatically.

tomlankhorst:
What's also cool is that the template type can be deducted from the first parameter:
```
uint32_t a = 0x100;
auto b = rol(a,8); // 0x10000
```

Gymnasiast:
From what I gather, this can be merged? @IntelOrca

IntelOrca:
Depending on how big of a change that would be to do, it could be a follow up PR.

tomlankhorst:
Like, openrct2/util/Rotate.h?

Gymnasiast:
.h_pp_ , but yeah.

IntelOrca:
or if we wanted to generalise more, `Numerics.hpp`

IntelOrca:
Would you be able to move the hpp to core? util is sort of obsolete now.

src/openrct2/common.h
```diff
@@ -41,14 +43,14 @@ using utf16string = utf16*;
...
 
-#define rol8(x, shift) (((uint8_t)(x) << (shift)) | ((uint8_t)(x) >> (8 - (shift))))
-#define ror8(x, shift) (((uint8_t)(x) >> (shift)) | ((uint8_t)(x) << (8 - (shift))))
-#define rol16(x, shift) (((uint16_t)(x) << (shift)) | ((uint16_t)(x) >> (16 - (shift))))
-#define ror16(x, shift) (((uint16_t)(x) >> (shift)) | ((uint16_t)(x) << (16 - (shift))))
-#define rol32(x, shift) (((uint32_t)(x) << (shift)) | ((uint32_t)(x) >> (32 - (shift))))
-#define ror32(x, shift) (((uint32_t)(x) >> (shift)) | ((uint32_t)(x) << (32 - (shift))))
-#define rol64(x, shift) (((uint64_t)(x) << (shift)) | ((uint64_t)(x) >> (64 - (shift))))
-#define ror64(x, shift) (((uint64_t)(x) >> (shift)) | ((uint64_t)(x) << (64 - (shift))))
+const constexpr auto rol8 = rol<uint8_t>;
```
Gymnasiast: I think (but I'd like him to confirm it) that @IntelOrca also meant that these consts should be moved.

janisozaur: :+1:

IntelOrca: It doesn't really matter since they are going to be deleted eventually.

src/openrct2/core/Numerics.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
```
janisozaur: Missing copyright header. When copying, remember it's 2019

tomlankhorst: Sure. Will add it to my IDE's template while I'm at it. 
Are we going to update the year numbering every now and then? 
I feel like that's horrible for version control and CI.

PR #8640, opened 2019-01-26
[Feedback] Inlining of frequently called functions

tomlankhorst:
Frequently called simple functions like `TileElement->GetType` cannot be inlined because their definition is not available at compile time. I've identified a number of functions in the paint subroutines:

- `get_current_rotation`
- `TerrainSurfaceObject::GetImageId`
- `tile_element_paint_setup`
- `bitscanforward`
- `PathElement::IsSloped`
- `PathElement::GetPathEntry`
- `PathElement::GetRailingEntry`
- `rotate_map_coordinates`
- `TileElementBase::GetType`
- `SurfaceElement::GetSurfaceStyle`
- `openrct2_assert_fwd`

I've checked `llvm`'s optimization output to validate that these functions will indeed be inlined. 
I've ran the `gfxbench` and `benchspritesort` between `devel` and this branch, compiled with `-O2`. 

|            | gfxbench (rep.40) | spritesort (rep.8) |
|------------|-------------------|--------------------|
| devel      | 323.6 k/s         | 7.49 s             |
| inlined    | 346.8 k/s         | 7.37 s             |
| difference | +7.1%             | +1.6% (1/s)        |

There seems to be quite some performance to gain. What do you think about this? Should we extend this to all kinds of getters and setters throughout the code, only do this for simple functions (getters, arithmetic, etc.) that are often called or not at all?

### DIFF
--- a/src/openrct2/core/Guard.cpp
+++ b/src/openrct2/core/Guard.cpp
@@ -24,14 +24,6 @@
 #include <cstdio>
 #include <cstdlib>
 
-void openrct2_assert_fwd(bool expression, const char* message, ...)
-{
-    va_list va;
-    va_start(va, message);
-    Guard::Assert_VA(expression, message, va);
-    va_end(va);
-}
-
 namespace Guard
 {
     constexpr const utf8* ASSERTION_MESSAGE = "An assertion failed, please report this to the OpenRCT2 developers.";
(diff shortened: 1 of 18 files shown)


tomlankhorst:
Now we get a bunch of 'potentional nullptr deref.' warnings that were already there but just not exposed.

IntelOrca:
Is it possible to get LTO working instead? I can't remember why it isn't enabled.

tomlankhorst:
... Got it to work with `clang`, will try GCC later.

tomlankhorst:
Closing in favor of #8641

PR #8641, opened 2019-01-26
Detect availability of LTO, apply when available

tomlankhorst:
Follow up of #8640. Pushing it through CI.


tomlankhorst:
Hm... `c++` complaining about `-Wsuggest-final-methods` in `GameAction.h`, while: 

```
#ifdef __WARN_SUGGEST_FINAL_METHODS__
#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wsuggest-final-methods"
#    pragma GCC diagnostic ignored "-Wsuggest-final-types"
#endif
```

tomlankhorst:
So... seems like the linking moves the code away from the `#pragma`'s. 
I ended up moving the diagnostic warning compiler options around: I'm using `target_compile_options` now to supply them to just the compiler, not the linker.

tomlankhorst:
So, running some simple back of the envelope calculations on a Ubuntu 64 machine with GCC 7.2.

|            | spritesort (rep.4) | benchgfx (rep.120) |
|------------|--------------------|--------------------|
| LTO        | 2.054M/s           | 19.34s             |
| devel      | 1.928M/s           | 20.45s             |
| difference | +6.5%              | +5.7% (1/s)        |

That's quite a neat gain IMHO. The binary ends up ~6% smaller at 8.5MB versus 9.0MB. 

[openrct-lto.zip](https://github.com/OpenRCT2/OpenRCT2/files/2799364/openrct-lto.zip)


Sure, this needs to be a `Rel*` step only.

tomlankhorst:
Implemented the conditional enabling of IPO based on build type.

janisozaur:
Let's see if Travis likes the changes and then it can be merged

janisozaur:
Travis is green. It seems LTO is not enabled properly for mingw job, but we're not using optimised builds on Travis, so it doesn't concern us too much and it can be fixed or worked around in a future PR.

janisozaur:
@tomlankhorst sadly, I've only just learned this invalidated all our extended CXX flags detection stuff. The flags were detected, but never applied anymore.

PR #8647, opened 2019-01-27
Fix marketing entrance/ride price threshold

### LINKED ISSUE #8642: Advertisement campaigns -  weird entrance fee check
**OS:** Windows 10
**Version:** 0.2.1
**Commit/Build:** 3b176cd

In the function that determines the spawn rate from advertisement campaigns, the function checks if the park entrance fee is below 4 for free entrance campaigns, or below 6 for half price entrance campaigns. For a money16 variable, this would translate to € 0.4 and €0.6. If they are lower, then the guest spawn rates for these advertisement campaigns are severly limited.

park_get_entrance_fee() returns gParkEntranceFee, which is also a Money16 variable, so a fixed point number with 1 decimal. 1.0 € would be stored as 10. Since the park entrance fee always gets set in increments of 1€, it doesn't make sense for the function to check if the entrance fee is below € 0.4 or € 0.6. I assume it was meant to check if the park entrance fee was below €4 or €6 (which are both still very low entrance fee prices).

I'm curious as to whether this was an oversight in the original game or if this is something introduced by OpenRCT2..

```
int32_t marketing_get_campaign_guest_generation_probability(int32_t campaign)
{
    int32_t probability = AdvertisingCampaignGuestGenerationProbabilities[campaign];
    Ride* ride;

    // Lower probability of guest generation if price was already low
    switch (campaign)
    {
        case ADVERTISING_CAMPAIGN_PARK_ENTRY_FREE:
            if (park_get_entrance_fee() < 4)
                probability /= 8;
            break;
        case ADVERTISING_CAMPAIGN_PARK_ENTRY_HALF_PRICE:
            if (park_get_entrance_fee() < 6)
                probability /= 8;
            break;
        case ADVERTISING_CAMPAIGN_RIDE_FREE:
            ride = get_ride(gMarketingCampaignRideIndex[campaign]);
            if (ride->price < 3)
                probability /= 8;
            break;
    }

    return probability;
}
```

tomlankhorst:
Fix for price/fee threshold in marketing guest generation probability computation. Details: #8642

### DIFF
--- a/src/openrct2/management/Marketing.cpp
+++ b/src/openrct2/management/Marketing.cpp
@@ -46,18 +46,18 @@ int32_t marketing_get_campaign_guest_generation_probability(int32_t campaign)
     switch (campaign)
     {
         case ADVERTISING_CAMPAIGN_PARK_ENTRY_FREE:
-            if (park_get_entrance_fee() < MONEY(4, 0))
+            if (park_get_entrance_fee() < MONEY(4, 00))
                 probability /= 8;
-            break;
+        break;
         case ADVERTISING_CAMPAIGN_PARK_ENTRY_HALF_PRICE:
-            if (park_get_entrance_fee() < MONEY(6, 0))
+            if (park_get_entrance_fee() < MONEY(6, 00))
                 probability /= 8;
-            break;
+        break;
         case ADVERTISING_CAMPAIGN_RIDE_FREE:
             ride = get_ride(gMarketingCampaignRideIndex[campaign]);
-            if (ride->price < MONEY(3, 0))
-                probability /= 8;
-            break;
+        if (ride->price < MONEY(0, 30))
+            probability /= 8;
+        break;
     }
 
     return probability;
--- a/src/openrct2/management/Marketing.cpp
+++ b/src/openrct2/management/Marketing.cpp
@@ -48,16 +48,16 @@ int32_t marketing_get_campaign_guest_generation_probability(int32_t campaign)
         case ADVERTISING_CAMPAIGN_PARK_ENTRY_FREE:
             if (park_get_entrance_fee() < MONEY(4, 00))
                 probability /= 8;
-        break;
+            break;
         case ADVERTISING_CAMPAIGN_PARK_ENTRY_HALF_PRICE:
             if (park_get_entrance_fee() < MONEY(6, 00))
                 probability /= 8;
-        break;
+            break;
         case ADVERTISING_CAMPAIGN_RIDE_FREE:
             ride = get_ride(gMarketingCampaignRideIndex[campaign]);
-        if (ride->price < MONEY(0, 30))
-            probability /= 8;
-        break;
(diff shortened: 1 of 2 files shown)


src/openrct2/management/Marketing.cpp
```diff
@@ -46,16 +46,16 @@ int32_t marketing_get_campaign_guest_generation_probability(int32_t campaign)
     switch (campaign)
     {
         case ADVERTISING_CAMPAIGN_PARK_ENTRY_FREE:
-            if (park_get_entrance_fee() < 4)
+            if (park_get_entrance_fee() < MONEY(4,0))
```
Gymnasiast: `MONEY(4,00)`. We're civilised people.

src/openrct2/management/Marketing.cpp
```diff
@@ -46,16 +46,16 @@ int32_t marketing_get_campaign_guest_generation_probability(int32_t campaign)
...
             break;
         case ADVERTISING_CAMPAIGN_PARK_ENTRY_HALF_PRICE:
-            if (park_get_entrance_fee() < 6)
+            if (park_get_entrance_fee() < MONEY(6,0))
                 probability /= 8;
             break;
         case ADVERTISING_CAMPAIGN_RIDE_FREE:
             ride = get_ride(gMarketingCampaignRideIndex[campaign]);
-            if (ride->price < 3)
+            if (ride->price < MONEY(3,0))
```
Gymnasiast: I'm not 100% sure if this occurrence should be modified as well. After all, rides can easily cost £0.30 if they're older, for example.

src/openrct2/management/Marketing.cpp
```diff
@@ -46,16 +46,16 @@ int32_t marketing_get_campaign_guest_generation_probability(int32_t campaign)
     switch (campaign)
     {
         case ADVERTISING_CAMPAIGN_PARK_ENTRY_FREE:
-            if (park_get_entrance_fee() < 4)
+            if (park_get_entrance_fee() < MONEY(4, 0))
```
Gymnasiast: Er, that's not what I said.

AaronVanGeffen: `clang-format` enforces the space though, doesn't it?

tomlankhorst: This was just the `clang-format` patch. Hadn't read your comment at the time.

Gymnasiast:
Could you rebase and bump the network version? I'll merge it then.

tomlankhorst:
Rebased onto <develop>, bumped network version.

src/openrct2/management/Marketing.cpp
deurklink: Is there a way to check how this was set in vanilla RCT2?

duncanspumpkin:
Just for note this is fixing an original bug not an OpenRCT2 bug. See #8642 for more info.

Issue #8634, opened 2019-01-24
[Feature Request] Chat button on the UI


TheEssemCraft:
Currently, to chat in multplayer games, the only option is to press the C button on the keyboard. This can be an issue for users that use touchscreens (like Android users) and can't use a keyboard on anything but textboxes. Adding a chat button on the UI would be useful for accessibility, as it can not only give these users a way to chat, but can also help with general accessibility.

TheEssemCraft:
EzioisAwesome56 managed to get the mute button to work as a chat button with his fork: https://github.com/EzioisAwesome56/OpenRCT2/commit/ecae8af61b47efa8cbdc6d06cdaac136788b76d0

Xkeeper0:
Since #8648 got merged, should this be resolved?

PR #8648, opened 2019-01-27
Chat toggle button

tomlankhorst:
So, I added this toggleable chat button for #8634. 
I feel like I'm almost there, except for adding a new 'Chat' sprite. 

 It has a blue matte and reduced color palette because it will pop up on the blue tabs to the left. 

I tried to follow the Wiki page about adding sprites but this was not very clear. Could someone please show me the way ;). Resources are added. [Original SVG](https://github.com/OpenRCT2/OpenRCT2/files/2800228/chat.svg.zip). 

chat 

> Doh, I was aiming at 'Chat' but I highlighed 'Cheats'.


tomlankhorst:
I'm setting up a clang-formatter githook...

IntelOrca:
I can't see anything unusual about the code, you have done it exactly like the other toolbar buttons so it should work. I can only assume that the import code is not processing your .png file correctly and losing the colour information. The offset can be easily fixed though either in the code or by adding it to the G2 json.

Gymnasiast:
Put the icon before the glyphs. There are three sets of them, but only one set of constants. It's easier to keep them at the end of G2.

tomlankhorst:
@Gymnasiast Have done that as well... 
When I `export` from the sprite file I get an empty PNG too. 

What should the format of the input PNG actually be? PNG-8/24?

Gymnasiast:
I'll look into it tonight, if you send me a reminder through Gitter at around 19:00.

IntelOrca:
@tobiaskohlbau Any PNG format should work as we use libpng to decode. It then matches the colours exactly or finds the closest possible ones to the game palette, although you can set a flag to keep the index the same if it is a 8bpp PNG which we do for font glyphs.

ZehMatt:
This icon sticks out a lot, might be the brightness.

tomlankhorst:
@ZehMatt Better?
image

Gymnasiast:
Now it's missing some pixels. It also has a greenish hue that it doesn't have on my screenshot.

tomlankhorst:
GIMP to the rescue... image.

Gymnasiast:
Nearly there - the top right and bottom right of the rightmost speech bubble appear to be missing some pixels. Should be easy to fix.

tomlankhorst:
@Gymnasiast , it seems like it does, I must admin, but this is the output of a dithered converted icon:

image

I'll try to fix up the dithering.

tomlankhorst:
image
@Gymnasiast

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -389,6 +393,16 @@ static void window_top_toolbar_mouseup(rct_window* w, rct_widgetindex widgetInde
...
             audio_toggle_all_sounds();
             break;
+        case WIDX_CHAT:
+            if (chat_available())
+            {
+                chat_toggle();
+            }
+            else
+            {
+                context_show_error(STR_CHAT_UNAVAILABLE, STR_NONE);
```
Gymnasiast: Will this ever happen? The button is not present anyway if there is no network. I would just remove the else clause and keep the `if`.

tomlankhorst: Whenever `network_get_status` is not 'connected' or `network_get_authstatus` is not 'auth status ok'. I suppose this could happen during a multiplayer game (e.g., connection lost).

tomlankhorst:
@gymnasiast didn't know about the tag.. thanks. Will do tomorrow.

distribution/changelog.txt
```diff
@@ -17,6 +17,7 @@
 - Feature: [#8377] Add option to adjust amount of autosaves to keep.
 - Feature: [#8458] Add sprite sorting benchmark.
 - Feature: [#8583] Add boosters to water coaster.
+- Feature: [#8648] Add chat button to top toolbar.
```
AaronVanGeffen:
```suggestion
- Feature: [#8648] Add optional chat button to top toolbar in multiplayer games.
```

PR #8650, opened 2019-01-29
[WIP] Use CMake for MacOS bundle, drop Xcode

tomlankhorst:
I'll use the CI for testing now.


Gymnasiast:
I'm reluctant about dropping Xcode support. But I'm willing to be convinced otherwise if it means we can improve macOS this way.

Gymnasiast:
Well, as long as we can still generate an `.app` and the documentation (especially for installing the dependencies) is up to scratch, let's get rid of it then. I have to admit that updating the Xcode project does slow down development.

IntelOrca:
> I'm reluctant about dropping Xcode support

Presumably xcode can still generate the xcode project still though?

If the generate .app has nothing missing from the original it seems like a good idea. @LRFLEW did most of the xcode project setup, he should take a look at this to make sure nothing has been missed out.

CMakeLists.txt
```diff
@@ -283,3 +287,78 @@ install(FILES "distribution/linux/openrct2-scenario.desktop" DESTINATION "${CMAK
...
+    set(BUNDLE_RESOURCE "${PROJECT}.app/Contents/Resources")
+    set(SOURCE_DATA_DIR "${CMAKE_CURRENT_SOURCE_DIR}/data")
+    set(SOURCE_ICON_DIR "${CMAKE_CURRENT_SOURCE_DIR}/resources/logo")
+    set(ICON_TARGET "${CMAKE_CURRENT_SOURCE_DIR}/distribution/macos/AppIcon.iconset")
+    set(ICON_OUTPUT "${CMAKE_CURRENT_SOURCE_DIR}/distribution/macos/AppIcon.icns")
+
+    add_custom_command(OUTPUT ${ICON_OUTPUT}
+        COMMAND sips -z 16 16     icon_x1024.png --out ${ICON_TARGET}/icon_16x16.png
+        COMMAND sips -z 32 32     icon_x1024.png --out ${ICON_TARGET}/icon_16x16@2x.png
+        COMMAND sips -z 32 32     icon_x1024.png --out ${ICON_TARGET}/icon_32x32.png
```
LinusU: Instead of creating and bundling the same image file twice, you can just name it `icon_32x32.png` and refer to that file both from `16x16@2x` and `32x32@1x` in the `AppIcon.iconset/Contents.json` file:

```json
    ...
      "scale" : "1x"
    },
    {
      "size" : "16x16",
      "idiom" : "mac",
      "filename" : "icon_32x32.png",
      "scale" : "2x"
    },
    {
      "size" : "32x32",
      "idiom" : "mac",
      "filename" : "icon_32x32.png",
      "scale" : "1x"
    },
    {
      "size" : "32x32",
    ...
```

pizza2004:
I personally always use the Xcode project to build and I don't think I even know how to set up these other methods, so I vote if this goes through that the Xcode project not be removed entirely, just made unnecessary to the production build and then it can be updated when people have the time instead of holding up development.

tomlankhorst:
@pizza2004 
I'll try to get the XCode generator working too.
Meaning that the following command will result in an `.xcodeproj`:

```bash
cmake <path/to/source/dir> -G Xcode
```

Gymnasiast:
I also generally prefer to actually provide support for stuff we put into the game, rather than allowing it to fall into disrepair. It's also clearer to the users.

pizza2004:
Oh I know how Xcode works.  For a good 2 years my main job on Commander Genius was just keeping the Xcode project up to date and building the Mac release.  I've just never had any good results doing much of anything with the command line and I hate how much even git requires me to use the command line. :D

LRFLEW:
Figured I should chime in, since I did create the original Xcode project. I'm not *against* the idea of using CMake instead of .xcproj files. Though if we do this, I think we should go all the way and drop the Visual Studio project files as well. That way we have a single build system.

I'm not familiar enough with CMake to know for sure if this is doing the right thing by just looking at it. I can test this when I get a chance to sit down with it. The main worry I can see with this is the dependency handling. For the non-bundle version (which I see you're handling as the "non-gui" build? It might be worth allowing for a "gui, non-bundle" build as well), it should use system-installed dependencies, such as from HomeBrew. However, the bundled version needs to use specifically built versions of the libraries that have the dylib id setup to correctly use the rpath. (I'm currently working on getting that switched to use Vcpkg, but it's waiting on some things like Microsoft/vcpkg#4480). I need to look into it and make sure that is all setup the way necessary for the program to work when run on a non-dev setup.

AaronVanGeffen:
> However, the bundled version needs to use specifically built versions of the libraries that have the dylib id setup to correctly to use the rpath.

Unless it statically links everything into a single binary, rather than shipping dylibs. This is what we ended up doing for OpenLoco, anyway.

LRFLEW:
@AaronVanGeffen Static libraries are an option, but I'd argue they aren't the *right* option. macOS Application Bundles are **designed** for handling dynamic libraries in a simple, user-friendly way. I would be against switching to all static libraries (though I also think the Windows builds should be using DLL's).

janisozaur:
I second @LRFLEW's comments

ZehMatt:
Visual Studio 2017 also has bultin CMake support so we could probably ditch the project files.

ccoors:
What about using something like https://github.com/kronenthaler/mod-pbxproj to update the .pbxproj? Maybe it can even be automated?

AaronVanGeffen:
> What about using something like https://github.com/kronenthaler/mod-pbxproj to update the .pbxproj?

We tried this in the past, without much success, unfortunately.

Traace:
Including cmake as a vaild build option might also improve support for older osx installs that only run xcode 3  | They could install current cmake from macports,homebrew or whatever.

AaronVanGeffen:
@Traace Interesting thought, but macOS installs that only support Xcode 3 ship with GCC 4.2. That isn't going to be able to compile our code base, due to it lacking C++14 support.

Traace:
@AaronVanGeffen No problem at all, this powerpc osx leopard installation is running gcc6 and gcc7 just fine :)

janisozaur:
GCC7 (or more likely: libstdc++7) is required, this is the lowest known version to ship with C++17's `string_view` header. You may get away with other compilers/C++ libraries, the lowest versions we have in our travis array is GCC7, Clang 5 and whatever Apple uses.

Gymnasiast:
Since this is over a year old and not finished, I will close this.
Better CMake support for MacOS is welcome, though, and I'll gladly reopen the PR if you want to work on it again.

PR #8652, opened 2019-01-29
Network window

ZehMatt:
This adds a new window for network debugging/stats. Currently it only implements a net graph tab, later on I plan to add desync debugging tools in there.

2019-02-04_15-48-51f27a9ac21fa3a333e

Also perhaps someone could help me out with the resizing, I haven't quite figured out yet how to get the small resize button on the corner working.

Things that I still need to do:
- [x] Add legend for all colors on bottom
- [x] Add resize support
- [x] Perhaps scroll the graph from right to left at the beginning.


src/openrct2-ui/windows/Network.cpp
```diff
@@ -0,0 +1,409 @@
...
+        dpi = &clippedDPI;
+
+        int32_t x = borderGap;
+        int32_t y = 50;
+
+        char textBuffer[128];
+
+        {
+            sprintf(
+                textBuffer, "%u %*s %.2f B/s", _networkCurrentStats.deltaBytesReceived[NETWORK_STATISTICS_GROUP_TOTAL], 4, "",
```
AaronVanGeffen: I'm afraid "b/s" should be translatable as well... The French use octets rather than bytes. :-)

AaronVanGeffen:
This looks great!

Another potential improvement would be to automatically convert >= 1024 bytes to kilobytes, and >= 1024 kilobytes to megabytes, etc. I imagine the numbers to be quite high for servers with a higher uptime.

src/openrct2-ui/windows/Network.cpp
ZehMatt: Learned something new, and seriously?

Gymnasiast: In general, you should use OpenRCT2-strings with placeholders rather than sprintf.

ZehMatt: I didn't think that it would matter this much in such case but turns out I'm wrong.

src/openrct2/core/CircularBuffer.h
```diff
@@ -0,0 +1,112 @@
...
+    {
+        return _size == 0;
+    }
+
+    size_type capacity() const
+    {
+        return _elements.size();
+    }
+
+    void push_back(const value_type& val)
```
tomlankhorst: I think this can be reduced to: 

```cpp
void push_back(const value_type& val)
{
  if(_size == capacity()) {
    _head = (_head + 1) % capacity();
  } else {
    _size++;
  }
  _elements[_tail] = val;
  _tail = (_tail + 1) % capacity();
}
```

or maybe add `full()` and replace `_size == capacity()`.

src/openrct2/core/CircularBuffer.h
```diff
@@ -0,0 +1,112 @@
...
+    {
+        return _size;
+    }
+
+    bool empty() const
+    {
+        return _size == 0;
+    }
+
+    size_type capacity() const
```
tomlankhorst: this can be `constexpr`

src/openrct2/core/CircularBuffer.h
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    const_reference back() const
+    {
+        return _elements[_tail];
+    }
+
+    reference operator[](size_type idx)
+    {
+        idx = (_head + idx) % _elements.size();
```
tomlankhorst: could reuse the class' (constexpr) `capacity()` here, and next

src/openrct2-ui/windows/Network.cpp
AaronVanGeffen: We should probably have separate strings for kB, MB etc… Then standardise their use through the code base, e.g. in string 5759 (`Downloading map ... ({INT32} / {INT32}) KiB`).

src/openrct2-ui/windows/Network.cpp
```diff
@@ -0,0 +1,510 @@
...
+        _bytesOutSec = (double)_bytesOut / statsTimeElapsed;
+
+        _networkAccumulatedStats = {};
+    }
+
+    _networkStats = curStats;
+    _networkHistory.push_back(_networkLastDeltaStats);
+}
+
+static void format_readable_size(char* buf, size_t bufSize, uint64_t bytes)
```
AaronVanGeffen: I think this function should go into Localisation.cpp, so other windows can make use of it easily as well.

ZehMatt: Alrite I'll see how it fits in.

ZehMatt: Fits in very well :)

ZehMatt:
I've also added a test for the CircularBuffer I didn't feel too good leaving this untested and already caught a bug, so I guess this was the right choice. I would say I'm done at this point.

test/tests/CircularBuffer.cpp
```diff
@@ -0,0 +1,67 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2018 OpenRCT2 developers
```
janisozaur: 2019

data/language/en-GB.txt
```diff
@@ -3727,11 +3727,39 @@ STR_6276    :{RED}{STRINGID} has guests getting stuck, possibly due to invalid r
 STR_6277    :{WINDOW_COLOUR_2}Station index: {BLACK}{COMMA16}
 STR_6278    :Autosave amount
 STR_6279    :{SMALLFONT}{BLACK}Number of autosaves that should be kept
+<<<<<<< HEAD
```
janisozaur: Remove

data/language/en-GB.txt
```diff
@@ -3727,11 +3727,39 @@ STR_6276    :{RED}{STRINGID} has guests getting stuck, possibly due to invalid r
...
+STR_6285    :Network Information
+STR_6286    :Bytes in
+STR_6287    :Bytes out
+STR_6288    :Total in
+STR_6289    :Total out
+STR_6290    :Base protocol
+STR_6291    :Commands
+STR_6292    :Map
+	
+=======
```
janisozaur: Remove

src/openrct2-ui/windows/Network.cpp
```diff
@@ -0,0 +1,480 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2018 OpenRCT2 developers
```
janisozaur: 2019

data/language/en-GB.txt
```diff
@@ -3731,6 +3731,21 @@ STR_6280    :{SMALLFONT}{BLACK}Chat
...
+STR_6285    :Network Information
+STR_6286    :Receive
+STR_6287    :Send
+STR_6288    :Total received
+STR_6289    :Total sent
+STR_6290    :Base protocol
+STR_6291    :Commands
+STR_6292    :Map
+STR_6293    :B
+STR_6294    :KB
```
Gymnasiast: Elsewhere in the codebase, we use `KiB`, `MiB`, …
(which means "binary" kilobyte, or 1024 bytes, unlike the ambiguous kB, MB, …)

Gymnasiast:
I also think it should default to light colours, like the rest of the windows. Dark window colours have their place, but those should be in a theme, so they can be applied consistently.

AaronVanGeffen:
@Gymnasiast In the interest of moving things along, I think we should defer that to another PR. Especially considering the consequences of changing background colours: the graph will need other colours as well. When this window is made themeable, it would ideally get at least two sets of contrast-rich graph colours available, to be used depending on the background colour.

Gymnasiast:
I'm okay with that. I think the 'About OpenRCT2' isn't themeable either -- we should just create an issue and tie it to the 0.2.2 milestone.

AaronVanGeffen:
Then I'm fine with this being merged as-is. I think the network version needs incrementing, still, though?

Gymnasiast:
Hm, why? It doesn't look like it affects the communication between clients and servers.

ZehMatt:
Yeah sorry removed the label, I bumped it since its a feature and I felt like if they should at least run with the same feature set, but its not mandatory.

Gymnasiast:
Good, could you rebase? When the CIs are green, you can merge it.

Issue #8593, opened 2019-01-15
Sign contents on mouse hover


Richard-L:
Just a quick suggestion: how about automatically showing a sign's text in appropriate complementary colour, on hovering the mouse, where currently it already shows the sign type and "right-click to modify".

![image](https://user-images.githubusercontent.com/4934209/51198410-770df100-18f4-11e9-8609-61166ddef4b1.png)

Thanks for considering.

Liampie:
I think this is a great idea, but in the case of banner facing away from the viewer like in this example, you're not supposed to read the banner. I think ideally this feature should only work from banners that are facing you.

Xkeeper0:
Something like this?

![image](https://user-images.githubusercontent.com/1077535/52020319-b9633080-24a5-11e9-80b3-6ebc0387d26c.png)

Gymnasiast:
@Xkeeper0 Open a pull request and we can help you get it up to scratch.

Richard-L:
@Xkeeper0 I think that's awesome !

Umdlye:
Implemented by #8659

PR #8659, opened 2019-01-31
Better tooltips for banners and signs

Xkeeper0:
Here is an implementation of the request in issue #8593.

* Adds tooltips to signs that are visible on hover.
  * Scrolling signs that have colored text use the text's color.
  * No entry signs show "No entry".
  * 3D landscape signs and other scrolling signs display their text.
* Changes some of the localization strings.
  * Addition of a new string for sign tooltips that resets the tooltip formatting.
  * Banner messages that ended with ` - - ` have had that trimmed.
    * `STR_RIDE_ENTRANCE_NAME`, `STR_RIDE_ENTRANCE_CLOSED`, `STR_PARK_CLOSED`, `STR_NO_ENTRY` ...
  * Those banners now use `STR_BANNER_TEXT_FORMAT` or `STR_SCROLLING_SIGN_TEXT` to provide the final ` - - `.

One small issue is that new signs that _have not been edited yet_ will show the text "Sign" instead of the default nearby-ride name, but the tile inspector does this as well.

----

Scrolling sign with colored text
No Entry sign
3D sign

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -551,8 +551,8 @@ STR_1167    :Can't raise water level here...
 STR_1168    :Options
 STR_1169    :(None)
 STR_1170    :{STRING}
-STR_1171    :{RED}Closed - - 
-STR_1172    :{YELLOW}{STRINGID} - - 
+STR_1171    :{RED}Closed
+STR_1172    :{YELLOW}{STRINGID}
 STR_1173    :{SMALLFONT}{BLACK}Build footpaths and queue lines
 STR_1174    :Banner sign in the way
 STR_1175    :Can't build this on sloped footpath
@@ -1108,7 +1108,7 @@ STR_1726    :Land not for sale!
 STR_1727    :Construction rights not for sale!
 STR_1728    :Can't buy construction rights here...
 STR_1729    :Land not owned by park!
-STR_1730    :{RED}Closed - - 
+STR_1730    :{RED}Closed
 STR_1731    :{WHITE}{STRINGID} - - 
 STR_1732    :Build
 STR_1733    :Mode
@@ -2237,7 +2237,7 @@ STR_2977    :Staff member name
 STR_2978    :Enter new name for this member of staff:
 STR_2979    :Can't name staff member...
 STR_2980    :Too many banners in game
-STR_2981    :{RED}No entry - - 
+STR_2981    :{RED}No entry
 STR_2982    :Banner text
 STR_2983    :Enter new text for this banner:
 STR_2984    :Can't set new text for banner...
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3756,7 +3756,6 @@ STR_6305    :Multithreading
 STR_6306    :{SMALLFONT}{BLACK}Experimental option to use multiple threads to render, may cause instability.
 STR_6307    :Colour scheme: {BLACK}{STRINGID}
 STR_6308    :“{STRINGID}{OUTLINE}{TOPAZ}”{NEWLINE}{STRINGID}
-STR_6309    :{RED}No entry
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3755,6 +3755,8 @@ STR_6304    :Open scenery picker
 STR_6305    :Multithreading
 STR_6306    :{SMALLFONT}{BLACK}Experimental option to use multiple threads to render, may cause instability.
 STR_6307    :Colour scheme: {BLACK}{STRINGID}
+STR_6308    :“{STRINGID}{OUTLINE}{TOPAZ}”{NEWLINE}{STRINGID}
+STR_6309    :{RED}No entry
 
 #############
 # Scenarios #
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,6 +3,7 @@
 - Feature: [#7296] Allow assigning a keyboard shortcut for the scenery picker.
 - Feature: [#8029] Add the Hungarian Forint (HUF) to the list of available currencies.
 - Feature: [#8481] Multi-threaded rendering.
+- Feature: [#8659] Banner and sign texts are now shown in tooltips.
 - Feature: [#8919] Allow setting ride price from console.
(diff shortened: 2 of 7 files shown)


data/language/en-GB.txt
```diff
@@ -3727,6 +3727,7 @@ STR_6276    :{RED}{STRINGID} has guests getting stuck, possibly due to invalid r
 STR_6277    :{WINDOW_COLOUR_2}Station index: {BLACK}{COMMA16}
 STR_6278    :Autosave amount
 STR_6279    :{SMALLFONT}{BLACK}Number of autosaves that should be kept
+STR_6280    :"{STRINGID}{OUTLINE}{TOPAZ}"{NEWLINE}{STRINGID}
```
Gymnasiast:
```suggestion
STR_6280    :{OUTLINE}{TOPAZ}“{STRINGID}”{NEWLINE}{STRINGID}
```
Always use proper curly quotes in localisation.

Gymnasiast:
The CI fails because of the formatting. Specifically, the following changes should be made: https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/487190950

Formatting can be automatically done via the `clang-format` utility, or by `git add`ing files, then doing `git clang-format`, followed by another `git add`.

ZehMatt:
I do appreciate the banner text on hovering them but not a big fan of using the banner colors, I think it would be enough to know what the banner text is.

ZehMatt:
They are not too bad but when you hover over a banner with red text for like a short time during scroll/zoom you can potentially think that you just saw an error. When it comes to text I would a maximum have 3 colors I would ever use which is different when it comes to real world banners. How hard would it be to just skip color formatting for a specific string.

Gymnasiast:
@ZehMatt Probably harder than you think. The banner text is saved in a stringid with the formatting code baked in. It requires copying and processing the string. All of this because of what I consider to be a cosmetic thing.

ZehMatt:
I think that is not correct, the banner stores only the raw string id, in Banner.Paint uses combines that string with another to the color.
See here: https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/paint/tile_element/Paint.Banner.cpp#L101

Patrik356b:
Either what @Gymnasiast said, or duplicate strings for each sign. Neither is really worth it I'd say.

Gymnasiast:
@ZehMatt That's not what the code says. It always combines with string 1731, which is:
```
STR_1731    :{WHITE}{STRINGID} - - 
```
This combines the string-with-colour with the ` - - `. 

Since I have done more stuff with banners, I'm very sure it saves the colour formatting tag in the stringid itself.

jensj12:
As far as I know, this is the only map tooltip with 3 lines. Does it work properly with the bottom toolbar enabled (RCT1 style)?

data/language/en-GB.txt
Xkeeper0: Thanks for this! I didn't realize that there were curly quotes in this (as they appear extremely similar to standard quotes).

Xkeeper0:
Thank you for the feedback! I'm going to work on implementing the suggestions here and should have an update Soon™, hopefully.

----

> As far as I know, this is the only map tooltip with 3 lines. Does it work properly with the bottom toolbar enabled (RCT1 style)?

I had checked it before submitting it (one of my roommates uses the bottom toolbar, thankfully) and it ... _almost_ works. The text does fit (though it is a tight squeeze). But the expanded text uses tooltip formatting, which doesn't match the bottom toolbar. (It's also missing the outline...)

image

This seems to be because `openrct2-ui/windows/MapTooltip.cpp`:168 uses `STR_MAP_TOOLTIP_STRINGID`, but `openrct2-ui/windows/GameBottomToolbar.cpp`:721 handles drawing the tooltip itself using `STR_STRINGID`. I'm not sure if there's an easy way to say "reset to the default tooltip styling" in formatting without having to check against `UITHEME_FLAG_USE_FULL_BOTTOM_TOOLBAR` every time, though.

----

> They are not too bad but when you hover over a banner with red text for like a short time during scroll/zoom you can potentially think that you just saw an error. When it comes to text I would a maximum have 3 colors that I use which is different when it comes to real world banners. How hard would it be to just skip color formatting for a specific string.

For normal tooltips, you have to hover over them for a short period of time before the tooltip displays. For the bottom toolbar, though, that is true.

----

> The CI fails because of the formatting. Specifically, the following changes should be made: https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/487190950
> 
> Formatting can be automatically done via the `clang-format` utility, or by `git add`ing files, then doing `git clang-format`, followed by another `git add`.

I knew it had a good chance of failing the code format tool. At the time, I didn't have `clang-format` handy. I'm going to follow this advice and push out an update to fix the formatting :)

----

I did encounter an issue with some testing last night. Some rides, apparently most tracked ones, suffer an amusing bug in that when the ride is unnamed, the entry hut banner shows "(Ride Type) #", except instead of being the auto-incremented number, it's a rapidly-incrementing number related to the scrolling of the text. I suspect it's something I may have fouled up with `openrct2/paint/tile_element/Paint.Entrance.cpp`:169, but I'm not immediately sure how (nor why it only seems to break for certain ride types). The queue banners display the names as expected.

janisozaur:
@Xkeeper0 thanks for your contributions, but it appears you have to rebase in order to resolve conflicts with translations.

Xkeeper0:
> @Xkeeper0 thanks for your contributions, but it appears you have to rebase in order to resolve conflicts with translations.

I used Github's web UI to resolve the conflict by adjusting the numbers I used.

janisozaur:
@Gymnasiast is it ok now?

Xkeeper0:
> You accidentally stripped whitespace in the en-GB file.

Can you clarify where the whitespace was stripped? Github's "See review" doesn't seem to show any comments or changes and when I checked with diff, I didn't see any unintentional whitespace deletions:

image

In this case, the change of ` - - ` after `Closed`, `No entry` etc. is "fixed" by recycling the generic banner formats `STR_BANNER_TEXT_FORMAT` or `STR_SCROLLING_SIGN_TEXT`, which include the spaces themselves.

AaronVanGeffen:
@Xkeeper0 Thanks for pointing that out, I missed that bit!

PR #8661, opened 2019-02-01
[WIP] Refactor random engine

tomlankhorst:
Introduce RotateEngine and Rct2Engine, FixedSeedSequence and Rct2Seed.
Adhere respectively to requirements `RandomNumberEngine` and `SeedSequence`.
Can be used with C++11 adaptors and distributions in <random>.


src/openrct2/scenario/Scenario.cpp
```diff
@@ -66,8 +68,7 @@ uint16_t gSavedAge;
 uint32_t gLastAutoSaveUpdate = 0;
 
 uint32_t gScenarioTicks;
-uint32_t gScenarioSrand0;
-uint32_t gScenarioSrand1;
+static Random::Rct2Engine rnd;
```
ZehMatt: I would rather call this gScenarioRand

src/openrct2/scenario/Scenario.cpp
```diff
@@ -476,6 +477,22 @@ static int32_t scenario_create_ducks()
     return 1;
 }
 
+std::pair<uint32_t, uint32_t> scenario_rand_seed()
```
ZehMatt: Perhaps declare a struct for the prng state and return a const reference. Also I'm not a fan of the stringstream usage few lines below.

src/openrct2/scenario/Scenario.cpp
tomlankhorst: Sure, I've exposed this `gScenarioRand` in `Scenario.hpp` as well.

src/openrct2/scenario/Scenario.cpp
tomlankhorst: You'd rather use no tuples at all I suppose? I like that the pair allows access like `.second` and tying with `std::tie`.

ZehMatt: I mean the state is essentially uint32_t[2] so why not use POD when you can?

src/openrct2/core/Random.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+        value_type s0;
+        value_type s1;
+    };
+
+    /**
+     * RotateEngine adheres to the _Named Requirement_ `RandomNumberEngine`
+     * https://en.cppreference.com/w/cpp/named_req/RandomNumberEngine
+     */
+    template<typename UIntType, UIntType __x, size_t __r1, size_t __r2>
```
janisozaur: Identifiers starting with double underscore are reserved for implementation.

janisozaur: Same with underscore followed by a capital letter.

tomlankhorst:
@ZehMatt Done, and added to style guide.

```cpp
/// Template parameters should be prefixed with `_T`, e.g. `_TFoo`
template<typename _TGenerator, typename _TSizeType = size_t>
void take(_TGenerator g, _TSizeType num)
{
}
```

PR #8664, opened 2019-02-03
Update changelog according to #8647

tomlankhorst:
This is a _change_ since it involves an _original bug_.


Gymnasiast:
No, it's a fix. If you take a look at the changelog, you'll see (almost) all original bugs are noted as fixes.

Also, again, you need to put `[ci skip]` in your commit title if you're not modifying code to avoid useless CI runs.

PR #8665, opened 2019-02-03
Fix misaligned access

janisozaur:
Platforms like ARM are very sensitive to unaligned access. This change
makes sure the addresses obey the C++ standard and are properly aligned.

### DIFF
--- a/src/openrct2/ride/Ride.h
+++ b/src/openrct2/ride/Ride.h
@@ -95,6 +95,8 @@ struct rct_ride_name
 };
 assert_struct_size(rct_ride_name, 4);
 
+#pragma pack(pop)
+
 /**
  * Ride type structure.
  * size: unknown
@@ -134,8 +136,6 @@ struct rct_ride_entry
     rct_string_id capacity;
     void* obj;
 };
-
-#pragma pack(pop)
 
 struct RideStation
 {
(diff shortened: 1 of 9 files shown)


IntelOrca:
It looks like we now always prefix `std::` for our memcpys and memsets.

src/openrct2/ride/Ride.h
```diff
@@ -135,8 +137,6 @@ struct rct_ride_entry
     void* obj;
 };
 
-#pragma pack(pop)
-
```
IntelOrca: This change probably breaks testpaint.

janisozaur: This PR:
```
[  PASSED  ] 2847 tests.
[  FAILED  ] 202 tests, listed below:
```
develop:
```
[  PASSED  ] 2847 tests.
[  FAILED  ] 202 tests, listed below:
```

janisozaur:
@IntelOrca re memcpy: the CIs pass, should I still prefix it?

IntelOrca:
@janisozaur I think @ZehMatt said it guaranteed inlining or something, it was done for performance.

src/openrct2/ride/Ride.h
IntelOrca: ok cool

janisozaur:
@ZehMatt what macro? I updated the calls as requested.

PR #8666, opened 2019-02-04
Implement game action for sign set style

cgdolan:


src/openrct2/actions/SignSetStyleAction.hpp
```diff
@@ -0,0 +1,119 @@
...
+    {
+        rct_banner* banner = &gBanners[_bannerIndex];
+
+        int32_t x = banner->x << 5;
+        int32_t y = banner->y << 5;
+
+        if (_is_large)
+        {
+            TileElement* tileElement = banner_get_tile_element((BannerIndex)_bannerIndex);
+            if (tileElement == nullptr || tileElement->GetType() != TILE_ELEMENT_TYPE_LARGE_SCENERY)
```
ZehMatt: This part should be probably in Query as well.

src/openrct2/actions/SignSetStyleAction.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include "../world/Banner.h"
+#include "GameAction.h"
+
+DEFINE_GAME_ACTION(SignSetStyleAction, GAME_COMMAND_SET_SIGN_STYLE, GameActionResult)
+{
+private:
+    int32_t _bannerIndex;
+    uint8_t _mainColour;
+    uint8_t _textColour;
+    bool _is_large;
```
ZehMatt:
```suggestion
    bool _isLarge;
```

src/openrct2/actions/SignSetStyleAction.hpp
```diff
@@ -0,0 +1,119 @@
...
+                    banner->x * 32, banner->y * 32, tileElement->base_height, tileElement->GetDirection(),
+                    tileElement->AsLargeScenery()->GetSequenceIndex(), _mainColour, _textColour))
+            {
+                return MakeResult();
+            }
+        }
+        else
+        {
+            TileElement* tileElement = map_get_first_element_at(x / 32, y / 32);
+            bool wall_found = false;
```
ZehMatt:
```suggestion
            bool wallFound = false;
```

ZehMatt:
I left a few suggestions/comments otherwise its looking good. You also need to increase the network version, see here: https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/network/Network.cpp#L33

src/openrct2/actions/SignSetStyleAction.hpp
cgdolan: Done.

src/openrct2/actions/SignSetStyleAction.hpp
cgdolan: Done.

src/openrct2/actions/SignSetStyleAction.hpp
cgdolan: Done.

It seems like failing validations in Query need to return an error status to prevent Execute logic from running, so I've modified some of the MakeResults() accordingly. Does this look correct?

cgdolan:
Thanks for the quick review!

janisozaur:
This is still failing on CI due to missing includes.

src/openrct2/actions/SignSetStyleAction.hpp
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        return MakeResult();
+    }
+
+    GameActionResult::Ptr Execute() const override
+    {
+        rct_banner* banner = &gBanners[_bannerIndex];
+
+        int32_t x = banner->x << 5;
```
Gymnasiast: While we're at it, could you save the x and y as `CoordsXY`? Code should look something like
`CoordsXY coords = { banner->x * 32, banner->y * 32 };`

The advantage of Coords/TileCoords is that it actually tells whether it saves big or small x/y coordinates, as well as keeping them together. You should also be able to pass `CoordsXY` into `map_get_first_element_at()` directly.

tomlankhorst:
https://github.com/cgdolan/OpenRCT2/pull/1

Gymnasiast:
Thanks for helping us out with this PR. Migrating to GameActions should reduce desyncs, and the code is a heck of a lot cleaner.

PR #8668, opened 2019-02-05
Implement Game Action Staff Set Order

duncanspumpkin:
I've also split up costume and order game commands as it didn't make sense to have them together.


duncanspumpkin:
I thought the same but the game command was originally labelled order so i continued the naming scheme. You would normally only change the order one at a time.

ZehMatt:
Could you also create replays for the new game actions?

Gymnasiast:
> I thought the same but the game command was originally labelled order so i continued the naming scheme.

Well, our old code contains lots of naming oddities -- this is one of the milder examples, obviously.

> You would normally only change the order one at a time.

Via the interface: yes. But again, _this function_ sets them all. Implying otherwise because we don't expose the option to the interface is only confusing.

duncanspumpkin:
I've made 2 replays that use just the 2 game actions. Do we want to add them to the repo?

ZehMatt:
If you want them to be tested by the CI's yes

PR #8670, opened 2019-02-05
Auto-download missing objects

IntelOrca:
This adds a new button to the missing object window to automatically download the missing objects.

Sometimes I get a chain of 502 errors when querying for objects I am wondering whether that is me, curl or the API. Would anyone be able to test this and let me know if they hit the same issue?

This downloads objects synchronously, this can be improved but I don't want to over-complicate the initial feature PR.

image

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3747,6 +3747,7 @@ STR_6296    :GiB
 STR_6297    :TiB
 STR_6298    :{STRING}/sec
 STR_6299    :Download all
+STR_6300    :Downloading object ({COMMA16} / {COMMA16}): [{STRING}]
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3576,7 +3576,7 @@ STR_6125    :Object type
 STR_6126    :Unknown type
 STR_6127    :File: {STRING}
 STR_6128    :The file could not be loaded as some of the objects referenced in it are missing or corrupt. A list of these objects is given below.
-STR_6129    :Copy selected item to clipboard
+STR_6129    :Copy item to clipboard
 STR_6130    :Copy entire list to clipboard
 STR_6131    :Object source
 STR_6132    :Ignore research status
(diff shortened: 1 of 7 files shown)


janisozaur:
@IntelOrca can you update your x86-64 compiler configuration to enable the warning all other CI jobs raised? (Comparison of signed to unsigned values.)

AaronVanGeffen:
This is really neat feature. Thanks for working on it.

Just for the record, what does the repository on our end consist of? We should make sure not to inadvertently distribute any copyrighted (official) objects, of course.

IntelOrca:
@AaronVanGeffen It is an endpoint that returns a URL of where to download the object, currently NEDesigns and they do not allow you to download the vanilla objects.

ZehMatt:
This seems cool but how would I actually test it?

telk5093:
Nice feature, seems like online contents download of OpenTTD.
btw is there any plan to _upload_ objects like BaNaNaS of OpenTTD?

Gymnasiast:
>  btw is there any plan to _upload_ objects like BaNaNaS of OpenTTD?

Ultimately, yes.

data/language/en-GB.txt
```diff
@@ -3576,7 +3576,7 @@ STR_6125    :Object type
 STR_6126    :Unknown type
 STR_6127    :File: {STRING}
 STR_6128    :The file could not be loaded as some of the objects referenced in it are missing or corrupt. A list of these objects is given below.
-STR_6129    :Copy selected item to clipboard
+STR_6129    :Copy item to clipboard
 STR_6130    :Copy entire list to clipboard
```
AaronVanGeffen: Perhaps 'Copy entire list to clipboard' should be shortened as well. It barely fits in English, but I think both Dutch and German will be in trouble for this string, too.

Gymnasiast: Shortening it in English does not solve any problems for Dutch and English. Texts should not be translated literally, they should make the idea behind the message clear. (Also, AFAIK, the current Dutch does fit.)

AaronVanGeffen: I don't disagree, but the localisation aspect was not the main point I was trying to make. The three buttons are equally wide, but their strings are not — especially with two being overly full. I'm hoping to get some balance back.

Gymnasiast: Perhaps we should change the buttons to `Copy`, `Copy all` and `Download all`, then, with the current button texts as tooltips.

IntelOrca: Done

PR #8671, opened 2019-02-05
Link targets with platform specific libs

tomlankhorst:
The 'platform' codebase might require linking with some shared libs / frameworks of the OS. I've tried to centralize this in a function such that it's also easy applied to tests and the like. 

This gets test working on MacOS when using CMake.

PR #8673, opened 2019-02-05
CMake cleanup

tomlankhorst:
(based on #8671)

The goal of this PR is to make the CMake scripts a bit more clear and easy to understand and use. 

- Consistent use of lowercase `set_check_cxx_flags()`
- Remove duplicate call to `target_link_platform_libraries()`
- [Don't allow](https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-is-it-not-recommended-to-install-a-pre-compiled-copy-of-google-test-for-example-into-usrlocal) system gtest, automatically pull and make a fresh build (1.8.1)
- Clean-up CMakeLists.txt of tests, improve dryness.


AaronVanGeffen:
Did you intend to close this PR?

tomlankhorst:
Yeah, it was a bit messy because it targeted too many things at the same
time.

PR #8675, opened 2019-02-06
.sv6r version 2: introducing compression

tomlankhorst:
This effort reduces the size of replays by a factor 2 to 10 depending on
the park complexity. ZLIB is used as the compression backend.

- The replay file version is bumped to 2.
- An `sv6r` file still starts with a magic number and a version.
- If version = 1, the file stream is consumed by the ReplayRecordData
serializer.
- If version > 1, the file stream is decompressed and the resulting
stream is consumed by the ReplayRecordData serializer.
- Introduced `Compatible` that indicates whether a file with a version
mismatch is compatible. E.g.: v1 is compatible with this `ReplayManager

In this case, the replay files in the repo on average reduce to 16% of their original size.


tomlankhorst:
Some interesting observations: 

|                    | v1 RLE=1 | v2 RLE=1 | v2 RLE=0 |
|--------------------|----------|----------|----------|
| issue5953.sv6r     |     481K |      82K |          |
| gzip( .. )         |      82K |      58K |          |
| gzip( gzip( .. ) ) |      58K |      58K |          |

ZehMatt:
I do like the changes however now that we have two versions I feel like we should split the ReplayRecordData apart into maybe two structs. I would have a layout like this:

v1:
- ReplayRecordHeader
- ReplayRecordData
v2:
- ReplayRecordHeader
- ReplayCompressionInfo // The added data in v2, could contain uint32 uncompressedSize, uint32 compressedSize
- ReplayRecordData // Compressed data

Where at the end you make version specific structs containing those structs. Hope that makes sense to you

src/openrct2/ReplayManager.cpp
```diff
@@ -77,6 +79,14 @@ namespace OpenRCT2
         }
     };
 
+    struct ReplayRecordFile
+    {
+        uint32_t magic;
+        uint16_t version;
+        uint64_t uncompressedSize;
```
tomlankhorst: `uint32_t`?

ZehMatt: Sure.

ZehMatt: Actually nevermind, would result in a useless cast from zlib.

PR #8676, opened 2019-02-06
Move Ride Set Appearance into Game Action framework

duncanspumpkin:
I've moved it over to the game action but I don't like the style too much and plan to refactor it a little.

I'm unsure what the `network_set_player_last_action_coord(network_get_player_index(game_command_playerid), coord);` part was meant to achieve. Is it for anything more than the game action result position. From looking through the code base I can't remember why there is a player last action coord and also a game action coord.


src/openrct2/actions/RideSetAppearanceAction.hpp
```diff
@@ -0,0 +1,182 @@
...
+                return std::make_unique<GameActionResult>(GA_ERROR::INVALID_PARAMETERS, STR_NONE);
+        }
+
+        return std::make_unique<GameActionResult>();
+    }
+
+    GameActionResult::Ptr Execute() const override
+    {
+        Ride* ride = get_ride(_rideIndex);
+        if (!ride)
```
ZehMatt:
```suggestion
        if (!ride || ride->type == RIDE_TYPE_NULL)
```

ZehMatt:
Left two suggestions but thats about it, looking good 👍

src/openrct2/actions/RideSetAppearanceAction.hpp
```diff
@@ -22,19 +22,32 @@
 #include "../world/Park.h"
 #include "GameAction.h"
 
+enum class RideSetAppearanceType : uint8_t
```
ZehMatt: You could move this into the new RideTypes.h files and instead of casting it later to uint8_t you could add DataSerialiserTraits<RideSetAppearanceType>.

src/openrct2/actions/RideSetAppearanceAction.hpp
```diff
@@ -50,8 +63,8 @@ DEFINE_GAME_ACTION(RideSetAppearanceAction, GAME_COMMAND_SET_RIDE_APPEARANCE, Ga
     void Serialise(DataSerialiser & stream) override
     {
         GameAction::Serialise(stream);
-
-        stream << DS_TAG(_rideIndex) << DS_TAG(_type) << DS_TAG(_value) << DS_TAG(_index);
+        auto type = static_cast<uint8_t>(_type);
```
ZehMatt: See above.

src/openrct2/actions/RideSetAppearanceAction.hpp
duncanspumpkin: I don't think this enum should be moved to RideTypes.h it is only used for this game action and for calling this game action. I'll see if I can add it to the DataSerialiserTraits.

duncanspumpkin: Its a major pain to add to DataSerialiserTraits without messing around with the headers loads. I think I'll keep the cast.

src/openrct2/actions/RideSetAppearanceAction.hpp
```diff
@@ -0,0 +1,182 @@
...
+    void Serialise(DataSerialiser & stream) override
+    {
+        GameAction::Serialise(stream);
+        auto type = static_cast<uint8_t>(_type);
+        stream << DS_TAG(_rideIndex) << DS_TAG(type) << DS_TAG(_value) << DS_TAG(_index);
+    }
+
+    GameActionResult::Ptr Query() const override
+    {
+        if (_rideIndex >= MAX_RIDES)
```
Gymnasiast: Check for both MAX_RIDES and RIDE_ID_NULL.

duncanspumpkin: They are the same value. But i can certainly add an additional check if you think its necessary.

Gymnasiast: Yes, please. They are the same value currently, but that won't necessarily always be true. Better to be explicit.

src/openrct2/actions/RideSetAppearanceAction.hpp
ZehMatt: Thats true but thats for what we get for using headers only for game actions, you get to include everything now. But I would still appreciate having it under the data serialiser now that its own type.

duncanspumpkin: The other option would be to drop the enum class and make it a normal enum.

ZehMatt: The reason I would like to see this in the serialiser is because its also shared by logging, if you implement the log member you also get nice output in your log.

duncanspumpkin: Well if you can figure out a simple way to achieve that I'm all ears.

ZehMatt: Fair point, for the time being this is functional so lets leave it for refactoring.

PR #7889, opened 2018-08-13
Refactor openrct2-ui/input/ files

SijmenSchoon:
**Warning: Massive PR**

I spent the last two weeks on refactoring the files inside openrct2-ui/input/ and some files around that that caught my eye.

I tried to make everything easier to read by splitting the big MouseInput.cpp and Input.cpp files into smaller files, and by splitting up the massive functions inside of them, which makes them easier to test in the future.

I tried to test everything as thoroughly as I can, but of course some bugs could have slipped through, in the form of a regression or similar.

PR #8684, opened 2019-02-06
Rename event_07 to periodic_update

AaronVanGeffen:
Initially split off from #7889, but opted for `periodic_update` rather than `fixed_update`.


janisozaur:
What does `fixed_update` mean here?

Gymnasiast:
Judging from the code, it gets called every 1000 ticks. (As opposed to other update functions which get fired whenever the window _needs_ updating, rather than on a fixed schedule.)

AaronVanGeffen:
I agree `fixed_update` isn't the most intuitive either. As the event in question only fires every 1000 ticks, perhaps something like `deferred_update` or even `delayed_update` would be more appropriate?

IntelOrca:
`periodic_update`

PR #8686, opened 2019-02-06
Ensure finances window's months don't overflow

janisozaur:
Some functions (e.g. window_finances_summary_scrollpaint) will try using
the returned index, which makes it overflow.

### DIFF
--- a/src/openrct2-ui/windows/Finances.cpp
+++ b/src/openrct2-ui/windows/Finances.cpp
@@ -611,14 +611,14 @@ static void window_finances_summary_mousedown(rct_window* w, rct_widgetindex wid
     }
 }
 
-static uint16_t summary_num_months_available()
+static uint16_t summary_max_available_month()
 {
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Are you referring to the overview that is visible when scrolling back to the earliest month?

src/openrct2-ui/windows/Finances.cpp
```diff
@@ -737,7 +737,7 @@ static void window_finances_summary_scrollpaint(rct_window* w, rct_drawpixelinfo
 
     // Expenditure / Income values for each month
     int16_t currentMonthYear = gDateMonthsElapsed;
-    for (int32_t i = summary_num_months_available(); i >= 0; i--)
+    for (int32_t i = summary_max_available_month(); i >= 0; i--)
```
ZehMatt: Wouldn't it be actually better to just use minus one here?

Issue #5512, opened 2017-05-28
Suggestion: Quick reconnect button for multiplayer


hypocube:
Recently a friend and I have had an issue with desynchronizations in multiplayer, so we came up with an idea of a "Quick reconnect" button in the multiplayer menu that would simply refresh the connection with the server. Otherwise we would have to re-enter the IP every time (because we were doing a LAN thing and not a public server).

Thanks for reading, I hope I put this in the right place.

Gymnasiast:
I think this was suggested before, but discouraged because it increases load on the server. (Of course, for a LAN server this would not apply as much.)

Thing is that we'd much prefer to work on actually getting rid of the desyncs, rather than working around it.

janisozaur:
I think there is some merit to this request: rather than having a "resync" button (to which I'm very much against), we could store the last entered server information so that user doesn't have to re-enter it when he opens "multiplayer" window. Especially when the program has not been restarted.

hypocube:
Thanks for replying. I get what you mean about desyncs, so janisozaur's idea would be work perfectly. Our idea really just boils down to having something to make reconnecting to the same server more convenient. 

Thank you for making RCT2 multiplayer a reality.

Nubbie:
Mentioned in #3134 and suggested at #2695 (closed) - A idea like Janisozaur would work better then avoiding the problem

Issue #3134, opened 2016-03-12
Add a 'retry'/'reconnect' console command


JarnoVgr:
Desyncs are annoying and sometimes you have to reconnect to a server. But it would be handy to support typing 'retry' or 'reconnect' in the console to quickly reconnect.

Source engine games have support for 'retry'.

janisozaur:
I believe this was raised before. It is not really feasible for the quite large map re-download with the amount of desyncs we still have.

JarnoVgr:
Even if there weren't desyncs, there could be tons of reasons to simply want to reconnect. Especially if server plugins/scripts are ever going to be supported. Makes testing mods/scripts easier. etc.

Gymnasiast:
I haven't done much multiplayer, so could you verify this: do we currently have a button to reconnect? If so, adding a console option would be easy and not intrusive.

janisozaur:
We don't have such button yet. Client has to restart/drop from MP game and connect again.

I also agree it shouldn't be hard to add it, but I fear adding it before a lot more desync causes get solved will cause an unnecessary burden on servers hosting parks, degrading the experience for others.

I'm all for having such button, as I see value in it, but only once we have addressed more glaring issues with MP, like sending custom objects every time we sync map and desyncs.

zsilencer:
There should be an undertaking to find the root source of the desyncs and fix that.  I think I found about five of them but it is really hard and it would be good to have more people to search.  I will describe some of the things that I have found can cause them:

1. The game logic can be affected by variables or memory outside of the game save state.  Look at _guestGenerationProbability for an example that had to be added to scenario_save_network/game_load_network.

2. There could also be asynchronous data affecting the game, such as the playback position of ride music which was causing a desync, since that runs in a different thread and might be different for each client.  Another example was in climate.c, a flag was being set based on if the thunder sound was playing.  This was not only being affected by asynchronicity but also whether the player had sound on or off in their config.  The flag was determining if scenario_rand would be called.  Anytime scenario_rand is called it will affect the rest of the game state, so util_rand needed to be used instead.

There are also issues with game command ghosts affecting peep behavior on client only, like placing paths, that will cause desync.  The desyncs that happen from just running the game in the background described in #1968 should be highest priority.  You can test for them faster by uncommenting part of toolbar code to enable high speed game in multiplayer, because they sometimes will not appear for hundreds of game years.

IntelOrca:
@zsilencer
[...]

Issue #2695, opened 2016-01-11
One-click reconnect in multiplayer


devnoname120:
It would help a lot to add a reconnect button to the disconnected window.

janisozaur:
I think there is some other underlying issue preventing this, in my (limited) testing, I was not able to reconnect at all without restarting the game.

Both of these could be solved at the same time.

devnoname120:
In my testing, I could reconnect right after without problems (unless when the game crashed of course).

JohannesHoffmann:
+1
i agree with devnoname120. We built a big park and every time we ran out of money we noticed the host has still 1.000 money left.
A reconnect helps but this takes to long, quit, open server list, find the server in the list and connect.
A simple reconnect would be so nice. I gues it makes only sense in the beta. In future versions the network code should be stable enough to make a reconnect unnessesary.

PR #8687, opened 2019-02-07
Multiplayer dropdown additions

Xkeeper0:
This is mostly a quality-of-life change for multiplayer players.

* Multiplayer toolbar icon changed from "two people on ride" to the globe used on the main menu.
* Toolbar icon has a "badge" overlay for current number of players.
* Toolbar icon has an "↔️" sync indicator underneath it. If the gamestate becomes desynced, a :warning: and a yellow strikethrough will appear.
* Multiplayer dropdown has a "Reconnect" option that immediately disconnects and reconnects to the server, without having to exit to the main menu, open the multiplayer menu, fetch the server list, ...

Toolbar button, showing 2 players and a synced game state.
Toolbar button with dropdown, showing a desynced game state and the "Reconnect" option


janisozaur:
#5512, #2695

Xkeeper0:
#2695 is over three years old, and was only closed after nearly 7 months without further discussion.
#3134 is nearly three years old and is still open.
#5512 is nearly two years old and is still open.

I will fix the clang-format issue tomorrow.

Gymnasiast:
The circumstances that led us to hold back adding such a button have not changed.

AaronVanGeffen:
Personally, I don't mind having the option to reconnect in the multiplayer menu. In its current implementation, it is quite unintrusive.

While fixing desyncs should remain the goal, there may always be connection issues, unrelated to the desyncs that occur due to the current implementation. As @Xkeeper0 said, this is a well-overdue quality-of-life change for multiplayer players.

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -993,8 +996,14 @@ static void window_top_toolbar_paint(rct_window* w, rct_drawpixelinfo* dpi)
         y = w->y + window_top_toolbar_widgets[WIDX_NETWORK].top + 0;
         if (widget_is_pressed(w, WIDX_NETWORK))
             y++;
-        imgId = SPR_SHOW_GUESTS_ON_THIS_RIDE_ATTRACTION;
-        gfx_draw_sprite(dpi, imgId, x, y, 0);
+        imgId = SPR_G2_MULTIPLAYER_TOOLBAR;
+        gfx_draw_sprite(dpi, imgId, x - 2, y - 2, 0);
+        gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
+        imgId = (network_is_desynchronised() ? SPR_G2_MULTIPLAYER_DESYNC : SPR_G2_MULTIPLAYER_SYNC);
```
AaronVanGeffen: This should be guarded for network availability, e.g. through:
```
#ifndef DISABLE_NETWORK
...
#endif
```

IntelOrca:
The code for reconnecting is still useful even if we don't have the button. The reason is that if we eventually do fix all the desyncs to the point where they happen rarely we can then do auto-reconnect on desync.

IntelOrca:
It should not show the reconnect button if you are not desynced.

src/openrct2-ui/windows/TopToolbar.cpp
Xkeeper0: The rest of the toolbar code doesn't check for `#ifndef DISABLE_NETWORK`, so I'm unsure if you mean the _entirety_ of the toolbar code involving network options should be wrapped in this.

However, it _is_ indirectly protected; the toolbar button is only available if `network_get_mode()` returns `NETWORK_MODE_CLIENT` or `NETWORK_MODE_SERVER` (see line 789), and if `DISABLE_NETWORK` is set, `network_get_mode()` will _always_ return `NETWORK_MODE_NONE`. This same outcome applies to the Chat button as well.

Xkeeper0:
> The code for reconnecting is still useful even if we don't have the button. The reason is that if we eventually do fix all the desyncs to the point where they happen rarely we can then do auto-reconnect on desync.

It may also be useful for adding to the error popup you get when you become disconnected from a server due to other reasons (internet hiccup, etc.)

A potential future addition would be tracking if the current game _was_ a network game that became disconnected, and keeping the multiplayer toolbar button but perhaps with a :x: and only the option to reconnect.

src/openrct2-ui/windows/TopToolbar.cpp
AaronVanGeffen: Please have a look at the [CI output for the networkless build](https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/489833657).

Xkeeper0: That was because two of the functions I added to `openrct2/network/Network.cpp` were missing from its `DISABLE_NETWORK` section. I've resolved that in the latest commit.

Xkeeper0:
> This is a good change for consistency, I think. One minor thing: the globe overlaps the button border, particularly the light part. I'd prefer it to be trimmed a little.

I'll trim the left side (it is only a few pixels, anyway). The right side will still overlap a little, but I am pretty sure at least one other toolbar button does as well (and it's the dark side)

> I like this idea, but I think the icons can still be improved upon. More specifically, I think the contrast between the green and red hues is too low. This makes them appear a little bit off, especially on top of the globe, I think.

Admittedly I did not try to find a _great_ palette and merely settled for a _decent_ one, because testing sprites in this so far has been a _huge_ pain.

The wiki is not clear on how to make sprites and I can't honestly remember _how_ I did it at this point.

> `openrct2 sprite append <sprite file> <image> [x offset] [y offset]` Adds the specified image to the end of the sprite file. If supplied, the x and y offset will be used. Since OpenRCT2 has a limited colour palette, the image has to be converted. Currently, there are three modes of conversion, which can be specified by the option -m: (default, closest, ...)

This makes it sound like this command would work:
```
openrct2 sprite append g2.dat <image.png> 0 0 -m closest
```
but when I try that, I get a usage hint and nothing of value:
```
openrct2/build$ ./openrct2 sprite append g2.dat ../resources/g2/icons/multiplayer_toolbar.png 0 0 -m closest
usage: sprite append <spritefile> <input> [<x offset> <y offset>]
```

Maybe `sprite` needs a `convert <in> <out> -m <method>` option that takes an image and simply outputs it using the specified color method or something.

Xkeeper0:
> It should not show the reconnect button if you are not desynced.

I don't quite agree with this but the latest commit addresses this as well.

image

image

Gymnasiast:
> It should not show the reconnect button if you are not desynced.

Why? It's only confusing. I much prefer the earlier solution of greying it out. (It's a menu entry, not a button, by the way.)

AaronVanGeffen:
This PR has stalled a bit, likely due the discussion surrounding it. I'd like to pick it up again. The following things still need sorting, as far as I'm concerned:

- [ ] Rebase on top of current develop.
- [ ] The globe in the toolbar image should be smaller than it is currently — it covers part of the frame.
- [ ] The toolbar image should be a complete, framed image instead of just an overlay.
- [ ] An image for the pressed toolbar state needs to be introduced as well.

I aim to take care of this tomorrow.

AaronVanGeffen:
Sync / desync status:
Screenshot_20190502_122713 Screenshot_20190502_122531

Pressed:
Screenshot_20190502_122452

Raw unpressed / pressed:
multiplayer_toolbar multiplayer_toolbar_pressed

PR #8689, opened 2019-02-07
Move Land Set Height into GameAction Framework

duncanspumpkin:


src/openrct2/actions/LandSetHeightAction.hpp
```diff
@@ -0,0 +1,334 @@
...
+            TileElement* tileElement = map_get_first_element_at(_coords.x / 32, _coords.y / 32);
+            do
+            {
+                int32_t elementType = tileElement->GetType();
+
+                if (elementType == TILE_ELEMENT_TYPE_WALL)
+                    continue;
+                if (elementType == TILE_ELEMENT_TYPE_SMALL_SCENERY)
+                    continue;
+                if (tileElement->flags & 0x10)
```
Gymnasiast: Should be replaced with the appropriate constant.

AaronVanGeffen: It should, but this is moved code, not new code. Not blocking IMO.

src/openrct2/world/Map.cpp
```diff
@@ -2627,8 +2393,8 @@ void game_command_place_large_scenery(
 
         if (!gCheatsDisableClearanceChecks
             && !map_can_construct_with_clear_at(
-                   curTile.x, curTile.y, zLow, zHigh, &map_place_scenery_clear_func, bl, flags, &supportsCost,
-                   CREATE_CROSSING_MODE_NONE))
+                curTile.x, curTile.y, zLow, zHigh, &map_place_scenery_clear_func, bl, flags, &supportsCost,
```
Gymnasiast: Looks like you have an autoformatter on that has different settings than clang-format. It's better to turn it off and just do `git clang-format` after making changes and doing `git add`.

duncanspumpkin: I'm using clang format just it doesn't like to match the other clang format.

src/openrct2/actions/LandSetHeightAction.hpp
Gymnasiast: For me, it is. It's trivial to fix.

src/openrct2/actions/LandSetHeightAction.hpp
```diff
@@ -0,0 +1,380 @@
...
+                }
+            }
+            if (!map_can_construct_with_clear_at(
+                    _coords.x, _coords.y, _height, zCorner, &map_set_land_height_clear_func, 0xF, 0, nullptr,
+                    CREATE_CROSSING_MODE_NONE))
+            {
+                return std::make_unique<GameActionResult>(GA_ERROR::DISALLOWED, gGameCommandErrorText);
+            }
+
+            // This code possibly does nothing? TODO: REMOVE IF ASSERTS NOT HIT
```
duncanspumpkin: Unsure about this code I think it's dead.

src/openrct2/actions/LandSetHeightAction.hpp
```diff
@@ -0,0 +1,385 @@
...
+    uint16_t GetActionFlags() const override
+    {
+        return GameAction::GetActionFlags() | GA_FLAGS::EDITOR_ONLY;
+    }
+
+    void Serialise(DataSerialiser & stream) override
+    {
+        GameAction::Serialise(stream);
+
+        stream << DS_TAG(_coords.x) << DS_TAG(_coords.y) << DS_TAG(_height) << DS_TAG(_style);
```
ZehMatt: Could add CoordsXY to DataSerialiserTraits to simplify this case.

duncanspumpkin:
Lol so amusingly LandSetHeight has never actually been a game command that is sent over the network. Its always been inside another game command.

tomlankhorst:
https://github.com/duncanspumpkin/OpenRCT2/pull/16

ZehMatt:
```
GAME_COMMAND_EDIT_LAND_SMOOTH -> game_command_smooth_land -> smooth_land -> 
	GAME_COMMAND_RAISE_LAND,
	GAME_COMMAND_LOWER_LAND,
	GAME_COMMAND_SET_LAND_HEIGHT
GAME_COMMAND_RAISE_WATER -> game_command_raise_water -> raise_water -> 
	GAME_COMMAND_SET_WATER_HEIGHT
GAME_COMMAND_LOWER_WATER -> game_command_lower_water -> lower_water ->
	GAME_COMMAND_SET_WATER_HEIGHT

janisozaur:
> Lol so amusingly LandSetHeight has never actually been a game command that is sent over the network. Its always been inside another game command.

@duncanspumpkin when @ZehMatt first implemented replays, someone reported that they created a hill and it was not getting replayed. On top of that, the replay did not report out-of-sync, which uses sprite checksum. I realise it's only _sprite_ checksum, but terrain is another large piece of game state.

PR #8694, opened 2019-02-08
Start moving track place into GameAction framework

duncanspumpkin:
Track place game action


tomlankhorst:
https://github.com/duncanspumpkin/OpenRCT2/pull/18

AaronVanGeffen:
Is this deliberately not removing the old code path yet?

duncanspumpkin:
Yes i need to get the callbacks working then I can start dismantling the old code. Its a messy function this one.

duncanspumpkin:
Currently there is an issue with networked games. When placing the first piece the game is closing the construction window and destroying the ride. Likely due to checking for track piece before the piece has been placed.

pizza2004:
I see the commit about preventing ghost pieces from showing up in multiplayer and I was wondering if it would be possible now with game actions to instead synchronize the display of ghost pieces across the network?

duncanspumpkin:
No that would not be possible. Ghosts are drawn and removed multiple times a game tick it would clog the network. All of the saving us designed around removing ghosts as well. On top of all that I don't think people would want that.

PR #8697, opened 2019-02-09
Localization rework

Xkeeper0:
I changed the way language files work.

* Rather than a simple enum in `StringIds.h`, string ids are retrieved by including `RctStrings.h`, which has a goofy macro to generate the enum using preprocessor witchcraft.
* This mess simultaneously creates the enum, but also creates a `std::unordered_map<string, rct_string_id>`, using the enum constant name as the key and the enum value for the value.
* When parsing the language file, it first checks if there's a value in the map for the identifier, and uses that if so. Otherwise, it falls back to using the numeric id. If there's a performance hit, it's only during the initial loading of the language file -- the same old `rct_string_id`s are still passed around everywhere.
* The `language/en-GB.txt` file has been swapped from using `STR_(string id)` to the actual name of the string constant, if it was set. The ability to use `STR_(string id)` still exists for strings that aren't in the enumeration or haven't been converted. Notably, all other language files still work normally.
* A `printf` statement drops an `Asked for undefined string %d` when `GetString()` is called but no string matches the id given. I originally wanted to be able to show this in-game as "`(undefined string 0925)`", but due to the way that function works I don't think it's possible without assigning that as a string to any unused string id between 0 and _(whatever)_.

Mostly this was done because I got really tired of playing the game of having to keep multiple files open to know what a string was. With this, it's much easier to tell -- if you want to use `{WINDOW_COLOUR_2}{CURRENCY}`, rather than having to go check `StringIds.h` and search for what string constant has id `3246`, it's _right there_ in the language file -- `STR_CURRENCY_FORMAT_LABEL`.

It also makes additions that add new strings less of a pain to add, and maybe less prone to merge conflicts. In theory, the values for the enum constants could be dropped entirely and allowed to auto-increment again, since you can just use the const name, but the macro may need a little tweaking for this since it _expects_ a value. I think. Maybe.

----

I'm opening this PR mostly to see if there's any interest in merging it, to know if I should polish it up more. Particularly:

* I have a script that I (sadly) don't have handy that converted the old format to the new format, which I could run on other languages in the Localisation repo and PR there.
* The padding length for identifiers in `en-GB.txt` was chosen completely arbitrarily based on what looked okay without being _too_ wide. Some of the string identifiers are _really long_.
* The console liked to print out `Asked for undefined string id 954` (`STR_CONSTRUCTION_ERR_UNKNOWN`) even though it exists and is defined, it's just empty. I'm not sure if this is a weird bug I introduced or what. In theory the output could see if the hash map contains a key for that value and output the const name, but it doesn't really matter, probably.
* I did a quick skim through GitHub's diff tool and didn't _see_ any whitespace issues, but it's possible that some of the lines may have gotten trimmed by accident.
* Maybe some other things, who knows. I did this while bored and without asking anyone, and it's kind of a big change, so I'm not expecting much.

Edit: Oh yeah, I tested this only on Ubuntu with cmake so if I had to guess some other project files will need updating because of the new header file or whatever.


ZehMatt:
Frankly I have no issue with the way the data is stored I have however some minor nitpick regarding formatting. If there is room for improvement its probably the way they are formatted given that it still uses alot of C for that. As an example we could have a better format function something like format_string(buf, len, str_id, args..) instead of setting a global/local array of fields this can done in a better way.

tomlankhorst:
Nonetheless, I like that strings are identified by a representative string key, instead of a numerical id.

ZehMatt:
Sadly a few conflicts came up, can you rebase it?

Gymnasiast:
Now, we certainly want to update the language files and introduce textual identifiers instead of the current numeric ones, but I'd prefer YAML or Gettext instead of what amounts to a custom format. The PR has also been abandoned. For those reasons, I'll close this.

PR #8702, opened 2019-02-10
Fix #8696: Setting land height costs money in no-money scenarios

### LINKED ISSUE #8696: Raising land costs money in No-Money scenarios
**Version:** 0.2.1-750
**Commit/Build:** f32531c

**Steps to reproduce:**
1. Open any scenario
2. Disable money (cheats or scenario options)
3. Raise or lower land

Gymnasiast:


duncanspumpkin:
Can you add an issue to refactor this as we shouldn't be setting the price to zero in every single game command it should all be in just the one finance payment location.

Gymnasiast:
@duncanspumpkin I agree, but this was causing a lot of problems, so it had to be patched quickly.

Could _you_ open that issue, by the way? You understand game actions better than I do, so I think the issue would be more informative and more likely to be correct.

ZehMatt:
I'm a little puzzled how this fixes the issue, except the scenario editor flag and cheats flag its already checked here https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/actions/GameAction.cpp#L291

Gymnasiast:
@ZehMatt Does it actually execute that line? Also, the check is wrong anyway, since it does not check if the user is in the Scenario Editor.

ZehMatt:
Yes I know, but the bug is caused in normal no-money scenarios so it should work, thats why im slightly confused right now, currently debugging the issue.

ZehMatt:
So the problem is relatively simple, raising/lowering land does not directly use LandSetHeightAction instead its nested from the game commands game_command_raise_land/game_command_lower_land which never go thru GameActions::Execute given that they are nested. We can either fix this per action basis by simply resulting 0 or we can add checks here: https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/Game.cpp#L538

Gymnasiast:
If it only happens in situations where it's nested inside an old, clunky game command, then it should probably be the latter.

ZehMatt:
Thats not the case tho, game actions also need to consider the scenario editor, CheckActionAffordability also doesn't ignore ghosts etc. There is a lot to cleanup.

duncanspumpkin:
@Gymnasiast sorry was a little busy and didn't have time at that moment to write up an issue.

PR #8703, opened 2019-02-10
GameAction cleanup part 1

ZehMatt:
Part 1 of the cleanup, this is only for the money quirks.

This ensures that no money is ever used under given conditions:
- Scenario Editor
- No Money scenario
- Ghost piece
- Preview piece.

This applies to game actions and legacy game commands. I moved the function from GameActions.cpp to Finance.cpp so game commands can use the same logic. Also invoking game actions from game commands will no longer consume any money as well, no need to worry in further GA porting.

### DIFF
--- a/src/openrct2/actions/LandSetHeightAction.hpp
+++ b/src/openrct2/actions/LandSetHeightAction.hpp
@@ -134,11 +134,7 @@ public:
             }
         }
         auto res = std::make_unique<GameActionResult>();
-        res->Cost = 0;
-        if (!(gScreenFlags & SCREEN_FLAGS_EDITOR) && !(gParkFlags & PARK_FLAGS_NO_MONEY))
-        {
-            res->Cost = sceneryRemovalCost + GetSurfaceHeightChangeCost(surfaceElement);
-        }
+        res->Cost = sceneryRemovalCost + GetSurfaceHeightChangeCost(surfaceElement);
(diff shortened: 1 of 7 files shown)


src/openrct2/actions/LandSetHeightAction.hpp
```diff
@@ -134,11 +134,7 @@ DEFINE_GAME_ACTION(LandSetHeightAction, GAME_COMMAND_SET_LAND_HEIGHT, GameAction
             }
         }
         auto res = std::make_unique<GameActionResult>();
-        res->Cost = 0;
-        if (!(gScreenFlags & SCREEN_FLAGS_EDITOR) && !(gParkFlags & PARK_FLAGS_NO_MONEY))
-        {
-            res->Cost = sceneryRemovalCost + GetSurfaceHeightChangeCost(surfaceElement);
-        }
+        res->Cost = sceneryRemovalCost + GetSurfaceHeightChangeCost(surfaceElement);
```
duncanspumpkin: Is this really the only game action that needs modified?

ZehMatt: Actually FootpathRemoveAction.hpp has GetRefundPrice which doesn't need any of those conditions anymore.

duncanspumpkin:
Can you remove 9f8a2eb and squash into key commits.

PR #8704, opened 2019-02-11
GameAction cleanup part 2

ZehMatt:
This will eliminate the need to always call network_set_player_last_action_coord. It includes the commits from part 1 so I won't run into conflicts later, I'll rebase this once part 1 is merged.


duncanspumpkin:
L96-105 of TrackPlaceAction still need adjustment

ZehMatt:
Oh yeah, totally ignored gCommandPosition.

PR #8707, opened 2019-02-13
Fix 8706. Mountain tool clipping issues after action refactor

duncanspumpkin:
Ultimately the mountain tool needs further refactoring when ported to a game action. For now this will fix the issue.

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -1580,8 +1580,12 @@ static money32 smooth_land_tile(
     }
 
     auto landSetHeightAction = LandSetHeightAction({ x, y }, targetBaseZ, slope);
-    auto res = (flags & GAME_COMMAND_FLAG_APPLY) ? landSetHeightAction.Execute() : landSetHeightAction.Query();
-
+    auto res = landSetHeightAction.Query();
+    if (res->Error == GA_ERROR::OK && flags & GAME_COMMAND_FLAG_APPLY)
+    {
+        // This is required as we want Query to not fail the parent game command
+        res = landSetHeightAction.Execute();
+    }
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Map.cpp
```diff
@@ -2393,8 +2402,8 @@ void game_command_place_large_scenery(
 
         if (!gCheatsDisableClearanceChecks
             && !map_can_construct_with_clear_at(
-                   curTile.x, curTile.y, zLow, zHigh, &map_place_scenery_clear_func, bl, flags, &supportsCost,
-                   CREATE_CROSSING_MODE_NONE))
+                curTile.x, curTile.y, zLow, zHigh, &map_place_scenery_clear_func, bl, flags, &supportsCost,
+                CREATE_CROSSING_MODE_NONE))
```
Broxzier: Clang-format wants the indention here restored.

If your clang-format formatted it this way... I noticed a similar issue earlier. The version on Travis probably doesn't match your local machine.

duncanspumpkin: Yes it's quite annoying. I seem to be hitting a few clang format bugs.

duncanspumpkin:
One thing i found in testing is it actually meant you could change surfaces that you don't own!

src/openrct2/world/Map.cpp
IntelOrca: Might be better to make it a new nested if block.

ZehMatt:
I've been lately thinking that its actually safe to just use GameActions::Execute if we always pass down the flags. Given that when its executed in MP it will carry the network flag otherwise requires a round trip, so top level commands should invoke nested ones with the flags always set.

PR #8709, opened 2019-02-14
Fix 8698

ZehMatt:
I meant to eliminate the whole expression not just the "true" keyword.

### DIFF
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -30,7 +30,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "35"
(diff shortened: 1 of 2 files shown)


Broxzier:
The previous behaviour, though difficult to read, was correct. Right now it fails at intersections:

Before (release):
rc f multiplayer squares 23 2019-02-14 19-00-53

After (this branch):
rc f multiplayer squares 23 2019-02-14 18-59-42

Guests will only path-find over the non-ghost paths, but as you can see they stop.

ZehMatt:
Thats odd, I'll have another look, I wonder if a wrong flag perhaps slipped thru.

ZehMatt:
Okay so heres what I get from 2.1
openrct2_2019-02-14_21-20-160cbeaf36be8d88f15

And with this PR:
openrct2_2019-02-14_21-23-41e816530f995d67249

Can you validate that its now correct?

PR #8711, opened 2019-02-15
Refactor direct calls to Query and Execute on game actions.

ZehMatt:
Previously it was unsafe to directly call Execute as Query would always validate the input and conditions. Instead of calling it directly I added an optional "topLevel" parameter to GameActions::Query/Execute which is by default set to true, meaning it is the root command and should obey network/queue, when set to false it will Query/Execute without the consideration of any round trips, this should be used when used within an action.

Now every Execute will also use Query before finally going to execute the action, this might solve a few crashes reported from backtrace, I would need to go over it.

This will also probably supersede #8707


ZehMatt:
Done, and yes it looks better code wise.

PR #8712, opened 2019-02-15
Decouple network updates from game ticks.

ZehMatt:
This greatly reduces the delay on every action by allowing the network to queue up multiple commands/actions for the next tick instead of waiting for the next tick to happen before the messages actually go out.


ZehMatt:
Not so compatible with the new track actions, I have to first figure out whats happening.

ZehMatt:
I will continue this in #8785

PR #8715, opened 2019-02-16
Fix #8264: Rides and scenery placeable outside of map with ZC and Sandbox mode enabled

### LINKED ISSUE #8264: Rides and scenery placeable outside of map with ZC and Sandbox mode enabled
**OS:** Windows 7, Ubuntu 18.04
**Version:** 0.2.1+
**Commit/Build:** de12a8d

With zero clearance and sandbox mode, rides and scenery that can use the scattered placement do not produce an "Off edge of map" error like pathing does and can be constructed outside of the boundaries of the map. This crashes the game when an entity would be placed on an X or Y greater than 255 or less than 0. Rides operate just fine out of the edges of the map so long as the absolute boundaries are not reached. It does not seem like a dump file is produced.

**Steps to reproduce:**
1. Turn on sandbox mode and zero clearance
2. Begin constructing a ride or scattering scenery out of the edges of the map 
3. Witness crash past 0 or 255 X/Y

jensj12:
map_can_construct_with_clear_at first checks whether it is inside the map, and then checks gCheatsDisableClearanceChecks, making earlier checks for gCheatsDisableClearanceChecks unnecessary.

Note: I have removed a check for the ghost flag in Track.cpp as that fixed the issue for track designs, but I don't know why it was there in the first place. I haven't found any side effects yet (I can still build tracked rides and place track designs normally).


duncanspumpkin:
I like the changes so that the global is only checked once. But we should in addition be validating coordinates before passing to the construct with clear function. That way we can set an appropriate error message.

Also the network version has updated a good few times since this. I like to keep the network version increment as a separate commit as it inevitably goes stale before merging.

jensj12:
I don't see yet why we would need to check the coords before passing it to the function. The sole reason I removed the check for `gCheatsDisableClearanceChecks` is to make it check for the coords inside `map_can_construct_with_clear_at` in any case. Error messages work perfectly fine.

Still, I don't know why `(GetFlags() & GAME_COMMAND_FLAG_GHOST)` was checked before calling `map_can_construct_with_clear_at` in TrackPlaceAction.cpp.

duncanspumpkin:
The track place special situation is because we want ghosts to be shown going through obstacles. This is unlike the other calls.

The way i see game actions is all parameters should be checked for validity first. If they are invalid a ga error invalid parameter would be set. Can construct with clear is not checking parameters it's checking clearance and i would expect a ga error disallowed. 

This doesn't mean that this is currently how the code base works. It's just how i would expect it to work. Does that make sense?

Gymnasiast:
> The track place special situation is because we want ghosts to be shown going through obstacles. This is unlike the other calls.

RCT2 will only show ghosts if the whole track design can be placed. That's how most player expect it to work. Changing that is most certainly not something that "we" want.

duncanspumpkin:
Turned out I had the wrong end of the stick anyway I had another think about this code and I think the changes are correct. I'll play with it some more. This is the track place not track design place. Rules are slightly different for the two.

Gymnasiast:
> This is the track place not track design place.

I misread that. Still, as far as I know, RCT2 does not show a preview if the part cannot be placed, and I think that is the best way to do it -- it makes it immediately clear to the user whether the part can be placed or not.

duncanspumpkin:
Not always the case 
image

Gymnasiast:
Is that vanilla?

duncanspumpkin:
Yes. I think thats the reason for the original ghost flag check but I need to look more into it.

IntelOrca:
@Gymnasiast yeah, it only hides the track if the action would result in an error message showing. Note however that this is only checked when the ghost is first shown. If you happen to then run out of money afterwards you will still see the ghost until you attempt to buy.

Gymnasiast:
> yeah, it only hides the track if the action would result in an error message showing.

^ This is how I think it should work

> Note however that this is only checked when the ghost is first shown. If you happen to then run out of money afterwards you will still see the ghost until you attempt to buy.

That seems correct (and consistent) behaviour to me. What isn't correct behaviour is that you can place stuff in the way and the ghost won't update -- a vanilla bug that may or may not still be present in OpenRCT2.

duncanspumpkin:
Anyway going back to this PR this is correct. I've tested the behaviour and the ghost check does not change anything. It performs as expected and is fine to merge (after fixing the clang format)

PR #8716, opened 2019-02-16
Refactor marketing to be a single vector

IntelOrca:


src/openrct2/world/Park.cpp
```diff
@@ -967,15 +963,12 @@ void Park::GenerateGuests()
     }
 
     // Extra guests generated by advertising campaigns
-    for (int32_t campaign = 0; campaign < ADVERTISING_CAMPAIGN_COUNT; campaign++)
+    for (const auto& campaign : gMarketingCampaigns)
     {
-        if (gMarketingCampaignDaysLeft[campaign] != 0)
+        // Random chance of guest generation
+        if ((int32_t)(scenario_rand() & 0xFFFF) < marketing_get_campaign_guest_generation_probability(campaign.Type))
```
ZehMatt:
```suggestion
        if (scenario_rand_max(0xFFFF) < marketing_get_campaign_guest_generation_probability(campaign.Type))
```

ZehMatt: Few lines above can be changed too.

IntelOrca: That would cause sign mis-match wouldn't it?

ZehMatt: By type yes but marketing_get_campaign_guest_generation_probability never returns a negative value and using "and" is not a good distribution.

IntelOrca: I have changed the function but you still need the type cast.

ZehMatt: The cast is not required, it will only exhibit the compiler warning but the logic remains equivalent in this case, see this little snippet https://onlinegdb.com/B1OUdCUr4 . Ideally marketing_get_campaign_guest_generation_probability should return uint32_t.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1542,7 +1542,7 @@ static constexpr const utf8* console_variable_table[] = {
     "guest_initial_thirst",
     "guest_prefer_less_intense_rides",
     "guest_prefer_more_intense_rides",
-    "forbid_marketing_campagn",
+    "forbid_marketing_campaign",
```
Gymnasiast:
```suggestion
    "forbid_marketing_campaigns",
```

src/openrct2/world/Park.cpp
IntelOrca: I have changed it to a `uint16_t` since it representing a fraction between 0 and 0xFFFF. You could go further and have some sort of `probability<uint16_t>` type but that can be done in a separate PR.

PR #8718, opened 2019-02-18
Remove logic for required objects

IntelOrca:
This allows TARMAC to be removed when `remove_unused_objects` is called.


Gymnasiast:
Why, though?
And is there logic to prevent selecting 0 footpaths?

IntelOrca:
@Gymnasiast yes when I tried it out it said you must select at least one footpath.

This is a request from NEdesigns.

PR #8719, opened 2019-02-18
Fix error value checked from ride_get_smallest_station_length

janisozaur:
ride_get_smallest_station_length returns the int32_t::max instead of -1
on error.

See https://github.com/OpenRCT2/OpenRCT2/blob/671016c24ec9bc94f602602a2bc39142ad553ce9/src/openrct2/ride/Ride.cpp#L7241

### DIFF
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -31,7 +31,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "23"
+#define NETWORK_STREAM_VERSION "24"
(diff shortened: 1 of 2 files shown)


IntelOrca:
@tomlankhorst managed to get optional working for macOS. Could we take the parts from that and then return a `std::optional<int32_t>` instead?

AaronVanGeffen:
928f49537bf0871416c6641d412bdfb161923356 from https://github.com/OpenRCT2/OpenRCT2/pull/8611 should contain what you need for that. :)

janisozaur:
@IntelOrca I like that, in fact, I was wondering if we can already do such a thing. With @tomlankhorst away for now, I will cherry-pick relevant commit.

janisozaur:
I've only added the header for now, I have yet to adapt the code to use it.

IntelOrca:
I've been wanting to use `optional` for ages but thought we couldn't due to macOS / xcode.

PR #8722, opened 2019-02-19
Quarter tile refactor

duncanspumpkin:
Quarter tiles had a lot of duplicate code for rotating. This will hopefully reduce it.

`map_can_construct_with_clear_at` still needs some rework to use this. I think I am going to change the name of it to `TileQuarter` instead of `TrackTileQuarter` as it is not used only for tracks. The only annoying thing left is the fact that a map element has a half `TileQuarter` shoved into its `flags`.


IntelOrca:
This should probably be merged asap. After rebase.

PR #8728, opened 2019-02-19
Refactor peep_interact_with_path to prevent null ride warnings

IntelOrca:
My recent `get_ride` refactor meant this function was raising warnings frequently. I have refactored the method so it only calls get_ride when necessary which in most cases should be when ride index is not 255. I also reduced the number of returns moving closer to "single entry, single exit". It should be easier to follow now.


IntelOrca:
`outside_of_park` is defined as a `uint8_t` and there are many references to it for `==` and `!=`. It should be changed in a separate PR, although probably after my sprite import / export field by field refactor.

PR #8733, opened 2019-02-20
Fix pause

duncanspumpkin:


src/openrct2/actions/GameAction.cpp
```diff
@@ -154,7 +154,7 @@ namespace GameActions
 
         if (result->Error == GA_ERROR::OK)
         {
-            if (finance_check_affordability(result->Cost, action->GetFlags()) == false)
+            if (result->Cost != 0 && finance_check_affordability(result->Cost, action->GetFlags()) == false)
```
ZehMatt: Shouldn't this be checked within finance_check_affordability perhaps?

duncanspumpkin: Hmm yeah maybe.

PR #8735, opened 2019-02-20
Refactor sprite import / export

IntelOrca:
Import and export all sprites on a field by field basis so that we are free to change the in-game structs for sprites which will help with moving to new save format.

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -1174,7 +1174,7 @@ private:
         dst->ride_subtype = ride->subtype;
 
         dst->vehicle_type = vehicleEntryIndex;
-        dst->is_child = src->is_child;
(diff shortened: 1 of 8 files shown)


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1174,7 +1174,7 @@ class S4Importer final : public IParkImporter
         dst->ride_subtype = ride->subtype;
 
         dst->vehicle_type = vehicleEntryIndex;
-        dst->is_child = src->is_child;
+        dst->is_child = src->type;
```
Gymnasiast: Perhaps it's a good idea to also rename `is_child` to `type` on the regular sprite.

IntelOrca: I have already done that on my follow up PR along with a new getter method, IsChild().

PR #8737, opened 2019-02-21
Add LoadOrQuitAction

duncanspumpkin:
I'm not completely sure why this has to be a game action. But I'm not going to rock the boat.
This command isn't networked so I'm not sure if there is a point to doing the serialisation method. Maybe useful for replay?


ZehMatt:
Yeah for replays to work you need to serialise the data but I highly doubt this should be replayed to begin with. Well for server logging perhaps?

src/openrct2/actions/LoadOrQuitAction.hpp
```diff
@@ -0,0 +1,74 @@
...
+{
+    OpenSavePrompt,
+    CloseSavePrompt
+};
+
+DEFINE_GAME_ACTION(LoadOrQuitAction, GAME_COMMAND_LOAD_OR_QUIT, GameActionResult)
+{
+private:
+    uint8_t _mode{ 0 };
+    uint8_t _savePromptMode{ 0 };
```
ZehMatt: I think this should be a bool.

src/openrct2/actions/LoadOrQuitAction.hpp
```diff
@@ -0,0 +1,74 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2018 OpenRCT2 developers
```
ZehMatt: 2019

src/openrct2/actions/LoadOrQuitAction.hpp
duncanspumpkin: Turns out its of type enum `PROMPT_MODE` so I've kept it as `uint8_t` and add the appropriate constants.

ZehMatt: I see, since it was either 0 or 1 I sort of assumed its nothing more than that.

ZehMatt:
Do we need for CLIENT_ONLY a new network version? The server doesn't care about this action.

duncanspumpkin:
No probably don't need to make it a network increment

Gymnasiast:
The code suggests that the server can send this to the clients, but not the other way round.

ZehMatt:
Can you point at the given code? CLIENT_ONLY will only act locally.

Gymnasiast:
@ZehMatt Meh, ignore what I said, it was just what I gathered from what the old code did.

Has this been tested with a server and client, closing the server down, and closing the client down?

duncanspumpkin:
Yes its tested on both.

PR #8738, opened 2019-02-21
Implement RideSetVehicleAction

duncanspumpkin:
Move all callees to the new action.
Required rework on how the ride type change worked.


src/openrct2/actions/RideSetVehiclesAction.hpp
```diff
@@ -0,0 +1,276 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2018 OpenRCT2 developers
```
ZehMatt: 2019

src/openrct2/actions/RideSetVehiclesAction.hpp
```diff
@@ -0,0 +1,276 @@
...
+        : _rideIndex(rideIndex)
+        , _type(static_cast<uint8_t>(type))
+        , _value(value)
+        , _colour(colour)
+    {
+    }
+
+    uint16_t GetActionFlags() const override
+    {
+        return GameAction::GetActionFlags();
```
ZehMatt: Missing GA_FLAGS::ALLOW_WHILE_PAUSED

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1960,13 +1961,17 @@ static money32 place_track_design(int16_t x, int16_t y, int16_t z, uint8_t flags
 
     if (entryIndex != 0xFF)
     {
-        game_do_command(0, flags | (2 << 8), 0, ride->id | (entryIndex << 8), GAME_COMMAND_SET_RIDE_VEHICLES, 0, 0);
+        auto colour = ride_get_unused_preset_vehicle_colour(entryIndex);
+        auto rideSetVehicleAction = RideSetVehicleAction(ride->id, RideSetVehicleType::RideEntry, entryIndex, colour);
+        GameActions::ExecuteNested(&rideSetVehicleAction);
```
duncanspumpkin: Missing set flags!

duncanspumpkin: Flags it seems are not used in this function but there should have been an QueryNested if.

ZehMatt: Well the game action base class uses some flags to determine the game action flags, but yeah since its nested it doesn't matter.

PR #8740, opened 2019-02-22
Ride Set Setting GameAction

duncanspumpkin:
Another of the game actions


src/openrct2/actions/RideSetSetting.hpp
```diff
@@ -0,0 +1,331 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2018 OpenRCT2 developers
```
ZehMatt: 2019

src/openrct2/actions/RideSetSetting.hpp
```diff
@@ -0,0 +1,331 @@
...
+    RideSetSettingAction(ride_id_t rideIndex, RideSetSetting setting, uint8_t value)
+        : _rideIndex(rideIndex)
+        , _setting(static_cast<uint8_t>(setting))
+        , _value(value)
+    {
+    }
+
+    uint16_t GetActionFlags() const override
+    {
+        return GameAction::GetActionFlags();
```
ZehMatt: GA_FLAGS::ALLOW_WHILE_PAUSED missing

duncanspumpkin: I don't think you should be able to change these settings whilst paused.

PR #8746, opened 2019-02-23
Fix #8739: savegame from original game crashes when cruising through map

### LINKED ISSUE #8739: savegame from original game crashes when cruising through map
b7db6cb
MacOS

hey guys,

first of all, thank you all so much for developing may childhood fav game. 
Im currently playing on OpenRCT2-0.2.1-develop-b7db6cb-macos, but had this issue with previous builds of 0.21 (before that I didnt know about openrct).

when Im browsing through a map of certain old save games, the game crashes instantly. I found out, that it depended on certain attractions that the game couldn't show. I have the clipboard of the error that macOS gave me:

Application Specific Information:
Assertion failed: (false), function vehicle_visual_splash_effect, file /Users/travis/build/OpenRCT2/OpenRCT2/src/openrct2/ride/VehiclePaint.cpp, line 3096.
( i have more, if you need the longer version just say it)

The attractions which caused the crash were "launched free fall" (this is weird, because I can build a new free fall tower within this exact loaded game and it doesnt crash) and "observation tower".

would be great if you guys can fix this, so my games wont crash anymore!

thanks in advance!

IntelOrca:


duncanspumpkin:
Which part fixes it? Or is it both?

IntelOrca:
Mainly the switch but I also made sure it didn't go out of bounds when selecting the vehicle entry to use. It will still might use a undefined vehicle since we don't keep a count of how many there are in a ride entry but at least it won't go past the size of the array.

PR #8761, opened 2019-02-24
Fix being unable to refund when in negative money.

duncanspumpkin:


duncanspumpkin:
Thanks. I went out for a cycle and forgot I still hadn't bumped the version.

PR #8768, opened 2019-02-24
Fix interactive console staff set costume command.

duncanspumpkin:
After the refactor to GameActions, forgot to actually issue the command after generating it.


duncanspumpkin:
No servers will still accept the commands.

Gymnasiast:
Shouldn't this do a network version bump?

PR #8772, opened 2019-02-25
Refactor in-game sprite structs

IntelOrca:
In particular, change all sprite structs to inherit a common base and remove all the padding.

### DIFF
--- a/src/openrct2/ride/CableLift.cpp
+++ b/src/openrct2/ride/CableLift.cpp
@@ -39,7 +39,7 @@ rct_vehicle* cable_lift_segment_create(
         move_sprite_to_list((rct_sprite*)current, SPRITE_LIST_TRAIN * 2);
         ride->cable_lift = current->sprite_index;
     }
-    current->is_child = head ? 0 : 1;
+    current->type = head ? 0 : 1;
     current->var_44 = var_44;
     current->remaining_distance = remaining_distance;
     current->sprite_width = 10;
--- a/src/openrct2/ride/CableLift.cpp
+++ b/src/openrct2/ride/CableLift.cpp
@@ -39,7 +39,7 @@ rct_vehicle* cable_lift_segment_create(
         move_sprite_to_list((rct_sprite*)current, SPRITE_LIST_TRAIN * 2);
         ride->cable_lift = current->sprite_index;
     }
-    current->type = head ? 0 : 1;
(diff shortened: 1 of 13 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -3526,7 +3526,7 @@ static void vehicle_update_collision_setup(rct_vehicle* vehicle)
     if (!(ride->lifecycle_flags & RIDE_LIFECYCLE_CRASHED))
     {
         rct_vehicle* frontVehicle = vehicle;
-        while (frontVehicle->is_child != 0)
+        while (frontVehicle->IsChild() != 0)
```
Gymnasiast:
```suggestion
        while (frontVehicle->IsChild())
```

duncanspumpkin:
Should we drop `rct_` from the names or is that too big a change.

Gymnasiast:
We definitely should, since they don't even match their RCT1/2 counterparts any more.

IntelOrca:
Dropping the `rct_` will cause at least 5000 and in total almost 10000 changes across 388 files. Should we do that in it's own PR?

Gymnasiast:
@IntelOrca Sure, that would make reviewing easier.

ZehMatt:
It could be a general code maintenance PR where it also updates all our copyright dates since we already touch hundreds of files.

src/openrct2/ride/CableLift.cpp
```diff
@@ -39,7 +39,7 @@ rct_vehicle* cable_lift_segment_create(
         move_sprite_to_list((rct_sprite*)current, SPRITE_LIST_TRAIN * 2);
         ride->cable_lift = current->sprite_index;
     }
-    current->is_child = head ? 0 : 1;
+    current->type = head ? 0 : 1;
```
Gymnasiast: It would be better to use an enum for `type`.

IntelOrca: How can I do that when type is different for each sprite identifier?

Gymnasiast: I meant to say "define an enum with constants for 0 and 1, and leave type as an uint8_t".

IntelOrca: done

PR #8773, opened 2019-02-25
Implement SmallSceneryPlaceAction

duncanspumpkin:
Not quite finished yet.


src/openrct2/actions/SmallSceneryPlaceAction.hpp
```diff
@@ -0,0 +1,422 @@
...
+            baseHeight = _loc.z;
+            res->Position.z = baseHeight;
+        }
+
+        if (!map_check_free_elements_and_reorganise(1))
+        {
+            return MakeResult(GA_ERROR::NO_FREE_ELEMENTS, STR_CANT_POSITION_THIS_HERE);
+        }
+
+        if (!byte_9D8150 && (_loc.x > gMapSizeMaxXY || _loc.y > gMapSizeMaxXY))
```
ZehMatt: I believe byte_9D8150 is set when it wants the preview only. We could call this something like 'previewMode' but thats better for a later PR.

duncanspumpkin: Yes that is correct. I thought about that as well. Really it should be passed as a game command flag but I'll wait until we are doing the track place action to change it.

PR #8774, opened 2019-02-25
Use IsGhost() to access ghost state

duncanspumpkin:
Added set and clear ghost to further prevent direct access to flags

### DIFF
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -2185,7 +2185,7 @@ static void window_tile_inspector_scrollpaint(rct_window* w, rct_drawpixelinfo* 
 
         const int32_t baseHeight = tileElement->base_height;
         const int32_t clearanceHeight = tileElement->clearance_height;
-        const bool ghost = (tileElement->IsGhost()) != 0;
+        const bool ghost = tileElement->IsGhost();
         const bool broken = (tileElement->flags & TILE_ELEMENT_FLAG_BROKEN) != 0;
         const bool last = (tileElement->flags & TILE_ELEMENT_FLAG_LAST_TILE) != 0;
 
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -2185,7 +2185,7 @@ static void window_tile_inspector_scrollpaint(rct_window* w, rct_drawpixelinfo* 
 
         const int32_t baseHeight = tileElement->base_height;
         const int32_t clearanceHeight = tileElement->clearance_height;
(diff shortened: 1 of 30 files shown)


src/openrct2/paint/tile_element/Paint.Entrance.cpp
```diff
@@ -157,7 +157,7 @@ static void ride_entrance_exit_paint(paint_session* session, uint8_t direction,
         paint_util_push_tunnel_left(session, height, TUNNEL_6);
     }
 
-    if (!is_exit && !(tile_element->flags & TILE_ELEMENT_FLAG_GHOST) && tile_element->AsEntrance()->GetRideIndex() != 0xFF
+    if (!is_exit && !(tile_element->IsGhost()) && tile_element->AsEntrance()->GetRideIndex() != 0xFF
```
Gymnasiast:
```suggestion
    if (!is_exit && !(tile_element->IsGhost()) && tile_element->AsEntrance()->GetRideIndex() != RIDE_ID_NULL
```

src/openrct2/paint/tile_element/Paint.Entrance.cpp
```diff
@@ -157,7 +157,7 @@ static void ride_entrance_exit_paint(paint_session* session, uint8_t direction,
...
         && stationObj->ScrollingMode != 0xFF)
```
Gymnasiast:
```suggestion
        && stationObj->ScrollingMode != SCROLLING_MODE_NONE)
```

ZehMatt:
Is there a reason we use two functions as in, Set/Clear. I would rather see SetGhost(bool state) imho.

duncanspumpkin:
I thought of having just one but didn't like the inefficiency of testing the bool. Besides it's much clearer setghost/clearghost than setghost true/false

ZehMatt:
Most systems I have worked with usually just have basic Get/Set rather than Get/Set/Clear but thats way too trivial to argue about, its just my opinion, but I also see your point of checking before applying.

Gymnasiast:
The thing is that all tile element methods that I introduced use @ZedMatt's approach. I think it's more important to be consistent.

duncanspumpkin:
Okay will change it

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -2185,7 +2185,7 @@ static void window_tile_inspector_scrollpaint(rct_window* w, rct_drawpixelinfo*
 
         const int32_t baseHeight = tileElement->base_height;
         const int32_t clearanceHeight = tileElement->clearance_height;
-        const bool ghost = (tileElement->flags & TILE_ELEMENT_FLAG_GHOST) != 0;
+        const bool ghost = (tileElement->IsGhost()) != 0;
```
jensj12: No need for `!= 0` anymore here.

Issue #8318, opened 2018-11-30
BACKTRACE 83f40c ride_shop_connected -> scenario_week_update


IntelOrca:
<table>
    <tr>
        <td><img src='https://backtrace.io/images/icon.png' width=30 height=30 /></td>
        <td><a href='https://openrct2.sp.backtrace.io/dashboard/openrct2/project/openrct2/group/83f40c982c5cf4f775e590b9c9fb4aa3e24eda2df1bd86e3e136f02dd29f6607?_t=W3sibmFtZSI6IndvcmtmbG93In1d'><b>New unique error: [openrct2]</b></a><br /><b>1</b> occurrence(s) <b></b> since (Fri Nov 30 2018 02:58:12)<br /></td>
    </tr>
</table>

<table>
<tr><td><strong>Application</strong></td><td>openrct2.exe</td></tr>

<tr><td><strong>Attributes</strong></td><td><strong>attachment_park.sv6:</strong> *: <em>1</em><br><strong>attachment_config.ini:</strong> *: <em>1</em><br><strong>attachment_park.sv6.gz:</strong> E:\\Documents\\OpenRCT2\\d6f7a206-829b-41fb-a444-1299af10f629.sv6.gz: <em>1</em><br><strong>commit:</strong> eeaca61: <em>1</em><br></td></tr>
<tr><td><strong>Classifiers</strong></td><td><strong>invalid-read</strong> </td></tr>
<tr><td><strong>Frames</strong></td><td><code>ride_shop_connected -> scenario_week_update -> scenario_update -> OpenRCT2::GameState::UpdateLogic -> OpenRCT2::GameState::Update -> OpenRCT2::Context::Update -> OpenRCT2::Context::RunVariableFrame -> OpenRCT2::Context::RunGameLoop -> OpenRCT2::Context::RunOpenRCT2 -> NormalisedMain -> LaunchOpenRCT2</code></td></tr>
</table>

janisozaur:
Unsure, I couldn't reproduce it on `develop` or eeaca61, but I don't recall any fix in that area.

PR #8776, opened 2019-02-25
Fix ride_shop_connected dereferencing null.

ZehMatt:
I did not find the issue on the tracker the closest was #8318 which seems to be the same problem. Also given that I'm not sure how to reproduce this error and which realm is affected should I still bump the network version to be safe?

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -3463,7 +3463,6 @@ static void ride_entrance_exit_connected(Ride* ride)
 static void ride_shop_connected(Ride* ride)
 {
     int32_t x, y, count;
-    TileElement* tileElement;
 
     LocationXY8 coordinates = ride->stations[0].Start;
     if (coordinates.xy == RCT_XY8_UNDEFINED)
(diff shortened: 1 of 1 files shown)


IntelOrca:
You also need to return if no track element is found. At the moment it just falls through.

ZehMatt:
Not sure I follow, where should I return? I'm checking if the tileElement is null at the beginning and post-loop I validate if the specific element is the track piece and is associated with the ride if not it returns.

src/openrct2/ride/Ride.cpp
```diff
@@ -3472,13 +3471,24 @@ static void ride_shop_connected(Ride* ride)
...
-    tileElement = map_get_first_element_at(x, y);
+    TileElement* tileElement = map_get_first_element_at(x, y);
+    if (tileElement == nullptr)
+    {
+        return;
+    }
+
     do
     {
         if (tileElement->GetType() == TILE_ELEMENT_TYPE_TRACK && tileElement->AsTrack()->GetRideIndex() == ride->id)
```
duncanspumpkin: Break this up into two if statements much easier to read. I would introduce a bool found and then check that after the loop

ZehMatt: Ha the good old trick from the game actions huh?

duncanspumpkin: Yeah

ZehMatt: I did it slightly different which avoids two further calls to "AsTrack()" by simply having a TrackElement pointer being assigned or not.

PR #8778, opened 2019-02-25
Remove redundant move

janisozaur:


duncanspumpkin:
I think i even read somewhere that this can be less efficient as it means the compiler has less optimisation options

PR #8781, opened 2019-02-26
Flag refactor

duncanspumpkin:
Continues on from the IsGhost PR.

Found one or two incorrect uses of flags that have been corrected. No change in function just the wrong flag name that has the same value.

### DIFF
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -2186,7 +2186,7 @@ static void window_tile_inspector_scrollpaint(rct_window* w, rct_drawpixelinfo* 
         const int32_t baseHeight = tileElement->base_height;
         const int32_t clearanceHeight = tileElement->clearance_height;
         const bool ghost = tileElement->IsGhost();
-        const bool broken = (tileElement->flags & TILE_ELEMENT_FLAG_BROKEN) != 0;
+        const bool broken = tileElement->AsPath() == nullptr ? false : tileElement->AsPath()->IsBroken();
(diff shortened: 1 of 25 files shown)


src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -2186,8 +2186,8 @@ static void window_tile_inspector_scrollpaint(rct_window* w, rct_drawpixelinfo*
         const int32_t baseHeight = tileElement->base_height;
         const int32_t clearanceHeight = tileElement->clearance_height;
         const bool ghost = tileElement->IsGhost();
-        const bool broken = (tileElement->flags & TILE_ELEMENT_FLAG_BROKEN) != 0;
-        const bool last = (tileElement->flags & TILE_ELEMENT_FLAG_LAST_TILE) != 0;
+        const bool broken = tileElement->AsPath() == nullptr ? false : tileElement->AsPath()->IsBroken();
```
Broxzier: Not only path elements have the broken flag, track elements can be broken too, and they share the same bit. This change does not look right.

duncanspumpkin: They aren't broken though they are closed block brakes. Does anyone use this to check the state of block brakes?

Broxzier: Not sure, though it's misleading if the tile inspector shows that the "broken" flag is not set while it really is. If no other elements use it, it may be better to move that information from the column to their respective information group boxes instead.

duncanspumpkin: Yes that would make more sense. I'll make an issue for it as I don't want this PR to modify too much.

Issue #8782, opened 2019-02-26
Version info in Travis builds is incomplete


AaronVanGeffen:
The git environment on Travis does not seem to be working as expected. `git describe` yields an empty string, and branch is set to HEAD instead.

Expected version line (local build):
```
OpenRCT2, v0.2.1-903-g5b0367 (g5b0367 on master, DEBUG)
```

Instead reads (Travis build):
```
OpenRCT2,  (g5b0367 on HEAD, DEBUG)
```

This is probably related to git clone depth: https://github.com/travis-ci/travis-ci/issues/7422. Hence, releases weren't affected when tested.

We may have to run `git fetch --unshallow` to get this info to show up properly, but I'd prefer not to do that as it probably slows down the CI builds quite a bit.

Related issue: https://github.com/OpenLoco/OpenLoco/issues/283

AaronVanGeffen:
| Command | # Commits |
| ---- | --- |
| git rev-list v0.0.1..v0.0.2 --count   | 3573 |
| git rev-list v0.0.2..v0.0.3 --count   | 1595 |
| git rev-list v0.0.3..v0.0.3.1 --count | 11   |
| git rev-list v0.0.3.1..v0.0.4 --count | 1534 |
| git rev-list v0.0.4..v0.0.5 --count   | 3592 |
| git rev-list v0.0.5..v0.0.6 --count   | 453  |
| git rev-list v0.0.6..v0.0.7 --count   | 624  |
| git rev-list v0.0.7..v0.1.0 --count   | 730  |
| git rev-list v0.1.0..v0.1.1 --count   | 284  |
| git rev-list v0.1.1..v0.1.2 --count   | 1756 |
| git rev-list v0.1.2..v0.2.0 --count   | 795  |
| git rev-list v0.2.0..v0.2.1 --count   | 334  |
| git rev-list v0.2.1..develop --count   | 890  |

I think we'd be fine if we increased the depth to the last 1000 commits.

Travis only allows specifying up to 50 commits, however. We _can_ manually run `git fetch --depth=1000`, or indeed go all the way with `git fetch --unshallow`.

Thoughts? @janisozaur

duncanspumpkin:
What are the implications of the change? Does it just take longer to compile? Are we talking seconds, minutes, hours?

AaronVanGeffen:
[...]

PR #8788, opened 2019-02-27
Fetch more repository history when building on CI

AaronVanGeffen:
Proposed fix for #8782.


AaronVanGeffen:
```
$ if [[ $TRAVIS_JOB_NAME != "clang-format" ]]; then git config remote.origin.fetch +refs/heads/*:refs/remotes/origin/*; fi
install.2
$ if [[ $TRAVIS_JOB_NAME != "clang-format" ]]; then git fetch --depth=1000; fi
```

My original attempt. Takes 20.43s, but just fetches the last 1000 commits.

```
$ if [[ $TRAVIS_JOB_NAME != "clang-format" ]]; then git config remote.origin.fetch +refs/tags/*:refs/tags/*; fi
install.2
$ if [[ $TRAVIS_JOB_NAME != "clang-format" ]]; then git fetch; f
```

Takes just 17.19s, and isn't tied to a constant number of commits. I think we have a winner.

PR #8790, opened 2019-02-27
Refactor peep struct

IntelOrca:


Gymnasiast:
I would just use `Guest` and `Staff`, rather than `GuestPeep` and `StaffPeep`.

IntelOrca:
Yes, some caution is needed though for example RemoveFromRide and RemoveFromQueue were assumed to be guest only but I think they apply to mechanics as well.

IntelOrca:
@Gymnasiast done

Issue #8055, opened 2018-10-05
Game falsly detects desync


Broxzier:
**Version:** 0.2.1
**Commit/Build:** 0231fd7

Usually shortly after joining a server, the game detects a desync. After staying connected for a couple of minutes, popping balloons still works. It would be way too coincidental that the sprite ID on the server matches the ones in my local game exactly.

- [ ] Reproducible in RCT2 (vanilla)?
- [x] Specific to multiplayer?

**Steps to reproduce:**

0. Enable the option to stay connected after desyncing
1. (Host and then) join a multiplayer server
2. Wait for the desync to happen (usually only takes a few seconds on my end)
3. Wait for a couple of minutes
4. Perform any game action that works with Sprite IDs, and notice how it works perfectly fine even for sprites that spawned minutes after the desync was detected.

Broxzier:
Unfortunately, it still happened on 46ae093.

AaronVanGeffen:
https://github.com/OpenRCT2/OpenRCT2/pull/8803 hasn't been merged yet. As I understand it, it should be fixed when it is?

duncanspumpkin:
#8803 is just an additional test for the ci's the fix was already merged.

ZehMatt:
The park loaded might play a role in this. I used to get desyncs quite consistently doing this:
"openrt2 host parks/electric-fields.sv6"
"openrct2 join 127.0.0.1"

Which seems to be fixed with the export/import changes.

Broxzier:
After a bunch of testing and trying things out with @ZehMatt we found out that this has to do with **autosaves**. When hosting a self-made save (possible requiring objects to be exported) there is no desync, but when hosting from an autosave it does desync immediately.

PR #8792, opened 2019-02-27
Temporary fix for false desyncs.

ZehMatt:
One issue has been that the network_update was called from within LogicUpdate which might be used from the TitleScreen->Update, so after the map load it would be still on the title screen state and desyncs. Thats why network_update was moved out from there. The other reason for the desync remains unknown, in this PR it is required to have two ticks mismatch before its considered fully desynced which seems to solve the issue of #8055 
I also removed some unnecessary code, the function NETWORK_SEND_TICK is only called from game ticks, no need to double time it.

I have to yet figure out why its doing the false ones, for now this seems to do the job to hide the false positives.

### DIFF
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -306,6 +306,7 @@ private:
     std::vector<uint8_t> chunk_buffer;
     std::string _password;
     bool _desynchronised = false;
+    bool _desyncStarted = false;
     uint32_t server_connect_time = 0;
     uint8_t default_group = 0;
     uint32_t game_commands_processed_this_tick = 0;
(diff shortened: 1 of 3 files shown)


src/openrct2/network/Network.cpp
```diff
@@ -965,20 +964,31 @@ bool Network::CheckSRAND(uint32_t tick, uint32_t srand0)
...
+                intent.putExtra(INTENT_EXTRA_MESSAGE, std::string{ str_desync });
+                context_open_intent(&intent);
 
-        if (!gConfigNetwork.stay_connected)
+                if (!gConfigNetwork.stay_connected)
+                {
+                    Close();
+                }
+            }
+            _desyncStarted = true;
```
Broxzier: It'd be nice to explain in a comment why this bool is necessary, possible with link to this PR. For someone who doesn't know what this is about, it appears to be simply skipping one tick before truely desyncing.

ZehMatt:
Looks like the root cause for this specific desync is actually the export/import as seen here on the tests: #8803

ZehMatt:
This is no longer required. The real fix was done in #8812

PR #8796, opened 2019-02-28
Entranceexit place ga

duncanspumpkin:
Following on from remove.

For some reason track design place was using a game action when it really didn't need to. I've created a static function for it to call instead of using the game action framework for that situation.


src/openrct2/actions/RideEntranceExitPlaceAction.hpp
```diff
@@ -0,0 +1,338 @@
...
+        , _direction(direction)
+        , _rideIndex(rideIndex)
+        , _stationNum(stationNum)
+        , _isExit(isExit)
+    {
+    }
+
+    uint16_t GetActionFlags() const override
+    {
+        return GameAction::GetActionFlags();
```
ZehMatt: I couldn't find any place where the GAME_COMMAND_ALLOW_WHILE_PAUSED flag was set conditionally, could add it here and ditch the passing of the flag.

duncanspumpkin: You shouldn't be able to place a entranceexit when in paused but you should be able to show a ghost. I think thats the reason behind this.

ZehMatt: Now that you mention that, I wonder if we should always add the allow paused action flag if the ghost flag is specified.

duncanspumpkin: Would have to run some tests but I think that makes sense. Maybe we need to have some sort of central location that specifies all the implications each flag has.

ZehMatt: Agreed

src/openrct2/actions/RideEntranceExitPlaceAction.hpp
```diff
@@ -0,0 +1,338 @@
...
+        {
+            return MakeResult(GA_ERROR::NO_CLEARANCE, errorTitle, gGameCommandErrorText, gCommonFormatArgs);
+        }
+
+        if (gMapGroundFlags & ELEMENT_IS_UNDERWATER)
+        {
+            return MakeResult(GA_ERROR::DISALLOWED, errorTitle, STR_RIDE_CANT_BUILD_THIS_UNDERWATER);
+        }
+
+        if (z / 8 > 244)
```
ZehMatt: Magic value, could define this in Map.h like the other GA.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1251,9 +1252,8 @@ static int32_t track_design_place_maze(rct_track_td6* td6, int16_t x, int16_t y,
 
                     if (_trackDesignPlaceOperation == PTD_OPERATION_1)
                     {
-                        cost = game_do_command(
-                            mapCoord.x, 0 | rotation << 8, mapCoord.y, ((z / 16) & 0xFF) | (1 << 8),
-                            GAME_COMMAND_PLACE_RIDE_ENTRANCE_OR_EXIT, -1, 0);
+                        auto res = RideEntranceExitPlaceAction::TrackPlaceQuery({ mapCoord.x, mapCoord.y, z }, true);
```
ZehMatt: How come this requires a static function rather than calling QueryNested?

duncanspumpkin: The game command was really odd in that it called a separate part of the query code and never called the execute section of that code. It didn't make sense for it to be a game command. You will notice it takes slightly different parameters and validates them differently. I'm not sure what the best approach would be.

ZehMatt: Hm yeah, I can only think of separating the path in Query for when the ride id is -1 (0xFF) but then its quite the annoyance to pass temporary parameters for everything else.

ZehMatt:
Testing shows that it works as expected, only one little nitpick and its good to go from my end.

src/openrct2/actions/RideEntranceExitPlaceAction.hpp
```diff
@@ -0,0 +1,285 @@
...
+        {
+            return MakeResult(GA_ERROR::NO_CLEARANCE, errorTitle, gGameCommandErrorText, gCommonFormatArgs);
+        }
+
+        if (gMapGroundFlags & ELEMENT_IS_UNDERWATER)
+        {
+            return MakeResult(GA_ERROR::DISALLOWED, errorTitle, STR_RIDE_CANT_BUILD_THIS_UNDERWATER);
+        }
+
+        if (baseZ > 244)
```
ZehMatt: MaxRideEntranceOrExitHeight

PR #8797, opened 2019-02-28
Fix Ride Entrance Exit Removal Network Issue

duncanspumpkin:
Wrong sized parameters used in DataSerialiser caused bad data to be input into game action.


duncanspumpkin:
Integrated into #8796

PR #8803, opened 2019-02-28
Add network load and save test.

ZehMatt:
This tests the exact same path as you would have during network play. The server exports the map data and includes the spatial sprite indices so the client will have the same linked list as the server. This basically loads a map, exports it as the network variant with the checksum and imports it right away and compares if the checksum matches the one used before exporting, this ensures that server and client will have the same state.


duncanspumpkin:
Is this now working reliably? It would be good to have some more tests. (not sure why I never get round to making any...)

ZehMatt:
I just force pushed the rebased version, seems like its still failing. I will keep digging until this thing passes.

ZehMatt:
Now its looking good.

PR #8806, opened 2019-03-01
Implement rate limiting times for game actions.

ZehMatt:


duncanspumpkin:
Could you add the overrides for SmallSceneryPlace and DemolishRide as those were the current game command ones.

duncanspumpkin:
The game commands also allowed the server to not listen to the countdowns. Could we maybe have someway of allowing the server to vary this per user? Would that be useful?

ZehMatt:
1. Sure can do
2. I can't find any code that would suggest its ignoring it, but we can introduce that if I would know how to determine this.

ZehMatt:
This more or less leaves it the way it always has been, we can introduce cooldowns per groups/user in another PR. Might as well look at the whole permission/group code anyway.

src/openrct2/network/Network.cpp
```diff
@@ -667,6 +685,11 @@ void Network::Update()
 {
     _closeLock = true;
 
+    // Update is not unnecessarily called per game tick, maintain our own delta time
```
duncanspumpkin: necessarily

duncanspumpkin:
Maybe hold off until after release.

src/openrct2/actions/SmallSceneryPlaceAction.hpp
```diff
@@ -49,6 +49,11 @@ DEFINE_GAME_ACTION(SmallSceneryPlaceAction, GAME_COMMAND_PLACE_SCENERY, GameActi
     {
     }
 
+    uint32_t GetCooldownTime() const override
+    {
+        return 20;
```
ZehMatt: I wonder if this is correct, I took the value that was previously used but this doesn't even cover a single game tick of 25ms.

duncanspumpkin: It's to block the cluster tool that issues 16+ actions in one tick.

ZehMatt: I see, I'm not even sure if this still works when nested actions are used since this is only checked at network level.

duncanspumpkin: Cluster isn't a nested action (yet).

ZehMatt: Just saying, also is this ready to be merged?

duncanspumpkin:
I've not tested it but looks fine

ZehMatt:
Should be fine, just did a final test.

PR #8807, opened 2019-03-01
Footpath GameAction

duncanspumpkin:


duncanspumpkin:
I've tested this quite a bit on network (client and server) and non network and I think it is ready for merging. This does change how footpath scenery works. It will now no longer reconnect path edges 
https://files.gitter.im/OpenRCT2/OpenRCT2/9vlq/image.png
In above image I removed the seats marked. Before this would have reconnected the paths. This was not helpful when you spent time getting the paths connected how you like.

ZehMatt:
2019 :)

ZehMatt:
It needs to update the network version but it looks otherwise good to me except those copyright dates, we should probably update them all at once so we will stop copying the outdated ones.

PR #8814, opened 2019-03-02
Always link threads, fix linking cURL

### LINKED ISSUE #8810: -DDISABLE_NETWORK=ON requires pthreads too
**OS:** Linux

Building with -DDISABLE_NETWORK=ON results in error "undefined reference to `pthread_create'"

tomlankhorst:
Threads are currently used throughout the codebase so they need to be
required. CURL is required when using HTTP, not when using Networking.

Fixes #8810

Current situation:
- `DISABLE_HTTP` Disables cURL and `Http.cpp`.
- `DISABLE_NETWORK` Disabled `Network.cpp` and friends.

I think it's a bit vague that disabling networking does not disable HTTP. 
Maybe, we could refactor it to `DISABLE_MULTIPLAYER`?


Gymnasiast:
This will need to have the changelog updated as well, but looks fine otherwise.

PR #8820, opened 2019-03-03
Fix #8804: Raising water shows money effect at the bottom

### LINKED ISSUE #8804: Raising water shows money effect at the bottom rather than new height.
2019-03-01_08-12-14950eca2339200d9f5

The original doesn't do this, instead it would show the money at the new height.

AaronVanGeffen:


AaronVanGeffen:
Should it be added to the changelog as well?

duncanspumpkin:
Probably.

PR #8837, opened 2019-03-09
Fix more of these backtrace reports

IntelOrca:


src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -578,12 +575,15 @@ static void window_footpath_invalidate(rct_window* w)
         : WWT_EMPTY;
 
     // Set footpath and queue type button images
-    selectedPath = gFootpathSelectedId;
-    pathType = get_path_surface_entry(selectedPath);
-
-    int32_t pathImage = pathType->preview;
-    // Editor-only paths might lack a queue image
-    int32_t queueImage = (pathType->flags & FOOTPATH_ENTRY_FLAG_SHOW_ONLY_IN_SCENARIO_EDITOR) ? pathImage : pathImage + 1;
+    auto pathImage = (uint32_t)SPR_NONE;
```
AaronVanGeffen:
```suggestion
    uint32_t pathImage = SPR_NONE;
```

IntelOrca: That won't compile on VS, the cast is required because `SPR_NONE` is -1.

PR #8842, opened 2019-03-10
README: Add Julius to similar projects

devnoname120:


AaronVanGeffen:
The selection of similar projects is rather arbitrary, but I think the row is already quite wide enough as-is. I don't think we should extend it.

Not meant at the detriment of the Julius project, of course.

Gymnasiast:
Indeed. We had a similar request lately.

janisozaur:
Maybe we should have a catch-all case that links to https://osgameclones.com and https://en.m.wikipedia.org/wiki/List_of_open-source_video_games ?

devnoname120:
I understand. I find osgameclones a bit hard to navigate as it doesn't properly distinguish high quality projects with a full reimplementation such as Julius from half-done and buggy implementations.

Gymnasiast:
I suppose we could take a third way and have a page on the wiki, listing only the better-quality examples. Though that in itself in hard to define, and could lead to a big list that everyone just piles stuff onto.

PR #8845, opened 2019-03-10
Fix #8524: Do not show RCT1 land edges when CSG1 is not loaded

### LINKED ISSUE #8524: Vertical Edges Not Showing
**OS:** Windows 10, 1803
**Version:** v0.2.1-474 
**Commit/Build:** (2530cd1 on develop)

I just noticed that when I try to add vertical edges to land, that only the plain, wood, black wood, and ice are showing.  If I try to apply any other edge, such as stone, metal, colored wall, skyscraper, etc. it charges me the money but doesn't change.  

Also, the game displays only those four existing edges.  I can load a scenario such as Razor Rocks that should have the grey walls, but they appear as normal dirt walls.  This is happening on new scenarios as well as saved games.  

I could be mistaken, but I think these were the only four edges available in all of RCT2.  In that case, the edges I am not seeing are the ones brought over from RCT1.  The "release" version of oRCT2 only has the original four available.  

I thought for sure the new vertical land faces were working earlier this week, but I could be wrong.  

Reproducible in RCT2 (vanilla)?  No
Specific to multiplayer? No

**Steps to reproduce:**
1.  Open a park in oRCT2 "develop"
2.  See the new land choices
3.  Try to apply one - does not change, but charges money

Please let me know if you have any questions.

Gymnasiast:


src/openrct2/object/ObjectJsonHelpers.cpp
```diff
@@ -513,12 +514,16 @@ namespace ObjectJsonHelpers
...
                     auto images = ParseImages(context, s);
+                    if (!images.empty())
+                        numImagesLoaded += images.size();
                     allImages.insert(
                         allImages.end(), std::make_move_iterator(images.begin()), std::make_move_iterator(images.end()));
                 }
                 else if (json_is_object(el))
                 {
                     auto images = ParseImages(context, el);
+                    if (!images.empty())
```
ZehMatt: I think this statement is not required, if its empty it will simply add 0.

src/openrct2/object/ObjectJsonHelpers.cpp
```diff
@@ -513,12 +514,16 @@ namespace ObjectJsonHelpers
                 {
                     auto s = json_string_value(el);
                     auto images = ParseImages(context, s);
+                    if (!images.empty())
```
ZehMatt: I think this statement is not required, if its empty it will simply add 0.

src/openrct2/object/TerrainEdgeObject.h
```diff
@@ -18,6 +18,7 @@ class TerrainEdgeObject final : public Object
     rct_string_id NameStringId{};
     uint32_t IconImageId{};
     uint32_t BaseImageId{};
+    uint32_t numImagesLoaded = 0;
```
ZehMatt: Could default initialize it as above.

IntelOrca: It should also be TitleCase.

IntelOrca:
I don't think you need any changes to ObjectJsonHelpers as you can just get the number of loaded images by doing `GetImageTable().GetCount()`

src/openrct2/object/TerrainEdgeObject.h
Gymnasiast: > Could default initialize it as above.

Initialising integers with `{}` is ugly, IMO.

Gymnasiast:
I will make changes tonight. I hope we can merge it then, in order to be in time for the release.

src/openrct2/object/TerrainEdgeObject.h
AaronVanGeffen: > Initialising integers with `{}` is ugly, IMO.

Maybe, but the other integers in the file are also default initialised. Consistency is good practice.

ZehMatt:
I've been wondering, if I were to join a server using RCT1 content and I only have RCT2 content would that cause a desync given that I don't have those sprites? If we use replacements/hiding them for the missing ones it will generate a different sprite checksum, is this correct?

src/openrct2/object/TerrainEdgeObject.h
Gymnasiast: Then we should change the code style, IMO. Anyway, I changed it.

PR #8852, opened 2019-03-11
Implement TrackSetBrakeSpeedAction

duncanspumpkin:


src/openrct2/actions/GameActionRegistration.cpp
```diff
@@ -51,6 +51,7 @@
 #include "TrackRemoveAction.hpp"
 #include "WallRemoveAction.hpp"
 #include "WaterSetHeightAction.hpp"
+#include "TrackSetBrakeSpeedAction.hpp"
```
ZehMatt: Gotta also register the action.

Gymnasiast:
Have you also tested if setting booster speeds still works?

duncanspumpkin:
This action is untested at present.

duncanspumpkin:
Part of #8878

PR #8855, opened 2019-03-11
Change call to strncat(dst, src, strlen(src)) to strcat(dst, src)

janisozaur:
Calling strncat with third argument equal strlen(src) triggers GCC's
stringop-overflow warning


ZehMatt:
strcat is the unsafe version however, it will [[maybe]] write past the buffer

IntelOrca:
Lets change it all to use `std::stringstream` 😄

PR #8860, opened 2019-03-12
Desync fixes.

ZehMatt:
One of the desyncs was caused by not actually using the correct tick, it would use server_srand0_tick which was only assigned when it has been 0 and would only be reset to 0 within the desync check, now we'll use the real server tick as everything should use the real server tick.
The other issue that this PR fixes is that it would check desyncs from within the title screen or previously loaded map where the ticks may match up to the server loaded map but will obviously desync since they are different things.


ZehMatt:
Doesn't fix it yet, ill have to investigate this further.

ZehMatt:
I can't reproduce this on any other server than @Broxzier . I've setup my own server and its not happening there with the exact save file, there might be a configuration quirk somewhere. I had to also actually buffer the tick and checksums as the client will never be at the server tick at the time it sends the tick packet.

ZehMatt:
Ran another test with @Krutonium on his Server and no desyncs upon connect.
2019-03-13_01-32-3097a689afa86a66e12

Gymnasiast:
I'd like to merge this. Could you rebase it and bump the network version to 1?

PR #8864, opened 2019-03-12
Apply better guards for download queueing

janisozaur:
Follow-up to #8821.
Use mutex to synchronize access rather than make the bool atomic to
avoid TOCTOU issues.

### DIFF
--- a/src/openrct2-ui/windows/ObjectLoadError.cpp
+++ b/src/openrct2-ui/windows/ObjectLoadError.cpp
@@ -34,6 +34,7 @@ private:
     std::vector<rct_object_entry> _downloadedEntries;
     size_t _currentDownloadIndex{};
     std::mutex _downloadedEntriesMutex;
+    std::mutex _queueMutex;
     bool _nextDownloadQueued{};
 
     // TODO static due to INTENT_EXTRA_CALLBACK not allowing a std::function
@@ -61,6 +62,7 @@ public:
 
     void Update()
     {
+        std::lock_guard guard(_queueMutex);
         if (_nextDownloadQueued)
         {
             _nextDownloadQueued = false;
@@ -87,6 +89,7 @@ private:
 
     void QueueNextDownload()
     {
+        std::lock_guard guard(_queueMutex);
         _nextDownloadQueued = true;
     }
 


IntelOrca:
What is the benefit of this?

IntelOrca:
@janisozaur ?

janisozaur:
Removal of the data race

IntelOrca:
I don't see any data race. You will need to explain.

The update is routinely it called, if it misses the result the first time, it will see it the second time.

janisozaur:
`ObjectDownloader::Update()` is being called from the UI thread through https://github.com/OpenRCT2/OpenRCT2/blob/6911a7f3b0391d1433023a2aafd6bd8e513fca08/src/openrct2-ui/windows/ObjectLoadError.cpp#L414 and reads from `_nextDownloadQueued` https://github.com/OpenRCT2/OpenRCT2/blob/6911a7f3b0391d1433023a2aafd6bd8e513fca08/src/openrct2-ui/windows/ObjectLoadError.cpp#L66, whilst `ObjectDownloader::QueueNextDownload()` is being called from a helper thread through `Http::DoAsync`'s lambda: https://github.com/OpenRCT2/OpenRCT2/blob/6911a7f3b0391d1433023a2aafd6bd8e513fca08/src/openrct2-ui/windows/ObjectLoadError.cpp#L124 which writes to `_nextDownloadQueued`: https://github.com/OpenRCT2/OpenRCT2/blob/6911a7f3b0391d1433023a2aafd6bd8e513fca08/src/openrct2-ui/windows/ObjectLoadError.cpp#L93

IntelOrca:
Yep that is how it is supposed to work. Why is the mutex necessary?

janisozaur:
> that is how it is supposed to work

Then it requires mutex to not cause the data race on `_nextDownloadQueued`, which is currently being shared between threads without any locks whatsoever.

IntelOrca:
It doesn't need a lock, one thread sets (only to true), the other thread periodically checks to see if it has been set to true and is the only thing that sets it back to false.

ZehMatt:
Might be worth to use std::atomic::compare_exchange_weak/compare_exchange_strong instead, this has one producer and one consumer and since its only two states this will do nicely.

janisozaur:
@IntelOrca You're wrong. What you said would happen if there was a mutex or cmpxchg, like @ZehMatt suggested. What happens now instead is one thread sets it to false and then another one invokes undefined behaviour.

IntelOrca:
Is it not atomic?

janisozaur:
No, have you made it atomic?

IntelOrca:
I guess not, my apologies. It should use `std::atomic::compare_exchange_weak` like @ZehMatt suggested.

PR #8886, opened 2019-03-16
Split path surface objects into paths and queues

Gymnasiast:


Gymnasiast:
@duncanspumpkin I will open a separate PR for that.

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -613,14 +613,7 @@ static void window_footpath_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
         int32_t selectedPath = gFootpathSelectedId;
```
duncanspumpkin: int32_t selectedPath = `(gFootpathSelectedType << 7) +  gFootpathSelectedId;` otherwise it will only select the regular path.

Gymnasiast: I think that should be `<< 5` or `* 32`. There are also other approaches, though, since I haven't done the split 100%.

Issue #8688, opened 2019-02-07
List Tile Inspector, Object Selection and Inventions list in Cheats menu


Gymnasiast:
The three mentioned items are currently only in the debugging tools menu. Which is understandable, but it does mean that a lot of people have to turn on debugging tools to do things that aren't debugging. Especially if we're going to add more debugging tools (the peep debugging PR springs to mind), I want to change that.

My plan is to put the items in both menus, allowing access to them even if the user hasn't turned on debugging tools.

AaronVanGeffen:
I'd prefer the items to just be in the cheat menu, to be honest. Most devs have the cheat menu enabled anyway, as cheats are a help in debugging, too.

Gymnasiast:
Fine by me.

IntelOrca:
Tile inspector was designed as a debugging tool. Object Selection and Inventions list might be better under map and ride / research respectively?

AaronVanGeffen:
Hmm, but they're only unavailable when debugging tools (or cheats, for that matter) are unavailable, right? Filing them under different menus would hinder their discoverability, I think.

Gymnasiast:
> Tile inspector was designed as a debugging tool.

Certainly, but it has become a very useful cheating tool as well.

As @AaronVanGeffen already pointed out: many cheats are also useful debugging tools. Stuff like "ignore research status" and "win scenario" spring to mind.

IntelOrca:
Perhaps tile inspector should go under the map icon then.

Gymnasiast:
I don't think that would be a good idea. It's not something you want to be available by default, as it can break your game.

IntelOrca:
I didn't imply it would be available by default, you can hide individual menu items.

AaronVanGeffen:
I can see the logic in filing the options under categorically related toolbar items, but I really think we shouldn't.

In its current implementation, the configurable options affect just one button on the toolbar. This makes it easy to see what's added or removed. This does not apply to individual (sub)menu items — thereby hindering discoverability.

Therefore, personally, I'd much rather see the options affect just one menu, instead of hiding individual menu items.

AaronVanGeffen:
Has since been resolved.

PR #8892, opened 2019-03-16
Move common actions from debug menu into cheats menu

AaronVanGeffen:
This aims to make it less necessary for users to enable debugging tools.

Screenshot_20190316_163937

Implements #8688. Feedback is welcome.


Gymnasiast:
Could you also update the changelog?

Gymnasiast:
I think this implementation is easiest for most people, but we'd like to have @IntelOrca 's approval first.

PR #8895, opened 2019-03-16
Desync fix.

ZehMatt:
Those fields are only required for rendering/invalidation and shall be ignored.


Gymnasiast:
Why u no changelog? 🙁

PR #8901, opened 2019-03-16
Fix #8900: Implement GuestSetFlagsAction.

### LINKED ISSUE #8900: Peep tracking is not synchronized.
Enabling peep tracking on a multiplayer session will cause a desync as the flags are set directly without the server knowing about it.

ZehMatt:
Also while updating the twitch part I noticed that it also directly modifies the name of the peeps, it should probably use the game action to modify the name also.


src/openrct2-ui/windows/Guest.cpp
```diff
@@ -636,8 +637,13 @@ void window_guest_overview_mouse_up(rct_window* w, rct_widgetindex widgetIndex)
             window_scroll_to_viewport(w);
             break;
         case WIDX_TRACK:
-            get_sprite(w->number)->peep.peep_flags ^= PEEP_FLAGS_TRACKING;
-            break;
+        {
+            uint32_t flags = get_sprite(w->number)->peep.peep_flags ^ PEEP_FLAGS_TRACKING;
```
duncanspumpkin: `peep` is already a variable in this function no need for the `get_sprite`.

duncanspumpkin:
Just a thought but shouldn't tracking be tied to individual clients? Not everyone wants to track the same peeps.

ZehMatt:
I talked to @Broxzier about this, the main argument was that when people leave and join again it should still have it, but yeah to me it made more sense to have it per client, so I proposed to fix the desync for now and then we can discuss later what the ideal way would be.

Gymnasiast:
Please update the changelog and bump the network version.

distribution/changelog.txt
```diff
@@ -3,7 +3,7 @@
 - Fix: [#5579] Network desync immediately after connecting.
 - Fix: [#7884] Unfinished preserved rides can be demolished with quick demolish.
 - Fix: [#8873] Potential crash when placing footpaths.
-- Change: [#8688] Move common actions from debug menu into cheats menu.
+- Fix: [#8900] Synchronize peep tracking to prevent desyncs.
```
Gymnasiast: You unintentionally removed another entry.

ZehMatt: Hm thats strange, I recall keeping it when the rebase had conflicts, ill fix it.

Gymnasiast: ... now you moved it up the list :/
Can you just keep it where it was, i.e. after the fixes?

ZehMatt: It was ordered like that from the previous release, but okay.

Gymnasiast: You're right, that was a mistake on my part.

PR #8914, opened 2019-03-17
Add landset/buyrights action

duncanspumpkin:
I've reworked how these two commands work quite a considerable amount and this will need testing.

The new way these two work is that there is one action for normal play and a different action for editor/sandbox mode. That was how the old game commands were setup but due to the way they worked it was possible to use editor commands in normal play.


duncanspumpkin:
Note to self. Should anything be accessing the master set ownership command. The checks should be performed on all editor commands and maybe they should access the correct individual one.

src/openrct2/Editor.cpp
```diff
@@ -193,7 +194,10 @@ namespace Editor
     {
         int32_t mapSize = gMapSize;
 
-        game_do_command(64, 1, 64, 2, GAME_COMMAND_SET_LAND_OWNERSHIP, (mapSize - 3) * 32, (mapSize - 3) * 32);
+        auto landSetRightsAction = LandBuyRightsAction(
+            { 64, 64, (mapSize - 3) * 32, (mapSize - 3) * 32 }, LandBuyRightSetting::BuyLand);
+        landSetRightsAction.SetFlags(GAME_COMMAND_FLAG_5);
+        GameActions::Execute(&landSetRightsAction);
```
duncanspumpkin: This is likely wrong. Should not just buy land should set it to bought

duncanspumpkin:
Note to self still doesn't work

src/openrct2/actions/LandSetRightsAction.hpp
```diff
@@ -0,0 +1,227 @@
...
+                {
+                    if (tileElement->GetType() == TILE_ELEMENT_TYPE_ENTRANCE
+                        && tileElement->AsEntrance()->GetEntranceType() == ENTRANCE_TYPE_PARK_ENTRANCE)
+                    {
+                        // Do not allow ownership of park entrance.
+                        if (_ownership == OWNERSHIP_OWNED || _ownership == OWNERSHIP_AVAILABLE)
+                            return res;
+                        // Allow construction rights available / for sale on park entrances on surface.
+                        // There is no need to check the height if _ownership is 0 (unowned and no rights available).
+                        if ((_ownership == OWNERSHIP_CONSTRUCTION_RIGHTS_OWNED
```
ZehMatt: I think splitting this into multiple statements would be easier to read.

src/openrct2/actions/LandBuyRightsAction.hpp
```diff
@@ -0,0 +1,185 @@
...
+                    surfaceElement->SetOwnership(surfaceElement->GetOwnership() | OWNERSHIP_CONSTRUCTION_RIGHTS_OWNED);
+                    uint16_t baseHeight = surfaceElement->base_height * 8;
+                    map_invalidate_tile(loc.x, loc.y, baseHeight, baseHeight + 16);
+                }
+                res->Cost = gConstructionRightsPrice;
+                return res;
+
+            default:
+                log_warning("Tried calling buy land rights with an incorrect setting. setting = %u", _setting);
+                assert(false);
```
ZehMatt: Guard::Assert

src/openrct2/actions/LandBuyRightsAction.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+    GameActionResult::Ptr map_buy_land_rights_for_tile(const CoordsXY loc, bool isExecuting) const
+    {
+        if (_setting >= static_cast<uint8_t>(LandBuyRightSetting::Count))
+        {
+            log_warning("Tried calling buy land rights with an incorrect setting. setting = %u", _setting);
+            return MakeResult(GA_ERROR::INVALID_PARAMETERS, _ErrorTitles[0], STR_NONE);
+        }
+
+        SurfaceElement* surfaceElement = map_get_surface_element_at(loc)->AsSurface();
```
ZehMatt: If map_get_surface_element_at returns nullptr this will try to dereference null when calling AsSurface()

src/openrct2/world/Map.cpp
```diff
@@ -1952,7 +1907,13 @@ void map_remove_out_of_range_elements()
         {
             if (x == 0 || y == 0 || x >= mapMaxXY || y >= mapMaxXY)
             {
-                map_buy_land_rights(x, y, x, y, BUY_LAND_RIGHTS_FLAG_UNOWN_TILE, GAME_COMMAND_FLAG_APPLY);
+                // Note this purposely does not use LandSetRightsAction as X Y coordinates are outside of normal range.
```
duncanspumpkin: This still does not work

src/openrct2/actions/LandBuyRightsAction.hpp
```diff
@@ -0,0 +1,184 @@
...
+        if (isExecuting)
+        {
+            map_count_remaining_land_rights();
+        }
+        return res;
+    }
+
+    GameActionResult::Ptr map_buy_land_rights_for_tile(const CoordsXY loc, bool isExecuting) const
+    {
+        if (_setting >= static_cast<uint8_t>(LandBuyRightSetting::Count))
```
ZehMatt: I think this whole block can be moved to Query

src/openrct2/actions/LandSetRightsAction.hpp
```diff
@@ -0,0 +1,231 @@
...
+                                }),
+                            gPeepSpawns.end());
+                    }
+                    surfaceElement->SetOwnership(_ownership);
+                    update_park_fences_around_tile(loc);
+                    gMapLandRightsUpdateSuccess = true;
+                }
+                return res;
+            }
+            default:
```
ZehMatt: Should be done in Query

src/openrct2/actions/LandSetRightsAction.hpp
```diff
@@ -0,0 +1,231 @@
...
+    SetOwnershipWithChecks,
+    Count
+};
+
+DEFINE_GAME_ACTION(LandSetRightsAction, GAME_COMMAND_SET_LAND_OWNERSHIP, GameActionResult)
+{
+private:
+    MapRange _range;
+    uint8_t _setting = static_cast<uint8_t>(LandSetRightSetting::Count);
+    uint8_t _ownership;
```
ZehMatt: Should have a default invalid value and tested in Query

src/openrct2/actions/GameActionRegistration.cpp
```diff
@@ -138,6 +140,7 @@ namespace GameActions
         Register<LandLowerAction>();
         Register<LandRaiseAction>();
         Register<LandSetHeightAction>();
+        Register<LandSetRightsAction>();
         Register<LandSmoothAction>();
```
ZehMatt: Missing LandBuyRightsAction registration

src/openrct2/actions/LandSetRightsAction.hpp
```diff
@@ -0,0 +1,231 @@
...
+            {
+                if (_ownership == surfaceElement->GetOwnership())
+                {
+                    return res;
+                }
+
+                TileElement* tileElement = map_get_first_element_at(loc.x / 32, loc.y / 32);
+                do
+                {
+                    if (tileElement->GetType() != TILE_ELEMENT_TYPE_ENTRANCE)
```
ZehMatt: Potential null dereference.

src/openrct2/actions/GameActionRegistration.cpp
duncanspumpkin: I think its when I rebase that it sometimes deletes my register as I've tested these a few times. Thanks for noticing.

src/openrct2/actions/LandSetRightsAction.hpp
duncanspumpkin: Its default `0` from the two main constructors. I don't think it makes sense to default invalid this one as there isn't really an invalid value.

src/openrct2/actions/LandBuyRightsAction.hpp
duncanspumpkin: It already is in the Query. Its also in the execute though. I don't think it would read better if it had an additional if and there isn't much of a performance issue here either.

src/openrct2/actions/LandSetRightsAction.hpp
duncanspumpkin: This is the Query and Execute. As above.

src/openrct2/actions/LandSetRightsAction.hpp
ZehMatt: When it comes from the network it will use the default constructor where this value is uninitialized, this is just to catch serialization issues.

src/openrct2/actions/LandSetRightsAction.hpp
ZehMatt: Execute is never called without Query first unless one would specifically invoke Execute without the game action framework so its just doing it twice at the moment.

src/openrct2/actions/LandBuyRightsAction.hpp
```diff
@@ -0,0 +1,184 @@
...
+DEFINE_GAME_ACTION(LandBuyRightsAction, GAME_COMMAND_BUY_LAND_RIGHTS, GameActionResult)
+{
+private:
+    MapRange _range;
+    uint8_t _setting = static_cast<uint8_t>(LandBuyRightSetting::Count);
+
+    constexpr static rct_string_id _ErrorTitles[] = { STR_CANT_BUY_LAND, STR_CANT_BUY_CONSTRUCTION_RIGHTS_HERE };
+
+public:
+    LandBuyRightsAction()
```
ZehMatt:
```suggestion
    LandBuyRightsAction() = default;
```

src/openrct2/actions/LandSetRightsAction.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+DEFINE_GAME_ACTION(LandSetRightsAction, GAME_COMMAND_SET_LAND_OWNERSHIP, GameActionResult)
+{
+private:
+    MapRange _range;
+    uint8_t _setting = static_cast<uint8_t>(LandSetRightSetting::Count);
+    uint8_t _ownership;
+
+public:
+    LandSetRightsAction()
```
ZehMatt:
```suggestion
    LandSetRightsAction() = default;
```

src/openrct2/actions/LandSetRightsAction.hpp
duncanspumpkin: I don't really see the point in adding extra logic so that it can only execute during Query.

src/openrct2/actions/LandSetRightsAction.hpp
duncanspumpkin: Okay I will default construct it to `0`. I can't do a check on that though.

ZehMatt: Thats why its best to have it to some invalid value like 0xFF which can be checked in Query, but its okay for now, I'll run the fuzzer soon and see what kind of issues popup when network receives garbage/missing data.

src/openrct2/actions/LandSetRightsAction.hpp
ZehMatt: What I meant is it has to be done only in Query as long no one directly uses Execute, but its fine this way as well.

src/openrct2/actions/LandSetRightsAction.hpp
duncanspumpkin: I kind of want to just get rid of the whole `SetOwnershipWithChecks` which would make this field pointless as its too powerful a command. Its a job for another day though. 0xFF would also be a valid value thats the problem. Its a collection of flags and it will do sensible things with any value.

ZehMatt: Alrite, not gonna block this PR any longer and my lazyness already stalled it for too long.

PR #8919, opened 2019-03-18
Console command to set price of all rides of type

duncanspumpkin:
RCT and friends suggested it would be useful to be able to set all rides of a certain type to a single price. 

This command can be used in 3 different ways:
single ride:
rides set price `<rideIndex> <price>`

All rides:
rides set price all `<price>`

All rides of type:
rides set price all `<type> <price>`


Gymnasiast:
Nice idea.
Needs to be put into the changelog.

duncanspumpkin:
The empty box is the game action. It will be fixed as the game actions are improved with better strings.

Issue #7296, opened 2018-03-17
Feature request: shortcut key for scenery picker


Umdlye:
This would be especially useful if it could be used without having the scenery window open, considering object rotating can already be done with shortcut keys.

This would allow builders to make an object palette in the corner of the map, look at it in an extra viewport and easily pick objects from it.

xp4xbox:
Il take a look at this.

PR #8926, opened 2019-03-20
Add scenery picker shortcut

benhhopkins:
Allows players to set a shortcut key for the scenery picker as requested in #7296

Opens the scenery builder first if not open already, then checks to see if the picker is already selected before pressing the button (so it doesn't toggle).

Follows the formatting of the similar shortcut functions for opening the scenery builder and rotation, including nested-ifs. Let me know if that isn't desired.


src/openrct2-ui/input/KeyboardShortcut.cpp
```diff
@@ -767,6 +768,38 @@ static void shortcut_advance_to_next_tick()
     gDoSingleUpdate = true;
 }
 
+static void shortcut_open_scenery_picker()
+{
+    if (gScreenFlags & SCREEN_FLAGS_TITLE_DEMO)
```
Gymnasiast:
```suggestion
    if ((gScreenFlags & (SCREEN_FLAGS_TITLE_DEMO | SCREEN_FLAGS_TRACK_DESIGNER | SCREEN_FLAGS_TRACK_MANAGER) || (gScreenFlags & SCREEN_FLAGS_SCENARIO_EDITOR && gS6Info.editor_step != EDITOR_STEP_LANDSCAPE_EDITOR))
```

That get rid if the nested if. Make sure you use clang-format to autoformat.

benhhopkins:
Thanks, I think this looks better.

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -579,6 +580,7 @@ static void window_scenery_mouseup(rct_window* w, rct_widgetindex widgetIndex)
             gWindowSceneryPaintEnabled = 0;
             gWindowSceneryClusterEnabled = 0;
             gWindowSceneryEyedropperEnabled = !gWindowSceneryEyedropperEnabled;
+            scenery_remove_ghost_tool_placement();
```
Gymnasiast: Why did you make this change? (What does it do if it's left out?)

benhhopkins: Before I added that change, I found the preview ghost of the scenery object currently being placed would stay on the map when switching to the scenery picker via hotkey. I suppose it wasn't needed before since you could only get the picker when mousing over the scenery window, so the ghost was already removed.

Gymnasiast: Ah, that sounds logical.

PR #8928, opened 2019-03-20
Make cheats menu partially accessible in scenario editor

### LINKED ISSUE #8927: Tile inspector no longer available in Scenario Editor
**OS:** [Windows 10]
**Version:** [0.2.2]
**Commit/Build:** [e64aafa]

I am not sure if tile inspector was intended to be used in scenario editor, but i used it there to remove park fences, or make the park entrance invisible, etc.
Since it was moved to the cheats menu, it is no longer available in scenario editor. Should some cheats maybe be available in scenario editor? I think disable clearance checks and remove support limits (to partly bury trees) could also be helpful there.

AaronVanGeffen:
This makes the cheats menu partially accessible in the scenario editor. Irrelevant items have been disabled.

Screenshot_20190320_200259

As the majority of the items are greyed out, it might be better to outright hide them instead. However, that would mean their dropdown indices change as well. I'm afraid that would make things a bit messier behind the scenes, so I've kept it like this, for now.

Fixes #8927.

### DIFF
--- a/src/openrct2-ui/windows/TopToolbar.cpp
+++ b/src/openrct2-ui/windows/TopToolbar.cpp
@@ -712,7 +712,7 @@ static void window_top_toolbar_invalidate(rct_window* w)
         window_top_toolbar_widgets[WIDX_CHAT].type = WWT_EMPTY;
     }
 
-    if (gScreenFlags & (SCREEN_FLAGS_SCENARIO_EDITOR | SCREEN_FLAGS_TRACK_DESIGNER | SCREEN_FLAGS_TRACK_MANAGER))
+    if (gScreenFlags & SCREEN_FLAGS_EDITOR)
     {
         window_top_toolbar_widgets[WIDX_PAUSE].type = WWT_EMPTY;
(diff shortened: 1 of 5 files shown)


src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -3268,6 +3267,15 @@ static void top_toolbar_init_cheats_menu(rct_window* w, rct_widget* widget)
         dropdown_set_disabled(DDIDX_INVENTIONS_LIST, true);
     }
 
+    if (gScreenFlags & (SCREEN_FLAGS_SCENARIO_EDITOR | SCREEN_FLAGS_TRACK_DESIGNER | SCREEN_FLAGS_TRACK_MANAGER))
```
Gymnasiast:
```suggestion
    if (gScreenFlags & SCREEN_FLAGS_EDITOR)
```

Krutonium:
I prefer having them disabled, it signals that they are disabled on purpose, and it's not a bug better imo.

PR #8930, opened 2019-03-20
Remove gTrackGroundFlags by returning it within GameActionResult

Nazey:
Right now, res on `RideConstruction.cpp#L1817` does not have member GroundFlagCount, even though the new subclass `TrackPlaceActionResult` includes it. Not building because of this (obviously)


src/openrct2/actions/TrackPlaceAction.hpp
```diff
@@ -68,31 +92,31 @@ DEFINE_GAME_ACTION(TrackPlaceAction, GAME_COMMAND_PLACE_TRACK, GameActionResult)
...
         {
             log_warning("Invalid direction for track placement, direction = %d", _origin.direction);
-            return std::make_unique<GameActionResult>(
-                GA_ERROR::INVALID_PARAMETERS, STR_RIDE_CONSTRUCTION_CANT_CONSTRUCT_THIS_HERE, STR_NONE);
+            return std::make_unique<TrackPlaceActionResult>(
+                GA_ERROR::INVALID_PARAMETERS, STR_NONE);
         }
 
-        auto res = std::make_unique<GameActionResult>();
+        auto res = std::make_unique<TrackPlaceActionResult>();
```
duncanspumpkin: All of the references to `gTrackGroundFlags` in this function and the Execute should be changed to our new variable `res->GroundFlagCount`

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1814,7 +1814,7 @@ static void window_ride_construction_construct(rct_window* w)
         _currentTrackSelectionFlags |= TRACK_SELECTION_FLAG_TRACK_PLACE_ACTION_QUEUED;
     }
 
-    if (gTrackGroundFlags & TRACK_ELEMENT_LOCATION_IS_UNDERGROUND)
+    if (res->GroundFlagCount & TRACK_ELEMENT_LOCATION_IS_UNDERGROUND)
```
duncanspumpkin: Ah sorry forgot this wasn't inside the callback. You will have to do a cast to the correct type. `dynamic_cast<TrackPlaceActionResult*>(res)->GroundFlagCount` will compile.

data/language/en-GB.txt
```diff
@@ -3585,7 +3585,7 @@ STR_6134    :Clear Scenery
 STR_6135    :Client sent invalid request
 STR_6136    :Server sent invalid request
 STR_6137    :OpenRCT2, a free and open source clone of Roller Coaster Tycoon 2.
-STR_6138    :OpenRCT2 is the work of many authors, a full list can be found in “contributors.md”. For more information, visit http://github.com/OpenRCT2/OpenRCT2
+STR_6138    :OpenRCT2 is the work of many authors, a full list can be found in “contributors.md”. For more information, visit http://github.com/OpenRCT2/OpenRCT2.
```
duncanspumpkin: Any reason for this change? This isn't really part of this PR. If you think it should be changed it would be best to address in a different PR.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1723,10 +1723,10 @@ static void RideConstructPlacedForwardGameActionCallback(const GameAction* ga, c
     CloseConstructWindowOnCompletion(ride);
 }
 
-static void RideConstructPlacedBackwardGameActionCallback(const GameAction* ga, const GameActionResult* result)
+static void RideConstructPlacedBackwardGameActionCallback(const GameAction* ga, const TrackPlaceActionResult* result)
 {
     if (result->Error != GA_ERROR::OK)
-    {
+    {
```
duncanspumpkin: Try to avoid adding whitespace. `clang_format` will take care of this.

duncanspumpkin:
I see that you have commited straight to develop. There is nothing wrong with that but you will tend to find it is easier to work with git if you make a feature branch off of develop and work on that. It makes it easier to rebase and fix issues. If you want to learn a bit more about it read this article. https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow

When we commit this to master we will likely squash all of your 15 commits into a single one as then the history is easier to read, bisecting is easy and if required reverting is easier. Using a feature branch its easier to stop having loads of merge from develop commits.

duncanspumpkin:
Oh and one final note add yourselves to `https://github.com/OpenRCT2/OpenRCT2/blob/develop/contributors.md` in the bug fixes section.

src/openrct2-ui/windows/RideConstruction.cpp
duncanspumpkin: Ah no that wont work as this is a smart pointer.

duncanspumpkin: `dynamic_cast<TrackPlaceActionResult*>(res.get())->GroundFlagCount` now that will compile.

data/language/en-GB.txt
Nazey: What I didn’t realize was that I could use the feature branch instead of directly merging. For our class, we essentially had to put our names in that string, so I just removed it. If there is a way to remove this part, I would be all for it.

duncanspumpkin:
There is a reference to `gTrackGroundFlags` in `_legacy.cpp` for maze elements. Mazes can never be placed underground and `gTrackGroundFalgs` will always be 0 so you can safely remove the reference and choose the correct static value.

data/language/en-GB.txt
duncanspumpkin: Just correct it in another commit. When we squash merge this it will be ignored.

Nazey: Ok, sweet.

Nazey:
Ok, everything builds correctly, but the game still has trouble switching into underground mode. It will eventually switch, but it is finicky. Like, in some cases it works flawlessly, but others fails. Take this one for example. Shouldn't it change to underground mode?
Screen Shot 2019-03-20 at 7 19 54 PM

duncanspumpkin:
So close.

Do a search for references to the variable. I know of one last place you still need to adjust. That should fix it.

Nazey:
I swear I have looked up and down for it, but I cannot find any other references to `gTrackGroundFlags'. And I couldn't see anything with `GroundFlag` that needed to be changed either. Am I missing something in my project again? I know I am a couple 100 commits behind develop, but it shouldn't really affect the functionality of underground mode.

duncanspumpkin:
`L118` of `src/openrct2/windows/_legacy.cpp`. You could also just delete the definition of gTrackGroundFlags and it would quickly show the problem when you compile.

Nazey:
I didn't even realize that I hadn't added `_legacy.cpp` to my commits. That was already changed a long time ago, but it still refuses to go into underground mode.

duncanspumpkin:
Well the mistake is still in _legacy.cpp. have a think about how this should work.

Nazey:
Got it. Had an extra 0 that needed to be dynamically casted. That is a wicked line right there just for a variable lol.

duncanspumpkin:
Okay this is almost finished. The last task is to delete the definition and declaration of the global variable. Then finally add yourself to the contributors file.

Nazey:
Yeah I keep not committing the changes that I make. Need to work on that.

duncanspumpkin:
`git commit -a` will select all changes then you won't forget.

src/openrct2/windows/_legacy.cpp
```diff
@@ -134,7 +134,7 @@ money32 place_provisional_track_piece(
         _unkF440C5.z = z;
         _unkF440C5.direction = trackDirection;
         _currentTrackSelectionFlags |= TRACK_SELECTION_FLAG_TRACK;
-        viewport_set_visibility((gTrackGroundFlags & TRACK_ELEMENT_LOCATION_IS_UNDERGROUND) ? 1 : 3);
+        viewport_set_visibility((0 & TRACK_ELEMENT_LOCATION_IS_UNDERGROUND) ? 1 : 3);
```
AaronVanGeffen: `((0 & TRACK_ELEMENT_LOCATION_IS_UNDERGROUND) ? 1 : 3)` will always be `3`, so this not seem correct to me.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1790,7 +1790,7 @@ static void window_ride_construction_construct(rct_window* w)
     auto trackPlaceAction = TrackPlaceAction(
         rideIndex, trackType, { x, y, z, static_cast<uint8_t>(trackDirection) }, (properties)&0xFF, (properties >> 8) & 0x0F,
         (properties >> 12) & 0x0F, liftHillAndAlternativeState);
-
+
```
AaronVanGeffen:
```suggestion

```

src/openrct2/windows/_legacy.cpp
Nazey: Isn't this line for a maze, and mazes cannot ever be put underground?

src/openrct2-ui/windows/RideConstruction.cpp
AaronVanGeffen:
```suggestion
    {
```

src/openrct2/actions/TrackPlaceAction.hpp
```diff
@@ -19,7 +19,31 @@
...
+    TrackPlaceActionResult(GA_ERROR error, rct_string_id message)   
+    : GameActionResult(error, STR_RIDE_CONSTRUCTION_CANT_CONSTRUCT_THIS_HERE, message)
+    {
+    }
+     TrackPlaceActionResult(GA_ERROR error, rct_string_id message, uint8_t* args)  
+    : GameActionResult(error, STR_RIDE_CONSTRUCTION_CANT_CONSTRUCT_THIS_HERE, message, args)
+    {
+    }
+
+    uint8_t GroundFlag{ 0 };
```
ZehMatt: I think it would be better to call this GroundFlags

src/openrct2/windows/_legacy.cpp
ZehMatt: Yeah this can be always 3 as the maze does not/did not affect gTrackGroundFlags at all.

src/openrct2/actions/TrackPlaceAction.hpp
Nazey: Makes more sense for consistency I guess.

Nazey:
Anything else I should be looking for?

janisozaur:
You still need to apply the formatting and some commits have overlong subject lines, see https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

AaronVanGeffen:
I think we're going to squash merge anyway due to the merge commits, so no need to adjust any commits.

Nazey:
Yeah, from now on I'm gonna just branch off of develop and commit to that instead of develop like @duncanspumpkin said. It looks way cleaner.

PR #8960, opened 2019-03-25
Water level is raised abnormally

Nazey:
`WaterRaiseAction.hpp` was missing an if statement and therefore would place water tiles on inclines where it shouldn't have.


Nazey:
Don't know why all of the past commits are included in this one, I made a new branch but it didn't seem like it removed the past ones.

fromthere:
Making new branch doesn't remove old commits. you should rebase with upstream(in this case, this repo) and force-push commits.

duncanspumpkin:
Yeah the problem is you need to do a `git reset --hard origin/develop` to make it all back to normal.

PR #8964, opened 2019-03-25
Fix #8882: Submarine Ride does not count as indoors

### LINKED ISSUE #8882: Submarine Ride does not count as indoors
**OS:** Windows 10
**Version:** 0.2.2
**Commit/Build:** 57b8fba32

The Submarine Ride, despite having enclosed vehicles, does not count as an indoor ride; consequently, guests will refuse to ride it when it is raining.

The attached save was made during rain, so reproducing this should be as simple as checking what guests are thinking about "Space Subs".

**Steps to reproduce:**
1.  Open any scenario that starts you with the Submarine Ride
2. Build one of any length and open it
3. Wait for rainfall or use the cheats menu to force it to rain, and watch as guests refuse to ride the fully-enclosed subs.

**Dump file**

**Screenshots / Video:**
Leafy Lake 2019-03-15 18-44-41

**Save game:**
[SubBug.zip](https://github.com/OpenRCT2/OpenRCT2/files/2973160/SubBug.zip)

Gymnasiast:


src/openrct2/ride/RideRatings.cpp
```diff
@@ -4127,7 +4127,8 @@ static void ride_ratings_calculate_submarine_ride(Ride* ride)
     ride->window_invalidate_flags |= RIDE_INVALIDATE_RIDE_INCOME;
 
     ride->inversions &= 0x1F;
-    ride->inversions |= 0 << 5;
+    // Originally, this was always to zero, even though the default vehicle is completely enclosed.
+    ride->inversions |= get_num_of_sheltered_eighths(ride) << 5;
```
duncanspumpkin: Should really use `ride->sheltered_eighths`

duncanspumpkin: Ahh I see none of rideRatings uses `ride->sheltered_eighths` never mind. Will make a refactor issue.

PR #8967, opened 2019-03-25
Water level is raised abnormally

Nazey:
A missing if statement caused the water levels to all rise proportionally on the first "rise" action.


duncanspumpkin:
Good. You will need to increment network version in network.cpp. also in your commit message you should refer to the issue you are fixing so you would write `Fix #xxx: description` that way everything links nicely in github. I'll fix it when merging for this one though.

Nazey:
Oh boy, I accidentally pushed everything to my waterbug branch again

Issue #8942, opened 2019-03-23
Water level is raised abnormally (in the first)


telk5093:
**OS:** Windows 10
**Version:** 0.2.2
**Commit/Build:** 6d99426

Water is raised on the every land level in the same time when I first raise them.

**Steps to reproduce:**
1. Raise the water

**Screenshots / Video:**
![녹화_2019_03_23_18_09_31_300](https://user-images.githubusercontent.com/10726524/54864187-e1c92600-4d96-11e9-83e7-e09aaaedbcae.gif)

Nazey:
Yea it is definitely something new within version 0.2.2, because 0.2.1 did not have this issue at all.

Nazey:
I'm gonna try and give this one a go

duncanspumpkin:
@nazey your welcome to have a shot but i do know where the issue is. I changed this code to game action two weeks ago. The mistake will be in the lower water game action.

duncanspumpkin:
I've accidently removed `if(surfaceElement->base_height > maxHeight) continue;` from the water raise game action loop.

Nazey:
Dang you work fast. I was messing around with `WaterRaiseAction.hpp` yesterday and thought I was getting close, but your snippet fixed it.

duncanspumpkin:
Well i only touched this code last week so had a big headstart

Nazey:
@duncanspumpkin I've tried cleaning my repo and performing a hard reset on the repo for like the past hour, but the old commits still show up when I try to pull request from the new branch. Is it because I am on my own forked repo? Or is there something I'm still missing.

Broxzier:
When you've already pushed your branch before, you need to force push to overwrite it on the remote repo (GitHub). If your branch is named `bugfix/water-level` for instance, you'd run:

```
git checkout bugfix/water-level # your branch
git fetch upstream
git reset --hard upstream/develop # resets your branch to be the same as upstream/develop
# make your changes and commit them
git push --force origin bugfix/water-level # push your branch to your fork
```

Where `origin` points at your fork and `upstream` points at this repository.

If you already have made commits and just want to apply it on a newer version of the develop branch, you should rebase instead.

duncanspumpkin:
In addition to the above i expect the issue is the commits you are trying to get rid of are still in your feature branch. You could do a `git rebase -i upstream/develop` and delete all of the bad commits then force push to github or just delete the feature branch and recreate it from upstream/develop.

[...]

PR #8971, opened 2019-03-26
Fix bug #8942 by adding missing if statement

Nazey:
A missing if statement caused the water levels to all rise proportionally on the first "rise" action. Should be fine now

### DIFF
--- a/src/openrct2/actions/WaterRaiseAction.hpp
+++ b/src/openrct2/actions/WaterRaiseAction.hpp
@@ -87,7 +87,7 @@ private:
                 SurfaceElement* surfaceElement = tileElement->AsSurface();
                 uint8_t height = surfaceElement->GetWaterHeight();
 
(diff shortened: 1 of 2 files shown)


src/openrct2/actions/WaterRaiseAction.hpp
```diff
@@ -86,6 +86,10 @@ DEFINE_GAME_ACTION(WaterRaiseAction, GAME_COMMAND_RAISE_WATER, GameActionResult)
 
                 SurfaceElement* surfaceElement = tileElement->AsSurface();
                 uint8_t height = surfaceElement->GetWaterHeight();
+
+                if(surfaceElement->base_height > maxHeight)
```
AaronVanGeffen:
```suggestion
                if (surfaceElement->base_height > maxHeight)
```

PR #8978, opened 2019-03-27
Close #8965: Refactor ride->inversions

### LINKED ISSUE #8965: Refactor Ride.inversions
Ride.inversions has three different types of data that is stuffed into it. ~~We will ignore `holes` for now as golf rides are super special.~~
```
    union
    {
        uint8_t inversions; // (???X XXXX)
        uint8_t holes;      // (???X XXXX)
        // This is a very rough approximation of how much of the ride is undercover.
        // It reaches the maximum value of 7 at about 50% undercover and doesn't increase beyond that.
        uint8_t sheltered_eighths; // (XXX?-????)
    };
```
~~There should be an getter and setter for inversions and sheltered_eighths (and probably holes).~~

These should be split into 3 different fields that are set correctly from s4/s6 importer/exporter.

Gymnasiast:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -1781,14 +1781,21 @@ static void vehicle_update_measurements(rct_vehicle* vehicle)
...
         {
-            uint8_t inversions = ride->inversions & 0x1F;
-            if (inversions != 0x1F)
-                inversions++;
-
-            ride->inversions &= ~0x1F;
-            ride->inversions |= inversions;
+            if (track_flags & TRACK_ELEM_FLAG_IS_GOLF_HOLE)
+            {
+                if (ride->holes < 255)
```
duncanspumpkin: This brakes compatibility with vanilla. Stick to 0x1F

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1781,14 +1781,21 @@ static void vehicle_update_measurements(rct_vehicle* vehicle)
...
+                    ride->holes++;
+            }
+        }
+        else
+        {
+            if (track_flags & TRACK_ELEM_FLAG_NORMAL_TO_INVERSION)
+            {
+                if (ride->inversions < 255)
```
duncanspumpkin: Again brakes vanilla

src/openrct2/ride/Vehicle.cpp
Gymnasiast: How does it break compatibility? It gets &'ed with 0x1F when exporting.

Gymnasiast: How does it break compatibility? It gets truncated on export. I suppose I could use a `min(31, inversions)` on export rather than truncating.

IntelOrca: `min(31, inversions)` would be better.

Gymnasiast: Done.

IntelOrca:
Is there any reason for `inversions` and `holes` to stay as `uint8_t` now? We can probably make them 16 or 32-bit.

Gymnasiast:
@IntelOrca I don't think there is any specific reason not to... would you like me to change it in this PR?

IntelOrca:
@Gymnasiast yeah, I noticed it after I saw the uint8_t cast in the min function which you can now change.

src/openrct2/ride/RideRatings.cpp
```diff
@@ -1892,12 +1894,11 @@ static void ride_ratings_calculate_miniature_railway(Ride* ride)
     ride->upkeep_cost = ride_compute_upkeep(ride);
     ride->window_invalidate_flags |= RIDE_INVALIDATE_RIDE_INCOME;
 
-    int32_t edx = get_num_of_sheltered_eighths(ride);
-    if (((edx >> 8) & 0xFF) >= 4)
+    auto shelteredEighths = get_num_of_sheltered_eighths(ride);
+    if ((shelteredEighths.TrackShelteredEighths & 0xFF) >= 4)
```
duncanspumpkin: No need for the 0xff

src/openrct2/ride/RideRatings.cpp
```diff
@@ -1928,12 +1929,11 @@ static void ride_ratings_calculate_monorail(Ride* ride)
...
```
duncanspumpkin: As above

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: You have to change it here to 31 otherwise you modify state when saving. This will break multiplayer and also mean the save performs differently in vanilla and open

Gymnasiast: How will it break multiplayer, apart from needing a network version bump?

Really, I don't care at all that vanilla behaves differently for the few months we still export to it: it only happens with more than 31 inversions/holes, and even then it is handled gracefully: RCT2 wouldn't know the difference.

src/openrct2/ride/RideRatings.cpp
Gymnasiast: I was a bit loathe to remove it - why was it there in the first place?

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: The checksums wont match between clients and hosts as the files are sent as S6's.

duncanspumpkin: May as well keep it compatible for the next few months before increasing the limits.

src/openrct2/ride/RideRatings.cpp
duncanspumpkin: Because you had to mask off the upper 8bits. Now that the function returns a struct its not required.

duncanspumpkin:
Thanks. I know it was a bit of a pain and unlikely to cause issues.

Gymnasiast:
No problem, I didn't know it would cause issues with multiplayer. At least we can now very easily increase it, and it also eliminates another hardcoded limit.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -560,8 +560,11 @@ void S6Exporter::ExportRide(rct2_ride* dst, const Ride* src)
     dst->turn_count_default = src->turn_count_default;
     dst->turn_count_banked = src->turn_count_banked;
     dst->turn_count_sloped = src->turn_count_sloped;
-    // Includes holes and (for some strange reason?!) sheltered_eights
-    dst->inversions = src->inversions;
+    if (dst->type == RIDE_TYPE_MINI_GOLF)
+        dst->inversions = std::min((uint8_t)src->holes, (uint8_t)RCT12_MAX_GOLF_HOLES);
+    else
+        dst->inversions = std::min((uint8_t)src->inversions, (uint8_t)RCT12_MAX_INVERSIONS);
```
IntelOrca: This won't give you nice results when you have more than 255 inversions / holes. You should cast after the min function has applied, let min operate on the larger sized integers.

PR #8981, opened 2019-03-27
Fix #5905: Urban Park MGR has entrance/exit swapped

### LINKED ISSUE #5905: (RCT1) Urban Park Entrance Problem
There is a Problem in the Urban Park from Added Attractions. The entrance and exit are switched, so the queue line goes away from the exit, instead of to the entrance. 
This happens since Chris Sawyer made a mistake when he builds the park. (so also in RCT1). 
screenshot_12

I don't know if it would be possible to fix this, without changing the SC4 file. (Similar to the missing land purchases)

Gymnasiast:


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2940,6 +2943,50 @@ class S4Importer final : public IParkImporter
...
+
+            auto entranceElement = map_get_ride_exit_element_at(
+                entranceCoords.x * 32, entranceCoords.y * 32, entranceCoords.z, false);
+            entranceElement->SetEntranceType(ENTRANCE_TYPE_RIDE_ENTRANCE);
+            auto exitElement = map_get_ride_entrance_element_at(exitCoords.x * 32, exitCoords.y * 32, exitCoords.z, false);
+            exitElement->SetEntranceType(ENTRANCE_TYPE_RIDE_EXIT);
+
+            // Trigger footpath update
+            footpath_queue_chain_reset();
+            footpath_connect_edges(
```
duncanspumpkin: Just a thought but does this spend money? GAME_COMMAND_FLAG_5 is normally passed in a fixing function as that tells it not to spend money (or network it).

ZehMatt: It only uses the flags but its not really a game command, it should be fine.

Gymnasiast: Reconnecting footpath edges does not cost money.

> GAME_COMMAND_FLAG_5 is normally passed in a fixing function as that tells it not to spend money (or network it).

Shouldn't we rename that flag?

duncanspumpkin: I thought for a while that it should be renamed GAME_COMMAND_TRACK_PREVIEW. But now we use it for quite a few things not related to track previews. I guess its like GHOST in that it forces it to be client side only so that property of it should be documented but not part of the name. So maybe NO_COST would be appropriate but its a very oddly used command.

Issue #8219, opened 2018-11-03
Faulty folder recreation in "save"


Richard-L:
>Place an SV6 in a subfolder in the default "save" directory.
>Open the park and close OpenRCT2.
>Delete the subfolder and its contents. Open OpenRCT2 back up.
>An empty subfolder with the same name as before is created, as it looks for that folder in the "open park" dialogue.

It should instead notice the folder has been deleted, and go to the default "save".

Thanks

hokasha2016:
My team and i would like to attempt to fix this issue

bdupree5:
We will be submitting a pull request for this issue today.

PR #8982, opened 2019-03-27
Fix bug OpenRCT2#8219

hokasha2016:
This is another pull request, the previous one was closed. This fix is to prevent the game from automatically creating a new sub folder even though it was deleted. If the sub folder exists, then it will take you there when you try to load  game. If it was moved or deleted, then it takes you to the default path.

### DIFF
--- a/src/openrct2-ui/windows/LoadSave.cpp
+++ b/src/openrct2-ui/windows/LoadSave.cpp
@@ -240,16 +240,11 @@ static const char* getFilterPatternByType(const int32_t type, const bool isSave)
 static int32_t window_loadsave_get_dir(const int32_t type, char* path, size_t pathSize)
 {
     const char* last_save = getLastDirectoryByType(type);
-    if (last_save && platform_ensure_directory_exists(last_save))
+    // checks for path existance
+    if (platform_directory_exists(last_save))
         safe_strcpy(path, last_save, pathSize);
     else
         getInitialDirectoryByType(type, path, pathSize);
-
-    if (!platform_ensure_directory_exists(path))
-    {
-        log_error("Unable to create save directory.");
-        return 0;
-    }
 
     return 1;
 }
--- a/src/openrct2-ui/windows/LoadSave.cpp
+++ b/src/openrct2-ui/windows/LoadSave.cpp
@@ -242,6 +242,7 @@ static int32_t window_loadsave_get_dir(const int32_t type, char* path, size_t pa
     const char* last_save = getLastDirectoryByType(type);
     // checks for path existance
     if (platform_directory_exists(last_save))
+    //If the last saved path is found
         safe_strcpy(path, last_save, pathSize);
     else
         getInitialDirectoryByType(type, path, pathSize);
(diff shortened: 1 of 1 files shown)


ZehMatt:
I think the wrong issue was referenced, the issue referenced is the AVX2 detection.

Gymnasiast:
It's issue #8219, not #8947.

Also, your commit title is wrong, it should mentioned the issue number and the problem it solves (not the solution, the problem).

In this case, the commit title should be `Fix #8219: Faulty folder recreation in "save"`

hokasha2016:
please excuse me, i thought the issue was right. ive fixed the title and issue number

duncanspumpkin:
@hokasha2016 sorry we were talking about the commit message. You can adjust that by doing `git commit -amend -m"Fix #8219: Faulty folder recreation in save"

hokasha2016:
excuse my stupidity, im still trying to get the hang of it but they're all the same. @duncanspumpkin I tried what you told me and it didnt work.

duncanspumpkin:
Argh it doesn't matter too much we can squash it and fix the title.

src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -240,17 +240,12 @@ static const char* getFilterPatternByType(const int32_t type, const bool isSave)
 static int32_t window_loadsave_get_dir(const int32_t type, char* path, size_t pathSize)
 {
     const char* last_save = getLastDirectoryByType(type);
-    if (last_save && platform_ensure_directory_exists(last_save))
+    // checks for path existance.
```
Gymnasiast: This comment is of no use - `platform_directory_exists()` speaks for itself.

src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -240,17 +240,11 @@ static const char* getFilterPatternByType(const int32_t type, const bool isSave)
 static int32_t window_loadsave_get_dir(const int32_t type, char* path, size_t pathSize)
 {
     const char* last_save = getLastDirectoryByType(type);
-    if (last_save && platform_ensure_directory_exists(last_save))
+    if (platform_directory_exists(last_save))
```
Gymnasiast: last_save can still be nullptr.
```suggestion
    if (last_save != nullptr && platform_directory_exists(last_save))
```

hokasha2016: When platform_ensure_directory_exists runs, it creates a new empty folder even though the folder was deleted. What i did was that it checks for the existence of a last_save path and that will return a true or false. it will return a false if last_save is a nullptr or doesnt exist anymore, then it shows the user the default path which is
 C:\Documents\OpenRCT2\save

Gymnasiast: That's not what I was talking about. `getLastDirectoryByType()` can return nullptr. This needs to be checked, because `platform_directory_exists()` does not handle nullptr. (That's basic C stuff.)

hokasha2016:
@Gymnasiast You were correct.

Gymnasiast:
This would also have needed a changelog entry, but I will address that in a follow-up commit.

Issue #8480, opened 2018-12-18
BACKTRACE fbed11 Object::GetLegacyData -> track_design_mirror_scenery


IntelOrca:
<table>
    <tr>
        <td><img src='https://backtrace.io/images/icon.png' width=30 height=30 /></td>
        <td><a href='https://openrct2.sp.backtrace.io/dashboard/openrct2/project/openrct2/group/fbed11eadb9624c32efb13ce6399eba129a3d7a38ac7d5217cee0667ebfeace3?_t=W3sibmFtZSI6IndvcmtmbG93In1d'><b>New unique error: [openrct2]</b></a><br /><b>1</b> occurrence(s) <b></b> since (Tue Dec 18 2018 13:58:05)<br /></td>
    </tr>
</table>

<table>
<tr><td><strong>Application</strong></td><td>openrct2.exe</td></tr>

<tr><td><strong>Attributes</strong></td><td><strong>attachment_park.sv6:</strong> *: <em>1</em><br><strong>attachment_config.ini:</strong> C:\\Users\\a\\Documents\\OpenRCT2\\731e9d88-3149-404b-a91f-70a1fc49ace7.ini: <em>1</em><br><strong>attachment_park.sv6.gz:</strong> C:\\Users\\a\\Documents\\OpenRCT2\\731e9d88-3149-404b-a91f-70a1fc49ace7.sv6.gz: <em>1</em><br><strong>commit:</strong> 0060c88: <em>1</em><br></td></tr>
<tr><td><strong>Classifiers</strong></td><td><strong>unhandled-exception</strong> </td></tr>
<tr><td><strong>Frames</strong></td><td><code>Object::GetLegacyData -> track_design_mirror_scenery -> window_track_place_mouseup -> input_state_widget_pressed -> game_handle_input_mouse -> game_handle_input -> OpenRCT2::GameState::Update -> OpenRCT2::Context::Update -> OpenRCT2::Context::RunGameLoop -> OpenRCT2::Context::Launch -> OpenRCT2::Context::RunOpenRCT2 -> NormalisedMain -> LaunchOpenRCT2 -> wmain</code></td></tr>
</table>

janisozaur:
Still valid, see https://github.com/OpenRCT2/OpenRCT2/issues/8535

PR #8983, opened 2019-03-27
Refactor track

duncanspumpkin:
Moving around the scenery placement to make it a bit clearer.


src/openrct2/ride/TrackDesign.cpp
```diff
@@ -701,467 +701,506 @@ static void track_design_update_max_min_coordinates(int16_t x, int16_t y, int16_
     gTrackPreviewMax.z = std::max(gTrackPreviewMax.z, z);
 }
 
-/**
- *
- *  rct2: 0x006D0964
- */
-static int32_t track_design_place_scenery(
-    rct_td6_scenery_element* scenery_start, int32_t originX, int32_t originY, int32_t originZ)
+static bool TrackDesignPlaceSceneryTileDrawOutline(LocationXY8 tile)
```
Gymnasiast: Don't use Location8/16/32 for anything new. It's deprecated, because it can be either (big) Coords or TileCoords. Use an explicit struct type.

janisozaur: `[[deprecated]]`

src/openrct2/Game.cpp
```diff
@@ -446,7 +446,7 @@ int32_t game_do_command_p(
     {
         // Check funds
         int32_t insufficientFunds = 0;
-        if (gGameCommandNestLevel == 1 && !(flags & GAME_COMMAND_FLAG_2) && !(flags & GAME_COMMAND_FLAG_5) && cost != 0)
```
Gymnasiast: Are you really sure that GAME_COMMAND_FLAG_2 is no longer needed?

duncanspumpkin: It was used in one function as a no money flag. I changed that function to use the no money flag instead

duncanspumpkin:
Note to self: Look at #8480

PR #8984, opened 2019-03-27
Fix non-networked game actions not showing errors.

ZehMatt:
The player id should default to -1 on game actions which is currently considered an invalid player id, the server will assign the id when its networked back. So for errors to show up when the action won't make it to the server this condition has to be checked since currently it would test the default value 0 against player id 1.


ZehMatt:
Not sure I like it this way, I'll have another look.

PR #8991, opened 2019-03-29
Fix #8988: Improve lookups for codepoint offsets

### LINKED ISSUE #8988: Opening the “New Ride” window noticably slows down game
**OS:** Ubuntu 18.04
**Version:** 0.2.2+
**Commit/Build:** 731af7473202b931ce243321bc7c42f9800dafc9

Using a debug build, I get 37 FPS at 1920x1200 in a certain park. When opening the New Ride window on the roller coasters page, this drop down to 15 FPS. This is using software rendering mode with no hardware display.

ZehMatt:
This change keeps the FPS constant at 40 on debug builds at 1920x1080. Park I used was Crazy Castle with nothing built and 10 Guests.

### DIFF
--- a/src/openrct2/drawing/Font.cpp
+++ b/src/openrct2/drawing/Font.cpp
@@ -17,7 +17,7 @@
 #include "TTF.h"
 
 #include <iterator>
-#include <map>
+#include <unordered_map>
 
 static constexpr const int32_t SpriteFontLineHeight[FONT_SIZE_COUNT] = { 6, 10, 10 };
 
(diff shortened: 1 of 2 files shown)


src/openrct2/drawing/Font.cpp
```diff
@@ -197,12 +197,23 @@ static const std::map<char32_t, int32_t> codepointOffsetMap = {
...
  *  rct2: 0x006C19AC
  */
 void font_sprite_initialise_characters()
 {
+    // Compute min and max that helps avoiding lookups for no reason.
+    _smallestCodepointValue = std::numeric_limits<char32_t>::max();
+    for (const auto entry : codepointOffsetMap)
+    {
+        _smallestCodepointValue = std::min(_smallestCodepointValue, entry.first);
+        _biggestCodepointValue = std::max(_biggestCodepointValue, entry.first);
```
janisozaur: Isn't it just `std::minmax_element`?

janisozaur: Hmm, I guest not in `std::unordered_map`.

PR #9002, opened 2019-03-31
Remove duplicate buttons in Cheats window

Gymnasiast:


AaronVanGeffen:
Nice catch on the 'reset date' button — that was indeed duplicate functionality in the same window.

I don't think the same goes for the sandbox button, however. Its functionality is also available in the cheats menu, but not in some other place in the cheats window.

Gymnasiast:
@AaronVanGeffen It's the only option that is duplicated between the two, though -- there is no corresponding button for "disable clearance checks" or "disable support limits".

Issue #8110, opened 2018-10-18
Three different folder names are used for title sequences


Gymnasiast:
This is something that stood me out when writing a tutorial about the user folder. The precreated folder in the user's directory is named `sequence`. The default ones are in `title`, and `GetUserSequencesPath()` uses `title sequences`.

This makes no sense. It should use one folder name. I propose using `sequence`.

IntelOrca:
If `GetUserSequencesPath()` returns `title sequences` why is a folder called `sequence` precreated?

IntelOrca:
Probably due to C and C++ cross.

Gymnasiast:
I don't know, both functions follow C++ naming convention.

Anyway, the reason I wanted to opt for `sequence` is because of three reasons:
- It's precreated already, making it easier for the users
- Its naming is consistent with the rest, unlike `title sequences`.
- It's easier to understand than `title`.

bdupree5:
Hey guys i'm trying to work on this issue, what exactly is the problem? In the function GetUserSequencesPath() needs to call on the table? Or needs to return "sequence" instead of "title sequence"? I'm just a bit confused.

IntelOrca:
@bdupree5 I think for now just replace all variations with `sequence`.

bdupree5:
I will get on it thanks.

PR #9009, opened 2019-04-01
Fix bug OpenRCT2#8110

hokasha2016:
Replaced all variations of "title sequence" with "sequence". Only 1 variation was found.


src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -244,6 +244,7 @@ static int32_t window_loadsave_get_dir(const int32_t type, char* path, size_t pa
         safe_strcpy(path, last_save, pathSize);
     else
         getInitialDirectoryByType(type, path, pathSize);
+
```
Gymnasiast: Stray added line.

src/openrct2/title/TitleSequenceManager.cpp
```diff
@@ -285,7 +285,7 @@ namespace TitleSequenceManager
 
     static void GetUserSequencesPath(utf8* buffer, size_t bufferSize)
     {
-        platform_get_user_directory(buffer, "title sequences", bufferSize);
+        platform_get_user_directory(buffer, "sequence", bufferSize);
```
Gymnasiast: This function should not be needed any more, and code that calls this function should use `PlatformEnvironment::DirectoryNamesOpenRCT2` instead.

Gymnasiast:
Issue #8219 clearly states in the opening post that there are two directory names need changing: `title sequences` and `title`. You did not address the last one.

hokasha2016:
I do apologize, issue #8219 refers to my previous issue fix which was the folder save; this is issue #8110 .. i will clean the commits and find all the variations of "title sequences"

hokasha2016:
I have found 2 methods, both with different titles
```

**Comment by hokasha2016** [comment]
I have found 2 variations, one above the other in the same file:
```
    static void GetDataSequencesPath(utf8* buffer, size_t bufferSize)
    {
        platform_get_openrct_data_path(buffer, bufferSize);
        Path::Append(buffer, bufferSize, "title");
    }

    static void GetUserSequencesPath(utf8* buffer, size_t bufferSize)
    {
        platform_get_user_directory(buffer, "title sequences", bufferSize);
    }
```
Should i change them both to be "sequence"? to match the table created in PlatformEnviroment.cpp:
```
const char * PlatformEnvironment::DirectoryNamesOpenRCT2[] =
{
    "data",                 // DATA
    "landscape",            // LANDSCAPE
    "language",             // LANGUAGE
    "chatlogs",             // LOG_CHAT
    "serverlogs",           // LOG_SERVER
    "keys",                 // NETWORK_KEY
    "object",               // OBJECT
    "save",                 // SAVE
    "scenario",             // SCENARIO
    "screenshot",           // SCREENSHOT
    "sequence",             // SEQUENCE
    "shaders",              // SHADER
    "themes",               // THEME
    "track",                // TRACK
    "heightmap",            // HEIGHTMAP
    "replay",               // REPLAY
};
```

**Comment by Gymnasiast** [comment]
> I do apologize, issue #8219 refers to my previous issue fix which was the folder save; this is issue #8110 .. i will clean the commits and find all the variations of "title sequences"

I also mistakingly typed #8219 in my post here. I meant #8110, indeed.

> Should i change them both to be "sequence"? to match the table created in PlatformEnviroment.cpp:

You identified the right functions, but you should remove them entirely and replace their callers with calls to `GetDirectoryPath`. By searching for that function, you should be able to find out how it's used.

I will handle the necessary changes to the build systems.

**Comment by hokasha2016** [comment]
Great, I will work on that and preform another pull request for this issue when I have completed it. Thank you. I apologize for the confusion and such.

**Comment by hokasha2016** [comment]
So in ``` Context.cpp ``` we found the two methods being used above starting at line 1332-1356:
```
/**
 * This function is deprecated.
 * Use IPlatformEnvironment instead.
 */
void platform_get_user_directory(utf8* outPath, const utf8* subDirectory, size_t outSize)
{
    auto env = GetContext()->GetPlatformEnvironment();
    auto path = env->GetDirectoryPath(DIRBASE::USER);
    if (!String::IsNullOrEmpty(subDirectory))
    {
        path = Path::Combine(path, subDirectory);
    }
    String::Set(outPath, outSize, path.c_str());
}

/**
 * This function is deprecated.
 * Use IPlatformEnvironment instead.
 */
void platform_get_openrct_data_path(utf8* outPath, size_t outSize)
{
    auto env = GetContext()->GetPlatformEnvironment();
    auto path = env->GetDirectoryPath(DIRBASE::OPENRCT2);
    String::Set(outPath, outSize, path.c_str());
}
```
When trying to simply replace the instances of the two methods and their callers (not these ones, the ones above), as well as including the method doesn't get detected which means that i don't know how to call the function correctly.

PR #9017, opened 2019-04-02
Hungarian Forint OpenRCT2#8029

hokasha2016:
Feature request OpenRCT2#8029.This commit adds the Hungarian Forint, HUF, to the list of available currencies in the game. I am unsure of the exchange rate in ```Currency.cpp```, however to my knowledge they do not have an equivalent to cents or penny's. I placed 1000, but it could be easily altered.


duncanspumpkin:
Changelog as well.

PR #9039, opened 2019-04-03
Don't store multithreading option permanently.

ZehMatt:
I need to look at all the bugs that have been submitted and come up with better automated tests regarding languages like Korean that use TTF rendering.

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3753,7 +3753,7 @@ STR_6302    :{SMALLFONT}{BLACK}Copy the entire list of missing objects to the cl
 STR_6303    :Downloading object ({COMMA16} / {COMMA16}): [{STRING}]
 STR_6304    :Open scenery picker
 STR_6305    :Multithreading
-STR_6306    :{SMALLFONT}{BLACK}Use multiple threads to render
+STR_6306    :{SMALLFONT}{BLACK}Experimental option to use multiple threads to render, may cause instability.
 
 #############
 # Scenarios #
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Rather than making it non-persistent without UI indication, I think we should just default to off instead. Is that an option at this stage, or do you actively want to reset existing config files?

janisozaur:
Maybe make it false if using TTF only? Defaulting to off is not really an option once it's stored in config file.

ZehMatt:
There might be more unrelated bugs, so I still want to give people a chance to turn it on when they go to the options just not keep it saved so it would crash next start.

janisozaur:
It's worth taking Aaron's suggestion and at least modify the text of the option with mention of the issue # on GitHub.

AaronVanGeffen:
Or change the option's name so all users are reset to false once, as @IntelOrca suggested in chat.

ZehMatt:
Good idea actually

ZehMatt:
Renamed the option to "multi_threading" and extended the tooltip string since the option window doesn't have much space to put it on the label.

duncanspumpkin:
We change it back to the original. Eventually the old entry will go away.

janisozaur:
Do we remove options not consumed by our structures? Now that I think of it, it should work regardless of we do or not. In case we do, we will set it to true and if we don't, it would have already been true, unless user manually turned it off. Looks fine to me then

IntelOrca:
@janisozaur we do, it re-writes the whole file.

PR #9059, opened 2019-04-06
Implement Banner Place/Remove Actions

duncanspumpkin:
The end is in sight.


src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -2597,20 +2612,36 @@ static money32 try_place_ghost_scenery(
...
             // Banners
             // 6e2612
-            cost = game_do_command(
-                map_tile.x, parameter_1 | 0x69, map_tile.y, parameter_2, GAME_COMMAND_PLACE_BANNER, parameter_3, 0);
+            uint8_t direction = (parameter_2 >> 8) & 0xFF;
+            int32_t z = (parameter_2 & 0xFF) * 16;
+            CoordsXYZD loc{ map_tile.x, map_tile.y, z, direction };
+            auto primaryColour = gWindowSceneryPrimaryColour;
+            auto bannerType = (parameter_1 & 0xFF00) >> 8;
+            auto bannerIndex = create_new_banner(0);
```
ZehMatt: I'm a little worried about what if two clients are placing banners at the same time, this would collide

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -1906,12 +1907,26 @@ static void window_top_toolbar_scenery_tool_down(int16_t x, int16_t y, rct_windo
...
-            gGameCommandErrorTitle = STR_CANT_POSITION_THIS_HERE;
-            game_command_callback = game_command_callback_place_banner;
-            game_do_command(
-                gridX, flags, gridY, parameter_2, GAME_COMMAND_PLACE_BANNER, parameter_3, gWindowSceneryPrimaryColour);
+            uint8_t direction = (parameter_2 >> 8) & 0xFF;
+            int32_t z = (parameter_2 & 0xFF) * 16;
+            CoordsXYZD loc{ gridX, gridY, z, direction };
+            auto primaryColour = gWindowSceneryPrimaryColour;
+            auto bannerType = (parameter_1 & 0xFF00) >> 8;
+            auto bannerIndex = create_new_banner(0);
```
ZehMatt: I'm a little worried about what if two clients are placing banners at the same time, this would collide

duncanspumpkin: It would fail the check in the game action if two tried at the same tick.

ZehMatt: Not sure how I feel about that if I were to fight with someone placing banners at the same time I am, either I or the other person would basically get randomly some error, I mean its not that likely to happen but if two people are currently mass placing banners they will notice. One way to compensate this a little bit could by trying it a few times if the callback says via error message that there is a collision of banner index and then just use the next free one and try again, but then it would better if ghost sprites were handled differently, best not stored in the sprite list at all rather having a vector of sprites that are supposed to be ghost pieces.

duncanspumpkin: This is the same principle used for all of the banner index allocations.

src/openrct2/actions/BannerPlaceAction.hpp
```diff
@@ -0,0 +1,199 @@
...
+#include "../world/MapAnimation.h"
+#include "../world/Scenery.h"
+#include "GameAction.h"
+
+DEFINE_GAME_ACTION(BannerPlaceAction, GAME_COMMAND_PLACE_BANNER, GameActionResult)
+{
+private:
+    CoordsXYZD _loc;
+    uint8_t _bannerType{ std::numeric_limits<uint8_t>::max() };
+    uint8_t _bannerIndex{ BANNER_INDEX_NULL };
```
ZehMatt: It seems we have the type 'BannerIndex' for this, if we ever change the type to raise the limit would be probably better to use that.

src/openrct2/actions/BannerPlaceAction.hpp
```diff
@@ -0,0 +1,199 @@
...
+    uint8_t _primaryColour;
+
+public:
+    BannerPlaceAction() = default;
+    BannerPlaceAction(CoordsXYZD loc, uint8_t bannerType, uint8_t bannerIndex, uint8_t primaryColour)
```
ZehMatt: Also BannerIndex can be used here.

src/openrct2/world/Scenery.cpp
```diff
@@ -236,9 +237,11 @@ void scenery_remove_ghost_tool_placement()
     if (gSceneryGhostType & SCENERY_GHOST_FLAG_4)
     {
         gSceneryGhostType &= ~SCENERY_GHOST_FLAG_4;
-        constexpr uint32_t flags = GAME_COMMAND_FLAG_APPLY | GAME_COMMAND_FLAG_GHOST | GAME_COMMAND_FLAG_ALLOW_DURING_PAUSED
-            | GAME_COMMAND_FLAG_NO_SPEND;
-        game_do_command(x, flags, y, z | (gSceneryPlaceRotation << 8), GAME_COMMAND_REMOVE_BANNER, 0, 0);
+
+        auto removeSceneryAction = BannerRemoveAction({ x, y, z * 8, gSceneryPlaceRotation });
+        removeSceneryAction.SetFlags(
+            GAME_COMMAND_FLAG_GHOST | GAME_COMMAND_FLAG_ALLOW_DURING_PAUSED | GAME_COMMAND_FLAG_NO_SPEND);
```
ZehMatt: Is GAME_COMMAND_FLAG_NO_SPEND really required here given that its already a ghost?

duncanspumpkin: Its more for consistency than anything else. The no spend flag is always (i think) sent with the ghost flag.

PR #9065, opened 2019-04-07
Update String.cpp

hokasha2016:
This is implementing code fix given in OpenRCT2#7871 in String.cpp.

### DIFF
--- a/src/openrct2/core/String.cpp
+++ b/src/openrct2/core/String.cpp
@@ -184,9 +184,9 @@ namespace String
     {
         if (ignoreCase)
         {
-            while (*str != '\0' && *match != '\0')
+            while (*match != '\0')
             {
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This would fix the issue, but I personally would rather see the logic inverted:

```diff
            while (*str != '\0' && *match != '\0')
            {
-               if (tolower(*str++) != tolower(*match++))
+               if (tolower(*str++) == tolower(*match++))
                {
                    return true;
                }
            }
            return false;
```

Same for the other case.

Gymnasiast:
Ideally, we'd want to include a test that ensures it is fixed (and doesn't get broken again). OpenRCT2 has a test suite for this, and it already includes numerous localisation tests.

It is not an _absolute_ requirement, but perhaps it's a nice thing for you guys to figure out.

PR #9070, opened 2019-04-07
Fix #9068: Unable to place staff or peeps during multiplayer

### LINKED ISSUE #9068: Unable to place down staff in multiplayer server
**OS:** Windows 10
**Version:** 0.2.2
**Commit/Build:** 7376e16

When you pick up an employee in a multiplayer server, you cannot place them down.
An empty error message appears:
image

ZehMatt:
Small regression from #8905 as it didn't network the player pickup info. Also the reason is for try_get_sprite is that this info can be sen't before the map is fully loaded which would just trigger an assert otherwise.

### DIFF
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -1934,7 +1934,13 @@ void Network::ProcessPlayerInfo()
         auto* player = GetPlayerByID(it->second.Id);
         if (player != nullptr)
         {
-            *player = it->second;
+            const NetworkPlayer& networkedInfo = it->second;
+            player->Flags = networkedInfo.Flags;
+            player->Group = networkedInfo.Group;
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Does it really need to pass these variables every packet?

ZehMatt:
The other way would be to selectively copy them I guess, would be a lot easier if it were to use the DataSerialiser but good point.

duncanspumpkin:
If we just went and did it in a game action would that solve it better

ZehMatt:
The issue atm is that it overwrites the entire NetworkPlayer struct in Network::ProcessPlayerInfo, thats why I added them into the stream. But yeah it feels better not sending more, I'll overwrite the commit.

ZehMatt:
Done.

PR #9072, opened 2019-04-07
PlayerSetGroupAction

duncanspumpkin:
I'm unsure how this action is meant to work. How do I get a ref to gNetwork? I'm assuming its purposely not in the header file. Discussion welcome.


duncanspumpkin:
This works fine now. Its basically just a wrapper for the existing function.

ZehMatt:
Appears that no errors are shown for when I'm trying to change the hosts group or try to change a group without permissions.

ZehMatt:
It is functional, about the errors I will open a PR fixing it.

Issue #7729, opened 2018-06-25
Money Input Prompt break on certain values


Broxzier:
**OS:** Windows 10
**Version:** 0.2.0
**Commit/Build:** 9e052c7

Two issues:
1. Setting a ride's price to something ignores the decimal part for values over 3 when the decimal part is `.10` (3.10, 4.10, 5.10, etc..). Other decimal values seem to work fine.
2. Entering high values in the 'Add/set money' cheat gives back a seemingly random amount. For example, entering `100,000.00` gives you `34,464.00` cash.

jmyre1999:
Is this bug currently being worked on by anybody?

Gymnasiast:
@jmyre1999 No, as far as I know, nobody is working on it. If you would like to work on it, please do!

jmyre1999:
@Gymnasiast I tried to replicate part 1 and it seemed to work as intended

joshtucker132:
I am working with @jmyre1999 on this and I'm having issues with floating point division when trying to fix part 1.

https://github.com/OpenRCT2/OpenRCT2/blob/89330703e0f84b0e94dd4236c4b502a16d7fade8/src/openrct2/localisation/Localisation.cpp#L1513-L1516

For some entered money values (ex. 6.10), `number` evaluates to 6.099999. This causes `fraction` to evaluate to 0 instead of .1. I have tried multiple ways to round `number` to two decimal places (ceiling/floor functions, casting to int and dividing by 100.0, etc.) but no method works as expected. Any particular reason for this error? Even simple division like 610 / 100.0 evaluates to 6.099999 which seems odd.

ZehMatt:
Have a look at this example http://cpp.sh/9ao5v you want to use long double and suffix L.

joshtucker132:
`auto rounded = (static_cast<long>(number * 100 + .5) - (whole * 100)) / 100.0;
    auto fraction = static_cast<uint8_t>(rounded * 100);`

@ZehMatt I tried to implement your solution, but it still did not give me the rounded value that I wanted. I found a work-around by performing the subtraction before the division. It is a little messy but fixes the issue.

PR #9086, opened 2019-04-10
Issue #7729: Money Input Prompt break on certain values fix

jmyre1999:
Fixed errors in the string_to_money function that caused larger input values to not be properly converted and caused decimal values of 10 to not get carried through the conversion (ex. 4.10 -> 4.00).

Additional commit was done to add "thread_local" back to variables that lost it for an unknown reason.

### DIFF
--- a/src/openrct2/localisation/Localisation.cpp
+++ b/src/openrct2/localisation/Localisation.cpp
@@ -35,9 +35,9 @@
 #include <iterator>
 #include <limits.h>
 
-char gCommonStringFormatBuffer[512];
-uint8_t gCommonFormatArgs[80];
-uint8_t gMapTooltipFormatArgs[40];
+thread_local char gCommonStringFormatBuffer[512];
+thread_local uint8_t gCommonFormatArgs[80];
+thread_local uint8_t gMapTooltipFormatArgs[40];
 
 #ifdef DEBUG
 // Set to true before a string format call to see details of the formatting.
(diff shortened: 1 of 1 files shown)


src/openrct2/localisation/Localisation.cpp
```diff
@@ -1512,8 +1512,9 @@ money32 string_to_money(const char* string_to_monetise)
 
     auto number = std::stod(processedString, nullptr);
     number /= (currencyDesc->rate / 10.0);
-    auto whole = static_cast<uint16_t>(number);
-    auto fraction = static_cast<uint8_t>((number - whole) * 100);
```
Gymnasiast: What you will likely work, but it's not really the _right_ way to fix it. It's something I'd only want to do if there is no other way.

What happens if you change `100` to `100.0`? And if that alone doesn't work, what happens if you do `static_cast<uint8_t>(ceil((number - whole) * 100.0))`?

joshtucker132: In my testing, none of the expected solutions worked. I tried changing to 100.0, using ceiling, floor, rounding functions, etc. I'm not sure if it could have been a problem with my machine, but every fix resulted in a floating point error unless I performed the subtraction before assigning to variable.

Gymnasiast: I have done some tests myself, and found a way to do it using `ceil`. I'll therefore close this PR, but I might reopen it should my solution not be universal.

PR #9087, opened 2019-04-11
Fix #8800: ensure tile_element_height is used correctly

### LINKED ISSUE #8800: Ensure tile_element_height is used correctly
`tile_element_height` is often used incorrectly. It returns the surface height `int16_t` and also the water height `int16_t` in one `sint32_t`.

The function needs to be modified to return a tuple and all callees need to be checked to ensure they are using the correct parts.

nicolewright:
Created tile_element_water_height function. Updated old references to match with the new function.


src/openrct2/actions/SmallSceneryPlaceAction.hpp
```diff
@@ -75,7 +75,8 @@ DEFINE_GAME_ACTION(SmallSceneryPlaceAction, GAME_COMMAND_PLACE_SCENERY, GameActi
         {
             supportsRequired = true;
         }
-        int32_t baseHeight = tile_element_height(_loc.x, _loc.y);
+        int32_t baseHeight = tile_element_height(_loc.x, _loc.y); // this is really just land height now
+        int16_t waterHeight = tile_element_water_height(_loc.x, _loc.y);
         // If on water
         if (baseHeight & 0xFFFF0000)
```
duncanspumpkin: This function is using the water height and will need adjustment

src/openrct2/scenario/Scenario.cpp
```diff
@@ -435,7 +435,7 @@ static int32_t scenario_create_ducks()
     if (!map_is_location_in_park({ x, y }))
         return 0;
 
-    centreWaterZ = (tile_element_height(x, y) >> 16) & 0xFFFF;
+    centreWaterZ = (tile_element_height(x, y) >> 16);
```
duncanspumpkin: Water height not surface height

src/openrct2/scenario/Scenario.cpp
```diff
@@ -447,7 +447,7 @@ static int32_t scenario_create_ducks()
     {
         for (j = 0; j < 7; j++)
         {
-            waterZ = (tile_element_height(x2, y2) >> 16) & 0xFFFF;
+            waterZ = (tile_element_height(x2, y2) >> 16);
```
duncanspumpkin: Water height

src/openrct2/world/Map.cpp
```diff
@@ -1080,6 +1106,362 @@ uint8_t map_get_highest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, in
     return max_height;
 }
 
+
+static money32 raise_land(
```
duncanspumpkin: This function is now removed delete

src/openrct2/world/Map.cpp
```diff
@@ -1080,6 +1106,362 @@ uint8_t map_get_highest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, in
...
+    _currentTrackSelectionFlags |= TRACK_SELECTION_FLAG_RECHECK;
+
+    gCommandExpenditureType = RCT_EXPENDITURE_TYPE_LANDSCAPING;
+    gCommandPosition.x = x;
+    gCommandPosition.y = y;
+    gCommandPosition.z = z;
+    return cost;
+}
+
+static money32 lower_land(
```
duncanspumpkin: Function removed delete

src/openrct2/world/Map.cpp
```diff
@@ -1080,6 +1106,362 @@ uint8_t map_get_highest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, in
...
+
+    gCommandExpenditureType = RCT_EXPENDITURE_TYPE_LANDSCAPING;
+    gCommandPosition.x = x;
+    gCommandPosition.y = y;
+    gCommandPosition.z = z;
+    return cost;
+}
+
+money32 raise_water(int16_t x0, int16_t y0, int16_t x1, int16_t y1, uint8_t flags)
+{
```
duncanspumpkin: Function removed

src/openrct2/world/Map.cpp
```diff
@@ -1080,6 +1106,362 @@ uint8_t map_get_highest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, in
...
+        }
+    }
+
+    // Force ride construction to recheck area
+    _currentTrackSelectionFlags |= TRACK_SELECTION_FLAG_RECHECK;
+
+    return cost;
+}
+
+money32 lower_water(int16_t x0, int16_t y0, int16_t x1, int16_t y1, uint8_t flags)
```
duncanspumpkin: Removed

src/openrct2/world/Map.cpp
```diff
@@ -1080,6 +1106,362 @@ uint8_t map_get_highest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, in
...
+    _currentTrackSelectionFlags |= TRACK_SELECTION_FLAG_RECHECK;
+
+    return cost;
+}
+
+/**
+ *
+ *  rct2: 0x0068C542
+ */
+void game_command_raise_land(
```
duncanspumpkin: Removed

src/openrct2/world/Map.cpp
```diff
@@ -1080,6 +1106,362 @@ uint8_t map_get_highest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, in
...
+    *ebx = raise_land(
+        *ebx, *eax, *ecx, tile_element_height(*eax, *ecx), (int16_t)(*edx & 0xFFFF), (int16_t)(*ebp & 0xFFFF), *edx >> 16,
+        *ebp >> 16, *edi & 0xFFFF);
+}
+
+/**
+ *
+ *  rct2: 0x0068C6D1
+ */
+void game_command_lower_land(
```
duncanspumpkin: Removed

src/openrct2/world/Map.h
```diff
@@ -151,7 +151,10 @@ SmallSceneryElement* map_get_small_scenery_element_at(int32_t x, int32_t y, int3
 EntranceElement* map_get_park_entrance_element_at(int32_t x, int32_t y, int32_t z, bool ghost);
 EntranceElement* map_get_ride_entrance_element_at(int32_t x, int32_t y, int32_t z, bool ghost);
 EntranceElement* map_get_ride_exit_element_at(int32_t x, int32_t y, int32_t z, bool ghost);
-int32_t tile_element_height(int32_t x, int32_t y);
+int16_t tile_element_height(int32_t x, int32_t y);
+uint8_t map_get_highest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, int32_t yMax);
+uint8_t map_get_lowest_land_height(int32_t xMin, int32_t xMax, int32_t yMin, int32_t yMax);
+int16_t tile_element_water_height(int32_t x, int32_t y);
```
duncanspumpkin: This should only be the one new function and the change in return value

Nazey: Did you relocate the other three functions? A bit confused on this

Nazey: I removed them but the build still fails. Any reason for this?

janisozaur: The build error is quite clear:
```
../src/openrct2-ui/../openrct2/actions/SmallSceneryRemoveAction.hpp:135:27: error: 'tile_element_height' was not declared in this scope
         res->Position.z = tile_element_height(res->Position.x, res->Position.y);
                           ^~~~~~~~~~~~~~~~~~~
```

janisozaur:
0c7dbeb8a96529f3f6aa2ebd2e5cbede926db5a1 does not conform to the commit message format: https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

src/openrct2/world/Map.h
Nazey: @janisozaur So, tile_element_height still must remain? I know one of my other members pushed an update to fix all of these conflicts, but it doesn't seem like that fixed the appveyor build nor the traviscl.

Nazey:
Very confused as to what to do from here.

duncanspumpkin:
Have a look at SmallSceneryPlaceAction.hpp there are a few issues with it. L81 - 83 need changed, L136 - 141 need changed same for L276, L323. I've not checked the whole PR but I expect there are other functions like that which need some more care and attention.

Nazey:
Think we got everything. We made sure to go back through every file and check the callees to `tile_element_height`

Nazey:
@duncanspumpkin There is some more &ing in a couple of different files, like `Viewport.cpp` such as line 224 

`height = tile_element_height((0xFFFF) & pos.x, (0xFFFF) & pos.y);` 

and line 641

` int32_t height = (tile_element_height(0xFFFF & sprite->generic.x, 0xFFFF & sprite->generic.y)) - 16;` 

Everywhere where the x, y, or result is &'d with 0xFFFF should be changed, correct?

duncanspumpkin:
For now i was only caring about the result.

PR #9089, opened 2019-04-11
Fix #9083: Wrong sign index is set for clients

### LINKED ISSUE #9083: Sign texts overwrite newest sign in multiplayer
**OS:** Windows 10
**Version:** 0.2.2
**Commit/Build:** e01b5f9

When someone renames a sign in multiplayer, the newest sign in the map will get the text, and the older one will just be renamed "sign".

This is very annoying because we use signs to show who is building where in multiplayer.

Possibly related, when someone renamed a shop one of the rides automatically got the same name as the shop.

Broxzier:
This issue was caused by the ghost for banners, which already create an entry on the banner. The game command is then received while the ghost still exists, giving the placed banner another ID.

This fix is basically a copy a workaround from `game_do_command_p` that tackles this same issue. Having moved over the `GAME_COMMAND_PLACE_LARGE_SCENERY` command to game actions, this fix was no longer working.

https://github.com/OpenRCT2/OpenRCT2/blob/8632eb7297de719f6f192fe34eeac6173bd6aa00/src/openrct2/Game.cpp#L411-L414

### DIFF
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -31,7 +31,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "18"
+#define NETWORK_STREAM_VERSION "19"
 #define NETWORK_STREAM_ID OPENRCT2_VERSION "-" NETWORK_STREAM_VERSION
 
 static Peep* _pickup_peep = nullptr;
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -1976,6 +1976,17 @@ void Network::ProcessGameCommands()
 
         if (gc.action != nullptr)
         {
+            // Remove ghost scenery so it doesn't interfere with incoming network command
+            switch (gc.action->GetType())
+            {
+                case GAME_COMMAND_PLACE_WALL:
+                case GAME_COMMAND_PLACE_LARGE_SCENERY:
+                case GAME_COMMAND_PLACE_BANNER:
+                case GAME_COMMAND_PLACE_SCENERY:
+                    scenery_remove_ghost_tool_placement();
+                    break;
+            }
+
             GameAction* action = gc.action.get();
             action->SetFlags(action->GetFlags() | GAME_COMMAND_FLAG_NETWORKED);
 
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
Yes that would fix it. I'd like to combine it with https://github.com/duncanspumpkin/OpenRCT2/tree/fix_signs as we shouldn't just blindly allocate the ID.

duncanspumpkin:
I've not tested this change. 

What I'd like to do is make it so that we don't remove ghosts every update. But need to make sure we are able to place actual pieces ontop without causing collision issues and that removing the ghost doesn't remove the actual.

src/openrct2/network/Network.cpp
```diff
@@ -1948,6 +1948,9 @@ void Network::ProcessPlayerInfo()
 
 void Network::ProcessGameCommands()
 {
+    // Remove ghost scenery so it doesn't interfere with incoming network command
+    scenery_remove_ghost_tool_placement();
```
duncanspumpkin: This should be moved into L1982 and protected by an if gc->GetType() == GAME_COMMAND_PLACE_WALL, LARGE_SCENERY, BANNER... 

At present it just makes ghosts redraw every tick which makes ghosts that have some random element (such as rotation) constantly change.

Broxzier:
> What I'd like to do is make it so that we don't remove ghosts every update.

For signs and banners I agree that should be right, but for other elements this will require a lot of checks everywhere to make sure things stay in sync. The ideal situation is that ghost are purely visual and that nothing else is even aware of their existence. Right now, with paths for example, ghosts directly affect their neighbours edges. Not removing them and placing them every frame could make peeps walk differently.

src/openrct2/network/Network.cpp
```diff
@@ -3657,7 +3665,8 @@ void game_command_kick_player(
     NetworkPlayer* kicker = gNetwork.GetPlayerByID(game_command_playerid);
     if (player == nullptr)
     {
-        // Player might be already removed by the PLAYERLIST command, need to refactor non-game commands executing too early.
+        // Player might be already removed by the PLAYERLIST command, need to refactor non-game commands executing too
+        // early.
```
Broxzier: Not sure how this happened. I suspect clang-format, but it doesn't happen again when I try to trigger it.

duncanspumpkin:
yeah the ghosts staying will have to wait for a different PR. Could you cherry pick in my change as well? I would do it myself but I'm not at my dev PC.

Broxzier:
Sure thing, I'll do that when I get home tonight.

duncanspumpkin:
Argh I forgot to clang format my commit.

Issue #6808, opened 2017-12-10
Inventarisation of `interface/window.c` uses.


marijnvdwerf:
##### actions/GameActionCompat.cpp
  - window_close
  - window_close_by_class
  - window_close_by_number
  - window_find_by_class
  - window_find_by_number
  - window_invalidate_by_class
  - window_invalidate_by_number

##### actions/GameActionRegistration.cpp
  - window_close
  - window_close_by_class
  - window_close_by_number
  - window_find_by_class
  - window_find_by_number
  - window_invalidate_by_class
  - window_invalidate_by_number

##### audio/Audio.cpp
  - window_get_previous_viewport

##### Cheats.cpp
  - window_invalidate_by_class

##### cmdline/ConvertCommand.cpp
  - window_close_by_class

##### Context.cpp
  - window_close_all

##### drawing/lightfx.c
  - window_get_main

##### drawing/Rain.cpp
  - gWindowNextSlot
  - g_window_list

##### drawing/string.c
  - gCurrentWindowColours

##### drawing/X8DrawingEngine.cpp
  - window_draw_all
  - window_reset_visibilities
  - window_update_all
  - window_update_all_viewports

##### Editor.cpp
  - gWindowUpdateTicks
  - tool_cancel
  - window_close_all
  - window_find_by_class
  - window_get_main
  - window_invalidate
  - window_invalidate_by_class
  - window_set_location

##### Game.cpp
  - gWindowMapFlashingFlags
  - gWindowUpdateTicks
  - tool_cancel
  - window_close_by_class
  - window_dispatch_update_all
  - window_get_main
  - window_invalidate
  - window_invalidate_by_class
  - window_resize_gui
  - window_update_all

##### interface/console.c
  - gWindowUpdateTicks
  - window_close_all
  - window_get_main
  - window_set_location
  - window_set_window_limit

##### interface/Screenshot.cpp
  - window_get_main

##### interface/Theme.cpp
  - gWindowNextSlot
  - g_window_list

##### interface/viewport.c
  - gWindowNextSlot
  - g_window_list
  - textinput_cancel
  - window_draw_all
  - window_event_resize_call
  - window_find_from_point
  - window_get_main
  - window_init_all
  - window_invalidate
  - window_is_visible

##### interface/viewport_interaction.c
  - window_find_by_class
  - window_find_from_point

##### interface/widget.c
  - gCurrentTextBox
  - gTextBoxFrameNo
  - gTextBoxInput
[...]

Issue #8731, opened 2019-02-20
Issue with Host moving peeps or staff members desync multiplayer


ageekhere:
**OS:** Windows 10
**Version:** Version 0.2.1 build 971480b
**Commit/Build:** 8ac731e
**Map** crazy castle
**Number of players** 2

I am guessing this is already i known issue, just posting my findings. 

When the Host picks up a peep or staff member and moves it, it will often trigger a desync for the clients. However if a client moves a peep or staff member it does not seem to cause a desync (More testing is still needed to determine if a client can cause a desync by moving peeps and staff members).

It seems to be an issue with the host moving peeps or staff member (creating staff members too). Could be that the host is not sending the correct data to the clients.
 
- [ ] Reproducible in RCT2 (vanilla)? NA
- [ ] Specific to multiplayer? Yes

**Steps to reproduce:**
1. Host selects a peep or staff member and moves it to a new location
2. Client often desync

ZehMatt:
Thanks for the report. We are currently working on moving more game commands towards the new game action system to prevent desyncs, moving the peep is not ported yet so please bare with us, we are working on it.

ageekhere:
Has this issue been fixed?

ZehMatt:
Yes

PR #9098, opened 2019-04-15
Fix #5893: Invalidate widget only if it's visible

### LINKED ISSUE #5893: Assertion failed: (w->widgets[i].type != WWT_LAST), function widget_invalidate, file ../src/openrct2/interface/window.c, line 994.
**OS:** Mac OS X 10.12.5
**Version:** master
**Commit/Build:** 1aed33192 + 1 irrelevant commit of my own (#5892)

I didn't do anything specific when this assertion triggered. IIRC a peep just walked past a drink stand, I clicked it, it was thinking "I'm not thirsty", and a second later this happened. Perhaps this peep was just thinking about something to do, but maybe it is completely unrelated.

I don't know if this is easily reproducible either. Happened during a singleplayer game. Maybe the stack trace helps:

Application Specific Information:
Assertion failed: (w->widgets[i].type != WWT_LAST), function widget_invalidate, file ../src/openrct2/interface/window.c, line 994.
 

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fffb50b9d42 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fffb51a7457 pthread_kill + 90
2   libsystem_c.dylib             	0x00007fffb501f420 abort + 129
3   libsystem_c.dylib             	0x00007fffb4fe6893 __assert_rtn + 320
4   libopenrct2.dylib             	0x000000010c93987d widget_invalidate + 189
5   libopenrct2.dylib             	0x000000010c96ec0c peep_pick_ride_to_go_on + 1356
6   libopenrct2.dylib             	0x000000010c965e0b sub_68F41A + 1755
7   libopenrct2.dylib             	0x000000010c965387 peep_update_all + 151
8   libopenrct2.dylib             	0x000000010c91f6a4 game_logic_update + 196
9   libopenrct2.dylib             	0x000000010c91f4f7 game_update + 599
10  libopenrct2.dylib             	0x000000010c98cc0e rct2_update + 238
11  libopenrct2.dylib             	0x000000010ce68517 OpenRCT2::Context::RunFixedFrame() + 183
12  libopenrct2.dylib             	0x000000010ce68029 OpenRCT2::Context::RunGameLoop() + 153
13  libopenrct2.dylib             	0x000000010ce676ea OpenRCT2::Context::Launch() + 746
14  libopenrct2.dylib             	0x000000010ce66b89 OpenRCT2::Context::RunOpenRCT2(int, char**) + 57
15  openrct2                      	0x000000010c8b0c4c main + 236
16  libdyld.dylib                 	0x00007fffb4f8b235 start + 1

w-flo:
When reviewing, please take into account that I have looked at OpenRCT2 code for the first time a few days ago, so maybe this pull request is terrible. This fixes #5893 for me and I have verified using the software and OpenGL renderers that action labels in guest windows update correctly when guests leave a ride, then head to the same ride again and either re-enter the queue or fail to get in line because the queue is full. However, I'm having a hard time to verify that this doesn't break anything, because the action label seems to update without any invalidate calls anyway, maybe because of the viewport in the guest window that updates all the time and is so close to the label?

Note:
- This removes calls to "window_event_invalidate_call(w);". As far as I can tell, that function executes an event callback like [window_guest_overview_invalidate()](https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2-ui/windows/Guest.cpp#L1031), and that just fixes the guest window layout. The layout shouldn't change when a guest decides to head to some ride (or stops heading to a ride), so I thought that call could be dropped. If it can't, it could probably be added inside the new if body in window_guest_overview_update().
- This changes the peep->window_invalidate_flags value for most guests by setting one bit to 1 that was probably 0 before (unless the guest window for that guest is open). As far as I can tell, that value is stored in save files. It shouldn't cause issues in OpenRCT2, but I can't test / tell if it causes any issues in vanilla RCT2 when loading a save game that has this flag set for guests. The field name in OpenRCT2 sounds so innocent, but that's just OpenRCT2 I guess?

Also, the flag name is a bit weird, considering that it is only for "heading_to_ride has changed" right  now. When looking at it as a way for libopenrtc2 to signal to the -ui project that it should invalidate the action label it makes more sense (but the action label could change for other reasons and libopenrtc2 calls a full window_invalidate() on its own in those cases). It should be possible to use the flag in other places where libopenrtc2 currently calls window_invalidate() after changing the guest state, but I wanted to change only those cases where the existing code base calls widget_invalidate(), because that's where my game crashes because of the failed assertion.

Commit message:
Remove the widget_invalidate() call after changing a guest's
guest_heading_to_ride_id, because that call fails a debug assertion if
the guest window is open and the currently active tab is not the
"overview" tab. In Release builds (if assertion is disabled),
widget_invalidate() might access the widgets array out of bounds.

Instead, introduce a new flag PEEP_INVALIDATE_PEEP_ACTION for
window_invalidate_flags in the peep struct and set that flag. The guest
window update function then makes sure to invalidate the label if the
flag is set.

The flag could be used in other places to reduce libopenrct2 dependency
on window_*() calls (see #6808), but this commit only cares about cases
where the assertion would fail.


src/openrct2/peep/Peep.h
```diff
@@ -511,6 +511,7 @@ enum PeepInvalidate
     PEEP_INVALIDATE_PEEP_2 = 1 << 2,
     PEEP_INVALIDATE_PEEP_INVENTORY = 1 << 3,
     PEEP_INVALIDATE_STAFF_STATS = 1 << 4,
+    PEEP_INVALIDATE_PEEP_ACTION = 1 << 5     // Currently set only when guest_heading_to_ride_id is changed
```
AaronVanGeffen:
```suggestion
    PEEP_INVALIDATE_PEEP_ACTION = 1 << 5, // Currently set only when guest_heading_to_ride_id is changed
```

AaronVanGeffen:
Nice work! Please feel free to add yourself to `contributors.md` as well.

src/openrct2/peep/Peep.h
w-flo: Whoops, turns out I should actually add that pre-commit hook instead of pasting the command and accidentally dropping the *.h part. This & contributors.md is in the second commit (feel free to squash both commits into one).

AaronVanGeffen: You've accidentally left two spaces between the comma and slash, though. ;) As a result, the clang-format CI job still fails.

w-flo: Okay, this is now becoming embarrassing. The second space was a pre-commit hook test to make sure it works. And it worked! So I removed that second space again and committed, but forgot to "git add" the fixed file before committing. So I blame the pre-commit hook. :-)

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1090,6 +1089,8 @@ void window_guest_overview_invalidate(rct_window* w)
  */
 void window_guest_overview_update(rct_window* w)
 {
+    Peep* peep = GET_PEEP(w->number);
```
duncanspumpkin: Move this closer to when it is used

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1098,6 +1099,12 @@ void window_guest_overview_update(rct_window* w)
     widget_invalidate(w, WIDX_TAB_1);
     widget_invalidate(w, WIDX_TAB_2);
 
+    if (peep->window_invalidate_flags & PEEP_INVALIDATE_PEEP_ACTION)
```
duncanspumpkin: Do a nullptr check on peep.

duncanspumpkin:
Will need testing in multiplayer i think.

duncanspumpkin:
Oh also need to add a change log entry

w-flo:
Thanks for your comments!

> Will need testing in multiplayer i think.

That sounds important. I have never used multiplayer and I only have a vague idea how this would be tested. Like, is it enough to connect to a server and see if anything bad happens? Do I need to run my own server that has these changes, and then connect using a client with/without these changes? Should I simulate high network latency? Obviously I could try to do all these things, but I'm not sure if that is enough, I don't know how multiplayer is implemented in OpenRCT2 and which things this commit might break.

I will add a commit addressing your code comments and a changelog entry.

duncanspumpkin:
You just need to run two instances and make a local server and connect to that. Have one that has a peep window open so that it clears the invalidate flag. Just leave it running for a while and see if it desyncs. I'm pretty sure we already have code that prevents the invalidate flags from breaking multiplayer but i don't have the time right now to check.

w-flo:
It seems to work fine with 8 guest windows open in the server instance, and 0 windows in the connecting instance. Some of the 8 guests have already shown the "action label" message for heading_to_ride_id, so both instances should have different sets of peep window invalidate flags.

For fun, I decided to pick up peeps (both staff and guests) and sometimes this would result in a desync, but I assume that's unrelated to this PR and is simply #8731.

I also found this in the code for sprite_checksum(), which sounds promising:
```
                if (copy.generic.sprite_identifier == SPRITE_IDENTIFIER_PEEP)
                {
                    // We set this to 0 because as soon the client selects a guest the window will remove the
                    // invalidation flags causing the sprite checksum to be different than on server, the flag does not affect
                    // game state.
                    copy.peep.window_invalidate_flags = 0;
                }
```

duncanspumpkin:
Yup that was the bit of code i was thinking of. Think this is good to go then

PR #9100, opened 2019-04-15
Refactor font passing.

ZehMatt:
Pretty early thing, now would be the time to leave suggestions and ideas about what those fonts are and how they are used.

### DIFF
--- a/src/openrct2/drawing/Text.cpp
+++ b/src/openrct2/drawing/Text.cpp
@@ -141,12 +141,12 @@ static void DrawTextEllipsisedCompat(
     DrawText(dpi, x, y, &_legacyPaint, buffer);
 }
 
-void gfx_draw_string(rct_drawpixelinfo* dpi, const_utf8string buffer, uint8_t colour, int32_t x, int32_t y)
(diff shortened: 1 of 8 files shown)


src/openrct2/drawing/Text.cpp
```diff
@@ -141,12 +141,12 @@ static void DrawTextEllipsisedCompat(
     DrawText(dpi, x, y, &_legacyPaint, buffer);
 }
 
-void gfx_draw_string(rct_drawpixelinfo* dpi, const_utf8string buffer, uint8_t colour, int32_t x, int32_t y)
+void gfx_draw_string(rct_drawpixelinfo* dpi, const_utf8string buffer, uint8_t colour, int32_t x, int32_t y, TextFont font)
```
duncanspumpkin: Since the vast majority of callers use Font medium maybe that should be a default value.

ZehMatt: I also wonder if I should perhaps rename it to TextSettings/TextOptions and put the color in there as well

Gymnasiast: I agree with Duncan. It makes for more readable code IMO.

Gymnasiast:
Since this has been lying stale for 2 months I'm closing it, but feel free to reopen it if you end up working on it again.

PR #9103, opened 2019-04-17
Track importer/exporter

duncanspumpkin:
Complete. Some minor todo's

Things to do:
TrackDesign stream export/import (might leave this for now)

Things complete:
Td6 exporter
Td6 importer
Ride to new TrackDesign
Td4 importer
Td4 aa importer


src/openrct2/ride/TrackDesign.h
```diff
@@ -164,6 +164,62 @@ assert_struct_size(rct_track_td6, 0xbf);
...
+    uint8_t vehicle_additional_colour[RCT2_MAX_CARS_PER_TRAIN];
+    uint8_t lift_hill_speed;
+    uint8_t num_circuits;
+
+    std::vector<rct_td6_maze_element> maze_elements;
+    std::vector<rct_td6_track_element> track_elements;
+    std::vector<rct_td6_entrance_element> entrance_elements;
+    std::vector<rct_td6_scenery_element> scenery_elements;
+
+    std::unique_ptr<const utf8> name;
```
IntelOrca: Should this be `std::string`?

Gymnasiast:
I have dug a little deeper, and it seems like both kinds of TD4 have the same checksum magic value of `0x1A67C` (`108156`): http://tid.rctspace.com/Checksum.html . However, the SawyerEncoding::ValidateTrackChecksum() function, for some reason, incorrectly expects the value of `0x1A650` for TD4 version 0.

duncanspumpkin:
Its odd that `0x1A650` is already in the code base as a valid magic number. Must be from some legacy td4 thing.

src/openrct2/rct1/T4Importer.cpp
```diff
@@ -77,10 +77,9 @@ class TD4Importer final : public ITrackImporter
             throw IOException("Version number incorrect.");
         }
         // This is required if allow_loading_with_incorrect_checksums has created a bad value
```
Gymnasiast: It looks like you also need to remove this comment.

Gymnasiast:
Even after the last commits, TD4 designs still show the same error, whether they originate from RCT1 or AA/LL.

Do you have any TD4 designs to test with?

duncanspumpkin:
I only have one or two td4s that i downloaded a long time ago. It's possible i broke the import at some point. I'll check tomorrow

Gymnasiast:
The PR now seems to correctly handle every TD4 I threw at it, including mazes.

I did notice a bug when mirroring track designs. For example, take Crazy Cabs, mirror it, and you will notice that the signs it comes with are now shifted one tile.

duncanspumpkin:
@Gymnasiast looks like that was a bug in the current version of the game. I've fixed it now.

Issue #7700, opened 2018-06-16
park keeps crashing... Help!


scottnewm:
**OS:** MacOS
**Version:** 0.2.0
**Commit/Build:** 26666b1

im currently playing a park i download and just recently it has been crashing after a few mins of playing. the only way to make it stop crashing is to remove all guests from the park. Also if there are a bunch of guest in the park i have to keep the game paused. What should i do? I love the park and don't want to loose it. Thank you for your help!

- [ ] Reproducible in RCT2 (vanilla) no it won't load there. never has
- [ ] Specific to multiplayer? no

**Steps to reproduce:**
1. 
2. 

**Dump file**
 i have a DAT folder i attached

**Screenshots / Video:**

**Save game:**

[Six Flags Great America 2016 S01.sv6.zip](https://github.com/OpenRCT2/OpenRCT2/files/2107805/Six.Flags.Great.America.2016.S01.sv6.zip)

[DAT.zip](https://github.com/OpenRCT2/OpenRCT2/files/2107806/DAT.zip)

janisozaur:
Crash happens at 0025-05-04

https://github.com/OpenRCT2/OpenRCT2/blob/13224104e49c9f8b24070ace3e5fe87402b89ad3/src/openrct2/peep/Guest.cpp#L4101

```
>>> p vehicle_entry->peep_loading_positions
$3 = std::vector of length 0, capacity 0
```

Stack trace:
```
>>> bt full
#0  0x00007ffff6f3a94c in rct_peep::UpdateRideLeaveVehicle (this=0x7ffff7cd1ac0 <_spriteList+1618176>) at ../src/openrct2/peep/Guest.cpp:4101
        exitLocation = {
          x = 174, 
          y = 135, 
          z = 14, 
          direction = 0 '\000'
        }
        platformLocation = {
          x = 5532, 
          y = 4304, 
          z = 14, 
          direction = 2 '\002'
        }
        loadPosition = -1 '\377'
        ride = 0x7ffff78b8c90 <gRideList+96368>
        vehicle = 0x7ffff7cf94c0 <_spriteList+1780480>
        ride_station = 0 '\000'
        rideEntry = 0x7fffb5823a98
        vehicle_entry = 0x7fffb5823ab2
        __PRETTY_FUNCTION__ = "void rct_peep::UpdateRideLeaveVehicle()"
        exitLocation = {
          x = 5532, 
          y = 4304, 
          z = 14, 
[...]

PR #9120, opened 2019-04-20
Fix #8079: Crash when unloading buggy custom rides

### LINKED ISSUE #8079: Custom ride The ZAP! crashes when unloading passengers
**OS:** [e.g. Windows 10]
**Version:**  0.2.1
**Commit/Build:** 426e106

https://openrct2.org/forums/topic/3415-game-crashes-when-there-are-over-2000-guests-in-park/ If you download the save "Mystic Timbers" in the last post of this thread and open it up, it crashes within 1 second when the custom ride The ZAP! unloads. This isn't specific to the save, if you try and create it in a different map the same happens. I *assume* this ride worked for someone at sometime as otherwise it wouldn't have been uploaded (although maybe i'm wrong - I don't have vanilla around to test)

Stack trace:
 	openrct2.dll!std::vector<signed char,std::allocator<signed char> >::operator[](const unsigned __int64 _Pos) Line 1733	C++
	openrct2.dll!rct_peep::UpdateRideLeaveVehicle() Line 4092	C++
 	openrct2.dll!rct_peep::UpdateRide() Line 5065	C++
 	openrct2.dll!rct_peep::Update() Line 1267	C++
 	openrct2.dll!peep_update_all() Line 442	C++
 	openrct2.dll!OpenRCT2::GameState::UpdateLogic() Line 247	C++
 	openrct2.dll!OpenRCT2::GameState::Update() Line 141	C++
 	openrct2.dll!OpenRCT2::Context::Update() Line 940	C++
 	openrct2.dll!OpenRCT2::Context::RunFixedFrame() Line 858	C++
 	openrct2.dll!OpenRCT2::Context::RunFrame() Line 832	C++
 	openrct2.dll!OpenRCT2::Context::RunGameLoop() Line 802	C++
 	openrct2.dll!OpenRCT2::Context::Launch() Line 774	C++
 	openrct2.dll!OpenRCT2::Context::RunOpenRCT2(int argc, const char * * argv) Line 215	C++
 	openrct2.dll!NormalisedMain(int argc, const char * * argv) Line 60	C++
 	openrct2.dll!LaunchOpenRCT2(int argc, wchar_t * * argvW) Line 46	C++
 	openrct2.exe!wmain(int argc, wchar_t * * argvW, wchar_t * envp) Line 28	C++
 	[External Code]	

The problem is `vehicle_entry->peep_loading_position`s is empty so `vehicle_entry->peep_loading_positions[current_seat]` throws an out of range error. I'm not sure how this happened as I don't understand structure of rides enough.

- [] Reproducible in RCT2 (vanilla)?
- [] Specific to multiplayer?

**Steps to reproduce:**
1. Open attached save

2. Crashes within 1 second, which is exactly when The ZAP! is unloading

**Save game:**

[Mystic Timbers.zip](https://github.com/OpenRCT2/OpenRCT2/files/2466135/Mystic.Timbers.zip)

Gymnasiast:
Also fixes https://github.com/OpenRCT2/OpenRCT2/issues/8969, #7700 and https://openrct2.org/forums/topic/3642-crashing-when-playing-files/ .

Possibly also https://github.com/OpenRCT2/OpenRCT2/issues/8958 -- but that has missing objects that NE does not have.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -13,6 +13,7 @@
 - Fix: [#6006] Objects higher than 6 metres are considered trees (original bug).
 - Fix: [#7884] Unfinished preserved rides can be demolished with quick demolish.
 - Fix: [#7913] RCT1/RCT2 title sequence timing is off.
+- Fix: [#8079, #8969] Crash when unloading buggy custom rides.
 - Fix: [#8219] Faulty folder recreation in "save" folder.
 - Fix: [#8537] Imported RCT1 rides/shops are all numbered 1.
 - Fix: [#8649] Setting date does not work in multiplayer.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -13,7 +13,7 @@
 - Fix: [#6006] Objects higher than 6 metres are considered trees (original bug).
 - Fix: [#7884] Unfinished preserved rides can be demolished with quick demolish.
 - Fix: [#7913] RCT1/RCT2 title sequence timing is off.
-- Fix: [#8079, #8969] Crash when unloading buggy custom rides.
+- Fix: [#7700, #8079, #8969] Crash when unloading buggy custom rides.
 - Fix: [#8219] Faulty folder recreation in "save" folder.
 - Fix: [#8537] Imported RCT1 rides/shops are all numbered 1.
 - Fix: [#8649] Setting date does not work in multiplayer.
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -4049,6 +4049,12 @@ void Guest::UpdateRideLeaveVehicle()
                     break;
             }
         }
+        else
+        {
+            log_verbose(
+                "current_seat %d is too large! (Vehicle entry has room for %d.)", current_seat,
+                vehicle_entry->peep_loading_positions.size());
+        }
 
         platformLocation.z = ride->stations[current_ride_station].Height * 8;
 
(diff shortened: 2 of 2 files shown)


AaronVanGeffen:
Would it be possible to instead fix these rides in the S6 importer?

Broxzier:
Do you know how these indices or seat positions were set incorrectly? The thing with these checks is that, while fixing the issue, they bloat the codebase one by one.

I agree with Aaron that fixing them on load is preferable.

duncanspumpkin:
You could mess it up by changing the ride entry whilst there are peeps on the ride. I'm not sure if that is possible though.

duncanspumpkin:
Current seat is a field that is used for a few different things so i expect it might be possible that this is because of incorrectly clearing the variable.

Gymnasiast:
@broxzier I don't know what causes them. What all parks I investigated have in common is that the crash always happens when peeps exit a custom flat ride. It does not seem to be the result of changing the ride entry while the ride is open. I also have a scenario (LOTR - The Shire) that crashes very soon after opening, so it seems it's really down to badly made custom ride entries.

duncanspumpkin:
SEFLIGHT.dat is an example of a bugged object. I want to have a look to see how it is handled in vanilla.

duncanspumpkin:
In vanilla this would have returned 0. Assuming `current_seat` is `0` and `peep_loading_positions` is a zero sized array. This is because it would have accessed the next memory location which would be the size of `vehicle_entry->peep_loading_positions[1];` which is `0`.

`int8_t loadPosition = vehicle_entry->peep_loading_positions[current_seat];`

I'm not sure what would be the best solution as the object file is definitely at fault here. Could we perhaps add a verbose warning stating that the object is broken just in case we come up with a better solution one day.

Gymnasiast:
> Could we perhaps add a verbose warning stating that the object is broken just in case we come up with a better solution one day.

OK, I went with that.

duncanspumpkin:
Even though this is a logic mod and would normally need a network increment there is no need this time as the game would crash without this change so I'm merging it as is.

Issue #6116, opened 2017-08-04
Expose colour scheme in tile inspector


jensj12:
I have 10+ track pieces of the same coaster on one tile that I'd like to edit using the tile inspector. I use different colour schemes to differentiate between track pieces, but the colour scheme is not exposed in the tile inspector. 
I would like to see the colour scheme of the selected track piece in the tile inspector.

jensj12:
Closed by #9128

PR #9128, opened 2019-04-22
 Expose colour scheme in the tile inspector (#6116)

Broxzier:


Gymnasiast:
Why not just leave out the station index field for non-station elements?

Broxzier:
Because that leaves us with an empty line, and the UI code is not flexible enough to make its height depend on its content easily.

data/language/en-GB.txt
```diff
@@ -3724,7 +3724,7 @@ STR_6273    :Music
 STR_6274    :Can't set colour scheme...
 STR_6275    :{WINDOW_COLOUR_2}Station style:
 STR_6276    :{RED}{STRINGID} has guests getting stuck, possibly due to invalid ride type or operating mode.
-STR_6277    :{WINDOW_COLOUR_2}Station index: {BLACK}{COMMA16}
+STR_6277    :{WINDOW_COLOUR_2}Station index: {BLACK}{STRINGID}
```
duncanspumpkin: This change has broken all the other use for STR_6277. `TILE_INSPECTOR_PAGE_ENTRANCE` has been broken by this.

PR #9129, opened 2019-04-22
 Fix #9124: Disconnected clients crashing the server

### LINKED ISSUE #9124: Executing game action of disconnected user crashes server
**OS:** Ubuntu Server 18.04
**Version:** 0.2.2
**Commit/Build:** 

Just now one of my multiplayer servers crashed, and when looking in the logs to find the cause I found this:

```
[2019/04/21 21:47:48] Mitsubishi has disconnected (Connection Closed)
[2019/04/21 21:47:48] [sv] GA: TrackRemoveAction (00000004) (_networkId = 00000202; _flags = 80000000; _playerId = 2; _trackType = 00000000; _sequence = 00000000; _origin = CoordsXYZD(x = 2016, y = 3392, z = 144, direction = 3); ) OK
```

The action `TrackRemoveAction` has _playerId 2, looking up higher in the list this seems to be the same user as the one who just disconnected (this same user constantly keeps disconnecting because of bad internet).

**Steps to reproduce:**
1. Connect to a server with terrible internet
2. Perform a bunch of actions and hope to disconnect just before one gets executed

**Dump file**
Not sure if these are automatically generated on Ubuntu Server. It does say "Core dumped", but I have no idea where.

ZehMatt:
I made sure the correct order is now in place, first the game actions/commands are executed, then the player information if any and then the player list if it needs to. Also it was possible that one of the player list updates was skipped if two packets arrived at the same tick, it would have processed only the last one. I also made sure that only one player list update is now sent to the clients per tick which saves some bandwidth and makes it deterministic at when its sent.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -38,6 +38,7 @@
 - Fix: [#8947] Detection of AVX2 support.
 - Fix: [#8988] Character sprite lookup noticeably slows down drawing.
 - Fix: [#9000] Show correct error message if not enough money available.
+- Fix: [#9124] Disconnected clients can crash the server.
 - Fix: [#9132] System file browser cannot open SV4 files.
 - Fix: [#9152] Spectators can modify ride colours.
 - Fix: [#9202] Artefacts show when changing ride type as client or using in-game console.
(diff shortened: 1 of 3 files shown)


Gymnasiast:
@ZehMatt Looks good to me. Could you rebase this so it can be merged?

Issue #7905, opened 2018-08-14
Feature Request: Option to turn off Scroll Wheel Zooming (i.e. for trackpads)


Kuriibo:
On a laptop with a multitouch trackpad, where clicking with two fingers is a right-click, it is difficult to pan the camera, as the game often zooms in/out instead, since moving up/down with two fingers on the trackpad is treated as turning the scroll wheel. This could be alleviated easily by giving the player the option for the game to not zoom in/out when it receives input from the scroll wheel.

Gymnasiast:
To me, this sounds like a problem with the touchpad, rather than OpenRCT2, and in any case, it seems not many people want this. I'll therefore close this.

PR #9142, opened 2019-04-25
Allow disabling of scroll-wheel zooming

MichaelJBerk:
When this feature is enabled, scrolling up/down on the trackpad will not zoom in/out of the map

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3756,7 +3756,7 @@ STR_6305    :Multithreading
 STR_6306    :{SMALLFONT}{BLACK}Experimental option to use multiple threads to render, may cause instability.
 STR_6307    :Colour scheme: {BLACK}{STRINGID}
 STR_6308    :Scroll wheel behaviour:
-STR_6309    :Zoom From Centere of Screen
+STR_6309    :Zoom from centre of screen
 STR_6310    :Disabled
 STR_6311    :{SMALLFONT}{BLACK}When zooming in/out with the scroll wheel, whether to zoom at the cursor or screen centre. If set to "Disabled", scrolling will not zoom in/out (reccommended for trackpad users).
 
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3755,6 +3755,8 @@ STR_6304    :Open scenery picker
 STR_6305    :Multithreading
 STR_6306    :{SMALLFONT}{BLACK}Experimental option to use multiple threads to render, may cause instability.
 STR_6307    :Colour scheme: {BLACK}{STRINGID}
+STR_6308    :Trackpad Mode
+STR_6309    :When enabled, scrollling up/down will not zoom in/out
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3353,7 +3353,7 @@ STR_5902    :Show bounding boxes
 STR_5903    :Show paint debug window
 STR_5904    :Reset date
 STR_5905    :{SMALLFONT}{BLACK}A map generation tool that automatically creates a custom landscape
-STR_5906    :Zoom to cursor position
+STR_5906    :Zoom from cursor position
 STR_5907    :{SMALLFONT}{BLACK}When enabled, zooming in will centre around the cursor, as opposed to the screen centre.
 STR_5908    :Allow arbitrary ride type changes
 STR_5909    :{SMALLFONT}{BLACK}Allows changing ride type freely. May cause crashes.
@@ -3755,8 +3755,11 @@ STR_6304    :Open scenery picker
 STR_6305    :Multithreading
 STR_6306    :{SMALLFONT}{BLACK}Experimental option to use multiple threads to render, may cause instability.
 STR_6307    :Colour scheme: {BLACK}{STRINGID}
-STR_6308    :Trackpad Mode
-STR_6309    :When enabled, scrollling up/down will not zoom in/out
+STR_6308    :Scroll wheel behaviour:
+STR_6309    :Zoom From Centere of Screen
+STR_6310    :Disabled
+STR_6311    :{SMALLFONT}{BLACK}When zooming in/out with the scroll wheel, whether to zoom at the cursor or screen centre. If set to "Disabled", scrolling will not zoom in/out (reccommended for trackpad users).
+
 
 #############
 # Scenarios #
--- a/data/language/en-US.txt
+++ b/data/language/en-US.txt
@@ -924,11 +924,6 @@ STR_6206    :Gray stucco
 STR_6212    :Gray sandstone
 STR_6274    :Can't set color scheme...
 
-STR_6308    :Scroll wheel behavior:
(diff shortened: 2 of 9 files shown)


data/language/en-GB.txt
```diff
@@ -3755,6 +3755,8 @@ STR_6304    :Open scenery picker
 STR_6305    :Multithreading
 STR_6306    :{SMALLFONT}{BLACK}Experimental option to use multiple threads to render, may cause instability.
 STR_6307    :Colour scheme: {BLACK}{STRINGID}
+STR_6308    :Trackpad Mode
```
AaronVanGeffen: I think the label 'Trackpad Mode' has other, more involved associations for people. Might be best to stick to 'Disable scrolling'.

src/openrct2-ui/input/Input.cpp
```diff
@@ -108,7 +108,9 @@ static void game_handle_key_scroll()
     {
         window_unfollow_sprite(mainWindow);
     }
+    //scrollZOOM
```
AaronVanGeffen: Remove debug comments

AaronVanGeffen:
I'm not particularly sure what problem this intends to solve, but I'm open to the suggestion.

Instead of adding another checkbox, perhaps you could adapt the existing 'zoom to cursor' option into  a dropdown selection. Something like:

```
Scrollwheel behaviour:           [ Zoom to cursor  [▼]]
                                  [Disabled          ]
                                  [Zoom to cursor    ]
                                  [Viewport zoom     ]
```

data/language/en-US.txt
```diff
@@ -924,6 +924,11 @@ STR_6206    :Gray stucco
 STR_6212    :Gray sandstone
 STR_6274    :Can't set color scheme...
 
+STR_6308    :Scroll wheel behavior:
```
AaronVanGeffen: You should only change the en-UK file on develop. All other language files are updated through our Localisation repository. (If we'd change them here, that would lead to a conflict.)

src/openrct2/localisation/StringIds.h
```diff
@@ -3938,9 +3938,14 @@ enum
...
     STR_TILE_INSPECTOR_COLOUR_SCHEME = 6307,
+    
+    STR_SCROLL_MODE = 6308,
+    STR_SCROLL_MODE_STANDARD = 6309,
+    STR_SCROLL_MODE_DISABLED = 6310,
+    STR_SCROLL_MODE_TIP = 6311,
 
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    STR_COUNT = 32768
+    STR_COUNT = 32771
```
AaronVanGeffen: Please revert this change.

```suggestion
    STR_COUNT = 32769
```

AaronVanGeffen:
For the record, this is what it looks like at the moment.

Screenshot_20190425_233312

The options should be capitalised consistently. Personally, I prefer them to be not use title case. I also think 'Disabled' should be the first option, rather than the third.

I'd also like

src/openrct2-ui/windows/Options.cpp
```diff
@@ -1869,14 +1906,28 @@ static void window_options_invalidate(rct_window* w)
...
-            widget_set_checkbox_value(w, WIDX_ZOOM_TO_CURSOR, gConfigGeneral.zoom_to_cursor);
             widget_set_checkbox_value(w, WIDX_TOOLBAR_SHOW_FINANCES, gConfigInterface.toolbar_show_finances);
             widget_set_checkbox_value(w, WIDX_TOOLBAR_SHOW_RESEARCH, gConfigInterface.toolbar_show_research);
             widget_set_checkbox_value(w, WIDX_TOOLBAR_SHOW_CHEATS, gConfigInterface.toolbar_show_cheats);
             widget_set_checkbox_value(w, WIDX_TOOLBAR_SHOW_NEWS, gConfigInterface.toolbar_show_news);
             widget_set_checkbox_value(w, WIDX_TOOLBAR_SHOW_MUTE, gConfigInterface.toolbar_show_mute);
             widget_set_checkbox_value(w, WIDX_TOOLBAR_SHOW_CHAT, gConfigInterface.toolbar_show_chat);
-
+            
+            if (gConfigGeneral.zoom_to_cursor == false && gConfigGeneral.disable_scroll_zoom == false)
```
AaronVanGeffen: I'd prefer `zoom_to_cursor` to be replaced with a `scroll_wheel_behaviour` (integer) option, rather than adding another `disable_scroll_zoom` bool. This block can then use a switch or array-based string lookup instead.

src/openrct2-ui/interface/Window.cpp
```diff
@@ -435,9 +435,9 @@ static void window_viewport_wheel_input(rct_window* w, int32_t wheel)
 {
     if (gScreenFlags & (SCREEN_FLAGS_TRACK_MANAGER | SCREEN_FLAGS_TITLE_DEMO))
         return;
-    if (wheel < 0 && !gConfigGeneral.disable_scroll_zoom)
+    if (wheel < 0 && !gConfigGeneral.scroll_wheel_behaviour == 0)
```
AaronVanGeffen: Did you test this? `!gConfigGeneral.scroll_wheel_behaviour == 0` will bind the `!` operator to `gConfigGeneral.scroll_wheel_behaviour`, inverting the setting rather than the expression.

I'd preferred for the entire block to be nested in one `gConfigGeneral.scroll_wheel_behaviour != 0` if-block.

src/openrct2/config/Config.cpp
```diff
@@ -206,9 +205,9 @@ namespace Config
             model->last_save_landscape_directory = reader->GetCString("last_landscape_directory", nullptr);
             model->last_save_scenario_directory = reader->GetCString("last_scenario_directory", nullptr);
             model->last_save_track_directory = reader->GetCString("last_track_directory", nullptr);
+            model->scroll_wheel_behaviour = reader->GetInt32("scroll_wheel_behaviour", 1);
```
AaronVanGeffen: Zooming to cursor should still be the default. You've assigned this `2`, not `1`.

Gymnasiast:
Like Aaron, I would like to know what problem this pull request aims to solve. We generally only allow disabling functionality if it can cause problems to a significant amount of users. For mousewheel zooming, we haven't had any requests to disable it so far.

Additionally, Aaron highlighted some issues with the code itself, which you neither addressed nor replied to.

AaronVanGeffen:
It's been requested before; @MichaelJBerk mentioned #7905 in his original PR title, which was requested last year. The discussion label was never, though.

I don't object to the change, but I also wonder what problem it's solving. If the touchpad is too sensitive to zooming while panning around, I think the better solution would be to tweak that sensitivity instead.

Gymnasiast:
I don't think this feature is really useful. If a user's trackpad is too sensitive, I don't think OpenRCT2 is the correct place to mitigate it.

Besides that, the code quality of this PR is not quite up to snuff and you didn't react to Aaron's feedback.

Now, closing PRs is not something I do lightly, but I don't want to drag people around and giving them false hope.

PR #9154, opened 2019-04-27
Change map toolbar icon with current rotation

AaronVanGeffen:
While working on OpenLoco, I noticed a particular feature that we still hadn't ported over to OpenRCT2: the map icon changes depending on the current viewport rotation.

~Currently, the implementation overlays the original icon, rather than replacing it. Depending on the feedback here, I may create proper toolbar icons instead. This would include a highlighted version for the pressed state.~

## In action

Forest Frontiers 2019-04-27 16-53-26
[west](https://user-images.githubusercontent.com/604665/56851310-daed8000-690d-11e9-8b5a-52c3752986b2.png), [south](https://user-images.githubusercontent.com/604665/56851311-daed8000-690d-11e9-80ac-cc5280989d7f.png), [east](https://user-images.githubusercontent.com/604665/56851312-daed8000-690d-11e9-8bc5-751cd3e165c1.png)

## Bonus materials

As Locomotion's map icon is 24x24 pixels, rather than the 22x22 pixels in OpenRCT2's toolbar, I decided to redo the icons instead of just cropping them.

Here's what the bare map looks like without compass or grid lines. I interpolated by overlaying the Locomotion and RCT2 icons — confirming they are indeed showing the same map.

map_24x24

Added grid lines (22x22):
map_22x22_grid_only

Added compass (22x22):
map_22x22_north map_22x22_west map_22x22_south map_22x22_east

Framed (normal):
map_north map_west map_south map_east

Framed (pressed):
map_north_pressed map_west_pressed map_south_pressed map_east_pressed


src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -939,6 +939,25 @@ static void window_top_toolbar_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
 
+    // Draw map button overlay
+    if (window_top_toolbar_widgets[WIDX_MAP].type != WWT_EMPTY)
+    {
+        x = w->x + window_top_toolbar_widgets[WIDX_MAP].left + 4;
+        y = w->y + window_top_toolbar_widgets[WIDX_MAP].top + 1;
+        if (widget_is_pressed(w, WIDX_MAP))
+            y++;
+
+        static int32_t imageIdByRotation[] = {
```
IntelOrca: `const`

ZehMatt: I would even say this is constexpr. const can be still initialized during runtime, its just makes it immutable.

Broxzier:
What did you base the orientation for the compass on? Right now in many places in the code we refer to Noth/East/South/West based on the X/Y grid (positive X being South and Positive Y being East).

In a previous PR this confusion was brought up, and different conventions are apparently mixed in the code base as well (though I still haven't seen it being any different myself). Mixing things up further is not going to help (one referring to "North" as top of the screen, the other as a specific direction, and yet another based on the X/Y grid)

https://github.com/OpenRCT2/OpenRCT2/pull/8618#discussion_r251939841


In localization for the Tile Inspector, the later convention is used. https://github.com/OpenRCT2/OpenRCT2/blob/develop/data/language/en-GB.txt#L3406-L3409

These orientations are based on the flags used by some tile elements (park fences, path edges, quadrant scenery placement, etc.).

AaronVanGeffen:
Hmm, I hadn’t considered that, to be honest. The compass orientation is the same as used in  Locomotion / OpenLoco.

Not sure what the best course of action would be. I think a perpendicular compass would be clearest, though…

Gymnasiast:
Perhaps change the names to up/right/down/left for now, and sort out the naming mess later on.

AaronVanGeffen:
I appreciate the suggestion, but to me, up/right/down/left makes no sense for viewport rotation…

Gymnasiast:
I was talking about the sprite naming. I wasn't suggesting renaming the viewport rotation itself.

AaronVanGeffen:
Ah, alright. I am neutral towards the naming there. Could we please tackle that in a separate PR, though?

Gymnasiast:
@AaronVanGeffen Don't mind the naming either, but I thought Broxzier took issue, and I was trying to find a way to work around his (perceived) objections.

Broxzier:
I didn't mean that you'd have to rename anything, just change the sprites to match one orientation. But since it's not exactly clear what's best, I've created a new issue where we can discuss it. This is not a blocking matter for this PR, and could be updated later easily.

PR #9157, opened 2019-04-27
Fix #9152

### LINKED ISSUE #9152: People without permissions can change colors of rides
**OS:** Windows 10
**Version:** 0.2.2
**Commit/Build:** c8df801

Some user who didn't have permissions went on my server and turned all coasters orange. I think something similar happened on @Broxzier 's server where all rides were turned purple. I just checked and apparently you don't need permissions to recolor a ride.

From the server log:
`
[2019/04/27 04:48:15] [sv] GA: RideSetAppearanceAction (00000000) (_networkId = 00000003; _flags = 80000000; _playerId = 1 "Fridge"; _rideIndex = 111 "Wheel of Ra"; _type = 04; _value = 14; _index = 00000000; ) OK
`

ZehMatt:
The reason is pretty simple, NetworkAction uses std::array which initializes everything by default to 0, the Spectator group has only one command in it so the remaining entries would be all 0. Since GAME_COMMAND_SET_RIDE_APPEARANCE is also 0 it would of course think that the group has such command in it.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -25,7 +25,7 @@
 - Fix: [#8947] Detection of AVX2 support.
 - Fix: [#8988] Character sprite lookup noticeably slows down drawing.
 - Fix: [#9000] Show correct error message if not enough money available.
-- Fix: [#9152] Spectators being able to modify ride colours.
+- Fix: [#9152] Spectators can modify ride colours.
 - Improved: [#6116] Expose colour scheme for track elements in the tile inspector.
 - Improved: Allow the use of numpad enter key for console and chat.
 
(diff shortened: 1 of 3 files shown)


distribution/changelog.txt
```diff
@@ -25,6 +25,7 @@
 - Fix: [#8947] Detection of AVX2 support.
 - Fix: [#8988] Character sprite lookup noticeably slows down drawing.
 - Fix: [#9000] Show correct error message if not enough money available.
+- Fix: [#9152] Spectators being able to modify ride colours.
```
AaronVanGeffen: Did this really happen in past releases as well? I thought it was a recent regression.

ZehMatt: Yeah true, ill remove it.

AaronVanGeffen: As you pointed out on Gitter, it appears to be a regression of #8734 and so warrants a changelog item. :)

Gymnasiast:
@AaronVanGeffen :

> As you pointed out on Gitter, it appears to be a regression of #8734 and so warrants a changelog item. :)

I agree with Aaron here: since the issue was introduced before the 0.2.2 release, it needs a changelog entry,

Issue #8507, opened 2018-12-20
Vehicle speed underflows when going backward only in openRCT2


LordMarcel:
**OS:** Windows 10
**Version:** 0.2.1
**Commit/Build:** c457d1

When going forward there is a minimum speed of 1 km/h in all versions of the game. However, when going backward there is a minimum speed of 0 km/h in classic and vanilla, but in OpenRCT2 it will underflow to the minimum forward speed. Here are two examples:

Vanilla/Classic: https://www.youtube.com/watch?v=No22AJ7thXE

OpenRCT2: https://www.youtube.com/watch?v=HjKEKkZI3ZA

This has been in openrct2 since at least late 2016, which is when I first encountered it. For a long time I thought it a thing in vanilla as well but today someone found out it is not.

harrysbird:
This first project of openrct2 maded in 2014 not 2016

duncanspumpkin:
This will be due to replacing some arithmetic shifts with divides

bdupree5:
I would like to take a look at this any ideas where to start looking in the code?

duncanspumpkin:
@bdupree5 it will be very difficult to find where the problem is. Its to do with how the acceleration works at slow values. The code is very messy for this. vehicle.cpp is where the code resides.

bdupree5:
Ill take your word for it. I don't really want to deal with messy code lol

w-flo:
I believe I found a fix for this. The negative velocity goes up from e.g. -6000 at an acceleration of 1 until it hits -4096, then the "magical implicit forward creeping boosters" kick in to speed up the train to +32k (~1km/h) even though they shouldn't while the train is moving backwards. I can send a simple PR next week that will keep the negative speed at -4095 (about 12.5% of the forward creeping speed) that should probably mirror the RCT2 behavior.

PR #9167, opened 2019-04-29
Fix #8145: Mechanics fix vehicles that are not in the station

### LINKED ISSUE #8145: Mechanics walk to vehicles on the middle of the track
**OS:** Windows 10
**Version:** 0.2.1
**Commit/Build:** 2556171

Because a gif says more than a 1000 words:
jesusmechanic

Two weird things are happening here:
1. Obviously, the mechanic tries to walk to a coordinate in the middle of the track
2. The ride is not broken down yet, but the mechanic is already underway. A minute later or so, the ride will actually break down with a vehicle malfunction (if the mechanic hasn't visited) and will have the status 'calling mechanic'. I think the mechanic is called too early for some reasom, leading him to go to a vehicle coordinate that's somewhere in the middle of the track. If his speed is set to fast he will arrive at the track at a different time, making him go to a different part of the track.
I think this breakdown normally chooses a car to have the breakdown on, and waits for it to arrive at a station (but i haven't checked this).

Here's a save with the ride from the gif, just before it breaks down, with a mechanic already underway:
[mechanic.zip](https://github.com/OpenRCT2/OpenRCT2/files/2510099/mechanic.zip)

w-flo:
I put three independent commits in one pull request, because I believe they all need the network version bump that is only in the last commit. Obviously if that one turns out not to be acceptable, I can put the network version bump in a different commit. :-)

Two of the commits fix one bug each, and the third commit makes sure to "upgrade" a mechanic who is currently "heading for an inspection" to the "responding to breakdown call" status if that ride happens to break down. I believe this was not a reported bug, but is probably a change that makes sense in the context of fixing the "mechanic walks out of station to repair vehicle" bug. (Without that commit, the mechanic would keep the "heading for an inspection" status and then upgrade to "fixing" once they enter the station, but before that happens, players might wonder why no mechanic is called to fix the ride.)

Feel free to suggest any style changes, obviously. I hope clang-format likes my commits this time.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -2593,6 +2593,12 @@ static void ride_mechanic_status_update(Ride* ride, int32_t mechanicStatus)
                 ride->window_invalidate_flags |= RIDE_INVALIDATE_RIDE_MAINTENANCE;
                 ride_mechanic_status_update(ride, RIDE_MECHANIC_STATUS_CALLING);
             }
+            else if (
+                mechanic->state == PEEP_STATE_HEADING_TO_INSPECTION && (ride->lifecycle_flags & RIDE_LIFECYCLE_BROKEN_DOWN))
+            {
+                // Ride has broken down while mechanic was HEADING_TO_INSPECTION, upgrade to ANSWERING.
+                ride_call_mechanic(ride, mechanic, false);
+            }
             break;
         }
         case RIDE_MECHANIC_STATUS_FIXING:
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -2632,7 +2632,7 @@ static void ride_call_mechanic(Ride* ride, Peep* mechanic, int32_t forInspection
  */
 static void ride_call_closest_mechanic(Ride* ride)
 {
-    auto forInspection = (ride->lifecycle_flags & (RIDE_LIFECYCLE_BREAKDOWN_PENDING | RIDE_LIFECYCLE_BROKEN_DOWN)) == 0;
+    auto forInspection = (ride->lifecycle_flags & RIDE_LIFECYCLE_BROKEN_DOWN) == 0;
     auto mechanic = ride_find_closest_mechanic(ride, forInspection);
     if (mechanic != nullptr)
         ride_call_mechanic(ride, mechanic, forInspection);
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -5,6 +5,7 @@
 - Fix: [#475] Water sides drawn incorrectly (original bug).
 - Fix: [#6123, #7907, #9472, #11028] Cannot build some track designs with 4 stations (original bug).
 - Fix: [#7094] Back wall edge texture in water missing.
+- Fix: [#8145] Mechanics walk to vehicles in the middle of the track.
 - Fix: [#10928] File browser's date column is too narrow.
 - Fix: [#10951, #11160] Attempting to place park entrances creates ghost entrances in random locations.
 - Fix: [#11005] Company value overflows.
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -2067,8 +2067,7 @@ void Staff::UpdateFixing(int32_t steps)
     bool progressToNextSubstate = true;
     bool firstRun = true;
 
-    if ((state == PEEP_STATE_INSPECTING)
-        && (ride->lifecycle_flags & (RIDE_LIFECYCLE_BREAKDOWN_PENDING | RIDE_LIFECYCLE_BROKEN_DOWN)))
+    if (state == PEEP_STATE_INSPECTING && (ride->lifecycle_flags & RIDE_LIFECYCLE_BROKEN_DOWN))
     {
         // Ride has broken down since Mechanic was called to inspect it.
         // Mechanic identifies the breakdown and switches to fixing it.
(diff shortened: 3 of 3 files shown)


w-flo:
Some CI builds are apparently failing because of HTTP error 500 somewhere, and a checksum mismatch for a title-sequences.zip download. I guess those are unrelated to my commits.

janisozaur:
I restarted CI it's green now

src/openrct2/ride/Ride.cpp
```diff
@@ -2816,7 +2822,7 @@ static void ride_call_mechanic(Ride* ride, Peep* mechanic, int32_t forInspection
  */
 static void ride_call_closest_mechanic(Ride* ride)
 {
-    auto forInspection = (ride->lifecycle_flags & (RIDE_LIFECYCLE_BREAKDOWN_PENDING | RIDE_LIFECYCLE_BROKEN_DOWN)) == 0;
+    auto forInspection = (ride->lifecycle_flags & RIDE_LIFECYCLE_BROKEN_DOWN) == 0;
```
duncanspumpkin: This is a change from vanilla. Is it something we want to do make it that mechanics aren't called until the ride has actually broken down.

duncanspumpkin:
I don't like the fix to #8145 it is changing the way mechanics work too much. I think it would be better if when they are fixing / inspecting they just know to not go walking off the platform. If that means changing the fixing mode to a station brakes failure then that would be better.

duncanspumpkin:
#8507 fix is good. Agree with the change. Can you split up the two commits into different PR's as the mechanic one needs some further thinking and might take a while.

Gymnasiast:
> I don't like the fix to #8145 it is changing the way mechanics work too much. I think it would be better if when they are fixing / inspecting they just know to not go walking off the platform. If that means changing the fixing mode to a station brakes failure then that would be better.

Not so sure about that. I think the current logic looks fundamentally wrong.

I'm okay with splitting the PR up if that is necessary to get the other fix merged more quickly.

duncanspumpkin:
This logic has already been modified a few times and I suspect it was one of the previous changes that introduced the mechanic walking to a vehicle on the track glitch. Each time it becomes slightly harder to understand. 

What should happen is:

- Inspecting mechanics should clear any breakdowns (including pending).
- Rides can't call a mechanic if there is mechanic fixing or inspecting
- Inspecting or fixing mechanics should fall back to basic repair animation if ride hasn't broken down. But it should still fully clear all broken down bits after finishing repair animation.

src/openrct2/ride/Ride.cpp
w-flo: I will reply to this even though I've already seen your comments below, just to explain why I made the change on this line: It is possible to reproduce #8145 using the "force breakdown" feature without this change. While a mechanic is heading for an inspection, force an ideally long-lasting (i.e. train stuck somewhere on track) pending breakdown and then pick up / put down the mechanic. The inspection logic will try to call them again for the inspection through this function, but since there is a pending breakdown now, it will be a "respond to breakdown call" even though the breakdown is still pending. The mechanic will then walk to the vehicle that has the pending breakdown even though it's not in the station.

So I tried to fix this bug by making sure the mechanic would not get into the "fixing" state unless the ride has already broken down. So while a breakdown is pending and not visible to the player, I try to keep them in the inspection-related states (HEADING instead of ANSWERING). This function was one of the two places I identified that would put a mechanic into a fixing-related state "too early", and my change should prevent that from happening.

Obviously there are other ways to prevent this from happening, like handling the "fixing" process in the station differently if the breakdown is still pending when the mechanic arrives. It would be interesting to see how vanilla RCT2 handles this, but booting Windows is a hassle and it has no "force breakdown" feature AFAIK so I'm not really sure how it could be tested.

w-flo:
I agree it's not easy to understand the mechanics logic and I'm afraid every little change has a relatively high probability of introducing new bugs. It's quite complex and it's not clear which behavior is preferable when special cases happen (or which special cases are even possible).

> Inspecting mechanics should clear any breakdowns (including pending).

I implemented this in the PR, but I'm not too happy about the code style there. I didn't want to re-format the whole function and increase the chance of breaking things in the process. Basically the "return true" at the end of the "inspection" code path in UpdateFixingFinishFixOrInspect is skipped if there is a pending breakdown. So in that case, the "fixing" code path is executed in addition to the "inspection" code path, so the pending breakdown gets fixed even after an inspection.

> Rides can't call a mechanic if there is mechanic fixing or inspecting

This should already be the case with or without my PR (assuming there are no undiscovered bugs). Well, almost: My PR makes it so the ride will call the mechanic heading for an inspection if there is a breakdown, so that the mechanic changes to "answering breakdown call" instead of "heading for an inspection".

> Rides can call a mechanic whilst the breakdown is pending in the hope it gets fixed before the breakdown fully manifests.

I don't think this is implemented. I understood the "pending breakdown" flag to be an RCT2 implementation detail, not something that would actually be visible to staff or guests. Basically "randomness has decided that this ride should break down, and it should be a vehicle malfunction of vehicle V, but since that vehicle is most likely not in the station right now, I'll simply set this flag so the actual breakdown happens once vehicle V is in the station again". And in the case of safety cut out, I thought the pending flag was used to stop lifts, and a few seconds later the actual breakdown flag is set anyway, so that short delay from pending to breakdown would be the time it takes for staff to notice the breakdown and notify the mechanic.

If consensus is that pending breakdowns should actually be something that would be visible to staff or guests ("oh, look, that vehicle looks bad, better call a mechanic before it breaks!") I think we'd need some code changes. I believe right now, a mechanic is only called once the ride actually breaks down, unless the mechanic is already heading there for an inspection. Maybe this would be the right place for the change, but maybe it would add all sorts of new issues if the pending state was added to that if:
https://github.com/OpenRCT2/OpenRCT2/blob/c5b66c029483a8344889bbadf24f6ed196262dbf/src/openrct2/ride/Ride.cpp#L2753-L2757

So the question is which of these two "breakdown pending" interpretations is the one to choose. I actually like my own interpretation of the flag, and the fact that mechanics are only called once the actual breakdown happens seems to support it, but I don't have a strong opinion on it. So if you prefer the "mechanics should treat a pending breakdown exactly like a breakdown, the only difference is that the ride is still up and running and we possibly need a different fixing animation because the vehicle is not in the station" (if I understand correctly) interpretation, we should drop my fix for #8145 completely because it's plain wrong in that case.

> Inspecting or fixing mechanics should fall back to basic repair animation if ride hasn't broken down. But it should still fully clear all broken down bits after finishing repair animation.

This sounds like it could be another approach to fix #8145 if consensus is that mechanics are supposed to go and fix rides before they actually break down. Just use a different "sub_state_sequence_mask" for the vehicle breakdowns if the ride has not broken down yet.

I will send a PR for #8507 in isolation (probably tomorrow).

Gymnasiast:
We merged #8507 today. That leaves the mechanic issue.

Perhaps we should take the fixes in steps. And we should make sure that the code is properly documented, at least through comments.

AaronVanGeffen:
Fwiw, I just tested this patch. Sadly, it doesn't fix the issue I was having, #11210.

tupaschoal:
Seems like what remains here doesn't fix the issue, so I'm closing this PR. Please reopen if you can find a solution

PR #9184, opened 2019-05-01
Switch back to Windows 8.1 SDK

IntelOrca:
This should increase compatibility across different versions of Visual Studio. 10.0.14393.0 is also no longer available through the Visual Studio installer

### DIFF
--- a/openrct2.common.props
+++ b/openrct2.common.props
@@ -9,7 +9,7 @@
   <PropertyGroup>
     <!-- Allow any version of VS and Windows SDK -->
     <PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>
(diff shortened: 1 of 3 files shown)


ZehMatt:
I think the current CI errors can be fixed by defining WIN32_LEAN_AND_MEAN before including windows.h

IntelOrca:
@ZehMatt There are a couple of files which require definitions from windows.h that are excluded when using `WIN32_LEAN_AND_MEAN`.

The 8.1 SDK is not updated to handle the stricter compiler flags and there is no way of disabling the warnings just for the header in MSVC.

See https://developercommunity.visualstudio.com/content/problem/185399/error-c2760-in-combaseapih-with-windows-sdk-81-and.html

PR #9186, opened 2019-05-01
Fix track and screenshot rendering for OpenGL drawing engine

IntelOrca:


ZehMatt:
Interesting approach, I once thought about doing this but held my self back by wanting to have multiple render targets instead.

PR #9196, opened 2019-05-03
Refactor Peep Generate

nexgenration:
Move peep_generate into the `Peep` struct as a static member function.

### DIFF
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -1708,7 +1708,7 @@ static constexpr const uint8_t tshirt_colours[] = {
  *
  *  rct2: 0x0069A05D
  */
-Peep* peep_generate(int32_t x, int32_t y, int32_t z)
+Peep* Peep::Generate(int32_t x, int32_t y, int32_t z)
 {
     if (gSpriteListCount[SPRITE_LIST_NULL] < 400)
         return nullptr;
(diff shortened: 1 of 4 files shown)


src/openrct2/peep/Peep.cpp
```diff
@@ -1708,7 +1708,7 @@ static constexpr const uint8_t tshirt_colours[] = {
  *
  *  rct2: 0x0069A05D
  */
-Peep* peep_generate(int32_t x, int32_t y, int32_t z)
+Peep* Peep::Generate(int32_t x, int32_t y, int32_t z)
 {
     if (gSpriteListCount[SPRITE_LIST_NULL] < 400)
         return nullptr;
```
duncanspumpkin: Could you modify in this function `nausea_tolerance` , `happiness_delta`, `hunger_delta`, `thirst_delta`, `tshirt_colour`, `trouser_colour` to camelCase.

src/openrct2/peep/Peep.cpp
```diff
@@ -1708,7 +1708,7 @@ static constexpr const uint8_t tshirt_colours[] = {
  *
  *  rct2: 0x0069A05D
  */
-Peep* peep_generate(int32_t x, int32_t y, int32_t z)
+Peep* Peep::Generate(int32_t x, int32_t y, int32_t z)
```
duncanspumpkin: For the location the appropriate struct for this one would be `CoordsXYZ`. This is because it uses big coordinates.

duncanspumpkin:
Since this is your first commit to the project you should also add yourself to contributors.md under Bug Fixes.

PR #9206, opened 2019-05-04
Fix 9204: Making screenshots from command-line is broken

IntelOrca:


test/tests/ImageImporterTests.cpp
```diff
@@ -57,3 +67,38 @@ TEST_F(ImageImporterTests, Import_Logo)
...
+
+TEST_F(ImageImporterTests, Screenshot)
+{
+    auto parkPath = TestData::GetParkPath("bpb.sv6");
+    auto pngPath = OutputImagePath;
+
+    // Dump screenshot
+    ScreenshotOptions options;
+    const char* argv[] = { parkPath.c_str(), pngPath.c_str(), "800", "600" };
+    auto exitCode = cmdline_for_screenshot(argv, std::size(argv), &options);
```
Broxzier:
```suggestion
    auto exitCode = cmdline_for_screenshot(argv, (int32_t)std::size(argv), &options);
```

test/tests/ImageImporterTests.cpp
```diff
@@ -57,3 +67,38 @@ TEST_F(ImageImporterTests, Import_Logo)
...
+TEST_F(ImageImporterTests, GiantScreenshot)
+{
+    auto parkPath = TestData::GetParkPath("bpb.sv6");
+    auto pngPath = OutputImagePath;
+
+    // Dump screenshot
+    ScreenshotOptions options;
+    options.weather = 1 + WEATHER_SUNNY;
+    const char* argv[] = { parkPath.c_str(), pngPath.c_str(), "giant", "0", "0" };
+    auto exitCode = cmdline_for_screenshot(argv, std::size(argv), &options);
```
Broxzier:
```suggestion
    auto exitCode = cmdline_for_screenshot(argv, (int32_t)std::size(argv), &options);
```

Broxzier:
The calculated hashes in the tests are different for me. I think settings like tile smoothing, gridlines and uppercase banner text are used from the global `config.ini` file, because I got another different hash when changing some of these settings.

IntelOrca:
I have had to remove the tests for now, they first aren't going to run on the CI anyway and I didn't find an easy way of resetting the config as the OpenRCT2 context is not created within the test.

Issue #2339, opened 2015-11-18
Broadcast server to LAN


IntelOrca:
Source:
https://www.reddit.com/r/openrct2/comments/3ta95l/could_someone_tell_me_how_to_run_multiplayer/

A lot of old games used to broadcast the local address of the host on the local area network. Server lists would automatically listen to broadcast messages and show LAN servers.

~~I believe this can only be done for IPv4.~~
IPv4: Broadcast protocol
IPv6: Multicast protocol

Smeat:
I don't really know much about networking, but I think for IPv6 you could use the FF02::1 multicast address.

https://www.iana.org/assignments/ipv6-multicast-addresses/ipv6-multicast-addresses.xhtml

PFCKrutonium:
Minecraft Does this, and it rarely works as well as expected. We need to make sure OpenRCT2 punches temp holes in the firewall on all computers.

samdroid-apps:
One way that I've seen things do it on linux is using Avahai (mDNS).  According to the avahai zeroconf viewer tool, there are both chat and chromecasts broadcasting their presence on the lan right now.

I've only ever use avahai way up the abstraction layer (via telepathy salut, where I just asked it for a list of other people on the network).  I also don't know if mDNS is supported on windows, but just some food for thought.

janisozaur:
I's called "avahi", not "avahai" (just leaving a comment so it's easier to find this issue).

janisozaur:
Should this be closed now?

PR #9212, opened 2019-05-05
Implement searching of local servers

IntelOrca:
Implements #2339, for IPv4.

When fetching servers, a UDP packet is broadcast to all broadcast addresses on the LAN. Any servers on the LAN will reply with server details. This should make it very simple for anyone playing on the same network to join each other.


src/openrct2/network/ServerList.cpp
```diff
@@ -80,3 +233,170 @@ bool server_list_write(const std::vector<server_entry>& entries)
         return false;
     }
 }
+
+std::future<std::vector<ServerListEntry>> ServerList::FetchLocalServerListAsync(const INetworkEndpoint& broadcastEndpoint)
+{
+    auto broadcastAddress = broadcastEndpoint.GetHostname();
+    return std::async([broadcastAddress] {
```
ZehMatt: Perhaps use std::launch::async

src/openrct2/network/ServerList.cpp
```diff
@@ -80,3 +233,170 @@ bool server_list_write(const std::vector<server_entry>& entries)
...
+            }
+            platform_sleep(RECV_DELAY_MS);
+        }
+        return entries;
+    });
+}
+
+std::future<std::vector<ServerListEntry>> ServerList::FetchLocalServerListAsync()
+{
+    return std::async([&] {
```
ZehMatt: Also here std::launch::async

src/openrct2/network/ServerList.cpp
IntelOrca: Is there any difference? I couldn't see one during testing.

src/openrct2/network/Socket.h
```diff
@@ -31,6 +32,18 @@ enum NETWORK_READPACKET
     NETWORK_READPACKET_DISCONNECTED
 };
 
+/**
+ * Represents an address and port.
+ */
+interface INetworkEndpoint
+{
+    virtual ~INetworkEndpoint()
```
ZehMatt:
```suggestion
    virtual ~INetworkEndpoint() = default;
```

src/openrct2/network/Socket.h
```diff
@@ -59,10 +72,36 @@ interface ITcpSocket
...
 };
 
-std::unique_ptr<ITcpSocket> CreateTcpSocket();
+/**
+ * Represents a UDP socket / listener.
+ */
+interface IUdpSocket
+{
+public:
+    virtual ~IUdpSocket()
```
ZehMatt:
```suggestion
    virtual ~IUdpSocket() = default;
```

src/openrct2/core/Json.hpp
```diff
@@ -23,7 +24,7 @@ namespace Json
     json_t* ReadFromFile(const utf8* path, size_t maxSize = MAX_JSON_SIZE);
     void WriteToFile(const utf8* path, const json_t* json, size_t flags = 0);
 
-    json_t* FromString(const std::string& raw);
+    json_t* FromString(const std::string_view& raw);
```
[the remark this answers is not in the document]
IntelOrca: I don't understand.

janisozaur:
I'd like to test and review it

src/openrct2/network/Socket.cpp
```diff
@@ -532,6 +808,110 @@ void DisposeWSA()
...
+    int sock = socket(AF_INET, SOCK_DGRAM, 0);
+    if (sock == -1)
+    {
+        return {};
+    }
+
+    DWORD len = 0;
+    size_t capacity = 2;
+    std::vector<INTERFACE_INFO> interfaces;
+    for (;;)
```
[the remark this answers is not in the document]
IntelOrca: resolved

src/openrct2/network/Socket.cpp
```diff
@@ -532,6 +808,110 @@ void DisposeWSA()
...
+static std::vector<INTERFACE_INFO> GetNetworkInterfaces()
+{
+    int sock = socket(AF_INET, SOCK_DGRAM, 0);
+    if (sock == -1)
+    {
+        return {};
+    }
+
+    DWORD len = 0;
+    size_t capacity = 2;
```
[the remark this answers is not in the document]
IntelOrca: resolved

src/openrct2/network/ServerList.cpp
```diff
@@ -80,3 +233,170 @@ bool server_list_write(const std::vector<server_entry>& entries)
...
+            }
+        }
+        return mergedEntries;
+    });
+}
+
+std::future<std::vector<ServerListEntry>> ServerList::FetchOnlineServerListAsync()
+{
+#    ifdef DISABLE_HTTP
+    return std::async(std::launch::deferred, [] { return std::vector<ServerListEntry>(); });
```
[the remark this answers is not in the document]
IntelOrca: `std::launch::deferred` shouldn't spawn a thread... but I have changed it anyway. The `get()`s are already wrapped in try...catches.

src/openrct2/network/ServerList.cpp
```diff
@@ -80,3 +233,170 @@ bool server_list_write(const std::vector<server_entry>& entries)
...
+            throw std::runtime_error("Unable to broadcast server query.");
+        }
+
+        std::vector<ServerListEntry> entries;
+        char buffer[1024]{};
+        size_t recievedLen{};
+        std::unique_ptr<INetworkEndpoint> endpoint;
+        for (int i = 0; i < (RECV_WAIT_MS / RECV_DELAY_MS); i++)
+        {
+            auto p = udpSocket->ReceiveData(buffer, sizeof(buffer) - 1, &recievedLen, &endpoint);
```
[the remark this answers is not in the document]
IntelOrca: Buffer is now always initialised to zeros prior to receive data leaving the remaining byte always 0 in case received data is not null terminated or too big. Invalid JSON is now handled by try...catch.

src/openrct2/network/Socket.h
IntelOrca: Resolved

src/openrct2/network/Socket.h
IntelOrca: Resolved

src/openrct2/network/Network.cpp
```diff
@@ -3994,6 +3998,11 @@ NetworkStats_t network_get_stats()
     return gNetwork.GetStats();
 }
 
+json_t* network_get_server_info_as_json()
```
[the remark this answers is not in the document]
IntelOrca: Some of the data can change, e.g. players. I have to leave it like this.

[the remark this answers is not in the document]
ZehMatt: What I meant is that you can just feed the updated data to the advertiser in the Network::Update call

src/openrct2/core/Json.hpp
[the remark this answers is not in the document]
ZehMatt: Not sure what you would gain passing a reference to a borrow object

src/openrct2/network/Socket.cpp
```diff
@@ -446,60 +562,220 @@ class TcpSocket final : public ITcpSocket
...
+        }
+        NetworkEndpoint endpoint((const sockaddr*)&ss, ss_len);
+        return SendData(endpoint, buffer, size);
+    }
+
+    size_t SendData(const INetworkEndpoint& destination, const void* buffer, size_t size) override
+    {
+        if (_socket == INVALID_SOCKET)
+        {
+            _socket = socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP);
```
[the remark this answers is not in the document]
IntelOrca: resolved

src/openrct2/network/Network.cpp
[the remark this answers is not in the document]
IntelOrca: I will postpone this. GetHeartbeatJson already calls into the network class via `network_get_num_players`. I will do what you suggest in another PR using a new struct that holds all the server status info.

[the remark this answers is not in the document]
ZehMatt: Alrite

src/openrct2/network/ServerList.cpp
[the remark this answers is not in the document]
ZehMatt: Sorry I actually misread the documentation, it does indeed not spawn a new thread, what I've misread is that it can be used by another thread and then invokes the function.

ZehMatt:
I'll try to get this tested tomorrow in a real environment, I think code wise it looks okay.

src/openrct2/network/Socket.cpp
```diff
@@ -77,7 +77,123 @@ class SocketException : public std::runtime_error
...
+        }
+        else
+        {
+            return ((sockaddr_in6*)&_address)->sin6_port;
+        }
+    }
+
+    std::string GetHostname() const override
+    {
+        char hostname[256];
```
ZehMatt:
```suggestion
        char hostname[256]{};
```
Not sure if thats even possible but better better safe than sorry.

ZehMatt:
It somehow ends up showing the server twice on my end
openrct2_2019-05-11_13-46-33578160ca1c47316a4

IntelOrca:
@ZehMatt The server list should be removing any servers that have identical addresses. Can you check what the addresses are for those two duplicate servers?

`ServerList::Sort()` should be removing the duplicates, you could try debugging that.

ZehMatt:
Looks like its really two different ips per subnet, I guess this is very specific to my network setup and not sure how one would resolve this. Its probably fine for the other 99.9%

ZehMatt:
Also I'm okay with it, @janisozaur wanted to review it aswell.

src/openrct2/network/Socket.cpp
```diff
@@ -446,60 +562,211 @@ class TcpSocket final : public ITcpSocket
...
+        }
+
+        // Enable send and receiving of broadcast messages
+        if (!SetOption(sock, SOL_SOCKET, SO_BROADCAST, true))
+        {
+            log_warning("SO_BROADCAST failed. %d", LAST_SOCKET_ERROR());
+        }
+
+        // Turn off IPV6_V6ONLY so we can accept both v4 and v6 connections
+        if (!SetOption(_socket, IPPROTO_IPV6, IPV6_V6ONLY, false))
```
janisozaur:
```suggestion
        if (!SetOption(sock, IPPROTO_IPV6, IPV6_V6ONLY, false))
```

src/openrct2/network/Socket.cpp
```diff
@@ -446,60 +562,211 @@ class TcpSocket final : public ITcpSocket
...
+        {
+            log_warning("IPV6_V6ONLY failed. %d", LAST_SOCKET_ERROR());
+        }
+
+        if (!SetOption(_socket, SOL_SOCKET, SO_REUSEADDR, true))
```
janisozaur:
```suggestion
        if (!SetOption(sock, SOL_SOCKET, SO_REUSEADDR, true))
```

src/openrct2/network/Socket.cpp
```diff
@@ -446,60 +562,211 @@ class TcpSocket final : public ITcpSocket
...
 
-        addrinfo hints = {};
-        hints.ai_family = AF_UNSPEC;
-        if (address.empty())
+    SOCKET_STATUS GetStatus() override
+    {
+        return _status;
+    }
+
+    const char* GetError() override
```
janisozaur:
```suggestion
    const char* GetError() const override
```

src/openrct2/network/Socket.cpp
```diff
@@ -446,60 +562,211 @@ class TcpSocket final : public ITcpSocket
...
+    ~UdpSocket() override
     {
-        std::string serviceName = std::to_string(port);
+        CloseSocket();
+    }
 
-        addrinfo hints = {};
-        hints.ai_family = AF_UNSPEC;
-        if (address.empty())
+    SOCKET_STATUS GetStatus() override
```
janisozaur:
```suggestion
    SOCKET_STATUS GetStatus() const override
```

src/openrct2/network/ServerList.cpp
```diff
@@ -7,22 +7,155 @@
...
+    _serverEntries.insert(_serverEntries.end(), entries.begin(), entries.end());
+    Sort();
+}
+
+void ServerList::Clear()
+{
+    _serverEntries.clear();
+}
+
+std::vector<ServerListEntry> ServerList::ReadFavourites()
```
janisozaur: looks like a const method to me

IntelOrca:
Local servers are now green.

Issue #485, opened 2014-09-24
Add ghost train for testing while building mode (as seen in RCT3)


AngryHobo:
I was thinking about different functions or enhancements you could make to the game so I thought, how about adding a testing during build mode like in RCT3 that shows a ghost train on the coaster just to make sure the ride can get over all the hills and loops. I know we've all had that time where we're finished building a perfect coaster except its not able to make it over all the obstacles.

Gymnasiast:
Has been implemented recently, so this can be closed.

IntelOrca:
Well... not to the same extent as RCT3. Once vehicle update is done, it might be possible to add the ghost vehicle.

mrtnptrs:
Vehicle update is done now, so this would be possible now.

Gymnasiast:
I doubt that. Drawing of most roller coasters and vehicles is still not implemented.

devnoname120:
Can this be done now?

janisozaur:
Probably, but someone has to do it and we are more concerned with getting release out the door.

PFCKrutonium:
Basically, ask again when the release is out the door.

IntelOrca:
Not possible until we have our own save format. It would need to be a vehicle with specific flag - a new flag which would be incompatible with RCT2.

JohnEriksson00:
Does it really require a new save format? Isn't it possible to implement a train that continously rides the track (like testing does currently, although with the construction window closed) when you have the construction window open? You don't need to save this train, as the window closes (therefore removing the "ghost" train). Preventing crashes in testing mode is already in the game.

duncanspumpkin:
Feel free to implement it but it would be a pain to do in the current code.

PR #9222, opened 2019-05-07
Implement #485: ghost train simulation

IntelOrca:
I have attempted to replicate the behaviour from RCT3 as much as possible.


src/openrct2/paint/Paint.h
```diff
@@ -182,7 +182,7 @@ extern LocationXY8 gClipSelectionA;
 extern LocationXY8 gClipSelectionB;
 
 /** rct2: 0x00993CC4. The white ghost that indicates not-yet-built elements. */
-#define CONSTRUCTION_MARKER (COLOUR_DARK_GREEN << 19 | COLOUR_GREY << 24 | IMAGE_TYPE_REMAP);
+#define CONSTRUCTION_MARKER (COLOUR_DARK_GREEN << 19 | COLOUR_GREY << 24 | IMAGE_TYPE_REMAP)
```
AaronVanGeffen: Nice catch!

IntelOrca: I got compile error because of it 😄

Umdlye:
I'm not sure if the "test unfinished track" option is entirely obsolete, I believe people would still use it for spawning trains for decorative purposes (transfer tracks and such).

It'd be cool if the simulation mode didn't require an entrance and exit to be built since that's usually done at the very end of construction.

Gymnasiast:
> I'm not sure if the "test unfinished track" option is entirely obsolete, I believe people would still use it for spawning trains for decorative purposes (transfer tracks and such).

For that, rotating lift mode is commonly used, which continues to work. For other purposes, powered launch should probably do.

ocalhoun6:
>For other purposes, powered launch should probably do.

As long as it's possible to set the powered launch speed very low, even just 1 or 2 mph.

ZehMatt:
Not even sure if that is a valid argument to use some feature for different purposes than its meant to. Being able to place vehicles anywhere without having a track/finished ride should be its own feature, eventually plugins can do that at some point.

Gymnasiast:
Right, I have just tried this out.

The first thing probably sounds petty, but here goes: I really don't like how this looks when using the "traffic light" buttons.

The second thing is: what place will this have in the build-test-open workflow? Right now, you can let trains finish the entire track in simulate mode and it will update the statistics. That essentially makes the game easier: after all, you now don't have to be on the ball for crashing trains when testing. I suppose the exact mechanics of how this feature works and what it can and can't do need a bit of discussion.

Nevertheless, it's a very nice achievement and the points I raised are things that should not be hard to address.

IntelOrca:
I don't have any suggestions of my own, I would probably just go with what RCT3 does, but I can't remember how it works.

Let me know what changes to make.

Gymnasiast:
For that, I'll have to check what RCT3 does. That will hopefully be tomorrow.

AaronVanGeffen:
Perhaps we could add a blue flag and traffic light for the simulation mode, as a sort-of in-between mode? I'd be happy to conjure up some icons if you like.

Gymnasiast:
How about the following:
- When the track is not a complete circuit or the entrance/exit hasn't been built, the test button spawns a ghost train.
- Otherwise, the test button works in the normal way.

That would address both points for me.

IntelOrca:
If you are building a track that uses launched operation mode then you might want to keep building more track while it repeats simulations of launches. Basically it would come down to the mode being dependant on ride / entrance existence alone which doesn't seem particularly intuitive.

How about we keep it as it is but stop simulation from producing test result data?

Broxzier:
Ah cool, this is something I was also working on, but never finished it or pushed it to GitHub (WIP PR's are nice now). If I remember correctly, RCT3 allowed you to see how the trains would move even while the game was paused. In OpenRCT2 we could take the speed multiplier even into account while the game is paused. Here's a video of how I implemented it for OpenRCT2 at the time: https://cdn.discordapp.com/attachments/356546837945253888/527991864612093965/2018-12-28_00-26-47-1.mp4

https://github.com/Broxzier/OpenRCT2/commit/fc8355b1fa1e6166d3ef250caa3439ea11b31524
To perhaps take some inspiration from.

Gymnasiast:
@Broxzier Love the Raveleijn theme! :P

But besides that, what I find most interesting about the video is that you put a test button in the construction window. I think that's a brilliant idea: that's the place that most people would want to use it. And it avoids messing up the traffic light UI.

If we combine that with the ghost trains not yielding results, then all my points are solved, and in a very neat way as well.

sta5is:
Wonderful idea! Can this be expanded, as an option, where the train is simulated as if it is running with full capacity since the train goes faster with a full train? Possibly as a drop down option when clicking the yellow flag in the construction menu?

ZehMatt:
I also like the button in the construction window, I think its better than having a new test mode.

AaronVanGeffen:
Having the simulate button in the construction window looks like a huge improvement. I have two suggestions, though.

Currently, the construction window looks like this:
Screenshot_20190523_203754

The flag looks rather generic; it is not immediately clear what it does — it indicates the ride is closed, yes, but what happens when we click it?

It turns into a different icon:
Screenshot_20190523_203941

My first suggestion is to let the button's 'pressed state' indicate the simulation mode is active:
Screenshot_20190523_204017

My second suggestion is to change the flag to a more neutral colour. Currently, the entire icon is recoloured, including the stick. I think a shade of blue might be a nice colour for simulation mode. I'll draft up an icon.

ZehMatt:
Also I wonder how it would look like if the button would sit above the demolish button rather than being an lonely outsider on the right

Broxzier:
Perhaps it'll fit better with the Entrance/Exit groupbox, since those only appear when a station exists, and a station is required for a ride to be even testable.

mockups

Another thing that RCT3 has (though I may misremember), is that trains would also simulate while the game is paused.

Gymnasiast:
@AaronVanGeffen I like the blue colour. It also looks a bit ghosty ;-)

IntelOrca:
@AaronVanGeffen Your icon uses colours from the water palette (colours that change each tick), would you be able to update the image?

AaronVanGeffen:
@IntelOrca Ah, I hadn't thought about that. Sure, I'll swap the offending colours.

AaronVanGeffen:
Changed the icon. It's a bit less ghostly now, as @Gymnasiast put it, but still reasonably blue:

Screenshot 2019-05-25 at 11 54 34

AaronVanGeffen:
This PR has turned out quite nicely. While the button layout in the construction window is perhaps suboptimal, I think we should probably tackle that in a follow-up PR.

One thing I noticed is that, while you can no longer select it, the simulate flag *is* still used in the RCT2-style ride window:
Screenshot_20190525_130606

The RCT1-style window, however, just shows the ride as closed:
Screenshot_20190525_130701

I'm not sure what would be best. Perhaps the RCT1 style could default to 'testing' as well, showing a blue shade on the yellow light? Again, I'd be happy to provide sprites.

AaronVanGeffen:
Added the relevant sprites: rct1_simulate_off_pressed rct1_simulate_off rct1_simulate_on_pressed rct1_simulate_on

Screenshot 2019-05-25 at 14 11 58

ZehMatt:
Finally got around to test this, very cool feature. There is one small issue with multiplayer right now, when its in the new testing mode and I hover the entrance/exit on the stations the trains will vanish while for the other clients nothing happens and it instantly desyncs.

Gymnasiast:
@AaronVanGeffen I have been very slow in reacting, but I thought it would be cool to have a blinking orange light.

AaronVanGeffen:
Honestly, while I like the idea, I think having a blue light makes it more obvious that something else entirely is going on. It also makes it more consistent with the blue flag used elsewhere.

A recent merger has already caused this PR to require a rebase; let's not let it rot. We can optimise things in later follow-up PRs if needed.

Gymnasiast:
@AaronVanGeffen I still prefer the blinking, but I agree on not letting the PR rot. Let's go with the blue, at least for now.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2055,12 +2062,32 @@ static void window_ride_main_mouseup(rct_window* w, rct_widgetindex widgetIndex)
...
     if (theme_get_flags() & UITHEME_FLAG_USE_LIGHTS_RIDE)
-        minHeight = 200 + offset + RCT1_LIGHT_OFFSET
-            - (ride_type_has_flag(get_ride(w->number)->type, RIDE_TYPE_FLAG_NO_TEST_MODE) ? 14 : 0);
+    {
+        minHeight += 20 + RCT1_LIGHT_OFFSET;
+
+        auto ride = get_ride(w->number);
+        if (ride != nullptr)
+        {
+#ifdef __SIMULATE_IN_RIDE_WINDOW__
```
Gymnasiast: I might have missed something, but why is this behind an ifdef?

AaronVanGeffen:
@ZehMatt With the latest commit in, has the multiplayer issue been resolved?

src/openrct2-ui/windows/Ride.cpp
AaronVanGeffen: You can get four lights in the ride window with this compile-time flag.

Gymnasiast: Why? That's asking for bitrot, and it's not even useful.

IntelOrca: I want to be sure that it won't be requested some time later, it took a lot of effort to add in the first place.

ZehMatt:
I will check it later today, when I was testing this I had to change 2 game actions that would test if the ghost flag is set in order to prevent the trains from vanishing.

ZehMatt:
Still desyncs, there are two places where it needs to check if the ghost flag is set.
here:
https://github.com/OpenRCT2/OpenRCT2/blob/b17797d22fe3cdcddc9c0923dcb889b6fe51f658/src/openrct2/actions/RideEntranceExitPlaceAction.hpp#L156
and here:
https://github.com/OpenRCT2/OpenRCT2/blob/b17797d22fe3cdcddc9c0923dcb889b6fe51f658/src/openrct2/actions/RideEntranceExitRemoveAction.hpp#L126

ZehMatt:
Now the desync issue is resolved, another thing I noticed is that the graph will only update for the host while other clients don't see anything but its not an issue just something I've noticed.

AaronVanGeffen:
I think the idea was that the simulation mode shouldn't generate test results. In that case, this sounds like a bug.

ZehMatt:
There are no test results however the graph will update and resets the moment the vehicles start from the station

src/openrct2/actions/RideSetStatus.hpp
```diff
@@ -73,9 +74,16 @@ DEFINE_GAME_ACTION(RideSetStatusAction, GAME_COMMAND_SET_RIDE_STATUS, GameAction
...
             {
-                if (!ride_is_valid_for_test(ride, _status == RIDE_STATUS_OPEN, 0))
+                // Simulating will force clear the track, so make sure player can't cheat around a break down
+                res->Error = GA_ERROR::DISALLOWED;
+                res->ErrorMessage = STR_HAS_BROKEN_DOWN_AND_REQUIRES_FIXING;
+                return res;
+            }
+            else if (_status == RIDE_STATUS_TESTING || _status == RIDE_STATUS_SIMULATING)
+            {
+                if (!ride_is_valid_for_test(ride, _status, 0))
```
ZehMatt: Might be a good time to turn the last parameter into an actual bool since this PR touches both lines, its not used anywhere else.

src/openrct2/ride/thrill/LaunchedFreefall.cpp
```diff
@@ -35,48 +35,49 @@ void vehicle_visual_launched_freefall(
...
-    if (session->DPI.zoom_level < 2)
+    if (session->DPI.zoom_level < 2 && vehicle->num_peeps > 0 && !vehicle->IsGhost())
     {
-        if (vehicle->num_peeps > 0)
+        baseImage_id = vehicleEntry->base_image_id + 9;
+        if ((vehicle->restraints_position / 64) == 3)
+        {
+            baseImage_id += 2; // Draw peeps sitting without transparent area between them for restraints
+        }
+        image_id = (baseImage_id + ((((imageDirection / 8) + 0) & 3) * 3))
```
ZehMatt: Might be a good idea to divide imageDirection once as every statement below is doing it.

src/openrct2-ui/windows/Ride.cpp
Gymnasiast: If it gets requested, I will veto it anyway.

I'm ok with merging the PR with this intact (too important a feature to get bogged by this), but I do want to remove it at a later point.

AaronVanGeffen: If we end up removing it in a separate commit, it can easily be reverted. I'm not too fussed either way.

ZehMatt:
Found some funny things during my tests



Also it seems that when you place an entrance or exit during the simulation mode it will stop the simulation but the mode is still on while nothing happens.

Gymnasiast:
The observation tower is a known bug/limitation. It also happens in vanilla.

ZehMatt:
Even the first version where the doors open mid air? 😄

Gymnasiast:
Hm, no. ;)
But the one where is vibrates is vanilla behaviour. I would say it is better left to another PR.

AaronVanGeffen:
So, that means this is ready to be merged, yes?

KableK42:
This update broke this calculator https://imgur.com/a/PPaUQKi

Since this relies on one of the coasters crashing when in this new "blue flag" test mode it just resets both coasters when that happens. Would be nice to have some advanced option to enable the old functionality.

KableK42:
This update broke this calculator https://imgur.com/a/PPaUQKi

Since this relies on one of the coasters crashing when in this new "blue flag" test mode it just resets both coasters when that happens. Would be nice to have some advanced option to enable the old functionality.

AaronVanGeffen:
@KableK42 Please file a separate issue for this. Frankly, though, this is a fairly extreme use-case; I can't say we'll support it.

KableK42:
> @KableK42 Please file a separate issue for this. Frankly, though, this is a fairly extreme use-case; I can't say we'll support it.

I understand :)
https://xkcd.com/1172/

Issue #9198, opened 2019-05-03
Refactor vehicle_invalidate


duncanspumpkin:
`vehicle_invalidate` should be refactored to become a member function of the struct. This member function should be for the `rct_vehicle` struct. It will need to be a public function as it will be called from many non vehicle struct functions.
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Vehicle.cpp#L842 

The updated function should follow the coding style https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style Replace all variables with lower camelCase. Function names should be upper CamelCase.

Commit message should follow the commit message standard https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

PR #9226, opened 2019-05-10
Vehicle invalidate

nexgenration:
Fix issue #9198, (Refactor vehicle_invalidate).

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -1394,7 +1394,7 @@ static bool vehicle_close_restraints(rct_vehicle* vehicle)
                 continue;
             }
         }
-        vehicle->Invalidate();//updated function call to match becoming a member function of rct_vehicle struct//updated function call to match becoming a member function of rct_vehicle struct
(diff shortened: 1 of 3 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -1392,7 +1394,7 @@ static bool vehicle_close_restraints(rct_vehicle* vehicle)
                 continue;
             }
         }
-        vehicle_invalidate(vehicle);
+        vehicle->Invalidate();//updated function call to match becoming a member function of rct_vehicle struct
```
AaronVanGeffen: Please remove this comment and the others like it.

AaronVanGeffen:
Thanks for your PR.

Please remove the comments describing your changes. We can see what's changed in the commit diffs — the comments just clutter the code base.

src/openrct2/ride/Vehicle.cpp
nexgenration: oops my bad. i actually meant to take those out but forgot

Issue #9197, opened 2019-05-03
Refactor `peep_insert_new_thought`


duncanspumpkin:
`peep_insert_new_thought` should be refactored to become a member function of the struct. This member function should be for the ~~Guest~~ Peep struct. ~~It will therefore need to be moved from Peep.cpp to Guest.cpp with the relevant changes made to the struct~~. It will need to be a public function as peep thoughts are inserted from a number of places. 
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/peep/Peep.cpp#L2239
_Edit: Function can't be moved to the Guest struct yet. Leave within the Peep struct with a note that it is only for Guests_

The updated function should follow the coding style https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style Replace all variables with lower camelCase. Function names should be upper CamelCase. 

Commit message should follow the commit message standard https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

PR #9230, opened 2019-05-11
Peep insert new thought

nexgenration:
Fix issue #9197, (Refactor peep_insert_new_thought).


src/openrct2/peep/Peep.h
```diff
@@ -736,6 +736,13 @@ struct Peep : rct_sprite_common
 struct Guest : Peep
 {
 public:
+    /**
+     * rct2: 0x699F5A
+     * al:thought_type
+     * ah:thought_arguments
+     * esi: peep
+     */
+    void InsertNewThought(PeepThoughtType thought_type, uint8_t thought_arguments);
```
AaronVanGeffen: Ideally, this should be declared here, but not all instances of Peep are using the Guest struct yet. It should therefure be declared on the Peep struct, still.

The comment can be removed, too. It is only relevant for the implementation (cpp) file, not the header file.

src/openrct2/peep/Guest.cpp
```diff
@@ -5103,6 +5103,55 @@ void Guest::UpdateRide()
...
 static bool peep_find_ride_to_look_at(Peep* peep, uint8_t edge, uint8_t* rideToView, uint8_t* rideSeatToView);
 
+/**
+ *
+ *  rct2: 0x699F5A
+ * al:thoughtType
+ * ah:thoughtArguments
+ * esi: peep
+ */
+void Guest::InsertNewThought(PeepThoughtType thoughtType, uint8_t thoughtArguments)
```
AaronVanGeffen: Should be 
```suggestion
void Peep::InsertNewThought(PeepThoughtType thoughtType, uint8_t thoughtArguments)
```

src/openrct2/peep/Peep.h
duncanspumpkin: Yes sorry that was my mistake. I thought it was ready to go in Guest but no it must still be left inside of Peep.

src/openrct2/peep/Guest.cpp
```diff
@@ -382,7 +382,7 @@ void Guest::Tick128UpdateGuest(int32_t index)
             PeepThoughtType thought_type = crowded_thoughts[scenario_rand() & 0xF];
             if (thought_type != PEEP_THOUGHT_TYPE_NONE)
             {
-                peep_insert_new_thought(this, thought_type, PEEP_THOUGHT_ITEM_NONE);
+                this->InsertNewThought(thought_type, PEEP_THOUGHT_ITEM_NONE);
```
duncanspumpkin: You don't need to do `this->` on a member function when you are already calling it from within a member function.

duncanspumpkin:
Please try to follow the commit message standard https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

I had to rewrite the previous PR's message to follow it.

nexgenration:
> Please try to follow the commit message standard https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages
> 
> I had to rewrite the previous PR's message to follow it.

Sorry my bad. I thought I was following those commit message guidelines but turns out i was miss reading them.

> Please remove the unnecessary `this->` and move the function back to `Peep.cpp` as detailed by AaronVanGeffan. It was a mistake on my part to move it this early.

I will get to this shortly. Sorry for the delay.

src/openrct2/peep/Guest.cpp
```diff
@@ -443,7 +443,7 @@ void Guest::Tick128UpdateGuest(int32_t index)
 
                     if (thought_type != PEEP_THOUGHT_TYPE_NONE)
                     {
-                        peep_insert_new_thought(this, thought_type, PEEP_THOUGHT_ITEM_NONE);
+                        this->InsertNewThought(thought_type, PEEP_THOUGHT_ITEM_NONE);
```
duncanspumpkin: I meant remove all of the `this->`

src/openrct2/peep/Guest.cpp
```diff
@@ -5103,6 +5103,55 @@ void Guest::UpdateRide()
...
+/**
+ *
+ *  rct2: 0x699F5A
+ * al:thoughtType
+ * ah:thoughtArguments
+ * esi: peep
+ */
+void Peep::InsertNewThought(PeepThoughtType thoughtType, uint8_t thoughtArguments)
+{
+    PeepActionType action = PeepThoughtToActionMap[thoughtType].action;
```
duncanspumpkin: Because Peep already contains a field called `action` this is causing issues. Rename this variable to `newAction` and it should fix the compilation

src/openrct2/peep/Guest.cpp
```diff
@@ -5103,6 +5103,55 @@ void Guest::UpdateRide()
...
+    if (action != PEEP_ACTION_NONE_2 && this->action >= PEEP_ACTION_NONE_1)
+    {
+        this->action = action;
+        this->action_frame = 0;
+        this->action_sprite_image_offset = 0;
+        this->UpdateCurrentActionSpriteType();
+        this->Invalidate();
```
duncanspumpkin: Again all of these `this->` aren't required. Some people do like this style but this codebase does not do it like this.

src/openrct2/peep/Guest.cpp
```diff
@@ -5103,6 +5103,55 @@ void Guest::UpdateRide()
...
+    }
+
+    for (int32_t i = 0; i < PEEP_MAX_THOUGHTS; ++i)
+    {
+        rct_peep_thought* thought = &this->thoughts[i];
```
duncanspumpkin: Again `this->` isn't required here.

src/openrct2/peep/Guest.cpp
```diff
@@ -5103,6 +5103,55 @@ void Guest::UpdateRide()
...
+            // existing thought and placing it at the top.
+            if (i < PEEP_MAX_THOUGHTS - 2)
+            {
+                memmove(thought, thought + 1, sizeof(rct_peep_thought) * (PEEP_MAX_THOUGHTS - i - 1));
+            }
+            break;
+        }
+    }
+
+    memmove(&this->thoughts[1], &this->thoughts[0], sizeof(rct_peep_thought) * (PEEP_MAX_THOUGHTS - 1));
```
duncanspumpkin: As above. Also applies to the thoughts below.

PR #9232, opened 2019-05-11
Fix false positive desync when changing map during network play

ZehMatt:
This is because the client wont reconnect and keeps the tick data from the previous map, this PR simply clears the data whenever the map is loaded to ensure there are no colliding entries.


ZehMatt:
Its a minor regression from my end introduced by https://github.com/OpenRCT2/OpenRCT2/pull/8860

ZehMatt:
Thanks

PR #9235, opened 2019-05-11
Improve util_rand

ZehMatt:
util_rand was changed at some point to use rand() which only ever returns a maximum value with 15 bits. 
https://github.com/OpenRCT2/OpenRCT2/blob/c8f822ea70ba41deefdef2767235d7ecdd744281/src/openrct2/world/Climate.cpp#L178
Would always be zero and causes thunder sounds every 43 ticks with the same sound and lightining would never work because
https://github.com/OpenRCT2/OpenRCT2/blob/c8f822ea70ba41deefdef2767235d7ecdd744281/src/openrct2/world/Climate.cpp#L180
Would always be zero.
I also made sure that when generating a key for the advertisment it would seed it which didn't happen before.

### DIFF
--- a/src/openrct2/network/NetworkServerAdvertiser.cpp
+++ b/src/openrct2/network/NetworkServerAdvertiser.cpp
@@ -244,7 +244,7 @@ private:
         };
 
         std::random_device rd;
-        std::uniform_int_distribution<> dist(0, std::size(hexChars) - 1);
+        std::uniform_int_distribution<int32_t> dist(0, static_cast<int32_t>(std::size(hexChars) - 1));
 
         char key[17];
         for (int32_t i = 0; i < 16; i++)
--- a/src/openrct2/network/NetworkServerAdvertiser.cpp
+++ b/src/openrct2/network/NetworkServerAdvertiser.cpp
@@ -27,6 +27,7 @@
 
 #    include <iterator>
 #    include <memory>
+#    include <random>
 #    include <string>
 
 #    ifndef DISABLE_HTTP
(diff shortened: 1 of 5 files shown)


src/openrct2/util/Util.cpp
```diff
@@ -21,6 +21,9 @@
 #include <cctype>
 #include <cmath>
 #include <ctime>
+#include <random>
+
+static std::mt19937 _prng;
```
IntelOrca: Is this thread safe?

src/openrct2/network/NetworkServerAdvertiser.cpp
```diff
@@ -241,6 +242,11 @@ class NetworkServerAdvertiser final : public INetworkServerAdvertiser
         static constexpr char hexChars[] = {
             '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f',
         };
+
+        // Seed with non-deterministic random number.
+        std::random_device rd;
+        util_srand(rd());
+
```
IntelOrca: I think it would be better to just not use `util_rand` in this method at all and instead use the c++ random API directly.

src/openrct2/util/Util.cpp
ZehMatt: Are we calling this from across threads currently? rand() isn't either or at least its not guaranteed and implementation specific.

src/openrct2/network/NetworkServerAdvertiser.cpp
ZehMatt: Done, also uses now uniform_int_distribution instead of modulo.

src/openrct2/util/Util.cpp
ZehMatt: I've changed it to use thread_local and is seeded with a number from random_device during initialization, this should do it and eliminates the need to call util_srand before usage.

Issue #9240, opened 2019-05-12
Assertion when passing a directory as the park on command line


zachary-D:
**OS:** Ubuntu Server 18.04
**Version:** 0.2.2
**Commit/Build:**  298c9f523

I installed everything fresh and when trying to start a headless server, some assertions fail and (as far as I can tell) the game server does not start as I cannot connect to it.

serverhost@open-rct2:~/OpenRCT2$ ./openrct2-cli host ~/serversaves --port 11753
An assertion failed, please report this to the OpenRCT2 developers.
Version: OpenRCT2, v0.2.2 (298c9f523 on HEAD)
Failed to allocate array of 9223372036854775807 * h (1 bytes)
Attempted to read past end of file.
An assertion failed, please report this to the OpenRCT2 developers.
Version: OpenRCT2, v0.2.2 (298c9f523 on HEAD)
Failed to allocate array of 9223372036854775807 * h (1 bytes)
Attempted to read past end of file.
Unable to detect file type.
openrct2 $

- [ ~ ] Specific to multiplayer?

**Steps to reproduce:**
1. run openrct2-cli with the above arguments

janisozaur:
Can you share the file you use?

ZehMatt:
The size looks like its int64_t(-1) and since the argument points to a folder rather than a sv6 I wonder if thats the cause for this.

IntelOrca:
@zachary-D You are passing a directory rather than a single save file to load. It should be something like `./openrct2-cli host ~/serversaves/park.sv6`.

11753 is the default port so you don't need to specify that.

PR #9243, opened 2019-05-12
Protect FileStream opening directories on unix/linux systems

ZehMatt:
Apparently thats a thing and yields potential issues if the user accidentally supplies directories as input.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -37,6 +37,7 @@
 - Fix: [#9000] Show correct error message if not enough money available.
 - Fix: [#9152] Spectators can modify ride colours.
 - Fix: [#9202] Artefacts show when changing ride type as client or using in-game console.
+- Fix: [#9240] Crash when passing directory instead of save file.
 - Fix: Guests eating popcorn are drawn as if they're eating pizza.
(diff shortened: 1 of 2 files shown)


IntelOrca:
Is this the fix for #9240? If so then it can be added as a changelog entry.

ZehMatt:
Good point, also need to add the correct includes.

janisozaur:
Looks mostly ok, but you have to take into account possibility that the argument is a symlink. Either allow them or verify the linked file instead (`readlink()`). For latter case, it would make a good platform function.

ZehMatt:
If the passed file is a symlink wouldn't stat return the information about the file? I don't know much about the Linux/Unix filesystems perhaps you can take over on this one.

IntelOrca:
@ZehMatt If it did that, how would you find out information about the symlink itself?

ZehMatt:
From what I read is that stat follows the symlink where lstat does not, thats why I'm unsure about this right now.

IntelOrca:
If that is the case then I am not sure why you need to change anything.

ZehMatt:
@janisozaur can you explain what you want me to do?

PR #9250, opened 2019-05-13
Implement BalloonPressAction

duncanspumpkin:
Furthering reducing the number of game commands and moving them to the Game Action framework.

This one was a real tough one thats why I thought I'd save it till near the end.


src/openrct2/actions/BalloonPressAction.hpp
```diff
@@ -0,0 +1,61 @@
...
+        {
+            log_error("Tried getting invalid sprite for balloon: %u", spriteIndex);
+            return MakeResult(GA_ERROR::INVALID_PARAMETERS, STR_NONE);
+        }
+        return MakeResult();
+    }
+
+    GameActionResult::Ptr Execute() const override
+    {
+        rct_sprite* sprite = try_get_sprite(spriteIndex);
```
Gymnasiast: Looks like this should be `_spriteIndex` (ditto for line 53).

duncanspumpkin: As per usual I try make a joke in my pr about it being easy then make a simple mistake

PR #9252, opened 2019-05-13
Small cleanup

Gymnasiast:
Mostly debris from Game Action ports.


src/openrct2/actions/TrackRemoveAction.hpp
```diff
@@ -19,6 +19,10 @@
 #include "../world/Surface.h"
 #include "GameAction.h"
 
+/**
```
duncanspumpkin: I don't see much reason for keeping this comment anymore.

duncanspumpkin:
If you look at the `new_game_command_table` you can delete all game command stubs for the ones that are implemented. Originally we kept the game command and just made it assert false if it was reached but there is no point in keeping them so just remove them.

src/openrct2/Game.cpp
```diff
@@ -1192,9 +1186,15 @@ GAME_COMMAND_POINTER* new_game_command_table[GAME_COMMAND_COUNT] = {
...
-    game_command_set_maze_track,
+    nullptr,
+    nullptr,
+    nullptr,
+    nullptr,
+    nullptr,
+    nullptr,
+    nullptr,
+    nullptr,
     game_command_set_park_entrance_fee,
```
duncanspumpkin: This one can be deleted as well

PR #9259, opened 2019-05-14
Implement ScenarioSetSetting game action

duncanspumpkin:
Further work moving game commands into the game action framework to improve design.


src/openrct2-ui/windows/EditorScenarioOptions.cpp
```diff
@@ -652,18 +654,9 @@ static void window_editor_scenario_options_financial_mousedown(rct_window* w, rc
...
-                        0, GAME_COMMAND_FLAG_APPLY, EDIT_SCENARIOOPTIONS_SETANNUALINTERESTRATE, 80,
-                        GAME_COMMAND_EDIT_SCENARIO_OPTIONS, 0, 0);
-                }
-                else
-                {
-                    game_do_command(
-                        0, GAME_COMMAND_FLAG_APPLY, EDIT_SCENARIOOPTIONS_SETANNUALINTERESTRATE, gBankLoanInterestRate - 1,
-                        GAME_COMMAND_EDIT_SCENARIO_OPTIONS, 0, 0);
-                }
+                auto interest = std::max(80, gBankLoanInterestRate - 1);
```
Gymnasiast: Should be `std::min`. Otherwise it causes the _minimum_ interest rate to be 80%...

src/openrct2/actions/ScenarioSetSettingAction.hpp
```diff
@@ -0,0 +1,290 @@
...
+    CostToBuyLand,
+    CostToBuyConstructionRights,
+    ParkChargeMethod,
+    ParkChargeEntryFee,
+    ForbidTreeRemoval,
+    ForbidLandscapeChanges,
+    ForbidHighConstruction,
+    ParkRatingHigherDifficultyLevel,
+    GuestGenerationHigherDifficultyLevel,
+    Count
```
AaronVanGeffen: For an `enum class`, `Count` should not be part of the `enum`. The compiler will complain if you don't handle it as a case, etc.

duncanspumpkin: This is how we are doing it for OpenRCT2. We need to be able to know the size of the enum.

IntelOrca: Only those which aren't `enum class`. I suppose the idea is that `enum class` represents a value of which should never no actual meaningful underlying value. I have used `constexpr size_t` underneath to this in other cases.

duncanspumpkin: All of the other game actions are using this method. I don't especially like it either but was going for consistency.

IntelOrca: What do you need a count for?

IntelOrca: Since you are only using `count` to check if it is valid I would instead just return the invalid result using the default case.

duncanspumpkin: I'd rather keep it matching the rest of the actions for now and then refactor them all in a oner later.

Gymnasiast:
Looks good to me. Haven't tested it yet and I suppose the enum class discussion needs to be resolved as well. But otherwise, I don't want to drag our feet to long on this.

duncanspumpkin:
I tested it last night and worked as expected. I don't mind changing the enum class depending on what people want but would prefer its something that can easily be applied to all of the game actions that use enum classes

src/openrct2/actions/ScenarioSetSettingAction.hpp
ZehMatt: I think its perfectly fine to do this, switches should have a default block where you should report some error because its an unexpected case, also defaulting the value to Count will also provide a way to tell if the variable was never assigned.

PR #9264, opened 2019-05-15
Tile Modify GameAction

duncanspumpkin:
Further work moving game commands into the game action framework

5

~~Todo: I've change it so that it takes CoordsXY need to modify the functions to take CoordsXY.~~
~~The code is still horrible. Considering splitting into multiple game actions.~~ Code is less horrible but not pretty.


ZehMatt:
The builds are failing, can you address that?

src/openrct2/ReplayManager.cpp
```diff
@@ -537,6 +538,27 @@ namespace OpenRCT2
                     result.action->SetFlags(command.ebx & 0xFF);
                     break;
                 }
+                case GAME_COMMAND_MODIFY_TILE:
+                {
+                    int32_t param1 = command.edx;
+                    int32_t param2 = command.edi;
+                    CoordsXY loc = { (command.ecx & 0xFF) * 32, ((command.ecx >> 8) & 0xFF) * 32 };
```
AaronVanGeffen: Untested, but:
```suggestion
                    auto loc = CoordsXY(command.ecx & 0xFF) * 32, ((command.ecx >> 8) & 0xFF) * 32);
```
is the cleanest way to get rid of the compile errors on Travis, I think.

PR #9271, opened 2019-05-17
fix: Staff patrol areas getting clobbered when hiring new staff

jwarby:
Hi, I think I have found and fixed an issue with the resetting of staff patrol areas when hiring new staff members that end up reusing a fired staff member's old ID.  I noticed that some of my staff patrol areas would occasionally get messed up so decided to spend some time looking into it.

Steps to reproduce:

1. Hire a new staff member and assign them a patrol area
2. Fire the new staff member
3. Hire another new staff member
4. Now try to either:
  - set their patrol area to the same square - nothing is highlighted and the area is not assigned
  - or, click on a new square - the new square and the old square are highlighted, and some graphical issues may occur on the old square

It looks like the problem was introduced when the staff hire action was added in [`f96a1a1b`](https://github.com/OpenRCT2/OpenRCT2/commit/f96a1a1b5f0174e8736464a11e00ea0a3127b8d2#diff-3c354ecc83a4335deb4414c2a5fcb981L337); the loop indexing when resetting the patrol areas for the new staff member's ID is different in the new action code.  I think my commit changes the indexing logic back to what it was prior to the action being introduced.

### DIFF
--- a/src/openrct2/actions/StaffHireNewAction.hpp
+++ b/src/openrct2/actions/StaffHireNewAction.hpp
@@ -260,9 +260,9 @@ private:
 
             gStaffModes[staffIndex] = STAFF_MODE_WALK;
 
-            for (int32_t newStaffIndex = 0; newStaffIndex < STAFF_PATROL_AREA_SIZE; newStaffIndex++)
+            for (int32_t i = 0; i < STAFF_PATROL_AREA_SIZE; i++)
             {
-                gStaffPatrolAreas[newStaffIndex * STAFF_PATROL_AREA_SIZE + newStaffId] = 0;
+                gStaffPatrolAreas[staffIndex * STAFF_PATROL_AREA_SIZE + i] = 0;
             }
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Great find there.

src/openrct2/actions/StaffHireNewAction.hpp
```diff
@@ -262,7 +262,7 @@ DEFINE_GAME_ACTION(StaffHireNewAction, GAME_COMMAND_HIRE_NEW_STAFF_MEMBER, Staff
 
             for (int32_t newStaffIndex = 0; newStaffIndex < STAFF_PATROL_AREA_SIZE; newStaffIndex++)
```
duncanspumpkin: Change this variable to `i` to remove future confusion.

jwarby: OK those are both done, cheers :)

PR #9273, opened 2019-05-18
Fix desync when host changes map

ZehMatt:
The server sent the map before sorting the peeps by name. The function peep_update_names calls peep_sort which changes their order in the sprite list.


ZehMatt:
The game state snapshots are totally worth it :) Still a little of trial and error left but knowing which fields were affected helped narrowing it down a lot.

PR #9286, opened 2019-05-20
Fix #9270: Refactor Money Effect Cpp

### LINKED ISSUE #9270: Money Effect Cpp Refactor
The money effect sprite is still in its original form from when we implemented it. It could do with being refactored into something more cpp'y. 
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/world/MoneyEffect.cpp

For inspiration of what it should look like have a look at Balloon.cpp
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/world/Balloon.cpp

The money effect struct `rct_money_effect` should have the following `public` member functions:

1. `static void CreateAt(...)`
2. `static void Create(...)`
3. `void Update()`
4. `rct_string_id GetStringId(...) const` consider refactoring this to return a `std::pair`

Two new `public` member functions should be added to `rct_sprite` struct.

1. `bool isMoneyEffect()`
2. `rct_money_effect* AsMoneyEffect()`

This issue is mainly an exercise in renaming functions and should be easy to follow for anyone with a basic understanding of C++. If you require further help please leave a message here or on gitter.

All C++ functions should follow the coding style guidelines. https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style Make sure all local variables are in lower camelCase and all function names are in upper CamelCase.

aw20368:
Changed static functions to rct_money_effect member functions. GetStringId now returns std::pair.

### DIFF
--- a/src/openrct2/world/MoneyEffect.cpp
+++ b/src/openrct2/world/MoneyEffect.cpp
@@ -17,11 +17,27 @@
 
 static constexpr const LocationXY16 _moneyEffectMoveOffset[] = { { 1, -1 }, { 1, 1 }, { -1, 1 }, { -1, -1 } };
 
+bool rct_sprite::IsMoneyEffect()
+{
+    return this->money_effect.sprite_identifier == SPRITE_IDENTIFIER_MISC
+        && this->money_effect.type == SPRITE_MISC_MONEY_EFFECT;
+}
+
+rct_money_effect* rct_sprite::AsMoneyEffect()
+{
+    rct_money_effect* result = nullptr;
+    if (IsMoneyEffect())
+    {
+        result = (rct_money_effect*)this;
+    }
+    return result;
+}
+
 /**
  *
  *  rct2: 0x0067351F
  */
-void money_effect_create_at(money32 value, int32_t x, int32_t y, int32_t z, bool vertical)
+void rct_money_effect::CreateAt(money32 value, int32_t x, int32_t y, int32_t z, bool vertical)
 {
     if (value == MONEY(0, 00))
         return;
@@ -44,10 +60,9 @@ void money_effect_create_at(money32 value, int32_t x, int32_t y, int32_t z, bool
     int16_t offsetX = 0;
     if (!gOpenRCT2NoGraphics)
     {
-        // Construct string to display
-        rct_string_id stringId = money_effect_get_string_id(moneyEffect, &value);
+        auto [stringId, newValue] = moneyEffect->GetStringId();
         char buffer[128];
-        format_string(buffer, 128, stringId, &value);
+        format_string(buffer, 128, stringId, &newValue);
         gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
         offsetX = -(gfx_get_string_width(buffer) / 2);
     }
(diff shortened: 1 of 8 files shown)


src/openrct2/world/MoneyEffect.cpp
```diff
@@ -44,8 +59,10 @@ void money_effect_create_at(money32 value, int32_t x, int32_t y, int32_t z, bool
     int16_t offsetX = 0;
     if (!gOpenRCT2NoGraphics)
     {
-        // Construct string to display
-        rct_string_id stringId = money_effect_get_string_id(moneyEffect, &value);
+        auto stringIdValueTuple = moneyEffect->GetStringId();
+        rct_string_id stringId = stringIdValueTuple.first;
+        money32 newValue = stringIdValueTuple.second;
+        *(&value) = newValue;
```
Gymnasiast: Is this awkward code really necessary?

Gymnasiast:
You will need to fix the formatting. Check the failed clang-format test to see what you need to change exactly. (Tip: do `git clang-format` after `git add`-ing the files, then do `git add` again.)

src/openrct2/world/MoneyEffect.cpp
aw20368: Decomposing the tuple isn't necessary, I was just unfamiliar with structured bindings. That's fixed.

If you're referring to the `*(&value) = newValue;`, that's what the original code essentially does, it changes the "value" of the value inside GetStringId(). Passing by reference is also awkward due to
`CreateAt(-value,...)` being called, so I'm not sure of a better solution.

aw20368:
Sorry, I'm having trouble with the clang-format. When I try to run `git clang-format` it tells me that it's not a git command. I also tried the python script in /scripts, but that does not find the file. VisualStudio has clang support, and it fixes things for me sometimes on the fly, but I can't find where to FORCE it to do a final check.

src/openrct2/world/MoneyEffect.cpp
duncanspumpkin: I don't think this is necessary either. The get string id function is basically calling absolute on the value. Just pass newValue into format string. No need to overwrite value. Ideally the parameters of this function should be const as they shouldn't be modified.

duncanspumpkin:
Some setups don't have clang format integrated with git. You will also find that clang format from the VS addin doesn't format it quite correctly due to being a different version. I've never worked out a way to force a final check client side so if you do find a way please ping me a message. 

Apart from the create at discussion PR looks good. Thanks for the changes. Please add yourself to contributors.md as well.

src/openrct2/world/MoneyEffect.cpp
aw20368: I looked over the old code and I agree. I missed the fact that the change wasn't propagated up the call stack.

aw20368:
This commit is causing a bug (money sprite cut off at some x value, and an assertion at Paint.sprite.cpp). Will further investigate.

src/openrct2/world/MoneyEffect.cpp
```diff
@@ -79,66 +95,59 @@ void money_effect_create(money32 value)
...
-    int32_t z = moneyEffect->z;
-    moneyEffect->move_delay = 0;
+    move_delay = 0;
 
-    if (moneyEffect->vertical)
+    if (vertical)
     {
         z += 1;
     }
     y += _moneyEffectMoveOffset[get_current_rotation()].y;
```
duncanspumpkin: This will be the source of your problem. This is adding the value onto the instance value instead of a new variable used just for sprite move.

aw20368: Thanks. That fixed the problem. I pushed the amended version (along with the suggested change for CreateAt).

PR #9301, opened 2019-05-24
Fix #9288: Replace repeated 3d to 2d functions

### LINKED ISSUE #9288: Replace repeated 3d to 2d functions
Related to #8723. There is also repeated implementations of the 3d to 2d function `coordinate_3d_to_2d` https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/world/Map.cpp#L140-L164

For example https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/world/Sprite.cpp#L643-L664 should be replaced by using the already implemented function. To complete this issue the codebase should be searched for any other instances of this function. Perhaps a search such as `(*y + *x) / 2 + *z` where * is a wild card would show up instances of this function. Unsure what would be the best way to find them.

In addition make sure  all local variables are in lower camelCase and all function names are in upper CamelCase. https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style

aw20368:
Replaced duplicate 3d coords to 2d code with translate_3d_to_2d_with_z(...) function

### DIFF
--- a/src/openrct2/interface/Screenshot.cpp
+++ b/src/openrct2/interface/Screenshot.cpp
@@ -253,30 +253,13 @@ void screenshot_giant()
     int32_t centreX = (mapSize / 2) * 32 + 16;
     int32_t centreY = (mapSize / 2) * 32 + 16;
 
-    int32_t x = 0, y = 0;
     int32_t z = tile_element_height(centreX, centreY);
-    switch (rotation)
-    {
-        case 0:
-            x = centreY - centreX;
-            y = ((centreX + centreY) / 2) - z;
-            break;
-        case 1:
-            x = -centreY - centreX;
-            y = ((-centreX + centreY) / 2) - z;
-            break;
-        case 2:
-            x = -centreY + centreX;
-            y = ((-centreX - centreY) / 2) - z;
-            break;
-        case 3:
-            x = centreY + centreX;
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
Just a thought but are there any issues with replacing the shift with a divide.

aw20368:
My understanding is that it's equivalent (up to negligible speed difference). I could change it the other way, or just leave those two functions separate?

Also I'm not sure what's going on with the failed CI builds for my commit. The clang-format passed while the mac clang xcode failed?

duncanspumpkin:
There would be differences at values of -1. I'm not sure if that's an issue though. I have a nagging feeling in the back of my mind that there might be an issue with interpreting what the mouse is hovering over. It's probably nothing though. 

As for the ci's it looks like they are erroring which tends to suggest that Travis is having issues rather than the pr.

aw20368:
Oh yes, it seems like it differs at a negative odd number. And perhaps it is an issue...as I'm playtesting just now, one of my half-underground tracks doesn't appear, so I will be reverting the change to that specific function.

duncanspumpkin:
Since it did make a difference could you leave a note in the function that this has to be a right shift to prevent issues at -1.

PR #9302, opened 2019-05-24
fix: Unable to use second skyscraper terrain edge style

jwarby:
Code to check if selected edge style is valid was passing the wrong object type
(surface instead of edge) to GetLoadedObject, causing that method to warn that
the selected edget style index was out-of-range.

Reproduction steps:

1. Open the Land tool
2. Select the last edge type (second skyscraper style edge)
3. Click a tile to try to change it to the selected edge, and the message "Can't change land type" is shown and the tile edge is not changed

### DIFF
--- a/src/openrct2/actions/SurfaceSetStyleAction.hpp
+++ b/src/openrct2/actions/SurfaceSetStyleAction.hpp
@@ -88,7 +88,7 @@ public:
             }
 
             const auto edgeObj = static_cast<TerrainEdgeObject*>(
-                objManager.GetLoadedObject(OBJECT_TYPE_TERRAIN_SURFACE, _edgeStyle));
+                objManager.GetLoadedObject(OBJECT_TYPE_TERRAIN_EDGE, _edgeStyle));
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Can you make an entry in the changelog and increment the network version.

jwarby:
@duncanspumpkin that's done now :)

Gymnasiast:
@IntelOrca @duncanspumpkin Why? The Game Action was merged after 0.2.2 came out. According to our own rules, it should _not_ get a changelog entry.

IntelOrca:
I didn't know it was a recent regression.

jwarby:
Let me know if you guys want me to take the changelog entry back out.

duncanspumpkin:
The consensus is that it shouldn't have a changelog entry as there hasn't been a release from between this bug being introduced and it being fixed. This is in line with how we generally do fixes. I thought this was from before the last release but that was incorrect (I think I was even the person that introduced the mistake).

PR #9303, opened 2019-05-24
fix: Unable to change colours of scrolling wall banner sign via sign dialog

jwarby:
1. Check in SignSetStlyeAction query was accidentally inverting value of
   wallFound boolean
2. Code in action execution was not looping through the tile elements to find
   the correct wall piece like the query did (which meant just fixing 1. would
   cause a segfault)

Moved code to find relevant wall tile element into new helper function
banner_get_scrolling_wall_tile_element() in Banner; use in both query and action
to avoid any duplication of search logic.

Reproduction steps:

1. Build a small scrolling banner (the wall style one which can be placed on any edge of a tile)
2. Right-click the banner to open the sign dialog
3. Try to change either the primary or secondary colour of banner - the colours don't change

### DIFF
--- a/src/openrct2/actions/SignSetStyleAction.hpp
+++ b/src/openrct2/actions/SignSetStyleAction.hpp
@@ -77,9 +77,9 @@ public:
         }
         else
         {
-            TileElement* tileElement = banner_get_scrolling_wall_tile_element((BannerIndex)_bannerIndex);
+            WallElement* wallElement = banner_get_scrolling_wall_tile_element((BannerIndex)_bannerIndex);
 
-            if (!tileElement)
+            if (!wallElement)
             {
                 log_warning("Invalid game command for setting sign style, banner id '%d' not found", _bannerIndex);
                 return MakeResult(GA_ERROR::INVALID_PARAMETERS, STR_NONE);
@@ -107,11 +107,11 @@ public:
         }
         else
         {
-            TileElement* tileElement = banner_get_scrolling_wall_tile_element((BannerIndex)_bannerIndex);
+            WallElement* wallElement = banner_get_scrolling_wall_tile_element((BannerIndex)_bannerIndex);
 
-            tileElement->AsWall()->SetPrimaryColour(_mainColour);
-            tileElement->AsWall()->SetSecondaryColour(_textColour);
-            map_invalidate_tile(coords.x, coords.y, tileElement->base_height * 8, tileElement->clearance_height * 8);
+            wallElement->SetPrimaryColour(_mainColour);
+            wallElement->SetSecondaryColour(_textColour);
+            map_invalidate_tile(coords.x, coords.y, wallElement->base_height * 8, wallElement->clearance_height * 8);
         }
 
         auto intent = Intent(INTENT_ACTION_UPDATE_BANNER);
--- a/src/openrct2/actions/SignSetStyleAction.hpp
+++ b/src/openrct2/actions/SignSetStyleAction.hpp
@@ -77,23 +77,9 @@ public:
         }
         else
         {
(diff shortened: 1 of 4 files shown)


src/openrct2/world/Banner.cpp
```diff
@@ -128,6 +128,27 @@ TileElement* banner_get_tile_element(BannerIndex bannerIndex)
     return nullptr;
 }
 
+TileElement* banner_get_scrolling_wall_tile_element(BannerIndex bannerIndex)
```
jwarby: Not sure if adding this helper method was the right way to do it, but I didn't want to duplicate the logic for finding the correct wall piece in both the query & execute of the action.

IntelOrca: This is good, it should return a WallTileElement as we know that is all we get.

src/openrct2/world/Banner.cpp
```diff
@@ -128,6 +128,27 @@ TileElement* banner_get_tile_element(BannerIndex bannerIndex)
...
+{
+    rct_banner* banner = &gBanners[bannerIndex];
+    TileElement* tileElement = map_get_first_element_at(banner->x, banner->y);
+
+    do
+    {
+        if (tileElement->GetType() != TILE_ELEMENT_TYPE_WALL)
+            continue;
```
IntelOrca: You can improve the code by only casting to wall once, the As methods are type safe.
```
auto wallElement = tileElement->AsWall();
if (wallElement != nullptr)
{
    auto sceneryEntry = wallElement->GetEntry();
    // etc.
}
```

src/openrct2/world/Banner.cpp
```diff
@@ -128,6 +128,27 @@ TileElement* banner_get_tile_element(BannerIndex bannerIndex)
...
 }
 
+TileElement* banner_get_scrolling_wall_tile_element(BannerIndex bannerIndex)
+{
+    rct_banner* banner = &gBanners[bannerIndex];
+    TileElement* tileElement = map_get_first_element_at(banner->x, banner->y);
+
+    do
+    {
+        if (tileElement->GetType() != TILE_ELEMENT_TYPE_WALL)
```
ZehMatt: Potentially nullptr, please add a check.

jwarby:
Sorry, just realised CI failed because of indentation, will fix when I am back home later.

src/openrct2/actions/SignSetStyleAction.hpp
```diff
@@ -77,23 +77,9 @@ DEFINE_GAME_ACTION(SignSetStyleAction, GAME_COMMAND_SET_SIGN_STYLE, GameActionRe
...
-                if (scenery_entry->wall.scrolling_mode == SCROLLING_MODE_NONE)
-                    continue;
-                if (tileElement->AsWall()->GetBannerIndex() != (BannerIndex)_bannerIndex)
-                    continue;
-                wallFound = true;
-                break;
-            } while (!(tileElement++)->IsLastForTile());
-
-            if (!wallFound == false)
+            WallElement* wallElement = banner_get_scrolling_wall_tile_element((BannerIndex)_bannerIndex);
```
duncanspumpkin: try to prefer using the `static_cast<BannerIndex>(` style. Its not the end of the world to use C style cast but `xxx_casts` are clearer as to what you are trying to achieve.

jwarby: OK I've changed that now :)

Gymnasiast:
Could you bump the network version? We're going to merge it.

jwarby:
@Gymnasiast just added a commit to bump the network version :+1:

Issue #9295, opened 2019-05-23
Refactor Vehicle state setting


duncanspumpkin:
When you change state on a vehicle the vehicle window should also be invalidated. This set and invalidate should be wrapped within a member function so that it can't be forgotten. In addition sub_state is often reset to 0. The sub_state reset needs a bit of further investigation into if this should also be moved into the member function. This is very similar to the Peep::SetState member function.

Create a new public member function of `rct_vehicle` called `SetState` it should take a `vehicle_status` enum as input. You should name the currently unnamed enum (see `vehicle.h`). The function should: set the state; invalidate the window; and depending on investigation of `sub_state`, reset `sub_state` to zero;. After creating the new member function replace all instances of `vehicle->state = xxxx; vehicle_invalidate_window(vehicle); ...` with the new member function. There may be some instances of `vehicle->state = xxxx;` that do not at present invalidate the window. Most likely those are wrong and they should invalidate the window. This task will involve some investigation.

In addition make sure all local variables are in lower camelCase and all function names are in upper CamelCase. https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style

When commiting in git try to stick to the commit message format. https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

Spongeloaf:
I'll give it a shot.

Spongeloaf:
Question time: When you wrote `vehicle->state = xxxx;`, did you mean `vehicle->status = xxxx;` ? Unless I am crazy or blind (Both are entirely possible), I do not see a member or rct_vehicle (or it's parent, rct_sprite_common). However, many places where `vehicle_invalidate_window(vehicle);` is called, I also see `vehicle->status` and `vehicle->sub_state`.

duncanspumpkin:
Yes sorry i meant status. You could just have sub state default value of 0 instead of creating another function and overloading. I think it's important to always consider the sub state when setting status.

Spongeloaf:
My primary hesitation for setting sub_state automatically is that I found a handful of functions that change the status but don't modify the sub_state at the same time. What should be done about those?

duncanspumpkin:
Well you could pass in the current sub state. It's also possible that those functions aren't setting the state correctly.

Spongeloaf:
> There may be some instances of vehicle->state = xxxx; that do not at present invalidate the window. Most likely those are wrong and they should invalidate the window.

[...]

PR #9329, opened 2019-05-29
#9295 Refactor rct_vehcile->status to rct_vehicle::SetState()

Spongeloaf:
Created rct_vehicle::SetState() and refactored all assignments to
rct_vehcile->status. SetState also includes a call to
vehicle_invalidate_window(), so it cannot be forgotten.


duncanspumpkin:
Try not to open and close pr's. It's better to just keep it open and keep working on it.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -2579,53 +2560,42 @@ static void vehicle_update_waiting_to_depart(rct_vehicle* vehicle)
...
             vehicle_update_rotating(vehicle);
             break;
         case RIDE_MODE_FILM_AVENGING_AVIATORS:
         case RIDE_MODE_FILM_THRILL_RIDERS:
-            vehicle->status = VEHICLE_STATUS_SIMULATOR_OPERATING;
-            vehicle->sub_state = 0;
-            if (ride->mode == RIDE_MODE_FILM_THRILL_RIDERS)
-                vehicle->sub_state = 1;
-            vehicle_invalidate_window(vehicle);
+            vehicle->SetState(VEHICLE_STATUS_SIMULATOR_OPERATING, 1);
```
duncanspumpkin: This isn't quite right

src/openrct2/ride/Vehicle.cpp
```diff
@@ -10014,3 +9913,10 @@ const rct_vehicle* rct_vehicle::GetCar(size_t carIndex) const
     }
     return car;
 }
+
+void rct_vehicle::SetState(VEHICLE_STATUS vehicleStatus, uint8_t subState)
```
duncanspumpkin: I think a default value of 0 for the sub state would simplify this. The vast majority of situations it's 0

src/openrct2/ride/Vehicle.cpp
Spongeloaf: In the original code , vehicle->substate will always be set to 1, because `case RIDE_MODE_FILM_THRILL_RIDERS` and `if (ride->mode == RIDE_MODE_FILM_THRILL_RIDERS)` appear to be redundant. Unless I've missed something.

src/openrct2/ride/Vehicle.cpp
Spongeloaf: I'll fix that.

src/openrct2/ride/Vehicle.h
```diff
@@ -111,6 +111,41 @@ enum VEHICLE_TYPE : uint8_t
     VEHICLE_TYPE_TAIL = 1,
 };
 
+enum VEHICLE_STATUS
```
ZehMatt: I think it would be better to use enum class with uint8_t as type.

Spongeloaf: I was doing some reading about enum classes earlier today. I wasn't going to suggest it because no other enums in vehicle.h use them, and I'm still very much learning my around c++. However, It would allow s4import and s6import to make safer range checks.

ZehMatt: Most of the code is from the days where everything was compiled as C, now that we compile everything as C++ we should use enum class wherever appropriate. It should look like following:
```
enum class VehicleStatus : uint8_t
{
     MovingToEndOfStation = 0,
     WaitingForPassengers,
     ...
};
```

Spongeloaf: I can do that. I'll also figure out how to range check appropriately, so adding or removing enums later on doesn't break anything.

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: But what about RIDE_MODE_FILM_AVENGING_AVIATORS

src/openrct2/ride/Vehicle.h
duncanspumpkin: I was going to leave changing it to an enum class for a second pr. I think it requires quite a good few changes and I'd prefer to have this pr just for set state.

ZehMatt: Oh I kinda figured since the lines are already affected it might be a good time but another PR works for me aswell

src/openrct2/ride/Vehicle.cpp
Spongeloaf: Ah yes, I see now. I gave it's own case, for the sake of explicitness.

src/openrct2/ride/Vehicle.h
duncanspumpkin: Your forgetting all of the branching code that checks the status. There will be a lot of it scattered throughout the code base.

Spongeloaf: @duncanspumpkin Was `Your forgetting all of the branching code` directed at @ZehMatt or me? I did find one location where we were still checking the status by integer value, but that's been fixed. (Will be pushed shortly).

duncanspumpkin: It was directed at zehmatt.

Spongeloaf: Ok. I'll push the recent change. If the tests fail again, what do we do? I was unable to decipher anything meaningful out of the logs. I've never used Travis-ci before, not really sure what to look for.

ZehMatt: Just look at the log for errors/warnings https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/540830525 and address them.

Spongeloaf:
I keep failing on S4Import and S6Import with 
````
/src/openrct2/rct2/S6Importer.cpp:1136:25: error: comparison is always true due to limited range of data type [-Werror=type-limits]
         if (src->status >= 0 || src->status <= 30)
             ~~~~~~~~~~~~^~~~
cc1plus: all warnings being treated as errors
````

I've tried a few tricks, like casting the enum values to uint8_t, and comparing status to ints (like the sample above) but it hasn't worked. I could drop the range check and just cast it, but we'll get undefined behavior if the source value is outside of our enum range. When the time comes to refactor the enum into an enum class, we'd have a much easier time of range checking. Should we drop it for now? Or is there a simpler method I'm not aware of?

I think I can cast src->status to an int, but I'm not sure how portable that might be, or what other trouble we could cause.

File: src/openrct2/rct1/S4Importer.cpp [hunk 4]
```diff
@@ -1243,6 +1243,15 @@ class S4Importer final : public IParkImporter
             }
         }
 
+        VEHICLE_STATUS statusSrc = VEHICLE_STATUS_MOVING_TO_END_OF_STATION;
+
+        // This will need to be updated if we ever add more VEHICLE_STATUS enums.
+        if (src->status >= 0 || src->status <= 30)
```
duncanspumpkin: The problem is that src->status is unsigned. Just drop the `src->status>=0` from this comparison. This line should be:
```
if (src->status <= static_cast<uint8_t>(VEHICLE_STATUS_STOPPED_BY_BLOCK_BRAKES))
```
Ideally there would be a way to get the size of the enum but that is not currently something in the standard. When we switch it to an enum class we will add a constexpr for enum class size.

src/openrct2/ride/CableLift.cpp
```diff
@@ -113,6 +112,10 @@ void cable_lift_update(rct_vehicle* vehicle)
         case VEHICLE_STATUS_ARRIVING:
             cable_lift_update_arriving(vehicle);
             break;
+
+        // prevents compiler warnings
```
duncanspumpkin: Don't add this comment its obvious why we need this.

src/openrct2/ride/Ride.cpp
```diff
@@ -4678,8 +4678,10 @@ static rct_vehicle* vehicle_create_car(
 
         vehicle->track_type = tileElement->AsTrack()->GetTrackType() << 2;
         vehicle->track_progress = 0;
-        vehicle->status = 0;
-        vehicle->sub_state = 0;
+
+        // Spongeloaf: This used to be "vehicle->status = 0;". I swapped it for the 0 enum. I hope it's correct.
```
duncanspumpkin: Good catch. Remove the comment.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -2027,6 +2027,10 @@ static void vehicle_update(rct_vehicle* vehicle)
             break;
         case VEHICLE_STATUS_DOING_CIRCUS_SHOW:
             vehicle_update_doing_circus_show(vehicle);
+
+        // prevents compiler warnings
```
duncanspumpkin: Remove the comment.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1129,7 +1129,16 @@ class S6Importer final : public IParkImporter
...
+
+        VEHICLE_STATUS statusSrc = VEHICLE_STATUS_MOVING_TO_END_OF_STATION;
+
+        // This will need to be updated if we ever add more VEHICLE_STATUS enums.
+        if (src->status >= 0 || src->status <= 30)
+        {
+            statusSrc = static_cast<VEHICLE_STATUS>(src->status);
+        }
+
+        dst->status =statusSrc;
```
duncanspumpkin: dont forget the space after the `=`

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1243,6 +1243,15 @@ class S4Importer final : public IParkImporter
             }
         }
 
+        VEHICLE_STATUS statusSrc = VEHICLE_STATUS_MOVING_TO_END_OF_STATION;
+
+        // This will need to be updated if we ever add more VEHICLE_STATUS enums.
```
duncanspumpkin: I would just remove this comment for now it will be addressed by the enum class change.

src/openrct2/rct1/S4Importer.cpp
Spongeloaf: That was the first thing I tried, but it failed too. It's possible that I misinterpreted the errors that time, so I'll try it again.

duncanspumpkin:
Try not to merge develop into your branch. Prefer rebasing. You will need to remove the lower check its always going to be true a uint can never be below zero.

Spongeloaf:
> You will need to remove the lower check its always going to be true

 Yeah, I kinda feel like an idiot for not noticing that. But thanks. I'm trying again.

duncanspumpkin:
Good, youre starting to get some nice aggressive clang format commits :-).

src/openrct2/ride/Vehicle.cpp
```diff
@@ -10049,6 +9951,13 @@ const rct_vehicle* rct_vehicle::GetCar(size_t carIndex) const
     return car;
 }
 
+void rct_vehicle::SetState(VEHICLE_STATUS vehicleStatus, uint8_t subState = 0)
```
ZehMatt: Default parameter should go to the declaration

Spongeloaf:
> aggressive clang format commits :-)

I'm totally onboard with having a robotic format-nazi reject my sloppy work. It's annoying in the moment but definitely worth it overall.

> Default parameter should go to the declaration

Done. Assuming the tests passagain, what's the next step after this?

duncanspumpkin:
Note to self. Squash on merge.

PR #9331, opened 2019-05-30
Add compile time selection for copying data with specialised paths

ZehMatt:
This improves the performance of IStream/MemoryStream most where the effect is most prominent with the recently introduced gamestate snapshots system.

openrct2 host parks/bpb.sv6:
```
In-Game FPS bpb.sv6 at 1080p Software (HW):
  Average Before: 39
  Average After: 56

GameStateSnapshots::Capture:
  Average Before: 0.003934 secs
  Average After: 0.002075 secs
```

This should also affect other areas where those functions are used, there are 280~ references to Read/ReadValue and 110~ references to Write/WriteValue

### DIFF
--- a/src/openrct2/core/MemoryStream.h
+++ b/src/openrct2/core/MemoryStream.h
@@ -58,7 +58,51 @@ public:
     void Seek(int64_t offset, int32_t origin) override;
 
     void Read(void* buffer, uint64_t length) override;
+    void Read1(void* buffer) override;
+    void Read2(void* buffer) override;
+    void Read4(void* buffer) override;
+    void Read8(void* buffer) override;
+    void Read16(void* buffer) override;
+
+    template<size_t N> void Read(void* buffer)
+    {
+        uint64_t position = GetPosition();
+        if (position + N > _dataSize)
+        {
+            throw IOException("Attempted to read past end of stream.");
+        }
+
+        std::memcpy(buffer, _position, N);
+        _position = (void*)((uintptr_t)_position + N);
+    }
+
     void Write(const void* buffer, uint64_t length) override;
+    void Write1(const void* buffer) override;
+    void Write2(const void* buffer) override;
+    void Write4(const void* buffer) override;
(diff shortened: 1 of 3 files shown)


Gymnasiast:
Result of doing `openrct2 benchgfx sc0.sc4 100`:
```

Without:
Rendering 100 times with drawing engine Software took 1.74 seconds. = 57 FPS

With:
Rendering 100 times with drawing engine Software took 1.63 seconds. = 61 FPS
```

I love how you managed to squeeze out more FPS, even though OpenRCT2 went through a lot of optimisations already.

ZehMatt:
Interesting, I didn't think it would influence singleplayer but I won't complain 😆

src/openrct2/core/MemoryStream.h
```diff
@@ -58,7 +58,51 @@ class MemoryStream final : public IStream
...
+
+    template<typename T> void Read(T& buffer)
+    {
+        uint64_t position = GetPosition();
+        if (position + sizeof(T) > _dataSize)
+        {
+            throw IOException("Attempted to read past end of stream.");
+        }
+
+        buffer = *reinterpret_cast<T*>(_position);
```
janisozaur: This is illegal, as you end up with misaligned pointers.

src/openrct2/core/MemoryStream.h
```diff
@@ -58,7 +58,51 @@ class MemoryStream final : public IStream
...
+            {
+                EnsureCapacity((size_t)nextPosition);
+            }
+            else
+            {
+                throw IOException("Attempted to write past end of stream.");
+            }
+        }
+
+        *reinterpret_cast<T*>(_position) = buffer;
```
janisozaur: likewise

src/openrct2/core/MemoryStream.h
ZehMatt: I'm not sure I see the issue, if you are at offset 1 and read 4 bytes via normal memcpy you have the same access pattern.

Issue #1260, opened 2015-06-07
Giant screenshot keying


Belzebub:
Hi!

Maybe you know the giant aerials at New Element.
We used to stitch those together hand by hand with photoshop, so it's AWESOME to see a giant screenshot function - thanks a lot!
However we used an altered version of the G1.DAT so that the background is pink to key out the background more efficiently (because otherwise pixels on the map get marked as well). Would it be possible to change the standard rct grey/black-background to pink/something user defined or even save the giant screenshot image transparent already?

duncanspumpkin:
I actually came across the code to do this today. Can probably get a solution for it tomorrow.

Gymnasiast:
Since the screenshot format is PNG by default, saving it with transparency directly might indeed be possible. That would save you the extra step of taking out the pink.

duncanspumpkin:
Hmm that might not be too easy. I know the sprite it assigns for drawing the background. My plan was to just upload a g2 version in pink and then link to that when in giant screenshot. It would be a bit harder to change the whole drawing code to also create a transparency mask.

Gymnasiast:
Ah, I didn't know that (which is why I said 'might'). In that case pink will do, I think. :)

duncanspumpkin:
Doh just realised this wont work at present. Some old code still accesses g1 so this will fail. I have to implement that first but its going to take a couple weeks to get the code ready for it.

zsilencer:
just have to set height of the background element to 0 so it doesn't draw, and memset the screenshot buffer to 0 before drawing and make that the transparency key.

duncanspumpkin:
@zsilencer ahh that does sound like a better approach.

zsilencer:
so i put it in a branch
however, there is still the matter of some maps having custom scenery that looks like the background, giving the effect that the map is shaped.
a solution is needed for that

Belzebub:
^We changed the black tiles with custom pink tiles and manually colored over the custom scenery like that. Not the fastest solution but it worked kinda well.

zsilencer:
ok if you are able to change the black tiles with custom tiles that are colored with palette index 0 instead of pink, then the giant screenshot should make it transparent for you.  if you could test out my branch and see?

Belzebub:
How can I test out your branch?

zsilencer:
here is the built dll, just replace the existing one
http://wikisend.com/download/310948/openrct2.dll.zip

[...]

PR #7349, opened 2018-03-25
 Add option to select land style for outside of the map

Broxzier:
Tagging #2332 since it's related, but asks for colours rather than land styles.

**Screenshot:**
image

PR #9334, opened 2019-05-30
Implement #1260: Transparent Giant Screenshots

Broxzier:
Evergreen Gardens 2019-05-30 16-42-41
Evergreen Gardens 2019-05-30 16-42-10


Gymnasiast:
Do we need to make this optional, really?

Broxzier:
Could've phrased it differently if you just want to say you think this should always be enabled. There are two reasons why I chose to make this an option: To give players the ability to keep using how it was with the black background, and to make it easier to extend this further in the future to something like in this PR: https://github.com/OpenRCT2/OpenRCT2/pull/7349

Gymnasiast:
> Could've phrased it differently if you just want to say you think this should always be enabled. 

I'm not sure how else one would interpret my words. If I had something against the feature/enhancement itself, I would have said that.

> To give players the ability to keep using how it was with the black background,

Not a very compelling argument. That could justify making _bug fixes_ optional.

>  and to make it easier to extend this further in the future to something like in this PR: #7349

YAGNI. Though that is at least a case where one might at least want to have some control over what background gets rendered  in giant screenshot.

Broxzier:
This is not a bug fix, but an alteration. The previous behaviour was not considered bad, hence having an option to keep using that seems useful to me. I get that clustering the user interface is not desirable. Would you be okay with this being a hidden option then?

Gymnasiast:
I still think it doesn't really need to be optional.

Hidden options are not that much better - they still clutter the codebase itself. I would leave the option in for now - we can always reconsider it later.

janisozaur:
Having that an option seems a bit weird to me. I like the feature though. How about having another menu entry like 'giant screenshot with transparent background', rather than option? Ideally though, letting the user just select background colour would the best way to approach this issue - with one of the options there explicitly being named "transparent in screenshots".

Btw, do we still have any way of doing BMP screenshots?

AaronVanGeffen:
> How about having another menu entry like 'giant screenshot with transparent background', rather than option?

I think that's a little overkill. More intrusive than having the option as-is, tbh.

> Ideally though, letting the user just select background colour would the best way to approach this issue - with one of the options there explicitly being named "transparent in screenshots".

Agreed. This is what @Broxier was doing in #7349 — something we may yet see in a future PR.

> Btw, do we still have any way of doing BMP screenshots?

I don't think we do.

Gymnasiast:
> Btw, do we still have any way of doing BMP screenshots?

I took out that code a few years ago.

PR #9335, opened 2019-05-30
Add version info resource to avoid/reduce false positive virus detection

ZehMatt:
My tests on VirusTotal have shown that adding this alone reduces the detections on 32 bit from 13 to 3. I need to tinker with AppVeyor to enable the version patching
Edit: This is what it looks like now:
explorer_2019-05-31_00-29-057156d11f287b69f84


resources/OpenRCT2.rc
```diff
@@ -12,11 +12,61 @@
 /////////////////////////////////////////////////////////////////////////////
 #undef APSTUDIO_READONLY_SYMBOLS
 
+/////////////////////////////////////////////////////////////////////////////
+// English (United States) resources
```
Gymnasiast: I think we should put UK English here - that's what we develop OpenRCT2 in.

resources/OpenRCT2.rc
```diff
@@ -12,11 +12,61 @@
...
+
+#if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_ENU)
+LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
+#pragma code_page(1252)
```
Gymnasiast: Shouldn't we use UTF-8, rather than Windows-1252?

resources/OpenRCT2.rc
```diff
@@ -12,11 +13,59 @@
 /////////////////////////////////////////////////////////////////////////////
 #undef APSTUDIO_READONLY_SYMBOLS
 
+/////////////////////////////////////////////////////////////////////////////
+// English (United States) resources
+
+#if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_ENU)
```
Gymnasiast:
```suggestion
#if !defined(AFX_RESOURCE_DLL) || defined(AFX_TARG_ENG)
```

resources/OpenRCT2.rc
```diff
@@ -12,11 +13,59 @@
 /////////////////////////////////////////////////////////////////////////////
 #undef APSTUDIO_READONLY_SYMBOLS
 
+/////////////////////////////////////////////////////////////////////////////
+// English (United States) resources
```
Gymnasiast:
```suggestion
// English (United Kingdom) resources
```

resources/OpenRCT2.rc
```diff
@@ -12,11 +13,59 @@
...
+            VALUE "ProductVersion", OPENRCT2_PRODUCT_VERSION
+        END
+    END
+    BLOCK "VarFileInfo"
+    BEGIN
+        VALUE "Translation", 0x809, 1200
+    END
+END
+
+#endif    // English (United States) resources
```
Gymnasiast:
```suggestion
#endif    // English (United Kingdom) resources
```

resources/OpenRCT2.rc
```diff
@@ -12,11 +13,59 @@
...
+/////////////////////////////////////////////////////////////////////////////
+
+
 /////////////////////////////////////////////////////////////////////////////
```
Gymnasiast: Removes the following five lines (they're duplicates).

JarnoVgr:
Would be great if this can be merged a.s.a.p.
Thanks for the work @ZehMatt

PR #9339, opened 2019-05-31
Fix #7323 Tunnel entrances not rendering in 'highlight path issues'

### LINKED ISSUE #7323: Tunnel entrances not rendering in 'highlight path issues' mode if they have benches inside
**OS:** Windows 8
**Version:** e.g. 0.1.3
**Commit/Build:** ccb1b33

In the 'Highlight path issues mode' (which is fantastic, by the way), tunnel entrances where paths go underground are not being rendered, which makes it look like the path dead-ends into a wall of dirt instead.

Normal (with transparent ride and scenery so the tunnel entrance is visible):

Tunnel disappears in 'Highlight path issues mode':

(Ride tunnels also disappear, though that's not as big of an issue for this view mode.)

- [No] Reproducible in RCT2 (vanilla)?
- [No] Specific to multiplayer?

**Steps to reproduce:**
1. Build a path that goes underground
2. Place a bench just inside the tunnel entrance
3. View it in 'Highlight path issues mode'.

aw20368:
Hiding path scenery caused function that draws tunnel to prematurely return. Changed to only skip drawing scenery.

### DIFF
--- a/src/openrct2/paint/tile_element/Paint.Path.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Path.cpp
@@ -690,6 +690,8 @@ static void sub_6A3F61(
 
     if (dpi->zoom_level <= 1)
     {
+        bool paintScenery = true;
+
         if (!gTrackDesignSaveMode)
         {
             if (tile_element->AsPath()->HasAddition())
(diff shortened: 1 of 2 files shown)


src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -690,6 +690,8 @@ static void sub_6A3F61(
 
     if (dpi->zoom_level <= 1)
     {
+        bool bPaintScenery = true;
```
Gymnasiast: That seems like a bit of an odd variable name - is that Hungarian notation? `paintScenery` should be clear enough, especially in this day of IDEs and actually useful compiler messages.

aw20368: I'm not sure, it's just a convention I used in a previous job. Amended.

Issue #4568, opened 2016-10-09
OpenRCT2 unnecessarily removes walls when building tracked rides, even with clearance checks disabled


DonCuan:
When building next to a wall or fence, both vanilla and OpenRCT2 clear obstructing wall pieces. The difference is that OpenRCT2 is more radical about it.

Vanilla only removes adjacent walls when building a station platform, which doesn't allow any wall pieces in the same block or when passing through a wall.

OpenRCT2 also clears walls when building straight pieces (flat or sloped). It doesn't remove anything when building a sharp 1x1 turn next to the wall and only the first wall piece when building a 2x2 turn. This happens even when clearance checks are disabled with a cheat.

**Steps to reproduce:**
1. Open a scenario with prebuilt walls in both RCT2 and OpenRCT2. Crazy Castle works well.
2. Build a tracked ride such as Mini Helicopters on the inner side of the wall.
3. Compare results

DonCuan:
On the other hand, OpenRCT2 clears less wall pieces when building a curve through a wall. RCT2 would have removed the next one as well:

!(https://cloud.githubusercontent.com/assets/2139850/19220728/8908e5b0-8e34-11e6-8078-77a5a2531738.png)

Liampie:
Same problem with paths.

DonCuan:
I didn't have any issue with the path tool deleting adjacent scenery when building parallel to the wall, in either mode.

It does delete the wall when going through it in bridge/tunnel mode, even with the clearance cheat, but that's not really what the issue is about.

PR #9341, opened 2019-05-31
Fix #5896 Fences not always removed when building a tracked ride through

### LINKED ISSUE #5896: Fences are not always removed when building a tracked ride through them
**OS:** Windows 10
**Version:** 0.1.1-develop
**Commit/Build:** 553e161

When building a tracked ride though a fence, the fence is removed when placed along one axis, but _not_ when placed along the other axis.

- [ ] Multiplayer?

**Steps to reproduce:**
1.  Place fences in a square
2.  Build a tracked ride through two non-opposite sides.
3. On one side the fence is removed, on the other it is not.

**Screenshots / Video:**

rct2fence

aw20368:
Fence removal did not account for track direction. Added the rotation for the track/fence intersection test.

Note that this bug also caused extraneous fences to be removed when building curved track (the picture on the left is the behavior in v0.2.2, my branch on the right).

Fence

### DIFF
--- a/src/openrct2/core/Numerics.hpp
+++ b/src/openrct2/core/Numerics.hpp
@@ -31,6 +31,18 @@ namespace Numerics
     }
 
     /**
+     * Bitwise left rotate of lowest 4 bits
+     * @param x unsigned 8-bit integer value
+     * @param shift positions to shift
+     * @return rotated value
+     */
+    [[maybe_unused]] static constexpr uint8_t rol4(uint8_t x, size_t shift)
+    {
+        x &= 0x0F;
+        return (x << shift | x >> (4 - shift)) & 0x0F;
+    }
+
+    /**
      * Bitwise right rotate
      * @tparam _UIntType unsigned integral type
      * @param x value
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
Is there not a rol4 function? If there isn't can you just create one since you are using it a few times.

AaronVanGeffen:
There is not, yet. Since it's only rotating a nybble, you probably can't fit it in the templated rotate functions #8633 introduced, though…?

aw20368:
I can use rol8 if I copy over the lower 4 bits to the upper 4 (and then mask with 0x0F), but that might be slightly hacky.

Should I...
1) Do the above
2) Add rol4 and ror4 to Numerics.hpp (or somewhere else)
3) Leave as-is (plus changelog/network)?

aw20368:
Also looks like this will resolve #4568.

duncanspumpkin:
Create a rol4 in numerics.

src/openrct2/core/Numerics.hpp
```diff
@@ -30,6 +30,18 @@ namespace Numerics
         return (((_UIntType)(x) << shift) | ((_UIntType)(x) >> (limits::digits - shift)));
     }
 
+    /**
+     * Bitwise left rotate of lowest 4 bits
+     * @param x unsigned 8-bit integer value
+     * @param shift positions to shift
+     * @return rotated value
+     */
+    static constexpr uint8_t rol4(uint8_t x, size_t shift)
```
AaronVanGeffen: Ugly, but to silence a CI warning:
```suggestion
    [[maybe_unused]] static constexpr uint8_t rol4(uint8_t x, size_t shift)
```

PR #9392, opened 2019-06-09
Peep Pickup Game Action

duncanspumpkin:
Guest and Staff pickup back into one simple game action.
Tested seems to work.

4
3


src/openrct2/Game.cpp
```diff
@@ -96,8 +96,8 @@ static GAME_COMMAND_CALLBACK_POINTER * const game_command_callback_table[] = {
     nullptr,
     nullptr,
     nullptr,
-    game_command_callback_pickup_guest,
-    game_command_callback_pickup_staff
+    nullptr,
+    nullptr
```
Gymnasiast: Now this table is empty, might as well get rid of it if it isn't too much trouble.

duncanspumpkin: I thought about doing that but then decided I'd make a second pr for that as I've a feeling it won't be as simple as just deleting this bit

duncanspumpkin:
There is a bug that prevents networked games from letting you drop peeps.

ZehMatt:
There seems to be an issue with canceling, this is done locally only so the moment you cancel the pickup the other clients will still see the peep in a pickup state and immediately desync.

duncanspumpkin:
I'm going to change it so that I pass the peep owner id instead of using the game action player id.

duncanspumpkin:
Found a small mistake in the pickup code. Not ready for merge.

PR #9403, opened 2019-06-11
Fix #9396. Pass the ghost flag to banner remove to prevent invalid removal

### LINKED ISSUE #9396: Deleting an underground path with a banner causes crash
**OS:** Ubuntu 18.04.2 LTS
**Version:** `0.2.2-develop-0~7499-201906091155-g993709c~ubuntu18.04.1`

When you delete (i.e. rightclick on a path) going from the upper left to the lower right corner of the screen, the game crashes.
This issue might be related to issue #8526 though this one is single player. Could not test it on multiplayer.

**Steps to reproduce:**
1. Open an arbitrary scenario map (I tested it with RCT1 and RCT2 scenarios)
2. find or build a path
3. rotate your view so that the path goes from the upper left corner of the  screen to the lower right corner of the screen (see screenshot).
4. Open the Scenery(Szenerie in german, do not know english translation) dialog and select a banner.
5. While having the banner in your hand, right click (i.e. delete) a piece of the path (marked with red in the screenshots)
6. the game hangs and must be killed with SIGKILL.
7. If you right click on a path not going from upper left to lower right then everything works as expected, that is: the path vanishes and I get 10 Euros worth of money.

This error message appears on the command line infinitely:

`ERROR[/build/openrct2-Xk7X9u/openrct2-0.2.2-develop-0~7499-201906091155/src/openrct2-ui/../openrct2/actions/BannerRemoveAction.hpp:60 (Query)]: Invalid banner location, x = 2656, y = 2752, z = 64, direction = 3`

**Screenshots / Video:**
banner_crash2
banner_crash

duncanspumpkin:
Previously this would never have happened because the old code would do a scenery ghost removal before calling anything else. I think this is a better way of handling it though and don't want to revert to the old method.

One thing I noticed is that this is deleting banners for free. There is scope to further improve this.

### DIFF
--- a/src/openrct2/actions/FootpathRemoveAction.hpp
+++ b/src/openrct2/actions/FootpathRemoveAction.hpp
@@ -157,7 +157,8 @@ private:
 
             auto bannerRemoveAction = BannerRemoveAction(
                 { x, y, tileElement->base_height * 8, tileElement->AsBanner()->GetPosition() });
-            bannerRemoveAction.SetFlags(GetFlags());
+            auto bannerFlags = GetFlags() | (tileElement->IsGhost() ? static_cast<uint32_t>(GAME_COMMAND_FLAG_GHOST) : 0);
+            bannerRemoveAction.SetFlags(bannerFlags);
             GameActions::ExecuteNested(&bannerRemoveAction);
             tileElement--;
         }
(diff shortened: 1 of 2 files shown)


Gymnasiast:
> One thing I noticed is that this is deleting banners for free

Normally, you would get 3/4 of the purchase price back.

PR #9404, opened 2019-06-11
Switch from RelWithDebInfo to MinSizeRel + -g for Linux amd64

### LINKED ISSUE #9357: Travis-ci linux build script is missing `curl`
As can be viewed in the numerous Travis-ci Linux build logs, the `curl` command is missing. It's used to push builds to https://openrct2.org .
macOS builds are unaffected.

janisozaur:
In https://github.com/OpenRCT2/OpenRCT2/issues/9357 it was found that
the archive to be uploaded is too big, over 32MiB quota imposed by
Google App Engine. To alleviate that, switch GCC to `-Os -g` to help
reduce the code size a bit.

Fixes #9357

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,12 +27,12 @@ env:
 matrix:
     include:
         - os: linux
-          name: Ubuntu amd64 GCC RelWithDebInfo
+          name: Ubuntu amd64 GCC MinSizeRel
           if: type != cron
           services:
             - docker
           env:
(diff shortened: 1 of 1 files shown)


janisozaur:
Let's see how big the archive is.

Ping @JarnoVgr

janisozaur:
I have some other ideas how to further improve that:
1. Use system gtest. While there is a warning, I think I know what I'm doing™ and it should be fine.
  - This should have the benefit of exposing cmake's `gtest_discover_tests`
2. Group the tests into common binary. Multiple smaller binaries are usually quicker to build than large one and easier to start porting, but I'm willing to sacrifice that for the sake of being able to produce binaries we distribute

janisozaur:
Any of the gtest changes won't help, as tests are not being packaged.

janisozaur:
Resulting package is 31752767 bytes.

33554432 (32MiB) - 31752767 = 1801665 ≃ 1760kiB

PR #9410, opened 2019-06-12
 Use system gtest on Ubuntu AMD64 to help reduce distributable size

janisozaur:
The tests were kept separate because:
* it limited a scope for porting to a new platform,
* (in usual case) it would be quicker to create multiple smaller
  executables than single large one,
* gtest carried a warning that mentioned possibly incompatible builds.

While the above still hold, size of distributable grew too large and
exceeded the quota of Google App Engine we use to distribute builds and
it takes precedence.

### DIFF
--- a/test/tests/LanguagePackTest.cpp
+++ b/test/tests/LanguagePackTest.cpp
@@ -14,10 +14,6 @@
 
 #include <gtest/gtest.h>
 
-#ifndef _WIN32
-const language_descriptor LanguagesDescriptors[] = {};
-#endif
-
 class LanguagePackTest : public testing::Test
 {
 protected:
@@ -38,7 +34,7 @@ TEST_F(LanguagePackTest, create_mutable_id_1)
     ILanguagePack* lang = LanguagePackFactory::FromText(1, "STR_0000:\n");
     ASSERT_EQ(lang->GetId(), 1);
     ASSERT_EQ(lang->GetCount(), 1U);
-    ASSERT_STREQ(lang->GetString(0), nullptr);
+    ASSERT_STREQ(lang->GetString(0), "");
     lang->SetString(0, "xx");
     ASSERT_EQ(lang->GetCount(), 1U);
     ASSERT_STREQ(lang->GetString(0), "xx");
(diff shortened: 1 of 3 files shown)


janisozaur:
Gtest was only packaged correctly since cosmic, 18.10. I'll have to think what to do about this. Probably revert to own gtest and change how tests are registered with ctest.

test/tests/LanguagePackTest.cpp
```diff
@@ -38,7 +34,7 @@ TEST_F(LanguagePackTest, create_mutable_id_1)
     ILanguagePack* lang = LanguagePackFactory::FromText(1, "STR_0000:\n");
     ASSERT_EQ(lang->GetId(), 1);
     ASSERT_EQ(lang->GetCount(), 1U);
-    ASSERT_STREQ(lang->GetString(0), nullptr);
+    ASSERT_STREQ(lang->GetString(0), "");
```
janisozaur: For some reason this changed on Linux when I made this PR, but Windows version started the same. Needs some investigation.

janisozaur:
Nice, works as expected. There is a failing test that needs to be addressed first, which is a command id 66 is nullptr, but gets called anyway. No idea why it doesn't show up on `develop`. @ZehMatt @duncanspumpkin can you look into it? The test is `Replay/ReplayTests.RunReplay/issue8345`

test/tests/LanguagePackTest.cpp
ZehMatt: It will return nullptr because the string is empty and it seems this have been like this since 2017. Maybe there is something under ASSERT_STREQ that didn't work right.

janisozaur: Hmm, good point. Perhaps its usage got changed

Broxzier:
```
error: Expected equality of these values:
  lang->GetString(0)
    Which is: NULL
  ""
[  FAILED  ] LanguagePackTest.create_mutable_id_1 (42 ms)
```

janisozaur:
Not required anymore, also the tests are not part of the binary we produce, so the whole point is moot.

PR #9414, opened 2019-06-13
Fix #8602: Wall piece collision detection deviates from vanilla

### LINKED ISSUE #8602: Wall piece collision detection deviates from vanilla
This is something that has bothered me for multiple years now, but I never bothered to file the report. Here we go. :P 

The screen should make the differences clear. Wooden supports, either a coaster's, path's or an object's, only get removed because of the placement of a wall if it's one clearance below it or at the same height. If it's two clearances below it, it will still draw the support. This isn't the case in vanilla. Seems like a small issue and hardly noticable, and in this example screen it is, but in practice it means that if you open old parks there may be random black supports spawned by raised objects in places where they are very unwelcome.

wallsupports

Using v0.2.1-600 (92412d4) on Windows 8

aw20368:
Wall piece collision used the base height of the wall. Changed to use the clearance height.

Behaviour after change:

PathSupport

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -39,6 +39,7 @@
 - Fix: [#8537] Imported RCT1 rides/shops are all numbered 1.
 - Fix: [#8553] Scenery removal tool removes fences and paths while paused.
 - Fix: [#8598] Taking screenshots fails with some park names.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Nice work!

PR #9422, opened 2019-06-15
Don't clear the screen to black for giant screenshots

Broxzier:
Someone mentioned this issue on the OpenRCT2 forums: https://forums.openrct2.org/topic/3782-giant-screenshot-transparency/?tab=comments#comment-21826

When certain viewport flags were set (underground view, hide base land, hide vertical faces, clip view), the viewport would always be cleared, including for giant screenshots. With this fix, the background will also be transparent when these flags are set.

Current behaviour:
trnds-before

With the fix:
trnds-vertical trnds-base trnds-see-through

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -845,12 +845,14 @@ static void viewport_fill_column(paint_session* session)
 static void viewport_paint_column(paint_session* session)
 {
     if (session->ViewFlags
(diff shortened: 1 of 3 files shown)


Gymnasiast:
Was this issue introduced with the transparent giant screenshots feature? If not, please update the changelog.

Broxzier:
@Gymnasiast Yes, before that feature it was not noticeable since the blank tile and the clear colour are the same colour.

PR #8481, opened 2018-12-18
Implement partial multicore rendering

ZehMatt:
For testing purpose, perhaps a feature at some point.

PR #9424, opened 2019-06-15
Fix crash with truetype fonts and multi threading

ZehMatt:
This should fix the crashes, I've tried various languages and parks from the Backtrace saves and it doesn't crash anymore.


Gymnasiast:
Could you add this to the changelog?

ZehMatt:
Done

AaronVanGeffen:
Multi-threading (#8481) was introduced after the 0.2.2 release, so I don't think the bugfix should have made the changelog. Not too bothered, though.

PR #9426, opened 2019-06-15
Refactor windows code

ZehMatt:
This is basically a start for a long run of refactoring the windows code. This first PR moves a few functions out of the global scope into rct_struct for skeleton work. The idea is to first move out most of the functions into rct_struct and once all direct data access is eliminated and it has all the members functions we can create a new interface and move the entire code into the UI part of the project. Once that is completed the window manager should take care of all the window management also. The end goal is to have the window manager do all the heavy lifting while providing safe access to the window list at any time even during events.

Because this involves a lot of changes I decided to do this gradually.

### DIFF
--- a/src/openrct2/interface/Window_internal.cpp
+++ b/src/openrct2/interface/Window_internal.cpp
@@ -1,0 +1,7 @@
+#include "Window_internal.h"
+
+void rct_window::SetLocation(int32_t newX, int32_t newY, int32_t newZ)
+{
+    window_scroll_to_location(this, newX, newY, newZ);
+    flags &= ~WF_SCROLLING_TO_LOCATION;
+}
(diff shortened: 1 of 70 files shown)


src/openrct2/interface/Window_internal.cpp
```diff
@@ -0,0 +1,42 @@
+#include "Window_internal.h"
+
+#include "../world/Sprite.h"
+
```
AaronVanGeffen: Can we move this file in its entirety to the `openrct2-ui` subproject? Most (if not all) of the callees are in the ui subproject as well, right?

ZehMatt: Thats planned for when everything is complete and have all the private members hidden behind an interface

janisozaur:
Is this something you would want merged before next release? Are you going to keep extending this PR or create new ones?

ZehMatt:
Not gonna extend this PR, its just a start for more consecutive PRs

PR #9432, opened 2019-06-16
Fix #8064 Ride construction errors shown as (undefined string)

### LINKED ISSUE #8064: Construction error about local authority not allowing construction that high is shown as (undefined string) every other height
**OS:** Windows 10]
**Version:** 0.2.1 and latest develop
**Commit/Build:** f468bef6cce7b55e0b5104f7260da8bb07675913

"local authority won't allow construction above tree height" errors get displayed as (undefined string) - `STR_CONSTRUCTION_ERR_UNKNOWN` in code - every other height step (when you're holding shift). so eg in the attached park you click to build a swinging inverter pirate ship, you get (undefined string), hold shift and move it 1 up, you get the proper error, move it 1 higher again and you get the (undefined string) again. 

I did try to investigate but I find the error code stuff a little baffling. All I can say is the error is set on line 1093 of Track.cpp. 

Related to #7439 and #8014 

**Steps to reproduce:**
1. Open attached save
2. Try and build an inverted pirate ship

**Save game:**
[Raceway Park.zip](https://github.com/OpenRCT2/OpenRCT2/files/2457788/Raceway.Park.zip)

aw20368:
The search for height to build track sometimes ended at a half-height, leading to (undefined string) error. Changed to always end at whole height and search as high as possible.

The code previously does at most 41 searches (magic number afaik) for a z-value to build at, if the last search was at a half-height, the search aborts while displaying the last error, which would be null. Since the code did only 41 searches, the below occurs prior to this PR, rather than building on top of the paths. I based my numAttempts to search until the max allowable track height (based on the ride being too high `if (clearanceZ >= 255)` in TrackPlaceAction.hpp). Should I refactor 255 into a constant, similar to MAXIMUM_LAND_HEIGHT in Map.h?

TrackUndefined

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -3576,8 +3576,8 @@ void ride_construction_toolupdate_construct(int32_t screenX, int32_t screenY)
         return;
     }
 
-    // search for appropriate z value for ghost, up to max ride height (2^12 - 16)
-    int numAttempts = (z <= 2032 ? (2032 - z) / 8 + 1 : 2);
+    // search for appropriate z value for ghost, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
     for (int zAttempts = numAttempts; zAttempts > 1; zAttempts--)
     {
         window_ride_construction_update_state(
@@ -3838,8 +3838,8 @@ void ride_construction_tooldown_construct(int32_t screenX, int32_t screenY)
         return;
     }
 
-    // search for appropriate z value, up to max ride height (2^12 - 16)
-    int numAttempts = (z <= 2032 ? (2032 - z) / 8 + 1 : 2);
+    // search for z value to build at, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
 
     for (int32_t zAttempts = numAttempts; zAttempts > 0; zAttempts--)
     {
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -3576,7 +3576,9 @@ void ride_construction_toolupdate_construct(int32_t screenX, int32_t screenY)
         return;
     }
 
-    for (;;)
+    // search for appropriate z value for ghost, up to max ride height (2^12 - 16)
+    int numAttempts = (z <= 2032 ? (2032 - z) / 8 + 1 : 2);
+    for (int zAttempts = numAttempts; zAttempts > 1; zAttempts--)
     {
         window_ride_construction_update_state(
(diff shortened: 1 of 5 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3576,7 +3576,9 @@ void ride_construction_toolupdate_construct(int32_t screenX, int32_t screenY)
         return;
     }
 
-    for (;;)
+    // search for appropriate z value for ghost, up to max ride height (2^12 - 16)
+    int numAttempts = (z <= 2032 ? (2032 - z) / 8 + 1 : 2);
```
Gymnasiast: 2032 appears to be a significant number. One that needs a constexpr.

duncanspumpkin:
The whole searching for z height is such horrible code. Surely there is a better way of doing this.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3576,7 +3576,9 @@ void ride_construction_toolupdate_construct(int32_t screenX, int32_t screenY)
         return;
     }
 
-    for (;;)
+    // search for appropriate z value for ghost, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
```
duncanspumpkin: If you are going to make this change make the change for mazes as well its just inconsistent as it is at present.
As this is a complex equation it should really be a constexpr that is named.

aw20368:
I'm open to take a shot at refactoring the whole thing, I just don't see how we can avoid searching for z when the expected behaviour is for the ride to be placed at the lowest permissible height at the player's cursor.

Gymnasiast:
@aw20368 I don't think that Duncan objects to the Z searching in general - just to the current implementation. And I agree with him that the current state of the code could do with a bit of a refactor. It's not that your PR contains bad code, it's just that it's adding to an already hard-to-read bit of code.

aw20368:
I mean, having read the code I definitely agree that it could use some work on it, I just wasn't sure if he meant wholesale changes.

I'll take a look and see what I can do with it.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3841,7 +3838,10 @@ void ride_construction_tooldown_construct(int32_t screenX, int32_t screenY)
         return;
     }
 
-    for (int32_t zAttempts = 41; zAttempts >= 0; zAttempts--)
+    // search for z value to build at, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
```
ZehMatt: If you were to have MAX_TRACK_HEIGHT by default of multiple by 8 you can apply following:
```suggestion
    int numAttempts = std::clamp(MAX_TRACK_HEIGHT - z, 2, MAX_TRACK_HEIGHT );
```

Gymnasiast:
@aw20368 Do you have time to continue this PR?

aw20368:
No not at the moment. If someone wants to take it over, that'd be welcome.

AaronVanGeffen:
Superseded by #10590.

Issue #9245, opened 2019-05-12
Hosting a server with --headless should not apply a Rich Presence status on Discord


shinsasch:
**OS:** Windows 10
**Version:** 0.2.2-426
**Commit/Build:** 62e1885

When you run a multiplayer server using the steps in https://github.com/OpenRCT2/OpenRCT2/wiki/Multiplayer#dedicated-server with `--headless`, it will apply Rich Presence like below if you happen to have Discord running on your computer while running the server. I am aware that with just `--headless` you can connect to other servers, but when you use the application to host a server with `--headless`, there is no point to have it apply a Rich Presence status.

**Steps to reproduce:**
1. `openrct2 host <saved game path> --port 11753 --headless`
2. Go to Discord, you will see a Rich Presence status being applied on your profile.

**Screenshots / Video:**
![image](https://user-images.githubusercontent.com/16175575/57588881-a12a9500-751b-11e9-899e-767ce2639888.png)
![image](https://user-images.githubusercontent.com/16175575/57588877-996af080-751b-11e9-9844-1ab6c9897dc6.png)

PR #9443, opened 2019-06-18
Do not apply Discord Rich Presence in headless mode

nhair:
This pull request attempts to fix issue #9245.

Fix is that the DiscordService object is not created if the program is running in "headless" mode.  Otherwise, it is created and Rich Presence is applied like normal.


Gymnasiast:
Could you update the changelog? Add the following line in the correct place:
```
- Fix: [#9245] Headless servers apply Discord Rich Presence.
```

nhair:
No problem, @Gymnasiast. I amended the commit with that added to the changelog.  Let me know if I did not add it to the correct place and I'll fix it.

PR #9446, opened 2019-06-18
Fix 9440 Tile Coordinate in tile inspector not updating

duncanspumpkin:
For whatever reason there are two coordinates in the tile inspector for holding the current tile x y. I swaped to the other one in the tile action refactor but it seems it wasn't updated in every location.

Since I was touching this code thought I might as well refactor to use the CoordsXY struct.

There is still room for improvement in here removing the second x y variable but it is a bit more complex than you would guess.

### DIFF
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -1026,18 +1026,24 @@ static void window_tile_inspector_mousedown(rct_window* w, rct_widgetindex widge
     {
         case WIDX_SPINNER_X_INCREASE:
             windowTileInspectorTileX = std::min<uint32_t>(windowTileInspectorTileX + 1, MAXIMUM_MAP_SIZE_TECHNICAL - 1);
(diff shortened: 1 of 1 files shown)


src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -446,8 +446,7 @@ static bool windowTileInspectorTileSelected = false;
 static int32_t windowTileInspectorToolMouseX = 0;
 static int32_t windowTileInspectorToolMouseY = 0;
 static bool windowTileInspectorToolCtrlDown = false;
-static int32_t windowTileInspectorToolMapX = 0;
-static int32_t windowTileInspectorToolMapY = 0;
+static CoordsXY windowTileInspectorToolMap = {};
```
Gymnasiast: Any reason to use CoordsXY instead of TileCoordsXY? The latter seems to be more appropriate for the Tile Inspector.

duncanspumpkin: Because i was standardising the game actions to use coordsxy

PR #9448, opened 2019-06-19
Fix marketing/money effect bugs

### LINKED ISSUE #928: Vehicle type can be changed to one that hasn't been researched
After the addition of feature #913, I believe its now possible to change the ride to use a vehicle type that hasn't been researched yet.

### LINKED ISSUE #9287: Starting advertisements in multiplayer causes desync
OS: Windows 10
Version: v0.2.2-467
Commit/Build: 74fa30a

In multiplayer games as soon as someone starts an advertisement the clients desync nonstop and no amount of resync fixes the issue. After reconnecting the clients immediately desync and see no advertisements running. The host appears unaffected.

This has been happening for many months and seems to happen on every map.

The map we are on tonight is Australasia - Fun At The Beach and it's happening.

ZehMatt:
This PR fixes two things.
1. MarketingCampaignFlags::FIRST_WEEK was never imported/exported. Given that the maximum value of weeks can be only 12 I decided to store the flag in the campaign_weeks_left field on the 7th bit. The only downside to this approach is that when you would export the save and try to load it on RCT2 it would show up 72 weeks. I was also considering to use some padding bytes but I wasn't sure what the default value can be of those.
2. If the game action does not return a valid position it would use the mouse position of the client and translates that into world space which of course produces a desync given that every client has the cursor somewhere else on the screen. I've disabled picking the screen position in multiplayer for when there is no valid position, it will simply not show any money effect for things like marketing actions.

This PR will resolve #9287

### DIFF
--- a/src/openrct2/management/Marketing.h
+++ b/src/openrct2/management/Marketing.h
@@ -36,6 +36,7 @@ enum
 
 enum
 {
+    CAMPAIGN_FIRST_WEEK_FLAG = (1 << 6),
     CAMPAIGN_ACTIVE_FLAG = (1 << 7)
 };
 
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -33,7 +33,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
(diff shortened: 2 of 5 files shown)


IntelOrca:
I don't see any other way of solving this so we will have to accept the incompatibility.

PR #9449, opened 2019-06-19
Remove unnecessary multiplication and division for sprite lists

ZehMatt:
We only need to do that during s6 import/export now. I think this made the whole thing look more complex than it is

### DIFF
--- a/src/openrct2/rct12/RCT12.h
+++ b/src/openrct2/rct12/RCT12.h
@@ -377,25 +377,25 @@ assert_struct_size(RCT12EightCarsCorruptElement15, 8);
 
 struct RCT12SpriteBase
 {
-    uint8_t sprite_identifier;      // 0x00
-    uint8_t type;                   // 0x01
-    uint16_t next_in_quadrant;      // 0x02
-    uint16_t next;                  // 0x04
-    uint16_t previous;              // 0x06
-    uint8_t linked_list_index;      // 0x08
-    uint8_t sprite_height_negative; // 0x09
-    uint16_t sprite_index;          // 0x0A
-    uint16_t flags;                 // 0x0C
-    int16_t x;                      // 0x0E
-    int16_t y;                      // 0x10
-    int16_t z;                      // 0x12
-    uint8_t sprite_width;           // 0x14
-    uint8_t sprite_height_positive; // 0x15
-    int16_t sprite_left;            // 0x16
-    int16_t sprite_top;             // 0x18
-    int16_t sprite_right;           // 0x1A
-    int16_t sprite_bottom;          // 0x1C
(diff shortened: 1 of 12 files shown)


Gymnasiast:
Looks OK. One thing though: * 2 is just multiplying by two, not taking things to the power of two.

ZehMatt:
Yeah true I'll rename it

src/openrct2/world/Sprite.cpp
```diff
@@ -299,7 +299,7 @@ static void sprite_reset(rct_sprite_generic* sprite)
 
     std::memset(sprite, 0, sizeof(rct_sprite));
 
-    sprite->linked_list_type_offset = llto;
+    sprite->linked_list_index = llto;
```
Gymnasiast: Perhaps change the name of the `llto` variable to `lli` accordingly.

src/openrct2/world/Sprite.cpp
```diff
@@ -392,15 +392,13 @@ rct_sprite* create_sprite(uint8_t bl)
  * of the desired linked list in a uint16_t array. Known valid values are
  * 2, 4, 6, 8 or 10 (SPRITE_LIST_... * 2)
```
Gymnasiast: Comment needs changing.

ZehMatt:
Changed

Gymnasiast:
Merge after 0.2.3 release.

src/openrct2/rct12/RCT12.h
```diff
@@ -377,25 +377,25 @@ assert_struct_size(RCT12EightCarsCorruptElement15, 8);
...
-    int16_t sprite_top;              // 0x18
-    int16_t sprite_right;            // 0x1A
-    int16_t sprite_bottom;           // 0x1C
-    uint8_t sprite_direction;        // 0x1E
+    uint8_t sprite_identifier;      // 0x00
+    uint8_t type;                   // 0x01
+    uint16_t next_in_quadrant;      // 0x02
+    uint16_t next;                  // 0x04
+    uint16_t previous;              // 0x06
+    uint8_t linked_list_index;      // 0x08
```
duncanspumpkin: Technically this one should have stayed as offset. Then it would make sense why you can't just copy it across.

Gymnasiast: Agreed.

ZehMatt: Agreed, I'll change it later today.

duncanspumpkin: Changed it myself :)

ZehMatt: Ha thanks, I was just about to start.

Issue #9285, opened 2019-05-20
Save current group in scenario list window


AaronVanGeffen:
Something I've noticed playing Locomotion is that it remembers which scenario group (tab) was active last time the scenario list window was used. It uses the normal config struct to achieve this. We should be able to easily introduce something similar.

A potential pitfall is that OpenRCT2 allows you to change the scenario grouping type. The setting should be reset to the first tab when switching types. Such a reset should also happen if an invalid setting is encountered in general, e.g. because RCT1 is no longer linked, or if the config INI file has been tampered with.

Spongeloaf:
I'd like to give this one a shot. It sounds simple enough, but I'm sure I'll manage to crash the game a few dozen times.

Spongeloaf:
Should it be persistent across game sessions?

AaronVanGeffen:
> Should it be persistent across game sessions?

That is what I had in mind, yes. You can use the game config for this.

Spongeloaf:
I've got it working already, I just need to do a little more digging to ensure edge cases won't be missed.

PR #9458, opened 2019-06-21
#9285 Add feature to remember scernario tab

Spongeloaf:
[#9285 Save current group in scenario list window](https://github.com/OpenRCT2/OpenRCT2/issues/9285)

It seems to work well enough. It remembers your tab between sessions, but sets the saved value back to 0 if the sorting type is changed, or if the path to RCT1 is changed.

In any case, if the tab is invalid, it defaults to the first populated tab. Also, I tried removing all of the scenarios, and it didn't crash or complain.

### DIFF
--- a/src/openrct2-ui/windows/TitleScenarioSelect.cpp
+++ b/src/openrct2-ui/windows/TitleScenarioSelect.cpp
@@ -226,10 +226,17 @@ static void window_scenarioselect_init_tabs(rct_window* w)
         }
     }
 
-    int32_t firstPage = bitscanforward(showPages);
-    if (firstPage != -1)
+    if (showPages & 1 << gConfigInterface.scenarioselect_last_tab)
     {
-        w->selected_tab = firstPage;
+        w->selected_tab = gConfigInterface.scenarioselect_last_tab;
+    }
+    else
+    {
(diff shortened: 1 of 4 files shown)


src/openrct2-ui/windows/TitleScenarioSelect.cpp
```diff
@@ -226,10 +226,17 @@ static void window_scenarioselect_init_tabs(rct_window* w)
         }
     }
 
-    int32_t firstPage = bitscanforward(showPages);
-    if (firstPage != -1)
+    if (showPages & 1 << gConfigInterface.scenarioselect_last_tab)
```
AaronVanGeffen: For clarity, I would prefer you change this to:
```suggestion
    if (showPages & (1 << gConfigInterface.scenarioselect_last_tab))
```

Spongeloaf:
Does anything else need to be done with this? Is it good to go?

AaronVanGeffen:
We're in feature freeze at the moment. It'll likely be merged after the impending release.

AaronVanGeffen:
This is probably worth mentioning in the changelog for v0.2.3+, right?

Broxzier:
@AaronVanGeffen It sure is.

PR #9466, opened 2019-06-23
Fix #6515: Add rain rendering to OpenGL renderer

### LINKED ISSUE #6515: No rain and thunder animation on OpenGL renderer
**OS:** Windows 10 1607
**Version:** 0.1.2
**Commit/Build:** ec22d04

When using the OpenGL renderer rain and thunder is not animated. The option for rendering weather is on! I haven't found a bug report for this and the wiki page didn't state the problem. Maybe this is something new related to some of the recent changes that got introduced. 

- [ ] Multiplayer?

**Steps to reproduce:**
1. Start the game with the OpenGL renderer
2. Open a park
3. Force rain or wait until it rains
4. No animation rendered

**Screenshots / Video:**
screen

ZehMatt:
Bringing back the rain!


ZehMatt:
Would be nice if some people could test this on their "older" machines to see how much of an performance impact there is given that the rain is drawn using lines instead of setting the pixels.

IntelOrca:
These are my frame rates in Forest Frontiers:
```
Sunny: 760
Rain: 680
Heavy: 660
```
In Magic Mountain:
```
Sunny: 300
Heavy: 280
```
So not too much of a drop.

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -115,6 +114,72 @@ class OpenGLDrawingContext final : public IDrawingContext
...
+    {
+        // clang-format off
+        static constexpr const uint8_t RainPattern[] =
+        {
+            32, 32, 0, 12, 0, 14, 0, 16, 255, 0, 255, 0, 255, 0, 255, 0, 255,
+            0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0,
+            255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255,
+            0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 0, 0
+        };
+        // clang-format on
```
IntelOrca: You should share this data between the two rain drawers.

ZehMatt: I wasn't sure where to put it exactly as the rain interface is supposed to be abstract, I could introduce a header file like RainPatterns.h if thats alrite

IntelOrca: Rain.h and Rain.cpp is where I would put it.

ZehMatt: Good point, its oddly abstract in too many ways.

ZehMatt:
Performance could be slightly better but there is no good way without refactoring too much in the OpenGL renderer. At some point I might put this into a shader and give it something more visually appealing.

ZehMatt:
Should we put this in the next release as well?

ZehMatt:
I rebased and updated the changelog, any objections?

PR #9479, opened 2019-06-25
Use 0.1.2b release of title sequence

janisozaur:
0.1.2b contains the same title sequence as previous release, but the
parks no longer use RLE chunks, which improves overall compression when
distributing as zip file.

This aims to alleviate the problems with Linux builds on Travis going
over quota and failing to upload to https://openrct2.org/


Gymnasiast:
I didn't see you already created this one, but https://github.com/OpenRCT2/OpenRCT2/pull/9480 updates _all_ URLs and shasums. I would prefer merging mine, I think it's better to consistently update stuff like this.

janisozaur:
Right, I was not aware we have so many URLs to update. I'd prefer if we had just one build system that could be used for all the platforms. If only such thing existed…

PR #9489, opened 2019-06-27
Fix #9485: Staff cannot be hired in pause mode

### LINKED ISSUE #9485: Staff cannot be hired in pause mode
**OS:** Ubuntu 19.04
**Version:** 0.2.2+
**Commit/Build:** b25c1c80887d571a69989838cc7a2513903af700

You cannot hire staff in pause mode. This is possible in vanilla and used to be possible in OpenRCT2. Most likely caused by the Game Action port of the relevant function.

41northstudios:


ZehMatt:
Looking good, can you also bump up the network version? You'll need to raise the version of ```NETWORK_STREAM_VERSION``` in Network.cpp

41northstudios:
> Looking good, can you also bump up the network version? You'll need to raise the version of `NETWORK_STREAM_VERSION` in Network.cpp

Done. Is there a general rule for when the network version should be incremented? It that something that needs to happen for all PRs or just when certain code is altered?

ZehMatt:
We generally do it only when the logic changes that would affect the network play. Given that clients wouldn't be able to hire staff while the server would be capable, this ensures both run the same logic. But its not something you have to worry too much about, we'll usually request it when we feel like there is a need to do it.

41northstudios:
> Also your commit should be something like this:
> `Fix #9485: Staff cannot be hired in pause mode`
> So it will automatically reference the issue.

Ahh thank you!! I thought putting the issue # in the title would automatically link to the issue, didn't realize I need it in the commit messages.

PR #9490, opened 2019-06-27
Correctly determine size of wchar arrays

muemart:
This PR fixes some mistakes when determining the size of wchar arrays I found while browsing the code:
- Most WinAPI functions want the number of characters, not the number of bytes. But sizeof returns the number of bytes.
- String literals already include the null-terminator, so there's no need to add 1 to the sizeof


ZehMatt:
Nice catch 👍

PR #9504, opened 2019-06-29
Fix #9498: Only close the most recent window when using the hotkey

### LINKED ISSUE #9498: Backspace closing all windows
OS: Windows 10
Version: 0.2.2-583
Commit/Build: e149722

Backspace is currently closing all windows instead of only the most recently opened window. Shift+Backspace shortcut is behaving normally.

ZehMatt:
window_close_top would iterate in reverse and break after closing the first window. To avoid creating a complex new path I decided to introduce flags on how to close the window with window_close_by_condition. I think window_close_top is the only one that specifically requires reverse iteration because the most recent window is always on the top.

### DIFF
--- a/src/openrct2/interface/Window.cpp
+++ b/src/openrct2/interface/Window.cpp
@@ -51,7 +51,6 @@ TextInputSession* gTextInput;
 
 uint16_t gWindowUpdateTicks;
 uint16_t gWindowMapFlashingFlags;
-
 colour_t gCurrentWindowColours[4];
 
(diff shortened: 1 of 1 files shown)


Gymnasiast:
What does this do if you use the hotkey to close the construction window? I think it should then open the ride window, like it does on vanilla.

duncanspumpkin:
That window opening is handled by the window close event. This wouldn't have any issues with that.

Gymnasiast:
Tested it. Seems to work correctly.

PR #9540, opened 2019-07-08
Remove sloped curves from Spinning Wild Mouse

Generalcamo:
The Spinning Wild Mouse did not have sloped curves in either vanilla RCT2 or RCT Classic. Unfortunately the cars lack the proper sprites to make this work, so this pull request reverts this to vanilla behavior. Sloped curves can still be used with the enable all track pieces cheat, but this restricts them in scenario play.


duncanspumpkin:
Assume this is related to https://github.com/OpenRCT2/OpenRCT2/commit/d88c01de745b389b517c9fcfd08346c96e5c5b5

Gymnasiast:
> Unfortunately the cars lack the proper sprites to make this work

Do you have a screenshot or video that shows this? Generally, cars that can do slopes can also do sloped turns, but the Spinning WM may be an exception.

Otherwise I will have to check for myself.

PR #7266, opened 2018-03-11
Make stdin / stdout a game console

IntelOrca:
This PR refactors the in game console into three classes:
  * A base class containing the commands.
  * An implementation for stdin / stdout (allowing interactive console in headless mode)
  * An implementation for the in game console (which is now in openrct2-ui)

One small issue is that anytime there is a printf or log in headless mode, it will mess up the current prompt on stdout. Fixing this would require replacing all stdout calls with our own method which overwrites the current prompt and writes a new prompt with the current statement. I am not going to worry about this for now.

I originally wrote the interactive console for the plug-in branch, but it will be a long time before that goes into the game. So for now this should allow more interim console commands to be added and used in headless servers.

For the interactive console, I have used `linenoise` which comes as a single .hpp that we can include.

PR #9543, opened 2019-07-08
Disable warnings around incompatible function pointer cast

janisozaur:
This addresses a compilation issue with newer (8+) mingw.

### DIFF
--- a/src/openrct2/platform/Platform.Win32.cpp
+++ b/src/openrct2/platform/Platform.Win32.cpp
@@ -222,10 +222,14 @@ namespace Platform
         if (hModule != nullptr)
         {
             using RtlGetVersionPtr = NTSTATUS(WINAPI*)(PRTL_OSVERSIONINFOW);
-#    pragma GCC diagnostic push
-#    pragma GCC diagnostic ignored "-Wcast-function-type"
+#    if defined(__GNUC__) && __GNUC__ >= 8
+#        pragma GCC diagnostic push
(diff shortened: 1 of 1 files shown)


ZehMatt:
Why are we using kernel mode APIs again? This should rather use GetVersionEx or VerifyVersionInfo

janisozaur:
It was introduced here: https://github.com/OpenRCT2/OpenRCT2/pull/7266

ZehMatt:
Dunno what to think of that, strictly speaking those APIs are identical to the user mode ones when in user mode but we should perhaps address this at some point.

Gymnasiast:
This breaks the MinGW CI. Is it correct, and is there a way to satisfy both versions?

Gymnasiast:
Summoning @IntelOrca .

IntelOrca:
I can't remember. The documentation says to now use the following APIs:

https://docs.microsoft.com/en-gb/windows/win32/sysinfo/version-helper-apis

In particular:
https://docs.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindowsversionorgreater

Gymnasiast:
What shall we do with this?

IntelOrca:
I don't have a problem with it being merged.

Gymnasiast:
@IntelOrca Neither do I, but the CI failed at the time. Perhaps that's now no longer the case, but it would need green CIs before merging.

PR #9544, opened 2019-07-08
Fix #9524: Cannot place staff in multiplayer without guest permission

### LINKED ISSUE #9524: Cannot place staff in multiplayer without guest permission
**OS:** Windows 10
**Version:** v0.2.2-a7f8960

Attempting to place staff without the guest permission in Multiplayer does not work. With the staff permission but not the guest permission, hiring is successful, as is modifying patrol areas and jobs, but actually placing the staff member does not work.

**Steps to reproduce:**
1. Grant player staff permissions but not guest permissions
2. Ask to hire a staff member
3. Player cannot place staff after hiring

ZehMatt:
Instead of merging the permissions I've created some macro that creates an alias for another game action so the implementation can be re-used with different game command ids.

### DIFF
--- a/src/openrct2-ui/windows/Guest.cpp
+++ b/src/openrct2-ui/windows/Guest.cpp
@@ -699,7 +699,7 @@ void window_guest_overview_mouse_up(rct_window* w, rct_widgetindex widgetIndex)
                 return;
             }
             w->picked_peep_old_x = peep->x;
-            PeepPickupAction pickupAction{ PeepPickupType::Pickup, w->number, {}, network_get_current_player_id() };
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
This just looks a bit messy. Is there not another way we could do it?

ZehMatt:
I mean I agree but its a little finicky given that it references its self inside query and execute otherwise I could have moved the entire implementation out to be shared. Also just registering the game action as a new one with a different ID is also not gonna work because its stored internally so it can be serialized/deserialized, thats why I kinda went for the proxy/alias way.

ZehMatt:
Well this is pretty messy, gonna close it. Only option I see is to either split the action or create a new permission for pickup.

Issue #9110, opened 2019-04-18
Use a vector for station locations list in sub_6CB945


duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Ride.cpp#L7253-L7376

`TileCoordsXYZD locations[(MAX_STATIONS * 2) + 1];` should be replaced by a `std::vector<TileCoordsXYZD> locations;`

https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Ride.cpp#L7274
Can then be modified to use a ranged for loop.

IntelOrca:
You can use `for (x : y)` on a array too. `std::array` might be better in this case, it really depends on how often the function is called. We should try to avoid mallocs where possible in game / draw tick unless it is infrequent or re-used per tick.

duncanspumpkin:
I want to avoid the requirement for the end marker. Yes it should probably be checked how often this is called.

duncanspumpkin:
I've confirmed it is not used every tick. It is therefore okay to use a vector for this.

ZehMatt:
Too bad STL does not provide a fixed size vector container.

janisozaur:
Isn't that just `std::array`?

ZehMatt:
std::array has no push/erase. It acts more like a C array with iterators.

janisozaur:
Why would you want to have `push()`? It's a fixed size array. I don't quite understand the use case you want.

ZehMatt:
std::vector is dynamic in space where I want one with finite room but still have the model of being able to insert and remove elements just by taking a slot in our finite memory storage, so capacity would stay at fixed N where size is dynamic in the range from 0 to N thus avoiding allocations that std::array offers. Its not exactly a new concept.

IntelOrca:
You basically want `std::tuple<std::array<T, N>, size_t>` with all the push, pop method etc. We could make our own although perhaps unnecessary for this.

ZehMatt:
Yeah not for this maybe, but if we wan't to replace more arrays where elements are added/erased dynamically more often this might be a good idea to implement at some point.

lwatson2016:
I can work on this! To confirm, you want to use `std::vector` for this?

duncanspumpkin:
For this i think a vector will be fine. It's not ideal but it will work.

PR #9546, opened 2019-07-09
Refactor #9110: Use a vector for station locations list in sub_6CB945

aaruel:
#9110

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -7114,42 +7114,38 @@ void sub_6CB945(Ride* ride)
         }
     }
 
-    // Needs room for an entrance and an exit per station, plus one position for the list terminator.
-    TileCoordsXYZD locations[(MAX_STATIONS * 2) + 1];
-    TileCoordsXYZD* locationList = locations;
+    std::vector<TileCoordsXYZD> locations;
     for (uint8_t stationId = 0; stationId < MAX_STATIONS; ++stationId)
     {
         auto entrance = ride_get_entrance_location(ride, stationId);
         if (!entrance.isNull())
         {
-            *locationList++ = entrance;
+            locations.push_back(entrance);
             ride_clear_entrance_location(ride, stationId);
         }
 
         auto exit = ride_get_exit_location(ride, stationId);
         if (!exit.isNull())
         {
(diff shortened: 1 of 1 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -7114,42 +7114,38 @@ void sub_6CB945(Ride* ride)
...
-            *locationList++ = exit;
+            locations.push_back(exit);
             ride_clear_exit_location(ride, stationId);
         }
     }
-    (*locationList++).x = COORDS_NULL;
 
-    locationList = locations;
-    for (; !(*locationList).isNull(); locationList++)
+    for (auto locationList = locations.cbegin(); locationList != locations.cend(); locationList++)
```
duncanspumpkin: This doesn't look pretty a ranged for loop would work here.

src/openrct2/ride/Ride.cpp
```diff
@@ -7114,42 +7114,38 @@ void sub_6CB945(Ride* ride)
...
     {
-        TileCoordsXYZD* locationList2 = locationList;
+        auto locationList2 = locationList;
         locationList2++;
 
         bool duplicateLocation = false;
-        do
+        for (; locationList2 != locations.cend(); locationList2++)
```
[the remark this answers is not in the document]
aaruel: Agreed, which is why I didn't go for the ranged for loop above, figuring having access to the iterator would keep finding duplicates simple. I can think up a prettier solution if you wanted

src/openrct2/ride/Ride.cpp
```diff
@@ -7114,49 +7114,46 @@ void sub_6CB945(Ride* ride)
...
+            locations.push_back(exit);
             ride_clear_exit_location(ride, stationId);
         }
     }
-    (*locationList++).x = COORDS_NULL;
 
-    locationList = locations;
-    for (; !(*locationList).isNull(); locationList++)
+    auto locationListIter = locations.cbegin();
+    for (const TileCoordsXYZD& locationCoords : locations)
```
aaruel: Should I prefer `const auto&` here and on L7143?

aaruel:
@duncanspumpkin Thank you! Yes I tested and it works as intended.

PR #9548, opened 2019-07-11
Travis: push master builds to openrct2.org as well

AaronVanGeffen:
Release builds for Linux, macOS, and Android were not being pushed to OpenRCT2.org. I'm not sure why, but these changes should address this for the next release.

Thanks to @JarnoVgr for bringing this to my attention.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,7 @@ matrix:
               export PUSH_BRANCH=master ;
               else export PUSH_BRANCH=$TRAVIS_BRANCH ; export FILENAME_PART=-${TRAVIS_BRANCH}-$(git rev-parse --short HEAD) ;
               fi
-            - if [[ "z$OPENRCT2_ORG_TOKEN" != "z" && "$TRAVIS_PULL_REQUEST" == "false" && ("${TRAVIS_BRANCH}" =~ ^(develop|push/) || "z${TRAVIS_TAG}" != "z") ]] ; then
+            - if [[ "z$OPENRCT2_ORG_TOKEN" != "z" && "$TRAVIS_PULL_REQUEST" == "false" && ("${TRAVIS_BRANCH}" =~ ^(master|develop|push/) || "z${TRAVIS_TAG}" != "z") ]] ; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -116,7 +116,7 @@ matrix:
               export PUSH_BRANCH=master ;
               else export PUSH_BRANCH=$TRAVIS_BRANCH ; export FILENAME_PART=-${TRAVIS_BRANCH}-$(git rev-parse --short HEAD) ;
               fi
-            - if [[ "z$OPENRCT2_ORG_TOKEN" != "z" && "$TRAVIS_PULL_REQUEST" == "false" && ("${TRAVIS_BRANCH}" =~ ^(develop|push/) || "z${TRAVIS_TAG}" != "z") ]] ; then
+            - if [[ "z$OPENRCT2_ORG_TOKEN" != "z" && "$TRAVIS_PULL_REQUEST" == "false" && ("${TRAVIS_BRANCH}" =~ ^(master|develop|push/) || "z${TRAVIS_TAG}" != "z") ]] ; then
```
janisozaur: This should already be covered by `!pull_request && "tag" != ""`

AaronVanGeffen: Interesting. I think you're right, but then I wonder why no builds were pushed.

ich777:
Will the latest version of the linux build be pushed to the releases?

JarnoVgr:
@ich777 Probably not. The release of v0.2.3 was kinda problematic as macOS and Linux builds weren't automatically built. macOS was built manually.
This PR will only fix it for future releases.

AaronVanGeffen:
They were built: https://travis-ci.org/OpenRCT2/OpenRCT2/builds/556998142

Something went wrong transferring them to OpenRCT2.org, however.

IntelOrca:
Could we deploy straight to the release for both Travis and AppVeyor like we do for our other repositories?

https://docs.travis-ci.com/user/deployment/releases/

AaronVanGeffen:
@IntelOrca Yes, I believe @JarnoVgr has suggested we do this, too.

JarnoVgr:
> Could we deploy straight to the github release for both Travis and AppVeyor like we do for our other repositories?
Yes, that would be the best solution for both the https://openrct2.org website and for the project itself. My website already fetches the files from the GitHub releases page and not from it's own API.

PR #9549, opened 2019-07-11
Prevent unnecessary copies in certain loops

AaronVanGeffen:
These changes prevent a few unnecessary copies by using references in certain loops.

### DIFF
--- a/src/openrct2/drawing/Font.cpp
+++ b/src/openrct2/drawing/Font.cpp
@@ -208,7 +208,7 @@ void font_sprite_initialise_characters()
 {
     // Compute min and max that helps avoiding lookups for no reason.
     _smallestCodepointValue = std::numeric_limits<char32_t>::max();
-    for (const auto entry : codepointOffsetMap)
+    for (const auto& entry : codepointOffsetMap)
     {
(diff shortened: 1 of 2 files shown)


ZehMatt:
Thats actually a great catch, good one 👍

AaronVanGeffen:
I'm not seeing a measurable difference in benchgfx, no. The difference is likely in the order of milliseconds.

ZehMatt:
Well does benchgfx even have any UI rendered at all given that this is related to fonts?

duncanspumpkin:
I caught this in my track design branch as well. It looks like the xcode ci doesn't fail on all warnings.

IntelOrca:
@AaronVanGeffen your changes are not in any frequently called functions, one is initialisation the other is only when you move the cursor while placing a track design on the map.

AaronVanGeffen:
@IntelOrca That certainly explains the lack of impact. Still, these unnecessary copies are best eliminated, I think. :)

IntelOrca:
@AaronVanGeffen oh yeah I am sure there many like that. It is often good to remove the copy constructor for complex types for that reason.

PR #9560, opened 2019-07-13
Renamed sound enums for accuracy and organization

Generalcamo:
Per the request of @Gymnasiast, renamed the sound enums for better accuracy.


duncanspumpkin:
We could look into switching this to an enum class as well whilst your at it.

src/openrct2/audio/Audio.cpp
```diff
@@ -59,9 +59,9 @@ rct_vehicle_sound_params* gVehicleSoundParamsListEnd;
 // clang-format off
 static int32_t SoundVolumeAdjust[SOUND_MAXID] =
 {
-    0,      // SOUND_LIFT_1
-    0,      // SOUND_TRACK_FRICTION_1
-    0,      // SOUND_LIFT_2
+    0,      // SOUND_LIFT_CLASSIC
+    0,      // SOUND_TRACK_FRICTION_CLASSIC_WOOD
+    0,      // SOUND_FRICTION_CLASSIC
```
Gymnasiast: I suppose you mean friction wheels? That's indeed one usage (the Junior RC, for example), but it's also used by rides like the dinghy slide, which uses a conveyor belt.

src/openrct2/audio/Audio.cpp
```diff
@@ -71,27 +71,27 @@ static int32_t SoundVolumeAdjust[SOUND_MAXID] =
     0,      // SOUND_SCREAM_4
     0,      // SOUND_SCREAM_5
     0,      // SOUND_SCREAM_6
-    0,      // SOUND_LIFT_3
+    0,      // SOUND_LIFT_VEKOMA
```
Gymnasiast: Hm, why Vekoma? Isn't this sound also used for the Junior RC and dinghy slide?

src/openrct2/audio/audio.h
```diff
@@ -92,27 +92,27 @@ enum RCT2_SOUND
...
     SOUND_COUGH_3,
     SOUND_COUGH_4,
-    SOUND_RAIN_1,
+    SOUND_RAIN,
     SOUND_THUNDER_1,
     SOUND_THUNDER_2,
-    SOUND_RAIN_2,
-    SOUND_RAIN_3,
+    SOUND_TRACK_FRICTION_TRAIN,
+    SOUND_TRACK_FRICTION_WATER,
```
Gymnasiast: Could you clarify this one? You might be right, but I don't immediately recognise the sound.

Gymnasiast:
Most of these look all right. I have put line notes on two of them, though.

src/openrct2/audio/Audio.cpp
Generalcamo: Yes it is, though the sound is used in odd places anyway. Per discussion, I'll rename it FRICTION_WHEEL for now.

src/openrct2/audio/audio.h
Generalcamo: This is used for most of the water rides. Examples include the River Rapids and the Dinghy Slide:
https://github.com/OpenRCT2/objects/blob/72dfb9adc551a0b5a2e173dcf6536402e9a5acb0/objects/rct2/ride/rct2.ding1.json#L32

Exceptions include the river rafts.

src/openrct2/audio/Audio.cpp
Gymnasiast: `FRICTION_WHEELS`, they always come in pairs ;)

PR #9571, opened 2019-07-14
Replay cleanup

ZehMatt:
I started a new version and removed all game command code as only track design is left which will eventually become a game action anyway, so as it is it will be simply not recorded.

I also removed all the replays as some of them broke during game action refactor and we started a new repo https://github.com/OpenRCT2/replays.

Not sure what the best way is to use the new repository for the CI to test them.


AaronVanGeffen:
You can let CI download the replay repo as an archive, e.g. [master.zip](https://github.com/OpenRCT2/replays/archive/master.zip)

janisozaur:
https://github.com/OpenRCT2/replays/archive/master.tar.gz would be a better choice

duncanspumpkin:
I ended up making it the same as objects in that it creates releases of the replays. I think it makes sense for us to have that sort of versioning capabilities. (The other reason is i didn't read these comments until I'd done it)

janisozaur:
I'll try reviewing this today in the evening

duncanspumpkin:
We will need to do something like https://github.com/OpenRCT2/OpenRCT2/commit/c465b35037c898963f2234b2dff5c69ce86456f3#diff-a529678596d0f697db67d414f6585955 to download the replays.

duncanspumpkin:
Okay so that is downloading the replays but i think on travis it is dumping it in the wrong folder. Can someone with more experience with cmake tell me what is wrong.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# CMAKE project for openrct2
+﻿# CMAKE project for openrct2
```
janisozaur: revert

CMakeLists.txt
```diff
@@ -287,6 +291,12 @@ if (DOWNLOAD_OBJECTS)
             file(REMOVE \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/object/objects.zip)\n\
         endif ()")
 endif ()
+if (DOWNLOAD_REPLAYS)
+    set(REPLAY_DIR "\$ENV{DESTDIR}${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/testdata/replays")
+    install(CODE "file(DOWNLOAD ${REPLAYS_URL} ${REPLAY_DIR}/replays.zip EXPECTED_HASH SHA1=${REPLAYS_SHA1} SHOW_PROGRESS)")
+    install(CODE "execute_process(COMMAND \"${CMAKE_COMMAND}\" -E chdir ${REPLAY_DIR}/ \"${CMAKE_COMMAND}\" -E tar xvf replays.zip)")
```
janisozaur: No need for verbosity here

CMakeLists.txt
```diff
@@ -287,6 +291,12 @@ if (DOWNLOAD_OBJECTS)
             file(REMOVE \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/object/objects.zip)\n\
         endif ()")
 endif ()
+if (DOWNLOAD_REPLAYS)
+    set(REPLAY_DIR "\$ENV{DESTDIR}${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/testdata/replays")
```
janisozaur: This should follow the structure of other such entries – allow caching of downloaded files.

duncanspumpkin:
How can we make it so that #9601 does not break existing replays?

ZehMatt:
One way would be to serialize cheats with variable length using the name for lookups maybe.

ZehMatt:
I've updated how cheats are stored, they are now stored by count and type so adding more cheats in will no longer break old replays, also preserved some space for potential future park parameters, this makes them a little bit less volatile against updates. Game action parameters are something that can't be easily made bullet proof, if the layout changes it will break replays with those game actions, splitting every parameter by type would be pretty harsh.

duncanspumpkin:
Looks fine by me. Just need to make some more replays.

ZehMatt:
Good to go?

janisozaur:
This broke Linux builds:

```
Using cached title sequences
Using cached objects
CMake Error at cmake_install.cmake:68 (file):
  file DOWNLOAD HASH mismatch

    for file: [/<<PKGBUILDDIR>>/debian/openrct2//usr/share/openrct2/testdata/replays/replays.zip]
      expected hash: [6e034e5299a8ec30a8d0a1b93f6a30a3a6429fec]
        actual hash: [da39a3ee5e6b4b0d3255bfef95601890afd80709]
             status: [6;"Couldn't resolve host name"]
```

janisozaur:
Though I suppose fixing it is on me, I should make sure the relevant repo is updated with the zip in question. I'll look into that later tonight

ZehMatt:
So just for the future, what must be done when replays are updated/version increments? I wasn't aware the CI failed actually.

janisozaur:
It's not the Travis ci, but launchpad which builds Ubuntu packages. The builders have no internet access there and the files need to be uploaded to another repository (I think it's one in my account and @Gymnasiast should have access), then synced to launchpad and all will be fine.

CMakeLists.txt
```diff
@@ -287,6 +291,16 @@ if (DOWNLOAD_OBJECTS)
             file(REMOVE \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/object/objects.zip)\n\
         endif ()")
 endif ()
+if (DOWNLOAD_REPLAYS)
+    install(CODE
+        "if (EXISTS \$ENV{DESTDIR}${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/testdata/replays/)\n\
```
janisozaur: Oh, this is no good. It should inspect a path that exists _before_ we issue an `install` target, like this: https://github.com/OpenRCT2/OpenRCT2/blob/18e35c1c98b6a084f58264862d05cd56399e2f7b/CMakeLists.txt#L286

Otherwise I can't inject the replays into the build system and it will fail to build.

IntelOrca: The launchpad thing doesn't need the replays though as it just needs to build, not test. Can't you disable the download?

janisozaur: That's a fair point, the replays should only be downloaded when building with tests. Thanks for pointing that out

PR #9573, opened 2019-07-14
Fix #7883: server log stored incorrectly if server name contains CJK

### LINKED ISSUE #7883: Headless server log is stored incorrectly if server name contains CJK in Ubuntu
**OS:** Ubuntu 18.04.1 LTS
**Version:** 0.2.0
**Commit/Build:** b25cb8c

It is about #6547 again, in headless server, server log is incorrectly stored if server name contains CJK.
I pulled a request #6565 and worked fine in Windows 10.
But when I try to start server in **Ubuntu**, korean server name is not correctly handled.
I have no idea where is problem, so I ask you a help.

In _config.ini_:
```
(...)
server_name = "서버"
(...)
```

openrct2-cli displays:
```
(...)
WARNING[/home/orct/build/src/openrct2/platform/Posix.cpp:108 (platform_utf8_to_multibyte)]: truncated string /home/orct/serverlogs/����
(...)
```
And a directory with corrupted name is generated. I have no idea whether log text file is in or not.
.
.

Notes:
 1. _config.ini_ file is using UTF-8 encoding
 1. ``locale`` command shows:
 ```
 LANG=ko_KR.UTF-8
 LANGUAGE=
 LC_CTYPE="ko_KR.UTF-8"
 LC_NUMERIC="ko_KR.UTF-8"
 LC_TIME="ko_KR.UTF-8"
 LC_COLLATE="ko_KR.UTF-8"
 LC_MONETARY="ko_KR.UTF-8"
 LC_MESSAGES=POSIX
 LC_PAPER="ko_KR.UTF-8"
 LC_NAME="ko_KR.UTF-8"
 LC_ADDRESS="ko_KR.UTF-8"
 LC_TELEPHONE="ko_KR.UTF-8"
 LC_MEASUREMENT="ko_KR.UTF-8"
 LC_IDENTIFICATION="ko_KR.UTF-8"
 LC_ALL=
 ```

IntelOrca:
Use correct locale name.

### DIFF
--- a/src/openrct2/platform/Posix.cpp
+++ b/src/openrct2/platform/Posix.cpp
@@ -98,7 +98,7 @@ void platform_get_time_local(rct2_time* out_time)
 static size_t platform_utf8_to_multibyte(const utf8* path, char* buffer, size_t buffer_size)
 {
     wchar_t* wpath = utf8_to_widechar(path);
-    setlocale(LC_CTYPE, "UTF-8");
+    setlocale(LC_CTYPE, "en_US.UTF-8");
     size_t len = wcstombs(NULL, wpath, 0);
     bool truncated = false;
     if (len > buffer_size - 1)


src/openrct2/platform/Posix.cpp
```diff
@@ -98,7 +98,7 @@ void platform_get_time_local(rct2_time* out_time)
 static size_t platform_utf8_to_multibyte(const utf8* path, char* buffer, size_t buffer_size)
 {
     wchar_t* wpath = utf8_to_widechar(path);
-    setlocale(LC_CTYPE, "UTF-8");
+    setlocale(LC_CTYPE, "en_US.UTF-8");
```
AaronVanGeffen: This will fail if the `en_US` locale is absent, right? Not unheard of, unfortunately. (In the past, I only had `en_UK` and `nl_NL` locales installed to save a bit of space...)

IntelOrca: Sure, I was waiting for feedback on this. I just found that this fixed it - so an improvement on before but not yet good enough.

Any ideas?

ZehMatt: We could go by the return value, if its absent it will return NULL otherwise the locale assigned.

PR #9589, opened 2019-07-16
Do not store park name as user string

IntelOrca:
When looking at the new save format, I have come to the conclusion that it is not necessary to store strings in a user table. Strings should be stored on the objects they relate to as `std::string`.

Starting with park name, I have moved it to be stored as a simple `std::string` in the `Park` class. When exporting back to S6, it allocates the user string for backwards compatibility. In rare the occurrence all user strings are used up, the park name gets saved as "unnamed park".

The benefit now is that the park name can share the same name as a ride.

### DIFF
--- a/src/openrct2-ui/windows/LoadSave.cpp
+++ b/src/openrct2-ui/windows/LoadSave.cpp
@@ -411,7 +411,7 @@ static bool browse(bool isSave, char* path, size_t pathSize)
             auto buffer = park.Name;
             if (buffer.empty())
             {
-                // Use localized "Unnamed Park" if park name was empty
+                // Use localised "Unnamed Park" if park name was empty.
                 buffer = format_string(STR_UNNAMED_PARK, nullptr);
             }
             safe_strcat_path(path, buffer.c_str(), pathSize);
--- a/src/openrct2-ui/windows/LoadSave.cpp
+++ b/src/openrct2-ui/windows/LoadSave.cpp
@@ -17,6 +17,7 @@
 #include <openrct2/Editor.h>
 #include <openrct2/FileClassifier.h>
 #include <openrct2/Game.h>
+#include <openrct2/GameState.h>
 #include <openrct2/config/Config.h>
 #include <openrct2/core/FileScanner.h>
 #include <openrct2/core/Guard.hpp>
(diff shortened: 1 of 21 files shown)


src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -406,15 +407,14 @@ static bool browse(bool isSave, char* path, size_t pathSize)
...
-            // Use localized "Unnamed Park" if park name was empty
-            if (String::SizeOf(buffer) == 0)
-                format_string(buffer, pathSize, STR_UNNAMED_PARK, nullptr);
-
-            safe_strcat_path(path, buffer, pathSize);
+            auto& park = OpenRCT2::GetContext()->GetGameState()->GetPark();
+            auto buffer = park.Name;
+            if (buffer.empty())
+            {
+                // Use localized "Unnamed Park" if park name was empty
```
Gymnasiast:
```suggestion
                // Use localised "Unnamed Park" if park name was empty.
```

(Might as well... :P)

src/openrct2/world/Park.cpp
```diff
@@ -79,17 +77,6 @@ int32_t _suggestedGuestMaximum;
  */
 int32_t _guestGenerationProbability;
 
-/**
- *
- *  rct2: 0x00667104
- */
-void reset_park_entry()
```
Gymnasiast: Why did you remove this?

IntelOrca: It was only called from one place which already initialised the data.

src/openrct2/drawing/ScrollingText.cpp
```diff
@@ -19,20 +19,15 @@
 #include <algorithm>
 #include <mutex>
 
-#pragma pack(push, 1)
-/* size: 0xA12 */
 struct rct_draw_scroll_text
```
Gymnasiast: I take it this structure is no longer tied to SV6 (or wasn't in the first place)? In that case, it's perhaps better to drop the `rct_` prefix and use TitleCase.

IntelOrca: It was never tied, just a draw structure.

src/openrct2-ui/windows/Park.cpp
```diff
@@ -582,8 +583,11 @@ static void window_park_set_disabled_tabs(rct_window* w)
 
 static void window_park_prepare_window_title_text()
 {
-    set_format_arg(0, rct_string_id, gParkName);
-    set_format_arg(2, uint32_t, gParkNameArgs);
+    auto& park = OpenRCT2::GetContext()->GetGameState()->GetPark();
+    auto parkName = park.Name.c_str();
+
+    set_format_arg(0, rct_string_id, STR_STRING);
+    set_format_arg(2, const char*, parkName);
```
Broxzier: This may be what causes some of the backtrace reports. `parkName` points to a temporary buffer returned by `c_str` that is not guaranteed to exist after this function call. Forcing this function to be inlined and calling it just before `gCommonFormatArgs` gets used may fix it.

IntelOrca: Thanks for spotting this, I will try to resolve...

IntelOrca: On closer inspection of the disassembly, this is not a temporary. It is actually using the internal buffer of the std::string on the global Park instance. The std::string is not copied.

IntelOrca: Okay, I found the cause. It is because the price of the park is overwriting the park name args in the arg buffer.

`set_format_arg(6, uint32_t, parkEntranceFee);`

PR #9590, opened 2019-07-16
Fix #9559: benchspritesort is broken after #8481

### LINKED ISSUE #9559: benchspritesort got broken in #8481
```
$ ./openrct2 benchspritesort out/usr/local/share/openrct2/title/titleseq/CocoaBayou.sv6 --benchmark_min_time=0.1
INFO[../src/openrct2/cmdline/BenchSpriteSort.cpp:74 (extract_paint_session)]: Starting...
INFO[../src/openrct2/cmdline/BenchSpriteSort.cpp:128 (extract_paint_session)]: Obtaining sprite data...
INFO[../src/openrct2/cmdline/BenchSpriteSort.cpp:134 (extract_paint_session)]: Got 0 paint sessions.
2019-07-13 16:37:10
Run on (4 X 1500 MHz CPU s)
Load Average: 0.14, 2.24, 3.73
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
---------------------------------------------------------------------
Benchmark           Time             CPU   Iterations UserCounters...
---------------------------------------------------------------------
baseline         2175 ns         1987 ns        70500 items_per_second=503.291k/s
$
```

#8481 broke benchmarking with `benchspritesort` (https://github.com/OpenRCT2/OpenRCT2/pull/8458)

janisozaur:
This takes multithreading into account as well

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -833,27 +833,27 @@ static void record_session(const paint_session* session, std::vector<paint_sessi
     // This is done to extract the session for benchmark.
     // Place the copied session at provided record_index, so the caller can decide which columns/paint sessions to copy; there
     // is no column information embedded in the session itself.
-    if (recorded_sessions != nullptr)
-    {
-        (*recorded_sessions)[record_index] = (*session);
-        paint_session* session_copy = &recorded_sessions->at(record_index);
+    (*recorded_sessions)[record_index] = (*session);
+    paint_session* session_copy = &recorded_sessions->at(record_index);
 
-        // Mind the offset needs to be calculated against the original `session`, not `session_copy`
-        for (auto& ps : session_copy->PaintStructs)
-        {
-            ps.basic.next_quadrant_ps = (paint_struct*)(ps.basic.next_quadrant_ps ? int(ps.basic.next_quadrant_ps - &session->PaintStructs[0].basic) : std::size(session->PaintStructs));
-        }
-        for (auto& quad : session_copy->Quadrants)
-        {
-            quad = (paint_struct*)(quad ? int(quad - &session->PaintStructs[0].basic) : std::size(session->Quadrants));
-        }
+    // Mind the offset needs to be calculated against the original `session`, not `session_copy`
+    for (auto& ps : session_copy->PaintStructs)
+    {
+        ps.basic.next_quadrant_ps = (paint_struct*)(ps.basic.next_quadrant_ps ? int(ps.basic.next_quadrant_ps - &session->PaintStructs[0].basic) : std::size(session->PaintStructs));
+    }
+    for (auto& quad : session_copy->Quadrants)
+    {
+        quad = (paint_struct*)(quad ? int(quad - &session->PaintStructs[0].basic) : std::size(session->Quadrants));
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/openrct2/interface/Viewport.cpp
```diff
@@ -836,9 +836,26 @@ void viewport_render(
...
 
-static void viewport_fill_column(paint_session* session)
+static void viewport_fill_column(paint_session* session, std::vector<paint_session>* sessions, size_t index)
 {
     paint_session_generate(session);
+    // Perform a deep copy of the paint session, use relative offsets.
+    // This is done to extract the session for benchmark.
+    if (sessions != nullptr)
+    {
+        (*sessions)[index] = (*session);
```
ZehMatt: I think it would be better to move the code into a new function to better separate the testing code out visibly.

src/openrct2/interface/Viewport.cpp
```diff
@@ -937,6 +954,11 @@ void viewport_paint(
 
     // Splits the area into 32 pixel columns and renders them
     size_t index = 0;
+    uint16_t column_count = (rightBorder - floor2(dpi1.x, 32)) / 32 + 1;
```
ZehMatt: This should be only done when sessions != nullptr otherwise unused.

src/openrct2/interface/Viewport.cpp
```diff
@@ -836,9 +836,26 @@ void viewport_render(
 #endif
 }
 
-static void viewport_fill_column(paint_session* session)
+static void viewport_fill_column(paint_session* session, std::vector<paint_session>* sessions, size_t index)
```
ZehMatt: The thing between session and sessions is unclear the same applies to the index, would be better to give it more concrete names.

src/openrct2/interface/Viewport.cpp
ZehMatt: You can also copy it after getting session_copy

janisozaur:
Hey @ZehMatt can you take another look? I addressed all your comments

src/openrct2/interface/Viewport.cpp
```diff
@@ -827,9 +827,33 @@ void viewport_render(
...
 }
 
-static void viewport_fill_column(paint_session* session)
+static void record_session(const paint_session* session, std::vector<paint_session>* recorded_sessions, size_t record_index)
+{
+    // Perform a deep copy of the paint session, use relative offsets.
+    // This is done to extract the session for benchmark.
+    // Place the copied session at provided record_index, so the caller can decide which columns/paint sessions to copy; there
+    // is no column information embedded in the session itself.
+    if (recorded_sessions != nullptr)
```
ZehMatt: Makes more sense to check that before the call, also have you forgotten our coding style by now 😄

janisozaur: Yeah, I figured as this is TU-local function I would rather put this check inside, since all our builds use LTO anyway and not clobber up the caller. I'll update it if that's what you prefer.

Tha clang-format check passes, do you have anything else in mind?

ZehMatt: Looks otherwise good. The reason why I want the check moved is that you can avoid the call and it makes it more clear.

janisozaur: Updated

PR #9593, opened 2019-07-17
Fix #9592: Builds fails on Arch with -DSTATIC=ON flag

### LINKED ISSUE #9592: Builds fails on Arch with -DSTATIC=ON flag
**OS:** Arch Linux
**Version:** develop
**Commit/Build:** 7b317e312c0b8a24584a0403f33f4d722c136729

After pulling a new copy of OpenRCT2, building with -DSTATIC=ON fails with:

[100%] Linking CXX executable openrct2-cli
/usr/bin/ld: libopenrct2.a(String.cpp.o): in function `String::ToUtf8[abi:cxx11](std::basic_string_view<wchar_t, std::char_traits<wchar_t> > const&)':
String.cpp:(.text+0x26e): undefined reference to `icu_64::UnicodeString::fromUTF32(int const*, int)'
<snip>
/usr/bin/ld: libopenrct2.a(LanguagePack.cpp.o):LanguagePack.cpp:(.text._ZN12LanguagePack6FixRTLERNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE[_ZN12LanguagePack6FixRTLERNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE]+0x313): more undefined references to `icu_64::UnicodeString::~UnicodeString()' follow
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/openrct2-cli.dir/build.make:85: openrct2-cli] Error 1
make[1]: *** [CMakeFiles/Makefile2:142: CMakeFiles/openrct2-cli.dir/all] Error 2
make: *** [Makefile:130: all] Error 2

- [ No ] Reproducible in RCT2 (vanilla)? 
- [ No ] Specific to multiplayer?

**Steps to reproduce:**
1. Pull latest OpenRCT2
2. Build using: cmake -DSTATIC=ON -DDISABLE_HTTP_TWITCH=ON -DDISABLE_NETWORK=ON -DDISABLE_TTF=ON -DDISABLE_GUI=ON ../

Note: Building with -DWITH_TESTS=ON also has some failures in the test executables. Same issue not being linked against -licuuc
Ex:
<snip>
/usr/bin/ld: libopenrct2.a(LanguagePack.cpp.o):LanguagePack.cpp:(.text._ZN12LanguagePack6FixRTLERNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE[_ZN12LanguagePack6FixRTLERNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE]+0x313): more undefined references to `icu_64::UnicodeString::~UnicodeString()' follow
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/test_pathfinding.dir/build.make:102: test_pathfinding] Error 1
make[1]: *** [CMakeFiles/Makefile2:112: CMakeFiles/test_pathfinding.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

mremallin:
Fixed Issue #9592 by:
- Adding test-common to multiple test executables
- Adding ${ICU_LIBRARIES} to the library list of openrct2-cli

Please let me know if the fix is incomplete or not proper CMake style.

Thanks!


src/openrct2-cli/CMakeLists.txt
```diff
@@ -16,5 +16,5 @@ file(GLOB_RECURSE OPENRCT2_CLI_SOURCES
 add_executable(${PROJECT_NAME} ${OPENRCT2_CLI_SOURCES})
 target_include_directories(${PROJECT_NAME} PRIVATE "${CMAKE_CURRENT_LIST_DIR}/..")
 ipo_set_target_properties(${PROJECT_NAME})
-target_link_libraries(${PROJECT_NAME} libopenrct2 Threads::Threads)
+target_link_libraries(${PROJECT_NAME} ${ICU_LIBRARIES} libopenrct2 Threads::Threads)
```
janisozaur: It's also required for `openrct2`. In fact, it's required by `libopenrct2`.

test/tests/CMakeLists.txt
```diff
@@ -174,7 +174,7 @@ if (NOT DISABLE_NETWORK)
     add_executable(test_crypt "${CMAKE_CURRENT_LIST_DIR}/CryptTests.cpp"
                               "${CMAKE_CURRENT_LIST_DIR}/TestData.cpp")
     SET_CHECK_CXX_FLAGS(test_crypt)
-    target_link_libraries(test_crypt ${GTEST_LIBRARIES} libopenrct2)
+    target_link_libraries(test_crypt ${GTEST_LIBRARIES} test-common libopenrct2)
```
janisozaur: This doesn't look right. Why is it needed here, when `libopenrct2` is already present?

src/openrct2-cli/CMakeLists.txt
mremallin: It looks like ICU is already included by libopenrct2 as part of src/openrct2/CMakeLists.txt Line 139. If libopenrct2 is being built statically, is it expected that the linker will pick up ICU from libopenrct2?

janisozaur: The line you mentioned tries linking `${ICU_STATIC_LIBRARIES}`, which is empty and that is the root cause.

janisozaur: unlike `${ICU_LIBRARIES}`, used with `STATIC=off`

mremallin:
Thanks for the comments janisozaur! They helped me to sort out the static library paths instead.

mremallin:
Thanks for the pointer to `FindICU.cmake`. Looks like there's a similar issue already tracked in the CMake repo: https://gitlab.kitware.com/cmake/cmake/issues/16617

For OpenRCT2, would it make more sense to require the user provide the library paths manually for a static build? (-DICU_STATIC_LIBRARIES="/usr/lib/libicuuc.a;/usr/lib/libicudata.a")

janisozaur:
Based on your comments, the issue you linked to seems unrelated to me.

mremallin:
I took a look at FindICU.cmake in my local Arch install. From what I can understand:

- There's no reference to ICU_STATIC_LIBRARIES, only ICU_LIBRARIES
- There is some special logic (static_prefix variable) which is only used for MSVC because the static ICU build with MSVC uses a different naming convention (libsicu*)
- The MSVC static prefix is only used as part of find_library, which means any library found would be put in the same variable (ICU_LIBRARIES) as the dynamically linked ones

As an experiment, I modified CMAKE_FIND_LIBRARY_SUFFIXES to look for '.a' first. (This is another hack, I know) After this, ICU_LIBRARIES contained the correct path of /usr/lib/libicuuc.a.
The build continued further but did not complete as it was missing /usr/lib/libicudata.a.

Based on the above, this seems like FindICU.cmake is not behaving as expected, right? It does not populate ICU_STATIC_LIBRARIES. Turning on ICU_DEBUG also shows no reference to any static libraries.

tupaschoal:
Issue quoted above seems to state that this has since been fixed.
If it hasn't, feel free to reopen the PR and resume discussion

PR #9601, opened 2019-07-17
Add new Cheat option: Automatic ride pricing

Sqoops:
Adds a new Cheat option to the game that lets rides have their admission price automatically set and updated as time progresses to just a bit lower then "overpriced".

cheats


duncanspumpkin:
Undo your changes to the language files. You should only be adding a single item to engb. This will break replays. Will need to have a think about how to get that to work.

Sqoops:
> 
> 
> Undo your changes to the language files. You should only be adding a single item to engb. This will break replays. Will need to have a think about how to get that to work.

done - only added STR_6329 and STR_6330 to the GB version

duncanspumpkin:
some how you have messed up the format of both language files. I suspect you have changed the line endings without realising. You will need to fix that before continuing.

duncanspumpkin:
👍

Sqoops:
> 
> 
> some how you have messed up the format of both language files. I suspect you have changed the line endings without realising. You will need to fix that before continuing.

I've used notepad++ with UTF-8 encode - I guess they are not in that format. Changed the files with the browser now.

AaronVanGeffen:
> I've used notepad++ with UTF-8 encode - I guess they are not in that format. Changed the files with the browser now.

They _are_ encoded in UTF-8. You had merely changed the line endings, as Duncan said.

Generalcamo:
Could an option be made to always set prices to "good value" in addition to an option to always set prices to "just below overpriced?" It seems both would be useful options for people to have.

Sqoops:
> 
> 
> Could an option be made to always set prices to "good value" in addition to an option to always set prices to "just below overpriced?" It seems both would be useful options for people to have.

Yes that could be done. First though I want this to actually get merged before I invest even more time into it. 

I had these values set to "good value" at first but found it quite a slog progressing through missions with that setting. 
"good value" is 1/4th of the money that you will get compared to "just below overpriced". Some rides will produce a deficit.

duncanspumpkin:
It will take a while to get this merged as it is dependant on updating replays. I want to make it so that new features don't break replays in general.

src/openrct2/localisation/StringIds.h
```diff
@@ -3965,8 +3965,11 @@ enum
     STR_TRANSPARENT_SCREENSHOT = 6327,
     STR_TRANSPARENT_SCREENSHOT_TIP = 6328,
 
+    STR_CHEAT_AUTOMATIC_RIDE_PRICING = 6329,
+    STR_CHEAT_AUTOMATIC_RIDE_PRICING_TIP = 6330,
+
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    STR_COUNT = 32768
+    STR_COUNT = 32770
```
Gymnasiast: Should not have been changed.

ZehMatt:
I'm not sure if I'm the only one against adding more cheats but we already have a ton of them and I think the idea is to implement scripting which could control the pricing. Also this makes the game a lot easier at some point as you have to no longer manage that yourself.

ocalhoun6:
Isn't making the game easier the whole point of cheats?

And if we're going to limit ourselves to a certain number of cheats because there's 'a ton of them', I could think of several cheats I'd rather get rid of than this one. This is a really nifty cheat that can greatly reduce the labor involved in managing a park.

Gymnasiast:
I suppose we'd ultimately want to spin off some of the cheats into plug-ins. Of course, that would need the plug-in system to be finished first.

I can see Matt's point of view, though - basically, this makes making money a lot easier, but since it's less visible, it feels less like cheating - which I'd argue is worse than cheating and knowing it darn well. But since this cheat only touches a few lines, I'm not immediately opposed to it.

As for the cutting the amount of cheats in general: that is something we can talk about. I don't think we should necessarily do that on this PR, but if either of you wants to open an issue, we can whether to scrap some cheats (and especially which ones) there.

data/language/en-GB.txt
```diff
@@ -3776,6 +3776,8 @@ STR_6325    :{SMALLFONT}{BLACK}Simulate ride/attraction
 STR_6326    :Can't simulate {POP16}{POP16}{POP16}{STRINGID}...
 STR_6327    :Transparent background for giant screenshots
 STR_6328    :{SMALLFONT}{BLACK}With this option enabled, giant screenshots will have a transparent background instead of the default black colour.
+STR_6329    :Automatic ride pricing
+STR_6330    :{SMALLFONT}{BLACK}Rides will have their admission price automatically set and updated as time progresses to just a bit lower then "overpriced".
```
Gymnasiast:
```suggestion
STR_6330    :{SMALLFONT}{BLACK}Rides will have their admission price automatically set to the maximum amount guests are willing to pay. The price will be kept up-to-date as the ride gets older.
```

IntelOrca:
Well the main point (and already has been) of cheats, is to help developers test the game more easily. This only needs to include the fundamentals, getting a lot of cash, triggering a lot of guests to enter the park, removing litter, resetting the date and also anti-cheats like removing all guests, triggering weather conditions.

I would say any of the cheats that aren't useful for developers such as this one are unnecessary. On the other hand we don't yet have a plugin system but when we do we can start to move more of these out the main source code.

There is already a preview of the plugin system in action:
https://www.youtube.com/watch?v=jmjWzEhmDjk

ocalhoun6:
Then wouldn't these 'developer tools' be best suited to a plugin as well?

Gymnasiast:
> Well the main point (and always has been) of cheats, is to help developers test the game more easily. This only needs to include the fundamentals, getting a lot of cash, triggering a lot of guests to enter the park, removing litter, resetting the date and also anti-cheats like removing all guests, triggering weather conditions.
>
> I would say any of the cheats that aren't useful for developers such as this one are unnecessary. 

I have to disagree there. It was never _my_ intention, and I have introduced the most cheats. I'm also sure that most outside contributors that have added cheats did not do so to help development.

Finally, regardless of the intention, most people that use cheats do not use them for development purposes.

IntelOrca:
Well most of them were added to get feature parity with 8-cars which did not work with OpenRCT2. I am just trying to say, exploding guests should be moved to a plug in

Gymnasiast:
The plug-in system has been merged in the meantime. Since a majority of the team thinks this should be a plug-in, I will close this PR.

enoeone19:
Hi, sorry for the noob question but how do I implement this cheat in the game? Navigating to the files changed tab, do I need to change all the files shown here then re-compile the game? Is there a way to download the already modified files? If that is the case, do I need to recompile the game or can I just overwrite the existing game files?

duncanspumpkin:
This has been replaced by a plugin. To use the plugin just install it in the plugin folder.

enoeone19:
Thanks, that simplifies things a lot.

PR #9613, opened 2019-07-19
Add new Cheat option: Build Benches, bins etc.

Sqoops:
Adds a new Cheat option to the game that places benches, bins, lamps, sprinklers randomly on all empty footpaths in your park and lamps, queue line tv's on all empty queue lines.
c1
c2
c3


ocalhoun6:
Personally, I'd like it better without the lamps.

Since lamps are for aesthetics only, I'd much rather carefully place them for aesthetics' sake than just have them randomly strewn about.

Also, what does this cheat do when...
- there are multiple bench/trash bin objects loaded in the scenario?
- there are no bench/trash bin/lamp objects loaded in the scenario?
- there are already path scenery objects in place -- does it replace them or ignore path tiles that already have benches/lamps, etc?

duncanspumpkin:
This would be better suited to be in a plugin. (when we get round to finishing the plugin system)

PR #9614, opened 2019-07-20
Can now pop balloons and quack ducks in title screen.

XplosiveLugnut:
Was staring at the title screen the other day and saw a balloon. I couldn't pop it. Nor could I make the ducks quack. I figured, "why not?"


xavery:
This significantly improves my title screen experience. Kudos!

janisozaur:
Since the read-only parks are now being modified, won't that affect autosaves?

XplosiveLugnut:
Doesn't look like autosaves are an issue at the title screen.
See the following code snippet from src/openrct2/GameState.cpp line 201

    // Always perform autosave check, even when paused
    if (!(gScreenFlags & SCREEN_FLAGS_TITLE_DEMO) && !(gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER)
        && !(gScreenFlags & SCREEN_FLAGS_TRACK_MANAGER))
    {
        scenario_autosave_check();
    }

XplosiveLugnut:
Also, I'm just guessing that I merged the changelog correctly.

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -74,6 +74,15 @@ int32_t viewport_interaction_get_item_left(int32_t x, int32_t y, viewport_intera
     tileElement = info->tileElement;
     sprite = (rct_sprite*)tileElement;
 
+    // Allows only balloons to be popped and ducks to be quacked in title screen
+    if (gScreenFlags & SCREEN_FLAGS_TITLE_DEMO)
+    {
+        if (info->type == VIEWPORT_INTERACTION_ITEM_SPRITE && sprite->generic.sprite_identifier == SPRITE_IDENTIFIER_MISC)
+            return info->type;
+        else
+            return info->type = VIEWPORT_INTERACTION_ITEM_NONE;
```
AaronVanGeffen: Not too confident about this assignment here. Should perhaps just return `VIEWPORT_INTERACTION_ITEM_NONE` instead?

XplosiveLugnut:
Alright, when I saw your comment, and the code snippet, I thought to myself "yeah that does look weird, why did I do that?"

After checking the whole file and not just the commit changes, I have the answer. This is done numerous times in this file.

The reasoning appears to be that "info->type" needs to be assigned and the function needs to return what it was assigned to.

Upon a google, I found https://stackoverflow.com/questions/15292892/what-is-return-type-of-assignment-operator

So this is a valid way to do that (return what something was assigned to). `A = B = C;` is effectively the same as `B = C; A = B;` By the same logic, `return B = C;` is the same as `B = C; return B;`.

So, if assigning something is the last thing the function needs to do before returning what that thing was assigned to, this works fine.

duncanspumpkin:
I don't think anyone is debating if it works. More it just looks odd. Break it up into two lines.

XplosiveLugnut:
"Not too confident about this assignment"
Reads like uncertainty to me. That's why I confirmed that it is valid and works as intended, like it does everywhere else in the file already.

"Break it up into two lines."
Well it's not *that* simple. :wink:

Anyway, I got rid of *all* of them that I could find, which wasn't many, so I figured it's no big deal.
I do feel this is out of the scope of this pull request though.

AaronVanGeffen:
Oh, I was aware of the transitive properties of assignments. It was the assignment itself that seemed odd to me, hence the shout-out.

Refactoring all assignments in return clauses should really be done in a different PR, if at all. I personally do not think it is necessary.

XplosiveLugnut:
Yes, after much messing around with SSH keys and emails, I think it's good.

XplosiveLugnut:
Ohhhh. Ok. Yes. I completely misunderstood. I'll revert that last refactoring commit.

src/openrct2-ui/interface/ViewportInteraction.cpp
XplosiveLugnut: Still learning git hub here.

Ok, looking into *why* I'm making this assignment here. I'm sure I had reasoned it out. But I don't remember.

AaronVanGeffen:
I would prefer the last two commits to be dropped entirely, as adding-then-reverting merely pollutes the repository history. Alternatively, we could just squash-merge the PR.

XplosiveLugnut:
Agreed. What ever ends up being the cleanest. Guessing squash-merging turns the whole PR into a single commit? That sounds good to me.

src/openrct2-ui/interface/ViewportInteraction.cpp
XplosiveLugnut: Yup, looks like that assignment isn't necessary. Thanks. I'll fix that.

Ok, so what do I do here now? Talking about this little thread in particular. Do I click resolve/start a review? Do I wait until I've made the next commit?

AaronVanGeffen: You can 'resolve' the thread/conversation after pushing a new commit, no worries.

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -80,7 +80,7 @@ int32_t viewport_interaction_get_item_left(int32_t x, int32_t y, viewport_intera
         if (info->type == VIEWPORT_INTERACTION_ITEM_SPRITE && sprite->generic.sprite_identifier == SPRITE_IDENTIFIER_MISC)
             return info->type;
         else
-            return info->type = VIEWPORT_INTERACTION_ITEM_NONE;
+            return VIEWPORT_INTERACTION_ITEM_NONE;
```
Broxzier: Are you sure this assignment is unnecessary? It couild be that the caller doesn't check the return type of this function, but does check the `info` instance.

XplosiveLugnut: I should have documented how the code got to this point. I don't clearly remember why, but I believe I determined that the assignment was not only unnecessary, but _could_ actually cause a problem.
I'll look into it again and document it this time. It'll be a good exercise for me. Might even reach a different conclusion.

Gymnasiast: @XplosiveLugnut Well, was it? ;-)

distribution/changelog.txt
```diff
@@ -4,6 +4,7 @@
 - Feature: [#6553] Android version now runs in full screen.
 - Feature: [#7865] Transport rides can now be synchronised.
 - Feature: [#9073] Shortcut keys for the Tile Inspector.
+- Feature: [#9614] Can now pop balloons and quack ducks in title screen.
```
Gymnasiast, line 7:
```suggestion
- Feature: [#9614] Allow interaction with balloons and ducks in the title screen.
```
Also needs to be put under the appropriate version.

src/openrct2-ui/interface/ViewportInteraction.cpp
XplosiveLugnut: Heh, I suppose.

OK, so, the only thing the assignment can affect is the _window_title_command_editor_tool_down_ function.
This function doesn't use the return value, only the assigned value.

However, the Title Command Editor isn't usable in the title screen, so this doesn't matter.

In the other two cases where _viewport_interaction_get_item_left_ is used, returning _VIEWPORT_INTERACTION_ITEM_NONE_ causes the calling function to do nothing but return and never use the assigned value. So the lack of assignment doesn't matter here either.

TLDR: There _should_ be no noticeable difference between having or not having the assignment, since the only function that it could affect can't be called while in the title screen.

Gymnasiast:
@XplosiveLugnut Could you rebase the branch? It seems you merged develop into your branch, which you should never do (or at least not in our project) - just rebase.

XplosiveLugnut:
@Gymnasiast Yes, I'll rebase. Should I rebase onto the latest release? And when I rebase, is it possible for me to clean up the commits? I would think I could rebase and reset back a couple commits before pushing, correct?

Gymnasiast:
> Should I rebase onto the latest release? 

No, you should rebase onto the latest `develop`.

> And when I rebase, is it possible for me to clean up the commits? I would think I could rebase and reset back a couple commits before pushing, correct?

Yes. When you do a rebase, it will show a list of commits. You can then decide to squash them together, reword them, drop them altogether, ...

XplosiveLugnut:
Not sure I did that correctly, but I think I achieved the desired result after having done that merge by mistake.

src/openrct2-ui/interface/ViewportInteraction.cpp
Gymnasiast: I'd prefer a stricter check for balloons or ducks specifically. (I think we have functions like IsBalloon and IsDuck for that). While there aren't any other misc sprites one can interact with _now_, there is no guarantee there won't be in the future, and they might not be suitable for use in the title screen.

Gymnasiast:
@XplosiveLugnut It looks like the rebase went well. I do have some line notes.

distribution/changelog.txt
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#7648] "Enable all drawable track pieces" now enables more pieces for the Twister, Vertical and Air Powered Vertical coasters.
 - Feature: [#9029] Open doors with the tile inspector.
+- Feature: [#9614] Can now pop balloons and quack ducks in title screen.
```
Gymnasiast, line 5:
```suggestion
- Feature: [#9614] Allow popping balloons and quacking ducks in the title screen.
```

src/openrct2-ui/interface/ViewportInteraction.cpp
Gymnasiast: Still, doing it like this (which is more in line with how the code was) shouldn't cause problems, right?

```suggestion
            return info->type = VIEWPORT_INTERACTION_ITEM_NONE;
```

PR #9616, opened 2019-07-20
Fix #9574: Text overflow in scenario objective window

### LINKED ISSUE #9574: Korean texts are overflow in the scenario objective window
**OS:** Any OS
**Version:** 0.2.3
**Commit/Build:** 6eda049

Scenario objective tab(window) can't display whole korean texts when the description is very long as you can see below (since korean text's line height is wider than english)
I think it will be great if the window's size could be longer or flexible.

image

AaronVanGeffen:
As CJK languages use bigger font sizes, the objective window sometimes wasn't high enough to accommodate texts. This PR increases the window height slightly if TTF fonts are enabled. The 'enter name' button also sees a height increase to better accomodate its caption.

Before:
Screenshot_20190720_131952

After:
Screenshot_20190720_132511

Aside from the button height, English is unaffected:
Screenshot_20190720_132546

### DIFF
--- a/src/openrct2-ui/windows/Park.cpp
+++ b/src/openrct2-ui/windows/Park.cpp
@@ -124,7 +124,7 @@ static rct_widget window_park_stats_widgets[] = {
 
 static rct_widget window_park_objective_widgets[] = {
     MAIN_PARK_WIDGETS,
-    { WWT_BUTTON,           1,  7,      222,    209,    220,    STR_ENTER_NAME_INTO_SCENARIO_CHART,         STR_NONE },             // enter name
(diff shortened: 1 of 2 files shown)


src/openrct2-ui/windows/Park.cpp
```diff
@@ -1498,7 +1498,10 @@ static void window_park_objective_mouseup(rct_window* w, rct_widgetindex widgetI
  */
 static void window_park_objective_resize(rct_window* w)
 {
-    window_set_resize(w, 230, 226, 230, 226);
+    if (gCurrentTTFFontSet == nullptr)
```
Gymnasiast: This fails in DISABLE_TTF builds. In these builds, it should (obviously) always assume the sprite font.

PR #9618, opened 2019-07-20
Use enum class for RCT2_SOUND

duncanspumpkin:
Continuing on from #9560. This switches to an enum class for all of the sounds.

Has not been clang formatted yet.

### DIFF
--- a/src/openrct2/audio/AudioMixer.h
+++ b/src/openrct2/audio/AudioMixer.h
@@ -15,7 +15,7 @@
 #define MIXER_LOOP_NONE 0
 #define MIXER_LOOP_INFINITE (-1)
 
-enum class RCT2Sound : uint8_t;
+enum class SoundId : uint8_t;
 
 enum MIXER_GROUP
 {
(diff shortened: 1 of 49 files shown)


src/openrct2/audio/AudioMixer.h
```diff
@@ -10,11 +10,14 @@
 #pragma once
 
 #include "../common.h"
+#include "audio.h"
 
 #define MIXER_VOLUME_MAX 128
 #define MIXER_LOOP_NONE 0
 #define MIXER_LOOP_INFINITE (-1)
 
+enum class RCT_SOUND : uint32_t;
```
IntelOrca: What is this?

src/openrct2/audio/audio.h
```diff
@@ -78,74 +80,77 @@ struct rct_vehicle_sound_params
     uint16_t priority;
 };
 
-enum RCT2_SOUND
+enum class RCT2Sound : uint8_t
```
IntelOrca: We can remove RCT2, perhaps just call it `SoundType` or `SoundId` or `SoundKind`?

src/openrct2/audio/AudioMixer.h
duncanspumpkin: That is a mistake. :(

src/openrct2/audio/audio.h
duncanspumpkin: It is sort of tied to RCT2 though. If we ever wanted to add more it would break saves. I would probably call it `SoundId`.

duncanspumpkin: Changed to `SoundId`

PR #9622, opened 2019-07-21
Remove user string pool

IntelOrca:
Continuation of #9589 but this time for ride names and everything else.

PR to check CI pass.

This PR will form a single changeset to remove the user string pool from OpenRCT2. It will create one only to export to SV6 where there is a restriction of 1024 custom strings. These will get stored in the following order of priority: park name, ride names, banners, staff, guests with notifications, all other guests.

### DIFF
--- a/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
+++ b/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
@@ -1181,7 +1181,8 @@ static void window_editor_objective_options_rides_scrollpaint(rct_window* w, rct
         }
 
         // Ride name
-        gfx_draw_string_left(dpi, stringId, &ride->name, COLOUR_BLACK, 15, y);
+        ride->FormatNameTo(gCommonFormatArgs);
+        gfx_draw_string_left(dpi, stringId, gCommonFormatArgs, COLOUR_BLACK, 15, y);
     }
 }
 
--- a/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
+++ b/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
@@ -1144,9 +1144,6 @@ static void window_editor_objective_options_rides_paint(rct_window* w, rct_drawp
  */
 static void window_editor_objective_options_rides_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32_t scrollIndex)
 {
-    rct_string_id stringId;
-    Ride* ride;
-
     int32_t colour = ColourMapA[w->colours[1]].mid_light;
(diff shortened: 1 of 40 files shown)


src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -1181,7 +1181,8 @@ static void window_editor_objective_options_rides_scrollpaint(rct_window* w, rct
         }
 
         // Ride name
-        gfx_draw_string_left(dpi, stringId, &ride->name, COLOUR_BLACK, 15, y);
+        ride->FormatNameTo(gCommonFormatArgs);
```
Gymnasiast: Would it be possible to use a local variable here, rather than the global gCommonFormatArgs?

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -496,8 +496,27 @@ void S6Exporter::ExportRide(rct2_ride* dst, const Ride* src)
 
     // pad_046;
     dst->status = src->status;
-    dst->name = src->name;
-    dst->name_arguments = src->name_arguments;
+    if (src->custom_name.empty())
+    {
+        // Default name with number
+        dst->name = RideNaming[src->type].name;
```
Gymnasiast: This will (among other things) cause names like "Shop 1", "Shop 2", etc in exported saves.

IntelOrca: While this is true, using allocated string ids (which is required to get the object name) relies on vanilla allocating the same string IDs to those object names right?

src/openrct2/ride/Ride.cpp
```diff
@@ -7943,3 +7868,57 @@ void ride_clear_leftover_entrances(Ride* ride)
...
+            {
+                rideTypeName = rideEntry->naming.name;
+            }
+        }
+        else if (RideGroupManager::RideTypeHasRideGroups(type))
+        {
+            auto rideEntry = GetRideEntry();
+            if (rideEntry != nullptr)
+            {
+                auto rideGroup = RideGroupManager::GetRideGroup(type, rideEntry);
```
Gymnasiast: This can return a nullptr.

src/openrct2/rct2/S6Exporter.cpp
Gymnasiast: Yes, but so far that _has_ worked correctly.

Gymnasiast: I have just done a test. It works correctly if using a name from RideNaming or from a rideEntry, but if it is named after a ride group, it error trappers.

IntelOrca: Yeah, well we can basically deprecate compatibility with RCT2 vanilla at this point. We will soon be removing it altogether.

PR #9623, opened 2019-07-21
Remove user string pool

IntelOrca:
Continuation of #9589 but this time for ride names and everything else.

This PR removes the user string pool from OpenRCT2. It will create one only to export to SV6 where there is a restriction of 1024 custom strings.


Gymnasiast:
This is quite a big PR to do in one go. I'll try to finish reviewing it tomorrow. Will also test it.

src/openrct2-ui/windows/Player.cpp
```diff
@@ -475,7 +475,6 @@ void window_player_statistics_close(rct_window* w)
 {
     if (w->error.var_480)
     {
-        user_string_free(w->error.var_480);
```
Gymnasiast: Is this sufficient, considering the rest of `var_480` does not appear to have been touched?

IntelOrca: I don't know where this code even came from. This variable is for the error window timer and has nothing to do with this custom window. It is not used anywhere else in this custom window, so I might as well delete this whole close method.

ZehMatt:
This seems to break the network load/save test, have you looked into that yet?

IntelOrca:
Not yet, I haven't had any experience with these tests.

Gymnasiast:
This appears to work 99% correctly. The only issue that I found was that if you load enclosed save and move to the bottom left, the three banners under Noah's Ark briefly say "North Car Park" in purple before switching to their correct "Entrance" in yellow.

[Blackpool Pleasure Beach-testnamesave.zip](https://github.com/OpenRCT2/OpenRCT2/files/3443748/Blackpool.Pleasure.Beach-testnamesave.zip)
Blackpool Pleasure Beach

Gymnasiast:
Another bug: changing the text colour of a path banner does not work.

IntelOrca:
@Gymnasiast banner colours are screwed up because I have chosen to remove the text colour from the beginning of the user string and I haven't done a good job.

Gymnasiast:
Another thing: when modifying a path banner name, the colour code is copied into the text editing window. (It can even be deleted by holding backspace until the box is empty.)

IntelOrca:
@Gymnasiast I have improved things, would be able to test again?

Gymnasiast:
Sure, but it will be tomorrow evening. Need to get up early.

Gymnasiast:
The problem I mentioned earlier are gone.

I did notice another: the (frame) colour of path banners is not saved properly.

IntelOrca:
@Gymnasiast thanks, I have fixed that now

Gymnasiast:
Here is a patch for the Xcode project. I couldn't push from my macOS VM and it's getting late. With this, compilation in Xcode succeeds.

[xcode-user-strings-patch.zip](https://github.com/OpenRCT2/OpenRCT2/files/3448959/xcode-user-strings-patch.zip)

Gymnasiast:
I'll take it for one more test tonight.

Gymnasiast:
OK, tested it, but found another problem: the Status view of the ride list should list the rides in alphabetical order, but doesn't do so. Clicking the `Sort` button does not do anything.

IntelOrca:
@Gymnasiast thanks, I have fixed that now and the fact it did not display the status

duncanspumpkin:
This is not freeing banners anymore on removal. Banners have to have their type set to null for them to be freed. Causes #11523

PR #9629, opened 2019-07-22
Fix UTF-8 support on Windows

IntelOrca:
* Allow UTF-8 to be read / written to and from console. UTF-8 paths are now logged correctly to stdout when using `std::printf`, `std::fprintf`. Update some of the verbose logging to demonstrate that.
* Remove old `utf8_to_widechar` and `widechar_to_utf8` convert methods. These were our own sloppy implementation which didn't work for higher unicode characters like emoji. Everything now uses Windows API or libICU.
* Rename `String::ToUtf16` to `String::ToWideChar`. `wchar_t` is not always 16-bits. On Linux, `wchar_t` is typically 32 bits, hence UTF-32 not UTF-16.

### DIFF
--- a/src/openrct2-dll/openrct2-dll.cpp
+++ b/src/openrct2-dll/openrct2-dll.cpp
@@ -43,6 +43,8 @@ DLLEXPORT int LaunchOpenRCT2(int argc, wchar_t** argvW)
         return -1;
     }
 
+    SetConsoleCP(CP_UTF8);
+    SetConsoleOutputCP(CP_UTF8);
     int exitCode = NormalisedMain(argc, const_cast<const char**>(argv));
 
     FreeCommandLineArgs(argc, argv);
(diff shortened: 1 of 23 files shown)


IntelOrca:
> where the native functions take utf8

They do not take UTF-8 and aren't defined to in the C lib or posix spec. It all depends on the system or filesystem encoding.

Gymnasiast:
If I write a file to a FAT32 drive using OpenRCT2 on Linux, it will be written with the correct encoding. _Something_ takes care of it, if it's not the functions that OpenRCT2 calls themselves then it might be the filesystem driver, or even something else. So whether they're defined to take UTF-8 or not in the C or POSIX specs is not really relevant - you have to write UTF-8 to them either way.

IntelOrca:
The Linux filesystem drivers for FAT32 and NTFS will handle that and probably assume UTF-8 encoding.

Gymnasiast:
Exactly, so between that and every single Linux distro having used UTF-8 for ages, it means you always have to write UTF-8. To me, that counts as the native functions taking UTF-8. Anything else is just a technicality, as far as I'm concerned.

PR #9632, opened 2019-07-23
Fix #9625: Show correct cost in scenery selection

### LINKED ISSUE #9625: Some object's expected cost is estimated as $0
**OS:** Windows 10
**Version:** 0.2.3
**Commit/Build:** 5342a6d

The expected cost of some object that positioned in the bottom of the scenary window is estimated as $0. I think the only large objects' cost seems to be estimated as $0.

**Steps to reproduce:**
1. Open a scenary window
2. Select some object, especially large object such as mine buildings
3. Hover mouse cursor into main game view
4. See the scenary window's bottom.

**Screenshots / Video:**
See the "Cost:", which is positioned in the right bottom of scenary window.

ZehMatt:
While some actions did set the cost some did not, I've slightly refactored the part also for consistency. This does not require a network bump, its a client side thing.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,6 +1,7 @@
 0.2.3+ (in development)
 ------------------------------------------------------------------------
 - Improved: [#9466] Add the rain weather effect to the OpenGL renderer.
+- Fix: [#9625] Show correct cost in scenery selection.
 
 0.2.3 (2019-07-10)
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
I want to refactor the place scenery in general its a bit of a mess and now with game actions I think we can create a more consistent solution for all of the different place actions.

PR #9636, opened 2019-07-23
Make socket status atomic

janisozaur:
`TcpSocket::ConnectAsync` will call `Connect` and modify `_status` from
a different thread and must ensure it doesn't introduce a data race.

### DIFF
--- a/src/openrct2/network/Socket.cpp
+++ b/src/openrct2/network/Socket.cpp
@@ -197,7 +197,7 @@ private:
 class TcpSocket final : public ITcpSocket, protected Socket
 {
 private:
-    std::atomic<SOCKET_STATUS> _status = SOCKET_STATUS_CLOSED;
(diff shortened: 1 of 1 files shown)


src/openrct2/network/Socket.cpp
```diff
@@ -196,7 +197,7 @@ class Socket
 class TcpSocket final : public ITcpSocket, protected Socket
 {
 private:
-    SOCKET_STATUS _status = SOCKET_STATUS_CLOSED;
+    std::atomic<SOCKET_STATUS> _status = SOCKET_STATUS_CLOSED;
```
ZehMatt:
```suggestion
    std::atomic<SOCKET_STATUS> _status = ATOMIC_VAR_INIT(SOCKET_STATUS_CLOSED);
```

PR #9637, opened 2019-07-23
Remove conversion from UTF-8 to widechar and back again

Gymnasiast:
I have tested this using non-ASCII characters like é, writing to both an EXT4 and FAT32 filesystem. This all appears to work correctly.


Gymnasiast:
Besides this, I also tested this by creating a folder called `💩` and moving the RCT2 files over there. This works correctly (also checked if it correctly write `game_path = "/home/michael/💩/RollerCoaster Tycoon 2"` to `config.ini`, which it did).

IntelOrca:
This should fix fix #7883, so you can change the commit message to say that and update the changelog.

PR #9639, opened 2019-07-23
Limit scrolling text colour to valid options

janisozaur:
[scrollingtext.sv6.gz](https://github.com/OpenRCT2/OpenRCT2/files/3424222/scrollingtext.sv6.gz)

### DIFF
--- a/src/openrct2/drawing/ScrollingText.cpp
+++ b/src/openrct2/drawing/ScrollingText.cpp
@@ -153,6 +153,10 @@ static int32_t scrolling_text_get_matching_or_oldest(rct_string_id stringId, uin
 static uint8_t scrolling_text_get_colour(uint32_t character)
 {
     int32_t colour = character & 0x7F;
+    if (colour >= COLOUR_COUNT)
+    {
+        colour = COLOUR_BLACK;
+    }
     if (character & COLOUR_FLAG_TRANSLUCENT)
     {
         return ColourMapA[colour].light;


src/openrct2/drawing/ScrollingText.cpp
```diff
@@ -153,6 +153,10 @@ static int32_t scrolling_text_get_matching_or_oldest(rct_string_id stringId, uin
 static uint8_t scrolling_text_get_colour(uint32_t character)
 {
     int32_t colour = character & 0x7F;
+    if (colour >= COLOUR_COUNT)
```
ZehMatt:
```suggestion
    colour = std::min<int32_t>(colour, COLOUR_COUNT);
```

janisozaur: Why though? I would rather think it should be a specific colour, rather than "last one".

ZehMatt: We could also use modulo to be mean to those corrupt parks. Dunno, I don't like defaulting to black i guess?

ZehMatt: Also where is this park coming from, how did this value get corrupted, is the corruption still a thing in our code, etc. Thats why I think it would be more apparent to not simply go black if its still something that happens.

janisozaur: This was some multiplayer park. When connecting to the server hosting it, I got immediate desync, but haven't tried hosting it myself to reproduce the issue. No idea how it got corrupted, but it was being hosted yesterday, so maybe related to utf8 stuff?

ZehMatt: Does text somehow affect sprites? Like I can imagine a desync would happen if two systems had different sprites because of different fonts? The whole font stuff is very complex and I barely understand it.

janisozaur: I doubt it does.

ZehMatt: I mean does the desync have anything to do with the color at all?

janisozaur: I understood what you meant and I doubt it does

Gymnasiast: It needs to default to _something_. Black is fine IMO.

PR #9644, opened 2019-07-25
Refactor banners

IntelOrca:
Pulling out the banner refactor from #9623.

* Changes `rct_banner` to `Banner`.
* Isolates `gBanners` to just `get_banner`.

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -324,7 +324,7 @@ int32_t viewport_interaction_get_item_right(int32_t x, int32_t y, viewport_inter
             sceneryEntry = tileElement->AsWall()->GetEntry();
             if (sceneryEntry->wall.scrolling_mode != SCROLLING_MODE_NONE)
             {
-                auto banner = &gBanners[tileElement->AsWall()->GetBannerIndex()];
+                auto banner = tileElement->AsWall()->GetBanner();
                 set_map_tooltip_format_arg(0, rct_string_id, STR_MAP_TOOLTIP_BANNER_STRINGID_STRINGID);
                 set_map_tooltip_format_arg(2, rct_string_id, banner->string_idx);
                 set_map_tooltip_format_arg(4, rct_string_id, STR_MAP_TOOLTIP_STRINGID_CLICK_TO_MODIFY);
@@ -337,7 +337,7 @@ int32_t viewport_interaction_get_item_right(int32_t x, int32_t y, viewport_inter
             sceneryEntry = tileElement->AsLargeScenery()->GetEntry();
             if (sceneryEntry->large_scenery.scrolling_mode != SCROLLING_MODE_NONE)
             {
-                auto banner = &gBanners[tileElement->AsLargeScenery()->GetBannerIndex()];
+                auto banner = tileElement->AsLargeScenery()->GetBanner();
                 set_map_tooltip_format_arg(0, rct_string_id, STR_MAP_TOOLTIP_BANNER_STRINGID_STRINGID);
                 set_map_tooltip_format_arg(2, rct_string_id, banner->string_idx);
                 set_map_tooltip_format_arg(4, rct_string_id, STR_MAP_TOOLTIP_STRINGID_CLICK_TO_MODIFY);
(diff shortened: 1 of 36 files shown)


src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -325,7 +324,7 @@ int32_t viewport_interaction_get_item_right(int32_t x, int32_t y, viewport_inter
             sceneryEntry = tileElement->AsWall()->GetEntry();
             if (sceneryEntry->wall.scrolling_mode != SCROLLING_MODE_NONE)
             {
-                banner = &gBanners[tileElement->AsWall()->GetBannerIndex()];
+                auto banner = get_banner(tileElement->AsWall()->GetBannerIndex());
```
Gymnasiast: Perhaps also use/create a `GetBanner()` function here?

src/openrct2/rct2/RCT2.h
```diff
@@ -32,6 +31,7 @@
 #define RCT2_MAX_RESEARCHED_RIDE_ENTRY_QUADS 8 // With 32 bits per uint32_t, this means there is room for 256 entries.
 #define RCT2_MAX_RESEARCHED_SCENERY_ITEM_QUADS 56
 #define RCT2_MAX_RESEARCHED_SCENERY_ITEMS (RCT2_MAX_RESEARCHED_SCENERY_ITEM_QUADS * 32) // There are 32 bits per quad.
+#define RCT2_MAX_BANNERS_IN_PARK 250
```
Gymnasiast: Why was this line moved?

src/openrct2/world/Banner.h
```diff
@@ -18,23 +18,17 @@ constexpr BannerIndex BANNER_INDEX_NULL = (BannerIndex)-1;
...
-    };
-    uint8_t text_colour; // 0x05
-    uint8_t x;           // 0x06
-    uint8_t y;           // 0x07
+    uint8_t flags;
+    rct_string_id string_idx;
+    uint8_t colour;
+    uint8_t ride_index;
+    uint8_t text_colour;
+    uint8_t x;
```
Gymnasiast: Perhaps change this to a TileCoordsXY, while we're at it. (Optional)

src/openrct2/rct2/RCT2.h
IntelOrca: hmm, I don't know

src/openrct2-ui/windows/Banner.cpp
```diff
@@ -133,8 +133,8 @@ rct_window* window_banner_open(rct_windownumber number)
     window_init_scroll_widgets(w);
 
     auto banner = get_banner(w->number);
-    int32_t view_x = banner->x << 5;
-    int32_t view_y = banner->y << 5;
+    int32_t view_x = banner->position.x;
+    int32_t view_y = banner->position.y;
```
Gymnasiast: You removed the `<< 5` by mistake.

src/openrct2/world/Banner.cpp
```diff
@@ -309,3 +335,12 @@ void BannerElement::ResetAllowedEdges()
 {
     flags |= 0b00001111;
 }
+
+Banner* get_banner(BannerIndex id)
```
ZehMatt: Shouldn't we call this GetBanner instead? I know that a lot of code still uses underscores but that is not what we want if I'm no mistaken.

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -348,7 +347,8 @@ int32_t viewport_interaction_get_item_right(int32_t x, int32_t y, viewport_inter
             break;
 
         case VIEWPORT_INTERACTION_ITEM_BANNER:
-            banner = &gBanners[tileElement->AsBanner()->GetIndex()];
+        {
+            auto banner = tileElement->AsBanner()->GetBanner();
```
ZehMatt: Looks a little bit funny, but I'm guessing thats because both have different storage.

IntelOrca: The tile elements should be postfixed with Element to make it clearer, so SurfaceTileElement, FootpathTileElement, BannerTileElement, etc.

ZehMatt: Makes sense, not requesting changes here, just odd looking.

PR #9661, opened 2019-07-27
Add lots of guards to track block get next / previous

IntelOrca:
In attempt to fix a lot of the backtrace issues containing track block get next / previous.


ZehMatt:
If you look at the screenshots in the attachments all of them have the new ghost train testing in common.

IntelOrca:
@ZehMatt I know, all callers check result of these functions though, so returning false on any kind of bad input arguments should be alright.

PR #9667, opened 2019-07-28
Fix #8136. Fix incorrect calculation of excessive lateral G penalty.

### LINKED ISSUE #8136: The penalties for excessive lateral G's are significantly lower in vanilla and classic
If you look at these screenshots you can see that the penalty for having at least 2.81 lateral G's is about 3.75 in OpenRCT2 while it's only about 2 in vanilla.
image
image

The same goes for the penalty for having at least 3.11 lateral G's. In OpenRCT2 you get an extra 8.5 intensity while in vanilla it seems to be about 4.6.
image
image

Here is also the save file if you want to see it for yourself.
[Lateral G penalties.zip](https://github.com/OpenRCT2/OpenRCT2/files/2501624/Lateral.G.penalties.zip)

The extra nausea you get from those penalties is different as well. Classic has the same penalties as vanilla.

I know that the order of the stat requirement penalties and lateral G penalties was reversed in openrct2, but that shouldn't have an influence here since there are no stat requirement penalties.

duncanspumpkin:
Fix #8136. Fix incorrect calculation of excessive lateral G penalty.

This penalty had been modified for OpenRCT2 moving it from before certain other penalties were applied to after. The move accidently removed the ride type multiplier which caused a very large penalty to be applied. In addition the excitement penalty applied to the whole calculation instead of just the gforce bonus. This caused an excessive penalty to the excitement.

Below pictures are from the new calculations.

image

image

### DIFF
--- a/src/openrct2/ride/RideRatings.cpp
+++ b/src/openrct2/ride/RideRatings.cpp
@@ -1629,18 +1629,44 @@ static void ride_ratings_apply_max_lateral_g_penalty(
     }
 }
 
-static void ride_ratings_apply_excessive_lateral_g_penalty(rating_tuple* ratings, Ride* ride)
+static rating_tuple ride_ratings_get_excessive_lateral_g_penalty(Ride* ride)
 {
-#ifndef ORIGINAL_RATINGS
+    rating_tuple result = { 0 };
     if (ride->max_lateral_g > FIXED_2DP(2, 80))
     {
-        ride_ratings_add(ratings, 0, FIXED_2DP(3, 75), FIXED_2DP(2, 00));
+        result.intensity = FIXED_2DP(3, 75);
+        result.nausea = FIXED_2DP(2, 00);
     }
+
     if (ride->max_lateral_g > FIXED_2DP(3, 10))
     {
-        ratings->excitement /= 2;
-        ride_ratings_add(ratings, 0, FIXED_2DP(8, 50), FIXED_2DP(4, 00));
+        // Remove half of the ride_ratings_get_gforce_ratings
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
Will break replays i suspect.

src/openrct2/ride/RideRatings.cpp
```diff
@@ -1629,18 +1629,44 @@ static void ride_ratings_apply_max_lateral_g_penalty(
     }
 }
 
-static void ride_ratings_apply_excessive_lateral_g_penalty(rating_tuple* ratings, Ride* ride)
+static rating_tuple ride_ratings_get_excessive_lateral_g_penalty(Ride* ride)
 {
-#ifndef ORIGINAL_RATINGS
+    rating_tuple result = { 0 };
```
AaronVanGeffen: This is on the stack. When returning a pointer to the tuple, it will go out of scope and may lead to corrupted memory being read at some point.

ZehMatt:
I would ignore replays for the moment, theres a PR where we start over anyway.

Gymnasiast:
Could you upload a save with the two, so that I can check it in vanilla?

duncanspumpkin:
Check the save in 8136.

Gymnasiast:
Right, it seems it was deliberately changed, because vanilla's algorithm would cause odd results elsewhere.

Well, if we have to pick, then matching vanilla's weirdness is better than having our own weirdness that vanilla doesn't have.

duncanspumpkin:
Yes this change now does exactly what the previous pr said it would do (but actually didn't). So now it will correctly penalise the correct rides. The original change was to fix the issue where successive penalties actually decreased this particular penalty.

Issue #9676, opened 2019-07-28
Ride Ratings Refactoring


duncanspumpkin:
Ride ratings (src/openrct2/ride/RideRatings.cpp) has a number of variables that are labelled by their register name or by a hex number. All of these variables are now known and should be refactored to explain things easier.

For example:
```
    rating_tuple var_10E_rating = get_flat_turns_rating(ride);
    excitement += var_10E_rating.excitement;
    intensity += var_10E_rating.intensity;
    nausea += var_10E_rating.nausea;

    rating_tuple var_110_rating = get_banked_turns_rating(ride);
    excitement += var_110_rating.excitement;
    intensity += var_110_rating.intensity;
    nausea += var_110_rating.nausea;
```
`var_10E_rating` is now known to be used for flat turns. So this variable should be `flatTurnsRating`. There are numerous cases very similar to this throughout this source file.

In addition make sure all local variables are in lower camelCase ~~and all function names are in upper CamelCase.~~ **Don't change the function names at this time. They will renamed later when they are refactored into a more cpp structure.** https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style

When commiting in git try to stick to the commit message format. https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

jgottula:
I'll take a stab at this. Will post a PR when I have something ready.

PR #9697, opened 2019-07-31
Refactor #9676: Ride Ratings Refactoring

jgottula:
I wasn't 100% clear on the appropriate naming conventions for function-static const tables and the like, so I went with my best guess based on similar situations elsewhere in the codebase.

I threw in a couple additional minor cleanups (compound assignment operators, `std::clamp`) that weren't requested in the original issue, but which looked like they ought to be fixed up when I encountered them while scanning through the code.

----

One thing that I noticed, but didn't address in this PR (as the issue mentioned that function renaming will be handled separately later), is that there's annoying inconsistency in some of the function names: specifically `get_whatever_rating` vs `get_whatever_ratings`. All of the functions in question return a single `rating_tuple`, and I guess it's technically arguable whether such should be considered plural; but in any case it's inconsistent whether the `s` is there from one function to another.

### DIFF
--- a/src/openrct2/ride/RideRatings.cpp
+++ b/src/openrct2/ride/RideRatings.cpp
@@ -836,11 +836,11 @@ static uint16_t ride_compute_upkeep(Ride* ride)
     uint16_t upkeep = initialUpkeepCosts[ride->type];
 
     uint16_t trackCost = costPerTrackPiece[ride->type];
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Can I just say that I appreciate the very clean separation over five commits? It made reviewing this PR a breeze. :+1:

PR #9705, opened 2019-08-01
Refactor map animations and dynamically create on park load

IntelOrca:
More prep. for new save format. Map animations are not game state and therefore don't need to be stored in the save file. We can dynamically create them again on load.

### DIFF
--- a/src/openrct2/Context.cpp
+++ b/src/openrct2/Context.cpp
@@ -546,6 +546,7 @@ namespace OpenRCT2
                         gCurrentLoadedPath = path;
                         gFirstTimeSaving = true;
                         game_fix_save_vars();
+                        AutoCreateMapAnimations();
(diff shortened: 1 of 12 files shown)


Gymnasiast:
I would like to test this park against a scenery-heavy/hack-heavy NE park, just to see if it doesn't break anything.

AaronVanGeffen:
Nice to see this. Should also help restore animations in parks/saves that have somehow had their animation list corrupted.

Just to check: this won't affect jumping fountain states, right?

IntelOrca:
@AaronVanGeffen jumping fountains are sprites and therefore invalidate on their own update routine. Map animations are purely for animated tile elements and it only ensures invalidate is called repeatedly on the tile. In OpenGL this will make little difference as there is no dirty drawing.

PR #9708, opened 2019-08-01
[RFC] Change ride list to a table

IntelOrca:
This is an attempt to create common code for implementing list views which provide headers, highlighting, selecting and sorting. I have done it with the ride list so far.

Based on how native Win32 list views work, a window listens for list view messages (a new window event handler) where it returns the data requested by the window manager. It will receive a message (item index, column index) for every visible cell and the handler will populate an argument buffer. The window no longer needs to handle the input or painting, it just provides the data to show.

At the moment I have hacked this into the scroll view events. I have tried to keep changes to the window system as minimal as possible but at the same time add initial plumbing to start making all the windows into sub classes of `rct_window`.

As for the ride list table, some columns are too narrow, what option is best?
* Use a good icon and tooltip.
* Add several views which show different wider columns.
* Double the height of each row and move status to underneath the ride name. (See how Locomotion does its vehicle list window).

image

### DIFF
--- a/src/openrct2-ui/interface/Window.cpp
+++ b/src/openrct2-ui/interface/Window.cpp
@@ -87,7 +87,7 @@ static bool window_fits_on_screen(int32_t x, int32_t y, int32_t width, int32_t h
 
 rct_window* window_create(
     int32_t x, int32_t y, int32_t width, int32_t height, rct_window_event_list* event_handlers, rct_windowclass cls,
-    uint16_t flags)
+    uint16_t flags, std::unique_ptr<rct_window> newWindow)
 {
     // Check if there are any window slots left
     // include WINDOW_LIMIT_RESERVED for items such as the main viewport and toolbars to not appear to be counted.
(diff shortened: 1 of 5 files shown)


Gymnasiast:
This is a nice little addition, but it feels a little bit cluttered. It would be nice if it was an optional view, or (but that's hard to realise with our current window station) if one could choose which columns to show.

Gymnasiast:
Perhaps we should leave this kind of thing to a plug-in, making sure that it can do everything this PR does?

tupaschoal:
@IntelOrca do we intend on merging this? Is it better off as a plugin?

pizza2004:
Unless plugins can edit the standard windows I think it would make more sense to try and add the ability to choose which columns show, like was mentioned in another comment, but that would be something to create a new issue about.

AaronVanGeffen:
@pizza2004 You're commenting on a closed PR. Feel free to open a new issue for the feature you're proposing.

pizza2004:
I understand where I'm commenting, I simply didn't want to needlessly open an issue if it was something that would be deemed unnecessary.  It seems wasteful.

PR #9742, opened 2019-08-06
Fix 9729

duncanspumpkin:
Change peep pathfinding to take into account the height difference when deciding which entrance to pathfind to.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -7,8 +7,7 @@
 - Fix: [#9625] Show correct cost in scenery selection.
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
-- Fix: [#9729] Peeps do not take into account height difference when
-deciding to pathfind to a ride entrance. (original bug)
+- Fix: [#9729] Peeps do not take into account height difference when deciding to pathfind to a ride entrance. (original bug)
 - Improved: [#9466] Add the rain weather effect to the OpenGL renderer.
 
 0.2.3 (2019-07-10)
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -7,6 +7,8 @@
 - Fix: [#9625] Show correct cost in scenery selection.
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
+- Fix: [#9729] Peeps do not take into account height difference when
+deciding to pathfind to a ride entrance. (original bug)
 - Improved: [#9466] Add the rain weather effect to the OpenGL renderer.
 
 0.2.3 (2019-07-10)
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -2054,7 +2054,7 @@ int32_t guest_path_finding(Guest* peep)
     /* Find the ride's closest entrance station to the peep.
      * At the same time, count how many entrance stations there are and
      * which stations are entrance stations. */
-    uint16_t closestDist = 0xFFFF;
+    uint16_t bestScore = 0xFFFF;
     uint8_t closestStationNum = 0;
 
     int32_t numEntranceStations = 0;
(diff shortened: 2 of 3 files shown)


src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2073,11 +2073,11 @@ int32_t guest_path_finding(Guest* peep)
 
         int16_t stationX = (int16_t)(entranceLocation.x * 32);
         int16_t stationY = (int16_t)(entranceLocation.y * 32);
-        uint16_t dist = abs(stationX - peep->next_x) + abs(stationY - peep->next_y);
-
-        if (dist < closestDist)
+        uint16_t score = abs(stationX - peep->next_x) + abs(stationY - peep->next_y);
```
ZehMatt: I think distance makes still more sense in this context than score, its looking for the smallest distance here.

distribution/changelog.txt
```diff
@@ -7,6 +7,8 @@
 - Fix: [#9625] Show correct cost in scenery selection.
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
+- Fix: [#9729] Peeps do not take into account height difference when
+deciding to pathfind to a ride entrance. (original bug)
```
IntelOrca: Is this line break intentional?

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2073,11 +2073,11 @@ int32_t guest_path_finding(Guest* peep)
...
+        score += abs(entranceLocation.z - peep->next_z) * 2;
```
ZehMatt: Any reason only z is multiple of two? If I recall correctly calculating the distance between 2 points in 3d space goes something like
```
sqrt(pow((x - x2), 2) + pow((y - y2), 2) + pow((z - z2), 2));
```
Given we want integers we can just skip sqrt

duncanspumpkin: Its to match the other calculation of the score.

distribution/changelog.txt
duncanspumpkin: no

src/openrct2/peep/GuestPathfinding.cpp
Broxzier: Perhaps because peeps walk slower on sloped paths? A comment to explain the logic behind the equations would be helpful.

ZehMatt: Instead of score/distance I would also consider this a "cost" if the effort is included in the mix.

duncanspumpkin: Its to bring it more inline with this calc.
```
        // Calculate the heuristic score of this map element.
        uint16_t x_delta = abs(gPeepPathFindGoalPosition.x - loc.x) * 32;
        uint16_t y_delta = abs(gPeepPathFindGoalPosition.y - loc.y) * 32;
        if (x_delta < y_delta)
            x_delta >>= 4;
        else
            y_delta >>= 4;
        uint16_t new_score = x_delta + y_delta;
        uint16_t z_delta = abs(gPeepPathFindGoalPosition.z - loc.z);
        z_delta <<= 1;
        new_score += z_delta;
```

ZehMatt: So we have essentially duplicate code for calculating a cost between two points in 3d space where Z has a bigger impact on the cost.

IntelOrca: I always assumed it did Manhattan distance for performance reasons.

ZehMatt: As for path finding what the actual metric is doesn't matter, its just a cost/weight value. Question was more about the multiple of two, if the reason is for making it consistent with the other code then we should deduplicate that logic and perhaps mention that slopes are more expensive cost wise in the function (assuming that is the reason behind it)

duncanspumpkin: Okay. Its now been changed into just the one score generator for the two functions.

src/openrct2/peep/GuestPathfinding.cpp
duncanspumpkin: imho score matches the rest of the pathfinding file.

ZehMatt:
I was wondering something, shouldn't the CI fail the replay test after this change? I wonder if they currently work.

PR #9743, opened 2019-08-06
Fix #8584: Duck spawn search now uses entire map

### LINKED ISSUE #8584: Ducks are not spawning!
**OS:** Windows 10
**Version:** 0.2.1
**Commit/Build:** 746feaa

Ducks no longer seem to spawn!

**Steps to reproduce:**
1. Create large body of water (code says you need at least 25 tiles of water at same height in a 7x7 area)
2. Wait forever
3. No ducks seem to spawn

ZehMatt:
This one should be without crashes. I left a few asserts in there to be sure and they have no performance impact.

### DIFF
--- a/src/openrct2/scenario/Scenario.cpp
+++ b/src/openrct2/scenario/Scenario.cpp
@@ -435,12 +435,15 @@ static int32_t scenario_create_ducks()
     if (centreWaterZ == 0)
         return 0;
 
-    // Check 7x7 area around centre tile
-    CoordsXY innerPos{ centrePos.x - (32 * 3), centrePos.y - (32 * 3) };
+    // Check NxN area around centre tile defined by SquareSize
+    constexpr int32_t SquareSize = 7;
+    constexpr int32_t SquareCentre = SquareSize / 2;
+
+    CoordsXY innerPos{ centrePos.x - (32 * SquareCentre), centrePos.y - (32 * SquareCentre) };
     int32_t waterTiles = 0;
-    for (int32_t y = 0; y < 7; y++)
+    for (int32_t y = 0; y < SquareSize; y++)
     {
-        for (int32_t x = 0; x < 7; x++)
+        for (int32_t x = 0; x < SquareSize; x++)
         {
             if (!map_is_location_valid(innerPos))
                 continue;
@@ -454,7 +457,7 @@ static int32_t scenario_create_ducks()
 
             innerPos.x += 32;
         }
-        innerPos.x -= 224;
+        innerPos.x -= SquareSize * 32;
         innerPos.y += 32;
     }
 
--- a/src/openrct2/scenario/Scenario.cpp
+++ b/src/openrct2/scenario/Scenario.cpp
@@ -416,70 +416,66 @@ void scenario_update()
     }
     scenario_update_daynight_cycle();
 }
-
 /**
  *
  *  rct2: 0x006744A9
  */
 static int32_t scenario_create_ducks()
 {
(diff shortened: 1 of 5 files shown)


src/openrct2/scenario/Scenario.cpp
```diff
@@ -467,7 +472,13 @@ static int32_t scenario_create_ducks()
         r = scenario_rand();
         x2 = (r >> 16) & 0x7F;
         y2 = (r & 0xFFFF) & 0x7F;
-        create_duck(x + x2 - 64, y + y2 - 64);
+
+        int32_t targetX = x + x2 - 64;
```
duncanspumpkin: Create a CoordsXY for this.

src/openrct2/scenario/Scenario.cpp
```diff
@@ -423,11 +424,10 @@ static int32_t scenario_create_ducks()
 {
     int32_t i, j, r, c, x, y, waterZ, centreWaterZ, x2, y2;
 
-    r = scenario_rand();
-    x = ((r >> 16) & 0xFFFF) & 0x7F;
-    y = (r & 0xFFFF) & 0x7F;
-    x = (x + 64) * 32;
-    y = (y + 64) * 32;
+    x = 64 + (scenario_rand_max(MAXIMUM_MAP_SIZE_PRACTICAL) * 32);
+    y = 64 + (scenario_rand_max(MAXIMUM_MAP_SIZE_PRACTICAL) * 32);
```
duncanspumpkin: CoordsXY this

IntelOrca:
If this gets rebased and changed to use coords which might be easier now after @duncanspumpkin's refactor, we can merge.

ZehMatt:
Sorry got a little distracted by other things, will do this today.

ZehMatt:
Applied the requested changes and cleaned up the code a little bit. Also I bumped up the network version since scenario_rand is now called twice.

src/openrct2/scenario/Scenario.cpp
```diff
@@ -415,60 +416,66 @@ void scenario_update()
...
             if (waterZ == centreWaterZ)
-                c++;
+                waterTiles++;
 
-            x2 += 32;
+            innerPos.x += 32;
         }
-        x2 -= 224;
-        y2 += 32;
+        innerPos.x -= 224;
```
duncanspumpkin: 7*32

Gymnasiast:
Could you rebase this and rebump the network version? I think it's time to get this merged, and Duncan's requested changes have been made anyway.

ZehMatt:
Rebased and incremented the network version

src/openrct2/scenario/Scenario.cpp
```diff
@@ -415,60 +416,69 @@ void scenario_update()
...
-
 /**
  *
  *  rct2: 0x006744A9
  */
 static int32_t scenario_create_ducks()
 {
-    int32_t i, j, r, c, x, y, waterZ, centreWaterZ, x2, y2;
+    CoordsXY centrePos;
+    centrePos.x = 64 + (scenario_rand_max(MAXIMUM_MAP_SIZE_PRACTICAL) * 32);
```
Gymnasiast: It looks like this shifts the area from being in the centre and skipping the outer ring of invisible tiles, to one tile in both x and y directions. AFAICS, this should be changed to `32 + ...`.

ZehMatt: You have to consider that it subtracts 64 some lines below so that would go negative

PR #9751, opened 2019-08-07
Create RCT2 version of ride rating calculation data

IntelOrca:


src/openrct2/rct2/RCT2.h
```diff
@@ -593,6 +593,26 @@ union RCT2Sprite
...
+    uint8_t state;
+    uint8_t proximity_track_type;
+    uint8_t proximity_base_height;
+    uint16_t proximity_total;
+    uint16_t proximity_scores[26];
+    uint16_t num_brakes;
+    uint16_t num_reversers;
+    uint16_t station_flags;
+};
+assert_struct_size(RCT2RideRatingCalculationData, 76);
```
ZehMatt: Is there a reason that this has to be strict?

IntelOrca: Yes... this is all used for loading SV6.

ZehMatt: Yeah nevermind, misread the code slightly. I thought there was another struct specific to s6/s4 so we wouldn't need to worry about packing given we copy field by field.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -745,6 +745,30 @@ class S6Importer final : public IParkImporter
...
+        dst.proximity_z = src.proximity_z;
+        dst.proximity_start_x = src.proximity_start_x;
+        dst.proximity_start_y = src.proximity_start_y;
+        dst.proximity_start_z = src.proximity_start_z;
+        dst.current_ride = src.current_ride;
+        dst.state = src.state;
+        dst.proximity_track_type = src.proximity_track_type;
+        dst.proximity_base_height = src.proximity_base_height;
+        dst.proximity_total = src.proximity_total;
+        for (size_t i = 0; i < std::size(dst.proximity_scores); i++)
```
Gymnasiast: As discussed on Gitter, this needs to take the source size to allow for making our own struct bigger.

PR #9791, opened 2019-08-08
Update replays to v0.0.4

ZehMatt:


ZehMatt:
Could need a hand here @Gymnasiast or @janisozaur, I'm having trouble figuring out the correct path for the CMake configuration, it doesn't seem to placing them in the correct directory.
I did a few attempts myself but couldn't quite figure out it, there are a few things I noticed:
1. The replays.zip contains already a folder called replays so it would probably unpack it into testdata\replays\replays if I'm not mistaken.
2. If I read the CMake file correctly its packing them into the userdata folder which is not the correct path I think, it should be <binary>/testdata/replays

IntelOrca:
> The replays.zip contains already a folder called replays so it would probably unpack it into testdata\replays\replays if I'm not mistaken.

Yes, which means you probably want to drop replays from the output directory in `openrct2.proj` as well.

ZehMatt:
It seems that its fine with windows builds, the replays are actually ran. Its currently specific to the CMake configuration

IntelOrca:
It should still probably be changed. Bit strange to have replays/replays.

ZehMatt:
Finally managed to get it working, now this looks good.

ZehMatt:
Just realized with network disabled it wouldn't actually run replays for code coverage. Now I think its ready. I'll gradually add more over time

Issue #6304, opened 2017-09-25
Multiplayer "Guest" feature


ZehMatt:
I would like to propose a new feature for the multiplayer part where next to spectators there would be a new group called guests. A guest is essentially a peep that each client would get as their own avatar and be able to freely roam the park. To make this a little more fun the player will not be able zoom out and will only see things that he has already explored, perhaps players buying a map will make everything visible at once. The controls of the peep will closely resemble existing RTS games as in: ordering the peep around will be done via RMB, interacting with buildings requires the player to select the building via LMB, etc,. The peep/avatar should be able to do anything the normal peeps would/could do, this also means that the peep/avatar will get hungry and thirsty over time, walking and riding coasters consumes energy. In order to provide a generic interface I would implement something called PeepController that will essentially drive the peeps, on top of instance there will be either the PlayerPeepController or AIPeepController that handles the logic for either user input or AI. Since the peep code is pretty messy this would be also my first thing on the TODO towards the feature. And in case the feature will be rejected we will at least have some clean code left. I hope this is the right place to bring this up and I would love to hear about what you think, so please share your thoughts.

janisozaur:
How that ties to https://github.com/OpenRCT2/OpenRCT2/issues/2896?

ZehMatt:
It would boil down to TwitchPeepController : PlayerPeepController

PR #9865, opened 2019-08-12
Multiplayer guests

ZehMatt:
Based on #6304


ZehMatt:
Now with plugin support I'll rather add changes required to be able to do this via scripting.

Issue #8376, opened 2018-12-06
Refactor gErrorType, gErrorStringId


ZehMatt:
Only the S6Importer currently uses them, they can be removed by directly showing the error in there. Relatively easy refactor.

kennycastro007:
Is this still open? I've been looking for issues to work on all night, so it'd be nice to find one still open!

ZehMatt:
You can give it a shot

kennycastro007:
What exactly are gErrorType and gErrorStringId doing? I see they're being set but I can't see where they're being used. I checked Diagnostics.h but still couldn't tell where they were being used once set

AaronVanGeffen:
They are both used in GameState.cpp.

kennycastro007:
I've finished the refactor, it still compiles just fine. What are the next steps to completing the issue? i.e How do I start the pull request? Can someone create the branch for the issue or do I have to do something else?

AaronVanGeffen:
Fork the repository and push it to a branch on your fork. Then create a pull request.

PR #9866, opened 2019-08-12
Fixed issue #8376

kennycastro007:
I thought about making the error reporting a function to avoid rewriting it 6 times but decided against it since I wasn't sure if it was a good idea and would've only been used in this one file

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -49,7 +49,8 @@
             },
             "compilerPath": "/usr/bin/clang",
             "cStandard": "c11",
-            "cppStandard": "c++17"
+            "cppStandard": "c++17",
+            "configurationProvider": "vector-of-bool.cmake-tools"
         },
         {
             "name": "Win32",
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -49,8 +49,7 @@
             },
             "compilerPath": "/usr/bin/clang",
             "cStandard": "c11",
-            "cppStandard": "c++17",
-            "configurationProvider": "vector-of-bool.cmake-tools"
+            "cppStandard": "c++17"
         },
         {
             "name": "Win32",
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -1546,19 +1546,52 @@ void load_from_sv6(const char* path)
     }
     catch (const ObjectLoadException&)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+
+        rct_string_id title_text = STR_UNABLE_TO_LOAD_FILE;
+        rct_string_id body_text = gErrorStringId;
+        if (gErrorType == ERROR_TYPE_GENERIC)
+        {
+            title_text = gErrorStringId;
+            body_text = 0xFFFF;
+        }
+        gErrorType = ERROR_TYPE_NONE;
+
+        context_show_error(title_text, body_text);
     }
     catch (const IOException& loadError)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_GAME_SAVE_FAILED;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_GAME_SAVE_FAILED;
         log_error("Error loading: %s", loadError.what());
+
+        rct_string_id title_text = STR_UNABLE_TO_LOAD_FILE;
+        rct_string_id body_text = gErrorStringId;
+        if (gErrorType == ERROR_TYPE_GENERIC)
+        {
+            title_text = gErrorStringId;
+            body_text = 0xFFFF;
+        }
+        gErrorType = ERROR_TYPE_NONE;
+
+        context_show_error(title_text, body_text);
     }
     catch (const std::exception&)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+
+        rct_string_id title_text = STR_UNABLE_TO_LOAD_FILE;
+        rct_string_id body_text = gErrorStringId;
+        if (gErrorType == ERROR_TYPE_GENERIC)
+        {
+            title_text = gErrorStringId;
+            body_text = 0xFFFF;
+        }
+        gErrorType = ERROR_TYPE_NONE;
+
+        context_show_error(title_text, body_text);
     }
 }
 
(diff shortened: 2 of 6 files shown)


ZehMatt:
Nothing speaks against a static function for when you know you will have duplicate code.

kennycastro007:
I just changed the code to use a static function `show_error(uint8_t, const rct_string_id&)` to use instead, this would be better should the manner in which errors are handled here can be changed easily

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1528,6 +1528,20 @@ std::unique_ptr<IParkImporter> ParkImporter::CreateS6(IObjectRepository& objectR
     return std::make_unique<S6Importer>(objectRepository);
 }
 
+static void show_error(uint8_t gErrorType, const rct_string_id& gErrorStringId)
```
ZehMatt: We only prefix globals with g. We normally use camelCase for variable names.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1528,6 +1528,20 @@ std::unique_ptr<IParkImporter> ParkImporter::CreateS6(IObjectRepository& objectR
...
+{
+    rct_string_id title_text = STR_UNABLE_TO_LOAD_FILE;
```
ZehMatt: camelCase name

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1546,19 +1560,24 @@ void load_from_sv6(const char* path)
     }
     catch (const ObjectLoadException&)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+        show_error(gErrorType, gErrorStringId);
```
ZehMatt: Why not pass the directly the value to the parameter? You don't have to assign variables first.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1546,19 +1560,24 @@ void load_from_sv6(const char* path)
...
     }
     catch (const IOException& loadError)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_GAME_SAVE_FAILED;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_GAME_SAVE_FAILED;
         log_error("Error loading: %s", loadError.what());
+
+        show_error(gErrorType, gErrorStringId);
```
ZehMatt: Same as above

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1546,19 +1560,24 @@ void load_from_sv6(const char* path)
...
     }
     catch (const std::exception&)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+
+        show_error(gErrorType, gErrorStringId);
```
ZehMatt: Same as above

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1584,20 +1603,26 @@ void load_from_sc6(const char* path)
     }
     catch (const ObjectLoadException& loadError)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_GAME_SAVE_FAILED;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_GAME_SAVE_FAILED;
         log_error("Error loading: %s", loadError.what());
+
+        show_error(gErrorType, gErrorStringId);
```
ZehMatt: Same as above

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1584,20 +1603,26 @@ void load_from_sc6(const char* path)
...
     }
     catch (const IOException& loadError)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_GAME_SAVE_FAILED;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_GAME_SAVE_FAILED;
         log_error("Error loading: %s", loadError.what());
+
+        show_error(gErrorType, gErrorStringId);
```
ZehMatt: Same as above

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1584,20 +1603,26 @@ void load_from_sc6(const char* path)
...
     }
     catch (const std::exception&)
     {
-        gErrorType = ERROR_TYPE_FILE_LOAD;
-        gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+        uint8_t gErrorType = ERROR_TYPE_FILE_LOAD;
+        rct_string_id gErrorStringId = STR_FILE_CONTAINS_INVALID_DATA;
+
+        show_error(gErrorType, gErrorStringId);
```
ZehMatt: Same as above

src/openrct2/rct2/S6Importer.cpp
kennycastro007: Fixed it in a new commit

kennycastro007:
@ZehMatt Is that all the changes you wanted me to make?

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1528,6 +1528,20 @@ std::unique_ptr<IParkImporter> ParkImporter::CreateS6(IObjectRepository& objectR
     return std::make_unique<S6Importer>(objectRepository);
 }
 
+static void show_error(uint8_t errorType, const rct_string_id& errorStringId)
```
ZehMatt: Why did you choose to use ```const rct_string_id&``` here? You should pass small things like integers always by value, this is not performance critical code but its generally better to do.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1528,6 +1528,20 @@ std::unique_ptr<IParkImporter> ParkImporter::CreateS6(IObjectRepository& objectR
...
+{
+    rct_string_id titleText = STR_UNABLE_TO_LOAD_FILE;
+    rct_string_id bodyText = errorStringId;
+    if (errorType == ERROR_TYPE_GENERIC)
+    {
+        titleText = errorStringId;
+        bodyText = 0xFFFF;
+    }
+    errorType = ERROR_TYPE_NONE;
```
ZehMatt: This assignment seems redundant.

kennycastro007: It was left over from when I copy-pasted the common parts out, I've taken it out in my latest commit

src/openrct2/rct2/S6Importer.cpp
kennycastro007: I can see it being simpler to read and since the copy for ints is negligible, but is there another reason why it's better to pass by value? Also do you think the function should be inline?

ZehMatt:
Given that its your first time contributing you should add yourself to the contributors in 
 'contributors.md'

kennycastro007:
I went ahead and added myself to contributors.md, thanks for the reminder!

.vscode/c_cpp_properties.json
```diff
@@ -49,7 +49,8 @@
             },
             "compilerPath": "/usr/bin/clang",
             "cStandard": "c11",
-            "cppStandard": "c++17"
+            "cppStandard": "c++17",
+            "configurationProvider": "vector-of-bool.cmake-tools"
```
ZehMatt: Please revert this change, its unrelated.

kennycastro007:
Sorry about that, that was CMake Tools, I accidentally staged that change and pushed it in

PR #9872, opened 2019-08-13
Show the object download source when downloading object files

IntelOrca:
Requested by nedesigns. Progress bar now shows where the object file is being downloaded from.


janisozaur:
It should also print out the URL to stdout so it's possible to copy it

IntelOrca:
ok

PR #9874, opened 2019-08-14
Resolve #3154: Random Title Sequence

### LINKED ISSUE #3154: Random Title Sequence
It'd be cool to randomly select title sequence just like how you can randomly select music for it.

MichaelJBerk:
Added an option to use a random title sequence

### DIFF
--- a/src/openrct2/localisation/StringIds.h
+++ b/src/openrct2/localisation/StringIds.h
@@ -3873,12 +3873,14 @@ enum
 
     STR_STRING_STRINGID = 6329,
     STR_DOWNLOADING_OBJECTS_FROM = 6330,
+    
+    STR_OPTIONS_RANDOM_TITLE_SEQUENCE = 6330,
 
     STR_CREATE_DUCKS = 6331,
     STR_REMOVE_DUCKS = 6332,
 
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    STR_COUNT = 32768
+    STR_COUNT = 32770
 };
 
 #endif
--- a/src/openrct2/title/TitleScreen.cpp
+++ b/src/openrct2/title/TitleScreen.cpp
@@ -225,6 +225,13 @@ void TitleScreen::TitleInitialise()
     {
         _sequencePlayer = GetContext()->GetUiContext()->GetTitleSequencePlayer();
     }
+    if (gConfigInterface.random_title_sequence) {
+        int total = TitleSequenceManager::GetCount();
+        int random = rand() % total --;
+        const utf8* randomSequence = title_sequence_manager_get_name(random);
+        ChangePresetSequence(random);
+        
+    }
     size_t seqId = title_get_config_sequence();
     if (seqId == SIZE_MAX)
     {
--- a/src/openrct2/title/TitleScreen.cpp
+++ b/src/openrct2/title/TitleScreen.cpp
@@ -27,6 +27,7 @@
 #include "../scenario/Scenario.h"
 #include "../scenario/ScenarioRepository.h"
 #include "../ui/UiContext.h"
+#include "../util/Util.h"
 #include "TitleSequence.h"
 #include "TitleSequenceManager.h"
 #include "TitleSequencePlayer.h"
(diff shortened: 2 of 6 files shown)


janisozaur:
Any reason for closing this?

MichaelJBerk:
The builds failed testing, and I'm trying to fix it.

janisozaur:
Not a reason to close a PR

MichaelJBerk:
I have been playing around with the random title sequence and realized a problem: If you don't have RCT1 installed, and the random title screen is one from RCT1 (or expansion), then the game will go back to the default and display the regular RCT2 title screen. The result is that if, for example, you don't have any extra sequences installed, then there's a 80% chance you'll get the RCT2 title screen instead of the OpenRCT2 one (since there's three that can't be loaded, plus the RCT2 title) , when it really should be a 50% chance. Until there's a way to fix it, the feature won't work as intended unless you have RCT1 with all expansions installed. I'm going to close this PR for the time being so I can look into this and find a way to fix it.

duncanspumpkin:
Just switch it to a draft pr don't open and close. It creates tons of notifications for nothing

src/openrct2/localisation/StringIds.h
```diff
@@ -3967,9 +3967,11 @@ enum
 
     STR_STRING_STRINGID = 6329,
     STR_DOWNLOADING_OBJECTS_FROM = 6330,
+    
+    STR_OPTIONS_RANDOM_TITLE_SEQUENCE = 6331,
 
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    STR_COUNT = 32768
+    STR_COUNT = 32769
```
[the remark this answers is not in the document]
janisozaur: This is not the first time I see it modified, there should be a comment explaining things a bit.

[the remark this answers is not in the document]
ZehMatt: If the main purpose is to limit the maximum string count to 2^15 then we can simply use enum : int16_t and the compiler will throw a warning if it ever goes beyond.

MichaelJBerk:
I have fixed the issue by checking if there's any scenarios installed from RCT1 or its expansions, so if the game chooses a title sequence that isn't installed, it will try again until the random sequence is one that is installed.

src/openrct2/title/TitleScreen.cpp
```diff
@@ -225,6 +226,68 @@ void TitleScreen::TitleInitialise()
...
+        std::string RCT1AAString = format_string(STR_TITLE_SEQUENCE_RCT1_AA, gCommonFormatArgs);
+        std::string RCT1LLString = format_string(STR_TITLE_SEQUENCE_RCT1_AA_LL, gCommonFormatArgs);
+        //Ensure the random sequence chosen isn't from RCT1 or expansion if the player doesn't have it installed
+        while (!safeSequence) {
+            size_t total = TitleSequenceManager::GetCount();
+            random = util_rand() % (int32_t)total;
+            const utf8* scName = title_sequence_manager_get_name(random);
+            safeSequence = true;
+            if (scName == RCT1String) {
+                if (RCT1Installed) {
```
duncanspumpkin: No need for this nested ifs. Just do `safeSequence = RCTXXXInstalled;`

src/openrct2/title/TitleScreen.cpp
```diff
@@ -225,6 +226,68 @@ void TitleScreen::TitleInitialise()
...
+        
+        bool RCT1Installed = false, RCT1AAInstalled = false, RCT1LLInstalled = false;
+        int RCT1Count = 0, RCT1AACount = 0, RCT1LLCount = 0;
+        size_t scenarioCount = scenario_repository_get_count();
+        for (size_t s = 0; s < scenarioCount; s++ ) {
+            if (scenario_repository_get_by_index(s)->source_game == SCENARIO_SOURCE_RCT1) {
+                RCT1Count++;
+            }
+            if (scenario_repository_get_by_index(s)->source_game == SCENARIO_SOURCE_RCT1_AA) {
+                RCT1AACount++;
```
duncanspumpkin: Might as well just delete the `RCT1AACount` variable and use `RCT1AAInstalled` setting it to true as soon as there is a single park.

src/openrct2/title/TitleScreen.cpp
```diff
@@ -225,6 +226,68 @@ void TitleScreen::TitleInitialise()
...
+            }
+            if (scenario_repository_get_by_index(s)->source_game == SCENARIO_SOURCE_RCT1_LL) {
+                RCT1LLCount++;
```
duncanspumpkin: As above but with `RCT1LL`

janisozaur:
You still need to reformat

janisozaur:
And rebase

MichaelJBerk:
Would it be possible to reopen this PR?

duncanspumpkin:
I don't have the ability. I think this might be because you have kept everything on the same branch. Always create a feature branch.

MichaelJBerk:
What would you recommend that I do, then? Should I set the [draft PR](https://github.com/OpenRCT2/OpenRCT2/pull/10370) to “ready for review” or just leave it how it is?

duncanspumpkin:
yes we will just continue in #10370. In future always make a branch. It makes things much easier.

PR #9877, opened 2019-08-15
Cleanup some Peep/Guest/Staff code

ZehMatt:


src/openrct2/peep/Peep.cpp
```diff
@@ -641,15 +643,13 @@ std::optional<CoordsXY> Peep::UpdateAction(int16_t& xy_distance)
         action_sprite_image_offset = 0;
         action = PEEP_ACTION_NONE_2;
         UpdateCurrentActionSpriteType();
-        Invalidate();
         return { { x, y } };
     }
     action_sprite_image_offset = peepAnimation[action_sprite_type].frame_offsets[action_frame];
```
duncanspumpkin: Are you sure this is sensible. Peeps don't always move when the action is being updated so wouldn't always call invalidate.

ZehMatt: UpdateCurrentActionSpriteType will call Invalidate

duncanspumpkin: But UpdateCurrentActionSpriteType isn't called in this branch.

ZehMatt: Yes but it returns a location and calls MoveTo which then invalidates. I haven't spotted any drawing glitches so far but would be nice if you gave it a spin as well.

ZehMatt:
I've added Invalidate calls to where the return value is ignored, it should cover everything now. It might still call it twice here and there but now we have less code.

Issue #9477, opened 2019-06-25
Process exit code is always 0


ZehMatt:
The function NormalisedMain is currently returning gExitCode but gExitCode is never assigned anywhere. The process should return EXIT_FAILURE when something wen't wrong and EXIT_SUCCESS when everything goes according to plan.

So everything that currently results gExitCode has to be refactored.

DzikiChrzan:
I'd like to take this. I presume task is stall available?

AaronVanGeffen:
No, #10048 aims to tackle this already.

PR #9899, opened 2019-08-18
Fixed the exit code always being 0, Issue #9477

Sonic7662:
Replaced the uninitialized gExitCode with EXIT_SUCCESS and EXIT_FAILURE where applicable and removed all unneeded declarations of gExitCode in the game files.

OpenRCT2\test\testpaint\main.cpp also uses the name gExitCode but does so correctly and independently from the game so is left unchanged.

### DIFF
--- a/src/openrct2-ui/Ui.cpp
+++ b/src/openrct2-ui/Ui.cpp
@@ -59,8 +59,12 @@ int main(int argc, const char** argv)
 
             context->RunOpenRCT2(argc, argv);
         }
+        return EXIT_SUCCESS;
     }
-    return gExitCode;
+    else
+    {
+        return EXIT_FAILURE;
+    }
 }
 
 #ifdef __ANDROID__
--- a/contributors.md
+++ b/contributors.md
@@ -132,6 +132,7 @@ The following people are not part of the development team, but have been contrib
 * Steve Xu (stevexu-umich)
 * (aw20368)
 * Jim Armstrong (41northstudios)
+* Matthew Michaud (Sonic7662)
 
 ## Toolchain
 * (Balletie) - macOS
--- a/src/openrct2-cli/Cli.cpp
+++ b/src/openrct2-cli/Cli.cpp
@@ -28,6 +28,10 @@ int main(int argc, const char** argv)
         // Run OpenRCT2 with a plain context
         auto context = CreateContext();
         context->RunOpenRCT2(argc, argv);
+        return EXIT_SUCCESS;
     }
-    return gExitCode;
+    else
+    {
+        return EXIT_FAILURE;
+    }
 }
--- a/src/openrct2/Context.cpp
+++ b/src/openrct2/Context.cpp
@@ -238,8 +238,12 @@ namespace OpenRCT2
             if (Initialise())
             {
                 Launch();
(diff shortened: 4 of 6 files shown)


Sonic7662:
Don't know much about the workflow of OpenRTC2 and I'm brand new to open projects overall so sorry if I make things overcomplicated. I'd appreciate it if someone could help me make sense of the failed checks, I only understand them to be platform-specific tests.

Gymnasiast:
The workflow is simple: you open a PR and if it isn't good enough, you just keep it open and beat it into shape. A thing to note is that, in this project, you never merge develop into your branch - always use rebase.

The CIs include platform-specific checks, yes, but all Windows and macOS checks fail, and so do half of the Linux ones. Did you make sure to compile it yourself, and if so, what platform do you use on your development machine?

src/openrct2-ui/Ui.cpp
```diff
@@ -59,8 +59,12 @@ int main(int argc, const char** argv)
 
             context->RunOpenRCT2(argc, argv);
         }
+        return EXIT_SUCCESS;
+    }
+    else
+    {
+        return EXIT_FAILURE;
```
Gymnasiast: I'm not sure if this is correct. There are many cases where OpenRCT2 simply prints something to the console and does not run the game - extracting sprites for example.

Sonic7662: I've spent a little more time with the game's code and have a couple of questions. Firstly, if it's common for fatal errors in the game to simply print a message, I wonder if issue #9477 is misguided in wanting to replace the use of gExitCode, I'm not sure how the solution proposed was intended to work across all relevant files. Perhaps it would make sense to employ gExitCode everywhere as it might have originally been intended. I would appreciate it if @ZehMatt could elaborate on the original intent of the issue.

My second question is concerning the sometimes already used exit(), is this only acceptable in some cases for potential issues with threads or would it be worth considering using it as the standard error termination?

To speak to what might be obvious, this is a very big and complex project for me, but I tend to learn and improve best when I throw myself off the deep end, so again, apologies for being new.

Sonic7662:
Thank you for the feedback @Gymnasiast , yes I did compile and run it myself, I am using Windows 10 on x64 with Visual Studio 2019 Community.

src/openrct2-ui/Ui.cpp
IntelOrca: We could do `std::exit` but that will not clean up any stack objects. When a fatal error occurs (which should really be rare anyway), the best thing to do is throw an exception which can then bubble up the program.

For this PR, `cmdline_run` can return 1 or 0 which just controls whether the main game should run or not. That function and its call chain would need to return a struct which contains exit code and whether it should exit. Or maybe just `std::optional` would work where `nullopt` states that it should run the main game.

Gymnasiast:
I'm assuming you're still stuck on some stuff you need answers for. If so, please ask, since I'd like to get this PR moving again.

Gymnasiast:
I'm going to close this PR, as it unfortunately does not solve the issue. But don't be discouraged, I'm just cleaning our open PRs a bit. Feel free to reply, ask questions or open another one.

PR #9903, opened 2019-08-18
Extend tile element size to 16 bytes

Gymnasiast:
A rough, first version. Importing RCT1/RCT2 saves works, but autosaving creates a corrupted version. I'll need some help with that.

### DIFF
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -1331,12 +1331,11 @@ void S6Exporter::ExportTileElements()
         auto dst = &_s6.tile_elements[index];
         if (src->base_height == 0xFF)
         {
-            std::memcpy(dst, src, sizeof(*src));
+            std::memcpy(dst, src, sizeof(*dst));
         }
         else
         {
             auto tileElementType = (RCT12TileElementType)src->GetType();
-            // Todo: replace with setting invisibility bit
             if (tileElementType == RCT12TileElementType::Corrupt || tileElementType == RCT12TileElementType::EightCarsCorrupt14
                 || tileElementType == RCT12TileElementType::EightCarsCorrupt15)
                 std::memcpy(dst, src, sizeof(*dst));
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -194,10 +194,7 @@ void S6Exporter::Export()
     _s6.scenario_srand_0 = state.s0;
     _s6.scenario_srand_1 = state.s1;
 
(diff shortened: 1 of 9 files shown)


src/openrct2/world/TileElement.h
```diff
@@ -65,6 +65,8 @@ struct TileElementBase
     uint8_t base_height;      // 2
     uint8_t clearance_height; // 3
 
+    uint8_t testPad[8];
```
IntelOrca: This struct shouldn't have any padding. Padding should be added to all the other structs.

Gymnasiast: It was just a test, but I will change it.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1326,6 +1323,173 @@ void S6Exporter::ExportMapAnimations()
...
 
+void S6Exporter::ExportTileElements()
+{
+    for (uint32_t index = 0; index < RCT2_MAX_TILE_ELEMENTS; index++)
+    {
+        auto src = &gTileElements[index];
+        auto dst = &_s6.tile_elements[index];
+        if (src->base_height == 0xFF)
+        {
+            std::memcpy(dst, src, sizeof(*src));
```
duncanspumpkin: What is this doing? Won't this be too big?

Gymnasiast: It copies "null" tile elements over (mirroring S6Importer). It was indeed too big - I have corrected that now

src/openrct2/world/TileElement.h
```diff
@@ -83,6 +83,7 @@ struct TileElementBase
 struct TileElement : public TileElementBase
 {
     uint8_t pad_04[4];
+    uint8_t pad_08[8];
```
duncanspumpkin: I think the padding should **only** be added to `TileElement` and not to any of the others. All of the other structs should 16 or less bytes. `TileElement` is the only one that needs to be `16`. That way we don't have to clutter the code base with pragmas. Its how we have done the sprites (kind of they use a big union).

Gymnasiast: That will cause the assertions of the other tile element types to fail. Not an option.

IntelOrca: @duncanspumpkin While the others don't need to be 16 bytes, it is safer. For example is I were do `SurfaceElement elements[5];` I would probably want that to be 5 * 16 bytes.

duncanspumpkin:
Did you check the tile inspector in multiplayer and increment the network version?

ZehMatt:
Luckily it just refuses to load the map and disconnects currently, but a network bump would have been good.

PR #9907, opened 2019-08-19
Fix #9603: Don't render audio when master volume is turned off

### LINKED ISSUE #9603: Don't render audio when master volume is turned off
We have toggle-able master volume, yet when it is disabled and the games does not _output_ any sounds, all the music/sfx are still processed, only to be dumped at a late stage. When testing on Raspberry Pi, which is not capable of full 40Hz expected by the game, there is significant cost to be paid for resampling audio and a naive test that emptied `void AudioMixer::MixChannel(ISDLAudioChannel* channel, uint8_t* data, size_t length)` provided measurable improvement.

When master volume is off, we should not render any sound effects and resampling should be further investigated for possible improvements (e.g. offloading to job pool).

See raspi callgrind outputs:
* https://gist.github.com/janisozaur/481b435d3c4ccdbc51b98fb306c30d0d/raw/ebe02b9451d2591671d89c9a4a4b6efcf46ebb37/callgrind.out.no-audio for https://github.com/janisozaur/OpenRCT2/commit/8cad86911f7ba5edb42e0b149591bd6de50bef12
* https://gist.github.com/janisozaur/481b435d3c4ccdbc51b98fb306c30d0d/raw/ebe02b9451d2591671d89c9a4a4b6efcf46ebb37/rpi4-perf.callgrind for https://github.com/janisozaur/OpenRCT2/commit/22a96116747bd8f59063748ff1a81bdc4451e11c

s0r00t:
This fix attempts to stop mixing channels when master volume is disabled, and offloads mixing to a dedicated job pool.  

I am not sure of the efficiency of the pool, as I've been unable to run measurement tools on my workstation.

### DIFF
--- a/src/openrct2-ui/audio/AudioMixer.cpp
+++ b/src/openrct2-ui/audio/AudioMixer.cpp
@@ -223,7 +223,8 @@ namespace OpenRCT2::Audio
             {
                 auto channel = *it;
                 int32_t group = channel->GetGroup();
-                if (group != MIXER_GROUP_SOUND || gConfigSound.sound_enabled)
+                if ((group != MIXER_GROUP_SOUND || gConfigSound.sound_enabled) && gConfigSound.master_sound_enabled
+                    && gConfigSound.master_volume != 0)
                 {
                     MixChannel(channel, dst, length);
                 }
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -10,6 +10,7 @@
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
 - Fix: [#9729] Peeps do not take into account height difference when deciding to pathfind to a ride entrance. (original bug)
+- Fix: [#9603] Don't render audio when master volume is turned off
 - Improved: [#9466] Add the rain weather effect to the OpenGL renderer.
 
 0.2.3 (2019-07-10)


ZehMatt:
I don't think the job pool makes a big difference given that each task will lock so no task runs really in parallel. But the fix is very welcome.

s0r00t:
Thanks for the feedback.

I am not sure whether the job pool implementation is a good idea for two reasons :

* As you've noticed, the buffers are shared between jobs, so they just run sequentially. Creating individual buffers for each job would consume an absurd amount of memory.
* The other issue that stems from this implementation is the eventual audio latency. So far, I haven't noticed any, but I wouldn't be surprised if this happens on lower-end systems.

Sadly, I do not know how to improve the audio rendering. Suggestions are welcome :)

Gymnasiast:
I would just keep to fixing the bug in this PR. "Hiding" an addition in a commit/PR that only claims to fix an issue makes fixing any bugs that arise from it harder.

s0r00t:
Good point. I have just pushed my reworked commits, removing the refactor. I have included my minor change to JobPool::AddTask in a separate commit.

Gymnasiast:
You need to fix the formatting of the first commit. You can see what needs to be changed by following the link to the failed Travis tests.

Tip: after `git add`, do `git clang-format`, followed by another `git add` and it will autoformat your files.

s0r00t:
Thanks for the tip, this should be fixed :)

IntelOrca:
There is still a change to JobPool.hpp

s0r00t:
Should I just discard it, or submit it on another PR? Considering the size of the change, I might as well drop it.

Gymnasiast:
That would be better, yeah. PRs should address one thing.

s0r00t:
I've isolated the JobPool change onto another branch, and squashed my commits.

janisozaur:
Looks really nice, but I would like to test it. I will do so later this week.

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -223,7 +223,8 @@ namespace OpenRCT2::Audio
             {
                 auto channel = *it;
                 int32_t group = channel->GetGroup();
-                if (group != MIXER_GROUP_SOUND || gConfigSound.sound_enabled)
+                if ((group != MIXER_GROUP_SOUND || gConfigSound.sound_enabled) && gConfigSound.master_sound_enabled
+                    && gConfigSound.master_volume != 0)
```
ZehMatt: Having master_volume == 0 is essentially identical to gConfigSound.master_sound_enabled == false which seems a little silly to me to have two options, I'm not requesting changes just stating the condition here. We should perhaps ditch the master volume option and simply check if its not 0 at some point.

Gymnasiast: I disagree. It's nicer to be able to mute the volume and have it return to the old level when unmuting. 

Among other things, it would mean that clicking the Mute button in the toolbar, then unclicking it, would return the volume to full blast.

ZehMatt: Yes, but how often is one toggling it on and off? But I get your point for convenience reasons.

janisozaur:
I'm going to take @ZehMatt's word then.

PR #9914, opened 2019-08-20
Refactor network queue out and one desync fix

ZehMatt:
I started to cleanup the network part a little bit, the game actions/game commands don't belong in there so I moved it into the GameAction framework. This PR also fixes a desync which was caused by clearing the tick data too late, now the tick data is erased when the client receives the first bytes for the map allowing it to buffer up until the map is fully loaded which is the initial state as the server advances while the client receives the map and used the wrong data to compare against.

### DIFF
--- a/src/openrct2/actions/GameAction.cpp
+++ b/src/openrct2/actions/GameAction.cpp
@@ -19,6 +19,7 @@
 #include "../platform/platform.h"
 #include "../scenario/Scenario.h"
 #include "../world/Park.h"
+#include "../world/Scenery.h"
 
 #include <algorithm>
 #include <iterator>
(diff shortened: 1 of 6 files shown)


src/openrct2/actions/GameAction.cpp
```diff
@@ -66,6 +95,77 @@ namespace GameActions
         return false;
     }
 
+    void Enqueue(const GameAction* ga, uint32_t tick)
```
IntelOrca: Any reason not to declare as `const GameAction& ga`?

ZehMatt: Only for consistency reasons, Query/Execute also take the same parameter, should make them maybe references in another PR.

Issue #7423, opened 2018-04-17
Textures missing on scenario loading


MisterSchmidt:
OS: _Windows 10_
version: _0.1.3_
build: _11207f9_

Some weird path loading issues which can be fixed by starting the scenario multiple times in a row without closing down the game. I found this issue by looking at the Path menu and also noticed the Park Entrance disappearing.

- [ **No** ] Reproducible in RCT2 (vanilla)?
- [ **No** ] Specific to multiplayer?

1. Start game.
2. Open a scenario or savegame with preferable mutliple paths.
3. Look in the path menu. (some paths should be missing.)
4. Close scenario.
5. Open the same scenario again.
6. Take another look in the path menu. (less paths should be missing then before)

My Example, I also tried this with a savegame

**First load:**
![naamloos1](https://user-images.githubusercontent.com/11180804/38899526-67a56302-4298-11e8-9352-7764b35ba1e4.jpg)

**Second load after closing scenario:**
![naamloos2](https://user-images.githubusercontent.com/11180804/38899573-8f2f75f2-4298-11e8-9b75-84a9dce21829.jpg)

**Third load after closing scenario:**
![naamloos3](https://user-images.githubusercontent.com/11180804/38899582-95191554-4298-11e8-8153-78ff190052a3.jpg)

AaronVanGeffen:
Looks like your savegame is including too many objects. Try opening the object selection window and deselecting some. (Keep an eye on the limits in the status bar.)

MisterSchmidt:
There is a limit isn't there?

AaronVanGeffen:
There is indeed, but the image limit may still be exceeded if large objects are selected. We recently learned as much in https://github.com/OpenRCT2/OpenRCT2/issues/7294.

MisterSchmidt:
Uhm ok, an image limit, what do you suggest? disabling some Large Scenery as they are in the same category as paths or someting?

MisterSchmidt:
Ah, I get the problem now! I disabled a lot of rides which seem to fix the problem indeed. And here's a save file with the issues. 

[Jupiter Lawn (image limit).zip](https://github.com/OpenRCT2/OpenRCT2/files/1921874/Jupiter.Lawn.image.limit.zip)

Gymnasiast:
I can no longer reproduce this issue on the current develop version of OpenRCT2. Could you try for yourself to see if the issue is solved?

Gymnasiast:
[...]

Issue #6073, opened 2017-07-29
Multi-dimension coaster sprites corrupted + crash


jensj12:
**OS:** Windows 10
**Version:** 0.1.1
**Commit/Build:** 2ec0df1

The sprites for the multi-dimension coaster can become corrupted. After that, opening the rollercoaster tab in the build new ride window crashed my game.

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Multiplayer?

**Steps to reproduce:**
I have yet to find a reliable way to reproduce the issue. What I did: open lots of saves until you encounter some corruption, then open new ride window and switch to coaster tab.

**Dump file**
[OpenRCT2CrashDumpMDcoaster.zip](https://github.com/OpenRCT2/OpenRCT2/files/1185081/OpenRCT2CrashDumpMDcoaster.zip)

**Screenshots / Video:**
A video is available for download here: https://openrct2.org/forums/topic/2564-um/

**Save game:**
Any game with a MD coaster may or may not work. On this save the sprites got corrupted: 
[Bogus sandbox.txt](https://github.com/OpenRCT2/OpenRCT2/files/1185084/Bogus.sandbox.txt) (I may have switched to another save before the crash)

jensj12:
[Made my own video](https://youtu.be/YFNlNpGUozU). Where the video hangs, it has crashed. This is the dump it created and I saved right before starting the video. Loading lots of saves seems to work.
[OpenRCT2CrashDumpMDcoasterWithSave.zip](https://github.com/OpenRCT2/OpenRCT2/files/1185093/OpenRCT2CrashDumpMDcoasterWithSave.zip)

IntelOrca:
My guess is that it is related to the image list not correctly allocating the images which is likely for multi-dimensional coaster as it has so many images to allocate. The image list can become fragmented after many object loads and unloads reducing the number of free image slots available.

Nubbie:
Dump from OpenRCT2CrashDumpMDcoasterWithSave.zip

Error message;
Unhandled exception thrown: read access violation.
colourSrc was 0xFFFFFFFFFFFFFFFF.

Call stack;
```
>	openrct2.dll!gfx_draw_sprite_raw_masked_software(rct_drawpixelinfo * dpi, int x, int y, int maskImage, int colourImage=-1) Line 659	C++
[...]

PR #9918, opened 2019-08-21
Increase image list and fix assert when exiting scenario editor

IntelOrca:
By re-ordering the image IDs I can squeeze another ~100K or so available image slots for object images. This should help reduce likeliness of hitting the image limit.

I also fixed that "not all images freed" assert you get when exiting the game from object selection window in scenario editor.


Gymnasiast:
I was planning to test this with issues like https://github.com/OpenRCT2/OpenRCT2/issues/6073 and https://github.com/OpenRCT2/OpenRCT2/issues/7423. But oddly enough, these parks no longer seem to exhibit the issues in question.

src/openrct2/Context.cpp
```diff
@@ -148,13 +148,14 @@ namespace OpenRCT2
             // NOTE: We must shutdown all systems here before Instance is set back to null.
             //       If objects use GetContext() in their destructor things won't go well.
 
+            network_close();
```
Gymnasiast: Perhaps add a comment to explain why it needs to be done before unloading objects, to avoid regressions in the future.

PR #9921, opened 2019-08-22
Refactor network groups out of network

ZehMatt:
Another PR to clean up more code in the Network area, this will probably take a bit to complete.

### DIFF
--- a/src/openrct2-ui/WindowManager.cpp
+++ b/src/openrct2-ui/WindowManager.cpp
@@ -77,10 +77,6 @@ public:
                 return window_map_open();
             case WC_MAPGEN:
                 return window_mapgen_open();
-#ifndef DISABLE_NETWORK
-            case WC_MULTIPLAYER:
-                return window_multiplayer_open();
-#endif
             case WC_MUSIC_CREDITS:
                 return window_music_credits_open();
             case WC_CONSTRUCT_RIDE:
@@ -105,12 +101,6 @@ public:
                 return window_scenery_open();
             case WC_SCENERY_SCATTER:
                 return window_scenery_scatter_open();
-#ifndef DISABLE_NETWORK
-            case WC_SERVER_LIST:
-                return window_server_list_open();
-            case WC_SERVER_START:
-                return window_server_start_open();
-#endif
             case WC_KEYBOARD_SHORTCUT_LIST:
                 return window_shortcut_keys_open();
             case WC_STAFF_LIST:
@@ -136,6 +126,12 @@ public:
             case WC_WATER:
                 return window_water_open();
 #ifndef DISABLE_NETWORK
+            case WC_MULTIPLAYER:
+                return window_multiplayer_open();
+            case WC_SERVER_LIST:
+                return window_server_list_open();
+            case WC_SERVER_START:
+                return window_server_start_open();
             case WC_NETWORK:
                 return window_network_open();
 #endif
--- a/src/openrct2-ui/WindowManager.cpp
+++ b/src/openrct2-ui/WindowManager.cpp
@@ -77,8 +77,10 @@ public:
                 return window_map_open();
             case WC_MAPGEN:
                 return window_mapgen_open();
+#ifndef DISABLE_NETWORK
             case WC_MULTIPLAYER:
                 return window_multiplayer_open();
+#endif
             case WC_MUSIC_CREDITS:
                 return window_music_credits_open();
             case WC_CONSTRUCT_RIDE:
@@ -133,8 +135,10 @@ public:
                 return window_viewport_open();
             case WC_WATER:
                 return window_water_open();
+#ifndef DISABLE_NETWORK
             case WC_NETWORK:
                 return window_network_open();
+#endif
             default:
                 Console::Error::WriteLine("Unhandled window class (%d)", wc);
                 return nullptr;
@@ -193,10 +197,10 @@ public:
                 return window_sign_open(id);
             case WD_SIGN_SMALL:
                 return window_sign_small_open(id);
-
+#ifndef DISABLE_NETWORK
             case WD_PLAYER:
                 return window_player_open(id);
-
+#endif
             default:
                 return nullptr;
         }
--- a/src/openrct2-ui/windows/Multiplayer.cpp
+++ b/src/openrct2-ui/windows/Multiplayer.cpp
@@ -7,18 +7,20 @@
  * OpenRCT2 is licensed under the GNU General Public License version 3.
  *****************************************************************************/
 
-#include <openrct2-ui/interface/Dropdown.h>
-#include <openrct2-ui/interface/Widget.h>
-#include <openrct2-ui/windows/Window.h>
-#include <openrct2/Game.h>
-#include <openrct2/actions/NetworkModifyGroupAction.hpp>
-#include <openrct2/config/Config.h>
-#include <openrct2/drawing/Drawing.h>
-#include <openrct2/localisation/Localisation.h>
-#include <openrct2/network/NetworkGroups.h>
-#include <openrct2/network/network.h>
-#include <openrct2/sprites.h>
-#include <openrct2/util/Util.h>
+#ifndef DISABLE_NETWORK
+
+#    include <openrct2-ui/interface/Dropdown.h>
+#    include <openrct2-ui/interface/Widget.h>
+#    include <openrct2-ui/windows/Window.h>
+#    include <openrct2/Game.h>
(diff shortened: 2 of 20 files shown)


ZehMatt:
Some things I should mention what this PR does:
- Introduces a "Host" group which the server operates on, users can not modify or assign other users to this group.
- Fixes an actual bug that didn't allow the host to change the group of users because the player id is always -1 when not networked first.
- Keeps compatibility of the current groups.json

Also the "Admin" group is currently immutable which I assume has been only the case because the host was enforced to use the admin group, now that this is no longer the case should we allow users to modify the admin group?

src/openrct2-ui/WindowManager.cpp
```diff
@@ -77,8 +77,10 @@ class WindowManager final : public IWindowManager
                 return window_map_open();
             case WC_MAPGEN:
                 return window_mapgen_open();
+#ifndef DISABLE_NETWORK
```
duncanspumpkin: Can we move all of the network windows next to each other so that we don't have to have multiple ifdefs

src/openrct2-ui/windows/Multiplayer.cpp
```diff
@@ -360,32 +364,41 @@ static void window_multiplayer_set_pressed_tab(rct_window* w)
...
+    int32_t numItems = 0;
+    int32_t defaultIndex = -1;
 
     dropdownWidget = widget - 1;
 
-    numItems = network_get_num_groups();
+    NetworkGroups* networkGroups = network_get_groups();
+    auto groups = networkGroups->GetAll(true);
+
+    numItems = static_cast<int32_t>(groups.size());
```
duncanspumpkin: may as well auto this to avoid the cast.

src/openrct2-ui/windows/Multiplayer.cpp
```diff
@@ -749,15 +767,19 @@ static void window_multiplayer_groups_mouseup(rct_window* w, rct_widgetindex wid
...
-        case WIDX_RENAME_GROUP:;
-            int32_t groupIndex = network_get_group_index(_selectedGroup);
-            const utf8* groupName = network_get_group_name(groupIndex);
+        case WIDX_RENAME_GROUP:
+        {
+            auto* group = networkGroups->GetById(_selectedGroupId);
+            Guard::Assert(group != nullptr);
+
+            const utf8* groupName = group->GetName().c_str();
             window_text_input_raw_open(w, widgetIndex, STR_GROUP_NAME, STR_ENTER_NEW_NAME_FOR_THIS_GROUP, (utf8*)groupName, 32);
```
duncanspumpkin: is the additional cast still required?

src/openrct2-ui/windows/Window.h
```diff
@@ -76,7 +76,9 @@ rct_window* window_ride_demolish_prompt_open(Ride* ride);
 rct_window* window_ride_refurbish_prompt_open(Ride* ride);
 rct_window* window_sign_open(rct_windownumber number);
 rct_window* window_sign_small_open(rct_windownumber number);
+#ifndef DISABLE_NETWORK
 rct_window* window_player_open(uint8_t id);
```
duncanspumpkin: Move into the disable network section

src/openrct2/network/NetworkGroups.cpp
```diff
@@ -0,0 +1,304 @@
...
+    {
+        if (group == nullptr)
+            continue;
+        if (group->GetName() == name)
+            return group.get();
+    }
+    return nullptr;
+}
+
+std::vector<NetworkGroup*> NetworkGroups::GetAll(bool excludeHost) const
```
duncanspumpkin: I suggest we make this by default exclude host and have a constexpr value of false that includes the host.

ZehMatt: I wanted to have this explicit as for network serialization the host must be included. For the UI part its excluded. If it defaults to something the person who is tempted to call this might not be aware what it would do otherwise.

Broxzier:
> Also the "Admin" group is currently immutable which I assume has been only the case because the host was enforced to use the admin group, now that this is no longer the case should we allow users to modify the admin group?

Yes. I think starting off with a few default groups like "Admin", "User, and "Spectator" are good, but give full control about them to the user.

ZehMatt:
I've modified the way it loads and saves the configuration to use names which is much easier for the initial setup and doesn't require to back track files.

The new format for groups.json:
```
{
    "default_group": "User",
    "groups": [
        {
            "name": "Admin",
            "permissions": [
                "PERMISSION_CHAT",
                "PERMISSION_TERRAFORM",
                "PERMISSION_SET_WATER_LEVEL",
                "PERMISSION_TOGGLE_PAUSE",
                "PERMISSION_CREATE_RIDE",
                "PERMISSION_REMOVE_RIDE",
                "PERMISSION_BUILD_RIDE",
                "PERMISSION_RIDE_PROPERTIES",
                "PERMISSION_SCENERY",
                "PERMISSION_PATH",
                "PERMISSION_CLEAR_LANDSCAPE",
                "PERMISSION_GUEST",
                "PERMISSION_STAFF",
                "PERMISSION_PARK_PROPERTIES",
                "PERMISSION_PARK_FUNDING",
                "PERMISSION_KICK_PLAYER",
                "PERMISSION_MODIFY_GROUPS",
                "PERMISSION_SET_PLAYER_GROUP",
                "PERMISSION_CHEAT",
                "PERMISSION_TOGGLE_SCENERY_CLUSTER",
                "PERMISSION_PASSWORDLESS_LOGIN",
                "PERMISSION_MODIFY_TILE",
                "PERMISSION_EDIT_SCENARIO_OPTIONS"
            ]
        },
        {
            "name": "Spectator",
            "permissions": [
                "PERMISSION_CHAT"
            ]
        },
        {
            "name": "User",
            "permissions": [
                "PERMISSION_CHAT",
                "PERMISSION_TERRAFORM",
                "PERMISSION_SET_WATER_LEVEL",
                "PERMISSION_TOGGLE_PAUSE",
                "PERMISSION_CREATE_RIDE",
                "PERMISSION_REMOVE_RIDE",
                "PERMISSION_BUILD_RIDE",
                "PERMISSION_RIDE_PROPERTIES",
                "PERMISSION_SCENERY",
                "PERMISSION_PATH",
                "PERMISSION_CLEAR_LANDSCAPE",
                "PERMISSION_GUEST",
                "PERMISSION_STAFF",
                "PERMISSION_PARK_PROPERTIES",
                "PERMISSION_PARK_FUNDING",
                "PERMISSION_TOGGLE_SCENERY_CLUSTER"
            ]
        },
        {
            "name": "격리중",
            "permissions": []
        }
    ]
}
```
The new format for users.json:
```
[
    {
        "hash": "",
        "name": "Matt",
        "group": "Host",
    },
    {
        "hash": "4d0e90c173766a2535a53eb2f62c82dfa016d680",
        "name": "Matt #2",
        "group": "격리중",
    },
    {
        "hash": "5cadab5de83ffb4ed9e69ba488be8d18378cb1bb",
        "name": "Zeh Matt",
        "group": "Admin",
    },
    {
        "hash": "a433e321e2f164d0cc8ab938353a539b8f918219",
        "name": "Matt #3",
        "group": "User",
    }
]
```
As for the new rules:
Duplicate names are not allowed, if the configuration file has two records of the same name it will overwrite the previous one.
If a specified group was not found the user will fallback to the default, if the default is improperly configured it will use to the builtin defaults.

Old configuration should not need to be changed as it will still use the id for a one way import to the new format.

I've bumped up the network version and added a changelog entry, from my end its good to go but it would be nice if someone could do a few tests.

AaronVanGeffen:
> Old configuration should not need to be changed as it will still use the id for a one way import to the new format.

Thank you for remembering to do this. :)

IntelOrca:
If names are unique, why not use that as the key in an object?

```
{
    "default_group": "Spectator",
    "groups": {
        "Admin": {},
        "Player": {},
        "Spectator": {}
    }
}
```

AaronVanGeffen:
@ZehMatt How ready is this? Could you rebase this it can be thoroughly tested and such?

ZehMatt:
Last time I tested it worked correctly however @IntelOrca brought up a good point about the json storage. So the question would be: Should I adapt to the suggestion or leave it as is and do that the next time, the idea about this PR was a code cleanup initially.

AaronVanGeffen:
@ZehMatt Perhaps an option would be to merge the refactoring without the changes to the groups.json format? They seem to be limited to the last few commits, so it looks easy enough to split off to another PR.

PR #9925, opened 2019-08-22
Fix #9920: Ghost pieces break maps on multiplayer for the client

### LINKED ISSUE #9920: Ghost pieces break maps on multiplayer for the client
**OS:** Tested on both Windows 10 and Ubuntu 18.04
**Version:** v0.2.3-277
**Commit/Build:** af2ad80

When playing on a multiplayer server, everything seems to work fine. Actions get through, and nothing looks to be wrong, but upon reconnecting the map is very broken on the client side. On tiles where things got changed, surface elements with a base height of 255, which is likely a result of all elements having been removed from that tile as well as strange edges (possibly missing tiles?). On the server, all is good, however. After having messed around for a bit, I think it's ghost pieces that cause the issue.

In some cases it looks like a tile is rendering the objects of the tile next to it. Probably because its own element list is empty and it loops until an element with the end-of-tile flag set. This could possibly cause crashes too.

Everything worked fine on a build from 5 days ago: 406ec7f

**Steps to reproduce:**
1. Host any park
2. Connect to it as a client
3. On either the host or client, build something (paths, ride)
4. Disconnect the client and reconnect again
5. Notice how on the changed tiles there appear to be massive pillars of surface elements.

This can also be replicated more easily:
1. Host any park
2. With the pathing tool, hover over some owned tiles to make a ghost appear
3. Connect to the server as a client

**Screenshots / Video:**
Whispering Cliffs 2019-08-22 18-26-58

ZehMatt:
This reverts some things from #9903 as it broke multiplayer. I'm not exactly sure what the root cause is but reverting the function and struct size was necessary to get the expected behavior back. 

Things that I do know:
1. Reverting ```scenario_fix_ghosts``` fixed half of it, it seems that the way the pointer list is built doesn't quite match what gTileElementPointers have.
2. Reverting back to size of 8 fixed the other half of it, here I'm not sure what the issue really is.

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2392,7 +2392,7 @@ static void window_ride_construction_draw_track_piece(
 }
 
 static TileElement _tempTrackTileElement;
-static TileElement _tempSideTrackTileElement = { 0x80, 0x8F, 128, 128, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
+static TileElement _tempSideTrackTileElement = { 0x80, 0x8F, 128, 128, 0, 0, 0, 0 };
 static TileElement* _backupTileElementArrays[5];
 
 /**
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
Let's fix the issue without reverting. It will be something obvious.

ZehMatt:
Works for me, I was unable to find a fix so lets leave it broken then.

PR #9927, opened 2019-08-23
Fix #9926: Africa - Oasis park has wrong peep spawn

### LINKED ISSUE #9926: Africa - Oasis park has wrong peep spawn
**OS:** Windows 10
**Version:** 0.2.3
**Commit/Build:** 4cf1d1e

Peeps spawn somewhere inside the terrain close to the entrance, instead of on the path to the entrance.

**Steps to reproduce:**
1. Open WW Africa - Oasis park and send a large tram.

**Screenshots / Video:**
A screenshot has been posted on the [forums](https://forums.openrct2.org/topic/4478-mirage-madness-afrika-oase-guest-dont-spawn/), but it's too small to see the problem.

**Fix:**
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/rct2/S6Importer.cpp#L890

Broxzier:
I also noticed that this same scenario has a single tile outside of the park whose ownership is set to build rights. To avoid breaking existing parks and since there are no side effects, I left it unchanged. This made me realize though that these peep-spawn fixes could break parks that have been saved under the same name and with their entry points moved.

Also ping @Gymnasiast since you probably have access to more RCT2 installations, perhaps another filename may need to be checked too ("WW Africa - Oasis.SC6" I guess?).

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -10,8 +10,9 @@
 - Fix: [#9625] Show correct cost in scenery selection.
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
-- Fix: [#9729] Peeps do not take into account height difference when deciding to pathfind to a ride entrance. (original bug)
-- Fix: [#9603] Don't render audio when master volume is turned off
+- Fix: [#9729] Peeps do not take into account height difference when deciding to pathfind to a ride entrance (original bug).
+- Fix: [#9603] Don't render audio when master volume is turned off.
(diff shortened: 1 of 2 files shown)


Gymnasiast:
I just tested on vanilla, and the problem does not occur there. Haven't yet investigated why.

Broxzier:
That's odd, for me it does. This screenshot was made in vanilla RCT2: 
Maybe the scenario file is different. The md5 hash of Africa - Oasis.SC6 is: `666F4ECF4505AEF1490EF1E6630214BF`

Gymnasiast:
Yes, that's it. Specifically, it's broken on the CD version, but it works on the GOG version.

PR #9938, opened 2019-08-24
Refactor some accessors to tile element type and flags fields

Gymnasiast:
This also fixes the serialisation of tile elements that have data in the upper 8 bytes.

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -393,6 +393,10 @@ template<> struct DataSerializerTraits<TileElement>
         {
             stream->WriteValue(tileElement.pad_04[i]);
         }
+        for (int i = 0; i < 8; ++i)
+        {
+            stream->WriteValue(tileElement.pad_08[i]);
+        }
     }
     static void decode(IStream* stream, TileElement& tileElement)
     {
@@ -403,6 +407,10 @@ template<> struct DataSerializerTraits<TileElement>
         for (int i = 0; i < 4; ++i)
         {
             tileElement.pad_04[i] = stream->ReadValue<uint8_t>();
+        }
+        for (int i = 0; i < 8; ++i)
+        {
+            tileElement.pad_08[i] = stream->ReadValue<uint8_t>();
         }
     }
     static void log(IStream* stream, const TileElement& tileElement)
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -310,7 +310,7 @@ void map_init(int32_t size)
     {
         TileElement* tile_element = &gTileElements[i];
         tile_element->ClearAs(TILE_ELEMENT_TYPE_SURFACE);
-        tile_element->SetFlag(TILE_ELEMENT_FLAG_LAST_TILE, true);
+        tile_element->SetLastForTile(true);
         tile_element->base_height = 14;
         tile_element->clearance_height = 14;
         tile_element->AsSurface()->SetWaterHeight(0);
@@ -957,7 +957,7 @@ void tile_element_remove(TileElement* tileElement)
     }
 
     // Mark the latest element with the last element flag.
-    (tileElement - 1)->SetFlag(TILE_ELEMENT_FLAG_LAST_TILE, true);
+    (tileElement - 1)->SetLastForTile(true);
     tileElement->base_height = 0xFF;
 
     if ((tileElement + 1) == gNextFreeTileElement)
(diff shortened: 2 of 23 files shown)


src/openrct2/world/Map.cpp
```diff
@@ -1178,10 +1178,11 @@ TileElement* tile_element_insert(const TileCoordsXYZ& loc, int32_t flags)
         originalTileElement++;
         newTileElement++;
 
-        if ((newTileElement - 1)->flags & TILE_ELEMENT_FLAG_LAST_TILE)
+        if ((newTileElement - 1)->IsLastForTile())
         {
             // No more elements above the insert element
-            (newTileElement - 1)->flags &= ~TILE_ELEMENT_FLAG_LAST_TILE;
+            (newTileElement - 1)->SetFlag(TILE_ELEMENT_FLAG_LAST_TILE, false);
+            ;
```
IntelOrca: Spurious semicolon

test/testpaint/Compat.cpp
```diff
@@ -155,6 +155,19 @@ rct_sprite* get_sprite(size_t sprite_idx)
...
+{
+    if (on)
+        flags |= flag;
+    else
+        flags &= ~flag;
+}
+
 bool TileElementBase::IsLastForTile() const
 {
     return (this->flags & TILE_ELEMENT_FLAG_LAST_TILE) != 0;
```
IntelOrca: This could use `HasFlag`

duncanspumpkin:
Are we going to create a get occupied quadrant as well? I didn't realise paths used the occupied quadrant to indicate if it was a queue. Might end up being messy.

Gymnasiast:
The usage of the lower three bits of `flags` is most peculiar. It only appears to be set by track elements, yet it is checked for almost every tile element type.

In addition: small scenery does not use these to save the quadrant an item is in, as it uses bits in `type` for that instead. Despite this, the tile inspector exposes the quadrant-in-flags info only for small scenery.

Can anyone shed any light on this?

test/testpaint/Compat.cpp
Gymnasiast: Since 99% of actual flags already used dedicated methods, I have taken out HasFlag/SetFlag instead.

duncanspumpkin:
The lower 4 bits of flags is expected to be what quadrants are occupied by the tile element. Small scenery also have additional information within the type that is used to tell what quarter tile is in use. This is a different sort of quarter identifier (0-3). In theory there is no need for both (for small scenery) but you would need to use a counter to convert from the 4 bit flags to the 2 bit (0-3) quarter identifier. The quadrant data is only set during a tile element insert (or with the tile inspector). We should create an accessor for it to make it a bit more obvious how it is used.

Surface, wall, banner elements do not set the quadrants occupied as they don't make sense to use an occupied quadrant.

Also just some additional notes / actions:
1. `map_animation_invalidate_queue_banner` is incorrect it looks at flags to check if its a queue but queue is on type not on flags. This is incorrect. I've checked the assembly and its a mistake in OpenRCT2. This PR fixes that so thats good.
2. `scenery_update_age` IMHO is incorrect and is a vanilla bug. It is only checking for occupied quarters in 3 quarters when it should check 4 quarters.
3. `tile_element_insert` should have its `flags` parameter renamed to `quadrants`.
4. `tile_inspector_paste_element_at` should have a comment stating that quadrants not required to be set in `tile_element_insert` as it will be copied when the rest of the tile is copied over.
5. `tile_element_insert` calling functions have a variety of formats for saying all quadrants occupied `0b1111` and should be standardised to using binary literals.

src/openrct2/world/MapAnimation.cpp
```diff
@@ -131,6 +131,7 @@ static bool map_animation_invalidate_queue_banner(int32_t x, int32_t y, int32_t
             continue;
         if (tileElement->GetType() != TILE_ELEMENT_TYPE_PATH)
             continue;
+        // Change from original: originally checked type instead of flags, which was likely a bug.
```
duncanspumpkin: Sorry this was a mistake in open rct2 not in vanilla

Gymnasiast:
Ok, that should be it. @duncanspumpkin could you approve/request changes?

src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -404,6 +408,10 @@ template<> struct DataSerializerTraits<TileElement>
         {
             tileElement.pad_04[i] = stream->ReadValue<uint8_t>();
         }
+        for (int i = 0; i < 8; ++i)
```
ZehMatt: Breaks network compatibility, needs a network version bump.

PR #9939, opened 2019-08-24
Create new ImageId struct

IntelOrca:
Create new struct to represent an image + colour to replace our `uint32_t` image ids. When all is converted we can expand the struct to a large size so that we can increase the image limit.

### DIFF
--- a/src/openrct2/drawing/Drawing.h
+++ b/src/openrct2/drawing/Drawing.h
@@ -286,22 +286,22 @@ public:
 
     ImageId() = default;
 
-    explicit ImageId(uint32_t index)
+    explicit constexpr ImageId(uint32_t index)
         : _value(index & MASK_INDEX)
     {
     }
 
-    ImageId(uint32_t index, uint8_t primaryColourOrPalette)
+    constexpr ImageId(uint32_t index, uint8_t primaryColourOrPalette)
         : ImageId(ImageId(index).WithPrimary(primaryColourOrPalette))
     {
     }
 
-    ImageId(uint32_t index, colour_t primaryColour, colour_t secondaryColour)
+    constexpr ImageId(uint32_t index, colour_t primaryColour, colour_t secondaryColour)
         : ImageId(ImageId(index).WithPrimary(primaryColour).WithSecondary(secondaryColour))
     {
     }
 
-    ImageId(uint32_t index, colour_t primaryColour, colour_t secondaryColour, colour_t tertiaryColour)
+    constexpr ImageId(uint32_t index, colour_t primaryColour, colour_t secondaryColour, colour_t tertiaryColour)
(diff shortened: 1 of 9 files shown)


src/openrct2/drawing/Drawing.h
```diff
@@ -230,6 +230,181 @@ struct rct_size16
...
+
+    ImageId WithTertiary(colour_t tertiary)
+    {
+        ImageId result = *this;
+        result._value &= ~FLAG_PRIMARY;
+        if (!(_value & FLAG_SECONDARY))
+        {
+            // Tertiary implies primary and secondary, so if colour was remap (8-bit primary) then
+            // we need to zero the secondary colour.
+            result._value &= ~MASK_SECONDARY;
```
duncanspumpkin: Is this correct. Secondary mask would overlap remap mask

IntelOrca: I think so, you can't have a remap colour as well as a secondary or tertiary colour. It will zero some of the high bits of the remap colour.

src/openrct2/drawing/Drawing.h
```diff
@@ -230,6 +230,181 @@ struct rct_size16
...
+    {
+    }
+
+    constexpr ImageId(uint32_t index, colour_t primaryColour, colour_t secondaryColour)
+        : ImageId(ImageId(index).WithPrimary(primaryColour).WithSecondary(secondaryColour))
+    {
+    }
+
+    constexpr ImageId(uint32_t index, colour_t primaryColour, colour_t secondaryColour, colour_t tertiaryColour)
+        : ImageId(ImageId(index).WithPrimary(primaryColour).WithSecondary(secondaryColour).WithTertiary(tertiaryColour))
```
ZehMatt: This looks really weird.

ZehMatt:
I think it would make the code a lot easier if we just separate image id, color, tertiary color. The only time where we need it packed again is during export, all the bit magic is superfluous if we encapsulate the data.

duncanspumpkin:
That's the idea. This is just to keep the mods a little at a time.

ZehMatt:
We could also use bit fields in this case as we don't care how its stored internally as long import/export is doing it properly, we could keep the same size and still have different variables segmented bit wise. I mean it already has FromUInt32 all it would need is ToUInt32 which packs them back together, all the constructor overloads could also directly assign the fields then without chaining a bunch of calls together.

src/openrct2/drawing/Drawing.h
IntelOrca: I was quite chuffed with this.

IntelOrca:
@ZehMatt I see your point, but they are all minor implementation details of ImageId that can be changed at any time without affecting the rest of the code base. Are you satisifed with the outter API?

ZehMatt:
Yeah it looks good

PR #9944, opened 2019-08-27
Rename some bits to allow unity builds of libopenrct2

janisozaur:
Some names are clashing when doing a unity build, renaming them solves
the problem.

### DIFF
--- a/src/openrct2/interface/Screenshot.cpp
+++ b/src/openrct2/interface/Screenshot.cpp
@@ -498,61 +498,61 @@ int32_t cmdline_for_gfxbench(const char** argv, int32_t argc)
     return 1;
 }
 
-static void ApplyOptions(const ScreenshotOptions* _options, rct_viewport& viewport)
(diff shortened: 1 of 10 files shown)


src/openrct2/interface/Screenshot.cpp
```diff
@@ -498,61 +498,61 @@ int32_t cmdline_for_gfxbench(const char** argv, int32_t argc)
     return 1;
 }
 
-static void ApplyOptions(const ScreenshotOptions* options, rct_viewport& viewport)
+static void ApplyOptions(const ScreenshotOptions* _options, rct_viewport& viewport)
```
IntelOrca: Underscore prefix are for static declarations outside functions.

janisozaur: Is it better now?

IntelOrca: yes

PR #9946, opened 2019-08-28
Shrink PNG Files

mic-max:


AaronVanGeffen:
I'm not sure this is worth doing at this point, since the original binaries are already in the repository anyway. `git would be storing both copies.`

As is, this *increases* the file size for several files, by the way. (distribution/windows/top.png, distribution/windows/welcome.png)

IntelOrca:
Something to consider though if we ever move the files into another repository.

duncanspumpkin:
What about moving these files to OpenGraphics repo?

Gymnasiast:
I thought @janisozaur already ran `pngoptimize` or something like that on the PNGs.

janisozaur:
Apparently not on these. I usually check PRs introducing new pngs for optimization, that must've slipped.
As already mentioned, at this point it will only bloat the repo by storing images twice for little to no benefit in resulting artifact size (I haven't checked, just my assumption).

Gymnasiast:
Since this does not save a great deal of space (most of the saving are expressed in _bytes_) and will cause the repo size to increase, I'm afraid I'll have to close this PR.

Don't let this keep you from trying to help the project, we appreciate the effort nonetheless. :-)

PR #9952, opened 2019-08-30
Refactor/use that space

Gymnasiast:
This switches over the Large Scenery and Surface tile elements to use the available space, using https://github.com/IntelOrca/OpenRCT2/blob/new-save-format-2/docs/save-format.md#tile-element as a guide.

With this and a few tweaks (not in this branch), I can have it load all surface types, making BPB go from this:
Blackpool Pleasure Beach 2019-08-30 23-19-54
to this:
Blackpool Pleasure Beach 2019-08-30 23-20-00
For comparison, this is how RCT1 renders it:
afbeelding

### DIFF
--- a/src/openrct2/actions/RideDemolishAction.hpp
+++ b/src/openrct2/actions/RideDemolishAction.hpp
@@ -295,7 +295,7 @@ private:
             if (it.element->GetType() != TILE_ELEMENT_TYPE_TRACK)
                 continue;
 
-            if (it.element->AsTrack()->GetRideIndex() != _rideIndex)
+            if (it.element->AsTrack()->GetRideIndex() != (ride_idnew_t)_rideIndex)
                 continue;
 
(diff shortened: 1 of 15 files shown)


janisozaur:
I see no differences between the images, can you highlight them?

IntelOrca:
The roof of the building in the bottom left corner is different.

ocalhoun6:
Another difference between the two is that it dropped from 154fps to 59fps. Is that just a fluke, or does this branch have performance issues?

Gymnasiast:
@janisozaur @IntelOrca The different around the Dodgems and under the railway lines is also different.

@ocalhoun6 It's not caused by this branch. FPS tends to fluctuate, and this is also a debug build. If anything, this should be a hair faster (since it doesn't have to bitmask and bitshift), but don't expect it to be very noticeable.

AaronVanGeffen:
The network version still needs to be incremented.

PR #9954, opened 2019-08-31
Clarify first Windows build instructions in Readme

qcz:
If you clone the OpenRCT2 repository and follow the first build instructions for Windows, it will be successful, but it fails if you open it in Visual Studio because it won't find any external libraries. This is because VS selects Win32 as the target platform by default, but msbuild will run with x64 as the target platform. It is not obvious for the first sight that this is the cause of the failing build.

I clarified the build instructions, so other developers won't have this problem at the beginning.


duncanspumpkin:
Tbh we should change the default as well

PR #9959, opened 2019-09-01
Convert more code to use proper Direction symbols

richard-fine:
Convert more code to using the Direction type and associated symbols. In this PR specifically:

* Swap a bunch of `uint8_t` to `Direction` where it is semantically correct to do so
* Replace -1 and 0xFF with `INVALID_DIRECTION` constant
* Replace some `obj->GetDirectionWithOffset(2)` with `direction_reverse(obj->GetDirection())` as it's easier to read
* Introduce a couple more helper symbols - `direction_next` and `direction_prev`, and `ALL_DIRECTIONS` constant - for use in followup PR

Also, make the Endian-swapping code able to handle any arbitrary type, not just uint_t. This allows us to convert some typedef'd uint_ts (such as Direction) to little wrapper structs (e.g. `struct Direction { uint8_t value; }`) so that we get compiler-enforced type correctness, while still allowing the Endian code to swap bytes exactly as if they were still just typedefs. Added tests for the endian code, plus for this new behaviour.

### DIFF
--- a/src/openrct2/actions/FootpathPlaceAction.hpp
+++ b/src/openrct2/actions/FootpathPlaceAction.hpp
@@ -88,7 +88,7 @@ public:
             return MakeResult(GA_ERROR::DISALLOWED, STR_CANT_BUILD_FOOTPATH_HERE, STR_TOO_HIGH);
         }
 
-        if (_direction != 0xFF && _direction > 15)
+        if (_direction != INVALID_DIRECTION && !direction_valid(_direction))
         {
             log_error("Direction invalid. direction = %u", _direction);
             return MakeResult(GA_ERROR::INVALID_PARAMETERS, STR_CANT_BUILD_FOOTPATH_HERE);
--- a/src/openrct2/actions/FootpathPlaceAction.hpp
+++ b/src/openrct2/actions/FootpathPlaceAction.hpp
@@ -28,11 +28,11 @@ private:
     CoordsXYZ _loc;
     uint8_t _slope;
     uint8_t _type;
-    uint8_t _direction = 0xFF;
+    Direction _direction = INVALID_DIRECTION;
 
 public:
     FootpathPlaceAction() = default;
-    FootpathPlaceAction(CoordsXYZ loc, uint8_t slope, uint8_t type, uint8_t direction = 0xFF)
+    FootpathPlaceAction(CoordsXYZ loc, uint8_t slope, uint8_t type, Direction direction = INVALID_DIRECTION)
         : _loc(loc)
         , _slope(slope)
         , _type(type)
@@ -127,7 +127,7 @@ public:
 
         if (!(GetFlags() & GAME_COMMAND_FLAG_GHOST))
         {
-            if (_direction != 0xFF && !gCheatsDisableClearanceChecks)
+            if (_direction != INVALID_DIRECTION && !gCheatsDisableClearanceChecks)
             {
                 // It is possible, let's remove walls between the old and new piece of path
                 auto zLow = _loc.z / 8;
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -1740,7 +1740,7 @@ static void get_ride_queue_end(TileCoordsXYZ& loc)
     if (!found)
         return;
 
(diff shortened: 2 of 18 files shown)


src/openrct2/actions/FootpathPlaceAction.hpp
```diff
@@ -88,7 +88,7 @@ DEFINE_GAME_ACTION(FootpathPlaceAction, GAME_COMMAND_PLACE_PATH, GameActionResul
             return MakeResult(GA_ERROR::DISALLOWED, STR_CANT_BUILD_FOOTPATH_HERE, STR_TOO_HIGH);
         }
 
-        if (_direction != 0xFF && _direction > 15)
+        if (_direction != INVALID_DIRECTION && !direction_valid(_direction))
```
richard-fine: This is a change in behaviour - direction_valid() restricts to 0-3 - but as discussed on Gitter, the new behaviour is more correct.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -1971,7 +1971,7 @@ int32_t guest_path_finding(Guest* peep)
     if (!peep->HasFood() && (scenario_rand() & 0xFFFF) >= 2184)
     {
         uint8_t adjustedEdges = edges;
-        for (int32_t chosenDirection = 0; chosenDirection < 4; chosenDirection++)
+        for (Direction chosenDirection = 0; direction_valid(chosenDirection); chosenDirection++)
```
duncanspumpkin: May as well use your range based for loop that you created.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -1889,7 +1889,7 @@ int32_t guest_path_finding(Guest* peep)
         /* If this tileElement is adjacent to any non-wide paths,
          * remove all of the edges to wide paths. */
         uint8_t adjustedEdges = edges;
-        for (int32_t chosenDirection = 0; chosenDirection < 4; chosenDirection++)
+        for (Direction chosenDirection = 0; direction_valid(chosenDirection); chosenDirection++)
```
duncanspumpkin: As below.

richard-fine:
I've got the range-based for loops stashed locally, but thought possibly I should put this PR through first so that it doesn't get too big and cause endless merge conflicts - I can switch to the range-based for loops now if you prefer.

duncanspumpkin:
I don't think there will be much complaints about having that in this PR.

duncanspumpkin:
We shouldn't really be using Upper case for defining these variable names. I'm not sure what would be best as we don't have a namespace `Direction`. It could be suggested that `Direction` should become a `struct` and reverse could become a method of that and also group these constants within that. I'm not sure what would be best but apart from that all changes look fine.

richard-fine:
@duncanspumpkin which variables are you talking about? Do you mean types? I was following the lead set by the other types in Location.hpp - LocationXY8, TileCoordsXY, TileCoordsXYZD, MapRange, etc. I don't think the fact that it is a typedef (for now) rather than a struct should really change anything.

I do agree that it might be good, at a future point, for Direction to become a struct and these helper methods to become members - that's actually exactly how I have been working locally to identify places where the Direction type can/should be used. Once I've converted most of the usages over and identified the helpers that we need, it should be relatively easy to transform from a typedef into a full struct.

richard-fine:
Oh, are you talking about `INVALID_DIRECTION`?

richard-fine:
I spent a few hours yesterday playing around with making `Direction` be a full struct, with implicit conversion to/from `uint8_t`, and putting the 'constant' directions into a `Directions` namespace.

It works, but the big downside is that when `direction_reverse` and friends are members of `Direction`, then the types you're dealing with _must_ be `Direction`, i.e. if you have a `uint8_t dir` then you can no longer do `direction_reverse(dir)`, you _must_ change the type (or cast it), and the implicit conversions to/from Direction don't help you. In theory this is great because it forces us to be even more rigorous about using the correct types, but it does also mean I run smack into dealing with all the places where we are dealing with directions that are using more than just the bottom two bits (tracks + some sprite stuff, as far as I can see).

I'll keep pursuing it because I think it will be better in the long run, but I think it's going to require cleaning up these other use cases in the process...

Issue #9957, opened 2019-08-31
When using 'no money' cheat, guests complain of running out of cash.


leftie1989:
**OS:** Windows 10 Pro, version 1903 / 18362.295
**Version:** 0.2.3
**Commit/Build:** ac7a1eb

In game, i launched the Blackpool Pleasure Beach scenario from the 'real parks' tab, decdided to disable money. A while later, a large proportion of guests reported having no money when approaching a shop. I managed to reproduce this using another scenario, however the images and save provided is from the first time i encountered this, on that BPB scenario.
When this happens, they also do NOT use cash machines to rectify their issues, even if you use the cheat to give everyone $1000, they'll quickly go back to complainining of having zero.
may be of value, my game is a copy of RCT2 from Steam, and includes all expansions, it also includes RCT1 from GOG, again with both expansions. 

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Specific to multiplayer? 

**Steps to reproduce:**
1. Launch a scenario and use the no money cheat
2. Play it for long enough, and at least in my experience, and i managed to reproduce it, guests will eventually complain of having no money.

**Screenshots / Video:**

![2019-09-01 00_18_26-OpenRCT2](https://user-images.githubusercontent.com/8727023/64069968-0d0d4680-cc4e-11e9-8ae5-7d4d3ca8479f.png)
![2019-09-01 00_15_38-OpenRCT2](https://user-images.githubusercontent.com/8727023/64069969-0d0d4680-cc4e-11e9-8766-80e9081aef4b.png)

**Save game:**

[BPB.zip](https://github.com/OpenRCT2/OpenRCT2/files/3562922/BPB.zip)

PR #9961, opened 2019-09-01
Added check for No Money cheat for issue #9957

gill984:
Added a few checks in Guest.cpp. There might be some other spots that could also use this check. Made sure cmake and make ran but didn't do other testing.

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1549,7 +1549,7 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
 loc_69B119:
     if (!hasVoucher)
     {
-        if ((price != 0) && !(gParkFlags & PARK_FLAGS_NO_MONEY))
+        if (price != 0 && !(gParkFlags & PARK_FLAGS_NO_MONEY))
         {
             if (cash_in_pocket == 0)
             {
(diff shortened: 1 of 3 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -1551,12 +1551,12 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
     {
         if (price != 0)
         {
-            if (cash_in_pocket == 0)
+            if ((cash_in_pocket == 0) && !(gParkFlags & PARK_FLAGS_NO_MONEY))
```
Gymnasiast: The park flags check can be moved up to line 1552, rather than checking it twice in lines 1554 and 1559.

gill984: I've consolidated the check to only be on line 1552 now as you said. I pushed this commit. Hopefully I am doing things correctly.

src/openrct2/peep/Guest.cpp
```diff
@@ -1549,7 +1549,7 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
 loc_69B119:
     if (!hasVoucher)
     {
-        if (price != 0)
+        if ((price != 0) && !(gParkFlags & PARK_FLAGS_NO_MONEY))
```
Gymnasiast: This is a nitpick, but there is no need to put parentheses around `price != 0`.
Same applies to line 2644.

Issue #9955, opened 2019-08-31
Resizing map while game is paused may result in endless bug loop causing game to freeze


Solioss:
**OS:** Windows 10 (x86-64)
**Version:** 0.2.2
**Commit/Build:** 298c9f5

(This was brought from the OpenRCT2 forums : https://forums.openrct2.org/topic/4479-resizing-map-while-game-is-paused-may-result-in-endless-bug-loop-causing-game-to-freeze/ )

BUG : When shrinking the map with sandbox tools while the game is paused, game may start spamming the "Construction not possible while game is paused" error with the corresponding sound effect, resulting in loud constant noise and game freezing completely. This does not happen in the scenario editor (which, unsurprisingly, does not have a pause button). 

- [ ] Reproducible in RCT2 (vanilla)? **No** (requires use of cheats specific to OpenRCT2)
- [ ] Specific to multiplayer? **No**

**Steps to reproduce:**
1. Open the game.
2. Open any map containing objects (e. g. any scenario or saved game with progress). I've tested on "Build you own Six Flags park", "Six Flags Belguim" and "Crazy Castle" maps. The map's objects (trees, decorations, paths...) should preferably be close to its borders.
3. Pause the game.
4. Make sure the "Allow construction while paused" cheat is disabled.
5. Enable sandbox mode.
6. Open the map menu to access map editing tools.
6.1. Recommended before next step : Lower your volume (yeah, I jumped when it happened to me the first time)
7. Reduce the size of the map several times. It can go normally until the game has to delete a tile which has an object on it, in which case it will attempt to delete the object, causing the "Cannot build while game is paused" error and, apparently, making the game retry to delete the object again and again, resulting in a spam of the error's sound effect and in a complete game freeze. The game may or may not have had the time to display its last words in its ultimate render frame before it freezes : "Can't construct this here... Construction not possible while game is paused!"

**Known temporary fixes** (use any) : 
- Make sure the "Allow construction while game is paused" cheat is enabled
- Do not pause the game while shrinking the map
- Clean the map's edges before shrinking it

ZehMatt:
I can not reproduce this, should be fixed. I do recall we converted this to a game action to work properly.

PR #9969, opened 2019-09-06
Allow sandbox mode map resize in pause mode #9955

crorvig:
Since map_remove_out_of_range_elements is the culprit for causing the infinite loop, I toggled BuildInPauseMode to be on for the duration of the call.

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -1541,28 +1541,33 @@ void map_restore_provisional_elements()
 void map_remove_out_of_range_elements()
 {
     int32_t mapMaxXY = gMapSizeMaxXY;
-    //#9955 ensure that we can remove elements
+    // Ensure that we can remove elements
     bool buildState = gCheatsBuildInPauseMode;
     CheatsSet(CheatType::BuildInPauseMode, true);
 
-    for (int32_t y = 0; y < (MAXIMUM_MAP_SIZE_TECHNICAL * 32); y += 32)
+    // Since we might lack cheat permission, check if the cheat has really been turned on.
+    if (gCheatsBuildInPauseMode)
     {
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Map.cpp
```diff
@@ -1540,6 +1540,9 @@ void map_restore_provisional_elements()
 void map_remove_out_of_range_elements()
 {
     int32_t mapMaxXY = gMapSizeMaxXY;
+    //#9955 ensure that we can remove elements
+    bool buildState = gCheatsBuildInPauseMode;
+    CheatsSet(CheatType::BuildInPauseMode, true);
 
     for (int32_t y = 0; y < (MAXIMUM_MAP_SIZE_TECHNICAL * 32); y += 32)
```
Gymnasiast: This can still go wrong if the user does not have cheats privileges. Perhaps it should check `gCheatsBuildInPauseMode` after CheatsSet, moving the for loop into an if.

crorvig: There's also an alternative where we can just unpause the game for that tick.  Are there pause / unpause permissions for multiplayer as well?  The reason I didn't implement the fix by unpausing was to reduce any side effects on the game.  

For now I'll implement your suggestion and see if that fixes it.

ZehMatt:
Did you actually test multiplayer? This will still freeze the client because the deletion will actually happen later and the loop expects the element to be gone right away.

ZehMatt:
This apparently broke multiplayer, when joining a server you can crash or get an error that you don't have permission, it should not use game actions to temporarily set the state.

Gymnasiast:
Let's revert it, then. Do you want to do it or shall I do it?

ZehMatt:
I think I will modify this slightly but not exactly revert it.

PR #9979, opened 2019-09-12
Fix #9683: Cannot raise/lower water level if part of the tool's area of effect is off of the map

### LINKED ISSUE #9683: Cannot raise/lower water level if part of the tool's area of effect is off of the map
**OS:** Windows 8
**Version:** 0.2.3
**Commit/Build:** 8378790

I'm pretty sure this used to work properly before, but doesn't anymore.
If you try to raise/lower the water level while part of the tool's area is off of the map, it just won't do it. No error messages are given; the tool just doesn't do anything. This happens in both the scenario editor and regular play.

If I remember right, vanilla would give you an error message if you tried this, but previous versions of open allowed you to use the tool like normal, even when part of it was off the map.

Land tool is not affected and works properly on edge of map.

**Steps to reproduce:**
1. Open water tool.
2. Change the size to at least 2x2, preferably much larger.
3. Try to raise/lower the water level with part of the tool's area of effect overlapping the edge of the map.
4. The water level does not rise or lower, and no error messages are given.

**Dump file**

**Screenshots / Video:**

Unnamed park 2019-07-29 22-58-35
Forest Frontiers 2019-07-29 23-13-32

**Save game:**

Spacerat:
Fixes #9683 

The water raise/lower actions failed at the edge of the map because `GetLowestHeight` and `GetHighestHeight` were checking outside of the map. This passes in the `validRange` already computed to fix that.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -14,6 +14,7 @@
 - Fix: [#9625] Show correct cost in scenery selection.
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
 - Fix: [#9675] Guest entry point limit can be bypassed in scenario editor.
+- Fix: [#9683] Cannot raise water level if part of the tool's area of effect is off of the map.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
 - Fix: [#9729] Peeps do not take into account height difference when deciding to pathfind to a ride entrance (original bug).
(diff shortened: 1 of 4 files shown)


Spacerat:
In fact, _lowering_ at the edge was never actually broken, so technically the change for `WaterLowerAction` isn't needed. It seems sensible to me to keep the change there though.

Spacerat:
done

PR #9986, opened 2019-09-14
Fix ride station index related bugs.

duncanspumpkin:
There are still a few issues. Need to make the same changes to the import code

### DIFF
--- a/src/openrct2/rct12/RCT12.cpp
+++ b/src/openrct2/rct12/RCT12.cpp
@@ -190,7 +190,11 @@ uint8_t RCT12TrackElement::GetColourScheme() const
 
 uint8_t RCT12TrackElement::GetStationIndex() const
 {
-    return (sequence & RCT12_TRACK_ELEMENT_SEQUENCE_STATION_INDEX_MASK) >> 4;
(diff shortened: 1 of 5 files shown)


src/openrct2/rct12/RCT12.cpp
```diff
@@ -190,7 +190,11 @@ uint8_t RCT12TrackElement::GetColourScheme() const
 
 uint8_t RCT12TrackElement::GetStationIndex() const
 {
-    return (sequence & RCT12_TRACK_ELEMENT_SEQUENCE_STATION_INDEX_MASK) >> 4;
+    if (trackType == TRACK_ELEM_END_STATION || trackType == TRACK_ELEM_BEGIN_STATION || trackType == TRACK_ELEM_MIDDLE_STATION)
```
Gymnasiast: I'd still prefer a helper method for this, even if it just involves copying it from the regular tile element.

duncanspumpkin:
I was just thinking this approach is the best for preventing duplication and makes it easier for any other person to work out the tile element quirks

Gymnasiast:
Well, I'm ok with it. If you address the check for station elements (the line note), it's good to go AFAIC.

Gymnasiast:
I have merged it now to prevent more corruption of people's maps.

PR #9987, opened 2019-09-14
Fix #9970: Wait for quarter load fails

### LINKED ISSUE #9970: Wait for quarter load fails
**OS:** Windows 10
**Version:** 0.2.3-324
**Commit/Build:** 1d6e4d7

Wait for quarter load does not wait for any guests. Trains will still leave as if this option was not selected (even empty).

**Steps to reproduce:**
1. Open RCT1LL scenario Megaworld Park
2. Observe Viper (in the sand area)
Or build any coaster and enable 'wait for quarter load'. (I checked multiple coasters in different parks, but coincidentally only corkscrews)

aw20368:
Two commits:
* First commit only fixes the bug. Brings behaviour in line with vanilla.
* Second commit contains a refactor. This change will make rides will actually wait for >= 25% load (described in #9970), unlike vanilla.

Will squash if both commits approved.
Assuming network version bump + changelog entry (can repro in v0.0.5).

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -2363,60 +2363,24 @@ static void vehicle_update_waiting_for_passengers(rct_vehicle* vehicle)
                 return;
             }
 
+            // any load: load=4 , full: load=3 , 3/4s: load=2 , half: load=1 , quarter: load=0
             uint8_t load = ride->depart_flags & RIDE_DEPART_WAIT_FOR_LOAD_MASK;
-            if (load == 3)
-            {
-                train_ready_to_depart(vehicle, num_peeps_on_train, num_used_seats_on_train);
-                return;
-            }
 
-            uint8_t three_quater_seats = (3 * num_seats_on_train) / 4;
-            if (three_quater_seats != 0 && num_peeps_on_train >= three_quater_seats)
-            {
+            // We want to wait for ceiling((load+1)/4 * num_seats_on_train) peeps, the +3 below is used instead of
+            // ceil() to prevent issues on different cpus/platforms with floats. Note that vanilla RCT1/2 rounded
+            // down here; our change reflects the expected behaviour for waiting for a minimum load target (see #9987)
+            uint8_t peepTarget = ((load + 1) * num_seats_on_train + 3) / 4;
+
+            if (load == 4) // take care of "any load" special case
+                peepTarget = 1;
+
+            if (num_peeps_on_train >= peepTarget)
                 vehicle->update_flags |= VEHICLE_UPDATE_FLAG_TRAIN_READY_DEPART;
-                train_ready_to_depart(vehicle, num_peeps_on_train, num_used_seats_on_train);
-                return;
-            }
 
-            if (load == 2)
-            {
-                train_ready_to_depart(vehicle, num_peeps_on_train, num_used_seats_on_train);
-                return;
-            }
-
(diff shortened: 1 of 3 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -2363,61 +2363,20 @@ static void vehicle_update_waiting_for_passengers(rct_vehicle* vehicle)
...
-            }
-
-            uint8_t three_quater_seats = (3 * num_seats_on_train) / 4;
-            if (three_quater_seats != 0 && num_peeps_on_train >= three_quater_seats)
-            {
-                vehicle->update_flags |= VEHICLE_UPDATE_FLAG_TRAIN_READY_DEPART;
-                train_ready_to_depart(vehicle, num_peeps_on_train, num_used_seats_on_train);
-                return;
-            }
+            double peepTarget = (((double)load + 1) * num_seats_on_train) / 4;
```
ZehMatt: This can actually yield different results on different cpus/platforms, use integers only.

Gymnasiast:
I have decided to cherry-pick the fix while we sort out the refactor. Could you rebase this PR?

aw20368:
Done. Rebased and addressed review.

aw20368:
Does the refactor need another network version bump? It'll change when a train can leave (by a little bit).

ZehMatt:
If it changes the mechanics then it should be done or the game will diverge during multiplayer if you have two different versions.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -2363,60 +2363,20 @@ static void vehicle_update_waiting_for_passengers(rct_vehicle* vehicle)
...
-                return;
-            }
-
-            if (num_seats_on_train / 2 != 0 && num_peeps_on_train >= num_seats_on_train / 2)
-            {
-                vehicle->update_flags |= VEHICLE_UPDATE_FLAG_TRAIN_READY_DEPART;
-                train_ready_to_depart(vehicle, num_peeps_on_train, num_used_seats_on_train);
-                return;
-            }
+            uint8_t peepTarget = ((load + 1) * num_seats_on_train + 3) / 4;
```
Gymnasiast: `+3` looks very messy, and isn't in the original code.

aw20368: The intent was to compute ceiling( (load+1)/4 * num_seats), and the +3 is a trick to bypass the need of floats/doubles.

It's not in the original code because it did not actually wait for a quarter load. In #9970 there's a screenshot of vanilla RCT1 starting an 18-car train with 4 occupants at "wait for quarter load".

I mean, it's a bit of an argument over semantics, but I expected that 18-car train to leave at 6 occupants rather than 4. I could be convinced otherwise (removing the +3 reverts the behaviour to vanilla and doesn't affect the rest of the logic).

Gymnasiast: I wrote out the results for a train with room for 18 people:

Load | "Real" result | RCT1/2 | This PR
--- | --- | --- | ---
Quarter | 4.5 | 4 | 5
Half | 9 | 9 | 9
Three q. | 13.5 | 13 | 14

Since the option is _minimum_ load, rounding up is indeed more in line with what people expect. I am assuming that the compiler will first do the multiplication and only then the addition.

I'd like to see the following things:
- A comment that explains why the +3 is there
- A changelog entry: `- Improved: [#9987] Minimum load rounding.`
- A network version bump

aw20368: Done, force pushed.

PR #9990, opened 2019-09-15
Fix #9675: Inconsistent number of guest entry points

### LINKED ISSUE #9675: Inconsistent number of guest entry points
**Version:** 0.2.3
**Commit/Build:** 3ac4311

https://forums.openrct2.org/topic/4235-inconsistent-number-of-guest-spawn-points/

> Originally there was a maximum of 2 guest spawn ﻿points. Recently in v0.2.1-1020 and v0.2.2-255, I noticed that I could place at least 8 guest spawn points. Then after I removed spawn points by shrinking the map, I was able to do this﻿:﻿﻿
> 

Following these steps reported on the forums I was able to place place 32 guest entry points.

**Steps to reproduce:**
1. In the scenario editor or on a map with sandbox mode enabled, place some guest entry points near the edge that you can shrink/enlarge.
2. After having placed some, shrink the map so that the guest entry points disappear
3. Enlarge the map again
4. Place new guest entry points

Expected behaviour: After having placed 8 points, the first one would disappear.
Current behaviour: You can place up to 16 points before the first disappears the first time, and up to 32 the second time.

Spacerat:
Fixes #9675 by removing old spawns from `gPeepSawn` when the limit is reached instead of invalidating them.

----

`_nextPeepSpawnIndex` invalidated old spawns[0] instead of removing them, while `gPeepSpawn` grew indefinitely[1]. If _all_ spawns were erased while shrinking the map[2], `_nextPeepSpawnIndex` would cycle back to zero and start invalidating previously invalidated spawns instead of the oldest valid ones.

[0]
https://github.com/OpenRCT2/OpenRCT2/blob/1efc774483ccfa749037c00a36d2f86a058ed942/src/openrct2/actions/PlacePeepSpawnAction.hpp#L104-L111

[1]
https://github.com/OpenRCT2/OpenRCT2/blob/1efc774483ccfa749037c00a36d2f86a058ed942/src/openrct2/actions/PlacePeepSpawnAction.hpp#L127

[2] https://github.com/OpenRCT2/OpenRCT2/blob/1efc774483ccfa749037c00a36d2f86a058ed942/src/openrct2/world/Map.cpp#L1711-L1718

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -10,11 +10,12 @@
 - Fix: [#9179] Crash when modifying a ride occasionally.
 - Fix: [#9533] Door sounds not playing.
 - Fix: [#9574] Text overflow in scenario objective window when using CJK languages.
+- Fix: [#9603] Don't render audio when master volume is turned off.
 - Fix: [#9625] Show correct cost in scenery selection.
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
+- Fix: [#9675] Guest entry point limit can be bypassed in scenario editor.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
 - Fix: [#9729] Peeps do not take into account height difference when deciding to pathfind to a ride entrance (original bug).
-- Fix: [#9603] Don't render audio when master volume is turned off.
 - Fix: [#9926] Africa - Oasis park has wrong peep spawn (original bug).
 - Improved: [#9466] Add the rain weather effect to the OpenGL renderer.
 
--- a/src/openrct2/actions/PlacePeepSpawnAction.hpp
+++ b/src/openrct2/actions/PlacePeepSpawnAction.hpp
@@ -19,8 +19,6 @@
 #include "../world/Surface.h"
 #include "GameAction.h"
 
-static int32_t _nextPeepSpawnIndex = 0;
-
 DEFINE_GAME_ACTION(PlacePeepSpawnAction, GAME_COMMAND_PLACE_PEEP_SPAWN, GameActionResult)
 {
 private:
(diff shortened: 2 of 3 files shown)


distribution/changelog.txt
```diff
@@ -83,6 +83,7 @@
 - Fix: [#9411] Ad campaigns end too soon.
 - Fix: [#9476] Running `simulate` command on park yields `Completed: (null)`.
 - Fix: [#9520] Time Twister object artdec29 conversion problem.
+- Fix: [#9675] Guest entry point limit can be bypassed in scenario editor
```
Broxzier: Please add a full stop at the end.

src/openrct2/actions/PlacePeepSpawnAction.hpp
```diff
@@ -100,17 +98,12 @@ DEFINE_GAME_ACTION(PlacePeepSpawnAction, GAME_COMMAND_PLACE_PEEP_SPAWN, GameActi
...
-            _nextPeepSpawnIndex = (peepSpawnIndex + 1) % (int32_t)gPeepSpawns.size();
-
-            // Before the new location is set, clear the old location
-            int32_t prevX = gPeepSpawns[peepSpawnIndex].x;
-            gPeepSpawns[peepSpawnIndex].x = PEEP_SPAWN_UNDEFINED;
-
-            map_invalidate_tile_full(prevX, gPeepSpawns[peepSpawnIndex].y);
+            auto oldestSpawn = gPeepSpawns.begin();
+            gPeepSpawns.erase(oldestSpawn);
+            map_invalidate_tile_full((*oldestSpawn).x, (*oldestSpawn).y);
```
Broxzier: `vector::erase` invalidates all iterators at and after the given iterator. You need to store the coordinates before removing it.

distribution/changelog.txt
Spacerat: I also added the entry to the wrong section!

src/openrct2/actions/PlacePeepSpawnAction.hpp
Spacerat: 👍 made the fix.

Out of interest: this would also have been solved by swapping the `erase` and `invalidate` lines, but would that have been a safe thing to do? I noticed that `map_invalidate_tile_full` calls tend to be the last things functions do, but the conditions where that really matters are not entirely clear.

Broxzier: In this case, the order doesn't really matter since it simply tells the game what screen areas to redraw on the next frame. Though invalidating them after having made the changes it the way to go, in case the invalidating process gets used by something else later on.

Spacerat: Ah, that makes sense. Thanks for explaining!

Spacerat:
I pushed another commit up - I think it's a bit more idiomatic/clean? I'll admit my C++ is fairly rusty. Let me know if you prefer that and I either can squash or delete it.

Broxzier:
Either will do. Eventually, functions like this will take `TileCoords` objects rather than X/Y.

I just noticed that the changelog line should be moved up by one line, because its issue ID is lower. Then I think this is good to go!

distribution/changelog.txt
```diff
@@ -10,11 +10,12 @@
 - Fix: [#9179] Crash when modifying a ride occasionally.
 - Fix: [#9533] Door sounds not playing.
 - Fix: [#9574] Text overflow in scenario objective window when using CJK languages.
+- Fix: [#9603] Don't render audio when master volume is turned off.
```
Broxzier: Well spotted 😄

PR #9992, opened 2019-09-15
Fix #7572: Queue paths connect to regular paths through fences

### LINKED ISSUE #7572: Queue paths connect to regular paths through fences
**OS:** Windows 10
**Version:** 0.1.3
**Commit/Build:** 15ae358

Queue paths will connect to regular paths through fences
image

Spacerat:
Fixes #7572 by checking `fence_in_the_way` for both sides of the fence.

The bug specifically occurred when queue is _removed_ (which also happens during the path construction preview), causing the newly disconnected queue piece to reconnect through a fence if the fence was not actually on its tile. 

Before:

Screen Shot 2019-09-15 at 1 47 25 PM

After:

Screen Shot 2019-09-15 at 1 44 25 PM

### DIFF
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -452,9 +452,9 @@ void footpath_interrupt_peeps(int32_t x, int32_t y, int32_t z)
 /**
  * Returns true if the edge of tile x, y specified by direction is occupied by a fence
  * between heights z0 and z1.
- * 
+ *
  * Note that there may still be a fence on the opposing tile.
- *  
+ *
  *  rct2: 0x006E59DC
  */
 bool fence_in_the_way(int32_t x, int32_t y, int32_t z0, int32_t z1, int32_t direction)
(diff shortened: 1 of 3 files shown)


src/openrct2/world/Footpath.cpp
```diff
@@ -688,16 +692,25 @@ static TileElement* footpath_get_element(int32_t x, int32_t y, int32_t z0, int32
...
 {
     if (((tileElement->AsPath()->GetEdges() & (1 << direction)) == 0) ^ (action < 0))
         return false;
-    if ((action < 0) && fence_in_the_way(x, y, tileElement->base_height, tileElement->clearance_height, direction))
-        return false;
 
     int32_t x1 = x + CoordsDirectionDelta[direction].x;
     int32_t y1 = y + CoordsDirectionDelta[direction].y;
+
+    if (action < 0) {
```
AaronVanGeffen: Formatting.

duncanspumpkin:
Appveyor is not currently working as we have exceeded our quota of space.

Gymnasiast:
Thanks for the fix!

I like the small cleanup around the function.

Issue #9902, opened 2019-08-18
Doors/Portcullis do not check to make sure doors are open causing double opens


duncanspumpkin:
`vehicle_update_scenery_door` and `vehicle_update_handle_scenery_door` do not first check that a door is open before reopening it. This used to not be an issue as only one car in a train would open/close a door. Now that all vehicles open/close doors invisible cars can reopen a door causing duplicate animations/sound.

Spongeloaf:
I'll take a shot at this one. I can reproduce it easily enough.

duncanspumpkin:
Should be quite simple to do.

PR #9996, opened 2019-09-17
Issue #9902

Spongeloaf:
RE: https://github.com/OpenRCT2/OpenRCT2/issues/9902

Portcullises now open and close correctly. I discovered that the red colored castle wall doors do not make sound when opening and closing, but it seems like an unrelated issue given that the animation works.

[Here's a save file to demonstrate](https://drive.google.com/open?id=1VPlvfJqjAiQMUvaOSzkLdQFCpAN7kcwD).

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -7460,7 +7460,7 @@ static void vehicle_update_scenery_door(rct_vehicle* vehicle)
         return;
     }
 
-    if ((vehicle->next_vehicle_on_train != SPRITE_INDEX_NULL) && (tileElement->GetAnimationFrame() == '\0'))
+    if ((vehicle->next_vehicle_on_train != SPRITE_INDEX_NULL) && (tileElement->GetAnimationFrame() == 0))
     {
         tileElement->SetAnimationIsBackwards(false);
         tileElement->SetAnimationFrame(1);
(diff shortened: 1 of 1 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7460,14 +7460,15 @@ static void vehicle_update_scenery_door(rct_vehicle* vehicle)
         return;
     }
 
-    if (vehicle->next_vehicle_on_train != SPRITE_INDEX_NULL)
+    if ((vehicle->next_vehicle_on_train != SPRITE_INDEX_NULL) && (tileElement->GetAnimationFrame() == '\0'))
```
AaronVanGeffen: Why compare a frame number to a character? It's just a `uint8_t`.

Spongeloaf: Well I'm questioning my sanity now, because yesterday when I was looking at the definitions of  GetAnimationFrame() and SetAnimationFrame(), I saw that it was taking a char and bitshifting it. 

But now I'm poking around in Visual Studio and that's clearly not the case. Anyway, I'll fix it and test it again.

Spongeloaf: Appveyor failed on an unrelated file:

```ERROR[c:\projects\openrct2\src\openrct2\world\map.cpp:188 (map_get_first_element_at)]: Trying to access element outside of range```

Otherwise I think it's good to go.

Gymnasiast: That is unrelated. It is also an error printed by OpenRCT2 itself. (Not likely to be caused by your branch either.)

Gymnasiast:
Thanks for the fix. I have squashed your commit and changed the commit message to ` Fix #9902: Doors and portcullises can be double opened` (always mention the _problem_ you fixed, not the solution).

Issue #9014, opened 2019-04-02
Verify that nullptr check exists for calls like map_get_first_element_at


ZehMatt:
A lot of issues currently come from missing nullptr checks after calls to functions like map_get_first_element_at, map_get_nth_element_at, map_get_surface_element_at, map_get_path_element_at, map_get_banner_element_at. 
Each reference should be checked if it will handle the result correctly, currently the fixes are case by case, which can be avoided by doing all of them for once.

joshtucker132:
My team can work on this issue.

lwatson2016:
What should be the error/action if the result is NULL?
I.e. if null, then ....?

duncanspumpkin:
That will depend on the function. Game actions should always return a game action result. Some functions it will be safe to just error and return some functions it will be best to do a hard crash.

janisozaur:
If nullptr is returned, it means no such element we found for the given tile and effectively there is nothing to process. In other words it means:
1. Don't crash
2. Exit the function gracefully, possibly propagating error if that makes sense.

tupaschoal:
I'm willing to work on this :)

tupaschoal:
All `map_get_banner_element_at` are currently already checked for invalid returns

tupaschoal:
I've added another PR, #10091 with other places I found. Lesson of the day is not to trust VS's "Find all References" or "Call Hierarchy" :|

tupaschoal:
I guess this can be closed now :)

PR #10013, opened 2019-09-23
Avoid dereferencing map_get_first_element_at nullptr on libopenrct2

tupaschoal:
First commit to work  towards #9014 . As is, I've double checked and fixed all places in `libopenrct2`, please review and I'll continue doing it for `libopenrct2ui` and the other functions mentioned in the issue.

Note: I did a commit per file. Thought it would be easier to track failures and to review and if it was supposed to be one, it can just be squashed.

### DIFF
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -239,6 +239,8 @@ static uint8_t footpath_element_next_in_direction(TileCoordsXYZ loc, PathElement
     nextTileElement = map_get_first_element_at(loc.x, loc.y);
     do
     {
+        if (nextTileElement == nullptr)
+            break;
         if (nextTileElement->IsGhost())
             continue;
         if (nextTileElement->GetType() != TILE_ELEMENT_TYPE_PATH)
@@ -1209,6 +1211,8 @@ Direction peep_pathfind_choose_direction(TileCoordsXYZ loc, Peep* peep)
     bool isThin = false;
     do
     {
+        if (dest_tile_element == nullptr)
+            break;
         if (dest_tile_element->base_height != loc.z)
             continue;
         if (dest_tile_element->GetType() != TILE_ELEMENT_TYPE_PATH)
@@ -1767,6 +1771,8 @@ static void get_ride_queue_end(TileCoordsXYZ& loc)
         nextTile += TileDirectionDelta[direction];
 
         tileElement = map_get_first_element_at(nextTile.x, nextTile.y);
+        if (tileElement == nullptr)
+            break;
         found = false;
         do
         {
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -1771,9 +1771,9 @@ static void get_ride_queue_end(TileCoordsXYZ& loc)
         nextTile += TileDirectionDelta[direction];
 
         tileElement = map_get_first_element_at(nextTile.x, nextTile.y);
+        found = false;
         if (tileElement == nullptr)
             break;
-        found = false;
         do
         {
             if (tileElement == firstPathElement)
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -477,7 +477,7 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
 
     TileElement* tileElement = map_get_first_element_at(nextTile.x / 32, nextTile.y / 32);
     if (tileElement == nullptr)
-        return nextDirection;
+        return 0xFF;
     do
     {
         if (tileElement->base_height != nextZ)
@@ -493,7 +493,7 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
 
     tileElement = map_get_first_element_at(nextTile.x / 32, nextTile.y / 32);
     if (tileElement == nullptr)
-        return nextDirection;
+        return 0xFF;
 
     do
     {
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -268,6 +268,8 @@ bool staff_can_ignore_wide_flag(Peep* staff, int32_t x, int32_t y, uint8_t z, Ti
 
         /* Search through all adjacent map elements */
         TileElement* test_element = map_get_first_element_at(adjac_x / 32, adjac_y / 32);
+        if (test_element == nullptr)
+            return false;
         bool pathfound = false;
         bool widefound = false;
         do
@@ -474,7 +476,8 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
     int16_t nextZ = ((peep->z + 8) & 0xFFF0) / 8;
 
     TileElement* tileElement = map_get_first_element_at(nextTile.x / 32, nextTile.y / 32);
-
+    if (tileElement == nullptr)
+        return nextDirection;
     do
     {
         if (tileElement->base_height != nextZ)
@@ -489,6 +492,8 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
     nextTile.y = (peep->y & 0xFFE0) + CoordsDirectionDelta[nextDirection].y;
 
     tileElement = map_get_first_element_at(nextTile.x / 32, nextTile.y / 32);
+    if (tileElement == nullptr)
+        return nextDirection;
 
     do
     {
@@ -1250,6 +1255,8 @@ void Staff::UpdateWatering()
         int32_t actionY = next_y + CoordsDirectionDelta[var_37].y;
 
         TileElement* tile_element = map_get_first_element_at(actionX / 32, actionY / 32);
+        if (tile_element == nullptr)
+            return;
 
         do
         {
@@ -1315,6 +1322,8 @@ void Staff::UpdateEmptyingBin()
             return;
 
         TileElement* tile_element = map_get_first_element_at(next_x / 32, next_y / 32);
+        if (tile_element == nullptr)
+            return;
 
         for (;; tile_element++)
         {
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -6313,7 +6313,7 @@ void ride_get_entrance_or_exit_position_from_screen_position(
             mapY -= CoordsDirectionDelta[direction].y;
             tileElement = map_get_first_element_at(mapX >> 5, mapY >> 5);
             if (tileElement == nullptr)
-                continue;
+                break;
             bool goToNextTile = false;
 
             do
@@ -6915,7 +6915,7 @@ void sub_6CB945(Ride* ride)
                 }
                 tileElement = map_get_first_element_at(location.x >> 5, location.y >> 5);
                 if (tileElement == nullptr)
-                    continue;
+                    break;
 
                 bool trackFound = false;
                 do
@@ -6964,7 +6964,7 @@ void sub_6CB945(Ride* ride)
                 bool trackFound = false;
                 tileElement = map_get_first_element_at(blockLocation.x >> 5, blockLocation.y >> 5);
                 if (tileElement == nullptr)
-                    continue;
+                    break;
                 do
                 {
                     if (blockLocation.z != tileElement->base_height)
@@ -7031,7 +7031,7 @@ void sub_6CB945(Ride* ride)
 
         TileElement* tileElement = map_get_first_element_at(location.x >> 5, location.y >> 5);
         if (tileElement == nullptr)
-            return;
+            continue;
         do
         {
             if (tileElement->GetType() != TILE_ELEMENT_TYPE_ENTRANCE)
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -4944,6 +4944,8 @@ static bool ride_initialise_cable_lift_track(Ride* ride, bool isApplying)
 
     bool success = false;
     TileElement* tileElement = map_get_first_element_at(location.x, location.y);
+    if (tileElement == nullptr)
+        return success;
     do
     {
         if (tileElement->GetType() != TILE_ELEMENT_TYPE_TRACK)
@@ -5182,6 +5184,8 @@ static TileElement* loc_6B4F6B(ride_id_t rideIndex, int32_t x, int32_t y)
         return nullptr;
 
     TileElement* tileElement = map_get_first_element_at(x / 32, y / 32);
+    if (tileElement == nullptr)
+        return nullptr;
     do
     {
         if (tileElement->GetType() != TILE_ELEMENT_TYPE_TRACK)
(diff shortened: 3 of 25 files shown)


Gymnasiast:
> Note: I did a commit per file. Thought it would be easier to track failures and to review and if it was supposed to be one, it can just be squashed.

Github makes reviewing easy enough not to need this. In addition, commits should contain self-contained changes and not be split out per file. But like you say, it can just be squashed, so no harm done.

duncanspumpkin:
Please don't merge into I've looked into the changes to peep.cpp that could easily cause unexpected issues.

tupaschoal:
Any news?

ZehMatt:
I will have a look at it this weekend, sorry for the delay. Maybe @duncanspumpkin beats me to it.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -1767,6 +1771,8 @@ static void get_ride_queue_end(TileCoordsXYZ& loc)
         nextTile += TileDirectionDelta[direction];
 
         tileElement = map_get_first_element_at(nextTile.x, nextTile.y);
+        if (tileElement == nullptr)
```
duncanspumpkin: Move this to after the `found = false;`

src/openrct2/peep/Staff.cpp
```diff
@@ -474,7 +476,8 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
     int16_t nextZ = ((peep->z + 8) & 0xFFF0) / 8;
 
     TileElement* tileElement = map_get_first_element_at(nextTile.x / 32, nextTile.y / 32);
-
+    if (tileElement == nullptr)
+        return nextDirection;
```
duncanspumpkin: `return 0xFF;` The litter is on an invalid tile in this situation so we should not try to path towards it. This might cause some really odd things to happen in almost every case though so I guess it doesn't ultimately matter what is done.

src/openrct2/peep/Staff.cpp
```diff
@@ -489,6 +492,8 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
     nextTile.y = (peep->y & 0xFFE0) + CoordsDirectionDelta[nextDirection].y;
 
     tileElement = map_get_first_element_at(nextTile.x / 32, nextTile.y / 32);
+    if (tileElement == nullptr)
+        return nextDirection;
```
duncanspumpkin: `return 0xFF;` The  path to the litter is on an invalid tile in this situation so we should not try to path towards it. This might cause some really odd things to happen in almost every case though so I guess it doesn't ultimately matter what is done.

src/openrct2/ride/Ride.cpp
```diff
@@ -6306,6 +6312,8 @@ void ride_get_entrance_or_exit_position_from_screen_position(
             mapX -= CoordsDirectionDelta[direction].x;
             mapY -= CoordsDirectionDelta[direction].y;
             tileElement = map_get_first_element_at(mapX >> 5, mapY >> 5);
+            if (tileElement == nullptr)
+                continue;
```
duncanspumpkin: `break;` you would likely end up in an infinite loop with `continue;`.

src/openrct2/ride/Ride.cpp
```diff
@@ -6904,6 +6914,8 @@ void sub_6CB945(Ride* ride)
                     location.y -= CoordsDirectionDelta[direction].y;
                 }
                 tileElement = map_get_first_element_at(location.x >> 5, location.y >> 5);
+                if (tileElement == nullptr)
+                    continue;
```
duncanspumpkin: `break;` avoids the infinite loop again.

src/openrct2/ride/Ride.cpp
```diff
@@ -6951,6 +6963,8 @@ void sub_6CB945(Ride* ride)
 
                 bool trackFound = false;
                 tileElement = map_get_first_element_at(blockLocation.x >> 5, blockLocation.y >> 5);
+                if (tileElement == nullptr)
+                    continue;
```
duncanspumpkin: `break;`

src/openrct2/ride/Ride.cpp
```diff
@@ -7016,6 +7030,8 @@ void sub_6CB945(Ride* ride)
         CoordsXY location = { locationCoords.x * 32, locationCoords.y * 32 };
 
         TileElement* tileElement = map_get_first_element_at(location.x >> 5, location.y >> 5);
+        if (tileElement == nullptr)
+            return;
```
duncanspumpkin: Its unlikely that there is anything else useful in the vector but there might be so use `continue;`

src/openrct2/ride/RideRatings.cpp
```diff
@@ -207,6 +207,8 @@ static void ride_ratings_update_state_2()
     int32_t trackType = gRideRatingsCalcData.proximity_track_type;
 
     TileElement* tileElement = map_get_first_element_at(x, y);
+    if (tileElement == nullptr)
+        return;
```
duncanspumpkin: will need to set `gRideRatingsCalcData.state = RIDE_RATINGS_STATE_FIND_NEXT_RIDE;` prior to returning.

src/openrct2/ride/RideRatings.cpp
```diff
@@ -313,6 +315,8 @@ static void ride_ratings_update_state_5()
...
```
duncanspumpkin: As above

src/openrct2/ride/RideRatings.cpp
```diff
@@ -1438,6 +1448,8 @@ static int32_t ride_ratings_get_scenery_score(Ride* ride)
         {
             // Count scenery items on this tile
             TileElement* tileElement = map_get_first_element_at(xx, yy);
+            if (tileElement == nullptr)
+                return 0;
```
duncanspumpkin: `continue;` there might be useful scenery on other tiles in this range.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -433,6 +433,8 @@ rct_string_id TrackDesign::CreateTrackDesignMaze(const Ride& ride)
 
     CoordsXY exitLoc = { location.x * 32, location.y * 32 };
     tileElement = map_get_first_element_at(location.x, location.y);
+    if (tileElement == nullptr)
+        return STR_NONE;
```
duncanspumpkin: `return STR_TRACK_TOO_LARGE_OR_TOO_MUCH_SCENERY;` Yeah I know it isn't the correct string but `STR_NONE` is used to indicate success in this function we need to indicate failure.

src/openrct2/ride/TrackDesignSave.cpp
```diff
@@ -525,6 +525,8 @@ static void track_design_save_select_nearby_scenery_for_tile(ride_id_t rideIndex
         for (int32_t x = cx - TRACK_NEARBY_SCENERY_DISTANCE; x <= cx + TRACK_NEARBY_SCENERY_DISTANCE; x++)
         {
             tileElement = map_get_first_element_at(x, y);
+            if (tileElement == nullptr)
+                return;
```
duncanspumpkin: `continue` its possible there might be scenery that we want to add on other tiles.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1830,6 +1830,8 @@ static void vehicle_update_measurements(rct_vehicle* vehicle)
         // Set tile_element to first element. Since elements aren't always ordered by base height,
         // we must start at the first element and iterate through each tile element.
         auto tile_element = map_get_first_element_at(x / 32, y / 32);
+        if (tile_element == nullptr)
+            return;
```
duncanspumpkin: This is okay but it is a tad messy it should really continue onto the code below like when surfaceelment is null. It will never happen though since map_get_surface_element_at already calls get_first_element_at with these coordinates.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -4671,6 +4673,8 @@ static void vehicle_update_boat_location(rct_vehicle* vehicle)
 static bool vehicle_boat_is_location_accessible(const TileCoordsXYZ& location)
 {
     TileElement* tileElement = map_get_first_element_at(location.x, location.y);
+    if (tileElement == nullptr)
+        return true;
```
duncanspumpkin: should return false as the location doesn't exist so can't be accessed. Ultimately will never happen but best to be safe.

src/openrct2/world/Footpath.cpp
```diff
@@ -114,10 +114,10 @@ static bool entrance_has_direction(TileElement* tileElement, int32_t direction)
 TileElement* map_get_footpath_element(int32_t x, int32_t y, int32_t z)
 {
     TileElement* tileElement = map_get_first_element_at(x, y);
+    if (tileElement == nullptr)
```
duncanspumpkin: revert. Previous code works fine and tends to be the standard for `map_get...`

src/openrct2/world/Footpath.cpp
```diff
@@ -2080,6 +2100,8 @@ static void footpath_fix_corners_around(int32_t x, int32_t y, TileElement* pathE
                 continue;
 
             TileElement* tileElement = map_get_first_element_at(x + xOffset, y + yOffset);
+            if (tileElement == nullptr)
+                return;
```
duncanspumpkin: `continue;`

src/openrct2/world/Map.cpp
```diff
@@ -1096,6 +1099,8 @@ void map_reorganise_elements()
         for (int32_t x = 0; x < MAXIMUM_MAP_SIZE_TECHNICAL; x++)
         {
             TileElement* startElement = map_get_first_element_at(x, y);
+            if (startElement == nullptr)
+                return;
```
duncanspumpkin: This would leak memory. Use `continue;`. If you ever hit this i suspect leaking memory would be the least of your worries though.

src/openrct2/world/Map.cpp
```diff
@@ -2178,10 +2191,10 @@ TileElement* map_get_track_element_at_of_type(int32_t x, int32_t y, int32_t z, i
 TileElement* map_get_track_element_at_of_type_seq(int32_t x, int32_t y, int32_t z, int32_t trackType, int32_t sequence)
 {
     TileElement* tileElement = map_get_first_element_at(x >> 5, y >> 5);
+    if (tileElement == nullptr)
```
duncanspumpkin: revert. Previous code works fine and tends to be the standard for `map_get...`

src/openrct2/world/MapAnimation.cpp
```diff
@@ -443,6 +459,8 @@ static bool map_animation_invalidate_large_scenery(int32_t x, int32_t y, int32_t
 
     bool wasInvalidated = false;
     tileElement = map_get_first_element_at(x >> 5, y >> 5);
+    if (tileElement == nullptr)
+        return wasInvalidated;
```
duncanspumpkin: `return true;` Returning true deletes this animation so if it has bad data it wont be hit again.

src/openrct2/world/MapAnimation.cpp
```diff
@@ -536,6 +556,8 @@ static bool map_animation_invalidate_wall(int32_t x, int32_t y, int32_t baseZ)
...
```
duncanspumpkin: `return true;`

src/openrct2/world/Scenery.cpp
```diff
@@ -197,6 +199,8 @@ void scenery_remove_ghost_tool_placement()
     {
         gSceneryGhostType &= ~SCENERY_GHOST_FLAG_1;
         TileElement* tileElement = map_get_first_element_at(x / 32, y / 32);
+        if (tileElement == nullptr)
+            return;
```
duncanspumpkin: we will want the rest of this function to be called. I know the rest will also fail but it needs to clear the ghost flags and such and that will be the most clean code. So move this into the do while and make it a break.

src/openrct2/world/TileInspector.cpp
```diff
@@ -355,6 +355,8 @@ GameActionResult::Ptr tile_inspector_sort_elements_at(CoordsXY loc, bool isExecu
     if (isExecuting)
     {
         const TileElement* const firstElement = map_get_first_element_at(loc.x / 32, loc.y / 32);
+        if (firstElement == nullptr)
+            return {};
```
duncanspumpkin: return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);

src/openrct2/world/TileInspector.cpp
```diff
@@ -787,6 +789,8 @@ GameActionResult::Ptr tile_inspector_track_base_height_offset(
 
             bool found = false;
             TileElement* tileElement = map_get_first_element_at(elem.x >> 5, elem.y >> 5);
+            if (tileElement == nullptr)
+                return {};
```
duncanspumpkin: return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);

src/openrct2/world/TileInspector.cpp
```diff
@@ -891,6 +895,8 @@ GameActionResult::Ptr tile_inspector_track_set_chain(
...
```
duncanspumpkin: return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);

src/openrct2/ride/Vehicle.cpp
tupaschoal: Ok, I'm assuming nothing needs to be done here, for now.

src/openrct2/world/Footpath.cpp
tupaschoal: But it will crash on:
```cpp
        if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH && tileElement->base_height == z)
```
If `titleElement` is ever `nullptr`. Should we assert instead?

src/openrct2/world/Map.cpp
tupaschoal: Same question as above, wouldn't it lead to a crash if `titleElement` is null?

tupaschoal:
@duncanspumpkin thanks for the thorough review and for taking the time to explain some things, I really appreciate. I've left 3 unresolved (which I commented into), once we get them decided and done I can push the remaining fixes, if any.

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: Yeah it's fine

src/openrct2/world/Footpath.cpp
duncanspumpkin: It will leave the loop before that due to the break.

src/openrct2/world/Map.cpp
duncanspumpkin: As above. It never gets to that bit of the loop due to the break

src/openrct2/world/Footpath.cpp
tupaschoal: I reverted, but am not sure I follow, there is no break within the function and thus it will go into `do while` and try to dereference `tileElement`, which likely will crash if it's a `nullptr`

duncanspumpkin: Ah my mistake I meant revert both of your changes you made. It used to have a break on nullptr in the do while.

tupaschoal: So do I keep them as is or do I move them (this and the open issue below) to within the `do...while` and `break` instead of returning?

duncanspumpkin: I want them to be within the `do while` with a `break`. This will then match the format of other `map_get...`

tupaschoal:
I've fixed the pending issues and a clang-format one,

tupaschoal:
I've rebased it, should be good to go now

PR #10017, opened 2019-09-23
Fix ghost elements influencing ride excitement

ZehMatt:


tupaschoal:
Will most probably have merge conflicts with #10013 , just FYI

ZehMatt:
We can merge yours first so its less work for you to rebase

AaronVanGeffen:
I think it would be better not to wait for #10013. That PR looks like it will need some checking w.r.t. whether or not it will inadvertently *hide* bugs.

tupaschoal:
I don't mind rebasing, shouldn't be too painful as it only affects one of the commits. Just wanted to note here to track

PR #10029, opened 2019-09-24
Fix output result of remove_unused_objects

IntelOrca:
`remove_unused_objects` returned a nonsense number. Only process objects that are actually selected in (hidden) object manager.

Also fix a nullptr exception if a path addition doesn't exist.

### DIFF
--- a/src/openrct2/EditorObjectSelectionSession.cpp
+++ b/src/openrct2/EditorObjectSelectionSession.cpp
@@ -530,20 +530,23 @@ int32_t editor_remove_unused_objects()
     int32_t numUnselectedObjects = 0;
     for (int32_t i = 0; i < numObjects; i++)
     {
-        if (!(_objectSelectionFlags[i] & OBJECT_SELECTION_FLAG_IN_USE)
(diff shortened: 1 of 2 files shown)


src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -530,20 +530,23 @@ int32_t editor_remove_unused_objects()
...
-            const ObjectRepositoryItem* item = &items[i];
-            uint8_t objectType = object_entry_get_type(&item->ObjectEntry);
-
-            if (objectType >= OBJECT_TYPE_SCENERY_GROUP)
+            if (!(_objectSelectionFlags[i] & OBJECT_SELECTION_FLAG_IN_USE)
+                && !(_objectSelectionFlags[i] & OBJECT_SELECTION_FLAG_ALWAYS_REQUIRED))
             {
-                continue;
-            }
+                const ObjectRepositoryItem* item = &items[i];
```
ZehMatt: Could use reference over pointer here

Gymnasiast: Could you two get to something we can merge? I really want to cut down on PR turnover times.

PR #10032, opened 2019-09-27
Add to build instructions

Krutonium:
Suggests building in RelWithDebInfo in the Readme, and adds instructions for building with Ninja in place of make.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -164,10 +164,20 @@ The standard CMake build procedure is to install the [required libraries](https:
 ```
 mkdir build
 cd build
-cmake ../ # set your standard cmake options, e.g. build type here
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
These would make good additions to the wiki, but I think we should not add this to the readme. It's long enough as-is without explaining how CMake works.

Krutonium:
Hm, how about if I remove the Ninja bit, but leave the RelWithDebInfo? My experience is that I end up having to hunt for it when I can't remember it, and having it there would be very helpful

Gymnasiast:
That would probably be a good compromise. There is a link to the Wiki just below, so people who want to know more can easily find that information.

PR #10037, opened 2019-09-28
Fix #10028: Saved park--brakes all set to 0mph

### LINKED ISSUE #10028: Saved park-brakes all set to 0mph
**Windows 10**
**0.2.3**
**b82d1bb:**

 I upgraded from 6b21aac on the x64 version and my park's coasters' brakes are all set to 0mph, causing one coaster to rollback.

Reproducible in RCT2 (vanilla)?
- No, brakes aren't set to 0mph when I load it in vanilla, but the game crashes after less than a minute in vanilla when I load the scenario.] 

Specific to multiplayer? No

**Steps to reproduce:**
1. Load the scenario attached 
2. See the hyper coaster stall after hitting a trim brake.

**Save game:**

[Six Flags BYO 1 - Copy.txt](https://github.com/OpenRCT2/OpenRCT2/files/3644651/Six.Flags.BYO.1.-.Copy.txt)

Gymnasiast:
Issue was caused by BrakeBoosterSpeed and OnridePhotoBits sharing a field, and RCT12TrackElement::GetPhotoTimeout() now returning 0 if the element is not a photo.

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -2128,8 +2128,14 @@ private:
                 dst2->SetDoorBState(src2->GetDoorBState());
                 dst2->SetStationIndex(src2->GetStationIndex());
                 dst2->SetHasGreenLight(src2->HasGreenLight());
-                dst2->SetBrakeBoosterSpeed(src2->GetBrakeBoosterSpeed());
-                dst2->SetPhotoTimeout(src2->GetPhotoTimeout());
+                if (track_element_has_speed_setting(dst2->GetTrackType()))
+                {
+                    dst2->SetBrakeBoosterSpeed(src2->GetBrakeBoosterSpeed());
+                }
+                else
+                {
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Went does the current design not work?

Gymnasiast:
Keep in mind that BrakeBoosterSpeed and OnridePhotoBits share a field in the **new** track element struct.

If the track piece is a brake a booster, the following line will properly set their speed:
```dst2->SetBrakeBoosterSpeed(src2->GetBrakeBoosterSpeed());```

However, the speed is then immediately set to 0 by this line:
```dst2->SetPhotoTimeout(src2->GetPhotoTimeout());```

This is because you changed `RCT12TrackElement::GetPhotoTimeout()` to always return 0 if the track piece in question is not a photo piece.

Gymnasiast:
Since this is quite a game-breaking bug, I have merged this now, but we can still discuss what's the best solution now the fire has been put out.

duncanspumpkin:
Argh i didn't realise that the new format still shared memory for the photos.

PR #10057, opened 2019-10-07
Fix #10052: Create a TrackDesignEntranceElement struct to replace rct_td6_entrance_element

### LINKED ISSUE #10052: Create a TrackDesignEntranceElement struct to replace rct_td6_entrance_element
We recently created a new TD6 and TD4 importer that takes the existing CS structs (`rct_track_td6`, `rct_track_td4`) and imports them into the new internal `TrackDesign` struct. This is the start of creating a new track design format for OpenRCT2. As part of this process we need to create new structs for the internal structs of the TrackDesign struct. The new `TrackDesign` has four structs that are still in the original CS types. These are:

- rct_td46_maze_element
- rct_td46_track_element
- rct_td6_entrance_element
- rct_td6_scenery_element

We need to convert the above structs into our own format so that we can improve them for greater flexibility. The first stage of doing this is to just create an identical struct with a new name and use it when importing and for the TrackDesign struct. I suggest we call them the following:

- TrackDesignMazeElement
- TrackDesignTrackElement
- TrackDesignEntranceElement
- TrackDesignSceneryElement

This refactor has been split into 4 parts. See also #10051, #10053, #10054.
This issue is only for the changes to implement `TrackDesignEntranceElement`.
Make the new struct in `ride/TrackDesign.h`. Make sure it has the exact same fields as `rct_td6_entrance_element`. Look at every use of `rct_td6_entrance_element` and if it is NOT in the following files change it to `TrackDesignEntranceElement`: T6Exporter.cpp, T6Importer.cpp, T4Importer.cpp. 

For T6Exporter.cpp, T6Importer.cpp, T4Importer.cpp you will need to modify the importers/exporter to convert between the two structs. Do this field by field. 

Example T6Importer
```
            rct_td6_entrance_element t6EntranceElement{};
            _stream.Read(&t6EntranceElement, sizeof(rct_td6_entrance_element));
            TrackDesignEntranceElement entranceElement{};
            entranceElement.x = t6EntranceElement.x;
            entranceElement.y = t6EntranceElement.y;
            entranceElement.z = t6EntranceElement.z;
            entranceElement.direction = t6EntranceElement.direction;
```

kodmord:


src/openrct2/ride/TrackDesign.h
```diff
@@ -21,6 +21,16 @@ struct Ride;
...
 
+/* Track Entrance entry size: 0x06 */
+struct TrackDesignEntranceElement
+{
+    int8_t z;          // 0x00
+    uint8_t direction; // 0x01
+    int16_t x;         // 0x02
+    int16_t y;         // 0x04
+};
+assert_struct_size(TrackDesignEntranceElement, 0x06);
```
duncanspumpkin: As this is no longer read from the file directly the struct size isn't important. So the assert can be deleted. This will allow us in the future to add or adjust the fields.

duncanspumpkin:
Oh one last thing. If this is your first contribution can you add yourself to the contributors.md.

PR #10058, opened 2019-10-08
Create TrackDesignTrackElement struct

### LINKED ISSUE #10054: Create a TrackDesignTrackElement struct to replace rct_td46_track_element
We recently created a new TD6 and TD4 importer that takes the existing CS structs (`rct_track_td6`, `rct_track_td4`) and imports them into the new internal `TrackDesign` struct. This is the start of creating a new track design format for OpenRCT2. As part of this process we need to create new structs for the internal structs of the TrackDesign struct. The new `TrackDesign` has four structs that are still in the original CS types. These are:

- rct_td46_maze_element
- rct_td46_track_element
- rct_td6_entrance_element
- rct_td6_scenery_element

We need to convert the above structs into our own format so that we can improve them for greater flexibility. The first stage of doing this is to just create an identical struct with a new name and use it when importing and for the TrackDesign struct. I suggest we call them the following:

- TrackDesignMazeElement
- TrackDesignTrackElement
- TrackDesignEntranceElement
- TrackDesignSceneryElement

This refactor has been split into 4 parts. See also #10052, #10051, #10054.
This issue is only for the changes to implement `TrackDesignTrackElement`.
Make the new struct in `ride/TrackDesign.h`. Make sure it has the exact same fields as `rct_td46_track_element`. Look at every use of `rct_td46_track_element` and if it is NOT in the following files change it to `TrackDesignTrackElement`: T6Exporter.cpp, T6Importer.cpp, T4Importer.cpp. 

For T6Exporter.cpp, T6Importer.cpp, T4Importer.cpp you will need to modify the importers/exporter to convert between the two structs. Do this field by field. 

Example T6Importer
```
                _stream.Read(&t6TrackElement, sizeof(rct_td46_track_element));
                TrackDesignTrackElement trackElement{};
                trackElement.type = t6TrackElement.type;
                trackElement.flags = t6TrackElement.flags;
                td->track_elements.push_back(trackElement);
```

tupaschoal:
Closes #10054

I failed to comply with the full instructions and created conversion constructors instead of doing the field by field initialization within the importers. It's easy enough to do that, so let me know if you want me to throw away the ctors and do that instead.

### DIFF
--- a/src/openrct2/rct1/T4Importer.cpp
+++ b/src/openrct2/rct1/T4Importer.cpp
@@ -266,11 +266,14 @@ private:
         }
         else
         {
-            rct_td46_track_element trackElement{};
+            rct_td46_track_element t4TrackElement{};
             for (uint8_t endFlag = _stream.ReadValue<uint8_t>(); endFlag != 0xFF; endFlag = _stream.ReadValue<uint8_t>())
             {
                 _stream.SetPosition(_stream.GetPosition() - 1);
-                _stream.Read(&trackElement, sizeof(rct_td46_track_element));
+                _stream.Read(&t4TrackElement, sizeof(rct_td46_track_element));
+                TrackDesignTrackElement trackElement{};
+                trackElement.type = t4TrackElement.type;
+                trackElement.flags = trackElement.flags;
                 td->track_elements.push_back(trackElement);
             }
         }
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
I think for this one it would have been fine either way. In some ways consistency is more important though so appreciate the change to the explicit conversion.

PR #10062, opened 2019-10-08
Avoid dereferencing map_get_surface_element_at nullptr on libopenrct2

tupaschoal:
Helps #9014 

There are multiple places on `MapHelpers.cpp` where the call to `map_get_surface_element_at` is done directly on the function parameters, didn't know what to do with those.

### DIFF
--- a/src/openrct2-ui/windows/Map.cpp
+++ b/src/openrct2-ui/windows/Map.cpp
@@ -1197,6 +1197,12 @@ static CoordsXYZD place_park_entrance_get_map_position(ScreenCoordsXY screenCoor
         return parkEntranceMapPosition;
 
     auto surfaceElement = map_get_surface_element_at(mapCoords);
+    if (surfaceElement == nullptr)
+    {
+        parkEntranceMapPosition.x = LOCATION_NULL;
+        return parkEntranceMapPosition;
+    }
+
     parkEntranceMapPosition.z = surfaceElement->GetWaterHeight() * 8;
     if (parkEntranceMapPosition.z == 0)
     {
@@ -1541,6 +1547,8 @@ static constexpr const uint8_t RideColourKey[] = {
 static uint16_t map_window_get_pixel_colour_peep(CoordsXY c)
 {
     auto* surfaceElement = map_get_surface_element_at(c);
+    if (surfaceElement == nullptr)
+        return 0;
     uint16_t colour = TerrainColour[surfaceElement->GetSurfaceStyle()];
     if (surfaceElement->GetWaterHeight() > 0)
         colour = WaterColour;
@@ -1580,6 +1588,9 @@ static uint16_t map_window_get_pixel_colour_ride(CoordsXY c)
     TileElement* tileElement = reinterpret_cast<TileElement*>(map_get_surface_element_at(c));
     do
     {
+        if (tileElement == nullptr)
+            break;
+
         if (tileElement->IsGhost())
         {
             colourA = MAP_COLOUR(PALETTE_INDEX_21);
--- a/src/openrct2/actions/LandSetHeightAction.hpp
+++ b/src/openrct2/actions/LandSetHeightAction.hpp
@@ -102,6 +102,9 @@ public:
         }
 
         auto* surfaceElement = map_get_surface_element_at(_coords);
+        if (surfaceElement == nullptr)
+            return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);
+
         TileElement* tileElement = CheckFloatingStructures(reinterpret_cast<TileElement*>(surfaceElement), _height);
         if (tileElement != nullptr)
         {
@@ -155,6 +158,9 @@ public:
         }
 
         auto* surfaceElement = map_get_surface_element_at(_coords);
+        if (surfaceElement == nullptr)
+            return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);
+
         cost += GetSurfaceHeightChangeCost(surfaceElement);
         SetSurfaceHeight(reinterpret_cast<TileElement*>(surfaceElement));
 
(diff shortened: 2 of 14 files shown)


src/openrct2/actions/LandSmoothAction.hpp
```diff
@@ -368,55 +368,83 @@ DEFINE_GAME_ACTION(LandSmoothAction, GAME_COMMAND_EDIT_LAND_SMOOTH, GameActionRe
                 // Smooth the 4 corners
                 { // top-left
                     auto surfaceElement = map_get_surface_element_at({ validRange.GetLeft(), validRange.GetTop() });
-                    int32_t z = std::clamp((uint8_t)tile_element_get_corner_height(surfaceElement, 2), minHeight, maxHeight);
-                    res->Cost += SmoothLandRowByCorner(
-                        isExecuting, { validRange.GetLeft(), validRange.GetTop() }, z, -32, -32, 0, 2);
+                    if (surfaceElement != nullptr)
```
ZehMatt: This will likely introduce a different bug if the surface element can not be found, this can be triggered with the Tile Inspector by deleting one of the surface elements and then using the Land tool afterwards. Could you perhaps check this with a debug build?

src/openrct2/actions/LandSetHeightAction.hpp
```diff
@@ -155,8 +158,11 @@ DEFINE_GAME_ACTION(LandSetHeightAction, GAME_COMMAND_SET_LAND_HEIGHT, GameAction
         }
 
         auto* surfaceElement = map_get_surface_element_at(_coords);
-        cost += GetSurfaceHeightChangeCost(surfaceElement);
-        SetSurfaceHeight(reinterpret_cast<TileElement*>(surfaceElement));
+        if (surfaceElement != nullptr)
```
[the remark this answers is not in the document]
tupaschoal: In this case it's avoiding exactly the crash, isn't it?

src/openrct2/actions/LandSmoothAction.hpp
tupaschoal: I don't really know how  to do this :(

tupaschoal: image
So I deleted the middle element from the surface there and it starts flickering on display (showing a tall terrain at times or showing none at all). Trying to elevate the terrain works, but trying to lower it sometimes yield the error shown above.

ZehMatt: This is to be somewhat expected if you erase the surface element, the rendering artifact is another issue on its own. I recall that std::clamp did not like the inverted min/max values, are you using a debug build?

tupaschoal: Yup, to make sure I just reran after this:
```
set platform=x64
msbuild openrct2.proj /t:clean
msbuild openrct2.proj /t:build /p:configuration=debug
```
And the results were the same

ZehMatt: Have you tried smooth land?

tupaschoal: Actually seemed to work as well
image

tupaschoal:
I've double checked all the places that call the function and this patch seems to handle all of them. Do kindly review, especially the ones with more intricate behavior, such as the TileInspetor, or the MapHelper, where I had to introduce an auxiliary function to avoid making the code unreadable.

tupaschoal:
Can I get some review in here? Really wanted to close the related issue and move on.

src/openrct2/actions/LandSetHeightAction.hpp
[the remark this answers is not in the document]
duncanspumpkin: It would be better to just return an error as done in the query. (ultimately this can never actually happen as it would be caught by the query)

src/openrct2/actions/LandSmoothAction.hpp
duncanspumpkin: I think this is fine.

src/openrct2/actions/TrackPlaceAction.hpp
```diff
@@ -351,6 +353,8 @@ DEFINE_GAME_ACTION(TrackPlaceAction, GAME_COMMAND_PLACE_TRACK, TrackPlaceActionR
 
             // 6c5648 12 push
             auto surfaceElement = map_get_surface_element_at(mapLoc);
+            if (surfaceElement == nullptr)
+                break;
```
duncanspumpkin: `return std::make_unique<TrackPlaceActionResult>(GA_ERROR::UNKNOWN, STR_NONE);`

src/openrct2/world/MapHelpers.cpp
```diff
@@ -240,11 +242,14 @@ int32_t tile_smooth(int32_t x, int32_t y)
 
             // Get neighbour height. If the element is not valid (outside of map) assume the same height
             auto* neighbourSurfaceElement = map_get_surface_element_at(x + x_offset, y + y_offset);
-            neighbourHeightOffset.baseheight[index] = neighbourSurfaceElement ? neighbourSurfaceElement->base_height
-                                                                              : surfaceElement->base_height;
+            if (neighbourSurfaceElement != nullptr)
```
duncanspumpkin: Not required the ternary already checks for if the element is nullptr. Perhaps make the ternary more explicit by adding an explicit comparison with nullptr

tupaschoal:
There as merge conflict, so I rebased it. Hope it is good for merging now, so we can finally close the related issue :D

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1197,6 +1197,8 @@ static CoordsXYZD place_park_entrance_get_map_position(ScreenCoordsXY screenCoor
         return parkEntranceMapPosition;
 
     auto surfaceElement = map_get_surface_element_at(mapCoords);
+    if (surfaceElement == nullptr)
+        return;
```
janisozaur:
```
../src/openrct2-ui/windows/Map.cpp: In function 'CoordsXYZD place_park_entrance_get_map_position(ScreenCoordsXY)':
../src/openrct2-ui/windows/Map.cpp:1201:9: error: return-statement with no value, in function returning 'CoordsXYZD' [-fpermissive]
         return;
         ^~~~~~
```

tupaschoal: Bad rebase, I'm sorry. Fixed it now.

PR #10066, opened 2019-10-08
Fix #10064: Refactor TrackDesignEntranceElement

### LINKED ISSUE #10064: Refactor TrackDesignEntranceElement
#10052 added the TrackDesignEntranceElement. We can now start modifying the structure of the entrance element to make things easier to follow. The first item that needs to be addressed is how the TrackDesignEntranceElement indicates whether it represents an exit or an entrance. At present this is done by stuffing the value in the highest bit in the direction. We want to move this stuffing into the import and export of TD6's. 

- Add a new field to the `TrackDesignEntranceElement` struct for `isExit`.
- Modify T6Importer modify the import so that isExit is set from the highest bit of the direction.
- Modify T6Exporter (do the opposite of T6Importer stuff the value into direction).
- Modify `track_design_place_ride` to use the new field.
- Modify `CreateTrackDesignTrack` to use the new field.
- Modify `window_track_place_draw_mini_preview_track` to use the new field.

I've added the bug label as whilst I was writing this up I've found a bug that the above refactor will likely solve. `track_design_mirror_ride` is silently destroying the isExit data in the direction field. This causes the entrance element to be invalid and this will cause them not to be placed. This is precisely why refactors like this are so useful.

kodmord:


src/openrct2/rct2/T6Importer.cpp
```diff
@@ -172,6 +172,7 @@ class TD6Importer final : public ITrackImporter
                 entranceElement.direction = t6EntranceElement.direction;
```
duncanspumpkin: Clear out the extra bit as well

src/openrct2/ride/TrackDesign.h
```diff
@@ -28,6 +28,7 @@ struct TrackDesignEntranceElement
     uint8_t direction;
     int16_t x;
     int16_t y;
+    uint8_t isExit;
```
duncanspumpkin: May as well make this a bool

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -330,7 +330,7 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
             // If this is the exit version
             if (i == 1)
             {
-                entrance.direction |= (1 << 7);
+                entrance.isExit = 1;
```
duncanspumpkin: After changing the field to a bool this can be made true

src/openrct2/rct2/T6Exporter.cpp
```diff
@@ -108,7 +108,7 @@ bool T6Exporter::SaveTrack(IStream* stream)
         for (const auto& entranceElement : _trackDesign->entrance_elements)
         {
             tempStream.WriteValue<uint8_t>(entranceElement.z);
-            tempStream.WriteValue<uint8_t>(entranceElement.direction);
+            tempStream.WriteValue<uint8_t>(entranceElement.direction | (entranceElement.isExit << 7));
```
kodmord: I have doubts. I think this code does not work correctly. If the `highest bit entranceElement.direction == 1` and `entranceElement.isExit == 0`, then the highest bit will be incorrect. We always want to fill it with the value `entranceElement.isExit`?

duncanspumpkin: Direction is always a value 0-3. The higher bits are never set.

PR #10067, opened 2019-10-09
Remove duplicate functions

tupaschoal:
Just some small changes that extract equal code to functions or calls another one instead of doing exactly the same thing.

### DIFF
--- a/src/openrct2/network/Twitch.cpp
+++ b/src/openrct2/network/Twitch.cpp
@@ -273,10 +273,7 @@ namespace Twitch
         // });
     }
 
-    /**
-     * GET /channel/:channel/audience
-     */
-    static void GetFollowers()
+    static void Get(const std::string& requestFormat, int operation)
     {
         char url[256];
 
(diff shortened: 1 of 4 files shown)


src/openrct2/network/Twitch.cpp
```diff
@@ -286,12 +283,12 @@ namespace Twitch
...
             return;
         }
-        snprintf(url, sizeof(url), "%s/channel/%s/audience", gConfigTwitch.api_url, gConfigTwitch.channel);
+        snprintf(url, sizeof(url), requestFormat.c_str(), gConfigTwitch.api_url, gConfigTwitch.channel);
 
         _twitchState = TWITCH_STATE_WAITING;
         _twitchIdle = false;
 
-        Http::DoAsync({ url }, [](Http::Response res) {
+        Http::DoAsync({ url }, [&](Http::Response res) {
```
janisozaur: This is for actually modifying `_twitchState`, `_twitchJsonResponse` I presume?

tupaschoal: Kind of, is for the lambda to have access to `operation` to use it on the assignment

janisozaur: It would have had access to `operation` with value capture anyway.

tupaschoal: Indeed, didn't give too much thought, will change it

tupaschoal:
Is this ok for merging now @janisozaur ?

janisozaur:
I think so, as long as Travis is green

tupaschoal:
Ok!

src/openrct2/world/Map.h
```diff
@@ -176,6 +176,7 @@ TileElement* tile_element_insert(const TileCoordsXYZ& loc, int32_t occupiedQuadr
 
 using CLEAR_FUNC = int32_t (*)(TileElement** tile_element, int32_t x, int32_t y, uint8_t flags, money32* price);
 
+int32_t map_place_clear_func(TileElement** tile_element, int32_t x, int32_t y, uint8_t flags, money32* price, bool is_scenery);
```
duncanspumpkin: Why was this added to the header? It will never be used.

tupaschoal: You're correct, it was not needed, I removed it.

tupaschoal:
I guess it does need the function on the header (which I suspected it would - see all the failures on Travis). So I added it back again

duncanspumpkin:
No that isn't right. Its a static function. You need to declare it in the file.  Just add `static`.

PR #10072, opened 2019-10-10
Close #10071: Update Twitch icon

### LINKED ISSUE #10071: Update Twitch Logo
As of September 26, Twitch has refreshed it's visual identity and changed the logo [(source)](https://blog.twitch.tv/en/2019/09/26/nice-to-meet-you-again-for-the-first-time/).

The twitch icon on the settings should be updated to the new one:
twitch

tupaschoal:
This is how it looks with the new twitch icon
new_twitch_icon
This is how it looked with the old:
old_twitch_icon

Note that as I mentioned in the issue and on Gitter I had some trouble testing these, I could only make it work after I deleted `g2.dat` and `config.ini`. Not sure if this is exposing some dependency problem.


duncanspumpkin:
The icon is too big it's overlapping the top of the button.

AaronVanGeffen:
To be honest, I would rather keep the old, colourful one, either way. It matches the style of the existing tabs much more closely.

tupaschoal:
> The icon is too big it's overlapping the top of the button.

Indeed. It actually has the exact same dimensions as the former icon (31x27 if I'm not mistaken), but the former had a lot more margin. I inspected the tree icon (2nd tab) and it is as large as it can get, with a 1px margin on every direction, so I could make something similar.
How important is it to keep the same aspect ratio/dimensions as the previous one?

> To be honest, I would rather keep the old, colourful one, either way. It matches the style of the existing tabs much more closely.

We could go for the same icon, but with their new purple color, something like [this](https://images.app.goo.gl/36hux5KKNBHoEtw67)

AaronVanGeffen:
>  How important is it to keep the same aspect ratio/dimensions as the previous one?

Not at all. Just offset it in the json file.

tupaschoal:
I did try with these two icons:
twitch2 twitch3
Sadly, I can't get the second to work, maybe the game can't represent all colors?

With the first, it's like this:
icon_twitch_black

AaronVanGeffen:
You'd have to use one of the palette colours, yes. An option would be to import it using either 'dithering' or 'closest colour' mode, then exporting the resulting sprite.

The icon is still too big, imo. I also much prefer the gradient from the current icon. Again, I think the solid fill colour does not fit with the rest of the current icons. Is this really worth all the trouble?

tupaschoal:
> Is this really worth all the trouble?

Dunno, if you guys think it's not worth it, we can drop it, no hard feelings

Gymnasiast:
In that case, let's keep the old one, then.

IntelOrca:
Given Twitch integration doesn't really work anymore (at least now without setting up your own chatbot servers) we will probably remove it soon anyway. It is something that should be in a plugin.

PR #10075, opened 2019-10-10
Little Refactor

duncanspumpkin:
Was looking at the latest track design changes and realised that Track Elem 255 can be labelled.
Also corrected a few cases where the wrong name was being used.


src/openrct2/ride/Vehicle.cpp
```diff
@@ -5684,7 +5684,7 @@ GForces vehicle_get_g_forces(const rct_vehicle* vehicle)
         case TRACK_ELEM_60_DEG_DOWN_COVERED:
         case TRACK_ELEM_BRAKES:
         case TRACK_ELEM_ROTATION_CONTROL_TOGGLE:
-        case TRACK_ELEM_INVERTED_90_DEG_UP_TO_FLAT_QUARTER_LOOP:
+        case TRACK_ELEM_MAZE:
```
duncanspumpkin: Line 6119 needs corrected as well

PR #10076, opened 2019-10-10
Use white colour for labels in Tile Inspector

Broxzier:
The dark-green colour appears white and comes from the days of hardcoded numbers, where it was most likely copied from somewhere else.


duncanspumpkin:
Can you do a before and after picture

Broxzier:
There shouldn't be any visual differences, but I'll make some screenshots this evening to verify that.

PR #10077, opened 2019-10-10
Fix #9985: Mistake from #9128

### LINKED ISSUE #9985: Fix mistake from #9128
Need to apply the change from #9128 to `TILE_INSPECTOR_PAGE_ENTRANCE` for its use of the entrance id.

Broxzier:
This was likely a test for working with STRINGID before creating a new string entry, and slipped in the previous PR. Station numbers are now displayed correctly for ride entrances and exits.


Broxzier:
While this fixes it for entrances, it breaks it again for tracks. Proper fix incoming.

PR #10082, opened 2019-10-11
[WIP] Use mingw on fedora.

### LINKED ISSUE #7079: Consider replacement for Arch-based docker images with mingw toolchain
mingw-w64 packages got removed from community repository of Arch, as per discussion in https://bbs.archlinux.org/viewtopic.php?id=233373.

While the already-built images will keep mingw builds alive, they may soon require updating wrt https://github.com/OpenRCT2/OpenRCT2/pull/6034

This has already started affecting us by inability to build image containing Boost for OpenLoco.

Consider moving docker images to Fedora, which has decent mingw packages available, e.g. https://src.fedoraproject.org/rpms/mingw-gcc.

notxarb:
I want to resolve #7079
This includes a Dockerfile based off of Feodra 30, and some fixes so it compiles.


CMakeLists.txt
```diff
@@ -200,6 +200,7 @@ else ()
     ADD_CHECK_CXX_COMPILER_FLAG(CMAKE_CXX_FLAGS CXX_WARN_RANGE_LOOP_ANALYSIS -Wrange-loop-analysis)
     ADD_CHECK_CXX_COMPILER_FLAG(CMAKE_CXX_FLAGS CXX_WARN_TAUTOLOGICAL_ZERO_COMPARE -Wtautological-unsigned-zero-compare)
     ADD_CHECK_CXX_FLAG(CMAKE_CXX_FLAGS CXX_WNO_CLOBBERED -Wno-clobbered)
+    ADD_CHECK_CXX_FLAG(CMAKE_CXX_FLAGS CXX_WNNO_CAST_FUNCTION_TYPE -Wno-cast-function-type)
```
janisozaur: There's a typo: `WNNO`, double `N`.

Shouldn't this also only be enabled for mingw builds?

dockerfiles/mingw-fedora/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+        fakeroot \
+        file \
+        patch \
+        ninja-build
+
+# Install MingW packages
+RUN dnf install -y \
+        mingw32-binutils \
+        mingw32-boost \
+#        mingw32-cmake \
```
janisozaur: This looks off. I doubt you can insert comments in the middle of a list using escaped new lines.

dockerfiles/mingw-fedora/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+        mingw32-libpng \
+        mingw32-libzip \
+        mingw32-openssl \
+        mingw32-pkg-config \
+        mingw32-SDL2 \
+        mingw32-SDL2_ttf \
+        mingw32-speexdsp \
+        mingw-w64-tools \
+        mingw32-winpthreads \
+        mingw32-win-iconv \
```
janisozaur: I think we don't use iconv on Windows, what is it required for?

dockerfiles/mingw-fedora/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+        mingw32-zlib
+
+RUN useradd --create-home --comment "Fedora Build User" build && \
+    groupadd sudo && \
+    echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers; \
```
janisozaur: I don't see a reason to use `sudo`, please explain

notxarb: I did a basic copy from the mingw-arch, it was in there, so I kept it. But you are probably right as I think it was used to compile extra packages that aren't in the arch package repository.

dockerfiles/mingw-fedora/Dockerfile
notxarb: I don't know if iconv is being used or not. It was in the mingw-arch list of packages, so I got the fedora version of the package.

dockerfiles/mingw-fedora/Dockerfile
notxarb: You can comment in the middle of escaped new lines in Dockerfiles. I was trying to copy the list from the mingw-arch Dockerfile. It's commented out because there is no mingw cmake package on fedora. I had it commented out to get the Dockerfile to build, but it can be removed now as it probably doesn't need to be there.

CMakeLists.txt
notxarb: Indeed there is an extra N in there, and yes it probably does only need to be enabled for mingw.

Gymnasiast:
@notxarb Could you rebase this pull request?

tupaschoal:
This PR has become stale as  the docker images have been moved out of this repository. 
If you can resume ot based on Janis comments and reopen it, feel free to reopen it.

PR #10090, opened 2019-10-13
Use ScreenCoordsXY for Keyboard shortcuts

tupaschoal:


src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -232,9 +232,9 @@ void keyboard_shortcuts_format_string(char* buffer, size_t bufferSize, int32_t s
     String::Set(buffer, bufferSize, str.c_str());
 }
 
-void get_keyboard_map_scroll(const uint8_t* keysState, int32_t* x, int32_t* y)
+void get_keyboard_map_scroll(const uint8_t* keysState, ScreenCoordsXY& screenCoords)
```
ZehMatt:
```suggestion
ScreenCoordsXY get_keyboard_map_scroll(const uint8_t* keysState)
```

src/openrct2-ui/input/KeyboardShortcuts.h
```diff
@@ -131,7 +132,7 @@ namespace OpenRCT2
 
             void Set(int32_t key);
             int32_t GetFromKey(int32_t key);
-            void GetKeyboardMapScroll(const uint8_t* keysState, int32_t* x, int32_t* y) const;
+            void GetKeyboardMapScroll(const uint8_t* keysState, ScreenCoordsXY& screenCoords) const;
```
ZehMatt:
```suggestion
            ScreenCoordsXY GetKeyboardMapScroll(const uint8_t* keysState) const;
```

tupaschoal:
Thanks for the suggestions @ZehMatt , sometimes people prefer return values, sometimes not, thus I didn't change at first.

PR #10091, opened 2019-10-13
Avoid dereferencing invalid map_get_first_element_at return

tupaschoal:
Helps #9014 and continue #10013 

I found out there were some places that weren't checking still, so decided to do a full run-through on all the usages.

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -528,6 +528,8 @@ static void viewport_interaction_remove_footpath(TileElement* tileElement, int32
         footpath_provisional_update();
 
     tileElement2 = map_get_first_element_at(x / 32, y / 32);
+    if (tileElement2 == nullptr)
+        return;
     do
     {
         if (tileElement2->GetType() == TILE_ELEMENT_TYPE_PATH && tileElement2->base_height == z)
--- a/src/openrct2-ui/windows/Banner.cpp
+++ b/src/openrct2-ui/windows/Banner.cpp
@@ -137,6 +137,8 @@ rct_window* window_banner_open(rct_windownumber number)
     int32_t view_y = banner->position.y << 5;
 
     TileElement* tile_element = map_get_first_element_at(view_x / 32, view_y / 32);
+    if (tile_element == nullptr)
+        return nullptr;
     while (1)
     {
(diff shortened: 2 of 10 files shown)


src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -530,6 +530,8 @@ static void viewport_interaction_remove_footpath(TileElement* tileElement, int32
     tileElement2 = map_get_first_element_at(x / 32, y / 32);
     do
     {
+        if (tileElement2 == nullptr)
```
Gymnasiast: You can just return before going into the `do` function, like you already do elsewhere in the PR.

tupaschoal: Done!

src/openrct2-ui/windows/Banner.cpp
```diff
@@ -176,6 +178,8 @@ static void window_banner_mouseup(rct_window* w, rct_widgetindex widgetIndex)
     int32_t y = banner->position.y << 5;
 
     TileElement* tile_element = map_get_first_element_at(x / 32, y / 32);
+    if (tile_element == nullptr)
+        return;
```
duncanspumpkin: This would technically block a window close that could be valid. Its not worth changing though. Keep like this.

src/openrct2/actions/ClearAction.hpp
```diff
@@ -146,6 +146,8 @@ DEFINE_GAME_ACTION(ClearAction, GAME_COMMAND_CLEAR_SCENERY, GameActionResult)
             tileElement = map_get_first_element_at(x, y);
             do
             {
+                if (tileElement == nullptr)
+                    break;
```
duncanspumpkin: This will enter an infinite loop best to `return totalCost;`

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -3004,6 +3004,8 @@ class S4Importer final : public IParkImporter
                     continue;
                 do
                 {
+                    if (tileElement == nullptr)
```
duncanspumpkin: Already accounted for above on line 3003

src/openrct2-ui/windows/Banner.cpp
tupaschoal: Ok, not changing anything on this one then.

tupaschoal:
Done!

PR #10092, opened 2019-10-13
Remapped keypad Enter to return scancode

### LINKED ISSUE #9684: Entering custom size for water/land tool allows confirmation with main enter key, but not numpad enter key
**OS:** Windows 8
**Version:** 0.2.3
**Commit/Build:** 8378790

When you're entering the desired size of the tool in the water or land tool with the keyboard, it lets you confirm entry with the main enter key, but _not_ with the numpad enter key.

I'm talking about when you click the number in one of these windows:
Forest Frontiers 2019-07-29 23-21-51 
Unnamed park 2019-07-29 22-58-51
and then a dialog box pops up to let you type in a number between 1 and 64.

You can then submit the number by clicking the OK button on that window, _or_ by pressing the keyboard's main enter key, but _not_ the numpad enter key. This is inconvenient when typing in the number on the numpad.

**Steps to reproduce:**
1. Select land or water tool.
2. Click the number in the center of it.
3. Type in a valid number.
4. Attempt to submit by pressing the numpad enter key.

**Dump file**

**Screenshots / Video:**

**Save game:**

### LINKED ISSUE #9690: The keyboard shortcut for rotating the game view can be set to Enter or KP Enter, but not both
**OS:** Windows 8.1
**Commit/Build:** 77f1905

----

Original RCT2 allowed rotating the view via either the normal Enter key or the keypad Enter key, interchangeably.

OpenRCT2, in its present state, does not allow this. If the keyboard shortcut is bound to "Return" (the default), then only the normal Enter key will rotate the view. If the keyboard shortcut is instead bound to "Keypad Enter", then only the keypad Enter key will rotate the view. There does not appear to be any way to allow both Enter keys to trigger the action.

----

I did look for duplicate reports on this, but didn't find anything recent. Way back in 2014, issue #651 was filed, concerning this same problem, and there was a fix for it in #653. But in the years since, my understanding is that basically all of the relevant code has been refactored and rewritten, making that moot.

I have no idea when this became a problem again, but I would assume that it was probably when the current keyboard shortcut system code was written up. I could have done a big bisect to pin it down for sure; but it doesn't seem terribly important, since the reason for the issue in the current code is quite straightforward.

In the current code, the keyboard shortcut system works via a series of simple 1:1 mappings between a **single** SDL scancode (possibly OR'd with a modifier key) and the function to be executed when that scancode is matched. And of course, since there are two distinct SDL scancodes for normal Enter and keypad Enter (`SDL_SCANCODE_RETURN` and `SDL_SCANCODE_KP_ENTER`), only one or the other can be assigned to the mapping for `SHORTCUT_ROTATE_VIEW_CLOCKWISE`; not both.

----

So... it seems that a resolution to this issue would probably either involve (a) a more complicated keyboard shortcut system that allows multiple keys to be bound to the same action (which would be plenty of work, and would make the UI a bit more complicated for users, but would ultimately allow for greater flexibility); or (b) some special-casing for the Enter keys (which might be a bit ugly, but would be simpler). The latter approach appears to be more-or-less what was done way back in #653.

Any special-casing, if done, I reckon would probably be put into `keyboard_shortcut_handle` and/or `keyboard_shortcut_handle_command`, essentially spoofing equality for `SDL_SCANCODE_RETURN` and `SDL_SCANCODE_KP_ENTER`. And then some additional code would probably be desirable in `KeyboardShortcuts::Set` and/or `KeyboardShortcuts::GetFromKey` to force user input of `SDL_SCANCODE_KP_ENTER` in the keyboard shortcut change window to be stored as `SDL_SCANCODE_RETURN`, lest users think they are mapping two separate actions to Enter and KP Enter when in fact only the non-keypad one will actually do anything.

----

I don't have a PR for this at the moment; if you want me to write one up, I can take a stab at it.

Really though, at this point, I'm more interested to know which direction you guys would prefer to take this (in part because it's potentially arguable whether the two Enter keys *should* be treated interchangeably for keyboard shortcuts, depending on preference):
- No fix
- Special-casing the Enter scancodes in the keyboard shortcut system, unilaterally
- Special-casing the Enter scancodes in the keyboard shortcut system, but only for the view rotation shortcuts in particular
- Enhancing the keyboard shortcut system to allow multiple keys to be mapped to the same action
- Something else

kodmord:
I added a scancode override for keypad Enter. This affected not only issues #9684 and #9690, but also the keyboard shortcut settings. The keypad Enter shortcut settings are now defined as Enter.

Fix #9684: Entering custom size for water/land tool allows confirmation
with main enter key, but not numpad enter key

Fix #9690: The keyboard shortcut for rotating the game view can be set
to Enter or KP Enter, but not both

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -17,6 +17,8 @@
 - Fix: [#9669] The tile inspector shortcut key does not work with debugging tools disabled.
 - Fix: [#9675] Guest entry point limit can be bypassed in scenario editor.
 - Fix: [#9683] Cannot raise water level if part of the tool's area of effect is off of the map.
+- Fix: [#9684] Entering custom size for water/land tool allows confirmation with main enter key, but not numpad enter key.
+- Fix: [#9690] The keyboard shortcut for rotating the game view can be set to Enter or KP Enter, but not both.
 - Fix: [#9717] Scroll bars do not render correctly when using OpenGL renderer.
 - Fix: [#9729] Peeps do not take into account height difference when deciding to pathfind to a ride entrance (original bug).
 - Fix: [#9902] Doors/Portcullis do not check to make sure doors are open causing double opens.
--- a/src/openrct2-ui/TextComposition.cpp
+++ b/src/openrct2-ui/TextComposition.cpp
@@ -106,13 +106,15 @@ void TextComposition::HandleMessage(const SDL_Event* e)
 
             uint16_t modifier = e->key.keysym.mod;
             SDL_Keycode key = e->key.keysym.sym;
+            SDL_Scancode scancode = e->key.keysym.scancode;
             if (key == SDLK_KP_ENTER)
             {
                 // Map Keypad enter to regular enter.
                 key = SDLK_RETURN;
+                scancode = SDL_SCANCODE_RETURN;
             }
 
-            GetContext()->GetUiContext()->SetKeysPressed(key, e->key.keysym.scancode);
+            GetContext()->GetUiContext()->SetKeysPressed(key, scancode);
 
             // Text input
             if (_session.Buffer == nullptr)


AaronVanGeffen:
Could you add a changelog entry for this fix?

PR #10096, opened 2019-10-14
Add define for ARM64

janisozaur:


janisozaur:
MacOS job fails on upload, but builds correctly

PR #10103, opened 2019-10-16
Fix some compilation warnings

tupaschoal:
I was exploring the solution a bit and have found plenty of warnings, they're probably not all applicable, but I'm creating this draft PR to see whether this is something you're interested in having fixed/dealt with.

- The loop tried to access `ScenarioOriginStringIds[7]`, which doesn't exist.
- `json_is_string` already checks for `palette` pointer validity
- `WallPlaceAction` had uninitialized member variables

### DIFF
--- a/src/openrct2-ui/drawing/engines/opengl/TextureCache.cpp
+++ b/src/openrct2-ui/drawing/engines/opengl/TextureCache.cpp
@@ -61,6 +61,7 @@ void TextureCache::InvalidateImage(uint32_t image)
     }
 }
 
+// Note: for performance reasons, this returns a BasicTextureInfo over an AtlasTextureInfo (also to not expose the cache)
 BasicTextureInfo TextureCache::GetOrLoadImageTexture(uint32_t image)
 {
     uint32_t index;
--- a/src/openrct2-ui/windows/TitleScenarioSelect.cpp
+++ b/src/openrct2-ui/windows/TitleScenarioSelect.cpp
@@ -442,7 +442,7 @@ static void window_scenarioselect_paint(rct_window* w, rct_drawpixelinfo* dpi)
                                                                                      : STR_WINDOW_COLOUR_2_STRINGID;
 
     // Text for each tab
-    for (int32_t i = 0; i < 8; i++)
+    for (uint32_t i = 0; i < std::size(ScenarioOriginStringIds); i++)
     {
         rct_widget* widget = &window_scenarioselect_widgets[WIDX_TAB1 + i];
         if (widget->type == WWT_EMPTY)
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -608,7 +608,7 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
             }
 
             json_t* path = json_object_get(sprite_description, "path");
-            if (!path || !json_is_string(path))
+            if (!json_is_string(path))
             {
                 fprintf(stderr, "Error: no path provided for sprite %lu\n", (unsigned long)i);
                 json_decref(sprite_list);
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -621,7 +621,7 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
             // Get palette option, if present
             bool keep_palette = false;
             json_t* palette = json_object_get(sprite_description, "palette");
-            if (palette && json_is_string(palette))
+            if (json_is_string(palette))
             {
                 const char* option = json_string_value(palette);
                 if (strncmp(option, "keep", 4) == 0)
--- a/src/openrct2/actions/WallPlaceAction.hpp
+++ b/src/openrct2/actions/WallPlaceAction.hpp
@@ -55,15 +55,13 @@ private:
     int32_t _wallType{ -1 };
     CoordsXYZ _loc;
     Direction _edge{ INVALID_DIRECTION };
-    int32_t _primaryColour;
-    int32_t _secondaryColour;
-    int32_t _tertiaryColour;
+    int32_t _primaryColour{ 0 };
+    int32_t _secondaryColour{ 0 };
+    int32_t _tertiaryColour{ 0 };
     BannerIndex _bannerId{ BANNER_INDEX_NULL };
 
 public:
-    WallPlaceAction()
-    {
-    }
+    WallPlaceAction() = default;
 
     WallPlaceAction(
         int32_t wallType, CoordsXYZ loc, uint8_t edge, int32_t primaryColour, int32_t secondaryColour, int32_t tertiaryColour)
--- a/src/openrct2/actions/WallPlaceAction.hpp
+++ b/src/openrct2/actions/WallPlaceAction.hpp
@@ -55,9 +55,9 @@ private:
     int32_t _wallType{ -1 };
     CoordsXYZ _loc;
     Direction _edge{ INVALID_DIRECTION };
-    int32_t _primaryColour{ 0 };
-    int32_t _secondaryColour{ 0 };
-    int32_t _tertiaryColour{ 0 };
+    int32_t _primaryColour{ COLOUR_BLACK };
+    int32_t _secondaryColour{ COLOUR_BLACK };
+    int32_t _tertiaryColour{ COLOUR_BLACK };
     BannerIndex _bannerId{ BANNER_INDEX_NULL };
 
 public:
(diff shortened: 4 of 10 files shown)


src/openrct2-ui/windows/TitleScenarioSelect.cpp
```diff
@@ -442,7 +442,7 @@ static void window_scenarioselect_paint(rct_window* w, rct_drawpixelinfo* dpi)
                                                                                      : STR_WINDOW_COLOUR_2_STRINGID;
 
     // Text for each tab
-    for (int32_t i = 0; i < 8; i++)
+    for (int32_t i = 0; i < SCENARIO_CATEGORY_COUNT; i++)
```
Gymnasiast: `std::size(ScenarioOriginStringIds)` is probably better, since it more explicitly makes sure not to loop beyond the size of `ScenarioOriginStringIds`.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -621,7 +621,7 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
             // Get palette option, if present
             bool keep_palette = false;
             json_t* palette = json_object_get(sprite_description, "palette");
-            if (palette && json_is_string(palette))
+            if (json_is_string(palette))
```
ZehMatt: There are a few more like this

janisozaur:
How about turning then on on our CIs?

janisozaur:
Are you building with MSVC? Can you provide exact warnings that you hit?

tupaschoal:
I am using the `msbuild` example on the readme page, but nonetheless, I'm getting these warnings from a visual studio option which is something like "look for errors/earnings", I don't recall the exact wording. It yields ~1k warnings

duncanspumpkin:
A whole bunch of the warnings are to do with not initialising game action variables.

src/openrct2/CmdlineSprite.cpp
tupaschoal: On the same file, I actually found no other `A && json_is_\(.*\)(A)` combination

src/openrct2-ui/windows/TitleScenarioSelect.cpp
tupaschoal: Done

tupaschoal:
I've fixed a bunch of them, but I do not intend to increment this PR anymore, so I marked it ready for review.

I want to work on enabling warnings that are actually caught during compilation, otherwise it just piles up a while later. Let me know if you have any tips to start with,

src/openrct2/actions/WallPlaceAction.hpp
```diff
@@ -55,15 +55,13 @@ DEFINE_GAME_ACTION(WallPlaceAction, GAME_COMMAND_PLACE_WALL, WallPlaceActionResu
     int32_t _wallType{ -1 };
     CoordsXYZ _loc;
     Direction _edge{ INVALID_DIRECTION };
-    int32_t _primaryColour;
-    int32_t _secondaryColour;
-    int32_t _tertiaryColour;
+    int32_t _primaryColour{ 0 };
```
Gymnasiast: COLOUR_BLACK

src/openrct2-ui/drawing/engines/opengl/TextureCache.cpp
```diff
@@ -61,7 +61,7 @@ void TextureCache::InvalidateImage(uint32_t image)
     }
 }
 
-BasicTextureInfo TextureCache::GetOrLoadImageTexture(uint32_t image)
+AtlasTextureInfo TextureCache::GetOrLoadImageTexture(uint32_t image)
```
Gymnasiast: Could you elaborate on this?

src/openrct2/network/Network.cpp
```diff
@@ -1356,6 +1356,11 @@ void Network::BeginServerLog()
     {
         format_string(logMessage, sizeof(logMessage), STR_LOG_SERVER_STARTED, nullptr);
     }
+    else
+    {
+        logMessage[0] = '\0';
+        assert(0 && "Uninitialized mode");
```
Gymnasiast:
```suggestion
        Guard::Assert(false, "Unknown network mode!")
```

src/openrct2/network/Network.cpp
```diff
@@ -1379,6 +1384,11 @@ void Network::CloseServerLog()
     {
         format_string(logMessage, sizeof(logMessage), STR_LOG_SERVER_STOPPED, nullptr);
     }
+    else
+    {
+        logMessage[0] = '\0';
+        assert(0 && "Uninitialized mode");
```
Gymnasiast:
```suggestion
        Guard::Assert(false, "Unknown network mode!")
```

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -168,8 +168,8 @@ void S6Exporter::Export()
         auto temp = utf8_to_rct2(gS6Info.details);
         safe_strcpy(_s6.info.details, temp.data(), sizeof(_s6.info.details));
     }
-    uint32_t researchedTrackPiecesA[128];
-    uint32_t researchedTrackPiecesB[128];
+    uint32_t researchedTrackPiecesA[128] = { 0 };
```
ZehMatt: Do you want zero initialize the whole array or just the first member here?

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -168,8 +168,8 @@ void S6Exporter::Export()
...
+    uint32_t researchedTrackPiecesB[128] = { 0 };
```
ZehMatt: Same as above

src/openrct2/CmdlineSprite.cpp
ZehMatt: Yeah don't use regex for such things instead read the actual file maybe. See line 610

src/openrct2-ui/drawing/engines/opengl/TextureCache.cpp
tupaschoal: The function actually creates an `AtlasTextureInfo`, not a `BasicTextureInfo`, so returning it by value actually slices it.

src/openrct2/rct2/S6Exporter.cpp
tupaschoal: The whole array, and it does so: https://godbolt.org/z/Q9o4RE

src/openrct2/CmdlineSprite.cpp
tupaschoal: To be fair, line 610 is an `||` clause, rather than an `&&` :P

src/openrct2/rct2/S6Exporter.cpp
ZehMatt: This is the C way of doing it, its also fine for C++ but you want to default initialize via {}

tupaschoal: Done

src/openrct2-ui/drawing/engines/opengl/TextureCache.cpp
ZehMatt: Please revert this, this is done on purpose for performance reasons.

Gymnasiast: We should add a comment, then.

tupaschoal: I can revert it, but I agree with Michael that we should leave a comment with a good explanation there, then.

Gymnasiast: Let's do that: revert the change and add a comment: `Note: for performance reasons, this functions returns a BasicTextureInfo instead of an AtlasTextureInfo.`

ZehMatt: Its also not strictly performance reasons, the data is the internal state of the cache and the callee shouldn't care about that.

tupaschoal: Done

src/openrct2-ui/drawing/engines/opengl/TextureCache.h
```diff
@@ -217,6 +217,7 @@ class TextureCache final
     TextureCache();
     ~TextureCache();
     void InvalidateImage(uint32_t image);
+    // Note: for performance reasons, this returns a BasicTextureInfo over an AtlasTextureInfo (also to not expose the cache)
```
ZehMatt: The comment is a little misplaced, this interface doesn't have to provide justification to why it results this structure. It makes more sense to move this comment into the implementation.

tupaschoal: I pondered about that for a little and took the wrong choice hahah. Fixed it now, thanks.

PR #10104, opened 2019-10-16
Fix possible cause for desyncs on ride crashes

Broxzier:
`scenario_rand` was used twice between two sequence points. The order of evaluation is unspecified in C++, meaning that these calls could be done in both forward and reverse order. Storing them in variables guarantees their order, making this cross-platform.


janisozaur:
Nice!

duncanspumpkin:
Hmm never thought of that.. I've probably done similar in a few places.

Broxzier:
@duncanspumpkin I've checked all lines with `scenario_rand` twice in it. This was the only one that had them inside one function call. It could be that I missed some, where the function arguments are split over multiple lines.

PR #10107, opened 2019-10-17
Fix #10106: Reset number of circuits if unsupported

### LINKED ISSUE #10106: Number of circuits is not reset to 1 when changing to mode that doesn't support multiple circuits
**OS:** Windows 10
**Version:** 0.2.3
**Commit/Build:** 8c1ccfd

I think the Number of circuits setting is not resetted to once after changing operation mode from "Powered launched (passing station)" into "~ (without passing station)".

Additional: The coaster tries to launch the car as many as the number of circuit is setted, but it slightly overruns behind the station just before it re-launched.

**Steps to reproduce:**
1. Build a (looping) coaster like this:
image
2. Set its operation mode as "Powered launch (passing station)" first (no test/open yet)
3. Set ``Number of circuits`` as 2.
4. Set the operation mode as "Powered launch (without passing station)" again.
5. Open or Test it, then the coaster crashes.
 If there are more tracks behind the station, it will not crash since there is more rooms, but coaster is launched twice.

tupaschoal:
Closes #10106

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -30,6 +30,7 @@
 - Fix: [#9994] Game action tick collision during server connect and map load.
 - Fix: [#10017] Ghost elements influencing ride excitement.
 - Fix: [#10036] Do not allocate large chunks of memory for save file classification.
+- Fix: [#10106]: Ride circuits should not be used for modes that do not support it.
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
Just hope that ne doesn't use this for some unknown reason.

tupaschoal:
Done!

Issue #10065, opened 2019-10-08
Refactor `get_map_coordinates_from_pos`


duncanspumpkin:
`get_map_coordinates_from_pos` and `get_map_coordinates_from_pos_window` are used to convert a screen coordinate position that a mouse is over into a map position that the game uses and tell you what item you are hovering over. It is quite complex how it works everything out but this refactor is about modifying its calling parameters. To make things easier to understand in the project we are trying to identify the different coordinate systems. This function uses both the map coordinates `CoordsXY` and the screen coordinates `ScreenCoords`. 

Create an overload of `get_map_coordinates_from_pos_window` that takes the current 9 parameters. Make that function call the original. This will be our shim to the old version.

Modify `get_map_coordinates_from_pos_window` to take a `ScreenCoordsXY` instead of `screenX` and `screenY`. Modify it further to take a `&CoordsXY` instead of `*x` and `*y`. Make sure the shim that was created in the first step still works correctly converting between the two.

Repeat the above for `get_map_coordinates_from_pos` and make it so that it calls our new `get_map_coordinates_from_pos_window`.

Later on refactors will look at calling the new version for all functions that call these functions. If you come across some simple ones feel free to do them whilst doing this refactor. Try not to do anything too crazy as it will take longer to review.

ahmedkrmn:
Can I work on this one?

duncanspumpkin:
Sure go ahead.

ahmedkrmn:
What do you think about using [std::pair](https://en.cppreference.com/w/cpp/utility/pair) to represent both  `ScreenCoordsXY`  and `CoordsXY`?

duncanspumpkin:
I don't think it would be useful for this. What did you have in mind.

ahmedkrmn:
@duncanspumpkin I think we can use either a pair, array, vector or a struct.

ahmedkrmn:
@duncanspumpkin If a [std::pair](https://en.cppreference.com/w/cpp/utility/pair) is used, `get_map_coordinates_from_pos_window()` can be overloaded to be as follows:
```
void get_map_coordinates_from_pos_window(
    rct_window* window, std::pair<int32_t, int32_t> ScreenCoordsXY, int32_t flags, std::pair<int16_t*, int16_t*>& CoordsXY,
    int32_t* interactionType, TileElement** tileElement, rct_viewport** viewport);
```

duncanspumpkin:
Ah no the structs already exist.

[...]

PR #10109, opened 2019-10-18
Refactor get_map_coordinates_from_pos_window

bentglasstube:
This changes get_map_coordinates_from_pos_window to use the
ScreenCoordsXY and CoordsXY structs instead of bare ints as suggested
in #10065


src/openrct2/interface/Viewport.cpp
```diff
@@ -1665,25 +1665,40 @@ void get_map_coordinates_from_pos(
 void get_map_coordinates_from_pos_window(
     rct_window* window, int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y, int32_t* interactionType,
     TileElement** tileElement, rct_viewport** viewport)
+{
+  ScreenCoordsXY screenCoords(screenX, screenY);
+  CoordsXY mapCoords(0, 0);
```
tupaschoal: `CoordsXY` are already initialized to {0, 0} on their default constructor, so this seems superfluous

bentglasstube: Good point, changed to use the default constructor.

src/openrct2/interface/Viewport.cpp
```diff
@@ -1665,25 +1665,40 @@ void get_map_coordinates_from_pos(
 void get_map_coordinates_from_pos_window(
     rct_window* window, int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y, int32_t* interactionType,
     TileElement** tileElement, rct_viewport** viewport)
+{
+  ScreenCoordsXY screenCoords(screenX, screenY);
+  CoordsXY mapCoords();
```
duncanspumpkin: `{}` or just don't put anything. At present it is trying to interpret this as a function definition.

duncanspumpkin:
just need to clang format the changes.

duncanspumpkin:
argh. You've got the same version of clang format as I have. It isn't the same version as what is installed on the CI. You will have to manually fix the file yourself.

Gymnasiast:
Could you rebase this?

duncanspumpkin:
I've fixed the conflict and will squash merge to remove the merge commit.

bentglasstube:
oh thanks, I didn't have a chance to rebase it.

--
Alan Berndt

PR #10112, opened 2019-10-19
Fix importing user strings with umlauts

Gymnasiast:
The problem was caused by unsigned characters getting interpreted as signed ones. For example, 'ä' was interpreted as `-28`, and thus discarded as a format code. `char32_t` is unsigned and also a better fit for the functions in question.

There was also a bug in the S4 importer, which meant that ride names were converted twice.

### DIFF
--- a/src/openrct2/localisation/FormatCodes.cpp
+++ b/src/openrct2/localisation/FormatCodes.cpp
@@ -96,7 +96,7 @@ const char* format_get_token(uint32_t code)
     return nullptr;
 }
 
-bool utf8_should_use_sprite_for_codepoint(int32_t codepoint)
+bool utf8_should_use_sprite_for_codepoint(char32_t codepoint)
 {
     switch (codepoint)
     {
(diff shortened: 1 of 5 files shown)


Gymnasiast:
> If the type change is purely cosmetic, I would prefer to see it in a separate commit.

It isn't purely cosmetic, otherwise I wouldn't have said it fixed the problem.

Gymnasiast:
I suppose that forcing unsignedness by using `char32_t` means that negative signed 8-bit values are first interpreted as unsigned ones. While you are correct that this isn't _supposed_ to be the solution, it does work and it is still a more appropriate data type for the functions in question (after all, codepoints are unsigned).

As for the _actual_ source - apparently the 8-bit `char` can be signed or unsigned, depending on implementation. Might be worth looking there.

AaronVanGeffen:
> As for the _actual_ source - apparently the 8-bit `char` can be signed or unsigned, depending on implementation. Might be worth looking there.

This is what I was getting at — your code isn't changing any of those. I believe you when you say this PR solves the bug; I just don't see how the type change helps here.

PR #10115, opened 2019-10-19
Audio clean-up and style/usage fixes

tomlankhorst:
- use appropriate static casts
- use appropriate dynamic cast
- remove unused headers
- mark 'pure' getters as 'nodiscard'
- remove redundant virtual
- mark override
- default initialize some structs

### DIFF
--- a/src/openrct2-ui/audio/AudioChannel.cpp
+++ b/src/openrct2-ui/audio/AudioChannel.cpp
@@ -10,12 +10,9 @@
 #include "AudioContext.h"
 #include "AudioFormat.h"
 
-#include <SDL.h>
 #include <algorithm>
 #include <cmath>
-#include <openrct2/audio/AudioChannel.h>
 #include <openrct2/audio/AudioSource.h>
-#include <openrct2/audio/audio.h>
 #include <openrct2/common.h>
 #include <speex/speex_resampler.h>
 
@@ -118,12 +115,12 @@ namespace OpenRCT2::Audio
             return false;
         }
 
-        [[nodiscard]] virtual int32_t GetLoop() const override
+        [[nodiscard]] int32_t GetLoop() const override
         {
             return _loop;
         }
 
-        virtual void SetLoop(int32_t value) override
+        void SetLoop(int32_t value) override
         {
             _loop = value;
         }
@@ -227,7 +224,7 @@ namespace OpenRCT2::Audio
 
         void Play(IAudioSource* source, int32_t loop) override
         {
-            _source = static_cast<ISDLAudioSource*>(source);
+            _source = dynamic_cast<ISDLAudioSource*>(source);
             _loop = loop;
             _offset = 0;
             _done = false;
--- a/src/openrct2-ui/audio/AudioChannel.cpp
+++ b/src/openrct2-ui/audio/AudioChannel.cpp
@@ -18,10 +18,12 @@
 
 namespace OpenRCT2::Audio
 {
-    class AudioChannelImpl : public ISDLAudioChannel
+    template<typename AudioSource_ = ISDLAudioSource> class AudioChannelImpl : public ISDLAudioChannel
     {
+        static_assert(std::is_base_of_v<IAudioSource, AudioSource_>);
+
     private:
-        ISDLAudioSource* _source = nullptr;
+        AudioSource_* _source = nullptr;
         SpeexResamplerState* _resampler = nullptr;
 
         int32_t _group = MIXER_GROUP_SOUND;
@@ -63,9 +65,12 @@ namespace OpenRCT2::Audio
             }
         }
 
-        [[nodiscard]] IAudioSource* GetSource() const override { return _source; }
+        [[nodiscard]] IAudioSource* GetSource() const override
+        {
+            return _source;
+        }
 
-            [[nodiscard]] SpeexResamplerState* GetResampler() const override
+        [[nodiscard]] SpeexResamplerState* GetResampler() const override
         {
             return _resampler;
         }
@@ -75,21 +80,30 @@ namespace OpenRCT2::Audio
             _resampler = value;
         }
 
-        [[nodiscard]] int32_t GetGroup() const override { return _group; }
+        [[nodiscard]] int32_t GetGroup() const override
+        {
+            return _group;
+        }
 
         void SetGroup(int32_t group) override
         {
             _group = group;
         }
 
-        [[nodiscard]] double GetRate() const override { return _rate; }
+        [[nodiscard]] double GetRate() const override
+        {
+            return _rate;
+        }
 
         void SetRate(double rate) override
         {
             _rate = std::max(0.001, rate);
         }
 
-        [[nodiscard]] uint64_t GetOffset() const override { return _offset; }
+        [[nodiscard]] uint64_t GetOffset() const override
+        {
+            return _offset;
+        }
 
         bool SetOffset(uint64_t offset) override
         {
@@ -103,30 +117,42 @@ namespace OpenRCT2::Audio
             return false;
         }
 
-        [[nodiscard]] int32_t GetLoop() const override { return _loop; }
+        [[nodiscard]] int32_t GetLoop() const override
+        {
+            return _loop;
+        }
 
         void SetLoop(int32_t value) override
         {
             _loop = value;
         }
 
-        [[nodiscard]] int32_t GetVolume() const override { return _volume; }
+        [[nodiscard]] int32_t GetVolume() const override
+        {
+            return _volume;
+        }
 
-            [[nodiscard]] float GetVolumeL() const override
+        [[nodiscard]] float GetVolumeL() const override
         {
             return _volume_l;
         }
 
-        [[nodiscard]] float GetVolumeR() const override { return _volume_r; }
+        [[nodiscard]] float GetVolumeR() const override
+        {
+            return _volume_r;
+        }
 
-            [[nodiscard]] float GetOldVolumeL() const override
+        [[nodiscard]] float GetOldVolumeL() const override
         {
             return _oldvolume_l;
         }
 
-        [[nodiscard]] float GetOldVolumeR() const override { return _oldvolume_r; }
+        [[nodiscard]] float GetOldVolumeR() const override
+        {
+            return _oldvolume_r;
+        }
 
-            [[nodiscard]] int32_t GetOldVolume() const override
+        [[nodiscard]] int32_t GetOldVolume() const override
         {
             return _oldvolume;
         }
@@ -136,7 +162,10 @@ namespace OpenRCT2::Audio
             _volume = std::clamp(volume, 0, MIXER_VOLUME_MAX);
         }
 
-        [[nodiscard]] float GetPan() const override { return _pan; }
+        [[nodiscard]] float GetPan() const override
+        {
+            return _pan;
+        }
 
         void SetPan(float pan) override
         {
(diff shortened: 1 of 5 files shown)


tomlankhorst:
I might be making a mistake... but Travis' clang-format seems to be inconsistent with mine.

tomlankhorst:
Misconfigured .clang-format, ofcourse.

AaronVanGeffen:
> Travis' clang-format seems to be inconsistent with mine.

We're probably still using an old version of clang-format. I think we have to manually update, and it's been a while.

tomlankhorst:
@AaronVanGeffen Seems @IntelOrca did the latest docker image update of `openrct2/openrct2:format`. Maybe he can run a `docker build`, `docker push`? Or even better, we automate it.

janisozaur:
That's not the right way to do. Build a static version of clang-format or use one already available in angular repos, don't build full OS just for that. I will eventually do that, once I have some more time.

tomlankhorst:
@janisozaur Actually, building an image for a single executable is a typical use-case. And although the `openrct2/openrct2:format` includes many packages that are not used, the corresponding _image layers_ will be reused from other `openrct2/openrct2` images. It's not a full-blown OS in the end, it's just some namespacing with about 3MiB of supporting binaries (in the case of Alpine).

IntelOrca:
> don't build full OS just for that

OS implies a having a kernel right? That is no kernel building involved with docker. As @tomlankhorst pointed out, docker has always intended to be for running single apps, in particular web apps.

src/openrct2-ui/audio/AudioChannel.cpp
```diff
@@ -118,42 +103,30 @@ namespace OpenRCT2::Audio
...
         }
 
-        float GetVolumeR() const override
-        {
-            return _volume_r;
-        }
+        [[nodiscard]] float GetVolumeR() const override { return _volume_r; }
 
-        float GetOldVolumeL() const override
+            [[nodiscard]] float GetOldVolumeL() const override
```
ZehMatt: Is it just GitHub or are there extra spaces, clang-format doesn't complain.

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -481,7 +479,7 @@ namespace OpenRCT2::Audio
             for (int32_t i = 0; i < length; i++)
             {
                 float t = (float)i / length;
-                data[i] = (uint8_t)(data[i] * ((1 - t) * startvolume_f + t * endvolume_f));
+                data[i] = static_cast<float>(data[i]) * ((1 - t) * startvolume_f + t * endvolume_f);
```
ZehMatt: Why not 1.0?

src/openrct2-ui/audio/AudioChannel.cpp
```diff
@@ -220,14 +181,11 @@ namespace OpenRCT2::Audio
...
-        bool IsPlaying() const override
-        {
-            return !_done;
-        }
+        [[nodiscard]] bool IsPlaying() const override { return !_done; }
 
         void Play(IAudioSource* source, int32_t loop) override
         {
-            _source = static_cast<ISDLAudioSource*>(source);
+            _source = dynamic_cast<ISDLAudioSource*>(source);
```
ZehMatt: Whats the reason for dynamic_cast here? Its a much heavier operation than static_cast

src/openrct2-ui/audio/AudioChannel.cpp
tomlankhorst: This must be a bug in an old clang-format since this ís the clang-format diff applied to my original.

src/openrct2-ui/audio/AudioMixer.cpp
tomlankhorst: Consistency with the other casts, I guess, but multiplication with a float works too. Specific preference?

src/openrct2-ui/audio/AudioChannel.cpp
tomlankhorst: Theoretically one might pass a non-ISDLAudioSource in Play which is not what we want. 
The `dynamic_cast` sets `_source` to a `nullptr` when `source` is not a `ISDLAudioSource`. 
However, I guess we know that an `ISDLAudioSource` is passed? In that case we could keep the `static_cast` for performance and drop the little extra provided safety.

ZehMatt: Yeah I don't think its worth it, highly doubt anyone will ever swap it.

src/openrct2-ui/audio/AudioMixer.cpp
ZehMatt: I would prefer that the floats are used with floats and I think that your cast is the wrong way, data is uint8_t and not float

tomlankhorst: Ouch, that is a plain error. I will check this in context and fix it.

src/openrct2-ui/audio/AudioChannel.cpp
tomlankhorst: Let's change this back then.

Gymnasiast:
@tomlankhorst Do you plan to continue work on this PR?

tupaschoal:
I was going to review, but the builds are failing, please take a look :)

tomlankhorst:
> I was going to review, but the builds are failing, please take a look :)

Sorry for the delay. I try to fix it today.

tupaschoal:
No problem/rush, I just didn't know if you saw it, so I commented

Gymnasiast:
I think this should have been split into multiple parts. Some of the plain fixes (like [nodiscard]) are likely to be merged with very little discussion, while other things might take a long time.

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -484,7 +481,7 @@ namespace OpenRCT2::Audio
             for (int32_t i = 0; i < length; i++)
             {
                 float t = static_cast<float>(i) / length;
-                data[i] = static_cast<uint8_t>(data[i] * ((1 - t) * startvolume_f + t * endvolume_f));
+                data[i] = static_cast<float>(data[i]) * ((1 - t) * startvolume_f + t * endvolume_f);
```
tupaschoal, line 484:
```suggestion
                data[i] = static_cast<uint8_t>(data[i] * ((1.0f - t) * startvolume_f + t * endvolume_f));
```
Would this suffice @ZehMatt ?

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -470,8 +467,8 @@ namespace OpenRCT2::Audio
             float endvolume_f = static_cast<float>(endvolume) / SDL_MIX_MAXVOLUME;
             for (int32_t i = 0; i < length; i++)
             {
-                float t = static_cast<float>(i) / length;
-                data[i] = static_cast<int16_t>(data[i] * ((1 - t) * startvolume_f + t * endvolume_f));
+                float t = (float)i / length;
+                data[i] = static_cast<float>(data[i]) * ((1 - t) * startvolume_f + t * endvolume_f);
```
tupaschoal, line 471:
```suggestion
                data[i] = static_cast<int16_t>(data[i] * ((1.0f - t) * startvolume_f + t * endvolume_f));
```

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -470,8 +467,8 @@ namespace OpenRCT2::Audio
             float endvolume_f = static_cast<float>(endvolume) / SDL_MIX_MAXVOLUME;
             for (int32_t i = 0; i < length; i++)
             {
-                float t = static_cast<float>(i) / length;
-                data[i] = static_cast<int16_t>(data[i] * ((1 - t) * startvolume_f + t * endvolume_f));
+                float t = (float)i / length;
```
tupaschoal, line 470:
```suggestion
                float t = static_cast<float>(i) / length;
```
Why would we revert back to c-style cast?

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -432,16 +429,16 @@ namespace OpenRCT2::Audio
...
             float volumeR = channel->GetOldVolumeR();
             const float d_left = dt * (channel->GetVolumeL() - channel->GetOldVolumeL());
             const float d_right = dt * (channel->GetVolumeR() - channel->GetOldVolumeR());
 
             for (int32_t i = 0; i < length * 2; i += 2)
             {
-                data[i] = static_cast<int16_t>(data[i] * volumeL);
-                data[i + 1] = static_cast<int16_t>(data[i + 1] * volumeR);
+                data[i + 0] = volumeL * static_cast<float>(data[i + 0]);
+                data[i + 1] = volumeR * static_cast<float>(data[i + 1]);
```
tupaschoal, line 441:
```suggestion
                data[i + 0] = static_cast<int16_t>(volumeL * static_cast<float>(data[i + 0]));
                data[i + 1] = static_cast<int16_t>(volumeR * static_cast<float>(data[i + 1]));
```

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -456,7 +453,7 @@ namespace OpenRCT2::Audio
 
             for (int32_t i = 0; i < length * 2; i += 2)
             {
-                float t = static_cast<float>(i) / (length * 2);
+                float t = static_cast<float>(i) / static_cast<float>(length * 2);
```
tupaschoal, line 456:
```suggestion
                float t = static_cast<float>(i) / static_cast<float>(length * 2.0f);
```

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -432,16 +429,16 @@ namespace OpenRCT2::Audio
 
         static void EffectPanS16(const IAudioChannel* channel, int16_t* data, int32_t length)
         {
-            const float dt = 1.0f / (length * 2);
+            const float dt = 1.0f / static_cast<float>(length * 2);
```
tupaschoal, line 432:
```suggestion
            const float dt = 1.0f / static_cast<float>(length * 2.0f);
```

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -384,7 +381,7 @@ namespace OpenRCT2::Audio
         int32_t ApplyVolume(const IAudioChannel* channel, void* buffer, size_t len)
         {
             float volumeAdjust = _volume;
-            volumeAdjust *= gConfigSound.master_sound_enabled ? (gConfigSound.master_volume / 100.0f) : 0;
+            volumeAdjust *= gConfigSound.master_sound_enabled ? (static_cast<float>(gConfigSound.master_volume) / 100.0f) : 0;
```
tupaschoal, line 384:
```suggestion
            volumeAdjust *= gConfigSound.master_sound_enabled ? (static_cast<float>(gConfigSound.master_volume) / 100.0f) : 0.0f;
```

src/openrct2-ui/audio/AudioMixer.cpp
```diff
@@ -384,7 +381,7 @@ namespace OpenRCT2::Audio
         int32_t ApplyVolume(const IAudioChannel* channel, void* buffer, size_t len)
         {
             float volumeAdjust = _volume;
-            volumeAdjust *= gConfigSound.master_sound_enabled ? (gConfigSound.master_volume / 100.0f) : 0;
+            volumeAdjust *= gConfigSound.master_sound_enabled ? (static_cast<float>(gConfigSound.master_volume) / 100.0f) : 0.0f;
```
tupaschoal, line 384:
```suggestion
            volumeAdjust *= gConfigSound.master_sound_enabled ? (static_cast<float>(gConfigSound.master_volume) / 100.0f)
                                                              : 0.0f;
```

PR #6764, opened 2017-12-02
Simplify UI scaling display quality options

AaronVanGeffen:
As suggested in a [comment](https://github.com/OpenRCT2/OpenRCT2/pull/6588#issuecomment-346169198) by @Gymnasiast, this PR aims to simplify the UI scaling display options.

This is achieved by removing two options:
* 'Use NN scaling at linear scales', as NN is decidedly preferable at integer scales.
* Anisotropic scaling, as it produces results similar (if not equal) to linear scaling.

Having removed the option for NN scaling at linear scales, perhaps we should consider disabling the dropdown if such a linear scale is selected, to prevent any confusion?

I have only tested this on macOS, where everything seems to be working correctly.

PR #10118, opened 2019-10-20
Cleanup OpenGL code and fix HiDpi OpenGL window size

tomlankhorst:
This fixes an issue where the framebuffer size of the OpenGL engine was set incorrectly. 
Furthermore, this commit improves type safety and code-deduplication. 

image

### DIFF
--- a/src/openrct2-ui/drawing/engines/opengl/DrawCommands.h
+++ b/src/openrct2-ui/drawing/engines/opengl/DrawCommands.h
@@ -27,7 +27,7 @@ public:
         : _numInstances(0)
     {
     }
-    bool empty() const
+    [[nodiscard]] bool empty() const
     {
         return _numInstances == 0;
     }
@@ -51,7 +51,7 @@ public:
         }
         return _instances[_numInstances++] = value;
     }
-    size_t size() const
+    [[nodiscard]] size_t size() const
     {
         return _numInstances;
     }
@@ -114,9 +114,9 @@ struct DrawRectCommand
 
     enum
     {
-        FLAG_NO_TEXTURE = (1 << 2),
-        FLAG_MASK = (1 << 3),
-        FLAG_CROSS_HATCH = (1 << 4),
+        FLAG_NO_TEXTURE = (1U << 2U),
+        FLAG_MASK = (1U << 3U),
+        FLAG_CROSS_HATCH = (1U << 4U),
     };
 };
 
--- a/src/openrct2-ui/drawing/engines/opengl/ApplyPaletteShader.h
+++ b/src/openrct2-ui/drawing/engines/opengl/ApplyPaletteShader.h
@@ -28,7 +28,7 @@ public:
     ApplyPaletteShader();
     ~ApplyPaletteShader() override;
 
-    void SetTexture(GLuint texture);
+    static void SetTexture(GLuint texture);
     void SetPalette(const vec4* glPalette);
 
     void Draw();
--- a/src/openrct2-ui/drawing/engines/opengl/ApplyTransparencyShader.h
+++ b/src/openrct2-ui/drawing/engines/opengl/ApplyTransparencyShader.h
@@ -31,7 +31,8 @@ public:
     ApplyTransparencyShader();
     ~ApplyTransparencyShader() override;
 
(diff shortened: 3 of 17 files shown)


AaronVanGeffen:
Thanks for your PR ­— most of these look like nice improvements. I don't necessarily agree with re-introducing the NN option, however. After some discussion, it was explicitly removed in #6764.

tomlankhorst:
If it's decided this way I will remove it. I'll follow up later today.

Gymnasiast:
The explicit NN option was removed because NN is the _best_ fit at integer scales and rubbish otherwise. As such, we forced it on for integer scales and removed it from the dropdown for non-integer ones. (In fact, I would even argue that we could lose the scaling algorithm dropdown altogether and just force SNN for non-integer scales. But that's something for another PR at the very least.)

data/language/en-US.txt
```diff
@@ -240,6 +240,7 @@ STR_6212    :Gray sandstone
 STR_6274    :Can't set color scheme...
 STR_6307    :Color scheme: {BLACK}{STRINGID}
 STR_6328    :{SMALLFONT}{BLACK}With this option enabled, giant screenshots will have a transparent background instead of the default black color.
+STR_6333    :Nearest Neighbor
```
Gymnasiast: It's not too much of a problem since you're going to remove it anyway, but we *never* update language files from this PR (except en-GB). Those are updated in the Localisation repository and updated nightly.

tomlankhorst:
Thanks for the useful info.

Op zo 20 okt. 2019 om 19:15 schreef Michael Steenbeek <
notifications@github.com>:

AaronVanGeffen:
@tomlankhorst Would you mind splitting off c813374c88c31b8da8a04215efca7476e2d3094b ("Pass SDL_WINDOW_ALLOW_HIGHDPI to SDL_CreateWindow") to a separate PR? I'd like to see it merged for the upcoming release, while the refactor work is a bit more risky.

AaronVanGeffen:
Rebased to leave out the commits that were already been merged in #10125. Should make reviewing this a little easier.

src/openrct2-ui/drawing/engines/opengl/DrawCommands.h
```diff
@@ -114,9 +114,9 @@ struct DrawRectCommand
 
     enum
     {
-        FLAG_NO_TEXTURE = (1 << 2),
-        FLAG_MASK = (1 << 3),
-        FLAG_CROSS_HATCH = (1 << 4),
+        FLAG_NO_TEXTURE = (1U << 2U),
```
Gymnasiast: This is a minor detail, but we always write `1u` etc. for unsigned values.

AaronVanGeffen: Certainly not, as there are no less than 781 instances of `1U` in the codebase. (778 of which are `1UL`.)

Gymnasiast: Hm, in that case I think we should pick one, and apply it throughout. That is something for another PR, though.

tomlankhorst:
Okay, so besides the erroneously failing clang-format check this PR is fine.

Gymnasiast:
@tomlankhorst Would you be able to find out how to modify clang-format so that it passes? (Or alternately, add the corresponding clang-format off tags if it proves too much bother.)

AaronVanGeffen:
We should probably update our clang-format docker image, as it has not been updated in quite a while. This is cause the likely the cause for the differences @tomlankhorst finds.

tomlankhorst:
@IntelOrca maybe you can `docker build`, `docker push`, as you seem to have access to the Docker Hub account of OpenRCT. 
@Gymnasiast I could modify it such that it passes, but the resulting code would be styled incorrectly... and adding the `clang-format off` would disable proper checking as soon as we've updated the image.

IntelOrca:
@tomlankhorst I did not create the original dockerfile, I don't know where the dockerfile for it is.

Gymnasiast:
@tomlankhorst Yes, but unless the docker image gets updated very soon, I don't want to hold up this PR for it. Adding `// clang-format off` around the "offending" code at least makes this mergeable.

IntelOrca:
I have updated the clang-format docker image now, it is now using clang-format 8

tomlankhorst:
Great, let me apply the corresponding changes and then we should be fine.

PR #10120, opened 2019-10-20
Use ScreenCoordsXY for MouseInput

tupaschoal:


src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -109,26 +109,27 @@ void game_handle_input()
...
+        game_handle_input_mouse(screenCoords, state & 0xFF);
     }
 
     if (_inputFlags & INPUT_FLAG_5)
     {
-        game_handle_input_mouse(x, y, state);
+        game_handle_input_mouse(screenCoords, state);
     }
-    else if (x != MONEY32_UNDEFINED)
+    else if (screenCoords.x != MONEY32_UNDEFINED)
```
duncanspumpkin: This looks v. odd. We should find out if this information can be passed back in a nicer manor. Might not be appropriate to change it in this refactor though.

src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -1587,25 +1589,25 @@ void game_handle_edge_scroll()
...
+    input_scroll_viewport(ScreenCoordsXY(scrollX, scrollY));
 }
 
 bool input_test_place_object_modifier(PLACE_OBJECT_MODIFIER modifier)
 {
     return gInputPlaceObjectModifier & modifier;
 }
 
-void input_scroll_viewport(int32_t scrollX, int32_t scrollY)
+void input_scroll_viewport(ScreenCoordsXY scrollScreenCoords)
```
duncanspumpkin: `Screen` is a bit superfluous in the variable name but don't really mind.

src/openrct2-ui/input/MouseInput.cpp
ZehMatt: This looks like dead code, no one ever assigns MONEY32_UNDEFINED to the coordinates.

tupaschoal: It is is very, if you think it's worth investigating, open an issue and I can follow up on that later.

ZehMatt: I've created the issue https://github.com/OpenRCT2/OpenRCT2/issues/10121

PR #10123, opened 2019-10-20
Use ScreenCoordsXY for interface\Widget.h

tupaschoal:


src/openrct2-ui/windows/EditorInventionsList.cpp
```diff
@@ -271,8 +271,9 @@ static ResearchItem* get_research_item_at(int32_t x, int32_t y, int32_t* outScro
         if (widgetIndex == WIDX_PRE_RESEARCHED_SCROLL || widgetIndex == WIDX_RESEARCH_ORDER_SCROLL)
         {
             gPressedWidget.widget_index = widgetIndex;
-            int32_t outX, outY, outScrollArea;
-            widget_scroll_get_part(w, widget, x, y, &outX, &outY, &outScrollArea, outScrollId);
+            int32_t outScrollArea;
+            ScreenCoordsXY outScrollCoords;
```
tupaschoal: Fun fact, these out coordinates are not used in this method

src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -612,16 +612,17 @@ static void input_scroll_begin(rct_window* w, rct_widgetindex widgetIndex, Scree
     gTooltipCursorX = screenCoords.x;
     gTooltipCursorY = screenCoords.y;
 
-    int32_t eax, ebx, scroll_area, scroll_id;
+    int32_t scroll_area, scroll_id;
+    ScreenCoordsXY newScreenCoords;
```
duncanspumpkin: These `newScreenCoords` are the coordinates on the widget. Perhaps `widgetCoords` would be a more descriptive name. I leave it up for debate though as I guess you might think that was the coordinates of the widget and not the coordinates of the mouse on the widget.

src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -928,14 +928,15 @@ static void input_widget_over(ScreenCoordsXY screenCoords, rct_window* w, rct_wi
 
     if (w != nullptr && widgetIndex != -1 && widget->type == WWT_SCROLL)
     {
-        int32_t eax, ebx, scroll_part, edx;
-        widget_scroll_get_part(w, widget, screenCoords.x, screenCoords.y, &eax, &ebx, &scroll_part, &edx);
+        int32_t scroll_part, edx;
```
duncanspumpkin: rename edx whilst your at it to `scrollId`

src/openrct2-ui/windows/EditorInventionsList.cpp
duncanspumpkin: I wander if `scrollY` is just `outScrollCoords.y + 6`

src/openrct2-ui/input/MouseInput.cpp
tupaschoal: I like `scrollCoords` and I think it fits better.

src/openrct2-ui/windows/EditorInventionsList.cpp
tupaschoal: It looks like `outScrollCoords.y + 4`, no?

duncanspumpkin: You have the sign of the 1 wrong. So its `-1 -> 5` which would be 6 different. I think it would be worthwhile changing it in this PR to not recalculate it cause its not like anyone will realise again for a long time.

tupaschoal: Changed it, it's too little to create another PR.

At least these refactorings are exposing some simplifications to us :D

src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -705,8 +706,7 @@ static void input_scroll_continue(rct_window* w, rct_widgetindex widgetIndex, Sc
         return;
     }
 
-    screenCoords.x = x2;
-    screenCoords.y = y2;
+    screenCoords = newScreenCoords;
```
duncanspumpkin: This doesn't need to be done. The scroll coords can be passed directly to the mouse drag call

src/openrct2-ui/windows/EditorInventionsList.cpp
duncanspumpkin: Yes thats one of the main ways of improving the codebase as many of the files go years without being reviewed.

src/openrct2-ui/input/MouseInput.cpp
tupaschoal: Done

PR #10124, opened 2019-10-20
Remove comparison of ScreenCoords with MONEY32_UNDEFINED

### LINKED ISSUE #10121: Investigate and remove potential dead code
The check below doesn't appear to make much sense:
https://github.com/OpenRCT2/OpenRCT2/blob/86474df361a2ba76be74e22ddb3c52cb55b266a2/src/openrct2-ui/input/MouseInput.cpp#L122
It needs investigation to weather or not this branch is ever hit. In case this code is still used one has to figure out where the value is assigned and use a proper constant for the condition.

tupaschoal:
As per investigation on the related issue, it seems like it:
- Doesn't make sense comparing a coordinate with this define
- The `x` is never set as `MONEY32_UNDEFINED`

Thus, I just removed the if, as I'd say it is very likely that it fell into this else very frequently, since it only wouldn't if it was equal to `MONEY32_UNDEFINED`

Closes #10121

### DIFF
--- a/src/openrct2-ui/input/MouseInput.cpp
+++ b/src/openrct2-ui/input/MouseInput.cpp
@@ -120,7 +120,7 @@ void game_handle_input()
     {
         game_handle_input_mouse(screenCoords, state);
     }
-    else if (screenCoords.x != MONEY32_UNDEFINED)
+    else
     {
         int32_t screenWidth = context_get_width();
         int32_t screenHeight = context_get_height();


duncanspumpkin:
Just to be extra safe we will merge it after the next release.

tupaschoal:
Alright. Are we expecting to release soon?

duncanspumpkin:
Yeah there was some chat of doing it this weekend but that usually means next weekend.

tupaschoal:
hahaha, ok

PR #10125, opened 2019-10-20
Pass SDL_WINDOW_ALLOW_HIGHDPI to SDL_CreateWindow

tomlankhorst:
Per @AaronVanGeffen 's request, the hi-dpi handling split of #10118


AaronVanGeffen:
Cheers! I think this deserves a mention in the changelog, too.

IntelOrca:
I think this only applies to macOS? Maybe say that in the changelog.

tomlankhorst:
@IntelOrca I have no HiDPI non-macOS device to test, but the SDL documentation does not mention any MacOS specifics here.

IntelOrca:
@tomlankhorst I think I only found it out by checking the source code. Although it was awhile aback now, but I doubt it has changed.

PR #10127, opened 2019-10-21
Fix #10126: window movement inversed

tupaschoal:
Closes #10126 

Fixes regression introduced by #10120 

My bad, I just copied the implementation of the `operator-` of `CoordsXY` and it was reversed, for some reason.

PR #10128, opened 2019-10-21
Fix CoordsXY sum and subtract operators

tupaschoal:
I found this out when developing #10127 
Luckily, it doesn't make a difference for `operator+`, which was being used in a handful of places, and it does make a huge difference for `operator-`, which was not being used anywhere.


janisozaur:
Nice, I was going to make that suggestion in #10127, but saw you implemented it here. The failing Travis job is fail to upload, it compiles fine.

PR #10131, opened 2019-10-21
Use ScreenCoordsXY for tool_* functions

tupaschoal:
This will free multiple places to migrate once merged and also gives me room while #10123  is not.

### DIFF
--- a/src/openrct2-ui/windows/StaffList.cpp
+++ b/src/openrct2-ui/windows/StaffList.cpp
@@ -45,7 +45,7 @@ static void window_staff_list_resize(rct_window *w);
 static void window_staff_list_mousedown(rct_window *w, rct_widgetindex widgetIndex, rct_widget* widget);
 static void window_staff_list_dropdown(rct_window *w, rct_widgetindex widgetIndex, int32_t dropdownIndex);
 static void window_staff_list_update(rct_window *w);
-static void window_staff_list_tooldown(rct_window *w, rct_widgetindex widgetIndex, int32_t x, int32_t y);
+static void window_staff_list_tooldown(rct_window *w, rct_widgetindex widgetIndex, ScreenCoordsXY screenCoords);
 static void window_staff_list_toolabort(rct_window *w, rct_widgetindex widgetIndex);
 static void window_staff_list_scrollgetsize(rct_window *w, int32_t scrollIndex, int32_t *width, int32_t *height);
 static void window_staff_list_scrollmousedown(rct_window *w, int32_t scrollIndex, int32_t x, int32_t y);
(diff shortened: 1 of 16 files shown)


src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -334,19 +334,20 @@ void window_staff_list_update(rct_window* w)
...
     if (widgetIndex == WIDX_STAFF_LIST_SHOW_PATROL_AREA_BUTTON)
     {
         int32_t selectedPeepType = _windowStaffListSelectedTab;
 
         int32_t direction;
         TileElement* tileElement;
-        footpath_get_coordinates_from_pos(x, y, &x, &y, &direction, &tileElement);
-        if (x == LOCATION_NULL)
+        footpath_get_coordinates_from_pos(
+            screenCoords.x, screenCoords.y, &screenCoords.x, &screenCoords.y, &direction, &tileElement);
```
duncanspumpkin: Create a new local variable for the return values (the &screenCoords) as the return value may not be in `ScreenCoordsXY`

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1216,18 +1216,19 @@ static void window_tile_inspector_dropdown(rct_window* w, rct_widgetindex widget
...
-static void window_tile_inspector_tool_update(rct_window* w, rct_widgetindex widgetIndex, int32_t x, int32_t y)
+static void window_tile_inspector_tool_update(rct_window* w, rct_widgetindex widgetIndex, ScreenCoordsXY screenCoords)
 {
     map_invalidate_selection_rect();
 
     gMapSelectFlags |= MAP_SELECT_FLAG_ENABLE;
 
-    int16_t mapX = x;
-    int16_t mapY = y;
+    int16_t mapX = screenCoords.x;
```
duncanspumpkin: This is horrible. We should refactor this.

mapX and mapY are CoordsXY values but you need to set them to screenCoords.x and screenCoords.y if clickedElement is nullptr below as `screen_pos_to_map_pos` is a horrible function that takes in screen coords and spits out coordsxy to the same variable.

src/openrct2-ui/windows/StaffList.cpp
tupaschoal: Done

src/openrct2-ui/windows/TileInspector.cpp
tupaschoal: I'm assuming this is part of the refactoring and nothing needs to be done here, for now?

duncanspumpkin: Yeah I didn't realise when I wrote this how prevalent the abuse caused by `screen_pos_to_map_pos` was. Leave for now then fix in next refactoring.

src/openrct2-ui/windows/StaffList.cpp
duncanspumpkin: Not too sure if this will compile because although the value should be in CoordsXY range I suspect the function expects a int16_t pointer not a int32_t.

tupaschoal: It did compile incrementally and from clean, but it should, as it expects `int32_t*`

Issue #2468, opened 2015-12-11
Add option to move the [x] to the other side of in game windows.


PFCKrutonium:
For OS X and Various Linux Desktops. Helps keep everything consistent with the OS.

IntelOrca:
* OpenTTD did not do this... and they actually have the X on the other side.
* We can't do this until we re-write the window system anyway because all the widget placement is hard coded.

Gymnasiast:
What OpenTTD did is not really relevant as far as I'm concerned. I think we should implement this eventually, though it's a low priority thing to me.

vzybilly:
For the future, on ubuntu linux, I had a tweak program where I could move
all the elements of the title bar around. I could place them into the left,
center, right and the elements I had to choose from was the Icon, Menu,
Title, minimize, maximize, close. I could even remove all of them which
made for an awkward experience but I could. This might be how we would want
to set this feature up...

side note, it was an old version of ubuntu, while it was still using gnome
and before unity replaced it...

IntelOrca:
@vzybilly this feature is for **in-game** windows.

vzybilly:
I know, I was using a real life example to show how it could be done in
game. it's still a good bit away before we can get coding on it but having
a plan is still good...

PR #10133, opened 2019-10-23
Feature/left closebox

tassaron2:
I play OpenRCT2 on a Linux desktop with window controls on the left (customized KDE). Having the close button in the right of windows within the game is mildly inconvenient because of this. So to help get myself familiar with the code of OpenRCT, I've tried to implement a left-closebox feature. It works but uses a cmake option so it's only useable if you re-compile the game (although it could be the default option for macOS where left-closebox is guaranteed).

I think it would be better as a runtime choice instead of compile-time (so add an --option to the commandline interface of orct2 instead of cmake). But doing that would require more substantial changes to the source code (I think) and I'm not sure if it's a good idea or not. Feedback on this? I am willing to rewrite my pull request if the feature is valid and someone can tell me how they'd prefer it was done. I don't have much experience with C++ but I'll do what I can to help this fantastic project 😀 (Playing RCT natively on Linux is a dream come true.)

I've edited every window except for the multiplayer windows, but I will do the multiplayer windows soon once I've had a chance to play the multiplayer mode. Here's a screenshot of the game with a bunch of windows open:
leftcloseboxrct2


AaronVanGeffen:
This is a feature that was first requested a long time ago: #2468. Nice to see someone picking it up.

Having said that, I'm not a big fan of `#define`-ing this sort of thing with a compile-time switch. I would prefer the buttons were moved depending on a config setting in, say, the window invalidate event.

Other team members might have more suggestions, though. Don't just take my opinion and change everything. ;)

Gymnasiast:
Having a runtime way to switch (with, on Linux, intelligent to detect the default) would be better. But on the other hand, this is better than what we have and if modifying the code to allow for it takes too much effort, I think this could work as well.

If we go down that route, I do prefer a more sophisticated define for the title bar, though, rather than the if-else construction the PR currently uses.

Like Aaron, I'm interested in hearing what the other team members think.

IntelOrca:
This is a prime example of how inflexible the window system is, and why no one has yet done it. I have attempted a few window system re-writes in the past but they are huge undertakings. All windows repeat the same code for the same things including but not limited to the title bar and close button. As @AaronVanGeffen suggested, the cleanest way to go about doing this with the current system is to modify the window shim on the fly in the invalidate event thus allowing you to add a runtime setting. There is already a shared method for resizing the shim widgets to the size of the window, the same method could also place the close button on the desired side.

ocalhoun6:
Seems like the kind of thing that would make a good plugin?

AaronVanGeffen:
> Should all windows have anchor_border_widgets or invalidate?

The two are not mutually exclusive. The former is a static function called by the latter. Consider `anchor_border_widgets` a helper function rather than an event, like `invalidate`.

As for the name, I agree `invalidate` is a little confusing. For OpenLoco, we ended up renaming this same event to `prepare_draw` instead.

Feel free to join us [on Gitter](https://gitter.im/OpenRCT2/OpenRCT2) for development questions, by the way :)

src/openrct2-ui/windows/About.cpp
```diff
@@ -44,10 +44,20 @@ enum WINDOW_ABOUT_WIDGET_IDX {
...
+    #define CLOSEBOX_WIDGET \
+        { WWT_CLOSEBOX,         0,  2,                  12,         2,          13,         STR_CLOSE_X,                            STR_CLOSE_WINDOW_TIP }
+#endif
+
 #define WIDGETS_MAIN \
       WWT_FRAME,            0,  0,              WW - 1,         0,          WH - 1,     0xFFFFFFFF,                             STR_NONE },             /* panel / background */ \
-    { WWT_CAPTION,          0,  1,              WW - 2,         1,          14,         STR_ABOUT,                              STR_WINDOW_TITLE_TIP }, /* title bar */ \
-    { WWT_CLOSEBOX,         0,  WW - 13,        WW - 3,         2,          13,         STR_CLOSE_X,                            STR_CLOSE_WINDOW_TIP }, /* close x button */ \
+    { WWT_CAPTION,          0,  1,              WW - 2,         1,          14,         STR_ABOUT,                              STR_WINDOW_TITLE_TIP }, \
+    CLOSEBOX_WIDGET, \
```
Gymnasiast: I think it should be possible to replace the first three definitions of every window with a define that takes a window title, window height and window width. This define can then take left/right close button status into account.

src/openrct2-ui/windows/Window.h
```diff
@@ -168,3 +168,17 @@ void window_tooltip_reset(ScreenCoordsXY screenCoords);
 void window_tooltip_show(rct_string_id id, ScreenCoordsXY screenCoords);
 void window_tooltip_open(rct_window* widgetWindow, rct_widgetindex widgetIndex, ScreenCoordsXY screenCoords);
 void window_tooltip_close();
+
+#ifndef LEFT_CLOSEBOX
+    // close button on right of window
+    #define WINDOW_SKELETON \
+        { WWT_FRAME,    0,  0,  WW - 1, 0,  WH, 0xFFFFFFFF,                                 STR_NONE }, \
```
Gymnasiast: This is indeed roughly what I had in mind. I'd like to see two things changed:
- `WINDOW_SKELETON` should be a "function" that takes the window name, width and height: `#define WINDOW_SKELETON(title, width, height) \` and change the body accordingly.
- On the old `WIDGETS_MAIN`, we deliberately left out the first and last brace, so that we could use `{ WIDGETS_MAIN }` in our window definitions. Please replicate that.

When you apply these changes, the window definitions themselves will look like this:
```
static rct_widget window_staff_overview_widgets[] = {
    { WINDOW_SKELETON(STR_STRINGID, WW, WH) },
    { WWT_RESIZE,   1, 0,       WW - 1,     43,         WH - 1, 0xFFFFFFFF,             STR_NONE },             // Resize
```

Gymnasiast:
Superseded by https://github.com/OpenRCT2/OpenRCT2/pull/10133 .

plastmassor33:
Sorry for asking here, I wasn't able to create an account on the forums.
Can I still compile a version of OpenRCT2 with the **left-side close button** commits? I'm no developer and it's hard for me to figure out how to clone it properly and what arguments should I give to CMake to compile it this way. 
I'm also on kubuntu can't stand this button being on the right :)

tassaron:
> Can I still compile a version of OpenRCT2 with the **left-side close button** commits? I'm no developer and it's hard for me to figure out how to clone it properly and what arguments should I give to CMake to compile it this way.

If you want to use a 3-years-old version, you can probably compile my fork. The extra cmake flag is `-DLEFT_CLOSEBOX`.

Applying the commits to a newer version wouldn't be trivial. I'm busy the next 8 hours so I don't have time to read the code right now, unfortunately.

AaronVanGeffen:
Over the past year or two, OpenRCT2 has slowly been transitioning towards a new window system. It should be much cleaner (and therefore easier) to add left-handed close buttons once this is done.

Gymnasiast:
I ported some bits over to #11653. Later on, I also added code to resize the shim, which was the most important missing piece. I just never got around to doing the last bit. It should not be _that_ hard to hack where the button is drawn, although you’d still need to compile. If you ask very nicely, I might take a look at it once some of my open PRs are finally merged.

AaronVanGeffen:
I was thinking about a dynamic option rather than a compile-time option. The introduction of `WINDOW_SHIM` makes adjusting it much easier, though, for sure. (Not sure if it's also used for plugin windows, though?)

Gymnasiast:
Sorry, I should have been clearer. What I meant was that I never got around to finishing the last bit, namely exposing it as a setting/dynamic option. It already uses a common function to dynamically resize the window frame/shim for all windows as well (this was previously just duplicate code in every single window). My point about compiling was merely that you could have it now by doing an easy local hack.

> Not sure if it's also used for plugin windows, though?

I am quite sure it does - I think plugins provide a window skeleton through CustomWindow.cpp. Haven’t checked though, so definitely something to keep in mind while testing. One or two high-profile plugins such as the Scenery Manager, Ride Vehicle Editor or Peep Editor should be sufficient testing material, I reckon.

Issue #10114, opened 2019-10-19
Refactor SteamParticleOffsets


duncanspumpkin:
Super easy refactor. `SteamParticleOffsets` should be an array of 16 offsets for 3 different types. At present it is an array of 48 offsets.

`Vehicle.cpp`

SijmenSchoon:
Can be closed

PR #10134, opened 2019-10-23
Refactor SteamParticleOffsets

cskrisz:
Turn flat array of 48 offsets into arrays of 16 offsets for 3 different
types.

#10114

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -7259,8 +7259,10 @@ static void vehicle_update_additional_animation(rct_vehicle* vehicle)
                                 switch (vehicle->vehicle_sprite_type)
                                 {
                                     case 2:
+                                        // uphill
                                         return 1;
                                     case 6:
+                                        // downhill
                                         return 2;
                                     default:
                                         return 0;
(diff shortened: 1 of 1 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7251,18 +7255,20 @@ static void vehicle_update_additional_animation(rct_vehicle* vehicle)
...
-                            {
-                                index += 32;
-                            }
-                            steam_particle_create(
-                                vehicle->x + SteamParticleOffsets[index].x, vehicle->y + SteamParticleOffsets[index].y,
-                                vehicle->z + SteamParticleOffsets[index].z);
+                            int32_t typeIndex = [&] {
+                                switch (vehicle->vehicle_sprite_type)
+                                {
+                                    case 2:
```
AaronVanGeffen: Do we have any constants for these yet? If not, there really should be.

duncanspumpkin: Case 2 is for going uphill, case 6 is for going downhill. I don't think there are constants for these though. It would require quite a lot of investigation to work out. I think for now just leave a comment stating that case 2 is uphill and case 6 is downhill.

PR #10135, opened 2019-10-24
Use constexpr on openrct2/*

tupaschoal:
I don't know your feelings wrt this changeset. Some things I took into consideration:
- I only changed #defines on source files
- I checked their usage to define the correct type (which in turn allowed for some static_casts to be removed)
- While checking, I found some defines that were never used.

There are still other `#define`s on the `src/openrct2` tree, but they're mostly for macros.


duncanspumpkin:
Ideally they would lose there all caps name and end up in a namespace but this is also an improvement. Some of the defines you have deleted are important such as the td6 one. I need to add that to the exporter to make sure it's not exceeded.

tupaschoal:
> Ideally they would lose there all caps name and end up in a namespace but this is also an improvement.

That's easy enough to do, I can do it on another PR, if you wish. What would you name the namespace and which convention for the variables, like `g_` to denote global?

> Some of the defines you have deleted are important such as the td6 one. I need to add that to the exporter to make sure it's not exceeded.

You mean important, but unused right? Because it compiles cleanly (they were the only reference in the whole code)

duncanspumpkin:
Yeah i mean important but have been forgotten.

Don't change the case right now. We can fix it later.

src/openrct2/ride/TrackDesignSave.cpp
```diff
@@ -32,9 +32,8 @@
 
 #include <algorithm>
 
-#define TRACK_MAX_SAVED_TILE_ELEMENTS 1500
-#define TRACK_NEARBY_SCENERY_DISTANCE 1
-#define TRACK_TD6_MAX_ELEMENTS 8192
```
tupaschoal: @duncanspumpkin this is kind of repeated here:
https://github.com/OpenRCT2/OpenRCT2/blob/e9bcad193ac098ecd8bb73db6fbb24fb94dd86b3/src/openrct2/ride/TrackDesign.cpp#L249-L251

I know it's not on the exporter code, but maybe it's what you meant?
`8192` is not found on any exporter code. If you know where to place it, I can change it as part of the PR

duncanspumpkin: Ah cool it's in a better location than i thought. Delete away then.

tupaschoal: Alright!

src/openrct2/ride/Vehicle.cpp
```diff
@@ -86,12 +86,14 @@ static SoundId vehicle_update_scream_sound(rct_vehicle* vehicle);
 static void vehicle_kill_all_passengers(rct_vehicle* vehicle);
 static bool vehicle_can_depart_synchronised(rct_vehicle* vehicle);
 
-#define VEHICLE_INVALID_ID (-1)
+constexpr int VEHICLE_INVALID_ID = -1;
```
duncanspumpkin: int32_t

tupaschoal: Addressed.
@duncanspumpkin Do you mean everywhere that addresses `trainIndex` should be refactored to `int32_t`? I can do that on a follow-up PR

duncanspumpkin: yeah.

tupaschoal:
> Ideally they would lose there all caps name and end up in a namespace but this is also an improvement. Some of the defines you have deleted are important such as the td6 one. I need to add that to the exporter to make sure it's not exceeded.

Hey @duncanspumpkin, what is the guideline in here, convergin to "_someConstExpr" or to "gConstExpr"? I'm assuming the first. Also, did you think about a generic namespace or something specific to each case? I intend to tackle this next and wanted to clarify

PR #10136, opened 2019-10-24
Use constexpr on openrct2-ui/*

tupaschoal:
Same considerations as in #10135:

> I don't know your feelings wrt this changeset. Some things I took into consideration:
> 
> * I only changed #defines on source files
> * I checked their usage to define the correct type (which in turn allowed for some static_casts to be removed)
> * While checking, I found some defines that were never used.
> 
> There are still other `#define`s on the `src/openrct2` tree, but they're mostly for macros.


src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -275,10 +275,10 @@ static rct_widget DefaultWidgets[] = {
 };
 
 // Group boxes .top and .bottom for a given window height offsets from the bottom
-#define SUR_GBPB PADDING_BOTTOM                 // Surface group box properties bottom
-#define SUR_GBPT (SUR_GBPB + 16 + 4 * 21)       // Surface group box properties top
-#define SUR_GBDB (SUR_GBPT + GROUPBOX_PADDING)  // Surface group box details bottom
-#define SUR_GBDT (SUR_GBDB + 20 + 4 * 11)       // Surface group box details top
+constexpr int SUR_GBPB = PADDING_BOTTOM;                   // Surface group box properties bottom
```
Broxzier:
```suggestion
constexpr int32_t SUR_GBPB = PADDING_BOTTOM;               // Surface group box properties bottom
```

tupaschoal: Done

PR #10141, opened 2019-10-26
Properly fix usage of station index string

Broxzier:
This basically reverts the change in en-GB.txt from #10077 and fixes the usage of this string where it was broken.


Gymnasiast:
If we merge this so close before the release, perhaps we should also modify the other languages ourselves. If needed, we should even modify them in this repo.

Broxzier:
Yes; since this doesn't touch any localised words it should be easy enough to do. Best to do it in the Localisation repo of course. I can do that as soon as this gets merged.

Gymnasiast:
@Broxzier Good, your time starts now! ;-)

PR #10144, opened 2019-10-26
Refactor vehicle_visual_virginia_reel

cskrisz:


duncanspumpkin:
If we are going to do a few of these refactors we should probably define what each value means:

Sprite Type:
0 : 0 deg
1 : 15 deg Up
2 : 30 deg Up
3 : 45 deg Up
4 : 60 deg Up
5 : 15 deg Down
6 : 30 deg Down
7 : 45 deg Down
8 : 60 deg Down
9 : 75 deg Up
10 : 90 deg Up
11 : 105 deg Up
12 : 120 deg Up
13 : 135 deg Up
14 : 150 deg Up
15 : 165 deg Up
16 : 180 deg
17 : 75 deg Down
18 : 90 deg Down

After this your on your own for working it out but I'm pretty sure it just does a bunch of down direction versions of them.

Issue #10145, opened 2019-10-26
Key names are not translated


Wuzzy2:
**OS:** GNU/Linux
**Version:** 0.2.3-462
**Commit/Build:** e9bcad1

- [ ] Reproducible in RCT2 (vanilla)? I don't know.
- [ ] Specific to multiplayer? No.

**Description:**
The special key names like “Delete”, “Page up”, “Backspace”, etc. in the key binding window are not translated, despite having translations available in the language files.
However, CTRL and SHIFT keys are translated (as part of key combinations).

**Steps to reproduce:**
1. Select German
2. Open the key binding window
3. Look for keys like Delete Page up, Backspace, etc.

cskrisz:
Translations for those seem to be unimplemented. See [StringIds.h](https://github.com/OpenRCT2/OpenRCT2/blob/e9bcad193ac098ecd8bb73db6fbb24fb94dd86b3/src/openrct2/localisation/StringIds.h#L1932-L2086).

IntelOrca:
Duplicate of #1164

Issue #1164, opened 2015-05-29
Keys in option 'Shortcut keys...' not translatable


danidoedel:
![2015-05-29 19_02_17-openrct2](https://cloud.githubusercontent.com/assets/10995998/7888049/ef867e16-0635-11e5-8053-91a06898c68c.png)

Gymnasiast:
A clarification: some keyboards do translate buttons: on German keyboards, Ctrl is labeled Strg., for example. (On French keyboards, it isn't renamed, oddly enough.)

Patrik356b:
Not so odd, keys aren't renamed on Swedish keyboards either.

Gymnasiast:
@Patrick356: I said 'oddly enough' since France tends to have more language purism than many other countries.

Gymnasiast:
Hm, the string _are_ in the language files. But not used, apparently.

IntelOrca:
Requires changes to:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/interface/keyboard_shortcut.c#L103

Gymnasiast:
Fixed by https://github.com/OpenRCT2/OpenRCT2/pull/10147.

PR #10147, opened 2019-10-26
Use available translations for shortcut keys

AaronVanGeffen:
Implements #1164, #10145. The former has been open for a long time, so I figured I'd have a go.

Before:
Screenshot_20191026_204828

After:
Screenshot_20191026_204543

To do:
- [x] I think we should do some housekeeping. There are many '???' strings that are not used, even after this PR. Moreover, keys like 'A', '0', or even 'F1' don't need a translation, so perhaps we should just rely on SDL for their names, too?
- [x] Move the `specialCharNames` map elsewhere?


cskrisz:
We should map to `SDL_Keycode`s as that takes the keyboard layout into account. In the current form, if a german player sets up 'Y' as a shortcut key then a 'Z' would appear.

AaronVanGeffen:
That would be a great change, but that's way bigger in scope than what I intend to do with this PR.

cskrisz:
Using scancodes for input makes sense for most games. I only mean the display of the characters in the shortcut settings. The keys in `specialCharNames` would be `SDLK_*` constants and it could be indexed with the result of `SDL_GetKeyFromScancode`.

Gymnasiast:
@AaronVanGeffen Could you amend the changelog? I'll merge it as soon as that's done.

@cskrisz That sounds like a good approach - would you like to make a follow-up PR for that?

AaronVanGeffen:
@Gymnasiast I have two follow-ups I would like to tackle in this PR first. Could you comment on the removal of unused key strings?

Gymnasiast:
Removing them sounds like a good idea to me. I think you're right about F1-F24, letter and special character keys not needing translation.

data/language/en-GB.txt
```diff
@@ -1858,39 +1858,38 @@ STR_2521    :Show staff list
...
-### The following need to be reordered to match SDL_keycode layout.
-STR_2525    :???
-STR_2526    :???
-STR_2527    :???
-STR_2528    :???
-STR_2529    :???
-STR_2530    :???
-STR_2531    :???
-STR_2532    :???
+STR_2525    :<not used anymore>
```
AaronVanGeffen: Should we just leave these out of the file altogether? I only later noticed there were only very few of these in the file.

AaronVanGeffen:
For reference: regex to easily find and remove affected strings from other language files:

```
^STR_(2525|2526|2527|2528|2529|2530|2531|2532|2535|2536|2539|2540|2541|2542|2546|2547|2548|2549|2550|2551|2553|2554|2555|2556|2573|2574|2575|2576|2577|2578|2579|2580|2581|2582|2583|2584|2585|2586|2587|2588|2589|2590|2591|2592|2593|2594|2595|2596|2597|2598|2599|2600|2601|2602|2603|2604|2605|2606|2607|2608|2609|2610|2611|2612|2613|2614|2615|2616|2617|2619|2620|2633|2637|2638|2639|2640|2641|2642|2643|2644|2645|2646|2647|2648|2649|2650|2651|2652|2653|2654|2655|2656|2657|2658|2659|2660|2661|2662|2663|2664|2665|2666|2667|2668|2671|2672|2673|2674|2675|2676|2677|2678|2679|2697|2698|2699|2711|2712|2713|2714|2715|2716|2717|2744|2745|2746|2747|2748)
```

data/language/en-GB.txt
Gymnasiast: Might as well.

AaronVanGeffen:
@Gymnasiast Good to go?

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -153,7 +199,17 @@ std::string KeyboardShortcuts::GetShortcutString(int32_t shortcut) const
...
+
+    SDL_Scancode scanCode = static_cast<SDL_Scancode>(shortcutKey & 0xFF);
+    auto keyPair = specialCharNames.find(scanCode);
+    if (keyPair != specialCharNames.end())
+    {
+        format_string(formatBuffer, sizeof(formatBuffer), keyPair->second, nullptr);
+        String::Append(buffer, sizeof(buffer), formatBuffer);
+    }
+    else
+        String::Append(buffer, sizeof(buffer), SDL_GetKeyName(SDL_GetKeyFromScancode(scanCode)));
```
Gymnasiast: Could you put parenthese around this statement? I prefer keeping it consistent within one if-else block.

src/openrct2/localisation/StringIds.h
```diff
@@ -2704,6 +2608,8 @@ enum
     STR_SET_PATROL_AREA = 3445,
     STR_CLEAR_PATROL_AREA = 3446,
 
+    // 3447--5119 unused
```
Gymnasiast: I believe the user strings used to be put here (or still are).

Gymnasiast:
Almost :). I have left one line note I'd like to see addressed (the one about the braces), and one line note about a comments that is optional.

src/openrct2/localisation/StringIds.h
AaronVanGeffen: I believe they used to, but no longer are. @IntelOrca, do you remember?

src/openrct2-ui/input/KeyboardShortcuts.cpp
AaronVanGeffen: Heh, I prefer the opposite. I'm not a fan of placing unessential braces. As both styles are used in the code base, I'd like to leave it as-is.

Gymnasiast: I think a block should either do
```
if (foo)
    bar();
else
    baz();
```
OR
```
if (foo)
{
    bar();
}
else
{
    baz();
}
```

Not mixing the two. That is just ugly and hard to read.

src/openrct2/localisation/StringIds.h
IntelOrca: I don't remember, sorry.

src/openrct2-ui/input/KeyboardShortcuts.cpp
AaronVanGeffen: Whether or not it is "just ugly and hard to read" really is a matter of opinion, but not a hill I want this to die on. I have added an amending commit.

src/openrct2/localisation/StringIds.h
AaronVanGeffen: I have rephrased the comment to mention user strings. We can free up the block once we are certain it can be used.

PR #10149, opened 2019-10-27
Fix desync in headless mode with rides that create smoke particles

ZehMatt:
ride_has_station_shelter checked that BaseImageId != 0 but when the server runs in headless mode this is always 0 while on clients its non-zero. To compensate for network play it will only check if the station object exists.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -7399,6 +7399,11 @@ bool ride_has_adjacent_station(Ride* ride)
 bool ride_has_station_shelter(Ride* ride)
 {
     auto stationObj = ride_get_station_object(ride);
+    if (network_get_mode() != NETWORK_MODE_NONE)
+    {
+        // The server might run in headless mode so no images will be loaded, only check for stations.
+        return stationObj != nullptr;
+    }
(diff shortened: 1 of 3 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -7399,6 +7399,11 @@ bool ride_has_adjacent_station(Ride* ride)
 bool ride_has_station_shelter(Ride* ride)
 {
     auto stationObj = ride_get_station_object(ride);
+    if (network_get_mode() != NETWORK_MODE_NONE)
+    {
+        // The server might run in headless mode so no images will be loaded, only check for stations.
+        return stationObj != nullptr;
```
Gymnasiast: This means that plain stations will be regarded as sheltered in network play. After all, this can be the only reason it even checks for `BaseImageId`. I would prefer if it could check for the number of images in the object, or if that isn't possible, if we could add a flag to station objects that determines this status.

ZehMatt: In headless mode there are no image information at all. But yeah I would think its better to flag specific objects as sheltered since such metadata shouldn't rely on the image.

PR #10161, opened 2019-10-29
Name field var_C8 of vehicle struct. Refactor to use it as uint32

duncanspumpkin:


src/openrct2/GameStateSnapshots.cpp
```diff
@@ -364,8 +364,7 @@ struct GameStateSnapshots : public IGameStateSnapshots
         {
             COMPARE_FIELD(rct_vehicle, pad_C6[i]);
         }
-        COMPARE_FIELD(rct_vehicle, var_C8);
-        COMPARE_FIELD(rct_vehicle, var_CA);
+        COMPARE_FIELD(rct_vehicle, additional_animaition_var);
```
AaronVanGeffen:
```suggestion
        COMPARE_FIELD(rct_vehicle, additional_animation_var);
```

src/openrct2/rct1/RCT1.h
```diff
@@ -306,8 +306,15 @@ struct rct1_vehicle : RCT12SpriteBase
...
     uint8_t pad_C6[0x2];
-    uint16_t var_C8;
-    uint16_t var_CA;
+    union
+    {
+        uint32_t additional_animation_var;
+        struct
+        {
+            uint16_t var_C8;
+            uint16_t var_CA;
```
ZehMatt: Do we still need this variable? It looks like it was only ever used a temporary variable.

src/openrct2/rct1/RCT1.h
```diff
@@ -306,8 +306,15 @@ struct rct1_vehicle : RCT12SpriteBase
...
     uint8_t animation_frame;
     uint8_t pad_C6[0x2];
-    uint16_t var_C8;
-    uint16_t var_CA;
+    union
+    {
+        uint32_t additional_animation_var;
+        struct
+        {
+            uint16_t var_C8;
```
Gymnasiast: Is it necessary to split this out? If nothing directly references it any more, a comment would do IMO.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7354,17 +7353,16 @@ static void vehicle_update_additional_animation(rct_vehicle* vehicle)
...
-                if (vehicle->var_C8 <= 0xCCCC)
+                if (vehicle->additional_animation_var <= 0xCCCC)
                 {
-                    vehicle->var_C8 += 0x3333;
+                    vehicle->additional_animation_var += 0x3333;
                 }
                 else
                 {
-                    vehicle->var_C8 += 0x3333;
+                    vehicle->additional_animation_var = 0;
```
Gymnasiast: Why this change?

duncanspumpkin: Previously this relied on rolling over. This just makes it explicit that it is resetting to 0.

Gymnasiast: What if the previously value is not exactly 0xCCCC, but higher? It wouldn't be 0 then.

spacek531: Current behavior:

On the first cycle, it updates the frame when animationState is 0xFFFF (0x3333 + 0xCCCC). 0x3333 + 0xFFFF is 0x3332.
On the next cycle, it updates the frame when animationState is 0xFFFE (0x3333 + 0xCCCB), and decreases by one every future update.

After 0x3333 frame updates, animationState will finally reach zero. I was concerned that it might insert a 6th frame somewhere, but I think it won't do that.

duncanspumpkin: Even if it did insert that extra frame every 0x3333 cycles it would be silly to keep it like that.

duncanspumpkin:
Superseeded by #15267

PR #10167, opened 2019-10-30
Refactor probability

duncanspumpkin:
Still playing about with this. Not too sure on my approach.

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -515,7 +515,7 @@ void Guest::UpdateEasterEggInteractions()
 
     if (peep_flags & PEEP_FLAGS_JOY)
     {
-        if (scenario_rand() <= 1456)
+        if (scenario_rand() <= 1456) // Mistake should be 0xFFFF
         {
             if (action == PEEP_ACTION_NONE_1 || action == PEEP_ACTION_NONE_2)
             {
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -515,7 +515,7 @@ void Guest::UpdateEasterEggInteractions()
 
     if (peep_flags & PEEP_FLAGS_JOY)
     {
-        if (scenario_rand() <= 1456) // Mistake should be 0xFFFF
+        if (!scenario_rand_probability16(Probability::_02_2_Percent))
         {
             if (action == PEEP_ACTION_NONE_1 || action == PEEP_ACTION_NONE_2)
             {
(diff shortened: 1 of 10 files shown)


AaronVanGeffen:
How about making a `constexpr` function that returns the right probability using a switch case instead? The percentage (?) could just be an argument, which I believe it would improve readability considerably.

IntelOrca:
We can just use floating point literals to get the right hex probabilities.
E.g.

```
constexpr unsigned short probability(double f) {
    return 65535 * (0.142 / 100);
}

unsigned short test() {
    return probability(0.142);
}
```

test() will return compile time constant 93.

duncanspumpkin:
```
GetProbability16(0.142));
GetProbability16(0.182));
GetProbability16(0.2));
GetProbability16(0.333));
GetProbability16(0.5));
GetProbability16(1.0));
GetProbability16(1.4283));
GetProbability16(2.0));
GetProbability16(2.221714));
GetProbability16(2.5));
GetProbability16(3.333));
GetProbability16(3.516));
GetProbability16(3.7034));
GetProbability16(4.0));
GetProbability16(4.166));
GetProbability16(4.3473));
GetProbability16(5.0));
GetProbability16(6.2501));
GetProbability16(6.667));
GetProbability16(8.333));
GetProbability16(10.0));
GetProbability16(11.1101));
GetProbability16(12.5002));
GetProbability16(14.2855));
GetProbability16(20.0));
GetProbability16(16.666));
GetProbability16(25.0004));
GetProbability16(33.333333333333333));
GetProbability16(50.001));
GetProbability16(62.501));
```
So the above numbers will give the same values. Some of them are a bit annoying.

AaronVanGeffen:
Instead of `33.333333333333333` you could just write `100.0 / 3.0`, right?

duncanspumpkin:
Yeah you could but is it easier to read.

AaronVanGeffen:
Haha, I was trying to argue the opposite. Personal preference, I guess. :)

Gymnasiast:
I agree your initial approach in this PR isn't the best way to refactor this (it gets harder to read IMO). I'm interested in how @IntelOrca 's approach will work out.

By the way, I don't think probability necessarily _needs_ refactoring, though I have nothing against it either (provided the code stays readable, of course).

AaronVanGeffen:
The constants are a little awkward. Is the plan to drop them altogether in favour of the constexpr `GetProbability16`? I believe doing so would alleviate the issues raised w.r.t. readability.

duncanspumpkin:
Yeah i was thinking of reverting basically all my changes except get probability 16 is used to calculate the probability

Gymnasiast:
If you get round to that, it's probably better to create a new PR, as this one has been untouched for 4 months now.

Issue #9943, opened 2019-08-27
Refactor out oddball coordinates


Gymnasiast:
Apart from big and small coordinates (Coords and TileCoords), there are a few oddballs, where the game uses other coordinates. This includes stuff like using TileCoords, but with z increasing half as quickly as normal, and possibly even mixing components from big and small coordinates within a struct or function. All structs and functions should use either big or small coordinates.

Additionally, the use of Location8/16/32 should be limited to SV6/TD6 structs. Eventually, these structs should be moved to `RCT12.h`.

Gymnasiast:
SurfaceElement::G/SetWaterHeight() uses "half-tilez". Most calls then multiply it by 16, the rest usually by 2. Going to change it to return big Zs.

Gymnasiast:
As of 2025, I have not encountered one for years. Going to assume this is done.

PR #10169, opened 2019-10-30
Change a sizable portion of LocationXY(Z) to (Tile)CoordsXY(Z)

SijmenSchoon:
This is the first working state from changing the type of `paint_session.MapPosition` to `CoordsXY`. One small step towards fixing #9943 

I double checked to not mix up any `Coords` and `TileCoords` variables, but I might have missed some.


src/openrct2/paint/Supports.cpp
```diff
@@ -16,16 +16,16 @@
...
-    {4,  4},
-    {28, 4},
-    {4,  28},
-    {28, 28},
-    {16, 16},
-    {16, 4},
-    {4,  16},
-    {28, 16},
-    {16, 28}
+static constexpr const CoordsXY SupportBoundBoxes[] = {
```
tupaschoal: You could have changed only the type of the `constexpr` to `CoordsXY`, it would know how to initialize itself with the initializer list above (see [example](https://godbolt.org/z/NvYi1w)). This applies to many other places in the patch.

Also if you hadn't changed the variable name, it would make it easier to review, as there would be less formatting changes due to the new width ;P

SijmenSchoon: I know, but I found using the constructors prettier. I can change it back if others agree though

duncanspumpkin: I don't think we really do it like this anywhere else in the codebase. I think it would look neater keeping the old format.

SijmenSchoon: Alright, I will change it back tonight

src/openrct2/paint/Supports.cpp
```diff
@@ -772,8 +772,8 @@ bool metal_a_supports_paint_setup(
     }
     else
     {
-        int8_t xOffset = loc_97AF20[segment].x;
-        int8_t yOffset = loc_97AF20[segment].y;
+        int8_t xOffset = SupportBoundBoxes[segment].x;
```
ZehMatt: We can avoid this assignment if we refactor sub_98196C to also use CoordsXY, but this should be done in another PR. Right now it accesses the memory 3 times which isn't too nice.

SijmenSchoon: What do you suggest I do right now?

ZehMatt: Nothing, like I stated this should be done in another PR honestly as it will change the declaration of those functions and requires a lot more modifications.

src/openrct2/ride/TrackPaint.cpp
```diff
@@ -29,15 +29,15 @@
 
 // clang-format off
 /* rct2: 0x007667AC */
-static LocationXY16 loc_7667AC[] = {
+static TileCoordsXY EntranceOffsetEdgeNE[] = {
```
ZehMatt: Can use constexpr here.

src/openrct2/ride/TrackPaint.cpp
```diff
@@ -29,15 +29,15 @@
...
     { -1, 0 },
     { 0,  -1 },
     { 1,  0 },
     { 0,  1 },
 };
 
 /* rct2: 0x007667AE */
-static LocationXY16 loc_7667AE[] = {
+static TileCoordsXY EntranceOffsetEdgeNW[] = {
```
ZehMatt: Can also use constexpr.

src/openrct2/paint/Supports.cpp
SijmenSchoon: Alright, cool

src/openrct2/ride/TrackPaint.cpp
SijmenSchoon: Can't currently, since `TileCoordsXY(int32_t, int32_t)` is not `constexpr` (unlike `TileCoordsXYZ(int32_t, int32_t, int32_t)`. I suppose this is an error, I'll change that too

SijmenSchoon: There's a bunch of `const CoordsXY(Z)` declarations in this file - should I also change these to `constexpr`s?

SijmenSchoon: There's a bunch of `const CoordsXY(Z)`s in this file, should I change these to `constexpr`s too?

SijmenSchoon:
@tupaschoal I wanted to request a review, but the button does not work

duncanspumpkin:
I suspect you can only request review from team members it's the same for assigning issues.

PR #10170, opened 2019-10-31
Use int32_t instead of int for train_index

tupaschoal:
As proposed by @duncanspumpkin on #10135

> int32_t (not sure why trainIndex has been defined as an int should have been int32_t)

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -86,8 +86,6 @@ static SoundId vehicle_update_scream_sound(rct_vehicle* vehicle);
 static void vehicle_kill_all_passengers(rct_vehicle* vehicle);
 static bool vehicle_can_depart_synchronised(rct_vehicle* vehicle);
 
-constexpr int32_t VEHICLE_INVALID_ID = -1;
-
 constexpr int16_t VEHICLE_MAX_SPIN_SPEED = 1536;
 constexpr int16_t VEHICLE_MIN_SPIN_SPEED = -VEHICLE_MAX_SPIN_SPEED;
 constexpr int16_t VEHICLE_MAX_SPIN_SPEED_FOR_STOPPING = 700;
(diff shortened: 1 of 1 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -2215,7 +2215,7 @@ static void train_ready_to_depart(rct_vehicle* vehicle, uint8_t num_peeps_on_tra
     vehicle->SetState(VEHICLE_STATUS_WAITING_FOR_PASSENGERS);
 }
 
-static int ride_get_train_index_from_vehicle(Ride* ride, uint16_t spriteIndex)
+static int32_t ride_get_train_index_from_vehicle(Ride* ride, uint16_t spriteIndex)
 {
     uint32_t trainIndex = 0;
```
duncanspumpkin: change this type to `int32_t` as well

tupaschoal: I had, actually, but when comparing it to `std::size(ride->num_vehicles)` it errored out due to narrowing. I thought about doing `static_cast` but was unsure. What do you think?

duncanspumpkin: What about refactoring the whole thing to use a std::optional? It would then get rid of the super edge cases.

tupaschoal: Will try this tonight, if it's too big

tupaschoal: I might have misunderstood, did you mean change the `ride->vehicles` to be an optional?
I think it only errored on the `else if (trainIndex >= std::size(ride->vehicles))`, not sure about the `if (trainIndex >= ride->num_vehicles)`. Sorry for the confusion.

duncanspumpkin: I was meaning for the return of the function. It gets rid of the invalid train index.

duncanspumpkin:
I've updated this to use std::optional for the return.

PR #5979, opened 2017-07-23
Create new GameState class

IntelOrca:
This is the start of a new sequence of PRs which will gradually remove the global state used for the game state and replace them with classes which can be constructed as plain objects but updated with the required dependencies. As an example, `Park::Update` now takes a `Date` parameter so it does not need to access `gMonthTicks` and `gMonthsElapsed`.

This will then allow multiple game states to coexist and OpenRCT2 will be able to switch from one to another very quickly. The title sequence for example could pre-load the next park so the switch is seamless. Loading any game can return a new `GameState`.

PR #10180, opened 2019-11-01
Refactor screen state swapping

ZehMatt:
Some initial work towards having multiple screen states. This will allow to have at some point loading screens that will do all the things in the background and simply swap the screen state after it has loaded all the required things, its still pretty raw and I would like to get some feedback like if we want to go this route.


ZehMatt:
Also I was wondering if we maybe should call this 'Scene' instead of 'Screen', so we get TitleScene, GameScene, LoadingScene etc.

AaronVanGeffen:
This sound very similar to the (ambitious) goals the initial GameState PR #5979 set out with. We never got quite that far. I haven't looked at the code in detail yet, but good to see this resurface.

`Scene` sounds like a better name to me as well.

src/openrct2/Context.cpp
```diff
@@ -103,8 +104,12 @@ namespace OpenRCT2
         Networking::Http::Http _http;
 #endif
 
-        // Game states
+        // Screens.
         std::unique_ptr<TitleScreen> _titleScreen;
+        std::unique_ptr<GameScreen> _gameScreen;
+        IScreenState* _activeScreen = nullptr;
```
tupaschoal: What about making the `std::unique_ptr` into `shared` and having `_activeScreen` be a `weak_ptr`, for safety?

tupaschoal: Should these unique be changes into shared and the `_acriveScreen` be a weak?

tjwrona1992: Since you really only want one instance of each screen I would think `std::unique_ptr` would be the appropriate choice here. Everywhere the raw pointer is used it is checked to see if it is null so I don't think it can cause a problem. Ideally you would want to make `_activeScreen` a `std::observer_ptr` to emphasize that it is non-owning but I don't think observer pointers have been officially adopted to the standard yet so a raw pointer is probably the best you can really do for now.

src/openrct2/screens/game/GameScreen.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include "../../core/Console.hpp"
+#include "../../drawing/Drawing.h"
+#include "../../interface/Screenshot.h"
+#include "../../interface/Viewport.h"
+#include "../../interface/Window.h"
+#include "../../localisation/Localisation.h"
+#include "../../network/network.h"
+#include "../../scenario/Scenario.h"
+#include "../../scenario/ScenarioRepository.h"
+#include "../../ui/UiContext.h"
```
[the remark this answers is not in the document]
ZehMatt: It was a pretty quick draft, and no, not at this point. I want to eventually deduplicate some code that also the title screen uses because both are more or less just simulating the park with a few minor differences.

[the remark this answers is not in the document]
tjwrona1992: I agree, this does seem a bit odd if this is a new file.

src/openrct2/Context.cpp
ZehMatt: I would agree if it wouldn't be clear when the memory gets released, the context drives the lifetime, but yeah I'm not a fan of pointers. I wan't to cleanup the whole initialization at some point its a mess atm.

tupaschoal: Alright

tjwrona1992: Not a fan of pointers at all??? or just not a fan of raw pointers? It's hard to imagine being a C++ programmer and not liking pointers at all lol

ZehMatt: Yes don't use pointers if they can be avoided, and in this case they probably can be avoided by making them basic members, so when the context would be destroyed so is everything with it. If you have some function called Initialize in a class you are breaking the core concept of C++ that wants objects to be usable after construction.

tjwrona1992: I guess I can agree with that, although wouldn't any `unique_ptr` class members automatically be destroyed along with the class too?

ZehMatt: Yes they would be destroyed automatically but they are still dynamic allocations which in this case is not required at all if I would just have them as basic members, ex.: "GameScreen _gameScreen" and pass the GameState object during construction and then everything is nicely in-place in the Context class.

tjwrona1992: Ah I see what you mean. Sorry if I ask a lot of questions, its refreshing to talk to people who actually deeply understand C++ haha. I can finally get answers to some questions people I work with don't have clear answers to.

ZehMatt:
This needs a fresh start

PR #10191, opened 2019-11-03
Refactor benchgfx

ZehMatt:
The old benchgfx was a little bit odd and did not really tell anything other than how much time really elapsed, on top there was a pretty good variance in it.

A few things this PR does:
- It will generate now a screenshoft by zoom levels * rotations * iterations.
- Pre-initialize viewports and dpis before making screenshots, before they had been included in the time but initialization usually happens only once, also fixing a bug where the viewport zoom was incorrectly initialized resulting in always having a viewport at zero.
- The new default value for the iteration count is now defaulted to 20 to reflect the changes.
- Raises accuracy by using double instead of float.
- The variance is now gone because initialization can vary a bit due to memory allocation.
- More meaningful output which is also easier to parse if one must.
```
>openrct2 benchgfx parks/bpb.sv6 20
Engine: Software
Render Count: 240
Zoom[0] average: 0.058009s, 17 FPS
Zoom[1] average: 0.026519s, 38 FPS
Zoom[2] average: 0.012787s, 78 FPS
Total average: 0.032438s, 31 FPS
Time: 7.78519s
```
Having multiple runs now has at best a variance of 1 FPS which is now more than okay.

### DIFF
--- a/src/openrct2/interface/Screenshot.cpp
+++ b/src/openrct2/interface/Screenshot.cpp
@@ -320,6 +320,34 @@ static int32_t GetTallestVisibleTileTop(int32_t mapSize, int32_t rotation)
     return minViewY - 256;
 }
 
+static rct_drawpixelinfo CreateDPI(const rct_viewport& viewport)
+{
+    rct_drawpixelinfo dpi;
+    dpi.width = viewport.width;
+    dpi.height = viewport.height;
+    dpi.bits = new (std::nothrow) uint8_t[dpi.width * dpi.height];
+    if (dpi.bits == nullptr)
+    {
+        throw std::runtime_error("Giant screenshot failed, unable to allocate memory for image.");
+    }
+
+    if (viewport.flags & VIEWPORT_FLAG_TRANSPARENT_BACKGROUND)
+    {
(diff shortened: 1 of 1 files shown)


src/openrct2/interface/Screenshot.cpp
```diff
@@ -320,6 +321,34 @@ static int32_t GetTallestVisibleTileTop(int32_t mapSize, int32_t rotation)
...
+static rct_drawpixelinfo CreateDPI(const rct_viewport& viewport)
+{
+    rct_drawpixelinfo dpi;
+    dpi.width = viewport.width;
+    dpi.height = viewport.height;
+    dpi.bits = new uint8_t[dpi.width * dpi.height]; //(uint8_t*)malloc((size_t)dpi.width * dpi.height);
+    if (dpi.bits == nullptr)
+    {
+        throw std::runtime_error("Giant screenshot failed, unable to allocate memory for image.");
+    }
```
tupaschoal: It makes sense to check the pointer if you're using ```malloc```, but if you're using ```new```, then you have to do a ```try...catch(std::bad_alloc& e)```, unless you use the `nothrow` version of `new`.

PR #10198, opened 2019-11-06
ifdef-out WinNT 6.0+ APIs

janisozaur:
Occasionally I need to build a binary compatible with WinNT 5.1, e.g.
when testing out Wine or ReactOS. There are just a few minor changes
required to have a working build, but as they are not part of the
repository, I always end up doing them from scratch. I would like to
upstream them instead so I don't have to recreate them each time, even
if this is not the most common use case.

### DIFF
--- a/src/openrct2-ui/UiContext.Win32.cpp
+++ b/src/openrct2-ui/UiContext.Win32.cpp
@@ -38,7 +38,7 @@
 static std::wstring SHGetPathFromIDListLongPath(LPCITEMIDLIST pidl)
 {
     std::wstring pszPath(MAX_PATH, 0);
-    while (!SHGetPathFromIDListEx(pidl, &pszPath[0], (DWORD)pszPath.size(), 0))
+    while (!SHGetPathFromIDListW(pidl, &pszPath[0]))
     {
         if (pszPath.size() >= SHRT_MAX)
         {
--- a/src/openrct2/core/String.cpp
+++ b/src/openrct2/core/String.cpp
@@ -706,6 +706,7 @@ namespace String
     std::string ToUpper(const std::string_view& src)
     {
 #ifdef _WIN32
+#    if _WIN32_WINNT >= 0x0600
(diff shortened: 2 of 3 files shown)


src/openrct2-ui/UiContext.Win32.cpp
```diff
@@ -38,7 +38,7 @@
 static std::wstring SHGetPathFromIDListLongPath(LPCITEMIDLIST pidl)
 {
     std::wstring pszPath(MAX_PATH, 0);
-    while (!SHGetPathFromIDListEx(pidl, &pszPath[0], (DWORD)pszPath.size(), 0))
+    while (!SHGetPathFromIDListW(pidl, &pszPath[0]))
```
tupaschoal: I'm assuming this changes it to a more backward compatible call (?)

src/openrct2/core/String.cpp
```diff
@@ -706,6 +706,7 @@ namespace String
     std::string ToUpper(const std::string_view& src)
     {
 #ifdef _WIN32
+#    if _WIN32_WINNT >= 0x0600
```
tupaschoal: Is it a standard in the whole project to have macros with the content indented but not the `#`? I see it in the includes and though I know it doesn't make a difference, it still bugs me

IntelOrca: I think it bugs quite a few people. There needs to be a change to our clang-format config and then a bulk format edit to the code base though.

src/openrct2-ui/UiContext.Win32.cpp
janisozaur: Yes, the extended version only makes sense if you use the non-standard parameters. In our case, it was using the standard ones

janisozaur:
@tupaschoal until such feature gets implemented, I'm going to merge this as-is to address the current needs.

tupaschoal:
Sure, I didn't mean for you to wait, just speaking out loud on things I could do in the future

PR #10213, opened 2019-11-10
src/openrct2/config/Config.cpp: Search additional pathes for rct data.

ConiKost:
With '/usr/share/openrct2/rct{1,2}_data' added into the search path list,
it makes possible to install rct{1,2} game files globally,
so every user on a linux system can run OpenRCT2 directly.

### DIFF
--- a/src/openrct2/config/Config.cpp
+++ b/src/openrct2/config/Config.cpp
@@ -615,6 +615,7 @@ namespace Config
         log_verbose("config_find_rct1_path(...)");
 
         static constexpr const utf8* searchLocations[] = {
+            R"(/usr/share/openrct2/rct1_data)",
             R"(C:\Program Files\Steam\steamapps\common\Rollercoaster Tycoon Deluxe)",
             R"(C:\Program Files (x86)\Steam\steamapps\common\Rollercoaster Tycoon Deluxe)",
             R"(C:\GOG Games\RollerCoaster Tycoon Deluxe)",
@@ -659,6 +660,7 @@ namespace Config
         log_verbose("config_find_rct2_path(...)");
 
         static constexpr const utf8* searchLocations[] = {
+            R"(/usr/share/openrct2/rct2_data)",
             R"(C:\Program Files\Steam\steamapps\common\Rollercoaster Tycoon 2)",
             R"(C:\Program Files (x86)\Steam\steamapps\common\Rollercoaster Tycoon 2)",
             R"(C:\GOG Games\RollerCoaster Tycoon 2 Triple Thrill Pack)",


src/openrct2/config/Config.cpp
```diff
@@ -615,6 +615,7 @@ namespace Config
         log_verbose("config_find_rct1_path(...)");
 
         static constexpr const utf8* searchLocations[] = {
+            R"(/usr/share/openrct2)",
```
janisozaur: I'm pretty sure this is not correct. These are locations for `RollerCoaster Tycoon Deluxe`, part 1 of the series.

src/openrct2/config/Config.cpp
```diff
@@ -659,6 +660,7 @@ namespace Config
         log_verbose("config_find_rct2_path(...)");
 
         static constexpr const utf8* searchLocations[] = {
+            R"(/usr/share/openrct2)",
```
janisozaur: This one is a bit odd. While I somewhat understand the problem you're trying to solve, I'm not convinced this is the correct way. These paths are locations for known default places where RCT2 would be installed. `/usr/share/openrct2` is not one such location.

There should be a more platform-specific function already you should be able to use and then keep in mind the distinction between `OpenRCT2`, this project, and RCT2, the original game we need for its assets.

src/openrct2/config/Config.cpp
ConiKost: A question: Is is possible to install both data files, from RCT1 and RCT2 at the same time?
Or only one at a time?

src/openrct2/config/Config.cpp
ConiKost: I am open for suggestions :-)
I agree, that my path is not an original installation path.
Just didn't found a better way to patch that.

Where should I instead add the path?
Sounds like you have something in mind?

src/openrct2/config/Config.cpp
Gymnasiast: They serve different purposes. The RCT2 files are needed for the game, while the RCT1 files are optional and provide some extra graphics. So a user will either have RCT2 installed, or RCT2 _and_ RCT1.

ConiKost: Ah! Allright, that clarifies it for me.

src/openrct2/config/Config.cpp
ConiKost: Looking further into the code, I don't see any other places, where this could be definied. This seems to be exactly the place, where it just works, as intended.

I just updated for two paths, since you are right, that RCT1 and RCT2 data is not the same. 
But the functions seems to be exactly to do, in order to achieve that?

When I place the needed folders from RCT1 and RCT2 in both directories, applied with my patch, OpenRCT2 seems to work perfectly fine. It detects the stuff correctly and paths for `config.ini` are mapped correctly automatically.

About the subfolders, the names can be of course something different, maybe more `rct1_data` and `rct2_data` or something.

ConiKost: @janisozaur @Gymnasiast ping

Gymnasiast:
I plan to look at this PR at some point. I haven't done it yet because I need to check some stuff. (Just know I have not forgotten about this PR.)

ConiKost:
Thanks :-)

Gymnasiast:
I finally got round to doing some checks and discussing some stuff. Sorry for the long wait.

The feature you are extending is meant to check common paths for RCT2 installs on Windows system. On Linux, lacking native versions of RCT2, this isn't really applicable, everybody uses a different directory. It means that only a very small group of people will benefit, namely people that are on multi-user systems and that are aware of this special location. After all, for people who are the only user of their system (which I'd say is the most common scenario), it won't really add anything, as they will only have one OpenRCT2 install anyway.

Personally, I think that using `locate` or `find` would be more beneficial, as though would _significantly_ increase the chances of OpenRCT2 automatically picking up the RCT2 assets.

AaronVanGeffen:
> Personally, I think that using `locate` or `find` would be more beneficial, as though would _significantly_ increase the chances of OpenRCT2 automatically picking up the RCT2 assets.

`locate` could work, but you cannot assume its daemon is running. (It's not on my pc, for example.) 

Using `find` is no different from implementing recursively traversing directories ourselves.

ConiKost:
> The feature you are extending is meant to check common paths for RCT2 installs on Windows system.

While is may be designed in mind for windows versions, its basically a generic function to search through definied directories for RCT* data.

> It means that only a very small group of people will benefit, namely people that are on multi-user systems and that are aware of this special location.

Well, that depends. The idea behind this is, if a linux distribution wan'ts to include OpenRCT2, it offers the possibility to use those data files for all users. I don't think, that this is only a small group. As speaking for Gentoo, after installing OpenRCT2, the user gets an info, where he can choose, if he wan't to install globally or just for his just locally. I am more looking as a linux distribution developer on that scenario and less in terms of a user.

> After all, for people who are the only user of their system (which I'd say is the most common scenario), it won't really add anything, as they will only have one OpenRCT2 install anyway.

Sure, but on the other side, it won't harm too 👍
Btw, the path `/usr/share/openrct2` is pretty generic and according to FHS, it would be the correct purpose for using it globally accros various linux distributions.

If you really don't want to add such change, I have another proposal: How about making that a compile option? That means, I can pass via configure options a path, which `can` be used by OpenRCT2 to find the data globally. And If as option nothing is passed, it stays, as it does currently and does not search for any data globally. So it's at the end up to the distro dev, what to do.

> Personally, I think that using locate or find would be more beneficial, as though would significantly increase the chances of OpenRCT2 automatically picking up the RCT2 assets.

No, please don't do that. That doesn't sound right, to run external tools in order to find something. Also you would have so search through whole `/` and locate isn't guaranteed to be available on all systems.

AaronVanGeffen:
> Well, that depends. The idea behind this is, if a linux distribution wan'ts to include OpenRCT2, it offers the possibility to use those data files for all users. I don't think, that this is only a small group. As speaking for Gentoo, after installing OpenRCT2, the user gets an info, where he can choose, if he wan't to install globally or just for his just locally. I am more looking as a linux distribution developer on that scenario and less in terms of a user.

For the OpenRCT2 files, this is already possible. For the RCT{1,2} files, this is not yet the case. The scenario you lay out is unlikely to ever apply, though, as the RCT{1,2} files are inherently proprietary. A Linux distribution shipping them is near inconceivable, to me.

IntelOrca:
It makes more sense to include the default paths of Steam, but since users have to manually extract the files in someway on Linux - they are very likely to know where they have placed the files and can easily choose that location using the folder picker when the game launches for the first time.

Alternatively there is a command line switch to set the path which can be done on installation:
`openrct2 set-rct2 /path/to/rct2`

Gymnasiast:
> It makes more sense to include the default paths of Steam, but since users have to manually extract the files in someway on Linux - they are very likely to know where they have placed the files and can easily choose that location using the folder picker when the game launches for the first time.

It already does that, assuming the user has used the Steam console to download the files.

ConiKost:
> Any Linux distribution shipping them is near inconceivable, to me.

Yes, I wasn't talking about shipping those files but just providing a comfort feature for installing for all users.

> Alternatively there is a command line switch to set the path which can be done on installation:
openrct2 set-rct2 /path/to/rct2

What about rct1 data?

IntelOrca:
> What about rct1 data?

That needs implementing.

Gymnasiast:
OK, I have given it some thought.

I'm okay with defining the two locations to search in, but I would like to see it coupled with a refactor. It should only check the `C:\` paths on Windows, and it should only check the `/usr/share` paths on Linux/BSD. For that, you'd need to specify `searchLocations` in the platform-specific files. Do you think you can pull that off?

ConiKost:
I will try, if I can do something.

tupaschoal:
I'll close this as it's been a while and it hasn't been resumed. Feel free to reopen if you do or to file an issue tracking what you had in mind.

PR #10220, opened 2019-11-12
Use ScreenCoordsXY for windows/Footpath

tupaschoal:


src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -717,10 +717,10 @@ static void window_footpath_set_provisional_path_at_point(int32_t x, int32_t y)
     TileElement* tileElement{};
     LocationXY16 mapCoord = {};
     get_map_coordinates_from_pos(
-        x, y, VIEWPORT_INTERACTION_MASK_FOOTPATH & VIEWPORT_INTERACTION_MASK_TERRAIN, &mapCoord.x, &mapCoord.y,
-        &interactionType, &tileElement, nullptr);
-    x = mapCoord.x;
-    y = mapCoord.y;
+        screenCoords.x, screenCoords.y, VIEWPORT_INTERACTION_MASK_FOOTPATH & VIEWPORT_INTERACTION_MASK_TERRAIN, &mapCoord.x,
+        &mapCoord.y, &interactionType, &tileElement, nullptr);
+    screenCoords.x = mapCoord.x;
```
duncanspumpkin: No don't do this. Screen cords and map cords should never mix.

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -853,10 +853,10 @@ static void window_footpath_place_path_at_point(int32_t x, int32_t y)
...
+    screenCoords.y = mapCoord.y;
```
duncanspumpkin: As above don't mix the coordinates. Use mapCoord for everything below

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -888,7 +888,7 @@ static void window_footpath_place_path_at_point(int32_t x, int32_t y)
 
     // Try and place path
     gGameCommandErrorTitle = STR_CANT_BUILD_FOOTPATH_HERE;
-    auto footpathPlaceAction = FootpathPlaceAction({ x, y, z * 8 }, currentType, selectedType);
+    auto footpathPlaceAction = FootpathPlaceAction({ screenCoords.x, screenCoords.y, z * 8 }, currentType, selectedType);
```
duncanspumpkin: No game actions take screen cords. Any time you see something like this it is obvious something is wrong

src/openrct2-ui/windows/Footpath.cpp
tupaschoal: What do you suggest? I mean because the assignment was already there, it just became more explicit.

src/openrct2-ui/windows/Footpath.cpp
tupaschoal: What to do? Build a `CoordsXY` or `TileCoordsXY`? I mean, they'd still be disguised screen coordinates.

src/openrct2-ui/windows/Footpath.cpp
duncanspumpkin: Is the assignment actually needed? Can't you just use mapcoord directly. The aim of this refactoring is to clearly indicate what is screen and what is map.

src/openrct2-ui/windows/Footpath.cpp
duncanspumpkin: Well it's the same issue as above remove the assignments and use the map cord parameters directly.

src/openrct2-ui/windows/Footpath.cpp
tupaschoal: I don't think it is strictly needed, I'll update it and we'll see.

tupaschoal:
I solved all the issues and also reverted to do the `+= 16` in place, as some of them did reuse the value below.

PR #10171, opened 2019-10-31
Make Viewport::screen_to_map_pos() return coords

tupaschoal:
As proposed by @duncanspumpkin on #10131

> This is horrible. We should refactor this.
>
> mapX and mapY are CoordsXY values but you need to set them to screenCoords.x and screenCoords.y if clickedElement is nullptr below as screen_pos_to_map_pos is a horrible function that takes in screen coords and spits out coordsxy to the same variable.

Notes to future:
- Make `screen_get_map_xy_side` and `screen_get_map_xy` return `CoordsXY`.
- Make `screen_pos_to_map_pos`, `screen_get_map_xy_side` receive `ScreenCoordsXY`

PR #10221, opened 2019-11-12
Receive ScreenCoordsXY and return CoordsXY in Viewport

tupaschoal:
Following on my own comments in #10171 

> Notes to future:
> Make `screen_get_map_xy_side` and `screen_get_map_xy` return `CoordsXY`.
> Make `screen_pos_to_map_pos`, `screen_get_map_xy_side` receive `ScreenCoordsXY`

### DIFF
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -1222,8 +1222,8 @@ static void window_tile_inspector_tool_update(rct_window* w, rct_widgetindex wid
 
     gMapSelectFlags |= MAP_SELECT_FLAG_ENABLE;
 
-    int16_t mapX = screenCoords.x;
-    int16_t mapY = screenCoords.y;
+    int16_t mapX = 0;
+    int16_t mapY = 0;
     TileElement* clickedElement = nullptr;
     if (input_test_place_object_modifier(PLACE_OBJECT_MODIFIER_COPY_Z))
     {
@@ -1234,7 +1234,7 @@ static void window_tile_inspector_tool_update(rct_window* w, rct_widgetindex wid
     // Even if Ctrl was pressed, fall back to normal selection when there was nothing under the cursor
     if (clickedElement == nullptr)
     {
-        CoordsXY mapCoords = screen_pos_to_map_pos({ mapX, mapY }, nullptr);
+        CoordsXY mapCoords = screen_pos_to_map_pos(screenCoords, nullptr);
         mapX = mapCoords.x;
         mapY = mapCoords.y;
     }
@@ -1273,8 +1273,8 @@ static void window_tile_inspector_update_selected_tile(rct_window* w, int32_t x,
     windowTileInspectorToolMouseY = y;
     windowTileInspectorToolCtrlDown = ctrlIsHeldDown;
 
-    int16_t mapX = x;
-    int16_t mapY = y;
+    int16_t mapX = 0;
+    int16_t mapY = 0;
     TileElement* clickedElement = nullptr;
     if (ctrlIsHeldDown)
     {
@@ -1284,7 +1284,7 @@ static void window_tile_inspector_update_selected_tile(rct_window* w, int32_t x,
     // Even if Ctrl was pressed, fall back to normal selection when there was nothing under the cursor
     if (clickedElement == nullptr)
     {
-        CoordsXY mapCoords = screen_pos_to_map_pos({ mapX, mapY }, nullptr);
+        CoordsXY mapCoords = screen_pos_to_map_pos({ x, y }, nullptr);
 
         if (mapCoords.x == LOCATION_NULL)
         {
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -1234,7 +1234,7 @@ static void window_tile_inspector_tool_update(rct_window* w, rct_widgetindex wid
     // Even if Ctrl was pressed, fall back to normal selection when there was nothing under the cursor
     if (clickedElement == nullptr)
     {
-        CoordsXY mapCoords = screen_pos_to_map_pos(mapX, mapY, nullptr);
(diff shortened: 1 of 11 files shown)


src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -2080,17 +2083,17 @@ static void top_toolbar_tool_update_land(int16_t x, int16_t y)
...
+    CoordsXY mapTile;
     uint8_t side;
 
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE;
     if (tool_size == 1)
     {
         int32_t selectionType;
         // Get selection type and map coordinates from mouse x,y position
-        CoordsXY mapCoords = screen_pos_to_map_pos(x, y, &selectionType);
+        CoordsXY mapCoords = screen_pos_to_map_pos({ x, y }, &selectionType);
```
duncanspumpkin: mapCoords is never actually used. This function doesn't need to save the result it is only been used to get selectionType.

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -2080,17 +2083,17 @@ static void top_toolbar_tool_update_land(int16_t x, int16_t y)
...
         mapTile = { static_cast<int16_t>(mapCoords.x), static_cast<int16_t>(mapCoords.y) };
```
duncanspumpkin: This doesn't do anything. Remove

src/openrct2/interface/Viewport.cpp
```diff
@@ -1047,11 +1043,11 @@ CoordsXY screen_pos_to_map_pos(int16_t x, int16_t y, int32_t* direction)
     return mapCoords;
 }
 
-LocationXY16 screen_coord_to_viewport_coord(rct_viewport* viewport, uint16_t x, uint16_t y)
+LocationXY16 screen_coord_to_viewport_coord(rct_viewport* viewport, ScreenCoordsXY screenCoords)
```
duncanspumpkin: A viewport coord is also a screen coord just adjusted for zoom. So in a future refactor this could be refactored to return a ScreenCoord struct.

src/openrct2/interface/Window.cpp
```diff
@@ -916,7 +916,9 @@ void window_rotate_camera(rct_window* w, int32_t direction)
 
     // has something to do with checking if middle of the viewport is obstructed
     rct_viewport* other;
-    screen_get_map_xy(x, y, &x, &y, &other);
+    CoordsXY coords = screen_get_map_xy({ x, y }, &other);
+    x = coords.x;
```
duncanspumpkin: We should just use `coords` in all of the below calcs instead of reusing x and y

src/openrct2/world/MoneyEffect.cpp
```diff
@@ -94,9 +94,10 @@ void rct_money_effect::Create(money32 value)
             return;
 
         rct_viewport* mainViewport = window_get_viewport(mainWindow);
-        screen_get_map_xy(
-            mainViewport->x + (mainViewport->width / 2), mainViewport->y + (mainViewport->height / 2), &mapPosition.x,
-            &mapPosition.y, nullptr);
+        CoordsXY mapPositionXY = screen_get_map_xy(
+            { mainViewport->x + (mainViewport->width / 2), mainViewport->y + (mainViewport->height / 2) }, nullptr);
+        mapPosition.x = mapPositionXY.x;
```
duncanspumpkin: mapPosition should be changed to a CoordsXYZ instead of a LocationXYZ16. This will simplify some of this.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1234,7 +1234,7 @@ static void window_tile_inspector_tool_update(rct_window* w, rct_widgetindex wid
     // Even if Ctrl was pressed, fall back to normal selection when there was nothing under the cursor
     if (clickedElement == nullptr)
```
duncanspumpkin: Okay I've been looking into this a bit more. This is still really messy but due to another bug we never actually encounter the bug that would cause screencoords to get clobered with map coords. Thats not a reason to fix the issue though. This function should be taking `screenCoords` not `{mapX, mapY}`. Then we need to get rid of the assignment in L1225-1226. That will then mean we can simplify it so that mapX and mapY mean CoordsXY. I'll fix the bug in `get_map_coordinates_from_pos` myself.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1284,7 +1284,7 @@ static void window_tile_inspector_update_selected_tile(rct_window* w, int32_t x,
...
     {
-        CoordsXY mapCoords = screen_pos_to_map_pos(mapX, mapY, nullptr);
+        CoordsXY mapCoords = screen_pos_to_map_pos({ mapX, mapY }, nullptr);
```
duncanspumpkin: same issue as above make the same changes but with x,y instead of screenCoords.

src/openrct2/interface/Viewport.cpp
tupaschoal: I saved it to do it next.

src/openrct2-ui/windows/TileInspector.cpp
tupaschoal: Ok, if you want me to do it, let me know

tupaschoal: I couldn't get away with the whole assignment, as the variables need to be created outside the `if` scope. I did assign just `0`, which I think is what you meant.

I thought about changing it all to `CoordsXY`, but the `get_map_coordinates_from_pos` requires an `int16_t`. I'll work on making sure it is refactored as soon as this is merged.

tupaschoal:
I changed based on the latest comments, let me know if it's alright now, I had to made some assumptions that I can't tell if they're correct.

duncanspumpkin:
You'll need to rebase due to me merging that other pr

tupaschoal:
Rebase done @duncanspumpkin

PR #10222, opened 2019-11-12
Use ScreenCoordsXY for MazeConstruction

tupaschoal:


src/openrct2-ui/windows/MazeConstruction.cpp
```diff
@@ -363,15 +363,16 @@ static void window_maze_construction_entrance_tooldown(int32_t x, int32_t y, rct
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE_ARROW;
 
     int32_t direction = 0;
-    ride_get_entrance_or_exit_position_from_screen_position(x, y, &x, &y, &direction);
```
Gymnasiast: This doesn't look correct. The output parameters are map coordinates and not ScreenCoords. You'll need to add an extra TileCoordsXY or CoordsXY (depending on what `ride_get_entrance_or_exit_position_from_screen_position()` outputs).

tupaschoal: This is a fun story, while doing this changeset, I even thought about updating `ride_get_entrance_or_exit_position_from_screen_position` to return a coords `struct`, but I didn't know which to use, because of these places that are mixing them up.

I can save the output coords to a new struct, the thing is that sometimes `ride_get_entrance_or_exit_position_from_screen_position` early exits, changing the value of `x`, but not `y`, so this behavior would be kind of different (but dunno if it would impact something).

duncanspumpkin: Yes save the output into a CoordsXY and change any below uses of the value to use CoordsXY var. There should be a check on the return value to make sure the value is not `0x8000` (LOCATION_NULL)

src/openrct2-ui/windows/MazeConstruction.cpp
```diff
@@ -363,16 +363,19 @@ static void window_maze_construction_entrance_tooldown(ScreenCoordsXY screenCoor
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE_ARROW;
 
     int32_t direction = 0;
+    CoordsXY entranceOrExitCoords;
     ride_get_entrance_or_exit_position_from_screen_position(
-        screenCoords.x, screenCoords.y, &screenCoords.x, &screenCoords.y, &direction);
+        screenCoords.x, screenCoords.y, &entranceOrExitCoords.x, &entranceOrExitCoords.y, &direction);
+    if (entranceOrExitCoords.x == 0x8000)
```
duncanspumpkin: Use the location_null macro. Could also fix ride get entrance or exit position to use the macro at the same time

tupaschoal: I am using the macro already, note that I did two commits, you probably saw an outdated version

duncanspumpkin: How odd. Must have been some cache issue

PR #10223, opened 2019-11-12
Remove global variables from get_map_coordinates_from_pos_window

duncanspumpkin:
Noticed that there was an easy way to nuke a good few globals. Note this will probably cause a few bugs to change from subtly failing to null dereferencing.

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -55,8 +55,6 @@ uint8_t gCurrentRotation;
 
 static uint32_t _currentImageType;
 
-static rct_drawpixelinfo _viewportDpi1;
-static rct_drawpixelinfo _viewportDpi2;
 struct InteractionInfo
 {
     InteractionInfo() = default;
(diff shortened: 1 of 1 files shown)


src/openrct2/interface/Viewport.cpp
```diff
@@ -58,10 +57,12 @@ static uint32_t _currentImageType;
...
 static rct_drawpixelinfo _viewportDpi1;
 static rct_drawpixelinfo _viewportDpi2;
-static uint8_t _interactionSpriteType;
-static int16_t _interactionMapX;
-static int16_t _interactionMapY;
-static uint16_t _interactionFlags;
+struct InteractionInfo
+{
+    CoordsXY loc;
+    TileElement* element;
```
tupaschoal:
```suggestion
    TileElement* element { nullptr };
```
Initialize it, to keep the old behavior

src/openrct2/interface/Viewport.cpp
```diff
@@ -1260,12 +1261,12 @@ void viewport_set_visibility(uint8_t mode)
...
  *  rct2: 0x00688697
  */
-static void store_interaction_info(paint_struct* ps)
+static std::optional<InteractionInfo> GetInteractionInfoFromPS(paint_struct* ps, uint16_t flags)
 {
     if (ps->sprite_type == VIEWPORT_INTERACTION_ITEM_NONE
         || ps->sprite_type == 11 // 11 as a type seems to not exist, maybe part of the typo mentioned later on.
         || ps->sprite_type > VIEWPORT_INTERACTION_ITEM_BANNER)
-        return;
+        return {};
```
tupaschoal: Thoughts on returning this instead?
```suggestion
        return std::nullopt;
```

duncanspumpkin:
Refactor to return bool and use constructor.

PR #10227, opened 2019-11-13
Fix desync because of ride status

ZehMatt:
ride_set_status requires a round trip, clients without permission to set the ride status would get an error whenever a vehicle crashes. Client and Server now use the nested variant during simulation, we expect it to be deterministic.


duncanspumpkin:
There should really be a comment saying why this is done as this isn't really what you would expect

ZehMatt:
Ref #10249

PR #10229, opened 2019-11-13
Refactor ride_get_entrance_or_exit_position_from_screen_position()

tupaschoal:
Some considerations:
- I narrowed `direction` in many of its usages to favor the `Direction` typedef, but it seems alright, as the larger ones used `Direction` operations anyway.
- I removed the global `_unkF44188`, maybe there are others to remove soon

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -3688,7 +3688,6 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
  */
 void ride_construction_toolupdate_entrance_exit(ScreenCoordsXY screenCoords)
 {
-    int32_t x, y, direction;
     uint8_t stationNum;
 
     map_invalidate_selection_rect();
@@ -3696,7 +3695,7 @@ void ride_construction_toolupdate_entrance_exit(ScreenCoordsXY screenCoords)
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE;
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE_CONSTRUCT;
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE_ARROW;
-    ride_get_entrance_or_exit_position_from_screen_position(screenCoords.x, screenCoords.y, &x, &y, &direction);
+    CoordsXYZD entranceOrExitCoords = ride_get_entrance_or_exit_position_from_screen_position(screenCoords);
     if (gRideEntranceExitPlaceDirection == 255)
     {
         ride_construction_invalidate_current_track();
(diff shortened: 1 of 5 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3705,26 +3704,30 @@ void ride_construction_toolupdate_entrance_exit(ScreenCoordsXY screenCoords)
...
+    gMapSelectArrowPosition.z = entranceOrExitCoords.z * 8;
     map_invalidate_selection_rect();
 
-    direction = direction_reverse(gRideEntranceExitPlaceDirection);
+    entranceOrExitCoords.direction = direction_reverse(gRideEntranceExitPlaceDirection);
     stationNum = gRideEntranceExitPlaceStationIndex;
-    if (!(_currentTrackSelectionFlags & TRACK_SELECTION_FLAG_ENTRANCE_OR_EXIT) || x != gRideEntranceExitGhostPosition.x
-        || y != gRideEntranceExitGhostPosition.y || direction != gRideEntranceExitGhostPosition.direction
+    if (!(_currentTrackSelectionFlags & TRACK_SELECTION_FLAG_ENTRANCE_OR_EXIT)
+        || entranceOrExitCoords.x != gRideEntranceExitGhostPosition.x
```
Gymnasiast: Doesn't CoordsXYZD have a definition for the `==` method, so that you can do `entranceOrExitCoords == gRideEntranceExitGhostPosition`?

tupaschoal: It does not. The only struct on `Location.hpp` that defines the `operator==` is `TyleCoordsXYZ`

src/openrct2/ride/Ride.cpp
```diff
@@ -6352,14 +6350,18 @@ void ride_get_entrance_or_exit_position_from_screen_position(
             entranceMaxY = mapY;
         }
 
-        direction = loc_6CD18E(*outX, *outY, entranceMinX - 32, entranceMinY - 32, entranceMaxX + 32, entranceMaxY + 32);
-        if (direction != -1 && direction != stationDirection && direction != direction_reverse(stationDirection))
+        auto direction = loc_6CD18E(
+            entranceExitCoords.x, entranceExitCoords.y, entranceMinX - 32, entranceMinY - 32, entranceMaxX + 32,
+            entranceMaxY + 32);
+        if (entranceExitCoords.direction != -1 && entranceExitCoords.direction != stationDirection
```
tupaschoal:
```suggestion
        if (direction != -1 && entranceExitCoords.direction != stationDirection
```

src/openrct2/ride/Ride.cpp
```diff
@@ -6352,14 +6350,18 @@ void ride_get_entrance_or_exit_position_from_screen_position(
             entranceMaxY = mapY;
         }
 
-        direction = loc_6CD18E(*outX, *outY, entranceMinX - 32, entranceMinY - 32, entranceMaxX + 32, entranceMaxY + 32);
-        if (direction != -1 && direction != stationDirection && direction != direction_reverse(stationDirection))
+        auto direction = loc_6CD18E(
+            entranceExitCoords.x, entranceExitCoords.y, entranceMinX - 32, entranceMinY - 32, entranceMaxX + 32,
+            entranceMaxY + 32);
+        if (direction != -1 && entranceExitCoords.direction != stationDirection
```
tupaschoal:
```suggestion
        if (direction != -1 && direction != stationDirection
```

src/openrct2/ride/Ride.cpp
```diff
@@ -6352,14 +6350,18 @@ void ride_get_entrance_or_exit_position_from_screen_position(
...
+            && entranceExitCoords.direction != direction_reverse(stationDirection))
```
tupaschoal:
```suggestion
            && direction != direction_reverse(stationDirection))
```

src/openrct2-ui/windows/RideConstruction.cpp
Gymnasiast: Then perhaps it's a good idea to introduce it.

tupaschoal: Will do

src/openrct2/ride/Ride.cpp
```diff
@@ -6208,50 +6207,48 @@ void ride_get_entrance_or_exit_position_from_screen_position(
 
     ride = get_ride(gRideEntranceExitPlaceRideIndex);
     if (ride == nullptr)
-        return;
+        return entranceExitCoords;
```
duncanspumpkin: Set .x to be `LOCATION_NULL`

src/openrct2/ride/Ride.cpp
```diff
@@ -6208,50 +6207,48 @@ void ride_get_entrance_or_exit_position_from_screen_position(
...
     }
 
     word_F4418C = mapX;
     word_F4418E = mapY;
-    _unkF44188.x = floor2(mapX, 32);
-    _unkF44188.y = floor2(mapY, 32);
-    *outX = _unkF44188.x;
-    *outY = _unkF44188.y;
+
+    entranceExitCoords = { floor2(mapX, 32), floor2(mapY, 32), entranceExitCoords.z, INVALID_DIRECTION };
```
duncanspumpkin: set z to 0

src/openrct2/ride/Ride.cpp
```diff
@@ -6208,50 +6207,48 @@ void ride_get_entrance_or_exit_position_from_screen_position(
...
-        return;
+        return entranceExitCoords;
 
     LocationXY8 stationStart = ride->stations[gRideEntranceExitPlaceStationIndex].Start;
     if (stationStart.xy == RCT_XY8_UNDEFINED)
-        return;
+        return entranceExitCoords;
 
-    _unkF44188.z = stationHeight;
+    entranceExitCoords.z = stationHeight;
```
duncanspumpkin: remove move to above

src/openrct2/ride/Ride.cpp
```diff
@@ -6208,50 +6207,48 @@ void ride_get_entrance_or_exit_position_from_screen_position(
...
-    _unkF44188.x = floor2(mapX, 32);
-    _unkF44188.y = floor2(mapY, 32);
-    *outX = _unkF44188.x;
-    *outY = _unkF44188.y;
+
+    entranceExitCoords = { floor2(mapX, 32), floor2(mapY, 32), entranceExitCoords.z, INVALID_DIRECTION };
 
     if (ride->type == RIDE_TYPE_NULL)
-        return;
+        return entranceExitCoords;
```
duncanspumpkin: For these returns that look like failures I think it should really be setting .x to `LOCATION_NULL`. Not 100% sure on that. Need to investigate it further.

src/openrct2/ride/Ride.cpp
tupaschoal: I'll assume you mean remove this as it'll already be initialized to `stationHeight` above

src/openrct2/ride/Ride.cpp
tupaschoal: I can set them to `LOCATION_NULL`, no hassle.

tupaschoal:
I've applied the fixes suggested by @Gymnasiast and @duncanspumpkin 

Note that in my latest commit I narrowed `stationHeight` to `uint_8t`. Every time it read or was assigned to, was also to `uint_8t`, so doing this made sense and allowed construction of `TileCoordsXYZD` without getting a narrowing warning.

src/openrct2/ride/Ride.cpp
```diff
@@ -6221,16 +6225,20 @@ CoordsXYZD ride_get_entrance_or_exit_position_from_screen_position(ScreenCoordsX
     word_F4418C = mapX;
     word_F4418E = mapY;
 
-    entranceExitCoords = { floor2(mapX, 32), floor2(mapY, 32), entranceExitCoords.z, INVALID_DIRECTION };
+    entranceExitCoords = { floor2(mapX, 32), floor2(mapY, 32), entranceExitCoords.z, stationHeight };
```
duncanspumpkin: Didn't realise station height was a type of direction 😜

tupaschoal: Wow, my bad, thanks for the catch! I've pushed the fix now

PR #10231, opened 2019-11-13
Refactor get_map_coordinates to enforce Coords struct

duncanspumpkin:
Little bit of tidying up. Removed the overload on `get_map_coordinates_from_pos_window` refactored all uses to use the Coords structs. Added an overload to `get_map_coordinates_from_pos` to use Coords struct. Will slowly move code to use that struct.

### DIFF
--- a/src/openrct2-ui/windows/TitleCommandEditor.cpp
+++ b/src/openrct2-ui/windows/TitleCommandEditor.cpp
@@ -122,7 +122,7 @@ static void window_title_command_editor_textinput(rct_window * w, rct_widgetinde
 static void scenario_select_callback(const utf8 * path);
 static int32_t get_command_info_index(int32_t index);
 static TITLE_COMMAND_ORDER get_command_info(int32_t index);
-static LocationXY16 get_location();
+static TileCoordsXY get_location();
 static uint8_t get_zoom();
 
 static rct_window_event_list window_title_command_editor_events = {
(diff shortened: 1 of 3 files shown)


src/openrct2-ui/windows/TitleCommandEditor.cpp
```diff
@@ -187,26 +187,25 @@ static TITLE_COMMAND_ORDER get_command_info(int32_t index)
...
+        CoordsXY mapCoord;
         get_map_coordinates_from_pos_window(
-            w, w->viewport->view_width / 2, w->viewport->view_height / 2, VIEWPORT_INTERACTION_MASK_TERRAIN, &mapCoord.x,
-            &mapCoord.y, &interactionType, &tileElement, nullptr);
+            w, { w->viewport->view_width / 2, w->viewport->view_height / 2 }, VIEWPORT_INTERACTION_MASK_TERRAIN,
+            mapCoord, &interactionType, &tileElement, nullptr);
         mapCoord.x -= 16;
-        mapCoord.x /= 32;
         mapCoord.y -= 16;
-        mapCoord.y /= 32;
```
tupaschoal: Why were these `/= 32` dropped?

duncanspumpkin: Because it converts the coordsxy into a tilecoordsxy. I've moved that into the constructor

tupaschoal: Ah, ok, I hadn't noticed `TileCoords` did that internally. Thanks!

src/openrct2/interface/Viewport.cpp
```diff
@@ -1619,21 +1619,20 @@ static void set_interaction_info_from_paint_session(paint_session* session)
...
 }
 
-void get_map_coordinates_from_pos_window(
-    rct_window* window, int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y, int32_t* interactionType,
+void get_map_coordinates_from_pos(
+    int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y, int32_t* interactionType,
     TileElement** tileElement, rct_viewport** viewport)
 {
-    ScreenCoordsXY screenCoords(screenX, screenY);
+    ScreenCoordsXY screenCoords{ screenX, screenY };
```
Gymnasiast: Why not change the function to take ScreenCoordsXY in the first place?

duncanspumpkin: Thats what I have done. This is making a shim so that I don't need to change every use of the function in one go. After this is merged I'll start moving the calling functions to use the new form. This function is used for a lot of things ~100+ and it would be very messy to change them all in one go.

Gymnasiast: Doh, I should have paid more attention!

duncanspumpkin: The diff made this look odd so not surprised

PR #10233, opened 2019-11-14
Use ScreenCoordsXY for windows/Map

tupaschoal:
- Refactor `place_park_entrance_get_map_position()` to receive `ScreenCoordsXY` and return `CoordsXYZD`
- Refactor `sub_68A15E()` to receive `ScreenCoordsXY` and return `CoordsXY`. Some of its usage are narrowing the result on purpose, as same variable is used for contexts where `int16_t` is still a hard requirement.

### DIFF
--- a/src/openrct2-ui/windows/Map.cpp
+++ b/src/openrct2-ui/windows/Map.cpp
@@ -187,11 +187,11 @@ static void window_map_paint_hud_rectangle(rct_drawpixelinfo* dpi);
 static void window_map_inputsize_land(rct_window* w);
 static void window_map_inputsize_map(rct_window* w);
 
-static void window_map_set_land_rights_tool_update(int32_t x, int32_t y);
-static void window_map_place_park_entrance_tool_update(int32_t x, int32_t y);
-static void window_map_set_peep_spawn_tool_update(int32_t x, int32_t y);
-static void window_map_place_park_entrance_tool_down(int32_t x, int32_t y);
-static void window_map_set_peep_spawn_tool_down(int32_t x, int32_t y);
+static void window_map_set_land_rights_tool_update(ScreenCoordsXY screenCoords);
+static void window_map_place_park_entrance_tool_update(ScreenCoordsXY screenCoords);
+static void window_map_set_peep_spawn_tool_update(ScreenCoordsXY screenCoords);
+static void window_map_place_park_entrance_tool_down(ScreenCoordsXY screenCoords);
+static void window_map_set_peep_spawn_tool_down(ScreenCoordsXY screenCoords);
 static void map_window_increase_map_size();
 static void map_window_decrease_map_size();
 static void map_window_set_pixels(rct_window* w);
@@ -467,13 +467,13 @@ static void window_map_toolupdate(rct_window* w, rct_widgetindex widgetIndex, Sc
     switch (widgetIndex)
     {
         case WIDX_SET_LAND_RIGHTS:
-            window_map_set_land_rights_tool_update(screenCoords.x, screenCoords.y);
+            window_map_set_land_rights_tool_update(screenCoords);
             break;
         case WIDX_BUILD_PARK_ENTRANCE:
-            window_map_place_park_entrance_tool_update(screenCoords.x, screenCoords.y);
+            window_map_place_park_entrance_tool_update(screenCoords);
             break;
         case WIDX_PEOPLE_STARTING_POSITION:
-            window_map_set_peep_spawn_tool_update(screenCoords.x, screenCoords.y);
+            window_map_set_peep_spawn_tool_update(screenCoords);
             break;
     }
 }
@@ -487,10 +487,10 @@ static void window_map_tooldown(rct_window* w, rct_widgetindex widgetIndex, Scre
     switch (widgetIndex)
     {
         case WIDX_BUILD_PARK_ENTRANCE:
-            window_map_place_park_entrance_tool_down(screenCoords.x, screenCoords.y);
+            window_map_place_park_entrance_tool_down(screenCoords);
             break;
         case WIDX_PEOPLE_STARTING_POSITION:
-            window_map_set_peep_spawn_tool_down(screenCoords.x, screenCoords.y);
+            window_map_set_peep_spawn_tool_down(screenCoords);
             break;
     }
 }
@@ -1156,13 +1156,13 @@ static void window_map_paint_hud_rectangle(rct_drawpixelinfo* dpi)
  *
  *  rct2: 0x0068D24E
  */
-static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
+static void window_map_set_land_rights_tool_update(ScreenCoordsXY screenCoords)
 {
     rct_viewport* viewport;
 
     map_invalidate_selection_rect();
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE;
-    CoordsXY mapCoords = screen_get_map_xy({ x, y }, &viewport);
+    CoordsXY mapCoords = screen_get_map_xy(screenCoords, &viewport);
     if (mapCoords.x == LOCATION_NULL)
         return;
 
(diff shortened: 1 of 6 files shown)


src/openrct2-ui/windows/Map.cpp
```diff
@@ -1189,89 +1189,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
-    sub_68A15E(x, y, mapX, mapY);
-    if (*mapX == LOCATION_NULL)
-        return;
-
-    auto surfaceElement = map_get_surface_element_at(*mapX >> 5, *mapY >> 5);
-    *mapZ = surfaceElement->GetWaterHeight();
-    if (*mapZ == 0)
+    CoordsXYZD parkEntranceMapPosition{};
+    const CoordsXY mapCoords = sub_68A15E(screenCoords);
+    parkEntranceMapPosition = { mapCoords.x, mapCoords.y, parkEntranceMapPosition.z, parkEntranceMapPosition.direction };
```
duncanspumpkin: Keep the initialised values as explicit 0 and invalid direction

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1189,89 +1189,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
+    if (parkEntranceMapPosition.x == LOCATION_NULL)
+        return parkEntranceMapPosition;
+
+    auto surfaceElement = map_get_surface_element_at(parkEntranceMapPosition.x >> 5, parkEntranceMapPosition.y >> 5);
```
duncanspumpkin: I think the is an override that takes a coordsxy

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1189,89 +1189,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
+    parkEntranceMapPosition.z = surfaceElement->GetWaterHeight();
+    if (parkEntranceMapPosition.z == 0)
```
duncanspumpkin: All of this z code should be setting the value in terms of big coordinates not micro. Will need the appropriate *8's

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1189,89 +1189,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
+    gMapSelectionTiles.push_back({ parkEntrancePosition.x - CoordsDirectionDelta[sideDirection].x,
+                                   parkEntrancePosition.y - CoordsDirectionDelta[sideDirection].y });
 
-    gMapSelectArrowPosition.x = mapX;
-    gMapSelectArrowPosition.y = mapY;
-    gMapSelectArrowPosition.z = mapZ * 16;
-    gMapSelectArrowDirection = direction;
+    gMapSelectArrowPosition.x = parkEntrancePosition.x;
+    gMapSelectArrowPosition.y = parkEntrancePosition.y;
+    gMapSelectArrowPosition.z = parkEntrancePosition.z * 16;
```
duncanspumpkin: Conversion from micro. Once anniversary changes are made this needs modified

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1189,89 +1189,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
+        && parkEntrancePosition.y == gParkEntranceGhostPosition.y
+        && parkEntrancePosition.direction == gParkEntranceGhostDirection)
     {
         return;
     }
 
     park_entrance_remove_ghost();
-    park_entrance_place_ghost(mapX, mapY, mapZ, direction);
+    park_entrance_place_ghost(
+        parkEntrancePosition.x, parkEntrancePosition.y, parkEntrancePosition.z, parkEntrancePosition.direction);
```
duncanspumpkin: Place ghost expects a micro z

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1302,16 +1305,15 @@ static void window_map_set_peep_spawn_tool_update(int32_t x, int32_t y)
...
-    int16_t mapX, mapY, mapZ;
-    int32_t direction;
-    place_park_entrance_get_map_position(x, y, &mapX, &mapY, &mapZ, &direction);
-    if (mapX != LOCATION_NULL)
+    CoordsXYZD parkEntrancePosition = place_park_entrance_get_map_position(screenCoords);
+    if (parkEntrancePosition.x != LOCATION_NULL)
     {
-        money32 price = place_park_entrance(mapX, mapY, mapZ, direction);
+        money32 price = place_park_entrance(
+            parkEntrancePosition.x, parkEntrancePosition.y, parkEntrancePosition.z, parkEntrancePosition.direction);
```
duncanspumpkin: Micro z

src/openrct2-ui/windows/Map.cpp
tupaschoal: There is, thanks!

tupaschoal:
I've fixed the issues you mentioned @duncanspumpkin and also rebased to solve conflicts. Please double check the micro/big z changes to make sure I got it right

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1188,89 +1188,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
+    auto surfaceElement = map_get_surface_element_at(mapCoords);
+    parkEntranceMapPosition.z = surfaceElement->GetWaterHeight() * 8;
+    if (parkEntranceMapPosition.z == 0)
     {
-        *mapZ = surfaceElement->base_height / 2;
+        parkEntranceMapPosition.z = surfaceElement->base_height * 8;
         if ((surfaceElement->GetSlope() & TILE_ELEMENT_SLOPE_ALL_CORNERS_UP) != 0)
         {
-            (*mapZ)++;
+            parkEntranceMapPosition.z++;
```
duncanspumpkin: += 16

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1188,89 +1188,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
             if (surfaceElement->GetSlope() & TILE_ELEMENT_SLOPE_DOUBLE_HEIGHT)
             {
-                (*mapZ)++;
+                parkEntranceMapPosition.z++;
```
duncanspumpkin: 16

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1188,89 +1188,92 @@ static void window_map_set_land_rights_tool_update(int32_t x, int32_t y)
...
 
     map_invalidate_selection_rect();
     map_invalidate_map_selection_tiles();
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE;
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE_ARROW;
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE_CONSTRUCT;
-    place_park_entrance_get_map_position(x, y, &mapX, &mapY, &mapZ, &direction);
-    if (mapX == (int16_t)-1)
+    CoordsXYZD parkEntrancePosition = place_park_entrance_get_map_position(screenCoords);
+    if (parkEntrancePosition.x == -1)
```
duncanspumpkin: Location null

src/openrct2-ui/windows/Map.cpp
tupaschoal:
```suggestion
            parkEntranceMapPosition.z += 16;
```

src/openrct2-ui/windows/Map.cpp
tupaschoal:
```suggestion
                parkEntranceMapPosition.z += 16;
```

src/openrct2-ui/windows/Map.cpp
tupaschoal:
```suggestion
    if (parkEntrancePosition.x == LOCATION_NULL)
```

tupaschoal:
Solved the pending issues, should be alright now

src/openrct2-ui/windows/Map.cpp
Gymnasiast: Just to check: normally you would use a factor of 8 when converting from small to big coordinates.

Was parkEntranceMapPosition.z using half-tile units before?

Gymnasiast:
Some of the diff looks weird, but that it caused by the park entrance structures using coordinates that are neither big nor small, but an oddball mixture. That will need to be refactored, but not in this PR.

PR #10238, opened 2019-11-14
Refactor screen_get_map_xy_with_z to use CoordsXY structs

duncanspumpkin:
Continuing on the refactoring.

### DIFF
--- a/src/openrct2/interface/Viewport.h
+++ b/src/openrct2/interface/Viewport.h
@@ -174,7 +174,7 @@ void sub_68B2B7(paint_session* session, CoordsXY mapCoords);
 void viewport_invalidate(rct_viewport* viewport, int32_t left, int32_t top, int32_t right, int32_t bottom);
 
 CoordsXY screen_get_map_xy(ScreenCoordsXY screenCoords, rct_viewport** viewport);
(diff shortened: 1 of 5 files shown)


src/openrct2/interface/Viewport.h
```diff
@@ -174,7 +174,7 @@ void sub_68B2B7(paint_session* session, int32_t x, int32_t y);
 void viewport_invalidate(rct_viewport* viewport, int32_t left, int32_t top, int32_t right, int32_t bottom);
 
 CoordsXY screen_get_map_xy(ScreenCoordsXY screenCoords, rct_viewport** viewport);
-void screen_get_map_xy_with_z(int16_t screenX, int16_t screenY, int16_t z, int16_t* mapX, int16_t* mapY);
+CoordsXY screen_get_map_xy_with_z(ScreenCoordsXY screenCoords, int16_t z);
```
tupaschoal: What does it really mean a screen coordinate with z, I was wondering

duncanspumpkin: Its some poor naming thats what it is. The none z version will get the tile that the item you are hovering over is on so if hovering on a tree that would be the base of the tree. The with z version gets the tile ignoring any objects but adds the z coordinate to the y component of your position. Not sure if that fully makes sense but place some scenery then hold down z or ctrl to see the different modes.

PR #10245, opened 2019-11-15
Changed logic in ride_prepare_breakdown to ensure Ride::inspection_station is set.

### LINKED ISSUE #10214: Mechanics can be called to an uninitialized station id
`ride_call_mechanic` uses the `Ride::inspection_station` field, which isn't always initialized. Specifically, not all cases in the `ride_prepare_breakdown` function sets it.

NoLongerBreathedIn:
Fixes #10214

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -2430,6 +2430,7 @@ void ride_prepare_breakdown(Ride* ride, int32_t breakdownReason)
     ride->breakdown_reason_pending = breakdownReason;
     ride->breakdown_sound_modifier = 0;
     ride->not_fixed_timeout = 0;
+    ride->inspection_station = 0; // ensure set to something.
 
     switch (breakdownReason)
     {


src/openrct2/ride/Ride.cpp
```diff
@@ -2542,6 +2539,7 @@ void ride_prepare_breakdown(Ride* ride, int32_t breakdownReason)
                         vehicle->update_flags |= VEHICLE_UPDATE_FLAG_BROKEN_CAR;
                 }
             }
+	    ride->inspection_station = 0;
```
Gymnasiast: The indentation on this line (and most others) is not correct. This is also why the CI fails.

As for the line itself: can't this breakdown occur at *any* station? Resetting it to 0 doesn't look quite right.

NoLongerBreathedIn: AAGH spaces/tabs [I have emacs set at 8 space tab stops]. Well, that's fixed now.

And this is a breakdown of a vehicle, not of a station, so it doesn't really matter what station.

duncanspumpkin:
Rebase, this will probably need a new Replay for bpb

tupaschoal:
I've rebased it.

tupaschoal:
Yay, it didn't break anything (should it have? heheh)

PR #10254, opened 2019-11-17
Coords for ride construction

tupaschoal:
I also updated the global variable `_trackPlaceShiftStart`. Seems fairly straight forward to remove this global, if only it had a class-like structure.

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -447,8 +447,7 @@ static constexpr const rct_string_id RideConfigurationStringIds[] = {
 static bool _trackPlaceCtrlState;
 static int32_t _trackPlaceCtrlZ;
 static bool _trackPlaceShiftState;
-static int32_t _trackPlaceShiftStartScreenX;
-static int32_t _trackPlaceShiftStartScreenY;
+static ScreenCoordsXY _trackPlaceShiftStart;
 static int32_t _trackPlaceShiftZ;
 static int32_t _trackPlaceZ;
 static money32 _trackPlaceCost;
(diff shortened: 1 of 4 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2120,7 +2119,8 @@ static bool ride_get_place_position_from_screen_position(int32_t screenX, int32_
         if (gInputPlaceObjectModifier & PLACE_OBJECT_MODIFIER_COPY_Z)
         {
             TileElement* tileElement;
-            get_map_coordinates_from_pos(screenX, screenY, 0xFCCA, &mapX, &mapY, &interactionType, &tileElement, &viewport);
+            get_map_coordinates_from_pos(
+                screenCoords.x, screenCoords.y, 0xFCCA, &mapX, &mapY, &interactionType, &tileElement, &viewport);
```
duncanspumpkin: You can start using the new ScreenCoords and CoordsXY version of `get_map_coordinates_from_pos`

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3558,27 +3553,27 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
...
     gMapSelectArrowPosition.z = z;
     bx = 41;
-    _currentTrackBegin.x = x;
-    _currentTrackBegin.y = y;
+    _currentTrackBegin.x = mapCoords.x;
+    _currentTrackBegin.y = mapCoords.y;
     _currentTrackBegin.z = z;
-    if ((_currentTrackSelectionFlags & TRACK_SELECTION_FLAG_TRACK) && x == _previousTrackPiece.x && y == _previousTrackPiece.y
-        && z == _previousTrackPiece.z)
+    if ((_currentTrackSelectionFlags & TRACK_SELECTION_FLAG_TRACK) && mapCoords.x == _previousTrackPiece.x
```
duncanspumpkin: A comparison between currentTrackBegin and previousTrackPiece could simplify this instead of breaking it into the components.

src/openrct2-ui/windows/RideConstruction.cpp
tupaschoal: I will, thought it hadn't been merged yet.

Gymnasiast:
@duncanspumpkin 

> I'm not really a fan of creating more code that uses LOCATION_NULL we should be trying to reduce the use of it.

What should be used instead?

IntelOrca:
`std::optional`

Gymnasiast:
@IntelOrca You can't save that into a variable, though, can you? Currently, coords.x == LOCATION_NULL is often used to mark a certain location as not set. It's not just return values.

IntelOrca:
> You can't save that into a variable, though, can you?

You mean this?:
```
std::optional<CoordsXY> coords;
```

Yes you can do that.

Gymnasiast:
What I meant was that it isn't possible to do `std::optional` on coords**.x**.

But that doesn't really matter, because it answered my question anyway: the whole coords variable should be marked as std::optional.

duncanspumpkin:
Sorry should have clarified. This function at present doesn't indicate failure at all. This pr correctly introduces a failure state by setting location null. I would prefer it did a std optional as that means you have to check for failure. Yes a lot of code uses location null but we should try to avoid adding more.

tupaschoal:
I've updated the code to address the review comments, including returning `std::optional`

PR #10264, opened 2019-11-19
Prototype OSSpecific

tupaschoal:
I once talked about OS specific things on https://github.com/OpenRCT2/OpenRCT2/pull/10198#pullrequestreview-313135711 and this is kind of what I was thinking about.

The idea is to remove the defines from the source code and move the implementation of OS specific stuff to their OS counterpart, like done here, so the actual implementation is agnostic to the OS and cleaner. I'm opening it here, as a PoC, for discussion.

(At the time, I didn't know there was `src/openrct2/platform`, hence why I did it in a separate folder/files. This could all be merged into `platform`)

---- Edit ----

This is the first part, I think it benefits from doing incrementally, just like all the other big projects, but my ultimate desire is that all places that branch out on ifdefs actually start calling something like `OSSpecific` and the code there is not ugly.

I removed these two specific `mac` calls to begin with.

### DIFF
--- a/src/openrct2/platform/Platform.Android.cpp
+++ b/src/openrct2/platform/Platform.Android.cpp
@@ -44,6 +44,16 @@ namespace Platform
         Guard::Assert(false, "GetCurrentExecutablePath() not implemented for Android.");
         return std::string();
     }
+
+    uintptr_t StrDecompToPrecomp(utf8* input)
+    {
+        return reinterpret_cast<uintptr_t>(input);
+    }
+
+    bool HandleSpecialCommandLineArgument(const char* argument)
+    {
+        return false;
+    }
 } // namespace Platform
 
 #endif
--- a/src/openrct2-ui/windows/LoadSave.cpp
+++ b/src/openrct2-ui/windows/LoadSave.cpp
@@ -701,11 +701,7 @@ static void window_loadsave_paint(rct_window* w, rct_drawpixelinfo* dpi)
     safe_strcpy(ch, _shortenedDirectory, sizeof(buffer) - (ch - buffer));
 
     // Draw path text
-#ifdef __APPLE__
-    set_format_arg(0, uintptr_t, (uintptr_t)macos_str_decomp_to_precomp(buffer));
-#else
-    set_format_arg(0, uintptr_t, (uintptr_t)buffer);
-#endif
+    set_format_arg(0, uintptr_t, Platform::StrDecompToPrecomp(buffer));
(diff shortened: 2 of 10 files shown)


ZehMatt:
I like the idea of cleaning up platform part a bit. However you should adopt to the currently existing files, theres already Platform.Win32.cpp, Platform.Posix.cpp etc.

Also please don't introduce a function called toUintptr_t, the place and name is rather abstract for a helper function that no one will use.

tupaschoal:
Yes, the idea is to merge it into their platform counterparts. When I started doing this PoC I didn't know it exist.

tupaschoal:
The failures seem unrelated to the changeset.

I'd love getting feedback on this to see if I'll pursue more PRs

IntelOrca:
Why not keep our current pattern of using `Platform::` rather than making a new name that serves the same purpose called OSSpecific?

tupaschoal:
It works too, I can adapt it to do so. What's the difference between `Platform.<OS>` and `<OS>` files?

IntelOrca:
So the OS files are old legacy. The new ones are Platform.Win32, Platform.Linux and Platform.MacOS which are done by having namespaces for each platform which define the same functions. We also have Platform.Posix for any shared logic.

All old legacy platform functions should be moved to the new Platform files and anything still #ifdefed like what you found in this PR can also be moved to Platform.* files.

tupaschoal:
Alright, sounds great, thanks for the explanation. I'll get onto it

src/openrct2/platform/Platform.Android.cpp
```diff
@@ -44,6 +44,16 @@ namespace Platform
         Guard::Assert(false, "GetCurrentExecutablePath() not implemented for Android.");
         return std::string();
     }
+
+    uintptr_t StrDecompToPrecomp(utf8* input)
+    {
+        return platform_str_decomp_to_precomp(input);
```
Gymnasiast: Do not redirect to a C-style function. Either handle it directly or create a namespaced C++ style helper function.

tupaschoal: All of the `Platform.*`  are implementing the functions declared in ´Platform2.h`. but I didn't want to replicate the same code for instances where there was no special handling (such as everything that is not a MacOS here), so I made them call things in `Shared`

Gymnasiast: Yes, but this function is so trivial that redirecting is not useful. And even if it was, it still shouldn't redirect to a C-style function, but to a C++-style one.

In any case: please replace the redirection with the actual implementation here.

tupaschoal: Done.

PR #10265, opened 2019-11-19
Build versioned AppImages

TheAssassin:
This tiny PR makes sure AppImages built by the build script contain a version number, both in the AppImage's filename as well as in the metadata (i.e., the desktop file embedded in the AppImage).

The script still works just fine as-is, by the way.

### DIFF
--- a/scripts/linux/build-appimage.sh
+++ b/scripts/linux/build-appimage.sh
@@ -26,6 +26,8 @@ OLD_CWD=$(readlink -f .)
 
 pushd "$BUILD_DIR"
 
+export VERSION=$(cd "$REPO_ROOT" && git describe --tags)
+
 # standard linuxdeploy pattern
 #see https://docs.appimage.org/packaging-guide/from-source/index.html for more information
 cmake "$REPO_ROOT" -DCMAKE_INSTALL_PREFIX=/usr -G Ninja -DCMAKE_BUILD_TYPE=Release


Gymnasiast:
Line 13 still contains `XXXXXX`, which looks like a placeholder.

TheAssassin:
@Gymnasiast that line is part of an `mktemp` command. In our case, the command generates a temporary directory. The placeholder is filled by `mktemp` with a unique, random string, probably with some collision avoidance built in.

The temporary directory is used for building to ensure every new AppImage build cycle is isolated from the source directory and all previous builds. This allows for simulating CI-like conditions locally, and you don't have to invest a lot of thought while writing your build scripts. Just make a change and start all over.

IntelOrca:
@TheAssassin FYI, for our released app images I do not use this script as I need it to work in a particular way for the GitHub CI environment.

The CI job is here: https://github.com/OpenRCT2/OpenRCT2/blob/develop/.github/workflows/ci.yml

TheAssassin:
@IntelOrca already guessed so, after having sent this PR while looking at this weird new GitHub job thingy. But I wanted to fix the existing script nevertheless.

By the way, what's this Discord RPC stuff about? I'm pretty sure that thing starts transmitting even if not used by the user, which might violate some laws. Note for self: I'll have to check myself tonight, when downloading the offical AppImage.

IntelOrca:
@TheAssassin Discord-RPC is used to post game status to Discord. It uses local loopback TCP to connect to a running local Discord RPC server (a.k.a. Discord).

TheAssassin:
I see, that answers my question. I was thinking of another Discord related component. This one is harmless as long as you don't have Discord running, I guess. Thanks for clarifying.

PR #10266, opened 2019-11-19
Simplify assignment in RideConstruction.cpp

tupaschoal:
Woke up too late to incorporate this into #10254 before it was merged 😄


duncanspumpkin:
Oh it was more a thought for future. Its fine as is.

PR #10271, opened 2019-11-19
Support NetBSD and simplify platform ifdefs.

niacat:
For files intended to support unixy freedesktopy systems, reduce the ifdef down to `__unix__ && !__APPLE__ && !__ANDROID__`, rather than naming linux and all the BSDs and so on by hand.

Should simplify adding support for systems like illumos in the future.

NetBSD support depends on #10270

### DIFF
--- a/src/openrct2/platform/Platform.Linux.cpp
+++ b/src/openrct2/platform/Platform.Linux.cpp
@@ -130,7 +130,11 @@ namespace Platform
             log_fatal("failed to read /proc/self/exe");
         }
 #    elif defined(__FreeBSD__) || defined(__NetBSD__)
+#        if defined(__FreeBSD__)
         const int32_t mib[] = { CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1 };
+#        else
+        const int32_t mib[] = { CTL_KERN, KERN_PROC_ARGS, -1, KERN_PROC_PATHNAME };
+#        endif
         auto exeLen = sizeof(exePath);
         if (sysctl(mib, 4, exePath, &exeLen, nullptr, 0) == -1)
         {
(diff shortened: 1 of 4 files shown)


Gymnasiast:
@ibara I expect this to work correctly on OpenBSD, but could you perhaps check?

src/openrct2/platform/Platform.Linux.cpp
```diff
@@ -129,7 +129,7 @@ namespace Platform
         {
             log_fatal("failed to read /proc/self/exe");
         }
-#    elif defined(__FreeBSD__)
+#    elif defined(__FreeBSD__) || defined(__NetBSD__)
         const int32_t mib[] = { CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1 };
```
krytarowski: NetBSD needs different parameters here.

ibara:
> @ibara I expect this to work correctly on OpenBSD, but could you perhaps check?

This works for us, yes.

src/openrct2/platform/Platform.Linux.cpp
Gymnasiast: Can you elaborate? Which parameters does it need?

krytarowski: https://nxr.netbsd.org/xref/src/lib/libexecinfo/backtrace.c#63

Gymnasiast:
@niacat Could you address the line note, so that we can merge this?

src/openrct2/platform/Platform.Linux.cpp
Gymnasiast: So if I understand correctly, on NetBSD, this line should read as follows:
```C++
const int32_t mib[] =  { CTL_KERN, KERN_PROC_ARGS, -1, KERN_PROC_PATHNAME };
```
?

krytarowski: Yes.

PR #10281, opened 2019-11-23
Added a benchmark for game logic

frutiemax:
Google benchmark parameters seem to work. Still need to figure out what to do with CPU iterations. The command to be called is openrct2 benchlogic xxx.sv6

### DIFF
--- a/src/openrct2/cmdline/BenchLogic.cpp
+++ b/src/openrct2/cmdline/BenchLogic.cpp
@@ -52,7 +52,8 @@ static void BM_gamelogic(benchmark::State& state, IContext* context)
         static_cast<double>(microseconds) };
     auto start = std::chrono::high_resolution_clock::now();
 
-    context->GetGameState()->UpdateLogic();
+    for (auto _ : state)
+        context->GetGameState()->UpdateLogic();
 
     auto end = std::chrono::high_resolution_clock::now();
     auto elapsed_seconds = std::chrono::duration_cast<std::chrono::duration<double>>(end - start);
--- a/src/openrct2/cmdline/BenchLogic.cpp
+++ b/src/openrct2/cmdline/BenchLogic.cpp
@@ -1,0 +1,136 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2019 OpenRCT2 developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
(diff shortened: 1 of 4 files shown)


src/openrct2/cmdline/BenchLogic.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include "../GameState.h"
+
+
+#include <cstdlib>
+#include <chrono>
+#include <memory>
+
+using namespace OpenRCT2;
+
+static void BM_gamelogic(benchmark::State& state, IContext* context)
```
janisozaur: Why can't it use the standard form
```
for (auto _ : state)
{
    context->GetGameState()->UpdateLogic();
}
```
?

janisozaur:
Other than the comment I posted above:
1. Make sure you formatted your code
2. Make sure your commit message conforms to https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

The xcode failure needs to be handled by someone with access to xcode. Ping @AaronVanGeffen @Gymnasiast

src/openrct2/cmdline/BenchLogic.cpp
frutiemax: Actually, you found the problem with my code, you actually need this. Why?

janisozaur:
That's not what I meant. I think in current form it will cause problems because you enabled manual timing and provide your own measurements, while all this should be handled by `state`. Why is it necessary to do your own timing instead of using the benchmark's one?

frutiemax:
It is not necessary to do manual timing. But it works nonetheless and it is one of the examples in the documentation of the Google Benchmark library.

janisozaur:
Since it is not necessary, I would prefer to use the simpler invocation

AaronVanGeffen:
@frutiemax I've rebased this PR and pushed a fix for the Xcode project. Do you intend to continue to work on this?

src/openrct2/cmdline/BenchLogic.cpp
chfast: The loop is actually measuring the time, you don't need your timer. I'm just not sure it is safe to call `context->GetGameState()->UpdateLogic()` repeatedly in a loop.

src/openrct2/cmdline/BenchLogic.cpp
```diff
@@ -0,0 +1,137 @@
...
```
chfast: You also don't need to call the function with `BM_` prefix.

AaronVanGeffen:
This is essentially doing the same thing as #10700. We need to decide which to merge.

PR #10282, opened 2019-11-23
Fixed compilation errors with Visual Studio 2019 Preview 2.0

tjwrona1992:
Fix errors where "std::max" was not recognized while compiling with the latest Visual Studio (2019 16.4.0 Preview 2.0) due to a missing "#include" statement.

image

### DIFF
--- a/src/openrct2/core/MemoryStream.h
+++ b/src/openrct2/core/MemoryStream.h
@@ -10,8 +10,9 @@
 #pragma once
 
 #include "../common.h"
-#include <algorithm>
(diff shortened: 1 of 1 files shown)


janisozaur:
I have doubts this will pass clang-format, as it sorts the includes

tjwrona1992:
@janisozaur it looks like it did not. What exactly is clang-format and why would it fail?

janisozaur:
Make sure your commits match https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

PR #10285, opened 2019-11-24
Remove raw coordinates overload of get_map_coordinates_from_pos

tupaschoal:
Part of https://github.com/OpenRCT2/OpenRCT2/projects/9

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -1632,20 +1632,6 @@ void get_map_coordinates_from_pos(
     get_map_coordinates_from_pos_window(window, screenCoords, flags, mapCoords, interactionType, tileElement, viewport);
 }
 
-void get_map_coordinates_from_pos(
-    int32_t screenX, int32_t screenY, int32_t flags, int16_t* x, int16_t* y, int32_t* interactionType,
-    TileElement** tileElement, rct_viewport** viewport)
-{
-    ScreenCoordsXY screenCoords{ screenX, screenY };
-    CoordsXY mapCoords;
-    get_map_coordinates_from_pos(screenCoords, flags, mapCoords, interactionType, tileElement, viewport);
-
-    if (x != nullptr)
-        *x = mapCoords.x;
-    if (y != nullptr)
-        *y = mapCoords.y;
-}
-
 void get_map_coordinates_from_pos_window(
     rct_window* window, ScreenCoordsXY screenCoords, int32_t flags, CoordsXY& mapCoords, int32_t* interactionType,
     TileElement** tileElement, rct_viewport** viewport)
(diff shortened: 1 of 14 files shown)


duncanspumpkin:
You will need to rebase due to the other PR's being merged.

src/openrct2/interface/Viewport.cpp
```diff
@@ -1774,26 +1760,25 @@ static rct_viewport* viewport_find_from_point(ScreenCoordsXY screenCoords)
...
-    CoordsXY map_pos = { my_x + 16, my_y + 16 };
+    map_pos = { map_pos.x + 16, map_pos.y + 16 };
 
     for (int32_t i = 0; i < 5; i++)
     {
         int32_t z = tile_element_height(map_pos);
         map_pos = viewport_coord_to_map_coord(start_vp_pos.x, start_vp_pos.y, z);
-        map_pos.x = std::clamp<int16_t>(map_pos.x, my_x, my_x + 31);
-        map_pos.y = std::clamp<int16_t>(map_pos.y, my_y, my_y + 31);
+        map_pos.x = std::clamp<int32_t>(map_pos.x, map_pos.x, map_pos.x + 31);
```
duncanspumpkin: This is subtly different. my_x and my_y are the initial position. map_pos is the new position. map_pos is initialised with a value that will round up to the nearest tile.

tupaschoal: Nice catch!

tupaschoal:
Rebased and solved issue!

src/openrct2/interface/Viewport.cpp
```diff
@@ -1773,26 +1759,25 @@ static rct_viewport* viewport_find_from_point(ScreenCoordsXY screenCoords)
...
         int32_t z = tile_element_height(map_pos);
-        map_pos = viewport_coord_to_map_coord(start_vp_pos.x, start_vp_pos.y, z);
-        map_pos.x = std::clamp<int16_t>(map_pos.x, my_x, my_x + 31);
-        map_pos.y = std::clamp<int16_t>(map_pos.y, my_y, my_y + 31);
+        modifiedPos = viewport_coord_to_map_coord(start_vp_pos.x, start_vp_pos.y, z);
+        modifiedPos.x = std::clamp<int32_t>(modifiedPos.x, map_pos.x, map_pos.x + 31);
+        modifiedPos.y = std::clamp<int32_t>(modifiedPos.y, map_pos.y, map_pos.y + 31);
     }
 
     if (viewport != nullptr)
```
duncanspumpkin: Still not right you need to return `modifiedPos`

src/openrct2/interface/Window.cpp
```diff
@@ -959,7 +959,8 @@ void window_viewport_get_map_coords_by_cursor(
     context_get_cursor_position_scaled(&mouse_x, &mouse_y);
 
     // Compute map coordinate by mouse position.
-    get_map_coordinates_from_pos(mouse_x, mouse_y, VIEWPORT_INTERACTION_MASK_NONE, map_x, map_y, nullptr, nullptr, nullptr);
+    CoordsXY mapCoords;
+    get_map_coordinates_from_pos({ mouse_x, mouse_y }, VIEWPORT_INTERACTION_MASK_NONE, mapCoords, nullptr, nullptr, nullptr);
 
     // Get viewport coordinates centring around the tile.
     int32_t z = tile_element_height({ *map_x, *map_y });
```
duncanspumpkin: will need to be modified to use the `mapCoords` see also below

tupaschoal:
Should be fixed now, but I won't say for sure, because I was wrong twice already :X

tupaschoal:
Is this ok for merge?

src/openrct2/interface/Window.cpp
duncanspumpkin: Saved the wrong coordinates

src/openrct2/interface/Viewport.cpp
```diff
@@ -1773,32 +1759,31 @@ static rct_viewport* viewport_find_from_point(ScreenCoordsXY screenCoords)
...
         return { LOCATION_NULL, 0 };
     }
 
     LocationXY16 start_vp_pos = screen_coord_to_viewport_coord(myViewport, screenCoords);
-    CoordsXY map_pos = { my_x + 16, my_y + 16 };
+    CoordsXY modifiedPos = { map_pos.x + 16, map_pos.y + 16 };
 
     for (int32_t i = 0; i < 5; i++)
     {
         int32_t z = tile_element_height(map_pos);
```
duncanspumpkin: This should have been modifiedPos

PR #10297, opened 2019-11-27
Prefer std::optional over LOCATION_NULL on Viewport

tupaschoal:


src/openrct2-ui/windows/LandRights.cpp
```diff
@@ -295,9 +295,9 @@ static void window_land_rights_tool_update_land_rights(ScreenCoordsXY screenCoor
     map_invalidate_selection_rect();
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE;
 
-    CoordsXY mapTile = screen_get_map_xy(screenCoords, nullptr);
+    auto mapTile = screen_get_map_xy(screenCoords, nullptr);
 
-    if (mapTile.x == LOCATION_NULL)
+    if (!mapTile)
```
Gymnasiast: We always compare with `nullptr`.

src/openrct2-ui/windows/LandRights.cpp
```diff
@@ -328,35 +328,35 @@ static void window_land_rights_tool_update_land_rights(ScreenCoordsXY screenCoor
     int16_t tool_length = (tool_size - 1) * 32;
 
     // Move to tool bottom left
-    mapTile.x -= (tool_size - 1) * 16;
-    mapTile.y -= (tool_size - 1) * 16;
-    mapTile.x &= 0xFFE0;
-    mapTile.y &= 0xFFE0;
+    mapTile->x -= (tool_size - 1) * 16;
+    mapTile->y -= (tool_size - 1) * 16;
+    mapTile->x &= 0xFFE0;
```
Gymnasiast: This is such a common pattern that I think we should create a method for it. Called `Truncate` or something like that.

Not necessarily something for this PR, though.

src/openrct2-ui/windows/LandRights.cpp
duncanspumpkin: This isn't a nullptr it's a optional. This is the way we have done this elsewhere

src/openrct2-ui/windows/LandRights.cpp
duncanspumpkin: Agreed. Should be a method from CoordsXY. That and centre coordinate as there is tons of x+16, y+16

Gymnasiast: Yeah, I though about that one too. I created https://github.com/OpenRCT2/OpenRCT2/issues/10298 for that.

src/openrct2-ui/windows/LandRights.cpp
Gymnasiast: I wonder if we should. Before optionals came along, we used explicit comparisons for anything that wasn't a boolean. I think explicit comparisons are clearer.

AaronVanGeffen: A `std::optional` has well-defined behaviour through the type system, though. I think this does add semantic value over a simple nullptr.

ZehMatt: Strictly speaking it is a boolean operation, see https://en.cppreference.com/w/cpp/utility/optional/operator_bool

tupaschoal: Since it was marked as resolved, I'm assuming nothing needs to be done.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2192,7 +2192,15 @@ static std::optional<CoordsXY> ride_get_place_position_from_screen_position(Scre
...
         auto mapZ = _trackPlaceCtrlZ;
-        mapCoords = screen_get_map_xy_with_z(screenCoords, mapZ);
+        auto mapXYCoords = screen_get_map_xy_with_z(screenCoords, mapZ);
+        if (mapXYCoords)
+        {
+            mapCoords = *mapXYCoords;
+        }
+        else
+        {
+            mapCoords.x = LOCATION_NULL;
```
duncanspumpkin: return a nullopt since thats all that is going to happen after this section of code.

src/openrct2/interface/Window.cpp
```diff
@@ -915,11 +915,12 @@ void window_rotate_camera(rct_window* w, int32_t direction)
 
     // has something to do with checking if middle of the viewport is obstructed
     rct_viewport* other;
-    CoordsXYZ coords{ screen_get_map_xy({ x, y }, &other), 0 };
+    auto mapXYCoords = screen_get_map_xy({ x, y }, &other);
+    CoordsXYZ coords{};
 
     // other != viewport probably triggers on viewports in ride or guest window?
     // x is LOCATION_NULL if middle of viewport is obstructed by another window?
```
duncanspumpkin: update this comment

tupaschoal:
Done

Issue #10298, opened 2019-11-27
Create functions for common operations on CoordsXY(Z)(D)


Gymnasiast:
There are two common operations on CoordsXY.

The first:
```
coords.x &= ~0x1F;
coords.y &= ~0x1F;
```

This resets them to the beginning of the tile.

The second:
```
coords.x &= ~0x1F;
coords.x += 16;
coords.y &= ~0x1F;
coords.y += 16;
```

This changes the coordinates to the centre of the tile.

Considering they're quite common, I think we should create methods for them. Some things to discuss are:
- Their naming
- Whether they should just concern x and y, or also modify z.
- Whether they should modify the object itself, or return a modified version and leave the original intact.

duncanspumpkin:
I agree. For the other modifying function (rotate) it returns a modified new object retaining the original object. I suggest we do the same.

Z coordinates don't factor into this. I don't think there are any instances where z is truncated.

Also note for any person making this change ~0x1F is the same as 0xFFE0 which is also used in lots of places.

tupaschoal:
Sounds like good functions indeed. Would naming them `toTileStart` and `toTileCentre` be good enough?

IntelOrca:
> Whether they should modify the object itself, or return a modified version and leave the original intact.

Definitely return a new one, leave original intact.

tupaschoal:
I didn't find many combinations of what you described to be `toTileCentre`, unless these multiplication operations also mean that
```
    mapTile.x -= (tool_size - 1) * 16;
    mapTile.y -= (tool_size - 1) * 16;
    mapTile.x &= 0xFFE0;
    mapTile.y &= 0xFFE0;
```

duncanspumpkin:
@tupaschoal that would be a subtly different tile centre. A lot of the code that does centering isn't converted to using a CoordsXY at the moment. You will see it a bit in the peep code. Some of the time its not really centring as much as it is rounding up.

If you do a search for 0xFFE0 or ~0x1F you will see that there are a lot of instances of setting to tile start. I've also thought of another variation of the same thing floor2(..., 32) is the same as `& 0xFFE0`.

tupaschoal:
K searched for all the quoted `&` operations, and changed every place that was using `CoordsXY`. The `floor2` option I didn't seek

tupaschoal:
I guess this can be closed now

PR #10303, opened 2019-11-29
Create CoordsXY::ToTileStart()

tupaschoal:
Helps #10298

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -4981,13 +4981,11 @@ void Guest::UpdateRideMazePathfinding()
             targetLoc.y = destination_y;
             if (chosenEdge & 1)
             {
-                targetLoc.x &= 0xFFE0;
-                targetLoc.x += 16;
+                targetLoc.x = targetLoc.ToTileCentre().x;
             }
             else
             {
-                targetLoc.y &= 0xFFE0;
-                targetLoc.y += 16;
+                targetLoc.y = targetLoc.ToTileCentre().y;
             }
             destination_x = targetLoc.x;
             destination_y = targetLoc.y;
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -3078,8 +3078,8 @@ void Peep::PerformNextAction(uint8_t& pathing_result, TileElement*& tile_result)
     }
 
     auto newLoc = *loc;
-    CoordsXY centeredNewLoc = newLoc.ToTileStart();
-    if (centeredNewLoc.x == next_x && centeredNewLoc.y == next_y)
+    CoordsXY truncatedNewLoc = newLoc.ToTileStart();
+    if (truncatedNewLoc.x == next_x && truncatedNewLoc.y == next_y)
     {
         int16_t height = GetZOnSlope(newLoc.x, newLoc.y);
         MoveTo(newLoc.x, newLoc.y, height);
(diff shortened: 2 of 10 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -4981,12 +4981,12 @@ void Guest::UpdateRideMazePathfinding()
             targetLoc.y = destination_y;
             if (chosenEdge & 1)
             {
-                targetLoc.x &= 0xFFE0;
+                targetLoc.x = targetLoc.ToTileStart().x;
                 targetLoc.x += 16;
```
Gymnasiast: This is a candidate for `ToTileCentre`.

src/openrct2/world/Location.hpp
```diff
@@ -139,6 +139,11 @@ struct CoordsXY
 
         return rotatedCoords;
     }
+
+    CoordsXY ToTileStart() const
+    {
+        return { x & ~0x1F, y & ~0x1F };
```
duncanspumpkin: Would `floor2( x, 32)` be better its slightly clearer why it is doing this.

src/openrct2/peep/Guest.cpp
tupaschoal: Done

src/openrct2/world/Location.hpp
tupaschoal: I have mixed feelings, after I got around what `floor2` does, it seems ok to use it instead, but I'm not sure. What's your take @Gymnasiast ?

tupaschoal:
I added `CoordsXY::ToTileCentre()` and also used `CoordsXY::ToTileStart()` in all places where `floor2` was being applied to a `CoordsXY`

src/openrct2/world/Location.hpp
Gymnasiast: I wonder if `floor2(x, 32)` get properly optimised to `x & ~0x1F` by the compiler. If it doesn't, I prefer `x & 0x1F`.

Another option is adding a comment: `Tiles are divided into 32 x coordinates and 32 y coordinates. We truncate it to a multiple of 32 to go back to the start of the tile.`

tupaschoal:
I solved the `clang-format` issue and rebased.

src/openrct2/world/Location.hpp
duncanspumpkin: Floor2 is just a macro doing exactly that. I suspect it's the same

src/openrct2/peep/Peep.cpp
```diff
@@ -3078,7 +3078,8 @@ void Peep::PerformNextAction(uint8_t& pathing_result, TileElement*& tile_result)
     }
 
     auto newLoc = *loc;
-    if ((newLoc.x & 0xFFE0) == next_x && (newLoc.y & 0xFFE0) == next_y)
+    CoordsXY centeredNewLoc = newLoc.ToTileStart();
```
Gymnasiast: You call the variable `centered`, but it isn't centred. `truncatedNewLoc` is a better name IMO.

(Also note that in `centered` is American English and that it's `centred` in British English, which is used for OpenRCT2.)

tupaschoal: I did know that, because I proposed `ToTileCentre`, just failed here hahaha

src/openrct2/world/Location.hpp
tupaschoal: I updated it to use `floor2` instead, will submit another PR to improve `floor2` documentation and implementation

PR #10309, opened 2019-11-30
Make floor2 and ceil2 constexpr functions

tupaschoal:
Based on the discussion here https://github.com/OpenRCT2/OpenRCT2/pull/10303#discussion_r352280292 , it seemed worth turning these functions into `constexpr` ones for improved readability and also because we can now guarantee users are sending in a power of two.

### DIFF
--- a/src/openrct2/common.h
+++ b/src/openrct2/common.h
@@ -59,14 +59,14 @@ constexpr bool is_power_of_2(int v)
 constexpr int floor2(const int x, const int y)
 {
     assert(is_power_of_2(y));
-    return ((x) & (~((y)-1)));
+    return x & ~(y - 1);
 }
 
 // Rounds an integer up to the given power of 2. y must be a power of 2.
 constexpr int ceil2(const int x, const int y)
 {
     assert(is_power_of_2(y));
-    return (((x) + (y)-1) & (~((y)-1)));
+    return (x + y - 1) & ~(y - 1);
 }
 
 // Gets the name of a symbol as a C string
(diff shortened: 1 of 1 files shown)


ZehMatt:
The guarantee you are giving only works for debug builds, you could add a compile time check for when its passed by value however.

tupaschoal:
I can't `static_assert` the arguments, or do you mean just a simple `if` statement?

src/openrct2/common.h
```diff
@@ -50,11 +50,24 @@ const constexpr auto ror32 = ror<uint32_t>;
...
+    assert(is_power_of_2(y));
+    return ((x) & (~((y)-1)));
+}
 
 // Rounds an integer up to the given power of 2. y must be a power of 2.
-#define ceil2(x, y) (((x) + (y)-1) & (~((y)-1)))
+constexpr int ceil2(const int x, const int y)
+{
+    assert(is_power_of_2(y));
+    return (((x) + (y)-1) & (~((y)-1)));
```
janisozaur:
```suggestion
    return (x + y - 1) & (~(y - 1));
```
since this is no longer a macro, there's no need for the additional parentheses

tupaschoal:
I removed most of the parentheses, to improve readability, as these are not macros anymore

IntelOrca:
@ZehMatt I don't think you can add a compile time check unless it is a template argument. We could throw `std::logic_error` if we want it to fail in release builds though.

ZehMatt:
Well its fine, I'm just saying there are no guarantees really.

tupaschoal:
Incorporate suggestion by @IntelOrca

IntelOrca:
Should `int` be changed to a template parameter?

tupaschoal:
Dunno if you're requesting or asking for the opinion of other members, @IntelOrca

IntelOrca:
@tupaschoal sorry for the delay

PR #10316, opened 2019-12-01
Make lights work for all rotations.

duncanspumpkin:
Mistake made with the coordinates confusing viewport and screen coordinates. In addition occlusion code that is depretiated was culling all lights in two rotations due to dead code.

### DIFF
--- a/src/openrct2/drawing/LightFX.cpp
+++ b/src/openrct2/drawing/LightFX.cpp
@@ -22,10 +22,15 @@
 #    include "../world/Map.h"
 #    include "../world/Sprite.h"
 #    include "Drawing.h"
+#include "../paint/Paint.h"
+#include "../interface/Window_internal.h"
 
 #    include <algorithm>
 #    include <cmath>
 #    include <cstring>
+
+struct InteractionInfo;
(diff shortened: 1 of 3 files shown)


src/openrct2/drawing/LightFX.cpp
```diff
@@ -276,7 +272,7 @@ void lightfx_prepare_light_list()
         // clang-format on
 #    endif // LIGHTFX_UNKNOWN_PART_1
 
-        if (true)
+        if (false)
```
duncanspumpkin: This kills the last of the occlusion code from running. Will try and actually fix the occlusion code eventually.

Broxzier:
It's working much better than how it was before, though on my PC it's still very buggy. Half the screen doesn't get updated or very late, and from some rotations, the lights seem to have an offset. The lights also seem to affect the UI (probably because the palette changes for these pixels or something?)

openrct2 com_2019-12-01_22-59-07

Since this didn't work at all for me before, these issues probably already existed.

duncanspumpkin:
Can't do anything about the interference with the interface sadly. Can improve the offsets though.

Gymnasiast:
Merged because the problems Broxzier mentioned were already present before, and can be tackled in another PR.

Issue #10267, opened 2019-11-19
Refactoring Coordinates Structures: Removing LocationXY{Z}{8/16}


duncanspumpkin:
Part of the refactoring project.

This issue is to confirm that all instances of LocationXY{Z}{8/16} are removed from the general codebase. They should only be used when importing / exporting to files / network or where memory size is critical.

James103:
Can you please consider updating the "Use CoordsXY{ZD} / ScreenCoordsXY" project, moving all items whose PRs are merged to "Complete" and adding new items based on PRs related to the project?

duncanspumpkin:
There are so many and when i forget to do it in the review stage the auto close doesn't kick in. If you're looking to work on this there are still lots to do

tupaschoal:
Well. most are mine, if you want I can track down the ones we forgot to add

PR #10326, opened 2019-12-04
Use CoordsXY for rotate_map_coordinates

tupaschoal:
Helps #10267


duncanspumpkin:
I'd kind of not done this one on purpose to point out instances that should be converted to coordsxy

duncanspumpkin:
All of these should be replaced with the rotate method and the function should be depreciated

tupaschoal:
😭 
Let me do the other way around then

PR #10328, opened 2019-12-04
Use CoordsXY and ScreenCoordsXY for rotate in Ride*.*

tupaschoal:
Helps #10267

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -2899,8 +2899,8 @@ static void ride_music_update(Ride* ride)
         return;
     }
 
-    CoordsXYZ rideCoords{ ride->stations[0].Start.x * 32 + 16, ride->stations[0].Start.y * 32 + 16,
-                          ride->stations[0].Height * 8 };
+    CoordsXYZ rideCoords{ ride->stations[0].Start.x * 32, ride->stations[0].Start.y * 32, ride->stations[0].Height * 8 };
+    rideCoords = { rideCoords.ToTileCentre(), rideCoords.z };
 
     int32_t sampleRate = 22050;
 
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -2899,8 +2899,8 @@ static void ride_music_update(Ride* ride)
         return;
     }
 
-    TileCoordsXYZ rideTileCoords{ ride->stations[0].Start.x, ride->stations[0].Start.y, ride->stations[0].Height };
-    CoordsXYZ rideCoords{rideTileCoords.ToCoordsXYZ()};
+    TileCoordsXYZ stationTileCoords{ ride->stations[0].Start.x, ride->stations[0].Start.y, ride->stations[0].Height };
+    CoordsXYZ rideCoords{ stationTileCoords.ToCoordsXYZ() };
     rideCoords = { rideCoords.ToTileCentre(), rideCoords.z };
 
     int32_t sampleRate = 22050;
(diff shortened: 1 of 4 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -2899,9 +2899,8 @@ static void ride_music_update(Ride* ride)
         return;
     }
 
-    int32_t x = ride->stations[0].Start.x * 32 + 16;
-    int32_t y = ride->stations[0].Start.y * 32 + 16;
-    int32_t z = ride->stations[0].Height * 8;
+    CoordsXYZ rideCoords{ ride->stations[0].Start.x * 32 + 16, ride->stations[0].Start.y * 32 + 16,
```
Gymnasiast: This can be achieved much more cleanly by converting `ride->stations[0].Start` to `CoordsXYZ` first, then calling `ToTileCentre()`.

tupaschoal: I was unsure if I should have omitted both the multiplication and sum or just the sum, please double check on my latest commit.

Gymnasiast: I meant both. After all, it is a bog standard conversion from TileCoords to Coords.

There should be a method to create `Coords` from `TileCoords` (and if there isn't, there should be one).

tupaschoal: Done. These weren't really `TileCoords` and I didn't want to change them here, as they might propagate a lot, but though it doesn't look ideal, I created the method and it's cleaner anyway. Thanks for the suggestion!

src/openrct2/world/Location.hpp
```diff
@@ -273,6 +273,11 @@ struct TileCoordsXYZ
         return !(*this == other);
     }
 
+    CoordsXYZ ToCoordsXYZ()
+    {
+        return { x * 32, y * 32, z * 8 };
+    }
+
```
Gymnasiast: What I meant is that you had to create a method like the one on line 248 (with TileCoords and Coords reversed).

tupaschoal: But that will create a circular dependency between both

Gymnasiast: Hm, that's where my knowledge of C++ hits its limits. @duncanspumpkin could you comment?

src/openrct2/ride/Ride.cpp
```diff
@@ -2899,9 +2899,9 @@ static void ride_music_update(Ride* ride)
         return;
     }
 
-    int32_t x = ride->stations[0].Start.x * 32 + 16;
-    int32_t y = ride->stations[0].Start.y * 32 + 16;
-    int32_t z = ride->stations[0].Height * 8;
+    TileCoordsXYZ rideTileCoords{ ride->stations[0].Start.x, ride->stations[0].Start.y, ride->stations[0].Height };
```
duncanspumpkin: `stationTileCoords`

src/openrct2/ride/Ride.cpp
```diff
@@ -7658,12 +7657,10 @@ StationObject* ride_get_station_object(const Ride* ride)
     return static_cast<StationObject*>(objManager.GetLoadedObject(OBJECT_TYPE_STATION, ride->entrance_style));
 }
 
-LocationXY16 ride_get_rotated_coords(int16_t x, int16_t y, int16_t z)
+ScreenCoordsXY ride_get_rotated_coords(CoordsXYZ coords3d)
```
duncanspumpkin: This function is now just a wrapper for `translate_3d_to_2d_with_z`. Its a little pointless.

src/openrct2/world/Location.hpp
duncanspumpkin: In addition we were trying to avoid converting up and down between the types. That is why there wasn't anything like this before and why the constructor is explicit. I think this is fine but it should be discouraged from use.

Gymnasiast: @duncanspumpkin Yes, but would it be possible to add an explicit constructor to create `Coords` from `TileCoords`?

duncanspumpkin: You wouldn't be able to resolve CoordsXY. So yes agree with @tupaschoal its not possible.

tupaschoal:
Dealt with issues

PR #10329, opened 2019-12-04
Use CoordsXY on TopToolbar and TrackDesignPlace rotations

tupaschoal:
Helps #10267

So my goal was to just remove the usage of `LocationXY` from `TopToolbar` and `TrackDesignPlace`, but then to avoid casting like crazy, I had to also update `map_offset_with_rotation`.

I made it return `TileCoordsXY` as it was what was implemented inside, but it's very confusing because users on the outside sometimes use this type, and at other times, `CoordsXY`, so I'd appreciate some feedback on whether this is one of the cases where it's mixing up and shouldn't.

### DIFF
--- a/src/openrct2-ui/windows/TrackDesignPlace.cpp
+++ b/src/openrct2-ui/windows/TrackDesignPlace.cpp
@@ -125,9 +125,9 @@ static void window_track_place_attempt_placement(
 static void window_track_place_clear_mini_preview();
 static void window_track_place_draw_mini_preview(TrackDesign* td6);
 static void window_track_place_draw_mini_preview_track(
-    TrackDesign* td6, int32_t pass, LocationXY16 origin, LocationXY16* min, LocationXY16* max);
+    TrackDesign* td6, int32_t pass, CoordsXY origin, CoordsXY& min, CoordsXY& max);
 static void window_track_place_draw_mini_preview_maze(
-    TrackDesign* td6, int32_t pass, LocationXY16 origin, LocationXY16* min, LocationXY16* max);
+    TrackDesign* td6, int32_t pass, CoordsXY origin, CoordsXY& min, CoordsXY& max);
 static LocationXY16 draw_mini_preview_get_pixel_position(int16_t x, int16_t y);
 static bool draw_mini_preview_is_pixel_in_bounds(LocationXY16 pixel);
 static uint8_t* draw_mini_preview_get_pixel_ptr(LocationXY16 pixel);
@@ -502,11 +502,11 @@ static void window_track_place_draw_mini_preview(TrackDesign* td6)
     window_track_place_clear_mini_preview();
 
     // First pass is used to determine the width and height of the image so it can centre it
-    LocationXY16 min = { 0, 0 };
-    LocationXY16 max = { 0, 0 };
+    CoordsXY min = { 0, 0 };
+    CoordsXY max = { 0, 0 };
     for (int32_t pass = 0; pass < 2; pass++)
     {
-        LocationXY16 origin = { 0, 0 };
+        CoordsXY origin = { 0, 0 };
         if (pass == 1)
         {
             origin.x -= ((max.x + min.x) >> 6) * 32;
@@ -515,17 +515,17 @@ static void window_track_place_draw_mini_preview(TrackDesign* td6)
 
         if (td6->type == RIDE_TYPE_MAZE)
         {
-            window_track_place_draw_mini_preview_maze(td6, pass, origin, &min, &max);
+            window_track_place_draw_mini_preview_maze(td6, pass, origin, min, max);
         }
         else
         {
-            window_track_place_draw_mini_preview_track(td6, pass, origin, &min, &max);
+            window_track_place_draw_mini_preview_track(td6, pass, origin, min, max);
         }
     }
 }
 
 static void window_track_place_draw_mini_preview_track(
-    TrackDesign* td6, int32_t pass, LocationXY16 origin, LocationXY16* min, LocationXY16* max)
+    TrackDesign* td6, int32_t pass, CoordsXY origin, CoordsXY& min, CoordsXY& max)
 {
     uint8_t rotation = (_currentTrackPieceDirection + get_current_rotation()) & 3;
 
@@ -543,20 +543,20 @@ static void window_track_place_draw_mini_preview_track(
         const rct_preview_track* trackBlock = trackBlockArray[trackType];
         while (trackBlock->index != 255)
         {
-            int16_t x = origin.x;
-            int16_t y = origin.y;
-            map_offset_with_rotation(&x, &y, trackBlock->x, trackBlock->y, rotation);
+            auto rotatedAndOffsetTrackBlock = map_offset_with_rotation(
+                { origin.x, origin.y }, { trackBlock->x, trackBlock->y }, rotation);
 
             if (pass == 0)
             {
-                min->x = std::min(min->x, x);
-                max->x = std::max(max->x, x);
-                min->y = std::min(min->y, y);
-                max->y = std::max(max->y, y);
+                min.x = std::min(min.x, rotatedAndOffsetTrackBlock.x);
+                max.x = std::max(max.x, rotatedAndOffsetTrackBlock.x);
+                min.y = std::min(min.y, rotatedAndOffsetTrackBlock.y);
+                max.y = std::max(max.y, rotatedAndOffsetTrackBlock.y);
             }
             else
             {
-                LocationXY16 pixelPosition = draw_mini_preview_get_pixel_position(x, y);
+                LocationXY16 pixelPosition = draw_mini_preview_get_pixel_position(
+                    rotatedAndOffsetTrackBlock.x, rotatedAndOffsetTrackBlock.y);
                 if (draw_mini_preview_is_pixel_in_bounds(pixelPosition))
                 {
                     uint8_t* pixel = draw_mini_preview_get_pixel_ptr(pixelPosition);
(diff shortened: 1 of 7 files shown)


src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -543,20 +543,20 @@ static void window_track_place_draw_mini_preview_track(
         const rct_preview_track* trackBlock = trackBlockArray[trackType];
         while (trackBlock->index != 255)
         {
-            int16_t x = origin.x;
-            int16_t y = origin.y;
-            map_offset_with_rotation(&x, &y, trackBlock->x, trackBlock->y, rotation);
+            auto rotatedAndOffsetTrackBlock = map_offset_with_rotation(
```
duncanspumpkin: This isn't correct a `rct_preview_track*` coordinates are in `CoordsXY`. I think this would be simplified if we did the following:

```
CoordsXY trackCoords = origin + CoordsXY{trackBlock->x, trackBlock->y}.Rotate(rotation);
```

src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -589,7 +589,8 @@ static void window_track_place_draw_mini_preview_track(
         const rct_track_coordinates* track_coordinate = &TrackCoordinates[trackType];
 
         trackType *= 10;
-        map_offset_with_rotation(&origin.x, &origin.y, track_coordinate->x, track_coordinate->y, rotation);
+        auto rotatedAndOfffsetTrack = map_offset_with_rotation(
```
duncanspumpkin: As above. Do the sum and rotate directly.

src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -598,28 +599,27 @@ static void window_track_place_draw_mini_preview_track(
         }
         if (!(rotation & 4))
         {
-            origin.x += CoordsDirectionDelta[rotation].x;
-            origin.y += CoordsDirectionDelta[rotation].y;
+            origin.x = rotatedAndOfffsetTrack.x + CoordsDirectionDelta[rotation].x;
+            origin.y = rotatedAndOfffsetTrack.x + CoordsDirectionDelta[rotation].y;
```
duncanspumpkin: When you fix the type this can be just a simple `+`. Also note you've done .x for the .y coordinate.

src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -598,28 +599,27 @@ static void window_track_place_draw_mini_preview_track(
...
         }
     }
 
     // Draw entrance and exit preview.
     for (const auto& entrance : td6->entrance_elements)
     {
-        int16_t x = origin.x;
-        int16_t y = origin.y;
-        map_offset_with_rotation(&x, &y, entrance.x, entrance.y, rotation);
+        auto rotatedAndOffsetEntrance = map_offset_with_rotation({ origin.x, origin.y }, { entrance.x, entrance.y }, rotation);
```
duncanspumpkin: As above. Do the sum and rotate directly.

src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -637,28 +637,26 @@ static void window_track_place_draw_mini_preview_track(
...
+    TrackDesign* td6, int32_t pass, CoordsXY origin, CoordsXY& min, CoordsXY& max)
 {
     uint8_t rotation = (_currentTrackPieceDirection + get_current_rotation()) & 3;
     for (const auto& mazeElement : td6->maze_elements)
     {
-        int16_t x = mazeElement.x * 32;
-        int16_t y = mazeElement.y * 32;
-        rotate_map_coordinates(&x, &y, rotation);
+        CoordsXY mazeCoords{ mazeElement.x * 32, mazeElement.y * 32 };
+        auto rotatedMazeCoords = mazeCoords.Rotate(rotation);
```
duncanspumpkin: May as well do the addition as part of this as well to be consistent with the others.

src/openrct2/ride/Ride.cpp
```diff
@@ -6977,8 +6977,9 @@ void sub_6CB945(Ride* ride)
             const rct_preview_track* trackBlock = get_track_def_from_ride(ride, tileElement->AsTrack()->GetTrackType());
             while ((++trackBlock)->index != 0xFF)
             {
-                LocationXYZ16 blockLocation = location;
-                map_offset_with_rotation(&blockLocation.x, &blockLocation.y, trackBlock->x, trackBlock->y, direction);
+                auto blockLocationXY = map_offset_with_rotation(
```
duncanspumpkin: As per the other functions just do the sum and rotate directly. You should be able to update the type of location whilst you are at it to CoordsXYZ.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1472,8 +1472,7 @@ static bool track_design_place_ride(TrackDesign* td6, int16_t x, int16_t y, int1
             case PTD_OPERATION_DRAW_OUTLINES:
                 for (const rct_preview_track* trackBlock = trackBlockArray[trackType]; trackBlock->index != 0xFF; trackBlock++)
                 {
-                    LocationXY16 tile = { x, y };
-                    map_offset_with_rotation(&tile.x, &tile.y, trackBlock->x, trackBlock->y, rotation);
+                    auto tile = map_offset_with_rotation({ x, y }, { trackBlock->x, trackBlock->y }, rotation);
```
duncanspumpkin: As above. Do the sum and rotate directly.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1552,16 +1551,13 @@ static bool track_design_place_ride(TrackDesign* td6, int16_t x, int16_t y, int1
                 int32_t tempZ = z - TrackCoordinates[trackType].z_begin;
                 for (const rct_preview_track* trackBlock = trackBlockArray[trackType]; trackBlock->index != 0xFF; trackBlock++)
                 {
-                    int16_t tmpX = x;
-                    int16_t tmpY = y;
-                    map_offset_with_rotation(&tmpX, &tmpY, trackBlock->x, trackBlock->y, rotation);
-                    CoordsXY tile = { tmpX, tmpY };
+                    auto tile = map_offset_with_rotation({ x, y }, { trackBlock->x, trackBlock->y }, rotation);
```
duncanspumpkin: As above. Do the sum and rotate directly.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1552,16 +1551,13 @@ static bool track_design_place_ride(TrackDesign* td6, int16_t x, int16_t y, int1
...
                     if (tile.x < 0 || tile.y < 0 || tile.x >= (256 * 32) || tile.y >= (256 * 32))
                     {
                         continue;
                     }
 
-                    auto surfaceElement = map_get_surface_element_at(tile);
+                    auto surfaceElement = map_get_surface_element_at({ tile.x, tile.y });
```
duncanspumpkin: Undo when you use the correct type.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1593,7 +1589,10 @@ static bool track_design_place_ride(TrackDesign* td6, int16_t x, int16_t y, int1
         }
 
         const rct_track_coordinates* track_coordinates = &TrackCoordinates[trackType];
-        map_offset_with_rotation(&x, &y, track_coordinates->x, track_coordinates->y, rotation);
+        auto offsetAndRotatedTrack = map_offset_with_rotation(
```
duncanspumpkin: As above. Do the sum and rotate directly.

src/openrct2/world/Map.cpp
```diff
@@ -2353,14 +2353,10 @@ TileElement* map_get_track_element_at_with_direction_from_ride(
     return nullptr;
 };
 
-void map_offset_with_rotation(int16_t* x, int16_t* y, int16_t offsetX, int16_t offsetY, uint8_t rotation)
+TileCoordsXY map_offset_with_rotation(TileCoordsXY position, TileCoordsXY offset, uint8_t rotation)
```
duncanspumpkin: Types are all CoordsXY

src/openrct2/world/Map.h
```diff
@@ -231,7 +231,7 @@ bool map_large_scenery_get_origin(
     int32_t x, int32_t y, int32_t z, int32_t direction, int32_t sequence, int32_t* outX, int32_t* outY, int32_t* outZ,
     LargeSceneryElement** outElement);
 
-void map_offset_with_rotation(int16_t* x, int16_t* y, int16_t offsetX, int16_t offsetY, uint8_t rotation);
+TileCoordsXY map_offset_with_rotation(TileCoordsXY position, TileCoordsXY offset, uint8_t rotation);
```
duncanspumpkin: Types are all CoordsXY

tupaschoal:
Fixed all issues and removed `map_offset_with_rotation`

src/openrct2/ride/Ride.cpp
```diff
@@ -6915,11 +6915,8 @@ void sub_6CB945(Ride* ride)
...
-            LocationXYZ16 location = {
-                (int16_t)(ride->stations[stationId].Start.x * 32),
-                (int16_t)(ride->stations[stationId].Start.y * 32),
-                (ride->stations[stationId].Height),
-            };
+            CoordsXYZ location = { ride->stations[stationId].Start.x * 32, ride->stations[stationId].Start.y * 32,
+                                   ride->stations[stationId].Height * 8 };
             uint8_t direction = 0xFF;
 
             bool specialTrack = false;
```
duncanspumpkin: you will need to check the rest of the function for instances of location.z to make sure they take into account the change to it.

tupaschoal: Solved. I used the ctor of `TileCoordsXYZ(CoordsXYZ)` itself to abstract away the divisions by `8` so that all the logic is on the Location objects themselves.

PR #10331, opened 2019-12-05
add debian package list for Linux compiling to readme.md

lightray22:
Added an example package list to the Linux prerequisites section.  A corresponding "brew install" line already exists in the macOS section.  Finding the list of actual packages takes some time for Linux novices.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -117,6 +117,11 @@ The program can also be built as a command line program using CMake. This type o
 - gl (commonly provided by Mesa or GPU vendors; only for UI client, can be disabled)
 - cmake
 
+For example on a Debian-based system:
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
As you have probably already surmised, such a list of packages is already present on our wiki, which is linked in the relevant section of the readme file.

IntelOrca:
The line should go under compiling if it is to be like the brew install line. The problem though is that if we add this then we would be expected to have all the distros which would bloat the readme. I wouldn't mind this being in a separate file with more detailed instructions, as @AaronVanGeffen pointed out, there is already a wiki page as well.

PR #10333, opened 2019-12-05
Use CoordsXY for Paint rotations

tupaschoal:
Helps #10267

### DIFF
--- a/src/openrct2/paint/Paint.cpp
+++ b/src/openrct2/paint/Paint.cpp
@@ -71,8 +71,7 @@ static void paint_session_add_ps_to_quadrant(paint_session* session, paint_struc
  * Extracted from 0x0098196c, 0x0098197c, 0x0098198c, 0x0098199c
  */
 static paint_struct* sub_9819_c(
-    paint_session* session, uint32_t image_id, const CoordsXYZ& offset, LocationXYZ16 boundBoxSize,
-    LocationXYZ16 boundBoxOffset)
+    paint_session* session, uint32_t image_id, const CoordsXYZ& offset, CoordsXYZ boundBoxSize, CoordsXYZ boundBoxOffset)
 {
     if (session->NextFreePaintStruct >= session->EndOfPaintStructArray)
         return nullptr;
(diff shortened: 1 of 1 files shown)


src/openrct2/paint/Paint.cpp
```diff
@@ -891,9 +890,9 @@ paint_struct* sub_98197C(
 
     session->LastRootPS = ps;
 
-    LocationXY16 attach = { (int16_t)ps->bounds.x, (int16_t)ps->bounds.y };
+    CoordsXY attach = { ps->bounds.x, ps->bounds.y };
 
-    rotate_map_coordinates(&attach.x, &attach.y, session->CurrentRotation);
+    attach = attach.Rotate(session->CurrentRotation);
```
duncanspumpkin: You could just do `CoordsXY attach = CoordsXY{ ps->bounds.x, ps->bounds.y }.Rotate(session->CurrentRotation);`
Not sure what is better readability wise.

duncanspumpkin:
Was there any measuarble decrease in performance with this change? This is a function that is called a lot.

src/openrct2/paint/Paint.cpp
tupaschoal: I think this line is small enough for this to be ok, so I'm following your suggestion.

tupaschoal:
> Was there any measuarble decrease in performance with this change? This is a function that is called a lot.

I didn't measure, really. How should I?

duncanspumpkin:
There is a benchmark command but I can't remember it. @ZehMatt could you check to see if there is any change in performance?

ZehMatt:
I can and I did,
develop:
```
>openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.059566s, 17 FPS
Zoom[1] average: 0.027072s, 37 FPS
Zoom[2] average: 0.012890s, 78 FPS
Total average: 0.033176s, 30 FPS
Time: 1.99055s
```
rotate-paint:
```
>openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.059929s, 17 FPS
Zoom[1] average: 0.027087s, 37 FPS
Zoom[2] average: 0.012765s, 78 FPS
Total average: 0.033260s, 30 FPS
Time: 1.99562s
```
No big difference really, super happy that benchgfx is now reliable, not so happy about CoordsXY::Rotate, I think there should be two versions such as: ```CoordsXY CoordsXY::GetRotated``` and ```void CoordsXY::Rotate```, the temporary should be avoided imho but not for this PR obviously.

PR #10334, opened 2019-12-05
Use CoordsXY for TrackDesign* rotations

tupaschoal:
Helps #10267

### DIFF
--- a/src/openrct2/ride/TrackDesign.cpp
+++ b/src/openrct2/ride/TrackDesign.cpp
@@ -277,10 +277,9 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
                 continue;
             }
 
-            int16_t x = location.x * 32;
-            int16_t y = location.y * 32;
+            CoordsXY mapLocation{ location.x * 32, location.y * 32 };
 
-            TileElement* tileElement = map_get_first_element_at(x >> 5, y >> 5);
+            TileElement* tileElement = map_get_first_element_at(location.x, location.y);
             if (tileElement == nullptr)
                 continue;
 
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/TrackDesign.cpp
```diff
@@ -277,10 +277,9 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
                 continue;
             }
 
-            int16_t x = location.x * 32;
-            int16_t y = location.y * 32;
+            CoordsXY mapLocation{ location.x * 32, location.y * 32 };
 
-            TileElement* tileElement = map_get_first_element_at(x >> 5, y >> 5);
+            TileElement* tileElement = map_get_first_element_at(location.x, location.y);
```
duncanspumpkin: This is wrong. It still needs the `/32`

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -301,13 +300,13 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
...
-            x -= gTrackPreviewOrigin.x;
-            y -= gTrackPreviewOrigin.y;
+            mapLocation.x -= gTrackPreviewOrigin.x;
+            mapLocation.y -= gTrackPreviewOrigin.y;
 
             // Rotate entrance coordinates backwards to the correct direction
-            rotate_map_coordinates(&x, &y, (0 - _saveDirection) & 3);
-            entrance.x = x;
-            entrance.y = y;
+            auto rotatedMapLocation = mapLocation.Rotate(0 - _saveDirection);
```
duncanspumpkin: I think you will still need `& 3` otherwise you will throw an assert.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -535,19 +534,17 @@ rct_string_id TrackDesign::CreateTrackDesignScenery()
             }
         }
 
-        int16_t x = ((uint8_t)scenery.x) * 32 - gTrackPreviewOrigin.x;
-        int16_t y = ((uint8_t)scenery.y) * 32 - gTrackPreviewOrigin.y;
-        rotate_map_coordinates(&x, &y, (0 - _saveDirection) & 3);
-        x /= 32;
-        y /= 32;
+        CoordsXY sceneryMapPos{ scenery.x * 32 - gTrackPreviewOrigin.x, scenery.y * 32 - gTrackPreviewOrigin.y };
+        CoordsXY rotatedSceneryMapPos = sceneryMapPos.Rotate(0 - _saveDirection);
```
duncanspumpkin: again `& 3`

src/openrct2/ride/TrackDesign.cpp
tupaschoal: Does it? `location.x` and `location.y` were being multiplied by `32` and saved into `{x, y` just to then be shifted left by `5`, which takes away exactly the 32. I just started using them directly.

tupaschoal:
I don't think we should revert the removal of the `& 3`, this is done inside the `Rotate` function and is inner logic that the user shouldn't need to care about, in my opinion:
https://github.com/OpenRCT2/OpenRCT2/blob/137d47a3759d6a236915fae3d59ba98f0740abb6/src/openrct2/world/Location.hpp#L116-L119

About the "removal of the /32", it wasn't really removed, if I'm correct, so there won't be a thing to change either.

src/openrct2/ride/TrackDesign.cpp
duncanspumpkin: ah right you are. I read that as `mapLocation` not `location`. This is fine.

PR #10341, opened 2019-12-06
Scale shortcut

0Cufox0:
it functions normally, only thing that's missing is the shortcut name in the shortCutBindingList, right now it's empty, I could add the text to eng-gb.txt but I thought that localization would take of that, do you think I should add it?


IntelOrca:
Yes always add new strings to en-GB.txt, it is only the other languages that are updated via the localisation repository.

duncanspumpkin:
Just needs a clang format and this will be good to go. In future stick to the commit message format. https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

AaronVanGeffen:
In addition to the clang-format issues, please restore the whitespace surrounding the strings. It's there for clarity, and I'd prefer it to remain. :)

src/openrct2/network/Network.cpp
```diff
@@ -31,7 +31,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "2"
+#define NETWORK_STREAM_VERSION "3"
```
janisozaur: Unnecessary, please revert this change. This only needs changing when the change could cause two clients to run different logic and lead to them inferring different state from same input.

0Cufox0: Alright! thx for the info! btw is there any resources I can read to get a better understanding of how everything works,apart from the source code of course.

duncanspumpkin:
Argh could you also add an entry to the changelog.

0Cufox0:
> Argh could you also add an entry to the changelog.

alright! don't worry Ima read the wiki in a bit to stop making these stupid mistakes :>

distribution/changelog.txt
```diff
@@ -3,7 +3,7 @@
 - Change: [#1164] Use available translations for shortcut key bindings.
 - Fix: [#10228] Can't import RCT1 Deluxe from Steam.
 - Fix: [#10325] Crash when banners have no text.
-
+- Feature: [#10305] Add two shortcuts for increasing and decreasing the sacling factor.
```
janisozaur: Leave a blank line between versions. Also, there's a typo, "sacling"

distribution/changelog.txt
```diff
@@ -3,6 +3,7 @@
 - Change: [#1164] Use available translations for shortcut key bindings.
 - Fix: [#10228] Can't import RCT1 Deluxe from Steam.
 - Fix: [#10325] Crash when banners have no text.
+- Feature: [#10305] Add two shortcuts for increasing and decreasing the scaling factor.
```
AaronVanGeffen: Features should be placed at the top of the changelog.

PR #10346, opened 2019-12-08
Name field var_CD of vehicle struct. Refactor to enum

modulo7:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -764,73 +764,73 @@ static const struct
 
 // clang-format on
 
-static bool vehicle_move_info_valid(int32_t cd, int32_t typeAndDirection, int32_t offset)
+static bool vehicle_move_info_valid(int32_t track_status, int32_t typeAndDirection, int32_t offset)
```
Gymnasiast: Use camelCase for variable names.

Gymnasiast:
What is the reasoning behind calling it "track status"?

src/openrct2/ride/Vehicle.h
```diff
@@ -243,7 +264,7 @@ struct rct_vehicle : rct_sprite_common
     uint16_t var_C8;
     uint16_t var_CA;
     SoundId scream_sound_id; // 0xCC
-    uint8_t var_CD;
+    VEHICLE_TRACK_STATUS track_status;// 0xCD
```
duncanspumpkin: In general we haven't been changing the type except when its a scoped enum.

duncanspumpkin:
I don't think track status is really a great name. Its used by a couple different things for slightly different purposes. The chairlift ones for example are to tell it which side of the cable it is. The go karts are related to where they are on the track tile but also used for working out overtaking. Boats might also use this variable for something. Reversers are using it for the bogies position. I just don't think track status really describes this. Its related to the vehicles position on the track.

Gymnasiast:
@duncanspumpkin How about `trackPosition`, then?

duncanspumpkin:
`trackPosition` suggests the X, Y coordinates and would be a better name for the already available `track_x`, `track_y`, `track_z` variables. Perhaps `TrackSubPosition` with a note that this is for special case rides (Chairlifts, GoKarts, Reversers)

Gymnasiast:
Continuing here: https://github.com/OpenRCT2/OpenRCT2/pull/10584

PR #10356, opened 2019-12-11
Change the label on known var to reflect use.

duncanspumpkin:
Just a minor refactoring. 
Simplified the logic as vars not used

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2326,9 +2326,9 @@ static void window_ride_construction_paint(rct_window* w, rct_drawpixelinfo* dpi
         return;
 
     ride_id_t rideIndex;
-    int32_t trackType, trackDirection, liftHillAndAlternativeState;
+    int32_t trackType, trackDirection, liftHillAndInvertedState;
     if (window_ride_construction_update_state(
-            &trackType, &trackDirection, &rideIndex, &liftHillAndAlternativeState, nullptr, nullptr, nullptr, nullptr))
+            &trackType, &trackDirection, &rideIndex, &liftHillAndInvertedState, nullptr, nullptr, nullptr, nullptr))
         return;
 
     // Draw track piece
@@ -2339,7 +2339,7 @@ static void window_ride_construction_paint(rct_window* w, rct_drawpixelinfo* dpi
     if (clip_drawpixelinfo(&clipdpi, dpi, x, y, width, height))
     {
         window_ride_construction_draw_track_piece(
-            w, &clipdpi, rideIndex, trackType, trackDirection, liftHillAndAlternativeState, width, height);
+            w, &clipdpi, rideIndex, trackType, trackDirection, liftHillAndInvertedState, width, height);
     }
 
     // Draw cost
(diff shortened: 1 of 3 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2456,15 +2453,15 @@ static void sub_6CBCE2(
         // Set the temporary track element
         _tempTrackTileElement.SetType(TILE_ELEMENT_TYPE_TRACK);
         _tempTrackTileElement.SetDirection(trackDirection);
-        _tempTrackTileElement.AsTrack()->SetHasChain((edx & 0x10000) != 0);
+        _tempTrackTileElement.AsTrack()->SetHasChain((liftHillAndInvertedState & CONSTRUCTION_LIFT_HILL_SELECTED) != 0);
```
Gymnasiast: You changed the value here from 0x10000 to 0x1. Are all callers of this function also modified?

duncanspumpkin: yes thats the point of the change.

src/openrct2/ride/Ride.h
```diff
@@ -993,6 +993,7 @@ struct rct_ride_properties
 #define TURN_MASK_4_PLUS_ELEMENTS 0xF800
 
 #define CONSTRUCTION_LIFT_HILL_SELECTED 1
+#define CONSTRUCTION_INVERTED_TRACK_SELECTED 2
```
tupaschoal: Why not a `constexpr int32_t` instead?

src/openrct2-ui/windows/RideConstruction.cpp
Gymnasiast: I didn't see any changed callers to `sub_6CBCE2`, so I was wondering if it was complete.

src/openrct2/ride/Ride.h
Gymnasiast: Should also use the `1 << 1` or `0b00000001` notation, I think.

src/openrct2-ui/windows/RideConstruction.cpp
duncanspumpkin: `window_ride_construction_draw_track_piece` is the only caller and it used to stuff a load of unused information in that var as well. I've removed the unneeded information and removed the << 16.

src/openrct2/ride/Ride.h
```diff
@@ -992,7 +992,8 @@ struct rct_ride_properties
 #define TURN_MASK_3_ELEMENTS 0x0700
 #define TURN_MASK_4_PLUS_ELEMENTS 0xF800
 
-#define CONSTRUCTION_LIFT_HILL_SELECTED 1
+constexpr uint32_t CONSTRUCTION_LIFT_HILL_SELECTED = 1 << 0;
+constexpr uint32_t CONSTRUCTION_INVERTED_TRACK_SELECTED = 1 << 1;
```
tupaschoal: Are they really unsigned? Because on all places they're being compared to `int32_t`

duncanspumpkin: I think we tend to use unsigned for flags for edge cases at 1 << 31

Gymnasiast: I think using unsigned for flags makes most sense. After all, there is no such thing as a "negative flag".

Issue #10246, opened 2019-11-15
Resizable Tree Scatter Tool


dr-dirt:
Request for making the treeing / tree scatter tool resizable; currently is only a single size.  Suggestion made to Gymnasiast.

scrapes:
Hey i tried making something out of fun:
https://github.com/scrapes/OpenRCT2/tree/scenery-scatter

Added an extra Window with Scatter options if you activate the Scenery-Scatter(Cluster) option. 
Some feedback would be nice :)

AaronVanGeffen:
@scrapes Feel free to open a PR for your branch. Makes it easier for us to give feedback on particular bits of code. :)

scrapes:
Alright, will do it as soon as I can!

Umdlye:
Implemented by #10357

Issue #7504, opened 2018-05-11
Quarter tile scenery placement on water differs from full size


Gymnasiast:
As noted in https://github.com/OpenRCT2/OpenRCT2/pull/5911 , when placing quarter tile scenery, one can place it directly on water without holding SHIFT, and it will not have a support structure.

This is different from regular scenery, but might be intended behaviour - it allows placing reed at water sides without a support structure spoiling the view.

ocalhoun6:
This also causes problems when placing quarter-tile scenery in random batches. They can be randomly placed over water, when full-square scenery items will automatically avoid placing themselves in water.

Example:
![leafy lake 2018-05-11 14-01-59](https://user-images.githubusercontent.com/26573303/39946667-ed2edf10-5523-11e8-8d17-f2625a1a0583.png)

PR #10357, opened 2019-12-11
Scenery-Scatter Tool Window

scrapes:
Scenery-Scatter Tool Window (WIP).

Hey i build a Tool Window For the Scenery Scatter Tool to change size and amount of scenery.
Maybe someone can take a look at it and tell me what i have to improve.
This is my first OpenSource Implementation.

This references Issue #10246 
Attached an Image for Quick design feedback.
SceneryScatter
The Tool Window Title is not Tree Scatter but Scatter tool as seen in the Commit.


ocalhoun6:
Maybe call it something like "scenery scatter" or "scatter tool", since it can be used for things other than trees?

And while I'm sure most people will figure it out quickly, the 'amount' input is a bit mysterious at first glance if you don't already know what it is and what it does.

-----------------

Also, at the risk of scope creep: This new tool window could be an excellent place to place a check-box control that lets you turn on or off placing the scenery items on the surface of water, and that would resolve things like issue #7504

scrapes:
> Maybe call it something like "scenery scatter" or "scatter tool", since it can be used for things other than trees?
> 
> And while I'm sure most people will figure it out quickly, the 'amount' input is a bit mysterious at first glance if you don't already know what it is and what it does. Might be a good idea to label it in the GUI. (Especially if the size field is increased to the point where it starts showing numbers rather than squares, both controls will look the same, making it hard to remember which does what.)
> 
> Also, at the risk of scope creep: This new tool window could be an excellent place to place a check-box control that lets you turn on or off placing the scenery items on the surface of water, and that would resolve things like issue #7504
> 
> (If you don't feel like addressing that part now, I could always just submit it as a new feature request after this PR goes through.)

Like mentioned in the PR the Tool itself called Scatter Tool or Scenery Scatter. The screenshot has just a misleading title. 
The labeling is a good idea! I was thinking about make the two selections on the same horizontal pane but I think that just adds more confusion to it. 

The other question is how I change the Windows background color. I looked in the scenery window file and found no reference to any background color besides "1" and "0". 

When I am done to the point it's release ready I will see if I address issue #7504 also.

Thanks for the feedback so far i am having a great first contribution experience!

Gymnasiast:
You're using tabs, but our codebase uses spaces. In general, the `clang-format` CI fails.

You can use the `clang-format` tool in order to autoformat your code. It's also possible to do `git clang-format` on all files you `git add`ed. (You will then need to `git add` them again.)

Two other points, which are optional:
- You haven't set your email address correctly in your local Git settings, so Github doesn't recognise your commits.
- We always rebase and never merge the develop branch into a feature branch. Since this is your first PR, we can handle the rebase if you're unfamiliar with Git (let's concentrate on the code itself first).

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -19,7 +19,7 @@
 #include <openrct2/localisation/Localisation.h>
 #include <openrct2/management/Research.h>
 #include <openrct2/network/network.h>
-#include <openrct2/object/ObjectList.h>
+#include <openrct2/object/ObjectList.h>  // Unused?
```
Gymnasiast: This is not related to your PR, so please revert it.

Also, this kind of comment is not very useful: if there is any doubt a header is used, it should be checked and removed if it isn't used.

scrapes: Deleted Comment

scrapes:
> You're using tabs, but our codebase uses spaces. In general, the `clang-format` CI fails.
> 
> You can use the `clang-format` tool in order to autoformat your code. It's also possible to do `git clang-format` on all files you `git add`ed. (You will then need to `git add` them again.)
> 
> Two other points, which are optional:
> 
> * You haven't set your email address correctly in your local Git settings, so Github doesn't recognise your commits.
> * We always rebase and never merge the develop branch into a feature branch. Since this is your first PR, we can handle the rebase if you're unfamiliar with Git (let's concentrate on the code itself first).

I have used clang-format on all modified files. The format tool formats a big chunk of the TopToolbar.cpp from 
´´´
case x:
   {
´´´
to
´´´
case x:  {
´´´

i didn´t commit those chagens because i dont know if these were intentional or not. 
I had my name and email added to the git config but CLion seems to not add these Information in the Commits itself, i fixed this and updated my email settings on GitHub.

I reworked the design a little bit: 
SceneScatter

scrapes:
I also apologize for the merge rather than the rebase. I will only do rebases from now on.

AaronVanGeffen:
> I have used clang-format on all modified files. The format tool formats a big chunk of the TopToolbar.cpp

If there are differences in formatting in chunks of code that you did not change, you are likely not using the same version of `clang-format` as we are in our CI job. (Which version are you using?)

scrapes:
I am currently on clang-format version 9.0.0 installed from brew

scrapes:
> > I have used clang-format on all modified files. The format tool formats a big chunk of the TopToolbar.cpp
> 
> If there are differences in formatting in chunks of code that you did not change, you are likely not using the same version of `clang-format` as we are in our CI job. (Which version are you using?)

i used the docker command from the ci and i seem to get no errors. Rebased the branch and pushed the new string-ids and the last format error.

Gymnasiast:
I will be reviewing this later, but some quick points:

- Sorry this is taking so long - many of us are quite busy.
- One CI came back red, but that was a time out, so I restarted it and it should be green once it finishes
- I see you have access to Xcode! That is very nice, because adding files to Xcode is always a major problem for us. (Hope you stay with the project for a while).
- I'm not sold on the window layout and text - it doesn't look very RCTy and it's also quite cluttered. It needs further thinking or discussion, so don't change it just yet.

scrapes:
> I will be reviewing this later, but some quick points:
> 
> * Sorry this is taking so long - many of us are quite busy.
> * One CI came back red, but that was a time out, so I restarted it and it should be green once it finishes
> * I see you have access to Xcode! That is very nice, because adding files to Xcode is always a major problem for us. (Hope you stay with the project for a while).
> * I'm not sold on the window layout and text - it doesn't look very RCTy and it's also quite cluttered. It needs further thinking or discussion, so don't change it just yet.

No Problem! I can also see what you are meaning with the design, it was also not so good in my eyes.
I will try to stay a bit on the Project!

Gymnasiast:
Could you rebase this PR?

src/openrct2-ui/windows/SceneryScatter.cpp
```diff
@@ -0,0 +1,255 @@
...
+static void window_scenery_scatter_mouseup(rct_window *w, rct_widgetindex widgetIndex);
+static void window_scenery_scatter_mousedown(rct_window *w, rct_widgetindex widgetIndex, rct_widget *widget);
+static void window_scenery_scatter_update(rct_window *w);
+static void window_scenery_scatter_invalidate(rct_window *w);
+static void window_scenery_scatter_paint(rct_window *w, rct_drawpixelinfo *dpi);
+static void window_scenery_scatter_textinput(rct_window *w, rct_widgetindex widgetIndex, char *text);
+static void window_scenery_scatter_inputsize(rct_window* w, rct_widgetindex widgetindex);
+
+static rct_window_event_list window_clear_scenery_events = {
+		window_scenery_scatter_close,
```
AaronVanGeffen:
```suggestion
		nullptr,
```

src/openrct2-ui/windows/SceneryScatter.cpp
```diff
@@ -0,0 +1,255 @@
...
+    window->hold_down_widgets = (1 << WIDX_INCREMENT) | (1 << WIDX_DECREMENT) | (1 << WIDX_AINCREMENT) | (1 << WIDX_ADECREMENT);
+    window_init_scroll_widgets(window);
+    window_push_others_below(window);
+
+    window->colours[2] = COLOUR_DARK_GREEN;
+
+    return window;
+}
+
+static void window_scenery_scatter_close([[maybe_unused]] rct_window* w)
```
AaronVanGeffen: You can remove this function and its declaration above.

danielmeinert:
It's nice to see some improvements to the scarttering tool! 

I have two suggestions for features that might make it even more powerful:
- have an option that the individual objects are placed with a random orientation. This would be handy when making forests as you get automatically more diverse foliage.

- open the scattering tool for large scenery objects. I don't know if this is possible but lots of CSO trees are LS and it would be handy wo be able to scatter them, too.

ocalhoun6:
>     * have an option that the individual objects are placed with a random orientation. This would be handy when making forests as you get automatically more diverse foliage.

This would be especially handy when scattering cacti! Those are very rotation dependent and it looks weird if all the ones you scattered are facing the same way.

(For that matter, I think having random rotation should probably be the default setting of the scatter tool, though it could still be nice to be able to deselect it in certain circumstances.)

AaronVanGeffen:
I have rebased the PR on top of develop.

Thanks for the suggestions for future improvements. For now, I think our goal should be to get the window up to the game's standards. Let's not feature creep this PR; we can add more features to the window later. Feel free to log them to our issue tracker, if they aren't already. :)

Gymnasiast:
I fear the feature is becoming a bit too complicated.

Is the "amount" spinner really necessary? If you want more items, you can always click again.

AaronVanGeffen:
Hmm, being able to set density through both area and amount seems like a decent feature to me. I do see an issue in their presentation, however.

The area widget works intuitively, as it uses the same paradigm as the terraforming tools. I think we can therefore drop its label. For the amount, it is likely better to switch to a regular spinner widget, while retaining the label.

AaronVanGeffen:
It would be fairly straightforward to implement random rotations by patching:

https://github.com/OpenRCT2/OpenRCT2/blob/1dc7bc2466fcd60d345921fb978a61a1fffbadfb/src/openrct2-ui/windows/TopToolbar.cpp#L1809-L1812

https://github.com/OpenRCT2/OpenRCT2/blob/1dc7bc2466fcd60d345921fb978a61a1fffbadfb/src/openrct2-ui/windows/TopToolbar.cpp#L1832-L1834

to not use `gSceneryPlaceRotation`, but a local variable instead. This would be set to random rotation for each cluster element, and set to `gSceneryPlaceRotation` otherwise. Of course, `SMALL_SCENERY_FLAG_ROTATABLE` would have to be taken into account as well.

duncanspumpkin:
I would leave it simple for now. The whole scatter mechanic needs to be refactored into a game action soon and it would be best we don't make that harder than it needs to be.

scrapes:
Sorry for being offline so long had a pretty rough exam phase. 
Is there now still work to be done or should i wait till the game action refractor to implement any changes? The only thing i could imagine now that i try to rework the design a bit to make it more RCTy.  

Thanks for rebasing and working on this PR so far!

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -486,7 +489,11 @@ rct_window* window_scenery_open()
     gSceneryPlaceRotation = 0;
     gWindowSceneryPaintEnabled = 0; // repaint coloured scenery tool state
     gWindowSceneryEyedropperEnabled = false;
-    gWindowSceneryClusterEnabled = 0; // build cluster tool state
+    gWindowSceneryScatterEnabled = 0; // build cluster tool state
+    gWindowSceneryScatterWindow = nullptr;
+
+    gWindowSceneryScatterSize = 2;
+    gWindowSceneryScatterAmount = 10;
```
Gymnasiast: Needs to be raised to 35 again, which is originally was.

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -486,7 +489,11 @@ rct_window* window_scenery_open()
     gSceneryPlaceRotation = 0;
     gWindowSceneryPaintEnabled = 0; // repaint coloured scenery tool state
     gWindowSceneryEyedropperEnabled = false;
-    gWindowSceneryClusterEnabled = 0; // build cluster tool state
+    gWindowSceneryScatterEnabled = 0; // build cluster tool state
+    gWindowSceneryScatterWindow = nullptr;
+
+    gWindowSceneryScatterSize = 2;
```
Gymnasiast: Needs to be raised to the old size again.

src/openrct2-ui/windows/Scenery.cpp
scrapes: Why does it specifically need to be raised to 35 again?

Gymnasiast: Because that was the old value. The scenery scatter window should default to the old parameters (number of trees and tool size).

AaronVanGeffen:
I've changed the lay-out a little so it resembles other windows a bit more, but I'm not particularly   satisfied with the results yet. Perhaps we should drop one or both of the labels?

Feedback welcome.

Forest Frontiers 2020-02-17 23-09-28

Gymnasiast:
Do we really need an `amount` spinner? IMO we should keep it simple and just stick with the area selection (deriving the amount of trees from the area size), allowing us to drop all labels.

AaronVanGeffen:
I've been mulling as well. Initially, I was against that, but it does not seem to be adding much over increasing or decreasing the number of elements proportionally to the area. I no longer object to this.

However, what would such a heuristic look like? Just double the area value to get the number of elements? (16 * 2 = 32, 2 * 2 = 4 ... it would fit, I guess.)

Umdlye:
If a toggle for randomized rotation were added later, would that need a label?

I'd agree that an amount spinner wouldn't be too useful since you can already just click multiple times to get the desired amount, and people would probably want such a spinner to scale with the area size or you'd have to keep adjusting both.

AaronVanGeffen:
> If a toggle for randomized rotation were added later, would that need a label?

That's what tooltips are used for in this game, I think. As an example, look at how the mountain tool is handled.

> and people would probably want such a spinner to scale with the area size or you'd have to keep adjusting both.

This is a good point. I agree this needs work, yes.

Gymnasiast:
@AaronVanGeffen Doubling the area value would work, yes.

Regarding the rotation: trees currently "autorotate" when placing one of them. I think we should apply that to the cluster tool as well - no need to add another button. But we're leaving this for another PR anyway.

Umdlye:
@AaronVanGeffen If people still desperately wanted such a spinner it might be better to call it "density"  with a 1-100% range.

Gymnasiast:
Nah, it's not necessary, I'd say. Even if there are too few trees, you can just click again.

ocalhoun6:
> 
> 
> Nah, it's not necessary, I'd say. Even if there are too few trees, you can just click again.

But if there's too many?

Yes, you can always click more times to get more (though if you have to do that a lot, it could get tedious). But if you want _less_ than the default amount, what then? You have to place the default amount and then go in and delete some. _Very_ tedious.

I say, since we already have the ability to do this, why give the player less options?

(Although, I do agree that having it be a 'density' spinner instead, adjustable from 1% to 100% would be better. Then you don't have to struggle as much to tune the amount compared to how large of an area you're using.)

AaronVanGeffen:
> But if you want _less_ than the default amount, what then? You have to place the default amount and then go in and delete some. _Very_ tedious.

Hmm, yes. Using a smaller area would place less scenery, but with the same density. Not ideal, either.

Designing a good UI is hard.

Broxzier:
Instead of amount, I would use a slider to determine the likeliness something gets placed for each tile. near 0 almost nothing would get placed, near 100 almost all tiles will receive an object.

With such a likeliness slider, it's possible that no items are placed at all, so rerunning the function until at least one item gets placed might be desirable.

AaronVanGeffen:
I have an idea. What about introducing three density modes instead, with buttons representing them? I'm thinking 25%, 50%, and 75% density. These could be displayed similarly to the rolling/banking buttons. I'll try to make time for making a mock-up sometime today.

Umdlye:
scatter
(Ignore the window title)

AaronVanGeffen:
Thank you. That's more or less what I had in mind, yes. Probably with a border around it, like the roll/banking button set in the ride construction window.

Umdlye:
Oh, like that. That's nice, would keep them clearly separated from any future additional toggle buttons. I thought you meant a border around the buttons themselves for a second 😛 

Here are the buttons I made in case anyone wants to use them.
medium
low

Gymnasiast:
I'm not sure it's needed, but it beats the spinner, I guess.

Featuritis and buttonitis is a big problem in open source projects and I wish to keep stuff simple.

ocalhoun6:
> Featuritis and buttonitis is a big problem in open source projects

I don't know... Maybe it's just my personal opinion, but I _like_ having lots of features, buttons, and options. And I have a big problem with a lot of large-scale commercial software when they tend to think that taking options away from the user is 'user-friendly' ... even doing updates to take away features it previously had.

I'd say, as long as you can follow the rule of 'if you don't know what it does, don't mess with it' and everything turns out fine, the more options and buttons the better. (Well, at least until you get to the point where it's just taking up too much space on the screen or something.) So as long as leaving all these controls in the default setting will result in the default scenery scatter behavior, I really don't see the problem with it.

(And at the _very_ worst, hide extra features behind an 'advanced options' tab or something, so basic users aren't confused by it, but it's still available to those who know what it does.)

pizza2004:
The problem is simply that the more features the more code, and the more code the more there is to maintain.  If for instance there was a need to move to a new windowing system you'd have more to reconstruct, and it introduces more places for bugs to crop up.

AaronVanGeffen:
This what I ended up with, in the end. The density buttons don't work yet, but it looks much better than what we had, I think. :)

Forest Frontiers 2020-02-18 23-07-10

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -486,7 +489,11 @@ rct_window* window_scenery_open()
     gSceneryPlaceRotation = 0;
     gWindowSceneryPaintEnabled = 0; // repaint coloured scenery tool state
     gWindowSceneryEyedropperEnabled = false;
-    gWindowSceneryClusterEnabled = 0; // build cluster tool state
+    gWindowSceneryScatterEnabled = 0; // build cluster tool state
```
Gymnasiast: Use true/false.

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -486,7 +489,11 @@ rct_window* window_scenery_open()
...
+    gWindowSceneryScatterWindow = nullptr;
+
+    gWindowSceneryScatterSize = 16;
+    gWindowSceneryScatterAmount = 35;
```
Gymnasiast: Considering we've gone for a density selection, is this variable still used? (And if so, does this default value still make sense?)

AaronVanGeffen: You were reviewing before I had implemented density. This variable has been replaced now.

AaronVanGeffen:
I believe this is now ready for review.

src/openrct2-ui/windows/SceneryScatter.cpp
```diff
@@ -0,0 +1,238 @@
...
+            TextInputDescriptionArgs[1] = MAXIMUM_TOOL_SIZE;
+            maxlen = 3;
+            break;
+    }
+    window_text_input_open(w, widgetindex, STR_SELECTION_SIZE, STR_ENTER_SELECTION_SIZE, STR_NONE, STR_NONE, maxlen);
+}
+
+static void window_scenery_scatter_update(rct_window* w)
+{
+    w->frame_no++;
```
AaronVanGeffen: I think we can safely remove this event as well, as the window does not use any animations. (I already removed the empty `close` event.)

src/openrct2-ui/windows/SceneryScatter.cpp
```diff
@@ -0,0 +1,238 @@
...
+}
+
+static void window_scenery_scatter_invalidate(rct_window* w)
+{
+    // Set the preview image button to be pressed down
+    w->pressed_widgets = (1 << WIDX_PREVIEW);
+
+    // TODO: press density buttons
```
AaronVanGeffen: Stray comment

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -1783,8 +1798,16 @@ static void window_top_toolbar_scenery_tool_down(int16_t x, int16_t y, rct_windo
...
-                    cur_grid_x += ((util_rand() % 16) - 8) * 32;
-                    cur_grid_y += ((util_rand() % 16) - 8) * 32;
+                    if (gWindowSceneryScatterSize % 2 != 0)
+                    {
+                        cur_grid_x += ((util_rand() % (gWindowSceneryScatterSize)) - (gWindowSceneryScatterSize / 2)) << 5;
+                        cur_grid_y += ((util_rand() % (gWindowSceneryScatterSize)) - (gWindowSceneryScatterSize / 2)) << 5;
+                    }
+                    else
+                    {
+                        cur_grid_x += ((util_rand() % (gWindowSceneryScatterSize)) - (gWindowSceneryScatterSize / 2) + 1) << 5;
```
AaronVanGeffen: Deduplicate these two branches.

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -2680,12 +2708,23 @@ static void top_toolbar_tool_update_scenery(int16_t x, int16_t y)
...
+                if (gWindowSceneryScatterSize % 2 != 0)
+                {
+                    gMapSelectPositionA.x = mapTile.x - cluster_size / 2;
+                    gMapSelectPositionA.y = mapTile.y - cluster_size / 2;
+                    gMapSelectPositionB.x = mapTile.x + cluster_size / 2;
+                    gMapSelectPositionB.y = mapTile.y + cluster_size / 2;
+                }
+                else
+                {
+                    gMapSelectPositionA.x = mapTile.x - cluster_size / 2;
```
AaronVanGeffen: Deduplicate these two branches.

Umdlye:
Closing the scatter window with the X button disables the scatter tool but pressing backspace just closes the window while keeping the tool enabled.

src/openrct2-ui/windows/Window.h
```diff
@@ -19,10 +19,14 @@ using scenarioselect_callback = void (*)(const utf8* path);
 struct Peep;
 struct TileElement;
 struct Vehicle;
+enum class ScatterToolDensity : uint8_t;
 
 extern uint16_t gWindowSceneryTabSelections[];
 extern uint8_t gWindowSceneryActiveTabIndex;
-extern uint8_t gWindowSceneryClusterEnabled;
+extern rct_window* gWindowSceneryScatterWindow;
```
AaronVanGeffen: Can be removed now.

data/language/en-GB.txt
```diff
@@ -3668,6 +3668,12 @@ STR_6346    :Tile Inspector: Decrease element height
 STR_6347    :Path additions cannot be placed on level crossings!
 STR_6348    :Remove level crossing first!
 STR_6349    :Random title sequence
+STR_6350    :Scatter
+STR_6351    :Scenery Scatter Tool
+STR_6352    :Density
+STR_6353    :{SMALLFONT}{BLACK}Low density (25%)
+STR_6354    :{SMALLFONT}{BLACK}Medium density (50%)
+STR_6355    :{SMALLFONT}{BLACK}High density (75%)
```
AaronVanGeffen: These lines should probably more closely reflect what is actually happening, e.g. `1x`, `2x`, `3x`. (Right now, `25%` could be interpreted as 16 * 16 * 0.25 = 64, while in reality only 16 elements are placed.)

Gymnasiast: Just get rid of the percentages and title them `Low density`, etc. The 1x/2x suffices don't add anything informative.

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -2680,12 +2707,18 @@ static void top_toolbar_tool_update_scenery(int16_t x, int16_t y)
...
+            if (gWindowSceneryScatterEnabled)
+            {
+                uint16_t cluster_size = (gWindowSceneryScatterSize - 1) << 5;
+                gMapSelectPositionA.x = mapTile.x - cluster_size / 2;
+                gMapSelectPositionA.y = mapTile.y - cluster_size / 2;
+                gMapSelectPositionB.x = mapTile.x + cluster_size / 2;
+                gMapSelectPositionB.y = mapTile.y + cluster_size / 2;
+                if (gWindowSceneryScatterSize % 2 == 0)
+                {
+                    gMapSelectPositionB.x += (1 << 5);
```
Gymnasiast: What's the significance of the `<< 5`?

AaronVanGeffen: These shifts are already present in the code as seen in develop. I don't mind changing to `COORDS_XY_STEP`, but changing the data types feels a bit out of scope for this PR.

AaronVanGeffen:
Thanks. I'll squash a few of the smaller commits and merge once CI passes, then.

PR #10361, opened 2019-12-12
Improve cmake config for Windows and Visual Studio

IntelOrca:
- Ensure static CRT is used when compiling.
- Allow Ninja generator to work (remove use of `CMAKE_VS_PLATFORM_NAME`).
- Move CMakeSettings.json to new standard location for VS.

Issues still remaining (not to be done for this PR):
- No support for getting discord-rpc via `find_library` (vcpkg).
- Build errors when using `-DWITH_TESTS`
- install target is far too geared towards unix and does not lay out our current portable distributed zip.


CMakeLists.txt
```diff
@@ -179,13 +179,18 @@ function (ADD_CHECK_CXX_FLAG _CXXFLAGS _CACHE_VAR _FLAG)
...
+    set(CMAKE_CXX_FLAGS_RELEASE "/MT /O2 /Ob2 /DNDEBUG")
+
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /utf-8 /permissive- /Zc:externConstexpr /EHsc /WX")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4244") # C4244: 'conversion_type': conversion from 'type1' to 'type2', possible loss of data
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4068") # C4068: unknown pragma
 
     add_definitions(-D_CRT_SECURE_NO_WARNINGS)
     add_definitions(-D_SCL_SECURE_NO_WARNINGS)
-    if ((CMAKE_VS_PLATFORM_NAME STREQUAL "Win32") OR (CMAKE_VS_PLATFORM_NAME STREQUAL "x64"))
+    if (CMAKE_SIZEOF_VOID_P EQUAL 8)
```
janisozaur: Arm64 doesn't have sse nor avx

CMakeLists.txt
```diff
@@ -291,12 +296,12 @@ endif ()
 if (DOWNLOAD_OBJECTS)
     # If rct2.wtrcyan.json or data/object/ exists, assume all the objects are already present
     install(CODE
-        "if (EXISTS \"\$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/object/rct2/water/rct2.wtrcyan.json\" OR EXISTS ${CMAKE_SOURCE_DIR}/data/object/)\n\
+        "if (EXISTS \"\$ENV{DESTDIR}${CMAKE_INSTALL_FULL_DATADIR}/${PROJECT}/object/rct2/water/rct2.wtrcyan.json\" OR EXISTS ${CMAKE_SOURCE_DIR}/data/object/)\n\
```
janisozaur: Why is it necessary? You mentioned that install target doesn't produce the windows zip, so why change it at all?

CMakeSettings.json
```diff
@@ -0,0 +1,22 @@
...
+  "configurations": [
+    {
+      "name": "x64-Debug",
+      "generator": "Ninja",
+      "configurationType": "Debug",
+      "inheritEnvironments": [ "msvc_x64_x64" ],
+      "buildRoot": "${projectDir}\\out\\build\\${name}",
+      "installRoot": "${projectDir}\\out\\install\\${name}",
+      "cmakeCommandArgs": "",
+      "buildCommandArgs": "-v",
```
janisozaur: Is this file really meant to be part of the repo? It seems overly specific and I would prefer not to have it.

CMakeLists.txt
IntelOrca: Because it is wrong, it means you can't use blank as DESTDIR, it will just write to `/...`

CMakeSettings.json
IntelOrca: This is so VS passes the right config to cmake, in particular the vcpkg triplet.

CMakeLists.txt
IntelOrca: So what can I put here instead to fix this?

IntelOrca: ?

janisozaur: `CMAKE_SYSTEM_PROCESSOR`?

IntelOrca: The way cmake expects to work for Windows and MSVC is that you launch the correct VS environment for your target platform then run cmake.

`CMAKE_SYSTEM_PROCESSOR` always returns AMD64 for both an x86 and AMD64 visual studio development environment.

Do you have any other alternatives?

IntelOrca: If not then I will merge this as we don't yet support Windows on ARM, it can be changed when we do.

janisozaur: But we do and this (current cmake) is how it is achieved.

IntelOrca: How about this?

`if (CMAKE_SIZEOF_VOID_P EQUAL 8 AND CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64")`

janisozaur: You mentioned earlier that the variable was not set to expected value when using x86_64 and x86, does it work for amd64?

IntelOrca: Do you mean ARM64? It always returns AMD64 for Win32 and x64 which is why the pointer size check is also needed to differentiate between the platforms. I would expect it to return ARM64 if you were targeting ARM, but I have no way of checking this.

janisozaur: Yeah, misread the value there. Checking should be trivial, create a dummy cmake project and have it print the value and pass on the relevant generator, both msbuild and ninja.

IntelOrca: I think I need the ARM cross compiling tools for Visual Studio for me to try this and I don't want to install that as it is quite large.

PR #10363, opened 2019-12-12
Refactor to reduce LocationXY usage

duncanspumpkin:
A couple odds and ends.

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -534,7 +534,7 @@ static void viewport_interaction_remove_footpath(TileElement* tileElement, Coord
     {
         if (tileElement2->GetType() == TILE_ELEMENT_TYPE_PATH && tileElement2->base_height == z)
         {
-            footpath_remove({ mapCoords, z }, GAME_COMMAND_FLAG_APPLY);
+            footpath_remove({ mapCoords, z * 8 }, GAME_COMMAND_FLAG_APPLY);
             break;
         }
     } while (!(tileElement2++)->IsLastForTile());
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -534,7 +534,7 @@ static void viewport_interaction_remove_footpath(TileElement* tileElement, Coord
     {
         if (tileElement2->GetType() == TILE_ELEMENT_TYPE_PATH && tileElement2->base_height == z)
         {
-            footpath_remove(mapCoords.x, mapCoords.y, z, GAME_COMMAND_FLAG_APPLY);
+            footpath_remove({ mapCoords, z }, GAME_COMMAND_FLAG_APPLY);
             break;
         }
     } while (!(tileElement2++)->IsLastForTile());
--- a/src/openrct2-ui/windows/Footpath.cpp
+++ b/src/openrct2-ui/windows/Footpath.cpp
@@ -183,7 +183,7 @@ static void window_footpath_start_bridge_at_point(ScreenCoordsXY screenCoords);
 static void window_footpath_construct();
 static void window_footpath_remove();
 static void window_footpath_set_enabled_and_pressed_widgets();
-static void footpath_get_next_path_info(int32_t* type, int32_t* x, int32_t* y, int32_t* z, int32_t* slope);
+static void footpath_get_next_path_info(int32_t* type, CoordsXYZ& footpathLoc, int32_t* slope);
 static bool footpath_select_default();
 
 /**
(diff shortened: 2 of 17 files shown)


src/openrct2/world/MoneyEffect.cpp
```diff
@@ -79,7 +79,7 @@ void rct_money_effect::Create(money32 value)
 {
     CoordsXYZ mapPosition = { gCommandPosition.x, gCommandPosition.y, gCommandPosition.z };
 
-    if (mapPosition.x == LOCATION_NULL)
+    if (mapPosition.x == COORDS_NULL)
```
tupaschoal: Hadn't even realised that we had `COORDS_NULL`. Some of the Coords object implement `isNull()`, should we extend to the others so that the code doesn't have to reference `COORDS_NULL`?

duncanspumpkin: Yeah i think i might need to revert this change anyway as need to check how the value is set.

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -534,7 +534,7 @@ static void viewport_interaction_remove_footpath(TileElement* tileElement, Coord
     {
         if (tileElement2->GetType() == TILE_ELEMENT_TYPE_PATH && tileElement2->base_height == z)
         {
-            footpath_remove(mapCoords.x, mapCoords.y, z, GAME_COMMAND_FLAG_APPLY);
+            footpath_remove({ mapCoords, z }, GAME_COMMAND_FLAG_APPLY);
```
Gymnasiast: You're putting a TileCoords z coordinate into a Coords without multiplying by 8.

src/openrct2/world/Footpath.cpp
```diff
@@ -184,15 +182,15 @@ money32 footpath_provisional_set(int32_t type, int32_t x, int32_t y, int32_t z,
...
         {
             // Going either straight on, or down.
-            virtual_floor_set_height(gFootpathProvisionalPosition.z * 8);
+            virtual_floor_set_height(gFootpathProvisionalPosition.z);
         }
         else
         {
             // Going up in the world!
-            virtual_floor_set_height((gFootpathProvisionalPosition.z + 2) * 8);
+            virtual_floor_set_height(gFootpathProvisionalPosition.z + 2 * 8);
```
Gymnasiast: Please revert. This makes the order ambiguous. In standard math rules, this would be equivalent to `gFootpathProvisionalPosition.z + 16`.

duncanspumpkin: you mean `gFootpathProvisionalPosition.z + 16` i assume. It used to be a small z now its a big z so only the 2 needs the * 8.

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -995,20 +996,14 @@ static void window_footpath_construct()
...
             }
 
+            gFootpathConstructFromPosition = footpathLoc;
             // If we have just built an upwards slope, the next path to construct is
             // a bit higher. Note that the z returned by footpath_get_next_path_info
             // already is lowered if we are building a downwards slope.
             if (gFootpathConstructSlope == 2)
             {
-                gFootpathConstructFromPosition.z = (z + 2) * 8;
+                gFootpathConstructFromPosition.z += 2 * 8;
```
Gymnasiast:
```suggestion
                gFootpathConstructFromPosition.z += 2;
```

src/openrct2/world/Entrance.cpp
```diff
@@ -56,8 +55,7 @@ void park_entrance_remove_ghost()
     if (gParkEntranceGhostExists)
     {
         gParkEntranceGhostExists = false;
-        auto parkEntranceRemoveAction = ParkEntranceRemoveAction(
-            { gParkEntranceGhostPosition.x, gParkEntranceGhostPosition.y, gParkEntranceGhostPosition.z * 16 });
```
Gymnasiast: I agree to refactoring out this oddball coordinate, but it will need to be tested thoroughly.

src/openrct2/world/Footpath.cpp
Gymnasiast: Ok, but then I do prefer if you put brackets around the `2 * 8` bit (for legibility and disambiguation).

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -1053,18 +1046,15 @@ static void footpath_remove_tile_element(TileElement* tileElement)
...
-    x = gFootpathConstructFromPosition.x - CoordsDirectionDelta[edge].x;
-    y = gFootpathConstructFromPosition.y - CoordsDirectionDelta[edge].y;
-    gFootpathConstructFromPosition.x = x;
-    gFootpathConstructFromPosition.y = y;
-    gFootpathConstructFromPosition.z = z << 3;
-    gFootpathConstructDirection = edge;
+    gFootpathConstructFromPosition.x -= CoordsDirectionDelta[edge].x;
+    gFootpathConstructFromPosition.y -= CoordsDirectionDelta[edge].y;
+    gFootpathConstructFromPosition.z = z * 8;
+    gFootpathConstructDirection = direction_reverse(edge);
```
tupaschoal: Because edge stoppped being updated before the `{ x ,y }` assignment, this caused #10426

tupaschoal:
Helps #10267

PR #10364, opened 2019-12-12
Use ScreenCoordsXY on context cursor operations

tupaschoal:
After this it looks like turning `gInputDragLastX` and `gInputDragLastY` on `MouseInput.cpp` into a single `ScreenCoordsXY` variable will be fairly easy and it will clean up a lot of calls.


tupaschoal:
Is this good to merge? I wanted to follow-up on the global variables I mentioned above.

PR #10369, opened 2019-12-13
Use CoordsXY on Map::map_surface_is_blocked()

tupaschoal:


duncanspumpkin:
You have a little clang format issue

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -149,7 +149,7 @@ static int32_t guest_surface_path_finding(Peep* peep)
 
         if (!fence_in_the_way(x, y, z, z + 4, backwardsDirection))
         {
-            if (!map_surface_is_blocked(x, y))
+            if (!map_surface_is_blocked({x, y}))
```
tupaschoal:
```suggestion
            if (!map_surface_is_blocked({ x, y }))
```

src/openrct2/world/Map.cpp
```diff
@@ -2054,12 +2054,12 @@ int32_t map_get_tile_quadrant(int32_t mapX, int32_t mapY)
  *
  *  rct2: 0x00693BFF
  */
-bool map_surface_is_blocked(int16_t x, int16_t y)
+bool map_surface_is_blocked(CoordsXY mapCoords)
 {
-    if (x >= 8192 || y >= 8192)
+    if (mapCoords.x >= 8192 || mapCoords.y >= 8192)
```
Gymnasiast: While you're at it: this is a magic value of 256 × 32. We have a constant called MAX_MAP_SIZE_TECHNICAL or something like that with a value of 256. Could you modify the code to use it?

duncanspumpkin: There is a function that can be called instead. That checks for outside of range.

tupaschoal: Could you point me to the function, I couldn't find it :\

In the meantime, I changed it to use `MAXIMUM_MAP_SIZE_TECHNICAL` instead.

Gymnasiast: I think it's `map_is_location_valid()`.

tupaschoal: Thanks!
Small note, it also checks if the coords are `>= 0` which wasn't before, but I guess this is a plus.

tupaschoal: Thanks!
Small note, it also checks if the coords are >= 0 which wasn't before, but I guess this is a plus.

Gymnasiast: Previously, it didn't need to, because the input parameters couldn't be negative:
```
bool map_surface_is_blocked(int16_t x, int16_t y)
```
so previously, just checking for >=8192 was sufficient to cover both.

tupaschoal:
Helps #10267

PR #10371, opened 2019-12-13
Make map_get_surface_element_at only accept Coords objects

tupaschoal:
This abstracts away the `/ 32` from all of the calls. Every place that was calling the `x, y` alternative was moved to call the `TileCoordsXY` one and the others that were already calling with `CoordsXY` were just made explicit to disambiguate.

### DIFF
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -765,7 +765,7 @@ bool Peep::Place(TileCoordsXYZ location, bool apply)
     TileElement* tileElement = reinterpret_cast<TileElement*>(pathElement);
     if (!pathElement)
     {
-        tileElement = reinterpret_cast<TileElement*>(map_get_surface_element_at(location.x, location.y));
+        tileElement = reinterpret_cast<TileElement*>(map_get_surface_element_at(TileCoordsXY{ location.x, location.y }));
     }
 
     if (!tileElement)
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -765,8 +765,7 @@ bool Peep::Place(TileCoordsXYZ location, bool apply)
     TileElement* tileElement = reinterpret_cast<TileElement*>(pathElement);
     if (!pathElement)
     {
-        tileElement = reinterpret_cast<TileElement*>(
-            map_get_surface_element_at(TileCoordsXY{ location.x, location.y }.ToCoordsXY()));
+        tileElement = reinterpret_cast<TileElement*>(map_get_surface_element_at(location.ToCoordsXYZ()));
     }
 
     if (!tileElement)
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -765,7 +765,8 @@ bool Peep::Place(TileCoordsXYZ location, bool apply)
     TileElement* tileElement = reinterpret_cast<TileElement*>(pathElement);
     if (!pathElement)
     {
-        tileElement = reinterpret_cast<TileElement*>(map_get_surface_element_at(TileCoordsXY{ location.x, location.y }));
+        tileElement = reinterpret_cast<TileElement*>(
+            map_get_surface_element_at(TileCoordsXY{ location.x, location.y }.ToCoordsXY()));
     }
 
     if (!tileElement)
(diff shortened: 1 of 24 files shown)


src/openrct2/world/MapGen.cpp
```diff
@@ -349,7 +349,8 @@ static void mapgen_place_trees()
         pos = availablePositions[i];
 
         int32_t type = -1;
-        auto* surfaceElement = map_get_surface_element_at(pos.x, pos.y);
+        // TODO: This is Converting CoordsXY to TileCoordsXY directly without /32
+        auto* surfaceElement = map_get_surface_element_at(TileCoordsXY{ pos.x, pos.y });
```
tupaschoal: Please comment on this. `pos` is a `CoordsXY` object, so I'm thinking this is a bug.

src/openrct2/world/Map.cpp
```diff
@@ -203,9 +203,9 @@ void map_set_tile_elements(int32_t x, int32_t y, TileElement* elements)
     gTileElementTilePointers[x + y * MAXIMUM_MAP_SIZE_TECHNICAL] = elements;
 }
 
-SurfaceElement* map_get_surface_element_at(int32_t x, int32_t y)
+SurfaceElement* map_get_surface_element_at(const TileCoordsXY& tileCoords)
```
duncanspumpkin: Remove the overload for `TileCoordsXY` it should only be used with `CoordsXY`.

src/openrct2/world/MapGen.cpp
duncanspumpkin: L307 and L309 are incorrect this is a `TileCoordsXY` object but really we should be modifying this function to make this `CoordsXY`

src/openrct2/world/Map.cpp
tupaschoal: What about all of the places that use it, we'll convert to `CoordsXY`, doing the multiplications, or are they just wrong?
(Including the one I commented when opening the PR)

duncanspumpkin: As it stands its a bit confusing having the two versions when types can be easily converted between the two. `CoordsXY` should be preferred in almost all situations. I'm not really sure what is best because this function is probably one of the ones that is called a ridiculous amount of times per tick so it should probably made to be as simple as possible.

tupaschoal:
Done

src/openrct2/world/MapGen.cpp
```diff
@@ -349,7 +349,8 @@ static void mapgen_place_trees()
         pos = availablePositions[i];
 
         int32_t type = -1;
-        auto* surfaceElement = map_get_surface_element_at(pos.x, pos.y);
+        // TODO: This is Converting CoordsXY to TileCoordsXY directly without /32
```
Gymnasiast: Does this apply?

tupaschoal: It doesn't anymore

Gymnasiast: Could you remove it, then?

tupaschoal:
Fixed pending issue and rebased

src/openrct2/peep/Peep.cpp
```diff
@@ -765,7 +765,8 @@ bool Peep::Place(TileCoordsXYZ location, bool apply)
     TileElement* tileElement = reinterpret_cast<TileElement*>(pathElement);
     if (!pathElement)
     {
-        tileElement = reinterpret_cast<TileElement*>(map_get_surface_element_at(location.x, location.y));
+        tileElement = reinterpret_cast<TileElement*>(
```
Gymnasiast: `tileElement = reinterpret_cast<TileElement*>(map_get_surface_element_at(location.ToCoordsXYZ));` should be sufficient.

test/tests/Pathfinding.cpp
```diff
@@ -151,7 +152,8 @@ class PathfindingTestBase : public testing::Test
     {
         const uint32_t forbiddenSurfaceStyle = 8u;
 
-        const uint32_t style = map_get_surface_element_at(location.x, location.y)->GetSurfaceStyle();
+        const uint32_t style = map_get_surface_element_at(TileCoordsXY{ location.x, location.y }.ToCoordsXY())
```
Gymnasiast: location.ToCoordsXYZ()

tupaschoal:
Done!

PR #10373, opened 2019-12-13
Prefer map_is_location_valid over hardcoded map limits

tupaschoal:


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2423,7 +2423,7 @@ static void sub_6CBCE2(
     gMapSizeUnits = 255 * 32;
     gMapSizeMinus2 = (255 * 32) + 286;
     gMapSize = 256;
-    gMapSizeMaxXY = (256 * 32) - 1;
+    gMapSizeMaxXY = (MAXIMUM_MAP_SIZE_TECHNICAL * 32) - 1;
```
Gymnasiast: You can also apply `MAXIMUM_MAP_SIZE_TECHNICAL` (or `MAXIMUM_MAP_SIZE_TECHNICAL - 1`) to the lines above.

tupaschoal: Done

PR #10374, opened 2019-12-14
Use CoordsXYZD on map_get_wall_element_at()

tupaschoal:


src/openrct2/world/Map.cpp
```diff
@@ -2324,18 +2324,19 @@ TileElement* map_get_track_element_at_with_direction_from_ride(
...
+    auto tileWallCoords = TileCoordsXYZ(wallCoords);
+    TileElement* tileElement = map_get_first_element_at(tileWallCoords.x, tileWallCoords.y);
     if (tileElement == nullptr)
         return nullptr;
     do
     {
         if (tileElement->GetType() != TILE_ELEMENT_TYPE_WALL)
             continue;
-        if (tileElement->base_height != z)
+        if (tileElement->base_height != tileWallCoords.z)
```
duncanspumpkin: This should be `/8`. A CoordsXYZ.z value is *8 of a base_height.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7432,12 +7432,11 @@ static void vehicle_update_scenery_door(rct_vehicle* vehicle)
         trackBlock++;
     }
     const rct_track_coordinates* trackCoordinates = &TrackCoordinates[trackType];
-    int32_t x = floor2(vehicle->x, 32);
-    int32_t y = floor2(vehicle->y, 32);
-    int32_t z = (vehicle->track_z - trackBlock->z + trackCoordinates->z_end) >> 3;
+    auto wallCoords = CoordsXYZ{ CoordsXY{ vehicle->x, vehicle->y }.ToTileStart(),
+                                 vehicle->track_z - trackBlock->z + trackCoordinates->z_end };
```
Gymnasiast: I prefer `CoordsXYZ { vehicle->x, vehicle->y, vehicle->track_z - trackBlock->z + trackCoordinates->z_end}.ToTileStart()`. If CoordsXYZ does not yet have that method, it should be added.

tupaschoal: Done!

PR #10375, opened 2019-12-14
Make map entrance/exit fetching functions use Coords objects

tupaschoal:


src/openrct2/actions/PlaceParkEntranceAction.hpp
```diff
@@ -79,13 +79,12 @@ DEFINE_GAME_ACTION(PlaceParkEntranceAction, GAME_COMMAND_PLACE_PARK_ENTRANCE, Ga
...
         for (uint8_t index = 0; index < 3; index++)
         {
             if (index == 1)
             {
-                entranceLoc.x += CoordsDirectionDelta[(_loc.direction - 1) & 0x3].x;
-                entranceLoc.y += CoordsDirectionDelta[(_loc.direction - 1) & 0x3].y;
+                entranceLoc += CoordsDirectionDelta[(_loc.direction - 1) & 0x3];
             }
             else if (index == 2)
             {
```
duncanspumpkin: You can change L91 as well

src/openrct2/world/Map.cpp
```diff
@@ -1790,17 +1791,17 @@ EntranceElement* map_get_park_entrance_element_at(int32_t x, int32_t y, int32_t
     return nullptr;
 }
 
-EntranceElement* map_get_ride_entrance_element_at(int32_t x, int32_t y, int32_t z, bool ghost)
+EntranceElement* map_get_ride_entrance_element_at(const TileCoordsXYZ& entranceTileCoords, bool ghost)
```
duncanspumpkin: Make it `CoordsXYZ`

src/openrct2/world/Map.cpp
```diff
@@ -1815,17 +1816,17 @@ EntranceElement* map_get_ride_entrance_element_at(int32_t x, int32_t y, int32_t
     return nullptr;
 }
 
-EntranceElement* map_get_ride_exit_element_at(int32_t x, int32_t y, int32_t z, bool ghost)
+EntranceElement* map_get_ride_exit_element_at(const TileCoordsXYZ& exitTileCoords, bool ghost)
```
duncanspumpkin: Make it `CoordsXYZ`

src/openrct2/actions/PlaceParkEntranceAction.hpp
tupaschoal: I can't there is a multiplication first, only if I implement `operator*`

tupaschoal:
> Change it so that the types of these functions generally are `CoordsXYZ`. I don't see this as being anywhere near as special as get surface

Personally I think it looks uglier, people have to remember to do the conversions, but if the interface should always get a `CoordsXY`, it makes sense.

Gymnasiast:
@tupaschoal We're generally moving towards using `CoordsXY(ZDE) for everything. At some point, the ugliness should go away.

However, for now, I do still think there has to be a better way than calling `map_get_entrance_exit_element_at()` with bitshifts. Perhaps you could add another conversion method to one of the coordinate objects?

tupaschoal:
Ok, I've created a helper function for them to convert into each other.

PR #10383, opened 2019-12-15
Remove usage of LocationXY on TrackDesign

tupaschoal:
I took the chance to remove the `extern` of the three variables, as they were not being used anywhere outside `TrackDesign`

### DIFF
--- a/src/openrct2/ride/TrackDesign.cpp
+++ b/src/openrct2/ride/TrackDesign.cpp
@@ -73,9 +73,9 @@ struct map_backup
 
 TrackDesign* gActiveTrackDesign;
 bool gTrackDesignSceneryToggle;
-LocationXYZ16 gTrackPreviewMin;
-LocationXYZ16 gTrackPreviewMax;
-LocationXYZ16 gTrackPreviewOrigin;
+CoordsXYZ gTrackPreviewMin;
+CoordsXYZ gTrackPreviewMax;
+CoordsXYZ gTrackPreviewOrigin;
 
 bool byte_9D8150;
 static uint8_t _trackDesignPlaceOperation;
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/TrackDesign.cpp
```diff
@@ -351,13 +348,12 @@ rct_string_id TrackDesign::CreateTrackDesignMaze(const Ride& ride)
...
 
-    gTrackPreviewOrigin = { static_cast<int16_t>(startLoc.x), static_cast<int16_t>(startLoc.y),
-                            (int16_t)(startLoc.element->base_height * 8) };
+    gTrackPreviewOrigin = { startLoc.x, startLoc.y, startLoc.element->base_height * 8 };
 
     // x is defined here as we can start the search
     // on tile start_x, start_y but then the next row
     // must restart on 0
-    for (int16_t y = startLoc.y, x = startLoc.y; y < 8192; y += 32)
+    for (int32_t y = startLoc.y, x = startLoc.y; y < 8192; y += 32)
```
Gymnasiast: Magic value for maximum map size * 32.

We should have `constexpr`s for both this value and for 32, I think.

This may be picked up in another PR if desired.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2103,10 +2096,8 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
     td6->cost = cost;
     td6->track_flags = flags & 7;
 
-    CoordsXYZ centre;
-    centre.x = (gTrackPreviewMin.x + gTrackPreviewMax.x) / 2 + 16;
-    centre.y = (gTrackPreviewMin.y + gTrackPreviewMax.y) / 2 + 16;
-    centre.z = (gTrackPreviewMin.z + gTrackPreviewMax.z) / 2;
+    CoordsXYZ centre = { (gTrackPreviewMin.x + gTrackPreviewMax.x) / 2 + 16, (gTrackPreviewMin.y + gTrackPreviewMax.y) / 2 + 16,
```
Gymnasiast: Use `ToTileCentre`. (Add it if CoordsXYZ does not yet have it.)

src/openrct2/ride/TrackDesign.cpp
tupaschoal: I'll pick it up on a new PR

src/openrct2/ride/TrackDesign.cpp
tupaschoal: I mean, the variable is called centre, indeed, but `ToTileCentre` floors the value first and then centers it, it doesn't look like it will be the same.

Gymnasiast: It depends on whether `gTrackPreviewMin` and `gTrackPreviewMax` are pre-floored, I'd say.

Let's keep it like this for now, then.

tupaschoal:
Helps #10267

PR #10387, opened 2019-12-15
Fix #10376: No ratings generated when a shop and track intersect

### LINKED ISSUE #10376: No ratings generated when a shop and track intersect
OS: Windows 8.1
Version Open RCT2: e.g. 0.2.4 - 174
**ommit/Build: 18ba9b2 (automatically updateding launcher)

 In the challenge "Extreme Heights" I noticed that manny of my rides did not have a rating (Excitment, Intensity, nausia). They all remain on "Not yet available".

I removed every roller coast with this issue, but I can't make new roller coasters without this problem.

- [x] Reproducible in RCT2 (vanilla)? Yes
- [ ] Specific to multiplayer? no, this is single player

**Dump file**

**Screenshots / Video:**

Stats not yet available

**Save game:**

[Ratings not generated Extreme Heights - kopie.txt](https://github.com/OpenRCT2/OpenRCT2/files/3964057/Ratings.not.generated.Extreme.Heights.-.kopie.txt)

IntelOrca:
Just limiting rides with no buildable track for now so that you still get ratings for rides that change track type. This may cause behaviour changes to hacked flat rides but I am hoping they aren't nearly as common.

### DIFF
--- a/src/openrct2/ride/RideRatings.cpp
+++ b/src/openrct2/ride/RideRatings.cpp
@@ -220,6 +220,12 @@ static void ride_ratings_update_state_2()
             continue;
         if (tileElement->base_height != z)
             continue;
+        if (tileElement->AsTrack()->GetRideIndex() != ride->id)
+        {
+            // Only check that the track belongs to the same ride if ride does not have buildable track
+            if (!ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_HAS_TRACK))
+                continue;
+        }
 
         if (trackType == 255
             || (tileElement->AsTrack()->GetSequenceIndex() == 0 && trackType == tileElement->AsTrack()->GetTrackType()))
@@ -331,6 +337,12 @@ static void ride_ratings_update_state_5()
             continue;
         if (tileElement->base_height != z)
             continue;
+        if (tileElement->AsTrack()->GetRideIndex() != ride->id)
+        {
+            // Only check that the track belongs to the same ride if ride does not have buildable track
+            if (!ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_HAS_TRACK))
+                continue;
+        }
 
         if (trackType == 255 || trackType == tileElement->AsTrack()->GetTrackType())
         {
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -12,6 +12,7 @@
 - Fix: [#10228] Can't import RCT1 Deluxe from Steam.
 - Fix: [#10313] Path furniture can be placed on level crossings.
 - Fix: [#10325] Crash when banners have no text.
+- Fix: [#10376] No ratings generated when a shop and track intersect.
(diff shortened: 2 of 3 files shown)


Gymnasiast:
> This may cause behaviour changes to hacked flat rides but I am hoping they aren't nearly as common.

They are common, as an invisibility hack.

duncanspumpkin:
The track element numbers are shared between the two. So you can't filter like that.

Gymnasiast:
Then I think we should call `get_first_non_null_ride_type()` on `ride->subtype` and check if that ride type has track.

IntelOrca:
Hacked flat rides may not suffer too much but I would need an example save to check.

> Then I think we should call `get_first_non_null_ride_type()` on `ride->subtype` and check if that ride type has track.

This function is only ever used prior to a ride being created. Once it is created, `ride->type` is then the concrete behaviour that should be used for all ride logic. I thought multiple ride types is just allow a vehicle to run on several different coasters anyway.

Gymnasiast:
I have asked NE to supply a save. Again, setting the ride type to Crooked House is a common way to make a ride invisible. Combine that with merging and you have a situation where I expect this PR will break ratings generation.

I'll get back when I have that save (or I'll make one myself if they don't bother).

Gymnasiast:
Here is a save to test with. The ride in question is "Schans" and attempting to regenerate ratings does work on the current `develop`.

[WINTER IN 'T PARK MET DE PAKHUIS-INGANG.zip](https://github.com/OpenRCT2/OpenRCT2/files/4100270/WINTER.IN.T.PARK.MET.DE.PAKHUIS-INGANG.zip)

Gymnasiast:
Ok, that one didn't fail. It's perhaps because Crooked House ratings generation doesn't look at the track in the first place... that would make my point moot.

src/openrct2/ride/RideRatings.cpp
```diff
@@ -220,6 +220,12 @@ static void ride_ratings_update_state_2()
             continue;
         if (tileElement->base_height != z)
             continue;
+        if (tileElement->AsTrack()->GetRideIndex() != ride->id)
+        {
+            // Only check that the track belongs to the same ride if ride does not have buildable track
+            if (!ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_HAS_TRACK))
```
Gymnasiast: I can't quite wrap my head around why this solves the problem in #10376 - after all, this will only affect the Fried Rice Noodles Stall, which begs the question: why on earth does it even try to generate ratings for it?

IntelOrca: I guess because upkeep calculation is part of ride ratings and shops have upkeep.

IntelOrca: You actually want to find a save where someone has mixed ride types for a flat ride. That is all track for a "flat ride" has mixed ride IDs. This is probably quite rare though.

PR #10397, opened 2019-12-19
Remove gCommandPosition. Use GameActionRes instead.

duncanspumpkin:
Simplified some game action code to reduce reliance on gCommandPosition. Also modified network.cpp to save CoordsXYZ instead of LocationXYZ of the last game command.


src/openrct2/actions/RideEntranceExitPlaceAction.hpp
```diff
@@ -129,7 +127,7 @@ DEFINE_GAME_ACTION(RideEntranceExitPlaceAction, GAME_COMMAND_PLACE_RIDE_ENTRANCE
         auto res = MakeResult();
         res->Position.x = _loc.x + 16;
         res->Position.y = _loc.y + 16;
-        res->Position.z = tile_element_height(_loc);
+        res->Position.z = z;
```
tupaschoal: Is this the same as `ride->stations[_stationNum].Height * 8` ? They look like very different functions, I'm confused.

duncanspumpkin: Not sure why it was doing tile element height as it is the wrong height. It should have been station height from the beginning

src/openrct2/ReplayManager.cpp
```diff
@@ -768,11 +768,11 @@ namespace OpenRCT2
                 }
 
                 // Focus camera on event.
-                if (isPositionValid && gCommandPosition.x != LOCATION_NULL)
+                if (isPositionValid && result->Position.x != LOCATION_NULL)
```
Gymnasiast: Shouldn't we just call `isNull` on the coords?

duncanspumpkin: No it would be incorrect that checks for coords null this is location null. Also isnull is a function of coordsxyzd this is coordsxyz.

Gymnasiast: > No it would be incorrect that checks for coords null this is location null.

Why do we have two ways of marking null coords anyway?

> Also isnull is a function of coordsxyzd this is coordsxyz.

Then it should be made available on CoordsXY and CoordsXYZ. There is no reason why it should be CoordsXYZD-only.

src/openrct2/actions/RideEntranceExitPlaceAction.hpp
tupaschoal: How curious, I wonder if it fixes any bugs. Thanks for explaining :)

src/openrct2/ReplayManager.cpp
duncanspumpkin: While I agree I don't think I will do it in this PR. I need to work out what functions are using which version of null before doing that and it will not be pretty until it is resolved.

Gymnasiast: That is okay.

Gymnasiast:
I suppose this will need a network version bump because of the change discussed in @tupaschoal 's line note?

duncanspumpkin:
It doesn't need a network bump as it only changes the height of audio and money effects. As audio is client side there is no issue and this action is 0 cost so money effects do not do anything either.

duncanspumpkin:
Just to be safe I am going to network bump.

tupaschoal:
Helps #10267

PR #10399, opened 2019-12-19
Refactor MapAnimation to use CoordsXYZ

duncanspumpkin:
Found another instance where CoordsXYZ z component was in TileCoordsXYZ format. Fixed that as well at the same time.

### DIFF
--- a/src/openrct2/actions/LargeSceneryPlaceAction.hpp
+++ b/src/openrct2/actions/LargeSceneryPlaceAction.hpp
@@ -313,7 +313,7 @@ public:
             TileElement* newTileElement = tile_element_insert(
                 { curTile.x / 32, curTile.y / 32, zLow }, quarterTile.GetBaseQuarterOccupied());
             Guard::Assert(newTileElement != nullptr);
-            map_animation_create(MAP_ANIMATION_TYPE_LARGE_SCENERY, curTile.x, curTile.y, zLow);
+            map_animation_create(MAP_ANIMATION_TYPE_LARGE_SCENERY, { curTile, zLow * 8 });
             newTileElement->SetType(TILE_ELEMENT_TYPE_LARGE_SCENERY);
             newTileElement->clearance_height = zHigh;
             auto newSceneryElement = newTileElement->AsLargeScenery();
--- a/src/openrct2/world/MapAnimation.cpp
+++ b/src/openrct2/world/MapAnimation.cpp
@@ -37,7 +37,7 @@ static bool DoesAnimationExist(int32_t type, const CoordsXYZ& location)
 {
     for (const auto& a : _mapAnimations)
     {
-        if (a.type == type && a.location.x == location.x && a.location.y == location.y && a.location.z == location.z)
+        if (a.type == type && a.location == location)
         {
             // Animation already exists
             return true;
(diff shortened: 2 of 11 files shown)


src/openrct2/actions/LargeSceneryPlaceAction.hpp
```diff
@@ -313,7 +313,7 @@ DEFINE_GAME_ACTION(LargeSceneryPlaceAction, GAME_COMMAND_PLACE_LARGE_SCENERY, La
             TileElement* newTileElement = tile_element_insert(
                 { curTile.x / 32, curTile.y / 32, zLow }, quarterTile.GetBaseQuarterOccupied());
             Guard::Assert(newTileElement != nullptr);
-            map_animation_create(MAP_ANIMATION_TYPE_LARGE_SCENERY, curTile.x, curTile.y, zLow);
+            map_animation_create(MAP_ANIMATION_TYPE_LARGE_SCENERY, { curTile, zLow * 8 });
```
tupaschoal: I think there should be `z` and then `zLow = z/8`, there are multiple places above doing `zLow * 8`

(This will probably apply to other files as well, but not sure if worth doing in this PR)

src/openrct2/world/MapAnimation.cpp
```diff
@@ -88,14 +87,14 @@ void map_animation_invalidate_all()
  *
  *  rct2: 0x00666670
  */
-static bool map_animation_invalidate_ride_entrance(int32_t x, int32_t y, int32_t baseZ)
+static bool map_animation_invalidate_ride_entrance(CoordsXYZ loc)
 {
-    auto tileElement = map_get_first_element_at(x >> 5, y >> 5);
+    auto tileElement = map_get_first_element_at(loc.x >> 5, loc.y >> 5);
```
tupaschoal: What I've been doing is creating a `auto tileLoc = TileCoordsXYZ{loc}`. Then all the shifting is implied and the code also is more clear on where you're using `loc` or `tileLoc`. I guess it's a good format/convention to stick to (and applies to most of the changes in this file)

src/openrct2/actions/LargeSceneryPlaceAction.hpp
duncanspumpkin: I don't think I'll address that in this PR. I need to come up with a standard for the z's. I guess we could standardise around using `z` for big Z and `baseZ`, `clearZ` or `lowZ`,`highZ` for small Z.

src/openrct2/world/MapAnimation.cpp
```diff
@@ -88,14 +87,15 @@ void map_animation_invalidate_all()
...
 {
-    auto tileElement = map_get_first_element_at(x >> 5, y >> 5);
+    TileCoordsXY tileLoc{ loc };
+    auto tileElement = map_get_first_element_at(tileLoc.x, tileLoc.y);
     if (tileElement == nullptr)
         return true;
     do
     {
-        if (tileElement->base_height != baseZ)
+        if (tileElement->base_height != loc.z / 8)
```
tupaschoal: This (and all the others) can now become `tileLoc.z`

src/openrct2/actions/LargeSceneryPlaceAction.hpp
tupaschoal: Fair enough

PR #10401, opened 2019-12-20
Standardize Coords and TileCoords objects

tupaschoal:
- Made all `TileCoords` inherit from each other
- Moved `isNull()` to each of their base classes
- Moved variable declarations to the top
- Used delegating ctors

### DIFF
--- a/src/openrct2/world/Location.hpp
+++ b/src/openrct2/world/Location.hpp
@@ -149,27 +149,38 @@ struct CoordsXY
     {
         return { floor2(x, 32), floor2(y, 32) };
     }
+
+    bool isNull() const
+    {
+        return x == COORDS_NULL;
+    };
 };
 
 struct TileCoordsXY
 {
+    int32_t x = 0;
+    int32_t y = 0;
+
     TileCoordsXY() = default;
     constexpr TileCoordsXY(int32_t x_, int32_t y_)
         : x(x_)
         , y(y_)
     {
     }
+
     explicit TileCoordsXY(CoordsXY c)
         : x(c.x / 32)
         , y(c.y / 32)
     {
     }
+
     TileCoordsXY& operator+=(const TileCoordsXY rhs)
     {
         x += rhs.x;
         y += rhs.y;
         return *this;
     }
+
     TileCoordsXY& operator-=(const TileCoordsXY rhs)
     {
         x -= rhs.x;
@@ -209,7 +220,10 @@ struct TileCoordsXY
         return rotatedCoords;
     }
 
-    int32_t x = 0, y = 0;
+    bool isNull() const
+    {
+        return x == COORDS_NULL;
+    };
 };
 
 struct CoordsXYZ : public CoordsXY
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Location.hpp
```diff
@@ -149,27 +149,38 @@ struct CoordsXY
     {
         return { floor2(x, 32), floor2(y, 32) };
     }
+
+    bool isNull() const
+    {
+        return x == COORDS_NULL;
+    };
```
Gymnasiast: While we're at it, could you also define a setNull() function?

duncanspumpkin:
The only problem being that `IsNull` wont work for the vast amount of items as `COORDS_NULL` is defined as `-1` and `LOCATION_NULL` and most instance are expecting `0x8000`.

src/openrct2/world/Location.hpp
AaronVanGeffen: I'm not sure I like that, semantically. Is that a common use-case? Perhaps re-assigning from a singleton would be more appropriate?

tupaschoal:
> The only problem being that `IsNull` wont work for the vast amount of items as `COORDS_NULL` is defined as `-1` and `LOCATION_NULL` and most instance are expecting `0x8000`.

This is not a problem that the changeset is inserting, right? I can work on other PRs unifying things to `COORDS_NULL`.

src/openrct2/world/Location.hpp
tupaschoal: Unsure if I create the `setNull()` or not

duncanspumpkin:
We can't unify without breaking a few things (replays, saves, network, ...)

duncanspumpkin:
Also -1 is a valid value for a lot of x coordinates. I think its a bad value to use for null.

src/openrct2/world/Location.hpp
Gymnasiast: @AaronVanGeffen Code that uses Coords should not have too intimate knowledge of how to mark coordinates as null. That's why it needs to be a function. Adding a `setNull()` function is the most straightforward way to do so.

Gymnasiast:
@duncanspumpkin It is already *used* for null. All this PR does is make it more broadly available.

Also: how is a negative x coordinate valid?

duncanspumpkin:
It is only used for null in import and export code. In normal play `LOCATION_NULL` is used. In normal play it is quite reasonable to have negative coordinates as we do all sort of sums with them. Move peep one tile left is done with a negative coord for example.

duncanspumpkin:
I think we should change `LocationXY16` to use this `-1` null and make `CoordsXY` and `TileCoordsXY` take the `0x8000` null.

Gymnasiast:
> It is only used for null in import and export code. In normal play LOCATION_NULL is used.

A cursory search through the codebase shows that that is simply untrue. There are *loads* of places where this is used already.

Here, for example: https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2-ui/windows/Ride.cpp#L2833-L2836

duncanspumpkin:
That is due to the import and export code though. Its is trivial to fix the entrances not conforming to the pattern that the rest sticks to.

duncanspumpkin:
hmm some of the import code also use 0x8000. Okay the whole things a mess.

Gymnasiast:
Besides `LOCATION_NULL` and `COORDS_NULL`, there is also `RCT_XY8_UNDEFINED` (defined as 0xFFFF).

Now we have all the information on the table, we can talk about how to solve it. I agree that we need to have *one* way of marking coordinates as NULL. I'm also not very fussed as to what that value is, as long as it doesn't cause valid coordinates to be interpreted as null. (Perhaps we should set both x and y to a specific value.)

tupaschoal:
> This is being approved but please do not use the IsNull function just yet until we resolve the issues related to `-1`, `0x8000` which will be addressed in the next few PR's.

ok

PR #10402, opened 2019-12-21
Android: fix Gradle, CMake, asset copying; upgrade to AndroidX, fullscreen mode

NuclearFej:
Android build now functions again. Compiled on Windows, tested on a Pixel 2 XL running Android 10.

Gradle now pulls from the main OpenRCT2 /bin/data folder to insert the assets into the APK. MainActivity now properly copies assets from within the APK. Dependencies have been updated and the support library has been upgraded to AndroidX. CMake libraries now link properly. Network (multiplayer), Discord, and HTTP_Twitch were disabled, in addition to the already-disabled HTTP, OpenGL, and Twitch, due to compile-time errors. Revision number incremented.

-Werror and -Wall were disabled (in Android only) for the time being due to compile-time errors caused by the disabled features.

### DIFF
--- a/src/openrct2-android/app/build.gradle
+++ b/src/openrct2-android/app/build.gradle
@@ -6,7 +6,7 @@ android {
 
     defaultConfig {
         applicationId 'website.openrct2'
-        minSdkVersion 16
+        minSdkVersion 19
         targetSdkVersion 29
 
         versionCode 2
@@ -94,6 +94,14 @@ dependencies {
     implementation 'commons-io:commons-io:2.6'
     implementation 'androidx.appcompat:appcompat:1.1.0'
     implementation 'com.google.android.gms:play-services-analytics:17.0.0'
+}
+
+allprojects {
+    gradle.projectsEvaluated {
+        tasks.withType(JavaCompile) {
+            options.compilerArgs << "-Xlint:deprecation"
+        }
+    }
 }
 
 apply plugin: 'com.google.gms.google-services'
--- a/src/openrct2-android/app/build.gradle
+++ b/src/openrct2-android/app/build.gradle
@@ -1,16 +1,16 @@
 apply plugin: 'com.android.application'
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion '28.0.2'
+    compileSdkVersion 29
+    buildToolsVersion '29.0.2'
 
     defaultConfig {
         applicationId 'website.openrct2'
         minSdkVersion 16
-        targetSdkVersion 25
+        targetSdkVersion 29
 
         versionCode 2
-        versionName '0.2.4'
+        versionName '0.2.5'
 
         externalNativeBuild {
             cmake {
@@ -18,6 +18,10 @@ android {
                 targets 'openrct2', 'openrct2-ui', 'openrct2-cli'
             }
         }
+    }
+
+    sourceSets {
+        main.assets.srcDirs += 'src/main/assets/'
     }
 
     signingConfigs {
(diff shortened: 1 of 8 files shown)


AaronVanGeffen:
Thanks for your work on this. It's nice to see the Android port getting some love.

src/openrct2-android/app/build.gradle
```diff
@@ -77,41 +81,19 @@ android {
...
-        }
-        download {
-            src 'https://github.com/OpenRCT2/objects/releases/download/v1.0.12/objects.zip'
-            dest new File(buildDir, 'objects.zip')
-        }
-        copy {
-            from zipTree(new File(buildDir, 'objects.zip'))
-            into "$variant.mergeAssets.outputDir/data/object"
-        }
-    }
```
AaronVanGeffen: These are essential steps for generating (release) builds) in CI jobs. Why were these steps removed? It seems the code that replaces it does not do the same thing.

NuclearFej:
The download plugin does not work anymore. My best guess is that it doesn't support TLSv1.2 which GitHub now requires. These assets are all duplicated elsewhere in the build, so I just took them from there.

variant.mergeassets.outputDir no longer exists and I'm not sure why it was used in the first place - these files are copied out of the APK and into /sdcard/openrct2 and that's it, so I just placed the files into the standard assets folder which is included in all builds.

src/openrct2-android/app/build.gradle
NuclearFej: The download plugin does not work anymore. My best guess is that it doesn't support TLSv1.2 which GitHub now requires. These assets are all duplicated elsewhere in the build, so I just took them from there.

variant.mergeassets.outputDir no longer exists and I'm not sure why it was used in the first place - these files are copied out of the APK and into /sdcard/openrct2 and that's it, so I just placed the files into the standard assets folder which is included in all builds.

NuclearFej:
One thing to note about fullscreen mode is that phones with rounded corners cut off bits of the UI elements in those positions. It ought to be fixed at some point but the effect is minor and the trade-off is well worth it for the extra screen real estate.

Gymnasiast:
Since @AaronVanGeffen appears to be busy and hasn't replied to the line note, I figured I'd merge it - after all, the Android port was in bad shape so there wasn't much that could go wrong here.

Gymnasiast:
Unfortunately, the CI job still fails: 

```
 /usr/local/android-sdk/ndk-bundle/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/include/c++/v1/filesystem:1070: error: undefined reference to 'std::__ndk1::__fs::filesystem::path::__extension() const'
```

https://travis-ci.org/OpenRCT2/OpenRCT2/jobs/636595094?utm_medium=notification&utm_source=github_status

AaronVanGeffen:
Sounds like it may require explicit linking in a way similar to #10522

NuclearFej:
Sorry for taking a while to respond -

Which version of NDK is being used for the builds? It does NOT work on the latest version. I am using NDK 18.

AaronVanGeffen:
Does this help?

https://github.com/OpenRCT2/OpenRCT2/blob/f1dcf8573138116e80544433e34548a952d06028/.travis.yml#L127

https://github.com/OpenRCT2/OpenRCT2/blob/f1dcf8573138116e80544433e34548a952d06028/.travis.yml#L135-L136

NuclearFej:
It's targeting API 28, not 25.

AaronVanGeffen:
Then these paths will require updating. Could you PR?

Invictaz:
@NuclearFej Do you mean it will require Android 9? That's ruling a lot of devices out.

janisozaur:
@Invictaz that also enables a whole lot more devices than no builds. We'd welcome a PR to use older target.

Invictaz:
After inspection it targets minver 4.4 so I hope the online inspection tool I used is correct.
Backporting stuff is an insanely huge workload for an Android build that is mostly neglected.

Gymnasiast:
I took this screenshot on an Samsung SM-G110H running Android 4.4.2. As you can see, it works:

Screenshot_2020-02-07-23-24-38

Invictaz:
@Gymnasiast 

Greatly appreciated

NuclearFej:
Good to see. Honestly, I wasn't worried about supporting devices that old, but if nothing requires raising the min SDK level then I wouldn't change it.

Invictaz:
Please don't. I come by piles of "older" Android devices wearing Android 4. 
In essence, they are from 2014 and were never updated. 2014 isn't that old imo.

Gymnasiast:
@NuclearFej Judging from the IDs, it looks like the Android version is compiled as debug. Could you  verify if that's the case, and if so, change it to `-O2` or something similar?

src/openrct2-android/app/build.gradle
```diff
@@ -1,16 +1,16 @@
...
     defaultConfig {
         applicationId 'website.openrct2'
-        minSdkVersion 16
-        targetSdkVersion 25
+        minSdkVersion 19
+        targetSdkVersion 29
 
         versionCode 2
-        versionName '0.2.4'
+        versionName '0.2.5'
```
janisozaur: This should not have been modified.

PR #10407, opened 2019-12-21
Fixing COORDS_NULL: Refactor stations

duncanspumpkin:
Moving the internal stations.Start to TileCoordsXY so that we can use the preferred NULL value in next refactor.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -3289,14 +3289,12 @@ static void ride_entrance_exit_connected(Ride* ride)
 
 static void ride_shop_connected(Ride* ride)
 {
-    LocationXY8 coordinates = ride->stations[0].Start;
-    if (coordinates.isNull())
+    TileCoordsXY shopLoc = ride->stations[0].Start;
+    if (shopLoc.isNull())
         return;
 
-    TileCoordsXY loc = { coordinates.x, coordinates.y };
-
     TrackElement* trackElement = nullptr;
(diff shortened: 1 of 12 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -6657,12 +6655,11 @@ static int32_t ride_get_track_length(Ride* ride)
 
     for (int32_t i = 0; i < MAX_STATIONS && !foundTrack; i++)
     {
-        LocationXY8 location = ride->stations[i].Start;
```
Gymnasiast: Saving the station start into a variable makes the code easier to read IMO.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9282,16 +9282,14 @@ static int32_t vehicle_update_track_motion_mini_golf(rct_vehicle* vehicle, int32
 
 loc_6DCE68:
     _vehicleMotionTrackFlags |= VEHICLE_UPDATE_MOTION_TRACK_FLAG_VEHICLE_AT_STATION;
-    regs.al = vehicle->track_x >> 5;
-    regs.ah = vehicle->track_y >> 5;
-    regs.dl = vehicle->track_z >> 3;
```
tupaschoal: maybe if you did
```c++
auto vehicleTileCoords = TileCoordsXYZ{CoordsXYZ{vehicle->track_x, vehicle->track_y, vehicle->track_z}};
```

It would make the comparisons below more readable

tupaschoal:
Helps #10267

PR #10409, opened 2019-12-21
Make Map::map_set_tile_elements() use TileCoordsXY

tupaschoal:


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2435,20 +2435,19 @@ static void sub_6CBCE2(
...
-        _backupTileElementArrays[1] = map_get_first_element_at(tileX + 1, tileY + 0);
-        _backupTileElementArrays[2] = map_get_first_element_at(tileX - 1, tileY + 0);
-        _backupTileElementArrays[3] = map_get_first_element_at(tileX + 0, tileY + 1);
-        _backupTileElementArrays[4] = map_get_first_element_at(tileX + 0, tileY - 1);
-        map_set_tile_elements(tileX + 0, tileY + 0, &_tempTrackTileElement);
-        map_set_tile_elements(tileX + 1, tileY + 0, &_tempSideTrackTileElement);
-        map_set_tile_elements(tileX - 1, tileY + 0, &_tempSideTrackTileElement);
-        map_set_tile_elements(tileX + 0, tileY + 1, &_tempSideTrackTileElement);
-        map_set_tile_elements(tileX + 0, tileY - 1, &_tempSideTrackTileElement);
+        _backupTileElementArrays[0] = map_get_first_element_at(tileCoords.x + 0, tileCoords.y + 0);
```
duncanspumpkin: Can some of this be refactored to use `TileDirectionDelta` not sure if it would be more readable.

src/openrct2/world/Map.cpp
```diff
@@ -191,14 +191,14 @@ TileElement* map_get_nth_element_at(int32_t x, int32_t y, int32_t n)
     return nullptr;
 }
 
-void map_set_tile_elements(int32_t x, int32_t y, TileElement* elements)
+void map_set_tile_elements(const TileCoordsXY& tilePos, TileElement* elements)
```
duncanspumpkin: This is really only setting a single tile element. Perhaps it should be renamed.

src/openrct2-ui/windows/RideConstruction.cpp
tupaschoal: I think it would, but I have an incoming PR that makes `map_get_first_element_at` receive a single Coords object, it would be clearer then, if we can delay it.

tupaschoal: Just published #10413. If this is merged, I'll make the changes to use `TileDirectionDelta` there

src/openrct2/world/Map.cpp
Gymnasiast: Yes and no. It sets the pointer to the first element of the tile, which can then be `++`'ed to get to the other tiles. Basically it replaces one chain of elements with another.

PR #10412, opened 2019-12-21
Remove MakeXY16

Gymnasiast:


src/openrct2-ui/windows/Map.cpp
```diff
@@ -1639,7 +1639,8 @@ static void map_window_set_pixels(rct_window* w)
     int32_t x = 0, y = 0, dx = 0, dy = 0;
 
     int32_t pos = (_currentLine * (MAP_WINDOW_MAP_SIZE - 1)) + MAXIMUM_MAP_SIZE_TECHNICAL - 1;
-    LocationXY16 destinationPosition = MakeXY16(pos % MAP_WINDOW_MAP_SIZE, pos / MAP_WINDOW_MAP_SIZE);
+    LocationXY16 destinationPosition = { static_cast<int16_t>(pos % MAP_WINDOW_MAP_SIZE),
```
duncanspumpkin: This has no need to be a `LocationXY16` we can safely change it to a `TileCoordsXY`.

Gymnasiast: I haven't done that because it's not a tile coordinate (nor a big coordinate). The very idea of TileCoords and Coords is that you can be absolutely certain about what they contain.

duncanspumpkin: Good point.

PR #10414, opened 2019-12-21
Create getters for GetBaseHeight()/GetClearanceHeight()

Gymnasiast:
This creates getters for base height and clearance height. The value returned is a "sprite z", not a "tile z". This makes moving everything to Coords easier.

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -505,7 +505,7 @@ int32_t viewport_interaction_right_click(ScreenCoordsXY screenCoords)
 static void viewport_interaction_remove_scenery(TileElement* tileElement, CoordsXY mapCoords)
 {
     auto removeSceneryAction = SmallSceneryRemoveAction(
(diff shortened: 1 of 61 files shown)


Gymnasiast:
Merge note: use squash and make sure the commit title is correct.

IntelOrca:
Only concern I have is that a dev might think oh, there is a getter for base_height, lets use that... oh wait this isn't just returning base_height. GetBaseZ might help differentiate since we don't have different typedefs for sprite Z and tile Z.

Gymnasiast:
> Only concern I have is that a dev might think oh, there is a getter for base_height, lets use that... oh wait this isn't just returning base_height. GetBaseZ might help differentiate since we don't have different typedefs for sprite Z and tile Z.

I see this as a stepping stone in moving everything towards `Coords`.

Having said that, I will change the function names to GetBaseZ/GetClearanceZ.

duncanspumpkin:
Is that the consistent naming we should be heading towards: `baseZ` means big Z and `baseHeight` means small Z?

PR #10416, opened 2019-12-21
New maze pathfinding

philgekni:
In the current maze pathfinding algorithm when a guest encounter a intersection, this one actually picks a random new direction. So, virtually more the maze contains intersections more difficult will be for the guests to get out, this independently of the complexity of the maze design.
Also it is a pretty common thing for guests to loop several time in the same area :

Considering such behaviors as visually silly (or even really dumb), giving a bad entrance throughput and presenting the risk for some guest to haunts forever moderately large mazes, I offer a new version of the maze pathfinding.

## New algorithm details
The principle of the new algorithm is based on path marking: at each intersection visitor will try to take a path that he doesn't already visited. A path is considered visited if all the dead-ends included into this path has been reached:

mazeExploreAllPathsGraph

### Direction picking and possible other features
As say before guests will try to avoid choosing paths already visited: each time before picking a direction, a random check is run to determinate if the guest succeed to remember already visited path.

Each guest can remember the visited paths at the 5 last encountered intersections.

The probability to pick a wrong way is the result of a random value compared to a threshold value, which depends of how long time the guest has visited this intersection the last time.

If a guest jump / try to peek his position, at the next intersection he is more likely to choose an unvisited path. Also the guests have less probability to choose the first edge from where they entered in the intersections (to encourage them to visit all the ways in a direction before retracing their steps).

After some discussion on the dev Gitter, here some features / changes suggestions :

- [x] Randomly forget some visited paths by the guest. (@tupaschoal).
   _Implemented: guests remember the last 5 intersections and forget the others._
- [x] The probability to pick a wrong way also based on the time passed since the guest visited the intersection the last time. (@IntelOrca).
  _Implemented: The probabilities change in function of the position of the intersection into the guest memory of the last intersections_.
- *Open to new suggestions.*

## Implementation details
This version of the maze pathfinding is based on the feature that each guest remembers details concerning the 5 last intersections encountered on their way.

mazeData

All the data used for this algorithm live inside the `Peep` structure in the `pathfind_history` fields (which are not used when the guest is in a ride) meaning that everything is saved into the save file and transmitted to new clients when joining a server.

## Results
mazeResultV2:CurrentAlgorithm:204guestPerHour,NewAlgoritm:924guestPerHour
(_Results after one year from the same save file and with the weather locked to "Sunny"._) 

According to my first observations, I observed an large improvement of the throughput of the mazes. Also guests inside the maze are less complaining about their desire to get out of the maze:

mazeThoughts

## Try by yourself
* [MazeTest.zip](https://github.com/OpenRCT2/OpenRCT2/files/5792795/MazeTest.zip): Include a save with some mazes I used to test this algorithm (new version 10/01/21).

## Roadmap
In side of the possibles review comments / changes and new ideas discussions here my agenda regarding this PR.

- [x] Plays with Peep pathfinding fields and write a new version of the algorithm.
- [x] Fix initial tests failures.
- [x] Coding style enforcement.
- [x] Add more tests / replace some `assert` to check in release possible errors situation due to hack done with cheating tools (like changing the height of a maze tile with the tile inspector).
- [ ] Tuning threshold values in regards of throughput and satisfaction results.


janisozaur:
There are failures highlighted in our CIs, please address them. Some compilation failures and some test suite failures.

philgekni:
Yup, on the way.

ZehMatt:
Heres what I think:
- For multiplayer when new clients join the list should not be cleared, players should not influence the game state like that.
- Storage can be probably simplified, the peep already has 4 slots for the path finding history, additionally you can also use pathfind_goal which makes a total of 5 junctions you could remember and allowing the data to be serialized in the save files. Having a total of 5 junctions in the peeps memory should already greatly improve some maze situations without having a perfect maze explorer.

philgekni:
I updated my branch with the last @tupaschoal `Map::map_get_track_element_at()` refactor and I took a look to the CIs.
All builds with tests fail due to MultiLaunch, result expected because for the moment I didn't yet implemented the data cleaning when the game is reloaded (see Roadmap). My only problem comes from one remaining compilation error on the Travis build job "Ubuntu i686 GCC Debug m32 no-build-shared-libs" (`undefined reference to 'map_get_track_element_at(CoordsXYZ const&)'`). As far I know I correctly include "world/Map.h" and didn't investigate yet too much.
@ZehMatt, thanks for the input and the suggestion about the pathfinding fields of the `Peep` structure (as I tried to avoid editing existing structures, I was too shy to dare to modify their values). I think that I will be indeed able to use these fields to save memory and to avoid inopportune data cleaning / regeneration in my code.
I will do some hack in this way in the next days and read a little the functions accessing to these values in order to understand what I risk to break.

janisozaur:
The link failure is in one of our special targets, `testpaint`. I think one of the team members can fix this for you.

philgekni:
I updated my branch with a new version of my algorithm with as key improvements:
- Complies with @ZehMatt concerns. Stop using extra memory which need to be wipe when reloading a game / a client joins the server. Everything lives inside the `Peep` structure and use `pathfind_history` memory. No memory allocation nor extra memory is needed.
- The new version is simpler. The intersection tracking is done by each guest for the last 5 encountered intersections. There is no more functions looping on maze tiles for data initialization or updating.
- I implemented in some ways the ideas that @tupaschoal and @IntelOrca gave me on Gitter (regarding paths and intersections forgetting).
- The guests are encouraged to discover new paths and not retracing their steps while they didn't completely visited the part of the maze where they are.
- The dead ends in form of loop are now detected and marked as visited after one loop.
- The debug panel of the guest window show the current state of the data used for the maze pathfinding.

I updated the PR description accordingly.

ZehMatt:
Looks pretty promising now except the coding style, please follow this guide regarding our code style: https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style

I'll get to test and review your code shortly, probably this weekend.

src/openrct2/peep/Guest.cpp
```diff
@@ -4909,17 +4915,35 @@ void Guest::UpdateRideMazePathfinding()
...
-        // TODO: Reimplement 'wrong way' probability using entry index
-        openEdgesFlags &= ~currentEntry.getVisited();
+        // TODO: Tune these values
+        static const std::array<std::array<uint16_t, 4>, 6> rememberThreshold = { {
+            // visited,           visited+peeked,     origin,             origin+peeked
+            { maze_set_odd(0.04), maze_set_odd(0.00), maze_set_odd(0.02), maze_set_odd(0.00) }, // Last intersection
+            { maze_set_odd(0.09), maze_set_odd(0.03), maze_set_odd(0.03), maze_set_odd(0.01) }, // -1
+            { maze_set_odd(0.27), maze_set_odd(0.07), maze_set_odd(0.18), maze_set_odd(0.10) }, // -2
+            { maze_set_odd(0.39), maze_set_odd(0.15), maze_set_odd(0.25), maze_set_odd(0.17) }, // -3
+            { maze_set_odd(0.50), maze_set_odd(0.31), maze_set_odd(0.45), maze_set_odd(0.32) }, // -4
```
pizza2004: Why is it that visiting is less effective than origin, but visited+peeked is more effective than origin+peeked a couple intersections back?  I would expect the relation between the two to be related, ie either you tend to forget your origin faster than your visited or peeking helps them both equally.

ZehMatt:
Sorry for the late response. I've tested this offline and online and seems to be working great, this improved the overall throughput for a few maze designs I tried out and had no issues. Before reviewing the code it would help if you could rebase and and change the coding style to match the base, overall pretty good job.

philgekni:
Thanks @ZehMatt, no worry for the delay, I have been myself occupied the last weeks. I'll find a moment this week to update my branch.

pizza2004:
This will need to be rebased again before it can be merged, but I'm just wondering if anyone has done anything more with this in the past month?

philgekni:
Hi! No you had the right feeling, I haven't had time to work on this PR these last week but I'm now available.

I currently struggle to apply correctly the coding style. I believe that `clang-format` returns advises which are conflicting with the coding style, like:
```diff
 @@ -342,9 +342,8 @@
         item.free();
 }
 
-std::tuple<MazePathfindingEntry&, uint8_t, bool> MazePathfindingHistory::meetIntersection(const CoordsXY& coords,
-                                                                                          uint8_t subTileIndex,
-                                                                                          Direction source)
+std::tuple<MazePathfindingEntry&, uint8_t, bool> MazePathfindingHistory::meetIntersection(
+    const CoordsXY& coords, uint8_t subTileIndex, Direction source)
 {
     uint8_t i;
     bool hasPeeked = Stack[0].hasPeeked();
```
with from the [coding-style doc](https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style):
```c
/// If a function definition runs over the line length limit, the arguments
/// should be broken up as follows:
void ShowAnimation(sint32 columns,
                   sint32 rows,
                   sint32 isResizable)
{
```

Let me know if I miss something.

pizza2004:
If @philgekni no longer has time to work on this I might be interested in cleaning up the formatting and such so it could be merged and then any other improvements could be submitted later as new PRs.

src/openrct2/peep/Guest.cpp
philgekni: You right, the odds relations are her inconstant and I fixed that in 65573e6.
Regarding the fact that guests tends to forget less easily the origin than other visited directions, this is to encourage them to explore completely an area before turning back and so avoiding multiples travels between 2 area of the maze.

philgekni:
Hi! I think that my PR is enough mature after the last refactorings and cleanings.
I tested the network locally and didn't encountered desync and also updated the mazeTest save (in the description) with new 'testcases'. If needed I can join a new version of the bpb1000Ticks replay taking in account the new guests behavior.

tupaschoal:
Nice. We'll probably be releasing a version this weekend, so I marked it as after 0.3.0, give us a couple of days

src/openrct2/ride/gentle/Maze.h
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Data linked to an intersection perviously encountered by a guest.
+ */
+class MazePathfindingEntry
+{
+    union
+    {
+        struct
+        {
+            uint32_t SubTileIdx : 2;       // Index of the subTile corresponding to the intersection
```
duncanspumpkin, line 60: Do not use bit masks in the code. It is not multi-platform friendly. You will find that we never use them across the whole codebase for this reason. Instead we have lots of getters and setters. I'm not a fan of stuffing all this information into the struct like this. I would much rather we wait for the new save format and used individual variables of a sensible size.

pizza2004: If it's gonna be much longer before the new save file, it might be nice to just find a cross platform way to make it work in the meantime for the sake of allowing it to be merged.

philgekni: Yeah, indeed, the bitfields are implementation dependent, my bad. So I think for the moment reverting aeba1ff and continuing to use bitwise ops, shifts and masks until the new save format. That should fixes the trouble regarding on how the different compilers want to organize the memory and lets only the endianness which is handled by the serialization module.

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+    assert(coords.x > 0 && coords.y > 0);
+
+    reset();
+    SubTileIdx = subTileIndex;
+    Origin = dirOrigin;
+    XTile = coords.x / COORDS_XY_STEP;
+    YTile = coords.y / COORDS_XY_STEP;
+}
+
+std::tuple<CoordsXY, uint8_t> MazePathfindingEntry::getCoords() const
```
ZehMatt, line 234: Use std::pair for pairs

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+        item.reset();
+}
+
+void MazePathfindingHistory::free()
+{
+    for (auto& item : Stack)
+        item.free();
+}
+
+std::tuple<MazePathfindingEntry&, uint8_t> MazePathfindingHistory::meetIntersection(
```
ZehMatt, line 290: Use std::pair for pairs

ZehMatt:
We generally use PascalCase for the function names

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
 
     return maze_paint_setup;
 }
+
+#pragma region Maze Pathfinding
+
+void swap(MazePathfindingEntry& a, MazePathfindingEntry& b)
```
ZehMatt, line 207: I don't think this function should exist

src/openrct2/ride/gentle/Maze.cpp
IntelOrca: What is the benefit of using pair of tuple?

ZehMatt: The compiler can optimize std::pair knowing it has two elements potentially into a single register, this doesn't seem to work with tuples.

IntelOrca: The optimisation problem is a shame. Otherwise `std::pair` would be obsolete since `std::tuple` does the same thing but also allows more things. `std::get<N>` is the way I get items from both pair and tuple.

ZehMatt: I find tuples to be generally a bad concept, just use a struct if you have more than 2 things. Having std::get<0>(var) is not exactly great code to read and one can only guess what the field could be.

duncanspumpkin: You shouldn't need to use get<...> Very often when you could use structured bindings.
```
auto [var1, var2] = functionReturningTupleOrStructOrPair();```

Comment (ZehMatt) [reply src/openrct2/ride/gentle/Maze.cpp:234]: This is true but that still means the caller needs to be aware of whats returned, all solved by structs.

File: src/openrct2/peep/Guest.cpp [hunk 6]
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
+        PeepMazeRegister.DirectionAtLastIntersection = chosenEdge;
+
     targetLoc.x = CoordsDirectionDelta[chosenEdge].x / 2;
     targetLoc.y = CoordsDirectionDelta[chosenEdge].y / 2;
 
     targetLoc.x += DestinationX;
     targetLoc.y += DestinationY;
 
-    enum class maze_type
+    enum class maze_next_tile_type
```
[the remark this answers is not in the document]
philgekni: This enum is used locally (so that why I permitted me this modification) and I changed its name to reflect better that the code searches to characterize the type of the next tile and not the type of a maze in its entirety.

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
 
     return maze_paint_setup;
 }
+
+#pragma region Maze Pathfinding
+
+void MazePathfindingEntry::Swap(MazePathfindingEntry& entry)
```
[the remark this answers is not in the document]
philgekni: Right, I don't know why I was so attached to this function. Fixed by 1faeb39.

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+    for (auto& item : Stack)
+        item.Free();
+}
+
+std::pair<MazePathfindingEntry&, uint8_t> MazePathfindingHistory::MeetIntersection(
+    const CoordsXY& coords, uint8_t subTileIndex, Direction source)
+{
+    uint8_t i;
+
+    for (i = 0; i < 5; ++i)
```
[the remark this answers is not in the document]
philgekni: Fixed by 1faeb39.

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+    {
+        if (Stack[i].MatchCoords(coords, subTileIndex))
+            break;
+    }
+
+    for (int j = std::min<int>(4, i); j > 0; --j)
+        Stack[j - 1].Swap(Stack[j]);
+
+    if (i == 5)
```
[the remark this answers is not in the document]
philgekni: Fixed by 1faeb39.

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+{
+    uint8_t i;
+
+    for (i = 0; i < 5; ++i)
+    {
+        if (Stack[i].MatchCoords(coords, subTileIndex))
+            break;
+    }
+
+    for (int j = std::min<int>(4, i); j > 0; --j)
```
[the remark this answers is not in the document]
philgekni: Fixed by 1faeb39.

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+        Stack[j - 1].Swap(Stack[j]);
+
+    if (i == 5)
+        Stack[0].Set(coords, subTileIndex, source);
+
+    return std::pair(std::ref(Stack[0]), i);
+}
+
+std::optional<std::reference_wrapper<MazePathfindingEntry>> MazePathfindingHistory::UpdateIntersectionEntry(
```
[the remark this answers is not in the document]
philgekni: This name was to precise the R/W access to the entry. Anyway it looks that I don't use anymore this function and so cleaned it (71352ba).

src/openrct2/peep/Guest.cpp
```diff
@@ -4859,19 +4858,22 @@ void Guest::UpdateRideMazePathfinding()
     }
 
     auto ride = get_ride(CurrentRide);
-    if (ride == nullptr)
+    if (!ride)
```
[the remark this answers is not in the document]
philgekni: I updated the commit ca6c8fc in this way as well as make explicit another condition with a pointer at line 4894.

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,132 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+
+    return {};
+}
+
+const MazePathfindingEntry& MazePathfindingHistory::GetLast() const
+{
+    return Stack[0];
+}
+
+const std::array<MazePathfindingEntry, 5>& MazePathfindingHistory::ReadHistory() const
```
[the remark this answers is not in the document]
philgekni: As you wish, fixed by 71352ba.

src/openrct2/ride/gentle/Maze.cpp
philgekni: Fixed by Fixed by 1faeb39.

src/openrct2/ride/gentle/Maze.cpp
philgekni: Fixed by ebb203f.

src/openrct2/ride/gentle/Maze.cpp
philgekni: Fixed by ebb203f.

ZehMatt:
Well looking much better now, the bitfields still need to be addressed however.

src/openrct2/peep/Guest.cpp
```diff
@@ -4830,21 +4837,13 @@ void Guest::UpdateRideLeaveSpiralSlide()
...
-    { 4, 5, 6, 2 },
-    { 6, 8, 9, 10 },
-    { 14, 10, 12, 13 },
-};
+/**
+ * Helper function to define maze pathfinding thresholds
+ */
+static constexpr uint16_t maze_forget_odds(double odds)
+{
+    return odds * UINT16_MAX;
```
ZehMatt, line 4845: I'm a bit worried that different compilers can produce different results here. This either needs a way to validate that the desired outcome is stored or better yet using integers only.

ZehMatt:
Sorry for the late response, I left one last comment otherwise looks good to me. Given that this is a rather larger change some of the other team members should also review this.

src/openrct2/peep/Guest.cpp
tupaschoal:
```suggestion
    return odds * std::numeric_limits<uint16_t>::max();
```

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
     }
 
     CoordsXY targetLoc = { DestinationX & 0xFFE0, DestinationY & 0xFFE0 };
+    uint16_t stationBaseZ = ride->stations[0].GetBaseZ();
```
tupaschoal, line 4885:
```suggestion
    uint16_t stationBaseZ = static_cast<uint16_t>(ride->stations[0].GetBaseZ());
```

`GetBaseZ()` is an `int32_t`, implicitly casting to `uint16_t` might be a bad idea. Do you need to remove the signed though? (it was previously `int16_t`)

You only use it for the `map_get_track_element_at` which expects a `CoordsXYZ` object, so I'd say just call this directly there, without casting and such

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
-    int16_t stationBaseZ = ride->stations[0].GetBaseZ();
-
-    // Find the station track element
-    auto trackElement = map_get_track_element_at({ targetLoc, stationBaseZ });
-    if (trackElement == nullptr)
+    // Get info on open edges at the current subTile
     {
-        return;
-    }
+        TrackElement* trackElement = map_get_track_element_at({ targetLoc, stationBaseZ });
```
tupaschoal, line 4893:
```suggestion
        auto trackElement = map_get_track_element_at({ targetLoc, stationBaseZ });
```

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
-    {
-        if (!(mazeEntry & (1 << _MazeCurrentDirectionToOpenHedge[Var37 / 4][i])) && i != mazeReverseLastEdge)
+        uint16_t mazeEntry = trackElement->GetMazeEntry();
+        auto nearEntranceTile = ride->stations[0].Entrance.ToCoordsXYZD();
+        Direction entranceDirection = nearEntranceTile.direction;
+
+        // If the subTile is connected to the maze entrance, close the edge giving to it
+        nearEntranceTile += CoordsDirectionDelta[entranceDirection];
+        entranceDirection = direction_reverse(entranceDirection);
+        if (nearEntranceTile.x == targetLoc.x && nearEntranceTile.y == targetLoc.y
```
tupaschoal, line 4904:
```suggestion
        if (nearEntranceTile == targetLoc
```

This likely works

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
+            && (subTileIndex == entranceDirection || subTileIndex == direction_next(entranceDirection)))
+        {
+            mazeEntry |= 0x1 << gMazeCurrentDirectionToHedge[subTileIndex][direction_reverse(nearEntranceTile.direction)];
```
tupaschoal, line 4907:
```suggestion
            mazeEntry |= 0x1 << gMazeCurrentDirectionToHedge[subTileIndex][entranceDirection];
```
`entranceDirection` is already `direction_reverse(nearEntranceTile.direction)`, is it not?

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
+        }
+
+        // Check open edges (without hedge)
+        for (int dir = 3; dir >= 0; --dir)
```
tupaschoal, line 4911: Do we need to iterate reversely? If not it would be great to use `ALL_DIRECTIONS` array,

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
+        auto [currentEntry, entryIndex] = MazePathfindHistory.MeetIntersection(targetLoc, subTileIndex, sourceDirection);
+        bool peekFlag = PeepMazeRegister.CheckAndResetPeekFlag();
+
+        // Mark the last edge as visited if previously in a dead-end or the previous intersection completly visited
+        if (currentEntry.MatchCoords(previousEntry) || previousEntry.IsCompletlyVisited())
         {
-            return;
+            if (currentEntry.IsCompletlyVisited())
+            {
+                // Offer amnesia to completly lost guest
```
tupaschoal, line 4940:
```suggestion
                // Offer amnesia to completly lose guest
```
?

src/openrct2/peep/Guest.cpp
```diff
@@ -4830,21 +4837,13 @@ void Guest::UpdateRideLeaveSpiralSlide()
...
-static constexpr const uint8_t _MazeCurrentDirectionToOpenHedge[][4] = {
-    { 1, 2, 14, 0 },
-    { 4, 5, 6, 2 },
-    { 6, 8, 9, 10 },
-    { 14, 10, 12, 13 },
-};
+/**
+ * Helper function to define maze pathfinding thresholds
+ */
+static constexpr uint16_t maze_forget_odds(double odds)
```
tupaschoal, line 4843:
```suggestion
static constexpr uint16_t MazeForgetOdds(double odds)
```

We're slowly converting everything to TitleCase, so better not adding any function in snake case.

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
+              // -3
+              { maze_forget_odds(0.39), maze_forget_odds(0.25), maze_forget_odds(0.15), maze_forget_odds(0.10) },
+              // -4
+              { maze_forget_odds(0.50), maze_forget_odds(0.45), maze_forget_odds(0.32), maze_forget_odds(0.30) },
+              // New or forgotten intersection
+              { maze_forget_odds(1.00), maze_forget_odds(1.00), maze_forget_odds(1.00), maze_forget_odds(1.00) } }
+        };
+
+        // If the guest succeeded his memory roll, remove visited edges (other that the origin edge)
+        if ((scenario_rand() & 0xffff) >= rememberThreshold[entryIndex][peekFlag * 2])
```
tupaschoal, line 4971:
```suggestion
        if ((scenario_rand() & 0xFFFF) >= rememberThreshold[entryIndex][peekFlag * 2])
```

src/openrct2/peep/Guest.cpp
```diff
@@ -4880,55 +4882,131 @@ void Guest::UpdateRideMazePathfinding()
...
+            openEdgesFlags &= ~currentEntry.GetVisited();
+
+        // Second dice roll, if succeeded remove also the origin edge (unless all the other edges are already visited)
+        if (bitcount(openEdgesFlags) > 1 && (scenario_rand() & 0xffff) >= rememberThreshold[entryIndex][1 + peekFlag * 2])
```
tupaschoal, line 4975:
```suggestion
        if (bitcount(openEdgesFlags) > 1 && (scenario_rand() & 0xFFFF) >= rememberThreshold[entryIndex][1 + peekFlag * 2])
```

src/openrct2/ride/gentle/Maze.h
```diff
@@ -0,0 +1,178 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2019 OpenRCT2 developers
```
tupaschoal, line 2:
```suggestion
 * Copyright (c) 2014-2020 OpenRCT2 developers
```

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,223 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+}
+
+bool MazePathfindingEntry::MatchCoords(const MazePathfindingEntry& e) const
+{
+    return (GetXTile() == e.GetXTile()) && (GetYTile() == e.GetYTile()) && (GetSubTileIndex() == e.GetSubTileIndex());
+}
+
+void MazePathfindingHistory::Initialize()
+{
+    for (auto& item : Stack)
```
tupaschoal, line 383:
```suggestion
    for (const auto& item : Stack)
```

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,223 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+        item.Reset();
+}
+
+void MazePathfindingHistory::Free()
+{
+    for (auto& item : Stack)
```
tupaschoal, line 389:
```suggestion
    for (const auto& item : Stack)
```

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,223 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+std::pair<MazePathfindingEntry&, uint8_t> MazePathfindingHistory::MeetIntersection(
+    const CoordsXY& coords, uint8_t subTileIndex, Direction source)
+{
+    uint8_t i;
+
+    for (i = 0; i < Stack.size(); ++i)
+    {
+        if (Stack[i].MatchCoords(coords, subTileIndex))
+            break;
+    }
```
tupaschoal, line 402:
```suggestion
    auto matchingStack = std::find_if(std::begin(Stack), std::end(Stack), [&](const auto& entry){
        return entry.MatchCords(coords, subTileIndex);
    });
    auto i = matchingStack != std::end(Stack) ? std::distance(Stack.begin(), matchingStack) : 4;
```

I'm not convinced it's clearer, but I had already typed it anyway

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,223 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+
+    for (uint8_t j = std::min((uint8_t)(Stack.size() - 1), i); j > 0; --j)
```
tupaschoal, line 404:
```suggestion
    for (uint8_t j = std::min(Stack.size() - 1, i); j > 0; --j)
```
I think you can let it implicitly cast safely

src/openrct2/ride/gentle/Maze.cpp
```diff
@@ -181,3 +201,223 @@ TRACK_PAINT_FUNCTION get_track_paint_function_maze(int32_t trackType, int32_t di
...
+        item.Reset();
+}
+
+void MazePathfindingHistory::Free()
+{
+    for (auto& item : Stack)
+        item.Free();
+}
+
+std::pair<MazePathfindingEntry&, uint8_t> MazePathfindingHistory::MeetIntersection(
```
tupaschoal, line 393: If this doesn't change the entry, it should probably return a const ref

src/openrct2/ride/gentle/Maze.cpp
Gymnasiast: I think the `for` loop is a lot easier to read.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -2241,20 +2242,41 @@ void window_guest_debug_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+        screenCoords.y += LIST_ROW_HEIGHT;
 
-    screenCoords.x += 10;
-    for (auto& point : peep->PathfindHistory)
+        screenCoords.x += 10;
+        for (auto& maze_pathfind_item : peep->MazePathfindHistory.GetHistory())
+        {
+            auto [coords, subTileIndex] = maze_pathfind_item.GetCoords();
+            uint16_t args[] = {
+                (uint16_t)(coords.x >> 5),      (uint16_t)(coords.y >> 5),       subTileIndex,
```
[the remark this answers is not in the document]
Gymnasiast: That PR has now been merged.

ZehMatt:
@philgekni are you still working on this?

philgekni:
Hi @ZehMatt, I have unfortunately not the time currently to work on this task due to my work. I think that I'll be able to finish in March.

RobertWHurst:
So looking forward to this when you have time again. It's fantastic!

src/openrct2/peep/Guest.cpp
philgekni: It doesn't look. I was tempted to get `nearEntranceTile` via `ToCoordsXY()` but I need the direction even if I don't care about the Z.

src/openrct2/peep/Guest.cpp
philgekni: I think so since I store the directions in `openEdgesFlags` with a shift. If needed I could rewrite this loop to use `ALL_DIRECTIONS`.

src/openrct2/ride/gentle/Maze.cpp
philgekni: I don't think that works as `item.Reset()` is non-const method.

src/openrct2/ride/gentle/Maze.cpp
philgekni: Idem with `item.Free()`.

src/openrct2/ride/gentle/Maze.cpp
philgekni: In the Guest.c `UpdateRideMazePathfinding()` function, I modify the state of the returned entry via the method `MarkVisited`.

src/openrct2/ride/gentle/Maze.cpp
philgekni: I agree with @Gymnasiast.

src/openrct2/ride/gentle/Maze.cpp
philgekni: This is also my opinion but not msvc one.

src/openrct2/ride/gentle/Maze.cpp
philgekni: This is also my opinion but not msvc one.

philgekni:
Thanks for the waiting, I rebased the branch and addressed your review comments.
I also fixed an error that I found in the pathfinding algorithm (b93363c).
Let me know if there is any problem, I should be more available now than by the past.

src/openrct2/ride/gentle/Maze.cpp
ZehMatt: The purpose of std::find_if is to make it clear that this statement serves a single purpose and that is to find something, this also prevents users from adding up more code into the loop where it doesn't belong. std::find_if should be the preferred way in the C++ environment.

philgekni:
I continuing to work on the PR. I checked after record desynchronization.
I noticed that this kind of appears when a guest jumps into a maze. It is not at the frame where the function `Guest::UpdateRideMazePathfinding()` set `Action = PeepActionType::Jump;` but in the next frames during the animation.
I didn't yet narrow down the problem.

Dubaifreak:
The plugin can be useful, thanks for the download link!

karst:
Is this still planned on being finished?

rik-smeets:
@philgekni Do you still intend to work on this? If not, I will close this PR.

AaronVanGeffen:
The 'new' save format would allow for the maze pathfinding structs to be serialised into park saves, which was the original holdup. I hope this can be finished indeed.

philgekni:
Sorry I don't have anymore the time to work on this feature. Please feel free to close it or to take the ownership.

rik-smeets:
Thanks for responding, @philgekni.

I do agree it would be a neat feature, @AaronVanGeffen. Anyone wants to pick this one up? I don't want to keep this in limbo forever. Unfortunately, my limited C++ skillset won't allow me to do this one, so hopefully someone else wants to continue work on this.

tupaschoal:
I opened https://github.com/OpenRCT2/OpenRCT2/discussions/18599 to track this feature improvement and will close this PR for the time being, until someone can pick it up.

PR #10422, opened 2019-12-22
Refactor out gCommandExpenditureType

ZehMatt:
Another global be gone! Had to slightly re-arrange includes to avoid cycles.

### DIFF
--- a/src/openrct2/management/Finance.h
+++ b/src/openrct2/management/Finance.h
@@ -10,14 +10,11 @@
 #pragma once
 
 #include "../common.h"
-#include "../peep/Staff.h"
 #include "Research.h"
 
-using rct_expenditure_type = int32_t;
-
-enum
+enum rct_expenditure_type : int32_t
 {
-    RCT_EXPENDITURE_TYPE_RIDE_CONSTRUCTION,
+    RCT_EXPENDITURE_TYPE_RIDE_CONSTRUCTION = 0,
     RCT_EXPENDITURE_TYPE_RIDE_RUNNING_COSTS,
     RCT_EXPENDITURE_TYPE_LAND_PURCHASE,
     RCT_EXPENDITURE_TYPE_LANDSCAPING,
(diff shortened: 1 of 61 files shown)


src/openrct2/management/Finance.h
```diff
@@ -10,14 +10,11 @@
 #pragma once
 
 #include "../common.h"
-#include "../peep/Staff.h"
 #include "Research.h"
 
-using rct_expenditure_type = int32_t;
-
-enum
+enum rct_expenditure_type : int32_t
```
duncanspumpkin: can we enum class this?

ZehMatt: That will have about 600+ edits then, I can still do it tho.

duncanspumpkin: I think it makes sense to do it for this because its discrete values that dont (normally) do any maths on the underlying type. If it looks to be a huge load of changes then just leave it for now.

ZehMatt: Its not too bad I think

AaronVanGeffen:
Given that the enum already inherits from `int32_t`, are these `static_cast<int32_t>` really required? They're not too aesthetically pleasing…

ZehMatt:
> Given that the enum already inherits from `int32_t`, are these `static_cast<int32_t>` really required? They're not too aesthetically pleasing…

enum class has no implicit conversation to integers, so yeah.

AaronVanGeffen:
Perhaps we add a utility like:

```cpp
#include <type_traits>

template <typename E>
constexpr auto AsInt(E e) noexcept
{
    return static_cast<std::underlying_type_t<E>>(e);
}
```

This would be considerably shorter and arguably easier to read.

[(via)](https://stackoverflow.com/questions/8357240/how-to-automatically-convert-strongly-typed-enum-into-int#answer-33083231)

ZehMatt:
I agree, however might be a bit out of scope already for this PR. There are other usages like this all across the project and should be done in another PR perhaps?

ZehMatt:
I also added the clang-format version output on our script, was slightly annoying to not know which specific version is used on our CI.

AaronVanGeffen:
> I agree, however might be a bit out of scope already for this PR. There are other usages like this all across the project and should be done in another PR perhaps?

Hmm, I agree it may be a little out of scope for this PR, yes. Let's not forget, either, though.

ZehMatt:
I'll look into it once this is merged, the helper function should also only accept enums so more template magic is required.

ZehMatt:
Any objection on the new commits?

PR #10428, opened 2019-12-22
Use integer math for guest initial happiness calculation

ZehMatt:
This ensures that every platform and compiler will yield the same result when it comes to calculating the initial happiness of guests where for floating point math the rounding can become an issue. This also checks compile time that the up scaled PI is correct.

### DIFF
--- a/src/openrct2/world/Park.cpp
+++ b/src/openrct2/world/Park.cpp
@@ -657,7 +657,10 @@ uint8_t Park::CalculateGuestInitialHappiness(uint8_t percentage)
     // This sequence can be defined as PI*(9+n)/2 (the value is floored)
     for (uint8_t n = 1; n < 55; n++)
     {
-        if ((3.14159 * (9 + n)) / 2 >= percentage)
+        // Avoid floating point math by rescaling PI up.
+        constexpr int32_t SCALE = 100000;
+        constexpr int32_t PI_SCALED = 314159; // PI * SCALE;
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Park.cpp
```diff
@@ -657,7 +657,11 @@ uint8_t Park::CalculateGuestInitialHappiness(uint8_t percentage)
     // This sequence can be defined as PI*(9+n)/2 (the value is floored)
     for (uint8_t n = 1; n < 55; n++)
     {
-        if ((3.14159 * (9 + n)) / 2 >= percentage)
+        // Avoid floating point math by rescaling PI up.
+        constexpr int32_t SCALE = 100000;
+        constexpr int32_t PI_SCALED = static_cast<int32_t>(3.141590 * SCALE);
+        static_assert(PI_SCALED == 314159);
+        if (((PI_SCALED * (9 + n)) / SCALE) / 2 >= percentage)
```
IntelOrca: Why not just do `constexpr int32_t PI_SCALED = 314159`?

ZehMatt: To clarify that this is PI but I guess I could just comment it.

IntelOrca: Hmm, that would only make sense if you used the `M_PI` define.

ZehMatt: Changed it to the computed constant and added a comment to how it came to be.

PR #10430, opened 2019-12-23
Add Keyboard Shortcuts to the Tile Inspector

bitman2049:
Add Keyboard Shortcuts to the Tile Inspector

The Tile Inspector's common functions now have unmapped keyboard shortcuts. The rotate button (default Z) now rotates element in the Tile Inspector.


janisozaur:
Please ensure your commit messages adhere to https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages and fix your history so there are no unnecessary files in there

data/language/en-GB.txt
```diff
@@ -3654,6 +3654,18 @@ STR_6331    :Create Ducks
...
+STR_6337	:(Tile Inspector) Paste element
+STR_6338	:(Tile Inspector) Delete element
+STR_6339	:(Tile Inspector) Move element up
+STR_6340	:(Tile Inspector) Move element down
+STR_6341	:(Tile Inspector) Increase X coordinate
+STR_6342	:(Tile Inspector) Decrease X coordinate
+STR_6343	:(Tile Inspector) Increase Y coordinate
+STR_6344	:(Tile Inspector) Decrease Y coordinate
+STR_6345	:(Tile Inspector) Increase element height
+STR_6346	:(Tile Inspector) Decrease element height
```
AaronVanGeffen: Prefixing these is probably a good idea, but I'm not sure about the parentheses. Personally, I think colons would be better. For instance:

```suggestion
STR_6346	:Tile Inspector: Decrease element height
```

Perhaps other team members disagree, though.

AaronVanGeffen:
Thanks for your PR! I'm sure these changes will be appreciated by many players. It also reminds me we really need to change our keyboard shortcut file format… It's current not possible to change the shortcut order without losing assigned key bindings.

Also, a quick heads-up: it looks like the commit author details do not match your GitHub profile's. This has no effect on the submitted code, but may make it more difficult for to find (or attribute) them at a later point.

data/language/en-GB.txt
ZehMatt: I agree

bitman2049: Just committed the change.

src/openrct2-ui/input/KeyboardShortcut.cpp
```diff
@@ -814,6 +823,201 @@ static void shortcut_scale_down()
     context_update_cursor_scale();
 }
 
+// Tile inspector shortcuts
+static void shortcut_insert_corrupt_element()
+{
+    rct_window* w = window_find_by_class(WC_TILE_INSPECTOR);
+    if (w != nullptr && !widget_is_disabled(w, WC_TILE_INSPECTOR__WIDX_BUTTON_CORRUPT)
```
AaronVanGeffen: @Broxzier The current PR exposes all these widget ids through `Window.h`. Is there a better way of accessing the status of these buttons?

bitman2049: So I based it on how some of the other window-specific widgets are accessed, such as staff tools or the scenery eyedropper. If you look above my changes in Window.h, you'll see other widgets accessed in a similar way.

PR #10431, opened 2019-12-23
Reduce LocationXY usage in vehicle.cpp.

duncanspumpkin:
Note changed logic on boat hire (Now uses `ToTileCentre()` before didn't do the `& ~0x1F`)
and chairlift (Now actually checks the z coordinate of the bullwheel when going backwards {not sure if that is even possible}).
Both likely minor bugs


duncanspumpkin:
This needs a network bump

tupaschoal:
Helps #10267

PR #10413, opened 2019-12-21
Make Map::map_get_first_element_at() use CoordsXY

tupaschoal:
This will need rebasing once #10409 is pushed, but is already rebased on all my other PRs.

It also showcases lots of places that receive `{x ,y}` on function name that can be refactored to use `CoordsXY`

PR #10432, opened 2019-12-23
Make ClearAction::ClearSceneryFromTile() receive a CoordsXY object

tupaschoal:
To address https://github.com/OpenRCT2/OpenRCT2/pull/10413#discussion_r360790284

### DIFF
--- a/src/openrct2/actions/ClearAction.hpp
+++ b/src/openrct2/actions/ClearAction.hpp
@@ -104,7 +104,7 @@ private:
             {
                 if (MapCanClearAt(x, y))
                 {
-                    auto cost = ClearSceneryFromTile(x / 32, y / 32, executing);
+                    auto cost = ClearSceneryFromTile({ x, y }, executing);
                     if (cost != MONEY32_UNDEFINED)
                     {
                         noValidTiles = false;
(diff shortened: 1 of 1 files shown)


src/openrct2/actions/ClearAction.hpp
```diff
@@ -171,7 +172,7 @@ DEFINE_GAME_ACTION(ClearAction, GAME_COMMAND_CLEAR_SCENERY, GameActionResult)
                         if (_itemsToClear & CLEARABLE_ITEMS::SCENERY_SMALL)
                         {
                             auto removeSceneryAction = SmallSceneryRemoveAction(
-                                { x * 32, y * 32, tileElement->base_height * 8 },
+                                { tilePos.x, tilePos.y, tileElement->base_height * 8 },
```
Gymnasiast: If I'm not mistaken, you can create a CoordsXYZ(D) from a CoordsXY. If that is indeed the case, please modify the code accordingly.

tupaschoal: They do, and I actually used this on the other PRs today haha
Dealt with them, thanks for the tip :)

PR #10433, opened 2019-12-23
Make Map::check_max_allowable_land_rights_for_tile() receive CoordsXYZ

tupaschoal:
I thought it would be better to consistently wrap it into a `CoordsXYZ` than to send a `CoordsXY` and a `z`, let me know what you think


Gymnasiast:
> I thought it would be better to consistently wrap it into a CoordsXYZ than to send a CoordsXY and a z, let me know what you think

Absolutely agree.

PR #10435, opened 2019-12-23
Make Map::map_get_track_element_at*() use CoordsXYZ*

tupaschoal:
By themselves, each function would be a very small PR and with lots of merge between each other, so I decided to do a PR grouping them all, but each on their own commit, let me know if that's too bad.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -4692,15 +4692,13 @@ static void vehicle_unset_update_flag_b1(rct_vehicle* head)
 static void ride_create_vehicles_find_first_block(Ride* ride, CoordsXYE* outXYElement)
 {
     rct_vehicle* vehicle = GET_VEHICLE(ride->vehicles[0]);
-    int32_t firstX = vehicle->track_x;
-    int32_t firstY = vehicle->track_y;
-    int32_t firstZ = vehicle->track_z;
-    auto firstElement = map_get_track_element_at(firstX, firstY, firstZ / 8);
+    auto firstPos = CoordsXYZ{ vehicle->track_x, vehicle->track_y, vehicle->track_z };
+    auto firstElement = map_get_track_element_at(firstPos);
 
     assert(firstElement != nullptr);
 
-    int32_t x = firstX;
-    int32_t y = firstY;
+    int32_t x = firstPos.x;
+    int32_t y = firstPos.y;
     auto trackElement = firstElement;
     track_begin_end trackBeginEnd;
     while (track_block_get_previous(x, y, reinterpret_cast<TileElement*>(trackElement), &trackBeginEnd))
@@ -4708,7 +4706,7 @@ static void ride_create_vehicles_find_first_block(Ride* ride, CoordsXYE* outXYEl
         x = trackBeginEnd.end_x;
         y = trackBeginEnd.end_y;
         trackElement = trackBeginEnd.begin_element->AsTrack();
-        if (x == firstX && y == firstY && trackElement == firstElement)
+        if (x == firstPos.x && y == firstPos.y && trackElement == firstElement)
         {
             break;
         }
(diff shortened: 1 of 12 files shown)


test/tests/TileElements.cpp
```diff
@@ -82,7 +82,7 @@ TEST_F(TileElementWantsFootpathConnection, SlopedPath)
 TEST_F(TileElementWantsFootpathConnection, Stall)
 {
     // Stalls usually have one path direction flag, but can have multiple (info kiosk for example)
-    const TrackElement* const stallElement = map_get_track_element_at(19 << 5, 15 << 5, 14);
+    const TrackElement* const stallElement = map_get_track_element_at({ 19 << 5, 15 << 5, 14 << 3 });
```
Gymnasiast: This is becoming more ugly with every change ;). Can you put the data into a TileCoordsXYZ and then convert that to CoordsXYZ?

src/openrct2/world/Map.cpp
```diff
@@ -2131,16 +2131,17 @@ void map_clear_all_elements()
...
+    TileElement* tileElement = map_get_first_element_at(trackPos);
     if (tileElement == nullptr)
         return nullptr;
+    auto trackTilePos = TileCoordsXYZ{ trackPos };
     do
     {
         if (tileElement->GetType() != TILE_ELEMENT_TYPE_TRACK)
             continue;
-        if (tileElement->base_height != z)
+        if (tileElement->base_height != trackTilePos.z)
```
Gymnasiast: This could have been so much cleaner if https://github.com/OpenRCT2/OpenRCT2/pull/10414 had already been merged. Ah well.

test/tests/TileElements.cpp
tupaschoal: Done

src/openrct2/ride/Ride.cpp
```diff
@@ -4701,23 +4701,21 @@ static void vehicle_unset_update_flag_b1(rct_vehicle* head)
 static void ride_create_vehicles_find_first_block(Ride* ride, CoordsXYE* outXYElement)
 {
     rct_vehicle* vehicle = GET_VEHICLE(ride->vehicles[0]);
-    int32_t firstX = vehicle->track_x;
-    int32_t firstY = vehicle->track_y;
-    int32_t firstZ = vehicle->track_z;
-    auto firstElement = map_get_track_element_at(firstX, firstY, firstZ / 8);
+    auto firstPos = CoordsXYZ{ vehicle->track_x, vehicle->track_y, vehicle->track_z };
```
duncanspumpkin: `curTrackPos`

src/openrct2/ride/Ride.cpp
```diff
@@ -4701,23 +4701,21 @@ static void vehicle_unset_update_flag_b1(rct_vehicle* head)
...
+    auto firstElement = map_get_track_element_at(firstPos);
```
duncanspumpkin: `curTrackElement`

src/openrct2/ride/Ride.cpp
```diff
@@ -4701,23 +4701,21 @@ static void vehicle_unset_update_flag_b1(rct_vehicle* head)
 static void ride_create_vehicles_find_first_block(Ride* ride, CoordsXYE* outXYElement)
 {
     rct_vehicle* vehicle = GET_VEHICLE(ride->vehicles[0]);
```
duncanspumpkin: This functions starts at the current track element that the vehicle is on and walks backwards up the track until it finds the first block brake or station or lift hill. I suggest we rename the `first` variables to `curTrack` as I feel it is more descriptive.

tupaschoal: Done!

Gymnasiast:
I'll merge this once it's rebased.

tupaschoal:
Rebased and recompiled.

PR #10437, opened 2019-12-23
Make Map::map_invalidate*() use CoordsXY

tupaschoal:


src/openrct2/world/Map.cpp
```diff
@@ -1983,41 +1983,41 @@ static void map_invalidate_tile_under_zoom(int32_t x, int32_t y, int32_t z0, int
  *
  *  rct2: 0x006EC847
  */
-void map_invalidate_tile(int32_t x, int32_t y, int32_t z0, int32_t z1)
+void map_invalidate_tile(const CoordsXY& tilePos, int32_t z0, int32_t z1)
```
Gymnasiast: This function, and its siblings, receives CoordsXY, coupled with TileCoords z values. I want to get rid of this mixed kind of functions.

One thing, of course, is that we don't have a fitting coords type for functions that need both a low and high Z. I propose creating CoordsXYZZ for this purpose.

@duncanspumpkin

duncanspumpkin: I'm not sure cordsxyzz makes that much sense. It's not exactly used that often either.

Gymnasiast: I think there are at least 20 functions that have this pattern. I really think such a struct should be created.

I'm not wedded to the CoordsXYZZ name, though. It just seemed a straightforward name to me.

tupaschoal: Maybe `CoordsXYRangedZ` or something alike

tupaschoal:
I've rebased so it can be mergeable once again. I can add the new `struct` for `CoordsXYZZ` in this PR or another, whichever you prefer.

src/openrct2/world/Map.cpp
duncanspumpkin: @Gymnasiast thoughts on `CoordsXYRangedZ`? I think that is a bit better. I'm not overly fussed on what is chosen. We should perhaps make a `RangedZ` struct though as there are some instances I can think of (map is clear at functions) where you pass in a base and clearance.

Gymnasiast: `CoordsXYRangedZ` is fine.

As for the `RangedZ` struct: are there places where one needs to pass _just_ a ranged Z and no X/Y?

duncanspumpkin: Yes the map is clear functions use it to work out collisions. There might be a few others

Gymnasiast: Do you mean `map_can_construct_with_clear_at()`? That one also takes x and y.

I haven't been able to find a function that _just_ takes a ranged z (and no x/y).

duncanspumpkin: Oh yeah I'm a twit. Yes of course it does have x and y as well. Yes just create a new struct called `CoordsXYRangedZ` and this will cover all cases I think.

tupaschoal: Implemented and replaced all usages with `CoordsXYRangedZ`

Gymnasiast: Good!

I think it's best done in a separate PR. I will review this one and merge it if nothing else pops up.

src/openrct2/actions/BannerSetColourAction.hpp
```diff
@@ -103,7 +103,7 @@ DEFINE_GAME_ACTION(BannerSetColourAction, GAME_COMMAND_SET_BANNER_COLOUR, GameAc
 
             auto banner = GetBanner(index);
             banner->colour = _primaryColour;
-            map_invalidate_tile_zoom1(_loc.x, _loc.y, _loc.z, _loc.z + 32);
+            map_invalidate_tile_zoom1({ _loc, _loc.z, _loc.z + 32 });
```
Gymnasiast: This looks odd. In BannerRemoveAction, the z value is divided by 8, but here, it isn't.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7952,7 +7952,8 @@ static bool vehicle_update_track_motion_forwards_get_new_track(
                         SoundId::BlockBrakeRelease, { vehicle->track_x, vehicle->track_y, vehicle->track_z });
                 }
             }
-            map_invalidate_element(vehicle->track_x, vehicle->track_z, tileElement);
+            // This is very weird because it uses Z
```
Gymnasiast: This is a very weird comment :P 

Seriously though: what is weird about it using Z?

src/openrct2/world/Location.hpp
```diff
@@ -162,6 +162,27 @@ struct CoordsXY
...
 
+struct CoordsXYRangedZ : public CoordsXY
+{
+    int32_t zBegin = 0;
+    int32_t zEnd = 0;
+
+    CoordsXYRangedZ() = default;
+    constexpr CoordsXYRangedZ(int32_t _x, int32_t _y, int32_t _zBegin, int32_t _zEnd)
+        : CoordsXY(_x, _y)
+        , zBegin(_zBegin)
```
Gymnasiast: zLow

src/openrct2/ride/Vehicle.cpp
tupaschoal: I meant that all other calls use `x,y` but this uses `x,z`

Gymnasiast: Could it be a refactoring typo? IIRC, Duncan recently refactored this, it might have been using y previously.

tupaschoal: It seems like it used to be `x, y`, didn't go as far as finding out who changed, but I fixed it.

tupaschoal:
Rebased, will wait for the decision on `zLow` or `baseZ`

Gymnasiast:
Could you also look into the other line note (about the parameters to `map_invalidate_tile_zoom1()`)?

src/openrct2/actions/BannerSetColourAction.hpp
tupaschoal: I think the weirdness is in `BannerRemoveAction.hpp` instead, because this function receives `CoordsXYZZ` and all other places actually seem to send Z coords that are multiplied by 8

Gymnasiast: Looks like it, yeah. Can you correct the call in BannerRemoveAction.hpp, then?

tupaschoal: Done

src/openrct2/world/Location.hpp
tupaschoal: I've gone with `baseZ` and `clearanceZ`, given the recent PR you did, if people think otherwise, I can change, it's not a big one.

src/openrct2/peep/Guest.cpp
```diff
@@ -6320,7 +6320,7 @@ static void peep_update_walking_break_scenery(Peep* peep)
 
     tileElement->AsPath()->SetIsBroken(true);
 
-    map_invalidate_tile_zoom1(peep->next_x, peep->next_y, (tileElement->GetBaseZ()) + 32, tileElement->GetBaseZ());
+    map_invalidate_tile_zoom1({ peep->next_x, peep->next_y, (tileElement->GetBaseZ()) + 32, tileElement->GetBaseZ() });
```
duncanspumpkin: remove the additional brackets around GetBaseZ()

duncanspumpkin: This is also very odd why is the base higher than the clearance.

duncanspumpkin: Just checked the assembly this is the wrong way round. should be `tileElement->GetBaseZ(), tileElement->GetBaseZ() + 32`

tupaschoal: Nice catch, just fixed it!

tupaschoal:
> For a future refactoring job we could do with using `map_invalidate_element` more often. Its a helper function that does the GetBaseZ and GetClearanceZ calls that a number of the changes in this PR could take advantage of.

It would use a different zoom though that would always take to invalidation, wouldn't it?

src/openrct2/ride/Ride.cpp
```diff
@@ -1381,7 +1381,7 @@ void ride_construction_invalidate_current_track()
         case RIDE_CONSTRUCTION_STATE_MAZE_FILL:
             if (_currentTrackSelectionFlags & TRACK_SELECTION_FLAG_ARROW)
             {
-                map_invalidate_tile_full(_currentTrackBegin.x & 0xFFE0, _currentTrackBegin.y & 0xFFE0);
+                map_invalidate_tile_full({ _currentTrackBegin.x & 0xFFE0, _currentTrackBegin.y & 0xFFE0 });
```
Gymnasiast:
```suggestion
                map_invalidate_tile_full( _currentTrackBegin.x.ToTileStart());
```

tupaschoal: Done.

src/openrct2/peep/Guest.cpp
Gymnasiast: Nice catch indeed - would this fix reduce the number of Z fighting issues?

tupaschoal:
Rebased to resolve conflicts pushed today. I guess this is good for merging?

PR #10438, opened 2019-12-24
Refactor ride remove LocationXY8

duncanspumpkin:
Couple little changes to remove the last of the LocationXY infection from the ride struct. 
I found no instances or way in which the bullwheel could have a null value or the boat hire return position so have not provided conversions for such states in the import/export code.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -2126,14 +2126,14 @@ void Ride::UpdateChairlift()
     if (old_chairlift_bullwheel_rotation == speed / 8)
         return;
 
-    x = chairlift_bullwheel_location[0].x * 32;
-    y = chairlift_bullwheel_location[0].y * 32;
-    z = chairlift_bullwheel_z[0] * 8;
+    x = ChairliftBullwheelLocation[0].x * 32;
+    y = ChairliftBullwheelLocation[0].y * 32;
+    z = ChairliftBullwheelLocation[0].z * 8;
     map_invalidate_tile_zoom1(x, y, z, z + (4 * 8));
 
-    x = chairlift_bullwheel_location[1].x * 32;
-    y = chairlift_bullwheel_location[1].y * 32;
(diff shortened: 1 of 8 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -2126,14 +2126,14 @@ void Ride::UpdateChairlift()
     if (old_chairlift_bullwheel_rotation == speed / 8)
         return;
 
-    x = chairlift_bullwheel_location[0].x * 32;
-    y = chairlift_bullwheel_location[0].y * 32;
-    z = chairlift_bullwheel_z[0] * 8;
+    x = ChairliftBullwheelLocation[0].x * 32;
```
Gymnasiast: Use `ToCoordsXYZ`.

src/openrct2/ride/Ride.cpp
```diff
@@ -4214,9 +4214,9 @@ static bool ride_check_start_and_end_is_station(CoordsXYE* input)
     {
         return false;
     }
-    ride->chairlift_bullwheel_location[1].x = trackFront.x >> 5;
-    ride->chairlift_bullwheel_location[1].y = trackFront.y >> 5;
-    ride->chairlift_bullwheel_z[1] = trackFront.element->base_height;
+    ride->ChairliftBullwheelLocation[1].x = trackFront.x >> 5;
```
Gymnasiast: Create a CoordsXYZ from trackFront.x, trackFront.y and trackFront.element->GetBaseZ, then convert that to TileCoordsXYZ.

Gymnasiast:
Looks good in general, just two line notes.

src/openrct2/ride/Ride.cpp
duncanspumpkin: I had kind of left this out as it would cause collision with the map invalidate pr but I've made this change now anyway.

tupaschoal:
Helps #10267

PR #10440, opened 2019-12-24
Coords for map_get_tile*

tupaschoal:


src/openrct2/world/Map.cpp
```diff
@@ -2046,17 +2046,17 @@ void map_invalidate_region(const CoordsXY& mins, const CoordsXY& maxs)
     }
 }
 
-int32_t map_get_tile_side(int32_t mapX, int32_t mapY)
+int32_t map_get_tile_side(const CoordsXY& mapPos)
 {
-    int32_t subMapX = mapX & (32 - 1);
-    int32_t subMapY = mapY & (32 - 1);
+    int32_t subMapX = mapPos.x & (32 - 1);
```
Gymnasiast: Change subMap{X,Y} into a CoordsXY and use ToTileStart().

tupaschoal: It doesn't look the same as `ToTileStart()`. `floor2` does `x & ~(32 - 1)` effectively rounding down to `32` or `0`. This one doesn't have the not, it is `x & (32 - 1)`, so it is kind of like doing `mapPos.x % 32`
Of course you can tell me that this is a bug and should have behaved as you said from the very beginning

Gymnasiast: Looks like you're right.

PR #10441, opened 2019-12-24
Make Map::sign_set_colour() use CoordsXYZD

tupaschoal:


src/openrct2/actions/SignSetStyleAction.hpp
```diff
@@ -99,7 +99,7 @@ DEFINE_GAME_ACTION(SignSetStyleAction, GAME_COMMAND_SET_SIGN_STYLE, GameActionRe
         {
             TileElement* tileElement = banner_get_tile_element((BannerIndex)_bannerIndex);
             if (!sign_set_colour(
-                    coords.x, coords.y, tileElement->base_height, tileElement->GetDirection(),
+                    { coords, tileElement->base_height << 3, tileElement->GetDirection() },
```
duncanspumpkin: GetBaseZ()

src/openrct2/world/Map.cpp
```diff
@@ -1900,15 +1900,13 @@ std::optional<CoordsXYZ> map_large_scenery_get_origin(
  *
  *  rct2: 0x006B9B05
  */
-bool sign_set_colour(
-    int32_t x, int32_t y, int32_t z, int32_t direction, int32_t sequence, uint8_t mainColour, uint8_t textColour)
+bool sign_set_colour(const CoordsXYZD& signPos, int32_t sequence, uint8_t mainColour, uint8_t textColour)
```
duncanspumpkin: Can you rename this to add that this is for `Large` scenery signs only.

tupaschoal:
Done!

PR #10442, opened 2019-12-24
Use CoordsXYZ for gMapSelectArrowPosition

tupaschoal:
This turns `gMapSelectArrowPosition` and `gMapSelectArrowDirection` into a single, non-LocationXYZ variable. They were always assigned to/from together, so it also makes the code more consistent.

Helps #10267

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2611,13 +2611,11 @@ void sub_6C94D8()
             z = _currentTrackBegin.z;
             direction = _currentTrackPieceDirection;
             type = _currentTrackPieceType;
-            gMapSelectArrowPosition.x = x;
-            gMapSelectArrowPosition.y = y;
-            gMapSelectArrowPosition.z = z;
             if (direction >= 4)
                 direction += 4;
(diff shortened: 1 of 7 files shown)


tupaschoal:
Argh, that's so sad

tupaschoal:
I've reworked it to stop using `CoordsXYZD` as @duncanspumpkin suggested it and just made `gMapSelectArrowPosition` use `CoordsXYZ`

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3709,10 +3703,8 @@ void ride_construction_toolupdate_entrance_exit(ScreenCoordsXY screenCoords)
     gMapSelectType = MAP_SELECT_TYPE_FULL;
     gMapSelectPositionA = entranceOrExitCoords;
     gMapSelectPositionB = entranceOrExitCoords;
+    gMapSelectArrowPosition = CoordsXYZ{ entranceOrExitCoords.x, entranceOrExitCoords.y, entranceOrExitCoords.z };
```
duncanspumpkin: Can't this just assign from a CoordsXYZD to a CoordsXYZ?

tupaschoal: It can and it now is.

PR #10444, opened 2019-12-24
Remove LocationXY usage on ride and peep

tupaschoal:
Helps #10267

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1228,11 +1228,8 @@ void Guest::UpdateSitting()
         if (!(pathingResult & PATHING_DESTINATION_REACHED))
             return;
 
-        LocationXYZ16 loc = {
-            (int16_t)((x & 0xFFE0) + BenchUseOffsets[var_37 & 0x7].x),
-            (int16_t)((y & 0xFFE0) + BenchUseOffsets[var_37 & 0x7].y),
-            z,
-        };
+        auto loc = CoordsXYZ{ x, y, z }.ToTileStart()
+            + CoordsXYZ{ BenchUseOffsets[var_37 & 0x7].x, BenchUseOffsets[var_37 & 0x7].y, 0 };
 
         MoveTo(loc.x, loc.y, loc.z);
 
@@ -3537,7 +3534,7 @@ void Guest::UpdateRideAtEntrance()
 }
 
 /** rct2: 0x00981FD4, 0x00981FD6 */
-static constexpr const LocationXY16 _MazeEntranceStart[] = {
+static constexpr const CoordsXY _MazeEntranceStart[] = {
     { 8, 8 },
     { 8, 24 },
     { 24, 24 },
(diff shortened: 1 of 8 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -3574,7 +3574,7 @@ static void vehicle_update_collision_setup(rct_vehicle* vehicle)
 }
 
 /** rct2: 0x009A3AC4, 0x009A3AC6 */
-static constexpr const LocationXY16 stru_9A3AC4[] = {
+static constexpr const CoordsXY stru_9A3AC4[] = {
```
duncanspumpkin: These are not coords. I'm not really sure what they represent yet but they aren't coords

tupaschoal: It looks like they are a collection of constant used to calculate coords for the vehicle crash

src/openrct2/peep/Guest.cpp
```diff
@@ -1227,9 +1227,9 @@ void Guest::UpdateSitting()
         if (!(pathingResult & PATHING_DESTINATION_REACHED))
             return;
 
-        LocationXYZ16 loc = {
-            (int16_t)((x & 0xFFE0) + BenchUseOffsets[var_37 & 0x7].x),
-            (int16_t)((y & 0xFFE0) + BenchUseOffsets[var_37 & 0x7].y),
+        CoordsXYZ loc = {
+            (x & 0xFFE0) + BenchUseOffsets[var_37 & 0x7].x,
```
Gymnasiast: Use ToTileStart()

src/openrct2/peep/Guest.cpp
```diff
@@ -3633,7 +3633,7 @@ static void peep_update_ride_leave_entrance_waypoints(Peep* peep, Ride* ride)
     Guard::Assert(!entranceLocation.isNull());
     uint8_t direction_entrance = entranceLocation.direction;
 
-    LocationXY16 waypoint;
+    CoordsXY waypoint;
     waypoint.x = ride->stations[peep->current_ride_station].Start.x * 32 + 16;
```
Gymnasiast: Use ToTileCentre()

src/openrct2/peep/Staff.cpp
```diff
@@ -445,8 +445,7 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
         return 0xFF;
     }
 
-    LocationXY16 litterTile = { static_cast<int16_t>(nearestLitter->x & 0xFFE0),
-                                static_cast<int16_t>(nearestLitter->y & 0xFFE0) };
+    CoordsXY litterTile = { nearestLitter->x & 0xFFE0, nearestLitter->y & 0xFFE0 };
```
Gymnasiast: Use ToTileStart()

src/openrct2/world/Footpath.cpp
```diff
@@ -70,7 +70,7 @@ const LocationXY16 BenchUseOffsets[8] = {
 };
 
 /** rct2: 0x00981D6C, 0x00981D6E */
-const LocationXY16 word_981D6C[4] = { { -1, 0 }, { 0, 1 }, { 1, 0 }, { 0, -1 } };
+const CoordsXY word_981D6C[4] = { { -1, 0 }, { 0, 1 }, { 1, 0 }, { 0, -1 } };
```
Gymnasiast: These look like TileCoords offsets, not Coords.

src/openrct2/peep/Guest.cpp
tupaschoal: I've done it, but note that it only sums the 16, but doesn't `floor`s the coords first.

Gymnasiast: It doesn't need to - it multiplies by 32 first, so the value is a multiple of 32 already - already floored.

src/openrct2/ride/Vehicle.cpp
Gymnasiast: The code using them is very hard to read, but it does look like they are indeed base values for vehicle crash coords.

Not sure if we should save them in a CoordsXY value - perhaps it's better to revert this change, for now at least.

tupaschoal: Ok, since it was the only change in that commit, I dropped it altogether.

PR #10453, opened 2019-12-27
Create getters for station height and start

Gymnasiast:


src/openrct2/ride/Station.cpp
```diff
@@ -434,3 +432,19 @@ void ride_set_exit_location(Ride* ride, const int32_t stationIndex, const TileCo
...
+void RideStation::SetBaseZ(int32_t newZ)
+{
+    Height = newZ / 8;
+}
+
+CoordsXYZ RideStation::GetStart() const
+{
+    TileCoordsXYZ stationTileCoords{ Start.x, Start.y, Height };
+    return stationTileCoords.ToCoordsXYZ();
+}
```
tupaschoal: I'm thinking if it maybe shouldn't return the `TileCoordsXYZ` and we get to call `ToCoordsXYZ` on the oustide.

src/openrct2/ride/Ride.cpp
```diff
@@ -7081,23 +7080,16 @@ void sub_6CB945(Ride* ride)
                     if (!ride_get_exit_location(ride, stationId).isNull())
                         break;
 
-                    ride_set_exit_location(
-                        ride, stationId,
-                        { location.x / 32, location.y / 32, ride->stations[stationId].Height,
-                          (uint8_t)tileElement->GetDirection() });
+                    CoordsXYZD loc = { location, ride->stations[stationId].GetBaseZ(), (uint8_t)tileElement->GetDirection() };
```
tupaschoal: While you're at it, do `static_cast<Direction>(tileElement->GetDirection())` instead. I've been trying to get away with these C casts.

src/openrct2/ride/Ride.cpp
```diff
@@ -7081,23 +7080,16 @@ void sub_6CB945(Ride* ride)
...
                         break;
 
-                    TileCoordsXYZD entranceLocation = {
-                        location.x / 32,
-                        location.y / 32,
-                        ride->stations[stationId].Height,
-                        (uint8_t)tileElement->GetDirection(),
-                    };
-                    ride_set_entrance_location(ride, stationId, entranceLocation);
+                    CoordsXYZD loc = { location, ride->stations[stationId].GetBaseZ(), (uint8_t)tileElement->GetDirection() };
```
tupaschoal: Same as above.

src/openrct2/ride/Station.cpp
Gymnasiast: I haven't done that because we're moving everything to Coords.

src/openrct2/ride/Ride.cpp
Gymnasiast: GetDirection() already returns a `Direction`, so casting that to `Direction` is a bit pointless.

tupaschoal: Even better, we can drop the casts and be happy :)

PR #10456, opened 2019-12-27
Create definitions for some coords related stuff

Gymnasiast:


duncanspumpkin:
We should also define the (2 * 8) for sloped paths and (4 * 8) for diagonal surfaces. Maybe leave for a different PR though.

PR #10471, opened 2019-12-29
Name word_981D6C

Gymnasiast:


src/openrct2/world/Footpath.cpp
```diff
@@ -70,7 +70,7 @@ const LocationXY16 BenchUseOffsets[8] = {
 };
 
 /** rct2: 0x00981D6C, 0x00981D6E */
-const CoordsXY word_981D6C[4] = { { -1, 0 }, { 0, 1 }, { 1, 0 }, { 0, -1 } };
+const CoordsXY DirectionOffsets[4] = { { -1, 0 }, { 0, 1 }, { 1, 0 }, { 0, -1 } };
```
AaronVanGeffen: Are these ever changed? If not, these really should be `const`.

duncanspumpkin:
Technically the value and structure is identical to the tilecoordsdirectiondelta but is used exclusively with a CoordsXY.

Issue #10468, opened 2019-12-29
Ride's Entrance/Exit is not buildable above a certain height level


telk5093:
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 98110c0

I can't build the entrance/exit of a ride which is at height equals or more than 9.
If it is at height +8, I can build entrance/exit of it. But if it is at +9, I can't.

**Steps to reproduce:**
1. Build any ride platform at the height >= 9
2. Point the mouse cursor next to the ride to build the entrace or exit
3. The ghost entrance/exit is not displayed

**Screenshots / Video:**
![b](https://user-images.githubusercontent.com/10726524/71556673-bd654d00-2a7e-11ea-9980-23eca5f0b9f5.gif)
(I disabled the support limits to show you this issue, but it doesn't matter if it is enabled or not)

Gymnasiast:
@duncanspumpkin @tupaschoal

PR #10473, opened 2019-12-29
Fix #5249: No collision detection when building ride entrance at heights > 85.5m

### LINKED ISSUE #5249: Ride entrance/exit may be built in the same position over 85.5m(=57 units)
**OS:** Windows 10
**Version:** 0.0.7
**Commit/Build:** 7005d5d

It is issued with #5248 together in korean community,
ride entrance/exit could be built in the same position if their height is over 85.5m (=57 unit).

- [ ] Multiplayer?

**Steps to reproduce:**
1. Disable support limit
2. Build any ride with height as 85.5m (=57 unit)
3. Then the clearance check of their entrance/exit isn't checked correctly. They could be built in the same position

**Dump file**

**Screenshots / Video:**
image

**Save game:**

Gymnasiast:
When doing https://github.com/OpenRCT2/OpenRCT2/issues/10468 , I remembered a similar bug, so I decided to look into that one as well. As I expected, it was caused by an overflowing 8-bit signed integer.

### DIFF
--- a/src/openrct2/actions/RideEntranceExitPlaceAction.hpp
+++ b/src/openrct2/actions/RideEntranceExitPlaceAction.hpp
@@ -105,7 +105,7 @@ public:
             return MakeResult(GA_ERROR::NOT_OWNED, errorTitle);
         }
 
-        int8_t clear_z = (z / 8) + (_isExit ? 5 : 7);
+        auto clear_z = (z / 8) + (_isExit ? 5 : 7);
         auto cost = MONEY32_UNDEFINED;
         if (!map_can_construct_with_clear_at(
                 _loc.x, _loc.y, z / 8, clear_z, &map_place_non_scenery_clear_func, { 0b1111, 0 }, GetFlags(), &cost,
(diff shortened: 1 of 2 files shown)


src/openrct2/actions/RideEntranceExitPlaceAction.hpp
```diff
@@ -105,7 +105,7 @@ DEFINE_GAME_ACTION(RideEntranceExitPlaceAction, GAME_COMMAND_PLACE_RIDE_ENTRANCE
             return MakeResult(GA_ERROR::NOT_OWNED, errorTitle);
         }
 
-        int8_t clear_z = (z / 8) + (_isExit ? 5 : 7);
+        auto clear_z = (z / 8) + (_isExit ? 5 : 7);
```
duncanspumpkin: Make the same changes to L176 for this to actually work.

Gymnasiast: It _does_ work, as I have tested it...

PR #10480, opened 2019-12-30
Reduce direct usage of LOCATION_NULL

duncanspumpkin:
Ideally we want to refactor a lot of these out to use std::optional as much as possible but for now move to using the standard interfaces.

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2174,7 +2174,7 @@ static std::optional<CoordsXY> ride_get_place_position_from_screen_position(Scre
     if (!_trackPlaceCtrlState)
     {
         mapCoords = sub_68A15E(screenCoords);
-        if (mapCoords.x == LOCATION_NULL)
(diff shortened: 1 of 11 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3578,6 +3578,7 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
                 break;
 
             _currentTrackBegin.z -= 8;
+            // FIXME: Unexpected & LOCATION_NULL. This will be hit on any negative value
```
Gymnasiast: Does the check (for negative values) make sense? I can imagine that you'd want to stop a loop once you go under zero.

duncanspumpkin:
I suspect it does make sense to check for negative numbers but it should really do that by comparing with 0 rather than anding with 0x8000. It requires a little more thought though so that was why I hadn't touched it.

duncanspumpkin:
Okay I had a look at the assembly and it is a check for negative so I will change it in this PR and remove the FIXME's

PR #10481, opened 2019-12-30
Fix #10477. Fix upper max height limit being very low for large scenery

### LINKED ISSUE #10477: Cannot place the Egyptian pyramid in the air
**OS:** Ubuntu 19.10
**Version:** 0.2.4+
**Commit/Build:** 4b0dff92aa871d3d64dad94856ba62e38a5c11c4

Once you start holding shift, the preview will disappear and attempting to place it will only yield `Cannot position this here...` with no further explanation.

Tagging @duncanspumpkin and @tupaschoal .

duncanspumpkin:
Mistake made while improving mouse control when zoomed out caused the upper limit for multi tile scenery to be limited by the sum of all clearance z values instead of the greatest clearance z value.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -8,7 +8,7 @@
 - Fix: [#10228] Can't import RCT1 Deluxe from Steam.
 - Fix: [#10325] Crash when banners have no text.
 - Fix: [#10420] Money effect causing false positive desync.
-- Fix: [#10477] Multitile Scenery unable to be placed higher using shift.
+- Fix: [#10477] Large Scenery cannot be placed higher using SHIFT.
 
 0.2.4 (2019-10-28)
 ------------------------------------------------------------------------
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -8,6 +8,7 @@
 - Fix: [#10228] Can't import RCT1 Deluxe from Steam.
 - Fix: [#10325] Crash when banners have no text.
 - Fix: [#10420] Money effect causing false positive desync.
+- Fix: [#10477] Multitile Scenery unable to be placed higher using shift.
 
 0.2.4 (2019-10-28)
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -1250,13 +1250,12 @@ static void sub_6E1F34(
         rct_scenery_entry* scenery_entry = get_large_scenery_entry(selected_scenery);
         if (scenery_entry)
         {
-            int16_t tileZ = 0;
+            int16_t maxClearZ = 0;
             for (int32_t i = 0; scenery_entry->large_scenery.tiles[i].x_offset != -1; ++i)
             {
-                assert(i < MAXIMUM_MAP_SIZE_TECHNICAL);
-                tileZ += scenery_entry->large_scenery.tiles[i].z_clearance;
+                maxClearZ = std::max(maxClearZ, static_cast<int16_t>(scenery_entry->large_scenery.tiles[i].z_clearance));
```
AaronVanGeffen: Perhaps more readable as:

```suggestion
                maxClearZ = std::max<int16_t>(maxClearZ, scenery_entry->large_scenery.tiles[i].z_clearance);
```

(Untested, mind.)

distribution/changelog.txt
```diff
@@ -8,6 +8,7 @@
 - Fix: [#10228] Can't import RCT1 Deluxe from Steam.
 - Fix: [#10325] Crash when banners have no text.
 - Fix: [#10420] Money effect causing false positive desync.
+- Fix: [#10477] Multitile Scenery unable to be placed higher using shift.
```
Gymnasiast:
```suggestion
- Fix: [#10477] Large scenery cannot be placed higher using SHIFT.
```

AaronVanGeffen: Any particular reason for capitalising 'shift' as such?

duncanspumpkin: As it's meaning the SHIFT key

AaronVanGeffen: Of course, but I've never seen it referred to in text in all-caps. Just the initial letter, usually.

Gymnasiast: We refer to it in ALL CAPS in our shortcut keys dialog, too.

PR #10484, opened 2019-12-30
Refactor calls to staff_is_patrol_area_set()

Gymnasiast:


Gymnasiast:
Travis is timing out again :-/

src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -347,7 +347,7 @@ static void window_staff_list_tooldown(rct_window* w, rct_widgetindex widgetInde
         if (footpathCoords.x == LOCATION_NULL)
             return;
 
-        bool isPatrolAreaSet = staff_is_patrol_area_set(200 + selectedPeepType, footpathCoords.x, footpathCoords.y);
+        bool isPatrolAreaSet = staff_is_patrol_area_set_for_type((STAFF_TYPE)selectedPeepType, footpathCoords);
```
duncanspumpkin: do a `static_cast<STAFF_TYPE>(` instead of the C style cast. Same goes for the other instances.

src/openrct2/peep/Staff.cpp
```diff
@@ -386,8 +378,12 @@ void staff_reset_stats()
     }
 }
 
-bool staff_is_patrol_area_set(int32_t staffIndex, int32_t x, int32_t y)
+static bool staff_is_patrol_area_set(int32_t staffIndex, int32_t x, int32_t y)
 {
+    // Patrol quads are stored in a bit map (8 patrol quads per byte)
+    // Each patrol quad is 4x4
+    // Therefore there are in total 64 x 64 patrol quads in the 256 x 256 map
+
```
duncanspumpkin: May as well add a comment that staff types have a common patrol area in the array as well after the `200 (STAFF_MAX_COUNT)`.

PR #10488, opened 2019-12-31
Refactor map_get_footpath_element() and fix #10486

### LINKED ISSUE #10486: Path breaks between level and inclined parts
**OS:** macOS (x86-64)
**Version:** 0.2.4
**Commit/Build:** 2c359736e on develop

   - No: this would be a bug in RCT2 vanilla
   - No: single player

**Steps to reproduce:**
1. Build a level path
2. Build an inclined path into the air

**Save game:**

Zipped save game: [path-break-bug.sv6.zip](https://github.com/OpenRCT2/OpenRCT2/files/4011343/path-break-bug.sv6.zip)

Gymnasiast:


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3616,13 +3616,12 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
...
     {
         TileElement* pathsByDir[4];
-        constexpr TileCoordsXY DirOffsets[4] = { { -1, 0 }, { 0, 1 }, { 1, 0 }, { 0, -1 } };
 
         bool keepOrientation = false;
         for (int8_t i = 0; i < 4; i++)
         {
-            pathsByDir[i] = map_get_footpath_element(
-                (mapCoords->x >> 5) + DirOffsets[i].x, (mapCoords->y >> 5) + DirOffsets[i].y, z >> 3);
+            pathsByDir[i] = map_get_footpath_element({ mapCoords->x + (DirectionOffsets[i].x * COORDS_XY_STEP),
```
duncanspumpkin: Use the CoordsXY direction delta instead then you don't need to multiply and can use the + overload

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3632,8 +3631,9 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
             // Sloped path on the level below
             if (!pathsByDir[i])
             {
-                pathsByDir[i] = map_get_footpath_element(
-                    (mapCoords->x >> 5) + DirOffsets[i].x, (mapCoords->y >> 5) + DirOffsets[i].y, (z >> 3) - 2);
+                pathsByDir[i] = map_get_footpath_element({ mapCoords->x + (DirectionOffsets[i].x * COORDS_XY_STEP),
+                                                           mapCoords->y + (DirectionOffsets[i].y * COORDS_XY_STEP),
+                                                           z - PATH_HEIGHT_STEP });
```
duncanspumpkin: As above

src/openrct2-ui/windows/RideConstruction.cpp
Gymnasiast: Nice, I wasn't too happy about these lines either. :+1:

erikhuizinga:
Thanks for picking this up quickly, as always. Enjoy your new year's eves!

src/openrct2-ui/windows/RideConstruction.cpp
duncanspumpkin: `CoordsDirectionDelta` is the variable you should use. `pathsByDir[i] = map_get_footpath_element({ *mapCoords + CoordsDirectionDelta[i], z });

PR #10490, opened 2019-12-31
Fix #10489: Hosts last action not being synchronized

### LINKED ISSUE #10489: Last action of host player is not synchronized
Every action the host does is not synchronized on the clients. The condition as seen below is incorrect.
https://github.com/OpenRCT2/OpenRCT2/blob/29922da871c0a33f89cf43a694afb1da0c4ef068/src/openrct2/actions/GameAction.cpp#L421
Execute is ran on server and client so this should be shared code.

ZehMatt:
This code should be shared and not be exclusive for the server.


ZehMatt:
No network bump is required, this fixes the client.

PR #10492, opened 2019-12-31
Refactor update_sound_params to be member functions

duncanspumpkin:
Moving a few items around to make it a bit easier to change how this works.

### DIFF
--- a/src/openrct2/ride/Vehicle.h
+++ b/src/openrct2/ride/Vehicle.h
@@ -148,6 +148,8 @@ enum VEHICLE_STATUS
     VEHICLE_STATUS_STOPPED_BY_BLOCK_BRAKES
 };
 
+struct rct_vehicle_sound_params;
+
 struct rct_vehicle : rct_sprite_common
 {
     uint8_t vehicle_sprite_type; // 0x1F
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Vehicle.h
```diff
@@ -274,6 +276,12 @@ struct rct_vehicle : rct_sprite_common
     void Invalidate();
     void SetState(VEHICLE_STATUS vehicleStatus, uint8_t subState = 0);
     bool IsGhost() const;
+    void UpdateSoundParams() const;
+
+private:
+    bool SoundCanPlay() const;
```
tupaschoal: All the others are like spoken word `GetSoundPriority`, `UpdateSoundParams`, `CreateSoundParam`. This one is a bit Yoda-ish with `SoundCanPlay`, I guess? Just a random comment

Gymnasiast: It makes sense if you put it into an if statement, though: 
```C++
if (SoundCanPlay)
    doStuff();
```

tupaschoal: But then it would sound even better, like a question, if it was `CanPlaySound`, no?
```c++
if (CanPlaySound())
    doStuff();
```

duncanspumpkin: I was originally going to prefix all of these sound functions so that it was SoundXXXX. I still don't really like the name anyway and will think of something better.

Gymnasiast: @tupaschoal That sounds less like a question, IMO. (Or at least: less like an English question. It might be different for Portuguese.)

tupaschoal: Hmm, It's no big deal anyway, I think the message gets through on both, and I'm not 100% happy with either :P

PR #10494, opened 2020-01-01
Fixed #10464 error message showing wrong amount when placing track with not enough cash

### LINKED ISSUE #10464: Placing a track design with insufficient funds shows incorrect message
image

klaframboise:
#10464 has been fixed. Added my name in contributors under bug fixes.


duncanspumpkin:
We really need to go through and refactor out the global variables for error messages. This wont be the last time something like this happens.

Issue #10358, opened 2019-12-11
Clean up leftover references to openrct2.website (and website.openrct2)


Gymnasiast:

Gymnasiast:
Fixed by https://github.com/OpenRCT2/OpenRCT2/pull/10503.

PR #10503, opened 2020-01-02
Implement #10358: Clean up leftover references to openrct2.website

Gymnasiast:


Gymnasiast:
@NuclearFej Could you test for me if this doesn't break anything on Android?

NuclearFej:
Yes, I'll take a look at it later. Doubt it'll break anything though.

NuclearFej:
There are a couple of problems.

First is that Java requires that the directory structure match the name of the package, so the folder "website" has to be renamed to "io". That part's easy.

Your repo doesn't actually build (the Android version) since it lacks (at least one of?) the changes I made that have _just_ been merged - so I made identical changes to my own local repo. It builds but crashes immediately. The debugger breaks on an exception involving JNI. Not sure why, going to have to take another look at it later.

Gymnasiast:
@NuclearFej The JNI had some leftover references to website as well, but separated with slashes instead of dots.

I have fixed those occurrences and renamed the folder. Could you retest?

src/openrct2/network/Network.cpp
```diff
@@ -1767,7 +1767,7 @@ void Network::Server_Send_GAMEINFO(NetworkConnection& connection)
     json_t* jsonProvider = json_object();
     json_object_set_new(jsonProvider, "name", json_string(gConfigNetwork.provider_name.c_str()));
     json_object_set_new(jsonProvider, "email", json_string(gConfigNetwork.provider_email.c_str()));
-    json_object_set_new(jsonProvider, "website", json_string(gConfigNetwork.provider_website.c_str()));
+    json_object_set_new(jsonProvider, "io", json_string(gConfigNetwork.provider_website.c_str()));
```
Gymnasiast: Revert

src/openrct2/network/Network.cpp
```diff
@@ -3141,7 +3141,7 @@ void Network::Client_Handle_GAMEINFO([[maybe_unused]] NetworkConnection& connect
     {
         ServerProviderName = json_stdstring_value(json_object_get(jsonProvider, "name"));
         ServerProviderEmail = json_stdstring_value(json_object_get(jsonProvider, "email"));
-        ServerProviderWebsite = json_stdstring_value(json_object_get(jsonProvider, "website"));
+        ServerProviderWebsite = json_stdstring_value(json_object_get(jsonProvider, "io"));
```
Gymnasiast: Also revert

NuclearFej:
I'm still working on this, wanted to let you know just so you know I'm not ghosting it.

It's still crashing. The first problem was of my own design, for which I apologize - the Android build requires that a desktop build be built first, so that build artifacts are present in the /bin folder which can be copied to the Android assets folder. This will get fixed "eventually". The second... well it _shouldn't_ be related. It isn't able to create the folder /sdcard/openrct2. This used to work, and the app has the Storage permission so it should be fine. I'm going to have to look more into this.

NuclearFej:
Sorry for the wait, I've been real busy the past few days. The issue is that SDK 29 doesn't like it when you write files directly to /sdcard, they're deprecating access to shared storage - which is insane, [but here you go.](https://developer.android.com/reference/android/os/Environment#getExternalStorageDirectory()) All of the code that touches the filesystem will have to be rewritten eventually to deal with this. In the meantime I've changed it to target SDK 28 in my fork, as well as to fix a hardcoded path (which might break on some devices?) These fixes make it build fine. I'm sorry for not fixing this sooner.

Once you merge this I'll make a PR with the fix.

PR #10510, opened 2020-01-03
Fix ride fence drawing when stations are above each other

Gymnasiast:
The code didn't check for the Z coordinate, and because of that, it would remove the fence on both lift platforms.


Gymnasiast:
Closing as it seems that the issue is down to the game sharing the entrance between the stations.

PR #10462, opened 2019-12-28
Make Map::map_can_construct_*() use CoordsXYRangedZ

tupaschoal:

PR #10511, opened 2020-01-04
 Improve variable names in map_can_construct_with_clear_at() and use big Z

tupaschoal:
This is a followup to @duncanspumpkin on https://github.com/OpenRCT2/OpenRCT2/pull/10462#discussion_r361832868

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -1266,8 +1266,8 @@ bool map_can_construct_with_clear_at(
     const CoordsXYRangedZ& pos, CLEAR_FUNC clearFunc, QuarterTile quarterTile, uint8_t flags, money32* price,
     uint8_t crossingMode)
 {
-    int32_t al, ah, bh, cl, ch, water_height;
-    al = ah = bh = cl = ch = water_height = 0;
+    int32_t northZ, eastZ, baseHeight, southZ, westZ, water_height;
+    northZ = eastZ = baseHeight = southZ = westZ = water_height = 0;
     uint8_t slope = 0;
 
     gMapGroundFlags = ELEMENT_IS_ABOVE_GROUND;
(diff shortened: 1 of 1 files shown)


src/openrct2/world/Map.cpp
```diff
@@ -1312,14 +1312,11 @@ bool map_can_construct_with_clear_at(
     loc_68B9B7:
         if (gParkFlags & PARK_FLAGS_FORBID_HIGH_CONSTRUCTION)
         {
-            al = pos.clearanceZ / 8 - tileElement->base_height;
-            if (al >= 0)
+            northZ = pos.clearanceZ - tileElement->GetBaseZ();
+            if (northZ >= 0 && northZ > (18 * 8))
```
Gymnasiast:
```suggestion
            if (northZ >= 0 && northZ > (18 * COORDS_Z_STEP))
```

src/openrct2/world/Map.cpp
```diff
@@ -1340,43 +1337,43 @@ bool map_can_construct_with_clear_at(
...
-                ch = al;
+                northZ = tileElement->GetBaseZ();
+                eastZ = northZ;
+                southZ = northZ;
+                westZ = northZ;
                 slope = tileElement->AsSurface()->GetSlope();
                 if (slope & TILE_ELEMENT_SLOPE_N_CORNER_UP)
                 {
-                    al += 2;
+                    northZ += (2 * 8);
```
Gymnasiast: Also use COORDS_Z_STEP on all these.

src/openrct2/world/Map.cpp
duncanspumpkin: Also this instance of variable al should be renamed to `heightFromGround`. I was making that change anyway in the #10501

tupaschoal: Ok, did both.

Issue #10508, opened 2020-01-03
Improvement: Add Cycle Detection to Ride Ratings Calculation


bitman2049:
**Version:** 0.2.4

An issue that gets reported often here and elsewhere is some variation of ride ratings not being calculated. The issue, almost always, has to do with merged tracks. More specifically that if you merge tracks and create a cycle that doesn't end in a station platform, the ratings function will get stuck in the loop and never calculate any ratings for the park until the offending ride(s) are closed.

Simple cycle detection could be added to the part of the ride ratings task that traverses tracked rides. It would only need to update the state to RIDE_RATINGS_STATE_FIND_NEXT_RIDE if a cycle is detected. This would help in parks with shoestringed rides and other hacked attractions.

- [*] Reproducible in RCT2 (vanilla)?
- [ ] Specific to multiplayer?

**Steps to reproduce:**
1. Build a hacked ride that includes a cycle in the track, then open or test the ride.
2. Build any other ride (flat ride, coaster, etc.) and open/test that ride as well.

Expected result: The second ride will get excitement/intensity/nausea ratings calculated.
Actual result: The second ride will never get ratings calculated.

**Related issues:**
https://github.com/OpenRCT2/OpenRCT2/issues/4695
https://github.com/OpenRCT2/OpenRCT2/issues/5257
https://github.com/OpenRCT2/OpenRCT2/issues/8532

bitman2049:
My own proposed solution is simple, but it could miss some edge cases. It's a variation of the tortoise and hare algorithm.

I propose adding a second XYZ iterator that gets updated every other step while the regular XYZ iterator is traversing the track. At each step, check if x1==x2, y1==y2, and z1==z2. If all those cases are met, it means that the slower iterator has caught up to the normal speed iterator. Since the regular speed iterator will stop in normal cases when it hits a station or a tile with no valid track, the slower iterator can only catch up if the normal speed iterator hits a cycle.

Having a second iterator would slow the traversal speed by ~1/3 in all cases, but it's cheap in memory since it doesn't require keeping a history of visited track pieces.

jensj12:
The tortoise and hare algorithm already exists in a few places. Here are the related PR's: #4359 #7062

bitman2049:
It looks like the iterator has to be a different type depending on whether you're iterating forward or backwards, am I reading that right?

PR #10512, opened 2020-01-04
Add Cycle Detection to Ride Rating Calculation

bitman2049:
Hacked rides no longer block the ride rating calculation function.

Issue link: https://github.com/OpenRCT2/OpenRCT2/issues/10508

### DIFF
--- a/src/openrct2/ride/RideRatings.h
+++ b/src/openrct2/ride/RideRatings.h
@@ -54,6 +54,10 @@ struct RideRatingCalculationData
     uint16_t num_brakes;
     uint16_t num_reversers;
     uint16_t station_flags;
+};
+
+struct RideRatingCycleDetectionData
+{
     uint16_t cycle_proximity_x;
     uint16_t cycle_proximity_y;
     uint16_t cycle_proximity_z;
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/RideRatings.h
```diff
@@ -54,6 +54,11 @@ struct RideRatingCalculationData
     uint16_t num_brakes;
     uint16_t num_reversers;
     uint16_t station_flags;
+    uint16_t cycle_proximity_x;
+    uint16_t cycle_proximity_y;
+    uint16_t cycle_proximity_z;
+    uint8_t cycle_proximity_track_type;
+    bool cycle_detection_toggle;
 };
```
IntelOrca: You have broken the SV6 format doing this. This is why we haven't tried doing this before.

bitman2049: Ah. Would it work if I put these in a different struct?

IntelOrca: Yes, remember that it also needs to reset for everyone when a new player joins a multiplayer game too.

PR #10515, opened 2020-01-04
Use CoordsXY* on world/Footpath.cpp

tupaschoal:
I'm trying to get rid of non coords usage in Footpath. Some notes:

- Some places seemed like they could use a `CoordsXYRangedZD` with the `Direction`, but I was not sure if this should be done.
- Some other places could benefit from Map's `CoordsXYE`, would that be ok?
- The only left are `footpath_update_path_wide_flags(int32_t x, int32_t y);` which I couldn't tell for sure are `CoordsXY` and `footpath_get_coordinates_from_pos` + `footpath_bridge_get_info_from_pos` which would involve changing from altering parameter to returning coords.

### DIFF
--- a/src/openrct2/Editor.cpp
+++ b/src/openrct2/Editor.cpp
@@ -500,12 +500,9 @@ namespace Editor
 
         for (const auto& parkEntrance : gParkEntrances)
         {
-            int32_t x = parkEntrance.x;
-            int32_t y = parkEntrance.y;
-            int32_t z = parkEntrance.z / 8;
             int32_t direction = direction_reverse(parkEntrance.direction);
 
-            switch (footpath_is_connected_to_map_edge(x, y, z, direction, 0))
+            switch (footpath_is_connected_to_map_edge(parkEntrance, direction, 0))
             {
                 case FOOTPATH_SEARCH_NOT_FOUND:
                     gGameCommandErrorText = STR_PARK_ENTRANCE_WRONG_DIRECTION_OR_NO_PATH;
@@ -516,7 +513,7 @@ namespace Editor
                     return false;
                 case FOOTPATH_SEARCH_SUCCESS:
                     // Run the search again and unown the path
-                    footpath_is_connected_to_map_edge(x, y, z, direction, (1 << 5));
+                    footpath_is_connected_to_map_edge(parkEntrance, direction, (1 << 5));
                     break;
             }
         }
--- a/src/openrct2/actions/LargeSceneryPlaceAction.hpp
+++ b/src/openrct2/actions/LargeSceneryPlaceAction.hpp
@@ -304,7 +304,7 @@ public:
 
             if (!(GetFlags() & GAME_COMMAND_FLAG_GHOST))
             {
-                footpath_remove_litter(curTile.x, curTile.y, zLow * 8);
+                footpath_remove_litter({ curTile, zLow * COORDS_Z_STEP });
                 if (!gCheatsDisableClearanceChecks)
                 {
                     wall_remove_at({ curTile, zLow * 8, zHigh * 8 });
--- a/src/openrct2/actions/PlaceParkEntranceAction.hpp
+++ b/src/openrct2/actions/PlaceParkEntranceAction.hpp
@@ -171,7 +171,7 @@ public:
 
             if (!(flags & GAME_COMMAND_FLAG_GHOST))
             {
-                footpath_connect_edges(entranceLoc.x, entranceLoc.y, newElement, 1);
+                footpath_connect_edges(entranceLoc, newElement, GAME_COMMAND_FLAG_APPLY);
             }
 
             update_park_fences(entranceLoc);
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -137,20 +137,19 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
  */
 static int32_t guest_surface_path_finding(Peep* peep)
 {
-    int16_t x = peep->next_x;
-    int16_t y = peep->next_y;
-    int16_t z = peep->next_z;
+    auto pathPos = CoordsXYRangedZ{ peep->next_x, peep->next_y, peep->next_z * COORDS_Z_STEP,
+                                    (peep->next_z * COORDS_Z_STEP) + PATH_HEIGHT };
     Direction randDirection = scenario_rand() & 3;
 
-    if (!fence_in_the_way(x, y, z, z + 4, randDirection))
+    if (!fence_in_the_way(pathPos, randDirection))
     {
-        x += CoordsDirectionDelta[randDirection].x;
-        y += CoordsDirectionDelta[randDirection].y;
+        pathPos.x += CoordsDirectionDelta[randDirection].x;
+        pathPos.y += CoordsDirectionDelta[randDirection].y;
         Direction backwardsDirection = direction_reverse(randDirection);
 
-        if (!fence_in_the_way(x, y, z, z + 4, backwardsDirection))
+        if (!fence_in_the_way(pathPos, backwardsDirection))
         {
-            if (!map_surface_is_blocked({ x, y }))
+            if (!map_surface_is_blocked(pathPos))
             {
                 return peep_move_one_tile(randDirection, peep);
             }
@@ -165,17 +164,17 @@ static int32_t guest_surface_path_finding(Peep* peep)
     }
     randDirection &= 3;
 
-    x = peep->next_x;
-    y = peep->next_y;
-    if (!fence_in_the_way(x, y, z, z + 4, randDirection))
+    pathPos.x = peep->next_x;
+    pathPos.y = peep->next_y;
+    if (!fence_in_the_way(pathPos, randDirection))
     {
-        x += CoordsDirectionDelta[randDirection].x;
-        y += CoordsDirectionDelta[randDirection].y;
+        pathPos.x += CoordsDirectionDelta[randDirection].x;
+        pathPos.y += CoordsDirectionDelta[randDirection].y;
         Direction backwardsDirection = direction_reverse(randDirection);
 
-        if (!fence_in_the_way(x, y, z, z + 4, backwardsDirection))
+        if (!fence_in_the_way(pathPos, backwardsDirection))
         {
-            if (!map_surface_is_blocked({ x, y }))
+            if (!map_surface_is_blocked(pathPos))
             {
                 return peep_move_one_tile(randDirection, peep);
             }
@@ -185,17 +184,17 @@ static int32_t guest_surface_path_finding(Peep* peep)
     randDirection -= 2;
     randDirection &= 3;
 
-    x = peep->next_x;
-    y = peep->next_y;
-    if (!fence_in_the_way(x, y, z, z + 4, randDirection))
+    pathPos.x = peep->next_x;
+    pathPos.y = peep->next_y;
+    if (!fence_in_the_way(pathPos, randDirection))
     {
-        x += CoordsDirectionDelta[randDirection].x;
-        y += CoordsDirectionDelta[randDirection].y;
+        pathPos.x += CoordsDirectionDelta[randDirection].x;
+        pathPos.y += CoordsDirectionDelta[randDirection].y;
         Direction backwardsDirection = direction_reverse(randDirection);
 
-        if (!fence_in_the_way(x, y, z, z + 4, backwardsDirection))
+        if (!fence_in_the_way(pathPos, backwardsDirection))
         {
-            if (!map_surface_is_blocked({ x, y }))
+            if (!map_surface_is_blocked(pathPos))
             {
                 return peep_move_one_tile(randDirection, peep);
             }
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -727,10 +727,16 @@ static uint8_t staff_direction_surface(Peep* peep, uint8_t initialDirection)
 
         direction &= 3;
 
-        if (fence_in_the_way(peep->next_x, peep->next_y, peep->next_z, peep->next_z + 4, direction))
+        if (fence_in_the_way(
+                { peep->next_x, peep->next_y, peep->next_z * COORDS_Z_STEP,
+                  (peep->next_z * COORDS_Z_STEP) + PEEP_CLEARANCE_HEIGHT },
+                direction))
             continue;
 
-        if (fence_in_the_way(peep->next_x, peep->next_y, peep->next_z, peep->next_z + 4, direction_reverse(direction)))
+        if (fence_in_the_way(
+                { peep->next_x, peep->next_y, peep->next_z * COORDS_Z_STEP,
+                  (peep->next_z * COORDS_Z_STEP) + PEEP_CLEARANCE_HEIGHT },
+                direction_reverse(direction)))
             continue;
 
         CoordsXY chosenTile = { peep->next_x + CoordsDirectionDelta[direction].x,
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -421,9 +421,9 @@ void footpath_remove_litter(int32_t x, int32_t y, int32_t z)
  *
  *  rct2: 0x0069A48B
  */
-void footpath_interrupt_peeps(int32_t x, int32_t y, int32_t z)
+void footpath_interrupt_peeps(const CoordsXYZ& footpathPos)
 {
-    uint16_t spriteIndex = sprite_get_first_in_quadrant(x, y);
+    uint16_t spriteIndex = sprite_get_first_in_quadrant(footpathPos.x, footpathPos.y);
     while (spriteIndex != SPRITE_INDEX_NULL)
     {
         Peep* peep = &get_sprite(spriteIndex)->peep;
@@ -432,7 +432,7 @@ void footpath_interrupt_peeps(int32_t x, int32_t y, int32_t z)
         {
             if (peep->state == PEEP_STATE_SITTING || peep->state == PEEP_STATE_WATCHING)
             {
-                if (peep->z == z)
+                if (peep->z == footpathPos.z)
                 {
                     peep->SetState(PEEP_STATE_WALKING);
                     peep->destination_x = (peep->x & 0xFFE0) + 16;
@@ -917,7 +917,7 @@ static void loc_6A6D7E(
         }
         if (!(flags & (GAME_COMMAND_FLAG_GHOST | GAME_COMMAND_FLAG_ALLOW_DURING_PAUSED)))
         {
-            footpath_interrupt_peeps(targetPos.x, targetPos.y, tileElement->GetBaseZ());
+            footpath_interrupt_peeps({ targetPos, tileElement->GetBaseZ() });
         }
         map_invalidate_element(targetPos, tileElement);
     }
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -1055,34 +1055,36 @@ void footpath_connect_edges(const CoordsXY& footpathPos, TileElement* tileElemen
  *  rct2: 0x006A742F
  */
 void footpath_chain_ride_queue(
-    ride_id_t rideIndex, int32_t entranceIndex, int32_t x, int32_t y, TileElement* tileElement, int32_t direction)
+    ride_id_t rideIndex, int32_t entranceIndex, const CoordsXY& initialFootpathPos, TileElement* const initialTileElement,
+    int32_t direction)
 {
     TileElement *lastPathElement, *lastQueuePathElement;
-    int32_t lastPathX = x, lastPathY = y, lastPathDirection = direction;
+    auto tileElement = initialTileElement;
+    auto curQueuePos = initialFootpathPos;
+    auto lastPath = curQueuePos;
+    int32_t baseZ = tileElement->GetBaseZ();
+    int32_t lastPathDirection = direction;
 
     lastPathElement = nullptr;
     lastQueuePathElement = nullptr;
-    int32_t z = tileElement->base_height;
     for (;;)
     {
         if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH)
         {
             lastPathElement = tileElement;
-            lastPathX = x;
-            lastPathY = y;
+            lastPath = curQueuePos;
             lastPathDirection = direction;
             if (tileElement->AsPath()->IsSloped())
             {
                 if (tileElement->AsPath()->GetSlopeDirection() == direction)
                 {
-                    z += 2;
+                    baseZ += LAND_HEIGHT_STEP;
                 }
             }
         }
 
-        x += CoordsDirectionDelta[direction].x;
-        y += CoordsDirectionDelta[direction].y;
-        tileElement = map_get_first_element_at({ x, y });
+        auto targetQueuePos = curQueuePos + CoordsDirectionDelta[direction];
+        tileElement = map_get_first_element_at(targetQueuePos);
         if (tileElement != nullptr)
         {
             do
@@ -1091,7 +1093,7 @@ void footpath_chain_ride_queue(
                     continue;
                 if (tileElement->GetType() != TILE_ELEMENT_TYPE_PATH)
                     continue;
-                if (tileElement->base_height == z)
+                if (tileElement->GetBaseZ() == baseZ)
                 {
                     if (tileElement->AsPath()->IsSloped())
                     {
@@ -1100,7 +1102,7 @@ void footpath_chain_ride_queue(
                     }
                     goto foundNextPath;
                 }
-                if (tileElement->base_height == z - 2)
+                if (tileElement->GetBaseZ() == baseZ - LAND_HEIGHT_STEP)
                 {
                     if (!tileElement->AsPath()->IsSloped())
                         break;
@@ -1108,7 +1110,7 @@ void footpath_chain_ride_queue(
                     if (direction_reverse(tileElement->AsPath()->GetSlopeDirection()) != direction)
                         break;
 
-                    z -= 2;
+                    baseZ -= LAND_HEIGHT_STEP;
                     goto foundNextPath;
                 }
             } while (!(tileElement++)->IsLastForTile());
@@ -1136,7 +1138,8 @@ void footpath_chain_ride_queue(
             tileElement->AsPath()->SetRideIndex(rideIndex);
             tileElement->AsPath()->SetStationIndex(entranceIndex);
 
-            map_invalidate_element({ x, y }, tileElement);
+            curQueuePos = targetQueuePos;
+            map_invalidate_element(targetQueuePos, tileElement);
 
             if (lastQueuePathElement == nullptr)
             {
@@ -1164,7 +1167,7 @@ void footpath_chain_ride_queue(
             lastPathElement->AsPath()->SetHasQueueBanner(true);
             lastPathElement->AsPath()->SetQueueBannerDirection(lastPathDirection); // set the ride sign direction
 
-            map_animation_create(MAP_ANIMATION_TYPE_QUEUE_BANNER, { lastPathX, lastPathY, lastPathElement->GetBaseZ() });
+            map_animation_create(MAP_ANIMATION_TYPE_QUEUE_BANNER, { lastPath, lastPathElement->GetBaseZ() });
         }
     }
 }
@@ -1222,7 +1225,7 @@ void footpath_update_queue_chains()
                         continue;
 
                     Direction direction = direction_reverse(tileElement->GetDirection());
-                    footpath_chain_ride_queue(rideIndex, i, location.x << 5, location.y << 5, tileElement, direction);
+                    footpath_chain_ride_queue(rideIndex, i, location.ToCoordsXY(), tileElement, direction);
                 } while (!(tileElement++)->IsLastForTile());
             }
         }
@@ -1902,7 +1905,7 @@ void footpath_update_queue_entrance_banner(const CoordsXY& footpathPos, TileElem
                 {
                     if (tileElement->AsPath()->GetEdges() & (1 << direction))
                     {
-                        footpath_chain_ride_queue(255, 0, footpathPos.x, footpathPos.y, tileElement, direction);
+                        footpath_chain_ride_queue(255, 0, footpathPos, tileElement, direction);
                     }
                 }
                 tileElement->AsPath()->SetRideIndex(RIDE_ID_NULL);
@@ -1912,8 +1915,7 @@ void footpath_update_queue_entrance_banner(const CoordsXY& footpathPos, TileElem
             if (tileElement->AsEntrance()->GetEntranceType() == ENTRANCE_TYPE_RIDE_ENTRANCE)
             {
                 footpath_queue_chain_push(tileElement->AsEntrance()->GetRideIndex());
-                footpath_chain_ride_queue(
-                    255, 0, footpathPos.x, footpathPos.y, tileElement, direction_reverse(tileElement->GetDirection()));
+                footpath_chain_ride_queue(255, 0, footpathPos, tileElement, direction_reverse(tileElement->GetDirection()));
             }
             break;
     }
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -1921,44 +1921,43 @@ void footpath_update_queue_entrance_banner(const CoordsXY& footpathPos, TileElem
  *  rct2: 0x006A6B7F
  */
 static void footpath_remove_edges_towards_here(
-    int32_t x, int32_t y, int32_t z, int32_t direction, TileElement* tileElement, bool isQueue)
+    const CoordsXYZ& footpathPos, int32_t direction, TileElement* tileElement, bool isQueue)
 {
     if (tileElement->AsPath()->IsQueue())
     {
         footpath_queue_chain_push(tileElement->AsPath()->GetRideIndex());
     }
 
-    int32_t d = direction_reverse(direction);
+    auto d = direction_reverse(direction);
     tileElement->AsPath()->SetEdges(tileElement->AsPath()->GetEdges() & ~(1 << d));
     int32_t cd = ((d - 1) & 3);
     tileElement->AsPath()->SetCorners(tileElement->AsPath()->GetCorners() & ~(1 << cd));
     cd = ((cd + 1) & 3);
     tileElement->AsPath()->SetCorners(tileElement->AsPath()->GetCorners() & ~(1 << cd));
-    map_invalidate_tile({ x, y, tileElement->GetBaseZ(), tileElement->GetClearanceZ() });
+    map_invalidate_tile({ footpathPos, tileElement->GetBaseZ(), tileElement->GetClearanceZ() });
 
     if (isQueue)
-        footpath_disconnect_queue_from_path({ x, y }, tileElement, -1);
+        footpath_disconnect_queue_from_path(footpathPos, tileElement, -1);
 
-    direction = (direction + 1) & 3;
-    x += CoordsDirectionDelta[direction].x;
-    y += CoordsDirectionDelta[direction].y;
+    Direction shiftedDirection = (direction + 1) & 3;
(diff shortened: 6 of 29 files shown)


src/openrct2/actions/LargeSceneryPlaceAction.hpp
```diff
@@ -304,7 +304,7 @@ DEFINE_GAME_ACTION(LargeSceneryPlaceAction, GAME_COMMAND_PLACE_LARGE_SCENERY, La
 
             if (!(GetFlags() & GAME_COMMAND_FLAG_GHOST))
             {
-                footpath_remove_litter(curTile.x, curTile.y, zLow * 8);
+                footpath_remove_litter({ curTile, zLow * 8 });
```
Gymnasiast:
```suggestion
                footpath_remove_litter({ curTile, zLow * COORDS_Z_STEP });
```

src/openrct2/actions/PlaceParkEntranceAction.hpp
```diff
@@ -171,7 +171,7 @@ DEFINE_GAME_ACTION(PlaceParkEntranceAction, GAME_COMMAND_PLACE_PARK_ENTRANCE, Ga
 
             if (!(flags & GAME_COMMAND_FLAG_GHOST))
             {
-                footpath_connect_edges(entranceLoc.x, entranceLoc.y, newElement, 1);
+                footpath_connect_edges(entranceLoc, newElement, 1);
```
Gymnasiast:
```suggestion
                footpath_connect_edges(entranceLoc, newElement, GAME_COMMAND_FLAG_APPLY);
```

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -142,13 +142,13 @@ static int32_t guest_surface_path_finding(Peep* peep)
     int16_t z = peep->next_z;
     Direction randDirection = scenario_rand() & 3;
 
-    if (!fence_in_the_way(x, y, z, z + 4, randDirection))
+    if (!fence_in_the_way({ x, y, z * 8, (z * 8) + 4 * 8 }, randDirection))
```
Gymnasiast: Can you create a constexpr called `PATH_HEIGHT` and define it as 4 * COORDS_Z_STEP, then use that here?

```suggestion
    if (!fence_in_the_way({ x, y, z * COORDS_Z_STEP, (z * COORDS_Z_STEP) + PATH_HEIGHT }, randDirection))
```

src/openrct2/world/Footpath.cpp
```diff
@@ -696,25 +692,26 @@ static TileElement* footpath_get_element(int32_t x, int32_t y, int32_t z0, int32
...
-        if (fence_in_the_way(x1, y1, tileElement->base_height, tileElement->clearance_height, direction_reverse(direction)))
+        if (fence_in_the_way(
+                { shiftedFootpathPos, tileElement->GetBaseZ(), tileElement->GetClearanceZ() }, direction_reverse(direction)))
             return false;
     }
 
-    int32_t z = tileElement->base_height;
-    TileElement* otherTileElement = footpath_get_element(x1, y1, z - 2, z, direction);
+    int32_t z = tileElement->GetBaseZ();
+    TileElement* otherTileElement = footpath_get_element({ shiftedFootpathPos, z - (2 * 8), z }, direction);
```
Gymnasiast: *cough* COORDS_Z_STEP *cough*

src/openrct2/world/Footpath.cpp
```diff
@@ -1279,39 +1273,39 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
...
-    x += CoordsDirectionDelta[direction].x;
-    y += CoordsDirectionDelta[direction].y;
+    auto shiftedFootpathPos = CoordsXYZD{ CoordsXY{ footpathPos } + CoordsDirectionDelta[footpathPos.direction], footpathPos.z,
+                                          footpathPos.direction };
     if (++level > 250)
         return FOOTPATH_SEARCH_TOO_COMPLEX;
 
     // Check if we are at edge of map
-    if (x < 32 || y < 32)
+    if (shiftedFootpathPos.x < 32 || shiftedFootpathPos.y < 32)
```
Gymnasiast: Use COORDS_XY_STEP.

src/openrct2/world/Footpath.cpp
```diff
@@ -1279,39 +1273,39 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
...
 
-        if (tileElement->AsPath()->IsSloped() && (slopeDirection = tileElement->AsPath()->GetSlopeDirection()) != direction)
+        if (tileElement->AsPath()->IsSloped()
+            && (slopeDirection = tileElement->AsPath()->GetSlopeDirection()) != shiftedFootpathPos.direction)
         {
-            if (direction_reverse(slopeDirection) != direction)
+            if (direction_reverse(slopeDirection) != shiftedFootpathPos.direction)
                 continue;
-            if (tileElement->base_height + 2 != z)
+            if (tileElement->GetBaseZ() + (2 * 8) != shiftedFootpathPos.z)
```
Gymnasiast: PATH_HEIGHT_STEP

Gymnasiast:
> Some places seemed like they could use a CoordsXYRangedZD with the Direction, but I was not sure if this should be done.

If they represent a thing together (e.g. an entrance, exit), then I would say yes.

> Some other places could benefit from Map's CoordsXYE, would that be ok?

Certainly, that's why it was created.

src/openrct2/world/Footpath.cpp
```diff
@@ -524,55 +524,51 @@ static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileEle
...
 
-        x += CoordsDirectionDelta[direction].x;
-        y += CoordsDirectionDelta[direction].y;
-        tileElement[1] = footpath_connect_corners_get_neighbour(x, y, z, (1 << direction_reverse(direction)));
+        currentPos += CoordsDirectionDelta[direction];
+        tileElement[1] = footpath_connect_corners_get_neighbour({ currentPos, z }, (1 << direction_reverse(direction)));
         if (tileElement[1] == nullptr)
             continue;
 
         direction = (direction + 1) & 3;
```
duncanspumpkin: direction_next

src/openrct2/world/Footpath.cpp
```diff
@@ -524,55 +524,51 @@ static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileEle
...
         tileElement[3]->AsPath()->SetCorners(tileElement[3]->AsPath()->GetCorners() | (1 << (direction)));
-        map_invalidate_element({ x, y }, tileElement[3]);
+        map_invalidate_element(currentPos, tileElement[3]);
 
         direction = (direction - 1) & 3;
         tileElement[2]->AsPath()->SetCorners(tileElement[2]->AsPath()->GetCorners() | (1 << (direction)));
 
-        map_invalidate_element({ x, y }, tileElement[2]);
+        map_invalidate_element(currentPos, tileElement[2]);
```
duncanspumpkin: This and subsequent calls to map_invalidate_element are calling with the wrong position. (note this is a bug in the current code)

src/openrct2/world/Footpath.cpp
```diff
@@ -524,55 +524,51 @@ static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileEle
...
+        tileElement[3] = footpath_connect_corners_get_neighbour({ currentPos, z }, (1 << ((direction + 1) & 3)));
         if (tileElement[3] == nullptr)
             continue;
 
         direction = (direction + 1) & 3;
         tileElement[3]->AsPath()->SetCorners(tileElement[3]->AsPath()->GetCorners() | (1 << (direction)));
-        map_invalidate_element({ x, y }, tileElement[3]);
+        map_invalidate_element(currentPos, tileElement[3]);
 
         direction = (direction - 1) & 3;
```
duncanspumpkin: direction_prev

src/openrct2/world/Footpath.cpp
```diff
@@ -731,13 +728,13 @@ static bool footpath_reconnect_queue_to_path(int32_t x, int32_t y, TileElement*
             otherTileElement->AsPath()->SetEdges(otherTileElement->AsPath()->GetEdges() | (1 << ((direction + 2) & 3)));
```
duncanspumpkin: These direction+2's should be direction_reverse()

src/openrct2/world/Footpath.cpp
```diff
@@ -881,59 +877,59 @@ static void loc_6A6D7E(
...
-            map_invalidate_element({ initialX, initialY }, initialTileElement);
+            initialTileElement->AsPath()->SetEdges(initialTileElement->AsPath()->GetEdges() | (1 << pos.direction));
+            map_invalidate_element(pos, initialTileElement);
         }
     }
 }
 
 static void loc_6A6C85(
-    int32_t x, int32_t y, int32_t direction, TileElement* tileElement, int32_t flags, bool query,
+    const CoordsXY& mapPos, int32_t direction, TileElement* tileElement, int32_t flags, bool query,
```
duncanspumpkin: mapPos is a bad name this is the `footpathPosition`

src/openrct2/world/Footpath.cpp
```diff
@@ -772,31 +769,29 @@ static bool footpath_disconnect_queue_from_path(int32_t x, int32_t y, TileElemen
...
  */
 static void loc_6A6D7E(
-    int32_t initialX, int32_t initialY, int32_t z, int32_t direction, TileElement* initialTileElement, int32_t flags,
-    bool query, rct_neighbour_list* neighbourList)
+    const CoordsXYZD& pos, TileElement* initialTileElement, int32_t flags, bool query, rct_neighbour_list* neighbourList)
 {
-    int32_t x = initialX + CoordsDirectionDelta[direction].x;
-    int32_t y = initialY + CoordsDirectionDelta[direction].y;
-    if (((gScreenFlags & SCREEN_FLAGS_SCENARIO_EDITOR) || gCheatsSandboxMode) && map_is_edge({ x, y }))
+    auto shiftedPos = CoordsXY{ pos } + CoordsDirectionDelta[pos.direction];
```
duncanspumpkin: shiftedPos represents the coordinates of of the footpath/ride/entrance/mapEdge that is hopefully being connected to. `targetPos` might be a better name.

src/openrct2/world/Footpath.cpp
```diff
@@ -772,31 +769,29 @@ static bool footpath_disconnect_queue_from_path(int32_t x, int32_t y, TileElemen
  *  rct2: 0x006A6D7E
  */
 static void loc_6A6D7E(
-    int32_t initialX, int32_t initialY, int32_t z, int32_t direction, TileElement* initialTileElement, int32_t flags,
-    bool query, rct_neighbour_list* neighbourList)
+    const CoordsXYZD& pos, TileElement* initialTileElement, int32_t flags, bool query, rct_neighbour_list* neighbourList)
```
duncanspumpkin: pos is the initialTileElements position

src/openrct2/world/Footpath.cpp
```diff
@@ -881,59 +877,59 @@ static void loc_6A6D7E(
...
         {
-            initialTileElement->AsPath()->SetEdges(initialTileElement->AsPath()->GetEdges() | (1 << direction));
-            map_invalidate_element({ initialX, initialY }, initialTileElement);
+            initialTileElement->AsPath()->SetEdges(initialTileElement->AsPath()->GetEdges() | (1 << pos.direction));
+            map_invalidate_element(pos, initialTileElement);
         }
     }
 }
 
 static void loc_6A6C85(
```
duncanspumpkin: Future refactoring job would be to change this into a simple check that validates if the direction should be fully checked with `loc_6A6D7E` and move the calling of `loc_6A6D7E` into the parent function.

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
...
-    ride_id_t rideIndex, int32_t entranceIndex, int32_t x, int32_t y, TileElement* tileElement, int32_t direction)
+    ride_id_t rideIndex, int32_t entranceIndex, const CoordsXY& footpathPos, TileElement* tileElement, int32_t direction)
 {
     TileElement *lastPathElement, *lastQueuePathElement;
-    int32_t lastPathX = x, lastPathY = y, lastPathDirection = direction;
+    auto lastPath = CoordsXYZD{ footpathPos, tileElement->GetBaseZ(), static_cast<Direction>(direction) };
 
     lastPathElement = nullptr;
     lastQueuePathElement = nullptr;
-    int32_t z = tileElement->base_height;
```
duncanspumpkin: undo this change. The z coordinate is not part of the last path coordinates.

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
...
-                    z += 2;
+                    lastPath.z += (2 * 8);
                 }
             }
         }
 
-        x += CoordsDirectionDelta[direction].x;
-        y += CoordsDirectionDelta[direction].y;
-        tileElement = map_get_first_element_at({ x, y });
+        auto shiftedPos = footpathPos + CoordsDirectionDelta[direction];
```
duncanspumpkin: `targetPos`

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
  *  rct2: 0x006A742F
  */
 void footpath_chain_ride_queue(
-    ride_id_t rideIndex, int32_t entranceIndex, int32_t x, int32_t y, TileElement* tileElement, int32_t direction)
+    ride_id_t rideIndex, int32_t entranceIndex, const CoordsXY& footpathPos, TileElement* tileElement, int32_t direction)
```
duncanspumpkin: Change footpathPos to `initialFootpathPos`. Rename `tileElement` to `initialTileElement` make it const.

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
...
 {
     TileElement *lastPathElement, *lastQueuePathElement;
```
duncanspumpkin: will also require `tileElement` set to the value of `initialTileElement`

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
...
-    int32_t lastPathX = x, lastPathY = y, lastPathDirection = direction;
+    auto lastPath = CoordsXYZD{ footpathPos, tileElement->GetBaseZ(), static_cast<Direction>(direction) };
```
duncanspumpkin: Create a new variable called `curQueuePos` initial value of `initialFootpathPos`

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
...
-            lastPathDirection = direction;
+            lastPath.x = footpathPos.x;
+            lastPath.y = footpathPos.y;
+            lastPath.direction = direction;
             if (tileElement->AsPath()->IsSloped())
             {
                 if (tileElement->AsPath()->GetSlopeDirection() == direction)
                 {
-                    z += 2;
+                    lastPath.z += (2 * 8);
```
duncanspumpkin: undo as above

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
...
-    int32_t z = tileElement->base_height;
     for (;;)
     {
         if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH)
         {
             lastPathElement = tileElement;
-            lastPathX = x;
-            lastPathY = y;
-            lastPathDirection = direction;
+            lastPath.x = footpathPos.x;
```
duncanspumpkin: value will be `curQueuePos`

src/openrct2/world/Footpath.cpp
duncanspumpkin: `targetQueuePos` = `curQueuePos` + ...

src/openrct2/world/Footpath.cpp
```diff
@@ -1130,7 +1124,7 @@ void footpath_chain_ride_queue(
             tileElement->AsPath()->SetRideIndex(rideIndex);
             tileElement->AsPath()->SetStationIndex(entranceIndex);
 
-            map_invalidate_element({ x, y }, tileElement);
+            map_invalidate_element(shiftedPos, tileElement);
```
duncanspumpkin: add `curQueuePos = targetQueuePos;` pass `curQueuePos` into this function

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1045,32 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
  *  rct2: 0x006A742F
  */
 void footpath_chain_ride_queue(
```
duncanspumpkin: The changes you have made will break this function. It might be easier to start again. It needs to be reworked this function as its a confusing mess. I've made some suggestions but they need to be thought through a bit more.

src/openrct2/world/Footpath.cpp
Gymnasiast: Hm, are you sure we shouldn't save the z along with it? How will it handle paths above each other?

duncanspumpkin: Its not part of the last path coordinate its a separate variable that is handled differently. Coupling it to last path would be confusing.

src/openrct2/world/Footpath.cpp
```diff
@@ -1279,39 +1273,39 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
...
-    int32_t x, int32_t y, int32_t z, int32_t direction, int32_t flags, int32_t level, int32_t distanceFromJunction,
-    int32_t junctionTolerance)
+    const CoordsXYZD& footpathPos, int32_t flags, int32_t level, int32_t distanceFromJunction, int32_t junctionTolerance)
 {
     TileElement* tileElement;
     int32_t edges, slopeDirection;
 
-    x += CoordsDirectionDelta[direction].x;
-    y += CoordsDirectionDelta[direction].y;
+    auto shiftedFootpathPos = CoordsXYZD{ CoordsXY{ footpathPos } + CoordsDirectionDelta[footpathPos.direction], footpathPos.z,
```
duncanspumpkin: This is not a good name. `targetPos` while still vague would be better.

src/openrct2/world/Footpath.cpp
```diff
@@ -1279,39 +1273,39 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
  *              (1 << 7): Ignore no entry signs
  */
 static int32_t footpath_is_connected_to_map_edge_recurse(
-    int32_t x, int32_t y, int32_t z, int32_t direction, int32_t flags, int32_t level, int32_t distanceFromJunction,
-    int32_t junctionTolerance)
+    const CoordsXYZD& footpathPos, int32_t flags, int32_t level, int32_t distanceFromJunction, int32_t junctionTolerance)
```
duncanspumpkin: I'm not convinced that direction should be part of footpathPos. It is used independently and represents the targetDirection and not the `footpathPos`

src/openrct2/world/Footpath.cpp
```diff
@@ -1396,26 +1392,27 @@ static int32_t footpath_is_connected_to_map_edge_recurse(
...
+        } while (get_next_direction(edges, &newDirection));
 
         return FOOTPATH_SEARCH_INCOMPLETE;
     }
 }
 
-int32_t footpath_is_connected_to_map_edge(int32_t x, int32_t y, int32_t z, int32_t direction, int32_t flags)
+int32_t footpath_is_connected_to_map_edge(const CoordsXYZD& footpathPos, int32_t flags)
 {
     flags |= (1 << 0);
```
duncanspumpkin: Future job will be to use the `GAME_COMMAND_FLAGS`

src/openrct2/world/Footpath.cpp
```diff
@@ -1912,45 +1909,44 @@ void footpath_update_queue_entrance_banner(int32_t x, int32_t y, TileElement* ti
  *
  *  rct2: 0x006A6B7F
  */
-static void footpath_remove_edges_towards_here(
-    int32_t x, int32_t y, int32_t z, int32_t direction, TileElement* tileElement, bool isQueue)
+static void footpath_remove_edges_towards_here(const CoordsXYZD& footpathPos, TileElement* tileElement, bool isQueue)
```
duncanspumpkin: Don't combine direction into the coordinate. Again this represents a direction not the footpathPos.

src/openrct2/world/Footpath.cpp
```diff
@@ -1912,45 +1909,44 @@ void footpath_update_queue_entrance_banner(int32_t x, int32_t y, TileElement* ti
...
 
     if (isQueue)
-        footpath_disconnect_queue_from_path(x, y, tileElement, -1);
+        footpath_disconnect_queue_from_path(footpathPos, tileElement, -1);
 
-    direction = (direction + 1) & 3;
-    x += CoordsDirectionDelta[direction].x;
-    y += CoordsDirectionDelta[direction].y;
+    Direction shiftedDirection = (footpathPos.direction + 1) & 3;
+    auto shiftedFootpathPos = CoordsXYZD{ CoordsXY{ footpathPos } + CoordsDirectionDelta[shiftedDirection], footpathPos.z,
```
duncanspumpkin: This is a `targetFootpathPos` and the D value is not part of the target

src/openrct2/world/Footpath.cpp
```diff
@@ -1912,45 +1909,44 @@ void footpath_update_queue_entrance_banner(int32_t x, int32_t y, TileElement* ti
...
-    tileElement = map_get_first_element_at({ x, y });
+    tileElement = map_get_first_element_at(shiftedFootpathPos);
     if (tileElement == nullptr)
         return;
     do
     {
         if (tileElement->GetType() != TILE_ELEMENT_TYPE_PATH)
             continue;
-        if (tileElement->base_height != z)
+        if (tileElement->GetBaseZ() != footpathPos.z)
```
duncanspumpkin: use the `targetFootpathPos.z`

src/openrct2/world/Footpath.cpp
tupaschoal: I've noted on my personal notes, do you want me to add a `TODO` as well?

duncanspumpkin: Yeah you could add a todo as it won't be for a while

src/openrct2/world/Footpath.cpp
tupaschoal: I'm guessing this is because the coords here should be `footpathPos + delta` for each of their directions, correct? This is what I've used.

If so, shouldn't all the places doing `currentPos += delta` actually be doing `currentPos = footpathPos + delta`?

src/openrct2/world/Footpath.cpp
tupaschoal: I've added a `TODO` to track it.

tupaschoal:
I've fixed all opened issues in their respective commits, for ease of reviewing. There is only one issue pending, which I need confirmation.

> For a lot of these functions direction can't be combined into the Coordinate as you will end up coupling two variables that are not related. As these functions are already confusing it will just make the issue worse.

There's none left.

> The changes you have made will break this function. It might be easier to start again. It needs to be reworked this function as its a confusing mess. I've made some suggestions but they need to be thought through a bit more.

This refers to `footpath_chain_ride_queue`. I've fixed all issues and moved it last in the commit order, in case we want to drop it.

src/openrct2/peep/Staff.cpp
```diff
@@ -727,10 +727,14 @@ static uint8_t staff_direction_surface(Peep* peep, uint8_t initialDirection)
 
         direction &= 3;
 
-        if (fence_in_the_way(peep->next_x, peep->next_y, peep->next_z, peep->next_z + 4, direction))
+        if (fence_in_the_way(
+                { peep->next_x, peep->next_y, peep->next_z * COORDS_Z_STEP, (peep->next_z * COORDS_Z_STEP) + PATH_HEIGHT },
```
duncanspumpkin: This isn't the path height it is the peep clearance height.

src/openrct2/peep/Staff.cpp
```diff
@@ -727,10 +727,14 @@ static uint8_t staff_direction_surface(Peep* peep, uint8_t initialDirection)
...
+                direction))
             continue;
 
-        if (fence_in_the_way(peep->next_x, peep->next_y, peep->next_z, peep->next_z + 4, direction_reverse(direction)))
+        if (fence_in_the_way(
+                { peep->next_x, peep->next_y, peep->next_z * COORDS_Z_STEP, (peep->next_z * COORDS_Z_STEP) + PATH_HEIGHT },
```
duncanspumpkin: as above

src/openrct2/world/Footpath.cpp
```diff
@@ -514,7 +514,7 @@ static TileElement* footpath_connect_corners_get_neighbour(int32_t x, int32_t y,
  *
  *  rct2: 0x006A70EB
  */
-static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileElement* initialTileElement)
+static void footpath_connect_corners(const CoordsXY& footpathPos, TileElement* initialTileElement)
```
duncanspumpkin: Can I be a pain and request this takes in a `PathElement*` instead of a `TileElement*`. It will reduce the AsPath() and mean the errors involved are higher up.

src/openrct2/world/Footpath.cpp
```diff
@@ -524,55 +524,50 @@ static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileEle
...
 
     tileElement[0] = initialTileElement;
-    int32_t z = initialTileElement->base_height;
+    int32_t z = initialTileElement->GetBaseZ();
     for (int32_t initialDirection = 0; initialDirection < 4; initialDirection++)
     {
-        int32_t x = initialX;
-        int32_t y = initialY;
-        int32_t direction = initialDirection;
+        auto currentPos = footpathPos;
```
duncanspumpkin: The logic here has been changed. currentPos should be footpathPos + CoordsDirectionDelta[direction]

src/openrct2/world/Footpath.cpp
```diff
@@ -524,55 +524,50 @@ static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileEle
         return;
 
     tileElement[0] = initialTileElement;
-    int32_t z = initialTileElement->base_height;
+    int32_t z = initialTileElement->GetBaseZ();
     for (int32_t initialDirection = 0; initialDirection < 4; initialDirection++)
     {
-        int32_t x = initialX;
-        int32_t y = initialY;
-        int32_t direction = initialDirection;
```
duncanspumpkin: I would keep this line in.

src/openrct2/world/Footpath.cpp
```diff
@@ -524,55 +524,50 @@ static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileEle
...
-        tileElement[1] = footpath_connect_corners_get_neighbour(x, y, z, (1 << direction_reverse(direction)));
+        tileElement[1] = footpath_connect_corners_get_neighbour({ currentPos, z }, (1 << direction_reverse(initialDirection)));
         if (tileElement[1] == nullptr)
             continue;
 
-        direction = (direction + 1) & 3;
-        x += CoordsDirectionDelta[direction].x;
-        y += CoordsDirectionDelta[direction].y;
-        tileElement[2] = footpath_connect_corners_get_neighbour(x, y, z, (1 << direction_reverse(direction)));
+        auto directionNext = (initialDirection + 1) & 3;
```
duncanspumpkin: I meant use the `direction_next()` function.

src/openrct2/world/Footpath.cpp
```diff
@@ -524,55 +524,50 @@ static void footpath_connect_corners(int32_t initialX, int32_t initialY, TileEle
...
+        tileElement[3] = footpath_connect_corners_get_neighbour({ currentPos, z }, (1 << ((directionNext + 1) & 3)));
         if (tileElement[3] == nullptr)
             continue;
 
-        direction = (direction + 1) & 3;
-        tileElement[3]->AsPath()->SetCorners(tileElement[3]->AsPath()->GetCorners() | (1 << (direction)));
-        map_invalidate_element({ x, y }, tileElement[3]);
+        directionNext = (directionNext + 1) & 3;
+        tileElement[3]->AsPath()->SetCorners(tileElement[3]->AsPath()->GetCorners() | (1 << (directionNext)));
+        map_invalidate_element(footpathPos + CoordsDirectionDelta[directionNext], tileElement[3]);
```
duncanspumpkin: This is too confusing to follow. Instead of recording just the tileElement in the array record the position as well using a CoordsXYE struct then you don't need to recalculate the position. (Note: direction will still need to be calculated as I can't at this moment follow if it stays the same)

src/openrct2/world/Footpath.cpp
```diff
@@ -696,48 +691,51 @@ static TileElement* footpath_get_element(int32_t x, int32_t y, int32_t z0, int32
...
-static bool footpath_reconnect_queue_to_path(int32_t x, int32_t y, TileElement* tileElement, int32_t action, int32_t direction)
+static bool footpath_reconnect_queue_to_path(
+    const CoordsXY& footpathPos, TileElement* tileElement, int32_t action, int32_t direction)
 {
     if (((tileElement->AsPath()->GetEdges() & (1 << direction)) == 0) ^ (action < 0))
         return false;
 
-    int32_t x1 = x + CoordsDirectionDelta[direction].x;
-    int32_t y1 = y + CoordsDirectionDelta[direction].y;
+    auto shiftedFootpathPos = footpathPos + CoordsDirectionDelta[direction];
```
duncanspumpkin: `targetQueuePos` would be a nicer name

src/openrct2/world/Footpath.cpp
```diff
@@ -696,48 +691,51 @@ static TileElement* footpath_get_element(int32_t x, int32_t y, int32_t z0, int32
...
-        if (fence_in_the_way(x1, y1, tileElement->base_height, tileElement->clearance_height, direction_reverse(direction)))
+        if (fence_in_the_way(
+                { shiftedFootpathPos, tileElement->GetBaseZ(), tileElement->GetClearanceZ() }, direction_reverse(direction)))
             return false;
     }
 
-    int32_t z = tileElement->base_height;
-    TileElement* otherTileElement = footpath_get_element(x1, y1, z - 2, z, direction);
+    int32_t z = tileElement->GetBaseZ();
+    TileElement* otherTileElement = footpath_get_element({ shiftedFootpathPos, z - LAND_HEIGHT_STEP, z }, direction);
```
duncanspumpkin: `targetFootpathElement`. I would then make a new variable called `targetQueueElement` after confirming it is not a nullptr and that it is a queue. Then we can get rid of all of the `AsPath()`

src/openrct2/world/Footpath.cpp
```diff
@@ -1049,34 +1050,34 @@ void footpath_connect_edges(int32_t x, int32_t y, TileElement* tileElement, int3
...
     lastQueuePathElement = nullptr;
-    int32_t z = tileElement->base_height;
     for (;;)
     {
         if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH)
         {
             lastPathElement = tileElement;
-            lastPathX = x;
-            lastPathY = y;
-            lastPathDirection = direction;
```
duncanspumpkin: I think you still need the `lastPathDirection`

src/openrct2/world/Footpath.cpp
```diff
@@ -1156,9 +1158,9 @@ void footpath_chain_ride_queue(
         if (lastPathElement->AsPath()->IsQueue())
         {
             lastPathElement->AsPath()->SetHasQueueBanner(true);
-            lastPathElement->AsPath()->SetQueueBannerDirection(lastPathDirection); // set the ride sign direction
+            lastPathElement->AsPath()->SetQueueBannerDirection(direction); // set the ride sign direction
```
duncanspumpkin: I think this needs to stay as `lastPathDirection`

src/openrct2/world/Footpath.cpp
tupaschoal: Argh, sure, my bad,

src/openrct2/world/Footpath.cpp
tupaschoal: I stored a pair `std:;pair<PathElement*, CoordsXY>`, because I had to start storing `PathElement*` to comply with a suggestion above and they don't inherit from each other (thus we have to add some `reinterpret_cast`)

src/openrct2/world/Footpath.cpp
tupaschoal: Done

src/openrct2/peep/Staff.cpp
tupaschoal: Add this constant to `Peep.h`

src/openrct2/world/Footpath.cpp
tupaschoal: @duncanspumpkin can you weigh on this? If it's not clear I can explain it better.

tupaschoal:
I've fixed all the new issues. Please give me feedback on the one still open: https://github.com/OpenRCT2/OpenRCT2/pull/10515#discussion_r363987668

src/openrct2/peep/Staff.cpp
duncanspumpkin: Yes

src/openrct2/world/Footpath.cpp
```diff
@@ -514,65 +515,65 @@ static TileElement* footpath_connect_corners_get_neighbour(int32_t x, int32_t y,
...
+        direction = direction_prev(direction);
+        tileElements[1].first->SetCorners(tileElements[1].first->GetCorners() | (1 << (direction)));
 
-        map_invalidate_element({ x, y }, tileElement[1]);
+        map_invalidate_element(tileElements[1].second, reinterpret_cast<TileElement*>(tileElements[1].first));
 
-        direction = initialDirection;
-        tileElement[0]->AsPath()->SetCorners(tileElement[0]->AsPath()->GetCorners() | (1 << (direction)));
-        map_invalidate_element({ x, y }, tileElement[0]);
+        direction = direction_prev(direction);
```
duncanspumpkin: Its the same value but can you make this the same as the original i.e. set it to initialDirection.

src/openrct2/Editor.cpp
```diff
@@ -500,12 +500,9 @@ namespace Editor
 
         for (const auto& parkEntrance : gParkEntrances)
         {
-            int32_t x = parkEntrance.x;
-            int32_t y = parkEntrance.y;
-            int32_t z = parkEntrance.z / 8;
             int32_t direction = direction_reverse(parkEntrance.direction);
 
-            switch (footpath_is_connected_to_map_edge(x, y, z, direction, 0))
+            switch (footpath_is_connected_to_map_edge(parkEntrance, direction, 0))
```
Gymnasiast: CoordsXYZD would be preferable here, I'd say.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -142,13 +142,13 @@ static int32_t guest_surface_path_finding(Peep* peep)
     int16_t z = peep->next_z;
```
Gymnasiast: Wouldn't it be better to just create a CoordsXYRangedZ and feed that into the functions? That should make the code easier to read.

Alternatively, making `z` use big coordinates would also clean up a lot already.

src/openrct2/Editor.cpp
duncanspumpkin: The direction represents where you want to check where as the xyz represents the footpath tile you are looking from. They shouldn't be tied together.

tupaschoal: Not changing anything then

src/openrct2/world/Footpath.cpp
tupaschoal: Solving based on discussions on gitter.

tupaschoal:
Should be alright now.

PR #10518, opened 2020-01-05
Fix line that causes build to fail on Mac

X123M3-256:
When trying to build the game on Mac, I got the following error:

```
/Users/ec2618/Programming/RCT2/OpenRCT2/src/openrct2/platform/Platform.macOS.mm:111:24: error: implicit conversion of NULL constant to 'uintptr_t' (aka 'unsigned long') [-Werror,-Wnull-conversion]
                return NULL;
                ~~~~~~ ^~~~
                       0
1 error generated.
```
I asked about it on the Gitter and was told to submit a PR.

### DIFF
--- a/src/openrct2/platform/Platform.macOS.mm
+++ b/src/openrct2/platform/Platform.macOS.mm
@@ -108,7 +108,7 @@ namespace Platform
         @autoreleasepool {
             if (input == NULL)
             {
-                return NULL;
+                return 0;
             }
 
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Regression from #10264.

tupaschoal:
Why didn't it fail on the CI? (I mean, the mac job failed, but it seems it was after it had already compiled)

AaronVanGeffen:
This is speculation, but I think fewer warnings are enabled in the Xcode project compared to the CMake project.

PR #10519, opened 2020-01-05
Fix #10313: Path furniture can be placed on level crossings

### LINKED ISSUE #10313: Path items such as benches can be placed on railroad grade crossings
**OS:** Windows 8
**Version:** 0.2.4
**Commit/Build:** 8378790

It's possible to place benches and other path items on a path tile that's part of a railroad's grade crossing. (See screenshot.) This doesn't really hurt anything, of course, but it looks really stupid. I think it would be better if a path tile with a railroad grade crossing on it were treated like a path tile that's a 4-way intersection: error message if you try to place something on it, and if you place something on it before building the crossing, that item will remain invisible and not really there until the crossing is removed.

Edit: the one exception might be for jumping fountains. Just like on a 4-way path intersection, those could still work on a grade crossing and still look fine.

**Steps to reproduce:**
1. Build path
2. Build miniature railroad across path
3. Build bench on the tile where the railroad crosses the path.

Fast Foods Factory 2019-11-30 18-13-23

Gymnasiast:
Specifically, this changes the following things:

- Building path furniture on level crossings is no longer allowed
- When building a miniature railway track through a path to create a level crossing, it will automatically remove any path furniture
- Lowering the land on a level crossing (which will cause handrails to be built) is no longer allowed

### DIFF
--- a/src/openrct2/actions/LandSetHeightAction.hpp
+++ b/src/openrct2/actions/LandSetHeightAction.hpp
@@ -105,6 +105,15 @@ public:
         if (surfaceElement == nullptr)
             return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);
 
+        // We need to check if there is _currently_ a level crossing on the tile. For that, we need the old height,
+        // so we can't use the _height variable.
+        auto oldCoords = CoordsXYZ{ _coords, surfaceElement->GetBaseZ() };
+        auto* pathElement = map_get_footpath_element(oldCoords);
+        if (pathElement != nullptr && pathElement->AsPath()->IsLevelCrossing(oldCoords))
+        {
+            return MakeResult(GA_ERROR::DISALLOWED, STR_REMOVE_LEVEL_CROSSING_FIRST);
+        }
+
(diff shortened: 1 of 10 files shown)


src/openrct2/actions/LandSetHeightAction.hpp
```diff
@@ -105,6 +105,15 @@ DEFINE_GAME_ACTION(LandSetHeightAction, GAME_COMMAND_SET_LAND_HEIGHT, GameAction
         if (surfaceElement == nullptr)
             return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);
 
+        // We need to check if there is _currently_ a level crossing on the tile. For that, we need the old height,
+        // so we can't use the _height variable.
```
AaronVanGeffen: This is very minor, but this would probably flow better with each sentence on their own line.

AaronVanGeffen:
It's nice to see the fruits of your refactoring work. These additions were likely made much easier by it.

Gymnasiast:
Yes, but not just my refactoring work: the game action refactor also helped a lot.

PR #10521, opened 2020-01-05
Find Jansson

### LINKED ISSUE #10355: CMake Mac Build failes for MacOS 10.14+
Mac OS Catalina 10.15.1 (Suspected since 10.14+)
Current Master Branch / 0.2.4
commit 432ad7867a76363534131f478b7d58247a40e6b8

MacOS Xcode linker doesn´t find  libraries like jansson.
Linker failes with " library not found"

**Dump file**
 ```
    [ 72%] Built target libopenrct2
    Scanning dependencies of target openrct2-cli
    Scanning dependencies of target openrct2
    [ 72%] Building CXX object CMakeFiles/openrct2-cli.dir/src/openrct2-cli/Cli.cpp.o
    [ 73%] Linking CXX executable openrct2-cli
    ld: library not found for -ljansson
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
    make[2]: *** [openrct2-cli] Error 1
    make[1]: *** [CMakeFiles/openrct2-cli.dir/all] Error 2
    make[1]: *** Waiting for unfinished jobs....
 ```

Referenced StackOverflow Question:
https://stackoverflow.com/questions/54068035/linking-not-working-in-homebrews-cmake-since-mojave

The new XCode Linker doesnt use the standard lib path /usr/local/lib where the Homebrew libs are installed. Instead it seems like the Linker searches in the path relative to the Project directory.
Like : $PWD//usr/local/lib 
This could be due to using the -issysroot option.

Tested working hotfix would be adding following lines somewhere in the head of CMakeLists.txt:
 ```
IF(APPLE)
    LINK_DIRECTORIES(/usr/local/lib)
ENDIF()
 ```

Building after change:
 ```
    [ 72%] Built target libopenrct2
    [ 73%] Linking CXX executable openrct2-cli
    [ 73%] Linking CXX executable openrct2
    [ 73%] Built target openrct2-cli
    [100%] Built target openrct2
    Build finished
 ```

this fixes the issue with the standard libary path not searched due to the before mentioned stackoverflow thread.

Would be nice to discuss this Issue since the "Fix" seems rather hacky.

tomlankhorst:
Find Jansson with pkg-config on non-MSVC. 
Fix #10355


src/openrct2-ui/CMakeLists.txt
```diff
@@ -9,10 +9,10 @@ option(DISABLE_OPENGL "Disable OpenGL support.")
 
 # Third party libraries
 if (MSVC)
-    find_package(jansson REQUIRED)
```
AaronVanGeffen: Why remove the dependency for the MSVC target?

tomlankhorst: Sorry, I moved from a FindFile.cmake to a PKG way. Will update this.

tomlankhorst:
> The `FindJansson.cmake` file mentioned has not been included in this PR.

This is fixed. I use pkg config now.

PR #10522, opened 2020-01-05
Link stdc++fs with GCC

### LINKED ISSUE #10352: gcc requires linking stdc++fs
**OS:** Debian buster x86_64 (10.2)
**Version:** develop
**Commit/Build:** e72d2aba851da17629316c4cb83efce57c32a264
**gcc:** (Debian 8.3.0-6) 8.3.0
**cmake:** 3.13.4

As mentioned in pull request #10261, older versions of gcc require linking `stdc++fs`. I do a personal build of OpenRCT2 about once a month, so I've just hit this after pulling current develop. As-is, anyone using a version of gcc older than 9.0 can't compile the project. This is potentially a lot of people, as for example Debian buster ships with gcc 8.3.

Locally I've just added a `target_link_libraries` line to the CMakeLists.txt and I'm able to build again, but a better solution is needed.

**Steps to reproduce:**
1. Checkout current develop
2. Build with gcc < 9

**Error**

```
[ 73%] Linking CXX executable openrct2-cli
/usr/bin/ld: libopenrct2.a(Path.cpp.o): in function `std::filesystem::__cxx11::path::path(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&&, std::filesystem::__cxx11::path::format)':
Path.cpp:(.text._ZNSt10filesystem7__cxx114pathC2EONSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS1_6formatE[_ZNSt10filesystem7__cxx114pathC5EONSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS1_6formatE]+0x4f): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/bin/ld: libopenrct2.a(Path.cpp.o): in function `std::filesystem::__cxx11::path::extension() const':
Path.cpp:(.text._ZNKSt10filesystem7__cxx114path9extensionEv[_ZNKSt10filesystem7__cxx114path9extensionEv]+0x19): undefined reference to `std::filesystem::__cxx11::path::_M_find_extension() const'
/usr/bin/ld: libopenrct2.a(Path.cpp.o): in function `std::filesystem::__cxx11::path::path<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::filesystem::__cxx11::path>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::filesystem::__cxx11::path::format)':
Path.cpp:(.text._ZNSt10filesystem7__cxx114pathC2INSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES1_EERKT_NS1_6formatE[_ZNSt10filesystem7__cxx114pathC5INSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES1_EERKT_NS1_6formatE]+0x64): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/openrct2-cli.dir/build.make:88: openrct2-cli] Error 1
make[1]: *** [CMakeFiles/Makefile2:142: CMakeFiles/openrct2-cli.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

**Quick fix**
```
gibmat@aule:~/OpenRCT2$ git diff
diff --git a/src/openrct2/CMakeLists.txt b/src/openrct2/CMakeLists.txt
index 97674911c..a5b24aaea 100644
--- a/src/openrct2/CMakeLists.txt
+++ b/src/openrct2/CMakeLists.txt
@@ -217,3 +217,5 @@ else ()
     # Dummy target to ease invocation
     add_custom_target(${PROJECT_NAME}-headers-check)
 endif ()
+
+target_link_libraries(libopenrct2 stdc++fs)
```

tomlankhorst:
Fix #10352

### DIFF
--- a/src/openrct2/CMakeLists.txt
+++ b/src/openrct2/CMakeLists.txt
@@ -18,7 +18,7 @@ set_target_properties(${PROJECT_NAME} PROPERTIES PREFIX "")
 SET_CHECK_CXX_FLAGS(${PROJECT_NAME})
 
 # GCC likes us to pass the -lstdc++fs flag to link C++17 filesystem implementation.
-if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT MINGW)
     target_link_libraries(${PROJECT_NAME} stdc++fs)
 endif()
 
--- a/src/openrct2/CMakeLists.txt
+++ b/src/openrct2/CMakeLists.txt
@@ -17,6 +17,11 @@ add_library(${PROJECT_NAME} ${OPENRCT2_CORE_SOURCES} ${OPENRCT2_CORE_MM_SOURCES}
 set_target_properties(${PROJECT_NAME} PROPERTIES PREFIX "")
 SET_CHECK_CXX_FLAGS(${PROJECT_NAME})
 
+# GCC likes us to pass the -lstdc++fs flag to link C++17 filesystem implementation.
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    target_link_libraries(${PROJECT_NAME} stdc++fs)
+endif()
+
 if (NOT DISABLE_NETWORK OR NOT DISABLE_HTTP)
     if (APPLE)
         # Needed for linking with non-broken OpenSSL on Apple platforms


Gymnasiast:
It looks like MinGW doesn't like this.

src/openrct2/CMakeLists.txt
```diff
@@ -17,6 +17,11 @@ add_library(${PROJECT_NAME} ${OPENRCT2_CORE_SOURCES} ${OPENRCT2_CORE_MM_SOURCES}
 set_target_properties(${PROJECT_NAME} PROPERTIES PREFIX "")
 SET_CHECK_CXX_FLAGS(${PROJECT_NAME})
 
+# GCC likes us to pass the -lstdc++fs flag to link C++17 filesystem implementation.
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
janisozaur:
```suggestion
if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT MINGW)
```

janisozaur: Ping @tomlankhorst . I would commit this myself, but have no rights to do so

tomlankhorst: Yes :) Will do now.

src/openrct2/CMakeLists.txt
```diff
@@ -17,6 +17,11 @@ add_library(${PROJECT_NAME} ${OPENRCT2_CORE_SOURCES} ${OPENRCT2_CORE_MM_SOURCES}
 set_target_properties(${PROJECT_NAME} PROPERTIES PREFIX "")
 SET_CHECK_CXX_FLAGS(${PROJECT_NAME})
 
+# GCC likes us to pass the -lstdc++fs flag to link C++17 filesystem implementation.
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT MINGW)
```
chfast: This is also sometimes needed when Clang is used on linux. By default clang is using libstdc++ (the same as GCC) on linux and it also needs this flag. In CMake, there seems not to be an easy way to check what standard library is being used. I usually put this condition as `if(UNIT NOT APPLE)`.

xbjfk:
Hey, actually this breaks compilation on my system that doesnt have gcc/libstdc++ installed at all, and everything is purely compiled with clang/libcxx. Maybe instead you should check whether libstdc++ is being used rather than adding this for both clang and gcc? I think this might also have some bad effects when libcxx is default and manually specifying a libstdc++ library 
Thanks.

PR #10408, opened 2019-12-21
Make Map::map_get_banner_element_at() use CoordsXYZ

tupaschoal:

PR #10528, opened 2020-01-06
Fix #10524: Banners not placeable above certain height

### LINKED ISSUE #10524: Cannot place banners on paths above a certain height
**OS:** Windows 10
**Version:** v0.2.4-365
**Commit/Build:** 08ab15e

When trying to build a banner on a path above a certain height, the banner will not place and only a ghost piece is visible. The banner-window very briefly opens and then closes again. The ghost-piece does not remain when saving and reloading a park. You **do** lose money when trying to build the banner.

**Steps to reproduce:**
1. Place a path at at least +8 units
2. Try place a banner on the path

**Screenshots / Video:**

banners

tupaschoal:
Closes #10524 

This fixes an overflow introduced by #10408 with the move to BigZ coords.


duncanspumpkin:
Can you bump the network version

duncanspumpkin:
Thought this needed a network bump but it doesn't.

PR #10536, opened 2020-01-08
Add Cycle Detection to Ride Rating Calculation

bitman2049:
Hacked rides no longer block the ride rating calculation function.

Issue link: https://github.com/OpenRCT2/OpenRCT2/issues/10508

### DIFF
--- a/data/language/fr-FR.txt
+++ b/data/language/fr-FR.txt
@@ -116,10 +116,10 @@ STR_0530    :Des voitures suspendues à un câble métallique font des allers-re
 STR_0531    :Montagnes russes compactes dont les voies en acier prennent la forme de tire-bouchons et de boucles
 STR_0532    :Labyrinthe construit à partie de haies ou de murs de 2m de haut duquel les visiteurs ne peuvent sortir que lorsqu'ils en trouvent la sortie
 STR_0533    :Structure en bois constitué d'un escalier intérieur et d'un toboggan extérieur en spirale utilisable avec des tapis de glisse
-STR_0534    :Karts individuels constitués de moteurs à essence
+STR_0534    :Les visiteurs font la course dans des karts sur un terrain en bitume
 STR_0535    :Bateaux en forme de bûches géantes parcourant un tube aquatique dont les pentes abruptes auront pour effet de tremper les passagers
 STR_0536    :Des bateaux de forme circulaire serpentent le long d'un large cours d'eau, puis dévalent des rapides et des chutes d'eau spectaculaires
-STR_0537    :Auto tamponneuses électriques en libre service
+STR_0537    :Les visiteurs se rentrent dedans dans des auto tamponneuses électriques
 STR_0538    :Grand bateau pirate à balancier
 STR_0539    :Bateau attaché à un bras avec contrepoids à l'autre extrémité, capable de se balancer en effectuant une rotation à 360°
 STR_0540    :Échoppe vendant de la nourriture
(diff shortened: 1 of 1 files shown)


ZehMatt:
I dislike the idea of influencing the game state upon player connect/disconnect, the main reason for this is that the more components we need to synchronize the harder it becomes to maintain the already complex multiplayer aspect, given that peeps will have different decisions on what to ride based on the ride rating it would be also required to match the synchronization tick perfect as in, the simulation requires the exact same data at a specific tick/mid-tick for every client. Maybe @IntelOrca has some thoughts on this.

IntelOrca:
Well the only other thing that can be done is to add it to the end of the map stream that is sent alongside other things like cheats.

bitman2049:
> Well the only other thing that can be done is to add it to the end of the map stream that is sent alongside other things like cheats.

I think this is a better option. The struct would be an additional 8 bytes of info sent to joining players, and once it's initialized it should update the same for all players.

duncanspumpkin:
Surely this doesn't need additional information. Why can't the algorithm just know to skip the bad rides.

IntelOrca:
@duncanspumpkin it needs more state to work out that the ride is bad. The discussion is about synchronising that state across all clients in network games.

duncanspumpkin:
But why does it need more state.

bitman2049:
The state is the slower iterator for the tortoise and hare algorithm. The "bad" rides can switch from bad to good in a single game tick if you use the tile inspector to break the loop, so it can't just skip checking those rides.

duncanspumpkin:
But cant we just do a quick check of the ride to see if it will encounter this issue (i.e. tortoise and hare each ride on every update) rather than keeping a tortoise and hare of the ride ratings. That way it avoids the additional state.

IntelOrca:
Isn't that expensive? You could do the whole ride update in 1 tick if you wanted but presumebly it is spread over multiple ticks to avoid slowing down the game in heavy parks.

duncanspumpkin:
I think the slow bit is the calculation of the bonuses near a ride piece. You don't need to do that calculation just do the look for loops.

bitman2049:
But that's the calculation that kept getting stuck in cycles. If it only checked when switching from one ride to the next, then games loaded with the calculator stuck in a loop would still have it stuck. This makes the check happen while traversing the track for prox bonuses, so it also fixes it when loading from a saved game.

Additionally, this fix doesn't just skip to the next ride when it detects a cycle. It treats a cycle the same way the end of track is treated in non-circuit tracks. It does result in double-counting some elements for proximity bonuses, but it also means that coasters like [this](https://imgur.com/uvNAFyp) can be tested and ridden by guests.

duncanspumpkin:
The ratings calc keeps track of where on the track it is in calculating bonuses. This is the part you have added a second pointer that is a tortoise. From my understanding the expensive bit of the rating calc is checking for proximity bonuses next to this track piece.

What I suggest we do is have a loop checker that runs every tic update. It would not perform the expensive proximity bonus checks. This would inform you that the track has a loop. It could even return a pointer to where the tortoise and hare had failed. You could then calculate bonuses upto that loop failure pointer and count that as the end of the track (like a non-circuited track).

Would that work? I think it might, then it wouldn't involve anything complex with adding more state. There would be a minor performance hit but I doubt that would be noticeable. Should be checked on a ride with very long track.

bitman2049:
That could work, I'd need to try it out. I'm not as experienced with the performance of a raw traversal, but I know the game does it whenever you test the track or go into construction. For really huge coasters (like edge cases, i.e. LordMarcel's record breakers), does that traversal cause lag?

duncanspumpkin:
Its hard to say without testing. Its never going to be free to do but I don't imagine it will be much worse.

bitman2049:
Are there perf tests? Or a ticket for making them? I'd be willing to jump in and make some if they're needed/wanted.

duncanspumpkin:
We have a benchmark command @ZehMatt would know more about it than me.

ZehMatt:
Unfortunately it only benchmarks drawing, there is currently no benchmark for the simulation.

AaronVanGeffen:
For what it's worth, #10281 comes to mind.

bitman2049:
Messed my forks up on github. I'll get back to this issue later.

Gymnasiast:
@bitman2049 Make sure you create a separate branch for your feature and don't push to your develop branch again. That tends to cause headaches.

bitman2049:
Yeah, thank you. I'm taking this as a learning experience. I didn't lose my changes, they're just on another branch now. And they're not ready for merge, I still need to do some testing.

PR #10540, opened 2020-01-09
Fix #10496: Game can't handle path with non-english string

### LINKED ISSUE #10496: Game can't handle path with non-english string
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** dcdd312

It seems that game can't handle path with non-english string, such as Korean.
If the name of Windows account is written in Korean, I can't run the game. If the account name is english, game is run correctly.
So I concluded that recent path refactoring causes this issue, so maybe it is related to #10283, too.
I've tested this at Windows 10 32bit on VMware.

**Steps to reproduce:**
1. Name the Windows account with a non-english characters, in my case, Korean: ```테스트```(means "test")
2. Check that user directory name is non-english. (eg) ```C:\Users\테스트```
3. Try to run the game
4. It shows a white windows few seconds, then it is terminated forcely.

**Screenshots / Video:**
When I try to run the game with Windows account name ```테스트```:
asdf

Directories in ```C:\Users```. ```TELK-win10x86``` works, and ```테스트``` doesn't:
image

ps. Happy New Year!

IntelOrca:


IntelOrca:
There is a problem with export objects, the code treats object names as UTF-8 but some are not legal UTF-8 strings so we will probably have to assume CP-1252 and convert to UTF-8 string when constructing a new filename for the object.

Gymnasiast:
@IntelOrca I don't think I have ever seen an object with an ID that consisted of non-ASCII characters. Do you have one for me that I can use to test?

IntelOrca:
There is an object embedded in the SV6 export import test. The name looks corrupt to me but essentially it has characters in the name above char 127.

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -575,40 +573,45 @@ class ObjectRepository final : public IObjectRepository
...
+        return fullPath;
+    }
+
+    std::string GetFileNameForNewObject(const std::string_view& name)
     {
+        // Trim name
         char normalisedName[9] = { 0 };
-        for (int32_t i = 0; i < 8; i++)
+        auto maxLength = std::min<size_t>(name.size(), 8);
+        for (int32_t i = 0; i < maxLength; i++)
```
Gymnasiast:
```suggestion
        for (size_t i = 0; i < maxLength; i++)
```

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -575,40 +573,45 @@ class ObjectRepository final : public IObjectRepository
...
         {
             if (name[i] != ' ')
             {
-                normalisedName[i] = toupper(name[i]);
+                normalisedName[i] = std::toupper(name[i]);
```
Gymnasiast: Seems to fail CI.

PR #10544, opened 2020-01-11
Fix #7784: Vehicle tab takes 1st car colour instead of tab_vehicle's colour

### LINKED ISSUE #7784: vehicles tab takes 1st car color instead of tab_vehicle's color
**OS:** Windows 10
**Version:** 0.2.0
**Commit/Build:** 0aff800

image

When painting Different Colors per Vehicle, rides with a nonzero `tab_vehicle` choose the colors from the 0th vehicle instead of the `tab_vehicle`th vehicle. This means the vehicles tab vehicle can be painted a different color than the train on the map.

**Steps to reproduce:**
1. Build a car ride, mini helicopters, or steel wild mouse
2. Change the color scheme to 'different colors per vehicle'
3. Change the color of the invisible 1st vehicle
4. The vehicle tab will show the wrong color

Gymnasiast:


duncanspumpkin:
Have you checked to make sure this doesn't cause desyncs

Gymnasiast:
@duncanspumpkin I haven't, no. But the `ride_get_vehicle_colour()` function is only called by the ride window and no actual game state code.

src/openrct2/ride/Ride.cpp
```diff
@@ -3108,8 +3108,16 @@ vehicle_colour ride_get_vehicle_colour(Ride* ride, int32_t vehicleIndex)
 {
     vehicle_colour result;
 
-    // Prevent indexing array out of bounds
-    vehicleIndex = std::min(vehicleIndex, MAX_CARS_PER_TRAIN);
+    if (ride->colour_scheme_type == VEHICLE_COLOUR_SCHEME_PER_VEHICLE)
```
duncanspumpkin: This is wrong. It should be a special case for when VEHICLE_COLOUR_SCHEME_SAME not for PER_VEHICLE

PR #10562, opened 2020-01-13
Add assertion failure reason to the Backtrace service integration

ZehMatt:
When I was doing #10175 I was hoping that I will find the string in the minidump on the stack memory but it seems that this assumption did not work out. This PR will attach a new parameter for the Backtrace service adding the last assertion message.

### DIFF
--- a/src/openrct2/core/Guard.cpp
+++ b/src/openrct2/core/Guard.cpp
@@ -45,6 +45,8 @@ namespace Guard
 #endif
         ;
 
+    static std::optional<std::string> _lastAssertMessage = std::nullopt;
+
 #ifdef _WIN32
     static void GetAssertMessage(char* buffer, size_t bufferSize, const char* formattedMessage);
     static void ForceCrash();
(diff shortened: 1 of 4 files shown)


IntelOrca:
Another thing that isn't sent is the message of a C++ exception when one is thrown but not caught. Is that possible to add?

ZehMatt:
Not sure if there is a easy way, only thing I can think of is having an try/catch block at the entrypoint and store the message somewhere before re-throwing.

IntelOrca:
@ZehMatt yeah, I am happy with this as it is anyways

ZehMatt:
While we are at it, at some point I want to have actual log files per instance and instead of writing to stdout/stderr, then we would be able to attach those logs for way better information.

IntelOrca:
@ZehMatt yes, I would assume you would implement this alongside your better logging interface.

ZehMatt:
If I ever get to finish it, yep.

PR #10567, opened 2020-01-13
Fix peep choosing to path to shops with NO_MONEY mode

duncanspumpkin:


src/openrct2/peep/Guest.cpp
```diff
@@ -2272,7 +2272,7 @@ bool Guest::ShouldGoToShop(Ride* ride, bool peepAtShop)
 
         // The amount that peeps are willing to pay to use the Toilets scales with their bathroom stat.
         // It effectively has a minimum of $0.10 (due to the check above) and a maximum of $0.60.
-        if (ride->price * 40 > toilet)
+        if (!(gParkFlags & PARK_FLAGS_NO_MONEY) && ride->price * 40 > toilet)
```
Gymnasiast: We have the `ride_get_price()` function, which checks this for you. Using it would make the code a bit cleaner.

duncanspumpkin: Argh i thought there was something like that

src/openrct2/peep/Guest.cpp
```diff
@@ -2298,7 +2298,7 @@ bool Guest::ShouldGoToShop(Ride* ride, bool peepAtShop)
     }
 
     // Basic price checks
-    if (ride->price != 0 && ride->price > cash_in_pocket)
+    if (ride_get_price(ride) != 0 && ride_get_price(ride) > cash_in_pocket)
```
Gymnasiast: Perhaps better to save the result in a variable, rather than calling this twice.

PR #10568, opened 2020-01-13
Android: target SDK 28, remove absolute path

NuclearFej:
Android version: an absolute path was replaced with a call to the Environment class. The target SDK of 29 was decremented to a target SDK of 28 to fix an incompatibility with an SDK change. Game now builds; however a desktop version must be built first so that build artifacts are present in /bin.

SDK 29 doesn't like it when you write files directly to /sdcard, they're deprecating access to shared storage - which is insane, but out of our control. All of the code that touches the filesystem will have to be rewritten eventually to deal with this. In the meantime, builds compiled to SDK 28 or lower are still allowed to touch /sdcard.

Builds and runs fine (with deprecation warnings which I enabled in an earlier PR), tested working on a Pixel 2 XL on Android 10/Q. Again, a desktop version must be built prior to building the Android build at this time so that data is present in OpenRCT2/bin/data which is copied into the Android assets. This will be fixed eventually but for the time being I would like to focus on getting the game to a playable state without mouse/keyboard, if that's okay.

### DIFF
--- a/src/openrct2-android/app/build.gradle
+++ b/src/openrct2-android/app/build.gradle
@@ -7,7 +7,7 @@ android {
     defaultConfig {
         applicationId 'io.openrct2'
         minSdkVersion 19
-        targetSdkVersion 29
+        targetSdkVersion 28
 
         versionCode 2
         versionName '0.2.5'
--- a/src/openrct2-android/app/src/main/java/io/openrct2/MainActivity.java
+++ b/src/openrct2-android/app/src/main/java/io/openrct2/MainActivity.java
@@ -11,6 +11,8 @@ import android.os.Bundle;
 import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
 import androidx.appcompat.app.AppCompatActivity;
+
+import android.os.Environment;
 import android.text.TextUtils;
 import android.util.DisplayMetrics;
 import android.util.Log;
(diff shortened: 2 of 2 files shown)


IntelOrca:
With GitHub actions it is possible to create a pipeline where we can grab g2.dat from another CI, alternatively we build a cut down version for x86-64 that can generate the resources.

What do they recommending doing instead of using /sdcard?

Good work by the way!

NuclearFej:
Thanks!

Here's the api doc on the deprecation: https://developer.android.com/reference/android/os/Environment#getExternalStorageDirectory()

I don't do much Android development, but here's what I've been able to glean - 

There is no straight replacement. For the /bin/data files, the use of https://developer.android.com/reference/android/content/Context.html#getExternalFilesDir(java.lang.String) would probably be fine as the user needs not see or edit them. They would be removed on uninstallation. Accessing the rct2 files is a bit harder, we could use the storage access framework to access the rct2 files on first run and copy them to the directory returned by getExternalFilesDir, but the user has to select that folder on first run using the system file picker. Not the end of the world and we probably should let the user pick the rct2 folder instead of hardcoding it anyway. User would be informed that they can delete the original files they copied if so desired.

Game saves are the worst part. If we use getExternalFilesDir, they are very hard to find (to copy to a PC - would be in something like /sdcard/android/data/io.openrct2) and are destroyed on uninstallation. If we use the storage access framework, we have to skip the in-game file picker and use the system file picker which could be confusing. (Do note that, at present, the "use system file picker" button in the game causes the app to crash instantly.) We _can_ have the user choose a directory instead of a file so once they pick one I _think_ we can access it until the app is closed but *don't quote me on that.*

I don't think there are any other options, maybe MediaStore but it doesn't seem meant for this use case.

IntelOrca:
Hmm okay, there are a few options then. Might be an idea to see what other games to for these sort of scenarios. I think using the system file picker is better anyway and gives players more choice on where their data is.

PR #10570, opened 2020-01-14
Prefer CoordsXY* on more places

tupaschoal:


src/openrct2/world/Banner.cpp
```diff
@@ -93,7 +93,7 @@ static uint8_t banner_get_ride_index_at(int32_t x, int32_t y, int32_t z)
         if (ride == nullptr || ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_IS_SHOP))
             continue;
 
-        if ((tileElement->GetClearanceZ()) + 32 <= z)
+        if ((tileElement->GetClearanceZ()) + COORDS_XY_STEP <= bannerCoords.z)
```
Gymnasiast: Z is not X or Y, so don't use COORDS_XY_STEP.

```suggestion
        if ((tileElement->GetClearanceZ()) + (4 * COORDS_Z_STEP) <= bannerCoords.z)
```

src/openrct2/world/Banner.cpp
```diff
@@ -209,15 +209,21 @@ WallElement* banner_get_scrolling_wall_tile_element(BannerIndex bannerIndex)
...
  */
-uint8_t banner_get_closest_ride_index(int32_t x, int32_t y, int32_t z)
+uint8_t banner_get_closest_ride_index(const CoordsXYZ& mapPos)
 {
-    static constexpr const LocationXY16 NeighbourCheckOrder[] = { { 32, 0 },    { -32, 0 },   { 0, 32 },
-                                                                  { 0, -32 },   { -32, +32 }, { +32, -32 },
-                                                                  { +32, +32 }, { -32, +32 }, { 0, 0 } };
-
-    for (size_t i = 0; i < std::size(NeighbourCheckOrder); i++)
+    static constexpr const std::array<CoordsXY, 9> NeighbourCheckOrder = { CoordsXY{ COORDS_XY_STEP, 0 },
```
Gymnasiast: Why an std::array, rather than keeping it as a C array? To me, it seems like the latter is more efficient.

Also, _if_ we decide to stick with std::array, then you should also modify its counterpart in TestPaint.

duncanspumpkin: Std::array let's us use the ranged for loop and is no more or less efficient.

tupaschoal: `std::array` is just a wrapper for C array that is STL compliant and provides convenient functions and usage, such as the one mentioned by @duncanspumpkin

src/openrct2/world/Banner.cpp
tupaschoal: Argh, my bad. Fixed it!

src/openrct2/world/Banner.cpp
Gymnasiast: Unresolving this - you haven't modified its counterpart in Testpaint yet, which is why the CI is failing.

tupaschoal:
Done

test/testpaint/Compat.cpp
```diff
@@ -38,15 +38,15 @@ uint32_t gScenarioTicks;
...
-const CoordsXY CoordsDirectionDelta[] = {
-    { -32, 0 },
-    { 0, +32 },
-    { +32, 0 },
-    { 0, -32 },
-    { -32, +32 },
-    { +32, +32 },
-    { +32, -32 },
-    { -32, -32 },
+static constexpr const std::array<CoordsXY, 8> CoordsXY CoordsDirectionDelta = {
```
Gymnasiast:
```suggestion
static constexpr const std::array<CoordsXY, 8> CoordsDirectionDelta = {
```

test/testpaint/Compat.cpp
```diff
@@ -38,15 +38,15 @@ uint32_t gScenarioTicks;
...
-const CoordsXY CoordsDirectionDelta[] = {
-    { -32, 0 },
-    { 0, +32 },
-    { +32, 0 },
-    { 0, -32 },
-    { -32, +32 },
-    { +32, +32 },
-    { +32, -32 },
-    { -32, -32 },
+static constexpr const std::array<CoordsXY, 8> CoordsDirectionDelta = {
```
Gymnasiast:
```suggestion
constexpr const std::array<CoordsXY, 8> CoordsDirectionDelta = {
```

It can't be static, as that violates the declaration in `TileElement.h`.

tupaschoal: Correct, changed it,

src/openrct2/world/Map.cpp
```diff
@@ -63,8 +63,14 @@ using namespace OpenRCT2;
 /**
  * Replaces 0x00993CCC, 0x00993CCE
  */
-const CoordsXY CoordsDirectionDelta[] = { { -32, 0 },   { 0, +32 },   { +32, 0 },   { 0, -32 },
-                                          { -32, +32 }, { +32, +32 }, { +32, -32 }, { -32, -32 } };
+const std::array<CoordsXY, 8> CoordsDirectionDelta = { CoordsXY{ -COORDS_XY_STEP, 0 },
+                                                       CoordsXY{ 0, +COORDS_XY_STEP },
```
Gymnasiast: Can you add a line break after `CoordsDirectionDelta = {` and indent the contents _one_ step (i.e. 4 spaces)? Makes for a more legible block.

test/testpaint/Compat.cpp
```diff
@@ -38,15 +38,15 @@ uint32_t gScenarioTicks;
...
-    { -32, 0 },
-    { 0, +32 },
-    { +32, 0 },
-    { 0, -32 },
-    { -32, +32 },
-    { +32, +32 },
-    { +32, -32 },
-    { -32, -32 },
+constexpr const std::array<CoordsXY, 8> CoordsDirectionDelta = {
+    CoordsXY{ -32, 0 },
```
Gymnasiast: Needs COORDS_XY_STEP.

src/openrct2/world/Map.cpp
tupaschoal: Done, but that required turning off clang-format for that piece of code.

src/openrct2/world/Banner.cpp
duncanspumpkin: This is just a fancy array rearranged of CoordsDirectionDelta (Except it also checks tile 0). It doesn't actually matter the order that much. Is there a way we could define this as CoordsDirectionDelta + 1 coords for the tile 0?

src/openrct2/world/Duck.cpp
```diff
@@ -279,12 +278,11 @@ void rct_duck::UpdateFlyAway()
         Invalidate();
 
         int32_t direction = sprite_direction >> 3;
-        int32_t newX = x + (DuckMoveOffset[direction].x * 2);
-        int32_t newY = y + (DuckMoveOffset[direction].y * 2);
-        int32_t newZ = std::min(z + 2, 496);
-        if (map_is_location_valid({ newX, newY }))
+        auto destination = CoordsXYZ{ x + (DuckMoveOffset[direction].x * 2), y + (DuckMoveOffset[direction].y * 2),
+                                      static_cast<int32_t>(std::min(z + 2, 496)) };
```
duncanspumpkin: Can you remove the cast if you specify the type in the `std::min<int32_t>`

src/openrct2/world/Entrance.cpp
```diff
@@ -128,15 +128,14 @@ money32 ride_entrance_exit_place_ghost(
...
-void maze_entrance_hedge_replacement(int32_t x, int32_t y, TileElement* tileElement)
+void maze_entrance_hedge_replacement(const CoordsXYE& entrance)
 {
-    int32_t direction = tileElement->GetDirection();
-    x += CoordsDirectionDelta[direction].x;
-    y += CoordsDirectionDelta[direction].y;
-    int32_t z = tileElement->GetBaseZ();
-    ride_id_t rideIndex = tileElement->AsEntrance()->GetRideIndex();
+    int32_t direction = entrance.element->GetDirection();
+    auto targetPos = entrance + CoordsDirectionDelta[direction];
```
duncanspumpkin: `hedgePos`

src/openrct2/world/Entrance.cpp
```diff
@@ -166,15 +165,14 @@ void maze_entrance_hedge_replacement(int32_t x, int32_t y, TileElement* tileElem
...
-void maze_entrance_hedge_removal(int32_t x, int32_t y, TileElement* tileElement)
+void maze_entrance_hedge_removal(const CoordsXYE& entrance)
 {
-    int32_t direction = tileElement->GetDirection();
-    x += CoordsDirectionDelta[direction].x;
-    y += CoordsDirectionDelta[direction].y;
-    int32_t z = tileElement->GetBaseZ();
-    ride_id_t rideIndex = tileElement->AsEntrance()->GetRideIndex();
+    int32_t direction = entrance.element->GetDirection();
+    auto targetPos = entrance + CoordsDirectionDelta[direction];
```
duncanspumpkin: `hedgePos`

src/openrct2/world/Fountain.cpp
```diff
@@ -38,12 +38,25 @@ namespace FOUNTAIN_FLAG
     const uint32_t DIRECTION = 1 << 7;
 }; // namespace FOUNTAIN_FLAG
 
-static constexpr const CoordsXY _fountainDirectionsNegative[] = {
-    { -32, 0 }, { -32, -32 }, { 0, 0 }, { -32, 0 }, { 0, 0 }, { 0, -32 }, { 0, -32 }, { -32, -32 },
+static constexpr const std::array<CoordsXY, 8> _fountainDirectionsNegative = {
```
duncanspumpkin: If these arrays were defined in terms of CoordsDirectionDelta positions would that be more readable?

src/openrct2/world/Map.cpp
```diff
@@ -63,8 +63,18 @@ using namespace OpenRCT2;
...
  * Replaces 0x00993CCC, 0x00993CCE
  */
-const CoordsXY CoordsDirectionDelta[] = { { -32, 0 },   { 0, +32 },   { +32, 0 },   { 0, -32 },
-                                          { -32, +32 }, { +32, +32 }, { +32, -32 }, { -32, -32 } };
+// clang-format off
+const std::array<CoordsXY, 8> CoordsDirectionDelta = {
+    CoordsXY{ -COORDS_XY_STEP, 0 },
+    CoordsXY{ 0, +COORDS_XY_STEP },
+    CoordsXY{ +COORDS_XY_STEP, 0 },
+    CoordsXY{ 0, -COORDS_XY_STEP },
```
duncanspumpkin: Might be more readable if this aligned up.

src/openrct2/world/Banner.cpp
tupaschoal: It's possible, but it would be something like (in a simplified example):
```c++
    static constexpr std::array<int, 8> a {1,2,3,4,5,6,7,8};
    static constexpr std::array<int, 9> b {a[0], a[2], a[3], a[4], a[5], a[6], a[7], 0};
```

src/openrct2/world/Duck.cpp
tupaschoal: Seems ok, these cast problems are typically warnings that don't show up on `msbuild`, but do on `cmake`, so let's see what the CI says.

src/openrct2/world/Fountain.cpp
tupaschoal: I'm not certain, because they're not a 1-1 copy of the `CoordsDirectionDelta`

src/openrct2/world/Map.cpp
tupaschoal: What do you mean?

duncanspumpkin: if the comma was in the same location on each line it would be more readable

Gymnasiast: If it was arranged like a table, with the x coordinates right beneath one another and the y coordinates too.

tupaschoal: Done

tupaschoal:
We only need to converge on the open issues now

src/openrct2/world/Fountain.cpp
duncanspumpkin: Okay thats fine.

src/openrct2/world/Banner.cpp
duncanspumpkin: leave as is.

PR #10574, opened 2020-01-15
Prefer CoordsDirectionDelta on footpath_update_path_wide_flags()

tupaschoal:
I recommend looking at each commit, to see what's best.

- On the first one, I just started using the `CoordsDirectionDelta`
- On the second one, I simplified it into a loop

### DIFF
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -1716,24 +1716,23 @@ void footpath_update_path_wide_flags(const CoordsXY& footpathPos)
         // Spanned from 0x00F3EFA8 to 0x00F3EFC7 (8 elements) in the original
         TileElement* pathList[8];
 
-        // TODO: Use DirectionDelta
-        auto pathPos = footpathPos - CoordsXY{ COORDS_XY_STEP, COORDS_XY_STEP };
+        auto pathPos = footpathPos + CoordsDirectionDelta[7];
         pathList[0] = footpath_can_be_wide(pathPos, height);
-        pathPos.y += COORDS_XY_STEP;
+        pathPos += CoordsDirectionDelta[1];
         pathList[1] = footpath_can_be_wide(pathPos, height);
-        pathPos.y += COORDS_XY_STEP;
+        pathPos += CoordsDirectionDelta[1];
         pathList[2] = footpath_can_be_wide(pathPos, height);
-        pathPos.x += COORDS_XY_STEP;
+        pathPos += CoordsDirectionDelta[2];
         pathList[3] = footpath_can_be_wide(pathPos, height);
-        pathPos.x += COORDS_XY_STEP;
+        pathPos += CoordsDirectionDelta[2];
         pathList[4] = footpath_can_be_wide(pathPos, height);
-        pathPos.y -= COORDS_XY_STEP;
+        pathPos += CoordsDirectionDelta[3];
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
I don't think this is quite right. This should be doing a look around every adjacent tile. I would expect the loop to be 1->8 and be: pos = footpathPos + CoordsDirectionDelta[direction]
Then the wide check.
The double addition looks incorrect.

src/openrct2/world/Footpath.cpp
```diff
@@ -1716,24 +1716,17 @@ void footpath_update_path_wide_flags(const CoordsXY& footpathPos)
...
+        Direction direction = 1;
+        int32_t i = 0;
+        while (i < 8)
+        {
+            pathList[i++] = footpath_can_be_wide(pathPos, height);
+            pathPos += CoordsDirectionDelta[direction];
+            pathList[i++] = footpath_can_be_wide(pathPos, height);
+            pathPos += CoordsDirectionDelta[direction];
+            direction = direction_next(direction);
+        }
```
duncanspumpkin: This likely works but we could do it simpler by having direction being 0->7. And the pos being a sum of the base footpathPos plus the Delta at the specified direction. It would be a different layout of the array though. Will need a little think to check if that is an issue.

tupaschoal: For reference, this is what the delta looks like at each position:
```
[0] = {-32, -32}
[1] = {-32, 0}
[2] = {-32, 32}
[3] = {0, 32}
[4] = {32, 32}
[5] = {32, 0}
[6] = {32, -32}
[7] = {0, -32}
```

tupaschoal: Which translates to:
```
[0] = footpathPos + CoordsDirectionDelta[7];
[1] = footpathPos + CoordsDirectionDelta[0];
[2] = footpathPos + CoordsDirectionDelta[4];
[3] = footpathPos + CoordsDirectionDelta[1];
[4] = footpathPos + CoordsDirectionDelta[5];
[5] = footpathPos + CoordsDirectionDelta[2];
[6] = footpathPos + CoordsDirectionDelta[6];
[7] = footpathPos + CoordsDirectionDelta[3];
```

tupaschoal: I've simplified it on the latest commit using the conversions above, please double check @duncanspumpkin .

src/openrct2/world/Footpath.cpp
```diff
@@ -1714,32 +1714,19 @@ void footpath_update_path_wide_flags(const CoordsXY& footpathPos)
...
-        pathPos.y -= COORDS_XY_STEP;
-        pathList[5] = footpath_can_be_wide(pathPos, height);
-        pathPos.y -= COORDS_XY_STEP;
-        pathList[6] = footpath_can_be_wide(pathPos, height);
-        pathPos.x -= COORDS_XY_STEP;
-        pathList[7] = footpath_can_be_wide(pathPos, height);
-        pathPos.y += COORDS_XY_STEP;
+        std::array<TileElement*, 8> pathList;
+
+        auto pathPos = footpathPos + CoordsDirectionDelta[7];
```
duncanspumpkin: This line can now be removed.

Gymnasiast: This is also why the CI fails.

tupaschoal:
Fixed and they all seem good to go now.

PR #10575, opened 2020-01-15
Prepare track piece handling code for more than 256 pieces

Gymnasiast:


src/openrct2/actions/TrackSetBrakeSpeedAction.hpp
```diff
@@ -16,12 +16,12 @@ DEFINE_GAME_ACTION(TrackSetBrakeSpeedAction, GAME_COMMAND_SET_BRAKES_SPEED, Game
 {
 private:
     CoordsXYZ _loc;
-    uint8_t _trackType;
+    track_type_t _trackType;
```
duncanspumpkin: This will break the network version so make sure you increment it.

PR #10576, opened 2020-01-16
Use a game action result for clear at function

duncanspumpkin:
I've been meaning to do this for a while.
The idea is to make it so that the error message args and ground flags are no longer passed around in global variables. Note this does not work at the moment as I need to modify the other callers of `map_obstruction_set_error_text`.

Argh can't believe I finally did what I always tell people not to do and recreated a pull request.

### DIFF
--- a/src/openrct2/actions/WallPlaceAction.hpp
+++ b/src/openrct2/actions/WallPlaceAction.hpp
@@ -279,16 +279,16 @@ public:
         bool wallAcrossTrack = false;
         if (!(GetFlags() & GAME_COMMAND_FLAG_PATH_SCENERY) && !gCheatsDisableClearanceChecks)
         {
-            if (!WallCheckObstruction(wallEntry, targetHeight / 8, clearanceHeight, &wallAcrossTrack))
+            if (auto result = WallCheckObstruction(wallEntry, targetHeight / 8, clearanceHeight, &wallAcrossTrack);
+                result->Error != GA_ERROR::OK)
             {
-                return std::make_unique<WallPlaceActionResult>(
-                    GA_ERROR::NO_CLEARANCE, gGameCommandErrorText, gCommonFormatArgs);
+                return result;
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/openrct2/actions/WallPlaceAction.hpp
```diff
@@ -380,16 +380,16 @@ DEFINE_GAME_ACTION(WallPlaceAction, GAME_COMMAND_PLACE_WALL, WallPlaceActionResu
         bool wallAcrossTrack = false;
         if (!(GetFlags() & GAME_COMMAND_FLAG_PATH_SCENERY) && !gCheatsDisableClearanceChecks)
         {
-            if (!WallCheckObstruction(wallEntry, targetHeight / 8, clearanceHeight, &wallAcrossTrack))
+            if (auto result = WallCheckObstruction(wallEntry, targetHeight / 8, clearanceHeight, &wallAcrossTrack);
```
Gymnasiast: I think it's better not to stuff the assignment into the `if` - it makes for harder-to-read code IMO.

src/openrct2/actions/WallPlaceAction.hpp
```diff
@@ -603,18 +603,16 @@ DEFINE_GAME_ACTION(WallPlaceAction, GAME_COMMAND_PLACE_WALL, WallPlaceActionResu
      *
      *  rct2: 0x006E5C1A
      */
-    bool WallCheckObstruction(rct_scenery_entry * wall, int32_t z0, int32_t z1, bool* wallAcrossTrack) const
+    GameActionResult::Ptr WallCheckObstruction(rct_scenery_entry * wall, int32_t z0, int32_t z1, bool* wallAcrossTrack) const
     {
         int32_t entryType, sequence;
         rct_scenery_entry* entry;
         rct_large_scenery_tile* tile;
-
```
tupaschoal: Seems like unrelated change

PR #10589, opened 2020-01-19
Remove clang-format from Travis

Gymnasiast:
Now clang-format has been moved to Github Actions, remove it from Travis to avoid running it twice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ before_install:
     - if [[ $TRAVIS_OS_NAME == "linux" ]]; then bash scripts/linux/install.sh; export OPENRCT_MAKE_OPTS="-k 10 all openrct2-cli" ; fi
 
 install:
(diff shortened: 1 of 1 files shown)


tupaschoal:
It also looks to be a lot faster on GH Actions

duncanspumpkin:
The best thing about it on GH actions is that it runs first.

tupaschoal:
I also like being able to go directly to the job that failed, instead of having to go to Travis first and then finding out what.

IntelOrca:
@duncanspumpkin I am not sure if GitHub actions actually priorities them, but since all jobs start almost instantly, the format job ends very soon after push.

IntelOrca:
@Gymnasiast looks like you pushed the branch to origin. Not sure if that was intentional but you can remove it now.

Gymnasiast:
@IntelOrca You mean I created the branch on the OpenRCT2 project? That's correct, but that was because I did this edit via the Github editor, which does not allow picking another origin (easily).

PR #10590, opened 2020-01-19
Fix #8064 Ride construction errors shown as (undefined string)

### LINKED ISSUE #8064: Construction error about local authority not allowing construction that high is shown as (undefined string) every other height

tupaschoal:
I'm picking up on stale #9432 

I did apply the changes to the maze, as @duncanspumpkin suggested:
> If you are going to make this change make the change for mazes as well its just inconsistent as it is at present.
> As this is a complex equation it should really be a constexpr that is named.

I didn't convert it to a `constexpr` because I was unsure if @ZehMatt 's comment was applicable without changing other places or not:
https://github.com/OpenRCT2/OpenRCT2/pull/9432/files#r294541067

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -3550,7 +3550,6 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
     z -= bx;
 
     gMapSelectArrowPosition.z = z;
-    bx = 41;
     _currentTrackBegin.x = mapCoords->x;
     _currentTrackBegin.y = mapCoords->y;
     _currentTrackBegin.z = z;
@@ -3561,9 +3560,12 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
     }
 
     _previousTrackPiece = _currentTrackBegin;
+    // search for appropriate z value for ghost, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
+
     if (ride->type == RIDE_TYPE_MAZE)
     {
-        for (;;)
+        for (int zAttempts = numAttempts; zAttempts > 1; zAttempts--)
         {
             window_ride_construction_update_state(
                 &trackType, &trackDirection, &rideIndex, &liftHillAndAlternativeState, &mapCoords->x, &mapCoords->y, &z,
@@ -3573,16 +3575,11 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
             if (_currentTrackPrice != MONEY32_UNDEFINED)
                 break;
 
-            bx--;
-            if (bx == 0)
-                break;
-
             _currentTrackBegin.z -= 8;
             if (_currentTrackBegin.z < 0)
                 break;
 
-            if (bx >= 0)
-                _currentTrackBegin.z += 16;
+            _currentTrackBegin.z += 16;
         }
 
         auto intent = Intent(INTENT_ACTION_UPDATE_MAZE_CONSTRUCTION);
@@ -3591,8 +3588,6 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
         return;
     }
 
-    // search for appropriate z value for ghost, up to max ride height
-    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
     for (int zAttempts = numAttempts; zAttempts > 1; zAttempts--)
     {
         window_ride_construction_update_state(
@@ -3793,9 +3788,12 @@ void ride_construction_tooldown_construct(ScreenCoordsXY screenCoords)
         z = _trackPlaceZ;
     }
 
+    // search for z value to build at, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
+
     if (ride->type == RIDE_TYPE_MAZE)
     {
-        for (int32_t zAttempts = 41; zAttempts >= 0; zAttempts--)
+        for (int32_t zAttempts = numAttempts; zAttempts > 0; zAttempts--)
         {
             _rideConstructionState = RIDE_CONSTRUCTION_STATE_MAZE_BUILD;
             _currentTrackBegin.x = mapCoords.x;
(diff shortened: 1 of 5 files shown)


Gymnasiast:
I want to test it.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3797,9 +3789,12 @@ void ride_construction_tooldown_construct(ScreenCoordsXY screenCoords)
         z = _trackPlaceZ;
     }
 
+    // search for z value to build at, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
+
     if (ride->type == RIDE_TYPE_MAZE)
     {
-        for (int32_t zAttempts = 41; zAttempts >= 0; zAttempts--)
+        for (int32_t zAttempts = numAttempts; zAttempts > 0; zAttempts--)
```
ZehMatt: Is there any reason this is done backwards?

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3859,7 +3854,7 @@ void ride_construction_tooldown_construct(ScreenCoordsXY screenCoords)
         return;
     }
 
-    for (int32_t zAttempts = 41; zAttempts >= 0; zAttempts--)
+    for (int32_t zAttempts = numAttempts; zAttempts > 0; zAttempts--)
```
ZehMatt: Same as above, this can be probably done forward.

src/openrct2-ui/windows/RideConstruction.cpp
tupaschoal: I don't think so, maybe the original author thought that counting down is more intuitive for attempts?

tupaschoal:
I applied the changes you requested @ZehMatt and everything seems to be working fine. I tested #8064 without my patch, saw the undefined strings and, applying the patch, they're gone.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3561,9 +3560,12 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
     }
 
     _previousTrackPiece = _currentTrackBegin;
+    // search for appropriate z value for ghost, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
+
     if (ride->type == RIDE_TYPE_MAZE)
     {
-        for (;;)
+        for (int zAttempts = 1; zAttempts < numAttempts; ++zAttempts)
```
ZehMatt: Whats the reason here to start at 1? The other loops seem to be starting at 0

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3844,7 +3839,7 @@ void ride_construction_tooldown_construct(ScreenCoordsXY screenCoords)
                     context_broadcast_intent(&intent2);
                     break;
                 }
-                else if (zAttempts >= 0)
+                else if (zAttempts < numAttempts)
```
ZehMatt: zAttempts < numAttempts will be always true

src/openrct2-ui/windows/RideConstruction.cpp
tupaschoal: The previous code had:
```
bx--;
if (bx == 0)
    break;
```
So it meant that the `currentTrackBegin.z += 16` was only done `bx - 1` times. I guess that's why the for starts at `1`

ZehMatt: Something doesn't quite add up, a few lines below it also had the check for when ```bx >= 0``` which is impossible to reach given that it would break out if ```bx == 0```. Can you test what happens if you were to start every loop at 0?

tupaschoal: Seems to not make a difference, at least in what I'm doing. I'm using `Raceway Park` and testing building above the tree line.

tupaschoal:
> This PR allows me to demolish trees on tiles that I only have construction rights for.
> 
> Can be tested easily by trying to build a flat ride on a construction-rights-owned piece of land with a tree on it.

I found out something very interesting. If you enable the sandbox mode to set the construction rights (and do), you can do this even without applying this branch. I'm thinking whether the game doesn't reset the land to just construction rights (kind of does OWN & CONSTRUCTION RIGHTS), I find it likely.

Gymnasiast:
Hm, it might be that I didn't turn off sandbox mode properly.

I'd like to double-check.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3561,9 +3560,12 @@ void ride_construction_toolupdate_construct(ScreenCoordsXY screenCoords)
     }
 
     _previousTrackPiece = _currentTrackBegin;
+    // search for appropriate z value for ghost, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
```
Gymnasiast: This is hard to read and uses hard coded `8`s instead of constants.

I'd prefer the following:
```suggestion
    const auto smallZ = z / COORDS_Z_STEP;
    int numAttempts = (smallZ <= MAX_TRACK_HEIGHT ? (MAX_TRACK_HEIGHT - smallZ + 1) : 2);
```

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3796,9 +3788,12 @@ void ride_construction_tooldown_construct(ScreenCoordsXY screenCoords)
         z = _trackPlaceZ;
     }
 
+    // search for z value to build at, up to max ride height
+    int numAttempts = (z <= (8 * MAX_TRACK_HEIGHT) ? MAX_TRACK_HEIGHT - (z / 8) + 1 : 2);
```
Gymnasiast: See previous comment

tupaschoal:
Done

PR #10591, opened 2020-01-19
Sprite Refactor

duncanspumpkin:
This modifies the size of rct_sprite to `0x200`. I've also moved a good number of functions to use the SpriteBase base class as then we can avoid a bunch of pointless casting.

I've also renamed `rct_vehicle` to `Vehicle` and renamed all of the other sprites that didn't have normal names.

In general I noticed a good few little mistakes along the way. I've tried to break up the PR into a bunch of smaller commits but I suspect it will be hard to follow no matter what.


src/openrct2/world/Sprite.h
```diff
@@ -138,7 +138,7 @@ union rct_sprite
     rct_money_effect* AsMoneyEffect();
     Peep* AsPeep();
 };
-assert_struct_size(rct_sprite, 0x100);
+assert_struct_size(rct_sprite, 0x200);
```
tupaschoal: Out of curiosity, I thought that `structs` with these asserts couldn't be changed due to export/import limitations, isn't that true?

src/openrct2/world/Sprite.cpp
```diff
@@ -689,26 +689,26 @@ void sprite_set_coordinates(int16_t x, int16_t y, int16_t z, rct_sprite* sprite)
  *
  *  rct2: 0x0069EDB6
  */
-void sprite_remove(rct_sprite* sprite)
+void sprite_remove(SpriteBase* sprite)
 {
-    auto peep = sprite->AsPeep();
+    auto peep = ((rct_sprite*)sprite)->AsPeep();
```
tupaschoal: This one sounds weird

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1518,19 +1518,17 @@ static int32_t cc_mp_desync(InteractiveConsole& console, const arguments_t& argv
...
     for (int i = 0; i < MAX_SPRITES; i++)
     {
         rct_sprite* sprite = get_sprite(i);
         if (sprite->generic.sprite_identifier == SPRITE_IDENTIFIER_NULL)
             continue;
 
-        if (sprite->generic.sprite_identifier == SPRITE_IDENTIFIER_PEEP)
-            peeps.push_back(sprite);
-        else if (sprite->generic.sprite_identifier == SPRITE_IDENTIFIER_VEHICLE)
-            vehicles.push_back(sprite);
```
tupaschoal: Why is the `vehicles` vector gone?

src/openrct2/world/Sprite.h
duncanspumpkin: Not quite true. These are the in memory structs. We can change their sizes and structure but any new fields can't be saved as that would break export/import.

src/openrct2/world/Sprite.cpp
duncanspumpkin: Yeah I really want to refactor so that `rct_sprite` is basically never used except when you need to know the max size of a sprite. For now this is just a stupid way of getting to the function I want.

src/openrct2/interface/InteractiveConsole.cpp
duncanspumpkin: It wasn't used for anything so thought there was no harm removing it. This function went through a few different versions during testing so there was a bit of dead code.

IntelOrca:
What do you intend to do with the extra 0x100 bytes?

duncanspumpkin:
> What do you intend to do with the extra 0x100 bytes?

Space for moving to bigger structures for the coordinates for one. I've chosen 0x100 just as a number that was bigger no real thought in it.

IntelOrca:
@duncanspumpkin well I think it makes sense to keep it a power of 2

src/openrct2/GameStateSnapshots.cpp
```diff
@@ -308,83 +308,82 @@ struct GameStateSnapshots : public IGameStateSnapshots
...
+        COMPARE_FIELD(Vehicle, var_44);
+        COMPARE_FIELD(Vehicle, mass);
+        COMPARE_FIELD(Vehicle, update_flags);
+        COMPARE_FIELD(Vehicle, swing_sprite);
+        COMPARE_FIELD(Vehicle, current_station);
+        COMPARE_FIELD(Vehicle, swinging_car_var_0);
+        COMPARE_FIELD(Vehicle, var_4E);
+        COMPARE_FIELD(Vehicle, status);
+        COMPARE_FIELD(Vehicle, sub_state);
         for (int i = 0; i < 32; i++)
```
Gymnasiast: While we're at it, could you get rid of these hard coded maximums?

src/openrct2/world/Sprite.h
tupaschoal: Ah, alright, thanks for explaining :)

src/openrct2/GameStateSnapshots.cpp
duncanspumpkin: I'll address that in a follow up pr.

PR #10070, opened 2019-10-09
Issue 9113 refactor g vehicle sound params list

UltimaBGD:
This is a refactor of the gVehicleSoundParams List as referenced in #9113.

I have renamed it to vehicleSoundParams and initialized it within the vehicle_sound_update, and pass it to vehicle_update_sound_params.

PR #10596, opened 2020-01-20
Remove global variable for sound param list

### LINKED ISSUE #9113: Refactor gVehicleSoundParamsList into a non global variable
gVehicleSoundParamsList is only used within the vehicle param update function. It has no reason to be a global variable and is cleared every tick. It should be replaced by a local variable that is passed into the required function. At the same time it could be replaced with a std::array.

tupaschoal:
Closes #9113 
Continues #10070 

This took some heavy rebasing, as the functions had changed pretty heavily. @duncanspumpkin concern on the original PR was:
> Hmm this certainly works but I'm thinking std::array is horrible for this. std::vector isn't great either due to the memory allocations.

I don't think it's that big of a deal, as it's a small array, but I'm open to suggestions

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -998,35 +998,36 @@ rct_vehicle_sound_params Vehicle::CreateSoundParam(uint16_t priority) const
  *
  *  rct2: 0x006BB9FF
  */
-void Vehicle::UpdateSoundParams(
-    std::array<rct_vehicle_sound_params, AUDIO_MAX_VEHICLE_SOUNDS> vehicleSoundParamsList,
-    rct_vehicle_sound_params* vehicleSoundParamsListEnd) const
+void Vehicle::UpdateSoundParams(std::vector<rct_vehicle_sound_params>& vehicleSoundParamsList) const
 {
     if (!SoundCanPlay())
         return;
 
     uint16_t soundPriority = GetSoundPriority();
-    rct_vehicle_sound_params* soundParam;
     // Find a sound param of lower priority to use
-    for (soundParam = &vehicleSoundParamsList[0];
-         soundParam < vehicleSoundParamsListEnd && soundPriority <= soundParam->priority; soundParam++)
-        ;
+    auto soundParamIter = std::find_if(
+        vehicleSoundParamsList.begin(), vehicleSoundParamsList.end(),
+        [soundPriority](rct_vehicle_sound_params param) { return soundPriority > param.priority; });
 
-    if (soundParam >= &vehicleSoundParamsList[std::size(vehicleSoundParamsList)])
-        return;
-
-    if (vehicleSoundParamsListEnd < &vehicleSoundParamsList[std::size(vehicleSoundParamsList)])
+    if (soundParamIter == std::end(vehicleSoundParamsList))
     {
-        vehicleSoundParamsListEnd++;
+        if (vehicleSoundParamsList.size() < AUDIO_MAX_VEHICLE_SOUNDS)
+        {
+            vehicleSoundParamsList.push_back(CreateSoundParam(soundPriority));
+        }
     }
-
-    // Shift all sound params down one if using a free space
-    if (soundParam != vehicleSoundParamsListEnd)
+    else
     {
-        std::memmove(
-            soundParam + 1, soundParam, ((vehicleSoundParamsListEnd - soundParam) - 1) * sizeof(rct_vehicle_sound_params));
+        if (vehicleSoundParamsList.size() < AUDIO_MAX_VEHICLE_SOUNDS)
+        {
+            // Shift all sound params down one if using a free space
+            vehicleSoundParamsList.insert(soundParamIter, CreateSoundParam(soundPriority));
+        }
+        else
+        {
+            *soundParamIter = CreateSoundParam(soundPriority);
+        }
     }
-    *soundParam = CreateSoundParam(soundPriority);
 }
 
 static void vehicle_sounds_update_window_setup()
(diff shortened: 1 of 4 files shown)


tupaschoal:
Any performance checks to be specific or just test it manually?

Personally I don't like the list pointer tracker as well hahah

ZehMatt:
I guess thats why he suggested vector. I've been looking around for a small vector/stack vector/fixed vector implementations but I haven't come to testing them. Ideally we have a stack vector that operates exactly like the normal std::vector but storage is on stack and it can not grow dynamically. There are more patterns like this across the code base so I think it would make sense to introduce such container at some point. Still worth to check if std::vector has a huge impact or not, ideally try to use bpb.sv6 from our test data, it has a bunch of rides and peeps so a lot of audio should be played.

tupaschoal:
I'm sorry that it took me a while to catch up on this @ZehMatt and @duncanspumpkin . I converted it to use `std::vector` and not keep track of the end pointer anymore. As far as testing on `bpb.sv6` goes, there seems to be no performance drawback.

tupaschoal:
In hindsight, it might not be 100% accurate yet, but it's close

src/openrct2/ride/Vehicle.cpp
```diff
@@ -998,34 +998,29 @@ rct_vehicle_sound_params Vehicle::CreateSoundParam(uint16_t priority) const
...
+        vehicleSoundParamsList.begin(), vehicleSoundParamsList.end(),
+        [&soundPriority](rct_vehicle_sound_params param) { return soundPriority > param.priority; });
 
-    if (soundParam >= &gVehicleSoundParamsList[std::size(gVehicleSoundParamsList)])
+    if (soundParamIter == std::end(vehicleSoundParamsList))
         return;
 
-    if (gVehicleSoundParamsListEnd < &gVehicleSoundParamsList[std::size(gVehicleSoundParamsList)])
+    // Shift all sound params down one if using a free space
+    if (vehicleSoundParamsList.size() < vehicleSoundParamsList.capacity())
```
ZehMatt: I'm not sure that this is safe to rely on. Calling reserve means that it will reserve a minimum of N and not the exact.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -998,34 +998,29 @@ rct_vehicle_sound_params Vehicle::CreateSoundParam(uint16_t priority) const
...
     {
-        gVehicleSoundParamsListEnd++;
+        vehicleSoundParamsList.insert(soundParamIter, CreateSoundParam(soundPriority));
```
ZehMatt: Is the order actually really important here? I think a normal push back will probably do, but I haven't tested that theory.

src/openrct2/ride/Vehicle.cpp
tupaschoal: Agreed, I'll change to compare size with the `define`

src/openrct2/ride/Vehicle.cpp
tupaschoal: I assumed it is, but didn't really experiment without, I imagined it would be something like "play only 14 sounds, if any of the other are of lower priority, they just get ignored to avoid becoming a mess of sounds"

src/openrct2/ride/Vehicle.cpp
```diff
@@ -998,34 +998,36 @@ rct_vehicle_sound_params Vehicle::CreateSoundParam(uint16_t priority) const
...
+            // Shift all sound params down one if using a free space
+            vehicleSoundParamsList.insert(soundParamIter, CreateSoundParam(soundPriority));
+        }
+        else
+        {
+            *soundParamIter = CreateSoundParam(soundPriority);
+        }
     }
-
-    *soundParam = CreateSoundParam(soundPriority);
```
tupaschoal: So what I understood for the old behavior is that:
- It had an `AUDIO_MAX_VEHICLE_SOUNDS`-sized array
- This array was sorted in descending order by sound priority
- When a new sound was added, it kept the order, as long as the array was not full already, otherwise...
- It would replace the first sound found with lower priority

This last point seems wrong to me, suppose `AUDIO_MAX_VEHICLE_SOUNDS = 3` and thus a vector of priorities `{4, 2, 1}`. If we try to add a sound with priority `3`, the old code would then have the remaining vector as `{4, 3, 1}` instead of `{4, 3, 2}`.

The new code is doing the same thing, but it is easy to fix if you think this makes sense. Let me know!

ZehMatt: I see, yeah leave it as it is.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -998,34 +998,36 @@ rct_vehicle_sound_params Vehicle::CreateSoundParam(uint16_t priority) const
...
 
     uint16_t soundPriority = GetSoundPriority();
-    rct_vehicle_sound_params* soundParam;
     // Find a sound param of lower priority to use
-    for (soundParam = &gVehicleSoundParamsList[0];
-         soundParam < gVehicleSoundParamsListEnd && soundPriority <= soundParam->priority; soundParam++)
-        ;
+    auto soundParamIter = std::find_if(
+        vehicleSoundParamsList.begin(), vehicleSoundParamsList.end(),
+        [&soundPriority](rct_vehicle_sound_params param) { return soundPriority > param.priority; });
```
ZehMatt: Capture by copy, no need for reference here.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1311,21 +1314,20 @@ void vehicle_sounds_update()
         }
     }
 
-    for (rct_vehicle_sound_params* vehicleSoundParams = &gVehicleSoundParamsList[0];
-         vehicleSoundParams < gVehicleSoundParamsListEnd; vehicleSoundParams++)
+    for (auto vehicleSoundParams : vehicleSoundParamsList)
```
ZehMatt:
```suggestion
    for (auto& vehicleSoundParams : vehicleSoundParamsList)
```

tupaschoal:
Done

PR #10599, opened 2020-01-20
Fix water sides glitch

Gymnasiast:
Fixes this problem:

Diamond Heights 2020-01-20 20-50-11


duncanspumpkin:
Yes this looks correct but at somepoint we need to refactor these functions to use bigZ. I'm surprised these are using tinyZ and not even smallZ.

PR #10604, opened 2020-01-22
Increase height of track save buttons

AaronVanGeffen:
This increases the height of the track save buttons in the ride window. While previous PRs have covered most of these, this will finally allow for CJK languages to fit the button captions with ease. For languages using the sprite font, it makes the window look less cramped as a fortunate side-effect.

Before:
Forest Frontiers 2020-01-22 12-28-04

After:
Forest Frontiers 2020-01-22 12-33-45


AaronVanGeffen:
This PR aside, I wonder if it would be better to just lay out the buttons vertically, instead of in the current four-corners approach. I don't want to ruffle too many feathers at once, though.

Gymnasiast:
> This PR aside, I wonder if it would be better to just lay out the buttons vertically, instead of in the current four-corners approach.

The current approach has the advantage of grouping the selection options and the save/cancel options together. I think it's more user-friendly this way. Another advantage is that adding more options to the selection group will keep it neat.

PR #10606, opened 2020-01-22
Fix #10598. Mistake made when fixing #7784.

### LINKED ISSUE #10598: All trains in the ride-window appear to have to colors of the first train
**OS:** Windows 10
**Version:** v0.2.4-458-gf1dcf8573
**Commit/Build:** f1dcf85

When trying to set the colors of each train of a ride separately, the colors do not appear properly in the ride window. All trains will appear to have the same colors as "Train 1", even if you have set different colors for "Train 2". The color of the train on the track does change properly.

Another place where this is visible is when building a Go-Karts. All vehicles in the ride-window appear to have the same color, but in reality they all have different colors (as expected)

**Steps to reproduce:**
1.  Build a tracked ride with more than 1 train
2.  Go to the color-tab in the ride-window
3. Set "Different colors per train"
4. Change the color of "Train 2"

**Screenshots / Video:**

colors1
colors2

duncanspumpkin:
Fix #10598. Mistake made when fixing #7784.

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -1357,7 +1357,8 @@ static void window_ride_draw_tab_vehicle(rct_drawpixelinfo* dpi, rct_window* w)
             ride->subtype, ride->num_cars_per_train, rideEntry->tab_vehicle);
         rct_ride_entry_vehicle* rideVehicleEntry = &rideEntry->vehicles[vehicle];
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Okay but #10544 fix breaks it for all rides. I think it better that only car rides are broken.

duncanspumpkin:
Okay I've got a better fix that shouldn't cause any other issues since it reverts the changes to the function.

PR #10608, opened 2020-01-22
Convert gResearchLast/NextItem to std::optional

Gymnasiast:


src/openrct2/rct2/S6Importer.cpp
```diff
@@ -262,11 +262,22 @@ class S6Importer final : public IParkImporter
...
+        if (_s6.last_researched_item_subject != RCT12_RESEARCHED_ITEMS_SEPARATOR)
+            gResearchLastItem->rawValue = _s6.last_researched_item_subject;
+        else
+            gResearchLastItem = std::nullopt;
         // pad_01357CF8
-        gResearchNextItem.rawValue = _s6.next_research_item;
+        if (_s6.next_research_item != RCT12_RESEARCHED_ITEMS_SEPARATOR)
+        {
+            gResearchNextItem->rawValue = _s6.next_research_item;
+            gResearchNextItem->category = _s6.next_research_category;
```
IntelOrca: Setting the fields directly on an optional does not change the has_value flag. Make sure you use something like `std::make_optional` first otherwise `gResearchNextItem` will still be `std::nullopt`.

Gymnasiast: Do I need to use `std::make_optional` on all assignments, or just on those where I set fields directly?

IntelOrca: `std::make_optional` will give you the constructor of the type. If you have a constructor that lets you set all the fields, then you can just make one single call to `std::make_optional`. If you want to set fields 1 by 1, then you need to call `std::make_optional` to reset it to the default initial value of an initialised `ResearchItem` then set the fields.

src/openrct2/management/Research.cpp
```diff
@@ -175,7 +175,7 @@ static void research_next_design()
         it++;
     }
 
-    gResearchNextItem = researchItem;
+    gResearchNextItem = std::make_optional(researchItem);
```
IntelOrca: `std::optional` is not require here as you are setting an optional field to an actual value of a ResearchItem. It is like doing:

```
std::optional<int> x = std::make_optional(5);
```
rather than just:
```
std::optional<int> x = 5;
```

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -262,11 +262,18 @@ class S6Importer final : public IParkImporter
 
         gResearchPriorities = _s6.active_research_types;
         gResearchProgressStage = _s6.research_progress_stage;
-        gResearchLastItem.rawValue = _s6.last_researched_item_subject;
+        if (_s6.last_researched_item_subject != RCT12_RESEARCHED_ITEMS_SEPARATOR)
+            gResearchLastItem = std::make_optional(
+                ResearchItem{ _s6.last_researched_item_subject, RESEARCH_CATEGORY_TRANSPORT });
```
IntelOrca: If you make a proper constructor for `ResearchItem`, then I think you can just do:

```
std::make_optional<ResearchItem>(_s6.last_researched_item_subject, RESEARCH_CATEGORY_TRANSPORT);
```

src/openrct2/scenario/Scenario.h
```diff
@@ -162,9 +162,9 @@ struct rct_s6_data
     // SC6[10]
     uint8_t active_research_types;
     uint8_t research_progress_stage;
-    uint32_t last_researched_item_subject;
+    int32_t last_researched_item_subject;
     uint8_t pad_01357CF8[1000];
-    uint32_t next_research_item;
+    int32_t next_research_item;
```
IntelOrca: How come you decided to go signed? Wouldn't it be better to make a constant for `0xFFFFFFFF` and keep it unsigned?

Gymnasiast: Mostly because the rest of the research items were signed as well. I could have changed all of them, but that would have been more work.

PR #10612, opened 2020-01-23
Move Linux portable actions to GitHub CI

IntelOrca:


.github/workflows/ci.yml
```diff
@@ -9,11 +9,12 @@ jobs:
     name: Check code formatting
     runs-on: ubuntu-latest
     container:
-      image: openrct2/openrct2:format
+      image: openrct2/openrct2-build:0.2.4-format
```
AaronVanGeffen: Any particular reason for attaching a version number to this task?

IntelOrca: I thought it would be good to start versioning our build images so we are able to pull old ones if we need them.

Some of our images have a moving base image and also download third party sources, this means you can not guarantee that old dockerfiles will still build which is actually a problem we have with our mingw one at the moment.

PR #10613, opened 2020-01-23
Implement As and Is functions for SpriteBase

duncanspumpkin:
This is a little refactoring to bring in functions that work similar to OpenLoco's thing get functions.

I plan to change the `get_sprite` functions to return a `SpriteBase*` and these functions will be used to get the class. When making this change extra checks will be implemented to make sure the returned value is not being blindly converted to a different type.


src/openrct2-ui/windows/Guest.cpp
```diff
@@ -602,7 +602,7 @@ static void window_guest_common_invalidate(rct_window* w)
  */
 void window_guest_disable_widgets(rct_window* w)
 {
-    Peep* peep = &get_sprite(w->number)->peep;
+    Peep* peep = get_sprite(w->number)->As<Peep>();
```
ZehMatt: While you are at it we could template get_sprite to avoid casting afterwards.

duncanspumpkin:
Binary file accidently added in `Change get_sprite to return a SpriteBase`

duncanspumpkin:
This will cause crashes until all of the loops and code properly handles the nullptrs returned by the get function. (Especially GET_PEEP and GET_VEHICLE)

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4353,7 +4353,7 @@ static void window_ride_maintenance_paint(rct_window* w, rct_drawpixelinfo* dpi)
             }
             else
             {
-                auto peep = (&(get_sprite(ride->mechanic)->peep))->AsStaff();
+                auto peep = get_sprite(ride->mechanic)->As<Peep>()->AsStaff();
```
Gymnasiast, line 4356: It would be great if one could just do 
```suggestion
                auto peep = get_sprite(ride->mechanic)->As<Staff>();
```

src/openrct2/drawing/LightFX.cpp
```diff
@@ -694,7 +694,6 @@ uint32_t lightfx_get_light_polution()
 void lightfx_add_lights_magic_vehicle(const Vehicle* vehicle)
 {
     uint16_t vehicleID = vehicle->sprite_index;
-
```
Gymnasiast, line 697: Unintended change.

duncanspumpkin:
I'm closing this for now. I'm still planning on doing this but will do it slightly differently.

PR #10620, opened 2020-01-25
Fix testing giga coaster with cable lift crashing the game

Gymnasiast:
Originally reported here: https://forums.openrct2.org/topic/4750-game-crashes-when-using-ghost-train


duncanspumpkin:
Yes this is correct but I'm pretty sure it's a regression that was made recently so no change log entry.

PR #10640, opened 2020-02-01
Enable Android builds with GitHub Actions

IntelOrca:
Android builds building, working and uploading (both to CI and openrct2.org) again.

I have removed building of x86 and x86-64 builds for now as they double the CI time (+20 minutes) and I doubt they are that likely to be used.

Android builds can now be built every push and pull request in around 10 minutes, similar time to Windows.


AaronVanGeffen:
This PR has been merged, but has the cron been removed from Travis as well?

IntelOrca:
@AaronVanGeffen I have removed it now, thanks.

PR #10651, opened 2020-02-03
Place constexpr functions in anonymous namespace

janisozaur:


janisozaur:
I got a comment requesting I do the namespace thing instead of `static` previously. I think it was this project, but can't be bothered to find it.

IntelOrca:
Could you explain the reason you are doing this in your commit?

janisozaur:
`constexpr` is not required to evaluate on compile time and could be a regular function in such case. It should then be marked as `static` or put in anonymous namespace.

AaronVanGeffen:
I'd prefer sticking with `constexpr` if it means we can avoid the -frankly- ugly `[[maybe_unused]]` annotations.

janisozaur:
No, `constexpr` can still be unused, as you could see in one of the previous runs, reported by clang.

tupaschoal:
> I got a comment requesting I do the namespace thing instead of `static` previously. I think it was this project, but can't be bothered to find it.

Alright, no big deal

PR #10653, opened 2020-02-03
Prevent shadowing of member fields

janisozaur:


janisozaur:
What's your suggestion then?

IntelOrca:
Well first of all, is there any reason to change it? I know it is a compiler warning if enabled but many language allow shadowing because it is quite useful or avoids this issue of finding different names.

My suggestion would be to use TitleCase for the fields of the class / struct since they are public.

Broxzier:
clang-tidy has a check for identifier naming: `readability-identifier-naming`. Perhaps we could use this for consistent naming all over the project.

Gymnasiast:
This PR is getting way too big for its own good, especially since there seems to be no consensus about whether it actually solves a problem or not.

The mass renaming of fields to adhere to our newer code style should really be put into a separate PRs. The peep field rename warrants a PR of its own.

tupaschoal:
The renaming does solve the shadowing, but I agree it is big, I intended to cherry pick each class into their own PR, I just didn't have any time yet.

If you can cherry pick the money effect and serverlist, I think they're quite small and good to go.

IntelOrca:
The diff also shows changes to string literals... were the changes automatically done?

tupaschoal:
Yeah, I mostly used "rename" from Visual Studio and changed only the things it stated as "confirmed references"

tupaschoal:
I'm closing this one in favour of the three PRs mentioned above.

ServerList and Money Effect are pretty straightforward and were done on a single PR each. As for Peep I'll take it in batches of 10, as doing it all at once might be too disruptive and a hell of merges.

PR #10654, opened 2020-02-03
Replace cURL with WinHttp implementation on Windows

IntelOrca:
This reduces the number of third party dependencies for Windows builds. WinHttp is quite a nice straight forward API so doesn't involve too much extra code.

### DIFF
--- a/src/openrct2/core/Http.WinHttp.cpp
+++ b/src/openrct2/core/Http.WinHttp.cpp
@@ -1,0 +1,223 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2020 OpenRCT2 developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#if !defined(DISABLE_HTTP) && defined(_WIN32)
+
+#    include "Http.h"
+
+#    include "../Version.h"
+#    include "String.hpp"
+
+#    include <cstdio>
+#    include <stdexcept>
+#    define WIN32_LEAN_AND_MEAN
+#    include <windows.h>
+#    include <winhttp.h>
+
+namespace Http
+{
+    static constexpr char OPENRCT2_USER_AGENT[] = "OpenRCT2/" OPENRCT2_VERSION;
+
+    static void ThrowWin32Exception(const char* methodName)
+    {
+        auto errorCode = GetLastError();
+        auto msg = String::StdFormat("%s failed, error: %d.", methodName, errorCode);
+        throw std::runtime_error(msg);
+    }
+
+    static const wchar_t* GetVerb(Method method)
+    {
+        switch (method)
+        {
+            case Method::GET:
+                return L"GET";
+            case Method::POST:
+                return L"POST";
+            case Method::PUT:
+                return L"PUT";
+            default:
+                throw std::runtime_error("Unsupported verb.");
+        }
+    }
+
+    static std::wstring ReadHeader(HINTERNET hRequest, DWORD infoLevel, const wchar_t* name)
+    {
+        wchar_t headerBuffer[256]{};
+        auto headerBufferLen = (DWORD)std::size(headerBuffer);
+        if (!WinHttpQueryHeaders(hRequest, infoLevel, name, headerBuffer, &headerBufferLen, WINHTTP_NO_HEADER_INDEX))
+            ThrowWin32Exception("WinHttpQueryHeaders");
+        return std::wstring(headerBuffer, headerBufferLen);
+    }
+
+    static int32_t ReadStatusCode(HINTERNET hRequest)
+    {
+        auto wStatusCode = ReadHeader(hRequest, WINHTTP_QUERY_STATUS_CODE, L"StatusCode");
+        return std::stoi(wStatusCode);
+    }
+
+    static std::map<std::string, std::string> ReadHeaders(HINTERNET hRequest)
+    {
+        DWORD dwSize{};
+        WinHttpQueryHeaders(
+            hRequest, WINHTTP_QUERY_RAW_HEADERS, WINHTTP_HEADER_NAME_BY_INDEX, NULL, &dwSize, WINHTTP_NO_HEADER_INDEX);
+
+        std::wstring buffer;
+        buffer.resize(dwSize);
+        if (!WinHttpQueryHeaders(
+                hRequest, WINHTTP_QUERY_RAW_HEADERS, WINHTTP_HEADER_NAME_BY_INDEX, buffer.data(), &dwSize,
+                WINHTTP_NO_HEADER_INDEX))
+            ThrowWin32Exception("WinHttpQueryHeaders");
+
+        std::map<std::string, std::string> headers;
+        std::wstring wKey, wValue;
+        int state = 0;
+        int index = 0;
+        for (auto c : buffer)
+        {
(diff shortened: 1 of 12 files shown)


Gymnasiast:
You upvoted your _own_ PR?!

src/openrct2/core/Http.WinHttp.cpp
```diff
@@ -0,0 +1,223 @@
...
+            auto reqBody = (LPVOID)req.body.data();
+            auto reqBodyLen = (DWORD)req.body.size();
+            if (!WinHttpSendRequest(hRequest, WINHTTP_NO_ADDITIONAL_HEADERS, 0, reqBody, reqBodyLen, reqBodyLen, 0))
+                ThrowWin32Exception("WinHttpSendRequest");
+
+            if (!WinHttpReceiveResponse(hRequest, NULL))
+                ThrowWin32Exception("WinHttpReceiveResponse");
+
+            auto statusCode = ReadStatusCode(hRequest);
+            auto contentType = ReadHeader(hRequest, WINHTTP_QUERY_CONTENT_TYPE, L"Content-Type");
```
ZehMatt: Could probably save some code by moving this line after ReadHeaders and then use the map to get this field.

src/openrct2/core/Http.WinHttp.cpp
```diff
@@ -0,0 +1,223 @@
...
+        std::string body;
+        do
+        {
+            // Check for available data.
+            if (!WinHttpQueryDataAvailable(hRequest, &dwSize))
+                ThrowWin32Exception("WinHttpQueryDataAvailable");
+
+            // No more available data.
+            if (dwSize == 0)
+                break;
```
ZehMatt: Remarks from MSDN:
```
Important  Do not use the return value of WinHttpQueryDataAvailable to determine whether the end of a response has been reached, because not all servers terminate responses properly, and an improperly terminated response causes WinHttpQueryDataAvailable to anticipate more data.
```
They explicitly say
```
Use the return value of WinHttpReadData to determine when a response has been completely read.
```
See https://docs.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpquerydataavailable

src/openrct2/core/Http.WinHttp.cpp
```diff
@@ -0,0 +1,223 @@
...
+        return body;
+    }
+
+    Response Do(const Request& req)
+    {
+        HINTERNET hSession{}, hConnect{}, hRequest{};
+        try
+        {
+            URL_COMPONENTS url;
+            ZeroMemory(&url, sizeof(url));
```
ZehMatt: std::memset, ZeroMemory does not count as an intrinsic.

src/openrct2/core/Http.WinHttp.cpp
```diff
@@ -0,0 +1,223 @@
...
+        if (!WinHttpQueryHeaders(
+                hRequest, WINHTTP_QUERY_RAW_HEADERS, WINHTTP_HEADER_NAME_BY_INDEX, buffer.data(), &dwSize,
+                WINHTTP_NO_HEADER_INDEX))
+            ThrowWin32Exception("WinHttpQueryHeaders");
+
+        std::map<std::string, std::string> headers;
+        std::wstring wKey, wValue;
+        int state = 0;
+        int index = 0;
+        for (auto c : buffer)
```
ZehMatt: The whole code for iterating the headers seems overkill. Consider using WinHttpQueryHeaders correctly with the index parameter. You can easily iterate all headers by starting at index 0, after the call the variable will be set to the next index and if it runs out it will have ERROR_WINHTTP_HEADER_NOT_FOUND.

src/openrct2/core/Http.WinHttp.cpp
```diff
@@ -0,0 +1,223 @@
...
+        auto headerBufferLen = (DWORD)std::size(headerBuffer);
+        if (!WinHttpQueryHeaders(hRequest, infoLevel, name, headerBuffer, &headerBufferLen, WINHTTP_NO_HEADER_INDEX))
+            ThrowWin32Exception("WinHttpQueryHeaders");
+        return std::wstring(headerBuffer, headerBufferLen);
+    }
+
+    static int32_t ReadStatusCode(HINTERNET hRequest)
+    {
+        auto wStatusCode = ReadHeader(hRequest, WINHTTP_QUERY_STATUS_CODE, L"StatusCode");
+        return std::stoi(wStatusCode);
```
ZehMatt: The status code can never be a negative and is illegal.

src/openrct2/core/Http.WinHttp.cpp
IntelOrca: Hmm, this snippet of code was copied from their example.

src/openrct2/core/Http.WinHttp.cpp
```diff
@@ -0,0 +1,223 @@
...
+            auto contentType = ReadHeader(hRequest, WINHTTP_QUERY_CONTENT_TYPE, L"Content-Type");
+            auto headers = ReadHeaders(hRequest);
+            auto body = ReadBody(hRequest);
+
+            Response response;
+            response.body = std::move(body);
+            response.status = (Status)statusCode;
+            response.content_type = String::ToUtf8(contentType);
+            response.header = headers;
+            return response;
```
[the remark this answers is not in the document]
IntelOrca: You're right, I thought I was in a finally block but that of course isn't a thing in C++. It should do because not all APIs have RAII.

src/openrct2/core/Http.WinHttp.cpp
IntelOrca: The index parameter is only if you are dealing with multiple headers with the same key. Since we store them in a map, we can't store more than one anyway.
It also does not separate the key and value apart.

src/openrct2/core/Http.WinHttp.cpp
ZehMatt: The documentation says do not use it, dunno if the example is bad or not but I would rather trust the documentation over snippets made up by interns. It has probably to do with servers that do not transfer the Content-Length field and relies on terminating the connection when the response completes.

src/openrct2/core/Http.WinHttp.cpp
ZehMatt: Looks like you are right, sorry. I also wondered if we actually have to parse all headers, only the encoding is relevant so far.

src/openrct2/core/Http.WinHttp.cpp
IntelOrca: Can you suggest the change to make?

ZehMatt: According to MSDN:
```If you are using WinHttpReadData synchronously, and the return value is TRUE and the number of bytes read is zero, the transfer has been completed and there are no more bytes to read on the handle. This is analogous to reaching end-of-file in a local file.```

File: src/openrct2/core/Http.WinHttp.cpp [hunk 7]
```diff
@@ -0,0 +1,234 @@
...
+            else
+            {
+                state = STATE_EXPECT_VALUE;
+                wValue.push_back(c);
+            }
+        }
+        return headers;
+    }
+
+    static std::string ReadBody(HINTERNET hRequest)
```
ZehMatt: I think it makes more sense to use std::vector<uint8_t> as a result since we are dealing with binary.

IntelOrca: I get your point but it would just been changing the http response struct used in curl code and everywhere else.

ZehMatt: I see, nevermind then.

PR #10660, opened 2020-02-05
Replace OpenSSL with Windows BCrypt API

IntelOrca:
Removing another third party dependency from Windows builds. This also reduces the size of builds by ~2MB.

Unfortunately there is no support for reading or writing DER / PEM files so I had to write enough code to read/write the current key files we use.


Gymnasiast:
Looks alright - have you tested it throughly? (Preferably also cross-platform - Windows clients connecting to a Linux server and vice versa.)

IntelOrca:
Yes

Issue #714, opened 2015-01-21
Allow more than 31 trains per ride


tehplanefreak:
In RollerCoaster Tycoon 2, roller coasters had a cap on how many cars per train you could set, and no one could bypass such a limit, even with a trainer. I'd like to see this cap removed in OpenRCT2, if the logic involving this hasn't been decompiled already.

AaronVanGeffen:
We'd all like to go there, of course. Unfortunately, much more of the game's logic needs to be decompiled before we can — OpenRCT2 still heavily relies on RCT2's routines in those areas.

Gymnasiast:
8 cars per trainer has a function for 8 cars per train*, so even with RCT2's routines in place, there is some room around this limit. Although completely lifting the limit would indeed require much more decompilation. It's quite high on my wishlist as well. :)

* Which works on Windows 98 and on Linux via Wine, but I'm not sure whether it still works on Windows XP and higher, since I recall more functions suddenly not working, like 10 foot shops.

IntelOrca:
These sorts of limits along with more than 255 rides, more than 10000 sprites, larger maps, more available objects in one park and more scenery themes require changing of the game structures which we can't do almost the whole game is decompiled. It will also result in saved games files which are no longer compatible with the original game. These changes will come in the distant future.

tehplanefreak:
Thank you all for the comments. :) 

Gymnasiast, as a side note, I tried using the 8 cars per train function (1.302) on RCT2 running on a Windows 98 virtual machine, but it doesn't appear to work.

Gymnasiast:
It might not work with non-US versions of the game, and not with all rides. Are you using the US version? And which rides are you trying it on?

tehplanefreak:
I tried it with the Triple Thrill Pack (2.01.043, distributed by GOG) on a Corkscrew Coaster.

Gymnasiast:
I think that the GOG version differs a little from the original CD version (which I have).

Strangely, it stopped working on my Wine installation as well.
Anyway, your best bet is probably waiting until OpenRCT2 can ease/eliminate this limit (after full decompilation, as IntelOrca said).

sjoerddebruin:
Is this possible now?

Gymnasiast:
@sjoerddebruin No. I have looked into it, but it looks like an SV6 limitation.

Nubbie:
[...]

Issue #681, opened 2015-01-12
Allow More Specific Patrol Area


pizza2004:
I don't know what people's opinions on this would be, but I've been playing RCT1 lately and it really bugs me that I can't adjust the patrol area one square at a time.  I want to be able to restrict each handyman to a very specific area, but it's hard because you can only set patrol area in 4x4 squares at a time.

vzybilly:
This would be amazing. If possible, I could also see one way or one way (go the same way that they came in) with turns (left, right, turn around) to stop some of the crazy things they do. I've seen a sweeper go back and forth on the same 3 paths while his service area was larger, I ended up needing 3 people for a single 'coaster exit from that...

pizza2004:
Another thing that would be nice is to be able to assign the mechanics to work on specific rides, and they'd just stay somewhat nearby the rides they're assigned.  That feels like it would be a lot better than giving them a patrol area.

vzybilly:
that does sound pretty nifty as well, but I think to get all of the awesome out of the staff and paths, the pathfinding will need to be done and from what I hear, it isn't...

Gymnasiast:
Sounds good.
As for the handyman issues: these may be solved with better pathfinding code (going to nearby litter instead of just randomly walking), which I think will eventually be in OpenRCT2 (though only after the whole game has been reverse engineered).

duncanspumpkin:
Its not quite as easy as just making handymen clean litter whenever there is one nearby. If that change was implemented the whole dynamic of handymen would be changed. If we improve handymen a.i. then something will need to balance this out. Maybe having to introduce handymen having lunch to prevent the game from becoming too easy. Wages would need to be adjusted and...

So the a.i. could be improved but the knock on effects will change the game a lot. 

Bugs in the a.i. preventing peeps from finding destinations when fixed shouldn't cause too many issues same with making some way for staff to walk their whole patrol area.

For anyone who does want to look into the a.i. more than have a look through 0x0068FD3A which i think might be the function that works out the direction for the peep to walk.

IntelOrca:
As for patrol area, map sizes can go up to 256 x 256. This means there are 65536 possible tiles. Using a bit mask, 65536 / 8 = 8192 (8KB) would be needed per staff member. This is why the current limit is 4x4 which only requires (0.5KB) per staff member. This is also probably why there is a staff limit too.

[...]

PR #10664, opened 2020-02-07
.park implementation

Gymnasiast:
# **This is not a general Q&A thread. Ask any questions on Discord. Questions placed here will be removed. Any issues you find should be reported as regular issues, making sure to mention they only affect the new save format.**

This is still in development, but it's easier to do proper discussion if we have a PR. Hence the "Draft PR".

Still to implement:
- [x] Entities
    - [x] Remove all unnecessary fields from SpriteBase (~~LinkedListId, next, previous, next_in_quadrant,~~ flags)
    - [x] Replace sprite_identifier with EntityType (remove sub types for MiscEntities)
    - [x] Replace EntityLists with lists (remove `next` field)
    - ~~[ ] Invert ownership of entities so that they store in the lists~~
    - ~~[ ] Make entity removal/addition happen at end of tick~~
    - ~~[ ] Replace EntityLists with vectors created on load (remove `next` field)~~
    - [x] Rework checksum to compute by entity type
    - [x] Rework replays to save by entity type
    - [x] Rework s6 save test to use GameState
    - ~~[ ] Investigate best way to hold all Entity types~~
    - ~~[ ] Move sprite flashing to ui project~~
    - [x] Remove all for loops over entities that use GetEntity (except where required)
    - [x] split guests and staff
- [x] Research list
- [x] Cheats (implemented by ZehMatt)
- [x] Fix crash when raising object limits (#10881)
- [x] Invisibility flag instead of corrupted elements
- [x] Revert RidesBeenOn hack
- [x] Remove `gScenarioExpansionPacks` (no longer used)
- [x] Export to .park in crash handling
- [x] Export to .parkrep in crash handling
- [x] Handle packing of objects (dat, json, parkobj)
- [x] Update replay to use .park
- [x] Allow using new _and_ old object identifiers
- [x] Handle new object types (station, surface, edge, music)
- [x] Reserve room for tile element owner
- [x] Fix scenery window to handle more scenery groups
  - [x] Use full width of window to show list items
  - [x] Fix incorrect title
- [x] Send park instead of sv6 over network (cheats now in .park data)
- [x] Splitting paths into path surfaces and path railings
  - [x] Update tile inspector
  - [x] Update plugin framework
  - [x] Use correct yellow bamboo supports
  - [x] Footpath surface objects
    - [x] JSON files
    - [x] Loading code in OpenRCT2
  - [x] Make old-style/legacy paths work as well.
  - [x] Paths without slope railings should not have railings underground
  - [x] Support colours: grey for space supports, yellow for bamboo supports
  - [x] Preview images for path surface and path supports
  - [x] Object Selection window:
    - [x] Preview for path surfaces and railings
    - [x] Tab images for path surfaces and railings
    - [x] Tooltips for legacy footpaths, surfaces and railings
    - [x] Window width
    - [x] Hide legacy footpaths from default view
    - [x] Require at least one regular path, one queue and one path railing selected. (And don't require selecting legacy paths.)
  - [x] Fix paths in track designs
  - [x] Reset footpath selection when loading a new game
  - [x] Path preview for yellow bamboo and grey space supports is broken (Ted)
  - [x] Loading SV6 files with custom objects is broken (Ted)
  - [x] Optionally: make the RCT1 space supports use the images from RCT2 + g1.dat in order to work for everyone (Michael)
- [x] Set researched ride types, entries and scenery automatically (build pre-researched list automatically)
- [x] Do we still need `gMapBaseZ`? Yes
- [x] Fix money32 / money64 issues
- [x] Replace `gScenarioTicks` with `gCurrentTicks`
- [x] Staff patrol areas
- [x] Fix read/write of notifications.
- [x] Call `scenario_remove_trackless_rides()` and `scenario_fix_ghosts()` again (or integrate them)
- [x] Miscellaneous tab in Scenery window is broken
- [x] Warn when trying to export JSON objects to TD6
- [x] Change guest history to real values
- [x] Scenario editor: Add pay park / pay rides
- [x] Fix mouse input on zoom x0.5 on half of the map (easy to repro on Electric Fields, left side)
- [x] Investigate string ID exhaustion
- [x] Fix park name always being Six Flags
- [x] Rename objects to `origin.type.name`.
  - [x] Rename space footpath surface to "red tarmac"
- [x] Remove loading of JSON legacy footpaths (DAT support for them should remain).
- [x] Fix raised circuit and art deco footpath railings (they should still have queue banners).
- [x] Allow railings to omit banners (needed for the invisible railings)
- [x] Include an invisible footpath surface

PRs to be merged first:
- [x] Refactor of remaining tile elements: https://github.com/OpenRCT2/OpenRCT2/pull/10866
- [x] Translate SV6 object offsets to OpenRCT2 ones #10881
- [x] Rework scenery handling #10922
- [x] Ride id (needed for more than 255 rides in a park) https://github.com/OpenRCT2/OpenRCT2/pull/14411

Crashes/faults to address:
- [x] News history gets corrupted

Limits to be raised in this PR:
- [x] 255 stations
- [x] 2047 small objects
- [x] 2047 large objects
- [x] 2047 wall objects
- [x] 2047 ride types
- [x] 255 combined path types
- [x] 255 banner objects
- [x] 255 path addition objects
- [x] 16777216 tile elements
- [x] 65535 sprites
- [x] 1000 rides in a park
- [x] 255 trains per ride (from 31) (#714)
- [x] 256 spawn points (up from 2)

Stuff to be left for another PR:
- [ ] More granular Z coordinates for tile elements (making them twice as granular allows for 100% RCT1 compatibility in this area).
- [ ] Splitting path elements and path additions

To be decided:
- [x] Map size (x and y are often saved in 8-bit unsigned variables, which will overflow)
- [x] Usage of old identifiers: https://github.com/OpenRCT2/OpenRCT2/issues/13080

Other stuff this PR tackles:
- [x] Making the Fruity Ices Stall colourable
- [x] Allow freezing a ride's ratings

Test notes:
- [x] Test by loading up an SV6, saving it to .PARK, then reloading that file.
- [x] Test placing track design (currently crashes in some cases)

Stuff for the changelog:
```
- Feature: [#10664] New save format with increased limits. (put this right on top!)
- Feature: [#714] Allow up to 255 trains per ride.
- Feature: [#2253] Path surfaces and railings can be mixed and matched, like in RCT1.
- Improved: [#10664] Invisibility status is now shown in the Tile Inspector.
```


src/openrct2/world/Entrance.cpp
```diff
@@ -218,6 +218,27 @@ void fix_park_entrance_locations(void)
...
+    tile_element_iterator it;
+    tile_element_iterator_begin(&it);
+    while (tile_element_iterator_next(&it))
+    {
+        auto entranceElement = it.element->AsEntrance();
+        if (entranceElement != nullptr && entranceElement->GetEntranceType() == ENTRANCE_TYPE_PARK_ENTRANCE
+            && entranceElement->GetSequenceIndex() == 0 && !entranceElement->IsGhost())
+        {
+            CoordsXYZD entrance;
+            entrance.x = it.x * 32;
```
Gymnasiast: Do a proper conversion here.

AaronVanGeffen:
With regards to the format itself, we should probably try to incorporate company ids in from the get go, even if we currently only support one company in-game. It would allow us to extend to different (competitive) forms of multiplayer games, eventually.

As for what this could look like, I'm thinking of Locomotion.

pizza2004:
Does this mean we might finally see #681 implemented?  It feels like I've been waiting forever for the new save format!  Not that I've been in any rush, it's fascinating just to watch you guys work at all and it's not like you're getting paid or like the save format is the top priority.

Gymnasiast:
@pizza2004 Not yet. I want to avoid scope creep and only remove a handful of limits initially.

Gymnasiast:
@AaronVanGeffen I'm not familiar with how Locomotion achieves this. could you be more specific?

Also note that there isn't much room in most tile elements - the Path element being the worst. Where do you propose saving this value?

pizza2004:
So is the goal here to make the new save format extensible in a way where removing limits won't break compatibility with older files?  I always assumed we'd want to determine all the new limits and such before settling on the new save format so it wouldn't need to change again in the future.

pizza2004:
I imagine companies should own separate plots of land and everything on them, but alternatively you could tie it directly to rides and staff and peeps and the like.  I imagine the overall goal is to allow competitive multiplayer just however it makes the most sense in the context of Rollercoaster Tycoon.

Gymnasiast:
@pizza2004 Getting rid of limits involves more than just changing the save format.

In any case, this PR is not meant as a general Q&A thread, this is a discussion about getting the code ready for merging. If you have any questions, please just ask them on Gitter.

IntelOrca:
@AaronVanGeffen like @Gymnasiast said, with 16 bytes per tile element, some have room for company ID while others like footpath do not. Although we did suggest doing some like banners and have two tile elements to represent footpaths, the second being just for the additions (bins, benches).

This otherwise shouldn't affect the design of the save format as we would dump the tile element data out right, we have 8 bytes of spare bytes for each element at the moment.

src/openrct2/world/Surface.h
```diff
@@ -28,7 +28,12 @@ enum
...
     TERRAIN_SAND_DARK,
     TERRAIN_SAND_LIGHT,
-    TERRAIN_COUNT_REGULAR = 14, // The amount of surface types the user can actually select - what follows are technical types
+    TERRAIN_ROOF_BROWN,
+    TERRAIN_ROOF_LOG,
+    TERRAIN_ROOF_IRON,
+    TERRAIN_ROOF_GREY,
+
+    // Technical types (not selectable)
     TERRAIN_CHECKERBOARD_INVERTED = 14,
```
Gymnasiast: Check if this needs to be hardcoded to 14.

Gymnasiast:
@IntelOrca @ZehMatt The build currently fails on targets that have network disabled (like Android), because the new save calls Crypt::CreateSHA1(), which is hidden behind an ifdef (DISABLE_NETWORK). What would be the best way to fix it?

janisozaur:
How about we move to sha3 and use dedicated library for that, at least for the no network builds? We still need authentication for network and the reason they are separated is to ease porting, which doesn't require openssl in the initial phase

IntelOrca:
@Gymnasiast Now that Windows uses native APIs to do the cryptography we can enable the crypt header for all builds. For non-Windows builds you will have to enable crypt or at least the hashing part in all builds. Implementing a SHA1 on all platforms should be easy enough, you can probably find a small copy and paste solution to that as well.

SHA1 is only for a checksum, if we want something more trivial to implement we can, I would not choose something more complex though. This is just a checksum, not a security measure.

janisozaur:
I would prefer moving on to the current preferred standard even in such case, it doesn't seem that difficult, especially when we design a new format.

That said, SHA1 (and later) implementations are dime a dozen. Quick Google yields e.g. https://github.com/983/SHA1

IntelOrca:
You will most likely have to add our own implementation of SHA-3. I don't think the Windows API supports that, and I really don't want to re-introduce OpenSSL just for this.

ZehMatt:
We could use a simpler hash such as xxHash for the save file, there is no need for security here.

Gymnasiast:
Report about saving as .PARK and reloading leading to a crash: https://github.com/OpenRCT2/OpenRCT2/issues/10872

src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -219,13 +219,13 @@ static const char* getFilterPatternByType(const int32_t type, const bool isSave)
...
-            return isSave ? "*.sv6" : "*.sv6;*.sc6;*.sc4;*.sv4;*.sv7";
+            return isSave ? "*.park" : "*.park;*.sv6;*.sc6;*.sc4;*.sv4;*.sv7";
 
         case LOADSAVETYPE_LANDSCAPE:
-            return isSave ? "*.sc6" : "*.sc6;*.sv6;*.sc4;*.sv4;*.sv7";
+            return isSave ? "*.park" : "*.park;*.sc6;*.sv6;*.sc4;*.sv4;*.sv7";
 
         case LOADSAVETYPE_SCENARIO:
-            return "*.sc6";
+            return "*.park";
```
Gymnasiast:
```suggestion
            return isSave ? "*.park" : "*.park;.sc6;*.sc4";
```

src/openrct2/ride/Ride.h
```diff
@@ -38,7 +38,7 @@ struct Staff;
 #define DOWNTIME_HISTORY_SIZE 8
 #define CUSTOMER_HISTORY_SIZE 10
 #define MAX_CARS_PER_TRAIN 255
-#define MAX_STATIONS 4
+#define MAX_STATIONS 8 // Increase to 254 once the tile elements can handle it
```
Gymnasiast: 255

Umdlye:
I don't know if scenario stuff is supposed to be implemented yet, but just in case:

- Scenarios created in the new format don't show up in the scenario menu when placed in the `scenario` folder
- Loading the file directly doesn't bring up the objective screen like loading a .sc6 would
- The scenario editor doesn't require a peep spawn point to be set
- Scenario descriptions don't seem to save, the description becomes the one from Haunted Harbour regardless of what you enter
- Loading .park files through the "load landscape" option in the scenario editor gives the `File contains invalid data!` error

Slightly related:

- The default size of the object selection window needs to be bigger to prevent the new category buttons from overlapping the "filter" button

X123M3-256:
I just tried to build this, and I get an error "array index 1048576 is past the end of the array (which contains 196608 elements)" on line 1112 of map.c. I'm guessing this has to do with the raised object limit?

Umdlye:
The path selection menu doesn't seem to like the new limit yet 😛 

17-32 paths works as expected 🎉 
33 paths crashes the game when you open the path selection
34-35 paths show the paths in one long line instead of a square menu
36-47 paths crashes the game when you open the path selection
48-49 paths show a little bit of a menu on the side of the screen
50 paths show an empty menu
51 paths show them all in one line again
52-53 paths show no menu at all but doesn't crash
54-55 paths show them all in one line
56-58 show no menu at all
59 paths show them all in one line
60 paths works as expected 🎉 
61 paths show them all in one line

Didn't test 62-253 yet

Finally 254-255 paths crashes the game without making a dump when you open the path selection

Again, sorry if this is too early, it's just cool to see that it partially works already!

Gymnasiast:
@Umdlye Most likely caused by trying to beyond the confines of `gAppropriateImageDropdownItemsPerRow`, which is only 34 items long.

Umdlye:
Yay! May be worth noting that trying to any save made in 5cbc5f9 crashes the game in 585e68d. (Windows 10, x64)

Gymnasiast:
@Umdlye Assume that every single commit in this branch breaks your old .PARK saves. The format will only be stable and backwards compatible once it's released.

James103:
Will the new save format ever incorporate a save game version system (such as the one found in OpenTTD)? This should allow for better forward compatibility, especially when new features are added and limits are raised.

Gymnasiast:
@James103 Yes, but again: **this is not a general Q&A thread.**

X123M3-256:
Is saving of the object selection implemented? I noticed when trying to load a save created yesterday (with the same version that was used to create it), that the void terrain type was not selected and the terrain rendered incorrectly as a result. Not sure if this is a known issue or a possible bug.

Umdlye:
openrct2_2020-03-08_21-51-33

Looks like there aren't enough page name strings to fit the increased amount of guests.

Gymnasiast:
The guest window will be addressed here: https://github.com/OpenRCT2/OpenRCT2/pull/10901

Dropdowns can only have 128 items (up from 64), so that makes sense: 96 stations + 31 trains + overall view = 128 items. We should perhaps increase the maximum size, or even look into replacing it with another widget.

Umdlye:
The guest counter appears to overflow at 32767 guests as reported by Marcel Vos on Discord.

https://cdn.discordapp.com/attachments/356546837945253888/686322163065683982/Test_Park_2020-03-08_22-18-55.png
https://cdn.discordapp.com/attachments/356546837945253888/686327824650469572/unknown.png

Gymnasiast:
@Umdlye Could you test if changing `{COMMA16}` to `{COMMA32}` on strings 1846, 1847 and 1848 fixes that problem? You don't need to recompile anything, just changing the strings in en-GB.txt and/or nl-NL.txt (depending on the language you're using) is enough.

Umdlye:
Doing that turned -32,505 into 16,744,711 😬

duncanspumpkin:
@gymnasiast you will need to modify the callers of that sprite to push a 32bit value onto the stack as well to fix the peep count.

Gymnasiast:
@duncanspumpkin Since there are more occurrences of the guest count being used (like in the Park window), is it perhaps a better idea to change `gNumGuestsInPark` to a 32-bit value?

duncanspumpkin:
Ah sorry thought it worked differently and was manually pushing the value on the args stack. That was incorrect. Yes you will need to change the type to uint32_t

src/openrct2/world/Sprite.h
```diff
@@ -194,9 +194,9 @@ enum
 rct_sprite* try_get_sprite(size_t spriteIndex);
 rct_sprite* get_sprite(size_t sprite_idx);
 
-extern uint16_t gSpriteListHead[6];
-extern uint16_t gSpriteListCount[6];
-extern uint16_t gSpriteSpatialIndex[0x10001];
+extern uint16_t gSpriteListHead[SPRITE_LIST_COUNT];
+extern uint16_t gSpriteListCount[SPRITE_LIST_COUNT];
+extern uint16_t gSpriteSpatialIndex[MAX_SPRITES + 1];
```
duncanspumpkin: This is incorrect the spatial index should still be 0x10001

src/openrct2/world/Sprite.cpp
```diff
@@ -31,9 +31,9 @@ static rct_sprite _spriteList[MAX_SPRITES];
 
 static bool _spriteFlashingList[MAX_SPRITES];
 
-#define SPATIAL_INDEX_LOCATION_NULL 0x10000
+#define SPATIAL_INDEX_LOCATION_NULL MAX_SPRITES
 
-uint16_t gSpriteSpatialIndex[0x10001];
+uint16_t gSpriteSpatialIndex[MAX_SPRITES + 1];
```
duncanspumpkin: See below 0x10001 is correct

src/openrct2/world/Sprite.cpp
```diff
@@ -31,9 +31,9 @@ static rct_sprite _spriteList[MAX_SPRITES];
 
 static bool _spriteFlashingList[MAX_SPRITES];
 
-#define SPATIAL_INDEX_LOCATION_NULL 0x10000
+#define SPATIAL_INDEX_LOCATION_NULL MAX_SPRITES
```
duncanspumpkin: revert

Gymnasiast:
@Umdlye Ok, should now work properly - could you re-test it?

Umdlye:
Looks like it works everywhere except the guests window:

openrct2_2020-03-09_21-12-05
openrct2_2020-03-09_21-11-53

duncanspumpkin:
That will be caused by stuffing the count of guests into var_492 which is 16 bit.

src/openrct2/world/Sprite.h
IntelOrca: Nah, it should be `MAX_MAP_SIZE * MAX_MAP_SIZE + `

Gymnasiast:
@Umdlye One of my latest commits should fix the number of guests in the guest list. Could you test that?

Umdlye:
The overall guest count is now fixed :) The thoughts and actions counts are still broken though, not sure if this was meant to fix those too.
openrct2_2020-03-20_20-57-05
openrct2_2020-03-20_20-57-17

Gymnasiast:
@Umdlye Thanks! I forgot about the summarised actions/thought counts.

Gymnasiast:
@Umdlye I have just pushed a commit to fix the thoughts and actions counts. Could you retest it?
(I appreciate your help a lot, btw!)

Umdlye:
Looks good!
https://user-images.githubusercontent.com/11657101/77229191-b127f200-6b8c-11ea-9d2e-a9c3ad35db80.png
https://user-images.githubusercontent.com/11657101/77229192-b2f1b580-6b8c-11ea-8f3c-21c0b6d3647e.png
(No problem, glad I could help!)

Umdlye:
openrct2_2020-03-26_22-44-47

Umdlye:
Did some more testing, the dropdown disappears when there are more than 61537 guests. At 61537 the game just crashes when you try to open the dropdown menu.

James103:
@Umdlye Why exactly does the dropdown disappear above (and crash at) 61537 guests? Could it be another signed/unsigned mishap (since 61537 is congruent to -3999 modulo 65536)

duncanspumpkin:
@James103 you'll get bad lag opening the guest list if you increase the length of the names by much more. Would be more performant to just increase the number of first names.

duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/pull/11228 will be adding a little fix for balloon frame numbers to prevent them exceeding a byte in value. This can be reverted in the new save format. Balloon.cpp L50.

Fred-1044276:
Any information on what the new limit of scenery groups will be? In OpenRCT2, it's currently still capped at 19.

James103:
The cap of 19 scenery groups is due to interface limitations (width of scenery selection window for example). In order to increase this cap, the scenery selection UI would have to be redesigned to accommodate 20+ scenery groups (ideally at least 255).

Gymnasiast:
@James103 Not true. Please don't answer questions you don't know the answer to, it's only going to confuse people.

IntelOrca:
The scenery group limit is a limit with the SV6 save format. However if we do increase this limit, the interface will likely need tweaking to handle many more tabs. We may have to completely rethink the window design altogether.

src/openrct2/ParkFile.cpp
```diff
@@ -531,8 +531,8 @@ namespace OpenRCT2
                     cs.ReadWrite(ride.custom_name);
                     cs.ReadWrite(ride.default_name_number);
 
-                    cs.ReadWrite(ride.price);
-                    cs.ReadWrite(ride.price_secondary);
+                    cs.ReadWrite(ride.price[0]);
+                    cs.ReadWrite(ride.price[1]);
```
IntelOrca, line 535: This should write an array now.

Gymnasiast: Like so: `cs.ReadWrite(ride.price);` ?

Also: the rebase meant that using std::begin, end and size on gNewsItems does not work any more, and I'm not sure how to fix it - do you have any idea?

IntelOrca: Yes.

A shame, you would have to implement a begin and end method with iterators to sort it out. Or maybe you could return just pointers. Otherwise you have to iterate it with an index.

Gymnasiast:
@Ruedii Regular walls have only two images (or six if they can be put on a slope). Path railings have far more. Corners have inner railings, dead ends have shorter-than-regular railings, for example.

frutiemax:
I just pulled this branch, try to save the Magic Mountain scenario to a .park file and load it and the game logic freezes. I've debugged it and when I comment line 287 in GameState.cpp i.e. peep_update_all(), the game is no more freezed.

Gymnasiast:
I replaced corrupt elements in the last commit with an invisibility flag. I also updated the Tile Inspector to use it. It looks like this:

afbeelding

Umdlye:
Awesome!

There is a small typo in the code by the way: `AnyToggleInvisilibity`.

Gymnasiast:
Yeah, I think it would be better. Actually, I actually tried it, but it seemed like the sprite was not compatible - there were dark bands around it. Of course, we could solve that by adding a fixed sprite to g2.dat.

docs/save-format.md
```diff
@@ -0,0 +1,255 @@
...
+  expenditure-history:      money32[][]
+
+Inventions:
+  
+
+Tiles:
+  map-size:                 xy32
+  tile-elements:            tile-element[]
+
+Things:
```
duncanspumpkin, line 143: Rename to Entities

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1930,6 +1930,14 @@ class S4Importer final : public IParkImporter
             }));
         LoadObjects(ObjectType::ParkEntrance, std::vector<const char*>({ "PKENT1  " }));
         LoadObjects(ObjectType::Water, _waterEntry);
+        LoadObjects(
```
Gymnasiast, line 1933: Should be removed - has been done in develop in the meantime.

src/openrct2/peep/Peep.h
```diff
@@ -635,7 +635,7 @@ struct Peep : SpriteBase
     };
     ride_id_t InteractionRideIndex;
     uint16_t TimeInQueue;
-    uint8_t RidesBeenOn[32];
+    ride_id_t RidesBeenOn[32];
```
Gymnasiast, line 638: Port to develop.

duncanspumpkin: This is going to be a breaking change. I don't think it can be ported

Gymnasiast: Hm, that's annoying. I guess there are no peep changes planned for the immediate future?

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1981,7 +1981,7 @@ static void window_ride_construction_entrance_click(rct_window* w)
     else
     {
         gRideEntranceExitPlaceType = ENTRANCE_TYPE_RIDE_ENTRANCE;
-        gRideEntranceExitPlaceRideIndex = w->number & 0xFF;
+        gRideEntranceExitPlaceRideIndex = w->number % MAX_RIDES;
```
IntelOrca, line 1984: This doesn't make sense. The original code was being silly and using & 0xFF instead of a type cast. When this window becomes a class, I would expect a field like _rideIndex or something to be used.

Gymnasiast: I wasn't sure about that, so I figured I'd play it safe until the window gets refactored.

IntelOrca: Yeah, it would just cause confusion if it did wrap.

duncanspumpkin:
The money64 changes need a good bit of additional work to them. `gParkValueHistory` needs updating to 64. Need to sort out the formatter.

src/openrct2/core/Crypt.OpenRCT2.cpp
```diff
@@ -7,190 +7,95 @@
...
+        {
+            // We have remainder, so fill rest of it with bytes from src
+            auto fillLen = sizeof(uint64_t) - _remLen;
+            assert(_remLen + fillLen <= sizeof(uint64_t));
+            std::memcpy(_rem + _remLen, src, fillLen);
+            src = reinterpret_cast<const uint64_t*>(reinterpret_cast<const uint8_t*>(src) + fillLen);
+            _remLen += fillLen;
+            dataLen -= fillLen;
+            ProcessRemainder();
+        }
```
ZehMatt, line 65: Why is this needed? The loop of mine copies just the bytes into uint64_t, if its less than 8 bytes its null padded.

IntelOrca: Would your code cope with:
  Update(5 bytes)
  Update(3 more bytes)

and give same result as if you did Update(all 8 bytes)?

ZehMatt: No. A lot of hash algorithms that use blocks will not do this.
https://godbolt.org/z/1875K31nG

There is also really no reason to have this, this is not a cryptographically secure hash, they also can have different schemas.

IntelOrca: > There is also really no reason to have this

Sorry I disagree with this. The API is so that you can stream the data into the hash algorithm. It should not give you different results depending on the size of the chunks you feed it.

ZehMatt: How would you implement lets say Murmur2 like that?

IntelOrca: Exactly the same way.

ZehMatt: I mean okay, it doesn't matter too much. The thing that sticks out the most is reinterpret_cast, you should avoid it when possible, since we are not on C++20 yet there is no std::bit_cast so the std::memcpy will have to do.

Umdlye:
Early observations from 27e83dc:
* Invention list changes in the scenario editor/through the cheat are not saved properly. Only the 'pre-invented' selection is saved. The 'to invent' selection is not, and the list there is duplicated multiple times
* Loading a .park scenario as a saved game does not bring up the objective window like loading a .sc4/.sc6 would

LordMarcel:
The sprite limit has been raised and seems to work fine, but the staff limit is still 200. According to Gymnasiast this is an oversight and he told me to report it here. Amazing work on all this stuff by the way.

oli414:
The sign limit appears to still be 200, got the message for too many banners on a 14x14 section of path with banners placed in 2 directions:
image

nottud:
image

Corruption on the main menu.

nottud:
image

Corruption on main menu

nottud:
image

May be issue exclusively for RCT1 scenario files?

nottud:
image

Seems to also occur on newly placed path too.

nottud:
Least one ride tested does not take advantage of the new max limit of vehicles by default. It currently does 32 but can be raised to 58 (Cannot raise above this due to max platform length limit). Possible situation is similar on other rides.

Richard-L:
We noticed some sprite culling difficulties with RCT1 parks. This Top Spin one was reported by https://github.com/Xtreme97-RCT.

Park link: https://www.nedesigns.com/park/152/harakiris-islands-of-adventure/

* [ ] Sprite culling
RCTLL:
image
OpenRCT2, v0.3.3 (417e5ae on develop):
image
NSF build from last night:
image

giraty:
- [ ] A bit of issue with rides with wooden style base supports 
[img]https://cdn.discordapp.com/attachments/691736850980667514/840615182022738010/unknown.png[/img]

[img]https://cdn.discordapp.com/attachments/691736850980667514/840615231884361738/unknown.png[/img]
- [ ] Don't know if this is intended or not, but new type scenario has their own category in Other Parks
image

Krutonium:
Profit
Seems to be on at least RCT1 Scenarios.

Krutonium:
After saving and reloading that same park; 
Research

[Forest Frontiers.zip](https://github.com/OpenRCT2/OpenRCT2/files/6446503/Forest.Frontiers.zip)

Fred-1044276:
- [ ] Path additions from other scenery groups (e.g. Mine Themeing, Jurassic Themeing, Roaring Twenties Themeing) do not show up in any of the default scenery groups (Trees, Shrubs and Ornaments, Gardens, Fences and Walls, Walls and Roofs, Signs and Items for Footpaths) at all:
Metropolis 2 2021-05-08 13-50-48

loweml:
I've found an issue with the average cash guests come in with. I've tested it with the most recent version that doesn't use the new save format and didn't have this problem. When guests come into the park though, their initial cash is as it should be. It only seems to be an error in the Scenario options menu. Not a problem at all, at least so far, but I just thought you should know about it. 

9594b1723416a6037b916773d8c76193
e43b3af725f77da6c7a1c89a3be41ade

simaticable:
Woodworm Park has a locomotive instead of the tenders on the trains.

keiyakins:
- [ ] Incorrect scenario marked complete in scenario list on completion of game loaded from .park save. Tested with Trinity Islands and Katie's World saves. TI marked Six Flags Magic Mountain, then itself. Katie's World marked Funtopia then Aqua Park. Will test an RCT2 scenario now...

LurkerRCT:
That title screen thing would make sense, every scenario that can be finished on my game already is, and my scenario was a small one with a short goal so what might have happened was it tried to mark one of those and it had a higher company value than the one I played, so no name option.

Gymnasiast:
- [ ] Restricting scenery on RCT1 scenarios also hides items that attach to a scenery group, like rct1.wfw0.

esojka:
Since the elimination of the object limits is a request that is like 100 years old, you know the first thing people are going to do will be to select-all on the Object Selection screen.  A couple of issues arise in the GUI when a user does this:

- [ ] The existing filters help in ths regard, but it would be nice to have a "Select All" on the selection screens to avoid lots of clicking to add multiple items.
- [ ] Once multiple scenery groups are selected, if the number of scenery groups exceeds the width of the Add Scenery window, only as many as can be displayed are shown, with the individual objects from the other SGs placed into the "?" tab.

esojka:
Using the eyedropper tool in the scenery selection screen causes a crash with this build

Fred-1044276:
- [ ] It is impossible to scroll through the Recent Messages window in the .park test build. The scrollbar refuses to budge.
- [ ] money64 issue in the water tool.

captnpineapple:
- [ ] > Using the eyedropper tool in the scenery selection screen causes a crash with this build

Adding checkbox - as I encounter this error too, and the checkbox was not on the last person's reply

AWildShinx:
image
It seems generator extremely obsessed with arborginal art pieces 
Reproduction steps: 
1. Select all objects or choose arborginal art piece 
2. Generate terrain with trees
3. You should see many arborginal art pieces there

Gymnasiast:
@AWildShinx That also affects develop: https://github.com/OpenRCT2/OpenRCT2/issues/14155

StrangeAnomoly:
- [ ]  > Using the eyedropper tool in the scenery selection screen causes a crash

ridetester101:
New here so not too sure of where to feedback stuff yet: just posted some related here

https://github.com/OpenRCT2/OpenRCT2/issues/14663

Two other thoughts- as paths and footpath surfaces are two different things are there any potential pitfalls here with custom paths vs footpath surfaces plus railings, similarly will sharing files be a problem as it could be a lot of new objects in one go for the other player.

Love the idea of this update- it will be worth the headaches getting it working (I'm not a coder but could think and help with logical input such as what might be good for quality of life improvements etc as handling lots of objects is great for creativity in game but a massive headache in terms of time and having to scroll through 000s of items which may or may not end up in your game, so fixing some of these things could also be a chance to add some quality of life elements such as scenery management)

One would be being able to select too many items whilst making your choices to then whittle the long list down, thus a sort of in-progress landscape editor- at this early stage it would just need to save which choices are on/off and have a state where you can't progress until you've cut any excess, eg you may have say 120 too many so you can go back later and see what you can cut - reason is with so many choices now this is a big burnout as it's kind of designed to have to do it in all one go currently which is too much, so a quality of life workaround to this would be great, as would others like being able to define your own tags/or and groups, say trains or cars or sci fi etc, but that's a bit wishful thinking for now, one for when other stuff is resolved.

Thanks for the epic work on this so far and keeping the game alive in 2021.

ridetester101:
Re esojka's screenshot of blanks in scenery box I had some like that with a small mystery image and then the bottom of the long scrolling previews were totally blank- some preview images have problems so you got some blanks etc even before this limit-breaking update, it may be designers setting up preview images wrong or preview images going a bit awol or some other problem with specific scenery items and how they were set up, again before this limit-breaking there were still problems with scrolling selected items- you could pick scenery items and paths that still didn't appear in game, possibly as the scrolling windows had run out of slots/ something else capacity related. Happens in some other older games on the official releases- one, maybe locomotion, let you select more available trains than would actually appear in the vehicle selection window.

So any limit breaking also needs to match what will/won't work within any in-game limits. Plus of course working out what those limits are and how to work most efficiently within them/ if they can be worked around.

LordMarcel:
Two more bugs here:

Normal path without railings is invisible. Queue lines work fine.
Screenshot_353

When building a coaster vertically into the ground the underground track sometimes doesn't get rendered.
Screenshot_352

Gieneq:
image
Very messy. i7-9750HF, RTX 2060 Win10

Gymnasiast:
@Gieneq Not a bug. You haven't linked your RCT1 install properly.

jmciver92:
Im not sure if this matter but I am testing on the MacOS build.

- [ ] Screen Crash when using the Eye-Dropper Tool

ridetester101:
Have run some further tests in scenario and landscape editor- does seem broken- made various attempts to use the Blackpool Pleasure Beach scenario, all ultimately failed

First related issue is replacing original files with the .park.sv6 file- this in itself is problematic as if the .park.sv6 file has issues you can't go back to the original so means a lot of multiple backing up/ repeating and the .park.sv6 files it seems are sometimes unloadable- of course it could be that they shouldn't have the double extension and just be .park depending how you've implemented the new save format, but due to issues it would be worth preserving the original file rather than replacing it if it is a non-park file.

First tried loading the park from a save to convert a saved park to a scenario- worked and loaded into scenario editor but as soon as I saved it the problems began- would not then reload anything when I tested exiting and seeing if could reload into editor (just got the blank empty map again), interestingly even after making the save again from the game and then trying to import the save game into the scenario editor I got nothing (maybe for now we also need option to save non-park games as either park format or original save format, it seems to be the park format it dislikes) so weirdly even the thing that had worked in test 1 failed in test 1b, the repeating of test 1.

Tried various things a few more times and eventually through the load landscape method got the landscape without the rides successfully, the entrances and everything seemed fine with pathing to map edge but again after saving it and exiting it did not want to reload the new park file into the editor- wanted to test if it works if you do the next steps like invention list setup but it has since refused to reload the park either as just landscape or with rides, seems the park format is causing a problem the editor can't handle.

Not sure if this being an RCT1 scenario originally is any part of the issue but it is similar to other problems I had trying a few things out with the editor, somewhere along the lines you seem to hit a point where you can't load files previously saved in the editor.

One thing I noticed is the create scenario from saved park option has no 'set up the objectives' bit- just skips to save scenario as the first 'next' option after loading park, IF you can get a landscape only to load from such a save it does have the correct next step of invention list setup, will try a few more experiments, including for rct2 parks that aren't real parks in case there are issues with either rct1 parks or real parks.

ridetester101:
Scenario/landscape editor test updates.

Found even though editor refuses to reload them (sc6 files) the game will through the load saved game tab so the data it seems is in tact if can fix editor and other issues, when checking in game:

Bottom of item selection window cuts off at higher zoom (window scale) levels in options

It will load a park where guest spawn path doesn't reach edge of map but this will have a glitched out of park path area very full of peeps- still works and didn't crash but wonder if it would eventually fail. If a design queue wasn't set it randomizes one. Positive as lots of attractions= can get innovations over decades so could make for a long and satisfying game, or test players patience! Invented shops etc were getting added and available and worked. Not sure if any were blanks as only checked a few.

Some blank rides could be placed and were some of the ones with pre-built layout/ design your own layout choices, but there could still be glitches as didn't check if these blank ones worked properly. Some had multiple options for flat rides but blank ride build window options were unavailable after placing.

Not available in next point is ones showing blanks or rides with unavailable vehicle options- didn't pick up all of these which should explain my only accounting for 670 attraction types. These tend to just plonk down nothing, a triangle, or a flat ride base with no ride, usually glitched, but possibly allowing a different flat ride to be selected in the ride's options menu.

1560 available scenery etc pieces versus 3122 selected scenery etc pieces, no missing scenery groups but had only selected six, missing items were not just in misc tab though so these other tabs had some missing too

Approx 492 available attractions with 178 not available which were often glitchy if placed (=670) out of 718 selected

MORE INFO TO FOLLOW - WILL BE BACK LATER (run out of time for now)

Gymnasiast:
@jmciver92 That has been reported several times already. Also, Windows != PC.

Gymnasiast:
@ridetester101 Can you add checkboxes to your points like everyone else? Makes it easier for us to check points that have been fixed.

KevsterW21:
Ive come across an issue in this version where the rides "Chase" camera doesnt seem to work anymore. Ive installed this version fresh & havent had any issues with any other game versions of this not working. 
Openrct2 Screenshot 2021 05 17 - 03 40 47 21

Gymnasiast:
@KevsterW21 I think that was already reported here: https://github.com/OpenRCT2/OpenRCT2/pull/10664#issuecomment-841208899

src/openrct2/ParkFile.cpp
```diff
@@ -0,0 +1,2091 @@
...
+            });
+            if (!found)
+            {
+                throw std::runtime_error("No tiles chunk found.");
+            }
+        }
+
+        void ReadWriteBannersChunk(OrcaStream& os)
+        {
+            os.ReadWriteChunk(ParkFileChunkType::BANNERS, [this, &os](OrcaStream::ChunkStream& cs) {
```
Gymnasiast, line 840: `this` is unused. Clang chokes on this.

cuyama:
The scenery eyedropper causes a crash when I try to pick an object. As far as I can tell this is independent of the object you're trying to select, map size, and whether the current park was loaded from an SV6 or a .park.

Fred-1044276:
- [ ] Attempting to start up any preexisting .sc6, .sv6, or .park file containing RCT2 Time Twister footpaths will fail (confirmed on the newest build from the "[Actions](https://github.com/OpenRCT2/OpenRCT2/actions)" hub). This is because the game is still using old object names for the RCT2 Time Twister footpath railings. The scenario I tried to open is Mythological - Cradle of Civilization 2 (Mythological Madness 2):
Vertigo Vale 2021-05-20 17-24-33

Fred-1044276:
- [ ] Guests' time in park stats are imported incorrectly when starting up a scenario or loading up a save game that already has guests in the park:
Luminous Landscapes 2021-05-21 10-23-47
Animatronic Antics 2 2021-05-21 10-35-45

As you can see, in the top picture, Guests 1 and 12 are claimed to have spent 1,085 h 24 m and 1,085 h 23 m, respectively, in the park, which just looks wrong (the code apparently added almost 65,536 minutes to the time in park stats of all guests in the park). The picture below shows Mythological - Animatronic Film Set 2 (Animatronic Antics 2), with Guests 6537 and 8130 having time in park stats of 1,018 h 6 m and 1,016 h 16 m, respectively. If I start up the same scenarios in the current develop or release builds, the issue doesn't occur as Guests 1 and 12 have correctly spent 13 and 11 minutes in the Luminous Landscapes park, respectively, and Guests 6537 and 8130 have correctly spent 2 h 9 m and 19 m in Animatronic Antics 2, respectively.
Luminous Landscapes 2021-05-21 10-30-33
Animatronic Antics 2 2021-05-21 10-34-59

- [ ] In the build scenery window, any scenery objects that appear in **two** different scenery groups, one of them usually being a default scenery group (Trees, Shrubs and Ornaments, Gardens, Fences and Walls, Walls and Roofs, or Signs and Items for Footpaths), do not appear in any of the default scenery groups if their unique scenery groups are available to use (Giant Candy and Wonderland Themeings are unaffected though). This is because all affected scenery objects' json files have a "sceneryGroup" flag which doesn't work correctly in current NSF build:
Animatronic Antics 2 2021-05-21 10-38-33
Animatronic Antics 2 2021-05-21 10-38-38

On develop and release, the "sceneryGroup" flags function properly.

Apologies for having reported these issues earlier, but I wanted to report them again with some clarification this time around.

Fred-1044276:
- [ ] Scenarios with park value objectives are shown with an incorrect end date on the scenario selector screen. Top pic shows a park value goal of $550,000 at the end of October, Year **11** and was taken in the current NSF test build. Bottom pic was taken in current develop build and shows the same park value goal but with the correct end date of October, Year **3**.
Six Flags Magic Mountain 2021-05-21 11-13-58
Six Flags Magic Mountain 2021-05-21 11-14-39
This affects all scenarios with a "Park value at a given date" objective.

Porthor:
- [ ] Completing the park Rainbow Valley (Pulled from the Deluxe version of RCT1) checks off Haunted Harbor.

2021-05-21 21_14_59-Start

ZeeMajora:
Fruity ice cream stalls are imported with colors when importing .sc6 scenarios and .sea (rct classic) scenarios.
Tested .sc6 with ayer's rock and six flags over texas.
Tested .sea with gentle glen and haunted harbor.
Curiously the rct classic scenarios i tested still seem to have rct1's color data present for these stalls so it may be a good idea to exclude them from being imported with light blue paint as the colors are the same as they are in rct1's equivalent scenarios.
DeepinScreenshot_openrct2_20210523214941
DeepinScreenshot_openrct2_20210523215402
DeepinScreenshot_openrct2_20210523214425
DeepinScreenshot_openrct2_20210523214327

Gymnasiast:
**To everyone: since this issue has become hard to go through, we are changing the way bugs are reported. You should now simply create an issue as you would for a regular bug, making sure to mention it is in the new-save-format branch.

src/openrct2/ride/Ride.cpp
```diff
@@ -2326,6 +2330,7 @@ static void ride_breakdown_update(Ride* ride)
 {
     if (gCurrentTicks & 255)
```
AaronVanGeffen, line 2331: MAX_RIDES?

ZehMatt: I wanted to keep the replays functional, its also not 100% clear what this is supposed to achieve, it ticks updates 254 times leaving only the first tick out. In another test run the vehicles got stuck mid way for some reason, so its for another time.

src/openrct2/ride/Ride.cpp
```diff
@@ -2078,8 +2078,12 @@ void Ride::Update()
 
     // Various things include news messages
     if (lifecycle_flags & (RIDE_LIFECYCLE_BREAKDOWN_PENDING | RIDE_LIFECYCLE_BROKEN_DOWN | RIDE_LIFECYCLE_DUE_INSPECTION))
-        if (((gCurrentTicks >> 1) & 255) == static_cast<uint32_t>(id))
+    {
+        // Breakdown updates are distributed, only one ride can update the breakdown status per tick.
```
duncanspumpkin, line 2082: But why. That seems like a bad idea. Perhaps if we did `(gCurrentTicks & 255) == (rideId & 255)` then it does not change anything balance wise from the base game but allows for the new limits

duncanspumpkin, line 2082: But why. That seems like a bad idea. Perhaps if we did `(gCurrentTicks & 255) == (rideId & 255)` then it does not change anything balance wise from the base game but allows for the new limits

src/openrct2/ride/Ride.cpp
IntelOrca: I am not sure 255 is representing `MAX_RIDES` here, it is just trying to make this method run once per 256 ticks. Similar to the every 8192 ticks below.

ZehMatt: It runs whenever value > 0, but I can see how one can be fooled by this line, it got me too.

src/openrct2/ride/Ride.cpp
ZehMatt: I think the reason it was built like this is to avoid having news popups overwriting each other. If I wrap the ride id its possible that two or more rides breakdown each adding the news entry in the same tick where normally the distance of breakdowns is delayed by ride id distance.

Porthor:
- [ ] Top Spin vehicle renders on top of everything. Rides, scenery, ect.

2021-05-26 22_09_18-OpenRCT2

esojka:
Elevator ride is wonky in this build - Elevators require multiple stations.  If you build second/third/etc. stations and then add  entrance/exit to each, all previously built stations disappear.  Build entrance/exit on first station, then close construction window, that E/E stays.  Go back into construction mode, build subsequent E/E all of them disappear.

src/openrct2/object/FootpathObject.cpp
```diff
@@ -100,6 +100,7 @@ void FootpathObject::ReadJson(IReadObjectContext* context, json_t& root)
                 { "hasSupportImages", RAILING_ENTRY_FLAG_HAS_SUPPORT_BASE_SPRITE },
                 { "hasElevatedPathImages", RAILING_ENTRY_FLAG_DRAW_PATH_OVER_SUPPORTS },
                 { "editorOnly", FOOTPATH_ENTRY_FLAG_SHOW_ONLY_IN_SCENARIO_EDITOR },
+                { "noSlopeRailings", FOOTPATH_ENTRY_FLAG_NO_SLOPE_RAILINGS },
```
Gymnasiast, line 103: Can be taken out, as we never use this for legacy footpaths.

src/openrct2/drawing/ImageImporter.h
```diff
@@ -45,13 +45,17 @@ namespace OpenRCT2::Drawing
             RLE = 1 << 1,
         };
 
+        ImportResult Import(
+            const Image& image, int32_t srcX, int32_t srcY, int32_t width, int32_t height, int32_t offsetX, int32_t offsetY,
```
Gymnasiast, line 49: Using ScreenCoords here might be better.

src/openrct2/FileClassifier.cpp
```diff
@@ -62,6 +69,29 @@ bool TryClassifyFile(OpenRCT2::IStream* stream, ClassifiedFileInfo* result)
...
 }
 
+static bool TryClassifyAsPark(OpenRCT2::IStream* stream, ClassifiedFileInfo* result)
+{
+    bool success = false;
+    uint64_t originalPosition = stream->GetPosition();
+    try
+    {
+        auto magic = stream->ReadValue<uint32_t>();
+        if (magic == 0x4B524150)
```
Gymnasiast, line 79: This needs a constant.

src/openrct2/peep/Guest.cpp
```diff
@@ -7459,6 +7461,6 @@ void Guest::RemoveRideFromMemory(ride_id_t rideId)
         // Last slot is now free.
         auto& lastEntry = Thoughts.back();
         lastEntry.type = PeepThoughtType::None;
-        lastEntry.item = PEEP_THOUGHT_ITEM_NONE;
+        lastEntry.argument = 0;
```
duncanspumpkin, line 7464: Should be std::numeric_limits<uint32_t>::max() to match L6800

src/openrct2-ui/windows/MapGen.cpp
```diff
@@ -818,7 +817,7 @@ static void window_mapgen_simplex_mouseup(rct_window* w, rct_widgetindex widgetI
             mapgenSettings.mapSize = _mapSize;
 
             mapgenSettings.height = _baseHeight;
-            mapgenSettings.water_level = _waterLevel + 2;
+            mapgenSettings.water_level = _waterLevel + MINIMUM_WATER_HEIGHT;
```
Gymnasiast, line 820: The changes in this file (and similar ones) should be backported.

src/openrct2/world/Footpath.h
```diff
@@ -49,6 +44,23 @@ struct rct_footpath_entry
     RailingEntrySupportType support_type; // 0x0A
     uint8_t flags;                        // 0x0B
     uint8_t scrolling_mode;               // 0x0C
+
+    constexpr uint32_t GetQueueImage() const
```
Gymnasiast, line 48: Parts of the footpath refactor could also be backported.

src/openrct2-ui/windows/MapGen.cpp
Broxzier: This was introduces in https://github.com/OpenRCT2/OpenRCT2/pull/14700, when raising the water height limit. This commit can safely be cherry-picked: 6a64b5a63cf24d7a9baf8f398445d341fc65c179

Gymnasiast: OK, can you explain the usage `MINIMUM_WATER_HEIGHT + MAXIMUM_WATER_HEIGHT` to me? It’s in a place where I would expect it to just read `MAXIMUM_WATER_HEIGHT`.

Broxzier: Good question, I was thinking the same thing yesterday. The game originally stores the water level in 5 bits, each step increases the base height by 2, and 0 means there is no water at all. This leaves us with 31 usable water levels. Since in the original games, surface element cannot have a base height under 2, a minimum water level of the same base height (value 1) would be useless. Instead of wasting this value, Chris Sawyer decided to add a minimum water level which essentially is an offset. With this offset, we gain one additional usable water level while still only using 5 bits.

The naming of `MAXIMUM_WATER_HEIGHT` is a bit confusing, as it's not the absolute height, but the maximum offset from the minimum water level.

Broxzier: | Value | Water surface height |
| --- | ---
| 0 | None |
| 1 | 4 |
| 2 | 6 |
| 3 | 8 |
| ... | ... |
| 29 | 60 |
| 30 | 62 |
| 31 | 64 |

Gymnasiast: I see. That sounds like a good candidate for refactoring - the offset is a detail that should only be known to the surface tile element itself. I might create an issue for that.

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -631,14 +657,16 @@ bool IsObjectCustom(const ObjectRepositoryItem* object)
 
     // Do not count our new object types as custom yet, otherwise the game
     // will try to pack them into saved games.
-    auto type = object->ObjectEntry.GetType();
-    if (type > ObjectType::ScenarioText)
+    if (object->Type > ObjectType::ScenarioText)
```
Gymnasiast, line 660: Can be removed now we no longer export to SV6.

src/openrct2/paint/tile_element/Paint.Entrance.cpp
```diff
@@ -14,6 +14,8 @@
 #include "../../drawing/LightFX.h"
 #include "../../interface/Viewport.h"
 #include "../../localisation/Localisation.h"
+#include "../../object/FootpathObject.h"
```
Gymnasiast, line 17: revert

src/openrct2/actions/RideCreateAction.cpp
```diff
@@ -297,7 +297,13 @@ GameActions::Result::Ptr RideCreateAction::Execute() const
     ride->income_per_hour = MONEY64_UNDEFINED;
     ride->profit = MONEY64_UNDEFINED;
     ride->connected_message_throttle = 0;
-    ride->entrance_style = 0;
+
+    ride->entrance_style = OBJECT_ENTRY_INDEX_NULL;
+    if (rtd.HasFlag(RIDE_TYPE_FLAG_HAS_ENTRANCE_EXIT))
+    {
+        ride->entrance_style = gLastEntranceStyle;
```
ZehMatt, line 304: I think it might be better to pass this as a parameter to the game action, this global may be different on other clients, I haven't looked  where this is assigned or when.

src/openrct2/interface/Window.h
```diff
@@ -318,19 +316,6 @@ struct map_variables
     uint16_t var_488;
 };
 
-struct ride_variables
```
duncanspumpkin, line 321: These should be reverted

src/openrct2/actions/LargeSceneryPlaceAction.cpp
```diff
@@ -220,6 +221,12 @@ GameActions::Result::Ptr LargeSceneryPlaceAction::Execute() const
 
     res->Position.z = maxHeight;
 
+    if (!CheckMapCapacity(sceneryEntry->tiles, totalNumTiles))
```
duncanspumpkin, line 224: These changes can be reverted. Its checked in the query.

src/openrct2/core/IStream.cpp
```diff
@@ -57,6 +54,17 @@ namespace OpenRCT2
         }
     }
 
+    void IStream::WriteString(std::string_view str)
```
duncanspumpkin, line 57: This can be backported.

src/openrct2/core/MemoryStream.cpp
```diff
@@ -262,6 +262,12 @@ namespace OpenRCT2
         Write<16>(buffer);
     }
 
+    void MemoryStream::Clear()
```
duncanspumpkin, line 265: backport.

distribution/openrct2.d.ts
```diff
@@ -517,7 +519,7 @@ declare global {
         result: boolean;
     }
 
-    type VehicleCrashIntoType = "another_vehicle" | "land" | "water";
+    type VehicleCrashIntoType = "another_vehicle"|"land"|"water";
```
Gymnasiast, line 522: Formatting changes in this file should be reverted.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -781,7 +781,7 @@ static std::unique_ptr<TrackDesign> _trackDesign;
 // Cached overall view for each ride
 // (Re)calculated when the ride window is opened
 struct ride_overall_view {
-    int16_t x, y, z;
+    int32_t x, y, z;
```
Gymnasiast, line 784: Can be backported.
Better yet, changed to CoordsXYZ or TileCoordsXYZ.

duncanspumpkin: yeah doing that atm.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1278,15 +1278,12 @@ static int32_t cc_show_limits(InteractiveConsole& console, [[maybe_unused]] cons
...
 
-    int32_t staffCount = GetEntityListCount(EntityType::Staff);
-
     auto bannerCount = GetNumBanners();
 
     console.WriteFormatLine("Sprites: %d/%d", spriteCount, MAX_ENTITIES);
     console.WriteFormatLine("Map Elements: %zu/%d", tileElementCount, MAX_TILE_ELEMENTS);
     console.WriteFormatLine("Banners: %d/%zu", bannerCount, MAX_BANNERS);
     console.WriteFormatLine("Rides: %d/%d", rideCount, MAX_RIDES);
-    console.WriteFormatLine("Staff: %d/%d", staffCount, STAFF_MAX_COUNT);
```
duncanspumpkin, line 1289: revert. The staff count is still valid but the max is max_entities

src/openrct2/config/Config.cpp
```diff
@@ -475,7 +475,7 @@ namespace Config
         }
     }
 
-    static void ReadWriteNotificationsChunk(IIniWriter* writer)
+    static void WriteNotificationsChunk(IIniWriter* writer)
```
duncanspumpkin, line 478: This is still not fully reverted. Should drop the Chunk part of the change.

src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -566,11 +566,7 @@ class StaffListWindow final : public Window
 
             if (isPatrolAreaSet)
             {
-                if (gStaffModes[peep->StaffId] != StaffMode::Patrol)
-                {
-                    continue;
-                }
-                if (!peep->IsLocationInPatrol(footpathCoords))
+                if (!peep->HasPatrolArea() || peep->IsLocationInPatrol(footpathCoords))
```
duncanspumpkin, line 569: IsLocationInPatrol has been inverted. Causes #14824

src/openrct2/Context.cpp
```diff
@@ -495,7 +495,6 @@ namespace OpenRCT2
 #endif
             }
 
-            gCurrentTicks = 0;
```
Gymnasiast, line 498: Needs to be reverted, I think.

src/openrct2/ride/transport/Monorail.cpp
```diff
@@ -460,7 +460,6 @@ static void paint_monorail_station(
 {
     uint32_t imageId;
     const StationObject* stationObject = nullptr;
-
```
Gymnasiast, line 463: revert

src/openrct2/ride/thrill/TopSpin.cpp
```diff
@@ -44,7 +44,7 @@ static int8_t TopSpinSeatPositionOffset[] = {
  */
 static void top_spin_paint_vehicle(
     paint_session* session, int32_t al, int32_t cl, const Ride* ride, uint8_t direction, int32_t height,
-    const TrackElement& tileElement)
+    const TrackElement& trackElement)
```
Gymnasiast, line 47: revert

src/openrct2/object/ImageTable.cpp
```diff
@@ -214,7 +237,7 @@ std::vector<std::unique_ptr<ImageTable::RequiredImage>> ImageTable::LoadObjectIm
         // Log place holder information
         if (placeHoldersAdded > 0)
         {
-            std::string msg = "Adding " + std::to_string(placeHoldersAdded) + " placeholders";
+            std::string msg = "PAdding " + std::to_string(placeHoldersAdded) + " placeholders";
```
Gymnasiast, line 240: revert

src/openrct2/localisation/StringIds.h
```diff
@@ -3896,6 +3893,16 @@ enum
...
+    STR_AT_LEAST_ONE_FOOTPATH_NON_QUEUE_SURFACE_OBJECT_MUST_BE_SELECTED = 6441,
+    STR_AT_LEAST_ONE_FOOTPATH_QUEUE_SURFACE_OBJECT_MUST_BE_SELECTED = 6442,
+    STR_AT_LEAST_ONE_FOOTPATH_RAILING_OBJECT_MUST_BE_SELECTED = 6443,
+    STR_OBJECT_SELECTION_FOOTPATH_SURFACES = 6444,
+    STR_OBJECT_SELECTION_FOOTPATH_RAILINGS = 6445,
+    STR_TILE_INSPECTOR_FOOTPATH_SURFACE_NAME = 6446,
+    STR_TILE_INSPECTOR_FOOTPATH_RAILINGS_NAME = 6447,
+
+    STR_UNSUPPORTED_OBJECT_FORMAT = 6448,
+
```
Gymnasiast, line 3905: STR_AT_LEAST_ONE_PATH_OBJECT_MUST_BE_SELECTED is no longer used and should be removed as well.

src/openrct2/Context.cpp
ZehMatt: GameState::InitAll takes care of this.

src/openrct2/ride/Ride.cpp
ZehMatt: I did not notice the return, what Ted says.

src/openrct2/GameState.cpp
```diff
@@ -58,6 +58,7 @@ void GameState::InitAll(int32_t mapSize)
     gInMapInitCode = true;
     gCurrentTicks = 0;
 
+    gCurrentTicks = 0;
```
ZehMatt, line 61: This one can be be deleted.

src/openrct2/ride/Ride.h
```diff
@@ -232,7 +233,7 @@ struct Ride
     uint8_t proposed_num_cars_per_train;
     uint8_t max_trains;
 
-private:
+public: // private:
     uint8_t min_max_cars_per_train;
```
Gymnasiast, line 237: This variable needs to be either split so that it can actually properly save the two, or removed altogether.

src/openrct2/core/IStream.cpp
```diff
@@ -12,9 +12,6 @@
 #include "../object/Object.h"
 #include "Memory.hpp"
 #include "String.hpp"
-
-#include <vector>
```
duncanspumpkin, line 16: revert this as everyone seems to enjoy having it here.

src/openrct2/ParkFile.cpp
```diff
@@ -0,0 +1,4594 @@
...
+
+        if (os.GetHeader().TargetVersion <= 1)
+        {
+            return;
+        }
+
+        cs.ReadWrite(entity.AssignedStaffType);
+        cs.ReadWrite(entity.MechanicTimeSinceCall);
+        cs.ReadWrite(entity.HireDate);
+        cs.ReadWrite(entity.StaffId);
```
janisozaur, line 2010: Is this field still present?

janisozaur:
```
../src/openrct2/ParkFile.cpp:2010:29: error: 'struct Staff' has no member named 'StaffId'; did you mean 'Staff'?
         cs.ReadWrite(entity.StaffId);
                             ^~~~~~~
                             Staff
```

ZehMatt: No, can be removed.

src/openrct2/ride/Ride.cpp
Gymnasiast: Can we get a resolution for this?

src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -1083,7 +1083,8 @@ void path_paint_box_support(
             session, image_id | imageFlags, { 0, 0, height }, { boundBoxSize, 0 },
             { boundBoxOffset, height + boundingBoxZOffset });
 
-        if (!pathElement.IsQueue() && !pathElement.ShouldDrawPathOverSupports())
+        // TODO: Revert this when path import works correctly.
```
Gymnasiast, line 1086:
```suggestion
```

src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -1223,7 +1224,8 @@ void path_paint_pole_support(
             session, bridgeImage | imageFlags, { 0, 0, height }, { boundBoxSize, 0 },
             { boundBoxOffset, height + boundingBoxZOffset });
 
-        if (pathElement.IsQueue() || pathElement.ShouldDrawPathOverSupports())
+        // TODO: Revert this when path import works correctly.
```
Gymnasiast, line 1227:
```suggestion
```

docs/save-format.md
```diff
@@ -0,0 +1,255 @@
...
+
+Rides:
+
+Banners:
+
+Animations:
+
+Staff:
```
duncanspumpkin, line 151: remove

docs/save-format.md
```diff
@@ -0,0 +1,255 @@
...
+  (spare: -2)
+
+track-element:
+  type:                     uint16          (straight, 25 deg up, brakes etc.)
+  sequence:                 uint8
+  style:                    uint8           (wooden, intamin, b&m etc.)
+  station-index:            uint8
+  colour:                   uint8
+  flags:                    uint8           (station / on-ride)
+  (spare: 5)
```
duncanspumpkin, line 200: out of date

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -101,19 +101,21 @@ static constexpr const ObjectPageDesc ObjectSelectionPages[] = {
...
 
-    // Currently hidden until new save format arrives:
-    // { STR_OBJECT_SELECTION_TERRAIN_SURFACES,            SPR_G2_TAB_LAND,            true  },
-    // { STR_OBJECT_SELECTION_TERRAIN_EDGES,               SPR_G2_TAB_LAND,            true  },
-    // { STR_OBJECT_SELECTION_STATIONS,                    SPR_TAB_PARK,               true  },
-    // { STR_OBJECT_SELECTION_MUSIC,                       SPR_TAB_MUSIC_0,            false },
-    // { STR_OBJECT_SELECTION_FOOTPATH_SURFACES,           SPR_TAB_SCENERY_PATHS,      false },
-    // { STR_OBJECT_SELECTION_FOOTPATH_RAILINGS,           SPR_G2_PATH_RAILINGS_TAB,   false },
+    // Dummy place holder for string objects
+    { STR_NONE,                   static_cast<uint32_t>(SPR_NONE),                  false },
```
duncanspumpkin, line 111: remove cast?

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -148,7 +150,7 @@ validate_global_widx(WC_EDITOR_OBJECT_SELECTION, WIDX_TAB_1);
 static bool _window_editor_object_selection_widgets_initialised;
 static std::vector<rct_widget> _window_editor_object_selection_widgets = {
     WINDOW_SHIM(WINDOW_TITLE, WW, WH),
-    MakeWidget({  0, 43}, {600, 357}, WindowWidgetType::Resize,       WindowColour::Secondary                                                                  ),
+    MakeWidget({  0, 43}, {WW, 357}, WindowWidgetType::Resize,       WindowColour::Secondary                                                                  ),
```
duncanspumpkin, line 153: realign

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -1371,26 +1368,27 @@ static void window_editor_object_selection_manage_tracks()
...
+            auto descriptor = ObjectEntryDescriptor(*item);
+            const auto* loadedObject = objManager.GetLoadedObject(descriptor);
             if (loadedObject == nullptr)
             {
-                loadedObject = object_manager_load_object(entry);
+                loadedObject = objManager.LoadObject(descriptor);
                 if (loadedObject == nullptr)
                 {
-                    log_error("Failed to load entry %.8s", entry->name);
+                    log_error("Failed to load entry %s", std::string(descriptor.GetName()).c_str());
```
duncanspumpkin, line 1386: Unsure if this is safe. Lifetime of std::string

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -749,6 +773,39 @@ static void window_footpath_show_footpath_types_dialog(rct_window* w, rct_widget
         gDropdownDefaultIndex = static_cast<int32_t>(*defaultIndex);
 }
 
+static void window_footpath_show_railings_types_dialog(rct_window* w, rct_widget* widget)
```
duncanspumpkin, line 776: All new functions to be TitleCase

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1870,10 +1818,38 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
             case TILE_ELEMENT_TYPE_PATH:
             {
                 // Details
-                // Path name
-                auto ft = Formatter();
-                ft.Add<rct_string_id>(tileElement->AsPath()->GetSurfaceDescriptor()->Name);
-                DrawTextBasic(dpi, screenCoords, STR_TILE_INSPECTOR_PATH_NAME, ft, { w->colours[1] });
+                auto pathEl = tileElement->AsPath();
```
duncanspumpkin, line 1821: auto* for all this case

src/openrct2/ParkFile.cpp
```diff
@@ -0,0 +1,4593 @@
...
+    // Current version that is saved.
+    constexpr uint32_t PARK_FILE_CURRENT_VERSION = 0x5;
+
+    // The minimum version that is forwards compatible with the current version.
+    constexpr uint32_t PARK_FILE_MIN_VERSION = 0x5;
+
+    namespace ParkFileChunkType
+    {
+        // clang-format off
+//        constexpr uint32_t RESERVED_0       = 0x00;
```
duncanspumpkin, line 83: realig

src/openrct2/management/Research.cpp
```diff
@@ -718,18 +724,17 @@ void research_remove_flags()
     }
 }
 
-void research_fix()
+static void ResearchRemoveNullItems(std::vector<ResearchItem>& items)
```
duncanspumpkin, line 727: Replace with `std::erase_if`

src/openrct2/world/Entrance.cpp
```diff
@@ -217,6 +217,23 @@ void fix_park_entrance_locations(void)
         gParkEntrances.end());
 }
 
+void UpdateParkEntranceLocations()
```
duncanspumpkin, line 220: Rename to `ResetParkEntranceLocations` or something else. Its not really updating them its setting them on first use. Should perhaps be moved to the parkfile as its only used by that.

src/openrct2/world/Scenery.cpp
```diff
@@ -394,3 +395,41 @@ void RestrictAllMiscScenery()
         }
     }
 }
+
+ObjectType GetObjectTypeFromSceneryType(uint8_t type)
```
duncanspumpkin, line 399: These are only used by the parkfile and then only for legacy interop. Suggest moving them back.

src/openrct2-ui/windows/EditorObjectSelection.cpp
ZehMatt: Lifetime here ends after the call

Broxzier: The `std::string` is a temporary and is valid until the end of the full expression, which is after `log_error` gets called, so this is perfectly safe.

src/openrct2/management/Research.cpp
duncanspumpkin: hmm erase_if is c++20. Will use a std::remove_if, erase.

src/openrct2/ride/Ride.cpp
duncanspumpkin: I still feel we should do it `(gCurrentTicks & 0xFF) == (rideId & 0xFF)` as is atm rides will breakdown much less frequently.

ZehMatt: Makes sense

duncanspumpkin (approved):

janisozaur (approved):
Done the usual test of compiling with latest upstream versions of compilers

PR #10674, opened 2020-02-08
Refactor/more tile elements

Gymnasiast:
Follow-up to https://github.com/OpenRCT2/OpenRCT2/pull/10667.

### DIFF
--- a/src/openrct2/rct12/RCT12.h
+++ b/src/openrct2/rct12/RCT12.h
@@ -48,7 +48,7 @@ constexpr uint16_t const RCT12_MAX_INVERSIONS = 31;
 constexpr uint16_t const RCT12_MAX_GOLF_HOLES = 31;
 constexpr uint16_t const RCT12_MAX_HELICES = 31;
 
-constexpr uint8_t RCT12_BANNER_INDEX_NULL = (uint8_t)-1;
+constexpr uint8_t RCT12_BANNER_INDEX_NULL = std::numeric_limits<uint8_t>::max();
 
 enum class RCT12TrackDesignVersion : uint8_t
 {
--- a/src/openrct2/rct12/RCT12.h
+++ b/src/openrct2/rct12/RCT12.h
@@ -48,6 +48,8 @@ constexpr uint16_t const RCT12_MAX_INVERSIONS = 31;
 constexpr uint16_t const RCT12_MAX_GOLF_HOLES = 31;
 constexpr uint16_t const RCT12_MAX_HELICES = 31;
 
+constexpr uint8_t RCT12_BANNER_INDEX_NULL = (uint8_t)-1;
+
 enum class RCT12TrackDesignVersion : uint8_t
 {
     TD4,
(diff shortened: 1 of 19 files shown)


duncanspumpkin:
This is failing as the replay is now broken. Delete the banner place replay in this PR and make an issue in the replay repo to recreate it. It should also increment the network version.

Gymnasiast:
@duncanspumpkin Are we sure that the test failing is not down to errors that I made?

duncanspumpkin:
The banner placement code looks fine as far as i can tell apart from the fact i can tell it will definitely break the replay as the variable size change will break it.

duncanspumpkin:
Argh how annoying the replays are still broken. I thought banner place was the only one that had issues but looks like banner set colour is broken as well. It must end up calling banner set style.

Gymnasiast:
With the replay tests (what's left of them) now succeeding, I get the following error with Xcode and Android:
```
 /__w/OpenRCT2/OpenRCT2/src/openrct2/object/../world/Banner.h:31:17: error: implicit instantiation of undefined template 'std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> >'
      std::string text;
                  ^
  /android-dev/ndk/sources/cxx-stl/llvm-libc++/include/iosfwd:193:32: note: template is declared here
      class _LIBCPP_TEMPLATE_VIS basic_string;
```

I have no idea why. I didn't even touch the line it complains about.

src/openrct2/rct12/RCT12.h
```diff
@@ -48,6 +48,8 @@ constexpr uint16_t const RCT12_MAX_INVERSIONS = 31;
 constexpr uint16_t const RCT12_MAX_GOLF_HOLES = 31;
 constexpr uint16_t const RCT12_MAX_HELICES = 31;
 
+constexpr uint8_t RCT12_BANNER_INDEX_NULL = (uint8_t)-1;
```
tupaschoal: Maybe use `std::numeric_limits<uint8_t>::max() instead?

src/openrct2/world/Banner.h
```diff
@@ -10,7 +10,15 @@
...
-#include "TileElement.h"
+#include "../ride/RideTypes.h"
+#include "Location.hpp"
+
+#include <string>
+
+struct TileElement;
+struct WallElement;
+
+typedef uint16_t BannerIndex;
```
tupaschoal: Prefer `using`
```suggestion
using BannerIndex = uint16_t;
```

Gymnasiast: Our code uses `typedef` almost throughout.

tupaschoal: Well, one could argue that you just moved this from another file and changed that type and that one was with `using` :P

IntelOrca: > Our code uses typedef almost throughout.

This is mostly due to our codebase originally being in C.

Gymnasiast: I don't see the point, but well...

tupaschoal: `using` is the C++ way of doing `typedef` since C++11. Technically they behave exactly like one another, but in terms of readability, I think `using` is friendlier to new programmers, due to the `X = Y` construction.

It's a point made by [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rt-using) and some of the examples show that it really adds value in some cases

Gymnasiast: Can you open a PR to replace the rest, then? I prefer to be consistent.

tupaschoal: Will do!

Issue #9472, opened 2019-06-23
Certain 4-station rides can't be saved properly


Gymnasiast:
**OS:** Ubuntu 19.04
**Version:** 0.2.2+
**Commit/Build:** a0b77c4bf

Copied from https://forums.openrct2.org/topic/3833-certain-4-station-rides-cant-be-saved-properly/

Original text:
` After saving, some 4 station rides will be blank in the ride previewer. When trying to build them, nothing happens except an empty ride construction window opening and, if money is involved, an immediate build and delete transaction. From my testing, it seems that a faulty saved ride can be made by creating a 4 station ride in a way such that station 1 of the saved ride would be longer than 2 segments, though I don't know if it's the only way. Original RCT2 seems to have the same problem but simply raises an access violation upon selecting the faulty saved ride.

I've attached a save file with the same coaster in 4 rotations, with one station being 3 segments long and the others 2 segments long. The one named "4station fail" produces a faulty track file. I've also attached the faulty and non-faulty track files.﻿`

**Save files**
Can be downloaded from https://forums.openrct2.org/topic/3833-certain-4-station-rides-cant-be-saved-properly/

jensj12:
Probably related: #6123

duncanspumpkin:
Its to do with there being 4 stations and it tries to allocate a 5th. Still trying to work out why its doing that. Looks like some sort of issue in the code that checks the station pieces or perhaps the backwards check that works out the start of a coaster.

duncanspumpkin:
Okay I've worked it out. When you have four stations you pretty much are guaranteed to cause it break if your first station is >2 pieces long. This happens because the start of the ride is marked by the first station piece it then iterates forward around the track. For every station it encounters it checks to see if its connected to another station and if not allocates it a new station number. The problem happens when it gets back to the first station. By this point it has already allocated every station number and it doesn't realise that it is connected to the first station.

Ryder17z:
don't close it, if it's still an issue.
This one can't be built: [Steel Twister 1.zip](https://github.com/OpenRCT2/OpenRCT2/files/4146416/Steel.Twister.1.zip)
created with: 3f837f5 today

PR #10677, opened 2020-02-09
Part of #9472: Convert vehicle_update to Vehicle struct method

TusharSariya:
…member of the Vehicle struct

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -1979,7 +1979,8 @@ void Vehicle::Update()
     {
         _vehicleBreakdown = curRide->breakdown_reason_pending;
         auto vehicleEntry = &rideEntry->vehicles[vehicle_type];
-        if ((vehicleEntry->flags & VEHICLE_ENTRY_FLAG_POWERED) && curRide->breakdown_reason_pending == BREAKDOWN_SAFETY_CUT_OUT)
+        if ((vehicleEntry->flags & VEHICLE_ENTRY_FLAG_POWERED) &&
+            curRide->breakdown_reason_pending == BREAKDOWN_SAFETY_CUT_OUT)
         {
             if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_WATER_RIDE)
                 || (vehicle_sprite_type == 2 && velocity <= 0x20000))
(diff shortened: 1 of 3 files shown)


src/openrct2/ride/Vehicle.h
```diff
@@ -294,6 +294,7 @@ struct Vehicle : SpriteBase
     {
         return type == VEHICLE_TYPE_HEAD;
     }
+    void vehicle_update();
```
ZehMatt:
```suggestion
    void Update();
```
There is no need to for a prefix anymore.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1955,114 +1954,114 @@ static SoundIdVolume sub_6D7AC0(SoundId currentSoundId, uint8_t currentVolume, S
  *
  *  rct2: 0x006D77F2
  */
-static void vehicle_update(Vehicle* vehicle)
+void Vehicle::vehicle_update()
 {
     // The cable lift uses the ride type of NULL
-    if (vehicle->ride_subtype == RIDE_TYPE_NULL)
+    if (this->ride_subtype == RIDE_TYPE_NULL)
```
duncanspumpkin: You don't need to use `this` `ride_subtype` is already available.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1955,114 +1954,114 @@ static SoundIdVolume sub_6D7AC0(SoundId currentSoundId, uint8_t currentVolume, S
...
     }
 
-    auto rideEntry = get_ride_entry(vehicle->ride_subtype);
+    auto rideEntry = get_ride_entry(this->ride_subtype);
     if (rideEntry == nullptr)
         return;
 
-    auto ride = get_ride(vehicle->ride);
-    if (ride == nullptr)
+    auto r = get_ride(ride);
```
duncanspumpkin: `curRide` might be a better name but this is fine as well.

duncanspumpkin:
As you are a first time contributor you should also add yourself to https://github.com/OpenRCT2/OpenRCT2/blob/develop/contributors.md under bug fixes.

src/openrct2/ride/Vehicle.cpp
TusharSariya: right so I named it `r` because that is how it was implemented in other places, i will change the naming convention to `curRide` in the whole `vehcle.cpp` file.

duncanspumpkin:
You just need to fix the code formatting and this is good to go. Check the ci for what is wrong. https://github.com/OpenRCT2/OpenRCT2/pull/10677/checks?check_run_id=434634365

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: The variable names should be consistent but don't just blindly change them all. Context is key.

PR #10690, opened 2020-02-12
Hide paint_arrange_structs_helper from global symbols

janisozaur:


janisozaur:
There was a teeny tiny improvement in the timings across the board (`benchspritesort` and `benchgfx`) in my tests

IntelOrca:
Do we not use LTO? I thought that would make changes like this make no difference to run-time performance.

PR #10695, opened 2020-02-13
Cleanup some code around CoordsXY

ZehMatt:
It appears that only LLVM is somewhat capable of properly optimizing the code for when the struct is passed by value. I've cleaned up a bunch of functions to simply use const ref, this adds roughly 5 to 8 FPS back on my end running bpb.sv6

I've only touched functions that didn't require a lot of editing or ended up in my list of bottlenecks during profiling.


duncanspumpkin:
Originally i didn't `const &` then i did then i didn't. I think I reflected that in my comments on your PRs so it's probably my fault.

PR #10697, opened 2020-02-13
Clean up scenery window variables

Gymnasiast:
This moves many scenery window globals to the UI project and removes three globals that were never read, only assigned.


AaronVanGeffen:
I wish you'd done this after #10357 was merged… Ah well.

Gymnasiast:
@AaronVanGeffen Totally forgot about that. I have now rebased that PR.

PR #10698, opened 2020-02-13
Confine use of LocationXY8 to RCT12 load/save code

Gymnasiast:


src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1309,6 +1309,7 @@ class S6Importer final : public IParkImporter
         dst->colours = src->colours;
         dst->track_progress = src->track_progress;
         dst->track_direction = src->track_direction;
+        dst->BoatLocation = TileCoordsXY{ src->boat_location.x, src->boat_location.y }.ToCoordsXY();
```
Gymnasiast: Check for null location.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1005,7 +1007,11 @@ void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
     dst->vehicle_type = src->vehicle_type;
     dst->colours = src->colours;
     dst->track_progress = src->track_progress;
-    dst->track_direction = src->track_direction;
+    if (ride != nullptr && ride->mode == RIDE_MODE_BOAT_HIRE)
```
Gymnasiast: Check for null location.

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1222,6 +1222,7 @@ class S4Importer final : public IParkImporter
         dst->track_z = src->track_z;
         dst->current_station = src->current_station;
         dst->track_type = src->track_type;
+        dst->BoatLocation = TileCoordsXY{ src->boat_location.x, src->boat_location.y }.ToCoordsXY();
```
Gymnasiast: Check for null location.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -4582,12 +4581,12 @@ static void vehicle_update_boat_location(Vehicle* vehicle)
     TileCoordsXY returnPosition = ride->boat_hire_return_position;
     uint8_t returnDirection = ride->boat_hire_return_direction & 3;
 
-    TileCoordsXY location{ CoordsXY{ vehicle->x, vehicle->y } + CoordsDirectionDelta[returnDirection] };
+    CoordsXY location = CoordsXY{ vehicle->x, vehicle->y } + CoordsDirectionDelta[returnDirection];
 
-    if (location == returnPosition)
+    if (location == returnPosition.ToCoordsXY())
```
duncanspumpkin: You probably want to do location.ToTileStart(). As you want it to pass this condition if it is on the same tile.

src/openrct2/rct2/S6Importer.cpp
duncanspumpkin: On `ride->mode != RIDE_MODE_BOAT_HIRE` this should set it to a null value.

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1222,6 +1222,14 @@ class S4Importer final : public IParkImporter
         dst->track_z = src->track_z;
         dst->current_station = src->current_station;
         dst->track_type = src->track_type;
+        if (src->boat_location.isNull() || ride->mode == RIDE_MODE_BOAT_HIRE)
```
Gymnasiast:
```suggestion
        if (src->boat_location.isNull() || ride->mode != RIDE_MODE_BOAT_HIRE)
```

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1308,6 +1310,14 @@ class S6Importer final : public IParkImporter
         dst->colours = src->colours;
         dst->track_progress = src->track_progress;
         dst->track_direction = src->track_direction;
+        if (src->boat_location.isNull() || ride.mode == RIDE_MODE_BOAT_HIRE)
```
Gymnasiast:
```suggestion
        if (src->boat_location.isNull() || ride.mode != RIDE_MODE_BOAT_HIRE)
```

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -996,6 +996,8 @@ void S6Exporter::ExportSpriteCommonProperties(RCT12SpriteBase* dst, const Sprite
 
 void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
 {
+    auto ride = get_ride(src->ride);
```
tupaschoal: Should this be `const auto&` like the other places?

ZehMatt: If at all it should be ```const auto*```

src/openrct2/rct12/RCT12.h
```diff
@@ -50,6 +50,15 @@ constexpr uint16_t const RCT12_MAX_HELICES = 31;
 
 constexpr uint8_t RCT12_BANNER_INDEX_NULL = std::numeric_limits<uint8_t>::max();
 
+constexpr uint16_t const RCT12_XY8_UNDEFINED = 0xFFFF;
+
+// Everything before this point has been researched
+#define RCT12_RESEARCHED_ITEMS_SEPARATOR 0xFFFFFFFF
```
ZehMatt: Could make those constexpr too

Gymnasiast: I think you're looking at the wrong PR ;-)
The one about `constexpr`ing stuff is here: https://github.com/OpenRCT2/OpenRCT2/pull/10855

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1135,7 +1135,7 @@ class S4Importer final : public IParkImporter
 
     void ImportVehicle(Vehicle* dst, rct1_vehicle* src)
     {
-        auto ride = get_ride(src->ride);
+        const auto& ride = get_ride(src->ride);
```
tupaschoal: `const auto*` to be consistent with the other changes

src/openrct2/ride/Vehicle.cpp
```diff
@@ -4576,12 +4575,12 @@ static void vehicle_update_boat_location(Vehicle* vehicle)
     TileCoordsXY returnPosition = ride->boat_hire_return_position;
     uint8_t returnDirection = ride->boat_hire_return_direction & 3;
 
-    TileCoordsXY location{ CoordsXY{ vehicle->x, vehicle->y } + CoordsDirectionDelta[returnDirection] };
+    CoordsXY location = CoordsXY{ vehicle->x, vehicle->y } + CoordsDirectionDelta[returnDirection];
 
-    if (location == returnPosition)
+    if (location == returnPosition.ToCoordsXY().ToTileStart())
```
duncanspumpkin: location should be at tile start. return position is fine. This will cause issues with pathfinding to exits.

PR #10699, opened 2020-02-13
Replace some usages of LocationXY(Z)16

Gymnasiast:


src/openrct2/paint/VirtualFloor.cpp
```diff
@@ -200,7 +198,7 @@ bool virtual_floor_tile_is_floor(int16_t x, int16_t y)
 }
 
 static void virtual_floor_get_tile_properties(
-    int16_t x, int16_t y, int16_t height, bool* outOccupied, bool* tileOwned, uint8_t* outOccupiedEdges, bool* outBelowGround,
+    CoordsXY loc, int16_t height, bool* outOccupied, bool* tileOwned, uint8_t* outOccupiedEdges, bool* outBelowGround,
```
tupaschoal: `const CoordsXY& loc`

src/openrct2/paint/VirtualFloor.cpp
```diff
@@ -213,8 +211,8 @@ static void virtual_floor_get_tile_properties(
     // See if we are a selected tile
     if ((gMapSelectFlags & MAP_SELECT_FLAG_ENABLE))
     {
-        if (x >= gMapSelectPositionA.x && y >= gMapSelectPositionA.y && x <= gMapSelectPositionB.x
-            && y <= gMapSelectPositionB.y)
+        if (loc.x >= gMapSelectPositionA.x && loc.y >= gMapSelectPositionA.y && loc.x <= gMapSelectPositionB.x
+            && loc.y <= gMapSelectPositionB.y)
```
tupaschoal: Maybe we should introduce the `operator>=`

src/openrct2/paint/VirtualFloor.cpp
```diff
@@ -177,9 +175,9 @@ bool virtual_floor_tile_is_floor(int16_t x, int16_t y)
 
     // Check if map selection (usually single tiles) are enabled
     //  and if the current tile is near or on them
-    if ((gMapSelectFlags & MAP_SELECT_FLAG_ENABLE) && x >= gMapSelectPositionA.x - _virtualFloorBaseSize
-        && y >= gMapSelectPositionA.y - _virtualFloorBaseSize && x <= gMapSelectPositionB.x + _virtualFloorBaseSize
-        && y <= gMapSelectPositionB.y + _virtualFloorBaseSize)
+    if ((gMapSelectFlags & MAP_SELECT_FLAG_ENABLE) && loc.x >= gMapSelectPositionA.x - _virtualFloorBaseSize
+        && loc.y >= gMapSelectPositionA.y - _virtualFloorBaseSize && loc.x <= gMapSelectPositionB.x + _virtualFloorBaseSize
+        && loc.y <= gMapSelectPositionB.y + _virtualFloorBaseSize)
```
tupaschoal: Would it look better like:
```c++
    const auto virtualFloorBaseCoords = CoordsXY{_virtualFloorBaseSize, _virtualFloorBaseSize};
    if ((gMapSelectFlags & MAP_SELECT_FLAG_ENABLE) && loc >= (gMapSelectPositionA - virtualFloorBaseCoords)
        && loc <= (gMapSelectPositionB + virtualFloorBaseCoords)
```
I'm unsure if putting it into `CoordsXY` are a bit of stretch

Gymnasiast: Another option would be to overload the CoordsXY + operator to also allow integers. But I don't think that should go into this PR, if I'm honest.

tupaschoal: I'm fine with it being done on another pr

duncanspumpkin: What would a + overload mean? Add integer to x coordinate, add to both x and y? It's a little ambiguous. I'm not 100% sure if the <= overloads make great sense either. What does it mean for a coordinate to be less than another.

ZehMatt: I'm with @duncanspumpkin, it adds some ambiguity, if theres some distance measurement involved it should be its own distance function outside of the class or make it a static to accept A and B as parameters.

PR #10700, opened 2020-02-13
Add benchsimulate command

janisozaur:
Adds a new command, `benchsimulate`, that starts benchmarking
UpdateLogic() function.

### DIFF
--- a/src/openrct2/GameState.cpp
+++ b/src/openrct2/GameState.cpp
@@ -369,7 +369,7 @@ void GameState::UpdateLogic(LogicTimings* timings)
 
     if (timings != nullptr)
     {
-        timings->CurrentIdx = (timings->CurrentIdx + 1) % LOGIC_UPDATE_MEASURMENTS_COUNT;
+        timings->CurrentIdx = (timings->CurrentIdx + 1) % LOGIC_UPDATE_MEASUREMENTS_COUNT;
     }
 }
 
--- a/src/openrct2/GameState.cpp
+++ b/src/openrct2/GameState.cpp
@@ -12,6 +12,7 @@
 #include "Context.h"
 #include "Editor.h"
 #include "Game.h"
+#include "GameState.h"
 #include "GameStateSnapshots.h"
 #include "Input.h"
 #include "OpenRCT2.h"
@@ -37,6 +38,7 @@
 #include "world/Scenery.h"
 
 #include <algorithm>
+#include <chrono>
 
 using namespace OpenRCT2;
 using namespace OpenRCT2::Scripting;
@@ -225,8 +227,17 @@ void GameState::Update()
     gInUpdateCode = false;
 }
 
-void GameState::UpdateLogic()
+void GameState::UpdateLogic(LogicTimings* timings)
 {
+    auto start_time = std::chrono::high_resolution_clock::now();
+
+    auto report_time = [timings, start_time](LogicTimePart part) {
+        if (timings != nullptr)
+        {
+            timings->TimingInfo[part][timings->CurrentIdx] = std::chrono::high_resolution_clock::now() - start_time;
+        }
+    };
+
     gScreenAge++;
     if (gScreenAge == 0)
         gScreenAge--;
@@ -234,6 +245,7 @@ void GameState::UpdateLogic()
     GetContext()->GetReplayManager()->Update();
 
     network_update();
+    report_time(LogicTimePart::NetworkUpdate);
 
     if (network_get_mode() == NETWORK_MODE_SERVER)
     {
@@ -277,33 +289,51 @@ void GameState::UpdateLogic()
 
     date_update();
     _date = Date(static_cast<uint32_t>(gDateMonthsElapsed), gDateMonthTicks);
+    report_time(LogicTimePart::Date);
 
     scenario_update();
+    report_time(LogicTimePart::Scenario);
     climate_update();
+    report_time(LogicTimePart::Climate);
     map_update_tiles();
+    report_time(LogicTimePart::MapTiles);
     // Temporarily remove provisional paths to prevent peep from interacting with them
     map_remove_provisional_elements();
+    report_time(LogicTimePart::MapStashProvisionalElements);
     map_update_path_wide_flags();
+    report_time(LogicTimePart::MapPathWideFlags);
     peep_update_all();
+    report_time(LogicTimePart::Peep);
     map_restore_provisional_elements();
+    report_time(LogicTimePart::MapRestoreProvisionalElements);
     vehicle_update_all();
+    report_time(LogicTimePart::Vehicle);
     sprite_misc_update_all();
+    report_time(LogicTimePart::Misc);
     Ride::UpdateAll();
+    report_time(LogicTimePart::Ride);
 
     if (!(gScreenFlags & SCREEN_FLAGS_EDITOR))
     {
         _park->Update(_date);
     }
+    report_time(LogicTimePart::Park);
 
     research_update();
+    report_time(LogicTimePart::Research);
     ride_ratings_update_all();
+    report_time(LogicTimePart::RideRatings);
     ride_measurements_update();
+    report_time(LogicTimePart::RideMeasurments);
     News::UpdateCurrentItem();
+    report_time(LogicTimePart::News);
 
     map_animation_invalidate_all();
+    report_time(LogicTimePart::MapAnimation);
     vehicle_sounds_update();
     peep_update_crowd_noise();
     climate_update_sound();
+    report_time(LogicTimePart::Sounds);
     editor_open_windows_for_current_step();
 
     // Update windows
@@ -316,9 +346,11 @@ void GameState::UpdateLogic()
     }
 
     GameActions::ProcessQueue();
+    report_time(LogicTimePart::GameActions);
 
     network_process_pending();
     network_flush();
+    report_time(LogicTimePart::NetworkFlush);
 
     gCurrentTicks++;
     gScenarioTicks++;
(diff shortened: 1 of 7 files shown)


ZehMatt:
Is there a way to annotate/tag a certain event during the profiling? The UpdateLogic function has like a bunch of sub-updates and it would be nice to know which one of them performs terrible, heres a list:
```
date_update
scenario_update
map_update_tiles
map_remove_provisional_elements
map_update_path_wide_flags
peep_update_all
map_restore_provisional_elements
vehicle_update_all
sprite_misc_update_all
research_update
ride_ratings_update_all
ride_measurements_update
```
There are quite a few, so it would be nice to know what is what.

janisozaur:
I see two ways to provide this:
1. Add explicit benchmarks for each of those
  - note that testing a single function out of the bunch could lead to results different than the current approach of using `UpdateLogic()`, as they likely touch same data (think: updating the peeps but not vehicles they're in could lead to unexpected results)
2. Use manual timing (a feature of gbench) and collect timings for each relevant section
  - this could be reused to expose some timing details in-game (see how it's done in OpenTTD https://imgur.com/gallery/tyI3Htc)
  - collect them to a map so they values can be given names without having to change underlying structure
  - display them using gbench's custom counters

src/openrct2/cmdline/BenchUpdate.cpp
```diff
@@ -0,0 +1,119 @@
...
+            // Register benchmark for sv6 if valid
+            benchmark::RegisterBenchmark(argv[i], BM_update, argv[i]);
+        }
+        else
+        {
+            argv_for_benchmark.push_back((char*)argv[i]);
+        }
+    }
+    // Update argc with all the changes made
+    argc = (int)argv_for_benchmark.size();
```
tupaschoal: Use named cast?

src/openrct2/cmdline/BenchUpdate.cpp
```diff
@@ -0,0 +1,119 @@
...
+    core_init();
+    gOpenRCT2Headless = true;
+
+    ::benchmark::RunSpecifiedBenchmarks();
+    return 0;
+}
+
+static exitcode_t HandleBenchUpdate(CommandLineArgEnumerator* argEnumerator)
+{
+    const char** argv = (const char**)argEnumerator->GetArguments() + argEnumerator->GetIndex();
```
tupaschoal: Use named cast?

AaronVanGeffen:
This is essentially doing the same thing as #10281. We need to decide which to merge.

src/openrct2/GameState.cpp
```diff
@@ -277,33 +289,51 @@ void GameState::UpdateLogic()
...
     _date = Date(static_cast<uint32_t>(gDateMonthsElapsed), gDateMonthTicks);
+    report_time(LogicTimePart::Date);
 
     scenario_update();
+    report_time(LogicTimePart::Scenario);
     climate_update();
+    report_time(LogicTimePart::Climate);
     map_update_tiles();
+    report_time(LogicTimePart::MapTiles);
     // Temporarily remove provisional paths to prevent peep from interacting with them
```
tupaschoal, line 300: I wonder if it would be better to do something like:

```c++
void GameState::UpdateLogic(LogicTimings* timings)
{
    for (...)
        RunStep(step);
}

void GameLogic::RunStep(LogicTimePart step)
{
    switch(step)
    {
        case LogicTimePart::Network: ...
        case ...
    }
    report_time(step);
}
```

This would force us to write more enums and to divide the code better, but also would be a little bit of less repetition? It would also be clearer which steps are and aren't being measured?

src/openrct2/GameState.h
```diff
@@ -11,12 +11,54 @@
...
+        GameActions,
+        NetworkFlush,
+        Scripts,
+    };
+
+    // ~6.5s at 40Hz
+    constexpr size_t LOGIC_UPDATE_MEASURMENTS_COUNT = 256;
+
+    // In order not to cause allocations, collect multiple samples into single pre-allocated struct
+    using LogicTimingInfo = std::map<LogicTimePart, std::array<std::chrono::duration<double>, LOGIC_UPDATE_MEASURMENTS_COUNT>>;
```
tupaschoal, line 54:
```suggestion
    using LogicTimingInfo = std::unordered_map<LogicTimePart, std::array<std::chrono::duration<double>, LOGIC_UPDATE_MEASURMENTS_COUNT>>;
```
Doesn't seem like we really need the ordering

src/openrct2/GameState.h
```diff
@@ -11,12 +11,54 @@
 
 #include "Date.h"
 
+#include <array>
+#include <chrono>
+#include <map>
```
tupaschoal, line 16:
```suggestion
#include <unordered_map>
```

src/openrct2/cmdline/BenchUpdate.cpp
```diff
@@ -0,0 +1,160 @@
...
+#    include <vector>
+
+using namespace OpenRCT2;
+
+static void BM_update(benchmark::State& state, const std::string& filename)
+{
+    std::unique_ptr<IContext> context(CreateContext());
+    if (context->Initialise())
+    {
+        if (!filename.empty() && !context->LoadParkFromFile(filename))
```
tupaschoal, line 32: What if the filename is empty? The function will just run, is that expected?

src/openrct2/cmdline/BenchUpdate.cpp
```diff
@@ -0,0 +1,160 @@
...
+        {
+            state.SkipWithError("Failed to load file!");
+        }
+
+        std::vector<LogicTimings> timings(1);
+        timings.reserve(100);
```
tupaschoal, line 38:
```suggestion
        std::vector<LogicTimings> timings(100);
```

src/openrct2/cmdline/BenchUpdate.cpp
```diff
@@ -0,0 +1,160 @@
...
+            context->GetGameState()->UpdateLogic(timingToUse);
+        }
+        state.SetItemsProcessed(state.iterations());
+        auto accumulator = [timings](LogicTimePart part) -> double {
+            std::chrono::duration<double> timesum;
+            for (const auto& timing : timings)
+            {
+                for (const auto& PartTime : timing.TimingInfo.at(part))
+                    timesum += PartTime;
+            }
```
tupaschoal, line 57: Can't we use `std::acumulate`?

src/openrct2/cmdline/BenchUpdate.cpp
```diff
@@ -0,0 +1,160 @@
...
+        state.counters["ParkAcc_ms"] = accumulator(LogicTimePart::Park);
+        state.counters["ResearchAcc_ms"] = accumulator(LogicTimePart::Research);
+        state.counters["RideRatingsAcc_ms"] = accumulator(LogicTimePart::RideRatings);
+        state.counters["RideMeasurmentsAcc_ms"] = accumulator(LogicTimePart::RideMeasurments);
+        state.counters["NewsAcc_ms"] = accumulator(LogicTimePart::News);
+        state.counters["MapAnimationAcc_ms"] = accumulator(LogicTimePart::MapAnimation);
+        state.counters["SoundsAcc_ms"] = accumulator(LogicTimePart::Sounds);
+        state.counters["GameActionsAcc_ms"] = accumulator(LogicTimePart::GameActions);
+        state.counters["NetworkFlushAcc_ms"] = accumulator(LogicTimePart::NetworkFlush);
+        state.counters["ScriptsAcc_ms"] = accumulator(LogicTimePart::Scripts);
```
tupaschoal, line 81: Maybe we should have something that converts from an enum of `LogicTimePart` into the corresponding string, to be less error prone. That would also allows us to do a for on the enum updating all counters
```c++
for (<all_enums>)
    state.counters[ToLogicPartString(enum)] = accumulator(enum);
```

src/openrct2/cmdline/BenchUpdate.cpp
```diff
@@ -0,0 +1,160 @@
...
+    }
+    else
+    {
+        state.SkipWithError("Context initialization failed.");
+    }
+}
+
+static int cmdline_for_bench_sprite_sort(int argc, const char* const* argv)
```
tupaschoal, line 89: Should it be
```suggestion
static int CmdlineForBenchSpriteSort(int argc, const char* const* argv)
```
?

tupaschoal (commented):
This is looking very neat

src/openrct2/GameState.cpp
janisozaur: I'm not quite fond of this, even with the code moved out to `switch`, there's little benefit - same code, just in different place. On top of that, the `report_time` is a lambda with captures, which could cause problems if it were to be passed to the other function - or possibly incur penalty of recreating it every time we call the `GameLogic::RunStep`

src/openrct2/cmdline/BenchUpdate.cpp
janisozaur: It will fail to load in `LoadParkFromFile` and get skipped. I believe that's fine. Additionally, as this is an explicitly provided parameter by a user to the `benchsimulate` command, I think the user will be aware of what to do with such problem.

src/openrct2/cmdline/BenchUpdate.cpp
janisozaur: But this isn't the same. `timings(100)` will create 100 objects, while the `.reserve(100)` will only allocate space for as many without changing vector's reported size. I'm not going to merge this

src/openrct2/cmdline/BenchUpdate.cpp
tupaschoal: Ok

src/openrct2/GameState.cpp
tupaschoal: Alright

tupaschoal (commented):
I've dropped the issues where you commented, but there are still a few to tackle.

janisozaur:
I've just updated, think it should be ready now

tupaschoal (approved):

PR #10706, opened 2020-02-15
Increase size of ride_id_t to 16 bits

Gymnasiast:
This does not yet modify the RIDE_ID_NULL value, which is planned to be addressed later.

### DIFF
--- a/src/openrct2/ride/RideTypes.h
+++ b/src/openrct2/ride/RideTypes.h
@@ -12,7 +12,7 @@
 #include <cstdint>
 #include <limits>
 
-typedef uint16_t ride_id_t;
+using ride_id_t = uint16_t;
 struct Ride;
 
 constexpr const ride_id_t RIDE_ID_NULL = 0xFF; // std::numeric_limits<ride_id_t>::max();
--- a/src/openrct2/ride/RideTypes.h
+++ b/src/openrct2/ride/RideTypes.h
@@ -10,9 +10,10 @@
 #pragma once
 
 #include <cstdint>
+#include <limits>
 
-typedef uint8_t ride_id_t;
-typedef uint16_t ride_idnew_t; // Temporary, old one can be removed after switching save format.
+typedef uint16_t ride_id_t;
 struct Ride;
 
-constexpr const ride_idnew_t RideIdNewNull = 0xFFFF;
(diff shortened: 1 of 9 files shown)


src/openrct2/ride/RideTypes.h
```diff
@@ -10,9 +10,10 @@
 #pragma once
 
 #include <cstdint>
+#include <limits>
 
-typedef uint8_t ride_id_t;
-typedef uint16_t ride_idnew_t; // Temporary, old one can be removed after switching save format.
+typedef uint16_t ride_id_t;
```
AaronVanGeffen: Now's as good a time as any to switch to C++ style `using` instead of C style `typedef`.

Gymnasiast:
```suggestion
using ride_id_t = uint16_t;
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -2682,15 +2682,12 @@ static void vehicle_update_waiting_to_depart(Vehicle* vehicle)
...
 }
 
-#pragma pack(push, 1)
 struct rct_synchronised_vehicle
 {
     ride_id_t ride_id;
     uint8_t station_id;
     uint16_t vehicle_id;
 };
-assert_struct_size(rct_synchronised_vehicle, 4);
```
tupaschoal: Why removing instead of updating? Just wondering

src/openrct2/world/TileElement.h
```diff
@@ -189,7 +189,7 @@ struct PathElement : TileElementBase
     union
     {
         uint8_t additionStatus; // 7
-        ride_id_t rideIndex;
+        uint8_t rideIndex;
```
tupaschoal: This looks weird, it's the only place you let go of the `ride_id_t` alias, even though. Why?

IntelOrca: Changes to save import / export are needed before we can change the tile element structs, so the ride id in tile elements has to remain 8-bit until we make those further changes.

src/openrct2/ride/Vehicle.cpp
IntelOrca: They were originally used to ensure we kept size compatibility with vanilla when interoping. We don't introp anymore and this struct is not used in the save file, or in networking so we can let the compiler adjust the struct however it likes.

Issue #10693, opened 2020-02-13
Add drawing code for boosters to B&M Flyer


Gymnasiast:
Request from NE.

(Note that this is about adding the drawing code - enabling by default is not planned.)

Gymnasiast:
Added by #10707.

PR #3501, opened 2016-05-05
Idea: Ride type descriptor

IntelOrca:
So now that we have individual source files for each type of ride, I thought it would be a good place to move all the constant data specific to a ride type in each source.

Currently a lot of the ride data is broken up into different structs and arrays, so this is essentially combining them all into one struct, `ride_type_desc`.

One problem though is the MSVC does not support `.field =` syntax in C++, which is a shame because its a nice way of, not only specifying required fields, but also annotating each value without using comments.

``` c
ride_type_desc RideTypeDescMerryGoRound = {
    .flags                      = RIDE_TYPE_DESC_FLAG_MUSIC_ON_DEFAULT |
                                  RIDE_TYPE_DESC_FLAG_ALLOW_MUSIC |
                                  RIDE_TYPE_DESC_FLAG_HAS_ENTRANCE_EXIT |
                                  RIDE_TYPE_DESC_FLAG_SINGLE_SESSION |
                                  RIDE_TYPE_DESC_FLAG_FLAG4_13,

    .operating_modes            = (1 << RIDE_MODE_ROTATION),
    .breakdowns                 = (1 << BREAKDOWN_SAFETY_CUT_OUT) | (1 << BREAKDOWN_CONTROL_FAILURE),

    .default_music_style        = MUSIC_STYLE_FAIRGROUND_ORGAN,

    .max_height                 = 12,
    .clearance_height           = 64,
    .z_offset                   = 3,
    .max_friction               = 255,
    .z                          = 2,
    .typical_price_multiplier   = 1,
    .bonus_value                = 45,

    .lift_hill_sound            = 255,
    .lift_hill_min_speed        = 5,
    .lift_hill_max_speed        = 5,

    // Prices
    .initial_upkeep_cost        = 50,
    .cost_per_track_piece       = 0,
    .price_primary              = 10,
    .price_secondary            = 0,
    .price_track                = 115,
    .price_supports             = 2,

    .value_rating_multiplier    = { 50, 10, 0 },

    // Strings
    .vehicle_name               = 1264,
    .structure_name             = 1278,
    .station_name               = 1257
};
```

PR #10707, opened 2020-02-15
Create ride type descriptor

Gymnasiast:
This is based on the idea of https://github.com/OpenRCT2/OpenRCT2/pull/3501 .

The plan is to migrate the information to RideTypeDescriptors field by field.

Also addresses https://github.com/OpenRCT2/OpenRCT2/issues/10693.


IntelOrca:
I see you have moved all the paint code for coasters into `ride/coaster/paint/...`. I think it would be better for these to be moved to `paint/ride/...` or `paint/ride/coaster`, the reason being is that we are likely to separate all the paint code into another code module (static library) separate from the core sources only need for running servers.

duncanspumpkin:
We should split this PR up and make it so that this is just for the ride type descriptors. Put all the file movement in a different PR.

Gymnasiast:
I have come up with an alternative solution for the designated initializers. This way, the correctness is enforced on Xcode/Clang/GCC, while Windows devs will at least still know which field should be set.

PR #6034, opened 2017-07-27
C++17 string_view test

janisozaur:
This is a dummy PR to test state of C++17 support.

PR #10710, opened 2020-02-15
Use c++latest in MSVC

Gymnasiast:


janisozaur:
I'm not a fan, it will cause failures when MSVC users start using post-17 features and they will get rejected by CI by other compilers. I'd rather move explicitly to 20 when all the CI jobs support it.

Gymnasiast:
For pull requests like https://github.com/OpenRCT2/OpenRCT2/pull/10707 , I would very much like to be able to use designated initializers, which are only supported this way.

janisozaur:
There was some discussion prior in https://github.com/OpenRCT2/OpenRCT2/pull/6034 and exactly the opposite PR to this merged a while ago: https://github.com/OpenRCT2/OpenRCT2/pull/7751

I uphold my point about moving to specific standard versions when they become available.

ZehMatt:
I'm with @janisozaur at this one, C++20 is very very new and I doubt we'll get a lot support especially on OSX/Android. I also can't wait for the switch, it will be great.

Gymnasiast:
@janisozaur I do see your point, it's just that I would really like to use designated initializers.

But if we don't merge this PR: would you be able to come up an alternative approach for PRs like https://github.com/OpenRCT2/OpenRCT2/pull/10707 ? (Some macro/C++ trickery perhaps?)

ZehMatt:
You could probably do it with some macro magic but I highly discourage from doing that. You could maybe design an intermediate format so you could load all that data also from a file without hard coding it but I'm not sure if thats wanted.

IntelOrca:
What about with a constexpr method that returns the struct where we can ensure the compiler serialises into raw data?

PR #10712, opened 2020-02-16
Replace typedef aliasing by using

tupaschoal:
Follow up to #10674


ZehMatt:
You might want to avoid changing the thirdparty files.

tupaschoal:
> changing

Argh, I had noticed this at some point and then forgot, thanks for checking, I reverted them.

tupaschoal:
Rebased.

PR #10715, opened 2020-02-16
Fix #10705: Apply multithreaded rendering to all viewports

### LINKED ISSUE #10705: Multithreading causes huge performance degrade on i9
**Processor:** i9-7980XE (18c/36t)

With one window open containing a viewport, and an average FPS of 200 on OpenGL. When enabling multithreading, FPS drops to 25.

This only happens unless there is more than one viewport open including the main viewport.

ZehMatt:
This fixes also the issue that it would reset the multithreading system every time any other viewport was rendered other than the main one which caused to always re-spawn threads every frame.

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -934,8 +934,6 @@ void viewport_paint(
     // this as well as the [x += 32] in the loop causes signed integer overflow -> undefined behaviour.
     const int16_t rightBorder = dpi1.x + dpi1.width;
     const int16_t alignedX = floor2(dpi1.x, 32);
-    const uint16_t columSize = rightBorder - alignedX;
-    const uint16_t columnCount = (columSize / 32);
 
     std::vector<paint_session*> columns;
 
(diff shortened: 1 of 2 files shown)


janisozaur:
Nice. I previously complained how multithreading didn't yield any speedup, especially on raspi, I would like to test it.

ZehMatt:
I almost forgot to mention, this also fixes the test part having 1 entry too much, it would always allocate +1, was that on purpose somehow?

IntelOrca:
This fixes the issue. Although with even with this many cores, the boost is not particularly high with multithreading enabled. The boost is higher for OpenGL than software rendering.

ZehMatt:
The reason for software being somewhat bad at performance is because of the bitmap copying and then applying the palette color, but frankly I have no idea how one could make this better.

IntelOrca:
@ZehMatt I understand that, I was just hoping that multithreading would make the whole process number of threads times faster.

IntelOrca:
I have already tested it and it is definitely an improvement, so :shipit:

ZehMatt:
Great, should I ignore the Travis builds?

IntelOrca:
Travis is dead to me, so sure.

src/openrct2/interface/Viewport.cpp
```diff
@@ -1,4 +1,4 @@
-/*****************************************************************************
+﻿/*****************************************************************************
```
Gymnasiast: BOM?

PR #10729, opened 2020-02-17
Extend object validation

janisozaur:
In attempt to solve quite commonly seen https://github.com/OpenRCT2/OpenRCT2/issues/10702, I wanted to have some systematic method of rejecting the objects that could lead to such crashes, this is my draft solution. I would like to hear your opinion before I dig myself too deep in it.

I tried reusing as much current code as possible without impacting performance, I came up with the `if constexpr` approach and I quite like it, though I had to duplicate some functions because of the different return types. My initial approach was to use `auto` deduced type, but I ran into some problems when there is `void` in the mix.

### DIFF
--- a/src/openrct2/object/RideObject.cpp
+++ b/src/openrct2/object/RideObject.cpp
@@ -147,6 +147,10 @@ void RideObject::ReadLegacy(IReadObjectContext* context, IStream* stream)
 
     GetImageTable().Read(context, stream);
 
+    if (!ValidateObject())
+    {
+        context->LogError(OBJECT_ERROR_BAD_IMAGE_TABLE, "Object failed to validate");
+    }
     // Validate properties
     if (_legacyType.excitement_multiplier > 75)
     {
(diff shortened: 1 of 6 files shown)


src/openrct2/object/RideObject.cpp
```diff
@@ -370,6 +374,53 @@ void RideObject::DrawPreview(rct_drawpixelinfo* dpi, [[maybe_unused]] int32_t wi
...
+    rct_drawpixelinfo dpi;
+    dpi.bits = pixels8;
+    dpi.pitch = 0;
+    dpi.x = 0;
+    dpi.y = 0;
+    dpi.width = max_width;
+    dpi.height = max_height;
+    dpi.zoom_level = 0;
+
+    uint32_t imageId = _legacyType.images_offset;
```
janisozaur: It was copied from https://github.com/OpenRCT2/OpenRCT2/blob/0c4623a39fa5dec8eac1c45f358e0fd45fd2f1df/src/openrct2/object/RideObject.cpp#L360-L368 and this is likely wrong/insufficient, would appreciate hints on how to get correct values.

Gymnasiast:
What happens if it "rejects" an object? Will it prevent the file from being loaded, or will it just not draw the object?

janisozaur:
I would reject it, as in "not load, provide error message". This aims at addressing crashes with given object, I argue that not having an object is better than rogue object taking down the entire game.

ZehMatt:
How would I actually test this, do you have some of those corrupt objects that causes some of the mentioned issues?

src/openrct2/object/RideObject.cpp
```diff
@@ -370,6 +374,53 @@ void RideObject::DrawPreview(rct_drawpixelinfo* dpi, [[maybe_unused]] int32_t wi
...
+        {
+            max_width = GetImageTable().GetImages()[i].width;
+        }
+    }
+    if (max_width < 0 || max_height < 0)
+    {
+        return false;
+    }
+    size_t numPixels = max_width * max_height;
+    auto pixels8 = new uint8_t[numPixels];
```
ZehMatt: Use vector or unique_ptr.

Gymnasiast:
> I would reject it, as in "not load, provide error message". This aims at addressing crashes with given object, I argue that not having an object is better than rogue object taking down the entire game.

Do you mean not loading the object, or also not loading the entire save? The latter could potentially affect a large number of saves.

janisozaur:
You mean the saves that crash OpenRCT2? What do we do with saves that don't have required objects now? Iirc display a window with offender list, right?

That said I get your sentiment and if you come up with a decent way of fixing up some of the errors (I suppose not all the errors are fixable) then it could be included

janisozaur:
@zehmatt I don't think I have any such object at hand right now, but that can be addressed somehow. I will provide some way of testing it on real data by the time I un-wip it. I would like to add a cli for validating a given object in a future commit.

janisozaur:
I was also wondering if we can extract the object data from dump on backtrace. Since we don't have access to actual minidump, I think I would like to investigate possibility of including offending obj in files sent to backtrace.

src/openrct2/object/RideObject.cpp
```diff
@@ -370,6 +374,53 @@ void RideObject::DrawPreview(rct_drawpixelinfo* dpi, [[maybe_unused]] int32_t wi
...
+
+    for (auto rideType : _legacyType.ride_type)
+    {
+        if (rideType != RIDE_TYPE_NULL)
+            break;
+        else
+            imageId++;
+    }
+
+    delete[] pixels8;
```
duncanspumpkin: You are deleting the ptr that you are about to draw to.

src/openrct2/object/RideObject.cpp
duncanspumpkin: Looks fine to me. This will slip over all the preview images for the ride and get to the sprites

duncanspumpkin:
Is there not already some sort of validate that checks you can't make a profit deleting scenery.

rik-smeets:
Since this PR hasn't seen any activity in 2+ years and the referenced issue has been closed, I'm closing this.

PR #10733, opened 2020-02-17
Add mingw to GitHub CI

IntelOrca:


.github/workflows/ci.yml
```diff
@@ -57,6 +57,20 @@ jobs:
...
+    name: Windows (win32) using mingw
+    runs-on: ubuntu-latest
+    container:
+      image: openrct2/openrct2-build:0.2.4-mingw
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Build OpenRCT2
+        run: |
+          sudo su
```
tupaschoal: Why does it need `sudo`?

IntelOrca: Ugh, well the container switches to a new user and that has a different uid / gid to the github actions checkout directory so you get permission issues. My quick fix was to change back to root.

Unfortunately the image we build in does not build anymore due to its third party dependencies changing or not existing so I can't make changes to it. This ideally needs to be resolved.

tupaschoal: I see, that's unfortunate, thanks for the explanation :)

.github/workflows/ci.yml
```diff
@@ -57,6 +57,20 @@ jobs:
...
               echo 'Not going to push build'
           fi
+  windows-mingw:
+    name: Windows (win32) using mingw
+    runs-on: ubuntu-latest
+    container:
+      image: openrct2/openrct2-build:0.2.4-mingw
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
```
Margen67: v2 should be used.

AaronVanGeffen: I had to look it up, so fwiw: https://github.com/actions/checkout#checkout-v2

PR #10739, opened 2020-02-18
Mountain tool overlay in land tool window

AaronVanGeffen:
When the mountain is activated, the old 'tool size 0' is overlaid over the selection grid. This works fine for odd-numbered selections, but not for even-numbered ones:

mountain_misaligned_overlap_odd mountain_misaligned_overlap_even

To fix this, I have made two new sprites for the mountain tool grid: one for odd numbers, and one for even. The result is a much clearer mountain tool:

mountain_aligned_overlap_odd mountain_aligned_overlap_even

The keen-eyed will spot a few stray pixels overlaying the [+] and [-] buttons. This was already present in the old ('odd') situation, but even more so in the new 'even' one. It is my intention to adjust both sprites such that they no longer overlap. (Alternatively, we could paint the buttons twice, but that seems a hack-ish solution to me.)

For what it's worth, the new sprites by themselves (with a solid background):

_mountain_tool_odd_bg _mountain_tool_even_bg

~~(This PR is intended to be merged after #10735, hence the temporary first commit. I will rebase eventually.)~~


tupaschoal:
Is the number of the PR you quoted at then end correct?

src/openrct2-ui/windows/Land.cpp
```diff
@@ -354,7 +354,10 @@ static void window_land_paint(rct_window* w, rct_drawpixelinfo* dpi)
     {
         x = w->x + previewWidget->left;
         y = w->y + previewWidget->top;
-        gfx_draw_sprite(dpi, SPR_LAND_TOOL_SIZE_0, x, y, 0);
+        if (gLandToolSize % 2 == 1)
```
tupaschoal: Maybe we should use `is_power_of_2(gLandToolSize)` instead. Then I'd advise swapping the statements so we don't have to negate.

Or we can do a ternary:
```c++
gfx_draw_sprite(dpi, is_power_of_2(gLandToolSize) ? SPR_G2_MOUNTAIN_TOOL_EVEN : SPR_G2_MOUNTAIN_TOOL_ODD, x, y, 0);
```

duncanspumpkin: Is power of two isn't the same 2, 4, 8,... This should be 2,4,6,8,...

Broxzier: I'd prefer keeping it as is over a single one-liner. The compiler is probably smart enough to check the least significant bit and optimize both calls into one anyway.

AaronVanGeffen: I wouldn't mind determining the sprite to use in a separate statement, either. But I agree with Broxzier in that I would rather not use ternary expressions inside a function call.

AaronVanGeffen:
Good question. The old mountain tool sprite included the 1x1 selection (as you may remember from vanilla). While not strictly necessary, the new sprite removes this selection.

On top of that:
> The keen-eyed will spot a few stray pixels overlaying the [+] and [-] buttons. This was already present in the old ('odd') situation, but even more so in the new 'even' one. It is my intention to adjust both sprites such that they no longer overlap.

Which means we will need new a sprite, anyway, unless:
> (Alternatively, we could paint the buttons twice, but that seems a hack-ish solution to me.)

src/openrct2-ui/windows/Land.cpp
```diff
@@ -356,6 +356,8 @@ static void window_land_paint(rct_window* w, rct_drawpixelinfo* dpi)
         y = w->y + previewWidget->top;
         int32_t sprite = gLandToolSize % 2 == 0 ? SPR_G2_MOUNTAIN_TOOL_EVEN : SPR_G2_MOUNTAIN_TOOL_ODD;
         gfx_draw_sprite(dpi, sprite, x, y, 0);
+        widget_draw(dpi, w, WIDX_DECREMENT);
```
Gymnasiast: I thought you wanted to tweak the sprites, rather than draw twice?

AaronVanGeffen: That was my original intent, yes, but I realised that it might be problematic if we ever want to adjust the container width. Suddenly, there would be pixels missing. This is a solution without a noticeable performance impact, anyway.

PR #10743, opened 2020-02-18
Fix #10740: Crash using benchspritegfx with a dimension not dividable by 32

### LINKED ISSUE #10740: BACKTRACE b1ebfb
<table>
    <tr>
        <td></td>
        <td><a href='https://openrct2.sp.backtrace.io/dashboard/openrct2/project/openrct2/group/b1ebfb013e4626c40dfb817faff6384441b90320a1fd3f382434f372a83371ae?_t=W3sibmFtZSI6IndvcmtmbG93In1d'><b>New unique error: [openrct2]</b></a><br /><b>1</b> occurrence(s) <b></b> since (Tue Feb 18 2020 22:28:56)<br /></td>
    </tr>
</table>

<table>
<tr><td><strong>Application</strong></td><td>openrct2.com</td></tr>

<tr><td><strong>Attributes</strong></td><td><strong>attachment_park.sv6.gz:</strong> C:\\Users\\andrzej\\Documents\\OpenRCT2\\f9d65173-cab3-4ab7-8688-7a118b935131.sv6.gz: <em>1</em><br><strong>attachment_park.sv6:</strong> *: <em>1</em><br><strong>attachment_config.ini:</strong> C:\\Users\\andrzej\\Documents\\OpenRCT2\\f9d65173-cab3-4ab7-8688-7a118b935131.ini: <em>1</em><br><strong>assert_failure:</strong> *: <em>1</em><br><strong>attachment_screenshot.png:</strong> *: <em>1</em><br><strong>commit:</strong> a8b5304: <em>1</em><br></td></tr>
<tr><td><strong>Classifiers</strong></td><td><strong>invalid-write</strong> </td></tr>
<tr><td><strong>Frames</strong></td><td><code>SetTEBLangID</code></td></tr>
</table>

ZehMatt:
Closes #10740

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -952,7 +952,7 @@ void viewport_paint(
     if (recorded_sessions != nullptr)
     {
         const uint16_t columnSize = rightBorder - alignedX;
-        const uint16_t columnCount = (columnSize / 32);
+        const uint16_t columnCount = (columnSize + 31) / 32;
         recorded_sessions->resize(columnCount);
     }
 


src/openrct2/interface/Viewport.cpp
```diff
@@ -952,7 +952,7 @@ void viewport_paint(
     if (recorded_sessions != nullptr)
     {
         const uint16_t columnSize = rightBorder - alignedX;
-        const uint16_t columnCount = (columnSize / 32);
+        const uint16_t columnCount = (columnSize / 32) + (columnSize % 32 == 0 ? 0 : 1);
```
tupaschoal: The problematic scenario is when `columCount == 0`, is it not? This change is making `columnCount = (columnSize / 32) + 1` for every `columnSize` above `32` (excluding multiples of 32), along with smaller than `32`. Why not just:
```suggestion
        const uint16_t columnCount = std::max((columnSize / 32), 1);
```

ZehMatt: I don't quite follow. It has to only add +1 when there is a remainder, ex 33. If its 0 then it should stay at 0.

Gymnasiast: @tupaschoal If you divide integers in C, it will always floor them. So `31 / 32` = 0. What you propose might work if you convert the lot to float, but that comes with its own set of nasties.

The code seems correct, the only thing that I would change is stick a comment above saying `Divide and round up` or something like that.

AaronVanGeffen: @Gymnasiast I don't think that's what he's proposing at all? `std::max` would work with the same integer division you've outlined.

Gymnasiast: @AaronVanGeffen It won't produce the same results as those of ZehMatt. Take a `columnSize` of 48 for example - Matt's code will return 2, while Tulio's will return 1.

tupaschoal: I was assuming it was crashing only when `columnCount` ended up being 0, hence the max proposition that would always return at least 1.

IntelOrca: If you want to round up why not do: `(columnSize + 31) / 32`?

janisozaur: Yup, @IntelOrca's is the canonical version of ceiling division.

src/openrct2/interface/Viewport.cpp
```diff
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
```
ZehMatt: I'm not sure what this change is about, locally it doesn't exist.

AaronVanGeffen: Byte order mark?

ZehMatt: Hm apparently clang-format is doing this, I have no control over it as any other clang version is not really working.

Gymnasiast: Load it up in a hex editor and it should become clear very soon.

ZehMatt: It is BOM but that is due to clang-format modifying it, initially its just UTF8 after clang-format its UTF8-BOM

IntelOrca: How are you using clang-format? Doesn't that output to stdout therefore leaving it to the caller to handle?

PR #10802, opened 2020-02-21
Fix #10713: Correctly encode network ride id over network

### LINKED ISSUE #10713: Rides are disappeared when it is built in multiplayer
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 38907fd

Rides are disappears when build another rides. It doesn't matter it is a flat rides or track rides.
I think it is occured from 723f658. It is fine at df28bca

**Steps to reproduce:**
1. Join any multiplayer server
2. Build any ride
3. Build another ride.

**Screenshots / Video:**

d

ZehMatt:
It byte swapped uint16_t and then during decode byte swapped it as uint32_t, now it always uses uint32_t for ride/player id. I could use the underlying type but that would require to re-create all replays again.

Closes #10713

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -141,14 +141,15 @@ template<> struct DataSerializerTraits<NetworkPlayerId_t>
 {
     static void encode(IStream* stream, const NetworkPlayerId_t& val)
     {
-        uint32_t temp = ByteSwapBE(val.id);
+        uint32_t temp = static_cast<uint32_t>(val.id);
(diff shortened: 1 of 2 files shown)


src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -1,4 +1,4 @@
-/*****************************************************************************
+﻿/*****************************************************************************
```
IntelOrca: BOM?

ZehMatt: God damn Visual Studio

PR #10804, opened 2020-02-21
Use CoordsXY(Z) in more places

tupaschoal:
I recommend viewing it commit by commit, the first two are trivial, the last is a bit bigger.

Regarding the last, I was wondering if `news_item_get_subject_location()` shouldn't just receive a `News*` or `News&` if everyone always send the same stuff.

### DIFF
--- a/src/openrct2-ui/windows/Footpath.cpp
+++ b/src/openrct2-ui/windows/Footpath.cpp
@@ -1057,26 +1057,24 @@ static void footpath_remove_tile_element(TileElement* tileElement)
 static TileElement* footpath_get_tile_element_to_remove()
 {
     TileElement* tileElement;
-    int32_t x, y, z, zLow;
+    int32_t z, zLow;
 
-    x = gFootpathConstructFromPosition.x / 32;
-    y = gFootpathConstructFromPosition.y / 32;
-    if (x >= 256 || y >= 256)
+    if (map_is_location_valid(gFootpathConstructFromPosition))
     {
         return nullptr;
     }
 
-    z = (gFootpathConstructFromPosition.z >> 3) & 0xFF;
-    zLow = z - 2;
+    z = std::min(255 * COORDS_Z_STEP, gFootpathConstructFromPosition.z);
+    zLow = z - (2 * COORDS_Z_STEP);
 
-    tileElement = map_get_first_element_at(TileCoordsXY{ x, y }.ToCoordsXY());
+    tileElement = map_get_first_element_at(gFootpathConstructFromPosition);
     do
     {
         if (tileElement == nullptr)
             break;
         if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH)
         {
-            if (tileElement->base_height == z)
+            if (tileElement->GetBaseZ() == z)
             {
                 if (tileElement->AsPath()->IsSloped())
                 {
@@ -1088,7 +1086,7 @@ static TileElement* footpath_get_tile_element_to_remove()
 
                 return tileElement;
             }
-            else if (tileElement->base_height == zLow)
+            else if (tileElement->GetBaseZ() == zLow)
             {
                 if (!tileElement->AsPath()->IsSloped())
                 {
--- a/src/openrct2-ui/windows/Footpath.cpp
+++ b/src/openrct2-ui/windows/Footpath.cpp
@@ -1059,7 +1059,7 @@ static TileElement* footpath_get_tile_element_to_remove()
     TileElement* tileElement;
     int32_t z, zLow;
 
-    if (map_is_location_valid(gFootpathConstructFromPosition))
+    if (!map_is_location_valid(gFootpathConstructFromPosition))
     {
         return nullptr;
     }
--- a/src/openrct2-ui/windows/GameBottomToolbar.cpp
+++ b/src/openrct2-ui/windows/GameBottomToolbar.cpp
@@ -192,17 +192,15 @@ static void window_game_bottom_toolbar_mouseup(rct_window* w, rct_widgetindex wi
 
             {
                 newsItem = news_item_get(0);
-                int32_t x, y, z;
-                int32_t subject = newsItem->Assoc;
 
-                news_item_get_subject_location(newsItem->Type, subject, &x, &y, &z);
+                auto subjectLoc = news_item_get_subject_location(newsItem->Type, newsItem->Assoc);
 
-                if (x == LOCATION_NULL)
+                if (!subjectLoc)
                     break;
 
                 rct_window* mainWindow = window_get_main();
                 if (mainWindow != nullptr)
-                    window_scroll_to_location(mainWindow, x, y, z);
+                    window_scroll_to_location(mainWindow, subjectLoc->x, subjectLoc->y, subjectLoc->z);
             }
             break;
         case WIDX_RIGHT_OUTSET:
@@ -333,11 +331,9 @@ static void window_game_bottom_toolbar_invalidate(rct_window* w)
         w->disabled_widgets &= ~(1 << WIDX_NEWS_LOCATE);
 
         // Find out if the news item is no longer valid
-        int32_t y, z;
-        int32_t subject = newsItem->Assoc;
-        news_item_get_subject_location(newsItem->Type, subject, &x, &y, &z);
+        auto subjectLoc = news_item_get_subject_location(newsItem->Type, newsItem->Assoc);
 
-        if (x == LOCATION_NULL)
+        if (!subjectLoc)
             w->disabled_widgets |= (1 << WIDX_NEWS_LOCATE);
 
         if (!(news_type_properties[newsItem->Type] & NEWS_TYPE_HAS_SUBJECT))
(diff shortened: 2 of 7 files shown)


src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -1057,26 +1057,24 @@ static void footpath_remove_tile_element(TileElement* tileElement)
...
-    if (x >= 256 || y >= 256)
+    if (map_is_location_valid(gFootpathConstructFromPosition))
     {
         return nullptr;
     }
 
-    z = (gFootpathConstructFromPosition.z >> 3) & 0xFF;
-    zLow = z - 2;
+    z = gFootpathConstructFromPosition.z & std::numeric_limits<int32_t>::max();
+    zLow = z - (2 * COORDS_Z_STEP);
```
tupaschoal: This code can be a lot simpler if we find out the bitwise and is not necessary.

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -1057,26 +1057,24 @@ static void footpath_remove_tile_element(TileElement* tileElement)
...
-    y = gFootpathConstructFromPosition.y / 32;
-    if (x >= 256 || y >= 256)
+    if (map_is_location_valid(gFootpathConstructFromPosition))
     {
         return nullptr;
     }
 
-    z = (gFootpathConstructFromPosition.z >> 3) & 0xFF;
-    zLow = z - 2;
+    z = gFootpathConstructFromPosition.z & std::numeric_limits<int32_t>::max();
```
Gymnasiast: This is not the same. There has to be a reason it maxes out at 255, and you just raised that limit enormously.

src/openrct2/management/NewsItem.cpp
```diff
@@ -207,73 +207,72 @@ static int32_t news_item_get_new_history_slot()
  *
  *  rct2: 0x0066BA74
  */
-void news_item_get_subject_location(int32_t type, int32_t subject, int32_t* x, int32_t* y, int32_t* z)
+CoordsXYZ news_item_get_subject_location(int32_t type, int32_t subject)
```
Gymnasiast: It might be better to use std::optional, rather than checking for `isNull()`.

src/openrct2-ui/windows/Footpath.cpp
duncanspumpkin: Yeah it should be something like `z = std::min(255 * COORDS_Z_STEP, gFootpathConstructFromPosition.z)`

src/openrct2-ui/windows/Footpath.cpp
duncanspumpkin: The & is most likely due to dl being traditionally the register for base height. Its not required but the old code had the other feature of maxing out at 255 (smallZ) which we will need to replicate.

src/openrct2/management/NewsItem.cpp
```diff
@@ -207,73 +207,72 @@ static int32_t news_item_get_new_history_slot()
...
             if (ride == nullptr || !(ride->lifecycle_flags & RIDE_LIFECYCLE_ON_TRACK))
             {
-                *x = LOCATION_NULL;
+                subjectLoc.setNull();
                 break;
             }
 
             // Find the first car of the train peep is on
-            vehicle = &(get_sprite(ride->vehicles[peep->current_train])->vehicle);
+            Vehicle* vehicle = &(get_sprite(ride->vehicles[peep->current_train])->vehicle);
```
duncanspumpkin: can you change the  `&(get_sprite(ride->vehicles[peep->current_train])->vehicle)` to `GET_VEHCILE(ride->vehicles[peep->current_train])`. I'm going to change it anyway in a latter refactor but this would look neater.

src/openrct2/management/NewsItem.cpp
```diff
@@ -207,73 +207,72 @@ static int32_t news_item_get_new_history_slot()
...
             // Find the actual car peep is on
             for (int32_t i = 0; i < peep->current_car; i++)
             {
                 vehicle = &(get_sprite(vehicle->next_vehicle_on_train)->vehicle);
```
duncanspumpkin: as above `GET_VEHICLE`

src/openrct2/management/NewsItem.cpp
```diff
@@ -207,73 +207,72 @@ static int32_t news_item_get_new_history_slot()
...
+            Peep* peep = GET_PEEP(subject);
+            subjectLoc = CoordsXYZ{ peep->x, peep->y, peep->z };
             break;
+        }
         case NEWS_ITEM_BLANK:
-            *x = subject;
-            *y = subject >> 16;
-            *z = tile_element_height({ *x, *y });
+        {
+            auto subjectXY = CoordsXY{ subject, subject >> 16 };
```
duncanspumpkin: This look wrong. I think someone has broken this in the past. I'm going to guess subject in this case has been created from packing to int16_t's into a int32_t. Eitherway this new code and the old code is wrong because x needs to be &'ed with 0xFFFF.

src/openrct2-ui/windows/Footpath.cpp
tupaschoal: Somehow I thought it would be equivalent, I'll update

src/openrct2/management/NewsItem.cpp
tupaschoal: And what to do with the `y`?

Gymnasiast: No need to change the y, as you already did a bitshift by 16, which ensures that the upper 16 bits are 0 anyway.

tupaschoal:
Fixed the issues

src/openrct2/management/NewsItem.cpp
```diff
@@ -207,73 +207,70 @@ static int32_t news_item_get_new_history_slot()
...
+            Peep* peep = GET_PEEP(subject);
+            subjectLoc = CoordsXYZ{ peep->x, peep->y, peep->z };
             break;
+        }
         case NEWS_ITEM_BLANK:
-            *x = subject;
-            *y = subject >> 16;
-            *z = tile_element_height({ *x, *y });
+        {
+            auto subjectXY = CoordsXY{ subject & 0xFF, subject >> 16 };
```
duncanspumpkin: `& 0xFFFF`

tupaschoal: Done

src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -1057,26 +1057,24 @@ static void footpath_remove_tile_element(TileElement* tileElement)
 static TileElement* footpath_get_tile_element_to_remove()
 {
     TileElement* tileElement;
-    int32_t x, y, z, zLow;
+    int32_t z, zLow;
 
-    x = gFootpathConstructFromPosition.x / 32;
-    y = gFootpathConstructFromPosition.y / 32;
-    if (x >= 256 || y >= 256)
+    if (map_is_location_valid(gFootpathConstructFromPosition))
```
Gymnasiast: You reversed the condition here!

src/openrct2-ui/windows/GameBottomToolbar.cpp
```diff
@@ -333,11 +331,9 @@ static void window_game_bottom_toolbar_invalidate(rct_window* w)
         w->disabled_widgets &= ~(1 << WIDX_NEWS_LOCATE);
 
         // Find out if the news item is no longer valid
-        int32_t y, z;
-        int32_t subject = newsItem->Assoc;
-        news_item_get_subject_location(newsItem->Type, subject, &x, &y, &z);
+        auto subjectLoc = news_item_get_subject_location(newsItem->Type, newsItem->Assoc);
 
-        if (x == LOCATION_NULL)
+        if (!subjectLoc)
```
Gymnasiast: I prefer an explicit comparison (with nullopt), in the same way we explicitly compare to nullptr.

src/openrct2-ui/windows/News.cpp
```diff
@@ -165,10 +165,10 @@ static void window_news_update(rct_window* w)
             }
             else if (w->news.var_482 > 1)
             {
-                news_item_get_subject_location(newsItem->Type, newsItem->Assoc, &x, &y, &z);
-                if (x != LOCATION_NULL && (w = window_get_main()) != nullptr)
+                auto subjectLoc = news_item_get_subject_location(newsItem->Type, newsItem->Assoc);
+                if (subjectLoc && (w = window_get_main()) != nullptr)
```
Gymnasiast: See previous remark.

src/openrct2/management/NewsItem.cpp
```diff
@@ -207,73 +207,70 @@ static int32_t news_item_get_new_history_slot()
...
-            *y = peep->y;
-            *z = peep->z;
-            if (*x != LOCATION_NULL)
+        {
+            Peep* peep = GET_PEEP(subject);
+            subjectLoc = CoordsXYZ{ peep->x, peep->y, peep->z };
+            if (subjectLoc->x != LOCATION_NULL)
                 break;
 
             if (peep->state != 3 && peep->state != 7)
```
Gymnasiast: Do we have constants for these? If so, please use them.

tupaschoal: I don't know, I can update them on another PR, as I didn't touch these lines, so I think it'd be weird.

Gymnasiast: I'd count it as "Boy Scout principle", you're modifying the function anyway.

I'm ok with doing them in another PR, though.

PR #10807, opened 2020-02-21
Add two new zoom in levels

IntelOrca:
I have started work on adding two new zoom in levels but it currently only works for OpenGL. I have stubs for the software rendering as it will require probably a full re-write to magnify the game. Would anyone be able to help with this?

Another thing I can't get working is the money text at the right position on any zoom level that isn't 0. This currently doesn't work, not sure it worked in vanilla or not.

- [x] BMP magnify
- [x] RLE magnify
- [x] Rain
- [x] Zoom to cursor
- [x] Window divide remainder (visible when opening a drop down)
- [x] Floating money text

**Software only:**
- [ ] Viewport scrolling issue
- [ ] Window divide remainder (visible when opening a drop down)


pizza2004:
Does this allow closer or farther levels of zoom?  And if you mean the floating money effect, I believe that wasn't in any version of vanilla.

IntelOrca:
Two closer zooms.
Floating red money.

Gymnasiast:
@pizza2004 Not sure what you mean - floating money was present in the first version of RCT1.

pizza2004:
I thought he meant the floating text that appears when peeps spend money.  I'm working on Xcode so I can test the build myself.  Finally got a new MacBook Pro so maybe I can try helping with development again.

ocalhoun6:
> Two closer zooms.

If the new save format is going to allow much larger maps, it might be a good idea to expand this project into farther zooms as well. Players might want to zoom out much farther in order to better view a much larger map.

(Probably shouldn't actually implement the super-far zoom levels yet, though ... but just have them ready to go when the maximum map size is increased.)

IntelOrca:
> it might be a good idea to expand this project into farther zooms as well.

Higher zooms will be very expensive, frame rate will drop a lot.

src/openrct2/interface/ZoomLevel.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+#include <cstdint>
+
+struct ZoomLevel
+{
+private:
+    int8_t _level{};
+
+public:
+    constexpr ZoomLevel()
```
ZehMatt: You can use default here, _level is defaulted already.

janisozaur:
@pizza2004 if you're willing to help, you can always submit commits to fix the Xcode project file. Usually that involves just adding a new source file to the build for PRs marked as https://github.com/OpenRCT2/OpenRCT2/labels/Xcode%20Fix%20Required

pizza2004:
Well, making git work was a pain, but I did the Xcode project.

Is it freezing on the software renderer as well for you?

Umdlye:
I don't know if it's too early to report bugs but rain doesn't appear correctly on the new zoom levels. https://i.imgur.com/LNb7yUT.mp4

Windows 10 (x64)

mrtnptrs:
I encounter a few issues with this amazing WIP build:
- @pizza2004 Yes, also for me as soon as I zoom in on the software (hardware) renderer the game freezes. This doesn't happen of course on OpenGL. 
-Also when I was zoomed in to one of the two new zoom-levels while using the OpenGL-rederer and then switched to a software-renderer, the game crashes. It seems like it then tries to go to that closer zoom-level, which isn't available yet of course. So, this would be fixed of course when it also works in the software-renderer. But maybe a safeguard could be added that the game automatically zooms to the closest correct zoom-level if the current one is not recognized by the renderer? Maybe avoids future crashes whenever you guys wanna maybe implement another renderer.

AaronVanGeffen:
This is very exciting!

It looks like zooming to cursor doesn't work correctly when switching between the new zoom levels. Perhaps just an issue with the bit shifting?

janisozaur (approved):

src/openrct2/drawing/Drawing.h
```diff
@@ -108,7 +109,14 @@ struct rct_drawpixelinfo
     int16_t pitch{}; // note: this is actually (pitch - width)
     ZoomLevel zoom_level{};
 
+    uint8_t remX{};
```
Gymnasiast, line 112: What does this variable hold? It isn't clear from the name or context, and there is no comment either.

Gymnasiast (approved):

tupaschoal:
Should be good to go, just need solving of the merge conflicts

PR #10818, opened 2020-02-22
Use ScreenCoordsXY for some window interactions

tupaschoal:


src/openrct2/interface/Viewport.cpp
```diff
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
```
tupaschoal: I guess I inherited @ZehMatt's curse

ZehMatt: This happened ever since we switched to VS2019, I never had this once before that. I installed some extension called "Fix file encoding" and it seems to work 50% of the time, so you might wanna give that a try.

tupaschoal:
Should be fine now.

PR #10820, opened 2020-02-22
Fix YaHei detection

Gymnasiast:
On Linux at least, the old name caused both YaHei and SimSun to be regarded as "not present".

I would like someone on Windows to verify that this does not break anything. Perhaps @telk5093 could help?


AaronVanGeffen:
Only the filename is relevant to Windows. In contrast, the font name is the only thing relevant to Linux and macOS.

In other words, this change should not negatively impact Windows.

telk5093:
I think it has no problems with simplified chinese:
image
(I found my _C:\Windows\Fonts_ directory, I found ``msyhl.ttc``, ``msyhbd.ttc`` and ``msyh.ttc``. It is collected as ``Microsoft YaHei``)

Issue #682, opened 2015-01-12
Underwater Patrol Area Difficult to Work With


pizza2004:
I know this isn't necessarily a priority, but I've been playing RCT lately and in order to maximize my space I've built out over water with paths and coasters and when trying to set patrol areas for my staff I've run into a couple of issues:

1) It's impossible to properly distinguish between the patrol area set for your currently selected staff member and the gray version representing what's been set before.

2) It's very difficult to tell whether you've set the patrol area in a way that encompasses whatever path is up on supports over the water.

My suggestion is that you just show the patrol area on top of the water instead of under it, and make it so when you click on the water it selects the square you clicked on, instead of whatever land is visible through the one you clicked on.  I feel like this change could make setting patrol areas for staff much much easier.  I honestly don't mind other ways of handling the problem either, I just think it's something that should be addressed.

RollingStar:
Would another color of patrol area work better over water? If the water is blue, for example, maybe the patrol area marker could be red when placed over water, to provide better contrast. But then it may create user confusion to have two colors for the same thing.

Note that there are 5 water color options.

IntelOrca:
@RollingStar I think the issue is from an input point of view. It would be easier to click the surface of the water and have the surface tile be the tile you clicked on for the patrol area where as currently you have to click the land tile the water surface tile is over which is not straight forward particular if the water is deep.

pizza2004:
Implemented by #10830.

PR #10830, opened 2020-02-23
Implement #682: Staff patrol area now drawn on the water

Gymnasiast:
Diamond Heights 2020-02-23 11-51-14

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -28,6 +28,7 @@
 - Fix: [#10705] Apply multithreaded rendering to all viewports.
 - Fix: [#10739] Mountain tool overlay for even-numbered selections.
 - Fix: [#10752] Mute button state not correctly set at startup.
+- Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
 - Removed: [#6898] LOADMM and LOADRCT1 title sequence commands (use LOADSC instead).
 
 0.2.4 (2019-10-28)
--- a/src/openrct2/paint/tile_element/Paint.Surface.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Surface.cpp
@@ -902,6 +902,36 @@ static void viewport_surface_draw_water_side_top(
     viewport_surface_draw_tile_side_top(session, edge, height, terrain, self, neighbour, true);
 }
 
+static std::pair<int32_t, int32_t> surface_get_height_above_water(
+    const SurfaceElement& surfaceElement, const int32_t height, const int32_t surfaceShape)
+{
+    int32_t local_surfaceShape = surfaceShape;
+    int32_t local_height = height;
+
+    if (surfaceElement.GetWaterHeight() > 0)
+    {
+        int32_t waterHeight = surfaceElement.GetWaterHeight();
+        if (waterHeight > height)
+        {
+            local_height += (2 * COORDS_Z_STEP);
+
+            if (waterHeight != local_height || !(local_surfaceShape & TILE_ELEMENT_SURFACE_DIAGONAL_FLAG))
+            {
+                local_height = waterHeight;
+                local_surfaceShape = 0;
+            }
+            else
+            {
+                int32_t bl = (surfaceShape ^ 0xF) << 2;
+                int32_t bh = bl >> 4;
+                local_surfaceShape = (bh & 0x3) | (bl & 0xC);
+            }
+        }
+    }
+
+    return { local_height, local_surfaceShape };
+}
+
 /**
  *  rct2: 0x0066062C
  */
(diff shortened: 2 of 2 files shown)


mrtnptrs:
You references to the wrong issue: should be 682 instead of 982.

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -902,6 +902,38 @@ static void viewport_surface_draw_water_side_top(
     viewport_surface_draw_tile_side_top(session, edge, height, terrain, self, neighbour, true);
 }
 
+static std::pair<int32_t, int32_t> surface_get_height_above_water(
+    const TileElement& tileElement, const int32_t height, const int32_t surfaceShape)
+{
+    int32_t local_surfaceShape = surfaceShape;
+    int32_t local_height = height;
+
+    if (tileElement.AsSurface()->GetWaterHeight() > 0)
```
IntelOrca: Since you are expecting the tile element to be a surface, the parameter into the method should just be `SurfaceTileElement`.

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -902,6 +902,38 @@ static void viewport_surface_draw_water_side_top(
...
+            {
+                local_height = waterHeight;
+                local_surfaceShape = 0;
+            }
+            else
+            {
+                int16_t bl, bh;
+
+                bl = (surfaceShape ^ 0xF) << 2;
+                bh = bl >> 4;
```
IntelOrca:
```suggestion
                int32_t bl = (surfaceShape ^ 0xF) << 2;
                int32_t bh = bl >> 4;
```

distribution/changelog.txt
```diff
@@ -28,6 +28,7 @@
 - Fix: [#10705] Apply multithreaded rendering to all viewports.
 - Fix: [#10739] Mountain tool overlay for even-numbered selections.
 - Fix: [#10752] Mute button state not correctly set at startup.
+- Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
```
AaronVanGeffen: I think you can file this as a feature, tbh. That, or move it up a bit, so it doesn't get lost in the sea of bugfixes.

Gymnasiast: It's not really a feature, though, in that it doesn't give you any extra possibilities or even information - it just makes it easier to see.

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -902,6 +902,36 @@ static void viewport_surface_draw_water_side_top(
     viewport_surface_draw_tile_side_top(session, edge, height, terrain, self, neighbour, true);
 }
 
+static std::pair<int32_t, int32_t> surface_get_height_above_water(
+    const SurfaceElement& surfaceElement, const int32_t height, const int32_t surfaceShape)
+{
+    int32_t local_surfaceShape = surfaceShape;
```
duncanspumpkin: use camelCase for local variables.

distribution/changelog.txt
AaronVanGeffen: Alright. Semantics aside, a lot of people will like this change, though. I really think improvements ought to be higher up in the changelog.

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -902,6 +902,36 @@ static void viewport_surface_draw_water_side_top(
...
+
+            if (waterHeight != local_height || !(local_surfaceShape & TILE_ELEMENT_SURFACE_DIAGONAL_FLAG))
+            {
+                local_height = waterHeight;
+                local_surfaceShape = 0;
+            }
+            else
+            {
+                int32_t bl = (surfaceShape ^ 0xF) << 2;
+                int32_t bh = bl >> 4;
```
duncanspumpkin: This is just a 4 bit rotate should use the rotate macros.

distribution/changelog.txt
pizza2004: If this one were moved now it would mean that either all other improvements in the changefile needed to be moved or just that the ordering would be different than it always has been on this one.

AaronVanGeffen: I'm fairly sure the way @Gymnasiast filed it is consistent with how it was done before. However, I believe that's a historical mistake that we should rectify at some point. :wink:

pizza2004: I actually originally said the same about him following the conventions but edited my comment because I scrolled through the whole file and noticed that ordering and naming (some just say Improve) have changed over time once or twice.

src/openrct2/paint/tile_element/Paint.Surface.cpp
Gymnasiast: So if I'm reading this correctly, I'd have to replace the entire else clause with
```
local_surfaceShape = ror4(surfaceShape ^ 0xF, 2);
```
?

duncanspumpkin: yeah that would be the same I believe.

Gymnasiast: https://github.com/OpenRCT2/OpenRCT2/pull/10853

PR #10846, opened 2020-02-29
Prefer const ref to send ScreenCoordsXY over

tupaschoal:
Most of these places are just changing `ScreenCoordsXY` to `const ScreenCoordsXY&`. Some have to create new variables because they used to modify the received copy, but there were less than I feared and it seems like a welcome change anyway.


AaronVanGeffen:
Is there a noticeable (positive) performance impact to these changes?

src/openrct2-ui/windows/EditorInventionsList.cpp
```diff
@@ -786,16 +788,17 @@ static void window_editor_inventions_list_drag_cursor(
...
  *  rct2: 0x00685412
  */
-static void window_editor_inventions_list_drag_moved(rct_window* w, ScreenCoordsXY screenCoords)
+static void window_editor_inventions_list_drag_moved(rct_window* w, const ScreenCoordsXY& screenCoords)
 {
     ResearchItem* researchItem;
 
     int32_t scrollId;
     // Skip always researched items, so that the dragged item gets placed underneath them
+    auto newScreenCoords = screenCoords;
```
Gymnasiast: Is there any advantage to doing this over leaving the signature non-const?

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1688,12 +1688,13 @@ static void map_window_set_pixels(rct_window* w)
...
+static CoordsXY map_window_screen_to_map(const ScreenCoordsXY& screenCoords)
 {
-    screenCoords.x = ((screenCoords.x + 8) - MAXIMUM_MAP_SIZE_TECHNICAL) / 2;
-    screenCoords.y = ((screenCoords.y + 8)) / 2;
-    int32_t x = (screenCoords.y - screenCoords.x) * 32;
-    int32_t y = (screenCoords.x + screenCoords.y) * 32;
+    auto newScreenCoords = screenCoords;
+    newScreenCoords.x = ((screenCoords.x + 8) - MAXIMUM_MAP_SIZE_TECHNICAL) / 2;
+    newScreenCoords.y = ((screenCoords.y + 8)) / 2;
+    int32_t x = (newScreenCoords.y - newScreenCoords.x) * 32;
```
Gymnasiast:
```suggestion
    auto location = TileCoordsXY(newScreenCoords.y - newScreenCoords.x, newScreenCoords.x + newScreenCoords.y).ToCoordsXY();
```

src/openrct2-ui/windows/News.cpp
```diff
@@ -199,37 +199,39 @@ static void window_news_scrollgetsize(rct_window* w, int32_t scrollIndex, int32_
...
  */
-static void window_news_scrollmousedown(rct_window* w, int32_t scrollIndex, ScreenCoordsXY screenCoords)
+static void window_news_scrollmousedown(rct_window* w, int32_t scrollIndex, const ScreenCoordsXY& screenCoords)
 {
     int32_t itemHeight = window_news_get_item_height();
     int32_t i, buttonIndex;
 
     buttonIndex = 0;
+    auto mutableScreenCoords = screenCoords;
     for (i = 11; i < 61; i++)
```
Gymnasiast: While we're here, could you use a constant or expression instead of the hardcoded 61 here?

src/openrct2/interface/Viewport.cpp
```diff
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
```
Gymnasiast: *cough*

src/openrct2/interface/Viewport.cpp
```diff
@@ -1639,35 +1639,36 @@ InteractionInfo set_interaction_info_from_paint_session(paint_session* session,
...
     InteractionInfo info{};
+    auto dynamicScreenCoords = screenCoords;
     if (window != nullptr && window->viewport != nullptr)
     {
         rct_viewport* myviewport = window->viewport;
-        screenCoords.x -= (int32_t)myviewport->x;
-        screenCoords.y -= (int32_t)myviewport->y;
-        if (screenCoords.x >= 0 && screenCoords.x < (int32_t)myviewport->width && screenCoords.y >= 0
-            && screenCoords.y < (int32_t)myviewport->height)
+        dynamicScreenCoords.x -= (int32_t)myviewport->x;
```
Gymnasiast: This function seems especially pointless to use const, considering how heavily it modifies the coords.

janisozaur:
I was wondering the same as Aaron, perhaps https://github.com/OpenRCT2/OpenRCT2/pull/10700 could provide the answer

tupaschoal:
> Is there a noticeable (positive) performance impact to these changes? (Good work, either way.)

> I was wondering the same as Aaron, perhaps #10700 could provide the answer

I didn't notice any gain on the average fps of `bpb.sv6`. I just tried to keep it consistent that it always receives as const-ref after #10695 proved to be useful

src/openrct2-ui/windows/News.cpp
tupaschoal: Does it exist already? I'm unsure of what it means.

src/openrct2-ui/windows/EditorInventionsList.cpp
tupaschoal: Other than making interfaces consistent, I think there is not.

src/openrct2/interface/Viewport.cpp
tupaschoal: Weirdly, it seems to always happen in this file.

tupaschoal:
I've reverted all places that modified the parameter to receive a copy again, except:
- Cases where it modified, but didn't need to
- Implementations of `rct_window_event_list` functions, because most can be `const &`, so I think it's better overall

If we merge, instead of squashing, we can later revert this last commit and have them all as `const &`

PR #10847, opened 2020-02-29
Prefer const ref to send (Tile)CoordsXY(ZD) over

tupaschoal:


AaronVanGeffen:
Is there a noticeable (positive) performance impact to these changes?

src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -575,14 +578,14 @@ static void window_track_place_draw_mini_preview_track(
         const rct_preview_track* trackBlock = trackBlockArray[trackType];
         while (trackBlock->index != 255)
         {
-            auto rotatedAndOffsetTrackBlock = origin + CoordsXY{ trackBlock->x, trackBlock->y }.Rotate(rotation);
+            auto rotatedAndOffsetTrackBlock = mutableOrigin + CoordsXY{ trackBlock->x, trackBlock->y }.Rotate(rotation);
```
Gymnasiast: Like in the other PR: why create a copy instead of not `const`ing the signature?

tupaschoal:
> Is there a noticeable (positive) performance impact to these changes? (Good work, either way.)

I didn't notice any gain on the average fps of `bpb.sv6`. I just tried to keep it consistent that it always receives as const-ref after #10695 proved to be useful

tupaschoal:
I've reverted all places that modified the parameter to receive a copy again. 
If we merge, instead of squashing, we can later revert this last commit and have them all as `const &`

PR #10851, opened 2020-02-29
Finance charts horizontal lines

pRizz:
Fixes a pet peeve of mine where it is difficult to tell where along the graph the line is. This adds horizontal lines to make it easier to tell, for all the financial graphs.

Before:
Screen Shot 2020-02-29 at 3 20 23 PM

After:
Screen Shot 2020-02-29 at 3 19 47 PM

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -7,7 +7,6 @@
 - Feature: [#10305] Add two shortcuts for increasing and decreasing the scaling factor.
 - Feature: [#10189] Make Track Designs work in multiplayer.
 - Feature: [#10357] Added window for scenery scatter tool, allowing for area and density selection.
-- Feature: [#10858] Finance charts horizontal lines
 - Change: [#1164] Use available translations for shortcut key bindings.
 - Fix: [#2485] Hide Vertical Faces not applied to the edges of water.
 - Fix: [#5249] No collision detection when building ride entrance at heights > 85.5m.
@@ -32,6 +31,7 @@
 - Fix: [#10739] Mountain tool overlay for even-numbered selections.
 - Fix: [#10752] Mute button state not correctly set at startup.
 - Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
+- Improved: [#10858] Added horizontal grid lines to finance charts.
 - Removed: [#6898] LOADMM and LOADRCT1 title sequence commands (use LOADSC instead).
 
 0.2.4 (2019-10-28)
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -7,6 +7,7 @@
 - Feature: [#10305] Add two shortcuts for increasing and decreasing the scaling factor.
 - Feature: [#10189] Make Track Designs work in multiplayer.
 - Feature: [#10357] Added window for scenery scatter tool, allowing for area and density selection.
+- Feature: [#10858] Finance charts horizontal lines
 - Change: [#1164] Use available translations for shortcut key bindings.
 - Fix: [#2485] Hide Vertical Faces not applied to the edges of water.
 - Fix: [#5249] No collision detection when building ride entrance at heights > 85.5m.
--- a/src/openrct2-ui/windows/Finances.cpp
+++ b/src/openrct2-ui/windows/Finances.cpp
@@ -879,6 +879,7 @@ static void window_finances_financial_graph_paint(rct_window* w, rct_drawpixelin
     {
         money32 axisValue = axisBase << yAxisScale;
         gfx_draw_string_right(dpi, STR_FINANCES_FINANCIAL_GRAPH_CASH_VALUE, &axisValue, COLOUR_BLACK, x + 70, y);
+        gfx_draw_line(dpi, x + 70, y + 5, graphLeft + 482, y + 5, w->colours[0]);
         y += 39;
     }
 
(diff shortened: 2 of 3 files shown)


IntelOrca:
> make it grey instead of black?

+1

pRizz:
@tupaschoal Thank you for the quick feedback!

- I think grey is good too 👍 
- I had considered this too, but it seemed like a good deal more effort to implement, unless I'm missing a basic library call.
- I also considered this, but yeah, I think it's worthy to have always on.

AaronVanGeffen:
If we're going down this route, I would prefer other graphs in game to sport these ticks as well. I'm thinking about the park window, which shows the park rating and number of guests over time. (Come to think of it, they currently have no vertical axis as all. Not sure how easy it would be to implement them.)

AaronVanGeffen:
Grey (window colour 2) instead of black (window colour 0):

Screenshot_20200301_145828

pizza2004:
I think it needs to be moved to make sure the lines don't intersect with the horizontal ticks.  You could move those labels to the bottom where most graphs keep them possibly.  Also I wish we could use a color that would blend into the background better.  Is it possible to make it use the color that is used along the bottom/right edges of the part of the window that makes it look inlaid?  Something that changes with the background color of the window would look best IMO.

AaronVanGeffen:
Using the insert rectangle drawing functions, we can get the lines to blend to:

Screenshot_20200301_161632

src/openrct2-ui/windows/Finances.cpp
```diff
@@ -879,6 +879,7 @@ static void window_finances_financial_graph_paint(rct_window* w, rct_drawpixelin
     {
         money32 axisValue = axisBase << yAxisScale;
         gfx_draw_string_right(dpi, STR_FINANCES_FINANCIAL_GRAPH_CASH_VALUE, &axisValue, COLOUR_BLACK, x + 70, y);
+        gfx_draw_line(dpi, x + 70, y + 5, graphLeft + 482, y + 5, w->colours[0]);
```
AaronVanGeffen: To get to what I proposed in my last screenshot:

```suggestion
        gfx_fill_rect_inset(dpi, x + 70, y + 5, graphLeft + 482, y + 5, w->colours[2], INSET_RECT_FLAG_BORDER_INSET);
```


Screenshot_20200301_161632

pRizz:
@AaronVanGeffen Very nice. I like that style.

Also, I went ahead and implemented dashed lines for fun 😄here's how it looks. I'm trying to figure out how to color them gray/yellow but it seems like there is no gray in the `w->colours` array. How can I use other colors? I tried using one of the enums but it didn't work.

Screen Shot 2020-03-01 at 12 37 29 PM

AaronVanGeffen:
Not too keen on the dashes, to be honest, but I appreciate the effort.

As for making the lines grey, `w->colours[2]` should be fine — it's used for the window frame. I left a review suggestion here on GitHub earlier with the code I used to get the amber coloured lines, too. Personally, I prefer that.

James103:
Is it possible to apply a similar update when it comes to the other graphs (number of guests, park rating, etc), if not already implemented?

Gymnasiast:
@James103 Look, can we just make sure this first-time contributor gets their first PR right? It can be hard enough without all kinds of requests, especially considering you are not a team member.

pRizz:
Current images from this PR:

Screen Shot 2020-03-01 at 4 43 14 PM
Screen Shot 2020-03-01 at 4 43 20 PM
Screen Shot 2020-03-01 at 4 43 25 PM

scripts/check-code-formatting
```diff
@@ -2,7 +2,8 @@
 set -e
 
 # Ensure we are in root directory
-basedir="$(readlink -f `dirname $0`/..)"
+readlink_bin=$([[ "$OSTYPE" == "darwin"* ]] && printf greadlink || printf readlink)
```
AaronVanGeffen: `greadlink` is not present on macOS by default. Instead, it is supplied by e.g. Homebrew. I think that, instead, we should check whether `greadlink` exists and if not, use `readlink`. However, this is outside of the scope for this PR.

src/openrct2/drawing/NewDrawing.cpp
```diff
@@ -209,6 +211,27 @@ void gfx_draw_line(rct_drawpixelinfo* dpi, int32_t x1, int32_t y1, int32_t x2, i
     }
 }
 
+void gfx_draw_dashed_line(
```
AaronVanGeffen: This function is no longer used, so please drop these changes.

scripts/check-code-formatting
Gymnasiast: I don't think this change belongs in this PR at all. It has nothing to do with the feature/enhancement at hand.

distribution/changelog.txt
```diff
@@ -7,6 +7,7 @@
 - Feature: [#10305] Add two shortcuts for increasing and decreasing the scaling factor.
 - Feature: [#10189] Make Track Designs work in multiplayer.
 - Feature: [#10357] Added window for scenery scatter tool, allowing for area and density selection.
+- Feature: [#10858] Finance charts horizontal lines
```
Gymnasiast: This is not a feature. It should be listed under "Improved".

scripts/check-code-formatting
AaronVanGeffen: That's what I said..?

AaronVanGeffen:
I've taken the liberty of dropping the relevant commits. Once CI passes, I intend to squash-merge this PR.

scripts/check-code-formatting
Gymnasiast: I don't think you were being clear at all. What you said could easily be interpreted as "leave this for now and the greadlink/readlink stuff can be handled in another PR" (which is the way that I interpreted it).

PR #10854, opened 2020-03-01
Increase usage of ScreenCoordsXY on viewport functions and structs

tupaschoal:
I changed `rct_viewport` and `rct_window` to use `ScreenCoordsXY`, not sure if this is a problem, as they used to have comments on the size and paddings

### DIFF
--- a/src/openrct2-ui/interface/Window.cpp
+++ b/src/openrct2-ui/interface/Window.cpp
@@ -38,13 +38,13 @@ static bool window_fits_between_others(int32_t x, int32_t y, int32_t width, int3
         if (w->flags & WF_STICK_TO_BACK)
             continue;
 
-        if (x + width <= w->x)
+        if (x + width <= w->windowPos.x)
             continue;
-        if (x >= w->x + w->width)
+        if (x >= w->windowPos.x + w->width)
             continue;
-        if (y + height <= w->y)
+        if (y + height <= w->windowPos.y)
             continue;
-        if (y >= w->y + w->height)
+        if (y >= w->windowPos.y + w->height)
             continue;
         return false;
     }
@@ -145,8 +145,7 @@ rct_window* window_create(
     }
 
     w->number = 0;
-    w->x = screenCoords.x;
-    w->y = screenCoords.y;
+    w->windowPos = screenCoords;
     w->width = width;
     w->height = height;
     w->viewport = nullptr;
@@ -227,43 +226,43 @@ rct_window* window_create_auto_pos(
         if (w->flags & WF_STICK_TO_BACK)
             continue;
 
-        x = w->x + w->width + 2;
-        y = w->y;
+        x = w->windowPos.x + w->width + 2;
+        y = w->windowPos.y;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
 
-        x = w->x - w->width - 2;
-        y = w->y;
+        x = w->windowPos.x - w->width - 2;
+        y = w->windowPos.y;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
 
-        x = w->x;
-        y = w->y + w->height + 2;
+        x = w->windowPos.x;
+        y = w->windowPos.y + w->height + 2;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
 
-        x = w->x;
-        y = w->y - w->height - 2;
+        x = w->windowPos.x;
+        y = w->windowPos.y - w->height - 2;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
 
-        x = w->x + w->width + 2;
-        y = w->y - w->height - 2;
+        x = w->windowPos.x + w->width + 2;
+        y = w->windowPos.y - w->height - 2;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
 
-        x = w->x - w->width - 2;
-        y = w->y - w->height - 2;
+        x = w->windowPos.x - w->width - 2;
+        y = w->windowPos.y - w->height - 2;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
 
-        x = w->x + w->width + 2;
-        y = w->y + w->height + 2;
+        x = w->windowPos.x + w->width + 2;
+        y = w->windowPos.y + w->height + 2;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
 
-        x = w->x - w->width - 2;
-        y = w->y + w->height + 2;
+        x = w->windowPos.x - w->width - 2;
+        y = w->windowPos.y + w->height + 2;
         if (window_fits_within_space(x, y, width, height))
             goto foundSpace;
     }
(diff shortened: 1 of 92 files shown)


src/openrct2-ui/windows/Banner.cpp
```diff
@@ -133,10 +133,9 @@ rct_window* window_banner_open(rct_windownumber number)
     window_init_scroll_widgets(w);
 
     auto banner = GetBanner(w->number);
-    int32_t view_x = banner->position.x << 5;
-    int32_t view_y = banner->position.y << 5;
+    auto bannerViewPos = banner->position.ToCoordsXY();
```
Gymnasiast:
```suggestion
    auto bannerViewPos = banner->position.ToCoordsXY().ToTileStart();
```
and remove the += at line 154.

src/openrct2-ui/windows/Banner.cpp
```diff
@@ -347,15 +346,14 @@ static void window_banner_viewport_rotate(rct_window* w)
 
     auto banner = GetBanner(w->number);
 
-    int32_t view_x = (banner->position.x << 5) + 16;
-    int32_t view_y = (banner->position.y << 5) + 16;
-    int32_t view_z = w->frame_no;
+    auto bannerViewPos = CoordsXYZ{ banner->position.ToCoordsXY() + CoordsXY{ 16, 16 }, w->frame_no };
```
Gymnasiast:
```suggestion
    auto bannerViewPos = CoordsXYZ{ banner->position.ToCoordsXY().ToTileStart(), w->frame_no };
```

src/openrct2-ui/windows/Player.cpp
```diff
@@ -581,7 +580,7 @@ static void window_player_set_page(rct_window* w, int32_t page)
     {
         if (w->viewport == nullptr)
         {
-            viewport_create(w, w->x, w->y, w->width, w->height, 0, 128 * 32, 128 * 32, 0, 1, SPRITE_INDEX_NULL);
+            viewport_create(w, { w->x, w->y }, w->width, w->height, 0, { 128 * 32, 128 * 32, 0 }, 1, SPRITE_INDEX_NULL);
```
Gymnasiast:
```suggestion
            viewport_create(w, { w->x, w->y }, w->width, w->height, 0, TileCoordsXYZ(128, 128, 0).ToCoordsXYZ, 1, SPRITE_INDEX_NULL);
```

src/openrct2-ui/windows/Sign.cpp
```diff
@@ -156,10 +156,9 @@ rct_window* window_sign_open(rct_windownumber number)
     window_init_scroll_widgets(w);
 
     auto banner = GetBanner(w->number);
-    int32_t view_x = banner->position.x << 5;
-    int32_t view_y = banner->position.y << 5;
+    auto signViewPos = banner->position.ToCoordsXY();
```
Gymnasiast: Use ToTileStart()

src/openrct2-ui/windows/Sign.cpp
```diff
@@ -365,15 +364,14 @@ static void window_sign_viewport_rotate(rct_window* w)
 
     auto banner = GetBanner(w->number);
 
-    int32_t view_x = (banner->position.x << 5) + 16;
-    int32_t view_y = (banner->position.y << 5) + 16;
-    int32_t view_z = w->frame_no;
+    auto signViewPos = CoordsXYZ{ banner->position.ToCoordsXY() + CoordsXY{ 16, 16 }, w->frame_no };
```
Gymnasiast: ditto

src/openrct2-ui/windows/Sign.cpp
```diff
@@ -433,14 +430,14 @@ rct_window* window_sign_small_open(rct_windownumber number)
     w->var_492 = tile_element->AsWall()->GetSecondaryColour();
     w->var_48C = tile_element->AsWall()->GetEntryIndex();
 
-    view_x += 16;
-    view_y += 16;
+    signViewPos += CoordsXY{ 16, 16 };
```
Gymnasiast: ditto

src/openrct2-ui/windows/Viewport.cpp
```diff
@@ -94,13 +94,13 @@ rct_window* window_viewport_open()
     w->number = _viewportNumber++;
 
     // Create viewport
-    viewport_create(w, w->x, w->y, w->width, w->height, 0, 128 * 32, 128 * 32, 0, 1, SPRITE_INDEX_NULL);
+    viewport_create(w, { w->x, w->y }, w->width, w->height, 0, { 128 * 32, 128 * 32, 0 }, 1, SPRITE_INDEX_NULL);
```
Gymnasiast: See other remark about using proper conversions.

Gymnasiast:
I think the comments on the size and paddings are from the era where we were still calling RCT2 functions.

I think the PR is good, but I did find some bits you need to address.

src/openrct2-ui/windows/Banner.cpp
tupaschoal: Can't this theoretically return a different `TileElement*`? Because the sum happens after the `map_get_first_element_at`

src/openrct2-ui/windows/Banner.cpp
tupaschoal: Did you mean `ToTileCentre`?

I remembered about this function, but I was unsure if flooring would be alright.

src/openrct2-ui/windows/Banner.cpp
Gymnasiast: No. Both represent the same tile, and the structure holding them divides the x and y by 32 anyway.

src/openrct2-ui/interface/Window.cpp
```diff
@@ -133,8 +133,6 @@ rct_window* window_create(
 
     // Setup window
     w->classification = cls;
-    w->var_4B8 = -1;
```
Gymnasiast: Why was this removed?

src/openrct2-ui/windows/GameBottomToolbar.cpp
```diff
@@ -421,8 +421,8 @@ static void window_game_bottom_toolbar_draw_left_panel(rct_drawpixelinfo* dpi, r
     if (!(gParkFlags & PARK_FLAGS_NO_MONEY))
     {
         rct_widget widget = window_game_bottom_toolbar_widgets[WIDX_MONEY];
-        int32_t x = w->x + (widget.left + widget.right) / 2;
-        int32_t y = w->y + (widget.top + widget.bottom) / 2 - (line_height == 10 ? 5 : 6);
+        int32_t x = w->windowPos.x + (widget.left + widget.right) / 2;
+        int32_t y = w->windowPos.y + (widget.top + widget.bottom) / 2 - (line_height == 10 ? 5 : 6);
```
Gymnasiast: Note to self or @AaronVanGeffen : shouldn't we just divide `line_height` by 2 here?

src/openrct2-ui/interface/Window.cpp
tupaschoal: No one was reading from it.

src/openrct2-ui/windows/GameBottomToolbar.cpp
AaronVanGeffen: No, some fonts use a line height of 14 or 15 while the glyphs are offset to compensate. It's a little messy, but the main reason why I hardcoded this.

src/openrct2/interface/Viewport.cpp
```diff
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
```
Gymnasiast: :bomb:

PR #10861, opened 2020-03-03
Fix #10662: Fixed font issue on create/remove ducks tooltip

### LINKED ISSUE #10662: Duck-cheat tooltips has incorrect font style
**Version:** 0.2.4
**Commit/Build:** Every build since 7fa41bb82d26cf483964cb88d0038cfa42d08645

The tooltip that appears when hovering over the "Create Ducks" and the "Remove Ducks" buttons in the cheat window, uses a font style that does not match the other tooltips.
The tooltips themselves don't really explain what these buttons do (creating ducks could explain that they only spawn on the edge of the map and only if there is water for example). This could be improved too.

In `data/language/en-GB.txt` you'll have to add a new string for the tooltip. Tooltips are prefixed by `{SMALLFONT}{BLACK}`. You don't need to edit the other language files, that will get done in the [localisation repository ](https://github.com/OpenRCT2/Localisation)after this fix gets merged.

You'll also need to update the widget definition for the buttons in `src/openrct2-ui/windows/Cheats.cpp`. The tooltips are the last parameter in the initialization list.

daandeheij:
- Correct tooltip font shows now
- Extended the tooltip string to be more informative


thversfelt:
**Before**
Wrong tooltip font and not an informative tooltip:

image

**After**
The `create ducks` tooltip:

image

The `remove ducks` tooltip:

image

thversfelt:
I added an issue for the added strings of this PR in the localization repository [here](https://github.com/OpenRCT2/Localisation/issues/1809).

IntelOrca:
Thank you for your contribution, if you like - you can also add your names to:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/contributors.md

thversfelt:
> 
> 
> Thank you for your contribution, if you like - you can also add your names to:
> https://github.com/OpenRCT2/OpenRCT2/blob/develop/contributors.md
> 
> This fixes a bug present in the last release so you will want to add an entry to the changelog as well:
> https://github.com/OpenRCT2/OpenRCT2/blob/develop/distribution/changelog.txt

Thanks! We will add our entry to the changelog as soon as possible. Also, we will continue working on other issues and try to tackle harder ones.

data/language/en-GB.txt
```diff
@@ -3674,6 +3674,8 @@ STR_6352    :Density
 STR_6353    :{SMALLFONT}{BLACK}Low density
 STR_6354    :{SMALLFONT}{BLACK}Medium density
 STR_6355    :{SMALLFONT}{BLACK}High density
+STR_6356    :{SMALLFONT}{BLACK}Spawns ducks if park contains water
+STR_6357    :{SMALLFONT}{BLACK}Removes ducks if park contains ducks
```
Gymnasiast: This seems a bit redundant to me. I would change it to this:

```suggestion
STR_6357    :{SMALLFONT}{BLACK}Removes all ducks from the map
```

thversfelt: Indeed, that seems like a better tooltip.

Issue #10637, opened 2020-01-30
Remove floating objects


Richard-L:
Was certain I'd posted this before but could've been on the discord:
![image](https://user-images.githubusercontent.com/4934209/73488520-60c0bf00-43a9-11ea-9747-a62673419238.png)

Would it be possible to either hide all floating objects (balloons, ducks, waving money, ...?) like you would hide scenery/rides/supports, or to add a console command "remove_floating_objects" to do away with the current air traffic?

This would be very helpful for us content creators when using "Giant Screenshot".

Thanks for considering.

ocalhoun6:
There already is a setting for turning on or off the waving money ... but the other parts of this could be useful.

Gymnasiast:
@ocalhoun6 Not true. The only thing you can toggle is whether guest's purchases also trigger waving money. Building will _always_ create a money effect.

Richard-L:
I think building isn't even that important for now. It's more something to go in conjunction with giant screenshot.

It'd be amazing if a potential "remove_floating_objects" console command would also have effect while the game is paused btw. Sometimes people time certain things to the moment when you open their park (ie explosions, or a coaster having just finished a lifthill to great you). Therefore we always take giant screenshots with the game paused.

Gymnasiast:
Has now been added.

Issue #7417, opened 2018-04-15
Console Commands do not Sync Across Multiplayer


saxman1089:
This probably isn't a bug, but wanted to bring it up here anyway. It appears that when using console commands these commands do not sync across multiplayer. For more aesthetically-oriented players like myself who use frozen staff as scenery or need to correct certain ratings for a coaster, and who play on MP, it makes it difficult to make some of these changes, as you need to pull the save off of MP and do it in single player, then place the save back onto MP. Is there any reason for not having these commands sync across MP?

janisozaur:
It's much easier to have those commands modify the game state directly rather than doing it through the game command/game action (which is how most part of the engine works and enables multiplayer).

I think for some things, e.g. entertainer costumes, we also skip some checks that would otherwise take place, not sure about other ones. This is still solvable with some amount of work, PRs welcome.

Margen67:
#2987

PR #10863, opened 2020-03-03
Feature: Add console command for removing all floating objects

jeysbach:
Feature requested in issue #10637

Console command "remove_floating_objects" removes all balloon sprites, money effects and ducks with state FLY_AWAY or FLY_TO_WATER. It returns how many objects were removed.

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -815,6 +815,38 @@ void litter_remove_at(int32_t x, int32_t y, int32_t z)
 }
 
 /**
+ * Loops through all sprites, finds floating objects and removes them.
+ * Returns the amount of removed objects as feedback.
+ */
+uint16_t remove_floating_sprites()
+{
+    uint16_t removed = 0;
+    for (uint16_t i = 0; i < MAX_SPRITES; i++)
+    {
+        rct_sprite* rctSprite = get_sprite(i);
+        if (rctSprite->IsBalloon())
+        {
+            sprite_remove(rctSprite->AsBalloon());
+            removed++;
+        }
+        if (rctSprite->IsDuck())
+        {
+            if (rctSprite->AsDuck()->IsFlying())
+            {
+                rctSprite->duck.Remove();
+                removed++;
+            }
+        }
+        if (rctSprite->IsMoneyEffect())
+        {
+            sprite_remove(rctSprite->AsMoneyEffect());
+            removed++;
+        }
+    }
+    return removed;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


AaronVanGeffen:
This looks like it will not work in a networked game. The question is whether that is required at this point — I believe most of the console commands aren't networked anyway.

jeysbach:
Having the console commands sync across network play has its own issue #7417 so I suppose it could be considered out of scope for this PR if it has not been implemented for other console commands already.

src/openrct2/world/Sprite.cpp
```diff
@@ -814,6 +814,41 @@ void litter_remove_at(int32_t x, int32_t y, int32_t z)
...
+{
+    uint16_t removed = 0;
+    for (uint16_t i = 0; i < MAX_SPRITES; i++)
+    {
+        rct_sprite* rctSprite = get_sprite(i);
+        if (rctSprite->IsBalloon())
+        {
+            sprite_remove(rctSprite->AsBalloon());
+            sprite_misc_update(rctSprite);
+            removed++;
```
AaronVanGeffen: Either `continue` here, or …

src/openrct2/world/Sprite.cpp
```diff
@@ -814,6 +814,41 @@ void litter_remove_at(int32_t x, int32_t y, int32_t z)
...
+        }
+        if (rctSprite->IsDuck())
```
AaronVanGeffen:
```suggestion
        else if (rctSprite->IsDuck())
```

src/openrct2/world/Sprite.cpp
```diff
@@ -814,6 +814,41 @@ void litter_remove_at(int32_t x, int32_t y, int32_t z)
...
+        {
+            if (rctSprite->AsDuck()->IsFlying())
+            {
+                rctSprite->duck.Remove();
+                sprite_misc_update(rctSprite);
+                removed++;
+            }
+        }
+        if (rctSprite->IsMoneyEffect())
```
AaronVanGeffen:
```suggestion
        else if (rctSprite->IsMoneyEffect())
```

duncanspumpkin:
Shouldn't this be moved to plugin. It is surely something that lends itself to the plugin system and i think we are very very close to having the plugin system out.

IntelOrca:
> Shouldn't this be moved to plugin. It is surely something that lends itself to the plugin system and i think we are very very close to having the plugin system out.

Let's merge this one, as soon as plugin system is "merged" we can remove all this stuff.

Richard-L:
@jeysbach thank you so much !!!

PR #10864, opened 2020-03-04
Change station starts to CoordsXY

Gymnasiast:


src/openrct2/ride/RideRatings.cpp
```diff
@@ -1460,9 +1454,9 @@ static int32_t ride_ratings_get_scenery_score(Ride* ride)
 
     // Count surrounding scenery items
     int32_t numSceneryItems = 0;
-    for (int32_t yy = std::max(y - 5, 0); yy <= std::min(y + 5, 255); yy++)
+    for (int32_t yy = std::max(location.y - 5, 0); yy <= std::min(location.y + 5, 255); yy++)
```
Gymnasiast: Not correct, needs TileCoordsXY. Also replace 255 with a constant.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9258,7 +9258,7 @@ static int32_t vehicle_update_track_motion_mini_golf(Vehicle* vehicle, int32_t*
 
     for (int32_t i = 0; i < MAX_STATIONS; i++)
     {
-        auto trackLoc = TileCoordsXY(vehicle->TrackLocation);
+        auto trackLoc = vehicle->TrackLocation;
         if (trackLoc != ride->stations[i].Start)
```
Gymnasiast: Can be compared directly.

src/openrct2/ride/Ride.cpp
```diff
@@ -4902,9 +4898,9 @@ static bool ride_initialise_cable_lift_track(Ride* ride, bool isApplying)
         }
     }
 
-    int32_t x = location.x * 32;
-    int32_t y = location.y * 32;
-    int32_t z = ride->stations[stationIndex].GetBaseZ();
+    int32_t x = location.x;
+    int32_t y = location.y;
+    int32_t z = location.z;
```
tupaschoal: Sounds like these could benefit from becoming a `CoordsXYZ`

PR #10866, opened 2020-03-04
Refactor/more tile element prep

Gymnasiast:
This does most (all?) of the remaining tile element refactors. Importantly, it makes room for some extra flags in the tile element - like in invisibility flag.


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1988,7 +1988,12 @@ class S4Importer final : public IParkImporter
...
         dst->ClearAs(tileElementType);
         dst->SetDirection(src->GetDirection());
-        dst->flags = src->flags;
+
+        // All saved in "flags"
+        dst->SetOccupiedQuadrants(src->GetOccupiedQuadrants());
+        dst->SetGhost(src->IsGhost());
+        dst->SetLastForTile(src->IsLastForTile());
+
         dst->SetBaseZ(src->base_height * 4);
```
Gymnasiast: RCT1_COORDS_Z_STEP

src/openrct2/ride/Track.h
```diff
@@ -66,6 +66,8 @@ enum
     TRACK_ELEMENT_FLAGS2_CABLE_LIFT = 1 << 2,
     TRACK_ELEMENT_FLAGS2_HIGHLIGHT = 1 << 3,
     TRACK_ELEMENT_FLAGS2_HAS_GREEN_LIGHT = 1 << 4,
+    TRACK_ELEMENT_FLAG2_BLOCK_BRAKE_CLOSED = 1 << 5,
```
Gymnasiast: TRACK_ELEMENT_FLAGS2_BLOCK_BRAKE_CLOSED

src/openrct2/world/LargeScenery.h
```diff
@@ -14,3 +14,8 @@
 #include "TileElement.h"
 
 rct_scenery_entry* get_large_scenery_entry(int32_t entryIndex);
+
+enum
+{
+    LARGE_SCENERY_ELEMENT_FLAGS2_ACCOUNTED = 1 << 0,
```
tupaschoal: Why an enum?

Gymnasiast: More flags might follow. We use enums for all flag fields.

IntelOrca: > We use enums for all flag fields.

Not true. I have been using namespaces for ages because it should most of the time force you to write the prefix (`LARGE_SCENERY_ELEMENT_FLAGS2::`) and still allows you to do bitwise ops if you use constants rather than enum class.

Gymnasiast: Well, the tile element stuff still seems to use it pretty consistently, at any rate.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1832,7 +1832,7 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
             {
                 // Details
                 // Path name
-                rct_string_id pathNameId = tileElement->AsPath()->GetPathEntry()->string_idx;
+                rct_string_id pathNameId = tileElement->AsPath()->GetSurfaceEntry()->string_idx;
```
Broxzier: This doesn't look right. This is about getting the string ID for the path's name. Assuming it still gives back the right string, why was this renamed to `GetSurfaceEntry`?

Gymnasiast: There are path surfaces and path railings. Calling the former just "path" is confusing in its own right.

PR #10881, opened 2020-03-05
Translate SV6 object offsets to OpenRCT2 ones

Gymnasiast:
Previously, raising one of our own object limits (like 252 small scenery objects) resulted in a crash. This PR translates the object offsets between RCT2 and OpenRCT2, putting it "null objects" as necessary.

### DIFF
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -35,6 +35,7 @@
 #include "../peep/Staff.h"
 #include "../rct12/SawyerChunkReader.h"
 #include "../rct12/SawyerEncoding.h"
+#include "../rct2/RCT2.h"
 #include "../ride/Ride.h"
 #include "../ride/RideRatings.h"
 #include "../ride/ShopItem.h"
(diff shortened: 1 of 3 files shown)


src/openrct2/rct2/S6Importer.cpp
```diff
@@ -170,7 +171,9 @@ class S6Importer final : public IParkImporter
 
         _s6Path = path;
 
-        return ParkLoadResult(std::vector<rct_object_entry>(std::begin(_s6.objects), std::end(_s6.objects)));
+        std::vector<rct_object_entry> objects;
```
duncanspumpkin: This isn't used. Assume it is left over.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1620,6 +1623,28 @@ class S6Importer final : public IParkImporter
...
     }
+
+    std::vector<rct_object_entry> GetRequiredObjects()
+    {
+        std::vector<rct_object_entry> result;
+        rct_object_entry nullEntry = {};
+        std::memset(&nullEntry, 0xFF, sizeof(nullEntry));
+
+        int objectIt = 0;
+        for (int16_t objectType = OBJECT_TYPE_RIDE; objectType <= OBJECT_TYPE_WATER; objectType++)
```
duncanspumpkin: I take it Scenerio text isn't included for a reason.

Gymnasiast: Yes. Ted and I removed support for that type of object long ago.

PR #10884, opened 2020-03-07
Axis labels and horizontal guides for park window

AaronVanGeffen:
This PR adds Y axis labels and horizontal guides to the park window, more specifically to the park rating and guests tabs.

Alignment isn't 100% perfect yet, in particularly for the guests tab. This needs a bit more tinkering.

rating_800

guests_2000

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3673,7 +3673,7 @@ STR_6356    :{SMALLFONT}{BLACK}Spawns ducks if park contains water
 STR_6357    :{SMALLFONT}{BLACK}Removes all ducks from the map
 STR_6358    :Page {UINT16}
 STR_6359    :{POP16}{POP16}Page {UINT16}
-STR_6360    :{SMALLFONT}{BLACK}{UINT16}
+STR_6360    :{SMALLFONT}{BLACK}{COMMA32}
 
 #############
 # Scenarios #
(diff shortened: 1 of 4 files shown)


AaronVanGeffen:
Coming back to this, I think the chart alignment is actually not incorrect. What I had not taken into account is that the current rating or number of guests (as mentioned above the charts) does not always match with the last data point in the chart. If I allow enough time to pass in game, the charts eventually match expectations.

Could someone else verify this is working as intended?

data/language/en-GB.txt
```diff
@@ -3673,6 +3673,7 @@ STR_6356    :{SMALLFONT}{BLACK}Spawns ducks if park contains water
 STR_6357    :{SMALLFONT}{BLACK}Removes all ducks from the map
 STR_6358    :Page {UINT16}
 STR_6359    :{POP16}{POP16}Page {UINT16}
+STR_6360    :{SMALLFONT}{BLACK}{UINT16}
```
tupaschoal: What's this being used for?

AaronVanGeffen: This string is used to display the y-axis labels.

Come to think of it, perhaps I should have used `{COMMA16}` instead, so we'd also get separators; e.g. `1,000` instead of just `1000`.

Gymnasiast: For the guest numbers, you will need to use COMMA32 - the new save format will allow more than the 32767 guests that a COMMA16 can represent.

AaronVanGeffen: I don't mind the change, but at that point, the scales will require changing, anyway…

IntelOrca: At least in that case, it can't overflow...

Gymnasiast: Yes, but since we know this is coming up and we're touching this code anyway, I would argue we should future-proof it now. The new save format PR is big and far-reaching enough as it is.

PR #10886, opened 2020-03-07
Refactor/more big coords

Gymnasiast:


src/openrct2/actions/RideEntranceExitPlaceAction.hpp
```diff
@@ -182,7 +182,7 @@ DEFINE_GAME_ACTION(RideEntranceExitPlaceAction, GAME_COMMAND_PLACE_RIDE_ENTRANCE
         res->Position = { _loc.ToTileCentre(), z };
         res->Expenditure = ExpenditureType::RideConstruction;
 
-        TileElement* tileElement = tile_element_insert(TileCoordsXYZ(CoordsXYZ{ _loc, z }), 0b1111);
+        TileElement* tileElement = tile_element_insert((CoordsXYZ{ _loc, z }), 0b1111);
```
Gymnasiast: redundant parentheses

src/openrct2/actions/TrackPlaceAction.hpp
```diff
@@ -348,7 +348,7 @@ DEFINE_GAME_ACTION(TrackPlaceAction, GAME_COMMAND_PLACE_TRACK, TrackPlaceActionR
             }
             if ((entranceDirections & TRACK_SEQUENCE_FLAG_ORIGIN) && trackBlock->index == 0)
             {
-                if (!track_add_station_element(mapLoc.x, mapLoc.y, baseZ / COORDS_Z_STEP, _origin.direction, _rideIndex, 0))
+                if (!track_add_station_element({ mapLoc, baseZ / COORDS_Z_STEP, _origin.direction }, _rideIndex, 0))
```
Gymnasiast:
```suggestion
                if (!track_add_station_element({ mapLoc, baseZ, _origin.direction }, _rideIndex, 0))
```

src/openrct2/actions/WallPlaceAction.hpp
```diff
@@ -392,11 +392,11 @@ DEFINE_GAME_ACTION(WallPlaceAction, GAME_COMMAND_PLACE_WALL, WallPlaceActionResu
         {
             return MakeResult(GA_ERROR::NO_FREE_ELEMENTS, STR_TILE_ELEMENT_LIMIT_REACHED);
         }
-
-        TileElement* tileElement = tile_element_insert({ _loc.x / 32, _loc.y / 32, targetHeight / 8 }, 0b0000);
+
```
Gymnasiast:
```suggestion

```

src/openrct2/peep/Staff.cpp
```diff
@@ -604,49 +603,49 @@ static int32_t staff_handyman_direction_rand_surface(Peep* peep, uint8_t validDi
     // If it tries all directions this is required
     // to make it back to the first direction and
     // override validDirections
-    direction &= 3;
-    return direction;
+    newDirection %= NumOrthogonalDirections;
```
tupaschoal: Why the change from & to %?

src/openrct2/ride/Track.cpp
```diff
@@ -619,12 +619,12 @@ static TileElement* find_station_element(int32_t x, int32_t y, int32_t z, int32_
     return nullptr;
 }
 
-static void ride_remove_station(Ride* ride, int32_t x, int32_t y, int32_t z)
+static void ride_remove_station(Ride* ride, CoordsXYZ location)
```
tupaschoal: `const &`

src/openrct2/peep/Staff.cpp
Gymnasiast: Because that's what the `& 3` does. But `% NumOrthogonalDirections` is much clearer, since it shows that we're truncating the direction to a valid one.

IntelOrca: Bare in mind that unless `newDirection` is unsigned, it will not be as efficient / generate the same instructions as `&`.

Gymnasiast: Good point. In this case, though, it uses an unsigned 8-bit integer.

src/openrct2/ride/Ride.cpp
```diff
@@ -2755,7 +2755,7 @@ Peep* find_closest_mechanic(int32_t x, int32_t y, int32_t forInspection)
         }
 
         if (map_is_location_in_park({ x, y }))
-            if (!staff_is_location_in_patrol(peep, x & 0xFFE0, y & 0xFFE0))
+            if (!peep->AsStaff()->IsLocationInPatrol({ x & 0xFFE0, y & 0xFFE0 }))
```
tupaschoal: Isn't this `ToTileStart()`?

PR #10887, opened 2020-03-07
Fix #10691 Improve Tile Inspector Delete 

### LINKED ISSUE #10691: Improve Tile Inspector Delete
When deleting tiles with the tile inspector it shouldn't be possible to delete all the tiles at one coordinate. There should always be at least one (it could also be argued there should always be a surface element as many functions assume that).

kylegarza21:
**When a user attempts to remove the last element on a tile, they are warned of the illegal action**
prevent_removal_of_last_element

**A user is still able to remove elements to allow for a 'floating' effect**
floating_element_allowed


src/openrct2/world/TileInspector.cpp
```diff
@@ -151,12 +151,14 @@ GameActionResult::Ptr tile_inspector_remove_element_at(const CoordsXY& loc, int1
 {
     if (isExecuting)
     {
-        // Forcefully remove the element
-        TileElement* const tileElement = map_get_nth_element_at(loc, elementIndex);
-        if (tileElement == nullptr)
+        // At least one surface element must exist on a tile
+        if (1 >= windowTileInspectorElementCount)
```
AaronVanGeffen: We don't use [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) in this project.

```suggestion
        if (windowTileInspectorElementCount < 1)
```

src/openrct2/localisation/StringIds.h
```diff
@@ -3905,9 +3905,10 @@ enum
 
     STR_CREATE_DUCKS_TIP = 6356,
     STR_REMOVE_DUCKS_TIP = 6357,
+    STR_TILE_NEEDS_ATLEAST_ONE_ELEMENT = 6358,
 
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    STR_COUNT = 32768
+    STR_COUNT = 32769
```
AaronVanGeffen: This constant should not be changed.
```suggestion
    STR_COUNT = 32768
```

AaronVanGeffen:
Looks good to me, but I'd like another team member to review it, preferably @duncanspumpkin or @Broxzier

data/language/en-GB.txt
```diff
@@ -3676,6 +3676,7 @@ STR_6354    :{SMALLFONT}{BLACK}Medium density
 STR_6355    :{SMALLFONT}{BLACK}High density
 STR_6356    :{SMALLFONT}{BLACK}Spawns ducks if park contains water
 STR_6357    :{SMALLFONT}{BLACK}Removes all ducks from the map
+STR_6358    :A tile must have at least one surface element
```
Broxzier:
```suggestion
STR_6358    :A tile must have at least one tile element
```

src/openrct2/world/TileInspector.cpp
```diff
@@ -151,12 +151,15 @@ GameActionResult::Ptr tile_inspector_remove_element_at(const CoordsXY& loc, int1
 {
     if (isExecuting)
     {
-        // Forcefully remove the element
-        TileElement* const tileElement = map_get_nth_element_at(loc, elementIndex);
-        if (tileElement == nullptr)
+        // At least one surface element must exist on a tile
```
Broxzier:
```suggestion
        // At least one tile element must exist on a tile
```

data/language/en-GB.txt
```diff
@@ -3676,6 +3676,7 @@ STR_6354    :{SMALLFONT}{BLACK}Medium density
 STR_6355    :{SMALLFONT}{BLACK}High density
 STR_6356    :{SMALLFONT}{BLACK}Spawns ducks if park contains water
 STR_6357    :{SMALLFONT}{BLACK}Removes all ducks from the map
+STR_6358    :A tile must have at least one tile element
```
AaronVanGeffen: I'm not too keen on this new phrasing, as it now repeats the word 'tile'. What would be clearer? Not sure, myself.

src/openrct2/localisation/StringIds.h
```diff
@@ -3905,6 +3905,7 @@ enum
 
     STR_CREATE_DUCKS_TIP = 6356,
     STR_REMOVE_DUCKS_TIP = 6357,
+    STR_TILE_NEEDS_ATLEAST_ONE_ELEMENT = 6358,
```
Broxzier: A small nitpicky thing, since this string is unrelated to the two above please add a white line this one.

data/language/en-GB.txt
Broxzier: I think it's fine, "Tile" and "Tile element" are two separate things. In Dutch the latter would be one word.

tupaschoal: "A tile must have at least one element" looks fine to me, but I'm not too concerned with the repetition

distribution/changelog.txt
```diff
@@ -27,6 +27,7 @@
 - Fix: [#10547] RCT1 parks have too many rides available.
 - Fix: [#10587] Update last action coordinates on correct player.
 - Fix: [#10662] Duck cheat tooltips look odd and do not explain anything.
+- Fix: [#10691] Improved tile inspector delete functionality.
```
AaronVanGeffen: I think this should be more specific. Perhaps something like:
```suggestion
- Fix: [#10691] Improved error reporting for deleting elements in the tile inspector.
```

Broxzier: This would fit better below with the other 'Improved' items, and more descriptive. Something like:
`- Improved: [#10691] Tile elements can now only be removed when there are other tile elements left.`

Broxzier:
Ideally, the remove-button would be greyed out when it cannot be used, but that can be done later with another PR.

src/openrct2/world/TileInspector.cpp
```diff
@@ -151,12 +151,15 @@ GameActionResult::Ptr tile_inspector_remove_element_at(const CoordsXY& loc, int1
 {
     if (isExecuting)
     {
-        // Forcefully remove the element
-        TileElement* const tileElement = map_get_nth_element_at(loc, elementIndex);
-        if (tileElement == nullptr)
+        // At least one tile element must exist on a tile
+        if (windowTileInspectorElementCount <= 1)
```
duncanspumpkin: Although this does work this is not network safe. This variable is not passed over the network. You will need to recalculate the number of elements in this function to make it safe. Also this check only needs to be performed when in the non isExecuting branch.

kylegarza21: @Broxzier thanks for the lead. I plan on working this, this weekend. Will let you know if there's any questions

duncanspumpkin:
For some reason there isn't a link to this in #10691. This comment is just to show the link.

tupaschoal:
@kylegarza21 I'm closing this due to inactivity. If you can resume the work to address the notes, feel free to reopen the PR

PR #10889, opened 2020-03-08
Fix #10822: Can place too many peep spawns

### LINKED ISSUE #10822: Can place more than two peep spawns
**OS:** Windows 10 (x64)
**Version:** 0.2.4-587-g86a79e742
**Commit/Build:** 86a79e7

I can place more than two peep spawns, they are all functional. Saving and reloading only keeps the first two as expected. When you place more than 8, the first one will disappear.

**Steps to reproduce:**
1. Place two peep spawns (in sandbox mode or scenario editor)
2. Place some more

Gymnasiast:


mrtnptrs:
Wouldn't it be possible with the WIP-new save format to add support for more than two peep spawn points? So that you could like make a park with 4 entrances and 4 peep spawn points? Or is two peep spawn points more than enough?

Gymnasiast:
@mrtnptrs The new save format will raise the limit again.

PR #10901, opened 2020-03-09
Prepare guest window for more than 5 pages

Gymnasiast:
Fixed the problem mentioned in https://github.com/OpenRCT2/OpenRCT2/pull/10664#issuecomment-596251700.

Of course, if somebody were to reach 65536 pages of guests, this wouldn't work, but at that point they are probably busy running away from their molten computer anyway.


IntelOrca:
We should probably start replacing all the 16-bit format specifiers with 32-bit ones.

Gymnasiast:
@IntelOrca Right, but that's a PR on its own, I would say. 
Is this one fine as-is?

PR #10905, opened 2020-03-09
Fix #10904: RCT1/LL-scenarios with red water won't open

### LINKED ISSUE #10904: RCT1/LL-scenarios with red water won't open
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 8737d3c

RCT1/LL scenarios with red water turned on (Crazy Craters, Icarus Park, Venus Ponds) won't load when clicked at in the "new scenario" window. There is just a delay before the main menu resets entirely. No error message will be prompted.

This bug only affects RCT1/LL-scenarios. RCT2 scenarios with red or green water are fine, no issues for these ones.

Gymnasiast:


Gymnasiast:
For some reason, `size()` returns 1, and thus the entry gets added at index 1. Which is not possible, since you can only have one water entry.

Perhaps it's a broader issue with `GetOrAddEntry()` or the vector it uses, but I don't know enough about C++ and vectors to tell.

IntelOrca:
I will have a look at this. I can't see an obvious place where two water entries would end up getting added so I will have to debug through.

IntelOrca:
Problem 1: game version was not assessed before first call to `CreateAvailableObjectMappings`
Problem 2: `CreateAvailableObjectMappings` was being called twice.
Problem 3: `InitialiseEntryMaps` was being called twice, but it doesn't clear the entry lists.

I have just removed the second call. Import really means, take what we have already loaded and set the game state. Load actually decides what objects we want to use probably so it knows in advance if any are missing before trying to import.

PR #10906, opened 2020-03-09
Refactor rct_track_preview flags

Gymnasiast:


src/openrct2/ride/Track.h
```diff
@@ -52,6 +52,13 @@ struct rct_track_coordinates
     int16_t y;             // 0x08
 };
 
+enum
+{
+    RCT_PREVIEW_TRACK_FLAG_1 = (1 << 0),
```
duncanspumpkin: Is this coincident through the code base i thought we normally started at 0

Gymnasiast: Wasn't sure myself. I'll change it to 0.

Gymnasiast:
> Looks okay. Any ideas on what the flags might mean?

I'm relatively certain about the `(1 << 2)` flag, which is set on vertical pieces, so I did name that one. `(1 << 0)` does not appear to be read at all and while `(1 << 1)` is read, I am not sure of its purpose.

duncanspumpkin:
1<<1 looks to be set on all base sections like flat rides or tower bases. Not that that helps name it.

Gymnasiast:
Making matters more complicated is that it is typically only set on one of the sequences - e.g. not set on index 0-7, but set on 8.

Gymnasiast:
I have addressed the line note. Is this one good to go?

PR #10911, opened 2020-03-11
Deduplicate flat/regular ride track blocks

Gymnasiast:
Apart from track pieces 95, 110, 111, 115, 116, 118, 119, 121, 122, 123 and 255, the regular and ride track block definitions were the same. This PR deduplicates the track block definitions, which takes out 1800 LOC.


duncanspumpkin:
Good idea.

PR #10913, opened 2020-03-11
Fix #10602. Fix track scenery additions when far from station start

### LINKED ISSUE #10602: "Ride is too large/too much elements" warning on coasters with scenery
**OS:** Windows 7
**Version:** 0.2.4-421
**Commit/Build:** 8a9f0ee

butwhycantisaveit

This track design with scenery brings up the "Ride is too large, contains too many elements, or scenery is too spread out" error message when one tries to save it. This seems like it wouldn't be the case, as all the chosen scenery is on/right next to the track, and it's normal-sized for a Corkscrew coaster. Perhaps there's something else going on that I don't know.

I've attached the .td6 file containing the track and all the scenery it was saved with.
[The Boss(Louis).zip](https://github.com/OpenRCT2/OpenRCT2/files/4095197/The.Boss.Louis.zip)

duncanspumpkin:
Fix #10602. Fix track scenery additions when far from station start
Issue caused due to stuffing a temporary unsigned variable in a signed variable. This fix is not a permenant fix and this field in the future should be split up into two seperate fields: tile position and relative tile position.

### DIFF
--- a/src/openrct2/ride/TrackDesign.cpp
+++ b/src/openrct2/ride/TrackDesign.cpp
@@ -529,8 +529,8 @@ rct_string_id TrackDesign::CreateTrackDesignScenery()
         }
 
         // Cast the value into a uint8_t as this value is not signed yet.
-        CoordsXY sceneryMapPos{ ((uint8_t)scenery.x) * 32 - gTrackPreviewOrigin.x,
-                                ((uint8_t)scenery.y) * 32 - gTrackPreviewOrigin.y };
+        auto sceneryPos = TileCoordsXY(static_cast<uint8_t>(scenery.x), static_cast<uint8_t>(scenery.y)).ToCoordsXY();
+        CoordsXY sceneryMapPos = sceneryPos - gTrackPreviewOrigin;
         CoordsXY rotatedSceneryMapPos = sceneryMapPos.Rotate(0 - _saveDirection);
         TileCoordsXY sceneryTilePos{ rotatedSceneryMapPos };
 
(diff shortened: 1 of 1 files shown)


src/openrct2/ride/TrackDesign.cpp
```diff
@@ -528,7 +528,9 @@ rct_string_id TrackDesign::CreateTrackDesignScenery()
             }
         }
 
-        CoordsXY sceneryMapPos{ scenery.x * 32 - gTrackPreviewOrigin.x, scenery.y * 32 - gTrackPreviewOrigin.y };
+        // Cast the value into a uint8_t as this value is not signed yet.
+        CoordsXY sceneryMapPos{ ((uint8_t)scenery.x) * 32 - gTrackPreviewOrigin.x,
```
tupaschoal: Use `static_cast`

Gymnasiast:
Is there any reason why we can't turn the x, y and z coordinates of `TrackDesignSceneryElement` into `uint8_t`s?

duncanspumpkin:
Yes there is. What we should really do is split it up into two sets of fields. The relative fields and the absolute fields. The TD6 file only saves the relative fields (signed) but whilst adding/removing scenery the field is absolute (unsigned). I just think it might be slightly confusing why there are two sets of fields and only one are saved.

src/openrct2/ride/TrackDesign.cpp
Gymnasiast: Also use COORDS_XY_STEP instead of 32. (Bit of camp floor cleaning ;))

PR #10922, opened 2020-03-13
Cleanup scenery

Gymnasiast:
~~This contains a set of only tangentially related commits. It is therefore recommended to view them one by one.~~

This changes the way that scenery is managed. This is very important for allowing more than 256 scenery items of any type.


Gymnasiast:
I have no idea why some tests segfault. Reverting the changes to the tests doesn't seem to make a difference.

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -191,25 +191,25 @@ static rct_widget window_scenery_widgets[] = {
...
 // rct2: 0x00F64F2C
-static uint16_t window_scenery_tab_entries[SCENERY_WINDOW_TABS][SCENERY_ENTRIES_BY_TAB + 1];
+static ScenerySelection window_scenery_tab_entries[SCENERY_WINDOW_TABS][SCENERY_ENTRIES_BY_TAB + 1];
 
 /**
  * Was part of 0x006DFA00
  * The same code repeated five times for every scenery entry type
  */
-static void init_scenery_entry(rct_scenery_entry* sceneryEntry, int32_t index, uint8_t sceneryTabId)
+static void init_scenery_entry(rct_scenery_entry* sceneryEntry, ScenerySelection index, uint8_t sceneryTabId)
```
tupaschoal: `index` is not the best name for this anymore, maybe `currentScenery`?

AaronVanGeffen:
Thanks for splitting off the other commits to a separate PR.

You've listed the end goal, but could you describe the PR in a bit more detail? What was changed, and why?

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -1134,12 +1133,11 @@ static void scenery_eyedropper_tool_down(int16_t x, int16_t y, rct_widgetindex w
         }
         case VIEWPORT_INTERACTION_ITEM_WALL:
         {
-            int32_t entryIndex = tileElement->AsWall()->GetEntryIndex();
+            uint16_t entryIndex = tileElement->AsWall()->GetEntryIndex();
             rct_scenery_entry* sceneryEntry = get_wall_entry(entryIndex);
```
tupaschoal: Maybe just `auto`? (Would apply elsewhere)

src/openrct2/management/Research.h
```diff
@@ -129,15 +129,15 @@ void research_insert_scenery_group_entry(uint8_t entryIndex, bool researched);
...
+void scenery_set_invented(ScenerySelection sceneryItem);
+void scenery_set_not_invented(ScenerySelection sceneryItem);
 bool ride_type_is_invented(uint32_t rideType);
 bool ride_entry_is_invented(int32_t rideEntryIndex);
 uint64_t get_available_track_pieces_for_ride_type(uint8_t rideType);
 bool track_piece_is_available_for_ride_type(uint8_t rideType, int32_t trackType);
 bool scenery_group_is_invented(int32_t sgIndex);
 void scenery_group_set_invented(int32_t sgIndex);
-bool scenery_is_invented(uint16_t sceneryItem);
+bool scenery_is_invented(ScenerySelection sceneryItem);
```
IntelOrca: You should probably pass structs using `const ScenerySelection&`.

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -1211,8 +1206,8 @@ static void scenery_eyedropper_tool_down(int16_t x, int16_t y, rct_widgetindex w
  * edi : parameter_3
  */
 static void sub_6E1F34(
-    int16_t x, int16_t y, uint16_t selected_scenery, CoordsXY& gridPos, uint32_t* parameter_1, uint32_t* parameter_2,
-    uint32_t* parameter_3)
+    int16_t x, int16_t y, uint8_t sceneryType, uint16_t sceneryIndex, CoordsXY& gridPos, uint32_t* parameter_1,
```
tupaschoal: Can't this receive the struct you created instead?

src/openrct2/management/Research.h
```diff
@@ -129,15 +129,15 @@ void research_insert_scenery_group_entry(uint8_t entryIndex, bool researched);
 
 void ride_type_set_invented(uint32_t rideType);
 void ride_entry_set_invented(int32_t rideEntryIndex);
-void scenery_set_invented(uint16_t sceneryItem);
-void scenery_set_not_invented(uint16_t sceneryItem);
+void scenery_set_invented(ScenerySelection sceneryItem);
```
tupaschoal: `const &`? Maybe all these functions should

src/openrct2/world/Scenery.h
```diff
@@ -228,7 +215,27 @@ enum
     SCENERY_TYPE_PATH_ITEM,
     SCENERY_TYPE_WALL,
     SCENERY_TYPE_LARGE,
-    SCENERY_TYPE_BANNER
+    SCENERY_TYPE_BANNER,
+
+    SCENERY_TYPE_COUNT,
+};
+
+constexpr const int8_t ObjectTypeToSceneryType[OBJECT_TYPE_COUNT] = {
```
tupaschoal: I think this should be a function instead, you can assert if you're calling from a not scenery object

Gymnasiast:
@AaronVanGeffen Previously, it used an uint16_t, with the upper 8 bits being the scenery type (small, large, wall, path bit, banner) and the lower 8 bits being the entry index. I have replaced that with a struct with separate fields, which now provides 16 bits for the entry index.

src/openrct2/world/ScenerySelection.h
```diff
@@ -0,0 +1,41 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2019 OpenRCT2 developers
```
IntelOrca: 2020?

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -874,7 +874,11 @@ class S6Importer final : public IParkImporter
             bool invented = (_s6.researched_scenery_items[quadIndex] & ((uint32_t)1 << bitIndex));
 
             if (invented)
-                scenery_set_invented(sceneryEntryIndex);
+            {
+                ScenerySelection scenerySelection = { static_cast<uint8_t>((sceneryEntryIndex >> 8) & 0xFF),
+                                                      static_cast<uint16_t>(sceneryEntryIndex & 0xFF) };
```
janisozaur: You [say](https://github.com/OpenRCT2/OpenRCT2/pull/10922#issuecomment-599203500) that 

> upper 8 bits being the scenery type (small, large, wall, path bit, banner)

but https://github.com/OpenRCT2/OpenRCT2/blob/7219c880e2db58a8a8d85cafafb5958a09512107/src/openrct2/management/Research.h#L67 and https://github.com/Gymnasiast/OpenRCT2/blob/567e70bfb67c06ed301e7869401e40d78fceef97/src/openrct2/rct2/RCT2.h#L35-L37 equals to `0x700`, so you end up setting `SceneryType` to 7.

Previously this code iterated over whole array, which did not even count a multiple of 5 (`small`, `large`, `wall`, `path bit`, `banner`), so there must be something off with your initial assumption
```suggestion
                                                      static_cast<uint16_t>(sceneryEntryIndex & 0xFF) };
                assert(scenerySelection.SceneryType < SCENERY_TYPE_COUNT);
```

Gymnasiast: Thanks, writing beyond the confines of an array was indeed the cause. I have put the assert in `scenery_set_invented()` and used an `if` here, for reasons I explained on lines 881 and 882.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -880,7 +880,9 @@ void S6Exporter::ExportResearchedSceneryItems()
 
     for (uint16_t sceneryEntryIndex = 0; sceneryEntryIndex < RCT2_MAX_RESEARCHED_SCENERY_ITEMS; sceneryEntryIndex++)
```
janisozaur: There's still similar define around: `MAX_RESEARCHED_SCENERY_ITEMS`, not used any more. Can you delete it now?

PR #10923, opened 2020-03-13
Allow opening rides with unsuitable track when "Enable all drawable track pieces" is enabled.

X123M3-256:
Currently, certain rides (most notably the hyper and hyper twister coasters) cannot be opened if the layout contains inversions and/or banked track. As far as I'm aware, there's currently no way to override this behaviour.

This PR disables that check when the "Enable all drawable track pieces" cheat is selected.

I am not sure if it might be better to disable it entirely, as I'm not sure there's any way you can actually see this error message if cheats are turned off - trains that can't run on inversions usually don't allow building them either. However, I'm not sure if there are exceptions, so I've left the default behaviour as it is.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -5224,7 +5224,7 @@ int32_t ride_is_valid_for_test(Ride* ride, int32_t status, bool isApplying)
         }
     }
 
-    if (ride->subtype != RIDE_ENTRY_INDEX_NULL)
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Network version should have been bumped for this

Fred-1044276:
Testing the ride does work but you can't open it sadly.
Harpy Gardens 2020-04-13 11-09-58

Gymnasiast:
@Fred-1044276 Please create an issue for that, and make sure to include the park.

PR #10924, opened 2020-03-13
Refactor/even more coords stuff

Gymnasiast:
Split off from https://github.com/OpenRCT2/OpenRCT2/pull/10922

### DIFF
--- a/src/openrct2/actions/FootpathPlaceAction.hpp
+++ b/src/openrct2/actions/FootpathPlaceAction.hpp
@@ -76,12 +76,12 @@ public:
             return MakeResult(GA_ERROR::DISALLOWED, STR_CANT_BUILD_FOOTPATH_HERE, STR_LAND_SLOPE_UNSUITABLE);
         }
 
-        if (_loc.z / 8 < FootpathMinHeight)
+        if (_loc.z < FootpathMinHeight)
         {
             return MakeResult(GA_ERROR::DISALLOWED, STR_CANT_BUILD_FOOTPATH_HERE, STR_TOO_LOW);
         }
 
-        if (_loc.z / 8 > FootpathMaxHeight)
+        if (_loc.z > FootpathMaxHeight)
         {
             return MakeResult(GA_ERROR::DISALLOWED, STR_CANT_BUILD_FOOTPATH_HERE, STR_TOO_HIGH);
         }
--- a/src/openrct2/paint/tile_element/Paint.Surface.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Surface.cpp
@@ -518,7 +518,7 @@ static bool tile_is_inside_clip_view(const tile_descriptor& tile)
     if (tile.tile_element == nullptr)
         return false;
 
-    if (tile.tile_element->base_height > gClipHeight)
+    if (tile.tile_element->GetBaseZ() > gClipHeight / COORDS_Z_STEP)
         return false;
 
     auto coords = tile.tile_coords.ToCoordsXY();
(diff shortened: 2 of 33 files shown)


src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -1624,15 +1620,11 @@ static int32_t guest_path_find_leaving_park(Peep* peep, uint8_t edges)
...
-    int16_t y = peepSpawn->y & 0xFFE0;
-    uint8_t z = peepSpawn->z / 8;
-    Direction direction = peepSpawn->direction;
+    const auto peepSpawnLoc = gPeepSpawns[chosenSpawn].ToTileStart();
+    Direction direction = peepSpawnLoc.direction;
 
-    gPeepPathFindGoalPosition = { x / 32, y / 32, z };
-    if (x == peep->NextLoc.x && y == peep->NextLoc.y)
+    gPeepPathFindGoalPosition = TileCoordsXYZ(peepSpawnLoc);
+    if (peepSpawnLoc.x == peep->NextLoc.x && peepSpawnLoc.y == peep->NextLoc.y)
```
tupaschoal: Are these different objects or could the equality be simplified?

src/openrct2/actions/FootpathPlaceAction.hpp
```diff
@@ -76,12 +76,12 @@ DEFINE_GAME_ACTION(FootpathPlaceAction, GAME_COMMAND_PLACE_PATH, GameActionResul
             return MakeResult(GA_ERROR::DISALLOWED, STR_CANT_BUILD_FOOTPATH_HERE, STR_LAND_SLOPE_UNSUITABLE);
         }
 
-        if (_loc.z / 8 < FootpathMinHeight)
+        if (_loc.z < FootpathMinHeight)
```
tupaschoal: Isn't this a potentially different result?

Gymnasiast: Doubt it, as it won't let you build at z coordinates that aren't multiples of 1.5 metres anyway.

src/openrct2/peep/GuestPathfinding.cpp
Gymnasiast: `peepSpawnLoc` is a CoordsXYZD and `peep->NextLoc` is a CoordsXYZ. Comparing them directly will also compare the Z coordinate, which is not supposed to happen.

tupaschoal: Ok, couldn't easily check on phone, thanks

src/openrct2/actions/FootpathPlaceAction.hpp
tupaschoal: Alright

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -518,7 +518,7 @@ static bool tile_is_inside_clip_view(const tile_descriptor& tile)
     if (tile.tile_element == nullptr)
         return false;
 
-    if (tile.tile_element->base_height > gClipHeight)
+    if (tile.tile_element->GetBaseZ() > gClipHeight / COORDS_Z_STEP)
```
duncanspumpkin: Assume this has been picked up. Incorrect comparison.

src/openrct2/paint/tile_element/Paint.TileElement.cpp
```diff
@@ -245,7 +245,7 @@ static void sub_68B3FB(paint_session* session, int32_t x, int32_t y)
     do
     {
         // Only paint tile_elements below the clip height.
-        if ((session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW) && (tile_element->base_height > gClipHeight))
+        if ((session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW) && (tile_element->GetBaseZ() > gClipHeight / COORDS_Z_STEP))
```
duncanspumpkin: Incorrect comparison.

PR #10925, opened 2020-03-14
Show hovered values on finance charts

pRizz:
Implemented showing the values under the cursor on finance charts. Here is a demo gif:

finance-value-hover

### DIFF
--- a/src/openrct2-ui/interface/Graph.cpp
+++ b/src/openrct2-ui/interface/Graph.cpp
@@ -8,6 +8,7 @@
  *****************************************************************************/
 
 #include <openrct2-ui/interface/Graph.h>
+#include <openrct2/Context.h>
 #include <openrct2/localisation/Date.h>
 #include <openrct2/localisation/Localisation.h>
 
@@ -182,10 +183,128 @@ static void graph_draw_line_b_money32(
     }
 }
 
+struct FinancialTooltipInfo
+{
+    const ScreenCoordsXY coords;
+    const money32 money{};
+};
+
+struct ScreenRect
+{
+    const ScreenCoordsXY LeftTop;
+    const ScreenCoordsXY RightBottom;
+
+    int32_t GetLeft() const
+    {
+        return LeftTop.x;
+    }
+    int32_t GetTop() const
+    {
+        return LeftTop.y;
+    }
+    int32_t GetRight() const
+    {
+        return RightBottom.x;
+    }
+    int32_t GetBottom() const
+    {
+        return RightBottom.y;
+    }
+    int32_t GetWidth() const
+    {
+        return RightBottom.x - LeftTop.x;
+    }
+    int32_t GetHeight() const
+    {
+        return RightBottom.y - LeftTop.y;
+    }
+    bool Contains(const ScreenCoordsXY& coords) const
+    {
+        return coords.x >= GetLeft() && coords.x <= GetRight() && coords.y >= GetTop() && coords.y <= GetBottom();
+    }
+};
+
+static constexpr auto CHART_MAX_DATA_COUNT = 64;
+static constexpr auto CHART_MAX_INDEX = CHART_MAX_DATA_COUNT - 1;
+static constexpr auto CHART_DATA_WIDTH = 6;
+static constexpr auto CHART_MAX_WIDTH = CHART_MAX_INDEX * CHART_DATA_WIDTH;
+static constexpr auto CHART_MAX_HEIGHT = 164;
+static constexpr auto CURSOR_X_OFFSET = 3;
+static constexpr auto DEFAULT_DASHED_LENGTH = 2;
+
+static int32_t IndexForCursorAndHistory(const int32_t historyCount, const int32_t cursorX, const int32_t chartX)
+{
+    const auto offsettedCursorX = cursorX + CURSOR_X_OFFSET;
+    return (historyCount - 1) - (offsettedCursorX - chartX) / CHART_DATA_WIDTH;
+}
+
+static const ScreenCoordsXY ScreenCoordsForHistoryIndex(
+    const int32_t index, const money32* history, const int32_t chartX, const int32_t chartY, const int32_t modifier,
+    const int32_t offset)
+{
+    ScreenCoordsXY coords;
+    coords.x = chartX + CHART_DATA_WIDTH * (CHART_MAX_INDEX - index);
+    coords.y = chartY + CHART_MAX_HEIGHT - ((((history[index] >> modifier) + offset) * 170) / 256);
+    return coords;
+}
+
+static const FinancialTooltipInfo finance_tooltip_info_from_money(
+    const money32* history, const int32_t historyCount, const int32_t modifier, const int32_t offset,
+    const ScreenRect& chartFrame, const ScreenCoordsXY& cursorPosition)
+{
+    if (!chartFrame.Contains(cursorPosition))
+    {
+        return { {}, MONEY32_UNDEFINED };
+    }
+
+    const auto historyIndex = IndexForCursorAndHistory(historyCount, cursorPosition.x, chartFrame.GetLeft());
+    const auto coords = ScreenCoordsForHistoryIndex(
+        historyIndex, history, chartFrame.GetLeft(), chartFrame.GetTop(), modifier, offset);
+
+    return { { coords.x, coords.y }, history[historyIndex] };
+}
+
+static void graph_draw_hovered_value(
+    rct_drawpixelinfo* dpi, const money32* history, const int32_t historyCount, const int32_t baseX, const int32_t baseY,
+    const int32_t modifier, const int32_t offset)
+{
+    const auto cursorPosition = context_get_cursor_position_scaled();
+    const ScreenRect chartFrame{ { baseX, baseY }, { baseX + CHART_MAX_WIDTH, baseY + CHART_MAX_HEIGHT } };
+
+    if (!chartFrame.Contains(cursorPosition))
+    {
+        return;
+    }
+
+    const auto info = finance_tooltip_info_from_money(
+        history, CHART_MAX_DATA_COUNT, modifier, offset, chartFrame, cursorPosition);
+
+    if (info.money == MONEY32_UNDEFINED)
+    {
+        return;
+    }
+
+    gfx_draw_dashed_line(dpi, info.coords.x, chartFrame.GetTop(), info.coords.x, info.coords.y, DEFAULT_DASHED_LENGTH, 0);
+    gfx_draw_dashed_line(dpi, chartFrame.GetLeft() - 10, info.coords.y, info.coords.x, info.coords.y, DEFAULT_DASHED_LENGTH, 0);
+
+    if (cursorPosition.y > info.coords.y)
+    {
+        gfx_draw_dashed_line(dpi, info.coords.x, info.coords.y, info.coords.x, cursorPosition.y, DEFAULT_DASHED_LENGTH, 0);
+    }
+
+    gfx_draw_string_centred(
+        dpi, STR_FINANCES_SUMMARY_EXPENDITURE_VALUE, info.coords.x, info.coords.y - 16, COLOUR_BLACK, &info.money);
+
+    gfx_fill_rect(dpi, info.coords.x - 2, info.coords.y - 2, info.coords.x + 2, info.coords.y + 2, PALETTE_INDEX_10);
+    gfx_fill_rect(dpi, info.coords.x - 1, info.coords.y - 1, info.coords.x + 1, info.coords.y + 1, PALETTE_INDEX_21);
+}
+
 void graph_draw_money32(
-    rct_drawpixelinfo* dpi, money32* history, int32_t count, int32_t baseX, int32_t baseY, int32_t modifier, int32_t offset)
+    rct_drawpixelinfo* dpi, const money32* history, const int32_t count, const int32_t baseX, const int32_t baseY,
+    const int32_t modifier, const int32_t offset)
 {
     graph_draw_months_money32(dpi, history, count, baseX, baseY);
     graph_draw_line_a_money32(dpi, history, count, baseX, baseY, modifier, offset);
     graph_draw_line_b_money32(dpi, history, count, baseX, baseY, modifier, offset);
+    graph_draw_hovered_value(dpi, history, count, baseX, baseY, modifier, offset);
 }
--- a/src/openrct2/drawing/NewDrawing.cpp
+++ b/src/openrct2/drawing/NewDrawing.cpp
@@ -19,6 +19,8 @@
 #include "IDrawingContext.h"
 #include "IDrawingEngine.h"
 
+#include <cmath>
+
 using namespace OpenRCT2;
 using namespace OpenRCT2::Drawing;
 using namespace OpenRCT2::Paint;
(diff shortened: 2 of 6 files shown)


pRizz:
Hmm the build errors don't give much help as to what the problem is. Can someone help?

AaronVanGeffen:
The verbosity makes it harder to find, but the errors seem pretty clear:

```
../src/openrct2-ui/interface/Graph.cpp:195:8: error: 'const' type qualifier on return type has no effect [-Werror,-Wignored-qualifiers]
static const int32_t indexForCursorAndHistory(const int32_t& historyCount, const int32_t& cursorX, const int32_t& chartX)
       ^~~~~~
../src/openrct2-ui/interface/Graph.cpp:248:16: error: unused variable 'chartX' [-Werror,-Wunused-variable]
    const auto chartX = chartFrame.GetLeft() + chartMaxWidth - chartWidth;
               ^
2 errors generated.
```

pRizz:
Wierd. For some reason, the log wasn't showing me those errors. I'll fix it. Thanks.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,90 @@ static void graph_draw_line_b_money32(
     }
 }
 
+struct financial_tooltip_info
+{
+    const CoordsXY coords;
+    const money32 money;
```
IntelOrca:
```suggestion
    const money32 money{};
```

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,90 @@ static void graph_draw_line_b_money32(
...
+};
+
+static const auto chartMaxWidth = 63 * 6;
+static const auto chartMaxHeight = 164;
```
IntelOrca: Use `constexpr`

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,90 @@ static void graph_draw_line_b_money32(
...
+{
+    CoordsXY coords;
+    coords.x = chartX + 6 * (63 - index);
+    coords.y = chartY + 170 - 6 - ((((history[index] >> modifier) + offset) * 170) / 256);
+    return coords;
+}
+
+static const financial_tooltip_info finance_tooltip_info_from_money(
+    const money32* history, const int32_t& historyCount, const int32_t& modifier, const int32_t& offset,
+    const MapRange& chartFrame, const ScreenCoordsXY& cursorPosition)
```
IntelOrca: Seems odd to use `MapRange` for something that isn't the map.

src/openrct2/drawing/NewDrawing.cpp
```diff
@@ -209,6 +211,28 @@ void gfx_draw_line(rct_drawpixelinfo* dpi, int32_t x1, int32_t y1, int32_t x2, i
...
+    const int32_t dashedLineSegmentLength, const int32_t colour)
+{
+    const auto drawingEngine = dpi->DrawingEngine;
+    if (drawingEngine != nullptr)
+    {
+        const auto dashedLineLength = hypot(x2 - x1, y2 - y1);
+        const auto lineSegmentCount = dashedLineLength / dashedLineSegmentLength / 2;
+        const auto angle = atan2(y2 - y1, x2 - x1);
+        const auto dx = dashedLineSegmentLength * cos(angle);
+        const auto dy = dashedLineSegmentLength * sin(angle);
```
IntelOrca: So far we have avoided floating point numbers everywhere in the game apart from the OpenGL coding. Is it possible for you to make this integer based?

pRizz: Don't these get coerced to integers later, anyways?

src/openrct2-ui/interface/Graph.cpp
pRizz: Are there better structs for this kind of abstraction?

src/openrct2/drawing/NewDrawing.cpp
janisozaur: Doesn't matter, the intermediate step with floats is enough to introduce inconsistencies across clients.

If possible, please avoid them.

IntelOrca: > introduce inconsistencies across clients.

This is drawing code so that isn't a problem. But it is just that we have done so well to avoid them, we would like to keep that if possible for the software renderer.

Using triganometry for a dashed line seems overkill anyway. You should just be able to just work out the width and height ratios for each dash. Multiply them up to large integers and then down again to avoid losing precision.

src/openrct2-ui/interface/Graph.cpp
IntelOrca: Hmm I don't think so, but you are welcome to make a `Rect` struct.

src/openrct2/drawing/NewDrawing.cpp
pRizz: I tested diagonals with your idea of multiplying big and going back down to avoid losing precision. It seems to work decently, so I will commit that.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,129 @@ static void graph_draw_line_b_money32(
...
+    {
+        return RightBottom.x - LeftTop.x;
+    }
+    int32_t GetHeight() const
+    {
+        return RightBottom.y - LeftTop.y;
+    }
+    bool Contains(const CoordsXY& coords) const
+    {
+        return coords.x >= LeftTop.x && coords.x <= RightBottom.x && coords.y >= LeftTop.y && coords.y <= RightBottom.y;
```
tupaschoal: I'd use `GetLeft`, `GetRight`, etc instead of the hardcoded internal variables, but that's my personal preference.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,129 @@ static void graph_draw_line_b_money32(
...
+struct financial_tooltip_info
+{
+    const CoordsXY coords;
+    const money32 money{};
+};
+
+struct Rect
+{
+    const CoordsXY LeftTop;
+    const CoordsXY RightBottom;
```
tupaschoal: Both of these should be `ScreenCoordsXY` I guess.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,129 @@ static void graph_draw_line_b_money32(
...
+    bool Contains(const ScreenCoordsXY& coords) const
+    {
+        return Contains(CoordsXY(coords.x, coords.y));
+    }
+};
+
+static constexpr auto chartMaxWidth = 63 * 6;
+static constexpr auto chartMaxHeight = 164;
+
+static int32_t indexForCursorAndHistory(const int32_t& historyCount, const int32_t& cursorX, const int32_t& chartX)
```
tupaschoal: No need for `const &` on `int32_t` and primitives smaller than integers.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,129 @@ static void graph_draw_line_b_money32(
...
+{
+    const auto cursorOffset = 3;
```
tupaschoal: Make it a `constexpr` and move out of the function.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,129 @@ static void graph_draw_line_b_money32(
...
+    const auto offsettedCursorX = cursorX + cursorOffset;
+    return (historyCount - 1) - (offsettedCursorX - chartX) / 6;
+}
+
+static const CoordsXY coordsForHistoryIndex(
+    const int32_t& index, const money32* history, const int32_t& chartX, const int32_t& chartY, const int32_t& modifier,
+    const int32_t& offset)
+{
+    CoordsXY coords;
+    coords.x = chartX + 6 * (63 - index);
```
tupaschoal: `63` and `6` should be extracted to constants, they're used all over, including on other constants, though I didn't check if it's just a coincidence

src/openrct2-ui/interface/Graph.cpp
pRizz: I was wondering about the convention for this 👍

IntelOrca: It probably doesn't matter after all compiler optimisation has taken place but avoid `const &` for primitives but always do it for structs. Any struct that fits in a register should be optimised by compiler like so otherwise you always force a copy to happen which if can be expensive depending on the copy constructor of the type.

tupaschoal: A very simple explanation:

A reference `&` is inherently a pointer.
Most primitives `bool`, `int`, `float` and etc will be the size of the pointer or smaller, so copying them or copying the pointer is kind of the same. For those, it's better to just copy them altogether.

Classes and structs, on the other hand, will typically contain multiple variables which makes using the reference a lot better for performance reasons, as you avoid copying stuff.

janisozaur: @IntelOrca it does matter for compiler optimisation. Taking a reference forces instantiation on stack, passing by value lets compiler work its magic.

Verify it yourself: https://godbolt.org/z/AWr_iM

pRizz: Sounds good. I made the variables `const` since they don't need to be modified.

IntelOrca: @janisozaur When I said after "all" compiler optimisation, I am including link time optimisation. Try calling `bar` from `foo` and you will see it optimise.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,131 @@ static void graph_draw_line_b_money32(
...
+static constexpr auto cursorXOffset = 3;
+static constexpr auto defaultDashedLength = 2;
+
+static int32_t indexForCursorAndHistory(const int32_t historyCount, const int32_t cursorX, const int32_t chartX)
+{
+    const auto offsettedCursorX = cursorX + cursorXOffset;
+    return (historyCount - 1) - (offsettedCursorX - chartX) / chartDataWidth;
+}
+
+static const CoordsXY coordsForHistoryIndex(
```
IntelOrca: A lot of files still use the old lower case with underscore function names but our newer code uses TitleCase. This follows neither convention so I suggest making it Title Case.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,131 @@ static void graph_draw_line_b_money32(
     }
 }
 
+struct financial_tooltip_info
```
IntelOrca:
```suggestion
struct FinancialTooltipInfo
```

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,131 @@ static void graph_draw_line_b_money32(
...
+    }
+};
+
+static constexpr auto chartMaxDataCount = 64;
+static constexpr auto chartDataWidth = 6;
+static constexpr auto chartMaxIndex = chartMaxDataCount - 1;
+static constexpr auto chartMaxWidth = chartMaxIndex * chartDataWidth;
+static constexpr auto chartMaxHeight = 164;
+static constexpr auto cursorXOffset = 3;
+static constexpr auto defaultDashedLength = 2;
```
IntelOrca: We tend to use UPPER_CASE for constants.

src/openrct2/drawing/NewDrawing.cpp
```diff
@@ -209,6 +211,39 @@ void gfx_draw_line(rct_drawpixelinfo* dpi, int32_t x1, int32_t y1, int32_t x2, i
...
+
+        const int32_t dashedLineLength = hypot(x2 - x1, y2 - y1);
+        const int32_t lineSegmentCount = dashedLineLength / dashedLineSegmentLength / 2;
+        if (lineSegmentCount == 0)
+        {
+            return;
+        }
+
+        const int32_t lineXDist = abs(x2 - x1);
+        const int32_t lineYDist = abs(y2 - y1);
```
IntelOrca:
```suggestion
        const int32_t lineYDist = std::abs(y2 - y1);
```

src/openrct2-ui/interface/Graph.cpp
AaronVanGeffen: Not in the newer cpp code, though, right?

src/openrct2/drawing/NewDrawing.cpp
```diff
@@ -209,6 +211,39 @@ void gfx_draw_line(rct_drawpixelinfo* dpi, int32_t x1, int32_t y1, int32_t x2, i
...
+    const int32_t dashedLineSegmentLength, const int32_t colour)
+{
+    assert(dashedLineSegmentLength > 0);
+
+    const auto drawingEngine = dpi->DrawingEngine;
+    if (drawingEngine != nullptr)
+    {
+        constexpr int32_t precisionFactor = 1000;
+
+        const int32_t dashedLineLength = hypot(x2 - x1, y2 - y1);
```
IntelOrca: `hypot` is still a floating point operation.

src/openrct2-ui/interface/Graph.cpp
IntelOrca: Example?

AaronVanGeffen: Too numerous:
```
$ grep -RE 'constexpr [^ ]+? [A-Z][a-z]+' src/ | wc -l
273
```

Across a number of files, too
```
$ grep -RE 'constexpr [^ ]+? [A-Z][a-z]+' -l src/ | wc -l
135
```

IntelOrca: You are probably picking up constexpr functions as well though right? I see arrays also tend to be TitleCase, but there are a lot also UPPER_CASE.

src/openrct2-ui/interface/Graph.cpp
pRizz: Should I make all the new functions TitleCase, or just the ones that don't follow either convention?

src/openrct2-ui/interface/Graph.cpp
pRizz: I went ahead and made the constants UPPER_CASE

pRizz:
I found what seems to be a sufficient utility rect struct in `SDL_rect.h`, `SDL_Rect`. Should I use this instead of my own?

duncanspumpkin:
No the SDL one should be kept on its own. No point making SDL more intertwined into the codebase.

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,132 @@ static void graph_draw_line_b_money32(
...
+    }
+    int32_t GetWidth() const
+    {
+        return RightBottom.x - LeftTop.x;
+    }
+    int32_t GetHeight() const
+    {
+        return RightBottom.y - LeftTop.y;
+    }
+    bool Contains(const CoordsXY& coords) const
```
duncanspumpkin: This is dangerous and shouldn't be performed. CoordsXY represent map coordinates this Rect represents screen coordinates they should never be compared directly (apart from on the map window but this isn't used for the map window)

src/openrct2-ui/interface/Graph.cpp
```diff
@@ -182,10 +183,132 @@ static void graph_draw_line_b_money32(
...
+static constexpr auto CURSOR_X_OFFSET = 3;
+static constexpr auto DEFAULT_DASHED_LENGTH = 2;
+
+static int32_t IndexForCursorAndHistory(const int32_t historyCount, const int32_t cursorX, const int32_t chartX)
+{
+    const auto offsettedCursorX = cursorX + CURSOR_X_OFFSET;
+    return (historyCount - 1) - (offsettedCursorX - chartX) / CHART_DATA_WIDTH;
+}
+
+static const CoordsXY CoordsForHistoryIndex(
```
duncanspumpkin: ScreenCoordsXY

src/openrct2-ui/interface/Graph.cpp
IntelOrca: It's up to you. I don't mind.

duncanspumpkin:
@pRizz are you able to rebase this?

PR #10930, opened 2020-03-14
Fix code formatting on macOS

pRizz:
This command break by default because the `readlink` command on macOS does not support the `-f` flag. Recommend installing `greadlink` from Homebrew as a workaround.

### DIFF
--- a/scripts/check-code-formatting
+++ b/scripts/check-code-formatting
@@ -1,8 +1,18 @@
 #!/bin/sh
 set -e
 
+readlink_bin='readlink'
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  if ! [ -x "$(command -v greadlink)" ]; then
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Running this script on macOS, I get clang-format output for the `thirdparty` folder as well, despite the script explicitly excluding it. Does this happen for you as well, @pRizz?

pRizz:
@AaronVanGeffen Mine does not get output for the `thirdparty` folder.
Screen Shot 2020-03-15 at 4 41 05 PM

tupaschoal:
@AaronVanGeffen can you comment on next steps, please?

janisozaur:
Why do we need `-f`? Looking at the patch context it's only to canonicalize the symlinks, but nothing really depends on it and it should work just as well with non-canonical path: we cd to it anyway… so why is it required?

PR #10931, opened 2020-03-14
Add script for formatting code in place

pRizz:


AaronVanGeffen:
Is there any advantage of this over `git clang-format`?

pRizz:
@AaronVanGeffen Interesting 😅I was not aware of that. I wish this was documented somewhere when I was doing other pull requests. How about I just document this?

AaronVanGeffen:
Yes, it should really be on the wiki, if it isn't already. The [Coding Style](https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style) lemma ought to reflect this at least.

tupaschoal:
A note on CONTRIBUTING.md sounds like a good addition as well.

pRizz:
I just added this to the Coding Style page:
https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style/ff1f60e0eedde158ca065967e7d6026f9230d58f

Do we also want it in the CONTRIBUTING.md ?

PR #10941, opened 2020-03-15
Fix Clear Scenery refunding ghost tile elements

Gymnasiast:


AaronVanGeffen:
Nice catch. How did you get ghost elements to exist while using the clear scenery tool, though?

Gymnasiast:
@AaronVanGeffen You can have path ghosts while using the Clear Scenery tool. The video that reported this bug can be reported here: https://www.reddit.com/r/rct/comments/fj15i7/came_across_this_money_glitch_in_open_rct2_is/

duncanspumpkin:
Did you increment the network version after this?

Gymnasiast:
I have now.

PR #10875, opened 2020-03-05
Refactor more TileCoords

Gymnasiast:
Also converts a lot of Z calculation to use big Zs.

PR #10945, opened 2020-03-16
Fix #10942: Cut-away view is broken

### LINKED ISSUE #10942: cut-away view is broken
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 7219c880e

When trying to use the cut-away view, the entire screen turns black. Some sprites are still visible, but as soon as you try to select an area or when you clear the selection, these disappear as well.

**Edit**: I just noticed that tile elements with the lowest possible base height (2) are still visible.

**Steps to reproduce:**
1. Open any map
2. Open the cutaway view from the View options toolbar button (eye)
3. Open the cut-away View
4. See how the entire map turns black, but sprites are still visible
5. Clear/set new selection
6. Observe how the sprites now disappear too

Gymnasiast:


tupaschoal:
I found it a bit weird that the fixes are on lines not touched by #10875 (the culprit, according to #10942)

src/openrct2-ui/windows/ViewClipping.cpp
```diff
@@ -242,7 +242,7 @@ static void window_view_clipping_mouseup(rct_window* w, rct_widgetindex widgetIn
             _previousClipSelectionA = gClipSelectionA;
             _previousClipSelectionB = gClipSelectionB;
             gClipSelectionA = { 0, 0 };
-            gClipSelectionB = { MAXIMUM_MAP_SIZE_TECHNICAL - 1, MAXIMUM_MAP_SIZE_TECHNICAL - 1 };
+            gClipSelectionB = { MAXIMUM_MAP_SIZE_BIG - 1, MAXIMUM_MAP_SIZE_BIG - 1 };
```
Broxzier: I don't think this is right. `gClipSelectionA` and `gClipSelectionB` are of type `CoordsXY`, but `MAXIMUM_MAP_SIZE_BIG` is for big coords, right?

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -518,7 +518,7 @@ static bool tile_is_inside_clip_view(const tile_descriptor& tile)
     if (tile.tile_element == nullptr)
         return false;
 
-    if (tile.tile_element->GetBaseZ() > gClipHeight / COORDS_Z_STEP)
+    if (tile.tile_element->GetBaseZ() > gClipHeight * COORDS_Z_STEP)
```
Broxzier: I think if you revert the assignment in ViewClipping and just remove the code `/ COORDS_Z_STEP`, we have the same result. Not tested yet.

src/openrct2/paint/tile_element/Paint.TileElement.cpp
```diff
@@ -245,7 +245,7 @@ static void sub_68B3FB(paint_session* session, int32_t x, int32_t y)
     do
     {
         // Only paint tile_elements below the clip height.
-        if ((session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW) && (tile_element->GetBaseZ() > gClipHeight / COORDS_Z_STEP))
+        if ((session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW) && (tile_element->GetBaseZ() > gClipHeight * COORDS_Z_STEP))
```
Broxzier: Same here.

src/openrct2-ui/windows/ViewClipping.cpp
Gymnasiast: CoordsXY **are** big coords.

src/openrct2/paint/tile_element/Paint.Surface.cpp
Gymnasiast: Absolutely not. gClipHeight contains small Z coordinates, except for the one or two assignments and comparisons I mistakingly changed in earlier PRs.

PR #10947, opened 2020-03-16
Name remaining ride type flags

Gymnasiast:


src/openrct2/ride/Ride.h
```diff
@@ -869,7 +869,8 @@ enum ride_type_flags : uint32_t
     RIDE_TYPE_FLAG_NO_VEHICLES = 1 << 13, // used only by maze, spiral slide and shops
     RIDE_TYPE_FLAG_HAS_LOAD_OPTIONS = 1 << 14,
     RIDE_TYPE_FLAG_HAS_NO_TRACK = 1 << 15,
-    RIDE_TYPE_FLAG_16 = 1 << 16, // something to do with vehicle colour scheme
+    RIDE_TYPE_FLAG_VEHICLE_IS_INTEGRAL = 1 << 16, // Set by flat rides where the vehicle is integral to the structure, like
```
tupaschoal: I wonder if "monolithic" is better than "integral"

Gymnasiast: Many vehicles that are monolithic are not integral. For example, coasters boats, rowing boats, but also dodgems - they're all monolithic (consisting of one, unarticulated piece), but none of those are integral to the ride structure.

pizza2004: I think monolithic is meant to refer to the ride as a whole, in that you can't separate the vehicle from the ride at all since it's one big piece.  That said, I think it makes enough sense either way not to matter.

PR #10948, opened 2020-03-16
Fix #10914: Missing sounds, including guest purchase sound

### LINKED ISSUE #10914: Missing sounds, including "ching" sound effect when a guest makes a purchase.
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** a2d8078

Even if the park has no rides playing music and the guest count is below 400, the "Ching" sound effect heard with a guest purchase either rarely plays or no longer plays. Although I still make money from guests making purchases in the park, the lack of the "Ching" sound effect can be offputting.

This issue currently does not exist in the v0.2.4 stable release.

- Reproducible in RCT2 (vanilla)? No.
- Specific to multiplayer? No.

**Steps to reproduce:**
1. Start up a new scenario. Cannot be a scenario with no money.

2. Build some rides and stalls and build up your guest count.

3. Select a ride or stall that's currently making money and click the "View of ride/attraction" tab again to hear only the ride/stall (for rides, a station/docking platform with an entrance should be selected).

4. If you don't hear the "ching" sound effect at all, then you have successfully reproduced the bug/issue.

5. I have tested this bug on both densely populated parks (+2,000 guests) and sparsely populated parks (1~400 guests). The bug occurs regardless of whether or not the "Show guest purchases as animation" option is toggled on.

**Dump file**
None here, as the bug does not crash the game.

**Screenshots / Video:**
None here.

**Save game:**
[No ching sound bug.zip](https://github.com/OpenRCT2/OpenRCT2/files/4322410/No.ching.sound.bug.zip)
This zip file contains a custom scenario I made and posted in the Marcel's Hangout Discord server. I also used this to reproduce this bug.

AaronVanGeffen:
Regression from 09fad0ef7f15ce074c387d1bd9c27789ff827828 (part of #10854)

### DIFF
--- a/src/openrct2/audio/Audio.cpp
+++ b/src/openrct2/audio/Audio.cpp
@@ -218,8 +218,8 @@ static AudioParams audio_get_params_from_location(SoundId soundId, const CoordsX
     {
         if (viewport->flags & VIEWPORT_FLAG_SOUND_ON)
         {
-            int16_t vy = pos2.y - viewport->viewPos.x;
-            int16_t vx = pos2.x - viewport->viewPos.y;
+            int16_t vy = pos2.y - viewport->viewPos.y;
+            int16_t vx = pos2.x - viewport->viewPos.x;
             params.pan = viewport->pos.x + (vx >> viewport->zoom);
             params.volume = SoundVolumeAdjust[static_cast<uint8_t>(soundId)]
                 + ((-1024 * viewport->zoom - 1) * (1 << volumeDown)) + 1;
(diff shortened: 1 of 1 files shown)


src/openrct2/audio/Audio.cpp
```diff
@@ -218,8 +218,8 @@ static AudioParams audio_get_params_from_location(SoundId soundId, const CoordsX
     {
         if (viewport->flags & VIEWPORT_FLAG_SOUND_ON)
         {
-            int16_t vy = pos2.y - viewport->viewPos.x;
-            int16_t vx = pos2.x - viewport->viewPos.y;
+            int16_t vy = pos2.y - viewport->viewPos.y;
+            int16_t vx = pos2.x - viewport->viewPos.x;
```
Gymnasiast: I'd prefer if you also switched these two lines around. That's what induced the mistake in the first place.

tupaschoal:
Thanks!

PR #10949, opened 2020-03-16
Fix setting locale on linux

Gymnasiast:
Using the nl_NL.UTF8 locale, previously:

afbeelding

Now:

afbeelding

### DIFF
--- a/src/openrct2/platform/Platform.Posix.cpp
+++ b/src/openrct2/platform/Platform.Posix.cpp
@@ -13,6 +13,7 @@
 #    include "Platform2.h"
 #    include "platform.h"
 
+#    include <clocale>
 #    include <cstdlib>
 #    include <cstring>
 #    include <ctime>
@@ -72,6 +73,7 @@ namespace Platform
 
     std::string FormatShortDate(std::time_t timestamp)
     {
+        setlocale(LC_TIME, "");
         char date[20];
         std::strftime(date, sizeof(date), "%x", std::localtime(&timestamp));
         return std::string(date);
@@ -79,6 +81,7 @@ namespace Platform
 
     std::string FormatTime(std::time_t timestamp)
     {
+        setlocale(LC_TIME, "");
         char time[20];
         std::strftime(time, sizeof(time), "%X", std::localtime(&timestamp));
         return std::string(time);


AaronVanGeffen:
I don't have this problem on macOS, but I'm not convinced whether it's necessary to single out Linux like this. It wouldn't do any harm to other platforms, right?

Gymnasiast:
@AaronVanGeffen I'm not entirely sure. I seem to recall there were once some problems on macOS that were resolved by removing or changing a call to setlocale(). 

I can change it, but the first macOS locale bug we get, you get shot. ;-)

AaronVanGeffen:
@Gymnasiast I'm just worried that the whitelist will grow quite rapidly — the BSD platforms might require it as well, for example. However, apparently, you've already had to make an exception for Android as well. Any particular reason?

Gymnasiast:
@AaronVanGeffen It won't compile under Android.

As for BSD: @ibara could you say if the code in this PR would also be proper for OpenBSD?

AaronVanGeffen:
> It won't compile under Android.

Oh, that's odd. Do you still have the compilation error logged somewhere?

Gymnasiast:
It was saying that `LC_TIME` was not defined. (Don't have access to the logs any more.)

AaronVanGeffen:
That would probably be resolved by explicitly including `<clocale>`.

ibara:
> @AaronVanGeffen It won't compile under Android.
> 
> As for BSD: @ibara could you say if the code in this PR would also be proper for OpenBSD?

The code as-is won't catch on OpenBSD (or any of the BSDs) but there is no harm in this for the BSDs (assuming you have the proper includes, as is referenced above).

Gymnasiast:
@ibara Hm, could you elaborate on why this code would not catch on OpenBSD?

ibara:
```
#if defined(__linux__) && !defined(__ANDROID__)
```
None of the BSDs define a __linux__ macro.

Gymnasiast:
@ibara That is obvious. I was asking about the code _inside_ the ifdef block - will it be okay if I extend the ifdef so that the code will also be executed on OpenBSD?

ibara:
Yes that would be fine.

IntelOrca:
My only concern with this, is whether PlatformEnvironment is the right place. Most of the code here is not platform specific, this kind of stuff goes in Platform, Platform.*.cpp etc.

Gymnasiast:
@IntelOrca Perhaps we should add a method to Platform called `DoPlatformSetup()` or something similar.

IntelOrca:
We already have calls to `setlocale` in `platform_get_locale_currency`, `platform_get_locale_measurement_format`. They are used to set our own internal format for various things.

I think we could perhaps limit this to `Platform.Posix.cpp`. In the routines: FormatShortDate, FormatTime, we could do the `setlocale` call, or lazy initialise it (initialise it first time through but not consecutive calls). Maybe it is good practice to always call `setlocale` before a format in case it is changed. I don't know how expensive it is or how thread safe it is.

Gymnasiast:
@IntelOrca I was thinking it should be the other way round: use them once in the general platform setup, so we don't have to execute them in `platform_get_locale_currency()`.

IntelOrca:
That works for those functions, but the date and time uses `std::strftime` which relies on the global value of locale for every call to it. You can set the locale in the init function but if it is ever changed, your screwed.

Anyway I don't mind there being a platform initialisation function for every platform.

PR #10953, opened 2020-03-17
Fix #10928: File browser's date column is too narrow

### LINKED ISSUE #10928: File browser 'date' column is too narrow
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 9b5e814

It doesn't fit most 2020 dates.
openrct2_2020-03-14_14-14-08

AaronVanGeffen:
The width of this column is font-dependent, and so I initially reasoned, at the time, that taking the width of the widest date I could think of would be most appropriate. However, I made a mistake by taking October (10) as the 'widest' month, as indeed February (02) or March (03) are both wider with their leading zeros. January (01) doesn't have this problem.

So, 2020 was a red herring. This comment illustrates the fix:

```diff
-    // Generate a time object for a relatively wide time: 2000-10-20 00:00:00
+    // Generate a time object for a relatively wide time: 2000-02-20 00:00:00
```

This assumes leading zeroes are included in the platform's locale, though, which isn't always the case. For Japanese, for example, I'd get 2020年３月17日. I have added an additional check that does use the date I originally coded in. The widest of the two widths is used.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -5,6 +5,7 @@
 - Fix: [#475] Water sides drawn incorrectly (original bug).
 - Fix: [#6123, #7907, #9472, #11028] Cannot build some track designs with 4 stations (original bug).
 - Fix: [#7094] Back wall edge texture in water missing.
+- Fix: [#10928] File browser's date column is too narrow.
 - Fix: [#11005] Company value overflows.
 - Fix: [#11027] Third color on walls becomes black when saving.
 
--- a/src/openrct2-ui/windows/LoadSave.cpp
+++ b/src/openrct2-ui/windows/LoadSave.cpp
@@ -637,16 +637,16 @@ static void window_loadsave_textinput(rct_window* w, rct_widgetindex widgetIndex
 
 static void window_loadsave_compute_max_date_width()
 {
-    // Generate a time object for a relatively wide time: 2000-10-20 00:00:00
+    // Generate a time object for a relatively wide time: 2000-02-20 00:00:00
     std::tm tm;
     tm.tm_sec = 0;
     tm.tm_min = 0;
     tm.tm_hour = 0;
     tm.tm_mday = 20;
-    tm.tm_mon = 9;
+    tm.tm_mon = 2;
     tm.tm_year = 100;
     tm.tm_wday = 5;
-    tm.tm_yday = 294;
+    tm.tm_yday = 51;
     tm.tm_isdst = -1;
 
     std::time_t long_time = mktime(&tm);
--- a/src/openrct2-ui/windows/LoadSave.cpp
+++ b/src/openrct2-ui/windows/LoadSave.cpp
@@ -635,6 +635,8 @@ static void window_loadsave_textinput(rct_window* w, rct_widgetindex widgetIndex
     }
 }
 
+constexpr uint16_t DATE_TIME_GAP = 2;
+
 static void window_loadsave_compute_max_date_width()
 {
     // Generate a time object for a relatively wide time: 2000-02-20 00:00:00
@@ -653,7 +655,7 @@ static void window_loadsave_compute_max_date_width()
 
     // Check how how this date is represented (e.g. 2000-02-20, or 00/02/20)
     std::string date = Platform::FormatShortDate(long_time);
-    maxDateWidth = gfx_get_string_width(date.c_str());
+    maxDateWidth = gfx_get_string_width(date.c_str()) + DATE_TIME_GAP;
 
     // Some locales do not use leading zeros for months and days, so let's try October, too.
     tm.tm_mon = 10;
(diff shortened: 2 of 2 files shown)


AaronVanGeffen:
@Umdlye Could you confirm this indeed fixes the bug on your end?

Umdlye:
Some dates are still cut off:
openrct2_2020-03-17_18-56-31

The missing numbers from top to bottom: `07`, `23`, `24`

AaronVanGeffen:
Thanks for the feedback. I feel like I'm missing something, but I'm not sure what yet.

Gymnasiast:
@AaronVanGeffen Is the gap between the date and the time taken into account?

AaronVanGeffen:
@Gymnasiast Yes, both have their own limits specified:

https://github.com/OpenRCT2/OpenRCT2/blob/e27a1a230a0c6e8b73c46dc04893068edb64837f/src/openrct2-ui/windows/LoadSave.cpp#L775-L779

Note that the spacing/offsets are not part of these widths.

Umdlye:
I tried changing `tm.tm_hour` to `20`, now the cut off dates in the above screenshot appear normally, but the seconds for the highlighted save in the screenshot and the one below it are cut off (`44` and `04`).

AaronVanGeffen:
I tried brute-forcing it:

```cpp
    std::tm tm;
    /* ... */
    std::time_t long_time;

    for (int month = 1; month <= 12; month++)
    {
        tm.tm_mon = month;
        for (int day = 1; day <= 31; day++)
        {
            tm.tm_mday = day;

            long_time = mktime(&tm);

            std::string date = Platform::FormatShortDate(long_time);
            int32_t dateWidth = gfx_get_string_width(date.c_str());

            if (dateWidth > maxDateWidth)
            {
                printf("%dM-%dD yields wider string (%dpx) than previous record (%dpx)\n", month, day, dateWidth, maxDateWidth);
                printf("%s\n", date.c_str());
                maxDateWidth = dateWidth;
            }
        }
    }
```

Off-by-one errors aside, this only prints:
```
1M-1D yields wider string (56px) than previous record (0px)
02/01/00
1M-2D yields wider string (58px) than previous record (56px)
02/02/00
```
to the console for me. I obtain the same width with the code from this PR.

I still need to verify this with other date formats, but I don't think the problem is in the heuristics.

AaronVanGeffen:
@Umdlye I don't know what to tell you… I just can't reproduce the problem over here. :confused:

Six Flags over Texas 2020-03-18 21-17-12

Six Flags over Texas 2020-03-18 21-17-23

These were generated using the PR in its current state, save for the use of custom (longer) date formats. Works with the normal ones, too.

Umdlye:
Could you zip up that tests folder and send it over by any chance?

AaronVanGeffen:
Sure: [tests.zip](https://github.com/OpenRCT2/OpenRCT2/files/4351284/tests.zip)
Generated using:
```
seq 1 29 | { while read i; do touch --date="2020-02-$i" "test$i.sv6"; done; }
```

Umdlye:
openrct2_2020-03-18_21-47-52
😭

AaronVanGeffen:
That's completely unacceptable. Thanks for testing.

I notice you've got some extra free space on the right of the canvas. Why does the text drawing routines decide to cut off text on your pc, but not on mine? I wonder…

Umdlye:
After changing locale settings to UK defaults:
openrct2_2020-03-18_21-59-53

AaronVanGeffen:
There must be something in what the Windows platform functions are doing that is different from their Linux counterparts… :thinking:

AaronVanGeffen:
@Umdlye I've added some 'grace pixels' to both width specifiers. Could you try whether this resolves the issue for you?

Umdlye:
Doesn't look like it: 
openrct2_2020-03-28_22-13-07

AaronVanGeffen:
Thanks for testing. Okay, this is grasping at straws, but let's try two grace pixels, then…

(I'm inclined to rewrite the whole thing, but if this works, I'll happily ship it in the mean time.)

Umdlye:
That did it :) 
openrct2_2020-03-28_23-24-57

AaronVanGeffen:
Thanks for testing again. Glad that worked.

Effectively, this means it's up to four pixels short. That's the same number of pixels as the gap in between date and time. This can't be a coincidence. Why this does not present an issue on macOS or Linux is the real mystery.

I want to get to the bottom of this (and clean all this up a bit) before merging. Stay tuned.

AaronVanGeffen:
Possibly diving too deep into this, but for completeness sake, I've decided to make a direct comparison with the 'grace pixels' in place. If I change my date formatting to match @Umdlye's, I get the following load/save listing:

loadsave_aaron

Now, if we count the number of pixels between date and time, as well as time and scrollbar, we get the following results:

| User | Date-time distance (px) | Time-scroll distance |
| ------ | ------------------------------- | -------------------------- |
| Umdlye |147 | 5 | 2 |
| Aaron | 152 | 5 | 4 |

I think this is very interesting. Why do the column widths differ between platforms if the formatting is identical? I want to find out before we find another edge case that these 'grace pixels' *don't* solve. To be continued…

AaronVanGeffen:
I've rebased this on top of current develop, and cleaned up a little. Not entirely satisfied, but it will do.

Currently, this includes work from #11152. I will rebase this PR after that's been merged, but in the mean time, I would like to get some feedback from testing. @Umdlye, could you confirm things still work on your end?

Umdlye:
Looks good:
openrct2_2020-03-30_22-06-36

AaronVanGeffen:
Squashed the lot into three atomic commits (previously nine, including two from #11152).

I believe this is now ready for review.

PR #10957, opened 2020-03-17
Fix #10623: AppImages are broken

### LINKED ISSUE #10623: AppImages are broken
Running the latest built AppImages on a system that does not have all the OpenRCT2 dependencies installed will result in an error such as: `error while loading shared libraries: libzip.so.4`.

This is caused by a bug in patchelf which is used internally by linuxdeploy to build the AppImages.

https://github.com/linuxdeploy/linuxdeploy/issues/120

IntelOrca:
Try setting rpath on binary during compile, then the appimage creator doesn't have to patch the rpath which inflates the filesize.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,9 +171,7 @@ jobs:
         run: . scripts/setenv -q && get-discord-rpc
       - name: Build OpenRCT2
         shell: bash
-        env:
-          DESTDIR: AppDir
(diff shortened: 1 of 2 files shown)


Gymnasiast:
Something for the changelog?

IntelOrca:
The last release is alright, it got broken only between releases.

IntelOrca:
I was going to change the message to the title of this PR but it has already been merged now.

PR #10958, opened 2020-03-17
Fix regression in boat hire pathfinding

### LINKED ISSUE #10902: Resetting ride let guest out in the air
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 8737d3c

Guest are exited with incorrect position when resetting rides by double clicking red flag/sign.

**Steps to reproduce:**
1. Build any ride
2. Let guest ride it.
3. Reset it by double clicking red signal button/red flag button
4. Guests are spawned/exited in the air.

**Screenshots / Video:**

AaronVanGeffen:


AaronVanGeffen:
Looks like this isn't enough, after all.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -4580,7 +4580,7 @@ static void vehicle_update_boat_location(Vehicle* vehicle)
 
     CoordsXY location = CoordsXY{ vehicle->x, vehicle->y } + CoordsDirectionDelta[returnDirection];
 
-    if (location == returnPosition.ToCoordsXY().ToTileStart())
+    if (location == returnPosition.ToCoordsXY())
```
duncanspumpkin: Location is the one that needs to be to tile start

AaronVanGeffen:
@duncanspumpkin This still doesn't appear to resolve the guests falling into the water. Different bug?

Gymnasiast:
@tupaschoal Yes: changing the coordinates from small to big.

PR #10963, opened 2020-03-18
Fix lightFX being off center in some rotations

WantDiscussion:
Fix lightFX being off center in some rotations

### DIFF
--- a/src/openrct2/drawing/LightFX.cpp
+++ b/src/openrct2/drawing/LightFX.cpp
@@ -674,30 +674,30 @@ void lightfx_add_3d_light(uint32_t lightID, uint16_t lightIDqualifier, int16_t x
 void lightfx_add_3d_light_magic_from_drawing_tile(
     const CoordsXY& mapPosition, int16_t offsetX, int16_t offsetY, int16_t offsetZ, uint8_t lightType)
 {
-    int16_t x = mapPosition.x + offsetX;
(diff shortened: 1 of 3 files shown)


src/openrct2/drawing/LightFX.cpp
```diff
@@ -674,30 +674,30 @@ void lightfx_add_3d_light(uint32_t lightID, uint16_t lightIDqualifier, int16_t x
...
-            x += 16;
-            y -= 16;
-            break;
-        default:
-            return;
-    }
+    int16_t x = mapPosition.x + offsetX + 16;
+    int16_t y = mapPosition.y + offsetY + 16;
+
+    // switch (get_current_rotation())
```
Gymnasiast: Never comment out code. Remove it entirely. We have source control if we ever want it back.

Gymnasiast:
Could you post a before and after screenshot?

WantDiscussion:
> 
> 
> Could you post a before and after screenshot?

Before:
Annotation 2020-03-19 001035

After:
Annotation 2020-03-19 001125

Gymnasiast:
Very good!

Could you also add an entry to the changelog?
`- Fix: [#10963] Light effects are drawn off-centre in some rotations.`

duncanspumpkin:
Yeah I noticed this before but guessed it was there for a reason.

PR #10967, opened 2020-03-18
Refactor the ride type to entry map

IntelOrca:
Use C++ vectors and allow larger ride entry indices (> 256)

### DIFF
--- a/src/openrct2-ui/windows/NewRide.cpp
+++ b/src/openrct2-ui/windows/NewRide.cpp
@@ -23,6 +23,7 @@
 #include <openrct2/management/Research.h>
 #include <openrct2/network/network.h>
 #include <openrct2/object/ObjectLimits.h>
+#include <openrct2/object/ObjectManager.h>
 #include <openrct2/rct1/RCT1.h>
 #include <openrct2/ride/RideData.h>
 #include <openrct2/ride/RideGroupManager.h>
@@ -212,7 +213,7 @@ static void window_new_ride_scrollmouseover(rct_window *w, int32_t scrollIndex, 
 static void window_new_ride_invalidate(rct_window *w);
 static void window_new_ride_paint(rct_window *w, rct_drawpixelinfo *dpi);
 static void window_new_ride_scrollpaint(rct_window *w, rct_drawpixelinfo *dpi, int32_t scrollIndex);
-static void window_new_ride_list_vehicles_for(const uint8_t rideType, const rct_ride_entry * rideEntry, char * out);
+static void window_new_ride_list_vehicles_for(uint8_t rideType, const rct_ride_entry* rideEntry, char* buffer, size_t bufferLen);
 
 // 0x0098E354
 static rct_window_event_list window_new_ride_events = {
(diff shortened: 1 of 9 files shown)


src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -212,7 +213,7 @@ static void window_new_ride_scrollmouseover(rct_window *w, int32_t scrollIndex,
 static void window_new_ride_invalidate(rct_window *w);
 static void window_new_ride_paint(rct_window *w, rct_drawpixelinfo *dpi);
 static void window_new_ride_scrollpaint(rct_window *w, rct_drawpixelinfo *dpi, int32_t scrollIndex);
-static void window_new_ride_list_vehicles_for(const uint8_t rideType, const rct_ride_entry * rideEntry, char * out);
+static void window_new_ride_list_vehicles_for(uint8_t rideType, const rct_ride_entry* rideEntry, char* buffer, size_t bufferLen);
```
tupaschoal: Do we need to drop the `const` on `rideType`? Should we?

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -346,15 +347,15 @@ static ride_list_item* window_new_ride_iterate_over_ride_group(
...
     bool allowDrawingOverLastButton = false;
-    uint8_t* rideEntryIndexPtr = get_ride_entry_indices_for_ride_type(rideType);
 
     char preferredVehicleName[DAT_NAME_LENGTH + 1];
     safe_strcpy(preferredVehicleName, "        ", sizeof(preferredVehicleName));
 
     // For each ride entry for this ride type
-    while (*rideEntryIndexPtr != RIDE_ENTRY_INDEX_NULL)
+    auto& objManager = OpenRCT2::GetContext()->GetObjectManager();
+    auto rideEntries = objManager.GetAllRideEntries(rideType);
```
tupaschoal: `const auto&` as the method returns a reference. Applies to all other calls

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -1056,27 +1057,20 @@ static void window_new_ride_select(rct_window* w)
...
 }
 
-static void window_new_ride_list_vehicles_for(const uint8_t rideType, const rct_ride_entry* rideEntry, char* out)
+static void window_new_ride_list_vehicles_for(uint8_t rideType, const rct_ride_entry* rideEntry, char* buffer, size_t bufferLen)
 {
-    rct_ride_entry* currentRideEntry;
-    const RideGroup *rideGroup, *currentRideGroup;
-    int32_t rideEntryIndex;
-
+    std::fill_n(buffer, bufferLen, 0);
```
tupaschoal: Do we need to?

src/openrct2-ui/windows/NewRide.cpp
IntelOrca: Thanks, very easy to make this mistake.

tupaschoal:
PR doesn't include the fix for the `auto &` yet

src/openrct2-ui/windows/NewRide.cpp
IntelOrca: I decided the code was simpler doing this at the beginning than first checking for a buffer length >= 0 before putting a null terminator on the first return.

This way is just simpler and safer. The routine isn't called often so it isn't much of a performance issue.

src/openrct2-ui/windows/NewRide.cpp
IntelOrca: We rarely make copy-by-value parameters immutable, it would add so much code bloat if we applied this to everything. In this case I just felt it was out of place, inconsistent with everything else here.

IntelOrca:
Pushed changes now.

PR #10970, opened 2020-03-19
Light Effects for rides reintegrated with less lag and option to turn off.

WantDiscussion:
Light effects now only adds lights for vehicles on screen. Only minor lag when many vehicles are on screen. Added option to turn on/off ride lights.

Before:
Annotation 2020-03-19 110021

After:
Annotation 2020-03-19 105916

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3675,7 +3675,7 @@ STR_6358    :Page {UINT16}
 STR_6359    :{POP16}{POP16}Page {UINT16}
 STR_6360    :{SMALLFONT}{BLACK}{COMMA32}
 STR_6361    :Enable lighting effects on rides (experimental)
-STR_6362    :{SMALLFONT}{BLACK} Rides will be lit up at night.
+STR_6362    :{SMALLFONT}{BLACK}If enabled, vehicles for tracked rides will be lit up at night.
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3674,6 +3674,8 @@ STR_6357    :{SMALLFONT}{BLACK}Removes all ducks from the map
 STR_6358    :Page {UINT16}
 STR_6359    :{POP16}{POP16}Page {UINT16}
 STR_6360    :{SMALLFONT}{BLACK}{COMMA32}
+STR_6361    :Enable lighting effects on rides (experimental)
+STR_6362    :{SMALLFONT}{BLACK} Rides will be lit up at night.
 
 #############
 # Scenarios #
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -41,7 +41,7 @@
 - Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
 - Improved: [#10858] Added horizontal grid lines to finance charts.
 - Improved: [#10884] Added y-axes and labels to park window charts.
-- Improved: [#10970] Included vehicle light effect, added option to turn them on/off.
+- Improved: [#10970] Introduced optional light effects for vehicles at night.
 - Removed: [#6898] LOADMM and LOADRCT1 title sequence commands (use LOADSC instead).
 
 0.2.4 (2019-10-28)
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -41,7 +41,7 @@
 - Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
 - Improved: [#10858] Added horizontal grid lines to finance charts.
 - Improved: [#10884] Added y-axes and labels to park window charts.
-- Improved: [#10970] Included vehicle light effect lag and option to turn on/off vehicle light effects.
+- Improved: [#10970] Included vehicle light effect, added option to turn them on/off.
 - Removed: [#6898] LOADMM and LOADRCT1 title sequence commands (use LOADSC instead).
 
 0.2.4 (2019-10-28)
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -41,6 +41,7 @@
 - Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
 - Improved: [#10858] Added horizontal grid lines to finance charts.
 - Improved: [#10884] Added y-axes and labels to park window charts.
+- Improved: [#10970] Included vehicle light effect lag and option to turn on/off vehicle light effects.
 - Removed: [#6898] LOADMM and LOADRCT1 title sequence commands (use LOADSC instead).
 
 0.2.4 (2019-10-28)
--- a/src/openrct2-ui/windows/Options.cpp
+++ b/src/openrct2-ui/windows/Options.cpp
@@ -263,7 +263,7 @@ static rct_widget window_options_rendering_widgets[] = {
     { WWT_GROUPBOX,         1,  5,      304,    FRAME_EFFECTS_START + 0,        FRAME_EFFECTS_START + 93,       STR_EFFECTS_GROUP,              STR_NONE },                             // Rendering group
     { WWT_CHECKBOX,         1,  10,     290,    FRAME_EFFECTS_START + 15,       FRAME_EFFECTS_START + 26,       STR_CYCLE_DAY_NIGHT,            STR_CYCLE_DAY_NIGHT_TIP },              // Cycle day-night
     { WWT_CHECKBOX,         1,  25,     290,    FRAME_EFFECTS_START + 30,       FRAME_EFFECTS_START + 41,       STR_ENABLE_LIGHTING_EFFECTS,    STR_ENABLE_LIGHTING_EFFECTS_TIP },      // Enable light fx
-    { WWT_CHECKBOX,         1,  30,     290,    FRAME_EFFECTS_START + 45,       FRAME_EFFECTS_START + 56,       STR_ENABLE_LIGHTING_VEHICLES,   STR_ENABLE_LIGHTING_VEHICLES_TIP },     // Enable light fx for vehicles
+    { WWT_CHECKBOX,         1,  40,     290,    FRAME_EFFECTS_START + 45,       FRAME_EFFECTS_START + 56,       STR_ENABLE_LIGHTING_VEHICLES,   STR_ENABLE_LIGHTING_VEHICLES_TIP },     // Enable light fx for vehicles
     { WWT_CHECKBOX,         1,  10,     290,    FRAME_EFFECTS_START + 60,       FRAME_EFFECTS_START + 71,       STR_RENDER_WEATHER_EFFECTS,     STR_RENDER_WEATHER_EFFECTS_TIP },       // Render weather effects
     { WWT_CHECKBOX,         1,  25,     290,    FRAME_EFFECTS_START + 75,       FRAME_EFFECTS_START + 86,       STR_DISABLE_LIGHTNING_EFFECT,   STR_DISABLE_LIGHTNING_EFFECT_TIP },     // Disable lightning effect
 #undef FRAME_EFFECTS_START
--- a/src/openrct2-ui/windows/Options.cpp
+++ b/src/openrct2-ui/windows/Options.cpp
@@ -1778,9 +1778,9 @@ static void window_options_invalidate(rct_window* w)
                 w->disabled_widgets |= (1 << WIDX_ENABLE_LIGHT_FX_CHECKBOX);
             }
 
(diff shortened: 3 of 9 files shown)


AaronVanGeffen:
Thanks for working on this.

It's not necessary to close pull requests and open a new one whenever you make a change. GitHub will update your PR whenever you push.

src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -17,6 +17,7 @@
 #include "../../ride/VehiclePaint.h"
 #include "../../world/Sprite.h"
 #include "../Paint.h"
+#include <openrct2\drawing\LightFX.h>
```
AaronVanGeffen:
```suggestion
#include <openrct2/drawing/LightFX.h>
```

src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -111,6 +112,12 @@ void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t
         {
             case SPRITE_IDENTIFIER_VEHICLE:
                 vehicle_paint(session, (Vehicle*)spr, image_direction);
+                #ifdef __ENABLE_LIGHTFX__
+                                if (lightfx_for_vehicles_is_available())
+                                {
+                                    lightfx_add_lights_magic_vehicles((Vehicle*)spr, sprite_idx);
+                                }
+                #endif
```
AaronVanGeffen: I believe the indentation here is incorrect. Are you using clang-format?

src/openrct2-ui/windows/Options.cpp
```diff
@@ -105,6 +105,7 @@ enum WINDOW_OPTIONS_WIDGET_IDX {
     WIDX_EFFECTS_GROUP,
     WIDX_DAY_NIGHT_CHECKBOX,
     WIDX_ENABLE_LIGHT_FX_CHECKBOX,
+    WIDX_ENABLE_LIGHT_FX_FOR_VEHICLES_CHECKBOX,
```
tupaschoal: Indentation looks incorrect

src/openrct2-ui/windows/Options.cpp
```diff
@@ -546,6 +548,7 @@ static uint64_t window_options_page_enabled_widgets[] = {
     (1 << WIDX_VIRTUAL_FLOOR_DROPDOWN) |
     (1 << WIDX_DAY_NIGHT_CHECKBOX) |
     (1 << WIDX_ENABLE_LIGHT_FX_CHECKBOX) |
+    (1 << WIDX_ENABLE_LIGHT_FX_FOR_VEHICLES_CHECKBOX) |
```
tupaschoal: Indentation looks incorrect

src/openrct2/config/Config.cpp
```diff
@@ -181,8 +181,8 @@ namespace Config
 
             // Default config setting is false until the games canvas can be separated from the effect
             model->day_night_cycle = reader->GetBoolean("day_night_cycle", false);
-
```
tupaschoal: Revert this change, as it has nothing to do with the PR

tupaschoal:
I also would appreciate gifs of before/after the change so we have a better idea of things in motion.

WantDiscussion:
> 
> 
> I also would appreciate gifs of before/after the change so we have a better idea of things in motion.

https://imgur.com/a/uJMyMUQ

duncanspumpkin:
Is additional options required? Would anyone actually want the vehicle lights to be off?

pizza2004:
I don't think the additional option is necessary personally.  Most people won't even play with the day/night cycle (especially given that it has nothing to do with the actual days in game), let alone the light effects.  This seems like such a tiny preference one way or the other that if it doesn't work well for you as far as FPS goes you probably just shouldn't be using the light effects at all.

WantDiscussion:
> 
> 
> Is additional options required? Would anyone actually want the vehicle lights to be off?

There is a bit of lag when using uncapped fps so I thought I'd add the option.

src/openrct2-ui/windows/Options.cpp
WantDiscussion: It looks like 4 spaces to me. If there is an issue I can't see it.

Gymnasiast:
Hm, I'm not sure why the vehicles should give off light. That's not typically what happens at night - vehicles stay dark.

AaronVanGeffen:
I think it helps see where vehicles are in the dark, and on top of that, it's very aesthetically pleasing.

WantDiscussion:
> 
> 
> Hm, I'm not sure why the vehicles should give off light. That's not typically what happens at night - vehicles stay dark.

Well transports rides would have interior lights, and possibly the mini cars. But true the go karts and boats probably wouldn't.

IntelOrca:
If you look at the original PR for the lights, the steam trains gave off light and it looked really good in the dark fog. I am not sure how you could decide what vehicles should give off light or not - but anything to brighen the game helps.

AaronVanGeffen:
What backend are you all using to test the lighting effects?

On Linux, I get screen/drawing invalidation problems as soon as I enable the lighting effects with the 'software (hardware passthrough)' backend. Having just the day/night cycle on does not present these problems. (This happens on develop as well, fwiw.)

src/openrct2/drawing/LightFX.cpp
```diff
@@ -676,7 +681,35 @@ void lightfx_add_3d_light_magic_from_drawing_tile(
 {
     int16_t x = mapPosition.x + offsetX + 16;
     int16_t y = mapPosition.y + offsetY + 16;
-
+<<<<<<< HEAD
```
duncanspumpkin: Looks like you had a rebase conflict.

src/openrct2/drawing/LightFX.cpp
```diff
@@ -685,20 +718,16 @@ uint32_t lightfx_get_light_polution()
     return _lightPolution_front;
 }
 
-void lightfx_add_lights_magic_vehicles()
+void lightfx_add_lights_magic_vehicles(const Vehicle* vehicle, uint16_t spriteIndex)
```
duncanspumpkin: spriteIndex is basically unused. Just remove it completely this is not required any more.

duncanspumpkin: This is no longer putting the lights on all vehicles but on just a singular vehicle change the name of the function.

src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -111,6 +112,12 @@ void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t
         {
             case SPRITE_IDENTIFIER_VEHICLE:
                 vehicle_paint(session, (Vehicle*)spr, image_direction);
+#ifdef __ENABLE_LIGHTFX__
```
duncanspumpkin: The below functions should only be called on Head vehicles. The function is assuming it is only passed heads.

src/openrct2/drawing/LightFX.cpp
```diff
@@ -685,20 +718,16 @@ uint32_t lightfx_get_light_polution()
...
 {
-    uint16_t spriteIndex = gSpriteListHead[SPRITE_LIST_VEHICLE_HEAD];
-    while (spriteIndex != SPRITE_INDEX_NULL)
+    if (spriteIndex != SPRITE_INDEX_NULL)
```
duncanspumpkin: remove this if. spriteIndex will never be null nor will vehicle be nullptr.

src/openrct2/drawing/LightFX.cpp
```diff
@@ -685,20 +718,16 @@ uint32_t lightfx_get_light_polution()
...
     {
-        Vehicle* vehicle = &(get_sprite(spriteIndex)->vehicle);
         uint16_t vehicleID = spriteIndex;
-        spriteIndex = vehicle->next;
-
-        Vehicle* mother_vehicle = vehicle;
+        const Vehicle* mother_vehicle = vehicle;
```
duncanspumpkin: this variable is only used for the below comparison remove the new variable and just use it directly.

src/openrct2/paint/sprite/Paint.Sprite.cpp
duncanspumpkin: On second thoughts that wouldn't work ignore the above comment.

src/openrct2/drawing/LightFX.cpp
```diff
@@ -685,20 +718,16 @@ uint32_t lightfx_get_light_polution()
...
 
         if (mother_vehicle->ride_subtype == RIDE_ENTRY_INDEX_NULL)
         {
-            continue;
+            return;
         }
 
         for (uint16_t q = vehicleID; q != SPRITE_INDEX_NULL;)
```
duncanspumpkin: This for loop is very odd. I don't think it is doing what it looks like it is doing.

src/openrct2/paint/sprite/Paint.Sprite.cpp
duncanspumpkin: On third thoughts this is a mess and I am no longer sure what it is meant to do.

distribution/changelog.txt
```diff
@@ -41,6 +41,7 @@
 - Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
 - Improved: [#10858] Added horizontal grid lines to finance charts.
 - Improved: [#10884] Added y-axes and labels to park window charts.
+- Improved: [#10970] Included vehicle light effect lag and option to turn on/off vehicle light effects.
```
AaronVanGeffen: Included vehicle light effect lag? That doesn't make sense.
```suggestion
- Improved: [#10970] Introduced optional light effects for vehicles at night.
```

src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -111,6 +112,12 @@ void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t
...
+                if (lightfx_for_vehicles_is_available())
+                {
+                    lightfx_add_lights_magic_vehicle((Vehicle*)spr);
```
tupaschoal: Please use a named cast (`static_cast`, `dynamic_cast`, ...) instead of c-style cast. In this case looks like static would do

src/openrct2/paint/sprite/Paint.Sprite.cpp
tupaschoal: (I'm marking this as resolved then Duncan)

data/language/en-GB.txt
```diff
@@ -3674,6 +3674,8 @@ STR_6357    :{SMALLFONT}{BLACK}Removes all ducks from the map
 STR_6358    :Page {UINT16}
 STR_6359    :{POP16}{POP16}Page {UINT16}
 STR_6360    :{SMALLFONT}{BLACK}{COMMA32}
+STR_6361    :Enable lighting effects on rides (experimental)
+STR_6362    :{SMALLFONT}{BLACK} Rides will be lit up at night.
```
AaronVanGeffen: I hate to keep moving the goal posts, but the space here is unnecessary. Perhaps it would best be replaced with something more descriptive, anyway:
```suggestion
STR_6362    :{SMALLFONT}{BLACK}If enabled, vehicles for tracked rides will be lit up at night.
```

AaronVanGeffen:
An idea for a future PR: add light sources to the shops/stalls, which are currently unlit.

src/openrct2/paint/sprite/Paint.Sprite.cpp
WantDiscussion: static_cast<Vehicle*>(spr);

gives an "invalid type conversion" error

ocalhoun6:
> Hm, I'm not sure why the vehicles should give off light. That's not typically what happens at night - vehicles stay dark.

Some certainly would, like miniature railroad trains, streamlined monorail trains, and lift cabins. Heck, for that matter, a lot of flat rides might be lit up like christmas trees:
photo-1523113501336-6ea4a7d6660d



> 
> 
> An idea for a future PR: add light sources to the shops/stalls, which are currently unlit.

Come to think of it, wouldn't most rides have their entire station platform lit up at night? For safety reasons if nothing else.

Gymnasiast:
@ocalhoun6 On funfairs, yes. In amusement parks: not so much.

src/openrct2/paint/sprite/Paint.Sprite.cpp
WantDiscussion: I changed it to

`lightfx_add_lights_magic_vehicle(reinterpret_cast<Vehicle*>(const_cast<rct_sprite*>(spr)));`

hope that's right

tupaschoal: I think it's fine for now, thanks

PR #10977, opened 2020-03-19
Test build with updated mingw docker image

janisozaur:


janisozaur:
It seems the way I created the Dockerfile prevents GitHub actions from working. I suspect this is a problem regarding switching users inside container. I wish a bit this stayed on Arch, but I strongly consider switching it to Fedora instead. We use Fedora-based mingw image already in OpenLoco, they provide a good choice of prebuilt mingw packages and like our other images it is not a rolling release, which makes some sense for CI to use.

janisozaur:
Replaced with a new docker image: https://github.com/OpenRCT2/openrct2-docker-build/pull/1

PR #6440, opened 2017-10-11
Initial prototype of A* for the heuristic search.

zaxcav:
First signs of life from the A* heuristic search I've been working on.

The A* code can either be enabled for all peeps (with significant performance degradation in parks with lots of peeps) or only for selected peeps.

In Peep.cpp the variable gAStarEnableAll (shortly after the #includes at the top of the file) controls this.
The A* code is currently enabled for all peeps be default.

When the A* code is enabled for only selected peeps, peeps are selected as follows:
- guests use A* when they are being tracked;
- mechanics use A* when they are named "Mechanic Star".

The A* code is substantially simpler than the existing heuristic search despite implementing similar specific handling of overlaid paths (zero clearance) to cater for really creative builders.

I've done rudimentary testing using save games for all of the path finding issues I've fixed over the last months, and the A* code works at least as well as the existing heuristic and lots better where the wide paths are causing "unnatural" behavior in the existing heuristic.

The various A* lists (open, closed,etc) are currently a hack with O(n) performance for search operations, so replacing these is absolutely necessary before this can even be remotely considered for merging.

Next steps:
- ~~rebase to the current development branch~~ done
- refactor/clean up the current A* logic
- implement a more efficient data structure for the A* lists
- start experimenting with fun & interesting stuff like using transport rides, letting the search leave paths (no more catastrophic path finding failures from disconnected ride entrances/exits, no more need for separate "surface finding" vs "path finding" vs all other "* finding")
- tuning: appropriate search limits?
- optimisation, incl. save state.

PR #3767, opened 2016-05-29
Optional expanded peep behaviour for walking and queuing

JeroenDStout:
I have created a branch in which I made some modifications to peep walking (not AI itself), which are exposed through (for now) five hidden settings in config.ini; they are off by default.

The idea of this alternate behaviour is to move peeps more off-grid, both for walking and queuing, and to add a little bit of extra character to them. Please note what I am showing here is not meant to be final. I would like to hear your thoughts on it!

messy time

Just to emphasize: _all_ of these options are written to have _no_ impact on path-finding or other higher AI. My main idea is to allow a player to _opt in_ to additional peep behaviour.

There are five changes:
- `peep_allow_sidestepping` (boolean)

Currently peeps walk straight lines and diagonals. This option allows them to 'side-step' by a pixel, meaning they have more freedom in walking lines. They also have a 'preferred' direction, meaning they are less inclined to alternate constantly between directions.
- `guest_messy_queuing` (boolean)

Currently peeps queue in perfect straight lines. This options allows them to queue messily, occasionally side by side. They try to keep a distance from the person ahead of them, which they prefer to be large; but if they are in a full, slow queue they succumb to pressure and start packing closely.

You can at a glance tell the difference between a ride with a high through-put and a ride which has more people queuing than places:

messy queuing fast
messy queuing busy

People (as is seen here) also skip around corners and queue on diagonals if the paths are quickly winding. Because of their dynamic preferences they constantly adjust the queue. In my experience this is what the rear of a queue experiences in a theme park; a constant slightly movement, whereas the head of the queue has great speed-ups and slow-downs.
- `peep_messy_walking` (boolean)

Peeps currently walk to the centre of a tile, but have a randomised threshold for getting there. This gives them some way of walking off-centre, but they frequently correct through (flickering) diagonals. The new option makes peeps walk the least possible to get to the new tile, meaning they always try to cut corners. They are also programmed to not take a turn instantly and instead incorporate a tiny soft turn.

messy walking

Especially when looking at a main square, they move less like cars in a grid-based city, and more like strollers in a park. When very nauseous they walk slightly less straight lines.
- `peep_messy_congestion` (boolean)

Currently peeps base walking pace only on their state. This option enables peeps who get congested (many peeps on one tile) to slow down. This makes avoiding tile-wide bottlenecks more interesting and shows bad design choices very quickly.
- `guest_max_time_in_park` (in minutes)

[...]

PR #10982, opened 2020-03-20
Improvement: Option for Guests to use more of wide path

WantDiscussion:
Improvement: Option for Guests to use more of wide path

-Instead of only one wide flag, each path can have multiple levels of flags

-The first level runs the basic algorithm 8 times (or "groups") in
multiple directions for peeps to use paths symmetrically.

-Levels 2 and above each has 8 groups coresponding to the previous flags
to fill in the remaining spaces (Still incomplete)

-When peeps are path finding, the flag they consider when checking if a
path is wide is based on their ID. The original wide flag still keeps
track of the default update order. Staff peeps use the default wide
Flag.

-Each group and level of flags can be viewed with the debugger.

-Flags are stored in a array of vectors separate from the
tile structures

-In place of saving all flags, when loading a map the game will instead
calculate all wide flags when configured to use 1 or more levels. This means
the game state is not perfectly preserved upon saving when this feature
is on.

Future Improvements:
-Improve algorithm for widepath levels 2 and up. Some gaps are not yet covered.


src/openrct2/world/TileElement.h
```diff
@@ -272,7 +277,7 @@ struct PathElement : TileElementBase
 
     bool IsLevelCrossing(const CoordsXY& coords) const;
 };
-assert_struct_size(PathElement, 16);
+assert_struct_size(PathElement, 17);
```
janisozaur: I doubt this will pass. These asserts are here to maintain SV6 compatibility. You can see a number of recent contributions to see how to deal with that: read old structs from files, use them to set new ones.

duncanspumpkin: Its not to maintain SV6 compatibility as that was 8 bytes. These asserts are because we don't want this struct to exceed 16 bytes. So its still the same outcome in that we can't make this change.

src/openrct2/peep/Staff.cpp
```diff
@@ -293,7 +293,6 @@ bool staff_can_ignore_wide_flag(Peep* staff, int32_t x, int32_t y, uint8_t z, Ti
                 pathfound = true;
                 pathcount++;
             }
-
```
tupaschoal: This is unrelated, so please revert.

src/openrct2/world/TileElement.h
```diff
@@ -199,15 +199,17 @@ struct PathElement : TileElementBase
...
     uint8_t Flags2;         // 9
     uint8_t SlopeDirection; // 10
+
     union
     {
         uint8_t AdditionStatus; // 11, only used for litter bins
         ride_id_t rideIndex;    // 11
     };
     ::StationIndex StationIndex; // 13
+    uint8_t isWideFlags;         // 14
```
tupaschoal:
```suggestion
    uint8_t WideFlags;         // 14
```
I think the "is" should be dropped. "is..." variables usually imply boolean value.

src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -850,7 +850,7 @@ void path_paint(paint_session* session, uint16_t height, const TileElement* tile
     }
 
     // Draw wide flags as ghosts, leaving only the "walkable" paths to be drawn normally
-    if (gPaintWidePathsAsGhost && tile_element->AsPath()->IsWide())
+    if (gPaintWidePathsAsGhost && tile_element->AsPath()->IsWideForGroup(gDateMonthTicks / 8192))
```
tupaschoal: What's with this division?

src/openrct2/world/Footpath.cpp
```diff
@@ -1630,7 +1647,11 @@ static void footpath_clear_wide(const CoordsXY& footpathPos)
     {
         if (tileElement->GetType() != TILE_ELEMENT_TYPE_PATH)
             continue;
-        tileElement->AsPath()->SetWide(false);
+        tileElement->AsPath()->SetWideForGroup(wideGroup, false);
+
+        if (wideGroup == 0 && !(tileElement->AsPath()->IsWideForGroup(0)))
```
tupaschoal:
```suggestion
        if (wideGroup == 0 && !tileElement->AsPath()->IsWideForGroup(wideGroup))
```

AaronVanGeffen:
Hey @WantDiscussion,

Thank you for your working on improving the pathfinding. We know the current pathing algorithm is suboptimal. Fortunately, various contributors have worked hard to improve these algorithms before, from ironing out obvious bugs to documenting existing behaviour. This has made them better than vanilla RCT2. However, there is certainly room for improvement.

At present, we are only considering improvements to the pathfinding algorithm that are bugfixes. By this I mean changes that are relatively self-contained. I speak for the team when I say that, unfortunately, that means we have decided not accept this PR at this time.

I am very sorry to break this news to you, because you have obviously been working very hard on this PR.

The next step on our roadmap for the peep pathfinding is to make the entire system pluggable. In this way, players could choose to experience the vanilla pathfinding algorithm, or an entirely different algorithm. Such an algorithm could trade more optimal pathing for more intensive CPU usage, for example.

This is quite an undertaking, however, and work on this is on hold as other changes are prioritised (the new save format and plug-in engine, for example). However, until this is done, we will reject PRs like #3767, #6440, and this one.

If you have smaller, self-contained changes to improve the pathing algorithm, I would like to encourage you to extract them into a new PR. However, please keep the above in mind. In general, it is good practice to ask around on Discord or Gitter before undertaking big changes.

Finally, I would like to point out that this decision is not meant, in any way, to reflect your coding abilities. Indeed, it is obvious that you worked very hard and enthusiastically on tackling this. I hope this decision will not discourage you from future contributions.

WantDiscussion:
Thank you for your response.

PR #10984, opened 2020-03-20
Convert objects limits to constexpr

Gymnasiast:
Also backported from #10664.

### DIFF
--- a/src/openrct2/object/ObjectLimits.h
+++ b/src/openrct2/object/ObjectLimits.h
@@ -11,21 +11,21 @@
 
 #include <cstdint>
 
-#define MAX_RIDE_OBJECTS 128
-#define MAX_SMALL_SCENERY_OBJECTS 252
-#define MAX_LARGE_SCENERY_OBJECTS 128
-#define MAX_WALL_SCENERY_OBJECTS 128
-#define MAX_BANNER_OBJECTS 32
-#define MAX_PATH_OBJECTS 16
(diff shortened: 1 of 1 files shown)


src/openrct2/object/ObjectLimits.h
```diff
@@ -11,21 +11,21 @@
...
-#define MAX_PATH_ADDITION_OBJECTS 15
-#define MAX_SCENERY_GROUP_OBJECTS 19
-#define MAX_PARK_ENTRANCE_OBJECTS 1
-#define MAX_WATER_OBJECTS 1
-#define MAX_SCENARIO_TEXT_OBJECTS 1
-#define MAX_TERRAIN_SURFACE_OBJECTS 14
-#define MAX_TERRAIN_EDGE_OBJECTS 255
-#define MAX_STATION_OBJECTS 255
-#define MAX_MUSIC_OBJECTS 0
+constexpr const uint16_t MAX_RIDE_OBJECTS = 128;
```
AaronVanGeffen: A `constexpr` is already `const` by definition, isn't it?

Gymnasiast: Don't know, but our codebase seems to use `constexpr const` in 95% of the cases.

tupaschoal: Yeah, there are a few corner cases where the implied constness is not where you would expect, but in these cases the second const is redundant

Issue #10631, opened 2020-01-28
Game bugs out and crashes if you get too many stations via copying stations with the tile inspector


LordMarcel:
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 550da84

You can copy a station with the tile inspector so that you have two station 1's, or two station 3's. When you right click on the copied station and go into build mode, the game makes that station its own station, giving it a number that hasn't been used yet. If you already have four stations there are no unused station numbers, so the game will bug out causing the colors of the ride to change and a whole load of station numbers to be added to the ride. You can't always manage to see that as the game now also has a high chance of crashing whenever you click on the dropdown menu, which is what happened in this video: https://www.youtube.com/watch?v=fLppoAehtXs

Here is a screenshot of the ride having 126 stations after that happened:
![image](https://user-images.githubusercontent.com/25930360/73313142-b589fb80-422a-11ea-9e8a-f9bb891ebb05.png)

Gymnasiast:
Dropdowns can only show 128 options. Most likely, the station tiles it places have index 255 ("-1").

PR #10986, opened 2020-03-20
Allow up to 512 items per dropdown

Gymnasiast:
Also backported from #10664.

### DIFF
--- a/src/openrct2-ui/windows/Dropdown.cpp
+++ b/src/openrct2-ui/windows/Dropdown.cpp
@@ -45,8 +45,8 @@ static bool _dropdown_list_vertically;
 int32_t gDropdownNumItems;
 rct_string_id gDropdownItemsFormat[DROPDOWN_ITEMS_MAX_SIZE];
 int64_t gDropdownItemsArgs[DROPDOWN_ITEMS_MAX_SIZE];
-static bool _dropdownItemsChecked[DROPDOWN_ITEMS_MAX_SIZE];
-static bool _dropdownItemsDisabled[DROPDOWN_ITEMS_MAX_SIZE];
+static std::bitset<DROPDOWN_ITEMS_MAX_SIZE> _dropdownItemsChecked = {};
+static std::bitset<DROPDOWN_ITEMS_MAX_SIZE> _dropdownItemsDisabled = {};
 bool gDropdownIsColour;
 int32_t gDropdownLastColourHover;
 int32_t gDropdownHighlightedIndex;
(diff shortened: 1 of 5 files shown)


src/openrct2-ui/windows/Dropdown.cpp
```diff
@@ -457,9 +457,14 @@ void window_dropdown_show_colour(rct_window* w, rct_widget* widget, uint8_t drop
...
         w->windowPos.x + widget->left, w->windowPos.y + widget->top, widget->bottom - widget->top + 1, dropdownColour,
-        DROPDOWN_FLAG_STAY_OPEN, COLOUR_COUNT, 12, 12, gAppropriateImageDropdownItemsPerRow[COLOUR_COUNT]);
+        DROPDOWN_FLAG_STAY_OPEN, COLOUR_COUNT, 12, 12, _appropriateImageDropdownItemsPerRow[COLOUR_COUNT]);
 
     gDropdownIsColour = true;
     gDropdownLastColourHover = -1;
     gDropdownDefaultIndex = defaultIndex;
 }
+
+uint32_t dropdown_get_appropriate_image_dropdown_items_per_row(uint32_t numItems)
```
duncanspumpkin: Should probably use CamelCase for this new function. Although might not make too much sense since it would be out of place in this file.

Gymnasiast: I don't think we have any C-style function that uses CamelCase - only C++ methods seem to use it.

duncanspumpkin: There are a good bunch scattered throughout the code base. The coding style which we (sometimes) follow assumes all functions are CamelCase.

AaronVanGeffen:
That's one way to fix #10631… :wink:

src/openrct2-ui/windows/Dropdown.cpp
```diff
@@ -22,8 +22,8 @@ constexpr int32_t DROPDOWN_TEXT_MAX_ROWS = 32;
 
 constexpr int32_t DROPDOWN_ITEM_HEIGHT = 12;
 
-int32_t gAppropriateImageDropdownItemsPerRow[] = {
-    1, 1, 1, 1, 2, 2, 3, 3, 4, 3, 5, 4, 4, 5, 5, 5, 4, 5, 6, 5, 5, 7, 4, 5, 6, 5, 6, 6, 6, 6, 6, 8, 8, 0,
+static int32_t _appropriateImageDropdownItemsPerRow[34] = {
```
tupaschoal: Can be made const as well, can't it?

Could probably use an std::array instead, as well.

tupaschoal: Can be made const as well, can't it?

Could probably use an std::array instead, as well.

Gymnasiast: Is there a point to using std::array for a constant array, anyway?

duncanspumpkin: Well you wouldnt need to use `std::size` as you can use `.size()`. I don't really see much point changing it unless you want to iterate over and use the ranged for loop.

AaronVanGeffen: You can use ranged for loops for built-in arrays: https://gcc.godbolt.org/z/TVioqG

No need to use std::array containers when that's your goal.

tupaschoal: I think it makes the code more like C++

tupaschoal: I think it makes the code more like C++

PR #10991, opened 2020-03-21
Fix #10631: Assertion fails when modifying copied station pieces 

### LINKED ISSUE #10631: Game bugs out and crashes if you get too many stations via copying stations with the tile inspector

Gymnasiast:


Gymnasiast:
@AaronVanGeffen The only idea I have is making them `log_verbose`.

AaronVanGeffen:
That might be better, yes. It would hide the error during normal playthroughs, though. But then again, this is an edge case, anyway.

PR #10992, opened 2020-03-21
Use StationIndex in many more places

Gymnasiast:


src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2094,11 +2094,24 @@ int32_t guest_path_finding(Guest* peep)
         int32_t select = peep->no_of_rides % numEntranceStations;
         while (select > 0)
         {
-            closestStationNum = bitscanforward(entranceStations);
-            entranceStations &= ~(1 << closestStationNum);
-            select--;
+            for (size_t i = 0; i < entranceStations.size(); i++)
```
Gymnasiast: This was done because MAX_STATION will be raised to 255, which makes it too large to stuff into a uint8_t or even a uint64_t.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2094,11 +2094,24 @@ int32_t guest_path_finding(Guest* peep)
...
+        for (size_t i = 0; i < entranceStations.size(); i++)
+        {
+            if (entranceStations[i])
+            {
+                closestStationNum = i;
+                break;
+            }
         }
-        closestStationNum = bitscanforward(entranceStations);
     }
```
IntelOrca: I think it would be nice if this whole loop and comment above was moved to a new function.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2094,11 +2094,24 @@ int32_t guest_path_finding(Guest* peep)
...
+                    select--;
+                    break;
+                }
+            }
+        }
+        for (size_t i = 0; i < entranceStations.size(); i++)
+        {
+            if (entranceStations[i])
+            {
+                closestStationNum = i;
```
tupaschoal: This is failing on Windows build, because it's narrowing the value of i.

src/openrct2/ride/Ride.cpp
```diff
@@ -3729,14 +3729,14 @@ money32 set_operating_setting_nested(ride_id_t rideId, RideSetSetting setting, u
  *
  *  rct2: 0x006B4CC1
  */
-static int32_t ride_mode_check_valid_station_numbers(Ride* ride)
+static StationIndex ride_mode_check_valid_station_numbers(Ride* ride)
 {
-    uint8_t no_stations = 0;
-    for (uint8_t station_index = 0; station_index < MAX_STATIONS; ++station_index)
+    uint8_t numStations = 0;
```
tupaschoal: Seems likely to blow up in the future, I'd just use a large primitive instead.
```suggestion
    uint32_t numStations = 0;
```

src/openrct2/peep/Guest.cpp
```diff
@@ -4260,7 +4260,7 @@ void Guest::UpdateRideLeaveVehicle()
 
     TileElement* trackElement = ride_get_station_start_track_element(ride, current_ride_station);
 
-    uint8_t station_direction = (trackElement == nullptr ? 0 : trackElement->GetDirection());
+    Direction station_direction = (trackElement == nullptr ? 0 : trackElement->GetDirection());
```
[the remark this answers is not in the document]
Gymnasiast: A bit. I consider it camp floor cleaning.

src/openrct2/ride/Ride.cpp
Gymnasiast: `uint16_t` should more than enough - that allows for 65535 stations _per ride_. (I'm not sure we're going to increase the maximum amount of stations beyond 255, but we're _certainly_ not going to increase it beyond 65535...)

src/openrct2/peep/Guest.cpp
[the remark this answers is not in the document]
tupaschoal: Ok.

src/openrct2/ride/Ride.cpp
tupaschoal: Alright.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -1839,6 +1839,45 @@ static void get_ride_queue_end(TileCoordsXYZ& loc)
...
+ * different sections of the ride.
+ * In this case, the ride's various entrance stations will typically,
+ * though not necessarily, be adjacent to one another and consequently
+ * not too far for the peep to walk when cycling between them.
+ * Note: the same choice of station must made while the peep navigates
+ * to the station. Consequently a truly random station selection here is not
+ * appropriate.
+ */
+static StationIndex guest_pathfinding_select_random_station(
+    const Guest* guest, int32_t numEntranceStations, std::bitset<MAX_STATIONS> entranceStations, StationIndex closestStationNum)
```
tupaschoal:
```suggestion
    const Guest* guest, int32_t numEntranceStations, std::bitset<MAX_STATIONS>& entranceStations, StationIndex& closestStationNum)
```
Because you're altering `entranceStations` and `closestStationNum`, you should receive a reference to them instead.

src/openrct2/world/Footpath.cpp
```diff
@@ -1502,7 +1502,7 @@ void PathElement::SetIsBlockedByVehicle(bool isBlocked)
     }
 }
 
-uint8_t PathElement::GetStationIndex() const
+::StationIndex PathElement::GetStationIndex() const
```
duncanspumpkin: why the extra `::`

src/openrct2/peep/GuestPathfinding.cpp
Gymnasiast: No.

entranceStations is not modified after the call to `guest_pathfinding_select_random_station()`, so a reference is not necessary.
The new value of `closestStationNum` is _returned_.

src/openrct2/world/Footpath.cpp
Gymnasiast: I believe it's in order to differentiate between the field and the typedef. At least that why I had to do so for `SetStationIndex()` - I assumed that the same would apply for `GetStationIndex()`.

duncanspumpkin: oh yeah right you are.

src/openrct2/peep/GuestPathfinding.cpp
tupaschoal: You're right, it only assigns to `closestStationNum`, but never reads from it inside the method, so you don't even need to send this parameter if you change the last return to `return 0` and the `closestStationNum = i;` to `return i;`. Did I interpret it correctly?

As for `entranceStations`, I get that it is not being read after the function is called, but if it ever gets to, it'll be hard to debug, unless we don't expect the function to alter the outer bitset at all.

Gymnasiast: OK, addressed both points.

PR #11006, opened 2020-03-23
Docs: Add missing directories in readme.md

daandeheij:
I am not sure whether PRs of this nature are appreciated, but I decided to offer this nevertheless:

I noticed recently that the documentation `readme.md` in `/src/openrct2/` is **missing** many directories. I think it would be very helpful to developers new to OpenRCT2 if this specific documentation would be a bit more complete.

Therefore, I have added missing directories, and wrote a short description for each. My apologies if some of the descriptions are not quite right or to the point, I am still new to OpenRCT2 :)


src/openrct2/readme.md
```diff
@@ -46,4 +90,4 @@
 
 - **world**
 
-  World objects and mechanics such as the climate, landscape, sprites and park.
+  World objects and mechanics such as the climate, landscape, sprites and park.
```
janisozaur: You truncated the new line here, please restore.

duncanspumpkin:
This is still not really correct as there is no mention of the fact its split into two main folders `openrct2` and `openrct2-ui`. A few of the existing descriptions are also incorrect as all the ui code is in the ui folder.

Gymnasiast:
> I am not sure whether PRs of this nature are appreciated, but I decided to offer this nevertheless:

They are appreciated. Documentation can frequently take the back burner for many open source projects and ours is no exception. We also try to keep our development as open as possible.

AaronVanGeffen:
@duncanspumpkin This is specifically for the `src/openrct2` directory.

I wasn't aware there was a separate readme in there, to be honest. I'm not convinced there should be, either. Perhaps the wiki is a more appropriate place for this kind of documentation.

duncanspumpkin:
@AaronVanGeffen argh oh yeah. Well its still got the wrong description for windows and interface.

Gymnasiast:
I merged this because we needed a new build at openrct2.org. Since this was an improvement upon the old state already, I figured it would be okay to use this PR for that.

PR #11007, opened 2020-03-23
Fix #11001: Rides list does not use natural sorting

### LINKED ISSUE #11001: Rides list does not use natural sorting
**OS:** Ubuntu 20.04
**Version:** 0.2.4
**Commit/Build:** https://github.com/OpenRCT2/OpenRCT2/commit/2a94632b32564426ebbddff318f3055718266d71

Discovered in a Reddit post and reproduced using above version.

afbeelding

EDIT:
Use the folllowing save file to reproduce/validate:
[sorting_attractions.zip](https://github.com/OpenRCT2/OpenRCT2/files/4371669/sorting_attractions.zip)

tupaschoal:
Closes #11001 

~I'm adding a library called "[alphanum](http://www.davekoelle.com/files/alphanum.hpp)", it is described in detail on [davekoelle's page](http://www.davekoelle.com/alphanum.html), but is summarized as:~

~> People sort strings with numbers differently than software does. Most sorting algorithms compare ASCII values, which produces an ordering that is inconsistent with human logic. Here's how to fix it.~

~I've made just three changes to it:~
~- Use `#pragma once`~
~- Suppress "assignment within conditional expression" warning~
~- Remove a usage of `std::binary_function` which is deprecated in C++17~

image
image
image

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -41,6 +41,7 @@
 - Fix: [#10904] RCT1/LL-scenarios with red water won't open.
 - Fix: [#10941] The Clear Scenery tool gives refunds for ghost elements.
 - Fix: [#10963] Light effects are drawn off-centre in some rotations.
+- Fix: [#11001] Rides list does not use natural sorting
 - Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
 - Improved: [#10858] Added horizontal grid lines to finance charts.
 - Improved: [#10884] Added y-axes and labels to park window charts.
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
We have `strlogicalcmp` in Util.cpp already. I think the extra library is a little overkill.

tupaschoal:
> I think the extra library is a little overkill. We have `strlogicalcmp` in Util.cpp already.

hah, if only I had known that, it makes things a LOT simpler :D, I have just modified to reuse it instead

PR #11008, opened 2020-03-23
Fix #10993: Guest Count Intent Not Listened To

### LINKED ISSUE #10993: Guest Count Intent Not Listened To
The guest count is meant to be redrawn whenever a guest leaves. It looks like when the intent system was implemented this was incorrectly handled. 

Within windowmanager.cpp the intent `INTENT_ACTION_UPDATE_GUEST_COUNT` needs to be handled by invalidating the bottom toolbar (gToolbarDirtyFlags), the guest list and the park window.

daandeheij:
This fixes the bug mentioned in #10993. 

`INTENT_ACTION_UPDATE_GUEST_COUNT` was not handled by `WindowManager.cpp`, causing the guest count not to redraw whenever a guest leaves. 

I just tested and the toolbar, guest list, and park information window all redraw correctly on guest leave.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -41,6 +41,7 @@
 - Fix: [#10904] RCT1/LL-scenarios with red water won't open.
 - Fix: [#10941] The Clear Scenery tool gives refunds for ghost elements.
 - Fix: [#10963] Light effects are drawn off-centre in some rotations.
+- Fix: [#10993] Bottom toolbar not refreshing when a guest leaves the park.
 - Fix: [#11001] Rides list does not use natural sorting.
 - Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
 - Improved: [#10858] Added horizontal grid lines to finance charts.
--- a/src/openrct2-ui/WindowManager.cpp
+++ b/src/openrct2-ui/WindowManager.cpp
@@ -413,6 +413,12 @@ public:
                 window_invalidate_by_class(WC_GUEST_LIST);
                 break;
 
+            case INTENT_ACTION_UPDATE_GUEST_COUNT:
+                gToolbarDirtyFlags |= BTM_TB_DIRTY_FLAG_PEEP_COUNT;
+                window_invalidate_by_class(WC_GUEST_LIST);
+                window_invalidate_by_class(WC_PARK_INFORMATION);
+                break;
+
             case INTENT_ACTION_UPDATE_PARK_RATING:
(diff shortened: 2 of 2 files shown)


Gymnasiast:
It's not necessary for this PR, since we can squash it, but in general: always make sure your PRs are rebased, so no merge commits are present.

daandeheij:
> It's not necessary for this PR, since we can squash it, but in general: always make sure your PRs are rebased, so no merge commits are present.

Yes, thank you! Will make sure to do so from now on. Should you also rebase before adding additional commits to the PR later on? This workflow is new to me.

Additionally, I added the entry to the changelog.

Gymnasiast:
> Should you also rebase before adding additional commits to the PR later on? This workflow is new to me.

No, that's not necessary. We just want a set of clean commits with no merges, that is all.

distribution/changelog.txt
```diff
@@ -1,4 +1,4 @@
-0.2.4+ (in development)
+﻿0.2.4+ (in development)
```
Gymnasiast: It looks like your editor (probably one created by Microsoft) inserted a Byte Order Mark here. Please remove it.

daandeheij:
Evidently, I completely messed up when trying to fix the changelog. I'm not sure how to fix this now. Sorry!

tupaschoal:
Make sure your develop branch is up-to-date, do so by switching to it, doing `git fetch upstream` then `git rebase upstream/develop` (assuming you have upstream configured.

Then switch to the PR branch and do `git rebase -i develop", it should show you just your commits and will rebase your PR. If that succeeds, `git push -f origin` will sync your local branch with the GitHub one.

I didn't take a close look to if that will suffice, but it's worth giving a try.

AaronVanGeffen:
It looks like the mishap was indeed caused by an outdated develop branch. I've rebased your branch for you, in the hope this fix can still make it to the upcoming release.

daandeheij:
Thank you so much for the assistance, I highly appreciate! I'll make sure to do this right in the future.

Gymnasiast:
I have a hard time testing this, because I cannot reproduce the bug on develop in the first place. As such, I'd like @duncanspumpkin to confirm this solves the problem.

Umdlye:
It solves the problem here :D (Win10 x64)
Occurs in develop but not here.

(If it seems to work normally in develop it might be because the guest leaving coincided with something else that updated the bottom toolbar like the date or money changing.)

duncanspumpkin:
I've not tested it but this is doing exactly what i was expecting it to do so on umdlye's confirmation this is good to go.

PR #11010, opened 2020-03-24
Fix: Object selector x button behaviour

WantDiscussion:
Previously in scenario editor and track designer, the X button for the object selector would close the entire scenario editor and track designer, returning the user to the main menu. This is counter intuitive for the x button to close everything down instead of just the window it belongs to. Now it closes the object selector window and advance to the next stage of the scenario editor/track designer

The x button for the object selector in the cheat menu remains the same.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -32,6 +32,7 @@
 - Fix: [#10928] File browser's date column is too narrow.
 - Fix: [#10951, #11160] Attempting to place park entrances creates ghost entrances in random locations.
 - Fix: [#11005] Company value overflows.
+- Fix: [#11010] #6024 Object selector X button closes scenario and track editors.
 - Fix: [#11027] Third color on walls becomes black when saving.
 - Fix: [#11063] Scrolling position persists when switching tabs in the scenery window.
 - Fix: [#11106] Crash on getting invalid vehicle index.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -32,7 +32,6 @@
 - Fix: [#10928] File browser's date column is too narrow.
 - Fix: [#10951, #11160] Attempting to place park entrances creates ghost entrances in random locations.
 - Fix: [#11005] Company value overflows.
-- Fix: [#11010] #6024 Object selector X button closes scenario and track editors.
 - Fix: [#11027] Third color on walls becomes black when saving.
 - Fix: [#11063] Scrolling position persists when switching tabs in the scenery window.
 - Fix: [#11106] Crash on getting invalid vehicle index.
@@ -44,6 +43,7 @@
 - Fix: [#11258] Properly remove format codes from imported strings.
 - Fix: [#11286] Fix banner tooltip colour.
 - Fix: Small red gardens in RCT1 saves are imported in the wrong colour.
+- Improved: [#6024] The close button in the object selection now advances to the next step.
 - Improved: [#11157] Slimmer virtual floor lines.
 
 0.2.5 (2020-03-24)
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -462,15 +462,10 @@ static void window_editor_object_selection_mouseup(rct_window* w, rct_widgetinde
     switch (widgetIndex)
     {
         case WIDX_CLOSE:
+            window_close(w);
             if (gScreenFlags & SCREEN_FLAGS_EDITOR)
             {
-                window_close(w);
                 finish_object_selection();
-            }
-            else
-            {
-                // Used for in-game object selection cheat
-                window_close(w);
             }
             break;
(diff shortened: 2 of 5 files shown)


src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -464,8 +464,23 @@ static void window_editor_object_selection_mouseup(rct_window* w, rct_widgetinde
         case WIDX_CLOSE:
             if (gScreenFlags & SCREEN_FLAGS_EDITOR)
             {
-                auto loadOrQuitAction = LoadOrQuitAction(LoadOrQuitModes::OpenSavePrompt, PM_SAVE_BEFORE_QUIT);
-                GameActions::Execute(&loadOrQuitAction);
+                window_close(w);
+                if (gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER)
+                {
+                    set_every_ride_type_invented();
```
Gymnasiast: Is this a copy of the code that is executed when you press the "Next" button? If so, it should be a function that is called from both locations.

WantDiscussion: Done. I've also changed it so the x doesn't open up a map or the ride selection window like the next button.

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -375,6 +376,14 @@ void reset_selected_object_count_and_size()
     }
 }
 
+void finish_object_selection(uint8_t editorStep)
```
AaronVanGeffen:
```suggestion
void finish_object_selection(RCT2_EDITOR_STEP editorStep)
```

AaronVanGeffen: It's nice that you've applied this change directly from GitHub, but you'll need to adjust the headers as well, or this won't compile.

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -375,6 +376,14 @@ void reset_selected_object_count_and_size()
     }
 }
 
+void finish_object_selection(const RCT2_EDITOR_STEP editorStep)
+{
+    set_every_ride_type_invented();
+    set_every_ride_entry_invented();
+    gS6Info.editor_step = editorStep;
+    gfx_invalidate_screen();
```
Gymnasiast: You put the code from the Track Designer here, which now causes this to run in _all_ situations. You need to copy the whole if-else construct.

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -375,6 +376,24 @@ void reset_selected_object_count_and_size()
...
+    if (gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER)
+    {
+        set_every_ride_type_invented();
+        set_every_ride_entry_invented();
+        gS6Info.editor_step = EDITOR_STEP_ROLLERCOASTER_DESIGNER;
+        gfx_invalidate_screen();
+    }
+    else
+    {
+        set_every_ride_type_invented();
```
Gymnasiast: This _still_ doesn't reflect what the old if-else did: https://github.com/OpenRCT2/OpenRCT2/pull/11010/files#diff-af125fc02d11204f8f2787620a406bb7L238

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -464,8 +464,8 @@ static void window_editor_object_selection_mouseup(rct_window* w, rct_widgetinde
         case WIDX_CLOSE:
             if (gScreenFlags & SCREEN_FLAGS_EDITOR)
             {
-                auto loadOrQuitAction = LoadOrQuitAction(LoadOrQuitModes::OpenSavePrompt, PM_SAVE_BEFORE_QUIT);
-                GameActions::Execute(&loadOrQuitAction);
+                window_close(w);
```
Gymnasiast, line 467: `window_close(w)` is called in both the if and the else clause. Move it before the if and remove the else clause.

tupaschoal:
Hey @WantDiscussion can you work on @Gymnasiast suggestions? (and mark as solved the ones you already dealt with?)

distribution/changelog.txt
```diff
@@ -9,6 +9,7 @@
 - Fix: [#10928] File browser's date column is too narrow.
 - Fix: [#10951, #11160] Attempting to place park entrances creates ghost entrances in random locations.
 - Fix: [#11005] Company value overflows.
+- Fix: [#11010] #6024 Object selector X button closes scenario and track editors.
```
Gymnasiast, line 12: Should be under improved, and should list the issue number, not the PR number.

src/openrct2/EditorObjectSelectionSession.h
```diff
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include "Editor.h"
```
Gymnasiast, line 12: Unnecessary include, it seems.

PR #11012, opened 2020-03-24
Fix #11005: Company value overflows

### LINKED ISSUE #11005: Company value overflows
**OS:** Windows 10
**Version:** 0.2.4
**Commit/Build:** 2a94632

I think it is not solvable until we have new save file format, but I'd like to issue here to notice to you.

**Steps to reproduce:**
1. Set money as ``$214,748,364.70``, which indicates ``INT_MAX``
2. See the finance window

**Screenshots / Video:**
TELK PARK 2020-03-22 20-16-11

thversfelt:
In #11005, the company value overflows when the park cash is equal to `INT_MAX`, a ride is built and opened. This is fixed by clamping the company value between `INT_MIN` and `INT_MAX`. I just tested it and it works.

fix11005

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -4,6 +4,7 @@
 - Fix: [#475] Water sides drawn incorrectly (original bug).
 - Fix: [#6123, #7907, #9472, #11028] Cannot build some track designs with 4 stations (original bug).
 - Fix: [#7094] Back wall edge texture in water missing.
+- Fix: [#11005] Company value overflows.
 - Fix: [#11027] Third color on walls becomes black when saving.
 
 0.2.5 (2020-03-24)
--- a/src/openrct2/world/Park.cpp
+++ b/src/openrct2/world/Park.cpp
@@ -517,11 +517,10 @@ money32 Park::CalculateRideValue(const Ride* ride) const
 
 money32 Park::CalculateCompanyValue() const
 {
-    money32 result = finance_get_current_cash();
+    money32 result = gParkValue - gBankLoan;
 
-    // Clamp addition and substraction to prevent overflow
-    result = add_clamp_money32(result, -gBankLoan);
-    result = add_clamp_money32(result, gParkValue);
+    // Clamp addition to prevent overflow
+    result = add_clamp_money32(result, finance_get_current_cash());
 
     return result;
 }
--- a/src/openrct2/world/Park.cpp
+++ b/src/openrct2/world/Park.cpp
@@ -33,6 +33,7 @@
 #include "../ride/RideData.h"
 #include "../ride/ShopItem.h"
 #include "../scenario/Scenario.h"
+#include "../util/Util.h"
 #include "../windows/Intent.h"
 #include "Entrance.h"
 #include "Map.h"
@@ -516,14 +517,13 @@ money32 Park::CalculateRideValue(const Ride* ride) const
 
 money32 Park::CalculateCompanyValue() const
 {
-    // Cast the sum to a 64-bit value
-    money64 result = (money64)finance_get_current_cash() + (money64)gParkValue - (money64)gBankLoan;
+    money32 result = finance_get_current_cash();
 
-    // Clamp the result to the range of a 32-bit value
-    result = result < INT32_MIN ? INT32_MIN : result;
-    result = result > INT32_MAX ? INT32_MAX : result;
+    // Clamp addition and substraction to prevent overflow
+    result = add_clamp_money32(result, -gBankLoan);
+    result = add_clamp_money32(result, gParkValue);
 
-    return (money32)result;
+    return result;
 }
 
 money16 Park::CalculateTotalRideValueForMoney() const
--- a/src/openrct2/world/Park.cpp
+++ b/src/openrct2/world/Park.cpp
@@ -516,7 +516,14 @@ money32 Park::CalculateRideValue(const Ride* ride) const
 
 money32 Park::CalculateCompanyValue() const
 {
-    return finance_get_current_cash() + gParkValue - gBankLoan;
+    // Cast the sum to a 64-bit value
+    money64 result = (money64)finance_get_current_cash() + (money64)gParkValue - (money64)gBankLoan;
+
+    // Clamp the result to the range of a 32-bit value
+    result = result < INT32_MIN ? INT32_MIN : result;
+    result = result > INT32_MAX ? INT32_MAX : result;
+
+    return (money32)result;
 }
 
 money16 Park::CalculateTotalRideValueForMoney() const
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -31,7 +31,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "1"
+#define NETWORK_STREAM_VERSION "2"
 #define NETWORK_STREAM_ID OPENRCT2_VERSION "-" NETWORK_STREAM_VERSION
 
 static Peep* _pickup_peep = nullptr;


src/openrct2/world/Park.cpp
```diff
@@ -516,7 +516,14 @@ money32 Park::CalculateRideValue(const Ride* ride) const
 
 money32 Park::CalculateCompanyValue() const
 {
-    return finance_get_current_cash() + gParkValue - gBankLoan;
+    // Cast the sum to a 64-bit value
+    money64 result = (money64)finance_get_current_cash() + (money64)gParkValue - (money64)gBankLoan;
+
+    // Clamp the result to the range of a 32-bit value
+    result = result < INT32_MIN ? INT32_MIN : result;
```
tupaschoal: `std::max`

src/openrct2/world/Park.cpp
```diff
@@ -516,7 +516,14 @@ money32 Park::CalculateRideValue(const Ride* ride) const
...
+    result = result > INT32_MAX ? INT32_MAX : result;
```
tupaschoal: `std::min`

src/openrct2/world/Park.cpp
```diff
@@ -516,7 +516,14 @@ money32 Park::CalculateRideValue(const Ride* ride) const
 
 money32 Park::CalculateCompanyValue() const
 {
-    return finance_get_current_cash() + gParkValue - gBankLoan;
+    // Cast the sum to a 64-bit value
+    money64 result = (money64)finance_get_current_cash() + (money64)gParkValue - (money64)gBankLoan;
```
tupaschoal: Use `static_cast` instead of c-style cast

src/openrct2/world/Park.cpp
```diff
@@ -516,7 +516,14 @@ money32 Park::CalculateRideValue(const Ride* ride) const
...
+
+    // Clamp the result to the range of a 32-bit value
+    result = result < INT32_MIN ? INT32_MIN : result;
+    result = result > INT32_MAX ? INT32_MAX : result;
+
+    return (money32)result;
```
tupaschoal: This cast can be omitted and done implicitly

thversfelt: > 
> 
> I'm not sure if this is the way to go, but I have added some comments of best practices

I will modify and commit with best practices.

thversfelt:
Alternatively, it can also be clamped without casting to a 64-bit integer. See [here](https://www.geeksforgeeks.org/int_max-int_min-cc-applications/) for an example. What would be preferred?

janisozaur:
I recall adding some overflow/clamp functions to the code. Please check somewhere in the math/utils, I'm on mobile now and can't do that myself. Also note it is not required to extend value to larger type to check overflow.

thversfelt:
> 
> 
> I recall adding some overflow/clamp functions to the code. Please check somewhere in the math/utils, I'm on mobile now and can't do that myself. Also note it is not required to extend value to larger type to check overflow.

Thanks, I am searching now. I will add it in a couple of hours.

thversfelt:
Thanks @janisozaur, there was indeed a clamp function in util. I have replaced the casting with clamping, tested it, and it works correctly.

Gymnasiast:
LGTM.

Could you add the following changelog entry?

`Fix: [#11005] Company value overflows`

distribution/changelog.txt
```diff
@@ -42,6 +42,7 @@
 - Fix: [#10941] The Clear Scenery tool gives refunds for ghost elements.
 - Fix: [#10963] Light effects are drawn off-centre in some rotations.
 - Fix: [#11001] Rides list does not use natural sorting.
+- Fix: [#11005] Company value overflows
```
tupaschoal:
```suggestion
- Fix: [#11005] Company value overflows.
```
Very minor thing :P

thversfelt: Haha, but still important :)

src/openrct2/world/Park.cpp
```diff
@@ -516,7 +517,13 @@ money32 Park::CalculateRideValue(const Ride* ride) const
 
 money32 Park::CalculateCompanyValue() const
 {
-    return finance_get_current_cash() + gParkValue - gBankLoan;
+    money32 result = finance_get_current_cash();
+
+    // Clamp addition and substraction to prevent overflow
+    result = add_clamp_money32(result, -gBankLoan);
+    result = add_clamp_money32(result, gParkValue);
```
janisozaur: This might alter the outcome and I would like to know why it overflows in the first place.

I'm going to assume `finance_get_current_cash()` can only return positive values or zero.

The problem is you can have maxed bank loan, `gBankLoan` much larger than `finance_get_current_cash()`, but it will get clamped and then `gParkValue > 0`, which would skew the calculated company value. Previously such case _could_ wrap twice (both ways) and end up being correct.

This is just my speculation, I don't know if this can actually happen.

For parks that wrap the value, what's the most probable reason? Wouldn't it make more sense to calculate `pvbl = gParkValue - gBankLoad` first and then add `result = current_cash + pvbl`?

thversfelt: I believe `finance_get_current_cash()` returns negative values as well:
image

I am not sure if I follow the part about `gParkValue`. The `CalculateParkValue()` method, which sets `gParkValue`, only takes the number of guests and the ride values into account. So, I don't see how that would skew the calculated company value.

If `gBankLoan` is much larger than `finance_get_current_cash()`, then the company value will be clamped to `INT32_MIN` instead of overflowing.

thversfelt: Also, the reasoning behind why it overflows in the first place is as follows. Given that `result = finance_get_current_cash() + pvbl` with `pvbl = gParkValue - gBankLoan` . So:

1. Use cheat to max cash: `finance_get_current_cash() = INT32_MAX`
2. A situation arises where `pvbl = gParkValue - gBankLoan > 0`. This can happen when the park value is updated when a ride is opened (which explains why it overflows as seen in #11005).
3. Then, `result = INT32_MAX + pvbl` and since `pvbl > 0`, it must hold that `result > INT32_MAX`. Thus, overflow happens.

thversfelt:
After reconsidering the review by @janisozaur, it turned out that indeed the company value could be skewed when `finance_get_current_cash() = INT32_MIN` and `gBankLoan > gParkValue`. The resulting company value would not get clamped to the expected `INT32_MIN`, but would become `INT32_MIN + gParkValue`.

so I implemented the proposed change. First `result = gParkValue - gBankLoan` is calculated. And since `0 < gParkValue < INT32_MAX` and `0 < gBankLoan < INT_32MAX`, the result will not overflow. Then, when adding the cash, the addition is clamped to prevent overflow in cases where `finance_get_current_cash() + gParkValue - gBankLoan > INT32MAX`.

I tested it just now and it works as expected.

src/openrct2/world/Park.cpp
```diff
@@ -516,7 +517,12 @@ money32 Park::CalculateRideValue(const Ride* ride) const
 
 money32 Park::CalculateCompanyValue() const
 {
-    return finance_get_current_cash() + gParkValue - gBankLoan;
+    money32 result = gParkValue - gBankLoan;
```
janisozaur: Are these small enough to not use `add_clamp_money32`?

janisozaur:
Many thanks for your investigation, I'm happy for this to be merged once you confirm my last comment.

src/openrct2/world/Park.cpp
thversfelt: I dug through the code and concluded that `gBankLoan` is clamped by `gMaxBankLoan`, which is set to `gMaxBankLoan = 5000000` so that is small enough. `gParkValue` is calculated through the `CalculateParkValue()` method:

```
money32 Park::CalculateParkValue() const
{
    // Sum ride values
    money32 result = 0;
    for (const auto& ride : GetRideManager())
    {
        result += CalculateRideValue(&ride);
    }

    // +7.00 per guest
    result += gNumGuestsInPark * MONEY(7, 00);

    return result;
}
```

There is no overflow protection/clamping present, so theoretically it could overflow if there is a massive amount of rides and guest. I don't think that is a realistic scenario, but then again, getting max cash is also a pretty unrealistic scenario and it still drove me to make this fix.

janisozaur:
I just realized there is one more thing required: increase the network version, as the calculated value may be used in logic code. Just increase the number in `Network.cpp`

janisozaur:
Rebased + increased network version + moved changelog entry to new version.

thversfelt:
> Rebased + increased network version + moved changelog entry to new version.

Thanks!

Issue #10955, opened 2020-03-17
Copy/Paste-ability of Input Strings


Richard-L:
OpenRCT2, v0.2.4-719-gdcdc4f829 (dcdc4f8 on develop) provided by GitHub
Windows (x86-64)

![Geauga Knockoff 2020-03-17 19-25-04](https://user-images.githubusercontent.com/4934209/76889228-86d0ee80-6885-11ea-86d0-579012b30fa7.png)

^would it mean much trouble to allow highlighting and copy/pasting functionality with CTRL+C/X/V?

Thank you.

AaronVanGeffen:
Thank you, this is a good feature request. Unfortunately, it isn't trivial to add -- as of yet, text selection hasn't been implemented at all.

IntelOrca:
Just to clarify, we already have paste (CTRL+V) support but no selection support which is needed for cut, delete, copy, and delete-and-paste.

janisozaur:
We could try having ^C copy entire current buffer, before text selection is implemented

daandeheij:
I'm working on a simple implementation now, as suggested by @janisozaur

duncanspumpkin:
Implemented in #11013.

PR #11013, opened 2020-03-24
Simple implementation of copy input to clipboard (Ctrl+C)

daandeheij:
The ability to copy input strings to the clipboard was requested in #10955. This is a very simple implementation using Ctrl+C, as suggested by @janisozaur.

I wanted to include a simple acknowledgement that the text is copied (as this is not very clear as selecting text is not possible). I have used the `context_show_error` in this PR, but I am not sure whether this is the most appropriate way. Of course, this message can also be omitted.   

The following screenshot shows the `context_show_error`. Note that the string in this commit is `copied text to clipboard`, contrary to this screenshot (I changed the string after taking the screenshot).

image


tupaschoal:
Can you add a screenshot of what the message looks like to the PR description, please?

In any case, I'm wondering whether we need an information pop-up or not.

daandeheij:
I just added a screenshot. I think ideally would be to have some sort of pop-up/message box notify the user that the text was copied, and then to disappear on its own. This current message sticks around until the next key/mouse press, which is slightly inconvenient.

src/openrct2-ui/TextComposition.cpp
```diff
@@ -175,6 +175,13 @@ void TextComposition::HandleMessage(const SDL_Event* e)
                     CursorRight();
                     console.RefreshCaret();
                     break;
+                case SDLK_c:
+                    if (modifier & KEYBOARD_PRIMARY_MODIFIER && _session.Length)
```
janisozaur: Just to be explicit about evaluation order here, I would suggest adding parentheses, as done in other such constructs around:
```suggestion
                    if ((modifier & KEYBOARD_PRIMARY_MODIFIER) && _session.Length)
```

AaronVanGeffen:
Could you add a changelog entry as well? (We just released v0.2.5 tonight, so it'll have to go under v0.2.5+.)

daandeheij:
Added a changelog entry just now under v.0.2.5+

duncanspumpkin:
@daandeheij in future have as your commit message `Fix #<number of issue>: description...` That way it will auto close the issue on merge.

PR #11014, opened 2020-03-24
Rework GuestList to use a vector.

duncanspumpkin:
This is part of a chain of refactors that will be used to rework peep order in the sprite list.
I basically want to change it so that peep order in the sprite list does not matter as that will make it easier to handle.

It has the added bonus of simplifying the code for the guest list.
Same change still needs to be made to the staff list before peep order can be changed.

~~Todo: Can remove assignments to `var_492` to simplify code as well.~~

There will be a small performance penalty with this new version of the code but its hard to tell if its any greater than the performance of the old code.

### DIFF
--- a/src/openrct2-ui/WindowManager.cpp
+++ b/src/openrct2-ui/WindowManager.cpp
@@ -417,6 +417,7 @@ public:
                 gToolbarDirtyFlags |= BTM_TB_DIRTY_FLAG_PEEP_COUNT;
                 window_invalidate_by_class(WC_GUEST_LIST);
                 window_invalidate_by_class(WC_PARK_INFORMATION);
+                window_guest_list_refresh_list();
                 break;
 
             case INTENT_ACTION_UPDATE_PARK_RATING:
--- a/src/openrct2-ui/windows/GuestList.cpp
+++ b/src/openrct2-ui/windows/GuestList.cpp
@@ -20,6 +20,7 @@
 #include <openrct2/sprites.h>
 #include <openrct2/util/Util.h>
 #include <openrct2/world/Sprite.h>
+#include <set>
 
 // clang-format off
 enum {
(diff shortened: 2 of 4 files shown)


src/openrct2-ui/windows/GuestList.cpp
```diff
@@ -482,29 +518,14 @@ static void window_guest_list_update(rct_window* w)
  */
 static void window_guest_list_scrollgetsize(rct_window* w, int32_t scrollIndex, int32_t* width, int32_t* height)
 {
-    int32_t i, y, numGuests, spriteIndex;
-    Peep* peep;
-
+    int32_t i, y;
```
Gymnasiast: Declaration of `i` can be moved to the assignment. And you might want to initialise `y` to 0.

src/openrct2-ui/windows/GuestList.cpp
```diff
@@ -549,35 +570,21 @@ static void window_guest_list_scrollgetsize(rct_window* w, int32_t scrollIndex,
...
 static void window_guest_list_scrollmousedown(rct_window* w, int32_t scrollIndex, const ScreenCoordsXY& screenCoords)
 {
-    int32_t i, spriteIndex;
-    Peep* peep;
+    int32_t i;
 
     switch (_window_guest_list_selected_tab)
     {
         case PAGE_INDIVIDUAL:
             i = screenCoords.y / SCROLLABLE_ROW_HEIGHT;
```
Gymnasiast: Declaration of `i` can be moved here.

src/openrct2-ui/WindowManager.cpp
```diff
@@ -417,6 +417,7 @@ class WindowManager final : public IWindowManager
                 gToolbarDirtyFlags |= BTM_TB_DIRTY_FLAG_PEEP_COUNT;
                 window_invalidate_by_class(WC_GUEST_LIST);
                 window_invalidate_by_class(WC_PARK_INFORMATION);
+                window_guest_list_refresh_list();
```
Gymnasiast: Is it necessary/useful to have this function separate from the invalidation function?

tupaschoal: And why not `window_refresh_guest_list`?

duncanspumpkin: Yes, invalidate is for forcing a redraw this is forcing a recalc of the data. Calling this function on every invalidate would be expensive (i tasted that already). I guess you could do the reverse invalidate on all calls to this function but not really a huge need.

src/openrct2/peep/Peep.cpp
```diff
@@ -3241,7 +3241,7 @@ rct_string_id get_real_name_string_id_from_id(uint32_t id)
     return dx;
 }
 
-static int32_t peep_compare(const void* sprite_index_a, const void* sprite_index_b)
+int32_t peep_compare(const void* sprite_index_a, const void* sprite_index_b)
```
tupaschoal: Why? I'm assuming there are more changes to follow

duncanspumpkin: ? Not sure what you mean. I need this function in the guest list that's why it's no longer static.

tupaschoal: I misunderstood, my bad.

src/openrct2-ui/windows/GuestList.cpp
duncanspumpkin: its used by both cases of the switch so I think I'll just leave it. I'll init it to 0 though.

PR #11029, opened 2020-03-25
Fix #11027: Third color on walls becomes black when saving

### LINKED ISSUE #11027: Third color on objects becoming black when saving
**OS:** Windows 7
**Version:** 0.2.5
**Commit/Build:** 4f6e77e and 302fe1c
Applies both on release and develop.

When we place a scenery object with a third color different from black, then save the game, then open the save, it becomes black.

- Reproducible in RCT2 (vanilla)? **No**
- Specific to multiplayer? **No**

**Steps to reproduce:**
1. Place a scenery object with a third color.
2. Save the game.
3. Open the save.

**Screenshots / Video:**
Before:
image
After:
image

**Save game:**
[test color.zip](https://github.com/OpenRCT2/OpenRCT2/files/4379258/test.color.zip)

Gymnasiast:
The problem here was that setting the banner index on walls would overwrite `colour_3`. Changed import and export to only do something with banner index if the object in question is one.


AaronVanGeffen:
Thanks for figuring this out so quickly.

We should make a replay test for this to ensure it doesn’t happen again.

duncanspumpkin:
A replay test wouldn't help. It's a import export mistake

PR #11031, opened 2020-03-25
Fix #6123: Cannot build some track designs with 4 stations

### LINKED ISSUE #6123: This track design fails to save/load correctly
**OS:** Windows 10
**Version:** 0.1.1 develop
**Commit/Build:** 2dc8746

I built a rollercoaster with 4 stations and a dummy coaster to sync them. Whenever I try to save and load the coaster, ~$30000 is charged but the coaster is not built. Disabling clearance checks does not help. While building, I switched from inverted coaster to steel looping coaster. #6094 does not fix this.

- [ ] Multiplayer?

**Steps to reproduce:**
1. Save track design
2. Load track design

**Screenshots / Video:**
happy park 2017-08-04 22-48-47

**Save game:**
[Quadruple steel racer.td6.txt](https://github.com/OpenRCT2/OpenRCT2/files/1202328/Quadruple.steel.racer.td6.txt)
[Racers Park.sv6.txt](https://github.com/OpenRCT2/OpenRCT2/files/1202329/Racers.Park.sv6.txt)

Gymnasiast:


duncanspumpkin:
This was an original bug we should mention that in the changelog

PR #11035, opened 2020-03-25
Update changelog to state fix of original bug (#11031)

tupaschoal:


Gymnasiast:
I severely doubt what Duncan said, so I'm going to test it myself first.

Issue #10675, opened 2020-02-08
Filter track designs by available scenery/vehicles


Superjustinbros:
Simply put, something on the track design selection window that lets you filter the selection of track designs that do or do not include the required scenery or the required vehicle type in the current scenario.

This way you could round the list down to only rides that contain neither "Design includes scenery which is unavailable" and/or "Vehicle Design Unavailable- ride performance may be affected" when choosing a design.

PR #11040, opened 2020-03-25
[WIP] Filter track designs by available scenery/vehicles

thversfelt:
I am attempting to implement the feature that was requested in #10675, by adding a checkbox to the track list which allows the player to filter the designs based on the availability of scenery and vehicles.

image

However, in its current state it is not functioning correctly. I filter the track list by looping through all track designs, loading each of them and attempting to read the `TRACK_DESIGN_FLAG_SCENERY_UNAVAILABLE` and `TRACK_DESIGN_FLAG_VEHICLE_UNAVAILABLE` flags. If one of these flags is set for a design, it will not pass the filter.

Moreover, to make sure it works with filtering on the name, I have modified the string filter such that a track design passes through the two filters. First, it passes through the string filter and then it passes through the availability filter. If it manages to pass these filters, the track design gets added to the list of filtered track id's.

Right now, it appears as though the availability flags are both not set, even though they should be set for some rides, so no availibility filtering is actually happening. Perhaps I am forgetting something? I am not sure. Also, I get occasional crashes when filtering by name, this has only happened with my modified code, so it is not a bug in `develop`.

Any help would be appreciated, thanks.

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3676,6 +3676,8 @@ STR_6359    :{POP16}{POP16}Page {UINT16}
 STR_6360    :{SMALLFONT}{BLACK}{COMMA32}
 STR_6361    :Enable lighting effects on rides (experimental)
 STR_6362    :{SMALLFONT}{BLACK}If enabled, vehicles for tracked rides will be lit up at night.
+STR_6363    :Hide unavailable designs
+STR_6364    :{SMALLFONT}{BLACK}Hides track designs with unavailable scenery or vehicles
 
 #############
 # Scenarios #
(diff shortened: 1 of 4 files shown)


tupaschoal:
As a side note, when the implementation is not finished, please open it as a draft PR instead

duncanspumpkin:
I'm not sure this approach would be good. This is going to be quite expensive to check as you need to do a full load of all of the track designs. You might not notice this on an ssd or when there aren't a lot of designs but for a hdd you will get a big slow down. The better approach might be to load and save more data into the track design index.

thversfelt:
> 
> 
> As a side note, when the implementation is not finished, please open it as a draft PR instead

I wasn't aware of this feature, thanks for letting me know! Will do this correctly next time.

thversfelt:
> 
> 
> I'm not sure this approach would be good. This is going to be quite expensive to check as you need to do a full load of all of the track designs. You might not notice this on an ssd or when there aren't a lot of designs but for a hdd you will get a big slow down. The better approach might be to load and save more data into the track design index.

Thank you for your feedback, and that sounds more efficient yes. I am going to look into loading and saving data to the index.

In the current state, the code is also parsing the filter string to lower case during every loop iteration, which only has to be done once before entering the loop, so that can also still be a performance improvement.

tupaschoal:
As this hasn't been touched in a while, I'm closing this PR. Please reopen if you rework on it and get to the final solution

Issue #11039, opened 2020-03-25
Map(s) Crash shortly or directly after loading


Victorvdvliet:
**OS:** Windows 10
**Version:** 0.2.5-15
**Commit/Build:** g3f473ec3f (3f473ec on develop) 

Map crashes as soon as, or very soon after, loading it.

- [not sure ] Reproducible in RCT2 (vanilla)?
- [no] Specific to multiplayer?

**Steps to reproduce:**
1. Load attached map
2. Game crashes (main exutable doesnt work)
3. ??
4. Profit

**Dump file**

[cca3573a-da8b-4866-8072-618770aa67bf(3f473ec_x86-64).dmp.gz](https://github.com/OpenRCT2/OpenRCT2/files/4382003/cca3573a-da8b-4866-8072-618770aa67bf.3f473ec_x86-64.dmp.gz)

**Screenshots / Video:**

![Error mini contest](https://user-images.githubusercontent.com/48918199/77558431-3710a980-6ebb-11ea-8d61-589b3bcd457b.png)

**Save game:**

[Deurklink mini entry _Victator - kopie.txt](https://github.com/OpenRCT2/OpenRCT2/files/4382030/Deurklink.mini.entry._Victator.-.kopie.txt)

Gymnasiast:
I have a hunch as to what might cause this. Could you temporarily switch to the 0.2.5 release and tell if it happens there as well?

Victorvdvliet:
I'm using the launcher, don't see any options to switch to an older version. On the openrct2 website 0.2.5 was updated 20 hours ago. I assume that is not the version you mean right?

Gymnasiast:
The launcher provides an option to switch between release and develop.

Victorvdvliet:
It works in release, thanks!

P.S. For some reason the elevator stops working as soon as you load the file, despite working prior to saving (I let it run for an hour before saving to be sure). Not as important, but I thought you might like to know.

Gymnasiast:
This should now also have been fixed in `develop`. Could you update to 268e81c21baab50959ca4e00e7af4bd7679c7ee5 or newer and see if the issue is now solved?

duncanspumpkin:
I've tested with the attached save. It's safe to close this

PR #11044, opened 2020-03-25
Fix mistake in #11029

Gymnasiast:


duncanspumpkin:
ah you just beat me by a few seconds was just writing my commit message.

duncanspumpkin:
Fixes #11039. And i expect a whole bunch of backtrace reports.

PR #10573, opened 2020-01-14
Move Windows CI from AppVeyor to GitHub actions

IntelOrca:

PR #11046, opened 2020-03-25
Restore uploading of debug symbols to backtrace.io

janisozaur:
Regression from ee09ffadf7c8e93865e943015b442adb1c2c8869 (https://github.com/OpenRCT2/OpenRCT2/pull/10573)


IntelOrca:
If we want to conceal the token we can add a GitHub secret like how openrct2.org is done.

janisozaur:
Any sign-offs for the updated version?

PR #11052, opened 2020-03-25
Fix undefined behaviour in ZoomLevel.hpp

janisozaur:
Cf. https://docs.microsoft.com/en-us/cpp/cpp/left-shift-and-right-shift-operators-input-and-output?view=vs-2019#footnotes


IntelOrca:
Why has this only been applied to `*` but not `/`?

Doesn't this generate less efficient instructions? It is no longer going to generate a single `SHL`, `SHR`, `SAL` or `SAR`.

ZehMatt:
I would also suggest to simply cast it to unsigned type, also use std::abs.

brenoguim:
> Doesn't this generate less efficient instructions? It is no longer going to generate a single SHL, SHR, SAL or SAR.

Compilers are really good at detecting operations on power of two:  https://godbolt.org/z/TZJA47

> I would also suggest to simply cast it to unsigned type, also use std::abs.

I'm not sure you can cast the lhs to unsigned because of code like this: 
```
Audio.cpp:225 -1024 * viewport->zoom
```

Personally, I would recommend creating a "magnitude" function and "magnify()" functions:
```
unsigned magnitude() const { return 1U << std::abs(_level); }
bool magnify() const { return _level >= 0; }
```
And then implement the operators as:
```
 template<typename T> friend T operator*(const T& lhs, const ZoomLevel& rhs)
{
    return magnify() ? lhs * magnitude() : lhs / magnitude();
}
 template<typename T> friend T operator*(const T& lhs, const ZoomLevel& rhs)
{
    return magnify() ? lhs / magnitude() : lhs * magnitude();
}
```
Inliner and optimizer will eat those for breakfast :D

IntelOrca:
Based on https://godbolt.org/z/Nb_-J-
This change will cause several more instructions including an `idiv` instruction to be generated instead of a single `asr` instruction.

I suggest at least doing something like:
```
int asr(int x, unsigned int y)
{
#if -1 >> 1 != -1
    return x / (1 << y);
#else
    return x >> y;
#endif
}
```

So `ASR` continues to get generated.

brenoguim:
But then you are back to undefined behavior land. One cannot shift a signed integer without checking if it's negative first. If you tell the compiler you want to work with unsigned integers then it will happily generate the efficient code: https://godbolt.org/z/DcwsME

This is how the assembly for the option I suggested looks like: https://godbolt.org/z/yXU--g

janisozaur:
I don't understand what the preprocessor has to do with undefined behaviour, as explicitly stated in the standard. @IntelOrca can you explain?

IntelOrca:
This will happen for signed numbers too so we can't rely on logical shift. We want an ASR instruction generated. Prove the compiler generates this with this code.

brenoguim:
You're right. Doing right shift of a negative number is not undefined behavior, it's implementation defined behavior.
That means vendors may choose how to implement that.

It just a habit of mine not to rely on "implementation defined behavior" because then you gotta check on all compilers you support and all architectures that they are actually doing what you would want. (And you have to remember to double check that when adding more support)

My personal recommendation is to not rely on that because unless it has measurable performance impact.
That is why my suggestion was to check if the number was negative and behave differently then.

IntelOrca:
But that will incur a branch instruction. My code checks if the behaviour of the code will be the correct behaviour and allow the ASR to generate. If it isn't it will do the div instruction.

brenoguim:
I would take the branch instruction over the hassle of checking compiler documentation :) I say that, because I don't think the preprocessor is required to adhere to the exact same semantics... I'm not sure that check is valid.

IntelOrca:
Then I would then just allow certain compilers and platforms which we know generate an ASR.

brenoguim:
I'm not a member of the project, so that's entirely up to you and the team to decide.

Personally, I would not create special cases for micro-optimizations unless I saw a measurable improvement. I tend to prefer a simpler high-level code in most places, and fine-tune specific places after profiling.

IntelOrca:
@brenoguim this is a core struct used throughout all our viewport and software rendering code. This should be micro-optimised.

janisozaur:
My proposition:
1. Do `if constexpr` here, the shift is indeed implementation-defined and that removes doubts I have about @IntelOrca 's suggestion of employing preprocessor
2. Add a test that would exercise this function and fail if the shift-by-negative does not produce expected values (this could use slow version for verification), for porter's benefit

IntelOrca:
Didn't know about `if constexpr`, that would be better than `#if`.

brenoguim:
That would be fine. Another possibly relevant action item is to see what is the degradation seen in the rendering benchmarks.

PR #11059, opened 2020-03-26
Rework StaffList to use a vector

duncanspumpkin:
This is a part of a reworking that removes the need for the peep list to be sorted by name/number order. Ultimately simplifying the sprite create code.

There will be a very minor performance impact of this new code as it is running every tick that the staff list is open. As the max number of staff is low this is not at present an issue. It would need looked at in the future to use intents to invalidate the list on fire/hire/death to handle this more efficient (like the guest list does).

### DIFF
--- a/src/openrct2-ui/windows/StaffList.cpp
+++ b/src/openrct2-ui/windows/StaffList.cpp
@@ -188,9 +188,6 @@ void WindowStaffListRefresh()
     {
         return;
     }
-
-    w->no_list_items = 0;
-
     StaffList.clear();
     Peep* peep = nullptr;
     uint16_t spriteIndex;
--- a/src/openrct2-ui/windows/StaffList.cpp
+++ b/src/openrct2-ui/windows/StaffList.cpp
@@ -28,6 +28,7 @@
 #include <openrct2/world/Footpath.h>
 #include <openrct2/world/Park.h>
 #include <openrct2/world/Sprite.h>
+#include <vector>
 
 // clang-format off
 enum {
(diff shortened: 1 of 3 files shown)


src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -176,9 +177,36 @@ rct_window* window_staff_list_open()
...
 
+void WindowStaffListRefresh()
+{
+    auto w = window_find_by_class(WC_STAFF_LIST);
+    if (w == nullptr)
+    {
+        return;
+    }
+
+    w->no_list_items = 0;
```
duncanspumpkin: This is never actually used. Remove.

src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -176,9 +177,36 @@ rct_window* window_staff_list_open()
     window->max_height = MAX_WH;
     _quick_fire_mode = false;
 
+    WindowStaffListRefresh();
     return window;
 }
 
+void WindowStaffListRefresh()
```
Gymnasiast: Wouldn't a static function on Window be better (`Window::RefreshStaffList()` or something)?

duncanspumpkin: What's our plan with Windows? Are we going to make each window a sub class of window?

Gymnasiast: That would probably be the way to go. It would allow us to define invalidate/mouse-up/etc. functions without a lookup table, among other things.

IntelOrca: Yeah, I experimented with this a couple of years back:
https://github.com/IntelOrca/OpenRCT2/blob/refactor/window-3/src/openrct2-ui/windows/Cheats.cpp

duncanspumpkin: I'm going to leave this until the windows refactor then.

PR #11067, opened 2020-03-26
Add missing check for valid scenery type

janisozaur:
Regression from #10922

[10922.sv6.gz](https://github.com/OpenRCT2/OpenRCT2/files/4389978/10922.sv6.gz)

### DIFF
--- a/src/openrct2/management/Research.cpp
+++ b/src/openrct2/management/Research.cpp
@@ -601,6 +601,7 @@ void ride_entry_set_invented(int32_t rideEntryIndex)
 
 bool scenery_is_invented(const ScenerySelection& sceneryItem)
 {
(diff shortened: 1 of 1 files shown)


Gymnasiast:
@tupaschoal Also on release.
This guards something that *should* never happen (because import guards against it), but could cause a lot of hard-to-detect headaches if it ever *did* happen. A second-line defence if you will.

AaronVanGeffen:
@Gymnasiast No, a normal `assert` will only be compiled in debug builds, and ignored in release builds. This is why we have our own OpenRCT2 guard functions.

IntelOrca:
For something like this, I would just return false. We should make our code robust and asserts do not do that.

PR #11080, opened 2020-03-27
Add extra null checks to banner window code

Gymnasiast:
Since it iterates over tile elements, check for null on every iteration. Also adds a few extra checks for null scenery entries.


duncanspumpkin:
There is no way for the new tile element checks to be hit. You can't get a nullptr from incrementing a valid pointer unless you are talking about some super edge case. We don't perform this sort of check on any other loop.

PR #11094, opened 2020-03-27
Fix #4617: signs causing crashes

### LINKED ISSUE #4617: Signs Causing Crashes
**OS:** Windows 10
**Version:** e.g. 0.0.5
**Commit/Build:** 5fa5f29 and many previous ones
**Explanation of the issue:** 
I'm having several issues with the "3D Sign" and "Scrolling Signs." The first issue I noticed was after I had placed 4 of the 3D Signs. When I placed the Scrolling Sign, it changed one of my 3D Signs to the text that was on the Scrolling Sign. When I tried to switch that 3D Sign back, it changed the Scrolling Sign as well. Then, oddly enough, I tried to nickname a Janitor and it changed a different 3D Sign to his name.  And occasionally when I try to edit the signs I will crash upon right clicking them. The bug seems to vary over time, as certain signs will cause crashes one day and then different ones will cause crashes the next day while the old ones are fine. I haven't been able to reproduce it or have it occur in the multiplayer.
**Reproducible in RCT2 (vanilla)?** Unknown
**Multiplayer?** Unknown
**Steps to reproduce:**
1. Glitch varies, difficult to reproduce. To see it, open my saved game and right click the four signs on the two buildings. One of them will likely cause a crash.
**Dump file:**
[8b7db1df-c3fc-4f7a-a56b-a56add8d724a(5fa5f29).zip](https://github.com/OpenRCT2/OpenRCT2/files/530302/8b7db1df-c3fc-4f7a-a56b-a56add8d724a.5fa5f29.zip)

**Save game:**
[Monsanto World.zip](https://github.com/OpenRCT2/OpenRCT2/files/530303/Monsanto.World.zip)

janisozaur:
This addresses the crash, but I will leave it to @Gymnasiast to figure out if it's the proper way of handling it.

### DIFF
--- a/src/openrct2-ui/windows/Sign.cpp
+++ b/src/openrct2-ui/windows/Sign.cpp
@@ -178,6 +178,10 @@ rct_window* window_sign_open(rct_windownumber number)
             }
         }
         tile_element++;
+        if (tile_element >= &gTileElements[std::size(gTileElements)])
+        {
+            return nullptr;
+        }
     }
 
     int32_t view_z = tile_element->GetBaseZ();
@@ -232,6 +236,10 @@ static void window_sign_mouseup(rct_window* w, rct_widgetindex widgetIndex)
                     }
                 }
                 tile_element++;
+                if (tile_element >= &gTileElements[std::size(gTileElements)])
+                {
+                    return;
+                }
             }
 
             auto sceneryRemoveAction = LargeSceneryRemoveAction(


Gymnasiast:
This indeed addresses the crash when right clicking the banners. Do I understand correctly it's because it increments beyond the size of the gTileElements array?

There is still a crash when trying to remove the bugged banners, though.

janisozaur:
yes, that's correct. I added a fix for deletion of signs too, but you still can change the text they carry, so I guess there is a code somewhere which handles it correctly.

src/openrct2-ui/windows/Sign.cpp
```diff
@@ -232,6 +236,10 @@ static void window_sign_mouseup(rct_window* w, rct_widgetindex widgetIndex)
                     }
                 }
                 tile_element++;
+                if (tile_element >= &gTileElements[std::size(gTileElements)])
```
duncanspumpkin: This should only loop through tile elements at this coordinate. So we should break the loop and return if on last element.

duncanspumpkin:
I think for both of these it shouldn't be going past the last element for that tile. That prevents us from doing anything unusual.

src/openrct2-ui/windows/Sign.cpp
Gymnasiast: I have tried that, but that seems to cause _more_ crashes, since it then cannot find the large scenery element.

Gymnasiast:
I found another crash.

1. Load the save in #4617.
2. Rotate 180 degrees.
3. Right-click the right on the right. It will open a window with no viewport.
4. Rotate the view 90 degrees - this crashes OpenRCT2.

src/openrct2-ui/windows/Sign.cpp
duncanspumpkin: But that doesn't make any sense. All of the tiles that comprise of the large scenery element will have the banner index set if it is a scrolling banner. How would the LargeSceneryRemoveAction work if the x,y coordinate is incorrect?

Gymnasiast:
I decided to merge what was here, since it at least fixes *some* crashes, and leave #4617 open.

PR #11105, opened 2020-03-28
Rename SPRITE_LIST_TRAIN_HEAD to reduce car head confusion

duncanspumpkin:
A single car on a coaster can comprise of multiple vehicles. One of those vehicles will be the car head vehicle. The train can comprise of multiple cars. Therefore not all vehicle->IsHead 's will be in SPRITE_LIST_TRAIN_HEAD. This rename makes that slightly more clear


Gymnasiast:
So `vehicle->IsHead()` can also refer to car read, while SPRITE_LIST_TRAIN_HEAD will always refer to the train head?

duncanspumpkin:
Yeah there is another renaming i need to make as we have a get head function that gets the car head and a get head function that gets the train head.

PR #11112, opened 2020-03-28
Fix #11111: Finance graph hover values are glitchy with software renderer

### LINKED ISSUE #11111: Finance graph hover values are glitchy with software renderer
**OS:** Windows 10 (x64)
**Version:** 0.2.5
**Commit/Build:** 2806104

This only happens with the software renderer, I don't have these issues with OpenGL.

**Steps to reproduce:**
1. Mouse your cursor over the finance graph while using software renderer

**Save game:**
[graph.zip](https://github.com/OpenRCT2/OpenRCT2/files/4397357/graph.zip)

AaronVanGeffen:
The new finance graph guides (#10925) did not invalidate the window, so they were rendered incompletely or not at all if no sprites were active behind the window.

The window already had update events for both graph tabs, but these were only updating the tab widgets for their animations. This changes it such that the entire window is invalidated. I'm not sure this is the best solution; this seems a little wasteful.

It would be better to only invalidate the window if mouse movement is detected, but to my knowledge, we don't support a `mouseover` event.

### DIFF
--- a/src/openrct2-ui/windows/Finances.cpp
+++ b/src/openrct2-ui/windows/Finances.cpp
@@ -811,7 +811,8 @@ static void window_finances_financial_graph_update(rct_window* w)
     // Tab animation
     if (++w->frame_no >= WindowFinancesTabAnimationLoops[w->page])
         w->frame_no = 0;
-    widget_invalidate(w, WIDX_TAB_2);
+
+    w->Invalidate();
 }
 
 /**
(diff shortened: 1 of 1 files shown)


janisozaur:
You could create ad-hoc one by caching previous cursor position and comparing it with current, only invalidating when the two are different

AaronVanGeffen:
I would prefer to create a mouseover event instead. We currently have three 'unknown' events that do not appear to be in use: 05, 08, and 15. (05 is declared for the staff window, but not really used, I think).

My point is that there's room for a new event — but I think that's beyond the scope of this PR.

PR #11136, opened 2020-03-29
Use named casts on openrct2-ui/windows

tupaschoal:
I basically tried using static_cast for everything, unless it errored out, then I just went with reinterpret_cast for ease, it's easy to track the latter afterwards.

PS: It turned out bigger than I expected doing for the whole `windows` folder, sorry

Helps #11158

### DIFF
--- a/src/openrct2-ui/windows/Cheats.cpp
+++ b/src/openrct2-ui/windows/Cheats.cpp
@@ -658,26 +658,26 @@ static void window_cheats_money_mousedown(rct_window* w, rct_widgetindex widgetI
             break;
         case WIDX_MONTH_UP:
             _monthSpinnerValue++;
-            _monthSpinnerValue = std::clamp(_monthSpinnerValue, 1, static_cast<int>(MONTH_COUNT));
-            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int>(days_in_month[_monthSpinnerValue - 1]));
+            _monthSpinnerValue = std::clamp(_monthSpinnerValue, 1, static_cast<int32_t>(MONTH_COUNT));
+            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int32_t>(days_in_month[_monthSpinnerValue - 1]));
             widget_invalidate(w, WIDX_MONTH_BOX);
             widget_invalidate(w, WIDX_DAY_BOX);
             break;
         case WIDX_MONTH_DOWN:
             _monthSpinnerValue--;
-            _monthSpinnerValue = std::clamp(_monthSpinnerValue, 1, static_cast<int>(MONTH_COUNT));
-            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int>(days_in_month[_monthSpinnerValue - 1]));
+            _monthSpinnerValue = std::clamp(_monthSpinnerValue, 1, static_cast<int32_t>(MONTH_COUNT));
+            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int32_t>(days_in_month[_monthSpinnerValue - 1]));
             widget_invalidate(w, WIDX_MONTH_BOX);
             widget_invalidate(w, WIDX_DAY_BOX);
             break;
         case WIDX_DAY_UP:
             _daySpinnerValue++;
-            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int>(days_in_month[_monthSpinnerValue - 1]));
+            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int32_t>(days_in_month[_monthSpinnerValue - 1]));
             widget_invalidate(w, WIDX_DAY_BOX);
             break;
         case WIDX_DAY_DOWN:
             _daySpinnerValue--;
-            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int>(days_in_month[_monthSpinnerValue - 1]));
+            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, static_cast<int32_t>(days_in_month[_monthSpinnerValue - 1]));
             widget_invalidate(w, WIDX_DAY_BOX);
             break;
         case WIDX_DATE_SET:
--- a/src/openrct2-ui/windows/Cheats.cpp
+++ b/src/openrct2-ui/windows/Cheats.cpp
@@ -188,17 +188,17 @@ constexpr int32_t OPTW = 220;                                            // Opti
 constexpr int32_t OPTH = 10;                                             // Option (checkbox) height (two columns)
 constexpr int32_t GROUP_SPACE = 6;
 
-#define YPL(ROW)        ((int16_t)(YOS + ((BTNH + YSPA) * ROW)))
-#define HPL(ROW)        ((int16_t)(YPL(ROW) + BTNH))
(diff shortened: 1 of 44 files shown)


src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -391,7 +391,7 @@ rct_window* window_editor_object_selection_open()
 
     window->enabled_widgets = (1 << WIDX_ADVANCED) | (1 << WIDX_INSTALL_TRACK) | (1 << WIDX_FILTER_DROPDOWN)
         | (1 << WIDX_FILTER_STRING_BUTTON) | (1 << WIDX_FILTER_CLEAR_BUTTON) | (1 << WIDX_CLOSE) | (1 << WIDX_LIST_SORT_TYPE)
-        | (((uint32_t)1) << WIDX_LIST_SORT_RIDE);
+        | ((static_cast<uint32_t>(1)) << WIDX_LIST_SORT_RIDE);
```
Gymnasiast, line 394:
```suggestion
        | (1UL << WIDX_LIST_SORT_RIDE);
```

src/openrct2-ui/windows/Map.cpp
```diff
@@ -86,7 +86,7 @@ static rct_widget window_map_widgets[] = {
       SPINNER_WIDGETS      (1,  104,    198,    229,    240,    STR_MAP_SIZE_VALUE,                     STR_NONE), // NB: 3 widgets
     { WWT_FLATBTN,          1,  4,      27,     1,      24,     SPR_BUY_LAND_RIGHTS,                    STR_SELECT_PARK_OWNED_LAND_TIP },
     { WWT_FLATBTN,          1,  4,      27,     1,      24,     SPR_PARK_ENTRANCE,                      STR_BUILD_PARK_ENTRANCE_TIP },
-    { WWT_FLATBTN,          1,  28,     51,     1,      24,     (uint32_t) SPR_NONE,                      STR_SET_STARTING_POSITIONS_TIP },
+    { WWT_FLATBTN,          1,  28,     51,     1,      24,     static_cast<uint32_t>( SPR_NONE),                      STR_SET_STARTING_POSITIONS_TIP },
```
Gymnasiast, line 89: Could you fix the alignment of this line?

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1431,7 +1431,7 @@ static void window_park_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
     y += LIST_ROW_HEIGHT;
 
     // Draw number of rides / attractions
-    if (w->list_information_type != (uint16_t)-1)
+    if (w->list_information_type != static_cast<uint16_t>(-1))
```
Gymnasiast, line 1434:
```suggestion
    if (w->list_information_type != 0xFFFF)
```

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1613,7 +1613,7 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
     // Objective outcome
     if (gScenarioCompletedCompanyValue != MONEY32_UNDEFINED)
     {
-        if ((uint32_t)gScenarioCompletedCompanyValue == 0x80000001)
+        if (static_cast<uint32_t>(gScenarioCompletedCompanyValue) == 0x80000001)
```
Gymnasiast, line 1616: I prefer a constant for this.
If you don't want to do that in this PR, could you create an issue?

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -909,7 +909,7 @@ static void window_editor_object_selection_invalidate(rct_window* w)
             for (int32_t i = 0; i < 6; i++)
             {
                 if (_filter_flags & (1 << (_numSourceGameItems + i)))
-                    w->pressed_widgets |= (uint64_t)(1ULL << (WIDX_FILTER_RIDE_TAB_TRANSPORT + i));
+                    w->pressed_widgets |= static_cast<uint64_t>(1ULL << (WIDX_FILTER_RIDE_TAB_TRANSPORT + i));
```
Gymnasiast, line 912: AFAIK, 1ULL is supposed to be 64-bit. So I think we can just get rid of this cast altogether.

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -720,15 +720,15 @@ static void window_scenery_dropdown(rct_window* w, rct_widgetindex widgetIndex,
 
     if (widgetIndex == WIDX_SCENERY_PRIMARY_COLOUR_BUTTON)
     {
-        gWindowSceneryPrimaryColour = (uint8_t)dropdownIndex;
+        gWindowSceneryPrimaryColour = static_cast<uint8_t>(dropdownIndex);
```
Gymnasiast, line 723: gWindowSceneryPrimaryColour and friends are `colour_t`, so I'd like all three casts to cast to `colour_t` instead.

src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -531,7 +531,8 @@ void window_staff_list_invalidate(rct_window* w)
     {
         window_staff_list_widgets[WIDX_STAFF_LIST_UNIFORM_COLOUR_PICKER].type = WWT_COLOURBTN;
         window_staff_list_widgets[WIDX_STAFF_LIST_UNIFORM_COLOUR_PICKER].image = SPRITE_ID_PALETTE_COLOUR_1(
-                                                                                     (uint32_t)staff_get_colour(tabIndex))
+                                                                                     static_cast<uint32_t>(
+                                                                                         staff_get_colour(tabIndex)))
```
Gymnasiast, line 535: This looks very silly like this. It's better to use an intermediary variable, I'd say.

src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -658,26 +658,26 @@ static void window_cheats_money_mousedown(rct_window* w, rct_widgetindex widgetI
             break;
         case WIDX_MONTH_UP:
             _monthSpinnerValue++;
-            _monthSpinnerValue = std::clamp(_monthSpinnerValue, 1, (int)MONTH_COUNT);
-            _daySpinnerValue = std::clamp(_daySpinnerValue, 1, (int)days_in_month[_monthSpinnerValue - 1]);
+            _monthSpinnerValue = std::clamp(_monthSpinnerValue, 1, static_cast<int>(MONTH_COUNT));
```
duncanspumpkin, line 661: this should be using `int32_t` instead of `int`. Same goes for the other clamps.

tupaschoal:
Is it good to go now @Gymnasiast or @duncanspumpkin ?

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1787,7 +1787,7 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
                 // Edge texture name
                 uint32_t idx = tileElement->AsSurface()->GetEdgeStyle();
                 openrct2_assert(
-                    (uint32_t)idx < std::size(TerrainEdgeTypeStringIds),
+                    static_cast<uint32_t>(idx) < std::size(TerrainEdgeTypeStringIds),
```
Gymnasiast, line 1790: `idx` is an `uint32_t` already. You can get rid of this cast.

PR #11139, opened 2020-03-29
Use named casts on openrct2 root folder

tupaschoal:
Helps #11158

### DIFF
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -82,13 +82,13 @@ static FILE* fopen_utf8(const char* path, const char* mode)
 static void sprite_entries_make_absolute()
 {
     for (uint32_t i = 0; i < spriteFileHeader.num_entries; i++)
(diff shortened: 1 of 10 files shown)


src/openrct2/CmdlineSprite.cpp
```diff
@@ -130,13 +130,14 @@ static bool sprite_file_open(const utf8* path)
         }
 
         int32_t entryTableSize = spriteFileHeader.num_entries * sizeof(rct_g1_element);
-        spriteFileEntries = (rct_g1_element*)malloc(entryTableSize);
+        spriteFileEntries = static_cast<rct_g1_element*>(malloc(entryTableSize));
```
duncanspumpkin: There's not any point anymore to use malloc over new. This would get rid of the need for the cast. Guess it's slightly out of the scope of this pr but might be an idea to open an issue for it so that it gets tracked

tupaschoal:
I completely agree, I just didn't do it here because it was already a big task. Will create an issue later.

PR #11140, opened 2020-03-29
Fix build on FreeBSD

pkubaj:
Otherwise build fails, this will be also the case for other systems where OpenRCT2 builds, like OpenBSD.

### DIFF
--- a/src/openrct2/thirdparty/filesystem.hpp
+++ b/src/openrct2/thirdparty/filesystem.hpp
@@ -45,6 +45,8 @@
 #define GHC_OS_MACOS
 #elif defined(__linux__)
 #define GHC_OS_LINUX
+#elif defined(__FreeBSD__)
+#define GHC_OS_FREEBSD
 #if defined(__ANDROID__)
 #define GHC_OS_ANDROID
 #endif


AaronVanGeffen:
Interesting. I would expect FreeBSD to support the std::filesystem library natively. If this is indeed, the case, I would rather the /src/openrct2/core/FileSystem.hpp file be updated accordingly. 

Either way, I would prefer we don’t change anything in third-party libraries. Chances are we will forget to leave the change(s) when we inevitably update them from their upstream sources.

pkubaj:
I just looked at FileSystem.hpp. std::filesystem is definitely available, so it's a detection tool issue.

janisozaur:
How about taking the latest release from https://github.com/gulrak/filesystem/releases ?

AaronVanGeffen:
@janisozaur I recently did, see #10939 

@pkubaj Could you instead edit `src/openrct2/core/FileSystem.hpp` such that std::filesystem is used instead?

janisozaur:
Right. Well, they've added (Free)BSD support upstream across https://github.com/gulrak/filesystem/pull/49 and https://github.com/gulrak/filesystem/pull/51, but it's yet to be released.

janisozaur:
I've asked about the release plan in https://github.com/gulrak/filesystem/issues/59

AaronVanGeffen:
My point is that we shouldn't need the `ghc::filesystem` library if FreeBSD supports `std::filesystem` natively. As @pkubaj says, we should address the detection issue instead.

pkubaj:
I have modified FileSystem.hpp to look like this:
```
/*****************************************************************************
 * Copyright (c) 2014-2019 OpenRCT2 developers
 *
 * For a complete list of all authors, please refer to contributors.md
 * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
 *
 * OpenRCT2 is licensed under the GNU General Public License version 3.
 *****************************************************************************/

/// @file
/// This file enables access to std::filesystem as `fs` namespace
/// or injects third-party drop-in replacement on build configs
/// where std::filesystem is not available.

#pragma once

// Find out if std::filesystem is available:
#if defined(_MSC_VER) // Visual Studio supports <filesystem>
#    define HAVE_STD_FILESYSTEM 1
#elif defined(__APPLE__) // XCode has the header, but reports error when included.
#    define HAVE_STD_FILESYSTEM 0
#elif defined(__FreeBSD__)
#    define HAVE_STD_FILESYSTEM 1
#elif defined(__has_include) // For GCC/Clang check if the header exists.
#    if __has_include(<filesystem>)
#        define HAVE_STD_FILESYSTEM 1
#    else
#        define HAVE_STD_FILESYSTEM 0
#    endif
#else // By default assume not supported.
#    define HAVE_STD_FILESYSTEM 0
#endif

#if HAVE_STD_FILESYSTEM
#    include <filesystem>
# error 1
namespace fs = std::filesystem;
#else
#    include "../thirdparty/filesystem.hpp"
namespace fs = ghc::filesystem;
#endif

#undef HAVE_STD_FILESYSTEM // Not needed any more, don't make it public.
```

So it assumes blindly that FreeBSD supports filesystem (which is true except for a release versions of some uncommon platforms, like MIPS, SPARC64 or PPC, this will change in next version).

It also errors printing "1" after trying to include filesystem. This to check whether the value is correctly defined.

Compiling:
```
root@amd64-default:/wrkdirs/usr/ports/games/openrct2/work/.build # /usr/bin/c++ -DOPENGL_NO_LINK -D__ENABLE_LIGHTFX__ -I/usr/local/include/freetype2 -I/usr/local/include/uuid -I/usr/local/include/libpng16 -O2 -pipe -DLIBICONV_PLUG -fstack-protector-strong -isystem /usr/local/include -fno-strict-aliasing -D_GNU_SOURCE -DLIBICONV_PLUG -isystem /usr/local/include -fstrict-overflow -fstrict-aliasing -Werror -Wundef -Wmissing-declarations -Winit-self -Wall -Wextra -Wshadow -Wno-unknown-pragmas -Wno-missing-braces -Wno-comment -Wnonnull -Wno-unused-parameter -DDEBUG=0 -O2 -pipe -DLIBICONV_PLUG -fstack-protector-all -isystem /usr/local/include -fno-strict-aliasing -D_GNU_SOURCE -DLIBICONV_PLUG -isystem /usr/local/include -fPIC -x c++ -Wno-pragma-once-outside-header -Wno-unused-const-variable -x c++ -Wno-pragma-once-outside-header -Wno-unused-const-variable -std=gnu++17 -MD -MT CMakeFiles/libopenrct2-headers-check.dir/src/openrct2/core/FileSystem.hpp.o -MF CMakeFiles/libopenrct2-headers-check.dir/src/openrct2/core/FileSystem.hpp.o.d -o CMakeFiles/libopenrct2-headers-check.dir/src/openrct2/core/FileSystem.hpp.o -c /wrkdirs/usr/ports/games/openrct2/work/OpenRCT2-0.2.5/src/openrct2/core/FileSystem.hpp
/wrkdirs/usr/ports/games/openrct2/work/OpenRCT2-0.2.5/src/openrct2/core/FileSystem.hpp:36:3: error: 1
# error 1
  ^
1 error generated.
```

So it finds std::filesystem, hurray. But for some reason, gulrak's implementation is still  chosen later in the build and will cause build error.

janisozaur:
@pkubaj please review https://github.com/OpenRCT2/OpenRCT2/pull/11142 if it fixes your problem.

PR #11141, opened 2020-03-29
Use named casts on openrct2/world

tupaschoal:
Helps #11158


src/openrct2/world/TileElement.h
```diff
@@ -643,4 +644,4 @@ bool tile_element_is_underground(TileElement* tileElement);
 void tile_element_set_banner_index(TileElement* tileElement, BannerIndex bannerIndex);
 void tile_element_remove_banner_entry(TileElement* tileElement);
 
-uint8_t tile_element_get_ride_index(const TileElement* tileElement);
+uint8_t tile_element_get_ride_index(const TileElement* tileElement);
```
Gymnasiast, line 647: Revert

tupaschoal: Thanks for doing it :)

PR #11142, opened 2020-03-29
Fix #11140: exclude thirdparty headers from headers-check target

janisozaur:
ping @pkubaj


pkubaj:
This fixes one build issue, there's still one more but that's for the next PR (it's unrelated).

PR #11155, opened 2020-03-30
Avoid duplicate starting points and remove starting point with path

### LINKED ISSUE #10876: Issues with "Set starting positions for people" in scenario editor
Windows 10, 64 Bit
v0.2.4
8351a04

I have issues with the "Set starting positions for people" function in the scenario editor.

It is permitted to set a starting position on a path:
image

When I build a path and press the "Set starting positions for people" button a starting position is automatically created:
image

When I delete the path the starting position is still there:
image

The behavior is identical for manually created starting positions. 

Also, I am unable to delete starting positions with right-click (though I don't think this is necessary when the removal of a path also removes the starting position).

- [?] Reproducible in RCT2 (vanilla)?
- [No] Specific to multiplayer?

Hummel95:
To fix #10876

### DIFF
--- a/src/openrct2/actions/FootpathRemoveAction.hpp
+++ b/src/openrct2/actions/FootpathRemoveAction.hpp
@@ -95,6 +95,17 @@ public:
             map_invalidate_tile_full(_loc);
             tile_element_remove(footpathElement);
             footpath_update_queue_chains();
+
+            // Remove the spawn point (if there is one in the current tile)
+            gPeepSpawns.erase(
+                std::remove_if(
+                    gPeepSpawns.begin(), gPeepSpawns.end(),
+                    [this](const CoordsXYZ& spawn) {
+                        {
+                            return spawn.ToTileStart() == _loc.ToTileStart();
+                        }
+                    }),
+                gPeepSpawns.end());
         }
         else
         {
--- a/src/openrct2/actions/PlacePeepSpawnAction.hpp
+++ b/src/openrct2/actions/PlacePeepSpawnAction.hpp
@@ -98,6 +98,35 @@ public:
         res->Expenditure = ExpenditureType::LandPurchase;
         res->Position = _location;
 
+        // Shift the spawn point to the edge of the tile
+        auto spawnPos = CoordsXY{ _location.ToTileCentre() }
+            + CoordsXY{ DirectionOffsets[_location.direction].x * 15, DirectionOffsets[_location.direction].y * 15 };
+
+        PeepSpawn spawn;
+        spawn.x = spawnPos.x;
+        spawn.y = spawnPos.y;
+        spawn.z = _location.z;
+        spawn.direction = _location.direction;
+
+        // When attempting to place a peep spawn on a tile that already contains it,
+        // remove that peep spawn instead.
+        if (!gPeepSpawns.empty())
+        {
+            // When searching for existing spawns, ignore the direction.
+            auto foundSpawn = std::find_if(gPeepSpawns.begin(), gPeepSpawns.end(), [spawn](const CoordsXYZ& existingSpawn) {
+                {
+                    return existingSpawn.ToTileStart() == spawn.ToTileStart();
+                }
+            });
+
+            if (foundSpawn != std::end(gPeepSpawns))
+            {
+                gPeepSpawns.erase(foundSpawn);
+                map_invalidate_tile_full(spawn);
+                return res;
+            }
+        }
+
         // If we have reached our max peep spawns, remove the oldest spawns
         while (gPeepSpawns.size() >= MAX_PEEP_SPAWNS)
         {
@@ -106,16 +135,7 @@ public:
             map_invalidate_tile_full(oldestSpawn);
         }
 
-        // Shift the spawn point to the edge of the tile
-        auto spawnPos = CoordsXY{ _location.ToTileCentre() }
-            + CoordsXY{ DirectionOffsets[_location.direction].x * 15, DirectionOffsets[_location.direction].y * 15 };
-
         // Set peep spawn
-        PeepSpawn spawn;
-        spawn.x = spawnPos.x;
-        spawn.y = spawnPos.y;
-        spawn.z = _location.z;
-        spawn.direction = _location.direction;
         gPeepSpawns.push_back(spawn);
 
         // Invalidate tile
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -1776,7 +1776,7 @@ static void clear_elements_at(const CoordsXY& loc)
     gPeepSpawns.erase(
         std::remove_if(
             gPeepSpawns.begin(), gPeepSpawns.end(),
-            [x = loc.x, y = loc.y](const auto& spawn) { return floor2(spawn.x, 32) == x && floor2(spawn.y, 32) == y; }),
+            [loc](const CoordsXY& spawn) { return spawn.ToTileStart() == loc.ToTileStart(); }),
         gPeepSpawns.end());
 
     TileElement* tileElement = map_get_first_element_at(loc);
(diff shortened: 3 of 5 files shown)


src/openrct2/actions/PlacePeepSpawnAction.hpp
```diff
@@ -98,24 +98,33 @@ DEFINE_GAME_ACTION(PlacePeepSpawnAction, GAME_COMMAND_PLACE_PEEP_SPAWN, GameActi
...
         spawn.x = spawnPos.x;
         spawn.y = spawnPos.y;
         spawn.z = _location.z;
         spawn.direction = _location.direction;
+
+        // Don't allow multiple peep spawns on same tile
+        if (!gPeepSpawns.empty()){
+            if (std::find(std::begin(gPeepSpawns), std::end(gPeepSpawns), spawn) != std::end(gPeepSpawns))
+            {
+                return res;
```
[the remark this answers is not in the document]
Hummel95: I rely on the spawn object though which is currently not created in the query...

src/openrct2/actions/FootpathRemoveAction.hpp
```diff
@@ -95,6 +95,15 @@ DEFINE_GAME_ACTION(FootpathRemoveAction, GAME_COMMAND_REMOVE_PATH, GameActionRes
             map_invalidate_tile_full(_loc);
             tile_element_remove(footpathElement);
             footpath_update_queue_chains();
+
+            // Remove the spawn point (if there is one in the current tile)
+            gPeepSpawns.erase(
+                std::remove_if(
+                    gPeepSpawns.begin(), gPeepSpawns.end(),
+                    [x = _loc.x, y = _loc.y](const auto& spawn) {
+                        return floor2(spawn.x, 32) == x && floor2(spawn.y, 32) == y;
```
Gymnasiast: You should use `ToTileStart()` to floor the coordinates.

src/openrct2/world/Map.cpp
```diff
@@ -1747,7 +1747,9 @@ static void clear_elements_at(const CoordsXY& loc)
     gPeepSpawns.erase(
         std::remove_if(
             gPeepSpawns.begin(), gPeepSpawns.end(),
-            [x = loc.x, y = loc.y](const auto& spawn) { return floor2(spawn.x, 32) == x && floor2(spawn.y, 32) == y; }),
+            [x = loc.x, y = loc.y](const auto& spawn) {
+                return spawn.ToTileStart().x == x && spawn.ToTileStart().y == y;
```
Gymnasiast: If you can change line 1750 to take `const TileCoordsXY& spawn)` (I don't know if that compiles), then you could do `return spawn.ToTileStart() == loc;`

Hummel95: ...\OpenRCT2\src\openrct2\actions\FootpathRemoveAction.hpp(104,1): error C2039: 'ToTileStart': is not a member of 'TileCoordsXY' (compiling source file world\Map.cpp)

image

duncanspumpkin: You should always try to avoid passing `this` into a lambda it rarely does what you expect. I think what @Gymnasiast meant to say was do it like this (also not tested for compiliation):

```
[loc](const CoordsXY& spawn){ return spawn.ToTileStart() == loc; }
```

Hummel95: For the change in the map.cpp your code works fine. For the FootpathRemoveAction.hpp I can only get it to work with passing this (I guess it's because _loc is private?).

Gymnasiast: I wrote TileCoordsXY where I meant to write CoordsXY. Sorry about that.

src/openrct2/actions/PlacePeepSpawnAction.hpp
```diff
@@ -98,24 +98,34 @@ DEFINE_GAME_ACTION(PlacePeepSpawnAction, GAME_COMMAND_PLACE_PEEP_SPAWN, GameActi
...
         spawn.y = spawnPos.y;
         spawn.z = _location.z;
         spawn.direction = _location.direction;
+
+        // Don't allow multiple peep spawns on same tile
+        if (!gPeepSpawns.empty())
+        {
+            if (std::find(std::begin(gPeepSpawns), std::end(gPeepSpawns), spawn) != std::end(gPeepSpawns))
+            {
+                return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_ERR_CANT_PLACE_PEEP_SPAWN_HERE);
```
Gymnasiast, line 116: Rather than showing an error, it would be very handy if we could remove peep spawns by clicking existing ones. Currently there is no way to remove them easily.

Hummel95: That's above my capabilities (yet) 😃

Gymnasiast: It shouldn't be too difficult. I can take a look at it, but I can't promise any timescale for it.

tupaschoal:
I'm labelling it as WIP, given @Gymnasiast 's comment

src/openrct2/actions/PlacePeepSpawnAction.hpp
Gymnasiast: I have now made the change in question.

src/openrct2/actions/PlacePeepSpawnAction.hpp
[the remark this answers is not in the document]
Gymnasiast: This conversation is outdated, since it no longer returns an error here.

src/openrct2/actions/PlacePeepSpawnAction.hpp
```diff
@@ -98,24 +98,38 @@ DEFINE_GAME_ACTION(PlacePeepSpawnAction, GAME_COMMAND_PLACE_PEEP_SPAWN, GameActi
...
 
-        // Set peep spawn
         PeepSpawn spawn;
         spawn.x = spawnPos.x;
         spawn.y = spawnPos.y;
         spawn.z = _location.z;
         spawn.direction = _location.direction;
+
+        // When attempting to place a peep spawn on a tile that already contains it,
+        // remove that peep spawn instead. This makes it easy
```
Gymnasiast, line 112:
```suggestion
        // remove that peep spawn instead.
```

Gymnasiast:
Needs squash on merge.

@duncanspumpkin Could you re-review this?

src/openrct2/actions/FootpathRemoveAction.hpp
```diff
@@ -95,6 +95,17 @@ DEFINE_GAME_ACTION(FootpathRemoveAction, GAME_COMMAND_REMOVE_PATH, GameActionRes
...
             tile_element_remove(footpathElement);
             footpath_update_queue_chains();
+
+            // Remove the spawn point (if there is one in the current tile)
+            gPeepSpawns.erase(
+                std::remove_if(
+                    gPeepSpawns.begin(), gPeepSpawns.end(),
+                    [this](const CoordsXYZ& spawn) {
+                        {
+                            return spawn.ToTileStart() == _loc;
```
duncanspumpkin, line 105: If you are doing a compare like this use the same function on both. Can you change loc so that it uses to tile start.

src/openrct2/world/Map.cpp
```diff
@@ -1768,8 +1768,7 @@ static void clear_elements_at(const CoordsXY& loc)
     // Remove the spawn point (if there is one in the current tile)
     gPeepSpawns.erase(
         std::remove_if(
-            gPeepSpawns.begin(), gPeepSpawns.end(),
-            [x = loc.x, y = loc.y](const auto& spawn) { return floor2(spawn.x, 32) == x && floor2(spawn.y, 32) == y; }),
+            gPeepSpawns.begin(), gPeepSpawns.end(), [loc](const CoordsXY& spawn) { return spawn.ToTileStart() == loc; }),
```
duncanspumpkin, line 1771: As above both should use to tile start.

Issue #11159, opened 2020-03-30
Purge usage of malloc() and free()


tupaschoal:
OpenRCT2 historically used C and many places still use `malloc` and `free` for allocation and deallocation, which is far from ideal. The goal is to replace these with `new` and `delete`, but even better to wrap this content into one of `std::unique_ptr` or `std::shared_ptr`, using the corresponding `std::make_unique` and `std::make_shared` where applicable.

It's recommended to break into multiple PRs inside each of these folders in `src`, for ease of review and better progress:
- [ ] openrct2
- [ ] openrct2-cli
- [ ] openrct2-ui
- [ ] openrct2-win

BasJansenDev:
Sounds like a good way for me to get familiar with the code. Mind if I attempt it?

AaronVanGeffen:
This relates to our dependencies as well, most prominently the JSON library we use (cf. #9088). While the calls to malloc and free are wrapped by these dependencies, I'd rather avoid them entirely by using a C++ library.

tupaschoal:
Sure, give it a spin @BasJansenDev

IntelOrca:
@AaronVanGeffen Ideally we would use rapidjson as that is what the discord API uses but that does not use any exception handling. If you attempt to obtain a property that doesn't exist or a property of the wrong data type you will get a segfault. I'd much rather use a safer JSON library.

AaronVanGeffen:
@IntelOrca I completely agree, fast as it may be, that's off the table IMO. Instead, I suggested [JSON for Modern C++](https://github.com/nlohmann/json#examples) in #9088, as it has lush syntax combined with good error handling.

Vipulsh:
I would like to pick it up and give it a try. Will try to wrap it within smart pointers as that will be a good learning and understanding for me too

tupaschoal:
Ok, take it slow and try to start from a file you can compile and that doesn't overlap with the PR above

rzhao271:
Just realized I'm supposed to comment on an issue to claim it, though the commits above look abandoned.
I'd like to work on this issue. 
I submitted a PR above to remove some occurrences of `free()` in the `openrct2-ui` folder, but most of the `malloc()` and `free()` calls are in the `openrct2` folder.

PR #11162, opened 2020-03-30
Purge free/malloc from openrct2-win.cpp

BasJansenDev:
First attempt at a fix for #11159

### DIFF
--- a/src/openrct2-win/openrct2-win.cpp
+++ b/src/openrct2-win/openrct2-win.cpp
@@ -60,7 +60,6 @@ static char** GetCommandLineArgs(int argc, wchar_t** argvW)
 static void FreeCommandLineArgs(int argc, char** argv)
 {
     // Free argv
-    delete[] *argv;
     delete[] argv;
 }
 
@@ -68,7 +67,7 @@ static char* ConvertWideChartoUTF8(const wchar_t* src)
 {
     auto srcLen = lstrlenW(src);
     auto sizeReq = WideCharToMultiByte(CP_UTF8, 0, src, srcLen, nullptr, 0, nullptr, nullptr);
-    auto result = std::make_unique<char[]>(sizeReq+1);
+    auto result = new char[sizeReq];
     WideCharToMultiByte(CP_UTF8, 0, src, srcLen, result, sizeReq, nullptr, nullptr);
     return result;
 }
--- a/src/openrct2-win/openrct2-win.cpp
+++ b/src/openrct2-win/openrct2-win.cpp
@@ -49,7 +49,6 @@ static char** GetCommandLineArgs(int argc, wchar_t** argvW)
 {
     // Allocate UTF-8 strings
     auto argv = new char*[argc];
-    //char** argv = (char**)malloc(argc * sizeof(char*));
     // Convert to UTF-8
     for (int i = 0; i < argc; i++)
     {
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Thanks, this is what we're looking for in the context of #11159. Feel free to submit a PR with several files next time, too.

src/openrct2-win/openrct2-win.cpp
```diff
@@ -48,33 +48,29 @@ int wmain(int argc, wchar_t** argvW, [[maybe_unused]] wchar_t* envp)
 static char** GetCommandLineArgs(int argc, wchar_t** argvW)
 {
     // Allocate UTF-8 strings
-    auto argv = (char**)malloc(argc * sizeof(char*));
-    if (argv != nullptr)
+    auto argv = new char*[argc];
+    //char** argv = (char**)malloc(argc * sizeof(char*));
```
AaronVanGeffen: Please remove this comment. (Git allows us to see or roll back changes if needed.)

src/openrct2-win/openrct2-win.cpp
```diff
@@ -48,33 +48,29 @@ int wmain(int argc, wchar_t** argvW, [[maybe_unused]] wchar_t* envp)
...
+    delete[] argv*;
+    delete argv;
 }
 
 static char* ConvertWideChartoUTF8(const wchar_t* src)
 {
     auto srcLen = lstrlenW(src);
     auto sizeReq = WideCharToMultiByte(CP_UTF8, 0, src, srcLen, nullptr, 0, nullptr, nullptr);
-    auto result = static_cast<char*>(calloc(1, static_cast<size_t>(sizeReq) + 1));
+    //auto result = static_cast<char*>(calloc(1, static_cast<size_t>(sizeReq) + 1));
```
AaronVanGeffen: Please remove this comment as well.

BasJansenDev:
Will do, as far as I was aware this was the only file in the openrct2-win folder, hence there only being one to follow the structure proposed in #11159. I however did not get to test if the behavior was preserved as I work on Linux. @tupaschoal offered to build it on Windows so I'll work together with him on that.

AaronVanGeffen:
Yes, our Windows CI is failing for this PR. It looks like there's a syntax error at least, and you may need to `#include <memory>` in order to use `make_unique`.

src/openrct2-win/openrct2-win.cpp
```diff
@@ -10,7 +10,7 @@
 // Windows.h needs to be included first
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
-
+#include <memory>
```
tupaschoal: You don't need these, as you're not using anything from `memory` anymore.

This would also satisfy the code formatting tool. Make sure to revert to having the line break as before.

src/openrct2-win/openrct2-win.cpp
```diff
@@ -48,33 +48,26 @@ int wmain(int argc, wchar_t** argvW, [[maybe_unused]] wchar_t* envp)
 static char** GetCommandLineArgs(int argc, wchar_t** argvW)
 {
     // Allocate UTF-8 strings
-    auto argv = (char**)malloc(argc * sizeof(char*));
-    if (argv != nullptr)
```
tupaschoal: Any reason for dropping this check?

janisozaur: It's no longer valid. Unless you use noexcept new, it will throw with std::bad_alloc instead of returning nullptr.

BasJansenDev:
I thought it would be do-able to make this quick-fix on Linux and checking on godbolt if it would compile. I'll setup VS to properly test the building locally.

frutiemax:
Is this still in progress?

tupaschoal:
I don't think so, you can take over I'd you want

Issue #11158, opened 2020-03-30
Stop using c-style casts


tupaschoal:
Because OpenRCT2 historically used C, many places still use C-cast for doing stuff, but that's far from ideal, as it is generally more permissive than it should. The idea here is to remove all C-casts in the code in favour of one of the named casts: `static_cast`, `dynamic_cast`, `const_cast` or `reinterpret_cast`.

We can then easily evaluate the code to remove the most permissive ones, like `reinterpret_cast`, if applicable.

This is a lot of work, so it's recommended to break into multiple PRs inside each of these folders in `src`:

- [ ] openrct2
- [ ] openrct2-cli
- [x] openrct2-ui
- [x] openrct2-win

ffrogman:
I think this is resolved. I can build with -Wold-style-cast. Maybe we should add this flag to prevent any C style casts from slipping in in the future?

tupaschoal:
I think we have already (?)

Lastorder-DC:
#11493

`-Wold-style-cast` is already enabled.

tupaschoal:
Thanks for the check

PR #11163, opened 2020-03-30
Use named casts on openrct2/ride

tupaschoal:
Excluding subfolders

Helps #11158

### DIFF
--- a/src/openrct2/ride/RideRatings.cpp
+++ b/src/openrct2/ride/RideRatings.cpp
@@ -454,7 +454,8 @@ static void ride_ratings_score_close_proximity_in_direction(TileElement* inputTi
                 }
                 break;
             case TILE_ELEMENT_TYPE_PATH:
-                if (abs((int32_t)inputTileElement->base_height - (int32_t)tileElement->base_height) <= 2)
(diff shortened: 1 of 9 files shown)


src/openrct2/ride/RideRatings.cpp
```diff
@@ -454,15 +454,18 @@ static void ride_ratings_score_close_proximity_in_direction(TileElement* inputTi
...
                     proximity_score_increment(PROXIMITY_PATH_SIDE_CLOSE);
                 }
                 break;
             case TILE_ELEMENT_TYPE_TRACK:
                 if (inputTileElement->AsTrack()->GetRideIndex() != tileElement->AsTrack()->GetRideIndex())
                 {
-                    if (abs((int32_t)inputTileElement->base_height - (int32_t)tileElement->base_height) <= 2)
+                    if (abs(static_cast<int32_t>(inputTileElement->base_height)
+                            - static_cast<int32_t>(tileElement->base_height))
+                        <= 2)
```
Gymnasiast, line 468: Once this is replaced with `GetBaseZ()`, we can get rid of the casts. (Note for later)

tupaschoal: Noted, will followup

PR #11169, opened 2020-03-31
Use named casts on openrct2/rct*

tupaschoal:
Helps #11158

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -527,7 +527,7 @@ private:
                             EntryList* entries = GetEntryList(objectType);
 
                             // Check if there are spare entries available
-                            size_t maxEntries = (size_t)object_entry_group_counts[objectType];
+                            size_t maxEntries = static_cast<size_t>(object_entry_group_counts[objectType]);
                             if (entries != nullptr && entries->GetCount() < maxEntries)
                             {
                                 entries->GetOrAddEntry(objectName);
@@ -1115,7 +1115,7 @@ private:
                 rct1_vehicle* srcVehicle = &_s4.sprites[i].vehicle;
                 if (srcVehicle->x != LOCATION_NULL)
                 {
-                    Vehicle* vehicle = (Vehicle*)create_sprite(SPRITE_IDENTIFIER_VEHICLE);
+                    Vehicle* vehicle = reinterpret_cast<Vehicle*>(create_sprite(SPRITE_IDENTIFIER_VEHICLE));
                     spriteIndexMap[i] = vehicle->sprite_index;
                     vehicles.push_back(vehicle);
 
@@ -1330,7 +1330,7 @@ private:
             if (_s4.sprites[i].unknown.sprite_identifier == SPRITE_IDENTIFIER_PEEP)
             {
                 rct1_peep* srcPeep = &_s4.sprites[i].peep;
-                Peep* peep = (Peep*)create_sprite(SPRITE_IDENTIFIER_PEEP);
+                Peep* peep = reinterpret_cast<Peep*>(create_sprite(SPRITE_IDENTIFIER_PEEP));
                 spriteIndexMap[i] = peep->sprite_index;
 
                 ImportPeep(peep, srcPeep);
@@ -1341,7 +1341,7 @@ private:
             rct_sprite* sprite = get_sprite(i);
             if (sprite->generic.sprite_identifier == SPRITE_IDENTIFIER_VEHICLE)
             {
-                Vehicle* vehicle = (Vehicle*)sprite;
+                Vehicle* vehicle = reinterpret_cast<Vehicle*>(sprite);
                 FixVehiclePeepLinks(vehicle, spriteIndexMap);
             }
         }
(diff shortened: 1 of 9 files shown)


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1812,7 +1813,7 @@ class S4Importer final : public IParkImporter
             if (_s4.marketing_status[i] & CAMPAIGN_ACTIVE_FLAG)
             {
                 MarketingCampaign campaign;
-                campaign.Type = (uint8_t)i;
+                campaign.Type = static_cast<uint8_t>(i);
```
Gymnasiast: Perhaps it's better to just change the loop variable to a `uint8_t`.

src/openrct2/rct12/RCT12.cpp
```diff
@@ -483,7 +483,7 @@ bool is_user_string_id(rct_string_id stringId)
 
 std::string RCT12::RemoveFormatCodes(const std::string_view& s)
 {
-    constexpr auto RCT12_MULTIBYTE_PREFIX = (char)(uint8_t)0xFF;
+    constexpr auto RCT12_MULTIBYTE_PREFIX = static_cast<char>(static_cast<uint8_t>(0xFF));
```
Gymnasiast:
```suggestion
    constexpr char RCT12_MULTIBYTE_PREFIX = 0xFF;
```
?

src/openrct2/rct12/RCT12.h
```diff
@@ -276,6 +276,7 @@ struct RCT12TileElement : public RCT12TileElementBase
     uint8_t pad_04[4];
     template<typename TType, RCT12TileElementType TClass> TType* as() const
     {
+        // TODO: CAST-IMPROVEMENT-NEEDED
         return (RCT12TileElementType)GetType() == TClass ? (TType*)this : nullptr;
```
duncanspumpkin: static_cast<RCT12TileElementType>(GetType()) == TClass ? reinterpret_cast<TType*>(this) : nullptr;

Gymnasiast: Perhaps GetType should just return `RCT12TileElementType`, so we don't need a cast for that.

src/openrct2/rct12/RCT12.cpp
tupaschoal: I'm doing this update in batches, so I don't pay too much attention on what could actually be simpler, because it's easier to go after named casts later, than C-style casts.

tl;dr Doing your change yields a warning that becomes an error

ZehMatt: Because 0xFF is not of type char, '\xFF' would work.

src/openrct2/rct2/T6Importer.cpp
```diff
@@ -173,7 +173,9 @@ class TD6Importer final : public ITrackImporter
                 _stream.SetPosition(_stream.GetPosition() - 1);
                 _stream.Read(&t6EntranceElement, sizeof(rct_td6_entrance_element));
                 TrackDesignEntranceElement entranceElement{};
-                entranceElement.z = (t6EntranceElement.z == (int8_t)(uint8_t)0x80) ? -1 : t6EntranceElement.z;
+                entranceElement.z = (t6EntranceElement.z == static_cast<int8_t>(static_cast<uint8_t>(0x80)))
+                    ? -1
+                    : t6EntranceElement.z;
```
Gymnasiast, line 178: There **has** to be a better way. This is so ugly.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -871,7 +871,7 @@ class S6Importer final : public IParkImporter
         {
             int32_t quadIndex = sceneryEntryIndex >> 5;
             int32_t bitIndex = sceneryEntryIndex & 0x1F;
-            bool invented = (_s6.researched_scenery_items[quadIndex] & ((uint32_t)1 << bitIndex));
+            bool invented = (_s6.researched_scenery_items[quadIndex] & (static_cast<uint32_t>(1) << bitIndex));
```
Gymnasiast, line 874: `1ULL << bitIndex` would be a heck of a lot clearer and less ugly.

src/openrct2/rct12/RCT12.cpp
```diff
@@ -335,7 +335,7 @@ colour_t RCT12SmallSceneryElement::GetSecondaryColour() const
 
 bool RCT12SmallSceneryElement::NeedsSupports() const
 {
-    return (bool)(colour_1 & MAP_ELEM_SMALL_SCENERY_COLOUR_FLAG_NEEDS_SUPPORTS);
+    return static_cast<bool>(colour_1 & MAP_ELEM_SMALL_SCENERY_COLOUR_FLAG_NEEDS_SUPPORTS);
```
Gymnasiast, line 338:
```suggestion
    return (colour_1 & MAP_ELEM_SMALL_SCENERY_COLOUR_FLAG_NEEDS_SUPPORTS) != 0;
```

duncanspumpkin: I don't think you even need that it should decay fine to a bool without anything elese.

tupaschoal:
Issues are dealt with

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -841,7 +841,7 @@ class S6Importer final : public IParkImporter
         {
             int32_t quadIndex = rideType >> 5;
             int32_t bitIndex = rideType & 0x1F;
-            bool invented = (_s6.researched_ride_types[quadIndex] & ((uint32_t)1 << bitIndex));
+            bool invented = (_s6.researched_ride_types[quadIndex] & (static_cast<uint32_t>(1) << bitIndex));
```
Gymnasiast, line 844: I probably should have been clearer, but I would like you to replace all occurrences of `static_cast<uint32_t>(1)` with `1ULL` in this file (not just the one on line 874).

tupaschoal:
Found 3 other places to be simplified with `1UL`. As soon as CI is green I'll squash

PR #11171, opened 2020-03-31
Improve bound checking on scenery invention

tupaschoal:
Follow-up of #11067

### DIFF
--- a/src/openrct2/management/Research.cpp
+++ b/src/openrct2/management/Research.cpp
@@ -601,20 +601,39 @@ void ride_entry_set_invented(int32_t rideEntryIndex)
 
 bool scenery_is_invented(const ScenerySelection& sceneryItem)
 {
-    assert(sceneryItem.SceneryType < SCENERY_TYPE_COUNT);
-    return _researchedSceneryItems[sceneryItem.SceneryType][sceneryItem.EntryIndex];
+    if (sceneryItem.SceneryType < SCENERY_TYPE_COUNT)
+    {
+        return _researchedSceneryItems[sceneryItem.SceneryType][sceneryItem.EntryIndex];
+    }
+    else
+    {
+        log_warning("Invalid Scenery Type");
+        return false;
+    }
 }
 
 void scenery_set_invented(const ScenerySelection& sceneryItem)
 {
-    assert(sceneryItem.SceneryType < SCENERY_TYPE_COUNT);
-    _researchedSceneryItems[sceneryItem.SceneryType][sceneryItem.EntryIndex] = true;
+    if (sceneryItem.SceneryType < SCENERY_TYPE_COUNT)
+    {
+        _researchedSceneryItems[sceneryItem.SceneryType][sceneryItem.EntryIndex] = true;
+    }
+    else
+    {
+        log_warning("Invalid Scenery Type");
+    }
 }
 
 void scenery_set_not_invented(const ScenerySelection& sceneryItem)
 {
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
I'm the first to defend ternary operators, but I agree with @Gymnasiast here. Once they start spanning several lines, it's time to reconsider their use.

tupaschoal:
Done

src/openrct2/management/Research.cpp
```diff
@@ -601,20 +601,29 @@ void ride_entry_set_invented(int32_t rideEntryIndex)
 
 bool scenery_is_invented(const ScenerySelection& sceneryItem)
 {
-    assert(sceneryItem.SceneryType < SCENERY_TYPE_COUNT);
-    return _researchedSceneryItems[sceneryItem.SceneryType][sceneryItem.EntryIndex];
+    Guard::Assert(sceneryItem.SceneryType < SCENERY_TYPE_COUNT, "Invalid Scenery Type");
```
Gymnasiast: Move the assert to the else as well and just have it assert `false`.

tupaschoal: Done.

src/openrct2/management/Research.cpp
```diff
@@ -601,20 +601,29 @@ void ride_entry_set_invented(int32_t rideEntryIndex)
...
+        Guard::Assert(false, "Invalid Scenery Type");
+        return false;
+    }
 }
 
 void scenery_set_invented(const ScenerySelection& sceneryItem)
 {
-    assert(sceneryItem.SceneryType < SCENERY_TYPE_COUNT);
-    _researchedSceneryItems[sceneryItem.SceneryType][sceneryItem.EntryIndex] = true;
+    Guard::Assert(sceneryItem.SceneryType < SCENERY_TYPE_COUNT, "Invalid Scenery Type");
```
duncanspumpkin, line 617: I've always viewed Guard::Assert as an exit point of the program. I realise that isn't the case if you hit ignore but should we really be continuing. Wouldn't it be better to just `log_warning`

tupaschoal: Sure, I'd ditch the ignore anyway, if we could. In any case this does seem a bit out of the scope of this PR, no?

Gymnasiast: I prefer replacing Guard::Assert with `log_warning` as well.

tupaschoal: Ok, done

src/openrct2/management/Research.cpp
```diff
@@ -601,20 +601,29 @@ void ride_entry_set_invented(int32_t rideEntryIndex)
...
+        log_warning("Invalid Scenery Type");
+        return false;
+    }
 }
 
 void scenery_set_invented(const ScenerySelection& sceneryItem)
 {
-    assert(sceneryItem.SceneryType < SCENERY_TYPE_COUNT);
-    _researchedSceneryItems[sceneryItem.SceneryType][sceneryItem.EntryIndex] = true;
+    log_warning("Invalid Scenery Type");
```
Gymnasiast, line 617: Er, what?! You're warning on every single call to `scenery_set_invented()`, including for valid data.

tupaschoal: Argh, my bad

tupaschoal: I've fixed it now.

Gymnasiast:
Nice, that's three fewer places where it could potentially read and write out of bounds.

Issue #3210, opened 2016-04-01
Scenery window scrolls wrongly after selecting item


Superjustinbros:
**OS:** Windows 7

Explanation of the issue...

- [Unsure] Reproducable in RCT2 (vanilla)?
- [Occurs in both SP and MP] Multiplayer?

**Steps to reproduce:**
1. Select a scenery object on the last on one of the last rows.
2. Let the window auto-adjust itself, and when you scroll the cursor back over the  scenery window, it will crop wrong and at least one of more of the rows underneath the selected object would be blank.

**Screenshots / Video:**
![rct2 selectino problem](https://cloud.githubusercontent.com/assets/4167951/14211561/b9718fb4-f7fc-11e5-9cbc-e1344d4caef2.png)

Gymnasiast:
I have had this bug plenty of times in vanilla, so marking as `original bug`.

marcovmun:
I think that this commit solves the bug, This bug is very annoying when placing objects

PR #11178, opened 2020-03-31
Fix #11063: Exchange hack for call to update_scroll

### LINKED ISSUE #11063: Scenery window scrolling issue
**OS:** Windows 10 (x64)
**Version:** 0.2.5
**Commit/Build:** da2ba0b

When you scroll down a long scenery tab, then switch to a different tab, the scroll position often persists. This can lead to a scenery tab appearing empty if the scroll position is below the last item in the tab.

This doesn't seem to happen all the time, but often enough to be an annoyance.

**Steps to reproduce:**
1. Open scenery window
2. Open a scenery tab with a lot of items and scroll down
3. Switch to a different scenery tab, observe that it will scroll down to where you were in the previous tab

jeysbach:
Fixes #11063 

Removed an old hack and everything seems to be working correctly now.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -9,6 +9,7 @@
 - Fix: [#11005] Company value overflows.
 - Fix: [#11027] Third color on walls becomes black when saving.
 - Improved: [#11157] Slimmer virtual floor lines.
+- Fix: [#11063] Scrolling position persists when switching tabs in the scenery window.
 
 0.2.5 (2020-03-24)
 ------------------------------------------------------------------------
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -8,8 +8,8 @@
 - Fix: [#10928] File browser's date column is too narrow.
 - Fix: [#11005] Company value overflows.
 - Fix: [#11027] Third color on walls becomes black when saving.
-- Improved: [#11157] Slimmer virtual floor lines.
 - Fix: [#11063] Scrolling position persists when switching tabs in the scenery window.
+- Improved: [#11157] Slimmer virtual floor lines.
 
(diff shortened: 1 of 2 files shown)


Gymnasiast:
It would be nice if one-liner was the solution. I do want to test it though, and also check if it doesn't regress the issue that the hack aimed to solve.

An aside: it's not a problem since we will just squash it, but it looks like you have set up your local Git with no e-mail address, or with one that is different from the one on Github. (You should be able to have it work properly by setting it to `jeysbach@users.noreply.github.com` if you don't want to give away your actual e-mail address.

jeysbach:
Right, thanks for the heads up. E-mail thing should be fixed now.

I was also surprised this worked, but it seems like the hack stopped working at some point and update_scroll was not called anymore.

Gymnasiast:
OK, this seems to work for me. I have also checked if it regresses #3210, which it doesn't.

Now, two things left:

- I'd like @Umdlye to test it.
- Could you amend the changelog? `Fix: [#11063] Scrolling position persists when switching tabs in the scenery window.`

Umdlye:
Works correctly here too :)

distribution/changelog.txt
```diff
@@ -9,6 +9,7 @@
 - Fix: [#11005] Company value overflows.
 - Fix: [#11027] Third color on walls becomes black when saving.
 - Improved: [#11157] Slimmer virtual floor lines.
+- Fix: [#11063] Scrolling position persists when switching tabs in the scenery window.
```
Gymnasiast: The changelog is grouped by change type (fix/feature/improved/etc.). This needs to be moved up one row.

PR #11179, opened 2020-03-31
Fix #11072: Land and water tools work out of bounds

### LINKED ISSUE #11072: Water and Land tool shouldn't react to out of bounds actions
**OS:** Windows 10
**Version:** 0.2.5
**Commit/Build:** 357044171

(related to #828)

The Water and Land tool yield errors of land not owned and emit sounds when you try to do anything out of the map bounds. Crazily enough, this is an original bug, I found out.

- [X] Reproducible in RCT2 (vanilla)?

**Steps to reproduce:**
1. Open the land or water tool
2. Try to raise/lower the void for land tool, or raise for water tool

**Screenshots / Video:**

image image image

oli414:
This ignores land- and water-raising/lowering interactions when interacting outside of the map. 

Currently, land and water interactions outside of the map bounds apply to the last valid tile element that the cursor touched. This was inconsistent with how the land painting and scenery clearing tools work, this is fixed with this PR.
As a result #11072 is also fixed with this PR.

Edit: Sorry for all the commits, I always seem to have some trouble getting all set up for OpenRCT2 again.


tupaschoal:
You need to rebase to solve the merge conflict though

Gymnasiast:
This _looks_ all right to me, but since it doesn't a fix _bug_ per se (only some untidy behaviour), I personally prefer only merging this after the 0.2.6. But if other team members think the risk of regression is low, then I am willing to with that as well.

AaronVanGeffen:
A vote for 0.2.6 from me, sorry. Better to be on the safe side this time.

tupaschoal:
Sorry, it needs another rebase :'(
(we merged lots of stuff with the release, so it was bound to happen)

PR #11181, opened 2020-03-31
Fix #6530: Make land tools work consistently on park borders

### LINKED ISSUE #6530: Make all land tools behave the same
All tools that act on land regions behave slightly different.

1. Clear Scenery/Random tree cluster: When a region inside and outside of the park is selected, it works as specified on the region inside the park and the outside region is ignored.
2. Land/Water tool: When using the mountain option regions outside of the park area are not affected. When not using the mountain tool, you can raise or lower the land until area outside of the park is affected, then the whole action is blocked.
3. ~~Land rights: If the selected region has a single square that can not be bought, the action is blocked.~~ (fixed in #6610)

I would actually prefer if all tools behaved like 1. Thoughts?

oli414:
This PR makes it possible to adjust the water and land height inside of the park even when the area includes land that is not owned. The land that isn't owned is ignored, only when none of the land in the selected area is owned will the error message (Like: "Can't raise land here... Land not owned by park!") appear.

Fixes #6530.

Ignore out of park messages when overlapping with the inside of the park

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -4,6 +4,7 @@
 - Feature: [#11231] Change shortcut window list order to be more intuitive, and split it into logical sections.
 - Fix: [#11072] Land and water tools working out of bounds (original bug).
 - Fix: [#11315] Ride that has never opened is shown as favorite ride of many guests.
+- Improved: [#6530] Allow water and land height changes on park borders.
 
 0.2.6 (2020-04-17)
 ------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


oli414:
That doesn't look right hold on

oli414:
Alright, that's better, I shouldn't have rebased there.

AaronVanGeffen:
Rather than doing a rebase, you've merged `develop` into your branch, I'm afraid.

oli414:
@AaronVanGeffen Should I be reverting the merge in that case?

AaronVanGeffen:
@oli414 It will probably be dropped in a rebase, anyway. If it isn't, for some reason, an interactive rebase would allow you to drop the commit entirely, without reverting.

oli414:
@AaronVanGeffen Thanks! That seemed to have worked. Been a while since I have dealt with forks

tupaschoal:
This has to be rebased again 🙈 
(but just because I merged your other PR ahahaha)

Issue #11126, opened 2020-03-28
Cannot place Frightmare (stock track design)


Gymnasiast:
**OS:** Ubuntu 20.04
**Version:** 0.2.5+
**Commit/Build:** 3613c660a0b4d2bc7d81329424f525eac7e3c7e3

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Specific to multiplayer?

**Steps to reproduce:**
1. Pick the Steel Twister.
2. Try placing the "Frightmare" track design
3. Game will show the error "Can't construct this here... in the way" (sic)

![Funtopia 2020-03-28 22-51-59](https://user-images.githubusercontent.com/1478678/77834705-d5fd0600-7146-11ea-8bc3-36246cd79048.png)

Fred-1044276:
I have had this bug for a while now. And it isn't just Frightmare that is affected. Several of my other designs, even ones I created and saved, have fallen victim to this bug.

James103:
Have you tried placing the affected designs in an empty park on flat land that you fully own?

Fred-1044276:
I tried that, and it always ends up with the same result of "Can't construct this here... in the way".

Gymnasiast:
@Fred-1044276 Could you upload a handful of tracks that don't work in a ZIP? Make sure to include the ride type in the filenames to make it easier to test.

Gymnasiast:
Bisecting points to be13c1fc87eaee50c0425a843af6b89698cb8f59.

Fred-1044276:
[broken TD6s.zip](https://github.com/OpenRCT2/OpenRCT2/files/4399489/broken.TD6s.zip)
@Gymnasiast  Here is the zip file with the tracks that don't work, which are:
- CrookedPirate - Spunknit (Giga Coaster)
- Empire Express (Inverted Roller Coaster) — also a stock design
- Flying 600 6 (Flying Roller Coaster)
- Frightmare (Steel Twister/Floorless Twister Trains) — the design that is mentioned in this issue
- Garden Hyper-Twist (Hyper-Twister/Hyper-Twister Trains (wide cars))
- Hypercoaster Supertall (Hypercoaster) — built using the "Disable support limits" and "Allow chain lifts on all track pieces" cheats
- LodeRunner - Spunknit (Giga Coaster)
- TKO Floorless (Steel Twister/Floorless Twister Trains)
- Twister 5 trains (Steel Twister/Twister Trains)
[...]

PR #11183, opened 2020-03-31
Fix mistake with wall coordinates in #10875

Gymnasiast:
While investigating https://github.com/OpenRCT2/OpenRCT2/issues/11126 , I came across a mistake I made in TrackPlaceAction (sadly not the mistake that caused #11126).


Gymnasiast:
Included in #11186.

PR #10579, opened 2020-01-17
Make GetWaterHeight return big Z

Gymnasiast:

PR #11184, opened 2020-03-31
Fix mistake made in #10579

Gymnasiast:


duncanspumpkin:
Can we do a check of the codebase for narrowing conversions like this?

IntelOrca:
We probably disabled the warning probably because there are so many. Will be a good effort to go through them all.

Gymnasiast:
Is this one good to go? Does it perhaps need a network version change?

duncanspumpkin:
Yes it will need network version update.

tupaschoal:
> Can we do a check of the codebase for narrowing conversions like this?

There are plenty I'm seeing when removing C-style casts.

PR #11186, opened 2020-03-31
Fix #11126: Cannot place Frightmare track design

### LINKED ISSUE #11126: Cannot place Frightmare (stock track design)

Gymnasiast:
Was this one ever tedious to fix...

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -10,6 +10,7 @@
 - Fix: [#11005] Company value overflows.
 - Fix: [#11027] Third color on walls becomes black when saving.
 - Fix: [#11063] Scrolling position persists when switching tabs in the scenery window.
+- Fix: [#11126] Cannot place Frightmare track design.
 - Improved: [#11157] Slimmer virtual floor lines.
 
 0.2.5 (2020-03-24)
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
So what was the issue?

Gymnasiast:
In the old situation, it converted from big to small Z, did the addition, then converted it back. This basically meant it truncated the Z value to whole small Z increments. I changed the code to use big Z throughout, which also meant it no longer truncated the values in question.

duncanspumpkin:
Which field is the bad Z coming from there might be other code we need to consider that this impacts. This has always been a little worry of mine when we increased the field sizes like with the pathfinding of the boat hire.

Gymnasiast:
@duncanspumpkin I don't know what you mean. I `floor2`ed two variables (baseZ and clearanceZ) and that appeared to fix the problem.

Gymnasiast:
I have bumped the network version. Is this good to go?

duncanspumpkin:
Its good but do we know which field is the one that has the z values that are not good.

Gymnasiast:
@duncanspumpkin I'll investigate. It's either baseZ, clearanceZ or one of the fields that feeds those.

PR #11187, opened 2020-03-31
Fix format string

duncanspumpkin:
Unsure if this will actually fix it as I can't reproduce the problem. This will at least cause the issue to be ignored I hope.


janisozaur:
Let's wait a bit after #11190 is in.

duncanspumpkin:
As #11190's bug is fixed we can now merge this.

Issue #10572, opened 2020-01-14
Cheat: Allow building track at half-height intervals


bitman2049:
It would be nice if there was a cheat to disable the Z check that happens whenever track is placed between normal height intervals. Stacking track for trackitecture gets really tedious when you have to copy/paste and adjust the height for every other level. A cheat to enable building at those levels would be really helpful.

PR #11188, opened 2020-03-31
Cheat: Allow placing track at any height

bitman2049:
Add a cheat to enable building track at all height intervals.

Issue link: https://github.com/OpenRCT2/OpenRCT2/issues/10572


janisozaur:
Thanks for your PR. We're currently working on getting as many fixes as we can to do another release and it will be a while longer until we will consider merging your feature.

duncanspumpkin:
Why have this as a separate option? In what situation does someone want this but not full support clearance off

deurklink:
@duncanspumpkin Clearance checks are usually disabled on our server. We often make structures by putting many monorail walls on top of each other, for example. For that, this would be a huge timesaver. However, when building an actual roller coaster, we wouldn't want this option turned on, because it would be really annoying to find out later that you built your rollercoaster at a height where you would not be able to connect a path to it.

duncanspumpkin:
@deurklink okay if you think it is necessary to have this as a separate option I'm not going to argue.

Umdlye:
The cheat will be called "Allow building at any height" in the menu but you refer to it as building track at invalid heights, which is more descriptive IMO.

tupaschoal:
This needs rebasing, I'd also take @Umdlye 's comment into consideration

bitman2049:
I just rebased with latest

Gymnasiast:
Needs changelog entry:
`Feature: [##10572] Cheat to allow building at invalid heights.`

Otherwise: LGTM.

src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -1074,6 +1077,9 @@ static void window_cheats_rides_mouseup(rct_window* w, rct_widgetindex widgetInd
         case WIDX_ENABLE_ALL_DRAWABLE_TRACK_PIECES:
             CheatsSet(CheatType::EnableAllDrawableTrackPieces, !gCheatsEnableAllDrawableTrackPieces);
             break;
+        case WIDX_ALLOW_TRACK_PLACE_INVALID_HEIGHTS:
+            CheatsSet(CheatType::AllowTrackPlaceInvalidHeights, !gCheatsAllowTrackPlaceInvalidHeights);
```
Gymnasiast, line 1081: Put this above:
```
            if (!gCheatsAllowTrackPlaceInvalidHeights)
            {
                context_show_error(STR_WARNING_IN_CAPS, STR_THIS_FEATURE_IS_CURRENTLY_UNSTABLE);
            }
```

bitman2049:
K, made those changes and resolved any conflicts. Should be good to go

distribution/changelog.txt
```diff
@@ -14,6 +14,7 @@
 - Fix: RCT1 scenarios have more items in the object list than are present in the park or the research list.
 - Improved: [#6530] Allow water and land height changes on park borders.
 - Improved: [#11390] Build hash written to screenshot metadata.
+- Feature: [#10572] Cheat to allow building at invalid heights.
```
Gymnasiast, line 17: Should be grouped together with the other features.

bitman2049: Got it, fixed now.

distribution/changelog.txt
```diff
@@ -3,6 +3,7 @@
 - Feature: [#9029] Open doors with the tile inspector.
 - Feature: [#11231] Change shortcut window list order to be more intuitive, and split it into logical sections.
 - Feature: [#11306] Path additions are now kept when replacing the path.
+- Feature: [#10572] Cheat to allow building at invalid heights.
```
AaronVanGeffen, line 6: Keep feature ids in sequential order:

```suggestion
- Feature: [#10572] Cheat to allow building at invalid heights.
- Feature: [#11231] Change shortcut window list order to be more intuitive, and split it into logical sections.
- Feature: [#11306] Path additions are now kept when replacing the path.
```

PR #11182, opened 2020-03-31
Fix #11166: Null pointer read in Guest::UpdateRideAdvanceThroughEntrance

Gymnasiast:

PR #11144, opened 2020-03-29
Fix #11128: null tile element in track_block_get_previous

janisozaur:

PR #11090, opened 2020-03-27
Fix #11084: null broken vehicle

janisozaur:

PR #11190, opened 2020-03-31
Upload park recording in case of crash

janisozaur:


janisozaur:
The filename argument should be changed to accept path instead and it should be caller's responsibility to extend it with replay directory path if needed.

AaronVanGeffen:
Is this intended to be merged, or just for testing purposes? I imagine there's quite a performance penalty to always recording.

ZehMatt:
How will it recover from the crash to start a recording?

duncanspumpkin:
@ZehMatt I think the idea is that it is always recording. We will need to be careful that we don't merge all of the fixes before this is merged in and tested as otherwise we wont get a crash to trigger the upload.

janisozaur:
Yes, I intend to merge it. `USE_BREAKPAD` is defined for all our current CI builds that have this library enabled, i.e. Windows MSVC ones. They generate enough reports as is. There's little performance penalty to recording of actions and the primary drop is from getting the sprite checksums. I chose to do this every 100th tick, or each 2.5s (assuming the logic runs at pretty constant 40Hz) and when ran through profiling it amounted to 0.1% cost (using `mci.sv6` from one of the title sequences, which is utilising about a quarter of all available sprites): https://gist.github.com/janisozaur/334ddd64d504bbd292de0d745ea40e60#file-perf-txt-L102

@ZehMatt 

> *the* crash to start a recording

Are you saying there is one? In that case it's best to address it first. If you meant _a_ crash, then it will fail to stop recording and not upload it (the recording), but the dump should still be. If we somehow break it so it won't even start, if it it's a logic error on our side and all the players are affected, we will surely be notified. We could also change change the format a bit to make the compression optional and flush the recording on every action/sprite addition to uncompressed file. That would serve to improve crash-proofness of code here.

@duncanspumpkin indeed, I was even thinking of taking some of the fixes out like https://github.com/OpenRCT2/OpenRCT2/pull/11090 or https://github.com/OpenRCT2/OpenRCT2/pull/11144

janisozaur:
I tested CI builds, which reported https://github.com/OpenRCT2/OpenRCT2/issues/11199 and a future one too, replay correctly. I've also modifed backtrace's reporting to include presence of sv6r in the issues it posts

src/openrct2/ReplayManager.cpp
```diff
@@ -98,6 +98,8 @@ namespace OpenRCT2
         static constexpr uint16_t ReplayVersion = 3;
         static constexpr uint32_t ReplayMagic = 0x5243524F; // ORCR.
         static constexpr int ReplayCompressionLevel = 9;
+        static constexpr int NormalRecordingChecksumTicks = 1;
+        static constexpr int SilentRecordingChecksumTicks = 100; // Same as network server
```
janisozaur, line 102: I'm tempted to lower this number a bit. Thoughts?

IntelOrca:
Do these recordings ever get deleted? You write to the result `tmpnam` and that isn't necessarily going to be tmpfs, certainly not for Windows.

src/openrct2/Game.cpp
```diff
@@ -845,3 +845,44 @@ void game_load_or_quit_no_save_prompt()
...
+    replayManager->GetCurrentReplayInfo(info);
+
+    if (replayManager->StopRecording())
+    {
+        const char* logFmt = "Replay recording stopped: (%s) %s\n"
+                             "  Ticks: %u\n"
+                             "  Commands: %u\n"
+                             "  Checksums: %u";
+
+        log_info(logFmt, info.Name.c_str(), info.FilePath.c_str(), info.Ticks, info.NumCommands, info.NumChecksums);
```
IntelOrca, line 882: This could get annoying to see in the console, I would prefer `std::printf` to be used, just so you don't constantly see the source filename and line number.

src/openrct2/ReplayManager.h
```diff
@@ -35,18 +35,27 @@ namespace OpenRCT2
     interface IReplayManager
     {
     public:
+        enum class RecordType
+        {
+            NORMAL = 0,
```
IntelOrca, line 40: ` = 0` is redundant.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1366,6 +1368,14 @@ static int32_t cc_replay_startrecord(InteractiveConsole& console, const argument
 
     std::string name = argv[0];
 
+    if (name.size() < 5 || name.substr(name.size() - 5) != ".sv6r")
```
IntelOrca, line 1371: Extension check is not case insensitive.

janisozaur:
@IntelOrca good point about temp-ness of files. I actually thought I was using `tmpfile`, not `tmpnam`, will address this.

janisozaur:
Thinking even more about this, I will just re-use a single file inside OpenRCT2's directory. This is partially because apparently using `tmpfile` requires elevated privileges on Windows.

I won't be able to finish this today, if anyone wants to pick it up before I finish up tomorrow evening, please do.

src/openrct2/interface/InteractiveConsole.cpp
janisozaur: done

src/openrct2/ReplayManager.h
janisozaur: done

src/openrct2/Game.cpp
janisozaur: done

src/openrct2/ReplayManager.cpp
janisozaur: I changed it to 40 and as expected, impacted increased from 0.1 to 0.23%

janisozaur:
List of reverted fixes, to be re-merged once root-caused:
- [ ] https://github.com/OpenRCT2/OpenRCT2/pull/11078
- [ ] https://github.com/OpenRCT2/OpenRCT2/pull/11090
- [ ] https://github.com/OpenRCT2/OpenRCT2/pull/11144
- [ ] https://github.com/OpenRCT2/OpenRCT2/pull/11182

janisozaur:
Please don't squash-merge this one, due to the reverts. Use regular merge instead.

janisozaur:
I think you're missing the point of this change. The reverts are so that we get reports and can _fix it_ properly, at the root cause, not just address the symptoms.

Gymnasiast:
@janisozaur I'm not missing the point at all. That's why I said I'm okay with reverting them _after_ the 0.2.6 release, so we can then get bug reports with recordings and fix them properly.

I don't want them in the 0.2.6 release because it's going to mean 6 months of crashes that these fixes prevent. The same happened with the 0.2.4 release: crash and bug reports kept coming in, even when the issues in question had long been fixed in develop.

janisozaur:
The plan is to have this merged, wait a few days until we get record that can be used to address the root bug, then once it's fixed merge those fixes back. Afterwards, we still want to change backtrace.io token so we don't get too crashes from old releases. All this is is meant to be included in v0.2.6.

Gymnasiast:
If we can expect the issues to be fixed properly before 0.2.6 goes out, then I have no objection.

PR #11206, opened 2020-04-01
Fix #9029: Set animation frame on doors using the tile inspector

### LINKED ISSUE #9029: Feature: Open doors and portcullis doors with tile inspector
**OS:** Windows 10
**Version:** 0.2.2
**Commit/Build:** dc90b28

To open doors, we now have to build a door, let a ride drive through it and remove the ride when the door is open. Then the open door can be copied with tile inspector and placed wherever we want.

It would save a lot of work if we could just click a doorway or portcullis door with tile inspector, and set the open state from there.

image

oli414:
This PR adds a spinner to the tile inspector that allows you to set the animation frame on certain wall pieces. 
This makes it possible to open doors without the need for creating a ride as is explained in #9029. 
The spinner showing the animation frame is seen on all wall-type tile elements, but is currently only enabled for walls with the is_door flag (Currently that is the only instance of an animation that is frozen that I know of).

Tile inspector animation frame

Tile inspector disabled animation frame

I expect this to be a post 0.2.6 addition, so for now:
- [x]  Test multiplayer.
- [x]  Add changelog entry.
- [x]  Increase network version.
- [ ]  Add an issue on the language repo for the "Animation frame" label.


src/openrct2/world/TileInspector.cpp
```diff
@@ -742,6 +742,32 @@ GameActionResult::Ptr tile_inspector_wall_set_slope(
...
+
+    if (isExecuting)
+    {
+        uint8_t animationFrame = wallElement->AsWall()->GetAnimationFrame();
+        wallElement->AsWall()->SetAnimationFrame(animationFrame + animationFrameOffset);
+
+        map_invalidate_tile_full(loc);
+
+        rct_window* const tileInspectorWindow = window_find_by_class(WC_TILE_INSPECTOR);
+        if (tileInspectorWindow != nullptr && (uint32_t)(loc.x / 32) == windowTileInspectorTileX
```
duncanspumpkin, line 761: I assume you have copied this from somewhere else? I don't see the point in performing all these checks I would just invalidate the TileInspector not much to be gained from checking that it is hovering over the correct tile.

oli414: Yes, the other tile inspector game actions do it as well, it's meant for multiplayer, so it only invalidates the tile inspector windows for the users that have the tile selected.

oli414:
@duncanspumpkin that should do!

tupaschoal:
Needs rebase again (and might require another depending on which of yours gets merged first :()

oli414:
@tupaschoal luckily it is just the changelog

AaronVanGeffen:
By rebasing we don’t mean merging develop back in. This resolves the conflicts, yes, but we want to prevent the extraneous merge conflicts.

duncanspumpkin:
We can just squash merge this one since its not that big but for future try to do rebase's.

oli414:
Oh okay, since it was just the changelog I figured I could just do it through Github, will do in the future.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1652,14 +1672,25 @@ static void window_tile_inspector_invalidate(rct_window* w)
...
+            w->widgets[WIDX_WALL_SPINNER_ANIMATION_FRAME].top = GBBT(propertiesAnchor, 2) + 3;
+            w->widgets[WIDX_WALL_SPINNER_ANIMATION_FRAME].bottom = GBBB(propertiesAnchor, 2) - 3;
+            w->widgets[WIDX_WALL_SPINNER_ANIMATION_FRAME_INCREASE].top = GBBT(propertiesAnchor, 2) + 4;
+            w->widgets[WIDX_WALL_SPINNER_ANIMATION_FRAME_INCREASE].bottom = GBBB(propertiesAnchor, 2) - 4;
+            w->widgets[WIDX_WALL_SPINNER_ANIMATION_FRAME_DECREASE].top = GBBT(propertiesAnchor, 2) + 4;
+            w->widgets[WIDX_WALL_SPINNER_ANIMATION_FRAME_DECREASE].bottom = GBBB(propertiesAnchor, 2) - 4;
             const auto wallType = tileElement->AsWall()->GetEntryIndex();
             const rct_wall_scenery_entry wallEntry = get_wall_entry(wallType)->wall;
             const bool canBeSloped = !(wallEntry.flags & WALL_SCENERY_CANT_BUILD_ON_SLOPE);
+            const bool hasAnimation = wallEntry.flags & WALL_SCENERY_IS_DOOR;
```
duncanspumpkin, line 1684: Use the `flags2 & WALL_SCENERY_2_ANIMATED` instead of checking to see if its a door. Then this can be used for any other animated walls (not that I know of any that exist).

oli414: My logic here is that the door is currently the only type of animated scenery that has a frozen animation. Otherwise, I'd add this feature for any animating objects, but for now, there isn't a reason to do so since they immediately move on anyways.

duncanspumpkin: Is it possible people might want to set up a park so that other non frozen animations are all at a certain frame? (This would be done by pausing then setting all of the frame numbers.)

oli414: I suppose there might be someone on NE who would want that. I'd have to make some drastic changes to the PR, perhaps something to expand upon later?

duncanspumpkin: okay lets consider it for the future.

duncanspumpkin:
You'll need to increment the network version one more time. Then this is good to go.

PR #11216, opened 2020-04-03
Change backtrace.io submission token

janisozaur:
This is to cut off v0.2.5 traces of which we have plenty and collect
ones that have #11190 in them.


janisozaur:
Both are active at this time, we can also change it in the future if needed.

janisozaur:
I've disabled the old token now. We still have the option to re-enable it

PR #11218, opened 2020-04-03
Load_Park feature for headless server

Juli199696:
New feature "load_park name" from "save" folder. Allows to load a new map while playing with headless server. Clients won't disconnect anymore as no server restart is needed. 
The save files need to be in the "save" directory to get detected by the server.

### DIFF
--- a/src/openrct2/interface/InteractiveConsole.cpp
+++ b/src/openrct2/interface/InteractiveConsole.cpp
@@ -39,6 +39,7 @@
 #include "../object/ObjectManager.h"
 #include "../object/ObjectRepository.h"
 #include "../peep/Staff.h"
+#include "../platform/platform.h"
 #include "../ride/Ride.h"
 #include "../ride/RideData.h"
 #include "../util/Util.h"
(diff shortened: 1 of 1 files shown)


src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1687,6 +1701,7 @@ static constexpr const console_command console_command_table[] = {
                                     "Loading a scenery group will not load its associated objects.\n"
                                     "This is a safer method opposed to \"open object_selection\".",
                                     "load_object <objectfilenodat>" },
+    { "load_park", cc_load_park, "Load park", "load_park [name]" },
```
janisozaur, line 1704: Please mention it will read a file from user's `save` directory.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1316,6 +1317,19 @@ static int32_t cc_for_date([[maybe_unused]] InteractiveConsole& console, [[maybe
...
 }
 
+static int32_t cc_load_park([[maybe_unused]] InteractiveConsole& console, [[maybe_unused]] const arguments_t& argv)
+{
+    if (argv.size() > 0)
+    {
+        char savePath[MAX_PATH];
+        platform_get_user_directory(savePath, "save", sizeof(savePath));
+        safe_strcat_path(savePath, argv[0].c_str(), sizeof(savePath));
+        path_append_extension(savePath, ".sv6", sizeof(savePath));
```
janisozaur, line 1327: It should allow user to include `.sv6` in the name. Personally I would even expect a filename here, without altering it. I guess it should also allow for `.sc6`. I recently was told there's `String::EndsWith` function that can be used for that.

src/openrct2/interface/InteractiveConsole.cpp
janisozaur: Why is the name marked as optional? Makes no sense to me without the park name and it will fail anyway.

Juli199696: This was a copy from save_park. I did not change that brackets. Sorry. should be just "nameofsave".

src/openrct2/interface/InteractiveConsole.cpp
Juli199696: Gonna add that soon and will do another PR.

AaronVanGeffen: This PR won't be merged until after the v0.2.6 release, so please amend this PR instead.

PR #11221, opened 2020-04-04
Create Sprite Refactor

duncanspumpkin:
This refactor builds a vector based sprite list that is ultimately the same as the existing linked list sprite list. The idea is that code can be moved over to use the new sprite lists.

I am planning on allocating memory for each CreateSprite rather than a bulk lump of contiguous memory. As there would be no need to store anything for free sprite id's this refactor gets ready for this future and splits off free sprite id's to a separate vector.

The two structures (linked list and vector) hold the same data but the order will not be the same. The vector is guaranteed to be ordered by sprite_id. The linked list can be in any order. For this reason they cannot at this time be used interchangeably. The advantage of the vector being ordered is that there is no need to save the list as it can be easily rebuilt (see `ResetSpriteLists()`). 

Before killing off the existing sprite lists we will need a way to rebuild the linked list sprite list for S6 export. This may be added in this PR or in a later PR.

After killing off the linked lists the `next, previous, linked_list_index` will no longer be used within the sprite struct. You will also be able to iterate through a sprite list using a ranged for loop `for (auto &peep : GetSpriteList(SPRITE_LIST_PEEP))`.

### DIFF
--- a/src/openrct2/actions/StaffHireNewAction.hpp
+++ b/src/openrct2/actions/StaffHireNewAction.hpp
@@ -158,8 +158,6 @@ private:
         }
         else
         {
-            move_sprite_to_list(newPeep, SPRITE_LIST_PEEP);
-
             newPeep->sprite_identifier = 1;
             newPeep->window_invalidate_flags = 0;
             newPeep->action = PEEP_ACTION_NONE_2;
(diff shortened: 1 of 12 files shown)


tupaschoal:
Needs to be rebased

duncanspumpkin:
This requires a rethink for how it will work after the issues with deleting elements in the sequence whilst iterating over it. Will close for now.

PR #11224, opened 2020-04-04
Add missing Expenditure type setting in LargeSceneryPlaceAction

janisozaur:
Claiming first bounty of #11190

### DIFF
--- a/src/openrct2/actions/LargeSceneryPlaceAction.hpp
+++ b/src/openrct2/actions/LargeSceneryPlaceAction.hpp
@@ -211,6 +211,7 @@ public:
     {
         auto res = std::make_unique<LargeSceneryPlaceActionResult>();
         res->ErrorTitle = STR_CANT_POSITION_THIS_HERE;
+        res->Expenditure = ExpenditureType::Landscaping;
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Is landscaping the correct expenditure type

janisozaur:
I copied it from query. I have also checked all the other actions to see if they set out correctly - they do.

janisozaur:
Small scenery uses the same: https://github.com/OpenRCT2/OpenRCT2/blob/12ab701a0fc89e1e12a3724ebef164ae4c111172/src/openrct2/actions/SmallSceneryPlaceAction.hpp#L428

janisozaur:
It seems that it is the best fit: https://github.com/OpenRCT2/OpenRCT2/blob/518780c3e805329489094a688b7851c163a47406/src/openrct2/management/Finance.h#L15-L32

duncanspumpkin:
okay sounds good just wanted to check.

Gymnasiast:
Did this cause a crash because the field contained garbage data?

duncanspumpkin:
It would have tried to set the expenditure as `ExpenditureType::Count` which would I think walk over the end of an array.

PR #11226, opened 2020-04-05
Fix track design replays.

duncanspumpkin:
The replay state was not being passed to all sub actions this prevented the replay from working. Note replays made before this still work its just they couldn't be replayed.

This is a quick and dirty fix. In the future I would like to refactor everything so that this additional global is not required.


duncanspumpkin:
There was also a small mistake with large scenery that this fixes. That will require a network version bump.

PR #11228, opened 2020-04-05
Fix replay manager checksums

duncanspumpkin:
This will likely fail the CI's on some of the replays as we haven't been checking the checksums.

`_nextChecksumTick` is a variable only used when recording so it is never incremented and so `gCurrentTicks` is only equal on the very first tick.

### DIFF
--- a/src/openrct2/ReplayManager.cpp
+++ b/src/openrct2/ReplayManager.cpp
@@ -388,9 +388,9 @@ namespace OpenRCT2
             GameStateCompareData_t cmpData = snapshots->Compare(replaySnapshot, localSnapshot);
 
             // Find out if there are any differences between the two states
-            auto res = std::find_if(cmpData.spriteChanges.begin(), cmpData.spriteChanges.end(), [](const GameStateSpriteChange_t& diff) {
-                return diff.changeType != GameStateSpriteChange_t::EQUAL;
-            });
+            auto res = std::find_if(
+                cmpData.spriteChanges.begin(), cmpData.spriteChanges.end(),
+                [](const GameStateSpriteChange_t& diff) { return diff.changeType != GameStateSpriteChange_t::EQUAL; });
 
             // If there are difference write a log to the desyncs folder
             if (res != cmpData.spriteChanges.end())
--- a/src/openrct2/ReplayManager.cpp
+++ b/src/openrct2/ReplayManager.cpp
@@ -91,8 +91,7 @@ namespace OpenRCT2
         std::multiset<ReplayCommand> commands;
         std::vector<std::pair<uint32_t, rct_sprite_checksum>> checksums;
         uint32_t checksumIndex;
-        MemoryStream startGameStateSnapshot;
-        MemoryStream endGameStateSnapshot;
+        MemoryStream gameStateSnapshots;
     };
 
     class ReplayManager final : public IReplayManager
@@ -206,6 +205,17 @@ namespace OpenRCT2
             }
         }
 
+        void TakeGameStateSnapshot(MemoryStream& snapshotStream)
+        {
+            IGameStateSnapshots* snapshots = GetContext()->GetGameStateSnapshots();
+
+            auto& snapshot = snapshots->CreateSnapshot();
+            snapshots->Capture(snapshot);
+            snapshots->LinkSnapshot(snapshot, gCurrentTicks, scenario_rand_state().s0);
+            DataSerialiser snapShotDs(true, snapshotStream);
+            snapshots->SerialiseSnapshot(snapshot, snapShotDs);
+        }
+
         virtual bool StartRecording(
             const std::string& name, uint32_t maxTicks /*= k_MaxReplayTicks*/, RecordType rt /*= RecordType::NORMAL*/) override
         {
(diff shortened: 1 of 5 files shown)


janisozaur:
There's `normalize` command that I think is supposed to address this issue, but I've never used it

janisozaur:
https://github.com/OpenRCT2/OpenRCT2/wiki/Replay-System#normalisation

duncanspumpkin:
Normalise will not fix the issue as i can't get a consistent sprite checksum.

janisozaur:
Oh dear. Isn't the whole point of replays to provide consistent checksums? What's the source of the desync?

duncanspumpkin:
I haven't worked that out yet.

janisozaur:
Some of them pass:
```
[ RUN      ] Replay/ReplayTests.RunReplay/WallRemoveAction
WARNING[../src/openrct2/ReplayManager.cpp:683 (Serialise)]: Replay network version mismatch: '0.2.3-4', expected: '0.2.5-10'
For security reasons, it is not recommended to run OpenRCT2 with elevated permissions.
[       OK ] Replay/ReplayTests.RunReplay/WallRemoveAction (129 ms)
[ RUN      ] Replay/ReplayTests.RunReplay/WaterLowerAction
WARNING[../src/openrct2/ReplayManager.cpp:683 (Serialise)]: Replay network version mismatch: '0.2.3-4', expected: '0.2.5-10'
For security reasons, it is not recommended to run OpenRCT2 with elevated permissions.
[       OK ] Replay/ReplayTests.RunReplay/WaterLowerAction (123 ms)
[ RUN      ] Replay/ReplayTests.RunReplay/WaterRaiseAction
WARNING[../src/openrct2/ReplayManager.cpp:683 (Serialise)]: Replay network version mismatch: '0.2.3-4', expected: '0.2.5-10'
For security reasons, it is not recommended to run OpenRCT2 with elevated permissions.
[       OK ] Replay/ReplayTests.RunReplay/WaterRaiseAction (134 ms)
```

duncanspumpkin:
@janisozaur very simple maps works fine. Its just that anything complex seems to fail. I'll not merge this until I work out why.

duncanspumpkin:
Any map with a vehicle has mismatching sprite checksums. Not sure why.

janisozaur:
https://github.com/OpenRCT2/OpenRCT2/blob/919a70361b787b9660220be16a97c65c83faa0d1/src/openrct2/ReplayManager.cpp#L190-L202

Normalisation only takes actions into account. I don't think that's what we want - for bpb1000Ticks there are zero actions and the updated recording is empty.

I would also like to complain about how long it took me to figure out that while replays are by default taken from `replay/` dir, the updated one is created in CWD, even though `replay_startrecord` places them in `replay/` as well. :angry:

src/openrct2/GameStateSnapshots.cpp
```diff
@@ -233,7 +233,7 @@ struct GameStateSnapshots : public IGameStateSnapshots
         COMPARE_FIELD(Peep, nausea_tolerance);
         COMPARE_FIELD(Peep, window_invalidate_flags);
         COMPARE_FIELD(Peep, paid_on_drink);
-        for (int i = 0; i < PEEP_MAX_THOUGHTS; i++)
+        for (int i = 0; i < 16; i++)
```
Gymnasiast, line 236: Why not use a constant or `std::size`? (I know `PEEP_MAX_THOUGHTS` was the wrong constant.)

tupaschoal: I'm also wondering this.

duncanspumpkin: probably because there isn't a constant for it right now.

Gymnasiast: There is still std::size.

duncanspumpkin:
Fixes to non replay code have been moved to #11246 as this requires remaking a lot of the replays.

src/openrct2/ReplayManager.cpp
```diff
@@ -382,6 +381,10 @@ namespace OpenRCT2
                 return false;
             }
 
+            // _mode is checked by some sub functions of LoadReplayDataMap
+            if (_mode != ReplayMode::NORMALISATION)
+                _mode = ReplayMode::PLAYING;
+
             if (!LoadReplayDataMap(*replayData))
             {
                 log_error("Unable to load map.");
```
duncanspumpkin, line 390: reset the mode to not playing here.

duncanspumpkin:
Remember to remove spatial index as it is no longer required.

janisozaur:
I would prefer the format allowed arbitrary snapshots, just like we do with checksums, even if we take only two. This is so we could make the replays denser and pinpoint the problems more easily in the future.

duncanspumpkin:
@janisozaur it is already designed to handle multiple snapshots. You would just add them to the memory stream.

src/openrct2/ReplayManager.cpp
```diff
@@ -379,6 +433,8 @@ namespace OpenRCT2
 
             gCurrentTicks = replayData->tickStart;
 
+            LoadAndCompareSnapshot(replayData->gameStateSnapshots);
```
janisozaur, line 436: @duncanspumpkin You only ever load snapshots at predefined places. While the stream may hold more than two snapshots, you don't have a way of knowing where (in terms of ticks) they were taken. There might be such info embedded in the state itself, but I don't see how it could be consumed as it is now. That makes me believe the current implementation is not sufficient for holding arbitrary amounts of snapshots.

The call to LinkSnapshots has odd arguments, perhaps there is some magic happening inside there, but I can't tell.

duncanspumpkin: The first uint32_t of each snapshot is the tick number. If you wanted to use more you could take deserialise all of them using the existing deserialiser converting them into an array of `GameStateSnapshot_t`. You can then do what is needed.

janisozaur: OK, that sounds like indeed extending the replays with more snapshots could be done without breaking the format.

PR #11229, opened 2020-04-05
Fix #11147: incorrect saves in backtrace

### LINKED ISSUE #11147: Corrupt park data
Extracted from #11090:

So… that one is a bit odd. While the game no longer crashes, there still must've been something that corrupted a lot of elements.

Using the park from #11132, a duplicate, see this screenshot from backtrace: 
screenshot(10) and the park attached: 
[park_11132.sv6.gz](https://github.com/OpenRCT2/OpenRCT2/files/4399711/park_11132.sv6.gz)

But once fixed it looks like this (scaled from giant screenshot): 
Volcania 2020-03-29 22-14-52_scaled

Full screenshot (large):

[Volcania 2020-03-29 22-14-52](https://user-images.githubusercontent.com/550290/77859639-de1e7980-720a-11ea-961f-56c5e2eebd91.png)

I recall seeing some issues regarding such misplaced elements, perhaps the vehicle in question also got overwritten and was not the root cause, but a symptom instead?

_Originally posted by @janisozaur in https://github.com/OpenRCT2/OpenRCT2/pull/11090#issuecomment-605694935_

duncanspumpkin:
Fix #11147 

This is the source of the corrupted saves in backtrace. Backtrace does not call the normal scenario_save function and therefore never triggers the reorg function. Same goes for replays.

### DIFF
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -196,6 +196,8 @@ void S6Exporter::Export()
     _s6.scenario_srand_0 = state.s0;
     _s6.scenario_srand_1 = state.s1;
 
+    // Map elements must be reorganised prior to saving otherwise save may be invalid
+    map_reorganise_elements();
     ExportTileElements();
     ExportSprites();
     ExportParkName();


duncanspumpkin:
Network saves already call this. Only replays and backtrace do not call this. The only other way of doing this would be to change it so that it's reorganised as it's converted into s6.

Gymnasiast:
Regular saves call it too - the somewhat confusingly named `scenario_save()` (which is also used for regular saves, not just scenarios) calls it before calling the `Export` function.

IntelOrca:
You could reorganise the map elements after exporting it to the S6 buffer so you aren't changing the game state as well.

Gymnasiast:
> You could reorganise the map elements after exporting it to the S6 buffer so you aren't changing the game state as well. It could actually be as simple as just looping through every tile element stack in coordinate order as you are exporting to S6.

Is there any advantage to that? It frequently calls the reorganise function during normal gameplay as well - not sure if there's any harm.

duncanspumpkin:
The advantage would be no need to do another memory allocation whilst crashing. But then I'm not sure if that is much of an advantage.

janisozaur:
For current saves/dumps, is there anything we can do at load time? To recognise the problem and address it there? If calling reorg twice results in the very same structure and it is saved as such to SV6, is it safe to call it once more right after the import?

duncanspumpkin:
@janisozaur we don't save the information (TileElementPointers) required to rebuild it on load.

Issue #11223, opened 2020-04-04
Shortcut keys window improvements


Umdlye:
I had a few ideas for usability improvements to the shortcut keys window. Your mileage may vary, discussion is welcome.

- [X] _Right click entry to clear shortcut_
Currently there is no way (that I know of) to clear shortcut keys other than overwriting it with another key. Right clicking the shortcut entry to clear it seems appropriate, the tooltip could be updated to reflect this.

- [X] _Add confirmation window to 'reset keys' button_
The button currently resets all of your shortcut keys without warning, AFAIK it is common practice to require confirmation for stuff like that.

- [ ] _Add shortcut keys window to options toolbar dropdown_
Giving the shortcuts window its own entry in the dropdown (below 'options' perhaps) would increase its visibility and make editing shortcuts during gameplay more convenient.

- [x] _Reorder shortcuts and add dividers_
Half of the shortcuts are in an arbitrary order, to make the window easier to read it could do with some reorganizing, preferably with lines to divide categories if at all possible. This seems like the most logical order to me, mostly based on the order of the top toolbar (much of the list is currently already sorted that way):

```
Close top-most
Close all
Cancel construction mode
Toggle visibility of toolbars
----------
Pause
Game speed -
Game speed +
Load game
Save game
Options
Screenshot
Mute
Cheats
----------
Zoom -
zoom +
Rotate clockwise
Rotate counterclockwise
Show map
----------
Clear scenery
Adjust land
Adjust water
Build scenery
Build paths
Build ride
----------
Financial
Research
Rides
Park
Guest
Staff
Recent messages
----------
Multiplayer window
Send message
----------
Underground view
Base land
Vertical faces
See through rides
See through scenery
See through paths
Invisible supports
Invisible people
Height marks on land
Height marks on ride tracks
Height marks on paths
Cut-away view
Highlight path issues
Gridlines display toggle
----------
Open scenery picker
Rotate construction object
Ride construction: ...
----------
Scroll map ...
----------
Toggle windowed mode
Scale factor -
Scale factor +
----------
Open tile inspector
Tile inspector: ...
----------
Advance to next tick
Track drawing options
Paint debug
```

AaronVanGeffen:
I think these are all good suggestions.

[...]

PR #11231, opened 2020-04-05
Change shortcut window list order

AaronVanGeffen:
This changes the order of shortcuts displayed in the key binding window. Accomplished as follows:

* The list order in the shortcut window no longer depends on the internal implementation order.
* Added functionality for adding separators to the new list.
* Reordered shortcuts per @Umdlye's suggestions in #11223.

To close off, a few screenshots of the improved window in action. (It has been resized for screenshot purposes; the initial dimensions have not been changed.)

Over The Edge 2020-04-05 11-57-43
Over The Edge 2020-04-05 11-57-49
Over The Edge 2020-04-05 11-57-59


IntelOrca:
I don't think these have to be done as part of this PR, since this doesn't change the file format. But I was thinking:

For the plugin system, I would like to allow custom shortcuts to be registered. For this I was thinking of making the shortcut IDs strings instead and storing them in a new text based file format (json or yaml) where we can preserve all the shortcuts even if you haven't loaded the plugin that registered it.

Additional to that, I would like to allow multiple bindings per shortcut and for mouse buttons to be possible bindings so that you can bind mouse 3 and 4 (usually the navigational buttons on the mouse).

AaronVanGeffen:
These are great ideas. Could you also post them to the discussion in #11223? As you said, these are a little out of scope for this PR.

src/openrct2-ui/windows/ShortcutKeys.cpp
```diff
@@ -283,6 +328,7 @@ static void window_shortcut_scrollmouseover(rct_window* w, int32_t scrollIndex,
 static void window_shortcut_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32_t scrollIndex)
 {
     gfx_fill_rect(dpi, dpi->x, dpi->y, dpi->x + dpi->width - 1, dpi->y + dpi->height - 1, ColourMapA[w->colours[1]].mid_light);
+    const auto scrollWidth = w->width - SCROLLBAR_WIDTH - 10;
```
AaronVanGeffen, line 331: Workaround for #11238

Gymnasiast: Can you add that as a comment?

AaronVanGeffen: Done.

tupaschoal:
Needs rebase

WackedSwe:
Somewhere along the way the copy /paste shortcuts in the list have been switched. Copy is not Paste and vice versa

PR #11232, opened 2020-04-05
Fix #11230: Seat Rotation not imported correctly for hacked rides

### LINKED ISSUE #11230: Seat rotation not imported correctly
**OS:** Ubuntu 20.04
**Version:** 0.2.5+
**Commit/Build:** 12ab701a0fc89e1e12a3724ebef164ae4c111172

Copied from https://forums.openrct2.org/topic/4869-4ds-now-broken-when-opening-saves/?tab=comments#comment-23904

 A little more research shows the bug appears in 0.2.3. Before them it wasn't present

A couple more examples of this bug in action taken from this park that was released a couple of years ago.

https://www.nedesigns.com/park/4198/laguna/

Since the release several of the rides no longer operate as they were designed.

version 0.1.2 - Cars on Mack-E-Motion coaster tilt on corners

afbeelding

version 0.2.5 - Cars no longer tilt and are locked to position

afbeelding

version 0.1.2 - Cars on Ghost Train rotate to look at different scenes

afbeelding

version 0.2.5 - Cars are locked to one position and all look the same way
afbeelding

Gymnasiast:


Gymnasiast:
Are you sure it needs calling for the importers as well? (I have increased it anyway, just to be on the safe side.)

duncanspumpkin:
Yes it needs the network increment as the server will have a different value to the client in the old version as the client would get the wrong value from the importer.

duncanspumpkin:
This should fix a big source of desyncs.

PR #11245, opened 2020-04-07
Allow logging in single player too

janisozaur:
@duncanspumpkin that's proven useful with replays stuff

### DIFF
--- a/src/openrct2/Context.cpp
+++ b/src/openrct2/Context.cpp
@@ -587,6 +587,7 @@ namespace OpenRCT2
                         {
                             network_send_map();
                         }
+                        network_start_server_log();
 #ifdef USE_BREAKPAD
                         if (network_get_mode() == NETWORK_MODE_NONE)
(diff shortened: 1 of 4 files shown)


AaronVanGeffen:
Not to dispute its usefulness, but won’t this generate a bunch of unexpected logs for players who occasionally host locally?

duncanspumpkin:
I don't think this is necessary with the silent recordings.

janisozaur:
The purpose here is the replay only tells you (if you look at stack trace) that there was `GameAction:: Execute`, but there is no info on what the action was. At least not directly in the stack trace, but perhaps I wasn't looking at what the derived type of action was? Anyway, I found it easier to debug this way and it was specifically for replay.

tupaschoal:
I'm closing this for seemingly lack of agreement on its usefulness or impact in the game. Please reopen if you think it's applicable Janis

PR #11246, opened 2020-04-07
Fix Various Import/Export issues

duncanspumpkin:
This splits off the fixes from #11228 into a separate PR so that I don't need to remake all the replays immediately.

### DIFF
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -1017,7 +1017,7 @@ void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
     dst->vehicle_type = src->vehicle_type;
     dst->colours = src->colours;
     dst->track_progress = src->track_progress;
-    if (ride != nullptr && ride->mode == RIDE_MODE_BOAT_HIRE)
+    if (ride != nullptr && ride->mode == RIDE_MODE_BOAT_HIRE && src->status == VEHICLE_STATUS_TRAVELLING_BOAT)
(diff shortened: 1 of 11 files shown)


src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1031,9 +1031,10 @@ void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
     }
     else
     {
+        // Track direction and type are in the same field
         dst->track_direction = src->track_direction;
+        // dst->track_type = src->track_type;
```
tupaschoal, line 1036: I found it a bit weird that it doesn't export to `track_type` anymore, but it does read from it when importing.

janisozaur: https://github.com/OpenRCT2/OpenRCT2/blob/0c6828f69204a539d497e2b80c642dff95bd4440/src/openrct2/rct2/S6Exporter.cpp#L1034
https://github.com/OpenRCT2/OpenRCT2/blob/0c6828f69204a539d497e2b80c642dff95bd4440/src/openrct2/ride/Vehicle.h#L198-L202

janisozaur: The code is full of such optimisations

tupaschoal: I see.

duncanspumpkin: And also boat location user the same field in sv6. So before it was writing boat location info into the track type.

PR #11252, opened 2020-04-08
Change the spatial insert to make it ordered

duncanspumpkin:
This is so that the spatial index can be rebuilt and guaranteed to be in the correct order

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -657,13 +657,8 @@ static void SpriteSpatialRemove(SpriteBase* sprite)
     *index = sprite->next_in_quadrant;
 }
 
-static void SpriteSpatialMove(SpriteBase* sprite, CoordsXY newLoc)
+static void SpriteSpatialMove(SpriteBase* sprite, const CoordsXY& newLoc)
 {
-    if (!map_is_location_valid(newLoc))
-    {
-        newLoc.setNull();
-    }
-
     size_t newIndex = GetSpatialIndexOffset(newLoc.x, newLoc.y);
     size_t currentIndex = GetSpatialIndexOffset(sprite->x, sprite->y);
     if (newIndex == currentIndex)
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -352,6 +352,8 @@ void sprite_clear_all_unused()
     }
 }
 
+static void SpriteSpatialInsert(SpriteBase* sprite, CoordsXY newLoc);
+
 static constexpr uint16_t MAX_MISC_SPRITES = 300;
 
 rct_sprite* create_sprite(SPRITE_IDENTIFIER spriteIdentifier)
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Sprite.cpp
```diff
@@ -613,6 +614,60 @@ void sprite_misc_update_all()
...
+}
+
+static void SpriteSpatialRemove(SpriteBase* sprite)
+{
+    size_t currentIndex = GetSpatialIndexOffset(sprite->x, sprite->y);
+    auto* index = &gSpriteSpatialIndex[currentIndex];
+
+    // This indicates that the spatial index data is all incorrect.
+    // Suggest rebuilding the sprite spatial index.
+    Guard::Assert(*index != SPRITE_INDEX_NULL, "Bad sprite spatial Index");
```
IntelOrca, line 640: So call the rebuild function?

duncanspumpkin: Yes I guess that would work.

src/openrct2/world/Sprite.cpp
```diff
@@ -613,6 +614,60 @@ void sprite_misc_update_all()
...
+        sprite2 = &get_sprite(*index)->generic;
+    }
+    *index = sprite->next_in_quadrant;
+}
+
+static void SpriteSpatialMove(SpriteBase* sprite, CoordsXY newLoc)
+{
+    if (!map_is_location_valid(newLoc))
+    {
+        newLoc.setNull();
```
IntelOrca, line 659: I feel like this will cause `GetSpatialIndexOffset` to throw an assert. Perhaps just remove it from the index and not try to add it?

duncanspumpkin: It still needs to be in the index as it needs to end up in `SPATIAL_INDEX_LOCATION_NULL`.
It should work fine performing setNull() as the value is the same as LOCATION_NULL. Although I guess thats not guaranteed to be the same.

IntelOrca: How about just removing this check and instead have `GetSpatialIndexOffset` return `SPATIAL_INDEX_LOCATION_NULL` if it is out of range?

IntelOrca:
Then finally do not send the spatial index over network as that should not be necessary anymore.

src/openrct2/world/Sprite.cpp
```diff
@@ -613,6 +614,60 @@ void sprite_misc_update_all()
     }
 }
 
+// Performs a search to ensure that insert keeps next_in_quadrant in sprite_index order
+static void SpriteSpatialInsert(SpriteBase* sprite, CoordsXY newLoc)
```
tupaschoal, line 618:
```suggestion
static void SpriteSpatialInsert(SpriteBase* sprite, const CoordsXY& newLoc)
```

src/openrct2/world/Sprite.cpp
```diff
@@ -613,6 +614,60 @@ void sprite_misc_update_all()
...
+        if (*index == SPRITE_INDEX_NULL)
+        {
+            break;
+        }
+        sprite2 = &get_sprite(*index)->generic;
+    }
+    *index = sprite->next_in_quadrant;
+}
+
+static void SpriteSpatialMove(SpriteBase* sprite, CoordsXY newLoc)
```
tupaschoal, line 655:
```suggestion
static void SpriteSpatialMove(SpriteBase* sprite, const CoordsXY& newLoc)
```

duncanspumpkin: This one isn't `const` unfortunately.

duncanspumpkin:
I can't remove the spatial index from the replays just yet as it would require remaking all of the replays. Will do that under the other PR.

PR #11254, opened 2020-04-09
Smoke position & animation speed

spacek531:
extending the abilities of objects slightly to move the steam effect of steam trains, and (hopefully) tune the speed that animations play at.

The attached object has properties set to utilize the features
[TESTDAT.zip](https://github.com/OpenRCT2/OpenRCT2/files/4455317/TESTDAT.zip)

### DIFF
--- a/src/openrct2/object/ObjectJsonHelpers.h
+++ b/src/openrct2/object/ObjectJsonHelpers.h
@@ -31,6 +31,7 @@ namespace ObjectJsonHelpers
     float GetFloat(const json_t* obj, const std::string& name, const float& defaultValue = 0);
     std::vector<std::string> GetJsonStringArray(const json_t* arr);
     std::vector<int32_t> GetJsonIntegerArray(const json_t* arr);
+    std::vector<int32_t> GetJsonRealArray(const json_t* arr);
     colour_t ParseColour(const std::string_view& s, colour_t defaultValue = COLOUR_BLACK);
     uint8_t ParseCursor(const std::string& s, uint8_t defaultValue);
     rct_object_entry ParseObjectEntry(const std::string& s);
--- a/src/openrct2/object/ObjectJsonHelpers.h
+++ b/src/openrct2/object/ObjectJsonHelpers.h
@@ -31,7 +31,7 @@ namespace ObjectJsonHelpers
     float GetFloat(const json_t* obj, const std::string& name, const float& defaultValue = 0);
     std::vector<std::string> GetJsonStringArray(const json_t* arr);
     std::vector<int32_t> GetJsonIntegerArray(const json_t* arr);
-    std::vector<int32_t> GetJsonRealArray(const json_t* arr);
+    std::vector<float> GetJsonFloatArray(const json_t* arr);
     colour_t ParseColour(const std::string_view& s, colour_t defaultValue = COLOUR_BLACK);
     uint8_t ParseCursor(const std::string& s, uint8_t defaultValue);
     rct_object_entry ParseObjectEntry(const std::string& s);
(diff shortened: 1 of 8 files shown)


src/openrct2/object/ObjectJsonHelpers.h
```diff
@@ -29,6 +29,7 @@ namespace ObjectJsonHelpers
     float GetFloat(const json_t * obj, const std::string &name, const float &defaultValue = 0);
     std::vector<std::string> GetJsonStringArray(const json_t * arr);
     std::vector<int32_t> GetJsonIntegerArray(const json_t * arr);
+    std::vector<float_t> GetJsonRealArray(const json_t * arr);
```
janisozaur, line 32: Please don't use floating point values for code that touches logic, different compilers and different optimisation flags will lead to divergent results.

src/openrct2/ride/Vehicle.h
```diff
@@ -282,6 +284,9 @@ enum {
     VEHICLE_ENTRY_ANIMATION_MULTI_DIM_COASTER
 };
 
+#define ANIMATION_SPEED_MODIFIER_COEFFICIENT 32
```
Gymnasiast, line 287: All new code should use the C++ counterpart: constexpr.

```suggestion
constexpr const uint8 ANIMATION_SPEED_MODIFIER_COEFFICIENT = 32;
```

src/openrct2/ride/Vehicle.h
```diff
@@ -282,6 +284,9 @@ enum {
...
+#define STEAM_EFFECT_MODIFIER_COEFFICIENT 32
```
Gymnasiast, line 288:
```suggestion
constexpr const uint8 STEAM_EFFECT_MODIFIER_COEFFICIENT = 32;
```
```

File: src/openrct2/object/RideObject.cpp [hunk 4]
```diff
@@ -434,7 +434,11 @@ void RideObject::ReadLegacyVehicle(
     vehicle->effect_visual = stream->ReadValue<uint8_t>();
     vehicle->draw_order = stream->ReadValue<uint8_t>();
     vehicle->num_vertical_frames_override = stream->ReadValue<uint8_t>();
-    stream->Seek(4, STREAM_SEEK_CURRENT);
+    stream->Seek(1, STREAM_SEEK_CURRENT);
+    vehicle->animation_speed_modifier = stream->ReadValue<uint8_t>();
+    vehicle->steam_effect_modifier[0] = stream->ReadValue<uint8_t>();
+    vehicle->steam_effect_modifier[1] = stream->ReadValue<uint8_t>();
+//    stream->Seek(0, STREAM_SEEK_CURRENT);
```
IntelOrca, line 441: New fields should only be added to the JSON format so this needs to be reverted.

src/openrct2/ride/Vehicle.h
```diff
@@ -71,11 +71,13 @@ struct rct_ride_entry_vehicle {
...
-    uint8_t draw_order;
+    uint8_t effect_visual;            // 0x5E , 0x78
+    uint8_t draw_order;               // 0x5F , 0x79
     uint8_t num_vertical_frames_override; // 0x60 , 0x7A, A custom number that can be used rather than letting RCT2 determine it. Needs the VEHICLE_ENTRY_FLAG_OVERRIDE_NUM_VERTICAL_FRAMES flag to be set.
-    uint8_t peep_loading_waypoint_segments; // 0x61 new
-    uint8_t pad_62[6] = {};          // 0x62 , 0x7B
+    uint8_t peep_loading_waypoint_segments; // 0x61, 0x7B new
+    int8_t animation_speed_modifier;  // 0x62 , 0x7C, A multiplier to change animation speed on most animation types
+    int8_t steam_effect_modifier[2] = {}; // 0x63, 0x7D, A multiplier to change smoke particle effect positions in lateral and vertical directions
+    uint8_t pad_65[3] = {};           // 0x65 , 0x7E
```
IntelOrca, line 80: You can remove this pad field.

src/openrct2/object/ObjectJsonHelpers.h
spacek531: this should be resolved now

spacek531:
rebasing in progress

PR #11260, opened 2020-04-10
Fix consistency issues in capitalization and punctuation

Umdlye:
Some strings ended in a period where all similar ones don't (as reported on Discord). Also, hyphens don't start a new sentence but were mostly followed by a capital letter anyway. This mostly occurred in scenario descriptions.

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -230,7 +230,7 @@ STR_0825    :Chosen name in use already
 STR_0826    :Too many names defined
 STR_0827    :Not enough cash - requires {CURRENCY2DP}
 STR_0828    :{SMALLFONT}{BLACK}Close window
-STR_0829    :{SMALLFONT}{BLACK}Window title - drag this to move window
+STR_0829    :{SMALLFONT}{BLACK}Window title - Drag this to move window
 STR_0830    :{SMALLFONT}{BLACK}Zoom view in
 STR_0831    :{SMALLFONT}{BLACK}Zoom view out
 STR_0832    :{SMALLFONT}{BLACK}Rotate view 90° clockwise
@@ -2241,6 +2241,7 @@ STR_3126    :{WINDOW_COLOUR_2}Intensity Factor: {BLACK}+{COMMA16}%
 STR_3127    :{WINDOW_COLOUR_2}Nausea Factor: {BLACK}+{COMMA16}%
 STR_3128    :Save Track Design
 STR_3129    :Save Track Design with Scenery
+STR_3130    :Save
 STR_3131    :Cancel
 STR_3132    :{BLACK}Click items of scenery to select them to be saved with track design...
 STR_3133    :Unable to build this on a slope
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -230,7 +230,7 @@ STR_0825    :Chosen name in use already
 STR_0826    :Too many names defined
 STR_0827    :Not enough cash - requires {CURRENCY2DP}
 STR_0828    :{SMALLFONT}{BLACK}Close window
-STR_0829    :{SMALLFONT}{BLACK}Window title - Drag this to move window
+STR_0829    :{SMALLFONT}{BLACK}Window title - drag this to move window
 STR_0830    :{SMALLFONT}{BLACK}Zoom view in
 STR_0831    :{SMALLFONT}{BLACK}Zoom view out
 STR_0832    :{SMALLFONT}{BLACK}Rotate view 90° clockwise
@@ -890,7 +890,7 @@ STR_1506    :{SMALLFONT}“The litter here is really bad”
 STR_1507    :{SMALLFONT}“I can't find the park exit”
 STR_1508    :{SMALLFONT}“I want to get off {STRINGID}”
 STR_1509    :{SMALLFONT}“I want to get out of {STRINGID}”
-STR_1510    :{SMALLFONT}“I'm not going on {STRINGID} - It isn't safe”
+STR_1510    :{SMALLFONT}“I'm not going on {STRINGID} - it isn't safe”
 STR_1511    :{SMALLFONT}“This path is disgusting”
 STR_1512    :{SMALLFONT}“It's too crowded here”
 STR_1513    :{SMALLFONT}“The vandalism here is really bad”
@@ -2246,7 +2246,7 @@ STR_3131    :Cancel
 STR_3132    :{BLACK}Click items of scenery to select them to be saved with track design...
 STR_3133    :Unable to build this on a slope
 STR_3134    :{RED}(Design includes scenery which is unavailable)
-STR_3135    :{RED}(Vehicle design unavailable - Ride performance may be affected)
+STR_3135    :{RED}(Vehicle design unavailable - ride performance may be affected)
 STR_3136    :Warning: This design will be built with an alternative vehicle type and may not perform as expected
 STR_3137    :Select Nearby Scenery
 STR_3138    :Reset Selection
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
I’m not sure I like these changes, to be honest. The majority of these capitals seem to serve a function at least.

Gymnasiast:
This is the kind of thing that I'd like a native speaker's opinion on.

ocalhoun6:
> 
> 
> This is the kind of thing that I'd like a native speaker's opinion on.

I've looked over the changes, and the new versions _are_ more grammatically correct, as long as you consider them all to be one sentence. Then again, to be proper _English_ grammar, punctuation should be added to all the sentences, rather than removed from the few that have it. But RCT grammar isn't necessarily the same as English grammar.

Personally, I'd lean toward standardizing the capitalization to be the most consistent with the most examples in Vanilla RCT. For punctuation, though, I do think it makes sense to standardize them ... again, probably leaning toward whichever was more common in Vanilla RCT.

pizza2004:
As a native speaker (although American) I would normally say to standardize around proper punctuation, but in this case there's at least two separate things: Peep thoughts, which should follow standard grammar rules, and stuff like descriptions of titles or such, which I think need to be consistent, but don't need to follow grammar rules.  I'm gonna go through and make comments on some of the changes specifically.

data/language/en-GB.txt
```diff
@@ -890,7 +890,7 @@ STR_1506    :{SMALLFONT}“The litter here is really bad”
 STR_1507    :{SMALLFONT}“I can't find the park exit”
 STR_1508    :{SMALLFONT}“I want to get off {STRINGID}”
 STR_1509    :{SMALLFONT}“I want to get out of {STRINGID}”
-STR_1510    :{SMALLFONT}“I'm not going on {STRINGID} - It isn't safe”
+STR_1510    :{SMALLFONT}“I'm not going on {STRINGID} - it isn't safe”
```
pizza2004, line 893: In this case I think all peep thoughts should have proper punctuations, with added periods, and proper capitalization.  I know it won't matter most of the time, but in the instances where peeps are saying something excitedly (ie, "I really liked that ride!") they would have punctuation for emphasis and it's more consistent to always have punctuation on thoughts.

As far as the capitalization, I think the same thing applies, since grammar rules are about replicating human thoughts in the most readable fashion.

data/language/en-GB.txt
```diff
@@ -230,7 +230,7 @@ STR_0825    :Chosen name in use already
 STR_0826    :Too many names defined
 STR_0827    :Not enough cash - requires {CURRENCY2DP}
 STR_0828    :{SMALLFONT}{BLACK}Close window
-STR_0829    :{SMALLFONT}{BLACK}Window title - Drag this to move window
+STR_0829    :{SMALLFONT}{BLACK}Window title - drag this to move window
```
pizza2004, line 233: This is an instance where the capital serves to show two separate thoughts that are simply not using standard punctuation.  Namely, the full thought is "This is the window title.  If you drag this it will move the window."  We're just condensing the information down to the minimum readable size.  I would say that either leave it as it was or change the - punctuation to something that makes the split more obvious.

data/language/en-GB.txt
```diff
@@ -2246,7 +2246,7 @@ STR_3131    :Cancel
 STR_3132    :{BLACK}Click items of scenery to select them to be saved with track design...
 STR_3133    :Unable to build this on a slope
 STR_3134    :{RED}(Design includes scenery which is unavailable)
-STR_3135    :{RED}(Vehicle design unavailable - Ride performance may be affected)
+STR_3135    :{RED}(Vehicle design unavailable - ride performance may be affected)
```
pizza2004, line 2249: Same as above.

data/language/en-GB.txt
```diff
@@ -2419,7 +2419,7 @@ STR_3327    :Starting positions for people not set
 STR_3328    :Can't advance to next editor stage...
 STR_3329    :Park entrance not yet built
 STR_3330    :Park must own some land
-STR_3331    :Path from park entrance to map edge either not complete or too complex - Path must be single-width with as few junctions and corners as possible
+STR_3331    :Path from park entrance to map edge either not complete or too complex - path must be single-width with as few junctions and corners as possible
```
pizza2004, line 2422: This feels like two complete sentences, so I would say that we should replace the - with a period.  Technically you don't need a period to end the other one if you think it looks better.

data/language/en-GB.txt
```diff
@@ -2448,7 +2448,7 @@ STR_3357    :{WINDOW_COLOUR_2}Are you sure you want to permanently delete {STRIN
 STR_3358    :Can't delete track design...
 STR_3359    :{BLACK}No track designs of this type
 STR_3360    :Warning!
-STR_3361    :Too many track designs of this type - Some will not be listed.
+STR_3361    :Too many track designs of this type - some will not be listed
```
pizza2004, line 2451: Same as above, keep capitalization, remove period.  This one is also basically two sentences, so you could change it to a period instead of a -.

data/language/en-GB.txt
```diff
@@ -2469,7 +2469,7 @@ STR_3380    :Unable to install this track design...
 STR_3381    :File is not compatible or contains invalid data
 STR_3382    :File copy failed
 STR_3383    :Select new name for track design
-STR_3384    :An existing track design already has this name - Please select a new name for this design:
+STR_3384    :An existing track design already has this name - please select a new name for this design:
```
pizza2004, line 2472: Change the dash to period, keep capitalization.

data/language/en-GB.txt
```diff
@@ -3721,7 +3721,7 @@ STR_DTLS    :Several islands form the basis for this new park
 <Katie's Dreamland>
 STR_SCNR    :Katie's Dreamland
 STR_PARK    :Katie's Dreamland
-STR_DTLS    :A small theme park with a few rides and room for expansion - Your aim is to double the park value
+STR_DTLS    :A small theme park with a few rides and room for expansion - your aim is to double the park value
```
pizza2004, line 3724: Dash to period.  Keep caps.

data/language/en-GB.txt
```diff
@@ -3761,7 +3761,7 @@ STR_DTLS    :Convert the Egyptian Ruins tourist attraction into a thriving theme
 <Crumbly Woods>
 STR_SCNR    :Crumbly Woods
 STR_PARK    :Crumbly Woods
-STR_DTLS    :A large park with well-designed but rather old rides - Replace the old rides or add new rides to make the park more popular
+STR_DTLS    :A large park with well-designed but rather old rides - replace the old rides or add new rides to make the park more popular
```
pizza2004, line 3765: Same as above.  Most of these are the same.

data/language/en-GB.txt
```diff
@@ -4296,7 +4296,7 @@ STR_DTLS    :Hired by a large amusement park chain, your task is to build them a
 <Gravity Gardens>
 STR_SCNR    :Gravity Gardens
 STR_PARK    :Gravity Gardens
-STR_DTLS    :Your challenge is to build a roller coaster park in the beautiful Gravity Gardens - No other rides, just roller coasters!
+STR_DTLS    :Your challenge is to build a roller coaster park in the beautiful Gravity Gardens - no other rides, just roller coasters!
```
pizza2004, line 4299: This one has exclamation points, which makes for a unique instance.  I would say all the scenario descriptions should be consistent, with periods if any have punctuation at all.  In this particular instance and the others below I think it's more appropriate to replace the - with a comma.

tupaschoal:
I like @pizza2004's considerations

Gymnasiast:
I'd still prefer if someone with _British English_ as their native language would look at this.

duncanspumpkin:
I agree with @pizza2004's comments.

tupaschoal:
I'm marking this as WIP, as it need to take @pizza2004 's comments into consideration

PR #11263, opened 2020-04-10
Fix #5302: Reset rides with looped vehicles when loading

### LINKED ISSUE #5302: Infinite Loop with Splash Boats 1
**OS:** Win 7, 10
**Version:** 0.0.7
**Commit/Build:** 76c154a

Loop occurs when selecting Splash Boats 1, or zooming out and waiting so long after the save loads.

**Steps to reproduce:**
1. Open attached save
2. Select Splash Boats 1

**Dump file** - Very large
https://drive.google.com/open?id=0B-m62aoqIb2BTXVENGotam82NVk

**Save game:**
[NE1_Sat_Mar_11_2017_074220.zip](https://github.com/OpenRCT2/OpenRCT2/files/835831/NE1_Sat_Mar_11_2017_074220.zip)

janisozaur:


janisozaur:
Ping @Sweet-Tan

src/openrct2/world/Sprite.cpp
```diff
@@ -1181,3 +1182,42 @@ int32_t check_for_spatial_index_cycles(bool fix)
...
+
+            if (fast == slow)
+            {
+                loops_found = true;
+                Ride* ride = get_ride(fast->ride);
+                ride->FormatNameTo(gCommonFormatArgs);
+                auto name = ride->GetName() + "'s vehicle list contained loop. Vehicles were removed.";
+                news_item_add_to_queue(NEWS_ITEM_RIDE, STR_RIDE_VEHICLES_LOOPED_REMOVED, fast->ride);
+                log_error(name.c_str());
+                ride_clear_for_construction(ride);
```
AaronVanGeffen, line 1217: Suggesting these lines are swapped for clarity:
```suggestion
                Ride* ride = get_ride(fast->ride);
                ride->FormatNameTo(gCommonFormatArgs);
                news_item_add_to_queue(NEWS_ITEM_RIDE, STR_RIDE_VEHICLES_LOOPED_REMOVED, fast->ride);
                auto name = ride->GetName() + "'s vehicle list contained loop. Vehicles were removed.";
                log_error(name.c_str());
                ride_clear_for_construction(ride);
```

The `news_item_add_to_queue` function uses `gCommonFormatArgs`, while `log_error` uses the `name` string. (Come to think of it, perhaps `message` would be a more appropriate name for this string.)

janisozaur:
Hmm, the tests timed out, I wonder if they entered some infinite loop themselves.

src/openrct2/world/Sprite.cpp
```diff
@@ -1181,3 +1182,42 @@ int32_t check_for_spatial_index_cycles(bool fix)
     }
     return -1;
 }
+
+bool check_for_vehicle_ride_loops()
+{
+    bool loops_found = false;
+    for (uint32_t i = 0; i < RCT2_MAX_SPRITES; i++)
```
duncanspumpkin, line 1189: Use the sprite list so that you dont need to go through each sprite.

src/openrct2/world/Sprite.cpp
```diff
@@ -1181,3 +1182,42 @@ int32_t check_for_spatial_index_cycles(bool fix)
...
+            fast = try_get_vehicle(fast->next_vehicle_on_ride);
+            Guard::Assert(fast != nullptr, "Ride's vehicle list contains non-vehicles");
+
+            slow = try_get_vehicle(slow->next_vehicle_on_ride);
+            Guard::Assert(slow != nullptr, "Ride's vehicle list contains non-vehicles");
+
+            if (fast == slow)
+            {
+                loops_found = true;
+                Ride* ride = get_ride(fast->ride);
```
duncanspumpkin, line 1212: Guard::Assert allows you to get to this point with a `nullptr`

duncanspumpkin:
I'm not sure this would work. As far as I know `next_vehicle_on_ride` is always a loop. Every single write to `next_vehicle_on_ride` involves a sprite_index of an existing vehicle. I think it therefore means there cant be a `NULL`.

janisozaur:
In #5302 the problem is vehicle to be drawn is splash boat and the vehicle to be drawn is also a splash boat and it overflows the stack trying to figure out what to draw: https://github.com/OpenRCT2/OpenRCT2/blob/b5336fc6b6344af921a7bf4733fb2da98d4a9f1c/src/openrct2/ride/water/SplashBoats.cpp#L1255

https://github.com/OpenRCT2/OpenRCT2/blob/b5336fc6b6344af921a7bf4733fb2da98d4a9f1c/src/openrct2/ride/water/SplashBoats.cpp#L1265

https://github.com/OpenRCT2/OpenRCT2/blob/f1dcf8573138116e80544433e34548a952d06028/src/openrct2/ride/VehiclePaint.cpp#L3193

janisozaur:
Indeed the rides in other parks are also loops, but in this particular sv6, all the vehicles are heads, unlike every other case. I think this should be enough of a test.

Following the links and printing out some information along the way for valid park (e.g. our current title sequence):
```
vehicle->id = 183
    vehicle->id = 183, ishead = 1
    vehicle->id = 184, ishead = 0
    vehicle->id = 59, ishead = 0
    vehicle->id = 88, ishead = 1
    vehicle->id = 199, ishead = 0
    vehicle->id = 200, ishead = 0
    vehicle->id = 201, ishead = 1
    vehicle->id = 202, ishead = 0
    vehicle->id = 203, ishead = 0
    vehicle->id = 87, ishead = 1
head
```
The pattern is always the same: head, followed by two non-heads, regardless of vehicle count.

In the offending park all vehicle in the loop are heads.

duncanspumpkin:
All heads is perfectly acceptable for a ride with only single car length trains. A splash boat should always have 3 cars though (1 head 2 tail).
https://github.com/OpenRCT2/objects/blob/master/objects/rct2/ride/rct2.spboat.json#L17-L18

I guess you could check to make sure that the min and max values are set but people have been known to override this. I think we will have to come up with a different way of doing this.

tupaschoal:
As this is still broken and stalled for a while, I'll close it

Issue #10806, opened 2020-02-21
Bugged Lift Cabins on Loading SV6


maddkattunge:
**OS:** Windows 10
**Version:** 0.2.4 develop (also on release)
**Commit/Build:** 12ab1e9

Short Lifts composed of only two station tracks with no other track in between will bug out upon opening the park's SV6 save file. The Lift cabin will not stop at Station 2 and will instead immediately reverse direction to then be caught in an infinite loop bouncing between stations.

- [No] Reproducible in RCT2 (vanilla)?
- [No] Specific to multiplayer?

**Steps to reproduce:**
1. Build a Lift composed of only two stations.
2. Operate Lift in either Test or Open mode.
3. Save and reload.

**Video:**
https://www.youtube.com/watch?v=3jXRMtOiYn0

**Save game:**
[mxi_bug_1_brokenlift.zip](https://github.com/OpenRCT2/OpenRCT2/files/4237478/mxi_bug_1_brokenlift.zip)

Gymnasiast:
Duplicate of https://github.com/OpenRCT2/OpenRCT2/issues/2128.

maddkattunge:
This does not happen in Vanilla RCT2 and I do not understand how this is a duplicate of the #2128  Chairlift bug since these are two different rides. Could you help me understand why this is a duplicate bug report?

duncanspumpkin:
I think on first read of your issue it sounded like it was talking about chairlifts. I see from your video it is talking about lifts though.
Either way its still a duplicate but this is actually a duplicate of this issue: https://github.com/OpenRCT2/OpenRCT2/issues/10509

Gymnasiast:
@maddkattunge I misread it, my bad.

@duncanspumpkin I'm not sure if it's a duplicate of that - #10509 has a "normal" lift.

PR #11267, opened 2020-04-11
Fix #10372, #10509, #10806. Lift base sections did not export correctly

### LINKED ISSUE #10372: Guests enter elevator through the base entrance but appear at the top, then they ride the elevator down.
**OS:** Windows 10
**Version:** 0.2.4+
**Commit/Build:** fed417da033fe81d35be7a337e7f5ce4d762a0e1

I built a park with a simple elevator ride trying to reproduce a different bug and stumbled upon some weird behavior. When guests enter the bottom entrance of the elevator they appear at the top and ride the elevator down.

**Steps to reproduce:**
Open the attached save file and check out the elevator ride
[elevator_bug.zip](https://github.com/OpenRCT2/OpenRCT2/files/3962672/elevator_bug.zip)

**Screenshots / Video:**
[OpenRCT2 2019-12-13 15-11-55_Trim.zip](https://github.com/OpenRCT2/OpenRCT2/files/3962677/OpenRCT2.2019-12-13.15-11-55_Trim.zip)

duncanspumpkin:
Fix #10372, #10509, #10806. Lift base sections did not export correctly
Mistake made whilst refactoring that assumed that station indexes are set on only station pieces.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -7,6 +7,7 @@
 - Fix: [#6238] Invalid tile elem iteration in Guest::UpdateUsingBin
 - Fix: [#7094] Back wall edge texture in water missing.
 - Fix: [#9719] Hacked walls in RCT1 saves are imported incorrectly. 
+- Fix: [#10372, #10509, #10806] Lift base sections incorrectly exporting, causing various lift related bugs.
 - Fix: [#10928] File browser's date column is too narrow.
 - Fix: [#10951, #11160] Attempting to place park entrances creates ghost entrances in random locations.
 - Fix: [#11005] Company value overflows.
@@ -16,7 +17,7 @@
 - Fix: [#11208] Cannot export parks with RCT2 DLC objects.
 - Fix: [#11230] Seat Rotation not imported correctly for hacked rides.
 - Fix: [#11225] Replay manager cannot handle track designs.
-- Fix: [#11246] Fix Various Import/Export issues
+- Fix: [#11246] Fix Various Import/Export issues with Boat locations, balloon frame number.
 - Fix: Small red gardens in RCT1 saves are imported in the wrong colour.
 - Improved: [#11157] Slimmer virtual floor lines.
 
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
This is another import/export bug so I wasn't going to add an entry to the changelog. But I could append the extra numbers if we feel it would be worthwhile.

AaronVanGeffen:
Hmm, I do think it's worthwhile to mention! Some people who have ran into this bug may only be reading the changelog to see if it's been fixed, after all. (I also think we should probably elaborate on the `Various Import/Export issues` that the other PR addressed.)

distribution/changelog.txt
```diff
@@ -7,6 +7,8 @@
 - Fix: [#6238] Invalid tile elem iteration in Guest::UpdateUsingBin
 - Fix: [#7094] Back wall edge texture in water missing.
 - Fix: [#9719] Hacked walls in RCT1 saves are imported incorrectly. 
+- Fix: [#10372, #10509, #10806] Lift base sections incorrectly exporting. This
+caused various lift related bugs.
```
AaronVanGeffen, line 11:
```suggestion
- Fix: [#10372, #10509, #10806] Lift base sections incorrectly exporting, causing various lift-related bugs.
```

distribution/changelog.txt
```diff
@@ -16,7 +18,8 @@
 - Fix: [#11208] Cannot export parks with RCT2 DLC objects.
 - Fix: [#11230] Seat Rotation not imported correctly for hacked rides.
 - Fix: [#11225] Replay manager cannot handle track designs.
-- Fix: [#11246] Fix Various Import/Export issues
+- Fix: [#11246] Fix Various Import/Export issues (Boat locations, balloon
+frame number)
```
AaronVanGeffen, line 22: Should be on one line:
```suggestion
- Fix: [#11246] Fix import/export issues with boat locations, balloon frame number.
```

Issue #11262, opened 2020-04-10
Mechanic Glitch-Walking Across Water


Oricuss:
Windows 10
v0.2.5 
(4f6e77e)

One of my mechanics is walking across the water in my game where there is no path and I didn't make anything invisible there. I didn't see what caused him to do this, I only noticed it when I saw him in the middle of my lake. He is twitching back and a forth and traveling diagonal. He cannot be picked up but can be fired. I also didn't notice him until he was halfway across the map from the ride he was fixing so I don't know what caused it. But his status bar says he is mid fixing the ride. The ride has invisible exits using the No Entrance option in the drop down menu. The Entrance and Exit face the normal directions. But otherwise it is the pre-made Flying Dutchman Gold Mine Wild Mouse. I watched him until he hit the edge of the map and he fell off the side and reset at that point.

- I don't know if it is reproducible in Vanilla as I don't know what caused it.
- Single Player Game
- The game didn't crash so I don't have a dump file.

[Infernal Views Save Game](https://github.com/OpenRCT2/OpenRCT2/files/4463271/Infernal.Views.Glitch.txt)

![Capture](https://user-images.githubusercontent.com/54409340/79020536-11d09c00-7b47-11ea-8e52-6ba9f38d8fab.PNG)

AaronVanGeffen:
Yeah, this is an interesting one… Duplicate of #8145 (and, by extension, #11210), though.

Issue #11259, opened 2020-04-10
Custom JSON object breaks saves


Umdlye:
**OS:** Windows 10 (x64)
**Version:** 0.2.5
**Commit/Build:** fdf9806

I created a custom JSON scenery group object following @Gymnasiast's instructions the other day and it works as expected, but it breaks saves it's included in regardless of export settings.

When exporting custom objects is enabled and the park is saved, saving aborts while giving the error `Header found in object file does not match object to pack.`: [save log.txt](https://github.com/OpenRCT2/OpenRCT2/files/4461493/save.log.txt)
The created save is in the attached .zip (export.sv6).

When exporting custom objects is disabled and the park is saved, the park appears to save normally. The created save is in the attached .zip (no export.sv6).
When you try to load this save it appears to export the object while loading, resulting in the same error: [load log.txt](https://github.com/OpenRCT2/OpenRCT2/files/4461507/load.log.txt)

This .zip has the object file and the two save files: [objectissue.zip](https://github.com/OpenRCT2/OpenRCT2/files/4461510/objectissue.zip)

This is the .json file used for the object:

```
{
    "id": "#UMDSCGT",
    "authors": ["Umdlye"],
    "version": "1.0",
	"originalId": "00000087|#UMDSCGT|00000000",
    "sourceGame": "custom",
    "objectType": "scenery_group",
    "properties": {
        "entries": ["TGE1", "GEOROOF2", "LAMP1"],
        "priority": 20
    },
    "images": [
        { "path": "images/0.png", "x": 0, "y": 0 },
        { "path": "images/1.png", "x": 0, "y": 0 } 
    ],
    "strings": {
        "name": {
            "en-GB": "Test group",
            "fr-FR": "Test group",
            "de-DE": "Test group",
            "es-ES": "Test group",
            "it-IT": "Test group",
            "nl-NL": "Test group",
            "sv-SE": "Test group",
            "ko-KR": "Test group",
            "zh-CN": "Test group",
            "zh-TW": "Test group",
            "pt-BR": "Test group",
            "cs-CZ": "Test group",
            "ja-JP": "Test group",
            "pl-PL": "Test group",
            "ru-RU": "Test group"
        }
    }
}
```

duncanspumpkin:
That is expected behavior but it should be properly warned.

Umdlye:
So custom JSON objects are not intended to be used with the .sv6 format at all, even if they're not exported? That seems a bit weird.

[...]

PR #11269, opened 2020-04-11
steam position modifier and animation speed properties

spacek531:
This PR is for custom rides.

Steam Position: moves the spawn point of steam effects forward/backward and up/down by multiplying the existing steam offsets by the specified values

Animation speed: changes the way vehicle animations (specifically, ones that are dependent on velocity of the vehicle) to a unified timing mechanism.


duncanspumpkin:
Please do not open and close change requests we loose information and it spams our notifications with extra items. It looked like you had messed up a rebase but that isn't unrecoverable.

spacek531:
This new PR is not only a rebase fix, it's a complete rewrite and changes the implementation a substantial amount.

tupaschoal:
Since it doesn't mention the old PR, can you please edit the PR description and state what you are doing, why and ideally add some screenshots/gifs? That makes it easier for us to review and makes for a good documentation afterwards.

ZehMatt:
> This new PR is not for a rebase fix, it's a complete rewrite and changes the implementation a substantial amount.

If this is a complete rewrite u could also have done that in the other PR by force pushing.

tupaschoal:
Formatting is broken, can you please fix it?

duncanspumpkin:
This requires fixing the object exporter before merging.

spacek531:
Can you elaborate on that duncan?

duncanspumpkin:
We can't merge until https://github.com/OpenRCT2/OpenRCT2/issues/11259 is fixed. I'm not expecting you @spacek531 to fix it but its more a warning to any reviewers or myself.

src/openrct2/object/RideObject.cpp
```diff
@@ -775,6 +781,22 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(const json_t* jCar)
     car.draw_order = ObjectJsonHelpers::GetInteger(jCar, "drawOrder");
     car.num_vertical_frames_override = ObjectJsonHelpers::GetInteger(jCar, "numVerticalFramesOverride");
 
+    auto jAnimationSpeed = json_object_get(jCar, "animationSpeed");
+    car.animation_speed_multiplier = ObjectJsonHelpers::GetInteger(
+        jAnimationSpeed, "multiplier", VehicleDefaultAnimationSpeeds[car.animation] >> 5);
+    car.animation_speed_exponent = ANIMATION_SPEED_MAX_EXPONENT
+        - ObjectJsonHelpers::GetInteger(
```
Gymnasiast, line 788: This is now spanning four lines. For legibility, this should be broken up. `ANIMATION_SPEED_MAX_EXPONENT - (VehicleDefaultAnimationSpeeds[car.animation] & 0x1F)` can be saved in a separate constexpr, and the result of `ObjectJsonHelpers::GetInteger(...` in a separate variable as well.

src/openrct2/ride/RideData.cpp
```diff
@@ -1045,7 +1045,7 @@ const uint8_t RideAvailableModes[] = {
     RIDE_MODE_CIRCUS_SHOW, 0xFF,                                                                                                                            // 31 Circus Show
     RIDE_MODE_CONTINUOUS_CIRCUIT, 0xFF,                                                                                                                     // 32 Ghost Train
     RIDE_MODE_CONTINUOUS_CIRCUIT, RIDE_MODE_CONTINUOUS_CIRCUIT_BLOCK_SECTIONED, 0xFF,                                                                       // 33 Twister Roller Coaster
-    RIDE_MODE_CONTINUOUS_CIRCUIT, RIDE_MODE_CONTINUOUS_CIRCUIT_BLOCK_SECTIONED, RIDE_MODE_REVERSE_INCLINE_LAUNCHED_SHUTTLE, 0xFF,                                                                       // 34 Wooden Roller Coaster
```
Gymnasiast, line 1048: This line, and 1069, need to be reverted.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7229,8 +7229,7 @@ static void steam_particle_create(int16_t x, int16_t y, int16_t z)
  */
 static void vehicle_update_additional_animation(Vehicle* vehicle)
 {
-    uint8_t al, ah;
-    uint32_t eax;
+    uint8_t al;
```
Gymnasiast, line 7232: If anywhere possible, could you come up with a more descriptive name for `al`?

src/openrct2/ride/Vehicle.h
```diff
@@ -372,6 +375,25 @@ enum : uint32_t
     VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT = 1u << 31,
 };
 
+constexpr const uint8_t ANIMATION_SPEED_MAX_EXPONENT = 32;
+constexpr const uint8_t STEAM_EFFECT_TRANSLATION_COEFFICIENT = 32;
+
+const uint8_t VehicleDefaultAnimationSpeeds[] = {
+    // first 5 bits: bitshift value where 31 is slowest, 0 is fastest
+    // last 3 bits: multiplier 0-7 (be warned a multiplier of 0 means animations will not play)
+    // Exponent value in JSON objects is 32 - bitshift, such that "Exponent":12 translates to bitshift value of 20
```
Gymnasiast, line 384: This comment should be above the declaration, in an /* */ block.

Otherwise: good job on annotating and commenting on it! :+1:

src/openrct2/ride/Vehicle.h
```diff
@@ -372,6 +375,25 @@ enum : uint32_t
     VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT = 1u << 31,
 };
 
+constexpr const uint8_t ANIMATION_SPEED_MAX_EXPONENT = 32;
+constexpr const uint8_t STEAM_EFFECT_TRANSLATION_COEFFICIENT = 32;
+
+const uint8_t VehicleDefaultAnimationSpeeds[] = {
```
Gymnasiast, line 381:
```suggestion
constexpr const uint8_t VehicleDefaultAnimationSpeeds[] = {
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7205,8 +7205,7 @@ static void steam_particle_create(int16_t x, int16_t y, int16_t z)
  */
 void Vehicle::UpdateAdditionalAnimation()
 {
-    uint8_t al, ah;
-    uint32_t eax;
+    uint8_t current_frame;
```
ZehMatt, line 7208: Can u change this to camelCase

src/openrct2/ride/RideData.cpp
```diff
@@ -729,7 +729,10 @@ const rct_ride_entry_vehicle CableLiftVehicle = {
     /* .effect_visual = */ 1,
     /* .draw_order = */ 14,
     /* .num_vertical_frames_override = */ 0,
-    /* .peep_loading_positions = */ 0
```
Gymnasiast, line 732: Why did you remove this line?

spacek531: I don't know why... I never noticed because the code worked.

tupaschoal:
@spacek531 can you mark the issues that are dealt with as solved? There's also a formatting issue on the CI, would you please check?

src/openrct2/ride/Vehicle.h
```diff
@@ -91,12 +91,15 @@ struct rct_ride_entry_vehicle
...
+    uint8_t effect_visual;        // 0x5E , 0x78
+    uint8_t draw_order;           // 0x5F , 0x79
     uint8_t num_vertical_frames_override;   // 0x60 , 0x7A, A custom number that can be used rather than letting RCT2 determine
                                             // it. Needs the VEHICLE_ENTRY_FLAG_OVERRIDE_NUM_VERTICAL_FRAMES flag to be set.
     uint8_t peep_loading_waypoint_segments; // 0x61 new
-    uint8_t pad_62[6] = {};                 // 0x62 , 0x7B
+    uint8_t animation_speed_multiplier;     // 0x62, multiply Var_C8 before bit-shifting
+    uint8_t animation_speed_exponent;       // 0x63, bits to shift when converting speed into animation frames
+    int8_t steam_effect_translation[2];     // 0x64-0x65, multiplier for steam effect spawn point
+    uint8_t pad_65[2] = {};                 // 0x66-0x67
```
duncanspumpkin, line 102: remove the padding its no longer required.

src/openrct2/ride/RideData.cpp
duncanspumpkin: can you put it back in it will still compile afaik.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7218,15 +7217,12 @@ void Vehicle::UpdateAdditionalAnimation()
...
-                animation_frame = ah;
-                al &= 0x02;
-                ah &= 0x02;
-                if (al != ah)
+            currentFrame = ((*curVar_C8 * vehicleEntry->animation_speed_multiplier) >> vehicleEntry->animation_speed_exponent)
+                & 3;
+            if (animation_frame != currentFrame)
+            {
+                animation_frame = currentFrame;
+                if ((currentFrame & 0x01) == 0) // every even frame create a steam particle
```
duncanspumpkin, line 7225: This is subtly different to the original. The original only shows new steam if the oldFrame and the newFrame have a mismatched second bit. This new version shows new steam if the newFrame is odd. Is this a conscious change that is being made or just for the sake of simplification. I would prefer we matched the original as much as possible but understand you might have made this change for a reason.

src/openrct2/ride/Vehicle.h
```diff
@@ -91,12 +91,15 @@ struct rct_ride_entry_vehicle
...
-    uint8_t draw_order;
+    uint8_t effect_visual;        // 0x5E , 0x78
+    uint8_t draw_order;           // 0x5F , 0x79
     uint8_t num_vertical_frames_override;   // 0x60 , 0x7A, A custom number that can be used rather than letting RCT2 determine
                                             // it. Needs the VEHICLE_ENTRY_FLAG_OVERRIDE_NUM_VERTICAL_FRAMES flag to be set.
     uint8_t peep_loading_waypoint_segments; // 0x61 new
-    uint8_t pad_62[6] = {};                 // 0x62 , 0x7B
+    uint8_t animation_speed_multiplier;     // 0x62, multiply Var_C8 before bit-shifting
+    uint8_t animation_speed_exponent;       // 0x63, bits to shift when converting speed into animation frames
+    int8_t steam_effect_translation[2];     // 0x64-0x65, multiplier for steam effect spawn point
```
duncanspumpkin, line 101: first byte is for x/y multiplier second byte is for z multiplier.

spacek531:
The old way required a number of variable swaps to compare the old frame index to the new one. Changing it to this method allowed me to remove 2 variables from the function, specifically ones with placeholder names. There are a few edge cases that considered when I changed the method, but they require sufficient hacks as to make code readability more important. Driving backwards or going so fast the animation plays backwards have the effect of changing smoke generation from even frames to odd frames in the old method, whereas the new method always generates smoke on even frames.

duncanspumpkin:
Okay that sounds reasonable.

spacek531:
Removal of the padding fails the Linux i686 portable build.

duncanspumpkin (dismissed):
Approved but i want to investigate why removing padding fails the ci at some point.

duncanspumpkin:
Prior to merging we just need to rebase and increment network version

duncanspumpkin:
I've squashed everything into two commits to make it easier to follow.

src/openrct2/ride/Vehicle.h
```diff
@@ -453,6 +457,27 @@ enum : uint32_t
...
 
+constexpr const uint8_t ANIMATION_SPEED_MAX_EXPONENT = 32;
+constexpr const uint8_t STEAM_EFFECT_TRANSLATION_COEFFICIENT = 32;
+
+/*
+    First value is bitshift value where 31 is slowst and 0 is fastest
+    Second value is multiplier 0-7 (be warned a multiplier of 0 means animations will not play)
+    Exponent value in JSON objects is 32 - bitshift, such that "Exponent":12 translates to bitshift value of 20
+*/
+constexpr const uint8_t VehicleDefaultAnimationSpeeds[][2] = {
```
Gymnasiast, line 468: Sorry that I forgot about this, but I would like you to create a struct with two `uint8_t`s, rather than this multidimensional array.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7259,48 +7255,57 @@ void Vehicle::UpdateAdditionalAnimation()
...
                 Invalidate();
             }
             break;
         case VEHICLE_ENTRY_ANIMATION_CANOES: // loc_6D6482
             *curVar_C8 += _vehicleVelocityF64E08;
-            eax = ((*curVar_C8 >> 13) & 0xFF) * 6;
-            ah = (eax >> 8) & 0xFF;
-            if (animation_frame != ah)
+            currentFrame = ((*curVar_C8 * vehicleEntry->animation_speed_multiplier) >> vehicleEntry->animation_speed_exponent)
+                % 6;
```
Gymnasiast, line 7285: This looks like a deviation from the old code - it used to do modulo 256, now it does modulo 6.

duncanspumpkin: It is actually the same but the original did it via some bit shifting

src/openrct2/ride/Vehicle.h
```diff
@@ -453,6 +457,27 @@ enum : uint32_t
     VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT = 1u << 31,
 };
 
+constexpr const uint8_t ANIMATION_SPEED_MAX_EXPONENT = 32;
+constexpr const uint8_t STEAM_EFFECT_TRANSLATION_COEFFICIENT = 32;
+
+/*
+    First value is bitshift value where 31 is slowst and 0 is fastest
```
ZehMatt, line 464:
```suggestion
    First value is bitshift value where 31 is slowest and 0 is fastest
```

src/openrct2/object/RideObject.cpp
```diff
@@ -763,6 +767,24 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(const json_t* jCar)
...
+
+    car.animation_speed_multiplier = animationMultiplier;
+    car.animation_speed_exponent = ANIMATION_SPEED_MAX_EXPONENT - animationExponent;
+
+    auto jSteamTranslation = json_object_get(jCar, "steamPositionModifier");
+    auto longitudeSteam = ObjectJsonHelpers::GetFloat(jSteamTranslation, "longitudinal", 1.0);
+    auto verticalSteam = ObjectJsonHelpers::GetFloat(jSteamTranslation, "vertical", 1.0);
+
+    car.steam_effect_translation[0] = std::clamp<int8_t>(
+        longitudeSteam * STEAM_EFFECT_TRANSLATION_COEFFICIENT, INT8_MIN, INT8_MAX);
```
ZehMatt, line 784: Use std::numeric_limits over the C constants.

src/openrct2/object/RideObject.cpp
```diff
@@ -763,6 +767,24 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(const json_t* jCar)
...
+    auto animationMultiplier = ObjectJsonHelpers::GetInteger(
+        jAnimationSpeed, "multiplier", VehicleDefaultAnimationSpeeds[car.animation][1]);
+    auto animationExponent = ObjectJsonHelpers::GetInteger(
+        jAnimationSpeed, "exponent", ANIMATION_SPEED_MAX_EXPONENT - (VehicleDefaultAnimationSpeeds[car.animation][0]));
+
+    car.animation_speed_multiplier = animationMultiplier;
+    car.animation_speed_exponent = ANIMATION_SPEED_MAX_EXPONENT - animationExponent;
+
+    auto jSteamTranslation = json_object_get(jCar, "steamPositionModifier");
+    auto longitudeSteam = ObjectJsonHelpers::GetFloat(jSteamTranslation, "longitudinal", 1.0);
```
ZehMatt, line 780: Not exactly a fan of having floats around, different compilers and platforms may bring different results, there are no guarantees that this will perform exactly the same way. I recall removing the misc sprites from the network checksum but I rather have the simulation deterministic for all aspects.

src/openrct2/object/RideObject.cpp
```diff
@@ -763,6 +767,24 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(const json_t* jCar)
...
+    auto verticalSteam = ObjectJsonHelpers::GetFloat(jSteamTranslation, "vertical", 1.0);
+
+    car.steam_effect_translation[0] = std::clamp<int8_t>(
+        longitudeSteam * STEAM_EFFECT_TRANSLATION_COEFFICIENT, INT8_MIN, INT8_MAX);
+    car.steam_effect_translation[1] = std::clamp<int8_t>(
+        verticalSteam * STEAM_EFFECT_TRANSLATION_COEFFICIENT, INT8_MIN, INT8_MAX);
```
ZehMatt, line 786: Same as above

src/openrct2/object/ObjectJsonHelpers.cpp
```diff
@@ -164,6 +164,25 @@ namespace ObjectJsonHelpers
         }
         return result;
     }
+    std::vector<float> GetJsonFloatArray(const json_t* arr)
```
ZehMatt, line 167: This function seems to be not used?

src/openrct2/ride/Vehicle.h
```diff
@@ -23,6 +23,18 @@
 
 enum class SoundId : uint8_t;
 
+struct rct_animation_speed
```
duncanspumpkin, line 26: Use CamelCase for new structs. `AnimationSpeed`

src/openrct2/ride/Vehicle.h
```diff
@@ -23,6 +23,18 @@
...
+{
+    uint8_t exponent;
+    uint8_t multiplier;
+};
+
+struct steam_effect_translation
```
duncanspumpkin, line 32: `SteamEffectTranslation`

spacek531:
This changes animations to have a variable number of frames.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7280,26 +7280,42 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
             animationState += _vehicleVelocityF64E08;
-            targetFrame = (animationState >> 20) & 3;
+            // Step 1: shift the value right by 8 less than you want
+            // Step 2: mask by 0xFF to get only 8 bits of value
+            // Step 3: multiply by the number of frames
+            // Step 4: shift right by 8 to remove the masked bits
+            // Step 5: you are left with the frame number
+            targetFrame = ((animationState >> std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32) & 0xFF)
+                           * vehicleEntry->AnimationFrames)
+                >> 8;
```
spacek531, line 7306: This is necessary to prevent skipped frames of non-powers of two frame lengths.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7280,26 +7280,42 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
     auto vehicleEntry = Entry();
     if (vehicleEntry == nullptr)
     {
         return;
     }
+    // similar animation styles all update the same
     switch (vehicleEntry->animation)
     {
-        case VEHICLE_ENTRY_ANIMATION_MINITURE_RAILWAY_LOCOMOTIVE: // loc_6D652B
+        case VEHICLE_ENTRY_ANIMATION_MINITURE_RAILWAY_LOCOMOTIVE:
```
spacek531, line 7292: Note for a future PR: refactor this enum and fix spelling error

src/openrct2/ride/Vehicle.h
```diff
@@ -427,17 +459,17 @@ enum : uint32_t
     VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED = 1 << 8, // Not used any more - every vehicle will now work with doors.
     VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2 = 1 << 9,
     VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS = 1 << 10, // Only used during loading of the objects.
-    VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES = 1
-        << 11, // Instead of the default 32 rotation frames. Only used for boat hire and works only for non sloped sprites.
```
spacek531, line 431: I don't know why clang format wants to move this around, but I don't argue because it's clang format

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1525,13 +1525,13 @@ bool Vehicle::OpenRestraints()
...
             {
-                vehicle->animationState += 0x3333;
+                vehicle->animationState += vehicleEntry->AnimationSpeed;
             }
             else
             {
                 vehicle->animationState = 0;
                 vehicle->animation_frame++;
-                vehicle->animation_frame &= 7;
+                vehicle->animation_frame %= vehicleEntry->AnimationFrames;
```
ZehMatt, line 1534: This will now wrap instead of a cap, isn't this changing the logic?

Broxzier: `&= 7` is the same as `%= 8`. Assuming that AnimationFrames is 8 here.

ZehMatt: Right I forgot power of two thingy since it was 7. Should be alright if this holds 8 i guess.

spacek531: the default value is 8 yes

src/openrct2/ride/Vehicle.h
Broxzier: Because of the line length limit. The comment is pretty long.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7280,26 +7280,41 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
-        case VEHICLE_ENTRY_ANIMATION_MINITURE_RAILWAY_LOCOMOTIVE: // loc_6D652B
+        case VEHICLE_ENTRY_ANIMATION_MINITURE_RAILWAY_LOCOMOTIVE:
+        case VEHICLE_ENTRY_ANIMATION_SWAN:
+        case VEHICLE_ENTRY_ANIMATION_CANOES:
+        case VEHICLE_ENTRY_ANIMATION_ROW_BOATS:
+        case VEHICLE_ENTRY_ANIMATION_WATER_TRICYCLES:
+        case VEHICLE_ENTRY_ANIMATION_HELICARS:
             animationState += _vehicleVelocityF64E08;
-            targetFrame = (animationState >> 20) & 3;
+            // Step 1: shift the value right by 8 less than you want
```
ZehMatt, line 7298: A good comment describes what the purpose of the code and not what each operation of the statement does.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7280,26 +7280,41 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
+            // Step 2: mask by 0xFF to get only 8 bits of value
+            // Step 3: multiply by the number of frames
+            // Step 4: shift right by 8 to remove the masked bits
+            // Step 5: you are left with the frame number
+            targetFrame = ((animationState >> std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32) & 0xFF)
```
ZehMatt, line 7303: This line is not clear if you want to have
```cpp
(animationState >> std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32)) & 0xFF
```
or
```cpp
animationState >> (std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32) & 0xFF)
```
Make this explicit.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7322,92 +7337,72 @@ void Vehicle::UpdateAdditionalAnimation()
...
+            // similar to a normal animation but only animates when a rider is present
+        case VEHICLE_ENTRY_ANIMATION_MONORAIL_CYCLES: // loc_6D64B6
+            if (num_peeps != 0)
             {
-                animation_frame = targetFrame;
-                Invalidate();
+                animationState += _vehicleVelocityF64E08;
+                targetFrame = ((animationState >> std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32 - 8) & 0xFF)
+                               * vehicleEntry->AnimationFrames)
+                    >> 8;
```
ZehMatt, line 7379: Duplicate code, move this into a function.

src/openrct2/ride/VehicleEntry.h
```diff
@@ -15,6 +15,12 @@
 #include <array>
 #include <vector>
 
+struct SteamEffectType
```
ZehMatt, line 18: Not the best name, something like SteamEffectData might be a bit better, the ```Type``` part would rather apply to an enum with various steam effect types.

src/openrct2/ride/Vehicle.cpp
ZehMatt: Also there are magic values like 8 and 32, they should be constants.

spacek531:
I meant, why is clang format moving this line when I didn't remove any `// clang-format off` comments

src/openrct2/ride/Vehicle.cpp
spacek531: > This line is not clear if you want to have
> 
> ```c++
> (animationState >> std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32)) & 0xFF
> ```
> 
> or
> 
> ```c++
> animationState >> (std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32) & 0xFF)
> ```
> 
> Make this explicit.

The second option is the same as not having `& 0xFF` because the value is already lower than 256.

spacek531: > Also there are magic values like 8 and 32, they should be constants.

They are the sizes of integer values, but I can do that.

ZehMatt: > > This line is not clear if you want to have
> > ```c++
> > (animationState >> std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32)) & 0xFF
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or
> > ```c++
> > animationState >> (std::clamp(vehicleEntry->AnimationSpeed - 8, 0, 32) & 0xFF)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Make this explicit.
> 
> The second option is the same as not having `& 0xFF` because the value is already lower than 256.

That is why I'm saying use parenthesis to make your intent clear here.

spacek531: I already use parenthesis so I'm not sure why you are saying this.

ZehMatt: Because who ever reads the code may question if want to shift with a masked value of 0xFF or if you want to mask the resulting value with 0xFF. There is a lot going on with this single line already, but also as suggested its best to move this into a function and it might be best to use a few variables to make this even easier to read.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -50,6 +50,8 @@
 #include <algorithm>
 #include <iterator>
 
+#define UINT8_WIDTH 8
```
ZehMatt, line 53: Properly define this with constexpr.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7273,6 +7275,16 @@ void Vehicle::UpdateAnimationAnimalFlying()
     }
 }
 
+uint8_t Vehicle::GetTargetFrame(rct_ride_entry_vehicle* vehicleEntry)
```
ZehMatt, line 7278: Would be better to have this a static function, you are exposing this function in the header for no reason.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7273,6 +7275,16 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
+{
+    animationState += _vehicleVelocityF64E08;
+    // Similar to a modulus, but using a mask and multiplying before a final bit shift to get rid of the mask bits. You are left
+    // with a value from 0 to AnimationFrames - 1. Prevents skipped frames on non powers-of-two AnimationFrames.
+    return (((animationState >> std::max(vehicleEntry->AnimationSpeed - UINT8_WIDTH, 0)) & UINT8_MAX)
```
ZehMatt, line 7283: Don't use C99, std::numeric_limits<T> is the C++ way.

ZehMatt (changes requested):

src/openrct2/ride/Vehicle.cpp
spacek531: I thought about this but it also `animationState +=` and it would cause a bunch of duplicate unnecessary lines.

src/openrct2/ride/Vehicle.cpp
spacek531: this comment is greek to me

spacek531: What member function do I call to get width? `std::numeric_limits<uint8_t>::width()`?

src/openrct2/ride/Vehicle.cpp
spacek531: Even if I remove `animationState +=` there's use for this as a member function for https://github.com/OpenRCT2/OpenRCT2/pull/15398

Is it better to rewrite it from member function to static function in this PR then rewrite it from static function to member function in the follow-up PR, or keep it a member function from the get-go?

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7273,33 +7276,38 @@ void Vehicle::UpdateAnimationAnimalFlying()
     }
 }
 
+uint8_t Vehicle::GetTargetFrame(rct_ride_entry_vehicle* vehicleEntry)
+{
+    // Similar to a modulus, but using a mask and multiplying before a final bit shift to get rid of the mask bits. You are left
+    // with a value from 0 to AnimationFrames - 1. Prevents skipped frames on non powers-of-two AnimationFrames.
+    return (((animationState >> std::clamp(vehicleEntry->AnimationSpeed - UINT8_WIDTH, 0, UINT32_WIDTH))
+             & std::numeric_limits<uint8_t>::max())
```
spacek531, line 7284: Using numeric limits actually makes this less legible in my opinion, but who am I to read my own code?

spacek531:
```suggestion
                 return (((animationState >> std::clamp(
                  vehicleEntry->AnimationSpeed - std::numeric_limits<uint8_t>::digits, 0,
                  std::numeric_limits<uint32_t>::digits))
             & std::numeric_limits<uint8_t>::max())
            * vehicleEntry->AnimationFrames)
        >> std::numeric_limits<uint8_t>::digits;
```
Here it is in full C++ incomprehensibility

src/openrct2/ride/Vehicle.cpp
spacek531: Alright, you got your wish.

src/openrct2/ride/Vehicle.cpp
ZehMatt: > Using numeric limits actually makes this less legible in my opinion, but who am I to read my own code?

No one forced you do use it in a horrible way. Just split up the single line into multiple statements and it will become a lot easier already, for the limits you can define constexpr constants.

distribution/changelog.txt
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#3868] Initial support for using TTF in OpenGL mode (still contains bugs).
 - Feature: [#7682] Follow ride/guest/staff in main window viewport.
+- Feature: [#11262] Speed property for vehicle animations in .parkobj
```
Broxzier, line 5: https://github.com/OpenRCT2/OpenRCT2/issues/11262 has nothing to do with this.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7419,15 +7412,25 @@ void Vehicle::UpdateAdditionalAnimation()
                     else
                         seat_rotation++;
 
-                    animation_frame = (seat_rotation - 4) & 7;
-                    Invalidate();
+                    int8_t targetSeatRotation = (seat_rotation - 4);
```
Broxzier, line 7415: What's the -4 for?

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7419,15 +7412,25 @@ void Vehicle::UpdateAdditionalAnimation()
...
+                    if (targetSeatRotation != animation_frame)
+                    {
+                        if (targetSeatRotation >= 0)
+                            animation_frame = targetSeatRotation;
+                        else
+                            // % is remainder function in C. This requires some reworking to use as a modulus for
+                            // negative numbers
+                            animation_frame = (vehicleEntry->AnimationFrames - 1)
+                                - ((-targetSeatRotation - 1) % vehicleEntry->AnimationFrames);
```
Broxzier, line 7424: The module operator works fine for negative integers too. You may be having troubles because this seems to be working with offsets and masks before. This could use some rewriting to make the code understandable.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7272,6 +7272,23 @@ void Vehicle::UpdateAnimationAnimalFlying()
         animationState = frameWaitTimes[animation_frame];
     }
 }
+/**
+ * Get the frame of animation for the current animationState based on animation speed and animation frames
+ */
+static uint8_t GetTargetFrame(rct_ride_entry_vehicle* vehicleEntry, uint32_t animationState)
```
Broxzier, line 7278: What does `animationState` mean? It's not very descriptive.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7322,92 +7334,73 @@ void Vehicle::UpdateAdditionalAnimation()
                             }();
                             int32_t directionIndex = sprite_direction >> 1;
                             auto offset = SteamParticleOffsets[typeIndex][directionIndex];
-                            SteamParticle::Create({ x + offset.x, y + offset.y, z + offset.z });
+                            SteamParticle::Create(CoordsXYZ(
+                                x + offset.x * vehicleEntry->SteamEffect.Longitudinal / SteamEffectTranslationCoefficient,
+                                y + offset.y * vehicleEntry->SteamEffect.Longitudinal / SteamEffectTranslationCoefficient,
+                                z + offset.z * vehicleEntry->SteamEffect.Vertical / SteamEffectTranslationCoefficient));
```
Broxzier, line 7340: "Coefficient" doesn't seem right when it's used for division.

Broxzier (changes requested):

distribution/changelog.txt
spacek531: This was a typo

src/openrct2/ride/Vehicle.cpp
spacek531: Interpreting `-1 % 8` was returning 255 and failing replays. This does not fail replays. Dunno why.

src/openrct2/ride/Vehicle.cpp
spacek531: It depends on the animation type. For most animation types, animationState is the sum of _vehicleVelocityF64E08. For others, it's a countup value. For one, it's a countdown value.

src/openrct2/ride/Vehicle.cpp
spacek531: What would you call it then?

src/openrct2/ride/Vehicle.cpp
spacek531: A seat rotation of 4 is 0 degrees:
https://github.com/OpenRCT2/OpenRCT2/blob/1a35cdafb364fe5f94c0edfb4e763371c61b8208/src/openrct2-ui/windows/RideConstruction.cpp#L478

src/openrct2/ride/Vehicle.cpp
spacek531: The only animation types that call this number are the first. Since this was converted from a member function to a static function, I kept the original name.

src/openrct2/ride/Vehicle.cpp
spacek531: I found a full explanation of why the modulus operator doesn't work, although not why `-1 % 8` returns 255.

C integer division rounds toward zero.
C also defines the modulus operator as `a = a / b + a % b`

Since a = -1 and b = 8, the equation becomes:
`-1 = (-1) / 8 + (-1) % 8`
Integer division is easy to solve, it's just zero. Therefore:
`-1 = (-1) % 8`

spacek531:
Any news on this? It's less than 24 hours until the deadline.

duncanspumpkin:
We are pretty much in a feature freeze with trying to get the NSF ready to go. This is too complex to think about atm.

spacek531:
I wish that was communicated earlier.

spacek531:
I'm trying to clean out old PRs. I hope this passes muster now.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9858,3 +9855,41 @@ void Vehicle::Serialise(DataSerialiser& stream)
...
+    { 0x3333, 8, VehicleEntryAnimation::MultiDimCoaster },  // Multi Dimension Coaster
+    { 24, 4, VehicleEntryAnimation::AnimalFlying },         // Animal Flying
+};
+
+const std::string VehicleEntryAnimationNames[VehicleEntryAnimation::Count] = {
+    "none",          "simpleVehicle",   "steamLocomotive",  "swanBoat",
+    "monorailCycle", "multiDimCoaster", "observationTower", "animalFlying",
+};
+
+uint8_t GetAnimationTypeFromString(std::string animationType)
```
ZehMatt, line 9878:
```suggestion
uint8_t GetAnimationTypeFromString(std::string_view animationType)
```

src/openrct2/ride/Vehicle.h
```diff
@@ -410,21 +410,35 @@ enum class MiniGolfAnimation : uint8_t
     Putt,
 };
 
-enum
+// Steam spawn position is multiplied by vehicle entry multiplier then divided by this value
+constexpr const uint8_t SteamEffectTranslationCoefficient = 64;
+
+enum VehicleEntryAnimation : uint8_t
```
ZehMatt, line 416:
```suggestion
enum class VehicleEntryAnimation : uint8_t
```

src/openrct2/ride/Vehicle.cpp
ZehMatt: Also the only place this is used is RideObject.cpp, better move it in there as a static function.

ZehMatt: Also this should result ```VehicleEntryAnimation``` and not some integer type, currently its unclear what this is even supposed to do, assuming thats what this function is doing currently?

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9858,3 +9855,41 @@ void Vehicle::Serialise(DataSerialiser& stream)
...
+    { 1 << 11, 7, VehicleEntryAnimation::SimpleVehicle },   // Rowboat
+    { 1 << 10, 2, VehicleEntryAnimation::SimpleVehicle },   // Water Tricycle
+    { 0x3333, 8, VehicleEntryAnimation::ObservationTower }, // Observation Tower
+    { 1 << 10, 4, VehicleEntryAnimation::SimpleVehicle },   // Mini Helicopter
+    { 1 << 11, 4, VehicleEntryAnimation::MonorailCycle },   // Monorail Cycle
+    { 0x3333, 8, VehicleEntryAnimation::MultiDimCoaster },  // Multi Dimension Coaster
+    { 24, 4, VehicleEntryAnimation::AnimalFlying },         // Animal Flying
+};
+
+const std::string VehicleEntryAnimationNames[VehicleEntryAnimation::Count] = {
```
ZehMatt, line 9873: Use ```EnumMap<VehicleEntryAnimation>``` instead, see Cursors.cpp on how its used.

ZehMatt (changes requested):
Left some comments

src/openrct2/ride/Vehicle.cpp
spacek531: yes, it is intended to get `VehicleEntryAnimation` from string. I forgot that I could return `VehicleEntryAnimation` as a type

spacek531:
currently passing my own replays with the existing objects as well as these modified versions
[modified animation.zip](https://github.com/OpenRCT2/OpenRCT2/files/8103454/modified.animation.zip)

src/openrct2/scripting/bindings/object/ScObject.hpp
```diff
@@ -318,7 +318,7 @@ namespace OpenRCT2::Scripting
             auto entry = GetEntry();
             if (entry != nullptr)
             {
-                return entry->animation;
+                return static_cast<uint8_t>(entry->animation);
```
ZehMatt, line 321:
```suggestion
                return EnumValue(entry->animation);
```

ZehMatt (approved):
Code wise it looks fine to me, there are some open discussions from other team members however.

spacek531:
It's been 31 days since the last update. Have the discussions come to any conclusions in this time?

src/openrct2/ride/Vehicle.cpp
tupaschoal: @Broxzier

src/openrct2/ride/Vehicle.cpp
tupaschoal: @Broxzier

src/openrct2/ride/Vehicle.cpp
tupaschoal: @Broxzier

tupaschoal:
I think he meant some conversations that were still unresolved. I tagged people there to get their attention

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: Rewrite this section completely its a mess
do something like the following there is no need to modulo anything as we already have an if:
```cpp
if (seat_rotation - 4 < 0){
   animation_frame = seat_rotation + (vehicleEntry->AnimationFrames - 1) - 4;
}else{
    animation_frame = seat_rotation - 4;
}
Invalidate();
```

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: I think its fine. I've used factor in some places but coefficient / factor they are all fine.

spacek531:
Would it mess with the metrics if I force-pushed the feature to the old PR and left this as the refactor?

spacek531 (commented):
This fails on any multim animation with fewer than 4 frames

tupaschoal (commented):
> Would it mess with the metrics if I force-pushed the feature to the old PR #11254 and left this as the refactor? Is it desirable to do so at this point?

I think it would be messy indeed. If it's relatively straight forward for you to squash some commits and reorder so that these PR is comprised of 

1. refactor commit(s)
2. implementation commits(s)

Then I think it's worth, because we can merge without squashing and this should ease any future bisecting effort.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7151,114 +7160,101 @@ void Vehicle::UpdateAdditionalAnimation()
                             }();
                             int32_t directionIndex = sprite_direction >> 1;
                             auto offset = SteamParticleOffsets[typeIndex][directionIndex];
-                            SteamParticle::Create({ x + offset.x, y + offset.y, z + offset.z });
+                            SteamParticle::Create(CoordsXYZ(
+                                x + offset.x * vehicleEntry->SteamEffect.Longitudinal / SteamEffectTranslationCoefficient,
+                                y + offset.y * vehicleEntry->SteamEffect.Longitudinal / SteamEffectTranslationCoefficient,
+                                z + offset.z * vehicleEntry->SteamEffect.Vertical / SteamEffectTranslationCoefficient));
```
spacek531, line 7166: I just remembered OpenLoco exists. I think backporting it could be a better way to handle this, under the assumption it looks better than this integer division.

src/openrct2/network/NetworkBase.cpp
```diff
@@ -42,7 +42,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "23"
+#define NETWORK_STREAM_VERSION "25"
```
spacek531, line 45: Do not squash merge

src/openrct2/network/NetworkBase.cpp
```diff
@@ -42,7 +42,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "24"
+#define NETWORK_STREAM_VERSION "26"
```
spacek531, line 45: Do not squash merge

spacek531:
I mentioned using OpenLoco code to calculate the steam positions and now I've gone ahead and done it. It will break replays, but the steam effect is better matched with the visuals of the vehicle and improves the effect in my opinion.

src/openrct2/ride/SteamPosition.hpp
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+#include "../world/Location.hpp"
+
+namespace OpenRCT2::Math::Trigonometry
+{
+    // 0x00503B6A
```
duncanspumpkin, line 15: Remove all loco offsets they mean nothing in this code

src/openrct2/ride/SteamPosition.hpp
```diff
@@ -0,0 +1,99 @@
...
+        return (yawToDirectionVector[yaw] * magnitude) / 256;
+    }
+    constexpr CoordsXY computeXYVector(int32_t height, uint8_t pitch, uint8_t yaw)
+    {
+        return computeXYVector(computeXYMagnitude(height, pitch), yaw);
+    }
+    constexpr CoordsXYZ computeXYZVector(int32_t height, int32_t run, uint8_t pitch, uint8_t yaw)
+    {
+        int32_t sin = pitchSin[static_cast<uint8_t>(pitch)];
+        int32_t cos = pitchCos[static_cast<uint8_t>(pitch)];
```
duncanspumpkin, line 93: Avoid using sin and cos although not protected words they are in the c api so may cause name collisions.

src/openrct2/ride/SteamPosition.hpp
```diff
@@ -39,7 +39,7 @@ namespace OpenRCT2::Math::Trigonometry
      * Where Y1 represents the angle of pitch in degrees
      * Using physical pitch calculated by gravity
      */
-    constexpr CoordsXY pitchToDirectionVector[60] = {
+    constexpr CoordsXY pitchOffsetsFromPhysics[60] = {
```
duncanspumpkin, line 41: This name means a lot less than before. Previously it meant converting a pitch to a direction vector now it just means garbage and physics has nothing to do with this. Can you also apply the coding style as mentioned in the last comment.

duncanspumpkin:
These last few commits don't make any sense. Why are you making the sums more hard to follow. Previously it was all labelled as per normal maths what does `pitchOffsetsFromGeometry` and `pitchOffsetsFromPhysics` mean. Why are they identical???

spacek531:
The physics of the game are not identical to how the sprites are drawn. I was using the physics table, but since the steam particles should match the sprites, it isn't correct to use that table. The notable differences are the diagonal slopes and helix slopes.

What do you want me to change to make it match the style? I realized that OpenRCT2 uses PascalCase constants, are there any other style issues?

duncanspumpkin:
Yes we use UpperCamelCase for pretty much everything. I should be able to read this and understand what mathematical functions are being used. It's meant to be calculating a unit vector based off of a magnitude and direction where direction is yaw and pitch. The part where you mess about with alternative definitions for pitch should not be part of this function that is very specific to your steam code and would never be used for anything else so it shouldn't be under Trig.

spacek531:
I don't know which table you are talking about not being part of the namespace. Locomotion also has this table, but it didn't have as many discrepancies between the physics and the visuals as RCT2 does making two tables useful.

One table uses the physics of RCT2 to calculate the components, using the AccelerationFromPitch table and calculating what angle the gravity constant represents using the formula asin(gravity/642000)

The other table uses the geometry of RCT2 to calculate the components. Trigonometry is used to determine the angles that vehicle sprites should be at for each pitch, and the components of those angles are recorded accordingly. I used X123M3-256's work for his ride generator, found here https://github.com/X123M3-256/RCT2RideGenerator/blob/39b3847a8f9277f5c5af23320f6308f01ac14fb4/src/dat.h#L13-L34

This refactor originally used the former, the physics table, to calculate the steam position, but it is better to use the other table, the geometry table, because the purpose of the feature is to match the sprite renders for position, not the physics.

I'm fine with moving one table out, if you feel it doesn't deserve to be in the Trigonometry namespace, I just don't know which one you consider to be 'real' trigonometry.

src/openrct2/ride/SteamPosition.hpp
```diff
@@ -0,0 +1,195 @@
...
+    };
+
+    /**
+     * The cos and sin of vehicle pitch
+     * Calculated using gravity constants
+     * X represents vertical component and Y represents horizontal component
+     * COS((Y1/360)*2*PI())*256,-SIN((Y1/360)*2*PI())*256
+     * Where Y1 represents the angle of pitch in degrees
+     */
+    constexpr CoordsXY PItchToDirectionVectorFromPhysics[60] = {
```
spacek531, line 41: Fixed, just waiting on a bigger change to run the CI for.

rik-smeets:
Do you still want to continue working on this PR @spacek531? Or are you awaiting @duncanspumpkin's reply?

spacek531:
I do want to complete this PR yes

spacek531:
[steam test.zip](https://github.com/OpenRCT2/OpenRCT2/files/10892772/steam.test.zip)

This park has packed objects which use the new animation system. 

I don't have anything else for this PR now, so I am marking it as ready for review.

src/openrct2/object/RideObject.cpp
```diff
@@ -383,8 +433,7 @@ uint8_t RideObject::CalculateNumVerticalFrames(const CarEntry& carEntry)
     {
         if (!(carEntry.flags & CAR_ENTRY_FLAG_SPINNING_ADDITIONAL_FRAMES))
         {
-            if (carEntry.flags & CAR_ENTRY_FLAG_VEHICLE_ANIMATION
-                && carEntry.animation != CAR_ENTRY_ANIMATION_OBSERVATION_TOWER)
+            if (carEntry.flags & CAR_ENTRY_FLAG_VEHICLE_ANIMATION && carEntry.animation != CarEntryAnimation::ObservationTower)
```
ZehMatt, line 436:
```suggestion
            if ((carEntry.flags & CAR_ENTRY_FLAG_VEHICLE_ANIMATION) && carEntry.animation != CarEntryAnimation::ObservationTower)
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6644,156 +6610,124 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
-    uint32_t eax{};
-
     auto carEntry = Entry();
     if (carEntry == nullptr)
     {
         return;
     }
+    if (carEntry->AnimationFrames == 0)
+        return;
     switch (carEntry->animation)
```
ZehMatt, line 6620: Could you perhaps move all those cases out into new functions? Having it all slammed into one function feels quite wrong.

ZehMatt (changes requested):
Just some minor things otherwise I think its reasonable

src/openrct2/ride/Vehicle.cpp
spacek531: What is your preference for doing that? I can have function calls in a switch or I could put all the functions in an array and call based on `static_cast<uint8_t>(animation)` similar to choosing the vehicle paint pitch function.

ZehMatt: You can either keep the switch or use a function table with a common signature, both works for me. It will be a lot easier to make out the functionality of each animation type when they are clearly separated.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6720,13 +6720,13 @@ static void AnimateMultiDimension(Vehicle& vehicle, const CarEntry& carEntry)
             else
                 vehicle.seat_rotation++;
 
-            int8_t targetSeatRotation = (vehicle.seat_rotation - 4);
+            int16_t targetSeatRotation = vehicle.seat_rotation - 4;
+            // cpp modulo allows negative values. This is the workaround.
+            while (targetSeatRotation < 0)
+                targetSeatRotation += carEntry.AnimationFrames;
+
```
spacek531, line 6727: This seems like a much simpler workaround to c modulo than before.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6638,163 +6579,191 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
+    {
+        return;
     }
+    if (carEntry->AnimationFrames == 0 || carEntry->animation >= CarEntryAnimation::Count)
+        return;
+    using AnimateFunction = void (*)(Vehicle & vehicle, const CarEntry& carEntry);
+    constexpr static AnimateFunction animationFunctions[EnumValue(CarEntryAnimation::Count)]{
+        AnimateNone,          AnimateSimpleVehicle,  AnimateSteamLocomotive,  AnimateSwanBoat,
+        AnimateMonorailCycle, AnimateMultiDimension, AnimateObservationTower, AnimateAnimalFlying
+    };
```
ZehMatt, line 6765: I personally think this should be outside the function, also if you add a comma to the end clang-format will have this like a table, in theory.

spacek531: I can do that. Trailing comma does not change formatting though.

src/openrct2/ride/Vehicle.cpp
ZehMatt: This is unfortunately much slower than a modulo, what exactly was the problem with negative modulo? You could always convert it to a positive with std::abs which is also almost zero cost.

src/openrct2/ride/Vehicle.cpp
ZehMatt: Too bad, at least outside the function it doesn't really get in the way of reading the code, thanks.

src/openrct2/ride/Vehicle.cpp
spacek531: The problem is it doesn't wrap very well. The existing *should* work like this:

seat_rotation | animation_frame (8 sprites: vanilla multidim) | animation_frame (3 sprites: example)
---|---|---
0 | 4 | 2
1 | 5 | 0
2 | 6 | 1
3 | 7 | 2
4 | 0 | 0
5 | 1 |1
6 | 2|2
7 | 3|0
8 | 4|1
9 | 5|2
10 | 6|0
11 | 7|1
12 | 0|2
13 | 1|0
14 | 2|1
15 | 3|2

So I am trying to make animation_frame loop seamlessly with 4->0 being the anchor.

Doing something like:

```cpp
int16_t targetSeatRotation = std::abs((vehicle.seatRotation - 4) % carEntry.AnimationFrames);
```
will cause the values to go backwards when below 4, and doing something like:
```cpp
uint16_t targetSeatRotation = vehicle.seatRotation - 4
targetSeatRotation %= carEntry.AnimationFrames;
```
will work with powers of two, but does weird stuff when it isn't.

ZehMatt: Your table doesn't show any negative inputs, so this function would work for 8 sprites:
```cpp
int transform(int input) {
    int output = (input + 4) % 8;
    return output;
}```
for 3 sprites this should work:
```cpp
int transform(int input) {
    int output = (input % 3 + 2) % 3;
    return output;
}```
thats according to your truth table.

Comment (spacek531) [reply src/openrct2/ride/Vehicle.cpp:6731]: I've pushed an update that unrolls the loop. The issue only happens when AnimationFrames is 1, 2, or 3. 1 is pointless because it's basically no animation, so 2 and 3 are the only ones that need special treatment:
```cpp
            int16_t targetSeatRotation = (vehicle.seat_rotation - 4) % carEntry.AnimationFrames;
            if (targetSeatRotation < 0)
            {
                targetSeatRotation += carEntry.AnimationFrames;
                // fix negative values when AnimationFrames is less than 4
                if (targetSeatRotation < 0)
                    targetSeatRotation += carEntry.AnimationFrames;
            }
            ```

Comment (ZehMatt) [reply src/openrct2/ride/Vehicle.cpp:6731]: Not sure I fully grasp the animation stuff but if you were to add 4 instead of subtracting then you would never have any negative values, its only negative for when seat_rotation is below 4 but the wrapping is the same if you use an addition of 4 instead. Maybe the calculation should be different per animation type? I can't quite tell for sure.

File: src/openrct2/math/Trigonometry.hpp [hunk 52]
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    constexpr CoordsXYZ ComputeSteamOffset(int32_t height, int32_t length, uint8_t pitch, uint8_t yaw)
+    {
+        uint8_t trueYaw = OpenRCT2::Entity::Yaw::YawTo64(yaw);
+        auto offsets = PitchToDirectionVectorFromGeometry[pitch];
+        int32_t projectedRun = (offsets.x * length - offsets.y * height) / 256;
+        int32_t projectedHeight = (offsets.x * height + offsets.y * length) / 256;
+        return { ComputeXYVector(projectedRun, trueYaw), projectedHeight };
+    }
```
ZehMatt, line 125: This probably belongs to SteamParticle.hpp

spacek531: Instead of creating a file just for this one function, could it be in Vehicle.cpp?

ZehMatt: I didn't realize you deleted it, yes.

ZehMatt (approved):
Just one thing that needs to be resolved after that I'm fine with this PR.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6638,163 +6580,209 @@ void Vehicle::UpdateAnimationAnimalFlying()
...
-        case CAR_ENTRY_ANIMATION_ANIMAL_FLYING:
-            UpdateAnimationAnimalFlying();
-            // makes animation play faster with vehicle speed
-            targetFrame = abs(_vehicleVelocityF64E08) >> 24;
-            animationState = std::max(animationState - targetFrame, 0u);
-            break;
+constexpr static const AnimateFunction animationFunctions[EnumValue(CarEntryAnimation::Count)]{
+    AnimateNone,          AnimateSimpleVehicle,  AnimateSteamLocomotive,  AnimateSwanBoat,
+    AnimateMonorailCycle, AnimateMultiDimension, AnimateObservationTower, AnimateAnimalFlying,
+};
```
Broxzier, line 6770: Using `static_assert` is a better way to assure the size is filled correctly, otherwise it'll just zero-initialize new entries when `Count` increases.
```suggestion
constexpr static const AnimateFunction animationFunctions[]{
    AnimateNone,          AnimateSimpleVehicle,  AnimateSteamLocomotive,  AnimateSwanBoat,
    AnimateMonorailCycle, AnimateMultiDimension, AnimateObservationTower, AnimateAnimalFlying,
};
static_assert(size::size(animationFunctions) == EnumValue(CarEntryAnimation::Count));
```

The same goes for the `SpriteGroupMultiplier` array defined much earlier in the code.

Broxzier: Also use TitleCase for static globals.

src/openrct2/object/RideObject.cpp
```diff
@@ -43,6 +43,55 @@ static const uint8_t SpriteGroupMultiplier[EnumValue(SpriteGroupType::Count)] =
...
+    { 1 << 10, 2, CarEntryAnimation::SwanBoat },        // Swan Boat
+    { 1 << 11, 6, CarEntryAnimation::SimpleVehicle },   // Canoe
+    { 1 << 11, 7, CarEntryAnimation::SimpleVehicle },   // Rowboat
+    { 1 << 10, 2, CarEntryAnimation::SimpleVehicle },   // Water Tricycle
+    { 0x3333, 8, CarEntryAnimation::ObservationTower }, // Observation Tower
+    { 1 << 10, 4, CarEntryAnimation::SimpleVehicle },   // Mini Helicopter
+    { 1 << 11, 4, CarEntryAnimation::MonorailCycle },   // Monorail Cycle
+    { 0x3333, 8, CarEntryAnimation::MultiDimension },   // Multi Dimension Coaster
+    { 24, 4, CarEntryAnimation::AnimalFlying },         // Animal Flying
+};
```
Broxzier, line 80: Also use static_assert to verify the size here, and leave `NumLegacyAnimationTypes` out of the brackets.

```suggestion
constexpr const LegacyAnimationParameters VehicleEntryDefaultAnimation[] = {
    { 0, 1, CarEntryAnimation::None },                  // None
    { 1 << 12, 4, CarEntryAnimation::SteamLocomotive }, // Miniature Railway Locomotive
    { 1 << 10, 2, CarEntryAnimation::SwanBoat },        // Swan Boat
    { 1 << 11, 6, CarEntryAnimation::SimpleVehicle },   // Canoe
    { 1 << 11, 7, CarEntryAnimation::SimpleVehicle },   // Rowboat
    { 1 << 10, 2, CarEntryAnimation::SimpleVehicle },   // Water Tricycle
    { 0x3333, 8, CarEntryAnimation::ObservationTower }, // Observation Tower
    { 1 << 10, 4, CarEntryAnimation::SimpleVehicle },   // Mini Helicopter
    { 1 << 11, 4, CarEntryAnimation::MonorailCycle },   // Monorail Cycle
    { 0x3333, 8, CarEntryAnimation::MultiDimension },   // Multi Dimension Coaster
    { 24, 4, CarEntryAnimation::AnimalFlying },         // Animal Flying
};
static_assert(std::size(VehicleEntryDefaultAnimation) == NumLegacyAnimationTypes);
```

src/openrct2/ride/Vehicle.cpp
spacek531: Sprite Group Multiplier would be more appropriately fixed in https://github.com/OpenRCT2/OpenRCT2/pull/19567

src/openrct2/math/Trigonometry.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+namespace OpenRCT2::Math::Trigonometry
+{
+    /**
+     * The cos and sin of sprite direction
+     * ROUND(COS((32/64+(L1/64))*(2*PI()))*256,0), ROUND(SIN(((L1/64))*(2*PI())) * 256,0)
+     * Where L1 represents an incrementing column 0 - 63
+     * Note: Must be at least 32bit to ensure all users do not overflow
+     */
+    static constexpr CoordsXY YawToDirectionVector[64] = {
```
Broxzier, line 22: Could you also apply the array size check to this one?

src/openrct2/math/Trigonometry.hpp
```diff
@@ -0,0 +1,118 @@
...
+        { -181, -181 }, { -198, -162 }, { -213, -142 }, { -226, -121 }, { -237, -98 },  { -245, -74 },  { -251, -50 },
+        { -255, -25 },
+    };
+
+    /**
+     * The cos and sin of vehicle pitch based on vehicle sprite angles
+     * COS((Y1/360)*2*PI())*256,-SIN((Y1/360)*2*PI())*256
+     * Where Y1 represents the angle of pitch in degrees
+     */
+    constexpr CoordsXY PitchToDirectionVectorFromGeometry[] = {
```
Broxzier, line 40: And this one. Size = 61?
All new static arrays should have this.

spacek531:
😄 feels like the end of an era

PR #11273, opened 2020-04-11
Make casualty notifications optional

### LINKED ISSUE #11272: Notification for how many people died in an accident cannot be disabled
**OS:** [Windows 10]
**Version:** [0.2.5]
**Commit/Build:** [b5336fc]

I disabled all notifications on the different tabs that are available, but the message for how many people died in an accident still pops up every time.

rideaccident

AaronVanGeffen:
This adds an option for toggling notifications for 'Ride casualties' and 'Stuck or stalled vehicles' to the Notification Settings window:

Diamond Heights 2020-04-11 14-41-23

Fixes #11272

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -5206,8 +5206,11 @@ static void vehicle_kill_all_passengers(Vehicle* vehicle)
 
     if (numFatalities != 0)
     {
-        ride->FormatNameTo(gCommonFormatArgs + 2);
-        news_item_add_to_queue(NEWS_ITEM_RIDE, STR_X_PEOPLE_DIED_ON_X, vehicle->ride);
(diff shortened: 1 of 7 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -5207,7 +5207,11 @@ static void vehicle_kill_all_passengers(Vehicle* vehicle)
     if (numFatalities != 0)
     {
         ride->FormatNameTo(gCommonFormatArgs + 2);
```
Gymnasiast, line 5209: This can probably be moved inside the `if` as well.

AaronVanGeffen: Good point.

Gymnasiast:
Seems low-risk enough to get into 0.2.6. Could you amend the changelog?

PR #11274, opened 2020-04-11
Fix #11075 SpriteListHead Refactor

### LINKED ISSUE #11075: SpriteListHead Refactor Peep / Staff
Really easy little refactor. We use the following macro to loop over all peeps:
```
FOR_ALL_PEEPS(sprite_index, peep) 
```
It basically does something similar to:
```
  for (spriteIndex = gSpriteListHead[SPRITE_LIST_PEEP]; spriteIndex != SPRITE_INDEX_NULL; spriteIndex = nextSpriteIndex)
        {
            auto peep = &(get_sprite(spriteIndex)->peep);
            nextSpriteIndex = peep->next;
```
But there are a number of cases in the codebase where we have just written out the whole big for loop again. This issue is to refactor all of those needless repeats to use the macro. There is also `FOR_ALL_GUESTS` and `FOR_ALL_STAFF` it should be pretty easy to work out when those macros can be used.

To get you started https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/actions/SetCheatAction.hpp#L688-L696 is an example of a `FOR_ALL_GUESTS` refactor location. Easiest way to find more is do a search for `gSpriteListHead[SPRITE_LIST_PEEP]`.

TomasZilinek:
Sorry it lasted so long. I was swamped for some time. Hope it's fine.

### DIFF
--- a/src/openrct2/actions/SetCheatAction.hpp
+++ b/src/openrct2/actions/SetCheatAction.hpp
@@ -642,7 +642,7 @@ private:
 
     void RemoveAllGuests() const
     {
-        uint16_t spriteIndex, nextSpriteIndex;
+        uint16_t spriteIndex;
         for (auto& ride : GetRideManager())
         {
             ride.num_riders = 0;
@@ -685,14 +685,12 @@ private:
             }
         }
 
-        for (spriteIndex = gSpriteListHead[SPRITE_LIST_PEEP]; spriteIndex != SPRITE_INDEX_NULL; spriteIndex = nextSpriteIndex)
+        Peep* peep;
+        uint16_t sprite_index;
+
+        FOR_ALL_GUESTS (sprite_index, peep)
         {
-            auto peep = &(get_sprite(spriteIndex)->peep);
-            nextSpriteIndex = peep->next;
-            if (peep->type == PEEP_TYPE_GUEST)
-            {
-                peep->Remove();
-            }
+            peep->Remove();
         }
 
         window_invalidate_by_class(WC_RIDE);
--- a/src/openrct2/actions/SetCheatAction.hpp
+++ b/src/openrct2/actions/SetCheatAction.hpp
@@ -686,9 +686,7 @@ private:
         }
 
         Peep* peep;
-        uint16_t sprite_index;
-
-        FOR_ALL_GUESTS (sprite_index, peep)
+        FOR_ALL_GUESTS (spriteIndex, peep)
         {
             peep->Remove();
         }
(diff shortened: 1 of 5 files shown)


tupaschoal:
Hello @TomasZilinek, can you please update the commit message/PR description to better match our guidelines? 

There's some info on https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

duncanspumpkin:
@TomasZilinek this looks good. When we merge we will squash all of these commits into one.

duncanspumpkin:
It might be a little while until we merge though as we have a few other items right now that need completed.

src/openrct2/actions/SetCheatAction.hpp
```diff
@@ -685,14 +685,12 @@ DEFINE_GAME_ACTION(SetCheatAction, GAME_COMMAND_CHEAT, GameActionResult)
             }
         }
 
-        for (spriteIndex = gSpriteListHead[SPRITE_LIST_PEEP]; spriteIndex != SPRITE_INDEX_NULL; spriteIndex = nextSpriteIndex)
+        Peep* peep;
+        uint16_t sprite_index;
```
duncanspumpkin, line 689: you can just reuse the already defined `spriteIndex` instead of creating another one.

src/openrct2/peep/Peep.cpp
```diff
@@ -404,19 +404,15 @@ int32_t peep_get_staff_count()
...
 
     if (gScreenFlags & SCREEN_FLAGS_EDITOR)
         return;
-
-    spriteIndex = gSpriteListHead[SPRITE_LIST_PEEP];
-    i = 0;
-    while (spriteIndex != SPRITE_INDEX_NULL)
+    FOR_ALL_PEEPS (spriteIndex, peep)
     {
         peep = &(get_sprite(spriteIndex)->peep);
```
duncanspumpkin, line 415: This line is no longer required, it will be performed by the macro.

src/openrct2/peep/Staff.cpp
```diff
@@ -152,8 +153,7 @@ void staff_update_greyed_patrol_areas()
             gStaffPatrolAreas[staffPatrolOffset + i] = 0;
         }
 
-        for (uint16_t sprite_index = gSpriteListHead[SPRITE_LIST_PEEP]; sprite_index != SPRITE_INDEX_NULL;
-             sprite_index = peep->next)
+        FOR_ALL_STAFF (sprite_index, peep)
         {
             peep = GET_PEEP(sprite_index);
```
duncanspumpkin, line 158: As above this is no longer required and can be removed.

tupaschoal:
Labelling as WIP while there are still pending comments

duncanspumpkin:
I have made the review changes and squashed the commits as I want this in before some sprite rework can you review @tupaschoal.

PR #11280, opened 2020-04-12
Avoid fast-forwarding peep to ride as it may not have funds to pay

brenoguim:
The code being removed in the patch tries to fast forward a peep into the ride when it is the second peep for a vehicle that is used in pairs. Problem is that funds checking does not happen, so it happens that a peep may pay against its will.
Lets say a rich peep enters in line and a poor peep enters in line right after.
If the price of the ride is such that the rich peep can pay and the poor peep can't, it will be dragged into the ride because funds checking only happened for the first.
The second part of the patch just adjusts we consider the vehicle a full car if the second position is filled.
As a bonus, the function "Guest::FindVehicleToEnter" now really only *finds* a vehicle, and never actually enters.

(I just noticed I didn't branch off "develop" but added the commit on develop direcly. Sorry for that. Let me know if I have to fix that for this PR.)

I noticed this bug by inspecting the code and came up with this scenario:

[peep_pays_more_than_it_has.zip](https://github.com/OpenRCT2/OpenRCT2/files/4466492/peep_pays_more_than_it_has.zip)
In this scenario, two peeps are in queue. One is rich and the second has only 1 dolar. I raised the ride cost to 1.10 dolars.
before

Without the patch, the peep will pay 1.10 even though it doesn't have it.
After

With the patch, the second peep turns around and leave, and the third peep goes into the ride.
Now added the third peep in line too.
Before_fixed
The poor peep goes back and leave the queue (in yellow shirt) and third one goes in.
After_fixed

This also happens in the original game. Thanks @tupaschoal for reproducing this in the original game for me :)

[OriginalSave_PeepOnVergeOverspending.zip](https://github.com/OpenRCT2/OpenRCT2/files/4466496/OriginalSave_PeepOnVergeOverspending.zip)
before_original
after_original

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -2595,24 +2595,13 @@ bool Guest::FindVehicleToEnter(Ride* ride, std::vector<uint8_t>& car_array)
         vehicle = GET_VEHICLE(vehicle_id);
 
         uint8_t num_seats = vehicle->num_seats;
-        if (vehicle_is_used_in_pairs(vehicle))
-        {
-            num_seats &= VEHICLE_SEAT_NUM_MASK;
-            if (vehicle->next_free_seat & 1)
-            {
-                current_car = i;
-                peep_choose_seat_from_car(this, ride, vehicle);
-                GoToRideEntrance(ride);
-                return false;
-            }
-        }
         if (num_seats == vehicle->next_free_seat)
             continue;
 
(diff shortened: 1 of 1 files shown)


brenoguim:
[peep_pays_more_than_it_has.zip](https://github.com/OpenRCT2/OpenRCT2/files/4466463/peep_pays_more_than_it_has.zip)

brenoguim:
I just noticed that another side effect is the peep correctly getting the thought of not having enough money:
thought

brenoguim:
Closing PR in favor of https://github.com/OpenRCT2/OpenRCT2/pull/11290 where I adjusted the source branch to not be develop

PR #11281, opened 2020-04-12
Add add_news_item to the console

janisozaur:
Feature requested in our discord.


janisozaur:
I was told that previously external tools were able to generate multiline messages and alter colour as well. We have that ability when passing the language data, but not from the console - even entering `{` is prevented, so it's not feature complete right now. But I probably won't be working on adding that.

Issue #11258, opened 2020-04-10
Text on scrolling banners disappears after saving the game


deurklink:
**OS:** Windows 10
**Version:** 0.2.5 - was already present in 0.2.4
**Commit/Build:** fdf9806

![image](https://user-images.githubusercontent.com/6469413/78975812-3f9ced00-7b15-11ea-984e-ee585919cfdf.png)

After saving, the banners in my megapark all lose their text. I also had this problem a few times but with the banners only keeping the first word or first character. I also saw this happen in some other parks.

1. Open attached save file
2. Change the banner text into something
3. Save the game
4. Reload, and watch the text on the banner disappear

**Save game:**

[Deurklink MegaPark.zip](https://github.com/OpenRCT2/OpenRCT2/files/4460598/Deurklink.MegaPark.zip)

deurklink:
When i remove the banner and replace it with a new one, the problem is gone for the new banner. But I think at some point, the value for the banner gets overwritten somewhere and then the problem persists until I replace the banner again.

AaronVanGeffen:
Had a quick look — here are only 84 banners in the park (out of 250), so it's not like the limit is being hit there.

deurklink:
I got rid of this problem by removing all the banners and replacing them with new ones. But then after a few days the problem came back. I don't know what causes it.

janisozaur:
Oh cool, it is not something odd with my own setup. I noticed this with recent replay parks as well. Sometimes the banner scrolled and other times it did not. A banner above the park entrance stopped scrolling in the main viewport, but was scrolling properly in the small one with park/ride details.

janisozaur:
Regression from f539fdd79

janisozaur:
https://github.com/OpenRCT2/OpenRCT2/blob/0cbd7002902e91d5e5023a91aa87dd94332fc07b/src/openrct2/rct2/S6Exporter.cpp#L1625-L1632


```
auto rct2encoded = utf8_to_rct2("dfgdfgdfg");
[...]

Issue #11286, opened 2020-04-13
Banner tooltip colour on hover lacks update


Richard-L:
OpenRCT2, v0.2.5-127-g0cbd70029 (0cbd700 on develop) provided by GitHub
Windows 7 (x86-64)

![image](https://user-images.githubusercontent.com/4934209/79131437-e351fb80-7da8-11ea-89ac-4df23d79f4f1.png)
Place a banner (sign) and hover over it to directly see what it says (cf https://github.com/OpenRCT2/OpenRCT2/issues/8593)

Even if the text colour is set to "white", the hover preview will show in orange.

Save park and exit, then reopen.

![image](https://user-images.githubusercontent.com/4934209/79131345-c1587900-7da8-11ea-8130-01a42733764b.png)
^the hover preview colour is now in the correct white.

Can this update correctly without the need for save/reopen?

Thanks

ps. up for debate I feel is capitalisation like "Right-Click to Modify", which I would say is more adequate as "Right-click to modify".

PR #11289, opened 2020-04-13
Use utf8_remove_format_codes to remove format codes from user strings

brenoguim:
The previous code used RCT12::RemoveFormatCodes to remove format codes from RCT strings. But that function uses utf8 methods to identify format codes. In this new version, rct2_to_utf8 is called first to obtain an utf8 string, and then utf8_remove_format_codes can be used.

Fixes  https://github.com/OpenRCT2/OpenRCT2/issues/11258.  With this patch saving and restoring a park will not erase banner names. Restoring the park in the game causes all blank banners to simply get the default value.

Important to notice that this change makes #11286 even worse. Now the tool tip doesn't get coloured even if you save/restore the map. However, the fix for that should probably be on the tooltip printing routine instead of the save routine, right?
The pull request https://github.com/OpenRCT2/OpenRCT2/pull/11291 shows this idea.

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -2945,8 +2945,9 @@ private:
     {
         const auto originalString = _s4.string_table[(stringId - USER_STRING_START) % 1024];
         std::string_view originalStringView(originalString, USER_STRING_MAX_LENGTH);
-        auto withoutFormatCodes = RCT12::RemoveFormatCodes(originalStringView);
-        return rct2_to_utf8(withoutFormatCodes, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        auto asUtf8 = rct2_to_utf8(originalStringView, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        utf8_remove_format_codes(asUtf8.data(), /*allow colour*/ false);
+        return asUtf8.data();
     }
 
     void FixLandOwnership()
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -1631,8 +1631,9 @@ public:
     {
         const auto originalString = _s6.custom_strings[(stringId - USER_STRING_START) % 1024];
         std::string_view originalStringView(originalString, USER_STRING_MAX_LENGTH);
-        auto withoutFormatCodes = RCT12::RemoveFormatCodes(originalStringView);
-        return rct2_to_utf8(withoutFormatCodes, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        auto asUtf8 = rct2_to_utf8(originalStringView, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        utf8_remove_format_codes(asUtf8.data(), /*allow colour*/ false);
+        return asUtf8.data();
     }
 
     std::vector<rct_object_entry> GetRequiredObjects()
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -19,6 +19,7 @@
 - Fix: [#11225] Replay manager cannot handle track designs.
 - Fix: [#11246] Fix Various Import/Export issues with Boat locations, balloon frame number.
 - Fix: Small red gardens in RCT1 saves are imported in the wrong colour.
+- Fix: [#11258] Properly remove format codes from imported strings.
 - Improved: [#11157] Slimmer virtual floor lines.
 
 0.2.5 (2020-03-24)
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -31,7 +31,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "14"
+#define NETWORK_STREAM_VERSION "15"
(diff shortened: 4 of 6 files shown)


brenoguim:
Marked as ready to review so that CI would kick in

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1629,10 +1629,13 @@ class S6Importer final : public IParkImporter
 
     std::string GetUserString(rct_string_id stringId)
     {
-        const auto originalString = _s6.custom_strings[(stringId - USER_STRING_START) % 1024];
-        std::string_view originalStringView(originalString, USER_STRING_MAX_LENGTH);
-        auto withoutFormatCodes = RCT12::RemoveFormatCodes(originalStringView);
-        return rct2_to_utf8(withoutFormatCodes, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        const auto originalStringC = _s6.custom_strings[(stringId - USER_STRING_START) % 1024];
+        std::string_view originalString(originalStringC, USER_STRING_MAX_LENGTH);
```
tupaschoal, line 1633: Diff would look cleaner if you hadn't touched these lines

brenoguim: Done, thanks!

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2945,8 +2945,11 @@ class S4Importer final : public IParkImporter
     {
         const auto originalString = _s4.string_table[(stringId - USER_STRING_START) % 1024];
         std::string_view originalStringView(originalString, USER_STRING_MAX_LENGTH);
-        auto withoutFormatCodes = RCT12::RemoveFormatCodes(originalStringView);
-        return rct2_to_utf8(withoutFormatCodes, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        auto asUtf8 = rct2_to_utf8(originalStringView, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        auto buf = std::make_unique<char[]>(asUtf8.size() + 1);
+        std::strcpy(buf.get(), asUtf8.data());
```
janisozaur, line 2950: Why can't this be string with pointer to utf8 data and it's length? You still need to convert it to string upon returning

janisozaur:
Looks fine code-wise, but I would like someone to test our banners against vanilla too, to see if the format matches it's expectations.

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2945,8 +2945,11 @@ class S4Importer final : public IParkImporter
...
+        utf8_remove_format_codes(buf.get(), /*allow colour*/ false);
+        return buf.get();
```
janisozaur, line 2952:
```suggestion
        auto asUtf8 = rct2_to_utf8(originalStringView, RCT2_LANGUAGE_ID_ENGLISH_UK);
        std::string buf(asUtf8.data(), asUtf8.size() + 1);
        utf8_remove_format_codes(buf.data(), /*allow colour*/ false);
        return buf;
```

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1631,8 +1631,11 @@ class S6Importer final : public IParkImporter
     {
         const auto originalString = _s6.custom_strings[(stringId - USER_STRING_START) % 1024];
         std::string_view originalStringView(originalString, USER_STRING_MAX_LENGTH);
-        auto withoutFormatCodes = RCT12::RemoveFormatCodes(originalStringView);
-        return rct2_to_utf8(withoutFormatCodes, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        auto asUtf8 = rct2_to_utf8(originalStringView, RCT2_LANGUAGE_ID_ENGLISH_UK);
+        auto buf = std::make_unique<char[]>(asUtf8.size() + 1);
+        std::strcpy(buf.get(), asUtf8.data());
+        utf8_remove_format_codes(buf.get(), /*allow colour*/ false);
+        return buf.get();
```
janisozaur, line 1638:
```suggestion
        auto asUtf8 = rct2_to_utf8(originalStringView, RCT2_LANGUAGE_ID_ENGLISH_UK);
        std::string buf(asUtf8.data(), asUtf8.size() + 1);
        utf8_remove_format_codes(buf.data(), /*allow colour*/ false);
        return buf;
```

janisozaur:
Oh, right, and the changelog entry.

src/openrct2/rct1/S4Importer.cpp
brenoguim: I avoided that because utf8_remove_format_codes will insert a "\0" in the middle of the string. So this string will have a `size()` that is different from `strlen(str.c_str())`. I wasn't sure if the code receiving this string is prepared to handle that. Sometimes developers will assume that.

janisozaur: I'm aware. But in this context we use it as string anyway and don't change the actual buffer length. It still gets converted to string in the end, so the buffer would be truncated at that point. You can add a note to let this be known.

brenoguim: Fair enough. I'll change that.

brenoguim: I actually had to return `buf.data()` instead of `buf` to strip the remaining bytes after the '\0' in the string. The test was still broken if I don't do that. :(

brenoguim:
I think it's all fine now. With your suggestion @janisozaur there is an intermediate variable less. Code is simpler now. I still have to return `asUtf8.data()` so another `std::string` will be created up until the '\0' character only.

PR #11290, opened 2020-04-14
Avoid fast-forwarding peep to ride as it may not have funds to pay

brenoguim:
The code being removed in the patch tries to fast forward a peep into the ride when it is the second peep for a vehicle that is used in pairs. Problem is that funds checking does not happen, so it happens that a peep may pay against its will.
Lets say a rich peep enters in line and a poor peep enters in line right after.
If the price of the ride is such that the rich peep can pay and the poor peep can't, it will be dragged into the ride because funds checking only happened for the first.
The second part of the patch just adjusts we consider the vehicle a full car if the second position is filled.
As a bonus, the function "Guest::FindVehicleToEnter" now really only *finds* a vehicle, and never actually enters.

I noticed this bug by inspecting the code and came up with this scenario:

[peep_pays_more_than_it_has.zip](https://github.com/OpenRCT2/OpenRCT2/files/4466492/peep_pays_more_than_it_has.zip)
In this scenario, two peeps are in queue. One is rich and the second has only 1 dolar. I raised the ride cost to 1.10 dolars.
before

Without the patch, the peep will pay 1.10 even though it doesn't have it.
After

With the patch, the second peep turns around and leave, and the third peep goes into the ride.
Now added the third peep in line too.
Before_fixed
The poor peep goes back and leave the queue (in yellow shirt) and third one goes in.
After_fixed

This also happens in the original game. Thanks @tupaschoal for reproducing this in the original game for me :)

[OriginalSave_PeepOnVergeOverspending.zip](https://github.com/OpenRCT2/OpenRCT2/files/4466496/OriginalSave_PeepOnVergeOverspending.zip)
before_original
after_original


janisozaur:
Nice find

duncanspumpkin:
Can you rebase this. I'm pretty sure this will fail the replay tests.

brenoguim:
@duncanspumpkin indeed a replay test failed. How do you suggest we handle this? I'm quite new here, so I don't know if there is a way to debug this, or verify that the first change in the replay is intended.

I would also be glad to put a specific replay test for this scenario. I mean, creating a small park with just the two peeps and verify that the second one doesn't enter the ride. Then it should be simple for whoever breaks that replay test to verify it.

Another thing that I thought of doing is to add an assert in the code to make sure no peep ever spends more than it has. But I don't know if that is valid some situation.

duncanspumpkin:
Well the good thing is that this proves the replay tests now work. They have been broken for over a year! I need to have a think of any other repercussions of this pr. Give me a day or so. To fix the replay we will have to rerecord it. It's not too hard but a little fiddly there isn't any point doing it until we are sure this pr is good to go though.

brenoguim:
I agree, take your time :) Thanks!

src/openrct2/peep/Guest.cpp
```diff
@@ -2595,24 +2595,13 @@ bool Guest::FindVehicleToEnter(Ride* ride, std::vector<uint8_t>& car_array)
         vehicle = GET_VEHICLE(vehicle_id);
 
         uint8_t num_seats = vehicle->num_seats;
-        if (vehicle_is_used_in_pairs(vehicle))
-        {
-            num_seats &= VEHICLE_SEAT_NUM_MASK;
-            if (vehicle->next_free_seat & 1)
-            {
-                current_car = i;
```
duncanspumpkin, line 2603: You can't remove this bit of code without breaking all other rides (non ferris wheels).
From tests it looks like you could take this case as a success and push_back into the car_array if it gets here instead of fast forwarding to the entrance. That way all of the money checks will still take place.

brenoguim: Good point! Thanks for the attention. I think your suggestion is probably the right choice there. I'll make some more testing and get back with results. :)

brenoguim:
I added the tests to this PR because it makes sense to push the fix + test together showing that it passes.

The text from the other PR was:

This test verifies that a guest, when second in queue, won't be forced to enter the ride if it has not enough money to pay for it.
To simulate this scenario, two guests (a rich and a poor) are encouraged to enter the ride queue, and then the price is raised such that the second guest in line (the poor one) cannot pay. The poor guest should not enter the ride.

This test covers the behavior being fixed by #11290 .

Ok, I might have gone a little bit rogue here. I wrote this PR to start a discussion on whether this type of testing is valuable or not. (and to be honest, it was great for learning about the codebase)

The idea here is to create very small games that are programatically played to cover a specific scenario. I have less then a week of experience on the codebase so I have no idea how many bugs can actually be covered with this type of tests.
I feel they can sometimes be an improvement over the replay tests as they can be easier to debug when they break.
Another interesting aspect of this style of testing is to exercise the APIs and perhaps reveal interesting operations that could be converted to game actions.

src/openrct2/peep/Guest.cpp
brenoguim: Seems that it works better now. I'm still doing more tests with ferris wheel and non ferris wheels rides to be sure.

duncanspumpkin:
Yeah i can fix bpb. Yes these changes are what i was thinking as well.

tupaschoal:
I took the WIP label out, as it seemingly now just requires replay update.

brenoguim:
@tupaschoal can you add the WIP label back again? I'm working on some infrastructure (mainly print functions for sprites) to be able to see exactly what is the first difference in the replay. I wanto to verify that the change is actually expected.
So far I have been able to tell that only a few changes happened (by dumping the hexa code of all sprites before and after the change in the failing tick). Now I'm implementing dump functions to understand what was this change.

janisozaur:
@brenoguim have you seen https://github.com/OpenRCT2/OpenRCT2/blob/d5c4d78d493c0238742f196bbe2d04f3116e0c75/test/tests/S6ImportExportTests.cpp#L128 and other code in that file? It's used for comparing gamestate snapshots you can have with replays.

duncanspumpkin:
You don't need to do that. We already have that capability. It's called the game state snapshot. At present we take a snapshot at the start and end of the replay but you can just create one for every tick if you want to see every little change.

brenoguim:
Thanks for the tips @duncanspumpkin and @janisozaur . So I would use these capabilities to generate the two snapshots to be compared and use functions in S6ImportExportTests.cpp to compare the two snapshots. Is that correct?
I saw your comments a bit too late. Implemented a function that dumps in text form all Peep attributes and I'm using this to compare them:

image

janisozaur:
You can see https://github.com/OpenRCT2/OpenRCT2/pull/11228 where replays were extended with snapshots. You could use the same to collect the data and have OpenRCT2 compare it for you, the only requirement could be adding more snapshots to the stream, in Duncan's tests they cost 400kiB in compressed output per snapshot, likely highly dependent on what the park is

brenoguim:
Ok, more updates to this issue. I have pushed two new commits:

1 - This commit makes the flow more similar to what it was before. Instead of just adding the car to the list of possible cars, I now clear the list and make that car the only choice.
I suspect this *might* be a bug fix in my patch because if the cars are supposed to be filled in pairs, we want to make sure the next guest always fills that gap right?

2 - The second commit is just a proof that now the only reason for the replay test to break is because of the extra `scenario_rand()` calls that happen because now the correct flow is applied.
*this patch is not supposed to go through!*. I'll remove the second patch as soon as the CI passes showing the point and you acknowledge it :)

What a ride!

janisozaur:
@brenoguim please keep subject line of the commit messages a touch shorter. There's a good reason GitHub ellipsises them after 70 cols.

See https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

brenoguim:
Thanks @janisozaur ! I will update these messages and pay attention to that in the future.

tupaschoal:
I think this just needs the Replay to be done @duncanspumpkin (and then undoing the last commit)

duncanspumpkin:
I've not tested it to destruction but have had a look at the related functionality and believe it will work correctly. @brenoguim how confident are you in it? If you are happy remove that last commit and I'll tak on the fixed Replay.

brenoguim:
@duncanspumpkin I'm confident of the change given that the last change made the replay work. I'll remove the last change.

test/tests/tests.vcxproj
```diff
@@ -77,4 +78,4 @@
     <ClCompile Include="TileElements.cpp" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-</Project>
\ No newline at end of file
+</Project>
```
duncanspumpkin, line 81: Can you revert this new line change.

duncanspumpkin:
That should be your replay issue fixed.

test/tests/tests.vcxproj
brenoguim: I'm trying hard... I've tried visual studio and vim. Both insert this change automatically. :(
I'll keep trying.

brenoguim: `:set nofixendofline` did the trick

tupaschoal:
I think you forgot to increment the network version

brenoguim:
> I think you forgot to increment the network version

I didn't. It just happened that someone incremented it in other PR and my change git "smartly" removed my change.

tupaschoal:
Thanks!

PR #11291, opened 2020-04-14
Fix coloring of Banners on tooltip

brenoguim:
This patch moves the function that adds coloring to the Buffer into the Buffer struct with a flag.
It also changes both Banner printing and Banner tooltip to use the colored version

I added a mutable buffer which is used for caching the string version of the formatting codepoint. Not sure if that is ideal.

This is a fix for #11286

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -20,6 +20,7 @@
 - Fix: [#11246] Fix Various Import/Export issues with Boat locations, balloon frame number.
 - Fix: Small red gardens in RCT1 saves are imported in the wrong colour.
 - Fix: [#11258] Properly remove format codes from imported strings.
+- Fix: [#11286] Fix banner tooltip colour
 - Improved: [#11157] Slimmer virtual floor lines.
 
 0.2.5 (2020-03-24)
--- a/src/openrct2/world/Banner.cpp
+++ b/src/openrct2/world/Banner.cpp
@@ -39,6 +39,29 @@ std::string Banner::GetText() const
     return format_string(STR_STRINGID, args);
 }
 
+size_t Banner::FormatTextTo(void* argsV, bool addColour) const
+{
+    auto args = static_cast<uint8_t*>(argsV);
+
+    int numColourArgs = 0;
+    if (addColour)
+    {
+        textColourUtf8.resize(5); // one code point in utf8 takes at most 4 bytes
+        auto terminator = utf8_write_codepoint(textColourUtf8.data(), FORMAT_COLOUR_CODE_START + text_colour);
+        *terminator = '\0';
+
+        set_format_arg_on(args, numColourArgs, rct_string_id, STR_STRING_STRINGID);
+        numColourArgs += sizeof(rct_string_id);
+
+        set_format_arg_on(args, numColourArgs, const char*, textColourUtf8.data());
+        numColourArgs += sizeof(const char*);
+
+        args += numColourArgs;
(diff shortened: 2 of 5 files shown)


duncanspumpkin:
Hmm it's not ideal to carry around that extra buffer just to fix this little issue. @intelOrca what are your thoughts on this

IntelOrca:
I don't like the fact that it is a fixed size of 32, we are trying to move away from fixed buffer sizes. I would at least use another std::string to store the coloured version.

Another approch is to have a string ID for each colour. For example replacing the following strings with: `{COLOUR}{STRINGID}` and then you can emit the correct string ID followed by the string ID for `{STRING}` followed then by the address of the banner string.

https://github.com/OpenRCT2/OpenRCT2/blob/develop/data/language/en-GB.txt#L2127

Of course the whole situation is already quite messy.

brenoguim:
Thanks for the feedback. Let me work with those suggestions and see what I can come up with.

brenoguim:
In this new version I removed the buffer by asking the compiler to generate the string table for me at compile time. Not sure if I went too far there :) what do you think?

If you like it, this CodePoint2Utf8 class could eventually be moved to a different place to be reused. I suppose other places could use that as well.

src/openrct2/world/Banner.cpp
```diff
@@ -32,13 +32,56 @@
...
 std::string Banner::GetText() const
 {
     uint8_t args[32]{};
     FormatTextTo(args);
     return format_string(STR_STRINGID, args);
 }
 
+size_t Banner::FormatTextTo(void* argsV, bool add_colour) const
+{
+    auto args = (uint8_t*)argsV;
```
tupaschoal, line 68: Use a named cast please :D

distribution/changelog.txt
```diff
@@ -19,6 +19,7 @@
 - Fix: [#11225] Replay manager cannot handle track designs.
 - Fix: [#11246] Fix Various Import/Export issues with Boat locations, balloon frame number.
 - Fix: Small red gardens in RCT1 saves are imported in the wrong colour.
+- Fix: [#11291] Fix banner tooltip colour
```
tupaschoal, line 22:
```suggestion
- Fix: [#11286] Fix banner tooltip colour
```
Should be the number of the issue that it fixes, not the PR number.

AaronVanGeffen:
I’m not against it, but I’d prefer refactoring to be done in a follow-up PR. We could merge the bugfix for the upcoming 0.2.6 release, but won’t be merging any further refactoring until street the release, in principle.

brenoguim:
@AaronVanGeffen , you mean that I should make this PR duplicate the code for adding the colour to get it merged to the current release, and for the following release I would clean that up with another PR?

AaronVanGeffen:
No, I would much prefer the short, hackier version for the initial bugfix PR. Just with the std::string instead of the char[] buffer.

brenoguim:
Oh ok. Shall be done :)

src/openrct2/world/Banner.h
```diff
@@ -42,6 +43,7 @@ struct Banner
     }
 
     std::string GetText() const;
+    size_t FormatTextTo(void* args, bool add_colour) const;
```
ZehMatt, line 46: We use camelCase for variables

brenoguim: Oh great! That's what I'm actually used to :)

PR #11296, opened 2020-04-14
Fix #11101: Server is not online

### LINKED ISSUE #11101: Server is not online under 0.2.5
**OS:** [Windows 10]
**Version:** (0.2.5)

Hello,

when i host a server under 0.2.5 develop and release my server is not online and does not appear in https://servers.openrct2.io/

Nobody is able to connect to me. 

When i host a server under 0.2.4 my server is online and does appear in 
https://servers.openrct2.io/

I downloaded the versions here :
https://openrct2.org/downloads/releases/latest
https://openrct2.org/downloads/releases/v0.2.4

I used the launcher too.

- [ X ] Specific to multiplayer?

IntelOrca:
Implement forceIPv4 in Http.WinHttp.

As I am not able to reproduce the problem, I can not guarantee this fixes the issue.

### DIFF
--- a/src/openrct2/core/Http.WinHttp.cpp
+++ b/src/openrct2/core/Http.WinHttp.cpp
@@ -19,6 +19,10 @@
 #    define WIN32_LEAN_AND_MEAN
 #    include <windows.h>
 #    include <winhttp.h>
+#    include <winsock2.h>
+#    include <ws2tcpip.h>
+
+bool InitialiseWSA();
 
 namespace Http
 {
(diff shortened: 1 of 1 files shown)


tupaschoal:
Actually I can see the server listed with and without the changes. I'm testing without password, as with would not work.
Without changes, on b5336fc6b, this is the output on my terminal. It fails to advertise, but connects:
develop
With the PR checked out and compiled:
with_pr

PR #11304, opened 2020-04-16
Move flags to RideTypeDescriptors

Gymnasiast:


src/openrct2/ride/RideData.h
```diff
@@ -104,33 +103,76 @@ struct RideTypeDescriptor
...
+    RIDE_TYPE_FLAG_SELLS_DRINKS = 1ull << 24ull,
+    RIDE_TYPE_FLAG_IS_TOILET = 1ull << 25ull,
+    RIDE_TYPE_FLAG_HAS_VEHICLE_COLOURS = 1ull << 26ull, // whether or not vehicle colours can be set
+    RIDE_TYPE_FLAG_CHECK_FOR_STALLING = 1ull << 27ull,
+    RIDE_TYPE_FLAG_HAS_TRACK = 1ull << 28ull,
+    RIDE_TYPE_FLAG_ALLOW_EXTRA_TOWER_BASES = 1ull << 29ull, // Only set by lift
+    RIDE_TYPE_FLAG_HAS_LARGE_CURVES = 1ull << 30ull,        // whether the ride supports large (45 degree turn) curves
+    RIDE_TYPE_FLAG_SUPPORTS_MULTIPLE_TRACK_COLOUR = 1ull << 31ull,
+
+    RIDE_TYPE_FLAG_ALLOW_DOORS_ON_TRACK = 1ull << 32ull,
```
AaronVanGeffen, line 158: I'd like to suggest using `ULL` instead of `ull`. In the latter, the 1 and l look awfully similar in many monotype fonts, unfortunately. (This was discussed before, but I can't find it for reference right now.)

src/openrct2/ride/coaster/meta/BobsleighCoaster.h
```diff
@@ -18,5 +18,13 @@ constexpr const RideTypeDescriptor BobsleighCoasterRTD =
     SET_FIELD(EnabledTrackPieces, (1ULL << TRACK_STRAIGHT) | (1ULL << TRACK_STATION_END) | (1ULL << TRACK_LIFT_HILL) | (1ULL << TRACK_FLAT_ROLL_BANKING) | (1ULL << TRACK_SLOPE) | (1ULL << TRACK_S_BEND) | (1ULL << TRACK_CURVE_SMALL) | (1ULL << TRACK_CURVE) | (1ULL << TRACK_HELIX_SMALL) | (1ULL << TRACK_BRAKES) | (1ULL << TRACK_BLOCK_BRAKES) | (1ULL << TRACK_ON_RIDE_PHOTO)),
     SET_FIELD(ExtraTrackPieces, 0),
     SET_FIELD(TrackPaintFunction, get_track_paint_function_bobsleigh_rc),
+    SET_FIELD(Flags, RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_MAIN | RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_ADDITIONAL |
```
duncanspumpkin, line 21: Is there any sort of standard set of flags a large amount of rides use. Perhaps we could make this a bit easier to follow if they were grouped.

IntelOrca: Good idea, you could also group together similar flags, e.g.:

```
RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_MAIN |
RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_ADDITIONAL | RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_SUPPORTS`

to

`RIDE_TYPE_FLAG_HAS_ALL_TRACK_COLOURS `

Comment (Gymnasiast) [reply src/openrct2/ride/coaster/meta/BobsleighCoaster.h]: I have created some groups (see third commit).

File: src/openrct2/ride/Ride.cpp [hunk 3]
```diff
@@ -5656,9 +5656,10 @@ rct_ride_name get_ride_naming(const uint8_t rideType, rct_ride_entry* rideEntry)
     }
 }
 
-bool ride_type_has_flag(int32_t rideType, uint32_t flag)
+bool ride_type_has_flag(int32_t rideType, uint64_t flag)
 {
-    return (RideProperties[rideType].flags & flag) != 0;
+    Guard::Assert(rideType < static_cast<int32_t>(std::size(RideTypeDescriptors)));
+    return (RideTypeDescriptors[rideType].Flags & flag) != 0;
```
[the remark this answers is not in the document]
Gymnasiast: Guard::Assert will always crash the game, so returning something is pointless. You're thinking of `assert()`, for which your suggestion would be more appropriate.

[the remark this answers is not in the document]
tupaschoal: Guard assert crashes even in release? I didn't know that. I thought it was just an enhanced `assert`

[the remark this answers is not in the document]
Gymnasiast: Hm, on closer inspection of the code, it might not actually crash the game, or not always. So your suggestion might be of use. By the way, you can just `Guard::Assert(false)`, since you moved the check to the if/else.

Issue #10483, opened 2019-12-30
Keep path objects when changing pathtype


WackedSwe:
When replacing paths it removes benches and lamps. Would there be a possible way to incorporate a texture switch instead of replacing the whole object when using paths?

frutiemax:
If anyone wants to do a commit; commenting line #193 in FootpathPlaceAction.hpp make this work, the benches don't get removed.

```
GameActionResult::Ptr ElementUpdateExecute(PathElement * pathElement, GameActionResult::Ptr res) const
    {
        const int32_t newFootpathType = (_type & (FOOTPATH_PROPERTIES_TYPE_MASK >> 4));
        const bool newPathIsQueue = ((_type >> 7) == 1);
        if (pathElement->GetSurfaceEntryIndex() != newFootpathType || pathElement->IsQueue() != newPathIsQueue)
        {
            res->Cost += MONEY(6, 00);
        }

        footpath_queue_chain_reset();

        if (!(GetFlags() & GAME_COMMAND_FLAG_PATH_SCENERY))
        {
            footpath_remove_edges_at(_loc, reinterpret_cast<TileElement*>(pathElement));
        }

        pathElement->SetSurfaceEntryIndex(_type & ~FOOTPATH_ELEMENT_INSERT_QUEUE);
        if (_type & FOOTPATH_ELEMENT_INSERT_QUEUE)
        {
            pathElement->SetIsQueue(true);
        }
        else
        {
            pathElement->SetIsQueue(false);
        }

        //Issue #10483 fix
        //pathElement->SetAddition(0);
        pathElement->SetIsBroken(false);

        RemoveIntersectingWalls(pathElement);
        return res;
    }
```

duncanspumpkin:
I think this might be done for (possibly questionable) realism.

ocalhoun6:
@frutiemax: Still have to make sure it removes path items when switching between queue and normal paths, though, right?

You don't want a flow of: build trash can on regular path --> replace regular path with queue path --> now have trash can on queue path. (Or vice versa withe queue line TV's.)

I'm not quite conversant enough with the way this code works to be sure, but it looks to me like this could happen if all you do is comment out that line. I think it would need an additional conditional in there, to trigger "pathElement->SetAddition(0)" if the queue/not queue state of the new/old paths don't match.

frutiemax:
Oh right forgot about the queue line. I'll check this out.

PR #11306, opened 2020-04-16
Changing footpath type won't remove the elements on it.

frutiemax:
Changing footpath type for an existing footpath does not remove elements on it, such as benches, lamps, etc.

https://github.com/OpenRCT2/OpenRCT2/issues/10483


src/openrct2/actions/FootpathPlaceAction.hpp
```diff
@@ -185,12 +185,19 @@ DEFINE_GAME_ACTION(FootpathPlaceAction, GAME_COMMAND_PLACE_PATH, GameActionResul
         if (_type & FOOTPATH_ELEMENT_INSERT_QUEUE)
         {
             pathElement->SetIsQueue(true);
+
+            //remove all but TVs
+            if (pathElement->GetAddition() != 0x6 && pathElement->GetAddition() != 0x5)
```
AaronVanGeffen, line 190: Please use constants rather than hardcoded numbers.

frutiemax: I am not sure how to define those constants to be honest. Sometimes the TV addition put a value of 0x5 sometimes 0x6... Should I declare constants ADDITION_TV_1 = 0x5 and ADDITION_TV_2 = 0x6?

duncanspumpkin: @frutiemax the reason they change is because they are not constant. These are index's into the objects. What you should do is use GetAdditionEntryIndex and use that to index the object. Then the object will tell you what type it is.

frutiemax:
I used the function to get the object entry, and I check with the path bits to see if there is a TV.

src/openrct2/actions/FootpathPlaceAction.hpp
```diff
@@ -182,15 +183,26 @@ DEFINE_GAME_ACTION(FootpathPlaceAction, GAME_COMMAND_PLACE_PATH, GameActionResul
...
+            else
+            {
+                //remove all TVs
+                if ((elem->path_bit.flags & PATH_BIT_FLAG_IS_QUEUE_SCREEN) != 0)
+                    pathElement->SetAddition(0);
+            }
         }
-        pathElement->SetAddition(0);
+        
         pathElement->SetIsBroken(false);
```
ocalhoun6, line 206: Maybe this line should be removed as well? (Or maybe moved inside the conditional between lines 196 and 197, to ensure that the broken flag gets set false when non-TV path items are removed. Since TV's can't be broken -- and therefore shouldn't have that flag in the first place -- we don't need to worry about that part.)

If changing the path type isn't getting rid of the path furniture, maybe it shouldn't magically repair the path furniture, either?

As the code stands now, it looks like if you changed the path type underneath broken path furniture, it would repair the path furniture for free. Not sure if that's the desired behavior or not, since that would normally cost money to fix.

frutiemax: Good catch! I think the repair flag should be set to false when resetting the Addition to 0.

frutiemax:
I am not sure how to fix the formatting and add change to changelog...

tupaschoal:
@frutiemax you need to go to `distribution/changelog.txt` and add an entry there, under the current revision. See how others have added in the past to get a grasp, it's sorted by the issue number and type of PR

src/openrct2/actions/FootpathPlaceAction.hpp
```diff
@@ -182,16 +183,31 @@ DEFINE_GAME_ACTION(FootpathPlaceAction, GAME_COMMAND_PLACE_PATH, GameActionResul
...
+        bool isQueue = _type & FOOTPATH_ELEMENT_INSERT_QUEUE;
+        pathElement->SetIsQueue(isQueue);
+
+        rct_scenery_entry* elem = pathElement->GetAdditionEntry();
+        if (elem != nullptr)
         {
-            pathElement->SetIsQueue(false);
+            if (isQueue)
+            {
+                // remove any addition that isn't a TV
```
Gymnasiast, line 194: Queues can also handle lamps.

src/openrct2/actions/FootpathPlaceAction.hpp
```diff
@@ -182,16 +183,31 @@ DEFINE_GAME_ACTION(FootpathPlaceAction, GAME_COMMAND_PLACE_PATH, GameActionResul
...
+        pathElement->SetIsQueue(isQueue);
+
+        rct_scenery_entry* elem = pathElement->GetAdditionEntry();
+        if (elem != nullptr)
         {
-            pathElement->SetIsQueue(false);
+            if (isQueue)
+            {
+                // remove any addition that isn't a TV or a lamp
+                if ((elem->path_bit.flags & PATH_BIT_FLAG_IS_QUEUE_SCREEN) == 0 && (elem->path_bit.flags & PATH_BIT_FLAG_LAMP) == 0)
```
tupaschoal, line 195: if ((elem->path_bit.flags & PATH_BIT_FLAG_IS_QUEUE_SCREEN) == 0
+                    && (elem->path_bit.flags & PATH_BIT_FLAG_LAMP) == 0)
```suggestion
                if ((elem->path_bit.flags & PATH_BIT_FLAG_IS_QUEUE_SCREEN) == 0
                    && (elem->path_bit.flags & PATH_BIT_FLAG_LAMP) == 0)
```
Just to fix formatting for everything to be ok, CI-wise

distribution/changelog.txt
```diff
@@ -1,5 +1,6 @@
 0.2.5+ (in development)
 ------------------------------------------------------------------------
+- Feature: [#11306] Replacing path textures do not remove additions on it. It handles the switch from queue line to standard path and vice versa.
```
Gymnasiast, line 3: This needs to be moved to the 0.2.6+ section. But in order to do so, you must first rebase this pull request.

janisozaur:
That didn't quite work. Do you need want help from us with the rebase?

frutiemax:
On my side it seemed to have worked... Is it master or dev that should be the base?


TurtoiseGit_rebase

tupaschoal:
`develop`, master is never pushed to, save when we're releasing

janisozaur:
@frutiemax I rebased it for you, also moved changelog entry to new section

duncanspumpkin:
This still needs a network version bump.

tupaschoal:
I've bumped the network version, will merge if CI is green

Gymnasiast:
I would have preferred if you had waited with merging until I had actually given my review.

distribution/changelog.txt
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#9029] Open doors with the tile inspector.
 - Feature: [#11231] Change shortcut window list order to be more intuitive, and split it into logical sections.
+- Feature: [#11306] Replacing path textures do not remove additions on it. It handles the switch from queue line to standard path and vice versa.
```
Gymnasiast, line 5: This description is too long. Since it was merged before I got the change to approve it, I will modify it in `develop`.

tupaschoal:
My bad then

distribution/changelog.txt
janisozaur: I had pretty much the same gripe, but assumed it would be amended before getting merged

PR #11307, opened 2020-04-16
Add #11218, #11281 to changelog

janisozaur:
[ci skip]


IntelOrca:
I don't think `[ci skip]` is a thing anymore.

Issue #5451, opened 2017-05-07
Guest Coaster Screaming Bug


HolyMythos:
**OS:** Windows 10
**Version:** 0.0.8 (also in 0.0.7)
**Commit/Build:** any 0.0.8

When guests are on any coaster that has a drop of any kind they will scream every time as opposed to vanilla where there is a height limit on screaming. On vanilla, guests would not scream on small drops on a coaster (5ft-15ft) however on larger drops (20ft+) the scream would activate. Due to this rides with multiple small carts and small drops (like Wooden Wild Mouse) cause nonstop screams that cut through all other sounds in the game. Keeping these coasters on screen causes the sound to be overloaded with high pitch screaming. Fixing the screams to only activate on a drop of 20ft or more will fix this.

**Steps to reproduce:**
1.  Start the Crazy Castle scenario (or load save below).
2.  Build the Wooden Wild Mouse roller coaster called Mouse Towers and set to free to fill multiple carts.
3.  When the ride is full and running all carts you should hear the nonstop screaming.

**Save game:**
[WoodenTesttext.zip](https://github.com/OpenRCT2/OpenRCT2/files/981875/WoodenTesttext.zip)

rwjuk:
Screams appear to be generated based on speed, not drop height (if the vehicle's velocity goes above 44 (internal units), then a scream is generated).

HolyMythos:
For the ride in question, the speed of the mice is slow (if you're talking
about mph) is way below 44 so no scream should generate but they do, unless
I'm understanding this incorrectly.

Cruplezone:
>  the speed of the mice is slow (if you're talking about mph) is way below 44

AFAIK 44 internal units != 44 mph.

HolyMythos:
IIRC the scream however is not in the original game for this particular ride on the small 5ft drops so I don't see why it would be now, unless there is a percentage on how often the scream is generated, but even if that's the case it always generates on OpenRCT2 100% of the time.

carlu93:
I can confirm the issue. I recognized it on Splash Mine, where they start screaming everytime, which started to annoy me. This definitely didn't happen that much on vanilla.

Broxzier:
In this video from Plokite_Wolf it's very noticeable that one of the screams is used way more frequently than the others: https://www.youtube.com/watch?v=1KG9JpNDEnE

[...]

PR #11308, opened 2020-04-16
reduce amount of screaming on rides

alexcai10:
My first time using GitHub, please let me know if I am on the right track/changes you would like to see.  Thanks


duncanspumpkin:
I think there is still some debate over if the scream amount is different to vanilla. If it is different to vanilla we should look into if that is because of a bug or not.

tupaschoal:
@alexcai10 I'll close this for now, we need to settle on a solution on #5451 first. Feel free to investigate it or to pick another issue to work on and resume this PR after it has been solved.

PR #11278, opened 2020-04-12
Fixed grammatical error described in #11275

mkstowe:

PR #11309, opened 2020-04-16
Adding singular sentences for rider deaths in different languages

alexcai10:


AaronVanGeffen:
Thanks for what I assume is a well-intended PR, but these are invalid string ids. Moreover, this is not the localisation repository, where translations are maintained see the [contributing guidelines](https://github.com/OpenRCT2/OpenRCT2/blob/develop/CONTRIBUTING.md#translation)). 

Finally, please only contribute to translations you are a native speaker of. I do not believe you're a native speaker of all 26 of these...

tupaschoal:
FYI, this message will appear when https://github.com/OpenRCT2/OpenRCT2/pull/11278 is merged, but like @AaronVanGeffen said, the translations for those will be made over at https://github.com/OpenRCT2/Localisation

PR #11310, opened 2020-04-17
Fix String Ids for localisation skipped on #11273

tupaschoal:
It skipped 100 ids


AaronVanGeffen:
Thank you for spotting this. I’m terribly embarrassed.

tupaschoal:
Don't be!

PR #11312, opened 2020-04-17
Make issue template support new syntax

tupaschoal:
This updates our issue template to the latest github syntax, which will make this go away:

image

It also enables us to do different templates depending on whether it's a feature/bug/discussion, so when user clicks "new issue", it will be like this:
image


duncanspumpkin:
Should we also take the opportunity to make a feature request version of the template.

AaronVanGeffen:
According to [this page](https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates) the new format uses a directory to store templates and a config file. Therefore, shouldn't the file also be renamed to e.g. `.github/ISSUE_TEMPLATE/bug_report.md`?

tupaschoal:
> Should we also take the opportunity to make a feature request version of the template.

I thought about it and would do it up next, but if you agree I can do it here too.

duncanspumpkin:
As discussed on discord. We will do follow on prs for feature template and possibly generic task template.

PR #11316, opened 2020-04-17
Fix 11050: macOS release 0.2.5 has invalid permissions

IntelOrca:
Okay it still doesn't give you a unix zip as an artefact, but at least you can just take it out the dos zip wrapper.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,10 +40,10 @@ jobs:
           build-symbols
           build-installer -i
       - name: Upload artifacts (CI)
-        uses: actions/upload-artifact@v1
+        uses: actions/upload-artifact@v2-preview
         with:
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This zip in a zip will most likely break the launcher, surely?

Gymnasiast:
Good point. Though to be precise, it will _continue_ to be broken, unless we manually fix the artifact before releasing.

AaronVanGeffen:
Right, we can manually intervene for the release... I don't agree with saying it will continue to be broken, though, as macOS 10.14 was unaffected by the bug. This is a regression for them.

Gymnasiast:
Well, I remember the launcher not being able to extract the ZIP *at all*. Not sure if that was ever fixed for the 0.2.5 release.

PR #11317, opened 2020-04-17
Fix #11315: Never opened ride shown as favourite ride of 74 guests

### LINKED ISSUE #11315: Ride that has never opened is shown as favorite ride of many guests
**OS:** [Windows 10]
**Version:** [0.2.5]
**Commit/Build:** [d342423]

From what I could see in a quick scan of the code, the guests favorite ride gets set when they exit a ride. This ride has never opened but is shown as the favorite ride of 74 guests.

image

[_GuestFavoriteCount.zip](https://github.com/OpenRCT2/OpenRCT2/files/4492811/_GuestFavoriteCount.zip)

Gymnasiast:


tupaschoal:
Marking as WIP till we fix replays

duncanspumpkin:
I'll fix the replays tomorrow for this.

PR #11326, opened 2020-04-17
11311 : Peep spawn locations can be placed elsewhere than on the edge of the map

frutiemax:
- In the scenario editor, it is now possible to place peep spawn locations elsewhere than on the edge of the map.
- The recursive routine that tracks the path from the entrance to the edge of the map has been adapted to check for peep spawn locations
- In game, if the peep is on the edge of the map, it exhibits the same behaviour as before, i.e. it quits the world on the edge of the tile.
- In game, if the peep is elsewhere, the moment it enters the tile with a spawn point with the same direction it quits the world.

### DIFF
--- a/src/openrct2/actions/PlacePeepSpawnAction.hpp
+++ b/src/openrct2/actions/PlacePeepSpawnAction.hpp
@@ -62,12 +62,12 @@ public:
             return std::make_unique<GameActionResult>(GA_ERROR::NO_FREE_ELEMENTS, STR_ERR_CANT_PLACE_PEEP_SPAWN_HERE, STR_NONE);
         }
 
-        if (_location.x <= 16 || _location.y <= 16 || _location.x >= (gMapSizeUnits - 16)
+        /*if (_location.x <= 16 || _location.y <= 16 || _location.x >= (gMapSizeUnits - 16)
             || _location.y >= (gMapSizeUnits - 16))
         {
             return std::make_unique<GameActionResult>(
                 GA_ERROR::INVALID_PARAMETERS, STR_ERR_CANT_PLACE_PEEP_SPAWN_HERE, STR_OFF_EDGE_OF_MAP);
-        }
+        }*/
 
         // Verify footpath exists at location, and retrieve coordinates
         auto pathElement = map_get_path_element_at(TileCoordsXYZ{ _location });
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -3076,44 +3076,52 @@ void Peep::PerformNextAction(uint8_t& pathing_result, TileElement*& tile_result)
     }
 
     //check if peep is on peep spawn location
-    for (auto& elem : gPeepSpawns)
+    bool outside = false;
+    if (newLoc.x <= 32 || newLoc.y <= 32 || newLoc.x >= gMapSizeUnits || newLoc.y >= gMapSizeUnits)
+        // apply same old behaviour
+        outside = true;
+
+    else
     {
-        CoordsXYZ coord1, coord2;
-        coord1.x = elem.ToTileCentre().x;
-        coord1.y = elem.ToTileCentre().y;
-        coord1.z = elem.ToTileCentre().z;
-
-        CoordsXYZ tileCoord;
-        tileCoord.x = coord1.x / COORDS_XY_STEP;
-        tileCoord.y = coord1.y / COORDS_XY_STEP;
-        tileCoord.z = coord1.z;
-
-        //check if either peep is on the edge of the map or anywhere else
-        bool outside = false;
-        
-        if (tileCoord.x == 0 || tileCoord.x == (gMapSizeUnits - 1) / COORDS_XY_STEP || tileCoord.y == 0
-            || tileCoord.y == (gMapSizeUnits - 1) / COORDS_XY_STEP)
+        for (auto& elem : gPeepSpawns)
         {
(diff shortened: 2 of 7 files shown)


src/openrct2/peep/Peep.cpp
```diff
@@ -3068,22 +3068,60 @@ void Peep::PerformNextAction(uint8_t& pathing_result, TileElement*& tile_result)
...
+            // check if either peep is on the edge of the map or anywhere else
+            if (tileCoord.x == 0 || tileCoord.x == (gMapSizeUnits / COORDS_XY_STEP)-1 || tileCoord.y == 0
+                || tileCoord.y == (gMapSizeUnits / COORDS_XY_STEP)-1)
+            {
+                //do nothing
+            }
+
+            else if (
+                newLoc.ToTileCentre().x == coord1.x && newLoc.ToTileCentre().y == coord1.y && height == coord1.z
+                && direction == elem.direction)
```
tupaschoal, line 3107: (not sure if it will be the final formatting, just simplifying the comparison)
```suggestion
            else if (newLoc.ToTileCentre() == coord1 && height == coord1.z && direction == elem.direction)
```

src/openrct2/peep/Peep.cpp
```diff
@@ -3068,22 +3068,60 @@ void Peep::PerformNextAction(uint8_t& pathing_result, TileElement*& tile_result)
...
+            coord1.z = elem.ToTileCentre().z;
+
+            CoordsXYZ tileCoord;
+            tileCoord.x = coord1.x / COORDS_XY_STEP;
+            tileCoord.y = coord1.y / COORDS_XY_STEP;
+            tileCoord.z = coord1.z;
+
+            // check if either peep is on the edge of the map or anywhere else
+            if (tileCoord.x == 0 || tileCoord.x == (gMapSizeUnits / COORDS_XY_STEP)-1 || tileCoord.y == 0
+                || tileCoord.y == (gMapSizeUnits / COORDS_XY_STEP)-1)
```
tupaschoal, line 3100: Couldn't you just do this?
```suggestion
            if (map_is_edge(coord1))
```
And is it ok to do nothing? Won't peep start wandering off? Replay is broken so I'm guessing the reason might be here.

duncanspumpkin:
This is expected to fail replays. The previous code deleted a peep as soon as they were about to leave the map bounds. This deletes the peep when they enter the peep spawn which is a tile earlier. I would prefer if it was kept the same as before but i don't quite know how that would be achieved.

frutiemax:
duncan, currently the code checks if the spawn point is on the edge of the map, if it is, it behaves the same as before i.e. it checks the peep's location is outside of the map.

The problem comes when the spawn point is not on the edge of the map. Paths seem to autoconnect to the edge of the map so peeps can get at the edge of the spawn point. However, when a spawn point is put in the middle of the map, there is no guarantee the path is connected to the edge. This is why I took the decision to delete the peep when it gets to the center of the tile.

frutiemax:
Applied some of the changes suggested.

tupaschoal:
Please mark solved issues as resolved and either fix the pending ones or comment in case you have a doubt or concern, so we can do other reviews :)

janisozaur:
Please mind your commit messages: https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

They are supposed to be concise, informative and glanceable.

src/openrct2/peep/Peep.cpp
```diff
@@ -3068,22 +3068,55 @@ void Peep::PerformNextAction(uint8_t& pathing_result, TileElement*& tile_result)
...
         MoveTo(newLoc.x, newLoc.y, height);
         return;
     }
 
-    if (newLoc.x < 32 || newLoc.y < 32 || newLoc.x >= gMapSizeUnits || newLoc.y >= gMapSizeUnits)
+
+    bool outside = false;
+    if (newLoc.x <= 32 || newLoc.y <= 32 || newLoc.x >= gMapSizeUnits || newLoc.y >= gMapSizeUnits)
+        // apply same old behaviour
+        outside = true;
```
duncanspumpkin, line 3082: Please don't use if's like this. Its a classic way of falling into the goto fail trap. https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

tupaschoal:
@frutiemax Please mark issues as resolved once you deal with them, like I said in the other PR.

frutiemax:
I'd like to mark that this requires additional testing before merging as this feature can break the game if buggy.

src/openrct2/actions/PlacePeepSpawnAction.hpp
```diff
@@ -62,12 +62,12 @@ DEFINE_GAME_ACTION(PlacePeepSpawnAction, GAME_COMMAND_PLACE_PEEP_SPAWN, GameActi
...
         }
 
-        if (_location.x <= 16 || _location.y <= 16 || _location.x >= (gMapSizeUnits - 16)
+        /*if (_location.x <= 16 || _location.y <= 16 || _location.x >= (gMapSizeUnits - 16)
             || _location.y >= (gMapSizeUnits - 16))
         {
             return std::make_unique<GameActionResult>(
                 GA_ERROR::INVALID_PARAMETERS, STR_ERR_CANT_PLACE_PEEP_SPAWN_HERE, STR_OFF_EDGE_OF_MAP);
-        }
+        }*/
```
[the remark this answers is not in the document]
Gymnasiast: This was incorrectly marked as "resolved".

PR #11330, opened 2020-04-18
Add a manually played test to verify that a guest cannot enter a ride it cannot pay

brenoguim:
This test verifies that a guest, when second in queue, won't be forced to enter the ride if it has not enough money to pay for it.
To simulate this scenario, two guests (a rich and a poor) are encouraged to enter the ride queue, and then the price is raised such that the second guest in line (the poor one) cannot pay. The poor guest should not enter the ride.

This test covers the behavior being fixed by #11290 .

-------------------------------------------------------

Ok, I might have gone a little bit rogue here. I wrote this PR to start a discussion on whether this type of testing is valuable or not. (and to be honest, it was great for learning about the codebase)

The idea here is to create very small games that are programatically played to cover a specific scenario. I have less then a week of experience on the codebase so I have no idea how many bugs can actually be covered with this type of tests.
I feel they can sometimes be an improvement over the replay tests as they can be easier to debug when they break.
Another interesting aspect of this style of testing is to exercise the APIs and perhaps reveal interesting operations that could be converted to game actions.


duncanspumpkin:
@brenoguim you might be interested also in our replay tests that replay the game and checks the checksums at every tick. Its not quite the same as this as this is to check that a peep with no money can't skip the checks and spend money they don't have.

test/tests/PlayTests.cpp
```diff
@@ -0,0 +1,139 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2018 OpenRCT2 developers
```
tupaschoal, line 2: 2014-2020

test/tests/tests.vcxproj
```diff
@@ -66,6 +66,7 @@
     <ClCompile Include="Localisation.cpp" />
     <ClCompile Include="MultiLaunch.cpp" />
     <ClCompile Include="ReplayTests.cpp" />
+    <ClCompile Include="PlayTests.cpp" />
```
tupaschoal, line 69: If this is merged we probably want a more meaningful name so we avoid people from always adding to the same test? Or do we want the big file with tests?

test/tests/tests.vcxproj
```diff
@@ -77,4 +78,4 @@
     <ClCompile Include="TileElements.cpp" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-</Project>
\ No newline at end of file
+</Project>
```
tupaschoal: Unrelated change.

test/tests/PlayTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+TEST_F(PlayTests, SecondGuestInQueueShouldNotRideIfNoFunds)
+{
+    /* This test verifies that a guest, when second in queue, won't be forced to enter
+     * the ride if it has not enough money to pay for it.
+     * To simulate this scenario, two guests (a rich and a poor) are encouraged to enter
+     * the ride queue, and then the price is raised such that the second guest in line
+     * (the poor one) cannot pay. The poor guest should not enter the ride.
+     */
+    std::string initStateFile = TestData::GetParkPath("small_park_with_ferris_wheel.sv6");
```
tupaschoal, line 83: Didn't you forget to include this?

test/tests/PlayTests.cpp
```diff
@@ -0,0 +1,139 @@
...
+    // Open it for free
+    ride_set_status(&ferrisWheel, RIDE_STATUS_OPEN);
+    GameActions::Execute(std::make_unique<RideSetPriceAction>(ferrisWheel.id, 0, true).get());
+
+    // Ignore intesity to stimulate peeps to queue into ferris wheel
+    gCheatsIgnoreRideIntensity = true;
+
+    // Insert a rich guest
+    auto richGuest = gs->GetPark().GenerateGuest();
+    richGuest->cash_in_pocket = 3000;
```
tupaschoal, line 112: Really rich for `OpenRCT2` standards :D

brenoguim: That value actuallly means 300.0 in the game. that should be fine, no?

test/tests/tests.vcxproj
brenoguim: I was thinking that this can start as a single file with the tests names being the descritive part (just like ReplayTests.cpp is not descriptive to what the test is verifying). If this turns out to grow too much, then it coulde be rethinked.

test/tests/PlayTests.cpp
brenoguim: This is in the PR isn't it? It's a binary file.

test/tests/PlayTests.cpp
tupaschoal: Sure, I was just poking at you 🤣

test/tests/tests.vcxproj
brenoguim: I have no idea what is going on here. I think Visual Studio added a new line, but doesn't show me. I'll take a look

brenoguim:
@duncanspumpkin I looked into the ReplayTests too. Indeed I could not figure out how to do this kind of "semantic testing" with the replay infra. That's why I created the "Play" tests :D

brenoguim:
Moved this PR to #11290 so the test and the fix go together

PR #11332, opened 2020-04-18
Fix #11331: Create constant for failure on ScenarioCompletedCompanyValue

### LINKED ISSUE #11331: Create constant for ScenarioCompletedCompanyValue
>I prefer a constant for this.
If you don't want to do that in this PR, could you create an issue?

https://github.com/OpenRCT2/OpenRCT2/blob/1af55d5894dfc14765c4a115f93a9d5318a90002/src/openrct2-ui/windows/Park.cpp#L1616

Taken from #11136

tupaschoal:


src/openrct2/scenario/Scenario.h
```diff
@@ -366,6 +366,8 @@ enum
 #define AUTOSAVE_PAUSE 0
 #define DEFAULT_NUM_AUTOSAVES_TO_KEEP 10
 
+static constexpr money32 INVALID_VALUE_ON_SCENARIO_FAILURE = 0x80000001;
```
Gymnasiast, line 369:
```suggestion
static constexpr money32 COMPANY_VALUE_FAILED_OBJECTIVE = 0x80000001;
```

tupaschoal: Thanks for the suggestion, I actually went with `COMPANY_VALUE_ON_FAILED_OBJECTIVE`, let me know if you disagree.

PR #11334, opened 2020-04-18
Fix refactor mistake that prevents compilation

Gymnasiast:
This used to be a cast to `uintptr_t` and was changed to `const uintptr_t` in a PR merged today. This prevents the game from compiling on my machine. Changing it back to `uintptr_t`.


tupaschoal:
Weird, I wonder how none of the CIs failed on compilation

Issue #9568, opened 2019-07-14
Fixing ride chain sounds


Gymnasiast:
Ride | Current sound | Proposal
-- | -- | --
SLC | Friction wheels | Classic chain sound or Arrow sound
Reverser RC | Classic chain sound | Side-friction RC sound

iking96:
@Gymnasiast What does SLC stand for?

Gymnasiast:
It stands for Suspended Looping Coaster, indeed. In RCT2 and OpenRCT2, it is referred to as "Compact Inverted Coaster".

PR #11347, opened 2020-04-18
Move LiftData to RTDs

Gymnasiast:
It is advised to look commit by commit. Two of reduce direct ride type comparisons and replace it with a RIDE_TYPE flag - this should be helpful in the future, if/when we want to turn ride types into objects.

There is one deliberate change, and that is changing the reverser rc sound to the sound already used by the side-friction and wooden rc.


tupaschoal:
This is breaking `bpb` and since it was only moving things around, I don't think it should, right?

Gymnasiast:
For some reason, they work fine on my system. :/
I'm going to split up this PR.

Gymnasiast:
It seems that even changing the sound of the lift hill affects replays. I have reverted that change - there is an issue filed for that, and I will deal with it at a later date.

Needs squash on merge.

src/openrct2/ride/coaster/meta/ReverseFreefallCoaster.h
```diff
@@ -22,5 +22,6 @@ constexpr const RideTypeDescriptor ReverseFreefallCoasterRTD =
     SET_FIELD(Flags, RIDE_TYPE_FLAGS_TRACK_HAS_3_COLOURS | RIDE_TYPE_FLAGS_COMMON_COASTER | RIDE_TYPE_FLAGS_COMMON_COASTER_NON_ALT),
     SET_FIELD(NameConvention, { RIDE_COMPONENT_TYPE_CAR, RIDE_COMPONENT_TYPE_TRACK, RIDE_COMPONENT_TYPE_STATION }),
     SET_FIELD(AvailableBreakdowns, (1 << BREAKDOWN_SAFETY_CUT_OUT) | (1 << BREAKDOWN_RESTRAINTS_STUCK_CLOSED) | (1 << BREAKDOWN_RESTRAINTS_STUCK_OPEN) | (1 << BREAKDOWN_VEHICLE_MALFUNCTION)),
+    SET_FIELD(LiftData, { SoundId::Null, 5, 5 }),
```
tupaschoal, line 25: Note for future: also sounds weird for this coaster to have no sound

src/openrct2/ride/coaster/meta/LIMLaunchedRollerCoaster.h
```diff
@@ -24,5 +24,6 @@ constexpr const RideTypeDescriptor LIMLaunchedRollerCoasterRTD =
                      RIDE_TYPE_FLAG_PEEP_CHECK_GFORCES | RIDE_TYPE_FLAG_ALLOW_MULTIPLE_CIRCUITS),
     SET_FIELD(NameConvention, { RIDE_COMPONENT_TYPE_TRAIN, RIDE_COMPONENT_TYPE_TRACK, RIDE_COMPONENT_TYPE_STATION }),
     SET_FIELD(AvailableBreakdowns, (1 << BREAKDOWN_SAFETY_CUT_OUT) | (1 << BREAKDOWN_RESTRAINTS_STUCK_CLOSED) | (1 << BREAKDOWN_RESTRAINTS_STUCK_OPEN) | (1 << BREAKDOWN_VEHICLE_MALFUNCTION) | (1 << BREAKDOWN_BRAKES_FAILURE)),
+    SET_FIELD(LiftData, { SoundId::Null, 4, 6 }),
```
tupaschoal, line 27: This is the only one that is null, but not `5, 5`. Should it?

(Note for future: sounds weird for this coaster to have no sound)

Gymnasiast:
> (Note for future: sounds weird for this coaster to have no sound)

Both roller coasters you commented this on do not have lift hills, so there's no sound set for them.

The issue for the lift hill sounds is https://github.com/OpenRCT2/OpenRCT2/issues/9568.

PR #11349, opened 2020-04-18
Create a compile-time table for converting color codepoints to utf8

brenoguim:
For this to be possible, the function to convert a codepoint to urf8 was
moved to the header and made constexpr.

Now it's not necessary to store a std::string just to serve as a buffer
for holding this string. All the computation is done at compile time.

### DIFF
--- a/src/openrct2/world/Banner.cpp
+++ b/src/openrct2/world/Banner.cpp
@@ -35,13 +35,13 @@ static Banner _banners[MAX_BANNERS];
 
 namespace
 {
-    template<uint32_t From, uint32_t To> struct CodePointToUtf8
+    template<uint32_t TFrom, uint32_t TTo> struct CodePointToUtf8
     {
         constexpr CodePointToUtf8()
         {
-            for (uint32_t i = From; i <= To; ++i)
+            for (uint32_t i = TFrom; i <= TTo; ++i)
             {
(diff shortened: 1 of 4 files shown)


src/openrct2/world/Banner.cpp
```diff
@@ -33,6 +33,30 @@
...
+            }
+        }
+
+        constexpr auto operator()(uint8_t colourId) const
+        {
+            return m_colors[colourId];
+        }
+
+        using Utf8Colour = utf8[5]; // A 32bit codepoint uses at most 4 bytes in utf8
+        Utf8Colour m_colors[To - From + 1]{};
```
tupaschoal, line 54:
```suggestion
    template<uint32_t TFrom, uint32_t TTo> struct CodePointToUtf8
    {
        constexpr CodePointToUtf8()
        {
            for (uint32_t i = TFrom; i <= TTo; ++i)
            {
                utf8_write_codepoint(m_colors[i - TFrom], i);
            }
        }

        constexpr auto operator()(uint8_t colourId) const
        {
            return m_colors[colourId];
        }

        using Utf8Colour = utf8[5]; // A 32bit codepoint uses at most 4 bytes in utf8
        Utf8Colour m_colors[TTo - TFrom + 1]{};
```

PR #11350, opened 2020-04-19
Replacing set_format_arg_on macro in favor of a more C++zy solution

brenoguim:
This is a start to make the formatting infrastructure simpler. There are further changes that enables even more simplifications (like passing the Formatter to FormatTextTo functions).

Technically it should be possible to allow the template to deduce the type (so instead of writing `add<rct_string_id>(STR_NONE)` one would just write `add(STR_NONE)`) but I wanted to start with the more conservative version.

### DIFF
--- a/src/openrct2/localisation/Localisation.h
+++ b/src/openrct2/localisation/Localisation.h
@@ -98,13 +98,50 @@ template<typename T> void constexpr set_format_arg(uint8_t* args, size_t offset,
         set_format_arg_body(gCommonFormatArgs, offset, (uintptr_t)(value), sizeof(type));                                      \
     } while (false)
 
-#define set_format_arg_on(args, offset, type, value) set_format_arg_body(args, offset, (uintptr_t)(value), sizeof(type))
-
 #define set_map_tooltip_format_arg(offset, type, value)                                                                        \
     do                                                                                                                         \
     {                                                                                                                          \
         static_assert(sizeof(type) <= sizeof(uintptr_t), "Type too large");                                                    \
         set_format_arg_body(gMapTooltipFormatArgs, offset, (uintptr_t)(value), sizeof(type));                                  \
     } while (false)
+
+class Formatter
+{
+    const uint8_t* StartBuf;
+    uint8_t* CurrentBuf;
+
+public:
+    explicit Formatter(uint8_t* buf)
+        : StartBuf(buf)
+        , CurrentBuf(buf)
+    {
+    }
+
+    auto buf()
+    {
+        return CurrentBuf;
+    }
+
+    std::size_t bytes() const
+    {
+        return CurrentBuf - StartBuf;
+    }
+
+    template<typename TSpecified, typename TDeduced> Formatter& add(TDeduced value)
+    {
+        uintptr_t convertedValue;
+        if constexpr (std::is_integral_v<TSpecified>)
+        {
+            convertedValue = static_cast<uintptr_t>(value);
+        }
+        else
+        {
+            convertedValue = reinterpret_cast<uintptr_t>(value);
+        }
+        set_format_arg_body(CurrentBuf, 0, convertedValue, sizeof(TSpecified));
+        CurrentBuf += sizeof(TSpecified);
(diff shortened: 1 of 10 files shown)


tupaschoal:
Looks like a great simplification, but it still fails on formatting and some of the CIs, please take a look.

IntelOrca:
This is nice. Is it possible to also be also write something like:
```
FormatString<T0,T1,T2,T3...>(const char* buffer, const char* argsBuffer, rct_string_id stringId, T0 arg0, T1 arg1, T2 arg2, T3 arg3);
```

Using vararg generics?

brenoguim:
Cool idea! I think it might be possible to do it. I play with this idea for the next PR!

src/openrct2/localisation/Localisation.h
```diff
@@ -98,13 +98,41 @@ template<typename T> void constexpr set_format_arg(uint8_t* args, size_t offset,
...
+    {
+        return _end;
+    }
+
+    std::size_t bytes() const
+    {
+        return _end - _start;
+    }
+
+    template<typename T, typename V> Formatter& add(V value)
```
tupaschoal, line 130: Following the short discussion on #11354 , these should probably receive more meaningful names and be preceded with `T`

src/openrct2/localisation/Localisation.h
```diff
@@ -98,13 +98,41 @@ template<typename T> void constexpr set_format_arg(uint8_t* args, size_t offset,
...
+    }
+
+    std::size_t bytes() const
+    {
+        return CurrentBuf - StartBuf;
+    }
+
+    template<typename TSpecified, typename TDeduced> Formatter& add(TDeduced value)
+    {
+        set_format_arg_body(CurrentBuf, 0, (uintptr_t)(value), sizeof(TSpecified));
```
janisozaur, line 132: static cast

brenoguim: I can't static cast because sometimes it needs reinterpret cast:
```
D:\github\OpenRCT2\src\openrct2\localisation\Localisation.h(132,1): error C2440: 'static_cast': não é possível converter de 'TDeduced' para 'uintptr_t' [D:\github\OpenRCT2\src\openrct2\libopenrct2.vcxproj]
          with
          [
              TDeduced=char *
          ] (compilando arquivo de origem network\Network.cpp)
```
I can't reinterpret_cast because compiler says that it's ok to use static cast:

```
D:\github\OpenRCT2\src\openrct2\localisation\Localisation.h(132,1): error C2440: 'reinterpret_cast': não é possível converter de 'TDeduced' para 'uintptr_t' [D:\github\OpenRCT2\src\openrct2\libopenrct2.vcxproj]
          with
          [
              TDeduced=money32
          ] (compilando arquivo de origem actions\GameAction.cpp)
...
D:\github\OpenRCT2\src\openrct2\localisation\Localisation.h(132,44): message : Conversão é uma conversão padrão válida, que pode ser realizada implicitamente ou pelo uso de static_cast, conversão C-style ou con
versão function-style (compilando arquivo de origem actions\GameAction.cpp) [D:\github\OpenRCT2\src\openrct2\libopenrct2.vcxproj]
```

brenoguim: Oh, that's in portuguese.
The first one says it can't converte `char*` to `uintptr_t` without reinterpret cast.
The second one says it won't convert `money32` ot `uintptr_t` using reinterpret cast because static_cast will do the job.

Make up your mind compiler!

janisozaur: It looks to me like there are two options out of this:
1. Template-fu
2. Comment
I will leave the decision up to you.

brenoguim: I added a comment. I'm not entirely sure what you meant with 'template-fu'. Anyway, there is still more to be done in this code, like adding some static asserts and removing more macros if possible. I'll surely get back to this.

brenoguim: Found a way to solve it :)

PR #11358, opened 2020-04-19
Switch tile inspector copy paste button placement

oli414:
The current button placement in the tile inspector has the paste button on the left, with the copy button on the right. This, however, doesn't seem in line with how most software seems to place these buttons.

This PR changes this to make it copy-paste, instead of paste-copy.

image

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,6 +3,7 @@
 - Feature: [#9029] Open doors with the tile inspector.
 - Feature: [#11231] Change shortcut window list order to be more intuitive, and split it into logical sections.
 - Feature: [#11306] Path additions are now kept when replacing the path.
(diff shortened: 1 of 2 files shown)


oli414:
I often find myself intuitively clicking the wrong button. One example where the buttons are placed in the paste-copy order is the Microsoft Office package. However in Windows Explorer it is copy-paste. All context menu based copy-paste buttons follow the copy-paste order.

In UI copy-paste buttons don't seem to be very common anymore, but let me know if you disagree and have examples of the paste-copy button orders.

Gymnasiast:
MS Office 2003 and LibreOffice (and, I would assume, almost all other office software) do use cut-copy-paste. It will cause some muscle memory issues briefly, but this makes more sense.

Gymnasiast:
I would like a changelog entry. It can be put under "Change" - let's keep it neutral.

PR #11360, opened 2020-04-19
Refactor/rtd more

Gymnasiast:
This contains some additional cleanup of RideData.cpp/.h. It is probably easier to review when looking at commits separately.


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2540,8 +2541,9 @@ void window_ride_construction_update_enabled_track_pieces()
     if (rideEntry == nullptr)
         return;
 
-    int32_t rideType = (_currentTrackAlternative & RIDE_TYPE_ALTERNATIVE_TRACK_TYPE) ? RideData4[ride->type].alternate_type
-                                                                                     : ride->type;
+    int32_t rideType = (_currentTrackAlternative & RIDE_TYPE_ALTERNATIVE_TRACK_TYPE)
+        ? RideTypeDescriptors[ride->type].AlternateType
+        : ride->type;
```
Gymnasiast, line 2546: Duplication of the `ride_get_alternative_type()` function.

tupaschoal: Is this a note for future or do you intend to fix this in the patch?

Gymnasiast: I intend to fix it in this patch.

Gymnasiast:
> did we never use the content in "Remove unused bits from rct_ride_data_5" or is it already present somewhere else?

It removed the bonus_value parameter (which was unused, the comment said it was replaced with the `rideBonusValue` array) and a padding variable.

PR #11378, opened 2020-04-20
Adjust function names according to coding style

brenoguim:


src/openrct2/localisation/Localisation.h
```diff
@@ -117,17 +117,17 @@ class Formatter
...
     }
 
-    auto buf()
+    auto Buf()
     {
         return CurrentBuf;
     }
 
-    std::size_t bytes() const
+    std::size_t NumBytes() const
```
tupaschoal, line 125: I see what you did here, but I don't mind, it seems clearer.

PR #11381, opened 2020-04-21
11162 : Usage of more modern C++ dynamic allocation methods

frutiemax:
Usage of more modern C++ dynamic allocation methods. IN PROGRESS


src/openrct2/cmdline/BenchSpriteSort.cpp
```diff
@@ -121,12 +121,14 @@ static std::vector<paint_session> extract_paint_session(const std::string parkFi
         dpi.width = resolutionWidth;
         dpi.height = resolutionHeight;
         dpi.pitch = 0;
-        dpi.bits = (uint8_t*)malloc(dpi.width * dpi.height);
+        //dpi.bits = (uint8_t*)malloc(dpi.width * dpi.height);
+        dpi.bits = new uint8_t[dpi.width * dpi.height];
```
tupaschoal, line 125: Please modify the line instead of duplicating it and commenting the old one

src/openrct2/cmdline/BenchSpriteSort.cpp
```diff
@@ -121,12 +121,14 @@ static std::vector<paint_session> extract_paint_session(const std::string parkFi
...
 
         log_info("Obtaining sprite data...");
         viewport_render(&dpi, &viewport, 0, 0, viewport.width, viewport.height, &sessions);
 
-        free(dpi.bits);
+        //free(dpi.bits);
```
tupaschoal, line 130: Please modify the line instead of duplicating it and commenting the old one

src/openrct2/cmdline/BenchSpriteSort.cpp
```diff
@@ -121,12 +121,14 @@ static std::vector<paint_session> extract_paint_session(const std::string parkFi
         dpi.width = resolutionWidth;
         dpi.height = resolutionHeight;
         dpi.pitch = 0;
-        dpi.bits = (uint8_t*)malloc(dpi.width * dpi.height);
+        //dpi.bits = (uint8_t*)malloc(dpi.width * dpi.height);
```
brenoguim, line 124: Perhaps you want to remove old code?

src/openrct2/cmdline/BenchSpriteSort.cpp
brenoguim, line 130: remove old code

src/openrct2/drawing/TTFSDLPort.cpp
```diff
@@ -995,6 +995,7 @@ void TTF_CloseFont(TTF_Font* font)
             fclose(font->src);
         }
         free(font);
+        delete font;
```
brenoguim, line 998: You should either delete or free the font. Never both. Also, you gotta make sure whoever created this font used `new` for `delete` to be used.

tupaschoal, line 998: This change is wrong, all of the allocations of font are done using `malloc`. Either you have to update them to use `new` or you have to leave as it was.

This way you're doing what's called as mismatched free()/delete. `malloc` always goes with `free` and `new` with `delete`, you shouldn't mix and match, especially double deletion like here.

src/openrct2/platform/Windows.cpp
```diff
@@ -222,15 +221,14 @@ bool platform_get_steam_path(utf8* outPath, size_t outSize)
         return false;
     }
 
-    wSteamPath = (wchar_t*)malloc(size);
-    result = RegQueryValueExW(hKey, L"SteamPath", nullptr, &type, (LPBYTE)wSteamPath, &size);
+    std::unique_ptr<wchar_t> wSteamPath(new wchar_t[size]);
```
brenoguim, line 224: This is creating a `std::unique_ptr<wchar_t>` which will call `delete` on the destructor. But the resource being handled is an array, so `delete[]` should be called.
Hence the correct syntax would be `auto wSteamPath = std::make_unique<wchar_t[]>(size)`

tupaschoal, line 224:
```suggestion
    auto wSteamPath = std::make_unique<wchar_t[]>(size);
```

src/openrct2/util/SawyerCoding.cpp
```diff
@@ -42,7 +43,7 @@ uint32_t sawyercoding_calculate_checksum(const uint8_t* buffer, size_t length)
  */
 size_t sawyercoding_write_chunk_buffer(uint8_t* dst_file, const uint8_t* buffer, sawyercoding_chunk_header chunkHeader)
 {
-    uint8_t *encode_buffer, *encode_buffer2;
+    std::unique_ptr<uint8_t> encode_buffer, encode_buffer2;
```
brenoguim, line 46: These are pointing to arrays, therefore they must be declared as `std::unique_ptr<uint8_t[]>`

src/openrct2/util/SawyerCoding.cpp
```diff
@@ -61,35 +62,29 @@ size_t sawyercoding_write_chunk_buffer(uint8_t* dst_file, const uint8_t* buffer,
             // fwrite(buffer, 1, chunkHeader.length, file);
             break;
         case CHUNK_ENCODING_RLE:
-            encode_buffer = static_cast<uint8_t*>(malloc(0x600000));
-            chunkHeader.length = static_cast<uint32_t>(encode_chunk_rle(buffer, encode_buffer, chunkHeader.length));
+            encode_buffer = std::unique_ptr<uint8_t>(new uint8_t[0x600000]);
```
brenoguim, line 65: Use `std::make_unique<uint8_t[]>(0x600000)`

brenoguim: Review all array allocations. This repeats throughout the change.

src/openrct2/util/SawyerCoding.cpp
```diff
@@ -109,14 +104,14 @@ size_t sawyercoding_decode_sc4(const uint8_t* src, uint8_t* dst, size_t length,
     size_t decodedLength = decode_chunk_rle_with_size(src, dst, length - 4, bufferLength);
 
     // Decode
-    for (size_t i = 0x60018; i <= std::min(decodedLength - 1, static_cast<size_t>(0x1F8353)); i++)
+    for (size_t i = 0x60018; i <= std::min(decodedLength - 1, (size_t)0x1F8353); i++)
```
brenoguim, line 107: dont convert a static_cast into a C-Style cast

brenoguim: Don't convert any named cast into a C-Style cast (this repeats throughout the change).

Named casts are safer than C-Style casts.

src/openrct2/util/SawyerCoding.cpp
```diff
@@ -61,35 +62,29 @@ size_t sawyercoding_write_chunk_buffer(uint8_t* dst_file, const uint8_t* buffer,
             // fwrite(buffer, 1, chunkHeader.length, file);
             break;
         case CHUNK_ENCODING_RLE:
-            encode_buffer = static_cast<uint8_t*>(malloc(0x600000));
-            chunkHeader.length = static_cast<uint32_t>(encode_chunk_rle(buffer, encode_buffer, chunkHeader.length));
+            encode_buffer = std::unique_ptr<uint8_t>(new uint8_t[0x600000]);
+            chunkHeader.length = (uint32_t)encode_chunk_rle(buffer, encode_buffer.get(), chunkHeader.length);
```
tupaschoal, line 66: Please don't convert it back to c-style cast.

tupaschoal:
I've labelled it as Work in progress and converted to draft. Please click "ready to review" when it's finished.

src/openrct2/drawing/ImageImporter.cpp
```diff
@@ -129,18 +130,30 @@ std::tuple<void*, size_t> ImageImporter::EncodeRLE(const int32_t* pixels, uint32
     };
 
     auto src = pixels;
-    auto buffer = (uint8_t*)std::malloc((height * 2) + (width * height * 16));
-    if (buffer == nullptr)
+   // auto buffer = (uint8_t*)std::malloc((height * 2) + (width * height * 16));
```
Gymnasiast, line 133: Don't comment out code - just delete it.

src/openrct2/util/SawyerCoding.cpp
frutiemax: Something screwed up when I checkout the code, I've never convert these back to C-style...

src/openrct2/drawing/ImageImporter.cpp
```diff
@@ -129,18 +132,21 @@ std::tuple<void*, size_t> ImageImporter::EncodeRLE(const int32_t* pixels, uint32
...
-    auto buffer = static_cast<uint8_t*>(std::malloc((height * 2) + (width * height * 16)));
-    if (buffer == nullptr)
+    auto buffer = std::vector<uint8_t>();
+    buffer.resize((height * 2) + (width * height * 16));
+
+    if (buffer.data() == nullptr)
+
     {
         throw std::bad_alloc();
     }
```
brenoguim, line 142: This code can be removed.  `std::vector` will already throw a  `std::bad_alloc` if it can't get the memory

src/openrct2/drawing/ImageImporter.cpp
```diff
@@ -129,18 +132,21 @@ std::tuple<void*, size_t> ImageImporter::EncodeRLE(const int32_t* pixels, uint32
...
+    std::fill_n(buffer.begin(), (height * 2) + (width * height * 16), 0);
```
brenoguim, line 143: This can be passed to the vector constructor as well

src/openrct2/object/ObjectManager.h
```diff
@@ -33,7 +33,7 @@ interface IObjectManager
     virtual Object* LoadObject(const rct_object_entry* entry) abstract;
     virtual void LoadObjects(const rct_object_entry* entries, size_t count) abstract;
     virtual void LoadDefaultObjects() abstract;
-    virtual void UnloadObjects(const rct_object_entry* entries, size_t count) abstract;
+    virtual void UnloadObjects(std::vector<rct_object_entry> & entries) abstract;
```
brenoguim, line 36: remove space between `>` and `&`

src/openrct2/platform/Windows.cpp
```diff
@@ -222,15 +221,14 @@ bool platform_get_steam_path(utf8* outPath, size_t outSize)
         return false;
     }
 
-    wSteamPath = (wchar_t*)malloc(size);
-    result = RegQueryValueExW(hKey, L"SteamPath", nullptr, &type, (LPBYTE)wSteamPath, &size);
+    auto wSteamPath = std::make_unique<wchar_t[]>(size);
```
brenoguim, line 224: This `size` refer to the number of bytes to fit `wchar_t` elements. So either this was a bug in the previous code or this should be called with `size/sizeof(wchar_t)`. Please investigate a bit where this `size` comes from and if it is already adjusted to  `wchar_t`.

src/openrct2/util/SawyerCoding.cpp
```diff
@@ -112,7 +107,7 @@ size_t sawyercoding_decode_sc4(const uint8_t* src, uint8_t* dst, size_t length,
     for (size_t i = 0x60018; i <= std::min(decodedLength - 1, static_cast<size_t>(0x1F8353)); i++)
         dst[i] = dst[i] ^ 0x9C;
 
-    for (size_t i = 0x60018; i <= std::min(decodedLength - 1, static_cast<size_t>(0x1F8350)); i += 4)
+    for (size_t i = 0x60018; i <= std::min(decodedLength - 1, static_cast<size_t> (0x1F8350)); i += 4)
```
brenoguim, line 110: unintended change

src/openrct2/platform/Windows.cpp
frutiemax: It seems that it stores the size of the data in bytes, so it might be a bug.

tupaschoal:
Hey @frutiemax, can you mark the issues people open as resolved, once you fix them? It makes the page a lot clearer and we can see if you have acted upon them or not.

Thanks!

tupaschoal:
@frutiemax can you please rebase and squash the commits if you intend to still work on this?

PR #11386, opened 2020-04-21
Fix strlogicalcmp

janisozaur:
We recently had reports about invalid sorting of such items

### DIFF
--- a/src/openrct2/util/Util.cpp
+++ b/src/openrct2/util/Util.cpp
@@ -323,37 +323,32 @@ int32_t strcicmp(char const* a, char const* b)
 // - Guest 100
 // - John v2.0
 // - John v2.1
-int32_t strlogicalcmp(char const* a, char const* b)
+int32_t strlogicalcmp(const char* s1, const char* s2)
 {
-    for (;; a++, b++)
+    for (;;)
     {
-        int32_t result = tolower(*a) - tolower(*b);
-        bool both_numeric = *a >= '0' && *a <= '9' && *b >= '0' && *b <= '9';
-        if (result != 0 || !*a || both_numeric)
-        { // difference found || end of string
-            if (both_numeric)
-            { // a and b both start with a number
-                // Get the numbers in the string at current positions
-                int32_t na = 0, nb = 0;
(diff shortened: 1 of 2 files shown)


AaronVanGeffen:
Your new rewrite appears to have done away with case insensitivity. Could you restore this?

janisozaur:
Can you provide the suggestion to tests?

IntelOrca:
`EXPECT_EQ(strlogicalcmp("foo13", "FOO13"), 0);`

AaronVanGeffen:
When you said "reports about invalid sorting of such items", were you thinking of any issues in particular? I recall there were issues with lists not being ordered as expected, but they turned out to be using alphabetical order rather than logical order.

janisozaur:
I think there was a report on reddit, not here.

AaronVanGeffen:
I wish you'd posted it in the OP, as it's hard to see exactly what this PR is trying to fix now. Combined with the regression mentioned above, how would you like to proceed?

src/openrct2/util/Util.cpp
```diff
@@ -323,37 +323,32 @@ int32_t strcicmp(char const* a, char const* b)
...
-                return na - nb;
-            }
+        if (*s2 == '\0')
+            return *s1 != '\0';
+        else if (*s1 == '\0')
+            return -1;
+        else if (!(isdigit(*s1) && isdigit(*s2)))
+        {
+            if (tolower(*s1) != tolower(*s2))
+                return (int)*s1 - (int)*s2;
```
IntelOrca, line 337:
```suggestion
                return tolower(*s1) - tolower(*s2);
```

IntelOrca: EXPECT_LT(strlogicalcmp("A", "b"), 0);
EXPECT_LT(strlogicalcmp("a", "B"), 0);
EXPECT_GT(strlogicalcmp("B", "a"), 0);
EXPECT_GT(strlogicalcmp("b", "A"), 0);

IntelOrca:
What is this actually fixing again?

PR #11388, opened 2020-04-21
Replacing usage of set_format_arg by Formatter::Common()

brenoguim:


Gymnasiast:
It would be nice if we could eventually get rid of the global format args. Is that the/an eventual goal of your refactors?

src/openrct2/paint/tile_element/Paint.Path.cpp
```diff
@@ -447,17 +447,20 @@ static void sub_6A4101(
             uint16_t scrollingMode = railingEntry->scrolling_mode;
             scrollingMode += direction;
 
-            set_format_arg(0, uint32_t, 0);
-            set_format_arg(4, uint32_t, 0);
+            // FIXME this is a bit hackish. We clear the 8 first bytes just in case
+            // we don't write to all of them later on. Check if really necessary.
+            Formatter::Common().Add<uint32_t>(0).Add<uint32_t>(0);
```
duncanspumpkin, line 452: It is required due to `scrolling_text_setup` doing a memcmp of another set of args to see if it matches this set of args.
Eventually when `scrolling_text_setup` takes the formatter variable this can be removed.

brenoguim: I updated the comment to describe that so the next developer has this hint. Hopefully not many developers will pass here. I hope to get these things cleaned up. It will be beautiful :)

src/openrct2/peep/Peep.cpp
```diff
@@ -2047,28 +2048,29 @@ bool Peep::SetName(const std::string_view& value)
  */
 void peep_thought_set_format_args(const rct_peep_thought* thought)
 {
-    set_format_arg(0, rct_string_id, PeepThoughts[thought->type]);
+    auto ft = Formatter::Common();
+    ft.Add<rct_string_id>(PeepThoughts[thought->type]);
```
tupaschoal, line 2052: Why not joining these in the same line, like you did on `Paint.Path.cpp `?

brenoguim: I start a new formatter when the "set_format_arg" is writing to the position 0, and the next set_format_args write incrementally (2, 4...).

brenoguim: You can see that in Paint.Path.cpp it writes to 0, then 4, then starts again writing to 0.

tupaschoal: I think it could've been collapsed, but this is not a huge deal, so let's move on

PR #11395, opened 2020-04-21
 Close #11209: Show a warning when the user is running OpenRCT2 in Wine

### LINKED ISSUE #11209: Detect when user is running OpenRCT2 via Wine
If a user is running OpenRCT2 via Wine, it is likely to be a mistake. Perhaps they assumed they needed the Windows version because they extract their RCT2 files that way.

This could have lead to crashes like https://github.com/OpenRCT2/OpenRCT2/issues/11205 .

When the game detects Wine, it could then display a warning in the same way that it already warns for running as root.

ifimfree:
Adds functionality to display a warning (similar to that of the warning displayed when running on root) when the user is detected to be using Wine to run the application instead of the better supported Linux per Issue #11209

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3681,6 +3681,7 @@ STR_6364    :{RED}{COMMA16} person has died in an accident on {STRINGID}
 STR_6365    :Ride casualties
 STR_6366    :Stuck or stalled vehicles
 STR_6367    :{WINDOW_COLOUR_2}Animation frame:
+STR_6368    :For compatibility reasons, it is not recommended to run OpenRCT2 with Wine. OpenRCT2 has native support for macOS, Linux, FreeBSD and OpenBSD.
 
 #############
 # Scenarios #
--- a/src/openrct2/localisation/StringIds.h
+++ b/src/openrct2/localisation/StringIds.h
@@ -3918,6 +3918,8 @@ enum
 
     STR_TILE_INSPECTOR_WALL_ANIMATION_FRAME = 6367,
 
+    STR_WINE_NOT_RECOMMENDED = 6368,
+
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
     /* MAX_STR_COUNT = 32768 */ // MAX_STR_COUNT - upper limit for number of strings, not the current count strings
 };
--- a/src/openrct2/platform/Platform2.h
+++ b/src/openrct2/platform/Platform2.h
@@ -45,6 +45,7 @@ namespace Platform
     bool IsOSVersionAtLeast(uint32_t major, uint32_t minor, uint32_t build);
 #endif
 
+    bool IsRunningInWine();
     bool IsColourTerminalSupported();
     bool HandleSpecialCommandLineArgument(const char* argument);
     uintptr_t StrDecompToPrecomp(utf8* input);
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,6 +3,7 @@
 - Feature: [#9029] Open doors with the tile inspector.
 - Feature: [#11231] Change shortcut window list order to be more intuitive, and split it into logical sections.
 - Feature: [#11306] Path additions are now kept when replacing the path.
+- Change: [#11209] Warn when user is running OpenRCT2 through Wine.
 - Change: [#11358] Switch copy and paste button positions in tile inspector.
 - Fix: [#1148] Research funding dropdown not shown in finances window.
 - Fix: [#6119] Advertising campaign for ride window not updated properly (original bug).
--- a/src/openrct2/Context.cpp
+++ b/src/openrct2/Context.cpp
@@ -52,6 +52,7 @@
 #include "object/ObjectRepository.h"
 #include "paint/Painter.h"
 #include "platform/Crash.h"
+#include "platform/Platform2.h"
 #include "platform/platform.h"
 #include "ride/TrackDesignRepository.h"
 #include "scenario/Scenario.h"
@@ -386,6 +387,19 @@ namespace OpenRCT2
                 else
                 {
                     _uiContext->ShowMessageBox(elevationWarning);
+                }
+            }
+
+            if (Platform::IsRunningInWine())
+            {
+                std::string wineWarning = _localisationService->GetString(STR_WINE_NOT_RECOMMENDED);
+                if (gOpenRCT2Headless)
+                {
+                    Console::Error::WriteLine(wineWarning.c_str());
+                }
+                else
+                {
+                    _uiContext->ShowMessageBox(wineWarning);
                 }
             }
 
--- a/src/openrct2/platform/Platform.Posix.cpp
+++ b/src/openrct2/platform/Platform.Posix.cpp
@@ -99,6 +99,11 @@ namespace Platform
         }
         return isSupported;
     }
+
+    bool IsRunningInWine()
+    {
+        return false;
+    }
 } // namespace Platform
 
 #endif
--- a/src/openrct2/platform/Platform.Win32.cpp
+++ b/src/openrct2/platform/Platform.Win32.cpp
@@ -241,6 +241,17 @@ namespace Platform
         return result;
     }
 
+    bool IsRunningInWine()
+    {
+        HMODULE ntdllMod = GetModuleHandleW(L"ntdll.dll");
+
+        if (ntdllMod && GetProcAddress(ntdllMod, "wine_get_version"))
+        {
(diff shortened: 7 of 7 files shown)


src/openrct2/platform/platform.h
```diff
@@ -124,6 +124,7 @@ uint8_t platform_get_locale_temperature_format();
 uint8_t platform_get_locale_date_format();
 bool platform_process_is_elevated();
 bool platform_get_steam_path(utf8* outPath, size_t outSize);
+bool is_running_in_wine();
```
AaronVanGeffen, line 127: platform.h is a legacy file. Please use Platform2.h instead. We are in the process of transitioning, sorry about that…

A version of this function will need to be implemented for other platforms, too, even if it just returns false. I think you can put it in Platform.Posix.cpp.

src/openrct2/platform/Windows.cpp
```diff
@@ -693,6 +693,17 @@ void platform_remove_file_associations()
     SHChangeNotify(SHCNE_ASSOCCHANGED, SHCNF_IDLIST, nullptr, nullptr);
 }
 
+bool is_running_in_wine()
```
AaronVanGeffen, line 696: Please define this in Platform.Win32.cpp instead.

AaronVanGeffen:
Thanks for tackling this. Could you split off the unrelated (flag) changes to another PR?

src/openrct2/localisation/StringIds.h
```diff
@@ -3911,8 +3911,10 @@ enum
 
     STR_COPY_INPUT_TO_CLIPBOARD = 6363,
 
+    STR_WINE_NOT_RECOMMENDED = 6368,
+
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    STR_COUNT = 32768
+    STR_COUNT = 32769
```
Gymnasiast, line 3917: I understand why you incremented this, but it shouldn't be incremented when you add strings.

data/language/en-GB.txt
```diff
@@ -3677,6 +3677,7 @@ STR_6360    :{SMALLFONT}{BLACK}{COMMA32}
 STR_6361    :Enable lighting effects on rides (experimental)
 STR_6362    :{SMALLFONT}{BLACK}If enabled, vehicles for tracked rides will be lit up at night.
 STR_6363    :Copied text to clipboard
+STR_6368    :For compatability reasons, it is not recommended to run OpenRCT2 with WINE. Instead, Linux is supported.
```
tupaschoal, line 3680: A few comments:
- The number should be the increment from the previous
- Typo fix and a change on the last sentence which I think makes it clearer.
```suggestion
STR_6364    :For compatibility reasons, it is not recommended to run OpenRCT2 with WINE. Linux is natively supported.
```

src/openrct2/localisation/StringIds.h
tupaschoal, line 3917: Just adding a suggestion reflecting what Gymnasiast said.
```suggestion
    STR_COUNT = 32768
```

src/openrct2/localisation/StringIds.h
```diff
@@ -3911,8 +3911,10 @@ enum
 
     STR_COPY_INPUT_TO_CLIPBOARD = 6363,
 
+    STR_WINE_NOT_RECOMMENDED = 6368,
```
tupaschoal, line 3914:
```suggestion
    STR_WINE_NOT_RECOMMENDED = 6364,
```
To reflect the suggestion above.

src/openrct2/platform/Platform2.h
```diff
@@ -44,7 +44,7 @@ namespace Platform
 #ifdef _WIN32
     bool IsOSVersionAtLeast(uint32_t major, uint32_t minor, uint32_t build);
 #endif
-
+    bool is_running_in_wine();
```
tupaschoal, line 47: This should follow the style of the file (which is actually our coding guideline
```suggestion
    bool IsRunningInWine();
```

data/language/en-GB.txt
```diff
@@ -3677,7 +3677,7 @@ STR_6360    :{SMALLFONT}{BLACK}{COMMA32}
 STR_6361    :Enable lighting effects on rides (experimental)
 STR_6362    :{SMALLFONT}{BLACK}If enabled, vehicles for tracked rides will be lit up at night.
 STR_6363    :Copied text to clipboard
-
+STR_6368    :For compatibility reasons, it is not recommended to run OpenRCT2 with WINE. Linux is natively supported.
```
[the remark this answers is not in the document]
janisozaur: Wine is available for other platforms: BSD, OS X, Android… don't assume Linux.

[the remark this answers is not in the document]
Gymnasiast: @tupaschoal STR_6364 has already been used. The number is correct, but the OP has branched off an outdated `develop`.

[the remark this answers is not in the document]
Gymnasiast: The Wine project writes its name like that - with only a capital at the start.

To address this and janisozaur's point:
```suggestion
STR_6368    :For compatibility reasons, it is not recommended to run OpenRCT2 with Wine. OpenRCT2 has native support for macOS, Linux, FreeBSD and NetBSD.
```

Gymnasiast:
You reverted a few commits, but you should have dropped them using a rebase. Since there are some conflicts, you will need to do so anyway.

If you're unsure how to do it, we can help you with it.

src/openrct2/localisation/StringIds.h
pizza2004: I see people make this mistake all the time, perhaps it would be a good idea to change this name to something like MAX_STR_COUNT instead?

data/language/en-GB.txt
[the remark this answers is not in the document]
tupaschoal: Ah, I see, thanks for clarifying Michael

src/openrct2/localisation/StringIds.h
tupaschoal: It's a good idea indeed, or to add a comment explaining. I'll see if I can tend to it today

ifimfree:
Thanks all for being so patient with me, I'm quite new both to this project and (honestly) also to Git.

In terms of needing to rebase, I've never had to do that sort of thing before.  Will the following work?  This is what I found by looking around but want to be sure this will give the intended result.

git remote add upstream https://github.com/OpenRCT2/OpenRCT2.git
git fetch upstream
git rebase upstream/develop
git push origin wine-detection

src/openrct2/localisation/StringIds.h
```diff
@@ -3911,6 +3911,8 @@ enum
 
     STR_COPY_INPUT_TO_CLIPBOARD = 6363,
 
+    STR_WINE_NOT_RECOMMENDED = 6364,
```
Gymnasiast, line 3914:
```suggestion
    STR_WINE_NOT_RECOMMENDED = 6368,
```

tupaschoal:
> Thanks all for being so patient with me, I'm quite new both to this project and (honestly) also to Git.
> 
> In terms of needing to rebase, I've never had to do that sort of thing before. Will the following work? This is what I found by looking around but want to be sure this will give the intended result.
> 
> git remote add upstream https://github.com/OpenRCT2/OpenRCT2.git
> git fetch upstream
> git rebase upstream/develop
> git push origin wine-detection

There are multiple ways of doing so, typically the upstream is already configured, you can check it out with `git remote -v`. Then you need to:
1 - `git checkout develop` checkout your local develop
2 - `git fetch upstream` fetch changes from the main `OpenRCT2` repo
3 - `git rebase upstream/develop` rebase your develop onto OpenRCT2/develop
4 - `git checkout wine-detection` checkout your branch again
5 - `git rebase -i develop` this is to interactively rebase your current branch atop your own develop, you can omit the `-i` i you want

ifimfree:
I seem to have messed up what/how I rebased - currently trying to fix this.

src/openrct2/platform/Posix.cpp
```diff
@@ -488,6 +489,11 @@ bool platform_process_is_elevated()
 #    endif // __EMSCRIPTEN__
 }
 
+bool Platform::IsRunningInWine()
```
Gymnasiast, line 492: This needs to be in Platform.Posix.cpp, in the same way it had to go into Platform.Win32.cpp.

Gymnasiast:
I have nuked your branch and cherry-picked the commits. They'll still need to be squashed on merging, and there are still two line notes to address, along with the formatting.

Issue #9473, opened 2019-06-24
Move Vehicle Update into member function


duncanspumpkin:
The vehicle update functions are currently static functions in the Vehicle.cpp file. All of them should really be `rct_vehicle` only member functions and should therefore be moved to improve readability. To begin with move only the `vehicle_update` function into a public member function.
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Vehicle.cpp#L1923-L2030

Create a new public member function of the `rct_vehicle` struct called `Update`. Rename the existing `vehicle_update` function and correct all of the variables that are used in the function. After this has been completed you can continue on to move every update function into the `rct_vehicle` struct. All of the subsequent member functions should probably be private. Don't try to move them all at once take your time and split them up into a couple compilable commits, it will be easier to review and then you wont get lost.

`cable_lift_update`, `vehicle_update_measurements`, `vehicle_update_moving_to_end_of_station`,` vehicle_update_waiting_for_passengers`, `vehicle_update_waiting_to_depart`, `vehicle_update_crash`, `vehicle_update_dodgems_mode`, `vehicle_update_swinging`, `vehicle_update_simulator_operating`, `vehicle_update_top_spin_operating`, `vehicle_update_ferris_wheel_rotating`, `vehicle_update_space_rings_operating`, `vehicle_update_haunted_house_operating`, `vehicle_update_crooked_house_operating`, `vehicle_update_rotating`, `vehicle_update_departing` , `vehicle_update_travelling`, ` vehicle_update_travelling_cable_lift`, `vehicle_update_travelling_boat`, `vehicle_update_arriving`, `vehicle_update_unloading_passengers`, `vehicle_update_waiting_for_cable_lift`, `vehicle_update_showing_film`, `vehicle_update_doing_circus_show`, ` vehicle_update_sound`

In addition make sure all local variables are in lower camelCase and all function names are in upper CamelCase. https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style

When commiting in git try to stick to the commit message format. https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

dairemcdonald:
Is it okay if I work on this?

Gymnasiast:
@dairemcdonald Certainly! These issues are specifically created for first-time contributors like you, in order to become familiar with our code.

If you need any help, you can ask us on Gitter. The link is in the readme.md file.

[...]

PR #11402, opened 2020-04-22
Name Flags in PeepThoughtToActionMap (#9473)

mustaphaelghoul:
Temporary flag names so far; progress check


src/openrct2/peep/Peep.cpp
```diff
@@ -2050,7 +2050,7 @@ void peep_thought_set_format_args(const rct_peep_thought* thought)
 {
     set_format_arg(0, rct_string_id, PeepThoughts[thought->type]);
 
-    uint8_t flags = PeepThoughtToActionMap[thought->type].flags;
+    PeepThoughtToAction flags = PeepThoughtToActionMap[thought->type].flags;
     if (flags & 1)
```
duncanspumpkin, line 2054: Change this to use PEEP_THOUGHT_ACTION_FLAG_1

src/openrct2/peep/Peep.h
```diff
@@ -529,6 +529,14 @@ enum PeepRideDecision
     PEEP_RIDE_DECISION_THINKING = 1 << 2,
 };
 
+// Flags used by PeepThoughtToActionMap
+enum PeepThoughtToAction : uint8_t
+{
+    PEEP_THOUGHT_ACTION_FLAG_1 = (1 << 0),
+    PEEP_THOUGHT_ACTION_FLAG_2 = (1 << 1),
+    PEEP_THOUGHT_ACTION_FLAG_4 = (1 << 2),
```
duncanspumpkin, line 537: Flag 3. 

We have been a bit inconsistent in the codebase with what we call flags. We sometimes zero index them so that they are the amount we want to left shit. Other times we have one indexed them so that its the number - 1 of the amount left shited.

src/openrct2/peep/Peep.h
```diff
@@ -529,6 +529,14 @@ enum PeepRideDecision
     PEEP_RIDE_DECISION_THINKING = 1 << 2,
 };
 
+// Flags used by PeepThoughtToActionMap
```
duncanspumpkin, line 532: This is only used in Peep.cpp so it can be moved to peep.cpp just above the `PeepThoughtToActionMap`

src/openrct2/peep/Peep.cpp
```diff
@@ -2050,8 +2058,8 @@ void peep_thought_set_format_args(const rct_peep_thought* thought)
 {
     set_format_arg(0, rct_string_id, PeepThoughts[thought->type]);
 
-    uint8_t flags = PeepThoughtToActionMap[thought->type].flags;
-    if (flags & 1)
+    PeepThoughtToAction flags = PeepThoughtToActionMap[thought->type].flags;
+    if (flags == PEEP_THOUGHT_ACTION_FLAG_1)
```
duncanspumpkin, line 2062: Although this will work you normally would `&` bit flags. Can you revert that change. (The reason you do this is because often a variable would have more than one flag at the same time. That is not the case this time but its always best to be consistent in the codebase)

src/openrct2/peep/Peep.cpp
```diff
@@ -82,186 +82,195 @@ static void peep_128_tick_update(Peep* peep, int32_t index);
...
+    PEEP_THOUGHT_ACTION_FLAG_2 = (1 << 1),
+    PEEP_THOUGHT_ACTION_FLAG_3 = (1 << 2),
+};
+
 /** rct2: 0x00981DB0 */
 static struct
 {
     PeepActionType action;
-    uint8_t flags;
+    PeepThoughtToAction flags;
```
Gymnasiast, line 98: I'd call this `PeepThoughtToActionFlag`. After all, this field does not contain the entire peep-thought-to-action map.

src/openrct2/peep/Peep.cpp
```diff
@@ -82,186 +82,195 @@ static void peep_128_tick_update(Peep* peep, int32_t index);
 static void peep_release_balloon(Guest* peep, int16_t spawn_height);
 // clang-format off
 
+// Flags used by PeepThoughtToActionMap
+enum PeepThoughtToAction : uint8_t
+{
+    PEEP_THOUGHT_ACTION_NO_FLAGS = 0,
+    PEEP_THOUGHT_ACTION_FLAG_1 = (1 << 0),
```
Gymnasiast, line 89:
```suggestion
    PEEP_THOUGHT_ACTION_FLAG_RIDE = (1 << 0),
```

src/openrct2/peep/Peep.cpp
```diff
@@ -82,186 +82,195 @@ static void peep_128_tick_update(Peep* peep, int32_t index);
...
+    PEEP_THOUGHT_ACTION_FLAG_2 = (1 << 1),
```
Gymnasiast, line 90:
```suggestion
    PEEP_THOUGHT_ACTION_FLAG_SHOP_ITEM_SINGULAR = (1 << 1),
```

src/openrct2/peep/Peep.cpp
```diff
@@ -82,186 +82,195 @@ static void peep_128_tick_update(Peep* peep, int32_t index);
...
+    PEEP_THOUGHT_ACTION_FLAG_3 = (1 << 2),
```
Gymnasiast, line 91:
```suggestion
    PEEP_THOUGHT_ACTION_FLAG_SHOP_ITEM_INDEFINITE = (1 << 2),
```

PR #11408, opened 2020-04-22
Replace malloc() and free() in openrct2-win.cpp

Vipulsh:
replaced the usage of malloc with unique pointer for the occurrences in openrct2-win

### DIFF
--- a/src/openrct2-win/openrct2-win.cpp
+++ b/src/openrct2-win/openrct2-win.cpp
@@ -17,12 +17,12 @@
     "\"/manifestdependency:type='win32' name='Microsoft.Windows.Common-Controls' version='6.0.0.0' processorArchitecture='*' publicKeyToken='6595b64144ccf1df' language='*'\"")
 
 // Then the rest
+#include <memory>
 #include <openrct2-ui/Ui.h>
 #include <stdio.h>
 #include <stdlib.h>
 
-static char** GetCommandLineArgs(int argc, wchar_t** argvW);
-static void FreeCommandLineArgs(int argc, char** argv);
+static std::unique_ptr<char*[]> GetCommandLineArgs(int argc, wchar_t** argvW);
 static char* ConvertWideChartoUTF8(const wchar_t* src);
 
 /**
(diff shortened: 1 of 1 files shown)


src/openrct2-win/openrct2-win.cpp
```diff
@@ -17,12 +17,16 @@
...
 // Then the rest
+#include <memory>
 #include <openrct2-ui/Ui.h>
 #include <stdio.h>
 #include <stdlib.h>
 
 static char** GetCommandLineArgs(int argc, wchar_t** argvW);
-static void FreeCommandLineArgs(int argc, char** argv);
+
+//-------removed as manually deallocating the memory is no longer required as using smart ptrs  ,BY VS
```
Gymnasiast, line 27: We have source control to keep track of removed code. Just remove it. Also, no need to graffiti your initials all over the place.

Vipulsh: Sure , thought of it , just have a habit of doing it  .

Vipulsh: I have done the changes as asked

src/openrct2-win/openrct2-win.cpp
```diff
@@ -57,19 +56,9 @@ static char** GetCommandLineArgs(int argc, wchar_t** argvW)
...
 
-static void FreeCommandLineArgs(int argc, char** argv)
-{
-    // Free argv
-    for (int i = 0; i < argc; i++)
-    {
-        free(argv[i]);
-    }
-    free(argv);
+    return argv.get();
```
brenoguim, line 60: Here the code is returning the pointer that is managed by `argv` to the caller. As soon as this function ends, `argv` is destroyed, and the resource is freed.
So in the callsite:
```
char** thePtr = GetCommandLineArgs(...)
```
`thePtr` no longer points to a valid resource, since it was destroyed when `GetCommandLineArgs` ended.

What you want to do here is to return `argv` and make this function return a `std::unique_ptr<char*[]>`. By doing so, the caller will still hold the resource it needs for as long as it needs.

Vipulsh: Changes made.Please review.

tupaschoal:
Hello @Vipulsh ,

The CI didn't succeed on Windows, so there's an outstanding problem to be fixed. Do note that the formatting is also broken. We can review once those are dealt with :)

Thanks

src/openrct2-win/openrct2-win.cpp
```diff
@@ -57,19 +56,9 @@ static char** GetCommandLineArgs(int argc, wchar_t** argvW)
...
-{
-    // Free argv
-    for (int i = 0; i < argc; i++)
-    {
-        free(argv[i]);
-    }
-    free(argv);
+    return argv;
 }
-
```
tupaschoal, line 72: Add this line back.

Vipulsh: I don't know why it's still failing

janisozaur: It's still miscormatted. Either run `clang-format` locally or apply the patch from CI.

Vipulsh: It worked this time .Thanks for the tip

tupaschoal:
@Vipulsh since this is incomplete, I'll be closing this. If you want to resume your work, please tend to the issue above and reopen the PR.

PR #11409, opened 2020-04-22
Rename and improve documentation on MAX_STR_COUNT

tupaschoal:
This is a follow-up to @pizza2004 's comment on https://github.com/OpenRCT2/OpenRCT2/pull/11395#discussion_r412962951

This specific piece of code is a common source of confusion for new contributors, so I figured renaming it and adding a comment might prevent some of those.

### DIFF
--- a/src/openrct2/localisation/StringIds.h
+++ b/src/openrct2/localisation/StringIds.h
@@ -3919,8 +3919,7 @@ enum
     STR_TILE_INSPECTOR_WALL_ANIMATION_FRAME = 6367,
 
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    // MAX_STR_COUNT - upper limit for number of strings, not the current count strings
-    MAX_STR_COUNT = 32768
+    /* MAX_STR_COUNT = 32768 */ // MAX_STR_COUNT - upper limit for number of strings, not the current count strings
 };
 
 #endif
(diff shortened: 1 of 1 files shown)


src/openrct2/localisation/StringIds.h
```diff
@@ -3919,7 +3919,8 @@ enum
     STR_TILE_INSPECTOR_WALL_ANIMATION_FRAME = 6367,
 
     // Have to include resource strings (from scenarios and objects) for the time being now that language is partially working
-    STR_COUNT = 32768
+    // MAX_STR_COUNT - upper limit for number of strings, not the current count strings
+    MAX_STR_COUNT = 32768
```
janisozaur, line 3923: Why do we even have this value? Renaming it breaks nothing, why not just change it to
```suggestion
    MAX_STR_COUNT_DO_NOT_CHANGE = 32768
```

Gymnasiast:
@janisozaur 's comment prompted me to look for through the codebase. It isn't used any more _now_ - I think it may have been used when we still had string tables for ride names and the like.

So basically, I think we can just remove it altogether. What would you say, @IntelOrca ?

janisozaur:
I'm fine with it staying, as it serves as a reminder - it's not 65535, which would likely cause some compilation issues due to types, but rather something semi-arbitrary. I'm just pointing out we can change the name to be more descriptive, as it is not used currently.

If it needs to be used and the name needs to be sensible, we can add `static_assert` somewhere in one of CPP files.

tupaschoal:
I guessed that even though it is not used, it might be acting as a warning there that something will break if new strings are added after the limit. I haven't really checked if that's the case.

IntelOrca:
You could make it a comment to show it isn't used.

tupaschoal:
I commented it out then.

tupaschoal:
Is this ok now @IntelOrca ?

PR #11410, opened 2020-04-22
Prototype Commit Lint integration

tupaschoal:
~I have no clue whether this will work or not, so I'm opening a draft PR to prototype~
**DON'T MERGE, IF WE SETTLE ON THIS I'LL REBASE AND SQUASH :D**

Apparently we can have a good control over commit messages with this `commitlint` extension, made possible through the `commitlint-github-action`. The rules are on [commitlint](https://commitlint.js.org/#/reference-rules), but I'm basically:

1. Erroring out if the commit message exceeds 72 chars
2. ~Warning~ Erroring if it's smaller than ~20~ 10 chars.
3. Making sure there is a line break between the message and the body and the footer.

This PR has 5 commits, some intentionally long/short just to test that it works.
```
Prototype Commit Lint integration
A very long commit message just to see how the lint reports failures when there are multiple patches in the PR
Short commit
Some normal commit with a regular size
Enable commitlint for pull requests and commits
```

This is the output of the CI job:
```
⧗   input: Short commit
⚠   header must not be shorter than 20 characters, current length is 12 [header-min-length]

⚠   found 0 problems, 1 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

⧗   input: A very long commit message just to see how the lint reports failures when there are multiple patches in the PR
✖   header must not be longer than 72 characters, current length is 110 [header-max-length]

✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

### DIFF
--- a/.commitlint.json
+++ b/.commitlint.json
@@ -32,8 +32,7 @@
           20
       ],
       "scope-empty": [
-          2,
-          "always"
+          0
       ],
       "subject-case": [
           0
--- a/.commitlint.json
+++ b/.commitlint.json
@@ -27,9 +27,9 @@
           72
       ],
       "header-min-length": [
-          1,
+          2,
           "always",
-          20
+          10
       ],
       "scope-empty": [
           0
--- a/.commitlint.json
+++ b/.commitlint.json
@@ -1,0 +1,69 @@
+
+{
+  "rules": {
+      "body-leading-blank": [
+          2,
+          "always"
+      ],
+      "body-max-length": [
+          0
+      ],
+      "body-min-length": [
+          0
+      ],
+      "footer-leading-blank": [
+          2,
+          "always"
+      ],
+      "footer-max-length": [
+          0
+      ],
+      "footer-min-length": [
+          0
+      ],
+      "header-max-length": [
+          2,
+          "always",
+          72
+      ],
+      "header-min-length": [
+          1,
+          "always",
+          20
+      ],
+      "scope-empty": [
+          2,
+          "always"
+      ],
+      "subject-case": [
+          0
+      ],
+      "subject-empty": [
+          0
+      ],
+      "subject-full-stop": [
+          0
+      ],
+      "subject-max-length": [
+          0
+      ],
+      "subject-min-length": [
+          0
+      ],
+      "type-case": [
+          0
+      ],
+      "type-empty": [
+          0
+      ],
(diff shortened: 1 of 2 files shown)


janisozaur:
Finally, sane commit messages for the masses!

Can't check it properly now, but some notes:
* This should be applied to all commits in a PR
* What does "warning" mean? Nobody reads those, especially when they are hidden in CI logs
* I'm not sure 20 chars minimum is fine. If you listed current commits with subject line shorter than this, "Release v0.2.6" would not fly. On the other hand we could use our codenames there.

It will likely also not check the PR title, which usually ends up being the commit's subject when squashing.

tupaschoal:
> * This should be applied to all commits in a PR

It is, as per the description notes


> * What does "warning" mean? Nobody reads those, especially when they are hidden in CI logs

Means the CI job will not fail if only warnings are emitted. It's just warning for "too short" commit messages for now anyway, so no big deal.

> * I'm not sure 20 chars minimum is fine. If you listed current commits with subject line shorter than this, "Release v0.2.6" would not fly. On the other hand we could use our codenames there.

Yeah, we didn't have a guideline for that, so I just set a rather low limit, we can discuss on what it ought to be.

> It will likely also not check the PR title, which usually ends up being the commit's subject when squashing.

I'd say not, but then upon merging it will fail if the title is wrong, so we can ammend if we notice that

janisozaur:
We historically never amended commits once in the repo, I doubt this will change.

tupaschoal:
I don't like the idea too, but if the lint fails for long commit messages and the author fixes it, then we won't have a problem in squash either, right?

duncanspumpkin:
Can we test the squash merge message by doing a pr to your repo and seeing what would happen.

tupaschoal:
> Can we test the squash merge message by doing a pr to your repo and seeing what would happen.

Here it is https://github.com/tupaschoal/OpenRCT2/runs/613976742

tupaschoal:
Hey @duncanspumpkin and @janisozaur , can we merge this? I've changed it to error out on any commit messages with `len < 10` or `len > 72`.

This would make the `Release vX.X.X` not fail, we can increase it later. Do note that I've made the squash test above and that we'd still need to squash this one on merge

Issue #11389, opened 2020-04-21
Refactor string formatting framework


brenoguim:
This is an issue to document all the changes for a more robust string formatting framework.

- [x] Introduce a `Formatter` class to serve as a stream that receives arguments. #11350 
- [x] Use the `Formatter` to replace `set_format_arg_on`. The `Formatter` class does the argument counting which simplifies the caller code quite a bit. #11350 
- [x] Introduce `Formatter::MapTooltip` to get an instance that works on the global buffer for formatting tooltips. Replace the usage of `set_map_tooltip_format_arg` #11379 
- [ ] Replace usage of  `set_format_arg` by  `Formatter::Common()` which works on the global buffer used for common argument formatting. This will take several PRs since there are too many callers.
- [ ] Remove all `Format*To` that takes overload

Ideas for investigation:
- Reduce the amount of calls of `Formatter::MapTooltip()` and `Formatter::Common()` by passing them through parameters. The code that wants to print a string, should be the code that creates the `Formatter` and pass it down for everyone to fill it.
- If it is possible to do the above, then it should be possible to store the number of arguments stored in the Formatter and cross-check that with the pattern for the string print.

IntelOrca:
Thanks for picking this up. One of our end goals is to have only numbered arguments rather than PUSH16, POP16 etc. in the string files.

duncanspumpkin:
The scrolling text uses its own array of formating arguments and the information is copied from the common format args into the scrolling text formatting arguments array. To prevent excessive amounts of scroll text work the game performs memcpr's of the scrolling text formatting arguments to see if the formatting arguments already exist and if it needs to add another one. This means that the whole 16? bytes of the formatting arguments must match (leading to odd pushes of zero prior to calling a scroll text function). It would be much better if scrolling text took a format args and this was copy constructed with any left over buffer space filled with zero.

tupaschoal:
> Replace usage of  `set_format_arg` by  `Formatter::Common()` which works on the global buffer used for common argument formatting. This will take several PRs since there are too many callers. 
@brenoguim I took the liberty of ticking this one and editing with the PRs that you and @Gymnasiast did to finish it.

tupaschoal:
Did another edit to track all `Format*To` PRs

frutiemax:
I feel like I really polluted the history log with all those commits on Ride FormatTo. Is there a way to avoid it?

tupaschoal:
I don't think there is @frutiemax

[...]

PR #11411, opened 2020-04-22
Let Ride::FormatStatusTo return the number of written bytes

brenoguim:
This will be useful later on as Formatter spreads throughout the code. It's for the caller code to be able to trust the "Formatter" is correctly updated when passed to a function.


tupaschoal:
Part of #11389

PR #11412, opened 2020-04-23
Replace usage of set_format_arg by Formatter::Common()

brenoguim:
Part of #11389

### DIFF
--- a/src/openrct2/paint/tile_element/Paint.Entrance.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Entrance.cpp
@@ -165,16 +165,17 @@ static void ride_entrance_exit_paint(paint_session* session, uint8_t direction, 
     if (!is_exit && !(tile_element->IsGhost()) && tile_element->AsEntrance()->GetRideIndex() != RIDE_ID_NULL
         && stationObj->ScrollingMode != SCROLLING_MODE_NONE)
     {
-        set_format_arg(0, rct_string_id, STR_RIDE_ENTRANCE_NAME);
-        set_format_arg(4, uint32_t, 0);
+        auto ft = Formatter::Common();
+        ft.Add<rct_string_id>(STR_RIDE_ENTRANCE_NAME);
+        ft.Add<uint32_t>(0);
 
         if (ride->status == RIDE_STATUS_OPEN && !(ride->lifecycle_flags & RIDE_LIFECYCLE_BROKEN_DOWN))
         {
-            ride->FormatNameTo(gCommonFormatArgs + 2);
+            ride->FormatNameTo(ft);
         }
         else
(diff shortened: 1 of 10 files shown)


src/openrct2/paint/tile_element/Paint.Entrance.cpp
```diff
@@ -270,20 +271,20 @@ static void park_entrance_paint(paint_session* session, uint8_t direction, int32
                 break;
 
             {
-                set_format_arg(0, uint32_t, 0);
-                set_format_arg(4, uint32_t, 0);
+                Formatter::Common().Add<uint32_t>(0).Add<uint32_t>(0);
```
tupaschoal, line 274: Should this get the same comment as #11388 ?

src/openrct2/paint/tile_element/Paint.LargeScenery.cpp
```diff
@@ -301,8 +301,7 @@ void large_scenery_paint(paint_session* session, uint8_t direction, uint16_t hei
         }
         // 6B8331:
         // Draw sign text:
-        set_format_arg(0, uint32_t, 0);
-        set_format_arg(4, uint32_t, 0);
+        Formatter::Common().Add<uint32_t>(0).Add<uint32_t>(0);
```
tupaschoal, line 304: Ditto

duncanspumpkin:
For your next pull request could you look into makeing a deep copy function for the formatter. We will need it for future refactorings especially with refactoring the scrolling text.

Gymnasiast:
ping @brenoguim

src/openrct2/paint/tile_element/Paint.Entrance.cpp
brenoguim: I wouldn't be able to tell. I don't know if the arguments are cleared for the same reason.

src/openrct2/paint/tile_element/Paint.LargeScenery.cpp
brenoguim: I wouldn't be able to tell. I don't know if the arguments are cleared for the same reason.

brenoguim:
> For your next pull request could you look into makeing a deep copy function for the formatter. We will need it for future refactorings especially with refactoring the scrolling text.

Hi @duncanspumpkin , the `Formatter` today doesn't own the buffer on which it writes to. The deep copy would imply in owning the buffer, right? How would that work?
Maybe I can't understand it because I didn't get to the code that needs that functionality yet.

tupaschoal:
Since it's for a follow-up PR, I'm assuming you can take the conversation to the parent issue @duncanspumpkin and @brenoguim

PR #11416, opened 2020-04-23
Fix #11405: Building path through walls does not always remove them

### LINKED ISSUE #11405: Building a path through walls does not always remove the walls
**OS:** Ubuntu 20.04
**Version:** 0.2.6+
**Commit/Build:** cfd94d4fa54525b3402093d2dc1a83dcb58dde49

afbeelding

afbeelding

Gymnasiast:
"Parentheses are maths' condoms. You can never use too many of them, but you can use too few." - My maths teacher in secondary school.


Gymnasiast:
@janisozaur It got a bit lost in translation. What he said in Dutch would perhaps better have been translated as "too often", rather than "too many".

PR #11418, opened 2020-04-23
Fix #11407: Setting an RCT1 path needs better error messages

### LINKED ISSUE #11407: Setting an RCT1 path needs better error messages
Setting an RCT1 path only fully works if (1) csg1i.dat is present. (2) it's a Loopy Landscapes install. When setting an RCT1 path, these points are not properly checked and warned for, leaving the user to wonder why it doesn't work.

So, when setting the RCT1 path, it should work like this:
- Check if csg1i.dat is present. If not, display a warning telling users to copy it from the CD to the RCT1 install folder and abort.
- Check if the supplied csg1.dat/csg1.1 is from Loopy Landscapes(/RCT Deluxe). If not, display a warning telling users this functionality only works with Loopy Landscapes or RCT Deluxe.

Gymnasiast:


src/openrct2/config/Config.cpp
```diff
@@ -837,3 +839,83 @@ bool config_find_or_browse_install_directory()
...
+}
+
+bool RCT1DataPresentAtLocation(const utf8* path)
+{
+    return Csg1datPresentAtLocation(path) && Csg1idatPresentAtLocation(path) && CsgAtLocationIsUsable(path);
+}
+
+bool CsgIsUsable(rct_gx csg)
+{
+    return csg.header.num_entries >= RCT1_NUM_LL_CSG_ENTRIES;
```
Gymnasiast, line 895:
```suggestion
    return csg.header.num_entries == RCT1_NUM_LL_CSG_ENTRIES;
```

Exactly 69917 slots are reserved, so this needs to be exact.

Umdlye:
May be a bit far-fetched but would it be possible to search the entire `\Hasbro Interactive\RollerCoaster Tycoon` folder for the CD files? I followed [this guide](https://www.nedesigns.com/topic/5781/rct-ll-no-cd-crack-works-for-xp-and-vista/?p=106504) to be able to launch LL without the CD so I have the CD files in `\RollerCoaster Tycoon\CD\Data`. I'm probably not the only one so that might save some people a bit of trouble.

Gymnasiast:
I think the overhead from doing so might get too big. At least you'll now get a nice warning saying the csg1i.dat file is missing from the install, and that you should copy it off the CD.

src/openrct2/config/Config.h
```diff
@@ -245,3 +246,11 @@ void config_set_defaults();
 void config_release();
 bool config_save_default();
 bool config_find_or_browse_install_directory();
+
+bool RCT1DataPresentAtLocation(const utf8* path);
+std::string FindCsg1datAtLocation(const utf8* path);
+bool Csg1datPresentAtLocation(const utf8* path);
+std::string FindCsg1idatAtLocation(const utf8* path);
+bool Csg1idatPresentAtLocation(const utf8* path);
```
tupaschoal, line 254: What's the `i` in `Csg1idatPresentAtLocation` that differs it from `Csg1datPresentAtLocation`? I found the subtle difference very confusing.

Gymnasiast: CSG1i.DAT is the file that contains the indexes for CSG1.DAT. (RCT2 did away with this split, having both the index table and actual sprites in one file.)

tupaschoal: Ah, ok

PR #11419, opened 2020-04-23
Fix RCT1 import importing too many vehicle types

Gymnasiast:


Gymnasiast:
Never try to understand RCT1 research. You'll go mad.

PR #11433, opened 2020-04-24
Android cleanup

FlukiestEmperor:
The ZipArchive class is also unused, but I left in it because I don't know the original intent for it and it may be needed in the future.

### DIFF
--- a/src/openrct2-android/app/src/main/java/io/openrct2/MainActivity.java
+++ b/src/openrct2-android/app/src/main/java/io/openrct2/MainActivity.java
@@ -28,8 +28,6 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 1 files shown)


src/openrct2-android/app/src/main/java/io/openrct2/MainActivity.java
```diff
@@ -65,34 +68,8 @@ private PointF getResolutionDips() {
 
         DisplayMetrics metrics = new DisplayMetrics();
         display.getMetrics(metrics);
-
-        boolean success = false;
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
```
janisozaur, line 70: which version do we have as minimum now?

FlukiestEmperor: We have the minimum set to 19, JELLY_BEAN_MR1 is 17

Issue #11426, opened 2020-04-23
Extra Viewport Enhancements


Richard-L:
> OpenRCT2, v0.2.6-146-g619b06254 (619b062 on develop) provided by GitHub
> Windows 7 (x86-64)

![image](https://user-images.githubusercontent.com/4934209/80159194-ab2d9280-85ca-11ea-872b-715582e0d119.png)

This is an amazing feature.

With content creators in mind, other than the already implemented zoom level, these should have all camera parameters independently available:

- [x] Zoom Level
- [ ] Camera Rotation
- [ ] Pixel Scaler

I would put the `Locate on map` button first. Next I would offer a `Sync with main view` toggle, followed by "Zoom Level", "Rotation" and "Pixel Scaler". If syncing is enabled, all three are greyed out. If syncing is disabled, they are available.

![image](https://user-images.githubusercontent.com/4934209/80160051-eb8e1000-85cc-11ea-959d-d1b01a489b64.png)
Further, I would suggest moving the `Extra Viewport` item from the current `Map` menu to `View options` (to the left), or create a new icon altogether with the camera one we know from ride windows.

Thank you

duncanspumpkin:
Pixel scaler can not be done for singular parts for now. If we could do that we would scale the ui differently.

AaronVanGeffen:
I think the View menu is way too full already. I'd prefer to split up the more advanced view options instead.

Richard-L:
@AaronVanGeffen I do agree actually. Hence the suggestion to perhaps use a dedicated button with the camera icon or similar.

Perhaps this is something to fall under menu buttons you can show/hide, like the mute button.

PR #11446, opened 2020-04-24
Create toggle for zoom toolbar button

Umdlye:
They're on by default.

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3664,7 +3664,7 @@ STR_6407    :This may take a few minutes.
 STR_6408    :Please install “innoextract” to extract GOG Installer, then restart OpenRCT2.
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
 STR_6410    :Zoom in/out
-STR_6411    :{SMALLFONT}{BLACK}Show separate buttons for the zoom options in the toolbar
+STR_6411    :Show buttons for zooming in and out in the toolbar
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3663,6 +3663,8 @@ STR_6406    :GOG RollerCoaster Tycoon 2 Installer
 STR_6407    :This may take a few minutes.
 STR_6408    :Please install “innoextract” to extract GOG Installer, then restart OpenRCT2.
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
+STR_6410    :Zoom in/out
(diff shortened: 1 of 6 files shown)


Gymnasiast:
What is the aim of this PR? Is it to conserve screen space? If so, we should perhaps have a broader discussion of that instead of adding toggles piecemeal.

IntelOrca:
We have been conservative about having too many settings, however maybe we should now allow more things to be configurable but now allow UI. Instead plugins could provide the UI to toggle the settings.

Gymnasiast:
And/or have a setting for "compact UI", that makes the layout more suitable for smaller screens.

Umdlye:
I like the idea about configuring UI through plug-ins, just figured this particular toggle would have a wide enough appeal to warrant inclusion in the settings. The aim is to save screen space (especially considering potentially moving some view options into a new toolbar menu was brought up in #11426) and eliminate redundant buttons (I believe most people change their zoom with the scroll wheel).

I asked if it was worth a PR on Discord and got a positive team member response so I went ahead with it.

IntelOrca:
I expect some people would want to also hide the rotate button, but maybe once I implement the ability to bind mouse button 3 and 4 to a shortcut (in this case rotate).

tupaschoal:
@IntelOrca how does this fare now that we have plugins? (Of course it needs to be rebased)

IntelOrca:
There is an API for changing settings but not many settings have been added yet. It is easy enough to do.
There are no APIs for altering the toolbars, that would have to be added.

Gymnasiast:
I have discussed this in the team chat and decided that I will merge it as-is. Could you rebase it?

Umdlye:
I'll try!

Umdlye:
Yay, it worked.

data/language/en-GB.txt
```diff
@@ -3663,6 +3663,8 @@ STR_6406    :GOG RollerCoaster Tycoon 2 Installer
 STR_6407    :This may take a few minutes.
 STR_6408    :Please install “innoextract” to extract GOG Installer, then restart OpenRCT2.
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
+STR_6410    :Zoom in/out
+STR_6411    :{SMALLFONT}{BLACK}Show separate buttons for the zoom options in the toolbar
```
Gymnasiast, line 3667:
```suggestion
STR_6411    :{SMALLFONT}{BLACK}Show buttons for zooming in and out in the toolbar
```

"Separate" implies that turning the option off will merge the buttons into one, rather than hide them.

Umdlye: Ah, right. I just matched the wording of the other tooltips in that group when I first created this, hadn't considered that yet.
Can I also drop `{SMALLFONT}{BLACK}` from the string or does that require additional work? (looking at \#13560)

Gymnasiast: Yeah, you can drop that. Good point, forgot about that.

PR #11480, opened 2020-04-26
Vehicle update 9473 02

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -6393,26 +6396,26 @@ int32_t Vehicle::UpdateMotionDodgems()
  *
  *  rct2: 0x006DD365
  */
-bool vehicle_update_dodgems_collision(Vehicle* vehicle, int16_t x, int16_t y, uint16_t* spriteId)
+bool Vehicle::UpdateDodgemsCollision(int16_t curX, int16_t curY, uint16_t* spriteId)
 {
-    uint16_t bp = (vehicle->var_44 * 30) >> 9;
-    uint32_t trackType = vehicle->track_type >> 2;
+    uint16_t bp = (var_44 * 30) >> 9;
```
duncanspumpkin, line 6401: bp is the dodgems vehicles radius. Could you rename the variable to something like that.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6393,26 +6396,26 @@ int32_t Vehicle::UpdateMotionDodgems()
  *
  *  rct2: 0x006DD365
  */
-bool vehicle_update_dodgems_collision(Vehicle* vehicle, int16_t x, int16_t y, uint16_t* spriteId)
+bool Vehicle::UpdateDodgemsCollision(int16_t curX, int16_t curY, uint16_t* spriteId)
```
duncanspumpkin, line 6399: another refactoring project is replacing these x and y variables with the respective structs. In this instance this is a `CoordsXY`. Can you replace x and y with a `CoordsXY`. All calls can construct this with `{x ,y}`

hdpoliveira: During this I actually noticed this does not Update the vehicle in any way, so I renamed it, and added a const to it.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6393,26 +6396,32 @@ int32_t Vehicle::UpdateMotionDodgems()
  *
  *  rct2: 0x006DD365
  */
-bool vehicle_update_dodgems_collision(Vehicle* vehicle, int16_t x, int16_t y, uint16_t* spriteId)
+static bool wouldCollideWithDodgemsTrackEdge(
+    CoordsXY coords, CoordsXY trackLocation, uint32_t trackType, uint16_t dodgemsCarRadius)
```
tupaschoal, line 6400:
```suggestion
    const CoordsXY& coords, cons CoordsXY& trackLocation, uint32_t trackType, uint16_t dodgemsCarRadius)
```
(formatting will probably need to be revisited as well)

src/openrct2/ride/Vehicle.h
```diff
@@ -305,7 +305,11 @@ struct Vehicle : SpriteBase
     void SetState(VEHICLE_STATUS vehicleStatus, uint8_t subState = 0);
     bool IsGhost() const;
     void UpdateSoundParams(std::vector<rct_vehicle_sound_params>& vehicleSoundParamsList) const;
+    bool DodgemsCarWouldCollideAt(CoordsXY coords, uint16_t* spriteId) const;
```
tupaschoal, line 308:
```suggestion
    bool DodgemsCarWouldCollideAt(const CoordsXY& coords, uint16_t* spriteId) const;
```

PR #11486, opened 2020-04-26
Implement Train class with iterators

hdpoliveira:
This change creates a Train class with iterators. It also creates a Mass function, replacing the get_train_mass function in Vehicle.cpp

For now this is on a unnamed namespace in Vehicle.cpp, but we may consider extract it to its own file.

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -815,18 +815,87 @@ void Vehicle::Invalidate()
     invalidate_sprite_2(this);
 }
 
-static int32_t get_train_mass(const Vehicle* first_vehicle)
+namespace
+{
+    template<typename T> class TrainIterator;
+    template<typename T> class Train
+    {
+    public:
+        explicit Train(T* vehicle)
+            : FirstCar(vehicle)
+        {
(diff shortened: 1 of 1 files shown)


IntelOrca:
I do like this. Is the idea for this to always be a temporary throw-away struct? Do you plan or using this for other pieces of logic and adding more helper methods?

hdpoliveira:
For now the idea is indeed for this to be a throw away struct. I noticed
many methods in Vehicle.cpp and Ride.cpp actually do calculation on the
entire train instead of only in a single Vehicle, so these functions could
become methods in the Train class.
For now I've put the Mass function in there just to illustrate, but this
could later be its own file.

duncanspumpkin:
The first car should always be a train head. Should we have extra protection making sure this is the case during construction.

hdpoliveira:
@duncanspumpkin just adding 
```cpp
            assert(vehicle->TrainHead() == FirstCar);
```
in the constructor is enough? (given https://github.com/OpenRCT2/OpenRCT2/pull/11480 is merged for TrainHead method)

duncanspumpkin:
IsHead is much cheaper to perform.

PR #11489, opened 2020-04-27
Implement SmallSceneryObject for getObject plugin API

IntelOrca:
```
openrct2 $ context.getObject("small_scenery", 0)
{ flags: 268435457,
  height: 80,
  price: 18,
  removalPrice: 10,
  type: 'small_scenery',
  index: 0,
  identifier: 'TAP     ',
  name: 'Aleppo Pine Tree' }
```

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -119,6 +119,7 @@ declare global {
          */
         getObject(type: ObjectType, index: number): Object;
         getObject(type: "ride", index: number): RideObject;
+        getObject(type: "small_scenery", index: number): SmallSceneryObject;
 
         getAllObjects(type: ObjectType): Object[];
         getAllObjects(type: "ride"): RideObject[];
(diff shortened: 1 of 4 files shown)


Gymnasiast:
This uses the old .DAT identifier, even for JSON objects. It would be preferable to use our own (JSON) identifier here, I think.

IntelOrca:
What about custom objects?

Gymnasiast:
Not sure. Perhaps they should be `custom.<OLDDATIDENTIER>`. Or perhaps we should specify both identifiers, where current JSON objects fill both fields, old .DAT objects fill only the old one and new JSON objects that don't have a .DAT header only fill the new one.

Or perhaps we should just use the JSON identifier, falling back to the .DAT identifier if it isn't a JSON object.

IntelOrca:
Okay, will add `legacyIdentifier` and `identifier` in another PR.

PR #11493, opened 2020-04-27
Enable -Wold-style-cast

janisozaur:
Now that casts have been changed to new style, prevent adding more
old-style cast


janisozaur:
I expect some fails in CI…

Gymnasiast:
I have rebased it and also cleaned up most remaining C-style casts. One remains, in `Memory.hpp`:
```
    template<typename T> static void Free(T* ptr)
    {
        free((void*)ptr);
    }
```

What should I do with it?

duncanspumpkin:
Can you reinterpret cast it?

Gymnasiast:
Ah, reinterpret casting to const void* and then const casting to void* seems to do the trick! :+1:

src/openrct2/drawing/TTFSDLPort.cpp
```diff
@@ -1,6 +1,10 @@
 // Adapted from freetype.h in order to avoid C-style casts.
 
 #define FT_LOAD_TARGET_ALT(x) (static_cast<FT_Int32>((x)&15) << 16)
+#define FT_IMAGE_TAG(value, _x1, _x2, _x3, _x4)                                                                                \
+    value                                                                                                                      \
+        = ((static_cast<unsigned long>(_x1) << 24) | (static_cast<unsigned long>(_x2) << 16)                                   \
+           | (static_cast<unsigned long>(_x3) << 8) | static_cast<unsigned long>(_x4))
```
tupaschoal, line 7: I didn't look this through, but because this is a macro, the callers might be sending things that aren't really `unsigned_long`, I think if we want to be safe, it should be a template instead

(Where is it coming from anyway? Because there are no red lines, so it seems like you're introducing this macro)

Gymnasiast: This is a macro that is deep in the guts of the Freetype library. We don't even call it ourselves. But since GCC seems to make no distinction between our own headers and those of dependencies, I _have_ to define it. For this reason, I really can't change how it works.

tupaschoal:
I've cleaned up for the PR to build properly on Windows and it's now green.

The two CIs still failing are doing so on:
- testpaint (which leads me to believe https://github.com/OpenRCT2/OpenRCT2/pull/11493/commits/a57397e267ffdd93611d4c796265285bdbbe97cc is not working)
- freetype include (we shouldn't apply compilation flags to dependencies)

Not particularly savvy of cmake, so if you can solve those out @janisozaur , this looks mostly ready to push

PR #11505, opened 2020-04-27
Refactor ride_list_item to RideSelection

Gymnasiast:
This is one extra place where it would sneakily hide two `uint8_t`s, one of them being the entry index, in a `uint16_t`. I gave it the same treatment as ScenerySelection.

### DIFF
--- a/src/openrct2-ui/windows/NewRide.cpp
+++ b/src/openrct2-ui/windows/NewRide.cpp
@@ -39,8 +39,8 @@ constexpr int32_t WH = 382;
 constexpr int32_t WW = 601;
 
 static uint8_t _windowNewRideCurrentTab;
-static ride_list_item _windowNewRideHighlightedItem[6];
-static ride_list_item _windowNewRideListItems[384];
+static RideSelection _windowNewRideHighlightedItem[6];
+static RideSelection _windowNewRideListItems[384];
 
 #pragma region Ride type view order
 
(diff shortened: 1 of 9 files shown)


src/openrct2/rct12/RCT12.cpp
```diff
@@ -1010,3 +1010,19 @@ bool RCT12ResearchItem::IsRandomEndMarker() const
 {
     return rawValue == RCT12_RESEARCHED_ITEMS_END_2;
 }
+
+ObjectEntryIndex RCTEntryIndexToOpenRCT2EntryIndex(RCT12ObjectEntryIndex index)
```
Gymnasiast, line 1014: Should not have been committed :/

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -445,12 +445,12 @@ static void window_new_ride_scroll_to_focused_ride(rct_window* w)
...
-    int32_t focusRideType = _windowNewRideHighlightedItem[_windowNewRideCurrentTab].ride_type_and_entry;
+    auto focusRideType = _windowNewRideHighlightedItem[_windowNewRideCurrentTab];
     int32_t count = 0, row = 0;
-    ride_list_item* listItem = _windowNewRideListItems;
-    while (listItem->type != RIDE_TYPE_NULL || listItem->entry_index != RIDE_ENTRY_INDEX_NULL)
+    RideSelection* listItem = _windowNewRideListItems;
+    while (listItem->Type != RIDE_TYPE_NULL || listItem->EntryIndex != RIDE_ENTRY_INDEX_NULL)
     {
-        if (listItem->type == focusRideType)
+        if (listItem->Type == focusRideType.Type && listItem->EntryIndex == 0)
```
Gymnasiast, line 453: This is what was done before. It could be that 

```suggestion
        if (listItem->Type == focusRideType.Type)
```
was intended.

src/openrct2/ride/RideTypes.h
```diff
@@ -17,3 +19,22 @@ struct Ride;
...
 constexpr const ride_id_t RIDE_ID_NULL = 0xFF; // std::numeric_limits<ride_id_t>::max();
 constexpr const ride_id_t RideIdNewNull = std::numeric_limits<ride_id_t>::max();
+
+/**
+ * Couples a ride type and subtype together.
+ */
+struct RideSelection
+{
+    uint8_t Type;
+    ObjectEntryIndex EntryIndex;
```
IntelOrca, line 29:
```suggestion
    ObjectEntryIndex EntryIndex = RIDE_ENTRY_INDEX_NULL;
```

src/openrct2/ride/RideTypes.h
```diff
@@ -17,3 +19,22 @@ struct Ride;
 
 constexpr const ride_id_t RIDE_ID_NULL = 0xFF; // std::numeric_limits<ride_id_t>::max();
 constexpr const ride_id_t RideIdNewNull = std::numeric_limits<ride_id_t>::max();
+
+/**
+ * Couples a ride type and subtype together.
+ */
+struct RideSelection
+{
+    uint8_t Type;
```
IntelOrca, line 28:
```suggestion
    uint8_t Type = RIDE_TYPE_NULL;
```

src/openrct2/ride/RideTypes.h
IntelOrca: Struct is used in union, so nevermind.

src/openrct2/ride/RideTypes.h
IntelOrca: Struct is used in union, so nevermind.

PR #11507, opened 2020-04-27

Gymnasiast:


src/openrct2/object/WallObject.cpp
```diff
@@ -25,7 +25,8 @@ void WallObject::ReadLegacy(IReadObjectContext* context, IStream* stream)
     _legacyType.wall.height = stream->ReadValue<uint8_t>();
     _legacyType.wall.flags2 = stream->ReadValue<uint8_t>();
     _legacyType.wall.price = stream->ReadValue<uint16_t>();
-    _legacyType.wall.scenery_tab_id = stream->ReadValue<uint8_t>();
+    _legacyType.wall.scenery_tab_id = OBJECT_ENTRY_INDEX_NULL;
+    stream->Seek(1, STREAM_SEEK_CURRENT);
```
tupaschoal, line 29: Does it need the `seek` ? It didn't have one before, like all the other places above (e. g. `FootpathItemObject.cpp`). Theseo other places also updated it to 2

src/openrct2/management/Research.h
```diff
@@ -19,32 +19,72 @@ struct rct_ride_entry;
...
             uint8_t type; // 0: scenery entry, 1: ride entry
-            uint8_t flags;
         };
     };
+    uint8_t flags;
     uint8_t category;
 
     bool IsNull() const;
+    void SetNull();
     bool Equals(const ResearchItem* otherItem) const;
```
IntelOrca, line 37: Is now a good time to replace `IsNull`, `SetNull`, `Equals` with operator overloads?

`if (item == ResearchItem::Null)` or `if (item == ResearchItem())` or `if (item == {})`
`item = {}` or `item = ResearchItem()`
`if (item == item2)`

src/openrct2/object/WallObject.cpp
duncanspumpkin: Yes it does. We are doing the entry. The others were also slipping another entry.

src/openrct2/world/Scenery.h
```diff
@@ -83,19 +80,16 @@ enum LARGE_SCENERY_TEXT_FLAGS
...
-    uint8_t scrolling_mode;        // 0x11
-    rct_large_scenery_text* text;  // 0x12
-    uint32_t text_image;           // 0x16
+    uint8_t tool_id;                 // 0x06
+    uint8_t flags;                   // 0x07
+    int16_t price;                   // 0x08
+    int16_t removal_price;           // 0x0A
+    rct_large_scenery_tile* tiles;   // 0x0C
+    ObjectEntryIndex scenery_tab_id; // 0x10
+    uint8_t scrolling_mode;          // 0x11
```
duncanspumpkin, line 89: Remove these comments they are no longer correct.

src/openrct2/world/Scenery.h
```diff
@@ -108,15 +102,14 @@ enum LARGE_SCENERY_FLAGS
...
-    uint8_t flags2;         // 0x09
-    int16_t price;          // 0x0A
-    uint8_t scenery_tab_id; // 0x0C
-    uint8_t scrolling_mode; // 0x0D 0xFF if no scrolling
+    uint8_t tool_id;                 // 0x06
+    uint8_t flags;                   // 0x07
+    uint8_t height;                  // 0x08
+    uint8_t flags2;                  // 0x09
+    int16_t price;                   // 0x0A
+    ObjectEntryIndex scenery_tab_id; // 0x0C
```
duncanspumpkin, line 110: Remove comments

Gymnasiast:
> What is still using the research item raw value? Can it be refactored to not use it?

Only news items and their related functions. These take a `uint32_t`. I have taken out all other occurrences.

Gymnasiast:
@duncanspumpkin This seems to break the replaytests. Can you help me address that?

duncanspumpkin:
Will need to recreate only FootpathPlaceAction.sv6r as far as i can tell. I will do that at the same time as  #11573

Gymnasiast:
Merged along with https://github.com/OpenRCT2/OpenRCT2/pull/11573.

PR #11509, opened 2020-04-27
Entity cleanup

duncanspumpkin:
A couple little refactors that are from my create sprite branch. These changes remove the ability for entities to change their list type after creation and also reduces the amount of code that actually uses the list type.

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -5482,7 +5482,7 @@ void Guest::UpdateWalking()
     {
         sprite = get_sprite(sprite_id);
 
-        if (sprite->generic.sprite_identifier != SPRITE_IDENTIFIER_PEEP)
+        if (!sprite->IsPeep())
             continue;
 
         if (sprite->peep.state != PEEP_STATE_WATCHING)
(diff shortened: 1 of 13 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -5482,7 +5482,7 @@ void Guest::UpdateWalking()
     {
         sprite = get_sprite(sprite_id);
 
-        if (sprite->generic.linked_list_index != SPRITE_LIST_PEEP)
+        if (sprite->generic.sprite_identifier != SPRITE_IDENTIFIER_PEEP)
```
Gymnasiast, line 5485: It would be neater if we used the `IsPeep()` function here.

duncanspumpkin: I'm about to replace all of this code but yes I can change that.

src/openrct2/world/Sprite.cpp
```diff
@@ -884,11 +889,10 @@ uint16_t remove_floating_sprites()
...
+    switch (sprite->generic.sprite_identifier)
     {
-        case SPRITE_LIST_PEEP:
-        case SPRITE_LIST_TRAIN_HEAD:
-        case SPRITE_LIST_VEHICLE:
+        case SPRITE_IDENTIFIER_PEEP:
+        case SPRITE_IDENTIFIER_VEHICLE:
             return true;
     }
     return false;
```
Gymnasiast, line 898: This function could also do `return sprite->IsPeep() || sprite->IsVehicle()`.

duncanspumpkin: we don't have an `isVehicle()` function yet. It will be changed as part of the is / as refactor.

PR #11511, opened 2020-04-27
Improve network plugin APIs

IntelOrca:
* Add `network.addGroup`, `network.removeGroup`.
* Add `network.numGroups` and `network.numPlayers`.
* Change `network.groups` and `network.players` to return a list.
* Add `player.ipAddress` and `player.publicKeyHash`.
* Call network.chat hook for server's chat messages.
* Implement `network.authenticate`, `network.join` and `network.leave` hooks.
* Implement `network.sendMessage(msg, players)`.

### DIFF
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -672,8 +672,8 @@ NetworkConnection* Network::GetPlayerConnection(uint8_t id)
     if (player != nullptr)
     {
         auto clientIt = std::find_if(
-                   client_connection_list.begin(), client_connection_list.end(),
-                   [player](const auto& conn) -> bool { return conn->Player == player; });
+            client_connection_list.begin(), client_connection_list.end(),
+            [player](const auto& conn) -> bool { return conn->Player == player; });
         return clientIt != client_connection_list.end() ? clientIt->get() : nullptr;
     }
     return nullptr;
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -1676,6 +1676,7 @@ void Network::Server_Send_CHAT(const char* text, const std::vector<uint8_t>& pla
 
     if (playerIds.empty())
     {
+        // Empty players / default value means send to all players
         SendPacketToClients(*packet);
     }
     else
@@ -3522,7 +3523,7 @@ money32 network_get_player_money_spent(uint32_t index)
 std::string network_get_player_ip_address(uint32_t id)
 {
     auto conn = gNetwork.GetPlayerConnection(id);
-    if (conn != nullptr)
+    if (conn != nullptr && conn->Socket != nullptr)
     {
         return conn->Socket->GetIpAddress();
     }
@@ -4002,6 +4003,7 @@ void network_send_chat(const char* text, const std::vector<uint8_t>& playerIds)
                 if (playerIds.empty()
                     || std::find(playerIds.begin(), playerIds.end(), gNetwork.GetPlayerID()) != playerIds.end())
                 {
+                    // Server is one of the recipients
                     chat_history_add(formatted);
                 }
                 gNetwork.Server_Send_CHAT(formatted, playerIds);
--- a/src/openrct2/network/Network.cpp
+++ b/src/openrct2/network/Network.cpp
@@ -671,10 +671,10 @@ NetworkConnection* Network::GetPlayerConnection(uint8_t id)
     auto player = GetPlayerByID(id);
     if (player != nullptr)
     {
-        return std::find_if(
+        auto clientIt = std::find_if(
                    client_connection_list.begin(), client_connection_list.end(),
-                   [player](const auto& conn) -> bool { return conn->Player == player; })
-            ->get();
+                   [player](const auto& conn) -> bool { return conn->Player == player; });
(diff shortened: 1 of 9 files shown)


src/openrct2/network/Network.cpp
```diff
@@ -1942,6 +1974,80 @@ void Network::ProcessPending()
...
+        eObj.Set("name", name);
+        eObj.Set("publicKeyHash", publicKeyHash);
+        eObj.Set("ipAddress", connection.Socket->GetIpAddress());
+        eObj.Set("cancel", false);
+        auto e = eObj.Take();
+
+        // Call the subscriptions
+        hookEngine.Call(OpenRCT2::Scripting::HOOK_TYPE::NETWORK_AUTHENTICATE, e, false);
+
+        // Check if any hook has cancelled the join
```
tupaschoal, line 1999: Does this mean if the user started authentication and left before actually doing so?

src/openrct2/network/Network.cpp
```diff
@@ -1993,19 +2103,35 @@ void Network::ProcessPlayerList()
             }
 
             // Remove any players that are not in newly received list
-            auto it = player_list.begin();
-            while (it != player_list.end())
+            for (const auto& player : player_list)
             {
-                if (std::find(activePlayerIds.begin(), activePlayerIds.end(), (*it)->Id) == activePlayerIds.end())
+                if (std::find(activePlayerIds.begin(), activePlayerIds.end(), player->Id) == activePlayerIds.end())
```
tupaschoal, line 2108: Shouldn't this be != end?

src/openrct2/network/Network.cpp
```diff
@@ -1654,12 +1671,27 @@ void Network::Client_Send_CHAT(const char* text)
...
+void Network::Server_Send_CHAT(const char* text, const std::vector<uint8_t>& playerIds)
 {
     std::unique_ptr<NetworkPacket> packet(NetworkPacket::Allocate());
     *packet << static_cast<uint32_t>(NETWORK_COMMAND_CHAT);
     packet->WriteString(text);
-    SendPacketToClients(*packet);
+
+    if (playerIds.empty())
+    {
+        SendPacketToClients(*packet);
```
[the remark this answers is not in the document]
IntelOrca: An empty list means send to all clients.

src/openrct2/network/Network.cpp
IntelOrca: No any of the hook event handlers can deny the user from entry. The same object is used for each event handler and they can fight over the cancel boolean amongst themselves. We then take the final result of cancel and if it is true, we deny the user from entry into the server. This is before they have "official" joined and been given a player id.

src/openrct2/network/Network.cpp
IntelOrca: No, we want to find all the players which are no longer in the server.

src/openrct2/network/Network.cpp
```diff
@@ -665,6 +666,19 @@ uint8_t Network::GetPlayerID()
...
 
+NetworkConnection* Network::GetPlayerConnection(uint8_t id)
+{
+    auto player = GetPlayerByID(id);
+    if (player != nullptr)
+    {
+        return std::find_if(
+                   client_connection_list.begin(), client_connection_list.end(),
+                   [player](const auto& conn) -> bool { return conn->Player == player; })
+            ->get();
```
tupaschoal, line 677:
```suggestion
        auto clientIt = std::find_if(
                   client_connection_list.begin(), client_connection_list.end(),
                   [player](const auto& conn) -> bool { return conn->Player == player; });
        return clientIt != client_connection_list.end() ? clientIt->get() : nullptr;
```
If it doesn't find the player, it won't try to dereference `std::end`

src/openrct2/network/Network.cpp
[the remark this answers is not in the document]
tupaschoal: Unless this is documented elsewhere (and I'm just being näive), it could use a comment

src/openrct2/network/Network.cpp
tupaschoal: I see, is the concurrency expected or something we'll have to deal with later ?

src/openrct2/network/Network.cpp
```diff
@@ -3385,6 +3519,26 @@ money32 network_get_player_money_spent(uint32_t index)
     return gNetwork.player_list[index]->MoneySpent;
 }
 
+std::string network_get_player_ip_address(uint32_t id)
+{
+    auto conn = gNetwork.GetPlayerConnection(id);
+    if (conn != nullptr)
+    {
+        return conn->Socket->GetIpAddress();
```
tupaschoal, line 3527: Should we check if `Socket` is ever `nullptr`?

src/openrct2/network/Network.cpp
IntelOrca: It isn't concurrent, all very synchronous. Yes it is by design.

tupaschoal:
Seems that just the formatting needs some fixing

IntelOrca:
@tupaschoal looks like it is caused by your code change.

tupaschoal:
I did suggest it on mobile, so formatting might have not been ideal. I can fix it myself but it will be a lot later today

PR #11514, opened 2020-04-27
Implement expenditureType on GameActionResult plugin API

IntelOrca:


src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -771,6 +777,44 @@ DukValue ScriptEngine::PositionToDuk(const CoordsXYZ& position)
...
+                                                    "food_drink_stock",
+                                                    "wages",
+                                                    "marketing",
+                                                    "research",
+                                                    "interest" };
+
+std::string_view ScriptEngine::ExpenditureTypeToString(ExpenditureType expenditureType)
+{
+    auto index = static_cast<size_t>(expenditureType);
+    if (index >= 0 && index < std::size(ExpenditureTypes))
```
Gymnasiast, line 798: A `size_t` is always bigger than or equal to 0. This is also why the CIs fail.

IntelOrca: Sorry index was previously an `int32_t` when I wrote the condition.

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -771,6 +777,44 @@ DukValue ScriptEngine::PositionToDuk(const CoordsXYZ& position)
...
+                                                    "park_entrance_tickets",
+                                                    "park_ride_tickets",
+                                                    "shop_sales",
+                                                    "shop_stock",
+                                                    "food_drink_sales",
+                                                    "food_drink_stock",
+                                                    "wages",
+                                                    "marketing",
+                                                    "research",
+                                                    "interest" };
```
Gymnasiast, line 793: If you put an trailing comma and put "ride_construction" on the next line, will it stay in shape after clang-format? I find this looking quite ugly and hard to read.

IntelOrca: I can't get it to do what you are suggesting. Are you able to make change?

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -771,6 +777,44 @@ DukValue ScriptEngine::PositionToDuk(const CoordsXYZ& position)
...
+}
+
+ExpenditureType ScriptEngine::StringToExpenditureType(const std::string_view expenditureType)
+{
+    size_t i = 0;
+    for (auto sz : ExpenditureTypes)
+    {
+        if (expenditureType == sz)
+        {
+            return static_cast<ExpenditureType>(i);
```
Gymnasiast, line 812: You never increase `i`, so this will always return 0 if the expenditure type exists.

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -754,6 +754,12 @@ std::unique_ptr<GameActionResult> ScriptEngine::DukToGameActionResult(const DukV
     result->ErrorTitle = AsOrDefault<std::string>(d["errorTitle"]);
     result->ErrorMessage = AsOrDefault<std::string>(d["errorMessage"]);
     result->Cost = AsOrDefault<int32_t>(d["cost"]);
+
+    auto expenditureType = AsOrDefault<std::string>(d["expenditureType"]);
+    if (!expenditureType.empty())
+    {
+        result->Expenditure = StringToExpenditureType(expenditureType);
```
Gymnasiast, line 761: This does not check if StringToExpenditureType returns `ExpenditureType::Count`. (I wonder if it should return that in the first place.)

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -771,6 +777,46 @@ DukValue ScriptEngine::PositionToDuk(const CoordsXYZ& position)
...
+    {
+        return ExpenditureTypes[index];
+    }
+    return {};
+}
+
+ExpenditureType ScriptEngine::StringToExpenditureType(const std::string_view expenditureType)
+{
+    size_t i = 0;
+    for (auto sz : ExpenditureTypes)
```
duncanspumpkin, line 810: I think you could do 
```
auto it = std::find(ExpenditureTypes.begin(), ExpenditureTypes.end(), sz); 
if (it != ExpenditureTypes.end())
{
    return  static_cast<ExpenditureType>(std::distance(ExpenditureTypes.begin(),it));
}

```

src/openrct2/scripting/ScriptEngine.cpp
IntelOrca: Well spotted! Thanks!

src/openrct2/scripting/ScriptEngine.cpp
IntelOrca: In the bindings I have been using `null` or not defining it for non-applicable values.

src/openrct2/scripting/ScriptEngine.cpp
tupaschoal: Just remember that though this looks cleaner, it iterates on the whole container twice, as both `find` and `distance` have ~`O(n)` complexity on worst case

duncanspumpkin: It doesn't iterate the list on `std::distance` if it is a random access iterator that std::array or std::vector would offer. Also distance wouldn't be doing string compares so is considerably less expensive.

Issue #7006, opened 2018-01-11
Submarine Ride is in the wrong research group


SidaRadis:
**OS:** Windows 7
**Version:** 0.1.2
**Commit/Build:** 47a2360

The Submarine Ride is researched as a gentle ride instead of a water ride.
Edit: Water Coaster is researched as a water ride but is part of the roller coaster group.

Liampie:
Isn't the submarine classified as both?

spacek531:
Not any more, with the current changes to OpenRCT2. Rides are now grouped into their track type, which now only exists in one tab in the rides menu (instead of two like some e.g. river rafts, row boats, submarine).

It means that the mat slides I made for CTR Christmas are a water ride and there's not a whole lot I can do about it.

Gymnasiast:
This has now been addressed, but for new games only. Old saves will continue to have them in the "wrong" group.

PR #11521, opened 2020-04-28
Make ResearchItem use ObjectEntryIndex

Gymnasiast:


src/openrct2/management/Research.cpp
```diff
@@ -540,8 +540,12 @@ bool research_insert_ride_entry(uint8_t rideType, ObjectEntryIndex entryIndex, u
 {
     if (rideType != RIDE_TYPE_NULL)
     {
-        research_insert(
-            { static_cast<uint32_t>(RESEARCH_ENTRY_RIDE_MASK | (rideType << 8) | entryIndex), category }, researched);
+        ResearchItem tmpItem = {};
```
duncanspumpkin, line 543: Why create a temporary then initialise it. Can't you create a constructor in research item for this.

src/openrct2/management/Research.cpp
```diff
@@ -560,7 +564,11 @@ void research_insert_ride_entry(ObjectEntryIndex entryIndex, bool researched)
 
 void research_insert_scenery_group_entry(ObjectEntryIndex entryIndex, bool researched)
 {
-    research_insert({ entryIndex, RESEARCH_CATEGORY_SCENERY_GROUP }, researched);
+    ResearchItem tmpItem = {};
```
duncanspumpkin, line 567: As above

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1094,7 +1094,8 @@ static int32_t cc_load_object(InteractiveConsole& console, const arguments_t& ar
             for (int32_t j = 0; j < MAX_RIDE_TYPES_PER_RIDE_ENTRY; j++)
             {
                 rideType = rideEntry->ride_type[j];
-                research_insert_ride_entry(rideType, groupIndex, rideEntry->category[0], true);
+                uint8_t category = RideTypeDescriptors[rideType].Category;
```
Gymnasiast, line 1097: Add RIDE_TYPE_NULL check.

Gymnasiast:
Because OpenRCT2 uses the ride type in most places to determine the category. This is necessary because it groups vehicles together. Vanilla had most vehicles as separate, so it in vanilla it made some sense to have it on the ride entry.

The commit finishes the job, solving https://github.com/OpenRCT2/OpenRCT2/issues/7006 in the process.

src/openrct2/management/Research.h
```diff
@@ -19,32 +19,80 @@ struct rct_ride_entry;
...
     {
     }
+    ResearchItem(uint8_t _type, ObjectEntryIndex _entryIndex, uint8_t _baseRideType, uint8_t _category, uint8_t _flags)
+        : entryIndex(_entryIndex)
+        , baseRideType(_baseRideType)
+        , type(_type)
+        , flags(_flags)
+        , category(_category)
+    {
+    }
```
tupaschoal, line 56: I think we should use delegating constructor as most as we can to avoid leaving fields uninitialized. If this can be used here, it would would mean converting this
```cpp
    constexpr ResearchItem(uint32_t _rawValue, uint8_t _category, uint8_t _flags)
        : rawValue(_rawValue)
        , flags(_flags)
        , category(_category)
    {
    }
```` 
to something like
```cpp
    constexpr ResearchItem(uint32_t _rawValue, uint8_t _category, uint8_t _flags)
        : ResearchItem(/* 5 parameters */)
    {
    }
```

src/openrct2/management/Research.h
```diff
@@ -19,32 +19,80 @@ struct rct_ride_entry;
...
+
+    RCT12ResearchItem ToRCT12ResearchItem() const
+    {
+        RCT12ResearchItem retItem = {};
```
tupaschoal, line 60:
```suggestion
        RCT12ResearchItem retItem {};
```

src/openrct2/management/Research.h
duncanspumpkin: There are two constructors one for each type of the Union. I think as I right now is as good as you can get.

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2548,7 +2548,7 @@ class S4Importer final : public IParkImporter
 
                 ResearchItem tmpResearchItem = {};
                 ConvertResearchEntry(&tmpResearchItem, researchItem, researchType);
-                dst->Assoc = static_cast<uint32_t>(tmpResearchItem.rawValue);
+                dst->Assoc = tmpResearchItem.rawValue;
```
duncanspumpkin, line 2551: Is this a conscious change to not copy the flags and category

Gymnasiast: This hasn't changed, really. dst->Assoc is a 32-bit integer, it never fitted the category and now it also doesn't fit the flags. Only the research type (8 bit), entry index (16 bit) and ride type (8 bit) fit. Which is fine, since flags and category are not used in news items anyway.

duncanspumpkin:
I'd like to see further work in the future to change the news items so that we can get rid of the raw value. Could you create an issue for it so it is not forgotten

src/openrct2/management/Research.h
Gymnasiast: Is this is a style thing, or does it make a difference in a technical way?

tupaschoal: Theoretically you're creating an empty object and assigning versus just initializing to empty, but the compiler should optimize it anyway, so it ends up being a style choice

PR #11528, opened 2020-04-28
Refactor game action binding for plugins

IntelOrca:
Adds parameter visiting for game actions to reduce code needed for binding game actions to JavaScript objects. This should make it much quicker now to support all the remaining game actions in the plugin system.

### DIFF
--- a/src/openrct2/actions/GameAction.h
+++ b/src/openrct2/actions/GameAction.h
@@ -166,6 +166,7 @@ public:
 
     template<typename T> void Visit(const std::string_view& name, T& param)
     {
+        static_assert(std::is_arithmetic<T>::value, "Not an arithmetic type");
         auto value = static_cast<int32_t>(param);
         Visit(name, value);
         param = static_cast<T>(value);
--- a/src/openrct2/actions/GameAction.h
+++ b/src/openrct2/actions/GameAction.h
@@ -140,6 +140,38 @@ public:
     std::string GetErrorMessage() const;
 };
 
+/**
+ *
+ */
+class GameActionParameterVisitor
+{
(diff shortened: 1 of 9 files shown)


src/openrct2/actions/GameAction.h
```diff
@@ -140,6 +140,38 @@ class GameActionResult
...
+    {
+        Visit("x", param.x);
+        Visit("y", param.y);
+        Visit("z", param.z);
+        Visit("direction", param.direction);
+    }
+
+    template<typename T> void Visit(const std::string_view& name, T& param)
+    {
+        auto value = static_cast<int32_t>(param);
```
tupaschoal, line 169: Maybe we should assert that it is number-like, to make compilation errors clearer, should we ever try to visit it with a `T` that is not
```suggestion
        static_assert(std::is_arithmetic<T>::value, "Not an arithmetic type");
        auto value = static_cast<int32_t>(param);
```

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -873,6 +916,51 @@ void ScriptEngine::RunGameActionHooks(const GameAction& action, std::unique_ptr<
     }
 }
 
+static std::unique_ptr<GameAction> CreateGameActionFromActionId(const std::string& actionid)
+{
+    const static std::map<std::string, uint32_t> ActionNameToType = {
```
tupaschoal, line 921: Won't make a huge difference, but we don't need the sorting anyway
```suggestion
    const static std::unordered_map<std::string, uint32_t> ActionNameToType = {
```

PR #11533, opened 2020-04-29
Fix #11526: Plugin: Crash when using sprite type in park.postMessage

### LINKED ISSUE #11526: Plugin: Crash when using sprite type in park.postMessage
**OS:** Windows 10
**Version:** 0.2.6-362
**Commit/Build:** 96222ebac

When using the function `park.postMessage` with a message type that required a sprite to be drawn, a crash occurs when the game tries to draw it. This seems to be caused by the sprite ID for the message having an invalid index of -1.

**Steps to reproduce:**
1. Create a simple script that posts a message using the "peep" type
2. Open any park

**Script**
```js
var main = function () {
	park.postMessage({
		type: 'peep_on_attraction',
		text: 'Hello Park!'
	});
};
registerPlugin({
    name: 'test',
    version: '0.1',
    authors: ['test'],
    type: 'local',
    main: main
});

```

IntelOrca:
- Add safety checks for subject.
- Implement subject as another property in `postMessage`.
- Support formatted strings in `postMessage`.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -652,6 +652,13 @@ declare global {
     interface ParkMessage {
         type: ParkMessageType;
         text: string;
+
+        /**
+         * Ride ID for attraction.
+         * Entity ID for peep_on_attraction or peep.
+         * Researched item for research.
+         */
+        subject?: number;
     }
 
     interface Park {
--- a/src/openrct2-ui/windows/GameBottomToolbar.cpp
+++ b/src/openrct2-ui/windows/GameBottomToolbar.cpp
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2014-2019 OpenRCT2 developers
+ * Copyright (c) 2014-2020 OpenRCT2 developers
  *
  * For a complete list of all authors, please refer to contributors.md
  * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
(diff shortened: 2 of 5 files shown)


src/openrct2/peep/Peep.cpp
```diff
@@ -329,6 +329,11 @@ bool rct_sprite::IsPeep() const
     return peep.sprite_identifier == SPRITE_IDENTIFIER_PEEP;
 }
 
+bool rct_sprite::IsVehicle() const
+{
+    return peep.sprite_identifier == SPRITE_IDENTIFIER_VEHICLE;
+}
```
Broxzier, line 335: Wouldn't this be better to defined in `src/openrct2/ride/Vehicle.cpp` and use the `vehicle` identifier rather than `peep`?

src/openrct2/peep/Peep.cpp
```diff
@@ -339,6 +344,16 @@ Peep* rct_sprite::AsPeep()
...
 
+Vehicle* rct_sprite::AsVehicle()
+{
+    Vehicle* result = nullptr;
+    if (IsVehicle())
+    {
+        return reinterpret_cast<Vehicle*>(this);
+    }
+    return result;
+}
```
Broxzier, line 355: Same as the previous comment.

IntelOrca: I wasn't using it in the end, so have removed it.

src/openrct2/peep/Peep.cpp
IntelOrca: Leaving it as `peep` was an unintentional, nice spot. I didn't actually use these functions in the end, so have removed them.

PR #11547, opened 2020-04-30
Remove code for Windows, macOS from Testpaint

Gymnasiast:


duncanspumpkin:
Test paint should be removed entirely.

Gymnasiast:
I disagree. We still have many cases where the drawing is not correct and TestPaint call help clear these up.

PR #11552, opened 2020-04-30
[Plugin] Implement custom window tabs

IntelOrca:
* Implement tabs on custom windows.
* Fix off-by-one error with control sizes.
* Improve window resize support.
* Add border property to button so image buttons can optionally have borders.
* Add `onUpdate` event to window.
* Fix UTF-8 strings not showing correctly.
* Fix re-positioning of widgets on a window.
* Add support for tooltips.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -733,6 +733,7 @@ declare global {
         width: number;
         height: number;
         name?: string;
+        tooltip?: string;
         isDisabled?: boolean;
     }
 
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -724,7 +724,7 @@ declare global {
      * Represents the type of a widget, e.g. button or label.
      */
     type WidgetType =
-        "button" | "checkbox" | "dropdown" | "groupbox" | "label" | "spinner" | "tabview" | "viewport";
+        "button" | "checkbox" | "dropdown" | "groupbox" | "label" | "spinner" | "viewport";
 
     interface Widget {
         type: WidgetType;
@@ -822,8 +822,7 @@ declare global {
         frameBase: number;
         frameCount?: number;
         frameDuration?: number;
-        offsetX?: number;
-        offsetY?: number;
+        offset?: Coord2;
     }
 
     interface WindowTabDesc {
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -737,6 +737,11 @@ declare global {
     }
 
     interface ButtonWidget extends Widget {
+        /**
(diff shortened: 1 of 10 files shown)


src/openrct2-ui/scripting/CustomWindow.cpp
```diff
@@ -172,6 +175,43 @@ namespace OpenRCT2::Ui::Windows
...
+        }
+    };
+
+    struct CustomTabDesc
+    {
+        ImageId imageFrameBase;
+        uint32_t imageFrameCount;
+        uint32_t imageFrameDuration;
+        int32_t offsetX;
+        int32_t offsetY;
```
tupaschoal, line 189: Can't (and shouldn't we) use `ScreenCoordsXY` for this?

src/openrct2-ui/scripting/CustomWindow.cpp
```diff
@@ -202,8 +201,13 @@ namespace OpenRCT2::Ui::Windows
...
                 result.imageFrameCount = AsOrDefault(dukImage["frameCount"], 0);
                 result.imageFrameDuration = AsOrDefault(dukImage["frameDuration"], 0);
-                result.offsetX = AsOrDefault(dukImage["offsetX"], 0);
-                result.offsetY = AsOrDefault(dukImage["offsetY"], 0);
+
+                auto dukCoord = dukImage["offset"];
+                if (dukCoord.type() == DukValue::Type::OBJECT)
+                {
+                    result.offset.x = AsOrDefault(dukCoord["x"], 0);
+                    result.offset.y = AsOrDefault(dukCoord["y"], 0);
```
tupaschoal, line 209: Alternatively:
```suggestion
                    result.offset = { AsOrDefault(dukCoord["x"], 0), AsOrDefault(dukCoord["y"], 0) };
```

distribution/openrct2.d.ts
```diff
@@ -822,8 +822,7 @@ declare global {
         frameBase: number;
         frameCount?: number;
         frameDuration?: number;
-        offsetX?: number;
-        offsetY?: number;
+        offset?: Coord2;
```
tupaschoal, line 825: We should rename `Coord2` and `Coord3` to either the same as their counterparts in C++ or something more meaningful at some point

IntelOrca: At the moment they are used for both map and screen coordinates. There is no strict type checking in JavaScript and TypeScript will be happy if the interface has the same fields which they will do.

Maybe I will change it, but in a separate PR.

tupaschoal: Yeah, I'm ok with it being in a different PR (I thought it would be), just don't want it to be forgotten.

Issue #11293, opened 2020-04-14
Tweak scream amount


Richard-L:
OpenRCT2, v0.2.5-127-g0cbd70029 (0cbd700 on develop) provided by GitHub
Windows 7 (x86-64)

Some ideas I had posted here: https://github.com/OpenRCT2/OpenRCT2/issues/11070#issuecomment-609507171

> Muting is a bad idea I agree, but the screaming should be tweaked. This has been brought up on both the RCF discord and in the project's new own server too:
> 
> > Making it so they don't scream over the slightest incline? Don't get me wrong, it's hilarious, but jesus christ they can get loud
> 
> The immediate parameters that would come to mind are:
> 
> * how long/deep is the drop (height difference)
> * how many peeps are on the train
> 
> So for example if the drop is very short, or not deep, and the train only has 4-6 peeps on it, the likelihood for a scream is 0-10%. And so forth...

Thanks for considering.

duncanspumpkin:
Is this actually still an issue. We fixed the bug in the scream amount. All of these parameters are sort of taken into account as well

annoyingdroid:
If that's the case then no problem. May want to go ahead and close the issue though

duncanspumpkin:
Have a look into it and see what you think. The scream generation code isn't the easiest to follow. This issue was original created before the bug fix but perhaps there is still tweaking to the generation that should be done. I don't play with sound on so maybe I'm not the best to speak on this.

Richard-L:
I think it's definitely much improved. [This park](https://www.nedesigns.com/park/4775/sommerrodelbahn-zullerhorn/) served as a decent example as it has single car coaster trains with a max of 2 peeps per train, and lots of small little drops where there would always be a scream which felt too much. Now it's much more downtoned.

Is something like the height difference of a drop a factor for scream likelihood right now? Could be worth considering. I can look for more example parks, or this could also just be closed.

duncanspumpkin:
At present the vehicle has to be at a certain angle and speed to produce a scream. The number of guests on a ride effects the chance of a scream.

duncanspumpkin:
You could also consider the g forces when creating a scream

ocalhoun6:
> At present the vehicle has to be at a certain angle and speed to produce a scream. The number of guests on a ride effects the chance of a scream.

The number of guests on the ride, or the number of guests on that train?

[...]

PR #11573, opened 2020-05-01
Fixed logic with returned screamId. Refactored to use enums.

tylerleamon:
When UpdateScreamSound() was called, the random number generated would return a NoScream SoundID. The two resulting if statements would be executed, causing a SoundId::NoScream to be read as Null instead. This Null would cause the UpdateScreamSound() to be called continually on the same drop until a Scream was generated from RNG. The resulting code fix assumes one scream check per drop.

Also updated some case statements for enum use.

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -5467,7 +5467,7 @@ void Vehicle::UpdateSound()
 
     switch (vehicleEntry->sound_range)
     {
-        case 3:
+        case SOUND_RANGE_WHISTLE:
             screamId = scream_sound_id;
             if (!(gCurrentTicks & 0x7F))
             {
@@ -5486,7 +5486,7 @@ void Vehicle::UpdateSound()
             screamVolume = 255;
             break;
 
-        case 4:
+        case SOUND_RANGE_BELL:
             screamId = scream_sound_id;
             if (!(gCurrentTicks & 0x7F))
             {
@@ -5511,7 +5511,7 @@ void Vehicle::UpdateSound()
                 screamId = UpdateScreamSound();
                 if (screamId == SoundId::NoScream)
                     screamId = SoundId::Null;
-                if (screamId == SoundId::Null)
+                else if (screamId == SoundId::Null)
                     goto loc_6D7A97;
                 break;
             }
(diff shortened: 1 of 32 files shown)


AaronVanGeffen:
Presumably, this fixes #11293?

Gymnasiast:
`Fix: [#11293]: Guests scream on every descent, no matter how small.`

Gymnasiast:
This will also need a network version bump, I think.

tylerleamon:
> Presumably, this fixes #11293?

Not necessarily. This fix is to prevent multiple scream checks on the same drop. That fix was a request to change the parameters that considered a valid scream during a drop.

tupaschoal:
I think it fixes #5451 actually

AaronVanGeffen:
Thanks, I knew there was an older bug, but couldn't find it while on my phone!

src/openrct2/ride/Vehicle.cpp
```diff
@@ -5511,10 +5511,12 @@ void Vehicle::UpdateSound()
                 screamId = UpdateScreamSound();
                 if (screamId == SoundId::NoScream)
                     screamId = SoundId::Null;
-                if (screamId == SoundId::Null)
+                else if (screamId == SoundId::Null)
                     goto loc_6D7A97;
                 break;
             }
+            break;
```
duncanspumpkin, line 5518: Revert this break.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -5511,10 +5511,12 @@ void Vehicle::UpdateSound()
                 screamId = UpdateScreamSound();
                 if (screamId == SoundId::NoScream)
                     screamId = SoundId::Null;
-                if (screamId == SoundId::Null)
+                else if (screamId == SoundId::Null)
```
duncanspumpkin, line 5514: agreed that this looks correct

duncanspumpkin:
Okay I checked vanilla assembly. Your if / else looks correct and fixes a mistake (looking at the assembly the jump this represents is very similar in number to 0x6d7a97 so I think when we implemented we misread it) but the extra break would deviate from vanilla. This code is quite different in assembly so its hard to follow completely but it definitely goes to `0x6d7a97` after the `if ((vehicleEntry->flags & VEHICLE_ENTRY_FLAG_RIDERS_SCREAM))`.

duncanspumpkin:
I'll need to create an update bpb1000Ticks replay to make the ci happy. Don't worry about that I'll fix it.

duncanspumpkin:
i've tried to be clever and do two replays at once but that of course does not work. I'll fix this tomorrow.

PR #11574, opened 2020-05-01

Gymnasiast:
Split off from #11507, because that one segfaults in places it doesn't on my system.


src/openrct2/actions/LargeSceneryPlaceAction.hpp
```diff
@@ -48,15 +48,15 @@ DEFINE_GAME_ACTION(LargeSceneryPlaceAction, GAME_COMMAND_PLACE_LARGE_SCENERY, La
 {
 private:
     CoordsXYZD _loc;
-    uint8_t _sceneryType{ std::numeric_limits<uint8_t>::max() };
+    ObjectEntryIndex _sceneryType{ std::numeric_limits<uint8_t>::max() };
```
Gymnasiast, line 51:
```suggestion
    ObjectEntryIndex _sceneryType{ OBJECT_ENTRY_INDEX_NULL };
```

PR #11586, opened 2020-05-02
Vehicle update 9473 05

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7804,19 +7803,19 @@ static bool vehicle_update_motion_collision_detection(
...
 {
-    Vehicle* previousVehicle = GET_VEHICLE(vehicle->prev_vehicle_on_ride);
-    Vehicle* nextVehicle = GET_VEHICLE(vehicle->next_vehicle_on_ride);
+    Vehicle* previousVehicle = GET_VEHICLE(prev_vehicle_on_ride);
+    Vehicle* nextVehicle = GET_VEHICLE(next_vehicle_on_ride);
 
-    vehicle->track_progress = 168;
-    vehicle->vehicle_type ^= 1;
+    track_progress = 168;
+    vehicle_type ^= 1;
```
duncanspumpkin, line 7812: Could you make this explicit the change. There are only two vehicle types

PR #11587, opened 2020-05-02
[Plugin] Fix game action hook player and result

IntelOrca:
- Player was not being attached to the game action event args.
- OpenRCT2 was looking for error on the event args object rather than the result sub object.


IntelOrca:
Apparently the build was interrupted...

> ** BUILD INTERRUPTED **

PR #11593, opened 2020-05-02
Part of #9473: Create Vehicle::TrainReadToDepart

hdpoliveira:
Converted from train_ready_to_depart


duncanspumpkin:
You've lost a 'y'. It should be train ready to depart

PR #11594, opened 2020-05-02
Extract function gNewsItems_first_open_slot

hdpoliveira:


src/openrct2/management/NewsItem.cpp
```diff
@@ -295,6 +295,20 @@ std::optional<CoordsXYZ> news_item_get_subject_location(int32_t type, int32_t su
     return subjectLoc;
 }
 
+static NewsItem* gNewsItems_first_open_slot()
```
tupaschoal, line 298:
```suggestion
static NewsItem* NewsItemsFirstOpenSlot()
```

src/openrct2/management/NewsItem.cpp
```diff
@@ -295,6 +295,20 @@ std::optional<CoordsXYZ> news_item_get_subject_location(int32_t type, int32_t su
...
+{
+    NewsItem* newsItem = gNewsItems;
+
+    while (newsItem->Type != NEWS_ITEM_NULL)
```
duncanspumpkin, line 302: Can we change this into a more normal loop like a for loop or even better use some std functions like a std::find

hdpoliveira: The issue is this function does two things: find the first open spot and close items. The "iterator" is not incremented on every iteration because a new spot may open by closing the current item.

To transform this in a regular loop we'd need to separate the two concerns: closing items and finding the first open spot.

duncanspumpkin: Splitting it into two would make more sense i think.

src/openrct2/management/NewsItem.cpp
```diff
@@ -295,6 +295,20 @@ std::optional<CoordsXYZ> news_item_get_subject_location(int32_t type, int32_t su
     return subjectLoc;
 }
 
+static NewsItem* news_item_first_open_slot()
```
Gymnasiast, line 298:
```suggestion
static NewsItem* news_item_first_open_queue_slot()
```
There is also a similar function for finding a slot in the history, so this hopefully keeps them separated better.

hdpoliveira:
Ok, I'm trying to understand how does this work, and why we have this weird condition on the 11th position of the queue. My conclusion is that we have actually two queues, one that goes from 0-10 (11 positions) with the most recent news, and one that goes from 11-60 (50 positions), with old news.

This is why at news_item_init_queue (NewsItem.cpp) we have:
```cpp
void news_item_init_queue()
{
    news_item_get(0)->Type = NEWS_ITEM_NULL;
    news_item_get(11)->Type = NEWS_ITEM_NULL;
```

and at window_news_update (News.cpp) we have:
```cpp
    for (i = 11; i < 61; i++)
    {
        if (news_item_is_empty(i))
            return;

        if (j == 0)
        {
            NewsItem* const newsItem = news_item_get(i);
```

So the check for the 11th position is actually the last position in the more recent news queue: if there is no slot in the more recent queue, move the oldest news in the more recent queue to the old news queue, and free the 11th position.

hdpoliveira:
@duncanspumpkin I don't think there is a way of changing this code to a more normal for loop given this entire infrastructure. One possibility would be to separate the arrays in two, then it would be something like:
```
auto pos = find_open_position_in_recent_news_array()
if (pos = end) {
    move_oldest_news_from_recent_news_array_to_old_news_array()
    pos = recent_news_array.back()
}
return pos
```

but this is a major refactoring effort comparing to the objective of this pull request (I just saw duplicated code, and I thought "ok, extract function!" :) )

My proposal: let's merge this PR, and file issues for refactoring the queue into two.
What do you think?

Gymnasiast:
I personally think that's a good idea. I tend to favour several smaller increments over one big bang anyway.

duncanspumpkin:
Yes that sounds sensible.

PR #11578, opened 2020-05-01
Keep scouring openrct2-ui to reduce individual {x, y} usage

tupaschoal:
I advise looking it on a commit-per-commit basis

PR #11598, opened 2020-05-02
Favour CoordsXYZ for ride related functions

tupaschoal:
I advise looking it on a commit-per-commit basis

Follow-up of @Gymnasiast 's [comment on #11578](https://github.com/OpenRCT2/OpenRCT2/pull/11578#discussion_r418723566) and more

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -1717,8 +1717,7 @@ static void RideConstructPlacedForwardGameActionCallback(const GameAction* ga, c
         }
 
         CoordsXYE next_track;
-        if (track_block_get_next_from_zero(
-                trackPos.x, trackPos.y, trackPos.z, ride, trackDirection, &next_track, &trackPos.z, &trackDirection, false))
+        if (track_block_get_next_from_zero(trackPos, ride, trackDirection, &next_track, &trackPos.z, &trackDirection, false))
         {
             _currentTrackBegin.x = next_track.x;
             _currentTrackBegin.y = next_track.y;
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -1761,7 +1761,7 @@ static void RideConstructPlacedBackwardGameActionCallback(const GameAction* ga, 
         }
 
         track_begin_end trackBeginEnd;
-        if (track_block_get_previous_from_zero(trackPos.x, trackPos.y, trackPos.z, ride, trackDirection, &trackBeginEnd))
+        if (track_block_get_previous_from_zero(trackPos, ride, trackDirection, &trackBeginEnd))
         {
             _currentTrackBegin.x = trackBeginEnd.begin_x;
             _currentTrackBegin.y = trackBeginEnd.begin_y;
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -1189,7 +1189,7 @@ static void window_ride_construction_resize(rct_window* w)
     if (_rideConstructionState == RIDE_CONSTRUCTION_STATE_FRONT)
     {
         disabledWidgets |= (1ULL << WIDX_NEXT_SECTION);
-        if (window_ride_construction_update_state(nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr))
+        if (window_ride_construction_update_state(nullptr, nullptr, nullptr, nullptr, nullptr, nullptr))
         {
             disabledWidgets |= (1ULL << WIDX_CONSTRUCT);
         }
@@ -1197,7 +1197,7 @@ static void window_ride_construction_resize(rct_window* w)
     else if (_rideConstructionState == RIDE_CONSTRUCTION_STATE_BACK)
     {
         disabledWidgets |= (1ULL << WIDX_PREVIOUS_SECTION);
-        if (window_ride_construction_update_state(nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr, nullptr))
+        if (window_ride_construction_update_state(nullptr, nullptr, nullptr, nullptr, nullptr, nullptr))
         {
             disabledWidgets |= (1ULL << WIDX_CONSTRUCT);
         }
@@ -1799,8 +1799,7 @@ static void window_ride_construction_construct(rct_window* w)
     _trackPlaceCost = MONEY32_UNDEFINED;
     ride_construction_invalidate_current_track();
     if (window_ride_construction_update_state(
-            &trackType, &trackDirection, &rideIndex, &liftHillAndAlternativeState, &trackPos.x, &trackPos.y, &trackPos.z,
-            &properties))
+            &trackType, &trackDirection, &rideIndex, &liftHillAndAlternativeState, &trackPos, &properties))
     {
         window_ride_construction_update_active_elements();
         return;
@@ -2323,7 +2322,7 @@ static void window_ride_construction_paint(rct_window* w, rct_drawpixelinfo* dpi
     ride_id_t rideIndex;
     int32_t trackType, trackDirection, liftHillAndInvertedState;
     if (window_ride_construction_update_state(
-            &trackType, &trackDirection, &rideIndex, &liftHillAndInvertedState, nullptr, nullptr, nullptr, nullptr))
+            &trackType, &trackDirection, &rideIndex, &liftHillAndInvertedState, nullptr, nullptr))
         return;
 
     // Draw track piece
@@ -2579,8 +2578,7 @@ void sub_6C94D8()
             if (!(_currentTrackSelectionFlags & TRACK_SELECTION_FLAG_TRACK))
             {
                 if (window_ride_construction_update_state(
-                        &type, &direction, &rideIndex, &liftHillAndAlternativeState, &trackPos.x, &trackPos.y, &trackPos.z,
-                        nullptr))
+                        &type, &direction, &rideIndex, &liftHillAndAlternativeState, &trackPos, nullptr))
                 {
                     ride_construction_remove_ghosts();
                 }
@@ -2668,7 +2666,7 @@ void sub_6C94D8()
 static void window_ride_construction_update_map_selection()
 {
     int32_t trackType, trackDirection;
-    CoordsXY trackPos{};
+    CoordsXYZ trackPos{};
 
     map_invalidate_map_selection_tiles();
     gMapSelectFlags |= MAP_SELECT_FLAG_ENABLE_CONSTRUCT;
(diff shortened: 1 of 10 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -4119,13 +4111,11 @@ static bool ride_check_start_and_end_is_station(CoordsXYE* input)
...
-    int32_t returnY = startElement->y;
-    int32_t startX = returnX;
-    int32_t startY = returnY;
-    TileElement* returnTrackElement = startElement->element;
+    auto returnPos = *startElement;
+    int32_t startX = returnPos.x;
+    int32_t startY = returnPos.y;
     track_begin_end trackBeginEnd;
-    while (track_block_get_previous(returnX, returnY, returnTrackElement, &trackBeginEnd))
+    while (track_block_get_previous(returnPos, &trackBeginEnd))
```
duncanspumpkin, line 4118: pass the start element into here this doesn't make much sense as is.

src/openrct2/ride/Ride.cpp
```diff
@@ -4119,13 +4111,11 @@ static bool ride_check_start_and_end_is_station(CoordsXYE* input)
 static void ride_set_boat_hire_return_point(Ride* ride, CoordsXYE* startElement)
 {
     int32_t trackType = -1;
-    int32_t returnX = startElement->x;
-    int32_t returnY = startElement->y;
-    int32_t startX = returnX;
-    int32_t startY = returnY;
-    TileElement* returnTrackElement = startElement->element;
+    auto returnPos = *startElement;
+    int32_t startX = returnPos.x;
```
duncanspumpkin, line 4115: remove these variables only trackBeginEnd should be defined here.

src/openrct2/ride/Ride.cpp
```diff
@@ -4134,16 +4124,15 @@ static void ride_set_boat_hire_return_point(Ride* ride, CoordsXYE* startElement)
...
-        auto newCoords = sub_6C683D({ trackCoords, static_cast<Direction>(direction) }, trackType, 0, &returnTrackElement, 0);
-        returnX = newCoords == std::nullopt ? trackCoords.x : newCoords->x;
-        returnY = newCoords == std::nullopt ? trackCoords.y : newCoords->y;
+        auto newCoords = sub_6C683D({ trackCoords, static_cast<Direction>(direction) }, trackType, 0, &returnPos.element, 0);
+        returnPos = newCoords == std::nullopt ? CoordsXYE{ trackCoords, returnPos.element }
+                                              : CoordsXYE{ *newCoords, returnPos.element };
     };
 
-    trackType = returnTrackElement->AsTrack()->GetTrackType();
+    trackType = returnPos.element->AsTrack()->GetTrackType();
```
duncanspumpkin, line 4132: this is the startelement

src/openrct2/ride/Ride.cpp
```diff
@@ -4588,16 +4577,14 @@ static void ride_create_vehicles_find_first_block(Ride* ride, CoordsXYE* outXYEl
 
     assert(curTrackElement != nullptr);
 
-    int32_t x = curTrackPos.x;
-    int32_t y = curTrackPos.y;
+    CoordsXY trackPos = curTrackPos;
```
duncanspumpkin, line 4580: variable not required.

src/openrct2/ride/Ride.cpp
```diff
@@ -6263,17 +6244,15 @@ bool ride_select_forwards_from_back()
...
         int32_t direction = direction_reverse(_currentTrackPieceDirection);
         CoordsXYE next_track;
-        if (track_block_get_next_from_zero(x, y, z, ride, direction, &next_track, &z, &direction, false))
+        if (track_block_get_next_from_zero(_currentTrackBegin, ride, direction, &next_track, &z, &direction, false))
         {
             _rideConstructionState = RIDE_CONSTRUCTION_STATE_SELECTED;
             _currentTrackBegin.x = next_track.x;
             _currentTrackBegin.y = next_track.y;
-            _currentTrackBegin.z = z;
+            _currentTrackBegin.z = _currentTrackBegin.z;
```
duncanspumpkin, line 6255: This is a change that i don't think you meant to make. it should still be z.

src/openrct2/ride/Ride.cpp
tupaschoal: The `while` alters the `returnPos` value. If I remove it and write to `startElement` instead, I'll be modifying a pointer which I don't own (and that the caller actually uses later). Is this really intended?

src/openrct2/ride/Ride.cpp
tupaschoal: Inside the loop we compare `trackPos` (which is often updated) with `curTrackPos` which is the initial one. Should we really remove it?

duncanspumpkin:
As discussed ignore my comments. The variable names for these aren't great but they never were.

Gymnasiast:
Why did you merge this without any approvals?

duncanspumpkin:
I approved it over discord

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1959,7 +1957,7 @@ static void window_ride_construction_mouseup_demolish(rct_window* w)
             if (ride != nullptr)
             {
                 _stationConstructed = ride->num_stations != 0;
-                window_ride_construction_mouseup_demolish_next_piece(newCoords->x, newCoords->y, newCoords->z, direction, type);
+                window_ride_construction_mouseup_demolish_next_piece({ *newCoords, static_cast<Direction>(direction) }, type);
```
Gymnasiast, line 1960: Would have been better to just change the type of `direction` to Direction.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3545,11 +3541,11 @@ void ride_construction_toolupdate_construct(const ScreenCoordsXY& screenCoords)
...
         for (int zAttempts = 0; zAttempts < numAttempts; ++zAttempts)
         {
+            CoordsXYZ trackPos{};
             window_ride_construction_update_state(
-                &trackType, &trackDirection, &rideIndex, &liftHillAndAlternativeState, &mapCoords->x, &mapCoords->y, &z,
-                nullptr);
+                &trackType, &trackDirection, &rideIndex, &liftHillAndAlternativeState, &trackPos, nullptr);
             _currentTrackPrice = place_provisional_track_piece(
-                rideIndex, trackType, trackDirection, liftHillAndAlternativeState, mapCoords->x, mapCoords->y, z);
+                rideIndex, trackType, trackDirection, liftHillAndAlternativeState, trackPos);
```
Gymnasiast, line 3548: Is `z` still used in this function? It doesn't get updated anymore after this change, even though it used to.

src/openrct2/windows/_legacy.cpp
```diff
@@ -55,14 +55,11 @@ money32 place_provisional_track_piece(
...
-        result = maze_set_track(x, y, z, flags, true, 0, rideIndex, GC_SET_MAZE_TRACK_BUILD);
+        result = maze_set_track(trackPos.x, trackPos.y, trackPos.z, flags, true, 0, rideIndex, GC_SET_MAZE_TRACK_BUILD);
         if (result == MONEY32_UNDEFINED)
             return result;
 
-        _unkF440C5.x = x;
-        _unkF440C5.y = y;
-        _unkF440C5.z = z;
-        _unkF440C5.direction = trackDirection;
+        _unkF440C5 = { trackPos, static_cast<Direction>(trackDirection) };
```
Gymnasiast, line 62: Should have changed the type of `trackDirection` to `Direction`.

src/openrct2/windows/_legacy.cpp
```diff
@@ -439,16 +427,15 @@ void window_ride_construction_do_station_check()
...
 }
 
-void window_ride_construction_mouseup_demolish_next_piece(int32_t x, int32_t y, int32_t z, int32_t direction, int32_t type)
+void window_ride_construction_mouseup_demolish_next_piece(const CoordsXYZD& piecePos, int32_t type)
 {
     if (gGotoStartPlacementMode)
     {
-        z &= 0xFFF0;
-        _currentTrackBegin.z = z;
+        _currentTrackBegin.z = piecePos.z & 0xFFF0;
```
Gymnasiast, line 434: Should have changed this to use floor2 with COORDS_Z_STEP.

src/openrct2/ride/Ride.cpp
```diff
@@ -559,22 +556,22 @@ bool track_block_get_next(CoordsXYE* input, CoordsXYE* output, int32_t* z, int32
  *  rct2: 0x006C63D6
  */
 bool track_block_get_previous_from_zero(
-    int16_t x, int16_t y, int16_t z, Ride* ride, uint8_t direction, track_begin_end* outTrackBeginEnd)
+    const CoordsXYZ& startPos, Ride* ride, uint8_t direction, track_begin_end* outTrackBeginEnd)
 {
     uint8_t directionStart = direction;
     direction = direction_reverse(direction);
+    auto trackPos = startPos;
```
Gymnasiast, line 563: Why copy this?

src/openrct2/ride/Ride.cpp
```diff
@@ -1460,11 +1455,8 @@ void ride_construction_set_default_next_piece()
                 && slope != TRACK_SLOPE_DOWN_60;
             break;
         case RIDE_CONSTRUCTION_STATE_BACK:
-            x = _currentTrackBegin.x;
-            y = _currentTrackBegin.y;
-            z = _currentTrackBegin.z;
             direction = direction_reverse(_currentTrackPieceDirection);
-            if (!track_block_get_next_from_zero(x, y, z, ride, direction, &xyElement, &z, &direction, false))
+            if (!track_block_get_next_from_zero(_currentTrackBegin, ride, direction, &xyElement, &z, &direction, false))
```
Gymnasiast, line 1459: `z` gets updated by this function, but is never written back to `_currentTrackBegin.z`.

src/openrct2/ride/Ride.cpp
tupaschoal: `trackPos` is updated in the if statement right below. `startPos` is a const-ref.

src/openrct2-ui/windows/RideConstruction.cpp
tupaschoal: It is, nice catch!

src/openrct2/ride/Ride.cpp
Gymnasiast: Then why make `startPos` a const-ref in the first place?!

tupaschoal: That's debatable, I feel like making the interface `const &` by design makes the interface more consistent and we tend to gain more than lose, as most places won't need to alter the object inside. Some cases will require duplication, but then you're most likely just delaying the duplication to the callee instead of the caller.

PR #11601, opened 2020-05-02
Add missing "openrct2_corrupt" element type

Broxzier:


distribution/openrct2.d.ts
```diff
@@ -356,7 +356,7 @@ declare global {
     }
 
     type TileElementType =
-        "surface" | "footpath" | "track" | "small_scenery" | "wall" | "entrance" | "large_scenery" | "banner";
+        "surface" | "footpath" | "track" | "small_scenery" | "wall" | "entrance" | "large_scenery" | "banner" | "openrct2_corrupt";
```
tupaschoal, line 359:
```suggestion
        "banner" | "entrance" | "footpath" | "large_scenery" | "openrct2_corrupt"| "small_scenery" |"surface" | "track" | "wall";
```
Can we take advantage of this change and also add another commit that sorts this list? :D

Broxzier:
@Gymnasiast This does not add the element, it merely adds the definition to the typescript definitions so that you can actually compile typescripts using them without errors. There's also the `CorruptElement` type, and the plugin system even returns "openrct2_corrupt" for corrupt elements. Besides they work different than any other "currupt" types, their behaviour is defined in the code. Without this change, you'd have to work around it to get rid of the typescript errors.

distribution/openrct2.d.ts
Broxzier: Might as well do that for all other lists. 😄 Currently, they are sorted by their type ID though, which might be nice to keep too.

Broxzier:
To give an example, this code would make the typescript compiler raise an error:

```js
const element: TileElement = tile.getElement(index);
if (element.type == 'openrct2_corrupt')
  ...
```

'openrct2_corrupt' is a valid type, because the plug-in system returns it, it's just that the definition is missing in the typescript definition file.

Once corrupt elements are superseded, and thus plug-in support for them has been removed, this line could be removed easily. Then "old" plug-ins that use them will raise this error and need updating.

IntelOrca:
I think the argument is that as long as it is not documented, plugin writers can't assume it will be returned.

Broxzier:
I get that you don't want to encourage players to create plug-ins using features that will ve superseded eventually, but they currently exist in a ton of parks and are useful. If there's no proper way to check for them in plug-ins we end up with hacky tricks that may break in another way (for example one that checks if the type is _not_ of the documented ones, and we introduce a new element type, that new element type will be seen as corrupt by that plug-in).

I suggest adding a comment stating that they will be dropped eventually rather than hiding their existence.

IntelOrca:
We plan for corrupt elements to be removed and a visibility flag being added to every other element that isn't corrupt. Perhaps we could simulate this in the plugin system.

Broxzier:
I'm aware, and while that sounds better I don't think that's entirely feasible. Corrupt elements have more uses than just hiding the next element in the list, they are also being used as invisible objects just after tower rides to hide the top part, often used for custom supports.

I bumped into this because I'm working on a plug-in that can remove all useless corrupt elements with a single click. They are only considered useless when there are two corrupt elements right after each other or is the last in the list, and is not placed directly above a track element.

They may even have more uses that I'm not aware of. I wouldn't be surprised if someone found another use neat trick with them.

Gymnasiast:
Above example looks like a bug to me - if I place, say, a path or scenery above the tower, it should not change shape. Neither should a corrupt element.

Also: are you sure the element was placed _after_ the track piece? To me, it looks like the corrupt element was placed just before it.

Broxzier:
With the last commit, I've added an `isHidden` field to `TileElement`. When changing this, it either removes or places a corrupt elements right below it.

This way, plug-in creators are encouraged to use `isHidden` rather than spawning corrupt elements to hide other tile elements, and their scripts won't need updating when `isHidden` becomes a real field that does not alter anything other than the element itself.

While adding a field to our new struct is possible already and we could add/remove corrupt elements on load and save, there's a pitfall that Duncan pointed out on Discord. When a park is close to the tile elements limit it cannot add new corrupt elements to the map. Therefore it's best to add them right away when using this, at least for now.

```ts
registerPlugin({
        name: 'Thanos Scenery',
        version: '1.0',
        authors: ['Broxzier'],
        type: 'remote',
        main: () => {
            for (let x: number = 0; x < map.size.x; x++) {
                for (let y: number = 0; y < map.size.y; y++) {
                    const tile: Tile = map.getTile(x, y);
                    for (let i: number = 0; i < tile.numElements; i++)
                    {
                        const element: TileElement = tile.getElement(i);
                        if (element.type == 'small_scenery' as TileElementType)
                        {
                            element.isHidden = Math.random() > 0.5;
                        }
                    }
                }
            }
        }
    } as PluginMetadata);
```

Note that changing the `isHidden` property inside a `tile.elements.forEach` callback can cause problems since the list of tile elements can get reallocated. Since `tile.elements` is defined as readonly in the API, I don't think that's really an issue though. I've added a comment for this in the API (the same warning applies to when a user alters the tile element list in another way - for example `Tile.insertElement` - this is just less obvious).

src/openrct2/scripting/ScTile.hpp
```diff
@@ -491,6 +497,82 @@ namespace OpenRCT2::Scripting
...
+                return;
+
+            if (hide)
+            {
+                // Get index of our current element (has to be done now before inserting the corrupt element)
+                const auto elementIndex = _element - map_get_first_element_at(_coords);
+
+                // Insert corrupt element at the end of the list for this tile
+                // Note: Z = 255 guarantee this
+                TileElement* insertedElement = tile_element_insert({ _coords, 0xFF }, 0);
```
Gymnasiast, line 534: Could you create a constant for this use of 0xFF. It is also used for discarded elements.

PR #11619, opened 2020-05-03
Part of #9473: Create Vehicle::TryReconnectBoatToTrack()

hdpoliveira:
Converted from loc_6DA9F9

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -4304,27 +4304,27 @@ void Vehicle::UpdateTravellingBoat()
     UpdateMotionBoatHire();
 }
 
-static void loc_6DA9F9(Vehicle* vehicle, int32_t x, int32_t y, int32_t trackX, int32_t trackY)
+void Vehicle::TryReconnectBoatToTrack(const CoordsXY& currentBoatLocation, const CoordsXY& trackCoords)
 {
-    vehicle->remaining_distance = 0;
-    if (!vehicle_update_motion_collision_detection(vehicle, x, y, vehicle->z, nullptr))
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -4238,27 +4238,27 @@ void Vehicle::UpdateTravellingBoat()
     UpdateMotionBoatHire();
 }
 
-static void loc_6DA9F9(Vehicle* vehicle, int32_t x, int32_t y, int32_t trackX, int32_t trackY)
+void Vehicle::Loc6DA9F9(const CoordsXY& coords, const CoordsXY trackCoords)
```
tupaschoal, line 4241:
```suggestion
void Vehicle::Loc6DA9F9(const CoordsXY& coords, const CoordsXY& trackCoords)
```

src/openrct2/ride/Vehicle.h
```diff
@@ -350,6 +350,7 @@ struct Vehicle : SpriteBase
     void UpdateTravellingCableLift();
     void UpdateTravellingBoat();
     void UpdateMotionBoatHire();
+    void Loc6DA9F9(const CoordsXY& coords, const CoordsXY trackCoords);
```
tupaschoal, line 353:
```suggestion
    void Loc6DA9F9(const CoordsXY& coords, const CoordsXY& trackCoords);
```

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: I think we can name this function TryReconnectBoatToTrack

PR #11620, opened 2020-05-04
Part of #9473: Create Vehicle::UpdateTrackMotionMiniGolf

hdpoliveira:
Converted from vehicle_update_track_motion_mini_golf

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -8656,28 +8656,25 @@ loc_6DBE7F:
  *
  *
  */
-static int32_t vehicle_update_track_motion_mini_golf(Vehicle* vehicle, int32_t* outStation)
+int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
 {
     registers regs = {};
 
-    auto ride = get_ride(vehicle->ride);
-    if (ride == nullptr)
+    auto curRide = get_ride(ride);
+    if (curRide == nullptr)
         return 0;
 
-    rct_ride_entry* rideEntry = get_ride_entry(vehicle->ride_subtype);
-    rct_ride_entry_vehicle* vehicleEntry = vehicle->Entry();
+    rct_ride_entry* rideEntry = get_ride_entry(ride_subtype);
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -8798,7 +8795,7 @@ static int32_t vehicle_update_track_motion_mini_golf(Vehicle* vehicle, int32_t*
         _vehicleBankEndF64E37 = TrackDefinitions[trackType].bank_end;
         tileElement = map_get_track_element_at_of_type_seq(vehicle->TrackLocation, trackType, 0);
     }
-    int16_t x, y, z;
+    int16_t curX, curY, curZ;
```
Gymnasiast, line 8798: While we're here, please turn these into a `CoordsXYZ`.

duncanspumpkin:
I think this is maybe too early to do this function. The function is a big mess of gotos and should be split into multiple member functions. I really don't like the using of vehicle pointer pointing to this or next car.

Gymnasiast:
@duncanspumpkin Okay, what would you say is the best course of action? Close this one, or merge it? Regardless of the action we pick, we should create some refactor issues.

duncanspumpkin:
I'd rather we close this. We need to pull out the member functions so that we aren't changing the vehicle pointer as it basically reads as modifying the this pointer.

hdpoliveira:
@duncanspumpkin  is it ok if I work on this?

I noticed this function is a huge for loop (using gotos) that either iterates from the first vehicle on the train towards the tail, or from the last vehicle on the train towards the head, the later if the speed is negative. First thing I think is to extract the for loop body to a function. Then there will be no more the change of "vehicle" pointer.

hdpoliveira:
@duncanspumpkin Please take a look at https://github.com/OpenRCT2/OpenRCT2/compare/develop...hdpoliveira:vehicle_update_9473_10, it is already much cleaner than this one. Let me know if I should keep going, open a PR, or drop it all :)

duncanspumpkin:
If you look at the non golf update move you can see how we extracted the functions. The two update moves are ultimately very similar.

PR #11621, opened 2020-05-04
Add drawing code for more APVC, Water Coaster and B&M pieces

Gymnasiast:
This adds drawing code for the following:
- Boosters on the Air Powered Vertical Coaster
- Downward water channel pieces on the Water Coaster
- Vertical Drop Coaster-specific pieces on the Twister RC and vice versa.

None of these are enabled by default - they can be built by enabling the "Enable all drawable track pieces" cheat or by changing the ride type.


Fred-1044276:
What speed range will the Air Powered Vertical Coaster boosters have?

Gymnasiast:
The normal range of 6 - 107 km/h. It could be doubled (like on the Giga Coaster), but only if it's more realistic.

tupaschoal:
@Gymnasiast I just merged it, but I feel like maybe it deserved a changelog?

Fred-1044276:
The WaterCoaster.h file wasn't updated. Also, is it really possible to use the large level-to-steep/steep-to-level transition pieces from the twister coaster on the vertical drop coaster and the small level-to-steep/steep-to-level transitions from the vertical drop coaster on the twister? Maybe not.

Gymnasiast:
> The WaterCoaster.h file wasn't updated.

Because it would not have made any difference. Covered pieces are not handled by the mechanism you're thinking of.

> Also, is it really possible to use the large level-to-steep/steep-to-level transition pieces from the twister coaster on the vertical drop coaster and the small level-to-steep/steep-to-level transitions from the vertical drop coaster on the twister? Maybe not.

No, or at least not with the construction window as it is. That's why I haven't enabled them.

PR #11635, opened 2020-05-04
[Plugin] Add named parameters for ride related game actions

IntelOrca:
Also changes `GameActionResult` slightly. It will now look like this:
```
{ action: 'trackplace',
  args:
   { x: 1408,
     y: 2464,
     z: 112,
     direction: 0,
     id: 2,
     trackType: 0,
     brakeSpeed: 0,
     colour: 0,
     seatRotation: 4,
     trackPlaceFlags: 0,
     isFromTrackDesign: false },
  player: -1,
  type: 3,
  isClientOnly: false,
  result:
   { cost: 170,
     position:
      { x: 1424,
        y: 2480,
        z: 112 },
     expenditureType: 'ride_construction' } }
```

### DIFF
--- a/src/openrct2/actions/GameAction.h
+++ b/src/openrct2/actions/GameAction.h
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2014-2019 OpenRCT2 developers
+ * Copyright (c) 2014-2020 OpenRCT2 developers
  *
  * For a complete list of all authors, please refer to contributors.md
  * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
@@ -148,12 +148,22 @@ class GameActionParameterVisitor
 public:
     virtual ~GameActionParameterVisitor() = default;
 
+    virtual void Visit(const std::string_view& name, bool& param)
+    {
+    }
+
     virtual void Visit(const std::string_view& name, int32_t& param)
     {
     }
 
     virtual void Visit(const std::string_view& name, std::string& param)
     {
+    }
+
+    void Visit(CoordsXY& param)
+    {
+        Visit("x", param.x);
+        Visit("y", param.y);
     }
 
     void Visit(CoordsXYZD& param)
@@ -170,6 +180,11 @@ public:
         auto value = static_cast<int32_t>(param);
         Visit(name, value);
         param = static_cast<T>(value);
+    }
+
+    template<typename T, size_t _TypeID> void Visit(const std::string_view& name, NetworkObjectId_t<T, _TypeID>& param)
+    {
+        Visit(name, param.id);
     }
 };
 
--- a/src/openrct2/actions/TrackDesignAction.h
+++ b/src/openrct2/actions/TrackDesignAction.h
@@ -53,6 +53,7 @@ public:
     void AcceptParameters(GameActionParameterVisitor & visitor) override
     {
         visitor.Visit(_loc);
+        // TODO visit the track design (it has a lot of sub fields)
     }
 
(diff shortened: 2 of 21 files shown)


src/openrct2/actions/GameAction.h
```diff
@@ -156,6 +160,12 @@ class GameActionParameterVisitor
     {
     }
 
+    void Visit(CoordsXY& param)
+    {
+        Visit("x", param.x);
+        Visit("y", param.y);
+    }
```
tupaschoal, line 167: Does `Visit` modify the `CoordsXY`?

src/openrct2/actions/TrackDesignAction.h
```diff
@@ -50,6 +50,11 @@ DEFINE_GAME_ACTION(TrackDesignAction, GAME_COMMAND_PLACE_TRACK_DESIGN, TrackDesi
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
+    {
+        visitor.Visit(_loc);
+    }
```
tupaschoal, line 56: Doesn't `TrackDesign` need to be visited as well?

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -934,6 +945,48 @@ class DukFromGameActionParameterVisitor : public GameActionParameterVisitor
...
+    { "trackdesign", GAME_COMMAND_PLACE_TRACK_DESIGN },
+    { "trackplace", GAME_COMMAND_PLACE_TRACK },
+    { "trackremove", GAME_COMMAND_REMOVE_TRACK },
+    { "tracksetbrakespeed", GAME_COMMAND_SET_BRAKES_SPEED },
+};
+
+static std::string GetActionName(uint32_t commandId)
+{
+    auto it = std::find_if(
+        ActionNameToType.begin(), ActionNameToType.end(), [commandId](const auto& kvp) { return kvp.second == commandId; });
```
tupaschoal, line 972: `kvp` ?  Can we use a more meaningful name? (or is it an acronym I missed?)

src/openrct2/actions/GameAction.h
IntelOrca: Yes, this function is used to get and set the parameters.

src/openrct2/actions/TrackDesignAction.h
IntelOrca: At some point, I am not implementing it for plugins right now as it is very complicated.

src/openrct2/scripting/ScriptEngine.cpp
IntelOrca: "key value pair", pretty common naming convention isn't it?

src/openrct2/actions/TrackDesignAction.h
tupaschoal: Ok, since all of the others visit every member, I think we can leave a TODO here to make it clear it's intentionally unimplemented

src/openrct2/scripting/ScriptEngine.cpp
tupaschoal: Seems alright

src/openrct2/actions/TrackDesignAction.h
IntelOrca: Done

IntelOrca:
@tupaschoal Yes but I don't know how you can prevent things slipping through. So far it will be a case of reviewing PRs carefully.

PR #11639, opened 2020-05-05
Fix #11455: Object Selection window cuts off scenery names

### LINKED ISSUE #11455: Object selection window cut off object names
In the object selection window, objects and things have their names cut off as if there were a second column of information. This makes sense for rides, but not for other tabs. This makes selecting small scenery problematic for example, because the names are often long. I have to do my selection purely by hovering my cursor over all the objects and watching the image preview.

Using version 0.2.6 6c3c857 on windows 8.

Gymnasiast:


src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -1160,10 +1160,11 @@ static void window_editor_object_selection_scrollpaint(rct_window* w, rct_drawpi
                 gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
             }
 
-            int32_t width_limit = (w->widgets[WIDX_LIST].right - w->widgets[WIDX_LIST].left - x) / 2;
+            int32_t width_limit = (w->widgets[WIDX_LIST].right - w->widgets[WIDX_LIST].left - x);
```
AaronVanGeffen, line 1163:
```suggestion
            int32_t width_limit = w->widgets[WIDX_LIST].right - w->widgets[WIDX_LIST].left - x;
```

PR #11641, opened 2020-05-05
Fix #11640: Object shows with blank description

### LINKED ISSUE #11640: Object shows with blank description
**OS:** Ubuntu 20.04
**Version:** 0.2.6+
**Commit/Build:** b03fbc086da9799be356182db3ba9f13b13bc625

The problem is as follows: the attached object (small scenery) has a proper description for American English, but the object creator has lazily put a space into the boxes for all other languages. This was fixed by @duncanspumpkin in 2015, but has since regressed.

Original bug report was here: https://github.com/OpenRCT2/OpenRCT2/issues/1533

[TTWDRC06.zip](https://github.com/OpenRCT2/OpenRCT2/files/4580205/TTWDRC06.zip)

Gymnasiast:


Gymnasiast:
No, the fallback system is already in place. The problem was that it would still _save_ the bad translations, which meant that when falling back, it would just pick the first translation available, even if it was one of the bad ones.

PR #11646, opened 2020-05-05
Fix #11644: Powered launch with blocks not available

### LINKED ISSUE #11644: Powered Launch (Block Section mode) missing from selectable operating modes when show all operating modes cheat is active
**OS:** [Windows 10]
**Version:** [0.2.6]
**Commit/Build:** [b03fbc0]

Powered launch (Block sectioned mode) can not be selected on rides when the Show all operating modes cheat is active.
image

It can only be selected on the LIM-launched coaster, and only when the cheat is turned off.
image

Gymnasiast:


Gymnasiast:
It was a little oversight from the RTD migration, which required my to put `1ULL <<` in front of all the supported ride modes. I forgot to do so for block sectioned mode.

PR #11650, opened 2020-05-05
Use named flags in calls to sub_6C683D()

Gymnasiast:


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2621,7 +2621,8 @@ void sub_6C94D8()
             type = _currentTrackPieceType;
             auto newCoords = sub_6C683D(
                 { _currentTrackBegin, static_cast<Direction>(direction) }, type, 0, nullptr,
-                _currentTrackSelectionFlags & TRACK_SELECTION_FLAG_ARROW ? 2 : 1);
+                _currentTrackSelectionFlags & TRACK_SELECTION_FLAG_ARROW ? TRACK_ELEMENT_SET_HIGHLIGHT_TRUE
```
AaronVanGeffen, line 2624: Can I suggest this parameter be split off into a const local variable? I really don't like ternaries in function calls, especially when they're spanning several lines.

AaronVanGeffen:
I really like the ongoing effort to name all these magic constants. However, I have left a comment to address a minor code style issue that becomes more pressing due to the use of these constants.

src/openrct2-ui/windows/RideConstruction.cpp
Gymnasiast: Neither do I ;-)

PR #11651, opened 2020-05-05
Fix #1013: negative length in Ride window

### LINKED ISSUE #1013: Integer overflow on track length
http://www.reddit.com/r/rct/comments/32ev9w/2_i_decided_to_make_a_long_ride_but_i_made_it_so/

Gymnasiast:
This changes length drawing not to cast to a signed 16-bit integer. It can now represent the whole unsigned 16-bit integer range.

The change on line 5834 of src/openrct2-ui/windows/Ride.cpp addresses what looks like a refactoring bug to me.


duncanspumpkin:
Doesn't line 5839 need to be changed you are now using a 32 bit field but only pushing in 16 bits.

Gymnasiast:
> Edit. no I see what you are doing thats a bit of a hack. Can't we do something better.

Since COMMA16 is always signed, and because having thousands separators is nice, there isn't much choice other than COMMA32, as far as the string itself is concerned. 

As for the rest of the code: the tidy way would be to change all the code that uses length to use int32_t, or even a typedef of that. But there are a lot of format arg calls to modify, and the potential to muck things up is very big, at least until the format arg refactor is done.

PR #11653, opened 2020-05-05
Introduce WINDOW_SHIM define

Gymnasiast:
Taken from https://github.com/OpenRCT2/OpenRCT2/pull/10133. It does not include the ifdefs to move the close box to the left, as that bit needs a rethink anyway.

### DIFF
--- a/src/openrct2-ui/windows/ClearScenery.cpp
+++ b/src/openrct2-ui/windows/ClearScenery.cpp
@@ -34,7 +34,7 @@ enum WINDOW_CLEAR_SCENERY_WIDGET_IDX {
 #define WH 93
 
 static rct_widget window_clear_scenery_widgets[] = {
-    WINDOW_SKELETON,
+    { WINDOW_SKELETON(WINDOW_TITLE, WW, WH) },
     { WWT_IMGBTN,   0,  27, 70, 17, 48, SPR_LAND_TOOL_SIZE_0,                       STR_NONE },                         // preview box
     { WWT_TRNBTN,   1,  28, 43, 18, 33, IMAGE_TYPE_REMAP | SPR_LAND_TOOL_DECREASE,        STR_ADJUST_SMALLER_LAND_TIP },      // decrement size
     { WWT_TRNBTN,   1,  54, 69, 32, 47, IMAGE_TYPE_REMAP | SPR_LAND_TOOL_INCREASE,        STR_ADJUST_LARGER_LAND_TIP },       // increment size
--- a/src/openrct2-ui/windows/ClearScenery.cpp
+++ b/src/openrct2-ui/windows/ClearScenery.cpp
@@ -30,8 +30,8 @@ enum WINDOW_CLEAR_SCENERY_WIDGET_IDX {
 };
 
 static constexpr const rct_string_id WINDOW_TITLE = STR_CLEAR_SCENERY;
-static constexpr const int32_t WW = 102;
-static constexpr const int32_t WH = 93;
+static constexpr const int32_t WW = 98;
+static constexpr const int32_t WH = 94;
 
 static rct_widget window_clear_scenery_widgets[] = {
     WINDOW_SHIM(WINDOW_TITLE, WW, WH),
@@ -99,7 +99,7 @@ rct_window* window_clear_scenery_open()
         return window;
 
     window = window_create(
-        ScreenCoordsXY(context_get_width() - 102, 29), 98, 94, &window_clear_scenery_events, WC_CLEAR_SCENERY, 0);
+        ScreenCoordsXY(context_get_width() - WW, 29), WW, WH, &window_clear_scenery_events, WC_CLEAR_SCENERY, 0);
     window->widgets = window_clear_scenery_widgets;
     window->enabled_widgets = (1 << WIDX_CLOSE) | (1 << WIDX_INCREMENT) | (1 << WIDX_DECREMENT) | (1 << WIDX_PREVIEW)
         | (1 << WIDX_SMALL_SCENERY) | (1 << WIDX_LARGE_SCENERY) | (1 << WIDX_FOOTPATH);
--- a/src/openrct2-ui/windows/ClearScenery.cpp
+++ b/src/openrct2-ui/windows/ClearScenery.cpp
@@ -98,7 +98,8 @@ rct_window* window_clear_scenery_open()
     if (window != nullptr)
         return window;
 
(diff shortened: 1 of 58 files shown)


src/openrct2-ui/windows/SavePrompt.cpp
```diff
@@ -53,7 +55,7 @@ enum WINDOW_QUIT_PROMPT_WIDGET_IDX {
 static rct_widget window_quit_prompt_widgets[] = {
     { WWT_FRAME,            0,  0,      176,    0,  37, STR_NONE,                   STR_NONE },                 // panel / background
     { WWT_CAPTION,          0,  1,      175,    1,  14, STR_QUIT_GAME_PROMPT_TITLE, STR_WINDOW_TITLE_TIP },     // title bar
-    { WWT_CLOSEBOX,         0,  164,    174,    2,  13, STR_CLOSE_X_WHITE,          STR_CLOSE_WINDOW_TIP },     // close x button
+    { WINDOW_SKELETON(WINDOW_TITLE, WW, WH) },     // close x button
```
Gymnasiast, line 58: Needs replacing of all three.

Also needs a define for a _white_ X.

src/openrct2-ui/windows/Window.h
```diff
@@ -190,3 +190,11 @@ void window_tooltip_open(rct_window* widgetWindow, rct_widgetindex widgetIndex,
 void window_tooltip_close();
 
 rct_window* window_scenery_scatter_open();
+
+// clang-format off
+// close button on right of window
+#define WINDOW_SKELETON(TITLE, WIDTH, HEIGHT) \
+      WWT_FRAME,    0,  0,          WIDTH - 1, 0, HEIGHT, 0xFFFFFFFF,  STR_NONE }, \
+    { WWT_CAPTION,  0,  1,          WIDTH - 2, 1, 14,     TITLE,       STR_WINDOW_TITLE_TIP }, \
+    { WWT_CLOSEBOX, 0,  WIDTH - 13, WIDTH - 3, 2, 13,     STR_CLOSE_X, STR_CLOSE_WINDOW_TIP
```
Broxzier, line 199: Wouldn't it be clearer to encapsulate this all with `{}`? That way it looks less like it's a define for just one widget, and is more in line with how `SPINNER_WIDGETS` works.

Gymnasiast: Hm, I changed it because I thought this was more in line with how the rest of the code did it :').

I have to admit I did overlook SPINNER_WIDGET. Anyway, I will change it to encapsulate it with `{}`.

IntelOrca:
For moving the close box. Well all windows (some probably already do) should be calling a common method for resizing those skeleton (or shim as I call them) widgets. In that common method, it could change the position of the close button based on user preference.

Gymnasiast:
It was my intention to create such a function, yes, especially since it avoids a lot of code duplication, even without the left/right handling. Is it ok if I leave that for another PR, or do you want me to address it in this one?

Also, if you think "shim" is a better name than "skeleton": I have no problem changing it.

IntelOrca:
> Is it ok if I leave that for another PR?

Yes, I didn't expect you to do this now.

src/openrct2-ui/windows/ClearScenery.cpp
```diff
@@ -97,7 +99,7 @@ rct_window* window_clear_scenery_open()
         return window;
 
     window = window_create(
-        ScreenCoordsXY(context_get_width() - 98, 29), 98, 94, &window_clear_scenery_events, WC_CLEAR_SCENERY, 0);
+        ScreenCoordsXY(context_get_width() - 102, 29), 98, 94, &window_clear_scenery_events, WC_CLEAR_SCENERY, 0);
```
Gymnasiast, line 102:
```suggestion
        ScreenCoordsXY(context_get_width() - WW, 29), WW, WH, &window_clear_scenery_events, WC_CLEAR_SCENERY, 0);
```

src/openrct2-ui/windows/Finances.cpp
```diff
@@ -27,6 +27,10 @@
 #include <openrct2/sprites.h>
 #include <openrct2/world/Park.h>
 
+static constexpr const rct_string_id WINDOW_TITLE = 0;
+static constexpr const int32_t WH = 0;
```
Gymnasiast, line 31: Not correct...

src/openrct2-ui/windows/SavePrompt.cpp
```diff
@@ -51,9 +53,7 @@ enum WINDOW_QUIT_PROMPT_WIDGET_IDX {
 };
 
 static rct_widget window_quit_prompt_widgets[] = {
-    { WWT_FRAME,            0,  0,      176,    0,  37, STR_NONE,                   STR_NONE },                 // panel / background
-    { WWT_CAPTION,          0,  1,      175,    1,  14, STR_QUIT_GAME_PROMPT_TITLE, STR_WINDOW_TITLE_TIP },     // title bar
-    { WWT_CLOSEBOX,         0,  164,    174,    2,  13, STR_CLOSE_X_WHITE,          STR_CLOSE_WINDOW_TIP },     // close x button
+    WINDOW_SHIM_WHITE(WINDOW_TITLE, WW, WH),
```
Gymnasiast, line 56: Needs own WW (177) and WH (38).

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -153,9 +154,7 @@ validate_global_widx(WC_SCENERY, WIDX_SCENERY_ROTATE_OBJECTS_BUTTON);
 validate_global_widx(WC_SCENERY, WIDX_SCENERY_EYEDROPPER_BUTTON);
 
 static rct_widget window_scenery_widgets[] = {
-    { WWT_FRAME, 0, 0, 633, 0, 141, 0xFFFFFFFF, STR_NONE },                             // 1                0x009DE298
-    { WWT_CAPTION, 0, 1, 632, 1, 14, 0xFFFFFFFF, STR_WINDOW_TITLE_TIP },                // 2                0x009DE2A8
-    { WWT_CLOSEBOX, 0, 621, 631, 2, 13, STR_CLOSE_X, STR_CLOSE_WINDOW_TIP },            // 4                0x009DE2B8
+    WINDOW_SHIM(WINDOW_TITLE, WINDOW_SCENERY_WIDTH, WINDOW_SCENERY_HEIGHT),
```
Gymnasiast, line 157: height was 142 originally. Check if this causes a problem.

Issue #11569, opened 2020-05-01
Use ScreenCoordsXY on gfx_draw_string_right


tupaschoal:
Create an overload of `gfx_draw_string_right` on `\src\openrct2\drawing\Text.cpp` with the same name, but that takes a `const ScreenCoordsXY&` object instead of an `int32_t x` and and an `int32_t y` and replace the calls to the `{x ,y}` pair by the one using `ScreenCoordsXY`.

Don't try to do all calls at once, unless it doesn't occur a lot, as this might lead to a huge PR. When we stop using the old function, we can drop it entirely in favour of the `ScreenCoordsXY` alternative.

Aryaman73:
Hey! I'd like to take this issue on. It's my first time contributing to OpenRCT2, so I was wondering if there were any documents other than [this one](https://github.com/OpenRCT2/OpenRCT2/blob/develop/CONTRIBUTING.md) about contributing to this project. (I couldn't find any details about building the code and testing it, in particular). 

 Awesome project btw 😄

tupaschoal:
There are some things about compiling in our main readme:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/readme.md#32-compiling-and-running

There are also links on our wiki such as [coding style](https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style) and [Commit messages](https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages)

Aryaman73:
@tupaschoal Thanks for the merge, and helping me out! What else needs to be done for this? I can't find any other calls to this function,

tupaschoal:
Really? If none is calling the old function anymore, just delete it and submit a pr. It should fail to compile if there are users still

Aryaman73:
Sorry, not sure what I did wrong earlier but you're right, there are definitely other calls to the function. As for fixing the other calls, the change should pretty much just include changing: 
 ` gfx_draw_string_right(dpi, stringId, nullptr, COLOUR_WHITE,  w->windowPos.x + w->width - 5, y );`
to this:
 ` gfx_draw_string_right(dpi, stringId, nullptr, COLOUR_WHITE, {w->windowPos.x + w->width - 5, y });`

Or is there more to it?

duncanspumpkin:
That's pretty much all you need to change. In some places you might have a ScreencoordsXY already so you obviously wouldn't need to create one for that.

Aryaman73:
[...]

PR #11654, opened 2020-05-06
Refactored gfx_draw_string_right (Issue 11569)

Aryaman73:
Refactored gfx_draw_string_right to use ScreenCoordsXY in Text.cpp and Drawing.h. Changed old function to call new function instead.

### DIFF
--- a/src/openrct2/drawing/Text.cpp
+++ b/src/openrct2/drawing/Text.cpp
@@ -164,9 +164,13 @@ void gfx_draw_string_centred(
 
 void gfx_draw_string_right(rct_drawpixelinfo* dpi, rct_string_id format, void* args, uint8_t colour, int32_t x, int32_t y)
 {
-    DrawTextCompat(dpi, x, y, format, args, colour, TextAlignment::RIGHT);
+    gfx_draw_string_right(dpi, format, args, colour, { x, y });
 }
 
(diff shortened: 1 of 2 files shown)


Aryaman73:
Issue: https://github.com/OpenRCT2/OpenRCT2/issues/11569
(Sorry, not sure if there's a better way to link issues with PR)

src/openrct2/drawing/Text.cpp
```diff
@@ -164,9 +164,13 @@ void gfx_draw_string_centred(
 
 void gfx_draw_string_right(rct_drawpixelinfo* dpi, rct_string_id format, void* args, uint8_t colour, int32_t x, int32_t y)
 {
-    DrawTextCompat(dpi, x, y, format, args, colour, TextAlignment::RIGHT);
+    gfx_draw_string_right(dpi, format, args, colour, { x, y });
 }
 
+void gfx_draw_string_right(rct_drawpixelinfo* dpi, rct_string_id format, void* args, uint8_t colour, const ScreenCoordsXY& coords)
```
tupaschoal, line 170:
```suggestion
void gfx_draw_string_right(
    rct_drawpixelinfo* dpi, rct_string_id format, void* args, uint8_t colour, const ScreenCoordsXY& coords)
```
The only thing missing is that the format is broken, you need to apply this change.

PR #11662, opened 2020-05-07
Fix #11632: Ride slot is not freed when demolishing it

### LINKED ISSUE #11632: Removing a ride without demolishing it doesn't free a ride slot
**OS:** Windows 10 64 bit
**Version:** 0.2.6 develop (but also previous releases)
**Commit/Build:** 9158b55

When the "remove ride" permission isn't granted to a player, removing every element of the ride will hide it in the in-game ride list, but the slot isn't freed up.

**Steps to reproduce:**
1. Connect to a multiplayer server where build permissions are granted but demolish permissions are not
2. Click the "build custom design" button for a ride type that supports it, but don't build anything.
3. Close the construction window. Observe the message that appears.
4. Repeat until all available ride slots are used up.

**Dump file**

**Screenshots / Video:**
permission denied

too many rides

**Save game:**

frutiemax:
The server now reallocate ride ids which are either invalid or empty i.e. 0 station and 0 in track length.

### DIFF
--- a/src/openrct2/actions/RideSetSetting.hpp
+++ b/src/openrct2/actions/RideSetSetting.hpp
@@ -81,7 +81,7 @@ public:
                     return MakeResult(GA_ERROR::DISALLOWED, STR_CANT_CHANGE_OPERATING_MODE, STR_MUST_BE_CLOSED_FIRST);
                 }
 
-                if (!ride_is_mode_valid(ride))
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
This is a pretty horrible solution. A better way would be just to override the permission check for ride demolish.

frutiemax:
Override the permission check on the client's side?

PR #11665, opened 2020-05-07
[Plugin] Split object identifier and add more ride object fields

IntelOrca:
Discussed here:
https://github.com/OpenRCT2/OpenRCT2/pull/11489#issuecomment-620014113

And most of the ride fields. @spacek531

### DIFF
--- a/src/openrct2/localisation/LanguagePack.cpp
+++ b/src/openrct2/localisation/LanguagePack.cpp
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2014-2019 OpenRCT2 developers
+ * Copyright (c) 2014-2020 OpenRCT2 developers
  *
  * For a complete list of all authors, please refer to contributors.md
  * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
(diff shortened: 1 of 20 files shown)


src/openrct2/object/ObjectFactory.cpp
```diff
@@ -104,15 +104,20 @@ class ReadObjectContext : public IReadObjectContext
     }
 
     ReadObjectContext(
-        IObjectRepository& objectRepository, const std::string& objectName, bool loadImages,
+        IObjectRepository& objectRepository, const std::string& identifier, bool loadImages,
         const IFileDataRetriever* fileDataRetriever)
         : _objectRepository(objectRepository)
         , _fileDataRetriever(fileDataRetriever)
-        , _objectName(objectName)
+        , _identifier(identifier)
```
Gymnasiast, line 111: Is this the new identifier, or the legacy one?

src/openrct2/localisation/LanguagePack.cpp
```diff
@@ -198,15 +198,14 @@ class LanguagePack final : public ILanguagePack
         }
     }
 
-    rct_string_id GetObjectOverrideStringId(const char* objectIdentifier, uint8_t index) override
+    rct_string_id GetObjectOverrideStringId(const std::string_view& objectIdentifier, uint8_t index) override
```
Gymnasiast, line 201: This takes a legacy identifier, right? If so: perhaps best to change the parameter name.

src/openrct2/object/ObjectFactory.cpp
IntelOrca: I think it is new for JSON, old for DAT, but we only use it for logging at the moment.

src/openrct2/localisation/LanguagePack.cpp
IntelOrca: sure

PR #11668, opened 2020-05-08
[Plugin] Add more stats to Peep interface

IntelOrca:
* Add id to the Entity interface.
* Add a few more stats to the Peep entity interface.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -655,6 +655,10 @@ declare global {
      */
     interface Entity {
         /**
+         * The entity index within the entity list.
+         */
+        readonly id: number;
+        /**
          * The type of entity, e.g. car, duck, litter, or peep.
          */
         readonly type: EntityType;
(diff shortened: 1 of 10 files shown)


distribution/openrct2.d.ts
```diff
@@ -599,14 +603,90 @@ declare global {
...
+
+        /**
+         * The target nausea value. Nausea will increase / decrease slowly towards this value.
+         */
+        nauseaTarget : number;
+
+        /**
+         * How hungry the guest is between 0 and 255. Lower is more hungry.
+         */
+        hunger: number;
```
Gymnasiast, line 654: A minor nitpick, but you seem to use inconsistent formatting (you use both `hunger: number;` and ``hunger : number;`).

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -158,6 +183,220 @@ namespace OpenRCT2::Scripting
...
+        }
+        void maxIntensity_set(uint8_t value)
+        {
+            ThrowIfGameStateNotMutable();
+            auto peep = GetPeep();
+            if (peep != nullptr)
+            {
+                auto minIntensity = peep->intensity & 0x0F;
+                auto maxIntensity = std::min<uint8_t>(value, 15);
+                peep->intensity = (maxIntensity << 4) | minIntensity;
```
Gymnasiast, line 366: I prefer if these implementation details were abstracted away by implementing them as methods on the `Guest` struct. The scripting layer should not have in-depth knowledge of implementation details of the Guest struct, IMO.

It is okay if you address this is another PR, but in that case: please create an issue as well.

duncanspumpkin:
What are we going to do for stats that share fields with staff stats?

IntelOrca:
@duncanspumpkin Not implemented any of those yet, but getEntity could always return a sub object: Guest or Staff which works fine. There might be a problem if you can change between guest and staff in the script.

src/openrct2/peep/Peep.h
```diff
@@ -541,6 +542,58 @@ struct rct_peep_thought
...
+    uint8_t _value{};
+
+public:
+    explicit IntensityRange(uint8_t value)
+        : _value(value)
+    {
+    }
+
+    IntensityRange(uint8_t min, uint8_t max)
+        : _value(std::min<uint8_t>(min, 15) | (std::max<uint8_t>(min, 15) << 4))
```
Gymnasiast, line 557:
```suggestion
        : _value(std::min<uint8_t>(min, 15) | (std::min<uint8_t>(max, 15) << 4))
```

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1528,12 +1528,12 @@ void window_guest_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
     y += LIST_ROW_HEIGHT;
 
     // Intensity
-    int32_t intensity = peep->intensity / 16;
+    auto intensity = peep->intensity.GetMaximum();
```
Gymnasiast, line 1531: Could you rename this variable to `maxIntensity`?

PR #11674, opened 2020-05-08
Fix #11632 : Rides ids get freed when the ride construction is aborted

### LINKED ISSUE #11632: Removing a ride without demolishing it doesn't free a ride slot

frutiemax:
Players with ride construction permissions can abort the construction window and launch a ride abort construction action to the server, which frees the ride ID for the future ones.


duncanspumpkin:
Can you please stick to one pr and not open and close new ones. It's much harder to track what comments have previously been said.

src/openrct2/actions/RideAbortConstructionAction.hpp
```diff
@@ -0,0 +1,342 @@
...
+
+using namespace OpenRCT2;
+
+DEFINE_GAME_ACTION(RideAbortConstructionAction, GAME_COMMAND_RIDE_ABORT_CONSTRUCTION, GameActionResult)
+{
+private:
+    NetworkRideId_t _rideIndex{ RideIdNewNull };
+
+public:
+    RideAbortConstructionAction()
```
duncanspumpkin, line 38: use `= default;` instead of an empty body.

src/openrct2/actions/RideAbortConstructionAction.hpp
```diff
@@ -0,0 +1,342 @@
...
+    {
+    }
+    RideAbortConstructionAction(ride_id_t rideIndex)
+        : _rideIndex(rideIndex)
+    {
+    }
+
+    uint32_t GetCooldownTime() const override
```
duncanspumpkin, line 46: Why do you need a cooldown on this?

duncanspumpkin:
I see this is just a copy of RideDemolishAction. That means a nefarious user could use this action to demolish any ride even if they only have creation permissions. This solution will also not work if a user disconnects. 

I think we will need to create a state on the server that records if a ride has been allocated to a user from opening the construction window (RideCreateAction). If that user closes the construction window (or disconnects) without building anything then the server can clean up by calling a RideDemolishAction on the ride. This would work similar to the peep pickup action so maybe have a look at that.

frutiemax:
How do you check if the client has closed the construction window?

duncanspumpkin:
I suspect we will still need to send some sort of game action for that so we can use this as a basis for that. In my head this is how I imagine it working:

1. Client requests a new ride id with RideCreateAction. This is assigned to the player within the server. Client has the construction window open from this point onwards.
- 2 a) Client closes the construction window after not creating anything. This game action `RideAbortConstructionAction` is sent. Server cleans up and frees the ride id. RideAbortConstructionAction is a seperate Permission that is always available with ride create and will only destroy a ride if it has no track otherwise it will fail. Server deregisters this ride id during this action.

- 2 b) Client closes the construction window after placing track. We will need some sort of command sent to the server to identify that this has happened. Server deregisters this ride id to the Client as it is in use.

- 2 c) Client disconnects or stops responding to packets. Server checks to see if registered ride id has track and if it does not sends out a `RideAbortConstructionAction`. Server deregisters this ride id during this action.

frutiemax:
duncan, I could just modify the RideAbortConstructionAction so it only deletes the ride when there is zero station and zero track length. This RideAbortConstructionAction is called when the construction window is closed.

duncanspumpkin:
That would be a good start. I think we still need to register the ride though eventually as you could issue other commands in-between opening the window and closing it. I had a thought earlier i think we can just do a nested call to ride demolish. I think then it would only check the new permission.

Gymnasiast:
@frutiemax Do you plan to continue work on this? We're fine either way but we would like to know.

frutiemax:
I'm not sure how to completely fix the problem tbh.

PR #11682, opened 2020-05-09
[Plugin] Ensure modified entities and tile elements are invalidated

IntelOrca:


src/openrct2/scripting/ScEntity.hpp
```diff
@@ -77,7 +77,9 @@ namespace OpenRCT2::Scripting
             auto entity = GetEntity();
             if (entity != nullptr)
             {
+                invalidate_sprite_2(entity);
```
tupaschoal, line 80: Why does it need to invalidate before? (Where most only invalidate after?)

Sounds like a good case for a small RAII class

Broxzier: Because the invalidate function has no way to know where it was before, making it unable to invalidate that portion of the screen.

tupaschoal: Hmm, I see. I wish we had used RAII here, it would be neat

IntelOrca: I don't think that RAII fits this use case that well. It is a very particular edge case with any kind of movement of stuff in the game, e.g. windows or entities. Ideally it should be in the sprite move function itself or rather a parameter or another method called moveandinvalidate.

PR #11686, opened 2020-05-09
Enable hypercoasters' boosters by default

Fred-1044276:
This updated RideGroupManager.cpp includes the following changes:
- The Hyper-Twister now has a maximum in-game support limit of 67 units (335ft/100.5m). This is because there is a real life B&M Hyper Coaster named Fury 325, which stands at 325 ft (99.1 m) tall, and the Hyper-Twister is based on the B&M Hyper Coaster.
- The Hypercoaster can now use boosters without the need to turn on "Enable all drawable track pieces". This change was done for RCT1 parity because in RCT1, the Hypercoaster Trains can be put on a Steel Corkscrew Roller Coaster with boosters (but no inversions).
- The Hyper-Twister now has access to boosters and launched lift hills by default. Like with the Hypercoaster, this change was also done for RCT1 parity as in RCT1, you can build a Steel Twister Roller Coaster, select the Roller Coaster Train (Non-Looping) (the RCT1 version of the Hyper-Twister Trains (wide cars)), and still run it on a track with boosters and launched lift hills.
- The Monster Trucks variant of the Car Ride can now be built with Spinning Tunnel sections.
- The copyright notices have been updated to read 2014-2020 instead of 2014-2019.

Regarding the boosters on the Hypercoaster and Hyper-Twister, although they may not look realistic, RCT1 still made the boosters usable on non-inverting train types, so why not make the Hypercoaster and Hyper-Twister have access to boosters (and launched lift hills for the latter) by default? The answer is: RCT1 parity.

### DIFF
--- a/src/openrct2/ride/RideGroupManager.cpp
+++ b/src/openrct2/ride/RideGroupManager.cpp
@@ -69,7 +69,7 @@ static constexpr const RideGroup ride_group_monster_trucks = {
 
 static constexpr const RideGroup ride_group_steel_twister_rc = {
     /*.RideType =*/RIDE_TYPE_TWISTER_ROLLER_COASTER,
-    /*.MaximumHeight =*/42,
+    /*.MaximumHeight =*/40,
     /*.AvailableTrackPieces =*/(1ULL << TRACK_FLAT) | (1ULL << TRACK_STRAIGHT) | (1ULL << TRACK_STATION_END)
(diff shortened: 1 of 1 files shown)


Gymnasiast:
All of the ride groups are cases where I **deliberately** deviated from RCT1 parity because RCT1 parity would have been utterly unrealistic. It is certainly not the be-all and end-all you make it out to be.

As for the spinning tunnel on Monster Trucks: have you actually tried what that looks like? The vehicles are much bigger than the spinning tunnel.

Also, the existence of a taller B&M ipso facto does not justify raising that limit. That at least needs discussion in order not to upset balance for non-cheating players.

Since that encompasses all of your changes, I'm afraid I have no other option than to close this pull request.

PR #11694, opened 2020-05-09
Name some tunnel constants

Gymnasiast:


Gymnasiast:
Yeah, I wasn't going to replace all the instances manually :P

PR #11700, opened 2020-05-10
[Plugin] Add support for SetCheatAction and cheat retrieval

IntelOrca:


src/openrct2/scripting/ScCheats.hpp
```diff
@@ -0,0 +1,326 @@
...
+
+#pragma once
+
+#ifdef ENABLE_SCRIPTING
+
+#    include "../Cheats.h"
+#    include "Duktape.hpp"
+#    include "ScriptEngine.h"
+
+namespace OpenRCT2::Scripting
```
tupaschoal, line 18: Why not?
```suggestion
using OpenRCT2::Scripting;
```

Then you won't have to open a huge namespace and with the extra 4 spaces maybe somethings won't break in as many lines hahaha

IntelOrca: You shouldn't do this in header files. Unless scoped which this is not.

tupaschoal: Ah, I hadn't noticed it was a header

tupaschoal:
There is one of the CI checks failing

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -369,7 +369,7 @@ ScriptEngine::ScriptEngine(InteractiveConsole& console, IPlatformEnvironment& en
 void ScriptEngine::Initialise()
 {
     auto ctx = (duk_context*)_context;
-    ScCheats::Register(ctx);
+    ScCheats::Register();
```
Broxzier, line 372: The `Register` function still needs it, right? Just not the constructor.

PR #11702, opened 2020-05-10
[Plugin] Add API to remove entities

IntelOrca:


src/openrct2/scripting/ScEntity.hpp
```diff
@@ -119,6 +119,40 @@ namespace OpenRCT2::Scripting
...
+                        else
+                        {
+                            peep->Remove();
+                        }
+                        break;
+                    }
+                    default:
+                        invalidate_sprite_2(entity);
+                        sprite_remove(entity);
+                        break;
```
tupaschoal, line 151: Do we want a `default` branch? There are only three other options, I tend to not like handling defaults because if something is added later it will automatically handle (and maybe incorrectly)

Issue #11210, opened 2020-04-02
Mechanic heading off into the air while repairing ride


AaronVanGeffen:
**OS:** Arch Linux
**Version:** 0.2.5+
**Commit/Build:** e1e8e2bec03e66cb278cde071a3cc8930ad7fe92

There seems to be an issue with mechanics repairing rides with small stations, or a separate station for entrance and exit, as is the case with 'Splashtastic' in the attached save file.

In this case, the mechanic entered the ride, but veered off into the air 'fixing' the ride, then off into the distance, presumably to find the exit.

Sacking the mechanic in question leaves the ride unfixed, and the next mechanic will be dispatched only to do the exact same thing.

**Screenshot**

![Okinawa Coast 2020-04-02 21-14-21](https://user-images.githubusercontent.com/604665/78292655-ab84c180-7527-11ea-9b8b-2a0476f21494.png)

**Save file**

[Japanese Coastal Reclaim.zip](https://github.com/OpenRCT2/OpenRCT2/files/4423162/Japanese.Coastal.Reclaim.zip)

AaronVanGeffen:
Duplicate of #8145; see also #9167.

PR #11704, opened 2020-05-10
Fix #11695. Mechanics walking to 0,0 on entrance only

### LINKED ISSUE #11695: Mechanics Walk to Tile 0,0 After Fixing Ride from an Entrance-Only Station
**OS:** Windows 10
**Version:** 0.2.6
**Commit/Build:** fa91b40 on develop

Working on a park with a bunch of rides that have entrances and exits in separate stations. I noticed that all of my mechanics would disappear as I was working on the park, then finally found them all hanging out off the map at tile 0,0. After a bunch of testing, I figured out that the mechanics were fixing the ride for a vehicle malfunction at the entrance station, and then walking all the way to the corner of the map. It looks to me like when the mechanics enter through an entrance in order to fix a ride, some flag or coordinate which indicates they should exit through the same way isn't set, and the game is putting them on tile 0,0.

I didn't want to share my park yet, so I made a small test park and was able to reproduce the issue using the junior coaster. This test park is attached.

- [ ] Reproducible in RCT2 (vanilla)? Don't think so, but don't know for sure
- [ ] Specific to multiplayer? No.

**Steps to reproduce:**
1.  Build a tracked ride with separate stations for entrance and exit.
2. Force (or wait for) a vehicle malfunction breakdown in the entrance station.
3. Wait for the mechanic to finish fixing the ride.
4. Watch the mechanic walk to coordinate 0,0 (off the map), and then "finish" fixing the ride and return to the "walking" state.

**Save game:**
[File_003.txt](https://github.com/OpenRCT2/OpenRCT2/files/4604992/File_003.txt)

duncanspumpkin:
Fix #11695. Mechanics walking to 0,0 on entrance only

Mistake made when refactoring that meant that null locations were converted into tile 0, 0. I've fixed the general case but it is preferred to try avoid using null states for coordinates if at all possible.

I haven't tested to see when the regression happened but as this would affect a whole bunch of different things I think it safe to assume these changes will fix a lot of them.

Note. This doesn't fix mechanics walking to middle of rides if a ride is fixed whilst the mechanic is heading to the ride.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -27,6 +27,7 @@
 - Fix: [#11455] Object Selection window cuts off scenery names.
 - Fix: [#11640] Objects with a blank description in one language do not fall back to other languages anymore.
 - Fix: [#11676] Spiral Roller Coaster has regular lift hill available.
+- Fix: [#11695] Mechanics walk to tile 0, 0 at entrance only stations when trying to fix them.
 - Fix: RCT1 scenarios have more items in the object list than are present in the park or the research list.
 - Improved: [#6530] Allow water and land height changes on park borders.
 - Improved: [#11390] Build hash written to screenshot metadata.
--- a/src/openrct2/world/Location.hpp
+++ b/src/openrct2/world/Location.hpp
@@ -254,6 +254,13 @@ struct TileCoordsXY
 
     CoordsXY ToCoordsXY() const
     {
+        if (isNull())
+        {
+            CoordsXY ret{};
+            ret.setNull();
+            return ret;
+        }
+
         return { x * COORDS_XY_STEP, y * COORDS_XY_STEP };
     }
 
@@ -404,6 +411,12 @@ struct TileCoordsXYZ : public TileCoordsXY
 
     CoordsXYZ ToCoordsXYZ() const
     {
+        if (isNull())
+        {
+            CoordsXYZ ret{};
+            ret.setNull();
+            return ret;
+        }
         return { x * COORDS_XY_STEP, y * COORDS_XY_STEP, z * COORDS_Z_STEP };
     }
 };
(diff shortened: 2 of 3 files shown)


AaronVanGeffen:
I wonder if this covers the case described in #11210.

(Leaving this here so I won't forget to test it.)

duncanspumpkin:
If it's a mechanic going out of bounds of a ride it is likely the same bug. If it's a mechanic going to somewhere in bounds of the ride it's the other bug.

distribution/changelog.txt
```diff
@@ -28,6 +28,7 @@
 - Fix: [#11640] Objects with a blank description in one language do not fall back to other languages anymore.
 - Fix: [#11676] Spiral Roller Coaster has regular lift hill available.
 - Fix: RCT1 scenarios have more items in the object list than are present in the park or the research list.
+- Fix: [#11695] Mechanics walk to tile 0, 0 at entrance only stations when trying to fix them.
```
AaronVanGeffen, line 31: Very minor, but these should probably be swapped:
```suggestion
- Fix: [#11695] Mechanics walk to tile 0, 0 at entrance only stations when trying to fix them.
- Fix: RCT1 scenarios have more items in the object list than are present in the park or the research list.
```

src/openrct2/world/Location.hpp
```diff
@@ -254,6 +254,11 @@ struct TileCoordsXY
 
     CoordsXY ToCoordsXY() const
     {
+        if (isNull())
+        {
+            return { COORDS_NULL, 0 };
```
Gymnasiast, line 259: I'd prefer if TileCoordsXY didn't handle implementation details of CoordsXY. CoordsXY should handle setting it to null itself.

duncanspumpkin: This has now been changed. I was going to create another method that returns a null from CoordsXY but decided that it would just make it more annoying when we come to remove all of the nulls.

PR #11712, opened 2020-05-10
[Plugin] Add more ride fields to API

IntelOrca:


MegaMech:
Please expose ride age if that's applicable to this PR.

Thanks

distribution/openrct2.d.ts
```diff
@@ -773,6 +773,16 @@ declare global {
...
          */
         buildDate: number;
+
+        /**
+         * The running cost of the ride billed every fortnight. Multiply this by 16 to get the cost per hour (~ 1 year).
+         */
+        runningCost: number;
+
+        /**
+         * How often the ride should inspected by a mechanic.
```
AaronVanGeffen, line 783:
```suggestion
         * How often the ride should be inspected by a mechanic.
```

src/openrct2/ride/Ride.cpp
```diff
@@ -891,7 +927,9 @@ void ride_init_all()
 void reset_all_ride_build_dates()
 {
     for (auto& ride : GetRideManager())
-        ride.build_date -= gDateMonthsElapsed;
+    {
+        ride.build_date = 0;
```
Gymnasiast, line 931: Doesn't this reset the build date to 1 March, Year 1?

IntelOrca: Okay, now I understand what the code was trying to do.

IntelOrca: Turns out this method is only called in scenario_start where `gDateMonthsElapsed` is set to 0 anyway. I have fixed it nonetheless.

src/openrct2/scripting/ScRide.hpp
```diff
@@ -131,6 +546,79 @@ namespace OpenRCT2::Scripting
             }
         }
 
+        int16_t buildDate_get() const
```
duncanspumpkin, line 549: Shouldn't we be getting age rather than build date as this value always needs to be combined with the current month to get a useful value

IntelOrca: At the same time, setting it would then need the reverse calculating in the plugin. Should I just add a new field for age?

Gymnasiast: I think what you did is fine.

Gymnasiast: I think what you did is fine.

IntelOrca: Okay I won't remove this. If @duncanspumpkin thinks there should also be age then I will add that as an additional field. It is possible to subtract elapsedmonths in the script though.

duncanspumpkin: Just add a note then that this is in terms of months and to use with elapsed months to get the age/time since last refurbish.

Gymnasiast: Would it be possible to create a JS getter for the age in addition to this buildDate field?

IntelOrca: Yes, would you want a setter as well?

src/openrct2/scripting/ScRide.hpp
```diff
@@ -56,6 +211,24 @@ namespace OpenRCT2::Scripting
             return ride != nullptr ? ride->type : 0;
         }
 
+        std::string classification_get() const
```
mgovea, line 214: Small oopsie here I discovered when playing with my plugin. While this function works just fine...

src/openrct2/scripting/ScRide.hpp
```diff
@@ -143,10 +637,31 @@ namespace OpenRCT2::Scripting
             dukglue_register_property(ctx, &ScRide::object_get, nullptr, "object");
             dukglue_register_property(ctx, &ScRide::type_get, nullptr, "type");
             dukglue_register_property(ctx, &ScRide::name_get, &ScRide::name_set, "name");
```
mgovea, line 639: ... it's never registered down here. Adding a quick
```
dukglue_register_property(ctx,` &ScRide::classification_get, nullptr, "classification");
```
will fix it up. If you want these to be in the same order as the function declarations above, then put it between `type` and `name`.

IntelOrca: Thanks, have fixed it!

mgovea:
Curious what the blockers are for this. `buildDate` and `readonly age` were added, and there wasn't any followup after that.

tupaschoal:
This PR breaks the ride age, see #12764

PR #11714, opened 2020-05-11
Part of #9473: Create Vehicle::KillPassengers()

hdpoliveira:
Converted from vehicle_kill_all_passengers

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -49,8 +49,6 @@ static bool vehicle_boat_is_location_accessible(const CoordsXYZ& location);
 static bool vehicle_update_motion_collision_detection(
     Vehicle* vehicle, int16_t x, int16_t y, int16_t z, uint16_t* otherVehicleIndex);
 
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -5258,42 +5240,58 @@ static void vehicle_kill_all_passengers(Vehicle* vehicle)
...
+    if (num_peeps != next_free_seat)
+        return;
 
-        if (curVehicle->num_peeps == 0)
-            continue;
+    if (num_peeps == 0)
+        return;
 
-        for (uint8_t i = 0; i < curVehicle->num_peeps; i++)
+    for (uint8_t i = 0; i < num_peeps; i++)
```
Gymnasiast, line 5280:
```suggestion
    for (auto i = 0; i < num_peeps; i++)
```

duncanspumpkin:
Causes https://github.com/OpenRCT2/OpenRCT2/issues/13222 don't decrement the number of riders the peep will handle that iteself.

duncanspumpkin:
After a second look i don't think it is related to the rider count issue

PR #11715, opened 2020-05-11
Create NewsItemQueue class

hdpoliveira:
With this change, the implementation of the News queue become encapsulated, and we are free to change the implementation away from C-array (to possibly std::array).

It also splits the two queues that were present, making the code clearer.

### DIFF
--- a/src/openrct2/management/NewsItem.cpp
+++ b/src/openrct2/management/NewsItem.cpp
@@ -76,6 +76,11 @@ NewsItem* news_item_get(int32_t index)
 
 NewsItem& NewsItemQueue::operator[](size_t index)
 {
+    return const_cast<NewsItem&>(const_cast<const NewsItemQueue&>(*this)[index]);
+}
+
+const NewsItem& NewsItemQueue::operator[](size_t index) const
+{
     if (index < NEWS_ITEM_HISTORY_START)
         return Recent[index];
     else
--- a/src/openrct2/management/NewsItem.cpp
+++ b/src/openrct2/management/NewsItem.cpp
@@ -51,12 +51,12 @@ const NewsItem& NewsItemQueue::Current() const
 
 NewsItem& NewsItemQueue::Oldest()
 {
-    return Old[0];
+    return Archived[0];
 }
 
 const NewsItem& NewsItemQueue::Oldest() const
 {
-    return Old[0];
+    return Archived[0];
 }
 
 bool news_item_is_valid_idx(int32_t index)
@@ -79,7 +79,7 @@ NewsItem& NewsItemQueue::operator[](size_t index)
     if (index < NEWS_ITEM_HISTORY_START)
         return Recent[index];
     else
-        return Old[index - NEWS_ITEM_HISTORY_START];
+        return Archived[index - NEWS_ITEM_HISTORY_START];
 }
 
 NewsItem* NewsItemQueue::At(int32_t index)
@@ -164,7 +164,7 @@ int32_t NewsItemQueue::RemoveTime() const
     return 320;
 }
 
-bool NewsItemQueue::IsCurrentOld() const
+bool NewsItemQueue::CurrentShouldBeArchived() const
 {
     return Current().Ticks >= RemoveTime();
 }
@@ -186,8 +186,8 @@ void news_item_update_current()
     news_item_tick_current();
 
     // Removal of current news item
-    if (gNewsItems.IsCurrentOld())
-        gNewsItems.MoveCurrentToOld();
+    if (gNewsItems.CurrentShouldBeArchived())
+        gNewsItems.ArchiveCurrent();
 }
 
 /**
@@ -196,16 +196,16 @@ void news_item_update_current()
  */
 void news_item_close_current()
 {
-    gNewsItems.MoveCurrentToOld();
+    gNewsItems.ArchiveCurrent();
 }
 
-void NewsItemQueue::MoveCurrentToOld()
+void NewsItemQueue::ArchiveCurrent()
 {
     // Check if there is a current message
     if (IsEmpty())
         return;
 
-    AppendToOld(Current());
+    AppendToArchive(Current());
 
     // Invalidate the news window
     window_invalidate_by_class(WC_RECENT_NEWS);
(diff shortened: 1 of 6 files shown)


src/openrct2/management/NewsItem.h
```diff
@@ -52,13 +52,75 @@ struct NewsItem
...
+    {
+        return Recent[0];
+    }
+    NewsItem& Oldest()
+    {
+        return Old[0];
+    }
+    const NewsItem& Oldest() const
+    {
+        return Old[0];
```
janisozaur, line 89: I'd prefer to move them to cpp for sake of compile times.

src/openrct2/management/NewsItem.cpp
```diff
@@ -52,10 +50,23 @@ bool news_item_is_valid_idx(int32_t index)
...
+
+NewsItem& NewsItemQueue::operator[](size_t index)
+{
+    if (index < MAX_RECENT_NEWS_ITEMS)
+        return Recent[index];
+    else
+        return Old[index - MAX_RECENT_NEWS_ITEMS];
+}
+
+NewsItem* NewsItemQueue::At(int32_t index)
```
janisozaur, line 65: Looks like this could be const

src/openrct2/management/NewsItem.cpp
```diff
@@ -295,18 +309,17 @@ std::optional<CoordsXYZ> news_item_get_subject_location(int32_t type, int32_t su
     return subjectLoc;
 }
 
-static NewsItem* news_item_first_open_queue_slot()
+NewsItem* NewsItemQueue::FirstOpenSlot()
```
janisozaur, line 312: The name suggests const, but the implementation actually mutates. Can you come up with better name for this method?

src/openrct2/management/NewsItem.h
hdpoliveira: Will do.
The reason I put them here is to allow inlining, and since these are straightforward I don't think they'd get in the way of compilation.

src/openrct2/management/NewsItem.cpp
hdpoliveira: It could but it shouldn't, because it returns a non-const access to an internal member.
I could create a const version of this method, if you'd like.

src/openrct2/management/NewsItem.cpp
hdpoliveira: Changed to FirstOpenOrCreateSlot

src/openrct2/management/NewsItem.h
janisozaur: Inlining… member functions? I doubt this would be done at compilation stage, but have no evidence to back it up. LTO on the other hand… that's not affected by the code being placed in the include or source file, though.

janisozaur: It seems I'm in the wrong about that: https://isocpp.org/wiki/faq/inline-functions#inline-member-fns-more

src/openrct2/management/NewsItem.cpp
janisozaur: The queue itself is immutable. Let's take the example of shared ptr: https://en.cppreference.com/w/cpp/memory/shared_ptr/get

Returns a non-const pointer from an immutable store, thus the method is marked const.

src/openrct2/management/NewsItem.cpp
janisozaur: That's an odd name, it reads as if "open" was a verb in this context. I'd suggest FirstOpenOrNewSlot

src/openrct2/management/NewsItem.cpp
hdpoliveira: This is with respect to bitwise constness vs logical constness (see chapter 3 in Effective C++ by Scott Meyers). 

> What does it mean for a member function to be const? There are two
prevailing notions: bitwise constness (also known as physical constness) and logical constness.
> ...
> Surely there is something wrong when you create a constant object with a particular value and you invoke only const member functions on it, yet you still change its value!
> This leads to the notion of logical constness. Adherents to this philosophy argue that a const member function might modify some of the bits in the object on which it's invoked, but only in ways that clients cannot detect.

The reason the committee chose to make shared_ptr::get() const is simply because if it wasn't the API would be much harder to use. The 'get' method already is a "let me do dangerous things" method, so there is no point in protecting it.

If you take STL containers (for example [std::vector](https://en.cppreference.com/w/cpp/container/vector/operator_at)) you will see all of them overload operator at with non-const and const versions, return non-const or const references respectively.

Is it ok if I add a const version of the method?

janisozaur: > Is it ok if I add a const version of the method?

Yes, that's fine.

I initially checked vector as well, but discarded it as here we deal explicitly with pointers to items, so I went to first thing that returns pointers I could think of. I do like your argumentation and let's do the overload then.

src/openrct2/management/NewsItem.h
```diff
@@ -52,14 +52,63 @@ struct NewsItem
...
+            p(newsItem);
+        }
+    }
+
+private:
+    int32_t RemoveTime() const;
+    void AppendToOld(NewsItem& item);
+
+    NewsItem Recent[NEWS_ITEM_HISTORY_START];
+    NewsItem Old[MAX_OLD_NEWS_ITEMS];
```
IntelOrca, line 108: I think these should be changed to vectors. All uses of `NEWS_ITEM_HISTORY_START` and `MAX_OLD_NEWS_ITEMS` should be removed with the exception of internal limits to ensure it never adds more than the current limit,  `NEWS_ITEM_HISTORY_START` and `MAX_OLD_NEWS_ITEMS` to the vectors.

IntelOrca:
In the plugin API, I refer to the news items as "archived" rather than "old", I think is a bit more meaningful.

src/openrct2/management/NewsItem.h
```diff
@@ -52,14 +52,63 @@ struct NewsItem
     uint16_t MonthYear;
     uint8_t Day;
     utf8 Text[256];
```
IntelOrca, line 54: Not sure whether you would like to include the extra work in the PR or not, if so then this should become `std::string` and support any length.

src/openrct2/management/NewsItem.h
hdpoliveira: The objective of this change is exactly to allow us to change these away from C-arrays (as I mentioned in https://github.com/OpenRCT2/OpenRCT2/pull/11715#issue-415850000).

That is the next thing I'll do once this PR is approved :)

src/openrct2/management/NewsItem.h
hdpoliveira: Nice point. I'll leave this for another PR, since here I'm focusing on the queue.

src/openrct2/management/NewsItem.cpp
hdpoliveira: Done 😄

hdpoliveira:
@IntelOrca I renamed most of the APIs with 'old' in the name to 'archive', as you mentioned.

PR #11716, opened 2020-05-11
Remove some of goto from vehicle_update_track_motion_mini_golf

hdpoliveira:
I thought I should open a PR for this before it becomes un-reviewable.
This cleans up some of the gotos in vehicle_update_track_motion_mini_golf and creates two methods in Vehicle: UpdateTrackMotionMiniGolf and UpdateTrackMotionMiniGolfVehicle
It also uses CoordsXYZ instead of x, y, and z, and cleans up some use of registers.

Items pending: Remove use of registers and extract functions to remove gotos (or at least reduce) in UpdateTrackMotionMiniGolfVehicle

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -26,6 +26,7 @@
 #include "../scenario/Scenario.h"
 #include "../util/Util.h"
 #include "../windows/Intent.h"
+#include "../world/Map.h"
 #include "../world/MapAnimation.h"
 #include "../world/Park.h"
 #include "../world/Scenery.h"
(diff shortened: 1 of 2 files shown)


tupaschoal:
1 test failed, so probably something isn't exactly as it was.

Also you need to rebase to solve conflicts

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9304,113 +9304,85 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
-    regs.edx = velocity >> 8;
-    regs.edx *= regs.edx;
-    if (regs.ebx < 0)
-    {
-        regs.edx = -regs.edx;
-    }
-    regs.edx >>= 4;
-    regs.eax = regs.edx / totalMass;
-    regs.ecx -= regs.eax;
+    int32_t newAcceleration = (sumAcceleration / numVehicles * 21) >> 9;
```
Gymnasiast, line 9295:
```suggestion
    int32_t newAcceleration = ((sumAcceleration / numVehicles) * 21) >> 9;
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9304,113 +9304,85 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
+    newAcceleration -= velocity >> 12;
+    newAcceleration -= (((velocity >> 8) * abs(velocity >> 8)) >> 4) / totalMass;
```
Gymnasiast, line 9297: (vehicle >> 8) was squared in the old code.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9304,113 +9304,85 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
-    if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_WATER_RIDE))
-    {
-        goto loc_6DD054;
-    }
+            if (update_flags & VEHICLE_UPDATE_FLAG_REVERSING_SHUTTLE)
+            {
+                poweredAcceleration = -poweredAcceleration;
+            }
+            poweredAcceleration -= velocity;
+            poweredAcceleration = poweredAcceleration * (powered_acceleration << 1) / ((speed * totalMass) >> 2);
```
Gymnasiast, line 9309: Having two variables that only differ by casing is not nice.

src/openrct2/ride/Vehicle.cpp
hdpoliveira: I thought so too at first glance, but the actuall math is `var * abs(var)`

This was done like
```
newvar = var * var
if (var < 0)
    newvar = -newvar
```

hdpoliveira: I can create an extra variable to keep it the old way, but I'd need your help coming up with a name for it 😃

src/openrct2/ride/Vehicle.cpp
hdpoliveira: I took this name from `static void vehicle_update_track_motion_powered_ride_acceleration`, and this problem exists there as well (powered_acceleration is a member variable of Vehicle)

In general this problem exists all over the Vehicle.cpp file ☹️ For example, I see this in many places:
```
uint32_t trackType = track_type >> 2
```
I believe we should keep this name as is, for keeping consistency with the other function, and file an issue to take a look at the more generic problem.

hdpoliveira:
@tupaschoal  I believe this is ready for review now 😃

tupaschoal:
I'll wait for @Gymnasiast to complete his first, as he is already involved

hdpoliveira:
@duncanspumpkin  I removed all gotos from the minigolf function, but I need some help coming up with names for the extracted functions. I noticed some Vehicle functions are still named Vehicle::LocSth, so I've named them this way for now.

Gymnasiast:
I must say that I find some commits hard to follow. I lost track of what was happening halfway through https://github.com/OpenRCT2/OpenRCT2/pull/11716/commits/fe2d3bf0b482c99386e8cc8cc3b6961f6d02f03e, which attempted to get rid of registers _and_ gotos at the same time.

In general, I think the PR is too extensive, given the complexity of the matter. I think I'd prefer splitting it up into multiple parts. But don't do that yet, I'd like to hear what the others think.

hdpoliveira:
I don't think there is any problem in breaking this PR into multiple parts. Just tell me a point you believe is ok and I'll break it :)

hdpoliveira:
@Gymnasiast , if you want, I can break this before fe2d3bf, break fe2d3bf into different changes (i.e. redo it again) and submit it in a different PR.

duncanspumpkin:
My main issue with this PR is its diverging further from the non golf version of these update functions. We want to make it converge with the non golf so that we can extract all of the common elements which there are many. 
What I ideally wanted is for there to be the forward and backward functions (i.e from loc_6DCA9A to loc_6DCD6B for backward and loc_6DC462 to loc_6DC9BC for forward).

If we make this PR just fe2d3bf I think we can do that as that is where it starts to get messy.

duncanspumpkin:
Not really related to this pr but https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Vehicle.cpp#L9673
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Vehicle.cpp#L9064
Both of these are incorrect. It should be `acceleration += dword....`. It doesn't seem to make any difference but it will make it easier to streamline the functions.

hdpoliveira:
@duncanspumpkin, I didn't quite understand: do you want all changes up to https://github.com/OpenRCT2/OpenRCT2/commit/fe2d3bf0b482c99386e8cc8cc3b6961f6d02f03e, all changes before https://github.com/OpenRCT2/OpenRCT2/commit/fe2d3bf0b482c99386e8cc8cc3b6961f6d02f03e or only https://github.com/OpenRCT2/OpenRCT2/commit/fe2d3bf0b482c99386e8cc8cc3b6961f6d02f03e?

duncanspumpkin:
Yes keep all changes up to fe2d3bf. That is where it starts to diverge from the non golf version more than I would like.

Gymnasiast:
Up to fe2d3bf, so not including it? That would be my preference as well, fe2d3bf is where I started to lose track. The ones before can probably be merged rather quickly.

src/openrct2/ride/Vehicle.cpp
Gymnasiast: Yes, certainly file that issue.

I would even prefer fixing the naming and only _then_ revisiting this, but I'm not going to force you to do that - I don't want to scare away contributors. ;-)

duncanspumpkin:
Up to and including

Gymnasiast:
I found fe2d3bf hard to review, so I'd prefer if only includes commits before fe2d3bf.

hdpoliveira:
My suggestion: I update this PR with changes up to fe2d3bf, not including it.
Once it is merged, I submit another PR with fe2d3bf.
Then I'll go slowly removing the rest of gotos in a way that doesn't diverge much from the other similar functions.

PR #11718, opened 2020-05-11
Added check to prevent overflow assertion error

MegaMech:
This PR is related to RIDE_STATUS


src/openrct2/actions/RideSetStatus.hpp
```diff
@@ -77,7 +77,14 @@ DEFINE_GAME_ACTION(RideSetStatusAction, GAME_COMMAND_SET_RIDE_STATUS, GameAction
             res->ErrorMessage = STR_NONE;
             return res;
         }
-
+        if (_status >= 4 || _status <= -1)
```
duncanspumpkin, line 80: status is a unsigned so the second check isn't required as it can never be negative.

We try to avoid magic numbers like that 4 but I'm going to guess we haven't defined a ride status count. Could you add RIDE_STATUS_COUNT to https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Ride.h#L632 and use RIDE_STATUS_COUNT instead of 4.

duncanspumpkin:
For any other reviewer this does not need a network increment as it's safe to make things less permissive.

src/openrct2/actions/RideSetStatus.hpp
Broxzier: In this case, `_status` gets used as an index for `_StatusErrorTitles`, so instead of using the defines you could use `std::size(_StatusErrorTitles)`.
Of course, this static array needs to be equal in size, so introducing `RIDE_STATUS_COUNT` and adding a `static_assert` there would be useful.

MegaMech: Wouldn't a static_assert stop the game from continuing? As the point of this change is to tell the JS operator that they chose an invalid status rather than crashing the game.

MegaMech: Would it also be relevant to add to the error message something like:
"Allowable values are range from zero to three"

src/openrct2/actions/RideSetStatus.hpp
```diff
@@ -77,7 +77,14 @@ DEFINE_GAME_ACTION(RideSetStatusAction, GAME_COMMAND_SET_RIDE_STATUS, GameAction
             res->ErrorMessage = STR_NONE;
             return res;
         }
-
+        if (_status >= RIDE_STATUS_COUNT)
```
Gymnasiast, line 80: Could you put a blank line above and below this block?

src/openrct2/ride/Ride.h
```diff
@@ -630,6 +630,7 @@ enum
     RIDE_STATUS_OPEN,
     RIDE_STATUS_TESTING,
     RIDE_STATUS_SIMULATING,
+    RIDE_STATUS_COUNT
```
Gymnasiast, line 633: Please put a trailing comma (for consistency).

src/openrct2/actions/RideSetStatus.hpp
```diff
@@ -78,6 +78,15 @@ DEFINE_GAME_ACTION(RideSetStatusAction, GAME_COMMAND_SET_RIDE_STATUS, GameAction
             return res;
         }
 
+        if (_status >= RIDE_STATUS_COUNT)
+        {
+            log_warning("Invalid ride status", uint32_t(_rideIndex));
```
Gymnasiast, line 83: This is not correct. The string does not contain any parameters, yet rideIndex is added.
```suggestion
            log_warning("Invalid ride status %u for ride %u", uint32_t(_status), uint32_t(_rideIndex));
```

Gymnasiast:
@duncanspumpkin Could you re-review?

PR #11720, opened 2020-05-11
Replace many C-style casts

Gymnasiast:


src/openrct2/world/Location.hpp
```diff
@@ -13,7 +13,7 @@
 
 #include <algorithm>
 
-#define LOCATION_NULL ((int16_t)(uint16_t)0x8000)
+constexpr const int16_t LOCATION_NULL = -32768;
```
Gymnasiast, line 16: Would appreciate a confirmation that this is indeed the same value.

Broxzier: Confirmed: http://cpp.sh/6tvae

tupaschoal:
Part of #11158 

(Forgot to say earlier :|)

PR #11721, opened 2020-05-11
Remove commented out code

Gymnasiast:


src/openrct2/drawing/LightFX.cpp
```diff
@@ -279,7 +279,6 @@ void lightfx_prepare_light_list()
         {
             int32_t totalSamplePoints = 5;
             int32_t startSamplePoint = 1;
-            // int32_t lastSampleCount = 0;
```
duncanspumpkin, line 282: Can you keep the lightfx.cpp dead code.

Gymnasiast: I'd argue we have source control for that.

PR #11722, opened 2020-05-11
Fix #11703: Remove nonfunctional "guest is lost" notification

### LINKED ISSUE #11703: Notification config for 'guest is lost' seems not working
**OS:** Windows 10
**Version:** 0.2.6
**Commit/Build:** c447cde

"Guest is lost" config at news message notification config seems not working.
Or, if it is affected to only a guest who is tracked, I think there should be another option to turn on/off the global guests' lost.

**Steps to reproduce:**
1. Uncheck "Guest is lost" at Nofitication config window
2. Open any map people lost.

**Screenshots / Video:**

image

Gymnasiast:


telk5093:
I thought I couldn't see any related code but it was actually none! :<

PR #11735, opened 2020-05-12
[Plugin] Allow plugins to update dropdown items

IntelOrca:


src/openrct2-ui/scripting/ScWidget.hpp
```diff
@@ -355,6 +355,61 @@ namespace OpenRCT2::Scripting
...
+
+    private:
+        int32_t selectedIndex_get() const
+        {
+            auto w = GetWindow();
+            if (w != nullptr)
+            {
+                return GetWidgetSelectedIndex(w, _widgetIndex);
+            }
+            return false;
```
tupaschoal, line 382: Looks weird to return a boolean on something that should be `int32_t`

src/openrct2-ui/scripting/CustomWindow.cpp
```diff
@@ -1070,6 +1058,95 @@ namespace OpenRCT2::Ui::Windows
...
+                {
+                    if (selectedIndex < 0 || static_cast<size_t>(selectedIndex) >= customWidgetInfo->Items.size())
+                    {
+                        selectedIndex = 0;
+                    }
+                }
+
+                if (selectedIndex == -1)
+                {
+                    widget.string = const_cast<utf8*>("");
```
tupaschoal, line 1100: Do we need this cast here?

IntelOrca: yes

PR #11747, opened 2020-05-14
[Plugin] Add language and split peep into guest and staff

IntelOrca:


distribution/openrct2.d.ts
```diff
@@ -744,6 +749,38 @@ declare global {
...
          */
         energyTarget: number;
+    }
+
+    type PeepType = "guest" | "staff";
+
+    /**
+     * Represents a guest.
+     */
+    interface Guest extends Entity {
```
duncanspumpkin, line 759: Shouldn't this extend peep?

IntelOrca: Yes

distribution/openrct2.d.ts
```diff
@@ -720,20 +720,25 @@ declare global {
...
          * Name of the peep.
          */
         name: string;
 
         /**
-         * Colour of the peep's t-shirt.
+         * Main flags for peep behaviour.
          */
-        tshirtColour: number;
+        flags: number;
```
AaronVanGeffen, line 736: While there are advantages to opening the flags up like this, there are a few drawbacks as well. Knowledge of which bit is meant for what flag is one. However, a more pressing one is that any plug-ins using these flags will break if we ever decide to change or add flags, right? The main code base will be refactored accordingly, but we won't have that kind of control over plug-ins.

IntelOrca: What do you suggest doing instead? It takes so long going through every field, and any large enumeration like this then requires expanding to many more fields all which require code and documentation or writing a routine to convert to and from strings.

IntelOrca: What about this way? (see latest commit)

AaronVanGeffen: Your latest commit has introduced a much better approach, I think. These two functions should be much more extensible and easier to support in the future, I think. Thanks. :)

PR #11748, opened 2020-05-14
Rename MoneyEffect member variables to prevent shadowing

tupaschoal:
Continues #10653

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -1705,11 +1705,11 @@ private:
 
     void ImportMoneyEffect(MoneyEffect* dst, MoneyEffect* src)
     {
-        dst->move_delay = src->move_delay;
-        dst->num_movements = src->num_movements;
-        dst->value = src->value;
-        dst->offset_x = src->offset_x;
-        dst->wiggle = src->wiggle;
+        dst->MoveDelay = src->MoveDelay;
+        dst->NumMovements = src->NumMovements;
+        dst->Value = src->Value;
+        dst->OffsetX = src->OffsetX;
+        dst->Wiggle = src->Wiggle;
     }
 
     void ImportSteamParticle(SteamParticle* dst, SteamParticle* src)
--- a/src/openrct2/world/MoneyEffect.cpp
+++ b/src/openrct2/world/MoneyEffect.cpp
@@ -47,16 +47,16 @@ void MoneyEffect::CreateAt(money32 value, int32_t x, int32_t y, int32_t z, bool 
     if (moneyEffect == nullptr)
         return;
 
-    moneyEffect->value = value;
-    moneyEffect->vertical = (vertical ? 1 : 0);
+    moneyEffect->Value = value;
+    moneyEffect->Vertical = (vertical ? 1 : 0);
     moneyEffect->sprite_width = 64;
     moneyEffect->sprite_height_negative = 20;
     moneyEffect->sprite_height_positive = 30;
     moneyEffect->sprite_identifier = SPRITE_IDENTIFIER_MISC;
     moneyEffect->MoveTo({ x, y, z });
     moneyEffect->type = SPRITE_MISC_MONEY_EFFECT;
-    moneyEffect->num_movements = 0;
-    moneyEffect->move_delay = 0;
+    moneyEffect->NumMovements = 0;
+    moneyEffect->MoveDelay = 0;
 
     int16_t offsetX = 0;
     if (!gOpenRCT2NoGraphics)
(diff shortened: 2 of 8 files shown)


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1705,11 +1705,11 @@ class S4Importer final : public IParkImporter
 
     void ImportMoneyEffect(MoneyEffect* dst, MoneyEffect* src)
     {
-        dst->move_delay = src->move_delay;
-        dst->num_movements = src->num_movements;
-        dst->value = src->value;
-        dst->offset_x = src->offset_x;
-        dst->wiggle = src->wiggle;
+        dst->MoveDelay = src->MoveDelay;
```
duncanspumpkin, line 1708: This isn't right. S4 should have a RCT12SpriteMoneyEffect. It's not a mistake with this pr though.

src/openrct2/world/MoneyEffect.cpp
```diff
@@ -47,16 +47,16 @@ void MoneyEffect::CreateAt(money32 value, int32_t x, int32_t y, int32_t z, bool
     if (moneyEffect == nullptr)
         return;
 
-    moneyEffect->value = value;
-    moneyEffect->vertical = (vertical ? 1 : 0);
+    moneyEffect->Value = value;
+    moneyEffect->Vertical = (vertical ? 1 : 0);
```
duncanspumpkin, line 51: I believe that vertical is actually whether it is a credit or debit. We should make this an enum class and define the two states.

src/openrct2/rct1/S4Importer.cpp
IntelOrca: I don't think it is a mistake, just we haven't got round to changing it yet. They all used to use the same structs, we have gradually been copying them and removing the push, 1 pop.

duncanspumpkin: Yes but for now we don't enforce packing on money effect so who knows if this works atm.

src/openrct2/world/MoneyEffect.cpp
tupaschoal: I think it has two meanings, I did convert it to an enum, but the `GetStringId` makes things look a bit weird. I've added the commit for you guys to see

src/openrct2/world/Sprite.h
```diff
@@ -74,14 +74,20 @@ struct Duck : SpriteGeneric
...
-    static void CreateAt(money32 value, int32_t x, int32_t y, int32_t z, bool vertical);
+    enum class TransactionType : bool
+    {
+        CREDIT,
+        DEBIT,
+    };
+
+    uint16_t MoveDelay;
+    uint8_t NumMovements;
+    TransactionType CashFlow;
```
Gymnasiast, line 85: I am not entirely convinced this is indeed credit/debit. Would have to check the calls. I am not sure about the name of this field either.

src/openrct2/world/MoneyEffect.cpp
```diff
@@ -147,11 +147,12 @@ void MoneyEffect::Update()
 
 std::pair<rct_string_id, money32> MoneyEffect::GetStringId() const
 {
-    rct_string_id spentStringId = vertical ? STR_MONEY_EFFECT_SPEND_HIGHP : STR_MONEY_EFFECT_SPEND;
-    rct_string_id receiveStringId = vertical ? STR_MONEY_EFFECT_RECEIVE_HIGHP : STR_MONEY_EFFECT_RECEIVE;
+    rct_string_id spentStringId = CashFlow == TransactionType::CREDIT ? STR_MONEY_EFFECT_SPEND_HIGHP : STR_MONEY_EFFECT_SPEND;
```
Gymnasiast, line 150:
```suggestion
    rct_string_id spentStringId = CashFlow == TransactionType::DEBIT ? STR_MONEY_EFFECT_SPEND_HIGHP : STR_MONEY_EFFECT_SPEND;
```

src/openrct2/world/MoneyEffect.cpp
```diff
@@ -147,11 +147,12 @@ void MoneyEffect::Update()
...
+    rct_string_id receiveStringId = CashFlow == TransactionType::CREDIT ? STR_MONEY_EFFECT_RECEIVE_HIGHP
+                                                                        : STR_MONEY_EFFECT_RECEIVE;
```
Gymnasiast, line 152: These lines show that `vertical` does not mean whether it credit/debit, IMO.

Gymnasiast: I dug a bit further. The `HIGHP` strings contain currency strings with 2 decimals and are used for guest purchases. The only odd bit is why it increases Z for guest purchases.

duncanspumpkin: Guest purchases were added later so they work differently. I'd forgotten about that. Renaming the variable might not make sense now.

tupaschoal:
I've dropped the renaming of `Vertical`, as we agreed that it wasn't really just credit/debit. I think this is good to go then.

src/openrct2/rct1/S4Importer.cpp
tupaschoal: I created #11757 to track this issue, feel free to add more details there.

PR #11749, opened 2020-05-14
Rename ServerList member variables to prevent shadowing

tupaschoal:
Continues #10653

### DIFF
--- a/src/openrct2/network/ServerList.cpp
+++ b/src/openrct2/network/ServerList.cpp
@@ -35,45 +35,45 @@ int32_t ServerListEntry::CompareTo(const ServerListEntry& other) const
     const auto& a = *this;
     const auto& b = other;
 
-    // Order by favourite
-    if (a.favourite != b.favourite)
+    // Order by Favourite
+    if (a.Favourite != b.Favourite)
     {
-        return a.favourite ? -1 : 1;
+        return a.Favourite ? -1 : 1;
     }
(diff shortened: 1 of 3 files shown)


src/openrct2/network/ServerList.cpp
```diff
@@ -35,45 +35,45 @@ int32_t ServerListEntry::CompareTo(const ServerListEntry& other) const
     const auto& a = *this;
     const auto& b = other;
 
-    // Order by favourite
-    if (a.favourite != b.favourite)
+    // Order by Favourite
```
Gymnasiast, line 38: All of these comments only state the obvious. We can get rid of them.

src/openrct2/network/ServerList.cpp
```diff
@@ -193,19 +193,19 @@ void ServerList::ReadAndAddFavourites()
...
+            _serverEntries.begin(), _serverEntries.end(), [](const ServerListEntry& entry) { return entry.Favourite; }),
         _serverEntries.end());
     auto entries = ReadFavourites();
     AddRange(entries);
 }
 
 void ServerList::WriteFavourites() const
 {
-    // Save just favourite servers
+    // Save just Favourite servers
```
Gymnasiast, line 204: revert

tupaschoal:
Fixed comments, do squash on merge.

PR #11755, opened 2020-05-15
[Plugin] Add API for editing park messages

IntelOrca:


src/openrct2/management/NewsItem.cpp
```diff
@@ -484,3 +484,30 @@ void news_item_add_to_queue_custom(NewsItem* newNewsItem)
     newsItem++;
     newsItem->Type = NEWS_ITEM_NULL;
 }
+
+void news_item_remove(int32_t index)
+{
+    if (index < 0 || index >= MAX_NEWS_ITEMS)
+        return;
```
tupaschoal, line 491: Should we log something in case this means someone doing crazy stuff with plugins?

src/openrct2/management/NewsItem.cpp
```diff
@@ -484,3 +484,30 @@ void news_item_add_to_queue_custom(NewsItem* newNewsItem)
...
+        gNewsItems[NEWS_ITEM_HISTORY_START - 1].Type = NEWS_ITEM_NULL;
+    }
+    else
+    {
+        for (size_t i = index; i < MAX_NEWS_ITEMS - 1; i++)
+        {
+            gNewsItems[i] = gNewsItems[i + 1];
+        }
+        gNewsItems[MAX_NEWS_ITEMS - 1].Type = NEWS_ITEM_NULL;
+    }
```
tupaschoal, line 512:
```suggestion
    auto newsBoundary = index < NEWS_ITEM_HISTORY_START ? NEWS_ITEM_HISTORY_START : MAX_NEWS_ITEMS;
    for (size_t i = index; i < newsBoundary - 1; i++)
    {
        gNewsItems[i] = gNewsItems[i + 1];
    }
    gNewsItems[newsBoundary - 1].Type = NEWS_ITEM_NULL;
```

src/openrct2/scripting/ScPark.hpp
```diff
@@ -24,6 +25,209 @@
 
 namespace OpenRCT2::Scripting
 {
+    static constexpr const char* ParkMessageTypeStrings[] = { "attraction", "peep_on_attraction", "peep",   "money",
+                                                              "blank",      "research",           "guests", "award",
+                                                              "chart" };
```
tupaschoal, line 30: This is so weird, I'd rather break after the `{` and have one per line than this. :X

src/openrct2/scripting/ScPark.hpp
```diff
@@ -24,6 +25,209 @@
...
+        uint8_t i = 0;
+        for (const auto& k : ParkMessageTypeStrings)
+        {
+            if (k == key)
+            {
+                return NEWS_ITEM_RIDE + i;
+            }
+            i++;
+        }
+        return NEWS_ITEM_BLANK;
```
tupaschoal, line 43: What about this (or similar?)
```suggestion
        auto it = std::find(ParkMessageTypeStrings.begin(), ParkMessageTypeStrings.end(), key);
        return it != ParkMessageTypeStrings.end() ? std::distance(ParkMessageTypeStrings.begin(), it) + NEWS_ITEM_RIDE : NEWS_ITEM_BLANK;
```

src/openrct2/scripting/ScPark.hpp
```diff
@@ -24,6 +25,209 @@
...
+    }
+
+    inline std::string GetParkMessageType(uint8_t type)
+    {
+        type--;
```
tupaschoal, line 48: Why do we have to decrement type?
Sounds like the type of witchery that could use a comment

src/openrct2/scripting/ScPark.hpp
```diff
@@ -75,6 +279,61 @@ namespace OpenRCT2::Scripting
             context_broadcast_intent(&intent);
         }
 
+        std::vector<std::shared_ptr<ScParkMessage>> messages_get() const
+        {
+            std::vector<std::shared_ptr<ScParkMessage>> result;
+            for (int32_t i = 0; i < 11; i++)
```
tupaschoal, line 285: Is 11 the enum you created, `NEWS_ITEM_HISTORY_START `?

src/openrct2/scripting/ScPark.hpp
```diff
@@ -75,6 +279,61 @@ namespace OpenRCT2::Scripting
...
+            {
+                if (news_item_is_empty(i))
+                    break;
+                result.push_back(std::make_shared<ScParkMessage>(i));
+            }
+            for (int32_t i = 11; i < MAX_NEWS_ITEMS; i++)
```
tupaschoal, line 291: Same

src/openrct2/management/NewsItem.cpp
IntelOrca: No you would use `duk_error` to throw a JavaScript exception.

src/openrct2/scripting/ScPark.hpp
IntelOrca: This is clang-format, not me.

Broxzier: Add a comma to the end, and clang-format will put them all one one line.

Issue #10634, opened 2020-01-29
peeps unable to use uphill paths out of toilets


DollyMixedUpWoman:
**OS:** Windows 10

**Version:** v0.1.2

**Commit/Build:** 6547116

Guests get stuck when exiting toilets onto a sloped path. They walk through the path rather than up the hill. When the path tile is deleted they fall through the base land and out of the game. 

- [unknown ] Reproducible in RCT2 (vanilla)?
- [no ] Specific to multiplayer?

**Steps to reproduce:**
1. build toilets at the bottom of a hill
2. build path tile
3. watch peeps get stuck
4. delete path tile to release them
5. watch them fall to their doom

**Screenshots / Video:**
![2020-01-29](https://user-images.githubusercontent.com/33560724/73380654-583e8a80-42bc-11ea-974e-fa536c5743fa.png)

**Save game:**
[Cracking Glacier Park broken bogs.zip](https://github.com/OpenRCT2/OpenRCT2/files/4129385/Cracking.Glacier.Park.broken.bogs.zip)

edit to add: i had the same issue in a different scenario but can't remember what it was

ocalhoun6:
Don't know if it's really a bug, but it might be related to this:

I've noticed that every time a peep exits a bathroom (and also first aid stalls), they first walk straight out for half a tile, then walk back to the bathroom door, and _then_ walk away from the bathroom normally.

My uneducated guess is that they're doing this here, too, but since the path is sloped upward and they walk straight outward from the bathroom, they walk straight through the path and end up getting stuck underneath it.

I just tried it with an upward-sloping path into the bathroom, and they do act weirdly when exiting, being drawn beneath the path for a while as they do their little in-and-out dance ... but on the upward-sloping paths, they don't get stuck.

PR #11758, opened 2020-05-16
Fix peeps leaving shop at upward slope.

tylerleamon:
Fix: https://github.com/OpenRCT2/OpenRCT2/issues/10634#event-3273204131

Fix follows logic for peeps exiting rides in UpdateRideLeaveExit().

Originally Peeps, once exiting the shop, would double-back towards the shop before resuming walking. In RCT Classic, Peeps still double-back, but find the sloped path. Not able to verify if issue exists in Vanilla where Peeps get stuck instead of finding path. Either way, this fix prevents Peeps from doubling-back towards the shop.

Included test file. Tests Toilets, First Aid, and Information Kiosk using both slope directions and flat. Use Cheats with Large Tram to fill park. Then execute Max Toilets, Nausea, or Rain. 
[Test_Shop.sv6.zip](https://github.com/OpenRCT2/OpenRCT2/files/4638456/Test_Shop.sv6.zip)

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -5113,7 +5113,7 @@ void Guest::UpdateRideShopLeave()
             return;
     }
 
-    SetState(PEEP_STATE_WALKING);
+    SetState(PEEP_STATE_FALLING);
 
     auto ride = get_ride(current_ride);
     if (ride != nullptr)
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -5113,7 +5113,7 @@ void Guest::UpdateRideShopLeave()
             return;
     }
 
-    // Previously SetState(PEEP_STATE_WALKING) caused Peeps to double-back to exit point of shop
+    //#11758 Previously SetState(PEEP_STATE_WALKING) caused Peeps to double-back to exit point of shop
     SetState(PEEP_STATE_FALLING);
 
     auto ride = get_ride(current_ride);
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -5113,6 +5113,7 @@ void Guest::UpdateRideShopLeave()
             return;
     }
 
+    // Previously SetState(PEEP_STATE_WALKING) caused Peeps to double-back to exit point of shop
     SetState(PEEP_STATE_FALLING);
 
     auto ride = get_ride(current_ride);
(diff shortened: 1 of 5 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -5113,7 +5113,7 @@ void Guest::UpdateRideShopLeave()
             return;
     }
 
-    SetState(PEEP_STATE_WALKING);
+    SetState(PEEP_STATE_FALLING);
```
Gymnasiast, line 5116: Put a comment above this saying it used to be `PEEP_STATE_WALKING`, and why it was changed. Link to the issue as well.

Especially for the black magic that is pathfinding, I prefer if changes are properly documented. :-)

tylerleamon: Updated. At this line in the pathfinding, Peeps are at the exit point of the shop/ride. It isn't a perfect solution to set them to falling, but it does prevent them from walking until they've truly exited the shop/ride.

duncanspumpkin:
Yes that seems pretty reasonable. Have you checked a variety of situations and types of shop to ensure that they do not fall through the floor?

Will need a replay update thats why CI's are failing

Gymnasiast:
We should also test this above water, preferably with normal path, an upwards slope and a downwards slope. Just to make sure that the peeps will not drown.

tylerleamon:
> Yes that seems pretty reasonable. Have you checked a variety of situations and types of shop to ensure that they do not fall through the floor?
> 
> Will need a replay update thats why CI's are failing

I'll do some testing and upload a new save file that includes testing above water and slopes that do/don't have land underneath.

tylerleamon:
Testing went well. I loaded 750 guests in the attached park with a variety of slopes and terrain with water, land, underground, above ground. No guests drowned or were lost to the abyss. I also stepped through the UpdateFalling() function when a peep exits a toilet on an upward slope directly over water. The potential for drowning exists until the Line 938 Condition:
`                    if (height - 4 >= z && height < z + 20)
`
Because the `height` of the water is equal to the Peep's `z`, the drowning condition fails and the loop attempts to find the next tile element at the Peep's location, if it exists, which is the sloped path. The loop finds the sloped path and the peep enters the walking state.

Also a quick note. UpdateFalling() has the comment `Note: peeps 'fall' after leaving a ride/enter the park.`. Looks like this logic has been in the code for a while for peeps entering the park.

[Test_Shop_Terrain.sv6.zip](https://github.com/OpenRCT2/OpenRCT2/files/4641149/Test_Shop_Terrain.sv6.zip)

duncanspumpkin:
Yes I wrote that comment. I think most likely CS just thought there was no need to do a fall for shops as it could mean that peeps return to the shop again. I'll fix the replay when I get a moment. This will requrie a network increment as well.

Gymnasiast:
I'd like to test it as well.

Gymnasiast:
I have tested it. It seems to work correctly. :+1:

It will need a changelog entry.
`- Fix: [#10634] Guests are unable to use uphill paths out of toilets.`

PR #11770, opened 2020-05-17
Fix #11568 - create draw_string_left_underline() overload

### LINKED ISSUE #11568: Use ScreenCoordsXY on draw_string_left_underline
Create an overload of `draw_string_left_underline` on `\src\openrct2\drawing\Text.cpp` with the same name, but that takes a `const ScreenCoordsXY&` object instead of an `int32_t x` and and an `int32_t y` and replace the calls to the `{x ,y}` pair by the one using `ScreenCoordsXY`.

TomasZilinek:
created the overload, changed all calls from the old function to the new (just 1 call) and deleted the old function


tupaschoal:
Well, this one was easy :D

TomasZilinek:
> Well, this one was easy :D

yep, I was shocked :D

PR #11771, opened 2020-05-17
Fix #11567 - create draw_string_centred_underline() overload

### LINKED ISSUE #11567: Use ScreenCoordsXY on draw_string_centred_underline
Create an overload of `draw_string_centred_underline` on `\src\openrct2\drawing\Text.cpp` with the same name, but that takes a `const ScreenCoordsXY&` object instead of an `int32_t x` and and an `int32_t y` and replace the calls to the `{x ,y}` pair by the one using `ScreenCoordsXY`.

TomasZilinek:
I created the overload and deleted the old function (there were no calls)


TomasZilinek:
Were there really no calls to the old function? I am confused

tupaschoal:
Seems like there weren't, maybe it was just written for "feature completion", but I can't really tell.

PR #11775, opened 2020-05-18
Extend the game action result properties

frutiemax:
So I've encountered some limitations while developing the territory rights plugin so I am attempting to fill the missing hooks:

- [x] Expose the map range of a raise land action result


duncanspumpkin:
This is not really the approach we were thinking of doing. For one this isn't guaranteed to be correct as it's the unvalidated range. This should be done through the visit interface. I'm thinking we will add a game action validate tile that can be hooked by the plugins which is more what you would want

PR #11776, opened 2020-05-18
gfx_draw_string_left_wrapped overload added

sohiladwani:
Planning to replace more calls just want to make sure I did the first correct.

### DIFF
--- a/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
+++ b/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
@@ -987,11 +987,7 @@ static void window_editor_objective_options_main_paint(rct_window* w, rct_drawpi
     ft = Formatter::Common();
     ft.Add<rct_string_id>(STR_STRING);
     ft.Add<const char*>(gS6Info.details);
-    ScreenCoordsXY screenCoords;
-    screenCoords.x = x;
-    screenCoords.y = y;
-    gfx_draw_string_left_wrapped(dpi, gCommonFormatArgs, screenCoords, width, STR_BLACK_STRING, COLOUR_BLACK);
-
+    gfx_draw_string_left_wrapped(dpi, gCommonFormatArgs, { x, y }, width, STR_BLACK_STRING, COLOUR_BLACK);
 
     // Scenario category label
     x = w->windowPos.x + 8;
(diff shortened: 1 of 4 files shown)


src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -987,7 +987,11 @@ static void window_editor_objective_options_main_paint(rct_window* w, rct_drawpi
     ft = Formatter::Common();
     ft.Add<rct_string_id>(STR_STRING);
     ft.Add<const char*>(gS6Info.details);
-    gfx_draw_string_left_wrapped(dpi, gCommonFormatArgs, x, y, width, STR_BLACK_STRING, COLOUR_BLACK);
+    ScreenCoordsXY screenCoords;
+    screenCoords.x = x;
+    screenCoords.y = y;
+    gfx_draw_string_left_wrapped(dpi, gCommonFormatArgs, screenCoords, width, STR_BLACK_STRING, COLOUR_BLACK);
+
```
tupaschoal, line 994:
```suggestion
    gfx_draw_string_left_wrapped(dpi, gCommonFormatArgs, { x, y }, width, STR_BLACK_STRING, COLOUR_BLACK);
```

You can create the objects in place following my suggestion. This function will probably have a `screenCoords` at some point, but then it will need to be used for all calls and you probably don't want to tackle it now.

src/openrct2/drawing/Text.cpp
```diff
@@ -229,6 +229,26 @@ int32_t gfx_draw_string_left_wrapped(
     return layout.GetHeight();
```
tupaschoal, line 229: Make the old function call the new one to avoid code duplication

src/openrct2/drawing/Text.h
```diff
@@ -1,3 +1,4 @@
+/*****************************************************************************
```
tupaschoal, line 1: Please revert this change

src/openrct2/drawing/Text.cpp
tupaschoal: Please make the old function call the new one, like I said, to avoid code duplication

sohiladwani: Sorry missed this, will do in next commit!

PR #11781, opened 2020-05-19
Merge cmdline and JSON object palette

Gymnasiast:


Gymnasiast:
Not sure why win32 fails to compile. All the required headers appear to be included.

duncanspumpkin:
Think its just a mistake in that one run of the build. I've told it to rerun.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -236,7 +215,7 @@ static bool sprite_file_export(int32_t spriteIndex, const char* outPath)
     dpi.pitch = 0;
     dpi.zoom_level = 0;
 
-    std::memcpy(spriteFilePalette, CmdlineSprite::_standardPalette, 256 * 4);
+    std::memcpy(spriteFilePalette, StandardPalette, 256 * sizeof(PaletteBGRA));
```
IntelOrca, line 218: `spriteFilePalette = StandardPalette`

IntelOrca:
Just the one suggestion. I think it would be nice to make a struct called GamePalette or something and expose that rather than an array. But up to you.

Gymnasiast:
I have now made the requested changes.

I have tested it with the following:
- Building g2.dat
- Extracting an image from g2.dat
- Reading PNG files from a JSON object (including remap colours)

PR #11798, opened 2020-05-23
Fix #11796: Unable to use sharedStorage

### LINKED ISSUE #11796: [plugin] Unable to use sharedStorage
Whenever I try to use plugin configuration, I am unable to set a value. Whether I use `context.sharedStorage` or `context.configuration`, I always hit [this "Property does not exist." error](https://github.com/OpenRCT2/OpenRCT2/blob/7e7b15812dabe6977f3573b2a3f52b1737c5527c/src/openrct2/scripting/ScConfiguration.hpp#L224).

It looks like `_isUserConfig` is always `true`.

Am I doing something wrong?

IntelOrca:
isUserConfig must only be true for configuration not sharedStorage.

### DIFF
--- a/src/openrct2/scripting/ScConfiguration.hpp
+++ b/src/openrct2/scripting/ScConfiguration.hpp
@@ -22,12 +22,17 @@ namespace OpenRCT2::Scripting
     class ScConfiguration
     {
     private:
-        bool _isUserConfig{ true };
+        bool _isUserConfig{};
         DukValue _backingObject;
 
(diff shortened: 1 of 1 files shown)


Gymnasiast:
> isUserConfig must only be true for configuration not sharedStorage.

Could you elaborate?

IntelOrca:
The same class is used for configuration and sharedStorage, this flag says which one we are talking about.

PR #11818, opened 2020-05-26
Create new PaletteMap type to replace byte pointers

IntelOrca:
Create new PaletteMap type to replace byte pointers

### DIFF
--- a/src/openrct2/Game.cpp
+++ b/src/openrct2/Game.cpp
@@ -24,6 +24,7 @@
 #include "config/Config.h"
 #include "core/FileScanner.h"
 #include "core/Path.hpp"
+#include "interface/Colour.h"
 #include "interface/Screenshot.h"
 #include "interface/Viewport.h"
 #include "interface/Window.h"
@@ -166,7 +167,7 @@ void update_palette_effects()
                 paletteOffset[(i * 4) + 1] = -((0xFF - g1->offset[(i * 3) + 1]) / 2) - 1;
                 paletteOffset[(i * 4) + 2] = -((0xFF - g1->offset[(i * 3) + 2]) / 2) - 1;
             }
-            platform_update_palette(gGamePalette, 10, 236);
+            platform_update_palette(gGamePalette, PALETTE_OFFSET_DYNAMIC, PALETTE_LENGTH_DYNAMIC);
         }
         gClimateLightningFlash++;
     }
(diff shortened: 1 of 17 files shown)


src/openrct2/libopenrct2.vcxproj
```diff
@@ -50,19 +50,720 @@
     <None Include="data\shaders\**\*.vert" />
     <None Include="data\shaders\**\*.frag" />
   </ItemGroup>
-  <ItemGroup>
-    <ClCompile Include="**\*.cpp" />
-  </ItemGroup>
   <ItemGroup>
     <ClInclude Include="..\..\resources\resource.h" />
-    <ClInclude Include="**\*.h" />
-    <ClInclude Include="**\*.hpp" />
```
Gymnasiast, line 59: This change is not related to the rest.
(Also, do we really need to do this?)

IntelOrca: I will exclude this change. But since the VS update it will be something that will keep ending up in PRs. Maybe I can disable tracking for the file...

duncanspumpkin: We will have to do this unfortunately. If we turn off tracking it's just going to be annoying when they eventually disable the feature completely.

duncanspumpkin:
There is far too much happening in this one pr. Couldn't you have split it into at least three (vs file change, game palettes, palette maps)

IntelOrca:
@duncanspumpkin sorry, I originally wanted to implement magnify zoom but decided to refactor palette maps and it spiralled out of control.

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -358,59 +358,43 @@ bool gfx_load_csg()
...
             }
         }
-        auto primary_palette = gfx_get_g1_element(primary_offset);
-        if (primary_palette != nullptr)
+
+        auto primaryPaletteMap = GetPaletteMapForColour(imageId.GetPrimary());
+        if (primaryPaletteMap)
         {
-            std::memcpy(palette_pointer + 0xF3, &primary_palette->offset[0xF3], 12);
+            paletteMap.Copy(0xF3, *primaryPaletteMap, 0xF3, 12);
```
Gymnasiast, line 388: The starts of the remap colours should be put into constants. I don't know if there is a constant for the number of shades per colour - it should either be used, or one should be created.

src/openrct2/Game.cpp
```diff
@@ -269,7 +270,7 @@ void update_palette_effects()
         if (g1 != nullptr)
         {
             uint8_t* vs = &g1->offset[j * 3];
-            uint8_t* vd = &gGamePalette[243 * 4];
+            uint8_t* vd = &gGamePalette[PALETTE_INDEX_243 * 4];
```
Gymnasiast, line 273: 243 == 0xF3. Isn't this just PALETTE_OFFSET_REMAP_PRIMARY?

IntelOrca: Yes, but as I was unsure how it was used I decided to leave it as 243.

Gymnasiast:
I saw this while testing. Currently checking if it is present on develop.

afbeelding

PR #11820, opened 2020-05-27
Remove (almost) non-functional Twitch integration

Gymnasiast:
Ever since the "middle server" was shut down, Twitch integration required people setting up their own. At this point, the game would be better served by a new implementation, perhaps even via a plug-in.


IntelOrca:
There are changes to sprite IDs etc. are they related?

AaronVanGeffen:
@IntelOrca I believe they're just a result of removing the Twitch icon from the G2 sprites.

Gymnasiast:
That is indeed the reason.

PR #11824, opened 2020-05-30
Fix #11298: Unable to advertise server (500): Retry with ipv4 only

### LINKED ISSUE #11298: Unable to advertise server (500): Retry with ipv4 only.
**OS:** Windows 7 Ultimate SP1
**Version:** OpenRCT2 0.2.5
**Commit/Build:** 4f6e77e

Whenever I try to start my OpenRCT2 server from the command line, I get this error. I think to solve it I have to force the server to host on ipv4? The server, however, still shows up locally. The issue also showed up in OpenRCT2 0.2.4.

**Steps to reproduce:**
1. Execute command _openrct2 host communitymultiplayer.sv6 --port 11753 --headless --password notactuallythepassword_

**Screenshots / Video:**

**Save game:**
[communitymultiplayer.zip](https://github.com/OpenRCT2/OpenRCT2/files/4480414/communitymultiplayer.zip)

IntelOrca:
Add new config option to allow any address to be advertised. This then doesn't rely on the master server retrieving the server IP address via the HTTP request which can often be IPv6 by default.

Also fixes WinHTTP not working with non TLS URLs.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -5,6 +5,7 @@
 - Feature: [#10572] Cheat to allow building at invalid heights.
 - Feature: [#11155] Guest entry points can now be removed by clicking them again.
 - Feature: [#11231] Change shortcut window list order to be more intuitive, and split it into logical sections.
+- Feature: [#11298] Custom IP address can now be advertised to the master server to work around IPv6 issues.
 - Feature: [#11306] Path additions are now kept when replacing the path.
 - Feature: [#11320] Support for custom JavaScript plugins.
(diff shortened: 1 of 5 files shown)


IntelOrca:
Maybe, depends how many people are likely needing to set it.

Gymnasiast:
Could you upload the changelog as well? Perhaps add it as a feature - if it is listed as a bugfix, people will probably assume it doesn't require any action from the user.

IntelOrca:
Updated changelog.

PR #11826, opened 2020-05-30
Fix compilation on Docker CI

Gymnasiast:
I have no idea why it complains about it being unused, seeing as that _is not even true_.


Gymnasiast:
I have no idea why it *still* does not want to work:
```
In file included from ../src/openrct2/drawing/Drawing.cpp:10:
../src/openrct2/drawing/../object/../drawing/Drawing.h:461:14: error: private field '_numMaps' is not used [-Werror,-Wunused-private-field]
    uint16_t _numMaps;
             ^
```
Especially since that instruction isn't even at line 461 any more.

PR #11834, opened 2020-05-30
Renames some Peep member variables to prevent shadowing

tupaschoal:
Continues #10653

It might be easier to review on a commit by commit basis

### DIFF
--- a/src/openrct2/peep/Peep.h
+++ b/src/openrct2/peep/Peep.h
@@ -714,7 +714,7 @@ struct Peep : SpriteBase
     {
         money16 paid_to_enter;
         uint16_t staff_lawns_mown;
-        uint16_t staff_rides_fixed;
+        uint16_t StaffRidesFixed;
     };
     union
     {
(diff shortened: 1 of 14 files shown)


src/openrct2/peep/Peep.h
```diff
@@ -709,35 +709,35 @@ struct Peep : SpriteBase
...
     };
-    uint8_t no_of_food;
-    uint8_t no_of_drinks;
-    uint8_t no_of_souvenirs;
-    uint8_t vandalism_seen; // 0xC0 vandalism thought timeout, 0x3F vandalism tiles seen
-    uint8_t voucher_type;
-    uint8_t voucher_arguments; // ride_id or string_offset_id
+    uint8_t NoOfFood;
+    uint8_t NoOfDrinks;
+    uint8_t NoOfSouvenirs;
```
duncanspumpkin, line 737: I think we decided that my `No` was a bad name for these variables. `Num` would be a better.

tupaschoal: I went with `AmountOf...` because `NumOfFood` seemed weird.

PR #11835, opened 2020-05-30
Part of #9473: Create Apply(Non)StopBlockBrake

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -6699,10 +6691,10 @@ void Vehicle::CheckAndApplyBlockSectionStopSite()
     switch (trackType)
     {
         case TRACK_ELEM_BLOCK_BRAKES:
-            if (curRide->IsBlockSectioned())
-                apply_block_brakes(this, trackElement->AsTrack()->BlockBrakeClosed());
+            if (curRide->IsBlockSectioned() && trackElement->AsTrack()->BlockBrakeClosed())
```
tupaschoal, line 6694: Nice Catch hahaha :D

PR #11836, opened 2020-05-30
Part of #9473: UpdateSwingingCar/UpdateSpinningCar

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -6945,21 +6945,21 @@ static int32_t vehicle_get_swing_amount(Vehicle* vehicle)
...
+    int32_t swingAmount = GetSwingAmount();
     if (swingAmount < 0)
     {
-        vehicle->var_4E -= dword_F64E08 >> (-swingAmount);
+        var_4E -= dword_F64E08 >> (-swingAmount);
     }
     else if (swingAmount > 0)
     {
-        vehicle->var_4E += dword_F64E08 >> swingAmount;
+        var_4E += dword_F64E08 >> swingAmount;
```
tupaschoal, line 6959: `var_4E` is probably the swinging boat position while `dword_F64E08` seems to be the increment base?

hdpoliveira:
I have a changeset with name changed in my stack right after this one. var_4E is the swing speed :)

tupaschoal:
Nice, then :D

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7135,21 +7135,21 @@ static const uint8_t TrackTypeToSpinFunction[256] = {
...
+    auto vehicleEntry = Entry();
     if (vehicleEntry == nullptr)
     {
         return;
     }
     int32_t spinningInertia = vehicleEntry->spinning_inertia;
-    int32_t trackType = vehicle->track_type >> 2;
+    int32_t trackType = track_type >> 2;
     int32_t dword_F64E08 = _vehicleVelocityF64E08;
     int32_t spinSpeed;
```
duncanspumpkin, line 7154: i know this is how it currently is but having two variables one spinSpeed and the other spin_speed is just a tad confusing. This one should have maybe been `newSpinSpeed`

duncanspumpkin: This variable is actually used for two different uses and it should have also been split into two variables.

duncanspumpkin:
I agree with the new var_4E name change can you make a PR with that change.

PR #11838, opened 2020-05-31
#9473: UpdateTrackMotionPoweredRideAcceleration

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -9448,21 +9448,21 @@ static int32_t vehicle_update_track_motion_mini_golf(Vehicle* vehicle, int32_t*
  *
  *  rct2: 0x006DC1E4
  */
-static void vehicle_update_track_motion_powered_ride_acceleration(
-    Vehicle* vehicle, rct_ride_entry_vehicle* vehicleEntry, uint32_t totalMass, int32_t* acceleration)
+void Vehicle::UpdateTrackMotionPoweredRideAcceleration(
+    rct_ride_entry_vehicle* vehicleEntry, uint32_t totalMass, int32_t* curAcceleration)
```
tupaschoal, line 9452: I think I'd stop sending the pointer and start returning the new acceleration value, I think the interface becomes clearer

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9478,38 +9478,38 @@ static void vehicle_update_track_motion_powered_ride_acceleration(
...
     }
     else if (trackType == TRACK_ELEM_RIGHT_QUARTER_TURN_1_TILE)
     {
-        speedModifier = (vehicle->TrackSubposition == VEHICLE_TRACK_SUBPOSITION_GO_KARTS_RIGHT_LANE) ? HALF_SPEED
+        speedModifier = (TrackSubposition == VEHICLE_TRACK_SUBPOSITION_GO_KARTS_RIGHT_LANE) ? HALF_SPEED
                                                                                                      : THREE_QUARTER_SPEED;
     }
 
-    uint8_t speed = vehicle->speed;
+    uint8_t curSpeed = speed;
```
duncanspumpkin, line 9494: technically this is the `newSpeed`

hdpoliveira: Not really. There is no assignment from curSpeed to speed.

The value is calculated to determine how big the acceleration will be. It is based on the current speed value and on the track.

I've renamed it to 'modifiedSpeed' and extracted a function for clarity.

hdpoliveira:
Ok I'll take a look

hdpoliveira:
@tupaschoal it is ok now 😄

PR #11839, opened 2020-05-31
[Plugin] Add API for vehicles

IntelOrca:


src/openrct2/peep/Peep.cpp
```diff
@@ -338,6 +338,21 @@ Peep* rct_sprite::AsPeep()
...
+Vehicle* rct_sprite::AsVehicle()
+{
+    Vehicle* result = nullptr;
+    if (IsVehicle())
+    {
+        return reinterpret_cast<Vehicle*>(this);
+    }
+    return result;
+}
+
```
tupaschoal, line 355: It looks so weird to have these on `Peep`, is this for the mini golf?

distribution/openrct2.d.ts
```diff
@@ -854,6 +854,123 @@ declare global {
...
+         */
+        bankRotation: number;
+
+        /**
+         * The colour of the car.
+         */
+        colours: VehicleColour;
+
+        speed: number;
+        poweredAcceleration: number;
```
tupaschoal, line 935: I feel like this one needs a comment (there are others more obvious that don't)

distribution/openrct2.d.ts
```diff
@@ -854,6 +854,123 @@ declare global {
...
+
+        status: VehicleStatus;
+        peeps: (number | null)[];
```
tupaschoal, line 938: It this an array of how many peeps are on each train or somehow an identifier of the peeps that are on the trains?

distribution/openrct2.d.ts
```diff
@@ -854,6 +854,123 @@ declare global {
...
+        "waiting_for_cable_lift" |
+        "travelling_cable_lift" |
+        "stopping" |
+        "waiting_for_passengers_17" |
+        "waiting_to_start" |
+        "starting" |
+        "operating_1a" |
+        "stopping_1b" |
+        "unloading_passengers_1c" |
+        "stopped_by_block_brake";
```
tupaschoal, line 972: If the current order doesn't offer any meaning, I'd sort this list.

src/openrct2/scripting/Duktape.hpp
```diff
@@ -219,7 +219,7 @@ namespace OpenRCT2::Scripting
             auto it = _s2n.find(k);
             if (it == _s2n.end())
             {
-                return 0;
+                return (T)0;
```
tupaschoal, line 222: why not `return {0};`?

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -176,6 +177,418 @@ namespace OpenRCT2::Scripting
...
+        { "waiting_for_cable_lift", VEHICLE_STATUS_WAITING_FOR_CABLE_LIFT },
+        { "travelling_cable_lift", VEHICLE_STATUS_TRAVELLING_CABLE_LIFT },
+        { "stopping", VEHICLE_STATUS_STOPPING },
+        { "waiting_for_passengers_17", VEHICLE_STATUS_WAITING_FOR_PASSENGERS_17 },
+        { "waiting_to_start", VEHICLE_STATUS_WAITING_TO_START },
+        { "starting", VEHICLE_STATUS_STARTING },
+        { "operating_1a", VEHICLE_STATUS_OPERATING_1A },
+        { "stopping_1b", VEHICLE_STATUS_STOPPING_1B },
+        { "unloading_passengers_1c", VEHICLE_STATUS_UNLOADING_PASSENGERS_1C },
+        { "stopped_by_block_brake", VEHICLE_STATUS_STOPPED_BY_BLOCK_BRAKES },
```
tupaschoal, line 211: At some point we should strive for a `toString()` of `fromString()` method, because if a value is added to the enum currently, we won't be able and multiple places will be left unhandled. If it's a `switch()...case` it would error out.

src/openrct2/peep/Peep.cpp
duncanspumpkin: Yes this code should have been put in vehicle.cpp.

distribution/openrct2.d.ts
IntelOrca: Any that don't have a comment is because I can't think of a good comment. Suggestions welcome.

src/openrct2/scripting/Duktape.hpp
IntelOrca: That produces errors. I have changed it to `static_cast` instead.

distribution/openrct2.d.ts
tupaschoal: Does this means the acceleration at launch?
`/* Acceleration value to be used when the train exits the station */`

src/openrct2/ride/Vehicle.cpp
```diff
@@ -805,6 +805,21 @@ Vehicle* try_get_vehicle(uint16_t spriteIndex)
...
+}
+
+Vehicle* rct_sprite::AsVehicle()
+{
+    Vehicle* result = nullptr;
+    if (IsVehicle())
+    {
+        return reinterpret_cast<Vehicle*>(this);
+    }
+    return result;
```
tupaschoal, line 820: I'd just do a ternary
```suggestion
    return IsVehicle() ? reinterpret_cast<Vehicle*>(this) : nullptr;
```

duncanspumpkin: It's going to be removed anyway when i do my is as refactor.

distribution/openrct2.d.ts
IntelOrca: @duncanspumpkin do you agree?

duncanspumpkin: Pretty sure powered acceleration is the field for vehicles that are self powered. Such as steam trains, boats, ...

IntelOrca: What is speed then?

duncanspumpkin: speed is the max speed that can be obtained from powered acceleration. We should rename it so that it matches the vehicle object name `powered_max_speed`

tupaschoal:
@IntelOrca can you please rebase it so we can merge?

IntelOrca:
Okay done.

PR #11843, opened 2020-05-31
Switch vcxproj files to remove wildcards

duncanspumpkin:
Unfortunately MS has removed the wild card feature from vcxproj files. This will mean all file deletions and creations will need to modify these files to be included in the VS builds.


src/openrct2-ui/libopenrct2ui.vcxproj
```diff
@@ -53,4 +213,4 @@
     </Lib>
   </ItemDefinitionGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-</Project>
+</Project>
```
Gymnasiast, line 216: *cough*

duncanspumpkin: The problem is vs is going to do this every time it edits the file.

tupaschoal: I'd rather leave this change and prevent us from opening a new issue every time someone edits those files on windows.

duncanspumpkin: The counter argument is that anyone not on Windows will insert the end of line when they edit the file.

tupaschoal: Will they, though? Because I remember there was one file which I kept doing this and reverting and when we settled on "letting it slip" we didn't have this problem anymore, as far as I can remember

PR #11849, opened 2020-05-31
Rename swing-related variables in Vehicle

hdpoliveira:


hdpoliveira:
Should I update the variable names in class rct1_vehicle as well?

duncanspumpkin:
Yes update it for the rct_vehicle struct as well

PR #11850, opened 2020-05-31
#9473: Create Vehicle::UpdateFlag

hdpoliveira:


duncanspumpkin:
UpdateFlag is not a great name for this function. It makes it sound like it would Update the flag  rather than the flags for the current update.

tupaschoal:
Yeah, I had mixed feelings about it too. It's the whole concept of avoiding or not things like "is", "Has", etc... The fact that the variable itself is called "update_flag" makes it more confusing. Dunno if we should go with "HasUpdateFlag", or something alike

hdpoliveira:
I chose the name because after this change I'd create "SetUpdateFlag/UnsetUpdateFlag"
The name of the variable is update_flags, so it makes sense for you to query the value of one single flag via the function "UpdateFlag"

Nevertheless, I'm ok changing it to "HasUpdateFlag" or "UpdateFlagValue" or "UpdateFlagVal".

What are the update flags anyway?

duncanspumpkin:
They are the flags that are set during the update functions.

PR #11852, opened 2020-05-31
Smart pointers

sohiladwani:
Was hoping someone could code review my code :)

### DIFF
--- a/src/openrct2/ride/TrackDesign.cpp
+++ b/src/openrct2/ride/TrackDesign.cpp
@@ -2111,7 +2111,9 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
  */
 static map_backup* track_design_preview_backup_map()
 {
-    map_backup* backup = static_cast<map_backup*>(malloc(sizeof(map_backup)));
+    //map_backup* backup = static_cast<map_backup*>(malloc(sizeof(map_backup)));
+    auto backup_ptr = std::make_unique<map_backup>();
+    map_backup* backup = backup_ptr.get();
     if (backup != nullptr)
     {
         std::memcpy(backup->tile_elements, gTileElements, sizeof(backup->tile_elements));
@@ -2138,8 +2140,6 @@ static void track_design_preview_restore_map(map_backup* backup)
     gMapSizeMinus2 = backup->map_size_units_minus_2;
     gMapSize = backup->map_size;
     gCurrentRotation = backup->current_rotation;
-
-    free(backup);
 }
 
 /**
--- a/src/openrct2/ride/TrackDesign.cpp
+++ b/src/openrct2/ride/TrackDesign.cpp
@@ -2112,18 +2112,18 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
  */
 static std::unique_ptr<map_backup> track_design_preview_backup_map()
 {
-    auto backup_ptr = std::make_unique<map_backup>();
-    if (backup_ptr != nullptr)
+    auto backup = std::make_unique<map_backup>();
+    if (backup != nullptr)
     {
-        std::memcpy(backup_ptr->tile_elements, gTileElements, sizeof(backup_ptr->tile_elements));
-        std::memcpy(backup_ptr->tile_pointers, gTileElementTilePointers, sizeof(backup_ptr->tile_pointers));
-        backup_ptr->next_free_tile_element = gNextFreeTileElement;
-        backup_ptr->map_size_units = gMapSizeUnits;
-        backup_ptr->map_size_units_minus_2 = gMapSizeMinus2;
-        backup_ptr->map_size = gMapSize;
-        backup_ptr->current_rotation = get_current_rotation();
+        std::memcpy(backup->tile_elements, gTileElements, sizeof(backup->tile_elements));
+        std::memcpy(backup->tile_pointers, gTileElementTilePointers, sizeof(backup->tile_pointers));
+        backup->next_free_tile_element = gNextFreeTileElement;
+        backup->map_size_units = gMapSizeUnits;
+        backup->map_size_units_minus_2 = gMapSizeMinus2;
+        backup->map_size = gMapSize;
+        backup->current_rotation = get_current_rotation();
     }
-    return backup_ptr;
+    return backup;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2111,7 +2112,9 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
  */
 static map_backup* track_design_preview_backup_map()
 {
-    map_backup* backup = static_cast<map_backup*>(malloc(sizeof(map_backup)));
+    // map_backup* backup = static_cast<map_backup*>(malloc(sizeof(map_backup)));
```
duncanspumpkin, line 2115: always just delete code. There is no need to comment out code.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2111,7 +2112,9 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
...
+    auto backup_ptr = std::make_unique<map_backup>();
```
duncanspumpkin, line 2116: You don't need to create a new variable name `backup` can still be accessed just the same.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2111,7 +2112,9 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
  */
 static map_backup* track_design_preview_backup_map()
```
duncanspumpkin, line 2113: You will need to modify this so that the return is a std::unique_ptr<map_backup> as we need the map_backup to survive for the whole of the calling function (`track_design_draw_preview`)

src/openrct2/world/Map.cpp
```diff
@@ -1076,8 +1077,10 @@ void map_reorganise_elements()
 {
     context_setcurrentcursor(CURSOR_ZZZ);
 
-    TileElement* new_tile_elements = static_cast<TileElement*>(malloc(MAX_TILE_ELEMENTS_WITH_SPARE_ROOM * sizeof(TileElement)));
-    TileElement* new_elements_pointer = new_tile_elements;
+    // TileElement* new_tile_elements = static_cast<TileElement*>(malloc(MAX_TILE_ELEMENTS_WITH_SPARE_ROOM *
```
duncanspumpkin, line 1080: as above just delete

duncanspumpkin:
The map backup needs a bit of work still. You need to modify the caller a bit. Did you test you're build because I think you will find it crashes when you try to place a track design as the memory has been used after freeing.

The map_reorganise_elements is good just need to delete the commented code.

tupaschoal:
Hello @sohiladwani ,
Please let us know if you'll be working on the issues pointed out. Thanks!

sohiladwani:
> Hello @sohiladwani ,
> Please let us know if you'll be working on the issues pointed out. Thanks!

@tupaschoal Yes! Hopefully by SaturdayI have them resolved sorry just been a bit swamped with work.

duncanspumpkin:
You appear to have merged develop into your branch instead of rebasing. If you don't know how to rebase let us know and we can help you out

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2109,20 +2110,20 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
...
-        std::memcpy(backup->tile_pointers, gTileElementTilePointers, sizeof(backup->tile_pointers));
-        backup->next_free_tile_element = gNextFreeTileElement;
-        backup->map_size_units = gMapSizeUnits;
-        backup->map_size_units_minus_2 = gMapSizeMinus2;
-        backup->map_size = gMapSize;
-        backup->current_rotation = get_current_rotation();
-    }
-    return backup;
+    auto backup_ptr = std::make_unique<map_backup>();
+    if (backup_ptr.get() != nullptr)
```
duncanspumpkin, line 2116: You only need to use get() when you absolutely need the pointer to the memory. None of these in this particular function require that. You can dereferrence a smart pointer as expected.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2000,7 +2001,7 @@ static bool track_design_place_preview(TrackDesign* td6, money32* cost, Ride** o
 void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
 {
     // Make a copy of the map
-    map_backup* mapBackup = track_design_preview_backup_map();
+    auto mapBackup = track_design_preview_backup_map().get();
```
duncanspumpkin, line 2004: this should not be a `get()` you want this to be the smart pointer

duncanspumpkin: L2021 and L2104 will either need to have a `get()` added or use std::move to move the smart pointer into that function and adjusting the function. I suggest just using get as it is safe to do in this context.

duncanspumpkin:
`map_reorganise_elements` is all correct nothing to change. The map backup is still not correct you need to modify the caller just a little more and remove the additional `get()`s

sohiladwani:
Think I got everything this time, I am a little confused about the map backup so i tried my best shot at understanding what you said. Hopefully its close now at least!

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2101,28 +2102,28 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
...
-        std::memcpy(backup->tile_elements, gTileElements, sizeof(backup->tile_elements));
-        std::memcpy(backup->tile_pointers, gTileElementTilePointers, sizeof(backup->tile_pointers));
-        backup->next_free_tile_element = gNextFreeTileElement;
-        backup->map_size_units = gMapSizeUnits;
-        backup->map_size_units_minus_2 = gMapSizeMinus2;
-        backup->map_size = gMapSize;
-        backup->current_rotation = get_current_rotation();
-    }
-    return backup;
+    auto backup_ptr = std::make_unique<map_backup>();
```
Gymnasiast, line 2115: Use camelCase.

src/openrct2/world/Map.cpp
```diff
@@ -1104,12 +1105,10 @@ void map_reorganise_elements()
         }
     }
 
-    num_elements = static_cast<uint32_t>(new_elements_pointer - new_tile_elements);
-    std::memcpy(gTileElements, new_tile_elements, num_elements * sizeof(TileElement));
+    num_elements = static_cast<uint32_t>(new_elements_pointer - new_tile_elements.get());
```
Gymnasiast, line 1108: Use camelCase.

duncanspumpkin:
I've updated the variable names. Squash merge on green CI's.

PR #11860, opened 2020-06-02
Create a hookable location checking function for actions

duncanspumpkin:
work in progress but this is so that you don't need to do really complex things to work out in a script if an action is working on a tile.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -224,6 +224,7 @@ declare global {
         subscribe(hook: "network.join", callback: (e: NetworkEventArgs) => void): IDisposable;
         subscribe(hook: "network.leave", callback: (e: NetworkEventArgs) => void): IDisposable;
         subscribe(hook: "ride.ratings.calculate", callback: (e: RideRatingsCalculateArgs) => void): IDisposable;
+        subscribe(hook: "action.location", callback: (e: ActionLocationArgs) => void): IDisposable;
     }
 
     interface Configuration {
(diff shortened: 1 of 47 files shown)


src/openrct2/scripting/HookEngine.h
```diff
@@ -37,6 +37,7 @@ namespace OpenRCT2::Scripting
         NETWORK_JOIN,
         NETWORK_LEAVE,
         RIDE_RATINGS_CALCULATE,
+        GAME_ACTION_LOCATION,
```
IntelOrca, line 40: I suggest removing `GAME_` so it is consistent with the other types which are just `ACTION_*`.

IntelOrca:
`map_is_location_valid` checks if it is actually a valid tile on the map. We probably don't even want to call the hook unless this is true right? I'd then perhaps set `result` to the normal checks, like whether the land is owned. And then the script can check if result is true, and override that to false if it isn't allowed.

Is this something to be called for both query and execute? How do we handle auto clearance removal, e.g. like if a tree is in the way but gets automatically removed.

duncanspumpkin:
I was imagining this being called in the query and the execute. Yes agree that the hook isn't worth calling if the check fails.

duncanspumpkin:
I've used the check on if location valid as that will be called on all tiles that are being modified. If you don't have permission due to the hook it would not allow that tile to be modified. Auto scenery removal wouldn't be an issue as the function would return false preventing that from happening.

distribution/openrct2.d.ts
```diff
@@ -350,6 +351,15 @@ declare global {
         nausea: number;
     }
 
+    interface ActionLocationArgs {
+        readonly x: number;
+        readonly y: number;
+        readonly playerId: number;
```
IntelOrca, line 357: The other network args just uses `player`.

duncanspumpkin: player for player Id?

PR #11866, opened 2020-06-03
Move calls to set_format_arg to Formatter::Common()

Gymnasiast:
This moves all remaining calls to set_format_arg() to use Formatter::Common(). This also means cleaning up many indirect usages of a C-style cast.

### DIFF
--- a/src/openrct2-ui/windows/Guest.cpp
+++ b/src/openrct2-ui/windows/Guest.cpp
@@ -1958,7 +1958,7 @@ static rct_string_id window_guest_inventory_format_item(Peep* peep, int32_t item
     switch (item)
     {
         case SHOP_ITEM_BALLOON:
-            ft.Increment(-ft.NumBytes());
+            ft.Rewind();
             ft.Add<uint32_t>(SPRITE_ID_PALETTE_COLOUR_1(peep->BalloonColour) | ShopItems[item].Image);
             break;
         case SHOP_ITEM_PHOTO:
@@ -1967,14 +1967,15 @@ static rct_string_id window_guest_inventory_format_item(Peep* peep, int32_t item
                 ride->FormatNameTo(gCommonFormatArgs + 6);
             break;
         case SHOP_ITEM_UMBRELLA:
-            ft.Increment(-ft.NumBytes());
+            ft.Rewind();
             ft.Add<uint32_t>(SPRITE_ID_PALETTE_COLOUR_1(peep->UmbrellaColour) | ShopItems[item].Image);
             break;
         case SHOP_ITEM_VOUCHER:
             switch (peep->VoucherType)
             {
                 case VOUCHER_TYPE_PARK_ENTRY_FREE:
-                    ft.Increment(-ft.NumBytes() + 6);
+                    ft.Rewind();
+                    ft.Increment(6);
                     ft.Add<rct_string_id>(STR_PEEP_INVENTORY_VOUCHER_PARK_ENTRY_FREE);
                     ft.Add<rct_string_id>(STR_STRING);
                     ft.Add<const char*>(parkName);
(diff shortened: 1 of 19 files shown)


src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1530,17 +1531,21 @@ void window_guest_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+        ft.Add<uint16_t>(peep->intensity.GetMinimum());
+        ft.Add<uint16_t>(maxIntensity);
         string_id = STR_GUEST_STAT_PREFERRED_INTESITY_BETWEEN;
         if (maxIntensity == 15)
             string_id = STR_GUEST_STAT_PREFERRED_INTESITY_ABOVE;
     }
+    else
+    {
+        ft.Add<uint16_t>(maxIntensity);
+    }
```
tupaschoal, line 1548: Alternatively
```
    if (peep->intensity.GetMinimum() != 0)
    {
        ft.Add<uint16_t>(peep->intensity.GetMinimum());
        string_id = STR_GUEST_STAT_PREFERRED_INTESITY_BETWEEN;
        if (maxIntensity == 15)
            string_id = STR_GUEST_STAT_PREFERRED_INTESITY_ABOVE;
    }
    ft.Add<uint16_t>(maxIntensity);
```

(I'd also use a ternary to simplify the `maxIntensity` stuff, but out of scope

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1553,7 +1558,7 @@ void window_guest_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
     };
     y += LIST_ROW_HEIGHT;
     int32_t nausea_tolerance = peep->nausea_tolerance & 0x3;
-    set_format_arg(0, rct_string_id, nauseaTolerances[nausea_tolerance]);
+    ft.Add<rct_string_id>(nauseaTolerances[nausea_tolerance]);
```
tupaschoal, line 1561:
```suggestion
    auto ft = Formatter::Common();
    ft.Add<rct_string_id>(nauseaTolerances[nausea_tolerance]);
```
To go back to 0

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1942,58 +1947,67 @@ static rct_string_id window_guest_inventory_format_item(Peep* peep, int32_t item
...
+    ft.Add<rct_string_id>(STR_STRING);
+    ft.Add<const char*>(parkName);
 
     // Special overrides
     Ride* ride{};
     switch (item)
     {
         case SHOP_ITEM_BALLOON:
-            set_format_arg(0, uint32_t, SPRITE_ID_PALETTE_COLOUR_1(peep->BalloonColour) | ShopItems[item].Image);
+            ft.Increment(-ft.NumBytes());
```
tupaschoal: Argh this is very hackish X_X
I'm going to think if there's a better way. Seems like here, and on other `case`s you're just trying to rewrite part of what you wrote before the `switch`. I'd save each of these arguments:
```
    ft.Add<uint32_t>(ShopItems[item].Image);
    ft.Add<rct_string_id>(ShopItems[item].Naming.Display);
    ft.Add<rct_string_id>(STR_STRING);
    ft.Add<const char*>(parkName);
```
in a corresponding variable, and update them on the cases, and then actually call:
```
    ft.Add<uint32_t>(ShopItems[item].Image);
    ft.Add<rct_string_id>(ShopItems[item].Naming.Display);
    ft.Add<rct_string_id>(STR_STRING);
    ft.Add<const char*>(parkName);
```
after the switch to get the properly formatted message with decrementing

src/openrct2/world/Map.cpp
```diff
@@ -1254,10 +1254,13 @@ void map_obstruction_set_error_text(TileElement* tileElement, GameActionResult&
             }
             break;
         case TILE_ELEMENT_TYPE_SMALL_SCENERY:
+        {
```
tupaschoal, line 1257: The changes in this file seems like they belong to another commit

src/openrct2/localisation/Localisation.h
```diff
@@ -112,11 +112,16 @@ class Formatter
         return CurrentBuf;
     }
 
-    void Increment(std::size_t count)
+    void Increment(size_t count)
```
tupaschoal, line 115: Why dropping the `std`?

src/openrct2/localisation/Localisation.h
```diff
@@ -112,11 +112,16 @@ class Formatter
...
     {
         CurrentBuf += count;
     }
 
+    void Rewind()
+    {
+        CurrentBuf -= NumBytes();
+    }
+
```
tupaschoal, line 124: Maybe?
```suggestion
    void Reset()
    {
        CurrentBuf = StartBuf;
    }

```

src/openrct2/localisation/Localisation.h
IntelOrca: Where else do we have `std::size_t`? I think it is just making things consistent.

src/openrct2-ui/windows/Guest.cpp
Gymnasiast: This had already been changed when you wrote this...

src/openrct2/localisation/Localisation.h
Gymnasiast: Does not compile.

src/openrct2/localisation/Localisation.h
Gymnasiast: I temporarily changed it to a `ssize_t` before opting for another solution. Could be changed back.

src/openrct2/world/Map.cpp
Gymnasiast: Er, no. Creating a variable in a `case` requires { and }.

duncanspumpkin:
Rewind seems fine but would prefer we don't use it. For most uses we don't need it but we can fix that eventually.

IntelOrca:
It is a bit counter-intuitive. You basically add arguments to a list, so if you can rewind back, add is no longer inserting them but overwriting old ones. I think each case should just build its own format args rather than build like a default one then overwrite bits of it.

src/openrct2/localisation/Localisation.h
duncanspumpkin: It doesn't compile because the constness of StartBuf is incorrect. It should be a const pointer not a pointer to const data.  `uint8_t* const StartBuf;`. I leave it to a later PR to correct this.

src/openrct2/world/Map.cpp
tupaschoal: No no, I meant that the changes you made in this file are not using `Formatter::Common()`, so it's a bit misleading with the PR title and commit messages. I'd just cherry pick the changes in here to another commit and rephrase the PR

src/openrct2-ui/windows/Guest.cpp
tupaschoal: I did review it commit by commit, maybe that's why

src/openrct2/world/Map.cpp
tupaschoal: :(

tupaschoal:
Part of #11389

PR #11867, opened 2020-06-03
Convert most remaining C-style casts to C++-style ones

Gymnasiast:


src/openrct2/PlatformEnvironment.cpp
```diff
@@ -124,35 +124,38 @@ std::unique_ptr<IPlatformEnvironment> OpenRCT2::CreatePlatformEnvironment()
 
     // Set default paths
     std::string basePaths[DIRBASE_COUNT];
-    basePaths[(size_t)DIRBASE::OPENRCT2] = Platform::GetInstallPath();
-    basePaths[(size_t)DIRBASE::USER] = Path::Combine(Platform::GetFolderPath(SPECIAL_FOLDER::USER_DATA), subDirectory);
-    basePaths[(size_t)DIRBASE::CONFIG] = Path::Combine(Platform::GetFolderPath(SPECIAL_FOLDER::USER_CONFIG), subDirectory);
-    basePaths[(size_t)DIRBASE::CACHE] = Path::Combine(Platform::GetFolderPath(SPECIAL_FOLDER::USER_CACHE), subDirectory);
-    basePaths[(size_t)DIRBASE::DOCUMENTATION] = Platform::GetDocsPath();
+    basePaths[static_cast<size_t>(DIRBASE::OPENRCT2)] = Platform::GetInstallPath();
```
AaronVanGeffen, line 127: This does not look like an improvement, or even desirable. Perhaps basePaths should just be a `std::map` instead?

Gymnasiast: Can't say I really agree. The point is to move away from C-style casts to more clearly defined behaviour, not to make it look pretty.

That said, you're welcome to push any improvements to this PR.

tupaschoal: I'm with @Gymnasiast here, we can change that afterwards, it's a bit out of scope in here

src/openrct2/drawing/Drawing.String.cpp
```diff
@@ -313,7 +313,7 @@ void draw_string_centred_raw(rct_drawpixelinfo* dpi, int32_t x, int32_t y, int32
 {
     ScreenCoordsXY screenCoords(dpi->x, dpi->y);
     gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
-    gfx_draw_string(dpi, (char*)"", COLOUR_BLACK, screenCoords);
+    gfx_draw_string(dpi, const_cast<char*>(""), COLOUR_BLACK, screenCoords);
```
ZehMatt, line 316: I dont think a cast is required, passing ```""``` should be of type ```const char*```

duncanspumpkin: Its removing the const I believe. Really all of the draw_strings need to be modified to take const.

ZehMatt: I looked at it, it already takes ```const_utf8string```

Issue #11523, opened 2020-04-28
Cannot place signs or banners in multiplayer after a while, until server gets restarted


deurklink:
**OS:** [Windows 10]
**Version:** [0.2.6]
**Commit/Build:** [06bbf5d]

A problem that has been around since 0.2.4, but couldn't find an issue for it. In multiplayer, after a while, no more banners or signs can be placed anymore. If you try to place one, nothing happens, even on the host. I don't know what triggers this state, but a server restart always fixes it.

duncanspumpkin:
I thought #9623 caused this but I'm not convinced anymore. I found a few instances where the allocated banner is not freed but none of those situations seem to be likely to be the cause of this issue. I'm not sure what to suggest to fix this.

giantmustache:
My friend just started a completely new server and I joined as a second player. From the start of the game I was unable to place signs. Tested with commit 8f29810.

duncanspumpkin:
That is interesting. Perhaps points to a bug in the new scenario code.

giantmustache:
We were playing crazy castle, might that help. I don't know if OP was also playing that scenario.

tylermcbride:
Having this issue. Only the host can place signs.

Gymnasiast:
Is this still an issue?

PR #11880, opened 2020-06-06
Banners Leak Protection

duncanspumpkin:
Maybe fix for https://github.com/OpenRCT2/OpenRCT2/issues/11523
Unsure how often players actually delete banners with the tile inspector.

### DIFF
--- a/src/openrct2/world/TileInspector.cpp
+++ b/src/openrct2/world/TileInspector.cpp
@@ -141,6 +141,52 @@ GameActionResult::Ptr tile_inspector_insert_corrupt_at(const CoordsXY& loc, int1
     return std::make_unique<GameActionResult>();
 }
 
+static int32_t numLargeScenerySequences(const CoordsXY& loc, const LargeSceneryElement* const largeScenery)
+{
+    const rct_scenery_entry* const largeEntry = largeScenery->GetEntry();
+    const auto* const tiles = largeEntry->large_scenery.tiles;
+    const auto direction = largeScenery->GetDirection();
+
+    const auto rotatedFirstTile = CoordsXYZ{
+        CoordsXY{ tiles[largeScenery->GetSequenceIndex()].x_offset, tiles[largeScenery->GetSequenceIndex()].y_offset }.Rotate(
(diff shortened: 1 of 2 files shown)


src/openrct2/world/TileInspector.cpp
```diff
@@ -157,6 +203,22 @@ GameActionResult::Ptr tile_inspector_remove_element_at(const CoordsXY& loc, int1
         {
             return std::make_unique<GameActionResult>(GA_ERROR::UNKNOWN, STR_NONE);
         }
+
+        auto largeScenery = tileElement->AsLargeScenery();
+        if (largeScenery)
+        {
+            // Only delete the banner entry if there are no other parts of the large scenery to delete
+            if (numLargeScenerySequences(loc, largeScenery) == 1)
```
HMorsinkGD, line 211: Can't test right now, but wouldn't this return 1 and thus remove the banner entry for large banner elements that have been rotated using the tile inspector?

Broxzier, line 211: Can't test right now, but wouldn't this return 1 and thus remove the banner entry for large banner elements that have been rotated using the tile inspector?

duncanspumpkin: Hmm never thought someone would do that.

duncanspumpkin: What do you propose as a solution. Should I check all 4 different rotations for the scenery?

duncanspumpkin: Ah well thats easy large scenery can't rotate!

Broxzier: > What do you propose as a solution. Should I check all 4 different rotations for the scenery?

Yeah, that'd be enough for 2x1 banners. If we later add new banner elements, that'd require another solution. If iterating over all elements on the map is not too expensive, that might be the best thing to do here.

> Ah well thats easy large scenery can't rotate!

Ah, that does sound like a missing feature. For the Goliath letters that would be useful, for instance.

PR #11881, opened 2020-06-06
Implement Sprite Is As functions

duncanspumpkin:
Okay think this might be the 3rd attempt at doing this. This time taking it in nice easy chunks.

Yes all this `->generic.Is` / `->generic.As` is horrible but it will be removed as the functions use the `SpriteBase` instead of `rct_sprite`. I've already started another branch that converts get_sprite to return a specific type or a SpriteBase.

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -288,7 +288,7 @@ rct_sprite_checksum sprite_checksum()
                         break;
                 }
 
-                if (copy.IsPeep())
+                if (copy.generic.Is<Peep>())
                 {
                     // Name is pointer and will not be the same across clients
                     copy.peep.name = {};
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -888,7 +888,7 @@ uint16_t remove_floating_sprites()
         }
         else if (rctSprite->generic.Is<Duck>())
         {
-            if (rctSprite->AsDuck()->IsFlying())
+            if (rctSprite->generic.As<Duck>()->IsFlying())
             {
                 rctSprite->duck.Remove();
                 sprite_misc_update(rctSprite);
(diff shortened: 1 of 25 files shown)


duncanspumpkin:
I've had a look at providing a `const` version of As. Its not too easy to do without just copying the code a second time so I propose adding the following as well to allow for a const version:
```
    template<typename T> const T* As() const
    {
        T* result = nullptr;
        if (Is<T>())
        {
            result = reinterpret_cast<const T*>(this);
        }
        return result;
    }
```

src/openrct2/world/SpriteBase.h
```diff
@@ -33,6 +33,16 @@ struct SpriteBase
...
     void MoveTo(const CoordsXYZ& newLocation);
+    template<typename T> bool Is() const;
+    template<typename T> T* As()
+    {
+        T* result = nullptr;
+        if (Is<T>())
+        {
+            result = reinterpret_cast<T*>(this);
+        }
+        return result;
```
tupaschoal, line 44: Personally I'd just use ternary, but this is a personal preference
```suggestion
        return Is<T>() ? reinterpret_cast<T*>(this) : result;
```

src/openrct2/world/Sprite.cpp
```diff
@@ -897,7 +897,7 @@ uint16_t remove_floating_sprites()
         }
         else if (rctSprite->generic.Is<MoneyEffect>())
         {
-            sprite_remove(rctSprite->AsMoneyEffect());
+            sprite_remove(&rctSprite->generic);
```
tupaschoal, line 900: This is on the "Remove AsBalloon" commit, but deals with `MoneyEffect`

duncanspumpkin:
As said the generic part of this will be removed in the next PR. Its all part of removing `rct_sprite`. I've already made all of those changes.

src/openrct2/world/Sprite.cpp
duncanspumpkin: yeah I didn't see the point converting it only to unconvert it so changed it all at the same time.

PR #11882, opened 2020-06-06
Rename some Peep member variables to use TitleCase

tupaschoal:
Continues #10653

It might be easier to review on a commit by commit basis

### DIFF
--- a/src/openrct2/peep/Peep.h
+++ b/src/openrct2/peep/Peep.h
@@ -669,7 +669,7 @@ struct Peep : SpriteBase
     };
     union
     {
-        uint8_t maze_last_edge;
+        uint8_t MazeLastEdge;
         Direction PeepDirection; // Direction ?
     };
     uint8_t InteractionRideIndex;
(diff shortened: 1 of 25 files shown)


src/openrct2/peep/Peep.h
```diff
@@ -664,27 +664,27 @@ struct Peep : SpriteBase
     uint8_t step_progress;
     union
     {
-        uint16_t mechanic_time_since_call; // time getting to ride to fix
-        uint16_t next_in_queue;
+        uint16_t MechanicTimeSinceCall; // time getting to ride to fix
+        uint16_t NextInQueue;
```
duncanspumpkin, line 668: Add a `Guest` prefix to this one.

src/openrct2/peep/Peep.h
```diff
@@ -664,27 +664,27 @@ struct Peep : SpriteBase
     uint8_t step_progress;
     union
     {
-        uint16_t mechanic_time_since_call; // time getting to ride to fix
-        uint16_t next_in_queue;
+        uint16_t MechanicTimeSinceCall; // time getting to ride to fix
```
duncanspumpkin, line 667: It would have been better design by CS if he had combined this with `TimeInQueue` but oh well.

src/openrct2/peep/Peep.h
tupaschoal: Done

PR #11885, opened 2020-06-06
Create methods XY and XYZ in SpriteBase

hdpoliveira:


duncanspumpkin:
I would prefer we just have a CoordsXYZ type in the sprite base.

hdpoliveira:
There is a difference here though, these functions return a new Coords object, which the caller can change at his will.

I think the functions are as necessary as the type in SpriteBase.

Also, this change makes it easier to do the change in SpriteBase (less places to modify)

PR #11892, opened 2020-06-07
GetSprite method instead of get_sprite

duncanspumpkin:
Returns a SpriteBase or a specified type.
Continues https://github.com/OpenRCT2/OpenRCT2/pull/11881

### DIFF
--- a/src/openrct2/paint/sprite/Paint.Sprite.cpp
+++ b/src/openrct2/paint/sprite/Paint.Sprite.cpp
@@ -35,12 +35,6 @@ void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t
         return;
     }
 
-    uint16_t sprite_idx = sprite_get_first_in_quadrant(x, y);
-    if (sprite_idx == SPRITE_INDEX_NULL)
-    {
-        return;
-    }
-
     rct_drawpixelinfo* dpi = &session->DPI;
     if (dpi->zoom_level > 2)
     {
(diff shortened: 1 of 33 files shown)


tupaschoal:
This can now be rebased as the other was merged

src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -49,22 +49,22 @@ void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t
 
     const bool highlightPathIssues = (session->ViewFlags & VIEWPORT_FLAG_HIGHLIGHT_PATH_ISSUES);
 
-    for (const rct_sprite* spr = get_sprite(sprite_idx); sprite_idx != SPRITE_INDEX_NULL;
-         sprite_idx = spr->generic.next_in_quadrant)
+    for (; sprite_idx != SPRITE_INDEX_NULL;)
```
tupaschoal, line 52: Well, this is basically a `while (sprite_idx != SPRITE_INDEX_NULL)` now, isn't it?

src/openrct2/peep/Peep.cpp
```diff
@@ -2678,13 +2678,14 @@ static void peep_footpath_move_forward(Peep* peep, int16_t x, int16_t y, TileEle
...
-    for (rct_sprite* sprite; sprite_id != SPRITE_INDEX_NULL; sprite_id = sprite->generic.next_in_quadrant)
+    for (uint16_t sprite_id = sprite_get_first_in_quadrant(x, y); sprite_id != SPRITE_INDEX_NULL;)
     {
-        sprite = get_sprite(sprite_id);
-        if (sprite->generic.Is<Peep>())
+        auto sprite = GetSprite(sprite_id);
+        sprite_id = sprite->next_in_quadrant;
+        auto other_peep = sprite->As<Peep>();
+        auto litter = sprite->As<Litter>();
+        if (other_peep != nullptr)
```
tupaschoal, line 2687: if the variable is only used within the scope, I think it looks clearer to do:
```c++
if (auto other_peep = sprite->As<Peep>())
    ...
else if (auto litter = sprite->As<Litter>())
    ...
```

src/openrct2/paint/sprite/Paint.Sprite.cpp
duncanspumpkin: I was trying to get all the quadrant loops to the same style and then replace them with some sort of ranged for loop. But yes this one is probably special.

duncanspumpkin:
> I find the changes of for loops to not contain the increment and sometimes the initialization a bit of a step back

The old version was equally hacky. I want them all in the same format so that in a later refactor they are easier to change. This is not the end of these refactors.

tupaschoal:
Some of the Linux CIs are failing

duncanspumpkin:
Yup I'm going to have to do a few other little refactors cause I don't want to go add nullptr checks everywhere.

src/openrct2/scripting/ScMap.hpp
```diff
@@ -131,49 +131,35 @@ namespace OpenRCT2::Scripting
...
+            for (auto spriteId = gSpriteListHead[targetList]; spriteId != SPRITE_INDEX_NULL;)
             {
-                auto sprite = get_sprite(spriteId);
-                if (sprite == nullptr)
-                {
-                    break;
-                }
-                else
+                auto sprite = GetEntity(spriteId);
+                spriteId = sprite->next;
```
duncanspumpkin, line 138: This will not work change

src/openrct2/world/Sprite.cpp
```diff
@@ -348,21 +360,14 @@ static void sprite_reset(SpriteBase* sprite)
...
-    uint16_t spriteIndex, nextSpriteIndex;
-
-    spriteIndex = gSpriteListHead[SPRITE_LIST_FREE];
-    while (spriteIndex != SPRITE_INDEX_NULL)
+    for (uint16_t spriteIndex = gSpriteListHead[SPRITE_LIST_FREE]; spriteIndex != SPRITE_INDEX_NULL;)
     {
-        sprite = &get_sprite(spriteIndex)->generic;
-        nextSpriteIndex = sprite->next;
+        auto sprite = GetEntity(spriteIndex);
+        spriteIndex = sprite->next;
```
duncanspumpkin, line 366: This will not work change

Issue #11868, opened 2020-06-03
Add JSDoc to all OpenRCT2 APIs


thyeggman:
There are large portions of (openrct2.d.ts)[https://github.com/OpenRCT2/OpenRCT2/blob/develop/distribution/openrct2.d.ts] that are missing doc comments.  Having these would help immensely for plug-in developers.

Gymnasiast:
Do you want to have a go? ;-)

thyeggman:
I have an idea for a plugin I'd like to write but am unfamiliar with the API so I'll give it a shot :)

PR #11894, opened 2020-06-07
Adding javadoc to plugin API

thyeggman:
First draft for #11868 

If anyone wants to answer questions I have in the TODOs it would save me some research!

Also open to any comments on formatting.  I just chose something that I thought looked good, but it wasn't consistent before.  I still need to go through from the beginning once more and make everything consistent but if anyone has comments before I do that would be great.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -64,30 +64,28 @@ declare global {
     }
 
     /**
-     * A coordinate within the game.
-     * Each in-game tile is a size of 32x32.
+     * A 2D coordinate within the game map.
      */
     interface CoordsXY {
+        /** The x location on the map. Each in-game tile is a size of 32x32 units. */
         x: number;
+        /** The y location on the map. Each in-game tile is a size of 32x32 units. */
         y: number;
     }
 
     /**
-     * A coordinate within the game.
-     * Each in-game tile is a size of 32x32.
-     * The z-coordinate raises 16 per land increment. A full-height wall is 32 in height.
+     * A 3D coordinate within the game map.
      */
     interface CoordsXYZ extends CoordsXY {
+        /** The z location on the map. One land increment is 16 z units. A full-height wall is 32 z units. */
         z: number;
     }
 
     /**
-     * A coordinate within the game.
-     * Each in-game tile is a size of 32x32.
-     * The z-coordinate raises 16 per land increment. A full-height wall is 32 in height.
-     * The direction is between 0 and 3.
+     * A 3D coordinate within the game map which includes direction.
      */
     interface CoordsXYZD extends CoordsXYZ {
+        /** A number between 0 and 3 indicating direction. */
         direction: number;
     }
 
@@ -100,28 +98,42 @@ declare global {
     }
 
     /**
-     * Represents information about the plugin such as type, name, author and version.
-     * It also includes the entry point.
+     * Information about the plugin such as type, name, author, version, and entry point.
      */
     interface PluginMetadata {
+        /** The plugin name. */
         name: string;
+        /** The plugin version. Consider using semantic versioning (semver.org). */
         version: string;
+        /** The plugin author(s). */
         authors: string | string[];
+        /** The plugin type. */
         type: PluginType;
+        /** The minimum API version required by the plugin. */
         minApiVersion?: number;
+        /** The entry point for the plugin. */
         main: () => void;
     }
 
     /**
-     * Console APIs
-     * Currently interact with stdout.
+     * Console APIs. Currently they only interact with stdout.
      */
     interface Console {
+        /**
+         * Clear the console.
+         */
         clear(): void;
+
+        /**
+         * Log a message in the console. TODO can this execute commands too?
+         * 
+         * @param message A message or content to print to the console.
+         * @param optionalParams An array of parameters to pass to the console. TODO?
+         */
         log(message?: any, ...optionalParams: any[]): void;
 
         /**
-         * Executes a command using the legacy console REPL. This should not be used
+         * Execute a command using the legacy console REPL. This should not be used
          * by plugins, and exists only for servers to continue using old commands until
          * all functionality can be accomplished with this scripting API.
          *
@@ -153,35 +165,64 @@ declare global {
         /**
          * Render the current state of the map and save to disc.
          * Useful for server administration and timelapse creation.
+         * 
          * @param options Options that control the capture and output file.
          */
         captureImage(options: CaptureOptions): void;
 
         /**
-         * Gets the loaded object at the given index.
+         * Get the loaded object at the given index.
+         * 
          * @param type The object type.
-         * @param index The index.
+         * @param index The index. TODO?
          */
         getObject(type: ObjectType, index: number): Object;
+
+        /**
+         * Get the loaded ride object at the given index.
+         * 
+         * @param type The "ride" object type.
+         * @param index The index. TODO?
+         */
         getObject(type: "ride", index: number): RideObject;
+
+        /**
+         * Get the loaded small scenery object at the given index.
+         * 
+         * @param type The "small_cenery" object type.
+         * @param index The index. TODO?
+         */
         getObject(type: "small_scenery", index: number): SmallSceneryObject;
 
+        /**
+         * Get all loaded objects of the given type.
+         * 
+         * @param type The object type.
+         */
         getAllObjects(type: ObjectType): Object[];
+
+        /**
+         * Get all loaded objects of the "ride" object type.
+         * 
+         * @param type The "ride" object type.
+         */
         getAllObjects(type: "ride"): RideObject[];
 
         /**
-         * Gets a random integer within the specified range using the game's pseudo-
+         * Get a random integer within the specified range using the game's pseudo-
          * random number generator. This is part of the game state and shared across
          * all clients, you therefore must be in a context that can mutate the game
          * state. Use this to generate random numbers instead of Math.Random during
          * game logic routines such as hooks and game actions.
+         * 
          * @param min The minimum value inclusive.
          * @param max The maximum value exclusive.
          */
         getRandom(min: number, max: number): number;
 
         /**
-         * Registers a new game action that allows clients to interact with the game.
+         * Register a new game action that allows clients to interact with the game.
+         * 
          * @param action The unique name of the action.
          * @param query Logic for validating and returning a price for an action.
          * @param execute Logic for validating and executing the action.
@@ -195,6 +236,7 @@ declare global {
         /**
          * Query the result of running a game action. This allows you to check the outcome and validity of
          * an action without actually executing it.
+         * 
          * @param action The name of the action.
          * @param args The action parameters.
          * @param callback The function to be called with the result of the action.
@@ -204,6 +246,7 @@ declare global {
         /**
          * Executes a game action. In a network game, this will send a request to the server and wait
          * for the server to reply.
+         * 
          * @param action The name of the action.
          * @param args The action parameters.
          * @param callback The function to be called with the result of the action.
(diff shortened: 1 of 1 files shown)


distribution/openrct2.d.ts
```diff
@@ -307,46 +475,71 @@ declare global {
...
 
+    /** The result of the game action. */
     interface GameActionResult {
+        /** TODO? */
         error?: number;
+        /** TODO? */
         errorTitle?: string;
+        /** The error message which is displayed in a pop-up, if specified. */
         errorMessage?: string;
+        /** The position of the error message, if any. TODO position of the error or just the position of the action? */
```
duncanspumpkin, line 500: It is the centre of the action (i.e. if the land modify action was 4x4 it would be the centre of the 4x4) or if the action does not make sense to have a position this will be null (This at least is how it works in C++ @IntelOrca  will need to confirm that is still how it works in plugins)

distribution/openrct2.d.ts
```diff
@@ -307,46 +475,71 @@ declare global {
...
         readonly args: object;
+        /** The result of the action. */
         result: GameActionResult;
     }
 
+    /** The result of the game action. */
     interface GameActionResult {
+        /** TODO? */
         error?: number;
+        /** TODO? */
```
duncanspumpkin, line 496: All error messages are comprised of a title and a message (and message arguments). Title is generally the same for all of an action type. Such as all land modify actions would have a title like "Land cannot be modified..." this is then followed by the message that could be "land not owned!".

distribution/openrct2.d.ts
```diff
@@ -307,46 +475,71 @@ declare global {
...
         readonly isClientOnly: boolean;
+        /** TODO */
         readonly args: object;
+        /** The result of the action. */
         result: GameActionResult;
     }
 
+    /** The result of the game action. */
     interface GameActionResult {
+        /** TODO? */
```
duncanspumpkin, line 494: This is a categorisation of the error. We should add an enum for it. Types can be `NOT_OWNED`, `INVALID_PARAMETER`, ...

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
      * APIs for the map.
      */
     interface GameMap {
+        /** The total size of the map. TODO in tiles? */
         readonly size: CoordsXY;
+        /** The number of rides on the game map. */
         readonly numRides: number;
+        /** The number of objects on the game map. */
```
duncanspumpkin, line 592: `entities`, think Guests, Vehicles, Ducks,...

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
+         * 
+         * @param x The x-coordinate of the tile.
+         * @param y The y-coordinate of the tile.
+         * 
+         * @return The game tile. TODO what if OOB?
+         */
         getTile(x: number, y: number): Tile;
+
+        /**
+         * Get the object with the given id. TODO what if doesn't exist?
```
duncanspumpkin, line 616: entity

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
+         * 
+         * @param id The object id.
+         * 
+         * @return The object. TODO what if doesn't exist?
+         */
         getEntity(id: number): Entity;
+
+        /**
+         * Get all map objects of the given type.
```
duncanspumpkin, line 625: entities

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
+         * 
+         * @param type The object type.
+         * 
+         * @return An array of objects.
+         */
         getAllEntities(type: EntityType): Entity[];
+
+        /**
+         * Get all peep objects.
```
duncanspumpkin, line 634: entities

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
+         * 
+         * @param type The "peep" object type.
```
duncanspumpkin, line 636: entity type

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
 
+    /**
+     * A surface element.
+     */
     interface SurfaceElement extends BaseTileElement {
+        /** TODO what does the number mean? */
         slope: number;
+        /** A number indicating the surface style. */
         surfaceStyle: number;
+        /** A number indicating the edge style (blended or not?) TODO: verify */
```
duncanspumpkin, line 673: This is an object index that refers to an edge object.

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         edgeStyle: number;
+        /** The surface height of the water. */
         waterHeight: number;
+        /** The length of the grass from X to Y. TODO */
         grassLength: number;
+        /** A number from x to y indicating the ownership status. TODO */
```
duncanspumpkin, line 679: In C++ this is in terms of flags @IntelOrca to confirm how this works in the plugins

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
     interface FootpathElement extends BaseTileElement {
+        /** A number indicating the footpath type. TODO */
         footpathType: number;
+        /** A number indicating the connection points on the path. TODO */
         edgesAndCorners: number;
+        /** A number between x and y indicating the slope direction, or null for a flat path. TODO */
         slopeDirection: number | null;
+        /** True if the path has a crossing which is occupied (e.g. by a train), false otherwise. */
         isBlockedByVehicle: boolean;
+        /** TODO? Is this for better pathfinding? */
```
duncanspumpkin, line 702: No its for bad pathfinding :P. This should never be adjusted as it is recalculated every couple ticks. Its used in the pathfinding to simplify the search space but its very hacky and one day will hopefully be removed.

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         isWide: boolean;
 
+        /** Indicates if the path is a queue. */
         isQueue: boolean;
+        /** A number between x and y indicating the queue banner direction, or null if no banner is present. TODO */
         queueBannerDirection: number | null;
+        /** TODO? */
```
duncanspumpkin, line 709: If it is a queue this is the ride index of the queue.

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         ride: number;
+        /** TODO? */
```
duncanspumpkin, line 711: If it is a queue this is the station index of the queue

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         station: number;
 
+        /** TODO? */
```
duncanspumpkin, line 714: Path additions such as benches and bins. Index of the type.

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         addition: number | null;
+        /** TODO? */
```
duncanspumpkin, line 716: broken benches

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         isAdditionBroken: boolean;
     }
 
+    /**
+     * A track element.
+     */
     interface TrackElement extends BaseTileElement {
+        /** A number indicating the track type. */
         trackType: number;
+        /** TODO? */
```
duncanspumpkin, line 726: multi tile track elements such as turns have a sequence number from 0 -> number of tiles that are occupied.

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         sequence: number;
+        /** The id of the ride the track belongs to. */
         ride: number;
+        /** TODO? */
```
duncanspumpkin, line 730: 0 - 3 for which station the track belongs to.

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         station: number;
+        /** True if the track has a chain lift, false otherwise. */
         hasChainLift: boolean;
     }
 
+    /**
+     * A scenery element a single tile or smaller.
+     */
     interface SmallSceneryElement extends BaseTileElement {
+        /** TODO? */
```
duncanspumpkin, line 740: object index

distribution/openrct2.d.ts
```diff
@@ -387,96 +585,211 @@ declare global {
...
         object: number;
+        /** The number from x to y indicating the primary color. TODO */
         primaryColour: number;
+        /** The number from x to y indicating the secondary color. TODO */
         secondaryColour: number;
     }
 
+    /** An entrance (or exit?) TODO */
```
duncanspumpkin, line 748: entrance, exit, park entrance

distribution/openrct2.d.ts
```diff
@@ -546,33 +884,47 @@ declare global {
...
         readonly description: string;
-        /**
-         * A text description describing the capacity of the ride in the player's current language.
-         */
+        /** A text description describing the capacity of the ride in the player's current language. */
         readonly capacity: string;
-
+        /** The ride object flags. TODO */
         readonly flags: number;
+        /** The ride type. TODO why is this an array */
```
duncanspumpkin, line 893: All rides have up to 3 different ride types they can be associated with

distribution/openrct2.d.ts
```diff
@@ -546,33 +884,47 @@ declare global {
...
         readonly rideType: number[];
+        /** The minimum number of cars per train. TODO w/ or w/o cheats? */
         readonly minCarsInTrain: number;
+        /** The maximum number of cars per train. TODO w/ or w/o cheats? */
         readonly maxCarsInTrain: number;
+        /** The maximum (?) number of cars per flat ride. TODO */
         readonly carsPerFlatRide: number;
+        /** TODO? */
         readonly zeroCars: number;
+        /** The vehicle shown in the train tab preview? TODO */
```
duncanspumpkin, line 903: yes

distribution/openrct2.d.ts
```diff
@@ -546,33 +884,47 @@ declare global {
...
         readonly vehicles: RideObjectVehicle[];
+        /** The exitement multiplier. */
         readonly excitementMultiplier: number;
+        /** The intensity multiplier. */
         readonly intensityMultiplier: number;
+        /** The nausea multiplier. */
         readonly nauseaMultiplier: number;
+        /** The max support height for the ride. TODO units */
         readonly maxHeight: number;
+        /** If the ride is a shop, a number indicating the item sold. TODO on ride photos? */
```
duncanspumpkin, line 925: on ride photo or food item or whatever the shop sells.

distribution/openrct2.d.ts
thyeggman: This would be a breaking change but should it ideally be `readonly`?

distribution/openrct2.d.ts
```diff
@@ -654,202 +1011,141 @@ declare global {
...
-        value: number;
     }
 
+    /** Indicates ride, stall, or facility. */
     type RideClassification = "ride" | "stall" | "facility";
 
+    /** Indicates operating status (open, closed, testing, or simulating). */
     type RideStatus = "closed" | "open" | "testing" | "simulating";
 
+    /** A track color scheme. */
```
duncanspumpkin, line 1093: british english `colour`

distribution/openrct2.d.ts
```diff
@@ -654,202 +1011,141 @@ declare global {
...
         supports: number;
     }
 
+    /** A color scheme used for a single vehicle. */
     interface VehicleColour {
+        /** The body color of the vehicle. TODO value */
         body: number;
+        /** The trim color of the vehicle. TODO value */
         trim: number;
+        /** The third color of the vehicle. TODO value, also I think this is intended to be tertiary */
```
duncanspumpkin, line 1109: yes it should be `tertiary`

distribution/openrct2.d.ts
IntelOrca: Returns null if it doesn't exist.

distribution/openrct2.d.ts
IntelOrca: Yeah I haven't changed the field from C++.

distribution/openrct2.d.ts
```diff
@@ -654,202 +1011,141 @@ declare global {
...
         ternary: number;
     }
 
+    /**
+     * Information associated with a ride station.
+     */
     interface RideStation {
+        /** The coordinate of the first tile of the station, in tiles. TODO start = has lights? */
```
duncanspumpkin, line 1117: yes the track element that has the lights

distribution/openrct2.d.ts
```diff
@@ -654,202 +1011,141 @@ declare global {
...
         exit: CoordsXYZD;
     }
 
+    /** The type of a entity that may move (car, duck, peep). */
     type EntityType =
         "car" | "duck" | "peep";
 
     /**
-     * Represents an object "entity" on the map that can typically moves and has a sub-tile coordinate.
+     * Represents an object "entity" on the map that can may move and has a sub-tile coordinate.
```
duncanspumpkin, line 1132: can typically move

distribution/openrct2.d.ts
thyeggman: Is that "leads to" or "departs from"?  Eg. s0->t1->t2->s1->t3->t4->(s0) -- Do t1 and t2 belong to s0 or s1?

duncanspumpkin: t1 and t2 in that example would belong to s0 i believe

distribution/openrct2.d.ts
thyeggman: @IntelOrca I assume that's the same for most/all OOB cases?

duncanspumpkin:
I know i just wrote a comment saying 0-3 for the stations but that will be possibly be raised for the new save format. I'm not sure if that should be mentioned or not. I guess we can just go back and correct the api docs but just to make sure users don't assume that the limit is always going to be 0-3

distribution/openrct2.d.ts
thyeggman: I'll double-check before resolving this

distribution/openrct2.d.ts
```diff
@@ -1359,158 +1795,369 @@ declare global {
...
         row: number;
+        /** The column number of the selected cell indexed from 0. */
         column: number;
     }
 
+    /** 
+     * Represents a list view widget and callbacks for interacting with the list.
+     */
     interface ListView extends Widget {
+        /** (Optional) The scollbar type to use for the list. TODO default? */
```
oli414, line 1941: Vertical

distribution/openrct2.d.ts
```diff
@@ -1359,158 +1795,369 @@ declare global {
...
         viewport: Viewport
     }
 
+    /**
+     * Represents a game window.
+     */
     interface Window {
+        /** TODO? */
         classification: number;
+        /** TODO? */
```
oli414, line 2003: I believe this is a unqiue identifier

distribution/openrct2.d.ts
Gymnasiast: Are we talking about the JS counterpart for TrackElement? Don't expect StationIndex to be set for pieces that don't do anything with it. It will only be set on station pieces and tower bases.

distribution/openrct2.d.ts
```diff
@@ -1359,158 +1795,369 @@ declare global {
...
         tooltip?: string;
+        /** Indicates if the widget can be interacted with. */
         isDisabled?: boolean;
     }
 
+    /**
+     * Represents a button widget, including callbacks.
+     */
     interface ButtonWidget extends Widget {
+        /** The button image. Commonly used for tabs. TODO what number for no image? */
```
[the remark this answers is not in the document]
thyeggman: It's not optional, though?  Otherwise it would be `image?: number` right? Is this an error the in API?

distribution/openrct2.d.ts
duncanspumpkin: Hey if people want to change the variable they are welcome to just don't expect anything useful to come from it.

distribution/openrct2.d.ts
duncanspumpkin: I checked the code earlier today and I'm pretty sure its set on all track element pieces. There is a `sub_6CB945` which sets it.

distribution/openrct2.d.ts
Gymnasiast: Not sure if I agree with that - making it readonly at least avoids the user wasting their time.

distribution/openrct2.d.ts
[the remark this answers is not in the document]
IntelOrca: It didn't used to be optional. Now it is, so you can make it `image?`.

thyeggman:
Going to try to finish this draft and open for proper review today.  It will probably still have a lot of TODOs but I'll continue to work on those.

thyeggman:
I made some progress today but it's taking a longer time to add all the enums and flags than I thought.

src/openrct2-ui/scripting/CustomWindow.cpp
```diff
@@ -49,6 +49,24 @@ namespace OpenRCT2::Ui::Windows
         { WWT_RESIZE, 1, 0, 0, 14, 0, 0xFFFFFFFF, STR_NONE },                // content panel
     };
 
+// TODO make a function that maps this.
+    static constexpr std::array<std::string_view, 99999> WindowClassifications = {
```
tupaschoal, line 53: Why creating a 100k sized array if you know exactly how many elements it contains?

If you don't want to count, use an `std::vector` instead. In any case, I don't see how this relates to the changeset.

thyeggman: I know, I just couldn't be bothered to count them at the time, WIP

IntelOrca:
I think your enums need to be defined as `const enum` otherwise using them will use insert their symbol rather than just the value it represents.

Gymnasiast:
@thyeggman I'd like to get this into a state that it can be merged. Do you have any plans to work on it soon(ish)?

IntelOrca:
Can we split this PR between adding just doc comments and actual new plugin APIs.

thyeggman:
Sure, I can try to finish it up this week.  Might take a bit since it's an old PR and I'll have to get it up to date, not sure how much things have changed.  Glad to see there's still interest :)

IntelOrca:
@thyeggman any update on this?

PR #11895, opened 2020-06-07
Create Vehicle::TrackType()

hdpoliveira:


src/openrct2/ride/Vehicle.h
```diff
@@ -321,6 +321,11 @@ struct Vehicle : SpriteBase
     Vehicle* TrainHead() const;
     Vehicle* TrainTail() const;
 
+    uint16_t TrackType() const
```
duncanspumpkin, line 324: In general getters in the code base would have `Get` as a prefix. Could you change it to add the prefix.

duncanspumpkin:
Looks good. I was meaning to do this ages ago. Good work.

tupaschoal:
To whomever merges this, squash and reword to match the `Get...` change

PR #11897, opened 2020-06-07
Remove usage of regs.edi and regs.di

Gymnasiast:
regs.edi was only ever written to, not read. regs.di was written to twice, but only used on the next respective line.


janisozaur:
Perhaps something worth checking against in vanilla?

PR #11909, opened 2020-06-08
Use more constants/specific types

Gymnasiast:


src/openrct2/management/Marketing.h
```diff
@@ -13,6 +13,7 @@
 #include "../common.h"
 #include "../peep/Peep.h"
 
+#include <src/openrct2/ride/ShopItem.h>
```
Gymnasiast, line 16: use ""

src/openrct2/peep/Peep.h
```diff
@@ -21,6 +21,7 @@
 #include <algorithm>
 #include <bitset>
 #include <optional>
+#include <src/openrct2/ride/ShopItem.h>
```
Gymnasiast, line 24: use ""

src/openrct2/peep/Peep.h
```diff
@@ -632,10 +634,14 @@ struct Peep : SpriteBase
...
-    uint8_t Photo3RideRef;
-    uint8_t Photo4RideRef;
-    uint8_t CurrentRide;
+    ride_id_t Photo2RideRef;
+    ride_id_t Photo3RideRef;
+    ride_id_t Photo4RideRef;
+    union
+    {
+        ride_id_t CurrentRide;
+        ParkEntranceIndex ChosenParkEntrance;
```
duncanspumpkin, line 643: I don't see any reason to make this into a union. Just have two separate fields. Or are you trying to simplify the import export code?

duncanspumpkin:
I checked the replays and they fail in trying to read the sprite structure which isn't much of a surprise. We would just need to rerecord them

src/openrct2/peep/Peep.h
Gymnasiast: The only reason I did this is because it was already like this. I think CS did it because `ChosenParkEntrance` is only used outside the park, while `CurrentRide` is only used in it.

duncanspumpkin:
I've updated the replays for this branch. Can you rebase and increment the network version.

PR #11932, opened 2020-06-11
Use ScreenCoordsXY for gfx_draw_string_right

### LINKED ISSUE #11569: Use ScreenCoordsXY on gfx_draw_string_right

tupaschoal:
Closes #11569

- First commit just stops using the old overload
- Second commit converts local variables to `ScreenCoordsXY`

### DIFF
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -952,22 +952,22 @@ static void window_editor_object_selection_invalidate(rct_window* w)
  */
 static void window_editor_object_selection_paint(rct_window* w, rct_drawpixelinfo* dpi)
 {
-    int32_t x, y, width;
+    int32_t width;
     rct_widget* widget;
     rct_string_id stringId;
 
     window_draw_widgets(w, dpi);
 
     // Draw tabs
+    ScreenCoordsXY screenPos{};
     for (size_t i = 0; i < std::size(ObjectSelectionPages); i++)
     {
         widget = &w->widgets[WIDX_TAB_1 + i];
         if (widget->type != WWT_EMPTY)
         {
             auto image = ObjectSelectionPages[i].Image;
-            x = w->windowPos.x + widget->left;
-            y = w->windowPos.y + widget->top;
-            gfx_draw_sprite(dpi, image, x, y, 0);
+            screenPos = w->windowPos + ScreenCoordsXY{ widget->left, widget->top };
+            gfx_draw_sprite(dpi, image, screenPos.x, screenPos.y, 0);
         }
     }
 
@@ -994,9 +994,8 @@ static void window_editor_object_selection_paint(rct_window* w, rct_drawpixelinf
             }
             spriteIndex += (i == 4 ? ThrillRidesTabAnimationSequence[frame] : frame);
 
-            x = w->windowPos.x + widget->left;
-            y = w->windowPos.y + widget->top;
-            gfx_draw_sprite(dpi, spriteIndex | (w->colours[1] << 19), x, y, 0);
+            screenPos = w->windowPos + ScreenCoordsXY{ widget->left, widget->top };
+            gfx_draw_sprite(dpi, spriteIndex | (w->colours[1] << 19), screenPos.x, screenPos.y, 0);
         }
     }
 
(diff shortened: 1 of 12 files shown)


src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -952,22 +952,22 @@ static void window_editor_object_selection_invalidate(rct_window* w)
...
 {
-    int32_t x, y, width;
+    int32_t width;
     rct_widget* widget;
     rct_string_id stringId;
 
     window_draw_widgets(w, dpi);
 
     // Draw tabs
+    ScreenCoordsXY screenPos{};
```
duncanspumpkin, line 962: I would rather we made a new variable for each of these.

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -1079,35 +1077,36 @@ static void window_editor_object_selection_paint(rct_window* w, rct_drawpixelinf
...
+        screenPos = { w->windowPos.x + w->widgets[WIDX_LIST].right + 4, screenPos.y + 15 };
+        width = w->windowPos.x + w->width - screenPos.x - 4;
 
-        gfx_draw_string_left_wrapped(dpi, gCommonFormatArgs, x, y + 5, width, STR_WINDOW_COLOUR_2_STRINGID, COLOUR_BLACK);
+        gfx_draw_string_left_wrapped(
+            dpi, gCommonFormatArgs, screenPos.x, screenPos.y + 5, width, STR_WINDOW_COLOUR_2_STRINGID, COLOUR_BLACK);
     }
 
-    y = w->windowPos.y + w->height - (12 * 4);
+    screenPos.y = w->windowPos.y + w->height - (12 * 4);
```
duncanspumpkin, line 1087: if you set screenPos.x to the value of `w->windowPos.x + w->width - 5` then that would simplify this considerably.

src/openrct2-ui/windows/Finances.cpp
```diff
@@ -1088,21 +1088,22 @@ static void window_finances_profit_graph_paint(rct_window* w, rct_drawpixelinfo*
...
     for (axisBase = MONEY(12, 00); axisBase >= MONEY(-12, 00); axisBase -= MONEY(6, 00))
     {
         money32 axisValue = axisBase << yAxisScale;
-        gfx_draw_string_right(dpi, STR_FINANCES_FINANCIAL_GRAPH_CASH_VALUE, &axisValue, COLOUR_BLACK, x + 70, y);
-        gfx_fill_rect_inset(dpi, x + 70, y + 5, graphLeft + 482, y + 5, w->colours[2], INSET_RECT_FLAG_BORDER_INSET);
-        y += 39;
+        gfx_draw_string_right(
+            dpi, STR_FINANCES_FINANCIAL_GRAPH_CASH_VALUE, &axisValue, COLOUR_BLACK, screenPos + ScreenCoordsXY{ 70, 0 });
+        gfx_fill_rect_inset(
+            dpi, screenPos.x + 70, screenPos.y + 5, graphLeft + 482, screenPos.y + 5, w->colours[2],
```
duncanspumpkin, line 1099: This doesn't look very clean at all. we need to get all of the values into a similar format.

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1046,34 +1046,34 @@ static void window_park_rating_paint(rct_window* w, rct_drawpixelinfo* dpi)
     window_draw_widgets(w, dpi);
     window_park_draw_tab_images(dpi, w);
 
-    int32_t x = w->windowPos.x;
-    int32_t y = w->windowPos.y;
+    auto screenPos = ScreenCoordsXY{ w->windowPos.x, w->windowPos.y };
```
duncanspumpkin, line 1049: windowPos is already a ScreenCoordsXY no need to access the x and y

duncanspumpkin:
I'd like to see the multiple redefinitions of the value of screenCoords be removed and replaced with new variables. we should be trying to split it up so that we can take functions out of these big chunks and that will help with it.

src/openrct2-ui/windows/Finances.cpp
tupaschoal: We kind of will, but I was expecting to do it when we convert these methods to receive a `ScreenCoordsXY` or a `SquareCoordsXY`

duncanspumpkin:
I'd still like to see less reuses of the variable in the other functions but we can refactor it later.

PR #11933, opened 2020-06-11
Patch unused variables

AaronVanGeffen:
Fix compilation failure due to unused vars. (Compilation only fails due to warnings-turned-errors.)

As a bonus, remove references to Twitch source files from Xcode project.

### DIFF
--- a/src/openrct2-ui/windows/Park.cpp
+++ b/src/openrct2-ui/windows/Park.cpp
@@ -33,7 +33,6 @@
 
 static constexpr const rct_string_id WINDOW_TITLE = STR_STRINGID;
 static constexpr const int32_t WH = 224;
-static constexpr const int32_t WW = 230;
 
 // clang-format off
 enum WINDOW_PARK_PAGE {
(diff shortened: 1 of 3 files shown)


src/openrct2-ui/windows/Park.cpp
```diff
@@ -33,7 +33,6 @@
 
 static constexpr const rct_string_id WINDOW_TITLE = STR_STRINGID;
 static constexpr const int32_t WH = 224;
-static constexpr const int32_t WW = 230;
```
AaronVanGeffen, line 36: NB: WW is no longer used; the MAIN_PARK_WIDGETS macro has recently been updated to take it as a parameter.

src/openrct2/core/FileWatcher.h
```diff
@@ -59,7 +59,9 @@ class FileWatcher
     ~FileWatcher();
 
 private:
+#if defined(_WIN32) || defined(__linux__)
```
Broxzier, line 62: Is the check for `defined(_WIN32)` really necessary here? `_finished` only seems to be used inside `__linux__` preprocessor blocks in the rouce file.

PR #11937, opened 2020-06-11
Remove regs use from minigolf vehicle functions

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -9312,94 +9309,81 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
-    regs.eax /= regs.ebx;
-    regs.ecx = (regs.eax * 21) >> 9;
-    regs.eax = velocity >> 12;
-    regs.ecx -= regs.eax;
-    regs.ebx = velocity;
-    regs.edx = velocity >> 8;
-    regs.edx *= regs.edx;
-    if (regs.ebx < 0)
+    int32_t newAcceleration = ((sumAcceleration / numVehicles) * 21) >> 9;
+    newAcceleration -= velocity >> 12;
```
Gymnasiast, line 9271: This does not appear to be correct. In the old situation, newacceleration was taken from (velocity >> 12) and you swapped that around here.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8859,21 +8856,19 @@ void Vehicle::UpdateTrackMotionMiniGolfVehicle(Ride* curRide, rct_ride_entry* ri
...
+            TrackSubposition--;
         }
-        TrackSubposition = regs.al;
     }
 
     ClearUpdateFlag(VEHICLE_UPDATE_FLAG_ON_LIFT_HILL);
     track_type = (tileElement->AsTrack()->GetTrackType() << 2) | (direction & 3);
     var_CF = tileElement->AsTrack()->GetBrakeBoosterSpeed();
-    regs.ax = 0;
+    track_progress = 0;
```
Gymnasiast, line 8869: Are you *entirely* sure that the contents of regs.ax are *always* written to `track_progress`?

src/openrct2/ride/Vehicle.cpp
hdpoliveira: newAcceleration is regs.ecx, and I removed regs.eax

It does seem confusing. Would you like me to have an auxiliary variable for regs.eax?

src/openrct2/ride/Vehicle.cpp
hdpoliveira: Yes.

What happened here is that on label loc_6DC743 we had track_progress = regs.ax
What I did was find all places that went to that label (only two places) and moved the assignment there. Then I cleaned up the register use.

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: I thought it looked fine. Isn't it also the same as the non golf version of this function

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: Same changes were made on the non golf version. Looks fine to me

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9043,9 +9022,9 @@ void Vehicle::UpdateTrackMotionMiniGolfVehicle(Ride* curRide, rct_ride_entry* ri
     goto loc_6DCD2B;
 
 loc_6DCA9A:
-    regs.ax = track_progress - 1;
-    if (static_cast<uint16_t>(regs.ax) != 0xFFFF)
+    if (track_progress)
```
Gymnasiast, line 9025:
```suggestion
    if (track_progress != 0)
```


ONLY use `if (var)` when handling a boolean.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8969,22 +8964,7 @@ void Vehicle::UpdateTrackMotionMiniGolfVehicle(
     trackPos = { TrackLocation.x + moveInfo->x, TrackLocation.y + moveInfo->y,
                  TrackLocation.z + moveInfo->z + RideData5[curRide->type].z_offset };
 
-    // Investigate redundant code
```
Gymnasiast, line 8972: History note: it is proper code in the regular vehicle update track motion function - it was probably just a copy-paste. (Possibly even by CS himself.)

PR #11941, opened 2020-06-13
Vehicle Motion Update

duncanspumpkin:
Remove some common code and small cleanup
Continuing on from @hdpoliveira

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -9205,6 +9205,26 @@ loc_6DCE02:
     }
 }
 
+static constexpr int32_t GetAccelerationDecrease2(const int32_t velocity, const int32_t totalMass)
+{
+    int32_t accelerationDecrease2 = velocity >> 8;
+    accelerationDecrease2 *= accelerationDecrease2;
+    if (velocity < 0)
+    {
+        accelerationDecrease2 = -accelerationDecrease2;
+    }
+    accelerationDecrease2 >>= 4;
+    // OpenRCT2: vehicles from different track types can have  0 mass.
+    if (totalMass != 0)
+    {
+        return accelerationDecrease2 / totalMass;
+    }
+    else
+    {
+        return accelerationDecrease2;
+    }
+}
+
 int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
 {
     auto curRide = get_ride(ride);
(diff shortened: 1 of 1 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -9296,35 +9309,32 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
+                    {
+                        spin_speed = 0;
+                    }
                 }
             }
         }
         else
         {
-        loc_6DD054:
             if (abs(velocity) > 0x10000)
```
Gymnasiast, line 9338: Previously, this code also ran when VEHICLE_ENTRY_FLAG_WATER_RIDE was set and vehicle_sprite_type was 00. So the `else` would have to be changed to `if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_WATER_RIDE) || vehicle_sprite_type == 0)`.

duncanspumpkin: I think I'll just copy the code into the above else block instead.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9296,35 +9309,39 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
-                    spin_speed = 0;
+                    if (vehicle_sprite_type == 2)
+                    {
+                        spin_speed = 0;
+                    }
+                }
+            }
+            else
+            {
+                if (abs(velocity) > 0x10000)
```
Gymnasiast, line 9337: Can be condensed into `else if...`

duncanspumpkin: I've modified it again to make it almost identical to the non golf version getting it ready to combine the two.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9296,39 +9309,36 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
+                if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SPINNING)
                 {
-                    spin_speed = 0;
+                    if (vehicle_sprite_type == 2)
+                    {
+                        spin_speed = 0;
+                    }
                 }
+                newAcceleration += poweredAcceleration;
+                goto loc_6DD069;
```
Gymnasiast, line 9335: I get why you did this, but I'd prefer if you didn't introduce a goto.

duncanspumpkin: Neither would I but it will be easier to refactor the two together now.

Gymnasiast: Do you plan to do that in this PR or in another?

duncanspumpkin: In another refactor

PR #11951, opened 2020-06-14
Close #11614: Require plugins to specify their license

### LINKED ISSUE #11614: plugins should have a required license field
Right now the plugins are all wild west with almost none of them specifying the license used. This should be a required field of manifest to clear up any confusion lack of it raises.

tupaschoal:
I might have been greedy about the number of licenses, but I think that [ChooseALicense](https://choosealicense.com/) is the official GitHub guide, so why not support all of the ones listed there?

I didn't save the license when parsing it, because we actually don't need it, we just want to make sure that people specify it.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -108,6 +108,7 @@ declare global {
         version: string;
         authors: string | string[];
         type: PluginType;
+        licence: string;
         minApiVersion?: number;
         main: () => void;
     }
--- a/distribution/scripting.md
+++ b/distribution/scripting.md
@@ -18,7 +18,7 @@ Local scripts can **not** alter the game state. This allows each player to enabl
 
 Remote scripts on the other hand can alter the game state in certain contexts, thus must be enabled for every player in a multiplayer game. Players **cannot** enable or disable remote scripts for multiplayer servers they join. Instead the server will upload any remote scripts that have been enabled on the server to each player. This allows servers to enable scripts without players needing to manually download or enable the same script on their end.
 
-The authors should also define a licence for the plugin, making it clear to the community whether that plugin can be altered, copied, etc. A good reference material is listed on [ChooseALlicense](https://choosealicense.com/appendix/), try to pick one of them and use its corresponding identifier, as listed on [SPDX](https://spdx.org/licenses/).
+The authors must also define a licence for the plug-in, making it clear to the community whether that plug-in can be altered, copied, etc. A good reference material is listed on [ChooseALlicense](https://choosealicense.com/appendix/), try to pick one of them and use its corresponding identifier, as listed on [SPDX](https://spdx.org/licenses/).
 
 ## Writing Scripts
 
--- a/distribution/scripting.md
+++ b/distribution/scripting.md
@@ -18,6 +18,8 @@ Local scripts can **not** alter the game state. This allows each player to enabl
 
 Remote scripts on the other hand can alter the game state in certain contexts, thus must be enabled for every player in a multiplayer game. Players **cannot** enable or disable remote scripts for multiplayer servers they join. Instead the server will upload any remote scripts that have been enabled on the server to each player. This allows servers to enable scripts without players needing to manually download or enable the same script on their end.
 
+The authors should also define a licence for the plugin, making it clear to the community whether that plugin can be altered, copied, etc. A good reference material is listed on [ChooseALlicense](https://choosealicense.com/appendix/), try to pick one of them and use its corresponding identifier, as listed on [SPDX](https://spdx.org/licenses/).
+
 ## Writing Scripts
 
 Scripts are written in ECMAScript 5 compatible JavaScript. OpenRCT2 currently uses the [duktape](https://duktape.org) library to execute scripts. This however does not mean you need to write your plug-in in JavaScript, there are many transpilers that allow you to write in a language of your choice and then compile it to JavaScript allowing it to be executed by OpenRCT2. JavaScript or [TypeScript](https://www.typescriptlang.org) is recommended however, as that will allow you to utilise the type definition file we supply (`openrct2.d.ts`). If you would like to use ECMAScript 6 or later which contain features such as the `let` keyword or classes, then you will need to use a transpiler such as [Babel](https://babeljs.io) or [TypeScript](https://www.typescriptlang.org).
@@ -38,6 +40,7 @@ registerPlugin({
     version: '1.0',
     authors: ['Your Name'],
     type: 'remote',
+    licence: 'MIT',
     main: main
 });
 ```
--- a/src/openrct2/scripting/Plugin.cpp
+++ b/src/openrct2/scripting/Plugin.cpp
@@ -11,6 +11,7 @@
 
 #    include "Plugin.h"
 
+#    include "../Diagnostic.h"
 #    include "../OpenRCT2.h"
 #    include "Duktape.hpp"
 
@@ -127,6 +128,7 @@ PluginMetadata Plugin::GetMetadata(const DukValue& dukMetadata)
         metadata.Name = TryGetString(dukMetadata["name"], "Plugin name not specified.");
         metadata.Version = TryGetString(dukMetadata["version"], "Plugin version not specified.");
         metadata.Type = ParsePluginType(TryGetString(dukMetadata["type"], "Plugin type not specified."));
+        CheckForLicence(dukMetadata["licence"], metadata.Name);
 
         auto dukMinApiVersion = dukMetadata["minApiVersion"];
         if (dukMinApiVersion.type() == DukValue::Type::NUMBER)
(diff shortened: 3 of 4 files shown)


distribution/scripting.md
```diff
@@ -18,6 +18,8 @@ Local scripts can **not** alter the game state. This allows each player to enabl
 
 Remote scripts on the other hand can alter the game state in certain contexts, thus must be enabled for every player in a multiplayer game. Players **cannot** enable or disable remote scripts for multiplayer servers they join. Instead the server will upload any remote scripts that have been enabled on the server to each player. This allows servers to enable scripts without players needing to manually download or enable the same script on their end.
 
+The authors should also define a valid license for the plugin, making it clear to the community whether that plugin can be altered, copied, etc. OpenRCT2 supports all licenses listed on [choosealicense](https://choosealicense.com/appendix/), make sure you pick one of them and use its identifier, as listed on [SPDX](https://spdx.org/licenses/).
```
Gymnasiast, line 21: OpenRCT2 does not "support" any licence.

If we are to _suggest_ a licence to people who don't know yet, make it MIT, I say - it's very common and allows others do as they please, including linking to GPL-incompatible software.

distribution/openrct2.d.ts
```diff
@@ -108,6 +108,7 @@ declare global {
         version: string;
         authors: string | string[];
         type: PluginType;
+        license: string;
```
Gymnasiast, line 111: `licence`. We use British English.

src/openrct2/scripting/Plugin.cpp
```diff
@@ -161,4 +162,57 @@ PluginType Plugin::ParsePluginType(const std::string_view& type)
     throw std::invalid_argument("Unknown plugin type.");
 }
 
+static std::vector<std::string> _validLicenses{ "0BSD",
```
Gymnasiast, line 165: Like I said earlier, I don't think we should force any licence upon the plug-in creator. The reason to create this field is to force plug-ins to make a decision in the first place.

I say change this to only check if the field is set.

distribution/scripting.md
```diff
@@ -38,6 +40,7 @@ registerPlugin({
     version: '1.0',
     authors: ['Your Name'],
     type: 'remote',
+    license: 'Unlicense',
```
AaronVanGeffen, line 43:
```suggestion
    license: 'Unlicensed',
```
?

src/openrct2/scripting/Plugin.cpp
AaronVanGeffen: I agree. We shouldn't limit this field to particular licences only. If someone wants to use a novelty licence (the "buy me a beer licence" comes to mind), they should be able to do so.

tupaschoal: Ok, I'll drop it later to let more people comment on it first.

distribution/scripting.md
tupaschoal: I was really going for the [Unlicense](https://choosealicense.com/licenses/unlicense/), but I changed the example to `MIT` as per Michael's suggestion

distribution/scripting.md
tupaschoal: I changed it to `MIT` in he example and rewrote this section slightly.

distribution/scripting.md
AaronVanGeffen: Ah, I wasn't familiar with that one. I think we shouldn't specify any licence by default, though. Instead, force the plugin author to specify one — and have the TypeScript transpiler throw an error if a non-empty string is provided, or some such.

Gymnasiast: No. This refers to "The Unlicense".

Gymnasiast: Personally, I think providing a little bit of guidance would be OK. Perhaps we should:
- throw an error if the field is missing or empty
- have it empty by default, so that everyone *has* to pick one
- suggest MIT in the .md file if the user is unsure

What do you think?

IntelOrca:
This will stop all current plugins from working. You probably want some feedback on the writers, although I think forcing the requirement of the field was disliked by some of them.

janisozaur:
It's trivial to fix using license the plugins should already have and something that was missing from original PR to enable plugins.

I don't see how liking this change or not affects anything. That said, do you have any specific quotes on that?

And to chime in on previous discussion: I also agree the license should be a free form.

Note: I never remember which spelling is British, 'c', or 's' one and sadly am probably inconsistent about it as well, please don't hold it against me.

IntelOrca:
It depends on the usage. Licence for the noun, License for the verb.

AaronVanGeffen:
> This will stop all current plugins from working. You probably want some feedback on the writers, although I think forcing the requirement of the field was disliked by some of them.

I suggested we do not require it in OpenRCT2 itself, but just require it via the TypeScript definitions. This will prevent existing plugins from breaking, while authors are nudged to fill out the field next time they edit their plugins.

Gymnasiast:
Or perhaps we should have a grace period - introduce the field now and say that people *have* to specify it, but only error out after the 0.3.0 release.

janisozaur:
That would mean people are free to create unlicensed plugins for 0.3.0 release, a case I specifically want to avoid.

IntelOrca:
They are still free to create unlicensed plugins with this change.

janisozaur:
As long as this is explicit

Gymnasiast:
Yes, but they won't be able to do so by accident, assuming we check if the field is present and not empty.

tupaschoal:
> They are still free to create unlicensed plugins with this change.

With the current diff, no right? Because it currently matches one of the valid licences. Most don't agree this should be the case, though, so I can change it to check for `licence` existing and being not empty, like Michael said.

IntelOrca:
@tupaschoal you're gonna have to since it is fine for people to use their own custom licence.

oli414:
Could this PR perhaps be split in two? 
First to merge the specifications, and later to add the actual check once the major plugins have transitioned?

Additionally, I highly recommend adding a message instead of simply not loading plugins. Something along the lines of:
"One or more of the installed plugins do not have a license specified. Please check to see if there's an update available if you wish to continue using this/these plugins."
Possibly even with a list of the plugins in question.

Just to make sure that the user experience is as smooth as possible

tupaschoal:
I've dropped requiring one of the licences in the list, it now only has to be present and not empty.

tupaschoal:
@oli414 I think we won't release `0.3.0` without requiring the field, so we can theoretically split it up, but it won't be that big of a grace period.

That being said, you can already add it to your plugins, even though it's not merged yet.

distribution/scripting.md
```diff
@@ -38,6 +40,7 @@ registerPlugin({
     version: '1.0',
     authors: ['Your Name'],
     type: 'remote',
+    license: 'MIT',
```
AaronVanGeffen, line 43: The code is using `licence`, but the example uses `license`. We should probably support both, but:
```suggestion
    licence: 'MIT',
```

src/openrct2/scripting/Plugin.cpp
```diff
@@ -161,4 +162,10 @@ PluginType Plugin::ParsePluginType(const std::string_view& type)
     throw std::invalid_argument("Unknown plugin type.");
 }
 
+void Plugin::CheckForLicence(const std::string_view& licence)
+{
+    if (licence.empty())
+        throw std::invalid_argument("Licence cannot be empty.");
```
AaronVanGeffen, line 168: Frankly, I'm torn on the client enforcing this. I'm fine with reminding devs by means of the definition in `openrct2.d.ts`, but this is of no help to the users.

Gymnasiast:
Let's take the Scooby Doo approach and split this PR in two - one PR that adds the licence fields and tells the developer they *must* include it (which is the uncontroversial part) and one that enforces this (the controversial) part. At least the requirement to specify a licence is then out in the open.

src/openrct2/scripting/Plugin.cpp
tupaschoal: I've changed it to just log an error for now

tupaschoal:
I'm now just `logging` an error. We can later enforce it. Does that sound ok?

distribution/scripting.md
```diff
@@ -18,6 +18,8 @@ Local scripts can **not** alter the game state. This allows each player to enabl
 
 Remote scripts on the other hand can alter the game state in certain contexts, thus must be enabled for every player in a multiplayer game. Players **cannot** enable or disable remote scripts for multiplayer servers they join. Instead the server will upload any remote scripts that have been enabled on the server to each player. This allows servers to enable scripts without players needing to manually download or enable the same script on their end.
 
+The authors should also define a licence for the plugin, making it clear to the community whether that plugin can be altered, copied, etc. A good reference material is listed on [ChooseALlicense](https://choosealicense.com/appendix/), try to pick one of them and use its corresponding identifier, as listed on [SPDX](https://spdx.org/licenses/).
```
Gymnasiast, line 21:
```suggestion
The authors must also define a licence for the plug-in, making it clear to the community whether that plug-in can be altered, copied, etc. A good reference material is listed on [ChooseALlicense](https://choosealicense.com/appendix/), try to pick one of them and use its corresponding identifier, as listed on [SPDX](https://spdx.org/licenses/).
```

tupaschoal:
(Squash on merge due to the suggestion I accepted)

PR #11954, opened 2020-06-14
Pass only TrackLocation instead of whole vehicle

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7989,7 +7988,7 @@ bool Vehicle::UpdateTrackMotionForwardsGetNewTrack(uint16_t trackType, Ride* cur
                 }
             }
             map_invalidate_element(TrackLocation, tileElement);
-            vehicle_update_block_brakes_open_previous_section(this, tileElement);
+            block_brakes_open_previous_section(*curRide, TrackLocation, tileElement);
```
tupaschoal, line 7991: Are we certain that `curRide` is valid in here?

Gymnasiast: This is also not exactly how it worked previously. It used `vehicle->ride`, not `curRide`.

hdpoliveira: > Are we certain that `curRide` is valid in here?

@tupaschoal Yes, it is verified some lines above.

> This is also not exactly how it worked previously. It used `vehicle->ride`, not `curRide`.

@Gymnasiast Yes, the actual Ride object was being recalculated within the function without need.

tupaschoal: Ok then. @Gymnasiast you cool with this?

Gymnasiast: I want to check it myself.

Gymnasiast: OK, I traced it back. `curRide` is snowballed from `Vehicle::UpdateTrackMotion`, which exits if it is null. It is indeed set from `vehicle->ride`.

PR #11955, opened 2020-06-14
Remove gotos from TrackMotionForwardsGetNewTrack

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7996,79 +7996,80 @@ bool Vehicle::UpdateTrackMotionForwardsGetNewTrack(uint16_t trackType, Ride* cur
     // Change from original: this used to check if the vehicle allowed doors.
     vehicle_update_scenery_door(this);
 
+    bool isGoingBack = false;
     switch (TrackSubposition)
     {
         default:
-            goto loc_6DB358;
+            break;
```
Gymnasiast, line 8003: Could you move the default case to the bottom?

PR #11956, opened 2020-06-15
Simplify code related to V-angle and bank

hdpoliveira:


duncanspumpkin:
I'm wanting to rename these fields roll and pitch. I need to check if the pitch values are the same as the ones for pitch that are within `vehicle->vehicle_sprite_type`

duncanspumpkin:
I've confirmed that the two types are the same. `TRACK_SLOPE_..` == `vehicle_sprite_type`. I would like to rename both so that they become `pitch`. `TRACK_BANK_...` == `bank_rotation` which should become `roll`. There is a bit of work to be done to make that renaming happen though.

src/openrct2/ride/Track.h
```diff
@@ -18,21 +18,37 @@ constexpr const uint16_t RideConstructionSpecialPieceSelected = 0x100;
...
 };
 assert_struct_size(rct_trackdefinition, 8);
 #pragma pack(pop)
 
+struct VAngleAndBank
+{
+    vangle_type_t VAngle;
+    bank_type_t Bank;
+};
+bool operator==(const VAngleAndBank& vb1, const VAngleAndBank& vb2)
```
Gymnasiast, line 43: Compiling fails because these are not declared first. I guess they should be moved up into the struct?

hdpoliveira: It is actually missing a constexpr in the two comparison functions. I was sure I compiled those before opening this PR, but I guess I messed something up...

hdpoliveira: Just noticed that I had this done on my working area, but I did not commit :facepalm:

hdpoliveira:
> I've confirmed that the two types are the same. `TRACK_SLOPE_..` == `vehicle_sprite_type`. I would like to rename both so that they become `pitch`. `TRACK_BANK_...` == `bank_rotation` which should become `roll`. There is a bit of work to be done to make that renaming happen though.

@duncanspumpkin let me see if I got this right:
- rename every reference to vangle in this PR to pitch
- rename every reference to bank in this PR to roll
- rename Vehicle::vehicle_sprite_type to Vehicle::Pitch
- rename Vehicle::bank_rotation to Vehicle::Roll
- rename all TRACK_SLOPE_... to TRACK_PITCH_...
- rename all TRACK_BANCK_... to TRACK_ROLL_...

Is this what you are asking? If yes, I prefer we do the last 4 in another PR, I can do the first two in this one.

duncanspumpkin:
Yes i think that sounds correct. Keep this just for the track. The renames on vehicles need a little rework as there are two fields that are slightly different so i want to introduce a union.

hdpoliveira:
Actually I just started doing this renaming and it looks a little bit weird. The code ends up in an inconsistent state, in which half of the things are Pitch and the other half are v-angle.

Let me push my change here then you can decide if it should be in this PR or a different one.

src/openrct2/ride/Track.cpp
```diff
@@ -560,6 +560,16 @@ const rct_trackdefinition FlatRideTrackDefinitions[256] =
 };
 // clang-format on
 
+PitchAndRoll track_pitch_and_roll_start(uint8_t trackType)
```
duncanspumpkin, line 563: All new functions should be in CamelCase

hdpoliveira:
@tupaschoal I think we can remove the Work in progress label 😃

tupaschoal:
Done. I'll leave for Duncan to review :)

tupaschoal:
Is this ok now @duncanspumpkin ?

PR #11960, opened 2020-06-16
Name track vehicle info lists

Gymnasiast:


Gymnasiast:
I think it's nice to them, so you it's easier to follow what kind of data you're dealing with. It also makes it easier to find all usages of subposition data if we ever add one.

PR #11972, opened 2020-06-17
Fix #7324: Research window shows vehicle name instead of ride name

### LINKED ISSUE #7324: Research window now shows the name of trains/cars for each ride, instead of the name of the ride itself.
**OS:** Windows 8
**Version:** e.g. 0.1.3
**Commit/Build:** ccb1b33

Research window now shows the name of trains/cars for each ride, instead of the name of the ride itself.

This happens not only for roller coasters, but even for flat rides that have vehicles, like the reverse freefall.

Shops/stalls, scenery, and rides without vehicles (like mazes or mini-golf) still appear in the research as normal.

amity airfield 2018-03-20 12-58-13

**Steps to reproduce:**
1. Open research window.
2. Research any ride that has vehicles.

Gymnasiast:
This iterates over the research list upon loading a save and determines whether a vehicle is the first to be invented for its type. If so, it sets a flag.

Six Flags Great Adventure 2020-06-17 16-50-02

### DIFF
--- a/src/openrct2-ui/windows/Research.cpp
+++ b/src/openrct2-ui/windows/Research.cpp
@@ -16,6 +16,7 @@
 #include <openrct2/management/Finance.h>
 #include <openrct2/management/NewsItem.h>
 #include <openrct2/management/Research.h>
+#include <openrct2/ride/RideData.h>
 #include <openrct2/sprites.h>
 #include <openrct2/world/Park.h>
 #include <openrct2/world/Scenery.h>
@@ -362,16 +363,33 @@ void window_research_development_page_paint(rct_window* w, rct_drawpixelinfo* dp
     else
     {
         // Research type
-        stringId = STR_RESEARCH_UNKNOWN;
+        rct_string_id strings[2] = { STR_RESEARCH_UNKNOWN, 0 };
+        rct_string_id label = STR_RESEARCH_TYPE_LABEL;
         if (gResearchProgressStage != RESEARCH_STAGE_INITIAL_RESEARCH)
         {
-            stringId = ResearchCategoryNames[gResearchNextItem->category];
+            strings[0] = ResearchCategoryNames[gResearchNextItem->category];
             if (gResearchProgressStage != RESEARCH_STAGE_DESIGNING)
             {
-                stringId = gResearchNextItem->GetName();
+                strings[0] = gResearchNextItem->GetName();
+                if (gResearchNextItem->type == RESEARCH_ENTRY_TYPE_RIDE)
+                {
+                    auto rtd = RideTypeDescriptors[gResearchNextItem->baseRideType];
+                    if (!rtd.HasFlag(RIDE_TYPE_FLAG_LIST_VEHICLES_SEPARATELY))
+                    {
+                        if (gResearchNextItem->flags & RESEARCH_ENTRY_FLAG_FIRST_OF_TYPE)
+                        {
+                            strings[0] = rtd.Naming.Name;
+                        }
+                        else
+                        {
+                            strings[1] = rtd.Naming.Name;
+                            label = STR_RESEARCH_TYPE_LABEL_VEHICLE;
+                        }
+                    }
+                }
             }
         }
-        gfx_draw_string_left_wrapped(dpi, &stringId, x, y, 296, STR_RESEARCH_TYPE_LABEL, COLOUR_BLACK);
+        gfx_draw_string_left_wrapped(dpi, &strings, x, y, 296, label, COLOUR_BLACK);
         y += 25;
 
         // Progress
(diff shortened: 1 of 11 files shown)


src/openrct2-ui/windows/Research.cpp
```diff
@@ -362,16 +363,33 @@ void window_research_development_page_paint(rct_window* w, rct_drawpixelinfo* dp
     else
     {
         // Research type
-        stringId = STR_RESEARCH_UNKNOWN;
+        rct_string_id strings[2] = { STR_RESEARCH_UNKNOWN, 0 };
```
tupaschoal, line 366:
```suggestion
        std::array<rct_string_id, 2> strings = { STR_RESEARCH_UNKNOWN, 0 };
```
If it works, prefer the `std::array`

src/openrct2-ui/windows/Research.cpp
```diff
@@ -402,11 +420,31 @@ void window_research_development_page_paint(rct_window* w, rct_drawpixelinfo* dp
     rct_string_id lastDevelopmentFormat;
```
tupaschoal, line 420: Move this to be inside the if, so we reduce its scope, as it's not needed anymore later. Also either initialize it to `STR_RESEARCH_SCENERY_LABEL` and change if `type != RESEARCH_ENTRY_TYPE_SCENERY` or to something invalid, so we know in case we break this someday.

src/openrct2/management/Research.cpp
```diff
@@ -910,3 +890,107 @@ bool ResearchItem::Exists() const
...
+static std::bitset<RIDE_TYPE_COUNT> _seenRideType = {};
+static std::bitset<RIDE_TYPE_COUNT* MAX_RIDE_GROUPS_PER_RIDE_TYPE> _seenRideGroup = {};
+
+static void research_update_first_of_type(ResearchItem* researchItem)
+{
+    if (researchItem->IsNull())
+        return;
+
+    if (researchItem->type != RESEARCH_ENTRY_TYPE_RIDE)
+        return;
```
tupaschoal, line 903: I'd merge these ifs with an `||`, but that's your call

src/openrct2/management/Research.cpp
```diff
@@ -910,3 +890,107 @@ bool ResearchItem::Exists() const
...
+
+    auto rideType = researchItem->baseRideType;
+    if (rideType >= RIDE_TYPE_COUNT)
```
tupaschoal, line 906: Should we log an error if this happens?

src/openrct2/management/Research.cpp
```diff
@@ -910,3 +890,107 @@ bool ResearchItem::Exists() const
...
+        }
+    }
+}
+
+void research_determine_first_of_type()
+{
+    _seenRideType.reset();
+    _seenRideGroup.reset();
+
+    for (auto& researchItem : gResearchItemsInvented)
```
tupaschoal, line 944:
```suggestion
    for (const auto& researchItem : gResearchItemsInvented)
```

src/openrct2/management/Research.cpp
```diff
@@ -910,3 +890,107 @@ bool ResearchItem::Exists() const
...
+        auto tmpVal = gResearchLastItem.value();
+        research_update_first_of_type(&tmpVal);
+        gResearchLastItem = tmpVal;
+    }
+    if (gResearchNextItem.has_value())
+    {
+        auto tmpVal = gResearchNextItem.value();
+        research_update_first_of_type(&tmpVal);
+        gResearchNextItem = tmpVal;
+    }
```
tupaschoal, line 990: Seems like you should extract a function and call `foo(gResearchLastItem)` and `foo(gResearchNextItem)`

src/openrct2/management/Research.cpp
```diff
@@ -910,3 +890,107 @@ bool ResearchItem::Exists() const
...
+
+    for (auto& researchItem : gResearchItemsUninvented)
```
tupaschoal, line 992:
```suggestion
    for (const auto& researchItem : gResearchItemsUninvented)
```

src/openrct2/management/Research.cpp
```diff
@@ -910,3 +890,107 @@ bool ResearchItem::Exists() const
     }
     return false;
 }
+
+static std::bitset<RIDE_TYPE_COUNT> _seenRideType = {};
+static std::bitset<RIDE_TYPE_COUNT* MAX_RIDE_GROUPS_PER_RIDE_TYPE> _seenRideGroup = {};
+
+static void research_update_first_of_type(ResearchItem* researchItem)
```
tupaschoal, line 897: Is there a thing such as a `Research` class? Just trying to understand the reason for the free function in here,

Gymnasiast: There is no Research class, no. (Most files are still a collection of C-style functions.)

src/openrct2/management/Research.cpp
Gymnasiast: No. It has to update the `researchItem`s.

src/openrct2-ui/windows/Research.cpp
```diff
@@ -362,16 +363,33 @@ void window_research_development_page_paint(rct_window* w, rct_drawpixelinfo* dp
     else
     {
         // Research type
-        stringId = STR_RESEARCH_UNKNOWN;
+        std::array<rct_string_id, 2> strings = { STR_RESEARCH_UNKNOWN, 0 };
```
duncanspumpkin, line 366: Shouldn't we be using the Formatter here instead of a std::array?

Gymnasiast: `strings` acts as a buffer here, not as a formatter.

The formatter would still have to format to _a_ buffer - either this one or gCommonFormatArgs. But I really don't want to introduce any more instances of the latter.

PR #11980, opened 2020-06-17
SpriteList enum class

duncanspumpkin:
Converted the SPRITE_LIST enum to an enum class. I'm not 100% convinced on it as there is now a ton of static_casts. I've also deleted the spatial index loop detection code as it was performing some bad array access with respect to the sprite lists. Also the spatial index data is recalculated on load so it should never end up in a situation where it is in a loop on load.

### DIFF
--- a/src/openrct2-ui/windows/EditorBottomToolbar.cpp
+++ b/src/openrct2-ui/windows/EditorBottomToolbar.cpp
@@ -310,8 +310,7 @@ static void window_editor_bottom_toolbar_mouseup([[maybe_unused]] rct_window* w,
     if (widgetIndex == WIDX_PREVIOUS_STEP_BUTTON)
     {
         if ((gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER)
-            || (gSpriteListCount[static_cast<uint8_t>(EntityListId::Free)] == MAX_SPRITES
-                && !(gParkFlags & PARK_FLAGS_SPRITES_INITIALISED)))
+            || (GetEntityListCount(EntityListId::Free) == MAX_SPRITES && !(gParkFlags & PARK_FLAGS_SPRITES_INITIALISED)))
         {
             previous_button_mouseup_events[gS6Info.editor_step]();
         }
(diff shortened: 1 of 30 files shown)


Gymnasiast:
Since we're renaming anyway, let's use the new name `entity` for the renamed structs/enums.

duncanspumpkin:
I've renamed it `EntityListId` so that it doesn't clash with `EntityList` which is the entity list iterator container.

src/openrct2-ui/windows/EditorBottomToolbar.cpp
```diff
@@ -310,7 +310,8 @@ static void window_editor_bottom_toolbar_mouseup([[maybe_unused]] rct_window* w,
     if (widgetIndex == WIDX_PREVIOUS_STEP_BUTTON)
     {
         if ((gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER)
-            || (gSpriteListCount[SPRITE_LIST_FREE] == MAX_SPRITES && !(gParkFlags & PARK_FLAGS_SPRITES_INITIALISED)))
+            || (gSpriteListCount[static_cast<uint8_t>(EntityListId::Free)] == MAX_SPRITES
```
Gymnasiast, line 313: This is a bit unwieldy. Is there no way around this casting requirement?

(If there isn't, I prefer saving `(gSpriteListCount[static_cast<uint8_t>(EntityListId::Free)]` into an intermediate variable.)

duncanspumpkin: There is no way to have it automatically convert into a uint8_t unfortunately but we could make an assessor function that gets the list count from a entity list id.

Gymnasiast: That would also work. I'll leave it to you which one you pick.

duncanspumpkin: I have applied the accessor function change. Eventually I'm going to remove the setters completely so I haven't created a Set function.

Issue #11964, opened 2020-06-16
Move remaining fields/arrays to RideTypeDescriptor


Gymnasiast:
- [x] RideProperties - https://github.com/OpenRCT2/OpenRCT2/pull/11967
- [x] rideBonusValue - https://github.com/OpenRCT2/OpenRCT2/pull/11967
- [x] RideRatings - https://github.com/OpenRCT2/OpenRCT2/pull/11975
- [x] RideRatingsCalculateFuncTable - https://github.com/OpenRCT2/OpenRCT2/pull/11983
- [x] RideColourPresets - https://github.com/OpenRCT2/OpenRCT2/pull/11987
- [x] RideColourKey - https://github.com/OpenRCT2/OpenRCT2/pull/11996
- [x] Covered track piece availability - https://github.com/OpenRCT2/OpenRCT2/pull/11981
- [x] RideTypeEnumNames - https://github.com/OpenRCT2/OpenRCT2/pull/11999

Additionally:
- [x] Move RideTypeTrackPaintFunctionsOld to TestPaint.

frutiemax:
I'll work on this if no one is.

Gymnasiast:
Yeah, that's fine. I have left some comments on your PR already.

Also, we still need to take a decision on what to do with `RideTypeEnumNames`, so don't do that one yet.

duncanspumpkin:
Great work!

PR #11983, opened 2020-06-17
Part of #11964 : Merge RideRatingsCalculateFuncTable in RideTypeDescriptor

frutiemax:
Merge RideRatingsCalculateFuncTable in RideTypeDescriptor


frutiemax:
I can't fix those linker errors for Linux i686 for the testpaint module, I would need some help there.

Gymnasiast:
I will look into that.

In the meantime, could you move the functions back into the RideRatings.cpp file? The functions will need a lot of revising anyway, and that is better left to another PR. Just make sure it compiles on Windows and regular Linux and I'll take care of TestPaint.

src/openrct2/ride/RideData.cpp
```diff
@@ -25,6 +25,7 @@
 #include "../localisation/Localisation.h"
 #include "../sprites.h"
 #include "Ride.h"
+#include "RideRatings.h"
```
Gymnasiast, line 28: Is this include necessary?

Gymnasiast:
Don't push anything to the branch yet until all CIs are green. I might need to add more files.

.github/workflows/ci.yml
```diff
@@ -164,7 +164,7 @@ jobs:
         shell: bash
         env:
           TESTPAINT: true
-        run: . scripts/setenv -q && build -DWITH_TESTS=on -DCMAKE_BUILD_TYPE=MinSizeRel -DBUILD_SHARED_LIBS=ON -DPORTABLE=ON -DFORCE32=ON -DCMAKE_CXX_FLAGS="-m32 -gz"
+        run: . scripts/setenv -q && build -DWITH_TESTS=on -DCMAKE_BUILD_TYPE=MinSizeRel -DBUILD_SHARED_LIBS=ON -DPORTABLE=ON -DFORCE32=ON -DENABLE_SCRIPTING=OFF -DCMAKE_CXX_FLAGS="-m32 -gz"
```
tupaschoal, line 167: Is this correct? Seems like we didn't need to disable scripting previously

Gymnasiast: No, but that was because we didn't include any file that includes references to scripting. So I would say disabling scripting for TestPaint is correct, it just happened to work by accident previously.

tupaschoal: Hmm, I see. Do we intend for it to work in the future? Because if so, we should open an issue to track this limitation

Gymnasiast: I can't think of a reason why scripting support would be useful for TestPaint. It would be like giving a bicycle to a fish.

duncanspumpkin: Also testpaint in its current form needs to die and be reborn completely different.

PR #11985, opened 2020-06-18
Increase widget heights in object selection window

AaronVanGeffen:
These changes accommodate CJK languages:
* Buttons from 11px to 13px.
* Text box from 11px to 13px.
* Like buttons, text boxes now paint their contents vertically centred.

Furthermore:
* Rename `WIDX_FILTER_STRING_BUTTON` constant to `WIDX_FILTER_TEXT_BOX`.
* Rename `WINDOW_STAFF_LIST_WIDGET_IDX` to `WINDOW_EDITOR_OBJECT_SELECTION_WIDGET_IDX`.

Before:
Unnamed park 2020-06-18 20-00-09

After:
Unnamed park 2020-06-18 20-18-16

### DIFF
--- a/src/openrct2-ui/interface/Widget.cpp
+++ b/src/openrct2-ui/interface/Widget.cpp
@@ -1063,6 +1063,13 @@ static void widget_text_box_draw(rct_drawpixelinfo* dpi, rct_window* w, rct_widg
     gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
     gCurrentFontFlags = 0;
 
+    // Figure out where the text should be positioned vertically.
+    int32_t height = (widget->bottom - widget->top);
+    if (height >= 10)
+        t = w->windowPos.y + std::max<int32_t>(widget->top, widget->top + (height / 2) - 5);
+    else
+        t = w->windowPos.y + widget->top - 1;
+
     if (!active || gTextInput == nullptr)
     {
         if (w->widgets[widgetIndex].text != 0)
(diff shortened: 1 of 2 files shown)


Gymnasiast:
Could you remove the dot from your commit messages? It's against the commit message format.

src/openrct2-ui/interface/Widget.cpp
```diff
@@ -1063,6 +1063,13 @@ static void widget_text_box_draw(rct_drawpixelinfo* dpi, rct_window* w, rct_widg
     gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
     gCurrentFontFlags = 0;
 
+    // Figure out where the text should be positioned vertically.
+    int32_t height = (widget->bottom - widget->top);
+    if (height >= 10)
+        t = w->windowPos.y + std::max<int32_t>(widget->top, widget->top + (height / 2) - 5);
```
tupaschoal, line 1069: Why the hardcoded `5`, is this half the font size?

AaronVanGeffen: It is half the font size, yes. Admittedly, I've copied these lines from the `widget_text` function:

https://github.com/OpenRCT2/OpenRCT2/blob/4974d1a9b06934955d9a3fb13fbc93d8b9501c01/src/openrct2-ui/interface/Widget.cpp#L349-L353

I added those originally, so I blame myself for not using a constant there either. However, there's a reason for this. At the time, I decided not to use `font_get_line_height`, because the line height is substantially larger than the base font height in pixels we need for this operation. I wanted to return to this at some point, but never got around to it — and it works, so it just hasn't been a priority.

PR #11993, opened 2020-06-19
Part of #9473: Scenery door functions

hdpoliveira:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7438,43 +7438,49 @@ static void play_scenery_door_close_sound(const CoordsXYZ& loc, WallElement* til
     }
 }
 
-/**
- *
- *  rct2: 0x006DEE93
- */
-static void vehicle_update_scenery_door(Vehicle* vehicle)
+static void animate_scenery_door(
+    const CoordsXYZD& doorLocation, const CoordsXYZ& trackLocation, bool isLastVehicle, bool isBackwards)
```
duncanspumpkin, line 7442: CamelCase for all new functions

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7479,8 +7479,8 @@ void Vehicle::UpdateSceneryDoor() const
     auto wallCoords = CoordsXYZ{ x, y, TrackLocation.z - trackBlock->z + trackCoordinates->z_end }.ToTileStart();
     int32_t direction = (track_direction + trackCoordinates->rotation_end) & 3;
 
-    AnimateSceneryDoor(
-        { wallCoords, static_cast<Direction>(direction) }, TrackLocation, next_vehicle_on_train == SPRITE_INDEX_NULL, false);
+    AnimateSceneryDoor<false>(
```
Gymnasiast, line 7482: What does this solve exactly? It still acts exactly like a regular parameter.

hdpoliveira: Since the parameter is always constant, there is no reason for passing it as a function parameter.

hdpoliveira: This allows the compiler to do better optimizations.
I can remove it if you want 😃

Gymnasiast: > This allows the compiler to do better optimizations.

Now, _that_ is a proper reason. Let's leave it like this, then.

duncanspumpkin: We've done similar in the software drawing functions. I don't think in this case it's really required but I don't mind it.

PR #11994, opened 2020-06-19
Get RTD from ride via a struct method, part 1

Gymnasiast:
The basic idea of this refactor is that the code will retrieve an RTD from a ride and use that for determining ride-related things. This means that ride types will behave in the same way that ride entries and other objects do already.

Access to the RideTypeDescriptors array should eventually be refactored out entirely - code should either use the Ride struct method or use a function like `get_ride_type_descriptor(ObjectEntryIndex index);` to retrieve a specific RTD.

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -1060,7 +1060,7 @@ static void window_ride_draw_tab_vehicle(rct_drawpixelinfo* dpi, rct_window* w)
         }
 
         // For any suspended rides, move image higher in the vehicle tab on the rides window
-        if (RideTypeDescriptors[ride->type].HasFlag(RIDE_TYPE_FLAG_IS_SUSPENDED))
+        if (ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_IS_SUSPENDED))
         {
             screenCoords.y /= 4;
         }
@@ -1138,8 +1138,9 @@ static void window_ride_disable_tabs(rct_window* w)
         return;
 
     uint8_t ride_type = ride->type; // ecx
+    const auto& rtd = ride->GetRideTypeDescriptor();
 
-    if (!ride_type_has_flag(ride_type, RIDE_TYPE_FLAG_HAS_DATA_LOGGING))
+    if (!rtd.HasFlag(RIDE_TYPE_FLAG_HAS_DATA_LOGGING))
         disabled_tabs |= (1 << WIDX_TAB_8); // 0x800
 
     if (ride_type == RIDE_TYPE_MINI_GOLF)
@@ -1148,19 +1149,17 @@ static void window_ride_disable_tabs(rct_window* w)
     if (ride_type_has_flag(ride_type, RIDE_TYPE_FLAG_NO_VEHICLES))
         disabled_tabs |= (1 << WIDX_TAB_2); // 0x20
 
-    if (!ride_type_has_flag(ride_type, RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_MAIN)
-        && !ride_type_has_flag(ride_type, RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_ADDITIONAL)
-        && !ride_type_has_flag(ride_type, RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_SUPPORTS)
-        && !ride_type_has_flag(ride_type, RIDE_TYPE_FLAG_HAS_VEHICLE_COLOURS)
-        && !(RideTypeDescriptors[ride->type].Flags & RIDE_TYPE_FLAG_HAS_ENTRANCE_EXIT))
+    if (!rtd.HasFlag(RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_MAIN) && !rtd.HasFlag(RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_ADDITIONAL)
+        && !rtd.HasFlag(RIDE_TYPE_FLAG_HAS_TRACK_COLOUR_SUPPORTS) && !rtd.HasFlag(RIDE_TYPE_FLAG_HAS_VEHICLE_COLOURS)
+        && !rtd.HasFlag(RIDE_TYPE_FLAG_HAS_ENTRANCE_EXIT))
     {
         disabled_tabs |= (1 << WIDX_TAB_5); // 0x100
     }
 
-    if (ride_type_has_flag(ride_type, RIDE_TYPE_FLAG_IS_SHOP))
+    if (rtd.HasFlag(RIDE_TYPE_FLAG_IS_SHOP))
         disabled_tabs |= (1 << WIDX_TAB_3 | 1 << WIDX_TAB_4 | 1 << WIDX_TAB_7); // 0x4C0
 
-    if (!(RideTypeDescriptors[ride->type].Flags & RIDE_TYPE_FLAG_ALLOW_MUSIC))
+    if (!rtd.HasFlag(RIDE_TYPE_FLAG_ALLOW_MUSIC))
     {
         disabled_tabs |= (1 << WIDX_TAB_6); // 0x200
     }
@@ -1874,8 +1873,10 @@ static void window_ride_show_view_dropdown(rct_window* w, rct_widget* widget)
     gDropdownItemsArgs[0] = STR_OVERALL_VIEW;
     int32_t currentItem = 1;
 
+    const auto& rtd = ride->GetRideTypeDescriptor();
+
     // Vehicles
-    int32_t name = GetRideComponentName(RideTypeDescriptors[ride->type].NameConvention.vehicle).number;
+    int32_t name = GetRideComponentName(rtd.NameConvention.vehicle).number;
     for (int32_t i = 1; i <= ride->num_vehicles; i++)
     {
         gDropdownItemsFormat[currentItem] = STR_DROPDOWN_MENU_LABEL;
@@ -1884,7 +1885,7 @@ static void window_ride_show_view_dropdown(rct_window* w, rct_widget* widget)
     }
 
     // Stations
-    name = GetRideComponentName(RideTypeDescriptors[ride->type].NameConvention.station).number;
+    name = GetRideComponentName(rtd.NameConvention.station).number;
     for (int32_t i = 1; i <= ride->num_stations; i++)
     {
         gDropdownItemsFormat[currentItem] = STR_DROPDOWN_MENU_LABEL;
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -1884,8 +1884,8 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
                 auto ride = get_ride(rideId);
                 if (ride != nullptr)
                 {
-                    auto rideType = RideTypeDescriptors[ride->type].Naming.Name;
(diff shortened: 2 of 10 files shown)


src/openrct2/world/Footpath.cpp
```diff
@@ -800,9 +800,15 @@ static void loc_6A6D7E(
                     if (initialTileElementPos.z == tileElement->GetBaseZ())
                     {
                         auto ride = get_ride(tileElement->AsTrack()->GetRideIndex());
-                        if (ride == nullptr || !ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_FLAT_RIDE))
+                        if (ride == nullptr)
                         {
-                            continue;
+                            return;
```
Gymnasiast, line 805: revert

src/openrct2/world/Footpath.cpp
```diff
@@ -800,9 +800,15 @@ static void loc_6A6D7E(
...
+                        }
+
+                        const auto& rtd = ride->GetRideTypeDescriptor();
+                        if (!rtd->HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
+                        {
+                            return;
```
Gymnasiast, line 811: continue;

test/testpaint/Compat.cpp
```diff
@@ -844,3 +844,8 @@ void ride_ratings_calculate_lim_launched_roller_coaster([[maybe_unused]] Ride* r
 void ride_ratings_calculate_drink_stall([[maybe_unused]] Ride* ride)
 {
 }
+
+const RideTypeDescriptor& Ride::GetRideTypeDescriptor() const
```
Gymnasiast, line 848:
```suggestion
const RideTypeDescriptor* Ride::GetRideTypeDescriptor() const
```

src/openrct2/ride/Ride.cpp
```diff
@@ -7761,10 +7768,10 @@ uint64_t Ride::GetAvailableModes() const
...
 
-    return GetRideTypeDescriptor().RideModes;
+    return GetRideTypeDescriptor()->RideModes;
 }
 
-const RideTypeDescriptor& Ride::GetRideTypeDescriptor() const
+const RideTypeDescriptor* Ride::GetRideTypeDescriptor() const
 {
-    return RideTypeDescriptors[type];
+    return &RideTypeDescriptors[type];
```
tupaschoal, line 7776: I find this change a bit weird. How can this be a `nullptr`? Do you not initialize the RTD array entirely?

Besides that, you're only checking for the return to be valid on a couple of places.

src/openrct2/ride/Ride.cpp
```diff
@@ -1409,6 +1412,8 @@ void ride_construction_set_default_next_piece()
     if (ride == nullptr)
         return;
 
+    const auto& rtd = ride->GetRideTypeDescriptor();
```
tupaschoal, line 1415: No need for `&` if it's a pointer (applies to multiple other places).

You might not need to change this based on my other comment's outcome.

src/openrct2/ride/Ride.cpp
duncanspumpkin: Agreed this should return a `const RideTypeDescriptor& ` as there will be no situation where this is nullptr.

Gymnasiast: There *will*, once ride types are turned into objects.

tupaschoal: Hmm, then what about returning an optional?

Gymnasiast: I went for consistency with the other object entry functions (e.g. ride entry, footpath entry, wall entry, ...).

Gymnasiast:
I have updated the PR to use a reference instead of a pointer, at least for now.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -1048,6 +1048,7 @@ static void window_ride_draw_tab_vehicle(rct_drawpixelinfo* dpi, rct_window* w)
         if (rideEntry == nullptr)
             return;
 
+        const auto& rtd = ride->GetRideTypeDescriptor();
```
tupaschoal, line 1051: Move this closer to the if below (or do it directly on the if to reduce the scope of the variable)

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -1060,7 +1061,7 @@ static void window_ride_draw_tab_vehicle(rct_drawpixelinfo* dpi, rct_window* w)
         }
 
         // For any suspended rides, move image higher in the vehicle tab on the rides window
-        if (RideTypeDescriptors[ride->type].HasFlag(RIDE_TYPE_FLAG_IS_SUSPENDED))
+        if (rtd.HasFlag(RIDE_TYPE_FLAG_IS_SUSPENDED))
```
tupaschoal, line 1064:
```
        if (const auto& rtd = ride->GetRideTypeDescriptor(); rtd.HasFlag(RIDE_TYPE_FLAG_IS_SUSPENDED))
```
or 
```
        if (ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_IS_SUSPENDED))
```

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1884,7 +1884,8 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
                 auto ride = get_ride(rideId);
                 if (ride != nullptr)
                 {
-                    auto rideType = RideTypeDescriptors[ride->type].Naming.Name;
+                    const auto& rtd = ride->GetRideTypeDescriptor();
+                    auto rideType = rtd.Naming.Name;
```
tupaschoal, line 1888:
```suggestion
                    auto rideType = ride->GetRideTypeDescriptor().Naming.Name;
```

src/openrct2/actions/LandSetHeightAction.hpp
```diff
@@ -298,7 +298,8 @@ DEFINE_GAME_ACTION(LandSetHeightAction, GAME_COMMAND_SET_LAND_HEIGHT, GameAction
                         int32_t maxHeight = rideEntry->max_height;
                         if (maxHeight == 0)
                         {
-                            maxHeight = RideTypeDescriptors[ride->type].Heights.MaxHeight;
+                            const auto& rtd = ride->GetRideTypeDescriptor();
+                            maxHeight = rtd.Heights.MaxHeight;
```
tupaschoal, line 302:
```suggestion
                            maxHeight = ride->GetRideTypeDescriptor().Heights.MaxHeight;
```

src/openrct2/world/Footpath.cpp
```diff
@@ -875,7 +875,13 @@ static void loc_6A6D7E(
...
                     {
                         auto ride = get_ride(tileElement->AsTrack()->GetRideIndex());
-                        if (ride == nullptr || !ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_FLAT_RIDE))
+                        if (ride == nullptr)
+                        {
+                            continue;
+                        }
+
+                        const auto& rtd = ride->GetRideTypeDescriptor();
+                        if (!rtd.HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
```
tupaschoal, line 884:
```suggestion
                        if (ride == nullptr || !ride->GetRideTypeDescriptor(),HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
```

src/openrct2/world/Footpath.cpp
```diff
@@ -950,10 +956,17 @@ static void loc_6A6C85(
...
     {
         auto ride = get_ride(tileElementPos.element->AsTrack()->GetRideIndex());
-        if (ride == nullptr || !ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_FLAT_RIDE))
+        if (ride == nullptr)
         {
             return;
         }
+
+        const auto& rtd = ride->GetRideTypeDescriptor();
+        if (!rtd.HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
```
tupaschoal, line 965: Maybe merge it back into the if above

src/openrct2/world/Footpath.cpp
```diff
@@ -2031,7 +2044,8 @@ bool tile_element_wants_path_connection_towards(const TileCoordsXYZD& coords, co
                     if (ride == nullptr)
                         continue;
 
-                    if (!ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_FLAT_RIDE))
+                    const auto& rtd = ride->GetRideTypeDescriptor();
+                    if (!rtd.HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
```
tupaschoal, line 2048:
```suggestion
                    if (!ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
```

src/openrct2/world/Footpath.cpp
```diff
@@ -2120,7 +2134,11 @@ void footpath_remove_edges_at(const CoordsXY& footpathPos, TileElement* tileElem
     {
         auto rideIndex = tileElement->AsTrack()->GetRideIndex();
         auto ride = get_ride(rideIndex);
-        if (ride == nullptr || !ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_FLAT_RIDE))
+        if (ride == nullptr)
+            return;
+
+        const auto& rtd = ride->GetRideTypeDescriptor();
+        if (!rtd.HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
```
tupaschoal, line 2141:
```suggestion
        if (ride == nullptr || !ride->GetRideTypeDescriptor(),.HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
```

src/openrct2/world/Footpath.cpp
```diff
@@ -2271,5 +2289,6 @@ bool PathElement::IsLevelCrossing(const CoordsXY& coords) const
         return false;
     }
 
-    return RideTypeDescriptors[ride->type].HasFlag(RIDE_TYPE_FLAG_SUPPORTS_LEVEL_CROSSINGS);
+    const auto& rtd = ride->GetRideTypeDescriptor();
+    return rtd.HasFlag(RIDE_TYPE_FLAG_SUPPORTS_LEVEL_CROSSINGS);
```
tupaschoal, line 2293:
```suggestion
    return ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_SUPPORTS_LEVEL_CROSSINGS);
```

tupaschoal (commented):
This seems ok, though a few places have variables that I don't think we need (but might have been introduced due to the pointer nature of the previous iteration)

src/openrct2-ui/windows/Ride.cpp
duncanspumpkin: Think most people decided that using init statement wasn't preferred.

src/openrct2-ui/windows/TileInspector.cpp
duncanspumpkin: Is rideType really a good name for this surely rideTypeName or typeName would be better. rideType generally means the type used to get the rtd.

duncanspumpkin (approved):
All looks good.

tupaschoal:
I'd still like to see some of my suggested changes incorporated

src/openrct2-ui/windows/Ride.cpp
Gymnasiast: @tupaschoal I find both suggestions to have way worse readability.

tupaschoal: I don't find the second that bad, but your call...

PR #12001, opened 2020-06-19
Part of #9473: Vehicle sounds update functions

hdpoliveira:


src/openrct2/audio/audio.h
```diff
@@ -55,20 +55,21 @@ struct rct_ride_music_params
...
-    int16_t sound1_volume;
-    int16_t sound1_pan;
-    uint16_t sound1_freq;
-    SoundId sound2_id;
-    int16_t sound2_volume;
-    int16_t sound2_pan;
-    uint16_t sound2_freq;
-    void* sound1_channel;
-    void* sound2_channel;
+    Sound Sound1;
```
duncanspumpkin, line 71: Make this into an array rather than have a sound 1 and 2

hdpoliveira: I don't think this should be done.
Rather we should name the two sounds properly. Sound1 is TrackSounds and Sound2 is OtherSounds.

duncanspumpkin: Okay that would also work. Go for that. I was thinking we could use more ranged based for loops if they were an array

src/openrct2/audio/audio.h
```diff
@@ -55,20 +55,21 @@ struct rct_ride_music_params
...
+    SoundId Id;
+    int16_t Volume;
+    int16_t Pan;
+    uint16_t Frequency;
+    void* Channel;
+};
+
 struct rct_vehicle_sound
 {
     uint16_t id;
```
duncanspumpkin, line 69: this is `VehicleId` can you do a refactor after this renaming it.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1174,135 +1174,90 @@ static rct_vehicle_sound* vehicle_sounds_update_get_vehicle_sound(rct_vehicle_so
...
+{
+    if constexpr (type == SoundType::TrackNoises)
     {
-        sound->sound1_id = vehicle->sound1_id;
-        sound->sound1_pan = sound_params->pan_x;
-        sound->sound1_volume = volume;
-        sound->sound1_freq = sound_params->frequency;
-        uint16_t frequency = sound_params->frequency;
-        if (_soundParams[static_cast<uint8_t>(vehicle->sound1_id)][1] & 2)
+        if (_soundParams[static_cast<uint8_t>(id)][1] & 2)
```
duncanspumpkin, line 1187: In a future refactor _soundParams should be modified into a set of flags. There are only 3 flags: Looping, FrequencyMod1, FrequencyMod2. Have a think about a better name than FrequencyMod...

src/openrct2/audio/audio.h
```diff
@@ -55,20 +55,21 @@ struct rct_ride_music_params
     uint16_t frequency;
 };
 
+struct Sound
```
duncanspumpkin, line 58: I feel as though we might get name collisions on this name. Also not this structure is also used in `rct_ride_music`.

src/openrct2/audio/audio.h
hdpoliveira: Yes, I can :)

src/openrct2/ride/Vehicle.cpp
hdpoliveira: I prefer we have 3 functions, isLooping, trackSoundIsFrequencyModified, otherSoundIsFrequencyModified. All these functions will have switch statements, returning true or false, based on the sound id.

There are many advantages to this: the switch statements will fail in compilation if we add a new sound and don't specify them, we will avoid all those static_casts, the functions is well isolated and can be constexpr

src/openrct2/audio/audio.h
hdpoliveira: We should start looking into namespaces to avoid name collisions :)

I'll work on making rct_ride_music use Sound as well in a future PR

hdpoliveira: Actually, rct_ride_music has a field ```uint8_t tune_id```, instead of a ```SoundId Id``` (the underlying type is the same)

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: I'd prefer it to be an array as it's really meta data of the sound. If you see the rtd refactors we have been doing for the ride types we have been undoing functions like this and converting them into arrays of flags.

src/openrct2/audio/audio.h
duncanspumpkin: Yes namespaces would be good.

src/openrct2/ride/Vehicle.cpp
hdpoliveira: Ok!

PR #12002, opened 2020-06-19
Part of #9473: Create Vehicle::GetRide

hdpoliveira:


hdpoliveira:
It looks like there is something wrong in the dependencies of testpaint binary. It should be linking with Vehicle.cpp.o, but it isnt.

duncanspumpkin:
The way to fix test paint is you add your new function also to compat.cpp in the testpaint folder. Its a pain and at somepoint we will be redoing testpaint as it doesn't function correctly atm.

duncanspumpkin:
Could you therefore revert your revert and fix the testpaint compat.cpp instead.

PR #12005, opened 2020-06-20
Add ride wait time input

pbagot-1:
I made an input window open up when you click on the ride min waiting time or ride max waiting time widgets. Max length input is 3 so you can enter from 0 -> 999 but only values 0 -> 250 are allowed.

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -424,9 +424,11 @@ static constexpr const uint64_t window_ride_page_enabled_widgets[] = {
         (1ULL << WIDX_LOAD_CHECKBOX) |
         (1ULL << WIDX_LEAVE_WHEN_ANOTHER_ARRIVES_CHECKBOX) |
         (1ULL << WIDX_MINIMUM_LENGTH_CHECKBOX) |
+        (1ULL << WIDX_MINIMUM_LENGTH) |
         (1ULL << WIDX_MINIMUM_LENGTH_INCREASE) |
         (1ULL << WIDX_MINIMUM_LENGTH_DECREASE) |
         (1ULL << WIDX_MAXIMUM_LENGTH_CHECKBOX) |
+        (1ULL << WIDX_MAXIMUM_LENGTH) |
         (1ULL << WIDX_MAXIMUM_LENGTH_INCREASE) |
         (1ULL << WIDX_MAXIMUM_LENGTH_DECREASE) |
         (1ULL << WIDX_SYNCHRONISE_WITH_ADJACENT_STATIONS_CHECKBOX) |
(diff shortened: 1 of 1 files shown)


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -3651,6 +3665,42 @@ static void window_ride_operating_update(rct_window* w)
...
+    auto rideSetSetting = RideSetSettingAction(w->number, RideSetSetting::MaxWaitingTime, time);
+    GameActions::Execute(&rideSetSetting);
+}
+
+static void window_ride_operating_textinput(rct_window* w, rct_widgetindex widgetIndex, char* text)
+{
+    if ((widgetIndex != WIDX_MINIMUM_LENGTH && widgetIndex != WIDX_MAXIMUM_LENGTH) || text == nullptr)
+        return;
+
+    int32_t time32;
```
Gymnasiast, line 3685: We don't separate the declaration and definition anymore, unless necessary.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -3651,6 +3665,42 @@ static void window_ride_operating_update(rct_window* w)
...
+    char* end;
+
+    time32 = strtol(text, &end, 10);
+    if (*end == '\0')
+    {
+        uint8_t time = (uint8_t)std::clamp(time32, 0, 250);
```
Gymnasiast, line 3691: You need to use a C++-style cast.

But I'm a bit puzzled as to why you write to a `time32` and then cast to uint8_t in the first place.

prodscary: It was intended to make it so negative number input by the user would get turned into 0, but I suppose this case doesn't really matter

src/openrct2-ui/windows/Ride.cpp
prodscary: Fixed.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -3651,6 +3665,44 @@ static void window_ride_operating_update(rct_window* w)
...
+    uint8_t time = strtol(text, &end, 10);
+
+    if (*end == '\0')
+    {
+        uint8_t minTime = 0;
+        uint8_t maxTime = 250;
+        time = std::clamp(time, minTime, maxTime);
+
+        // if widget is min waiting time set it, else set max waiting time
+        if (widgetIndex == WIDX_MINIMUM_LENGTH)
```
Gymnasiast, line 3695: I prefer a switch case here. After all, we might want to add inputs for other widgets on that tab later on.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -3488,10 +3496,16 @@ static void window_ride_operating_mouseup(rct_window* w, rct_widgetindex widgetI
             set_operating_setting(
                 w->number, RideSetSetting::Departure, ride->depart_flags ^ RIDE_DEPART_WAIT_FOR_MINIMUM_LENGTH);
             break;
+        case WIDX_MINIMUM_LENGTH:
+            window_text_input_raw_open(w, WIDX_MINIMUM_LENGTH, STR_ENTER_NEW_VALUE, STR_ENTER_NEW_VALUE, _minWaitInputText, 4);
```
Gymnasiast, line 3500: Why the raw variant, specifically? The Water window uses `window_text_input_open` and it uses a simple inter input as well.

tupaschoal:
Hello @pbagot-1 ,

I'm converting this to a draft PR. If you can tackle the issues pointed out by Michael, please do (and also mark them as resolved when you do). After that, set it as ready for review, please :)

tupaschoal:
Hey @pbagot-1 this has been stalled for a while, but it's a cool feature. If you can resume and solve the issues, please reopen the PR!

PR #12007, opened 2020-06-20
Close #11971: Replace ride groups with new ride types

### LINKED ISSUE #11971: Convert ride groups to separate coaster types
Currently, some ride types are subdivided into two ride groups. For example, the Twister Roller Coaster is divided into the Steel Twister and the Hyper-Twister.

This division complicates the code somewhat, and might become more problematic once we start turning ride types into objects.

The refactor would consist of the following:
- Copy the RTD of every ride type that has ride groups.
- Modify the naming, available track pieces, maximum height and flags from the ride groups.
- Modify the code to remove references to ride groups and the ride group manager.
- Introduce code to convert ride types in the S4Importer and S6Importer (as well as the S6Export/ParkImporter, depending on whether this refactor is carried out before or after the move to the new save format).

Gymnasiast:
Currently, five ride types are subdivided using "ride groups". While they have worked properly, they do complicate the code a bit in places. They will also be hard to port once ride types become actual objects.

This PR therefore creates five new ride types, converting saves, objects and track designs at load/save time. The reduced complexity during playing means this PR ends up with more than 450 fewer lines of code. (Or almost 200 fewer, if you count the added RTDs as code.)

This is also where the work to move most ride type metadata to RTDs really shines: it wasn't difficult to add these five.


src/openrct2/libopenrct2.vcxproj
```diff
@@ -764,4 +763,4 @@
     <ClCompile Include="world\Wall.cpp" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
-</Project>
\ No newline at end of file
+</Project>
```
Gymnasiast, line 766: I will revert this. (IDE did this automatically.)

src/openrct2/rct1/RCT1.h
```diff
@@ -1281,5 +1281,6 @@ enum
 
 void load_from_sv4(const char* path);
 void load_from_sc4(const char* path);
+ObjectEntryIndex RCT1RideTypeToOpenRCT2RideType(uint8_t rct1RideType, const rct_ride_entry* rideEntry);
```
Gymnasiast, line 1284: Remove

src/openrct2/localisation/StringIds.h
```diff
@@ -3544,22 +3544,14 @@ enum
...
-    STR_HYPER_TWISTER_GROUP_DESC = 6117,
-    STR_JUNIOR_RC_GROUP_DESC = 6118,
-    STR_CLASSIC_MINI_COASTER_GROUP_DESC = 6119,
+    STR_RIDE_NAME_HYPERCOASTER = 6105,
+    STR_RIDE_NAME_MONSTER_TRUCKS = 6107,
+    STR_RIDE_NAME_HYPER_TWISTER = 6109,
+    STR_RIDE_NAME_CLASSIC_MINI_ROLLER_COASTER = 6111,
+    STR_RIDE_DESCRIPTION_HYPERCOASTER = 6113,
+    STR_RIDE_DESCRIPTION_MONSTER_TRUCKS_GROUP = 6115,
+    STR_RIDE_DESCRIPTION_TWISTER_ROLLER_COASTER = 6116,
```
pizza2004, line 3553: Shouldn't this be 6117?

Gymnasiast: No, this is an exception.

src/openrct2/ride/TrackDesignRepository.cpp
```diff
@@ -78,10 +78,15 @@ class TrackDesignFileIndex final : public FileIndex<TrackRepositoryItem>
         auto td6 = track_design_open(path.c_str());
         if (td6 != nullptr)
         {
+            auto* rawObject = object_repository_load_object(&td6->vehicle_object);
```
duncanspumpkin, line 81: This is causing a race condition when building the track design list (looks to be issues with the image list). I don't think you should be loading and unloading objects in threads. You will either have to have this protected by a mutex or fix object loading / unloading to be thread safe.

Gymnasiast: Don't think I can pull off the latter, so I'd want to do the former. How do I introduce one? Is there any other place in OpenRCT2 that uses a mutex?

IntelOrca: Yes, just search for `std::mutex`. I think the object downloader and server list use them.

Gymnasiast: I have now added one. I hope I did it correctly.

duncanspumpkin:
Its causing a barely noticeable problem but peep::RideTypesBeenOn is no longer correct. Would it be too much of a pain to provide an importer/exporter for it?

Gymnasiast:
Yeah, I'll try that.

duncanspumpkin:
I've had a good play about with this and I think there is no way to merge this without the new save format as it is will cause desyncs on multiplayer. The rideTypesHaveRidden cannot be rebuilt on load correctly. The only way I can see this working is that we make it so that the ride types been on not listen to the new ride types until the new save format.

Gymnasiast:
> The only way I can see this working is that we make it so that the ride types been on not listen to the new ride types until the new save format.

After some discussion via PMs, this is what we went for.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -507,7 +507,14 @@ class S6Importer final : public IParkImporter
     {
         *dst = {};
         dst->id = rideIndex;
-        dst->type = src->type;
+
+        ObjectEntryIndex rideType = src->type;
+        if (RCT2RideTypeNeedsConversion(src->type))
+        {
+            auto* rideEntry = get_ride_entry(src->subtype);
+            rideType = RCT2RideTypeToOpenRCT2RideType(src->type, rideEntry);
```
Gymnasiast, line 515: Add null check

PR #12010, opened 2020-06-20
Part of #11571 - Create gfx_draw_string_left overload

TomasZilinek:
Third and final part of issue #11571. I replaced the rest of the calls and removed the old function definition and declaration.


tupaschoal:
Nice, thanks :)

PR #12011, opened 2020-06-20
Fix #7037 : Roller coasters starting with helix can be saved

### LINKED ISSUE #7037: Cannot save ride design if first piece is an upwards or downwards helix.
**OS:** [Windows 8.1]
**Version:** [0.1.2]
**Commit/Build:** [3ba0cc3]

Possibly related to #400. 

I was looking in old issues, and found one where tracks that start with a banked piece end up a bit higher when they are saved and placed. I tried reproducing this, and now found that a track design that starts with either a small or large helix (up or down) cannot be saved at all, which is different behaviour than before. A normal banked turn seems to be okay. I tested this with both looping coasters and corkscrew coasters.

The message you get is: Can't save track design. Ride is too large, contains too many elements or scenery is too spread out. 

I don't think this is a recent issue, because no-one starts their ride with a helix so this could have been hidden for a long time. 

- [ ] Multiplayer?

**Steps to reproduce:**
1. Build a launched ride which starts with a banked helix (up or down), like a looping or corkscrew roller coaster.
2. Test ride, then try to save ride design.

**Screenshots / Video:**

frutiemax:
[BuggedRollercoaster3.zip](https://github.com/OpenRCT2/OpenRCT2/files/4808438/BuggedRollercoaster3.zip)

Both the roller coasters couldn't be saved to a track design, now they can. BuggedRollerCoaster2 also couldn't be placed without a z coordinates check on the entrance, exit and stations.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -1177,7 +1177,7 @@ std::optional<CoordsXYZ> sub_6C683D(
     const CoordsXYZD& location, int32_t type, uint16_t extra_params, TileElement** output_element, uint16_t flags)
 {
     // Find the relevant track piece, prefer sequence 0 (this ensures correct behaviour for diagonal track pieces)
-    auto trackElement = map_get_track_element_at_of_type_seq(location, type, 0);
+    auto trackElement = map_get_track_element_at_of_type_seq(location, type, 0, true);
     if (trackElement == nullptr)
     {
         trackElement = map_get_track_element_at_of_type(location, type);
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
I'd prefer we fixed the bug properly rather than introduce another hack.

Gymnasiast:
@frutiemax Did you manage to track down why/where it ends up with an incorrect Z coordinate in the first place?

frutiemax:
@Gymnasiast I found out that the track element base height is not the same than the tile element base height for the helix elements. The base height of the ending tile of all helix elements is at an invalid height, while the base height of the track element is at a valid one. I've also seen it using the tile inspector. So, in the track design saving code, it needs to convert the tile element to a track element, but the z coordinates do not match up if the roller coaster starts with an helix.

PR #12016, opened 2020-06-21
Make interfaces use ScreenCoordsXY instead of x-y

tupaschoal:
A collection of small changes to favour ScreenCoordsXY.

There are two weird pieces of code that might be bugs, I'll comment on them for inputs

### DIFF
--- a/src/openrct2/drawing/Drawing.String.cpp
+++ b/src/openrct2/drawing/Drawing.String.cpp
@@ -421,7 +421,7 @@ int32_t string_get_height_raw(char* buffer)
  * ticks    : ebp >> 16
  */
 void gfx_draw_string_centred_wrapped_partial(
-    rct_drawpixelinfo* dpi, int32_t x, int32_t y, int32_t width, int32_t colour, rct_string_id format, void* args,
+    rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, int32_t width, int32_t colour, rct_string_id format, void* args,
     int32_t ticks)
 {
     int32_t numLines, fontSpriteBase, lineHeight, lineY;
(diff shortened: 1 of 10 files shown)


src/openrct2/drawing/Drawing.String.cpp
```diff
@@ -934,18 +933,18 @@ static void ttf_process_initial_colour(int32_t colour, text_draw_info* info)
...
 
     text_draw_info info;
     info.font_sprite_base = gCurrentFontSpriteBase;
     info.flags = gCurrentFontFlags;
-    info.startX = x;
-    info.startY = x;
-    info.x = x;
-    info.y = y;
+    info.startX = coords.x;
+    info.startY = coords.y;
```
tupaschoal, line 945: Weird code previously

src/openrct2/drawing/Drawing.String.cpp
```diff
@@ -992,16 +991,16 @@ static int32_t ttf_get_string_width(const utf8* text)
...
 {
     text_draw_info info;
     info.font_sprite_base = gCurrentFontSpriteBase;
     info.flags = gCurrentFontFlags;
-    info.startX = x;
-    info.startY = x;
-    info.x = x;
-    info.y = y;
+    info.startX = coords.x;
+    info.startY = coords.y;
```
tupaschoal, line 1001: Weird code previously

tupaschoal:
This had gone off my radar, I rebased it. Do note my comments above though

src/openrct2/drawing/Drawing.String.cpp
Gymnasiast: I wouldn't be surprised if they were simply copy-and-paste errors. I wonder if it will make any noticeable change.

tupaschoal: I assumed so

PR #12028, opened 2020-06-21
Fix #12020 - Use ScreenCoordsXY on gfx_draw_sprite_palette_set_software

### LINKED ISSUE #12020: Use ScreenCoordsXY on gfx_draw_sprite_palette_set_software
- Create an overload of `gfx_draw_sprite_palette_set_software
` with the same name, but that takes a `const ScreenCoordsXY&` object instead of an `int32_t x` and and an `int32_t y` 

frutiemax:


src/openrct2-ui/drawing/engines/opengl/TextureCache.cpp
```diff
@@ -320,7 +321,9 @@ rct_drawpixelinfo TextureCache::GetGlyphAsDPI(uint32_t image, const PaletteMap&
     int32_t height = g1Element->height;
 
     rct_drawpixelinfo dpi = CreateDPI(width, height);
-    gfx_draw_sprite_palette_set_software(&dpi, ImageId::FromUInt32(image), -g1Element->x_offset, -g1Element->y_offset, palette);
+
+    const auto& glyphCoords = ScreenCoordsXY({ -g1Element->x_offset, -g1Element->y_offset });
```
tupaschoal, line 325:
```suggestion
    const auto glyphCoords = ScreenCoordsXY{ -g1Element->x_offset, -g1Element->y_offset };
```
No need for the reference in here, as you've just allocated the value

Also, either use uniform initialization `{x , y}` or the constructor directly `(x ,y)`. In here theoretically you're creating a `ScreenCoordsXY` object and then calling the copy ctor. The compiler probably optimizes this away, but why add a layer of complexity when we can be simple to begin with :)

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -410,7 +410,9 @@ void FASTCALL gfx_draw_sprite_software(rct_drawpixelinfo* dpi, ImageId imageId,
         {
             palette = PaletteMap::GetDefault();
         }
-        gfx_draw_sprite_palette_set_software(dpi, imageId, x, y, *palette);
+
+        const auto& spriteCoords = ScreenCoordsXY({ x, y });
```
tupaschoal, line 414:
```suggestion
        const auto spriteCoords = ScreenCoordsXY{ x, y };
```

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -423,9 +425,13 @@ void FASTCALL gfx_draw_sprite_software(rct_drawpixelinfo* dpi, ImageId imageId,
...
  * y (dx)
  */
+
 void FASTCALL gfx_draw_sprite_palette_set_software(
-    rct_drawpixelinfo* dpi, ImageId imageId, int32_t x, int32_t y, const PaletteMap& paletteMap)
+    rct_drawpixelinfo* dpi, ImageId imageId, const ScreenCoordsXY& coords, const PaletteMap& paletteMap)
 {
+    int32_t x = coords.x;
+    int32_t y = coords.y;
+
```
tupaschoal, line 434:
```suggestion
```

No need to do this, we can just shift them in place below

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -442,8 +448,10 @@ void FASTCALL gfx_draw_sprite_palette_set_software(
         zoomed_dpi.width = dpi->width >> 1;
         zoomed_dpi.pitch = dpi->pitch;
         zoomed_dpi.zoom_level = dpi->zoom_level - 1;
+
+        const auto& spriteCoords = ScreenCoordsXY({ x >> 1, y >> 1 });
         gfx_draw_sprite_palette_set_software(
-            &zoomed_dpi, imageId.WithIndex(imageId.GetIndex() - g1->zoomed_offset), x >> 1, y >> 1, paletteMap);
+            &zoomed_dpi, imageId.WithIndex(imageId.GetIndex() - g1->zoomed_offset), spriteCoords, paletteMap);
```
tupaschoal, line 454:
```suggestion
            &zoomed_dpi, imageId.WithIndex(imageId.GetIndex() - g1->zoomed_offset), {spriteCoords.x >> 1, spriteCoords.y >> 1 }, paletteMap);
```

(This will probably break formatting)

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -442,8 +448,10 @@ void FASTCALL gfx_draw_sprite_palette_set_software(
         zoomed_dpi.width = dpi->width >> 1;
         zoomed_dpi.pitch = dpi->pitch;
         zoomed_dpi.zoom_level = dpi->zoom_level - 1;
+
+        const auto& spriteCoords = ScreenCoordsXY({ x >> 1, y >> 1 });
```
tupaschoal, line 452:
```suggestion
```
Same

src/openrct2/drawing/X8DrawingEngine.cpp
```diff
@@ -743,13 +743,16 @@ void X8DrawingContext::DrawSpriteSolid(uint32_t image, int32_t x, int32_t y, uin
     uint8_t palette[256];
     std::fill_n(palette, sizeof(palette), colour);
     palette[0] = 0;
+
+    const auto& spriteCoords = ScreenCoordsXY({ x, y });
```
tupaschoal, line 747:
```suggestion
    const auto spriteCoords = ScreenCoordsXY{ x, y };
```

Same as before

src/openrct2/drawing/X8DrawingEngine.cpp
```diff
@@ -743,13 +743,16 @@ void X8DrawingContext::DrawSpriteSolid(uint32_t image, int32_t x, int32_t y, uin
...
     gfx_draw_sprite_palette_set_software(
-        _dpi, ImageId::FromUInt32((image & 0x7FFFF) | IMAGE_TYPE_REMAP), x, y, PaletteMap(palette));
+        _dpi, ImageId::FromUInt32((image & 0x7FFFF) | IMAGE_TYPE_REMAP), spriteCoords, PaletteMap(palette));
 }
 
 void X8DrawingContext::DrawGlyph(uint32_t image, int32_t x, int32_t y, const PaletteMap& paletteMap)
 {
-    gfx_draw_sprite_palette_set_software(_dpi, ImageId::FromUInt32(image), x, y, paletteMap);
+    const auto& glyphCoords = ScreenCoordsXY({ x, y });
+    gfx_draw_sprite_palette_set_software(_dpi, ImageId::FromUInt32(image), glyphCoords, paletteMap);
```
tupaschoal, line 755:
```suggestion
    gfx_draw_sprite_palette_set_software(_dpi, ImageId::FromUInt32(image), { x, y }, paletteMap);
```
This is one other option, makes it simpler

src/openrct2/drawing/Drawing.Sprite.cpp
frutiemax: At lines 469, 470, the values x and y are modified.

tupaschoal: Ah, I see.

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -423,9 +425,13 @@ void FASTCALL gfx_draw_sprite_software(rct_drawpixelinfo* dpi, ImageId imageId,
  * x (cx)
  * y (dx)
  */
+
```
tupaschoal, line 428:
```suggestion
```
Unrelated change, just revert it please

PR #12033, opened 2020-06-22
Fix #12019 : Use ScreenCoordsXY on gfx_draw_sprite_software

### LINKED ISSUE #12019: Use ScreenCoordsXY on gfx_draw_sprite_software
- Create an overload of `gfx_draw_sprite_software
` with the same name, but that takes a `const ScreenCoordsXY&` object instead of an `int32_t x` and and an `int32_t y` 

frutiemax:


src/openrct2-ui/drawing/engines/opengl/TextureCache.cpp
```diff
@@ -186,7 +186,7 @@ void TextureCache::GeneratePaletteTexture()
         if (g1Index)
         {
             auto element = gfx_get_g1_element(*g1Index);
-            gfx_draw_sprite_software(&dpi, ImageId(*g1Index), -element->x_offset, y - element->y_offset);
+            gfx_draw_sprite_software(&dpi, ImageId(*g1Index), ScreenCoordsXY{ -element->x_offset, y - element->y_offset });
```
tupaschoal, line 189:
```suggestion
            gfx_draw_sprite_software(&dpi, ImageId(*g1Index), { -element->x_offset, y - element->y_offset });
```

Issue #10960, opened 2020-03-17
Time control in multiplayer


dkoether:
Hello everyone,

is it possible to add the time speed controls to the multiplayer? 

Best regards and thanks in advance!

frutiemax:
Perhaps we need to add a dedicated game action for setting the game speed.

geoff-B:
@frutiemax What's the progress on this, and what still needs to be done to make this work?

frutiemax:
This should work as is.

PR #12044, opened 2020-06-24
Part of #10960: Pause and setting game speed enabled in multiplayer

frutiemax:
- [x] Enable pause actions for the users in group with toggle pause permission
- [x] Enable set game simulation speed for the users in group with toggle pause permission
- [x] Disable pause on client's side when he is not allowed to pause the game on the server side
- [x] Disable setting game speed on client's side when he is not allowed to set game speed on the server side
- [x] Invalidate the pause and fast forward buttons when receiving a TogglePause and SetSimulationSpeed action
- [ ] Rename toggle pause permission to set game simulation speed permission
- [x] Hide pause and fast forward buttons when the client has not the permissions. Seems to work partially as it doesn't redraw the top toolbar when permissions change
- [x] When client joins the server, make sure that the server is correctly sending the pause status and simulation speed
- [ ] When client joins the server when paused, force a redraw of all sprites
- [x] When sever pauses game, force the client to call toggle_pause to stop audio
- [x] When server pauses game when gameSpeed != 1, stop the audio
- [ ] When server is paused, changing togglePause permissions for the user do not invalidate the top toolbar

### DIFF
--- a/src/openrct2-ui/windows/TopToolbar.cpp
+++ b/src/openrct2-ui/windows/TopToolbar.cpp
@@ -30,6 +30,7 @@
 #include <openrct2/actions/BannerSetColourAction.hpp>
 #include <openrct2/actions/ClearAction.hpp>
 #include <openrct2/actions/FootpathSceneryPlaceAction.hpp>
+#include <openrct2/actions/GameSetSimulationSpeed.hpp>
 #include <openrct2/actions/LandLowerAction.hpp>
 #include <openrct2/actions/LandRaiseAction.hpp>
 #include <openrct2/actions/LandSmoothAction.hpp>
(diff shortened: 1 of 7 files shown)


frutiemax:
I need some help with forcing sprites to redraw when a client joins a paused server.

duncanspumpkin:
There is a `gfx_invalidate_screen` function that can force a full redraw of the screen. You can have that happen on a client joining. I'm surprised that that is an issue though I was pretty sure joining would force a redraw anyway.

duncanspumpkin:
I was pretty sure we disabled this for a reason. This will need a good bit of testing before merging as there might be issues.

frutiemax:
@duncanspumpkin I can't seem to find the correct place to insert the gfx_invalidate_screen call. Do you have any clue where I could place it?

Issue #11437, opened 2020-04-24
Refactor platform source files


AaronVanGeffen:
Currently, the platform source files are halfway in between refactoring. We started to rewrite routines to use more modern C++, but some legacy (C) routines were left behind. The wait was for `std::filesystem` to be available on all platforms, but that's been available since #10261.

Note that functions should not be moved as-is. We prefer to move to take the opportunity to rewrite them to use `std::string` and `std::filesystem` routines instead.

The following are the 'new' platform files:

- Platform.Android.cpp
- Platform.Linux.cpp
- Platform.macOS.mm
- Platform.Posix.cpp
- Platform.Win32.cpp
- Platform2.h

The following are files whose functions need to be refactored into the new files:

- Android.cpp
- Linux.cpp
- macos.mm
- platform.h
- Posix.cpp
- Shared.cpp
- Windows.cpp

Note that Crash.cpp and Crash.h are not on this list.

Please contact us here or on Discord if you'd like to work on this.

IntelOrca:
And `Platform2.h` will finally be renamed to `Platform.h`.

PR #12052, opened 2020-06-24
Part of #11437: Move GetDateLocal and GetTimeLocal in Platform2.h

frutiemax:


duncanspumpkin:
I'm not even sure this should be in the platform files as ctime has never changed in the life of the project. but there must have been a 'time' where this made sense.

IntelOrca:
The functions should be changed to return the struct they are writing to rather than using an out parameter.

src/openrct2/platform/Platform.Win32.cpp
```diff
@@ -35,6 +36,7 @@
 #    include "Platform2.h"
 #    include "platform.h"
 
+#    include <ctime>
```
Gymnasiast, line 39: Is this change necessary? You didn´t change anything else to this file.

PR #12054, opened 2020-06-25
Part of #11159: limit_autosave_count uses std::vector instead of utf8**

frutiemax:


src/openrct2/Game.cpp
```diff
@@ -726,47 +718,39 @@ static void limit_autosave_count(const size_t numberOfFilesToKeep, bool processL
...
-                safe_strcat_path(autosaveFiles[i], scanner->GetPathRelative(), sizeof(utf8) * MAX_PATH);
+                safe_strcat_path(autosaveFiles[i].data(), "autosave", sizeof(utf8) * MAX_PATH);
+                safe_strcat_path(autosaveFiles[i].data(), scanner->GetPathRelative(), sizeof(utf8) * MAX_PATH);
             }
         }
     }
 
-    qsort(autosaveFiles, autosavesCount, sizeof(char*), compare_autosave_file_paths);
+    std::sort(autosaveFiles.begin(), autosaveFiles.end(), [](const auto& saveFile0, const auto& saveFile1) {
+        return strcmp(saveFile0.c_str(), saveFile1.c_str()) < 0;
```
tupaschoal, line 744:
```suggestion
        return saveFile0.compare(saveFile1) < 0;
```

src/openrct2/Game.cpp
```diff
@@ -726,47 +718,39 @@ static void limit_autosave_count(const size_t numberOfFilesToKeep, bool processL
...
-
+    auto autosaveFiles = std::vector<std::string>(autosavesCount);
     {
         auto scanner = std::unique_ptr<IFileScanner>(Path::ScanDirectory(filter, false));
         for (size_t i = 0; i < autosavesCount; i++)
         {
-            autosaveFiles[i] = static_cast<utf8*>(malloc(sizeof(utf8) * MAX_PATH));
-            std::memset(autosaveFiles[i], 0, sizeof(utf8) * MAX_PATH);
-
+            autosaveFiles[i].resize(MAX_PATH, 0);
```
tupaschoal, line 726: Not sure we need this, but I'm a bit sleepy, :X

duncanspumpkin: You will need to do this because the string is being used as an array of data. I assume in the future the idea would be to correct the platform functions to return a std::string

PR #12056, opened 2020-06-25
Introduce widget::midX and widget::midY helper functions

AaronVanGeffen:
As #12053 was well-received, let's continue this train.

### DIFF
--- a/src/openrct2-ui/interface/Widget.cpp
+++ b/src/openrct2-ui/interface/Widget.cpp
@@ -345,12 +345,7 @@ static void widget_text_centred(rct_drawpixelinfo* dpi, rct_window* w, rct_widge
     int32_t r = w->windowPos.x + widget->right;
 
     if (widget->type == WWT_BUTTON || widget->type == WWT_TABLE_HEADER)
-    {
-        if (widget->height() >= 10)
-            topLeft.y += std::max<int32_t>(widget->top, widget->top + (widget->height() / 2) - 5);
-        else
-            topLeft.y += widget->top - 1;
-    }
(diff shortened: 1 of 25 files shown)


src/openrct2-ui/interface/Widget.cpp
```diff
@@ -346,9 +346,8 @@ static void widget_text_centred(rct_drawpixelinfo* dpi, rct_window* w, rct_widge
 
     if (widget->type == WWT_BUTTON || widget->type == WWT_TABLE_HEADER)
     {
-        int32_t height = widget->height();
-        if (height >= 10)
-            topLeft.y += std::max<int32_t>(widget->top, widget->top + (height / 2) - 5);
+        if (widget->height() >= 10)
+            topLeft.y += std::max<int32_t>(widget->top, widget->top + (widget->height() / 2) - 5);
```
Gymnasiast, line 350: This is not an improvement. Did you plan to use midY here?

AaronVanGeffen: At the moment, this is just inlining the `height` variable, as it is not needed now this function exists. However, using `midY` would be shorter and may be preferable, yes. I’m not sure it yields the exact same result though. Worth looking into.

Gymnasiast: I think *just* inlining the variable only makes for slightly harder to read code and does not have any advantages. I'd like to see this either reverted or modified in a way that makes it easier to read.

Review note: this also applies to other occurrences in this file. All other files look good to me.

PR #12057, opened 2020-06-25
Make tile element direction accessible for scripts

tmstieff:
Add the direction field from the tile element class to the ScTileElement class passed to the

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -420,6 +420,7 @@ declare global {
         baseHeight: number;
         clearanceHeight: number;
         isHidden: boolean; /** Take caution when changing this field, it may invalidate TileElements you have stored in your script. */
+        direction: 0 | 1 | 2 | 3;
     }
 
     interface SurfaceElement extends BaseTileElement {
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -420,7 +420,6 @@ declare global {
         baseHeight: number;
         clearanceHeight: number;
         isHidden: boolean; /** Take caution when changing this field, it may invalidate TileElements you have stored in your script. */
-        direction: 0 | 1 | 2 | 3;
     }
 
     interface SurfaceElement extends BaseTileElement {
@@ -450,6 +449,8 @@ declare global {
 
         addition: number | null;
         isAdditionBroken: boolean;
+
+        direction: 0 | 1 | 2 | 3;
     }
 
     interface TrackElement extends BaseTileElement {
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
You can't set a large tile element or park entrance rotation. It will really mess up a bunch of expectations. It's the reason you can't set them from the tile inspector.

Gymnasiast:
We should perhaps create a function `CanRotate()` on `TileElementBase` and have both the Tile Inspector and this new function call it.

IntelOrca:
I would only define `direction` on the tile element types that support it within `openrct2.d.ts`. Then as @duncanspumpkin said, make sure it only read / writes direction for those valid element types.

duncanspumpkin:
Reading the direction is still useful for all types. It's just the setting which is an issue.

tmstieff:
I think I would have to defer this feature to someone with more knowledge of the game data structures.

duncanspumpkin:
I did some more investigating and tracks seem to handle this safely after all. You can get in some weird situations with tracks but the game just prevents you from doing anything. For large scenery it will definitely cause issues if you rotate and delete parts so I would want large scenery to not have rotation. Park entrances I doubt anyone will notice too much rotation related issues.

distribution/openrct2.d.ts
```diff
@@ -449,6 +449,8 @@ declare global {
 
         addition: number | null;
         isAdditionBroken: boolean;
+
+        direction: 0 | 1 | 2 | 3;
```
IntelOrca, line 453: It might be nice to declare a type for this now. E.g.:
`type Direction = 0 | 1 | 2 | 3`

tmstieff: Good call, definitely clearer.

duncanspumpkin:
Is there no way to have the property readable but not writeable for the problem entries? I know at least one plugin developer was wanting to read the rotations for all elements.

tmstieff:
I can add a readonly direction field to the other elements. Do you know if it works as expected for the large scenery and path elements?

IntelOrca:
@duncanspumpkin yes, you can pass nullptr to the setter method.

src/openrct2/scripting/ScTile.hpp
```diff
@@ -1013,6 +1032,7 @@ namespace OpenRCT2::Scripting
                 ctx, &ScTileElement::isAdditionBroken_get, &ScTileElement::isAdditionBroken_set, "isAdditionBroken");
             dukglue_register_property(
                 ctx, &ScTileElement::isAdditionGhost_get, &ScTileElement::isAdditionGhost_set, "isAdditionGhost");
+            dukglue_register_property(ctx, &ScTileElement::direction_get, &ScTileElement::direction_set, "direction");
```
Broxzier, line 1035: This is added under the "Footpath only" section, but applies to all element types, right? This should be moved up to be listed under "All".

tmstieff: I missed that, I'll move it up.

PR #12060, opened 2020-06-26
Fix #11963 - Refactor Peep::OutsideOfPark

### LINKED ISSUE #11963: Refactor Peep::OutsideOfPark
Peep::OutsideOfPark is a bool and should be converted to a bool.

TomasZilinek:
(issue #11963) Changed the definition of Peep::OutsideOfPark from uint8_t to bool and refactored all its occurences in code accordingly.

### DIFF
--- a/src/openrct2-ui/windows/GuestList.cpp
+++ b/src/openrct2-ui/windows/GuestList.cpp
@@ -251,7 +251,7 @@ void window_guest_list_refresh_list()
     for (auto peep : EntityList<Guest>(SPRITE_LIST_PEEP))
     {
         sprite_set_flashing(peep, false);
-        if (peep->OutsideOfPark != 0)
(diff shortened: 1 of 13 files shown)


TomasZilinek:
> Looks fine. What is happening in the importers and exporters

What importers and exporters? I am not familiar with this terminology.

duncanspumpkin:
S4 and S6 importer and S6 exporter for the save files. You should convert this explicitly.

TomasZilinek:
> Looks fine to me, though I'm not 100% convinced on the import/export changes

What should I do differently? Or did I forget something?

duncanspumpkin:
I think it is fine to do it like that for import and export. I only wanted something explicit I don't feel a ternary would improve the readability.

PR #12061, opened 2020-06-27
Fix #11394: Refactor Vehicle::UpdateTrackMotion[Forwards|Backwards]

### LINKED ISSUE #11394: Refactor Vehicle::UpdateTrackMotionForwards/UpdateTrackMotionBackwards
`Vehicle::UpdateTrackMotionForwards` and `UpdateTrackMotionBackwards` in `src/openrct2/ride/Vehicle.cpp` has not had much attention since it was implemented apart from a few tidy up exercises. This issue is for refactoring it.

There are a number of issues with this function:

1. It uses the `registers` struct for variables.
2. It abuses goto.
3. Its far too long

For item 1. you will need to work out the lifetime of each of the registers that the `registers` struct is being used for. You will also need to understand what the `registers` struct represents (have a view of its implementation). For example regs.cx is set in L8110 but then never used. This can very easily be removed as it is clear that none of regs.cl, regs.ch, regs.ecx are being used in this function. Likewise for L8151 regs.eax is set and immediately used in L8153 and then set again in L8198. This means that regs.eax at L8198 represents a completely new variable. Go through the function and remove all of the registers uses and replace them with appropriately named variables.

For item 2. there are 2 gotos used you will need to take a careful look at the function and work out how to undo the gotos. Perhaps they can be replaced by splitting the function up into 2 functions.

Item 3 try look for similarities between Forwards and Backwards can functions be pulled out and combined.

In addition make sure all local variables are in lower camelCase and all function names are in upper CamelCase. I think this has already been done. https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style

Issue currently assigned to @mustaphaelghoul

glaubertadeu:
1. Remove use of registers and gotos on Vehicle::UpdateTrackMotion[Forwards|Backwards] functions.
2. Split a part of code to a function (VehicleColision) and remove duplicity.

### DIFF
--- a/src/openrct2-ui/windows/Guest.cpp
+++ b/src/openrct2-ui/windows/Guest.cpp
@@ -1124,7 +1124,7 @@ void window_guest_overview_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     screenPos.x = widget->right - widget->left - w->list_information_type;
     peep_thought_set_format_args(&peep->Thoughts[i]);
-    gfx_draw_string_left(&dpi_marquee, STR_WINDOW_COLOUR_2_STRINGID, gCommonFormatArgs, COLOUR_BLACK, screenPos.x, 0);
+    gfx_draw_string_left(&dpi_marquee, STR_WINDOW_COLOUR_2_STRINGID, gCommonFormatArgs, COLOUR_BLACK, { screenPos.x, 0 });
 }
 
 /**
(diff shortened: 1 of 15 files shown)


duncanspumpkin:
Split this up into different changes. It is too hard to follow when you change the registers and the goto's all in one.

glaubertadeu:
Ok, I'll do.

glaubertadeu:
I'm trying to commit only the regs replacement but I'm not familiar with git. Although I revert most of the code to original, it was not reflected on commit. I use "commit --amend" to update.

tupaschoal:
There are 3 pending issues with this PR:
- Needs commits to be split up
- Is breaking tests (which probably means that it messed something up)
- Needs rebase (https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2)

I'm converting it to draft for lack of activity

tupaschoal:
As this hasn't been touched in a while, I'm closing this PR. Please reopen if you rework on it and get to the final solution

PR #12074, opened 2020-06-28
Fix #12065 again. Fixed track design previews with diagonals

### LINKED ISSUE #12065: Pre-build ride minimap shows entrance and exit at incorrect location
**OS:** Windows 10
**Version:** 0.2.6
**Commit/Build:** d4a35c5

When building pre-build rides, the entrance and exit are sometimes either far away from the correct position, or not shown at all, on the little map in the top left corner.

Some rides seem to still have the entrance and exit on the map. Some in correct location, some not, and on other rides it's gone (probably outside of view/window bounds?).

**Screenshots / Video:**
This is the Cobra pre-build ride, from the Corkscrew Roller Coaster.
openrct2-incorrect-entrance-on-map

Another example of the Goliath Giga Coaster, where the minimap is showing track that doesn't even exist.
openrct2-not-existing-track

duncanspumpkin:
Fix #12065 again. Fixed track design previews with diagonals
Unsure when mistake was made but there was a variety of issues with this section of code. The rotation of the entrance was meant to be relative to the track design rotation not the last track pieces rotation. This caused issues for tracks that were not complete circuits. In addition there was a mistake for diagonal track pieces where the location of the end piece was not incremented this caused the diagonal piece to act as though it didn't exist breaking the preview.

### DIFF
--- a/src/openrct2-ui/windows/TrackDesignPlace.cpp
+++ b/src/openrct2-ui/windows/TrackDesignPlace.cpp
@@ -567,11 +567,12 @@ static void window_track_place_draw_mini_preview(TrackDesign* td6)
 static void window_track_place_draw_mini_preview_track(
     TrackDesign* td6, int32_t pass, const CoordsXY& origin, CoordsXY min, CoordsXY max)
 {
-    uint8_t rotation = (_currentTrackPieceDirection + get_current_rotation()) & 3;
+    const uint8_t rotation = (_currentTrackPieceDirection + get_current_rotation()) & 3;
 
     const rct_preview_track** trackBlockArray = (ride_type_has_flag(td6->type, RIDE_TYPE_FLAG_HAS_TRACK)) ? TrackBlocks
                                                                                                           : FlatRideTrackBlocks;
     CoordsXY curTrackStart = origin;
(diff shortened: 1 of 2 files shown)


src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -624,20 +625,19 @@ static void window_track_place_draw_mini_preview_track(
         }
 
         // Change rotation and next position based on track curvature
-        rotation &= 3;
+        curTrackRotation &= 3;
         const rct_track_coordinates* track_coordinate = &TrackCoordinates[trackType];
 
-        trackType *= 10;
```
tupaschoal, line 630: Is this gone intentionally?

duncanspumpkin: Yeah. I went and checked the assembly as didn't know what it was here for. This *10 was the size is the structure it's accessing. It hasn't been required for a good 2+ years.

tupaschoal: Ok, thanks!

PR #12075, opened 2020-06-28
Part of #11159: Remove malloc and free in openrct2-win.cpp

frutiemax:


Gymnasiast:
Would appreciate a second opinion of @duncanspumpkin or @IntelOrca .

src/openrct2-win/openrct2-win.cpp
```diff
@@ -17,64 +17,50 @@
...
     {
-        free(argv[i]);
+        argv[i] = ConvertWideChartoUTF8(argvW[i]);
     }
-    free(argv);
+    return argv;
 }
 
-static char* ConvertWideChartoUTF8(const wchar_t* src)
+static std::string ConvertWideChartoUTF8(const wchar_t* src)
```
duncanspumpkin, line 58: How is this different to https://github.com/OpenRCT2/OpenRCT2/blob/9ef8d6da421399e69a687bcaf3a596702ce27009/src/openrct2/core/String.cpp#L72

Can't we use that function.

src/openrct2-win/openrct2-win.cpp
```diff
@@ -17,64 +17,50 @@
...
-            argv[i] = ConvertWideChartoUTF8(argvW[i]);
-        }
-    }
-    return argv;
-}
-
-static void FreeCommandLineArgs(int argc, char** argv)
-{
-    // Free argv
+    std::vector<std::string> argv(argc);
```
duncanspumpkin, line 50: I know this is more efficient to give it the size first but we might as well stick to normal behaviour as this is not exactly critical code. That would be constructing empty and using push back instead of accessing the individual elements.

src/openrct2-win/openrct2-win.cpp
```diff
@@ -17,64 +17,50 @@
...
 
     SetConsoleCP(CP_UTF8);
     SetConsoleOutputCP(CP_UTF8);
-    auto exitCode = NormalisedMain(argc, const_cast<const char**>(argv));
 
-    FreeCommandLineArgs(argc, argv);
+    std::vector<const char*> argv;
+    std::transform(
+        argvStrings.begin(), argvStrings.end(), std::back_inserter(argv), [](const auto& string) { return string.c_str(); });
+    auto exitCode = NormalisedMain(argc, argv.data());
```
duncanspumpkin, line 43: Future refactoring would be to use an vector of strings in normalisedmain

PR #12077, opened 2020-06-28
Part of #11437: Move PlatformFileExists to Platform2.h

frutiemax:


src/openrct2/platform/Shared.cpp
```diff
@@ -90,6 +91,12 @@ namespace Platform
         outTime.second = localTime->tm_sec;
         return outTime;
     }
+
+    bool FileExists(const std::string path)
+    {
+        std::ifstream file(path.c_str());
```
duncanspumpkin, line 97: Can't you use std::filesystem::exists

frutiemax:
```
/Users/runner/runners/2.263.0/work/OpenRCT2/OpenRCT2/src/openrct2/platform/Shared.cpp:109:13: error: 'path' is unavailable: introduced in macOS 10.15
        fs::path file = path;
```

frutiemax:
I have some problems making std::filesystem works on every platform.

src/openrct2/platform/Shared.cpp
```diff
@@ -29,6 +29,14 @@
 #include "Platform2.h"
 
 #include <algorithm>
+
+#if ((defined(_WIN32) || defined(_WIN64) || defined(__APPLE__)))
```
duncanspumpkin, line 33: We have a core header just for this purpose. Look at `core/FileSystem.hpp`.

duncanspumpkin:
Forgot to mention we have a core header that handles filesystem and for compilers that don't yet support the standards correctly.

PR #12082, opened 2020-06-29
Fix #12079: incorrect drawing of attached paint structs

### LINKED ISSUE #12079: Fully or partly missing grid while building or removing scenery
OpenRCT2, v0.2.6-1004-g972c28619 (972c286 on develop) provided by GitHub
Windows (x86-64)

When placing rides with scenery or when lowering land or removing scenery the grid vanishes partly or completely.
This hasn't been just since last version, but a few versions before.

In the screenshots I used the Clear Scenery tool, increased its size to 12 and just placed the cursor in the middle of the screen. As you can see, as I just turned the view the grid vanishes.

Forest Frontiers  120-05-29 10-56-13
Forest Frontiers  120-05-29 10-56-18
Forest Frontiers  120-05-29 10-56-21
Forest Frontiers  120-05-29 10-56-24

duncanspumpkin:
Fix #12079: incorrect drawing of attached paint structs
The issue is caused by the paint_struct x and y coordinates being incorrectly marked as unsigned. Unfortunately they cannot easily be converted without careful rework. Therefore this is a stop gap until the type is changed to preferably a ScreenCoords.

### DIFF
--- a/src/openrct2/paint/Paint.cpp
+++ b/src/openrct2/paint/Paint.cpp
@@ -522,7 +522,8 @@ static void paint_attached_ps(rct_drawpixelinfo* dpi, paint_struct* ps, uint32_t
     attached_paint_struct* attached_ps = ps->attached_ps;
     for (; attached_ps; attached_ps = attached_ps->next)
     {
-        auto screenCoords = ScreenCoordsXY{ attached_ps->x + static_cast<int16_t>(ps->x), attached_ps->y + static_cast<int16_t>(ps->y) };
+        auto screenCoords = ScreenCoordsXY{ attached_ps->x + static_cast<int16_t>(ps->x),
+                                            attached_ps->y + static_cast<int16_t>(ps->y) };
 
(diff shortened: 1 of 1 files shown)


src/openrct2/paint/Paint.cpp
```diff
@@ -522,7 +522,7 @@ static void paint_attached_ps(rct_drawpixelinfo* dpi, paint_struct* ps, uint32_t
     attached_paint_struct* attached_ps = ps->attached_ps;
     for (; attached_ps; attached_ps = attached_ps->next)
     {
-        auto screenCoords = ScreenCoordsXY{ attached_ps->x + ps->x, attached_ps->y + ps->y };
+        auto screenCoords = ScreenCoordsXY{ attached_ps->x + static_cast<int16_t>(ps->x), attached_ps->y + static_cast<int16_t>(ps->y) };
```
tupaschoal, line 525:
```suggestion
        auto screenCoords = ScreenCoordsXY{ attached_ps->x + static_cast<int16_t>(ps->x),
                                            attached_ps->y + static_cast<int16_t>(ps->y) };
```

PR #12088, opened 2020-06-30
Generic coords structs for reusability

tupaschoal:
This is just making `MapRange` more useful by allowing multiple combinations of ranges.

It is strict with regards to each of its inheritors, such as `ScreenRect`, `MapRange`, `ScreenLine`

This will make it possible to convert functions like:
```c++
void gfx_draw_line(rct_drawpixelinfo* dpi, int32_t x1, int32_t y1, int32_t x2, int32_t y2, int32_t colour);
void gfx_fill_rect(rct_drawpixelinfo* dpi, int32_t left, int32_t top, int32_t right, int32_t bottom, int32_t colour);
```

into

```c++
void gfx_draw_line(rct_drawpixelinfo* dpi, const ScreenLine& line, int32_t colour);
void gfx_fill_rect(rct_drawpixelinfo* dpi, const ScreenRect& rect, int32_t colour);
```

### DIFF
--- a/src/openrct2-ui/interface/Graph.cpp
+++ b/src/openrct2-ui/interface/Graph.cpp
@@ -189,41 +189,6 @@ struct FinancialTooltipInfo
     const money32 money{};
 };
 
-struct ScreenRect
-{
-    const ScreenCoordsXY LeftTop;
-    const ScreenCoordsXY RightBottom;
-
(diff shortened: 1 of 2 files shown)


IntelOrca:
Seems fine to me.

PR #12094, opened 2020-07-02
Fix #12093. Incorrect position and animation of staff tab

### LINKED ISSUE #12093: Staff preview is broken
**OS:** Ubuntu 20.04
**Version:** 0.2.6+
**Commit/Build:** 8b6cb7e94

The image is in the corner instead of in the centre, and is not animated.

afbeelding

duncanspumpkin:
Fix #12093. Incorrect position and animation of staff tab.

Mistake made when refactoring transposed two digits. Additional mistake made when removing a variable that was unused that changed the memory layout of the window. This meant that a part of the code that accessed a different variable in a union fetched the wrong memory. This was a mistake ultimately made during implementation.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -41,6 +41,7 @@
 - Fix: [#11953] Incorrect banner text shade colour on wall text.
 - Fix: [#12062] Inconsistent lift hill sprites on Flying Coaster while inverted (original bug).
 - Fix: [#12068] Incorrect Entrance/Exit location on track design preview. Incorrect track design previews with track that contain diagonal track elements.
+- Fix: [#12093] Staff window tab animation was no longer updating.
 - Fix: RCT1 scenarios have more items in the object list than are present in the park or the research list.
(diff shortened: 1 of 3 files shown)


distribution/changelog.txt
```diff
@@ -41,6 +41,7 @@
 - Fix: [#11953] Incorrect banner text shade colour on wall text.
 - Fix: [#12062] Inconsistent lift hill sprites on Flying Coaster while inverted (original bug).
 - Fix: [#12068] Incorrect Entrance/Exit location on track design preview. Incorrect track design previews with track that contain diagonal track elements.
+- Fix: [#12093] Staff window tab animation was no longer updating.
```
Gymnasiast, line 44: Since this was not present in release 0.2.6, it shouldn't get a changelog entry. I'll fix it.

duncanspumpkin: It was as far as i could tell. Only the staff offset was recent

Gymnasiast: OK, re-added it.

PR #12102, opened 2020-07-02
Close #12087: Use ScreenCoordsXY on gfx_filter_pixel

### LINKED ISSUE #12087: Use ScreenCoordsXY on gfx_filter_pixel
- Create an overload of `gfx_filter_pixel` with the same name, but that takes a `const ScreenCoordsXY&` object instead of an `int32_t x` and and an `int32_t y` 

annoyingdroid:
This is my first pull request, so my apologies for any errors. 

Modified `gfx_filter_pixel` to now use `const ScreenCoordsXY&`, and changed calls to the old function to reference the new one (there are only 4 calls to the function).

### DIFF
--- a/src/openrct2-ui/windows/Tooltip.cpp
+++ b/src/openrct2-ui/windows/Tooltip.cpp
@@ -203,10 +203,10 @@ static void window_tooltip_paint(rct_window* w, rct_drawpixelinfo* dpi)
     gfx_filter_rect(dpi, left + 2, top + 0, right - 2, top + 0, PALETTE_DARKEN_3);
 
     // Corners
-    gfx_filter_pixel(dpi, ScreenCoordsXY{ left + 1, top + 1 }, PALETTE_DARKEN_3);
-    gfx_filter_pixel(dpi, ScreenCoordsXY{ right - 1, top + 1 }, PALETTE_DARKEN_3);
-    gfx_filter_pixel(dpi, ScreenCoordsXY{ left + 1, bottom - 1 }, PALETTE_DARKEN_3);
-    gfx_filter_pixel(dpi, ScreenCoordsXY{ right - 1, bottom - 1 }, PALETTE_DARKEN_3);
+    gfx_filter_pixel(dpi, { left + 1, top + 1 }, PALETTE_DARKEN_3);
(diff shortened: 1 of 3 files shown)


annoyingdroid:
Got it. With the rebase, is it necessary that I force the push after rebasing? I'm running into an issue where the push fails because the branch tip is behind the remote counterpart.

tupaschoal:
Yes

src/openrct2-ui/windows/Tooltip.cpp
```diff
@@ -203,10 +203,10 @@ static void window_tooltip_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
 
     // Corners
-    gfx_filter_pixel(dpi, left + 1, top + 1, PALETTE_DARKEN_3);
-    gfx_filter_pixel(dpi, right - 1, top + 1, PALETTE_DARKEN_3);
-    gfx_filter_pixel(dpi, left + 1, bottom - 1, PALETTE_DARKEN_3);
-    gfx_filter_pixel(dpi, right - 1, bottom - 1, PALETTE_DARKEN_3);
+    gfx_filter_pixel(dpi, ScreenCoordsXY{ left + 1, top + 1 }, PALETTE_DARKEN_3);
+    gfx_filter_pixel(dpi, ScreenCoordsXY{ right - 1, top + 1 }, PALETTE_DARKEN_3);
+    gfx_filter_pixel(dpi, ScreenCoordsXY{ left + 1, bottom - 1 }, PALETTE_DARKEN_3);
+    gfx_filter_pixel(dpi, ScreenCoordsXY{ right - 1, bottom - 1 }, PALETTE_DARKEN_3);
```
tupaschoal, line 209:
```suggestion
    gfx_filter_pixel(dpi, { left + 1, top + 1 }, PALETTE_DARKEN_3);
    gfx_filter_pixel(dpi, { right - 1, top + 1 }, PALETTE_DARKEN_3);
    gfx_filter_pixel(dpi, { left + 1, bottom - 1 }, PALETTE_DARKEN_3);
    gfx_filter_pixel(dpi, { right - 1, bottom - 1 }, PALETTE_DARKEN_3);
```
Compiler can infer the type, so we usually do it this way for shortness

annoyingdroid: Good to know! Let me make those changes.

PR #12104, opened 2020-07-02
Close #12086: Use ScreenCoordsXY on gfx_draw_pixel

### LINKED ISSUE #12086: Use ScreenCoordsXY on gfx_draw_pixel
- Create an overload of `gfx_draw_pixel` with the same name, but that takes a `const ScreenCoordsXY&` object instead of an `int32_t x` and and an `int32_t y` 

annoyingdroid:
Refactored `gfx_draw_pixel` to use `const ScreenCoordsXY&`, though I wasn't able to find any other references to the function aside from the two I modified. Is the function being used?

### DIFF
--- a/src/openrct2/drawing/Drawing.cpp
+++ b/src/openrct2/drawing/Drawing.cpp
@@ -578,9 +578,9 @@ void mask_init()
     }
 }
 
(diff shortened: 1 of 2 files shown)


annoyingdroid:
Ahhhhh, got it. I figured you had meant just a single branch for changes. Thanks for the clarification!

Issue #9542, opened 2019-07-08
CI builds should depend on clang-format result


ZehMatt:
We can avoid a lot of unnecessary building if the build process depends on the clang-format stage. This should also simplify the process of people trying to fix the format by getting faster CI responses.

duncanspumpkin:
But wont that mean we need to fix the formatting before we fix the compilation of the code?

ZehMatt:
Pretty much.

janisozaur:
I would also make clang-format not download any docker images, but use a (mostly?) static binary instead.

tupaschoal:
This is already the case

PR #12108, opened 2020-07-02
Part of #9542: CI builds only start after successful lint-commit and check-code-formatting jobs

caseyjohnsonwv:
I noticed that #9542 was still open from almost a year ago. A few months after that issue was opened, GitHub Actions added job-level conditionals, making that change much easier to implement. If `lint-commit` or `check-code-formatting` fails (or is cancelled), the 10 subsequent build jobs are skipped. As suggested in the issue, this will save build resources while giving faster feedback on PRs that fail CI due to formatting.

PR is titled "Part of -" because @janisozaur mentioned switching clang-format to a static binary rather than a docker image. I think that's a great idea, but probably beyond the scope of my permissions.


IntelOrca:
Doesn't this delay the other jobs from starting until the format check is done, increasing the total CI time?

duncanspumpkin:
They are both very quick.

PR #12110, opened 2020-07-02
Add "Hybrid Coaster" ride type

X123M3-256:
This PR adds a new ride type "Hybrid Coaster", which is an RMC IBox style coaster. In addition, it adds some extra wooden support sprites for large flat to steep transitions. The ride type includes all sloped and banked pieces and transitions (except for the single tile steep-to-flat), plus barrel rolls, quarter loops, and launched lifts. Booster sprites are included but only accessible when "Show all drawable track pieces" is checked.

rmcscreen

Since no vehicle is currently available for this ride type, arbitrary ride types must be enabled in order to test the ride type. I recommend testing with a default vehicle such as the LIM trains instead of the custom RMC train, because the latter is quite glitchy and it's hard to tell what's an issue with the track vs the train.

The following track pieces are known to glitch, but these glitches are present on the vanilla tracks and are likely to be difficult or impossible to fix:

* Small unbanked sloped turn

* Steep to vertical

* Medium banked turns, when fences are built directly adjacent.

Some things I think need looking at:

* The code formatting. I did try running `clang-format` as Janisozaur suggested on the Discord, but I didn't run it on every file I modified, only those that were newly added, in case it would mess up existing formatting, so there's probably still issues.

* The ride rating function. It was bothering me that the giga coaster would always get slightly higher ratings than this even on a very un-hyper like layout, so I decided to increase the excitement from turns and decrease that from drops. This had the intended effect, but the difference between the two ride types is still small, and looking at the existing tracks, most rides that get extra excitement from turns are things like wild mouse and spinning rides. So it might be worth reverting this change so the ride rating function more closely follows the pattern of the existing ride types.

* The build cost - I basically looked at the original values and picked something that seemed suitable without ever actually testing. I don't know how many people would actually use this ride type in scenario mode so it may not be that important.

* The booster acceleration - it's currently set a bit less than the twister coaster because I thought the twister boosters looked very strong, but perhaps it should be the set the same as most other rides are.

* The lift hill sound effect. I tested all of them and none of them really sound right, but I wasn't sure which one was best.

* The overall look of the sprites, although I do believe that further improvement in this area is likely to be very difficult.


duncanspumpkin:
How many mega bytes is all of the sprites? I feel we need to break this out into an object and keep it in the object repo as this will permanently increase the main repo size forever. Annoyingly because of all the sprites GitHub can't handle this pr correctly.

X123M3-256:
@duncanspumpkin According to my file browser there are 815 sprites totalling 1.6MiB.

I think adding support for track types as object files would be a major project in itself and should really be a separate PR.

Gymnasiast:
@duncanspumpkin Turning ride types into objects is a big project in itself, though the RTD moves have done a lot of preparation in that area. I don't think we should do that now.

src/openrct2/ride/Ride.cpp
```diff
@@ -7181,7 +7180,6 @@ RideClassification Ride::GetClassification() const
         case RIDE_TYPE_1F:
         case RIDE_TYPE_SHOP:
         case RIDE_TYPE_22:
-        case RIDE_TYPE_50:
```
Gymnasiast, line 7184: Revert this change and the one on 4353.

src/openrct2/localisation/StringIds.h
```diff
@@ -191,14 +192,14 @@ enum
     STR_RIDE_DESCRIPTION_MAGIC_CARPET = 589,
     STR_RIDE_DESCRIPTION_SUBMARINE_RIDE = 590,
     STR_RIDE_DESCRIPTION_RIVER_RAFTS = 591,
-
```
Gymnasiast, line 194: Revert.

src/openrct2/paint/Supports.cpp
```diff
@@ -473,10 +521,15 @@ bool wooden_a_supports_paint_setup(
     {
         special = (special - 1) & 0xFFFF;
 
-        int32_t imageId = WoodenCurveSupportImageIds[supportType];
+        int32_t imageId;
+        if (special <= 48)
+            imageId = WoodenCurveSupportImageIds[supportType];
+        else
+            imageId = WoodenCurveSupportImageIdsG2[supportType];
+
```
Gymnasiast, line 529: Could you explain what you did in this file?
And what is the meaning of 48/49?

src/openrct2/ride/Ride.cpp
X123M3-256: That's odd, when I look at that file the reference to RIDE_TYPE_50 isn't there.

src/openrct2/paint/Supports.cpp
X123M3-256: The original code assumes that all the support sprites have contiguous sprite IDs. But since the old sprites are in g1.dat and the sprites I added are in g2.dat, they cannot be contiguous. So I added a separate array to hold the additional sprites. 48 is, IIRC, the number of original support sprites - the idea is that support sprite number 49 maps to support sprite 0 in the g2.dat array. Since there's no mechanism to add new sprites to g1.dat, and even if there was it would change the locations of every subsequent sprite, I don't think there's a way around this even if it's a bit dirty.

Gymnasiast: Right, but what kind of support sprites have you added?

X123M3-256: I added support sprites for large flat to steep and steep to flat transitions. I did this because the base game doesn't have support sprites for those track pieces but I wanted them to be available on this one. The sprites I added are in resources/g2/support.

src/openrct2/ride/Ride.cpp
Gymnasiast: A stray whitespace change.

I'll take care of it.

X123M3-256: Oh, RIDE_TYPE_50 was removed not added. This should be fixed now.

X123M3-256:
I had to rebase again and force-push the branch because stuff has changed upstream - hopefully this doesn't break stuff.

X123M3-256:
Should I increase the network version for this? Trying to open a park using the new track in a version without it segfaults the game.

tupaschoal:
That sounds like another problem. The network version is just a counter to make sure all clients joining a server have compatible builds, as far as I'm aware

duncanspumpkin:
Yeah we would need to increment the network version. We also need to add a little warning that the ride type will have backwards compatibility issues.

X123M3-256:
@tupaschoal I know, I'm asking if it should be increased because this change is not compatible with previous builds. Or should that not be part of the PR?

Gymnasiast:
Yeah, you should bump the network version. Could you also add a changelog entry? `- Feature: [#12110] Add "Hybrid Coaster" ride type.`

X123M3-256:
@Gymnasiast OK, I'll do that. To be honest I think submitting this now was a bit premature, I think the sprites definitely need more work to look right, so I will probably revisit it at the weekend.

Gymnasiast:
@X123M3-256 OK, I converted the PR to "draft PR" to mark that. Once you're ready, please mark it as "ready for review" again.

src/openrct2/paint/Supports.cpp
duncanspumpkin: It would have been better to change it so that instead of using contiguous sprite id's it used a table lookup. That would prevent the additional branch and allow for future remapping. Leave as is for now but keep that in mind in future.

src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -0,0 +1,11617 @@
...
+static void hybrid_rc_track_flat(
+    paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
+    const TileElement* tileElement)
+{
+    if (tileElement->AsTrack()->HasChain())
+    {
+        switch (direction)
+        {
+            case 0:
+                sub_98197C_rotated(
```
duncanspumpkin, line 45: `sub_98197C_rotated` was designed to reduce this duplication on rotation.
Just do:
```
sub_98197C_rotated( session, direction, hybrid_rc_get_track_colour(session) | (SPR_G2_HYBRID_LIFT_TRACK_FLAT + direction), 0, 0, 32, 20, 3, height, 0, 6, height);
```
drop all of the switch.

src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -0,0 +1,11617 @@
...
+                break;
+        }
+    }
+    else
+    {
+        switch (direction)
+        {
+            case 0:
+            case 2:
+                sub_98197C_rotated(
```
duncanspumpkin, line 72: as above but with `SPR_G2_HYBRID_TRACK_FLAT + (direction & 1)`

src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -0,0 +1,11617 @@
...
+    paint_util_set_general_support_height(session, height + 32, 0x20);
+}
+
+static void hybrid_rc_track_25_deg_up(
+    paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
+    const TileElement* tileElement)
+{
+    if (tileElement->AsTrack()->HasChain())
+    {
+        switch (direction)
```
duncanspumpkin, line 131: as above

src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -0,0 +1,11617 @@
...
+                    20, 2, height, 0, 6, height + 3);
+                break;
+        }
+    }
+    else
+    {
+        switch (direction)
+        {
+            case 0:
+                sub_98197C_rotated(
```
duncanspumpkin, line 160: as above

src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -0,0 +1,11617 @@
...
+    }
+    paint_util_set_segment_support_height(session, paint_util_rotate_segments(SEGMENTS_ALL, direction), 0xFFFF, 0);
+    paint_util_set_general_support_height(session, height + 104, 0x20);
+}
+
+static void hybrid_rc_track_flat_to_25_deg_up(
+    paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
+    const TileElement* tileElement)
+{
+    if (tileElement->AsTrack()->HasChain())
```
duncanspumpkin, line 269: as above

resources/g2/sprites.json
```diff
@@ -2175,5 +2175,5049 @@
         "y_offset": -1,
         "palette": "keep",
         "forceBmp": true
+    },
```
Gymnasiast, line 2178: Please always keep the font sprites at the end. Not doing so risks running into problems later.

src/openrct2/paint/Supports.cpp
```diff
@@ -230,37 +246,39 @@ struct unk_supports_desc {
...
     {{{0,  0,  0}, {1,  1,  8}},  0, 1},
+    {{{0,  0,  0}, {1,  1,  8}},  0, 1}, // Gentle to flat
     {{{0,  0,  0}, {1,  1,  8}},  0, 1},
     {{{0,  0,  0}, {1,  1,  8}},  0, 1},
     {{{0,  0,  0}, {1,  1,  8}},  0, 1},
+    {{{0,  0,  0}, {1,  1,  8}},  0, 1}, // Gentle slope
     {{{0,  0,  0}, {1,  1,  8}},  0, 1},
     {{{0,  0,  0}, {1,  1,  8}},  0, 1},
     {{{0,  0,  0}, {1,  1,  8}},  0, 1},
-    {{{0,  0,  0}, {1,  1,  8}},  0, 1},
```
Gymnasiast, line 242: I am not too sure about this. We're deviating from vanilla here, and I'm not sure why, or whether it will cause unforeseen problems with supports drawing.

X123M3-256: The behaviour of the existing supports should not have been modified. I'll look over the file again, I might have changed something accidentally.

X123M3-256: @Gymnasiast Can you clarify exactly what you think deviates from vanilla? I don't see any changes to the content of that array except for the extra entries I added.

Gymnasiast: You inserted several lines in-between.

X123M3-256: Where? I compared the original version currently on develop and the version from this branch side by side and can't find any changes to the existing lines, other than adding some comments to mark which support each line belongs to (although I could remove these if you want).

Gymnasiast: I have now checked locally. You are correct, it's just the way the diff is presented.

Gymnasiast (approved):

janisozaur (changes requested):
The images need to be optimised

janisozaur (changes requested):
The images still are far too large. Use zopflipng to optimise them.

src/openrct2/paint/Supports.cpp
```diff
@@ -653,12 +706,18 @@ bool wooden_b_supports_paint_setup(
     {
         uint16_t specialIndex = (special - 1) & 0xFFFF;
 
-        uint32_t imageId = WoodenCurveSupportImageIds[supportType];
+        int32_t imageId;
+        if (specialIndex < NumVanillaWoodenSupportTypes)
+            imageId = WoodenCurveSupportImageIds[supportType];
```
duncanspumpkin, line 711: I don't like this exposing the g2 compared to normal elements. I'll submit a change that adjusts it

AaronVanGeffen:
This arguing over what PNG cruncher to use is a bit counterproductive, I think. It's been my experience that some PNG crunching may even lead to OpenRCT2 no longer loading the file properly when building `g2.dat`, so we need to be careful with it. To me, having standard, tested PNGs is much more important than shaving off a few more bytes.

janisozaur:
> may even lead to OpenRCT2 no longer loading the file properly

Haven't ever seen this in any of the projects where zopfli was used, none of which used `gAMA` or `pHYs`, but willing to get proven wrong.

My request still stands as there are quite a few images in this PR - the `g2.dat` using current and `zopfli`'d images will tell if any of those are wrong.

AaronVanGeffen:
@janisozaur I last dealt with the problem when optimising the PNGs for #9154. Anecdotal, to be sure, but I am not going to pour more energy in arguing the point.

src/openrct2/paint/Supports.cpp
X123M3-256: @duncanspumpkin Can you explain what you want changed? I can't add sprites to g1.dat so that's why i have this awkward conditional; the game expects all the support sprites to be in a contiguous block.

duncanspumpkin: Yes so I want to change it so that the game doesn't expect it to be in a contiguous block

duncanspumpkin:
I'm proposing this change https://github.com/duncanspumpkin/OpenRCT2/commit/3b9774ecc802d21d687b48d292b6626d10ac7fd1 to make the supports not contiguous. We should have really done it ages ago.

duncanspumpkin:
The function naming is consistent with the other rides but it doesn't really follow the coding style. I'd prefer we rename the functions to TitleCase. Ideally they should be in a namespace to simplify the names further.

duncanspumpkin:
I'd like to see an "official" object to be added to the object repo with this pr as well so people can use it immediately without cheats.

duncanspumpkin:
https://github.com/duncanspumpkin/OpenRCT2/commit/86ecb4dae05253d6c6107c47222fb0f4b0490c98 and here is a commit renaming all the functions to TitleCase and adding a namespace.

duncanspumpkin (approved):

AaronVanGeffen (commented):
It would be great to have an 'OpenRCT2 official' RMC vehicle added to the objects repo, so players can use this track without cheats. However, I appreciate this is outside of the scope of this PR. From the OP, it sounds like one may already be in the works, though? :slightly_smiling_face: 

The network version should be incremented before this is merged.

X123M3-256:
@duncanspumpkin Did you add those changes directly to this branch or do I need to do something to integrate them?

duncanspumpkin:
@X123M3-256 yes commited them to your branch after a little discussion on discord. No changes needed from yourself.

AaronVanGeffen (commented):
I'm getting surprisingly good recompression results from `zopflipng`, as recommended by @janisozaur — between 20% to 80% off. I'll test the recompressed sprites and push them when ready.

AaronVanGeffen:
Before recompression:
Amity Airfield 2020-10-10 16-31-43

After recompression:
Amity Airfield 2020-10-10 16-52-29

@janisozaur This is what I was afraid of…

AaronVanGeffen (approved):
Good to go, as far as I can tell.

X123M3-256:
@AaronVanGeffen So does the bug with the images still need fixing?

duncanspumpkin:
No it's all fine to go.

PR #12111, opened 2020-07-02
Part of #11159: util_zlib_deflate returns a optional<vector<uint8_t>>

frutiemax:


src/openrct2/network/Network.cpp
```diff
@@ -1625,9 +1625,10 @@ uint8_t* Network::save_for_network(size_t& out_size, const std::vector<const Obj
     const void* data = ms.GetData();
     int32_t size = ms.GetLength();
 
-    uint8_t* compressed = util_zlib_deflate(static_cast<const uint8_t*>(data), size, &out_size);
-    if (compressed != nullptr)
+    auto compressed = util_zlib_deflate(static_cast<const uint8_t*>(data), size);
+    if (compressed != std::nullopt)
```
tupaschoal, line 1629: Can't we just return the vector and just use `.empty()`?

```suggestion
    auto compressed = util_zlib_deflate(static_cast<const uint8_t*>(data), size);
    if (!compressed.empty())
```

duncanspumpkin: I don't think that is a good idea. What happens if we compress zero data?

tupaschoal: Can't we just not call it if `size == 0`?

duncanspumpkin: Std optional is meant to be zero overhead so there should be no penalty using it. It makes it clear that this function has a failure state that must be checked.

tupaschoal: Alright, I just wanted to be sure we were not overusing it.

PR #12114, opened 2020-07-02
Close #12112: Move TrackColourPreviews to RideTypeDescriptors

### LINKED ISSUE #12112: Move TrackColourPreviews to RideTypeDescriptors

frutiemax:
The query of RideCreateAction has been modified so that a constant reference to a colour preset instead of a pointer is used. This caused the query to throw a Invalid Ride Type error.


src/openrct2/ride/coaster/meta/AirPoweredVerticalCoaster.h
```diff
@@ -47,6 +48,9 @@ constexpr const RideTypeDescriptor AirPoweredVerticalCoasterRTD =
         { COLOUR_BRIGHT_RED, COLOUR_LIGHT_BLUE, COLOUR_WHITE },
         { COLOUR_LIGHT_ORANGE, COLOUR_BRIGHT_RED, COLOUR_BRIGHT_YELLOW },
     )),
+    SET_FIELD(ColourPreview, { SPR_RIDE_DESIGN_PREVIEW_AIR_POWERED_VERTICAL_COASTER_TRACK,
```
Gymnasiast, line 51: Could you put these on one line for all RTDs?

src/openrct2/ride/RideData.h
```diff
@@ -333,6 +340,7 @@ constexpr const RideTypeDescriptor DummyRTD =
     SET_FIELD(PhotoItem, SHOP_ITEM_PHOTO),
     SET_FIELD(BonusValue, 0),
     SET_FIELD(ColourPresets, DEFAULT_FLAT_RIDE_COLOUR_PRESET),
+    SET_FIELD(ColourPreview, {0, 0 }),
```
Gymnasiast, line 343:
```suggestion
    SET_FIELD(ColourPreview, { SPR_NONE, SPR_NONE }),
```

frutiemax: @Gymnasiast putting SPR_NONE does not work straight forward since it is equal to -1 and it's expecting a uint32_t.

PR #12124, opened 2020-07-04
Tiny GetEntity Refactor

duncanspumpkin:
Use default template parameter instead of specialising.

I've a feeling this might trigger a few additional nullptr warnings. This is somewhat a test.


duncanspumpkin:
Its interesting that it did indeed trigger a ton of additional nullptr deref warnings.

tupaschoal:
Seems ok, there are a bunch of null derefs still to fix

duncanspumpkin:
Will need a rebase that actually simplifies a couple of the nullderef issues.

PR #12127, opened 2020-07-04
Clarify CoordsRange-derived interface methods

tupaschoal:
When I  turned `MapRange` into `CoordsRange`, I kept the names `leftTop` and `rightBottom`, but these don't make sense from the `ScreeLine` point of view, as the point don't really need to follow this convention. In this PR:

- `LeftTop` and `RightBottom` are renamed to `PointOne` and `PointTwo` in the base class.
- Base class defines getters for the generic point variables
- `GeLeft()`, `GetRight()`, `GetTop()` and `GetBottom()` are redefined using these new generic getters.

As a side effect of changing the names of the variables, I also had to change `DataSeraliserTraits` to use the getters, which I think is a plus :D

### DIFF
--- a/src/openrct2/world/Location.hpp
+++ b/src/openrct2/world/Location.hpp
@@ -614,24 +614,24 @@ struct TileCoordsXYZD : public TileCoordsXYZ
  */
 template<class T> struct CoordsRange
 {
-    T PointOne{ 0, 0 };
-    T PointTwo{ 0, 0 };
+    T Point1{ 0, 0 };
+    T Point2{ 0, 0 };
 
-    int32_t GetXOne() const
+    int32_t GetX1() const
     {
-        return PointOne.x;
+        return Point1.x;
     }
-    int32_t GetYOne() const
+    int32_t GetY1() const
     {
-        return PointOne.y;
+        return Point1.y;
     }
-    int32_t GetXTwo() const
+    int32_t GetX2() const
     {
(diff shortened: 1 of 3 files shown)


src/openrct2/world/Location.hpp
```diff
@@ -614,35 +614,35 @@ struct TileCoordsXYZD : public TileCoordsXYZ
  */
 template<class T> struct CoordsRange
 {
-    T LeftTop{ 0, 0 };
-    T RightBottom{ 0, 0 };
+    T PointOne{ 0, 0 };
+    T PointTwo{ 0, 0 };
 
-    int32_t GetLeft() const
+    int32_t GetXOne() const
```
Gymnasiast, line 620:
```suggestion
    int32_t GetX1() const
```

and the same for the other three.

src/openrct2/world/Location.hpp
```diff
@@ -614,35 +614,35 @@ struct TileCoordsXYZD : public TileCoordsXYZ
  */
 template<class T> struct CoordsRange
 {
-    T LeftTop{ 0, 0 };
-    T RightBottom{ 0, 0 };
+    T PointOne{ 0, 0 };
```
Gymnasiast, line 617:
```suggestion
    T Point1{ 0, 0 };
```

tupaschoal:
Renamed them, do squash on merge

PR #12131, opened 2020-07-05
Part of #12100: Use ScreenLine on gfx_draw_line_software

arnimir:
* created overload
* old function calls new one
* replaced call in X8DrawEngine.cpp


duncanspumpkin:
@arnimir did something go wrong?

Issue #1999, opened 2015-10-02
Staff Ignore Queue Paths


FrostyJTD:
I posted this over in the OpenRCT2 community forum, but I figured I'd throw it up here as well.

Staff have always entered queue lines, since the first game, and it's always been annoying. There's no reason for them to be in there. I'd love it if I could prevent them from doing so. Maybe as an option or a cheat?

PFCKrutonium:
I could see this being an option, but it should definitely treat a queue connected to a ride different to a random piece of queue that isn't  - I've seen people use queue for paths before.

FrostyJTD:
I thought about that too, but it's a simple check. An actual queue has the banner at the end of it, marking where the queue starts. Guests approach it before deciding if they want to go on a ride, so there's obviously something there. Just have staff not cross over that queue marker, and then queue path used as regular path will be treated as normal for staff.

Patrik356b:
Maybe it's for janitors, if guests can litter/vomit in the ques. Never seen but that would explain it.

PFCKrutonium:
I had a guest stuck in a puke loop in a queue, and while the guest was puking was the only time I have ever seen vomit in a queue.

FrostyJTD:
guests don't litter nor puke in queues, under normal circumstances.

IntelOrca:
Mechanics need to use queue lines to get to the entrance of a ride that has a station with only an entrance and not an exit.

PFCKrutonium:
Could we detect that case and only allow them up the queue when it is true?

duncanspumpkin:
The thing is entertainers in lines makes sense. Handymen in queues are occasionally required when using a cheat. Mechanics in queues are occasionally required. Guards are never required. I think it would just add a bunch of code for little improvement. I would prefer if we improved the handyman code so that they checked the queue for litter before entering it.

IntelOrca:
Sure, although for the record I don't know if entertainers would enter the actual queue lines in real life. They also clip through all the peeps.

FrostyJTD:
[...]

PR #12139, opened 2020-07-06
Make handyman not enter queue lines with banners

Jimver:
Hey I implemented a feature which was requested in #1999. 

It does the following:
- Handymen won't go from a normal path to a queue path if there is a queue banner facing the handymen
- Handymen will still be able to exit queue lines should they somehow end up in one as I check the direction of the queue line banner.

In the picture below the handyman never goes into the queue line, I attached the save file below.
handyman-queue.

[sandbox-handyman.zip](https://github.com/OpenRCT2/OpenRCT2/files/4876081/sandbox-handyman.zip)

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -87,6 +87,7 @@ The following people are not part of the development team, but have been contrib
 * Chad Ian Anderson (pizza2004) - Added New Game option, bug fixes, misc.
 * Peter Ryszkiewicz (pRizz) - Added horizontal grid lines to finance charts.
 * Hudson Oliveira (hdpoliveira) - Misc.
+* Jim Verheijde (Jimver) - Prevent handymen from entering queue lines.
 
 ## Bug fixes
 * (halfbro)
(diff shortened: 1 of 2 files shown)


Jimver:
Will close because of CI failures

tupaschoal:
Please analyze the CI failure and fix on the same PR instead of opening new ones, whenever possible.

Jimver:
Yeah I didn't know how to change commit messages after they are made so I made a new branch with a shorter commit name

AaronVanGeffen:
You don't need to change the commit; you can just push another to the same branch. Please don't open new PRs for the same branch, in any case.

tupaschoal:
> Yeah I didn't know how to change commit messages after they are made so I made a new branch with a shorter commit name

You can use `git rebase -i develop` which will open the list of commits with the action prepended. Change `pick` to `reword` and rewrite the commit message as it pleases you. There are other ways too.

PR #12140, opened 2020-07-06
Make handymen not enter queue lines

Jimver:
Hey I implemented a feature which was requested in #1999. 

It does the following:
- Handymen won't go from a normal path to a queue path if there is a queue banner facing the handymen
- Handymen will still be able to exit queue lines should they somehow end up in one as I check the direction of the queue line banner.

In the picture below the handyman never goes into the queue line, I attached the save file below.
handyman-queue.

[sandbox-handyman.zip](https://github.com/OpenRCT2/OpenRCT2/files/4876081/sandbox-handyman.zip)

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -87,6 +87,7 @@ The following people are not part of the development team, but have been contrib
 * Chad Ian Anderson (pizza2004) - Added New Game option, bug fixes, misc.
 * Peter Ryszkiewicz (pRizz) - Added horizontal grid lines to finance charts.
 * Hudson Oliveira (hdpoliveira) - Misc.
+* Jim Verheijde (Jimver) - Prevent handymen from entering queue lines.
 
 ## Bug fixes
 * (halfbro)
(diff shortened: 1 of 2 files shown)


Jimver:
I will look at the clang formatting tomorrow

sta5is:
Unfortunately there are instances where handyman is necessary to go on queue paths. When a guest enters a queue and leaves for reasons like ride shut down or unhappy with the long wait, and if they happen to be sick or carrying trash, there's a chance they'll vomit or litter on the queue path as they are leaving.

James103:
> there's a chance they'll vomit or litter on the queue path as they are leaving.

Here is an example of that from 6c3c857:
image

These queue paths are actually dirty - when a handyman is placed there, they actually sweep the vomit/litter off of the queue path. This shows that handyman are necessary even in queue paths.

To reproduce:
1. Launch OpenRCT2 with the default OpenRCT2 title sequence.
2. Wait until the title sequence switches to another park (should be a Six Flags over Texas park).
3. `save_park a` then `load_park a` (or unzip and load [the attached park](https://github.com/OpenRCT2/OpenRCT2/files/4876821/a.zip)).
4. Go into the cheats menu and set the nausea value of all guests to the max possible.
5. Wait a bit.
6. Look around for any queue paths with litter or vomit on them.

duncanspumpkin:
Yeah as said above the reason this hasn't been implemented is litter can be deposited in a queue line.

tupaschoal:
I'm closing this PR due to what was said, sorry this can't be merged :\

PR #12142, opened 2020-07-06
Use forward declares and reduce header's dependencies on Location.hpp

### LINKED ISSUE #12118: Use forward declares and reduce header's dependencies on Location.hpp
Several headers include `Location.hpp` to use its `.*Coords.*` objects. This, however, triggers recompilation of lots of files whenever `Location.hpp` is changed, which is not ideal and could be avoided in many cases.

Use forward declaration to remove includes from header files where applicable, doing `struct ScreenCoordsXY`, for example, and moving the `#include ... Location.hpp` to the source file instead.

This is possible on header files that only declare a function receiving one of  the `.*Coords.*` objects as reference or pointer, such as seen in `Chat.h`.

Some can't be done, like in `ClearAction.hpp`, as it contains the declarations and definitions.

Some notes:
- Removing one include from a header file may require you to add includes in multiple source files. That's ok and expected, as transitive includes might have given the objects to lots of different files.
- Don't try to do it all at once if it feels like too much

thamara:
Prefer forward declaration instead of including the Location.hpp in header files.
Tested it by compiling :)

Closes #12118.

### DIFF
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -43,6 +43,7 @@
 #include <openrct2/title/TitleSequencePlayer.h>
 #include <openrct2/ui/UiContext.h>
 #include <openrct2/ui/WindowManager.h>
+#include <openrct2/world/Location.hpp>
 #include <vector>
 
 using namespace OpenRCT2;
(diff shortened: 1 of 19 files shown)


duncanspumpkin:
There is a tiny bit of work to be done to the news item queue to prevent the extra headers it includes but can be picked up at a later date.

Issue #3205, opened 2016-03-31
Handymen getting stuck in ride queues


LittleMikey:
**OS:** Windows 10 64 bit
**Version:** 0.0.4
**Commit:** 9ef4af5

Explanation of the issue...
After wondering why guests were still complaining about disgusting paths despite hiring about 15 handymen, I looked around the park and saw that there was a huge group of them stuck in the entrance tile to one of my rides. This isn't the first time I've seen this behaviour, I also had a similar problem on a different scenario, though it has never happened to this big a group before.

I've made sure that the handymen are not set to patrol that one tile, and they aren't

- [x] Reproducable in RCT2 (vanilla)?
- [ ] Multiplayer? Not tested

**Screenshots / Video:**
[http://i.imgur.com/iTonGwV.png]

**Save game:**
[Canyon Calamities.zip](https://github.com/OpenRCT2/OpenRCT2/files/197352/Canyon.Calamities.zip)

LRFLEW:
This is likely related to #3125 and #2499. I'm adding this to my list of things to look for.

IntelOrca:
Caused by litter being on adjacent paths to handymen.

LRFLEW:
@IntelOrca That sounds likely (and what was described by @Wirlie in #3125 ), but from my testing on that issue, it is not an original bug, so there is something that OpenRCT2 is handling differently/incorrectly.

duncanspumpkin:
This is an original bug. Handymen can't handle litter like this very well.

Patrik356b:
Only que paths that leads to a ride entrance should be ignored by staff as some people use que paths for artistic purposes

PR #12145, opened 2020-07-06
Increase handyman randomness in queue 

Jimver:
As mentioned in #3205 handymen can get stuck in queue lines when there is litter on an adjacent regular path.

This PR fixes this by increasing the chance that handymen will take a random direction instead of following the nearest litter **while in a queue**.

Previously the handyman will always walk the light blue highlighted line over and over. It can take around 30-40 cycles (according to my tests) until it chooses a random direction to break out of the cycle. (see save file below)
image

I increased this probability of choosing a (valid) random direction while a handyman is in a queue line such that it is much easier for them to break out of the cycle.

Right now I simply invert the probability when on a queue path `1-P(follow litter)`. Maybe there can be a more appropriate value for this. However it seems to work fine as the handyman now already breaks out of the cycle after only 0-2 cycles.

Of course it is not a proper fix but until the AI is reworked (as mentioned in https://github.com/OpenRCT2/OpenRCT2/issues/1999#issuecomment-654285482) this is the most simple thing I can think of which doesn't add too much complexity.

Attached save file below:
[Canyon Calamities-bug4.zip](https://github.com/OpenRCT2/OpenRCT2/files/4879970/Canyon.Calamities-bug4.zip)

### DIFF
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -643,7 +643,10 @@ bool Staff::DoHandymanPathFinding()
                 bool chooseRandom = true;
                 if (litterDirection != INVALID_DIRECTION && pathDirections & (1 << litterDirection))
                 {
-                    if ((scenario_rand() & 0xFFFF) >= 0x1999)
+                    /// When in a queue path make the probability of following litter much lower
+                    /// as handymen often get stuck when there is litter on a normal path next to a queue they are in
+                    uint32_t chooseRandomProbability = pathElement->IsQueue() ? (0xFFFF - 0x1999) : 0x1999;
+                    if ((scenario_rand() & 0xFFFF) >= chooseRandomProbability)
                     {
                         chooseRandom = false;
                         newDirection = litterDirection;
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -643,9 +643,9 @@ bool Staff::DoHandymanPathFinding()
                 bool chooseRandom = true;
                 if (litterDirection != INVALID_DIRECTION && pathDirections & (1 << litterDirection))
                 {
-                    /// When in a queue path make the probability of following litter much lower
+                    /// When in a queue path make the probability of following litter much lower (10% instead of 90%)
                     /// as handymen often get stuck when there is litter on a normal path next to a queue they are in
-                    uint32_t chooseRandomProbability = pathElement->IsQueue() ? (0xFFFF - 0x1999) : 0x1999;
+                    uint32_t chooseRandomProbability = pathElement->IsQueue() ? 0xE666 : 0x1999;
                     if ((scenario_rand() & 0xFFFF) >= chooseRandomProbability)
                     {
                         chooseRandom = false;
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
Note it is expected to fail the replay tests as it modifies state. Will require a network update and Replay recreation (by me).

Jimver:
> Will require a network update (by you)

Thanks for the note @duncanspumpkin! I just now updated this PR with the latest `develop` branch, is that what you mean by network update, or should I do something else?

duncanspumpkin:
First of all. Always rebase instead of merge. https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2

No sorry a network version increment is bumping up the network version in network.cpp. just put it up by one value. It means that if playing networked you need to update to the latest version.

src/openrct2/peep/Staff.cpp
```diff
@@ -643,7 +643,10 @@ bool Staff::DoHandymanPathFinding()
                 bool chooseRandom = true;
                 if (litterDirection != INVALID_DIRECTION && pathDirections & (1 << litterDirection))
                 {
-                    if ((scenario_rand() & 0xFFFF) >= 0x1999)
+                    /// When in a queue path make the probability of following litter much lower
+                    /// as handymen often get stuck when there is litter on a normal path next to a queue they are in
+                    uint32_t chooseRandomProbability = pathElement->IsQueue() ? (0xFFFF - 0x1999) : 0x1999;
```
duncanspumpkin, line 648: I think it would be better to check if it is a queue and it has a ride assigned to it. That way if someone has used a queue path for cosmetic reasons it would not be affected.

duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/network/Network.cpp#L35

src/openrct2/peep/Staff.cpp
duncanspumpkin, line 648: Also just use `0xE666` for the value. I've a branch that is refactoring these constants to be easier to understand. (0x1999 means 10%)

Jimver: Alright I will use `0XE666` for now

Jimver: I added a check with `GetRideIndex`

Jimver:
@duncanspumpkin I implemented your requested changes, is there anything else I should do such as adding an entry to the changelog?

duncanspumpkin:
yeah a changelog entry will be required for this as well. I'm waiting on a different PR until I fix the replays for this one.

duncanspumpkin:
Can you do a proper rebase of this ontop of develop and then i can create the replays. The other pr i mentioned has now been merged so it's safe for me to create the replays but i need this rebased first.

Jimver:
Sure! I'm not too familiar with rebasing though, is it still possible to do it on the same branch even if I did a merge before?

duncanspumpkin:
Yup it's always possible. It will also clean up and remove that merge commit. https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2 

You might hit a merge conflict on the network version as the other pr incremented it as well. You will need to bump the number up again.

Jimver:
Alright it should be good now!

duncanspumpkin:
👍

PR #12146, opened 2020-07-06
Fix #12144: Do not require lint commit to pass to run full CI

### LINKED ISSUE #12144: Commit message Lint shouldn't prevent build generation
When I introduced the commit message lint on the CI, I intended for it to be educational on the PR side, but occasionally we do push things that make it fail (such as when the squash merge adds the PR number at the end at this makes it exceed the allowed length).

That was a minor problem, as it would just show up as broken, but everything else would be alright. Now since #12108 it can actually prevent the other CIs from running, which is OK from a PR standpoint, but not so for the main develop stream, as preventing the build generation can mess up with launchers, I guess. Its own commit presented this behavior https://github.com/OpenRCT2/OpenRCT2/commit/8f77bd94e15c462c9d1c5a8c20aa5ec829d32def.

So I'm open to discussions, some alternatives I think of, but haven't looked in depth on how feasible each of them is:
1 - Not making the builds depend on the commit lint step
2 - Restricting the lint commit to be just on pull request, not on push

tupaschoal:
~I kept the requirement of it to still run first, it is just not checking the result anymore~

~PS: I did a long commit message intentionally to double check it works, will reword and force push once it is confirmed.~

Turns out that `needs` already requires the jobs before it to be successful [\[1\]][1], so I simplified the CI script in another commit:
> Identifies any jobs that must complete successfully before this job will run. It can be a string or array of strings. If a job fails, all jobs that need it are skipped unless the jobs use a conditional statement that causes the job to continue.

[1]: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,8 +33,8 @@ jobs:
   windows:
     name: Windows
     runs-on: windows-latest
-    needs: [lint-commit, check-code-formatting]
-    if: needs.lint-commit.result == 'success' && needs.check-code-formatting.result == 'success'
+    needs: [check-code-formatting]
+    if: needs.check-code-formatting.result == 'success'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Hm, I'm not sure if it's necessary to change it. It's nice that we don't have the CI run unnecessarily.

If I'm the only one who thinks this, I'm okay with it, though.

tupaschoal:
My concern, as stated in detial in #12144, is that eventually we might merge things that fail on lint commit and then this tip won't have builds to be fetched by anyone that wants to.

caseyjohnsonwv:
I authored #12108- I wasn't sure if commit linting was intended to be a full-stop fail or not, so I grouped it with code formatting. Seems a little overkill that way in hindsight. There are ways to differentiate between pull request and push events to keep the fail on PRs only, but I think this solution is clean and accomplishes what we need.

PR #12151, opened 2020-07-07
Fix potential null derference

janisozaur:
The error is reported by launchpad.net from its PPA builds

Launchpad is spamming me with error notifications and it's gotten annoying to the point where I need to take action. It might not be the only one, but the CI doesn't reproduce… :shrug:


janisozaur:
https://launchpadlibrarian.net/487548762/buildlog_ubuntu-focal-arm64.openrct2_0.2.6-develop-0~1512-202007071322-g94b3598~ubuntu20.04.1_BUILDING.txt.gz

PR #12154, opened 2020-07-08
Use only ScreenCoordsXY for Text.cpp

tupaschoal:
Now that most of the callers of these functions are using `ScreenCoordsXY`, it makes sense to change them too and probably reap some benefit.

### DIFF
--- a/src/openrct2/drawing/Text.cpp
+++ b/src/openrct2/drawing/Text.cpp
@@ -31,7 +31,7 @@ StaticLayout::StaticLayout(utf8string source, TextPaint paint, int32_t width)
     LineHeight = font_get_line_height(fontSpriteBase);
 }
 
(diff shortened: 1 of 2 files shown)


src/openrct2/drawing/Text.cpp
```diff
@@ -77,55 +76,60 @@ int32_t StaticLayout::GetLineCount()
...
+            alignedCoords.x -= (width - 1) / 2;
             break;
         case TextAlignment::RIGHT:
-            x -= width;
+            alignedCoords.x -= width;
             break;
     }
 
-    ttf_draw_string(dpi, text, paint->Colour, { x, y });
+    ttf_draw_string(dpi, text, paint->Colour, coords);
```
duncanspumpkin, line 96: Wrong variable

tupaschoal: Nice catch!

PR #12156, opened 2020-07-08
Change train creation to use CoordsXYZ

tupaschoal:
Easier to follow on a commit basis. There's a catch in the third commit where I actually get to use the `z` parameter being sent.

The fourth commit just prevents abuse of pointers

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -4675,7 +4675,7 @@ static void ride_create_vehicles_find_first_block(Ride* ride, CoordsXYE* outXYEl
  *
  *  rct2: 0x006DD84C
  */
-static bool ride_create_vehicles(Ride* ride, CoordsXYE* element, int32_t isApplying)
+static bool ride_create_vehicles(Ride* ride, const CoordsXYE& element, int32_t isApplying)
 {
     ride->UpdateMaxVehicles();
     if (ride->subtype == RIDE_ENTRY_INDEX_NULL)
@@ -4696,8 +4696,8 @@ static bool ride_create_vehicles(Ride* ride, CoordsXYE* element, int32_t isApply
         return true;
     }
 
-    TileElement* tileElement = element->element;
-    auto vehiclePos = CoordsXYZ{ *element, element->element->GetBaseZ() };
+    TileElement* tileElement = element.element;
+    auto vehiclePos = CoordsXYZ{ element, element.element->GetBaseZ() };
     int32_t direction = tileElement->GetDirection();
 
     //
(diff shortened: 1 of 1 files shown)


Gymnasiast:
> There's a catch in the third commit where I actually get to use the z parameter being sent.

Could you elaborate?

src/openrct2/ride/Ride.cpp
```diff
@@ -4400,7 +4400,7 @@ static Vehicle* vehicle_create_car(
         // loc_6DDCA4:
         vehicle->TrackSubposition = VEHICLE_TRACK_SUBPOSITION_0;
         int32_t direction = tileElement->GetDirection();
-        auto chosenLoc = CoordsXYZ{ CoordsXY{ carPosition } + word_9A3AB4[direction], tileElement->GetBaseZ() };
+        auto chosenLoc = carPosition + CoordsXYZ{ word_9A3AB4[direction], 0 };
```
tupaschoal: This is what I meant @Gymnasiast . The call stack is
```
vehicle_create_car
vehicle_create_train
vehicle_create_trains
```
Though the z position was being sent to these functions, it was not being used and it was initialized from the `tileElement` instead. I'm now using the `z` as sent. I'll annotate the other places

src/openrct2/ride/Ride.cpp
```diff
@@ -4458,7 +4458,7 @@ static Vehicle* vehicle_create_car(
         }
         vehicle->TrackSubposition = subposition;
 
-        auto chosenLoc = CoordsXYZ{ CoordsXY{ carPosition }, tileElement->GetBaseZ() };
+        auto chosenLoc = carPosition;
```
tupaschoal, line 4461: Another place simplified by sending z correctly

src/openrct2/ride/Ride.cpp
```diff
@@ -4704,24 +4704,21 @@ static bool ride_create_vehicles(Ride* ride, CoordsXYE* element, int32_t isApply
...
     {
-        x = element->x - CoordsDirectionDelta[direction].x;
-        y = element->y - CoordsDirectionDelta[direction].y;
+        vehiclePos -= CoordsXYZ{ CoordsDirectionDelta[direction], 0 };
 
-        tileElement = reinterpret_cast<TileElement*>(map_get_track_element_at({ x, y, z << 3 }));
+        tileElement = reinterpret_cast<TileElement*>(map_get_track_element_at(vehiclePos));
 
-        z = tileElement->base_height;
+        vehiclePos.z = tileElement->GetBaseZ();
```
tupaschoal: Using base Z simplifies those places as well

src/openrct2/ride/Ride.cpp
```diff
@@ -4400,29 +4400,26 @@ static Vehicle* vehicle_create_car(
...
         vehicle->track_progress = 0;
         vehicle->SetState(VEHICLE_STATUS_MOVING_TO_END_OF_STATION);
         vehicle->update_flags = 0;
 
-        CoordsXY chosenLoc;
         // loc_6DDD26:
         do
         {
             vehicle->sprite_direction = scenario_rand() & 0x1E;
-            chosenLoc.y = y + (scenario_rand() & 0xFF);
```
Gymnasiast, line 4421: This is not the same. Previously, it would take the value `x` and `y` had before the loop and then add a random number, but now they accumulate.

tupaschoal: Nice catch

PR #12163, opened 2020-07-08
Part of #11159: save_for_network returns unique_ptr

frutiemax:


src/openrct2/network/Network.cpp
```diff
@@ -1604,12 +1605,12 @@ void Network::Server_Send_MAP(NetworkConnection* connection)
             SendPacketToClients(*packet);
         }
     }
-    free(header);
 }
 
-uint8_t* Network::save_for_network(size_t& out_size, const std::vector<const ObjectRepositoryItem*>& objects) const
+std::unique_ptr<uint8_t[]> Network::save_for_network(
```
ZehMatt, line 1610: I think it would be better to use std::vector<uint8_t> as the result type, this also removes the first parameter ```size_t& out_size```

tupaschoal:
Can you please rebase @frutiemax ?

tupaschoal:
This seems alright, what do you think @ZehMatt ?

PR #12168, opened 2020-07-08
Fix null pointers reported by launchpad builders

janisozaur:


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4060,9 +4064,8 @@ static void window_ride_maintenance_dropdown(rct_window* w, rct_widgetindex widg
                         for (int32_t i = 0; i < ride->num_vehicles; ++i)
                         {
                             uint16_t spriteId = ride->vehicles[i];
-                            while (spriteId != SPRITE_INDEX_NULL)
+                            while (spriteId != SPRITE_INDEX_NULL && (vehicle = GetEntity<Vehicle>(spriteId)))
```
duncanspumpkin, line 4067: the MS compilers don't like this

PR #12172, opened 2020-07-08
Part of #11389: Use Formatter on Ride::FormatStatusTo

tupaschoal:
This is the last remaining `Format*To` that takes a `void*`

Wanna review @brenoguim ?

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -2658,13 +2658,13 @@ static void window_ride_main_invalidate(rct_window* w)
  *
  *  rct2: 0x006AF10A
  */
-static rct_string_id window_ride_get_status_overall_view(rct_window* w)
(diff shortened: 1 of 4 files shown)


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2658,13 +2658,14 @@ static void window_ride_main_invalidate(rct_window* w)
  *
  *  rct2: 0x006AF10A
  */
-static rct_string_id window_ride_get_status_overall_view(rct_window* w, void* arguments)
+static rct_string_id window_ride_get_status_overall_view(rct_window* w)
```
duncanspumpkin, line 2661: This is going backwards. We want to eventually stop using the global variables completely this is readding the global

tupaschoal: Fixed, do squash on merge

PR #12178, opened 2020-07-09
Close #12177: Extract patrol offset and index calculation into method

### LINKED ISSUE #12177: Extract Patrol Area Offset Index function
The patrol area used on `src\openrct2\peep\Staff.cpp` is calculated by offsetting the coordinates with bit shift. This code is repeated on three different functions and it should be extracted out into a separate function.

```
static std::pair<int32_t, int32_t> getPatrolAreaOffsetIndex(const CoordsXY& coords)
{
    // Patrol areas are 4 * 4 tiles (32 * 4) = 128 = 2^^7
    auto hash = ((coords.x & 0x1F80) >> 7) |((coords.y & 0x1F80) >> 1);
    return { hash >> 5, hash & 0x1F };
}

auto [ offset, index ] = getPatrolAreaOffsetIndex(coords);
```

_Originally posted by @duncanspumpkin in https://github.com/OpenRCT2/OpenRCT2/pull/12175#issuecomment-656117140_

Jimver:
Hi again,

As mentioned in #12177 the patrol offset and index calculation code was duplicated in 3 functions. They are now extracted into one method, hence removing this code duplication.

### DIFF
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -373,6 +373,13 @@ void staff_reset_stats()
     }
 }
 
+static std::pair<int32_t, int32_t> getPatrolAreaOffsetIndex(const CoordsXY& coords)
+{
+    // Patrol areas are 4 * 4 tiles (32 * 4) = 128 = 2^^7
+    auto hash = ((coords.x & 0x1F80) >> 7) | ((coords.y & 0x1F80) >> 1);
+    return { hash >> 5, hash & 0x1F };
+}
+
 static bool staff_is_patrol_area_set(int32_t staffIndex, const CoordsXY& coords)
 {
     // Patrol quads are stored in a bit map (8 patrol quads per byte).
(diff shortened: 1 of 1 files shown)


src/openrct2/peep/Staff.cpp
```diff
@@ -373,6 +373,13 @@ void staff_reset_stats()
     }
 }
 
+static std::pair<int32_t, int32_t> getPatrolAreaOffsetIndex(const CoordsXY& coords)
+{
+    // Patrol areas are 4 * 4 tiles (32 * 4) = 128 = 2^^7
+    auto hash = ((coords.x & 0x1F80) >> 7) | ((coords.y & 0x1F80) >> 1);
```
Gymnasiast, line 379: It's not a hash.

duncanspumpkin: I called it hash as it's an amalgamation of the two coords smashed into one variable. It's not really a coord or anything else at this point.

Jimver: Hmm maybe something like `coordsCombined` then?

Gymnasiast: Hm, fair enough. I don't think the name is ideal, but I get where it's coming from.

PR #12184, opened 2020-07-09
Add support for reading .sea (RCT classic) files

IntelOrca:
This implements the decryption algorithm for .sea files used in RollerCoaster Tycoon classic. The encryption key is the filename of the .sea file, so if the user renames the file, OpenRCT2 will not be able to read it.

Credit to @marijnvdwerf for his discovery and original PHP implementation.


pizza2004:
I think it would be good to add an error message specifically if it fails to decrypt.  You can mention the thing about renaming the files.

src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -218,10 +218,10 @@ static const char* getFilterPatternByType(const int32_t type, const bool isSave)
     switch (type & 0x0E)
     {
         case LOADSAVETYPE_GAME:
-            return isSave ? "*.sv6" : "*.sv6;*.sc6;*.sc4;*.sv4;*.sv7";
+            return isSave ? "*.sv6" : "*.sv6;*.sc6;*.sea;*.sc4;*.sv4;*.sv7";
 
         case LOADSAVETYPE_LANDSCAPE:
-            return isSave ? "*.sc6" : "*.sc6;*.sv6;*.sc4;*.sv4;*.sv7";
+            return isSave ? "*.sc6" : "*.sc6;*.sea;*.sv6;*.sc4;*.sv4;*.sv7";
```
duncanspumpkin, line 224: Keep the same order as above

src/openrct2/rct2/SeaDecrypt.cpp
```diff
@@ -0,0 +1,109 @@
+/*****************************************************************************
```
duncanspumpkin, line 1: Should we perhaps make a separate folder just for classic rather than put it in rct2. There are likely a few constants and such we will need to have.

src/openrct2/rct2/SeaDecrypt.cpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+        s1 = (s1 + (s1 << 5)) ^ fileName[i];
+    }
+
+    return EncryptionKey{ s0, s1 };
+}
+
+static std::unique_ptr<uint8_t[]> CreateMask(const EncryptionKey& key)
+{
+    auto result = std::make_unique<uint8_t[]>(MASK_SIZE);
```
duncanspumpkin, line 47: A std vector would work fine for this and is generally the preferred default unless absolute speed is required.

src/openrct2/rct2/SeaDecrypt.cpp
```diff
@@ -0,0 +1,109 @@
...
+        if (i >= 2)
+        {
+            *dst8++ = (s0 >> 11) & 0xFF;
+            if (i >= 3)
+            {
+                *dst8++ = (s0 >> 19) & 0xFF;
+                if (i >= 4)
+                {
+                    *dst8++ = (seed1 >> 24) & 0xFF;
+                    i -= 4;
```
duncanspumpkin, line 68: This loop is suspicious. It goes in an infinite loop of mask is not a multiple of 4. Couldn't we change this so that i does not exist. We know ahead of time that 0x1000 divides into 4 so all of these ifs will always been taken. You then just check for if dst equals last value.

src/openrct2/rct2/SeaDecrypt.cpp
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+    }
+    return result;
+}
+
+static void Decrypt(std::vector<uint8_t>& data, const EncryptionKey& key)
+{
+    auto mask = CreateMask(key);
+    const uint8_t* mask8 = (const uint8_t*)mask.get();
```
duncanspumpkin, line 79: This pointer is not required.

src/openrct2/rct2/SeaDecrypt.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+std::vector<uint8_t> DecryptSea(const fs::path& path)
+{
+    auto key = GetEncryptionKey(path.filename().u8string());
+    auto data = File::ReadAllBytes(path.u8string());
+
+    // Last 4 bytes is the checksum
+    size_t inputSize = data.size() - 4;
+    uint32_t checksum;
+    std::memcpy(&checksum, data.data() + inputSize, sizeof(checksum));
```
duncanspumpkin, line 104: Should we be performing some sort of validity check with the checksum?

src/openrct2-ui/windows/LoadSave.cpp
IntelOrca: Any particular reason?

src/openrct2/rct2/SeaDecrypt.cpp
IntelOrca: I haven't looked into the checksum. I can't do anything without knowing how to calculate it, generally we ignore checksums though as we like to load anything that we can. We could issue a warning, but as I say we first need to know how it calculates the checksum.

src/openrct2/rct2/SeaDecrypt.cpp
IntelOrca: Forgetting about the vector for a moment, how do you index into a `unique_ptr<uint8_t[]>`?

src/openrct2-ui/windows/LoadSave.cpp
duncanspumpkin: It's neater.

src/openrct2/rct2/SeaDecrypt.cpp
duncanspumpkin: Should be possible with just the normal [] operator https://en.cppreference.com/w/cpp/memory/unique_ptr/operator_at

distribution/changelog.txt
```diff
@@ -13,6 +13,7 @@
 - Feature: [#11788] Command to extract images from a .DAT file.
 - Feature: [#11959] Hacked go-kart tracks can now use 2x2 bends, 3x3 bends and S-bends.
 - Feature: [#12090] Boosters for the Wooden Roller Coaster (if the "Show all track pieces" cheat is enabled).
+- Feature: [#12184] .sea (RCT classic) scenario files can now be imported.
```
Gymnasiast, line 16:
```suggestion
- Feature: [#12184] .sea (RCT Classic) scenario files can now be imported.
```

src/openrct2-ui/scripting/CustomWindow.cpp
```diff
@@ -1205,7 +1205,7 @@ namespace OpenRCT2::Ui::Windows
                 {
                     if (customWidgetInfo->Name == name)
                     {
-                        return i;
+                        return static_cast<rct_widgetindex>(i);
```
Gymnasiast, line 1208: This change appears to be unrelated. Any particular reason?

src/openrct2/rct2/SeaDecrypt.cpp
Gymnasiast: Not sure - the RCT2 folder isn't that full, and apart from .sea and perhaps a few constants, do we expect to be adding more?

src/openrct2-ui/scripting/CustomWindow.cpp
IntelOrca: The solution wouldn't build for me in VS2019 preview, I quickly fixed it so I could get on with what I was doing and then forgot about it and ended up including those changes. I will separate it into a new commit or PR.

src/openrct2/rct2/SeaDecrypt.cpp
IntelOrca: I didn't want to add a new folder just for one single file. If there are more constants and stuff to be added, can that be done in a new PR? @AaronVanGeffen has already updated the xcodeproject now, wouldn't want to require another fix for that.

duncanspumpkin: Leave as is

src/openrct2-win/openrct2-win.cpp
```diff
@@ -18,6 +18,7 @@
 
 // Then the rest
 #include <algorithm>
+#include <iterator>
```
Gymnasiast, line 21: Why this change?

IntelOrca: To fix building in VS2019 preview.

src/openrct2/rct2/SeaDecrypt.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+constexpr int32_t MASK_SIZE = 0x1000;
+
+struct EncryptionKey
+{
+    uint32_t Seed0{};
+    uint32_t Seed1{};
+};
+
+static EncryptionKey GetEncryptionKey(const std::string_view& fileName)
```
janisozaur, line 27: So if user renames the file, it won't decrypt?

Gymnasiast: Have you read the PR description?

src/openrct2/rct2/SeaDecrypt.cpp
janisozaur: I'd like the warning to be presented to user, if possible.

janisozaur: Oh scratch that, haven't quite understood the problem initially.

src/openrct2/rct2/SeaDecrypt.cpp
janisozaur: Obviously I haven't. Thanks for bringing it to my attention.

IntelOrca:
I have added a message to say that the .sea file may have been renamed.

PR #12191, opened 2020-07-10
Improve Sprite Misc Update

duncanspumpkin:
I've added member functions for all of the misc sprites to improve the code. I've also removed the reinterpret cast and now use the safe conversion functions for all of the types.

### DIFF
--- a/src/openrct2/world/Particle.cpp
+++ b/src/openrct2/world/Particle.cpp
@@ -57,63 +57,61 @@ void crashed_vehicle_particle_create(rct_vehicle_colour colours, const CoordsXYZ
  *
  *  rct2: 0x00673298
  */
-void crashed_vehicle_particle_update(VehicleCrashParticle* particle)
+void VehicleCrashParticle::Update()
 {
-    particle->Invalidate0();
-    particle->time_to_live--;
-    if (particle->time_to_live == 0)
+    Invalidate0();
+    time_to_live--;
+    if (time_to_live == 0)
     {
-        sprite_remove(particle);
+        sprite_remove(this);
         return;
     }
 
     // Apply gravity
-    particle->acceleration_z -= 5041;
+    acceleration_z -= 5041;
 
     // Apply air resistance
-    particle->acceleration_x -= (particle->acceleration_x / 256);
-    particle->acceleration_y -= (particle->acceleration_y / 256);
-    particle->acceleration_z -= (particle->acceleration_z / 256);
+    acceleration_x -= (acceleration_x / 256);
+    acceleration_y -= (acceleration_y / 256);
+    acceleration_z -= (acceleration_z / 256);
 
     // Update velocity and position
-    int32_t vx = particle->velocity_x + particle->acceleration_x;
-    int32_t vy = particle->velocity_y + particle->acceleration_y;
-    int32_t vz = particle->velocity_z + particle->acceleration_z;
+    int32_t vx = velocity_x + acceleration_x;
+    int32_t vy = velocity_y + acceleration_y;
+    int32_t vz = velocity_z + acceleration_z;
 
-    int16_t x = particle->x + (vx >> 16);
-    int16_t y = particle->y + (vy >> 16);
-    int16_t z = particle->z + (vz >> 16);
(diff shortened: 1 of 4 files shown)


src/openrct2/world/Particle.cpp
```diff
@@ -57,63 +57,61 @@ void crashed_vehicle_particle_create(rct_vehicle_colour colours, const CoordsXYZ
...
+    velocity_z = vz & 0xFFFF;
 
     // Check collision with land / water
-    int16_t landZ = tile_element_height({ x, y });
-    int16_t waterZ = tile_element_water_height({ x, y });
+    int16_t landZ = tile_element_height(newLoc);
+    int16_t waterZ = tile_element_water_height(newLoc);
 
-    if (waterZ != 0 && particle->z >= waterZ && z <= waterZ)
+    if (waterZ != 0 && z >= waterZ && z <= waterZ)
```
duncanspumpkin, line 93: Second z should be newLoc

src/openrct2/world/Particle.cpp
```diff
@@ -57,63 +57,61 @@ void crashed_vehicle_particle_create(rct_vehicle_colour colours, const CoordsXYZ
...
-        crash_splash_create({ particle->x, particle->y, waterZ });
-        sprite_remove(particle);
+        audio_play_sound_at_location(SoundId::Water2, { x, y, waterZ });
+        crash_splash_create({ x, y, waterZ });
+        sprite_remove(this);
         return;
     }
 
-    if (particle->z >= landZ && z <= landZ)
+    if (z >= landZ && z <= landZ)
```
duncanspumpkin, line 102: Second z should be newLoc.z

src/openrct2/world/Sprite.cpp
```diff
@@ -607,51 +607,51 @@ void sprite_misc_explosion_flare_create(const CoordsXYZ& flarePos)
...
     }
 }
 
 /**
  *
  *  rct2: 0x006731CD
  */
-static void sprite_misc_update(rct_sprite* sprite)
+static void sprite_misc_update(SpriteBase* sprite)
 {
```
tupaschoal, line 625: Should this function be checking for pointer validity?

src/openrct2/world/Particle.cpp
tupaschoal:
```suggestion
    if (waterZ != 0 && z >= waterZ && newLoc.z <= waterZ)
```

src/openrct2/world/Particle.cpp
tupaschoal:
```suggestion
    if (z >= landZ && newLoc.z <= landZ)
```

src/openrct2/world/Sprite.h
```diff
@@ -61,15 +61,18 @@ struct Duck : SpriteGeneric
     int16_t target_y;
     uint8_t state;
 
+    void Update();
+    uint32_t GetFrameImage(int32_t direction) const;
+    void Invalidate();
+    bool IsFlying();
+    void Remove();
```
tupaschoal, line 68: Why unprotecting all of these in this changeset?

duncanspumpkin: They were always unprotected as it's a struct. I added protection to the update sub functions though as they no longer have to be public.

tupaschoal: Ah, I see, I had missed that `private` was one of the additions as well

src/openrct2/world/Sprite.cpp
duncanspumpkin: I'm hoping that that is not required as the As template should mean that part of it is taken care of by the switch. None of the compilers are complaining about that so i think it is happening like that.

duncanspumpkin: Well i checked the assembly of the version generated on my PC. It did not work out that its already validated the type so not quite as good as I hoped.

src/openrct2/world/Sprite.h
AaronVanGeffen: If you're adding private members, this should probably be a class instead.

duncanspumpkin: I was adding a public member. It's just the 4 existing public members no longer need to be public.

PR #12182, opened 2020-07-09
Close #12159: Use Coords for Ride functions

heliobatimarqui:

PR #12171, opened 2020-07-08
Close #12165: Use Coords for Vehicle functions

heliobatimarqui:
Minor alterations in function prototypes and calls.

PR #12192, opened 2020-07-10
Use big Z coords on some Ride station functions

tupaschoal:
This was pointed out by @duncanspumpkin on #12171 and #12182 . Note that these PRs themselves don't introduce the error, but they do make it harder to detect

FYI @heliobatimarqui we have two coordinates representations, which we joyfully call "big coords" and "small coords", respectively `CoordsXY(Z)` and `TileCoordsXY(Z)`. If you look closely at `TileCoordsXYZ`, you'll see it converts itself into `CoordsXYZ` by doing a multiplication:
https://github.com/OpenRCT2/OpenRCT2/blob/36c9259a5f867efa4c62259d37d246268eff2d39/src/openrct2/world/Location.hpp#L444

So when we convert to `CoordsXYZ`, we expect all coordinates it holds to be either all "big coords" or all "small coords", which was not happening where Duncan pointed out.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -7225,7 +7225,7 @@ TileElement* get_station_platform(int32_t x, int32_t y, int32_t z, int32_t z_tol
             if (!tileElement->AsTrack()->IsStation())
                 continue;
 
-            if (z - z_tolerance > tileElement->base_height || z + z_tolerance < tileElement->base_height)
+            if (z - z_tolerance > tileElement->GetBaseZ() || z + z_tolerance < tileElement->GetBaseZ())
             {
                 /* The base height if tileElement is not within
                  * the z tolerance. */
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -7225,7 +7225,7 @@ TileElement* get_station_platform(int32_t x, int32_t y, int32_t z, int32_t z_tol
             if (!tileElement->AsTrack()->IsStation())
                 continue;
 
-            if (z - z_tolerance > tileElement->base_height || z + z_tolerance < tileElement->base_height)
+            if (z - z_tolerance > tileElement->GetBaseZ() || z + z_tolerance < tileElement->GetBaseZ())
```
duncanspumpkin, line 7228: This again isn't an issue with this pr but might be on the other. Z_tolerance is not the same as clearanceZ which is what rangedZ is expecting.

tupaschoal: I don't get what you think needs or should be changed here. I get that `z_tolerance != clearanceZ`, so maybe we just shouldn't use `CoordsXYRangedZ` there?

tupaschoal: I see you commented there now, nvm

Gymnasiast: Just to make sure: are `z` and `z_tolerance` already big Zs here?

Issue #12161, opened 2020-07-08
Use CoordsXYE on Peep functions


tupaschoal:
Some Peep functions still take individual `x` and `y` parameters. The goal is to make them use `CoordsXYE`.

Note: You can do all of the functions in all PR (seems small overall), but please do each of them into a different commit.

- [x] - `peep_interact_with_path` should receive a `const CoordsXYE&` instead of `x` , `y` and `tile_element`.
- [x] - `peep_interact_with_entrance` should receive a `const CoordsXYE&` instead of `x` , `y` and `tile_element`.
- [x] - `peep_footpath_move_forward` should receive a `const CoordsXYE&` instead of `x` , `y` and `tile_element`.
- [ ] - `peep_interact_with_shop` should receive a `const CoordsXYE&` instead of `x` , `y` and `tile_element`.

Functions are at `src\openrct2\peep\Peep.cpp`

PR #12193, opened 2020-07-10
Part of #12161: Use CoordsXYE on peep_interact_with_path

frutiemax:


src/openrct2/peep/Peep.cpp
```diff
@@ -3061,7 +3062,7 @@ void Peep::PerformNextAction(uint8_t& pathing_result, TileElement*& tile_result)
 
         if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH)
         {
-            peep_interact_with_path(this, newLoc.x, newLoc.y, tileElement);
+            peep_interact_with_path(this, CoordsXYE{ newLoc, tileElement });
```
tupaschoal, line 3065:
```suggestion
            peep_interact_with_path(this, { newLoc, tileElement });
```

src/openrct2/peep/Peep.cpp
```diff
@@ -2732,9 +2732,10 @@ static void peep_footpath_move_forward(Peep* peep, int16_t x, int16_t y, TileEle
  *
  *  rct2: 0x0069455E
  */
-static void peep_interact_with_path(Peep* peep, int16_t x, int16_t y, TileElement* tile_element)
+static void peep_interact_with_path(Peep* peep, const CoordsXYE& coords)
 {
     // 0x00F1AEE2
+    auto* tile_element = coords.element;
```
tupaschoal, line 2738: No need for the `*` when the auto is deducing a pointer, it can deduce this just fine. You only ever need to type either `auto` or `auto&`
```suggestion
    auto tile_element = coords.element;
```

Issue #12160, opened 2020-07-08
Use Coords for Viewport functions


tupaschoal:
Some viewport functions still take individual `x` and `y` parameters. The goal is to make them use `ScreenCoordsXY`.

Note: You can do all of the functions in all PR (seems small overall), but please do each of them into a different commit.

- [x] - `viewport_move` should receive a `const ScreenCoordsXY&` instead of `x` and  `y`
- [x] - `is_sprite_interacted_with` should receive a `const ScreenCoordsXY&` instead of `x` and  `y`
- [x] - `viewport_coord_to_map_coord` should receive a `const ScreenCoordsXY&` instead of `x` and  `y`
- [x] - `viewport_redraw_after_shift` should receive a `const ScreenCoordsXY&` instead of `x` and  `y`
- [x] - `is_sprite_interacted_with_palette_set` should receive a `const ScreenCoordsXY&` instead of `x` and  `y`

Functions are at `src\openrct2\interface\Viewport.cpp`

tupaschoal:
Done by #12216

duncanspumpkin:
I'd like to also see this comment applied as part of this refactor https://github.com/OpenRCT2/OpenRCT2/commit/27734e29714ca5ee1b0aa6ed7befd78630c254fc#r40554133 it's just a little tidy up.

duncanspumpkin:
Reopening to keep track of my above request.

PR #12195, opened 2020-07-10
Part of #12160: Use Coords for viewport_move

frutiemax:


src/openrct2/interface/Viewport.cpp
```diff
@@ -620,7 +620,7 @@ void viewport_update_position(rct_window* window)
         y += viewport->viewPos.y;
     }
 
-    viewport_move(x, y, window, viewport);
+    viewport_move({ x, y }, window, viewport);
```
tupaschoal, line 623: Can you please convert the `x` and `y` on this function to a single `ScreenCoordsXY` variable and use it here?

duncanspumpkin: And preferrably make it two variables one for the scope L585 to L623 and another for L546 to L549 (viewportMidPoint)

src/openrct2/interface/Viewport.cpp
```diff
@@ -543,10 +543,10 @@ void viewport_update_position(rct_window* window)
 
     viewport_set_underground_flag(0, window, viewport);
 
-    int16_t x = window->savedViewPos.x + viewport->view_width / 2;
-    int16_t y = window->savedViewPos.y + viewport->view_height / 2;
+    auto coords = ScreenCoordsXY{ window->savedViewPos.x + viewport->view_width / 2,
+                                  window->savedViewPos.y + viewport->view_height / 2 };
```
tupaschoal, line 547: As suggested by Duncan previously:
```suggestion
    auto viewportMidPoint = ScreenCoordsXY{ window->savedViewPos.x + viewport->view_width / 2,
                                            window->savedViewPos.y + viewport->view_height / 2 };
```

PR #12199, opened 2020-07-10
Close #12174: Remove Formatter additions with uintptr_t

### LINKED ISSUE #12174: Remove Formatter additions with uintptr_t
The `Formatter` class has a method `Add` for writing to a format string. Some of the calls to this method are using aliased types, not their actual types.

This issue is to remove the `Add<uintptr_t>` calls and instead pass the type they want to be written (like `const char*` or `uint8_t`)

frutiemax:


src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -717,7 +717,7 @@ static void window_loadsave_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     // Draw path text
     auto ft = Formatter::Common();
-    ft.Add<uintptr_t>(Platform::StrDecompToPrecomp(buffer));
```
tupaschoal, line 720: The `Platform::StrDecompToPrecomp` is there for a reson, it seems, so instead of dropping it altogether, change its return type to be `utf8*` and stop doing the `reinterpret_cast` there

Then restore the usages of this function that you removed, please

Gymnasiast:
Have you made sure to at least test one place where this is used? For example, whether the load/save window is drawn correctly?

Gymnasiast:
https://github.com/OpenRCT2/OpenRCT2/pull/12203 seems to have caught a handful of occurrences that you haven't.

tupaschoal:
> #12203 seems to have caught a handful of occurrences that you haven't.

Yep, two occurrences on `TitleCommandEditor`. Just search the whole project for `Add<uintptr_t>(`

Gymnasiast:
@frutiemax Code looks OK, but could you answer my question?

frutiemax:
I still need to test this tbh.

PR #12202, opened 2020-07-10
Feature: key.pressed hook for the scripting engine

frutiemax:
This is a test script.

```
function onKeyPressed(keyArgs){
	console.log(keyArgs);
}

function main() {
	context.subscribe("key.pressed",onKeyPressed);
	return;
}

registerPlugin({
    name: 'KeyTest',
    version: '1.0',
    authors: ['Frutiemax'],
	type: 'remote',
	license: 'MIT',
    main: main
});
```

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -226,6 +226,7 @@ declare global {
         subscribe(hook: "network.leave", callback: (e: NetworkEventArgs) => void): IDisposable;
         subscribe(hook: "ride.ratings.calculate", callback: (e: RideRatingsCalculateArgs) => void): IDisposable;
         subscribe(hook: "action.location", callback: (e: ActionLocationArgs) => void): IDisposable;
+        subscribe(hook: "key.pressed", callback: (e: KeyArgs) => void): IDisposable;
     }
 
     interface Configuration {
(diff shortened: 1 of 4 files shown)


distribution/openrct2.d.ts
```diff
@@ -226,6 +226,7 @@ declare global {
         subscribe(hook: "network.leave", callback: (e: NetworkEventArgs) => void): IDisposable;
         subscribe(hook: "ride.ratings.calculate", callback: (e: RideRatingsCalculateArgs) => void): IDisposable;
         subscribe(hook: "action.location", callback: (e: ActionLocationArgs) => void): IDisposable;
+		subscribe(hook: "key.pressed", callback: (e: KeyArgs) => void): IDisposable;
```
Gymnasiast, line 229: Alignment is off here.

IntelOrca:
@frutiemax I have plans to implement registration of shortcuts which show up on the shortcut editor window and can be changed by the user. This would provide a better and more generic solution to this. The only thing that might not cover is adding new modifier keys, e.g. something like holding CTRL while using a tool. What is the particular plugin you need this hook for?

Umdlye:
I would personally enjoy/make plugins with custom modifier keys like using modifier + left click to pick up peeps without having to open their menu. This would come in handy for that.

frutiemax:
@IntelOrca no script in particular tbh. Your idea of implementation seems better.

IntelOrca:
@frutiemax I will try to work on it soon. It will take a bit longer as I first need to change the existing shortcut file into YAML.

PR #12203, opened 2020-07-11
Close #12174: Remove Formatter additions with uintptr_t

### LINKED ISSUE #12174: Remove Formatter additions with uintptr_t

heliobatimarqui:


Gymnasiast:
There was already a PR for this: https://github.com/OpenRCT2/OpenRCT2/pull/12199

I'm therefore closing this one.

Gymnasiast:
There was a PR to address #12174 already: https://github.com/OpenRCT2/OpenRCT2/pull/12199
I'm therefore closing this one.

src/openrct2-ui/windows/LoadSave.cpp
```diff
@@ -717,7 +717,7 @@ static void window_loadsave_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     // Draw path text
     auto ft = Formatter::Common();
-    ft.Add<uintptr_t>(Platform::StrDecompToPrecomp(buffer));
```
Gymnasiast, line 720: Why did you remove this call? It is there for a good reason.

heliobatimarqui: Hi. I've seen that you closed it now. I didn't see that there was a PR for this already. In regard of removing the call, if I recall correctly this funtion is purely a cast.

Gymnasiast: Note that it lives in `Platform`. It means the implementation is OS-dependent. Take a look at the macOS implementation and you will see that is not purely a cast.

PR #12205, opened 2020-07-11
Fix issue with korean file paths

duncanspumpkin:
fs::path incorrectly interpreted the type of std::string to require no conversion for the filesystem.
This is fixed by using the u8path function.
Will need to keep this in mind as more functions start to use the std::filesystem.

### DIFF
--- a/src/openrct2/platform/Shared.cpp
+++ b/src/openrct2/platform/Shared.cpp
@@ -94,7 +94,8 @@ namespace Platform
 
     bool FileExists(const std::string path)
     {
-        fs::path file = path;
+        fs::path file = fs::u8path(path);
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Should be good. I’d like to verify this doesn’t inadvertently break Linux or macOS, though. I won’t have time today, so if someone else would like to check, then please.

Gymnasiast:
@AaronVanGeffen What would constitute a good test of this?

AaronVanGeffen:
@Gymnasiast Hmm, I was thinking a test with a homedir consisting of Unicode chars, somehow. Like `/home/アーロン` or some such.

PR #12208, opened 2020-07-11
Part of #12161: Use CoordsXYE on peep_interact_with_entrance

frutiemax:


src/openrct2/peep/Peep.cpp
```diff
@@ -2481,16 +2482,16 @@ static void peep_interact_with_entrance(Peep* peep, int16_t x, int16_t y, TileEl
         }
 
         bool found = false;
-        auto entrance = std::find_if(gParkEntrances.begin(), gParkEntrances.end(), [x, y](const auto& e) {
-            return e.x == floor2(x, 32) && e.y == floor2(y, 32);
+        auto entrance = std::find_if(gParkEntrances.begin(), gParkEntrances.end(), [coords](const auto& e) {
+            return e.x == floor2(coords.x, 32) && e.y == floor2(coords.y, 32);
```
tupaschoal, line 2486: This can be simplified with:
```suggestion
            return e == coords.ToTileStart();
```

src/openrct2/peep/Peep.cpp
```diff
@@ -2481,16 +2482,16 @@ static void peep_interact_with_entrance(Peep* peep, int16_t x, int16_t y, TileEl
...
         });
         if (entrance != gParkEntrances.end())
         {
             int16_t z = entrance->z / 8;
             entranceDirection = entrance->direction;
 
-            int16_t next_x = (x & 0xFFE0) + CoordsDirectionDelta[entranceDirection].x;
-            int16_t next_y = (y & 0xFFE0) + CoordsDirectionDelta[entranceDirection].y;
+            int16_t next_x = (coords.x & 0xFFE0) + CoordsDirectionDelta[entranceDirection].x;
```
Gymnasiast, line 2493: Doing `coords.x & 0xFFE0` is the same as `floor2(coords.x, 32)` and the same as doing ToTileStart().

Change next_x and next_y to a CoordsXY nextLoc.

```suggestion
            CoordsXY nextLoc = coords.ToTileStart()+ CoordsDirectionDelta[entranceDirection];
```

This also allows you to simply the call on line 2497 to
```
             TileElement* nextTileElement = map_get_first_element_at(nextLoc);
```

src/openrct2/peep/Peep.cpp
frutiemax: e is a CoordsXYE while ToTileStart is CoordsXY. There is no operator == for these two types, should I add one?

Gymnasiast: CoordsXYE should decay to CoordsXY, so you should just be able to compare them. If that doesn't work, try switching the order around.

Issue #12164, opened 2020-07-08
Use Coords for TrackDesign functions


tupaschoal:
Some TrackDesign functions still take individual `x` and `y` parameters. The goal is to make them use `CoordsXY*`.

Note: You can do all of the functions in all PR (seems small overall), but please do each of them into a different commit.

- [x] - `track_design_update_max_min_coordinates` should receive a `const CoordsXYZ&` instead of `x`, `y` and `z`.
- [x] - `track_design_add_selection_tile` should receive a `const CoordsXY&` instead of `x` and `y`.
- [x] - `track_design_place_maze` should receive a `const CoordsXYZ&` instead of `x`, `y` and `z`.
- [ ] - `track_design_place_ride` should receive a `const CoordsXYZ&` instead of `x`, `y` and `z`.
- [ ] - `place_virtual_track` should receive a `const CoordsXYZ&` instead of `x`, `y` and `z`.

Functions are at `src\openrct2\ride\TrackDesign.cpp`

PR #12209, opened 2020-07-11
Part of #12164: Use Coords for track_design_add_selection_tile

frutiemax:


src/openrct2/ride/TrackDesign.cpp
```diff
@@ -843,16 +843,12 @@ void track_design_mirror(TrackDesign* td6)
...
-        if (tile.x == x && tile.y == y)
-        {
-            return;
-        }
-    }
-    gMapSelectionTiles.push_back(CoordsXY(x, y));
+    auto tileIterator = std::find(gMapSelectionTiles.begin(), gMapSelectionTiles.end(), coords);
+    if (tileIterator == gMapSelectionTiles.end())
+        return;
+    gMapSelectionTiles.push_back(coords);
```
tupaschoal, line 851:
```suggestion
    if (tileIterator == gMapSelectionTiles.end())
    {
        gMapSelectionTiles.push_back(coords);
    }
```

It should add when it doesn't find any already

PR #12211, opened 2020-07-11
Fix: Map Generator shows incorrect map sizes

Gymnasiast:
I added a changelog entry because this issue has been in the game from December 2019.


duncanspumpkin:
I think we should have used the formatter class for this. We need to really work on changing the draw string functions to stop using void* args.

PR #12217, opened 2020-07-12
Part of #12161: Use CoordsXYE on peep_footpath_move_forward

frutiemax:


Gymnasiast:
There are only five calls in our codebase.

I'm considering this a WIP.

PR #7579, opened 2018-05-28
Make widget spinners horizontal

AaronVanGeffen:
The current spinner widgets are rather tiny, leading to unfortunate misclicks and generally making them a pain to use. I'm talking, of course, about widgets like:
screenshot_20180528_224245

We can improve them dramatically by making the buttons horizontal, thereby increasing the click areas:
screenshot_20180528_224215

It appears these buttons aren't properly vertically aligned yet, but it works as a proof of concept. If we stick with the triangles, the widgets would currently look like:
screenshot_20180528_224332

I'd like some input before I start converting all spinner widgets… Your thoughts?

PR #12219, opened 2020-07-12
Refactor widget definitions

AaronVanGeffen:
Recently, there has been a discussion about the way we define widgets internally. Currently, they use four parameters with absolute positions (top, bottom, left, right). With this PR, I propose to change this to the way we do this in OpenLoco: one parameter for position (x, y) and dimensions (x, y).

Streamlining widgets throughout all windows will require quite a bit of refactoring. Generally, it's best to separate changing the underlying structure from the desired end result. I therefore propose we change all this with an intermediary function before we change the actual widget structure. 

The way I picture this, is to introduce a constexpr `MakeWidget` function that replaces the current 'hardcoded' widget definitions. This way, we don't have to change the widget structure yet ­— we can first focus on defining widgets 'new style' in our widget lists. I have [written a Python script](https://gist.github.com/AaronVanGeffen/fb2e5bf4612ca157cd8a29c2e09c6820) to assist in the conversion.

Aside from the `MakeWidget` function, I have also added a helper function `MakeRemapWidget`, which adds the `IMAGE_TYPE_REMAP` flag, as well as `MakeSpinnerWidgets`, which replaces the current `SPINNER_WIDGETS` macro. We can discuss their names; nothing here is set in stone.

For this PR, I have changed the widget definitions for the three terraforming tools: the clear scenery window, land tool, and water tool windows. This gives us a few concrete examples to talk about; I imagine there might be some discussion about how the new definitions should be ordered or formatted.

I hope this change will make it much easier to add and move widgets around. Personally, I also hope it will remove the need for many of the macros in the cheats and tile inspector windows, too. However, tackling those is beyond the scope of what I aim to achieve with this first PR.

### DIFF
--- a/src/openrct2-ui/windows/ClearScenery.cpp
+++ b/src/openrct2-ui/windows/ClearScenery.cpp
@@ -35,12 +35,12 @@ static constexpr const int32_t WH = 94;
 
 static rct_widget window_clear_scenery_widgets[] = {
     WINDOW_SHIM(WINDOW_TITLE, WW, WH),
-    { WWT_IMGBTN,   0,  27, 70, 17, 48, SPR_LAND_TOOL_SIZE_0,                       STR_NONE },                         // preview box
-    { WWT_TRNBTN,   1,  28, 43, 18, 33, IMAGE_TYPE_REMAP | SPR_LAND_TOOL_DECREASE,        STR_ADJUST_SMALLER_LAND_TIP },      // decrement size
-    { WWT_TRNBTN,   1,  54, 69, 32, 47, IMAGE_TYPE_REMAP | SPR_LAND_TOOL_INCREASE,        STR_ADJUST_LARGER_LAND_TIP },       // increment size
-    { WWT_FLATBTN,  1,  7,  30, 53, 76, IMAGE_TYPE_REMAP | SPR_G2_BUTTON_TREES,           STR_CLEAR_SCENERY_REMOVE_SMALL_SCENERY_TIP }, // small scenery
-    { WWT_FLATBTN,  1,  37, 60, 53, 76, IMAGE_TYPE_REMAP | SPR_G2_BUTTON_LARGE_SCENERY,   STR_CLEAR_SCENERY_REMOVE_LARGE_SCENERY_TIP }, // large scenery
-    { WWT_FLATBTN,  1,  67, 90, 53, 76, IMAGE_TYPE_REMAP | SPR_G2_BUTTON_FOOTPATH,        STR_CLEAR_SCENERY_REMOVE_FOOTPATHS_TIP }, // footpaths
+    MakeWidget     ({27, 17}, {44, 32}, WWT_IMGBTN,  0, SPR_LAND_TOOL_SIZE_0,        STR_NONE),                                   // preview box
+    MakeRemapWidget({28, 18}, {16, 16}, WWT_TRNBTN,  1, SPR_LAND_TOOL_DECREASE,      STR_ADJUST_SMALLER_LAND_TIP),                // decrement size
+    MakeRemapWidget({54, 32}, {16, 16}, WWT_TRNBTN,  1, SPR_LAND_TOOL_INCREASE,      STR_ADJUST_LARGER_LAND_TIP),                 // increment size
+    MakeRemapWidget({ 7, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_TREES,         STR_CLEAR_SCENERY_REMOVE_SMALL_SCENERY_TIP), // small scenery
+    MakeRemapWidget({37, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_LARGE_SCENERY, STR_CLEAR_SCENERY_REMOVE_LARGE_SCENERY_TIP), // large scenery
+    MakeRemapWidget({67, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_FOOTPATH,      STR_CLEAR_SCENERY_REMOVE_FOOTPATHS_TIP),     // footpaths
     { WIDGETS_END },
 };
 
--- a/src/openrct2-ui/windows/Land.cpp
+++ b/src/openrct2-ui/windows/Land.cpp
@@ -43,16 +43,13 @@ enum WINDOW_LAND_WIDGET_IDX {
 
 static rct_widget window_land_widgets[] = {
     WINDOW_SHIM(WINDOW_TITLE, WW, WH),
-
-    { WWT_FLATBTN,  1,  19, 42, 19, 42,         SPR_RIDE_CONSTRUCTION_SLOPE_UP,         STR_ENABLE_MOUNTAIN_TOOL_TIP }, // mountain mode
-    { WWT_FLATBTN,  1,  55, 78, 19, 42,         SPR_PAINTBRUSH,                         STR_DISABLE_ELEVATION },        // paint mode
-
-    { WWT_IMGBTN,   0,  27, 70, 48, 79,         SPR_LAND_TOOL_SIZE_0,                   STR_NONE },                     // preview box
-    { WWT_TRNBTN,   1,  28, 43, 49, 64,         IMAGE_TYPE_REMAP | SPR_LAND_TOOL_DECREASE,    STR_ADJUST_SMALLER_LAND_TIP },  // decrement size
-    { WWT_TRNBTN,   1,  54, 69, 63, 78,         IMAGE_TYPE_REMAP | SPR_LAND_TOOL_INCREASE,    STR_ADJUST_LARGER_LAND_TIP },   // increment size
-
-    { WWT_FLATBTN,  1,  2,  48, 106,    141,    0xFFFFFFFF,                             STR_CHANGE_BASE_LAND_TIP },     // floor texture
-    { WWT_FLATBTN,  1,  49, 95, 106,    141,    0xFFFFFFFF,                             STR_CHANGE_VERTICAL_LAND_TIP }, // wall texture
+    MakeWidget     ({19,  19}, {24, 24}, WWT_FLATBTN, 1, SPR_RIDE_CONSTRUCTION_SLOPE_UP, STR_ENABLE_MOUNTAIN_TOOL_TIP), // mountain mode
+    MakeWidget     ({55,  19}, {24, 24}, WWT_FLATBTN, 1, SPR_PAINTBRUSH,                 STR_DISABLE_ELEVATION),        // paint mode
+    MakeWidget     ({27,  48}, {44, 32}, WWT_IMGBTN,  0, SPR_LAND_TOOL_SIZE_0,           STR_NONE),                     // preview box
+    MakeRemapWidget({28,  49}, {16, 16}, WWT_TRNBTN,  1, SPR_LAND_TOOL_DECREASE,         STR_ADJUST_SMALLER_LAND_TIP),  // decrement size
+    MakeRemapWidget({54,  63}, {16, 16}, WWT_TRNBTN,  1, SPR_LAND_TOOL_INCREASE,         STR_ADJUST_LARGER_LAND_TIP),   // increment size
+    MakeWidget     ({ 2, 106}, {47, 36}, WWT_FLATBTN, 1, 0xFFFFFFFF,                     STR_CHANGE_BASE_LAND_TIP),     // floor texture
+    MakeWidget     ({49, 106}, {47, 36}, WWT_FLATBTN, 1, 0xFFFFFFFF,                     STR_CHANGE_VERTICAL_LAND_TIP), // wall texture
     { WIDGETS_END },
 };
 
--- a/src/openrct2/interface/Widget.h
+++ b/src/openrct2/interface/Widget.h
@@ -10,6 +10,7 @@
 #ifndef _WIDGET_H_
 #define _WIDGET_H_
 
+#include "../localisation/StringIds.h"
 #include "Window.h"
 
 enum WINDOW_WIDGET_TYPES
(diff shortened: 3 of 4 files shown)


src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,62 @@ enum
...
+constexpr rct_widget MakeRemapWidget(
+    ScreenCoordsXY origin, ScreenSize size, uint8_t type, uint8_t colour, uint32_t content = 0xFFFFFFFF,
+    rct_string_id tooltip = STR_NONE)
+{
+    rct_widget out = MakeWidget(origin, size, type, colour, content, tooltip);
+
+    // TODO: implement this as a constant.
+    out.content |= IMAGE_TYPE_REMAP;
+
+    return out;
```
AaronVanGeffen, line 90: Can be simplified to:
```suggestion
    return MakeWidget(origin, size, type, colour, IMAGE_TYPE_REMAP | content, tooltip);
```

src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,62 @@ enum
...
 
+constexpr rct_widget MakeWidget(
+    ScreenCoordsXY origin, ScreenSize size, uint8_t type, uint8_t colour, uint32_t content = 0xFFFFFFFF,
+    rct_string_id tooltip = STR_NONE)
+{
+    rct_widget out = {};
+    out.left = origin.x;
+    out.right = origin.x + size.width - 1;
+    out.top = origin.y;
+    out.bottom = origin.y + size.height - 1;
```
tupaschoal, line 72: I don't quite get why providing the `width` and the `height` with extra 1s and then subtracting immediately

src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,62 @@ enum
...
+    out.type = type;
+    out.colour = colour;
+    out.content = content;
+    out.tooltip = tooltip;
+
+    return out;
+}
+
+constexpr rct_widget MakeRemapWidget(
+    ScreenCoordsXY origin, ScreenSize size, uint8_t type, uint8_t colour, uint32_t content = 0xFFFFFFFF,
```
tupaschoal, line 82: I wonder if we shouldn't group `ScreenCoordsXY` and `ScreenSize` into some struct. Maybe `WindowStub` or something alike?

src/openrct2/interface/Widget.h
AaronVanGeffen: Future proofing, really. Right now, they're not necessary, as we translate them back to the old struct — for now. However, if left = right, you'd still want width = 1. Likewise, if top = bottom, you'd want height = 1.

src/openrct2/interface/Widget.h
AaronVanGeffen: Hmm, I rather like having them both separately, to be honest. Easy to pass them on separately to rendering functions that way. I'm also worried that it would impede readability. For example, if WindowStub is only going to contain a `ScreenCoordsXY` and `ScreenSize`, that'd make the initialisation look like `{ {54, 32}, {16, 16} }` or something.

src/openrct2-ui/windows/Land.cpp
```diff
@@ -44,15 +44,15 @@ enum WINDOW_LAND_WIDGET_IDX {
...
-    { WWT_IMGBTN,   0,  27, 70, 48, 79,         SPR_LAND_TOOL_SIZE_0,                   STR_NONE },                     // preview box
-    { WWT_TRNBTN,   1,  28, 43, 49, 64,         IMAGE_TYPE_REMAP | SPR_LAND_TOOL_DECREASE,    STR_ADJUST_SMALLER_LAND_TIP },  // decrement size
-    { WWT_TRNBTN,   1,  54, 69, 63, 78,         IMAGE_TYPE_REMAP | SPR_LAND_TOOL_INCREASE,    STR_ADJUST_LARGER_LAND_TIP },   // increment size
+    MakeWidget({27, 48}, {44, 32}, WWT_IMGBTN, 0, SPR_LAND_TOOL_SIZE_0, STR_NONE), // preview box
+    MakeRemapWidget({28, 49}, {16, 16}, WWT_TRNBTN, 1, SPR_LAND_TOOL_DECREASE, STR_ADJUST_SMALLER_LAND_TIP), // decrement size
+    MakeRemapWidget({54, 63}, {16, 16}, WWT_TRNBTN, 1, SPR_LAND_TOOL_INCREASE, STR_ADJUST_LARGER_LAND_TIP), // increment size
 
-    { WWT_FLATBTN,  1,  2,  48, 106,    141,    0xFFFFFFFF,                             STR_CHANGE_BASE_LAND_TIP },     // floor texture
-    { WWT_FLATBTN,  1,  49, 95, 106,    141,    0xFFFFFFFF,                             STR_CHANGE_VERTICAL_LAND_TIP }, // wall texture
+    MakeWidget({2, 106}, {47, 36}, WWT_FLATBTN, 1, 0xFFFFFFFF, STR_CHANGE_BASE_LAND_TIP), // floor texture
```
Gymnasiast, line 54: I'm sure we have a constant for the 0xFFFFFFFF.

src/openrct2/interface/Widget.h
Gymnasiast: Let's keep them separate.

src/openrct2-ui/windows/Land.cpp
AaronVanGeffen: I found `SPR_NONE = -1`, but the enum doesn't define a data type. If `int32_t`, it fits, of course. (It actually says `0x7FFFF` in the comments, but I believe that to be outdated.)

src/openrct2-ui/windows/ClearScenery.cpp
```diff
@@ -35,12 +35,12 @@ static constexpr const int32_t WH = 94;
...
-    { WWT_FLATBTN,  1,  37, 60, 53, 76, IMAGE_TYPE_REMAP | SPR_G2_BUTTON_LARGE_SCENERY,   STR_CLEAR_SCENERY_REMOVE_LARGE_SCENERY_TIP }, // large scenery
-    { WWT_FLATBTN,  1,  67, 90, 53, 76, IMAGE_TYPE_REMAP | SPR_G2_BUTTON_FOOTPATH,        STR_CLEAR_SCENERY_REMOVE_FOOTPATHS_TIP }, // footpaths
+    MakeWidget({27, 17}, {44, 32}, WWT_IMGBTN, 0, SPR_LAND_TOOL_SIZE_0, STR_NONE), // preview box
+    MakeRemapWidget({28, 18}, {16, 16}, WWT_TRNBTN, 1, SPR_LAND_TOOL_DECREASE, STR_ADJUST_SMALLER_LAND_TIP), // decrement size
+    MakeRemapWidget({54, 32}, {16, 16}, WWT_TRNBTN, 1, SPR_LAND_TOOL_INCREASE, STR_ADJUST_LARGER_LAND_TIP), // increment size
+    MakeRemapWidget({7, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_TREES, STR_CLEAR_SCENERY_REMOVE_SMALL_SCENERY_TIP), // small scenery
+    MakeRemapWidget({37, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_LARGE_SCENERY, STR_CLEAR_SCENERY_REMOVE_LARGE_SCENERY_TIP), // large scenery
+    MakeRemapWidget({67, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_FOOTPATH, STR_CLEAR_SCENERY_REMOVE_FOOTPATHS_TIP), // footpaths
```
AaronVanGeffen, line 43:
```suggestion
    MakeWidget     ({27, 17}, {44, 32},  WWT_IMGBTN, 0, SPR_LAND_TOOL_SIZE_0,        STR_NONE),                                   // preview box
    MakeRemapWidget({28, 18}, {16, 16},  WWT_TRNBTN, 1, SPR_LAND_TOOL_DECREASE,      STR_ADJUST_SMALLER_LAND_TIP),                // decrement size
    MakeRemapWidget({54, 32}, {16, 16},  WWT_TRNBTN, 1, SPR_LAND_TOOL_INCREASE,      STR_ADJUST_LARGER_LAND_TIP),                 // increment size
    MakeRemapWidget({ 7, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_TREES,         STR_CLEAR_SCENERY_REMOVE_SMALL_SCENERY_TIP), // small scenery
    MakeRemapWidget({37, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_LARGE_SCENERY, STR_CLEAR_SCENERY_REMOVE_LARGE_SCENERY_TIP), // large scenery
    MakeRemapWidget({67, 53}, {24, 24}, WWT_FLATBTN, 1, SPR_G2_BUTTON_FOOTPATH,      STR_CLEAR_SCENERY_REMOVE_FOOTPATHS_TIP),     // footpaths
```

@Gymnasiast How's this?

Gymnasiast: That's what I had in mind!

src/openrct2-ui/windows/Land.cpp
Gymnasiast: Perhaps leave it as it is then, and create a separate issue to sort it out.

src/openrct2-ui/windows/ClearScenery.cpp
```diff
@@ -35,12 +35,12 @@ static constexpr const int32_t WH = 94;
 
 static rct_widget window_clear_scenery_widgets[] = {
     WINDOW_SHIM(WINDOW_TITLE, WW, WH),
-    { WWT_IMGBTN,   0,  27, 70, 17, 48, SPR_LAND_TOOL_SIZE_0,                       STR_NONE },                         // preview box
-    { WWT_TRNBTN,   1,  28, 43, 18, 33, IMAGE_TYPE_REMAP | SPR_LAND_TOOL_DECREASE,        STR_ADJUST_SMALLER_LAND_TIP },      // decrement size
```
Gymnasiast, line 39: Looks like the sizes differ. Is that deliberate?

AaronVanGeffen: There will be a difference of 1 in width and height by design. Other than that, there shouldn't be any. Which sizes are different, exactly?

src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,57 @@ enum
...
+{
+    const int16_t xPos = origin.x + size.width - 25;
+    const int16_t yPos = origin.y + 1;
+    const uint16_t width = 12;
+    const uint16_t height = size.height - 2;
+
+    return MakeWidget({ xPos, yPos }, { width, height }, WWT_BUTTON, colour, STR_NUMERIC_DOWN, tooltip);
+}
+
+constexpr rct_widget MakeStepperIncreaseWidget(
```
Gymnasiast, line 103: These are called "spinner" in our code. Why are using a different term here? That's only going to cause confusion - the code should use one term.

(If you believe the spinner name is incorrect/worse than stepper, it should be an issue of its own, and it should be renamed throughout.)

AaronVanGeffen: Thanks. No, I did not intend to change the naming. Now that you mention it, I remember the discussion from #7579 and did not mean to drag that up again. I will change it to match the old naming tonight.

src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,57 @@ enum
 
 constexpr uint8_t SCROLLBAR_WIDTH = 10;
 
+constexpr rct_widget MakeWidget(
+    ScreenCoordsXY origin, ScreenSize size, uint8_t type, uint8_t colour, uint32_t content = 0xFFFFFFFF,
```
tupaschoal, line 65:
```suggestion
    const ScreenCoordsXY& origin, const ScreenSize& size, uint8_t type, uint8_t colour, uint32_t content = 0xFFFFFFFF,
```

src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,57 @@ enum
...
+    out.type = type;
+    out.colour = colour;
+    out.content = content;
+    out.tooltip = tooltip;
+
+    return out;
+}
+
+constexpr rct_widget MakeRemapWidget(
+    ScreenCoordsXY origin, ScreenSize size, uint8_t type, uint8_t colour, uint32_t content = 0xFFFFFFFF,
```
tupaschoal, line 82:
```suggestion
    const ScreenCoordsXY& origin, const ScreenSize& size, uint8_t type, uint8_t colour, uint32_t content = 0xFFFFFFFF,
```

src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,57 @@ enum
...
+    rct_string_id tooltip = STR_NONE)
+{
+    return MakeWidget(origin, size, type, colour, IMAGE_TYPE_REMAP | content, tooltip);
+}
+
+#define MakeSpinnerWidgets(...)                                                                                                \
+    MakeWidget(__VA_ARGS__), MakeSpinnerDecreaseWidget(__VA_ARGS__), MakeSpinnerIncreaseWidget(__VA_ARGS__)
+
+constexpr rct_widget MakeSpinnerDecreaseWidget(
+    ScreenCoordsXY origin, ScreenSize size, [[maybe_unused]] uint8_t type, uint8_t colour,
```
tupaschoal, line 92:
```suggestion
    const ScreenCoordsXY& origin, const ScreenSize& size, [[maybe_unused]] uint8_t type, uint8_t colour,
```

src/openrct2/interface/Widget.h
```diff
@@ -60,6 +61,57 @@ enum
...
+    const int16_t xPos = origin.x + size.width - 25;
+    const int16_t yPos = origin.y + 1;
+    const uint16_t width = 12;
+    const uint16_t height = size.height - 2;
+
+    return MakeWidget({ xPos, yPos }, { width, height }, WWT_BUTTON, colour, STR_NUMERIC_DOWN, tooltip);
+}
+
+constexpr rct_widget MakeSpinnerIncreaseWidget(
+    ScreenCoordsXY origin, ScreenSize size, [[maybe_unused]] uint8_t type, uint8_t colour,
```
tupaschoal, line 104:
```suggestion
    const ScreenCoordsXY&  origin, const ScreenSize& size, [[maybe_unused]] uint8_t type, uint8_t colour,
```

tupaschoal:
Should this be assigned to the "New Window Framework" project, for proper tracking?

Gymnasiast:
@tupaschoal Yeah, there will be plenty of PRs for the window system. Keeps them organised.

tupaschoal:
Alright, done

PR #12236, opened 2020-07-13
Close #12161: Use CoordsXYE on peep_interact_with_shop

### LINKED ISSUE #12161: Use CoordsXYE on Peep functions

frutiemax:


src/openrct2/peep/Peep.cpp
```diff
@@ -2915,8 +2916,10 @@ static bool peep_interact_with_shop(Peep* peep, int16_t x, int16_t y, TileElemen
             money16 money = 0;
             guest->SpendMoney(money, cost, ExpenditureType::ParkRideTickets);
         }
-        peep->DestinationX = (x & 0xFFE0) + 16;
-        peep->DestinationY = (y & 0xFFE0) + 16;
+
+        auto coordsCenter = coords.ToTileCentre();
```
Gymnasiast, line 2920:
```suggestion
        auto coordsCentre = coords.ToTileCentre();
```

We use British English throughout.

src/openrct2/peep/Peep.cpp
```diff
@@ -2862,8 +2862,9 @@ static void peep_interact_with_path(Peep* peep, const CoordsXYE& coords)
  *
  *  rct2: 0x00693F70
  */
-static bool peep_interact_with_shop(Peep* peep, int16_t x, int16_t y, TileElement* tile_element)
+static bool peep_interact_with_shop(Peep* peep, const CoordsXYE& coords)
 {
+    auto tile_element = coords.element;
     ride_id_t rideIndex = tile_element->AsTrack()->GetRideIndex();
```
Gymnasiast, line 2868: This is the only place `tile_element` is used. Rather than creating an intermediate variable, just do 
```suggestion
    ride_id_t rideIndex = coords.element->AsTrack()->GetRideIndex();
```

PR #12238, opened 2020-07-13
Close #12223: Assert that Formatter::Add() only receives desired types

### LINKED ISSUE #12223: Statically assert that Formatter::Add() only receives desired types
The `Formatter::Add()` method should only receive the actual types of the content being sent to them, without casts getting it the way. Its usage has just been cleaned up, so we should `static_assert` that the type being sent over to them is one of the below:

`char*`
`const char*`
`const utf8*`
`int16_t`
`int32_t`
`money32`
`rct_string_id`
`uint16_t`
`uint32_t`
`utf8*`

This will be a chain of `std::is_same<TSpecified, (one of the types above)>::value || `

frutiemax:


src/openrct2/localisation/Localisation.h
```diff
@@ -131,6 +131,16 @@ class Formatter
     {
         static_assert(sizeof(TSpecified) <= sizeof(uintptr_t), "Type too large");
         static_assert(sizeof(TDeduced) <= sizeof(uintptr_t), "Type too large");
+
+        // checking if the specified type is a valid one
+        constexpr bool validType = std::is_same<TSpecified, char*>::value || std::is_same<TSpecified, const char*>::value
```
duncanspumpkin, line 136: Since c++17 you can simplify this to std::is_same_v then you don't need the ::value

frutiemax:
The clang format makes a huge line of code when doing multiple std::is_save_v...

tupaschoal:
> The clang format makes a huge line of code when doing multiple std::is_same_v...

Argh, that's very sad, maybe there is a way of fooling clang-format for this (though it looks like a bug we should report back to them)

Gymnasiast:
You can use `clang-format off` and `clang-format on` to turn it off for the assert.

src/openrct2/localisation/Localisation.h
```diff
@@ -131,6 +131,18 @@ class Formatter
     {
         static_assert(sizeof(TSpecified) <= sizeof(uintptr_t), "Type too large");
         static_assert(sizeof(TDeduced) <= sizeof(uintptr_t), "Type too large");
+
+        // checking if the specified type is a valid one
```
Gymnasiast, line 135: This comment more or less states what was already obvious from the code. Just get rid of it.

src/openrct2/localisation/Localisation.h
```diff
@@ -131,6 +131,18 @@ class Formatter
...
+        // clang-format off
+        static_assert(
+            std::is_same_v<TSpecified, char*>           || std::is_same_v<TSpecified, const char*>  ||
```
Gymnasiast, line 138: Hm, maybe put all of them on separate lines? What do you think @tupaschoal ?

src/openrct2/localisation/Localisation.h
tupaschoal: Agreed

src/openrct2/localisation/Localisation.h
tupaschoal: I definitely prefer one at a line so we don't have to keep aligning the `||`, and also keeping alike types together, sort of this way:
```c++
static_assert(
    std::is_same_v<TSpecified, char*> || 
    std::is_same_v<TSpecified, const char*> ||
    std::is_same_v<TSpecified, int16_t> ||
    std::is_same_v<TSpecified, int32_t> ||
    std::is_same_v<TSpecified, money32> ||
    std::is_same_v<TSpecified, rct_string_id> ||
    std::is_same_v<TSpecified, uint8_t*> ||
    std::is_same_v<TSpecified, uint16_t> ||
    std::is_same_v<TSpecified, uint32_t> ||
    std::is_same_v<TSpecified, utf8*> ||
    std::is_same_v<TSpecified, const utf8*>
);
```

PR #12246, opened 2020-07-13
Small constants cleanup

Gymnasiast:


Gymnasiast:
That was more or less the reason, yeah. All constants in this PR have become obsolete because we changed our own implementation to differ from the RCT1/2 implementations. As such, these constants are only used during import and export and should be declared in RCT12.h.

PR #12247, opened 2020-07-13
Related to #12160: Refactor viewport_coord_to_map_coord using Rotate

frutiemax:
@duncanspumpkin This is not working as it is.


duncanspumpkin:
If you revert back to the first version the issue is the rotation is not normal you want to do map the current rotation to the following order 0 3 2 1

Issue #12243, opened 2020-07-13
Expand usage of Coords objects for maze functions


tupaschoal:
Some Maze functions still take individual coordinate parameters. The goal is to make them use `CoordsXY*`.

Note: You can do all of the functions in all PR (seems small overall), but please do each of them into a different commit.

- [x] - `MazeRemoveTrack` should receive a `const CoordsXYZD&` instead of `x`, `y`, `z` and `direction` (`src\openrct2\actions\RideDemolishAction.hpp`).
- [ ] - `maze_set_track` should receive a `const CoordsXYZD&` instead of `x`, `y`, `z` and `direction` (`src\openrct2\actions\GameActionCompat.cpp`).
- [ ] - `MazeGetSegmentBit` should receive a `const CoordsXY&` instead of `x` and `y` (`src\openrct2\actions\MazeSetTrackAction.hpp`).

duncanspumpkin:
Please take remember that any change to a GameAction will require a network version increment.

tupaschoal:
Thanks for the reminder!

GalBr:
Can I finish this issue?

PR #12251, opened 2020-07-13
Part of #12243: Use coords for MazeRemoveTrack

frutiemax:


src/openrct2/actions/RideDemolishAction.hpp
```diff
@@ -340,7 +340,7 @@ DEFINE_GAME_ACTION(RideDemolishAction, GAME_COMMAND_DEMOLISH_RIDE, GameActionRes
             for (Direction dir : ALL_DIRECTIONS)
             {
                 const CoordsXY& off = DirOffsets[dir];
-                money32 removePrice = MazeRemoveTrack(location.x + off.x, location.y + off.y, location.z, dir);
+                money32 removePrice = MazeRemoveTrack({ location + off, dir });
```
Gymnasiast, line 343: location + off results in a CoordsXY, with dir being interpreted as a Z.

You can prevent that by changing line 342 as follows:
```const CoordsXYZ& off = { DirOffsets[dir], 0 };```

File: src/openrct2/actions/RideDemolishAction.hpp [hunk 2]
```diff
@@ -339,8 +339,8 @@ DEFINE_GAME_ACTION(RideDemolishAction, GAME_COMMAND_DEMOLISH_RIDE, GameActionRes
 
             for (Direction dir : ALL_DIRECTIONS)
             {
-                const CoordsXY& off = DirOffsets[dir];
-                money32 removePrice = MazeRemoveTrack(location.x + off.x, location.y + off.y, location.z, dir);
+                const CoordsXYZ& off = { DirOffsets[dir], 0 };
```
Gymnasiast, line 342: Sorry, made a mistake. It should have been 
```suggestion
                const CoordsXYZ off = { DirOffsets[dir], 0 };
```

PR #12255, opened 2020-07-14
Refactor/more constants in track curve

Gymnasiast:


tupaschoal:
FYI, this is on my plate to review, but I'll take some time double checking everything

PR #12256, opened 2020-07-14
Remove goto from window_create_auto_pos

tupaschoal:
Easier to follow on a commit-by-commit basis, as always

### DIFF
--- a/src/openrct2-ui/interface/Window.cpp
+++ b/src/openrct2-ui/interface/Window.cpp
@@ -182,25 +182,6 @@ rct_window* window_create_auto_pos(
     auto screenWidth = uiContext->GetWidth();
     auto screenHeight = uiContext->GetHeight();
 
-    // TODO dead code, looks like it is cascading the new window offset from an existing window
-    // we will have to re-implement this in our own way.
-    //
-    // if (cls & 0x80) {
-    //  cls &= ~0x80;
-    //  rct_window *w = window_find_by_number(0, 0);
-    //  if (w != nullptr) {
-    //      if (w->x > -60 && w->x < screenWidth - 20) {
-    //          if (w->y < screenHeight - 20) {
-    //              int32_t x = w->x;
-    //              if (w->x + width > screenWidth)
-    //                  x = screenWidth - 20 - width;
-    //              int32_t y = w->y;
-    //              return window_create(x + 10, y + 10, width, height, event_handlers, cls, flags);
-    //          }
-    //      }
-    //  }
-    // }
-
     ScreenCoordsXY screenPos{};
 
     // Place window in an empty corner of the screen
--- a/src/openrct2-ui/interface/Window.cpp
+++ b/src/openrct2-ui/interface/Window.cpp
@@ -201,60 +201,44 @@ rct_window* window_create_auto_pos(
     //  }
     // }
 
+    ScreenCoordsXY screenPos{};
+
     // Place window in an empty corner of the screen
-    auto screenPos = ScreenCoordsXY{ 0, 30 };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
+    {
+        const std::array cornerPositions = {
+            ScreenCoordsXY{ 0, 30 },                                          // topLeft
+            ScreenCoordsXY{ screenWidth - width, 30 },                        // topRight
+            ScreenCoordsXY{ 0, screenHeight - 34 - height },                  // bottomLeft
+            ScreenCoordsXY{ screenWidth - width, screenHeight - 34 - height } // bottomRight
+        };
 
-    screenPos = { screenWidth - width, 30 };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
-
-    screenPos = { 0, screenHeight - 34 - height };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
-
-    screenPos = { screenWidth - width, screenHeight - 34 - height };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
+        for (const auto& cornerPos : cornerPositions)
+        {
+            screenPos = cornerPos;
+            if (window_fits_within_space(cornerPos, width, height))
+                goto foundSpace;
+        }
+    }
 
     // Place window next to another
     for (auto& w : g_window_list)
     {
         if (w->flags & WF_STICK_TO_BACK)
             continue;
-
-        screenPos = { w->windowPos.x + w->width + 2, w->windowPos.y };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x - w->width - 2, w->windowPos.y };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x, w->windowPos.y + w->height + 2 };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x, w->windowPos.y - w->height - 2 };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x + w->width + 2, w->windowPos.y - w->height - 2 };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x - w->width - 2, w->windowPos.y - w->height - 2 };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
-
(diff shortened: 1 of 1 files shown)


src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,56 @@ rct_window* window_create(
...
-        goto foundSpace;
-
-    screenPos = { 0, screenHeight - 34 - height };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
-
-    screenPos = { screenWidth - width, screenHeight - 34 - height };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
+    std::array cornerPositions = { ScreenCoordsXY{ 0, 30 }, ScreenCoordsXY{ screenWidth - width, 30 },
```
duncanspumpkin, line 199: I'd prefer if this was formatted with the four entries on four lines.

src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,56 @@ rct_window* window_create(
...
+                                   ScreenCoordsXY{ 0, screenHeight - 34 - height },
+                                   ScreenCoordsXY{ screenWidth - width, screenHeight - 34 - height } };
+    for (auto& cornerPos : cornerPositions)
+    {
+        if (window_fits_within_space(cornerPos, width, height))
```
duncanspumpkin, line 204: Can this function modify corner pos?

src/openrct2-ui/interface/Window.cpp
duncanspumpkin: Should this array be const?

src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,56 @@ rct_window* window_create(
     return w;
 }
 
+static ScreenCoordsXY ClampWindowToScreen(const ScreenCoordsXY& pos, int32_t screenWidth, int32_t width)
```
duncanspumpkin, line 178: const all of the variables.

src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,56 @@ rct_window* window_create(
...
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
-
-    screenPos = { screenWidth - width, screenHeight - 34 - height };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
+    std::array cornerPositions = { ScreenCoordsXY{ 0, 30 }, ScreenCoordsXY{ screenWidth - width, 30 },
+                                   ScreenCoordsXY{ 0, screenHeight - 34 - height },
+                                   ScreenCoordsXY{ screenWidth - width, screenHeight - 34 - height } };
+    for (auto& cornerPos : cornerPositions)
```
duncanspumpkin, line 202: const auto& depends on if window fits within space which i don't think it does

src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,56 @@ rct_window* window_create(
...
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x + w->width + 2, w->windowPos.y + w->height + 2 };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x - w->width - 2, w->windowPos.y + w->height + 2 };
-        if (window_fits_within_space(screenPos, width, height))
-            goto foundSpace;
+        std::array offsets = { ScreenCoordsXY{ w->width + 2, 0 },
```
duncanspumpkin, line 213: const

src/openrct2-ui/interface/Window.cpp
```diff
@@ -263,21 +233,16 @@ rct_window* window_create_auto_pos(
...
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x - w->width - 2, w->windowPos.y };
-        if (window_fits_on_screen(screenPos, width, height))
-            goto foundSpace;
-
-        screenPos = { w->windowPos.x, w->windowPos.y + w->height + 2 };
-        if (window_fits_on_screen(screenPos, width, height))
-            goto foundSpace;
+        std::array offsets = { ScreenCoordsXY{ w->width + 2, 0 }, ScreenCoordsXY{ -w->width - 2, 0 },
```
duncanspumpkin, line 236: Const

src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,56 @@ rct_window* window_create(
...
-    //              int32_t x = w->x;
-    //              if (w->x + width > screenWidth)
-    //                  x = screenWidth - 20 - width;
-    //              int32_t y = w->y;
-    //              return window_create(x + 10, y + 10, width, height, event_handlers, cls, flags);
-    //          }
-    //      }
-    //  }
-    // }
+    ScreenCoordsXY screenPos{};
```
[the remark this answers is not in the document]
tupaschoal: Ah, yeah, it became unused on the last commit, but it was still needed when there were gotos

src/openrct2-ui/interface/Window.cpp
tupaschoal: I made them so, in any case.

As for the 4-line format, I achieved it by commenting what each coord mean

src/openrct2-ui/interface/Window.cpp
tupaschoal: It can't

src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,56 @@ rct_window* window_create(
...
-    screenPos = { screenWidth - width, screenHeight - 34 - height };
-    if (window_fits_within_space(screenPos, width, height))
-        goto foundSpace;
+    std::array cornerPositions = { ScreenCoordsXY{ 0, 30 }, ScreenCoordsXY{ screenWidth - width, 30 },
+                                   ScreenCoordsXY{ 0, screenHeight - 34 - height },
+                                   ScreenCoordsXY{ screenWidth - width, screenHeight - 34 - height } };
+    for (auto& cornerPos : cornerPositions)
+    {
+        if (window_fits_within_space(cornerPos, width, height))
+            return window_create(ClampWindowToScreen(screenPos, screenWidth, width), width, height, event_handlers, cls, flags);
```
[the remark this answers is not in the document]
tupaschoal: Nice catch

src/openrct2-ui/interface/Window.cpp
tupaschoal: Done. I assume you'd want 4-line array in here as well. I don't know how to fool `clang-format` in this case, as I think comments are not useful. I can turn it off and on, but not sure this is desirable

duncanspumpkin: yeah it would be preferred. I don't see any harm turning it off and on if we cant get what we want.

AaronVanGeffen:
Note to self: check whether this resolves #12262.

src/openrct2-ui/interface/Window.cpp
tupaschoal: Alright, done

tupaschoal:
> Note to self: check whether this resolves #12262.

I'd say it won't, theoretically it is meant to do the exact same thing. It might just be a whole lot easier to investigate now :)

src/openrct2-ui/interface/Window.cpp
```diff
@@ -175,86 +175,58 @@ rct_window* window_create(
     return w;
 }
 
+static ScreenCoordsXY ClampWindowToScreen(const ScreenCoordsXY& pos, const int32_t screenWidth, const int32_t width)
+{
+    auto screenPos = pos;
+    if (screenPos.x < 0)
+        screenPos.x = 0;
+    if (screenPos.x + width > screenWidth)
+        screenPos.x = screenWidth - width;
```
AaronVanGeffen, line 184: If we add similar clauses for the Y coordinate, this would resolve #12262.

AaronVanGeffen:
@tupaschoal Yes, thanks. I'm a bit puzzled it's not checking the Y axis, to be honest. We recently implemented a similar function in OpenLoco that does do this.

https://github.com/OpenLoco/OpenLoco/blob/bc69993a06cbf73b11ce7247936b82c1df1ba793/src/openloco/ui/WindowManager.cpp#L601-L611

src/openrct2-ui/interface/Window.cpp
tupaschoal: I can add it in a different commit or on a different PR, what do you prefer?

AaronVanGeffen: Hmm, perhaps best to leave it for another PR, then. I don't want to cause this one to be scope-creeped.

PR #12258, opened 2020-07-14
Fix #12257: Fix refactoring of is_sprite_interacted_with_palette_set

### LINKED ISSUE #12257: Placing down or clicking on items in-game is now next to impossible
**OS:** Windows 10
**Version:** 0.2.6
**Commit/Build:** 821dc60

A recently introduced bug has made it next to impossible to click on rides, guests, park borders, etc.; place down scenery, footpaths, rides, etc.; or use any tool that involves an adjustable area size (e.g. scenery scatter, land and water tools). Attempting to place down or click on anything is now a matter of being very lucky with the click hitboxes. This bug essentially makes the game unplayable. I couldn't find this bug in the 0.2.6 stable release.

- Reproducible in RCT2 (vanilla)? **No**
- Specific to multiplayer? **No**

**Steps to reproduce:**
1. Start up the game and load up any scenario.
2. Try to place down a new ride, scenery object, or footpath or click on anything (e.g. guests, rides). You will now find it near impossible to even click on or place down anything. Tools that involve an adjustable area size (e.g. buying land and construction rights, scenery demolish, land and water tools) also become virtually unusable because of this. Constructing footpaths above ground and tracked rides still works normally, though, as does selecting rides and guests through the Rides and Guest list windows, respectively.

**Screenshots / Video:**
Sea Skimmer Park 2020-07-13 17-45-09
As you can see here, I have struggled to place down the Large Elm Tree on this map. I have wildly circled my mouse cursor around the screen, and very rarely am I actually able to place down said object (or if it blocks me, then it should say "Can't position this here... [object name] in the way").

frutiemax:
We should reopen #12160: is_sprite_interacted_with_palette_set has a nasty hidden behavior with those integer types I believe.

EDIT : It's fixed now, see my second commit for the correction.

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -1451,8 +1451,8 @@ static bool is_sprite_interacted_with_palette_set(
     }
 
     int32_t round = std::max(1, 1 * dpi->zoom_level);
-    auto origin = coords;
 
+    auto origin = coords;
     if (g1->flags & G1_FLAG_RLE_COMPRESSION)
     {
         origin.y -= (round - 1);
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
This will need a good bit of testing as it might have broken the normal painting functions. Annoyingly i knew about this a few weeks back but couldn't wait fix it because it would cause issues with this code at the time.

duncanspumpkin:
I've squashed your commits together to make it easier to follow since it was just a revert followed by a revert. I'm also going to change the types back to 16bit as the 32bit builds are failing. Its likely that packing isn't an issue anymore and we can get rid of this but I don't want to do that in this PR.

duncanspumpkin:
I've tested this a little bit but there could still be issues can someone else test this on a few parks. Need to make sure things draw correctly and that interactions are correct.

duncanspumpkin:
I've done further testing and I feel it is now correct and ready to merge.

PR #12264, opened 2020-07-15
Use MakeWidget for most Title windows and the ToolTip

tupaschoal:
I wanted to use @AaronVanGeffen's script just for fun. I double checked the results and it works (yay!) , which all seem fine. Some small notes, that we might need if we're in for lots of PRs still (like I believe).
- It doesn't work for widgets that don't use hardcoded numbers, such as https://github.com/OpenRCT2/OpenRCT2/commit/c2503a738ea8a2fda63156e8a706f677975c1d48 and just fails silently
- It adds `// none` comments where there were actually none. It wasn't clear to me if we intended to comment everything or if you just expected everything to have a comment

As a side comment, it's a bit annoying to align every column, but I agree it looks nice :D

### DIFF
--- a/src/openrct2-ui/windows/TitleLogo.cpp
+++ b/src/openrct2-ui/windows/TitleLogo.cpp
@@ -23,7 +23,7 @@ enum
 };
 
 static rct_widget window_title_logo_widgets[] = {
-    { WWT_IMGBTN, 0, 0, WW, 0, WH, STR_NONE, STR_NONE },
+    MakeWidget({ 0, 0 }, { WW + 1, WH + 1 }, WWT_IMGBTN, 0),
     { WIDGETS_END },
 };
 
--- a/src/openrct2-ui/windows/TitleMenu.cpp
+++ b/src/openrct2-ui/windows/TitleMenu.cpp
@@ -31,10 +31,10 @@ enum {
 };
 
 static rct_widget window_title_menu_widgets[] = {
-    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
-    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
-    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
-    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
+    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
+    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
+    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
+    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
     { WIDGETS_END },
 };
 
--- a/src/openrct2-ui/windows/TitleMenu.cpp
+++ b/src/openrct2-ui/windows/TitleMenu.cpp
@@ -31,10 +31,10 @@ enum {
 };
 
 static rct_widget window_title_menu_widgets[] = {
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_MENU_NEW_GAME,        STR_START_NEW_GAME_TIP          },
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_MENU_LOAD_GAME,       STR_CONTINUE_SAVED_GAME_TIP     },
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_G2_MENU_MULTIPLAYER,  STR_SHOW_MULTIPLAYER_TIP        },
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_MENU_TOOLBOX,         STR_GAME_TOOLS_TIP              },
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
     { WIDGETS_END },
 };
 
--- a/src/openrct2-ui/windows/TitleScenarioSelect.cpp
+++ b/src/openrct2-ui/windows/TitleScenarioSelect.cpp
@@ -73,16 +73,16 @@ enum {
 
 static rct_widget window_scenarioselect_widgets[] = {
     WINDOW_SHIM(WINDOW_TITLE, WW, WH),
-    { WWT_IMGBTN,   1,  0,      733,    50,     333,    0xFFFFFFFF,                 STR_NONE },             // tab content panel
-    { WWT_TAB,      1,  3,      93,     17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 1
-    { WWT_TAB,      1,  94,     184,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 2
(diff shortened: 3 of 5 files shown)


AaronVanGeffen:
Ah, this is a pleasant surprise. Thanks for the help.

Aligning the widget lists automatically is on my todo list, yes. I'd like to avoid doing it manually, particularly for the larger sets. Likewise, I'd like to be able to parse positions where constants or macros are used. We'll get there! :)

src/openrct2-ui/windows/TitleScenarioSelect.cpp
```diff
@@ -73,16 +73,16 @@ enum {
...
+    MakeWidget     ({  0, 50}, {734, 284}, WWT_IMGBTN, 1, 0xFFFFFFFF,      STR_NONE), // tab content panel
+    MakeRemapWidget({  3, 17}, { 91,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 1
+    MakeRemapWidget({ 94, 17}, { 91,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 2
+    MakeRemapWidget({185, 17}, { 91,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 3
+    MakeRemapWidget({276, 17}, { 91,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 4
+    MakeRemapWidget({367, 17}, { 91,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 5
+    MakeRemapWidget({458, 17}, {136,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 6
+    MakeRemapWidget({594, 17}, { 91,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 7
+    MakeRemapWidget({685, 17}, { 91,  34}, WWT_TAB,    1, SPR_TAB_LARGE,   STR_NONE), // tab 8
+    MakeWidget     ({  3, 54}, {553, 276}, WWT_SCROLL, 1, SCROLL_VERTICAL, STR_NONE), // level list
```
AaronVanGeffen, line 85: We can leave out the `STR_NONE` value for tooltips entirely, as it's the default value for MakeWidget.

src/openrct2-ui/windows/TitleScenarioSelect.cpp
```diff
@@ -73,16 +73,16 @@ enum {
...
-    { WWT_TAB,      1,  94,     184,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 2
-    { WWT_TAB,      1,  185,    275,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 3
-    { WWT_TAB,      1,  276,    366,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 4
-    { WWT_TAB,      1,  367,    457,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 5
-    { WWT_TAB,      1,  458,    593,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 6
-    { WWT_TAB,      1,  594,    684,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 7
-    { WWT_TAB,      1,  685,    775,    17,     50,     IMAGE_TYPE_REMAP | SPR_TAB_LARGE, STR_NONE },             // tab 8
-    { WWT_SCROLL,   1,  3,      555,    54,     329,    SCROLL_VERTICAL,            STR_NONE },             // level list
+    MakeWidget     ({  0, 50}, {734, 284}, WWT_IMGBTN, 1, 0xFFFFFFFF,      STR_NONE), // tab content panel
```
AaronVanGeffen, line 76: We can leave out `0xFFFFFFFF` and `STR_NONE` here, as they are the respective default values for the `content` and `tooltip` properties in `MakeWidget`.

src/openrct2-ui/windows/Tooltip.cpp
```diff
@@ -21,7 +21,7 @@ enum {
 };
 
 static rct_widget window_tooltip_widgets[] = {
-    { WWT_IMGBTN, 0, 0, 199, 0, 31, 0xFFFFFFFF, STR_NONE },
+    MakeWidget({0, 0}, {200, 32}, WWT_IMGBTN, 0, 0xFFFFFFFF, STR_NONE),
```
AaronVanGeffen, line 24: As above.

src/openrct2-ui/windows/TitleMenu.cpp
```diff
@@ -31,10 +31,10 @@ enum {
...
 
 static rct_widget window_title_menu_widgets[] = {
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_MENU_NEW_GAME,        STR_START_NEW_GAME_TIP          },
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_MENU_LOAD_GAME,       STR_CONTINUE_SAVED_GAME_TIP     },
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_G2_MENU_MULTIPLAYER,  STR_SHOW_MULTIPLAYER_TIP        },
-    { WWT_IMGBTN, 2, 0, 0, 0, 81, SPR_MENU_TOOLBOX,         STR_GAME_TOOLS_TIP              },
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
+    MakeWidget({0, 0}, {1, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
```
AaronVanGeffen, line 37: Hmm, these looks odd to me with a width of 1. They're probably set in an invalidate event, but I think we should adjust the definitions while we're at it.

src/openrct2-ui/windows/TitleLogo.cpp
```diff
@@ -23,7 +23,7 @@ enum
 };
 
 static rct_widget window_title_logo_widgets[] = {
-    { WWT_IMGBTN, 0, 0, WW, 0, WH, STR_NONE, STR_NONE },
+    MakeWidget({ 0, 0 }, { WW + 1, WH + 1 }, WWT_IMGBTN, 0, STR_NONE, STR_NONE),
```
AaronVanGeffen, line 26: I think we can leave out both `STR_NONE`s here, too.

src/openrct2-ui/windows/TitleMenu.cpp
tupaschoal: This is so because the third widget, for the multiplayer, can sometimes be disabled:
https://github.com/OpenRCT2/OpenRCT2/blob/01d0287ee2a135e8f337c75df4834f92da23093c/src/openrct2-ui/windows/TitleMenu.cpp#L91-L96
Thus, it does this logic to calculate the coordinates:
https://github.com/OpenRCT2/OpenRCT2/blob/01d0287ee2a135e8f337c75df4834f92da23093c/src/openrct2-ui/windows/TitleMenu.cpp#L99-L114

I'm a bit confuse as to whether we should fix it above. It will look mostly alright, but we will still have to keep the function above to readjust the positions in case the multiplayer is disabled.

AaronVanGeffen: Thanks — yes, that explains it. I propose we change it to the following:
```suggestion
    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
```

This way, the dimensions of the buttons are clear, even if their coordinates will be changed. Once the underlying widget structure has been changed, we can simplify the code that repositions them, too.

tupaschoal: Done.

Gymnasiast:
> It adds // none comments where there were actually none. It wasn't clear to me if we intended to comment everything or if you just expected everything to have a comment

You should *never* add a comment just for the sake of it. Comments are there to clarify what wasn't already obvious from the code. Most widget comments were introduced when all the widgets were still using naked string numbers and the like, so they were necessary back then.

src/openrct2-ui/windows/Tooltip.cpp
```diff
@@ -21,7 +21,7 @@ enum {
 };
 
 static rct_widget window_tooltip_widgets[] = {
-    { WWT_IMGBTN, 0, 0, 199, 0, 31, 0xFFFFFFFF, STR_NONE },
+    MakeWidget({0, 0}, {200, 32}, WWT_IMGBTN, 0),
```
Gymnasiast, line 24: Why did you replace STR_NONE with 0?

AaronVanGeffen: He didn't -- that's the colour.

Gymnasiast:
... I'm not sure why it says " Gymnasiast deleted a comment from AaronVanGeffen". I removed a comment of my own.

AaronVanGeffen:
@Gymnasiast I had replied to your comment, that's probably why. No worries; I assumed there was no malintent.

tupaschoal:
> ... I'm not sure why it says " Gymnasiast deleted a comment from AaronVanGeffen". I removed a comment of my own.

If it was a line comment, maybe because he replied to yours and GitHub just notifies when you remove content that is not your own?

PR #12265, opened 2020-07-15
Remove goto from input_state_widget_pressed

tupaschoal:


src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -1253,67 +1253,67 @@ void input_state_widget_pressed(
...
                         if (cursor_w_class != w->classification || cursor_w_number != w->number
                             || widgetIndex != cursor_widgetIndex)
-                            goto dropdown_cleanup;
-                        dropdown_index = -1;
-                        if (_inputFlags & INPUT_FLAG_DROPDOWN_STAY_OPEN)
                         {
-                            if (!(_inputFlags & INPUT_FLAG_DROPDOWN_MOUSE_UP))
+                            dropdownCleanup = true;
+                        }
+                        else
```
duncanspumpkin, line 1272: This wasn't an else before, is this correct?

tupaschoal: The diff here didn't become the most easy to see, but it kind of was, if you take a close look at the code:
https://github.com/OpenRCT2/OpenRCT2/blob/01d0287ee2a135e8f337c75df4834f92da23093c/src/openrct2-ui/input/MouseInput.cpp#L1277-L1289

If it matched the `if`, it went to the `goto`, which just skipped everything below, otherwise it did the `else` logic. So it does look like an `if...else` to me

PR #12266, opened 2020-07-15
Close #12252: removed goto from wall_remove_at

### LINKED ISSUE #12252: Remove goto from wall_remove_at
https://github.com/OpenRCT2/OpenRCT2/blob/1034aadb72fdf16ebfb4ecb18a227c046e32e20b/src/openrct2/world/Wall.cpp#L36-L58

heliobatimarqui:
Refactored function wall_remove_at with the intent of remove the goto statement.

### DIFF
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -2254,7 +2254,9 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
 static void window_tile_inspector_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32_t scrollIndex)
 {
     const int32_t listWidth = w->widgets[WIDX_LIST].width();
-    gfx_fill_rect(dpi, dpi->x, dpi->y, dpi->x + dpi->width - 1, dpi->y + dpi->height - 1, ColourMapA[w->colours[1]].mid_light);
+    gfx_fill_rect(
+        dpi, { { dpi->x, dpi->y }, { dpi->x + dpi->width - 1, dpi->y + dpi->height - 1 } },
+        ColourMapA[w->colours[1]].mid_light);
 
     ScreenCoordsXY screenCoords{};
     screenCoords.y = SCROLLABLE_ROW_HEIGHT * (windowTileInspectorElementCount - 1);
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Wall.cpp
```diff
@@ -35,26 +35,24 @@
...
 {
-    TileElement* tileElement;
-
-repeat:
-    tileElement = map_get_first_element_at(wallPos);
-    if (tileElement == nullptr)
-        return;
-    do
+    for (TileElement* tileElement = map_get_first_element_at(wallPos); tileElement;
+         tileElement = map_get_first_element_at(wallPos))
```
tupaschoal, line 39:
```suggestion
    for (auto tileElement = map_get_first_element_at(wallPos); tileElement != nullptr;
         tileElement = map_get_first_element_at(wallPos))
```

src/openrct2/world/Wall.cpp
```diff
@@ -35,26 +35,24 @@
...
+            if (wallPos.baseZ >= tileElement->GetClearanceZ())
+                continue;
+            if (wallPos.clearanceZ <= tileElement->GetBaseZ())
+                continue;
+
+            tile_element_remove_banner_entry(tileElement);
+            map_invalidate_tile_zoom1({ wallPos, tileElement->GetBaseZ(), tileElement->GetBaseZ() + 72 });
+            tile_element_remove(tileElement);
+            break;
+        } while (!(tileElement++)->IsLastForTile());
```
tupaschoal, line 54: Having a `break` just before the `while` seems weird. All it is doing at the beginning is validating whether it's ok to remove and if it is, it proceeds and exits. This is what I come up with, I'll leave it here for your consideration and other people in the team too
```c++
    bool removalDone = false;
    while (!removalDone)
    {
        if (tileElement->GetType() == TILE_ELEMENT_TYPE_WALL && wallPos.baseZ < tileElement->GetClearanceZ()
            && wallPos.clearanceZ > tileElement->GetBaseZ())
        {
            tile_element_remove_banner_entry(tileElement);
            map_invalidate_tile_zoom1({ wallPos, tileElement->GetBaseZ(), tileElement->GetBaseZ() + 72 });
            tile_element_remove(tileElement);
            removalDone = true;
        }
        else
        {
            removalDone = (tileElement++)->IsLastForTile();
        }
    }
```

duncanspumpkin: In general all of the tile functions are do while loops with a break like that.

heliobatimarqui: It seems way better like this though.

heliobatimarqui:
> Have you tested this? It looks like it would just enter an infinite loop. map_get_first_element_at will always return a pointer. It can never be nullptr unless the save file has been corrupted. The way i would have expected this function to be achieved is to create map_get_wall_element CoordsXYRangedZ overload and keep looping on that until it returns a nullptr. Have a look at map.cpp file and all of the other map_get_..._element functions.

Correct me if I'm wrong but, as it is just a refactoring of an existing function, I thought that the tests would cover it. If they do, the tests are still running.

duncanspumpkin:
This function i think is only used in game actions. I've not recorded a game action test that removes walls so it will not be caught in the test suite. It's not hard to add one for that though.

tupaschoal:
> Have you tested this? It looks like it would just enter an infinite loop. map_get_first_element_at will always return a pointer. It can never be nullptr unless the save file has been corrupted. The way i would have expected this function to be achieved is to create map_get_wall_element CoordsXYRangedZ overload and keep looping on that until it returns a nullptr. Have a look at map.cpp file and all of the other map_get_..._element functions.

Is just a matter of changing the `for` to also check `lastForTile`. In any case I think the overload solution is cleaner.

PR #12272, opened 2020-07-15
Refactored the goto statement

### LINKED ISSUE #12269: Remove goto from DrawAllDirtyBlocks
https://github.com/OpenRCT2/OpenRCT2/blob/0951c0c5cc3f69a6097251ddc6bdb6047f8f7959/src/openrct2/drawing/X8DrawingEngine.cpp#L407-L421

0ro8lu:
Close #12269

### DIFF
--- a/src/openrct2/drawing/X8DrawingEngine.cpp
+++ b/src/openrct2/drawing/X8DrawingEngine.cpp
@@ -406,19 +406,21 @@ void X8DrawingEngine::DrawAllDirtyBlocks()
             uint32_t yy;
             for (yy = y; yy < dirtyBlockRows; yy++)
             {
+                bool shouldBreak = false;
                 uint32_t yyOffset = yy * dirtyBlockColumns;
                 for (xx = x; xx < x + columns; xx++)
                 {
                     if (dirtyBlocks[yyOffset + xx] == 0)
                     {
-                        goto endRowCheck;
+                        uint32_t rows = yy - y;
+                        DrawDirtyBlocks(x, y, columns, rows);
+                        shouldBreak = true;
+                        break;
                     }
                 }
+                if(shouldBreak)
+                    break;
             }
-
-        endRowCheck:
-            uint32_t rows = yy - y;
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Thanks for your contribution! This is a good start.

Note that the original code behaves in a slightly different way. After both for loops have terminated, the control flow reaches `endRowCheck`. Your code changes this, so it

src/openrct2/drawing/X8DrawingEngine.cpp
```diff
@@ -406,19 +406,21 @@ void X8DrawingEngine::DrawAllDirtyBlocks()
             uint32_t yy;
             for (yy = y; yy < dirtyBlockRows; yy++)
             {
+                bool shouldBreak = false;
```
Gymnasiast, line 409: Rename this variable to `reachedEndOfRow`. (It's important to know _why_ we have to break.)

src/openrct2/drawing/X8DrawingEngine.cpp
```diff
@@ -406,19 +406,21 @@ void X8DrawingEngine::DrawAllDirtyBlocks()
...
+                        shouldBreak = true;
+                        break;
                     }
                 }
+                if(shouldBreak)
+                    break;
             }
-
-        endRowCheck:
-            uint32_t rows = yy - y;
```
Gymnasiast, line 420: To fix @AaronVanGeffen 's comment, you have to restore these two lines here.

AaronVanGeffen: Hmm, that wouldn't be the end of it, though. If the code `break`s above with these lines restored, `DrawDirtyBlocks` would be called twice. This can be avoided by using `return` instead of `break`, though, probably.

IntelOrca:
This routine is quite critical. Adding IF statements to avoid the goto will incur high a performance cost. In order to break out multiple loops, you will have to extract that out to a new function so you can then use `return`.

PR #12274, opened 2020-07-15
Close #12269: remove goto from x8_drawing_engine

### LINKED ISSUE #12269: Remove goto from DrawAllDirtyBlocks

0ro8lu:
refactored function DrawAllDirtyBlocks with the intent of removing a goto statement

### DIFF
--- a/src/openrct2/drawing/X8DrawingEngine.cpp
+++ b/src/openrct2/drawing/X8DrawingEngine.cpp
@@ -403,32 +403,32 @@ void X8DrawingEngine::DrawAllDirtyBlocks()
             uint32_t columns = xx - x;
 
             // Check rows
-            uint32_t yy;
-            bool reachedEndOfRow = false;
-            for (yy = y; yy < dirtyBlockRows; yy++)
-            {
-                uint32_t yyOffset = yy * dirtyBlockColumns;
-                for (xx = x; xx < x + columns; xx++)
-                {
-                    if (dirtyBlocks[yyOffset + xx] == 0)
-                    {
-                        uint32_t rows = yy - y;
-                        DrawDirtyBlocks(x, y, columns, rows);
-                        reachedEndOfRow = true;
-                        break;
-                    }
-                }
-                if (reachedEndOfRow)
-                    break;
-            }
-
-            if (!reachedEndOfRow)
+            uint32_t yy = y;
+            if(CheckRows(yy, y, xx, x, dirtyBlockRows, dirtyBlockColumns, columns, dirtyBlocks))
             {
                 uint32_t rows = yy - y;
                 DrawDirtyBlocks(x, y, columns, rows);
             }
         }
     }
+}
+
+bool X8DrawingEngine::CheckRows(uint32_t yy, uint32_t y, uint32_t xx, uint32_t x, uint32_t dirtyBlockRows, uint32_t dirtyBlockColumns, uint32_t columns, uint8_t* dirtyBlocks)
+{
+    for (yy = y; yy < dirtyBlockRows; yy++)
+    {
+        uint32_t yyOffset = yy * dirtyBlockColumns;
+        for (xx = x; xx < x + columns; xx++)
+        {
+            if (dirtyBlocks[yyOffset + xx] == 0)
+            {
+                uint32_t rows = yy - y;
+                DrawDirtyBlocks(x, y, columns, rows);
+                return false;
+            }
+        }
+    }
+    return true;
 }
 
 void X8DrawingEngine::DrawDirtyBlocks(uint32_t x, uint32_t y, uint32_t columns, uint32_t rows)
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
I ran some tests and the two branches do get hit and they are both required. It is looking for the largest rectangle of dirty boxes that it can find. So it would exit the loop early if the rectangle isn't the full height of a column and it would not exit the loop early if it is the full height of the column. So for example a full screen rerender would not exit early but a single sprite dirty box would exit early.

src/openrct2/drawing/X8DrawingEngine.cpp
```diff
@@ -403,24 +403,34 @@ void X8DrawingEngine::DrawAllDirtyBlocks()
             uint32_t columns = xx - x;
 
             // Check rows
-            uint32_t yy;
-            for (yy = y; yy < dirtyBlockRows; yy++)
+            uint32_t yy = y;
+            if (CheckRows(yy, y, xx, x, dirtyBlockRows, dirtyBlockColumns, columns, dirtyBlocks))
```
duncanspumpkin, line 407: After running my tests you do not need an if here. The next part of the code should always be run.

src/openrct2/drawing/X8DrawingEngine.cpp
```diff
@@ -403,24 +403,34 @@ void X8DrawingEngine::DrawAllDirtyBlocks()
...
+                DrawDirtyBlocks(x, y, columns, rows);
             }
+        }
+    }
+}
 
-        endRowCheck:
-            uint32_t rows = yy - y;
-            DrawDirtyBlocks(x, y, columns, rows);
+bool X8DrawingEngine::CheckRows(
```
duncanspumpkin, line 416: yy needs to be passed by reference as it is used in the calling function. xx does not need to be passed into this function. `uint32_t dirtyBlockRows, uint32_t dirtyBlockColumns, uint8_t* dirtyBlocks` do not need to be passed into this function as they are available from the class. `x`, `y` and `columns` are not modified in this function and should be `const`. This function does not need to be a `bool`

src/openrct2/drawing/X8DrawingEngine.cpp
```diff
@@ -403,24 +403,34 @@ void X8DrawingEngine::DrawAllDirtyBlocks()
...
+    uint8_t* dirtyBlocks)
+{
+    for (yy = y; yy < dirtyBlockRows; yy++)
+    {
+        uint32_t yyOffset = yy * dirtyBlockColumns;
+        for (xx = x; xx < x + columns; xx++)
+        {
+            if (dirtyBlocks[yyOffset + xx] == 0)
+            {
+                uint32_t rows = yy - y;
```
duncanspumpkin, line 427: Remove this just have the caller call the drawdirty code in every situation.

src/openrct2/drawing/X8DrawingEngine.cpp
duncanspumpkin, line 416: Rename the function GetNumDirtyRows. Then you could make it return `yy-y` and avoid the pass by reference discussed in the above. Would work something like this:

```
...::DrawAllDirtyBlocks{
.... L403
auto rows = GetNumDirtyRows(x, y, columns);
DrawDirtyBlocks(x, y, columns, rows);
... L411
}
uint32_t ...::GetNumDirtyRows(const uint32_t x, const uint32_t y, const uint32_t columns)
{
    const uint32_t dirtyBlockColumns = _dirtyGrid.BlockColumns;
    const uint32_t dirtyBlockRows = _dirtyGrid.BlockRows;
    const uint8_t* dirtyBlocks = _dirtyGrid.Blocks;
...
for...
    for...
      if ...
         return yy - y;
 ...
 return yy - y;
}

Comment (0ro8lu) [reply src/openrct2/drawing/X8DrawingEngine.cpp:416]: I'm sorry i don't get it: GetNumDirtyBlocks needs both xx and yy to function but they are not passed as arguments, also, if i move dirtyBlockColumns, dirtyBlockRows, and dirtyBlocks* inside the body of GetNumDirtyRows then DrawAllDirtyBlocks would not be able to run it's for loops in the beginning of the function. I'm probably missing something.

Comment (duncanspumpkin) [reply src/openrct2/drawing/X8DrawingEngine.cpp:416]: Xx is just a temporary variable for the for loop as is yy but for the outer loop. As for the dirty block variables I'm not saying move them I'm saying have a duplicate of them in this function.

File: src/openrct2/drawing/X8DrawingEngine.cpp [hunk 4]
```diff
@@ -377,50 +377,50 @@ void X8DrawingEngine::ConfigureDirtyGrid()
...
-            uint32_t yy;
-            for (yy = y; yy < dirtyBlockRows; yy++)
+            uint32_t columns = xx - x;
+            auto rows = GetNumDirtyRows(x, y, columns);
+            DrawDirtyBlocks(x, y, columns, rows);
+        }
+    }
+}
+
+uint32_t X8DrawingEngine::GetNumDirtyRows(uint32_t x, uint32_t y, uint32_t columns)
```
duncanspumpkin, line 408: One last thing add `const` to each of these parameters. You have done it in the class deceleration of the function but you still need to do it for the definition.

Issue #12046, opened 2020-06-24
Improve documentation


Gymnasiast:
Currently, there are some problems with documentation:

1. There is not nearly enough of it.
2. User documentation and development documentation is mixed.
3. Documentation is spread out over too many places. In many cases, there is duplication, instead of linking people to the correct place. These places are:
3.1. The OpenRCT2 wiki
3.2. docs.openrct2.io (not working since the move from .website to .io)
3.3. openrct2.org
3.4. openrct2.io (Getting Started)
3.5. readme.md

Instead, the documentation should be reorganised as follows:
1. User documentation should go to docs.openrct2.io, which should also revert to RST/ReadTheDocs.
2. Technical documentation should stay on the OpenRCT2 wiki - for now at least.
3. Both should be vastly expanded.

Umdlye:
I just stumbled upon @telk5093's Korean documentation: https://telk.kr/orct/manual/get_start/basic_interface?v=orct&m=landscape It's gorgeous!

Gymnasiast:
Current docs: https://docs.openrct2.io/playing/footpaths.html
Archived copy of old (RTD) docs: https://web.archive.org/web/20170519180539/https://docs.openrct2.website/en/latest/playing/staff/index.html

PR #12278, opened 2020-07-15
Readme wiki unification for build instructions

tupaschoal:
I've made sure that the wiki has the same information that is already in our README. Then, I pointed to the wiki. This way:

- We make sure these instructions (that are not for everyone) are contained in another place and not cluttering the README
- We get a single place to keep updated

Check https://github.com/tupaschoal/OpenRCT2/tree/readme-wiki-unification#3-building-the-game for a preview of how it looks live

I guess this is kind of part of #12046 (?) I can reword the commits if so

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -79,21 +79,13 @@ Some \*BSD operating systems offer native packages. These packages are usually t
 
 # 3. Building the game
 
-## 3.1 Building prerequisites
-
 OpenRCT2 requires original files of RollerCoaster Tycoon 2 to play. It can be bought at either [Steam](http://store.steampowered.com/app/285330/) or [GOG.com](http://www.gog.com/game/rollercoaster_tycoon_2).
-
----
-
-## 3.2 Compiling and running
 
 * [Building OpenRCT2 on Linux](https://github.com/OpenRCT2/OpenRCT2/wiki/Building-OpenRCT2-on-Linux)
 * [Building OpenRCT2 on macOS using CMake](https://github.com/OpenRCT2/OpenRCT2/wiki/Building-OpenRCT2-on-macOS-using-CMake)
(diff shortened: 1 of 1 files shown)


IntelOrca:
I prefer to keep the build instructions in the readme.md rather than cross referencing the wiki so that there is still an offline and versioned copy of it.

tupaschoal:
> I prefer to keep the build instructions in the readme.md rather than cross referencing the wiki so that there is still an offline and versioned copy of it.

Both of these are possible with the wiki too, it is versioned and it is possible to have it offline (I have it, in fact)

IntelOrca:
It is versioned but it isn't easy to see what the build instructions were for a particular commit in the history of the OpenRCT2 repo. The build instructions for Windows are very small, perhaps you could just reduce the size of Linux and macOS? I am not saying it should cover all the possible platforms and Linux distributions but I would like the basic ones to stay.

tupaschoal:
> It is versioned but it isn't easy to see what the build instructions were for a particular commit in the history of the OpenRCT2 repo. The build instructions for Windows are very small, perhaps you could just reduce the size of Linux and macOS? I am not saying it should cover all the possible platforms and Linux distributions but I would like the basic instructions to stay.

Would it please you if we instead kept a "wiki" folder on OpenRCT2 and synced over to the wiki repo daily, upon modification? (Just like we do with Localisation)

Then we'd have the history all set, a single place to push to and from and Pull requests to edit the wiki

IntelOrca:
Not really, I'd rather just have the build instructions left in the readme. So the basic instructions and then a wiki link for more information and details.

tupaschoal:
So what's the opinion of the rest of the team? If you agree with @IntelOrca I can try to just reduce the content instead of linking

ZehMatt:
I'm with @IntelOrca, having the build instruction in the readme is pretty straight forward.

Gymnasiast:
Closing this are there is no consensus.

PR #12279, opened 2020-07-15
Close #12252: remove of goto in function wall_remove_at

### LINKED ISSUE #12252: Remove goto from wall_remove_at

heliobatimarqui:
* Removed use of goto in function wall_remove_at
* Implemented overload of function map_get_wall_element_at

Replacing pull request #12266. (Sorry, I ended up force pushing). 
I feel that we should have a function to cast stuff to TileElement, as in this case I'd have to return a TileElement from the function map_get_wall_element_at which in turn would make it a lot more different than the other overload. I've made some casts in the function but I'm willing to find a propper solution to this so some suggestions would be great.

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -2396,6 +2396,25 @@ TileElement* map_get_track_element_at_with_direction_from_ride(const CoordsXYZD&
     return nullptr;
 };
 
+WallElement* map_get_wall_element_at(const CoordsXYRangedZ& coords)
+{
+    auto tileElement = map_get_first_element_at(coords);
+
+    if (tileElement != nullptr)
+    {
+        do
+        {
+            if (tileElement->GetType() == TILE_ELEMENT_TYPE_WALL && coords.baseZ < tileElement->GetClearanceZ()
+                && coords.clearanceZ > tileElement->GetBaseZ())
+            {
+                return tileElement->AsWall();
+            }
+        } while (!(tileElement++)->IsLastForTile());
+    }
+
+    return nullptr;
+}
+
 WallElement* map_get_wall_element_at(const CoordsXYZD& wallCoords)
 {
     auto tileWallCoords = TileCoordsXYZ(wallCoords);
--- a/src/openrct2/world/Wall.cpp
+++ b/src/openrct2/world/Wall.cpp
@@ -35,26 +35,13 @@
  */
 void wall_remove_at(const CoordsXYRangedZ& wallPos)
 {
-    TileElement* tileElement;
-
-repeat:
-    tileElement = map_get_first_element_at(wallPos);
-    if (tileElement == nullptr)
-        return;
-    do
+    for (auto wallElement = map_get_wall_element_at(wallPos); wallElement != nullptr;
+         wallElement = map_get_wall_element_at(wallPos))
     {
-        if (tileElement->GetType() != TILE_ELEMENT_TYPE_WALL)
-            continue;
-        if (wallPos.baseZ >= tileElement->GetClearanceZ())
-            continue;
-        if (wallPos.clearanceZ <= tileElement->GetBaseZ())
-            continue;
-
-        tile_element_remove_banner_entry(tileElement);
-        map_invalidate_tile_zoom1({ wallPos, tileElement->GetBaseZ(), tileElement->GetBaseZ() + 72 });
-        tile_element_remove(tileElement);
-        goto repeat;
-    } while (!(tileElement++)->IsLastForTile());
+        tile_element_remove_banner_entry(reinterpret_cast<TileElement*>(wallElement));
+        map_invalidate_tile_zoom1({ wallPos, wallElement->GetBaseZ(), wallElement->GetBaseZ() + 72 });
+        tile_element_remove(reinterpret_cast<TileElement*>(wallElement));
+    }
 }
 
 /**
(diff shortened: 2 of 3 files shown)


src/openrct2/world/Wall.cpp
```diff
@@ -35,26 +35,15 @@
...
 void wall_remove_at(const CoordsXYRangedZ& wallPos)
 {
-    TileElement* tileElement;
-
-repeat:
-    tileElement = map_get_first_element_at(wallPos);
-    if (tileElement == nullptr)
-        return;
-    do
+    WallElement* wallElement = map_get_wall_element_at(wallPos);
```
tupaschoal, line 38:
```suggestion
    auto wallElement = map_get_wall_element_at(wallPos);
```

src/openrct2/world/Wall.cpp
```diff
@@ -35,26 +35,15 @@
...
+    if (wallElement)
```
tupaschoal, line 39:
```suggestion
    if (wallElement != nullptr)
```

src/openrct2/world/Map.cpp
```diff
@@ -2396,6 +2396,23 @@ TileElement* map_get_track_element_at_with_direction_from_ride(const CoordsXYZD&
     return nullptr;
 };
 
+WallElement* map_get_wall_element_at(const CoordsXYRangedZ& coords)
+{
+    TileElement* tileElement = map_get_first_element_at(coords);
```
tupaschoal, line 2401:
```suggestion
    auto tileElement = map_get_first_element_at(coords);
```

src/openrct2/world/Map.cpp
```diff
@@ -2396,6 +2396,23 @@ TileElement* map_get_track_element_at_with_direction_from_ride(const CoordsXYZD&
...
+
+    if (tileElement)
```
tupaschoal, line 2403:
```suggestion
    if (tileElement != nullptr)
```

src/openrct2/world/Wall.cpp
heliobatimarqui: Suggested change taken.

src/openrct2/world/Wall.cpp
heliobatimarqui: Suggested change taken.

src/openrct2/world/Map.cpp
heliobatimarqui: Suggested change taken.

src/openrct2/world/Map.cpp
heliobatimarqui: Suggested change taken.

src/openrct2/world/Wall.cpp
```diff
@@ -35,26 +35,15 @@
...
 void wall_remove_at(const CoordsXYRangedZ& wallPos)
 {
-    TileElement* tileElement;
-
-repeat:
-    tileElement = map_get_first_element_at(wallPos);
-    if (tileElement == nullptr)
-        return;
-    do
+    auto wallElement = map_get_wall_element_at(wallPos);
```
duncanspumpkin, line 38: You need to do a while loop on this as we keep searching for wall elements at this coordinate

src/openrct2/world/Wall.cpp
```diff
@@ -35,26 +35,15 @@
...
-            continue;
-        if (wallPos.clearanceZ <= tileElement->GetBaseZ())
-            continue;
-
-        tile_element_remove_banner_entry(tileElement);
-        map_invalidate_tile_zoom1({ wallPos, tileElement->GetBaseZ(), tileElement->GetBaseZ() + 72 });
-        tile_element_remove(tileElement);
-        goto repeat;
-    } while (!(tileElement++)->IsLastForTile());
+        // Shoudn't we have a function to cast stuff back to TileElement*?
```
duncanspumpkin, line 41: TileElememt is the base class of WallElement you don't need to cast it will implicitly degrade to a TileElement pointer

src/openrct2/world/Map.cpp
```diff
@@ -2396,6 +2396,23 @@ TileElement* map_get_track_element_at_with_direction_from_ride(const CoordsXYZD&
...
 
+WallElement* map_get_wall_element_at(const CoordsXYRangedZ& coords)
+{
+    auto tileElement = map_get_first_element_at(coords);
+
+    if (tileElement != nullptr)
+    {
+        do
+        {
+            if (tileElement->GetType() == TILE_ELEMENT_TYPE_WALL && coords.baseZ < tileElement->GetClearanceZ()
```
duncanspumpkin, line 2407: I personally prefer having this split into multiple ifs and continues. I'm not overly fussed though.

src/openrct2/world/Map.cpp
```diff
@@ -2396,6 +2396,23 @@ TileElement* map_get_track_element_at_with_direction_from_ride(const CoordsXYZD&
...
+                && coords.clearanceZ > tileElement->GetBaseZ())
+                return tileElement->AsWall();
```
tupaschoal, line 2409:
```suggestion
            {
                return tileElement->AsWall();
            }
```

Let's add the braces because it looks clearer

src/openrct2/world/Wall.cpp
heliobatimarqui: It seems that TileElementBase is the base class of WallElement not TileElement. I noticed that the classes are padded to all be the same size, so while rebasing the commit I substituted it for a simple reinterpret_cast.

src/openrct2/world/Wall.cpp
heliobatimarqui: I used a for loop as I think it makes it more readable. Tell me what you think.

src/openrct2/world/Map.cpp
heliobatimarqui: Added!

src/openrct2/world/Wall.cpp
duncanspumpkin: Argh yes I forgot its still got this awkward interface. Leave it with the reinterpret_casts it will be fixed in a later refactoring.

PR #12280, opened 2020-07-15
Close #12096: Use ScreenRect on gfx_fill_rect on remaining calls

### LINKED ISSUE #12096: Use ScreenRect on gfx_fill_rect
- Create an overload of `gfx_fill_rect` with the same name, but that takes a `const ScreenRect&` object instead of `left`, `top`, `right`, `bottom`
- Replace the calls to the one using the 4 variables by the one using `ScreenRect`.

frutiemax:


tupaschoal:
Thanks!

Issue #12245, opened 2020-07-13
Use coords objects for some Window functions


tupaschoal:
Some window functions still take individual coordinate parameters. The goal is to make them use an appropriate object.

Note: Please do each of them into a different commit and PR.

- [x] - `window_scroll_to_location` should receive a `const CoordsXYZ&` instead of `x`, `y` and `z` (`src\openrct2\interface\Window.cpp`).
- [ ] - `SetLocation` should receive a `const CoordsXYZ&` instead of `x`, `y` and `z`
(`src\openrct2\interface\Window_internal.h`).

PR #12282, opened 2020-07-15
Part of #12245: Use coords objects for window_scroll_to_location

frutiemax:


src/openrct2/interface/Window_internal.cpp
```diff
@@ -4,13 +4,13 @@
...
 {
-    window_scroll_to_location(this, newX, newY, newZ);
+    window_scroll_to_location(this, { newX, newY, newZ });
     flags &= ~WF_SCROLLING_TO_LOCATION;
 }
 
 void rct_window::ScrollToViewport()
 {
-    int32_t newX = LOCATION_NULL, newY = LOCATION_NULL, newZ = LOCATION_NULL;
+    auto newCoords = CoordsXYZ{ LOCATION_NULL, LOCATION_NULL, LOCATION_NULL };
```
Gymnasiast, line 13: These values seem to get always overwritten. It's better to initialise it like so:
```suggestion
    CoordsXYZ newCoords = {};
```

src/openrct2/interface/Window_internal.cpp
```diff
@@ -4,13 +4,13 @@
...
 {
-    window_scroll_to_location(this, newX, newY, newZ);
+    window_scroll_to_location(this, { newX, newY, newZ });
     flags &= ~WF_SCROLLING_TO_LOCATION;
 }
 
 void rct_window::ScrollToViewport()
 {
-    int32_t newX = LOCATION_NULL, newY = LOCATION_NULL, newZ = LOCATION_NULL;
+    CoordsXYZ newCoords = {};
```
Gymnasiast, line 13: Move this to just above the `if (viewport_focus_sprite.type & VIEWPORT_FOCUS_TYPE_SPRITE)` statement.

Merge the `rct_window* mainWindow;` declaration with the definition on line 38.

src/openrct2/interface/Window.cpp
```diff
@@ -809,18 +809,18 @@ rct_window* window_get_main()
  * @param y (ecx)
  * @param z (edx)
  */
-void window_scroll_to_location(rct_window* w, int32_t x, int32_t y, int32_t z)
+void window_scroll_to_location(rct_window* w, const CoordsXYZ& coords)
 {
-    CoordsXYZ location_3d = { x, y, z };
+    CoordsXYZ location_3d = coords;
```
tupaschoal, line 814: This variable is not needed anymore, just use coords there instead.

Issue #12244, opened 2020-07-13
Profit calculation is wrong when using on-ride photos


chaitanyathengdi:
Below is profit and customer stats of the Shuttle Loop ride with an on-ride photo added:
![image](https://user-images.githubusercontent.com/14993220/87346946-0343b780-c570-11ea-81cc-0e964cbdd52f.png)
![image](https://user-images.githubusercontent.com/14993220/87346958-08a10200-c570-11ea-8ab4-21d7a6d2e98e.png)
These stats are around the 7-month mark, and 1 hour elapses in ~10 months.

Based on the numbers above, the profit seems to be calculated based on the assumption that every customer buys a ticket as well as a photo, whereas actual profit seems to be based on ticket OR photo(1600 tickets + 400 photos)

There should be a less-than-1 multiplier for adding profit of on-ride photos to the total profit, since not all guests end up buying photos.

tupaschoal:
Duplicate of https://github.com/OpenRCT2/OpenRCT2/issues/10946

PR #12285, opened 2020-07-16
Fix #10946: On-ride photo profit assumes every guest buys one

### LINKED ISSUE #10946: On-ride photo profit calculation assumes that every guest buys one
**OS:** Ubuntu 20.04
**Version:** 0.2.4+
**Commit/Build:** fd2e3f7b08a810d5c20466d794326e37c46f4b1e

Originally reported here: https://forums.openrct2.org/topic/4833-bugmiscalculation-with-on-ride-photos/

Quoting:

> I noticed that the calculation for On-Ride Photos makes the assumption that every single guest that rides the ride buys a Photo.
>
> Here it shows that 696 customers ride this ride per hour: https://imgur.com/a/ZtZVa3F
> Here it says that the Income per hour is $1531.20: https://imgur.com/a/F2qwt0u
>
> I had each photo priced at $2.40 which gives $2.20 profit per photo sold. 696 x 2.2 is 1531.20. So the game is calculating that every single guest is buying an On-Ride Photo. In reality 10 people out of 100 have bought an On-Ride Photo so far, so the actual income is more like $153/hour which is lower than the upkeep. The game isn't giving the extra money or anything, it just shows the wrong number for how much money is coming in. There would either need to be a "Customers that bought On-Ride Photo per hour" or after taking the "Customer" value per hour (in this case 696) it would need to be multiplied by the ratio of guests who went on the ride, compared to guests who bought an On-Ride Photo, so in this particular case 100 rode the ride, 10 bought a Photo, so the income off of On-Ride Photo's needs to be divided by 10.

chaitanyathengdi:
Use stored values of photos sold and total customers to calculate ratio
and use that to predict income per hour for rides that include
on-ride-photo sections.

Before:
photofix - before
After:
photofix - after

Instead of showing an incorrect profit which is just the addition of the two prices, it shows a ratio based on the till-date on-ride photos and ride customers.
Before profit:  $10,648.80 per hour = 1836 * (4+1.8) = 1836 * 5.8
After profit: $7610.4 per hour = 1812 * (4 + 1.8 * (19/140)) = 1812 * 4.2(rounded down)

Total customers are customers of tickets + customers of photos, so customers of photos are actually counted twice; hence the subtraction.

Reference: #12244 #10946 
Fixes: #10946


chaitanyathengdi:
This is the bare-bones version.
For now, I couldn't find anything that'll allow me to calculate a more precise ratio using stored history(similar to `num_customers` in `Ride`. Will update this branch with more code if I find that.
Feel free to merge this one in the meantime.

src/openrct2/ride/Ride.cpp
```diff
@@ -397,8 +397,20 @@ money32 Ride::CalculateIncomePerHour() const
...
     {
-        priceMinusCost += price[1];
-        priceMinusCost -= ShopItems[currentShopItem].Cost;
+        if (lifecycle_flags & RIDE_LIFECYCLE_ON_RIDE_PHOTO)
+        {
+            uint32_t photosSold = no_secondary_items_sold;
+            int32_t rideTicketsSold = total_customers - no_secondary_items_sold;
+
+            // for now, ratio is just using all-time values of tickets sold and photos sold instead of last 5 minutes'
+            float_t photosRatio = rideTicketsSold > 0 ? ((float_t)photosSold / rideTicketsSold) : 0;
```
duncanspumpkin, line 406: We avoid floats in the main codebase. You can move the divide into the next calculation to keep it in the integer world.

chaitanyathengdi: @duncanspumpkin You mean get rid of `photosRatio`?

chaitanyathengdi: I'll have to cast it to a float at least. Is there any alternative to that?

duncanspumpkin: when you do the `(photosRatio * (price[1] - ShopItems[currentShopItem].Cost))` change it to `(photosSold * (price[1] - ShopItems[currentShopItem].Cost)) / rideTicketsSold`. You will need to have it so that rideTicketsSold is set as 1 if `rideTickestSold == 0`. That way you never need to make it a float.

chaitanyathengdi: @duncanspumpkin I've done  a bit of refactoring to remove the float(basically what you've outlined). Imprecise, but it's okay I guess.

src/openrct2/ride/Ride.cpp
```diff
@@ -397,8 +397,23 @@ money32 Ride::CalculateIncomePerHour() const
...
+
+        if (lifecycle_flags & RIDE_LIFECYCLE_ON_RIDE_PHOTO)
+        {
+            uint32_t photosSold = no_secondary_items_sold;
+            int32_t rideTicketsSold = total_customers - no_secondary_items_sold;
+
+            // for now, ratio is just using all-time values of tickets sold and photos sold instead of last 5 minutes'
+            if (rideTicketsSold > 0)
+            {
+                priceMinusCost = priceMinusCost + ((photosSold * photoProfit) / rideTicketsSold);
```
Gymnasiast, line 410: Use a compound operator (+=).

Get rid of the `photosSold` variable, which is not necessary.

src/openrct2/ride/Ride.cpp
```diff
@@ -397,8 +397,23 @@ money32 Ride::CalculateIncomePerHour() const
...
+            }
+        }
+        else
+        {
+            priceMinusCost = priceMinusCost + photoProfit;
```
Gymnasiast, line 415: If the item is not an on-ride photo, then `photoProfit` is not a proper name. Rename `photoProfit` to `shopItemProfit`.

src/openrct2/ride/Ride.cpp
```diff
@@ -397,8 +397,23 @@ money32 Ride::CalculateIncomePerHour() const
 
     if (currentShopItem != SHOP_ITEM_NONE)
     {
-        priceMinusCost += price[1];
-        priceMinusCost -= ShopItems[currentShopItem].Cost;
+        money16 photoProfit = price[1] - ShopItems[currentShopItem].Cost;
+
+        if (lifecycle_flags & RIDE_LIFECYCLE_ON_RIDE_PHOTO)
```
Gymnasiast, line 402: I prefer this if to check for `shop_item_is_photo`.

src/openrct2/ride/Ride.cpp
```diff
@@ -397,8 +397,23 @@ money32 Ride::CalculateIncomePerHour() const
...
+        {
+            uint32_t photosSold = no_secondary_items_sold;
+            int32_t rideTicketsSold = total_customers - no_secondary_items_sold;
+
+            // for now, ratio is just using all-time values of tickets sold and photos sold instead of last 5 minutes'
```
Gymnasiast, line 407: This comment is stating the obvious, while it doesn't mention _why_ it uses a ratio in the first place.

```suggestion
            // Use the ratio between photo sold and total admissions to approximate the photo income (as not every guest will buy one.
            // TODO: use data from the last 5 minutes instead of all-time values for a more accurate calculation.
```

src/openrct2/ride/Ride.cpp
chaitanyathengdi: What do you mean? Do I change it?

chaitanyathengdi:
> Can you please squash all commits into a single one and reword as :
> 
> Fix #10946: On-ride photo profit assumes every guest buys one

Could you do a squash-and-merge instead?
I don't really like to force-push on my branches.

src/openrct2/ride/Ride.cpp
Gymnasiast: Yeah, use `shop_item_is_photo(currentShopItem)` instead of the current check.

chaitanyathengdi: @Gymnasiast I've updated the code.

src/openrct2/ride/Ride.cpp
```diff
@@ -397,8 +397,24 @@ money32 Ride::CalculateIncomePerHour() const
 
     if (currentShopItem != SHOP_ITEM_NONE)
     {
-        priceMinusCost += price[1];
-        priceMinusCost -= ShopItems[currentShopItem].Cost;
+        money16 shopItemProfit = price[1] - ShopItems[currentShopItem].Cost;
```
duncanspumpkin, line 400: const money16

src/openrct2/ride/Ride.cpp
```diff
@@ -397,8 +397,24 @@ money32 Ride::CalculateIncomePerHour() const
...
+
+        if (ShopItems[currentShopItem].IsPhoto())
+        {
+            int32_t rideTicketsSold = total_customers - no_secondary_items_sold;
```
duncanspumpkin, line 404: const int32_t

duncanspumpkin:
Can you add a changelog entry for this and if this is your first contribution please add yourself to contributors.md. Could you provide some before and after screenshots for a ride in this PR just for future reference. As for the actual change I think it is fine

chaitanyathengdi:
@duncanspumpkin Done as requested.

Issue #11982, opened 2020-06-17
Replace direct ride type checking


Gymnasiast:
Currently, the code still contains several explicit checks for ride types. If we want to move ride types to objects, these should be replaced. In many cases, a flag can be used instead.

RIDE_TYPE_MAZE is especially problematic. It can be left alone for now, since it's so different from everything else. Eventually, we might need to introduce an `IS_MAZE` flag,

frutiemax:
Should we move RideTypeViewOrder to a field in the RideTypeDescriptor?

Gymnasiast:
I'm not entirely sure. It's probably better to tackle other places first.

frutiemax:
```
static rct_string_id get_ride_type_name_for_dropdown(uint8_t rideType)
{
    switch (rideType)
    {
        case RIDE_TYPE_1D:
            return STR_RIDE_NAME_1D;
        case RIDE_TYPE_1F:
            return STR_RIDE_NAME_1F;
        case RIDE_TYPE_22:
            return STR_RIDE_NAME_22;
        case RIDE_TYPE_50:
            return STR_RIDE_NAME_50;
        case RIDE_TYPE_52:
            return STR_RIDE_NAME_52;
        case RIDE_TYPE_53:
            return STR_RIDE_NAME_53;
        case RIDE_TYPE_54:
            return STR_RIDE_NAME_54;
        case RIDE_TYPE_55:
            return STR_RIDE_NAME_55;
        case RIDE_TYPE_59:
            return STR_RIDE_NAME_59;
        default:
            return RideTypeDescriptors[rideType].Naming.Name;
    }
}
```

Is there any way to avoid the call to RideTypeDescriptors table there?

pizza2004:
I'm pretty sure the only reason the rest of that switch block exists is because those ride types don't have any actual data associated with them and the call to RideTypeDescriptors is the desired behavior.

pizza2004:
I imagine if you passed in the ride and then used the method you were using on those commits instead of passing in just the ride type you could avoid the call.

Gymnasiast:
> Is there any way to avoid the call to RideTypeDescriptors table there?

@frutiemax Why would you want to do that? It's not the call to the RideTypeDescriptors table that's the problem there, it's the direct checks of RIDE_TYPE_x - which are there because we haven't removed these dummy ride types yet.

frutiemax:
For the RideTypeView order, where should we place the field in the RideTypeDescriptor?

PR #12287, opened 2020-07-16
Part of #11982: Replace direct ride type checking #1

frutiemax:


Gymnasiast:
I don't think you really understood the gist of #11982. The idea of that is that we replace checks like `if (rideType == RIDE_TYPE_MINIATURE_RAILWAY || rideType == MONORAIL)` with an alternative.

Replacing calls to RideTypeDescriptors with calls to the ride itself to obtain an RTD is another project. But that bit is already being worked on in https://github.com/OpenRCT2/OpenRCT2/pull/11994. It also has to be noted that if RTD end up working like any other object, it will probably need null checks as well. All of all, it's better to leave this refactor for now.

PR #12288, opened 2020-07-16
Close #12245: Use coords objects for SetLocation (Window_internal.h)

### LINKED ISSUE #12245: Use coords objects for some Window functions

frutiemax:


src/openrct2/Editor.cpp
```diff
@@ -87,7 +87,7 @@ namespace Editor
         gS6Info.category = SCENARIO_CATEGORY_OTHER;
         viewport_init_all();
         rct_window* mainWindow = context_open_window_view(WV_EDITOR_MAIN);
-        mainWindow->SetLocation(2400, 2400, 112);
+        mainWindow->SetLocation({ 2400, 2400, 112 });
```
Gymnasiast, line 90:
```suggestion
        mainWindow->SetLocation(TileCoordsXYZ{ 75, 75, 14 }.ToCoordsXYZ());
```

Same for the other calls.

PR #12289, opened 2020-07-16
Use CoordsXYZ for sub_98199C in RiverRapids.cpp

frutiemax:


frutiemax:
testpaint doing its own thing...

src/openrct2/paint/Paint.cpp
```diff
@@ -986,6 +979,16 @@ paint_struct* sub_98199C(
...
+paint_struct* sub_98199C(
+    paint_session* session, uint32_t image_id, int8_t x_offset, int8_t y_offset, int16_t bound_box_length_x,
+    int16_t bound_box_length_y, int8_t bound_box_length_z, int16_t z_offset, int16_t bound_box_offset_x,
+    int16_t bound_box_offset_y, int16_t bound_box_offset_z)
+{
+    return sub_98199C(
+        session, image_id, { x_offset, y_offset, z_offset }, { bound_box_length_x, bound_box_length_y, bound_box_length_z },
+        { bound_box_offset_x, bound_box_offset_y, bound_box_offset_z });
+}
+
```
tupaschoal, line 991: Are there other users of this function? I don't get why this is being kept

src/openrct2/paint/Paint.cpp
```diff
@@ -954,24 +954,17 @@ paint_struct* sub_98198C(
...
  */
 paint_struct* sub_98199C(
-    paint_session* session, uint32_t image_id, int8_t x_offset, int8_t y_offset, int16_t bound_box_length_x,
-    int16_t bound_box_length_y, int8_t bound_box_length_z, int16_t z_offset, int16_t bound_box_offset_x,
-    int16_t bound_box_offset_y, int16_t bound_box_offset_z)
+    paint_session* session, uint32_t image_id, const CoordsXYZ& offset, const CoordsXYZ& boundBoxLength,
+    const CoordsXYZ& boundBoxOffset)
 {
-    assert(static_cast<uint16_t>(bound_box_length_x) == static_cast<int16_t>(bound_box_length_x));
-    assert(static_cast<uint16_t>(bound_box_length_y) == static_cast<int16_t>(bound_box_length_y));
```
tupaschoal, line 962: Why removing the asserts?

Gymnasiast: Well, are they needed now we use signed 32-bit integers?

tupaschoal: `sub_98197C` below still take `int16_t`s so i'd say it's safer to keep them for now

src/openrct2/paint/Paint.cpp
frutiemax: There are still a lot of remaining calls, too many for one PR.

duncanspumpkin:
Looks fine to me.

PR #11893, opened 2020-06-07
Part of #11159 (purge malloc() & free())

rdbaris:
Replacing uses of malloc and free in the code with new and delete in src/libopenrct2

PR #12293, opened 2020-07-16
Fix for smart pointer bug

rdbaris:
Resolving an issue found in #11893 that came up when adding smart pointers

### DIFF
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -103,8 +103,8 @@ static bool sprite_file_open(const utf8* path)
             fclose(file);
             return false;
         }
+        spriteFileData = new uint8_t(spriteFileHeader.total_size);
 
-        spriteFileData = static_cast<uint8_t*>(malloc(spriteFileHeader.total_size));
         if (fread(spriteFileData, spriteFileHeader.total_size, 1, file) != 1)
         {
             free(spriteFileData);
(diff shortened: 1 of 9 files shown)


src/openrct2/CmdlineSprite.cpp
```diff
@@ -90,17 +90,16 @@ static bool sprite_file_open(const utf8* path)
...
+        std::unique_ptr<rct_g1_element_32bit[]> openElements = std::make_unique<rct_g1_element_32bit[]>(
+            spriteFileHeader.num_entries);
         if (openElements == nullptr)
         {
             fclose(file);
             return false;
         }
 
-        if (fread(openElements, openEntryTableSize, 1, file) != 1)
+        if (fread(openElements.get(), spriteFileHeader.num_entries, 1, file) != 1)
```
ZehMatt, line 101: This is now incorrect, fread expects the total amount of bytes as the parameter, this passes only the amount of entries.

rdbaris: Sorry about that! This should fix it

ZehMatt:
Looks about right now, this branch needs to rebased and it should be good to go.

tupaschoal:
This page explains about sync and rebase @rdbaris 
https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2

PR #12300, opened 2020-07-17
Part of #12243: Expand usage of Coords objects for maze_set_track

frutiemax:


src/openrct2/ride/Ride.cpp
```diff
@@ -1276,34 +1276,31 @@ void ride_remove_provisional_track_piece()
...
+        maze_set_track({ coords, 0 }, flags, false, rideIndex, GC_SET_MAZE_TRACK_FILL);
+        maze_set_track({ coords + CoordsXYZ{ 0, 16, 0 }, 1 }, flags, false, rideIndex, GC_SET_MAZE_TRACK_FILL);
+        maze_set_track({ coords + CoordsXYZ{ 16, 16, 0 }, 2 }, flags, false, rideIndex, GC_SET_MAZE_TRACK_FILL);
+        maze_set_track({ coords + CoordsXYZ{ 16, 0, 0 }, 3 }, flags, false, rideIndex, GC_SET_MAZE_TRACK_FILL);
     }
     else
     {
-        int32_t direction = _unkF440C5.direction;
-        if (!(direction & 4))
+        if (!(coords.direction & 4))
```
duncanspumpkin, line 1291: _unkF440C5 should not have been converted to a CoordsXYZD. Can you change it to a CoordsXYZ with the direction component separate.

frutiemax: I haven't converted it, it was already a CoordsXYZD.

frutiemax: How do I get the direction if _unkFF440C5 doesn't have it?

duncanspumpkin: You would need a separate variable just for direction. Pretty sure this is a global just add another global for now.

src/openrct2/ride/Ride.h
```diff
@@ -1071,7 +1071,8 @@ extern CoordsXYZ _previousTrackPiece;
 extern uint8_t _currentBrakeSpeed2;
 extern uint8_t _currentSeatRotationAngle;
 
-extern CoordsXYZD _unkF440C5;
+extern CoordsXYZ _unkF440C5;
+extern Direction _unkF440C5_Direction;
```
duncanspumpkin, line 1075: Type should be uint8_t Direction implies it's just 0-3

frutiemax:
I might be mistaking but it looks like it holds the coordinates of the current position of the maze in construction.

duncanspumpkin:
The direction split off is for two reasons. One is that direction is not the Cardinal directions for tracks and secondly because it isn't being used like it describes a tile.

Gymnasiast:
I don't think I understand this. It moves several things to CoordsXYZD, but _unkF440C5 is moved from CoordsXYZD to CoordsXYZ + a separate direction?

Gymnasiast:
I really can't see why the direction would have to be split off. Especially since it gets fed into CoordsXYZDs, which seems to contradict the point that "the value can be higher than 3".

duncanspumpkin:
`ride_remove_provisional_track_piece` uses values higher than 3 so it should not be a `Direction`. Tracks always use >3 for its direction due to the diagonal pieces its going to really confuse things if we have Direction to mean < 4 but also  < 8 depending on if its a track or not.

Gymnasiast:
We should add a `using` for that, then.

ZehMatt:
I'm also with @Gymnasiast, what bothers me a bit is that this PR introduces a new global variable where we all know this is never a good thing.

Gymnasiast:
@ZehMatt It doesn't so much add one; it splits an existing one in two.

ZehMatt:
When one becomes two I would consider that an addition but I get the point

Gymnasiast:
Since this has stalled for more than a year, closing it.

PR #7345, opened 2018-03-24
Refactor cheats window widget definitions

AaronVanGeffen:
This refactors the cheats window to use groupbox-based offsets for its widget positions, rather than its custom macro system. A major benefit is widget flexibility without resorting to hacky positioning. This was previously seen in the use of decimal grid positions and added pixel offsets.

On top of this, by making the window adhere to the conventions used in the rest of the code base, this PR hopefully increases legibility without compromising widget mobility.

I hope to do some more redesigning of the cheats window in another PR after this.

PR #12303, opened 2020-07-17
Refactor cheats window

AaronVanGeffen:
This moves the cheats window into the new window framework. While different from my original approach in #7345, this has been something I've wanted to see for years.

The result is a bunch of widget code that is, to me, much more intuitive to read than the macros unique to this window.

I've marked this PR as a draft, however, as I think there is room for improvement. All buttons and checkboxes in the cheat window have (or should have) the same dimensions, for example. We can extract these to constants and reuse them.

Thoughts and suggestions are welcome.

### DIFF
--- a/src/openrct2-ui/windows/Cheats.cpp
+++ b/src/openrct2-ui/windows/Cheats.cpp
@@ -180,6 +180,11 @@ static constexpr const rct_string_id WINDOW_TITLE = STR_CHEAT_TITLE;
 static constexpr const int32_t WW = 249;
 static constexpr const int32_t WH = 300;
 
+static constexpr ScreenSize CHEAT_BUTTON = {110, 17};
+static constexpr ScreenSize CHEAT_CHECK = {221, 12};
+static constexpr ScreenSize CHEAT_SPINNER = {117, 14};
+static constexpr ScreenSize MINMAX_BUTTON = {55, 17};
+
 #pragma endregion
 
 #define MAIN_CHEATS_WIDGETS \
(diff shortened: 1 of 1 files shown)


src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -179,147 +179,123 @@ enum WINDOW_CHEATS_WIDGET_IDX
...
-    { WWT_BUTTON,           1,      XPL(1),                 WPL(1),                 YPL(16),        HPL(16),        STR_SHOP_ITEM_PLURAL_UMBRELLA,      STR_NONE },                             // give guests umbrellas
-    { WWT_BUTTON,           1,      XPL(0),                 WPL(0),                 YPL(18),        HPL(18),        STR_CHEAT_LARGE_TRAM_GUESTS,        STR_CHEAT_LARGE_TRAM_GUESTS_TIP },      // large tram
-    { WWT_BUTTON,           1,      XPL(1),                 WPL(1),                 YPL(18),        HPL(18),        STR_CHEAT_REMOVE_ALL_GUESTS,        STR_CHEAT_REMOVE_ALL_GUESTS_TIP },      // remove all guests
-    { WWT_BUTTON,           1,      XPL(0),                 WPL(0),                 YPL(19),        HPL(19),        STR_CHEAT_EXPLODE,                  STR_CHEAT_EXPLODE_TIP },                // explode guests
+    MakeWidget({  5,  48}, {238, 279}, WWT_GROUPBOX, 1, STR_CHEAT_SET_GUESTS_PARAMETERS                                 ), // Guests parameters group frame
+    MakeWidget({183,  69}, { 54,  17}, WWT_BUTTON,   1, STR_MAX                                                         ), // happiness max
```
Gymnasiast, line 215: Am I correct in assuming the width alternates between 54 and 56? Wouldn't it be better to just make everything 55?

AaronVanGeffen: Yes, I want to streamline these. I haven't changed any of the values from what they are on current develop.

AaronVanGeffen:
I have added a few constants for the most common widget sizes in the cheat window. Consequently, I have harmonised the buttons that previously deviated +/- 1px in width, depending on what column they were in. The only other change that is reflected in the window's appearance, is the increase of the height of the spinner widgets:  the +/- buttons are now square, like in other windows.

The keen-eyed may notice that I have not added a constant for the tab and dropdown widgets, even though they have the same width. In a follow-up PR, I would like to add functions `MakeTab` and `MakeDropdown` to make adding these widgets easier. We could shorten

```cpp
MakeRemapWidget({ 3, 17}, {31,  27}, WWT_TAB,    1, SPR_TAB,    STR_FINANCIAL_CHEATS_TIP)
```
to exclude common parameters to something like:
```
MakeTab({ 3, 17}, STR_FINANCIAL_CHEATS_TIP)
```

Similarly, the `MakeDropdown` function would add a `WWT_DROPDOWN` and accompanying `WWT_BUTTON` widget, much like `MakeSpinnerWidgets` already operates. Speaking of `MakeSpinnerWidgets`, I intend to rename this to `MakeSpinner`.

Any feedback for this PR and the musings above are welcome, of course.

src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -1204,56 +1185,35 @@ static void window_cheats_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
-        gfx_draw_string_left(
-            dpi, STR_YEAR, nullptr, COLOUR_BLACK, w->windowPos + ScreenCoordsXY{ XPL(0) + TXTO, YPL(7) + TXTO });
-        gfx_draw_string_left(
-            dpi, STR_MONTH, nullptr, COLOUR_BLACK, w->windowPos + ScreenCoordsXY{ XPL(0) + TXTO, YPL(8) + TXTO });
-        gfx_draw_string_left(
-            dpi, STR_DAY, nullptr, COLOUR_BLACK, w->windowPos + ScreenCoordsXY{ XPL(0) + TXTO, YPL(9) + TXTO });
-        gfx_draw_string_right(
-            dpi, STR_FORMAT_INTEGER, &_yearSpinnerValue, w->colours[1],
-            w->windowPos + ScreenCoordsXY{ WPL(1) - 34 - TXTO, YPL(7) + TXTO });
+        gfx_draw_string_left(dpi, STR_BOTTOM_TOOLBAR_CASH, gCommonFormatArgs, colour, w->windowPos + ScreenCoordsXY{ 14, 93 });
```
Gymnasiast, line 1188: I see this number `14` a lot in this PR. There must be some significance to it, so defining a constant would be appropriate IMO.

AaronVanGeffen: Shall we introduce an `ANCHOR_LCOL_X` = 14 and `ANCHOR_RCOL_X` = 208? They seem to be restricted to `window_cheats_paint`, so I propose we make them local constants there.

Gymnasiast: Local constants would be fine, if it is really unique to the Cheats window.

AaronVanGeffen: I meant local to the paint function, even, not just the file. Since the value doesn't occur in the same way elsewhere in the file, it makes sense to optimise this locally.

Issue #12312, opened 2020-07-19
Game soft-locks when loading a save file via command line fails


ZehMatt:
Something I noticed as well is that if you start OpenRCT2 via command line with a file that it unsuccessfully opens, you just get a black soft locked screen.

_Originally posted by @IntelOrca in https://github.com/OpenRCT2/OpenRCT2/issues/12229#issuecomment-657631998_

janisozaur:
I complained about the black screen years ago. What does "soft locked" mean, though?

ZehMatt:
It means you can no longer interact with the game being able to do nothing.

IntelOrca:
@janisozaur https://en.wiktionary.org/wiki/softlock

You may still be able to interact with the game, but it is impossible to advance. The game is stuck but it has not crashed and is still running through the game loop.

PR #12313, opened 2020-07-19
#12229: Show error message when loading save from command line fails

### LINKED ISSUE #12229: Show proper warning when unsuccesfully trying to load a save
Currently, no warning is shown in the GUI when you try to load a save (or a scenario via the load/save window), only in the console. In cases like this, RCT2 would warn "file contains invalid data".

Additionally, when fixing this, it would be nice if loading renamed .sea files would have an error message of its own.

ZehMatt:
This PR also fixes the issue that it would soft-lock on load failure.

Closes #12229, #12312


duncanspumpkin:
This PR removes specific error messages from the loader as it will throw on a load failure.

PR #12323, opened 2020-07-19
Close #12270 Remove goto from track_remove_station_element

### LINKED ISSUE #12270: Remove goto from track_remove_station_element
https://github.com/OpenRCT2/OpenRCT2/blob/0951c0c5cc3f69a6097251ddc6bdb6047f8f7959/src/openrct2/ride/Track.cpp#L927-L930

0ro8lu:
Remove use of goto from function track_remove_station_element

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -3154,7 +3154,7 @@ static void window_ride_vehicle_scrollpaint(rct_window* w, rct_drawpixelinfo* dp
     rct_ride_entry* rideEntry = ride->GetRideEntry();
 
     // Background
-    gfx_fill_rect(dpi, dpi->x, dpi->y, dpi->x + dpi->width, dpi->y + dpi->height, PALETTE_INDEX_12);
+    gfx_fill_rect(dpi, { { dpi->x, dpi->y }, { dpi->x + dpi->width, dpi->y + dpi->height } }, PALETTE_INDEX_12);
 
     rct_widget* widget = &window_ride_vehicle_widgets[WIDX_VEHICLE_TRAINS_PREVIEW];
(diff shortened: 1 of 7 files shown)


0ro8lu:
> Please rebase instead of merging one branch into another: [Rebase and Sync fork with OpenRCT2](https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2)
Ok, i've synced my forked version of the repo with the main devlop branch. So when in CLion i do pull from develop it's not actually updating it? How does it work?

tupaschoal:
When you sync your develop, it's only it that is changing. You then have to either rebase your branch, `track_remove_station_element` in this case, on top of the "new" `develop`, or merge the `develop` onto it.

What you seem to have done is the latter, but we rather prefer the first, because as you can see, merging leaves the PR with lots of commits, which kind of pollutes it a bit.

Remember that when you create your branch, it takes a snapshot from where it's being branched from and, from there onwards, it is a different thing altogether, which needs manual/separate sync.

0ro8lu:
Mh... how about removing this PR? I’ll redo the modification without polluting the PR with tons of commits

duncanspumpkin:
Remaking PRs is a real pain and doing that every time you make a mistake would lead to 100s of PRs. It's really not difficult to do a rebase.

0ro8lu:
Sorry, i thought it would be a valid solution, i did not want to look lazy. I've rebased it (hopefully)

duncanspumpkin:
I see now why you were talking about the do while being horrible to look at. This is indeed a horrible to follow function. It isn't a tile element loop so I have no issues with changing it away from being a do while into something completely different. But lets leave that for a second refactoring.

duncanspumpkin:
Please ensure for future work that you use rebase correctly its still not correct. I'll merge this for now but for future. https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2

PR #12326, opened 2020-07-19
Close #12078: New shortcut key for sloping walls

### LINKED ISSUE #12078: New shortcut key request: sloping walls
When making buildings, or support structures, slopeable objects are great to use, especially for NCSO builders. 

The tile inspector has a function to slope these walls, but it takes quite a bit of time to do them one by one. It would be great if there was a shortcut key to do that.

In this case, there would need to be three options for setting the shortcut keys:
- Flat
- Right side up
- Left side up

tileinspector

frutiemax:
The shortcut key works like this:

1.  When the scenery window opens, the normal wall slope is used
2.  When the shortcut key is pressed once, it forces to use the edge slope = 1
3.  When the shortcut key is pressed again, it forces to use the edge slope = 2
4.  When the shortcut key is pressed again, the normal wall slope is used


tupaschoal:
Can you please clarify your intent @frutiemax ?

frutiemax:
Original issue description:
The tile inspector has a function to slope these walls, but it takes quite a bit of time to do them one by one. **It would be great if there was a shortcut key to do that.**

tupaschoal:
@Broxzier please let us know if OP can move forward or not. If so, this needs rebase

tupaschoal:
I've talked to @Broxzier and decided to close this, sorry for taking so long

PR #12333, opened 2020-07-20
Remove assert on ScreenLine

tupaschoal:
Code is not always constructing a valid line as pointed out by https://github.com/OpenRCT2/OpenRCT2/commit/f725fff2dad526b55f3bc5386f54ef0d565f99ec#r40758993


tupaschoal:
> Yeah it's unfortunate we can't have this as it's a sensible idea but oh well.

I imagined we would hit a case eventually, but it's a "we can't have this **yet**", maybe someday we will :)

PR #12339, opened 2020-07-21
Fix/consistent ride naming

Gymnasiast:


Umdlye:
Just curious, what is the convention for ride names in the code? The ship ride is called "pirate ship" in most languages including English, but the object file name is `swsh1`.

Gymnasiast:
RCT1 called it Swinging Ship, and many object file names were derived from how RCT1 called it, even if RCT2 changed it.

It should be noted that "Swinging Ship" is used for the ride type here, not for the vehicles that use it. I changed it to "Swinging Ship" because it's more generic, and not every vehicle for this ride type will be a pirate ship - arguably, not even the default vehicle!

Issue #12166, opened 2020-07-08
The Great Ride Draw Refactor: Track Flat


duncanspumpkin:
The ride drawing functions have not had much attention ever since they were automatically generated.
There is much scope for refactoring. This issue is for the Track Flat drawing functions.

First of all an example of the issues with `track_flat`
We will be using the LayDownRollerCoaster as the example as it is quite a complex coaster so should be relavant for most other coasters.
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/coaster/LayDownRollerCoaster.cpp
```
switch (direction)
{
    case 0:
        sub_98197C_rotated(
            session, direction, session->TrackColours[SCHEME_TRACK] | 16226, 0, 0, 32, 20, 3, height, 0, 6, height);
        break;
    case....
}
```
The first switch is not needed. The only difference for each step is the direction and the image number. We can just pass that straight into `sub_98197C` and use an array for the image number.

I.e. this is the same:
```
uint32_t flatLiftSprites[4] = { 16226, 16227, 16228, 16229 } 
sub_98197C_rotated(
      session, direction, session->TrackColours[SCHEME_TRACK] | flatLiftSprites[direction], 0, 0, 32, 20, 3, height, 0, 6, height);
```

The next switch at L58 can also be refactored with the same style:
```
uint32_t flatSprites[4] = { 16224, 16225, 16224, 16225 } 
sub_98197C_rotated(
      session, direction, session->TrackColours[SCHEME_TRACK] | flatSprites[direction], 0, 0, 32, 20, 3, height, 0, 6, height);
```
This can be repeated on the next two switches as well!

Notice how the array has been named differently for each part. The laydown coaster has flat, flat with a lift chain, inverted, inverted with a chain. We want to use this naming in our arrays. You might come across some functions that do not refer to the sprites individually and do a sum such as `direction + 16226`. We want this to be turned into an array like the above as well.

The above refactoring can be applied to all of the following coasters:
If you are wanting to perform this refactoring then please start with just one file and create a PR. It will be much much easier to review and it means we can get it merged quicker.

- [x] air_powered_vertical_rc_track_flat
- [ ] bobsleigh_rc_track_flat (Take care its slightly different to the above)
- [ ] bolliger_mabillard_track_flat
- [ ] compact_inverted_rc_track_flat
- [ ] corkscrew_rc_track_flat
- [ ] flying_rc_track_flat
[...]

PR #12341, opened 2020-07-21
Add using definitions for the 4 different styles of track flat

duncanspumpkin:
This is for #12166. Names are still tbc but this was my general idea.

### DIFF
--- a/src/openrct2/ride/TrackPaint.h
+++ b/src/openrct2/ride/TrackPaint.h
@@ -15,6 +15,8 @@
 #include "../paint/tile_element/Paint.TileElement.h"
 #include "../world/Map.h"
 
+#include <array>
+
 class StationObject;
 
 extern const uint8_t track_map_2x2[][4];
@@ -27,6 +29,21 @@ extern const uint8_t track_map_4x4[][16];
 extern const uint8_t edges_4x4[];
 
 extern const uint8_t track_map_1x4[][4];
+
+namespace TrackPaint::Flat
+{
+    // Lift and No lift sprites
+    using BasicSprites = std::array<std::array<uint16_t, NumOrthogonalDirections>, 2>;
+
+    // No lift sprites
+    using NoLiftSprites = std::array<uint16_t, NumOrthogonalDirections>;
+
+    // Lift and No Lift (sprites * 2 {front, back})
+    using ComplexSprites = std::array<std::array<std::array<uint16_t, 2>, NumOrthogonalDirections>, 2>;
+
+    // Inverted and Not Inverted (Lift and No Lift) sprites
+    using InvertedSprites = std::array<std::array<std::array<uint16_t, NumOrthogonalDirections>, 2>, 2>;
+} // namespace TrackPaint::Flat
 
 enum
 {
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/TrackPaint.h
```diff
@@ -28,6 +30,21 @@ extern const uint8_t edges_4x4[];
...
+namespace TrackPaint::Flat
+{
+    // Lift and No lift sprites
+    using BasicSprites = std::array<std::array<uint16_t, NumOrthogonalDirections>, 2>;
+
+    // No lift sprites
+    using NoLiftSprites = std::array<uint16_t, NumOrthogonalDirections>;
+
+    // Lift and No Lift (sprites * 2 {front, back})
+    using ComplexSprites = std::array<std::array<std::array<uint16_t, 2>, NumOrthogonalDirections>, 2>;
```
Gymnasiast, line 42: Can you give an example of this?

duncanspumpkin: Bobsleigh or do you mean in my comment give an example

ZehMatt:
I'm not a big fan of having array within array honestly, can't we make this one continuous array and simply have access by row*col + i. This memory access pattern is not the best practice.

duncanspumpkin:
Under the hood this would work like one continuous array. https://godbolt.org/z/Eo4j5o

ZehMatt:
> Under the hood this would work like one continuous array. https://godbolt.org/z/o5TWY5

This only applies for smarter compilers such as Clang, MSVC is garbage in this regard.

duncanspumpkin:
If the array is const then all compilers store it as contiguous memory afaik and with optimisations then it all becomes pretty much the same.

ZehMatt:
Fair enough, no objections.

src/openrct2/ride/TrackPaint.h
Gymnasiast: You mean that every piece consists of two sprites, or?

duncanspumpkin: yes

tupaschoal:
@Gymnasiast can you review? I think that's what is missing to be merged

src/openrct2/ride/TrackPaint.h
```diff
@@ -28,6 +30,21 @@ extern const uint8_t edges_4x4[];
 
 extern const uint8_t track_map_1x4[][4];
 
+namespace TrackPaint::Flat
+{
+    // Lift and No lift sprites
```
Gymnasiast, line 35: I think it should start with "no lift" and then continue with yes lift. Same for uninverted/inverted. The basic idea is to start with the simple/"undecorated" first.

src/openrct2/ride/TrackPaint.h
```diff
@@ -28,6 +30,21 @@ extern const uint8_t edges_4x4[];
...
+    using BasicSprites = std::array<std::array<uint16_t, NumOrthogonalDirections>, 2>;
+
+    // No lift sprites
+    using NoLiftSprites = std::array<uint16_t, NumOrthogonalDirections>;
+
+    // Lift and No Lift (sprites * 2 {front, back})
+    using ComplexSprites = std::array<std::array<std::array<uint16_t, 2>, NumOrthogonalDirections>, 2>;
+
+    // Inverted and Not Inverted (Lift and No Lift) sprites
```
Gymnasiast, line 44: Needs a bit of clarification as to what is the outer and what in the inner array.

duncanspumpkin: Yes thats a good idea I'll make it nested BasicSprites.

PR #12344, opened 2020-07-21
Close #12335: Eliminate goto statement in footpath_chain_ride_queue()

### LINKED ISSUE #12335: Remove goto from footpath_chain_ride_queue
https://github.com/OpenRCT2/OpenRCT2/blob/d10b615eb33f6cd597340d83d68eaaa97c1a493e/src/openrct2/world/Footpath.cpp#L1093-L1100

Ron4u:
please verify and make changes wherever possible.
i am not familiar with clang-format. i tried my best to come up with a bug fix.
suggest changes if required

### DIFF
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -1082,6 +1082,7 @@ void footpath_chain_ride_queue(
 
         auto targetQueuePos = curQueuePos + CoordsDirectionDelta[direction];
         tileElement = map_get_first_element_at(targetQueuePos);
+        bool move = false;
         if (tileElement != nullptr)
         {
             do
@@ -1097,7 +1098,9 @@ void footpath_chain_ride_queue(
                         if (tileElement->AsPath()->GetSlopeDirection() != direction)
                             break;
                     }
-                    goto foundNextPath;
+                    move = true;
+                    break;
+                    
                 }
                 if (tileElement->GetBaseZ() == baseZ - LAND_HEIGHT_STEP)
                 {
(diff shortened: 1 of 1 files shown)


src/openrct2/world/Footpath.cpp
```diff
@@ -1090,13 +1093,15 @@ void footpath_chain_ride_queue(
...
 
                     baseZ -= LAND_HEIGHT_STEP;
-                    goto foundNextPath;
+                    move = true;
+                    break;
                 }
             } while (!(tileElement++)->IsLastForTile());
         }
+        if(!move)
         break;
```
tupaschoal, line 1103:
```suggestion
        if (!move)
            break;
```

src/openrct2/world/Footpath.cpp
```diff
@@ -1064,6 +1064,7 @@ void footpath_chain_ride_queue(
 
         auto targetQueuePos = curQueuePos + CoordsDirectionDelta[direction];
         tileElement = map_get_first_element_at(targetQueuePos);
+        bool move = false;
```
Broxzier, line 1067: Please use a more descriptive identifier that explains what it means when `move` is true or false. From what I can see from quickly looking at the code, this is used to verify that a queue was found and can be connected.

src/openrct2/world/Footpath.cpp
Ron4u: Thank you for the suggestion

src/openrct2/world/Footpath.cpp
Ron4u: Yes, earlier in the code goto statement was used to connect it to queue but there is no such requirement we can just do this by using simple break statements and apply condition checkings at point.

Broxzier: If that's the case, then a better identifier would be `foundQueue`, or something similar. Then the statement `if (!foundQueue)` makes immediate sense without having to know how the value was set.

Ron4u: Noted. Will create a pr with foundQueue variable. Thank you

ZehMatt: This should be changed in this PR not in new one, this PR will not be merged unless we are happy with the code. You can just add more commits on top of this PR.

tupaschoal:
@Ron4u I'm converting this to a draft PR, please tend to the comments and the failures and when it's good to go mark it as ready to review

PR #12345, opened 2020-07-21
Fix research import/export

duncanspumpkin:
Old version incorrectly tried to convert the research type not the base ride type causing crashes when new saves were used on older versions of OpenRCT2 or vanilla.


duncanspumpkin:
I'm merging this without review as it's critical and it's child pr had been reviewed but has some discussion still to play out.

Gymnasiast:
Thanks for fixing this. The code looks good, it's what I intended to do anyway. Not sure how I missed that. :/

PR #12346, opened 2020-07-21
Refactor ResearchItem Type to an enum class

duncanspumpkin:
Continuing on from #12345 this prevents any future issues by using a proper type for the item.

### DIFF
--- a/src/openrct2/management/Research.h
+++ b/src/openrct2/management/Research.h
@@ -17,10 +17,10 @@
 
 struct rct_ride_entry;
 
-enum
+enum class ResearchEntryType
 {
-    RESEARCH_ENTRY_TYPE_SCENERY = 0,
-    RESEARCH_ENTRY_TYPE_RIDE = 1,
+    Scenery = 0,
+    Ride = 1,
 };
 
 enum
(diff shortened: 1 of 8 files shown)


src/openrct2/management/Research.h
```diff
@@ -17,10 +17,10 @@
...
 struct rct_ride_entry;
 
-enum
+enum class ResearchEntryType : uint8_t
 {
-    RESEARCH_ENTRY_TYPE_SCENERY = 0,
-    RESEARCH_ENTRY_TYPE_RIDE = 1,
+    Scenery = 0,
+    Ride = 1,
 };
```
tupaschoal, line 24: If it's not much to ask, I really think we should start using namespaces more often, I'd do:
```c++
namespace Research
{
    enum class EntryType : uint8_t
    {
        Scenery = 0,
        Ride = 1,
    };
}
```

Usage would be then:
```c++
    Research::EntryType::Scenery
```

duncanspumpkin: Yes i was thinking that as well but shouldn't it be Management::Research::EntryType?

tupaschoal: Oops, I didn't reply fast enough. I think both are fine

PR #12347, opened 2020-07-21
Pester users to update every so often

janisozaur:
Every 24h check if latest release on github is user's current one. If it isn't display announcement. The check is triggered on park load and last check time is saved in config, so you can opt out of it by setting the value to some time in the future.


src/openrct2/Version.cpp
```diff
@@ -9,17 +9,30 @@
...
+#include "core/Http.h"
+#include "core/Json.hpp"
+
+#include <chrono>
 
 #ifdef OPENRCT2_BUILD_INFO_HEADER
 #    include OPENRCT2_BUILD_INFO_HEADER
 #endif
 
+const char gVersionInfoTag[] =
```
janisozaur, line 23: I made some tests by spoofing this value

src/openrct2/Version.cpp
```diff
@@ -39,3 +52,46 @@ const char gVersionInfoFull[] = OPENRCT2_NAME ", "
...
+    // If the check doesn't succeed, provide current version so we don't bother user
+    // with invalid data.
+    std::string version = gVersionInfoTag;
+#ifndef DISABLE_HTTP
+    auto now = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+    auto then = gConfigGeneral.last_version_check_time;
+    if (then < now - 24 * 60 * 60)
+    {
+        Http::Request request;
+        request.url = "https://api.github.com/repos/OpenRCT2/OpenRCT2/releases/latest";
```
janisozaur, line 67: https://developer.github.com/v3/repos/releases/#get-the-latest-release

ocalhoun6:
Will this make OpenRCT2 slower to start up for people with slow internet connections?

janisozaur:
@ocalhoun6 no. The version check is done asynchronously. But it's worth testing the PR with internet disabled and impaired.

@tupaschoal you can opt out already: just set a future date in config. Set it to 9 billion or prepend the current value with 9.

janisozaur:
I also feel the current notification is inadequate.

Other ideas I have include:
* Posting a notification to the bottom bar (just like awards)
* Adding an item to floppy menu, clicking which could open the URL for the release

duncanspumpkin:
I think showing a message on the title screen would be best. I assume ultimately this is the start of taking responsibility of the auto updater?

janisozaur:
The problem with showing it on title screen is I can see some users not viewing it at all (e.g. by launching a park from a file association) or spending little time there. It could accompany the notification in-game, but in my opinion it shouldn't be the only one.

If it were to be shown on title screen, how would you see it done? Another button next to 4 main ones?

I have no plans to extend this to auto-updater, this is the final form of this feature as I see it.

IntelOrca:
> If it were to be shown on title screen, how would you see it done? Another button next to 4 main ones?

The same as the changelog that shows when you update the game. I doubt many people enter the game without going through the title screen. If they do, it won't be necessarily every single time they play.

ocalhoun6:
> The problem with showing it on title screen is I can see some users not viewing it at all (e.g. by launching a park from a file association) or spending little time there. It could accompany the notification in-game, but in my opinion it shouldn't be the only one.

As long as the title screen notification is very visible, I think that's enough.

And remember -- the more annoying the notification, the more likely users are to configure the game to ignore it. I could see an in-scenario notification being pretty annoying.

janisozaur:
I have some ideas what to do next:

1. Store the data obtained from GitHub inside context for future reference
2. The data contains release info we put there, so I can actually show the changelog to the user. I guess the window should have some button "go to release page" and somehow differentiate itself from the "you're on new version changelog" so it's clear OpenRCT2 did **not** update
3. Add a button to bottom of the floppy menu linking to release page/changelog window
4. Add the update button below the 4 ones on title screen, spanning their width again linking to changelog window.
5. Add notification of a new kind with same behaviour as other buttons

I think I would only show such changelog window after user input, so as not to annoy users with large pop-ups and place less annoying hints throughout the game.

AaronVanGeffen:
> Add a button to bottom of the floppy menu linking to release page/changelog window

The about window is a better fit for this, IMO. The changelog window is already linked there.

> Add the update button below the 4 ones on title screen, spanning their width again linking to changelog window.

Let's not, please. I'd be happy to have it go into the toolbox, though.

janisozaur:
I've improved checking for the release - it is now stored as part of the context and fetched as soon as the game starts. I've also separated out the two smaller commits to their own PRs. There's still UI to be done now.

tupaschoal:
Seems like this PR could use a rebase to lose the first two commits, that were already merged.

janisozaur:
Sure, I will do that once I have more UI stuff done.

janisozaur:
I've added new version info window, mostly copied from changelog, but I want to change some things:
* title should read new version tag/name
* there should be a button to open URL for the release. I struggle with this a bit, can anyone help me? In particular: I tried adding a button based on the WH value, at pos 100, 375, but it is always then drawn incorrectly: 
image

I then tested with 475, guessing perhaps WW/WH values could be switched, but it was still invalid. How do I render the button correctly?

src/openrct2-ui/windows/NewVersionInfo.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    window_new_version_process_info();
+
+    int32_t screenWidth = context_get_width();
+    int32_t screenHeight = context_get_height();
+
+    window = window_create_centred(
+        screenWidth * 4 / 5, screenHeight * 4 / 5, &window_new_version_events, WC_NEW_VERSION, WF_RESIZABLE);
+    window->widgets = window_new_version_widgets;
+    window->enabled_widgets = (1 << WIDX_CLOSE);
```
AaronVanGeffen, line 122:
```suggestion
    window->enabled_widgets = (1 << WIDX_CLOSE) | (1 << WIDX_OPEN_URL);
```

AaronVanGeffen:
I think making this window a 'changelog light' will not deliver the most optimal user experience. I suggest hiding the changelog, initially, only revealing after clicking a button. The window can be resized to accommodate the extra scroll widget after.

Concretely, my suggestion is to do the following. When the algorithm detects a new version, we pop up a simple, modal confirmation window, in the same vein as the "save before quitting" confirmation window, but different colours. We show three choices / buttons:
1. Show changelog (expands to show)
2. Remind me later (closes window)
3. Update now (link to release)

src/openrct2/Version.cpp
AaronVanGeffen: I tried to test this PR by changing this line to read:
```diff
-const char gVersionInfoTag[] =
+const char gVersionInfoTag[] = "v0.2.5";
```

But I'm not getting a popup on the title screen. I made sure to delete the `last_version_check_time` from `config.ini` before testing.

Any ideas?

janisozaur: I did test it earlier with such change and seemed ok. That was few revisions ago, so I'll recheck

janisozaur:
I've tried implementing the modal window you talked about, but found it too annoying. Additionally, the approach with modal window with `[Show changelog] [Remind me later] [Open download page]` buttons has following issues:
* `[Remind me later]` is a useless button that does nothing else than quitting the dialog, it won't be included.
* That leaves just the two buttons, but we could show the changelog already which contains link to download page and limit new windows to single one.

The fixed approach with changelog shown is also annoying, but better than the dialog window.

Instead I pursued approach I mentioned somewhere before: adding a button to prompt user to update. This is done in title menu as well as the floppy menu and in "about" window. All those buttons open the changelog for new version, but I changed the colours a bit so it's less confusing about which changelog is show..

While extending the "about" window, I've also added "Join Discord" button.

The final thing I would change about it is to have the floppy menu feature the red notification dot (`SPR_G2_RCT1_CLOSE_BUTTON_0`), but I don't know how to best achieve that - I guess I'd have to explicitly draw it? I tried extending the dropdown entry with `{INLINE_SPRITE}` but hit an assert about palettes, I assume that's due to trying to draw a sprite on translucent BG?

Here are the screenshots:
scr4
scr3
scr2
scr1

ocalhoun6:
If you still want to have three buttons available, one for 'Disable Update Notifications' that automatically changes the config settings to a far-future date and thereby disables updates might be nice.

I guess it's just a pet peeve of mine -- software that _never_ stops nagging you for updates. Sometimes, for whatever reason, I prefer an earlier version of certain software, so update notifications can be extremely annoying. It might be nice to have the ability to disable update checks be less hidden.

janisozaur:
@ocalhoun6 This is not something I will implement. We provide a mechanism for putting off the update check, but it is in our (and users') very interest to keep users updated. The older versions will continue to work, we don't handicap them or prevent access. I find the update notification strikes a decent balance between being attention grabbing and unobtrusive.

Users can still manually change the update check time, if they are so inclined

src/openrct2/interface/Chat.cpp
```diff
@@ -79,16 +79,15 @@ void chat_update()
...
-void chat_draw(rct_drawpixelinfo* dpi, uint8_t chatBackgroundColor)
-{
     if (!chat_available())
     {
         gChatOpen = false;
         return;
     }
+}
 
+void chat_draw(rct_drawpixelinfo* dpi, uint8_t chatBackgroundColor)
```
AaronVanGeffen, line 89: I believe these changes can be reverted.

src/openrct2/network/NetworkBase.cpp
```diff
@@ -3481,6 +3481,14 @@ void network_chat_show_connected_message()
     chat_history_add(formatted);
 }
 
+void network_chat_display_announcement(const char* announcement)
```
AaronVanGeffen, line 3484: While perhaps useful for another PR, I believe this can be reverted.

src/openrct2-ui/windows/TitleMenu.cpp
```diff
@@ -113,7 +118,9 @@ rct_window* window_title_menu_open()
         i++;
     }
     window->width = x;
+    window->widgets[WIDX_NEW_VERSION].right = window->width;
     window->windowPos.x = (context_get_width() - window->width) / 2;
+    window->colours[1] = TRANSLUCENT(COLOUR_LIGHT_ORANGE);
```
AaronVanGeffen, line 123: We don't usually hardcode our window colours. This should be done using a theme class. Orange is a striking default, though — good choice.

src/openrct2-ui/windows/TitleMenu.cpp
```diff
@@ -27,21 +27,26 @@ enum {
...
 static rct_widget window_title_menu_widgets[] = {
-    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
-    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
-    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
-    MakeWidget({0, 0}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
+    MakeWidget({0, 28}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
+    MakeWidget({0, 28}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
+    MakeWidget({0, 28}, {82, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
+    MakeWidget({0, 28}, {82, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
+    MakeWidget({0, 0}, {82 * 4, 28}, WWT_EMPTY, 1, STR_UPDATE_AVAILABLE),
```
AaronVanGeffen, line 41: Please align the definitions:
```suggestion
    MakeWidget({0, 28}, {    82, 82}, WWT_IMGBTN, 2, SPR_MENU_NEW_GAME,       STR_START_NEW_GAME_TIP),
    MakeWidget({0, 28}, {    82, 82}, WWT_IMGBTN, 2, SPR_MENU_LOAD_GAME,      STR_CONTINUE_SAVED_GAME_TIP),
    MakeWidget({0, 28}, {    82, 82}, WWT_IMGBTN, 2, SPR_G2_MENU_MULTIPLAYER, STR_SHOW_MULTIPLAYER_TIP),
    MakeWidget({0, 28}, {    82, 82}, WWT_IMGBTN, 2, SPR_MENU_TOOLBOX,        STR_GAME_TOOLS_TIP),
    MakeWidget({0,  0}, {82 * 4, 28}, WWT_EMPTY,  1, STR_UPDATE_AVAILABLE),
```

AaronVanGeffen:
This is currently the only PR left for v0.3.0. As much as I'd like to see the points above addressed, I do not think they should be blocking the release.

Once this has been rebased, we should consider merging this as-is. We can add issues for the refactor work as laid out above.

Thoughts?

src/openrct2-ui/windows/Changelog.cpp
```diff
@@ -29,19 +31,21 @@ enum {
...
-constexpr int32_t MIN_WH = 200;
+constexpr int32_t MIN_WH = 250;
 
 static rct_widget window_changelog_widgets[] = {
     WINDOW_SHIM(WINDOW_TITLE, WW, WH),
-    MakeWidget({0, 14}, {500, 386}, WWT_RESIZE, 1             ), // content panel
-    MakeWidget({3, 16}, {495, 370}, WWT_SCROLL, 1, SCROLL_BOTH), // scroll area
+    MakeWidget({0, 14}, {500, 382}, WWT_RESIZE, 1             ), // content panel
+    MakeWidget({3, 16}, {495, 366}, WWT_SCROLL, 1, SCROLL_BOTH), // scroll area
+    MakeWidget({3, 473}, {300, 14}, WWT_PLACEHOLDER, 1, STR_NEW_RELEASE_DOWNLOAD_PAGE), // changelog button
```
AaronVanGeffen, line 48: Please align:
```suggestion
    MakeWidget({0,  14}, {500, 382}, WWT_RESIZE,      1                               ), // content panel
    MakeWidget({3,  16}, {495, 366}, WWT_SCROLL,      1, SCROLL_BOTH                  ), // scroll area
    MakeWidget({3, 473}, {300,  14}, WWT_PLACEHOLDER, 1, STR_NEW_RELEASE_DOWNLOAD_PAGE), // changelog button
```

src/openrct2-ui/windows/About.cpp
```diff
@@ -53,7 +56,9 @@ enum WINDOW_ABOUT_WIDGET_IDX {
 
 static rct_widget window_about_openrct2_widgets[] = {
     WIDGETS_MAIN,
-    MakeWidget({100, WH - TABHEIGHT}, {200, 14}, WWT_BUTTON, 1, STR_CHANGELOG_ELLIPSIS), // changelog button
+    MakeWidget({100, WH - TABHEIGHT - (14 + 3) * 2}, {200, 14}, WWT_BUTTON, 1, STR_CHANGELOG_ELLIPSIS),          // changelog button
+    MakeWidget({100, WH - TABHEIGHT - (14 + 3) * 1}, {200, 14}, WWT_BUTTON, 1, STR_JOIN_DISCORD),                 // "join discord" button
+    MakeWidget({100, WH - TABHEIGHT - (14 + 3) * 0}, {200, 14}, WWT_PLACEHOLDER, 1, STR_UPDATE_AVAILABLE), // "new version" button
```
AaronVanGeffen, line 61: Please align:
```suggestion
    MakeWidget({100, WH - TABHEIGHT - (14 + 3) * 2}, {200, 14}, WWT_BUTTON,      1, STR_CHANGELOG_ELLIPSIS), // changelog button
    MakeWidget({100, WH - TABHEIGHT - (14 + 3) * 1}, {200, 14}, WWT_BUTTON,      1, STR_JOIN_DISCORD      ), // "join discord" button
    MakeWidget({100, WH - TABHEIGHT - (14 + 3) * 0}, {200, 14}, WWT_PLACEHOLDER, 1, STR_UPDATE_AVAILABLE  ), // "new version" button
```

src/openrct2/interface/Window.cpp
```diff
@@ -1629,7 +1629,7 @@ void window_resize_gui(int32_t width, int32_t height)
     if (titleWind != nullptr)
     {
         titleWind->windowPos.x = (width - titleWind->width) / 2;
-        titleWind->windowPos.y = height - 154;
+        titleWind->windowPos.y = height - 182;
```
Gymnasiast, line 1632: Why this change?

AaronVanGeffen: The update notifier was added on top, making the menu larger. This change compensates for it.

(I wondered the same thing when reviewing, myself.)

janisozaur:
As the two windows have now been merged, I also urge you to check if both functionalities still work.

In general I'm quite happy with merging them together, as this could be further reused for a large textbox if needed later on. One thing I had to give up was setting the colour of the window, I remember being upset with the defaults being a bit hard to read.

src/openrct2-ui/windows/Changelog.cpp
```diff
@@ -85,43 +89,77 @@ static rct_window_event_list window_changelog_events = {
...
+            window_changelog_widgets[WIDX_OPEN_URL].type = WWT_BUTTON;
+            break;
+
+        case WV_CHANGELOG:
+            if (!window_changelog_read_file())
+            {
+                return nullptr;
+            }
+
+            _persnality = WV_CHANGELOG;
```
janisozaur, line 134: Typo: missing 'o', `persOnality`

ZehMatt:
I tested it and it does work however I got some suggestions. 
1. I think the button is a bit too prominent, I would probably place it next to the Options button.
2. In the About window I would like to see a "Check for updates" button, when nothing newer is available it should result in a small "error" message, otherwise show up the changelog window.
3. The added context entry is also a bit too much, it should be enough to have it in the main window.

AaronVanGeffen:
@ZehMatt I agree with your suggestions; I think there's room for improvement as well. However, we agreed it would be good to have an update mechanism in the upcoming release. I don't think it's worth delaying the release further, particularly as no-one will see it until the release after that.

ZehMatt:
> @ZehMatt I agree with your suggestions; I think there's room for improvement as well. However, we agreed it would be good to have an update mechanism in the upcoming release. I don't think it's worth delaying the release further, particularly as no-one will see it until the release after that.

Fair enough. Maybe just the button?

janisozaur:
I disagree, the reason it's explicitly so prominent is because of the amount of "I don't know if I'm my OpenRCT2 is up to date" responses in survey. Even though prominent, it doesn't get in the way of things user might want to do.

AaronVanGeffen:
> Fair enough. Maybe just the button?

@ZehMatt Unfortunately, the button is part of the TitleMenu window. Moving it to the top right would constitute either making a new window for it, or moving it into the TitleOptions window. Either way, a (perhaps surprisingly) major change.

ZehMatt:
> I disagree, the reason it's explicitly so prominent is because of the amount of "I don't know if I'm my OpenRCT2 is up to date" responses in survey. Even though prominent, it doesn't get in the way of things user might want to do.

I think most people would understand if the text in the right bottom corner would say "Newer version available" would already mitigate this? I don't see how this warrants a giant button.

ZehMatt:
Anyway, I'm fine with it as is, not going to be a blocker here.

Issue #12329, opened 2020-07-20
Refactor widget definitions


AaronVanGeffen:
We are currently in the process of refactoring the widget definitions for all widgets.  The old-style widget definitions read as follows:

```cpp
    { WWT_BUTTON,           1,  73,     83,     60, 69,     STR_DROPDOWN_GLYPH,     STR_NONE },                     // page dropdown button
    { WWT_DROPDOWN,         1,  120,    261,    59, 70,     0xFFFFFFFF,             STR_INFORMATION_TYPE_TIP },     // information type dropdown
    { WWT_BUTTON,           1,  250,    260,    60, 69,     STR_DROPDOWN_GLYPH,     STR_INFORMATION_TYPE_TIP },     // information type dropdown button
    { WWT_FLATBTN,          1,  273,    296,    46, 69,     SPR_MAP,                STR_SHOW_GUESTS_ON_MAP_TIP },   // map
    { WWT_FLATBTN,          1,  297,    320,    46, 69,     SPR_G2_SEARCH,          STR_GUESTS_FILTER_BY_NAME_TIP },// filter by name
```

The old parameter order is the same as the internal widget structure order: type, colour, left, right, top, bottom, content, and tooltip.

The new-style definitions change this to use a position (x, y) and dimensions (width, height) tuple instead, as follows:
```cpp
    MakeWidget({ 73, 60}, { 11,  10}, WWT_BUTTON,   1, STR_DROPDOWN_GLYPH                               ), // page dropdown button
    MakeWidget({120, 59}, {142,  12}, WWT_DROPDOWN, 1, 0xFFFFFFFF,         STR_INFORMATION_TYPE_TIP     ), // information type dropdown
    MakeWidget({250, 60}, { 11,  10}, WWT_BUTTON,   1, STR_DROPDOWN_GLYPH, STR_INFORMATION_TYPE_TIP     ), // information type dropdown button
    MakeWidget({273, 46}, { 24,  24}, WWT_FLATBTN,  1, SPR_MAP,            STR_SHOW_GUESTS_ON_MAP_TIP   ), // map
    MakeWidget({297, 46}, { 24,  24}, WWT_FLATBTN,  1, SPR_G2_SEARCH,      STR_GUESTS_FILTER_BY_NAME_TIP), // filter by name
```

However, we still have to go over all widget definitions manually. Not just to update the widget definitions, but also to verify the window is still behaving correctly!

[...]

PR #12354, opened 2020-07-22
Refactor more widget definitions

AaronVanGeffen:
Part of #12329


src/openrct2-ui/windows/EditorScenarioOptions.cpp
```diff
@@ -113,60 +112,53 @@ enum {
...
-    { WWT_DROPDOWN,         1,  188,    394,    99,     110,    STR_NONE,                               STR_SELECT_CLIMATE_TIP                      },
-    { WWT_BUTTON,           1,  383,    393,    100,    109,    STR_DROPDOWN_GLYPH,                     STR_SELECT_CLIMATE_TIP                      },
-    { WWT_CHECKBOX,         1,  8,      391,    116,    127,    STR_FORBID_TREE_REMOVAL,                STR_FORBID_TREE_REMOVAL_TIP                 },
-    { WWT_CHECKBOX,         1,  8,      391,    133,    144,    STR_FORBID_LANDSCAPE_CHANGES,           STR_FORBID_LANDSCAPE_CHANGES_TIP            },
-    { WWT_CHECKBOX,         1,  8,      391,    150,    161,    STR_FORBID_HIGH_CONSTRUCTION,           STR_FORBID_HIGH_CONSTRUCTION_TIP            },
-    { WWT_CHECKBOX,         1,  8,      391,    167,    178,    STR_HARD_PARK_RATING,                   STR_HARD_PARK_RATING_TIP                    },
-    { WWT_CHECKBOX,         1,  8,      391,    184,    195,    STR_HARD_GUEST_GENERATION,              STR_HARD_GUEST_GENERATION_TIP               },
+    WINDOW_SHIM(STR_SCENARIO_OPTIONS_PARK, 400, 200),
```
Gymnasiast, line 145: Could do with local constants. The `384`, used for the latest five widgets, also appears to be derived from this 400.

AaronVanGeffen: Related, yes. Derived, no. Naturally, all widgets are constrained by the window width. However, I believe that expressing widget width in terms of window width, just because the widget is on the wider side, would be overdoing it.

I think these numbers by themselves are more intuitive than hiding them behind constants.

src/openrct2-ui/windows/Map.cpp
```diff
@@ -81,23 +81,23 @@ validate_global_widx(WC_MAP, WIDX_ROTATE_90);
...
+    MakeRemapWidget   ({  3,  17}, { 31,  27}, WWT_COLOURBTN, 1, SPR_TAB,                         STR_SHOW_PEOPLE_ON_MAP_TIP                     ),
+    MakeRemapWidget   ({ 34,  17}, { 31,  27}, WWT_COLOURBTN, 1, SPR_TAB,                         STR_SHOW_RIDES_STALLS_ON_MAP_TIP               ),
+    MakeWidget        ({  3,  46}, {239, 180}, WWT_SCROLL,    1, SCROLL_BOTH                                                                     ),
+    MakeSpinnerWidgets({104, 229}, { 95,  12}, WWT_SPINNER,   1, STR_MAP_SIZE_VALUE                                                              ), // NB: 3 widgets
+    MakeWidget        ({  4,   1}, { 24,  24}, WWT_FLATBTN,   1, SPR_BUY_LAND_RIGHTS,             STR_SELECT_PARK_OWNED_LAND_TIP                 ),
+    MakeWidget        ({  4,   1}, { 24,  24}, WWT_FLATBTN,   1, SPR_PARK_ENTRANCE,               STR_BUILD_PARK_ENTRANCE_TIP                    ),
+    MakeWidget        ({ 28,   1}, { 24,  24}, WWT_FLATBTN,   1, 0xFFFFFFFF,                      STR_SET_STARTING_POSITIONS_TIP                 ),
```
AaronVanGeffen, line 91: We need a proper constant to define empty widget content. `SPR_NONE` is not defined as an `uint32_t`, so it does not suffice.

I'd prefer to tackle these all at once in a follow-up PR when all windows are done.

src/openrct2-ui/windows/EditorScenarioOptions.cpp
Gymnasiast: They can be left for another PR, if you create an issue.

PR #12367, opened 2020-07-23
Close #12338: Remove goto from Tick128UpdateGuest

### LINKED ISSUE #12338: Remove goto from Tick128UpdateGuest
https://github.com/OpenRCT2/OpenRCT2/blob/d10b615eb33f6cd597340d83d68eaaa97c1a493e/src/openrct2/peep/Guest.cpp#L820-L823

frutiemax:


src/openrct2/peep/Guest.cpp
```diff
@@ -820,7 +995,7 @@ void Guest::Tick128UpdateGuest(int32_t index)
                     HappinessTarget = std::max(HappinessTarget - 128, 0);
                     peep_leave_park(this);
                     peep_update_hunger(this);
-                    goto loc_68F9F3;
+                    UpdateStatistics();
```
duncanspumpkin, line 998: Must return after this statement

src/openrct2/peep/Guest.cpp
```diff
@@ -694,6 +694,181 @@ int32_t Guest::CheckEasterEggName(int32_t index) const
...
+            {
+                Action = PEEP_ACTION_THROW_UP;
+                ActionFrame = 0;
+                ActionSpriteImageOffset = 0;
+                UpdateCurrentActionSpriteType();
+            }
+        }
+    }
+
+    // Remaining content is executed every call.
```
duncanspumpkin, line 741: You have changed the logic here. This needs to be split up into two functions as this second part from here is called every time but the first part is only called sometimes. That's what this comment is for.

src/openrct2/peep/Guest.cpp
```diff
@@ -694,6 +694,181 @@ int32_t Guest::CheckEasterEggName(int32_t index) const
...
+    if (newNausea != Nausea)
+    {
+        Nausea = newNausea;
+        WindowInvalidateFlags |= PEEP_INVALIDATE_PEEP_2;
+    }
+}
+
 void Guest::Tick128UpdateGuest(int32_t index)
 {
     if (static_cast<uint32_t>(index & 0x1FF) == (gCurrentTicks & 0x1FF))
```
duncanspumpkin, line 874: Function should be split up into two here.

src/openrct2/peep/Guest.cpp
```diff
@@ -694,6 +694,181 @@ int32_t Guest::CheckEasterEggName(int32_t index) const
     return _stricmp(buffer, gPeepEasterEggNames[index]) == 0;
 }
 
+void Guest::UpdateStatistics()
```
duncanspumpkin, line 697: These aren't really statistics can you come up with a better name

src/openrct2/peep/Guest.cpp
frutiemax: You are right, I've pretty much reverted all my changes. I haven't found a proper name for these functions.

duncanspumpkin:
Its not great the new names but at least its easy to follow and someone can improve the names later. This refactor also helped @ZehMatt and myself realise there is a desync issue with the entity lists.

PR #12373, opened 2020-07-24
Improve connecting to servers

ZehMatt:
This PR adds a few quality of life things and fixes one minor bug.
1. I renamed a few constants to better reflect what is actually going on on the network such as NETWORK_COMMAND_OBJECTS which was a bit misleading, the reality was that the server sends the object list and the client used the same constant to request the map once the object list was received including the missing objects. The map request has now its own constant NETWORK_COMMAND_MAPREQUEST.
2. The server sends now the object list in segments rather than one large packet allowing the client to receive a status information on what is going. When connecting to a server with slow uploads it would seem as nothing is happening.
3. If the server upload speed is rather low and clients have a lot of missing objects the chance was pretty high that the server disconnected the clients before the client had the chance to fully receive the map, the reason for this is when maps are big with a lot of missing objects on the client it could take a while for the client to download it, during the download stage the client would not send any packets to the server which made the server believe the client timed out, the threshold is 7 seconds. To avoid such thing happening I implemented a heartbeat message which is sent every 3 seconds to ensure the server will not drop the connection during the stage of sending the object list/map.

Connecting to a server now looks like following:
4yZe3WX8j5


data/language/en-GB.txt
```diff
@@ -3664,6 +3664,7 @@ STR_6374    :C
 STR_6375    :Unknown Ride
 STR_6376    :{WINDOW_COLOUR_2}Ride vehicle:{NEWLINE}{BLACK}{STRINGID} for {STRINGID}
 STR_6377    :{WINDOW_COLOUR_2}Type: {BLACK}{STRINGID} for {STRINGID}
+STR_6378    :Receiving objects list {INT32} / {INT32}
```
Gymnasiast, line 3667:
```suggestion
STR_6378    :Receiving objects list: {INT32} / {INT32}
```

src/openrct2/localisation/StringIds.h
```diff
@@ -3906,6 +3906,7 @@ enum
     STR_RESEARCH_VEHICLE_LABEL = 6376,
     STR_RESEARCH_TYPE_LABEL_VEHICLE = 6377,
 
+    STR_MULTIPLAYER_RECEIVING_OBJECTS_LIST = 6378,
```
Gymnasiast, line 3909:
```suggestion
    STR_MULTIPLAYER_RECEIVING_OBJECTS_LIST = 6378,

```

ZehMatt:
Yeah, I'm hoping @Broxzier could take it for a ride.

src/openrct2/network/Network.cpp
```diff
@@ -310,10 +310,10 @@ class Network
     void Client_Handle_EVENT(NetworkConnection& connection, NetworkPacket& packet);
     void Client_Handle_TOKEN(NetworkConnection& connection, NetworkPacket& packet);
     void Server_Handle_TOKEN(NetworkConnection& connection, NetworkPacket& packet);
-    void Client_Handle_OBJECTS(NetworkConnection& connection, NetworkPacket& packet);
+    void Client_Handle_OBJECTS_LIST(NetworkConnection& connection, NetworkPacket& packet);
     void Client_Handle_SCRIPTS(NetworkConnection& connection, NetworkPacket& packet);
     void Client_Handle_GAMESTATE(NetworkConnection& connection, NetworkPacket& packet);
-    void Server_Handle_OBJECTS(NetworkConnection& connection, NetworkPacket& packet);
+    void Server_Handle_MAPREQUEST(NetworkConnection& connection, NetworkPacket& packet);
```
tupaschoal, line 320: Seems like things should either be unified as `MAP_REQUEST` or `MAPREQUEST`
```suggestion
    void Server_Handle_MAP_REQUEST(NetworkConnection& connection, NetworkPacket& packet);
```

src/openrct2/network/Network.cpp
```diff
@@ -1470,18 +1471,24 @@ void Network::Server_Send_TOKEN(NetworkConnection& connection)
...
 
-void Network::Server_Send_OBJECTS(NetworkConnection& connection, const std::vector<const ObjectRepositoryItem*>& objects) const
+void Network::Server_Send_OBJECTS_LIST(NetworkConnection& connection, const std::vector<const ObjectRepositoryItem*>& objects) const
 {
     log_verbose("Server sends objects list with %u items", objects.size());
-    std::unique_ptr<NetworkPacket> packet(NetworkPacket::Allocate());
-    *packet << static_cast<uint32_t>(NETWORK_COMMAND_OBJECTS_LIST) << static_cast<uint32_t>(objects.size());
-    for (auto object : objects)
+
+    for (size_t i = 0; i < objects.size(); ++i)
```
tupaschoal, line 1493: I wonder if isn't clearer if we use ranged-based for and then `std::distance` where the index is clearer.

src/openrct2/network/Network.cpp
```diff
@@ -1588,7 +1595,7 @@ void Network::Server_Send_MAP(NetworkConnection* connection)
         }
         return;
     }
-    size_t chunksize = CHUNK_SIZE;
+    size_t chunksize = 128; //CHUNK_SIZE;
```
tupaschoal: Seems to be a mistake? I'd expect it to introduce another constant, if `CHUNK_SIZE != 128` or use `CHUNK_SIZE` directly.

src/openrct2/network/Network.cpp
```diff
@@ -2532,40 +2539,64 @@ void Network::Server_Handle_TOKEN(NetworkConnection& connection, [[maybe_unused]
...
-    log_verbose("client received object list, it has %u entries", size);
-    if (size > OBJECT_ENTRY_COUNT)
+
+    uint32_t index = 0;
+    uint32_t totalObjects = 0;
+    packet >> index >> totalObjects;
+
+    if (index == 0)
+    {
+        // Start of objects list.
```
tupaschoal, line 2580: I wonder if we should make
```c++
if (index == OBJECT_LIST_START)
{
    _missingObjects.clear();
}
```

Where `OBJECT_LIST_START` is `0`

src/openrct2/network/Network.cpp
```diff
@@ -2532,40 +2539,64 @@ void Network::Server_Handle_TOKEN(NetworkConnection& connection, [[maybe_unused]
...
+        totalObjects,
+    };
+    format_string(objectListMsg, 256, STR_MULTIPLAYER_RECEIVING_OBJECTS_LIST, &args);
+
+    auto intent = Intent(WC_NETWORK_STATUS);
+    intent.putExtra(INTENT_EXTRA_MESSAGE, std::string{ objectListMsg });
+    intent.putExtra(INTENT_EXTRA_CALLBACK, []() -> void { gNetwork.Close(); });
+    context_open_intent(&intent);
+
+    const char* name = reinterpret_cast<const char*>(packet.Read(8));
```
tupaschoal, line 2604:
```suggestion
    auto name = reinterpret_cast<const char*>(packet.Read(8));
```
Type is already on the right

src/openrct2/network/Network.cpp
```diff
@@ -2532,40 +2539,64 @@ void Network::Server_Handle_TOKEN(NetworkConnection& connection, [[maybe_unused]
...
+    // Required, as packet has no null terminators.
+    std::string s(name, name + 8);
+    uint32_t checksum, flags;
```
tupaschoal, line 2607:
```suggestion
    uint32_t checksum = 0;
    uint32_t flags = 0;
```

To be consistent with something similar at the top of the function

src/openrct2/network/Network.cpp
```diff
@@ -2532,40 +2539,64 @@ void Network::Server_Handle_TOKEN(NetworkConnection& connection, [[maybe_unused]
...
+    packet >> checksum >> flags;
+
+    const ObjectRepositoryItem* ori = repo.FindObject(s.c_str());
```
tupaschoal, line 2610:
```suggestion
    auto ori = repo.FindObject(s.c_str());
```

src/openrct2/network/Network.cpp
ZehMatt: I agree, changing it to MAPREQUEST. That slipped on my end.

src/openrct2/network/Network.cpp
ZehMatt: I could swear I removed this before commiting, good catch.

src/openrct2/network/Network.cpp
ZehMatt: I don't like naked auto when its a pointer, also I would like to keep const, even if the function wouldn't return it the usage is const.

src/openrct2/network/Network.cpp
ZehMatt: I've changed it to send a bool if its the start of the list, this should clarify the purpose a bit better and the index is obsolete since I can use the count of _missingObjects.

ZehMatt:
Regarding naming consistency, I'll do this in another PR and do all of them at once.

IntelOrca:
Should it say `receiving objects` rather than `receiving object list`?

src/openrct2/network/Network.cpp
```diff
@@ -1449,11 +1464,11 @@ void Network::Client_Send_AUTH(
     _serverConnection->QueuePacket(std::move(packet));
 }
 
-void Network::Client_Send_OBJECTS(const std::vector<std::string>& objects)
+void Network::Client_Send_REQUESTMAP(const std::vector<std::string>& objects)
 {
     log_verbose("client requests %u objects", uint32_t(objects.size()));
     std::unique_ptr<NetworkPacket> packet(NetworkPacket::Allocate());
-    *packet << static_cast<uint32_t>(NETWORK_COMMAND_OBJECTS) << static_cast<uint32_t>(objects.size());
+    *packet << static_cast<uint32_t>(NETWORK_COMMAND_MAPREQUEST) << static_cast<uint32_t>(objects.size());
```
tupaschoal, line 1471: Now I'm confused whether it's `REQUESTMAP` or `MAPREQUEST` hahaha

src/openrct2/network/Network.cpp
```diff
@@ -1470,18 +1485,25 @@ void Network::Server_Send_TOKEN(NetworkConnection& connection)
...
+void Network::Server_Send_OBJECTS_LIST(
+    NetworkConnection& connection, const std::vector<const ObjectRepositoryItem*>& objects) const
 {
     log_verbose("Server sends objects list with %u items", objects.size());
-    std::unique_ptr<NetworkPacket> packet(NetworkPacket::Allocate());
-    *packet << static_cast<uint32_t>(NETWORK_COMMAND_OBJECTS) << static_cast<uint32_t>(objects.size());
-    for (auto object : objects)
+
+    bool startOfObjectList = true;
+    for (auto* object : objects)
```
tupaschoal, line 1494:
```suggestion
    for (const auto* object : objects)
```

ZehMatt:
I just realized there are 2 cases I'm currently not covering. It seems that the object list can be empty, and I can not rely on _missingObjects size to be equal to totalObjectCount, I need to revert this back to sending the index.

IntelOrca:
It looked like the progress was showing an object count, was it just bytes?

ZehMatt:
> It looked like the progress was showing an object count, was it just bytes?

Its the object count. The server sends a list of objects to the client, the client will check which objects are missing or are mismatching and sends the list with the map request which the server will pack into the save game.

ZehMatt:
Now its working as intended.

src/openrct2/network/Network.cpp
```diff
@@ -2529,43 +2578,71 @@ void Network::Server_Handle_TOKEN(NetworkConnection& connection, [[maybe_unused]
...
+        };
+        format_string(objectListMsg, 256, STR_MULTIPLAYER_RECEIVING_OBJECTS_LIST, &args);
+
+        auto intent = Intent(WC_NETWORK_STATUS);
+        intent.putExtra(INTENT_EXTRA_MESSAGE, std::string{ objectListMsg });
+        intent.putExtra(INTENT_EXTRA_CALLBACK, []() -> void { gNetwork.Close(); });
+        context_open_intent(&intent);
+
+        char objectName[12]{};
+        std::memcpy(objectName, packet.Read(8), 8);
```
tupaschoal, line 2618: Should we avoid `memcpy` here?

Also why does it have 3 extra positions?

ZehMatt: I could make it strictly 9 so the 9. would be the zero but it shouldn't matter much. Do you consider this worse than having a temporary std::string object? std::string::assign at the end is the same as a memcpy

tupaschoal: If it doesn't seem to be a critical piece of code, I'd just have a simple string assignment, I think it improves legibility and also makes it clear that the memcpy is not there for performance reasons

ZehMatt: In my eyes this clarifies this is a fixed length string, but I can revert this change if must.

tupaschoal: I don't want to make a fuss about it, your call

ZehMatt: Using std::string just didn't feel right for something rather trivial, but if anyone else has something to add please do.

frutiemax: Why are you not doing `auto objectName = packet.Read(8);` ?

ZehMatt: Because its not null terminated

PR #12375, opened 2020-07-24
Disable HTTP for targets that don't need it

janisozaur:


tupaschoal:
Out of curiosity, is this just for the sake of correctness/build speed or did you face an issue?

janisozaur:
This is extracted from #12347 where the tests failed to link

PR #12376, opened 2020-07-24
Part of #11389: Audit calls to memcpy with gCommonFormatArgs

frutiemax:
I've added a CopyInto method to do the value copying. I've also made it so when you call it with uint16_t*, the function doubles the number of bytes to pass into the std::memcpy. I've tested all places where this function is called.

### DIFF
--- a/src/openrct2-ui/windows/GuestList.cpp
+++ b/src/openrct2-ui/windows/GuestList.cpp
@@ -928,7 +928,9 @@ static FilterArguments get_arguments_from_peep(const Peep* peep)
             {
                 std::memset(gCommonFormatArgs, 0, sizeof(gCommonFormatArgs));
                 peep_thought_set_format_args(thought);
-                std::memcpy(result.args, gCommonFormatArgs, std::min(sizeof(gCommonFormatArgs), sizeof(result.args)));
+
+                ft = Formatter::Common();
+                ft.CopyInto<uint8_t*>(result.args, sizeof(result.args));
             }
             break;
         }
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -3764,9 +3764,9 @@ static void window_ride_operating_invalidate(rct_window* w)
         if (ride->type == RIDE_TYPE_TWIST)
         {
             uint16_t arg;
-            std::memcpy(&arg, gCommonFormatArgs + 18, sizeof(uint16_t));
             ft = Formatter::Common();
             ft.Increment(18);
+            ft.CopyInto<uint16_t*>(&arg, 1);
             ft.Add<uint16_t>(arg * 3);
         }
 
--- a/src/openrct2-ui/windows/RideList.cpp
+++ b/src/openrct2-ui/windows/RideList.cpp
@@ -708,7 +708,9 @@ static void window_ride_list_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, 
                 formatSecondary = STR_QUEUE_EMPTY;
                 {
                     uint16_t arg;
-                    std::memcpy(&arg, gCommonFormatArgs + 2, sizeof(uint16_t));
+                    ft.Rewind();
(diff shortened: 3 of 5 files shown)


src/openrct2-ui/windows/GuestList.cpp
```diff
@@ -928,7 +928,9 @@ static FilterArguments get_arguments_from_peep(const Peep* peep)
             {
                 std::memset(gCommonFormatArgs, 0, sizeof(gCommonFormatArgs));
                 peep_thought_set_format_args(thought);
```
duncanspumpkin, line 930: This function should take the formatter ft we shouldn't be playing around setting gCommonFormatArgs to zero or anything else.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -3764,9 +3764,9 @@ static void window_ride_operating_invalidate(rct_window* w)
         if (ride->type == RIDE_TYPE_TWIST)
         {
             uint16_t arg;
-            std::memcpy(&arg, gCommonFormatArgs + 18, sizeof(uint16_t));
             ft = Formatter::Common();
             ft.Increment(18);
+            ft.CopyInto<uint16_t*>(&arg, 1);
```
duncanspumpkin, line 3769: I don't like this. We should get the correct variable from earlier in the function not copy it from the global buffer.

src/openrct2-ui/windows/RideList.cpp
```diff
@@ -708,7 +708,9 @@ static void window_ride_list_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi,
                 formatSecondary = STR_QUEUE_EMPTY;
                 {
                     uint16_t arg;
-                    std::memcpy(&arg, gCommonFormatArgs + 2, sizeof(uint16_t));
+                    ft.Rewind();
```
duncanspumpkin, line 711: This code is ridiculous. First we get the total queue length and put it in the buffer. Then we rewind the buffer then we increment the buffer to the total queue length. Then we copy out the total queue length. How about we just use the total queue length directly!

src/openrct2-ui/windows/RideList.cpp
```diff
@@ -721,7 +723,9 @@ static void window_ride_list_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi,
                 formatSecondary = STR_QUEUE_TIME_LABEL;
                 {
                     uint16_t arg;
-                    std::memcpy(&arg, gCommonFormatArgs + 2, sizeof(uint16_t));
+                    ft.Rewind();
```
duncanspumpkin, line 726: as above

src/openrct2/drawing/ScrollingText.cpp
```diff
@@ -1462,7 +1462,10 @@ int32_t scrolling_text_setup(
     // Setup scrolling text
     auto scrollText = &_drawScrollTextList[scrollIndex];
     scrollText->string_id = stringId;
-    std::memcpy(scrollText->string_args, gCommonFormatArgs, sizeof(scrollText->string_args));
+
+    auto ft = Formatter::Common();
+    ft.CopyInto<uint8_t*>(scrollText->string_args, sizeof(scrollText->string_args));
```
duncanspumpkin, line 1467: scolling_text_setup should be modified to take a formatter making it not pass the values through the global variable. This will likely reduce the rendering issues that occur.

In a latter refactor all callers should be modified to create their own formatter not use the `Common` as this function is used multithreaded. Will require some further work to make this work so hence why it should be left for a later refactor.

src/openrct2/localisation/Localisation.h
```diff
@@ -127,6 +127,15 @@ class Formatter
         return CurrentBuf - StartBuf;
     }
 
+    template<typename TSpecified, typename TDeduced> void CopyInto(TDeduced value, size_t numElements)
```
duncanspumpkin, line 130: This function is not required for this PR.

PR #12382, opened 2020-07-25
Reduce use of get_sprite

duncanspumpkin:
This reduces the uses of the get_sprite function to just the vehicle.h and peep.h macros.

### DIFF
--- a/src/openrct2/GameStateSnapshots.cpp
+++ b/src/openrct2/GameStateSnapshots.cpp
@@ -31,7 +31,8 @@ struct GameStateSnapshot_t
     MemoryStream storedSprites;
     MemoryStream parkParameters;
 
-    void SerialiseSprites(rct_sprite* sprites, const size_t numSprites, bool saving)
+    // Must pass a function that can access the sprite.
+    void SerialiseSprites(std::function<rct_sprite*(const size_t)> getEntity, const size_t numSprites, bool saving)
     {
         const bool loading = !saving;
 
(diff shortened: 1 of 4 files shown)


src/openrct2/world/Sprite.cpp
```diff
@@ -284,11 +284,12 @@ rct_sprite_checksum sprite_checksum()
...
             // TODO create a way to copy only the specific type
-            auto sprite = get_sprite(i);
-            if (sprite->generic.sprite_identifier != SPRITE_IDENTIFIER_NULL
-                && sprite->generic.sprite_identifier != SPRITE_IDENTIFIER_MISC)
+            auto sprite = GetEntity(i);
+            if (sprite != nullptr && sprite->sprite_identifier != SPRITE_IDENTIFIER_NULL
+                && sprite->sprite_identifier != SPRITE_IDENTIFIER_MISC)
             {
-                auto copy = *sprite;
+                // Up convert it to rct_sprite so that the full size is copied.
```
Gymnasiast, line 291:
```suggestion
                // Upconvert it to rct_sprite so that the full size is copied.
```

src/openrct2/GameStateSnapshots.cpp
```diff
@@ -132,8 +133,9 @@ struct GameStateSnapshots final : public IGameStateSnapshots
 
     virtual void Capture(GameStateSnapshot_t& snapshot) override final
     {
-        // TODO refactor to not use this as a proxy for getting a pointer to the sprite array
-        snapshot.SerialiseSprites(get_sprite(0), MAX_SPRITES, true);
+        snapshot.SerialiseSprites(
+            [](const size_t index) -> rct_sprite& { return *reinterpret_cast<rct_sprite*>(GetEntity(index)); }, MAX_SPRITES,
```
tupaschoal, line 137: If this lambda ever tries to dereference a `nullptr` it will crash immediately.

duncanspumpkin: Hmm good point.

duncanspumpkin: Its never going to happen due to the limits involved but I've now changed it so that it will log an error when it gets to the state that can never happen.

PR #12383, opened 2020-07-25
Tile Inspector: remove hardcoded surface/edge style names

Gymnasiast:
When the surface styles were moved into objects, the tile inspector was not modified to use these. This is causing problems on the new save format branch, so this PR tackles that.


Umdlye:
The RCT1 edge style strings are still there after this update, are they still needed for something?

Gymnasiast:
@Umdlye Well spotted! I have now removed those as well.

Umdlye:
Working on copying translations from these strings to the object files and noticed some of the removed strings don't have matching object files (blue grid, inverted checkerboard, underground view). Removing those strings might cause issues.

Gymnasiast:
@Umdlye They won't, as those are only used for rendering. The blue grid of the Track Designer, for example, is grass in disguise. That's also what the TI will say.

@Broxzier The descriptions of the objects can be fixed in https://github.com/OpenRCT2/objects/tree/master/objects/rct2/terrain_surface

Gymnasiast:
@Broxzier Addressed here: https://github.com/OpenRCT2/objects/commit/ca912583fe20dca53289abdd0b40c2144f12b086

PR #12384, opened 2020-07-25
Make all game actions take an ObjectEntryIndex where applicable

Gymnasiast:


duncanspumpkin:
Will require updated replays.

ZehMatt:
We often use the term 'type' in the variables but our type for those variables is called ObjectEntryIndex, can someone clarify on whether its a type or index?

duncanspumpkin:
It's a type of object represented as an index into the object array.

duncanspumpkin (approved):

PR #12385, opened 2020-07-25
Disable vcpkg for all projects

ZehMatt:
Since we ship pre-compiled libraries this causes some issues, in my case I have gtest installed from Vcpkg and causes build errors. Since we don't use Vcpkg I simply disabled it in all projects.

### DIFF
--- a/openrct2.common.props
+++ b/openrct2.common.props
@@ -5,7 +5,6 @@
   </PropertyGroup>
 
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
-
   <PropertyGroup>
     <!-- Allow any version of VS and Windows SDK -->
     <PlatformToolset>$(DefaultPlatformToolset)</PlatformToolset>
@@ -113,4 +112,8 @@
       <EnableDpiAwareness>PerMonitorHighDPIAware</EnableDpiAwareness>
     </Manifest>
   </ItemDefinitionGroup>
+  
+  <PropertyGroup>
+    <VcpkgEnabled>false</VcpkgEnabled>
+  </PropertyGroup>
 </Project>


src/openrct2-cli/openrct2-cli.vcxproj
```diff
@@ -34,6 +34,18 @@
     <TargetName>openrct2-cli</TargetName>
     <LibraryPath>$(SolutionDir)bin;$(LibraryPath)</LibraryPath>
   </PropertyGroup>
+  <PropertyGroup Label="Vcpkg" Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
```
Broxzier, line 37: Can't these conditions simply be removed? The value of `VcpkgEnabled` doesn't change.

ZehMatt: Visual Studio did that, I changed this configuration via the Properties.

ZehMatt: I tested it without conditions, seems to be fine.

IntelOrca:
I think there are times where I use vcpkg, see:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/openrct2.common.props#L22

But if you are going to do this, put it in `openrct2.common.props`.

ZehMatt:
> I think there are times where I use vcpkg, see:
> https://github.com/OpenRCT2/OpenRCT2/blob/develop/openrct2.common.props#L22
> 
> But if you are going to do this, put it in `openrct2.common.props`.

What ports are you using? If I disable it entirely that will no longer work

IntelOrca:
@ZehMatt probably when I was trying new ones out

ZehMatt:
So are those lines you pointed out supposed to disable Vcpkg? I'm getting a conflict with gtest

ZehMatt:
Anyway I moved the attributes to the common.props and it seems to be fine.

IntelOrca:
The label on `PropertyGroup` is unnecessary by the way.

ZehMatt:
Didn't know, should I remove it?

PR #12386, opened 2020-07-25
Refactor PeepType to use strong enum

ZehMatt:
I'm planning to continuously refactor the enums to strong enums where possible, this is the first PR.

### DIFF
--- a/src/openrct2/peep/Peep.h
+++ b/src/openrct2/peep/Peep.h
@@ -52,12 +52,12 @@ struct Ride;
 class GameActionResult;
 using ParkEntranceIndex = uint8_t;
 
-enum PeepType : uint8_t
+enum class PeepType : uint8_t
 {
-    PEEP_TYPE_GUEST,
-    PEEP_TYPE_STAFF,
+    Guest,
+    Staff,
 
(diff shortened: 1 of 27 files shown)


src/openrct2/peep/Peep.h
```diff
@@ -52,12 +52,12 @@ struct Ride;
...
+enum class PeepType : uint8_t
 {
-    PEEP_TYPE_GUEST,
-    PEEP_TYPE_STAFF,
+    Guest,
+    Staff,
 
-    PEEP_TYPE_INVALID = 0xFF
+    Invalid = 0xFF
 };
```
tupaschoal, line 61: I'd have hoped we'd add a namespace here:

```c++
namespace Peep
{
    enum class Type : uint8_t
    {
        Guest,
        Staff,
        Invalid = 0xFF
    };
}
```

Can we do it for the following? Seems like a very good opportunity to increase our usage of namespaces.

pizza2004: I'm trying to figure out how to turn this into a namespace but Peep is clashing with the struct Peep lower in the file and giving me an error.  Any idea how to refactor around that?

PR #12464, opened 2020-07-25
Fix #12387: Refactor PeepThoughtType to use strong enum

### LINKED ISSUE #12387: Refactor PeepThoughtType to use strong enum

pizza2004:
I think this is more or less what was wanted.  I also tried to put it all in a namespace but it really didn't like Peep so I did PeepE temporarily.


pizza2004:
I see someone else has already done this now.  I should have said something about doing it myself.

PR #12465, opened 2020-07-25
Refactor Peep Enum into Namespace

pizza2004:
I changed the peep enums to use a namespace.

### DIFF
--- a/src/openrct2/peep/Peep.h
+++ b/src/openrct2/peep/Peep.h
@@ -52,13 +52,15 @@ struct Ride;
 class GameActionResult;
 using ParkEntranceIndex = uint8_t;
 
-enum class PeepType : uint8_t
+namespace Crowd
 {
-    Guest,
-    Staff,
-
-    Invalid = 0xFF
-};
+    enum class Type : uint8_t
+    {
+        Guest,
+        Staff,
+        Invalid = 0xFF
+    };
+}
 
 enum PeepThoughtType : uint8_t
 {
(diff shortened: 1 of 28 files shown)


ZehMatt:
Sorry but PeepE is not a very good name. The purpose of adding namespaces is to categorize the code its not just for the enums.

pizza2004:
I agree, but it wouldn't allow me to name it Peep because of the struct existing in the same file, and I just figured I'd change it once I had it up.  It'll only take a second to change the name once one is agreed on.

ZehMatt:
We discussed this briefly on Discord and we agreed on doing this as PRs come. @tupaschoal any suggestions for the Peep/Guest/Staff namespace?

tupaschoal:
I can think of some that might be used with uppercase without conflict:
- People
- Population
- Folks
- Crowd

pizza2004:
I've updated it now.

src/openrct2/peep/Peep.h
```diff
@@ -52,13 +52,15 @@ struct Ride;
...
-    Staff,
-
-    Invalid = 0xFF
-};
+    enum class Type : uint8_t
+    {
+        Guest,
+        Staff,
+        Invalid = 0xFF
+    };
```
tupaschoal, line 62: Sorry for the rework, we discussed a bit and I think the enums would suit well enough inside the `Peep` class themselves. Then we don't have to introduce a new namespace that is a synonym and it would look like:

```c++
struct Peep : SpriteBase
{
    enum class Type : uint8_t
    {
        ...
    };
    ...
}
```

And usage would be `Peep::Type::Staff`

pizza2004: I agree that would be a better way to handle it.

pizza2004:
Looking into it, I think I'll close this, since the naming on the struct would conflict with adding a namespace that the whole file can use, and I can't think of a different name to give the struct that would keep it from becoming confusing.

PR #12467, opened 2020-07-25
Remove uses of GET_PEEP macro and replace with GetEntity

duncanspumpkin:
This simplifies the code a whole bunch and reduces the chances that there is operations on the wrong types.

### DIFF
--- a/src/openrct2/interface/InteractiveConsole.cpp
+++ b/src/openrct2/interface/InteractiveConsole.cpp
@@ -469,9 +469,9 @@ static int32_t cc_staff(InteractiveConsole& console, const arguments_t& argv)
                 int_val[0] = console_parse_int(argv[2], &int_valid[0]);
                 int_val[1] = console_parse_int(argv[3], &int_valid[1]);
 
-                if (int_valid[0] && int_valid[1] && ((GET_PEEP(int_val[0])) != nullptr))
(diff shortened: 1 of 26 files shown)


src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -469,12 +469,14 @@ static int32_t cc_staff(InteractiveConsole& console, const arguments_t& argv)
...
 
-                if (int_valid[0] && int_valid[1] && ((GET_PEEP(int_val[0])) != nullptr))
+                if (int_valid[0] && int_valid[1] && ((GetEntity<Peep>(int_val[0])) != nullptr))
                 {
-                    Peep* peep = GET_PEEP(int_val[0]);
-
-                    peep->Energy = int_val[1];
-                    peep->EnergyTarget = int_val[1];
+                    Peep* peep = GetEntity<Peep>(int_val[0]);
+                    if (peep != nullptr)
```
Gymnasiast, line 475: This will never happen, as you already did this check in the `if` on line 472.

duncanspumpkin: Yeah but we will get nullptr deref compiler warnings otherwise

Gymnasiast: Then we should take it out of the if on line 472.

duncanspumpkin: Urgh sorry so tired today. Changed!

PR #12468, opened 2020-07-25
Close #12461: Refactor WINDOW_ABOUT_WIDGET_IDX to use strong enum

### LINKED ISSUE #12461: Refactor WINDOW_ABOUT_WIDGET_IDX to use strong enum

frutiemax:


src/openrct2-ui/windows/About.cpp
```diff
@@ -28,21 +28,16 @@ enum
...
+enum class WindowAboutWidgetIndex {
+    Background,
+    Title,
+    Close,
+    PageBackground,
+    TabAboutOpenRCT2,
+    TabAboutRCT2,
+    PageStart,
+    ChangeLog,
+    MusicCredits
```
Gymnasiast, line 40: You changed the values of Changelog and MusicCredits. These were set to PageStart for a good reason.

Also, it's Changelog and not ChangeLog.

src/openrct2-ui/windows/About.cpp
```diff
@@ -69,11 +64,11 @@ static rct_widget *window_about_page_widgets[] = {
 };
 
 #define DEFAULT_ENABLED_WIDGETS \
-    (1ULL << WIDX_CLOSE) | (1ULL << WIDX_TAB_ABOUT_OPENRCT2) | (1ULL << WIDX_TAB_ABOUT_RCT2)
+    (1ULL << static_cast<size_t>(WindowAboutWidgetIndex::Close)) | (1ULL << static_cast<size_t>(WindowAboutWidgetIndex::TabAboutOpenRCT2)) | (1ULL << static_cast<size_t>(WindowAboutWidgetIndex::TabAboutRCT2))
```
Gymnasiast, line 67: This is a lot uglier and harder to read than before. @ZehMatt what do you think about this? Got any suggestions to improve it?

ZehMatt: I just realized now that I accidentally added this as an issue, I wanted to avoid all UI parts, I agree.

ZehMatt:
I'm sorry to tell you this but the issue you referenced was a mistake, the UI part should not be touched as it will make it a lot more horrible with all the required casting.

PR #12470, opened 2020-07-25
Fix #11623: Erratic zoom behavior when pointing outside of the map

### LINKED ISSUE #11623: erratic zooming with mouse wheel
**OS:** Mac OSX Catalina 10.15.4
**Version:**  Version 0.2.6 (develop)
**Commit/Build:** 174b32c09

Zooming with the scroll wheel is somewhat erratic. It's expected to zoom in on the mouse pointer, and that's how Classic and previous versions of OpenRCT2 worked. Something changed in the develop version a couple of iterations ago and zooming in the current develop version is erratic, sometimes zooming to the opposite side, sometimes zooming to some random point, sometimes zooming as expected.

- [ ] Reproducible in RCT2 (vanilla)? nope
- [ ] Specific to multiplayer? unknown

**Steps to reproduce:**
1. open a map
2. zoom using the mouse wheel

**Screenshots / Video:**
Video of issue...  https://youtu.be/wR1T4N0ASlg

ZehMatt:
The way the zoom worked when 'Zoom to mouse position' is enabled was by looking for the tile the mouse pointed at, it used the same code as the interaction part, this would obviously fail if you point outside the map causing the function get_map_coordinates_from_pos to always result x=0, y=0 which is incorrect. Also get_map_coordinates_from_pos had to setup a paint session in order to find the tile pointing at which is not really required for this function. This PR instead just translates the screen coordinates to map coordinates which also adds a bit of a performance improvement which is sadly not really noticeable but its there.
Now the zoom works as expected.
b3yE6XlAJo

Closes #11623

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -37,6 +37,7 @@
 - Fix: [#11405] Building a path through walls does not always remove the walls.
 - Fix: [#11450] Rides with unsuitable track can't be opened even with "Enable all drawable track pieces" cheat.
 - Fix: [#11455] Object Selection window cuts off scenery names.
+- Fix: [#11623] Erratic zoom behavior when pointing outside of the map.
 - Fix: [#11640] Objects with a blank description in one language do not fall back to other languages anymore.
 - Fix: [#11676] Spiral Roller Coaster has regular lift hill available.
 - Fix: [#11695] Mechanics walk to tile 0, 0 at entrance only stations when trying to fix them.
--- a/src/openrct2/interface/Window.cpp
+++ b/src/openrct2/interface/Window.cpp
@@ -953,8 +953,8 @@ void window_viewport_get_map_coords_by_cursor(
     auto mouseCoords = context_get_cursor_position_scaled();
 
     // Compute map coordinate by mouse position.
-    CoordsXY mapCoords;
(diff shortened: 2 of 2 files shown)


frutiemax:
What happens if there are many elevation changes?

ZehMatt:
> What happens if there are many elevation changes? Can you check the magic mountains scenario?

I'm not sure what difference that would make, but I tested it and it works fine.

AaronVanGeffen:
Reportedly, this has been plaguing the game since a few releases ago. I propose we add a changelog entry.

duncanspumpkin:
I don't quite see how this could work with high elevations.

Gymnasiast:
@duncanspumpkin Could you test it with high elevations, then?

ZehMatt:
jvpKRCW3u6

I'm not sure of any other way to test this

frutiemax:
@ZehMatt It looks fine to me, I'll try it myself to get a feeling of it.

frutiemax:
I've tested it in Extreme Heights and there are some cases where you actually want to zoom on the mountain tile. With the new behavior, it always zooms at the center of the pointer which makes the view not target high enough.

I'd suggest to revert back to the old behavior, but if it doesn't find a tile, use the new behavior.

PR #12472, opened 2020-07-25
Reset window pointer when cleaning up after dropdown

janisozaur:


Gymnasiast:
What is the idea behind this change? To catch any instances of it being reused by accident?

janisozaur:
Yes. It gets reused just a bit down

PR #12473, opened 2020-07-25
Fix 11623: Zooming takes into account the tile height at the cursor

frutiemax:


Gymnasiast:
Just a thing about PR messages:
- if it's a fix, describe the problem, not the solution
- if it's a change, use phrasing like "make zooming take tile height into account"

PR #12475, opened 2020-07-26
Close #12460: Refactor KeyboardShortcut to use strong enum

### LINKED ISSUE #12460: Refactor KeyboardShortcut to use strong enum

frutiemax:


src/openrct2-ui/input/KeyboardShortcuts.h
```diff
@@ -24,101 +24,106 @@
...
+    IncreaseElementHeight,
+    DecreaseElementHeight,
+    ToggleClearanceChecks,
+
+    Count,
+
+    Undefined = 0xFFFF,
 };
 
+constexpr size_t get_shortcuts_count()
```
ZehMatt, line 122: We Use 'PascalCase' for functions and 'camelCase' for variables.

frutiemax: So the function with the underscore (_) are old ones?

src/openrct2-ui/windows/ShortcutKeys.cpp
```diff
@@ -87,127 +87,127 @@ struct ShortcutStringPair
...
+    { KeyboardShortcut::IncreaseYCoord,                      STR_SHORTCUT_INCREASE_Y_COORD },
+    { KeyboardShortcut::DecreaseYCoord,                      STR_SHORTCUT_DECREASE_Y_COORD },
+    { KeyboardShortcut::IncreaseElementHeight,                  STR_SHORTCUT_INCREASE_ELEM_HEIGHT },
+    { KeyboardShortcut::DecreaseElementHeight,                  STR_SHORTCUT_DECREASE_ELEM_HEIGHT },
+
+    { KeyboardShortcut::Undefined,                             STR_NONE },
+
+    { KeyboardShortcut::AdvanceToNextTick,                  STR_ADVANCE_TO_NEXT_TICK },
+    { KeyboardShortcut::PaintOriginalToggle,                 STR_SHORTCUT_PAINT_ORIGINAL },
+    { KeyboardShortcut::DebugPaintToggle,                    STR_SHORTCUT_DEBUG_PAINT_TOGGLE },
```
tupaschoal, line 210: The second arguments were aligned before, so I think we shoult try to keep them so.

src/openrct2-ui/input/KeyboardShortcuts.h
```diff
@@ -24,101 +24,106 @@
 
 struct ScreenCoordsXY;
 
-enum KeyboardShortcut
+enum class KeyboardShortcut
```
tupaschoal, line 27: There's already a class below with this exact name, but it's in the `OpenRCT2::Input` namespace, so it hints that this maybe should have a different name and also be moved there into the namespace.

Maybe this enum should become just shortcut, so that when it's moved there it becomes `OpenRCT2::Input::Shortcut`. What do you think @ZehMatt ?

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -219,7 +219,8 @@ std::string KeyboardShortcuts::GetShortcutString(int32_t shortcut) const
 ScreenCoordsXY KeyboardShortcuts::GetKeyboardMapScroll(const uint8_t* keysState) const
 {
     ScreenCoordsXY screenCoords;
-    for (int32_t shortcutId = SHORTCUT_SCROLL_MAP_UP; shortcutId <= SHORTCUT_SCROLL_MAP_RIGHT; shortcutId++)
+    for (int32_t shortcutId = static_cast<int32_t>(KeyboardShortcut::ScrollMapUp);
+         shortcutId <= static_cast<int32_t>(KeyboardShortcut::ScrollMapRight); shortcutId++)
```
tupaschoal, line 223: Isn't it better to save it as the proper enum type and cast below? The `for` condition will become shorter.

frutiemax: You cannot increment an enum type.

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -159,22 +159,22 @@ void KeyboardShortcuts::Set(int32_t key)
 
 int32_t KeyboardShortcuts::GetFromKey(int32_t key)
```
duncanspumpkin, line 160: This should return the new type instead of int32_t. that will reduce the amount of static casts.

src/openrct2-ui/input/KeyboardShortcuts.h
```diff
@@ -127,14 +32,114 @@ namespace OpenRCT2
...
+            DecreaseXCoord,
+            IncreaseYCoord,
+            DecreaseYCoord,
+            IncreaseElementHeight,
+            DecreaseElementHeight,
+            ToggleClearanceChecks,
+
+            Count,
+
+            Undefined = 0xFFFF,
```
duncanspumpkin, line 127: This should not be part of this enum it is not used for this enum it is used for SDL key undefined. Move it out into a seperate variable of type `uint16_t`

src/openrct2-ui/input/KeyboardShortcuts.h
```diff
@@ -127,14 +32,114 @@ namespace OpenRCT2
...
+        };
+
+        constexpr size_t GetShortcutsCount()
```
duncanspumpkin, line 130: I'd have to check but I thought in general we did a namespace with a constexpr value of type size_t:
```
constexpr size_t ShortcutsCount = static_cast<size_t>(Shortcut::Count);
```

src/openrct2-ui/input/KeyboardShortcuts.h
duncanspumpkin: On second reading this is used correctly in some places but not all. So you need to keep this for the undefined down in the `ShortcutList` but it should not be used in `KeyboardShortcuts::DefaultKeys`

src/openrct2-ui/windows/ShortcutKeys.cpp
```diff
@@ -302,11 +304,11 @@ static void window_shortcut_scrollmousedown(rct_window* w, int32_t scrollIndex,
         return;
 
     // Is this a separator?
-    if (ShortcutList[selected_item].ShortcutId == SHORTCUT_UNDEFINED)
+    if (ShortcutList[selected_item].ShortcutId == Input::Shortcut::Undefined)
         return;
 
     auto& shortcut = ShortcutList[selected_item];
-    window_shortcut_change_open(shortcut.ShortcutId, shortcut.StringId);
+    window_shortcut_change_open(static_cast<int32_t>(shortcut.ShortcutId), shortcut.StringId);
```
duncanspumpkin, line 311: change window_shortcut_change_open to take the correct type. Change gKeyboardShortcutChangeId to use the new type

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -159,22 +159,22 @@ void KeyboardShortcuts::Set(int32_t key)
...
         {
-            return i;
+            return static_cast<int32_t>(i);
         }
     }
-    return SHORTCUT_UNDEFINED;
+    return static_cast<int32_t>(Input::Shortcut::Undefined);
 }
 
 std::string KeyboardShortcuts::GetShortcutString(int32_t shortcut) const
```
duncanspumpkin, line 172: Use the new type

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -159,22 +159,22 @@ void KeyboardShortcuts::Set(int32_t key)
...
 {
     utf8 buffer[256] = { 0 };
     utf8 formatBuffer[256] = { 0 };
     uint16_t shortcutKey = _keys[shortcut];
-    if (shortcutKey == SHORTCUT_UNDEFINED)
+    if (shortcutKey == static_cast<uint16_t>(Input::Shortcut::Undefined))
```
duncanspumpkin, line 177: this is not Input::Shortcut::Undefined but a different variable with the same value.

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -219,7 +219,8 @@ std::string KeyboardShortcuts::GetShortcutString(int32_t shortcut) const
 ScreenCoordsXY KeyboardShortcuts::GetKeyboardMapScroll(const uint8_t* keysState) const
 {
     ScreenCoordsXY screenCoords;
-    for (int32_t shortcutId = SHORTCUT_SCROLL_MAP_UP; shortcutId <= SHORTCUT_SCROLL_MAP_RIGHT; shortcutId++)
+    for (int32_t shortcutId = static_cast<int32_t>(Input::Shortcut::ScrollMapUp);
+         shortcutId <= static_cast<int32_t>(Input::Shortcut::ScrollMapRight); shortcutId++)
     {
         uint16_t shortcutKey = _keys[shortcutId];
         uint8_t scancode = shortcutKey & 0xFF;
```
duncanspumpkin, line 226: L228 this is not Input::Shortcut::Undefined but a different variable with the same value.

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -299,78 +301,78 @@ ScreenCoordsXY get_keyboard_map_scroll(const uint8_t* keysState)
...
+    SDL_SCANCODE_EQUALS,                               // SHORTCUT_INCREASE_GAME_SPEED,
+    PLATFORM_MODIFIER | ALT | SDL_SCANCODE_C,          // SHORTCUT_OPEN_CHEAT_WINDOW,
+    SDL_SCANCODE_T,                                    // SHORTCUT_REMOVE_TOP_BOTTOM_TOOLBAR_TOGGLE,
+    SDL_SCANCODE_UP,                                   // SHORTCUT_SCROLL_MAP_UP
+    SDL_SCANCODE_LEFT,                                 // SHORTCUT_SCROLL_MAP_LEFT
+    SDL_SCANCODE_DOWN,                                 // SHORTCUT_SCROLL_MAP_DOWN
+    SDL_SCANCODE_RIGHT,                                // SHORTCUT_SCROLL_MAP_RIGHT
+    SDL_SCANCODE_C,                                    // SHORTCUT_OPEN_CHAT_WINDOW
+    PLATFORM_MODIFIER | SDL_SCANCODE_F10,              // SHORTCUT_QUICK_SAVE_GAME
+    static_cast<uint16_t>(Input::Shortcut::Undefined), // SHORTCUT_SHOW_OPTIONS
```
duncanspumpkin, line 348: this is not Input::Shortcut::Undefined but a different variable with the same value.

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -159,22 +159,22 @@ void KeyboardShortcuts::Set(int32_t key)
...
+    for (size_t i = 0; i < GetShortcutsCount(); i++)
     {
         if (key == _keys[i])
         {
-            return i;
+            return static_cast<int32_t>(i);
         }
     }
-    return SHORTCUT_UNDEFINED;
+    return static_cast<int32_t>(Input::Shortcut::Undefined);
```
duncanspumpkin, line 169: this is not Input::Shortcut::Undefined but a different variable with the same value.

src/openrct2-ui/input/KeyboardShortcuts.cpp
duncanspumpkin, line 160: This returns the new type

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -159,22 +159,22 @@ void KeyboardShortcuts::Set(int32_t key)
 
 int32_t KeyboardShortcuts::GetFromKey(int32_t key)
 {
-    for (int32_t i = 0; i < SHORTCUT_COUNT; i++)
+    for (size_t i = 0; i < GetShortcutsCount(); i++)
     {
         if (key == _keys[i])
         {
-            return i;
+            return static_cast<int32_t>(i);
```
duncanspumpkin, line 166: cast to the new type

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -146,10 +146,10 @@ bool KeyboardShortcuts::Save()
 void KeyboardShortcuts::Set(int32_t key)
 {
     // Unmap shortcut that already uses this key
-    int32_t shortcut = GetFromKey(key);
-    if (shortcut != SHORTCUT_UNDEFINED)
+    auto shortcut = GetFromKey(key);
+    if (shortcut != static_cast<int32_t>(Input::Shortcut::Undefined))
     {
-        _keys[shortcut] = SHORTCUT_UNDEFINED;
+        _keys[shortcut] = static_cast<int32_t>(Input::Shortcut::Undefined);
```
duncanspumpkin, line 152: this is not Input::Shortcut::Undefined but a different variable with the same value.

src/openrct2-ui/input/KeyboardShortcut.cpp
```diff
@@ -38,31 +38,34 @@
...
         keyboard_shortcut_handle_command(shortcut);
     }
 }
 
-void keyboard_shortcut_handle_command(int32_t shortcutIndex)
+void keyboard_shortcut_handle_command(Input::Shortcut shortcut)
 {
-    if (shortcutIndex >= 0 && static_cast<uint32_t>(shortcutIndex) < std::size(shortcut_table))
+    size_t shortcutIndex = static_cast<size_t>(shortcut);
+    if (shortcutIndex >= 0 && shortcutIndex < std::size(shortcut_table))
```
tupaschoal, line 68:
```suggestion
    if (shortcutIndex < std::size(shortcut_table))
```

Because the `enum` has `size_t` as underlying type, the `>=0` doesn't make sense anymore

src/openrct2-ui/input/KeyboardShortcuts.cpp
```diff
@@ -146,35 +146,35 @@ bool KeyboardShortcuts::Save()
 void KeyboardShortcuts::Set(int32_t key)
 {
     // Unmap shortcut that already uses this key
-    int32_t shortcut = GetFromKey(key);
-    if (shortcut != SHORTCUT_UNDEFINED)
+    auto shortcut = GetFromKey(key);
+    if (shortcut != Input::Shortcut::Undefined)
     {
-        _keys[shortcut] = SHORTCUT_UNDEFINED;
+        _keys[static_cast<size_t>(shortcut)] = static_cast<uint16_t>(Input::Shortcut::Undefined);
```
duncanspumpkin, line 152: `static_cast<uint16_t>(Input::Shortcut::Undefined)` this isn't correct. Well it is correct in terms of what it does but it's not correct because it should be `Input::ScanCodeUndefined`

src/openrct2-ui/windows/Window.h
```diff
@@ -9,12 +9,14 @@
...
 #pragma once
 
+#include <openrct2-ui/input/KeyboardShortcuts.h>
 #include <openrct2-ui/interface/Window.h>
 #include <openrct2/common.h>
 #include <openrct2/ride/Ride.h>
 #include <openrct2/windows/tile_inspector.h>
 #include <string_view>
 
+using namespace OpenRCT2;
```
duncanspumpkin, line 19: Don't add using namespace in a popular header like this your bringing the whole of the namespace in when you do not need to. Completely removing the point of namespaces.

PR #12479, opened 2020-07-26
Refactor NewsItem to use strong enums

### LINKED ISSUE #12397: Refactor NEWS_ITEM_* to use strong enum
The enum is currently unnamed and should be called NewsItemType

pizza2004:
I refactored all the NewsItem enums into a namespace and made them strong.

### DIFF
--- a/src/openrct2-ui/windows/News.cpp
+++ b/src/openrct2-ui/windows/News.cpp
@@ -152,7 +152,7 @@ static void window_news_update(rct_window* w)
         return;
 
     const auto& newsItem = gNewsItems.GetArchived()[j];
-    if (newsItem.Flags & NEWS_FLAG_HAS_BUTTON)
+    if (newsItem.HasButton())
         return;
     if (w->news.var_482 == 1)
     {
@@ -192,7 +192,7 @@ static void window_news_scrollmousedown(rct_window* w, int32_t scrollIndex, cons
     {
         if (mutableScreenCoords.y < itemHeight)
         {
-            if (newsItem.Flags & NEWS_FLAG_HAS_BUTTON || mutableScreenCoords.y < 14 || mutableScreenCoords.y >= 38
+            if (newsItem.HasButton() || mutableScreenCoords.y < 14 || mutableScreenCoords.y >= 38
                 || mutableScreenCoords.x < 328)
             {
                 buttonIndex = 0;
@@ -269,7 +269,7 @@ static void window_news_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32
         gfx_draw_string_left_wrapped(dpi, &text, { 2, y + lineHeight }, 325, STR_BOTTOM_TOOLBAR_NEWS_TEXT, COLOUR_BRIGHT_GREEN);
 
         // Subject button
-        if ((newsItem.TypeHasSubject()) && !(newsItem.Flags & NEWS_FLAG_HAS_BUTTON))
+        if ((newsItem.TypeHasSubject()) && !(newsItem.HasButton()))
         {
             auto screenCoords = ScreenCoordsXY{ 328, y + lineHeight + 4 };
 
@@ -345,7 +345,7 @@ static void window_news_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32
         }
 
         // Location button
-        if ((newsItem.TypeHasLocation()) && !(newsItem.Flags & NEWS_FLAG_HAS_BUTTON))
+        if ((newsItem.TypeHasLocation()) && !(newsItem.HasButton()))
         {
             auto screenCoords = ScreenCoordsXY{ 352, y + lineHeight + 4 };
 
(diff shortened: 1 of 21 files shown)


pizza2004:
I've managed to get it to build properly, but Mingw doesn't like that I am using default: instead of individual switch cases.  I made those changes on my end but I'm wondering if I should commit them.

src/openrct2/management/NewsItem.h
```diff
@@ -44,12 +47,17 @@ enum
     NEWS_FLAG_HAS_BUTTON = 1 << 0,
 };
 
+constexpr size_t get_news_item_type_count()
```
ZehMatt, line 50: We use 'PascalCase' for functions and 'camelCase' for variables

src/openrct2/management/NewsItem.h
```diff
@@ -18,38 +18,72 @@
...
+            return News::ItemTypeProperty::HasSubject;
+        case News::ItemType::Blank:
+            return News::ItemTypeProperty::HasLocation;
+        case News::ItemType::Research:
+        case News::ItemType::Peeps:
+        case News::ItemType::Award:
+        case News::ItemType::Graph:
+            return News::ItemTypeProperty::HasSubject;
+        default:
+            return News::ItemTypeProperty::Null;
```
tupaschoal, line 77: Handle `Count`, the only one missing, and remove the `default`

src/openrct2/management/NewsItem.h
```diff
@@ -59,16 +93,35 @@ struct NewsItem
...
-        return Type == NEWS_ITEM_NULL;
+        return Type == News::ItemType::Null;
+    }
+
+    constexpr bool HasTypeSubject() const
+    {
+        auto properties = GetNewsItemTypeProperties(Type);
+        if (properties == News::ItemTypeProperty::HasSubject || properties == News::ItemTypeProperty::HasSubjectAndLocation)
+            return true;
+        return false;
```
tupaschoal, line 104:
```suggestion
        return properties == News::ItemTypeProperty::HasSubject || properties == News::ItemTypeProperty::HasSubjectAndLocation;
```

src/openrct2/management/NewsItem.h
```diff
@@ -59,16 +93,35 @@ struct NewsItem
...
+    }
+
+    constexpr bool HasTypeLocation() const
+    {
+        auto properties = GetNewsItemTypeProperties(Type);
+        if (properties == News::ItemTypeProperty::HasLocation || properties == News::ItemTypeProperty::HasSubjectAndLocation)
+            return true;
+        return false;
```
tupaschoal, line 112:
```suggestion
        return properties == News::ItemTypeProperty::HasLocation || properties == News::ItemTypeProperty::HasSubjectAndLocation;
```

src/openrct2-ui/windows/News.cpp
```diff
@@ -285,11 +285,13 @@ static void window_news_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32
 
             switch (newsItem.Type)
             {
-                case NEWS_ITEM_RIDE:
+                case News::ItemType::Null:
+                    break;
```
tupaschoal, line 289: On this and other switches where you had to start handling extra enum values but not do anything, like `Null`, `Count` and `Blank`, I think we can make it clear if we group them all into a single `break`. What's your opinion @ZehMatt ?

Like
```c++
case News::ItemType::Null:
case News::ItemType::Blank:
case News::ItemType::Count:
    break;
```

pizza2004: I only did it because it was failing to build on anything but Mac.  I added a default and it still failed to build on MinGW.  If there's a method of making the default work on MinGW we can just use that instead.

tupaschoal: I'd rather not have a `default`, because then if we ever add a new enumeration, it will compile just fine. I prefer handling the cases individually, because then compilation reminds us to look into all cases.

That being said, when there are multiple cases doing the exact same thing, in this case nothing, I think we can group them for improved readability.

pizza2004:
I reworked stuff a little bit and applied the review fixes to each commit.  I figured there's no reason for the type properties to be a enum class.

pizza2004:
I have also done pizza2004/OpenRCT2#4 to put the entire file into a namespace if that's more preferable.

src/openrct2/scripting/ScPark.hpp
```diff
@@ -30,12 +30,16 @@ namespace OpenRCT2::Scripting
...
+        auto firstType = static_cast<uint8_t>(News::ItemType::Ride);
+
+        auto begin = std::begin(ParkMessageTypeStrings);
+        auto end = std::end(ParkMessageTypeStrings);
+
+        auto it = std::find(begin, end, key);
+        return it != end ? static_cast<News::ItemType>(firstType + std::distance(begin, it)) : News::ItemType::Blank;
     }
 
     inline std::string GetParkMessageType(uint8_t type)
```
duncanspumpkin, line 45: Make this take the specific type then static cast to `uint8_t` when required in the function.

PR #12481, opened 2020-07-26
#12422: Refactor NETWORK_COMMAND to use strong enum

ZehMatt:


src/openrct2/network/NetworkTypes.h
```diff
@@ -48,34 +48,34 @@ enum NETWORK_AUTH
     NETWORK_AUTH_UNKNOWN_KEY_DISALLOWED,
 };
 
-enum NETWORK_COMMAND
+enum class NetworkCommand : uint32_t
```
tupaschoal, line 51: As there is no `Network` class here. Should we do:
```c++
namespace Network
{
    enum class Command : uint32_t
    {
        ...
    };
}

ZehMatt: There is a Network class. I'm planning to further cleanup the network code and put that in a namespace in the next upcoming PR's

PR #12482, opened 2020-07-26
Close #12457: Refactor INTRO_STATE to use strong enum

### LINKED ISSUE #12457: Refactor INTRO_STATE to use strong enum

frutiemax:


src/openrct2/Intro.h
```diff
@@ -14,23 +14,23 @@
 
 struct rct_drawpixelinfo;
 
-enum INTRO_STATE
+enum class IntroState : uint8_t
```
tupaschoal, line 17: Note for future, I think we'd probably want to go as:
```c++
namespace OpenRCT2::Intro
{
    enum class State : uint8_t
    {
        ...
    };
}
```

PR #12485, opened 2020-07-26
Refactor the Network a bit, setting the stones for further refactor work

ZehMatt:
This branch is based on #12481 and I don't want to do too much refactoring at once to avoid rebasing all the time as there is a lot to be done.

I've categorized all the members a bit to make this process a bit easier for me, I will do this gradually as I learned my lesson from attempting this before and it ended up being a mess that I couldn't keep up with.

I will rebase this branch once #12481 is merged.


src/openrct2/network/Socket.cpp
```diff
@@ -22,6 +22,8 @@
 #include <cmath>
 
 #ifdef _WIN32
+    #pragma comment(lib, "Ws2_32.lib")
```
duncanspumpkin, line 25: What is this here for?

ZehMatt: This links the winsock library, used to be in Network.cpp rather than Socket

duncanspumpkin:
Rebase and this is good to go

PR #12488, opened 2020-07-26
Refactor RAIN_LEVEL to use strong enum 

wdarvin:


src/openrct2/world/Climate.h
```diff
@@ -38,19 +38,19 @@ enum WEATHER_EFFECT
     WEATHER_EFFECT_STORM,
 };
 
-enum RAIN_LEVEL
+enum class RainLevel_e
```
ZehMatt, line 41: We don't suffix/prefix our enums, please remove the suffix.

wdarvin: How should handle the conflict with WeatherState::RainLevel?

ZehMatt: You can rename it to Level as the type already states its RainLevel

wdarvin:
Did rebase/force push to get branch up to date because I messed that up initially.  I'm less familiar with rebasing than merging so here's hoping I didn't screw anything up, but everything should be good now.

Issue #12337, opened 2020-07-21
Remove goto from DecideAndBuyItem


tupaschoal:
https://github.com/OpenRCT2/OpenRCT2/blob/d10b615eb33f6cd597340d83d68eaaa97c1a493e/src/openrct2/peep/Guest.cpp#L1541-L1551

batorni:
I can pick this one up

tupaschoal:
Go ahead

rpstester:
@batorni @tupaschoal  is anyone still working on this issue?

tupaschoal:
No, we closed the PR above for lack of activity. If you intend on picking this up, please give the comments in the PR a read before starting 😃

rpstester:
It looks like all this needs is an else clause to avoid the " if ((shopItem != SHOP_ITEM_MAP)" clause, correct?  So 
if umbrella and raining
//do nothing
else if map and souvenir and no voucher...
end if
if no voucher...

tupaschoal:
It isn't hard, but it's not that simple I'd say, note that there are two gotos in the function. My recommendation is doing it in a series of commits, which we can individually review, but squash later.
Extract one of the `goto` blocks into a function and call it where it used to be.
Extract the other
Merge if applicable

rpstester:
How would you like it if I split the decide code from the buy code?  You know, so each function/method does one basic task.  Would that be outside the scope of this issue?

tupaschoal:
As long as you do it in separate commits, so it becomes easier to review, it can be in the same PR. I mean, only one of your commits can be "Fix #....", but you can have some before and after to refactor the vicinity if it makes for a more readable code.

PR #12490, opened 2020-07-26
Updated Guest.cpp per issue 12337 to remove gotos

batorni:
Updated Guest.cpp per issue #12337 to remove goto's from DecideAndBuyItem()

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1554,7 +1554,7 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
         else if (GuestNumRides < 3)
             return false;
     }
-    else 
+    else
     {
         if (!hasVoucher)
         {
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Replays are failing suggesting you have modified the logic of the function.

batorni:
@duncanspumpkin I'll take a look. Is there any way to run the regression tests locally? The testing page on the wiki didn't detail anything about running those tests.

janisozaur:
Yes, you can: just use `ctest`, after building with `WITH_TESTS=on`. The replay tests will auto-discover replays, so they need to be placed to proper data path beforehand, I think it's just `ln -s <openrct2_repo>/tests/testdata` in the dir where you have the executables built.

duncanspumpkin:
You can also manually Replay a replay in game with the console type `replay_start <replay name>`. It's just bpb1000Ticks which is failing which is our basic busy map that has to run for 1000 ticks without the sprite data changing from the expect. If you run the replay manually then it will generate a desync report at the first and last tick telling you what fields are different.

tupaschoal:
@batorni I'm converting this PR to draft, if you resume work and fix the failures, make sure to click "ready to review"

Gymnasiast:
I'll have to close this PR, for the following reasons:

- The OP has not replied since 27 July
- It fails tests
- The commit history is hard to follow.

Of course, we would welcome another attempt. If you feel like doing that, please keep every `goto` elimination in its own commit, so that we can properly review it - otherwise, it's way too likely that an error slips in.

PR #12491, opened 2020-07-26
Add OpenURL method, with implementation for Linux and Windows

janisozaur:


src/openrct2-ui/UiContext.Win32.cpp
```diff
@@ -97,6 +97,12 @@ namespace OpenRCT2::Ui
             ShellExecuteW(NULL, L"open", pathW.c_str(), NULL, NULL, SW_SHOWNORMAL);
         }
 
+        void OpenURL(const std::string& url) override
+        {
+            std::wstring urlW = String::ToWideChar(url);
+            ShellExecuteW(NULL, L"open", urlW.c_str(), NULL, NULL, SW_SHOWNORMAL);
```
Broxzier, line 103: Why not use `ShellExecuteA` instead? Then there is no need to convert to long strings.

janisozaur: This is how OpenFolder does it

Gymnasiast: @Broxzier That won't work properly if the string contains non-ASCII characters, right? At least using widechar means it should work with any valid URL.

IntelOrca: Yes, we generally assume all `std::string` and `char*` are UTF-8 in OpenRCT2, so we should use the widechar APIs.

PR #12492, opened 2020-07-26
Fix formatting of UiContext.macOS.mm

janisozaur:


AaronVanGeffen:
This was done by applying clang-format, right? We should probably add `*.mm` to the clang-format CI job on GitHub Actions, then.

PR #12493, opened 2020-07-26
Fix: bad kerning on lowercase 'j' character (original bug)

Gymnasiast:
Before:
afbeelding

After:
afbeelding


Gymnasiast:
I have now run them through optipng. I wanted to make sure it didn't cause any unwanted side effects, but it looks like it all went well.

PR #12494, opened 2020-07-27
Reduce the use of GET_VEHICLE

duncanspumpkin:
Added a load of nullptr checks as well.

### DIFF
--- a/src/openrct2-ui/windows/Map.cpp
+++ b/src/openrct2-ui/windows/Map.cpp
@@ -1097,7 +1097,11 @@ static void window_map_paint_train_overlay(rct_drawpixelinfo* dpi)
         for (auto vehicle_index = train->sprite_index; vehicle_index != SPRITE_INDEX_NULL;
              vehicle_index = vehicle->next_vehicle_on_train)
         {
-            vehicle = GET_VEHICLE(vehicle_index);
+            vehicle = GetEntity<Vehicle>(vehicle_index);
+            if (vehicle == nullptr)
+            {
+                break;
+            }
             if (vehicle->x == LOCATION_NULL)
                 continue;
 
--- a/src/openrct2-ui/windows/Map.cpp
+++ b/src/openrct2-ui/windows/Map.cpp
@@ -1093,15 +1093,8 @@ static void window_map_paint_train_overlay(rct_drawpixelinfo* dpi)
 {
     for (auto train : EntityList<Vehicle>(EntityListId::TrainHead))
     {
-        Vehicle* vehicle = nullptr;
-        for (auto vehicle_index = train->sprite_index; vehicle_index != SPRITE_INDEX_NULL;
-             vehicle_index = vehicle->next_vehicle_on_train)
+        for (Vehicle* vehicle = train; vehicle != nullptr; vehicle = GetEntity<Vehicle>(vehicle->next_vehicle_on_train))
         {
-            vehicle = GetEntity<Vehicle>(vehicle_index);
-            if (vehicle == nullptr)
-            {
-                break;
-            }
             if (vehicle->x == LOCATION_NULL)
                 continue;
 
--- a/src/openrct2/actions/SetCheatAction.hpp
+++ b/src/openrct2/actions/SetCheatAction.hpp
@@ -657,7 +657,11 @@ private:
                 spriteIndex = trainIndex;
                 while (spriteIndex != SPRITE_INDEX_NULL)
                 {
-                    auto vehicle = GET_VEHICLE(spriteIndex);
+                    auto vehicle = GetEntity<Vehicle>(spriteIndex);
+                    if (vehicle == nullptr)
+                    {
+                        break;
+                    }
                     for (size_t i = 0, offset = 0; i < vehicle->num_peeps; i++)
                     {
                         while (vehicle->peep[i + offset] == SPRITE_INDEX_NULL)
--- a/src/openrct2/actions/SetCheatAction.hpp
+++ b/src/openrct2/actions/SetCheatAction.hpp
@@ -641,7 +641,6 @@ private:
 
     void RemoveAllGuests() const
     {
-        uint16_t spriteIndex;
         for (auto& ride : GetRideManager())
         {
             ride.num_riders = 0;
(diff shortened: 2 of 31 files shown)


src/openrct2-ui/windows/Map.cpp
```diff
@@ -1097,7 +1097,11 @@ static void window_map_paint_train_overlay(rct_drawpixelinfo* dpi)
         for (auto vehicle_index = train->sprite_index; vehicle_index != SPRITE_INDEX_NULL;
              vehicle_index = vehicle->next_vehicle_on_train)
         {
-            vehicle = GET_VEHICLE(vehicle_index);
+            vehicle = GetEntity<Vehicle>(vehicle_index);
+            if (vehicle == nullptr)
```
ZehMatt, line 1101: If such sprite index ends up on the train then the state is already bad. We should have a warning here or it may go unnoticed.

duncanspumpkin: I see those sorts of checks as something that should happen at load.

duncanspumpkin: I also want to move these sorts of train iterators into the train super class.

src/openrct2/peep/Guest.cpp
```diff
@@ -2487,11 +2487,18 @@ static Vehicle* peep_choose_car_from_ride(Peep* peep, Ride* ride, std::vector<ui
...
+        return nullptr;
+    }
     for (int32_t i = peep->CurrentCar; i > 0; --i)
     {
-        vehicle = GET_VEHICLE(vehicle->next_vehicle_on_train);
+        vehicle = GetEntity<Vehicle>(vehicle->next_vehicle_on_train);
+        if (vehicle == nullptr)
+        {
+            return nullptr;
+        }
```
tupaschoal, line 2501: I wonder if it's clearer if we do:
```c++
    auto vehicle = GetEntity<Vehicle>(ride->vehicles[peep->CurrentTrain]);
    for (int32_t i = peep->CurrentCar; i > 0 && vehicle != nullptr; --i)
    {
        vehicle = GetEntity<Vehicle>(vehicle->next_vehicle_on_train);
    }
    return vehicle;
```

src/openrct2/peep/Guest.cpp
```diff
@@ -2503,6 +2510,10 @@ static Vehicle* peep_choose_car_from_ride(Peep* peep, Ride* ride, std::vector<ui
  */
 static void peep_choose_seat_from_car(Peep* peep, Ride* ride, Vehicle* vehicle)
 {
+    if (vehicle == nullptr)
```
tupaschoal, line 2513: It's kind of unrelated to this PR, but maybe we should check for the other pointers

src/openrct2/peep/Guest.cpp
```diff
@@ -3640,7 +3655,12 @@ static void peep_update_ride_leave_entrance_waypoints(Peep* peep, Ride* ride)
 
     uint8_t direction_track = (tile_element == nullptr ? 0 : tile_element->GetDirection());
 
-    auto vehicle = GET_VEHICLE(ride->vehicles[peep->CurrentTrain]);
+    auto vehicle = GetEntity<Vehicle>(ride->vehicles[peep->CurrentTrain]);
+    if (vehicle == nullptr)
+    {
+        // TODO: Goto ride exit on failure.
```
tupaschoal, line 3661: Don't forget to file an issue for this

src/openrct2/peep/Guest.cpp
```diff
@@ -3973,10 +3993,19 @@ void Guest::UpdateRideFreeVehicleCheck()
         return;
     }
 
-    Vehicle* vehicle = GET_VEHICLE(ride->vehicles[CurrentTrain]);
+    Vehicle* vehicle = GetEntity<Vehicle>(ride->vehicles[CurrentTrain]);
+    if (vehicle == nullptr)
+    {
+        // TODO: Leave ride on failure goes for all returns on nullptr in this function
```
tupaschoal, line 3999: Same thing about the issue

src/openrct2/peep/Guest.cpp
```diff
@@ -4126,12 +4168,19 @@ void Guest::UpdateRideLeaveVehicle()
     if (ride == nullptr)
         return;
 
-    Vehicle* vehicle = GET_VEHICLE(ride->vehicles[CurrentTrain]);
+    Vehicle* vehicle = GetEntity<Vehicle>(ride->vehicles[CurrentTrain]);
+    if (vehicle == nullptr)
+        return;
```
tupaschoal, line 4173: Just to be consistent with most of the other places.
```suggestion
    if (vehicle == nullptr)
    {
        return;
    }
```

src/openrct2/ride/CableLift.cpp
```diff
@@ -373,6 +385,10 @@ int32_t Vehicle::CableLiftUpdateTrackMotion()
 
     for (Vehicle* vehicle = frontVehicle;;)
     {
+        if (vehicle == nullptr)
+        {
+            break;
+        }
```
tupaschoal, line 391:
```suggestion
    for (Vehicle* vehicle = frontVehicle; vehicle != nullptr;)
    {
```

src/openrct2/ride/Ride.cpp
```diff
@@ -4618,13 +4629,17 @@ static void vehicle_unset_update_flag_b1(Vehicle* head)
     Vehicle* vehicle = head;
     while (true)
     {
+        if (vehicle == nullptr)
+        {
+            return;
+        }
```
tupaschoal, line 4635:
```suggestion
    while (vehicle != nullptr)
    {
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -2508,7 +2521,11 @@ void Vehicle::UpdateWaitingToDepart()
             uint16_t spriteId = sprite_index;
             for (Vehicle* curVehicle; spriteId != SPRITE_INDEX_NULL; spriteId = curVehicle->next_vehicle_on_train)
```
tupaschoal, line 2522:
```suggestion
            for (Vehicle* curVehicle = GetEntity<Vehicle>(spriteId); curVehicle != nullptr && spriteId != SPRITE_INDEX_NULL; spriteId = curVehicle->next_vehicle_on_train)
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -4092,7 +4145,10 @@ void Vehicle::UpdateUnloadingPassengers()
         uint16_t spriteId = sprite_index;
         for (Vehicle* train; spriteId != SPRITE_INDEX_NULL; spriteId = train->next_vehicle_on_train)
```
tupaschoal, line 4146: Initialize `train` on `for`? (Like suggested in some other places)

src/openrct2/ride/Vehicle.cpp
```diff
@@ -4118,7 +4174,9 @@ void Vehicle::UpdateUnloadingPassengers()
     uint16_t spriteId = sprite_index;
     for (Vehicle* train; spriteId != SPRITE_INDEX_NULL; spriteId = train->next_vehicle_on_train)
```
tupaschoal, line 4175: Initialize `train` on `for`? (Like suggested in some other places)

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8258,7 +8337,13 @@ bool Vehicle::UpdateTrackMotionForwards(rct_ride_entry_vehicle* vehicleEntry, Ri
     remaining_distance = -1;
 
     // Might need to be bp rather than this, but hopefully not
-    auto head = (GET_VEHICLE(otherVehicleIndex))->TrainHead();
+    auto otherVeh = GetEntity<Vehicle>(otherVehicleIndex);
+    if (otherVeh == nullptr)
+    {
+        // Should never happen
```
tupaschoal, line 8343: Add a `log` so we can find out if it does

src/openrct2/peep/Guest.cpp
duncanspumpkin: There's actually a function for replacing all of this. I'm leaving that for a later refactor though.

src/openrct2/peep/Guest.cpp
duncanspumpkin: I was being consistent within the function.

src/openrct2/actions/SetCheatAction.hpp
```diff
@@ -654,32 +653,21 @@ DEFINE_GAME_ACTION(SetCheatAction, GAME_COMMAND_CHEAT, GameActionResult)
 
             for (auto trainIndex : ride.vehicles)
             {
-                spriteIndex = trainIndex;
-                while (spriteIndex != SPRITE_INDEX_NULL)
+                for (Vehicle* vehicle = TryGetEntity<Vehicle>(trainIndex); vehicle != nullptr;
```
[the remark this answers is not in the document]
duncanspumpkin: Game actions shouldn't really be using the none try version as you don't want to crash the server from client input. It probably would be fine as this is not a client specified variable but i thought I'd just do a blanket rule that game actions should use try

tupaschoal:
@duncanspumpkin are you following up soon on the comments left open above? If not I think we should open issues to track them

PR #12496, opened 2020-07-27
Use -fPIE for amd64 builds for Linux

janisozaur:
This is standard on some distributions, a good practice to play along
with ASLR and changes a bit what compiler knows about possible null
values, for which it complained in Launchpad builds in the past.


Gymnasiast:
Why don't we put this in the CMake file so it does this by default?

janisozaur:
Not all the compilers/systems must support it. I've also found a better way to enable it, via cmake option, so used that instead

PR #12502, opened 2020-07-27
Fix #12486: `set-rct2` has a broken g1.dat check

### LINKED ISSUE #12486: Path check to g1.dat accidentally broken
**OS:** Linux
**Version:** 0.2.6
**Commit/Build:** 9db2aac3b

https://github.com/OpenRCT2/OpenRCT2/blob/ba10b8478961845efeeda6ff3b956d34b770c0b4/src/openrct2/cmdline/RootCommands.cpp#L370-L375

Currently the command `openrct2 set-rct2` fails if the path is valid. The `!` was deleted when the called function was changed, so I think it was accidentally.

autinerd:
This fixes #12486, where the path of g1.dat is valid in `openrct2 set-rct2` when it doesn't exist.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -148,6 +148,7 @@ The following people are not part of the development team, but have been contrib
 * Reid Baris (Rdbaris)
 * Deanna Baris (dbaris)
 * Chaitanya Thengdi (chaitanyathengdi)
+* Sidney Kuyateh (autinerd)
 
 ## Toolchain
 * (Balletie) - macOS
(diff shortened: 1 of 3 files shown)


Gymnasiast:
Could you add an entry to `distribution/changelog.txt`?
```
Fix: [#12486] `set-rct2` has a broken g1.dat check.
```
Put it together with the other "Fix" entries for 0.2.6+, sorted by the issue number.


If you like, you can also add your name to the `contributors.md` file, under "bugfixes".

autinerd:
Done, thank you!

PR #12503, opened 2020-07-27
Refactor another set of widget definitions

AaronVanGeffen:
Part of #12329. Let's get these done. :)

### DIFF
--- a/src/openrct2-ui/windows/TitleEditor.cpp
+++ b/src/openrct2-ui/windows/TitleEditor.cpp
@@ -135,55 +135,53 @@ enum WINDOW_TITLE_EDITOR_WIDGET_IDX {
 
 // Increase BW if certain languages do not fit
 // BW should be a multiple of 4
+static constexpr const rct_string_id WINDOW_TITLE = STR_TITLE_EDITOR_TITLE;
 static constexpr const int32_t WW = 320;
 static constexpr const int32_t WH = 270;
-constexpr int32_t BX = 8;
-constexpr int32_t BW = 72;
-constexpr int32_t BY = 52;
-constexpr int32_t BH = 63;
-constexpr int32_t BS = 18;
-constexpr int32_t SCROLL_WIDTH = 350;
 static constexpr const int32_t WH2 = 127;
+static constexpr const ScreenSize ACTION_BTN = { 91, 14 };
+static constexpr const ScreenSize EDIT_BTN = { 72, 14 };
+static constexpr const ScreenSize PLAYBACK_BTN = { 18, 17 };
+static constexpr const ScreenSize REORDER_BTN = { 36, 14 };
+
+static constexpr int32_t SCROLL_WIDTH = 350;
 
 static rct_widget window_title_editor_widgets[] = {
-    { WWT_FRAME,            0,  0,      WW-1,   0,      WH2-1,  0xFFFFFFFF,             STR_NONE },                             // panel / background
-    { WWT_CAPTION,          0,  1,      WW-2,   1,      14,     STR_TITLE_EDITOR_TITLE, STR_WINDOW_TITLE_TIP },                 // title bar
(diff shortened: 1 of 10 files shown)


src/openrct2-ui/windows/TitleEditor.cpp
```diff
@@ -135,55 +135,53 @@ enum WINDOW_TITLE_EDITOR_WIDGET_IDX {
...
-    { WWT_IMGBTN,           1,  BX+BW/4,BX+BW/2-1,  WH-32,  WH-16,      SPR_G2_TITLE_STOP,      STR_TITLE_EDITOR_ACTION_STOP_TIP }, // Stop
-    { WWT_IMGBTN,           1,  BX+BW/2,BX+BW*3/4-1,WH-32,  WH-16,      SPR_G2_TITLE_PLAY,      STR_TITLE_EDITOR_ACTION_PLAY_TIP }, // Play
-    { WWT_IMGBTN,           1,  BX+BW*3/4,BX+BW,    WH-32,  WH-16,      SPR_G2_TITLE_SKIP,      STR_TITLE_EDITOR_ACTION_SKIP_TIP }, // Skip
-
+    MakeWidget     ({   8,  52},     EDIT_BTN, WWT_BUTTON,          1, STR_TITLE_EDITOR_ACTION_INSERT,    STR_TITLE_EDITOR_ACTION_INSERT_TIP            ), // Insert
+    MakeWidget     ({   8,  70},     EDIT_BTN, WWT_BUTTON,          1, STR_TITLE_EDITOR_ACTION_EDIT,      STR_TITLE_EDITOR_ACTION_EDIT_TIP              ), // Edit
+    MakeWidget     ({   8,  88},     EDIT_BTN, WWT_BUTTON,          1, STR_TRACK_MANAGE_DELETE,           STR_TITLE_EDITOR_ACTION_DELETE_TIP            ), // Delete
+    // MakeWidget  ({   8, 106},     EDIT_BTN, WWT_DROPDOWN_BUTTON, 1, STR_TITLE_EDITOR_ACTION_RELOAD,    STR_TITLE_EDITOR_ACTION_RELOAD_TIP            ), // Reload
```
Gymnasiast, line 175: Perhaps get rid of this?

src/openrct2-ui/windows/ViewClipping.cpp
```diff
@@ -45,20 +45,20 @@ static DISPLAY_TYPE gClipHeightDisplayType = DISPLAY_TYPE::DISPLAY_UNITS;
 
 #pragma region Widgets
 
+static constexpr const rct_string_id WINDOW_TITLE = STR_VIEW_CLIPPING_TITLE;
```
Gymnasiast, line 48: Why did you define a constant for this, seeing as this is only used once?

src/openrct2-ui/windows/TitleEditor.cpp
AaronVanGeffen: I considered it, but commented references to `WIDX_TITLE_EDITOR_RELOAD` are found throughout the file. Perhaps a little out of scope.

src/openrct2-ui/windows/ViewClipping.cpp
AaronVanGeffen: Well, I took that to be the convention. All windows that use the `WINDOW_SHIM` macro have a `WINDOW_TITLE` constant defined, unless there are several tabs involved. (I used `grep -RE \\bWINDOW_TITLE\\b *.cpp src/openrct2-ui/windows` to confirm.)

Gymnasiast: It is mostly a remnant from when the original author of the window shim wrote it - initially, you had to #define the WINDOW_TITLE in the window definition file and this would then be picked up by the window shim #define. IOW, it would act like a global variable. I changed WINDOW_SHIM to simply take variables instead, but I didn't bother changing all the files. So it's not really a convention.

This can be left as-is, btw.

AaronVanGeffen: Ah, alright, thanks for the clarification. Perhaps we can simplify it eventually, then.

PR #12508, opened 2020-07-28
Fix #12505: Stores can only have the first product advertised

### LINKED ISSUE #12505: Stores selling multiple items can only have the first products advertised
I decided to get back into OpenRCT2 after many years and I'm so happy to be back!
I was wondering if it is possible to make it so that stores selling both food and drink can have one or the other in marketing campaigns. As of the moment, you can only advertise the top-most item. E.G: If you advertise the contents of an "Art Deco Food Stall", You will only be able to advertise Beef Noodles. Essentially, the result of adding the feature should allow advertising for either Beef Noodles or Lemonade.

Hopefully that explains enough. If you do need more context or info, just say so. It's great to be back! Looking forward to hearing from you!

Gymnasiast:


src/openrct2-ui/windows/NewCampaign.cpp
```diff
@@ -172,10 +172,12 @@ static void window_new_campaign_get_shop_items()
         auto rideEntry = ride.GetRideEntry();
         if (rideEntry != nullptr)
         {
-            auto itemType = rideEntry->shop_item[0];
-            if (itemType != SHOP_ITEM_NONE && ShopItems[itemType].IsFoodOrDrink())
+            for (auto itemType : rideEntry->shop_item)
```
tupaschoal, line 175:
```suggestion
            for (const auto itemType : rideEntry->shop_item)
```

janisozaur:
No network version change?

janisozaur:
I guess not needed, looked at it again and seems like the value is fully formed in UI and only passed to core later on

PR #12510, opened 2020-07-28
Fix #12506: Cannot advertise food if there are no rides in the park

### LINKED ISSUE #12506: Unable to select food/drink type when starting a marketing campaign for free food/drink vouchers if there are no rides in park
**OS:** Windows 8
**Version:** 0.2.6
**Commit/Build:** 59c2e01

I was trying to experiment with things to see if there's a workaround for #12505, I found that I actually couldn't start free food/drink marketing campaigns at all. I was running a pretty old build -- at least a few months old -- so I figured it might have been fixed since then, so I upgraded to the latest build ... and I was shocked to see that it still didn't work.

When I try to select which food item to give free vouchers for on this screen:
Forest Frontiers  120-06-27 23-08-36
every time I clicked on the item in the list, it would make the list go away, but the item selected would remain (not selected). 

This completely breaks this type of marketing campaign because it won't let you start the campaign without selecting an item first. I tried a variety of different stall types, having multiple stalls or just a single one, and tried making the campaign longer ... still exactly the same behavior.

Edit: tried some more experimentation and found out that it works fine on Dynamite Dunes. This made me suspicious, given that Dynamite Dunes has a coaster at the beginning.
Went back to a fresh Forest Frontiers, built only a food stall, tried to advertise free burgers -- didn't work.
Built a twist in that park, then tried to advertise the free burgers again -- it worked.

**Steps to reproduce:**
1. Open a new Forest Frontiers (or probably any scenario that doesn't start with any rides)
2. Build any food stall (do not build any rides)
3. Go to the marketing campaigns
4. Select 'vouchers for free food or drink'
5. Try to select which food/drink to make vouchers for.

Gymnasiast:


ocalhoun6:
It's awesome how fast you guys are at fixing bugs sometimes!

Issue #12395, opened 2020-07-25
Refactor PeepItem to use strong enum


ZehMatt:

danty178:
Hi everyone, it was my first contribute to the community, I'm not sure if I've pushed my commit to the right place. Please take account the my last commit and ignore the two others. I'm sorry for the inconvenience.

tupaschoal:
@danty178 you pushed to your own fork, you now need to open a pull request to ours: 
https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request

PR #12513, opened 2020-07-29
Refactor #12395 PeepItem to strong enum

danty178:
I hope this time I have pulled the commit to the right place. Please take account commit 3bb77f1 and ignore the rest. Thanks for the help of tupaschoal.


duncanspumpkin:
annoyingly this can't become a strong enum without a ton of casts. This is a limitation with c++ atm maybe it will be fixed in 2023.
I think our preferred style for arrays of flags like this is to do a namespace of constexprs. So it would be something like:

```
namespace PeepItem{
    constexpr uint32_t Balloon = (1 << 0);
    constexpr uint32_t Photo = (1 << 1);
    ...
}
```

ZehMatt:
I tried to deliberately leave enums used as flags out of the enum refactor project but this one sadly slipped. I think we should leave this as is for now. @duncanspumpkin @tupaschoal what are your thoughts?

ZehMatt:
Also on another thought it still makes sense to have PeepItem as a strong enum but not having each entry value defined as a flag. When we require a flag for such item we can always do something like
```
1u << static_cast<uint32_t>(item)
```
Because this usage is rather specific to how each item is stored.
So in general it would be probably best to have 
```Peep::AddItem, Peep::RemoveItem, Peep::HasItem, Peep::ClearItems``` hiding the internal data format entirely.

duncanspumpkin:
Yes that would be a good idea. This one is especially important due to the extra flags so the offsets overlap. We should fix that at load so that a 64bit field is used. Should we close this PR?

tupaschoal:
If there is an issue to track the prerequisite, I'm ok with closing it

tupaschoal:
@duncanspumpkin can this be resumed now that we have the method to convert enums to flag? If not, maybe we should close as you suggested

duncanspumpkin:
These fields are all flags at all times. We could swap it around though to change it so that they become enums. I think making it a namespace of constexprs would be the better approach personally.

duncanspumpkin:
I think this should be closed. `SHOP_ITEM_xxx` is the enum version of these where `PEEP_ITEM_xxx` is `1 << SHOP_ITEM_xxx`. We need to unify these two types.

tupaschoal:
Ok, closing then. Sorry @danty178

PR #12526, opened 2020-07-30
Eliminate Win32 garbage

ZehMatt:
I've added WIN32_LEAN_AND_MEAN to the common.props and it revealed that we have a bunch of include errors which I fix in this PR. This also revealed that we have a clash with ```define interface struct``` and ```interface IStream``` so I had to rename interface to INTERFACE, we should discuss on dropping this define. Also to eliminate the name clash of IStream I moved IStream, MemoryStream, FileStream into OpenRCT2 namespace, we want to re-organize into namespaces so its a good start.

### DIFF
--- a/openrct2.common.props
+++ b/openrct2.common.props
@@ -52,7 +52,7 @@
              C4549: 'operator': operator before comma has no effect; did you intend 'operator'?
              C4555: expression has no effect; expected expression with side-effect
       -->
-      <PreprocessorDefinitions>__AVX2__;__SSE4_1__;OPENGL_NO_LINK;_CRT_SECURE_NO_WARNINGS;_USE_MATH_DEFINES;SDL_MAIN_HANDLED;_WINSOCK_DEPRECATED_NO_WARNINGS;NOMINMAX;%(PreprocessorDefinitions)</PreprocessorDefinitions>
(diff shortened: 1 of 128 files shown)


Gymnasiast:
I'm not sure I like having use INTERFACE. Shouty typedefs is one of the Windows idiosyncrasies that should be limited to the platform code. Why does defining WIN32_LEAN_AND_MEAN even mean we have to redefine a language construct like `interface`?

IntelOrca:
Just replace all instances of `interface` with `struct`. I don't know why I thought it was a good idea.

ZehMatt:
Why a struct? I think interfaces should be class, structs should stay structs for POD things.

IntelOrca:
Because it doesn't make sense for interfaces to have private members. If you use class, you'd have to add `public:` everywhere.

ZehMatt:
> Because it doesn't make sense for interfaces to have private members. If you use class, you'd have to add `public:` everywhere.

Just once per interface really, but whatever, I'll eliminate this madness.

> I'm not sure I like having use INTERFACE. Shouty typedefs is one of the Windows idiosyncrasies that should be limited to the platform code. Why does defining WIN32_LEAN_AND_MEAN even mean we have to redefine a language construct like `interface`?

Windows is special, it seems to include different things while it should include less, its annoying and stupid, either way globa namespace is bad.

IntelOrca:
I don't think `WIN32_LEAN_AND_MEAN` should be defined as a project directive. Some files require some headers that are not included if you define `WIN32_LEAN_AND_MEAN`. Any files, which should mostly be limited to platform should define it if necessary above the windows.h and co includes.

ZehMatt:
Seems to be building fine on mingw.

PR #12527, opened 2020-07-30
Refactor usage of TextPaint

Gymnasiast:
Eliminating a global variable in the process.


duncanspumpkin:
It's a tiny structure the us almost no way there is performance issues. I only see this improving things

PR #12528, opened 2020-07-30
Pass the formatter directly to scrolling_text_setup

duncanspumpkin:
Just a little refactor to make it clear where this global var is coming from.

Noticed that we have a massive issue with multithreading in this function as well since the global data that this touches will be trashed by the next thread regardless of the mutex as the global data must be kept correct until the end of the rendering which would of course mean multithreading wouldn't work.

### DIFF
--- a/src/openrct2/drawing/Drawing.h
+++ b/src/openrct2/drawing/Drawing.h
@@ -694,8 +694,12 @@ void ttf_draw_string(rct_drawpixelinfo* dpi, const_utf8string text, int32_t colo
 // scrolling text
 void scrolling_text_initialise_bitmaps();
 void scrolling_text_invalidate();
+
+class Formatter;
+
 int32_t scrolling_text_setup(
-    struct paint_session* session, rct_string_id stringId, uint16_t scroll, uint16_t scrollingMode, colour_t colour);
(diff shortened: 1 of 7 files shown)


src/openrct2/drawing/Drawing.h
```diff
@@ -694,8 +694,11 @@ void ttf_draw_string(rct_drawpixelinfo* dpi, const_utf8string text, int32_t colo
 // scrolling text
 void scrolling_text_initialise_bitmaps();
 void scrolling_text_invalidate();
+
+class Formatter;
+
 int32_t scrolling_text_setup(
-    struct paint_session* session, rct_string_id stringId, uint16_t scroll, uint16_t scrollingMode, colour_t colour);
+    struct paint_session* session, rct_string_id stringId, Formatter& args, uint16_t scroll, uint16_t scrollingMode, colour_t colour);
```
tupaschoal, line 701:
```suggestion
    struct paint_session* session, rct_string_id stringId, Formatter& ft, uint16_t scroll, uint16_t scrollingMode, colour_t colour);
```

It's already called `ft` everywhere, so I think it would be useful keeping it as such

PR #12530, opened 2020-07-31
Reduce the use of rct_sprite

duncanspumpkin:
Couple little refactors getting everything to use the `SpriteBase`

### DIFF
--- a/src/openrct2-ui/windows/TitleCommandEditor.cpp
+++ b/src/openrct2-ui/windows/TitleCommandEditor.cpp
@@ -613,62 +613,55 @@ static void window_title_command_editor_tool_down(
 
     if (info.type == VIEWPORT_INTERACTION_ITEM_SPRITE)
     {
-        uint16_t spriteIndex = info.sprite->generic.sprite_index;
-        uint16_t spriteIdentifier = info.sprite->generic.sprite_identifier;
         bool validSprite = false;
-        if (info.sprite->generic.Is<Peep>())
+        auto peep = info.sprite->As<Peep>();
+        auto vehicle = info.sprite->As<Vehicle>();
+        auto litter = info.sprite->As<Litter>();
+        auto duck = info.sprite->As<Duck>();
+        auto balloon = info.sprite->As<Balloon>();
+        if (peep != nullptr)
         {
             validSprite = true;
-            auto peep = GetEntity<Peep>(spriteIndex);
-            if (peep != nullptr)
+            uint8_t formatArgs[32]{};
+            Formatter ft(formatArgs);
+            peep->FormatNameTo(ft);
+            format_string(command.SpriteName, USER_STRING_MAX_LENGTH, STR_STRINGID, &peep->Id);
+        }
+        else if (vehicle != nullptr)
+        {
+            validSprite = true;
+
+            auto ride = vehicle->GetRide();
+            if (ride != nullptr)
             {
                 uint8_t formatArgs[32]{};
                 Formatter ft(formatArgs);
-                peep->FormatNameTo(ft);
-                format_string(command.SpriteName, USER_STRING_MAX_LENGTH, STR_STRINGID, &peep->Id);
+                ride->FormatNameTo(ft);
+                format_string(command.SpriteName, USER_STRING_MAX_LENGTH, STR_STRINGID, formatArgs);
             }
         }
-        else if (spriteIdentifier == SPRITE_IDENTIFIER_VEHICLE)
+        else if (litter != nullptr)
         {
-            validSprite = true;
-            auto vehicle = GetEntity<Vehicle>(spriteIndex);
-            if (vehicle != nullptr)
-            {
-                auto ride = vehicle->GetRide();
-                if (ride != nullptr)
-                {
-                    uint8_t formatArgs[32]{};
-                    Formatter ft(formatArgs);
-                    ride->FormatNameTo(ft);
-                    format_string(command.SpriteName, USER_STRING_MAX_LENGTH, STR_STRINGID, formatArgs);
-                }
-            }
-        }
-        else if (spriteIdentifier == SPRITE_IDENTIFIER_LITTER)
-        {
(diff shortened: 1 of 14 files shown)


src/openrct2/world/Balloon.cpp
```diff
@@ -82,19 +82,22 @@ void Balloon::Pop()
...
-        sprite->balloon.colour = colour;
-        sprite->balloon.popped = (isPopped ? 1 : 0);
-    }
+    if (sprite == nullptr)
+        return;
+    sprite->generic.sprite_identifier = SPRITE_IDENTIFIER_MISC;
+    sprite->generic.type = SPRITE_MISC_BALLOON;
+    auto balloon = sprite->generic.As<Balloon>();
+    if (balloon == nullptr)
+        return; // can never happen
```
tupaschoal, line 91: Is this really needed? Did compiler complain?

Because you just set it to be a balloon so it seems kinda of useless.

src/openrct2/world/Duck.cpp
```diff
@@ -305,13 +305,16 @@ void create_duck(const CoordsXY& pos)
...
-    sprite->duck.sprite_width = 9;
-    sprite->duck.sprite_height_negative = 12;
-    sprite->duck.sprite_height_positive = 9;
-    sprite->duck.target_x = targetPos.x;
-    sprite->duck.target_y = targetPos.y;
+    sprite->generic.sprite_identifier = SPRITE_IDENTIFIER_MISC;
+    sprite->generic.type = SPRITE_MISC_DUCK;
+    auto duck = sprite->generic.As<Duck>();
+    if (duck == nullptr)
+        return; // can never happen
```
tupaschoal, line 312: Same as comment above

src/openrct2/world/Sprite.cpp
```diff
@@ -98,28 +98,27 @@ std::string rct_sprite_checksum::ToString() const
...
 {
-    rct_sprite* sprite = nullptr;
-    if (spriteIndex < MAX_SPRITES)
+    if (spriteIndex >= MAX_SPRITES)
     {
-        sprite = &_spriteList[spriteIndex];
+        return nullptr;
     }
-    return sprite;
+    return &_spriteList[spriteIndex].generic;
```
tupaschoal, line 107: why not just
```c++
return spriteIndex < MAX_SPRITES ? &_spriteList[spriteIndex].generic : nullptr
```

src/openrct2/paint/sprite/Paint.Misc.cpp
```diff
@@ -38,6 +38,8 @@ void misc_paint(paint_session* session, const SpriteBase* misc, int32_t imageDir
         case SPRITE_MISC_STEAM_PARTICLE: // 0
         {
             auto particle = misc->As<SteamParticle>();
+            if (particle == nullptr)
+                return;
```
tupaschoal, line 42: Are all of this really needed? Because if the type already matches, `Is<>` will return true and it will have something that isn't a `nullptr` anyway

src/openrct2/scripting/ScMap.hpp
```diff
@@ -183,8 +183,7 @@ namespace OpenRCT2::Scripting
                     return GetObjectAsDukValue(_context, std::make_shared<ScVehicle>(spriteId));
                 case SPRITE_IDENTIFIER_PEEP:
                 {
-                    auto peep = sprite->As<Peep>();
-                    if (peep->AssignedPeepType == PeepType::Staff)
+                    if (sprite->Is<Staff>())
```
tupaschoal: This was added on the first commit, so it should probably be moved there.

src/openrct2-ui/windows/TitleCommandEditor.cpp
```diff
@@ -613,62 +613,55 @@ static void window_title_command_editor_tool_down(
...
-        uint16_t spriteIndex = info.sprite->generic.sprite_index;
-        uint16_t spriteIdentifier = info.sprite->generic.sprite_identifier;
         bool validSprite = false;
-        if (info.sprite->generic.Is<Peep>())
+        auto peep = info.sprite->As<Peep>();
+        auto vehicle = info.sprite->As<Vehicle>();
+        auto litter = info.sprite->As<Litter>();
+        auto duck = info.sprite->As<Duck>();
+        auto balloon = info.sprite->As<Balloon>();
+        if (peep != nullptr)
```
tupaschoal, line 622: Can't this become a `switch()...case` instead?

src/openrct2/paint/sprite/Paint.Misc.cpp
duncanspumpkin: Compilers unfortunately weren't intelligent enough to know that it wouldn't be a nullptr.

src/openrct2-ui/windows/TitleCommandEditor.cpp
duncanspumpkin: I was getting fed up with all the extra nullptr checks involved with doing a switch.

src/openrct2/paint/sprite/Paint.Misc.cpp
tupaschoal: I thought so :\

src/openrct2/world/Balloon.cpp
duncanspumpkin: Again compiler wasn't clever enough to work that out. It will be fixed when i change create sprite so I'm not overly fussed with it atm

src/openrct2-ui/windows/TitleCommandEditor.cpp
ZehMatt: You can also do something like ```if(auto peep = info.sprite->As<Peep>(); peep != nullptr)``` since C++17.

duncanspumpkin: I used that before on another pr and it wasn't liked by a few folk.

PR #12531, opened 2020-07-31
Close #12399: Refactor MIXER_GROUP to strong enum

### LINKED ISSUE #12399: Refactor MIXER_GROUP to strong enum

pizza2004:


src/openrct2/audio/Audio.cpp
```diff
@@ -283,7 +283,7 @@ void audio_start_title_music()
     gTitleMusicChannel = Mixer_Play_Music(pathId, MIXER_LOOP_INFINITE, true);
     if (gTitleMusicChannel != nullptr)
     {
-        Mixer_Channel_SetGroup(gTitleMusicChannel, MIXER_GROUP_TITLE_MUSIC);
+        Mixer_Channel_SetGroup(gTitleMusicChannel, OpenRCT2::Audio::MixerGroup::TitleMusic);
```
AaronVanGeffen, line 286: The file itself should probably be in the `OpenRCT2::Audio` namespace rather than just using it. (Same below.)

pizza2004: Is that something I should tackle in this PR or would it be better to open an issue for it?

PR #12532, opened 2020-07-31
WIP: Use strong enums

autinerd:
This pull request is Work in progress to do some easy "convert to strong enum" tasks.


James103:
As per https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages, you may want to include the issue title in your commit message. Suggested commit message: "Fix #12455: Refactor MOUSE_STATE to use strong enum"

autinerd:
@James103 It will come, when it's ready. In https://github.com/OpenRCT2/OpenRCT2/issues/12460#issuecomment-665227127 it was said that small refactorings could be combined together in one PR. That's why this PR is still a draft.

src/openrct2/rct2/RCT2.h
```diff
@@ -542,7 +542,7 @@ struct RCT2SpriteVehicle : RCT12SpriteBase
     uint16_t var_C8;
     uint16_t var_CA;
     uint8_t scream_sound_id; // 0xCC
-    uint8_t TrackSubposition;
+    VehicleTrackSubposition TrackSubposition;
```
duncanspumpkin, line 545: in general we do not change the type of the RCT2 and RCT1 headers they would stay as `uint8_t` this is because this structure should only be used for reading a writing to the file not for performing any logic and it should be validated.

tupaschoal:
Also @autinerd please stop adding more commits to this PR, because it already touches 53 files and is quite long to review (though each change is split into a commit, thanks for that). Remember to reword the commit titles to be more meaningful and also to squas the ones that fix clang-format into their "Original"

autinerd:
@tupaschoal I will split it up in multiple PRs :)

PR #12537, opened 2020-07-31
Fix #12533: Track designs list does not use natural sorting

### LINKED ISSUE #12533: Feature request: Sort the saved ride designs list in the same way as the saved games list and the staff list.
Currently the saved designs that start with a number are still ordered by the first digit instead of the full number, which causes 10 to be between 1 and 2 (see screenshot). The staff list and saved games list used to be sorted this way too, but they've been changed to consider the whole number instead of just the first digit. I think it'd be a nice change if the saved designs list was organized that way too.

image

tupaschoal:


AaronVanGeffen:
In the game as-is, the list is already ordered in alphanumeric order, just not logical order. To me, this suggests we should replace a sorting algorithm, rather than add another on top.

tupaschoal:
Sure, I can change:
https://github.com/OpenRCT2/OpenRCT2/blob/cd1d573c392c8605849aa3b32ba4e26a95944174/src/openrct2/ride/TrackDesignRepository.cpp#L315-L324

Not sure if it will have other consequences, will have to analyze a bit (I think it won't). @Gymnasiast ?

tupaschoal:
I've changed the original sorting accordingly @AaronVanGeffen , as per your hint

src/openrct2/ride/TrackDesignRepository.cpp
```diff
@@ -319,7 +320,7 @@ class TrackDesignRepository final : public ITrackDesignRepository
             {
                 return a.RideType < b.RideType;
             }
-            return String::Compare(a.Name, b.Name) < 0;
+            return strlogicalcmp(a.Name.c_str(), b.Name.c_str()) < 0;
```
Gymnasiast, line 323: We should have a C++ equivalent for this.

ZehMatt: I agree, std::string does have an actual compare function

tupaschoal: But using std:: string compare would revert the "logical" sorting, wouldn't it? (Not arguing against am std::string overload though)

ZehMatt: I've never heard about a "logical" string sort, what makes it special?

ZehMatt: I wonder if what this is supposed to achive could be:
```
    std::sort(strings.begin(), strings.end(),[](auto& a, auto& b)->bool
    {
        return !std::lexicographical_compare(a.begin(), a.end(),
            b.begin(), b.end(),
            [](auto& c1, auto& c2)
            {
                return std::tolower(c1) < std::tolower(c2);
            });
    });
```

Gymnasiast: @ZehMatt It should be immediately obvious if you take a look at the `strlogicalcmp` function: it sorts `Mechanic 5` before `Mechanic 10` rather than after it, like regular sort would.

ZehMatt: See my comment above.

tupaschoal: I'm all for your solution if it achieves the same effect, I just reused what we already had

ZehMatt: I'm also wondering, is the sort actually working correctly with other languages than lets say en-gb/en-us/de-de etc. Like Korean/Russian/Chinese etc. Another aspect is the UTF8 encoding, using tolower would be the wrong thing here. I never really looked at what it does until now and I'm starting to think that the implementation is incorrect. Also regarding the logical sort, the common term is lexicographical, and we should change that at some point.

AaronVanGeffen: @ZehMatt I integrated ICU for character set conversions￼, but also for Unicode-aware case transformations. We should probably use it more throughout the code base.

PR #12538, opened 2020-07-31
Close #12497: Remove extra SPRITE_INDEX_NULL check

### LINKED ISSUE #12497: Remove extra SPRITE_INDEX_NULL check
There are a number of instances where the following occurs:

```
if (spriteIndex == SPRITE_INDEX_NULL)
    return / continue
auto entity = GetEntity<...>(spriteIndex);
if (entity == nullptr)
    return / continue
```
This is going to end up checking if the `spriteIndex == SPRITE_INDEX_NULL` twice as the GetEntity function also checks this and returns `nullptr`. Therefore it can be simplified to:
```
auto entity = GetEntity<...>(spriteIndex);
if (entity == nullptr)
    return / continue
```
Only perform this refactor for the `GetEntity` or `TryGetEntity` functions. Note: I'm halfway through a GET_VEHICLE to GetEntity refactor so there might be more after that is completed.

Lukasz-Pe:
Every double check I could find has been corrected. If there is something else to be done regarding this issue, please send me a message.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -835,12 +835,11 @@ void Ride::FormatStatusTo(Formatter& ft) const
     {
         ft.Add<rct_string_id>(STR_TEST_RUN);
     }
-    else if (
-        mode == RIDE_MODE_RACE && !(lifecycle_flags & RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING))
+    else if (mode == RIDE_MODE_RACE && !(lifecycle_flags & RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING))
     {
         auto peep = GetEntity<Peep>(race_winner);
-            ft.Add<rct_string_id>(STR_RACE_WON_BY);
-            peep->FormatNameTo(ft);
+        ft.Add<rct_string_id>(STR_RACE_WON_BY);
+        peep->FormatNameTo(ft);
     }
     else if (!ride_type_has_flag(type, RIDE_TYPE_FLAG_IS_SHOP))
     {
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -838,8 +838,16 @@ void Ride::FormatStatusTo(Formatter& ft) const
     else if (mode == RIDE_MODE_RACE && !(lifecycle_flags & RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING))
     {
         auto peep = GetEntity<Peep>(race_winner);
-        ft.Add<rct_string_id>(STR_RACE_WON_BY);
-        peep->FormatNameTo(ft);
+        if (peep != nullptr)
+        {
+            ft.Add<rct_string_id>(STR_RACE_WON_BY);
+            peep->FormatNameTo(ft);
+        }
+        else
+        {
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
Hmm thought there would be more of these. Good work

tupaschoal:
There are conflicts with the base branch, you need to sync your fork and rebase, please follow this: https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2

Lukasz-Pe:
Sorry for closing and reopening.

Lukasz-Pe:
hello @tupaschoal,
I did everything as in the link provided. Now it should be ok.

I would apretiate any help, as this is my first pull request.

tupaschoal:
I've rebased it, dropped the merge commits and etc. Do double check I haven't missed any change

src/openrct2/ride/Ride.cpp
```diff
@@ -835,11 +836,8 @@ void Ride::FormatStatusTo(Formatter& ft) const
     {
         ft.Add<rct_string_id>(STR_TEST_RUN);
     }
-    else if (
-        mode == RIDE_MODE_RACE && !(lifecycle_flags & RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING)
-        && race_winner != SPRITE_INDEX_NULL)
+    else if (mode == RIDE_MODE_RACE && !(lifecycle_flags & RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING) && peep != nullptr)
```
duncanspumpkin, line 839: There is no need for this check as well as the one lower down. I would suggest removing this one and moving the `auto peep = GetEntity<Peep>(race_winner);` into L842 where it was before.

Lukasz-Pe: I've refactored code as you suggested.

Lukasz-Pe:
Thanks @tupaschoal!

I've checke once more and integrated @duncanspumpkin's suggestion.

src/openrct2/ride/Ride.cpp
```diff
@@ -835,12 +835,11 @@ void Ride::FormatStatusTo(Formatter& ft) const
...
         ft.Add<rct_string_id>(STR_TEST_RUN);
     }
-    else if (
-        mode == RIDE_MODE_RACE && !(lifecycle_flags & RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING))
+    else if (mode == RIDE_MODE_RACE && !(lifecycle_flags & RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING))
     {
         auto peep = GetEntity<Peep>(race_winner);
-            ft.Add<rct_string_id>(STR_RACE_WON_BY);
-            peep->FormatNameTo(ft);
+        ft.Add<rct_string_id>(STR_RACE_WON_BY);
```
duncanspumpkin, line 841: Sorry I think I might have not been very clear. I meant you don't need both nullptr checks. So could you readd one of them.

Lukasz-Pe: I've restored the check.
Everything is ok :) I could have asked or read what you have written more carefully.

duncanspumpkin:
I changed the commit message to `Part of` instead of `Close` as there is still a couple more to address. These can be done in another PR.

Issue #12518, opened 2020-07-30
Refactor `get_map_coordinates_from_pos_window` to return an `InteractionInfo`


duncanspumpkin:
`get_map_coordinates_from_pos_window` currently gets an `InteractionInfo` but then splits it up into parts before returning it. It should just return the full `InteractionInfo`. This will have to then modify `viewport_interaction_get_item_right` and `viewport_interaction_get_item_left` to return an `InteractionInfo` the `viewport_interaction_info` struct can be removed eventually.

PR #12541, opened 2020-08-01
Viewport Refactor

duncanspumpkin:
Just cleaning up the functions `get_map_coordinates_from_pos` so that they don't take a ton of in and out parameters.

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -500,6 +500,9 @@ bool viewport_interaction_right_click(const ScreenCoordsXY& screenCoords)
     switch (info.SpriteType)
     {
         case VIEWPORT_INTERACTION_ITEM_NONE:
+        case VIEWPORT_INTERACTION_ITEM_TERRAIN:
+        case VIEWPORT_INTERACTION_ITEM_WATER:
+        case VIEWPORT_INTERACTION_ITEM_LABEL:
             return false;
 
         case VIEWPORT_INTERACTION_ITEM_SPRITE:
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -61,12 +61,12 @@ int32_t viewport_interaction_get_item_left(const ScreenCoordsXY& screenCoords, v
     if ((gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER) && gS6Info.editor_step != EDITOR_STEP_ROLLERCOASTER_DESIGNER)
         return info->type = VIEWPORT_INTERACTION_ITEM_NONE;
 
-    CoordsXY mapCoord = {};
-    get_map_coordinates_from_pos(
-        screenCoords, VIEWPORT_INTERACTION_MASK_SPRITE & VIEWPORT_INTERACTION_MASK_RIDE & VIEWPORT_INTERACTION_MASK_PARK,
-        mapCoord, &info->type, &info->tileElement);
-    info->x = mapCoord.x;
-    info->y = mapCoord.y;
+    auto info2 = get_map_coordinates_from_pos(
+        screenCoords, VIEWPORT_INTERACTION_MASK_SPRITE & VIEWPORT_INTERACTION_MASK_RIDE & VIEWPORT_INTERACTION_MASK_PARK);
+    info->x = info2.Loc.x;
+    info->y = info2.Loc.y;
+    info->tileElement = info2.Element;
+    info->type = info2.SpriteType;
     auto tileElement = info->tileElement;
     auto sprite = info->sprite;
 
@@ -227,7 +227,6 @@ int32_t viewport_interaction_left_click(const ScreenCoordsXY& screenCoords)
  */
 int32_t viewport_interaction_get_item_right(const ScreenCoordsXY& screenCoords, viewport_interaction_info* info)
 {
-    TileElement* tileElement;
     rct_scenery_entry* sceneryEntry;
     Ride* ride;
     int32_t i, stationIndex;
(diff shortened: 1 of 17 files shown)


duncanspumpkin:
Closes #12518

src/openrct2/ride/Ride.cpp
```diff
@@ -6175,17 +6170,19 @@ CoordsXYZD ride_get_entrance_or_exit_position_from_screen_position(const ScreenC
     {
         auto mapX = stationStart.x;
         auto mapY = stationStart.y;
-        entranceMinX = mapX;
-        entranceMinY = mapY;
+        auto entranceMinX = mapX;
+        auto entranceMinY = mapY;
+        auto entranceMaxX = mapX;
+        auto entranceMaxY = mapY;
```
tupaschoal, line 6176: This didn't exist before, I'm assuming you're playing safe or fixing a bug?

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -225,40 +224,35 @@ int32_t viewport_interaction_left_click(const ScreenCoordsXY& screenCoords)
...
-        screenCoords, ~(VIEWPORT_INTERACTION_MASK_TERRAIN & VIEWPORT_INTERACTION_MASK_WATER));
-    info->x = info2.Loc.x;
-    info->y = info2.Loc.y;
-    info->tileElement = info2.Element;
-    info->type = info2.SpriteType;
-    auto tileElement = info->tileElement;
-    auto sprite = info->sprite;
-
-    switch (info->type)
+        return info->SpriteType = VIEWPORT_INTERACTION_ITEM_NONE;
```
tupaschoal, line 239: I know this is how it was, but can you change this into either 
```c++
info->SpriteType == VIEWPORT_INTERACTION_ITEM_NONE
```
(if it was meant to be a comparison)
or
```c++
{
    info->SpriteType = VIEWPORT_INTERACTION_ITEM_NONE;
    return info->SpriteType;
}
```

(Argh, I just noticed it is all over this function)

src/openrct2/interface/Viewport.h
```diff
@@ -49,7 +49,7 @@ enum
     VIEWPORT_FLAG_TRANSPARENT_BACKGROUND = (1 << 19),
 };
 
-enum
+enum ViewportInteractionItem : uint8_t
```
tupaschoal, line 52: Do we want to help https://github.com/OpenRCT2/OpenRCT2/projects/14 and go for enum class now?

src/openrct2-ui/interface/ViewportInteraction.cpp
duncanspumpkin: Yeah i was thinking about that. I think I'm going to rejig it to be like that for a different reason. Change coming up shortly

src/openrct2/interface/Viewport.h
duncanspumpkin: I had a look at how much would be involved for this and decided to leave it for a follow on pr. Now that I've used the enum name it should be easier for anyone else to work out.

src/openrct2/ride/Ride.cpp
duncanspumpkin: It was previously a legacy c function with all the variable declarations at the start of the function.

src/openrct2/interface/Viewport.h
tupaschoal: Sure, sounds fine. I'll keep the comment unresolved so we can remember to open the issue later.

src/openrct2/ride/Ride.cpp
tupaschoal: I meant that `entranceMax*` variables were not initialized to `mapX` and `mapY`

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -391,34 +409,37 @@ int32_t viewport_interaction_get_item_right(const ScreenCoordsXY& screenCoords,
             banner->FormatTextTo(ft, /*addColour*/ true);
             ft.Add<rct_string_id>(STR_MAP_TOOLTIP_STRINGID_CLICK_TO_MODIFY);
             ft.Add<rct_string_id>(sceneryEntry->name);
-            return info->SpriteType;
+            return info;
         }
+        default:
+            break;
```
tupaschoal, line 415: Why did the `default` become neccessary? I'd rather we handled the extra cases explicitly, default might be painful in the future if we add more values to the enum

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -439,44 +460,47 @@ int32_t viewport_interaction_get_item_right(const ScreenCoordsXY& screenCoords,
...
+            return info;
 
         case VIEWPORT_INTERACTION_ITEM_LARGE_SCENERY:
             sceneryEntry = tileElement->AsLargeScenery()->GetEntry();
             ft.Add<rct_string_id>(STR_MAP_TOOLTIP_STRINGID_CLICK_TO_REMOVE);
             ft.Add<rct_string_id>(sceneryEntry->name);
-            return info->SpriteType;
+            return info;
+        default:
+            break;
```
tupaschoal, line 477: Same

src/openrct2-ui/interface/ViewportInteraction.cpp
duncanspumpkin: its a bit of an annoyance but this part of the function is split into two switches with some middle code and if it is not in the first switch then it will be in the second switch.

src/openrct2/ride/Ride.cpp
duncanspumpkin: oh right yeah they were initialised to mapX and mapY later in the function but need for them to exist from earlier so thought might as well do it here like this.

src/openrct2-ui/interface/ViewportInteraction.cpp
tupaschoal: Ok

src/openrct2/ride/Ride.cpp
tupaschoal: Ok

PR #12543, opened 2020-08-01
Fix #12352: Wrong date format for file names

### LINKED ISSUE #12352: Wrong date format for file name
**OS:** Windows 10, Ubuntu 18.04
**Version:** 0.2.6
**Commit/Build:** 205fc33

The year and month in the screenshot's file name is incorrect.
So does for auto savegame files.

 * Year: ``120`` instead of ``2020``
 * Month: ``6`` instead of ``7``

**Screenshots / Video:**
image
The screenshot is taken at ``2020-07-22 17:27:11`` but its file name is ``~-120-06-22-17-22-11.png``

telk5093:
This fixes #12352 
I think the issue has come because of 5ae592e.

I've tested that it works fine in Windows 10 and Ubuntu 18.04. (Not tested in macOS)
But I have no idea why @frutiemax removed the code that adds 1900 to year & 1 to month.
So please let me know if there is any problem or it was on purpose ;)

### DIFF
--- a/src/openrct2/platform/Shared.cpp
+++ b/src/openrct2/platform/Shared.cpp
@@ -75,8 +75,8 @@ namespace Platform
         rct2_date outDate;
         outDate.day = localTime->tm_mday;
         outDate.day_of_week = localTime->tm_wday;
-        outDate.month = localTime->tm_mon;
-        outDate.year = localTime->tm_year;
+        outDate.month = localTime->tm_mon + 1;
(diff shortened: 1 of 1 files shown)


Gymnasiast:
According to the C++ reference, what you did it correct: https://en.cppreference.com/w/cpp/chrono/c/tm

So it's very likely that this should also work on macOS.

PR #12546, opened 2020-08-01
Change maze pathfinding to not prefer a direction

duncanspumpkin:
This is just to mess up MarcelVos

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -61,6 +61,7 @@
 - Fix: 'j' character has broken kerning (original bug).
 - Fix: RCT1 scenarios have more items in the object list than are present in the park or the research list.
 - Fix: Brakes keep working during "Brakes failure".
+- Fix: Guests maze pathfinding prefers a specific direction (original bug).
 - Improved: [#6530] Allow water and land height changes on park borders.
 - Improved: [#11390] Build hash written to screenshot metadata.
 - Improved: [#3205] Make handymen less likely to get stuck in ride queues.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,8 +29,8 @@ set(TITLE_SEQUENCE_SHA1 "304d13a126c15bf2c86ff13b81a2f2cc1856ac8d")
 set(OBJECTS_URL  "https://github.com/OpenRCT2/objects/releases/download/v1.0.15/objects.zip")
 set(OBJECTS_SHA1 "dfd5864cf7d0449c0fb280c5c6b902a24816df6c")
 
-set(REPLAYS_URL  "https://github.com/OpenRCT2/replays/releases/download/v0.0.16/replays.zip")
-set(REPLAYS_SHA1 "8F5FABE403DE8DE7CC3C084C55EF44442E5CBA8E")
+set(REPLAYS_URL  "https://github.com/OpenRCT2/replays/releases/download/v0.0.17/replays.zip")
+set(REPLAYS_SHA1 "8DD6D8DBDB1EA50ABCB3A1FB55D2800E13486F44")
 
 option(FORCE32 "Force 32-bit build. It will add `-m32` to compiler flags.")
 option(WITH_TESTS "Build tests")
--- a/openrct2.proj
+++ b/openrct2.proj
@@ -48,8 +48,8 @@
     <TitleSequencesSha1>304d13a126c15bf2c86ff13b81a2f2cc1856ac8d</TitleSequencesSha1>
     <ObjectsUrl>https://github.com/OpenRCT2/objects/releases/download/v1.0.15/objects.zip</ObjectsUrl>
     <ObjectsSha1>dfd5864cf7d0449c0fb280c5c6b902a24816df6c</ObjectsSha1>
-    <ReplaysUrl>https://github.com/OpenRCT2/replays/releases/download/v0.0.16/replays.zip</ReplaysUrl>
-    <ReplaysSha1>8F5FABE403DE8DE7CC3C084C55EF44442E5CBA8E</ReplaysSha1>
+    <ReplaysUrl>https://github.com/OpenRCT2/replays/releases/download/v0.0.17/replays.zip</ReplaysUrl>
+    <ReplaysSha1>8DD6D8DBDB1EA50ABCB3A1FB55D2800E13486F44</ReplaysSha1>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/openrct2/network/NetworkBase.cpp
+++ b/src/openrct2/network/NetworkBase.cpp
@@ -33,7 +33,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "24"
+#define NETWORK_STREAM_VERSION "25"
 #define NETWORK_STREAM_ID OPENRCT2_VERSION "-" NETWORK_STREAM_VERSION
 
 static Peep* _pickup_peep = nullptr;
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -4895,44 +4895,31 @@ void Guest::UpdateRideMazePathfinding()
     {
         return;
     }
+
     uint16_t mazeEntry = trackElement->GetMazeEntry();
-    uint16_t openHedges = 0;
     // Var37 is 3, 7, 11 or 15
-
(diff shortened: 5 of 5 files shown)


janisozaur:
For reference: https://youtu.be/KVgoy_a_gWI

duncanspumpkin:
Technically in the first tick of the pathfinding the peeps may walk in an unexpected direction. This is because the mazelastdirection value is not set to an invalid number. This is very unlikely to be noticed or to cause issues.

distribution/changelog.txt
```diff
@@ -61,6 +61,7 @@
 - Fix: 'j' character has broken kerning (original bug).
 - Fix: RCT1 scenarios have more items in the object list than are present in the park or the research list.
 - Fix: Brakes keep working during "Brakes failure".
+- Fix: Guests maze pathfinding preferring a single direction (original bug).
```
Gymnasiast, line 64:
```suggestion
- Fix: Guests maze pathfinding prefers a specific direction (original bug).
```

LordMarcel:
Time for me to make an update video :P

syrusakbary:
I think there was a reason the algorithm was that way. For solving mazes, here's the most popular algorithm:
> Touch the wall or hedge with the hand nearest to it, left or right. Keep that same hand touching the wall and keep walking. This may take you on a horribly long route, but it will eventually get you out

Which I think it was reflected properly by the previous algorithm.

Rather than doing complete randomness, I think it would be actually better to assign an strategy for each player (once they enter the attraction), setting their default direction to "always clockwise" (as it was before this PR) or "always counter-clockwise".

That way, you still equilibrate probability weights, but with an algorithm better suited for mazes :)

jamesrowen:
> Which I think it was reflected properly by the previous implementation.

How so? The whole point of the video is that most of the guests failed to find the exit. With your algorithm they would all be guaranteed to proceed along a single path to the exit. Your algorithm is deterministic, and both of the implementations are nondeterministic. The previous implementation wasn't "always clockwise," it was "random, then clockwise if blocked."

The goal is not for all the guests to solve the maze, but to create a bunch of different paths that simulate people walking around semi-randomly.

syrusakbary:
@jamesrowen with the solution I'm proposing the guests will also find the exit for either path.

> How so? The whole point of the video is that most of the guests failed to find the exit. With your algorithm they would all be guaranteed to proceed along a single path to the exit. Your algorithm is deterministic, and both of the implementations are nondeterministic. The previous implementation wasn't "always clockwise," it was "random, then clockwise if blocked."

So, here's the difference: the solution I was proposing will weight more the paths that are in one strategy than the rest, which will tend to a maze-solving strategy.
With current implementation it actually doesn't weight one direction over other, it's just pure randomness which I believe is much worse for solving mazes.

Said in other way, in one strategy (the one I'm proposing) each direction has different weights (towards solving the maze clockwise, or solving the maze counter-clockwise), and in other strategy (implemented by this PR) all directions have the same weight.

Eckankar:
I don't know if there's any interest in this, but inspired by @LordMarcel's update video, I decided to explore a tweak to the algorithm, to allow the guests to wander around randomly, but still tend to make their way towards the exit. Either way, I might as well share my results.

I'm not into the codebase; but made [an implementation in Perl](https://gist.github.com/Eckankar/90a00ae5d77c0382db4869774daa7855) to test it out.

The idea is to slowly disallow tiles based on their distance to the goal, as the guest spends more time in the maze.

More specifically, don't allow walking to a tile `t` if `dist(t) > max_dist - floor(steps / fudge_factor)`, where:
*  `steps` counts the number of steps (from one tile to another) the person has taken in the maze; 
* `dist` gives the distance to the exit from the tile `t` - i.e. number of steps needed for shortest path to exit from `t`. This can be computed once each time the ride is modified, as that is the only time it changes.
* `fudge_factor` allows you to scale on how quickly they're pushed forward. For every `fudge_factor` steps, we close off 1 distance further away, nudging them closer towards the exit.

If there isn't any neighboring tile allowed under that constraint, pick the one with the lowest distance to nudge them forward.

This would guarantee that the guest completes the maze in `max_distance * fudge_factor` steps at most.

Running the above perl code on left-mazes (as in Marcel's video) - with a fudge factor of 5 - gives me the following numbers.  `n` is the length of the maze; and the output numbers are the number of steps used by the guests:

```
Testing on n == 10; 100 trials
old_algorithm
    Total: 110662
  Average: 1106.620
new_algorithm
    Total: 7062
  Average: 70.620
proposed_algorithm
    Total: 3220
  Average: 32.200

Testing on n == 100; 100 trials
new_algorithm
    Total: 665246
  Average: 6652.460
proposed_algorithm
    Total: 47546
  Average: 475.460
```

 I omitted the old algorithm in the larger test for obvious reasons. :)

I don't know if a fudge factor of 5 allows for enough random wandering, but with some adjustment, it could potentially be useful for guaranteeing that the guests complete the maze in a timely fashion.

jamesrowen:
> It's true that both are random, but the distribution for chosen paths vary significantly (since the weights of each direction are different... so one will make the citizens of the sample tend to solve/not solve it)

@syrusakbary In thinking about it more I see it now, in the previous implementation there was bias towards either solving or "anti-solving" the maze depending on the direction. I would still argue that in this scenario you don't necessarily want an algorithm that tends toward pushing them along that one path. In reality most people will exit the same place they came in, though I don't know what the goal is for RCT, maybe they do want to push people along toward the exit. True randomness may be a bit crude but with a little bit of help it's probably a better reflection of all the different ways people go.

michaelmior:
@Eckankar Suppose I have a maze with a split near the beginning and two long parallel paths but only one leads to the exit and the other goes to a dead end. Couldn't this end up with the algorithm getting stuck if not allowed to backtrack far enough?

pizza2004:
There's already PR #10416 open that tries to improve the pathfinding formula, so I wouldn't worry too much about designing another new one.

Ruedii:
Do you have a listing of how this fix works.

How I would approach it:
Current algorithm runs a fixed rotational direction from failed attempts.
Replace that with randomly choosing the rotation clockwise or counterclockwise.

ocalhoun6:
> > Touch the wall or hedge with the hand nearest to it, left or right. Keep that same hand touching the wall and keep walking. This may take you on a horribly long route, but it will eventually get you out

It's possible to design a maze that defeats this algorithm. For a simple example:
Forest Frontiers  120-07-03 21-19-43

syrusakbary:
> In a maze like that, a peep following the 'follow left hand wall' or 'follow right hand wall' would circle endlessly around the outer perimeter, never getting close to the exit in the center.

I believe that precise example will still have people entering the inner circle, since (straight -> straight) is still a possible path (as we are applying randomness to the "always clockwise" or "always counter-clockwise" algorithm) :)

VeaaC:
You could precompute distances to the exist for every space in the maze periodically, and then increase the probability for moving towards the exit vs away from it.

Eckankar:
> @michaelmior 
> Suppose I have a maze with a split near the beginning and two long parallel paths but only one leads to the exit and the other goes to a dead end. Couldn't this end up with the algorithm getting stuck if not allowed to backtrack far enough?

The distances are computed from the exit. So, the path not leading to the exit would have increasing distances, and the one leading towards the exit would have decreasing distances (as seen from the start.)

There is no chance of getting stuck, as falling back to taking a step towards a lower distance will always bring you closer to the exit.

duncanspumpkin:
You need to keep in mind that this code needs to tick all of these boxes: fast (we only have limited time per tick to calculate a direction), deterministic (needs to compute the exact same outcome for network games to not desync), require no additional state (there is no space to stuff additional state into the save format at this time without adding additional special import export code). When we have the new save format we can add additional state which is sort of what the pr which adds a bit of memory to the peeps does.

michaelmior:
@Eckankar I see. I was assuming you meant just how far the guest was from the exit, but if it's the length of the shortest path to the exit, this wouldn't be a problem.

Eckankar:
@michaelmior Well; if we know where the person is in the maze, that's the same thing. :) The good thing about mazes is, that the distance a tile is from the exit only changes whenever the ride is modified. At those times, the distances can be computed with a simple breadth-first search from the exit, and reused whenever a peep walks the maze.

The only state needed per peep is to know how long they've been on the ride for - be it number of steps or some measure of time. I don't know if something relevant to that is already stored on the person - if not, that obviously rules it out for not needing additional information to be stored.

michaelmior:
@Eckankar What I meant is the Euclidean distance to the exist vs the distance *through the maze* to the exist.

Eckankar:
@michaelmior Ah, right - I'm talking maze distance. :)

PR #12548, opened 2020-08-01
closes #12447: Refactor WEATHER_EFFECT to use strong enum

### LINKED ISSUE #12447: Refactor WEATHER_EFFECT to use strong enum

autinerd:
This closes #12447: Refactor WEATHER_EFFECT to use strong enum

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -2650,12 +2650,12 @@ private:
         gClimateUpdateTimer = _s4.climate_timer;
         gClimateCurrent.Temperature = _s4.temperature;
         gClimateCurrent.Weather = _s4.weather;
-        gClimateCurrent.WeatherEffect = WEATHER_EFFECT_NONE;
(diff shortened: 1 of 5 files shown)


src/openrct2/scenario/Scenario.h
```diff
@@ -282,8 +283,8 @@ struct rct_s6_data
     uint8_t next_weather;
     uint8_t temperature;
     uint8_t next_temperature;
-    uint8_t current_weather_effect;
-    uint8_t next_weather_effect;
+    WeatherEffectType current_weather_effect;
```
duncanspumpkin, line 286: Could I be annoying and request that these stay as `uint8_t` this struct should really be moved into the RCT2 header as its a RCT2 specific struct. We should always be validating the values before taking a value from the save and putting it into memory by keeping the types different this will ensure they do not directly convert.

autinerd: Oh, I thought I already changed it back, sorry, I missed this. I will correct it of course!

PR #12551, opened 2020-08-01
closes #12445: Refactor CLIMATE to use strong enum

### LINKED ISSUE #12445: Refacator CLIMATE to use strong enum

autinerd:
This closes #12445: Refactor CLIMATE to use strong enum

### DIFF
--- a/src/openrct2/actions/ClimateSetAction.hpp
+++ b/src/openrct2/actions/ClimateSetAction.hpp
@@ -14,15 +14,13 @@
 
 DEFINE_GAME_ACTION(ClimateSetAction, GAME_COMMAND_SET_CLIMATE, GameActionResult)
 {
-    using climate_t = decltype(gClimate);
-
 private:
-    climate_t _climate;
+    uint8_t _climate;
 
 public:
     ClimateSetAction() = default;
-    ClimateSetAction(climate_t climate)
-        : _climate(climate)
+    ClimateSetAction(ClimateType climate)
+        : _climate(static_cast<uint8_t>(climate))
     {
     }
 
@@ -40,7 +38,7 @@ public:
 
     GameActionResult::Ptr Query() const override
     {
-        if (_climate >= CLIMATE_COUNT)
+        if (_climate >= static_cast<uint8_t>(ClimateType::Count))
         {
             return std::make_unique<GameActionResult>(GA_ERROR::INVALID_PARAMETERS, STR_INVALID_CLIMATE_ID, STR_NONE);
         }
@@ -50,7 +48,7 @@ public:
 
     GameActionResult::Ptr Execute() const override
     {
-        gClimate = _climate;
+        gClimate = ClimateType{ _climate };
 
         gfx_invalidate_screen();
 
--- a/src/openrct2/interface/InteractiveConsole.cpp
+++ b/src/openrct2/interface/InteractiveConsole.cpp
@@ -637,7 +637,8 @@ static int32_t cc_get(InteractiveConsole& console, const arguments_t& argv)
         }
         else if (argv[0] == "climate")
         {
-            console.WriteFormatLine("climate %s  (%d)", ClimateNames[gClimate], gClimate);
+            console.WriteFormatLine(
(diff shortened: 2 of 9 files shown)


src/openrct2/actions/ClimateSetAction.hpp
```diff
@@ -35,12 +33,12 @@ DEFINE_GAME_ACTION(ClimateSetAction, GAME_COMMAND_SET_CLIMATE, GameActionResult)
     {
         GameAction::Serialise(stream);
 
-        stream << DS_TAG(_climate);
+        stream << DS_TAG(reinterpret_cast<uint8_t&>(_climate));
```
duncanspumpkin, line 36: I'd prefer we didn't do this. If you see all of the other game actions. I have purposely kept the type as a `uint8_t` and _then validated the value before using_ (okay maybe I didn't always validate it but that was the plan). So `_climate` would  be a `uint8_t` but `climate` in the constructor would be the specific enum class type.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -865,8 +866,9 @@ static int32_t cc_set(InteractiveConsole& console, const arguments_t& argv)
         {
             if (int_valid[0])
             {
-                const auto newClimate = int_val[0];
-                if (newClimate < 0 || newClimate >= CLIMATE_COUNT)
+                const ClimateType newClimate = ClimateType{ static_cast<uint8_t>(int_val[0]) };
```
duncanspumpkin, line 869: Technically you should validate that the `int_val[0]` is within the range of our `enum class` then if it is construct it into our new type. I.e. move this down to line 877.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -878,17 +880,17 @@ static int32_t cc_set(InteractiveConsole& console, const arguments_t& argv)
             }
             else
             {
-                for (i = 0; i < CLIMATE_COUNT; i++)
+                for (i = 0; i < static_cast<uint8_t>(ClimateType::Count); i++)
```
duncanspumpkin, line 883: use a different variable that is `uint8_t` to reduce the amount of casts.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -878,17 +880,17 @@ static int32_t cc_set(InteractiveConsole& console, const arguments_t& argv)
...
                     {
-                        gClimate = i;
+                        gClimate = ClimateType{ static_cast<uint8_t>(i) };
                         break;
                     }
                 }
             }
 
-            if (i == CLIMATE_COUNT)
+            if (i == static_cast<uint8_t>(ClimateType::Count))
```
duncanspumpkin, line 893: This code is odd. This whole function is wrong. Sorry could you fix it. It should be the following:
uint8_t newClimate = static_cast<uint8_t>(ClimateType::Count);
If `int_valid[0]` then `newClimate =  int_val[0]` else loop through comparing argv[1] and ClimateNames to find a newClimate.

After that if else branch validate the `newClimate` if it is >= ClimateType::Count write error message from L871 and sent invalidArgs to true. Else issue a `ClimateSetAction` with the new climate and finally do the console.Eecute from L894.

duncanspumpkin:
You seem to have merged again instead of rebasing

autinerd:
I wanted to rebase, but there was a problem I think. I will try to fix it.

src/openrct2/actions/ClimateSetAction.hpp
```diff
@@ -14,14 +14,12 @@
...
-    using climate_t = decltype(gClimate);
-
 private:
-    climate_t _climate;
+    uint8_t _climate;
 
 public:
     ClimateSetAction() = default;
-    ClimateSetAction(climate_t climate)
+    ClimateSetAction(uint8_t climate)
```
duncanspumpkin, line 22: Keep the constructor using ClimateType. Then static cast to uint8 for _climate.

tupaschoal:
@duncanspumpkin, can you please check whether this is alright now? (With regards to the changes you requested and etc)

PR #12553, opened 2020-08-01
Resolve GCC warnings for Ride.cpp

CorySanin:
Eliminate warnings from ride_create_vehicles and ride_create_vehicles_find_first_block that would prevent build from succeeding when using GCC.


AaronVanGeffen:
Thanks for the patch. Our CI didn't catch these; what version on GCC are you using?

CorySanin:
gcc 9.3.0 in Ubuntu and Alpine.

PR #12556, opened 2020-08-02
Allow game to run with GUI while without audio.

XplosiveLugnut:
I had hastily built a Gentoo system and was unable to play OpenRCT2 because I hadn't installed anything related to audio. SDL2 wasn't build with either ALSA or PulseAudio. An exception would be thrown immediately because SDL2 couldn't create an audio context.

I doubt this issue pops up very often, but this seems to solve it.

I'd like the sound options to be completely grayed out, but at the moment there's just a single "No Audio Device" device populated in the audio devices drop down box. This gives some indication of why there's no audio, should a user have this happen.

### DIFF
--- a/src/openrct2-ui/Ui.cpp
+++ b/src/openrct2-ui/Ui.cpp
@@ -61,9 +61,9 @@ int main(int argc, const char** argv)
             {
                 audioContext = ToShared(CreateAudioContext());
             }
-            catch (const SDLException&)
+            catch (const SDLException& e)
             {
-                log_warning("Failed to create audio context. Using dummy audio context.");
+                log_warning("Failed to create audio context. Using dummy audio context. Error message was: %s", e.what());
                 audioContext = ToShared(CreateDummyAudioContext());
             }
             auto uiContext = ToShared(CreateUiContext(env));
--- a/src/openrct2-ui/Ui.cpp
+++ b/src/openrct2-ui/Ui.cpp
@@ -9,6 +9,7 @@
 
 #include "Ui.h"
 
+#include "SDLException.h"
 #include "UiContext.h"
 #include "audio/AudioContext.h"
 #include "drawing/BitmapReader.h"
@@ -55,7 +56,16 @@ int main(int argc, const char** argv)
         {
             // Run OpenRCT2 with a UI context
             auto env = ToShared(CreatePlatformEnvironment());
-            auto audioContext = ToShared(CreateAudioContext());
+            std::shared_ptr<IAudioContext> audioContext;
+            try
+            {
+                audioContext = ToShared(CreateAudioContext());
+            }
+            catch (const SDLException&)
+            {
+                log_warning("Failed to create audio context. Using dummy audio context.");
+                audioContext = ToShared(CreateDummyAudioContext());
+            }
             auto uiContext = ToShared(CreateUiContext(env));
             context = CreateContext(env, audioContext, uiContext);
         }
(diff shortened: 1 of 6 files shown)


src/openrct2-ui/Ui.cpp
```diff
@@ -55,7 +56,15 @@ int main(int argc, const char** argv)
         {
             // Run OpenRCT2 with a UI context
             auto env = to_shared(CreatePlatformEnvironment());
-            auto audioContext = to_shared(CreateAudioContext());
+            std::shared_ptr<IAudioContext> audioContext;
+            try
+            {
+                audioContext = to_shared(CreateAudioContext());
+            }
+            catch (SDLException&)
```
janisozaur, line 64: const

src/openrct2-ui/Ui.cpp
```diff
@@ -55,7 +56,15 @@ int main(int argc, const char** argv)
...
+            {
+                audioContext = to_shared(CreateDummyAudioContext());
```
janisozaur, line 66: Failure to create an audio context should be reported to the user as a warning

XplosiveLugnut: So it should be `catch (const SDLException&)` ?

And would the warning use the log_warning macro? Or is this a pop up type warning?

src/openrct2/audio/Audio.cpp
```diff
@@ -169,12 +169,22 @@ void audio_populate_devices()
     devices.insert(devices.begin(), defaultDevice);
 #endif
 
-    gAudioDeviceCount = static_cast<int32_t>(devices.size());
-    gAudioDevices = Memory::AllocateArray<audio_device>(gAudioDeviceCount);
-    for (int32_t i = 0; i < gAudioDeviceCount; i++)
+    if (devices.size())
```
Gymnasiast, line 172: There is a special function to check if the vector is empty. This should already be used throughout our codebase.

src/openrct2/audio/Audio.cpp
```diff
@@ -169,12 +169,22 @@ void audio_populate_devices()
...
+            auto device = &gAudioDevices[i];
+            String::Set(device->name, sizeof(device->name), devices[i].c_str());
+        }
+    }
+    else
+    {
+        gAudioDeviceCount = 1;
+        gAudioDevices = Memory::AllocateArray<audio_device>(gAudioDeviceCount);
+        auto device = &gAudioDevices[0];
+        String::Set(device->name, sizeof(device->name), "No Audio Device");
```
Gymnasiast, line 187: I'm not sure if faking an audio device is a good idea, unless the alternative is dozens of checks elsewhere.

src/openrct2-ui/Ui.cpp
janisozaur: I meant https://en.cppreference.com/w/cpp/error/exception/what

janisozaur: It's always nicer to the user to have the UI error in the UI, but `log_warning` will suffice if it's problematic to have it UI.

janisozaur (changes requested):
The error still needs to be reported to the user

Gymnasiast:
@janisozaur The error is already written to `stderr` using `log_warning`. And you said [here](https://github.com/OpenRCT2/OpenRCT2/pull/12556#discussion_r485216308) that it was acceptable not to have a warning in the UI.

If `log_warning` is not difficult (or if it omits information), I think you'll have to be a bit more specific about what you want to have changed.

janisozaur:
Yes, I've written that already in a comment

XplosiveLugnut:
Experimenting with a pop-up error message on the title screen.

Gymnasiast:
@XplosiveLugnut Good. Note that you will still have to include the error message inside the `SDLException` (that is what janisozaur requested).

tupaschoal:
I fought a bit with SDL, turns out including `SDLException.h` would also bring `SDL.h` which wreaked all sorts of havoc with our main functions. I tried a few `ifdef`s but it turned out to be simpler to just have the exception code split into a source file, and now all builds are working just fine.

This seems ready to merge, considering it had already been approved last time and the only change was solving rebase conflicts and these few builds that were breaking.

janisozaur (approved):

PR #12569, opened 2020-08-02
Fix #12405: Make LIGHTFX_LIGHT_TYPE a strong enum (LightType)

### LINKED ISSUE #12405: Refactor LIGHTFX_LIGHT_TYPE to use strong enum

tombomp:
Made LIGHTFX_LIGHT_TYPE into a strong enum called LightType instead.

I removed a couple of members that are totally unused, tested with effects on at every zoom level and seems fine. Obviously very happy to get criticism if I've done stuff wrong

Also sorry if you saw my last pull request, I closed it and opened a new one because I'm still getting used to git and made a lot of mistakes

### DIFF
--- a/src/openrct2/drawing/LightFX.cpp
+++ b/src/openrct2/drawing/LightFX.cpp
@@ -395,7 +395,7 @@ void lightfx_prepare_light_list()
                 continue;
             }
 
-            SetLightTypeSize(
+            entry->lightType = SetLightTypeSize(
                 entry->lightType, GetLightTypeSize(entry->lightType) - static_cast<int8_t>(_current_view_zoom_front));
         }
     }
--- a/src/openrct2/drawing/LightFX.cpp
+++ b/src/openrct2/drawing/LightFX.cpp
@@ -389,14 +389,14 @@ void lightfx_prepare_light_list()
 
         if (_current_view_zoom_front > 0)
         {
-            if ((static_cast<uint8_t>(entry->lightType) & 0x3) < static_cast<int8_t>(_current_view_zoom_front))
+            if (GetLightTypeSize(entry->lightType) < static_cast<int8_t>(_current_view_zoom_front))
             {
                 entry->lightType = LightType::None;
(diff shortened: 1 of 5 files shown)


src/openrct2/drawing/LightFX.cpp
```diff
@@ -389,13 +389,14 @@ void lightfx_prepare_light_list()
...
-            if ((entry->lightType & 0x3) < static_cast<int8_t>(_current_view_zoom_front))
+            if ((static_cast<uint8_t>(entry->lightType) & 0x3) < static_cast<int8_t>(_current_view_zoom_front))
             {
-                entry->lightType = LIGHTFX_LIGHT_TYPE_NONE;
+                entry->lightType = LightType::None;
                 continue;
             }
 
-            entry->lightType -= static_cast<int8_t>(_current_view_zoom_front);
+            entry->lightType = static_cast<LightType>(
```
duncanspumpkin, line 398: I'm not really a fan of this. Its because we are using the light type as a type and also a size. Could you create a function that can get the light type size and also a function that sets the light type size?
```
constexpr uint8_t GetLightTypeSize(LightType type){ return static_cast<uint8_t>(type) & 0x3; }
constexpr LightType SetLightTypeSize(LightType type, uint8_t size) { return static_cast<LightType>(((static_cast<uint8_t>(type) & ~0x3) | size); }
```

tombomp: Thank you! I was thinking it was a bit ugly that makes a lot of sense. When I add it to the commit should I credit you in the commit message or something like that?

tupaschoal: No need to.

duncanspumpkin:
Did you check that worked? I'm sure it does but I'm not 100% sure. Should be easy to check make sure lights are the same size as before and after.

duncanspumpkin:
Ha no my fault for not testing it either. Good work

PR #12572, opened 2020-08-03
Fix #12561: Synchronized coasters not synced

### LINKED ISSUE #12561: Three coasters are not synchronized correctly, just among two of them
**OS:** Windows 10
**Version:** 0.2.6
**Commit/Build:** 733bc5

I was found this issue in Three Monkeys Park scenario which is well known as that there is nice three synchronized coasters.
These coasters were not synchronized correctly. Just two of them synchronized.

Since it was fine when I tested by four coasters, I think the synchronisation does not work properly if the number of coasters are odd number(3, 5, 7 ...).

**Steps to reproduce:**
1-1. Build three coasters to synchronize
1-2. Run them
2. Or, just open Three Monkeys Park scenario

**Screenshots / Video:**
Incorrect synchronisation
All of three coasters' are synchronized.

tupaschoal:
As stated on issue, regression from https://github.com/tupaschoal/OpenRCT2/commit/e5e70688

I agree with @janisozaur that it's a good candidate for a replay and am willing to do so @duncanspumpkin  (and maybe document for other developers too?)

### DIFF
--- a/src/openrct2/network/NetworkBase.cpp
+++ b/src/openrct2/network/NetworkBase.cpp
@@ -33,7 +33,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "25"
+#define NETWORK_STREAM_VERSION "26"
 #define NETWORK_STREAM_ID OPENRCT2_VERSION "-" NETWORK_STREAM_VERSION
 
 static Peep* _pickup_peep = nullptr;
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
This wasn't in the last release so changelog entry should be removed.

duncanspumpkin:
For the replay create as small a park as possible that exhibits the function. I can record an Replay for it and add it to the replay repo

distribution/changelog.txt
```diff
@@ -58,6 +58,7 @@
 - Fix: [#12498] Circus construction ghost does not rotate (original bug).
 - Fix: [#12505] Stores selling multiple items can only have the first product advertised.
 - Fix: [#12506] Cannot advertise food if there are no rides in the park.
+- Fix: [#12561] Synchronized coasters not departing synced
```
tupaschoal, line 61:
```suggestion
```

tupaschoal:
I've removed the changelog entry. Do squash merge. Should I have incremented the network version?
I'll create the park, but I'd very much like the process of doing the replay and/or do it myself to learn.

duncanspumpkin:
Excellent I've been getting fed up always doing the replays.

duncanspumpkin:
Yes bump the network version. Replays should be added before merging this.

duncanspumpkin:
Do the replays as a separate pr as a few folk are asking about this

PR #12573, opened 2020-08-03
Refactor footpath_is_connected_to_map_edge

### LINKED ISSUE #12334: Remove goto from footpath_is_connected_to_map_edge_recurse
https://github.com/OpenRCT2/OpenRCT2/blob/d10b615eb33f6cd597340d83d68eaaa97c1a493e/src/openrct2/world/Footpath.cpp#L1345-L1349

### LINKED ISSUE #12557: Scenario editor hangs when checking if park entrance path extends to map edge
**Summary:**
The `footpath_is_connected_to_map_edge_recurse` method in Footpath.cpp recurses for several minutes when there is a 2-tile wide path of length 15 and tiles are placed in certain configurations. It eventually terminates (on my machine it took roughly 7 minutes to begin responding again).

**Steps to reproduce:**
1. Launch the Scenario Editor and advance to the Landscape Editor screen.
2. Place a straight path of length 16, starting at the right side of the map extending parallel to the left side.
3. Place a second path of length 15 along the right side of the first path, again extending from the outermost tile of the map inward.
4. Place the park entrance connected to the first path of length 16.
5. Attempt to advance to the Invention List Set Up screen, the editor will hang for several minutes.

There are some other configurations that cause this issue, and it seems that the order in which the tiles are placed has a large effect on whether or not it hangs. As well, it does not hang if the second path is on the left side of the first path.

**Screenshot:**
EditorHang

NathanIkola:
- Fix #12334: remove use of goto in footpath_is_connected_to_map_edge

- Remove recursion in footpath_is_connected_to_map_edge

- Fix #12557: editor does not hang when searching for map edge

### DIFF
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -1287,137 +1287,175 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
  *              (1 << 5): Unown
  *              (1 << 7): Ignore no entry signs
  */
-static int32_t footpath_is_connected_to_map_edge_recurse(
-    const CoordsXYZ& footpathPos, int32_t direction, int32_t flags, int32_t level, int32_t distanceFromJunction,
+static int32_t footpath_is_connected_to_map_edge_helper(
+    CoordsXYZ footpathPos, int32_t direction, int32_t flags, int32_t level, int32_t distanceFromJunction,
     int32_t junctionTolerance)
 {
-    TileElement* tileElement;
+    // Struct for keeping track of tile state
+    struct TileState
+    {
+        //bool processed = false;
+        CoordsXYZ footpathPos;
+        int32_t direction;
+        int32_t level;
+        int32_t distanceFromJunction;
+        int32_t junctionTolerance;
+    };
+
+    // Vector of all of the child tile elements for us to explore
+    std::vector<TileState> tiles;
+    TileElement* tileElement = nullptr;
+    int numPendingTiles = 0;
+
+    // Captures the current state of the variables and stores them for iteration later
+    auto CaptureCurrentTileState = [&tiles, &footpathPos, &direction, &level, &distanceFromJunction, &junctionTolerance,
+                                    &numPendingTiles]() -> void {
+        tiles.push_back({ footpathPos, direction, level, distanceFromJunction, junctionTolerance });
+        ++numPendingTiles;
+    };
+
+    // Loads the next tile to visit into our variables
+    auto LoadNextTileElement = [&tiles, &footpathPos, &direction, &level, &distanceFromJunction, &junctionTolerance,
+                                &numPendingTiles]() -> void {
+        size_t ii = tiles.size();
+        --numPendingTiles;
+        footpathPos = tiles[ii - 1].footpathPos;
+        direction = tiles[ii - 1].direction;
+        level = tiles[ii - 1].level;
+        distanceFromJunction = tiles[ii - 1].distanceFromJunction;
+        junctionTolerance = tiles[ii - 1].junctionTolerance;
+        tiles.pop_back();
+    };
+
+    // Helper method for footpath_is_connected_to_map_edge_helper
+    // to help make the function more readable
+    auto SkipTileElement = [](int32_t ste_flags, TileElement* ste_tileElement, int32_t& ste_slopeDirection,
+                              int32_t ste_direction, const CoordsXYZ& ste_targetPos) {
+        // We are only interested in paths
+        if (ste_tileElement->GetType() != TILE_ELEMENT_TYPE_PATH)
+            return true;
+
+        if (ste_tileElement->AsPath()->IsSloped()
+            && (ste_slopeDirection = ste_tileElement->AsPath()->GetSlopeDirection()) != ste_direction)
+        {
+            if (direction_reverse(ste_slopeDirection) != ste_direction)
+                return true;
+            if (ste_tileElement->GetBaseZ() + PATH_HEIGHT_STEP != ste_targetPos.z)
+                return true;
+        }
+        else if (ste_tileElement->GetBaseZ() != ste_targetPos.z)
+            return true;
+
+        if (!(ste_flags & FOOTPATH_CONNECTED_MAP_EDGE_IGNORE_QUEUES))
+            if (ste_tileElement->AsPath()->IsQueue())
+                return true;
+        return false;
+    };
+
+    // Function to test whether or not the vector has been processed
+    auto DoneProcessing = [&numPendingTiles]() -> bool { return numPendingTiles == 0; };
+
     int32_t edges, slopeDirection;
 
-    auto targetPos = CoordsXYZ{ CoordsXY{ footpathPos } + CoordsDirectionDelta[direction], footpathPos.z };
-    if (++level > 250)
-        return FOOTPATH_SEARCH_TOO_COMPLEX;
+    // Capture the current tile state to begin the loop
+    CaptureCurrentTileState();
 
-    // Check if we are at edge of map
-    if (targetPos.x < COORDS_XY_STEP || targetPos.y < COORDS_XY_STEP)
-        return FOOTPATH_SEARCH_SUCCESS;
-    if (targetPos.x >= gMapSizeUnits || targetPos.y >= gMapSizeUnits)
-        return FOOTPATH_SEARCH_SUCCESS;
-
-    tileElement = map_get_first_element_at(targetPos);
-    if (tileElement == nullptr)
-        return level == 1 ? FOOTPATH_SEARCH_NOT_FOUND : FOOTPATH_SEARCH_INCOMPLETE;
-    do
+    // Loop on this until all tiles are processed or we return
+    while (!DoneProcessing())
     {
-        if (tileElement->GetType() != TILE_ELEMENT_TYPE_PATH)
-            continue;
+        LoadNextTileElement();
 
-        if (tileElement->AsPath()->IsSloped() && (slopeDirection = tileElement->AsPath()->GetSlopeDirection()) != direction)
-        {
-            if (direction_reverse(slopeDirection) != direction)
-                continue;
-            if (tileElement->GetBaseZ() + PATH_HEIGHT_STEP != targetPos.z)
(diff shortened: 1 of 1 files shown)


janisozaur:
@NathanIkola can you add a test for that?

NathanIkola:
I'm not sure where to start for adding a test for this, but I'd love to learn; any advice on where I can start?

NathanIkola:
No problem, I'll split those up into a few commits and get them pushed up.

Would you prefer that I go back entirely and make an intermediate commit of solely removing goto, or would it be appropriate to combine the first two changes into a single commit since that was ultimately still part of the refactor?

src/openrct2/world/Footpath.cpp
```diff
@@ -1348,65 +1348,62 @@ static int32_t footpath_is_connected_to_map_edge_recurse(
                 }
             }
         }
-        goto searchFromFootpath;
-    } while (!(tileElement++)->IsLastForTile());
-    return level == 1 ? FOOTPATH_SEARCH_NOT_FOUND : FOOTPATH_SEARCH_INCOMPLETE;
```
tupaschoal, line 1353: I don't think this change is 100% as it was, walk with me:

**Previously**
- When it reached the `goto`, it jumped to a label just after the function
- It essentially did `break` on the `while` block, but skipping the `return` that is just after it

**Now**
- It does `searchFromFootpath` on all iterations of the `while` 
- It essentially stopped "early exiting"

Does that make sense? If so you'll need to review your first commit

src/openrct2/world/Footpath.cpp
```diff
@@ -1279,7 +1279,7 @@ static int32_t footpath_is_connected_to_map_edge_helper(
     // Struct for keeping track of tile state
     struct TileState
     {
-        //bool processed = false;
+        bool processed = false;
         CoordsXYZ footpathPos;
```
tupaschoal: The description of this third commit states:

1. Fixed a bug that caused the scenario editor to hang when searching for the map edge
2. Also implements vanilla RCT2 behavior when unowning paths by bypassing the early returns when the UNOWN flag is set and letting the method continue to unown all connected path tiles

These are two changes right? So they should go into different commits. It might sound like nitpicking, but imagine that this commits introduce a bug, we won't be able to tell if it was introduced by 1 or 2. It also makes it a little bit harder to review

src/openrct2/world/Footpath.cpp
NathanIkola: I am not extremely comfortable with `goto` so it's possible my understanding is flawed, but I reviewed my first commit, and I believe the behavior should be identical. 

In this commit I simply substituted the code from `searchFromFootpath` to where the `goto` call was. Copying the code from `searchFromFootpath` verbatim to the call site of `goto searchFromFootpath` emulates almost all of the behavior that the `goto` call yields, except for the fact that, as you pointed out, the `goto` call breaks it out of the loop, whereas my implementation does not do so directly.

However, all code paths return a value in `searchFromFootpath`, so the `break` provided by the `goto` call is unneeded, since it is guaranteed to hit a return every time `goto searchFromFootpath` is called. Therefore, it should be safe to put inside of the while loop, since there is no way to hit the line where the `goto` would have been, without hitting a return before looping again.

As a proof of concept, remove the return statement on Footpath.cpp:1408 of the develop branch; you will experience a compilation failure due to not all code paths returning a value since the label is at the end of the function, and if you were able to execute the code within the label without returning, you would reach the end of a non-void function without returning.

Is this correct or have I gone wrong in my reasoning?

NathanIkola: I will also double check my iterative refactor to make sure that it preserves this behavior since I do modify this code slightly.

tupaschoal: > Is this correct or have I gone wrong in my reasoning?

I hadn't looked this way, but it does seem like you're correct :)

tupaschoal:
@duncanspumpkin you might want to check this, specially the fourth commit on vanillla behavior

src/openrct2/world/Footpath.cpp
NathanIkola: I thought it was right, but you definitely had me worried. Glad you said something though, because I was missing the appropriate `break` in my iterative refactor to emulate this behavior, so it very helpful; thank you!

src/openrct2/world/Footpath.cpp
```diff
@@ -1272,148 +1272,210 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
...
             {
-                for (int32_t i = 1; i < 6; i++)
+                // Determine how many elements to check
+                size_t limit = 0;
+                if (tileElement[1].GetType() == TILE_ELEMENT_TYPE_BANNER)
+                    limit = 4;
+                if (tileElement[2].GetType() == TILE_ELEMENT_TYPE_BANNER && tileElement[1].GetType() != TILE_ELEMENT_TYPE_PATH)
+                    limit = 6;
+
+                for (size_t ii(1); ii < limit; ++ii)
```
duncanspumpkin, line 1412: use a better name for the index `ii`.

src/openrct2/world/Footpath.cpp
```diff
@@ -1272,148 +1272,210 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
...
+            direction = direction_reverse(direction);
+            if (!(flags & FOOTPATH_CONNECTED_MAP_EDGE_IGNORE_NO_ENTRY))
             {
-                for (int32_t i = 1; i < 6; i++)
+                // Determine how many elements to check
+                size_t limit = 0;
+                if (tileElement[1].GetType() == TILE_ELEMENT_TYPE_BANNER)
+                    limit = 4;
+                if (tileElement[2].GetType() == TILE_ELEMENT_TYPE_BANNER && tileElement[1].GetType() != TILE_ELEMENT_TYPE_PATH)
+                    limit = 6;
```
duncanspumpkin, line 1410: This part is wrong but its currently wrong as well. The loop below is meant to be one of two options:
if the next element is a banner it will go through elements 1, 2, 3, 4 checking if they are a banner and if they are it will cull the edges. If not a banner stops processing the loop.
The other option is that it loops through elemnts 2, 3, 4, 5 checking the same if they are a banner.... So the limit should be 5 and 6 but also you need to adjust the start value so that its either 1 or 2.

duncanspumpkin: Tbh all of this code is very badly designed because it is possible to get a banner on a tile more than 1 element apart by placeing fences after the path is placed but before the banner is placed. But my above changes will bring it back in alignment with vanilla.

src/openrct2/world/Footpath.cpp
```diff
@@ -1272,148 +1272,210 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
...
-            junctionTolerance--;
-        }
-        junctionTolerance--;
-        if (junctionTolerance < 0)
-        {
-            return FOOTPATH_SEARCH_TOO_COMPLEX;
-        }
+            // Find next direction to go
+            if (!get_next_direction(edges, &direction))
+                continue;
```
duncanspumpkin, line 1428: This continue looks incorrect. You are going to repeat the do while but shouldn't this exit the do while and do another iteration of the while.

src/openrct2/world/Footpath.cpp
```diff
@@ -1272,148 +1272,210 @@ static bool get_next_direction(int32_t edges, int32_t* direction)
  *              (1 << 5): Unown
  *              (1 << 7): Ignore no entry signs
  */
-static int32_t footpath_is_connected_to_map_edge_recurse(
-    const CoordsXYZ& footpathPos, int32_t direction, int32_t flags, int32_t level, int32_t distanceFromJunction,
+static int32_t footpath_is_connected_to_map_edge_helper(
+    CoordsXYZ footpathPos, int32_t direction, int32_t flags, int32_t level, int32_t distanceFromJunction,
```
duncanspumpkin, line 1276: level, distanceFromJunction, junctionTolerance do not need to be parameters any more and can become local function variables.

duncanspumpkin (changes requested):
The general style of this is a bit different to the rest of the codebase. There isn't anything wrong with that but its a little harder for me to follow looking for differences. It looks like the existing code is checking the banners wrong so would appretiate if you could fix that for your version. I'm not 100% sure on one of the continue statements it just looks wrong but I can't quite work out how it is meant to work.

src/openrct2/world/Footpath.cpp
NathanIkola: Because of the issue you posed in the second comment about being able to get a banner on a tile more than 1 element apart, I take it that it would be more appropriate to have the loop `continue` rather than `break` if we find an element that is not a banner so that we are sure to check every element in our range and cull from all banners present?

duncanspumpkin: There should be a max of 4 banners. If there is a path element then it should stop checking. So you could come up with a loop that handles all that.

do
 If path break
 If not banner continue
 Cull edges
While not last element

src/openrct2/world/Footpath.cpp
NathanIkola: I believe you are correct; each separate pass of the do-while loop should correspond to a single recursive call in the original code, so where that return was should most certainly be a `break` rather than a `continue`.

src/openrct2/world/Footpath.cpp
NathanIkola: I have implemented this loop in but I do want to clarify one thing before committing my code: regarding the "not last element" condition of your do-while, was that implying the end of all elements in the tile, or simply the end of the ranges 1-4 or 2-5? Because we can't guarantee the banners are contiguous in the tile, I took that to mean the absolute end so that we are guaranteed to check all banners.

duncanspumpkin: You have implemented it correctly. Thanks makes more sense now

tupaschoal:
@NathanIkola can you mark the issues you fixed as resolved, please? It makes for an easier time for us to tell whether you've tended to the commens.

Thanks!

duncanspumpkin (approved):
could you just rebase this one last time

PR #12575, opened 2020-08-03
Fix launchpad builds

janisozaur:
```
/<<PKGBUILDDIR>>/src/openrct2-ui/interface/ViewportInteraction.cpp: In function ‘InteractionInfo viewport_interaction_get_item_left(const ScreenCoordsXY&)’:
/<<PKGBUILDDIR>>/src/openrct2-ui/interface/ViewportInteraction.cpp:85:29: error: potential null pointer dereference [-Werror=null-dereference]
   85 |             switch (sprite->sprite_identifier)
      |                     ~~~~~~~~^~~~~~~~~~~~~~~~~
```


janisozaur:
I initially did it the way you describe, but figured this looks nicer. Also: the check can't be _within_ the switch

duncanspumpkin:
It's two switches nested. I was suggesting it will be within the first switch. We've changed a nullptr warning to potentially undefined behaviour. We could just get rid of the variable completely it's just to simplify the body.

PR #12589, opened 2020-08-04
Fix #12416: Refactor CURRENCY_AFFIX to use strong enum

### LINKED ISSUE #12416: Refactor CURRENCY_AFFIX to use strong enum

matt-thomson:
Introduce a strong enum, `CurrencyAffix`, to replace the existing `CURRENCY_AFFIX`.

Closes #12416.

### DIFF
--- a/src/openrct2/config/IniWriter.hpp
+++ b/src/openrct2/config/IniWriter.hpp
@@ -31,6 +31,8 @@ interface IIniWriter
 
     template<typename T> void WriteEnum(const std::string& name, T value, const IConfigEnum<T>& configEnum)
     {
+        static_assert(sizeof(T) <= sizeof(int32_t), "Type too large");
+
         std::string key = configEnum.GetName(value);
         if (key.empty())
         {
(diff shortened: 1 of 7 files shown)


src/openrct2/config/IniWriter.hpp
```diff
@@ -34,7 +34,7 @@ interface IIniWriter
         std::string key = configEnum.GetName(value);
         if (key.empty())
         {
-            WriteInt32(name, value);
+            WriteInt32(name, static_cast<int32_t>(value));
```
matt-thomson, line 37: This is the first time we've had a config enum that wasn't backed by `int32_t`.  I *think* that, after we've made all of the config enums strong, this code path becomes unhittable...

ZehMatt: We could add a static_assert for when the underlying type is bigger.

matt-thomson: Added in 788bc15.

PR #12591, opened 2020-08-05
Feature #11817: Show authors field in object selection

### LINKED ISSUE #11817: Expose `authors` field in the Object Selection
JSON objects have a field that contains information about the author(s) of an object. As the collection of JSON objects grows and contains objects from more people than the usual suspects, it would be nice to expose this information in the Object Selection dialog. Currently, some objects (ab)use the name field for this, but this is hardly ideal.

tombomp:
This closes #11817 by adding the authors to the bottom right of the object selection dialogue. It's the last lines, other lines are shifted 1 up. If there's no authors field it just shows an empty space instead (I think basically every object does have one, but it seemed better to be consistent in positioning if not). It also serialises to the object index and bumps the object index version (I think that's enough to get it recreated?) 

Obviously very open to changing how it displays and willing to fix anything or completely change my approach on this.

image

### DIFF
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -1124,8 +1124,9 @@ static void window_editor_object_selection_paint(rct_window* w, rct_drawpixelinf
         }
         ft.Add<rct_string_id>(STR_STRING);
         ft.Add<const char*>(authorsString.c_str());
-        gfx_draw_string_right(
-            dpi, STR_WINDOW_COLOUR_2_STRINGID, gCommonFormatArgs, COLOUR_BLACK, { w->windowPos.x + w->width - 5, screenPos.y });
+        gfx_draw_string_right_clipped(
+            dpi, STR_WINDOW_COLOUR_2_STRINGID, gCommonFormatArgs, COLOUR_BLACK, { w->windowPos.x + w->width - 5, screenPos.y },
+            w->width - w->widgets[WIDX_LIST].right - 4);
     }
 }
 /**
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -1113,7 +1113,7 @@ static void window_editor_object_selection_paint(rct_window* w, rct_drawpixelinf
     // Draw object author (will be blank space if no author in file or a non JSON object)
     {
         auto ft = Formatter::Common();
-        std::string authorsString = "";
+        std::string authorsString;
         for (size_t i = 0; i < listItem->repositoryItem->Authors.size(); i++)
         {
             if (i > 0)
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -1100,14 +1100,34 @@ static void window_editor_object_selection_paint(rct_window* w, rct_drawpixelinf
     screenPos.y += LIST_ROW_HEIGHT;
 
     // Draw object dat name
-    const char* path = path_get_filename(listItem->repositoryItem->Path.c_str());
-    auto ft = Formatter::Common();
-    ft.Add<rct_string_id>(STR_STRING);
-    ft.Add<const char*>(path);
-    gfx_draw_string_right(
-        dpi, STR_WINDOW_COLOUR_2_STRINGID, gCommonFormatArgs, COLOUR_BLACK, { w->windowPos.x + w->width - 5, screenPos.y });
-}
+    {
+        const char* path = path_get_filename(listItem->repositoryItem->Path.c_str());
+        auto ft = Formatter::Common();
+        ft.Add<rct_string_id>(STR_STRING);
+        ft.Add<const char*>(path);
+        gfx_draw_string_right(
+            dpi, STR_WINDOW_COLOUR_2_STRINGID, gCommonFormatArgs, COLOUR_BLACK, { w->windowPos.x + w->width - 5, screenPos.y });
+        screenPos.y += LIST_ROW_HEIGHT;
+    }
 
+    // Draw object author (will be blank space if no author in file or a non JSON object)
+    {
+        auto ft = Formatter::Common();
+        std::string authorsString = "";
+        for (int i = 0; i < listItem->repositoryItem->Authors.size(); i++)
+        {
+            if (i > 0)
+            {
+                authorsString.append(", ");
+            }
+            authorsString.append(listItem->repositoryItem->Authors[i]);
+        }
+        ft.Add<rct_string_id>(STR_STRING);
+        ft.Add<const char*>(authorsString.c_str());
+        gfx_draw_string_right(
+            dpi, STR_WINDOW_COLOUR_2_STRINGID, gCommonFormatArgs, COLOUR_BLACK, { w->windowPos.x + w->width - 5, screenPos.y });
+    }
+}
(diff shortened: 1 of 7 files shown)


src/openrct2/object/Object.cpp
```diff
@@ -116,6 +116,16 @@ void rct_object_entry::SetName(const std::string_view& value)
     std::memcpy(name, value.data(), std::min(sizeof(name), value.size()));
 }
 
+std::vector<std::string> Object::GetAuthors()
```
duncanspumpkin, line 119: Return a const reference to the authors

src/openrct2/object/Object.cpp
```diff
@@ -116,6 +116,16 @@ void rct_object_entry::SetName(const std::string_view& value)
...
+{
+    return _authors;
+}
+
+void Object::SetAuthors(std::vector<std::string> authors)
```
duncanspumpkin, line 124: I would pass an rvalue reference of authors no point having all the extra copies

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -442,6 +442,31 @@ namespace ObjectFactory
...
+                {
+                    if (json_is_array(authors))
+                    {
+                        std::vector<std::string> authorVector;
+                        for (size_t j = 0; j < json_array_size(authors); j++)
+                        {
+                            json_t* tryString = json_array_get(authors, j);
+                            if (json_is_string(tryString))
+                            {
+                                authorVector.push_back(json_string_value(tryString));
```
tupaschoal, line 456:
```suggestion
                                authorVector.emplace_back(json_string_value(tryString));
```

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -442,6 +442,31 @@ namespace ObjectFactory
...
+                    }
+                    else if (json_is_string(authors))
+                    {
+                        result->SetAuthors({ json_string_value(authors) });
+                    }
+                }
+                else
+                {
+                    result->SetAuthors({ "" });
+                }
```
tupaschoal, line 469:
```suggestion
```
No need to set the vector to a vector with 1 empty string

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -128,13 +129,21 @@ class ObjectFileIndex final : public FileIndex<ObjectRepositoryItem>
...
         uint8_t sourceLength = static_cast<uint8_t>(item.Sources.size());
         stream->WriteValue(sourceLength);
         for (auto source : item.Sources)
         {
             stream->WriteValue(source);
         }
 
+        uint8_t authorsLength = static_cast<uint8_t>(item.Authors.size());
+        stream->WriteValue(authorsLength);
+        for (auto author : item.Authors)
```
tupaschoal, line 142:
```suggestion
        for (const auto& author : item.Authors)
```
To prevent unnecessary copy

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -165,13 +174,21 @@ class ObjectFileIndex final : public FileIndex<ObjectRepositoryItem>
...
         {
             auto value = stream->ReadValue<uint8_t>();
             item.Sources.push_back(value);
         }
 
+        auto authorsLength = stream->ReadValue<uint8_t>();
+        for (size_t i = 0; i < authorsLength; i++)
+        {
+            auto author = stream->ReadStdString();
+            item.Authors.push_back(author);
```
tupaschoal, line 189: Try:
```suggestion
            item.Authors.emplace_back(stream->ReadStdString());
```

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -1100,14 +1100,34 @@ static void window_editor_object_selection_paint(rct_window* w, rct_drawpixelinf
...
+        for (size_t i = 0; i < listItem->repositoryItem->Authors.size(); i++)
+        {
+            if (i > 0)
+            {
+                authorsString.append(", ");
+            }
+            authorsString.append(listItem->repositoryItem->Authors[i]);
+        }
+        ft.Add<rct_string_id>(STR_STRING);
+        ft.Add<const char*>(authorsString.c_str());
```
tupaschoal, line 1126: If there are too many authors, this will probably be drawn off-window or burst throught the object list.

@AaronVanGeffen can you suggest something to elide/clip?

src/openrct2/object/Object.cpp
tombomp: Thank you both for the reference advice, hadn't thought about that before

src/openrct2-ui/windows/EditorObjectSelection.cpp
AaronVanGeffen: `gfx_draw_string_right_clipped` takes care of this automatically. Just need to provide the maximum width.

src/openrct2/object/Object.cpp
```diff
@@ -116,12 +116,12 @@ void rct_object_entry::SetName(const std::string_view& value)
     std::memcpy(name, value.data(), std::min(sizeof(name), value.size()));
 }
 
-std::vector<std::string> Object::GetAuthors()
+const std::vector<std::string>& Object::GetAuthors()
```
tupaschoal, line 119:
```suggestion
const std::vector<std::string>& Object::GetAuthors() const
```

Make the method const as well, so compiler knows it won't change data

tombomp:
> Make sure you resolve the issues pointed out if you fixed them

Ah yeah of course, sorry, thank you for taking the time to review and helping me out!

src/openrct2-ui/windows/EditorObjectSelection.cpp
tombomp: > `gfx_draw_string_right_clipped` takes care of this automatically. Just need to provide the maximum width.

Thank you! I'll work out something using that

tombomp:
image
This is what it looks like when the string is too long. Tested at different window sizes too.

src/openrct2/object/Object.cpp
```diff
@@ -116,6 +116,16 @@ void rct_object_entry::SetName(const std::string_view& value)
     std::memcpy(name, value.data(), std::min(sizeof(name), value.size()));
 }
 
+const std::vector<std::string>& Object::GetAuthors() const
+{
+    return _authors;
+}
+
+void Object::SetAuthors(const std::vector<std::string>& authors)
```
duncanspumpkin, line 124: I still think you could make this more efficient using an r value reference. It would avoid one extra duplication of the vector

tombomp: Sorry, I had to read up on rvalues. To be clear does that just mean changing the argument signature to && and changing the set call to this (ie adding std::move)

`result->SetAuthors(std::move(authorVector));`

duncanspumpkin: Yup that would be all that is involved

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -442,6 +442,27 @@ namespace ObjectFactory
                 {
                     throw std::runtime_error("Object has errors");
                 }
+                auto authors = json_object_get(jRoot, "authors");
+                if (authors != nullptr)
+                {
+                    if (json_is_array(authors))
```
ZehMatt, line 448: You can move this this check to up to ```if (authors != nullptr)```, json_is_array will evaluate to false if ```authors == nullptr```, the nullptr test is redundant this way.

PR #12601, opened 2020-08-06
Close #12497 (Remove extra SPRITE_INDEX_NULL check)

### LINKED ISSUE #12497: Remove extra SPRITE_INDEX_NULL check

Lukasz-Pe:
I hope, that this time there are only my commits. I have been doing all updates accordingly to [this](https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2) document.

If there are any leftovers I will correct them, but I will need also help like e.g. file to check.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -1008,13 +1008,18 @@ static void ride_remove_cable_lift(Ride* ride)
     if (ride->lifecycle_flags & RIDE_LIFECYCLE_CABLE_LIFT)
     {
         ride->lifecycle_flags &= ~RIDE_LIFECYCLE_CABLE_LIFT;
-
-        for (Vehicle* vehicle = GetEntity<Vehicle>(ride->cable_lift); vehicle != nullptr;
-             vehicle = GetEntity<Vehicle>(vehicle->next_vehicle_on_train))
+        uint16_t spriteIndex = ride->cable_lift;
+        do
         {
+            Vehicle* vehicle = GetEntity<Vehicle>(spriteIndex);
+            if (vehicle == nullptr)
+            {
+                return;
+            }
             vehicle->Invalidate();
             sprite_remove(vehicle);
-        }
+            spriteIndex = vehicle->next_vehicle_on_train;
+        } while (spriteIndex != SPRITE_INDEX_NULL);
     }
 }
 
@@ -1031,12 +1036,19 @@ static void ride_remove_vehicles(Ride* ride)
 
         for (size_t i = 0; i <= MAX_VEHICLES_PER_RIDE; i++)
         {
-            for (Vehicle* vehicle = GetEntity<Vehicle>(ride->vehicles[i]); vehicle != nullptr;
-                 vehicle = GetEntity<Vehicle>(vehicle->next_vehicle_on_train))
+            uint16_t spriteIndex = ride->vehicles[i];
+            while (spriteIndex != SPRITE_INDEX_NULL)
             {
+                Vehicle* vehicle = GetEntity<Vehicle>(spriteIndex);
+                if (vehicle == nullptr)
+                {
+                    break;
+                }
                 vehicle->Invalidate();
                 sprite_remove(vehicle);
+                spriteIndex = vehicle->next_vehicle_on_train;
             }
+
             ride->vehicles[i] = SPRITE_INDEX_NULL;
         }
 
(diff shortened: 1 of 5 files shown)


James103:
Could you squish all of the "refactor: removing double check of SPRITE_INDEX_NULL" commits into one?

Lukasz-Pe:
I could do it, but I do not have idea how. Is there any tutorial which I could use?

On the other side, every commit is for separate file. I thought it would be good to commit them separately.

ZehMatt:
You have unrelated commits, you need to rebase the branch onto the latest develop.

Lukasz-Pe:
@ZehMatt, now should be fine.

src/openrct2/ride/Ride.cpp
```diff
@@ -1008,18 +1008,13 @@ static void ride_remove_cable_lift(Ride* ride)
     if (ride->lifecycle_flags & RIDE_LIFECYCLE_CABLE_LIFT)
     {
         ride->lifecycle_flags &= ~RIDE_LIFECYCLE_CABLE_LIFT;
-        uint16_t spriteIndex = ride->cable_lift;
-        do
+
+        for (Vehicle* vehicle = GetEntity<Vehicle>(ride->cable_lift); vehicle != nullptr;
+             vehicle = GetEntity<Vehicle>(vehicle->next_vehicle_on_train))
```
duncanspumpkin, line 1013: unfortunately we can't do this due to the `sprite_remove`. Its a bit of an annoying situation that will be fixed but you will need to revert this change for now. Basically any loop that contains a `sprite_remove` has to be handled very carefully as it can invalidate the iterators.

src/openrct2/ride/Ride.cpp
```diff
@@ -1036,19 +1031,12 @@ static void ride_remove_vehicles(Ride* ride)
 
         for (size_t i = 0; i <= MAX_VEHICLES_PER_RIDE; i++)
         {
-            uint16_t spriteIndex = ride->vehicles[i];
-            while (spriteIndex != SPRITE_INDEX_NULL)
+            for (Vehicle* vehicle = GetEntity<Vehicle>(ride->vehicles[i]); vehicle != nullptr;
+                 vehicle = GetEntity<Vehicle>(vehicle->next_vehicle_on_train))
```
duncanspumpkin, line 1035: Same for this loop as well unfortunately.

src/openrct2/ride/Ride.cpp
```diff
@@ -6462,70 +6431,70 @@ uint64_t ride_entry_get_supported_track_pieces(const rct_ride_entry* rideEntry)
...
+            VEHICLE_SPRITE_FLAG_FLAT,                                                                                                              // TRACK_WHIRLPOOL
+            VEHICLE_SPRITE_FLAG_FLAT | VEHICLE_SPRITE_FLAG_GENTLE_SLOPES | VEHICLE_SPRITE_FLAG_STEEP_SLOPES,                                       // TRACK_BRAKE_FOR_DROP
+            VEHICLE_SPRITE_FLAG_FLAT | VEHICLE_SPRITE_FLAG_FLAT_TO_GENTLE_SLOPE_BANKED_TRANSITIONS |  VEHICLE_SPRITE_FLAG_CORKSCREWS,              // TRACK_CORKSCREW_UNINVERTED
+            VEHICLE_SPRITE_FLAG_FLAT | VEHICLE_SPRITE_FLAG_FLAT_TO_GENTLE_SLOPE_BANKED_TRANSITIONS |  VEHICLE_SPRITE_FLAG_CORKSCREWS,              // TRACK_CORKSCREW_INVERTED
+            VEHICLE_SPRITE_FLAG_FLAT | VEHICLE_SPRITE_FLAG_GENTLE_SLOPES,                                                                          // TRACK_HEARTLINE_TRANSFER
+            0,                                                                                                                                     // TRACK_MINI_GOLF_HOLE
+        };
```
duncanspumpkin, line 6497: unrelated change. Could you revert.

Lukasz-Pe:
@duncanspumpkin 
I've removed those changes and thank for compliment :)

src/openrct2/ride/Ride.cpp
```diff
@@ -6511,10 +6492,10 @@ uint64_t ride_entry_get_supported_track_pieces(const rct_ride_entry* rideEntry)
...
-        VEHICLE_SPRITE_FLAG_VERTICAL_SLOPES,                                                                                                   // TRACK_QUARTER_LOOP_UNINVERTED   
-        VEHICLE_SPRITE_FLAG_VERTICAL_SLOPES,                                                                                                   // TRACK_QUARTER_LOOP_INVERTED     
+        VEHICLE_SPRITE_FLAG_FLAT | VEHICLE_SPRITE_FLAG_FLAT_BANKED |  VEHICLE_SPRITE_FLAG_INLINE_TWISTS,                                       // TRACK_INLINE_TWIST_UNINVERTED
+        VEHICLE_SPRITE_FLAG_FLAT | VEHICLE_SPRITE_FLAG_FLAT_BANKED |  VEHICLE_SPRITE_FLAG_INLINE_TWISTS,                                       // TRACK_INLINE_TWIST_INVERTED
+        VEHICLE_SPRITE_FLAG_VERTICAL_SLOPES,                                                                                                   // TRACK_QUARTER_LOOP_UNINVERTED
+        VEHICLE_SPRITE_FLAG_VERTICAL_SLOPES,                                                                                                   // TRACK_QUARTER_LOOP_INVERTED
```
tupaschoal, line 6498: Typically we try not to touch things that aren't related to what we're changing, even if it's beneficial, but this is minor so I'll let it slip :P

tupaschoal:
Also, don't forget to click "Resolve Conversation" when you push a new commit fixing issues pointed out by the reviewers.

Lukasz-Pe:
@tupaschoal 
Thanks for your insight :)
I'll remember to resolve conversations and be more strict in applying code autoformater.

PR #12605, opened 2020-08-07
Close #12439: Refactor VEHICLE_TYPE to use strong enum

### LINKED ISSUE #12439: Refactor VEHICLE_TYPE to use strong enum

ion232:
Closes #12439 - 'Refactor VEHICLE_TYPE to use strong enum'

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -4359,7 +4359,7 @@ static Vehicle* vehicle_create_car(
     vehicle->ride_subtype = ride->subtype;
 
     vehicle->vehicle_type = vehicleEntryIndex;
-    vehicle->type = carIndex == 0 ? VEHICLE_TYPE_HEAD : VEHICLE_TYPE_TAIL;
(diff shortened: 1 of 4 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -4359,7 +4359,7 @@ static Vehicle* vehicle_create_car(
     vehicle->ride_subtype = ride->subtype;
 
     vehicle->vehicle_type = vehicleEntryIndex;
-    vehicle->type = carIndex == 0 ? VEHICLE_TYPE_HEAD : VEHICLE_TYPE_TAIL;
+    vehicle->type = static_cast<uint8_t>(carIndex == 0 ? VehicleType::Head : VehicleType::Tail);
```
duncanspumpkin, line 4362: You need to change type of vehicle->type to the new type

ion232: Unless I'm mistaken, vehicle->type in Vehicle is inherited from SpriteBase, which has member 'type'. I didn't change this underlying type because it appeared to be used in other non-vehicle related instances.

duncanspumpkin: Argh you are correct. I'll fix it later when i remove this member variable.

Issue #12400, opened 2020-07-25
Refactor TEMPERATURE_FORMAT to use strong enum


ZehMatt:

sudo-michael:
Hi, I plan on working on this if no one else has already.

tupaschoal:
Go ahead

PR #12610, opened 2020-08-08
Refactor TEMPERATURE_FORMAT to use strong enum

sudo-michael:
Closes [#12400](https://github.com/OpenRCT2/OpenRCT2/issues/12400)  Refactor TEMPERATURE_FORMAT to use strong enum

### DIFF
--- a/src/openrct2/config/Config.h
+++ b/src/openrct2/config/Config.h
@@ -15,7 +15,7 @@
 
 #include <string>
 
-enum class TemperatureFormat : int32_t;
+enum class TemperatureUnit : int32_t;
 
 struct GeneralConfiguration
 {
@@ -56,7 +56,7 @@ struct GeneralConfiguration
     // Localisation
     int32_t language;
     int32_t measurement_format;
-    TemperatureFormat temperature_format;
+    TemperatureUnit temperature_format;
     bool show_height_as_units;
     int32_t date_format;
     int32_t currency_format;
@@ -208,10 +208,10 @@ enum SORT
     SORT_DATE_DESCENDING,
 };
 
-enum class TemperatureFormat : int32_t
+enum class TemperatureUnit : int32_t
 {
-    C,
-    F
+    Celsius,
+    Fahrenheit
 };
 
 enum SCALE_QUALITY
--- a/src/openrct2/config/Config.h
+++ b/src/openrct2/config/Config.h
@@ -15,7 +15,7 @@
 
 #include <string>
 
-enum class TEMPERATURE_FORMAT : int32_t;
+enum class TemperatureFormat : int32_t;
 
 struct GeneralConfiguration
 {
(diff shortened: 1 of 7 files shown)


src/openrct2/config/Config.h
```diff
@@ -206,10 +208,10 @@ enum SORT
     SORT_DATE_DESCENDING,
 };
 
-enum TEMPERATURE_FORMAT
+enum class TEMPERATURE_FORMAT : int32_t
```
duncanspumpkin, line 211: Use CamelCase for enum classes.

src/openrct2/config/Config.h
```diff
@@ -15,6 +15,8 @@
 
 #include <string>
 
+enum class TemperatureFormat : int32_t;
```
tupaschoal, line 18: Seems a bit overkill to have the underlying type as `int32_t` (even though that's what it was previously
```suggestion
enum class TemperatureFormat : bool;
```

I'd go for `bool`

duncanspumpkin: I think it was 32 due to it being a config option. There might be issues with changing the size.

tupaschoal: Compilation issues or running/replay issues? I'm alright if you want to keep it untouched for now

ZehMatt: Why would you pick a bool? That type has a different purpose and is not used for integers.

src/openrct2/config/Config.h
```diff
@@ -206,10 +208,10 @@ enum SORT
     SORT_DATE_DESCENDING,
 };
 
-enum TEMPERATURE_FORMAT
+enum class TemperatureFormat : int32_t
```
ZehMatt, line 211:
```suggestion
enum class TemperatureUnit : int32_t
```

src/openrct2/config/Config.h
```diff
@@ -206,10 +208,10 @@ enum SORT
...
 {
-    TEMPERATURE_FORMAT_C,
-    TEMPERATURE_FORMAT_F
+    C,
```
ZehMatt, line 213:
```suggestion
    Celcius,
```

src/openrct2/config/Config.h
```diff
@@ -206,10 +208,10 @@ enum SORT
...
+    F
```
ZehMatt, line 214:
```suggestion
    Fahrenheit
```

src/openrct2/config/Config.h
tupaschoal: Would fit just as well, unless we're planning to go for Kelvin eventually :¶

Not too attached, go for the the uint8 suggestion if it works

src/openrct2/config/Config.h
tupaschoal: Celsius, no?

sudo-michael: Yup that's the correct spelling

ZehMatt: Yup my bad.

src/openrct2/config/Config.h
ZehMatt: bool is just not logical for enums. I think int32_t will do fine.

PR #12623, opened 2020-08-10
Add releases in appdata XML file

takluyver:
This will (hopefully) allow the appdata file to be validated in the Flatpak build.

Ideally this would be updated with each new release. It's also possible to put a whole load of additional metadata with each release - details of changes, URLs, update urgency, etc. - but I believe that's all optional. This is used by software centre type applications (like Gnome software) to present applications and changes in updates. Detailed documentation is at: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases

### DIFF
--- a/distribution/linux/openrct2.appdata.xml
+++ b/distribution/linux/openrct2.appdata.xml
@@ -64,7 +64,26 @@
     <content_attribute id="money-gambling">none</content_attribute>
   </content_rating>
   <releases>
-      <release version="0.2.6" date="2020-04-17" />
-      <release version="0.2.5" date="2020-03-24" />
+      <release version="0.2.6" date="2020-04-17">
+        <url>https://github.com/OpenRCT2/OpenRCT2/releases/tag/v0.2.6</url>
+      </release>
+      <release version="0.2.5" date="2020-03-24">
+        <url>https://github.com/OpenRCT2/OpenRCT2/releases/tag/v0.2.5</url>
+      </release>
+      <release version="0.2.4" date="2019-10-28">
+        <url>https://github.com/OpenRCT2/OpenRCT2/releases/tag/v0.2.4</url>
+      </release>
+      <release version="0.2.3" date="2019-07-10">
+        <url>https://github.com/OpenRCT2/OpenRCT2/releases/tag/v0.2.3</url>
+      </release>
+      <release version="0.2.2" date="2019-03-13">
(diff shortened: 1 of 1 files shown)


distribution/linux/openrct2.appdata.xml
```diff
@@ -63,4 +63,8 @@
     <content_attribute id="money-purchasing">none</content_attribute>
     <content_attribute id="money-gambling">none</content_attribute>
   </content_rating>
+  <releases>
+      <release version="0.2.6" date="2020-04-17" />
+      <release version="0.2.5" date="2020-03-24" />
```
Gymnasiast, line 68: I don't know how important a complete release history is, but the dates are listed here: https://github.com/OpenRCT2/OpenRCT2/releases

distribution/linux/openrct2.appdata.xml
```diff
@@ -63,4 +63,8 @@
     <content_attribute id="money-purchasing">none</content_attribute>
     <content_attribute id="money-gambling">none</content_attribute>
   </content_rating>
+  <releases>
+      <release version="0.2.6" date="2020-04-17" />
```
Gymnasiast, line 67: So if I read the documentation correctly, to add the release notes, we could add a tag inside that reads as follows?
```<url>https://github.com/OpenRCT2/OpenRCT2/releases/tag/v0.2.6</url>```

distribution/linux/openrct2.appdata.xml
takluyver: I think it's used to describe updates from one version with the information to the next, so I don't think it's super important to have the whole back catalogue. But my knowledge of this is rather vague - if one of the Flathub maintainers tells me it needs the full list, I'd accept that (with some grumbling).

distribution/linux/openrct2.appdata.xml
takluyver: That's how I read it too. Would you like me to add that?

Gymnasiast: Yes, please. And also for 0.2.5. 

You can leave out the other versions if you prefer.

takluyver: OK, I've gone back as far as 0.2.0, which seemed like a more coherent starting point.

takluyver:
Thanks!

PR #12630, opened 2020-08-10
Fix #12611: Bad check for pay-per-ride ability

### LINKED ISSUE #12611: 'Monthly Income from ride tickets' in Scenario Editor is removed when park is not free entry.
I ran into an interesting issue while creating a scenario where if you set the park to have a fee, then move onto the next section where you choose an objective, the "Achieve a monthly income from tickets of at least x" option is absent. If you set the park to free entry then it reappears.

More interestingly though, if you set park to free, then set that objective, you can then go back and set the park to have a non-zero fee but the aforementioned objective will stay selected when you go forward again to that menu.

Not sure if this is an oversight or intentional behavior, but it feels unnecessarily limiting either way and it could be discussed whether it should be changed or not.

Gymnasiast:
This also refactors a few other cases where it did not check this properly.

### DIFF
--- a/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
+++ b/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
@@ -461,7 +461,7 @@ static void window_editor_objective_options_show_objective_dropdown(rct_window* 
         gDropdownItemsArgs[numItems] = STR_OBJECTIVE_DROPDOWN_PARK_VALUE_AT_A_GIVEN_DATE;
         numItems++;
 
-        if (parkFlags & PARK_FLAGS_PARK_FREE_ENTRY)
+        if (park_ride_prices_unlocked())
         {
             gDropdownItemsFormat[numItems] = STR_DROPDOWN_MENU_LABEL;
             gDropdownItemsArgs[numItems] = STR_OBJECTIVE_DROPDOWN_MONTHLY_INCOME_FROM_RIDE_TICKETS;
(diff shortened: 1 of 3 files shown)


src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -753,7 +753,7 @@ static void window_editor_objective_options_main_update(rct_window* w)
          && objectiveType != OBJECTIVE_GUESTS_AND_RATING && objectiveType != OBJECTIVE_10_ROLLERCOASTERS_LENGTH
          && objectiveType != OBJECTIVE_FINISH_5_ROLLERCOASTERS)
         // The park must be free for the monthly ride income objective
```
Gymnasiast, line 755: Remove this comment.

src/openrct2/management/Award.cpp
```diff
@@ -171,7 +171,7 @@ static bool award_is_deserved_best_value(int32_t activeAwardTypes)
     if (activeAwardTypes & (1 << PARK_AWARD_MOST_DISAPPOINTING))
         return false;
 
-    if (gParkFlags & (PARK_FLAGS_NO_MONEY | PARK_FLAGS_PARK_FREE_ENTRY))
+    if (gParkFlags & (PARK_FLAGS_NO_MONEY) || !park_entry_price_unlocked())
```
ZehMatt, line 174:
```suggestion
    if ((gParkFlags & PARK_FLAGS_NO_MONEY) || !park_entry_price_unlocked())
```
Is probably a bit better

Gymnasiast: Yeah, I agree.

Issue #12474, opened 2020-07-25
Harden the network code


ZehMatt:

Gymnasiast:
Could you add specific criteria/tasks? We need to be able to tell if this has been finished or not.

PR #12631, opened 2020-08-10
Add more guards in network code

ZehMatt:
This is a part of #12474, this could help for when new reports come in.

### DIFF
--- a/src/openrct2/network/NetworkBase.cpp
+++ b/src/openrct2/network/NetworkBase.cpp
@@ -3329,26 +3329,36 @@ int32_t network_get_num_players()
 
 const char* network_get_player_name(uint32_t index)
 {
+    Guard::IndexInRange(index, gNetwork.player_list);
+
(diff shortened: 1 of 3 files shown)


src/openrct2/network/NetworkBase.cpp
```diff
@@ -3329,26 +3329,36 @@ int32_t network_get_num_players()
 
 const char* network_get_player_name(uint32_t index)
 {
+    Guard::Assert(index >= gNetwork.player_list.size(), "Index %u out of bounds (%zu)", index, gNetwork.player_list.size());
```
tupaschoal, line 3332: Does `Guard::Assert` print the calling fn? Or are we expecting to get this from the stack trace?

All of these calls are equal, I'd maybe call this a `checkValidIndex` which would make changing these messages a lot easier in the future.

ZehMatt: I guess we can add a Guard::CheckBounds(idx, container) or something, the checks are around players and groups.

PR #12633, opened 2020-08-10
Refactor objective handling in Scenario Editor

Gymnasiast:
This moves some knowledge out of the window and into the main code and also cleans up the window a bit.

This also allows setting the "x guests by the end of year y" goal for no money scenarios, which was not the case previously.

### DIFF
--- a/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
+++ b/src/openrct2-ui/windows/EditorObjectiveOptions.cpp
@@ -439,51 +439,19 @@ static void window_editor_objective_options_show_objective_dropdown(rct_window* 
     dropdownWidget = &w->widgets[WIDX_OBJECTIVE];
     parkFlags = gParkFlags;
 
-    gDropdownItemsFormat[numItems] = STR_DROPDOWN_MENU_LABEL;
-    gDropdownItemsArgs[numItems] = STR_OBJECTIVE_DROPDOWN_HAVE_FUN;
-    numItems++;
-
-    if (!(parkFlags & PARK_FLAGS_NO_MONEY_SCENARIO))
+    for (auto i = 0; i < OBJECTIVE_COUNT; i++)
     {
-        gDropdownItemsFormat[numItems] = STR_DROPDOWN_MENU_LABEL;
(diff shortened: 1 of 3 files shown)


src/openrct2/scenario/Scenario.h
```diff
@@ -344,7 +344,9 @@ enum
     OBJECTIVE_10_ROLLERCOASTERS_LENGTH,
     OBJECTIVE_FINISH_5_ROLLERCOASTERS,
     OBJECTIVE_REPLAY_LOAN_AND_PARK_VALUE,
-    OBJECTIVE_MONTHLY_FOOD_INCOME
+    OBJECTIVE_MONTHLY_FOOD_INCOME,
+
+    OBJECTIVE_COUNT
 };
```
duncanspumpkin, line 350: Would it be too annoying to do the enum class refactor of the objectives in this PR?

Gymnasiast: I prefer doing that in a separate PR to keep things simple.

duncanspumpkin: Okay

src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -439,52 +439,20 @@ static void window_editor_objective_options_show_objective_dropdown(rct_window*
...
-
-        gDropdownItemsFormat[numItems] = STR_DROPDOWN_MENU_LABEL;
-        gDropdownItemsArgs[numItems] = STR_OBJECTIVE_DROPDOWN_PARK_VALUE_AT_A_GIVEN_DATE;
-        numItems++;
+        if (i == OBJECTIVE_NONE || i == OBJECTIVE_BUILD_THE_BEST)
+            continue;
 
-        if (park_ride_prices_unlocked())
+        const bool objectiveAllowedByMoneyUsage = !(parkFlags & PARK_FLAGS_NO_MONEY_SCENARIO) || !ObjectiveNeedsMoney(i);
+        if (objectiveAllowedByMoneyUsage && (i != OBJECTIVE_MONTHLY_RIDE_INCOME || park_ride_prices_unlocked()))
```
duncanspumpkin, line 448: This second part of the if is confusing. Can you break it out into a separate named variable to make it crystal clear why it's like this

PR #12636, opened 2020-08-10
Use a 16-bit integer for track type in more places 

Gymnasiast:


src/openrct2/ride/Track.cpp
```diff
@@ -1195,13 +1195,13 @@ bool track_element_is_covered(int32_t trackElementType)
     }
 }
 
-bool track_element_is_booster(uint8_t rideType, uint8_t trackType)
+bool track_element_is_booster(uint8_t rideType, track_type_t trackType)
```
duncanspumpkin, line 1198: These aren't track elements the function name should be TrackTypeIsBooster

src/openrct2/ride/Track.cpp
```diff
@@ -1195,13 +1195,13 @@ bool track_element_is_covered(int32_t trackElementType)
...
 {
     // Boosters share their ID with the Spinning Control track.
     return rideType != RIDE_TYPE_SPINNING_WILD_MOUSE && trackType == TRACK_ELEM_BOOSTER;
 }
 
-bool track_element_has_speed_setting(uint8_t trackType)
+bool track_element_has_speed_setting(track_type_t trackType)
```
duncanspumpkin, line 1204: TrackTypeHasSpeedSetting

PR #12638, opened 2020-08-11
Fix Z coordinate and spillover in guest debug tab

quale:
It used to show X Y X instead of X Y Z. The small minimum size (albeit barely smaller than the maximum) caused the pathfind history to spill out.


quale:
Done!

PR #12642, opened 2020-08-11
Fix #12402: Refactor MEASUREMENT_FORMAT to use strong enum

### LINKED ISSUE #12402: Refactor MEASUREMENT_FORMAT to use strong enum

matt-thomson:
Introduce a strong enum, MeasurementFormat, to replace the existing MEASUREMENT_FORMAT.

Closes #12402.

### DIFF
--- a/src/openrct2-ui/windows/Options.cpp
+++ b/src/openrct2-ui/windows/Options.cpp
@@ -1740,10 +1740,9 @@ static void window_options_invalidate(rct_window* w)
 
             // Distance: metric / imperial / si
             {
-                rct_string_id stringId;
+                rct_string_id stringId = STR_NONE;
                 switch (gConfigGeneral.measurement_format)
                 {
-                    default:
                     case MeasurementFormat::Imperial:
                         stringId = STR_IMPERIAL;
                         break;
--- a/src/openrct2-ui/windows/Options.cpp
+++ b/src/openrct2-ui/windows/Options.cpp
@@ -1143,7 +1143,7 @@ static void window_options_mousedown(rct_window* w, rct_widgetindex widgetIndex,
 
                     window_options_show_dropdown(w, widget, 3);
 
-                    dropdown_set_checked(gConfigGeneral.measurement_format, true);
+                    dropdown_set_checked(static_cast<int32_t>(gConfigGeneral.measurement_format), true);
                     break;
                 case WIDX_TEMPERATURE_DROPDOWN:
                     gDropdownItemsFormat[0] = STR_DROPDOWN_MENU_LABEL;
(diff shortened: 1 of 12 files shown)


src/openrct2/localisation/Localisation.cpp
```diff
@@ -903,13 +903,14 @@ static void format_velocity(char** dest, size_t* size, uint16_t value)
     switch (gConfigGeneral.measurement_format)
     {
         default:
```
tupaschoal, line 905: Shouldn't you erase this `default` given that you handled all cases?

src/openrct2-ui/windows/ViewClipping.cpp
```diff
@@ -430,14 +430,14 @@ static void window_view_clipping_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                    case MeasurementFormat::Metric:
+                    case MeasurementFormat::SI:
                         clipHeightValueInMeters = static_cast<fixed32_2dp>(
                             FIXED_2DP(gClipHeight, 0) / 2 * 1.5f - FIXED_2DP(10, 50));
                         gfx_draw_string_left(
                             dpi, STR_UNIT2DP_SUFFIX_METRES, &clipHeightValueInMeters, w->colours[0], screenCoords);
                         break;
-                    case MEASUREMENT_FORMAT_IMPERIAL:
+                    case MeasurementFormat::Imperial:
                     default:
```
tupaschoal, line 441: You can probably (and should) remove this `default` too, as the switch handles all cases.

src/openrct2-ui/windows/Options.cpp
```diff
@@ -1744,13 +1744,13 @@ static void window_options_invalidate(rct_window* w)
                 switch (gConfigGeneral.measurement_format)
                 {
                     default:
```
tupaschoal, line 1746: Should probably remove this `default`

src/openrct2/localisation/Localisation.cpp
matt-thomson: So I kept it in (here and elsewhere) as a fallback in case we add another value to the enum and forget to update the switches.  That said, it seems pretty unlikely we'd do that, and these methods wouldn't be doing the right thing anyway in that case.

Happy to remove them if you'd prefer that, let me know!

tupaschoal: You pretty much explained why I'd like not to have the default labels hahahah. I think it's good to not have it because the compilation will fail if we add a new enum, which will remind us to implement proper logic for it.

matt-thomson: Is that the case though? I just tried adding another value to the enum, and everything still compiled just fine...

duncanspumpkin: Maybe on your system it does but our ci's will definitely pick it up they have generally more warnings turned on

tupaschoal: Yup, Duncan is right, not all compilers complain

matt-thomson: That works for me, thanks! I will update.

PR #12658, opened 2020-08-12
Fix freezes on incomplete packet transmission

ZehMatt:
The loop would keep going for as long the packet was not fully received which never allowed the game to actually buffer the packets causing freezes, especially during a connect this is very noticeable. Now the loop only goes on for as long it has complete packets available and keeps buffering for when it requires more data.


ZehMatt:
Also had to increase the network buffer, while the freezing was gone it made the map download quite slow with only 1 KiB. Now it has a buffer size of what the maximum packet length can be, which makes this smooth as possible.

PR #5138, opened 2017-01-29
Disable Twitch tab in versions without Twitch support

AaronVanGeffen:
When compiling OpenRCT2 without Twitch support, the Twitch options are left intact in the options menu. This PR tackles this by disabling the tab, and drawing a greyed-out sprite instead. To draw the sprite greyed-out, I'm introducing additional widget logic, too.

Addresses #5128.

PR #12661, opened 2020-08-12
Fix #12308: Cannot use cheats in editor modes

### LINKED ISSUE #12308: Cannot use cheats in Track Designer (and Scenario Editor)
Cheats have been disabled in the Track Designer, even though there are several that would be useful, like the "show all track pieces" cheat.

Gymnasiast:
This allows opening the cheats window while in editor mode. Some tabs and options are disabled because they are not relevant to these modes.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -57,6 +57,7 @@
 - Fix: [#12221] Map Generation tool doesn't place any trees.
 - Fix: [#12285] On-ride photo profit assumes every guest buys one.
 - Fix: [#12297] OpenGL renderer causing artifacts.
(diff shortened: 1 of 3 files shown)


AaronVanGeffen:
I would like this to use the disabled tab sprite introduced in #5138. With Twitch support removed, I believe no window is currently using it. However, I think it would be a perfect fit for this. To illustrate:

Gymnasiast:
Since you said that handling greying out properly requires a PR of its own to do properly, I have opted for the simplest solution: updating the positions so that there is no awkward gap between tabs 1 and 4.

PR #12663, opened 2020-08-12
Fix #10549: 'Build the best ride you can' objective missing ride name

### LINKED ISSUE #10549: "Build the best <ride name> you can" objective is missing the ride name
**OS:** Ubuntu 19.10
**Version:** 0.2.4+
**Commit/Build:** f05aeb1fcd96953f01cf9a63bb749032c98a276c

While this objective is technically present in RCT2, I have never come across a scenario that uses it, and the code around it does not seem to work under any circumstance.

Since this problem will realistically only surface with RCT1 scenarios, I will apply that label.

Originally reported as part of https://github.com/OpenRCT2/OpenRCT2/issues/8875.

afbeelding

Gymnasiast:
This is implemented in a way that is compatible with RCT2 (as long as the ride type was not split).

SCR3

### DIFF
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -296,7 +296,12 @@ public:
         gScenarioObjectiveYear = _s6.objective_year;
         // pad_013580FA
         gScenarioObjectiveCurrency = _s6.objective_currency;
-        gScenarioObjectiveNumGuests = _s6.objective_guests;
+        // In RCT2, the ride string IDs start at index STR_0002 and are directly mappable.
+        // This is not always the case in OpenRCT2, so we use the actual ride ID.
+        if (gScenarioObjectiveType == OBJECTIVE_BUILD_THE_BEST)
+            gScenarioObjectiveNumGuests = _s6.objective_guests - RCT2_RIDE_STRING_START;
+        else
+            gScenarioObjectiveNumGuests = _s6.objective_guests;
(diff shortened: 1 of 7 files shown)


src/openrct2/rct2/S6Importer.cpp
```diff
@@ -296,7 +296,12 @@ class S6Importer final : public IParkImporter
         gScenarioObjectiveYear = _s6.objective_year;
         // pad_013580FA
         gScenarioObjectiveCurrency = _s6.objective_currency;
-        gScenarioObjectiveNumGuests = _s6.objective_guests;
+        // In RCT2, the ride string IDs start at index STR_0002 and are directly mappable.
+        // This is not always the case in OpenRCT2, so we use the actual ride ID.
+        if (gScenarioObjectiveType == OBJECTIVE_BUILD_THE_BEST)
+            gScenarioObjectiveNumGuests = _s6.objective_guests - RCT2_RIDE_STRING_START;
```
duncanspumpkin, line 302: If you are going to use this for not just guests we should rename this so that it is just a objective argument instead of num guests.

Gymnasiast: It is also used for the length requirement for roller coasters.

I have a pull request coming up that creates a struct for the objective, in which I also handle this.

duncanspumpkin: Okay that's fine then

PR #12665, opened 2020-08-12
Create Objective struct and use it for global objective

Gymnasiast:


src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -272,16 +272,19 @@ void S6Exporter::Export()
...
     // This is not always the case in OpenRCT2, so we use the actual ride ID.
-    if (gScenarioObjectiveType == OBJECTIVE_BUILD_THE_BEST)
-        _s6.objective_guests = gScenarioObjectiveNumGuests + RCT2_RIDE_STRING_START;
+    if (gScenarioObjective.Type == OBJECTIVE_BUILD_THE_BEST)
+        _s6.objective_guests = gScenarioObjective.RideId + RCT2_RIDE_STRING_START;
     else
-        _s6.objective_guests = gScenarioObjectiveNumGuests;
+        _s6.objective_guests = gScenarioObjective.NumGuests;
+
+    _s6.objective_currency = gScenarioObjective.Currency;
```
Gymnasiast, line 286: Remove this line and the following one.

src/openrct2/scenario/Scenario.h
```diff
@@ -349,6 +349,34 @@ enum
...
+
+struct Objective
+{
+    uint8_t Type;
+    uint8_t Year;
+    union
+    {
+        uint16_t NumGuests;
+        rct_string_id RideId;
+        uint16_t MinimumLength; // For the "Build 10 coasters of minimum length" objective.
```
Gymnasiast, line 362: Also add minimum excitement rating here.

src/openrct2/scenario/Scenario.cpp
```diff
@@ -713,13 +710,13 @@ void scenario_remove_trackless_rides(rct_s6_data* s6)
 
 static void scenario_objective_check_guests_by()
```
duncanspumpkin, line 711: Are you planning on making these member functions?

src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -362,30 +362,30 @@ static void window_editor_objective_options_set_objective(rct_window* w, int32_t
         case OBJECTIVE_10_ROLLERCOASTERS:
             break;
         case OBJECTIVE_GUESTS_BY:
-            gScenarioObjectiveYear = 3;
-            gScenarioObjectiveNumGuests = 1500;
+            gScenarioObjective.Year = 3;
```
duncanspumpkin, line 365: I don't feel any of this default behaviours should be in the window. It should be moved to the constructor.

src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -618,26 +618,26 @@ static void window_editor_objective_options_arg_1_decrease(rct_window* w)
 
 static void window_editor_objective_options_arg_2_increase(rct_window* w)
 {
-    if (gScenarioObjectiveYear >= 25)
+    if (gScenarioObjective.Year >= 25)
```
duncanspumpkin, line 621: I feel all of these limits should also be moved to the class. Not too sure how that would look though. I guess you could go with lots of inheritance but might be overkill.

src/openrct2-ui/windows/EditorObjectiveOptions.cpp
Gymnasiast: I'm not sure about that. They're defaults that the scenario editor finds appropriate. I think the struct should be "neutral" here.

src/openrct2-ui/windows/EditorObjectiveOptions.cpp
Gymnasiast: I think the Objective struct should allow more than these limits. In a technical sense, it can take 30 years or 240 guests without any problem, it's just that the Scenario Editor enforces round limits. This is why I think these limits should stay in the Scenario Editor.

src/openrct2/scenario/Scenario.cpp
Gymnasiast: Eventually, yes.

duncanspumpkin (approved):

Issue #9088, opened 2019-04-11
Switch to more modern JSON library


AaronVanGeffen:
I'd like us to switch to a more modern JSON backend to replace the C library we're currently using.

[JSON for Modern C++](https://github.com/nlohmann/json#examples) looks very promising. It's been packaged for Homebrew, vcpkg/nuget, Arch Linux (AUR), and even [Debian](https://packages.debian.org/stretch/nlohmann-json-dev) ships it.

tupaschoal:
I can vouch for https://github.com/nlohmann/json, I've seen pretty heavy usage of it with no problems

AaronVanGeffen:
Perhaps the best way to go about this, is to introduce the new library to the codebase and then rewrite the existing code in smaller refactor projects. It's more than a little daunting otherwise.

PR #12667, opened 2020-08-13
Migrate JSON library to JSON for Modern C++

mwnciau:
This PR removes all references to the old C JSON library, and replaces them with the [JSON for Modern C++](https://github.com/nlohmann/json) library, as discussed in #9088. The latter library will need to be [downloaded](https://github.com/nlohmann/json/releases) and added to the "lib/.../include" folder.

I realise I'm a new contributor, and this is a pretty massive PR, but I have some experience with PRs of a similar scale on other OS projects before, and I thought it would be a good way to familiarise myself with the codebase. I'm willing to put further work in (including going back on some of the design decisions I've made) to get this accepted.

### JSON for Modern C++
I thought I'd summarise a few quirks with the new library that I discovered in the making of this PR. Generally, it's a lot nicer to use than the Jannson library, especially when creating JSON objects where it takes advantage of initializer lists, e.g. from NetworkBase:

```
json_t jsonProvider = {
    {"name", gConfigNetwork.provider_name},
    {"email", gConfigNetwork.provider_email},
    {"website", gConfigNetwork.provider_website},
};
```
Traversing JSON is as simple as using the `[]` operator: 
```
jsonProvider["name"] = newName;
```
There are some caveats to using the `[]` operators, however. There are a few different types of JSON node that the library recognises, including null, array, object and the value types. Using the `[]` operators on a null, array or object node type is safe, and will return a null node if the operation doesn't make sense or the key isn't found. However, using the `[]` operator on a value node will throw an exception, so the node type always needs to be checked when traversing unknown JSON. Additionally, the behaviour of `[]` behaves differently for const variables: because the JSON object cannot be changed, a null node cannot be inserted when accessing an unknown key, so it will result in an exception.

### Design Decisions

**ObjectJsonHelpers**
This namespace seemed to be a mishmash of helper functions that should really have been attached to the classes or namespaces they dealt with. Especially the Get[...] functions which were being used outside of the object classes. I have completely removed this namespace, and migrated the functions as follows (see commit 68aff48):

- Moved ParseColour to Colour::FromString
- Moved ParseCursor to Cursor::FromString
- Moved LoadStrings to StringTable::ReadJson
- Moved LoadImages to ImageTable::ReadJson
- Moved ParseObjectEntry to Object::ParseObjectEntry
- Moved GetString, etc. to Json::GetString, etc.

**const**
In the previous implementation, const JSON variables were often used. Because of the issues with exceptions being thrown when using const variables, const JSON variables are not used.

**Json namespace**
The Json namespace provides various functions that make it easier to work with the JSON library:

- GetString, GetNumber<>, GetBoolean and GetEnum perform all necessary node-type checking and type casting when retrieving a value.
- The AsObject function ensures that a given node is an object - if it isn't, it returns an empty object.
- The AsArray function ensures that a given node is an array - objects are converted to arrays and values are returned as a JSON array containing that value.
- The From[...] functions load and parse JSON, performing all the necessary checks and catches.
- The GetFlags functions allows easy conversion between a JSON object containing boolean values to a binary flag variable.

**GetFlags**
I've added an override of GetFlags that allows for inverted flags by using a third value in the initializer list, e.g. from LargeSceneryObject:
```
tile.flags = Json::GetFlags<uint16_t>(
    jTile,
    {
        // NB: Flag inverted
        {"hasSupports", LARGE_SCENERY_TILE_FLAG_NO_SUPPORTS, Json::FLAG_INVERTED},
        {"allowSupportsAbove", LARGE_SCENERY_TILE_FLAG_ALLOW_SUPPORTS_ABOVE, Json::FLAG_NORMAL}
    });
```
This seemed a bit cleaner and clearer than the previous implementation which just manually set the inverted flags after the GetFlags call.

### Epilogue

If this gets merged, the dependencies zip will need to be updated with the JSON for Modern C++ library. On discord @simonj if you want to chat. There are a few places where I'm not sure what's intended with the code (or are potential bugs). I'll go through and comment on those lines to highlight them when I get the chance. I've tested as much as I can, and can't see any issues with the latest commit, but I'll continue to test this build.


src/openrct2/core/Json.cpp
```diff
@@ -15,53 +15,112 @@
...
+            json = json_t::parse(fileData, nullptr, true, true);
+        }
+        catch (json_t::exception e)
         {
-            throw JsonException(&jsonLoadError);
+            log_error("Unable to parse JSON file (%s)\n%s", path, e.what());
         }
 
-        return json;
+        return json;
```
tupaschoal, line 42:
```suggestion
        return json;
```

src/openrct2/core/Json.cpp
```diff
@@ -15,53 +15,112 @@
...
-            throw JsonException(&jsonLoadError);
+            log_error("Unable to parse JSON file (%s)\n%s", path, e.what());
         }
 
-        return json;
+        return json; 
     }
 
-    void WriteToFile(const utf8* path, const json_t* json, size_t flags)
+    void WriteToFile(const utf8* path, json_t  jsonData, int indentSize)
```
tupaschoal:
```suggestion
    void WriteToFile(const utf8* path, json_t jsonData, int indentSize)
```

src/openrct2/core/Json.hpp
```diff
@@ -11,36 +11,187 @@
...
 #include "../common.h"
 
-#include <jansson.h>
-#include <stdexcept>
+#include <nlohmann/json.hpp>
+
 #include <string>
 #include <string_view>
 
+typedef nlohmann::json json_t;
```
tupaschoal, line 19:
```suggestion
using json_t = nlohmann::json;
```

src/openrct2/core/Json.hpp
```diff
@@ -11,36 +11,187 @@
...
+                flags = static_cast<T>(flags | item.second);
+            }
+        }
+        return flags;
+    }
+
+    /**
+     * Used by the GetFlags function to allow for inverted values
+     */ 
+    enum
```
tupaschoal, line 156: Can we make it a named `enum class`?

src/openrct2/core/Json.hpp
```diff
@@ -11,36 +11,187 @@
...
+        {
+            // true means flag is enabled
+            if (std::get<2>(item) == FLAG_NORMAL)
+            {
+                if (Json::GetBoolean(jsonObj[std::get<0>(item)]))
+                {
+                    flags = static_cast<T>(flags | std::get<1>(item));
+                }
+            }
+            // false means flag is enabled
```
tupaschoal, line 186:
```suggestion
            // false means flag is disabled
```

The comment is the same as above, so I'll just assume this is a typo. I think the enum is pretty self explanatory anyway, so I don't see much value in these comments.

src/openrct2/core/Json.cpp
```diff
@@ -15,53 +15,112 @@
...
-
-        if (json == nullptr)
+        try
+        {
+            json = json_t::parse(fileData, nullptr, true, true);
+        }
+        catch (json_t::exception e)
         {
-            throw JsonException(&jsonLoadError);
+            log_error("Unable to parse JSON file (%s)\n%s", path, e.what());
```
tupaschoal, line 39: This is a bit of a behavior change; . Assuming `JsonExceptions` were uncaught, it would cause termination of the program, whereas yours only logs and moves ahead, supposedly. Not sure if it's in a sane state afterwards.

AaronVanGeffen:
Thank you very much for working on this. I'm really happy to see this has been picked up.

src/openrct2-ui/interface/Theme.cpp
```diff
@@ -355,41 +355,40 @@ void UITheme::RemoveEntry(rct_windowclass windowClass)
...
-
-    json_object_set_new(jsonTheme, "useLightsRide", json_boolean(Flags & UITHEME_FLAG_USE_LIGHTS_RIDE));
-    json_object_set_new(jsonTheme, "useLightsPark", json_boolean(Flags & UITHEME_FLAG_USE_LIGHTS_PARK));
-    json_object_set_new(
-        jsonTheme, "useAltScenarioSelectFont", json_boolean(Flags & UITHEME_FLAG_USE_ALTERNATIVE_SCENARIO_SELECT_FONT));
-    json_object_set_new(jsonTheme, "useFullBottomToolbar", json_boolean(Flags & UITHEME_FLAG_USE_FULL_BOTTOM_TOOLBAR));
+    json_t jsonTheme = {
+        {"name", Name},
+        {"entries", jsonEntries},
+        {"useLightsRide", 0 != (Flags & UITHEME_FLAG_USE_LIGHTS_RIDE)},
```
AaronVanGeffen, line 376: I would prefer not to use 'yoda coding', i.e.:
```suggestion
        {"useLightsRide", (Flags & UITHEME_FLAG_USE_LIGHTS_RIDE) != 0},
```

src/openrct2/object/TerrainSurfaceObject.cpp
```diff
@@ -74,24 +73,45 @@ void TerrainSurfaceObject::DrawPreview(rct_drawpixelinfo* dpi, int32_t width, in
...
+                entry.Underground = Json::GetBoolean(el["underground"]);
+                SpecialEntries.push_back(std::move(entry));
+            }
+        }
+    }
+
+    auto jDefault = root["default"];
+    if (jDefault.is_object())
+    {
+        // simonj ~ I assume these should be jDefault["normal"] etc.?
```
mwnciau, line 111: The existing code used properties["normal"] instead of jDefault["normal"]. The defaults aren't actually used as far as I could see in the data files, but I assume that this wasn't intended behaviour.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -708,66 +697,57 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
...
-            json_t* y_offset = json_object_get(sprite_description, "y_offset");
+            json_t x_offset = jsonSprite["x_offset"];
+            json_t y_offset = jsonSprite["y_offset"];
 
             // Get palette option, if present
             bool keep_palette = false;
-            json_t* palette = json_object_get(sprite_description, "palette");
-            if (json_is_string(palette))
+            std::string palette = Json::GetString(jsonSprite["palette"]);
+            // simonj ~ do we need to substring here? was originally strncmp
```
mwnciau, line 725: I couldn't find an example of this file so I don't know what `jsonSprite["palette"]` should contain

Gymnasiast: Take a look at `resources/g2/sprites.json`.

AaronVanGeffen: An example would be [resources/g2/sprites.json](https://github.com/OpenRCT2/OpenRCT2/blob/develop/resources/g2/sprites.json)

Gymnasiast:
Please do not mix unrelated changes in this PR. The Cursors.h rename should be its own PR.

mwnciau:
> Please do not mix unrelated changes in this PR. The Cursors.h rename should be its own PR.

Removed the Cursors.h rename from this PR

AaronVanGeffen:
> Removed the Cursors.h rename from this PR

Please cherry-pick it to another, though. It's a good change, for the reasons you mentioned earlier. 🙂

mwnciau:
> I've only really looked into the first commit, but this looks very promising. I found these a bit hard to follow:
> 
> > Additionally, the behaviour of [] behaves differently for const variables: because the JSON object cannot be changed, a null node cannot be inserted when accessing an unknown key, so it will result in an exception.
> 
> > const
> > In the previous implementation, const JSON variables were often used. Because of the issues with exceptions being thrown when using const variables, const JSON variables are not used.
> 
> Seems like our usage is mixing up reads and writes, am I right? Because I would rather have it split with either methods having `const &` or `&`, depending on usage, than everything being just `&`.

It was a problem I was running into early on into making this PR, and at the time I just decided to remove all const parameters and revisit it later. Actually, now that I understand the library a bit better, there are a few places where const parameters can easily be added so I'll make those changes. There are still a few places where a const parameters would make sense, but there's a cost in code complexity, e.g. compare the follow two methods:

```
void SomeObject::ReadJson(json_t& json)
{
    _var = Json::GetString(json["key"], "default");
}
```
Changing the above parameter to be const would cause the code to throw an exception if `json["key"]` didn't exist. So we check the key exists before trying to access it:
```
void SomeObject::ReadJson(const json_t& json)
{
    if (json.contains("key"))
    {
        _var = Json::GetString(json["key"], "default");
    }
    else
    {
        _var = "default";
    }
}
```
Not a huge deal if it's a one off, but when some of the object ReadJson methods contain 10+ lines of such assignments, it clutters the code a lot. On the other hand, an additional helper function like `Json::GetString(json_t json, std::string key, std::string default)` would shorten it, but then you lose a bit of consistency in how child nodes are accessed, which I was trying to avoid. I'll leave this one up to consensus.

> As this is a big PR, I have to thank you for doing it in lots of separate, well-defined commits. I'd still like to see some of them squashed, like [dedb26f](https://github.com/OpenRCT2/OpenRCT2/commit/dedb26f7f140a1b8ed8262b7698588019939b44b) into [711e336](https://github.com/OpenRCT2/OpenRCT2/commit/711e336b58e14f6b7cfc137909c43414320ed337). That will make it easier for you when we review, because as you'll see below, I've opened issues for things that are fixed down the commit range. So "fixes" to something that have been introduced in earlier commits would be very desirable.
> 
> All in all, thanks for the PR, like I said this looks very promising and I look forward to working to get it merged. Note that we do intend to release soon, so it's very likely that we will hold the merge for a bit.

I've just squashed all the minor fixes and typos into the original commits that they were introduced in.

janisozaur:
Many thanks for picking up this work. We're about to do a release and this certainly won't fit, but hope to review and merge it soon after.

src/openrct2/core/Json.cpp
mwnciau: This discussion may be a little out of the scope of this PR, but I've been thinking about the desired behaviour for exceptions since you mentioned it. You're right in that things get a little fuzzy when things don't load (you basically just get a black screen), but do we _always_ want to throw an exception if the JSON couldn't be parsed?

- It makes sense that the core files like rides, etc. would throw an exception as these will never change for 99% of users and anyone changing them would want to fix any problems.
- Do we want exceptions to be thrown for "less important" json files such as the user list or the plugin store?
- Most of the `Object::FromJson` methods for rides rely on having a "properties" key - do we want to throw an exception if this doesn't exist (this would actually make the diffs a bit easier to follow for this PR because I've added indentation for a check on the properties variable).
- We probably _don't_ want exceptions to be thrown by the `Json::FromString`/`Json::FromVector` functions so the game doesn't crash from, e.g. a bad packet.

The From[...] functions will return a `json_t` of type null only if the file couldn't load or the parsing failed, so it would be fairly easy to add an assertion on the returned value.

tupaschoal: You're right in the sense that we can definitely do better with regards to exception handling, we might not really need to throw every time something fails. That being said, we still want to throw for some stuff and getting the ideal "exception behavior" in this PR might add too much to handle in here.

My suggestions:
- Don't change the exception behavior in here, keep it as is. Let's make this PR about upgrading to modern c++ and doing the changes required for it to work, which will make it easier to review and speed-up the process of merging it.
- You can create an issue with the exact text above and we can all discuss over there on the next steps

AaronVanGeffen:
This is still marked as a draft, but it is more or less ready, right? When you think it's ready for review, we can update the dependency bundles and Docker images so this can be merged.

src/openrct2/core/Json.cpp
mwnciau: I've now reverted the above change and a few others so that the exception behaviour should be the same. I've realised that the actual behaviour is not far off from what I've written above.

mwnciau:
Yep, I think it's more or less ready to review. It's up to the latest commit of develop as of right now.

AaronVanGeffen:
Pending https://github.com/OpenRCT2/Dependencies/pull/22

Similar changes will be needed for Linux / our Docker images. @janisozaur, could you look into that?

AaronVanGeffen:
Just to update: macOS CI is now passing. Windows CI isn't yet, but I'm no Visual Studio adept. I'd appreciate it if someone else could have a look there. The Docker images also still need updating for our Linux and Android builds to pass.

mwnciau:
Currently the CI is failing with the following, and it's failing for the same reason on my local machine:

```
freetype.lib(sfnt.c.obj) : error LNK2001: unresolved external symbol BrotliDecoderDecompress [D:\a\OpenRCT2\OpenRCT2\src\openrct2-cli\openrct2-cli.vcxproj]
```

Possibly a problem with the latest version of freetype/its dependencies in the new dependency release?

IntelOrca:
brotlicommon-static.lib, brotlidec-static.lib and brotlienc-static.lib need adding to the lib list due to freetype now requiring brotli.

AaronVanGeffen:
Looks like that did the trick. That leaves the Docker images to take care of.

tupaschoal:
And some reviews too :P
I'm kinda waiting for the dependencies to be sorted to scour through the commits again

AaronVanGeffen:
@mwnciau We're seeing some error messages on the Linux CI side. The library does appear to be available, but certain member functions are triggering errors. Do you have an idea as to what may be causing the errors?

mwnciau:
> @mwnciau We're seeing some error messages on the Linux CI side. The library does appear to be available, but certain member functions are triggering errors. Do you have an idea as to what may be causing the errors?

Looks like the version of nlohmann-json that the ubuntu package manager has is quite old (v2.1.1-1.1) and the contains method was introduced in 3.6.0. Is there any way to manually load the library in the docker image? I can always undo the changes that caused these problems (change the parameter back to non-const so we don't need contains() checks).

janisozaur:
@mwnciau we've just installed the `nlohmann-json` package inside the container used for CI, but other images (e.g. mingw) just wget the file. You can do the same: https://github.com/OpenRCT2/openrct2-docker-build/pull/1

mwnciau:
Added a [PR on OpenRCT2/openrct2-docker-build](https://github.com/OpenRCT2/openrct2-docker-build/pull/4) to download the latest version. It now builds in my docker image (or at least, gets further - a couple of errors I need to look at).

AaronVanGeffen:
Looks like Android is the last holdout. I've filed https://github.com/OpenRCT2/openrct2-docker-build/pull/5 to -hopefully- address this.

janisozaur:
Needs a rebase before the CI kicks in. @mwnciau @AaronVanGeffen I checked what the conflicts and it's https://github.com/OpenRCT2/OpenRCT2/pull/12796. Can you please resolve that?

mwnciau:
Rebased to the latest dev branch and all tests are passing now. I've also merged a few of the smaller new commits to make reviewing a bit easier.

tupaschoal:
I'll do my best to review tomorrow

tupaschoal:
I won't be able to review those today, give me by the end of the week, please :)

janisozaur:
Just remembered: this will break our PPA builds. Please hold.

janisozaur:
@Gymnasiast could you check if focal's package is recent enough? If it is, I will drop the outdated versions from PPA recipe and it will require update to `debian/rules` file.

janisozaur:
Also ping @pkubaj @PureTryOut @joepie91 @geistesk @ibara @rmnvgr

ibara:
Will one of the following two scenarios be true:
1. This new library will live with the rest of OpenRCT2; in other words, when I download the OpenRCT2 repository, will this new JSON library be there or will it be a separate download?
2. Will OpenRCT2 detect and use a system installation of the new JSON library?

janisozaur:
The current state of PR does not include the library and relies on one from the system. There is no configure-time check for its presence.

As this is a single-header library, OpenRCT2 will consume it at compile-time. The reason for not including it in the repo is it's a large library (~1MB of code) with fairly frequent releases and one that's already packaged in the repositories.

janisozaur:
Ok, https://packages.ubuntu.com/focal/nlohmann-json3-dev suggests it is recent enough, just needs the rules update then.

src/openrct2/network/NetworkBase.cpp
```diff
@@ -4248,8 +4239,8 @@ NetworkServerState_t network_get_server_state()
 {
     return NetworkServerState_t{};
 }
-json_t* network_get_server_info_as_json()
+json_t network_get_server_info_as_json()
 {
-    return nullptr;
+    return json_t();
```
tupaschoal, line 4244:
```suggestion
    return {};
```

PS: I kind wonder how useful this function really is

src/openrct2/network/NetworkGroup.cpp
```diff
@@ -14,31 +14,27 @@
 #    include "NetworkAction.h"
 #    include "NetworkTypes.h"
 
-NetworkGroup NetworkGroup::FromJson(const json_t* json)
+NetworkGroup NetworkGroup::FromJson(json_t& jsonData)
```
tupaschoal, line 17:
```suggestion
NetworkGroup NetworkGroup::FromJson(const json_t& jsonData)
```

src/openrct2/network/NetworkGroup.cpp
```diff
@@ -14,31 +14,27 @@
...
     }
 
-    group.Id = static_cast<uint8_t>(json_integer_value(jsonId));
-    group._name = std::string(json_string_value(jsonName));
+    group.Id = Json::GetNumber<uint8_t>(jsonId);
+    group._name = Json::GetString(jsonName);
     std::fill(group.ActionsAllowed.begin(), group.ActionsAllowed.end(), 0);
 
-    for (size_t i = 0; i < json_array_size(jsonPermissions); i++)
+    for (auto& jsonValue : jsonPermissions)
```
tupaschoal, line 33:
```suggestion
    for (const auto& jsonValue : jsonPermissions)
```

src/openrct2/network/NetworkGroup.cpp
```diff
@@ -14,31 +14,27 @@
...
     {
-        json_t* jsonPermissionValue = json_array_get(jsonPermissions, i);
-        const char* perm_name = json_string_value(jsonPermissionValue);
-        if (perm_name == nullptr)
-        {
-            continue;
-        }
-        int32_t action_id = NetworkActions::FindCommandByPermissionName(perm_name);
+        std::string permission = Json::GetString(jsonValue);
```
tupaschoal, line 35:
```suggestion
        const std::string permission = Json::GetString(jsonValue);
```

src/openrct2/network/NetworkServerAdvertiser.cpp
```diff
@@ -221,86 +214,84 @@ class NetworkServerAdvertiser final : public INetworkServerAdvertiser
...
             this->OnHeartbeatResponse(root);
-            json_decref(root);
         });
     }
 
-    void OnRegistrationResponse(json_t* jsonRoot)
+    /**
+     * @param jsonRoot must be of JSON type object or null
+     */
+    void OnRegistrationResponse(json_t& jsonRoot)
```
tupaschoal, line 226:
```suggestion
    void OnRegistrationResponse(const json_t& jsonRoot)
```

src/openrct2/network/NetworkServerAdvertiser.cpp
```diff
@@ -221,86 +214,84 @@ class NetworkServerAdvertiser final : public INetworkServerAdvertiser
...
+                log_info("Retry with ipv4 only");
             }
         }
     }
 
-    void OnHeartbeatResponse(json_t* jsonRoot)
+    /**
+     * @param jsonRoot must be of JSON type object or null
+     */
+    void OnHeartbeatResponse(json_t& jsonRoot)
```
tupaschoal, line 262:
```suggestion
    void OnHeartbeatResponse(const json_t& jsonRoot)
```

src/openrct2/network/NetworkUser.cpp
```diff
@@ -21,50 +20,41 @@
 
 constexpr const utf8* USER_STORE_FILENAME = "users.json";
 
-NetworkUser* NetworkUser::FromJson(json_t* json)
+NetworkUser* NetworkUser::FromJson(json_t& jsonData)
```
tupaschoal, line 23:
```suggestion
NetworkUser* NetworkUser::FromJson(const json_t& jsonData)
```

src/openrct2/network/NetworkUser.cpp
```diff
@@ -21,50 +20,41 @@
...
 {
-    const char* hash = json_string_value(json_object_get(json, "hash"));
-    const char* name = json_string_value(json_object_get(json, "name"));
-    const json_t* jsonGroupId = json_object_get(json, "groupId");
+    std::string hash = Json::GetString(jsonData["hash"]);
+    std::string name = Json::GetString(jsonData["name"]);
```
tupaschoal, line 26:
```suggestion
    const std::string hash = Json::GetString(jsonData["hash"]);
    const std::string name = Json::GetString(jsonData["name"]);
```

src/openrct2/network/NetworkUser.cpp
```diff
@@ -92,18 +82,18 @@ void NetworkUserManager::Load()
 
         try
         {
-            json_t* jsonUsers = Json::ReadFromFile(path);
-            size_t numUsers = json_array_size(jsonUsers);
-            for (size_t i = 0; i < numUsers; i++)
+            json_t jsonUsers = Json::ReadFromFile(path);
+            for (auto& jsonUser : jsonUsers)
```
tupaschoal, line 86:
```suggestion
            for (const auto& jsonUser : jsonUsers)
```

src/openrct2/network/NetworkUser.cpp
```diff
@@ -92,18 +82,18 @@ void NetworkUserManager::Load()
...
             {
-                json_t* jsonUser = json_array_get(jsonUsers, i);
-                NetworkUser* networkUser = NetworkUser::FromJson(jsonUser);
-                if (networkUser != nullptr)
+                if (jsonUser.is_object())
                 {
-                    _usersByHash[networkUser->Hash] = networkUser;
+                    NetworkUser* networkUser = NetworkUser::FromJson(jsonUser);
```
tupaschoal, line 90:
```suggestion
                    auto networkUser = NetworkUser::FromJson(jsonUser);
```

src/openrct2/network/NetworkUser.cpp
```diff
@@ -129,36 +119,35 @@ void NetworkUserManager::Save()
...
-        const char* hash = json_string_value(json_object_get(jsonUser, "hash"));
-        if (hash != nullptr)
+        json_t jsonUser = *it;
+        if (!jsonUser.is_object())
+        {
+            continue;
+        }
+        std::string hashString = Json::GetString(jsonUser["hash"]);
+
+        const NetworkUser* networkUser = GetUserByHash(hashString);
```
tupaschoal, line 133:
```suggestion
        const auto networkUser = GetUserByHash(hashString);
```

src/openrct2/network/ServerList.cpp
```diff
@@ -70,35 +70,40 @@ bool ServerListEntry::IsVersionValid() const
     return Version.empty() || Version == network_get_version();
 }
 
-std::optional<ServerListEntry> ServerListEntry::FromJson(const json_t* server)
+std::optional<ServerListEntry> ServerListEntry::FromJson(json_t& server)
```
tupaschoal, line 73:
```suggestion
std::optional<ServerListEntry> ServerListEntry::FromJson(const json_t& server)
```

src/openrct2/network/ServerList.cpp
```diff
@@ -70,35 +70,40 @@ bool ServerListEntry::IsVersionValid() const
...
-
-    if (name == nullptr || version == nullptr)
+    auto port = Json::GetNumber<int32_t>(server["port"]);
+    auto name = Json::GetString(server["name"]);
+    auto description = Json::GetString(server["description"]);
+    auto requiresPassword = Json::GetBoolean(server["requiresPassword"]);
+    auto version = Json::GetString(server["version"]);
+    auto players = Json::GetNumber<uint8_t>(server["players"]);
+    auto maxPlayers = Json::GetNumber<uint8_t>(server["maxPlayers"]);
+    std::string ip = "";
```
tupaschoal, line 82:
```suggestion
    std::string ip;
```
No need to initialize a `std::string` object

src/openrct2/network/ServerList.h
```diff
@@ -35,7 +35,12 @@ struct ServerListEntry
     int32_t CompareTo(const ServerListEntry& other) const;
     bool IsVersionValid() const;
 
-    static std::optional<ServerListEntry> FromJson(const json_t* root);
+    /**
+     * Constructs a ServerListEntry from JSON
+     * @param server JSON representing server. Must be object type.
```
tupaschoal, line 40: We're not enforcing that it it an object type on the function, maybe we should

src/openrct2/object/StringTable.cpp
```diff
@@ -78,6 +78,43 @@ void StringTable::Read(IReadObjectContext* context, OpenRCT2::IStream* stream, O
...
+    if (s == "capacity")
+        return ObjectStringID::CAPACITY;
+    if (s == "vehicleName")
+        return ObjectStringID::VEHICLE_NAME;
+    return ObjectStringID::UNKNOWN;
+}
+
+void StringTable::ReadJson(json_t& root)
+{
+    // We trust the JSON type of root is object
```
tupaschoal, line 96: Should we? Maybe we should assert

src/openrct2/object/RideObject.cpp
```diff
@@ -523,305 +522,287 @@ uint8_t RideObject::CalculateNumHorizontalFrames(const rct_ride_entry_vehicle* v
...
             {
-                if (json_is_array(route))
+                if (jRoute.is_array())
                 {
-                    size_t j;
-                    json_t* waypoint;
                     std::array<CoordsXY, 3> entry;
-                    json_array_foreach(route, j, waypoint)
+
+                    for (size_t j = 0; j < 3; j++)
```
tupaschoal, line 762:
```suggestion
                    for (size_t j = 0; j < 3; ++j)
```
Why did we lose the range based here? Sounds dangerous to have the 3 hard coded

src/openrct2/CmdlineSprite.cpp
```diff
@@ -708,66 +696,44 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
...
-            json_t* forceBmpObject = json_object_get(sprite_description, "forceBmp");
-            if (palette && json_is_boolean(forceBmpObject))
-            {
-                forceBmp = json_boolean_value(forceBmpObject);
-            }
+            json_t x_offset = jsonSprite["x_offset"];
+            json_t y_offset = jsonSprite["y_offset"];
+
+            bool keep_palette = Json::GetString(jsonSprite["palette"]) == "keep";
+            bool forceBmp = Json::GetBoolean(jsonSprite["forceBmp"]);
```
tupaschoal, line 721: This used to be set only if `jsonSprite["palette"]` was set, which is not the case anymore.

readme.md
```diff
@@ -144,7 +144,7 @@ The recommended way of building OpenRCT2 for macOS is with Xcode. The Xcode buil
 #### CMake:
 A command line version of OpenRCT2 can be built using CMake. This type of build requires you to provide the dependencies yourself. The supported method of doing this is with [Homebrew](http://brew.sh). Once you have Homebrew installed, you can download all the required libraries with this command:
 ```
-brew install cmake duktape freetype icu4c jansson libpng libzip openssl pkg-config sdl2 speexdsp
+brew install cmake duktape freetype icu4c nlohmann-json libpng libzip openssl pkg-config sdl2 speexdsp
```
tupaschoal, line 147: The list of packages here used to be sorted

debian/control
```diff
@@ -4,7 +4,7 @@ Section: misc
 Priority: optional
 Standards-Version: 3.9.2
 Multi-Arch: same
-Build-Depends: debhelper (>= 9), cmake (>= 3.8), duktape-dev, libsdl2-dev, g++ (>= 4:7), pkg-config, libjansson4 (>= 2.5), libjansson-dev (>= 2.3), libspeex-dev, libspeexdsp-dev, libcurl4-openssl-dev, libcrypto++-dev, libfontconfig1-dev, libfreetype6-dev, libpng-dev, libssl-dev, libzip-dev (>= 1.0.0), libicu-dev (>= 59.0)
+Build-Depends: debhelper (>= 9), cmake (>= 3.8), duktape-dev, libsdl2-dev, g++ (>= 4:7), pkg-config, nlohmann-json3-dev (>= 3.7), libspeex-dev, libspeexdsp-dev, libcurl4-openssl-dev, libcrypto++-dev, libfontconfig1-dev, libfreetype6-dev, libpng-dev, libssl-dev, libzip-dev (>= 1.0.0), libicu-dev (>= 59.0)
```
tupaschoal, line 7: Readme change says 3.6 (https://github.com/OpenRCT2/OpenRCT2/pull/12667/commits/d028fa27cb3d1de10c7dbb4283261f1f36a1d19d) and this says 3.7, we should have it the same everywhere

tupaschoal (commented):
This is a great PR, thank you, once again!

I reviewed it commit by commit, so I advise you to fix as such. I opened lots of `const&` issues for the `json_t` functions and forgot about what you had originally written, so feel free to drop those. I'm still a bit concerned that someone in the future will just try to `const&` them and get into a pitfall.

Lots of commits at the end, sorting the removal of jansson and new build images, seem like they can be squashed for clarity and for the sake of having each commit compile on its own, otherwise we'll have to squash everything on merge, which I really don't want to do, as it's neatly divided and organized as it is.

Finally, https://github.com/OpenRCT2/OpenRCT2/pull/12667/commits/1a085bde998367e8daa4c89e250141e2a5eae195 seems like it doesn't belong here? Not sure why it was added

mwnciau:
Thanks for the review. I'll be away for the next week so won't be able to make any changes, but I'll have a look when I'm back.

> I'm still a bit concerned that someone in the future will just try to const& them and get into a pitfall.

It might be a good idea to add a comment in the documentation to explain why they're not const.

Gymnasiast:
@janisozaur 
> Gymnasiast could you check if focal's package is recent enough? If it is, I will drop the outdated versions from PPA recipe and it will require update to `debian/rules` file.

Focal has version `3.7.3-1`. Is that recent enough?

src/openrct2/network/NetworkBase.cpp
```diff
@@ -1030,41 +1029,40 @@ void NetworkBase::LoadGroups()
...
-        size_t groupCount = json_array_size(json_groups);
-        for (size_t i = 0; i < groupCount; i++)
+        json_t jsonGroups = jsonGroupConfig["groups"];
+        if (jsonGroups.is_array())
         {
-            json_t* jsonGroup = json_array_get(json_groups, i);
-
-            auto newgroup = std::make_unique<NetworkGroup>(NetworkGroup::FromJson(jsonGroup));
-            group_list.push_back(std::move(newgroup));
+            for (auto& jsonGroup : jsonGroups)
```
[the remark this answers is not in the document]
mwnciau: This one doesn't work because NetworkGroup::FromJson doesn't accept a const jsonGroup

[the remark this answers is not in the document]
Gymnasiast: Could the signature of `NetworkGroup::FromJson` be modified to accept it?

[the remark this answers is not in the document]
mwnciau: (Also brought up [here](https://github.com/OpenRCT2/OpenRCT2/pull/12667#issuecomment-673454586))

Not without extensive modification of this PR. A const json_t will throw an exception if you try to access a member that doesn't exist (i.e. `json["key"]` where "key" doesn't exist) whereas a non-const json_t will just return a null json_t object. Changing parameters to const would require that this:

```
std::string someString = Json::GetString(json["somestring"], "default value");
```
be changed to:
```
std::string someString = "default value";
if ( json.contains("somestring") )
{
    someString = Json::GetString(json["somestring"], "default value");
}
```
which is a little verbose, especially considering this is done 10-20 lines in a row in places.

Alternatively, we could define a Json::GetObject method:
```
std::string someString = Json::GetString(Json::GetObject(json, "somestring"), "default value");
```
but, having had to understand all the existing code where something similar is done, this is nowhere near as clear as the current implementation.

So not having const parameters is a compromise for code readability. As I mentioned [here](https://github.com/OpenRCT2/OpenRCT2/pull/12667#issuecomment-687581380), I do intend to add comments where const would otherwise be defined to make this clear. (@tupaschoal I'd forgotten that you'd mentioned to ignore these)

src/openrct2/network/ServerList.h
mwnciau: There are a few functions like this. With this and with the others, there's always a check further up the callstack that this is an object. I'm not sure whether we should just double up on the checks.

src/openrct2/object/RideObject.cpp
mwnciau: `entry` is of fixed size 3 (see defintion on line 760), and access of points beyond that results in undefined behaviour. The existing code didn't check for this.

mwnciau:
I've gone through and resolved most review comments, and commented on a couple of others. @tupaschoal Nice spots on a few of those issues. I've kept all the changes I've done as separate commits so it's easier to see what has changed, but they can be merged with the existing commits after review.

Couple of changes:
* I've added `Guard::Assert`s to any function that expects a json type object (the errors are very obtuse without the asserts)
* I've added comments to any function that has a parameter that isn't `const json_t&` (plus a few loops)
* `Json::FromVector` now handles exceptions like the other `From*` functions

@janisozaur I'll leave you/others to comment on 1a085bd and the [list of packages](https://github.com/OpenRCT2/OpenRCT2/pull/12667#discussion_r483853996)

Also rebased again to the latest version of develop to solve merge conflicts.

readme.md
tupaschoal: This is still pending

src/openrct2/core/Json.cpp
```diff
@@ -26,42 +25,120 @@ namespace Json
...
+
+    json_t AsObject(const json_t& jsonObj)
+    {
+        if (jsonObj.is_object())
+        {
+            return jsonObj;
+        }
+
+        return json_t::object();
+    }
```
tupaschoal, line 119:
```suggestion
    std::string GetString(const json_t& jsonObj, const std::string& defaultValue)
    {
        return jsonObj.is_string() ? jsonObj.get<std::string>() : defaultValue;
    }

    bool GetBoolean(const json_t& jsonObj, bool defaultValue)
    {
        return jsonObj.is_boolean() ? jsonObj.get<bool>() : defaultValue;
    }

    json_t AsObject(const json_t& jsonObj)
    {
        return jsonObj.is_object() ? jsonObj : json_t::object();
    }
```

I'd go for ternaries in here, but I know some of the other guys in the team don't love it, so you don't need to do anything in here if you don't want ;P

src/openrct2/object/ImageTable.cpp
```diff
@@ -97,6 +368,71 @@ void ImageTable::Read(IReadObjectContext* context, OpenRCT2::IStream* stream)
...
+{
+    Guard::Assert(root.is_object(), "ImageTable::ReadJson expects parameter root to be object");
+
+    if (context->ShouldLoadImages())
+    {
+        // First gather all the required images from inspecting the JSON
+        std::vector<std::unique_ptr<RequiredImage>> allImages;
+        auto jsonImages = root["images"];
+
+        // json_array_foreach(jsonImages, i, el)
```
tupaschoal, line 381: Does it make sense to keep this comment?

tupaschoal (commented):
I did a quick run through and it looks good to go. Is there any pending issue to solve?

If there isn't, I guess you can merge each fixup into its respective commit and I can do a complete review again. If not, we can wait for whatever is pending.

src/openrct2/object/ImageTable.cpp
Gymnasiast: Commented-out code should be taken out unless there is a damn good reason why it's there (and even then it should be taken out).

src/openrct2/core/Json.cpp
Gymnasiast: I don't hate ternaries per se, but I am far more cautious because they get hard to read very quickly.

In this case, though, I think using ternaries is fine - I think they actually improve legibility here.

AaronVanGeffen: +1 from me for ternaries here.

src/openrct2/object/ImageTable.cpp
IntelOrca: I doubt it is intentional, likely commented out when writing the new code and then forgot to delete it afterwards.

mwnciau:
I've now merged all the commits into the originals, squashed some of the later commits together in hopefully a sensible way, and rebased to the latest develop to solve some merging issues. Also implemented the ternary operator suggestion and removed the old commented line of code.

The 2 issues that remain I don't really have the knowledge to solve (regarding the homebrew dependencies and the last commit in the PR), but they're readme issues rather than code issues so it's probably ok to go ahead and do a complete review again.

readme.md
janisozaur:
```suggestion
brew install cmake duktape freetype icu4c libpng libzip nlohmann-json openssl pkg-config sdl2 speexdsp
```

janisozaur:
Regarding what was requested in https://github.com/OpenRCT2/OpenRCT2/pull/12667#issuecomment-692372796 of me: all ok. I updated the readme to make the list sorted, but don't see it as stopper.

src/openrct2/core/Json.hpp
```diff
@@ -11,36 +11,181 @@
...
+     * @param jsonObj JSON object holding the value
+     * @param defaultValue Default value to return if the JSON object is not a string
+     * @return Copy of the JSON value converted to std::string
+     */
+    std::string GetString(const json_t& jsonObj, const std::string& defaultValue = std::string());
+
+    /**
+     * Explicit type conversion between a JSON object and a boolean
+     * @param jsonObj JSON object holding the value
+     * @param defaultValue Default value to return if the JSON object is not a string
```
tupaschoal, line 97:
```suggestion
     * @param defaultValue Default value to return if the JSON object is not a bool
```

src/openrct2/core/Json.hpp
```diff
@@ -11,36 +11,181 @@
...
+        static_assert(std::is_arithmetic<T>::value, "GetNumber template parameter must be arithmetic");
+
+        return jsonObj.is_number() ? jsonObj.get<T>() : defaultValue;
+    }
+
+    /**
+     * Explicit type conversion between a JSON object and a compatible enum value
+     * @param T Destination enum
+     * @param jsonObj JSON object holding the value
+     * @param defaultValue Default value to return if the JSON object is not a number type
```
tupaschoal, line 76:
```suggestion
     * @param defaultValue Default value to return if the JSON object is not an enum type
```

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -390,98 +394,81 @@ namespace ObjectFactory
         log_verbose("CreateObjectFromJsonFile(\"%s\")", path.c_str());
 
         Object* result = nullptr;
+        json_t jRoot;
         try
         {
-            auto jRoot = Json::ReadFromFile(path.c_str());
+            jRoot = Json::ReadFromFile(path.c_str());
```
tupaschoal, line 400:
```suggestion
            json_t jRoot = Json::ReadFromFile(path.c_str());
```

Seems like you can reduce the scope of `jRoot` above

src/openrct2-ui/interface/Theme.cpp
```diff
@@ -266,45 +271,49 @@ static void ThrowThemeLoadException()
...
-UIThemeWindowEntry UIThemeWindowEntry::FromJson(const WindowThemeDesc* wtDesc, const json_t* json)
+UIThemeWindowEntry UIThemeWindowEntry::FromJson(const WindowThemeDesc* wtDesc, json_t& jsonData)
 {
-    json_t* jsonColours = json_object_get(json, "colours");
-    if (jsonColours == nullptr)
+    Guard::Assert(jsonData.is_object(), "UIThemeWindowEntry::FromJson expects parameter jsonData to be object");
+
+    auto jsonColours = Json::AsArray(jsonData["colours"]);
+
+    if (jsonColours.size() == 0)
```
tupaschoal, line 302:
```suggestion
    if (jsonColours.empty())
```

Can't we use `.empty()` here?

tupaschoal (commented):
Looks great to me, save for some very minor details now

mwnciau:
I've committed those four suggestions now so should be good to go, unless anyone else would like to do a complete review of the code.

Apologies for not sorting that list; I was just being a bit dense - I thought you meant sort it _out_, not sort it... 🤦

tupaschoal (approved):
Looks good to me, thank you very much!

If no other team member says anything I'll merge it over the next few days

janisozaur (approved):

src/openrct2/core/Json.cpp
```diff
@@ -26,42 +25,101 @@ namespace Json
...
-        json_error_t jsonLoadError;
-        json = json_loads(fileData, 0, &jsonLoadError);
-        Memory::Free(fileData);
+        json_t json;
 
-        if (json == nullptr)
+        try
         {
-            throw JsonException(&jsonLoadError);
+            json = json_t::parse(fileData, nullptr, true, true);
```
Gymnasiast, line 35: Any reason why you have specified all optional elements with their defaults?

Asking because specifying the fourth argument causes it to fail compilation on Ubuntu 20.04, using nlohmann/json version 3.7.3-1.

PR #12671, opened 2020-08-13
MakeTab wrapper for standard tab widgets

AaronVanGeffen:
This PR introduces a `MakeTab` wrapper function, covering common arguments for tab widgets. In the current state of this PR, it has been applied to only a few windows. I intend to cover the rest as well, after an initial round of feedback w.r.t. formatting.

To do:
- [x] Cheats
- [x] EditorInventionsList
- [x] EditorObjectiveOptions
- [x] EditorObjectSelection
- [x] EditorScenarioOptions
- [x] Finances
- [x] Guest
- [x] GuestList
- [x] MapGen
- [x] Multiplayer
- [x] Network
- [x] NewRide
- [x] NewsOptions
- [x] Options
- [x] Park
- [x] Player
- [x] Research
- [x] Ride
- [x] RideList
- [x] Scenery
- [x] Staff
- [x] StaffList
- [x] Themes
- [x] TitleEditor


Gymnasiast:
I'm all for a consistent approach between windows, but I'm not sold on greying out tabs instead of hiding them. But perhaps you can convince me otherwise.

pizza2004:
I like it personally as a user.  It makes it clear that the tabs are missing because they're disabled and not some sort of accidental bug, if nothing else.

tupaschoal:
> I'm all for a consistent approach between windows, but I'm not sold on greying out tabs instead of hiding them. But perhaps you can convince me otherwise.

I'm ok with greying out, but I think it looks too harsh and it seems like it misrendered

AaronVanGeffen:
From a UI perspective, I like seeing what choices I have available in the window, even I can't use them right now. Disabling the tabs makes it clearer that the scenario editor has fewer options available. Hiding them, as is the case right now, might make someone wonder where they have gone, and whether it's a bug. Therefore, I much prefer to grey out the options. However, I agree the effect looks a bit too harsh.

Perhaps I should explain how this came to be. Unfortunately, the sprite drawing functions do not support desaturating the image. The current method is basically using the image as a mask to draw a solid outline.

Back in 2017, my initial proposal in #5138 looked like this (with exaggerated colours):

image

@marijnvdwerf had the following advice to improve it: 
> * The inactive tab doesn't look disabled
> 
> * The inactive tab appears to hover in front of the content frame; breaking the illusion of tabs
> 
> * The embossed icon only has a highlight, whereas embossed/disabled text uses two colours to make it look recessed:
> image
> 
> By removing the tab icon and taking the window bg colour, you are able to get this effect
> image.
> 
> If you layer some filtered sprites, you're even able to have a monotone image, without resorting to colour-filling it:
> image
> 
> With some extra graphics, you might be able to add a 'disabled tab' image:
> image

It looked good enough for the Twitch icon, but I agree it doesn't look quite right in the cheats window.

Any ideas?

AaronVanGeffen:
An interesting issue has popped up. Due to how I'm currently assigning the sprite based on whether or not a tab is disabled, the 'Select rides to be preserved' tab is now always showing in the objective selection window. It is only used when converting an existing save, i.e. not in scenarios without rides.

Unnamed park 2020-08-17 17-31-52

For discoverability, I actually like that it shows up. Clearly, it needs work, in that the sprite is missing. However, I thought I'd put it here for discussion.

IntelOrca:
The ride window currently hides all non-applicable tabs, are you suggesting they now be visible and greyed out?

AaronVanGeffen:
> The ride window currently hides all non-applicable tabs, are you suggesting they now be visible and greyed out?
> Same goes for the money tab on all windows when money is disabled.

I had honestly forgotten about this. You're right, that shouldn't happen. Yet it is:

Diamond Heights 2020-08-17 20-10-26

Fixing this would cost more time than I would like to spend on it, so I concede. I'll rebase this PR to undo the changes that disable tabs.

src/openrct2-ui/windows/Research.cpp
```diff
@@ -62,31 +62,31 @@ enum {
...
-    MakeWidget     ({  3,  47}, {WW_DEVELOPMENT - 10,  70}, WWT_GROUPBOX, 2, STR_CURRENTLY_IN_DEVELOPMENT                                  ),
-    MakeWidget     ({  3, 124}, {WW_DEVELOPMENT - 10,  65}, WWT_GROUPBOX, 2, STR_LAST_DEVELOPMENT                                          ),
-    MakeWidget     ({265, 161}, {                 24,  24}, WWT_FLATBTN,  2, 0xFFFFFFFF,                   STR_RESEARCH_SHOW_DETAILS_TIP   ),
+    MakeWidget({  0,  43}, {     WW_DEVELOPMENT, 153}, WWT_RESIZE,   1),
+    MakeTab   ({  3,  17}, STR_RESEARCH_AND_DEVELOPMENT_TIP           ),
+    MakeTab   ({ 34,  17}, STR_FINANCES_RESEARCH_TIP                  ),
+    MakeWidget({  3,  47}, {WW_DEVELOPMENT - 10,  70}, WWT_GROUPBOX, 2, STR_CURRENTLY_IN_DEVELOPMENT                                  ),
+    MakeWidget({  3, 124}, {WW_DEVELOPMENT - 10,  65}, WWT_GROUPBOX, 2, STR_LAST_DEVELOPMENT                                          ),
+    MakeWidget({265, 161}, {                 24,  24}, WWT_FLATBTN,  2, 0xFFFFFFFF,                   STR_RESEARCH_SHOW_DETAILS_TIP   ),
```
tupaschoal, line 70:
```suggestion
    MakeWidget({  0,  43}, {     WW_DEVELOPMENT, 153}, WWT_RESIZE,   1                                                                ),
    MakeTab   ({  3,  17},                                                                            STR_RESEARCH_AND_DEVELOPMENT_TIP),
    MakeTab   ({ 34,  17},                                                                            STR_FINANCES_RESEARCH_TIP       ),
    MakeWidget({  3,  47}, {WW_DEVELOPMENT - 10,  70}, WWT_GROUPBOX, 2, STR_CURRENTLY_IN_DEVELOPMENT                                  ),
    MakeWidget({  3, 124}, {WW_DEVELOPMENT - 10,  65}, WWT_GROUPBOX, 2, STR_LAST_DEVELOPMENT                                          ),
    MakeWidget({265, 161}, {                 24,  24}, WWT_FLATBTN,  2, 0xFFFFFFFF,                   STR_RESEARCH_SHOW_DETAILS_TIP   ),
```

src/openrct2-ui/windows/Research.cpp
```diff
@@ -62,31 +62,31 @@ enum {
...
-    MakeWidget     ({  8, 147}, {WW_FUNDING - 16,  12}, WWT_CHECKBOX, 2, STR_RESEARCH_NEW_THRILL_RIDES,        STR_RESEARCH_NEW_THRILL_RIDES_TIP           ),
-    MakeWidget     ({  8, 160}, {WW_FUNDING - 16,  12}, WWT_CHECKBOX, 2, STR_RESEARCH_NEW_WATER_RIDES,         STR_RESEARCH_NEW_WATER_RIDES_TIP            ),
-    MakeWidget     ({  8, 173}, {WW_FUNDING - 16,  12}, WWT_CHECKBOX, 2, STR_RESEARCH_NEW_SHOPS_AND_STALLS,    STR_RESEARCH_NEW_SHOPS_AND_STALLS_TIP       ),
-    MakeWidget     ({  8, 186}, {WW_FUNDING - 16,  12}, WWT_CHECKBOX, 2, STR_RESEARCH_NEW_SCENERY_AND_THEMING, STR_RESEARCH_NEW_SCENERY_AND_THEMING_TIP    ),
+    MakeWidget({  0,  43}, {     WW_FUNDING, 164}, WWT_RESIZE,   1),
+    MakeTab   ({  3,  17}, STR_RESEARCH_AND_DEVELOPMENT_TIP       ),
+    MakeTab   ({ 34,  17}, STR_FINANCES_RESEARCH_TIP              ),
+    MakeWidget({  3,  47}, { WW_FUNDING - 6,  45}, WWT_GROUPBOX, 2, STR_RESEARCH_FUNDING_                                                             ),
+    MakeWidget({  8,  59}, {            160,  14}, WWT_DROPDOWN, 2, 0xFFFFFFFF,                           STR_SELECT_LEVEL_OF_RESEARCH_AND_DEVELOPMENT),
```
tupaschoal, line 80:
```suggestion
    MakeWidget({  0,  43}, {     WW_FUNDING, 164}, WWT_RESIZE,   1                                                                                    ),
    MakeTab   ({  3,  17},                                                                                STR_RESEARCH_AND_DEVELOPMENT_TIP            ),
    MakeTab   ({ 34,  17},                                                                                STR_FINANCES_RESEARCH_TIP                   ),
    MakeWidget({  3,  47}, { WW_FUNDING - 6,  45}, WWT_GROUPBOX, 2, STR_RESEARCH_FUNDING_     
    MakeWidget({  8,  59}, {            160,  14}, WWT_DROPDOWN, 2, 0xFFFFFFFF,                           STR_SELECT_LEVEL_OF_RESEARCH_AND_DEVELOPMENT),
```

src/openrct2/sprites.h
```diff
@@ -787,24 +787,23 @@ enum
...
+    SPR_G2_BM_BOOSTER_NW_SE = SPR_G2_BEGIN + 82,
 
-    SPR_G2_TOOLBAR_MUTE = SPR_G2_BEGIN + 86,
-    SPR_G2_TOOLBAR_MUTE_PRESSED = SPR_G2_BEGIN + 87,
-    SPR_G2_TOOLBAR_UNMUTE = SPR_G2_BEGIN + 88,
-    SPR_G2_TOOLBAR_UNMUTE_PRESSED = SPR_G2_BEGIN + 89,
+    SPR_G2_TOOLBAR_MUTE = SPR_G2_BEGIN + 83,
+    SPR_G2_TOOLBAR_MUTE_PRESSED = SPR_G2_BEGIN + 86,
+    SPR_G2_TOOLBAR_UNMUTE = SPR_G2_BEGIN + 87,
+    SPR_G2_TOOLBAR_UNMUTE_PRESSED = SPR_G2_BEGIN + 88,
```
tupaschoal, line 799: Either this should start at 85 (if you're just decrementing 1 from everyone after the disabled sprite) or all of the lines should then follow the 83 and be 84, 85, 86 (not skipping some numbers)

src/openrct2/sprites.h
```diff
@@ -787,24 +787,23 @@ enum
...
+    SPR_G2_GIGA_RC_BOOSTER_NE_SW = SPR_G2_BEGIN + 79,
+    SPR_G2_GIGA_RC_BOOSTER_NW_SE = SPR_G2_BEGIN + 80,
+    SPR_G2_BM_BOOSTER_NE_SW = SPR_G2_BEGIN + 81,
+    SPR_G2_BM_BOOSTER_NW_SE = SPR_G2_BEGIN + 82,
 
-    SPR_G2_TOOLBAR_MUTE = SPR_G2_BEGIN + 86,
-    SPR_G2_TOOLBAR_MUTE_PRESSED = SPR_G2_BEGIN + 87,
-    SPR_G2_TOOLBAR_UNMUTE = SPR_G2_BEGIN + 88,
-    SPR_G2_TOOLBAR_UNMUTE_PRESSED = SPR_G2_BEGIN + 89,
+    SPR_G2_TOOLBAR_MUTE = SPR_G2_BEGIN + 83,
```
AaronVanGeffen, line 796:
```suggestion
    SPR_G2_TOOLBAR_MUTE = SPR_G2_BEGIN + 85,
```

src/openrct2/sprites.h
AaronVanGeffen: Good catch, thanks! I wonder what went wrong there, haha. (Also, if the booster pieces for the junior roller coaster are unused, why are they still compiled into `g2.dat`…?)

tupaschoal: I really don't know :P

Gymnasiast: The booster pieces for the Junior RC are most definitely used. I'm curious as to why it appears they aren't.

Gymnasiast (changes requested):
You broke the booster graphics on the Junior RC.

afbeelding

AaronVanGeffen:
> This breaks the booster graphics on the Junior RC.

This will be fixed by patching the following two lines:

https://github.com/OpenRCT2/OpenRCT2/blob/02d1c5afcab8eb1c18253bc5969794d3f3b23b33/src/openrct2/ride/coaster/JuniorRollerCoaster.cpp#L638-L639

I really wish they were defined in `sprites.h` instead. But if we want to go down that route, that's a refactor for another day.

Gymnasiast (approved):

PR #12673, opened 2020-08-13
Fix #12401: Refactor SCALE_QUALITY to use strong enum

### LINKED ISSUE #12401: Refactor SCALE_QUALITY to use strong enum

tombomp:
Moved to a new enum class ScaleQuality.

Closes #12401

Tested and everything seems to work the same as before.

### DIFF
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -555,7 +555,7 @@ public:
         {
             scaleQuality = ScaleQuality::Linear;
         }
-        snprintf(scaleQualityBuffer, sizeof(scaleQualityBuffer), "%u", scaleQuality);
+        snprintf(scaleQualityBuffer, sizeof(scaleQualityBuffer), "%u", static_cast<int32_t>(_scaleQuality));
         SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, scaleQualityBuffer);
 
         int32_t width, height;
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -551,11 +551,11 @@ public:
         }
 
         ScaleQuality scaleQuality = _scaleQuality;
-        if (_scaleQuality == ScaleQuality::SmoothNN)
+        if (_scaleQuality == ScaleQuality::SmoothNearestNeighbour)
         {
             scaleQuality = ScaleQuality::Linear;
         }
-        snprintf(scaleQualityBuffer, sizeof(scaleQualityBuffer), "%u", static_cast<int32_t>(_scaleQuality));
+        snprintf(scaleQualityBuffer, sizeof(scaleQualityBuffer), "%u", static_cast<int32_t>(scaleQuality));
         SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, scaleQualityBuffer);
 
         int32_t width, height;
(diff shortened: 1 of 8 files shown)


src/openrct2/config/Config.h
```diff
@@ -214,11 +215,11 @@ enum class TemperatureUnit : int32_t
     Fahrenheit
 };
 
-enum SCALE_QUALITY
+enum class ScaleQuality : int32_t
 {
-    SCALE_QUALITY_NN,
-    SCALE_QUALITY_LINEAR,
-    SCALE_QUALITY_SMOOTH_NN
+    NN, //Nearest Neighbour
```
mwnciau, line 220: Comments should be on the line before, not on the same line as code.

Any reason this couldn't just be NearestNeighbour instead of NN? A bit long, but a lot clearer.

tombomp: You're right, that's a way better idea, was just coping the previous enum without thinking

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -465,11 +465,11 @@ class OpenGLDrawingEngine final : public IDrawingEngine
             delete _smoothScaleFramebuffer;
             _smoothScaleFramebuffer = nullptr;
         }
-        if (GetContext()->GetUiContext()->GetScaleQuality() > 0)
+        if (GetContext()->GetUiContext()->GetScaleQuality() != ScaleQuality::NearestNeighbour)
         {
             _scaleFramebuffer = new OpenGLFramebuffer(_width, _height, false, false);
         }
-        if (GetContext()->GetUiContext()->GetScaleQuality() == SCALE_QUALITY_SMOOTH_NN)
+        if (GetContext()->GetUiContext()->GetScaleQuality() == ScaleQuality::SmoothNN)
```
ZehMatt, line 472: You can change that to a simple else statement here, both conditions can not apply at the same time.

tombomp: Not sure that's right, first if is true if ScaleQuality is either SmoothNN or Linear, 2nd is true only if ScaleQuality is SmoothNN. The variable they change is different each time too. Or maybe I'm not thinking straight if so sorry

ZehMatt: Actually my bad, you are correct.

src/openrct2-ui/UiContext.cpp
```diff
@@ -547,15 +547,15 @@ class UiContext final : public IUiContext
...
-        int32_t scaleQuality = _scaleQuality;
-        if (_scaleQuality == SCALE_QUALITY_SMOOTH_NN)
+        ScaleQuality scaleQuality = _scaleQuality;
+        if (_scaleQuality == ScaleQuality::SmoothNN)
         {
-            scaleQuality = SCALE_QUALITY_LINEAR;
+            scaleQuality = ScaleQuality::Linear;
         }
-        snprintf(scaleQualityBuffer, sizeof(scaleQualityBuffer), "%u", scaleQuality);
+        snprintf(scaleQualityBuffer, sizeof(scaleQualityBuffer), "%u", static_cast<int32_t>(_scaleQuality));
```
tupaschoal, line 558:
```suggestion
        snprintf(scaleQualityBuffer, sizeof(scaleQualityBuffer), "%u", static_cast<int32_t>(scaleQuality));
```

This is `scaleQuality`

PR #12676, opened 2020-08-14
Fix #7037: saving rides starting with sloped turn

### LINKED ISSUE #7037: Cannot save ride design if first piece is an upwards or downwards helix.

mwnciau:
Fixed a bug in the save track code that would prevent a rollercoaster that started with a banked, sloped turn from being saved.

The `ride_get_start_of_track` function retrieves the coordinates and track element of the first piece of track. The coordinates it retrieves are of the first section of the first piece of track, but the z-location of the element it retrieves is for the last section of the first piece of track. This only affects a very small subset of track pieces: those with multiple sections and the starting z-position is different to the ending z-position - so essentially helixes and banked, sloped turns.

The `TrackDesign::CreateTrackDesignTrack` method passes the coordinates and z-position to `sub_6C683D` which finds the track position at those coordinates. Because the z-position was off, it couldn't find the track piece and so exits gracefully.

A lot of the above is an educated guess as I don't fully understand `how ride_get_start_of_track` works, but given how specific the z-position offsets needed to fix this are, I'm fairly confident that it's right.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,5 +1,6 @@
 0.3.0+ (in development)
 ------------------------------------------------------------------------
+- Fix: [#7037] Unable to save tracks starting with a sloped turn or helix.
 - Fix: [#12691] Ride graph tooltip incorrectly used count instead of number string.
 - Fix: [#12694] Crash when switching ride types with construction window open.
 - Fix: [#12701] Silent NSIS setup flag /S isn't silent, upgrade pop-up appears anyway.
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -5366,9 +5366,12 @@ void ride_get_start_of_track(CoordsXYE* output)
         bool moveSlowIt = true;
         do
         {
+            // Because we are working backwards, begin_element is the section at the end of a piece of track, whereas
+            // begin_x and begin_y are the coordinates at the start of a piece of track, so we need to pass end_x and
+            // end_y
             CoordsXYE lastGood = {
-                /* .x = */ trackBeginEnd.begin_x,
-                /* .y = */ trackBeginEnd.begin_y,
+                /* .x = */ trackBeginEnd.end_x,
+                /* .y = */ trackBeginEnd.end_y,
                 /* .element = */ trackBeginEnd.begin_element,
             };
 


src/openrct2/ride/Track.cpp
```diff
@@ -589,7 +589,7 @@ int32_t track_is_connected_by_shape(TileElement* a, TileElement* b)
     return aBank == bBank && aAngle == bAngle;
 }
 
-const rct_preview_track* get_track_def_from_ride(Ride* ride, int32_t trackType)
+const rct_preview_track* get_track_def_from_ride(const Ride* ride, int32_t trackType)
```
Gymnasiast, line 592: Not related to this PR, but certainly welcome as another PR. (Even better if that other PR makes a member function out of it.)

src/openrct2/ride/TrackData.cpp
```diff
@@ -1638,6 +1639,7 @@ static constexpr const rct_preview_track TrackBlocks086[] = {
     TRACK_BLOCK_END
 };
 
+// Helix Up Left Small
```
Gymnasiast, line 1642: Rather than adding comments, it is better to rename these blocks to something meaningful.

duncanspumpkin: But again leave for a different pr

src/openrct2/ride/Track.cpp
mwnciau: This change in particular was needed because the ride pointer in TrackDesign::CreateTrackDesignTrack was const, so this function needed to accept a const.

src/openrct2/ride/TrackData.cpp
mwnciau: That's a good point. I'm planning to do some refactoring of the Track code, starting with documentation - would you prefer I remove these comments from this PR and do a PR later that renames these blocks?

Gymnasiast: I would, yes.

src/openrct2/ride/Track.cpp
Gymnasiast: You can leave it like this, then.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -187,6 +187,16 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
     auto trackType = trackElement.element->AsTrack()->GetTrackType();
     uint8_t direction = trackElement.element->GetDirection();
     _saveDirection = direction;
+
+    // ride_get_start_of_track returns the x and y of the first section of track, but the z location
+    // of the last section so we correct it by adding the difference
+    const rct_preview_track* trackSection = get_track_def_from_ride(&ride, trackType);
+    int32_t firstSectionZ = trackSection->z;
+    trackSection += trackElement.element->AsTrack()->GetSequenceIndex();
```
Gymnasiast, line 195: I'd like @duncanspumpkin 's opinion on this block. I really can't say if it is correct or not.

duncanspumpkin: It is correct but i want to do a few tests of this pr before merging

mwnciau: If it helps, here's a save file where I've spammed rollercoasters with different starting pieces to test them: [PR12676.zip](https://github.com/OpenRCT2/OpenRCT2/files/5077572/PR12676.zip)

mwnciau: I think I've figured out exactly why it's broken. It's to do with how `track_block_get_previous` / `track_block_get_previous_from_zero` work. They use a struct to store their result that contains the 5 relevent fields: begin_x, begin_y, end_x, end_y, begin_element. However, because `track_block_get_previous` traverses backwards, the begin_element label is misleading: when the algorithm looks for the previous track piece, it looks for a track section in location begin_x and begin_y. The piece it finds is the last section of the previous piece of track, not the first. `ride_get_start_of_track` would then return the coordinates for the first track section, but the element for the last track section. See the below picture if that's confusing...

this picture

I've completely reworked this PR to just fix `ride_get_start_of_track` to return the coordinates for the last section instead of the coordinates for the first section (begin_element along with end_x and end_y). Both places where this function is used, only the first track piece is needed, not the first section of track, so this isn't a breaking change (the other place is for checking the refund amount for rides, which I also tested).

duncanspumpkin:
Yeah the new change looks good. I wasn't a huge fan of the old one it looked like a cludge. Annoyingly we can't easily modify the forward/backward get functions without very very careful thinking of the implications. There are a large number of hacks that people exploit that are based on the unique interactions of the forward and backward functions which we definitely do not want to break. As this change does not change those functions it looks safe.

Gymnasiast:
I would like to test this.

tupaschoal:
@Gymnasiast don't forget to test this

Gymnasiast (approved):

PR #12694, opened 2020-08-15
Fix crash when switching ride types with construct window open

duncanspumpkin:


tupaschoal:
Should this be on the changelog @duncanspumpkin ?

duncanspumpkin:
yeah probably. I'm thinking that maybe this isn't the best fix. It stops the crash but its not great.

tupaschoal:
Should we hold it for now?

duncanspumpkin:
I think it's probably fine

PR #12695, opened 2020-08-15
Remove full stop from localisation workflow commit message

AaronVanGeffen:
The OpenRCT2 commit guidelines actively discourage people from using a full stop in their commit messages. Let's apply the same principle to our OpenRCT2 Localisation Bot. :wink:


Gymnasiast:
Good catch! For some reason, I had never noticed.

PR #12697, opened 2020-08-15
Add more drawable parts for Flying coaster

gamrguy:
Imports a large selection of parts from the Steel Twister coaster, as the track types match.
Also provides inversible corkscrews and inversible quarter loops using the `BolligerMabillardTrack` functions.

Things to note:
- The newly drawable parts are available when using the "Enable all drawable track pieces" cheat.
- There is no available steep-to-vertical transition for the inverted track, so the inverted track cannot go vertical.
- When using the uninverted track, an invisible downwards Quarter Loop is available. Don't use it, there are no sprites for it or the vertical inverted track it would be connecting to.
- Barrel Rolls and Large Half Loops are a bit buggy; ideally, these would be replaced with inversible versions when the new save format is finished. They can still be used properly if you switch to something like the Steel Twister to build with them.
- The Flying Coaster trains are incapable of going through some of the inversions properly and will show incorrect animation frames.


duncanspumpkin:
When you say using code do you mean copy paste? Can't you just call the respective functions we are trying to undo duplication not increase it.

gamrguy:
Not quite, while the actual code is the same I needed to change which sprites were being used.

gamrguy:
- Added Vertical Curves to the build menu
- Added Vertical Loops (available on inverted and uninverted track)
- Code for inverted Vertical Loops in this track style relocated to `BolligerMabillardTrack.cpp`, with 3 additional parameters `trackHeightOffset`, `supportHeightOffset`, and `flipInverted`, currently only used for support with the Inverted coaster.
- - `trackHeightOffset` raises/lowers the track piece (added to `height`), Inverted Coaster uses `5`, others `0`
- - `supportHeightOffset` raises/lowers the general support height (added to `height`), Inverted Coaster uses `16`, others `0`
- - `flipInverted` is a bool that indicates whether the track should draw inverted when it isn't inverted and vice versa, Inverted Coaster uses `true`, others `false`

gamrguy:
In that case, perhaps it'd be best if I just close this, fix up my repository, and come back when I have something more presentable.

tupaschoal:
It's alright if you want to keep it open, we can mark the PR as draft, to show that the work is unfinished, but ongoing :)

gamrguy:
Since I'll probably be moving the commits to a new branch, this PR will likely become invalid anyway, but thanks!

PR #12707, opened 2020-08-16
Fix: Incomplete loop collision box allowed overlap (original bug)

mwnciau:
A bug in the track loop collision box allowed for the construction of colliding pairs of loops. This PR fixes the collision box to solve the bug. To reproduce this bug, use the following saved track, [Colliding Loops.zip](https://github.com/OpenRCT2/OpenRCT2/files/5081152/Colliding.Loops.zip), or see [this image](https://media.discordapp.net/attachments/691736850980667514/744557894116048906/Artisan_Valley_2020-08-16_21-06-11.png) (thanks OpenShinx).

The problem was that indices 4 and 5 of the track specification didn't have any quarters specified, so wouldn't block pieces from being created. See the below image, which is a mapping of the collision box of a right loop; the black boxes are the two collision boxes that are added in this PR.

untitled

The collision boxes are only checked/changed when you place the piece, so this won't affect any saved games or existing coasters (but will prevent any saved custom coasters from being built).


AaronVanGeffen:
Is this addressing a recent regression, or has this been in the game for a while?

mwnciau:
It's a long term problem (it existed in 0.1.0), but an edge case so likely has very rarely come up. Those blocks haven't been changed for at least a year.

duncanspumpkin:
Did it exist in vanilla? Might also be another thing that was fixed in the gog version.

mwnciau:
Just started up my Steam version of RCT2 and it existed there, too.

Added a changelog entry now.

duncanspumpkin:
If it's a vanilla bug you should add that to the changelog entry as well.

duncanspumpkin:
You will also need to increment the network version

mwnciau:
Sorted.

distribution/changelog.txt
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#12691] Ride graph tooltip incorrectly used count instead of number string.
 - Fix: [#12694] Crash when switching ride types with construction window open.
+- Fix: [#12707] Incomplete loop collision box allowed overlapping track (original bug).
```
tupaschoal, line 5:
```suggestion
- Fix: Incomplete loop collision box allowed overlapping track (original bug).
```

The number we quote in here is of an issue in the GitHub tracking that got fixed. Since this was not tracked (apparently, I did not search), then no need for the number.

Also, can you reword your commit to be what is here in the changelog? I think it's a lot nicer and more informative:
"Fix: Incomplete loop collision box allowed overlap (original bug)"

tupaschoal:
Also: kudos for the great explanation/documentation on the PR description!

mwnciau:
All done and rebased to the latest on the develop branch.

AaronVanGeffen:
We could also address it with a squash merge, if preferred.

mwnciau:
Sorry, I wasn't quite awake apparently and thought you meant the PR name. The first commit is changed now.

mwnciau:
Done, hopefully for the last time! :)

Gymnasiast:
This has broken some built-in track designs. One example is "Bumble Bee", a Looping Roller Coaster.

PR #12708, opened 2020-08-17
Make all GameAction types accessible to plugins

tombomp:
All GameActions are mapped to strings usable by plugins and there's a new type in the .d.ts file listing them all (ActionType)

By providing an ActionType signature for query/executeAction as well as just a plain string it enables autocomplete in things like Visual Studio Code etc while still allowing strings for custom game actions. Ultimately would want to provide overloads for most game actions because each game action has different arguments so would document it in the file. 

One obvious thing I wasn't sure about was the names of the game action strings - the ones already there were just the game action name in all lower case so I copied it but wasn't sure if it should be in something like C/camelCase or something else and wasn't sure where to get feedback - obviously happy to rename them to anything else and fix any issues here. And if this is a bad idea/wrong approach I totally understand.


AaronVanGeffen:
Assuming this is desirable, should this bump `OPENRCT2_PLUGIN_API_VERSION`?

IntelOrca:
I think this is good and the casing is consistent with what is currently there.

> Ultimately would want to provide overloads for most game actions because each game action has different arguments so would document it in the file.

Yeah, that can be done in follow up PR/s.

duncanspumpkin:
Is there no way that these can be generated from the game action. We already have a string of the game action.

tombomp:
> Is there no way that these can be generated from the game action. We already have a string of the game action.
To be honest I wasn't aware before you said that you could get the string from the action ID - I should have looked first and if I'd seen it I would have done it differently. Mostly I just was following the system already there but obviously if consensus is to change it I'm happy to do it another way. Although I'm not sure if there's a totally "nice" way to do it - using GameActionNameQuery still means filling an array like { GameActionNameQuery<BalloonPressAction> , GAME_COMMAND_BALLOON_PRESS } without changing the system around a lot. Unless there's a different way of getting a string to a name that I'm missing (quite likely!). Sorry for the trouble/missing stuff

IntelOrca (approved):

tupaschoal:
@tombomp can you please bump the plugin version, as requested by Aaron?

tombomp:
Oops sorry, updated the API version now

tupaschoal:
@AaronVanGeffen did you also mean to request a changelog entry?

tombomp:
Ah sorry again for forgetting that haha. Also added myself to contributors because I've had quite a few pull requests approved but kept forgetting to add myself, if that's ok.

duncanspumpkin (approved):

Issue #11555, opened 2020-05-01
Support websocket for plugin


bnnu:
websocket can be used to implement twitch plugins.
(like add money if received donation)

IntelOrca:
Not sure if websocket is the best way to go about this, certainly we will want some way of doing interprocess communication but to reduce security vulnerabilities it will likely be restricted to localhost TCP streams.

I am happy to discuss further in the discord plugin channel.

IntelOrca:
TCP communication is now implemented.

PR #12712, opened 2020-08-17
Plugin: Add API for listening and communicating over TCP

### LINKED ISSUE #11844: Add Plugin API for twitch(maybe other services)
Since #11820, Twitch integration is removed. however, with current API there are no way to create replacement plugin for twitch.(since plugins cannot connect to internet, and there are no twitch api)

If there are some api that I can use for twitch connection(maybe localhost socket that mentioned at #11555, or just http get/post api that can get twitch api) maybe I can make twitch integration plugin.

IntelOrca:
This adds new APIs based on [node.js](https://nodejs.org/api/net.html) for listening and communicating on TCP streams. The same APIs could potentially also be used for unix and named sockets in the future. Only connections to localhost (IPv4 or IPv6) are currently allowed.

Here is some rough sample script for a very basic crude HTTP server and client.
https://gist.github.com/IntelOrca/30d96975146b618b4c74865153142ccf

In practice this could be used to host a local HTTP dashboard for a server, or further outside integration such as Twitch support.

Closes #11844 and #11555.


distribution/scripting.md
```diff
@@ -199,4 +214,4 @@ This is up to you. The OpenRCT2 licence does not enforce any licence requirement
 
 > Is there a good place to distribute my script to other players?
 
-There is currently no official database for this. For now the recommendation is to upload releases of your script on GitHub alongside your source code (if public). Some people like to make a GitHub repository that just consists of a list of content (scripts in this case) which anyone can add to via pull requests.
+The recommendation is to upload releases of your script on GitHub alongside your source code (if public). There is a community driven repository for sharing plugins available at http://openrct2plugins.org/.
```
janisozaur, line 217:
```suggestion
The recommendation is to upload releases of your script on GitHub alongside your source code (if public). There is a community driven repository for sharing plugins available at https://openrct2plugins.org/.
```

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -1127,6 +1134,47 @@ void ScriptEngine::SaveSharedStorage()
...
+#    endif
+
+void ScriptEngine::UpdateSockets()
+{
+#    ifndef DISABLE_NETWORK
+    // Use simple for i loop as Update calls can modify the list
+    for (size_t i = 0; i < _sockets.size(); i++)
+    {
+        _sockets[i]->Update();
+        if (_sockets[i]->IsDisposed())
```
frutiemax, line 1151: i can be potentially set to -1 if _sockets[0]->IsDisposed()==true

oli414:
Just tested it with a Node.js server, works as expected.

src/openrct2/scripting/ScriptEngine.cpp
IntelOrca: I don't think it can.

src/openrct2/scripting/ScNetwork.hpp
```diff
@@ -447,6 +449,38 @@ namespace OpenRCT2::Scripting
...
+        std::shared_ptr<ScSocketServer> createServer()
+        {
+            auto& scriptEngine = GetContext()->GetScriptEngine();
+            auto plugin = scriptEngine.GetExecInfo().GetCurrentPlugin();
+            auto socket = std::make_shared<ScSocketServer>(plugin);
+            scriptEngine.AddSocket(socket);
+            return socket;
+        }
+#    else
+        void createServer()
```
janisozaur, line 462: Is it ok the two method have different signatures? Shouldn't this one return `std::shared_ptr<ScSocketServer> createServe{}`?

IntelOrca: This is conditional compilation though, so it isn't an overload.

src/openrct2/scripting/Duktape.hpp
```diff
@@ -269,6 +275,11 @@ namespace OpenRCT2::Scripting
         return DukValue::take_from_stack(ctx);
     }
 
+    template<> inline DukValue ToDuk(duk_context* ctx, const std::string& value)
```
ZehMatt, line 278: I think its better to have std::string_view as the parameter here otherwise it might construct temporary std::strings when passing const char*

src/openrct2/scripting/ScNetwork.hpp
```diff
@@ -447,6 +449,38 @@ namespace OpenRCT2::Scripting
...
+        {
+            duk_error(_context, DUK_ERR_ERROR, "Networking has been disabled.");
+        }
+#    endif
+
+#    ifndef DISABLE_NETWORK
+        std::shared_ptr<ScSocket> createSocket()
```
ZehMatt, line 469: I think it would be better to call this client rather than socket. A socket can be both client and server.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+        ScSocket(const std::shared_ptr<Plugin>& plugin, std::unique_ptr<ITcpSocket>&& socket)
+            : ScSocketBase(plugin)
+            , _socket(std::move(socket))
+        {
+        }
+
+    private:
+        ScSocket* destroy(const DukValue& error)
```
ZehMatt, line 142: I would just call this close instead of destroy its a bit misleading between dispose/destroy.

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -1127,6 +1134,47 @@ void ScriptEngine::SaveSharedStorage()
...
+            i--;
+        }
+    }
+#    endif
+}
+
+void ScriptEngine::RemoveSockets(const std::shared_ptr<Plugin>& plugin)
+{
+#    ifndef DISABLE_NETWORK
+    for (auto it = _sockets.begin(); it != _sockets.end();)
```
ZehMatt, line 1163: This loop can be replaced with std::remove_if, ```~ScSocketBase``` calls Dispose.

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -393,6 +394,10 @@ void ScriptEngine::Initialise()
     ScVehicle::Register(ctx);
     ScPeep::Register(ctx);
     ScGuest::Register(ctx);
+#    ifndef DISABLE_NETWORK
+    ScSocket::Register(ctx);
```
ZehMatt, line 398: I think ScSocketClient is a better name as its not a general purpose socket.

src/openrct2/scripting/ScriptEngine.cpp
ZehMatt: It will underflow temporarily but I don't think thats an issue here, its a bit odd tho.

src/openrct2/scripting/ScriptEngine.h
```diff
@@ -133,6 +137,9 @@ namespace OpenRCT2::Scripting
         };
 
         std::unordered_map<std::string, CustomActionInfo> _customActions;
+#    ifndef DISABLE_NETWORK
+        std::vector<std::shared_ptr<ScSocketBase>> _sockets;
```
ZehMatt, line 141: Given that the storage is shared_ptr it might be better to use std::list considering iterator invalidation matters for the Update loop and there is no penalty here.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        {
+            auto eventId = GetEventType(eventType);
+            if (eventId != EVENT_NONE)
+            {
+                _eventList.RemoveListener(eventId, callback);
+            }
+            return this;
+        }
+
+        uint32_t GetEventType(const std::string_view& name)
```
ZehMatt, line 438: It is recommended that std::string_view is passed by value, the copy is cheap.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        virtual void Update() = 0;
+
+        virtual void Dispose()
+        {
+        }
+
+        virtual bool IsDisposed() const = 0;
+    };
+
+    class ScSocket : public ScSocketBase
```
ZehMatt, line 115:
```suggestion
    class ScSocketClient final : public ScSocketBase
```

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+            dukglue_register_method(ctx, &ScSocket::setNoDelay, "setNoDelay");
+            dukglue_register_method(ctx, &ScSocket::connect, "connect");
+            dukglue_register_method(ctx, &ScSocket::end, "end");
+            dukglue_register_method(ctx, &ScSocket::write, "write");
+            dukglue_register_method(ctx, &ScSocket::on, "on");
+            dukglue_register_method(ctx, &ScSocket::off, "off");
+        }
+    };
+
+    class ScSocketServer : public ScSocketBase
```
ZehMatt, line 351:
```suggestion
    class ScSocketServer final : public ScSocketBase
```

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+                            break;
+                        case NETWORK_READPACKET_DISCONNECTED:
+                            CloseSocket();
+                            _disposed = true;
+                            break;
+                    }
+                }
+                else
+                {
+                    CloseSocket();
```
ZehMatt, line 322: Could call Dispose here since its the same code.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        void AddListener(uint32_t id, const DukValue& listener)
+        {
+            auto& listeners = GetListenerList(id);
+            listeners.push_back(listener);
+        }
+
+        void RemoveListener(uint32_t id, const DukValue& value)
+        {
+            auto& listeners = GetListenerList(id);
+            auto it = std::find(listeners.begin(), listeners.end(), value);
```
ZehMatt, line 65: Use std::remove

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        }
+    };
+
+    class ScSocketBase
+    {
+    private:
+        std::shared_ptr<Plugin> _plugin;
+
+    protected:
+        static bool IsLocalhostAddress(const std::string_view& s)
```
ZehMatt, line 85: Pass by value

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+        const std::shared_ptr<Plugin>& GetPlugin() const
+        {
+            return _plugin;
+        }
+
+        virtual void Update() = 0;
+
+        virtual void Dispose()
```
ZehMatt, line 108: It should probably be pure virtual as well.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        bool write(const std::string& data)
+        {
+            if (_disposed)
+            {
+                auto ctx = GetContext()->GetScriptEngine().GetContext();
+                duk_error(ctx, DUK_ERR_ERROR, "Socket is disposed.");
+            }
+            else if (_socket != nullptr)
+            {
+                _socket->SendData(data.c_str(), data.size());
```
ZehMatt, line 217: This send might be not successful if the remote host closed the connection, this should be handled.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        ScSocket* end(const DukValue& data)
+        {
+            if (_disposed)
+            {
+                auto ctx = GetContext()->GetScriptEngine().GetContext();
+                duk_error(ctx, DUK_ERR_ERROR, "Socket is disposed.");
+            }
+            else if (_socket != nullptr)
+            {
+                if (data.type() == DukValue::Type::STRING)
```
ZehMatt, line 199: Passing any other type will have no-op, this should be made clear by an error or support for those types.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+                    duk_error(ctx, DUK_ERR_ERROR, "Server is already listening.");
+                }
+                else
+                {
+                    if (dukHost.type() == DukValue::Type::STRING)
+                    {
+                        auto host = dukHost.as_string();
+                        if (IsLocalhostAddress(host))
+                        {
+                            _socket->Listen(host, port);
```
ZehMatt, line 402: This function might throw, this should be handled.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+            if (_socket != nullptr)
+            {
+                return _socket->GetStatus() == SOCKET_STATUS_LISTENING;
+            }
+            return false;
+        }
+
+        ScSocketServer* close()
+        {
+            Dispose();
```
ZehMatt, line 373: There is the ambiguity of close vs dispose here again, closing a socket should not destroy it as one might use it again.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+                    auto ctx = scriptEngine.GetContext();
+                    auto dukClientSocket = GetObjectAsDukValue(ctx, clientSocket);
+                    _eventList.Raise(EVENT_CONNECTION, GetPlugin(), { dukClientSocket }, false);
+                }
+            }
+        }
+
+        void Dispose() override
+        {
+            if (_socket != nullptr)
```
ZehMatt, line 480: This could call close once close stops calling Dispose.

src/openrct2/scripting/ScSocketServer.hpp
```diff
@@ -0,0 +1,505 @@
...
+        ScSocket(const std::shared_ptr<Plugin>& plugin, std::unique_ptr<ITcpSocket>&& socket)
+            : ScSocketBase(plugin)
+            , _socket(std::move(socket))
+        {
+        }
+
+    private:
+        ScSocket* destroy(const DukValue& error)
+        {
+            if (_socket != nullptr)
```
ZehMatt, line 144: I think that when having an ScSocket instance it should imply the underlying instance also always exists. Its probably best to construct the socket in ctor and only delete it in dtor which eliminates all those nullptr checks.

src/openrct2/scripting/ScriptEngine.cpp
IntelOrca: This is a pretty common way of removing items while iterating through. Even if you use an iterator (which I can't because update modifies the collection), you still typically subtract 1 to counter act the loop modifier.

src/openrct2/scripting/ScNetwork.hpp
IntelOrca: Well node.js has Socket and Server, hence my Socket and Server JS classes, so I was more or less trying to match that with these creation methods.

src/openrct2/scripting/ScSocketServer.hpp
IntelOrca: This is mirroring the node.js API, so unfortunately not something I'd like to change.

src/openrct2/scripting/Duktape.hpp
IntelOrca: I can't remove this as it causes compile errors, there is already another overload for string_view.

ZehMatt: I see, then it should be fine already.

IntelOrca:
@oli414 I have changed the API slightly. You now need to call `createListener` instead of `createServer`.

IntelOrca:
do not merge yet

src/openrct2/network/Socket.cpp
```diff
@@ -504,6 +504,11 @@ class TcpSocket final : public ITcpSocket, protected Socket
             throw std::runtime_error("Socket not closed.");
         }
 
+        // The status is  When connect is called, status is set to resolving but we want to make sure
+        // the status is changed before this async method exits otherwise the consumer
+        // might think the status has closed before it started to connect
```
Broxzier, line 509:
```suggestion
        // When connect is called, the status is set to resolving, but we want to make sure
        // the status is changed before this async method exits, otherwise the consumer
        // might think the status has closed before it started to connect.
```

AaronVanGeffen: Agreed with @Broxzier, although I think "otherwise" should start its own sentence:
```suggestion
        // When connect is called, the status is set to resolving, but we want to make sure
        // the status is changed before this async method exits. Otherwise, the consumer
        // might think the status has closed before it started to connect.
```

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -1127,6 +1134,44 @@ void ScriptEngine::SaveSharedStorage()
...
+    }
+#    endif
+}
+
+void ScriptEngine::RemoveSockets(const std::shared_ptr<Plugin>& plugin)
+{
+#    ifndef DISABLE_NETWORK
+    _sockets.erase(
+        std::remove_if(
+            _sockets.begin(), _sockets.end(), [&plugin](const auto& socket) { return socket->GetPlugin() == plugin; }),
```
terinjokes, line 1170: It doesn't look like this condition will match, so sockets are never closed.

IntelOrca: I tested this method and it works.

terinjokes: When I leave a game and return to the main menu listening sockets aren't closed.

IntelOrca: For some reason something, probably dukglue / duktape is holding on to the ScSocket reference. I have therefore changed back to explicitly closing the socket when the plugin is stopped.

terinjokes: I'll checkout this commit and test it out. Thanks for the quick response.

PR #12714, opened 2020-08-17
Remove duplicate TRACK_ELEMENT_LOCATION_IS_UNDERGROUND flag

mwnciau:
This flag is used in the same context as ELEMENT_IS_UNDERGROUND flag and has the same value so I'm fairly sure it's a duplicate. See, for example, this excerpt from TrackPlaceAction.hpp (lines 272-284):

```
uint8_t mapGroundFlags = gMapGroundFlags & (ELEMENT_IS_ABOVE_GROUND | ELEMENT_IS_UNDERGROUND);
...
res->GroundFlags = mapGroundFlags;
...
if (res->GroundFlags & TRACK_ELEMENT_LOCATION_IS_UNDERGROUND)
```


duncanspumpkin:
Looks fine and sensible to me

PR #12722, opened 2020-08-19
Create WindowColour enum for widget definition calls

tupaschoal:
This will have a crazy amount of merges with #12671 , but anyway.
Turns out that there was a colour `3`, so I created a `quaternary` value.
I have split into different commits just for the sake of review, but this should eventually be squash merged.


tupaschoal:
Now that your PR has been merged and that I've solved the merge conflicts, I think this is ready for review @AaronVanGeffen

AaronVanGeffen (approved):
This looks like a nice change for readability. Admittedly, I wish C++ would support leaving out the `WindowColour::` prefix. It is a little more verbose than I'd like in that respect.

Anyway, huge improvement over just having hardcoded numbers. Go for it.

PR #12366, opened 2020-07-23
Close #12336: Remove goto from loc_6A6D7E

frutiemax:

PR #12730, opened 2020-08-19
Fix #12728: Paths made in scenario editor don't connect to the map edge

### LINKED ISSUE #12728: Paths made in scenario editor don't connect to the map edge
**OS:**  Windows 10
**Version:** develop
**Commit/Build:** 8d141fa

In the scenario editor drawing a path that should connect to the map edge no longer creates a connection. This means you can't make a valid peep spawn. It works fine in 0.3 so although I haven't bisected I'd assume it was introduced in #12366

**Steps to reproduce:**
1. Open scenario editor
2. Press button for landscape editor
3. Create a path on an edge square - it doesn't connect

**Screenshots / Video:**
image

mwnciau:
Fix bug introduced in #12366 that prevented making paths to the edge of the map in the scenario editor. A function call was missed in the original refactor.

### DIFF
--- a/src/openrct2/world/Footpath.cpp
+++ b/src/openrct2/world/Footpath.cpp
@@ -837,6 +837,7 @@ static void loc_6A6D7E(
         {
             neighbour_list_push(neighbourList, 7, direction, 255, 255);
         }
+        loc_6A6FD2(initialTileElementPos, direction, initialTileElement, query);
     }
     else
     {


janisozaur:
Would you be able to come with a test for this regression?

tupaschoal:
Thanks for the quick fix, the `else` was so big I missed the call should also happen on the `if`.

I've merged it to fix the regression, but do see if you can come up with a test, please

PR #12731, opened 2020-08-19
Add more drawable parts to Flying Roller Coaster

gamrguy:
(Repost as I shifted the commits over to a new branch)

Adds a large selection of parts that can be used from the "Enable all drawable track pieces" cheat.
Also refactors `BolligerMabillardTrack`-style Vertical Loops to use a standardized function, which supports the inverted loops and the track style adjustments the Inverted Roller Coaster uses.

Parts for uninverted track:
- Long steep-to-flat transitions
- Half barrel roll
- Large half loop
- Steep lift hill
- Launched lift hill
- Holding brake for drop
- Vertical slope
- Vertical curve
- Half corkscrew
- Quarter loop
- Vertical loop

Parts for inverted track:
- Half corkscrew
- Quarter loop
- Booster
- Vertical loop

Things to note:
- There is no available steep-to-vertical transition for the inverted track, so the inverted track cannot go vertical.
- When using the uninverted track, an invisible downwards Quarter Loop is available. Don't use it, there are no sprites for it or the vertical inverted track it would be connecting to.
- Barrel Rolls and Large Half Loops are a bit buggy; ideally, these would be replaced with inversible versions when the new save format is finished. They can be used normally if you switch to something like the Steel Twister to build with them, but still cannot be used to transition to inverted track.
- The Flying Coaster trains are incapable of going through some of the inversions properly and will show incorrect animation frames.


Gymnasiast (commented):
Hm, I'm not sure sure how to properly review this. It is a very big diff and it is hard to see if no drawing function is broken. I would prefer having separate commits per track piece.

Fred-1044276:
This could also use an updated changelog file. A line like the one below would work:

`- Feature: [#12731] Add more drawable track pieces to the Flying Roller Coaster (enable the "Show all track pieces" cheat).`

AaronVanGeffen:
I agree the changelog should be updated, but probably as a `Change` rather than a `Feature`:

`- Change: [#12731] Add more drawable track pieces to the Flying Roller Coaster.`

The `(enable the "Show all track pieces" cheat)` bit, while helpful, is not in keeping with our changelog style, I think.

Gymnasiast:
As I said on 8 September, I cannot review this properly, as it hasn't been properly split into manageable commits. For this reason, unfortunately, I'll have to close it.

If you still feel like working on this further, please come to our Discord to work out the best approach.

PR #12736, opened 2020-08-20
Add regression test for #12728: map edge footpath connections in scenario editor

mwnciau:
Regression test for bug #12728. The test runs the `footpath_connect_edges` function on a footpath tile starting with no connections on the edge of the map after enabling the scenario editor and checks which edges are flagged. See also PR #12730.

Before #12730, 3 of 4 expects fail; after the fix, all four pass.

This is the closest function that was accessible outside of Footpath.cpp, but it does the job. I wasn't sure whether to tag this onto TileElementWantsFootpathConnection or create a new testing class, but the setup overlaps a lot, and the name is at least somewhat relevant so I went for it.


janisozaur (approved):
Perfect! The CI must be complaining about the commit message being a bit too long, can you shorten it a bit (70 chars)?

test/tests/TileElements.cpp
```diff
@@ -140,3 +147,28 @@ TEST_F(TileElementWantsFootpathConnection, DifferentHeight)
...
+    // Enable the scenario editor flag
+    gScreenFlags |= SCREEN_FLAGS_SCENARIO_EDITOR;
+
+    // Calculate the connected edges
+    footpath_connect_edges({ 16, 64 }, pathElement, 0);
+    auto edges = pathElement->AsPath()->GetEdges();
+
+    // The tiles alongside in the Y direction are both on the map edge so should be marked as an edge
+    EXPECT_TRUE(edges & (1 << 1));
+    EXPECT_TRUE(edges & (1 << 3));
```
mwnciau, line 165: Thought about this a bit, and I'm not sure this is desired behaviour. The `map_is_edge` call at Footpath.cpp:834 will return true if the coordinates are for a piece touching the edge of the map or off the edge of the map. This means a footpath piece will try to connect to empty tiles on the edge of the map, which I don't think is correct. I can't actually think of how this could break anything, though.

PR #12738, opened 2020-08-20
Fix #12737: Space Rings draw the same vehicle 4 times

### LINKED ISSUE #12737: Space Rings draw the same vehicle 4 times
**OS:** Windows 10, Linux (Manjaro)
**Version:** 0.3.0
**Commit/Build:** b51e98215

All 4 vehicles draw the passenger riding in vehicle 0 at the animation frame of vehicle 0

**Steps to reproduce:**
1. Build a Space Rings
2. Watch the people ride

**Dump file**

**Screenshots / Video:**

**Save game:**

CorySanin:
paint_space_rings_structure always used vehicle 0 for it's vehicle reference

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,6 +3,7 @@
 - Fix: [#12691] Ride graph tooltip incorrectly used count instead of number string.
 - Fix: [#12694] Crash when switching ride types with construction window open.
 - Fix: [#12701] Silent NSIS setup flag /S isn't silent, upgrade pop-up appears anyway.
+- Fix: [#12737] Space Rings draw the same vehicle 4 times.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
 
 0.3.0 (2020-08-15)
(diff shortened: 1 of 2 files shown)


CorySanin:
Here ya go!

Peek-2020-08-19-23-17_text

duncanspumpkin:
Probably my mistake in the entity refactor

CorySanin:
Actually, it should probably use `vehicleIndex` instead of `segment`, right? Let me fix that.

duncanspumpkin:
Yes vehicle index should be used. Segment index will get the wrong vehicle depending on rotation.

CorySanin:
Good thing we caught that 😉
Rotating around the ride behaves how I expect now. Looks good to me.

duncanspumpkin:
You'll need to add a changelog entry as this bug will have been in the release

distribution/changelog.txt
```diff
@@ -4,6 +4,7 @@
 - Fix: [#12694] Crash when switching ride types with construction window open.
 - Fix: [#12701] Silent NSIS setup flag /S isn't silent, upgrade pop-up appears anyway.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
+- Fix: [#12737] Space Rings draw the same vehicle 4 times.
```
AaronVanGeffen, line 7: Sorry to be the bearer of bad news, but we keep the numbered fixes above the unnumbered ones.

CorySanin: Oh, alrighty. Not a problem

CorySanin: Oh shoot, didn't see that you also fixed it 😬

PR #12749, opened 2020-08-22
Downgrade Mac build requirements of OpenRCT2 to macOS 10.13

awiebe:
I have downgraded the build target of the Xcode project file for the app bundle build target to OS 10.7

This allows people on much earlier versions of the OS to enjoy Open RTC2.🥂  The OS SDK basically only guarantees that some os features will be available.  Since OpenRTC basically uses nothing besides surface drawing, and the save panel, it is completely unnecessary to have it set to 10.14.

Any lower than this and Xcode complains because clang doesn't exist on 10.6.

For some reason even though the Xcode build file is using `--std=c++17`, build fails reporting that 3 const std::array initializers have an un-deducable size, rather than try to figure out why this is, I just named the sizes and types of those arrays explicitly.


YJSoft:
OS X 10.8(which is minimum version with coregraphic)
Result : crash due to missing symbol __ZNSt19bad_optional_accessD1Ev(which is destructor for `bad_optional_access`)

OS X 10.9 ~ macOS 10.13(Installing...)

janisozaur:
See https://github.com/canonical/multipass/issues/1258#issuecomment-580725681

janisozaur:
@YJSoft does it say where's the symbol referenced from?

See also: https://forums.factorio.com/viewtopic.php?p=340158#p340158

IntelOrca:
> Result : crash due to missing symbol __ZNSt19bad_optional_accessD1Ev(which is destructor for bad_optional_access)

That suggests that using new C++ features such as `std::optional` does still link to a newer version of libc++.

YJSoft:
@janisozaur https://gist.github.com/YJSoft/2a98b42a816b9507ccd3a31c5ddd0a86

YJSoft:
image
I tested with all version except 10.9(I don't have installer image), but all version below 10.13(osx 10.7 ~ macos 10.12) failed to launch Generated app file.

AaronVanGeffen:
Thank you very much for testing, @YJSoft. 

Considering our recent survey revealed there are still players that stick with macOS 10.13 and 10.14, I personally think it is worth holding off on the `std::array` type inference if it means retaining support for these OSes.

As I have argued in the team chat before, as long as the maintenance cost is low, we should try to support macOS versions at least while Apple still does. At the moment, that is versions of macOS starting from 10.13. This is likely to change when macOS 11 is released in October, though, but perhaps for v0.3.1 at least?

OpenRCT2.xcodeproj/project.pbxproj
```diff
@@ -4498,6 +4498,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CLANG_ENABLE_OBJC_WEAK = YES;
+				MACOSX_DEPLOYMENT_TARGET = 10.7;
```
AaronVanGeffen, line 4501: As we have established, this is an unrealistic target. Could you change all of these to read 10.13 instead?

(No need to amend the commit, we can squash the lot on merge.)

IntelOrca:
```suggestion
				MACOSX_DEPLOYMENT_TARGET = 10.13;
```

You are us can also accept this suggestion which is a quick way to commit the change.

AaronVanGeffen: It's not just one line, though. I didn't feel like making six suggestions.

duncanspumpkin:
Do we know what other features that this change will prevent us from using. I very much think we should be increasing the xcode version so we can use more c++20 features as and when they are made available by the major compilers.

AaronVanGeffen:
I'm sorry to say this, but I think it's a bit early days to start using C++20 features… :confused:

janisozaur:
I'm with @AaronVanGeffen on this one and it's generally been consensus so far, also in wake of C++17 support. To recap: we can switch to it when it becomes official and supported across enough CI setups (that means if old platforms don't make the cut, then it's fine).

So far I'd prefer we stick to C++17 and if there are ways to support older platforms as well without too much additional maintenance burden, then let's do that.

src/openrct2-ui/interface/Window.cpp
```diff
@@ -212,14 +212,16 @@ rct_window* window_create_auto_pos(
...
-        const std::array offsets = { ScreenCoordsXY{ w->width + 2, 0 },
-                                     ScreenCoordsXY{ -w->width - 2, 0 },
-                                     ScreenCoordsXY{ 0, w->height + 2 },
-                                     ScreenCoordsXY{ 0, -w->height - 2 },
-                                     ScreenCoordsXY{ w->width + 2, -w->height - 2 },
-                                     ScreenCoordsXY{ -w->width - 2, -w->height - 2 },
-                                     ScreenCoordsXY{ w->width + 2, w->height + 2 },
-                                     ScreenCoordsXY{ -w->width - 2, w->height + 2 } };
+
+        const ScreenCoordsXY offsets[] = { { w->width + 2, 0 },
```
AaronVanGeffen, line 216: `std::array` requires each of the elements to be explicitly constructed with ScreenCoordx{}, whereas a regular array does not require this. IMO this makes for more readable code.

If this is not desirable, we can drop the commit.

AaronVanGeffen: Splitting this patch off to another PR. Will open after this PR has been merged.

PR #12757, opened 2020-08-22
Strong enum operators

mwnciau:
See [this comment](https://github.com/OpenRCT2/OpenRCT2/pull/12757#issuecomment-685990463) for latest info.

---

I've been playing around with strong enums ([here](https://github.com/mwnciau/Strong-Enum-Overloading)) to find an efficient and easy way to use strong enums as flags or array indices. This is more of a proof of concept than an actual PR, but I thought it would be easier to show what it might look like in the current codebase. That said, these are fully working changes which compile/run as expected.

There are three commits: adding in a few macros and template classes, an example flag strong enum, and an example array index strong enum. For the latter two, I tried to find examples that were as small as possible.

### Macros

There are five defined macros in `Enum.hpp`:

`ENABLE_ENUM_BITWISE_OPERATIONS` and `ENABLE_ENUM_BITWISE_COMPATIBILITY`, used for flags, respectively enable use of the bitwise operators on the specified enum, and enable use of bitwise operators with a specified other enum or type. E.g.

```
enum class Example : uint32_t;
enum class Example2 : uint32_t;
ENABLE_ENUM_BITWISE_OPERATIONS(Example, uint32_t);
// This now works: Example::A | Example::B
ENABLE_ENUM_BITWISE_COMPATIBILITY(Example, Example2, uint32_t);
// This now works: Example::A | Example2::A
ENABLE_ENUM_BITWISE_COMPATIBILITY(Example, uint32_t, uint32_t);
// This now works: Example::A | 1
// But (Example2::A | Example::A) and (1 | Example::A) still don't work
```

Condition checking can either be done by using `!`, e.g. `if (!!(flag & Example::A))`, or by checking with null, e.g. `if ((flag & Example::A) == Example::Null)`.

`ENABLE_ENUM_ARITHMETIC_OPERATIONS` and `ENABLE_ENUM_ARITHMETIC_COMPATIBILITY`, used for array indices, work exactly as above, but for arithmetic operators, e.g. `+`, `-`, `++`, etc.

`ENABLE_ENUM_PLUS_SHORTCUT` overloads the unary `+` operator to return the underlying value of an enum as a constexpr. I.e. `arr[+Example::Count]` instead of `arr[static_cast<uint32_t>(Example::Count)]`.

### Helper Classes

As well as the macros, `Enum.hpp` also provides to helper template classes, `EnumeratedArray` and `ConstEnumeratedArray`. They work as wrappers on an array and allow strictly typed access to array elements.

```
uint32_t _treeHeights[10];
const EnumeratedArray<uint32_t, TreeType> treeHeights{ _data};
// We can now access elements of _treeHeights using enums
treeHeights[TreeType::Yew];
// Normal access to the array doesn't work
// treeHeights[0];
```

### Pros / Cons

Pros
- Strongly typed, preventing unintentional errors
- Works intuitively
- Aside from the EnumeratedArray classes, just as fast or faster than as using weak enums
- Can explicitly specify when two strong enums should interact

Cons
- Slight performance hit for using the EnumeratedArray classes (although some of it can be optimised away by the compiler)
- I have some reservations about using  `!!` and the unary `+`, but they _do_ do what you'd expect (and they're optional anyway)


mwnciau:
I was trying to find something that is:

- Strongly typed*
- Doesn't have a large performance overhead
- Doesn't require a lot of extra code to use an enum
- Interactions with other enums can be enabled**

\* Errors like [this](https://github.com/OpenRCT2/OpenRCT2/pull/12745/files#diff-e8f2c9f73fff3b5e0e307a5ff9da83fb) would be a lot easier to avoid
** e.g. the TRACK_CURVE_* enums use values 256-511 to overlap with the TRACK_ELEM_* enums, and it'd be useful to allow them to continue interacting without having to use a lot of casts.

Unfortunately, the macro solution is a bit messy, as you say. I also tried a couple of other things:

### Template Class

I created a couple of [template classes](https://github.com/mwnciau/Strong-Enum-Overloading/blob/master/Classes.hpp) that wrap a strong enum.

```
enum class ClassFlagExample : uint32_t { ... };
using ClassFlagExample_t = FlagEnum<ClassFlagExample>;
```

`ClassFlagExample_t` would then be a type that you'd use wherever you would've used `ClassFlagExample`. It works fairly well, but I couldn't figure out a way to merge the enum and the type while both keeping the type safety and not having to redefine all the operations for each enum you create.

This method has a couple of downsides. You need to define a lot more interactions (interactions between `ClassFlagExample_t` with itself and with `ClassFlagExample`), and in that just seems a bit messier. There also isn't a way to easily allow interactions with another enum (it could be done with another class definition, but it starts to get unwieldy).

### Include file

I came across a suggestion to use a [reusable header file](https://github.com/mwnciau/Strong-Enum-Overloading/blob/master/Enum.hxx), and use preprocessor variables instead of macro parameters to specify which enum to add the overloaded operators to. This way, they are actually part of the code instead hidden in macros, and any warnings or errors appear in the header file in the appropriate place rather than nebulously at a macro call. You can also add breakpoints and step through it, too. It's a bit unconventional, but I actually quite like how this turned out:

```
enum class FlagExample : uint32_t { ... };
#define ENABLE_ENUM_FLAG FlagExample
#define ENABLE_ENUM_FLAG_OTHER FlagExample2
// This uses the above variables to create all the operator overloads, then #undef's them so it can easily be reused
#include "Enum.hxx"

enum class IndexExample : uint32_t { ... };
#define ENABLE_ENUM_INDEX IndexExample
// Enum.hxx doesn't have #pragma once so it can be included multiple times
#include "Enum.hxx"
```

tupaschoal:
By the way, you might also want to read about [magic enum](https://github.com/Neargye/magic_enum) which is also header-only, but a lot more elegant

mwnciau:
One last push to this PR. I've implemented and pushed to this PR a class-based replacement for strong enums that provides all the functionality I was looking for without the use of macros. See the last section of this comment for more details on that.

I've done a general review of the options I considered below (sorry for the wall of text), and how they fit against the criteria I had in mind. Happy to write this off as wishful thinking if no consensus can be made, though - I was just interested in whether or not this sort of thing was possible.

## Criteria
1. Strongly typed
2. Allows bitwise operators
3. Allows arithmetic operators
4. Efficient
5. Easy to use
6. Optionally allow interaction with other enums

## Solutions

### namespace + constexpr

1. Strongly typed ❌
2. Allows bitwise operators ✔️
3. Allows arithmetic operators ✔️
4. Efficient ✔️
5. Easy to use ✔️
6. Optionally allow interaction with other enums ➖

**Examples**
```
namespace Index
{
    constexpr uint32_t Value1;
    ...
};
```

**Comments**
* These are just typed as their underlying type so there is no type safety. This does mean that they interact with other enums out of the box, but there are no restrictions on what is allowed to interact.


### Magic Enum
Magic Enum is a header only library that extends the functionality of C++ strong enums without any macro or boilerplate code.

1. Strongly typed ✔️
2. Allows bitwise operators ✔️
3. Allows arithmetic operators ❌
4. Efficient ✔️
5. Easy to use ✔️
6. Optionally allow interaction with other enums ❌

**Examples**
```
enum class Flag : uint32_t { ... };
enum class NotAFlag : uint32_t { ... };

// Enables bitwise operators for *all* enums
using namespace magic_enum::bitwise_operators;

Flag flag = Flag::Value1 | Flag::Value2;

// This is allowed but doesn't necessarily make sense
NotAFlag notAFlag = NotAFlag::Value1 | NotAFlag::Value2;

if ( +Flag != 0 ){ ... }
```

**Comments**

* The actual implementation of allowing bitwise operations in `magic_enum::bitwise_operators` is a series of fairly straightforward template functions which could be recreated.
* Enabling bitwise functions for all enums allows them to be used in contexts where it doesn't make sense for them to be able to be used.
* This library doesn't cover the case of enums being used as array indices. However, the operations to allow this could be added as per the above comment.

### Macro Definitions
This is a header file containing a series of macros that defines bitwise and arithmetic operations between strong enums passed as parameters. [Implementation can be found here](https://github.com/mwnciau/Strong-Enum-Overloading/blob/7cd1f9f0e3fb193bb77853ff6e5ed964ad459958/Macros.hpp).

1. Strongly typed ✔️
2. Allows bitwise operators ✔️
3. Allows arithmetic operators ✔️
4. Efficient ✔️
5. Easy to use ✔️
6. Optionally allow interaction with other enums ✔️
 
**Examples**
```
enum class Index : uint32_t { ... };
ENABLE_ENUM_ARITHMETIC_OPERATIONS(Index);

Index index = Index::Value1;
index++;
someArray[ +index ];

enum class Flag : uint32_t { ... };
ENABLE_ENUM_BITWISE_OPERATIONS(Flag);
ENABLE_ENUM_BITWISE_COMPATIBILITY(Flag, Index);

Flag flag = Flag::Value1 | Flag::Value2;
if ( !!flag ){ ... } // or +flag != 0 or flag != Flag::Null

// This is allowed because we have explicitly allowed it with the ENABLE_ENUM_BITWISE_COMPATIBILITY call
flag = Flag::Value1 & Index::Value1;
```

**Comments**

* This works along similar lines to Better Enums, but provides more functionality.
* The main downside to this solution is the use of macros, which can be a nightmare to debug, obfuscate code, and generally should be avoided.

### Include File
This is a file that is designed to be included multiple times in a project containing operator definitions, and controlled by definitions. [Implementation can be found here](https://github.com/mwnciau/Strong-Enum-Overloading/blob/2f7695f421e626d2fd96a583c440e205e8484619/Enum.hxx).

1. Strongly typed ✔️
2. Allows bitwise operators ✔️
3. Allows arithmetic operators ✔️
4. Efficient ✔️
5. Easy to use ✔️
6. Optionally allow interaction with other enums ✔️
 
**Examples**
```
enum class Index : uint32_t { ... };
#define ENABLE_ENUM_INDEX Index
// All definitions are undefined in Enum.hxx to make usage easier
#include "Enum.hxx"

Index index = Index::Value1;
index++;
someArray[ +index ];

enum class Flag : uint32_t { ... };
#define ENABLE_ENUM_FLAG Flag
#define ENABLE_ENUM_FLAG_OTHER Index
#include "Enum.hxx"

Flag flag = Flag::Value1 | Flag::Value2;

// This is allowed because we have explicitly allowed it with the ENABLE_ENUM_FLAG_OTHER definition
flag = Flag::Value1 & Index::Value1;
if ( !!flag ){ ... } // or +flag != 0 or flag != Flag::Null
```

**Comments**

* This solution is similar to the macros solution, but avoids the use of code inside macros. This means errors occur at the actual line of code, and breakpoints can be added.
* The use of a include file used in this way is very non-standard, and their function would be unclear to anyone new to the codebase. Additionally, because the code is shared, only changing based on macro definitions, errors could still be confusing.
 
### ClassEnum
This solution moves away from using strong enums entirely, and instead provides a template class that can be used to provide the desired functionality. [Implementation can be found here](https://github.com/mwnciau/Strong-Enum-Overloading/blob/ddd32c37f628f9bdb8451b8e6d1d8ea9efd11ce9/Classes.hpp).

1. Strongly typed ✔️
2. Allows bitwise operators ✔️
3. Allows arithmetic operators ✔️
4. Efficient ✔️
5. Easy to use ✔️
6. Optionally allow interaction with other enums ❌
 
**Examples**
```
struct Index
{
    using Type = IndexEnum<Index, uint32_t>;
    static constexpr Type Value1 = { 1 };
    ...
}

Index::Type index = Index::Value1;
index++;
someArray[ +index ];

struct Flag
{
    using Type = FlagEnum<Flag, uint32_t>;
    static constexpr Type Value1 = { 1 };
    ...
}

Flag::Type flag = Flag::Value1 | Flag::Value2;
if ( !!flag ){ ... } // or +flag != 0 or flag != Flag::Null
```

**Comments**

* In a release build, the performance is comparable to a strong enum. There is a small (on the nanosecond scale) performance overhead in debug builds.
* In the given implementation, the constructor for creating objects from the underlying is protected, but the wrapping struct (Index / Flag) are specified as friends so they can define the values. A FromUnderlying(val) function could be added to the wrapper if needed.

### General Purpose ClassEnum
This solution is similar to the previous solution, but uses variadic template functions to allow interactions with an arbitrary number of other types. See the first commit of this PR to see the implementation of this.

1. Strongly typed ✔️
2. Allows bitwise operators ✔️
3. Allows arithmetic operators ✔️
4. Efficient ✔️
5. Easy to use ✔️
6. Optionally allow interaction with other enums ✔️

**Examples**

```
struct Index
{
    using Type = IndexEnum<Index, uint32_t/*, AnotherClassEnum, ...*/>;
    static constexpr Type Value1 = { 1 };
    ...
}

Index::Type index = Index::Value1;
index++;
someArray[ +index ];

struct Flag
{
    using Type = FlagEnum<Flag, uint32_t, Index/*, AnotherClassEnum, ...*/>;
    static constexpr Type Value1 = { 1 };
    ...
}

Flag::Type flag = Flag::Value1 | Flag::Value2;
if ( !!flag ){ ... } // or +flag != 0 or flag != Flag::Null

// This is allowed because we have explicitly allowed it with the ENABLE_ENUM_FLAG_OTHER definition
flag = Flag::Value1 & Index::Value1;
```

**Comments**

* The comments for the previous class solution equally apply here. The only reason I've included this seperately is the increased complexity of its implementation, almost doubling the lines of code. I'm also not sure how happy other compilers will be with it.
* Performance wise, this still has no overhead on release builds, but has an even bigger overhead in debug builds (about double that of the previous solution, but still very small).
* Currently, interactions can only be defined between enums created with these functions; although, conceivably, allowing interaction with strong enums / primitive types could be implemented using template specialisations and std::enable_if.

src/openrct2/util/Enum.hpp
```diff
@@ -0,0 +1,510 @@
...
+ *****************************************************************************/
+
+#pragma once
+
+#include <stdint.h>
+
+/**
+ * Base class for all enums, holding the value thus allowing interaction between IndexEnum and FlagEnum
+ */
+template<typename T> struct EnumBase
```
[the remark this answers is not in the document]
tupaschoal: Also, you probably want to check that `T` is integral?

src/openrct2/PlatformEnvironment.h
```diff
@@ -10,24 +10,28 @@
...
-        OPENRCT2,      // Base directory for OpenRCT2 installation.
-        USER,          // Base directory for OpenRCT2 user content.
-        CONFIG,        // Base directory for OpenRCT2 configuration.
-        CACHE,         // Base directory for OpenRCT2 cache files.
-        DOCUMENTATION, // Base directory for OpenRCT2 doc files.
+        using Type = IndexEnum<DIRBASE, size_t>;
+
+        static constexpr Type Init = { 0 };
+        static constexpr Type RCT1 = { 0 };          // Base directory for original RollerCoaster Tycoon 1 content.
+        static constexpr Type RCT2 = { 1 };          // Base directory for original RollerCoaster Tycoon 2 content.
```
tupaschoal, line 26: I think one disadvantage of this approach is having to manually define all enumeration values, which we do lots of times, but not always.

it also allows overlapping values, which is a feature in this specific part, but might hide a bug in another

mwnciau:
Thanks for the feedback @tupaschoal. There were a few things left over in there from when I was testing it. (I'm also probably going to have a read through the C++ Core Guidelines, too)

Ultimately, while this solution is interesting/nice, I think it's probably too complicated to include as part of OpenRCT rather then a standalone library.

A solution based on how Magic Enum implements the bitwise operators might be more realistic. It has a greater scope for making mistakes than this class solution would and working with enums that interact with other enums will be slightly awkward, but it's closer to native C++, much simpler, and would allow us to use enum classes instead of the namespace/constexpr compromise. Basically, a series of definitions for each of the operators along these lines:

```
template <class Enum, typename = std::enable_if_t<std::is_enum<Enum>::value>>
constexpr Enum operator | (const Enum& lhs, const Enum& rhs)
{
    return static_cast<Enum>(static_cast<std::underlying_type_t<Enum>>(lhs) | static_cast<std::underlying_type_t<Enum>>(rhs));
}
```

Which would allow the bitwise | operator to be used between members of any enum class (and the unary + operator could also be defined to convert to the underlying type allowing for easy array access).

mwnciau:
Following my previous comment, I've decided to take this in a slightly different direction. Instead of a class wrapper, the PR now follows the Magic Enum example of defining operators for strong enums. This essentially re-implements the operators that were used for weak enums and so the current namespace/constexpr could be avoided.

I've done some cursory searches through the codebase to see which operators are used on enums, and thus have created:
* bitwise operators: `|`, `&`, `^`, `|=`, `&=`, `^=`, `~` (both sides will need to be of the same enum type)
* helper unary operators: `+` to convert to the underlying type and `!` to compare the underlying type to 0 (useful for getting boolean values of flags)
* bit shift operators: `enum << integral`, `enum >> integral` and `integral << enum`, always returning the type on the left hand side of the operator

This isn't quite as strict as the class-based solution, but that's what code reviews are for.

tupaschoal:
One innocent question: if it's "magic-enum" inspired, why not using it? Is it adding things that the former doesn't have or because it's a simpler version of it?

mwnciau:
A bit of both really. Magic Enum is an enum utility library, but it seems to be primarily for static reflection, and the bitwise operators are tagged on a little bit. It's not a huge library anyway (~1000 lines), but it is another dependency for what is essentially just ~35 lines of code in that library.

Additionally, Magic Enum implements strictly only the bitwise operators, so the `+` and `!` unary operators don't exist (if indeed it's agreed that we want them), the bit shifts don't exist (although I don't know how useful they'd actually be), and, probably more importantly, the increment operator `++` doesn't exist, which I think would be used quite a bit in our codebase to iterate through an enum. Plus, there's the flexibility to add other operators, if needed.

src/openrct2/PlatformEnvironment.cpp
```diff
@@ -21,20 +21,21 @@ using namespace OpenRCT2;
...
 private:
-    std::string _basePath[DIRBASE_COUNT];
+    std::string _underlyingBasePath[+DIRBASE::Count];
+    const EnumeratedArray<std::string, DIRBASE> _basePath{ _underlyingBasePath };
 
 public:
     explicit PlatformEnvironment(DIRBASE_VALUES basePaths)
     {
-        for (size_t i = 0; i < DIRBASE_COUNT; i++)
+        for (auto i = static_cast<DIRBASE>(0); i < DIRBASE::Count; i++)
```
Gymnasiast, line 30: Would something like
```suggestion
        for (DIRBASE i = 0; i < DIRBASE::Count; i++)
```
work?

src/openrct2/windows/_legacy.cpp
```diff
@@ -345,13 +345,13 @@ bool window_ride_construction_update_state(
     bool turnOffLiftHill = false;
     if (!(_enabledRidePieces & (1ULL << TRACK_LIFT_HILL_CURVE)))
     {
-        if (TrackFlags[trackType] & TRACK_ELEM_FLAG_CURVE_ALLOWS_LIFT)
+        if (!!(TrackFlags[trackType] & TrackElemFlag::AllowLiftCurve))
```
Gymnasiast, line 348: This doesn't seem necessary, considering we already have an `if` (which checks for truthy). If it is absolutely necessary, then please use `!= 0` (our convention).

richard-fine:
I have some concerns about this PR:

- Being able to define these operators for all enum classes once is kinda nice, but it really means ALL enum classes, which includes ones that are *not* flags enums. So this is reducing the safety that `enum class` is bringing for those types. (I'm also a bit curious about the impact these operators have on build times, but that's a minor concern).

- Defining the arithmetic operators: I know they are used for loop iteration a lot, but I think it is clearer in most cases to use a named `e_next()` function. In cases where the order of the enum values has some meaning then it gives you a chance to express that in the name (e.g., for the Direction type we could have `direction_next_clockwise()`), and it also allows for multiple strategies for traversing the enums (e.g. for RideType we could have both `ride_type_next(rid)` and `ride_type_next(rid, RideCategory::Gentle)`). Not every enum needs both of those, but I think it wouldn't be amazing to have some enums using arithmetic operators and others using named functions.

- The whole !! thing: It's not immediately obvious what it means, and I think it's going to be harder to spot at a glance when we mixed up `!` and `!!` (vs when we have `!` vs not having it). In my experience it's again better to have explicit functions that convert to bool with particular semantics that let you reduce the bitwise operations everywhere, like `HasAll` and `HasAny`.

duncanspumpkin:
https://godbolt.org/z/PWorqd I created this as an idea for the ride mode enum class i think it would be useful for enums that are used as both flags and regular enums.

ZehMatt:
I'm against defining the bitwise operators for all enums, enabling the operators should be explicit.

mwnciau:
I've tidied up one of my previous implementations of a class based approach, and incorporated a few of your suggestions, @richard-fine, here: https://godbolt.org/z/GPfsnE.

I've also made some changes to this PR. The bitwise operators now use SFINAE techniques to restrict their definition to only enum classes with Null defined as a member, reintroducing some of the safety that this PR had removed. I've also removed the bitshift operators in light of @duncanspumpkin's EnumToFlags functions.

I don't think there's an "ideal" solution, but I do think either this class-based approach or defining enum operators as in the current PR is a better approach than using namespaces without type safety, where errors when using similarly named namespace-constexpr enums are relatively easy to miss. Both approaches are constexpr, so there are no runtime differences in release builds, and a (very) small impact on debug builds. As you say, both will have an impact on build time, but I suspect that will be negligible.

### Class Based Approach

**Pros**
* Strongly Typed
* Allows defining helper member functions such as `HasAll` or `HasAny` allowing simple calls such as `flag.HasAny(Flag::One, Flag::Two)`
* Wrapper class allows helper functions such as the `Direction::Rotate` without needing to define functions in the global scope

**Cons**
* Non-standard approach to enums that might catch people off
  * Defining a flag enum follows a different paradigm
  * Having to use SomeFlag::Type instead of SomeFlag as the type
* When using an operator with the underlying type (e.g. `flag | 0`), the compiler will try to convert the underlying type to the FlagEnum, showing a 'constructor is protected' error instead of an 'operator doesn't exist' error
* Tooltips aren't always as helpful (at least in VS). E.g. if the underlying type is uint8_t, VS will display a character rather than the integer value: https://i.imgur.com/UFyMPMr.png. Slightly better for other integer types: https://i.imgur.com/WzELvbJ.png.
* No implicit conversion to/from primitive types

**Example**

```
struct Direction
{
    // Direction is made a friend of Type so can access and change the underlying value
    // directly in member methods
    using Type = FlagEnum < Direction, uint16_t > ;

    static constexpr Type Null  = { 0 };
    static constexpr Type North = { 1 << 0 };
    ...
}

Direction::Type dir = Direction::North | Direction::East;

// false
dir.HasAny(Direction::South);

// true
dir != Direction::Null;
```

### Defining operators for enum classes

**Pros**
* Strongly Typed
* Enum classes can be used as flags without any additional code changes

**Cons**
* bitwise operators are defined for any enum with Null defined
  * This is a bit unclear unless you know the codebase (although static_assert errors provide hints, but the errors are a bit messy)
  * Potentially still enabled if a non-flag enum has a Null member
* No implicit conversion to/from primitive types
* Helper functions cannot be defined in the same scope as the enums (e.g. for an enum `Direction`, the function `Rotate` couldn't be `Direction::Rotate`, in the "Direction" scope)

**Example**

```
enum class Direction
{
    // Defining Null enables the bitwise operators
    Null = 0,
    North = ( 1 << 0 ),
    ...
}

Direction dir = Direction::North | Direction::East;

// false
FlagHasAny(dir, Direction::South);

// true
dir != Direction::Null;
```

src/openrct2/windows/_legacy.cpp
mwnciau: I've taken a different approach to this in the latest version of the PR

src/openrct2/PlatformEnvironment.cpp
mwnciau: `DIRBASE` is a strong enum so doesn't allow implicit conversion from an integer

ZehMatt:
I don't like that one has to define 'Null' in the enum in order to enable the operators, enabling this operator should be explicit as I said before, there are enums that may have Null but dont use bitmasks which would have an undesirable result.

I think having something like https://godbolt.org/z/o3Mhro and the @duncanspumpkin for enums that not specifically define bitmasks is the way to go honestly, anything else seems to be over engineered.

richard-fine:
> I do think either this class-based approach or defining enum operators as in the current PR is a better approach than using namespaces without type safety, where errors when using similarly named namespace-constexpr enums are relatively easy to miss

Just to be clear, I do agree completely with this - I think the intention of improving our tools for working with flags-style enums is totally sound. We disagree about the solution, not about the problem.

I agree with @ZehMatt - I don't like the 'magic member' approach, I think we should make it be something that each enum class explicitly opts-in to. Just because our toolbox includes SFINAE does not mean we should use it :) 

From the POV of a newcomer to the codebase, if you have two `enum class` types, and one of them you can use bitwise operators with, and the other you can't... I think it would be very difficult to realise that the presence of a 'Null' member is what's causing the difference in behaviour. 

Plus, I think we could probably argue for a long time about whether `Null` is what it should be called... for many flags enums I think `None` will read more naturally, and I bet we could find other cases where some third option is also desirable, etc.

Wish we could do C#-style extension methods on enum class types...

tupaschoal:
I'll be closing this as it's been open for a while and also because @duncanspumpkin was implemented at some point and seems to be working well (though far simpler):

https://github.com/OpenRCT2/OpenRCT2/blob/93d13552a968f6945c829e3ee4a744aafc6896e1/src/openrct2/util/Util.h#L64-L78

Thanks for the investigation and the great explanations though, I learned a lot from it!

PR #12759, opened 2020-08-23
WinNT5.1 fixes

janisozaur:


AaronVanGeffen:
Please update your commit messages to follow our [guidelines for commit messages](https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages) :wink:

src/openrct2/platform/Platform.Win32.cpp
```diff
@@ -37,6 +37,10 @@
 
 #    include <iterator>
 
+#    if _WIN32_WINNT < 0x600
+#        define swprintf_s(a, b, c, d, ...) swprintf(a, b, c, ##__VA_ARGS__)
```
Broxzier, line 41: This drops the fourth argument, doesn't that mean that where this is used it prints garbage now?

janisozaur: The fourth argument is buffer length, unused in the non-`_s` version.

ZehMatt: I would rather like to see this not being a macro, you can use variadic template arguments and perfect forwarding.

janisozaur: That would just shift the ifdef into a template, is that what you mean?

ZehMatt: Just the function you still need the if _WIN32_WINNT < 0x600

AaronVanGeffen:
@janisozaur Is this ready to be merged?

janisozaur:
@AaronVanGeffen yes. I'm still trying to figure out the problem with our CI, but that's another PR.

Issue #12756, opened 2020-08-22
Scenario Editor crashing the game (Mac OS X, 0.3.0)


clong8887:
OS: Mac OS X Mojave 10.14.6
Version: 0.3.0
Commit/Build: Not sure what this means? But it says "135cc10b4 on HEAD, DEBUG" next to the version number

While working in Scenario Editor, just selecting items to have available, the game closes itself down without warning. It only seems to happen while I am selecting items under scenery items (I think mostly in the trees/walls/small items section but maybe not always). It has not happened while I am selecting rides.

clong8887:
Yeah it doesn’t say anything at all. I don’t think it was a specific scenery item that caused the issues but I suppose it’s possible.

clong8887:
Ok, it still doesn't give me any actual error messages but I think the problem item may be "Geometric Fountain". I can consistently get the game to crash if I move my cursor over the item when going through the Advanced selection menu. No other items seem to be causing the problem, as far as I can tell. I'm not clicking the geometric fountain, just hovering over it.

clong8887:
I tried deleting the relevant file from my ObjData folder but that didn't seem to help. I will try to just avoid hovering over that item for now but hopefully this helps others on the back end.

AaronVanGeffen:
Ah, since you're using a debug build, I'm fairly sure you're running into #11875. The macOS v0.3.0 shouldn't have been a debug build, though. I wonder what went wrong there.

clong8887:
Upon trying to avoid Geometric Fountain it not just that one object, something else (not sure what) in the early G's of that menu seems to also trigger it. I scrolled to the bottom of that list and worked upwards and didn't have any issues until I got to that area of the menu again.

AaronVanGeffen:
If my hypothesis is correct, the builds for #12761 should resolve the problem.

AaronVanGeffen:
The PR in question has been merged, so the assertion should no longer trigger on develop builds.

clong8887:
Hi- sorry I stopped having the problems by avoiding the objects I know to be an issue, but ran into it again today while in an actual park (rather than editor) so I came back.

I'm new to all of this, so could you please explain how I would download and implement the build at #12761?

AaronVanGeffen:
[...]

PR #12761, opened 2020-08-23
GitHub CI: use Release configuration for macOS

AaronVanGeffen:
Instead of using the release configuration, the macOS version of v0.3.0 is a debug build. This changes the GitHub CI such that `xcodebuild` generates release builds instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,7 +99,7 @@ jobs:
       - name: Build OpenRCT2
         run: |
           . scripts/setenv
-          xcodebuild
+          xcodebuild -configuration Release
           mkdir -p artifacts
           mv build/Release/OpenRCT2.app artifacts
           echo -e "\033[0;36mCompressing OpenRCT2.app...\033[0m"


janisozaur:
Does it produce symbols still?

AaronVanGeffen:
I would assume it doesn't include them, no. Much like our Linux/AppImage and Windows builds in that respect.

janisozaur:
We get enough reports from mac users to make sure that at least symbol names are present. You can test with `assert` or `dereference` using in-game console

AaronVanGeffen:
I'm counting on asserts _not_ being present for #12756 (which suffers from #12761).

We don't get backtrace reports from macOS, so I'm not that fussed about symbols being present in releases, to be honest.

janisozaur:
optimisation is orthogonal to symbols being present

tupaschoal:
Did you consider @janisozaur concerns?

Gymnasiast:
I did, but I found the points that Aaron made in https://github.com/OpenRCT2/OpenRCT2/pull/12761#issuecomment-678816912 to be more important.

janisozaur:
@Gymnasiast what does that non-answer mean? Are symbols available or not?

PR #12771, opened 2020-08-25
Fix #12764: Rides don't start aged anymore

### LINKED ISSUE #12764: "Year Built" bug in ride dialog upon loading scenarios from both RCT1 installation and RCT2.
Windows 10
0.3.0
135cc10

This is my first post here, or on any forum like this, so I apologize if I am doing this incorrectly. 

I believe I have found a bug. 

When attempting to open RCT1 scenarios from my Steam installation of RollerCoaster Tycoon Deluxe within OpenRCT2 (I have pointed the program to the proper folder using the options menu), all of the scenarios appear to properly load.

However, in all RCT1 scenarios, the rides show as being built “this year”, regardless of how old they are supposed to be. This behavior does not happen when using vanilla RCT1 deluxe; the screenshots below are scenarios created from the same files in the RCT1 deluxe directory that the OpenRCT2 installation points to (see screenshot 1). 

It appears that the other statistics about the ride; number of total customers, lifetime profits, relability, last breakdown, etc., do load properly (you can see this by looking at the side-by-side screenshots of OpenRCT2 and RCT1 Deluxe (see screenshot 2).   

I have also noticed this behavior when loading scenarios from the RCT2 vanilla game. For example, in Rock ‘n’ Roll revival, when I load it in vanilla RCT2, it shows the ride “Jet Plane Coaster” was built “10 years ago”, but in OpenRCT2, it shows it being built “This Year" (see screenshot 3). 

It appears that this might well just be a visual bug; i.e., the number of years is not being properly reported to whatever shows it in the ride dialog, since the other historical stats are correct. 

But this has prevented me from playing scenarios like Crumbly Woods where the aged status of the rides is a major part of the challenge. Without knowing how old the rides are, it makes it more difficult to play. (see Screenshot 4; upon loading the park, all the rides sow as being built "This year")

The “built” counter does appear to properly increase as time progresses in-game. 

This is my first time ever reporting a bug for any game and I have very little programming experience (I do web development on the side and know CSS, HTML and Javascript but don't know much about actually writing applications) so if there is anything you need from me (saves, the sc4/sc6 files themselves, etc.) please let me know. 

**Screenshots / Video:**
Screenshot 1:

screenshot 1
Screenshot 2:
screenshot 2
Screenshot 3:
screenshot 3
Screenshot 4: 
screenshot 4

tupaschoal:
https://github.com/OpenRCT2/OpenRCT2/pull/11712/commits/f6fbd272ecfe2aa994690dfcfe8c2bae11809b1a changed the fields to be unsigned, but many rides' ages were set to negative to indicate that they were quite old, so that the `getAge()` function could do `elapsedMonths - (-age)` = `elapsedMonths + Age`. Changing it to be ` = gDateMonthsElapsed`, would just be 0 on some cases. 

There are several ways to fix this, this is one of them, but I'm opening up for discussions

### DIFF
--- a/src/openrct2/rct1/RCT1.h
+++ b/src/openrct2/rct1/RCT1.h
@@ -176,7 +176,7 @@ struct rct1_ride
     uint8_t pad_11F[0x7];              // 0x11F
     uint8_t spiral_slide_progress;     // 0x126
     uint8_t pad_127[0x9];              // 0x127
-    uint16_t build_date;               // 0x130
+    int16_t build_date;                // 0x130
     money16 upkeep_cost;               // 0x131
     uint16_t race_winner;              // 0x132
     uint8_t unk_134[2];                // 0x134
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -932,7 +932,7 @@ void reset_all_ride_build_dates()
 {
     for (auto& ride : GetRideManager())
     {
-        ride.build_date = gDateMonthsElapsed;
+        ride.build_date -= static_cast<int32_t>(gDateMonthsElapsed);
     }
 }
 
@@ -7114,7 +7114,7 @@ void Ride::Delete()
 void Ride::Renew()
 {
     // Set build date to current date (so the ride is brand new)
-    build_date = gDateMonthsElapsed;
+    build_date = static_cast<int32_t>(gDateMonthsElapsed);
     reliability = RIDE_INITIAL_RELIABILITY;
 }
 
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -280,7 +280,7 @@ size_t Ride::GetNumPrices() const
 
 int32_t Ride::GetAge() const
 {
-    return static_cast<int32_t>(gDateMonthsElapsed) - build_date;
+    return gDateMonthsElapsed - build_date;
 }
 
 int32_t Ride::GetTotalQueueLength() const
@@ -932,7 +932,7 @@ void reset_all_ride_build_dates()
 {
     for (auto& ride : GetRideManager())
     {
-        ride.build_date -= static_cast<int32_t>(gDateMonthsElapsed);
+        ride.build_date -= gDateMonthsElapsed;
     }
 }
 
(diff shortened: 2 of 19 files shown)


duncanspumpkin:
Agree that these should be negative as gMonthsElapsed starts when the scenario starts and some rides should be old at the start so therefore build date should allow negative values.

src/openrct2/ride/Ride.cpp
```diff
@@ -278,9 +278,9 @@ size_t Ride::GetNumPrices() const
     return result;
 }
 
-int32_t Ride::GetAge() const
+int16_t Ride::GetAge() const
 {
-    return static_cast<int32_t>(gDateMonthsElapsed) - build_date;
+    return static_cast<int16_t>(gDateMonthsElapsed) - build_date;
```
duncanspumpkin, line 283: Why the size change?

duncanspumpkin:
You'll need to increment the network version

src/openrct2/rct1/RCT1.h
```diff
@@ -176,7 +176,7 @@ struct rct1_ride
     uint8_t pad_11F[0x7];              // 0x11F
     uint8_t spiral_slide_progress;     // 0x126
     uint8_t pad_127[0x9];              // 0x127
-    uint16_t build_date;               // 0x130
+    int16_t build_date;               // 0x130
```
James103, line 179: Your type change has resulted in misaligned comments, making the clang-format unhappy.

IntelOrca:
I didn't know this, looks fine. I recommend sticking with `int32_t` everywhere and only casting on import / export to SV6.

src/openrct2/ride/Ride.cpp
tupaschoal: Just wanted to keep the return type size consistent with the parameters, but I'm reverting this as per Ted's recommendation.

src/openrct2/rct1/RCT1.h
tupaschoal: Yeah

tupaschoal:
@IntelOrca or @duncanspumpkin is this ok now?

IntelOrca:
I suggest making `gDateMonthsElapsed` an `int32_t`.

tupaschoal:
> I suggest making `gDateMonthsElapsed` an `int32_t`.

Done

IntelOrca (approved):

PR #12036, opened 2020-06-22
Part of #11437: moved window_setup_file_association(s) in Platform2.hpp

frutiemax:
window_setup_file_association(s) now become SetupFileAssociation(s). I'd like to get a good review on this one as I'm not too sure if I did it correctly.

PR #12773, opened 2020-08-25
Part of #11437: Refactor Win32 RemoveFileAssociations to Platform2

janisozaur:
Additionally fixes missing changes from previous refactor in #12036 that
borked WinNT 5.1 support

### DIFF
--- a/src/openrct2/platform/Platform.Win32.cpp
+++ b/src/openrct2/platform/Platform.Win32.cpp
@@ -37,6 +37,10 @@
 
 #    include <iterator>
 
+#    if _WIN32_WINNT < 0x600
+#        define swprintf_s(a, b, c, d, ...) swprintf(a, b, c, ##__VA_ARGS__)
+#    endif
+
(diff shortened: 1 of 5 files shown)


AaronVanGeffen (commented):
Good to split this off from the other PR. It looks good to me, but I can't this myself. (It would be good to have confirmation of things working before this is merged, naturally.)

src/openrct2/platform/Platform.Win32.cpp
```diff
@@ -37,6 +37,10 @@
 
 #    include <iterator>
 
+#    if _WIN32_WINNT < 0x600
+#        define swprintf_s(a, b, c, d, ...) swprintf(a, b, c, ##__VA_ARGS__)
```
IntelOrca, line 41: Does `std::swprintf` work? Or rather is it possible to have `std::swprintf` be used instead of `swprintf_s` by default?

janisozaur: Is it worth it though?

IntelOrca (approved):

PR #12783, opened 2020-08-26
Use Formatter in windows/Ride.cpp

Gymnasiast:
This removes some diabolical casting with nice Formatter calls. This also turned out to be the last call to `set_format_arg_body()` outside of the `Formatter::Add()` function, so I inlined that.

### DIFF
--- a/src/openrct2/localisation/Localisation.h
+++ b/src/openrct2/localisation/Localisation.h
@@ -80,11 +80,6 @@ extern const rct_string_id DateDayNames[31];
 extern const rct_string_id DateGameMonthNames[MONTH_COUNT];
 extern const rct_string_id DateGameShortMonthNames[MONTH_COUNT];
 
-[[maybe_unused]] static inline void set_format_arg_body(uint8_t* args, size_t offset, uintptr_t value, size_t size)
-{
-    std::memcpy(args + offset, &value, size);
-}
-
 class Formatter
 {
     const uint8_t* StartBuf;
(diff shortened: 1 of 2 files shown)


src/openrct2/localisation/Localisation.h
```diff
@@ -156,7 +151,7 @@ class Formatter
         {
             convertedValue = reinterpret_cast<uintptr_t>(value);
         }
-        set_format_arg_body(CurrentBuf, 0, convertedValue, sizeof(TSpecified));
+        std::memcpy(CurrentBuf, &convertedValue, sizeof(TSpecified));
```
AaronVanGeffen, line 154: Perhaps we can avoid the C-style `memcpy`, e.g. by using:
```suggestion
        *static_cast<TSpecified*>(CurrentBuf) = convertedValue;
```

Gymnasiast: That just results in an error:
```
/home/michael/Programma's/Code/CLionProjects/OpenRCT2/src/openrct2/core/../localisation/Localisation.h:154:10: error: invalid static_cast from type ‘uint8_t*’ {aka ‘unsigned char*’} to type ‘short unsigned int*’
  154 |         *static_cast<TSpecified*>(CurrentBuf) = convertedValue;
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

AaronVanGeffen: I suppose a `reinterpret_cast` is in order, then.

Gymnasiast: Does not work either. I get a similar error, but involving `uintptr_t` this time.

ZehMatt: It should keep using std::memcpy for now, reinterpret_cast should be avoided.

ZehMatt (approved):
This will do nicely with the PR I have planned.

PR #12785, opened 2020-08-26
Start removal of gfx_draw_string_*()

Gymnasiast:


src/openrct2/drawing/Text.h
```diff
@@ -48,3 +49,16 @@ class StaticLayout
...
+
+void DrawTextBasic(
+    rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, rct_string_id format, const Formatter& ft, uint8_t colour,
+    TextAlignment alignment, bool underline = false);
+void DrawTextBasic(
+    rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, rct_string_id format, const void* args, uint8_t colour,
+    TextAlignment alignment, bool underline = false);
+void DrawTextEllipsised(
+    rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, int32_t width, rct_string_id format, const Formatter& ft,
+    uint8_t colour, TextAlignment alignment, bool underline = false);
```
Gymnasiast, line 61: `colour_t`

src/openrct2/drawing/Drawing.h
```diff
@@ -648,24 +649,23 @@ void FASTCALL gfx_draw_sprite_raw_masked_software(
 // string
 void gfx_draw_string(rct_drawpixelinfo* dpi, const_utf8string buffer, uint8_t colour, const ScreenCoordsXY& coords);
 
+/** @deprecated */
```
ZehMatt, line 652: You could use https://en.cppreference.com/w/cpp/language/attributes/deprecated, this will add some compiler warnings but will actually make it clear that its not supposed to be used anymore.

Gymnasiast: The problem with that is that compilation will fail if a compiler warning is output in our project.

ZehMatt: Is there no way to ignore this specific warning maybe?

Gymnasiast: I don't know, but I'm sure @janisozaur does.

ZehMatt: I mean its not that big of a deal, we are going to remove them anyway soonish I guess.

src/openrct2/localisation/Localisation.h
```diff
@@ -112,6 +112,11 @@ class Formatter
         return CurrentBuf;
     }
 
+    auto GetStartBuf() const
```
ZehMatt, line 115: Unrelated?

ZehMatt (approved):
Besides the one unrelated change, lgtm.

src/openrct2/localisation/Localisation.h
Gymnasiast: No. Take a look at `DrawTextBasic` and `DrawTextEllipsised`. They need to get the start buffer. If I just call `Buf()`, it will crash the game.

duncanspumpkin:
Why are we exposing the args overload? Nothing should be using that version of the function. Everything should use the formatter.

PR #12792, opened 2020-08-27
Enable scripting for MinGW CI job

janisozaur:
Uses new 0.3.0 images from https://github.com/OpenRCT2/openrct2-docker-build/pull/1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,7 +264,7 @@ jobs:
         uses: actions/checkout@v1
       - name: Build OpenRCT2
         shell: bash
-        run: . scripts/setenv && build -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=Debug -DDISABLE_NETWORK=ON -DDISABLE_HTTP_TWITCH=ON -DDISABLE_OPENGL=ON
+        run: . scripts/setenv && build -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=Debug -DDISABLE_NETWORK=ON -DDISABLE_HTTP=ON -DDISABLE_OPENGL=ON
   android:
     name: Android
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,26 @@ jobs:
           mkdir bin && cd bin
           cmake .. -G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DCMAKE_BUILD_TYPE=MinSizeRel -DDISABLE_IPO=on -DFORCE32=on -DBUILD_SHARED_LIBS=ON -DENABLE_SCRIPTING=OFF
           ninja -k0
+  windows-mingw-nt51:
+    name: Windows (win32, WinNT5.1) using mingw
+    runs-on: ubuntu-latest
+    needs: [check-code-formatting]
+    container:
+      image: openrct2/openrct2-build:0.3.0-mingw
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         run: |
           sudo su
           mkdir bin && cd bin
-          cmake .. -G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DCMAKE_BUILD_TYPE=MinSizeRel -DDISABLE_IPO=on -DFORCE32=on -DBUILD_SHARED_LIBS=ON -DENABLE_SCRIPTING=OFF
+          cmake .. -G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DCMAKE_BUILD_TYPE=MinSizeRel -DDISABLE_IPO=on -DFORCE32=on -DBUILD_SHARED_LIBS=ON -DENABLE_SCRIPTING=ON
```
janisozaur, line 90:
```suggestion
          cmake .. -G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DCMAKE_BUILD_TYPE=MinSizeRel -DDISABLE_IPO=on -DFORCE32=on -DBUILD_SHARED_LIBS=ON -DENABLE_SCRIPTING=OFF
```

janisozaur:
Interesting to see https://github.com/OpenRCT2/OpenRCT2/pull/12667 having mingw job succeed.

janisozaur:
OK, now I have the error I expected:
```
 -- Checking for module 'jansson>=2.5'
CMake Error at /usr/share/cmake-3.18/Modules/FindPkgConfig.cmake:545 (message):
--   Package 'jansson', required by 'virtual:world', not found
  A required package was not found
Call Stack (most recent call first):
  /usr/share/cmake-3.18/Modules/FindPkgConfig.cmake:733 (_pkg_check_modules_internal)
  src/openrct2/CMakeLists.txt:123 (PKG_CHECK_MODULES)
  CMakeLists.txt:283 (include)
```

I can't proceed with this until after #12667 is merged because there's no jansson package in the image anymore.

janisozaur:
cmake is drunk again.

It complains about missing pkg-config, which I fixed in https://github.com/OpenRCT2/openrct2-docker-build/pull/3, don't know what's going on here.

janisozaur:
image
I've played myself.

janisozaur:
Finding duktape somehow breaks pkg-config. Possibly a bug around https://github.com/OpenRCT2/OpenRCT2/blob/4030d9050ed75b21f4bb8874a53a2353eeb04d9a/cmake/duktapeConfig.cmake#L14-L15

janisozaur:
Requires https://github.com/OpenRCT2/openrct2-docker-build/pull/6

janisozaur:
Still needs ` -liconv -lp11-kit.dll -lhogweed -lgmp.dll -lnettle`

AaronVanGeffen:
As of this PR, MinGW builds are both networked and have scripting support, right? That's worth mentioning in the changelog and release announcement.

janisozaur:
@AaronVanGeffen testing it on Windows XP VM proved this is missing symbols. I wonder if that's due to me switching where I source my mingw packages or if perhaps they got updated since I last tried. As this was meant to still get into the release, I removed the changelog entry for now and we can re-enable it when it really happens.

I'd still like the mingw CI job to run, so we can get the coverage and breaking the 5.1 support will be caught quicker.

janisozaur:
For reference: the missing symbol is `GetFinalPathNameByHandleA` from `kernel32.dll`. In the docker image used to build it:

```
$ i686-w64-mingw32-nm /usr/i686-w64-mingw32/lib/libkernel32.a  | grep GetFinalPathNameByHandleA
00000000 T _GetFinalPathNameByHandleA@16
00000000 I __imp__GetFinalPathNameByHandleA@16
```

Which makes me suspect it somehow sneaked in. I can't quite tell where it's used though.

PR #12795, opened 2020-08-28
Add Ride.totalProfit to the plugin API

CorySanin:
Simple addition to the plugin API that allows for querying the total profit of a ride.


AaronVanGeffen:
Scripts using this wouldn't work on v0.3.0, so perhaps we should increase the plugin API version. @IntelOrca ?

PR #12796, opened 2020-08-28
Close #12430, refactor: OBJ_STRING_ID to use strong enum

### LINKED ISSUE #12430: Refactor OBJ_STRING_ID to use strong enum

Lukasz-Pe:
Closes Issue #12430

### DIFF
--- a/src/openrct2/object/BannerObject.cpp
+++ b/src/openrct2/object/BannerObject.cpp
@@ -26,7 +26,7 @@ void BannerObject::ReadLegacy(IReadObjectContext* context, OpenRCT2::IStream* st
     _legacyType.banner.scenery_tab_id = OBJECT_ENTRY_INDEX_NULL;
     stream->Seek(2, OpenRCT2::STREAM_SEEK_CURRENT);
 
-    GetStringTable().Read(context, stream, static_cast<uint8_t>(ObjectStringID::NAME));
+    GetStringTable().Read(context, stream, ObjectStringID::NAME);
 
     rct_object_entry sgEntry = stream->ReadValue<rct_object_entry>();
     SetPrimarySceneryGroup(&sgEntry);
--- a/src/openrct2/object/BannerObject.cpp
+++ b/src/openrct2/object/BannerObject.cpp
@@ -26,7 +26,7 @@ void BannerObject::ReadLegacy(IReadObjectContext* context, OpenRCT2::IStream* st
     _legacyType.banner.scenery_tab_id = OBJECT_ENTRY_INDEX_NULL;
     stream->Seek(2, OpenRCT2::STREAM_SEEK_CURRENT);
 
-    GetStringTable().Read(context, stream, OBJ_STRING_ID_NAME);
+    GetStringTable().Read(context, stream, static_cast<uint8_t>(ObjectStringID::NAME));
 
     rct_object_entry sgEntry = stream->ReadValue<rct_object_entry>();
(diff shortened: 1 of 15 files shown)


src/openrct2/object/ObjectJsonHelpers.cpp
```diff
@@ -462,14 +462,14 @@ namespace ObjectJsonHelpers
     static uint8_t ParseStringId(const std::string& s)
```
duncanspumpkin, line 462: Return the new enum type.

src/openrct2/object/StringTable.h
```diff
@@ -21,21 +21,21 @@ namespace OpenRCT2
...
+    PARK_NAME = 1,
+    SCENARIO_DETAILS = 2,
+    CAPACITY = 2,
+    VEHICLE_NAME = 3,
 };
 
 struct StringTableEntry
 {
-    uint8_t Id = OBJ_STRING_ID_UNKNOWN;
+    uint8_t Id = static_cast<uint8_t>(ObjectStringID::UNKNOWN);
```
duncanspumpkin, line 38: Use the new type

src/openrct2/object/RideObject.cpp
```diff
@@ -393,12 +393,12 @@ void RideObject::DrawPreview(rct_drawpixelinfo* dpi, [[maybe_unused]] int32_t wi
 
 std::string RideObject::GetDescription() const
 {
-    return GetString(OBJ_STRING_ID_DESCRIPTION);
+    return GetString(static_cast<uint8_t>(ObjectStringID::DESCRIPTION));
```
duncanspumpkin, line 396: Modify get string to use the new type

src/openrct2/object/BannerObject.cpp
```diff
@@ -26,7 +26,7 @@ void BannerObject::ReadLegacy(IReadObjectContext* context, OpenRCT2::IStream* st
     _legacyType.banner.scenery_tab_id = OBJECT_ENTRY_INDEX_NULL;
     stream->Seek(2, OpenRCT2::STREAM_SEEK_CURRENT);
 
-    GetStringTable().Read(context, stream, OBJ_STRING_ID_NAME);
```
duncanspumpkin, line 29: Modify string table read to take the new type. Modify the set function as well.

duncanspumpkin (commented):
Lots of changes to be made to string table to use the new type.

src/openrct2/object/BannerObject.cpp
Lukasz-Pe: There are following functions in StringTable.h, class StringTable:
```
void Read(IReadObjectContext* context, OpenRCT2::IStream* stream, uint8_t id);
std::string GetString(uint8_t id) const;
std::string GetString(uint8_t language, uint8_t id) const;
void SetString(uint8_t id, uint8_t language, const std::string& text);
```
and as I undertand field `uint8_t id` is the field for `ObjectStringID`.
I want to replace all `uint8_t id` fields with `ObjectStringID id`. Is that ok, or should I pay more attention here for something else?

Lukasz-Pe: I've also modified `Object::GetString` function in Object.cpp and modified `auto sz = GetOverrideString(index);` to `auto sz = GetOverrideString(static_cast<uint8_t>(index));`

duncanspumpkin (approved):

PR #12798, opened 2020-08-28
Fix wrong syntax of std::array in Window.cpp

gguedesaz:
Fixed a compilation error on Window.cpp caused by incorrect std::array syntax.

### DIFF
--- a/src/openrct2-ui/interface/Window.cpp
+++ b/src/openrct2-ui/interface/Window.cpp
@@ -194,7 +194,7 @@ rct_window* window_create_auto_pos(
     auto screenHeight = uiContext->GetHeight();
 
     // Place window in an empty corner of the screen
-    const std::array cornerPositions = {
+    const std::array<ScreenCoordsXY, 4> cornerPositions = {
         ScreenCoordsXY{ 0, 30 },                                          // topLeft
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This is already part of #12749.

Out of curiosity, on what platform are you seeing compilation errors?

gguedesaz:
@AaronVanGeffen I'm seeing this on macOS High Sierra 10.13.6, using XCode 10.0. Should I close this PR then? I'm actually migrating to cmake, so it won't hurt me that much.

AaronVanGeffen:
@gguedesaz Yeah, #12749 is addressing the same thing for the same reason.

Curious, does this change make the project fully compile for you on macOS 10.13? We're seeing some problems on our macOS CI (10.15) when we try to target 10.13 directly, because of code that's using C++17 `std::optional`. #12749 is making it work through a workaround, though.

gguedesaz:
I don't think I have the capability of fully answering your question.

It was the first time I tried to compile the project and it failed. I took a quick look and made this "hotfix" so I could just start playing around the code base.

It fully compiled with this change ("Build Succeeded" on XCode).
Either way, I'm dropping this PR because of #12749 

Thanks for your inputs!

PR #12799, opened 2020-08-28
Refactor window offset arrays into regular arrays

AaronVanGeffen:
In #12749, explicit typing was added to certain `std::array`s in `src/openrct2-ui/interface/Window.cpp`, used for computing window offsets. Due to how `std::array`s work, each of the elements has to be constructed individually. This leads to invoking the `ScreenCoordsXY` constructor explicitly, with ample repetition.

Regular arrays do not suffer from this. In this PR, propose we change to using regular arrays for readability where an `std` container is not strictly required.

If we do switch, we should make note of this in the coding guidelines.

### DIFF
--- a/src/openrct2-ui/interface/Window.cpp
+++ b/src/openrct2-ui/interface/Window.cpp
@@ -194,11 +194,11 @@ rct_window* window_create_auto_pos(
     auto screenHeight = uiContext->GetHeight();
 
     // Place window in an empty corner of the screen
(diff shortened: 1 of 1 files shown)


Gymnasiast:
I'm in favour.

We should perhaps make a note of the circumstances where an std::array _is_ required.

duncanspumpkin:
Much like @tupaschoal i do prefer std array but in this case the none array version is more elegant

Gymnasiast (approved):

Issue #8000, opened 2018-09-17
macOS - ICU libraries were not found


CatFlyFlyFly:
**OS:** macOS 
**Version:** 10.13.6
**Commit/Build:** c7632f29a

When trying to build on macOS using Cmake, this happened

```
-- Unsupported CXXFLAG: -Wsuggest-final-types
-- Unsupported CXXFLAG: -Wsuggest-final-methods
-- Unsupported CXXFLAG: -Wsuggest-override
-- Unsupported CXXFLAG: -Wduplicated-cond
-- Unsupported CXXFLAG: -Wduplicated-branches
-- Unsupported CXXFLAG: -Wrestrict
-- Unsupported CXXFLAG: -Wlogical-op
-- Unsupported CXXFLAG: -Wtautological-unsigned-zero-compare
No discord-rpc detected, to enable clone discord-rpc to root directory: /Users/catfly/Library/OpenRCT2-Make
-- The following ICU libraries were not found:
--   uc (required)
CMake Error at /usr/local/Cellar/cmake/3.12.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Failed to find all ICU components (missing: ICU_LIBRARY
  _ICU_REQUIRED_LIBS_FOUND) (found suitable version "62.1", minimum required
  is "59.0")
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.12.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.12.2/share/cmake/Modules/FindICU.cmake:317 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  src/openrct2/CMakeLists.txt:115 (find_package)
  CMakeLists.txt:207 (include)


-- Configuring incomplete, errors occurred!
See also "/Users/catfly/Library/OpenRCT2-Make/build/CMakeFiles/CMakeOutput.log".
See also "/Users/catfly/Library/OpenRCT2-Make/build/CMakeFiles/CMakeError.log".
```

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Specific to multiplayer?

**Steps to reproduce:**
[...]

PR #12801, opened 2020-08-28
Fix #12800: Update readme.md with cmake for MacOS 

### LINKED ISSUE #12800: Add "ICU libraries were not found" solution to readme.md
**OS:** macOS High Sierra
**Version:** 10.13.6

Following the cmake instructions described in readme.md is not enought to properly configure the environment via cmake. The solution for the problem can be easily seen on #8000, so I believe it would be a good thing to add it to the readme.md file.

**Steps to reproduce:**
1. Follow the cmake compilation steps from MacOs session in readme.md

gguedesaz:
This is a seemly common issue, already fixed by #8000. Is the wording good enough? Also, is there any other place where this information fits better (any troubleshooting page, etc...)?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -155,6 +155,10 @@ cmake ..
 make
 ln -s ../data data
 ```
+If cmake fails with the message "ICU libraries were not found":
+```
+export CMAKE_PREFIX_PATH=/usr/local/opt/icu4c
+```
 Then you can run the game by running `./openrct2`.
 
 ### Linux:


AaronVanGeffen:
This is a good point, but we deliberately keep the readme very general and concise. Your instructions are assuming ICU has been installed through Homebrew. That may not apply to other people's set-ups.

I think the wiki is a more appropriate place for this kind of troubleshooting, to be honest. I notice it has already been included in the [Building OpenRCT2 on macOS using CMake](https://github.com/OpenRCT2/OpenRCT2/wiki/Building-OpenRCT2-on-macOS-using-CMake#problem-5-cmake-cannot-find-the-required-icu-library) lemma.

Perhaps we should link more clearly to the wiki in the readme, but that's another issue.

gguedesaz:
Cool, I totally agree with you.
How about adding the link to the lemma in the readme? Something similar to it was already done in the Linux part.

AaronVanGeffen:
Yeah, that sounds like a good idea. Could you use the lemma title as the caption for the link? (I notice the same hasn't been done for the Linux part, so feel free to address it there, too.)

PR #12806, opened 2020-08-29
Add Esperanto diacritics

tellovishous:
I've created the Esperanto diacritics for the RCT fonts to use with the translation me and someone else are working on. I did my best to stay true to the original fonts.


Gymnasiast:
There are two problems with the PNG files:
- They often contain empty rows of pixels - they need to be cropped
- They are not indexed. Please index them all with the palette from the one of the existing characters.

tellovishous:
Okay I can crop them, but I thought I used the correct pallette for indexing. Is there something else I missed?

Gymnasiast:
Besides these two things: no.

> I thought I used the correct pallette for indexing

You didn't index them at all. You took _colours_ from the palette used on other letters, but didn't make them indexed PNGs. If you apply the palette using GIMP or some other tool, it should recognise the colours you used and convert them to the correct indices.

tellovishous:
Oooooh okay, my mistake, I understand now. I'll fix the letters to be properly cropped/indexed.

Gymnasiast:
For easier testing:


Ĉ ĉ Ĝ ĝ Ĥ ĥ Ĵ ĵ Ŝ ŝ Ŭ ŭ

Gymnasiast:
The indexing is not correct.

This is how it should be (taken from another characters):
afbeelding

This is what you did:
afbeelding

Like I said in my previous comment, you should have applied the palette from one of the existing characters, keeping the colours. Instead, you seem to have let it generate a new palette, which unsurprisingly has picked different indices from the existing ones. (Note that the colours are not relevant to the game - OpenRCT2 _only_ looks at the indices.)

tellovishous:
Ooooh sorry again, didn't realize they needed to be indexed in order. I'll try again.

Gymnasiast:
It works now: :+1:

afbeelding

Personally, I would have made the diacritics a bit smaller and the letter bigger (like I did for the other letters), but your approach (taking cues from the accented letters from vanilla) is not _wrong_.

I have now pushed a commit that adds drawing code for them, so you can test it yourself.

Gymnasiast:
Let me know when you're happy with how the letters look. The `ĥ` could use a bit of adjustment still, I'd say.

tellovishous:
Dankon! And yes there may be room for some improvements, I might consider looking into refinements as the translation progresses. Nevertheless I think this looks pretty good for now, will test it out myself soon.

tellovishous:
Okay so after some trial & error, I think I FINALLY got it all looking good and working properly! What do ya think?
Clipboard

Gymnasiast:
Two things:
- The bold and small `Ĵ` has a pixel too much spare room on the right: 
afbeelding
- The breve on the bold `ŭ` is a bit on the large side. (This might still be the best shape.)

tellovishous:
Yeah, the Ĵ letters have been a bit tricky to get right, I might try another tweak or two on them.
As for the bold ŭ, perhaps it's a tad big but yeah, I feel like it's the best shape for the character.

tellovishous:
The Ĵ letters look better now 
image
I guess that should be it then!

Gymnasiast:
Never merge to a branch again. I had squashed all your commits into one and rebased it against `develop`, and your merge introduced every single commit of yours again _on top of the squashed one_.

Gymnasiast (approved):

tellovishous:
Ooooh I'm sorry, my bad
Still new to dev stuff like this
Anyway, thanks for the help!

PR #12808, opened 2020-08-29
Part of #11159: Refactor on StringBuilder.hpp

gguedesaz:
## Overall Idea
Use RAII to prevent dangerous operations with raw pointers.

## What was done
As far as I understood, `StringBuilder` is a "glorified" `std::stringstream`. At first, I thought it would be possible to directly port to the standard solution, but in a matter of fact, `StringBuilder` does some low-level optimizations that are not made by `std::stringstream`: If a 32bit integer is appended to the builder, it would actually try to break it into (at most 4) 8bit chunks. That said, I just changed how the memory layout of this object looks, using a `std::vector<utf8>`.

### Edit
To handle strings more gently (and also fix some functional issues), I have changed the underlying implementation from `std::vector<utf8>` to `std::basic_string<utf8>` (a.k.a `std::string`).

## For the long term
Check the usage of allocations made by the `Memory::` namespace. Taking a quick look, I was able to find many opportunities to refactor data structures that are allocated using its functions.

### DIFF
--- a/src/openrct2/core/StringBuilder.hpp
+++ b/src/openrct2/core/StringBuilder.hpp
@@ -95,44 +95,6 @@ public:
     }
 
     /**
-     * Like Clear, only will guarantee freeing of the underlying buffer.
-     */
-    void Reset()
-    {
-        _length = 0;
-        _capacity = 0;
-        SafeFree(_buffer);
-    }
-
-    /**
-     * Resets the StringBuilder and returns the working buffer (resized to the string size).
-     */
-    utf8* StealString()
-    {
-        utf8* result = _buffer;
-        result = Memory::ReallocateArray<utf8>(result, _length + 1);
-        result[_length] = 0;
-
-        _length = 0;
-        _capacity = 0;
-        _buffer = nullptr;
-
-        return result;
-    }
-
-    /**
-     * Returns the current string buffer as a new fire-and-forget string.
-     */
-    utf8* GetString() const
-    {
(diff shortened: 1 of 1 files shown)


src/openrct2/core/StringBuilder.hpp
```diff
@@ -45,10 +42,9 @@ class StringBuilder final
     void Append(codepoint_t codepoint)
     {
         size_t codepointLength = String::GetCodepointLength(codepoint);
-        EnsureCapacity(_length + codepointLength + 1);
-        String::WriteCodepoint(_buffer + _length, codepoint);
-        _length += codepointLength;
-        _buffer[_length] = 0;
+        std::vector<utf8> data(codepointLength);
```
tupaschoal, line 45: Shouldn't this one be `std::basic_string`?

gguedesaz: It could be, but it is not mandatory:

The thing is that `String::WriteCodepoint` implementation will assign utf8 values for data()[0], ..., data()[3], so the "string" semantic is not needed here.

If you think it is more intuitive to use the basic_string, I can make the change.

tupaschoal: Yeah, I think it's better, because it conveys the meaning of "parts of a whole" better than vector, does it not? Vector kind of tells that the elements are correlated, but maybe not really connected to each other.

PR #12810, opened 2020-08-29
Create an enum class serializer

duncanspumpkin:
As part of the entity refactor there is going to be a need for using enums in the serializer traits so this adds that support.
I'm not sure what I should do with the naming. Ideally I wouldn't have to add a suffix but I can't get the SIFNAE to work otherwise. I believe in C++20 this can be improved to use concepts but as we aren't using that atm it will have to wait.
I still need to think how to change the visitor.

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -258,14 +258,14 @@ template<> struct DataSerializerTraits_t<OpenRCT2::MemoryStream>
 {
     static void encode(OpenRCT2::IStream* stream, const OpenRCT2::MemoryStream& val)
     {
-        DataSerializerTraits_t<uint32_t> s;
+        DataSerializerTraits<uint32_t> s;
         s.encode(stream, val.GetLength());
 
         stream->Write(val.GetData(), val.GetLength());
     }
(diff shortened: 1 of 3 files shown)


src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -28,13 +28,37 @@
...
+template<typename T> struct DataSerializerTraits_t
 {
     static void encode(OpenRCT2::IStream* stream, const T& v) = delete;
     static void decode(OpenRCT2::IStream* stream, T& val) = delete;
     static void log(OpenRCT2::IStream* stream, const T& val) = delete;
 };
 
+template<typename T> struct DataSerializerTraits_enum
+{
+    using underlying = std::underlying_type_t<T>;
```
tupaschoal, line 40: Reduce the scope of this to `log`, as it's the only function using it.

tupaschoal (approved):
This looks fine to me as is. Are there any other things you plan to do on this PR still? If not, I'm ok with merging

duncanspumpkin:
No I think this is everything that I was planning on doing. @IntelOrca is this fine how I've changed the visitor?

IntelOrca (approved):

PR #12815, opened 2020-08-29
Part of #11195: Refactor TitleSequence to use RAII model

gguedesaz:
This changeset adds a smart pointer to manage TitleSequence's storage.

There is one specific aspect that I'd like an opinion from the reviewers:
The `FreeTitleSequence` function could be implemented in two distinct ways:

1. Receives a non-const reference to the `std::unique_ptr<TitleSequence>`, and it frees the memory calling `::reset()`.
2. Receive a copy of the unique_ptr. Caller should explicitly move it:  `FreeTitleSequence(std::move(_sequence))`.

I arbitrarily chose the first option, but I'm good with both (or even a third one I haven`t envision). Please let me know your thoughts about it.

Thanks a lot.

PS.: Is there any tool (ASAN, Valgrind, etc...) to be sure I'm not messing up with memory management here?

### DIFF
--- a/src/openrct2-ui/windows/TitleEditor.cpp
+++ b/src/openrct2-ui/windows/TitleEditor.cpp
@@ -269,7 +269,7 @@ static void window_title_editor_close(rct_window* w)
     // Close the related windows
     window_close_by_class(WC_TITLE_COMMAND_EDITOR);
 
-    FreeTitleSequence(_editingTitleSequence);
+    _editingTitleSequence.reset();
     _sequenceName = nullptr;
 
     SafeFree(_renameSavePath);
@@ -335,7 +335,7 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
             {
                 if (w->selected_list_item != -1)
                 {
-                    TitleSequenceRemovePark(_editingTitleSequence.get(), w->selected_list_item);
+                    _editingTitleSequence->RemovePark(w->selected_list_item);
                     if (w->selected_list_item >= static_cast<int16_t>(_editingTitleSequence->NumSaves()))
                     {
                         w->selected_list_item--;
@@ -357,7 +357,7 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
         case WIDX_TITLE_EDITOR_LOAD_SAVE:
             if (w->selected_list_item >= 0 && w->selected_list_item < static_cast<int16_t>(_editingTitleSequence->NumSaves()))
             {
-                auto handle = TitleSequenceGetParkHandle(_editingTitleSequence.get(), w->selected_list_item);
+                auto handle = _editingTitleSequence->GetParkHandle(w->selected_list_item);
                 auto stream = static_cast<OpenRCT2::IStream*>(handle->Stream);
                 auto hintPath = String::ToStd(handle->HintPath);
                 bool isScenario = ParkImporter::ExtensionIsScenario(hintPath);
@@ -374,7 +374,7 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
                     else
                         game_load_init();
 
-                    TitleSequenceCloseParkHandle(handle);
+                    _editingTitleSequence->CloseParkHandle(handle);
                     window_title_editor_open(WINDOW_TITLE_EDITOR_TAB_SAVES);
                 }
                 catch (const std::exception&)
@@ -414,12 +414,12 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
                     && w->selected_list_item < static_cast<int16_t>(_editingTitleSequence->NumCommands()))
                 {
                     _editingTitleSequence->Commands.erase(_editingTitleSequence->Commands.begin() + w->selected_list_item);
-                    
+
                     if (w->selected_list_item >= static_cast<int16_t>(_editingTitleSequence->NumCommands()))
                     {
                         w->selected_list_item--;
                     }
-                    TitleSequenceSave(_editingTitleSequence.get());
+                    _editingTitleSequence->Save();
                 }
             }
             break;
@@ -441,11 +441,9 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
                 if (w->selected_list_item != -1
                     && w->selected_list_item < static_cast<int16_t>(_editingTitleSequence->NumCommands() - 1))
                 {
-                    TitleCommand& a = _editingTitleSequence->Commands[w->selected_list_item];
-                    TitleCommand& b = _editingTitleSequence->Commands[w->selected_list_item + 1];
-                    std::swap(a, b);
+                    _editingTitleSequence->SwapConsecutiveCommands(w->selected_list_item);
                     w->selected_list_item++;
-                    TitleSequenceSave(_editingTitleSequence.get());
+                    _editingTitleSequence->Save();
                 }
             }
             break;
@@ -455,11 +453,9 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
                 if (w->selected_list_item > 0
                     && w->selected_list_item < static_cast<int16_t>(_editingTitleSequence->NumCommands()))
                 {
-                    TitleCommand& a = _editingTitleSequence->Commands[w->selected_list_item - 1];
-                    TitleCommand& b = _editingTitleSequence->Commands[w->selected_list_item];
-                    std::swap(a, b);
+                    _editingTitleSequence->SwapConsecutiveCommands(w->selected_list_item - 1);
                     w->selected_list_item--;
-                    TitleSequenceSave(_editingTitleSequence.get());
+                    _editingTitleSequence->Save();
                 }
             }
             break;
@@ -865,7 +861,8 @@ static void window_title_editor_scrollpaint_saves(rct_window* w, rct_drawpixelin
     if (_editingTitleSequence == nullptr)
         return;
 
-    for (int32_t i = 0; i < static_cast<int32_t>(_editingTitleSequence->NumSaves()); i++, screenCoords.y += SCROLLABLE_ROW_HEIGHT)
+    for (int32_t i = 0; i < static_cast<int32_t>(_editingTitleSequence->NumSaves());
+         i++, screenCoords.y += SCROLLABLE_ROW_HEIGHT)
     {
         bool selected = false;
         bool hover = false;
@@ -1061,22 +1058,22 @@ static void window_title_editor_load_sequence(size_t index)
     if (index >= title_sequence_manager_get_count())
         return;
 
-    const char* path = title_sequence_manager_get_path(index);
-    auto titleSequence = LoadTitleSequence(path);
-    if (titleSequence == nullptr)
+    try
+    {
+        const char* path = title_sequence_manager_get_path(index);
+        auto titleSequence = std::make_unique<TitleSequence>(path);
+        _selectedTitleSequence = index;
+        size_t predefinedIndex = title_sequence_manager_get_predefined_index(index);
+        _isSequenceReadOnly = (predefinedIndex != SIZE_MAX);
+        _sequenceName = title_sequence_manager_get_name(index);
+        _editingTitleSequence = std::move(titleSequence);
+        window_close_by_class(WC_TITLE_COMMAND_EDITOR);
+    }
+    catch (std::exception&)
     {
         context_show_error(STR_FAILED_TO_LOAD_FILE_CONTAINS_INVALID_DATA, STR_NONE);
         return;
     }
-
-    _selectedTitleSequence = index;
-    size_t predefinedIndex = title_sequence_manager_get_predefined_index(index);
-    _isSequenceReadOnly = (predefinedIndex != SIZE_MAX);
-    _sequenceName = title_sequence_manager_get_name(index);
-    FreeTitleSequence(_editingTitleSequence);
-    _editingTitleSequence = std::move(titleSequence);
-
-    window_close_by_class(WC_TITLE_COMMAND_EDITOR);
 }
 
 static ITitleSequencePlayer* window_title_editor_get_player()
@@ -1102,7 +1099,7 @@ static bool window_title_editor_check_can_edit()
 
 static bool save_filename_exists(const utf8* filename)
 {
-    auto seq = _editingTitleSequence.get();
+    auto& seq = _editingTitleSequence;
     for (size_t i = 0; i < seq->NumSaves(); i++)
     {
         const utf8* savePath = seq->Saves[i].c_str();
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
The obvious improvement on this is to further refactor the TitleSequence so that it uses a RAII model and then we could remove all of the `FreeTitleSequence` calls. Do you feel that would be too much work within this PR?

gguedesaz:
The good news is that I've already done that, but I didn't want to overburden the reviewers with too much code at once.

It is actually not too much work, but changing this data structure touches many different places and I'm not really sure about the impact. For instance:

`seq->Path = String::duplicate (name)` will be a memory leak when `TitleSequence::Path` becomes a `std::string`.

This was something I found by luck, scrolling through the code. There might be other places doing similar stuff, so I need to be careful.

That said, is there any guidelines for running Valgrind/ASAN here? I feel like this is a must have if we want to refactor memory management in this project.

duncanspumpkin:
There are already a few memory leaks in the code. We don't at present have anything like Valgrind/ASAN running but yes we probably should.

gguedesaz:
Updated the PR to make TitleSequence use the RAII model.

src/openrct2/title/TitleSequence.h
```diff
@@ -38,18 +38,34 @@ struct TitleCommand
...
+    void AcquireSaves(std::vector<std::string>& saves) { _saves = std::move(saves); }
+
+  private:
+    std::string _path;
+    std::string _name;
+
+    std::vector<TitleCommand> _commands;
+    std::vector<std::string> _saves;
+
+    bool _isZip {false};
```
tupaschoal:
```suggestion
    std::string Path;
    std::string Name;

    std::vector<TitleCommand> Commands;
    std::vector<std::string> Saves;

    bool IsZip{false};
```

If you're taking the effort to make it into a class, let's also rename the parameters to be consistent with our coding style: https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style

src/openrct2/title/TitleSequence.h
```diff
@@ -40,16 +40,64 @@ struct TitleCommand
...
+    std::string _name;
 
-    size_t NumSaves;
-    utf8** Saves;
+    std::vector<TitleCommand> _commands;
+    std::vector<std::string> _saves;
 
-    bool IsZip;
+    bool _isZip{ false };
 };
```
duncanspumpkin, line 101: None of this is an invariant therefore i don't see much point making accessors. Just keep all the fields public.

src/openrct2/title/TitleSequence.cpp
```diff
@@ -89,43 +87,29 @@ TitleSequence* LoadTitleSequence(const utf8* path)
...
+    seq->SetPath(path);
+    seq->SetName(Path::GetFileNameWithoutExtension(seq->Path()));
+    seq->AcquireSaves(saves);
+    seq->AcquireCommands(commands);
+    seq->SetIsZip(isZip);
     return seq;
 }
 
-void FreeTitleSequence(TitleSequence* seq)
+void FreeTitleSequence(std::unique_ptr<TitleSequence>& seq)
```
duncanspumpkin, line 99: Just remove this function.

src/openrct2/title/TitleSequence.h
```diff
@@ -78,9 +126,9 @@ enum TITLE_SCRIPT
...
 constexpr uint8_t SAVE_INDEX_INVALID = UINT8_MAX;
 
-TitleSequence* CreateTitleSequence();
-TitleSequence* LoadTitleSequence(const utf8* path);
-void FreeTitleSequence(TitleSequence* seq);
+std::unique_ptr<TitleSequence> CreateTitleSequence();
+std::unique_ptr<TitleSequence> LoadTitleSequence(const utf8* path);
+void FreeTitleSequence(std::unique_ptr<TitleSequence>& seq);
 
 TitleSequenceParkHandle* TitleSequenceGetParkHandle(TitleSequence* seq, size_t index);
```
duncanspumpkin, line 133: This can become a member function

src/openrct2/title/TitleSequence.cpp
```diff
@@ -30,24 +30,22 @@
...
-static std::vector<TitleCommand> LegacyScriptRead(utf8* script, size_t scriptLength, std::vector<utf8*> saves);
+static std::vector<std::string> GetSaves(const utf8* path);
+static std::vector<std::string> GetSaves(IZipArchive* zip);
+static std::vector<TitleCommand> LegacyScriptRead(utf8* script, size_t scriptLength, const std::vector<std::string>& saves);
 static void LegacyScriptGetLine(OpenRCT2::IStream* stream, char* parts);
 static std::vector<uint8_t> ReadScriptFile(const utf8* path);
 static std::string LegacyScriptWrite(TitleSequence* seq);
 
-TitleSequence* CreateTitleSequence()
+std::unique_ptr<TitleSequence> CreateTitleSequence()
```
duncanspumpkin, line 40: I question if we need this function anymore.

src/openrct2/title/TitleSequence.cpp
```diff
@@ -30,24 +30,22 @@
...
 {
-    TitleSequence* seq = Memory::Allocate<TitleSequence>();
-    *seq = {};
-    return seq;
+    return std::make_unique<TitleSequence>();
 }
 
-TitleSequence* LoadTitleSequence(const utf8* path)
+std::unique_ptr<TitleSequence> LoadTitleSequence(const utf8* path)
```
duncanspumpkin, line 45: This can become a member function. It's basically a constructor that can fail.

src/openrct2-ui/windows/TitleEditor.cpp
```diff
@@ -336,8 +335,8 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
             {
                 if (w->selected_list_item != -1)
                 {
-                    TitleSequenceRemovePark(_editingTitleSequence, w->selected_list_item);
-                    if (w->selected_list_item >= static_cast<int16_t>(_editingTitleSequence->NumSaves))
+                    TitleSequenceRemovePark(_editingTitleSequence.get(), w->selected_list_item);
```
duncanspumpkin, line 338: Make this a member function

src/openrct2-ui/windows/TitleEditor.cpp
```diff
@@ -392,47 +391,43 @@ static void window_title_editor_mouseup(rct_window* w, rct_widgetindex widgetInd
...
-                    _editingTitleSequence->NumCommands--;
-                    if (w->selected_list_item >= static_cast<int16_t>(_editingTitleSequence->NumCommands))
+                    _editingTitleSequence->Commands().erase(_editingTitleSequence->Commands().begin() + w->selected_list_item);
+
+                    if (w->selected_list_item >= static_cast<int16_t>(_editingTitleSequence->NumCommands()))
                     {
                         w->selected_list_item--;
                     }
-                    TitleSequenceSave(_editingTitleSequence);
+                    TitleSequenceSave(_editingTitleSequence.get());
```
duncanspumpkin, line 422: Make member function

src/openrct2-ui/windows/TitleEditor.cpp
```diff
@@ -1140,7 +1132,7 @@ static void window_title_editor_add_park_callback(int32_t result, const utf8* pa
         return;
     }
 
-    TitleSequenceAddPark(_editingTitleSequence, path, filename);
+    TitleSequenceAddPark(_editingTitleSequence.get(), path, filename);
```
duncanspumpkin, line 1135: Make member function

src/openrct2/title/TitleSequence.cpp
```diff
@@ -260,15 +241,15 @@ bool TitleSequenceAddPark(TitleSequence* seq, const utf8* path, const utf8* name
 
 bool TitleSequenceRenamePark(TitleSequence* seq, size_t index, const utf8* name)
```
duncanspumpkin, line 242: Make member function

duncanspumpkin (commented):
A lot of these functions should become member functions it will simplify the code. Also we no longer require a free function as everything is handled correctly by the destructors and it's not like we need the memory as this isn't going to be huge.

src/openrct2/title/TitleSequence.h
tupaschoal: He converted the `struct` to a `class`, so it kind of makes sense to make those `private` according to the core guidelines: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-class

Do you suggest also converting it back to a `struct`?

(Personally I like it being a class, specially with all the methods that are about to become member functions)

duncanspumpkin: See what it looks like after making more of these member functions. I don't think it needs all of these assessors though.

src/openrct2/title/TitleSequence.h
gguedesaz: I went with the CoreGuildelines and kept it as a struct:
`C.2: Use class if the class has an invariant; use struct if the data members can vary independently`

Which was the case here: (Path, Name, Saves, Commands and IsZip are all independent).

src/openrct2/title/TitleSequence.cpp
gguedesaz: I 'changed' it for an empty constructor.

src/openrct2/title/TitleSequence.h
gguedesaz: See comment:
https://github.com/OpenRCT2/OpenRCT2/pull/12815#discussion_r481605081

duncanspumpkin (commented):
Starting to look really good

gguedesaz:
Hey @tupaschoal and @duncanspumpkin,

I just pushed a commit moving the TitleSequenceParkHandle to use the RAII model. With that, I want to finalize the job done in this PR (so it doesn't become too big). Please let me know if you guys think further refactor is needed here.

Thank you

src/openrct2/title/TitleSequence.cpp
```diff
@@ -202,22 +196,19 @@ bool TitleSequenceAddPark(TitleSequence* seq, const utf8* path, const utf8* name
...
     if (index == SIZE_MAX)
     {
-        seq->Saves = Memory::ReallocateArray(seq->Saves, seq->NumSaves + 1);
-        Guard::Assert(seq->Saves != nullptr, GUARD_LINE);
-        index = seq->NumSaves;
-        seq->NumSaves++;
+        index = seq->NumSaves();
     }
-    seq->Saves[index] = String::Duplicate(name);
+    seq->Saves.insert(seq->Saves.begin() + index, std::string(name));
```
tupaschoal: Not sure this is doing the same thing as before.

Previously it searched for an index to replace or increased the vector by one and used the last one if the vector was full. Now it is always adding a new element at index, which is alright for `SIZE_MAX`, but does add a duplicate when `index = i`, does it not?

(This is in commit https://github.com/OpenRCT2/OpenRCT2/pull/12815/commits/946a8fd52fd8899e373f1c63f11bd802014151c5)

src/openrct2/title/TitleSequence.cpp
```diff
@@ -28,25 +28,20 @@
...
-    return std::make_unique<TitleSequence>();
-}
-
-std::unique_ptr<TitleSequence> LoadTitleSequence(const utf8* path)
+TitleSequence::TitleSequence(const utf8* path)
 {
     std::vector<uint8_t> script;
     std::vector<std::string> saves;
-    bool isZip;
+    bool is_zip = false;
```
tupaschoal:
```suggestion
    bool isZip = false;
```
Use camelCase for local variables (won't conflict with the member variable because that is TitleCase)

(This is on commit https://github.com/OpenRCT2/OpenRCT2/pull/12815/commits/b0f7e8d558e04b4736b41257400ceab4cd73a38b)

src/openrct2/title/TitleSequence.cpp
```diff
@@ -168,20 +152,27 @@ void TitleSequenceCloseParkHandle(TitleSequenceParkHandle* handle)
     }
 }
 
-bool TitleSequenceSave(TitleSequence* seq)
+void TitleSequence::SwapConsecutiveCommands(size_t index)
+{
+    TitleCommand* first = &Commands[index];
+    TitleCommand* second = &Commands[index + 1];
+    std::swap(first, second);
```
tupaschoal:
```suggestion
    assert(Commands.size() - 1 >= index + 1);
    if (Commands.size() - 1 >= index + 1)
    {
        TitleCommand* first = &Commands[index];
        TitleCommand* second = &Commands[index + 1];
        std::swap(first, second);
    }
```
We should add an assertion and if here to make sure no one tries to access the vector out of bounds.

(This is on commit https://github.com/OpenRCT2/OpenRCT2/pull/12815/commits/b0f7e8d558e04b4736b41257400ceab4cd73a38b)

src/openrct2/title/TitleSequence.h
```diff
@@ -38,24 +38,45 @@ struct TitleCommand
...
+        return Commands.size();
+    }
+
     std::string Path;
     std::string Name;
 
     std::vector<TitleCommand> Commands;
-    size_t NumCommands() const { return Commands.size(); }
-
     std::vector<std::string> Saves;
```
tupaschoal: Maybe you want to swap the order of the vectors and strings to make it in descending size order? 

(this is on commit https://github.com/OpenRCT2/OpenRCT2/pull/12815/commits/b0f7e8d558e04b4736b41257400ceab4cd73a38b)

src/openrct2/title/TitleSequence.h
```diff
@@ -38,24 +38,45 @@ struct TitleCommand
...
-    size_t NumSaves() const { return Saves.size(); }
 
-    bool IsZip;
-};
-
-struct TitleSequenceParkHandle
-{
-    const utf8* HintPath;
-    void* Stream;
+    bool IsZip = false;
```
tupaschoal, line 82:
```suggestion
    bool IsZip {false};
```

Maybe use uniform initialization?

(This is on commit https://github.com/OpenRCT2/OpenRCT2/pull/12815/commits/b0f7e8d558e04b4736b41257400ceab4cd73a38b)

tupaschoal (commented):
This is looking great already, thanks.

I still find it weird to have some of the initialization of the member variables after the object is constructed and not within its ctor. I understand some might not be initialized at all times, but if there are some that do, I'd say these should be a ctor parameter. Let me know if this makes sense and whether you think it's doable in this PR or you want to tackle in another

tupaschoal:
@gguedesaz I'm moving this to draft, for lack of activity. If you resume, please mark as ready to review

AaronVanGeffen:
Has this been superseded by #13011?

tupaschoal:
It looks like, I asked @ZehMatt to confirm and he did

ZehMatt:
@gguedesaz I have to apologize, I did not see this PR and starting hacking into the same code as you did. Given the inactivity I will close this PR, I've added you as co-author in my PR to all the commits.

gguedesaz:
Hey Guys!

Sorry for the inactivity, had some personal issues and I couldn't touch the code for almost a month.
@ZehMatt , no need to apologize, thanks for the contribution. I'll find another place to modernize when I feel it is the time.

Thanks!

ZehMatt:
I left out the part of adding methods, you can still do that part if you like.

PR #12823, opened 2020-08-30
Move update notification to top-right corner

AaronVanGeffen:
_Just_ before the last release, in #12347, an update notification mechanism was introduced to OpenRCT2. To draw attention to the notification, it was placed on top of the title menu buttons. However, there has been discussion about whether this is too prominent. In this PR, I therefore propose to move it to the top-right corner, where it can sit next to the 'Options' button.

### v0.3.0:
Haunted Harbour 2020-08-30 20-24-56

### This PR:
Haunted Harbour 2020-08-30 20-19-45

### DIFF
--- a/src/openrct2-ui/windows/TitleMenu.cpp
+++ b/src/openrct2-ui/windows/TitleMenu.cpp
@@ -31,16 +31,13 @@ enum {
     WIDX_NEW_VERSION,
 };
 
-static ScreenRect _filterRect;
 static constexpr ScreenSize MenuButtonDims = { 82, 82 };
-static constexpr ScreenSize UpdateButtonDims = { MenuButtonDims.width * 4, 28 };
 
 static rct_widget window_title_menu_widgets[] = {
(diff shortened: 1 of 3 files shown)


janisozaur:
I liked it in my place and wanted it so prominent partially because of how many "don't know if my OpenRCT2 is up to date" responses in the survey.

If I'm overruled, that's fine. @AaronVanGeffen I guess as you implemented that change you prefer it top right as well?

AaronVanGeffen:
Hmm, I don't think the placement was _bad_ last time, but I do prefer the UI to be a bit more subtle, yes. I didn't want to hold back the release to make these changes last time, though.

Personally, I think the orange stands out enough to grab attention, whether it's front-and-centre or in the corner.

tupaschoal:
Hmm, I'm not sure. Orange does stand out, on the other hand, most people play the release, so nudging them to upgrade (in a bit more intrusive manner) might be better. I think I'm still leaning towards the big button

marijnvdwerf:
In case you find it too prominent, I think it would make more sense to move it to the bottom right near the version information. But since you want people to update ASAP, I think the prominent location is better, since it's near the buttons that people will press 'instantly' after opening the name.

AaronVanGeffen:
Thanks for the input. Let's leave the update button as-is, then.

PR #12832, opened 2020-09-01
Fix #12818 - ignore ride price in free-rides parks

### LINKED ISSUE #12818: ScRide.hpp doesn't prevent plugins changing the price of free rides in pay to enter park scenarios
**OS:** Windows 10
**Version:** 0.3.0-93-g26def8716 (26def87 on develop)

ScRide.hpp doesn't prevent plugins (such as Ride Price Manager) changing the price of free rides in pay to enter park scenarios. Though guests don't pay anything and the UI still claims the price for each ride is free, even a $1 entrance fee will generate "Your park entrance fee is too high!" messages!

The maximum park entrance fee that guests consider reasonable is calculated by Park::CalculateTotalRideValueForMoney, which values each open non-broken ride at (ride.value - ride.price[0]) * 2, if the value is greater than the price. Since the plugin puts the price at greater than the ride's value, all the rides end up generating $0 in this calculation!

Note that reloading the save with all plugins disabled doesn't fix anything.
**Save game:**
[Gentleman's Garden.txt](https://github.com/OpenRCT2/OpenRCT2/files/5146369/Gentleman.s.Garden.txt)
Scenario file: [Gentlemans Garden - Xedoh - Beginner.txt](https://github.com/OpenRCT2/OpenRCT2/files/5146456/Gentlemans.Garden.-.Xedoh.-.Beginner.txt)

chess123mate:
This change fixes the reason I opened #12818 by ensuring that, when calculating `CalculateTotalRideValueForMoney` (used in determining a fair entrance fee and in some awards), the ride prices are ignored if the park isn't pay-for-rides.

It occurs to me that the code could be made even more efficient by having two separate loops (so that checking whether ride prices should be considered would only be performed once), but I am assuming that copy/pasting the for loop would not be ideal. The for loop already is already in two places (along with a TODO to merge them). I have two ideas on how to improve this, though I don't know enough C++ to do it at the moment:

1. Some sort of accumulator function that receives a function pointer (and my understanding is that this is less efficeint)
2. RideManager could provide a different iterator object that only iterates over open and functioning rides

Or is the TODO just complaining that all these checks are happening 2x when the functions could be merged?

Side note: I find it a bit strange that `park_ride_prices_unlocked()` doesn't check if the park uses money or not. All the other places it's used also perform the NO_MONEY check, except for `marketing_is_campaign_type_applicable` - but this turns out to not matter in practice because there is a check for this in the UI code (though this sounds like the wrong place to check).

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -10,6 +10,7 @@
 - Fix: [#12701] Silent NSIS setup flag /S isn't silent, upgrade pop-up appears anyway.
 - Fix: [#12737] Space Rings draw the same vehicle 4 times.
 - Fix: [#12764] Rides don't start aged anymore.
+- Fix: [#12818] Ride price not ignored in free-rides parks.
 - Fix: [#12820] Title menu buttons not invalidating properly
 - Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
 - Fix: [#12857] Incorrect Peep thoughts in imported RCT1 parks.
--- a/src/openrct2/network/NetworkBase.cpp
+++ b/src/openrct2/network/NetworkBase.cpp
@@ -33,7 +33,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "3"
+#define NETWORK_STREAM_VERSION "4"
 #define NETWORK_STREAM_ID OPENRCT2_VERSION "-" NETWORK_STREAM_VERSION
 
 static Peep* _pickup_peep = nullptr;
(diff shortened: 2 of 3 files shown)


tupaschoal (approved):
It looks alright to me. I can't comment on the to-do cause it's a bit difficult to see it on the cellphone

tupaschoal (commented):
This needs a changelog entry

duncanspumpkin (commented):
Increment network version

tupaschoal:
I've added the changelog entry and bumped the network version, should be mergeable once checks are green

tupaschoal (approved):

chess123mate:
> I've added the changelog entry and bumped the network version, should be mergeable once checks are green

Thanks!

PR #12837, opened 2020-09-01
Windows: FriBidi for Bidi and shaping

Gymnasiast:
Fribidi should be a more palatable dependency on Windows.

I would like to have some help with handling the Windows and macOS side of things.


janisozaur:
Can you provide some details of the features for the two? Why can't we just keep using the current library? It's already supported on Windows - at least in terms of compilation in CI + vcpkg.

Gymnasiast:
FriBidi handles *just* bidirectional text and shaping, while ICU does a lot more.

We used to use ICU on Windows too, but @IntelOrca removed it at some point. It's probably best if he explains why himself.

AaronVanGeffen:
> Personally, I don't have a problem with ICU, but @IntelOrca wanted it gone from Windows (we used to use it on Windows too) and removed it. It's probably best if he explains why himself.

I think we never used ICU on Windows. When I introduced ICU for Linux and macOS, @IntelOrca implemented equivalent platform routines using Windows APIs.

Gymnasiast:
After discussion in the team chat, I have modified the code to use the new FriBidi implementation on Windows and the old ICU implementation on all other platforms.

This will require someone to add FriBidi to the Windows dependencies bundle and to test the implementation. @IntelOrca could you do that?

IntelOrca:
On discord, I mentioned another suggestion of doing something similar to curl and crypt where you put files in the core directory: `rte.h`, `rte.icu.cpp` and `rte.fribidi.cpp`.

What do you think?

Gymnasiast:
That should be `RTL.h`, `RTL.FriBidi.cpp` and `RTL.ICU.cpp`, I'd say.

But yeah, that sounds fine. I can change it if just moving the `FixRTL` function is enough. If you want it fancier, it's probably better if you handle it.

janisozaur (dismissed):
There needs to be a way to turn fribidi off entirely.

See https://github.com/microsoft/vcpkg/issues/11248 / https://github.com/microsoft/vcpkg/issues/11230

IntelOrca:
I'll wait for @Gymnasiast's changes to go to RTE.h etc then do the guard properly.

src/openrct2/localisation/LanguagePack.cpp
```diff
@@ -651,7 +659,29 @@ class LanguagePack final : public ILanguagePack
     std::string FixRTL(std::string& input)
     {
 #ifdef _WIN32
-        return input;
+        FriBidiChar utf32String[1024] = { 0 };
```
AaronVanGeffen, line 662: Can we use a (local) constant for this `1024`? It's used several times throughout this function.

src/openrct2/localisation/LanguagePack.cpp
```diff
@@ -651,7 +659,29 @@ class LanguagePack final : public ILanguagePack
...
+        auto len = static_cast<FriBidiStrIndex>(input.length() + 1);
+        fribidi_charset_to_unicode(FRIBIDI_CHAR_SET_UTF8, input.c_str(), len, utf32String);
+
+        FriBidiStrIndex utf32len = 0;
+        for (; utf32len < 1024; utf32len++)
```
AaronVanGeffen, line 667: Perhaps
```suggestion
        for (; utf32len < std::size(utf32String); utf32len++)
```

Gymnasiast:
OK, that _should_ have Janis', Intel's and Aaron's requests all fulfilled.

src/openrct2/localisation/RTL.FriBidi.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    char outputString[BufferLength];
+    fribidi_unicode_to_charset(FRIBIDI_CHAR_SET_UTF8, reorderedStr, len, outputString);
+
+    return std::string(outputString);
+}
+#elif defined(_WIN32)
+std::string FixRTL(std::string& input)
+{
+    log_warning("This build of OpenRCT2 cannot properly handle right-to-left text!");
```
IntelOrca, line 53: You will only want this to show once, otherwise the console will be spammed with the message continuously.

AaronVanGeffen:
My understanding is that this is pending on #12667 to be merged for updating the dependencies. Is that correct?

src/openrct2/localisation/RTL.ICU.cpp
```diff
@@ -0,0 +1,40 @@
+
```
janisozaur, line 1: missing header

Gymnasiast:
@AaronVanGeffen Both need a dependency bundle upgrade, so either that one or this one would need to be rebased after the other is merged. It is perfectly possible to merge this first, should #12667 be delayed.

AaronVanGeffen:
@Gymnasiast Perhaps, but deps bundle [version 26](https://github.com/OpenRCT2/Dependencies/releases/tag/v26) (introducing fribidi) builds on top of [version 25](https://github.com/OpenRCT2/Dependencies/releases/tag/v25) (replacing jansson with nlohmann-json). I'd really rather not fiddle with that.

I suggest we merge this in order. This isn't a critical bugfix, fortunately.

IntelOrca:
> It is perfectly possible to merge this first, should #12667 be delayed.

Unfortunately not, as @AaronVanGeffen says, jansson is no longer in the zip so it can not build this branch.

Gymnasiast:
Ah, I see.

Anyway: could you fix the MSVC bits and test it, @IntelOrca ?

openrct2.proj
```diff
@@ -37,10 +37,10 @@
   <!-- 3rd party libraries / dependencies -->
   <PropertyGroup>
     <DependenciesCheckFile>$(RootDir).dependencies</DependenciesCheckFile>
-    <LibsUrl Condition="'$(Platform)'=='Win32'">https://github.com/OpenRCT2/Dependencies/releases/download/v25/openrct2-libs-v25-x86-windows-static.zip</LibsUrl>
-    <LibsSha1 Condition="'$(Platform)'=='Win32'">c8c1a3052721c7dedde8c3cab73ed09b793a9b8e</LibsSha1>
-    <LibsUrl Condition="'$(Platform)'=='x64'">https://github.com/OpenRCT2/Dependencies/releases/download/v25/openrct2-libs-v25-x64-windows-static.zip</LibsUrl>
-    <LibsSha1 Condition="'$(Platform)'=='x64'">5384dffd21b4e87c36bb91c6882e277023285612</LibsSha1>
+    <LibsUrl Condition="'$(Platform)'=='Win32'">https://github.com/OpenRCT2/Dependencies/releases/download/v26/openrct2-libs-v21-x86-windows-static.zip</LibsUrl>
+    <LibsSha1 Condition="'$(Platform)'=='Win32'">d7445387108b70fb23ccc06b73c8bdd8a44a0706</LibsSha1>
+    <LibsUrl Condition="'$(Platform)'=='x64'">https://github.com/OpenRCT2/Dependencies/releases/download/v26/openrct2-libs-v21-x64-windows-static.zip</LibsUrl>
```
IntelOrca, line 42: the URL is wrong, it says v21 instead of v26

AaronVanGeffen:
This was tagged with the changelog tag, but the changelog hasn't been updated yet. Could you append it in a follow-up commit/PR?

Gymnasiast:
Done.

PR #12840, opened 2020-09-02
Add Park.entranceFee to the plugin API

pencil:
## Description

While looking through the code of [openrct2-ride-price-manager](https://github.com/mgovea/openrct2-ride-price-manager) I noticed the plugin has its own setting to determine if the park charges an entrance fee. It seems like it would make sense to just provide this as part of the plugin API instead.

## Test

I tested this by changing the aforementioned plugin to check for `park.entranceFee > 0` instead of [`config.getParkAdmissionEnabled()`](https://github.com/mgovea/openrct2-ride-price-manager/blob/b60aae47775fbc391a6b62a7789a5c982f8ac4b6/src/fixRidePrices.ts#L37). Additionally I tested updating the fee with this simple plugin which seemed to work as expected.

```
var main = function() {
  context.subscribe("interval.day", function () {
    park.entranceFee = date.day * 10;
  });
};

registerPlugin({
    name: 'Set Park Price',
    version: '1.0',
    authors: ['OpenRCT2'],
    type: 'remote',
    main: main
});
```

entranceFee

### DIFF
--- a/src/openrct2/scripting/ScPark.hpp
+++ b/src/openrct2/scripting/ScPark.hpp
@@ -280,11 +280,20 @@ namespace OpenRCT2::Scripting
             context_broadcast_intent(&intent);
         }
 
+        money16 entranceFee_get() const
+        {
+            return gParkEntranceFee;
+        }
+        void entranceFee_set(money16 value)
+        {
+            ThrowIfGameStateNotMutable();
+            gParkEntranceFee = value;
+        }
+
         std::string name_get() const
         {
             return GetContext()->GetGameState()->GetPark().Name;
         }
-
         void name_set(std::string value)
         {
             ThrowIfGameStateNotMutable();
@@ -383,6 +392,7 @@ namespace OpenRCT2::Scripting
             dukglue_register_property(ctx, &ScPark::rating_get, &ScPark::rating_set, "rating");
             dukglue_register_property(ctx, &ScPark::bankLoan_get, &ScPark::bankLoan_set, "bankLoan");
             dukglue_register_property(ctx, &ScPark::maxBankLoan_get, &ScPark::maxBankLoan_set, "maxBankLoan");
+            dukglue_register_property(ctx, &ScPark::entranceFee_get, &ScPark::entranceFee_set, "entranceFee");
             dukglue_register_property(ctx, &ScPark::name_get, &ScPark::name_set, "name");
             dukglue_register_property(ctx, &ScPark::messages_get, &ScPark::messages_set, "messages");
             dukglue_register_method(ctx, &ScPark::postMessage, "postMessage");
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,6 +1,7 @@
 0.3.0+ (in development)
 ------------------------------------------------------------------------
 - Feature: [#10807] Add 2x and 4x zoom levels (currently limited to OpenGL).
+- Feature: [#12840] Add Park.entranceFee to the plugin API.
 - Fix: [#400] Unable to place some saved tracks flush to the ground (original bug).
 - Fix: [#7037] Unable to save tracks starting with a sloped turn or helix.
 - Fix: [#12691] Ride graph tooltip incorrectly used count instead of number string.
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -1344,6 +1344,12 @@ declare global {
         rating: number;
         bankLoan: number;
         maxBankLoan: number;
+
+        /**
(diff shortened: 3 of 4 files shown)


src/openrct2/scripting/ScPark.hpp
```diff
@@ -280,11 +280,20 @@ namespace OpenRCT2::Scripting
...
         }
 
+        money16 entranceFee_get() const
+        {
+            return gParkEntranceFee;
+        }
+        void entranceFee_set(money16 value)
+        {
+            ThrowIfGameStateNotMutable();
+            park_set_entrance_fee(value);
```
duncanspumpkin, line 290: This is calling a game action. @intelOrca should this be instead be performed through the plugin game actions code and using the visit function.

IntelOrca: I don't have anything against having an `entranceFee_set`, it will only be able to be used when game state is mutable such as in a custom action or certain hooks, but it should not be performing a game action. It should just set the price directly.

duncanspumpkin: If it sets it directly then you'll have to duplicate all the validation code and it won't work in networked games. Or is that what the mutable protection ensures.

IntelOrca: The mutable thing already protects it from network issues, you use the game action if you want network sync. The validation can still be used, just extract that out to another function you can call.

pencil: I'll look into that. Out of curiosity: Is calling a game action from the plugin API generally not a good practice? Why so?

IntelOrca: Game actions are asynchronous, where as a plugin would expect this property to be changed and updated immediately.

pencil: Ah, I see. Other setters here either just copy the boundary checks from other places (e.g. `rating_set`) or don't perform any kind of validation/sanitization. Any suggestions for where to put a shared validation method and how to structure it? It has to be more than a single bool check because currently [`SetParkEntranceFeeAction`](https://github.com/OpenRCT2/OpenRCT2/blob/22f29bf192320ef78cabd8bc76672f1e85cad038/src/openrct2/actions/SetParkEntranceFeeAction.hpp#L47-L56) differentiates the "no entrance fee allowed" and "entrance fee too high/low" cases.

IntelOrca: Probably just `Park.cpp`

duncanspumpkin: We don't really want to get into the same issue as the ride price setting value issues so I think we still need to check for the no entrance fee allowed.

IntelOrca: But if you toggle that park flag you would probably expect it to keep the price it had before you toggled it. We just need to make sure we use the get method for everything that uses the price so we can return free if it is disabled.

pencil: Hadn't even noticed that. I should probably use `park_get_entrance_fee();` in the getter?

IntelOrca: Like I keep saying, this is supposed to represent the raw property, so I really don't think there is any need for disguise or validation.

pencil: Yeah, I guess it's a question of philosophy. If it should just allow access to the raw values (like the other fields), I'll just drop the validation for now and update `gParkEntranceFee` directly.

pencil: Updated the PR to just allow direct access to `gParkEntranceFee` like @IntelOrca suggested.

IntelOrca (approved):

duncanspumpkin (commented):
You'll need to increment the API version

pencil:
@duncanspumpkin Done 👍

tupaschoal (commented):
Please also add an entry to the changelog. PRs that are not tracked by issues go last on the changelog, see others for reference

IntelOrca:
@tupaschoal changelog is sorted by type (feature, bug) then by issue number, if there is no issue, PR number.

duncanspumpkin (approved):

tupaschoal (approved):

pencil:
@mgovea FYI this should allow Ride Price Manager to determine if the park charges an entrance fee and eliminate the need for that manual configuration in the plugin.

mgovea:
Siiiiiiiick dude. Thank you so much for the mention.

PR #12848, opened 2020-09-02
Fix #12845: Cancel marketing campaigns for demolished rides

### LINKED ISSUE #12845: Deleting ride during ad campaign creates incorrect notification
**OS:** Windows 10
**Version:** v0.3.0
**Commit/Build:** 135cc10

Deleting a roller coaster while an ad campaign for that ride is running results in an incorrect ticker message when the campaign completes. Deleting the pre-built coaster on Dynamite Dunes resulted in the message saying "Your advertising campaign for Inverted Impulse Coaster is complete." This happens in other scenarios with both pre-existing and player-built rides.

In vanilla RCT2, the game cancels the advertising campaign when the ride is deleted.

**Steps to reproduce:**
1. Open any scenario. For example, Forest Frontiers.
2. Build a roller coaster. For example, the Black Widow wooden roller coaster.
3. Start an advertising campaign for the roller coaster.
4. Delete the roller coaster. The ad campaign will not be canceled, but the ride name will be blank.
5. Fast forward to the end of the ad campaign. When it finishes, it will fill-in the missing ride in the ticker notification with a random ride type. The first time, I got "Roto-Drop;" the second time, I got "Vertical Drop Roller Coaster." Neither of these are available in Forest Frontiers.

**Screenshots / Video:**
After deleting the Black Widow coaster:
Missing Ride Name
After the ad campaign ended:
Incorrect Message

pencil:
Fixes #12845 

demolish

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -9,6 +9,7 @@
 - Fix: [#12737] Space Rings draw the same vehicle 4 times.
 - Fix: [#12764] Rides don't start aged anymore.
 - Fix: [#12820] Title menu buttons not invalidating properly
+- Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
 - Technical: The required version of macOS has been lowered to 10.13 (High Sierra).
 
--- a/src/openrct2/management/Marketing.cpp
+++ b/src/openrct2/management/Marketing.cpp
@@ -237,3 +237,18 @@ void marketing_new_campaign(const MarketingCampaign& campaign)
         gMarketingCampaigns.push_back(campaign);
     }
 }
+
+void MarketingCancelCampaignsForRide(const ride_id_t rideId)
+{
+    auto isCampaignForRideFn = [&rideId](MarketingCampaign& campaign) {
+        if (campaign.Type == ADVERTISING_CAMPAIGN_RIDE_FREE || campaign.Type == ADVERTISING_CAMPAIGN_RIDE)
+        {
+            return campaign.RideId == rideId;
+        }
+        return false;
+    };
+
+    auto& v = gMarketingCampaigns;
+    auto removedIt = std::remove_if(v.begin(), v.end(), isCampaignForRideFn);
+    v.erase(removedIt, v.end());
+}
--- a/src/openrct2/actions/RideDemolishAction.hpp
+++ b/src/openrct2/actions/RideDemolishAction.hpp
@@ -240,6 +240,8 @@ private:
             }
         }
 
+        MarketingCancelCampaignsForRide(_rideIndex);
+
         auto res = std::make_unique<GameActionResult>();
         res->Expenditure = ExpenditureType::RideConstruction;
         res->Cost = refundPrice;
--- a/src/openrct2/management/Marketing.h
+++ b/src/openrct2/management/Marketing.h
@@ -67,3 +67,4 @@ void marketing_set_guest_campaign(Peep* peep, int32_t campaign);
 bool marketing_is_campaign_type_applicable(int32_t campaignType);
 MarketingCampaign* marketing_get_campaign(int32_t campaignType);
 void marketing_new_campaign(const MarketingCampaign& campaign);
+void MarketingCancelCampaignsForRide(const ride_id_t rideId);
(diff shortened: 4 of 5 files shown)


src/openrct2/management/Marketing.cpp
```diff
@@ -237,3 +237,20 @@ void marketing_new_campaign(const MarketingCampaign& campaign)
         gMarketingCampaigns.push_back(campaign);
     }
 }
+
+void marketing_cancel_campaigns_for_ride(ride_id_t rideId)
```
duncanspumpkin, line 241: All new functions should be TitleCase. rideId should be const

duncanspumpkin (commented):
You'll also need to increment the network version

src/openrct2/management/Marketing.cpp
```diff
@@ -237,3 +237,20 @@ void marketing_new_campaign(const MarketingCampaign& campaign)
...
+{
+    for (auto it = gMarketingCampaigns.begin(); it != gMarketingCampaigns.end();)
```
gguedesaz, line 243: I think it would be less error-prone to use the [erase-remove](https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom) idiom:

```
auto isAdvertisingFn = [rideId](MarketingCampaign& campaign)
{
    if (campaign.Type == ADVERTISING_CAMPAIGN_RIDE_FREE || campaign.Type == ADVERTISING_CAMPAIGN_RIDE)
    {
        return campaign.RideId == rideId;
    }
    return false;
};

auto& v = gMarketingCampaign;
auto removedIt = std::remove_if(v.begin(), v.end(), isAdvertisingFn);
v.erase(removedIt, v.end());
```

src/openrct2/management/Marketing.cpp
Broxzier: I'ma bit torn between using and not using const for values passed by value. It clusters the code if used a lot and in some cases it makes it look like a value is const while it only makes the pointer const (`char* const c`), though it helps to avoid side-effects (accidentally using `=` instead of `==`).

I'd say we only do this in source files, the header file doesn't care whether or not the value gets changed.

duncanspumpkin: Adding more consts reduces errors and conveys more information. You can put it in both the header and the source file. I'll admit that some instances of very short functions it is a bit annoying to const the variables though this isn't one of those situations

src/openrct2/management/Marketing.cpp
pencil: Agreed, that does look a bit cleaner than my version 👍

pencil:
Hm, not sure why the build is failing on Linux i686. Probably an unrelated transient error:
>FAILED: CMakeFiles/openrct2.dir/src/openrct2-ui/CursorRepository.cpp.o 
[...]
ccache: error: /github/home/.ccache/ccache.conf: No such file or directory

janisozaur:
Yes, oddly lot of those recently @pencil

gguedesaz (approved):

src/openrct2/management/Marketing.cpp
```diff
@@ -237,3 +237,18 @@ void marketing_new_campaign(const MarketingCampaign& campaign)
         gMarketingCampaigns.push_back(campaign);
     }
 }
+
+void marketingCancelCampaignsForRide(const ride_id_t rideId)
```
tupaschoal, line 241:
```suggestion
void MarketingCancelCampaignsForRide(const ride_id_t rideId)
```

To be title case

tupaschoal (commented):
This will also require a changelog entry

duncanspumpkin (approved):

distribution/changelog.txt
```diff
@@ -9,6 +9,7 @@
 - Fix: [#12737] Space Rings draw the same vehicle 4 times.
 - Fix: [#12764] Rides don't start aged anymore.
 - Fix: [#12820] Title menu buttons not invalidating properly
+- Fix: [#12848] Deleting ride with active ad campaign creates incorrect notification.
```
tupaschoal, line 12:
```suggestion
- Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
```
Because there's an issue to track it

tupaschoal (approved):

PR #12853, opened 2020-09-04
Consistently render various height marks when zoomed in

pencil:
This PR adjusts the rendering logic for the various height markers so that they consistently render when zoomed in but not when zoomed out (due to them being hard/impossible to read).

### DIFF
--- a/src/openrct2/paint/tile_element/Paint.Surface.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Surface.cpp
@@ -1015,7 +1015,7 @@ void surface_paint(paint_session* session, uint8_t direction, uint16_t height, c
         descriptor.corner_heights.left = baseHeight + ch.left;
     }
 
-    if ((session->ViewFlags & VIEWPORT_FLAG_LAND_HEIGHTS) && (zoomLevel == 0))
+    if (PaintShouldShowHeightMarkers(session, VIEWPORT_FLAG_LAND_HEIGHTS))
     {
         const int16_t x = session->MapPosition.x;
         const int16_t y = session->MapPosition.y;
--- a/src/openrct2/paint/tile_element/Paint.Entrance.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Entrance.cpp
@@ -340,9 +340,7 @@ void entrance_paint(paint_session* session, uint8_t direction, int32_t height, c
 {
     session->InteractionType = VIEWPORT_INTERACTION_ITEM_LABEL;
 
-    rct_drawpixelinfo* dpi = &session->DPI;
-
-    if (session->ViewFlags & VIEWPORT_FLAG_PATH_HEIGHTS && dpi->zoom_level <= 0)
+    if (PaintShouldShowHeightMarkers(session, VIEWPORT_FLAG_PATH_HEIGHTS))
     {
         if (entrance_get_directions(tile_element) & 0xF)
         {
--- a/src/openrct2/paint/tile_element/Paint.Path.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Path.cpp
@@ -930,7 +930,7 @@ void path_paint(paint_session* session, uint16_t height, const TileElement* tile
         }
     }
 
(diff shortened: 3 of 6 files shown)


src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -1015,7 +1015,7 @@ void surface_paint(paint_session* session, uint8_t direction, uint16_t height, c
         descriptor.corner_heights.left = baseHeight + ch.left;
     }
 
-    if ((session->ViewFlags & VIEWPORT_FLAG_LAND_HEIGHTS) && (zoomLevel == 0))
+    if (session->ViewFlags & VIEWPORT_FLAG_LAND_HEIGHTS)
```
AaronVanGeffen, line 1018: I don't think we should be enabling them on all zoom levels. You already mentioned they aren't really readable when zoomed out, anyway. Only showing them when zoomed **in** would be best, I think:

```suggestion
    if ((session->ViewFlags & VIEWPORT_FLAG_LAND_HEIGHTS) && (zoomLevel <= 0))
```

distribution/changelog.txt
```diff
@@ -3,6 +3,7 @@
 - Feature: [#10807] Add 2x and 4x zoom levels (currently limited to OpenGL).
 - Feature: [#12712] Add TCP / socket plugin APIs.
 - Feature: [#12840] Add Park.entranceFee to the plugin API.
+- Feature: [#12853] Render "Height Marks on Land" on all zoom levels when enabled.
```
AaronVanGeffen, line 6: We would consider this a Change, not a Feature. However, if we change this PR to only affect the new 2x and 4x zoom levels, a changelog entry won't be needed.

oli414:
This happens with more stuff, note the grass growth for instance, mowed grass isn't shown on the new 2x and 4x zoom levels.

AaronVanGeffen:
@oli414 Could you file an issue for that?

src/openrct2/paint/tile_element/Paint.Surface.cpp
pencil: I adjusted this and ensured that all the other height markers follow the exact same logic.

pencil:
I adjusted the PR to only render height markers when zoomed in and applied that logic to all height markers consistently. The decision logic is now in a central `PaintShouldShowHeightMarkers` function so that it can be adjusted easily for all types.

AaronVanGeffen:
This is a nice change. It's breaking the testpaint job, though. Testpaint compares our C++ paint routines to the ones from vanilla

```
CMakeFiles/testpaint.dir/src/openrct2/ride/TrackPaint.cpp.o: In function `track_paint(paint_session*, unsigned char, int, TileElement const*)':
TrackPaint.cpp:(.text+0x3068): undefined reference to `PaintShouldShowHeightMarkers(paint_session const*, unsigned int)'
```

pencil:
Hm, I wasn't able to get testpaint working on my machine but I'll look it more later.

I'm guessing `testpaint` somehow swaps out the implementation of `Paint.cpp` for its own? If that's the case I could probably move `PaintShouldShowHeightMarkers` to a different place.

duncanspumpkin:
You only need to get test paint compiling it's output isn't useful anymore as too many of the structures have diverged.

pencil:
I got `test/testpaint` to compile locally after some tinkering
```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.15.6
BuildVersion:   19G2021
$ g++ -I ../../src -I ../../src/thirdparty -o testpaint -std=c++17 main.cpp
$ ./testpaint
Testpaint can only be properly executed on x86
```

I think most of the code is hidden behind a `#if defined(__i386__)` so probably won't be able to reproduce the build failure locally.

pencil:
Alright, I have moved the helper method and now `testpaint` is building successfully once again.

AaronVanGeffen (approved):
LGTM. Tested and everything appears to be working correctly — indeed, better than before on OpenGL, when zoomed in.

PR #12865, opened 2020-09-05
Fix #12434: Refactor ENTERTAINER_COSTUME to use strong enum 

### LINKED ISSUE #12434: Refactor ENTERTAINER_COSTUME to use strong enum

cfe316:
This is a pretty straightforward translation, following the example https://github.com/OpenRCT2/OpenRCT2/pull/12386.
There's some places to improve brevity, like returning EntertainerCostumes rather than converting back and forth to uint_8.

### DIFF
--- a/src/openrct2-ui/windows/Staff.cpp
+++ b/src/openrct2-ui/windows/Staff.cpp
@@ -282,7 +282,7 @@ static constexpr const uint32_t window_staff_page_enabled_widgets[] = {
 };
 // clang-format on
 
-static uint8_t _availableCostumes[ENTERTAINER_COSTUME_COUNT];
+static uint8_t _availableCostumes[static_cast<uint8_t>(EntertainerCostume::Count)];
 
 enum class PatrolAreaValue
 {
--- a/src/openrct2-ui/windows/Staff.cpp
+++ b/src/openrct2-ui/windows/Staff.cpp
@@ -282,7 +282,7 @@ static constexpr const uint32_t window_staff_page_enabled_widgets[] = {
 };
 // clang-format on
 
-static uint8_t _availableCostumes[static_cast<uint8_t>(EntertainerCostume::Count)];
+static EntertainerCostume _availableCostumes[static_cast<uint8_t>(EntertainerCostume::Count)];
 
 enum class PatrolAreaValue
 {
@@ -1472,12 +1472,13 @@ void window_staff_options_mousedown(rct_window* w, rct_widgetindex widgetIndex, 
     int32_t numCostumes = staff_get_available_entertainer_costume_list(_availableCostumes);
     for (int32_t i = 0; i < numCostumes; i++)
     {
-        uint8_t costume = _availableCostumes[i];
-        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costume)
+        EntertainerCostume costume = _availableCostumes[i];
+        uint8_t costume_value = static_cast<uint8_t>(costume);
+        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costume_value)
         {
             checkedIndex = i;
         }
-        gDropdownItemsArgs[i] = StaffCostumeNames[costume];
+        gDropdownItemsArgs[i] = StaffCostumeNames[costume_value];
         gDropdownItemsFormat[i] = STR_DROPDOWN_MENU_LABEL;
     }
 
@@ -1510,7 +1511,7 @@ void window_staff_options_dropdown(rct_window* w, rct_widgetindex widgetIndex, i
     if (dropdownIndex == -1)
         return;
 
-    uint8_t costume = _availableCostumes[dropdownIndex];
+    EntertainerCostume costume = _availableCostumes[dropdownIndex];
     auto staffSetCostumeAction = StaffSetCostumeAction(w->number, costume);
     GameActions::Execute(&staffSetCostumeAction);
 }
--- a/src/openrct2-ui/windows/Staff.cpp
+++ b/src/openrct2-ui/windows/Staff.cpp
@@ -1473,12 +1473,11 @@ void window_staff_options_mousedown(rct_window* w, rct_widgetindex widgetIndex, 
     for (int32_t i = 0; i < numCostumes; i++)
     {
         EntertainerCostume costume = _availableCostumes[i];
-        uint8_t costumeValue = static_cast<uint8_t>(costume);
-        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costumeValue)
+        if (peep->SpriteType == EntertainerCostumeToSprite(costume))
         {
             checkedIndex = i;
         }
-        gDropdownItemsArgs[i] = StaffCostumeNames[costumeValue];
+        gDropdownItemsArgs[i] = StaffCostumeNames[static_cast<uint8_t>(costume)];
         gDropdownItemsFormat[i] = STR_DROPDOWN_MENU_LABEL;
     }
 
--- a/src/openrct2-ui/windows/Staff.cpp
+++ b/src/openrct2-ui/windows/Staff.cpp
@@ -1473,12 +1473,12 @@ void window_staff_options_mousedown(rct_window* w, rct_widgetindex widgetIndex, 
     for (int32_t i = 0; i < numCostumes; i++)
     {
         EntertainerCostume costume = _availableCostumes[i];
-        uint8_t costume_value = static_cast<uint8_t>(costume);
-        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costume_value)
+        uint8_t costumeValue = static_cast<uint8_t>(costume);
+        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costumeValue)
         {
             checkedIndex = i;
         }
-        gDropdownItemsArgs[i] = StaffCostumeNames[costume_value];
+        gDropdownItemsArgs[i] = StaffCostumeNames[costumeValue];
         gDropdownItemsFormat[i] = STR_DROPDOWN_MENU_LABEL;
     }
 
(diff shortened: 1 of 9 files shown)


cfe316:
This is my first pull request to OpenRCT2, so I'd appreciate any feedback. As I wrote above I've made a straightforward translation following the model pull request for this enum-refactoring Project; there's probably some places where back and forth conversions could be improved or where code could be rewritten for brevity.

I did try running the built code, spawned some entertainers, and changed their types. Everything seemed normal.

src/openrct2/actions/StaffHireNewAction.hpp
```diff
@@ -55,15 +55,15 @@ DEFINE_GAME_ACTION(StaffHireNewAction, GAME_COMMAND_HIRE_NEW_STAFF_MEMBER, Staff
 private:
     bool _autoPosition = false;
     uint8_t _staffType = static_cast<uint8_t>(StaffType::Count);
-    uint8_t _entertainerType = ENTERTAINER_COSTUME::ENTERTAINER_COSTUME_COUNT;
+    uint8_t _entertainerType = static_cast<uint8_t>(EntertainerCostume::Count);
```
duncanspumpkin, line 58: I've just merged a change that means you can change the type of this class to the new enum

cfe316: Thanks. I was looking at that, but then saw the code << DS_TAG(_entertainerType) << down on line 99. Will the enum class be able to Serialise as before?

I'll try making that change.

tupaschoal: It should, that's exactly what Duncan's https://github.com/OpenRCT2/OpenRCT2/pull/12810 targeted

cfe316: Cool. I've changed this. There's still a place where I compare the type of a new/loaded entertainer to static_cast<uint8_t>(_entertainerCostume::Count), in order to check that it's valid. Is there a better way than `if (static_cast<uint8_t>(_entertainerType) >= static_cast<uint8_t>(EntertainerCostume::Count))` ?

src/openrct2/actions/StaffHireNewAction.hpp
```diff
@@ -204,7 +204,8 @@ DEFINE_GAME_ACTION(StaffHireNewAction, GAME_COMMAND_HIRE_NEW_STAFF_MEMBER, Staff
             PeepSpriteType spriteType = spriteTypes[_staffType];
             if (_staffType == static_cast<uint8_t>(StaffType::Entertainer))
             {
-                spriteType = static_cast<PeepSpriteType>(PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + _entertainerType);
+                spriteType = static_cast<PeepSpriteType>(
+                    PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + static_cast<uint8_t>(_entertainerType));
```
Broxzier, line 208: This code seems messy. It'd be better to introduce an array or function that maps a type to a sprite.

cfe316: I agree, it's a bit odd to do this sort of arithmetic, and I don't like how there's multiple casts in the one line.

I'll try writing a function that takes in one EntertainerCostume and returns a PeepSpriteType.

cfe316: In peep/Staff.cpp there's already a function 
`void Staff::SetCostume(uint8_t value)
{
    SpriteType = static_cast<PeepSpriteType>(value + PEEP_SPRITE_TYPE_ENTERTAINER_PANDA);
}`

Should I make another function there,
`PeepSpriteType Staff:GetEntertainerCostumeSprite(EntertainerCostume entertainerCostume)
{
   ...
   return entertainerCostumeSprite
}` ?
Or maybe there's a better file it to put it in? I'm new to the project and actually not that experienced with modifying large C++ projects.

src/openrct2/actions/StaffHireNewAction.hpp
duncanspumpkin: No you will still need to do that. It's not great but it's the only way to validate it fully

cfe316: Gotcha, thanks.

src/openrct2/object/SceneryGroupObject.cpp
```diff
@@ -150,28 +150,28 @@ uint32_t SceneryGroupObject::ReadJsonEntertainerCostumes(const json_t* jCostumes
 uint32_t SceneryGroupObject::ParseEntertainerCostume(const std::string& s)
```
tupaschoal, line 150: Make this function return `EntertainerCostume` and just cast to `uint32_t` on the single place that calls it. Will become neater.

src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -758,8 +758,8 @@ void window_staff_list_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32_
 
 static uint8_t window_staff_list_get_random_entertainer_costume()
 {
-    uint8_t result = ENTERTAINER_COSTUME_PANDA;
-    uint8_t costumeList[ENTERTAINER_COSTUME_COUNT];
+    uint8_t result = static_cast<uint8_t>(EntertainerCostume::Panda);
```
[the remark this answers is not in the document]
Broxzier:
```suggestion
    EntertainerCostume result = EntertainerCostume::Panda;
```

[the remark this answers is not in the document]
IntelOrca:
```suggestion
    auto result = EntertainerCostume::Panda;
```

src/openrct2/object/SceneryGroupObject.cpp
cfe316: Thanks, I've done so in the most recent commit.

src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -758,8 +758,8 @@ void window_staff_list_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int32_
...
+    uint8_t costumeList[static_cast<uint8_t>(EntertainerCostume::Count)];
```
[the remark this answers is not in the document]
cfe316: Thanks, this is fixed in the most recent commit.

src/openrct2-ui/windows/StaffList.cpp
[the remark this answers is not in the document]
cfe316: Thanks, this is fixed in the most recent commit. Should `auto` be used only where it should be obvious to humans as well as to the compiler?

src/openrct2-ui/windows/Staff.cpp
```diff
@@ -1472,12 +1472,13 @@ void window_staff_options_mousedown(rct_window* w, rct_widgetindex widgetIndex,
     int32_t numCostumes = staff_get_available_entertainer_costume_list(_availableCostumes);
     for (int32_t i = 0; i < numCostumes; i++)
     {
-        uint8_t costume = _availableCostumes[i];
-        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costume)
+        EntertainerCostume costume = _availableCostumes[i];
+        uint8_t costume_value = static_cast<uint8_t>(costume);
```
duncanspumpkin, line 1476: camelCase for function variables.

src/openrct2-ui/windows/StaffList.cpp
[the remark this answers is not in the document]
tupaschoal: It's debatable, I tend to use it when it's easy enough to guess or it doesn't really matter.

src/openrct2/object/SceneryGroupObject.h
```diff
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include "../peep/Staff.h"
```
tupaschoal, line 12: Move this include to the source file and add this after line 18:

```c++
enum class EntertainerCostume : uint8_t;
```

This is a forward declaration, so it compiles properly without us having to pull in all dependencies of `Staff.h`

src/openrct2/object/SceneryGroupObject.cpp
```diff
@@ -142,36 +142,36 @@ uint32_t SceneryGroupObject::ReadJsonEntertainerCostumes(const json_t* jCostumes
 
         // For some reason the flags are +4 from the actual costume IDs
         // See staff_get_available_entertainer_costumes
-        costumes |= 1 << (entertainer + 4);
+        costumes |= 1 << (static_cast<uint8_t>(entertainer) + 4);
```
tupaschoal, line 145: This is weirdle/oddly familiar to the calculation being done on `EntertainerCostumeToSprite` 🤔

tupaschoal (commented):
Seems mostly good to go. Note that when you do solve issues that we comment, you can (and should) mark them as resolved.

src/openrct2-ui/windows/StaffList.cpp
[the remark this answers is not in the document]
IntelOrca: Personal preference, I always use auto unless I can't.

src/openrct2/object/SceneryGroupObject.h
cfe316: Done!

src/openrct2/object/SceneryGroupObject.cpp
cfe316: Done!

src/openrct2-ui/windows/Staff.cpp
```diff
@@ -1472,12 +1472,13 @@ void window_staff_options_mousedown(rct_window* w, rct_widgetindex widgetIndex,
     int32_t numCostumes = staff_get_available_entertainer_costume_list(_availableCostumes);
     for (int32_t i = 0; i < numCostumes; i++)
     {
-        uint8_t costume = _availableCostumes[i];
-        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costume)
+        EntertainerCostume costume = _availableCostumes[i];
+        uint8_t costumeValue = static_cast<uint8_t>(costume);
+        if (peep->SpriteType == PEEP_SPRITE_TYPE_ENTERTAINER_PANDA + costumeValue)
```
tupaschoal, line 1477:
```suggestion
        if (peep->SpriteType == EntertainerCostumeToSprite(costume)
```
This can simplified as above, can't it? 
Then costumeValue can be moved to the inside the if (or just be dropped in favour of a direct cast in its single usage

cfe316: That makes sense; we can remove the costumeValue line. Done. I don't see any other extraneous uses of PEEP_SPRITE_TYPE_ENTERTAINER_PANDA.

tupaschoal (approved):

Issue #12875, opened 2020-09-07
Openrct2 Refuses to open my save file


Shysnx:
**OS: Windows 7
**Version: 0.3.0
**Commit/Build: 96d1db9

While I want to continue my project, the game refuses to load the save file. No errors or warning whatsoever. But other save files work. Filename: Drachenfest
- [ ] Reproducible in RCT2 (vanilla)? No
- [ ] Specific to multiplayer? No

**Steps to reproduce:**
1. Click open save
2. Click Drachenfest

**Screenshots / Video:**
https://youtu.be/ex2j2l1vpcI

**Save game:**
[Drachenfest.zip](https://github.com/OpenRCT2/OpenRCT2/files/5181313/Drachenfest.zip)

keithstellyes:
@janisozaur are you still looking at this? I was wanting to look into it

Can confirm it is happening on latest version on Linux

tupaschoal:
Please do!

keithstellyes:
So, while investigating I found a few things:

Running RollerCoaster Tycoon 2 from GOG, via WINE, this _does_ reproduce

![Screenshot from 2020-10-09 19-36-46](https://user-images.githubusercontent.com/13317074/95643898-30b4da00-0a67-11eb-8a9b-c4837b334fcd.png)

An exception is thrown that a zero-sized chunk was found, still looking further. I wonder if OpenRCT2 corrupted the file, then?

PR #12877, opened 2020-09-07
Use correct count of research items when importing park

janisozaur:


janisozaur:
https://github.com/OpenRCT2/OpenRCT2/blob/96d1db97e0bbc689ef9d5d48ee2514bec7c5c7f8/src/openrct2/scenario/Scenario.h#L241

https://github.com/OpenRCT2/OpenRCT2/blob/96d1db97e0bbc689ef9d5d48ee2514bec7c5c7f8/src/openrct2/rct12/RCT12.h#L799-L819

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -919,7 +919,7 @@ class S6Importer final : public IParkImporter
     void ImportResearchList()
     {
         bool invented = true;
-        for (size_t i = 0; i < sizeof(_s6.research_items); i++)
+        for (size_t i = 0; i < std::size(_s6.research_items); i++)
```
Broxzier, line 922: Since the size is known at compile-time and the index is not used for anything else, this could be refactored to a for-range loop.

```suggestion
        for (RCT12ResearchItem& ri : research_items)
```

janisozaur:
eh, just noticed this will need more fixes to update to new style. Will update the PR later

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -919,7 +919,7 @@ class S6Importer final : public IParkImporter
     void ImportResearchList()
     {
         bool invented = true;
-        for (size_t i = 0; i < sizeof(_s6.research_items); i++)
+        for (RCT12ResearchItem& ri : research_items)
```
tupaschoal, line 922:
```suggestion
        for (const auto& ri : research_items)
```

tupaschoal (commented):
Nice find!

tupaschoal:
By the way, is this related to any of the open issues or just something you found "by chance"?

src/openrct2/rct2/S6Importer.cpp
Broxzier: Nice use of auto. Not sure if `const` can be used here, since a pointer gets passed in the body. Can't try it right now though.

janisozaur:
By chance, when investigating #12875 (but that doesn't fix it)

src/openrct2/rct2/S6Importer.cpp
Gymnasiast: This doesn't compile. Should have been `_s6.research_items`.

tupaschoal (approved):
Looks fine to me. Not sure if @janisozaur wanted to do anything else here though

PR #12886, opened 2020-09-08
Add visitors for all scenery placement and removes

duncanspumpkin:
This should allow greater flexibility for plugin developers


IntelOrca:
Probably should increment plugin API version.

duncanspumpkin:
API now incremented, added a changelog entry and fixed CoordsXYZ visitor.

IntelOrca (approved):

Issue #400, opened 2014-09-03
Bug in original Rollercoaster Designer


ghost:
Test case. In Rollercoaster Designer, build a looping coaster to this specification:

Filename:   Spin Drier.TD6
Ride Type:  Looping Roller Coaster
Vehicle Type:   Looping rollercoaster trains (1 7-car train)
Operating Mode: Powered Launch Mode, speed 49 mph

1-10    10  Helix down left [R2](11 to 1)
11      Helix down left [R3](1 to 0)
12      Roll left to level (0)
13      On-ride photo section (0)
14      Level to up (0 to 1)
15      Vertical loop left (1 to 0)
16      Down to level (0)
17      Begin station (0)
18-21   4   Middle station (0)
22      End station (0)
23-25   3   Level (0)
26      Level to up (0 to 1)
27      Vertical loop right (1 to 0)
28      Down to level (0)
29      Level (0)
30      Level to up (0 to 1)
31      Up to up steep (1 to 3)
32-36   5   Up steep (3 to 23)

It looks like this

![spin drier](https://cloud.githubusercontent.com/assets/8623605/4140570/7dcdc3a4-33a6-11e4-9c9a-8ce0cdf94d80.png)

Set launch speed to 49mph. Test so you can then save.

Now in coaster designer try and plant another copy of the ride into the designer using your saved file.

Bug: It is 5 feet higher than the original and refuses to be lowered to zero height.

Further, if you place it in a park in a scenario, add a queue line, and 'save with scenery' a 2nd copy with the queue lines added 2 things happen with the copy.

1) You cannot see a preview picture when you select it to be added into a park from the coaster builder menu.
2) When you place it in the park, the queue lines that you saved with the ride are placed at the incorrect height and no longer join onto the entrance and exit.

frutiemax:
Just for fun, I've recreated this roller coaster in the track designer, tried to save it, but I get the Can't save track design... Ride is too large, contains too many elements, or scenery is too spread out.

![bugRollerCoaster](https://user-images.githubusercontent.com/12629817/85203561-a4767f80-b2dc-11ea-9bce-d26ea50df9d1.PNG)

This is on v0.2.6 Master.

frutiemax:
So I've analyzed this a bit, and I arrived to the conclusion that there is a problem when the first element on the roller coaster in the direction of the train is a small helix. When I add a straight segment, the roller coaster can save itself.
[...]

PR #12888, opened 2020-09-08
Fix #5753: Entertainers make themselves happy instead of the guests

### LINKED ISSUE #5753: Guests will not wait in queues
**OS:** [Windows 10]
**Version:** 0.0.8
**Commit/Build:** cb28381

I actually joined the forum for this reason. In RCT2 vanilla, guests would wait for hours (literally) as long as entertainers and TVs were in the line. In Open RCT2 that's not the case. They start complaining after a few minutes even with entertainers and TVs. It's a real bummer because I spend a lot of time on long, themed queues. I really like the realism of long queues, makes your big attractions seem that much bigger. Any word if there is a way to fix this or if a change is coming?
Everything about Open RCT2 is amazing, but I would love to see TVs and entertainers working like they did in vanilla. 

 

- [ ] Multiplayer?

Files were uploaded on the official OpenRCT2 forum.

https://openrct2.org/forums/topic/2531-guests-will-not-wait-in-queues/

richard-fine:
After identifying guests that are near an entertainer, modify the guest's HappinessTarget and TimeInQueue values instead of the entertainer's own values.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -4,6 +4,7 @@
 - Feature: [#12712] Add TCP / socket plugin APIs.
 - Feature: [#12840] Add Park.entranceFee to the plugin API.
 - Fix: [#400] Unable to place some saved tracks flush to the ground (original bug).
+- Fix: [#5753]: Entertainers make themselves happy instead of the guests.
 - Fix: [#7037] Unable to save tracks starting with a sloped turn or helix.
 - Fix: [#12691] Ride graph tooltip incorrectly used count instead of number string.
 - Fix: [#12694] Crash when switching ride types with construction window open.
--- a/src/openrct2/peep/Staff.cpp
+++ b/src/openrct2/peep/Staff.cpp
@@ -1009,19 +1009,19 @@ bool Staff::DoMiscPathFinding()
  *
  *  rct2: 0x006C086D
  */
-static void staff_entertainer_update_nearby_peeps(Peep* peep)
+void Staff::EntertainerUpdateNearbyPeeps() const
 {
     for (auto guest : EntityList<Guest>(EntityListId::Peep))
     {
         if (guest->x == LOCATION_NULL)
             continue;
 
-        int16_t z_dist = abs(peep->z - guest->z);
+        int16_t z_dist = abs(z - guest->z);
         if (z_dist > 48)
             continue;
 
-        int16_t x_dist = abs(peep->x - guest->x);
-        int16_t y_dist = abs(peep->y - guest->y);
+        int16_t x_dist = abs(x - guest->x);
+        int16_t y_dist = abs(y - guest->y);
 
         if (x_dist > 96)
             continue;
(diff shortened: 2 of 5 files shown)


duncanspumpkin (commented):
Could you take this opportunity to make this a staff member function. This will likely need a replay rerecording and will definitely need a network version bump. Also you need to add a changelog entry

src/openrct2/peep/Staff.cpp
```diff
@@ -1009,41 +1009,41 @@ bool Staff::DoMiscPathFinding()
...
+        if (guest->State == PEEP_STATE_WALKING)
         {
-            peep->HappinessTarget = std::min(peep->HappinessTarget + 4, PEEP_MAX_HAPPINESS);
+            guest->HappinessTarget = std::min(guest->HappinessTarget + 4, PEEP_MAX_HAPPINESS);
         }
-        else if (peep->State == PEEP_STATE_QUEUING)
+        else if (guest->State == PEEP_STATE_QUEUING)
         {
-            if (peep->TimeInQueue > 200)
+            if (guest->TimeInQueue > 200)
```
duncanspumpkin, line 1038: This can be refactored to `guest->TimeInQueue = std::max(0, guest->TimeInQueue - 200);`

Gymnasiast:
If we fix something, we describe the _problem_ that is being fixed. I have modified the title of your PR accordingly. That description should also go into the changelog.

Could you also fix the commit title?

src/openrct2/peep/Staff.cpp
richard-fine: Done. I got a little worried because `TimeInQueue` is unsigned, but I think the 'usual integer conversion' rules will apply here and it will be converted to signed int before subtraction/compare.

richard-fine:
I think my latest push has everything you guys are asking for, except for the replay re-recording. What do you need me to do for that?

distribution/changelog.txt
```diff
@@ -15,6 +15,7 @@
 - Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
 - Fix: [#12857] Incorrect Peep thoughts in imported RCT1 parks.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
+- Fix: [#5753]: Entertainers make themselves happy instead of the guests.
```
Gymnasiast, line 18: This needs to be in order, so between #400 and #7037.

Gymnasiast:
@richard-fine It's not certain a replay update is needed - we'll know once the CIs are done (if they are all green, it's good to go).

Gymnasiast (approved):
LGTM (assuming CIs will be green)

janisozaur:
Wouldn't a replay be useful to track regressions?

duncanspumpkin:
I'm surprised bpb didn't trigger the issue.

duncanspumpkin (approved):

PR #12890, opened 2020-09-09
Add stroke to lowercase 'L' to differentiate from capital 'I'

tellovishous:
The problem with many sans-serif fonts, such as those in RCT2, is that it can be difficult to tell between a lowercase L and an uppercase I, and can be visually ambiguous in cases like this: 
unknown (1)

What I've done is add a small stroke to the bottom of the lowercase L's so that it will stand out better when it's used next to capital I's. 
unknown


duncanspumpkin:
Could you show a before and after image of the same part of the interface

Gymnasiast:
Do you think you can pull off adding the drawing code as well? You can use https://github.com/OpenRCT2/OpenRCT2/pull/12493 as a guide, as it did exactly the same thing for the `j` character.

tellovishous:
Hmm perhaps, I'll see if I can add it like the j letter.

tellovishous:
Well I added it but it doesn't seem to build right. Idk how to fix that. ><

resources/g2/sprites.json
```diff
@@ -962,6 +962,10 @@
         "forceBmp": true
     },
     {
+        "path": "font/latin/l-small.png",
+        "palette": "keep",
+        "forceBmp": true
+    },
         "path": "font/latin/c-circumflex-uc-small.png",
```
AaronVanGeffen, line 969: There is a syntax error here now due to a missing opening brace:
```suggestion
    {
        "path": "font/latin/c-circumflex-uc-small.png",
```

AaronVanGeffen: Note that the order in which these characters are defined matters for the sprite ids! In order for them to match the definitions in `sprites.h`, you should *append* them the end of the file, not put them in between.

tellovishous: Ahh you're right I missed that, I'll add those fixes.

tellovishous:
Alright it's working now. Here's a before and after:
image

AaronVanGeffen:
Thanks! The changes look good in the screenshot.

I have pushed a commit to your branch that adds an entry to the changelog, including #12806 as well.

distribution/changelog.txt
```diff
@@ -16,6 +16,8 @@
 - Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
 - Fix: [#12857] Incorrect Peep thoughts in imported RCT1 parks.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
+- Change: [#12806] Add Esperanto diacritics to the sprite font.
```
Gymnasiast, line 19: @AaronVanGeffen I have always put new letters in as a "feature". I suppose that "improved" would also fit, but we tend to use "Change" for stuff that is more likely to be neutral or controversial.

tellovishous:
Nice, thanks very much!

Gymnasiast:
I wonder if the kerning would be improved by taking away a pixel on the right. What do you all think?

AaronVanGeffen:
@Gymnasiast Hmm, that would meld combinations of two l's together, right? That's not typically done in typesetting. I think it would look better to leave it as-is.

distribution/changelog.txt
AaronVanGeffen: @Gymnasiast Fair enough; I think "Improved" would work just as well. Feel free to address this as you see fit. I just didn't want it to be forgotten.

tellovishous:
While I can see where you're coming from, I second that it looks fine as is. It's not like we usually remove side pixels from letters like lowercase 't' for example.

Gymnasiast (commented):
The new shape of the 'l' will take some getting used to. Nevertheless, this PR works as intended, and if the general consensus is that this is a good idea, I am not going to object.

AaronVanGeffen (approved):

PR #12041, opened 2020-06-23
Part of #11389: Use formatter class for Ride::FormatNameTo

frutiemax:

PR #12891, opened 2020-09-09
Fix #12881: Guests' favourite rides not listed in the guest window

### LINKED ISSUE #12881: All guests in the park show "N/A" as their favorite ride
**OS:** Windows 10
**Version:** 0.3.0
**Commit/Build:** 6ac23fb

An issue that I have only started noticing today is that every guest in the park, no matter how many rides they have ridden, will always show "N/A" as their favorite ride, even though the numbers of guest favorites for the rides always change. I have noticed this issue in every scenario I play, from RCT1 and RCT2 scenarios to custom ones.

**Steps to reproduce:**
1. Open up any scenario or saved game with many guests in it.
2. Click on a guest and check their rides tab.
3. You should see a list of rides and "N/A" as their favorite ride, even if they have gone on a lot of rides/attractions.

**Screenshots / Video:**
Onyx Heights 2020-09-07 18-27-44

tupaschoal:
I scoured #12041 and this seems to be the only actual bug that slipped (reusing a formatter but without resetting it back to "0")

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -16,7 +16,7 @@
 - Fix: [#12820] Title menu buttons not invalidating properly
 - Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
 - Fix: [#12857] Incorrect Peep thoughts in imported RCT1 parks.
-- Fix: [#12881] Guests have no favorite ride.
+- Fix: [#12881] Guests' favourite rides are not listed in the guest window.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
 - Technical: The required version of macOS has been lowered to 10.13 (High Sierra).
 
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -16,6 +16,7 @@
 - Fix: [#12820] Title menu buttons not invalidating properly
 - Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
 - Fix: [#12857] Incorrect Peep thoughts in imported RCT1 parks.
+- Fix: [#12881] Guests have no favorite ride.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
 - Technical: The required version of macOS has been lowered to 10.13 (High Sierra).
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin (commented):
Can you make this into an if else. I'd prefer we limit use of rewind except when we absolutely need it

tupaschoal:
> Can you make this into an if else. I'd prefer we limit use of rewind except when we absolutely need it

Makes sense, I had discarded it at first because of the pair of equal elses

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1765,14 +1765,21 @@ void window_guest_rides_paint(rct_window* w, rct_drawpixelinfo* dpi)
     screenCoords.y = w->windowPos.y + window_guest_rides_widgets[WIDX_PAGE_BACKGROUND].bottom - 12;
 
     auto ft = Formatter::Common();
-    ft.Add<rct_string_id>(STR_PEEP_FAVOURITE_RIDE_NOT_AVAILABLE);
     if (peep->FavouriteRide != RIDE_ID_NULL)
```
duncanspumpkin, line 1768: I wonder if we even need this check. Could simplify the if else if get_ride returns nullptr on ride id null.

duncanspumpkin (approved):

distribution/changelog.txt
```diff
@@ -15,6 +15,7 @@
 - Fix: [#12820] Title menu buttons not invalidating properly
 - Fix: [#12845] Deleting ride with active ad campaign creates incorrect notification.
 - Fix: [#12857] Incorrect Peep thoughts in imported RCT1 parks.
+- Fix: [#12881] Guests have no favorite ride.
```
AaronVanGeffen, line 18: This suggests the favourite ride is missing entirely, implying that save games are affected. Since this is a UI issue, I suggest we phrase it as such, e.g.:
```suggestion
- Fix: [#12881] Guests' favourite rides are not listed in the guest window.
```

src/openrct2-ui/windows/Guest.cpp
Broxzier: Sounds like a good idea, then the two identical else blocks can be merged too.

duncanspumpkin: Just checked and it should work fine to make that change.

Gymnasiast (approved):

PR #12907, opened 2020-09-12
Staff code cleanup

richard-fine:
Various cleanups in the Staff code:

~- Convert the STAFF_ORDERS enum to an enum class. This is a 'flags' style enum, so to allow code to continue to easily set bits, I've defined bitwise-or and or-equal operators. I've also added a global `HasFlag` utility method which can be used to test if a particular bit is set, because we cannot define any implicit conversion to bool for an enum class and making bitwise-and return a bool felt like a bad idea.~ Removed for now, as it clashes with #12757.

- Added a bunch of `const` keywords, on methods and on parameters. It's now clearer to see which of the UpdateFixing*() steps will/won't actually modify the Ride, for example.

- Converted a bunch of usage of `uint8_t` to the `Direction` type, where appropriate.

- A couple of other minor refactorings / simplifications.


richard-fine:
Ah, looks like I might be slightly stepping on https://github.com/OpenRCT2/OpenRCT2/pull/12757 with this...

duncanspumpkin:
Remove the flags change for now. Until we decide what we are doing with 12757

richard-fine:
Done.

src/openrct2/peep/Staff.cpp
```diff
@@ -79,6 +79,9 @@ colour_t gStaffHandymanColour;
 colour_t gStaffMechanicColour;
 colour_t gStaffSecurityColour;
 
+// Maximum manhattan distance that litter can be for a handyman to seek to it
+const uint16_t MAX_LITTER_DISTANCE = 0x60;
```
Gymnasiast, line 83:
```suggestion
const uint16_t MAX_LITTER_DISTANCE = 3 * COORDS_XY_STEP;
```

This makes it clearer that it's 3 tiles (a tile is divided into 32x32 "big" x/y coords).

src/openrct2/peep/Staff.cpp
```diff
@@ -669,8 +671,8 @@ bool Staff::DoHandymanPathFinding()
         }
     }
 
-    // countof(CoordsDirectionDelta)
-    assert(newDirection < 8);
+    // newDirection can only contain a cardinal direction at this point, no diagonals
+    assert(direction_valid(newDirection));
```
Gymnasiast, line 675: This is not correct. `direction_valid()` will return false for directions 4-7, which the old assert considered valid.

src/openrct2/peep/Staff.cpp
```diff
@@ -904,7 +905,7 @@ bool Staff::DoMechanicPathFinding()
     }
 
     // countof(CoordsDirectionDelta)
-    assert(newDirection < 8);
+    assert(direction_valid(newDirection));
```
Gymnasiast, line 908: ditto

src/openrct2/peep/Staff.cpp
richard-fine: The old assert was wrong. Firstly, it's not possible for `newDirection` to contain values 4-7 at this point - every assignment to it comes via a codepath which is doing either `& 0x03` or `% NumOrdinalDirections`. Secondly, we are assigning it to PeepDirection below, and if you look through usages of that variable you will see many places where a value greater than 3 would break things.

src/openrct2/peep/Staff.cpp
richard-fine: As above.

src/openrct2/peep/Staff.cpp
duncanspumpkin: The old assert was to protect invalid array access. This change is correct

src/openrct2/peep/Peep.cpp
```diff
@@ -339,6 +339,11 @@ Staff* Peep::AsStaff()
     return AssignedPeepType == PeepType::Staff ? static_cast<Staff*>(this) : nullptr;
 }
 
+const Staff* Peep::AsStaff() const
```
duncanspumpkin, line 342: I'm trying to get rid of these in favour of `Is<Staff>` or whatever.

src/openrct2/peep/Staff.cpp
```diff
@@ -187,22 +190,21 @@ bool Staff::IsLocationInPatrol(const CoordsXY& loc) const
     return IsPatrolAreaSet(loc);
 }
 
-bool staff_is_location_on_patrol_edge(Peep* mechanic, const CoordsXY& loc)
+// Check whether the location x,y is inside and on the edge of the
+// patrol zone for mechanic.
+bool staff_is_location_on_patrol_edge(const Peep* mechanic, const CoordsXY& loc)
```
duncanspumpkin, line 195: Since you are refactoring this make it a member function of the `Staff`. Then there is no need to use the `AsStaff`

duncanspumpkin: Also change the function to TitleCase

src/openrct2/peep/Staff.cpp
```diff
@@ -187,22 +190,21 @@ bool Staff::IsLocationInPatrol(const CoordsXY& loc) const
...
         auto neighbourPos = loc + CoordsDirectionDelta[neighbourDir];
-        onZoneEdge = !mechanic->AsStaff()->IsLocationInPatrol(neighbourPos);
-        neighbourDir++;
+        onZoneEdge = !staff->IsLocationInPatrol(neighbourPos);
     }
     return onZoneEdge;
 }
 
-bool staff_can_ignore_wide_flag(Peep* staff, const CoordsXYZ& staffPos, TileElement* path)
+bool staff_can_ignore_wide_flag(const Peep* staff, const CoordsXYZ& staffPos, TileElement* path)
```
duncanspumpkin, line 207: make a member function

src/openrct2/peep/Staff.cpp
```diff
@@ -323,7 +325,7 @@ bool staff_can_ignore_wide_flag(Peep* staff, const CoordsXYZ& staffPos, TileElem
  *  rct2: 0x006C095B
  *  returns 0xF if not in a valid patrol area
  */
-static uint8_t staff_get_valid_patrol_directions(Staff* staff, const CoordsXY& loc)
+static uint8_t staff_get_valid_patrol_directions(const Staff* staff, const CoordsXY& loc)
```
duncanspumpkin, line 328: make a member function

src/openrct2/peep/Staff.cpp
```diff
@@ -431,7 +433,7 @@ void staff_toggle_patrol_area(int32_t staffIndex, const CoordsXY& coords)
  *
  * Returns INVALID_DIRECTION when no nearby litter or unpathable litter
  */
-static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
+static Direction staff_handyman_direction_to_nearest_litter(const Peep* peep)
```
duncanspumpkin, line 436: make a member function

src/openrct2/peep/Staff.cpp
```diff
@@ -463,7 +465,7 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
     int16_t x_diff = litterTile.x - peep->x;
     int16_t y_diff = litterTile.y - peep->y;
 
-    uint8_t nextDirection = 0;
+    Direction nextDirection = 0;
 
     if (abs(x_diff) <= abs(y_diff))
```
duncanspumpkin, line 470: I feel as though there is a function that does this already. Its used in a number of locations.

src/openrct2/peep/Staff.cpp
```diff
@@ -516,7 +518,7 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
  *
  *  rct2: 0x006BF931
  */
-static uint8_t staff_handyman_direction_to_uncut_grass(Peep* peep, uint8_t valid_directions)
+static uint8_t staff_handyman_direction_to_uncut_grass(const Peep* peep, uint8_t valid_directions)
```
duncanspumpkin, line 521: Make a member function

src/openrct2/peep/Staff.cpp
```diff
@@ -691,7 +693,7 @@ bool Staff::DoHandymanPathFinding()
     return false;
 }
 
-static uint8_t staff_direction_surface(Peep* peep, uint8_t initialDirection)
+static Direction staff_direction_surface(const Peep* peep, Direction initialDirection)
```
duncanspumpkin, line 696: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -734,9 +736,9 @@ static uint8_t staff_direction_surface(Peep* peep, uint8_t initialDirection)
  *
  *  rct2: 0x006BFF45
  */
-static uint8_t staff_mechanic_direction_surface(Peep* peep)
+static Direction staff_mechanic_direction_surface(const Peep* peep)
```
duncanspumpkin, line 739: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -770,7 +772,7 @@ static uint8_t staff_mechanic_direction_surface(Peep* peep)
  *
  *  rct2: 0x006C02D1
  */
-static uint8_t staff_mechanic_direction_path_rand(Peep* peep, uint8_t pathDirections)
+static Direction staff_mechanic_direction_path_rand(const Peep* peep, uint8_t pathDirections)
```
duncanspumpkin, line 775: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -794,9 +796,8 @@ static uint8_t staff_mechanic_direction_path_rand(Peep* peep, uint8_t pathDirect
  *
  *  rct2: 0x006C0121
  */
-static uint8_t staff_mechanic_direction_path(Peep* peep, uint8_t validDirections, PathElement* pathElement)
+static Direction staff_mechanic_direction_path(Peep* peep, uint8_t validDirections, PathElement* pathElement)
```
duncanspumpkin, line 799: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -926,9 +927,8 @@ bool Staff::DoMechanicPathFinding()
  *
  *  rct2: 0x006C050B
  */
-static uint8_t staff_direction_path(Peep* peep, uint8_t validDirections, PathElement* pathElement)
+static Direction staff_direction_path(const Peep* peep, uint8_t validDirections, PathElement* pathElement)
```
duncanspumpkin, line 930: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -1634,10 +1631,10 @@ static constexpr const CoordsXY _WateringUseOffsets[] = {
  *
  *  rct2: 0x006BF483
  */
-static int32_t peep_update_patrolling_find_watering(Peep* peep)
+static bool peep_update_patrolling_find_watering(Peep* peep)
```
duncanspumpkin, line 1634: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -1696,51 +1693,51 @@ static int32_t peep_update_patrolling_find_watering(Peep* peep)
...
-    return 0;
+    return false;
 }
 
 /**
  *
  *  rct2: 0x006BF3A1
  */
-static int32_t peep_update_patrolling_find_bin(Peep* peep)
+static bool peep_update_patrolling_find_bin(Peep* peep)
```
duncanspumpkin, line 1706: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -1764,23 +1761,23 @@ static int32_t peep_update_patrolling_find_bin(Peep* peep)
...
-    return 1;
+    return true;
 }
 
 /**
  *
  *  rct2: 0x006BF322
  */
-static int32_t peep_update_patrolling_find_grass(Peep* peep)
+static bool peep_update_patrolling_find_grass(Peep* peep)
```
duncanspumpkin, line 1771: member function

src/openrct2/peep/Staff.cpp
```diff
@@ -1793,20 +1790,20 @@ static int32_t peep_update_patrolling_find_grass(Peep* peep)
...
-    return 0;
+    return false;
 }
 
 /**
  *
  *  rct2: 0x006BF295
  */
-static int32_t peep_update_patrolling_find_sweeping(Peep* peep)
+static bool peep_update_patrolling_find_sweeping(Peep* peep)
```
duncanspumpkin, line 1803: member function

duncanspumpkin:
As you are improving these functions can you switch them to title case and make them member variables. Can you also get rid of the AsStaff const version.

src/openrct2/peep/Staff.cpp
richard-fine: I've had a look in Location.hpp and Util.h but couldn't see any obvious candidate.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -785,7 +786,7 @@ static void peep_pathfind_heuristic_search(
                 if (tileElement->AsPath()->IsWide())
                 {
                     /* Check if staff can ignore this wide flag. */
-                    if (!staff_can_ignore_wide_flag(peep, loc.ToCoordsXYZ(), tileElement))
+                    if (!peep->Is<Staff>() || !static_cast<Staff*>(peep)->CanIgnoreWideFlag(loc.ToCoordsXYZ(), tileElement))
```
AaronVanGeffen, line 789:
```suggestion
                    if (!peep->Is<Staff>() || !peep->AsStaff()->CanIgnoreWideFlag(loc.ToCoordsXYZ(), tileElement))
```

richard-fine: Deliberately not using `AsStaff` here, per @duncanspumpkin https://github.com/OpenRCT2/OpenRCT2/pull/12907#discussion_r488788491

AaronVanGeffen: Ah, I didn't realise. Thanks.

duncanspumpkin: Yeah but i meant use `As<Staff>`

richard-fine: Oh! My bad, you mentioned using `Is` but I didn't spot that there are `As` helpers as well. Will fix.

src/openrct2/peep/Staff.cpp
```diff
@@ -493,8 +491,8 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
         }
     } while (!(tileElement++)->IsLastForTile());
 
-    nextTile.x = (peep->x & 0xFFE0) + CoordsDirectionDelta[nextDirection].x;
-    nextTile.y = (peep->y & 0xFFE0) + CoordsDirectionDelta[nextDirection].y;
+    nextTile.x = (x & 0xFFE0) + CoordsDirectionDelta[nextDirection].x;
```
Gymnasiast, line 494:
```suggestion
    nextTile = CoordsXY(x, y).ToTileStart() + CoordsDirectionDelta[nextDirection];
```

If we're touching this line, might as well clean it up.

src/openrct2/peep/Staff.cpp
```diff
@@ -478,7 +476,7 @@ static uint8_t staff_handyman_direction_to_nearest_litter(Peep* peep)
     CoordsXY nextTile = { static_cast<int32_t>((nearestLitter->x & 0xFFE0) - CoordsDirectionDelta[nextDirection].x),
                           static_cast<int32_t>((nearestLitter->y & 0xFFE0) - CoordsDirectionDelta[nextDirection].y) };
 
-    int16_t nextZ = ((peep->z + 8) & 0xFFF0) / 8;
+    int16_t nextZ = ((z + 8) & 0xFFF0) / 8;
```
Gymnasiast, line 479:
```suggestion
    int16_t nextZ = ((z + COORDS_Z_STEP) & 0xFFF0) / COORDS_Z_STEP;
```

src/openrct2/peep/Staff.cpp
```diff
@@ -788,64 +785,63 @@ static uint8_t staff_mechanic_direction_path_rand(Peep* peep, uint8_t pathDirect
...
     if (pathDirections == 0)
     {
-        if (peep->State != PEEP_STATE_ANSWERING && peep->State != PEEP_STATE_HEADING_TO_INSPECTION)
+        if (State != PEEP_STATE_ANSWERING && State != PEEP_STATE_HEADING_TO_INSPECTION)
         {
             return direction;
         }
 
-        if (peep->SubState != 2)
+        if (SubState != 2)
```
Gymnasiast, line 821: I'm sure we have enums for substate. Could you use them here?

src/openrct2/peep/Staff.cpp
```diff
@@ -927,39 +923,35 @@ bool Staff::DoMechanicPathFinding()
...
+    if (pathDirections == (1 << direction))
     {
         return direction;
     }
 
-    pathDirections |= (1 << direction);
-
     direction = scenario_rand() & 3;
-    for (int32_t i = 0; i < 4; ++i, ++direction)
+    for (uint8_t i = 0; i < 4; ++i, direction = direction_next(direction))
```
Gymnasiast, line 953:
```suggestion
    for (uint8_t i = 0; i < NumOrthogonalDirections; ++i, direction = direction_next(direction))
```

src/openrct2/peep/Staff.cpp
duncanspumpkin: I'm not sure we do for staff sub states

richard-fine: Indeed, no enum exists for this yet.

src/openrct2/peep/Staff.cpp
richard-fine: Introduced `direction_from_to`.

src/openrct2/world/Location.hpp
```diff
@@ -503,6 +503,27 @@ constexpr Direction ALL_DIRECTIONS[] = { 0, 1, 2, 3 };
     return (dir - 1) & 0x03;
 }
 
+/**
+ * Given two positions, return the cardinal direction which is closest to the direction from 'from' to 'to'.
+ */
+[[maybe_unused]] static constexpr Direction direction_from_to(const CoordsXY& from, const CoordsXY& to)
```
duncanspumpkin, line 509: For all new functions make sure you use TitleCase

richard-fine: You don't think it's better to be consistent with the other `direction_*` functions?

duncanspumpkin: No if anything it will serve as a reminder to fix them

duncanspumpkin (approved):

PR #12908, opened 2020-09-12
Refactor PeepSittingSubState

rpstester:
Ref #12389 
This is my first PR; please let me know any feedback.

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1228,7 +1228,7 @@ void Guest::TryGetUpFromSitting()
  */
 void Guest::UpdateSitting()
 {
-    if (SubState == PEEP_SITTING_TRYING_TO_SIT)
+    if (SubState == PeepSittingSubState::TryingToSit)
     {
         if (!CheckForPath())
             return;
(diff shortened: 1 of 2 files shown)


richard-fine:
You can't assign an `enum class` value to a `uint8_t` (without casting) - but you also can't just turn `Peep::SubState` into a `PeepSittingSubState` because it's used for many other purposes too.

I think maybe turning it into a `union` is your best bet - a union of `uint8_t SubState` and `PeepSittingSubState SittingSubState` should let you share the storage while offering correctly-typed access.

rpstester:
Why did the macOS fail, and how do I fix it? It's not showing me any details.

rpstester:
Please let me know if this resolves #12389 and if you have any feedback. Thank you to @richard-fine for helping me when I was stuck.

duncanspumpkin:
Can you clean up your commits so that this is just one single commit. You shouldn't have any merge commits in this.

rpstester:
> Can you clean up your commits so that this is just one single commit. You shouldn't have any merge commits in this.

Sure! Only I'm not sure how to do so.  Would you please point me in the right direction?  I'm still learning Git(Hub).

duncanspumpkin:
As your just getting started we can take care of it this time. I think this pr looks about ready for merging so don't want to waste time.

For future refer to this https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2
You can use interactive rebase to squash commits together or reword them.

duncanspumpkin (approved):

PR #12916, opened 2020-09-12
Pathfinding cleanup

richard-fine:
Consolidate all the pathfinding code into GuestPathfinding.cpp, and introduce a GuestPathfinding.h for the externally visible members (instead of leaving their declarations in Peep.h). I've also commented what those externally-visible members are / do, to the best of my knowledge.

### DIFF
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -7,6 +7,8 @@
  * OpenRCT2 is licensed under the GNU General Public License version 3.
  *****************************************************************************/
 
+#include "GuestPathfinding.h"
+
 #include "../core/Guard.hpp"
 #include "../ride/RideData.h"
 #include "../ride/Station.h"
@@ -25,6 +27,17 @@ static int8_t _peepPathFindNumJunctions;
 static int8_t _peepPathFindMaxJunctions;
 static int32_t _peepPathFindTilesChecked;
 static uint8_t _peepPathFindFewestNumSteps;
+
+TileCoordsXYZ gPeepPathFindGoalPosition;
+bool gPeepPathFindIgnoreForeignQueues;
+ride_id_t gPeepPathFindQueueRideIndex;
+
+#if defined(DEBUG_LEVEL_1) && DEBUG_LEVEL_1
+// Use to guard calls to log messages
+bool gPathFindDebug = false;
+// Use to put the peep name in the log message
+utf8 gPathFindDebugPeepName[256];
+#endif // defined(DEBUG_LEVEL_1) && DEBUG_LEVEL_1
 
 static int32_t guest_surface_path_finding(Peep* peep);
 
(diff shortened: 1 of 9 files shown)


duncanspumpkin:
Don't change logic and refactor in one pr. Keep them separate.

richard-fine:
Split the logic changes off into https://github.com/OpenRCT2/OpenRCT2/pull/12917

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2217,3 +2230,81 @@ int32_t guest_path_finding(Guest* peep)
 #endif // defined(DEBUG_LEVEL_1) && DEBUG_LEVEL_1
     return peep_move_one_tile(direction, peep);
 }
+
+bool is_valid_path_z_and_direction(TileElement* tileElement, int32_t currentZ, int32_t currentDirection)
```
duncanspumpkin, line 2234: Since you are moving these functions can you go through and follow the coding style. 
Function names should be TitleCase. variable names should be lower camelCase (which I'm pretty sure they all are).

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2217,3 +2230,81 @@ int32_t guest_path_finding(Guest* peep)
...
+            return false;
+    }
+    return true;
+}
+
+/**
+ *
+ *  rct2: 0x0069A98C
+ */
+void peep_reset_pathfind_goal(Peep* peep)
```
duncanspumpkin, line 2265: This is not a `Guest` function but a `Peep` function I've never been too sure though what I was doing with splitting things up into a GuestPathfinding as its pretty much all Peep pathfinding. Eitherway this is a member function and should be changed to reflect that.

src/openrct2/peep/GuestPathfinding.h
```diff
@@ -0,0 +1,72 @@
...
+ * Copyright (c) 2014-2020 OpenRCT2 developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#ifndef _GUESTPATHFINDING_H_
+#define _GUESTPATHFINDING_H_
```
duncanspumpkin, line 11: we use #pragma once instead of an include guard

src/openrct2/peep/GuestPathfinding.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "../common.h"
+#include "../ride/RideTypes.h"
```
duncanspumpkin, line 14: why is this included?

richard-fine: For `ride_id_t` on line 31.

duncanspumpkin (approved):

PR #12917, opened 2020-09-12
Debunch peeps

richard-fine:
I've changed the logic in `peep_move_one_tile` to better de-bunch peeps. The previous implementation, giving the tile center as a target but using a randomized DestinationTolerance, never actually worked very well - it would spread peeps out every time they reached a path corner, but as they walked along a straight path, they had gradually increasing probability that they'd be given a low-tolerance destination at the center of the tile - after which they would stay on the center line until reaching the next corner. This behaviour also led to this slightly odd 'course correction' behaviour, usually shortly after a peep turns a corner, like this:

ezgif-1-b5c7a29a7342

I have changed the logic so that instead of randomizing the tolerance, instead we randomize the destination, but only on the axis that the peep is actually travelling. In the other axis, instead of using the tile center, I use their current position. This has the effect that a peep who is offset to one side of the path, will stay on that side of the path until they reach a corner:

ezgif com-resize

Peeps will still bunch up in other situations where the destination is set up (like exiting rides, or getting up from a bench), but I think this is a still a good start.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -52,6 +52,7 @@
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 - Improved: [#13098] Improvements to the maze construction window user interface
 - Improved: [#13125] Selecting the RCT2 files now uses localised dialogs.
+- Improved: Changed peep movement so that they stay more spread out over the full width of paths.
 
 0.3.1 (2020-09-27)
 ------------------------------------------------------------------------
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#13334] Uninitialised variables in CustomTabDesc
 - Fix: [#13342] Rename tabChange to onTabChange in WindowDesc interface
+- Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 
 0.3.2 (2020-11-01)
 ------------------------------------------------------------------------
@@ -49,7 +50,6 @@
 - Fix: [#13278] Desync caused by ghost tiles changing the ride mode.
 - Fix: [#13289] Litter and vomit sometimes not loading with RCT1 saved game or scenario
 - Fix: [#13292] Impossible excitement rating requirements with finish building 5 coasters goal
-- Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 - Improved: [#13098] Improvements to the maze construction window user interface
 - Improved: [#13125] Selecting the RCT2 files now uses localised dialogs.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -52,7 +52,7 @@
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 - Improved: [#13098] Improvements to the maze construction window user interface
 - Improved: [#13125] Selecting the RCT2 files now uses localised dialogs.
-- Improved: Changed peep movement so that they stay more spread out over the full width of paths.
+- Improved: Changed peep movement so that they stay more spread out over the full width of single tile paths.
 
 0.3.1 (2020-09-27)
 ------------------------------------------------------------------------
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -49,10 +49,10 @@
 - Fix: [#13278] Desync caused by ghost tiles changing the ride mode.
 - Fix: [#13289] Litter and vomit sometimes not loading with RCT1 saved game or scenario
 - Fix: [#13292] Impossible excitement rating requirements with finish building 5 coasters goal
+- Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 - Improved: [#13098] Improvements to the maze construction window user interface
 - Improved: [#13125] Selecting the RCT2 files now uses localised dialogs.
-- Improved: Changed peep movement so that they stay more spread out over the full width of single tile paths.
 
 0.3.1 (2020-09-27)
 ------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


richard-fine:
Note that this causes one of the replay tests to fail, so I guess I need to re-record it. Can anyone point me at notes on how to do that?

richard-fine:
PR to update bpb1000Ticks in the replays repository: https://github.com/OpenRCT2/replays/pull/5

duncanspumpkin:
Do we want this sort of change?

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -124,7 +124,31 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
     peep->DestinationTolerance = 2;
     if (peep->State != PEEP_STATE_QUEUING)
     {
-        peep->DestinationTolerance = (scenario_rand() & 7) + 2;
```
duncanspumpkin, line 127: Why get rid of this bit as well? I think keeping the tolerance as well would be best. Then you can clearly mark where your changes diverge from vanilla.

distribution/changelog.txt
```diff
@@ -22,6 +22,7 @@
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
 - Improved: [#12806] Add Esperanto diacritics to the sprite font.
 - Improved: [#12890] Add stroke to lowercase 'L' to differentiate from capital 'I'.
+- Improved: Changed peep movement so that they stay more spread out over the full width of paths.
```
duncanspumpkin, line 25: This would imply they use the full width of a multitile path but you mean the full width of a single path tile.

AaronVanGeffen:
> Do we want this sort of change?

I'm going by the gifs, but it appears to make the routing a bit more organic. Combined with the small scope of the change, I am in favour of the change.

ZehMatt:
> Do we want this sort of change?

I would vote for yes, it seems like a long overdue change to me.

src/openrct2/peep/GuestPathfinding.cpp
richard-fine: We *could* do, but I think that getting rid of it makes it easier to reason about the movement, as well as being better for performance. Instead of having two random factors to take into account for each tile move, you only have one. 

I'm also not sure that the resulting range of values is acceptable; if you roll a 9 here, and a -3 to offset below, the peep will reach their target a full 12 units away from the centre of the tile. I think this will lead to peeps being too far away from the centre line and walking off the sides of paths.

In principle I suppose we could keep this and drop the offset variable below, but I'm not a fan - DestinationTolerance is inherently 2D, but the point of this change is that we only want to allow a random tolerance in one dimension.

distribution/changelog.txt
richard-fine:
```suggestion
- Improved: Changed peep movement so that they stay more spread out over the full width of single tile paths.
```

richard-fine:
So, how do we move forward with this?

tupaschoal:
Seems like no one is against the change, so we can move forward with it. We'll only merge it after `v0.3.1` though.

I guess for now there's only one issue pending an answer from @duncanspumpkin

src/openrct2/peep/GuestPathfinding.cpp
duncanspumpkin: I think having some variability in the destination tolerance makes for a tiny bit of more natural movement.

richard-fine: Can you please clarify EXACTLY what you are expecting by "more natural" movement so I can go and test to see if you are right?

I am still convinced that varying the DestinationTolerance _in addition to_ the target position is going to do nothing but hurt performance.

duncanspumpkin: Calling scenario_rand is not an expensive operation so I don't see there being any issue with calling that more times and even if there was then you could get the random value once and use it for both the x/y offset and the tolerance. Have you tested it with the original line reinstated as well as your change to see how it would look?

richard-fine: This is what it looks like with the original line reinstated:

ezgif com-optimize

As you can see, this results in peeps now being too spread away from the centre lines of the paths, to the point that they walk on the wrong sides of bridge railings etc.

ZehMatt:
I think this is good to go, any objections @OpenRCT2/developers ?

duncanspumpkin:
Yeah okay with it

AaronVanGeffen:
@richard-fine Could you rebase this? :slightly_smiling_face:

richard-fine:
> @richard-fine Could you rebase this? 🙂

Done :)

ZehMatt (approved):

tupaschoal:
(Pending replay fix)

richard-fine:
Note that the failing replay test is fixed by re-recording the replay. I did this in https://github.com/OpenRCT2/replays/pull/5, I don't know what your landing process is between the two repositories.

tupaschoal:
We need to merge it there, then come here and update the code to reference the new replay

distribution/changelog.txt
```diff
@@ -30,6 +30,7 @@
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 - Improved: [#13098] Improvements to the maze construction window user interface
 - Improved: [#13125] Selecting the RCT2 files now uses localised dialogs.
+- Improved: Changed peep movement so that they stay more spread out over the full width of single tile paths.
```
Gymnasiast, line 33: Include the PR number here.

richard-fine: Done

duncanspumpkin:
This is going to have to go into the next version so unfortunately can't do anything about it till after the release.

tupaschoal:
@richard-fine can you please update vcxproj, cmakelists and any other relevant place to point to the newest replay release, which contains your recording?

https://github.com/OpenRCT2/replays/releases/tag/v0.0.20

tupaschoal:
You have to update the replays SHA too

tupaschoal (approved):

tupaschoal:
Sorry that it took us a while to merge it, but thanks for the patience and the PR!

PR #12925, opened 2020-09-13
RFC: Dropdown menus static data

richard-fine:
I'm playing around with refactoring the way dropdown menu items are specified. The current code has a few problems:

- The DDIDX values have to be kept in sync with the actual indices of the elements in the menu. Some menus solve this by setting up the `gDropdownItemsFormat` and `gDropdownItemsArgs` arrays using the DDIDX values as indices, but not all of them do this.
- In addition to the DDIDX value, every menu item typically requires two assignments: one for `gDropdownItemsFormat` and one for `gDropdownItemsArgs`. This is a bit verbose, and has the potential for getting out of sync.
- The values for the format and args arrays are often totally constant, and could just be copied from prepared buffers in the .rodata section; the current way of specifying the options doesn't make it easier for the optimiser to discover that.

I have a proposed solution in this PR. It allows you to build up menu items using constexpr and initialiser_list, which I think is much more concise and readable, and - from what I can tell - also results in slightly more optimal code (at least for Clang 10.0.1 on Godbolt). A static assert is also used to make sure that the IDs and indices are kept in sync.

I've migrated the View and Cheats menus to this style so you can see what the before/after looks like.

Seeking feedback on this idea before I invest more time into it...

### DIFF
--- a/src/openrct2-ui/interface/Dropdown.h
+++ b/src/openrct2-ui/interface/Dropdown.h
@@ -11,6 +11,7 @@
 
 #include <openrct2-ui/interface/Window.h>
 #include <openrct2/common.h>
+#include <openrct2/localisation/StringIds.h>
 
 constexpr const rct_string_id DROPDOWN_SEPARATOR = 0;
 constexpr const rct_string_id DROPDOWN_FORMAT_COLOUR_PICKER = 0xFFFE;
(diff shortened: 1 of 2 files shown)


Gymnasiast:
To me, this looks like a good refactor. Anything that makes our windowing system easier to work with is welcomed, and the slightly better optimised code is a cherry on top.

I have asked for a review from @AaronVanGeffen , who has done most of the recent refactors on the windowing system.

AaronVanGeffen (commented):
Thanks for this PR! I like that we find these parts of the code that could look nicer and then tackle the job.

From a high level point of view, I really like what you've done here. I appreciate that you've looked into the impact this has on the generated code — that's my main concern with this sort of refactor.

When you think it's ready, I think we should go ahead and merge this PR as-is, and convert the other dropdowns in another PR. Like the new widget constructor functions, the two can live side-by-side for a while.

richard-fine:
As far as I'm concerned, this is good to go - so, merge at your leisure.

PR #12933, opened 2020-09-15
More game action visitors added (all Park, Staff, Guest, Banner)

tombomp:
There's a lot of files changed here but it's all pretty rote stuff. There's still a few involving a MapRange that don't have a visitor, this is just all the simple ones. 

With the names if I noticed an already existing similar one was using a name different to the variable I used that, otherwise I mostly just copied the variable name. Obviously any better naming suggestions are welcome.

### DIFF
--- a/src/openrct2/actions/BannerSetStyleAction.hpp
+++ b/src/openrct2/actions/BannerSetStyleAction.hpp
@@ -42,6 +42,13 @@ public:
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
+    {
+        visitor.Visit("type", _type);
+        visitor.Visit("bannerIndex", _bannerIndex);
+        visitor.Visit("parameter", _parameter);
+    }
+
     uint16_t GetActionFlags() const override
     {
         return GameAction::GetActionFlags() | GA_FLAGS::ALLOW_WHILE_PAUSED;
--- a/src/openrct2/actions/FootpathPlaceFromTrackAction.hpp
+++ b/src/openrct2/actions/FootpathPlaceFromTrackAction.hpp
@@ -40,6 +40,14 @@ public:
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
+    {
+        visitor.Visit(_loc);
+        visitor.Visit("object", _type);
+        visitor.Visit("edges", _edges);
+        visitor.Visit("slope", _slope);
+    }
+
     uint16_t GetActionFlags() const override
     {
         return GameAction::GetActionFlags();
--- a/src/openrct2/actions/LoadOrQuitAction.hpp
+++ b/src/openrct2/actions/LoadOrQuitAction.hpp
@@ -36,6 +36,12 @@ public:
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
+    {
+        visitor.Visit("mode", _mode);
+        visitor.Visit("savePromptMode", _savePromptMode);
+    }
+
     uint16_t GetActionFlags() const override
     {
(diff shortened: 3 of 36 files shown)


src/openrct2/actions/BannerSetStyleAction.hpp
```diff
@@ -42,6 +42,13 @@ DEFINE_GAME_ACTION(BannerSetStyleAction, GAME_COMMAND_SET_BANNER_STYLE, GameActi
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
+    {
+        visitor.Visit("type", _type);
```
duncanspumpkin, line 47: Hmm what would plugin users pass for this?

src/openrct2/actions/FootpathPlaceFromTrackAction.hpp
```diff
@@ -40,6 +40,14 @@ DEFINE_GAME_ACTION(FootpathPlaceFromTrackAction, GAME_COMMAND_PLACE_PATH_FROM_TR
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
```
duncanspumpkin, line 43: Do not expose this action it should never be used by plugins it is only used by TrackDesigns.

src/openrct2/actions/LoadOrQuitAction.hpp
```diff
@@ -36,6 +36,12 @@ DEFINE_GAME_ACTION(LoadOrQuitAction, GAME_COMMAND_LOAD_OR_QUIT, GameActionResult
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
```
duncanspumpkin, line 39: Debatable if this is safe to perform from a plugin. Pretty sure this will unload the plugin.

src/openrct2/actions/ParkSetParameterAction.hpp
```diff
@@ -40,6 +40,12 @@ DEFINE_GAME_ACTION(ParkSetParameterAction, GAME_COMMAND_SET_PARK_OPEN, GameActio
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
+    {
+        visitor.Visit("parameter", _parameter);
```
duncanspumpkin, line 45: this action does not follow the order of other commands. value should be parameter and parameter should be some sort of type

src/openrct2/actions/ScenarioSetSettingAction.hpp
```diff
@@ -60,6 +60,12 @@ DEFINE_GAME_ACTION(ScenarioSetSettingAction, GAME_COMMAND_EDIT_SCENARIO_OPTIONS,
     {
     }
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override
+    {
+        visitor.Visit("setting", _setting);
+        visitor.Visit("value", _value);
```
duncanspumpkin, line 66: parameter?

duncanspumpkin:
There are a few here that I do not think should be usable by plugins and others that are dangerous to use. A lot of the parameters need some careful thought on their names as we should standardise it as much as possible to prevent issues with later renaming.

Gymnasiast:
@tombomp Are you still working on this?

tombomp:
> @tombomp Are you still working on this?

Ah sorry real life got in the way but I'm not, sorry. I'll close it, sorry for not finishing it. Some of these have been added by other PRs anyway and the rest has enough issues someone else should probably start from scratch re naming and which ones are worth exposing. Sorry once again.

tupaschoal:
No need to be sorry, it happens. Thanks for letting us know and feel free to come back when you're freer :)

Issue #12842, opened 2020-09-02
Add Booster and Brake speed setting via Tile Inspector


maddkattunge:
Currently, the only way to modify the booster and brake track speeds is from the track editor window, requiring the user to close the ride and reset the trains. On the other hand, chain speed can be set from the Operating Options tab of the ride window without closing the ride, and chains can be added or removed from specific track pieces via a toggle in the Tile Inspector, allowing for easy fine-tuning of the chains. I propose adding the ability to set speed for selected booster or brake track via the Tile Inspector, working around the need to close the ride and thus streamlining the process of fine tuning the speed of loaded trains operating normally, as well as shoestrung trains and hacked trains that can be difficult to reset.

PR #12946, opened 2020-09-17
Feature #12842: Add Booster and Brake speed setting via Tile Inspector

frutiemax:


src/openrct2/actions/TileModifyAction.hpp
```diff
@@ -95,11 +96,14 @@ DEFINE_GAME_ACTION(TileModifyAction, GAME_COMMAND_MODIFY_TILE, GameActionResult)
             return MakeResult(GA_ERROR::INVALID_PARAMETERS, STR_LAND_NOT_OWNED_BY_PARK);
         }
         auto res = MakeResult();
+
+        uint8_t speed;
+        uint32_t elementIndex;
```
Gymnasiast, line 101: There is no need for this. Use brackets inside your `case` statement, like all other cases already do.

src/openrct2/localisation/StringIds.h
```diff
@@ -3887,6 +3887,8 @@ enum
     STR_HEAVY_SNOW = 6385,
     STR_BLIZZARD = 6386,
 
+    STR_TILE_INSPECTOR_TRACK_SPEED = 6387
```
Gymnasiast, line 3890: Always use a trailing comma in enums.

src/openrct2/actions/TileModifyAction.hpp
```diff
@@ -158,104 +162,110 @@ DEFINE_GAME_ACTION(TileModifyAction, GAME_COMMAND_MODIFY_TILE, GameActionResult)
...
             {
-                const auto elementIndex = _value1;
+                elementIndex = _value1;
                 const bool setChain = _value2;
                 res = tile_inspector_track_set_chain(_loc, elementIndex, false, setChain, isExecuting);
                 break;
             }
+            case TileModifyType::TrackSetSpeed:
+                elementIndex = _value1;
+                assert(_value2 <= 0xFFFF);
```
duncanspumpkin, line 227: why check that is less than 0xFFFF when you then cast it to uint8_t. You should not have asserts in code like this either. What happens if a malicous individual sends a command to modify the track speed with a value > 0xFFFF, it crashes the server. Always validate safely and never throw an un caught error in action code.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1944,10 +2013,31 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                        screenCoords.y = w->windowPos.y + w->widgets[WIDX_TRACK_SPINNER_SPEED].top;
+                        gfx_draw_string_left(dpi, STR_TILE_INSPECTOR_TRACK_SPEED, nullptr, COLOUR_WHITE, screenCoords);
+
+                        // brakes/booster speed spinner value
+                        screenCoords.x = w->windowPos.x + w->widgets[WIDX_TRACK_SPINNER_SPEED].left + 3;
+                        screenCoords.y = w->windowPos.y + w->widgets[WIDX_TRACK_SPINNER_SPEED].top;
+                        int32_t speed = static_cast<int32_t>(tileElement->AsTrack()->GetBrakeBoosterSpeed());
+
+                        // convert the speed to a readable format
+                        gfx_draw_string_left(dpi, STR_FORMAT_INTEGER, &speed, COLOUR_WHITE, screenCoords);
```
duncanspumpkin, line 2031: always use a Formatter instead of passing the value directly as we are phasing out the void* pointer versions of these functions.

Gymnasiast: Also, `gfx_draw_string_left()` is deprecated. Use `DrawTextBasic()` instead.

data/language/en-GB.txt
```diff
@@ -3642,6 +3642,7 @@ STR_6383    :Open download page
 STR_6384    :Snow
 STR_6385    :Heavy Snow
 STR_6386    :Blizzard
+STR_6387    :Brake/Boost speed
```
AaronVanGeffen, line 3645: The Tile Inspector uses colons after headers, i.e.:
```suggestion
STR_6387    :Brake/Boost speed:
```

AaronVanGeffen: No need to amend and force push after every change, btw. We can squash everything nicely after reviewing.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1944,10 +2021,33 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                        screenCoords.y = w->windowPos.y + w->widgets[WIDX_TRACK_SPINNER_SPEED].top;
+                        gfx_draw_string_left(dpi, STR_TILE_INSPECTOR_TRACK_SPEED, nullptr, COLOUR_WHITE, screenCoords);
+
+                        // brakes/booster speed spinner value
+                        screenCoords.x = w->windowPos.x + w->widgets[WIDX_TRACK_SPINNER_SPEED].left + 3;
+                        screenCoords.y = w->windowPos.y + w->widgets[WIDX_TRACK_SPINNER_SPEED].top;
+                        int32_t speed = static_cast<int32_t>(tileElement->AsTrack()->GetBrakeBoosterSpeed());
+
+                        // convert the speed to a readable format
+                        auto ft = Formatter().Add<int32_t>(speed);
```
duncanspumpkin, line 2039: You're going to hit the copy constructor bug if the optimiser wants to mess with this. Make this two separate calls. I'm going to fix the copy constructor as well at some point but for now split this up.

src/openrct2/actions/TileModifyAction.hpp
```diff
@@ -95,11 +96,12 @@ DEFINE_GAME_ACTION(TileModifyAction, GAME_COMMAND_MODIFY_TILE, GameActionResult)
             return MakeResult(GA_ERROR::INVALID_PARAMETERS, STR_LAND_NOT_OWNED_BY_PARK);
         }
         auto res = MakeResult();
+
         switch (static_cast<TileModifyType>(_setting))
         {
             case TileModifyType::AnyRemove:
             {
-                const auto elementIndex = _value1;
+                uint32_t elementIndex = _value1;
```
Gymnasiast, line 104: Like I said: please revert these changes that have nothing to do with your addition.

src/openrct2/actions/TileModifyAction.hpp
```diff
@@ -158,104 +160,114 @@ DEFINE_GAME_ACTION(TileModifyAction, GAME_COMMAND_MODIFY_TILE, GameActionResult)
...
                 const bool setChain = _value2;
                 res = tile_inspector_track_set_chain(_loc, elementIndex, false, setChain, isExecuting);
                 break;
             }
+            case TileModifyType::TrackSetSpeed:
+            {
+                uint32_t elementIndex = _value1;
```
Gymnasiast, line 225: Use `const auto`, like the rest of the code.

Gymnasiast (changes requested):
Getting closer, but there are still some things to address.

duncanspumpkin (approved):

Gymnasiast (approved):

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1555,12 +1589,55 @@ static void window_tile_inspector_invalidate(rct_window* w)
...
+                w->widgets[WIDX_TRACK_SPINNER_SPEED].type = WWT_EMPTY;
+                w->widgets[WIDX_TRACK_SPINNER_SPEED_INCREASE].type = WWT_EMPTY;
+                w->widgets[WIDX_TRACK_SPINNER_SPEED_DECREASE].type = WWT_EMPTY;
+                w->widgets[WIDX_TRACK_CHECK_CHAIN_LIFT].top = GBBT(propertiesAnchor, 2);
+                w->widgets[WIDX_TRACK_CHECK_CHAIN_LIFT].bottom = GBBB(propertiesAnchor, 2);
+                w->widgets[WIDX_TRACK_CHECK_BLOCK_BRAKE_CLOSED].top = GBBT(propertiesAnchor, 3);
+                w->widgets[WIDX_TRACK_CHECK_BLOCK_BRAKE_CLOSED].bottom = GBBB(propertiesAnchor, 3);
+                w->widgets[WIDX_TRACK_CHECK_IS_INDESTRUCTIBLE].top = GBBT(propertiesAnchor, 4);
+                w->widgets[WIDX_TRACK_CHECK_IS_INDESTRUCTIBLE].bottom = GBBB(propertiesAnchor, 4);
+            }
```
Broxzier, line 1639: This entire if/(if/else)/else block is there because you're hiding the widget for track elements that aren't boosters or brakes, right? Imagine how this scales if we add another widget that needs hiding. The code would be much simpler if the widget simply gets disabled when it cannot be used for a specific element, just like how it's done with the frame animation for small walls for instance. It'd also prevent the interface from changing for elements of the same type.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1949,10 +2026,34 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                        screenCoords.y = w->windowPos.y + w->widgets[WIDX_TRACK_SPINNER_SPEED].top;
+                        int32_t speed = static_cast<int32_t>(tileElement->AsTrack()->GetBrakeBoosterSpeed());
+
+                        // convert the speed to a readable format
+                        auto ft = Formatter();
+                        ft.Add<int32_t>(speed);
+
+                        DrawTextBasic(dpi, screenCoords, STR_FORMAT_INTEGER, ft, COLOUR_WHITE);
+                    }
+                }
```
Broxzier, line 2049: Following with the above comment, this needs adjusting as well. Simply draw a "-" inside the disabled widget here.

src/openrct2/actions/TileModifyAction.hpp
```diff
@@ -218,6 +220,16 @@ DEFINE_GAME_ACTION(TileModifyAction, GAME_COMMAND_MODIFY_TILE, GameActionResult)
...
+            case TileModifyType::TrackSetSpeed:
+            {
+                const auto elementIndex = _value1;
+                uint8_t speed = 0xFF;
+                if (_value2 <= 0xFF)
+                    speed = static_cast<uint8_t>(_value2);
+                res = tile_inspector_track_set_speed(_loc, elementIndex, speed, isExecuting);
+                break;
+            }
+
```
Broxzier, line 232: nitpicky, but none of the other cases have an empty new line afterwards.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1555,12 +1589,55 @@ static void window_tile_inspector_invalidate(rct_window* w)
...
             w->widgets[WIDX_TRACK_SPINNER_HEIGHT_DECREASE].top = GBBT(propertiesAnchor, 1) + 4;
             w->widgets[WIDX_TRACK_SPINNER_HEIGHT_DECREASE].bottom = GBBB(propertiesAnchor, 1) - 4;
-            w->widgets[WIDX_TRACK_CHECK_CHAIN_LIFT].top = GBBT(propertiesAnchor, 2);
-            w->widgets[WIDX_TRACK_CHECK_CHAIN_LIFT].bottom = GBBB(propertiesAnchor, 2);
-            w->widgets[WIDX_TRACK_CHECK_BLOCK_BRAKE_CLOSED].top = GBBT(propertiesAnchor, 3);
-            w->widgets[WIDX_TRACK_CHECK_BLOCK_BRAKE_CLOSED].bottom = GBBB(propertiesAnchor, 3);
-            w->widgets[WIDX_TRACK_CHECK_IS_INDESTRUCTIBLE].top = GBBT(propertiesAnchor, 4);
-            w->widgets[WIDX_TRACK_CHECK_IS_INDESTRUCTIBLE].bottom = GBBB(propertiesAnchor, 4);
+
+            // check if track type is either a brake or boost section
```
Broxzier, line 1593:
```suggestion
            // Check if track type is either a brake or boost section
```

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -558,6 +562,14 @@ static void window_tile_inspector_select_element_from_list(rct_window* w, int32_
     else
     {
         windowTileInspectorSelectedIndex = index;
+
+        // update the brake speed to current one
```
Broxzier, line 566:
```suggestion
        // Update the brake speed to the current one
```

src/openrct2/world/TileInspector.cpp
```diff
@@ -35,6 +35,7 @@ uint32_t windowTileInspectorTileX;
 uint32_t windowTileInspectorTileY;
 int32_t windowTileInspectorElementCount = 0;
 int32_t windowTileInspectorSelectedIndex;
+uint8_t windowTileInspectorTrackSpeed;
```
Broxzier, line 38: Get rid of this, there is no need to store this value anywhere, this data already exists in the element itself. This just introduces code where you need to set it (currently on selecting a tile), making it quite phrone to errors (e.g. when refactoring or introducing a new way of selecting tiles for instance).

Broxzier (changes requested):
There is quite some code bloat that should be reduced. I haven't tested this in-game yet.

src/openrct2-ui/windows/TileInspector.cpp
Broxzier: This should be removed entirely once you get rid of the global variable `windowTileInspectorTrackSpeed` though.

Gymnasiast:
@frutiemax Are you still working on this?

Broxzier:
@frutiemax I recall you wanting to see this merged. Any plans on continuing this?

Gymnasiast:
Closing this as it’s inactive.

PR #12965, opened 2020-09-18
Allow compilation with nlohmann/json < 3.9.0

Gymnasiast:
According to the [API docs](https://github.com/nlohmann/json/blob/develop/single_include/nlohmann/json.hpp#L23219-L23222). the fourth argument (which allows ignoring comments) was only introduced in version 3.9.0 (released on 27 July 2020), which is not present in Ubuntu 20.04 (it only has 3.7.3-1).

Since comments are not part of the JSON spec, this option is not terribly useful anyway. I have removed it, so that it can now compile on Ubuntu 20.04, and possibly also earlier versions.

### DIFF
--- a/src/openrct2/core/Json.cpp
+++ b/src/openrct2/core/Json.cpp
@@ -32,7 +32,7 @@ namespace Json
 
         try
         {
-            json = json_t::parse(fileData, nullptr, true);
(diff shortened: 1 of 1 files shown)


mwnciau:
Might as well remove the other two parameters which are defaults as part of this change, too.

PR #12966, opened 2020-09-18
Replace gfx_draw_string_right() with DrawTextBasic

Gymnasiast:


tupaschoal (approved):

src/openrct2-ui/windows/Finances.cpp
```diff
@@ -769,18 +769,25 @@ static void window_finances_summary_scrollpaint(rct_window* w, rct_drawpixelinfo
...
             screenCoords.y += TABLE_CELL_HEIGHT;
         }
         screenCoords.y += 4;
 
         // Month profit
-        gfx_draw_string_right(
-            dpi, profit >= 0 ? STR_FINANCES_SUMMARY_INCOME_VALUE : STR_FINANCES_SUMMARY_LOSS_VALUE, &profit, COLOUR_BLACK,
-            screenCoords + ScreenCoordsXY{ EXPENDITURE_COLUMN_WIDTH, 0 });
+        const rct_string_id format = profit >= 0 ? STR_FINANCES_SUMMARY_INCOME_VALUE : STR_FINANCES_SUMMARY_LOSS_VALUE;
+        ft = Formatter();
```
AaronVanGeffen, line 786: I believe this causes the problem with #12968. We may need to explicitly scope it, at least for now.

PR #12971, opened 2020-09-19
Remove Map tooltip format args global

duncanspumpkin:
Following on from #12970 this removes one of the global format args and replaces it with a global Formatter but protected by accessor functions.


tupaschoal:
Can you rebase to drop the other commit that has been merged?

src/openrct2-ui/windows/MapTooltip.cpp
```diff
@@ -64,6 +64,18 @@ static int32_t _cursorHoldDuration;
 
 static void window_map_tooltip_open();
 
+static Formatter _mapTooltipArgs;
```
Gymnasiast, line 67: Effectively, this is still a global. Is there a plan to get rid of this variable eventually?

duncanspumpkin: It needs to stay but eventually it would become part of the window.

Gymnasiast (approved):

PR #12972, opened 2020-09-19
Remove Formatter::Common where not required

duncanspumpkin:
~~Note will likely have issues if merged before #12970~~

On second thoughts this will turn into a massive PR if I try do everything in one go. Therefore this is all I will change in this one PR.

### DIFF
--- a/src/openrct2-ui/interface/Graph.cpp
+++ b/src/openrct2-ui/interface/Graph.cpp
@@ -26,10 +26,9 @@ static void graph_draw_months_uint8_t(
         if (history[i] != 255 && yearOver32 % 4 == 0)
         {
             // Draw month text
-            auto ft = Formatter::Common();
+            auto ft = Formatter();
             ft.Add<uint32_t>(DateGameShortMonthNames[date_get_month((yearOver32 / 4) + MONTH_COUNT)]);
-            gfx_draw_string_centred(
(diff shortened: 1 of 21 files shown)


src/openrct2-ui/interface/Graph.cpp
```diff
@@ -26,10 +26,9 @@ static void graph_draw_months_uint8_t(
         if (history[i] != 255 && yearOver32 % 4 == 0)
         {
             // Draw month text
-            auto ft = Formatter::Common();
+            auto ft = Formatter();
             ft.Add<uint32_t>(DateGameShortMonthNames[date_get_month((yearOver32 / 4) + MONTH_COUNT)]);
-            gfx_draw_string_centred(
-                dpi, STR_GRAPH_LABEL, screenCoords - ScreenCoordsXY{ 0, 10 }, COLOUR_BLACK, gCommonFormatArgs);
+            gfx_draw_string_centred(dpi, STR_GRAPH_LABEL, screenCoords - ScreenCoordsXY{ 0, 10 }, COLOUR_BLACK, ft.Data());
```
Gymnasiast, line 31: You could use `DrawTextBasic`, so that you can pass `ft` directly.

Gymnasiast (approved):
Looks good. Might need to wait until after the 0.3.1 merge.

src/openrct2-ui/interface/Graph.cpp
duncanspumpkin: I thought about that as well and sort of came to the same conclusion. I think there is less chance of error doing this in two steps

PR #12973, opened 2020-09-19
Compact syntax for event lists

AaronVanGeffen:
Currently, the window event list struct only has one (default) constructor, which leads to a bunch of null-pointers having to be set for most windows. It's not obvious which events they represent, which leaves these lists hard to maintain. Unfortunately, we cannot use named initialisers. Therefore, I propose a replacement set of definitions for the event lists using functions, giving them a syntax [similar to OpenLoco](https://github.com/OpenLoco/OpenLoco/blob/9b419912746c1b4930b1165e6dd4c149aeb48704/src/OpenLoco/Windows/Options.cpp#L633-L643).

I generated all of this using [a Python script](https://gist.github.com/AaronVanGeffen/110509418eca7bd22f6fbfb93c1406d3), which hopefully reduces the chance of errors. From brief testing on my end, all appears to work correctly. However, we should test this properly before merging.

### DIFF
--- a/src/openrct2-ui/windows/Cheats.cpp
+++ b/src/openrct2-ui/windows/Cheats.cpp
@@ -333,65 +333,41 @@ static void window_cheats_paint(rct_window *w, rct_drawpixelinfo *dpi);
 static void window_cheats_set_page(rct_window *w, int32_t page);
 static void window_cheats_text_input(rct_window *w, rct_widgetindex widgetIndex, char *text);
 
-static rct_window_event_list window_cheats_money_events = {};
-
-static void window_cheats_money_events_init()
+static rct_window_event_list window_cheats_money_events([](auto& events)
 {
-    auto& events = window_cheats_money_events;
+    events.mouse_up = &window_cheats_money_mouseup;
+    events.mouse_down = &window_cheats_money_mousedown;
+    events.update = &window_cheats_update;
+    events.text_input = &window_cheats_text_input;
+    events.invalidate = &window_cheats_invalidate;
+    events.paint = &window_cheats_paint;
+});
 
-    events.mouse_up = window_cheats_money_mouseup;
-    events.mouse_down = window_cheats_money_mousedown;
-    events.update = window_cheats_update;
-    events.text_input = window_cheats_text_input;
-    events.invalidate = window_cheats_invalidate;
-    events.paint = window_cheats_paint;
-}
-
-static rct_window_event_list window_cheats_guests_events = {};
-
-static void window_cheats_guests_events_init()
+static rct_window_event_list window_cheats_guests_events([](auto& events)
 {
-    auto& events = window_cheats_guests_events;
+    events.mouse_up = &window_cheats_guests_mouseup;
+    events.update = &window_cheats_update;
+    events.invalidate = &window_cheats_invalidate;
+    events.paint = &window_cheats_paint;
+});
 
-    events.mouse_up = window_cheats_guests_mouseup;
-    events.update = window_cheats_update;
-    events.invalidate = window_cheats_invalidate;
-    events.paint = window_cheats_paint;
-}
-
-static rct_window_event_list window_cheats_misc_events = {};
-
-static void window_cheats_misc_events_init()
+static rct_window_event_list window_cheats_misc_events([](auto& events)
 {
-    auto& events = window_cheats_misc_events;
+    events.mouse_up = &window_cheats_misc_mouseup;
+    events.mouse_down = &window_cheats_misc_mousedown;
+    events.dropdown = &window_cheats_misc_dropdown;
+    events.update = &window_cheats_update;
+    events.invalidate = &window_cheats_invalidate;
+    events.paint = &window_cheats_paint;
+});
 
-    events.mouse_up = window_cheats_misc_mouseup;
-    events.mouse_down = window_cheats_misc_mousedown;
-    events.dropdown = window_cheats_misc_dropdown;
-    events.update = window_cheats_update;
-    events.invalidate = window_cheats_invalidate;
-    events.paint = window_cheats_paint;
-}
-
-static rct_window_event_list window_cheats_rides_events = {};
-
-static void window_cheats_rides_events_init()
+static rct_window_event_list window_cheats_rides_events([](auto& events)
 {
-    auto& events = window_cheats_rides_events;
-
-    events.mouse_up = window_cheats_rides_mouseup;
-    events.update = window_cheats_update;
-    events.invalidate = window_cheats_invalidate;
-    events.paint = window_cheats_paint;
-}
-
-static void window_cheats_events_init_all()
-{
-    window_cheats_money_events_init();
-    window_cheats_guests_events_init();
-    window_cheats_misc_events_init();
-    window_cheats_rides_events_init();
-}
+    events.mouse_up = &window_cheats_rides_mouseup;
+    events.update = &window_cheats_update;
+    events.invalidate = &window_cheats_invalidate;
+    events.paint = &window_cheats_paint;
+});
 
 
 static rct_window_event_list *window_cheats_page_events[] =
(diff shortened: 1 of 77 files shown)


ZehMatt:
Since you already touch quite a bit of the code would be nice to have them in PascalCase/camelCase where appropriate.

AaronVanGeffen:
I agree that would be nice, but I think that is really something for another PR. Preferably, we would rename the event functions into namespaces at the same time.

ZehMatt:
We can probably avoid explicit initialization with something like this: https://godbolt.org/z/PqaKPf , what do you think about that?

AaronVanGeffen:
I think it would be great if we can avoid explicitly calling an initialisation function. Your snippet looks like a great start in this respect. However, I am not sure I can adapt it directly. The problem I'm seeing is that the event handlers are stored as a pointer:
```cpp
struct rct_window
{
    rct_window_event_list* event_handlers;
```

Currently the event lists are passed as follows:
```
        window = window_create_auto_pos(windowWidth, 157, &window_guest_overview_events, WC_PEEP, WF_RESIZABLE);
```

I am wondering where the constructor (below) would be called here. Can we avoid making the assignment more explicit than this?
```cpp
    rct_window_event_list(auto fn)
    {
        fn(*this);
    }
```

src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -1332,4 +1265,4 @@ static void window_cheats_set_page(rct_window* w, int32_t page)
     w->widgets[WIDX_BACKGROUND].bottom = maxY - 1;
     w->widgets[WIDX_PAGE_BACKGROUND].bottom = maxY - 1;
     w->Invalidate();
-}
+}
```
Broxzier, line 1268: EOL

src/openrct2-ui/windows/ClearScenery.cpp
```diff
@@ -253,4 +238,4 @@ static void window_clear_scenery_paint(rct_window* w, rct_drawpixelinfo* dpi)
         screenCoords.y = window_clear_scenery_widgets[WIDX_PREVIEW].bottom + w->windowPos.y + 5 + 27;
         gfx_draw_string_centred(dpi, STR_COST_AMOUNT, screenCoords, COLOUR_BLACK, &gClearSceneryCost);
     }
-}
+}
```
Broxzier, line 241: EOL

src/openrct2-ui/windows/CustomCurrency.cpp
```diff
@@ -274,4 +257,4 @@ static void custom_currency_window_paint(rct_window* w, rct_drawpixelinfo* dpi)
                 + ScreenCoordsXY{ window_custom_currency_widgets[WIDX_AFFIX_DROPDOWN].left + 1,
                                   window_custom_currency_widgets[WIDX_AFFIX_DROPDOWN].top });
     }
-}
+}
```
Broxzier, line 260: EOL

src/openrct2-ui/windows/DebugPaint.cpp
```diff
@@ -184,4 +165,4 @@ static void window_debug_paint_invalidate(rct_window* w)
 static void window_debug_paint_paint(rct_window* w, rct_drawpixelinfo* dpi)
 {
     window_draw_widgets(w, dpi);
-}
+}
```
Broxzier, line 168: EOL

src/openrct2-ui/windows/DemolishRidePrompt.cpp
```diff
@@ -240,4 +208,4 @@ static void window_ride_refurbish_paint(rct_window* w, rct_drawpixelinfo* dpi)
         ScreenCoordsXY stringCoords(w->windowPos.x + WW / 2, w->windowPos.y + (WH / 2) - 3);
         gfx_draw_string_centred_wrapped(dpi, gCommonFormatArgs, stringCoords, WW - 4, stringId, COLOUR_BLACK);
     }
-}
+}
```
Broxzier, line 211: EOL

src/openrct2-ui/windows/Dropdown.cpp
```diff
@@ -471,4 +451,4 @@ void window_dropdown_show_colour(rct_window* w, rct_widget* widget, uint8_t drop
 uint32_t dropdown_get_appropriate_image_dropdown_items_per_row(uint32_t numItems)
 {
     return numItems < std::size(_appropriateImageDropdownItemsPerRow) ? _appropriateImageDropdownItemsPerRow[numItems] : 8;
-}
+}
```
Broxzier, line 454: EOL

src/openrct2-ui/windows/EditorBottomToolbar.cpp
```diff
@@ -522,4 +503,4 @@ void window_editor_bottom_toolbar_paint(rct_window* w, rct_drawpixelinfo* dpi)
             gfx_draw_string_centred(dpi, stringId, { textX, textY + 10 }, textColour, nullptr);
         }
     }
-}
+}
```
Broxzier, line 506: EOL

src/openrct2-ui/windows/EditorInventionsList.cpp
```diff
@@ -846,4 +822,4 @@ static std::pair<rct_string_id, Formatter> window_editor_inventions_list_prepare
     return std::make_pair(drawString, ft);
 }
 
-#pragma endregion
+#pragma endregion
```
Broxzier, line 825: EOL

Broxzier (changes requested):
Left a few comments on removed EOLs, this happened automatically by your editor, I presume. `insert_final_newline` is set in our `.editorconfig` file, so your editor seems to ignore this.

AaronVanGeffen:
@Broxzier Ah, you're right, I hadn't noticed. This is entirely due to my Python script omitting them in its output — an oversight. I will amend. (It's a pity `clang-format` doesn't tackle this.)

ZehMatt:
> I think it would be great if we can avoid explicitly calling an initialisation function. Your snippet looks like a great start in this respect. However, I am not sure I can adapt it directly. The problem I'm seeing is that the event handlers are stored as a pointer:
> 
> ```c++
> struct rct_window
> {
>     rct_window_event_list* event_handlers;
> ```
> 
> Currently the event lists are passed as follows:
> 
> ```
>         window = window_create_auto_pos(windowWidth, 157, &window_guest_overview_events, WC_PEEP, WF_RESIZABLE);
> ```
> 
> I am wondering where the constructor (below) would be called here. Can we avoid making the assignment more explicit than this?
> 
> ```c++
>     rct_window_event_list(auto fn)
>     {
>         fn(*this);
>     }
> ```

The construction should happen at the beginning of the application start, static initialization happens via a list the compiler constructed and iterates thru the constructors before any code is usually called, so that should be fine.

AaronVanGeffen:
@ZehMatt Unfortunately, it seems your example only works if C++20 concepts are available. When I change to C++17 in Godbolt, it no longer compiles:

```
error: use of ‘auto’ in parameter declaration only available with ‘-fconcepts-ts’ [-Werror]
  245 |     rct_window_event_list(auto fn)
      |                           ^~~~
```

Any ideas?

ZehMatt:
> @ZehMatt Unfortunately, it seems your example only works if C++20 concepts are available. When I change to C++17 in Godbolt, it no longer compiles:
> 
> ```
> error: use of ‘auto’ in parameter declaration only available with ‘-fconcepts-ts’ [-Werror]
>   245 |     rct_window_event_list(auto fn)
>       |                           ^~~~
> ```
> 
> Any ideas?

https://godbolt.org/z/Gv7Gn6 😏

AaronVanGeffen:
I think I've got it. Even more impressive diffstat now :+1:

ZehMatt (approved):

PR #12975, opened 2020-09-20
Fix #12974: unable to build fences on slopes

### LINKED ISSUE #12974: Unable to build sloped walls or fences
**OS:** Windows 10
**Version:** 0.3.0
**Commit/Build:** 124161b

Sloped walls and fences (for fence and wall objects that support the ability to be built on a slope) now can no longer be built even if the "Disable clearance checks" cheat is disabled. I suspect this is yet another recently introduced regression.

**Steps to reproduce:**
1. Attempt to place down a fence or wall on a slope (e.g. Roman Column Wall).
2. It will block you from placing down the fence.

**Screenshot:**
Hesperides Tree 2020-09-19 17-23-35

richard-fine:
Fix the casing of 'isAllowedOnSlope' to match the way it is cased in the JSON files, which causes the flag to be read correctly and re-enables building on slopes for the scenery pieces which have it specified.

### DIFF
--- a/src/openrct2/object/WallObject.cpp
+++ b/src/openrct2/object/WallObject.cpp
@@ -114,7 +114,7 @@ void WallObject::ReadJson(IReadObjectContext* context, json_t& root)
             properties,
             {
                 { "hasPrimaryColour",       WALL_SCENERY_HAS_PRIMARY_COLOUR,    Json::FlagType::Normal },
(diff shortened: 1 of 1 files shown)


richard-fine:
I've not added a CHANGELOG entry for this because it's fixing an issue that was introduced in 0.3.0+. Let me know if that's the wrong thing to do.

Fred-1044276:
A changelog entry isn't necessary.

duncanspumpkin (approved):

PR #12977, opened 2020-09-20
Ensure ccache's expected dirs exist

janisozaur:
This should address the intermittent issue where builds fail because of missing this file.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,12 +207,9 @@ jobs:
       - name: Get pre-reqs
         shell: bash
         run: . scripts/setenv -q && get-discord-rpc
-      - name: Fake ccache
-        shell: bash
-        run: mkdir -p $HOME/.ccache && touch $HOME/.ccache/ccache.conf
       - name: Build OpenRCT2
         shell: bash
-        run: . scripts/setenv -q && build -DCMAKE_BUILD_TYPE=Release -DAPPIMAGE=ON
+        run: . scripts/setenv -q && build -DCMAKE_BUILD_TYPE=Release -DAPPIMAGE=ON -DOPENRCT2_USE_CCACHE=off
       - name: Build AppImage
         shell: bash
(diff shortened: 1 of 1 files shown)


janisozaur:
Hmm, maybe it would be better to set `OPENRCT2_USE_CCACHE` to off instead.

AaronVanGeffen:
> Hmm, maybe it would be better to set `OPENRCT2_USE_CCACHE` to off instead.

Probably, yes. I don't think we have any persistent ccache storage on GitHub Actions.

tupaschoal (approved):
I wouldn't expect ccache to work on GH actions at all (it doesn't make sense, so I think we're really better off with disabling it)

janisozaur:
Why not? It's perfectly sound expectation and we used it with Travis to great results.

tupaschoal:
Presumably GitHub would have to be storing the cache somewhere, which I haven't seen yet.

I have tried porting a cached ruby setup from Travis to ghactions and there was no way, so I'm just assuming it isn't there (yet?)

janisozaur:
Oh it's just a matter of using https://github.com/actions/cache and configuring the restore keys and paths correctly.

janisozaur:
There's also https://cristianadam.eu/20200113/speeding-up-c-plus-plus-github-actions-using-ccache/ to be used for reference

tupaschoal:
Nice! I don't think it greatly benefit us all, as we have other setups that don't accept `ccache`, so it would just speed up a few.
But if you think it's worth trying and implementing it, maybe we can open an issue for the community to tackle, what do you say?

janisozaur:
Oh yes, you should always use ccache, I find it hard to think of scenario that wouldn't benefit from it.

Especially in CI, where most builds are incremental. Cf. https://github.com/scummvm/scummvm/pull/2243 or https://github.com/RPCS3/rpcs3/pull/2448 for some stats

PR #12982, opened 2020-09-21
Fix #12918: Cannot place "Blue Hurricane" (hypercoaster)

### LINKED ISSUE #12918: Cannot place "Blue Hurricane" (hyper coaster)
Even on completely flattened land, when trying to place the prebuilt coaster "Blue Hurricane" it says "Cannot build that here".

Gymnasiast:
The track design was recognised as a Corkscrew RC, not as a Hypercoaster. Moved the conversion code from the track design repository to the TD6Importer (where it should have been, really).

This also fixes the issue that Hypercoasters, Monster Trucks, Classic Mini Roller Coasters, Spinning Wild Mouses and Hyper-Twisters placed from a track design have the wrong ride type.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -24,7 +24,7 @@
 - Fix: [#12881] Guests' favourite rides are not listed in the guest window.
 - Fix: [#12910] Plugin API: getRide sometimes returns null for valid ride IDs.
 - Fix: [#12912] Plugin: selectedCell of CustomListView is being ignored on creation.
-- Fix: [#12918] Cannot place "Blue Hurricane" (hypercoaster).
+- Fix: [#12918] Cannot place vanilla TD6 tracks of the Hypercoasters, Monster Trucks, Classic Mini Roller Coasters, Spinning Wild Mouses and Hyper-Twisters types.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
 - Improved: [#12806] Add Esperanto diacritics to the sprite font.
 - Improved: [#12837] Arabic text is now drawn and shaped correctly on Windows.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -24,7 +24,7 @@
 - Fix: [#12881] Guests' favourite rides are not listed in the guest window.
 - Fix: [#12910] Plugin API: getRide sometimes returns null for valid ride IDs.
 - Fix: [#12912] Plugin: selectedCell of CustomListView is being ignored on creation.
-- Fix: [#12918] Cannot place vanilla TD6 tracks of the Hypercoasters, Monster Trucks, Classic Mini Roller Coasters, Spinning Wild Mouses and Hyper-Twisters types.
+- Fix: [#12918] Cannot place vanilla TD6 tracks of the Hypercoaster, Monster Trucks, Classic Mini Roller Coaster, Spinning Wild Mouse and Hyper-Twister types.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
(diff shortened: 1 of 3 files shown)


src/openrct2/rct2/T6Importer.cpp
```diff
@@ -200,8 +206,30 @@ class TD6Importer final : public ITrackImporter
...
+        UpdateRideType(td);
+
         return td;
     }
+
+    void UpdateRideType(std::unique_ptr<TrackDesign>& td)
+    {
+        if (RCT2RideTypeNeedsConversion(td->type))
+        {
+            std::scoped_lock<std::mutex> lock(_objectLookupMutex);
```
ZehMatt, line 219: Is this function actually used from multiple threads? The only things that use multiple threads is the index builder and the drawing code, this might not be needed here.

ZehMatt: Eh nevermind, I absolutely ignored I added this since this is in the file scanner.

distribution/changelog.txt
```diff
@@ -24,6 +24,7 @@
 - Fix: [#12881] Guests' favourite rides are not listed in the guest window.
 - Fix: [#12910] Plugin API: getRide sometimes returns null for valid ride IDs.
 - Fix: [#12912] Plugin: selectedCell of CustomListView is being ignored on creation.
+- Fix: [#12918] Cannot place "Blue Hurricane" (hypercoaster).
```
AaronVanGeffen, line 27: I realise this is a the title of the original issue, but I think it would be better if all of the "Hypercoasters, Monster Trucks, Classic Mini Roller Coasters, Spinning Wild Mouses and Hyper-Twisters" track types would be listed. As I understand it:
```suggestion
- Fix: [#12918] Cannot place vanilla TD6 tracks of the Hypercoasters, Monster Trucks, Classic Mini Roller Coasters, Spinning Wild Mouses and Hyper-Twisters types.
```

distribution/changelog.txt
```diff
@@ -24,7 +24,7 @@
 - Fix: [#12881] Guests' favourite rides are not listed in the guest window.
 - Fix: [#12910] Plugin API: getRide sometimes returns null for valid ride IDs.
 - Fix: [#12912] Plugin: selectedCell of CustomListView is being ignored on creation.
-- Fix: [#12918] Cannot place "Blue Hurricane" (hypercoaster).
+- Fix: [#12918] Cannot place vanilla TD6 tracks of the Hypercoasters, Monster Trucks, Classic Mini Roller Coasters, Spinning Wild Mouses and Hyper-Twisters types.
```
James103, line 27: When listing out track types, each track type should be listed as singular and not plural.
`- Fix: [#12918] Cannot place vanilla TD6 tracks of the Hypercoaster, Monster Truck, Classic Mini Roller Coaster, Spinning Wild Mouse and Hyper-Twister types.`

src/openrct2/rct2/T6Importer.cpp
```diff
@@ -200,8 +206,30 @@ class TD6Importer final : public ITrackImporter
         }
 
         td->name = _name;
+
+        UpdateRideType(td);
+
         return td;
     }
+
+    void UpdateRideType(std::unique_ptr<TrackDesign>& td)
```
duncanspumpkin, line 215: Why are you passing in the unique ptr just pass in a track design reference. This function isn't messing about with any memory allocation.

Gymnasiast: I was asked to do so. See the team chat.

PR #12999, opened 2020-09-26
Add .sea support to New Scenario list and extension handling

Gymnasiast:


src/openrct2/scenario/ScenarioRepository.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include <algorithm>
 #include <memory>
+#include <src/openrct2/core/MemoryStream.h>
```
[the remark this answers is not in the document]
Gymnasiast: It should, indeed. Unfortunately, this is how CLion auto-adds it and I didn't catch it myself.

src/openrct2/scenario/ScenarioRepository.cpp
```diff
@@ -203,6 +204,19 @@ class ScenarioFileIndex final : public FileIndex<scenario_index_entry>
     }
 
 private:
+    static IStream* GetStreamFromRCT2Scenario(const std::string& path)
```
tupaschoal, line 207: Can you make it return a `std::unique_ptr` instead? Then we don't need to do manual memory management.

Gymnasiast: It's what I tried initially, but that caused a crash. The same happened with shared_ptr.

Now, I'm not very knowledgeable in this area, so feel free to experiment with it yourself. If you happen to hit upon a better solution, you can push it to the PR.

IntelOrca: I am surprised it doesn't still crash. The problem is that your MemoryStream is still trying to access `data` which is automatically freed when `data` goes out of scope. You can tell MemoryStream that it is now the owner of the memory you pass into it, but I don't know how to take ownership away from `data`.

ZehMatt: The issue with this is the vector will not allow you to take ownership of the buffer. I simply copy the data into the MemoryStream now which solves this.

ZehMatt (approved):

PR #13000, opened 2020-09-26
Allow opening Objective options via console

Gymnasiast:
Ah, PR #13000. _Nice, round number, eh, Biggus?_


tupaschoal (approved):

duncanspumpkin:
Is this being added due to a request? Is there an associated issue this should be linked to? Should this be in the changelog?

Gymnasiast:
> Is this being added due to a request? 

No. I needed it when investigating something and thought it would be generally useful.

> Is there an associated issue this should be linked to?

No.

> Should this be in the changelog?

Yes.

PR #13012, opened 2020-09-27
Define NDEBUG for Xcode release targets

AaronVanGeffen:
Xcode does not define NDEBUG automatically for release targets. This means our current release builds ship with assertions.

This PR defines NDEBUG for Xcode release targets on a project basis. To make the project more maintainable, I have removed the `GCC_PREPROCESSOR_DEFINITIONS` clause from each of the individual targets, making them all inherit from the main project.

(Interestingly, DEBUG has to be declared manually as well, but we had taken care of that already.)


AaronVanGeffen:
Yep, this has the desired effect. Using a CI build:

Screenshot 2020-09-27 at 16 19 09
Screenshot 2020-09-27 at 16 19 09

Gymnasiast (approved):

PR #13020, opened 2020-09-28
Fix #8957: Error title missing when building w/ insufficient funds

### LINKED ISSUE #8957: Can't place stalls "Can't construct this here...(Undefined String)"
**OS:** [Windows 10]
**Version:** [0.2.2]
**Commit/Build:** [9cbc308]

Playing Lightning Peaks and get an error trying to place stalls in valid tiles and positions:
"Can't construct this here...(Undefined String)"

Was able to toggle disable clearance checks a few times and build stall successfully

**Steps to reproduce:**
1. Play Lightning Peaks, no cheats enabled
2. Try to place stall
3. Error pops up

**Screenshots / Video:**

OpenRCT2_Bug

**Save game:**

[Lightning Peaks_Bug.txt](https://github.com/OpenRCT2/OpenRCT2/files/3000792/Lightning.Peaks_Bug.txt)

tupaschoal:


tupaschoal:
When merging, please fix commit message "tile" -> "title"

src/openrct2/actions/GameAction.cpp
```diff
@@ -306,6 +306,7 @@ namespace GameActions
             if (!finance_check_affordability(result->Cost, action->GetFlags()))
             {
                 result->Error = GA_ERROR::INSUFFICIENT_FUNDS;
+                result->ErrorTitle = STR_RIDE_CONSTRUCTION_CANT_CONSTRUCT_THIS_HERE;
```
duncanspumpkin, line 309: This is not correct. The error title should always come from the game action.

tupaschoal: Note that the action actually returns `GA_ERROR::OK`, so it wouldn't have any of those anyway.

I think this is just a shortcut to check for funds in a single place, other than in each and every action. Seems like refactoring this out of here is a matter for another PR, don't you agree?

ZehMatt: Seems correct to me as all game actions have this checked before going into the actual game action, can you elaborate @duncanspumpkin

duncanspumpkin: This will mean that all actions that have insufficient funds will have an error title that is ride construction can't construct here

ZehMatt: I guess thats true but also on the other hand it would duplicate a lot of code if all game actions were to do this. I'm not sure what the optimal way is. I can only think of checking what the game action type is and based on that use a different title maybe?

duncanspumpkin: But isn't the real issue that one of the game actions is setting the title incorrectly?

ZehMatt: When the funds are checked the game action had no Query/Execute call yet, so no game action really assigns a title for the insufficient funds check. Maybe a generic title for this is the best way?

tupaschoal: Maybe `STR_1161    :Can't position this here...` ?

ZehMatt: That wouldn't fit when you try to remove something that costs money.

tupaschoal: Then I guess we're left with `STR_5637    :Can't do this...`

ZehMatt: Sounds okay, what do you think @duncanspumpkin?

duncanspumpkin: I still think there is something better we could do but that will be sufficient for now

tupaschoal: Ok

ZehMatt (approved):

Issue #13024, opened 2020-09-28
Console cursor doesn't follow actual cursor position


gsckoco:
**OS:** Ubuntu 20.04 and Windows 10
**Version:** 0.3.0 and 0.3.1
**Commit/Build:** Not sure.

Visual cursor doesn't follow cursor's actual position.

As far as I can tell, it only occurs in the console window, other text inputs such as the Park Name window has a working cursor.
**Steps to reproduce:**
1. Open debug console
2. Input a string, and move back through the string with arrow keys

**Screenshots / Video:**
https://youtu.be/AKAL3_Kj4YE

gsckoco:
I'm going to have a shot at fixing this myself

tupaschoal:
Closed via https://github.com/OpenRCT2/OpenRCT2/commit/2524e9201ac12cc3183667e53a2cf7d86667d52f

PR #13025, opened 2020-09-28
Fixed console cursor rendering.

gsckoco:
Based on my issue #13024 
Cursor now renders correctly at your current cursor position.

Here is the fix in action:
https://youtu.be/maBymn7jWPs


src/openrct2-ui/interface/InGameConsole.cpp
```diff
@@ -134,6 +134,12 @@ void InGameConsole::ScrollToEnd()
         _consoleScrollPos = std::max<int32_t>(0, static_cast<int32_t>(_consoleLines.size()) - maxLines);
 }
 
+void InGameConsole::RefreshCaret(size_t position)
+{
+    _consoleCaretTicks = 0;
+    _selectionStart = position;
+}
+
 void InGameConsole::RefreshCaret()
```
ZehMatt, line 143: You can eliminate this function by making the position parameter defaulting to 0.

ZehMatt:
I left a comment otherwise looking good. Also given that this is your first contribution you should add yourself to ```contributors.md``` also please add a changelog entry in ```changelog.txt```, this bug has been around since a while.

src/openrct2-ui/interface/InGameConsole.cpp
gsckoco: I did not know you could do this! I have got rid of the overloaded functions now.

gsckoco:
I have another open pull request with my name being added to the contributors.md, should I also add myself to the contributors on this PR, or wait for the other one to be merged?

ZehMatt:
> I have another open pull request with my name being added to the contributors.md, should I also add myself to the contributors on this PR, or wait for the other one to be merged?

You should only do this with your first PR, I wasn't aware you have another one open where you already did this, so no it is not required.

src/openrct2-ui/interface/InGameConsole.cpp
```diff
@@ -134,9 +134,10 @@ void InGameConsole::ScrollToEnd()
         _consoleScrollPos = std::max<int32_t>(0, static_cast<int32_t>(_consoleLines.size()) - maxLines);
 }
 
-void InGameConsole::RefreshCaret()
+void InGameConsole::RefreshCaret(size_t position = 0)
```
ZehMatt, line 137: The default parameter must be used only on the declaration, this will fail to compile. Consider building it locally before pushing your changes.

src/openrct2-ui/interface/InGameConsole.h
```diff
@@ -53,7 +54,7 @@ namespace OpenRCT2::Ui
         void WriteLine(const std::string& s, uint32_t colourFormat) override;
 
         void Input(CONSOLE_INPUT input);
-        void RefreshCaret();
+        void RefreshCaret(size_t position = 0); // Update _selectionStart to draw cursor in correct place
```
ZehMatt, line 57: The comment is not needed, the function is self explanatory.

distribution/changelog.txt
```diff
@@ -2,6 +2,8 @@
 ------------------------------------------------------------------------
 - Feature: [#13000] objective_options command for console
 - Fix: [#3200] Close Construction window upon selecting vehicle page
+- Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
```
AaronVanGeffen, line 5: The changelog is ordered by issue/PR number, grouped by category in the following order:

- Feature
- Change
- Fix
- Improved
- Technical

duncanspumpkin:
What is the performance of this like, get string width is a little expensive. I remember this being removed before. I'm sure there was some problem with how we did it before. @janisozaur i have a faint memory of discussing it with you.

gsckoco:
> What is the performance of this like, get string width is a little expensive. I remember this being removed before. I'm sure there was some problem with how we did it before. @janisozaur i have a faint memory of discussing it with you.

It shouldn't be any more expensive than any other text box in the game. However, I could always recalculate the string width in the `RefreshCaret()` method, this means it would only update when the player inputs a character or moves the cursor.

gsckoco:
I have moved the string width calculation from the `InGameConsole::Draw()` method to the `InGameConsole::RefreshCaret()` method. So now its not calculating the width of the string every frame.

ZehMatt (approved):
LGTM

distribution/changelog.txt
```diff
@@ -6,6 +6,7 @@
 - Fix: [#8015] RCT2 files are not found when put into the OpenRCT2 folder.
 - Fix: [#8957] Error title missing when building with insufficient funds
 - Fix: [#13021] Mowed grass and weeds don't show up in extra zoom levels.
+- Fix: [#13024] Console cursor now correctly renders at current cursor position.
```
tupaschoal, line 9:
```suggestion
- Fix: [#13024] Console cursor does not correctly render at current cursor position.
```

We typically state what was broken not what is now working, just a minor change

tupaschoal (approved):

tupaschoal:
Argh, forgot to include the "Close #..." as soon as I clicked squash :|

PR #13030, opened 2020-09-28
Fix #4617: Crash when rotating with invalid viewport open

### LINKED ISSUE #4617: Signs Causing Crashes

ZehMatt:
This fixes the issue described here: https://github.com/OpenRCT2/OpenRCT2/pull/11094#issuecomment-606820960 when clicking on those banners:
openrct2_2020-09-28_23-39-56
and then rotating.

The main issue is that the viewport is actually not working correctly with those, this just prevents the crash but the viewport issue remains.

### DIFF
--- a/src/openrct2-ui/windows/Sign.cpp
+++ b/src/openrct2-ui/windows/Sign.cpp
@@ -322,10 +322,7 @@ static void window_sign_paint(rct_window* w, rct_drawpixelinfo* dpi)
  */
 static void window_sign_viewport_rotate(rct_window* w)
 {
-    rct_viewport* view = w->viewport;
-    w->viewport = nullptr;
-
-    view->width = 0;
+    w->RemoveViewport();
 
     auto banner = GetBanner(w->number);
 
--- a/src/openrct2/interface/Window_internal.cpp
+++ b/src/openrct2/interface/Window_internal.cpp
@@ -42,3 +42,12 @@ void rct_window::Invalidate()
 {
     gfx_set_dirty_blocks({ windowPos, windowPos + ScreenCoordsXY{ width, height } });
 }
+
+void rct_window::RemoveViewport()
+{
+    if (viewport == nullptr)
+        return;
+
+    viewport->width = 0;
+    viewport = nullptr;
+}
--- a/src/openrct2/interface/Window_internal.h
+++ b/src/openrct2/interface/Window_internal.h
@@ -94,6 +94,7 @@ struct rct_window
     void SetLocation(const CoordsXYZ& coords);
     void ScrollToViewport();
     void Invalidate();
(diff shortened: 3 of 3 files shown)


src/openrct2-ui/windows/Sign.cpp
```diff
@@ -323,6 +323,9 @@ static void window_sign_paint(rct_window* w, rct_drawpixelinfo* dpi)
 static void window_sign_viewport_rotate(rct_window* w)
 {
     rct_viewport* view = w->viewport;
+    if (view == nullptr)
+        return;
+
     w->viewport = nullptr;
 
     view->width = 0;
```
Broxzier, line 331: I assume it crashed here. Why is this needed? Can't this just be removed, or moved down to the other null check?

IntelOrca: Yeah, I think this is the correct fix:
```c
if (w->viewport != nullptr)
{
    w->viewport->width = 0;
    w->viewport = nullptr;
}
```

ZehMatt: I applied @IntelOrca suggestion.

Gymnasiast (approved):
I can no longer get it to crash by rotating any viewport.

PR #13034, opened 2020-09-29
Remove common format formatter where possible

duncanspumpkin:
Eventually should have it limited to tooltips, window titles and widget args

This again does not fully cover everything but this is a nice chunk of them. Where I noticed that a formatter was not being used I've added on in as well.

### DIFF
--- a/src/openrct2-ui/windows/NewRide.cpp
+++ b/src/openrct2-ui/windows/NewRide.cpp
@@ -896,10 +896,10 @@ static void window_new_ride_paint_ride_information(
 
     // Ride name and description
     rideNaming = get_ride_naming(item.Type, rideEntry);
-    auto ft = Formatter::Common();
+    auto ft = Formatter();
     ft.Add<rct_string_id>(rideNaming.Name);
     ft.Add<rct_string_id>(rideNaming.Description);
-    gfx_draw_string_left_wrapped(dpi, gCommonFormatArgs, screenPos, width, STR_NEW_RIDE_NAME_AND_DESCRIPTION, COLOUR_BLACK);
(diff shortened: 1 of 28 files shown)


duncanspumpkin:
This is ready to go. Will not be adding more to this.

Gymnasiast (commented):
Code looks OK, will be doing a test as well.

Gymnasiast (approved):
Tested it. Ride window and Options window (two of the more complex windows) appear to work correctly.

Gymnasiast:
I took the liberty of rebasing, taking out a typo and squashing the last two commits (which both removed misused gCommonFormatArgs).

Can be merged as soon as the CIs are green.

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -896,26 +896,28 @@ static void window_new_ride_paint_ride_information(
...
         ft.Add<const char*>(drawString);
         DrawTextEllipsised(dpi, screenPos + ScreenCoordsXY{ 0, 39 }, WW - 2, STR_AVAILABLE_VEHICLES, ft, COLOUR_BLACK);
     }
 
+    ft = Formatter();
     if (item.Type != _lastTrackDesignCountRideType.Type || item.EntryIndex != _lastTrackDesignCountRideType.EntryIndex)
     {
         _lastTrackDesignCountRideType = item;
         _lastTrackDesignCount = get_num_track_designs(item);
+        ft.Add<int32_t>(_lastTrackDesignCount);
```
duncanspumpkin, line 920: Move to outside of if

Issue #12793, opened 2020-08-28
Add Excitement/Intensity/Nausea ratings to Rides view with sorting


kenperkins:
It would be really nice when using the Rides view that you could select Excitement, Intensity, or Nausea ratings as the view, with sorting for them.

![image](https://user-images.githubusercontent.com/730664/91508588-c831e500-e88c-11ea-8a6c-16a83bbc17d1.png)

IntelOrca:
I had a branch awhile ago that turned this into a multiple column table with excitement, intensity and nausea all sortable.

AaronVanGeffen:
I think @IntelOrca is referring to #9708

hjort96:
Is this something I could implement when #15836 is approved? Or are you set on using the new window @IntelOrca has mentioned?

Piperapk:
Had a go at it here #16655 as I saw that the window refactor of class RideList was done.

PR #13036, opened 2020-09-29
Feature #12793: Added Exciting/Intensity/Nausea sorts on Ride List

gsckoco:
Based on #12793 
image
image

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3651,6 +3651,7 @@ STR_6392    :Nausea rating: Not yet available
 STR_6393    :Excitement
 STR_6394    :Intensity
 STR_6395    :Nausea
+
 #############
 # Scenarios #
 ################
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3642,7 +3642,15 @@ STR_6383    :Open download page
 STR_6384    :Snow
 STR_6385    :Heavy Snow
 STR_6386    :Blizzard
-
+STR_6387    :Excitement rating: {COMMA2DP32}
+STR_6388    :Excitement rating: Not yet available
+STR_6389    :Intensity rating: {COMMA2DP32}
+STR_6390    :Intensity rating: Not yet available
+STR_6391    :Nausea rating: {COMMA2DP32}
+STR_6392    :Nausea rating: Not yet available
+STR_6393    :Excitement
+STR_6394    :Intensity
+STR_6395    :Nausea
 #############
 # Scenarios #
 ################
--- a/src/openrct2-ui/windows/RideList.cpp
+++ b/src/openrct2-ui/windows/RideList.cpp
@@ -117,6 +117,9 @@ enum {
     INFORMATION_TYPE_RELIABILITY,
     INFORMATION_TYPE_DOWN_TIME,
     INFORMATION_TYPE_GUESTS_FAVOURITE,
+    INFORMATION_TYPE_EXCITEMENT,
+    INFORMATION_TYPE_INTENSITY,
+    INFORMATION_TYPE_NAUSEA,
     DROPDOWN_LIST_COUNT
 };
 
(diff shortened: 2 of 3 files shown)


data/language/en-GB.txt
```diff
@@ -3642,7 +3642,15 @@ STR_6383    :Open download page
...
-
+STR_6387    :Excitement rating: {COMMA2DP32}
+STR_6388    :Excitement rating: Not yet available
+STR_6389    :Intensity rating: {COMMA2DP32}
+STR_6390    :Intensity rating: Not yet available
+STR_6391    :Nausea rating: {COMMA2DP32}
+STR_6392    :Nausea rating: Not yet available
+STR_6393    :Excitement
+STR_6394    :Intensity
+STR_6395    :Nausea
```
Broxzier, line 3653: Please keep the empty line after this block.

AaronVanGeffen:
Hmm, this seems to be partially duplicating existing strings:

```
STR_1473    :{WINDOW_COLOUR_2}Excitement rating: {BLACK}{COMMA2DP32}  ({STRINGID})
STR_1474    :{WINDOW_COLOUR_2}Excitement rating: {BLACK}Not yet available
STR_1475    :{WINDOW_COLOUR_2}Intensity rating: {BLACK}{COMMA2DP32}  ({STRINGID})
STR_1476    :{WINDOW_COLOUR_2}Intensity rating: {BLACK}Not yet available
STR_1477    :{WINDOW_COLOUR_2}Intensity rating: {OUTLINE}{RED}{COMMA2DP32}  ({STRINGID})
STR_1478    :{WINDOW_COLOUR_2}Nausea rating: {BLACK}{COMMA2DP32}  ({STRINGID})
STR_1479    :{WINDOW_COLOUR_2}Nausea rating: {BLACK}Not yet available
```

I wonder if it would be worth making the existing ones more flexible instead, e.g. by refactoring out the colours.

gsckoco:
> Hmm, this seems to be partially duplicating existing strings:
> ...
> I wonder if it would be worth making the existing ones more flexible instead, e.g. by refactoring out the colours.
Or would it be better to write a function that strips a string of its colour values?

AaronVanGeffen:
That sounds like allocating strings just for rendering? Unfortunately, the UI is redrawn every tick, so that may have a bit of an impact. In the current framework, it's easier to draw a string with e.g. two substrings. In this case, we'd use one string for the formatting, and have it process two strings by their string ids from passed format arguments.

gsckoco:
> That sounds like allocating strings just for rendering? Unfortunately, the UI is redrawn every tick, so that may have a bit of an impact. In the current framework, it's easier to draw a string with e.g. two substrings. In this case, we'd use one string for the formatting, and have it process two strings by their string ids from passed format arguments.

Good point, instead of splitting the strings into format strings and text strings. Would it be possible to instead pass an optional bool into whatever function handles setting the colour, so instead of having 2 strings for rendering just have 1 string and have it ignore rendering the colour.

AaronVanGeffen:
I see three colour variations and three labels. Something like:

1. `{WINDOW_COLOUR_2}{STRINGID}: {BLACK}{STRINGID}`
2. `{WINDOW_COLOUR_2}{STRINGID}: {OUTLINE}{RED}{STRINGID}`
3. `{STRINGID}: {STRINGID}` (introduced by this PR)

The first string id is the label, while the latter is either of:
* `{COMMA2DP32}  ({STRINGID})` (with the string id here being another label for the intensity)
* `{COMMA2DP32}` (variant introduced by this PR, but probably already exists)
* `Not yet available`

Perhaps I'm forgetting something, but it looks doable to me :)

duncanspumpkin:
Do we not have the ability to pass the colour in as an argument?

AaronVanGeffen:
We do, but (variable) window colours are not supported in this way. That, and there are two colours at play here.

gsckoco:
Yeah, the ability is there, but if there is a tag that overwrites it it becomes useless. 
image
This is using `STR_EXCITEMENT_RATING` and the colour is being set by the `{WINDOW_COLOUR_2}` tag
This is what it should look like
image

tupaschoal:
So where do we go from here? Is this on hold @AaronVanGeffen ?

gsckoco:
Yeah, I think this is on hold until the template strings are sorted

Gymnasiast:
@gsckoco When do you plan to make the changes Aaron requested?

gsckoco:
@Gymnasiast not sure, I have been busy with college and not had much time to work on any projects,

Gymnasiast:
Closing this as it has not been updated in two months. Let us know when/if you want to continue work and we can reopen it or open a new one.

IntelOrca:
Probably best left to first refactoring the window to a class then re-visiting #9708

PR #13042, opened 2020-09-30
Updated Readme.md

Piyussshh:
Updated Readme for better understanding
1."scenario" to "scenarios"
2."which also " to "that also"
3."and" to ", and"


Gymnasiast:
1. Seems ungrammatical to me.
2. Is not better than what it replaces.
3. Matter of taste.

PR #13043, opened 2020-09-30
Fix #8015: RCT2 files are not found when put into the OpenRCT2 folder

### LINKED ISSUE #8015: Game files are not found when put into the OpenRCT2 folder
**OS:** Windows 7
**Version:** 0.2.1

The readme notes that game files will be searched for in the loacation of openrct2.exe.  This is implemented in `Config::FindRCT2Path` by looking in `gExePath`, but after dcb92a466dbc1684d2dc6d827b894fa83d2d693b `gExePath` is no longer initialized.

**Steps to reproduce:**
1. Copy RTC2 game files to the folder where openrct2.exe is located.
2. Observe that openrct2.exe does not find the game files.

Another thing to note is that the RCT2 game files have a folder name `Data`, while OpenRCT2 has a folder named `data`.  It looks like on Linux this becomes two separate folders, while on Windows which uses a case insensitive filesystem it's one folder.  Not sure what to make of that, it trips my OCD.

Gymnasiast:
As mentioned in #8015, gExePath is never set, so I removed that variable altogether.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,6 +3,7 @@
 - Feature: [#12999] .sea (RCT Classic) scenarios are now listed in the “New Scenario” dialog.
 - Feature: [#13000] objective_options command for console.
 - Fix: [#3200] Close Construction window upon selecting vehicle page.
+- Fix: [#8015] RCT2 files are not found when put into the OpenRCT2 folder.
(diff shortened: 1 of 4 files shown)


tupaschoal (approved):
Nice find! And it even removes a global

PR #13045, opened 2020-09-30
Fix #13044: Rides in RCT1 saves all have "0 customers per hour"

### LINKED ISSUE #13044: Rides in RCT1 saves all have "0 customers per hour"
**OS:** Ubuntu 20.04
**Version:** 0.3.1+
**Commit/Build:** https://github.com/OpenRCT2/OpenRCT2/commit/b92328ce343688dbddfb8f55682133fb3744194f

Gymnasiast:
This issue was caused by misidentified fields. It didn't help that the TID was also incomplete (it only mentioned the first field). Luckily, RCT2 has the same data at the same place.

(Values cross-checked in RCT1)

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -6,6 +6,7 @@
 - Fix: [#8015] RCT2 files are not found when put into the OpenRCT2 folder.
 - Fix: [#13021] Mowed grass and weeds don't show up in extra zoom levels.
 - Fix: [#13029] Not all Junior Roller Coaster pieces are shown when "Show all track pieces" cheat is enabled.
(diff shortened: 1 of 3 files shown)


IntelOrca (approved):

duncanspumpkin:
So the age was just completely wrong?

IntelOrca:
I don't think so, `build_date` is what is used for ride age. I don't think the age field in the RCT1 struct was used.

Gymnasiast:
RCT1 has `build_date` in exactly the same place as RCT2 has, the `age` field was just misidentified (and never referenced).

PR #13049, opened 2020-09-30
Fix #13046: Simplify block sectioned comparisons

### LINKED ISSUE #13046: Simplify block sectioned comparisons
Lots of places in the code use the following comparison:
https://github.com/OpenRCT2/OpenRCT2/blob/47c72eaed3d9a14303dee5baa78f5772b63cd215/src/openrct2-ui/windows/Ride.cpp#L3826

But there is a simpler version which is a lot easier to read:
```c++
ride->IsBlockSectioned()
```

Please replace all the places in the code using the former for the latter

SGunner2014:


tupaschoal (commented):
You're on the right path, but I think you've missed some. From a quick count, I'd say there are:
- 5 candidates on  ` src/openrct2-ui/windows/Ride.cpp`
- 7 on `src/openrct2/ride/Ride.cpp`
- 3 on `src/openrct2/ride/Vehicle.cpp`

src/openrct2/ride/Ride.cpp
```diff
@@ -4936,7 +4936,7 @@ static bool ride_create_cable_lift(ride_id_t rideIndex, bool isApplying)
     if (ride == nullptr)
         return false;
 
-    if (ride->mode != RideMode::ContinuousCircuitBlockSectioned && ride->mode != RideMode::ContinuousCircuit)
+    if (!(ride->IsBlockSectioned()))
```
AaronVanGeffen: This statement contains superfluous parentheses. I suggest:
```suggestion
    if (!ride->IsBlockSectioned())
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -3431,8 +3431,7 @@ void Vehicle::FinishDeparting()
         audio_play_sound_at_location(SoundId::RideLaunch1, { x, y, z });
     }
 
-    if (curRide->mode != RideMode::Race && curRide->mode != RideMode::ContinuousCircuitBlockSectioned
-        && curRide->mode != RideMode::PoweredLaunchBlockSectioned)
+    if (curRide->mode != RideMode::Race && !(curRide->IsBlockSectioned()))
```
tupaschoal, line 3434:
```suggestion
    if (curRide->mode != RideMode::Race && !curRide->IsBlockSectioned())
```
Remove these parentheses as well.

One tip, when you fix the issue we open, make sure you set the conversation as resolved.

tupaschoal (commented):
This looks almost ready to go now, just need to remove some parentheses

src/openrct2/ride/Vehicle.cpp
SGunner2014: Okay, will do, thanks

tupaschoal (approved):

PR #13052, opened 2020-10-01
Close #12409: Refactor CONSOLE_INPUT to use strong enum

### LINKED ISSUE #12409: Refactor CONSOLE_INPUT to use strong enum

BryanDiLaura:
I didn't specify a type for the strong enum. Can add it if needed.


AaronVanGeffen (changes requested):
You also have to change the new enum class to conform to our code style, that is UpperCamelCase for class names.

BryanDiLaura:
D'oh. I also realized I didn't follow other pieces of the contributing guide. Fixing...

AaronVanGeffen:
Please re-use the same PR. Just push additional commits.

BryanDiLaura:
Sorry I'm a dingus 🤦‍♀️.  I think that should be in line with the style guide now. 

I notice that the style guide on the [wiki](https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style) doesn't mention classes. Is there someplace with a more complete style guide?

duncanspumpkin:
The style guide does mention classes. See the struct part.

AaronVanGeffen:
Looks much better now, thanks.

Per our contributing guidelines, we normally don't merge `develop` into branches. Instead, we would like PRs to be rebased when conflicts arise. That's not the case here, fortunately, so we can just squash-merge instead.

duncanspumpkin (approved):

PR #13063, opened 2020-10-01
Remove goto from DecideAndBuyItem to fix #12337

### LINKED ISSUE #12337: Remove goto from DecideAndBuyItem

rpstester:
I did it in two parts as you requested:
remove goto loc_69B119 by using an elseif statement
remove goto loc_69B221 by using a bool variable and some if statements to bypass the code

I am interested next in rewriting more of this function so that it flows better, as well as actually splitting it in two (decide and buy), but I don't want to get carried away to solve #12337.  

Please let me know if you have any questions/comments. Thanks!

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1580,7 +1580,7 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
         if (itemValue < price)
         {
             itemValue -= price;
-           
+
             if (!isRainingAndUmbrella)
             {
                 itemValue = -itemValue;
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1622,16 +1622,6 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
             HappinessTarget = std::min((HappinessTarget + happinessGrowth), PEEP_MAX_HAPPINESS);
             Happiness = std::min((Happiness + happinessGrowth), PEEP_MAX_HAPPINESS);
         }
-    }
-
-    if (!hasVoucher)
-    {
-        if (gClimateCurrent.Temperature >= 21)
-            itemValue = ShopItems[shopItem].HotValue;
-        else if (gClimateCurrent.Temperature <= 11)
-            itemValue = ShopItems[shopItem].ColdValue;
-        else
-            itemValue = ShopItems[shopItem].BaseValue;
 
         itemValue -= price;
         uint8_t satisfaction = 0;
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1546,9 +1546,11 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
     }
 
     if (shopItem == SHOP_ITEM_UMBRELLA && climate_is_raining())
-        goto loc_69B119;
-
-    if ((shopItem != SHOP_ITEM_MAP) && ShopItems[shopItem].IsSouvenir() && !hasVoucher)
+    {
+        //goto loc_69B119;  replaced by "else if" below
+        std::cout << "It's raining and I'm considering an umbrella\n";
+    }
+    else if ((shopItem != SHOP_ITEM_MAP) && ShopItems[shopItem].IsSouvenir() && !hasVoucher)
     {
         if (((scenario_rand() & 0x7F) + 0x73) > Happiness)
             return false;
@@ -1556,7 +1558,7 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
             return false;
     }
 
-loc_69B119:
+//loc_69B119: replaced by "else" above
     if (!hasVoucher)
     {
         if (price != 0 && !(gParkFlags & PARK_FLAGS_NO_MONEY))
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1548,9 +1548,7 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
 
     if (!isRainingAndUmbrella && (shopItem != SHOP_ITEM_MAP) && ShopItems[shopItem].IsSouvenir() && !hasVoucher)
     {
-        if (((scenario_rand() & 0x7F) + 0x73) > Happiness)
-            return false;
-        else if (GuestNumRides < 3)
+        if (((scenario_rand() & 0x7F) + 0x73) > Happiness || GuestNumRides < 3)
             return false;
     }
 
@@ -1585,11 +1583,11 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
             {
                 itemValue = -itemValue;
                 if (Happiness >= 128)
+                {
                     itemValue /= 2;
-
-                if (Happiness >= 180)
-                    itemValue /= 2;
-
+                    if (Happiness >= 180)
+                        itemValue /= 2;
+                }
                 if (itemValue > (static_cast<money16>(scenario_rand() & 0x07)))
                 {
                     // "I'm not paying that much for x"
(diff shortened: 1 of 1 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -1546,17 +1546,19 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
     }
 
     if (shopItem == SHOP_ITEM_UMBRELLA && climate_is_raining())
-        goto loc_69B119;
-
-    if ((shopItem != SHOP_ITEM_MAP) && ShopItems[shopItem].IsSouvenir() && !hasVoucher)
+    {
+        //goto loc_69B119;  replaced by "else if" below
+        std::cout << "It's raining and I'm considering an umbrella\n";
```
Gymnasiast, line 1553: Uhm, is this leftover test code?

Also, you can put the two if togethers:
```
auto consideringUmbrella = shopItem == SHOP_ITEM_UMBRELLA && climate_is_raining();
if (!consideringUmbrella && (shopItem != SHOP_ITEM_MAP) && ShopItems[shopItem].IsSouvenir() && !hasVoucher)
```

Also remove the `//loc_69B119: replaced by "else" above` comment.

src/openrct2/peep/Guest.cpp
```diff
@@ -1585,30 +1587,34 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
...
-                    goto loc_69B221;
+                //goto loc_69B221; now using below variable to skip all the code
+                //until the place where loc_69B221 was;
+                isRainingAndUmbrella = true;
             }
 
-            itemValue = -itemValue;
-            if (Happiness >= 128)
-                itemValue /= 2;
+            if (isRainingAndUmbrella == false)
```
tupaschoal, line 1597: It's a bit weird to compare a `bool` to a `bool` literal and it's not how we do it in the code base. Just do:
`!isRainingAndUmbrella`

src/openrct2/peep/Guest.cpp
```diff
@@ -1585,30 +1587,34 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
...
+                    PeepThoughtType thought_type = static_cast<PeepThoughtType>(
+                        (shopItem >= 32 ? (PEEP_THOUGHT_TYPE_PHOTO2_MUCH + (shopItem - 32))
+                                        : (PEEP_THOUGHT_TYPE_BALLOON_MUCH + shopItem)));
+                    InsertNewThought(thought_type, ride->id);
+                    return false;
+                }
             }
         }
-        else
+        else if (isRainingAndUmbrella == false)
```
tupaschoal, line 1617: This doesn't seem like it will ever be `false`, as the variable is only changed inside the scope above, `if`

src/openrct2/peep/Guest.cpp
```diff
@@ -1585,30 +1587,34 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
...
         {
             itemValue -= price;
-            if (shopItem == SHOP_ITEM_UMBRELLA)
+            if (shopItem == SHOP_ITEM_UMBRELLA && climate_is_raining())
             {
-                if (climate_is_raining())
-                    goto loc_69B221;
+                //goto loc_69B221; now using below variable to skip all the code
+                //until the place where loc_69B221 was;
+                isRainingAndUmbrella = true;
```
tupaschoal, line 1594: Not sure if this is is the best way, sounds like a break (or an even better refactor) would make the code clearer

src/openrct2/peep/Guest.cpp
rpstester: OK, I will clean it up.

src/openrct2/peep/Guest.cpp
rpstester: Yes, that makes sense!

src/openrct2/peep/Guest.cpp
rpstester: I will work on a refactor; it seems like it could be rewritten to be clearer.

rpstester:
What does it mean that these two tests failed, and what can I do about it?

tupaschoal:
It is a replay where we replay (duh) game actions and see whether the game state is the same as a file we use for comparison. When it fails, it means that the game did not behave exactly the same, which is ok if you're fixing a bug, but it's not if you're refactoring (like here), because that means some of the game logic changed.

rpstester:
OK, in that case I'll go over it again, as I probably made a mistake.  Thanks!

rpstester:
If you look at the last commit, I don't understand why I'm failing the tests. My small changes are all Boolean.  Can you help, please?

src/openrct2/peep/Guest.cpp
```diff
@@ -1583,30 +1585,32 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
...
+                    PeepThoughtType thought_type = static_cast<PeepThoughtType>(
+                        (shopItem >= 32 ? (PEEP_THOUGHT_TYPE_PHOTO2_MUCH + (shopItem - 32))
+                                        : (PEEP_THOUGHT_TYPE_BALLOON_MUCH + shopItem)));
+                    InsertNewThought(thought_type, ride->id);
+                    return false;
+                }
             }
         }
-        else
+        else if (!isRainingAndUmbrella)
```
duncanspumpkin, line 1613: drop this extra if and I think you should pass the tests.

rpstester: I missed it! Thanks so much!

rpstester:
Alright. I believe I have met the requirements of this issue. Please let me know when you get a chance. Thanks!

src/openrct2/peep/Guest.cpp
```diff
@@ -1627,17 +1620,6 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
...
-
-loc_69B221:
-    if (!hasVoucher)
-    {
-        if (gClimateCurrent.Temperature >= 21)
-            itemValue = ShopItems[shopItem].HotValue;
-        else if (gClimateCurrent.Temperature <= 11)
-            itemValue = ShopItems[shopItem].ColdValue;
-        else
-            itemValue = ShopItems[shopItem].BaseValue;
```
tupaschoal, line 1640: I'm a bit confused by the removal of these. Sounds like it was used to reset the `itemValue`, but it isn't anymore,

rpstester: In the old code, it did the same operation twice. I removed the redundancy.

tupaschoal: But look in the old code it first reset `itemValue` depending on the temperature, then it seems like it always subtracted the price, so I'll call `itemValue - price` as `itemValueP`. Afterwards, it could do any number of operations. For the first if:
`[-itemValueP, -itemValueP/2, -itemValueP/4]`
For the else:
`[max(8, itemValueP)]`
For the second if, the one that calculates satisfaction, it reset back to `itemValueP` , but now it is whatever the end result of the if/else pair above ends up with, which in most of the cases isn't it (only 1 out of the 5 options). Does that make sense?

rpstester: Yes, I see your point.  I have restored the code.

src/openrct2/peep/Guest.cpp
```diff
@@ -1583,31 +1578,36 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
...
         }
         else
         {
+            // reset itemValue for satisfaction calculation
+            if (gClimateCurrent.Temperature >= 21)
+                itemValue = ShopItems[shopItem].HotValue;
+            else if (gClimateCurrent.Temperature <= 11)
+                itemValue = ShopItems[shopItem].ColdValue;
+            else
+                itemValue = ShopItems[shopItem].BaseValue;
```
tupaschoal, line 1610: This is still misplaced, it should be after the happiness calculation, just before line 1631 when it starts the satisfaction calculation

rpstester: You're right. Sorry about that.

tupaschoal (approved):

PR #13066, opened 2020-10-02
Light Tile Inspector Improvements

tupaschoal:
Each commit is self contained and explanatory

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3642,6 +3642,9 @@ STR_6383    :Open download page
 STR_6384    :Snow
 STR_6385    :Heavy Snow
 STR_6386    :Blizzard
+STR_6387    :Can't lower element here...
+STR_6388    :Can't raise element here...
+STR_6389    :No clearance
 
 #############
 # Scenarios #
--- a/src/openrct2/world/TileInspector.cpp
+++ b/src/openrct2/world/TileInspector.cpp
@@ -474,6 +474,29 @@ GameActionResult::Ptr tile_inspector_sort_elements_at(const CoordsXY& loc, bool 
     return std::make_unique<GameActionResult>();
 }
 
+static GameActionResult::Ptr ValidateTileHeight(TileElement* const tileElement, int8_t heightOffset)
+{
+    int16_t newBaseHeight = static_cast<int16_t>(tileElement->base_height + heightOffset);
+    int16_t newClearanceHeight = static_cast<int16_t>(tileElement->clearance_height + heightOffset);
+    if (newBaseHeight < 0)
+    {
+        return std::make_unique<GameActionResult>(GA_ERROR::TOO_LOW, STR_CANT_LOWER_ELEMENT_HERE, STR_TOO_LOW);
+    }
+    else if (newBaseHeight > MAX_ELEMENT_HEIGHT)
+    {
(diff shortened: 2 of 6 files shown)


src/openrct2/world/TileInspector.cpp
```diff
@@ -474,20 +474,39 @@ GameActionResult::Ptr tile_inspector_sort_elements_at(const CoordsXY& loc, bool
...
+        return std::make_unique<GameActionResult>(GA_ERROR::TOO_LOW, STR_CANT_LOWER_ELEMENT_HERE, STR_TOO_LOW);
+    }
+    else if (newBaseHeight > MAX_ELEMENT_HEIGHT)
+    {
+        return std::make_unique<GameActionResult>(GA_ERROR::TOO_HIGH, STR_CANT_RAISE_ELEMENT_HERE, STR_TOO_HIGH);
+    }
+    else if (newClearanceHeight < 0)
+    {
+        return std::make_unique<GameActionResult>(GA_ERROR::NO_CLEARANCE, STR_CANT_LOWER_ELEMENT_HERE, STR_NO_CLEARANCE);
+    }
```
tupaschoal, line 486: I can't actually see why we would ever fall here, but I didn't want to change the previous behavior

Broxzier: This may be hit in case an element has a clearance height that is lower than its base height (who knows someone might have found a use of such a hack, I wouldn't be too surprised anymore).

tupaschoal: But would anything have a loww clearance*ok it's own" (without hacks)?

Broxzier: Without hacks that's not possible indeed, then it would error out on the base height check already.

ZehMatt: Is it perhaps possible with saved track designs?

tupaschoal: I mean, I'm handling it anyway, just wanted to know if it was really possible

ZehMatt (approved):
Looks good code wise, I did not test this.

tupaschoal:
I did plenty of tests, seems fine

data/language/en-GB.txt
```diff
@@ -3642,6 +3642,9 @@ STR_6383    :Open download page
 STR_6384    :Snow
 STR_6385    :Heavy Snow
 STR_6386    :Blizzard
+STR_6387    :Can't lower element here...
```
Gymnasiast, line 3645: Should have used an ellipsis here. (`…`)

data/language/en-GB.txt
```diff
@@ -3642,6 +3642,9 @@ STR_6383    :Open download page
...
+STR_6388    :Can't raise element here...
+STR_6389    :No clearance
```
Gymnasiast, line 3647: This is not correct. The message gets emitted when the clearance height ends up with a value < 0 or > MAX_HEIGHT, while "No clearance" would mean that the tile element is too high to fit between/under another (which the tile element very deliberately ignores!)

data/language/en-GB.txt
tupaschoal: To bem fair I copied from the "can't raise land here..." text, so we probably need to update it there

PR #13068, opened 2020-10-02
Fix #13055: Replace includes of Json.hpp in header files with json_fwd.hpp

### LINKED ISSUE #13055: Reduce Json.hpp includes
Since #12667, we've been using a new header-only C++ JSON library. While this leads to nicer code than the C library we were using before, we find compilation times for a clean build have tripled. This is unfortunate.

A quick investigation finds that Json.hpp is included in several header files as well, which in turn are included by loads of other files. This means that all of these will be parsing Json.hpp again in every compilation unit. We should try to remove the include for these files.

The following header files are affected:

* src/openrct2/network/ServerList.h
* src/openrct2/network/network.h
* src/openrct2/network/NetworkUser.h
* src/openrct2/network/NetworkBase.h
* src/openrct2/network/NetworkGroup.h
* src/openrct2/object/StringTable.h
* src/openrct2/object/RideObject.h
* src/openrct2/object/ImageTable.h

mwnciau:
JSON for Modern C++ provides a forward declaration header file since 3.1.0 that will forward declare the `nlohmann::json` type (note: this isn't trivial because it is a complicated template class). This PR replaces all includes of the `Json.hpp` file in header files with this forward declaration. This reduces the clean compile time from about 220s to 180s on my machine.

I tried a few other things, too: moving GetNumber into the cpp file and explicitly instantiating it for the types we use; removing the GetEnum template function in favour of using GetNumber and doing static casts; moving the GetFlags functions to a separate file so they don't always have to be included. The compilation time difference after making these changes was more or less the same, so I don't think it's worth the inconvenience.

IIRC, the docker images only pull `nlohmann/json.hpp` so they may need to be changed, too.


src/openrct2/object/LargeSceneryObject.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include <algorithm>
 #include <iterator>
+#include <nlohmann/json_fwd.hpp>
```
tupaschoal, line 25: Seems a bit suspicious to have the `fwd` on a source file, specially because there was no include before 🤔

tupaschoal (commented):
I  think we probably should have `core/Json_fwd.hpp`, so that we don't have to directly include `nlohmann` anywhere on the code, but the two `Json` files. We could also package the `using json_t = nlohmann::json` into this forward, which would make usage simpler too.

> I tried a few other things, too: moving GetNumber into the cpp file and explicitly instantiating it for the types we use; removing the GetEnum template function in favour of using GetNumber and doing static casts; moving the GetFlags functions to a separate file so they don't always have to be included. The compilation time difference after making these changes was more or less the same, so I don't think it's worth the inconvenience.

When you did that, did the header file still include the full `nlohmann/json`? Because I was thinking that if we were to split into `Json.h`, with the fwd declare and the function prototypes, and `Json.cpp` with the actual full include, that it might be faster. I just want to confirm it isn't exactly what you tried.

mwnciau:
> Because I was thinking that if we were to split into Json.h, with the fwd declare and the function prototypes, and Json.cpp with the actual full include, that it might be faster.

I gave that a go, too, but the forward declaration doesn't include any of the methods in json_t, just an empty forward declared object, so we'd still need to include the full thing wherever we used any of the methods.

In full, I tried (this was after changing to use json_fwd):

> moving GetNumber into the cpp file and explicitly instantiating it for the types we use

This prevents the compiler compiling the same instantiation of GetNumber multiple times, but it does mean you have to explicitly instantiate every type you would use - currently we have 7. There was no noticeable improvement so definitely not worth having to keep track of those 7 types.

```
// Json.hpp - forward declare without body
 template<typename T> T GetNumber(json_t, T);

// Json.cpp - define body and explicitly declare the types we use
template<typename T> T GetNumber(json_t, T) { ... };

template float GetNumber<float>(const json_t&, float);
...
```

> removing the GetEnum template function in favour of using GetNumber and doing static casts

This function is only actually used once so I just removed it and used GetNumber to the underlying type. No perceived improvement, not too surprisingly given how little it was used.

> moving the GetFlags functions to a separate file so they don't always have to be included

Created a new file, `Flags.hpp`, and moved these functions there so they only need to be included where needed. Again, no perceived improvement, not too surprisingly given its included where its used anyway.

----

This is slightly unrelated, but I discovered this when sorting out the includes. Currently NetworkGroup doesn't need a `JsonFwd.hpp` include because of this inheritence chain:

`NetworkGroup` from `NetworkPacket.h` from `NetworkTypes.h` from `RideTypes.h` (for `ride_id_t`) from `Object.h` (for `ObjectEntryIndex`)

Both the includes for `RideTypes.h` and `Object.h` are for a single type alias (`ride_id_t` and `ObjectEntryIndex`). Seems like it might be nice to have a single include file for the type aliases to reduce the include spaghetti a bit (although I don't know how this would look in the future of strong enums). That or just redefining `ride_id_t` and `ObjectEntryIndex` instead of including the whole thing, but I don't know how people feel about having duplicate definitions about the place.

tupaschoal (commented):
Alright, thanks for the explanation and investigation! I'm ok with waiting a bit on reducing the spaghetti code, because we still have to settle on the enum stuff.

This looks good to go in my opinion, we just need the docker changes merged so we can update here properly.

AaronVanGeffen:
Fwiw, https://github.com/OpenRCT2/openrct2-docker-build/pull/8 has been merged and images are building.

AaronVanGeffen:
Looks like the macOS dependencies bundle will need to be updated. I will take care of that tonight.

There are also CI errors for -I believe- the networkless build and Android build. Could you look into those?

src/openrct2/interface/Colour.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <unordered_map>
```
AaronVanGeffen, line 17: Added to fix macOS/Xcode compilation.

mwnciau:
Fixed the network build issue.

The android build issue is an issue with the docker image - OpenRCT2/openrct2-docker-build#9.

AaronVanGeffen (approved):

akien-mga:
Note that this requires having `nlohmann/json` built with the `JSON_MultipleHeaders` option, which is OFF by default, and thus not used explicitly by many distros.

See e.g. [Fedora](https://src.fedoraproject.org/rpms/json/blob/master/f/json.spec#_42) or [Mageia](https://svnweb.mageia.org/packages/cauldron/json/current/SPECS/json.spec?view=markup#l36).

On such a system, OpenRCT2 v0.3.2 therefore fails building:
```
[  0%] Building CXX object CMakeFiles/libopenrct2.dir/src/openrct2/Cheats.cpp.o
In file included from /home/akien/tmp/OpenRCT2/src/openrct2/actions/../management/../ride/../rct12/../object/Object.h:13,
                 from /home/akien/tmp/OpenRCT2/src/openrct2/actions/../management/../ride/../rct12/RCT12.h:15,
                 from /home/akien/tmp/OpenRCT2/src/openrct2/actions/../management/../ride/Ride.h:15,
                 from /home/akien/tmp/OpenRCT2/src/openrct2/actions/../management/Research.h:14,
                 from /home/akien/tmp/OpenRCT2/src/openrct2/actions/../management/Finance.h:13,
                 from /home/akien/tmp/OpenRCT2/src/openrct2/actions/ParkSetLoanAction.hpp:15,
                 from /home/akien/tmp/OpenRCT2/src/openrct2/Cheats.cpp:13:
/home/akien/tmp/OpenRCT2/src/openrct2/actions/../management/../ride/../rct12/../object/../core/JsonFwd.hpp:12:10: fatal error: nlohmann/json_fwd.hpp: No such file or directory
   12 | #include <nlohmann/json_fwd.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/libopenrct2.dir/build.make:82: CMakeFiles/libopenrct2.dir/src/openrct2/Cheats.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:209: CMakeFiles/libopenrct2.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
```

Mailaender:
Same problem on @openSUSE. It seems also older releases like  https://github.com/nlohmann/json/blob/v3.7.3/CMakeLists.txt#L20 come with this disabled by default. I am trying to address this in https://build.opensuse.org/request/show/856743

PR #13069, opened 2020-10-02
Close #13047: Track type normalising duplication in GameActions

### LINKED ISSUE #13047: Track type normalising duplication in Game Actions
Many game actions contain this code:
```
        switch (_trackType)
        {
            case TRACK_ELEM_BEGIN_STATION:
            case TRACK_ELEM_MIDDLE_STATION:
                comparableTrackType = TRACK_ELEM_END_STATION;
        }
```

This should be put into a function. "NormaliseTrackType", for example.

SGunner2014:


src/openrct2/actions/GameAction.h
```diff
@@ -318,6 +319,18 @@ struct GameAction
         return 0;
     }
 
+    int32_t NormaliseTrackType(int32_t _trackType) const
```
ZehMatt, line 322: This seems like the wrong place to put it in, this has hardly anything to do with game actions in general. It would be better suited with the TrackRemoveAction class.

ZehMatt (changes requested):

src/openrct2/actions/GameAction.h
IntelOrca: We also have `TrackElement::IsStation()` and `track_type_is_station` which could be used.
E.g.
```
if (track_type_is_station(_trackType)
    _trackType = TrackElemType::EndStation;
```

PR #13070, opened 2020-10-02
Close #12404: Refactor DRAWING_ENGINE to use strong enum

### LINKED ISSUE #12404: Refactor DRAWING_ENGINE to use strong enum

ju-pinheiro:


src/openrct2-ui/drawing/engines/DrawingEngineFactory.hpp
```diff
@@ -33,14 +33,14 @@ namespace OpenRCT2
             std::unique_ptr<IDrawingEngine> Create(
                 DRAWING_ENGINE_TYPE type, const std::shared_ptr<IUiContext>& uiContext) override
```
ZehMatt, line 34: The type of the parameter should be changed to DrawingEngine, I don't understand how this managed to compile now that the old type is gone.

ZehMatt: Just realized, this is another type declared IDrawingEngine.h, can you also remove that one and use the correct type?

ju-pinheiro: Yes, is another type! I found that strange too haha.
But since I'm new to the code I thought that changing it could affect the functionality.

So you suggest replacing DRAWING_ENGINE_TYPE with the new DrawingEngine enum class, right?
Sure, I'll enhance it.

ZehMatt: Yeah its an oddity, it should use the right type.

ZehMatt (approved):

PR #13073, opened 2020-10-02
Minor cleanup around audio code

ZehMatt:
This should make it a bit easier to move the audio code later in a class. I initially just wanted to get rid of that custom memory allocation but ended up cleaning a bit more, also removes 2 global variables and a struct that only really stored the name.


src/openrct2-ui/windows/Options.cpp
```diff
@@ -1164,15 +1164,16 @@ static void window_options_mousedown(rct_window* w, rct_widgetindex widgetIndex,
                     audio_populate_devices();
 
                     // populate the list with the sound devices
-                    for (size_t i = 0; static_cast<int32_t>(i) < gAudioDeviceCount; i++)
+                    for (int32_t i = 0; i < audio_get_device_count(); ++i)
                     {
+                        auto& deviceName = audio_get_device_name(i);
                         gDropdownItemsFormat[i] = STR_OPTIONS_DROPDOWN_ITEM;
-                        gDropdownItemsArgs[i] = reinterpret_cast<uintptr_t>(gAudioDevices[i].name);
+                        gDropdownItemsArgs[i] = reinterpret_cast<uintptr_t>(deviceName.c_str());
```
tupaschoal, line 1171: Is `gDropdownItemsArgs` an array of strings or of `char*´? Because if it's the latter, this would save the pointer of a soon to be unscoped variable (deviceName), so we'd have to `strdup`

tupaschoal (commented):
There's a sign comparison mismatch which is causing CI to fail

src/openrct2-ui/windows/Options.cpp
ZehMatt: It returns std::string as reference and the audio devices are populated at the very beginning so this pointer will be alive until the application closes, this should be fine and has been like that before. I know it aint pretty but its been like this before.

src/openrct2/audio/Audio.cpp
```diff
@@ -311,6 +316,22 @@ void audio_stop_all_music_and_sounds()
     audio_stop_weather_sound();
 }
 
+int32_t audio_get_device_count()
+{
+    return static_cast<int32_t>(_audioDevices.size());
```
tupaschoal, line 321: Do we need the cast here?

src/openrct2/audio/Audio.cpp
```diff
@@ -311,6 +316,22 @@ void audio_stop_all_music_and_sounds()
...
+}
+
+const std::string& audio_get_device_name(int32_t index)
+{
+    Guard::Assert(index >= 0 && index < audio_get_device_count());
+    return _audioDevices[index];
```
tupaschoal, line 327: This will still crash in a production build, because IIRC the default behavior for the guard assert is just to log

src/openrct2/audio/audio.h
```diff
@@ -170,6 +161,22 @@ extern rct_ride_music_params* gRideMusicParamsListEnd;
...
+ */
+int32_t audio_get_device_count();
+/**
+ * Returns the device name by index.
+ */
+const std::string& audio_get_device_name(int32_t index);
+/**
+ * Returns the currently used device index, -1 if not available.
+ */
+int32_t audio_get_device_index();
```
tupaschoal, line 179: Shouldn't we be taking this opportunity to use `TitleCase` in the new functions?

ZehMatt: I want to move all of those into a class, so better left for a new PR. Just wanted to make it easy to move.

src/openrct2/audio/Audio.cpp
ZehMatt: If I return size_t we would need quite a few casts else where.

ZehMatt:
Since this already stalled a bit I'll keep going on this PR. I'll move the functions into a namespace and convert the functions to PascalCase and address some minor potential issues.

src/openrct2/ride/Ride.cpp
```diff
@@ -3603,28 +3604,28 @@ int32_t ride_music_params_update(
...
-            rct_ride_music* ride_music = &gRideMusicList[0];
+            OpenRCT2::Audio::RideMusic* ride_music = &OpenRCT2::Audio::gRideMusicList[0];
             int32_t channel = 0;
             uint32_t a1;
             while (ride_music->ride_id != ride->id || ride_music->tune_id != *tuneId)
             {
                 ride_music++;
                 channel++;
-                if (channel >= AUDIO_MAX_RIDE_MUSIC)
+                if (channel >= OpenRCT2::Audio::MaxRideMusic)
```
tupaschoal, line 3614:
```suggestion
                if (static_cast<size_t>(channel) >= OpenRCT2::Audio::MaxRideMusic)
```

src/openrct2/ride/Ride.cpp
```diff
@@ -3690,10 +3691,10 @@ void ride_music_update_final()
                 }
                 rideMusic++;
                 channelIndex++;
-                if (channelIndex >= AUDIO_MAX_RIDE_MUSIC)
+                if (channelIndex >= OpenRCT2::Audio::MaxRideMusic)
```
tupaschoal, line 3694:
```suggestion
                if (static_cast<size_t>(channelIndex) >= OpenRCT2::Audio::MaxRideMusic)
```

tupaschoal (commented):
It looks alright, though because it's so big I might have missed or overlooked a thing or too.

I'll push the two fixes I suggested to try having a green CI overall. Maybe we should just have those globals as signed types, as they aren't big anyway. I didn't want to change because it might blow up somewhere else.

Other than that, lots of `type*` were converted to `auto*`. When it's a pointer, personally I'd use `auto`, but no big fuss

tupaschoal (approved):

ZehMatt:
> It looks alright, though because it's so big I might have missed or overlooked a thing or too.
> 
> I'll push the two fixes I suggested to try having a green CI overall. Maybe we should just have those globals as signed types, as they aren't big anyway. I didn't want to change because it might blow up somewhere else.
> 
> Other than that, lots of `type*` were converted to `auto*`. When it's a pointer, personally I'd use `auto`, but no big fuss

I'm not too big of a fan when you can not differentiate on auto if its pointer or reference, thats why I usually make it visible to the reader so there is no need having to lookup what the assignment is about.

ZehMatt:
Before merging it would be best if a third pair of eyes would have also a look.

tupaschoal:
> I'm not too big of a fan when you can not differentiate on auto if its pointer or reference, thats why I usually make it visible to the reader so there is no need having to lookup what the assignment is about.

The reference is never inferred, so it's always `auto&`, otherwise it's either a copy or a pointer.

duncanspumpkin (approved):

ZehMatt:
> > I'm not too big of a fan when you can not differentiate on auto if its pointer or reference, thats why I usually make it visible to the reader so there is no need having to lookup what the assignment is about.
> 
> The reference is never inferred, so it's always `auto&`, otherwise it's either a copy or a pointer.

The thing is about pointers
```
int var = 0;
auto a = &var;
auto* b = &var;
```
in both cases the type is ```int*``` where the first one hides the fact its a pointer. Thats why I mentioned the references.
```
int var = 0;
auto& a = var;
auto b = var;
```
Where ```a``` is clearly a reference and ```b``` a copy. Thats why I like to make it more strict, because a copy of a pointer is always a pointer type.

You might think "but doing &var is clearly a pointer", that is true. But if you have code like
```
auto a = function();
```
You would need to check what ```function``` returns, might be a pointer or might be a reference, I just find it easier to read and clarifying it for anyone else reading the code that the function does return a pointer vs copy.

PR #13074, opened 2020-10-02
Fix entrance and exit ghosts not being removed for mazes

mwnciau:
Currently, when placing the entrance or exit of a maze, the ghost of the entrance or exit will not disappear until you actually place it - even if you exit construction. This PR adds a call in the `ride_construction_invalidate_current_track` function to remove the ghost images for mazes when the track is invalidated.

This also solves a separate issue where `RideEntranceExitRemoveAction` would spew a lot of "Track Element not found" warnings in the console.

I couldn't see an existing issue about this. It's not an original bug, but it looks like it's always been in OpenRCT2.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -8,7 +8,7 @@
 - Fix: [#13021] Mowed grass and weeds don't show up in extra zoom levels.
 - Fix: [#13029] Not all Junior Roller Coaster pieces are shown when "Show all track pieces" cheat is enabled.
 - Fix: [#13044] Rides in RCT1 saves all have "0 customers per hour".
-- Fix: Entrance and exit ghosts for mazes not being removed.
+- Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 
 0.3.1 (2020-09-27)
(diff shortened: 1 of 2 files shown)


tupaschoal (commented):
This looks alright, just missing a changelog entry, something like:
Fix: [13074] ...

mwnciau:
Added a changelog. I didn't include #13074 because there is no issue for this bug - similar to the loop collision box issue I fixed in 0.3.1 where we decided not to include the PR number.

tupaschoal:
> Added a changelog. I didn't include #13074 because there is no issue for this bug - similar to the loop collision box issue I fixed in 0.3.1 where we decided not to include the PR number.

That was an oversight on our part then. We have decide to always have issue or PR # (if first doesn't exist), so one person reading the changelog can quickly go there and read about if, if interested.

distribution/changelog.txt
```diff
@@ -8,6 +8,7 @@
 - Fix: [#13021] Mowed grass and weeds don't show up in extra zoom levels.
 - Fix: [#13029] Not all Junior Roller Coaster pieces are shown when "Show all track pieces" cheat is enabled.
 - Fix: [#13044] Rides in RCT1 saves all have "0 customers per hour".
+- Fix: Entrance and exit ghosts for mazes not being removed.
```
tupaschoal, line 11:
```suggestion
- Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
```

PR #13077, opened 2020-10-03
Cleanup game actions

ZehMatt:


src/openrct2/actions/LandBuyRightsAction.hpp
```diff
@@ -132,11 +132,11 @@ DEFINE_GAME_ACTION(LandBuyRightsAction, GAME_COMMAND_BUY_LAND_RIGHTS, GameAction
         if (surfaceElement == nullptr)
         {
             log_error("Could not find surface. x = %d, y = %d", loc.x, loc.y);
-            return MakeResult(GA_ERROR::INVALID_PARAMETERS, _ErrorTitles[_setting], STR_NONE);
+            return MakeResult(GA_ERROR::INVALID_PARAMETERS, _ErrorTitles[EnumValue(_setting)], STR_NONE);
```
tupaschoal, line 135: I wonder if we shouldn't just make a function to convert from the setting to the error, seems cleaner (and less error prone).

Not something for this PR though

tupaschoal (approved):
Great cleanup!

PR #13079, opened 2020-10-03
Do not output error if the sourceGame parameter is omitted

Gymnasiast:
Since >90% of objects in the wild are custom objects, it make sense if this
parameter can be safely omitted. Only warn if this field contains garbage.

### DIFF
--- a/src/openrct2/object/ObjectFactory.cpp
+++ b/src/openrct2/object/ObjectFactory.cpp
@@ -408,6 +408,38 @@ namespace ObjectFactory
         return result;
     }
 
+    static void ExtractSourceGames(const std::string& id, json_t& jRoot, Object* result)
+    {
+        auto sourceGames = jRoot["sourceGame"];
+        if (sourceGames.is_array() || sourceGames.is_string())
+        {
+            std::vector<uint8_t> sourceGameVector;
+            for (const auto& jSourceGame : sourceGames)
+            {
+                sourceGameVector.push_back(ParseSourceGame(Json::GetString(jSourceGame)));
+            }
+            if (!sourceGameVector.empty())
+            {
+                result->SetSourceGames(sourceGameVector);
+            }
+            else
+            {
+                log_error("Object %s has an incorrect sourceGame parameter.", id.c_str());
+                result->SetSourceGames({ OBJECT_SOURCE_CUSTOM });
+            }
+        }
+        // >90% of objects are custom, so allow omitting the parameter without displaying an error.
+        else if (sourceGames.is_null())
+        {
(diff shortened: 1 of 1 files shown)


src/openrct2/object/ObjectFactory.cpp
```diff
@@ -408,6 +408,38 @@ namespace ObjectFactory
         return result;
     }
 
+    static void ExtractSourceGames(const std::string& id, const json_t& jRoot, Object* result)
```
tupaschoal, line 411:
```suggestion
    static void ExtractSourceGames(const std::string& id, json_t& jRoot, Object* result)
```
@mwnciau suggested that having `const &` on json objects could be dangerous when there might be exceptions involved. I guess you have to remove this here or check if the key `sourceGame` exists, before starting.

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -408,6 +408,38 @@ namespace ObjectFactory
...
+            {
+                result->SetSourceGames(sourceGameVector);
+            }
+            else
+            {
+                log_error("Object %s has an incorrect sourceGame parameter.", id.c_str());
+                result->SetSourceGames({ OBJECT_SOURCE_CUSTOM });
+            }
+        }
+        // >90% of objects is custom, so allow omitting the parameter without displaying an error.
```
tupaschoal, line 431:
```suggestion
        // >90% of objects are custom, allow omitting the parameter without displaying an error.
```

src/openrct2/object/ObjectFactory.cpp
Gymnasiast: I'd like @mwnciau to explain this a bit further.

tupaschoal: There's an explanation here https://github.com/OpenRCT2/OpenRCT2/pull/12667#issuecomment-673454586

mwnciau: There's also some more explanation [here](https://github.com/OpenRCT2/OpenRCT2/pull/12667#discussion_r488202243). If it'd be helpful, I can compile something more comprehensive together into a wiki article so we can reference it outside the context of that PR.

Basically, when a `json_t` child node is accessed with the `[]` operator, the library will return a reference to another `json_t`. If the key doesn't already exist, the library has to return a reference to something, so it adds a new null JSON node at that key location. This is fine if the parent node is mutable, but if it's const, the behaviour is undefined so the library will assert or throw an exception based on how it's configured.

tupaschoal: If you can write a page/section in our wiki under "Data structures", that would be great.

tupaschoal (approved):

PR #13084, opened 2020-10-03
Make some PVS-Studio fixes

### LINKED ISSUE #12523: Make PVS-Studio Fixes
Fix the issues identified by PVS Studio in the following post:

https://www.viva64.com/en/b/0749/

vjavs:
PVS-Studio is a tool for static code analysis. It has a free license for open-source projects.

Some warnings for this project were reported [here](https://www.viva64.com/en/b/0749/)

Close #12523

### DIFF
--- a/src/openrct2-ui/input/MouseInput.cpp
+++ b/src/openrct2-ui/input/MouseInput.cpp
@@ -1089,8 +1089,6 @@ void process_mouse_over(const ScreenCoordsXY& screenCoords)
 
     if (window != nullptr)
     {
-        int32_t ebx, edi;
-        rct_window* subWindow;
         rct_widgetindex widgetId = window_find_widget_from_point(window, screenCoords);
         if (widgetId != -1)
         {
@@ -1107,21 +1105,6 @@ void process_mouse_over(const ScreenCoordsXY& screenCoords)
                         break;
                     }
                     cursorId = gCurrentToolId;
-                    subWindow = window_find_by_number(
-                        gCurrentToolWidget.window_classification, gCurrentToolWidget.window_number);
-                    if (subWindow == nullptr)
-                        break;
-
-                    ebx = 0;
-                    edi = cursorId;
-                    // Window event WE_UNKNOWN_0E was called here, but no windows actually implemented a handler and
-                    // it's not known what it was for
-                    cursorId = edi;
-                    if ((ebx & 0xFF) != 0)
-                    {
-                        set_cursor(cursorId);
-                        return;
-                    }
                     break;
 
                 case WWT_FRAME:
--- a/src/openrct2/object/ObjectFactory.cpp
+++ b/src/openrct2/object/ObjectFactory.cpp
@@ -472,6 +472,7 @@ namespace ObjectFactory
             result->ReadJson(&readContext, jRoot);
             if (readContext.WasError())
             {
+                delete result;
                 throw std::runtime_error("Object has errors");
             }
 
--- a/src/openrct2/rct12/RCT12.h
+++ b/src/openrct2/rct12/RCT12.h
@@ -589,9 +589,9 @@ assert_struct_size(RCT12EntranceElement, 8);
 struct RCT12BannerElement : RCT12TileElementBase
 {
 private:
-    uint8_t index;    // 4
-    uint8_t position; // 5
-    uint8_t flags;    // 6
+    uint8_t index;        // 4
+    uint8_t position;     // 5
+    uint8_t AllowedEdges; // 6
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunused-private-field"
     uint8_t unused; // 7
--- a/src/openrct2/ride/gentle/ObservationTower.cpp
+++ b/src/openrct2/ride/gentle/ObservationTower.cpp
@@ -35,18 +35,23 @@ void vehicle_visual_observation_tower(
     int32_t baseImage_id = (vehicle->restraints_position / 64);
     if (vehicle->restraints_position >= 64)
     {
-        if ((imageDirection / 8) && (imageDirection / 8) != 3)
+        auto directionOffset = imageDirection / 8;
+        if ((directionOffset == 0) || (directionOffset == 3))
         {
-            baseImage_id *= 2;
-            baseImage_id += vehicleEntry->base_image_id + 28;
-            if ((imageDirection / 8) != 1)
-            {
-                baseImage_id -= 6;
-            }
+            baseImage_id = vehicleEntry->base_image_id + 8;
         }
         else
         {
-            baseImage_id = vehicleEntry->base_image_id + 8;
+            baseImage_id *= 2;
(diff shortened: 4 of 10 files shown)


src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -1089,7 +1089,8 @@ void process_mouse_over(const ScreenCoordsXY& screenCoords)
 
     if (window != nullptr)
     {
-        int32_t ebx, edi;
+        int32_t ebx;
+        [[maybe_unused]] int32_t edi;
```
duncanspumpkin, line 1093: just remove edi completely we don't need it anymore.

src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -1116,7 +1117,6 @@ void process_mouse_over(const ScreenCoordsXY& screenCoords)
                     edi = cursorId;
                     // Window event WE_UNKNOWN_0E was called here, but no windows actually implemented a handler and
                     // it's not known what it was for
-                    cursorId = edi;
                     if ((ebx & 0xFF) != 0)
```
duncanspumpkin, line 1120: This is always false. remove whole if case and comment.

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -440,6 +440,7 @@ namespace ObjectFactory
             result->ReadJson(&readContext, jRoot);
             if (readContext.WasError())
             {
+                delete result;
```
duncanspumpkin, line 443: This does fix this specific instance but we still need to track that this needs conversion to using something like a unique ptr. This PR is not to be merged without making an issue for that.

src/openrct2/rct12/RCT12.h
```diff
@@ -589,9 +589,9 @@ assert_struct_size(RCT12EntranceElement, 8);
 struct RCT12BannerElement : RCT12TileElementBase
 {
 private:
-    uint8_t index;    // 4
-    uint8_t position; // 5
-    uint8_t flags;    // 6
+    uint8_t index;        // 4
+    uint8_t position;     // 5
+    uint8_t banner_flags; // 6
```
duncanspumpkin, line 594: You will need to check every instance of RCT12BannerElement to make sure if it was meant to use banner_flags or flags.

src/openrct2/ride/gentle/ObservationTower.cpp
```diff
@@ -35,11 +35,11 @@ void vehicle_visual_observation_tower(
     int32_t baseImage_id = (vehicle->restraints_position / 64);
     if (vehicle->restraints_position >= 64)
     {
-        if ((imageDirection / 8) && (imageDirection / 8) != 3)
+        if (auto directionOffset = imageDirection / 8; (directionOffset != 0) && (directionOffset != 3))
```
duncanspumpkin, line 38: needs checked against the assembly. Also we tend not to use the if with initializer.

src/openrct2/object/ObjectFactory.cpp
vjavs: Please see #13100

src/openrct2/rct12/RCT12.h
vjavs: Luckily that member is private, so the changes were easily done.

src/openrct2/ride/gentle/ObservationTower.cpp
vjavs: I wrote two functions to get the difference between the assemblies.

First:
```
#include <cstdint>

auto f(int32_t imageDirection){
    return (imageDirection / 8) && (imageDirection / 8) != 3;
}
```
[https://godbolt.org/z/7Yczqe](https://godbolt.org/z/7Yczqe)

Second:
```
#include <cstdint>

auto f(int32_t imageDirection){
    auto directionOffset = imageDirection / 8;
    return (directionOffset != 0) && (directionOffset != 3);
}
```
[https://godbolt.org/z/hz61Te](https://godbolt.org/z/hz61Te)

The diff between assemblies can be seen here: [https://www.diffchecker.com/OJKsMcBr](https://www.diffchecker.com/OJKsMcBr)

I ran these functions in all range of ```int32_t``` and results were identical.
I've noticed that this same technique was already employed here: [LaunchedFreefall.cpp#L61](https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/thrill/LaunchedFreefall.cpp#L61)

duncanspumpkin: No I mean checked against the original rct2 assembly

vjavs: Here we have artifacts before and after the change:
 - [before](https://github.com/vjavs/OpenRCT2/actions/runs/288461434)
 - [after](https://github.com/vjavs/OpenRCT2/actions/runs/288494654)

But I have no clue how to compare them. Either against each other or against the original rct2 assembly.
Actually I checksum file `libopenrct2.so` and they diverged.

How could we do that? Also, I would love to know why is it necessary to check against the binary in such a condition.

duncanspumpkin: It's necessary because it looks wrong so I'd like to make sure that it is doing what it is meant to do. You can check the assembly with a debugger/dissassembler like olydbg or ida or you can just leave it for me to check

vjavs: > or you can just leave it for me to check

Sure, please.

duncanspumpkin: image
Yup all looks fine. Could you make it a bit more readable by moving around the if:

```
        auto directionOffset = imageDirection / 8;
        if ((directionOffset == 0) || (directionOffset == 3))
        {
            baseImage_id = vehicleEntry->base_image_id + 8;
        }
        else
        {
            baseImage_id *= 2;
            baseImage_id += vehicleEntry->base_image_id;
            if (directionOffset == 1)
            {
                 baseImage_id += 28;
            }
            else
            {
                baseImage_id += 22;
            }
        }

```

vjavs: Nice check! :)

src/openrct2/rct12/RCT12.h
duncanspumpkin: I question if these should even be called flags maybe should call it allowed edges. Give me a little while to check things

duncanspumpkin: after a review of the codebase I dont think these should be called flags. They are `AllowedEdges` can you please rename them. Can you also rename the `BannerElement` version of them as well `TileElement.h` will need modifcations in `Banner.cpp` as well. I realise that `AllowedEdges` wont match the code style surrounding this but that is how it should be styled as per our coding style document

vjavs: That totally makes sense.

duncanspumpkin (approved):

tupaschoal (approved):

Issue #7490, opened 2018-05-04
Missing Sound Effects


FlatoutDude:
**OS:** Windows 10
**Version:** V0.0.13
**Commit/Build:** Build dbbd5b8

No sound effects for various menu items and when building things.

- [no] Reproducible in RCT2 (vanilla)?
- [no] Specific to multiplayer?

**Steps to reproduce:**
1. load an RCT 1 Scenario, for me this was dynamite dunes
2. try placing paths and structures such as rides and stalls, and navigating menu panes while in-game.

**Dump file**
I don't have a dump file right now.

**Screenshots / Video:**
https://www.youtube.com/watch?v=TVp8J0V-NDw&feature=youtu.be

(i've noticed that sound effects appear to be working in the video, but they still persist to not work when actually in game.)

**Save game:**
[Dynamite Dunes.sv6.txt](https://github.com/OpenRCT2/OpenRCT2/files/1976128/Dynamite.Dunes.sv6.txt)

FlatoutDude:
edit: i can hear the sounds for when i am building coasters, it's just that they are very, very faint, how can i fix this? the sounds are maxed out and there isn't any music playing.

AaronVanGeffen:
Have you only looked at OpenRCT2's volume settings, or also in Windows' mixer settings? Since Windows Vista, Windows has allowed you to set the volume on a per-application basis. Perhaps something has accidentally gone wrong there?

Umdlye:
The sound effect volume is wrong here too. Was used to it/never really noticed it until I booted up the vanilla games yesterday. Some sounds play at full volume (raising/lowering water, opening windows), all the others are much quieter. In vanilla all the UI sound effects are equally loud as far as I can tell.

Windows 10 (x64), appears in release as well as develop.

meatbee:
I'm having the same issue. I miss the coaster build sound effects.

NicNamSam:
Same issue, 19 April, 2020. No "click" sounds when using menus or placing paths, tracks, or rides.

tupaschoal:
I ran every release on GitHub from `0.2.6` to `0.0.5` and in all of them the sounds of building appears to be muffled when compared to Vanilla
[muffled_build.zip](https://github.com/OpenRCT2/OpenRCT2/files/4535524/muffled_build.zip)

lah7:
[...]

Issue #7443, opened 2018-04-23
button click sound is smaller than vanilla version & yellow arrow flickring on build mode


guljam:
click sound "TOCK TOCK"
and
installing object sound "BOOK BOOK" is smaller than vanillla version.

![gif3](https://user-images.githubusercontent.com/15816034/39157236-e45bafe0-4794-11e8-811f-8e1fa4074ed2.gif)

Some ride's buildmode's yellow cursor is very blinking.(ex. Car Ride)

duncanspumpkin:
Unsure what you mean by sound issue.

Yellow flickering arrow has been annoying for quite some time. It needs reworked.

tupaschoal:
Sound effect seems to be the same as #7490 and will be tracked there

PR #13097, opened 2020-10-04
Construction arrow pulse improvements

mwnciau:
While working on #13074, I noticed that the arrow/maze section pulsing was inconsistent and would occasionally flash on and off in bursts. This PR changes the pulsing for ride and footpath construction to be time based. Additionally, the arrow pulse for station pieces has been restored.

The pulse speed of the arrow can be controlled with the `ARROW_PULSE_DURATION` static variable in `RideConstruction.cpp` and `Footpath.cpp`.

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -25,6 +25,7 @@
 #include <openrct2/config/Config.h>
 #include <openrct2/localisation/Localisation.h>
 #include <openrct2/network/network.h>
+#include <openrct2/platform/platform.h>
 #include <openrct2/ride/Ride.h>
 #include <openrct2/ride/RideData.h>
 #include <openrct2/ride/Track.h>
@@ -38,6 +39,7 @@
 static constexpr const rct_string_id WINDOW_TITLE = STR_RIDE_CONSTRUCTION_WINDOW_TITLE;
 static constexpr const int32_t WH = 394;
 static constexpr const int32_t WW = 166;
+static constexpr const uint16_t ARROW_PULSE_DURATION = 200;
 
 #pragma region Widgets
 
@@ -441,6 +443,7 @@ static uint8_t _currentlyShowingBrakeOrBoosterSpeed;
 static bool _boosterTrackSelected;
 
 static uint32_t _currentDisabledSpecialTrackPieces;
+static uint32_t _rideConstructionNextArrowPulse = 0;
 
 static void window_ride_construction_construct(rct_window* w);
 static void window_ride_construction_mouseup_demolish(rct_window* w);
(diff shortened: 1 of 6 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2546,6 +2546,8 @@ void sub_6C94D8()
         _currentTrackSelectionFlags &= ~TRACK_SELECTION_FLAG_RECHECK;
     }
 
+    uint32_t curTime;
```
ZehMatt, line 2549: Initialize the variable here

mwnciau: I wanted to avoid calling `platform_get_ticks()` in construction states where it isn't relevant. I've added scope to the case blocks instead so we don't have to declare it before the switch statement.

Gymnasiast:
Your local Git settings are incorrect - the email address does not match the one on Github. (Hint: you can set the local email address to `mwnciau@users.noreply.github.com` if you don't want to give away your email address.)

tupaschoal (commented):
@mwnciau would you be able to provide before/after comparison gifs?

ZehMatt:
I'm a bit hesitant over this, this now relies on the system clock and no longer on the game ticks, but its also not something that affects the game state as far I'm concerned.

mwnciau:
Pre-fix, low fps on the left (~10fps), high fps on the right (144fps - it's flashing so quickly that the gif can't quite capture it):

pre-fix low-fps maze pre-fix high-fps maze

Post-fix, low fps on the left and high fps on the right:

post-fix low-fps maze post-fix high-fps maze

mwnciau:
Pre-fix, low and high fps:

pre-fix low-fps ride pre-fix high-fps ride

Post-fix, low and high fps:

post-fix low-fps ride post-fix high-fps ride

These gifs aren't perfect - you'll get a better idea if you just build and run the fix and compare to a pre-fix build.

The flag that's set is visual only as far as I'm aware, so shouldn't affect game state/multiplayer at all.

tupaschoal (approved):
This looks alright to me, especially with the gifs. @ZehMatt are you ok with merging?

ZehMatt (approved):

tupaschoal:
I guess we could use two changelog entries here @mwnciau :
 
`Fix: [#13097] Construction arrows pulse at irregular intervals.`
`Fix: [#13097] Missing direction arrow for stations`

mwnciau:
It looks like an issue exists for the flickering: #7443. At least partially - it's not the best quality issue.

tupaschoal:
You can quote that one instead for the fix on the pulsing, both on the changelog and on the commit title. 

The sound issue mentioned there seems to be the same as #7490, so we can track it in this other.

mwnciau:
Sorted the changelog now

tupaschoal:
Thanks!

steamroll:
I noticed a slight graphical issue with the arrow when removing and adding track pieces for rides where the arrow will stay present after a track piece is placed. This also occurs with the grid curtain that appears when elevating scenery pieces, but I do not have a screenshot to show the issue.

I was playing on develop build 5ec5ece when this started happening.

Diamond Heights 2020-10-16 04-26-24

Issue #1324, opened 2015-06-13
Place ride entrance / exit tile glitch


IntelOrca:
When placing a ride entrance or ride exit, if the location is invalid the tiles where your next track piece would go gets highlighted. This can be reproduced by placing an entrance and then hovering the cursor for the ride exit over the ride entrance.

![image](https://cloud.githubusercontent.com/assets/1482259/8144014/e57bf6c6-11c0-11e5-9e77-aca00108df1b.png)

duncanspumpkin:
Original Bug.

PR #13098, opened 2020-10-04
Entrance/exit construction state fixes in ride and maze construction

mwnciau:
A couple of bug fixes for the RIDE_CONSTRUCTION_STATE_ENTRANCE_EXIT construction state in rides and mazes:

#### Ride Construction map selections

Currently, when placing an entrance or exit on a track ride that's incomplete, the map selection of the last ghost piece is visible:

image

The first commit, 8820d90, removes the existing selection when in the entrance exit construction state

#### Maze construction UX

Currently, when placing an entrance or exit on a maze, the arrow buttons are still usable even though the current position indicator is hidden. This PR makes a few changes to improve the UX of the maze construction window:

* The arrow buttons are disabled in the entrance/exit state
* When leaving the entrance/exit state, the previous state is restored
* When in the entrance/exit state, clicking one of the maze build mode buttons will disabled the active entrance/exit buttons
* The entrance/exit buttons now switch along with the widget switch when the first entrance/exit is placed

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -20,10 +20,13 @@
 - Fix: [#13044] Rides in RCT1 saves all have "0 customers per hour".
 - Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
 - Fix: [#13083] Dialog for renaming conflicting track design crops text out.
+- Fix: [#13098] UI buttons for entrance and exit don't toggle according to them being built.
+- Fix: [#13098] Maze can still be constructed while placing entrance and exit (original bug).
 - Fix: [#13118] Closing construction window resets ride viewport.
 - Fix: [#13129] Missing error message when waiting for train to leave station on the ride measurements graph.
 - Fix: [#13138] Fix logical sorting of list windows.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
+- Improved: [#13098] Improvements to the maze construction window user interface
 - Improved: [#13125] Selecting the RCT2 files now uses localised dialogs.
 
 0.3.1 (2020-09-27)
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -5,6 +5,7 @@
 - Feature: [#13000] objective_options command for console.
 - Feature: [#13096] Add Esperanto translation.
 - Change: [#9568] Change lift sounds of Reverser Roller Coaster and Compact Inverted Coaster to better fitting ones.
+- Fix: [#1324] Last track piece map selection still visible when placing ride entrance or exit (original bug).
 - Fix: [#3200] Close Construction window upon selecting vehicle page.
 - Fix: [#4041] Garbled park option on scenario editor with custom theme.
 - Fix: [#5904] Empty errors on tile inspector base height change.
--- a/src/openrct2-ui/windows/MazeConstruction.cpp
+++ b/src/openrct2-ui/windows/MazeConstruction.cpp
@@ -186,11 +186,12 @@ static void window_maze_construction_entrance_mouseup(rct_window* w, rct_widgeti
 
     ride_construction_invalidate_current_track();
 
-    // ???
-    uint8_t old_state = _rideConstructionState;
+    if (_rideConstructionState != RIDE_CONSTRUCTION_STATE_ENTRANCE_EXIT)
+    {
+        gRideEntranceExitPlacePreviousRideConstructionState = _rideConstructionState;
+    }
     _rideConstructionState = RIDE_CONSTRUCTION_STATE_ENTRANCE_EXIT;
-    if (old_state != RIDE_CONSTRUCTION_STATE_ENTRANCE_EXIT)
-        _rideConstructionState = old_state;
+
     window_maze_construction_update_pressed_widgets();
 }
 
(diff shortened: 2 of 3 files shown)


src/openrct2-ui/windows/MazeConstruction.cpp
```diff
@@ -254,21 +279,6 @@ static void window_maze_construction_resize(rct_window* w)
  */
 static void window_maze_construction_mousedown(rct_window* w, rct_widgetindex widgetIndex, rct_widget* widget)
 {
-    switch (widgetIndex)
```
duncanspumpkin, line 257: Why the switch from mouse down to mouse up

duncanspumpkin (commented):
I don't understand why you are changing when these actions occur

mwnciau:
I'm happy to change it back, but I felt that the previous behaviour was a bit inconsistent.

image

The Mode section actions on mouse down, and the Build and Entrance/Exit sections action on mouse up, meaning you can click on them and then move off before releasing the click to prevent the button press action.

The Mode and Build sections are visually similar so it's a little strange that they behave differently, but on the other hand they are functionally different so it can be argued that it makes sense.

The Mode and Entrance/Exit sections are functionally similar - they both act as radio button inputs (in fact, they more or less make a 5 element radio input for construction mode) - so again, it's a little strange that they behave differently, but on the other hand they are visually inconsistent so the case could be made.

Both together, however, feels a bit unpolished. The answer may be instead to change the Entrance/Exit buttons to action on mouse down instead of mouseup.

duncanspumpkin:
How does it compare to the ride construction window as I suspect its following the same style as that window in that the central box is a mouse down but all the outside elements are a mouse up

mwnciau:
For comparison:

image

Yeah, in the ride construction window, the direction/slope/bank are all mouse down, and the big build button is mouse up. They are a little different: they are greyed out when the entrance/exit build mode is enabled, and the maze mode buttons are a lot more functionally similar to the entrance/exit buttons.

Playing around with it some more, the radio button functionality does feel a lot cleaner on mouse down so I'll change it back. I do think the entrance/exit buttons should behave the same way, i.e. on mouse down.

mwnciau:
Changed it back to mouse down now.

Also added another minor fix (MazeConstruction line 393): previously, when you placed the first entrance or exit, the tool would change to placing the other, but the UI wouldn't change to reflect that (you'd be placing an exit with the entrance button depressed, and vice-versa).

mwnciau:
For the changelog, would something like this suffice?

```
- Fix: [#13098] Last track piece map selection still visible when placing ride entrance or exit
...
- Improved: [#13098] Improvements to the maze construction window user interface
```

Or something a little more generic:

```
- Fix: [#13098] Various minor issues with entrances and exits in the ride construction window
```

duncanspumpkin:
Go with the two separate items. The fix has been long overdue

mwnciau:
Added changelog.

Found the original issue for the fix: #1324

duncanspumpkin (approved):
Can we have a few testers on this?

distribution/changelog.txt
```diff
@@ -13,7 +13,9 @@
 - Fix: [#13029] Not all Junior Roller Coaster pieces are shown when "Show all track pieces" cheat is enabled.
 - Fix: [#13044] Rides in RCT1 saves all have "0 customers per hour".
 - Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
+- Fix: [#1324] Last track piece map selection still visible when placing ride entrance or exit (original bug).
```
AaronVanGeffen, line 16: Could you move this line up a little, so the fixes appear in numerical order?

mwnciau: oops, yep fixed now

tupaschoal (approved):

tupaschoal:
I've done some basic in-game testing and this looks great.

One note regarding the maze changes:
> Currently, when placing an entrance or exit on a maze, the arrow buttons are still usable even though the current position indicator is hidden. This PR makes a few changes to improve the UX of the maze construction window:
> - The arrow buttons are disabled in the entrance/exit state

This kind of happens on vanilla RCT as well, though with a slight difference.
- **Vanilla**: Clicking entrance/exit button does not disable the arrows. Clicking them doesn't do anything, but if you click any of the mode buttons, you can use the arrows normally, while still holding the entrance/exit object.
- **OpenRCT2**: Clicking entrance/exit button does not disable the arrows. Clicking them ~doesn't do anything, but if you click any of the mode buttons,~ you can use the arrows normally, while still holding the entrance/exit object.

So there's a slight difference in behavior, one might even call it an `original bug`, but I'm not vouching strongly for this here. I do think it could be split into more changelog entries:
- Fix: [#1324] Last track piece map selection still visible when placing ride entrance or exit (original bug).
- Fix: [#13098] UI buttons for entrance and exit don't toggle according to them being built.
- Fix: [#13098] Maze can still be constructed while placing entrance and exit (original bug).
- Improved: [#13098] Improvements to the maze construction window user interface.


It sounds a bit verbose, but after all these are multiple fixes and some enhancements. Personally I'd do each of them in a single commit, so that would also take to a changelog each, it also becomes more informative for the players. Let's see what the other team members think before you do these.

distribution/changelog.txt
```diff
@@ -15,6 +15,7 @@
 - Fix: [#13044] Rides in RCT1 saves all have "0 customers per hour".
 - Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
+- Improved: [#13098] Improvements to the maze construction window user interface
```
tupaschoal, line 18:
```suggestion
- Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
- Fix: [#13098] UI buttons for entrance and exit don't toggle according to them being built.
- Fix: [#13098] Maze can still be constructed while placing entrance and exit (original bug).
- Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
- Improved: [#13098] Improvements to the maze construction window user interface
```

tupaschoal:
I've modified the changelog, but there's a rebase conflict. Can you please solve it so we can merge it @mwnciau ?

You can fold my changelog commit into the relevant one

mwnciau:
Rebased to develop now

PR #13104, opened 2020-10-05
Use RemoveViewport helper function

duncanspumpkin:
I noticed that #13030 added the helper function but there was a lot of additional places where it could be used.


src/openrct2-ui/windows/Park.cpp
```diff
@@ -763,10 +763,8 @@ static void window_park_init_viewport(rct_window* w)
     }
     else
     {
-        viewport = w->viewport;
-        w->viewport = nullptr;
         viewportFlags = viewport->flags;
```
AaronVanGeffen, line 766: This variable is no longer assigned to `w->viewport`.

AaronVanGeffen (approved):

PR #13105, opened 2020-10-05
Fix #7210: Land tile smoothing occurs with edge tiles

### LINKED ISSUE #7210: Land tile smoothing occurs with edge tiles
**Version:** 0.1.2
**Commit/Build:** Develop d5d9441

With landscape smoothing turned on, the hidden edge tiles (which are grass by default and cannot be changed with landscape tools) will blend with the 'apparent'(visible) edge tiles at the same height. This looks quite odd. Ideally, the drawing engine should implement a check to disable blending between an invisible edge tile and a visible tile.

**Steps to reproduce:**
1. Turn on 'landscape smoothing' in the Options menu on the second tab
2. Lower one of the visible edge tiles down to the minimum height and paint it with a blendable texture. The tile will have a visible strip of grass blending next to the invisible tile.

**Screenshots / Video:**
once upon a time in america 2018-02-22 14-37-10

Gymnasiast:


tupaschoal (commented):
I tested it and it seems to have more side effects than benefits. I've recorded two gifs, notice how the tile on the corner doesn't have the smoothing effect in all camera angles:

**With this PR**
WithPatch

**Without this PR**
WithoutPatch

PR #13113, opened 2020-10-06
Port some OpenLoco helper functions

duncanspumpkin:
Added a few helper functions from OpenLoco to make viewport code a bit simpler

### DIFF
--- a/src/openrct2/interface/Window.h
+++ b/src/openrct2/interface/Window.h
@@ -132,8 +132,6 @@ struct rct_widget
     }
 };
 
-using ViewportCoordsXY = ScreenCoordsXY;
-
 /**
  * Viewport structure
  */
@@ -142,7 +140,7 @@ struct rct_viewport
     int16_t width;
     int16_t height;
     ScreenCoordsXY pos;
-    ViewportCoordsXY viewPos;
+    ScreenCoordsXY viewPos;
     int16_t view_width;
     int16_t view_height;
     uint32_t flags;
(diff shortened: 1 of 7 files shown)


src/openrct2/interface/Window.h
```diff
@@ -128,6 +128,8 @@ struct rct_widget
     }
 };
 
+using ViewportCoordsXY = ScreenCoordsXY;
```
ZehMatt, line 131: I think we should try to reduce the amount of coordinate types, we have already a lot of them imho.

duncanspumpkin: Would it be clear enough that Contains and ContainsScreen and when to use either?

ZehMatt: I think documenting the functions should provide a better way to distinguish what needs to be called.

duncanspumpkin: Okay so I'll just add a bigger comment stating when to use each of the functions.

tupaschoal (approved):

src/openrct2/interface/Window.h
```diff
@@ -147,6 +147,21 @@ struct rct_viewport
     ZoomLevel zoom;
     uint8_t var_11;
     VisibilityCache visibility;
+
+    // Use this function on coordinates that are relative to the viewport zoom i.e. a peeps x, y position after transforming from its x, y, z
```
tupaschoal, line 151:
```suggestion
    // Use this function on coordinates that are relative to the viewport zoom i.e. a peeps x, y position after transforming
    // from its x, y, z
```

PR #13114, opened 2020-10-06
Fix #10932: Extensive warnings in console when placing an entrance or exit

### LINKED ISSUE #10932: RideEntranceExitRemoveAction very verbose
**OS:** Linux
**Version:** 0.2.4+
**Commit/Build:** 9b5e8140

The RideEntranceExitRemoveAction is very verbose. Just attempting to place an entrance or exit in an empty park (Forest Frontiers) leads to this message spamming the console:

```
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
WARNING[../src/openrct2-ui/../openrct2/actions/../actions/RideEntranceExitRemoveAction.hpp:102 (Query)]: Track Element not found. x = 1440, y = 960, ride = 0, station = 0
```

mwnciau:
This PR fixes the issue raised in #10932. I think it was partially resolved in #13074, but this PR completes the fix.

The problem was that when moving an entrance or exit (i.e. the entrance/exit already exists), RideEntranceExitRemoveAction would look for the entrance/exit to remove. Because the entrance/exit already existed, it was never placed, and so not found hence incorrectly outputting the warning message. The PR simply checks for this edge case and returns without the message.

### DIFF
--- a/src/openrct2/actions/RideEntranceExitRemoveAction.cpp
+++ b/src/openrct2/actions/RideEntranceExitRemoveAction.cpp
@@ -48,13 +48,8 @@ void RideEntranceExitRemoveAction::Serialise(DataSerialiser& stream)
 static TileElement* FindEntranceElement(
     const CoordsXY& loc, RideId rideIndex, StationIndex stationNum, int32_t entranceType, uint32_t flags)
 {
-    const bool isGhost = flags & GAME_COMMAND_FLAG_GHOST;
     for (auto* entranceElement : TileElementsView<EntranceElement>(loc))
     {
-        // If we are removing ghost elements
-        if (isGhost && entranceElement->IsGhost() == false)
-            continue;
-
         if (entranceElement->GetRideIndex() != rideIndex)
             continue;
 
@@ -96,7 +91,12 @@ GameActions::Result RideEntranceExitRemoveAction::Query() const
     auto* entranceElement = FindEntranceElement(
         _loc, _rideIndex, _stationNum, _isExit ? ENTRANCE_TYPE_RIDE_EXIT : ENTRANCE_TYPE_RIDE_ENTRANCE, GetFlags());
 
-    if (entranceElement == nullptr)
+    // If we are trying to remove a ghost and the element we found is real, return an error, but don't log a warning
+    if (entranceElement != nullptr && (GetFlags() & GAME_COMMAND_FLAG_GHOST) && !(entranceElement->IsGhost()))
+    {
+        return GameActions::Result(GameActions::Status::InvalidParameters, STR_NONE, STR_NONE);
+    }
+    else if (entranceElement == nullptr)
     {
         log_warning(
             "Track Element not found. x = %d, y = %d, ride = %u, station = %u", _loc.x, _loc.y, _rideIndex.ToUnderlying(),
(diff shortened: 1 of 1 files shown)


tupaschoal (commented):
This looks fine to me, though I'd appreciate input from someone else

ZehMatt:
I wonder if the changelog is necessary, the logged information is normally not visible on release builds.

mwnciau:
Fair point - I've removed the changelog entry.

src/openrct2/actions/RideEntranceExitRemoveAction.hpp
```diff
@@ -110,7 +107,12 @@ DEFINE_GAME_ACTION(RideEntranceExitRemoveAction, GAME_COMMAND_REMOVE_RIDE_ENTRAN
             break;
         } while (!(tileElement++)->IsLastForTile());
 
-        if (!found)
+        // Silently prevent removing an existing entrance/exit if we're trying to place a ghost
```
duncanspumpkin, line 110: This is the remove action this doesn't make much sense. We aren't trying to place a ghost we are trying to remove a ghost but we have a real tile element here instead. Can you change the comment and use a different GA_ERROR as I don't see it as the correct one to use for this.

duncanspumpkin:
I'm not completely convinced that this is the right location to fix the issue. The game action is performing correctly. Its the calling code that needs to be adjusted to not try remove an entrance when it hasn't even placed one yet.

src/openrct2/actions/RideEntranceExitRemoveAction.hpp
mwnciau: Reworded so hopefully that's a bit clearer, and changed the error to `GA_ERROR::INVALID_PARAMETERS`.

mwnciau:
> I'm not completely convinced that this is the right location to fix the issue. The game action is performing correctly. Its the calling code that needs to be adjusted to not try remove an entrance when it hasn't even placed one yet.

`RideEntranceExitPlaceAction` is essentially using `RideEntranceExitRemoveAction::Query` to check whether the station can be removed or not. If it can't remove the existing element, then it can't place the new element and the error bubbles down. Other than calling the remove action, it doesn't have any interaction with the existing station, nor does anything else in the call stack.

I don't know if this was an intended use of game actions, but the alternative would basically mean copying the code from `RideEntranceExitRemoveAction` more or less verbatim, which seems a bit messy. That, or moving it all to a separate function, which seems a bit wasteful as you're then duplicating the work of finding the right tile element.

I'm not a huge fan of using `GA_ERROR::INVALID_PARAMETERS` for the error in this particular case - it's more like an invalid state than invalid parameters, but there doesn't seem to be a better fitting existing error enum.

duncanspumpkin:
I still feel that it should be the calling function that fixes this. Its how every other function works. But I'm not that fussed.

duncanspumpkin (approved):

PR #13120, opened 2020-10-07
Fix #10186: Placing multiple saved rides ignores design name

### LINKED ISSUE #10186: Placing multiple saved rides ignores the design name.
**OS:** [Windows 10 x64]
**Version:** [ 0.2.4]
**Commit/Build:** [d645338]

Placing more than one saved ride ignores the saved design name and treats the name like a generic new design.  This didn't occur in 0.2.3.

EX: If I place multiple saved mazes named "micromaze", it will just name the second maze "Maze 1" and the third "Maze 2" instead of just keeping the "micromaze" name. In the previous release (v0.2.3), it would just place all mazes as "micromaze." I cannot recreate in RCT2 to know which naming convention is desired, and I cannot remember how releases prior to 0.2.3 behaved.

If you change the name of the first placed coaster before placing a new one, the new one will remember the design name, so the game only ignores the design name if a ride is already placed with that name.  Renaming another ride to the same name flags a "name already exists" message, so perhaps an undocumented change was made to protect from having multiple rides with same name.   Perhaps it would be preferable to increment the design name, like "micromaze 2", "micromaze 3" instead of bypassing it entirely.  If the saved name is too long, truncating/replacing the final few characters is always an option.

Unknown, I cannot save rides in RCT2 due to a device I/O error leading into a crash-to-desktop.
no

**Steps to reproduce:**
1.  Place a saved ride and do not change name
2.  Place an identically-named ride
3.  View name of newly placed ride.

**screenshot**
image

I placed two of the "6x9 Classic Mini" designs and 4 of the "5x5 Spinning Box" rides.  This is how they were named after placement.  In v0.2.3 all the Spinning Wild Mouse coasters would've shared the "5x5 Spinning Box" name without any errors and both of the Classic Mini Coasters would've been named "6x9 Classic Mini".

**Save game:  I made simple numbers with pathing to show first ride placed vs second**
[naming example.txt](https://github.com/OpenRCT2/OpenRCT2/files/3801136/naming.example.txt

tupaschoal:
`RideSetNameAction` returns an uncaught error if the name already exists, thus the renaming never happened for custom rides.

### DIFF
--- a/src/openrct2/actions/TrackDesignAction.cpp
+++ b/src/openrct2/actions/TrackDesignAction.cpp
@@ -221,9 +221,13 @@ GameActionResult::Ptr TrackDesignAction::Execute() const
         ride->vehicle_colours[i].Ternary = _td.vehicle_additional_colour[i];
     }
 
-    auto gameAction = RideSetNameAction(ride->id, _td.name);
-    gameAction.SetFlags(GetFlags());
-    GameActions::ExecuteNested(&gameAction);
+    for (int32_t count = 1; count == 1 || r->Error != GA_ERROR::OK; ++count)
+    {
+        auto name = count == 1 ? _td.name : (_td.name + " " + std::to_string(count));
+        auto gameAction = RideSetNameAction(ride->id, name);
+        gameAction.SetFlags(GetFlags());
+        r = GameActions::ExecuteNested(&gameAction);
+    }
     res->Cost = cost;
     res->rideIndex = ride->id;
     return res;
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -9,6 +9,7 @@
 - Fix: [#6086] Cannot install existing track design with another name.
 - Fix: [#8015] RCT2 files are not found when put into the OpenRCT2 folder.
 - Fix: [#8957] Error title missing when building with insufficient funds
+- Fix: [#10186] Placing multiple saved rides ignores design name (original bug).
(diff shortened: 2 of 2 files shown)


src/openrct2/actions/TrackDesignAction.cpp
```diff
@@ -221,9 +221,13 @@ GameActionResult::Ptr TrackDesignAction::Execute() const
         ride->vehicle_colours[i].Ternary = _td.vehicle_additional_colour[i];
     }
 
-    auto gameAction = RideSetNameAction(ride->id, _td.name);
-    gameAction.SetFlags(GetFlags());
-    GameActions::ExecuteNested(&gameAction);
+    for (int32_t count = 0; count == 0 || r->Error != GA_ERROR::OK; ++count)
```
AaronVanGeffen, line 224: I wonder if it's better to start with `1`? With `0`, the first ends up e.g. `Crazy Mice`, while the second one gets `Crazy Mice 1`, third `Crazy Mice 2`, etc. It would be more intuitive if the second one were to be assigned `2` as well.

AaronVanGeffen (commented):
Nice solution. I think it'd be better to tweak the assigned number slightly, though.

src/openrct2/actions/TrackDesignAction.cpp
tupaschoal: Done!

AaronVanGeffen (approved):

PR #13122, opened 2020-10-07
Refactor PeepSpriteType to use strong enum

### LINKED ISSUE #12396: Refactor PeepSpriteType to use strong enum

vjavs:
Closes #12396


tupaschoal (approved):
Nice!

duncanspumpkin:
Nice work. Kept thinking ahh but did you remember this. Then would look and the answer would be yes.

PR #13126, opened 2020-10-08
Fix #4041: Garbled park option on scenario editor with custom theme

### LINKED ISSUE #4041: Doubly written text in scenario options when using custom theme color
**OS:** GNU/Linux (Arch Linux)
**Version:** 0.0.5-develop
**Commit/Build:** 59c3656
- Reproducible in RCT2 (vanilla)? I don't know.
- Multiplayer? N/A

**Steps to reproduce:**
1. Start scenario editor
2. Go all the way until the step “Options Selection”
3. Click on the park options tab
4. Select “Pay to enter park”
5. Open the options window (game options, not scenario options)
6. Create a new theme by duplicating the RCT2 one
7. Select the new theme
8. Edit the theme: Choose a different color for the second color in the entry “Scenario Options” of the “Editors” tab
9. Now look at the drop down list you changed in step 4

Expected: The text looks normal and is readable
Actual: The text looks weird and is hard to read

This is how it looks like for me (using my custom theme):
https://i.imgur.com/XPDXUX4.png

Funny is, this bug seems _only_ to manifest for only some particular colors. For a few colors, including gray (i.e. the color used in the “RCT1” and “RCT2” themes), the bug is invisible.

If you look closely, it seems that the other text of this drop-down list is just “written over” the other one. You can easily see it when you switch between the drop-down list entries after following my steps.

tupaschoal:
Regardless of what had been chosen, it was always drawing `STR_FREE_PARK_ENTER` first.
**Before:**
before
**After:**
after

### DIFF
--- a/src/openrct2-ui/windows/EditorScenarioOptions.cpp
+++ b/src/openrct2-ui/windows/EditorScenarioOptions.cpp
@@ -1360,7 +1360,6 @@ static void window_editor_scenario_options_park_paint(rct_window* w, rct_drawpix
         else
             stringId = STR_PAY_PARK_ENTER;
 
-        gfx_draw_string_left(dpi, stringId, nullptr, COLOUR_BLACK, screenCoords);
         gfx_draw_string_left(dpi, STR_WINDOW_COLOUR_2_STRINGID, &stringId, COLOUR_BLACK, screenCoords);
     }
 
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Why is it doing two draws of the same string?

tupaschoal:
I have no idea, I thought it might be drawing a shade or something, because when it was garbled, it was grey, but that might just be because it was not reading the colour.

AaronVanGeffen:
They're drawing on the same position. I think we can just remove the first call altogether.

duncanspumpkin:
I checked the assembly and the first call does not exist. Looks like we added this mistake during implementation

src/openrct2-ui/windows/EditorScenarioOptions.cpp
```diff
@@ -1361,6 +1360,7 @@ static void window_editor_scenario_options_park_paint(rct_window* w, rct_drawpix
         else
             stringId = STR_PAY_PARK_ENTER;
 
+        gfx_draw_string_left(dpi, stringId, nullptr, COLOUR_BLACK, screenCoords);
```
tupaschoal, line 1363:
```suggestion
```

tupaschoal:
I removed the call then, when merging, please squash

duncanspumpkin (approved):

PR #13127, opened 2020-10-08
Close #5520: Use std::filesystem over fts.h for recursive deletion

### LINKED ISSUE #5520: Refactor platform_directory_delete() to remove FTS dependency
`fts.h` is included, and FTS used, for `platform_directory_delete()` on POSIX systems.

This function is only called from `DeleteItem()` in `TitleSequenceManager.cpp` - the function that deletes title sequences.

Including FTS for one very infrequently called function is inefficient - `platform_directory_delete()` should be rewritten to avoid using it.

tupaschoal:
Full disclosure: I didn't test it, as I don't have a UNIX system

### DIFF
--- a/src/openrct2/platform/Posix.cpp
+++ b/src/openrct2/platform/Posix.cpp
@@ -9,17 +9,9 @@
 
 #if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__)) || defined(__FreeBSD__)
 
-#    include <cstring>
-#    include <ctype.h>
-#    include <dirent.h>
-#    include <errno.h>
-#    include <fcntl.h>
-#    include <fnmatch.h>
-#    if !defined(__EMSCRIPTEN__) && __has_include(<fts.h>)
-#        include <fts.h>
-#    endif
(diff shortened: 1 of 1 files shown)


janisozaur:
@PureTryOut you use musl-based alpine, right? Can you check? I think maybe title sequence uses that code?

AaronVanGeffen:
@janisozaur Yes, the title sequence manager is the only place where this is currently used.

PureTryOut:
@janisozaur yes I do. This PR compiles fine on my system if that is what you're asking.

tupaschoal:
I guess going to the `TitleSequenceManager` and trying to delete one of them, and make sure it works. (Duplicate one to make sure you don't lose yours :D)

AaronVanGeffen:
I wonder if this is even still needed. The title sequence manager uses zip files with a `.parkseq` extension nowadays, right? Although, I suppose directories are still supported as well…

Gymnasiast:
I manually created a title sequence using a folder instead of a ZIP/parkseq. Removing it through the Title Sequence Editor works correctly.

PR #13128, opened 2020-10-08
Fix #13125, make initial RCT2 directory picking not be hardcoded

### LINKED ISSUE #13125: "OpenRCT2 needs files from the original RollerCoaster Tycoon 2..." string is hardcoded
The string in the dialog that pops up when the game asset directory isn't set, the string is currently hardcoded, should probably be localizable

Specifically, this string in `Config.cpp`,

> OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work.
> Please select the directory where you installed RollerCoaster Tycoon 2.

keithstellyes:
An issue that was ran into is that at the point in the lifecycle where the directory is queried, language is not set.

In order to open a language in the `LocalisationService`, an `ObjectManager` was required in the API, of course don't have an `ObjectManager` yet at before we've selected the directory. The API was simply it was just calling `resetObjects()`, assumingly as a convenience measure. To me it makes sense for language and localization to be decoupled from the `ObjectManager` as we may need UI at times where we don't have it. (Like, say, picking an RCT2 directory.) Once decoupled, I simply moved language initialization logic to before picking the directory, then, finally pulling the strings dynamically.

**Testing:**
- deleted local `config.ini`
- ran program, ensured dialog was correct text
- picked bad directory, to get the "Can't find g1.dat" dialog
- picked correct directory
- changed language to Spanish at main menu, ensure texts changed
- started new save, then changed language back to English (US), ensuring text changed back

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3646,6 +3646,11 @@ STR_6387    :Can't lower element here…
 STR_6388    :Can't raise element here…
 STR_6389    :Invalid clearance
 
+# Picking RCT2 directory 
+STR_6390    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work. Please select the directory where you installed RollerCoaster Tycoon 2.
+STR_6391    :Please select your RCT2 directory
+STR_6392    :Could not find {STRING} at this path.
+
 #############
 # Scenarios #
 ################
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3645,8 +3645,6 @@ STR_6386    :Blizzard
 STR_6387    :Can't lower element here…
 STR_6388    :Can't raise element here…
 STR_6389    :Invalid clearance
-
-# Picking RCT2 directory 
 STR_6390    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work. Please select the directory where you installed RollerCoaster Tycoon 2.
 STR_6391    :Please select your RCT2 directory
 STR_6392    :Could not find {STRING} at this path.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -17,6 +17,7 @@
 - Fix: [#13044] Rides in RCT1 saves all have "0 customers per hour".
 - Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
 - Fix: [#13083] Dialog for renaming conflicting track design crops text out.
+- Fix: [#13125] Support localisation for RollerCoaster Tycoon 2 data picker dialogs
 - Fix: [#13129] Missing error message when waiting for train to leave station on the ride measurements graph.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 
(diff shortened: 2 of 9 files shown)


src/openrct2/Context.cpp
```diff
@@ -350,6 +350,24 @@ namespace OpenRCT2
...
+                _localisationService->OpenLanguage(gConfigGeneral.language);
+            }
+            catch (const std::exception& e)
+            {
+                log_error("Failed to open configured language: %s", e.what());
+                try
+                {
+                    _localisationService->OpenLanguage(LANGUAGE_ENGLISH_UK);
+                }
+                catch (const std::exception&)
```
janisozaur, line 364: Missing exception name here, causes to relog original exception

src/openrct2/config/Config.cpp
```diff
@@ -811,9 +815,8 @@ bool config_find_or_browse_install_directory()
                     return true;
                 }
 
-                std::string message = String::StdFormat(
-                    "Could not find %s" PATH_SEPARATOR "Data" PATH_SEPARATOR "g1.dat at this path", installPath.c_str());
-                uiContext->ShowMessageBox(message);
+                format_string(cantFindG1Dat, 1024, STR_COULD_NOT_FIND_AT_PATH, &g1DatPath);
```
IntelOrca, line 818: Change to the `std::string` variation of `format_string`, then you don't need the fixed length buffer.

IntelOrca:
Could you also update changelog.txt?

src/openrct2/Context.cpp
tupaschoal: (note it was already doing that, but since you're moving it, we can take the chance to fix it)

distribution/changelog.txt
```diff
@@ -17,6 +17,7 @@
 - Fix: [#13074] Entrance and exit ghosts for mazes not being removed.
 - Fix: [#13083] Dialog for renaming conflicting track design crops text out.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
+- Fix: [#13125] Support localisation for RollerCoaster Tycoon 2 data picker dialogs
```
AaronVanGeffen, line 20: We group by category, and then keep the list ordered by issue ID. Please move this line so the changelog retains it numerical order.

AaronVanGeffen:
I notice you have merged 'develop' into your branch. In this project, we have the convention of rebasing our branches before merging them, though. In this case, it's fine to squash-merge your PR, fortunately. Please keep it in mind in the future, though.

keithstellyes:
> I notice you have merged 'develop' into your branch. In this project, we have the convention of rebasing our branches before merging them, though. In this case, it's fine to squash-merge your PR, fortunately. Please keep it in mind in the future, though.

Just did a rebase

IntelOrca (approved):

data/language/en-GB.txt
```diff
@@ -3646,6 +3646,11 @@ STR_6387    :Can't lower element here…
 STR_6388    :Can't raise element here…
 STR_6389    :Invalid clearance
 
+# Picking RCT2 directory
```
Gymnasiast, line 3649: We don't typically put these kind of comments in the language file, to avoid making it too big. (Also note that many strings are scattered instead of grouped together, which is another reason adding comments won't really work well.)

Gymnasiast (commented):
I have one little line note to address, and I'd like to test it.

Gymnasiast (approved):
Appears to work correctly. The system language also gets picked up before this dialog is loaded, so it will appear in the user's native language.

Gymnasiast:
Thanks for the PR!
Do make sure you're using a branch next time instead of `develop` - you're way less likely to run into problems.

PR #13129, opened 2020-10-08
Fix missing error message on ride graph

duncanspumpkin:
Mistake made during implementation meant that the ride graph would not show the error message when the ride had not left the station. Another mistake meant that the tertiary colour of the ride window was not set to green. The tertiary colour was only used by this error message though so was not noticed.

image
New

rideold
Old


IntelOrca (approved):

tupaschoal (approved):
Needs a changelog entry

tupaschoal (approved):

PR #13131, opened 2020-10-08
Move formatter to seperate file. Rework GetMeasurement

duncanspumpkin:
Trying to get the tooltip functions to be a bit more sensible lead me off on a little detour.


tupaschoal (commented):
This looks fine, though I wonder how all other files are getting the `Formatter.h` include if they are not including it themselves, I guess it's all through the `Ride.h`

duncanspumpkin:
Yeah i was a bit disappointed that I didn't need to include anywhere. We really need to break up the big include files

tupaschoal:
Maybe define the OpenRCT2String in a tiny separate file? So you don't need to draw the Formatter into the headers just for it (as most of the functions will have `Formatter&` and thus will be forward declarable)

duncanspumpkin:
The OpenRCT2String takes a Formatter not a reference so can't be split from the file.

ZehMatt (approved):

PR #13132, opened 2020-10-08
Fix #7748. Move tooltip timeout reset to correct location

### LINKED ISSUE #7748: Tooltip timeout not working / counter intuitive code?
**OS:** Windows 10
**Version:** 0.2.0
**Commit/Build:** 1322410

game_handle_input_mouse -> input_widget_over -> input_update_tooltip has a section of code that seems to try to make tooltips disappear after being hovered over for a few seconds:

```
gTooltipTimeout += gTicksSinceLastUpdate;
        if (gTooltipTimeout >= 8000)
        {
            window_close_by_class(WC_TOOLTIP);
        }
```
However, right after the method is called in input_widget_over, gTooltipTimeout gets set back to 0 for some reason:

```
else
    {
        input_update_tooltip(w, widgetIndex, x, y);
    }

    gTooltipTimeout = 0;
    gTooltipCursorX = x;
    gTooltipCursorY = y;
```
As a result, gTooltipTimeout never reaches 8000 and the tooltips stay there forever. If I comment out `gTooltipTimeout = 0;` then it works and the tooltips disappear after a few seconds. Is there a reason for this line to be there?

As a note, I am particularly talking about ingame tooltips here. Tooltips on the title screen also don't go away, but apparently due to a method in TitleMenu.cpp:
```
static void window_title_menu_cursor(rct_window *w, rct_widgetindex widgetIndex, int32_t x, int32_t y, int32_t *cursorId)
{
    gTooltipTimeout = 2000;
}
```
I assume this is an intentional exception since there is a method specifically created for this purpose.

- [? ] Reproducible in RCT2 (vanilla)?

**Steps to reproduce:**
1. Load a scenario
2. Hover over a button or checkbox

duncanspumpkin:
Fix #7748. Move tooltip timeout reset to correct location

Mistake made in implementation

Not tested yet still a work in progress to see if this fixes the issues.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -8,6 +8,7 @@
 - Fix: [#4041] Garbled park option on scenario editor with custom theme.
 - Fix: [#5904] Empty errors on tile inspector base height change.
 - Fix: [#6086] Cannot install existing track design with another name.
+- Fix: [#7748] Tooltips would not timeout for normal UI elements.
 - Fix: [#8015] RCT2 files are not found when put into the OpenRCT2 folder.
 - Fix: [#8957] Error title missing when building with insufficient funds
 - Fix: [#10186] Placing multiple saved rides ignores design name (original bug).
--- a/src/openrct2-ui/input/MouseInput.cpp
+++ b/src/openrct2-ui/input/MouseInput.cpp
@@ -931,10 +931,6 @@ static void input_widget_over(const ScreenCoordsXY& screenCoords, rct_window* w,
     {
         input_update_tooltip(w, widgetIndex, screenCoords);
     }
-
-    gTooltipTimeout = 0;
-    gTooltipCursorX = screenCoords.x;
-    gTooltipCursorY = screenCoords.y;
 }
 
 /**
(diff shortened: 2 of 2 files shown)


tupaschoal (approved):
I tested and it works nicely.
Needs a changelog

duncanspumpkin:
So I looked into this quite a bit and the vanilla behaviour is quite different to what we have. I think this will do for now. In vanilla if you wait for a tooltip to show by hovering on a single UI element for a number of ticks it enters a tooltip showing mode where it does not wait any time for any UI tooltip to show. After you wait 8000 ticks on a single UI element it leaves tooltip show mode and you are back at the start. The difference we have is that we never enter this UI tooltip show mode we always have to wait a number of ticks between showing a tooltip. The number of ticks we have to wait was adjusted in OpenRCT2 compared to vanilla at some point.

pizza2004:
This sounds vaguely familiar.  I believe this was something that came up in a previous issue and was purposefully refactored because it wasn't exactly intuitive in vanilla.

Issue #12444, opened 2020-07-25
Refactor TILE_INSPECTOR_PAGE to use strong enum


ZehMatt:

autinerd:
This seems to need some more work, because it interferes with other enums

PR #13134, opened 2020-10-09
Tile inspector element type

amdoku:
should fix OpenRCT2#12444

introduced a convenience method to typesafe convert from the enum class to the underlying integer type, as the enum is used quite a lot as a lookup value for tables.


src/openrct2/windows/tile_inspector.h
```diff
@@ -12,20 +12,24 @@
...
+    Wall,
+    LargeScenery,
+    Banner,
+    Corrupt
 };
 
+std::underlying_type_t<TileInspectorPage> toIndex(TileInspectorPage page) {
+    return static_cast<std::underlying_type_t<TileInspectorPage>>(page);
+}
+
```
tupaschoal, line 32: We already have this in `src\openrct2\util\Util.h`:
https://github.com/OpenRCT2/OpenRCT2/blob/84c9ff833a778744c620cf7c4a56fa4eacfa2ac9/src/openrct2/util/Util.h#L84-L87

src/openrct2/interface/Window_internal.h
```diff
@@ -14,6 +14,8 @@
 #include <list>
 #include <memory>
 
+#include <src/openrct2/windows/tile_inspector.h>
+
```
tupaschoal, line 18:
```suggestion
enum class TileInspectorPage : int16_t;

```
Can you try forward declaring it, instead? If it works we avoid one include

tupaschoal (commented):
Can you please rebase and do two things:
- Drop the merge commit
- Fold the other commits into a single one, named:
`Close #12444: Refactor TILE_INSPECTOR_PAGE to use strong enum`

duncanspumpkin:
This would go in contrast to every other window. I don't think we should make this change.

src/openrct2/windows/tile_inspector.h
amdoku: Will use that one :+1:

src/openrct2/interface/Window_internal.h
amdoku: :+1:

amdoku:
This exploded quite a bit in size.
The culprit is that [this](https://github.com/OpenRCT2/OpenRCT2/blob/7c95f594d3f72d2d113490ec783e072418b4f1ef/src/openrct2/interface/Window_internal.h#L60) variable is accessed very often, and reused across different windows with different value ranges.

i.e. in [Cheats.cpp](https://github.com/OpenRCT2/OpenRCT2/blob/7c95f594d3f72d2d113490ec783e072418b4f1ef/src/openrct2-ui/windows/Cheats.cpp#L1008) it is used to get the value from a lookup table, which contains only four entries, TileInspectorPage containing 10.

This seems to be a reoccurring pattern in OpenRCT2, since there are a lot of union types already defined in Window_internal.h
My current approach is to replicate the union pattern for `page` as well.


I assume this is what @duncanspumpkin meant - please correct me if I am wrong.

duncanspumpkin:
The problem is that every window type has its own number of pages/tabs. If we use an enum class for each of them we will need a union of every window type page type. This would be annoying to use. Alternatively we could make a generic enum class for pages 1-X and have aliases that translate a page specific name to an enum class page number. But then it's not really giving us anything useful protection wise. 

I think we need to ask ourselves what are we trying to achieve and what does the final product look like. What do we want developers not to do and what do we want them to do. Off the top of my head we want to reduce macro usage to reduce name pollution. Reduce enum usage to reduce name pollution and allow for strong typing. Reduce code duplication. Prevent setting pages that don't exist. We have talked for a while about moving windows to individual classes and that will give us more tools to tackle this sort of thing but we still need to think about what the end goal should be.

amdoku:
Some thoughts:
- A blank pointer into a std::vector, which might reallocate, would result in a use after free bug, a fixed sized std::array would have the same storage layout and very similar API - but avoids the potential use after free.
- Any container approach would have to introduce polymorphism, otherwise you would not be able to store different window types in a container - I am not sure if this is something you would be opposed to (ignoring the `std::vector<std::variant<A,B,C,D,E,F> >` approach for now)
- Going down the road to remove macro usage and have the undesirable union type of all panes, might not be a bad way to go forward. All the logic for the panes is already in their own class, reusing the union member variable for individual state storage. First converting to strong enums would allow for all state sharing bugs to be found and/or fixed. Then a gradual transition on a per window basis is possible without having to worry about side effects in other windows.

This is from a first time contributor to this project, so I am very much aware that all of this has been talked about by people who have much more knowledge about this project already and I just could not find it.

duncanspumpkin:
Currently the big unions in window_internal are for accessing variables that should be private variables of the window if it was turned into class. In general writing to the page variable is done only from the specific window. Reading of the page value is also generally from the specific window but is at times also read in other locations. I'm not really suggesting anything here just writing down thoughts.

tupaschoal:
> I'm not really suggesting anything here just writing down thoughts.

Well we do need a consensus or a general direction at some point for the PR to be merged/closed ahahah

tupaschoal:
First of all, sorry for getting back to this after a while.

> Going down the road to remove macro usage and have the undesirable union type of all panes, might not be a bad way to go forward. All the logic for the panes is already in their own class, reusing the union member variable for individual state storage. First converting to strong enums would allow for all state sharing bugs to be found and/or fixed. Then a gradual transition on a per window basis is possible without having to worry about side effects in other windows.

I'm very convinced about that too. The union will look ugly, but decoupling will be nice when we have windows split up. I'm ok with merging this, are you @duncanspumpkin ?

(If so, we'll need merge conflicts to be resolved @amdoku )

duncanspumpkin:
okay but please keep in mind this is not the end goal

tupaschoal:
Decoupling is not the end goal? I thought ultimately we wanted a polymorphic window infra

duncanspumpkin:
I meant that the union we have isn't the end goal.

tupaschoal:
I know (?), for me it's a means to an end, we get a big ugly union but then we slowly move the corresponding variable to its actual window class.

amdoku:
No problem. Discussions take time.

I'll try to find time around new years eve.

tupaschoal (approved):

PR #13136, opened 2020-10-09
Fix #12413: VISIBILITY_CACHE strong enum change

### LINKED ISSUE #12413: Refactor VISIBILITY_CACHE to use strong enum

BryanDiLaura:
Fix for #12413. 

Let me know if there needs to be any changes!


tupaschoal (approved):

BryanDiLaura:
Ah that totally makes sense - will do from now on. Thanks so much!

Issue #12993, opened 2020-09-23
Remove hardcoded {SMALLFONT}{BLACK} from all tooltips


Gymnasiast:
In general, presentation should not be the responsibility of the strings. There are far more examples, but this one should be quite easy to do and make a nice difference.

duncanspumpkin:
How would you want to go about changing it?

Gymnasiast:
Not sure. I suppose you could render to a buffer and then to a `{SMALLFONT}{BLACK}{STRING}` string, but I'm sure there are other (and better) ways.

AaronVanGeffen:
Tooltips are all prepared in the same place, right? That means that we can indeed allocate a buffer of the right length when the tooltip is created. Then, add the code points for font and colour, and append the string in question. After that, we can just paint  the buffer.

gsckoco:
Are there any occurrences where tool tips use anything other than `{SMALLFONT}{BLACK}`? If not surely this could just be changed where you render the tooltip?

IntelOrca:
@gsckoco that is what this issue is suggesting.

I think I would do it how @Gymnasiast suggested. Use format string with a generic string ID that chains to the given tooltip string ID. Ideally these format strings would go directly in the code rather than being another resource string but we don't have that setup very easily at the moment.

duncanspumpkin:
String 1812 is small font format black string id  it can be used for this.

Lukasz-Pe:
I did it like @duncanspumpkin suggested.

PR #13137, opened 2020-10-09
Rework tooltip to be more logical

### LINKED ISSUE #13035: Modify rct_window_event_list::tooltip to return arguments
The `rct_window_event_list::tooltip` function currently returns a `string_id*` for its last parameter. It also passes back formatter arguments via the global variable `gCommonFormatArgs`. The `rct_window_event_list::tooltip` should instead return a `rct_string_id` and a `Formatter`. This reduces the usage of `gCommonFormatArgs` and its associated magic.

`std::pair<rct_string_id, Formatter> (*tooltip)(struct rct_window*, rct_widgetindex)`

duncanspumpkin:
Closes #13035. I'm not 100% convinced on the ride graph tooltips this adds them but they dont format correctly.

### DIFF
--- a/src/openrct2-ui/windows/GameBottomToolbar.cpp
+++ b/src/openrct2-ui/windows/GameBottomToolbar.cpp
@@ -68,7 +68,7 @@ static rct_widget window_game_bottom_toolbar_widgets[] =
 uint8_t gToolbarDirtyFlags;
 
 static void window_game_bottom_toolbar_mouseup(rct_window *w, rct_widgetindex widgetIndex);
-static void window_game_bottom_toolbar_tooltip(rct_window* w, rct_widgetindex widgetIndex, rct_string_id *stringId);
+static OpenRCT2String window_game_bottom_toolbar_tooltip(rct_window* w, const rct_widgetindex widgetIndex, const rct_string_id fallback);
 static void window_game_bottom_toolbar_invalidate(rct_window *w);
 static void window_game_bottom_toolbar_paint(rct_window *w, rct_drawpixelinfo *dpi);
 static void window_game_bottom_toolbar_update(rct_window* w);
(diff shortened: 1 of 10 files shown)


src/openrct2-ui/windows/Options.cpp
```diff
@@ -2107,21 +2107,23 @@ static void window_options_advanced_paint(rct_window* w, rct_drawpixelinfo* dpi)
     DrawTextEllipsised(dpi, screenCoords, 277, STR_STRING, ft, w->colours[1]);
 }
 
-static void window_options_advanced_tooltip(rct_window* w, rct_widgetindex widgetIndex, rct_string_id* stringid)
+static OpenRCT2String window_options_advanced_tooltip(rct_window* w, const rct_widgetindex widgetIndex, const rct_string_id fallback)
```
tupaschoal, line 2110:
```suggestion
static OpenRCT2String window_options_advanced_tooltip(
    rct_window* w, const rct_widgetindex widgetIndex, const rct_string_id fallback)
```

src/openrct2-ui/windows/GameBottomToolbar.cpp
```diff
@@ -189,10 +189,10 @@ static void window_game_bottom_toolbar_mouseup(rct_window* w, rct_widgetindex wi
     }
 }
 
-static void window_game_bottom_toolbar_tooltip(rct_window* w, rct_widgetindex widgetIndex, rct_string_id* stringId)
+static OpenRCT2String window_game_bottom_toolbar_tooltip(rct_window* w, const rct_widgetindex widgetIndex, const rct_string_id fallback)
```
tupaschoal, line 192:
```suggestion
static OpenRCT2String window_game_bottom_toolbar_tooltip(
    rct_window* w, const rct_widgetindex widgetIndex, const rct_string_id fallback)
```

tupaschoal (approved):
> Closes #13035. I'm not 100% convinced on the ride graph tooltips this adds them but they dont format correctly.

But had they never worked correctly or at all before? If so I'm in favour of merging this as is and opening a bug issue, if not then I guess we should tackle it here.

Changeset looks fine, though I do find it a bit weird that all methods receive a `fallback` and don't do anything with it other than returning when there is an error.

duncanspumpkin:
They had never worked before.

duncanspumpkin:
The fallback follows the style of the cursor event

tupaschoal:
Should we have a changelog entry for the graph tooltip or do we add it when it's fixed?

duncanspumpkin:
Add it when its fixed. I think doing https://github.com/OpenRCT2/OpenRCT2/issues/12993 will fix it.

PR #13140, opened 2020-10-09
Close #12408: Refactor CURSOR_ID to use strong enum

### LINKED ISSUE #12408: Refactor CURSOR_ID to use strong enum

ju-pinheiro:
Hi,

I have a question if that's ok :)

Some places were using the CURSOR_ID as int32_t, and others as uint8_t.
Since it had the UNDEFINED that equals to -1, I chose int8_t to represent it.
And then, to guarantee the uint8_t behavior in the places that expect it, I used some assertions.

Also, some data such as _legacyType.large_scenery.tool_id has the type uint8_t, and I was considering changing it to int8_t to avoid the need for one more static_cast. What do you think?

Thanks!

### DIFF
--- a/src/openrct2-ui/CursorData.cpp
+++ b/src/openrct2-ui/CursorData.cpp
@@ -11,9 +11,9 @@
 
 #include <openrct2/interface/Cursors.h>
 
-// clang-format off
 namespace OpenRCT2::Ui
 {
+    // clang-format off
     static constexpr const CursorData BlankCursorData =
     {
         { 0, 0 }, { 0 }, { 0 }
@@ -650,14 +650,14 @@ namespace OpenRCT2::Ui
         &HandClosedDownCursorData,  // CURSOR_HAND_CLOSED
     };
 
-    const CursorData * CursorRepository::GetCursorData(CURSOR_ID cursorId)
+    // clang-format on
+    const CursorData* CursorRepository::GetCursorData(CursorID cursorId)
     {
-        const CursorData * result = nullptr;
-        if (cursorId >= 0 && cursorId < CURSOR_COUNT)
+        const CursorData* result = nullptr;
+        if (cursorId != CursorID::Undefined && cursorId != CursorID::Count)
         {
-            result = RawCursorData[cursorId];
+            result = RawCursorData[EnumValue(cursorId)];
         }
         return result;
     }
-}
-// clang-format on
+} // namespace OpenRCT2::Ui
--- a/src/openrct2-ui/CursorRepository.h
+++ b/src/openrct2-ui/CursorRepository.h
@@ -13,6 +13,7 @@
 #include <functional>
 #include <map>
 #include <openrct2/interface/Cursors.h>
+#include <openrct2/util/Util.h>
 
 struct SDL_Cursor;
 
@@ -24,35 +25,35 @@ namespace OpenRCT2::Ui
         class CursorSetHolder
         {
         private:
-            SDL_Cursor* _cursors[CURSOR_COUNT] = { nullptr };
+            SDL_Cursor* _cursors[EnumValue(CursorID::Count)] = { nullptr };
 
         public:
-            CursorSetHolder(const std::function<SDL_Cursor*(CURSOR_ID)>& getCursor)
+            CursorSetHolder(const std::function<SDL_Cursor*(CursorID)>& getCursor)
             {
-                for (size_t i = 0; i < CURSOR_COUNT; i++)
+                for (size_t i = 0; i < EnumValue(CursorID::Count); i++)
                 {
-                    _cursors[i] = getCursor(static_cast<CURSOR_ID>(i));
+                    _cursors[i] = getCursor(static_cast<CursorID>(i));
                 }
             }
 
             ~CursorSetHolder()
             {
-                for (size_t i = 0; i < CURSOR_COUNT; i++)
+                for (size_t i = 0; i < EnumValue(CursorID::Count); i++)
                 {
                     SDL_FreeCursor(_cursors[i]);
                 }
             }
 
-            SDL_Cursor* getScaledCursor(CURSOR_ID cursorId)
+            SDL_Cursor* getScaledCursor(CursorID cursorId)
             {
-                return _cursors[cursorId];
+                return _cursors[EnumValue(cursorId)];
             }
         };
 
         constexpr static int32_t BASE_CURSOR_WIDTH = 32;
         constexpr static int32_t BASE_CURSOR_HEIGHT = 32;
 
-        CURSOR_ID _currentCursor = CURSOR_UNDEFINED;
+        CursorID _currentCursor = CursorID::Undefined;
         uint8_t _currentCursorScale = 1;
 
         std::map<uint8_t, CursorSetHolder> _scaledCursors;
@@ -60,13 +61,13 @@ namespace OpenRCT2::Ui
     public:
         ~CursorRepository();
         void LoadCursors();
-        CURSOR_ID GetCurrentCursor();
-        void SetCurrentCursor(CURSOR_ID cursorId);
+        CursorID GetCurrentCursor();
+        void SetCurrentCursor(CursorID cursorId);
         void SetCursorScale(uint8_t cursorScale);
 
     private:
         SDL_Cursor* Create(const CursorData* cursorInfo, uint8_t scale);
         void GenerateScaledCursorSetHolder(uint8_t scale);
-        static const CursorData* GetCursorData(CURSOR_ID cursorId);
+        static const CursorData* GetCursorData(CursorID cursorId);
     };
 } // namespace OpenRCT2::Ui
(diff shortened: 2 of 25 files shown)


src/openrct2/interface/Cursors.cpp
```diff
@@ -14,39 +14,40 @@
 
 namespace Cursor
 {
-    uint8_t FromString(const std::string& s, uint8_t defaultValue)
+    uint8_t FromString(const std::string& s, CursorID defaultValue)
```
duncanspumpkin, line 17: why isn't this returning a CursorID?

duncanspumpkin:
The int32_t's used in the windows are fine to change to whatever. The legacyType one is a bit more fiddly. Check how it is set I can't remember atm how it is read in the object loader but you might be able to change it to a CursorID.

src/openrct2/interface/Cursors.cpp
ju-pinheiro: At first, because this was always assigned to a _legacyType that was expecting a uint8_t.
That's why I asked about changing their type too.
Anyway, after refactoring the rct_small_scenery_entry::tool_id (and others) to CursorID type, this is resolved :)

ju-pinheiro:
Hi Duncan, changing the _legacyType.*tool_id to CursorID type worked. I gave the game a try, and the behavior seems the same.

src/openrct2/object/FootpathItemObject.cpp
```diff
@@ -25,7 +25,7 @@ void FootpathItemObject::ReadLegacy(IReadObjectContext* context, OpenRCT2::IStre
     stream->Seek(6, OpenRCT2::STREAM_SEEK_CURRENT);
     _legacyType.path_bit.flags = stream->ReadValue<uint16_t>();
     _legacyType.path_bit.draw_type = stream->ReadValue<uint8_t>();
-    _legacyType.path_bit.tool_id = stream->ReadValue<uint8_t>();
+    _legacyType.path_bit.tool_id = stream->ReadValue<CursorID>();
```
duncanspumpkin, line 28: can you make it so that all of the ReadValue calls that use CursorID stick to uint8_t. Wrap it in a static cast.
The reason we want to do this is so that if someone changes the type size we don't want to break object loading. In theory we would be validating the value that is read before converting it into a CursorID but...

src/openrct2-ui/input/MouseInput.cpp
```diff
@@ -1126,22 +1124,22 @@ void process_mouse_over(const ScreenCoordsXY& screenCoords)
                     ScreenCoordsXY scrollCoords;
                     widget_scroll_get_part(
                         window, &window->widgets[widgetId], screenCoords, scrollCoords, &output_scroll_area, &scroll_id);
-                    cursorId = scroll_id;
+                    cursorId = static_cast<CursorID>(scroll_id);
```
duncanspumpkin, line 1127: this line doesn't make any sense. Could you delete it. I think its just left over from the original.

duncanspumpkin (approved):

src/openrct2-ui/CursorData.cpp
```diff
@@ -650,12 +650,12 @@ namespace OpenRCT2::Ui
         &HandClosedDownCursorData,  // CURSOR_HAND_CLOSED
     };
 
-    const CursorData * CursorRepository::GetCursorData(CURSOR_ID cursorId)
+    const CursorData * CursorRepository::GetCursorData(CursorID cursorId)
```
AaronVanGeffen, line 653: This function is not following our code style. Because the entire namespace is wrapped between `// clang-format off` and `// clang-format on`, this has gone unnoticed.

I propose we move the clang-format toggle, starting just before `static constexpr const CursorData BlankCursorData =` and ending just before this function.

src/openrct2/interface/Cursors.h
```diff
@@ -11,42 +11,42 @@
 
 #include "../common.h"
 
-enum CURSOR_ID
```
Gymnasiast, line 14: Please change this to uint8_t, and define the Undefined cursor as 0xFF.

Your method is understandable and not wrong, but we almost always do it as I described above, so I prefer being consistent.

AaronVanGeffen:
This is looking good, thanks!

Unfortunately, you have added a merge commit to your branch. This is something we try to avoid. Please [rebase your branch](https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2) on top of current develop instead.

src/openrct2-ui/input/MouseInput.cpp
ju-pinheiro: It does not compile if I remove the cast. scroll_id's type is different from CursorID, isn't it?

duncanspumpkin: I meant delete the whole line. Scroll id has absolutely nothing to do with cursor id.

ju-pinheiro: oh ok, got it.

ju-pinheiro:
Hi Aaron, I'm trying to rebase as you mentioned and following the steps on the wiki page.
But I keep getting the following error after git rebase --continue:

[detached HEAD ce2b09669] Close #12408: Refactor CURSOR_ID to use strong enum
 25 files changed, 179 insertions(+), 172 deletions(-)
 rewrite src/openrct2/interface/Cursors.cpp (70%)
error: update_ref failed for ref 'refs/heads/cursor-id-strong-enum': cannot lock ref 'refs/heads/cursor-id-strong-enum': is at 1bbd551af2c2c401ec820744c27630a4bbf68839 but expected acd408436910a95e759ec6cec39d6e8d8f14f9e5
error: could not update refs/heads/cursor-id-strong-enum

So I typed git rebase --abort for now.
Any idea what's going on?

ju-pinheiro:
Hi Aaron, thanks for the git tip. It's rebased now :)

src/openrct2-ui/CursorRepository.h
```diff
@@ -24,49 +25,49 @@ namespace OpenRCT2::Ui
...
         private:
-            SDL_Cursor* _cursors[CURSOR_COUNT] = { nullptr };
+            SDL_Cursor* _cursors[EnumValue(CursorID::Count)] = { nullptr };
 
         public:
-            CursorSetHolder(const std::function<SDL_Cursor*(CURSOR_ID)>& getCursor)
+            CursorSetHolder(const std::function<SDL_Cursor*(CursorID)>& getCursor)
             {
-                for (size_t i = 0; i < CURSOR_COUNT; i++)
+                for (size_t i = 0; i < static_cast<size_t>(CursorID::Count); i++)
```
ZehMatt, line 33:
```suggestion
                for (size_t i = 0; i < EnumValue(CursorID::Count); i++)
```

src/openrct2-ui/CursorRepository.h
```diff
@@ -24,49 +25,49 @@ namespace OpenRCT2::Ui
...
                 {
-                    _cursors[i] = getCursor(static_cast<CURSOR_ID>(i));
+                    _cursors[i] = getCursor(static_cast<CursorID>(i));
                 }
             }
 
             ~CursorSetHolder()
             {
-                for (size_t i = 0; i < CURSOR_COUNT; i++)
+                for (size_t i = 0; i < static_cast<size_t>(CursorID::Count); i++)
```
ZehMatt, line 41:
```suggestion
                for (size_t i = 0; i < EnumValue(CursorID::Count); i++)
```

src/openrct2-ui/scripting/CustomMenu.h
```diff
@@ -46,7 +48,7 @@ namespace OpenRCT2::Scripting
     {
         std::shared_ptr<Plugin> Owner;
         std::string Id;
-        CURSOR_ID Cursor{};
+        CursorID Cursor{};
```
ZehMatt, line 51: This should be probably default initialize to Undefined and not the first value.

src/openrct2-ui/scripting/CustomMenu.cpp
```diff
@@ -20,15 +20,16 @@ namespace OpenRCT2::Scripting
...
         "arrow",         "blank",      "up_arrow",      "up_down_arrow", "hand_point", "zzz",         "diagonal_arrows",
         "picker",        "tree_down",  "fountain_down", "statue_down",   "bench_down", "cross_hair",  "bin_down",
         "lamppost_down", "fence_down", "flower_down",   "path_down",     "dig_down",   "water_down",  "house_down",
         "volcano_down",  "walk_down",  "paint_down",    "entrance_down", "hand_open",  "hand_closed",
     };
 
-    template<> DukValue ToDuk(duk_context* ctx, const CURSOR_ID& value)
+    template<> DukValue ToDuk(duk_context* ctx, const CursorID& cursorId)
     {
+        int8_t value = EnumValue(cursorId);
```
ZehMatt, line 32: Change the type to auto please, this avoids having type conflicts when the enum type changes.

src/openrct2-ui/scripting/CustomMenu.cpp
```diff
@@ -20,15 +20,16 @@ namespace OpenRCT2::Scripting
...
         if (value >= 0 && static_cast<size_t>(value) < std::size(CursorNames))
```
ZehMatt, line 33:
```suggestion
        if (EnumValue(value) < std::size(CursorNames))
```

ZehMatt (dismissed):

Gymnasiast (approved):

PR #13146, opened 2020-10-10
Close #8848 Refactor TimeInPark

### LINKED ISSUE #8848: Refactor Peep::TimeInPark for Staff/Guest.
For staff this field is used to track the employment date, for peeps its used to the track the time they have been in the park.

aw20368:
Split TimeInPark into a union: ParkEntryTime for guests, HireDate for staff.

Note: Duncan mentioned handling the rct2/rct1 structs in the issue. I updated only the import/export code and didn't change the fields of RCT2SpritePeep/rct1_peep, since I saw that previous refactors of Peep fields to unions (https://github.com/OpenRCT2/OpenRCT2/commit/e8d6383f and https://github.com/OpenRCT2/OpenRCT2/commit/0dd5b3bb)  did the same. If it's desired, I'll change all those fields as well.

### DIFF
--- a/src/openrct2-ui/windows/Guest.cpp
+++ b/src/openrct2-ui/windows/Guest.cpp
@@ -1423,10 +1423,11 @@ void window_guest_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     // Time in park
     screenCoords.y += LIST_ROW_HEIGHT + 1;
-    if (peep->TimeInPark != -1)
+    int32_t guestEntryTime = peep->GetParkEntryTime();
+    if (guestEntryTime != -1)
     {
         int32_t eax = gScenarioTicks;
-        eax -= peep->TimeInPark;
+        eax -= guestEntryTime;
         eax >>= 11;
         auto ft = Formatter();
         ft.Add<uint16_t>(eax & 0xFFFF);
@@ -1497,7 +1498,7 @@ void window_guest_rides_update(rct_window* w)
     }
 
     // Every 2048 ticks do a full window_invalidate
-    int32_t number_of_ticks = gScenarioTicks - guest->TimeInPark;
+    int32_t number_of_ticks = gScenarioTicks - guest->GetParkEntryTime();
     if (!(number_of_ticks & 0x7FF))
         w->Invalidate();
 
--- a/src/openrct2-ui/windows/Staff.cpp
+++ b/src/openrct2-ui/windows/Staff.cpp
@@ -1102,7 +1102,7 @@ void window_staff_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
     }
 
     auto ft = Formatter();
-    ft.Add<int32_t>(peep->TimeInPark);
+    ft.Add<int32_t>(peep->GetHireDate());
     gfx_draw_string_left(dpi, STR_STAFF_STAT_EMPLOYED_FOR, ft.Data(), COLOUR_BLACK, screenCoords);
     screenCoords.y += LIST_ROW_HEIGHT;
 
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -983,7 +983,7 @@ void Guest::Tick128UpdateGuest(int32_t index)
         if (State == PeepState::Walking && !OutsideOfPark && !(PeepFlags & PEEP_FLAGS_LEAVING_PARK) && GuestNumRides == 0
             && GuestHeadingToRideId == RIDE_ID_NULL)
         {
-            uint32_t time_duration = gScenarioTicks - TimeInPark;
+            uint32_t time_duration = gScenarioTicks - ParkEntryTime;
             time_duration /= 2048;
 
             if (time_duration >= 5)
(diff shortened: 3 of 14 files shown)


duncanspumpkin:
I would still like to see the field renamed in the rct2/rct1 structs. No need to add the union or change anything else just would like the field renamed as its clear that its name is not correct.

aw20368:
Makes sense....updated the structs.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1423,10 +1423,11 @@ void window_guest_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     // Time in park
     screenCoords.y += LIST_ROW_HEIGHT + 1;
-    if (peep->TimeInPark != -1)
+    int32_t guestEntryTime = peep->AsGuest()->GetParkEntryTime();
```
duncanspumpkin, line 1426: think this is already a guest you don't need to use AsGuest

src/openrct2-ui/windows/Staff.cpp
```diff
@@ -1102,7 +1102,7 @@ void window_staff_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
     }
 
     auto ft = Formatter();
-    ft.Add<int32_t>(peep->TimeInPark);
+    ft.Add<int32_t>(peep->AsStaff()->GetHireDate());
```
duncanspumpkin, line 1105: same for this pretty sure this is already a staff

aw20368:
You're right, AsGuest/AsStaff weren't needed in both places. I found another (in Staff::ResetStats) that I fixed as well.

duncanspumpkin (approved):

src/openrct2/peep/Guest.cpp
```diff
@@ -2399,6 +2399,16 @@ bool Guest::HasRiddenRideType(int32_t rideType) const
...
 }
 
+void Guest::SetParkEntryTime(int32_t entryTime)
+{
+    ParkEntryTime = entryTime;
+}
+
+int32_t Guest::GetParkEntryTime() const
+{
+    return ParkEntryTime;
```
Gymnasiast, line 2409: We never use "passthrough getters" for fields that are public.

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1489,7 +1489,8 @@ class S4Importer final : public IParkImporter
         dst->Id = src->id;
         dst->CashInPocket = src->cash_in_pocket;
         dst->CashSpent = src->cash_spent;
-        dst->TimeInPark = src->time_in_park;
+        // This doubles as staff hire date
+        dst->ParkEntryTime = src->park_entry_time;
```
Gymnasiast, line 1493: I'd find it more consistent of this also used a `union`.

@duncanspumpkin would that be OK with you?

src/openrct2/peep/Guest.cpp
aw20368: The getter was requested in the issue, I'll remove it if we confirm that it's not wanted. I assume the "setter" should also be removed in that case?

src/openrct2/rct1/S4Importer.cpp
duncanspumpkin: I don't like to modify the rct1/2 structs much except to update the naming. If you feel it would be useful then go ahead. I think eventually I'm going to want to split the import code into a staff and a guest import so it would help with that.

src/openrct2/peep/Guest.cpp
duncanspumpkin: @ZehMatt suggested including it. I'm okay with it staying

ZehMatt: Its usually better to have getters so we can change the underlying data at any point without compatibility issues.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1423,10 +1423,11 @@ void window_guest_stats_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     // Time in park
     screenCoords.y += LIST_ROW_HEIGHT + 1;
-    if (peep->TimeInPark != -1)
+    int32_t guestEntryTime = peep->GetParkEntryTime();
+    if (guestEntryTime != -1)
     {
         int32_t eax = gScenarioTicks;
```
ZehMatt, line 1429: While you are at it, could you rename this variable to something more sensible such as ```timeInPark```

src/openrct2/peep/Guest.cpp
tupaschoal: > Its usually better to have getters so we can change the underlying data at any point without compatibility issues.

I second that

ZehMatt (approved):
We can change the variable name later, PR looks good to me.

src/openrct2/peep/Guest.cpp
Gymnasiast: Can't say I agree, but I'm not going to overrule three other team members for something like this.

src/openrct2-ui/windows/Guest.cpp
tupaschoal: Please do this in a follow-up PR, so we don't forget :D

src/openrct2/rct1/S4Importer.cpp
aw20368: There's a few more union fields in Peep that are not union in the rct1/2 structs (eg voucher_ride_id, paid_on_souvenirs, paid_on_food)...should I make them all union? I can also split the import/export if we want to do so now, or are we waiting for the new save format.

Issue #13148, opened 2020-10-10
-Wold-enable-old-style-cast has broken GCC-9 builds (probably others)


keithstellyes:
There's a few errors like

```
error: use of old-style cast to ‘struct paint_struct*’ [-Werror=old-style-cast]
   43 | [i].PaintStructs[j].basic.next_quadrant_ps == (paint_struct*)paint_struct_entries)
      |                                                              ^~~~~~~~~~~~~~~~~~~~
      |                                               ---------------
      |                                               reinterpret_cast<paint_struct*> (  )
```

Introduced by `f418acfe1951749d31754553c52c5904a19ba3e1`, I can make a PR real quick fixing that, but I'm surprised this made it past CI?

duncanspumpkin:
Not sure why the CI isn't failing on this. Can you PR the reinterpret_cast

keithstellyes:
Will do real quick

Krutonium:
FWIW It's also broken for me on GCC 10.2.0

AaronVanGeffen:
Compiles fine for me on GCC 10.2.0, oddly enough. But good to get rid of these remaining old-style casts, anyway, of course.

AaronVanGeffen:
Some more breakage in `macos.mm`, which I'll file a PR for shortly.

keithstellyes:
@AaronVanGeffen Just as an FYI, I do have a PR for some of them here: https://github.com/OpenRCT2/OpenRCT2/pull/13150

AaronVanGeffen:
@keithstellyes Oddly enough, my compiler wasn't complaining about that file at all -- just macos.mm

keithstellyes:
Closing since my PR was closed

PR #13150, opened 2020-10-10
Fix old-style-cast build errors on BenchSpriteSort.cpp ( #13148 )

keithstellyes:
The turning on of `-Wold-style-casts` still had a few lines that it broke on, at least for GCC-9 on Linux, this fixes it

### DIFF
--- a/src/openrct2/cmdline/BenchSpriteSort.cpp
+++ b/src/openrct2/cmdline/BenchSpriteSort.cpp
@@ -40,7 +40,7 @@ static void fixup_pointers(paint_session* s, size_t paint_session_entries, size_
     {
         for (size_t j = 0; j < paint_struct_entries; j++)
         {
-            if (s[i].PaintStructs[j].basic.next_quadrant_ps == (paint_struct*)paint_struct_entries)
+            if (s[i].PaintStructs[j].basic.next_quadrant_ps == reinterpret_cast<paint_struct*>(paint_struct_entries))
             {
                 s[i].PaintStructs[j].basic.next_quadrant_ps = nullptr;
             }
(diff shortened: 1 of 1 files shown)


src/openrct2/cmdline/BenchSpriteSort.cpp
```diff
@@ -52,13 +52,13 @@ static void fixup_pointers(paint_session* s, size_t paint_session_entries, size_
...
         {
-            if (s[i].Quadrants[j] == (paint_struct*)quadrant_entries)
+            if (s[i].Quadrants[j] == reinterpret_cast<paint_struct*>(quadrant_entries))
             {
                 s[i].Quadrants[j] = nullptr;
             }
             else
             {
-                s[i].Quadrants[j] = &s[i].PaintStructs[(size_t)s[i].Quadrants[j]].basic;
+                s[i].Quadrants[j] = &s[i].PaintStructs[reinterpret_cast<size_t>(s[i].Quadrants[j])].basic;
```
duncanspumpkin, line 61: Think this can be a static_cast

keithstellyes: That actually gives a compile error

```
error: invalid static_cast from type ‘paint_struct*’ to type ‘size_t’ {aka ‘long unsigned int’}
   61 |                 s[i].Quadrants[j] = &s[i].PaintStructs[static_cast<size_t>(s[i].Quadrants[j])].basic;
```

It seems really odd to me that we just cast a big fat struct to an int, wouldn't it be better to just be using `image_id`? e.g.

`s[i].Quadrants[j] = &s[i].PaintStructs[s[i].Quadrants[j]->image_id].basic;`

I mean even here doing that weirdness of really just getting first struct member by casting:

https://github.com/OpenRCT2/OpenRCT2/blob/b628bba704b3b4dd0f5a1b548a4e718504259859/src/openrct2/cmdline/BenchSpriteSort.cpp#L49

Seems to me that also really should be 

`auto nextQuadrantPs = s[i].PaintStructs[j].basic.next_quadrant_ps->image_id;`

duncanspumpkin: Yeah this code is very odd thanks for looking into it properly. @janisozaur I can't understand what is trying to be achieved here. Has `s[i].Quadrants[j]` been loaded with offsets and this is converting them to pointers?

janisozaur: Yes, that's correct. They're converted between offsets and pointers

duncanspumpkin: Where is this loading code? I tried to find it earlier but couldn't work out how it ends up with offsets.

Gymnasiast: @duncanspumpkin Is it okay to merge this PR as-is, then? I imagine changing the offset/pointer conversion is a PR in its own right.

duncanspumpkin (approved):

Issue #13102, opened 2020-10-05
Underflow on height chart (Ride measurements)


Gymnasiast:
**OS:** Ubuntu 20.04
**Version:** 0.3.1+
**Commit/Build:** a284119990bb94fad7aff97ee2cebccf511556b8

![afbeelding](https://user-images.githubusercontent.com/1478678/95118040-e15e6900-0749-11eb-8ba1-112c20099cc4.png)

iking96:
I can try to look into this one this weekend! 📉

PR #13152, opened 2020-10-10
[Issue 13102] Update Localisation.format_string_code FORMAT_LENGTH to be signed

iking96:
## Description

Update cast of value input to `Localisation.format_string_code` from unsigned to signed. Addresses https://github.com/OpenRCT2/OpenRCT2/issues/13102

### 📷 Pic of issue
 afbeelding

### 📸 Pic of fix
Screen Shot 2020-10-10 at 2 01 46 PM

### Note
~This change has the potential to affect all displays which use the "length" type. They will not have a max signed value of 32,767.~ (Outdated)


iking96:
@Gymnasiast I believe this should address the issue.

distribution/changelog.txt
```diff
@@ -5,6 +5,7 @@
 - Feature: [#13000] objective_options command for console.
 - Feature: [#13096] Add Esperanto translation.
 - Change: [#9568] Change lift sounds of Reverser Roller Coaster and Compact Inverted Coaster to better fitting ones.
+- Fix: [#13102] Correct underflow when displaying negative length.
```
AaronVanGeffen, line 8: We keep the changelog ordered in numerical order, grouped by category. Could you move this line down, so the order is retained?

iking96: My bad, will fix.

Gymnasiast (commented):
I'd like to check if this fix does not cause https://github.com/OpenRCT2/OpenRCT2/issues/1013 to resurface. I'm afraid it might, because https://github.com/OpenRCT2/OpenRCT2/pull/11651 essentially did the opposite of this PR.

iking96:
> #11651

> I'd like to check if this fix does not cause #1013 to resurface. I'm afraid it might, because #11651 essentially did the opposite of this PR.

Yes, I see what your saying and think this will likely cause that issue to resurface. So it seem like we'd need more than an `int16_t` to avoid both issues.

iking96:
Hey @Gymnasiast,

Since we stuck using 16 bits for the value associated with "length" strings for now (pending a more substantial refactor), I added a new string ID for a signed int. What do you think of this.

For now I have made only the altitude stat use it, so it should not cause this issue to resurface; https://github.com/OpenRCT2/OpenRCT2/issues/1013.

For other localizations to work I believe this PR will also be needed: https://github.com/OpenRCT2/Localisation/pull/1964

duncanspumpkin:
Would it really be a substantial refactor? Perhaps we should look into properly handling this taking it up to int32_t.

src/openrct2/localisation/Localisation.cpp
```diff
@@ -1138,6 +1138,13 @@ static void format_string_code(uint32_t format_code, char** dest, size_t* size,
 
             format_length(dest, size, static_cast<uint16_t>(value));
             break;
+        case FORMAT_LENGTH_SIGNED:
+            // Pop argument
+            std::memcpy(&value, *args, sizeof(int16_t));
+            *args += 2;
+
+            format_length(dest, size, static_cast<int16_t>(value));
+            break;
```
AaronVanGeffen, line 1147: To be honest, I am not a fan of introducing a new format code for this. Can't length always be signed instead? If really necessary, we can increase it to a 32-bits value.

IntelOrca: I agree, as I am planning on refactoring out the arg buffer which eliminates strict sizes altogether. This will mean that in the language strings, we will be left with `{COMMA}`, `{INT}`, `{LENGTH}`.

iking96: Thank you all for the feedback. I have ben having fun looking into this project.

@AaronVanGeffen @duncanspumpkin 

I see where you are coming from and agree. I think to extend the length format to 32-bits I'd need to find all the places which ultimately use the length formatter and ensure they are sending in a pointer to a 32-bit value. Such as [here](https://github.com/OpenRCT2/OpenRCT2/blob/0bf851cfb290636969a95b9d84528962c536df82/src/openrct2-ui/windows/Ride.cpp#L6006-L6011). Is that your understanding as well?

@IntelOrca 

That sounds much more usable. Was that something you were planning to tackle relatively soon or just something that is planned for future work? If it is already in the works, perhaps it'd be best for me to just let your refactor resolve this issue.

IntelOrca: I am currently developing it. I wouldn't let that put you off from finding every occurrence that length is used and replacing the int16 with an int32. I doubt the are too many usages of LENGTH.

PR #13159, opened 2020-10-11
Tooltip 2

duncanspumpkin:
A couple of little refactors to the tooltips to make it a bit simpler.

### DIFF
--- a/src/openrct2-ui/windows/EditorInventionsList.cpp
+++ b/src/openrct2-ui/windows/EditorInventionsList.cpp
@@ -686,9 +686,6 @@ static void window_editor_inventions_list_scrollpaint(rct_window* w, rct_drawpix
 static void window_editor_inventions_list_drag_open(ResearchItem* researchItem)
 {
     char buffer[256], *ptr;
-    int32_t stringWidth;
-    rct_window* w;
-
     window_close_by_class(WC_EDITOR_INVENTION_LIST_DRAG);
     _editorInventionsListDraggedItem = *researchItem;
     rct_string_id stringId = researchItem->GetName();
@@ -710,13 +707,12 @@ static void window_editor_inventions_list_drag_open(ResearchItem* researchItem)
         format_string(ptr, 256, stringId, nullptr);
     }
 
-    stringWidth = gfx_get_string_width(buffer);
+    auto stringWidth = gfx_get_string_width(buffer);
     window_editor_inventions_list_drag_widgets[0].right = stringWidth;
 
-    w = window_create(
-        ScreenCoordsXY(gTooltipCursor.x - (stringWidth / 2), gTooltipCursor.y - 7), stringWidth, 14,
-        &window_editor_inventions_list_drag_events, WC_EDITOR_INVENTION_LIST_DRAG,
-        WF_STICK_TO_FRONT | WF_TRANSPARENT | WF_NO_SNAPPING);
+    auto* w = window_create(
+        gTooltipCursor - ScreenCoordsXY{ stringWidth / 2, 7 }, stringWidth, 14, &window_editor_inventions_list_drag_events,
+        WC_EDITOR_INVENTION_LIST_DRAG, WF_STICK_TO_FRONT | WF_TRANSPARENT | WF_NO_SNAPPING);
     w->widgets = window_editor_inventions_list_drag_widgets;
     w->colours[1] = COLOUR_WHITE;
     input_window_position_begin(w, 0, gTooltipCursor);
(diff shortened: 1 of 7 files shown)


duncanspumpkin:
This should fix the introduced issues with the tooltips now. I did a search for all tooltips that had parameters and ensured they were taken care of so I think that has caught all of the issues.

src/openrct2-ui/windows/EditorInventionsList.cpp
```diff
@@ -714,12 +714,12 @@ static void window_editor_inventions_list_drag_open(ResearchItem* researchItem)
     window_editor_inventions_list_drag_widgets[0].right = stringWidth;
 
     w = window_create(
-        ScreenCoordsXY(gTooltipCursorX - (stringWidth / 2), gTooltipCursorY - 7), stringWidth, 14,
+        ScreenCoordsXY(gTooltipCursor.x - (stringWidth / 2), gTooltipCursor.y - 7), stringWidth, 14,
```
tupaschoal, line 717:
```suggestion
        gTooltipCursor - ScreenCoordsXY{ stringWidth / 2, 7 }, stringWidth, 14,
```

src/openrct2-ui/windows/Tooltip.cpp
```diff
@@ -47,30 +47,35 @@ void window_tooltip_reset(const ScreenCoordsXY& screenCoords)
...
+    rct_window* w;
+    int32_t width, height;
+
+    w = window_find_by_class(WC_ERROR);
+    if (w != nullptr)
+        return;
+
+    int32_t textWidth = FormatTextForTooltip(message);
+    width = textWidth + 3;
+    height = ((_tooltipNumLines + 1) * font_get_line_height(FONT_SPRITE_BASE_MEDIUM)) + 4;
```
tupaschoal:
```suggestion
    auto* w = window_find_by_class(WC_ERROR);
    if (w != nullptr)
        return;

    int32_t textWidth = FormatTextForTooltip(message);
    int32_t width = textWidth + 3;
    int32_t height = ((_tooltipNumLines + 1) * font_get_line_height(FONT_SPRITE_BASE_MEDIUM)) + 4;
```

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2777,6 +2779,58 @@ static void window_ride_vehicle_update(rct_window* w)
...
+        return { STR_NONE, {} };
+
+    switch (widgetIndex)
+    {
+        case WIDX_VEHICLE_TRAINS:
+        case WIDX_VEHICLE_TRAINS_DECREASE:
+        case WIDX_VEHICLE_TRAINS_INCREASE:
+        {
+            auto ft = Formatter();
+            ft.Increment(12);
```
tupaschoal, line 2796: What's up with these increments that didn't exit before? There isn't any data prior nor any rewind later, so they seems a bit confusing (to me, at least)

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2777,6 +2779,58 @@ static void window_ride_vehicle_update(rct_window* w)
...
+            rct_string_id stringId = RideComponentNames[vehicleType].count;
+            if (ride->max_trains > 1)
+            {
+                stringId = RideComponentNames[vehicleType].count_plural;
+            }
+            ft.Add<rct_string_id>(stringId);
+            ft.Add<uint16_t>(ride->max_trains);
+            return { fallback, ft };
+        }
+        break;
```
tupaschoal, line 2808:
```suggestion
```

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2777,6 +2779,58 @@ static void window_ride_vehicle_update(rct_window* w)
...
+
+            rct_string_id stringId = RideComponentNames[RIDE_COMPONENT_TYPE_CAR].singular;
+            if ((ride->min_max_cars_per_train & 0xF) - rideEntry->zero_cars > 1)
+            {
+                stringId = RideComponentNames[RIDE_COMPONENT_TYPE_CAR].plural;
+            }
+            ft.Add<rct_string_id>(stringId);
+            return { fallback, ft };
+        }
+        break;
```
tupaschoal, line 2829:
```suggestion
```

tupaschoal:
I've fixed the build failure myself, but there are still some comments and suggestions there.

src/openrct2-ui/windows/Ride.cpp
duncanspumpkin: The string id associated with this needs the args in those positions. If you look at the chain of adds that this was removed from you will see where the number comes from

Issue #13147, opened 2020-10-10
PNG sprite importer does not support 4-bit colour maps


AaronVanGeffen:
**OS:** Arch Linux
**Version:** 0.3.1+
**Commit/Build:** 

The in-game PNG sprite importer, used to generate `g2.dat`, does not appear to support 4-bit colour maps.

Before recompressing the RMC/Hybrid Coaster PNGs (8-bit colours maps):
![Amity Airfield 2020-10-10 16-31-43](https://user-images.githubusercontent.com/604665/95658227-13404880-0b19-11eb-8010-be376fa88bb8.png)

Before recompressing the RMC/Hybrid Coaster PNGs (4-bit colours maps):
![Amity Airfield 2020-10-10 16-52-29](https://user-images.githubusercontent.com/604665/95658229-13d8df00-0b19-11eb-8126-ae7f3874deba.png)

PR #13162, opened 2020-10-11
Resources compression

ppamorim:
This PR applies a bunch of non lossy compression to the sprites inside the `/resources` folder. The script applied is based on the `oxipng` inside the `ImageOptim` application (By running `oxipng` standalone didn't compress much more, less than 10 bytes of difference). With this setup I was able to reduce the folder from 3073kb to 1708kb:

Screenshot 2020-10-11 at 12 46 38

You should not be able to see any difference between the old sprites and the new ones because the compression applied is not lossy.


AaronVanGeffen:
Are you sure there aren't any differences? Given #13147

I am not sure it's worth compressing these images at this stage. It's worth doing when a PR hasn't been merged yet, but at this point, the objects are already in the repository. Worse yet, recompressing would just increasing the repo size further, as new objects are needed for the recompressed objects.

ppamorim:
@AaronVanGeffen The sprites are using 8-bit colormap. I think there is a way to replace these files in older commits.

AaronVanGeffen:
Yes, but force pushing is not an option for our repository. All commit hashes would change, which is quite destructive for historical cross-references.

X123M3-256:
@AaronVanGeffen It wouldn't decrease the repo size but it would decrease the download size, no? I think most users aren't building from source.

AaronVanGeffen:
@X123M3-256 It wouldn't matter for download size, as the PNGs are RLE compressed instead (iirc) when processed into `g2.dat`.

X123M3-256:
@AaronVanGeffen Oh yeah, they are. I am an idiot.

Gymnasiast:
Given that this:
- Increases the repo size
- Does not reduce the download size\*

I won't merge it.

But I do think it does provide good insight on the different PNG optimisers, and it will be useful when deciding on which optimiser to use when the next PR that adds PNGs comes around. Last time, we tried `optipng` and `zopflipng` - the latter compressed better, but in a way that was incompatible with our PNG handling code. Next time, we can also compare `oxipng`, as it appears to compress better than optipng while still being compatible with our PNG handling code.

Gymnasiast:
I have used `oxipng` on https://github.com/OpenRCT2/objects/pull/135 . It reduced the size of the images from 3.5 MB to 1.1 MB. (Small caveat: I have not yet tested if they get imported correctly.)

Issue #13156, opened 2020-10-11
Headless GNU/Linux install fails to find objects


Berbe:
**OS:** GNU/Linux Debian equivalent
**Version:** 0.3.1
**Commit/Build:** d01dcea

Following [Wiki GNU/Linux build instructions](../wiki/Building-OpenRCT2-on-Linux#running-it-from-the-build-directory), you can either:
- [Run `openrct-cli` from the `build` directory](../wiki/Building-OpenRCT2-on-Linux#running-it-from-the-build-directory)
- [Run an installed `openrct-cli`](../wiki/Building-OpenRCT2-on-Linux#make-install)

While the former works, with:
```
openrct2/build# ls -l data/
total 128
-rw-r--r-- 1 root root 116634 Oct 11 02:04 g2.dat <-- obtained from `build` directory
drwxr-xr-x 2 root root   4096 Oct 11 01:08 language
drwxr-xr-x 7 root root   4096 Oct 11 02:05 object <-- obtained from [latest build](https://github.com/OpenRCT2/objects/releases)
drwxr-xr-x 2 root root   4096 Oct 11 01:08 shaders
```

The latter *does not* work, with:
```
openrct2/build# ls -l /usr/local/share/
total 28
drwxr-xr-x 3 root root 4096 Oct 11 02:36 doc
drwxr-xr-x 3 root root 4096 Oct 11 02:36 man
drwxr-xr-x 2 root root 4096 Oct 11 02:36 metainfo
drwxr-xr-x 3 root root 4096 Oct 11 02:36 mime
drwxr-xr-x 7 root root 4096 Oct 11 02:36 object
drwxr-xr-x 4 root root 4096 Oct 11 02:36 openrct2
drwxr-xr-x 2 root root 4096 Oct 11 02:36 sequence
```

It yields:
```
openrct2-cli saves/Crazy\ Castle.SC6 --headless
[BALLN   ] Object not found.
[BURGB   ] Object not found.
[CNDYF   ] Object not found.
[ATM1    ] Object not found.
[ARRT1   ] Object not found.
[DRNKS   ] Object not found.
[FAID1   ] Object not found.
[INFOK   ] Object not found.
[ZLDB    ] Object not found.
[LFB1    ] Object not found.
[...]
openrct2 $
```

The `object` & `sequence` directories seem ill-located. In fact, moving the `object` directory from `/usr/local/share/` to `/usr/local/share/openrct2/` yields:
```
[...]

PR #13163, opened 2020-10-11
Fix: CMake project

Berbe:
Fix how path are computed from project name/variables in CMake.

This allows installing resources (sequences & objects) at their proper location when built without GUI.

Fixes: #13156

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,7 +306,9 @@ add_custom_command(
     COMMAND ./openrct2-cli sprite build \"${CMAKE_BINARY_DIR}/g2.dat\" \"${ROOT_DIR}/resources/g2/sprites.json\"
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
 )
-add_custom_target(g2 DEPENDS ${PROJECT} g2.dat)
+add_custom_target(g2 DEPENDS ${PROJECT_NAME} g2.dat)
+
+project(openrct2 CXX)
 
 # Include tests
 if (WITH_TESTS)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -306,7 +306,9 @@ add_custom_command(
     COMMAND ./openrct2-cli sprite build \"${CMAKE_BINARY_DIR}/g2.dat\" \"${ROOT_DIR}/resources/g2/sprites.json\"
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
 )
-add_custom_target(g2 DEPENDS ${PROJECT} g2.dat)
+add_custom_target(g2 DEPENDS ${PROJECT_NAME} g2.dat)
+
+project(openrct2 CXX)
```
janisozaur, line 311: this already exists on line 22

janisozaur (changes requested):
I don't see what/how this fixes the issue. The only change other than `${PROJECT}` -> `${PROJECT_NAME}` is the one I find invalid. Please provide justification.

CMakeLists.txt
Berbe: `PROJECT` is not defined in this file; it inherits it from included `CMakeLists` (odd code organisation... but not my doing). It is specifically defined in the included `src/openrct2-ui/CMakeLists.txt`.
What happens when this file is not included? `PROJECT` is empty and that wreaks havoc later; please see the linked issue.

This change makes the `g2` build depend on the last `project` name: if you build with GUI, it should depend on `openrct2`, exactly as before.
Without GUI, it will depend on `openrct2-cli` (the last included sub-`CMakeLists.txt`), which did not seem to yield any problem on my side.

This is a quick fix to what looks like a broader mess of inter-dependencies, but I am no CMake expert, and this needed rework might be seen as broader than the scope of the issue at hand.

janisozaur (approved):

janisozaur:
Sorry it took so long, took a break from GitHub.

PR #13172, opened 2020-10-12
Part of #12307: GOG installer extraction for POSIX

### LINKED ISSUE #12307: Make retrieving assets through Steam and GOG easier
This is mostly an issue for users on Mac and Linux. The amount of questions we get about this illustrates that this too difficult for many users currently.

Since the actual install mostly involves running a few commands, it would be nice if OpenRCT2 could run these for the user.

When OpenRCT2 starts up and doesn't have any assets yet, it could, for example work like this:

```
OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work.

Select which option applies to yet:
1. I already have RCT2 installed on my computer.
2. I don't have RCT2 installed, but I own it on Steam.
3. I don't have RCT2 installed, but I own it on GOG.
4. I don't own RCT2 yet.
```
Option 1 will lead to the current dialog box where the user selects their directory.
Option 4 will display information on how and where to get RCT2.

Option 2 could do two things:
- only use Steam itself, starting it up with the Console tab and giving users instructions (easy to implement)
- use a script to actually download RCT2 for the user (more difficult to implement, but much more user friendly).

Option 3 would work similar to option 2:
- the simplest option would be to ask the user to download the full install and tell OpenRCT2 where it is, then use `innoextract` to extract it.
- the more user-friendly option would be to use `lgogdownloader` to download the game for the user.

keithstellyes:
# Overview
I tried to be more minimal with these changes, however a menu-selection didn't already exist, so this required implementation. So far, it is only implemented in Linux. As a result, this flow is only enabled on Linux. Mac support shouldn't be too hard to add in and it would be useful for Mac too. 

I do not feel strongly about wording on options, happy to hear thoughts, better wordings, etc.

`zenity` and `kdialog` don't actually support any kind of reasonable automatic width or height, so this approximates it.

* Menu selection (Linux only)
* Move UiContext's logic for checking if an app executes up to Platform, logic should be same across POSIX.
* Move UiContext's Execute logic up to Platform
* Extract GOG installer on behalf of user. The installer data is extracted to `OPEN_RCT2_CONFIG/rct2/`

`kdialog` (KDE)
Screenshot from 2020-10-11 20-21-51

`zenity` (GNOME)
Screenshot from 2020-10-11 20-22-20

If Linux user has neither `kdialog` or `zenity`, then it simply falls back to the non-Linux behavior

# Related future work

* Mac, other BSD's
* Similar option but for Steam as documented in #12307

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3651,6 +3651,15 @@ STR_6396    :Disable screensaver and monitor power saving
 STR_6397    :If checked, screensaver and other monitor power saving features will be inhibited while OpenRCT2 is running.
 STR_6398    :File contains unsupported ride types. Please update to a newer version of OpenRCT2.
 STR_6399    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work. Please set the “game_path” variable in config.ini to the directory where you installed RollerCoaster Tycoon 2, then restart OpenRCT2.
+STR_6400    :I have the GOG installer for RollerCoaster Tycoon 2 downloaded, but it is not installed
+STR_6401    :I have RollerCoaster Tycoon 2 installed already
+STR_6402    :OpenRCT2 Data Setup
+STR_6403    :Select which applies best to you
+STR_6404    :Please select GOG RollerCoaster Tycoon 2 installer
+STR_6405    :Select GOG Installer
+STR_6406    :GOG RollerCoaster Tycoon 2 Installer
+STR_6407    :This may take a few minutes
+STR_6408    :Please install innoextract to extract GOG Installer
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3651,15 +3651,15 @@ STR_6396    :Disable screensaver and monitor power saving
 STR_6397    :If checked, screensaver and other monitor power saving features will be inhibited while OpenRCT2 is running.
 STR_6398    :File contains unsupported ride types. Please update to a newer version of OpenRCT2.
 STR_6399    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work. Please set the “game_path” variable in config.ini to the directory where you installed RollerCoaster Tycoon 2, then restart OpenRCT2.
-STR_6400    :I have the GOG installer for RollerCoaster Tycoon 2 downloaded, but it is not installed
+STR_6400    :I have the GOG offline installer for RollerCoaster Tycoon 2 downloaded, but it is not installed
 STR_6401    :I have RollerCoaster Tycoon 2 installed already
 STR_6402    :OpenRCT2 Data Setup
 STR_6403    :Select which applies best to you
 STR_6404    :Please select GOG RollerCoaster Tycoon 2 installer
 STR_6405    :Select GOG Installer
 STR_6406    :GOG RollerCoaster Tycoon 2 Installer
-STR_6407    :This may take a few minutes
-STR_6408    :Please install innoextract to extract GOG Installer
+STR_6407    :This may take a few minutes.
+STR_6408    :Please install innoextract to extract GOG Installer.
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3655,11 +3655,12 @@ STR_6400    :I have the GOG offline installer for RollerCoaster Tycoon 2 downloa
 STR_6401    :I have RollerCoaster Tycoon 2 installed already
 STR_6402    :OpenRCT2 Data Setup
 STR_6403    :Select which applies best to you
-STR_6404    :Please select GOG RollerCoaster Tycoon 2 installer
+STR_6404    :Please select the GOG RollerCoaster Tycoon 2 installer.
 STR_6405    :Select GOG Installer
 STR_6406    :GOG RollerCoaster Tycoon 2 Installer
 STR_6407    :This may take a few minutes.
-STR_6408    :Please install innoextract to extract GOG Installer.
+STR_6408    :Please install “innoextract” to extract GOG Installer, then restart OpenRCT2.
+STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
 
 #############
 # Scenarios #
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -36,6 +36,7 @@
 
 0.3.2 (2020-11-01)
 ------------------------------------------------------------------------
+- Feature: [#12307] For Linux users, allow extraction of GOG installer via innoextract.
 - Feature: [#12110] Add Hybrid Coaster (Rocky Mountain Construction I-Box) track type.
 - Feature: [#12999] .sea (RCT Classic) scenarios are now listed in the “New Scenario” dialog.
 - Feature: [#13000] objective_options command for console.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,6 +3,7 @@
 - Feature: [#6677] Add Discord RPC to macOS builds.
 - Feature: [#6844] Enhanced track designer with ability to add/remove scenery and footpaths.
 - Feature: [#11859] Add on-ride photo section to Air Powered Vertical and Reverse Freefall Coaster.
+- Feature: [#12307] Allow extraction of GOG installer via innoextract (for Linux users).
 - Feature: [#13057] Make GameAction flags accessible by plugins.
 - Feature: [#13078] [Plugin] Add colour picker widget.
 - Feature: [#13376] Open custom window at specified tab.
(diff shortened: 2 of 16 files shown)


keithstellyes:
@ocalhoun6 you seemed to have thoughts on the wording of these options, tagging you

data/language/en-GB.txt
```diff
@@ -3651,6 +3651,16 @@ STR_6390    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in or
...
+STR_6394    :I have the GOG installer for RollerCoaster Tycoon 2 downloaded, but it is not installed
+STR_6395    :I have RollerCoaster Tycoon 2 installed already
+STR_6396    :OpenRCT2 Data Setup
+STR_6397    :Select which applies best to you
+STR_6398    :Please select GOG RollerCoaster Tycoon 2 installer
+STR_6399    :Select GOG Installer
+STR_6400    :GOG RollerCoaster Tycoon 2 Installer
+STR_6401    :This may take a few minutes
+STR_6402    :Please install innoextract to extract GOG Installer
+>>>>>>> Support GOG installer extraction during setup for POSIX systems
```
Gymnasiast, line 3663: Leftover VCS marker?

src/openrct2-ui/UiContext.Linux.cpp
```diff
@@ -97,14 +99,14 @@ namespace OpenRCT2::Ui
                 {
                     std::string cmd = String::Format(
                         "%s --title \"OpenRCT2\" --msgbox \"%s\"", executablePath.c_str(), message.c_str());
-                    Execute(cmd);
+                    Platform::Execute(cmd);
```
Gymnasiast, line 102: What is the rationale behind this?

src/openrct2-ui/UiContext.Linux.cpp
```diff
@@ -113,16 +115,72 @@ namespace OpenRCT2::Ui
...
+            DIALOG_TYPE dtype = GetDialogApp(&executablePath);
+
+            size_t longest_string = 0;
+            for(const auto& option : options) {
+                if(option.size() > longest_string) {
+                    longest_string = option.size();
+                }
+            }
+
+            // If zenity and kdialog were sane, they would support some kind of
```
Gymnasiast, line 130: Avoid putting opinions in the code. Just state they do not support automatic scaling and leave it at that.

src/openrct2/platform/Platform.macOS.mm
```diff
@@ -130,7 +130,7 @@ bool HandleSpecialCommandLineArgument(const char* argument)
         }
         return false;
     }
-
+
```
Gymnasiast, line 133: Stray change.

src/openrct2-ui/UiContext.Linux.cpp
```diff
@@ -113,16 +115,72 @@ namespace OpenRCT2::Ui
...
+                    Platform::Execute(sb.GetBuffer(), &buff);
+                    return std::find(options.begin(), options.end(), buff) - options.begin();
+                }
+                case DIALOG_TYPE::KDIALOG:
+                {
+                    auto sb = StringBuilder();
+                    sb.Append(reinterpret_cast<utf8*>(String::Format("kdialog --geometry %dx%d --title '%s' --menu ",
+                        width, height, title.c_str())));
+                    sb.Append(reinterpret_cast<utf8*>(String::Format(" '%s'", text.c_str())));
+                    for (const auto& option : options) {
```
Gymnasiast, line 157: This does not adhere to our code style. Make sure to use `clang-format` on every commit. If set up correctly, you can even do `git clang-format`.

src/openrct2/ui/UiContext.h
```diff
@@ -113,6 +113,7 @@ namespace OpenRCT2
             virtual void TriggerResize() abstract;
 
             virtual void ShowMessageBox(const std::string& message) abstract;
+            virtual int ShowMenuDialog(const std::vector<std::string>& options, const std::string& title, const std::string& text) abstract;
```
Gymnasiast, line 116: Can you add a docblock here? It is not clear from the context why it returns an integer and what it means.

src/openrct2/config/Config.cpp
```diff
@@ -798,13 +811,61 @@ bool config_find_or_browse_install_directory()
...
                 uiContext->ShowMessageBox(format_string(STR_NEEDS_RCT2_FILES, nullptr));
-
-                std::string installPath = uiContext->ShowDirectoryDialog(format_string(STR_PICK_RCT2_DIR, nullptr));
+                std::string gog = language_get_string(STR_OWN_ON_GOG);
+                std::string hdd = language_get_string(STR_INSTALLED_ON_HDD);
+
+                std::vector<std::string> options;
+                std::string chosenOption;
+                // so far, only Linux has ShowMenuDialog implemented
+#if (defined(__unix__) || defined(__EMSCRIPTEN__)) && !defined(__ANDROID__) && !defined(__APPLE__)
```
Gymnasiast, line 820: This should be a platform function. It could be named `HasMenuSupport` and return a bool.

Gymnasiast (commented):
This PR essentially deals with two separate things: adding the menu, and refactoring some stuff into Platform functions.

I would like you to spin off the `FindApp` and `Execute` functions into their own PR.

keithstellyes:
Hold for merge until confirmation it builds and can do initial set-up (i.e., picking the RCT2 data directory)  on non-Linux platforms

src/openrct2-ui/UiContext.Linux.cpp
```diff
@@ -250,6 +251,77 @@ namespace OpenRCT2::Ui
             return result;
         }
 
+        bool HasMenuSupport() override
+        {
+#    ifdef __linux__
```
Gymnasiast, line 256: zenity and kdialog can be available on *BSD too. There is no need to ifdef it.

src/openrct2/ui/DummyUiContext.cpp
```diff
@@ -90,6 +90,15 @@ namespace OpenRCT2::Ui
         void ShowMessageBox(const std::string& /*message*/) override
         {
         }
+        bool HasMenuSupport() override
+        {
+            return false;
+        }
+
+        int ShowMenuDialog(const std::vector<std::string>& options, const std::string& title, const std::string& text) override
```
Gymnasiast, line 98: We always use integers with a specific size, like `int8_t`.

src/openrct2/config/Config.cpp
```diff
@@ -712,6 +713,21 @@ namespace Config
         }
         return std::string();
     }
+
+    static bool ExtractGogInstaller(const utf8* installerPath, const utf8* targetPath)
+    {
+        std::string path;
+        std::string output;
+
+        if (!Platform::FindApp("innoextract", &path))
```
tupaschoal, line 722: This function is only defined on Posix/Linux plataform headers, that's why all of the windows builds are failing.

Gymnasiast:
`Stub Execute() for Win32` does compile, it's CCache acting up again.

data/language/en-GB.txt
```diff
@@ -3651,6 +3651,15 @@ STR_6396    :Disable screensaver and monitor power saving
 STR_6397    :{SMALLFONT}{BLACK}If checked, screensaver and other monitor power saving features will be inhibited while OpenRCT2 is running.
 STR_6398    :File contains unsupported ride types. Please update to a newer version of OpenRCT2.
 STR_6399    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work. Please set the “game_path” variable in config.ini to the directory where you installed RollerCoaster Tycoon 2, then restart OpenRCT2.
+STR_6400    :I have the GOG installer for RollerCoaster Tycoon 2 downloaded, but it is not installed
```
Gymnasiast, line 3654:
```suggestion
STR_6400    :I have the GOG offline installer for RollerCoaster Tycoon 2 downloaded, but it is not installed
```
People might otherwise confuse this with the GOG Galaxy downloader stub. (Of course, plenty still will, but hopefully this will prevent _some_ people from making that mistake.)

data/language/en-GB.txt
```diff
@@ -3651,6 +3651,15 @@ STR_6396    :Disable screensaver and monitor power saving
...
+STR_6401    :I have RollerCoaster Tycoon 2 installed already
+STR_6402    :OpenRCT2 Data Setup
+STR_6403    :Select which applies best to you
+STR_6404    :Please select GOG RollerCoaster Tycoon 2 installer
+STR_6405    :Select GOG Installer
+STR_6406    :GOG RollerCoaster Tycoon 2 Installer
+STR_6407    :This may take a few minutes
```
Gymnasiast, line 3661: Is this a notification? If so, please put a dot at the end.

Same for 6408.

distribution/changelog.txt
```diff
@@ -25,6 +25,7 @@
 
 0.3.2 (2020-11-01)
 ------------------------------------------------------------------------
+- Feature: [#12307] For Linux users, allow extraction of GOG installer via innoextract.
```
Gymnasiast, line 28: Needs to be moved up to 0.3.3.

src/openrct2/config/Config.cpp
```diff
@@ -807,13 +823,71 @@ bool config_find_or_browse_install_directory()
...
+                        // graceful fallback if app errors or user exits out of window
+                        chosenOption = hdd;
+                    }
+                    else
+                    {
+                        chosenOption = options[optionIndex];
+                    }
+                }
+
+                chosenOption = hdd;
```
Gymnasiast, line 849: Should be wrapped in an `else`, otherwise it always assumes HDD installs.

Gymnasiast:
This now needs testing of the following configurations:
- HDD
- GOG with innoextract installed
- GOG without innoextract installed

PR #13175, opened 2020-10-12
Remove incorrect variable from interval names

duncanspumpkin:
Not sure why this has ended up in the interval names as it is not a rct_string_id. Luckily nothing at present was accessing this.


duncanspumpkin:
Hacktoberfest worthy.

PR #13180, opened 2020-10-13
Refactor ObjectFactory to use unique_ptr

### LINKED ISSUE #13100: Refactor ObjectFactory::CreateObject(...) to use smart pointers
This does fix this specific instance but we still need to track that this needs conversion to using something like a unique ptr. This PR is not to be merged without making an issue for that.

_Originally posted by @duncanspumpkin in https://github.com/OpenRCT2/OpenRCT2/pull/13084#discussion_r499161466_

In issue #12523 was discussed that a better approach to fix the warning number 1 listed would be to use a smart pointer like `std::unique_ptr`. But it would involve more changes in code than it was expected. So I'm opening this issue to keep track of that.

vjavs:
Employs the smart pointer unique_ptr for safer memory management.

Classes involved:

 - ObjectRepository

 - ObjectManager

Closes #13100

### DIFF
--- a/src/openrct2/object/ObjectFactory.cpp
+++ b/src/openrct2/object/ObjectFactory.cpp
@@ -164,7 +164,7 @@ namespace ObjectFactory
      * @param jRoot Must be JSON node of type object
      * @note jRoot is deliberately left non-const: json_t behaviour changes when const
      */
-    static Object* CreateObjectFromJson(
+    static std::unique_ptr<Object> CreateObjectFromJson(
         IObjectRepository& objectRepository, json_t& jRoot, const IFileDataRetriever* fileRetriever);
 
     static uint8_t ParseSourceGame(const std::string& s)
@@ -183,11 +183,11 @@ namespace ObjectFactory
         return (result != LookupTable.end()) ? result->second : OBJECT_SOURCE_CUSTOM;
     }
 
-    static void ReadObjectLegacy(Object* object, IReadObjectContext* context, OpenRCT2::IStream* stream)
+    static void ReadObjectLegacy(Object& object, IReadObjectContext* context, OpenRCT2::IStream* stream)
     {
         try
         {
-            object->ReadLegacy(context, stream);
+            object.ReadLegacy(context, stream);
         }
         catch (const IOException&)
         {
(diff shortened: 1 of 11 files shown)


src/openrct2/object/ObjectFactory.cpp
```diff
@@ -183,7 +183,7 @@ namespace ObjectFactory
         return (result != LookupTable.end()) ? result->second : OBJECT_SOURCE_CUSTOM;
     }
 
-    static void ReadObjectLegacy(Object* object, IReadObjectContext* context, OpenRCT2::IStream* stream)
+    static void ReadObjectLegacy(const std::unique_ptr<Object>& object, IReadObjectContext* context, OpenRCT2::IStream* stream)
```
IntelOrca, line 186: This can just be `Object&`.

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -369,46 +364,38 @@ namespace ObjectFactory
...
         {
             Console::Error::WriteLine("Unable to open or read '%s': %s", path.c_str(), err.what());
         }
 
-        return result;
+        return nullptr;
     }
 
-    static void ExtractSourceGames(const std::string& id, json_t& jRoot, Object* result)
+    static void ExtractSourceGames(const std::string& id, json_t& jRoot, const std::unique_ptr<Object>& result)
```
IntelOrca, line 398: `Object&`

src/openrct2/object/ObjectFactory.cpp
vjavs: Got it.

src/openrct2/object/ObjectFactory.cpp
vjavs: Ok

IntelOrca (approved):

ZehMatt:
I wonder how this passed the CI

src/openrct2/object/ObjectManager.cpp
```diff
@@ -589,14 +589,14 @@ class ObjectManager final : public IObjectManager
         std::mutex commonMutex;
         ParallelFor(requiredObjects, [this, &commonMutex, requiredObjects, &objects, &badObjects, &loadedObjects](size_t i) {
             auto ori = requiredObjects[i];
-            Object* loadedObject = nullptr;
+            std::unique_ptr<Object> object;
             if (ori != nullptr)
             {
-                loadedObject = ori->LoadedObject;
+                auto loadedObject = ori->LoadedObject;
```
ZehMatt, line 595: There is a difference here. It will never assign the already loaded object into objects[i] in case ```ori != nullptr```

janisozaur:
This caused a regression: https://github.com/OpenRCT2/OpenRCT2/issues/13912

PR #13181, opened 2020-10-13
Ride window: position graph buttons properly for CJK

AaronVanGeffen:
This is a subtle change concerning the ride window, specifically the four buttons at the bottom of the graph tab. Their widget definitions were already set to the correct height; it's just the invalidate event that needed changing.

### Before/after (French)

Diamond Heights 2020-10-14 16-57-07 Diamond Heights 2020-10-14 16-59-10

### Before/after (Japanese)

Diamond Heights 2020-10-14 16-57-14 Diamond Heights 2020-10-14 16-59-03

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -5742,7 +5742,7 @@ static void window_ride_graphs_mouseup(rct_window* w, rct_widgetindex widgetInde
  */
 static void window_ride_graphs_resize(rct_window* w)
 {
-    window_set_resize(w, 316, 180, 500, 450);
+    window_set_resize(w, 316, 182, 500, 450);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


tupaschoal (commented):
Can you add before and after screenshots for the sake of documenting?

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -5918,7 +5918,7 @@ static void window_ride_graphs_invalidate(rct_window* w)
     window_ride_graphs_widgets[WIDX_GRAPH_ALTITUDE].top = y;
     window_ride_graphs_widgets[WIDX_GRAPH_VERTICAL].top = y;
     window_ride_graphs_widgets[WIDX_GRAPH_LATERAL].top = y;
-    y += 11;
+    y += 13;
```
Gymnasiast, line 5921: Is there a constant for this?

AaronVanGeffen: Right, I'll go with `BUTTON_FACE_HEIGHT + 1` then.

AaronVanGeffen:
Added a few screenshots to the OP. Amended the patch to use the `BUTTON_FACE_HEIGHT` constant.

PR #13184, opened 2020-10-14
GameActions cleanup #1

ZehMatt:
This one is pretty minor, it renames GA_ERROR to GameActions::Status and GA_FLAGS into GameActions::Flags and make it more consistent with our style.


src/openrct2-ui/windows/Footpath.cpp
```diff
@@ -1,4 +1,4 @@
-/*****************************************************************************
+﻿/*****************************************************************************
```
Gymnasiast, line 1: Looks like you introduced a BOM in all files. Please remove them.

Gymnasiast (commented):
Looks all right to me, but would like an approval of a team member who has done more with multiplayer.

tupaschoal (approved):
> renames GA_ERROR to GameActions::Status

THANK YOU! 🚀

Issue #13160, opened 2020-10-11
Improve organisation of the cheats window


AaronVanGeffen:
The past few days, I have been thinking about how to improve the cheats window. First, I would like to talk about the park cheats tab.

* First off, I think we should further categorise options, so they will be easier to find (back).  To this end, I propose we split off options from the 'general' frame into what I'm calling 'objective' options (name TBD). Likewise, I propose splitting off the buttons from 'staff' into a new 'maintenance' frame (name TBD).

* The 'Park parameters...' button has become superfluous, as the option is also present in the cheats menu. I propose we remove it.

* The 'Freeze weather' button acts as a toggle. I propose we turn it into a checkbox.

* We can group 'Water plants' with 'Plants do not age'. I don't feel they warrant their own group, but they can certainly be juxtaposed.

* The window is quite tall, but doesn't have to be: there is quite a bit of empty space. 

Below is a concrete proposal for what the park cheats tab could look like. (Old on the left, new on the right.)

I welcome any input.

![cheats_2_new](https://user-images.githubusercontent.com/604665/95675994-5ce16f80-0bbb-11eb-83c8-9682ddcc01cb.png)

duncanspumpkin:
The code could also do with general splitting up into tabs in some of the functions

knomepasi:
Could the weather dropdown be layout the same way as park rating forcing?

I would also swap the clear/mowed grass row with the litter/vandalism row, to keep the grass and plants next to each other.

AaronVanGeffen:
> Could the weather dropdown be layout the same way as park rating forcing?

I would prefer not to, as -to me- that would suggest you could only change the weather if you force it to stay. In this, it is different from the park rating cheat.

> I would also swap the clear/mowed grass row with the litter/vandalism row, to keep the grass and plants next to each other.

Nice idea! I'll add that.

ocalhoun6:
>as the option is also present in the cheats menu.

While we're on the subject of reorganizing things ... I've never really liked that there's a cheats menu _and_ a cheats window. Wouldn't it be nicer if all the cheats were in one place?

Obviously, some of these controls need to be in a window, so I'd propose ditching the cheats menu altogether and moving its options into the cheat window, integrating them into this organization as well. Maybe into a kind of a cheats 'home' tab in this window, so the items on that menu are some of the first things that pop up when you open the cheats window.

knomepasi:
[...]

PR #13186, opened 2020-10-14
Improve park cheats tab layout

AaronVanGeffen:
As proposed and discussed in #13160, I have improved the layout of the park cheats tab.

* The 'general' group has been split in two, introducing a new 'objective' group.
* The 'park parameters' button has been removed, as 'Scenario Options' is already accessible via the cheats menu.
* 'Force weather' has been relabelled 'Change weather to'.
* 'Freeze weather' is now a checkbox instead of a button.
* The 'staff' group has been split in two, introducing a new 'maintenance' group.

### Discussion points

* [x] Should we shrink the park rating spinner, so there is a little more wiggle room for translations?
* [x] Should we consider relabelling 'Freeze weather' with 'Disable weather changes'?
* [x] Should we consider relabelling 'Mowed grass' with 'Mow grass' — make it clear that it's an action, like 'Clear grass'?
* [x] In the proposal, 'Water plants' and 'Plants do not age' were on the same line, but this apparently causes some localisation problems. Deferred.

### Before/after (English)

cheats-old-english cheats-new-english

### Before/after (French)

cheats-old-french cheats-new-french


data/language/en-GB.txt
```diff
@@ -1967,7 +1967,7 @@ STR_2763    :???
 STR_2765    :Large Tram
 STR_2766    :Win scenario
 STR_2767    :Freeze weather
-STR_2768    :Unfreeze weather
+STR_2768    :<not used anymore>
```
Gymnasiast, line 1970: We used to do this, but for the last 1-2 years we have just removed them. (Strings don't get reused anyway and leaving them out makes it easier to detect old strings in the translations.)

Gymnasiast:
In the screenshot, "Staff" and "Maintenance" both have the same name. Has this since been fixed in the code?

src/openrct2/localisation/StringIds.h
```diff
@@ -2078,7 +2078,7 @@ enum
     STR_CHEAT_LARGE_TRAM_GUESTS = 2765,
     STR_CHEAT_WIN_SCENARIO = 2766,
     STR_CHEAT_FREEZE_WEATHER = 2767,
-    STR_CHEAT_UNFREEZE_WEATHER = 2768,
+    // STR_CHEAT_UNFREEZE_WEATHER = 2768,
```
Gymnasiast, line 2081: Remove this line entirely.

data/language/en-GB.txt
AaronVanGeffen: Ah, I found one more such entry in the file -- that's why I went with this format. I'll remove the lot tomorrow, then.

tupaschoal (commented):
> The 'staff' group has been split in two, introducing a new 'maintenance' group.

I found it a bit weird to have two groups with the same title in the same window

> Should we shrink the park rating spinner, so there is a little more wiggle room for translations?

I'd say yes, but it wouldn't help portuguese by a lot, so I guess if you like it being aligned with all other similar elements in the window, then no.

> Should we consider relabelling 'Freeze weather' with 'Disable weather changes'?

Yes!

> Should we consider relabelling 'Mowed grass' with 'Mow grass' — make it clear that it's an action, like 'Clear grass'?

Yes!

AaronVanGeffen:
I've updated the strings in question, and the screenshots in the OP to reflect the changes.

Gymnasiast:
> Should we shrink the park rating spinner, so there is a little more wiggle room for translations?

It would probably be better.

> Should we consider relabelling 'Freeze weather' with 'Disable weather changes'?

Good idea. It's clearer.

> Should we consider relabelling 'Mowed grass' with 'Mow grass' — make it clear that it's an action, like 'Clear grass'?

Yes, it's clearer and more consistent.

> In the proposal, 'Water plants' and 'Plants do not age' were on the same line, but this apparently causes some localisation problems. Deferred.

Having them on separate lines is best for localisation, I think.

Umdlye:
Unless there's a good reason for keeping them this way, I would love to see "Park" and "Ducks" in the cheat buttons written in lowercase for consistency.

AaronVanGeffen:
I believe that's everything for which consensus has been reached. The screenshots in the OP have been updated to reflect the latest changes.

Unless there are urgent matters, I would prefer to leave further improvements to another PR. This is done as far as I'm concerned.

Broxzier (approved):

Gymnasiast (approved):

PR #13197, opened 2020-10-15
Fix #13192: Crash because of objects not properly loading

### LINKED ISSUE #13192: Crash when Opening Saved Game
**OS: Windows 10** [e.g. Windows 10]
**Version: 0.3.1** [e.g. 0.3.1]
**Commit/Build: 711dd00** [e.g. 426e106]

Game crashes & creates dump file whenever opening new & saved games.  Cannot Alt+tab, cannot Alt+f4, only option is to shut down or restart computer.

- [N] Reproducible in RCT2 (vanilla)?
- [N] Specific to multiplayer?

**Steps to reproduce:**
1. Open OpenRCT2 V 0.3.1
2. Open Saved Game
3. Game Crash

**Dump file**

**Screenshots / Video:**

**Save game:**

ZehMatt:
Was caused because of #13180 

Closes #13192

### DIFF
--- a/src/openrct2/object/ObjectManager.cpp
+++ b/src/openrct2/object/ObjectManager.cpp
@@ -344,7 +344,7 @@ private:
         return result;
     }
 
-    void SetNewLoadedObjectList(std::vector<std::unique_ptr<Object>> newLoadedObjects)
+    void SetNewLoadedObjectList(std::vector<std::unique_ptr<Object>>&& newLoadedObjects)
     {
         if (newLoadedObjects.empty())
         {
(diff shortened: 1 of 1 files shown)


tupaschoal:
Is it possible that all those crashes on the swinging ship have the same root cause?

ZehMatt:
I'll check it out

ZehMatt:
No, different crash because of nullptr and some odd hacked ride. rideEntry is null at those crashes.

tupaschoal (approved):

tupaschoal:
> No, different crash because of nullptr and some odd hacked ride. rideEntry is null at those crashes. But also possible still, its missing one object I don't have so who knows whats inside.

Maybe @Gymnasiast can confirm? I'm just finding all this sudden burst of paint related crashes odd

ZehMatt:
The park uses a custom object which is also not on NEDesigns so its possible someone is fiddling around with custom objects, I wish the parks exported in those crashes would just include them.

vjavs:
Thanks for the fix.

Gymnasiast:
@tupaschoal Could you fill me in on the swinging ship crashes? Do we have an issue or a save somewhere?

tupaschoal:
All issues from 13191 to 13196 basically

PR #13205, opened 2020-10-15
Fix #7518 : Water isn't cut down by view clipping tool

### LINKED ISSUE #7518: Water isn't cut down by view clipping tool
**Version:** 0.1.3
**Commit/Build:** c89e321f352612fdfa9367d6476ced8645320e47

The water level is above the clip-height (you can see the hill and path being cut off), but the water is not cut off:
image

I can think of two solutions:
- Not drawing the water at all. This makes it clear to see what's under water.
- Cutting the water level off at clip height. Gives more of a sense of looking underwater.

raacker:
By checking GetWaterHeight for surface tile, it can be clipped properly since surface tile's base_height is not being modified and being used to show base surface level.

Just one concern, I was trying to make a virtual function call to make a clean api but it feels like too much work for this small bug fix. 

So even though [this line](https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/paint/tile_element/Paint.TileElement.cpp#L248) is the proper sanity check for clip height and also I can even just do a cast beside that line, I'm just checking it inside of surface_paint. 

Any better idea to check the value?

### DIFF
--- a/src/openrct2/paint/tile_element/Paint.Surface.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Surface.cpp
@@ -1288,45 +1288,44 @@ void surface_paint(paint_session* session, uint8_t direction, uint16_t height, c
         std::memcpy(session->RightTunnels, backupRightTunnels, sizeof(tunnel_entry) * TUNNEL_MAX_COUNT);
     }
 
-    if (tileElement->AsSurface()->GetWaterHeight() > 0)
+    const uint16_t waterHeight = tileElement->AsSurface()->GetWaterHeight();
+    const bool waterGetsClipped = (session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW) && (waterHeight > gClipHeight * COORDS_Z_STEP);
+
+    if (waterHeight > 0 && !gTrackDesignSaveMode && !waterGetsClipped)
     {
         // loc_6615A9: (water height)
         session->InteractionType = VIEWPORT_INTERACTION_ITEM_WATER;
 
         const uint16_t localHeight = height + 16;
-        const uint16_t waterHeight = tileElement->AsSurface()->GetWaterHeight();
 
-        if (!gTrackDesignSaveMode)
+        session->WaterHeight = waterHeight;
+
+        int32_t image_offset = 0;
+        if (waterHeight <= localHeight)
         {
-            session->WaterHeight = waterHeight;
+            image_offset = byte_97B740[surfaceShape & 0xF];
+        }
 
-            int32_t image_offset = 0;
-            if (waterHeight <= localHeight)
-            {
-                image_offset = byte_97B740[surfaceShape & 0xF];
-            }
+        const int32_t image_id = (SPR_WATER_MASK + image_offset) | IMAGE_TYPE_REMAP | IMAGE_TYPE_TRANSPARENT
+            | PALETTE_WATER << 19;
+        sub_98196C(session, image_id, 0, 0, 32, 32, -1, waterHeight);
 
-            const int32_t image_id = (SPR_WATER_MASK + image_offset) | IMAGE_TYPE_REMAP | IMAGE_TYPE_TRANSPARENT
-                | PALETTE_WATER << 19;
-            sub_98196C(session, image_id, 0, 0, 32, 32, -1, waterHeight);
+        paint_attach_to_previous_ps(session, SPR_WATER_OVERLAY + image_offset, 0, 0);
 
-            paint_attach_to_previous_ps(session, SPR_WATER_OVERLAY + image_offset, 0, 0);
+        if (!(session->ViewFlags & VIEWPORT_FLAG_HIDE_VERTICAL))
+        {
+            // This wasn't in the original, but the code depended on globals that were only set in a different conditional
+            const uint32_t edgeStyle = tileElement->AsSurface()->GetEdgeStyle();
+            // end new code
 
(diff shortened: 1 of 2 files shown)


Broxzier:
It might be nicer to instead draw the water at the clip-height when its surface is higher, instead of not drawing it at all.

raacker:
@Broxzier I thought it is a bit inconsistent? because most of the tiles are all clipped out by clip height. But I get what you mean also because if we always draw water at the clip height we can still see the water level

Broxzier:
I get what you mean, but I think for water it's appropriate since it's a volume, not just a layer at a set height. The only other thing in-game similar to this are supports, which _do_ get hidden when an element is higher than the cut-away height, so in a way it's not consistent.

raacker:
Okay I'm totally at the same page with you now. I'll try to change the behavior and push the commit. After that, will rebase

raacker:
@Broxzier Actually I found the further problem. I can just render at clipping height but what about surface tile actions and rendering? 

For the normal land action, we just don't show the tile and don't have to worry about it. But if we start showing it, action tool will remain as available. Maybe we can just simply disable this and add an another fix for merged water tile texture.

image

Broxzier:
It may be quite tricky to get this right, take a look at the edges in the middle of the 2x2 square. I remember running into some annoying issues when creating horizontal clipping.

I think for this PR it'll be best to _only_ hide the water, but show the surface anyway. This can be easily achieved by moving the check you made down to [here](https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/paint/tile_element/Paint.Surface.cpp#L1291). (I just tried setting the water height there to match the clip-height, but that resulted in the same edge issue).

https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/paint/tile_element/Paint.Surface.cpp#L1291

Also, make sure to first check the viewport flags for `VIEWPORT_FLAG_CLIP_VIEW`, otherwise you would always be clipping, even when the tool is closed.

raacker:
@Broxzier Actually you are right, it's just better not to show the water but surface. Only one problem left will be that your water raise action is still available at the water level height

Broxzier:
I don't really think that's an issue, and certainly not for this PR. Surfaces that are clipped can be altered by the land tool too when the tile is within the selection.

raacker:
@ Broxzier I see. Alright then. Even though this was a pretty simple fix, I learned a lot of the tile structure haha. Sorry for so many rebasing due to my lack of commit quality. Then could you confirm the behavior for the last time?

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -1296,6 +1296,10 @@ void surface_paint(paint_session* session, uint8_t direction, uint16_t height, c
         const uint16_t localHeight = height + 16;
         const uint16_t waterHeight = tileElement->AsSurface()->GetWaterHeight();
 
+        if ((session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW)
+            && (tileElement->AsSurface()->GetWaterHeight() > gClipHeight * COORDS_Z_STEP))
+            return;
```
Broxzier, line 1301: Don't just break out of this function, later on there is code that would draw the park fences. This would hide those. Only skip this section of the paint function when the water gets clipped.

Also, you can use waterHeight here instead of calling the function again. It would be even nicer if you moved that variable just before the parent if-condition and use it there as well.

Broxzier (changes requested):

src/openrct2/paint/tile_element/Paint.Surface.cpp
raacker: Oh that's true. My bad. Could you check the fix commit?

raacker: Actually, we can just merge all 3 conditions into one parent if statement as ParkFences check does

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -1288,45 +1288,43 @@ void surface_paint(paint_session* session, uint8_t direction, uint16_t height, c
         std::memcpy(session->RightTunnels, backupRightTunnels, sizeof(tunnel_entry) * TUNNEL_MAX_COUNT);
     }
 
-    if (tileElement->AsSurface()->GetWaterHeight() > 0)
+    const uint16_t waterHeight = tileElement->AsSurface()->GetWaterHeight();
+    if ((waterHeight > 0) && !gTrackDesignSaveMode
+        && !((session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW) && (waterHeight > gClipHeight * COORDS_Z_STEP)))
```
Broxzier, line 1293: That works pretty well, last suggestion would be to make it more readable:
```suggestion
    bool const waterGetsClipped = (session->ViewFlags & VIEWPORT_FLAG_CLIP_VIEW) && waterHeight > gClipHeight * COORDS_Z_STEP;
    if (waterHeight > 0 && !gTrackDesignSaveMode && !waterGetsClipped)
```

Broxzier:
Don't forget to also add a changelog entry.

raacker:
@Broxzier Thank you for all help and reminding me of changelog!

tupaschoal:
Our changelog is sorted by type of commit and number of issue. Can you amend your change to keep it sorted, please?

raacker:
@tupaschoal thanks for letting me know.

Broxzier (approved):

PR #13206, opened 2020-10-15
Close#5178: Lighting effects are disabled in software mode

Lukasz-Pe:
Lightning effect are now automatically turned off when software renderer is selected.

The change is made once in file `openrct2-ui/windows/Options.cpp`, but does not affect config.ini file.
Here it is only cosmetic change, which is visible right after renderer change.
Second check of selected renderer is done when loading settings from file. This changes the state of disabled checkboxes after game restart.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -9,6 +9,7 @@
 - Fix: [#1324] Last track piece map selection still visible when placing ride entrance or exit (original bug).
 - Fix: [#3200] Close Construction window upon selecting vehicle page.
 - Fix: [#4041] Garbled park option on scenario editor with custom theme.
+- Fix: [#5178] Lighting effects are disabled in software mode
 - Fix: [#5904] Empty errors on tile inspector base height change.
 - Fix: [#6086] Cannot install existing track design with another name.
 - Fix: [#7748] Tooltips would not timeout for normal UI elements.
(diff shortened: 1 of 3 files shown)


IntelOrca:
Can you update the changelog?

distribution/changelog.txt
```diff
@@ -9,6 +9,7 @@
 - Fix: [#1324] Last track piece map selection still visible when placing ride entrance or exit (original bug).
 - Fix: [#3200] Close Construction window upon selecting vehicle page.
 - Fix: [#4041] Garbled park option on scenario editor with custom theme.
+- Fix: [#5178] Lighting effects are disabled in software mode
```
AaronVanGeffen, line 12:
```suggestion
- Fix: [#5178] Lighting effects cannot be disabled in software mode.
```

src/openrct2/config/Config.cpp
```diff
@@ -187,9 +187,16 @@ namespace Config
...
+            if (model->drawing_engine != DrawingEngine::Software)
+            {
+                model->enable_light_fx = reader->GetBoolean("enable_light_fx", false);
+                model->enable_light_fx_for_vehicles = reader->GetBoolean("enable_light_fx_for_vehicles", false);
+            }
+            else
+            {
+                model->enable_light_fx = false;
+                model->enable_light_fx_for_vehicles = false;
+            }
```
AaronVanGeffen, line 199:
```suggestion
            bool isHardware = model->drawing_engine != DrawingEngine::Software;
            model->enable_light_fx = isHardware && reader->GetBoolean("enable_light_fx", false);
            model->enable_light_fx_for_vehicles = isHardware && reader->GetBoolean("enable_light_fx_for_vehicles", false);
```

AaronVanGeffen (approved):

PR #13209, opened 2020-10-15
Fix hybrid coaster

X123M3-256:
This PR fixes a few issues with the hybrid coaster that I only noticed after it was merged. These are:

* Train glitches on verticals and steep slopes

* On ride photo displays brake sprite when viewed from some angles

* Block brake sprite not shown on final station tile.

* JSON objects cannot use the hybrid ride type

* Several unnecessary comments were left in

### DIFF
--- a/src/openrct2/ride/coaster/HybridCoaster.cpp
+++ b/src/openrct2/ride/coaster/HybridCoaster.cpp
@@ -25,14 +25,6 @@
 #include "../TrackData.h"
 #include "../TrackPaint.h"
 
-// Dynamic glitches
-// Steep to gentle/gentle to steep - sprite splitting issue requiring manual fix
-
-// Glitches present on original (fixing optional)
-// Sloped turns
-// Steep to vertical
-// Helixes (check height of bounding box but probably a sprite mask issue)
-
 namespace HybridRC
 {
     static uint32_t GetTrackColour(paint_session* session)
@@ -10010,7 +10002,6 @@ namespace HybridRC
         paint_util_set_general_support_height(session, height + 32, 0x20);
     }
 
-    /** rct2: 0x008ADA44 */
     static void TrackOnRidePhoto(
         paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
         const TileElement* tileElement)
@@ -10026,7 +10017,6 @@ namespace HybridRC
         paint_util_set_general_support_height(session, height + 48, 0x20);
     }
 
-    /** rct2: 0x008ADED4 */
     static void TrackFlatTo60DegUpLongBase(
         paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
         const TileElement* tileElement)
@@ -10165,7 +10155,6 @@ namespace HybridRC
         }
     }
 
-    /** rct2: 0x008ADEE4 */
     static void Track60DegUpToFlatLongBase(
         paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
         const TileElement* tileElement)
@@ -10304,7 +10293,6 @@ namespace HybridRC
         }
     }
 
-    /** rct2: 0x008ADEF4 */
     static void TrackFlatTo60DegDownLongBase(
         paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
         const TileElement* tileElement)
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -490,9 +562,29 @@ namespace HybridRC
         paint_session* session, uint16_t rideIndex, uint8_t trackSequence, uint8_t direction, int32_t height,
         const TileElement* tileElement)
     {
-        sub_98197C_rotated(
-            session, direction, GetTrackColour(session) | (SPR_G2_HYBRID_TRACK_VERTICAL + direction + 4), 0, 0, 32, 20, 3,
-            height, 0, 6, height + 8);
+        switch (direction)
+        {
+            case 0:
+                sub_98197C_rotated(
```
Gymnasiast, line 568: The point of sub_xxyyzz_rotated() is not to have to write out all the directions. If you do need a separate call for each direction, then you should use the non-rotated version - but make sure to adjust the parameters accordingly if you do.

X123M3-256: The bounding boxes depend on the direction. That was what was causing the bug, I replaced the switch statement with a single call and that resulted in it using the same bounding boxes for every angle.

duncanspumpkin:
Okay i see what you need. Give me a wee while and i should be able to get this working without having to revert everything.

src/openrct2/ride/coaster/HybridCoaster.cpp
Gymnasiast: I know, but using the `_rotated` calls while also splitting out per direction is just unnecessary overhead. You can achieve what you want using regular `sub_98197C()` calls.

But let's wait until Duncan has his solution ready, as it might address this issue in another way.

duncanspumpkin:
@X123M3-256 can you double check I've refactored it correctly. This is how we would like all of the functions to end up like. Then we have a simple table that can be extracted into the potential future track object format.

src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -135,18 +127,41 @@ namespace HybridRC
...
+            { 0, 6, height },
+        };
+        static const CoordsXYZ boundBoxLengths[4] = {
+            { 32, 20, 3 },
+            { 2, 24, 93 },
+            { 2, 24, 93 },
+            { 32, 20, 3 },
+        };
+        static const uint32_t imageIds[2][4] = { {
+                                                     SPR_G2_HYBRID_LIFT_TRACK_STEEP + 12,
```
Gymnasiast, line 143: This is indented too much. 4 spaces compared the line before would be best.

src/openrct2/ride/coaster/HybridCoaster.cpp
```diff
@@ -135,18 +127,41 @@ namespace HybridRC
...
+                                                     SPR_G2_HYBRID_LIFT_TRACK_STEEP + 13,
+                                                     SPR_G2_HYBRID_LIFT_TRACK_STEEP + 14,
+                                                     SPR_G2_HYBRID_LIFT_TRACK_STEEP + 15,
+                                                 },
+                                                 {
+                                                     SPR_G2_HYBRID_TRACK_STEEP + 12,
+                                                     SPR_G2_HYBRID_TRACK_STEEP + 13,
+                                                     SPR_G2_HYBRID_TRACK_STEEP + 14,
+                                                     SPR_G2_HYBRID_TRACK_STEEP + 15,
+                                                 } };
```
Broxzier, line 153: Add a trailing comma here, and clang-format will format it properly.

X123M3-256:
@duncanspumpkin  It looks like your refactor sets `WoodenSupportsPrependTo` for all angles while the original code only does it for two. I am not sure if this is a problem because I'm not sure what that line actually does, but I know it broke when I removed it.

duncanspumpkin:
Its protected by an if so that its only on the two directions requried.

duncanspumpkin:
That line is so the wooden supports are forced to be drawn in the correct order. I think its more expensive to do thats why its not applied in all situations.

X123M3-256:
@duncanspumpkin Yeah, you're right, didn't notice that. I think your code is equivalent to the original.

Gymnasiast (commented):
LGTM. I'd like @X123M3-256 to confirm everything works correctly, though.

X123M3-256:
@Gymnasiast It's all working correctly for me.

PR #13212, opened 2020-10-15
Refactor formatting strings

IntelOrca:
Eradicate all coded format tokens. All internal strings now contain raw tokens such as `{STRINGID}` and `{RED}`. New iterators have been created to iterate the tokens and the UTF-8 codepoints.

Formatting strings has been re-written and a new template version is available.


tupaschoal:
Do we want this in before the next version is released Ted?

duncanspumpkin:
@tupaschoal this is bound to produce some bugs, I would say no.

IntelOrca:
@tupaschoal there is no hurry for this

tupaschoal:
Yep, I though so, just wanted to confirm

IntelOrca:
I will apply new naming / casing suggestions.

src/openrct2/core/String.hpp
```diff
@@ -42,6 +42,7 @@ namespace String
     bool IsNullOrEmpty(const utf8* str);
     int32_t Compare(const std::string& a, const std::string& b, bool ignoreCase = false);
     int32_t Compare(const utf8* a, const utf8* b, bool ignoreCase = false);
+    bool Equals(const std::string_view& a, const std::string_view& b, bool ignoreCase);
```
tupaschoal: Should this have an `ignoreCase` default equal to false? (all other similar alternatives do...

src/openrct2/localisation/FormatCodes.cpp
```diff
@@ -76,14 +77,12 @@ static constexpr const format_code_token format_code_tokens[] = {
...
-    for (uint32_t i = 0; i < std::size(format_code_tokens); i++)
-    {
-        if (_strcmpi(token, format_code_tokens[i].token) == 0)
-            return format_code_tokens[i].code;
-    }
-    return 0;
+    auto result = std::find_if(std::begin(format_code_tokens), std::end(format_code_tokens), [token](auto& fct) {
+        return String::Equals(token, fct.token, true);
+    });
+    return result != std::end(format_code_tokens) ? result->code : 0;
```
tupaschoal: Maybe use `cbegin` and `cend`?

src/openrct2/localisation/FormatCodes.h
```diff
@@ -11,7 +11,9 @@
 
 #include "../common.h"
 
-uint32_t format_get_code(const char* token);
+#include <string_view>
+
+uint32_t format_get_code(std::string_view token);
```
tupaschoal: `const &`? I don't think string views need them, but because you used above, I'm vouching for consistency.

src/openrct2/localisation/Formatting.cpp
```diff
@@ -0,0 +1,207 @@
...
+{
+    char GetDigitSeperator()
+    {
+        return ',';
+    }
+
+    char GetDecimalSeperator()
+    {
+        return '.';
+    }
```
tupaschoal: Maybe these would be better of as `constexpr`

In any case, seems like "seperetor" has been mispelled.

src/openrct2/localisation/Formatting.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    template<size_t TDecimalPlace, bool TDigitSep, typename T> void FormatNumber(std::stringstream& ss, T value)
+    {
+        char buffer[32];
+        int32_t i = 0;
+
+        size_t num;
+        if (value < 0)
+        {
+            // TODO handle edge case: std::numeric_limits<int64_t>::min();
```
tupaschoal, line 284: Handle it?

src/openrct2/localisation/Formatting.cpp
```diff
@@ -0,0 +1,207 @@
...
+            num = -value;
```
tupaschoal, line 285: Did you intend to negate what is already negative? Because then I think this code would be a lot clearer with:
```c++
if (value < 0)
{
    ss << '-';
}
size_t num = std::abs(value);

File: src/openrct2/localisation/Formatting.cpp [hunk 7]
```diff
@@ -0,0 +1,207 @@
...
+        {
+            num = value;
+        }
+
+        // Decimal digits
+        if constexpr (TDecimalPlace > 0)
+        {
+            while (num != 0 && i < sizeof(buffer) && i < TDecimalPlace)
+            {
+                buffer[i++] = (char)('0' + (num % 10));
```
tupaschoal: I guess this probably failed compilation due to the usage of c-style cast (but as I'm looking commit by commit, I won't know now :D)

src/openrct2/localisation/Formatting.cpp
```diff
@@ -0,0 +1,207 @@
...
+        {
+            if constexpr (TDigitSep)
+            {
+                if (groupLen >= 3)
+                {
+                    groupLen = 0;
+                    buffer[i++] = GetDigitSeperator();
+                }
+            }
+            buffer[i++] = (char)('0' + (num % 10));
```
tupaschoal: Same
> I guess this probably failed compilation due to the usage of c-style cast (but as I'm looking commit by commit, I won't know now :D)

src/openrct2/localisation/Formatting.cpp
```diff
@@ -0,0 +1,207 @@
...
+            case FORMAT_STRINGID2:
+                if constexpr (std::is_integral<T>())
+                {
+                    ss << language_get_string(arg);
+                }
+                break;
+        }
+    }
+
+    std::pair<std::string_view, uint32_t> FormatNextPart(std::string_view& fmt)
```
tupaschoal: I find the whole idea of changing `fmt` on this function a bit weird, can you elaborate, please?

src/openrct2/localisation/Formatting.h
```diff
@@ -0,0 +1,102 @@
...
+#include "../common.h"
+#include "Language.h"
+
+#include <sstream>
+#include <string>
+#include <string_view>
+#include <utility>
+
+namespace OpenRCT2
+{
```
tupaschoal, line 25: Maybe there should be one more namesape? `OpenRCT2::Formatting` or `OpenRCT2::Localisation`...

src/openrct2/localisation/Formatting.cpp
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#include "Formatting.h"
+
+#include "../config/Config.h"
+#include "../util/Util.h"
+#include "FormatCodes.h"
+#include "Language.h"
```
tupaschoal, line 15: Already included in header.

test/tests/FormattingTests.cpp
```diff
@@ -0,0 +1,118 @@
...
+{
+    gConfigGeneral.measurement_format = MeasurementFormat::Metric;
+    auto actual = FormatString("Train is going at {VELOCITY}.", 1024);
+    ASSERT_EQ("Train is going at 1,648 km/h.", actual);
+}
+
+TEST_F(FormattingTests, velocity_mps)
+{
+    gConfigGeneral.measurement_format = MeasurementFormat::SI;
+    auto actual = FormatString("Train is going at {VELOCITY}.", 1024);
```
tupaschoal, line 197: These were probably not passing (and got fixed later in the PR?)

src/openrct2/localisation/Formatting.h
```diff
@@ -23,6 +23,48 @@ namespace OpenRCT2
 {
     using FormatToken = uint32_t;
 
+    class FmtString
+    {
+    private:
+        std::string_view _str;
+        std::string _strOwned;
```
tupaschoal, line 32: Don't use use TitleCase for member variables?

src/openrct2/localisation/Formatting.cpp
```diff
@@ -212,6 +245,105 @@ namespace OpenRCT2
...
+    template<size_t TDecimalPlace, bool TDigitSep, typename T> void FormatCurrency(std::stringstream& ss, T rawValue)
+    {
+        auto currencyDesc = &CurrencyDescriptors[gConfigGeneral.currency_format];
+        auto value = static_cast<int64_t>(rawValue) * currencyDesc->rate;
+
+        // Negative sign
+        if (value < 0)
+        {
+            ss << '-';
+            value = -value;
```
tupaschoal, line 343: `std::abs`

src/openrct2/drawing/Drawing.String.cpp
```diff
@@ -20,6 +21,83 @@
...
+{
+private:
+    std::string_view _str;
+
+public:
+    class iterator
+    {
+    private:
+        std::string_view _str;
+        size_t _index;
```
tupaschoal: TitleCase?

src/openrct2/drawing/Drawing.String.cpp
```diff
@@ -20,6 +21,83 @@
 
 #include <algorithm>
 
+using namespace OpenRCT2;
+
+class CodepointView
+{
+private:
+    std::string_view _str;
```
tupaschoal: Don't use use TitleCase for member variables?

tupaschoal (commented):
I gave up on 3def53a5a13661d62a45310825518b48f849850b because I was dozing

src/openrct2/drawing/Drawing.String.cpp
IntelOrca: This isn't title case.

src/openrct2/drawing/Drawing.String.cpp
IntelOrca: This isn't title case.

src/openrct2/localisation/Formatting.cpp
IntelOrca: Since we already know that it is negative, `abs` is unnecessary.

src/openrct2/localisation/Formatting.h
IntelOrca: This is lower camel case, the convention for private fields.

test/tests/FormattingTests.cpp
IntelOrca: I am not sure what you are saying. Does this test fail for you?

src/openrct2/localisation/Formatting.cpp
IntelOrca: This would incur two checks, but I will need to do some clever stuff anyway to handle the edge case of negative long max value which can't be represented as positive.

src/openrct2/localisation/Formatting.cpp
IntelOrca: Yeah, I changed to C++ cast.

src/openrct2/localisation/Formatting.cpp
IntelOrca: This no longer exists.

src/openrct2/localisation/Formatting.cpp
IntelOrca: Fix spelling, constexpr not possible anymore.

src/openrct2/core/String.hpp
IntelOrca: Yes, that was done in a later commit

src/openrct2/localisation/FormatCodes.cpp
IntelOrca: What does cbegin and cend do?

src/openrct2/localisation/FormatCodes.h
IntelOrca: @ZehMatt said to make all `string_view` not `const &`.

tupaschoal: I know, it's just that there are other places in the diff doing const& (not sure if in the latest revision though)

ZehMatt:
I don't quite understand all the thread_local use, when std::string is local variable it should be thread safe. Maybe I'm missing something about the implementation?

src/openrct2/core/String.cpp
```diff
@@ -154,6 +154,32 @@ namespace String
         }
     }
 
+    bool Equals(const std::string_view& a, const std::string_view& b, bool ignoreCase)
```
ZehMatt, line 157: Remove the reference here.

ZehMatt, line 157:
```suggestion
	constexpr bool Equals(const std::string_view a, const std::string_view b, bool ignoreCase)
	{
		using CharType = std::string_view::value_type;
		auto cmp = ignoreCase ? [](CharType a, CharType b) {
			return std::tolower(a) == std::tolower(b);
		} : [](CharType a, CharType b) {
			return a == b;
		};
		return std::equal(a.begin(), a.end(), b.begin(), cmp);
	}
```

src/openrct2/drawing/TTF.cpp
```diff
@@ -181,12 +181,12 @@ static void ttf_close_font(TTF_Font* font)
     TTF_CloseFont(font);
 }
 
-static uint32_t ttf_surface_cache_hash(TTF_Font* font, const utf8* text)
+static uint32_t ttf_surface_cache_hash(TTF_Font* font, std::string_view text)
 {
     uint32_t hash = static_cast<uint32_t>(((reinterpret_cast<uintptr_t>(font) * 23) ^ 0xAAAAAAAA) & 0xFFFFFFFF);
-    for (const utf8* ch = text; *ch != 0; ch++)
+    for (size_t i = 0; i < text.size(); i++)
```
ZehMatt, line 187:
```suggestion
    for (auto& char : text)
```

src/openrct2/drawing/TTF.cpp
```diff
@@ -181,12 +181,12 @@ static void ttf_close_font(TTF_Font* font)
     TTF_CloseFont(font);
 }
 
-static uint32_t ttf_surface_cache_hash(TTF_Font* font, const utf8* text)
+static uint32_t ttf_surface_cache_hash(TTF_Font* font, std::string_view text)
```
ZehMatt, line 184:
```suggestion
static uint32_t ttf_surface_cache_hash(TTF_Font* font, const std::string_view text)
```

src/openrct2/interface/StdInOutConsole.cpp
```diff
@@ -108,19 +108,21 @@ void StdInOutConsole::Close()
     openrct2_finish();
 }
 
-void StdInOutConsole::WriteLine(const std::string& s, uint32_t colourFormat)
+void StdInOutConsole::WriteLine(const std::string& s, FormatToken colourFormat)
 {
     std::string formatBegin;
-    if (colourFormat != FORMAT_WINDOW_COLOUR_2)
+    if (colourFormat != FormatToken::ColourWindow2)
```
ZehMatt, line 114: That if statement seems a bit pointless followed by the switch.

IntelOrca:
@ZehMatt Formatting calls are called from drawing calls. Drawing calls can be multi-threaded, so I am using threadlocal to ensure it doesn't share the same buffer on the heap. I am avoid fixed static buffers because we are trying to move away from fixed numbers which may be exceeded in some scenarios or otherwise just hog stack space.

ZehMatt:
None of the buffers should be shared, the implementation should be thread-safe by design and not by choice. I understand the argument about allocations but the stringstream isn't helping here. The current thing with the drawing right now is that it shares a global variable as buffer which ofc requires it to be either thread_local or locking, but the goal of the Formatter was to eradicate this kind of usage. Also often when the strings are small you don't have to worry about allocations due to small string optimization.

IntelOrca:
@ZehMatt how do you suggest it should be done then if we want to avoid memory allocation every single frame and avoid fixed length stack buffers?

ZehMatt:
In one of my projects I do something like this
```
        template<typename... Args> void formatMsg(MsgType type, const char* fmt, Args&&... args)
        {
            char buffer[128];
            // Use local buffer first, for 90% of the time this is enough space.
            int actualLength = snprintf(buffer, sizeof(buffer), fmt, std::forward<Args&&>(args)...);
            if (actualLength >= sizeof(buffer))
            {
                // snprintf_s returns the size of the completely formatted string, if bigger than buffer
                // it was truncated allocate then temporary buffer.
                std::unique_ptr<char[]> tempBuf(new char[actualLength + 1]);
                _snprintf_s(tempBuf.get(), actualLength + 1, actualLength, fmt, std::forward<Args&&>(args)...);
                tempBuf[actualLength] = '\0';
                return printMsg(type, tempBuf.get());
            }
            return printMsg(type, buffer);
        }```

which covers 95% of all the things going in. Also aren't the current global buffers fixed size? We can probably measure what the maximum/minimum/average is and just allocate for the worst case like above.

**Comment by IntelOrca** [comment]
My new global buffer is a stringstream or string which is cleared and re-used each frame, so it can grow to any size. I would like to avoid a fixed length, I would like 100% of cases to be correct, not just 90%.

**Comment by ZehMatt** [comment]
> My new global buffer is a stringstream or string which is cleared and re-used each frame, so it can grow to any size. I would like to avoid a fixed length, I would like 100% of cases to be correct, not just 90%.
> 
> If you allocate a temp buffer for this one really long string, you are still going to get that every tick that long string might be formatted.

I don't know what to say, having dynamic strings comes with a cost, thread_local also comes with a cost, its probably not as expensive as an allocation but its not free either. I'm not sure why you want dynamic strings, the localisation is a fixed data set, the only exceptions for this is user supplied strings as arguments which is not often really.

**Comment by IntelOrca** [comment]
@ZehMatt why isn't it free? What is involved with thread_local?

File: src/openrct2/core/String.cpp (hunk 17 above)
Comment (IntelOrca) [reply src/openrct2/core/String.cpp:157]: This does not compile

File: src/openrct2/rct2/S6Exporter.cpp [hunk 21]
```diff
@@ -1371,7 +1371,7 @@ void S6Exporter::ExportBanner(RCT12Banner& dst, const Banner& src)
         if (!(src.flags & BANNER_FLAG_IS_WALL) && !(src.flags & BANNER_FLAG_IS_LARGE_SCENERY))
         {
             char codeBuffer[32]{};
-            utf8_write_codepoint(codeBuffer, FORMAT_COLOUR_CODE_START + src.text_colour);
+            utf8_write_codepoint(codeBuffer, 142 + src.text_colour);
```
Gymnasiast, line 1374: Don't think it's a good idea to replace constants with naked numbers.

src/openrct2/scenario/ScenarioRepository.cpp
```diff
@@ -283,6 +283,18 @@ class ScenarioFileIndex final : public FileIndex<scenario_index_entry>
         return false;
     }
 
+    static bool IsLikelyUtf8(const std::string_view s)
```
Gymnasiast, line 286: Please don't duplicate this function.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -187,6 +187,18 @@ class S6Importer final : public IParkImporter
         return false;
     }
 
+    static bool IsLikelyUtf8(const std::string_view s)
+    {
+        for (auto c : s)
+        {
+            if (static_cast<uint8_t>(c) >= 128)
```
Gymnasiast, line 194: This is wrong. CP1252 has plenty of characters >= 128. It will thus detect numerous CP1252 strings and (virtually) all non-UTF8 multibyte ones as UTF-8.

The old code checked for codepoints between FORMAT_COLOUR_CODE_START and FORMAT_COLOUR_CODE_END, because these don't typically appear in scenario titles and descriptions.

IntelOrca: Done, and improved further by also checking for arg format codes.

src/openrct2-ui/interface/InGameConsole.cpp
```diff
@@ -199,13 +199,13 @@ void InGameConsole::Toggle()
...
+void InGameConsole::WriteLine(const std::string& input, FormatToken colourFormat)
 {
     // Include text colour format only for special cases
     // The draw function handles the default text colour differently
-    utf8 colourCodepoint[4]{};
-    if (colourFormat != FORMAT_WINDOW_COLOUR_2)
-        utf8_write_codepoint(colourCodepoint, colourFormat);
+    auto colourCodepoint = "";
+    if (colourFormat != FormatToken::ColourWindow2)
+        colourCodepoint = "{";
```
Gymnasiast, line 208: Can you elaborate on this?

src/openrct2/drawing/Font.cpp
```diff
@@ -240,7 +240,7 @@ void font_sprite_initialise_characters()
             int32_t width = 0;
             if (g1 != nullptr)
             {
-                if (glyphIndex < (FORMAT_ARGUMENT_CODE_START - 32) || glyphIndex >= (FORMAT_COLOUR_CODE_END - 32))
+                if (glyphIndex < 91 || glyphIndex >= 109)
```
Gymnasiast, line 243: Please do not introduce naked numbers.

src/openrct2-ui/interface/InGameConsole.cpp
IntelOrca: I think this should be:
```suggestion
        colourCodepoint = "{WINDOW_COLOUR_2}";
```

src/openrct2/localisation/FormatCodes.cpp
```diff
@@ -9,116 +9,175 @@
...
             return false;
     }
 }
 
-#pragma endregion
+size_t FormatTokenGetTextColourIndex(FormatToken token)
+{
+    switch (token)
+    {
+        case FormatToken::ColourBlack:
```
Gymnasiast, line 138: Is this (roughly) as efficient as just doing a bounds check and subtracting?

Gymnasiast (changes requested):
Getting there. I have reviewed it and have one thing to address and two things that need explanation. I would also like to test import of some saves.

Also, does this make solving bugs like https://github.com/OpenRCT2/OpenRCT2/issues/13102 easier?

IntelOrca:
@Gymnasiast yes, I think that can be solved now if you use the new template format function.

src/openrct2/localisation/FormatCodes.cpp
IntelOrca: Format tokens are now in no particular order, you can change it freely and things still work. But it requires switch like this. There was already a debate between switch and static array in a different PR, switch ensure no extra changes are required when the values of FormatToken change.

Gymnasiast (commented):
OK, all my code-related questions/remarks have been solved or answered.

I still want to test this, though.

Gymnasiast (changes requested):
Banners from RCT1 are completely broken:

afbeelding

The colours are not imported correctly, there are large blank gaps and one sign even says "palelavender".

Edit: so are banners from RCT2 parks. E.g. the no entry banner on Judge Roy Scream from SF over Texas.

Gymnasiast (changes requested):
Most of the banner bugs are gone, but the tooltip is still not quite correct:

afbeelding

It should be in the colour of the banner text. In above case, it should be pale lavender.

Additionally, the banner on the left should display "Exit", but displays "Exitnce" (I made sure to check in RCT1).

afbeelding

Both issues are in Blackpool Pleasure Beach.

afbeelding

And the name of Judge Roy Scream (SF over Texas) also contains garbage.

Looks like a buffer is not cleared correctly.

IntelOrca:
Issues were due to my last fix not being particularly good enough. I have now resolved those issues.

Gymnasiast (changes requested):
Dropdown lists lack the padding on the left, and are missing the marker that denotes the selection:

afbeelding

Gymnasiast (approved):
Above error appears to be the result of an error in nl-NL.txt that did not surface before (it worked by coincidence). It should be fixed in Localisation.

Saving and loading appears to work properly.

PR #13213, opened 2020-10-15
Export loaded objects alongside inside the save for Backtrace reports

ZehMatt:


src/openrct2/platform/Crash.cpp
```diff
@@ -168,6 +168,15 @@ static bool OnCrash(
     try
     {
         auto exporter = std::make_unique<S6Exporter>();
+
+        // Make sure the save is using the current viewport settings.
+        viewport_set_saved_view();
+
+        // Export all loaded objects to avoid having custom objects missing in the reports.
+        auto context = GetContext();
```
janisozaur, line 176: This shadows function argument, but as the argument is not used I suggest commenting it out -> `void* /*context*/` in line 90

janisozaur (commented):
The custom objects can significantly increase the upload size, but we apply gzip to the saves. To achieve better ratio set `gUseRLE` to false, like we do in network.

ZehMatt:
Is everyone ok with this @OpenRCT2/developers? I think this is a great change as half of the Backtrace issues can not be used due to missing objects and probably most crashes happen due to invalid ones. I think the increase of size is rather minimal after the compression, servers with big parks are usually no bigger than 6 MiB, uploading useless reports is worse.

tupaschoal (approved):
I'll be merging this as no one was against it, and given that the size won't be a problem and the benefits might outweigh this only con

Issue #13226, opened 2020-10-17
Opening a corrupt save file closes dialog, no error dialog


keithstellyes:
**OS:** Linux
**Version:** 0.3.1
**Commit/Build:**d287653cb5b294a0aeeefd635bea3ba3d2c4c3c5

Trying to load a corrupt save file causes the save picker dialog to simply close, rather than present an error. Found while investigating #12875 .Vanilla behavior shows the error dialog

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Specific to multiplayer?

**Steps to reproduce:**
1. Load a known corrupt save file (or other file that simply isn't a valid save file

**Screenshots / Video:**

Vanilla behavior:
![Screenshot from 2020-10-17 09-01-59](https://user-images.githubusercontent.com/13317074/96347599-97358b80-1057-11eb-93c4-b6c0930e79b8.png)

OpenRCT2 behavior (video taken from #12875): https://youtu.be/ex2j2l1vpcI

**Save game:**
Note that this is literally random data

[corrupt.zip](https://github.com/OpenRCT2/OpenRCT2/files/5395994/corrupt.zip)

PR #13231, opened 2020-10-18
Fix for no error being shown when loading corrupted park save (#13226)

keithstellyes:
Tested by:

From both main menu, and in-scenario, did both:
* Load corrupted save, ensure error dialog pops up
* Load valid save, ensure loads correctly

Screenshot from 2020-10-17 20-16-57
Screenshot from 2020-10-17 20-17-19


Gymnasiast:

`Fix: [#13226] No error is shown when attempting to load a corrupted save.`

PR #13238, opened 2020-10-18
Fix #12940 : Windows cause issues with snow drawing

### LINKED ISSUE #12940: Windows cause issues with snow drawing
**OS:** Windows 10
**Version:** 0.3.0-develop
**Commit/Build:** 564cc7b

Snow above windows behaves differently from snow on other places. The speed might be different and snowflakes (dis)appear.

**Steps to reproduce:**
1. Force weather to snow with the cheats window open.
2. Move or open more windows if the problem is not visible.

**Screenshots / Video:**
Hard to see, but there is some distortion in the snow pattern above the vertical edges of the windows. Most notable are left sides of options and R&D window. It's easier to spot in-game.
Micro Sandox 2020-09-16 11-39-22

raacker:
Windows' left value inside of cosine function is causing glitch-feel movement for snows but in fact, x_start can simply modified by another magic numbers.

* One thing I wonder, do we add all the "Fix" to changelog with no exception? I'm just curious because I didn't add it

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -20,6 +20,7 @@
 - Fix: [#8015] RCT2 files are not found when put into the OpenRCT2 folder.
 - Fix: [#8957] Error title missing when building with insufficient funds
 - Fix: [#10186] Placing multiple saved rides ignores design name (original bug).
+- Fix: [#12940] Windows cause issues with snow drawing.
 - Fix: [#13021] Mowed grass and weeds don't show up in extra zoom levels.
 - Fix: [#13024] Console cursor does not correctly render at current cursor position.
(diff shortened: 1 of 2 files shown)


Broxzier (changes requested):
Fix works as expected. 👍 Please add an entry for it to the changelog, then this can be merged.

Broxzier:
> One thing I wonder, do we add all the "Fix" to changelog with no exception? I'm just curious because I didn't add it

When it fixes a bug that was present in the previous release, it deserves a changelog entry. If it's a bug that was introduced with the current development branch, it's not.

raacker:
@Broxzier That makes sense. Thanks for letting me know! changelog has been updated 👍

Broxzier (approved):

PR #13239, opened 2020-10-19
Fix #13222 : Vehicle Collision causes negative number of passengers

### LINKED ISSUE #13222: Vehicle collission causing negative number of passengers
**OS:** Linux
**Version:** 0.3.1
**Commit/Build:** 645289f4ec1b1c5ed1475da63190153d99f0b406

After a collision in which everyone died (28), info panel for the ride listed -28 passengers on ride. Save/reload fixed issue.

**Steps to reproduce:**
1. Have fatal crash of all trains on a ride

**Screenshots / Video:**
loopy-lenny
negative-customers

**Save game:**
[ll.zip](https://github.com/OpenRCT2/OpenRCT2/files/5395024/ll.zip)

raacker:
when peep_sprite_remove is called, RemoveFromRide() will decrease ride's num_riders

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -34,6 +34,7 @@
 - Fix: [#13129] Missing error message when waiting for train to leave station on the ride measurements graph.
 - Fix: [#13138] Fix logical sorting of list windows.
 - Fix: [#13158] Cursors are drawn incorrectly in text input fields.
+- Fix: [#13222] Vehicle Collision causes negative number of passengers.
 - Fix: [#13226, #7280] No error is shown when attempting to load a corrupted save.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 - Improved: [#13098] Improvements to the maze construction window user interface
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -34,7 +34,7 @@
 - Fix: [#13129] Missing error message when waiting for train to leave station on the ride measurements graph.
 - Fix: [#13138] Fix logical sorting of list windows.
 - Fix: [#13158] Cursors are drawn incorrectly in text input fields.
-- Fix: [#13222] Vehicle Collision causes negative number of passengers.
+- Fix: [#13222] Vehicle collision causes negative number of passengers.
 - Fix: [#13226, #7280] No error is shown when attempting to load a corrupted save.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
 - Improved: [#13098] Improvements to the maze construction window user interface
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
I think this will need a changelog entry. I'm pretty sure it's an original bug from my look at the assembly.

duncanspumpkin:
Also will need network increment.

raacker:
@duncanspumpkin Could you tell me what do you mean by "network increment"?

duncanspumpkin:
Network.cpp has a version number just increment the number by one. It prevents folk from getting desyncs due to different simulation states as they won't be able to connect with different network versions

tupaschoal:
I think it is `NetworkBase.cpp` now, but I'm not 100% sure

raacker:
@duncanspumpkin I see thanks. Just updated network version number.
@tupaschoal And it was correct. NetworkBase.cpp has NETWORK_STREAM_VERSION

@Broxzier Hey I also added this fix to changelog because it has been there for a while even before 3.1.0. Is it correct then?

Broxzier:
> @Broxzier Hey I also added this fix to changelog because it has been there for a while even before 3.1.0. Is it correct then?

Yep!

distribution/changelog.txt
```diff
@@ -33,6 +33,7 @@
 - Fix: [#13129] Missing error message when waiting for train to leave station on the ride measurements graph.
 - Fix: [#13138] Fix logical sorting of list windows.
 - Fix: [#13158] Cursors are drawn incorrectly in text input fields.
+- Fix: [#13222]  Vehicle Collision causes negative number of passengers.
```
AaronVanGeffen, line 36: Stray space:
```suggestion
- Fix: [#13222] Vehicle Collision causes negative number of passengers.
```

raacker: Thanks for the good catch :)

distribution/changelog.txt
```diff
@@ -34,6 +34,7 @@
 - Fix: [#13129] Missing error message when waiting for train to leave station on the ride measurements graph.
 - Fix: [#13138] Fix logical sorting of list windows.
 - Fix: [#13158] Cursors are drawn incorrectly in text input fields.
+- Fix: [#13222] Vehicle Collision causes negative number of passengers.
```
Gymnasiast, line 37:
```suggestion
- Fix: [#13222] Vehicle collision causes negative number of passengers.
```

distribution/changelog.txt
```diff
@@ -34,6 +34,7 @@
 - Fix: [#13129] Missing error message when waiting for train to leave station on the ride measurements graph.
 - Fix: [#13138] Fix logical sorting of list windows.
 - Fix: [#13158] Cursors are drawn incorrectly in text input fields.
+- Fix: [#13222] Vehicle collision causes negative number of passengers.
```
Gymnasiast, line 37: Aw, shite, should have added ` (original bug)`. :/

PR #13243, opened 2020-10-19
Close  #4865: Offer an option to disable inhibiting the monitor power

### LINKED ISSUE #4865: Offer an option to disable inhibiting the monitor power
**OS:** Ubuntu 16.04 (could be for all operating systems
**Commit/Build:** d1d6be0b9316fd8a0322024b55ef5fbb0f3f7d7f

If the player leaves a computer with this game running for a while, upon reaching any specified screensaver times, the screensaver never seems to activate, which can cause a monitor to be left on for the entire time until the user comes to play again. It seems that this has been left here for the entire timeline of this game's development.

1.  Launch the program,
2.  Wait for the amount of time the screensaver or monitor power off activates.
3. The screensaver and/or the monitor turning itself off never seems to happen.

Lukasz-Pe:
It works, but I have an unknown string instead of option text and thus as draft. After correction and resolving of eventual comments will be changed to normal PR.

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3649,6 +3649,7 @@ STR_6392    :Could not find {STRING} at this path.
 STR_6393    :Objective Selection
 STR_6394    :Objective
 STR_6395    :Maintenance
+STR_6396    :Disable Screensaver lock
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3649,7 +3649,8 @@ STR_6392    :Could not find {STRING} at this path.
 STR_6393    :Objective Selection
 STR_6394    :Objective
 STR_6395    :Maintenance
-STR_6396    :Disable Screensaver lock
+STR_6396    :Disable screensaver
+STR_6397    :{SMALLFONT}{BLACK}Disables screensaver{NEWLINE}If checked screensaver will be inhibited.
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3649,8 +3649,8 @@ STR_6392    :Could not find {STRING} at this path.
 STR_6393    :Objective Selection
 STR_6394    :Objective
 STR_6395    :Maintenance
-STR_6396    :Disable screensaver
-STR_6397    :{SMALLFONT}{BLACK}If checked, screensaver functionality will be inhibited while OpenRCT2 is running.
+STR_6396    :Disable screensaver and monitor power saving
+STR_6397    :{SMALLFONT}{BLACK}If checked, screensaver and other monitor power saving features will be inhibited while OpenRCT2 is running.
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3650,7 +3650,7 @@ STR_6393    :Objective Selection
 STR_6394    :Objective
 STR_6395    :Maintenance
 STR_6396    :Disable screensaver
-STR_6397    :{SMALLFONT}{BLACK}Disables screensaver{NEWLINE}If checked screensaver will be inhibited.
+STR_6397    :{SMALLFONT}{BLACK}If checked, screensaver functionality will be inhibited while OpenRCT2 is running.
 
 #############
 # Scenarios #
--- a/data/language/pl-PL.txt
+++ b/data/language/pl-PL.txt
@@ -3647,7 +3647,6 @@ STR_6390    :Do działania OpenRCT2 wymaga plików z orginalnego RollerCoaster T
 STR_6391    :Wybierz katalog z RCT2
 STR_6392    :Nie można znaleźć {STRING} w tej ścieżce
 STR_6393    :Wybór celu
-STR_6396    :Wyłącz blokadę wygaszacza ekranu
 
 #############
 # Scenarios #
--- a/data/language/pl-PL.txt
+++ b/data/language/pl-PL.txt
@@ -3647,6 +3647,7 @@ STR_6390    :Do działania OpenRCT2 wymaga plików z orginalnego RollerCoaster T
 STR_6391    :Wybierz katalog z RCT2
 STR_6392    :Nie można znaleźć {STRING} w tej ścieżce
 STR_6393    :Wybór celu
+STR_6396    :Wyłącz blokadę wygaszacza ekranu
 
 #############
 # Scenarios #
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -669,6 +669,8 @@ private:
             SDLException::Throw("SDL_CreateWindow(...)");
         }
 
+        gConfigGeneral.enable_screen_saver?SDL_EnableScreenSaver():SDL_DisableScreenSaver();
+
         SDL_SetWindowMinimumSize(_window, 720, 480);
         SetCursorTrap(gConfigGeneral.trap_cursor);
         _platformUiContext->SetWindowIcon(_window);
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -669,7 +669,7 @@ private:
             SDLException::Throw("SDL_CreateWindow(...)");
         }
 
-        gConfigGeneral.enable_screen_saver?SDL_EnableScreenSaver():SDL_DisableScreenSaver();
+        gConfigGeneral.disable_screensaver_lock ? SDL_DisableScreenSaver() : SDL_EnableScreenSaver();
 
         SDL_SetWindowMinimumSize(_window, 720, 480);
         SetCursorTrap(gConfigGeneral.trap_cursor);
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -669,7 +669,7 @@ private:
             SDLException::Throw("SDL_CreateWindow(...)");
         }
 
-        gConfigGeneral.disable_screensaver_lock ? SDL_DisableScreenSaver() : SDL_EnableScreenSaver();
+        disable_screensaver_lock();
 
         SDL_SetWindowMinimumSize(_window, 720, 480);
         SetCursorTrap(gConfigGeneral.trap_cursor);
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -669,7 +669,7 @@ private:
             SDLException::Throw("SDL_CreateWindow(...)");
         }
 
-        ToggleScreenSaverLock();
+        ApplyScreenSaverLockSetting();
 
         SDL_SetWindowMinimumSize(_window, 720, 480);
         SetCursorTrap(gConfigGeneral.trap_cursor);
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -669,7 +669,7 @@ private:
             SDLException::Throw("SDL_CreateWindow(...)");
         }
 
-        disable_screensaver_lock();
+        ToggleScreensaverLock();
 
         SDL_SetWindowMinimumSize(_window, 720, 480);
         SetCursorTrap(gConfigGeneral.trap_cursor);
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -669,7 +669,7 @@ private:
             SDLException::Throw("SDL_CreateWindow(...)");
         }
 
-        ToggleScreensaverLock();
+        ToggleScreenSaverLock();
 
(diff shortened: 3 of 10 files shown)


data/language/pl-PL.txt
```diff
@@ -3647,6 +3647,7 @@ STR_6390    :Do działania OpenRCT2 wymaga plików z orginalnego RollerCoaster T
 STR_6391    :Wybierz katalog z RCT2
 STR_6392    :Nie można znaleźć {STRING} w tej ścieżce
 STR_6393    :Wybór celu
+STR_6396    :Wyłącz blokadę wygaszacza ekranu
```
AaronVanGeffen, line 3650: For organisational reasons, please don't modify any language files but en-GB in this repo.

src/openrct2-ui/windows/Options.cpp
```diff
@@ -88,6 +88,7 @@ enum WINDOW_OPTIONS_WIDGET_IDX {
     WIDX_MULTITHREADING_CHECKBOX,
     WIDX_USE_VSYNC_CHECKBOX,
     WIDX_MINIMIZE_FOCUS_LOSS,
+    WIDX_ENABLE_SCREEN_SAVER,
```
AaronVanGeffen, line 91: Your string says "disable" but the widget uses "enable" internally. This should be made consistent throughout.

Lukasz-Pe: I will change the name to disable_screensaver_lock, as this seems clearer to me.
What do you think about that?

data/language/pl-PL.txt
```diff
@@ -4467,4 +4467,4 @@ STR_DTLS    :Na twoim terenie odbywa się wielki doroczny festiwal muzyczny. Wyb
 <Rock 'n' Roll - Rock 'n' Roll>
 STR_SCNR    :Rock 'n' Roll - Rock 'n' Roll
 STR_PARK    :Rock 'n' Roll Revival
-STR_DTLS    :Ten starzejący się lunapark ma najlepsze dni za sobą. Pomóż właścicielom nadać mu stylowy, rockandrollowy wygląd i przekształć to miejsce w dochodowe przedsięwzięcie.
\ No newline at end of file
+STR_DTLS    :Ten starzejący się lunapark ma najlepsze dni za sobą. Pomóż właścicielom nadać mu stylowy, rockandrollowy wygląd i przekształć to miejsce w dochodowe przedsięwzięcie.
```
AaronVanGeffen, line 4470: This EOL change will need to be reverted as well, otherwise we might still run into problems with the nightly localisation merge.

Lukasz-Pe: I've checked this file in repository and this line looks exactly the same. I will re-upload this file from original repository.

src/openrct2-ui/windows/Options.cpp
```diff
@@ -636,6 +639,11 @@ static void window_options_display_mouseup(rct_window* w, rct_widgetindex widget
             config_save_default();
             w->Invalidate();
             break;
+        case WIDX_DISABLE_SCREENSAVER_LOCK:
+            gConfigGeneral.disable_screensaver_lock ^= 1;
+            config_save_default();
+            w->Invalidate();
+            break;
```
AaronVanGeffen, line 646: I don't believe this is enough to actually toggle the screensaver if the relevant routine is only invoked in `CreateWindow`. Perhaps split the invocation off to a separate function, and call that function from here, too?

Lukasz-Pe: After toggling the game has to be restarted, otherwise it won't have any effect.
Toggling this option here requires including SDL.h in Options.cpp file. My solution will follow soon.

AaronVanGeffen: > After toggling the game has to be restarted, otherwise it won't have any effect.

Yes, it doesn't have any effect without restarting in its current state… However, if that's intended, you should at least present a warning to the user.

> Toggling this option here requires including SDL.h in Options.cpp file. My solution will follow soon.

Instead of invoking any of the SDL functions directly, you can expose a new method on e.g. the context object. (Though do keep in mind it can't be available outside of the `openrct2-ui` scope.)

Lukasz-Pe: I hope the solution presented resembles what you meant.

src/openrct2-ui/windows/Options.cpp
```diff
@@ -221,6 +222,7 @@ static rct_widget window_options_display_widgets[] = {
     MakeWidget        ({155, 176}, {136,  12}, WWT_CHECKBOX, WindowColour::Secondary, STR_MULTITHREADING,                    STR_MULTITHREADING_TIP                   ), // Multithreading
     MakeWidget        ({ 11, 176}, {143,  12}, WWT_CHECKBOX, WindowColour::Secondary, STR_USE_VSYNC,                         STR_USE_VSYNC_TIP                        ), // Use vsync
     MakeWidget        ({ 11, 191}, {280,  12}, WWT_CHECKBOX, WindowColour::Secondary, STR_MINIMISE_FULLSCREEN_ON_FOCUS_LOSS, STR_MINIMISE_FULLSCREEN_ON_FOCUS_LOSS_TIP), // Minimise fullscreen focus loss
+    MakeWidget        ({ 11, 206}, {280,  12}, WWT_CHECKBOX, WindowColour::Secondary, STR_ENABLE_SCREENSAVER,                STR_ENABLE_SCREENSAVER                   ), // Enable screensaver
```
tupaschoal, line 225: Should we have a tooltip with a bit more information?

tupaschoal (commented):
Please add screenshots of how it looks in the options menu

src/openrct2-ui/windows/Options.cpp
duncanspumpkin: Also tooltips (for now) need to specify that their size and colour so should always be a separate string.

Lukasz-Pe:
As mentioned in first message, I've got undefined string instead of right one.
The config below will inhibit turning screensaver on.
OpenRCT2

AaronVanGeffen:
Is the `data` folder in your build folder set up as as a symlink to the repository's?

Lukasz-Pe:
There was a hardcopy of this dir in my build directory. After correction it looks better.
Thanks @AaronVanGeffen !
OpenRCT2_2020-10-20 11-34-56

src/openrct2-ui/windows/Options.cpp
Lukasz-Pe: Corrected. Please see second image attached.

src/openrct2-ui/interface/Window.cpp
```diff
@@ -536,6 +537,11 @@ void window_all_wheel_input()
         }
     }
 }
+void toggle_screensaver_lock()
```
duncanspumpkin, line 540: All new functions are to be TitleCase. (Really we need to start a big renaming project to make everything use TitleCase)

data/language/en-GB.txt
```diff
@@ -3649,6 +3649,8 @@ STR_6392    :Could not find {STRING} at this path.
 STR_6393    :Objective Selection
 STR_6394    :Objective
 STR_6395    :Maintenance
+STR_6396    :Disable screensaver
+STR_6397    :{SMALLFONT}{BLACK}If checked, screensaver functionality will be inhibited while OpenRCT2 is running.
```
AaronVanGeffen, line 3653: Perhaps this should be amended to mention power saving as well, considering that was the original issue.

AaronVanGeffen (commented):
I've pushed a few minor changes, rather than adding pedantic comments. There are two more final issues to be resolved, as far as I'm concerned:

* Perhaps the strings should indicate this affects screen power saving functionality as well
* A changelog entry will need to be added

data/language/en-GB.txt
Lukasz-Pe: Maybe something like this:
STR_6397    :{SMALLFONT}{BLACK}If checked, screensaver functionality will be inhibited while OpenRCT2 is running.{NEWLINE}Unchecking allows screen saver to run and (if in system enabled) turn off the monitor.

Lukasz-Pe: and the main option string:
STR_6396    :Disable screensaver and monitor power saving mode

AaronVanGeffen: I was thinking among the lines of:
```suggestion
STR_6396    :Disable screensaver and monitor power saving
STR_6397    :{SMALLFONT}{BLACK}If checked, screensaver and other monitor power saving features will be inhibited while OpenRCT2 is running.
```
It would be good to wait for some input on this from others, I think.

Lukasz-Pe: Meanwhile screenshot to have a look at those changes:
OpenRCT2_2020-10-20 14-32-15

src/openrct2-ui/UiContext.cpp
```diff
@@ -669,7 +669,7 @@ class UiContext final : public IUiContext
             SDLException::Throw("SDL_CreateWindow(...)");
         }
 
-        ToggleScreensaverLock();
+        ToggleScreenSaverLock();
```
Broxzier, line 672: 'Toggle' suggests that it's changing the lock, but this function is only applying the setting, right?

data/language/en-GB.txt
AaronVanGeffen: I prefer the look of the first of the two screenshots. (Translations will look different, anyway, of course...)

src/openrct2-ui/UiContext.cpp
Lukasz-Pe: It actually changes the state of the lock. If you begin with enabled it will change setting to disabled and so on.

Broxzier: `gConfigGeneral.disable_screensaver` is read but not changed. It also wouldn't really make sense to toggle the setting when opening the window.

data/language/en-GB.txt
Lukasz-Pe: I'll wait with commit till discussion regarding `ToggleScreenSaverLock();` will be resolved, to stop commit spam :)

src/openrct2-ui/UiContext.cpp
Lukasz-Pe: `gConfigGeneral.disable_screensaver` is changed in lines 642 - 647 of `src/openrct2-ui/windows/Options.cpp`:
```
        case WIDX_DISABLE_SCREENSAVER_LOCK:
            gConfigGeneral.disable_screensaver ^= 1;
            ToggleScreenSaverLock();
            config_save_default();
            w->Invalidate();
            break;
```
I followed also a suggestion by @AaronVanGeffen to export this logic to new method and invoke it in relevant places.

The problem here (as I understand it) is that SDL by default inhibits screensaver and monitor power saving.
Leaving this function only in options window logic will cause to read this value and change related setting only if someone goes to options window and (un)checks the checkbox.
Without this, independently of what is saved in config.ini file, this setting won't be applied to the game at the start and thus screensaver will be always inhibited.

AaronVanGeffen: I think @Broxzier is suggesting the name `ToggleScreenSaverLock` implies a setting change, rather than applying the current setting. In that sense, I agree something like `ApplyScreenSaverLock` might be more appropriate. However, at the same time, SDL is still toggling the screensaver inhibition status. I therefore feel a name change is not necessarily needed.

Broxzier: > However, at the same time, SDL is still toggling the screensaver inhibition status.

That doesn't sound right. Shouldn't this be called on startup already rather than when the options window gets opened?

>  I therefore feel a name change is not necessarily needed.

The function calls the corresponding SDL function, the toggling happens in the user input handler. How I see it, 'toggle' doesn't make sense here.

AaronVanGeffen: > That doesn't sound right. Shouldn't this be called on startup already rather than when the options window gets opened?

Perhaps there is a misunderstanding here. It is called when the game starts: in `UiContext`'s `CreateWindow`. The only time it's called in the Options window, is when you click the relevant checkbox, in order to change the state.

Broxzier: > Perhaps there is a misunderstanding here. It is called when the game starts: in UiContext's CreateWindow. The only time it's called in the Options window, is when you click the relevant checkbox, in order to change the state.

Ah! That explains the confusion, I was indeed thinking these changes were made in the options window.

> Again, I don't feel strongly about this. If you think it needs changing, let's.

Alrighty. It's cleaner imo, identifiers should reflect what they do, not how they are used.

Lukasz-Pe: > Alrighty. It's cleaner imo, identifiers should reflect what they do, not how they are used.
Looking from that perspective maybe renaming to `ApplyScreenSaverLockStatusChange` should be more descriptive than Toggle.

AaronVanGeffen (approved):
LGTM as of d81ab5a8c27cd5e3f07175e239fc2990303450dc

PR #13247, opened 2020-10-20
Refactor Peep Bin Sub State #12391

rpstester:
Changed it to enum class and updated references.


tupaschoal (approved):

rpstester:
Wow! That was fast! Thanks!

PR #13249, opened 2020-10-20
Fix #13236: Researching new ride type appears like new vehicle type

### LINKED ISSUE #13236: Researching new ride type results in "New vehicle now available" message instead
**OS:** Windows 10
**Version:** 0.3.1
**Commit/Build:** 59d785e

I have noticed that in some scenarios, when a new ride type is researched, it does not even send me a message saying "New ride/attraction now available: [ride name]", but instead "New vehicle now available for [ride name]: [vehicle name]".

**EDIT (2020-10-19 18:12 PST):** tupaschoal has confirmed that the research window also shows the wrong info.

**Steps to reproduce:**
1. Load up the [scenario file shown in the screenshot below](https://drive.google.com/file/d/1Ky7mKdSILs3rCXrg2ibvEj7ya3e3Dhr1/view?usp=sharing) and wait until March 30, Year 1.
2. The Air Powered Vertical Coaster will be researched, but the game will think you've researched a new vehicle type instead of a new ride type. Researching the Looping Roller Coaster (which comes afterwards) still correctly results in the "New ride/attraction now available: Looping Roller Coaster" message though.
3. I have also been able to reproduce this by researching a new ride type, modifying the inventions list to move said ride type back to the "Items to invent during game" list, and then having said ride type be researched again. This can be done with any scenario, not just custom ones.

**Screenshots / Video:**
Extreme Heights 2 2020-10-18 15-00-52
Note that researching the Air Powered Vertical Coaster in the original Extreme Heights scenario still brings up the correct "New ride/attraction now available: Air Powered Vertical Coaster" message.

tupaschoal:
There are a couple of bugs here:
- `research_update_first_of_type` only checks if a ride has been seen or not and updates the flag accordingly, but it was erroneously also marking this ride as seen, so every uninvented item would be marked as such and there was never a new ride type, just new vehicles.
- Regardless of the above, `research_mark_ride_type_as_seen` was being called for the item still to be invented
- Regarding the code for the last item, it seems like it should be marked as seen first, then the first of type would be updated, but I didn't want to touch that for now, given the comments on the for loop above it (which seem weird for `gResearchLastItem`)
```
    if (gResearchLastItem.has_value())
    {
        research_update_first_of_type(&gResearchLastItem.value());
        research_mark_ride_type_as_seen(gResearchLastItem.value());
    }
```

I did some basic testing with the provided save and it fixed the issue.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -40,6 +40,7 @@
 - Fix: [#13158] Cursors are drawn incorrectly in text input fields.
 - Fix: [#13222] Vehicle collision causes negative number of passengers (original bug).
 - Fix: [#13226, #7280] No error is shown when attempting to load a corrupted save.
+- Fix: [#13236] Researching new ride type appears like new vehicle type instead.
 - Fix: [#13266] Plugin API: Deleting key of sharedStorage not working.
 - Fix: [#13278] Desync caused by to ghost tiles changing the ride mode.
 - Improved: [#13023] Made add_news_item console command last argument, assoc, optional.
--- a/src/openrct2/management/Research.cpp
+++ b/src/openrct2/management/Research.cpp
@@ -916,8 +916,6 @@ static void research_update_first_of_type(ResearchItem* researchItem)
 
     if (!_seenRideType[rideType])
         researchItem->flags |= RESEARCH_ENTRY_FLAG_FIRST_OF_TYPE;
-
-    _seenRideType[rideType] = true;
 }
 
 static void research_mark_ride_type_as_seen(const ResearchItem& researchItem)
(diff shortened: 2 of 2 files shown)


Gymnasiast:
I really need to have a good think about why this is going wrong.

Because it does make sense to mark even uninvented items as "seen". For example, imagine that you have a scenario where the Car Ride has not been invented yet. In the "uninvented" list are the Sports Cars and the Racing Cars. If you don't mark the Car Ride as "seen" after processing the Sports Cars, then it will report the Racing Cars as "Car Ride" instead of "Racing Cars for Car Ride".

tupaschoal:
> If you don't mark the Car Ride as "seen" after processing the Sports Cars

But doesn't this mean "Inventing" it or moving it to inventedList/lastInvention? Because if so when this happens it is marked as seen

tupaschoal:
Did you get a chance to look at this @Gymnasiast ? Or anyone?

Gymnasiast:
Not yet. :/

Gymnasiast:
I have done some more tests. It works correctly for newly started scenarios. I think the problem might lie with how it handles the previous/next research items - because those are generally still clean on newly started scenarios.

Gymnasiast:
https://github.com/OpenRCT2/OpenRCT2/pull/13841

PR #13258, opened 2020-10-21
Add Execute and FindApp to Platform class

keithstellyes:
Split off from this PR https://github.com/OpenRCT2/OpenRCT2/pull/13172

This adds `::Execute` and `::FindApp` to the Platform class, as it seems like that should be brought up, and because pre-existing logic was found in `UiContext.Linux` when it was needed outside the `UiContext`. There is equivalent logic in the Windows-specific code too.

HOLD for merge until I get confirmation the code still builds and runs on platforms other than Linux


tupaschoal (commented):
It looks alright, let us know when you get the confirmation

PR #13216, opened 2020-10-16
Close #12418: Refactor RESEARCH_CATEGORY_* to use strong enum

Lukasz-Pe:

PR #13270, opened 2020-10-23
Fix #13263: SceneryGroup problems

### LINKED ISSUE #13263: Scenery group research issues
**OS:** Windows 10
**Version:** 0.3.1
**Commit/Build:** 214f699

A regression bug has now made it impossible to select or deselect "Scenery & Themeing" from the research priority window even if scenery groups can be researched in the scenario. In addition, viewing scenery groups in the Inventions list wrongly shows "ALT +" as their research group.

**Steps to reproduce:**
1. Open the research window and open the tab that allows you to change research funding and priorities. You have to do this in a scenario where scenery groups can be researched (in this case, a custom scenario like the one shown below).
2. The "Scenery & Themeing" tab is now always grayed out.
3. Open the inventions list and hover over any scenery group.
4. You should see "ALT +" as said scenery group's research group.

**Screenshots:**
Antediluvian Antics 2020-10-22 11-54-14
Antediluvian Antics 2020-10-22 12-00-48
These screenshots are from my scenario [Antediluvian Antics](https://drive.google.com/file/d/1IE3Ma6imQfn4WafCEPxy9167MONAEEDy/view?usp=sharing).

tupaschoal:
- First I just started reading some values as `enum`s, rather than flags, like they were converted to in #13216
- I also renamed the enum back from `Group` to `SceneryGroup`

### DIFF
--- a/src/openrct2/management/Research.cpp
+++ b/src/openrct2/management/Research.cpp
@@ -519,7 +519,7 @@ bool research_insert_scenery_group_entry(ObjectEntryIndex entryIndex, bool resea
     if (entryIndex != OBJECT_ENTRY_INDEX_NULL)
     {
         auto tmpItem = ResearchItem(
-            Research::EntryType::Scenery, entryIndex, RIDE_TYPE_NULL, EnumToFlag(ResearchCategory::Group), 0);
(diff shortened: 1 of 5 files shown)


AaronVanGeffen (approved):
Second commit message has a typo. LGTM otherwise!

duncanspumpkin:
Should ResearchItem be using the new type?

duncanspumpkin (approved):

PR #13272, opened 2020-10-23
Fix #13266: Plugin API: Deleting key of sharedStorage not working

### LINKED ISSUE #13266: bug in plugin API: deleting key of sharedStorage not working
ScConfiguration.hpp L 254
`duk_del_prop_string(ctx, -1, key.c_str());`
is not working as intended and should be changed to
`duk_del_prop_lstring(ctx, -1, n.data(), n.size());`

Sadret:


IntelOrca (approved):

duncanspumpkin:
Can you adjust the commit message so that it is `Fix: #13266: Plugin API Deleting key of shared storage not working`

PR #13276, opened 2020-10-24
closes #12415 CURRENCY_TYPE enum refactor

### LINKED ISSUE #12415: Refactor CURRENCY_TYPE to use strong enum

BryanDiLaura:
#12415 CURRENCY_TYPE enum refactor to strong enum


BryanDiLaura:
Note this was a pretty surface-level refactor at this point. If desired, it may make sense to also refactor the config value to leverage the strong enum, but it would require more invasive surgery. Let me know!

src/openrct2/config/Config.cpp
```diff
@@ -50,24 +50,24 @@ namespace Config
     });
 
     static const auto Enum_Currency = ConfigEnum<int32_t>({
```
duncanspumpkin, line 52: type should be changed to `CurrencyType`

src/openrct2/localisation/Currency.h
```diff
@@ -13,29 +13,29 @@
...
+    HKD,          // Hong Kong Dollar
+    TWD,          // New Taiwan Dollar
+    Yuan,         // Chinese Yuan
+    Forint,       // Hungarian Forint
 
-    CURRENCY_CUSTOM, // Custom currency
+    Custom, // Custom currency
 
-    CURRENCY_END // Last item
+    End // Last item
```
duncanspumpkin, line 38: All other enums use `Count` instead of end to indicate the count of the enum. Please change the name.

src/openrct2/platform/Shared.cpp
```diff
@@ -220,18 +220,18 @@ uint8_t platform_get_currency_value(const char* currCode)
 {
```
duncanspumpkin, line 220: This function should return a `CurrencyType`

tupaschoal (approved):

Issue #13060, opened 2020-10-01
Add CMake requirement for nlohmann-json


AaronVanGeffen:
Since #12667, OpenRCT2 has used the JSON for Modern C++ library (`nlohmann-json`) to do its JSON processing. However, the availability of this library is not checked by CMake. This leads to unexpected errors when compiling if the package is not present.

We should add a directive to check for `nlohmann-json` to our main `CMakeLists.txt` file.

For inspiration, see https://nlohmann.github.io/json/integration/cmake/

tupaschoal:
Up for grabs

raacker:
@tupaschoal May I take over this? :)

tupaschoal:
Sure

raacker:
@AaronVanGeffen Hi, I got some question to solve this.

It actually started from "why current PRs' don't fail with CI even though they don't have nlohman json" and I found that **Win64 and Win32** are downloading 3rd party dependencies by [openrct2.proj](https://github.com/OpenRCT2/OpenRCT2/blob/develop/openrct2.proj#L40) file. But Linux, MacOS and Mingw version are not. 

So can I say nlohman dependency of cmake is only needed for those environments?

And also, I found that [Building OpenRCT2 on Linux](https://github.com/OpenRCT2/OpenRCT2/wiki/Building-OpenRCT2-on-Linux#required-packages-general) page has legacy json info for Fedora, Gentoo and etc.

AaronVanGeffen:
@raacker Thanks for the interest in this issue. 

For Windows and macOS, we manage dependencies using other package managers. Periodically, we release bundles in our own Dependencies repository.

For Linux and BSD platforms, we rely on the platform's package manager to do this job. Indeed, want to continue to do this, even for nlohmann-json. Retrieving a bundle for the project sounds technically interesting, but I don't think it's the best solution here.

raacker:
@AaronVanGeffen I see. But we still support cmake build for macOS right? the xcode build project is downloading zip files separately but not the cmake one.

If then, I will still add nlohman-json check for cmake but then have no idea how we can solve CI's build problem here. I might need to change CI env setup script also

AaronVanGeffen:
> @AaronVanGeffen I see. But we still support cmake build for macOS right? the xcode build project is downloading zip files separately but not the cmake one.

Oh, yes, we do still support cmake builds on macOS. That's what I use myself, most of the time, in fact! 🙂

[...]

PR #13277, opened 2020-10-24
Add compile-time check for nlohmann json version

keithstellyes:
Currently, the version of nlohmann json distributed with [certain Linux distros](https://repology.org/project/nlohmann-json/versions) is <3.6, notably this includes Raspbian.
A user following this guide for example: https://forums.openrct2.org/topic/5145-building-openrct2-on-rpi4-with-raspberry-pi-os/ will hit compile errors, due to the version being distributed is too old (3.5).

Just a bit ago was helping a user on Discord who following aforementioned guide wa shaving build errors caused by too old a version being distributed with Raspberry Pi OS.

### DIFF
--- a/src/openrct2/core/Json.hpp
+++ b/src/openrct2/core/Json.hpp
@@ -15,6 +15,10 @@
 #include <string>
 #include <string_view>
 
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This should be handled by CMake, imo.

janisozaur:
While I agree with @AaronVanGeffen , I think it's fine to have at least this check for now and it won't cause cause issues when cmake one is finally done.

janisozaur:
For reference: https://github.com/OpenRCT2/OpenRCT2/issues/13060

PR #13278, opened 2020-10-24
Fix desync caused due to ghost tiles changing the ride mode

manuelVo:
Steps to reproduce the desync:
1. Build a track starting with a chain lift
2. Somewhere after the chain lift place a block brake (don't actually build it. Just hover it so the ghost tile appears.)
3. Instead of the blockbrake place a regular straight tile
4. Complete the track

What happens is: When hovering the block brake, the operation mode of the ride will be set to block mode. (Only for the client building the ride though since the ghost tile isn't synchronized). When the block brack ghost is replaced by the ghost of a regular tile the mode isn't reset to circuit mode, since there still exist two blocks (due to the chain lift existing). After finishing the track the mode will be set to block mode for the builder and to circuit mode for all other clients. The desync window will appear when testing/opening the ride as soon as differences emerge due to the different mode.

This PR fixes that (and some related desyncs) by preventing the ghost tile from changing the operation mode. I've been a bit liberal with excluding code for ghost tiles and made it affect other code in the vicinity that appeared to be likely to cause unwanted side effects as well. However so far I've been able to find actual desyncs caused by the code within `case TrackElemType::BlockBrakes:`

### DIFF
--- a/src/openrct2/actions/TrackPlaceAction.hpp
+++ b/src/openrct2/actions/TrackPlaceAction.hpp
@@ -546,43 +546,46 @@ public:
 
             cost += ((supportHeight / (2 * COORDS_Z_STEP)) * RideTypeDescriptors[ride->type].BuildCosts.SupportPrice) * 5;
 
-            invalidate_test_results(ride);
-            switch (_trackType)
+            if (!(GetFlags() & GAME_COMMAND_FLAG_GHOST))
             {
-                case TrackElemType::OnRidePhoto:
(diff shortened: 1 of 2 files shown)


Gymnasiast (commented):
Code LGTM. (Hint: you can tell Github to ignore whitespace changes, which makes for far easier reviewing here.)

This definitely warrants a changelog entry:
```
Fix: [#13278] Desync caused by ghost tiles changing the ride mode.
```

 It will also need a network version update.

duncanspumpkin:
I know I never got round to reviewing it but this looks good. Sorry was busy on other things.

PR #13280, opened 2020-10-24
Fix crash when using a writable FileStream to a file in non-existing folder

Broxzier:
Something that I ran in to long ago, but got stuck at the time on an annoying issue with Windows's `CreateDirectory` macro. The second commit fixes this.

This change makes sure the directory exists before `_wfopen` is called.

### DIFF
--- a/src/openrct2/core/FileStream.hpp
+++ b/src/openrct2/core/FileStream.hpp
@@ -12,6 +12,7 @@
 #include "../common.h"
 #include "../localisation/Language.h"
 #include "IStream.hpp"
+#include "Path.hpp"
 #include "String.hpp"
 
 #include <algorithm>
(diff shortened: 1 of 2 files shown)


src/openrct2/core/FileStream.hpp
```diff
@@ -81,6 +82,16 @@ namespace OpenRCT2
                     throw;
             }
 
+            // Make sure the directory exists before writing to a file inside it
+            if (_canWrite)
+            {
+                std::string directory = Path::GetDirectory(path);
+                if (!Path::DirectoryExists(directory))
+                {
+                    Path::CreateDirectory(directory);
```
tupaschoal, line 91: Shouldn't we check the return and throw in case this fails?

Broxzier: `Path::CreateDirectory` is void and doesn't throw from the looks of it. It itself does call a function returning bool, however. It's probably a good idea to validate the return value there and return false or throw on failure.

tupaschoal: I don't mind if this is left for another PR, as it will take changing multiple places.

tupaschoal (commented):
Does this fix any open issue? Probably could use a changelog

Broxzier:
> Does this fix any open issue? Probably could use a changelog

I highly doubt it. With the current codebase, all folders for where FileStream gets currently used are created when launching the game. With this change, any future uses of FileStream (to files in folders that don't exist yet) would be working without crashing.

tupaschoal:
Okz then I guess it's fine to merge without changelog

PR #13281, opened 2020-10-24
Refactor LIGHTFX_LIGHT_QUALIFIER #12406

rpstester:
Now using strong enum (along with EnumValue)
Please let me know if you have any questions or feedback. Thanks!

### DIFF
--- a/src/openrct2/drawing/LightFX.cpp
+++ b/src/openrct2/drawing/LightFX.cpp
@@ -699,13 +699,17 @@ void lightfx_add_lights_magic_vehicle(const Vehicle* vehicle)
     {
         case RIDE_TYPE_OBSERVATION_TOWER:
             lightfx_add_3d_light(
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
Anytime when you do one of these refactors and you have to use the enumvalue function everywhere it suggests things should be refactored. I'd like to see a few changes to the light fx code so that it takes this new type and uses it correctly.

rpstester:
Thank you for the feedback. I was looking at approved refactor work and saw EnumValue used over and over again.  Would you please point me in the right direction of how you would like this enum class used properly, or give me an example to study?

tupaschoal:
There are some cases where we'll need to use `EnumValue`, such as indexing into an array or generic window/tab code. Function parameters, member variables, normal variables and other stuff can instead be converted to use the actual enum type instead, making the code safer, clearer and with fewer casts/conversions

rpstester:
This one looks a little beyond my skill level.  Someone else can take it on :)

PR #13286, opened 2020-10-24
Small research releated refactor

tupaschoal:
Might be clearer to look commit by commit
- Uses `ResearchCategory` as enum throughout
- Moves Research related string stuff into `ResearchItem`
- Add converter from `RideCategory` to `ResearchCategory` to reduce the number of casts (and prevent a possible number mismatch, in case those are ever reordered)

### DIFF
--- a/src/openrct2/ride/RideData.cpp
+++ b/src/openrct2/ride/RideData.cpp
@@ -23,6 +23,7 @@
 #include "../audio/audio.h"
 #include "../interface/Colour.h"
 #include "../localisation/Localisation.h"
+#include "../management/Research.h"
 #include "../sprites.h"
 #include "Ride.h"
 #include "ShopItem.h"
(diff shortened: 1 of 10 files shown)


duncanspumpkin (approved):

src/openrct2/ride/RideData.cpp
```diff
@@ -329,3 +330,26 @@ bool RideTypeDescriptor::SupportsTrackPiece(const uint64_t trackPiece) const
 {
     return GetAvailableTrackPieces() & (1ULL << trackPiece);
 }
+
+ResearchCategory RideTypeDescriptor::RideToResearchCategory() const
+{
```
Gymnasiast, line 335: Call this GetResearchCategory()

Gymnasiast:
What is the performance of using an unordered_map compared to an array?

AaronVanGeffen:
> What is the performance of using an unordered_map compared to an array?

According to [cppreference](https://en.cppreference.com/w/cpp/container/unordered_map) (emphasis mine):
> Search, insertion, and removal of elements have average constant-time complexity. 

So that should be fine.

IntelOrca:
It still won't be as fast as an array. It would be nice if we had a data structure which used the same implementation as a switch. Have a 1:1 lookup table within the bounds of the lowest and highest value.

duncanspumpkin:
Could do with the same structure for openloco

Gymnasiast:
Considering @IntelOrca 's comment, I think that an array would be more appropriate. After all, the source values are all adjacent.

tupaschoal:
It's true that though the map is supposed to be constant time, it will be slower than the array. I want to prevent us from relying on the ordering of the enum. If it ever receives another value or it's reordered, the result will be completely wrong. I'd vouch for the `switch...case`, according to quick bench it has the same performance as the array (actually a bit faster):
https://quick-bench.com/q/Jwwxqi2Gq1MAUwNAkP5MYRGCZAQ
image

tupaschoal:
Changed to a switch based on the result above

ZehMatt:
I'm curious to how a switch is faster than the array. switch tables usually have some code around them for bounds checking and based on the size of each case branch it may cause cache misses, strange stuff.

tupaschoal:
Would you still be ok with the switch in here anyway?

Gymnasiast:
I still prefer the array, but I'll let @ZehMatt decide.

tupaschoal:
Note that there are benefits to both array and switch, namely speed and "bound checking" respectively

ZehMatt:
I would like the array honestly but its not an absolute must, if you can avoid code you should, a switch is a bit different compared to having it as pure data in the array. The complexity of arrays is always linear, switches do not promise anything.

IntelOrca:
If this were a language like Python or F#, a switch / match would be much more elegant, and I would prefer that over an array because then it can't break if you change the order of the enum values.

Gymnasiast:
OK, it's time to take a decision. Since @ZehMatt and I both prefer an array, let's just go for that - an array with a bounds check to avoid reading invalid memory.

tupaschoal:
Restored the arrays, should be all good now @Gymnasiast

PR #13298, opened 2020-10-25
Refactor GoTo loc_6D7A97

rpstester:
Remove the goto for loc_6D7A97.
Moved the code after loc_6D7A97 to another function; what would be a good name for it?
As usual, please let me know if you have any questions or feedback.

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -5496,6 +5496,20 @@ void Vehicle::UpdateCrash()
         }
     }
 }
+
+void Vehicle::screamSubFunction(Ride* curRide, OpenRCT2::Audio::SoundId screamId, uint8_t screamVolume)
+{
+    scream_sound_id = OpenRCT2::Audio::SoundId::Null;
+    if (curRide->type < std::size(RideTypeDescriptors))
+    {
+        // Get lift hill sound
+        screamId = RideTypeDescriptors[curRide->type].LiftData.sound_id;
+        screamVolume = 243;
+        if (!(sound2_flags & VEHICLE_SOUND2_FLAGS_LIFT_HILL))
+            screamId = OpenRCT2::Audio::SoundId::Null;
+    }
+}
+
 /**
  *
  *  rct2: 0x006D7888
@@ -5534,7 +5548,10 @@ void Vehicle::UpdateSound()
             if (!(gCurrentTicks & 0x7F))
             {
                 if (velocity < 0x40000 || scream_sound_id != OpenRCT2::Audio::SoundId::Null)
-                    goto loc_6D7A97;
+                {
+                    screamSubFunction(curRide, screamId, screamVolume);
+                    break;
+                }
 
                 if ((scenario_rand() & 0xFFFF) <= 0x5555)
                 {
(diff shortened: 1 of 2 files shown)


duncanspumpkin (changes requested):

duncanspumpkin:
I don't think you have tested this please try make sure the changes you make are working. Also the CI's have caught your mistake as well.

tupaschoal (commented):
Make sure you mark the solved comments as such, when you fix the issues.

Also let us know when this is ready by commenting and clicking "Ready for review"

rpstester:
I will. I have not finished yet. Thank you for your patience.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1945,6 +1945,21 @@ static SoundIdVolume sub_6D7AC0(
     return { currentSoundId, currentVolume };
 }
 
+SoundIdVolume Vehicle::GetLiftHillSound(Ride* curRide)
+{
+    SoundIdVolume resultSounds{};
```
rpstester, line 1950: FYI, I gotta come back and change this because it may wipe out the default setting of Null and 255 set in the larger function (old lines 5509-5510).

src/openrct2/ride/Vehicle.cpp
```diff
@@ -5506,8 +5521,9 @@ void Vehicle::UpdateSound()
     uint8_t frictionVolume = 255;
     auto frictionId = OpenRCT2::Audio::SoundId::Null;
     // bh screamVolume should be set before hand
-    auto screamId = OpenRCT2::Audio::SoundId::Null;
-    uint8_t screamVolume = 255;
+    SoundIdVolume currentSound = { OpenRCT2::Audio::SoundId::Null, 255 };
```
rpstester, line 5524: Do you have a preference for the name of this variable? screamSound or currentSound?

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1945,6 +1945,21 @@ static SoundIdVolume sub_6D7AC0(
     return { currentSoundId, currentVolume };
 }
 
+SoundIdVolume Vehicle::GetLiftHillSound(Ride* curRide)
```
rpstester, line 1948: Do you have a preference for the name of this function, since there seems to be an effort to not name it sub_6xxxxx?

src/openrct2/ride/Vehicle.cpp
```diff
@@ -5534,7 +5548,10 @@ void Vehicle::UpdateSound()
             if (!(gCurrentTicks & 0x7F))
```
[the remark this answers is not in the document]
rpstester: Thanks. I think I got it now.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -5530,64 +5542,64 @@ void Vehicle::UpdateSound()
...
             {
-                screamId = UpdateScreamSound();
-                if (screamId == OpenRCT2::Audio::SoundId::NoScream)
-                    screamId = OpenRCT2::Audio::SoundId::Null;
-                else if (screamId == OpenRCT2::Audio::SoundId::Null)
-                    goto loc_6D7A97;
+                currentSound.id = UpdateScreamSound();
+                if (currentSound.id == OpenRCT2::Audio::SoundId::NoScream)
+                    currentSound.id = OpenRCT2::Audio::SoundId::Null;
+                else if (currentSound.id == OpenRCT2::Audio::SoundId::Null)
```
duncanspumpkin, line 5594: You can slightly clean this up by not repeating the get lift hill sound. Just rework the ifs so that it doesn't break.

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: It's ultimately used for setting sound2. So might as well call it `newSound2`

src/openrct2/ride/Vehicle.cpp
rpstester: Done

duncanspumpkin (approved):

PR #13300, opened 2020-10-25
Fix #13019: Prevent ghost trains from opening the construction window when crashing

### LINKED ISSUE #13019: Simulating with block brakes (most noticeable in multiplayer) causes everyone to open the ride edit panel
OS: Windows 10, Arch 5.8.6-1-Manjaro
Version: 0.3.1 (has been for several versions)
Commit/Build: 0201c0b

This is an issue that we have been having for a while now and just figured out tonight what was causing it. A few things need to be true for the issue to happen.

A coaster needs to be in the middle of construction.
Ghost trains need to be turned on.
After the ghost trains are on a block brake has to be placed.
When the ghost train hits the block brake every player will have the construction panel forcibly open for that ride.

This is most noticeable in multiplayer but is reproducible in single player as well if you place the block brake then close the construction panel.

manuelVo:
The forced open of the construction window reported in #13019 happens whenever a ride is in block mode and a ghost train reaches the end of the track (crashes). Usually this shouldn't happen, because starting a simulation in block mode isn't allowed. However such a situation may still occur when the track is being modified while the simulation is already running for two reasons:
1. If a track piece is removed during simulation in block mode (this tries to reset the simulation, which isn't possible due to the block mode. The construction window is probably opened while attempting to display the error)
2. If a block brake is added to a ride that was previously in circuit mode the ride will be switched to block mode without resetting the trains.

I addressed 1 by closing down the ride instead of resetting the simulation if the ride is in block mode. I preferred doing this over closing the ride already when removing a track piece as this allows players to still swap out small pieces during simulation (for example adding a photo section or banked curve).

While 1 would already be enough to fix the reported issue, I also added a second change: Currently when adding a block brake the operation mode is switched without the side effects that a mode switch would have when the mode was switched by the player (like clearing the trains). This means that there might be multiple trains within a single block (which should never happen for block operation mode). As such the simulation wouldn't provide any beneficial insight for the player and I think it's better to just clear out the track entirely. To accomplish this I switch the mode using the game command instead of directly modifying the mode variable.

### DIFF
--- a/src/openrct2/actions/TrackPlaceAction.hpp
+++ b/src/openrct2/actions/TrackPlaceAction.hpp
@@ -18,6 +18,7 @@
 #include "../world/MapAnimation.h"
 #include "../world/Surface.h"
 #include "GameAction.h"
+#include "RideSetSetting.hpp"
 
 class TrackPlaceActionResult final : public GameActions::Result
 {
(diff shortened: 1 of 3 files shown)


duncanspumpkin (approved):
Code looks fine not tested though. Will need network bump. Also might be worth a changelog entry

PR #13309, opened 2020-10-27
Fix #13306: Restore entrance/exit for prebuilt ride design ghosts

### LINKED ISSUE #13306: Prebuilt ride design ghost has no entrance or exit buildings
**OS:** Windows 10
**Version:** 0.3.1
**Commit/Build:** 03ef249

If you try to place down a prebuilt ride design anywhere where it can be built, its ride entrance and exit buildings will not show up. The station style it was saved with is still carried over though, and placing down the design is still possible without any major issue.

**Steps to reproduce:**
1. Select any prebuilt ride design for any ride type (e.g. Updraft for the Flying Roller Coaster).
2. Place it down anywhere where it can be built (assuming you have sufficient funds or if it was built with certain cheats, then either "Enable all drawable track pieces", "Allow chain lifts on all track pieces", "Disable clearance checks", or "Disable support limits").
3. You will now see that the entrance and exit buildings do not show up in the ride design's ghost, but they still show up correctly once the ride has actually been built.

**Screenshot:**
Termite Excavation 2020-10-26 18-08-49

manuelVo:
This fixes the regression introduced by c440dac where prebuilt design ghosts did not display their entrances. To fix this bug (while still getting rid of the desync bug) I've limited the changes I made in c440dac to only apply when the ride is in simulation mode. This fixes the desync - as it only occurs when simulating, while not affecting the prebuilt ghost (since you cannot simulate ride designs during placement).

### DIFF
--- a/src/openrct2/actions/TrackPlaceAction.hpp
+++ b/src/openrct2/actions/TrackPlaceAction.hpp
@@ -695,7 +695,10 @@ public:
                 }
             }
 
-            if (entranceDirections & TRACK_SEQUENCE_FLAG_ORIGIN && !(GetFlags() & GAME_COMMAND_FLAG_GHOST))
+            // If the placed tile is a station modify station properties.
+            // Don't do this if the ride is simulating and the tile is a ghost to prevent desyncs.
+            if (entranceDirections & TRACK_SEQUENCE_FLAG_ORIGIN
+                && !(ride->status == RIDE_STATUS_SIMULATING && GetFlags() & GAME_COMMAND_FLAG_GHOST))
             {
                 if (trackBlock->index == 0)
                 {
--- a/src/openrct2/actions/TrackRemoveAction.hpp
+++ b/src/openrct2/actions/TrackRemoveAction.hpp
@@ -429,7 +429,10 @@ public:
 
             cost += (_support_height / 2) * RideTypeDescriptors[ride->type].BuildCosts.SupportPrice;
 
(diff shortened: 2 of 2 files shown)


duncanspumpkin:
As this is getting a bit complex as to why it's like this we should at least have a comment next to this.

manuelVo:
Agreed. What's the exact meaning of the `entranceDirections & TRACK_SEQUENCE_FLAG_ORIGIN` and `tileElement->AsTrack()->GetSequenceIndex() == 0` checks? I haven't yet undestood those, but would like to add them to the comment properly.

duncanspumpkin:
GetSequenceIndex multi tile track pieces increment the sequence index for each one. This is saying the first track tile piece. TRACK_SEQUENCE_FLAG_ORIGIN not 100% sure on that one. Looks a bit odd that it's used with a direction but it's possible this makes sense.

manuelVo:
In this case I'll do some more testing to see if I can make any sense out of that so I can document it properly.

manuelVo:
Turns out that `entranceDirections & TRACK_SEQUENCE_FLAG_ORIGIN` in this context is a `isStation` check. And the way I see it `entranceDirections` would be better named `trackSequenceProperties`. However I still have no clue why the `tileElement->AsTrack()->GetSequenceIndex() == 0` is used in this context. So for now I've added a comment that ignored this line.

If anyone is able to put that into understandable terms I'd be happy to add that to the comment.

duncanspumpkin (approved):

PR #13310, opened 2020-10-27
Close #12450: Refactor SPRITE_IDENTIFIER to use strong enum

### LINKED ISSUE #12450: Refactor SPRITE_IDENTIFIER to use strong enum

ju-pinheiro:


ju-pinheiro:
Hi,

Some builds are failing now because several switch statements weren't handling all the SPRITE_IDENTIFIER enumerations.
For example: src/openrct2/world/Sprite.cpp:911:12: error: enumeration value ‘Litter’ not handled in switch [-Werror=switch].
Should I try to add some default cases and assertions? Then, I'd need help to make sure the behavior is still consistent :)

Thanks!

duncanspumpkin:
I'd rather you didn't add `default:` cases but add the specific missing cases (its usually just null that is missing which in almost all cases we want to nop). Do it as a separate commit to make it easy to review.

tupaschoal:
I agree with Duncan, I avoid `default` like the plague

duncanspumpkin (approved):

tupaschoal:
Note that it might take a week or so to merge, as we are about to release and will only merge bug fixed during this "code freeze",

PR #13311, opened 2020-10-27
Close #12395: Refactor PeepItem to use strong enum

### LINKED ISSUE #12395: Refactor PeepItem to use strong enum

Lukasz-Pe:
1st step is to combine SHOP_ITEM and PEEP_ITEM enums.
2nd step - Change SHOP_ITEM to use strong enum.


src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -2087,7 +2087,7 @@ int32_t guest_path_finding(Guest* peep)
     /* If there are still multiple directions to choose from,
      * peeps with maps will randomly read the map: probability of doing so
      * is much higher when heading for a ride or the park exit. */
-    if (peep->ItemStandardFlags & PEEP_ITEM_MAP)
+    if (peep->ItemStandardFlags & EnumToFlag(PeepItem::PEEP_ITEM_MAP))
```
ZehMatt, line 2090: They are already defined as flags, the purpose of EnumToFlag is that it does ```(1 << N)```

src/openrct2/peep/Peep.h
```diff
@@ -418,7 +418,7 @@ enum class PeepNauseaTolerance : uint8_t
     High
 };
 
-enum PeepItem
+enum class PeepItem : uint32_t
 {
     // ItemStandardFlags
     PEEP_ITEM_BALLOON = (1 << 0),
```
ZehMatt, line 424: The naming should be changed as well to adhere the other enums, also the elements should be not defined as flags which also means all code needs to be updated using EnumToFlags accordingly.

Lukasz-Pe: Renaming is done as well as flags are removed from this enum. It's now only uint32_t. And here is my doubt -  is that correct type for this enum? Maybe uint8_t would be better?

duncanspumpkin:
This isn't going to work nicely like this we need to make it so that standard and extra are combined into one nice 64bit variable.

Lukasz-Pe:
You mean uint64_t and continuous numbering?

duncanspumpkin:
Sorry i meant the enum stays as uint8 but the standard flags and extra flags need to be combined. It's quite a complex operation to combine them. This enum is going to be identical to the SHOP_ITEM_xxx enum. I'd like those two enums to end up as one. If you feel this is too complex we can split it up into multiple PRs.

Lukasz-Pe:
I'll give it a try :)
If I understand you correctly, PeepItem enum and SHOP_ITEM_ enum shold be combined as one of type uint8t?

duncanspumpkin:
It will need some careful thinking about the import and export code.

Lukasz-Pe:
I'll start refactoring from scratch and make a new PR for both PeepItem and ShopItem.

tupaschoal:
You can reset the commits but still continue on this pr, creating a new one makes us lose the comments and context, which is not ideal

Lukasz-Pe:
I've renamed all ShopItem enums. The code compiles on my machine.
Please review and comment. I would like to be sure, that `SET_FIELD` macro can be used as it is now.
There is a lot of `static_casts`, especially for iterators and I am not sure if this could be left that way.

tupaschoal (commented):
The CI is showing an out of bounds access (I haven't checked if it was introduced by you), but probably was (?)

Lukasz-Pe:
I haven't touched those loops, and thus I am very surprised with that behaviour. I was experimenting with calculation of max loop range, but those arrays, over which code loops, are initialised like this:
```
/** rct2: 0x0097EFCC */
static constexpr const uint8_t item_standard_litter[32] = {
    LITTER_TYPE_RUBBISH,          // ShopItem::Balloon
    LITTER_TYPE_RUBBISH,          // ShopItem::Toy
    LITTER_TYPE_RUBBISH,          // ShopItem::Map
    LITTER_TYPE_RUBBISH,          // ShopItem::Photo
    LITTER_TYPE_RUBBISH,          // ShopItem::Umbrella
    LITTER_TYPE_RUBBISH,          // ShopItem::Drink
    LITTER_TYPE_RUBBISH,          // ShopItem::Burger
    LITTER_TYPE_RUBBISH,          // ShopItem::Chips
    LITTER_TYPE_RUBBISH,          // ShopItem::IceCream
    LITTER_TYPE_RUBBISH,          // ShopItem::Candyfloss
    LITTER_TYPE_EMPTY_CAN,        // ShopItem::EmptyCan
    LITTER_TYPE_RUBBISH,          // ShopItem::Rubbish
    LITTER_TYPE_EMPTY_BURGER_BOX, // ShopItem::EmptyBurgerBox
    LITTER_TYPE_RUBBISH,          // ShopItem::Pizza
    LITTER_TYPE_RUBBISH,          // ShopItem::Voucher
    LITTER_TYPE_RUBBISH,          // ShopItem::Popcorn
    LITTER_TYPE_RUBBISH,          // ShopItem::HotDog
    LITTER_TYPE_RUBBISH,          // ShopItem::Tentacle
    LITTER_TYPE_RUBBISH,          // ShopItem::Hat
    LITTER_TYPE_RUBBISH,          // ShopItem::ToffeeApple
    LITTER_TYPE_RUBBISH,          // ShopItem::TShirt
    LITTER_TYPE_RUBBISH,          // ShopItem::Doughnut
    LITTER_TYPE_RUBBISH,          // ShopItem::Coffee
    LITTER_TYPE_EMPTY_CUP,        // ShopItem::EmptyCup
    LITTER_TYPE_EMPTY_BOX,        // ShopItem::Chicken
    LITTER_TYPE_EMPTY_BOTTLE,     // ShopItem::Lemonade
    LITTER_TYPE_EMPTY_BOX,        // ShopItem::EmptyBox
    LITTER_TYPE_EMPTY_BOTTLE,     // ShopItem::EmptyBottle
};

/** rct2: 0x0097EFE8 */
static constexpr const uint8_t item_extra_litter[32] = {
    LITTER_TYPE_RUBBISH,            // ShopItem::Photo2
    LITTER_TYPE_RUBBISH,            // ShopItem::Photo3
    LITTER_TYPE_RUBBISH,            // ShopItem::Photo4
    LITTER_TYPE_RUBBISH,            // ShopItem::Pretzel
    LITTER_TYPE_RUBBISH,            // ShopItem::Chocolate
    LITTER_TYPE_RUBBISH,            // ShopItem::IcedTea
    LITTER_TYPE_RUBBISH,            // ShopItem::FunnelCake
    LITTER_TYPE_RUBBISH,            // ShopItem::Sunglasses
    LITTER_TYPE_RUBBISH,            // ShopItem::BeefNoodles
    LITTER_TYPE_RUBBISH,            // ShopItem::FriedRiceNoodles
    LITTER_TYPE_RUBBISH,            // ShopItem::WontonSoup
    LITTER_TYPE_RUBBISH,            // ShopItem::MeatballSoup
    LITTER_TYPE_RUBBISH,            // ShopItem::FruitJuice
    LITTER_TYPE_RUBBISH,            // ShopItem::SoybeanMilk
    LITTER_TYPE_RUBBISH,            // ShopItem::SuJeongGwa
    LITTER_TYPE_RUBBISH,            // ShopItem::SubSandwich
    LITTER_TYPE_RUBBISH,            // ShopItem::Cookie
    LITTER_TYPE_EMPTY_BOWL_RED,     // ShopItem::EmptyBowlRed
    LITTER_TYPE_EMPTY_DRINK_CARTON, // ShopItem::EmptyDrinkCarton
    LITTER_TYPE_EMPTY_JUICE_CUP,    // ShopItem::EmptyJuiceCup
    LITTER_TYPE_RUBBISH,            // ShopItem::RoastSausage
    LITTER_TYPE_EMPTY_BOWL_BLUE,    // ShopItem::EmptyBowlBlue
};
```

and except comments, there was no change made.

BTW.
Why only windows MinGW CI fails, when the same code is compiled with other toolchains and there is no error?

Here is the original code I started with:
```
/** rct2: 0x0097EFCC */
static constexpr const uint8_t item_standard_litter[32] = {
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_BALLOON
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_TOY
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_MAP
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_PHOTO
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_UMBRELLA
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_DRINK
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_BURGER
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_CHIPS
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_ICE_CREAM
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_CANDYFLOSS
    LITTER_TYPE_EMPTY_CAN,        // PEEP_ITEM_EMPTY_CAN
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_RUBBISH
    LITTER_TYPE_EMPTY_BURGER_BOX, // PEEP_ITEM_EMPTY_BURGER_BOX
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_PIZZA
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_VOUCHER
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_POPCORN
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_HOT_DOG
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_TENTACLE
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_HAT
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_TOFFEE_APPLE
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_TSHIRT
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_DOUGHNUT
    LITTER_TYPE_RUBBISH,          // PEEP_ITEM_COFFEE
    LITTER_TYPE_EMPTY_CUP,        // PEEP_ITEM_EMPTY_CUP
    LITTER_TYPE_EMPTY_BOX,        // PEEP_ITEM_CHICKEN
    LITTER_TYPE_EMPTY_BOTTLE,     // PEEP_ITEM_LEMONADE
    LITTER_TYPE_EMPTY_BOX,        // PEEP_ITEM_EMPTY_BOX
    LITTER_TYPE_EMPTY_BOTTLE,     // PEEP_ITEM_EMPTY_BOTTLE
};

/** rct2: 0x0097EFE8 */
static constexpr const uint8_t item_extra_litter[32] = {
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_PHOTO2
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_PHOTO3
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_PHOTO4
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_PRETZEL
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_CHOCOLATE
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_ICED_TEA
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_FUNNEL_CAKE
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_SUNGLASSES
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_BEEF_NOODLES
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_FRIED_RICE_NOODLES
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_WONTON_SOUP
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_MEATBALL_SOUP
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_FRUIT_JUICE
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_SOYBEAN_MILK
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_SU_JONGKWA
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_SUB_SANDWICH
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_COOKIE
    LITTER_TYPE_EMPTY_BOWL_RED,     // PEEP_ITEM_EMPTY_BOWL_RED
    LITTER_TYPE_EMPTY_DRINK_CARTON, // PEEP_ITEM_EMPTY_DRINK_CARTON
    LITTER_TYPE_EMPTY_JUICE_CUP,    // PEEP_ITEM_EMPTY_JUICE_CUP
    LITTER_TYPE_RUBBISH,            // PEEP_ITEM_ROAST_SAUSAGE
    LITTER_TYPE_EMPTY_BOWL_BLUE,    // PEEP_ITEM_EMPTY_BOWL_BLUE
};
```

Only thing I can come up with now is the change of quantity of enums. There are 3 more positions in ShopItem enum in comparison with (now deleted) PeepItem enum. In ShopItem are additionally: `Admission = 31`, `Count = 56` and `None = 255`.

The loops code was not modified.

Lukasz-Pe:
I've taken the liberty of combining litter types into one variable. Please review.
A new variable is introduced to keep track of item_litter array size.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -1839,27 +1839,27 @@ void window_guest_inventory_update(rct_window* w)
...
+static std::pair<rct_string_id, Formatter> window_guest_inventory_format_item(Peep* peep, ShopItem item)
 {
     auto& park = OpenRCT2::GetContext()->GetGameState()->GetPark();
     auto parkName = park.Name.c_str();
 
     // Default arguments
     auto ft = Formatter();
-    ft.Add<uint32_t>(ShopItems[item].Image);
-    ft.Add<rct_string_id>(ShopItems[item].Naming.Display);
+    ft.Add<uint32_t>(ShopItems[static_cast<uint8_t>(item)].Image);
```
Gymnasiast, line 1849:
```suggestion
    ft.Add<uint32_t>(ShopItems[EnumValue(item)].Image);
```
Also applies to other occurrences.

Lukasz-Pe:
I want to repeair failing CI replay test. According to [wiki page](https://github.com/OpenRCT2/OpenRCT2/wiki/Replay-System) I have to rerecord the test. I have no experience with GitHub CI system and I would like to know what can I do to make the situation better.

tupaschoal:
First thing to consider is: did you introduce something that should have changed the behavior of the game? If not, then there might be a problem in the patch, rather than a need of re-recording the replays.

Sometimes they might fail because the underlying data changed, like in #13312

Lukasz-Pe:
Maybe this test fails due to merging of PeepItem and ShopItem. Except that, I do not find any code incursions that might have changed game behaviour.
I am bit confused by the fact, that only two of the tests are failing. I would expect all of them to fail in the same place.

Gymnasiast:
Perhaps we should just split up the PR. Have one PR that merges them, and then another which does the enum stuff. Would make for easier reviewing as well as a lower chance of bugs slipping in.

Lukasz-Pe:
I am resetting this PR.
Title will be changed to "Merge SHOP_ITEM and PEEP_ITEM to SHOP_ITEM". If everything will be ok, I'll make after merging next PR to change it to strong enum.

Lukasz-Pe:
@duncanspumpkin could you please look on this PR? An information about [mentioned](https://github.com/OpenRCT2/OpenRCT2/pull/13311#issuecomment-718472204) flags will be more than helpful.

duncanspumpkin:
The way i was imagining this working was a multi step process. From reviewing the pr you have sort of achieved most of this but i think accessor functions would simplify things still for the next step.

1. remove all direct reference to item standard flags by using accessor functions.

3. Replace any direct reference to the extra flags with accessor functions.

2. change all calls to the standard flags to use shop item enum instead of flags enum.

3. change all calls to the extra flags to use the shop item enum. Accessor functions will need to remember the 32 bit shift.

4. Delete flags enum as it shouldn't be in use now.

4. within the standard flags accessor function combine in the extra flags. I.e standardflags | (extra flags << 32).  Change name of functions to something more common.

5. Replace all calls to the extra flags with the calls to the common flags.

At this point you would pr the work and there would be no changes to the underlying data. Replays would work during all these steps.

After merging these changes there would be a follow up pr that combined the two fields on load into a single 64bit flags. This would require minimal changes to the accessor functions but will break replays. The replays would be rerecorded.

Lukasz-Pe:
@duncanspumpkin Thank you for the review & comment!
I'll make this PR as normal and integrate those two fields as you outlined.

Lukasz-Pe:
In meantime, what should be done with COMPARE_FIELD macro? Comment out nad make Item...Flags private? Or leave as is now?

Lukasz-Pe:
@duncanspumpkin Could you please look again on this PR?
I've managed to make those accessor functions, but if I use one 64bit variable there are CI errors (commit [039825c](https://github.com/OpenRCT2/OpenRCT2/pull/13311/commits/039825c130a2008f1eb88ba1a68ac823e530679d)).

src/openrct2/actions/ParkMarketingAction.hpp
```diff
@@ -77,7 +77,7 @@ DEFINE_GAME_ACTION(ParkMarketingAction, GAME_COMMAND_START_MARKETING_CAMPAIGN, G
         }
         else if (campaign.Type == ADVERTISING_CAMPAIGN_FOOD_OR_DRINK_FREE)
         {
-            campaign.ShopItemType = _item;
+            campaign.ShopItemType = static_cast<ShopItem>(_item);
```
duncanspumpkin, line 80: You can use the ShopItem constructor for this I think `ShopItem(_item)`

src/openrct2/actions/RideDemolishAction.hpp
```diff
@@ -174,41 +174,41 @@ DEFINE_GAME_ACTION(RideDemolishAction, GAME_COMMAND_DEMOLISH_RIDE, GameActions::
             }
 
             // remove any free voucher for this ride from peep
-            if (peep->ItemStandardFlags & PEEP_ITEM_VOUCHER)
+            if (peep->GetItemFlags() & EnumToFlag(ShopItem::Voucher))
             {
                 if (peep->VoucherType == VOUCHER_TYPE_RIDE_FREE && peep->VoucherRideId == _rideIndex)
                 {
-                    peep->ItemStandardFlags &= ~(PEEP_ITEM_VOUCHER);
+                    peep->SetItemFlags(peep->GetItemFlags() & ~EnumToFlag(ShopItem::Voucher), 0);
```
duncanspumpkin, line 181: Create a `ClearItemFlags` function then you don't need to do so much repetition.

src/openrct2/peep/Peep.cpp
```diff
@@ -3329,3 +3328,26 @@ void Peep::RemoveFromRide()
     }
     StateReset();
 }
+
+uint32_t Peep::GetItemFlags(bool ExtraItem) const
```
duncanspumpkin, line 3332: I would make this function like the following:

For cases where you are checking an individual flag:
i.e. to replace `peep->ItemStandardFlags & PEEP_ITEM_BALLOON`
```
bool Peep::HasItem(ShopItem item) const 
{
    return GetItemFlags() & EnumToFlag(item);
}
```

For cases where you need to check a few flags at the same time (unsure if there are any cases of this but need for the above either way):
i.e. `peep->ItemStandardFlags & (PEEP_ITEM_BALLOON | PEEP_ITEM_VOUCHER)`
```
uint64_t Peep::GetItemFlags() const
{
    return ItemStandardFlags | (ItemExtraFlags << 32);
}
```

For instances where you need to clear a flag:
i.e. `peep->ItemStandardFlags &= ~PEEP_ITEM_BALLOON;`
```
void Peep::RemoveItem(ShopItem item)
{
    // TODO: Join up standard and extra flags into a `uint64_t` to remove this if for the extra flags
    if (EnumValue(item) > 31)
    {
        ItemExtraFlags &= ~(1 << (EnumValue(item) - 31));
    } 
    else
    {
        ItemStandardFlags &= ~EnumToFlag(item);
    }
}
```

Then I'd also rename your `ResetItemFlags` to `RemoveAllItems`

src/openrct2/actions/RideSetPriceAction.hpp
```diff
@@ -181,7 +181,7 @@ DEFINE_GAME_ACTION(RideSetPriceAction, GAME_COMMAND_SET_RIDE_PRICE, GameActions:
             {
                 // If the shop item is the same or an on-ride photo
                 if (rideEntry->shop_item[1] == shopItem
-                    || (rideEntry->shop_item[1] == SHOP_ITEM_NONE && ShopItems[shopItem].IsPhoto()))
+                    || (rideEntry->shop_item[1] == ShopItem::None && ShopItems[EnumValue(shopItem)].IsPhoto()))
```
duncanspumpkin, line 184: `ShopItems[EnumValue(` will be used in a lot of locations. Make an accessor function.

```
GetShopItemDescriptor(ShopItem item)
{
    return ShopItems[EnumValue(item)];
}```

File: src/openrct2/management/Award.cpp [hunk 8]
```diff
@@ -305,9 +305,9 @@ static bool award_is_deserved_best_food(int32_t activeAwardTypes)
         auto rideEntry = get_ride_entry(ride.subtype);
         if (rideEntry != nullptr)
         {
-            if (!(shopTypes & (1ULL << rideEntry->shop_item[0])))
+            if (!(shopTypes & EnumToFlag(rideEntry->shop_item[0])))
```
duncanspumpkin, line 308: I'm surprised that ShopTypes is `uint32_t` it should be `uint64_t`. Investigate if this should be changed or leave a `TODO`

src/openrct2/peep/Guest.cpp
```diff
@@ -6803,37 +6816,37 @@ struct item_pref_t
 
 // clang-format off
 static item_pref_t item_order_preference[] = {
```
duncanspumpkin, line 6818: `item_pref_t` should be modified to use an `uint64_t` for item and remove of the `-ShopItem::Photo2` for the extra flags. You'll need to do a bit of work to make this work.

src/openrct2/peep/Peep.h
```diff
@@ -783,6 +726,7 @@ struct Peep : SpriteBase
     ride_id_t FavouriteRide;
     uint8_t FavouriteRideRating;
     uint32_t ItemStandardFlags;
+    uint64_t ItemFlags;
```
duncanspumpkin, line 729: remove for now. leave for a different PR.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1173,7 +1173,7 @@ void S6Exporter::ExportSpritePeep(RCT2SpritePeep* dst, const Peep* src)
     {
         dst->ride_types_been_on[i] = src->RideTypesBeenOn[i];
     }
-    dst->item_extra_flags = src->ItemExtraFlags;
+    dst->item_extra_flags = src->GetItemFlags(true);
```
duncanspumpkin, line 1176: will need reworking when you change the functions as suggested.

src/openrct2/ride/ShopItem.cpp
```diff
@@ -13,70 +13,85 @@
 #include "../localisation/StringIds.h"
 #include "../sprites.h"
 
+ShopItem& operator++(ShopItem& d, int)
```
duncanspumpkin, line 16: These shouldn't be required. Try to avoid requiring them.

duncanspumpkin (changes requested):
This is really good work and I appreciate what you have done so far. There is still a bit of work to do though to make it cleaner. Always think about how you want the code to eventually look. We don't want to have to expose the fact there is a difference between standard and extra flags and this PR should be hiding that fact to a few small functions. I've made some suggestions as to how I would structure the interface so that the peeps have an easier to understand interface of HasItem, RemoveItem, GiveItem.

src/openrct2/ride/ShopItem.cpp
Lukasz-Pe: I'll check tomorrow where I used it and delete if possible.

Lukasz-Pe:
Thank you for your review. I am still working on it and I'll try to make this interface more straightforward.

src/openrct2/rct2/S6Exporter.cpp
Lukasz-Pe: now this function has two parameters, which make it to work like directly accessed flags. On the next PR we could without much work make the function to return uint64_t with one simple change to funtion signature.

src/openrct2/ride/ShopItem.h
```diff
@@ -108,8 +115,10 @@ enum
...
-extern const ShopItemDescriptor ShopItems[SHOP_ITEM_COUNT];
+extern const ShopItemDescriptor ShopItems[EnumValue(ShopItem::Count)];
 extern uint64_t gSamePriceThroughoutPark;
 
-money32 shop_item_get_common_price(Ride* forRide, const int32_t shopItem);
-bool shop_item_has_common_price(const int32_t shopItem);
+money32 shop_item_get_common_price(Ride* forRide, const ShopItem shopItem);
+bool shop_item_has_common_price(const ShopItem shopItem);
+
+ShopItemDescriptor GetShopItemDescriptor(ShopItem item);
```
Gymnasiast, line 124: Always end your files with a newline.

src/openrct2/object/RideObject.cpp
```diff
@@ -83,8 +83,8 @@ void RideObject::ReadLegacy(IReadObjectContext* context, IStream* stream)
     _legacyType.max_height = stream->ReadValue<uint8_t>();
     // Skipping a uint64_t for the enabled track pieces and two uint8_ts for the categories.
     stream->Seek(10, STREAM_SEEK_CURRENT);
-    _legacyType.shop_item[0] = stream->ReadValue<uint8_t>();
-    _legacyType.shop_item[1] = stream->ReadValue<uint8_t>();
+    _legacyType.shop_item[0] = stream->ReadValue<ShopItem>();
```
Gymnasiast, line 86: Change this back to `uint8_t`. We might change the shop item to use uint16_t at some point, but in .DAT objects it will always be a `uint8_t`.

Lukasz-Pe: Do you mean `_legacyType.shop_item` to be `uint8_t` or only `ReadValue<>`?
I would like to do following: `_legacyType.shop_item[0] = static_cast<ShopItem>(stream->ReadValue<uint8_t>());`

Gymnasiast: Only `ReadValue`.

The code you suggested look good to me.

src/openrct2/peep/Guest.cpp
```diff
@@ -782,14 +782,11 @@ void Guest::loc_68FA89()
                 chosen_food = bitscanforward(HasFoodExtraFlag());
                 if (chosen_food != -1)
                 {
-                    ItemExtraFlags &= ~(1 << chosen_food);
+                    RemoveItem(static_cast<ShopItem>(chosen_food));
```
duncanspumpkin, line 785: this will be removing the wrong food item. You need to add 32 to the value. In a later refactor this will not be required

src/openrct2/peep/Guest.cpp
```diff
@@ -1591,9 +1569,9 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
                 if (itemValue > (static_cast<money16>(scenario_rand() & 0x07)))
                 {
                     // "I'm not paying that much for x"
-                    PeepThoughtType thought_type = static_cast<PeepThoughtType>(
-                        (shopItem >= 32 ? (PEEP_THOUGHT_TYPE_PHOTO2_MUCH + (shopItem - 32))
-                                        : (PEEP_THOUGHT_TYPE_BALLOON_MUCH + shopItem)));
+                    PeepThoughtType thought_type = static_cast<PeepThoughtType>((
+                        shopItem >= ShopItem::Photo2 ? (ShopItem(PEEP_THOUGHT_TYPE_PHOTO2_MUCH) + (shopItem - ShopItem::Photo2))
```
duncanspumpkin, line 1573: This is wrong. It should be:
```
PeepThoughtType thought_type = 
                        shopItem >= ShopItem::Photo2 ? PEEP_THOUGHT_TYPE_PHOTO2_MUCH + static_cast<uint8_t>((shopItem - ShopItem::Photo2))) 
                                                     : PEEP_THOUGHT_TYPE_BALLOON_MUCH + static_cast<uint8_t>(shopItem);
```

src/openrct2/peep/Guest.cpp
```diff
@@ -1610,8 +1588,8 @@ bool Guest::DecideAndBuyItem(Ride* ride, int32_t shopItem, money32 price)
                 {
                     // "This x is a really good value"
                     PeepThoughtType thought_item = static_cast<PeepThoughtType>(
-                        (shopItem >= 32 ? (PEEP_THOUGHT_TYPE_PHOTO2 + (shopItem - 32))
-                                        : (PEEP_THOUGHT_TYPE_BALLOON + shopItem)));
+                        (shopItem >= ShopItem::Photo2 ? (ShopItem(PEEP_THOUGHT_TYPE_PHOTO2) + (shopItem - ShopItem::Photo2))
```
duncanspumpkin, line 1591: as above

src/openrct2/peep/Guest.cpp
```diff
@@ -5366,7 +5339,7 @@ void Guest::UpdateWalking()
                 for (int32_t container = HasEmptyContainerExtraFlag(); pos_extr < 32; pos_extr++)
                     if (container & (1u << pos_extr))
                         break;
-                ItemExtraFlags &= ~(1u << pos_extr);
+                RemoveItem(static_cast<ShopItem>(pos_extr));
```
duncanspumpkin, line 5342: This is incorrect need to add 32 to the value

src/openrct2/peep/Guest.cpp
```diff
@@ -5963,7 +5936,7 @@ void Guest::UpdateUsingBin()
                     // switched to scenario_rand as it is more reliable
                     if ((scenario_rand() & 7) == 0)
                         space_left_in_bin--;
-                    ItemExtraFlags &= ~(1 << cur_container);
+                    RemoveItem(static_cast<ShopItem>(cur_container));
```
duncanspumpkin, line 5939: Need to add 32 to the value

src/openrct2/peep/Guest.cpp
```diff
@@ -5975,7 +5948,7 @@ void Guest::UpdateUsingBin()
                 int32_t litterY = y + (scenario_rand() & 7) - 3;
 
                 litter_create({ litterX, litterY, z, static_cast<Direction>(scenario_rand() & 3) }, litterType);
-                ItemExtraFlags &= ~(1 << cur_container);
+                RemoveItem(static_cast<ShopItem>(cur_container));
```
duncanspumpkin, line 5951: need to add 32 to value

duncanspumpkin (changes requested):
Almost there few little issues left.

src/openrct2/peep/Guest.cpp
Lukasz-Pe: I have corrected it, but `static_cast<PeepThoughtType>` had to be left to avoid compilation errors.

duncanspumpkin (approved):
There are a few things we can change in the future but this is good to go as is for now.

PR #13312, opened 2020-10-27
Use ride_id_t in more places

Gymnasiast:


duncanspumpkin:
Will need Replay rerecord and network bump

src/openrct2/rct12/RCT12.h
```diff
@@ -14,6 +14,7 @@
 #include "../common.h"
 #include "../object/Object.h"
 
+#include <src/openrct2/ride/RideTypes.h>
```
duncanspumpkin, line 17: `../ride/RideTypes.h`

Gymnasiast: Argh, my IDE keeps auto-adding that :/

duncanspumpkin: I'll fix it never mind.

Gymnasiast: Sorry, read that too late - I already pushed a fix for that and for formatting.

duncanspumpkin: never mind you got there first.

duncanspumpkin (approved):

Issue #13292, opened 2020-10-24
Impossible excitement rating requirements with finish building 5 coasters goal


Fred-1044276:
**OS:** Windows 10
**Version:** 0.3.1
**Commit/Build:** 4c39ee9

If you create a scenario that starts as a "Pay to enter park / Free rides" scenario and then attempt to turn it into a "Finish building 5 roller coasters scenario" after that, you will get extremely high excitement rating requirements like 19,666.60 even though you should be getting between 4.00 and 9.90 instead like if you did it with any of the other park types ("Pay to enter park / Pay per ride" or "Free park entry / Pay per ride"). The issue still persists even if the save was modified to change the park from "Pay to enter park / Free rides" to either "Pay to enter park / Pay per ride" or "Free park entry / Pay per ride".

I have also provided the problematic scenario and save file below. I eventually scrapped Vulcanodon Caldera's original "Finish building 5 roller coasters" goal and turned it into a "Number of guests in park" goal scenario instead.

**Steps to reproduce:**
There might be more ways to reproduce this bug, but these steps are how I got the bug to appear (minus the park value goal).
1. Create a new scenario and make it "Pay to enter park / Free rides". Also make it have a park value goal.
2. Save a game in that scenario and then convert the save file into a scenario.
3. Select the "Finish building 5 roller coasters" objective.
4. Now you will see values like 19,666.60 for the excitement.

**Screenshots / Video:**
![Vulcanodon Caldera 2020-10-24 15-21-39](https://user-images.githubusercontent.com/61885939/97094759-a9419c00-160c-11eb-9b6d-7b796f9e0556.png)

**Save game / Scenario:**
[Vulcanodon Caldera test.zip](https://github.com/OpenRCT2/OpenRCT2/files/5433865/Vulcanodon.Caldera.test.zip)

PR #13316, opened 2020-10-28
Fix: #13292: Impossible excitement rating requirements...

evilclownattack:
Incorrect argument was used when drawing the minimum excitement rating in the Objective Selection window. Appears to have been a typo.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -161,6 +161,7 @@ The following people are not part of the development team, but have been contrib
 * Stephan Spengler (Sadret)
 * Roger Seekell (rpstester)
 * Ben Johnston (gsckoco)
+* (evilclownattack)
 
 ## Toolchain
 * (Balletie) - macOS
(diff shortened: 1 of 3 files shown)


tupaschoal (approved):
This looks ok to me, though I think using the default instead of naming the type is something we should tend to later on.

tupaschoal:
Please add a changelog entry.

@Gymnasiast , can you please review it? Seems to have been caused by https://github.com/OpenRCT2/OpenRCT2/commit/6e22eade675117e19f4c905af9800568d460673b

evilclownattack:
Should I add any fixes to the changelog when first submitting a pull request in the future?

tupaschoal:
It depends, we only add stuff to changelog if it changes the behavior from what was seen in the last release (fix a bug, enhancement, feature, etc...)

Gymnasiast (changes requested):
Could you create a new case label for the "finish 5 coasters" objective and revert the default label to what it was?

Gymnasiast (approved):

PR #13329, opened 2020-10-31
Fix #12981: Heartline Twister Coaster (reverse car) does not appear

### LINKED ISSUE #12981: Heartline Twister Coaster (reverse car) does not appears during game
**OS:** Windows 10
**Version:** 0.3.0
**Commit/Build:** 4030d90

**Steps to reproduce:**
1. Make sure the Heartline Twister coaster is invented and there is reverse car in invention list
2. Research the reverse car
3. It shows that reverse car is invented but I cannot select it.
4. It is selectable when I save the map and load again.

I've tested in this map: https://rctsc.telk.kr/download?no=901 so I have no idea whether it is occured in certain map or it is a general issue.

**Screenshots / Video:**
image

evilclownattack:
Newly researched vehicle types were not appearing in the dropdown menu. Fixed by checking if research list has changed, or if "ignore research status" cheat was activated.

Marking this as a draft because I'm unsure about the implementation, admittedly this fix is less than elegant. An alternative solution would be to skip checking if the list has changed altogether, and just repopulate it every time.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -24,6 +24,7 @@
 - Fix: [#10186] Placing multiple saved rides ignores design name (original bug).
 - Fix: [#12368] Desync due to ghost station pieces affecting changing ride settings.
 - Fix: [#12940] Windows cause issues with snow drawing.
+- Fix: [#12981] Heartline Twister Coaster (reverse car) does not appears during game.
 - Fix: [#13019] Simulated trains sometimes open construction window when they crash.
 - Fix: [#13021] Mowed grass and weeds don't show up in extra zoom levels.
 - Fix: [#13024] Console cursor does not correctly render at current cursor position.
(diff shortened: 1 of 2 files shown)


tupaschoal (commented):
I don't think this is the best way to fix this. When the research finishes, we should instead broadcast an intent and make the Ride window react to that intent. That way, we only refresh it upon the research emitting the intent.

evilclownattack:
Would we also need to broadcast anytime an alteration is made in the Invention List/Object Selection? Initially I was just checking if `gResearchLastItem` had changed but since the Invention List resets it, there was the potential for any changes made via the cheat menu during a game to not register in the vehicle dropdown.

tupaschoal:
I would say so, though one has to investigate first whether this isn't already being done.

distribution/changelog.txt
```diff
@@ -24,6 +24,7 @@
 - Fix: [#10186] Placing multiple saved rides ignores design name (original bug).
 - Fix: [#12368] Desync due to ghost station pieces affecting changing ride settings.
 - Fix: [#12940] Windows cause issues with snow drawing.
+- Fix: [#12981] Heartline Twister Coaster (reverse car) does not appears during game.
```
thyeggman, line 27:
```suggestion
- Fix: [#12981] Heartline Twister Coaster (reverse car) does not appear during game.
```

tupaschoal:
The PR went inactive for too long, and also the issue this refers to has already been solved, so closing this PR.

PR #13332, opened 2020-10-31
Split declarations and definitions to improve compile times when editing

Broxzier:
When making changes to one of the .hpp files, all source files that include them (or that include headers that include them), you have to wait a few minutes before you can even test them. By moving the definitions to the source files, the compile-time is reduced significantly.

Abstract classes, templated classes, subclasses thereof, extremely small classes like `IOException` and `JsonException`, and plug-in .hpp files are skipped; the latter because I assume there's some reason for having them like this. Perhaps because those files are only included in a few source files.


src/openrct2/core/FileStream.h
```diff
@@ -53,87 +53,9 @@ namespace OpenRCT2
...
-                {
-                    fclose(_file);
-                }
-            }
-        }
+        FileStream(const std::string& path, int32_t fileMode);
+        FileStream(const utf8* path, int32_t fileMode);
+        ~FileStream() override;
 
         bool CanRead() const override
```
duncanspumpkin, line 60: why not move more of the bellow over so we can move the #ifdef macro stuff into the cpp file.

src/openrct2/core/JobPool.h
```diff
@@ -0,0 +1,53 @@
...
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include <atomic>
+#include <cassert>
```
duncanspumpkin, line 13: dont think cassert is needed here anymore.

src/openrct2/core/StringBuilder.h
```diff
@@ -24,92 +24,54 @@ class StringBuilder final
 {
```
duncanspumpkin, line 24: Look into the headers that are included pretty sure iterator and algorithm can be moved to the cpp file. Not sure if they are even used any more.

src/openrct2/core/FileStream.h
Broxzier: Not sure why I skipped these, maybe because they were small. I've moved them to the .cpp files now, as well as the includes and macros at the top.

duncanspumpkin (approved):

tupaschoal:
There are some weird errors in some of the CIs 🤔

duncanspumpkin:
The mac one is due to xcode needing project update.

Broxzier:
I don't get the errors in `LanguagePack.cpp`. The headers `StringBuilder.h` and `FileStream.h` are included, yet there are undefined references to their content.

Gymnasiast:
@Broxzier The errors occurred when building the tests. The necessary .cpp files needed to be added to the CMake file for the language test, as well as some others.

Broxzier:
Ah, that explains it. Thanks for fixing it @Gymnasiast!

PR #13333, opened 2020-11-01
Refactor add 3d light follow on from #13281

duncanspumpkin:
Close #13281 

Refactored the add 3d light functions to not ask for as many params as well.


tupaschoal (commented):
You'll either need to move the `place` definitions back to the beginning of the fn or add curly braces on lots of case statements.

duncanspumpkin:
Oh yeah i forgot i needed to do that.

IntelOrca (approved):

Issue #13334, opened 2020-11-01
uninitialised variables in CustomTabDesc


Sadret:
If the image is provided as a number instead of an ImageAnimation, the variables `imageFrameCount` and `imageFrameDuration` are not initialised. So, they will contain whatever is in the memory at that time. This leads to the tab image cycling through the available sprite images over time.

PR #13335, opened 2020-11-01
Fix: [#13334] Uninitialised variables in CustomTabDesc

Sadret:


distribution/changelog.txt
```diff
@@ -43,7 +43,7 @@
 - Fix: [#13266] Plugin API: Deleting key of sharedStorage not working.
 - Fix: [#13278] Desync caused by ghost tiles changing the ride mode.
 - Fix: [#13289] Litter and vomit sometimes not loading with RCT1 saved game or scenario
-- Fix: [#13292] Impossible excitement rating requirements with finish building 5 coasters goal
+- Fix: [#13334] Uninitialised variables in CustomTabDesc
```
Gymnasiast, line 46: Looks like you made a mistake here. But a release is imminent, so it's better to wait with fixing and rebasing.

Sadret:
Why can't it auto merge the changelog?

Gymnasiast:
Because it needs a few lines on either side that are the same. You don't want it to automerge where it shouldn't.

Sadret:
Ok, good to know.

PR #13350, opened 2020-11-02
Close #13346: Change FootpathScenery to FootpathAddition

### LINKED ISSUE #13346: footpath addition vs. footpath scenery
Benches, litter bins etc. are called "footpath additions" almost everywhere in the source code.
On the other hand, the corresponding game actions are called `FootpathScenery[Place|Remove]Action`.

I propose to change this, so that it is called "addition" everywhere. This would include changing the plugin API, which also uses e.g. `FootpathElement->addition` and `ObjectType "footpath_addition"` alongside `ActionType "footpathsceneryplace"` and `ActionType "footpathsceneryremove"`.

Sadret:
Changes:
- The filenames and classnames of `FootpathScenery[Place|Remove]Action` were changed to `FootpathAddition[Place|Remove]Action`.
- Instance variables of these classes were adjusted accordingly.
- The plugin API action names were adjusted: `footpathscenery[place|remove]` changed to `footpathaddition[place|remove]`. The API version was increased from 7 to 8.

Because of the last point, this update is not strictly backwards compatible for plugins.
In my opinion, this is not very important because of two aspects:
- First, I do not think that right now there are that many plugins that rely on this part of the API.
- Second, if an existing plugin uses this feature, the update to the new API does not involve a lot of effort.

Last remarks:
- The game and the plugin API were tested and yielded no problem.
- A typo in the .d.ts file was fixed: `foothpathremove` -> `footpathremove`.

Edit:
- Restored alphabetical order of includes in `GameActionRegistration`.


duncanspumpkin (approved):

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -963,8 +963,8 @@ const static std::unordered_map<std::string, uint32_t> ActionNameToType = {
     { "footpathplace", GAME_COMMAND_PLACE_PATH },
     { "footpathplacefromtrack", GAME_COMMAND_PLACE_PATH_FROM_TRACK },
     { "footpathremove", GAME_COMMAND_REMOVE_PATH },
-    { "footpathsceneryplace", GAME_COMMAND_PLACE_FOOTPATH_SCENERY },
-    { "footpathsceneryremove", GAME_COMMAND_REMOVE_FOOTPATH_SCENERY },
+    { "footpathadditionplace", GAME_COMMAND_PLACE_FOOTPATH_SCENERY },
```
Gymnasiast, line 966: Could you also rename the game command enum members accordingly?

Sadret: Sure, done.

Gymnasiast:
Should not have gone in the changelog, as it is something that only matters to the developers and is not something that will affect the users.

duncanspumpkin:
It would be in the changelog as its a change to the plugin api. So would want to draw attention to it just for that.

PR #13354, opened 2020-11-03
Refactor UI/Interface misc to TitleCase

pizza2004:
I included a few in this one again because they're all a bit small.

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -40,19 +40,19 @@
 #include <openrct2/world/Surface.h>
 #include <openrct2/world/Wall.h>
 
-static void viewport_interaction_remove_scenery(TileElement* tileElement, const CoordsXY& mapCoords);
-static void viewport_interaction_remove_footpath(TileElement* tileElement, const CoordsXY& mapCoords);
-static void viewport_interaction_remove_footpath_item(TileElement* tileElement, const CoordsXY& mapCoords);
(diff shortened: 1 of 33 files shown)


duncanspumpkin (approved):

duncanspumpkin:
Will need a rebase due to the other PRs

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -721,7 +721,7 @@ static Peep* viewport_interaction_get_closest_peep(ScreenCoordsXY screenCoords,
  *
  *  rct2: 0x0068A15E
  */
-CoordsXY sub_68A15E(const ScreenCoordsXY& screenCoords)
+CoordsXY ViewportInteractionGetTileStartAtCursor(const ScreenCoordsXY& screenCoords)
```
pizza2004, line 724: While I was trying to figure out what this function did in order to rename it, I found that removing lines 744 to 763 and simply returning the "initialPos" variable did not seem to change the functionality at all in game.  I tested just about every use case except placing an entrance, but I'm not sure what those lines are doing so I'm not sure how good my tests were.  Might be a place to remove unnecessary code, but I figured it wasn't the scope of this PR.

duncanspumpkin: Yes please do not do any code changes during these refactors no matter how tempting they might be. Make a follow on pr or issue so we can track it.

duncanspumpkin: That bit of code i think is hit when there is columns of land. I think it is required

PR #13357, opened 2020-11-03
Rename sub_6B5952 and remove one call of the function that is unreachable

manuelVo:
One of the calls to `sub_6B5952` is being done when Ride::Test is called with open state `RIDE_STATUS_OPEN`, which will will never happen, so I removed it.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -3893,14 +3893,14 @@ static int32_t ride_check_for_entrance_exit(ride_id_t rideIndex)
 }
 
 /**
- *
+ * Calls footpath_chain_ride_queue for all entrances of the ride
  *  rct2: 0x006B5952
  */
-static void sub_6B5952(Ride* ride)
+void Ride::ChainQueues() const
 {
     for (int32_t i = 0; i < MAX_STATIONS; i++)
     {
-        auto location = ride_get_entrance_location(ride, i);
+        auto location = ride_get_entrance_location(this, i);
         if (location.isNull())
             continue;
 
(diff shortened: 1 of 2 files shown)


tupaschoal (commented):
Can you elaborate on why this is being done?

> One of the calls to sub_6B5952 is being done when Ride::Test is called with open state RIDE_STATUS_OPEN, which will will never happen, so I removed it.

manuelVo:
I can explain this in two ways. One is a more high level, game logic approach (1), while the other is a low level approach that argues purely on the source code given (2).

1. A ride that is being tested is closed by definition. As such the action of setting a ride into testing mode while simultaneously opening it does not exist, which means that when changing to testing mode a condition that checks whether the ride should be opened must always evaluate to `false`.

2. There are two closely related functions that share quite a lot of similar code: `Ride::Open` and `Ride::Test`. As the names suggest they are being called to test or respectively open the rides. The only calls to those functions are made by the `RideSetStatus` game action. This game action decides to call Test or Open based on the new status (which in case of the `Test` function is then passed as parameter). If - and only if - the new status is either `RIDE_STATUS_TESTING` or `RIDE_STATUS_SIMULATING` the game action calls `Test`. This means that `Test` is never being called with the status `RIDE_STATUS_OPEN`, making the code I removed dead code. Since that code never gets executed it only adds complexity without providing any benefits so I decided to remove it.

I didn't track down where this dead code came from, but since the functions `Ride::Open` and `Ride::Test` are so closely related I suspect at some point code got copied from one function to the other, bringing over this snipped of code, which is only relevant for opening the ride but not for testing it.

tupaschoal (approved):
I tracked it down and your analysis make sense, thanks for the explanation!

PR #13359, opened 2020-11-03
Close #13057: Make GameAction flags accessible by plugins.

### LINKED ISSUE #13057: Make GameAction flags accessible by plugins
For e.g. ghost placement, it is neccessary to be able to set certain GameAction flags.
This should be available for plugins.

Sadret:
Following the recommendations of Duncanspumpkin.


duncanspumpkin (approved):
Won't you need another plugin version increment

Sadret:
Yes, I added that to the commit.

oli414:
This change caused the "flag" field to be required instead of optional, breaking any plugins that use "executeAction".

PR #13360, opened 2020-11-03
Rename ride_set_block_points to ride_open_block_brakes

manuelVo:


duncanspumpkin:
Could you title case the function whilst you are at it.

manuelVo:
The function isn't part of a struct. Should the name still be title-cased?

duncanspumpkin:
Yes, all functions should be TitleCase we used snake case when we first implemented everything in C before we decided on a coding style.

manuelVo:
Done

duncanspumpkin (approved):

PR #13363, opened 2020-11-03
Rename VEHICLE_UPDATE_FLAG_1

manuelVo:


src/openrct2/ride/Ride.cpp
```diff
@@ -4635,7 +4635,7 @@ static void vehicle_unset_update_flag_b1(Vehicle* head)
 {
```
duncanspumpkin, line 4635: Rename this function whilst you are doing this.

duncanspumpkin (approved):
May as well correct the function name whilst we are at it

manuelVo:
Done

tupaschoal (approved):

PR #13364, opened 2020-11-03
fix address comments in TileElement.h

Sadret:
Some addresses were off by one.
Also, `BannerElement->AllowedEdges` was listed as unused.


duncanspumpkin:
I mean we dont want people accessing the data like this. This was discussed in a previous PR. We should make accessors for the data rather than this as users shouldn't be dependant on the memory layout.

Sadret:
Yes, but right now this is still a mistake in the comments.

Sadret:
I didn't do this change with plugin developers in mind.
It would just bug me if I know there is a mistake and I don't fix it.

IntelOrca (approved):

PR #13368, opened 2020-11-04
Remove unnecessary parameter from Ride::Open

manuelVo:
Similarly to my explanation in https://github.com/OpenRCT2/OpenRCT2/pull/13357#issuecomment-721413793 the parameter `goingToBeOpen` of `Ride::Open` will always be set to `true`, so it's redudndant and can be removed.

I've tracked down where this redundant parameter comes from: In d0a50c43f4b6c7d0fa24e5c7c6232807a3f54d05, opening and testing a ride were done by the same function (back then named `ride_is_valid_for_open`) and the parameter `goingToBeOpen` indicated whether the ride should be opened (`true`) or tested (`false`). 37221f0aaeb90a37f135e0ae884ca5bc69651deb then split up testing and opening into two different functions, from which on `goingToBeOpen` lost it's meaning.

### DIFF
--- a/src/openrct2/actions/RideSetStatus.hpp
+++ b/src/openrct2/actions/RideSetStatus.hpp
@@ -110,7 +110,7 @@ public:
             }
             else if (_status == RIDE_STATUS_OPEN)
             {
-                if (!ride->Open(_status == RIDE_STATUS_OPEN, false))
+                if (!ride->Open(false))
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
Looks good. Pity that the true and false provide no useful guidance as to what they are for.

PR #13369, opened 2020-11-04
Refactor loc_6DDF9C

manuelVo:
This PR renames `loc_6DDF9C` and puts it into `Ride`. I also unified two occurances of `train->UpdateTrackMotion(nullptr);` that were called in the `if` and `else` case of an if into a single call before the if.

This PR also contains a commit that moves `ride_create_vehicles` into `Ride` to allow `loc_6DDF9C` to become a private function without causing visibility conflicts.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -138,7 +138,6 @@ static void ride_inspection_update(Ride* ride);
 static void ride_mechanic_status_update(Ride* ride, int32_t mechanicStatus);
 static void ride_music_update(Ride* ride);
 static void ride_shop_connected(Ride* ride);
-void loc_6DDF9C(Ride* ride, TileElement* tileElement);
 
 RideManager GetRideManager()
 {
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -4779,34 +4778,35 @@ static bool ride_create_vehicles(Ride* ride, const CoordsXYE& element, int32_t i
...
 
 /**
- *
+ * Move all the trains so each one will be placed at the block brake of a different block.
+ * The first vehicle will placed into the first block and all other vehicles in the blocks
+ * preceding that block.
  *  rct2: 0x006DDF9C
  */
-void loc_6DDF9C(Ride* ride, TileElement* tileElement)
+void Ride::MoveTrainsToBlockBrakes(const TileElement* firstBlock)
```
duncanspumpkin, line 4791: You can simplify the logic slightly by passing in a TrackElement reference instead of the generic TileElement ptr.

manuelVo:
Done

duncanspumpkin (approved):

PR #13371, opened 2020-11-04
Make TileElement conversion preserve the const qualifier

manuelVo:
Currently, `TileElement` (and `RCT12TileElement`) conversion functions silently drop the `const` qualifiers of tile elements it's called on. This commit adds more conversion functions so `const` qualifiers are kept as they are during conversion.

Additionally this commit fixes some constness issues revealed by this change.


Gymnasiast (approved):
Would like the approval of @duncanspumpkin as well.

duncanspumpkin (approved):

duncanspumpkin:
I don't think we really need the RCT12 ones but no harm adding them.

PR #13374, opened 2020-11-04
Create calls for min/max cars per train

Gymnasiast:
I also want to actually split the field later on. Not doing so now also means that this should pass the replay tests as-is.

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -2821,10 +2821,10 @@ static OpenRCT2String window_ride_vehicle_tooltip(
 
             auto ft = Formatter();
             ft.Increment(16);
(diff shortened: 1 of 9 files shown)


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2821,10 +2821,10 @@ static OpenRCT2String window_ride_vehicle_tooltip(
 
             auto ft = Formatter();
             ft.Increment(16);
-            ft.Add<uint16_t>(std::max(1, ride->min_max_cars_per_train & 0xF) - rideEntry->zero_cars);
+            ft.Add<uint16_t>(std::max(uint8_t(1), ride->GetMaxCarsPerTrain()) - rideEntry->zero_cars);
```
duncanspumpkin, line 2824: You can specify in the max function the type you want max<uint8_t>

duncanspumpkin (approved):

PR #13375, opened 2020-11-05
Convert a register to a variable

manuelVo:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -8322,10 +8322,10 @@ bool Vehicle::UpdateTrackMotionForwards(rct_ride_entry_vehicle* vehicleEntry, Ri
     }
     auto head = otherVeh->TrainHead();
 
-    regs.eax = abs(velocity - head->velocity);
+    auto velocityDifference = abs(velocity - head->velocity);
```
duncanspumpkin, line 8325: I always like the term delta for these.

manuelVo:
Yes, that's probably a bit more elegant.

duncanspumpkin (approved):

PR #13377, opened 2020-11-05
Close #13376: Open custom window at specified tab.

### LINKED ISSUE #13376: Plugin API feature request: open window at specified tab
It would be nice to be able to open a custom window at a specific tab.

Sadret:
Added WindowDesc->tabIndex to API.
Incremented API version.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,9 +1,10 @@
 0.3.2+ (in development)
 ------------------------------------------------------------------------
 - Feature: [#13057] Make GameAction flags accessible by plugins.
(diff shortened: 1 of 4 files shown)


duncanspumpkin:
I don't think this is very safe to do. Changing the tab on normal windows involves usually a good bit of logic happening. I know this is just for the custom window that works differently but if we ever move the normal windows to be like this we might have issues.

Sadret:
Why do you think it's unsafe? It's not even _changing_ the tab, it's setting the tab on window creation.

duncanspumpkin:
Yeah sorry for some reason i missed the "window creation" part of the pr. It is indeed fine to specify a tab on creation.

PR #13379, opened 2020-11-05
Allow object lookup using new JSON id

Gymnasiast:
This adds the JSON identifier to the object repository and adds some functions to load them.

This also changes the loading of the default surfaces, edges and stations to use the JSON ids.

It is still necessary to provide legacy entry in the JSON files, as much of the object loading still revolves around `rct_object_entry`. But this puts us a step closer towards making it optional.

### DIFF
--- a/src/openrct2/object/ObjectRepository.cpp
+++ b/src/openrct2/object/ObjectRepository.cpp
@@ -49,11 +49,10 @@ using namespace OpenRCT2;
 
 struct ObjectIdentifierHash
 {
-    size_t operator()(const std::string_view& identifier) const
+    size_t operator()(std::string_view identifier) const
     {
-        std::string copy = identifier.data();
         uint32_t hash = 5381;
-        for (auto i : copy)
+        for (auto i : identifier)
         {
             hash = ((hash << 5) + hash) + i;
         }
@@ -63,7 +62,7 @@ struct ObjectIdentifierHash
 
 struct ObjectIdentifierEqual
 {
-    bool operator()(const std::string_view& lhs, const std::string_view& rhs) const
+    bool operator()(std::string_view lhs, std::string_view rhs) const
     {
         return lhs == rhs;
     }
@@ -306,9 +305,9 @@ public:
         return nullptr;
     }
 
-    const ObjectRepositoryItem* FindObject(const std::string_view& identifier) const override final
+    const ObjectRepositoryItem* FindObject(std::string_view identifier) const override final
     {
-        auto kvp = _newItemMap.find(identifier.data());
+        auto kvp = _newItemMap.find(std::string(identifier));
         if (kvp != _newItemMap.end())
         {
             return &_items[kvp->second];
--- a/src/openrct2/object/ObjectRepository.cpp
+++ b/src/openrct2/object/ObjectRepository.cpp
@@ -47,27 +47,6 @@
 
 using namespace OpenRCT2;
 
-struct ObjectIdentifierHash
-{
-    size_t operator()(std::string_view identifier) const
-    {
-        uint32_t hash = 5381;
-        for (auto i : identifier)
-        {
-            hash = ((hash << 5) + hash) + i;
-        }
-        return hash;
-    }
-};
-
-struct ObjectIdentifierEqual
-{
-    bool operator()(std::string_view lhs, std::string_view rhs) const
-    {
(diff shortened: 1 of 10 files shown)


Gymnasiast:
Not sure why the win32 CI fails. Do you have any idea, @IntelOrca ?

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -47,6 +47,28 @@
 
 using namespace OpenRCT2;
 
+struct ObjectIdentifierHash
+{
+    size_t operator()(const std::string_view& identifier) const
+    {
+        std::string copy = identifier.data();
```
IntelOrca, line 54: This is not a good idea. If string_view is not null terminated, then this will cause issues. You should instead be doing `std::string copy = identifier`.

However this function can just iterate over the string_view. You don't need to copy it to a std::string.

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -279,6 +306,16 @@ class ObjectRepository final : public IObjectRepository
         return nullptr;
     }
 
+    const ObjectRepositoryItem* FindObject(const std::string_view& identifier) const override final
+    {
+        auto kvp = _newItemMap.find(identifier.data());
```
IntelOrca, line 311: Again, bad idea. You would have to do `std::string(identifier)`

IntelOrca:
Also as @ZehMatt has mentioned in chats, no `std::string_view` should be passed as `const &`. We will need to replace all existing ones at some point, but we can avoid adding any more of them.

Gymnasiast:
Sure, how _should_ I pass them? Just a naked `std::string_view` with no `const &`?

ZehMatt:
> Sure, how _should_ I pass them? Just a naked `std::string_view` with no `const &`?

Yes, string_view is a borrow type so its a cheap copy.

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -47,6 +47,27 @@
...
+        uint32_t hash = 5381;
+        for (auto i : identifier)
+        {
+            hash = ((hash << 5) + hash) + i;
+        }
+        return hash;
+    }
+};
+
+struct ObjectIdentifierEqual
```
ZehMatt, line 63: This struct should not be necessary, std::equal_to should work with string_view

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -457,6 +499,10 @@ class ObjectRepository final : public IObjectRepository
             auto copy = item;
             copy.Id = index;
             _items.push_back(copy);
+            if (item.Identifier != "")
```
ZehMatt, line 502:
```suggestion
            if (!item.Identifier.empty())
```

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -68,13 +89,14 @@ struct ObjectEntryEqual
     }
 };
 
+using ObjectIdentifierMap = std::unordered_map<std::string, size_t, ObjectIdentifierHash, ObjectIdentifierEqual>;
```
ZehMatt, line 92: Any particular reason you are using ObjectIdentifierHash isntead of std::hash<std::string>?

Gymnasiast: Lack of C++ skills.

src/openrct2/object/ObjectRepository.cpp
```diff
@@ -68,13 +68,15 @@ struct ObjectEntryEqual
     }
 };
 
+using ObjectIdentifierMap = std::unordered_map<
+    std::string, size_t, std::hash<std::string_view>, std::equal_to<std::string_view>>;
```
ZehMatt, line 72: You should be able to omit the last two template parameter, unordered-map has those as defaults
```class _Hasher = hash<_Kty>, class _Keyeq = equal_to<_Kty```

Gymnasiast: So this line should read `using ObjectIdentifierMap = std::unordered_map<std::string, size_t>`?

ZehMatt: Yeah

IntelOrca (approved):

PR #13381, opened 2020-11-05
Close #13359: Make GameAction flags optional in plugin API.

Sadret:
Also marks `WindowDesc->tabIndex` as optional in the .d.ts file.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -1866,7 +1866,7 @@ declare global {
         widgets?: Widget[];
         colours?: number[];
         tabs?: WindowTabDesc[];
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
You'll need another script api increment. I'd like @IntelOrca to take a look see if there is a better way for the flags to be optional

Sadret:
I thought we don't need another API increment because it is just a bug fix of a thing recently changed.
On the other hand, you are right and it doesn't hurt us to do it.
I'll push it.

I too would be interested in a better solution, I just wanted to have a fix as soon as possible, because the bug is actually breaking things.

IntelOrca (approved):

PR #13382, opened 2020-11-06
Use new identifier for default objects

Gymnasiast:


src/openrct2/EditorObjectSelectionSession.h
```diff
@@ -16,6 +16,7 @@
 
 enum EDITOR_INPUT_FLAGS
 {
+    INPUT_FLAG_EDITOR_OBJECT_0 = (1 << 0),
```
Gymnasiast, line 19: This flag is used to select an object - leaving it out deselects it.

src/openrct2/EditorObjectSelectionSession.h
```diff
@@ -16,6 +16,7 @@
...
     INPUT_FLAG_EDITOR_OBJECT_1 = (1 << 1),
     INPUT_FLAG_EDITOR_OBJECT_2 = (1 << 2),
```
Gymnasiast, line 21: This flag is used to select all objects from a scenario group.

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -342,9 +342,13 @@ static void window_editor_object_selection_select_default_objects()
 {
     if (_numSelectedObjectsForType[0] == 0)
     {
-        for (const auto& defaultSelectedObject : DefaultSelectedObjects)
+        for (auto defaultSelectedObject : DefaultSelectedObjects)
```
duncanspumpkin, line 345: This should still be `const`

IntelOrca (approved):

src/openrct2/EditorObjectSelectionSession.cpp
tupaschoal: This was marked as resolved but still isn't `const` as before

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -408,16 +412,15 @@ static void set_object_selection_error(uint8_t is_master_object, rct_string_id e
  *
  *  rct2: 0x006AB54F
  */
-int32_t window_editor_object_selection_select_object(uint8_t bh, int32_t flags, const rct_object_entry* entry)
+bool window_editor_object_selection_select_object(uint8_t isMasterObject, int32_t flags, const ObjectRepositoryItem* item)
```
tupaschoal, line 415: What is a "master object"? Seems weird to have something that "isSomething" but isn't a boolean.

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -483,44 +486,57 @@ int32_t window_editor_object_selection_select_object(uint8_t bh, int32_t flags,
...
 
         _numSelectedObjectsForType[objectType]++;
 
         *selectionFlags |= OBJECT_SELECTION_FLAG_SELECTED;
-        return 1;
+        return true;
     }
 }
 
+bool window_editor_object_selection_select_object(uint8_t isMasterObject, int32_t flags, std::string_view identifier)
```
tupaschoal, line 527: Should this be title case?

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -483,44 +486,57 @@ int32_t window_editor_object_selection_select_object(uint8_t bh, int32_t flags,
...
+{
+    auto& objectRepository = OpenRCT2::GetContext()->GetObjectRepository();
+    const auto* item = objectRepository.FindObject(identifier);
+    return window_editor_object_selection_select_object(isMasterObject, flags, item);
+}
+
+bool window_editor_object_selection_select_object(uint8_t isMasterObject, int32_t flags, const rct_object_entry* entry)
```
tupaschoal, line 534: Should this be title case?

src/openrct2/EditorObjectSelectionSession.h
```diff
@@ -16,8 +16,9 @@
 
 enum EDITOR_INPUT_FLAGS
 {
+    INPUT_FLAG_EDITOR_OBJECT_SELECT = (1 << 0), // Set when you want to select an object, not set when you want to deselect it.
```
tupaschoal, line 19:
```suggestion
    INPUT_FLAG_EDITOR_OBJECT_SELECT = (1 << 0), // Set when you want to select an object, unset when you want to deselect it.
```

src/openrct2/EditorObjectSelectionSession.cpp
Gymnasiast: I took the name from the underlying function. I'm not 100% sure it is correct, which is why I didn't make any changes to the data type yet.

src/openrct2/EditorObjectSelectionSession.cpp
Gymnasiast: Perhaps better taken as part of a window cleanup. (Would especially be nice to make it a method if we were to use subclassing.)

src/openrct2/EditorObjectSelectionSession.cpp
tupaschoal: Ah, sure, I'm not questioning, I'm trying to understand because I have no clue hahaha

PR #13389, opened 2020-11-07
Fix #12895: Prevent Mechanics from fixing just fixed rides

### LINKED ISSUE #12895: Mechanic Heading to Ride that isn't Broken
**OS:** Windows 10
**Version:** 0.3.0
**Commit/Build:** 11ebe78

I found some similar issues but not one quite like this.  I have mechanics called to fix a ride that is not broken down.  They are delayed in arriving, so I get a message that the ride still is not fixed.  (I suppose that message is generated by how long the mechanic is heading for the ride, not how long the ride has been broken down.) 

I guess in Alton Towers, due to the large size and twisted path layouts, this issue becomes more obvious.

- [ ] Reproducible in RCT2 (vanilla)? _no_
- [ ] Specific to multiplayer? _no_

**Steps to reproduce:**
1. Start Alton Towers
2. Fire all the mechanics and hire a few new ones (to remove patrol areas)
3. Set all rides to 10-min inspections, so mechanics are very busy
4. Wait and watch mechanics go to repair rides that aren't broken
5. Eventually, you'll get a message that a ride still is not fixed. You'll see that the mechanic is heading for a ride that isn't broken down. 
6. The mechanic visits the ride that is not broken down and does his repair routine.

**Dump file**

**Screenshots / Video:**

**Save game:**

I will try to capture it.

adam-bloom:
The state machine for finishing repairs previously unset the ride's mechanic status prior to advancing to the next state (and before marking the ride as repaired). This caused the ride's `ride_mechanic_status_update` to trigger a new mechanic call when the ride's mechanic state was undefined and the ride state was still broken down, which previously occurred while the mechanic was performing the call action to reopen the ride.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,7 +3,7 @@
 - Feature: [#13057] Make GameAction flags accessible by plugins.
 - Feature: [#13376] Open custom window at specified tab.
 - Change: [#13346] Change FootpathScenery to FootpathAddition in all occurrences.
-- Fix: [#12895] Mechanics no longer are called to repair rides that were just re-opened
+- Fix: [#12895] Mechanics are called to repair rides that have already been fixed.
 - Fix: [#13334] Uninitialised variables in CustomTabDesc.
 - Fix: [#13342] Rename tabChange to onTabChange in WindowDesc interface.
 - Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -3,6 +3,7 @@
 - Feature: [#13057] Make GameAction flags accessible by plugins.
 - Feature: [#13376] Open custom window at specified tab.
 - Change: [#13346] Change FootpathScenery to FootpathAddition in all occurrences.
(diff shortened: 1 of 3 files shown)


Gymnasiast (approved):

tupaschoal (commented):
This will require a changelog entry.

Edit this file with your fix: https://github.com/OpenRCT2/OpenRCT2/blob/develop/distribution/changelog.txt
Note that we make it sorted by the issue or PR number and type of change.

distribution/changelog.txt
```diff
@@ -3,6 +3,7 @@
 - Feature: [#13057] Make GameAction flags accessible by plugins.
 - Feature: [#13376] Open custom window at specified tab.
 - Change: [#13346] Change FootpathScenery to FootpathAddition in all occurrences.
+- Fix: [#12895] Mechanics no longer are called to repair rides that were just re-opened
```
Gymnasiast, line 6:
```suggestion
- Fix: [#12895] Mechanics are called to repair rides that have already been fixed.
```
We always describe the bug here, not the solution.

duncanspumpkin (approved):
Increment the network version and this is good to go

tupaschoal (approved):

PR #13391, opened 2020-11-08
Give all legacy objects a new-style identifier

Gymnasiast:


src/openrct2/object/ObjectFactory.cpp
```diff
@@ -38,6 +38,7 @@
 #include "WaterObject.h"
 
 #include <algorithm>
+#include <src/openrct2/util/Util.h>
```
Gymnasiast, line 41: Effing CLion...

src/openrct2/core/String.cpp
```diff
@@ -745,6 +745,50 @@ namespace String
 #endif
     }
 
+    std::string ToLower(const std::string_view& src)
+    {
+#ifdef _WIN32
```
ZehMatt, line 750:
```suggestion
#if defined(_WIN32) && _WIN32_WINNT >= 0x060
```

Gymnasiast: That would break compilation for WINNT5.1.

It should be noted that this function is a straight copy of the `ToUpper` function, with the relevant bits changed.

ZehMatt: So for NT 5.1 systems you never get a lowercase?

Gymnasiast: Indeed. Unless someone were to implement it.

IntelOrca: I think identifiers should be case sensitive or invariant culture ToLower. This means you want the implementation to be same regardless of the user's culture. A simple ASCII implementation may be best.

Gymnasiast:
After some deliberation I have decided it's better leave this to JSON objects. These provide a clear mapping from old DAT ids to new JSON ones.

I might still port the `ToUpper`/`ToLower` implementations to another PR.

PR #13394, opened 2020-11-08
Check if platform has a file picker

Gymnasiast:
Hides buttons and options related to using the native file picker if the OS does not provide one. This avoids some footshooting on Android and barebones Linux and BSD installs.


src/openrct2/ui/UiContext.h
```diff
@@ -117,6 +117,7 @@ namespace OpenRCT2
             virtual void OpenURL(const std::string& url) abstract;
             virtual std::string ShowFileDialog(const FileDialogDesc& desc) abstract;
             virtual std::string ShowDirectoryDialog(const std::string& title) abstract;
+            virtual bool HasFilePicker() abstract;
```
ZehMatt, line 120:
```suggestion
            virtual bool HasFilePicker() const abstract;
```
If the state of the object/class is not modified the functions should be marked const

PR #13415, opened 2020-11-13
Fix #13409: Peeps sometimes stray too far from the path centre

### LINKED ISSUE #13409: Peeps are sometimes a little bit *too* debunched.
**OS:** Kubuntu Linux 20.04
**Version:** ??? Version info still not showing up on Linux title screen -- updated as of this post
**Commit/Build:** ???

I'm a huge fan of  #12917, but it seems that things have gone a bit too far. In some cases, peeps are _so_ debunched that they walk completely off of the path. This sometimes even puts them on the wrong side of trash cans and guardrails.

This happens fairly rarely, but in even a moderately busy park, it's common enough that you can spot it happening frequently if you're looking for it. The screenshots below were all captured within a few minutes of each other in a park with only about 1000 guests in it.

Not that it isn't realistic to have some guests wandering outside the bounds of your paths, but I don't think this is the intended behavior. The peeps' viable path choices probably just need to be narrowed down a little bit.

**Steps to reproduce:**
1. Open a fairly busy park.
2. Wait and watch for peeps walking next to the path instead of on the path.

**Screenshots / Video:**

Quanahog Winter Games 2020-11-11 17-39-32
Quanahog Winter Games 2020-11-11 17-40-09
Quanahog Winter Games 2020-11-11 17-40-59
Quanahog Winter Games 2020-11-11 17-41-04
Quanahog Winter Games 2020-11-11 17-41-13

richard-fine:
The debunch-peeps improvement did not account for the fact that peeps may change direction when they are only at the very edge of a tile - it's not entirely clear to me why this happens, but it does. The previous code would push these peeps back towards the center line over time, but the new behaviour allows them to keep walking along these very edge-y lines, which means they sometimes appear to be walking on the wrong sides of benches, railings, etc.

To fix the problem, do not simply keep the target coordinate constant for the direction the peep is moving in, but clamp it, so that peeps in these outlier positions will get pulled back to a more acceptable position within one tile. Peeps who are already debunched within the reasonable center area of the path are unaffected.

### DIFF
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -148,18 +148,27 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
         //
         // What we want instead is to apply that randomness in the direction they are walking ONLY, and keep their
         // other coordinate constant.
+        //
+        // However, we have also seen some situations where guests end up too far from the center of paths. We've
+        // not identified exactly what causes this yet, but to limit the impact of it, we don't just keep the other
+        // coordinate constant, but instead clamp it to an acceptable range. This brings in 'outlier' guests from
+        // the edges of the path, while allowing guests who are already in an acceptable position to stay there.
 
         int8_t offset = (scenario_rand() & 7) - 3;
         if (direction == 0 || direction == 2)
         {
-            // Peep is moving along X, so apply the offset to the X position of the destination and keep their current Y
+            // Peep is moving along X, so apply the offset to the X position of the destination and clamp their current Y
             peep->DestinationX += offset;
-            peep->DestinationY = peep->y;
+            const uint16_t centerLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
+            peep->DestinationY = std::clamp(
+                peep->y, static_cast<int16_t>(centerLine - 3), static_cast<int16_t>(centerLine + 3));
         }
         else
         {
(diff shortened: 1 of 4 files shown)


richard-fine:
This will need the bpb replay updating (again).

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -148,18 +148,27 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
         //
         // What we want instead is to apply that randomness in the direction they are walking ONLY, and keep their
         // other coordinate constant.
+        //
+        // However, we have also seen some situations where guests end up too far from the center of paths. We've
+        // not identified exactly what causes this yet, but to limit the impact of it, we don't just keep the other
```
duncanspumpkin, line 153: Isn't it just when the destination tolerance is cranked up to a large value like during aimless wandering mode.

duncanspumpkin (approved):
I don't quite agree with the comment. I think its pretty much going to occur in one of the cases where the tolerance is cranked up to a large number. There are quite a few different tolerance values that can be used depending on a variety of different situations so I'm not too surprised its not seen often. Either way. The fix is fine and I'll get those replays merged.

richard-fine:
Yeah, I guess it's more accurate to say "@richard-fine does not know" rather than "we don't know" :) I did try a brief review of all assignments to destinationTolerance (as well as checking specific things like using bins/benches and spectating on rides) but I couldn't quickly find the ones that seemed big.

PR #13419, opened 2020-11-15
Disentangle sub_6E1F34() and try_place_ghost_scenery()

Gymnasiast:
Code might be easier to follow when looking at commits one by one.


duncanspumpkin:
I've puzzled over how to break up this code for a long time. Good work. I still think there must be a way to make it less repetitive but its good progress.

PR #13423, opened 2020-11-15
Remove built-in explode guests cheat

Gymnasiast:


src/openrct2/localisation/StringIds.h
```diff
@@ -2758,8 +2758,7 @@ enum
     STR_THEMES_OPTION_RCT1_RIDE_CONTROLS = 5282,
     STR_THEMES_OPTION_RCT1_PARK_CONTROLS = 5283,
     STR_THEMES_OPTION_RCT1_SCENARIO_SELECTION_FONT = 5284,
-    STR_CHEAT_EXPLODE = 5285,
-    STR_CHEAT_EXPLODE_TIP = 5286,
+
```
tupaschoal, line 2761: Do we want this new line?

tupaschoal (commented):
Looks ok, I found it kind of a useless cheat :|

src/openrct2/localisation/StringIds.h
Gymnasiast: I put it in because there is a gap, and also because the strings following are from a different area than the ones preceding.

Gymnasiast:
> Looks ok, I found it kind of a useless cheat :|

It was one of the first cheats, so it had a special status. But there is now a replacement plug-in available.

Stevoisiak:
> > Looks ok, I found it kind of a useless cheat :|
> 
> It was one of the first cheats, so it had a special status. But there is now a replacement plug-in available.

Where can I download this plugin?

Gymnasiast:
https://github.com/Gymnasiast/everything-must-die/releases

PR #13429, opened 2020-11-18
Add ride category to plugin API

mgovea:
This is a small change. I want to have different behavior for different ride categories for my plugin, so I added it.

One quirk I saw: there can be up to `RCT2_MAX_RIDE_TYPES_PER_RIDE_ENTRY` (3) ride types per ride entry, but in every case I tested, I got `[realCategory, "transport", "transport"]` (e.g. `["water", "transport", "transport"]` or `["shop", "transport", "transport"]` or `["transport", "transport", "transport"]`) so I just took the 1st category. This is the logic in `category_get`.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -775,6 +775,12 @@ declare global {
         type: number;
 
         /**
+         * The category of a ride. Same as the tab from ride building window.
+         * (e.g. Transport, Gentle, Coaster, Thrill, Water, Shop)
+         */
+        readonly category: RideCategory;
+
+        /**
          * Whether the ride is a ride, shop or facility.
          */
         readonly classification: RideClassification;
@@ -905,6 +911,8 @@ declare global {
          */
         value: number;
     }
+
+    type RideCategory = "transport" | "gentle" | "rollercoaster" | "thrill" | "water" | "shop" | "";
 
     type RideClassification = "ride" | "stall" | "facility";
 
--- a/src/openrct2/scripting/ScRide.hpp
+++ b/src/openrct2/scripting/ScRide.hpp
@@ -14,6 +14,7 @@
 #    include "../Context.h"
 #    include "../common.h"
 #    include "../ride/Ride.h"
+#    include "../ride/RideData.h"
 #    include "Duktape.hpp"
 #    include "ScObject.hpp"
 #    include "ScriptEngine.h"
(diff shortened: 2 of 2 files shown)


distribution/openrct2.d.ts
```diff
@@ -906,6 +912,8 @@ declare global {
         value: number;
     }
 
+    type RideCategory = "transport" | "gentle" | "rollercoaster" | "thrill" | "water" | "shop" | "";
```
mgovea, line 915: Thinking about `""`, I realize other enums like `RideClassification` don't include the empty string, even though it's possible in the code for the empty string to be passed under some circumstances. I might want to remove it. Thoughts?

Gymnasiast:
I will try to review this shortly.

distribution/openrct2.d.ts
tupaschoal: I'd remove it, seems like you wouldn't want to take any action on an empty type anyway.

src/openrct2/scripting/ScRide.hpp
```diff
@@ -229,6 +230,35 @@ namespace OpenRCT2::Scripting
             return "";
         }
 
+        std::string category_get() const
+        {
+            auto ride = GetRide();
+            if (ride != nullptr)
+            {
+                for (auto rideType : ride->GetRideEntry()->ride_type)
```
tupaschoal, line 238: I don't quite get this for, given that it seems like it will always return on the first iteration

Gymnasiast: It is incorrect anyway. You should read the ride type from the ride itself (instead of the ride entry), then use `gRideCategories` to look up the category.

Gymnasiast (changes requested):
@mgovea Please address the line note ( https://github.com/OpenRCT2/OpenRCT2/pull/13429#discussion_r529886324 ).

mgovea:
I've been planning on making changes, and I didn't anticipate you merging as-is. I just haven't gotten around to it.

Though I have one question. It makes sense to get the type directly from the ride, but I'm not sure what you're referring to when you said

> use `gRideCategories` to look up the category

I am confused. I searched around the codebase a bit, and I wasn't able to find what you're referring to. Can you give me more info on what you mean or pointers where to look?

Gymnasiast:
gRideCategories used to exist, but it has been superseded.

What you need to do is fetch the ride type descriptor from the ride, then take the category from that.

Gymnasiast:
Closing as there has been no update for more than a month. We can reopen it if you want to resume working on it.

PR #13434, opened 2020-11-20
closes #12446 Refactor WEATHER to use strong enum

### LINKED ISSUE #12446: Refactor WEATHER to use strong enum

transistive:
This is my first attempt at some real-world c++.

First, I ended up at the Weather.h file, where I found a lot of numbers. Being a little overenthusiastic, I thought this was what the issue was about, so I encapsulated it in an enum class of uint8_t.

Later on, I found the actual enum that needed to be refactored, so I did that as well. I tried to interpret the code as best I could and only change what was needed. Once I refactored the enumeration, I just fixed all the subsequent compiler errors until it was able to do its job. 

The game appears to run fine. The weather still changes and is changeable with cheats. But I do not know how to run tests or anything... All feedback is highly appreciated.

Kind regards,

Ghlen


James103:
In order for the tests to run automatically, the code must be in the correct format for this project, as outlined by [the coding style](https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style). You can apply the changes that the `clang-format` run gives to correctly format the code.

James103:
There's still a few problems with the code formatting.

transistive:
Thank you for the swift reply @James103! My bad.
I think I fixed it all, but the CI keeps telling me I'm wrong.
Is this the correct command with the corresponding output?

image

James103:
Could you try running the command, push changes, and see what happens? I'd recommend making a backup first.

transistive:
It finally got through the ci! Sorry for the inconvenience. I hope this will be accepted or at least get some honest feedback.

Kind regards,
Ghlen

src/openrct2/interface/Screenshot.cpp
```diff
@@ -550,15 +550,9 @@ int32_t cmdline_for_gfxbench(const char** argv, int32_t argc)
...
+    if (options->weather != WeatherType::Sunny && options->weather != WeatherType::Count)
     {
-        if (options->weather < 1 || options->weather > 6)
-        {
-            throw std::runtime_error("Weather can only be set to an integer value from 1 till 6.");
-        }
-
-        uint8_t customWeather = options->weather - 1;
-        climate_force_weather(customWeather);
+        climate_force_weather(weatherTypes[EnumValue(options->weather) - 1]);
```
tupaschoal, line 555: Not sure this `weatherTypes` solution is the best, why not just use it directly? You've changed `ScreenshotOptions` to use a WeatherTYpe already, so there shouldn't be any conversion.

src/openrct2/rct1/RCT1.h
```diff
@@ -685,8 +686,8 @@ struct rct1_s4
     uint8_t climate;
     uint8_t unk_1F432B;
     uint16_t climate_timer;
-    uint8_t weather;
-    uint8_t target_weather;
+    WeatherType weather;
+    WeatherType target_weather;
```
tupaschoal, line 690: @Gymnasiast  can correct me if I'm wrong, but we don't change `rct1` structs.

src/openrct2/drawing/Weather.h
```diff
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2014-2020 OpenRCT2 developers
+ * Copyright (c) 2N14-2N2N OpenRCT2 developers
```
tupaschoal, line 2:
```suggestion
 * Copyright (c) 2014-2020 OpenRCT2 developers
```

Guess this was just replacing more than intended

src/openrct2/world/Climate.cpp
```diff
@@ -230,8 +230,21 @@ bool climate_is_raining()
...
+    else
+    {
+        return false;
+    }
+}
+
+bool climate_is_dry()
+{
+    if (gClimateCurrent.Weather == WeatherType::Sunny || gClimateCurrent.Weather == WeatherType::PartiallyCloudy
+        || gClimateCurrent.Weather == WeatherType::Cloudy)
```
tupaschoal, line 247:
```suggestion
    return gClimateCurrent.Weather == WeatherType::Sunny || gClimateCurrent.Weather == WeatherType::PartiallyCloudy
        || gClimateCurrent.Weather == WeatherType::Cloudy;
```

This and all other `climate_is` functions can be a lot simpler by just returning directly.

tupaschoal (commented):
Hello Ghlen, thanks for your work! I think it is a good start, but it is a bit hard to review as you mixed different conceptual changes.

Can you split it into multiple commits? One for the actual enum to strong enum conversion, another for the flags/droplet sizes (and more), this will make it easier for us to review. Note this will require some git knowledge, if you need help, let us know in our Discord chat.

src/openrct2/interface/Screenshot.cpp
transistive: Thank you for the great feedback! I used the weatherTypes array here because I first needed the enum value to be able to subtract one from it. 

I kept the subtraction but to be fair, it seems like either a bug or a hack to me.

When you say I can use it directly, does that imply I am allowed to subtract 1 from an enumeration and then cast it back to an enum?

src/openrct2/rct1/RCT1.h
transistive: I will wait for feedback on this before I change it. I would also be interested as into why that would be. Seems like a learning opportunity.

transistive:
@tupaschoal I have now also removed the droplet enumeration all together. It seems the only reason I kept it was because I had already implemented it without really thinking about it. Not a good enough reason :shrug: 

Looking forward to your feedback.

src/openrct2/interface/Screenshot.cpp
tupaschoal: I hadn't noticed the subtraction, so you might be right, though I guess it's better to subtract and cast it back (?)

src/openrct2/rct1/RCT1.h
duncanspumpkin: No don't change the rct1 structs. Leave as uint8_t. These structures are only used for reading from the binary data. We need to keep the memory layout exactly the same. What is meant to happen is that after reading the data, it should be validated to ensure that it is correct for our type. (we might not do that last step very often). By keeping the types different it ensures that there has to be a conscious decision to validate the type on load.

src/openrct2/scenario/Scenario.h
```diff
@@ -278,8 +279,8 @@ struct rct_s6_data
     uint8_t climate;
     uint8_t pad_013CA747;
     uint16_t climate_update_timer;
-    uint8_t current_weather;
-    uint8_t next_weather;
+    WeatherType current_weather;
```
duncanspumpkin, line 282: Same as the rct1 struct please keep the type as `uint8_t`.

src/openrct2/world/Climate.cpp
```diff
@@ -450,45 +456,52 @@ const WeatherState ClimateWeatherData[WEATHER_COUNT] = {
     { -20, WeatherEffectType::Blizzard, 2, WeatherLevel::Heavy, SPR_WEATHER_SNOW },  // Blizzard
 };
 
+constexpr auto S = WeatherType::Sunny;
+constexpr auto P = WeatherType::PartiallyCloudy;
+constexpr auto C = WeatherType::Cloudy;
+constexpr auto R = WeatherType::Rain;
+constexpr auto H = WeatherType::HeavyRain;
+constexpr auto T = WeatherType::Thunder;
```
duncanspumpkin, line 464: Nice idea this would have been horrible to read otherwise.

src/openrct2/world/Climate.h
```diff
@@ -21,20 +22,26 @@ enum class ClimateType : uint8_t
...
+    Rain,
+    HeavyRain,
+    Thunder,
+    Snow,
+    HeavySnow,
+    Blizzard,
+    Count
 };
 
+constexpr WeatherType weatherTypes[EnumValue(WeatherType::Count)] = { WeatherType::Sunny,     WeatherType::PartiallyCloudy,
```
duncanspumpkin, line 39: I don't see the reason to have this and runs in contrast to the rest of the codebase. Just use EnumValue where required.

src/openrct2/interface/Screenshot.cpp
duncanspumpkin: This function is already performing all the validation to make sure that doing -1 on the value will produce the correct result. I'd prefer if you just used EnumValue directly.

src/openrct2/world/Climate.cpp
```diff
@@ -241,6 +241,12 @@ bool climate_is_snowing()
     }
 }
 
+bool climate_is_dry()
```
duncanspumpkin, line 244: All new functions are to be in TitleCase. Do you think this would be more useful to take an input of `WeatherType` and return if it is dry? i.e. the caller would pass in `gClimateCurrent.Weather`? Not sure myself I'll leave for you to decide.

transistive: You are absolutely correct. I just followed the examples of climate_is_snowing etc. I have now changed the function to `weatherIsDry(WeatherType)` as I feel that name makes a lot more sense as well in this case.

src/openrct2/world/Climate.h
transistive: It was there because I didn't know you could do this: `WeatherType{ <uint8_t argument> }`. It is fixed now

src/openrct2/scenario/Scenario.h
transistive: Done :)

src/openrct2/interface/Screenshot.cpp
transistive: I have changed it now :) I used the underlying enumvalue, subtracted one and then passed it to the constructor (?) if that's what this is called: `WeatherType{ }`

src/openrct2/world/Climate.cpp
duncanspumpkin: Just need to capitalise the w. WeatherIsDry

duncanspumpkin (approved):

src/openrct2/world/Climate.cpp
transistive: Done in #b29567c

src/openrct2/drawing/Weather.h
```diff
@@ -20,21 +20,21 @@ namespace OpenRCT2::Drawing
...
+        32, 32, 0, 32, 0, 32, 0, 16, 255, 0, 255, 0, 255, 0, 255, 0, 255,
+        0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0,
+        255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255,
+        0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 255, 0, 0, 0
+    };
 
 // clang-format on
 
-void DrawWeather(rct_drawpixelinfo* dpi, OpenRCT2::Drawing::IWeatherDrawer* weatherDrawer);
+void DrawWeather(rct_drawpixelinfo* dpi, OpenRCT2::Drawing::IWeatherDrawer* weatherDrawer);
```
duncanspumpkin, line 40: Not sure what has happened to this file but could you revert it. `git checkout develop src/openrct2/drawing/Weather.h`

transistive: I accidentally removed the newline. I added it back again.

transistive (commented):
Sorry for opening the review, I might have pressed it accidentally but I cannot remove it now. God, I am such a beginner :facepalm:

src/openrct2/rct1/RCT1.h
Gymnasiast: FWIW, I typically define an RCT1/2 counterpart - if it's a `typedef` and not an `enum`, at least.

PR #13438, opened 2020-11-21
Fix writing of objects.idx

IntelOrca:
Since 656cdda3, objects.idx is written in a corrupted state and prevents all objects loading correctly on next run of OpenRCT2. Use WriteString instead of WriteValue for std::string.


duncanspumpkin:
Will we need to force a rebuild of the object cache by incrementing its number?

IntelOrca:
@duncanspumpkin done

duncanspumpkin (approved):

PR #13444, opened 2020-11-23
Extract ifdefed functions

tupaschoal:
Easier to follow per commit. I'm just moving stuff that branches on platform to the actual platform

### DIFF
--- a/src/openrct2/core/Path.cpp
+++ b/src/openrct2/core/Path.cpp
@@ -212,11 +212,7 @@ namespace Path
 
     bool Equals(const utf8* a, const utf8* b)
     {
-        bool ignoreCase = false;
-#ifdef _WIN32
-        ignoreCase = true;
-#endif
-        return String::Equals(a, b, ignoreCase);
+        return String::Equals(a, b, Platform::ShouldIgnoreCase());
     }
 
     std::string ResolveCasing(const std::string& path)
--- a/src/openrct2/core/Path.cpp
+++ b/src/openrct2/core/Path.cpp
@@ -7,10 +7,7 @@
  * OpenRCT2 is licensed under the GNU General Public License version 3.
  *****************************************************************************/
 
-#include <algorithm>
-#ifndef _WIN32
-#    include <dirent.h>
-#endif
+#include "Path.hpp"
 
 #include "../localisation/Language.h"
 #include "../platform/Platform2.h"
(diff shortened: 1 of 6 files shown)


src/openrct2/core/FileScanner.cpp
```diff
@@ -23,335 +23,156 @@
...
+}
 
 IFileScanner* Path::ScanDirectory(const std::string& pattern, bool recurse)
 {
-#ifdef _WIN32
-    return new FileScannerWindows(pattern, recurse);
-#elif defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
-    return new FileScannerUnix(pattern, recurse);
-#endif
+    return Platform::ScanDirectory(pattern, recurse);
```
duncanspumpkin, line 175: Add a Todo saying replace with std::unique_ptr and create an issue to track it.

tupaschoal: Done

Gymnasiast:
I'd prefer if the string functions were more clearly labelled as such: `StringToUpper`, etc.
Also, the functions `ToWidechar` and `ToUTF8` lack some context and should be called `UTF8ToWideChar` and `WidecharToUTF8`. That is, if they even _belong_ here - they're only used on Windows, as macOS/Linux/BSD/Android all use UTF-8 directly.

IntelOrca:
> That is, if they even belong here - they're only used on Windows, as macOS/Linux/BSD/Android all use UTF-8 directly.

It might have been OpenLoco, but I think sometimes widechar is used for some APIs. Their widechar differs in that it is 32-bit rather than 16-bit - but the same named function can still be used in those instances.

tupaschoal:
Took another route first, but could move some stuff out to the Platform, finally.

src/openrct2/core/Path.cpp
```diff
@@ -202,42 +191,15 @@ namespace Path
         return lastDot;
     }
 
-    utf8* GetAbsolute(utf8* buffer, size_t bufferSize, const utf8* relativePath)
+    utf8* GetAbsolutePath(utf8* buffer, size_t bufferSize, const utf8* relativePath)
     {
```
duncanspumpkin, line 195: I don't agree with this rename. We are already in the path namespace. Also you've only renamed one of the overloads

tupaschoal: I thought I had dropped this renaming and left the new name only on Platform, but seems like I hadn't. Just rewrote it.

duncanspumpkin (approved):

PR #13453, opened 2020-11-25
Origin/12410

ffrogman:
Renamed and converted enum to a strong enum. Moved enum declaration from Widget.h to Window.h because the rct_widget struct should use the enumeration instead of a uint8_t.


src/openrct2-ui/scripting/ScWidget.hpp
```diff
@@ -69,40 +69,42 @@ namespace OpenRCT2::Scripting
...
+                    case WindowWidgetType::Scroll:
                         return "scroll_view";
-                    case WWT_CHECKBOX:
+                    case WindowWidgetType::Checkbox:
                         return "checkbox";
-                    case WWT_TEXT_BOX:
+                    case WindowWidgetType::TextBox:
                         return "textbox";
+                    default:
+                        break;
```
tupaschoal, line 107: I think we shouldn't have a default in this case, as it's a plugin API (specially if we add a new widget type later, it's easy to forget. How many enumerations are missing here?

src/openrct2/interface/Window.h
```diff
@@ -78,18 +78,45 @@ namespace WIDGET_FLAGS
     const WidgetFlags TOOLTIP_IS_STRING = 1 << 4;
 } // namespace WIDGET_FLAGS
 
+enum class WindowWidgetType : uint8_t
```
tupaschoal, line 81: Any reason for moving this over to `Window.h`? It's a widget thing, I think it should be kept there. If you need it here for the `rct_widget`, can't you just forward declare it as:
```c++
enum class WindowWidgetType : uint8_t;
```

And leave the definition there?

src/openrct2/interface/Widget.h
```diff
@@ -13,43 +13,16 @@
...
-    WWT_DROPDOWN = 16,
-    WWT_VIEWPORT = 17,
-    WWT_GROUPBOX = 19,
-    WWT_CAPTION = 20,
-    WWT_CLOSEBOX = 21,
-    WWT_SCROLL = 22,
-    WWT_CHECKBOX = 23,
-    WWT_PLACEHOLDER = 25,
-    WWT_TEXT_BOX = 27,
-    WWT_LAST = 26,
```
tupaschoal, line 40: Please restore the explicit number assignment for now. I know it is done automatically, but if you take a close look we skip some number like `4, 5, 11, 18, 24` and I don't know currently if that will do any harm.

tupaschoal (commented):
Thanks for this patch, specially for keeping the widget creation macro alignments the same!

I have a few minor comments.

src/openrct2-ui/scripting/ScWidget.hpp
ffrogman: That makes sense. It was "empty", "placeholder", and "last" left. I think these are sort of dummy values, but I added them and removed the default case.

PR #13457, opened 2020-11-25
Fix #13427: Newly created Go-Karts show "Race won by <blank>"

### LINKED ISSUE #13427: Newly opened Go-Karts track shows the status message of: "Race won by" with the name blank.
**OS:** Kubuntu Linux 20.04
**Version:** ??? Version info still not showing up on Ubuntu title screen (see #12909) updated as of this post
**Commit/Build:** ???

The first time a Go-karts is opened, the status message at the bottom of the window says 'Race won by' and just leaves it at that. Seems like it would make more sense for it to say 'X guests on ride' or something to that effect or even just 'open' if the track has never been run and it's never been won by anyone, right?

Really minor thing, really, but it seems like it could be improved ... and *maybe* it wouldn't be difficult to do so?

**Steps to reproduce:**
1. Build a complete go-karts track
2. Close the construction window (the ride window opens)
3. Open the ride (probably needs to be in the default race mode, not continuous circuit)
4. Ride status message now reads 'Race won by _______'

**Screenshots / Video:**
Forest Frontiers 2020-11-16 22-04-26

ffrogman:
Fix #13427. This was caused in the fix for issue #12497 0f78d452.

race_winner is null when the ride switches from closed to open. The fix for issue #12497 accidentally changed the logic here in an attempt to prevent a duplicate check. This resulted in "Race won by" displaying any time a ride switches from closed to open (not just the first time). This message remained until the ride actually started. For example it would jump up to say "7 people on the ride" immediately even if peeps are entering the ride slowly.

In vanilla for a ride that changed from closed to open the message says "0 people on the ride" and counts as peeps enter the ride. I tested and this fix seems to behave the same as vanilla.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -10,6 +10,7 @@
 - Fix: [#13454] Plug-ins do not load on Windows if the user directory contains non-ASCII characters.
 - Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 - Removed: [#13423] Built-in explode guests cheat (replaced by plug-in).
+- Fix: [#13427] Newly created Go-Karts show "Race won by <blank>".
 
 0.3.2 (2020-11-01)
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


duncanspumpkin (approved):

Gymnasiast (changes requested):
`- Fix: [#13427] Newly created Go-Karts show "Race won by <blank>".`

duncanspumpkin:
Sorry could you move the fix so that the fixes are grouped together in order of number.

duncanspumpkin (approved):

PR #13459, opened 2020-11-25
Plugins - return result from StaffHireAction

eezstreet:
The StaffHireAction will now return a usable result for subscribers.

### DIFF
--- a/src/openrct2/scripting/ScriptEngine.cpp
+++ b/src/openrct2/scripting/ScriptEngine.cpp
@@ -15,6 +15,7 @@
 #    include "../actions/CustomAction.hpp"
 #    include "../actions/GameAction.h"
 #    include "../actions/RideCreateAction.hpp"
+#    include "../actions/StaffHireNewAction.hpp"
 #    include "../config/Config.h"
 #    include "../core/File.h"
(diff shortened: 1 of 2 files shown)


src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -857,6 +858,14 @@ DukValue ScriptEngine::GameActionResultToDuk(const GameAction& action, const std
             obj.Set("ride", rideCreateResult.rideIndex);
         }
     }
+    else if (action.GetType() == GAME_COMMAND_HIRE_NEW_STAFF_MEMBER)
+    {
+        auto& staffHireResult = static_cast<StaffHireNewActionResult&>(*result.get());
+        if (staffHireResult.peepSriteIndex != SPRITE_INDEX_NULL)
+        {
+            obj.Set("staff", staffHireResult.peepSriteIndex);
```
IntelOrca, line 866:
```suggestion
            obj.Set("peep", staffHireResult.peepSriteIndex);
```

IntelOrca (approved):

duncanspumpkin:
Will need an api version bump for this (and three other plugin pr's)

PR #13461, opened 2020-11-26
Add scripting hooks for guest generation.

eezstreet:
Added 3 hooks for guest generation:

- `guest.generation.suggested`: This calculates the "suggested maximum" number of guests. Takes a "suggestedMaxGuests" as a parameter and returns a new value.
- `guest.generation.probability`: This calculates the probability of generating a new guest. Takes "numGuests" and "probability" and returns a probability.
- `guest.generation.marketingProbability`: This calculates the probability of generation a new guest from a marketing campaign. Takes "campaignType" and "probability" and returns a probability.

I can see a lot of potential for plugins here, such as ones that make the game harder or easier.


James103:
@eezstreet You should run `clang-format` on your changes so that the format is good and the tests can be run.

tupaschoal (commented):
This will need a PLUGIN API version increment.

Also, I'd love if those hooks were functions on their own, so that we call them from the methods but we don't pollute the methods with `IFDEF`s

James103:
> This will need a PLUGIN API version increment.

Change the `14` to a `15` in the following line:
https://github.com/OpenRCT2/OpenRCT2/blob/67f5ce62b307eec6bb27bf7177e9004c12705edd/src/openrct2/scripting/ScriptEngine.cpp#L47

IntelOrca:
Can you update `openrct2.d.ts` please.

Gymnasiast:
Since there has been no answers from OP since 26 November, I shall close it. It can be reopened if the OP becomes active again.

PR #13464, opened 2020-11-27
Update icon for macOS Big Sur

### LINKED ISSUE #13436: macOS Big Sur Icon
**OS:** macOS
**Version:** 11.0.1

Hi, macOS Big Sur has a new design for icons.
I found the SVG of the original icon, and made this alternate version for Big Sur.
If you want I can open a PR and replace the `.icns` file with this one and generate assets for all resolutions.

**Screenshots / Video:**

App Icons - @2x
Capture d’écran, le 2020-11-21 à 02 17 06

mrleblanc101:
fix #13436


pizza2004:
I don't think you need to actually include the icns file in the repository, as we weren't doing that before.  I think it just generates it when the project builds through Xcode.

mrleblanc101:
Yes, I've noticed. I don't know why you don't just use an .icns or the iconset directly, but I didn't want to take the risk to break anything. I don't think it's bad to have all the assets type, but I can remove it if you really want.

pizza2004:
It would probably make more sense to only include the icns file and the svg and then edit the project file to use the icns file.  I'm sure the reason for it to be the other way is for the sake of making sure that updating the icons is easy if necessary, but since this is Mac specific it doesn't need all the individual images unless for some reason that takes up more space.

Gymnasiast:
I think @pizza2004 's explanation is about right.

I think including an .icns and .svg file should be sufficient, if you modify the Xcode project accordingly. Testing whether something breaks should not be too difficult: just tell Xcode to compile it and see if the icon is correct.

tupaschoal:
Hey @mrleblanc101 , would you kindly tend to Gymnasiast's comments? Thanks!

pizza2004:
Since there has been no response on this for a couple of months, should someone else make the changes so it can be merged?

Gymnasiast:
@pizza2004 Yes. Would you be able to do so?

Gymnasiast:
Merged as https://github.com/OpenRCT2/OpenRCT2/pull/14268 .

Merci, monsieur Leblanc!

Issue #13318, opened 2020-10-28
Add missing ASCII characters to sprite font


IntelOrca:
The OpenRCT2 sprite fonts currently do not support some ascii characters, including:

* [ ] `{` and `}`

PR #13478, opened 2020-11-28
Added missing ASCII characters to sprite font #13318

zahramakki:
added ASCII characters: { and } and | and ~ and □ and � #13318


zahramakki:
need to add more changes. will reopen when finished!

myounus99:
We were unsure of what the official font library was for glyphs so we tried our best to match what we saw. Please let us know if we need to make any changes to the pngs, we're working on the indexing part right now as well.

src/openrct2/drawing/Font.cpp
```diff
@@ -186,6 +186,12 @@ static const std::unordered_map<char32_t, int32_t> codepointOffsetMap = {
     { UnicodeChar::narrow_non_breaking_space, ' ' - CS_SPRITE_FONT_OFFSET },
     { UnicodeChar::quote_open, CSChar::quote_open - CS_SPRITE_FONT_OFFSET },
     { UnicodeChar::quote_close, CSChar::quote_close - CS_SPRITE_FONT_OFFSET },
+    { UnicodeChar::left_curly_bracket, '{' - CS_SPRITE_FONT_OFFSET },
```
Gymnasiast, line 189: All added characters use `SPR_G2_<name> - SPR_CHAR_START`.

Gymnasiast:
You will have to modify `sprites.json` to add the characters as well. Add them after `u-breve`.

Additionally, all characters need to be added in three shapes: small (guest thoughts), bold (regular user interface) and tiny (ride banners). These have to be indexed PNGs.

All letters also need to be added to `sprites.h`.

You can use this PR to see everything that needs to be done: https://github.com/OpenRCT2/OpenRCT2/pull/12806/files

By the way, I should note that you didn't pick the easiest issue to work on, as it involves quite a bit of knowledge of how RCT2/OpenRCT2 handles sprites. If you prefer working on something else, please let us know.

myounus99:
We're converting this to a draft while we're testing out the indexing changes. We will change the status once it's ready for review and if we have any questions on how to proceed we'll definitely reach out. Thanks!

Gymnasiast:
The changes to `Font.cpp`, `FormatCodes.cpp` and `sprites.h` now look correct.

What you now need to do is make sure you have created a glyph for each character for each of the font types (small, bold and tiny). Use `resources/g2/font/j-bold.png`/`j-small.png`/`j-tiny.png` as an example. The PNGs should be indexed PNGs. It might be easiest to take a pre-existing character PNG and then modify it using [GNU Image Manipulation Program](https://www.gimp.org/).
Then, add those to `resources/g2/sprites.json`, after the bold/small/tiny definitions for the lowercase u breve, in the same order.

After that, you need to rebuild g2.dat. _How_ to do that depends on whether you use Microsoft Visual Studio, Xcode or CMake.

myounus99:
We were having some trouble rebuilding g2.dat so we're not sure if the palette indexing was done correctly. Can you take a look at that and let us know if there's anything we need to change?

src/openrct2/localisation/FormatCodes.h
```diff
@@ -322,6 +328,8 @@ namespace UnicodeChar
     constexpr char32_t tick = 0x2713;
     constexpr char32_t plus = 0x2795;
     constexpr char32_t minus = 0x2796;
+    constexpr char32_t white_square = 0x25A1;
```
Gymnasiast, line 331: These two entries should be moved to `Misc`.

src/openrct2/drawing/Font.cpp
```diff
@@ -203,6 +209,8 @@ static const std::unordered_map<char32_t, int32_t> codepointOffsetMap = {
     { UnicodeChar::tick, CSChar::tick - CS_SPRITE_FONT_OFFSET },
     { UnicodeChar::plus, '+' - CS_SPRITE_FONT_OFFSET },
     { UnicodeChar::minus, '-' - CS_SPRITE_FONT_OFFSET },
+    { UnicodeChar::white_square, SPR_G2_WHITE_SQUARE - SPR_CHAR_START },
```
Gymnasiast, line 212: Move to `Misc`.

Gymnasiast:
The changes to the code and JSON file seem to be mostly OK. The character PNGs still need a lot of work. They are not indexed yet and they also are not stylistically compatible with the rest of the font - for example, they need outline definitions in index colours 2 and 3. I modified the small apostrophe to give an example of what it should be like. I hope that makes things a bit clearer.

IntelOrca:
@zahramakki @myounus99 are you still working on this? If you no longer have time or you're finding it too difficult, we can take over the PR for you.

myounus99:
Yep feel free to take over, we've been swamped with some other stuff and haven't been able to work on this very much

Gymnasiast:
I'll take this over, but I will be working in a branch of my own, so I'll be closing this PR.

PR #13490, opened 2020-11-29
Fix #13489: Mechanics continue heading to inspect broken down rides

### LINKED ISSUE #13489: Mechanics heading to inspect a broken down ride do not update status
**OS:** macOS
**Version:** 0.3.2-117
**Commit/Build:** 55dc985c5

When a ride breaks down, if a mechanic is already heading to inspect it, then that mechanic will continue to display "heading to inspect ___" since its orders were never updated to repair the ride. However, when the mechanic arrives at the ride, it will begin to repair instead of inspect.

**Steps to reproduce:**
1. Force a breakdown while a mechanic is heading to inspect a ride.
2. Mechanic orders will not be modified to "responding".

PR incoming with the fix...

adam-bloom:
If a mechanic is already heading to a ride when it breaks down (for an inspection), the mechanic would continue to head for the ride for inspection, but would repair the ride when they arrived. With this change, the mechanic will receive a page for a repair and replace "heading for inspection" with "responding to breakdown" in the mechanic status.

Starting state:
image

Before:
image

After:
image
image

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -2627,6 +2627,7 @@ static void ride_mechanic_status_update(Ride* ride, int32_t mechanicStatus)
         case RIDE_MECHANIC_STATUS_HEADING:
         {
             auto mechanic = ride_get_mechanic(ride);
+            bool rideNeedsRepair = (ride->lifecycle_flags & (RIDE_LIFECYCLE_BREAKDOWN_PENDING | RIDE_LIFECYCLE_BROKEN_DOWN));
             if (mechanic == nullptr
                 || (mechanic->State != PeepState::HeadingToInspection && mechanic->State != PeepState::Answering)
                 || mechanic->CurrentRide != ride->id)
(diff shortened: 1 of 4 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -2635,6 +2636,11 @@ static void ride_mechanic_status_update(Ride* ride, int32_t mechanicStatus)
                 ride->window_invalidate_flags |= RIDE_INVALIDATE_RIDE_MAINTENANCE;
                 ride_mechanic_status_update(ride, RIDE_MECHANIC_STATUS_CALLING);
             }
+            // if the ride is broken down, but a mechanic was heading for an inspection, update orders to fix
+            else if (mechanic->State == PeepState::HeadingToInspection && !forInspection)
+            {
+                ride_call_mechanic(ride, mechanic, forInspection);
```
IntelOrca, line 2642:
```suggestion
                ride_call_mechanic(ride, mechanic, true);
```

adam-bloom: @IntelOrca should we also change the `forInspection` variable defined above to `forBreakdown`? I don't think changing this line on its own adds any additional clarity, as forInspection must be false to get into this conditional block.

IntelOrca: Rename the variable to whatever you think is clearest. But I thought it is clearer to pass a literal to the function since it will always be false. Ideally I would do `ride_call_mechanic(ride, mechanic, forInspection: false);`, but that is not possible in C++. That is the problem with boolean parameters, they don't tend to be as clear. An enum would be better or two distinct functions.

tupaschoal (commented):
Make sure to rebase to solve merge conflicts!

src/openrct2/ride/Ride.cpp
adam-bloom: @IntelOrca updated and added a few comments - hopefully it's clearer now

duncanspumpkin:
Has this been tested? All looks good to me.

adam-bloom:
> Has this been tested? All looks good to me.

@duncanspumpkin I did manual integration testing to get the screenshots included. I’ve played at least an hour or two on a build including this and have yet to see anything odd. Anything else you were hoping to see tested?

duncanspumpkin:
Increment the network version (NetworkBase.cpp) and this is good to go as far as i see.

duncanspumpkin:
And don't forget to rebase before that as we recently incremented the version.

duncanspumpkin (approved):

PR #13492, opened 2020-11-30
Fix #13476: Clean up unneeded colour variables in TopToolbar.cpp

### LINKED ISSUE #13476: Clean up return values of sub_6E1F34() and try_place_ghost_scenery()
Currently, these functions sometimes return primary/secondary colour, but these are just gSceneryPrimaryColour. There is little point in returning these from a function that does not even concern itself with colours.

ffrogman:
I also removed the _secondaryColour and _tertiaryColour variables. I don't think we had a need for them they seemed to just match the global values.

### DIFF
--- a/src/openrct2-ui/windows/TopToolbar.cpp
+++ b/src/openrct2-ui/windows/TopToolbar.cpp
@@ -314,10 +314,6 @@ static ClearAction GetClearAction();
 static bool _landToolBlocked;
 static uint8_t _unkF64F0E;
 static int16_t _unkF64F0A;
-// rct2: 0x00F64F15
-static colour_t _secondaryColour;
-// rct2: 0x00F64F16
-static colour_t _tertiaryColour;
 
 /**
  * Creates the main game top toolbar window.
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
I'm not sure this should be changed. We want to be moving things away from the global variables this is just further enforcing them

Gymnasiast:
Can you change the `try_place_` functions to take primary/secondary/tertiary colour as parameter (of type `colour_t`)?

IntelOrca:
I agree with @duncanspumpkin. It is nice that it removes those static globals and indeed also out parameters. But further improvement would be reducing or removing the use of those other global variables and returning a struct or tuple instead of out paramters.

Gymnasiast:
@IntelOrca Certainly, but the original issue the OP addresses did not mention this. If the OP modifies `try_place_` functions to take colour input parameters again, I think that would be sufficient in actually _reducing_ usage of globals.

ffrogman:
@Gymnasiast, ok I can modify the `try_place_` functions to take the appropriate color parameters. When I call the function should I still pass in the appropriate global color variables?

Gymnasiast:
Yeah, I think that is sufficient. Obviously it will have to be changed at _some_ point, but I prefer multiple smaller PRs that are easy to review over a big one.

ffrogman:
Ok the `try_place_` functions have colour parameters. 

Now that I look at it these gWindowsScenery* colour variables are only used in TopToolbar.cpp, Scenery.cpp, and Window.h. Probably not terribly difficult to clean this up. Although I am not sure what format this project prefers. Plus this may be out of scope for Issue #13476 as writte.

Gymnasiast:
> Now that I look at it these gWindowsScenery* colour variables are only used in TopToolbar.cpp, Scenery.cpp, and Window.h. Probably not terribly difficult to clean this up.

If you'd like to try your hand at removing them, please go ahead!

> Although I am not sure what format this project prefers.

Could you elaborate?

PR #13503, opened 2020-12-02
Create ObjectEntryDescriptor to properly handle mixed DAT and JSON

Gymnasiast:
ObjectEntryDescriptor contains an enum to tell whether the object is DAT or JSON, along with either a DAT header or JSON identifier.

In this PR, it is used for the list of objects in a scenery group. It is also useful for the new save format, allowing it to reference both old and new objects and allowing for a smooth transition from .DAT objects to JSON objects.

### DIFF
--- a/src/openrct2/core/IStream.cpp
+++ b/src/openrct2/core/IStream.cpp
@@ -9,6 +9,7 @@
 
 #include "IStream.hpp"
 
+#include "../object/Object.h"
 #include "Memory.hpp"
 #include "String.hpp"
 
@@ -59,6 +60,24 @@ namespace OpenRCT2
     void IStream::WriteString(const std::string& str)
     {
         WriteString(str.c_str());
+    }
+
+    ObjectEntryDescriptor IStream::ReadObjectEntryDescriptor()
+    {
+        auto generation = ReadValue<ObjectGeneration>();
+        if (generation == ObjectGeneration::DAT)
+            return ObjectEntryDescriptor(ReadValue<rct_object_entry>());
+
+        return ObjectEntryDescriptor(ReadStdString());
+    }
+
+    void IStream::WriteObjectEntryDescriptor(const ObjectEntryDescriptor& oed)
+    {
+        WriteValue<ObjectGeneration>(oed.Generation);
+        if (oed.Generation == ObjectGeneration::DAT)
+            WriteValue<rct_object_entry>(oed.Entry);
+        else
+            WriteString(oed.Identifier);
     }
 
 } // namespace OpenRCT2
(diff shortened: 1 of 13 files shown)


IntelOrca (approved):

src/openrct2/object/Object.h
```diff
@@ -126,6 +128,46 @@ struct rct_object_entry_group
...
+        Entry = newEntry;
+    }
+
+    explicit ObjectEntryDescriptor(std::string_view newIdentifier)
+    {
+        Generation = ObjectGeneration::JSON;
+        safe_strcpy(const_cast<char*>(Identifier), std::string(newIdentifier).c_str(), 64);
+    }
+
+    ObjectEntryDescriptor& operator=(const ObjectEntryDescriptor& newEntry)
```
duncanspumpkin, line 159: https://en.cppreference.com/w/cpp/language/rule_of_three

src/openrct2/core/IStream.cpp
```diff
@@ -61,4 +62,22 @@ namespace OpenRCT2
         WriteString(str.c_str());
     }
 
+    ObjectEntryDescriptor IStream::ReadObjectEntryDescriptor()
```
duncanspumpkin, line 65: This is not how any other custom structure reads and writes to the IStream. Its polluting the IStream with details that are relevant to a very small subset of the program.

src/openrct2/object/Object.h
```diff
@@ -126,6 +128,46 @@ struct rct_object_entry_group
...
+        Generation = ObjectGeneration::DAT;
+        Entry = newEntry;
+    }
+
+    explicit ObjectEntryDescriptor(std::string_view newIdentifier)
+    {
+        Generation = ObjectGeneration::JSON;
+        safe_strcpy(const_cast<char*>(Identifier), std::string(newIdentifier).c_str(), 64);
+    }
+
```
duncanspumpkin, line 158: Add a GetIdentifier function as its used multiple times.

Gymnasiast: I don't think I understand. This function is setting stuff, not getting.

duncanspumpkin: There are a two locations (so far) where you have to get either the identifier or the dat name. Just make a function called GetIdentifier and use that for those two locations.

Gymnasiast: No, it is either the JSON identifier (a string) or a DAT _header_ (which is a struct and not a string). And even if I could create such a function, there is still the issue that I have to _set_ the property based on whether we're handling a DAT or JSON object.

duncanspumpkin: You have two instances where you branch to get the identifier or the dat name. FindObject and window_editor_object_selection_select_object use a getter to get the identifier name.

Gymnasiast: Right, that makes sense. I was confused because you made this a line note.

PR #13505, opened 2020-12-03
Start implementing undo functionality

duncanspumpkin:
just a proof of concept for now
This enables undoing of actions. It still has a lot of thought required to be done.


src/openrct2/actions/GameAction.cpp
```diff
@@ -445,6 +457,14 @@ namespace GameActions
             ActionLogContext_t logContext;
             LogActionBegin(logContext, action);
 
+            if (!(actionFlags & GameActions::Flags::ClientOnly))
+            {
+                auto undo = action->Get_Undo(*result);
+                if (undo != nullptr)
+                {
+                    _undoQueue.emplace_back(std::move(undo));
```
duncanspumpkin, line 465: perhaps move the placing in a queue until after completion of the Execute command.

src/openrct2/actions/GameAction.cpp
```diff
@@ -272,6 +273,17 @@ namespace GameActions
...
 
+    void Undo()
+    {
+        if (_undoQueue.empty())
+        {
+            return;
+        }
+        auto action = std::move(_undoQueue.back());
+        _undoQueue.pop_back();
+        Execute(action.get());
```
duncanspumpkin, line 284: Add parameter to indicate that is from the undo queue so that we don't store undo actions.

IntelOrca:
I don't think we need to implement undo for all actions. But some actions like SetParkName would need to store the previous park name - I suppose it could return SetParkName(oldParkName) from the Get_Undo function.

More importantly, removing a scenery item is not quite the same as undo as you will not get the same cash back and it will be recorded in your finances. Undo would have to revert everything the Execute action did.

duncanspumpkin:
Yeah I need to add an undo flag so that I can make it free. No we don't need to do every action but even then its still a lot of things needed. As for the set names thats why I make it generate the undo function before the execute so that it can get the current state. Might need to store side effect state (i.e. nested actions) in the game action result so that it can also be packaged into the undo function

PR #13506, opened 2020-12-04
ObjectEntryDescriptor Serializer Trait

duncanspumpkin:
Ended up being quite a mess. Not fully tested.

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -14,6 +14,7 @@
 #include "../localisation/Localisation.h"
 #include "../network/NetworkTypes.h"
 #include "../network/network.h"
+#include "../object/Object.h"
 #include "../ride/Ride.h"
 #include "../ride/TrackDesign.h"
 #include "../world/Location.hpp"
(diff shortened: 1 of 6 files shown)


src/openrct2/object/Object.h
```diff
@@ -139,11 +139,23 @@ struct ObjectEntryDescriptor
     ObjectGeneration Generation;
     union
     {
+        std::string Identifier; // For JSON objects
```
duncanspumpkin, line 142: Might just break this union into two fields. The Union is causing a good few issues.

src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -730,3 +731,46 @@ template<> struct DataSerializerTraits_t<rct_vehicle_colour>
...
+            std::string id;
+            DataSerializerTraits<std::string> s;
+            s.decode(stream, id);
+            val = ObjectEntryDescriptor(id);
+        }
+    }
+    static void log(OpenRCT2::IStream* stream, const ObjectEntryDescriptor& val)
+    {
+        char msg[128] = {};
+        snprintf(msg, sizeof(msg), "ObjectEntryDescription (Generation = %d)", val.Generation);
```
Gymnasiast, line 773:
```suggestion
        snprintf(msg, sizeof(msg), "ObjectEntryDescriptor (Generation = %d)", val.Generation);
```

src/openrct2/object/Object.h
```diff
@@ -137,60 +152,30 @@ enum class ObjectGeneration : uint8_t
...
+        , Entry()
     {
-        *this = entry;
     }
 
     explicit ObjectEntryDescriptor(const rct_object_entry& newEntry)
     {
         Generation = ObjectGeneration::DAT;
         Entry = newEntry;
+        Identifier = std::string(newEntry.GetName());
```
Gymnasiast, line 169: Not entirely sure about this. I'd prefer having `Identifier` empty if there isn't one. (Also see the conversation in #plug-in with Bassie.)

src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -730,3 +731,46 @@ template<> struct DataSerializerTraits_t<rct_vehicle_colour>
...
+            std::string id;
+            DataSerializerTraits<std::string> s;
+            s.decode(stream, id);
+            val = ObjectEntryDescriptor(id);
+        }
+    }
+    static void log(OpenRCT2::IStream* stream, const ObjectEntryDescriptor& val)
+    {
+        char msg[128] = {};
+        snprintf(msg, sizeof(msg), "ObjectEntryDescriptor (Generation = %d)", static_cast<int32_t>(val.Generation));
```
Gymnasiast, line 773: This won't log the identifier/DAT name, which is the most interesting bit.

This _might_ be fine, I don't know how important this logging function is.

Gymnasiast (approved):
LGTM. I have one line note, but I'll let you decide if that needs addressing or not.

src/openrct2/core/DataSerialiserTraits.h
duncanspumpkin: Loging function is only used in network code as far as i am aware so you can't actually trigger this.

PR #13523, opened 2020-12-05
Fix #13517: Bad formatting of dates

### LINKED ISSUE #13517: Bad formatting of chat messages
**OS:** Windows 10
**Version:** [e.g. 0.3.2] 0.3.2
**Build:** e8611169a5eed155c90f1cfd42231e2b65ad801f

Sorry that it is not english in the screenshot below,
All chat messages are broken, since the new-format-string branch is merged, I think.

Also, It breaks scenario objective's date in Korean.
Please aware me if there is something I should change in Korean translation.

**Screenshots / Video:**
image

Hydro Hills, the date is broken. The right one is October, 3 but it is displayed as June, 7.
image

IntelOrca:


src/openrct2/localisation/Formatting.cpp
```diff
@@ -792,6 +794,22 @@ namespace OpenRCT2
...
+    {
+        thread_local std::vector<FormatArg_t> tempArgs;
+        tempArgs.clear();
+
+        auto fmt = GetFmtStringById(STR_DATE_FORMAT_MY);
+        Formatter ft;
+        ft.Add<uint16_t>(month);
+        ft.Add<uint16_t>(year);
+        const void* legacyArgs = ft.Data();
+        BuildAnyArgListFromLegacyArgBuffer(fmt, tempArgs, legacyArgs);
```
tupaschoal, line 808: Do we need to use the legacy version?

IntelOrca: Do you see another way?

tupaschoal: Idk, open ended question, I'm assuming that if there's a legacy function, there is also a non legacy one

IntelOrca: There is and it causes said bug I am trying to fix. The reason is the non-legacy version doesn't support PUSH16 POP16, but it has no concept of argument byte size.

tupaschoal (approved):

PR #13532, opened 2020-12-06
Get Testpaint operating again

duncanspumpkin:
This gets Testpaint working on x64 by ripping out all of the interop code. The plan is that the output from this will be used to create a baseline which will be compared against. This is in contrast to the original test paint which used the original game to compare functions with.


test/testpaint/TestPaint.cpp
```diff
@@ -51,46 +51,46 @@ namespace TestPaint
             vehicleEntry.base_image_id = 0x70000;
             rideEntry.vehicles[0] = vehicleEntry;
             gRideList[0] = {};
-            gRideEntries[0] = &rideEntry;
+            //gRideEntries[0] = &rideEntry;
```
duncanspumpkin, line 54: this is required

Gymnasiast:
?

duncanspumpkin:
Argh, i need to name my branches better. Thought this was one that was already merged.

Issue #13384, opened 2020-11-06
Expose all TileElement data to plugin API


Sadret:
As of now, every `Tile` has an int8 data array that stores the data of all its `TileElement`s.
A lot of this data is already also accessible through fields in the respective `XyElement extends BaseTileElement` interfaces.

This issue demands to make all of the data available.

In some cases, parts of the data is used in different ways, depending on the situation.
E.g. for footpaths, `AdditionStatus `and `rideIndex` share the same memory.

In a lot of other cases, one int8 is split up bitwise for different uses.
E.g. `flags` consists of 4 bit for occupied quadrants, 1 bit for the ghost flag, 2 unused bits, and 1 bit for the last tile flag.

A complete list can be found [here (Google Spreadsheets)](https://docs.google.com/spreadsheets/d/1UqbZRRDDazb0tiR5BqVx5_CBLKYgRJ6Q0obEVId-zCk).

Now, I think there are a few general points to discuss here:

1. Should we expose every single of these fields?
2. For flags etc, should we add a field for every flag or one field with the combined value of all flags?
3. If the getter and setters transform the value before returning/setting, should we use the transformed value or the raw one?
E.g. `TileElementBase` stores the `base_height` as `value / COORDS_Z_STEP`. The current API uses the raw value in this case.
4. Should we include other fields, that do not corresponds to a getter/setter of some TileElement data?
E.g. `isHidden()` just does calculations that are also possible within the plugins.
`SurfaceElement` has readonly fields for `hasOwnership `and `hasConstructionRights` although there are no getters for that.

Sadret:
This is currently my opinion:

1. Yes. For every field, sooner or later ther is a plugin developer who needs it.
2. One field for every flag. It is much more independent from the implementation, which might change with the new save file format. (Although this leads to roughly ten more fields in TrackElement)
3. Use the transformed values. As above, it is much more independent from the implementation.
4. No, I would keep this interface "pure". For the ownership flags, we could split them up like in no 2 and implement getters and setters.

duncanspumpkin:
I would prefer that there is a getter and setter for each of the fields so it reflects how they are used in the rest of the codebase. We wont be changing the size of `TileElement` it has already been resized for future use.

IntelOrca:
`isHidden` is planned to be a real flag once the new save format is available. It helps ensure plugins still work correctly after the transition.

Sadret:
I'd like to suggest the following changes:

#### Add
[...]

PR #13536, opened 2020-12-06
Implement #13384: Expose all TileElement data to plugin API

Sadret:
#### Adds the following fields:
TileElement.isGhost
TileElement.baseZ (= baseHeight * 8)
TileElement.clearanceZ (= clearanceHeight * 8)
Footpath.edges
Footpath.corners
Track.mazeEntry
Track.colourScheme
Track.seatRotation
Track.brakeBoosterSpeed ( = stored value * 2)
Track.isInverted
Track.hasCableLift
Wall.slope
Wall.primaryColour
Wall.secondaryColour
Wall.tertiaryColour
Wall.bannerIndex
LargeScenery.bannerIndex
LargeScenery.sequence
Banner.bannerIndex
Banner.direction

#### Removes the following fields:
Footpath.edgesAndCorners (replaced by Footpath.edges and Footpath.corners)
Footpath.railings (unused)
Footpath.direction (unused)

#### Changes to openrct2.d.ts
- adds/removes above fields
- removes FootpathElement->footpathType (unused / mistake?)
- adds nullable types where needed
- removes type TileElement (wrongly used and superfluous / mistake?)

#### Other changes:
- API returns null values for all fields that do not make sense for that element, e.g. corners for Walls.
- Banner.position is merged into TileElement.direction, as they have the same semantic meaning. (Note that above this change only affect the plugin API and nothing else.)
- Increased API version.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -540,8 +540,11 @@ declare global {
     interface BaseTileElement {
         type: TileElementType;
         baseHeight: number;
+        baseZ: number;
         clearanceHeight: number;
+        clearanceZ: number;
         occupiedQuadrants: number;
+        isGhost: boolean;
         isHidden: boolean; /** Take caution when changing this field, it may invalidate TileElements you have stored in your script. */
     }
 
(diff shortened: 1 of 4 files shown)


src/openrct2/scripting/ScTile.hpp
```diff
@@ -112,6 +112,17 @@ namespace OpenRCT2::Scripting
             Invalidate();
         }
 
+        uint8_t baseZ_get() const
+        {
+            return _element->GetBaseZ();
+        }
+        void baseZ_set(uint8_t value)
```
Gymnasiast, line 119: Should be an uin16_t at least. Valid range is `0 - (254 * COORDS_Z_STEP)`. (Because 255 is used to mark the tile element as discarded.

Sadret: Same for clearanceZ.
Changed it to uint16_t.

src/openrct2/scripting/ScTile.hpp
```diff
@@ -719,37 +910,113 @@ namespace OpenRCT2::Scripting
...
         }
 
-        bool railings_get() const
+        uint8_t edges_get() const
         {
             auto el = _element->AsPath();
-            return el != nullptr ? el->GetRailingEntryIndex() : false;
+            return el != nullptr ? el->GetEdges() : 0;
         }
-        void railings_set(bool value)
```
Gymnasiast, line 730: Why did you remove railings_get / railings_set ?

Sadret: It's unused. The getter is just an alias for `GetSurfaceEntryIndex` and the setter does nothing.

Gymnasiast: .. for now. But well, how path surfaces and railings work is bound to change a lot anyway.

Gymnasiast (approved):
LGTM, but would like either @IntelOrca or @duncanspumpkin to review it as well.

src/openrct2/scripting/ScTile.hpp
```diff
@@ -446,6 +501,114 @@ namespace OpenRCT2::Scripting
             }
         }
 
+        uint16_t mazeEntry_get() const
+        {
+            auto el = _element->AsTrack();
+            if (el != nullptr)
```
duncanspumpkin, line 507: Should we have an extra check to make sure this is a maze entry?

src/openrct2/scripting/ScTile.hpp
```diff
@@ -446,6 +501,114 @@ namespace OpenRCT2::Scripting
...
+                return el->GetMazeEntry();
+            return 0;
+        }
+        void mazeEntry_set(uint16_t value)
+        {
+            ThrowIfGameStateNotMutable();
+            auto el = _element->AsTrack();
+            if (el != nullptr)
```
duncanspumpkin, line 515: As above

src/openrct2/scripting/ScTile.hpp
Sadret: Probably. Consequently, we should then also check in `sequence_get` that it is _not_ a maze. Also:
`Footpath->AdditionStatus` vs `Footpath->rideIndex`
`Track->ColourScheme` (i.e. `colourScheme` and `seatRotation`) vs `Track->MazeEntry`
`Track->BrakeBoosterSpeed` vs it's not a break/booster.
That's all unions used.

duncanspumpkin: @IntelOrca what are your opinions

IntelOrca: I would say so.

distribution/openrct2.d.ts
```diff
@@ -531,66 +534,88 @@ declare global {
...
+    interface WallElement extends BaseTileElement {
+        direction: Direction;
         object: number;
-        sequence: number;
-        ride: number;
-        station: number;
+        primaryColour: number;
+        secondaryColour: number;
+        tertiaryColour: number;
+        bannerIndex: number;
```
duncanspumpkin, line 589: bannerIndex could be null

distribution/openrct2.d.ts
```diff
@@ -531,66 +534,88 @@ declare global {
...
+        sequence: number;
+        footpathObject: number;
     }
 
     interface LargeSceneryElement extends BaseTileElement {
+        direction: Direction;
         object: number;
         primaryColour: number;
         secondaryColour: number;
+        bannerIndex: number;
```
duncanspumpkin, line 607: bannerIndex could be null

distribution/openrct2.d.ts
Sadret: The way it is implemented right now, bannerIndex is never null. But bannerIndex == 0 is ambigious, as it can either mean no banner or the first banner.
I can change that by e.g. checking if GetBanner() == nullptr first.

distribution/openrct2.d.ts
Sadret: see above

duncanspumpkin (approved):
Don't forget to increment the api version

Sadret:
> Don't forget to increment the api version

I did, twice. Now a third time, I guess.

Sadret:
Changed
`tile_element_get_banner_index(_element)`
to
`_element->GetBannerIndex();`
after the latest rebase.

duncanspumpkin:
@Sadret for future what I do is wait until all of the reviews are in and no further changes are being asked and then bump the network/plugin version. Means you don't have to do it as many times. Doesn't always work out like that though :(.

PR #13548, opened 2020-12-08
Split actions hpp files into separate h and cpp files

adam-bloom:
This PR splits all actions *.hpp files into separate .h and .cpp files.
Various include paths are adjusted in other game files that used to (a bit sloppily) include headers via the actions hpp files that the action headers do not require.
Required for #13505.


James103:
@adam-bloom Could you try this [tutorial on how to rebase your fork](https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2)?

duncanspumpkin:
Good work. Completely unscientific but build time of Windows x64 was reduced by 2 minutes after this change.

tupaschoal (commented):
This is great, thanks! I love that you tried to split it into multiple commits. I'd suggest folding some of those like the clang-format ones and etc and only keeping commits that actually split files, but I understand it might be too much.

I've just looked at the first commit for now, but I'd say that if we are doing this, we should have as much as possible on the cpp, even the constructors and some simple getters. This can be done later if we think it's too much for this PR. Also seems like in most of these actions we have room for using delegating constructors.

duncanspumpkin:
@tupaschoal it was based off of my undo branch following the convention I had done.

adam-bloom:
> This is great, thanks! I love that you tried to split it into multiple commits. I'd suggest folding some of those like the clang-format ones and etc and only keeping commits that actually split files, but I understand it might be too much.
A lot of my recent work has been on a project that squashes PRs on merge, and I couldn't remember what the convention was in this repo. Easy enough to clean up a bit. I agree this is a few too many commits to go forward.

> I've just looked at the first commit for now, but I'd say that if we are doing this, we should have as much as possible on the cpp, even the constructors and some simple getters. This can be done later if we think it's too much for this PR. Also seems like in most of these actions we have room for using delegating constructors.
As @duncanspumpkin said, he had already started some changes. He picked a convention very similar to what my first attempt likely would have been, so I just went with it. I'm ok either way - merging the current convention and opening issues for future clean up, or if we think it is best to just tackle now and be done with it.

tupaschoal:
We don't have a convention, I'd say. If the commits are distinct enough and well contained, I'd rather merge than squash. Sometimes the whole PR is just doing a single change and I just squash for ease

adam-bloom:
@tupaschoal I just cleaned up the commits a bit, I think it'll be a bit easier to review now (and, they mostly have consistent naming).

Let me know if you think it's worth pursuing changes to the h/cpp style now, or if we want to put that on hold. cc @duncanspumpkin

duncanspumpkin:
I think its fine as is for now. Could you add new lines to the end of all the files that don't have them (they show as a red stop sign in github)

janisozaur:
Looked at this and seems quite straightforward. Was there ever really a reason to keep the GAs in include files?

tupaschoal (approved):
I took a glance at it and it seems ready to go. We can tackle other stuff like moving even more methods into the source file later. I find this a very welcome PR

duncanspumpkin:
@janisozaur it was for a few reasons none of them that sensible.

PR #13558, opened 2020-12-08
Fix #6677: Add discord RPC to macOS builds

### LINKED ISSUE #6677: Discord Rich Presence is missing from macOS builds
**OS:** macOS
**Commit/Build:** d649a3e8b

Support for Discord Rich Presence API was introduced in:
* https://github.com/OpenRCT2/OpenRCT2/pull/6651 (Windows)
* https://github.com/OpenRCT2/OpenRCT2/pull/6658 (Linux)

It should also be added to our macOS builds.

adam-bloom:
This PR adds discord-rpc to the Xcode build scripts. Unlike cmake, Xcode does not support flags in quite the same way (or at least, I don't know how to use them!), so discord-rpc is always cloned, built, and linked.

General question: why do we explicitly have the Xcode project file maintained, instead of generating it from cmake to bring the build in line with linux? See https://cmake.org/cmake/help/v3.9/generator/Xcode.html. Is there a good reason, or just that no one has tried to get it to work? I'll note that once I cloned the discord-rpc directory, the cmake build on macOS immediately worked.

### DIFF
--- a/OpenRCT2.xcodeproj/project.pbxproj
+++ b/OpenRCT2.xcodeproj/project.pbxproj
@@ -67,6 +67,9 @@
 		4CF67197206B7E720034ADDD /* object in Resources */ = {isa = PBXBuildFile; fileRef = 4CF67196206B7E720034ADDD /* object */; };
 		6341F4E12400AA0F0052902B /* Drawing.Sprite.RLE.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6341F4DF2400AA0E0052902B /* Drawing.Sprite.RLE.cpp */; };
 		6341F4E22400AA0F0052902B /* Drawing.Sprite.BMP.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6341F4E02400AA0F0052902B /* Drawing.Sprite.BMP.cpp */; };
+		662578A625803AA90002C77E /* discord_rpc.h in Headers */ = {isa = PBXBuildFile; fileRef = 662578A525803AA90002C77E /* discord_rpc.h */; };
+		662578AD25803CE50002C77E /* libdiscord-rpc.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 662578A325803A6C0002C77E /* libdiscord-rpc.a */; };
+		662578AE25803D040002C77E /* libdiscord-rpc.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 662578A325803A6C0002C77E /* libdiscord-rpc.a */; settings = {ATTRIBUTES = (Required, ); }; };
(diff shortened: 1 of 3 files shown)


Gymnasiast (approved):
Could you update the changelog?

`- Feature: [#6677] Add Discord RPC to macOS builds.`

Gymnasiast:
> General question: why do we explicitly have the Xcode project file maintained, instead of generating it from cmake to bring the build in line with linux? See https://cmake.org/cmake/help/v3.9/generator/Xcode.html. Is there a good reason, or just that no one has tried to get it to work? I'll note that once I cloned the discord-rpc directory, the cmake build on macOS immediately worked.

It was once attempted, but stranded. https://github.com/OpenRCT2/OpenRCT2/pull/8650

It would be wonderful if you could pull that off, seeing as it's currently a royal pain in the behind to update the Xcode project for every single file that gets added, renamed or moved. (And heaven forbid if you create a new directory.)

LRFLEW:
Ok, hold on.

a) Why is this using a static library? All the other dependencies are using dynamic libraries.
b) This is building discord-rpc for each build locally? Why wasn't this added to OpenRCT2/Dependencies?

I'm currently working on making https://github.com/OpenRCT2/Dependencies/pull/25 stable. I can add discord-rpc into that as well if this is something that's desired. I *did* say in that PR's description "let me know if there's anything else [such as discord-rpc] you want me to look into for this," but nobody responded.

adam-bloom:
@LRFLEW To answer your question, I was using the linux implementation as a model, which both builds locally and links statically. In fact, if you clone the discord-rpc repo (either manually or using the shell script I modified to work on macOS), building with cmake on macOS will also do that.

I spent a lot more time with Xcode builds today (I'm off work this week, so just exploring a bit!), and I now see how different that build is than the linux builds.

LRFLEW:
@adam-bloom That makes sense. Generally, the CMake and Xcode builds are different by how they handle dependencies. The CMake build uses system-installed dependencies (usually through Homebrew), while the Xcode build creates a self-contained app bundle that bundles its own dependencies. This is probably fine for now, but we should probably update it to use the dependency scheme that the other libraries are using.

And I was the person who made the original Xcode project years ago. At the time, I didn't know CMake very well, and knew that Xcode made making these sorts of app bundles simple. As I said before, I don't mind replacing the Xcode project with a CMake system, but *ideally* both build types (self-contained and system-install libraries) should probably still be possible. At this point, it's just a matter of someone who understands CMake making the required changes.

adam-bloom:
@LRFLEW totally agreed. I got started on the cmake changes today (they actually work for generating the Xcode project file too). It does system-installed libraries only right now....we need the next cmake release to embed libraries in app bundles via Xcode projects (or we can go for full replacement with cmake). Either way, let me open up a draft PR for that change so we can continue that conversation there....

PR #13560, opened 2020-12-09
Close #12993: Remove hardcoded {SMALLFONT}{BLACK} from all tooltips

### LINKED ISSUE #12993: Remove hardcoded {SMALLFONT}{BLACK} from all tooltips

Lukasz-Pe:
2020-12-09 11:46: Colour picker has to be reworked. Tooltip is not small and black.
2020-12-09 11:54: Done.

### DIFF
--- a/src/openrct2-ui/windows/Tooltip.cpp
+++ b/src/openrct2-ui/windows/Tooltip.cpp
@@ -71,9 +71,17 @@ void window_tooltip_show(const OpenRCT2String& message, ScreenCoordsXY screenCoo
         return;
 
     OpenRCT2String formattedMessage;
-    formattedMessage.str = STR_STRING_DEFINED_TOOLTIP;
-    formattedMessage.args = Formatter();
-    formattedMessage.args.Add<rct_string_id>(message.str);
+    if (message.args.NumBytes() == 0)
+    {
+        formattedMessage.str = STR_STRING_DEFINED_TOOLTIP;
+        formattedMessage.args = message.args;
+        formattedMessage.args.Add<rct_string_id>(message.str);
+    }
+    else
+    {
+        formattedMessage.str = message.str;
+        formattedMessage.args = message.args;
+    }
     int32_t textWidth = FormatTextForTooltip(formattedMessage);
     int32_t width = textWidth + 3;
     int32_t height = ((_tooltipNumLines + 1) * font_get_line_height(FONT_SPRITE_BASE_MEDIUM)) + 4;
--- a/src/openrct2-ui/windows/Tooltip.cpp
+++ b/src/openrct2-ui/windows/Tooltip.cpp
@@ -71,17 +71,9 @@ void window_tooltip_show(const OpenRCT2String& message, ScreenCoordsXY screenCoo
         return;
 
     OpenRCT2String formattedMessage;
-    if (message.args.NumBytes() == 0)
-    {
-        formattedMessage.str = STR_STRING_DEFINED_TOOLTIP;
-        formattedMessage.args = message.args;
-        formattedMessage.args.Add<rct_string_id>(message.str);
-    }
-    else
-    {
-        formattedMessage.str = message.str;
-        formattedMessage.args = message.args;
-    }
+    formattedMessage.str = STR_STRING_DEFINED_TOOLTIP;
+    formattedMessage.args = message.args;
+    formattedMessage.args.Add<rct_string_id>(message.str);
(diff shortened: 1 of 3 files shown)


Lukasz-Pe:
If you want I can move those changes to separate function.

src/openrct2-ui/windows/Tooltip.cpp
```diff
@@ -70,7 +70,11 @@ void window_tooltip_show(const OpenRCT2String& message, ScreenCoordsXY screenCoo
     if (w != nullptr)
         return;
 
-    int32_t textWidth = FormatTextForTooltip(message);
+    OpenRCT2String formattedMessage;
+    formattedMessage.str = STR_STRING_DEFINED_TOOLTIP;
+    formattedMessage.args = Formatter();
+    formattedMessage.args.Add<rct_string_id>(message.str);
```
IntelOrca, line 76: I don't know if any tooltips use arguments, but you are losing whatever was in message.args which might be an issue.

Lukasz-Pe: I've added an if statement. I am not sure if used NumBytes() funtion was intended for such tasks.

IntelOrca: Now tooltips with arguments don't have the right format: BLACK SMALL. You need to copy the arguments across to the formatted one.

Lukasz-Pe: If I do the following (basically if statement for message without arguments):
```
formattedMessage.str = STR_STRING_DEFINED_TOOLTIP;
formattedMessage.args = message.args;
formattedMessage.args.Add<rct_string_id>(message.str);
```
then the arguments are copied from message and string should be formatted as BLACK SMALL.
I would like to understand working of `formattedMessage.args.Add<rct_string_id>(message.str);` better. Specifically if this funcion (`.Add<>()`) concatenates given value of `.args` with anything given in braces of `.Add<>()` or simply overwrites `.args`?
I am asking because there is `std::memcpy` used in template of this function and as far as I understand the target buffer will be overwritten.

IntelOrca: I think it would be better to do something like:
```
format_string(buffer1, sizeof(buffer1), message.str, message.args.Data());
FormatString(buffer1, sizeof(buffer1), "{BLACK}{SMALL}{STRING}", buffer1);
```

Have a look at what happens in `FormatTextForTooltip`

Lukasz-Pe: I've looked there, but needed your suggestion to have that click in my head :)

src/openrct2-ui/windows/Tooltip.cpp
```diff
@@ -53,6 +53,10 @@ static int32_t FormatTextForTooltip(const OpenRCT2String& message)
     format_string(_tooltipText, sizeof(_tooltipText), message.str, message.args.Data());
     gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
 
+    OpenRCT2String formattedMessage{ STR_STRING_TOOLTIP, Formatter() };
+    formattedMessage.args.Add<const char*>(_tooltipText);
+    format_string(_tooltipText, sizeof(_tooltipText), formattedMessage.str, formattedMessage.args.Data());
```
IntelOrca, line 58: I am a bit nervous about this reading and writing to the same buffer. I think it is better to use a middle temp buffer for the first format call.

Gymnasiast:
You were a bit too eager with removing {SMALLFONT}{BLACK}, it seems:

image

It otherwise appears to work correctly. @IntelOrca are you happy with the implementation, or does that still need adjustment?

IntelOrca (approved):

tupaschoal:
Are you ok with merging then @Gymnasiast ?

Gymnasiast:
No, of course not. I pointed out a mistake and Lukasz-Pe hasn't yet come round to correct it.

Lukasz-Pe:
@Gymnasiast sorry for not answering to your comment. Yes I've been too eager with deleting {SMALLFONT}{BLACK}:)
I've restored all I could find damaged (graphs).

Gymnasiast:
> sorry for not answering to your comment.

Don't worry, I just assumed you were busy with other things.

PR #13569, opened 2020-12-10
[WIP] macOS cmake updates for OpenRCT2.app target

adam-bloom:
Continuing the work started in #8650 and discussion in #13558....

Currently, cmake on macOS will build an executable only (`./openrct2`) while the Xcode project is setup to build `OpenRCT2.app`. This updates the cmake project definition to build `OpenRCT2.app` on macOS instead of the CLI executable. Before this is merged, the app bundle should be fully contained (libraries and required objects/languages/etc).

Yet to do:
- [ ] Reconcile the differences in the dependency (title sequences, objects, etc) management in cmake and the current Xcode project (cmake just checks for directory existence, Xcode tracks specific versions)
- [ ] Move dependency download/unpack to build phase instead of install phase for macOS only (since they need to be bundled into the .app file....or figure out if cmake install phase can end up in the app bundle)
- [ ] Fix library links to use rpath
- [ ] Update CI to use generated Xcode project? Or update to use make directly? Or both!
- [ ] Add tests to the macOS target. If the tests will run on macOS, then they should be used in CI somewhere.

Already done:
- [x] Xcode project generation
- [x] App icon (set properly both in GNU make and Xcode)
- [x] Copying (but not linking) of dependent libraries
- [x] g2.dat generation
- [x] Copying of objects/lanaguages/title sequences (but they currently must already exist in ./data)
- [x] changed minimum macOS target to 10.14 (cmake can only support one version for all targets, so the mix of 10.13 and 10.14 that the current Xcode project has needs to be rounded up.)

### Build methods
Two build methods seem to work (but I haven't tested this in a clean workspace yet):
make:
```
cmake ..
make
```
This will create OpenRCT2.app in the directory the command is run from.

Xcode:
```
cmake -G "Xcode" ..
xcodebuild
```
This generates an Xcode project file from cmake, which can then be used to build. It can also be used for development/editing as well...just run that cmake step when checking out the repo and then open the generated Xcode project. I'll note that the auto-generation does not preserve the directory structure that is set up in the current project, but sources for each target are clearly visible and can be edited. When building, the default configuration is "Debug", so the target is available in Debug/OpenRCT2.app (of course, Release could be specified via the appropriate argument).

### Library linking
Existing Xcode project:
```
❯ otool -L Release/OpenRCT2.app/Contents/MacOS/OpenRCT2
Release/OpenRCT2.app/Contents/MacOS/OpenRCT2:
	/System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa (compatibility version 1.0.0, current version 23.0.0)
	/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation (compatibility version 300.0.0, current version 1770.106.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/usr/lib/libcurl.4.dylib (compatibility version 7.0.0, current version 9.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	@rpath/libcrypto.dylib (compatibility version 1.1.0, current version 1.1.0)
	@rpath/libduktape.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libicudata.dylib (compatibility version 64.0.0, current version 64.2.0)
	@rpath/libicuuc.dylib (compatibility version 64.0.0, current version 64.2.0)
	@rpath/libpng16.dylib (compatibility version 54.0.0, current version 54.0.0)
	@rpath/libSDL2.dylib (compatibility version 13.0.0, current version 13.0.0)
	@rpath/libfreetype.dylib (compatibility version 24.0.0, current version 24.1.0)
	@rpath/libspeexdsp.dylib (compatibility version 7.0.0, current version 7.1.0)
	@rpath/libzip.dylib (compatibility version 5.0.0, current version 5.1.0)
        ...
```
The libraries from [OpenRCT2/Dependencies](https://github.com/OpenRCT2/Dependencies) are all linked with rpath. By default, the Frameworks folder inside the app bundle is part of the search path (which already has the dylibs copied), as is the standard library directories.

Cmake defaults to linking against the library versions it discovers. Here is a similar excerpt from the current build:
```
otool -L OpenRCT2.app/Contents/MacOS/OpenRCT2
OpenRCT2.app/Contents/MacOS/OpenRCT2:
	@rpath/libSDL2.dylib (compatibility version 13.0.0, current version 13.0.0)
	/usr/local/opt/speexdsp/lib/libspeexdsp.1.dylib (compatibility version 7.0.0, current version 7.1.0)
	/System/Library/Frameworks/OpenGL.framework/Versions/A/OpenGL (compatibility version 1.0.0, current version 1.0.0)
	/usr/local/opt/openssl@1.1/lib/libssl.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/local/opt/openssl@1.1/lib/libcrypto.1.1.dylib (compatibility version 1.1.0, current version 1.1.0)
	/usr/lib/libcurl.4.dylib (compatibility version 7.0.0, current version 9.0.0)
	/usr/local/opt/duktape/lib/libduktape.206.so (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/opt/libpng/lib/libpng16.16.dylib (compatibility version 54.0.0, current version 54.0.0)
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
       ...
```
Notice that only `libSDL2` was linked with rpath. This PR shows the changes required to accomplish that, although I do want to continue to look for a better way/explore alternatives.

@LRFLEW @Gymnasiast comments/discussion is very welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,12 @@ if (CCache_FOUND)
     endif (OPENRCT2_USE_CCACHE)
 endif (CCache_FOUND)
 
-set(CMAKE_OSX_DEPLOYMENT_TARGET "10.14")
+if (APPLE)
+    set(CMAKE_OSX_ARCHITECTURES "x86_64")
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.14")
+    include_directories(${CMAKE_SOURCE_DIR}/libxc/inc)
+    link_directories(${CMAKE_SOURCE_DIR}/libxc/lib)
+endif ()
 
 project(openrct2 CXX)
 
@@ -328,23 +333,6 @@ if (MACOS_BUNDLE)
     endif ()
 
     # TODO: make the above routine a function, use it for objects, title sequences, and languages
-
-    # not yet used, but link against the dylibs (and embed them)
-    message("search dir: ${MACOS_DYLIBS_DIR}")
-    find_library(
-        DYLIB_SDL2
-        libSDL2.dylib
-        PATHS ${MACOS_DYLIBS_DIR}/lib
-        HINTS ${MACOS_DYLIBS_DIR}/lib
-        NO_DEFAULT_PATH
-        NO_PACKAGE_ROOT_PATH
-        NO_CMAKE_PATH
-        NO_CMAKE_ENVIRONMENT_PATH
-        NO_CMAKE_SYSTEM_PATH
-        NO_SYSTEM_ENVIRONMENT_PATH
-        REQUIRED
-    )
-    message("Found libsdl2.dylib: ${DYLIB_SDL2}")
 
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,8 @@ if (MINGW)
 endif ()
 
 option(DISABLE_GUI "Don't build GUI. (Headless only.)")
-
+CMAKE_DEPENDENT_OPTION(MACOS_BUNDLE "Build macOS application bundle (OpenRCT2.app)" ON
+    "APPLE; NOT DISABLE_GUI" OFF)
 if (FORCE32)
     set(TARGET_M "-m32")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TARGET_M}")
@@ -297,7 +298,7 @@ if (MINGW)
 endif ()
 
 # if we'll build a macOS app bundle, we need to pull dependencies
-if (NOT DISABLE_GUI AND APPLE)
+if (MACOS_BUNDLE)
     # update dylibs
     set(MACOS_DYLIBS_VERSION "26")
     set(MACOS_DYLIBS_ZIPFILE "openrct2-libs-v26-x64-macos-dylibs.zip")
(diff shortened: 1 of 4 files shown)


LRFLEW:
Just a thought / note, it looks like you're tying the app bundle build type with the GUI build type. I think those should be separated out into separate variables. There are cases where non-bundled GUI builds are desirable, particularly when testing if there's an issue with the Dependency project's builds, and while a bundled non-GUI build seems weird, there is some utility in having such builds available simply to use the pre-built dependencies. I would make a separate variable like `MACOS_APP_BUNDLE` (or simply `MACOS_BUNDLE`) defaulting to ON for this IMO.

adam-bloom:
> I would make a separate variable like `MACOS_APP_BUNDLE` (or simply `MACOS_BUNDLE`) defaulting to ON for this IMO.

Should this variable being set to ON replace the CLI target or supplement it?

And thanks for your thoughts on if Xcode project generation was worthwhile! I hadn’t thought enough about it yet.

adam-bloom:
I've spent some more time thinking about this, and looking at the build system in general. There's a very good chance that I end up abandoning this PR (this branch has turned more into experimentation/playground and the commits aren't the cleanest) and moving to a new one. Before doing so, I wanted to ask a few questions, since I think we need to make a change to how to build macOS builds to accomplish all of the goals above. Current state is the following:

- https://github.com/OpenRCT2/Dependencies supplies dylibs for the macOS .app, but does not contain macOS static libraries
- https://github.com/OpenRCT2/openrct2-docker-build has the containers for CI, but the macOS CI runs on a base macos-latest image (note: base macOS image for GitHub actions has both Xcode and brew installed)
- test executables are statically linked
- openrct2 CLI executable is statically linked
- OpenRCT2.app is dynamically linked

We cannot have both a static and dynamically linked target built from only dynamic libraries, and using two different sources for the dependencies seems like a bad idea. As does dynamic linking for the CLI executables - that might be ok for CI, but poses issues for other users - we'd need to add an install action to copy the dylibs, where they risk trampling on brew installed dylibs if we put them in /usr/local/lib.

So, I'd like your thoughts on eliminating the macOS dependencies dylibs zip. In its place, we'd create an openrct2-build container for macOS that used brew to install all of the dependencies (or just always run `brew install` as the first step, this seems to be more common than not, although it doesn't allow dependency version pinning...if that's a concern). This will install both the static and dynamic libraries in the build system. We can continue to dynamically link, but we'd use the brew installed dylibs to build the .app bundle (we can run the steps performed [here](https://github.com/OpenRCT2/Dependencies/blob/master/macos_collect.sh#L40-L53) when copying the dylibs into the bundle). We could also use the static versions of the same libraries to build the test executables and include them in the CI pipeline.

@LRFLEW - what do you think? Is there something I didn't think about and should consider? We could also add the static libraries to the dependency zip file, but I'm really not sure why we have that instead of relying on brew to manage our dependencies on macOS, especially since GitHub actions runners install brew by default.

On the bright side, I've got everything running locally - .app building and running correctly (with rpath), tests build and run successfully, etc, but they're using a mix of dependencies from the zip file and homebrew installed packages, so I can't easily port this into the build pipeline.

cc @IntelOrca @Gymnasiast - please add others who may have thoughts on the macOS build pipeline. If we do want to make these changes, I'm happy to help, but it'd be great to have someone who can do CI container releases available, in case we need to introduce the macOS build container. If we decide that's the right path (I haven't found an example of doing this elsewhere though, only examples of running brew install/bundle as the first step in the CI pipeline), I'll need someone who can merge that release.

pizza2004:
I really only want to make sure, would a clean checkout of the repository still include an Xcode file and not require me to use cmake at all?  I hate using the command line and needing to use cmake would actively discourage me from ever contributing to the project again.

adam-bloom:
> I really only want to make sure, would a clean checkout of the repository still include an Xcode file and not require me to use cmake at all? I hate using the command line and needing to use cmake would actively discourage me from ever contributing to the project again.

@pizza2004, the idea of this is that supporting the Xcode project file updates is tricky - whenever new files are added, they need to explicitly be added to the Xcode project, and that's not possible if the developer making the changes is on Windows or Linux. If we move forward with this change, we'd probably remove the Xcode project and drop Xcode support. (Accordingly, we'd also no longer need to bug someone with Xcode whenever changes are being made that add/remove source files.) However, there are other IDEs on macOS that work great (and don't require multiple gigabyte downloads). Have you used VSCode? This project is already partially setup to work with VSCode - we could easily add a build configuration to VSCode that invokes cmake/make under the hood for you. Thanks for the feedback though - this is a big change to development workflow, and we only want to move forward with it if others agree. It's not worth the time otherwise!

(In case it doesn't go without saying....if we make a change here, we'll have to update all of the READMEs/wikis/etc, and that can include a tutorial to build with VSCode's UI).

LRFLEW:
> using two different sources for the dependencies seems like a bad idea.

Is there a particular reason why? I felt that having this split of system-installed dependencies for the CLI builds and bundling-compatible dependencies for .app builds made sense to me.

> We can continue to dynamically link, but we'd use the brew installed dylibs to build the .app bundle (we can run the steps performed [here](https://github.com/OpenRCT2/Dependencies/blob/master/macos_collect.sh#L40-L53) when copying the dylibs into the bundle).

That is highly problematic. That script was made when I was inactive on the project for a few years, and if I was there, I would have contested its inclusion. The Homebrew devs have stated to me in the past that they don't really support using their tool as a building step for self-contained releases ([source](https://github.com/Homebrew/homebrew-core/issues/6161#issuecomment-255581821) specifically about Qt). The reason I made OpenRCT2/Dependencies#25 is because I wanted to avoid using that script altogether. I would be against making that script part of the core build system.

> we'd create an openrct2-build container for macOS

Is that a thing? I didn't think there were Docker containers for macOS.

This does get into the core reason why using Homebrew as the main dependency source sounds dangerous to me. package managers can introduce breaking updates to their updates repository that breaks things on our end unexpectedly. IIRC, the point of using containers for Linux is because it freezes at least the major release versions of libraries so that there are no unexpected dependency errors. Homebrew doesn't make any such guaruntees, and will update projects so long as the updates don't conflict with the other libraries it tracks.

Most solutions also have this problem to some extent. vcpkg will also update projects to the latest version as soon as they can, but adding the Dependency project as a buffer helps prevent the updates being unexpected, as we have to manually create a new release of the dependencies.

> We could also use the static versions of the same libraries

Figured I should address the usage of dynamic libraries: This was a decision I pushed for early on, but wasn't made for strictly practical purposes. I've never been a fan of using static libraries for Windows, but that decision was made long before I got involved with all this, so I didn't have any input on that. I wanted to use dynamic libraries for the .app build simply because the way app bundles work made it seem like the intended solution. We *could* switch to static libraries, but I'm still not generally *happy* with that idea, and my previous issues with switching to Homebrew for dependency management still applies.

pizza2004:
I mean, I have to have Xcode either way, so it's not saving me from the download not to use it, but I've literally never used anything *but* Xcode on Mac in the decade I've been programming on Mac, to the point where I'm so used to Xcode I can't even stand sitting down at a Windows machine to program because it's not Xcode.  It's not that I can't make it work other ways, but with my personal life issues it's just too exhausting to go through all the steps to do anything on the project if I can't use Xcode.

adam-bloom:
> Is there a particular reason why? I felt that having this split of system-installed dependencies for the CLI builds and bundling-compatible dependencies for .app builds made sense to me.

My concern is largely in line with the rest of your post - dependency version management. What happens when the test executable is built against a more recent version of a library than the one included in the app bundle? Or vice versa. I don't want to find out. :) (Unstated here, but I would like to change macOS CI to run the tests, as they're run for both Windows and Linux. That requires being able to build and execute the tests....and I was thinking they'd go outside the bundle since they don't need to be there. Unless you've got a good reason for dynamic linking there, I'd think static would do just fine.

> Is that a thing? I didn't think there were Docker containers for macOS.

There are not, so I'm not sure how this would work either.

I'm not very familiar with vcpkg, but I'd be concerned about vcpkg and Homebrew co-existing (looks like they can both write to /usr/local) - in other words, perhaps the system that is in place is the best we can do. From the little reading I've done, perhaps we just need two triplet files from vcpkg - one for x64 static and one for x64 dynamic? Maybe that's the best option. I've already written the logic in cmake to keep the local copy up to date (and download the release from GitHub if needed) - if we can also do that with the static libraries, then we can link against them for openrct2-cli and the test executables.

IntelOrca:
My thoughts:

* vcpkg is very convenient, particularly if it is pre-built on CI (whenever they change) and downloaded automatically at the beginning of a clean build on the developer's computer. This is what Windows/VS build does. CMake can be made to do the same thing but currently that hasn't been done yet, so it is up to Windows users to build the dependencies with vcpkg themselves. vcpkg is a library package manager (analogous to npm, nuget, or maven) instead of a system package manager (e.g. apt, pacman, homebrew).

* macOS has docker for Linux containers, but not macOS containers. So not sure there is anything useful here unless you can cross compile.

* Being convenient to use cmake (with vscode) would be nice on macOS. It is familiar across all platforms and an alternative to xcode. 

* Running the tests on macOS and macOS CI would be really good.

* Being able to add / remove source files without waiting for someone to edit xcode would be perfect. [OpenLoco*](https://github.com/OpenLoco/OpenLoco) only provides a cmake configuration, so it does not have this issue. If we use cmake for CI then the xcode project will most likely fall out of date. If that is an issue, the next xcode user could submit a PR to fix either by manually editing it or running a generate script. Alternatively you could just have a script that the developer runs after checkout to generate the xcode project.

By the looks of the changes, this PR currently only tweaks the cmake files, it isn't removing any current CI or xcode process. So only these cmake changes should be reviewed for now before debating removing xcode.

Gymnasiast:
> I mean, I have to have Xcode either way, so it's not saving me from the download not to use it, but I've literally never used anything _but_ Xcode on Mac in the decade I've been programming on Mac, to the point where I'm so used to Xcode I can't even stand sitting down at a Windows machine to program because it's not Xcode.

The problem here is that while it helps you contribute, having to spin up a rickety VM to add a file to the Xcode project (because none of us are on macOS) is incredibly tedious. It increases the already long time PRs are open. This probably sounds a bit cold, but I reckon it outweighs your contributions in terms of the increased time and effort it takes us to maintain.

> It's not that I can't make it work other ways, but with my personal life issues it's just too exhausting to go through all the steps to do anything on the project if I can't use Xcode.

You're right to put your personal life above contributing to a voluntary project - it is supposed to be voluntary and fun after all, and it should not feel like a chore. But obviously, the same applies to us: it shouldn't feel like work or a chore to us either.

adam-bloom:
Happy New Year, everyone! FYI, #13682 implements the first part of this (in a bit cleaner way!) to handle the dynamic linking and dependency fetching. After that is merged, the next step will be to implement the app bundling logic. Shall we leave this PR open as a draft until that is complete (mostly for the conversation)? Or close it? It'll be much cleaner to do the app bundle changes in a separate branch (and therefore new PR) to keep these changes a lot cleaner. I think the discussion here has been very useful, so I'm hesitant to close this PR and cut off conversation while these changes remain in progress.

Gymnasiast:
Keep it open for now and split off anything you want/need to separate branches.

adam-bloom:
I believe everything that this proposed is now done via other PRs.

PR #13572, opened 2020-12-11
Close #13386: Show a (hardcoded) GUI error message if en-GB.txt cannot be loaded

### LINKED ISSUE #13386: Show a (hardcoded) GUI error message if en-GB.txt cannot be loaded
Currently, OpenRCT2 will silently exit instead. This is especially a problem on Android, which requires a bit of an awkward setup.

This error message should tell the user their installation is damaged.

Lukasz-Pe:
That's my idea how it could be done.

Drawbacks:
 - Message is shown twice (as is the message in console)
 - I don't know if this method is good for Android, I followed code in `Drawing.Sprite.cpp` file.

Small things:
 - Icon shown in message box suggests an information (that is also valid for other error messages using this code)

### DIFF
--- a/src/openrct2/localisation/LanguagePack.cpp
+++ b/src/openrct2/localisation/LanguagePack.cpp
@@ -17,7 +17,6 @@
 #include "../core/String.hpp"
 #include "../core/StringBuilder.h"
 #include "../core/StringReader.h"
-#include "../ui/UiContext.h"
 #include "Language.h"
 #include "Localisation.h"
 
@@ -88,9 +87,6 @@ public:
         {
             Memory::Free(fileData);
             log_error("Unable to open %s: %s", path, ex.what());
-            auto uiContext = OpenRCT2::GetContext()->GetUiContext();
-            uiContext->ShowMessageBox(
-                "Missing language file: " + static_cast<std::string>(path) + "\n\nYour installation may be damaged.");
             return nullptr;
         }
 
--- a/src/openrct2/localisation/LanguagePack.cpp
+++ b/src/openrct2/localisation/LanguagePack.cpp
@@ -9,7 +9,6 @@
 
 #include "LanguagePack.h"
 
-#include "../Context.h"
 #include "../common.h"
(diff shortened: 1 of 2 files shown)


Lukasz-Pe:
Could the test fail because of opening window with error message?

src/openrct2/localisation/LanguagePack.cpp
```diff
@@ -86,6 +88,9 @@ class LanguagePack final : public ILanguagePack
         {
             Memory::Free(fileData);
             log_error("Unable to open %s: %s", path, ex.what());
+            auto uiContext = OpenRCT2::GetContext()->GetUiContext();
```
Gymnasiast, line 91: I think it would be better to move this call to the `Context.cpp`, line 366 - the catch clause that is called when it fails to load *any* language. That would also unbreak the Language Pack test.

Lukasz-Pe: Worked like a charm:)

tupaschoal:
Should we add an error code? Because then if someone doesn't know English they can search for that.

I don't think we have error codes at all, maybe we should :)

Gymnasiast:
That would require documentation to be available in their native language as well. We're currently struggling to provide _English_ documentation, never mind other languages.

Not that your idea of adding error codes is without merit, but it's not going to help us in this case, at this moment. I recommend creating a Github issue for it.

src/openrct2/localisation/LanguagePack.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "LanguagePack.h"
 
+#include "../Context.h"
```
Gymnasiast, line 12: Please revert this addition. It is no longer necessary.

Gymnasiast (commented):
Needs one little line note addressed,

I´m going to test this on an Android phone with missing language files (the "use" case I had in mind).

Gymnasiast (approved):
Correctly shows an error message on Android.

PR #13573, opened 2020-12-11
Move remaining game actions logic from header to source

tupaschoal:
A follow-up of #13548 and #13571 to make sure everything is defined inside the `cpp` file, preventing further compilation


Gymnasiast (approved):

duncanspumpkin:
I don't see a huge point of moving the constructors into the cpp files but I'm not that fussed.

PR #13575, opened 2020-12-11
Feature: Add allow_remote_hosts to plugin section of the config

CorySanin:
Gives plugin developers more flexibility while continuing to prevent users from accidentally shooting themselves in the foot.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -8,6 +8,7 @@
 - Feature: [#13495] [Plugin] Add properties for park value, guests and company value.
 - Feature: [#13509] [Plugin] Add ability to format strings using OpenRCT2 string framework.
 - Feature: [#13512] [Plugin] Add item separators to list view.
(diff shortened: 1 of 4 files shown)


Gymnasiast:
This is certainly _a_ solution to the problem, but I wonder if it might be a better idea if it was a whitelist instead. That would at least prevent the user from having to throw all protection out of the window when all they need is one external server to talk to.

But I'd like to hear from other people first, so don't modify it just yet.

Sadret:
In my understanding, the idea of the localhost restriction protects the user from plugins doing bad things.
This change would effectively nullify that. As a user, I would have no influence with which servers the plugin that I use are communicating with.
In my opinion, if we want to have any protection, it has to come in the shape that the user **actively** has to allow a plugin to have connections.

CorySanin:
Closing because it seems #13583 is universally preferred to this approach.

PR #13577, opened 2020-12-12
Fix #13576: Remove unrequired code and casts from sprite export

### LINKED ISSUE #13576: Sprite exporter fails on 3D sign objects
**OS:** Arch Linux
**Version:** 0.3.2 (develop)
**Commit/Build:** 3feefb0

When trying to export a 3D sign object (SSIG*.DAT) using `openrct2 sprite exportalldat`, the sprite exporter fails with the error "Image width is zero in IHDR" and only the first 12 sprites are successfully extracted. The sprites which fail to extract are the character glyphs.

**Steps to reproduce:**
1. `mkdir images`
2. `openrct2 sprite exportalldat SSIG2`

duncanspumpkin:
As not all objects load images in the same order the previous code could end up with sprite index's that were incorrect. There also was no reason to make the code object specific as we can just access the image table directly.

### DIFF
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -194,7 +194,7 @@ static void sprite_file_close()
     SafeFree(spriteFileData);
 }
 
-static bool sprite_file_export(rct_g1_element* spriteHeader, const char* outPath)
+static bool sprite_file_export(const rct_g1_element* spriteHeader, const char* outPath)
 {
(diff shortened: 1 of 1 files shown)


Gymnasiast (approved):
Appears to work correctly.

I wrote the offending code, but I have no idea why I did it like this. It's probably a copy-and-paste job that I beat until it worked correctly.

IntelOrca (approved):

PR #13583, opened 2020-12-14
Feature: Add allowed_hosts to plugin section of config

CorySanin:
It's like #13575 but it uses a whitelist instead. Probably not the cleanest implementation, but with both PRs open the ball is officially in your court 👀

### DIFF
--- a/src/openrct2/config/Config.h
+++ b/src/openrct2/config/Config.h
@@ -204,6 +204,7 @@ struct FontConfiguration
 struct PluginConfiguration
 {
     bool enable_hot_reloading;
+    std::string allowed_hosts;
 };
 
 enum SORT
--- a/src/openrct2/scripting/ScSocket.hpp
+++ b/src/openrct2/scripting/ScSocket.hpp
@@ -83,20 +83,20 @@ namespace OpenRCT2::Scripting
             return s == "localhost" || s == "127.0.0.1" || s == "::";
         }
 
-        static bool IsHostOnWhitelist(std::string_view s)
+        static bool IsOnWhiteList(std::string_view host)
         {
-            char delimited = ',';
+            constexpr char delimiter = ',';
             size_t start_pos = 0;
             size_t end_pos = 0;
-            while ((end_pos = gConfigPlugin.allowed_hosts.find(delimited, start_pos)) != std::string::npos)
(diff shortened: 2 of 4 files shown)


IntelOrca:
I think this is the better solution.

src/openrct2/scripting/ScSocket.hpp
```diff
@@ -83,6 +83,22 @@ namespace OpenRCT2::Scripting
             return s == "localhost" || s == "127.0.0.1" || s == "::";
         }
 
+        static bool IsHostOnWhitelist(std::string_view s)
```
tupaschoal, line 86:
```suggestion
        static bool IsOnWhitelist(std::string_view host)
```

I'd prefer this name, from the usage below you'll see that it would avoid a redundancy on having two hosts consecutively.

src/openrct2/scripting/ScSocket.hpp
```diff
@@ -83,6 +83,22 @@ namespace OpenRCT2::Scripting
...
+        {
+            char delimited = ',';
```
tupaschoal, line 88:
```suggestion
            constexpr char delimiter = ',';
```

src/openrct2/config/Config.h
```diff
@@ -204,6 +204,7 @@ struct FontConfiguration
 struct PluginConfiguration
 {
     bool enable_hot_reloading;
+    std::string allowed_hosts;
```
tupaschoal, line 207: Usage would be a lot clearer if we stored this as a container somehow, instead of a comma separated list, but I'm not sure if this is possible (may be a limitation of the serialiser)

We might want to store it as a container and "split" on load, because at least that will make querying easier.

Gymnasiast:
@IntelOrca What do you think about the implementation?

IntelOrca (approved):

IntelOrca:
Looks fine to me

src/openrct2/config/Config.h
Gymnasiast: I suppose it would be possible to add a global vector with whitelisted IPs, to be loaded/saved from the config.

Since IntelOrca is okay with the current implementation, I think that that is best left for another PR.

tupaschoal:
Should have had the PR number and `[plugin]` on the changelog.

Gymnasiast:
@tupaschoal Now fixed.

tupaschoal:
Thanks

PR #13586, opened 2020-12-14
Restore extra cheat track pieces for some rides; intended to fix #13324

### LINKED ISSUE #13324: Not all Corkscrew RC pieces shown with "Show all track pieces" on
**OS:** Ubuntu 20.04
**Version:** 0.3.1
**Commit/Build:** PPA, version 0.3.1-develop-0\~1993-202010300552-gbbf91d961\~ubuntu20.04.1

Just like #13029, except that this is now for the Corkscrew Roller Coaster. The flat to steep pieces that normally appear for the Hypercoaster no longer appear even with the "Show all track pieces" cheat enabled. The only way around this is to enable the "arbitrary ride types" cheat, and change a corkscrew roller coaster to a hypercoaster, in order to access these pieces.

**Steps to reproduce:**
1. Build any corkscrew roller coaster design, even a custom one.
2. Enable the "Show all track pieces" cheat. The flat to steep pieces are inaccessible, even with this cheat turned on.
3. The sign for this is that the steep buttons are not enabled when the selected piece (from custom design) is a flat piece.

RQWorldblender:
Includes:
* Corkscrew Roller Coaster
* Hypercoaster
* Steel Wild Mouse
* Spinning Wild Mouse
* Water Coaster
* Car Ride
* Monster Trucks

Some pieces are incorrectly labelled (log bumps and rotation control toggle),
and need to be investigated further

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2914,18 +2914,16 @@ static void window_ride_construction_update_widgets(rct_window* w)
         {
             // Enable helix
             window_ride_construction_widgets[WIDX_SLOPE_DOWN_STEEP].type = WindowWidgetType::FlatBtn;
-            window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::FlatBtn;
-            if (rideType != RIDE_TYPE_SPLASH_BOATS && !gCheatsEnableAllDrawableTrackPieces)
-                window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::Empty;
+            if (rideType != RIDE_TYPE_SPLASH_BOATS && rideType != RIDE_TYPE_RIVER_RAFTS)
+                window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::FlatBtn;
         }
     }
 
     if (is_track_enabled(TRACK_SLOPE_STEEP))
     {
         window_ride_construction_widgets[WIDX_SLOPE_DOWN_STEEP].type = WindowWidgetType::FlatBtn;
-        window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::FlatBtn;
-        if (rideType == RIDE_TYPE_SPLASH_BOATS && !gCheatsEnableAllDrawableTrackPieces)
-            window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::Empty;
+        if (rideType != RIDE_TYPE_SPLASH_BOATS && rideType != RIDE_TYPE_RIVER_RAFTS)
+            window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::FlatBtn;
     }
 
     int32_t x;
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2245,7 +2245,7 @@ static void window_ride_construction_invalidate(rct_window* w)
     if (_currentTrackCurve & RideConstructionSpecialPieceSelected)
     {
         stringId = RideConfigurationStringIds[_currentTrackCurve & ~RideConstructionSpecialPieceSelected];
-        if (stringId == STR_RAPIDS && ride->type == RIDE_TYPE_MONSTER_TRUCKS)
+        if (stringId == STR_RAPIDS && (ride->type == RIDE_TYPE_MONSTER_TRUCKS || ride->type == RIDE_TYPE_CAR_RIDE))
         {
             stringId = STR_LOG_BUMPS;
         }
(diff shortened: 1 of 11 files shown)


Gymnasiast:
> Some pieces are incorrectly labelled (log bumps and rotation control toggle), and need to be investigated further

Log bumps and rapids behave exactly the same, so the game considers them one piece, just with an overridden label for some ride types.

What is incorrect about the rotation control toggle?

RQWorldblender:
> > Some pieces are incorrectly labelled (log bumps and rotation control toggle), and need to be investigated further
> 
> Log bumps and rapids behave exactly the same, so the game considers them one piece, just with an overridden label for some ride types.
> 
> What is incorrect about the rotation control toggle?

It appears as if it is a booster, complete with the the ability to set the speed.

RQWorldblender:
I'm also looking into enabling slopes for the River Rafts, which I succeeded in, but I haven't gotten the same with the Splash Boats. But I don't know how I'm supposed to enable the upwards steep slopes, even with the all track pieces drawn cheat.

Gymnasiast:
> It appears as if it is a booster, complete with the the ability to set the speed. It's even labelled as a booster, not a spinning control toggle.
> I would prefer that the labels be correct, unless the different labels are intended behavior.

The reason is that the spinning control toggle and the booster share an ID. This is for RCT1/RCTC compatibility, but also because there is simply no more room. We do a simple ride type check: if it's a spinning wild mouse, then it's a rotation control toggle. Otherwise, it's a booster. You should search the code base for RIDE_TYPE_SPINNING_WILD_MOUSE and amend the checks as necessary. Then the labels should be correct as well.

RQWorldblender:
Finally, I've fixed the problems with Car Ride and Splash Boats, but the problem with Spinning Wild Mouse and Steel Wild Mouse is taking me a lot longer. Even when I think I set up the code correctly, all rides, including both Spinning Wild Mouse and Steel Wild Mouse, are all labelling Splnning Control Toggle as Boosters.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2913,16 +2915,18 @@ static void window_ride_construction_update_widgets(rct_window* w)
         {
             // Enable helix
             window_ride_construction_widgets[WIDX_SLOPE_DOWN_STEEP].type = WindowWidgetType::FlatBtn;
-            if (rideType != RIDE_TYPE_SPLASH_BOATS)
-                window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::FlatBtn;
+            window_ride_construction_widgets[WIDX_SLOPE_UP_STEEP].type = WindowWidgetType::FlatBtn;
```
Gymnasiast, line 2918: Why did you change these?

RQWorldblender: It was hard for me to come up with the best way to not inadvertently disable the steep slope up for other rides. I was trying to get the steep slope up button enabled for Splash Boats only when the show all drawable track pieces cheat is enabled. It will still be disabled when that cheat is turned off. I just could not easily determine the best way to code this without disabling it for all other rides.

Gymnasiast: Unless you swapped around enabled/disabled around by accident, I cannot make much sense of your answer:

> I was trying to get the steep slope up button enabled for Splash Boats

Why would you want to enable the steep slope up for Splash Boats anyway? The up pieces have rollers and look distinct from the down pieces, and I don't remember a sprite existing for a steep up piece.

RQWorldblender: During my testing, I learned that the Splash Boat vehicles can actually climb up the steep slope up pieces (the River Rafts vehicles can also traverse slopes, both up and down). Steep slope up reuses the same sprites as those for steep slope down, but it apparently works.

Gymnasiast: The track lacks them, which was my point:

afbeelding

The cheat is there to enable _drawable_ track pieces. I do not count this as such. Please revert the changes to the Splash Boats and this bit of code.

RQWorldblender: Okay then. Before I complete this, should I also apply these changes to the River Rafts, or leave them as-is? If I revert this code, the steep slopes will become accessible to the River Rafts, but not Splash Boats.

Gymnasiast: Restore the old code, but add an extension for River Rafts.

In the longer run, I want to introduce a flag that says the ride uses a water channel, which can be checked instead of doing a ride type check. You can also do that in this PR if you feel like trying that.

RQWorldblender: > In the longer run, I want to introduce a flag that says the ride uses a water channel, which can be checked instead of doing a ride type check. You can also do that in this PR if you feel like trying that.
Cool, I think that might be a better fit. First, I need to know how to identify a ride as having a water channel.
Side note: OpenGL rendering is causing a bunch of problem for me:
```
openrct2 $ openrct2: /home/richard/OpenRCT2/src/openrct2-ui/drawing/engines/opengl/TextureCache.h:116: AtlasTextureInfo Atlas::Allocate(int32_t, int32_t): Assertion `!_freeSlots.empty()' failed.
                    Aborted (core dumped)
```
Should I revert to an earlier commit to get around this problem, or use software rendering for now? I tried that, but the font looks all messed up! I know that other people have reported this problem.

Gymnasiast: That was a bad commit that was merged on Friday morning and fixed in the afternoon. I have already rebased your branch against the most recent commit, so you should not have any problems. If you still have, make sure you rebuild g2.dat.

Rides with water channels are Log Flume, Splash Boats and River Rafts.

RQWorldblender: I think I have this issue fixed now; however identifying which rides are using water channels looks like to be out of scope for what I want to merge here. It may appear in another pull request in the future.

PR #13587, opened 2020-12-14
Enhanced track designer with ability to add/remove scenery and footpaths

RQWorldblender:
Enhance the track designer for feature parity with Rollercoaster Tycoon Classic (RCTC). RCTC features the ability to add scenery and footpaths, unlike vanilla RCT2.

Includes:
* Re-enabled buttons: Ride List, Map, Scenery, Path, Clear Scenery
* Add all standard footpaths and core scenery objects to the object selection
    
The max limit of 4 rides has been removed (up to 128 rides can now be selected), and only a minimum set of scenery is enabled. I wish to add all of the standard scenery groups, but this would exceed the max limit of 19 scenery groups. Since I don't know which other scenery groups should be enabled, besides the core ones, I'll just leave the selection at the core set.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -24,6 +24,7 @@
 - Fix: [#13510] [Plugin] list view scroll resets when items is set.
 - Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 - Improved: [#13386] A GUI error message is now displayed if the language files are missing.
+- Improved: [#13587] Allow up to 128 ride objects to be selected in track designer.
 - Removed: [#13423] Built-in explode guests cheat (replaced by plug-in).
 
 0.3.2 (2020-11-01)
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -9,6 +9,7 @@
 - Feature: [#13509] [Plugin] Add ability to format strings using OpenRCT2 string framework.
 - Feature: [#13512] [Plugin] Add item separators to list view.
 - Feature: [#13583] Add allowed_hosts to plugin section of config.
+- Feature: [#13587] Enhanced track designer with ability to add/remove scenery and footpaths.
 - Change: [#13346] Change FootpathScenery to FootpathAddition in all occurrences.
 - Fix: [#12895] Mechanics are called to repair rides that have already been fixed.
 - Fix: [#13257] Rides that are exactly the minimum objective length are not counted.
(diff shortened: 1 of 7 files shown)


Gymnasiast:
> The max limit of 4 rides has not changed

Let's get rid of that one as well.

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -353,6 +355,21 @@ static void window_editor_object_selection_select_default_objects()
     }
 }
 
+static void select_designer_objects()
```
duncanspumpkin, line 358: All new functions are to be UpperCamelCase

RQWorldblender: It's done as of now. The new function is SelectDesignerObjects().

Gymnasiast (commented):
Looks OK codewise, but still needs testing.

Please add a changelog entry. Use this PR number in the list.

distribution/changelog.txt
```diff
@@ -9,6 +9,7 @@
 - Feature: [#13509] [Plugin] Add ability to format strings using OpenRCT2 string framework.
 - Feature: [#13512] [Plugin] Add item separators to list view.
 - Feature: [#13583] Add allowed_hosts to plugin section of config.
+- Feature: [#13587] Enhanced track designer with ability to add/remove scenery and footpaths.
```
duncanspumpkin, line 12: and allows for up to 128 ride types to be loaded. or something along those lines.

Gymnasiast: Perhaps create two entries, as this PR adds a feature (scenery) and an improvement (more rides).

distribution/changelog.txt
```diff
@@ -23,6 +24,7 @@
 - Fix: [#13510] [Plugin] list view scroll resets when items is set.
 - Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 - Improved: [#13386] A GUI error message is now displayed if the language files are missing.
+- Improved [#13587] Allow up to 128 ride objects to be selected in track designer.
```
Gymnasiast, line 27:
```suggestion
- Improved: [#13587] Allow up to 128 ride objects to be selected in track designer.
```

Gymnasiast (approved):

RQWorldblender:
Should I add more scenery groups, or is that core one good enough for now? I would like to add all of the standard ones, but they exceed the limit of 19 scenery groups.

Gymnasiast:
Leave it as-is for now, I'd say. This selection has the big advantage of being available in every scenario.

duncanspumpkin (approved):

RQWorldblender:
Hold on. I just discovered that the tracks with scenery can't be saved. I need to enable the button before all this gets merged.

RQWorldblender:
Thank you @Gymnasiast. I've been looking at the ride list window, and I noticed that some tabs could be disabled (shops and stalls, restrooms and information kiosks), along with hiding some of the sort options. Some of them aren't applicable here.

PR #13590, opened 2020-12-15
Cmdsprite 2

duncanspumpkin:
The previous fix identified that this file was in need of some attention.
I've modified it to use some more cpp style functions and classes.
Each commit should be compliable on its own and shows the incremental steps involved with modifying it.

### DIFF
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -146,27 +146,24 @@ bool SpriteFile::Save(const utf8* path)
     }
 }
 
-static bool sprite_file_export(const rct_g1_element* spriteHeader, const char* outPath)
+static bool SpriteImageExport(const rct_g1_element& spriteElement, const char* outPath)
 {
-    rct_drawpixelinfo dpi;
-    uint8_t* pixels;
-    int32_t pixelBufferSize;
-
-    pixelBufferSize = spriteHeader->width * spriteHeader->height;
+    const auto pixelBufferSize = spriteElement.width * spriteElement.height;
     std::unique_ptr<uint8_t[]> pixelBuffer(new uint8_t[pixelBufferSize]);
-    pixels = pixelBuffer.get();
+    auto pixels = pixelBuffer.get();
     std::fill_n(pixels, pixelBufferSize, 0x00);
 
+    rct_drawpixelinfo dpi;
     dpi.bits = pixels;
     dpi.x = 0;
     dpi.y = 0;
-    dpi.width = spriteHeader->width;
-    dpi.height = spriteHeader->height;
+    dpi.width = spriteElement.width;
+    dpi.height = spriteElement.height;
     dpi.pitch = 0;
     dpi.zoom_level = 0;
 
     DrawSpriteArgs args(
-        &dpi, ImageId(), PaletteMap::GetDefault(), *spriteHeader, 0, 0, spriteHeader->width, spriteHeader->height, pixels);
+        &dpi, ImageId(), PaletteMap::GetDefault(), spriteElement, 0, 0, spriteElement.width, spriteElement.height, pixels);
     gfx_sprite_to_buffer(args);
 
     auto const pixels8 = dpi.bits;
@@ -190,7 +187,7 @@ static bool sprite_file_export(const rct_g1_element* spriteHeader, const char* o
     }
 }
 
-static std::optional<ImageImporter::ImportResult> sprite_file_import(
+static std::optional<ImageImporter::ImportResult> SpriteImageImport(
     const char* path, int16_t x_offset, int16_t y_offset, bool keep_palette, bool forceBmp, int32_t mode)
 {
     try
@@ -298,8 +295,8 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
             return -1;
         }
 
-        rct_g1_element* spriteHeader = &sFile->Entries[spriteIndex];
-        if (!sprite_file_export(spriteHeader, outputPath))
+        const auto& spriteHeader = sFile->Entries[spriteIndex];
+        if (!SpriteImageExport(spriteHeader, outputPath))
         {
             fprintf(stderr, "Could not export\n");
             return -1;
@@ -357,8 +354,8 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
                 printf("\r%d / %d, %d%%", spriteIndex, maxIndex, spriteIndex / maxIndex);
             }
 
-            rct_g1_element* spriteHeader = &sFile->Entries[spriteIndex];
-            if (!sprite_file_export(spriteHeader, outputPath))
+            const auto& spriteHeader = sFile->Entries[spriteIndex];
+            if (!SpriteImageExport(spriteHeader, outputPath))
             {
                 fprintf(stderr, "Could not export\n");
                 return -1;
@@ -444,8 +441,8 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
 
         for (uint32_t spriteIndex = 0; spriteIndex < maxIndex; spriteIndex++)
         {
-            const rct_g1_element g1 = metaObject->GetImageTable().GetImages()[spriteIndex];
-            if (!sprite_file_export(&g1, outputPath))
+            const auto& g1 = metaObject->GetImageTable().GetImages()[spriteIndex];
+            if (!SpriteImageExport(g1, outputPath))
             {
                 fprintf(stderr, "Could not export\n");
                 return -1;
@@ -513,7 +510,7 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
             }
         }
 
-        auto importResult = sprite_file_import(imagePath, x_offset, y_offset, false, false, gSpriteMode);
+        auto importResult = SpriteImageImport(imagePath, x_offset, y_offset, false, false, gSpriteMode);
         if (importResult == std::nullopt)
             return -1;
 
@@ -604,7 +601,7 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
 
             auto imagePath = platform_get_absolute_path(strPath.c_str(), directoryPath);
 
-            auto importResult = sprite_file_import(
+            auto importResult = SpriteImageImport(
                 imagePath.c_str(), Json::GetNumber<int16_t>(x_offset), Json::GetNumber<int16_t>(y_offset), keep_palette,
                 forceBmp, gSpriteMode);
             if (importResult == std::nullopt)
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -23,8 +23,6 @@
 #include "object/ObjectRepository.h"
 #include "platform/platform.h"
 #include "util/Util.h"
-#include "world/Entrance.h"
-#include "world/Scenery.h"
 
 #include <cmath>
 #include <cstring>
@@ -36,22 +34,10 @@
 
 using namespace OpenRCT2::Drawing;
 
-#pragma pack(push, 1)
-
-struct rct_sprite_file_header
-{
-    uint32_t num_entries;
-    uint32_t total_size;
-};
-
-assert_struct_size(rct_sprite_file_header, 8);
-
-#pragma pack(pop)
-
 struct SpriteFile
 {
 public:
-    rct_sprite_file_header Header;
+    rct_g1_header Header;
     std::vector<rct_g1_element> Entries;
     std::vector<uint8_t> Data;
     void AddImage(ImageImporter::ImportResult& image);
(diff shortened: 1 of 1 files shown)


src/openrct2/CmdlineSprite.cpp
```diff
@@ -292,15 +293,15 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
         else if (argc == 2)
         {
             const char* spriteFilePath = argv[1];
-
-            if (!sprite_file_open(spriteFilePath))
+            auto sFile = sprite_file_open(spriteFilePath);
+            if (!sFile.has_value())
```
tupaschoal, line 228: (This is on commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/030ba011fa142136fd7ba7a9538911c9679a5eec)

FYI: In the rest of the code we generally compare to nullopt, I think

src/openrct2/CmdlineSprite.cpp
```diff
@@ -292,15 +293,15 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
         else if (argc == 2)
         {
             const char* spriteFilePath = argv[1];
-
-            if (!sprite_file_open(spriteFilePath))
+            auto sFile = sprite_file_open(spriteFilePath);
```
tupaschoal: (This is on commmit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/030ba011fa142136fd7ba7a9538911c9679a5eec)

I'm not sure I'd have changed `spriteFile` to ` sFile`, not sure it improved.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -46,9 +46,14 @@ assert_struct_size(rct_sprite_file_header, 8);
...
 #pragma pack(pop)
 
-static rct_sprite_file_header spriteFileHeader;
-static rct_g1_element* spriteFileEntries;
-static uint8_t* spriteFileData;
+struct SpriteFile
+{
+    rct_sprite_file_header Header;
+    rct_g1_element* Entries;
+    uint8_t* Data;
```
tupaschoal: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/ade064674ad5327edddbf9f5337eee72d7a5cdd6)

Ìnitialize pointers to `nullptr`

src/openrct2/CmdlineSprite.cpp
```diff
@@ -46,9 +46,14 @@ assert_struct_size(rct_sprite_file_header, 8);
 
 #pragma pack(pop)
 
-static rct_sprite_file_header spriteFileHeader;
-static rct_g1_element* spriteFileEntries;
-static uint8_t* spriteFileData;
+struct SpriteFile
+{
+    rct_sprite_file_header Header;
```
tupaschoal: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/ade064674ad5327edddbf9f5337eee72d7a5cdd6)

I can't comment on the `rct_sprite_file_header` declaration & definition, but it should have the member variables initialized to 0

src/openrct2/CmdlineSprite.cpp
```diff
@@ -543,9 +543,10 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
 
         const char* spriteFilePath = argv[1];
 
-        spriteFile.Header.num_entries = 0;
-        spriteFile.Header.total_size = 0;
-        sprite_file_save(spriteFilePath);
+        SpriteFile sFile;
+        sFile.Header.num_entries = 0;
+        sFile.Header.total_size = 0;
+        sFile.Save(spriteFilePath);
```
tupaschoal, line 468: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/3b39ab2299e88ae7cee242006590ab3ec0415b8f)

Shouldn't this be using the `static` variable `spriteFile`?

src/openrct2/CmdlineSprite.cpp
```diff
@@ -640,9 +641,10 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
 
         bool silent = (argc >= 4 && strcmp(argv[3], "silent") == 0);
 
-        spriteFile.Header.num_entries = 0;
-        spriteFile.Header.total_size = 0;
-        sprite_file_save(spriteFilePath);
+        SpriteFile baseFile;
+        baseFile.Header.num_entries = 0;
+        baseFile.Header.total_size = 0;
+        baseFile.Save(spriteFilePath);
```
tupaschoal, line 551: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/3b39ab2299e88ae7cee242006590ab3ec0415b8f)

Shouldn't this be using the `static` variable `spriteFile`?

src/openrct2/CmdlineSprite.cpp
```diff
@@ -53,6 +53,7 @@ struct SpriteFile
     rct_sprite_file_header Header;
     std::vector<rct_g1_element> Entries;
     std::vector<uint8_t> Data;
+    bool isAbsolute = false;
```
tupaschoal, line 48: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/48b3fe8944c9ece0036c71c037941a7fbc2ada17)

This should probably be private, as it's only needed inside the struct (which might become a class now).
That's probably true for the other member variables at this point too

src/openrct2/CmdlineSprite.cpp
```diff
@@ -114,55 +117,33 @@ std::optional<SpriteFile> SpriteFile::Open(const utf8* path)
...
-            outElement->offset = static_cast<uint32_t>(
-                (reinterpret_cast<uintptr_t>(inElement->offset) - reinterpret_cast<uintptr_t>(Data.data())));
-            outElement->width = inElement->width;
-            outElement->height = inElement->height;
-            outElement->x_offset = inElement->x_offset;
-            outElement->y_offset = inElement->y_offset;
-            outElement->flags = inElement->flags;
-            outElement->zoomed_offset = inElement->zoomed_offset;
-        }
+            bool wasAbsolute = isAbsolute;
```
tupaschoal, line 117: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/48b3fe8944c9ece0036c71c037941a7fbc2ada17)

`const bool`

src/openrct2/CmdlineSprite.cpp
```diff
@@ -114,55 +117,33 @@ std::optional<SpriteFile> SpriteFile::Open(const utf8* path)
...
+            stream.Write(Data.data(), Header.total_size);
 
-        if (fwrite(Data.data(), Header.total_size, 1, file) != 1)
-        {
-            fclose(file);
-            return false;
+            if (wasAbsolute)
+            {
+                MakeEntriesAbsolute();
+            }
```
tupaschoal, line 129: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/48b3fe8944c9ece0036c71c037941a7fbc2ada17)

If an exception is thrown, we never make it absolute again. I'd use an RAII solution for this
```c++
ScopedRelativeSpriteFile(SpriteFile& sFile, bool isAbsolute)
    : Spritefile(sFile), WasAbsolute(isAbsolute)
{
    if (WasAbsolute)
        sFile.MakeEntriesRelative();
}

~ScopedRelativeSpriteFile()
{
    if (WasAbsolute)
        sFile.MakeEntriesAbsolute();
}
```

And then just instantiate this before writing to the stream.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -77,12 +78,14 @@ void SpriteFile::MakeEntriesAbsolute()
...
     for (auto& entry : Entries)
         entry.offset += reinterpret_cast<uintptr_t>(Data.data());
+    isAbsolute = true;
 }
 
 void SpriteFile::MakeEntriesRelative()
 {
     for (auto& entry : Entries)
         entry.offset -= reinterpret_cast<uintptr_t>(Data.data());
+    isAbsolute = false;
```
tupaschoal: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/48b3fe8944c9ece0036c71c037941a7fbc2ada17)

We should probably only do each of these if `isAbsolute` is set to a valid value, e.g., only relativize if absolute and only make absolute if relative. Otherwise we might have some problems.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -88,6 +78,23 @@ void SpriteFile::MakeEntriesRelative()
...
 
+void SpriteFile::AddImage(ImageImporter::ImportResult& image)
+{
+    Header.num_entries++;
+    Header.total_size += static_cast<uint32_t>(image.Buffer.size());
+    Entries.reserve(Header.num_entries);
+
+    MakeEntriesRelative();
+    Data.reserve(Header.total_size);
+    MakeEntriesAbsolute();
```
tupaschoal: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/8a91330eb7df1e56de3ceddce95937d94396ce82)

Another case for the scoped relative/absolute RAII class.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -50,30 +50,20 @@ assert_struct_size(rct_sprite_file_header, 8);
...
-    auto pathW = String::ToWideChar(path);
-    auto modeW = String::ToWideChar(mode);
-    auto file = _wfopen(pathW.c_str(), modeW.c_str());
-    return file;
-}
-
-#    define fopen fopen_utf8
-
-#endif
-
```
tupaschoal: (On commit https://github.com/OpenRCT2/OpenRCT2/pull/13590/commits/8a91330eb7df1e56de3ceddce95937d94396ce82)

Not sure this became useless on previous commits or if it was never touched in this PR at all, but this cleanup seems a bit lost here.

src/openrct2/CmdlineSprite.cpp
```diff
@@ -36,22 +34,10 @@
 
 using namespace OpenRCT2::Drawing;
 
-#pragma pack(push, 1)
```
tupaschoal, line 37: There is a warning being disabled on line 10 (I can't really comment there). Maybe it's not applicable anymore? (I'm not sure, I haven't double checked)

src/openrct2/CmdlineSprite.cpp
duncanspumpkin: It was due to wanting to avoid a name clash whilst half way through the changes

src/openrct2/CmdlineSprite.cpp
duncanspumpkin: It could have been removed after the second file stream commit

src/openrct2/CmdlineSprite.cpp
duncanspumpkin: pretty sure we just do `if (!sFile)`. I've started to think that has_value is more clear though

src/openrct2/CmdlineSprite.cpp
duncanspumpkin: Removed in subsequent commits with a std::vector.

src/openrct2/CmdlineSprite.cpp
duncanspumpkin: no it is localised to just this branch of the if no need to be the global

src/openrct2/CmdlineSprite.cpp
duncanspumpkin: no it is localised to just this branch of the if no need to be the global

src/openrct2/CmdlineSprite.cpp
duncanspumpkin: Made private in subsequent commit

src/openrct2/CmdlineSprite.cpp
tupaschoal: > pretty sure we just do if (!sFile)

I don't think so, because I prefer that over the comparison to nullopt or nullptr, but I was dissuaded at the time :D

But `has_value()` is clearer, sure.

tupaschoal (approved):
Seems alright now!

Gymnasiast:
This was obviously NOT tested.





And it outright **crashes** when I try OpenGL mode.

@duncanspumpkin @tupaschoal Why did neither of you catch this?

duncanspumpkin:
I tested the cmd line sprite bits didn't think it would change anything in game as its standalone.

Gymnasiast:
Standalone or not, the command line bits are used to build g2.dat, which _is_ used in the game. But even if it wasn't: if you change anything, you should verify that it still produces the correct results. 
In this case, you should have tested if the resulting g2.dat still worked the same as before.

I'm going to revert the merge, so this can be fixed and reviewed **properly**.

duncanspumpkin:
I'm just about to fix it.

PR #13592, opened 2020-12-15
Fix #13567: Added ability for peeps to stop eating certain food constanly

### LINKED ISSUE #13567: several bugs with food and drink
**OS:** Windows10
**Version:** 0.3.2-196
**Commit/Build:** f09b14e

1. Guests buy multiple food or drink.
ex) Guests buy a cup of sujeonggwa even though they already have a sandwich. 

2. Only the last one actually functions. 
ex) If a guest buys a sandwich when he has a cup of hot chocoate already, only the hunger level goes down not thirsty even though both a sandwich and a cup of hot chocolate exist on his inventory.

4. Food and drink in the list of inventory of guests will never disappear. 
ex) All guests who have drink or food are sitting on the bench forever.  

**Steps to reproduce:**
1. Build any food stall or drink stall
2. Create guests and make them hungry or thirsty so that they buy something

**Screenshots / Video:**
미개척된 숲 2020-12-09 23-22-20
미개척된 숲 2020-12-09 23-22-56
As you can see, Iain C has been sitting for over an year.

미개척된 숲 2020-12-10 20-14-14
Guests buy multiple food or drink.

제목 없음
The stalls with bugs 
( +Chicken nuggets stall)

sorry for my terrible english :)

duncanspumpkin:
Fix #13567: Added ability for peeps to stop eating certain food constanly

Incorrect assumption that bitscanforward iterated over 64 bits meant that food that was previously within the ExtraItemFlags would never get removed from the peeps inventory. bitscanforward function has been replaced with a 64bit version

### DIFF
--- a/src/openrct2/util/Util.cpp
+++ b/src/openrct2/util/Util.cpp
@@ -172,12 +172,34 @@ int32_t bitscanforward(int32_t source)
     int32_t success = __builtin_ffs(source);
     return success - 1;
 #else
-#    pragma message "Falling back to iterative bitscan forward, consider using intrinsics"
+#    pragma message("Falling back to iterative bitscan forward, consider using intrinsics")
     // This is a low-hanging optimisation boost, check if your compiler offers
     // any intrinsic.
     // cf. https://github.com/OpenRCT2/OpenRCT2/pull/2093
     for (int32_t i = 0; i < 32; i++)
         if (source & (1u << i))
+            return i;
+
+    return -1;
+#endif
+}
+
+int32_t bitscanforward(int64_t source)
+{
+#if defined(_MSC_VER) && (_MSC_VER >= 1400) && defined(_M_X64) // Visual Studio 2005
+    DWORD i;
+    uint8_t success = _BitScanForward64(&i, static_cast<uint64_t>(source));
+    return success != 0 ? i : -1;
+#elif defined(__GNUC__)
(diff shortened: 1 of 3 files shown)


IntelOrca (approved):

tupaschoal (commented):
The 32-bit win build is complaining

src/openrct2/util/Util.cpp
```diff
@@ -172,12 +172,34 @@ int32_t bitscanforward(int32_t source)
...
     return success - 1;
 #else
-#    pragma message "Falling back to iterative bitscan forward, consider using intrinsics"
+#    pragma message("Falling back to iterative bitscan forward, consider using intrinsics")
     // This is a low-hanging optimisation boost, check if your compiler offers
     // any intrinsic.
     // cf. https://github.com/OpenRCT2/OpenRCT2/pull/2093
     for (int32_t i = 0; i < 32; i++)
-        if (source & (1u << i))
+        if (source & (1ull << i))
```
Gymnasiast, line 180: Is it necessary to also change the 32-bit version?

duncanspumpkin: It's not really necessary is because I thought i could get rid of the 32 bit version but realised it would prevent optimisations on x86-32

IntelOrca: Why are you casting to a 64-bit type in the 32-bit version?

PR #13593, opened 2020-12-15
[Plugin] Add track progress property to vehicles

Basssiiie:
Hey all, this is my draft attempt at adding a property for track progress to vehicles for plugins. This would be very useful for my RideVehicleEditor plugin to adjust spacing between vehicles, and maybe others can find some usefulness in it as well!

- Track progress can be read and set for Car-instances.
- ~~If the chosen value is beyond the maximum amount of the current track piece, the vehicle will go to the relative position on the next track piece.~~
- If the chosen value is negative, the vehicle will go to the relative position on the previous track piece.

What I'm still planning to do:
- [x] ~~Clamp track progress at the end of tracks (for non-complete circuits).~~
- [x] ~~Sometimes if it jumps to the next or previous track, the position is off by 1.~~
- [x] ~~Test it thoroughly to make sure it doesn't crash or mess up the game somehow (so far it doesn't seem to).~~
- [x] Run clang-format.

Update 21 dec:
- After talking with Duncan, the proposed implementation has been changed to modify the vehicles 'remaining distance' instead.
- Duncan provided the related #13607 update flag which this implementation uses.
- The 'track progress' setter has been removed and replaced by a 'move' method which moves the car a relative distance in either forward or backwards along the track.

A screenshot of it working ingame:
afbeelding

Right now it works surprisingly well and I haven't really found any issues so far. 

I'd love to hear feedback though. This is my first pull request and I'm not even sure if this is the best approach for it. Please let me know if this is something that would fit into the source code and if there is anything I should change.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -1064,6 +1064,11 @@ declare global {
          * Current status of the car or train.
          */
         status: VehicleStatus;
+		
+        /**
+         * The progress on the current track piece, in steps.
+         */
+        trackProgress: number;
 
         /**
          * List of peep IDs ordered by seat.
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -1085,7 +1085,7 @@ declare global {
          * position. A single visible step is about 8.000 to 14.000 in distance depending 
          * on the direction its moving in.
          */
-        move(distance: number): void;
+        travelBy(distance: number): void;
     }
 
     type VehicleStatus =
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -1064,12 +1064,12 @@ declare global {
          * Current status of the car or train.
          */
         status: VehicleStatus;
-		
+
         /**
          * The progress on the current track piece, in steps.
          */
         readonly trackProgress: number;
-		
+
         /**
          * The currently projected remaining distance the car will travel.
          */
@@ -1079,13 +1079,13 @@ declare global {
          * List of peep IDs ordered by seat.
          */
         peeps: (number | null)[];
-				
+
         /**
          * Moves the vehicle forward or backwards along the track, relative to its current 
-		 * position. A single visible step is about 8.000 to 14.000 in distance depending 
-		 * on the direction its moving in.
+         * position. A single visible step is about 8.000 to 14.000 in distance depending 
+         * on the direction its moving in.
          */
-		move(distance: number): void;
+        move(distance: number): void;
     }
 
     type VehicleStatus =
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -1068,12 +1068,24 @@ declare global {
         /**
          * The progress on the current track piece, in steps.
          */
-        trackProgress: number;
+        readonly trackProgress: number;
+		
+        /**
+         * The currently projected remaining distance the car will travel.
+         */
+        readonly remainingDistance: number;
 
         /**
          * List of peep IDs ordered by seat.
          */
         peeps: (number | null)[];
+				
+        /**
+         * Moves the vehicle forward or backwards along the track, relative to its current 
+		 * position. A single visible step is about 8.000 to 14.000 in distance depending 
+		 * on the direction its moving in.
+         */
+		move(distance: number): void;
     }
 
     type VehicleStatus =
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -825,16 +825,6 @@ void Vehicle::MoveRelativeDistance(int32_t distance)
     SetUpdateFlag(VEHICLE_UPDATE_FLAG_SINGLE_CAR_POSITION);
     UpdateTrackMotion(nullptr);
     ClearUpdateFlag(VEHICLE_UPDATE_FLAG_SINGLE_CAR_POSITION);
-
-    const Ride* curRide = GetRide();
-    const rct_vehicle_info* moveInfo = GetMoveInfo();
-
-    MoveTo({ TrackLocation.x + moveInfo->x, TrackLocation.y + moveInfo->y,
-             TrackLocation.z + moveInfo->z + RideTypeDescriptors[curRide->type].Heights.VehicleZOffset });
-
-    sprite_direction = moveInfo->direction;
-    bank_rotation = moveInfo->bank_rotation;
-    vehicle_sprite_type = moveInfo->vehicle_sprite_type;
 }
 
 Vehicle* try_get_vehicle(uint16_t spriteIndex)
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -818,6 +818,44 @@ uint16_t Vehicle::GetTrackProgress() const
     return vehicle_get_move_info_size(TrackSubposition, track_type);
 }
 
+void Vehicle::SetRelativeTrackProgress(int16_t trackProgress)
+{
+    Ride* curRide = GetRide();
+
+    const uint16_t maximumProgress = GetTrackProgress();
+    const uint16_t trackType = GetTrackType();
+
+    if (trackProgress < 0)
+    {
+        uint16_t new_progress = 0;
+        if (UpdateTrackMotionBackwardsGetNewTrack(trackType, curRide, &new_progress))
+        {
+            SetRelativeTrackProgress(trackProgress + new_progress);
+        }
+        return;
+    }
+
+    if (trackProgress >= maximumProgress)
+    {
+        if (UpdateTrackMotionForwardsGetNewTrack(trackType, curRide, GetRideEntry()))
+        {
+            SetRelativeTrackProgress(trackProgress - maximumProgress);
+        }
+        return;
+    }
+
+    track_progress = trackProgress;
+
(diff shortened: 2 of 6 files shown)


James103:
Have you tried running clang-format on your changes?

Basssiiie:
> Have you tried running clang-format on your changes?

No I haven't, I was looking into how to do that after the checks failed. I'll fix that tonight. :)

duncanspumpkin:
This wont be able to set the correct sprite for vehicles going through reverser elements or rotate elements like log flumes, heartline.

duncanspumpkin:
I think it will need to hook in early to the UpdateMotion functions. The vehicle placement code uses UpdateMotion to place the vehicles and this ensures that all of the values are correct. I'm not too sure how you would achieve that though.

Basssiiie:
@duncanspumpkin Thank you for your reply. I'll give the remaining_distance a shot. What does it exactly hold? The distance to travel within this frame? Or the distance to the next track piece? Or something else?

duncanspumpkin:
From memory the units are the same as track progress. It's a sort of left over amount after a vehicle update when things don't quite line up

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -518,6 +519,23 @@ namespace OpenRCT2::Scripting
...
+            auto vehicle = GetVehicle();
+            return vehicle != nullptr ? vehicle->track_progress : 0;
+        }
+        void trackProgress_set(int16_t value)
+        {
+            ThrowIfGameStateNotMutable();
+            auto vehicle = GetVehicle();
+            if (vehicle != nullptr)
+            {
+                vehicle->Invalidate2();
```
tupaschoal, line 533: Do we need this first invalidate, as nothing has been performed yet?

IntelOrca: yes, otherwise it will only invalidate the screen at its new position and not its old position.

Basssiiie:
@duncanspumpkin I realise I checked in your code from #13607 here as well. Should I remove it or will it merge fine after your PR has been merged?

distribution/openrct2.d.ts
```diff
@@ -1068,12 +1068,24 @@ declare global {
...
         /**
          * List of peep IDs ordered by seat.
          */
         peeps: (number | null)[];
+				
+        /**
+         * Moves the vehicle forward or backwards along the track, relative to its current 
+		 * position. A single visible step is about 8.000 to 14.000 in distance depending 
+		 * on the direction its moving in.
+         */
```
IntelOrca, line 1087: The formatting is a bit strange here.

Basssiiie: The formatting was fixed with commit 1cfaf49 right? Or did I miss something?

duncanspumpkin:
> @duncanspumpkin I realise I checked in your code from #13607 here as well. Should I remove it or should it merge fine after your PR has been merged?

Could you rebase now that my code has been merged.

distribution/openrct2.d.ts
IntelOrca: Yes, my browser probably hadn't updated to show the diff with that commit.

Basssiiie:
> Could you rebase now that my code has been merged.

Sure! I'll do that tonight or tomorrow when I have time. :)

src/openrct2/ride/Vehicle.cpp
```diff
@@ -818,6 +818,25 @@ uint16_t Vehicle::GetTrackProgress() const
...
 
+void Vehicle::MoveRelativeDistance(int32_t distance)
+{
+    remaining_distance += distance;
+
+    SetUpdateFlag(VEHICLE_UPDATE_FLAG_SINGLE_CAR_POSITION);
+    UpdateTrackMotion(nullptr);
+    ClearUpdateFlag(VEHICLE_UPDATE_FLAG_SINGLE_CAR_POSITION);
+
+    const Ride* curRide = GetRide();
```
duncanspumpkin, line 829: I'm pretty sure you don't need to do any of this  moveInfo changes. UpdateTrackMotion should take care of it.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7631,7 +7650,7 @@ void Vehicle::UpdateReverserCarBogies()
  */
 bool Vehicle::UpdateMotionCollisionDetection(const CoordsXYZ& loc, uint16_t* otherVehicleIndex)
 {
-    if (HasUpdateFlag(VEHICLE_UPDATE_FLAG_COLLISION_DISABLED))
+    if (HasUpdateFlag(VEHICLE_UPDATE_FLAG_COLLISION_DISABLED) || HasUpdateFlag(VEHICLE_UPDATE_FLAG_SINGLE_CAR_POSITION))
```
duncanspumpkin, line 7653: Any reason you are doing this rather than just setting `VEHICLE_UPDATE_FLAG_COLLISION_DISABLED` as well as `VEHICLE_UPDATE_FLAG_SINGLE_CAR_POSITION`

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -592,6 +607,18 @@ namespace OpenRCT2::Scripting
...
             return result;
         }
+
+        void move(int32_t value)
+        {
+            ThrowIfGameStateNotMutable();
+            auto vehicle = GetVehicle();
+            if (vehicle != nullptr)
+            {
+                vehicle->Invalidate2();
```
duncanspumpkin, line 617: I'm pretty sure `MoveRelativeDistance` will now take care of the invalidating.

src/openrct2/ride/Vehicle.cpp
Basssiiie: I can do that yes, that sounds like a better idea. Will it not interfere with any other code if I clear the `VEHICLE_UPDATE_FLAG_COLLISION_DISABLED`-flag again after `UpdateTrackMotion`? I am not sure when and where this flag is used elsewhere.

src/openrct2/scripting/ScEntity.hpp
Basssiiie: Yes you're right, I'll remove it. :)

src/openrct2/ride/Vehicle.cpp
Basssiiie: Yes you're right, I'll remove it. :)

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: I know it's used for vehicle placement (which is not an issue). Have a look see if it's set for anything else

Basssiiie: The only case I can find is the Giga Coaster cablelift, where it is enabled permanently. I don't think `UpdateMotionCollisionDetection` is called for the cablelift though, so it seems safe to clear it. :)

duncanspumpkin:
This looks reasonable. Assuming it all works fine on your plugin I've no issues with it. @IntelOrca do you have anything to add?

IntelOrca (approved):

Basssiiie:
> This looks reasonable. Assuming it all works fine on your plugin I've no issues with it. @IntelOrca do you have anything to add?

Yes it works perfect in my plugin. :) Only quirk I could find is that modifying a car on a slope sometimes moves other vehicles on the train as well. (Like I mentioned on Discord.) 

Other than that I tested a lot of special rides like Reverser Coaster, Chairlift, Giga Coaster, Lift, Go Karts, Heartline Twister, Log Flume reversers etc. as well and they all work perfectly!

IntelOrca:
Just wondering whether `move` is the best name for the method in both the source code and the plugin API. Maybe it should be `moveBy` or `travel` or `travelBy`.

Basssiiie:
> Just wondering whether `move` is the best name for the method in both the source code and the plugin API. Maybe it should be `moveBy` or `travel` or `travelBy`.

I can change it to `moveBy` if that is preferred. That'd also make it more in line with `[Viewport.moveTo()](https://github.com/OpenRCT2/OpenRCT2/blob/1d536e2aadb7141704ae4eab6a80bc9a906e35f8/distribution/openrct2.d.ts#L1975)`. 

`travel` and `travelBy` sounds a bit more unconventional, but that's just my personal feeling. I'd not be against it if it's the final verdict.

Who wants to make the final choice? 😃

Broxzier:
Just `move` doesn't really explain what the function does, especially since it only takes one argument. `travelBy` would make it more clear that it has to do with the track, though something even more verbose like `moveAlongTrack` would be even better IMO.

IntelOrca:
Yeah, my thoughts are that move is likely to mean, move by a coordinate relative to the map rather than an amount relative to the track. So I think it should just have a different name in case we have another function called `moveBy` or `moveTo` which move it by an x, y, z axis.

Basssiiie:
So `travelBy` is the final decision?

IntelOrca:
@duncanspumpkin any personal suggestions?

duncanspumpkin:
No opinions here

Basssiiie:
The `move` method is now renamed to `travelBy`. :)

IntelOrca:
Okay finally, you just need to increment the plugin API version number and add this to the changelog. Sorry, I should have mentioned that previously.

IntelOrca (approved):

Basssiiie:
Great, no worries! I think it should be properly rebased now? And the plugin API version and changelog have been updated. :)

PR #13604, opened 2020-12-18
Refactor TILE_INSPECTOR_ELEMENT_TYPE to use strong enum #12451

mysterious1223:


Broxzier (approved):

tupaschoal (commented):
@Broxzier should we remove those? A bit weird that they're unused...

mysterious1223:
I agree. I noticed that this enum was unused.

duncanspumpkin:
Its because we use the element type instead of the tile inspector element type. Why have two of the same thing.

Broxzier:
> @Broxzier should we remove those? A bit weird that they're unused...

Ah yes, of course. Unused stuff is better removed.

tupaschoal (changes requested):
Please remove this enum then, @mysterious1223 . Also make sure to mark the PR ready for review when it is.

tupaschoal (approved):

PR #13605, opened 2020-12-18
Fix: Docker build failure

CorySanin:
Sorry about that 😅


CorySanin:
Hmm, that's not right...

CorySanin:
Oh, is it just building what's in the develop branch? That would explain why it passed in my PR but then instantly failed when it was merged. Docker is passing locally for me with this change.

PR #13607, opened 2020-12-18
Add individual vehicle motion update flag

duncanspumpkin:
This can be used in conjunction with #13593 to change the placement of vehicles on a track.

Example use:
```
        veh->remaining_distance += distanceToMove;
        veh->SetUpdateFlag(VEHICLE_UPDATE_FLAG_SINGLE_CAR);
        veh->UpdateTrackMotion(nullptr);
        veh->ClearUpdateFlag(VEHICLE_UPDATE_FLAG_SINGLE_CAR);
```
remaining_distance is internal distance units.  Issue at present is that velocity of the vehicle is also applied. Unsure if that is useful or not. If not can easily fix that by changing how the override of the update code works.

### DIFF
--- a/src/openrct2/ride/Vehicle.h
+++ b/src/openrct2/ride/Vehicle.h
@@ -530,7 +530,9 @@ enum : uint32_t
     VEHICLE_UPDATE_FLAG_USE_INVERTED_SPRITES = (1 << 11), // Used on rides where trains can run for extended periods of time,
                                                           // i.e. the Flying, Lay-down and Multi-dimension RCs.
     VEHICLE_UPDATE_FLAG_12 = (1 << 12),
-    VEHICLE_UPDATE_FLAG_ROTATION_OFF_WILD_MOUSE = (1 << 13) // After passing a rotation toggle track piece this will enable
+    VEHICLE_UPDATE_FLAG_ROTATION_OFF_WILD_MOUSE = (1 << 13), // After passing a rotation toggle track piece this will enable
+    VEHICLE_UPDATE_FLAG_SINGLE_CAR_POSITION = (1 << 14), // OpenRCT2 Flag: Used to override UpdateMotion to move the position of
+                                                         // an individual car on a train
 };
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Vehicle.h
```diff
@@ -530,7 +530,9 @@ enum : uint32_t
     VEHICLE_UPDATE_FLAG_USE_INVERTED_SPRITES = (1 << 11), // Used on rides where trains can run for extended periods of time,
                                                           // i.e. the Flying, Lay-down and Multi-dimension RCs.
     VEHICLE_UPDATE_FLAG_12 = (1 << 12),
-    VEHICLE_UPDATE_FLAG_ROTATION_OFF_WILD_MOUSE = (1 << 13) // After passing a rotation toggle track piece this will enable
+    VEHICLE_UPDATE_FLAG_ROTATION_OFF_WILD_MOUSE = (1 << 13), // After passing a rotation toggle track piece this will enable
+    VEHICLE_UPDATE_FLAG_SINGLE_CAR = (1 << 14), // New flag used to override UpdateMotion to apply only to an individual vehicle
+                                                // on a train
```
tupaschoal, line 535:
```suggestion
    VEHICLE_UPDATE_FLAG_SINGLE_CAR = (1 << 14), // Flag to override UpdateMotion to apply only to an individual vehicle on a train
```

I guess the `new` will just become outdated overtime, so I'd just simplify this.

tupaschoal (approved):

src/openrct2/ride/Vehicle.h
Gymnasiast: Or specify that _we_ added it and didn't inherit it from RCT2.

duncanspumpkin: Changed it to "OpenRCT2 Flag: Used to...."

Gymnasiast (approved):

PR #13610, opened 2020-12-19
Validate all entity indexes on load

duncanspumpkin:
You get a lot of warnings doing this due to not fully clearing structures such as rides when creating.


IntelOrca:
Indexes or Indices, but not Indexs.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1685,6 +1686,167 @@ class S6Importer final : public IParkImporter
...
     }
+
+    void ValidateGuestEntityIndexs(Guest& guest)
+    {
+        if (guest.GuestNextInQueue != SPRITE_INDEX_NULL)
+        {
+            auto* next = TryGetEntity<Guest>(guest.GuestNextInQueue);
+            if (next == nullptr)
+            {
+                log_warning("Next in queue incorrect for guest: %d. Reseting to null", guest.sprite_index);
```
Gymnasiast, line 1697: resetting

PR #13613, opened 2020-12-20
Add "Single Rail Roller Coaster" ride type

X123M3-256:
This PR adds a new ride type "Single Rail Roller Coaster" which is an RMC Raptor style coaster. This ride type supports all slope angles and banked turn pieces, plus half loops, quarter loops, corkscrews, barrel rolls, and the small steep to flat. Large steep to flat and large half loop sprites are included but only accessible when "Show all drawable track pieces" is checked.

raptorpr

Due to the limitations of the ride construction window, the large and small steep to flat cannot be available simultaneously. I decided to enable the small one by default because I found that I used it more frequently in testing, and among vanilla rides the large steep to flat is found only on large coasters (which this isn't). I left the large half loop disabled by default because I felt it was awkwardly large for the ride type, but I am still not completely sure about this as this element is found on the real thing.

A [custom train](https://cdn.discordapp.com/attachments/656575075515891745/790028915615858698/RMCT2.DAT) has been made for this by @spacek531, which can be used for testing (this is not part of this PR).


data/language/en-GB.txt
```diff
@@ -175,6 +176,7 @@ STR_0599    :A compact roller coaster with individual cars and smooth twisting d
 STR_0600    :Powered mine trains career along a smooth and twisted track layout
 STR_0602    :Roller coaster trains are accelerated out of the station by linear induction motors to speed through twisting inversions
 STR_0603    :A wooden style roller coaster with a steel track, allowing for steep drops and inversions.
+STR_0603    :Riders ride single file on a narrow monorail track, as they race through tight inversions and direction changes.
```
Gymnasiast, line 179:
```suggestion
STR_0604    :Riders ride single file on a narrow monorail track, as they race through tight inversions and direction changes.
```

X123M3-256: This should be fixed.

Gymnasiast:
You will need to add an (empty) definition of `ride_ratings_calculate_single_rail_roller_coaster()` to Compat.cpp in order to make TestPaint compile again.

X123M3-256:
@Gymnasiast Where is Compat.cpp?

Gymnasiast:
In `test/testpaint`.

X123M3-256:
@Gymnasiast OK, I've done that

test/testpaint/Compat.cpp
```diff
@@ -861,6 +861,10 @@ void ride_ratings_calculate_hybrid_coaster([[maybe_unused]] Ride* ride)
 {
 }
 
+void ride_ratings_calculate_single rail_roller_coaster([[maybe_unused]] Ride* ride)
```
Gymnasiast, line 864:
```suggestion
void ride_ratings_calculate_single_rail_roller_coaster([[maybe_unused]] Ride* ride)
```

X123M3-256: @Gymnasiast These two lines look identical, what needs changing exactly?

YJSoft: @X123M3-256 It should be `single_rail`, not `single rail` (missing `_`)

X123M3-256: Oh yeah, not sure how I missed that.

duncanspumpkin:
I dont see an entry in `RideObject::ParseRideType` we will need this so that Spacek's object is loaded correctly. Suggest calling it `single_rail_rc`

X123M3-256:
@duncanspumpkin I think I've fixed that now, but I don't have the JSON version of SpaceK's train so I haven't tested it.

src/openrct2/localisation/StringIds.h
```diff
@@ -201,6 +202,7 @@ enum
 
     STR_RIDE_DESCRIPTION_LIM_LAUNCHED_ROLLER_COASTER = 602,
     STR_RIDE_DESCRIPTION_HYBRID_COASTER = 603,
+    STR_RIDE_DESCRIPTION_SINGLE_RAIL_ROLLER_COASTER = 603,
```
duncanspumpkin, line 205: 604

src/openrct2/ride/coaster/meta/SingleRailRollerCoaster.h
```diff
@@ -0,0 +1,54 @@
...
+#include "../../RideData.h"
+#include "../../Track.h"
+
+// clang-format off
+constexpr const RideTypeDescriptor SingleRailRollerCoasterRTD =
+{
+    SET_FIELD(AlternateType, RIDE_TYPE_NULL),
+    SET_FIELD(Category, RIDE_CATEGORY_ROLLERCOASTER),
+    SET_FIELD(EnabledTrackPieces, (1ULL << TRACK_FLAT) | (1ULL << TRACK_STRAIGHT) | (1ULL << TRACK_STATION_END) | (1ULL << TRACK_LIFT_HILL) | (1ULL << TRACK_LIFT_HILL_STEEP) | (1ULL << TRACK_FLAT_ROLL_BANKING) | (1ULL << TRACK_SLOPE) |(1ULL << TRACK_SLOPE_LONG) | (1ULL << TRACK_SLOPE_STEEP) | (1ULL << TRACK_SLOPE_CURVE) | (1ULL << TRACK_SLOPE_CURVE_STEEP) | (1ULL << TRACK_S_BEND) | (1ULL << TRACK_CURVE_SMALL) | (1ULL << TRACK_CURVE) | (1ULL << TRACK_CORKSCREW) | (1ULL << TRACK_HELIX_SMALL) | (1ULL << TRACK_BRAKES) | (1ULL << TRACK_ON_RIDE_PHOTO) | (1ULL << TRACK_SLOPE_VERTICAL) | (1ULL << TRACK_BARREL_ROLL) | (1ULL << TRACK_SLOPE_CURVE_BANKED) | (1ULL << TRACK_BLOCK_BRAKES) | (1ULL << TRACK_SLOPE_ROLL_BANKING) | (1ULL << TRACK_CURVE_VERTICAL) | (1ULL << TRACK_QUARTER_LOOP)| (1ULL << TRACK_HALF_LOOP)),
+    SET_FIELD(ExtraTrackPieces,  (1ULL << TRACK_SLOPE_STEEP_LONG) | (1ULL << TRACK_HALF_LOOP_LARGE)),
```
duncanspumpkin, line 21: The extra pieces that were removed such as the booster should they be added here or was that correcting a mistake before?

src/openrct2/ride/coaster/meta/SingleRailRollerCoaster.h
```diff
@@ -0,0 +1,54 @@
...
+    SET_FIELD(DefaultMusic, MUSIC_STYLE_ROCK),
+    SET_FIELD(PhotoItem, ShopItem::Photo),
+    SET_FIELD(BonusValue, 80),
+    SET_FIELD(ColourPresets, TRACK_COLOUR_PRESETS(
+        { COLOUR_LIGHT_ORANGE, COLOUR_LIGHT_ORANGE, COLOUR_WHITE },
+        { COLOUR_YELLOW, COLOUR_YELLOW, COLOUR_BLACK },
+        { COLOUR_LIGHT_BLUE, COLOUR_LIGHT_BLUE, COLOUR_DARK_GREEN }
+    )),
+  SET_FIELD(ColourPreview, { SPR_RIDE_DESIGN_PREVIEW_SINGLE_RAIL_ROLLER_COASTER_TRACK, SPR_RIDE_DESIGN_PREVIEW_SINGLE_RAIL_ROLLER_COASTER_SUPPORTS }),
+  SET_FIELD(ColourKey, RideColourKey::Ride),
```
Gymnasiast, line 52:
```suggestion
    SET_FIELD(ColourPreview, { SPR_RIDE_DESIGN_PREVIEW_SINGLE_RAIL_ROLLER_COASTER_TRACK, SPR_RIDE_DESIGN_PREVIEW_SINGLE_RAIL_ROLLER_COASTER_SUPPORTS }),
    SET_FIELD(ColourKey, RideColourKey::Ride),
```

duncanspumpkin:
Is square tunnels what you would want for this ride. I would have thought circular tunnels would look better. I'm not the biggest fan of the code not using the rotated functions correctly but as its just based on another ride I'm not going to request that it is changed. Code in general looks fine other than that. I've not tested the coaster.

src/openrct2/ride/coaster/meta/SingleRailRollerCoaster.h
X123M3-256: I removed those pieces because they weren't meant to be enabled - there is no sprite for them.

X123M3-256:
@duncanspumpkin I noticed the square tunnels and I had meant to change them to the circular ones, so I will have a look at that.

duncanspumpkin:
have these png's been compressed?

X123M3-256:
@duncanspumpkin No, shall I do that now? I remember there was some discussion last time about the best way to do it.

Gymnasiast:
It looks like `oxipng` had the best results that still produced images our code processed correctly.

X123M3-256:
@Gymnasiast I tried using that, but it messed up all the sprites so I've reverted it back to the unoptimized version. What's happening is that the optimizer tries to save space by reducing the palette to just the colors which are actually used, but this changes the palette indices. I am guessing this didn't affect the vehicle files because they don't have `"palette":"keep"` set in the JSON.

I can remove this for most of the sprites, but it is mandatory for lift sprites as the chain lift animation would otherwise be lost. Is there a way to make this tool preserve the palette indices for those sprites?

notright

Gymnasiast:
Hm, that's annoying. Can you use `optipng` on the chain lift sprites and use `oxipng` on the rest?

X123M3-256:
@Gymnasiast I certainly could do that, but I'm wondering, since only the palette indices are actually needed is there a way to drop the palette from the PNG file entirely?

Gymnasiast:
There isn't, no. Using palette indices requires an actual palette to be present.

X123M3-256:
@Gymnasiast Would you be able to change the palette used for importing so that the chain lift colors (index 1, 2, and 3) have distinct colors (they're all black at the moment)? Then it wouldn't be necessary to preserve the palette indices for loading those sprites.

Gymnasiast:
I could, yeah. Preferably, I'd take the same palette as the Object Editor and/or Trigger's tools if at all possible.

X123M3-256:
@Gymnasiast I asked on NE and they said the object editor doesn't support those colors (I myself haven't used it since it's a Windows program). I think any color not already in the palette would work - you could use (1,1,1), (2,2,2), (3,3,3) etc. If every palette index had a unique color, it wouldn't matter if the actual palette indices are preserved.

I can go through and optimize only the sprites that don't use the chain lift pixels, but I think it's a bit awkward to have these sprites treated as a special case, it would be nice if the same optimization could be applied to all of them.

Gymnasiast:
OK, I will do that. Is it just colours 1, 2 ,3 (counting from 0), or is it more than that?

Gymnasiast:
https://github.com/OpenRCT2/OpenRCT2/pull/13659

X123M3-256:
@Gymnasiast The chain lift colors are indices 1,2, and 3.

X123M3-256:
@Gymnasiast I've compressed the PNGs now, it seems to work OK.

duncanspumpkin:
Excellent. Are there any other changes you are wanting to make before merging

X123M3-256:
@duncanspumpkin I just pushed a few more tweaks to the sprites but as far as I'm concerned, this is ready to merge (apart from the network version).

Gymnasiast (approved):
Code looks good to me, and appears to work correctly.

duncanspumpkin:
I've no issues with this either.

duncanspumpkin (approved):

Gymnasiast:
Right, could you amend the changelog and do all rebases you want to do? I'll then do one last check and merge it.

X123M3-256:
@Gymnasiast What do you want added to the changelog (and where is the changelog)?

Issue #13615, opened 2020-12-20
Some Game Actions have incomplete implementations


csunday95:
**OS:** all
**Version:** 0.3.2
**Commit/Build:** latest develop - bcf0e15c97a70718c842c0e0129e7a58d20a4478

Some of the `GameAction` implementations are incomplete. For example, the `SetStaffNameAction` takes two arguments, `_spriteIndex` and `_name`, but the current code does not have an override for the base class `AcceptParameters` method. As a result, this action will access default-initialized values for its arguments and this will always result in an error when executed from the plugin interface. 

**Steps to reproduce:**
1. create a very simple plugin, such as:
```js
var setStaffName = function(newName) {
    for (var i = 0; i < map.numEntities; i++) {
        var entity = map.getEntity(i);
        if (entity && entity.type === 'peep' && entity.peepType == 'staff') {
            context.executeAction("staffsetname", {
                spriteIndex: entity.id,
                name: newName
            }, function (result) {
                console.log(result);
            });
            break;
        }
    }
}

var main = function() {
    // Add a menu item under the map icon on the top toolbar
    ui.registerMenuItem("Test Set Staff Name", function() {
        setStaffName("Test Name");
    });
};

registerPlugin({
    name: 'Example',
    version: '0.1',
    authors: ['OpenRCT2'],
    type: 'remote',
    main: main
});
```

2. Open a park with the plugin enabled 
3. Attempt to execute the new `Test Set Staff Name` menu entry in the map menu
4. observe the console output of
```
{ error: 1,
  errorTitle: 'Can't name staff member…',
  errorMessage: '',
  cost: 0,
  expenditureType: 14 }
  ```

I have a pull request I'm going to submit presently that has a fix specifically for staffsetname, but I was planning to quickly go through and see if any of the other GameActions are missing implementation.

sjoerddebruin:
This got fixed by #18826, right?

PR #13616, opened 2020-12-20
inserted missing implementation for some GameAction subclasses

csunday95:
Creating pull request for fixing issue #13615

So far only has SetStaffNameAction implementation, but am planning to browse through the actions and see if any others need the same additions


src/openrct2/actions/StaffSetNameAction.cpp
```diff
@@ -26,6 +26,12 @@ StaffSetNameAction::StaffSetNameAction(uint16_t spriteIndex, const std::string&
 {
 }
 
+void StaffSetNameAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("spriteIndex", _spriteIndex);
```
IntelOrca, line 31: I think just `id` would be better than `spriteIndex`. We no longer refer to entities as sprites.

duncanspumpkin: In `GuestSetNameAction` we used `peep`

csunday95: I just selected spriteIndex to line up with the member name, would it be preferred to have a more human readable name like "peep" or "peepId" for the outward facing API?

duncanspumpkin: I think `peep` would be a best. There are at least 2 other actions that use that.

src/openrct2/actions/FootpathPlaceAction.cpp
```diff
@@ -277,8 +277,9 @@ GameActions::Result::Ptr FootpathPlaceAction::ElementInsertQuery(GameActions::Re
 GameActions::Result::Ptr FootpathPlaceAction::ElementInsertExecute(GameActions::Result::Ptr res) const
 {
     bool entrancePath = false, entranceIsSamePath = false;
+    uint32_t flags = GetFlags();
 
-    if (!(GetFlags() & (GAME_COMMAND_FLAG_ALLOW_DURING_PAUSED | GAME_COMMAND_FLAG_GHOST)))
+    if (!(flags & (GAME_COMMAND_FLAG_ALLOW_DURING_PAUSED | GAME_COMMAND_FLAG_GHOST)))
```
duncanspumpkin, line 282: don't make unrelated changes in this PR save it for a different PR.

csunday95: wait... how did this end up here? dang I think I must have been working on the wrong local branch when I was investigating something. I'll revert this on my next commit to this PR

IntelOrca (approved):

csunday95:
ah shoot I seem to have messed up this PR while trying to rebase something else; not sure what to do in this situation since I committed on top of the rebase and pushed. Might just be best to close this PR and redo it from scratch? redoing these changes is pretty trivial with PR this as a reference

duncanspumpkin:
Just do a rebase of develop. It will remove all of the merge related commits that you have added and keep only your changes.

PR #13618, opened 2020-12-21
Fix #13611: Zero under decimal is not displayed

### LINKED ISSUE #13611: Zero under decimal is not displayed
**OS:** Windows 10
**Version:** 0.3.2
**Commit/Build:** 150f243

Zero under decimal is not displayed in Ratings, Money and so on.
I've also noticed that Ride's profit is displayed as ``$0.`` if it has no profit at all.

However I can't reproduce it always.
It does not matter whether it is multiplayer or not.

**Screenshots / Video:**
image

csunday95:
The existing code does not properly handle the formatting conversion case for fixed point decimal values. If a fixed point integer representation has fewer significant figures than the required fixed point precision, no zeros are placed between the decimal place and the first sig fig. This causes ride intensity, nausea etc. to display incorrectly when very small values, e.g. a nausea value of 0.05 will be displayed as 0.5.

### DIFF
--- a/src/openrct2/localisation/Formatting.cpp
+++ b/src/openrct2/localisation/Formatting.cpp
@@ -341,6 +341,11 @@ namespace OpenRCT2
                 buffer[i++] = static_cast<char>('0' + (num % 10));
                 num /= 10;
             }
+            // handle case where value has fewer sig figs than required decimal places
+            while (num == 0 && i < TDecimalPlace && i < sizeof(buffer))
+            {
+                buffer[i++] = static_cast<char>('0');
+            }
 
             auto decSep = GetDecimalSeparator();
             AppendSeparator(buffer, i, decSep);
--- a/src/openrct2/localisation/Formatting.cpp
+++ b/src/openrct2/localisation/Formatting.cpp
@@ -344,7 +344,7 @@ namespace OpenRCT2
             // handle case where value has fewer sig figs than required decimal places
             while (num == 0 && i < TDecimalPlace && i < sizeof(buffer))
             {
-                buffer[i++] = static_cast<char>('0');
+                buffer[i++] = '0';
             }
 
             auto decSep = GetDecimalSeparator();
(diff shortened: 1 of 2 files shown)


IntelOrca:
Are you able to write a unit test for this?

csunday95:
yes, but I was trying to figure out where the format for it is specified. I assume I can just add a TEST_F method to the tests/FormattingTests.cpp file?

src/openrct2/localisation/Formatting.cpp
```diff
@@ -341,6 +341,11 @@ namespace OpenRCT2
                 buffer[i++] = static_cast<char>('0' + (num % 10));
                 num /= 10;
             }
+            // handle case where value has fewer sig figs than required decimal places
+            while (num == 0 && i < TDecimalPlace && i < sizeof(buffer))
+            {
+                buffer[i++] = static_cast<char>('0');
```
duncanspumpkin, line 347: Surely the static cast is not required.

csunday95: yes certainly not required, just put it there for consistency/readability with the prior line that makes use of '0' char literal as an offset: `static_cast<char>('0' + (num % 10))`. Removing the static cast probably won't have any effect on code generation, but if you'd prefer it be removed for style I can.

IntelOrca: I agree, just remove it if it isn't required.

csunday95: I have removed the explicit cast; I made my commit summary message slightly too long though and got caught by the CI linter; is there anything I need to do about that (e.g. commit --amend) or is it more of a reminder?

duncanspumpkin: Just amend the commit the Mac os ci has failed as well. Might be unrelated though.

csunday95: Mac CI mysteriously had a connection timeout during upload phase from what I could tell, might just be a random glitch with CI

duncanspumpkin: Yeah Mac CI always seems to have issues.

IntelOrca (approved):

PR #13632, opened 2020-12-23
Some enum class conversions

tupaschoal:


src/openrct2/Editor.h
```diff
@@ -31,16 +31,17 @@ namespace Editor
...
-    EDITOR_STEP_OBJECTIVE_SELECTION,    // 4
-    EDITOR_STEP_SAVE_SCENARIO,          // 5
-    EDITOR_STEP_ROLLERCOASTER_DESIGNER, // 6
-    EDITOR_STEP_TRACK_DESIGNS_MANAGER   // 7
+    ObjectSelection,       // 0
+    LandscapeEditor,       // 1
+    InventionsListSetUp,   // 2
+    OptionsSelection,      // 3
+    ObjectiveSelection,    // 4
+    SaveScenario,          // 5
```
tupaschoal, line 41: Weirdly this one is never directly referenced

src/openrct2/scenario/Scenario.h
```diff
@@ -27,6 +27,8 @@
 
 using random_engine_t = Random::Rct2::Engine;
 
+enum class Rct2EditorStep : uint8_t;
```
duncanspumpkin, line 30: RCT2

src/openrct2-ui/input/KeyboardShortcut.cpp
```diff
@@ -98,7 +98,7 @@ static void ShortcutCloseAllFloatingWindow()
 {
     if (!(gScreenFlags & SCREEN_FLAGS_SCENARIO_EDITOR))
         window_close_all();
-    else if (gS6Info.editor_step == EDITOR_STEP_LANDSCAPE_EDITOR)
+    else if (gS6Info.editor_step == RCT2EditorStep::LandscapeEditor)
```
Gymnasiast, line 101: Why call this `RCT2EditorStep`? We only put this prefix if we use a different approach than RCT2.

tupaschoal: I just reused what was already there. What do you suggest, just `EditorStep`?

Gymnasiast: Yup, just `EditorStep`.

tupaschoal:
Done, I've also solved the merge conflict

Gymnasiast (changes requested):
Could you change `EditStep` to `EditorStep`?

src/openrct2/ride/Ride.cpp
```diff
@@ -1984,7 +1984,7 @@ void Ride::UpdateAll()
     // Remove all rides if scenario editor
     if (gScreenFlags & SCREEN_FLAGS_SCENARIO_EDITOR)
     {
-        if (gS6Info.editor_step <= EDITOR_STEP_INVENTIONS_LIST_SET_UP)
+        if (EnumValue(gS6Info.editor_step) <= EnumValue(EditStep::InventionsListSetUp))
```
Gymnasiast, line 1987: Can you create an issue to change the type of `gS6Info.editor_step`?

duncanspumpkin: Shouldn't this be a switch with all of the relevant steps.

tupaschoal: I can convert it to a switch no problem.

O have already changed the type to the enuk one, haven't I? Not sure I get you Gym

src/openrct2/scenario/Scenario.h
```diff
@@ -51,7 +53,7 @@ assert_struct_size(rct_s6_header, 0x20);
  */
 struct rct_s6_info
 {
-    uint8_t editor_step;
+    EditorStep editor_step;
```
tupaschoal, line 56: I had already changed this @Gymnasiast , is that what you meant?

src/openrct2/ride/Ride.cpp
Gymnasiast: I was thrown off by the use of `EnumValue`, but I now see why you had to use that.

Gymnasiast (approved):

PR #13633, opened 2020-12-23
Paint Setup refactor

duncanspumpkin:
Whilst implementing the same function for OpenLoco I noticed an opertunity to simplify the Paint Generate code. Then of course there was poor performance so had to introduce the template version to simplify the execution. Comments welcome to if this is easier to understand or not.


tupaschoal (approved):
Hmm, I guess the end result is indeed a bit clearer to follow. I look at each commit and it looks ok

duncanspumpkin:
I made a slight mistake in the middle commits but it was corrected at the end. I think there is two rotations of the coordinates.

tupaschoal:
Yea, I noticed that and the casts coming back

PR #13634, opened 2020-12-23
Add ability to choose a random color of selling shop items

### LINKED ISSUE #1438: Random balloon/hat color
Currently all balloons/hats are the same color, it would be nice to add an option to give the customer a random balloon/hat color.

8street:
I'd added ability to choose a random color of selling shop items: tshirts, hats, balloons, umbrellas. Save and load works correctly. Old saves also works.

random

This PR should Close #1438


duncanspumpkin:
I'm not a fan of adding more additional special cases to the sv6 format.

8street:
Сan you suggest your own solution, where to save the state of this option?

duncanspumpkin:
@8street i know its been a while but would you like to retarget this feature at the new save format branch. I've no issues with it on the NSF.https://github.com/OpenRCT2/OpenRCT2/pull/10664

8street:
Thank you for reminding. I will try do it next week or another week later.

src/openrct2/ride/Ride.h
```diff
@@ -375,6 +375,7 @@ struct Ride
     uint8_t num_circuits;
     CoordsXYZ CableLiftLoc;
     uint16_t cable_lift;
+    uint8_t selling_item_color_is_random;
```
duncanspumpkin, line 378: we use british english in this code base and also new fields should be CamelCase
`SellingItemColourIsRandom`

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4238,6 +4242,13 @@ static void window_ride_colour_mouseup(rct_window* w, rct_widgetindex widgetInde
         case WIDX_PAINT_INDIVIDUAL_AREA:
             tool_set(w, WIDX_PAINT_INDIVIDUAL_AREA, Tool::PaintDown);
             break;
+        case WIDX_SELL_ITEM_RANDOM_COLOUR_CHECKBOX:
+            auto ride = get_ride(w->number);
```
duncanspumpkin, line 4246: To make this work in multiplayer the change will have to go through the game action framework. I think you should be able to extend RideSetAppearanceAction to add the random setting.

8street:
For some reason, the game cannot load files after updating to `new-save-format` branch. Therefore, I cannot test the changes. The debug console outputs a lot of lines:

> Unable to load park: NinSFOT.sv6
> Failed to load: "NinSFOT.sv6" for the title sequence.
> [rct2.terrain_surface.grass] Object not found.
> [rct2.terrain_surface.sand] Object not found.
> [rct2.terrain_surface.dirt] Object not found.
> [rct2.terrain_surface.rock] Object not found.
> [rct2.terrain_surface.martian] Object not found.
> [rct2.terrain_surface.chequerboard] Object not found.
> [rct2.terrain_surface.grass_clumps] Object not found.
> ...

Also:

> Unable to load park: E:\Steam\steamapps\common\RollerCoaster Tycoon Deluxe\Scenarios\sc18.sc4
> Failed to load: "Ivory Towers" for the title sequence.
> Unable to load any parks from rct1.
> ERROR[E:\git\OpenRCT2\src\openrct2\rct1\S4Importer.cpp:615 (S4Importer::AddAvailableEntriesFromSceneryGroups)]: Cannot find object rct2.scenery_small.roof5
> ERROR[E:\git\OpenRCT2\src\openrct2\rct1\S4Importer.cpp:615 (S4Importer::AddAvailableEntriesFromSceneryGroups)]: Cannot find object rct2.scenery_small.roof7
> ERROR[E:\git\OpenRCT2\src\openrct2\rct1\S4Importer.cpp:615 (S4Importer::AddAvailableEntriesFromSceneryGroups)]: Cannot find object rct2.scenery_small.brbase
> ERROR[E:\git\OpenRCT2\src\openrct2\rct1\S4Importer.cpp:615 (S4Importer::AddAvailableEntriesFromSceneryGroups)]: Cannot find object rct2.scenery_small.roof1

Also:

> ...
>                : 'E:\git\OpenRCT2\bin\data\object\rct2ww\scenery_wall\rct2ww.scenery_wall.wwind03.json'
> Object conflict: 'E:\git\OpenRCT2\bin\data\object\rct2ww\scenery_wall\rct2.ww.wwind04.json'
>                : 'E:\git\OpenRCT2\bin\data\object\rct2ww\scenery_wall\rct2ww.scenery_wall.wwind04.json'
> Object conflict: 'E:\git\OpenRCT2\bin\data\object\rct2ww\scenery_wall\rct2.ww.wwind05.json'
>                : 'E:\git\OpenRCT2\bin\data\object\rct2ww\scenery_wall\rct2ww.scenery_wall.wwind05.json'
> Object conflict: 'E:\git\OpenRCT2\bin\data\object\rct2ww\scenery_wall\rct2.ww.wwind06.json'
>                : 'E:\git\OpenRCT2\bin\data\object\rct2ww\scenery_wall\rct2ww.scenery_wall.wwind06.json'
> 2087 object conflicts found.

The file paths are correct. The `develop` branch is working fine. Can someone help me?

duncanspumpkin:
Yes its a bit of a pain switching between develop and new-save-format due to the changes in the objects. You need to redownload your objects and delete the old ones. (I find the easiest way to test is to just download the portable build from the github actions saves having to delete things)

duncanspumpkin:
You'll also need to save your variable into the park file. i.e just modify here to add in the new member variable. https://github.com/OpenRCT2/OpenRCT2/blob/622c9d34d46535a953d98752154bbf8e80dd2174/src/openrct2/ParkFile.cpp#L1166

8street:
> I find the easiest way to test is to just download the portable build from the github actions saves having to delete things
I'd downloaded portable files and put them in OpenRCT2\bin folder. Objects are not loaded anyway. I see a black instead of a park. But main menu is visible. How can I delete objects manually? I have not found any menu that is responsible for objects.

8street:
Can be merged I guess.

src/openrct2/actions/RideSetAppearanceAction.h
```diff
@@ -21,7 +21,8 @@ enum class RideSetAppearanceType : uint8_t
     VehicleColourTrim,
     VehicleColourTernary,
     VehicleColourScheme,
-    EntranceStyle
+    EntranceStyle,
+    ToggleSellingItemColourRandomFlag
```
duncanspumpkin, line 25: Just looking at how this compares to the other game actions and the vast majority of boolean flags like this don't have a toggle but just a set function. This would be slightly more versatile for things like plugins that might not care what the current state the flag is in but wants to always set it to on or off.
Could you change it so this is setting the flag state based on the action _value

duncanspumpkin:
This would normally break replays but as this is NSF we don't have replays recorded. So no change required for that. Will require a network version bump though. Increment the number in NetworkBase.cpp

src/openrct2/actions/RideSetAppearanceAction.h
8street: Do you mean to do like this?

```C++
auto ride = get_ride(w->number);
if (ride != nullptr)
{
    auto rideSetAppearanceAction = RideSetAppearanceAction(w->number, RideSetAppearanceType::SellingItemColourIsRandom, !ride->SellingItemColourIsRandom, 0);
    GameActions::Execute(&rideSetAppearanceAction);
}
```

duncanspumpkin: Yup that looks great.

duncanspumpkin (approved):
Still need to do some testing of this but code is fine

distribution/changelog.txt
```diff
@@ -51,6 +51,7 @@
 - Feature: [#13593] [Plugin] Add ability to read and change the position of ride vehicles.
 - Feature: [#13613] Add single-rail roller coaster (Rocky Mountain Construction Raptor).
 - Feature: [#13614] Add terrain surfaces from RollerCoaster Tycoon 1.
+- Feature: [#13634] Add ability to choose a random color of the selling shop items.
```
Gymnasiast, line 54:
```suggestion
- Feature: [#13634] Add ability to sell merchandise in random colours.
```

src/openrct2/ParkFile.cpp
```diff
@@ -1164,6 +1164,8 @@ namespace OpenRCT2
                     cs.ReadWrite(ride.music);
                     cs.ReadWrite(ride.music_tune_id);
                     cs.ReadWrite(ride.music_position);
+
```
Gymnasiast, line 1167:
```suggestion
```

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4812,7 +4844,9 @@ static void window_ride_colour_paint(rct_window* w, rct_drawpixelinfo* dpi)
             + ScreenCoordsXY{ (widget->left + widget->right) / 2 - 8, (widget->bottom + widget->top) / 2 - 6 };
 
         ShopItem shopItem = rideEntry->shop_item[1] == ShopItem::None ? rideEntry->shop_item[0] : rideEntry->shop_item[1];
-        gfx_draw_sprite(dpi, ImageId(GetShopItemDescriptor(shopItem).Image, ride->track_colour[0].main), screenCoords);
+        uint8_t spriteColor = ride->SellingItemColourIsRandom ? scenario_rand_max(COLOUR_COUNT - 1)
```
Gymnasiast, line 4847: spriteColo**u**r

Gymnasiast (changes requested):
Apart from the line notes, there are some things I immediately noticed:
1. It allows setting a random colour for ice creams, which of course makes no sense. The best way to solve it, I think, is by adding a shop item flag (IS_RECOLOURABLE), set it for balloons/umbrellas/t-shirts/hats and only show the toggle if that flag is set.
2. You reused a "random" string that was completely unrelated. While it just _happens_ to be the same in English, this does not apply for other languages. Please just introduce a new string "Random colour" and use that.
3. The colour dropdown is not disabled when this option is checked, which is confusing
4. When the option is selected, the shop item preview goes berserk, cycling through hundreds of colours each second. Cycling through them might be nice, but obviously not at this rate - taking the date in days and modulo’ing it by the number of colours might be a good alternative.

8street:
> 3. The colour dropdown is not disabled when this option is checked, which is confusing

This color dropdown changes а store color. 
I'll see what I can do on your other comments later.

src/openrct2/ParkFile.cpp
8street: Are you sure? The `ride.music` block has already been separately highlighted with a new line a little higher.

8street:
I added the "IS_RECOLOURABLE" flag as requested. Not really a good solution, but it works. Perhaps someone will suggest a better solution for polling all shop items in the store for this flag.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4601,6 +4615,33 @@ static void window_ride_colour_invalidate(rct_window* w)
         window_ride_colour_widgets[WIDX_TRACK_ADDITIONAL_COLOUR].type = WindowWidgetType::Empty;
     }
 
+    // Selling item random colour checkbox
+    bool randomColourCheckboxIsVisible = false;
+    for (int item_index = 0; item_index < NUM_SHOP_ITEMS_PER_RIDE; item_index++)
+    {
+        const ShopItem *currItem = &rideEntry->shop_item[item_index];
```
Gymnasiast, line 4622: This looks a bit odd. You are creating a pointer, but on line 4625, you are referencing its value. It also doesn’t seem a pointer is necessary here.

I think you should be able to replace this with
```suggestion
        const ShopItem& currItem = rideEntry->shop_item[item_index];
```

and removing the asterisk on line 4625.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4601,6 +4615,33 @@ static void window_ride_colour_invalidate(rct_window* w)
...
+        if (currItem != nullptr)
+        {
+            randomColourCheckboxIsVisible |= GetShopItemDescriptor(*currItem).IsRecolourable();
+        }
```
Gymnasiast, line 4626: This looks a bit odd. You are creating a pointer, but on line 4625, you are referencing its value. It also doesn’t seem a pointer is necessary here - it’s a plain integer.

I think you should be able to replace this with
```suggestion
        const ShopItem& currItem = rideEntry->shop_item[item_index];
```

```suggestion
        const ShopItem currentItem = rideEntry->shop_item[item_index];
        if (currentItem != ShopItem::None)
        {
            randomColourCheckboxIsVisible |= GetShopItemDescriptor(currentItem).IsRecolourable();
        }
```

Gymnasiast:
> I added the "IS_RECOLOURABLE" flag as requested. Not really a good solution, but it works. Perhaps someone will suggest a better solution for polling all shop items in the store for this flag.

I’m curious, why do you think it’s not a good solution? We will most likely have more stalls in the future where the stall itself can be recoloured but the item cannot, so it seems to me that simply adding a flag is the cleanest way to go.

src/openrct2-ui/windows/Ride.cpp
8street: > pointer is necessary here - it’s a plain integer

Oh, I missed this moment. Sorry.

8street:
> I’m curious, why do you think it’s not a good solution?

Adding a flag is not bad in itself. But a best solution here would be like `GetRideDescriptor(ride).HasRecolourableItems()` method or something like that. Inside this method, we would iterate over all sold items. I'll try to implement this if you don't mind.

duncanspumpkin:
@8street we eventually want to make ride type descriptors into objects that can be generated by users. Using flags for this is generally the way we would handle things like this.

src/openrct2/ride/Ride.cpp
```diff
@@ -7372,6 +7372,25 @@ void Ride::IncreaseNumShelteredSections()
...
+    const auto rideEntry = GetRideEntry();
+    if (rideEntry == nullptr)
+        return false;
+
+    for (int itemIndex = 0; itemIndex < NUM_SHOP_ITEMS_PER_RIDE; itemIndex++)
+    {
+        const ShopItem currentItem = rideEntry->shop_item[itemIndex];
+        if (currentItem != ShopItem::None)
+        {
+            HasRecolourableShopItems |= GetShopItemDescriptor(currentItem).IsRecolourable();
```
Gymnasiast, line 7388: You can simply return true here if the item is recolourable and return false on line 7391. Might save a few cycles if shops have multiple items to sell.

Gymnasiast (changes requested):
Almost there. Only two things to address AFAIC:
1. The line note.
2. Could you slow down the rate of cycling through preview colours just a little bit more? (I just want to avoid it being too distracting/annoying)

8street:
> Could you slow down the rate of cycling through preview colours just a little bit more? (I just want to avoid it being too distracting/annoying)

What number in `!(gCurrentRealTimeTicks & 0xf)` should I put? I don’t know which number is not annoying, as it didn’t distracting/annoying me at all.

duncanspumpkin:
Go for something like every 1 second which would be about 40 ticks maybe 0x3F (63 ticks)

8street:
It seems to be ready. But because of the long commit name, one of CI tests fails. Can a maintainer rebase my PR on the merge in one commit?

Gymnasiast (approved):

Gymnasiast (approved):

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4778,7 +4808,24 @@ static void WindowRideColourPaint(rct_window* w, rct_drawpixelinfo* dpi)
                               (trackPreviewWidget.bottom + trackPreviewWidget.top) / 2 - 6 };
 
         ShopItem shopItem = rideEntry->shop_item[1] == ShopItem::None ? rideEntry->shop_item[0] : rideEntry->shop_item[1];
-        gfx_draw_sprite(dpi, ImageId(GetShopItemDescriptor(shopItem).Image, ride->track_colour[0].main), screenCoords);
+        static colour_t spriteColour = COLOUR_BLACK;
```
duncanspumpkin, line 4811: If you have two shops windows open they will be on the same colour each refresh. I assume this isn't an issue but just bringing it up.

duncanspumpkin (approved):

src/openrct2-ui/windows/Ride.cpp
Gymnasiast: Tweaked it a little more, should now be good.

PR #13635, opened 2020-12-23
Fix #13549: Ride name is truncated/resetted in Korean

### LINKED ISSUE #13549: Ride name is truncated/resetted in Korean
**OS:** Windows 10
**Version:** 0.3.2
**Commit/Build:** f4bd903

Ride name is truncated or resetted if I named it in Korean.
This issue occurs in not only singleplayer but also multiplayer.

**Steps to reproduce:**
1. Name a ride, such as Merry-go-round, in Korean, such as ``가나다``
2. Save the game, and reload it
3. Its name is truncated or resetted, such as ``회전목마 0``
4. This issue occurs identically in multiplayer when I rejoin instead of reloading.

**Screenshots / Video:**
I named a Merry-go-round as ``가나다``
image

After when I reload the map, its name became ``회전목마 0`` (means "Merry-go-round 0")
image

**In multiplayer game**
image
The red underlined names are weired ones.
``기`` is originally ``기가 코스터``(means "Giga coaster")
``미로 1에서 나`` is originally ``미로 1에서 나가고 싶어``(means "I want to get out of Maze 1")
``보라색 기`` is originally ``보라색 기가 코스터``(means "purple Giga coaster")

IMO, Rides before updates (maybe new-format-string updates?) seems okay.

IntelOrca:
Handle multi-byte RCT2 strings when converting to string_view.

### DIFF
--- a/src/openrct2/rct2/RCT2.h
+++ b/src/openrct2/rct2/RCT2.h
@@ -779,3 +779,9 @@ std::vector<uint8_t> DecryptSea(const fs::path& path);
 ObjectEntryIndex RCT2RideTypeToOpenRCT2RideType(uint8_t rct2RideType, const rct_ride_entry* rideEntry);
 bool RCT2RideTypeNeedsConversion(uint8_t rct2RideType);
 uint8_t OpenRCT2RideTypeToRCT2RideType(ObjectEntryIndex openrct2Type);
+
+/**
+ * Iterates an RCT2 string buffer and returns the length of the string in bytes.
+ * Handles single and multi-byte strings.
+ */
+size_t GetRCT2StringBufferLen(const char* buffer, size_t maxBufferLen);
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -3004,7 +3004,8 @@ private:
     std::string GetUserString(rct_string_id stringId)
     {
         const auto originalString = _s4.string_table[(stringId - USER_STRING_START) % 1024];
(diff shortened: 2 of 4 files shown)


Gymnasiast:
@telk5093 Could you test this?

telk5093:
@Gymnasiast It works for me ;)

Gymnasiast:
Hm, why is this only needed for import? Is the export already correct?

src/openrct2/rct2/RCT2.h
```diff
@@ -779,3 +779,9 @@ std::vector<uint8_t> DecryptSea(const fs::path& path);
 ObjectEntryIndex RCT2RideTypeToOpenRCT2RideType(uint8_t rct2RideType, const rct_ride_entry* rideEntry);
 bool RCT2RideTypeNeedsConversion(uint8_t rct2RideType);
 uint8_t OpenRCT2RideTypeToRCT2RideType(ObjectEntryIndex openrct2Type);
+
+/**
+ * Iterates an RCT2 string buffer and returns the length of the string in bytes.
+ * Handles single and multi-byte strings.
+ */
+size_t GetRCT2StringBufferLen(const char* buffer, size_t maxBufferLen);
```
Gymnasiast, line 787: Should be called GetRCT12StringBufferLen and be put in RCT12.cpp/.h instead.

IntelOrca:
> Hm, why is this only needed for import? Is the export already correct?

Yes, import was just truncating it at the first null terminator. But multi-byte strings do include null bytes which do not mark the end of the string.

PR #13639, opened 2020-12-23
Fix #13574: Object selection crashes if 'originalId' missing

### LINKED ISSUE #13574: Object selection crashes if JSON object is missing `originalId`
**OS:** Arch Linux
**Version:** 0.3.2 (develop)
**Commit/Build:** 3feefb0

If a JSON scenery object is installed which is missing the `originalId` field, the game will crash with an assertion failure upon opening the object selection.

This happens because the object selection uses the object type from the DAT header to determine which objects to display in the current tab (and not the one specified in the JSON). If the `originalId` field is omitted, this defaults to zero, which means the object will appear in the ride tab even if it isn't a ride. The crash occurs because `ride_type_get_string_id` returns `NULL`.

I know that the `originalId` field is currently mandatory but I think that the code should check that `originalId` is present and that the ride type matches the one specified in the JSON when building the object repository, and fail with an error otherwise (as it does if the JSON contains a syntax error). Currently, it's not obvious why the game is crashing.

Gymnasiast:
Not sure if this should be in the changelog or not.

### DIFF
--- a/src/openrct2/object/Object.h
+++ b/src/openrct2/object/Object.h
@@ -108,6 +108,12 @@ struct rct_object_entry
         return static_cast<ObjectType>(flags & 0x0F);
     }
 
+    void SetType(ObjectType newType)
+    {
+        flags &= ~0x0F;
+        flags |= (static_cast<uint8_t>(newType) & 0x0F);
+    }
+
     std::optional<uint8_t> GetSceneryType() const;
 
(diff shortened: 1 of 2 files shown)


IntelOrca (approved):

tupaschoal:
> Not sure if this should be in the changelog or not.

I guess it should, I think I've been seeing stacks with this crash for a while.
Btw, nice fix!

PR #13646, opened 2020-12-24
Fix #13644, #13102: Underflows in ride graph, temperature, cut-away view

### LINKED ISSUE #13644: Cut-away view shows 6,551.1 ft when going below 0 ft
**OS:** Windows 10
**Version:** 0.3.2-252-gf9e90ed84
**Commit/Build:** f9e90ed

When selecting a cut-away height less than 0 ft, the displayed value underflows and shows 6,551.1 ft (then 6,548.6 ft, then 6,546.1 ft, etc.)

**Steps to reproduce:**
1. Open cut-away view
2. Ensure the display is set to distance instead of units
3. Set the slider / value to less than 0.0 ft

**Screenshots / Video:**

Screenshot

Gymnasiast:
This silently reintroduces #1013. To fix that, we'll need to use 32 bits for length strings.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -12,6 +12,7 @@
 - Feature: [#13583] [Plugin] Add allowed_hosts to plugin section of config.
 - Feature: [#13614] Add terrain surfaces from RollerCoaster Tycoon 1.
 - Fix: [#12895] Mechanics are called to repair rides that have already been fixed.
+- Fix: [#13102] Underflow on height chart (Ride measurements).
 - Fix: [#13257] Rides that are exactly the minimum objective length are not counted.
 - Fix: [#13334] Uninitialised variables in CustomTabDesc.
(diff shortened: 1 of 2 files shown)


distribution/changelog.txt
```diff
@@ -12,6 +12,7 @@
 - Feature: [#13583] [Plugin] Add allowed_hosts to plugin section of config.
 - Feature: [#13614] Add terrain surfaces from RollerCoaster Tycoon 1.
 - Fix: [#12895] Mechanics are called to repair rides that have already been fixed.
+- Fix: [#13102] Underflow on height chart (Ride measurements).
```
duncanspumpkin, line 15: Isn't this just a regression?

Gymnasiast: #13644 is, but #13102 has been in the game since the 0.3.1 release.

IntelOrca:
Nice catch, we might be able to use the template Format function to fix #1013 so that we can use a larger signed integer.

PR #13647, opened 2020-12-24
Close #13637: refactor sprite compiler for filesystem efficiency

### LINKED ISSUE #13637: Refactor Sprite Compiler
The sprite compiler function `sprite build` is inefficient with the way it builds a new g2.dat file. As this function is run every time we compile on the CI's. It may reduce compile times by improving it. See this link for more info on the sprite compiler: https://github.com/OpenRCT2/OpenRCT2/wiki/Sprite-compiler

The problem with the sprite build is that it opens adds an image and finally saves the g2.dat file for each individual image. Instead it should add all of the images and then save. There is no reason for it to open the g2.dat file. This should just be a case of reordering the loop located at: https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/CmdlineSprite.cpp#L602

Have a look at the part before as well. I think there should only ultimately be one save for the whole function. Don't forget to test the result as its easy to forget (Self Doh! https://github.com/OpenRCT2/OpenRCT2/issues/13597)

csunday95:
Close #13637: refactor sprite compiler for filesystem efficiency

 - reordered sprite compilation loop so that the sprite file isn't reopened and closed for each image added
 - verified that: 
    - compilation is now much faster
    - g2.dat file generated from `resources/g2/sprites.json` has an identical hash before and after change
 - created a couple simple test cases for the sprite CLI


csunday95:
I made this as a draft because there are a couple points of discussion:

1. do we want to guarantee that a failed `sprite build [output] spritefile.json` will not result in modifying the output file? I think that both the original code and this PR code do not guarantee that

2. I created a couple simple test cases, but in doing so noticed that its not very convenient to create unit tests for the CLI interface. The lowest layer currently takes char** argv/int argc and is hardcoded to write to stdout/stdin/filesystem. Because of that, the test cases I generated needed to access actual files (which I usually try to avoid with unit tests for speed), so I added some to the testdata folder structure

src/openrct2/CmdlineSprite.cpp
```diff
@@ -598,6 +598,13 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
 
         json_t sprite_description;
 
+        auto spriteFile = SpriteFile::Open(spriteFilePath);
+        if (!spriteFile.has_value())
```
duncanspumpkin, line 602: There isn't any point opening this file. It was created on line 595. Just skip the save/load.

duncanspumpkin:
The layout of the output of that file is very important so it's better that it fails and doesn't output anything than for it to output a bad file that isn't noticed.

src/openrct2/CmdlineSprite.cpp
csunday95: ah I see, I missed that save call; that's also why its breaking an existing file even on failure since it instantly overwrites it with just the header

duncanspumpkin:
For your commit message could you try stick to this format: https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages I know this is just a draft so you're still playing about but just something to keep in mind. As its not a bug it would be `Close #...: ...`

csunday95:
> 
> 
> For your commit message could you try stick to this format: https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages I know this is just a draft so you're still playing about but just something to keep in mind. As its not a bug it would be `Close #...: ...`

Thanks, I missed this section in the wiki when looking for a contribution guide. One question I have is: For incremental commits while an issue is being worked on, would each individual commit still have the `Close #...` / `Fix ...` etc as a commit summary prefix, or just the merge commit? Personally I tend to make a lot of little commits while incrementally working on something, but I can try to contain myself to a single commit / squash the commits at the end if that's preferred.

duncanspumpkin:
We tend to squash merge small changes like this so in general just put your first commit as a message that would encompass the full PR.

duncanspumpkin (approved):

src/openrct2/CmdlineSprite.cpp
```diff
@@ -632,25 +633,18 @@ int32_t cmdline_for_sprite(const char** argv, int32_t argc)
...
-            }
+            spriteFile.AddImage(importResult.value());
 
             if (!silent)
                 fprintf(stdout, "Added: %s\n", imagePath.c_str());
         }
 
+        if (!spriteFile.Save(spriteFilePath))
+        {
+            fprintf(stderr, "Could not save sprite file: Canceling\n");
```
Gymnasiast, line 644:
```suggestion
            log_error("Could not save sprite file, cancelling.");
```

IntelOrca: I'd keep it using fprintf, log_error spits out the line and column and filename.

PR #13648, opened 2020-12-26
fix #13615 by inserting required action implementation

### LINKED ISSUE #13615: Some Game Actions have incomplete implementations

csunday95:
- add AcceptParameters() where missing to `GameAction` s exposed in the script engine
 - implement additional deserialization methods as needed in GameAction.h

### DIFF
--- a/src/openrct2/actions/FootpathPlaceFromTrackAction.cpp
+++ b/src/openrct2/actions/FootpathPlaceFromTrackAction.cpp
@@ -30,6 +30,14 @@ FootpathPlaceFromTrackAction::FootpathPlaceFromTrackAction(
 {
 }
 
+void FootpathPlaceFromTrackAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit(_loc);
+    visitor.Visit("slope", _slope);
+    visitor.Visit("type", _type);
+    visitor.Visit("edges", _edges);
+}
+
 void FootpathPlaceFromTrackAction::Serialise(DataSerialiser& stream)
 {
     GameAction::Serialise(stream);
--- a/src/openrct2/actions/ClearAction.cpp
+++ b/src/openrct2/actions/ClearAction.cpp
@@ -30,6 +30,12 @@ ClearAction::ClearAction(MapRange range, ClearableItems itemsToClear)
 {
 }
 
+void ClearAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit(_range);
+    visitor.Visit("items", _itemsToClear);
+}
+
 void ClearAction::Serialise(DataSerialiser& stream)
 {
     GameAction::Serialise(stream);
--- a/src/openrct2/actions/ClearAction.h
+++ b/src/openrct2/actions/ClearAction.h
@@ -34,6 +34,7 @@ public:
     ClearAction() = default;
     ClearAction(MapRange range, ClearableItems itemsToClear);
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override;
     void Serialise(DataSerialiser & stream) override;
     GameActions::Result::Ptr Query() const override;
     GameActions::Result::Ptr Execute() const override;
--- a/src/openrct2/actions/FootpathPlaceFromTrackAction.h
+++ b/src/openrct2/actions/FootpathPlaceFromTrackAction.h
@@ -23,6 +23,7 @@ public:
     FootpathPlaceFromTrackAction() = default;
     FootpathPlaceFromTrackAction(const CoordsXYZ& loc, uint8_t slope, ObjectEntryIndex type, uint8_t edges);
 
+    void AcceptParameters(GameActionParameterVisitor & visitor) override;
     uint16_t GetActionFlags() const override;
 
     void Serialise(DataSerialiser & stream) override;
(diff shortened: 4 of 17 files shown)


csunday95:
intended to replace #13616

duncanspumpkin:
Don't open a new PR for the same thing. You can pretty much always recover your branch. By opening a new one we loose all the comments that we made on the previous one.

src/openrct2/actions/FootpathPlaceFromTrackAction.cpp
```diff
@@ -30,6 +30,14 @@ FootpathPlaceFromTrackAction::FootpathPlaceFromTrackAction(
 {
 }
 
+void FootpathPlaceFromTrackAction::AcceptParameters(GameActionParameterVisitor& visitor)
```
duncanspumpkin, line 33: Remove this. This action should ONLY be used by track designs and never by plugins.

src/openrct2/actions/LoadOrQuitAction.cpp
```diff
@@ -18,6 +18,12 @@ LoadOrQuitAction::LoadOrQuitAction(LoadOrQuitModes mode, PromptMode savePromptMo
 {
 }
 
+void LoadOrQuitAction::AcceptParameters(GameActionParameterVisitor& visitor)
```
duncanspumpkin, line 21: debatable if this is safe to perform from a plugin as it will unload the plugin. This may cause unexpected behaviour.

src/openrct2/actions/PeepPickupAction.cpp
```diff
@@ -21,6 +21,14 @@ PeepPickupAction::PeepPickupAction(PeepPickupType type, uint32_t spriteId, const
 {
 }
 
+void PeepPickupAction::AcceptParameters(GameActionParameterVisitor& visitor)
```
duncanspumpkin, line 24: Will plugins understand this action. Its not a simple action to use.

duncanspumpkin:
Make sure all of the comments from https://github.com/OpenRCT2/OpenRCT2/pull/12933 are applied as well.

csunday95:
The actions that I am implementing are based entirely on the list of exposed actions in ScriptEngine.cpp:964. If there are actions we don't want plugins to use, they should be removed there rather than just having a broken implementation.

csunday95:
one comment from #12933 is I think worthy of discussion:

>There are a few here that I do not think should be usable by plugins and others that are dangerous to use. A lot of the parameters need some careful thought on their names as we should standardise it as much as possible to prevent issues with later renaming.

A standardized naming scheme for arguments would be nice, as would a document detailing all the exposed callbacks and their arguments. There's also the general concern of the plugin user needing to pass enumeration values, which will work under the current system, but be very arcane and require source code examination until a document is made.

src/openrct2/scripting/ScriptEngine.cpp
```diff
@@ -1027,8 +1025,6 @@ const static std::unordered_map<std::string, GameCommand> ActionNameToType = {
     { "staffsetname", GameCommand::SetStaffName },
     { "staffsetorders", GameCommand::SetStaffOrders },
     { "staffsetpatrolarea", GameCommand::SetStaffPatrol },
-    { "surfacesetstyle", GameCommand::ChangeSurfaceStyle },
-    { "tilemodify", GameCommand::ModifyTile },
     { "trackdesign", GameCommand::PlaceTrackDesign },
```
tupaschoal, line 1028: I think for these actions that have been removed from the plugin API, like `footpathplacefromtrack`, `loadorquit` and others, their `AcceptParameters` function should rather be no-ops and actually have a comment there stating why they're intentionally empty, so that people don't think we just forgot to add them a few moments from now.

Issue #4550, opened 2016-10-07
Owner rights on placed down coasters/sceanery/etc


Nubbie:
Some games gives the player a specific ID (in this case it could be our 'hash') and only let the person who have built it, touch it (great for greifers who delete/close/remove rides)

The only way for others to touch that coaster/store would be if they accept a permission for that player to make edits on the thing the original creator placed down

janisozaur:
Not possible until we get our own format where we can store such things.

IntelOrca:
Until you get greifers who obstruct other players with rides.

Nubbie:
@intelorca could be done now too

PR #13651, opened 2020-12-26
Reserve room for tile owner

Gymnasiast:
Part of #4550. In preparation for the new save format.

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -477,9 +477,10 @@ template<> struct DataSerializerTraits_t<TileElement>
         stream->WriteValue(tileElement.Flags);
         stream->WriteValue(tileElement.base_height);
         stream->WriteValue(tileElement.clearance_height);
-        for (int i = 0; i < 4; ++i)
+        stream->WriteValue(tileElement.owner);
+        for (int i = 0; i < 3; ++i)
         {
-            stream->WriteValue(tileElement.pad_04[i]);
+            stream->WriteValue(tileElement.pad_05[i]);
         }
         for (int i = 0; i < 8; ++i)
         {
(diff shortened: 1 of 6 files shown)


src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -477,9 +477,10 @@ template<> struct DataSerializerTraits_t<TileElement>
         stream->WriteValue(tileElement.Flags);
         stream->WriteValue(tileElement.base_height);
         stream->WriteValue(tileElement.clearance_height);
-        for (int i = 0; i < 4; ++i)
+        stream->WriteValue(tileElement.owner);
+        for (int i = 0; i < 5; ++i)
```
duncanspumpkin, line 481: shouldn't your go down from 4 to 3?

Gymnasiast: Good catch.

ZehMatt:
Player ids are not fixed in multiplayer, this is based on the time a player joins and cant be used for persistent storage, the only safe way is to store the hash but that would add quite some overhead.

IntelOrca:
You wouldn't use player IDs anyway as that prevents two players from working on the area. It would be company IDs which players assigned to a company.

Gymnasiast:
Exactly. This makes reserves room for 16 companies, which should be more than enough for the foreseeable future. The remaining 4 bits are not yet used, but I'm sure they will be very handy for something in the future.

With that bit cleared up, could I get a review and an approval/request for changes?

src/openrct2/world/Map.cpp
```diff
@@ -1197,7 +1197,8 @@ TileElement* tile_element_insert(const CoordsXYZ& loc, int32_t occupiedQuadrants
     newTileElement->SetLastForTile(isLastForTile);
     newTileElement->SetOccupiedQuadrants(occupiedQuadrants);
     newTileElement->SetClearanceZ(loc.z);
-    std::memset(&newTileElement->pad_04, 0, sizeof(newTileElement->pad_04));
+    newTileElement->owner = 0;
```
ZehMatt, line 1200: Might be worth to introduce a constant for this where 0 is undefined, ownerless should be perhaps the maximum possible value, this ensures we are able to tell the difference for when we actually use this field for compatibility with saves generated before.

Gymnasiast: I suppose 0 could mean "unowned" and 1 could mean "park 1". The advantage of this would be that it is trivial to increase the limit beyond 15 later on.

PR #13654, opened 2020-12-28
Close #13620: Add patrol area retrieval method for plugins

### LINKED ISSUE #13620: [Plugin] Add support for patrol areas to Staff
Hi there,

thanks for the great work on OpenRCT2!

I thought about writing a plugin that makes staff more intelligent by assigning patrol areas automatically. Maybe instead the wages are increased (to pay some imaginary middle-management).

Therefore, in the scripting support I'd need the possibility to
- get patrol area for staff members (`Staff.getPatrolArea()`)
- set patrol area for staff members (`Staff.setPatrolArea(coords)`)
- (optional) set wages for staff members (`Staff.setWage(money)`)
- (optional) set "visibility" of full bins to a high number (imagine "smart bins" that signal when they are full)

Unfortunately, I don't know C++ myself, so I'd be happy for someone to take this over.

csunday95:
- allows a script engine plugin to conveniently get the staff patrol
   area matrix for a particular staff member
 - returns a javascript Array of Arrays of bools representing a 64x64
   grid

### DIFF
--- a/src/openrct2/scripting/ScEntity.hpp
+++ b/src/openrct2/scripting/ScEntity.hpp
@@ -1001,6 +1001,7 @@ namespace OpenRCT2::Scripting
             auto peep = GetPeep();
             return peep != nullptr ? peep->Intensity.GetMinimum() : 0;
         }
+
         void minIntensity_set(uint8_t value)
         {
             ThrowIfGameStateNotMutable();
@@ -1077,7 +1078,6 @@ namespace OpenRCT2::Scripting
         }
 
     private:
-
         static inline int32_t patrolRegionHeight(const MapRange& region)
         {
             return (region.GetY2() - region.GetY1()) / (COORDS_XY_STEP * 4);
@@ -1090,24 +1090,24 @@ namespace OpenRCT2::Scripting
 
         static inline void mergeIntoRegion(MapRange& target, const MapRange& toMerge)
         {
-            //printf(
+            // printf(
             //    "target %d %d %d %d\n", target.Point1.x / 128, target.Point1.y / 128, target.Point2.x / 128,
             //    target.Point2.y / 128);
-            //printf(
+            // printf(
             //    "toMerge %d %d %d %d\n", toMerge.Point1.x / 128, toMerge.Point1.y / 128, toMerge.Point2.x / 128,
             //    toMerge.Point2.y / 128);
             target.Point1.x = std::min(target.Point1.x, toMerge.Point1.x);
             target.Point1.y = std::min(target.Point1.y, toMerge.Point1.y);
             target.Point2.x = std::max(target.Point2.x, toMerge.Point2.x);
             target.Point2.y = std::max(target.Point2.y, toMerge.Point2.y);
-            //printf(
+            // printf(
             //    "merged %d %d %d %d\n\n", target.Point1.x / 128, target.Point1.y / 128, target.Point2.x / 128,
             //    target.Point2.y / 128);
         }
 
         void mergeRegions()
         {
-            // from top left, go across horizontal scanlines to create as many one
+            // from top left, go across horizontal scanlines to create as many
             // horizontally maximized regions; then go through vertical scanlines and merge
             // as many of the horizontal regions as possible into bigger rectangles; should
             // produce minimum possible number of rectangular regions (though not
@@ -1130,7 +1130,7 @@ namespace OpenRCT2::Scripting
                 auto initialY = topLeft.y / PATROL_REGION_STEP;
                 auto finalX = bottomRight.x / PATROL_REGION_STEP;
                 auto finalY = bottomRight.y / PATROL_REGION_STEP;
-                //printf("%d %d %d %d\n", initialX, initialY, finalX, finalY);
+                // printf("%d %d %d %d\n", initialX, initialY, finalX, finalY);
                 for (int32_t y = initialY; y < finalY; y++)
                 {
                     for (int32_t x = initialX; x < finalX; x++)
@@ -1139,7 +1139,7 @@ namespace OpenRCT2::Scripting
                     }
                 }
             }
-        
+
             // horizontal scanning
             for (int32_t y = 0; y < 64; y++)
             {
@@ -1217,7 +1217,7 @@ namespace OpenRCT2::Scripting
                         // aren't currently merging, start merging on this new region
                         currentRegion = nextRegion;
                         currentWidth = patrolRegionWidth(*currentRegion);
-                        //printf("new region width: %d\n", currentWidth);
+                        // printf("new region width: %d\n", currentWidth);
                         continue;
                     }
                     else if (nextRegion && currentRegion)
@@ -1225,7 +1225,7 @@ namespace OpenRCT2::Scripting
                         // check if height matches, and merge if so
                         if (patrolRegionWidth(*nextRegion) == currentWidth)
                         {
-                            mergeRegions(*currentRegion, *nextRegion);
+                            mergeIntoRegion(*currentRegion, *nextRegion);
                             patrolBitmask[y][x] = currentRegion;
                             mergedRegions.erase(nextRegion);
                             mergedRegions.insert(currentRegion);
@@ -1241,9 +1241,9 @@ namespace OpenRCT2::Scripting
             }
             // assemble new region vector
             _regions.clear();
-            for(auto& regionPtr : mergedRegions)
+            for (auto& regionPtr : mergedRegions)
             {
-                //printf(
+                // printf(
                 //    "Final region: %d %d %d %d\n", regionPtr->GetLeft() / PATROL_REGION_STEP,
                 //    regionPtr->GetTop() / PATROL_REGION_STEP, regionPtr->GetRight() / PATROL_REGION_STEP,
                 //    regionPtr->GetBottom() / PATROL_REGION_STEP);
@@ -1289,8 +1289,8 @@ namespace OpenRCT2::Scripting
         {
             for (auto& region : _regions)
             {
-                if (toCheck.x >= region.GetLeft() && toCheck.x < region.GetRight() &&
-                    toCheck.y >= region.GetTop() && toCheck.y < region.GetBottom())
+                if (toCheck.x >= region.GetLeft() && toCheck.x < region.GetRight() && toCheck.y >= region.GetTop()
+                    && toCheck.y < region.GetBottom())
                 {
                     return true;
                 }
@@ -1460,7 +1460,6 @@ namespace OpenRCT2::Scripting
                     // to only step by whole coordinate -> left shift by 7
                     if (staff->IsPatrolAreaSet({ x * PATROL_REGION_STEP, y_index }))
                     {
-                        //printf("%d %d in patrol area\n", x, y);
                         // patrol region is multiplied by 4 to return to tile coordinates
                         auto patrolRegion = MapRange(
                             { x * PATROL_REGION_STEP, y * PATROL_REGION_STEP },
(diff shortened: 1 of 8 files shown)


src/openrct2/scripting/ScMap.hpp
```diff
@@ -168,6 +191,7 @@ namespace OpenRCT2::Scripting
             dukglue_register_method(ctx, &ScMap::getTile, "getTile");
             dukglue_register_method(ctx, &ScMap::getEntity, "getEntity");
             dukglue_register_method(ctx, &ScMap::getAllEntities, "getAllEntities");
+            dukglue_register_method(ctx, &ScMap::getStaffPatrolArea, "getStaffPatrolArea");
```
duncanspumpkin, line 194: Why would you make this part of map and not part of entity?

duncanspumpkin:
Please follow https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages for your commit messages.

IntelOrca:
I was hoping your function would return a list of tile co-coordinates that are in their patrol area. Returning a large boolean array is probably going to be a lot more inefficient. It might even be better to calculate rectangle ranges for the patrol area to help improve the performance.

It would probably also be useful to have a set of functions on a patrol area object. E.g.
```ts
interface PatrolArea {
    tiles: CoordsXY[];

    add(coords: CoordsXY[]): void;
    remove(coords: CoordsXY[]): void;
    contains(coord: CoordsXY): bool;
}

interface Staff extends Peep {
    // existing stuff ...

    readonly patrolArea: PatrolArea;
}
```

src/openrct2/scripting/ScMap.hpp
csunday95: Adding getPatrolArea as a dukglue method on the SCStaff type does make more sense; making that modification now. I'm also not clear on which part of the commit message guide I'm not following; just the keep it in present tense part?

csunday95:
I've moved the getPatrolArea method to the ScStaff object. I've also added a hasPatrolArea property to disambiguate the situation where the staff has "no patrol area", which is equivalent to having every possible location in its patrol area. I can also change to this to returning bounding rectangles as an Array of MapRange (with the topleft/bottom right map coordinates as the points), or create a different method for that.

csunday95:
Also, as it currently stands, a plugin will need to use scripting engine actions to do tasks like setting staff patrol area. E.g.:

```js
context.executeAction("staffsetpatrolarea", {
    peep: peep.id,
    loc: locationData
}, function (result) {...handle result...});
```

Would it be preferred to implement more intermediary duk methods, e.g. create a setPatrolArea function on ScStaff that handles creating/executing the requisite action(s)

saeltz:
> Would it be preferred to implement more intermediary duk methods, e.g. create a setPatrolArea function on ScStaff that handles creating/executing the requisite action(s)?

From a plugin author point of view this is much preferred.

csunday95:
okay, this latest iteration now returns an Array of TileRange objects, which specify the corners of the patrol region in tile coordinates. I've added a definition for it to the TypeScript file as

```ts
interface TileRange {
    x1: number;
    y1: number;
    x2: number; 
    y2: number;
}
```

Currently, I have the range as _inclusive_, i.e. a tile range for a patrol region will look like

```js
{x1: 4, y1: 8, x2: 7, y2: 11}
```

but switching it to the exclusive upper bound convention is trivial. I wasn't sure which convention is used for MapRange etc currently.

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -1188,6 +1190,39 @@ namespace OpenRCT2::Scripting
...
+            std::vector<DukValue> patrolArea;
+            auto ctx = GetContext()->GetScriptEngine().GetContext();
+            auto staff = GetStaff();
+            if (staff == nullptr)
+            {
+                return {};
+            }
+            for (int32_t y = 0; y < 64; y++)
+            {
+                auto y_index = y << 7;
```
duncanspumpkin, line 1211: don't use shifts unless you absolutely have to. Leave it to the compiler to optimise them.

duncanspumpkin: Also you could have just done +128 instead of ++ to avoid the multiplication.

src/openrct2/world/Location.hpp
```diff
@@ -729,6 +729,16 @@ struct MapRange : public RectRange<CoordsXY>
     }
 };
 
+struct TileRange : public RectRange<TileCoordsXY>
```
duncanspumpkin, line 732: We are trying to remove more instances of TileCoordsXY not add more. Leave in terms of `CoordsXY` (i.e. just use MapRange)

IntelOrca:
> Would it be preferred to implement more intermediary duk methods, e.g. create a setPatrolArea function on ScStaff that handles creating/executing the requisite action(s)?

Currently only some of the network commands actually execute game actions. This is mainly for convenience and they are a special case of game actions.

The game actions are designed around how the user edits the patrol areas. Plugins should instead create their own custom action which edits the patrol area more efficiently using a more direct API. This allows you to send one game action to set a large patrol area which would usually take many game actions.

I think it is important to get the API right first time otherwise it causes a lot of issues for existing plugins when we need to add extra functionality requiring the API to change. The API I mocked up above provides a simpler method for checking individual tile while allowing a batch of tiles to be added or removed. It will still behave correctly when we finally allow patrol areas to be smaller than 4x4 regions. What are your thoughts? Are you able go further and implement some kind of ScPatrolArea class or would you prefer me to take over?

src/openrct2/world/Location.hpp
csunday95: Well I see using MapRange in this scenario as going one of two ways:

1) plugin writers have to be aware of the 32x resolution per tile, i.e. when looking at a tile with the file inspector, the x/y values have to be multiplied by 32
2) OR we begin using MapRange with a completely different basis.

What's the rationale behind trying to remove TileCoordsXY? it seems logical to me to have both classes for the different coordinate systems

src/openrct2/scripting/ScEntity.hpp
csunday95: I was simply copying the shift logic used in the Staff class for consistency, but there's myriad ways to write this code of course. The most "readable" implementation, I think, would be keeping the single ++'s and using expanded multiplication like 
`y_index = y * 32 * 4;`

csunday95:
@IntelOrca I'm willing to implement as elaborate as a system as needed, I was simply trying to create a minimal system that would work without making too many assumptions as to what was required/desired. Creating a new Duktape class as you suggested seems like a good way to go, other than it also possibly needing to retain the staff sprite ID (unless we want PatrolArea 's to be entirely "abstract" and not associated with any particular staff member). So for clarity, the general consensus so far seems to be:

1) Create a new ScPatrolArea class for duktape
   - this class has an array of MapRange objects as its member data, possibly merged to minimize the number of array entries
   - has `Add()`, `Remove()` and `Contains()` methods 
   - Question: I think for efficiency, we would want add/remove/contain to execute on the plugin side and commit the new data to a staff member only when local operations are complete? 
 
2) make the getPatrolArea method on the ScStaff class return an ScPatrolArea instance
   - Question: do we want ScPatrolArea's to be independent of a particular staff member, or possibly store a staff sprite ID they correspond to?
 
3) >Plugins should instead create their own custom action which edits the patrol area more efficiently using a more direct API. This allows you to send one game action to set a large patrol area which would usually take many game actions.

 - so if I'm understanding correctly in this case we would create a `SetPatrolArea` method to complement the `Get` one on the ScStaff class. This `SetPatrolArea` would itself take a ScPatrolArea instance as its argument, and itself call `staff_toggle_patrol_area()` as needed/perform the required checks?

src/openrct2/scripting/ScEntity.hpp
Gymnasiast: The 32 is for the number of big x/y coordinates in 1 small coordinate. We have a constant for that: `COORDS_XY_STEP`. 

The 4 is for the number of tiles in a single piece of patrol area. I don't know if we have a constant for it, but if we haven't, please introduce one.

src/openrct2/world/Location.hpp
Gymnasiast: Mostly to avoid endlessly converting between the two. Entities always use CoordsXY, tile elements use a mixture. (Don't get us started on the numerous places that use X and Y from TileCoordsXYZ, but Z from CoordsXYZ...)

csunday95: so for the plugin interface, should we still send out serialized coordinates in the CoordsXY space (i.e. 32 steps per tile) or in tile space? I think very few plugins would actually make use of the higher precision coordinate system, but perhaps some.

IntelOrca:
I think `ScPatrolArea` should be created on-demand when `staffMember.patrolArea` is accessed. `ScPatrolArea` has a peep ID associated with it and it uses that alone to read/edit the patrol area. Given the add and remove members take one or more coordinates, you can commit after each call.

csunday95:
>I think ScPatrolArea should be created on-demand when staffMember.patrolArea is accessed

So something like a lazy property getter method in the typescript definitions? It appears that the openrct2.d.ts file is entirely declarative now, would we want to start adding some actual executable code to it. Otherwise I'm not sure how you'd do a lazy evaluation type property. 
```

**Comment by IntelOrca** [comment]
@csunday95 No I just mean the getter for it in the C++ code would return a new instance of ScPatrolArea. Have a look at how map elements work or the scenario options.

File: src/openrct2/world/Location.hpp (hunk 3 above)
Comment (duncanspumpkin) [reply src/openrct2/world/Location.hpp:732]: Everything at present plugin wise uses CoordsXY as far as I know. There shouldn't be anything that works in terms of TileCoordsXY.

File: src/openrct2/scripting/ScEntity.hpp [hunk 4]
```diff
@@ -1057,6 +1057,251 @@ namespace OpenRCT2::Scripting
...
+
+        void mergeRegions()
+        {
+            // from top left, go across horizontal scanlines to create as many one
+            // horizontally maximized regions; then go through vertical scanlines and merge
+            // as many of the horizontal regions as possible into bigger rectangles; should
+            // produce minimum possible number of rectangular regions (though not
+            // necessarily the most "square" ones)
+            constexpr int32_t PATROL_REGION_STEP = COORDS_XY_STEP * COORDS_XY_PATROL_STEP;
+            std::array<std::array<MapRange*, 64>, 64> patrolBitmask = { nullptr };
```
csunday95, line 1116: quite a large stack allocation; could be replaced with a `std::vector<std::vector<MapRange*>>` and populated in place on line 1138 below

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -1057,6 +1057,251 @@ namespace OpenRCT2::Scripting
...
+                {
+                    for (int32_t x = initialX; x < finalX; x++)
+                    {
+                        patrolBitmask[y][x] = &region;
+                    }
+                }
+            }
+        
+            // horizontal scanning
+            for (int32_t y = 0; y < 64; y++)
```
csunday95, line 1144: nested loops can be collapsed to a big loop with calculated indexing

tupaschoal:
Is this just missing the compilation error fixes or is it still a WIP?

csunday95:
This is still WIP; I need to implement the add/remove functions and create some unit tests for the merging algorithm. I was pretty busy during the holiday season, but I should have some time to work on it today and this week

IntelOrca:
I've labelled this as NSF, as that will change patrol areas to single tiles. It should wait until then and make use of the utility methods that will come with it.

Issue #13650, opened 2020-12-26
System libdiscord RPC


Mailaender:
It would be great if the cmake system could detect https://build.opensuse.org/package/show/games/discord-rpc instead of trying to download stuff from the internet, which is not allowed in such build environments.

janisozaur:
I don't think we're downloading anything from the internet for discord, at least the build system doesn't. It detects discord-rpc which is already present. That still sounds like a nice addition to detect the system one, but given discord-rpc was implemented back in the day when it was still supported and there weren't ready packages, I think this will depend on a PR by someone from outside of the team. Would you like to contribute such code @Mailaender ?

Mailaender:
It seems you currently rely on a manually patched version:
https://github.com/OpenRCT2/OpenRCT2/blob/c36ec5c5cebe2093396256259485d87d31529468/scripts/linux/install.sh#L102

Looks like that needs to get upstreamed first.

Mailaender:
Nevermind, you already did https://github.com/discord/discord-rpc/pull/179 and it is integrated in version 3.4.0.

PR #13658, opened 2020-12-29
Detect and link system wide installed Discord RPC on Linux

Mailaender:
Closes https://github.com/OpenRCT2/OpenRCT2/issues/13650. So far only available for @openSUSE and @ArchLinux https://pkgs.org/search/?q=discord-rpc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,7 +153,16 @@ if (ENABLE_SCRIPTING)
 endif ()
 
 if (NOT DISABLE_DISCORD_RPC)
-    if(EXISTS "${ROOT_DIR}/discord-rpc")
+    if (UNIX AND NOT APPLE)
+        find_package(DiscordRPC)
+        if(${DISCORDRPC_FOUND})
+            set(HAVE_DISCORD_RPC TRUE)
+            add_definitions(-D__ENABLE_DISCORD__)
+            include_directories(DISCORDRPC_PROCESS_INCLUDES)
+        endif()
+    endif()
+
+    if (NOT HAVE_DISCORD_RPC AND EXISTS "${ROOT_DIR}/discord-rpc")
         # Don't build discord's examples, some of which are in C and do not honour
         # the flags we set for C++. Also we don't use the provided examples.
         set(BUILD_EXAMPLES OFF CACHE BOOL "Build example apps")
--- a/cmake/FindDiscordRPC.cmake
+++ b/cmake/FindDiscordRPC.cmake
@@ -1,0 +1,25 @@
+include(LibFindMacros)
+
+libfind_pkg_check_modules(DISCORDRPC_PKGCONF DISCORDRPC)
+
+find_path(DISCORDRPC_INCLUDE_DIR
+	NAMES discord_rpc.h
+	PATHS
+		${DISCORDRPC_PKGCONF_INCLUDE_DIRS}
+		"/usr/include"
+		"/usr/local/include"
+)
+
+find_library(DISCORDRPC_LIBRARY
+	NAMES discord-rpc
+	PATHS
+		${DISCORDRPC_PKGCONF_LIBRARY_DIRS}
+        "/usr/lib"
+        "/usr/lib64"
+        "/usr/local/lib"
+        "/usr/local/lib64"
+)
+
+set(DISCORDRPC_PROCESS_INCLUDES DISCORDRPC_INCLUDE_DIR)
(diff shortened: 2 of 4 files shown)


janisozaur:
Is the intent here to move detection of pre-downloaded discord RPC to cmake module? In such way that module detects both the system library and the downloaded one?

Mailaender:
My intent that if it detects a system one, use that. Afterwards, search for a downloaded one.

adam-bloom:
@Mailaender why do we need to add custom cmake modules to find the discord-rpc library? Does `find_library` on its own not work?

Mailaender:
I otherwise get
```
[cmake] CMake Warning at CMakeLists.txt:157 (find_package):
[cmake]   By not providing "FindDiscordRPC.cmake" in CMAKE_MODULE_PATH this project
[cmake]   has asked CMake to find a package configuration file provided by
[cmake]   "DiscordRPC", but CMake did not find one.
[cmake] 
[cmake]   Could not find a package configuration file provided by "DiscordRPC" with
[cmake]   any of the following names:
[cmake] 
[cmake]     DiscordRPCConfig.cmake
[cmake]     discordrpc-config.cmake
[cmake] 
[cmake]   Add the installation prefix of "DiscordRPC" to CMAKE_PREFIX_PATH or set
[cmake]   "DiscordRPC_DIR" to a directory containing one of the above files.  If
[cmake]   "DiscordRPC" provides a separate development package or SDK, be sure it has
[cmake]   been installed.
```

However I am also not very experienced with `cmake`.

adam-bloom:
@Mailaender if you need to add a config.cmake file, look at https://github.com/OpenRCT2/OpenRCT2/blob/develop/cmake/duktapeConfig.cmake as an example. I don't think you need to add `LibFindMacros` at all.

adam-bloom:
I'm not entirely sure that `LibFindMacros` is necessary, but I haven't looked at this in detail yet. We don't need that to find any other package, system installed or not.

CMakeLists.txt
```diff
@@ -153,7 +153,14 @@ if (ENABLE_SCRIPTING)
...
 if (NOT DISABLE_DISCORD_RPC)
-    if(EXISTS "${ROOT_DIR}/discord-rpc")
+    if (UNIX AND NOT APPLE)
+        find_package(DiscordRPC)
+        if(${DISCORDRPC_FOUND})
+            set(HAVE_DISCORD_RPC ON)
+            add_definitions(-D__ENABLE_DISCORD__)
+            include_directories(DISCORDRPC_PROCESS_INCLUDES)
+        endif()
+    elseif(EXISTS "${ROOT_DIR}/discord-rpc")
```
janisozaur, line 163: This will only get check in case the if in line 156, which currently reads: `if (UNIX AND NOT APPLE)` fails, meaning all linux builds would require to have the package installed system-wide.

janisozaur (changes requested):
It seems the conditions are a bit misplaced to work as described in the comment above

CMakeLists.txt
Mailaender: Should be fixed now.

janisozaur (approved):

janisozaur:
I will keep this open for one more day in case @adam-bloom wants to comment

adam-bloom:
@janisozaur go for it. I’ll need to look at something similar for macOS as a part of #13682...if I’m able to come up with something else, I’ll reach out here to get someone to test it on Linux. I won’t be able to get to that for a few days. Thanks @Mailaender!

adam-bloom:
@Mailaender I ended up copying the duktapeConfig.cmake approach to find the library on macOS (when we use a pre-packaged dependency). Does this also work on linux? If so, we can clean up. See https://github.com/adam-bloom/OpenRCT2/commit/03955eabbbe5e65e881f3fe869803ada5fb5e49b

Mailaender:
As it is contained in an `elseif (APPLE AND MACOS_USE_DEPENDENCIES)` it shouldn't disturb Linux.

adam-bloom:
@Mailaender sorry, my question wasn’t clear. My question was: can you see if that will also find a system installed version on Linux? If so, we can clean this up a bit and remove one of the conditional sections.

Mailaender:
No, `find_package` did not work in my tests. The library contains too few metadata.

adam-bloom:
@Mailaender theres a new config file added to assist discovery in that commit. It follows the same pattern that was already in place for finding duktape (another dependency that doesn’t have pkgconfig)

PR #13659, opened 2020-12-29
Change standard palette colours for indices 1-9

Gymnasiast:


Gymnasiast:
@X123M3-256 Does this look good to you?

X123M3-256:
@Gymnasiast It looks fine to me. The colors I need are just 1-3, I'm not sure what the others do but I don't think it does any harm to make sure they also have a unique color.

tupaschoal (commented):
Looks ok, I guess, I'm a bit clueless as to what this is about

PR #13661, opened 2020-12-30
Close #13627: Refactor ItemTypeProperty to use strong enum

### LINKED ISSUE #13623: Refactor LightFXQualifier to use strong enum

### LINKED ISSUE #13627: Refactor ItemTypeProperty to use strong enum

ddevrien:
Close #13627: Refactor ItemTypeProperty to use strong enum
~~Close #13623: Refactor LightFXQualifier to use strong enum~~

The enum ItemTypeProperty was only being used to determine if a NewsItem Type has a subject and/or a location. There already exists a function TypeHasSubject() and TypeHasLocation() so I refactored those to return true for the correct Type. The enum becomes obsolete this way.

~~I also changed LightFXQualifier to a strong enum, same PR since it required no refactoring.~~

### DIFF
--- a/src/openrct2/management/NewsItem.cpp
+++ b/src/openrct2/management/NewsItem.cpp
@@ -19,7 +19,6 @@
 #include "../localisation/Localisation.h"
 #include "../management/Research.h"
 #include "../ride/Ride.h"
-#include "../util/Util.h"
 #include "../windows/Intent.h"
 #include "../world/Location.hpp"
 #include "../world/Sprite.h"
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
Undo the news change the previous version made more sense. Don't combine two unrelated refactors into one commit and preferably don't combine them into one pr

tupaschoal (commented):
The `LightFXQualifier ` one was done and merged in #13684.

Please rebase and rework the `ItemTypeProperty` change according to Duncan's comments

ddevrien:
Updated according to Duncan's comments.

duncanspumpkin:
Hmm these are just flags. In general we have kept flags as c style enums. @tupaschoal what do you think

tupaschoal (commented):
At this final stage we can realize this enum is actually meant to be a flag, so we don't really need to use `enum class`. I'll be closing the PR while we think on the final solution, sorry for the missed changes.

PR #13664, opened 2020-12-30
Move tile element banner functions to methods

Gymnasiast:


ZehMatt (approved):

src/openrct2/world/TileElement.h
```diff
@@ -183,6 +183,10 @@ struct TileElement : public TileElementBase
     void ClearAs(uint8_t newType);
 
     ride_id_t GetRideIndex() const;
+
```
duncanspumpkin, line 186: Why not add these to tileElementBase? The way I see it is that tileElement is just a catch all struct for when we aren't sure of the type and it should just be used to convert to the correct type.

Gymnasiast: TileElementBase does not have AsTrack(), AsBanner and friends, so I can't put them there.

PR #13665, opened 2020-12-30
Fix #11438: Crash when decreasing map size

### LINKED ISSUE #11438: Shrinking map freezes game
I was fucking around resizing an old savegame, and then the game froze with a loud repeating sound that may or may not be the error message sound. I tried singling out what caused the problem by isolating small scenery, large scenery and rides around the border, I couldn't figure out a pattern.

See for yourself, try resizing the map to 123x123. That's when it happens.

[game freezes.zip](https://github.com/OpenRCT2/OpenRCT2/files/4528951/game.freezes.zip)

May or may not be related to https://github.com/OpenRCT2/OpenRCT2/issues/9955.

Not quite an urgent thing for me, but I can imagine you want this bug gone.

Using version 0.2.6 6c3c857 on windows 8.

Gymnasiast:
The issue was caused by the code trying to remove a park entrance element that was not one that it knew about. It therefore refused to remove the tile element, caused code further upstream to get into an infinite loop.

Since `clear_element_at` is only called from functions that either clear the map completely or clear all the out of range elements, I figured that forcibly removing them was not too much of a problem.

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -1738,28 +1738,48 @@ static void clear_element_at(const CoordsXY& loc, TileElement** elementPtr)
                     break;
             }
             auto parkEntranceRemoveAction = ParkEntranceRemoveAction(CoordsXYZ{ seqLoc, element->GetBaseZ() });
-            GameActions::Execute(&parkEntranceRemoveAction);
+            auto result = GameActions::Execute(&parkEntranceRemoveAction);
+            // If asking nicely did not work, forcibly remove this to avoid an infinite loop.
+            if (result->Error != GameActions::Status::Ok)
+            {
+                tile_element_remove(element);
+            }
             break;
         }
         case TILE_ELEMENT_TYPE_WALL:
         {
             CoordsXYZD wallLocation = { loc.x, loc.y, element->GetBaseZ(), element->GetDirection() };
             auto wallRemoveAction = WallRemoveAction(wallLocation);
-            GameActions::Execute(&wallRemoveAction);
+            auto result = GameActions::Execute(&wallRemoveAction);
+            // If asking nicely did not work, forcibly remove this to avoid an infinite loop.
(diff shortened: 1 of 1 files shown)


IntelOrca:
If calling `tile_element_remove` works, why do we need to execute game actions at all?

Gymnasiast:
Because they also do stuff like cleaning up banner entries and removing park entrances from the global list. `tile_element_remove()` is used as a last resort.

ZehMatt:
How does it not know that the entrance exists? Seems like an issue that is not handled correctly elsewhere.

Gymnasiast:
There are countless reasons why the tile elements can be incorrect. One of them is simply copying stuff using the Tile Inspector.

ZehMatt:
I'm not a big fan about having a forced removal for all the other elements, given the entrance is not registered is the fault of the tile inspector, the removal logic is correct.

IntelOrca:
All code should be crash proof

Gymnasiast:
As I said, there are more reasons why it might be incorrect. Corruption of tile elements is very common and I'm frankly a bit surprised that you seem unaware of this.

ZehMatt:
I'm aware that things can get corrupted but that is exactly my point, fighting the code that corrupts the state rather than hiding it. I understand to why the park entrance can get into this situation but the cause is apparently known so we should fix it there. As for the other corruptions we need to identify the cause. In my experience attempts like this usually shift the issue elsewhere, that is why I think the fix is fine for the entrance only but not for the rest.

IntelOrca:
@ZehMatt it needs to handle all edge cases regardless of how they may have been introduced

ZehMatt:
I think it would be better to adjust the game actions in that regard, one thing that comes to my mind is have a flag that skips most of the things in query, the code needs to check of course for this specific flag. Its somewhat close to the clearance check. This is a cleaner solution and the plugins could also benefit to apply the specific flag to skip certain checks.

PR #13666, opened 2020-12-31
Refactor MISC_SPRITE to strong enum

ZehMatt:
I had to keep the type of ```type``` as its has a different representation based on the entity type.

### DIFF
--- a/src/openrct2/paint/sprite/Paint.Misc.cpp
+++ b/src/openrct2/paint/sprite/Paint.Misc.cpp
@@ -33,9 +33,9 @@ void misc_paint(paint_session* session, const SpriteBase* misc, int32_t imageDir
 {
     rct_drawpixelinfo* dpi = &session->DPI;
 
-    switch (misc->type)
+    switch (static_cast<MiscSpriteType>(misc->type))
     {
-        case SPRITE_MISC_STEAM_PARTICLE: // 0
+        case MiscSpriteType::SteamParticle: // 0
         {
             auto particle = misc->As<SteamParticle>();
             if (particle == nullptr)
@@ -45,7 +45,7 @@ void misc_paint(paint_session* session, const SpriteBase* misc, int32_t imageDir
             break;
         }
 
-        case SPRITE_MISC_MONEY_EFFECT: // 1
+        case MiscSpriteType::MoneyEffect: // 1
         {
             if (dpi->zoom_level > 0)
             {
@@ -61,7 +61,7 @@ void misc_paint(paint_session* session, const SpriteBase* misc, int32_t imageDir
             break;
         }
 
-        case SPRITE_MISC_CRASHED_VEHICLE_PARTICLE: // 2
+        case MiscSpriteType::CrashedVehicleParticle: // 2
         {
             if (dpi->zoom_level > 0)
             {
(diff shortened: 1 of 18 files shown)


duncanspumpkin:
This is pretty similar to the window code that @tupaschoal and myself were discussing. For the window code we created a union with the idea being that the field is eventually split in a future refactor. I feel we should be consistent with that.

ZehMatt:
I can't forward declare nested types, I initially attempted that but its better to split the variable later.

src/openrct2/world/MoneyEffect.cpp
```diff
@@ -43,7 +43,7 @@ void MoneyEffect::CreateAt(money32 value, const CoordsXYZ& effectPos, bool verti
     moneyEffect->sprite_height_positive = 30;
     moneyEffect->sprite_identifier = SpriteIdentifier::Misc;
     moneyEffect->MoveTo(effectPos);
-    moneyEffect->type = SPRITE_MISC_MONEY_EFFECT;
+    moneyEffect->type = EnumValue(MiscSpriteType::MoneyEffect);
```
Gymnasiast, line 46: Can we declare `type` as a MiscSpriteType instead?

Gymnasiast:
Since we renamed sprites to entities, `MiscEntityType` would be a more appropriate name.

src/openrct2/world/MoneyEffect.cpp
ZehMatt: Vehicle uses this as Vehicle::Type, thats what I meant by splitting up the variable, Vehicle should have its own type and Misc should have its own type.

src/openrct2/paint/sprite/Paint.Misc.cpp
```diff
@@ -134,7 +134,9 @@ void misc_paint(paint_session* session, const SpriteBase* misc, int32_t imageDir
                 isAntiClockwise = !isAntiClockwise;
             }
 
-            uint32_t baseImageId = (jumpingFountain->type == SPRITE_MISC_JUMPING_FOUNTAIN_SNOW) ? 23037 : 22973;
+            uint32_t baseImageId = (static_cast<MiscEntityType>(jumpingFountain->type) == MiscEntityType::JumpingFountainSnow)
+                ? 23037
```
Gymnasiast, line 138: These naked numbers should get a constant. Like `SPR_JUMPING_FOUNTAIN_WATER_START` and `SPR_JUMPING_FOUNTAIN_SNOW_START`.

src/openrct2/world/MoneyEffect.cpp
```diff
@@ -1,4 +1,4 @@
-/*****************************************************************************
+﻿/*****************************************************************************
```
Gymnasiast, line 1: *cough*

src/openrct2/world/Sprite.cpp
```diff
@@ -1,4 +1,4 @@
-/*****************************************************************************
+﻿/*****************************************************************************
```
Gymnasiast, line 1: Again.

Can you please stop your IDE from inserting these silly BOMs, which should not be present in UTF-8 files?

ZehMatt: I have no bloody idea how it keeps doing this, I always put it back but at some point it just screws me over again.

src/openrct2/paint/sprite/Paint.Misc.cpp
ZehMatt: I'll do that in another PR, there are more naked numbers in this file like that.

duncanspumpkin:
What about a getter function instead of the casts?

Gymnasiast (approved):

ZehMatt:
> What about a getter function instead of the casts?

The base class has no idea about how the type member is treated.

PR #13667, opened 2020-12-31
Simplify object invalidation automatically when possible

ZehMatt:
I refactored out the specific invalidation methods ```Invalidate0, Invalidate1, Invalidate2```, I unified the invalidation in a single routine that selects the correct invalidation as some entities are not drawn on a specific zoom level, this just guards useless invalidations for when the object is not visible at a specific zoom level. The other change is to automatically apply invalidation when the sprites move, Peep already had the MoveTo specialized to do exactly that, I moved that into SpriteBase as every sprite that uses MoveTo has to invalidate the old and new position, this allowed also to eliminate a lot of boilerplate code.

I touched some part of the scripting as the CI complained about the missing cases, I updated the documentation accordingly, it had a few types missing while its possible to obtain them I believe.

This PR is based on #13666

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -730,7 +730,9 @@ static void SpriteSpatialMove(SpriteBase* sprite, const CoordsXY& newLoc)
 }
 
 /**
- * Moves a sprite to a new location.
+ * Moves a sprite to a new location, invalidates the current position if valid
+ * and also the new position.
+ *
  *  rct2: 0x0069E9D3
  *
  * @param x (ax)
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -1,4 +1,4 @@
-/*****************************************************************************
+﻿/*****************************************************************************
  * Copyright (c) 2014-2020 OpenRCT2 developers
  *
  * For a complete list of all authors, please refer to contributors.md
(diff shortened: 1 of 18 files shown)


src/openrct2/world/Sprite.cpp
```diff
@@ -133,31 +133,54 @@ static void invalidate_sprite_max_zoom(SpriteBase* sprite, int32_t maxZoom)
...
+        case SpriteIdentifier::Peep:
+            maxZoom = 0;
+            break;
+        case SpriteIdentifier::Misc:
+            switch (static_cast<MiscSpriteType>(type))
+            {
+                case MiscSpriteType::CrashedVehicleParticle:
+                    [[fallthrough]];
+                case MiscSpriteType::JumpingFountainWater:
+                    [[fallthrough]];
```
duncanspumpkin, line 153: my understanding of when you use [[fallthrough]] would be if you wanted to perform a bit of work and then fall through to the next case and not for just bunching together cases like this.

```
case 0:
   i++;
   [[fallthrough]]
case 1:
   std::cout << i;
   break;
```

src/openrct2/world/Sprite.cpp
```diff
@@ -728,6 +755,10 @@ void SpriteBase::MoveTo(const CoordsXYZ& newLocation)
     {
         loc.x = LOCATION_NULL;
     }
+    else
```
duncanspumpkin, line 758: Why the else? If we are picking up a peep we move it to a LOCATION_NULL we would still want to invalidate its origin.

src/openrct2/world/Sprite.cpp
```diff
@@ -742,6 +773,8 @@ void SpriteBase::MoveTo(const CoordsXYZ& newLocation)
     {
         sprite_set_coordinates(loc, this);
     }
+
+    Invalidate(); // Invalidate new position.
```
duncanspumpkin, line 777: This invalidate could go in the else

src/openrct2/world/Sprite.cpp
```diff
@@ -847,7 +880,7 @@ void litter_create(const CoordsXYZD& litterPos, int32_t type)
     litter->sprite_identifier = SpriteIdentifier::Litter;
     litter->type = type;
     litter->MoveTo(offsetLitterPos);
```
duncanspumpkin, line 882: Are you going to go through every MoveTo and remove the invalidates?

src/openrct2/world/Sprite.cpp
ZehMatt: I remember a time where it complained, I can try it tho.

src/openrct2/world/Sprite.cpp
ZehMatt: Oh good catch, I misread the code a bit, I thought this is checking the old position but that makes no sense. I'll change it to what I thought: Invalidate if previous pos is valid and Invalidate when the new location is valid.

src/openrct2/world/Sprite.cpp
ZehMatt: Yeah I missed that one.

duncanspumpkin (approved):

Issue #13621, opened 2020-12-22
Increased lag makes large parks borderline unplayable


LordMarcel:
Over the past few years I have noticed the amount of lag in OpenRCT2 kept increasing and now it's come to a point where large parks have become borderline unplayable because of the lag. Yesterday I had to revert to version 0.2.5 (9 months old) to be able to livestream a large park without severe lag.

I have made this video to compare the amount of lag in four different OpenRCT2 builds: https://www.youtube.com/watch?v=tfpkhR_kfzg

If this trend continues, even medium sized parks of 3000 or 4000 guests will become very laggy and larger parks will become entirely unplayable. If we get an increased sprite limit after the custom save format is done, it won't be of much use since the lag will make it unplayable in 0.3.2 already.

This is a huge issue that has to be addressed.

James103:
There is a huge jump in lag from 0.2.5 to 0.3.2, but there are also several release versions in between (0.2.6, 0.3.0, and 0.3.1). Do you know the exact commit or release that causes OpenRCT2 to reduce in framerates severalfold, or is it just accumulated lag increases from many commits in between?

duncanspumpkin:
Why didn't you leave the fps on so that we could have some more empirical results. What graphics rendering are you using are you comparing that as well?

LordMarcel:
I could not test 0.2.6 since none of my big parks would load on that build, but all the others worked fine. The largest jump by far was between 0.3.0 and 0.3.1. There have of course been many small development builds between those two and I don't know on which of those the lag increased the most.

There is also quite a large jump between 0.2.0 and 0.2.5, which is especially noticable with the small park at x128 speed. The largest jumps here seem to be between 0.2.1 and 0.2.2, and between 0.2.4 and 0.2.5.

LordMarcel:
I have used 'software (hardware display)' for 0.1.0 (because OpenGL gave glitches) and OpenGL on all the others. I have tried switching between the two several times and never notice any difference in the amount of lag. I suspect that this is because this lag is not caused by the rendering but rather by the number of calculations that the game has to do for all the guests.

I forgot that having the fps on was a thing, apologies.

duncanspumpkin:
From my tests its the path finding that is the most expensive function when at faster speeds

csunday95:
Hey marcel, just saw the video. A similar test that would be really useful with regard to duncan's findings would be:

[...]

PR #13668, opened 2020-12-31
Avoid unnecessary calculations when the passed DPI is the active DPI

ZehMatt:
This squeezed out +8 FPS on bpb.sv6 at 1080p, its not a lot but its something.


src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -1,4 +1,4 @@
-/*****************************************************************************
+﻿/*****************************************************************************
```
Gymnasiast, line 1: Looks like you inserted a BOM here.

Gymnasiast (approved):

Gymnasiast:
It might not be a lot, but it is certainly welcome. Especially since lag seems to have crept in: https://github.com/OpenRCT2/OpenRCT2/issues/13621

Issue #9631, opened 2019-07-23
Text on banners and entrance buildings glitches


LordMarcel:
**OS:** Windows 10
**Version:** 0.2.3
**Commit/Build:** ab5f5b2

The text on banners, the park entrance, and entrance buildings sometimes randomly glitches like it does in the video I linked below. There seems to be no pattern to when this happens and sometimes it will stop for a few seconds and then start again.

**Screenshots / Video:**
https://www.youtube.com/watch?v=es0C28Z_vcQ

ZehMatt:
Are you using the multithreading option?

LordMarcel:
Yes, although disabling it doesn't fix it, it rather just reduces the
frequency at which it glitches (from about 10 times per second to 2 or so).

Broxzier:
Possibly a regression from #9589? I cannot reproduce this on my PC. Could you try it with a build from before that PR was merged @LordMarcel?

LordMarcel:
I tried a build from a while ago (0.2.2 ba040fc). I'm not sure how old it is but definitely older than 7 days. It's still there in the exact same way.

Gymnasiast:
Which rendering engine are you on? OpenGL or one of the software renderers?

LordMarcel:
I use the software renderer, but when I switch to software (hardware display) and OpenGL the glitching stays.

PR #13669, opened 2021-01-01
Refactor logic to allow to draw 256 banners at the same time

ZehMatt:
From this:

To this:


Gymnasiast:
@LordMarcel Can you check if this solves your problem in #9631 ?

ZehMatt:
I suspect this is the same issue

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -180,6 +180,7 @@ void mask_scalar(
 static rct_gx _g1 = {};
 static rct_gx _g2 = {};
 static rct_gx _csg = {};
+static rct_g1_element _scrollingText[256]{};
```
Gymnasiast, line 183: Shouldn't this be
```suggestion
static rct_g1_element _scrollingText[MAX_SCROLLING_TEXT_ENTRIES]{};
```
?

src/openrct2/drawing/ScrollingText.cpp
```diff
@@ -34,7 +34,7 @@ struct rct_draw_scroll_text
     uint8_t bitmap[64 * 40];
 };
 
-constexpr int32_t MAX_SCROLLING_TEXT_ENTRIES = 32;
+constexpr int32_t MAX_SCROLLING_TEXT_ENTRIES = SPR_SCROLLING_TEXT_END - SPR_SCROLLING_TEXT_START;
```
Gymnasiast, line 37: Normally, we would set this to a fixed value and define `SPR_SCROLLING_TEXT_END` as `SPR_SCROLLING_TEXT_START + 256`.

src/openrct2/drawing/Drawing.Sprite.cpp
ZehMatt: That constant is defined in ScrollingText.cpp

src/openrct2/drawing/ScrollingText.cpp
ZehMatt: I agree, however there is no header file specific for banners/scrolling text, Its a bit out of place to define it in sprites.h so I left it alone.

src/openrct2/drawing/Drawing.Sprite.cpp
Gymnasiast: That's no good reason to hardcode it here. It completely defeats the point of having a constant in the first place.

ZehMatt: Where do you suggest I define the constant? It would be required in sprites.h and ScrollingText.cpp

Gymnasiast: I suggest we ask Ted or Duncan. I would like one of them to review the PR anyway.

ZehMatt: I've introduced a header file called GameLimits.h, been wanting to do that for a while now. I think this simplifies such cases and makes it easier to control the compilation out of one place.

src/openrct2/drawing/ScrollingText.cpp
```diff
@@ -97,10 +96,11 @@ void scrolling_text_initialise_bitmaps()
...
     }
 
-    for (int32_t i = 0; i < MAX_SCROLLING_TEXT_ENTRIES; i++)
+    for (int32_t i = 0; i < OpenRCT2::MaxScrollingTextEntries; i++)
     {
-        int32_t imageId = SPR_SCROLLING_TEXT_START + i;
-        const rct_g1_element* g1original = gfx_get_g1_element(imageId);
+        const int32_t imageIdReference = SPR_SCROLLING_TEXT_LEGACY_START;
+        const int32_t imageId = SPR_SCROLLING_TEXT_START + i;
+        const rct_g1_element* g1original = gfx_get_g1_element(imageIdReference);
```
duncanspumpkin, line 103: This is always getting the same element why is it in the loop? Remove it from the loop and also its nullptr check. I don't quite understand why you are doing it like this. Is there some specific reason you get a different g1 element overwrite most of its elements and then place it in an image location? Wouldn't it make more sense to build a new g1 element and then write it to the `imageId`?

duncanspumpkin:
For my understanding of reading this you are removing the limit by ignoring the built in 32 entries and making a new 256 entry patch at a different location. Is that correct?

ZehMatt:
> For my understanding of reading this you are removing the limit by ignoring the built in 32 entries and making a new 256 entry patch at a different location. Is that correct?

Yes, I use the 256 ids before the dynamic image table, I wasn't sure if using it anywhere else would be safe.

src/openrct2/drawing/ScrollingText.cpp
ZehMatt: Not copying the original sprite(s) caused black artifacts, but yes having it in the loop is pointless I just need the data as reference.

duncanspumpkin: Might as well work out what the other fields you are copying are and remove the requirement to copy

src/openrct2/GameLimits.h
```diff
@@ -0,0 +1,19 @@
+/*****************************************************************************
```
Gymnasiast, line 1: Calling it GameLimits.h is too broad - there are dozens, if not hundreds of such limits. I would call it `src/openrct2/drawing/ScrollingText.h`.

ZehMatt: The purpose of this file is to be precisely that, broad. We should move all technical limits in there, right now its way too scattered.

src/openrct2/drawing/ScrollingText.cpp
```diff
@@ -98,23 +98,25 @@ void scrolling_text_initialise_bitmaps()
...
-            gfx_set_g1_element(imageId, &g1);
-        }
+
+        // Initialize the scrolling text sprite.
+        rct_g1_element g1{};
+        g1.offset = _drawScrollTextList[i].bitmap;
+        g1.x_offset = -32;
+        g1.y_offset = 0;
+        g1.flags = G1_FLAG_BMP;
+        g1.zoomed_offset = 766;
```
duncanspumpkin, line 109: Define an spr constant for this

src/openrct2/GameLimits.h
Gymnasiast: I have to disagree here .What we’ll end up with is a mega file that gets impossible to read. There are hundreds of limits.

ZehMatt: If there are so many limits its even more reason to have a central location for them and not have hundreds of constants scattered in my opinion.

IntelOrca:
You are reducing 256 images for objects to use now. You might be able to get a much larger limit by swapping out a single allocated g2 entry with a different dynamic set before drawing each banner.

src/openrct2/GameLimits.h
Gymnasiast: It will only complicate matters and make every part of the code tied to it IMO.

Gymnasiast (changes requested):
Please change the GameLimits.h file into ScrollingText.h.

ZehMatt:
> You are reducing 256 images for objects to use now. You might be able to get a much larger limit by swapping out a single allocated g2 entry with a different dynamic set before drawing each banner.

The current limit of images is 0x7FFFF (524287) which is half a million, I don't think the 256 gonna hurt that much. Adding more complexity for something like this is not feasible in my opinion.

duncanspumpkin:
I also think the reduction is not an issue we are decreasing the limit by way more than that in the RMC PR's anyway. We need to start thinking more about how we can move that limit up to 32bit anyway.

IntelOrca:
> I also think the reduction is not an issue we are decreasing the limit by way more than that in the RMC PR's anyway. We need to start thinking more about how we can move that limit up to 32bit anyway.

We need everywhere to use ImageId so that we can increase the ID mask.

Gymnasiast (approved):
All my requests have been fulfilled. I don't know if taking away 256 images is acceptable or not, so I'll leave that discussion to you, Duncan and Ted.

IntelOrca (approved):

PR #13671, opened 2021-01-01
Revert #13668

### LINKED ISSUE #13670: Dropdowns are broken in OpenGL mode
**OS:** Ubuntu 20.04
**Version:** 0.3.2+
**Commit/Build:** eda01dc6089a430bb3dddf5360f0b8441b0933be

afbeelding

Caused by eda01dc6089a430bb3dddf5360f0b8441b0933be . 
@ZehMatt

ZehMatt:
Small oversight, when the DPI sits on the stack the pointer can be identical while the data changed, so the check would cause false positives.

Closes #13670


Gymnasiast:
Since this inserted a BOM, I created my own: https://github.com/OpenRCT2/OpenRCT2/pull/13672

Issue #13679, opened 2021-01-03
Allow progress tracking of scenario objectives in the park menu


JonasKlose:
I've had a problem with the RCT2 scenario "Ghost Town" as I wasn't sure how the rules of the game worked.
The requirement was to build 10 roller coasters of different types that all were 1200 units long.
However I thought that doing multiple laps would increase the length of the rides like it does with other stats.

But I wasn't sure and saw no indicator if I managed to do any progress on the scenario or not.
So I looked into the source to find the exact requirements.

The proposal is to include a progress tracker in the main park ui.

OpenRCT2, v3.2.0

Makke94:
I was thinking of handling this issue, but I'll probably take care of #13792 and refactor the park window first as I figure that's where the progress will be shown.

PR #13681, opened 2021-01-03
Proposal for issue #13679: Allow progress tracking of scenario objectives in the park menu

JonasKlose:
- Added progress tracking in Park.cpp
- Moved duplicate code in Scenario.cpp in functions
- Added progress tracking UI strings for en-GB and de-DE
- Allow to print UINT32 numbers in the game UI
- Merge changes with current dev branch

### DIFF
--- a/data/language/de-DE.txt
+++ b/data/language/de-DE.txt
@@ -3648,6 +3648,15 @@ STR_6396    :Bildschirmschoner und -stromsparen deaktivieren
 STR_6397    :Falls ausgewählt, werden der Bildschirmschoner und andere Bildschirmstromsparfunktionen blockiert, solange OpenRCT2 läuft.
 STR_6398    :Datei enthält nicht unterstützte Fahrgeschäfttypen. Bitte aktualisieren Sie zu einer neueren Version von OpenRCT2.
 STR_6399    :OpenRCT2 benötigt Dateien vom Originalspiel RollerCoaster Tycoon 2, um zu funktionieren. Bitte setzen Sie die „game_path“-Variable in config.ini auf das Verzeichnis, in das Sie RollerCoaster Tycoon 2 installiert haben, dann starten Sie OpenRCT2 neu.
+STR_6400    :{WINDOW_COLOUR_2}Ihr Fortschritt:
+STR_6401    :{BLACK}-
+STR_6402    :{BLACK}Gäste: {UINT32} von {COMMA16}
+STR_6403    :{BLACK}Bewertung: {COMMA16} von {INT32}
+STR_6404    :{BLACK}Parkwert: {CURRENCY} von {CURRENCY}
+STR_6405    :{BLACK}Achterbahnen: {INT32} von {INT32}
+STR_6406    :{BLACK}Fahrteinnahmen: {CURRENCY} von {CURRENCY}
+STR_6407    :{BLACK}Darlehen: {CURRENCY} von {CURRENCY}
+STR_6408    :{BLACK}Gesamtverkäufe: {CURRENCY} von {CURRENCY}
 
 #############
 # Scenarios #
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3651,6 +3651,15 @@ STR_6396    :Disable screensaver and monitor power saving
 STR_6397    :If checked, screensaver and other monitor power saving features will be inhibited while OpenRCT2 is running.
 STR_6398    :File contains unsupported ride types. Please update to a newer version of OpenRCT2.
 STR_6399    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work. Please set the “game_path” variable in config.ini to the directory where you installed RollerCoaster Tycoon 2, then restart OpenRCT2.
+STR_6400    :{WINDOW_COLOUR_2}Your Progress:
+STR_6401    :{BLACK}-
+STR_6402    :{BLACK}Guests: {UINT32} of {COMMA16}
+STR_6403    :{BLACK}Rating: {COMMA16} of {INT32}
+STR_6404    :{BLACK}Park value: {CURRENCY} of {CURRENCY}
+STR_6405    :{BLACK}Roller coasters: {INT32} of {INT32}
+STR_6406    :{BLACK}Ride income: {CURRENCY} of {CURRENCY}
+STR_6407    :{BLACK}Current loan: {CURRENCY} of {CURRENCY}
+STR_6408    :{BLACK}Combined income: {CURRENCY} of {CURRENCY}
 
 #############
 # Scenarios #
--- a/src/openrct2-ui/windows/Park.cpp
+++ b/src/openrct2-ui/windows/Park.cpp
@@ -1394,10 +1394,10 @@ static void window_park_objective_resize(rct_window* w)
 {
 #ifndef NO_TTF
     if (gCurrentTTFFontSet != nullptr)
-        window_set_resize(w, 230, 270, 230, 270);
+        window_set_resize(w, 230, 300, 230, 300);
     else
 #endif
-        window_set_resize(w, 230, 226, 230, 226);
+        window_set_resize(w, 230, 250, 230, 250);
 }
 
 /**
@@ -1493,6 +1493,131 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     screenCoords.y += gfx_draw_string_left_wrapped(
         dpi, ft.Data(), screenCoords, 221, ObjectiveNames[gScenarioObjective.Type], COLOUR_BLACK);
+    screenCoords.y += 5;
+
+    // Your progress
+    gfx_draw_string_left(dpi, STR_OBJECTIVE_PROGRESS_LABEL, nullptr, COLOUR_BLACK, screenCoords);
+    screenCoords.y += LIST_ROW_HEIGHT;
+
+    // Progress
+    switch (gScenarioObjective.Type)
+    {
+        case OBJECTIVE_NONE:
+        case OBJECTIVE_HAVE_FUN:
+        case OBJECTIVE_BUILD_THE_BEST:
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, nullptr, screenCoords, 221, STR_OBJECTIVE_PROGRESS_NONE, COLOUR_BLACK);
+            break;
+        case OBJECTIVE_GUESTS_BY:
+        case OBJECTIVE_GUESTS_AND_RATING:
+        {
+            ft = Formatter();
+            ft.Add<uint32_t>(gNumGuestsInPark);
+            ft.Add<uint16_t>(gScenarioObjective.NumGuests);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_GUESTS, COLOUR_BLACK);
+            screenCoords.y += 1;
+
+            ft = Formatter();
+            ft.Add<uint16_t>(gParkRating);
+
+            if (gScenarioObjective.Type == OBJECTIVE_GUESTS_BY)
+            {
+                ft.Add<int32_t>(600);
+            }
+            else
+            {
+                ft.Add<int32_t>(700);
+            }
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_RATING, COLOUR_BLACK);
+
+            break;
+        }
+        case OBJECTIVE_PARK_VALUE_BY:
+        {
+            ft = Formatter();
+            ft.Add<money32>(gParkValue);
+            ft.Add<money32>(gScenarioObjective.Currency);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_PARK_VALUE, COLOUR_BLACK);
+
+            break;
+        }
+        case OBJECTIVE_10_ROLLERCOASTERS:
+        {
+            ft = Formatter();
+            ft.Add<int32_t>(gScenarioObjective.Get10RollerCoastersProgress());
+            ft.Add<int32_t>(10);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_ROLLER_COASTER_COUNT, COLOUR_BLACK);
+            break;
+        }
+        case OBJECTIVE_MONTHLY_RIDE_INCOME:
+        {
+            ft = Formatter();
+            ft.Add<money32>(gScenarioObjective.GetMonthlyRideIncomeProgress());
+            ft.Add<money32>(gScenarioObjective.Currency);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_MONTHLY_RIDE_INCOME, COLOUR_BLACK);
+            break;
+        }
+        case OBJECTIVE_10_ROLLERCOASTERS_LENGTH:
+        {
+            ft = Formatter();
+            ft.Add<int32_t>(gScenarioObjective.Get10RollerCoastersLengthProgress());
+            ft.Add<int32_t>(10);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_ROLLER_COASTER_COUNT, COLOUR_BLACK);
+            break;
+        }
+        case OBJECTIVE_FINISH_5_ROLLERCOASTERS:
+        {
+            ft = Formatter();
+            ft.Add<int32_t>(gScenarioObjective.GetFinish5RollerCoastersProgress());
+            ft.Add<int32_t>(5);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_ROLLER_COASTER_COUNT, COLOUR_BLACK);
+            break;
+        }
+        case OBJECTIVE_REPAY_LOAN_AND_PARK_VALUE:
+        {
+            ft = Formatter();
+            ft.Add<money32>(gBankLoan);
+            ft.Add<money32>(0);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_LOAN, COLOUR_BLACK);
+            screenCoords.y += 1;
+
+            ft = Formatter();
+            ft.Add<money32>(gParkValue);
+            ft.Add<money32>(gScenarioObjective.Currency);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_PARK_VALUE, COLOUR_BLACK);
+            break;
+        }
+        case OBJECTIVE_MONTHLY_FOOD_INCOME:
+        {
+            ft = Formatter();
+            ft.Add<money32>(gScenarioObjective.GetMonthlyShopIncomeProgress());
+            ft.Add<money32>(gScenarioObjective.Currency);
+
+            screenCoords.y += gfx_draw_string_left_wrapped(
+                dpi, ft.Data(), screenCoords, 221, STR_OBJECTIVE_PROGRESS_SHOP_INCOME, COLOUR_BLACK);
+            break;
+        }
+        default:
+            break;
+    }
     screenCoords.y += 5;
 
     // Objective outcome
--- a/src/openrct2/localisation/FormatCodes.cpp
+++ b/src/openrct2/localisation/FormatCodes.cpp
@@ -35,6 +35,7 @@ static const std::unordered_map<std::string_view, FormatToken> FormatTokenMap = 
     { "COMMA2DP32",           FormatToken::Comma2dp32,          },
     { "COMMA16",              FormatToken::Comma16,             },
     { "UINT16",               FormatToken::UInt16,              },
+    { "UINT32",               FormatToken::UInt32,              },
     { "CURRENCY2DP",          FormatToken::Currency2dp,         },
     { "CURRENCY",             FormatToken::Currency,            },
     { "STRINGID",             FormatToken::StringId,            },
@@ -119,6 +120,7 @@ bool FormatTokenTakesArgument(FormatToken token)
         case FormatToken::Comma2dp32:
         case FormatToken::Comma16:
         case FormatToken::UInt16:
+        case FormatToken::UInt32:
         case FormatToken::Currency2dp:
         case FormatToken::Currency:
         case FormatToken::StringId:
--- a/src/openrct2/localisation/FormatCodes.h
+++ b/src/openrct2/localisation/FormatCodes.h
@@ -33,6 +33,7 @@ enum class FormatToken
     Comma2dp32,
     Comma16,
     UInt16,
+    UInt32,
     Currency2dp,
     Currency,
     StringId,
--- a/src/openrct2/localisation/Formatting.cpp
+++ b/src/openrct2/localisation/Formatting.cpp
@@ -373,7 +373,7 @@ namespace OpenRCT2
         } while (num != 0 && i < sizeof(buffer));
 
         // Finally reverse append the string
-        for (int32_t j = static_cast<int32_t>(i - 1); j >= 0; j--)
+        for (int64_t j = static_cast<int64_t>(i - 1); j >= 0; j--)
         {
             ss << buffer[j];
         }
@@ -483,6 +483,7 @@ namespace OpenRCT2
         switch (token)
         {
             case FormatToken::UInt16:
+            case FormatToken::UInt32:
             case FormatToken::Int32:
                 if constexpr (std::is_integral<T>())
                 {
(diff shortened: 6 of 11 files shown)


data/language/de-DE.txt
```diff
@@ -3648,6 +3648,15 @@ STR_6396    :Bildschirmschoner und -stromsparen deaktivieren
 STR_6397    :Falls ausgewählt, werden der Bildschirmschoner und andere Bildschirmstromsparfunktionen blockiert, solange OpenRCT2 läuft.
 STR_6398    :Datei enthält nicht unterstützte Fahrgeschäfttypen. Bitte aktualisieren Sie zu einer neueren Version von OpenRCT2.
 STR_6399    :OpenRCT2 benötigt Dateien vom Originalspiel RollerCoaster Tycoon 2, um zu funktionieren. Bitte setzen Sie die „game_path“-Variable in config.ini auf das Verzeichnis, in das Sie RollerCoaster Tycoon 2 installiert haben, dann starten Sie OpenRCT2 neu.
+STR_6400    :{WINDOW_COLOUR_2}Ihr Fortschritt:
```
duncanspumpkin, line 3651: remove these changes. All non en-GB changes are done on the localisation repo.

src/openrct2/localisation/FormatCodes.cpp
```diff
@@ -35,6 +35,7 @@ static const std::unordered_map<std::string_view, FormatToken> FormatTokenMap =
     { "COMMA2DP32",           FormatToken::Comma2dp32,          },
     { "COMMA16",              FormatToken::Comma16,             },
     { "UINT16",               FormatToken::UInt16,              },
+    { "UINT32",               FormatToken::UInt32,              },
```
duncanspumpkin, line 38: Do not add new format codes we are working on removing these.

src/openrct2/localisation/FormatCodes.cpp
```diff
@@ -119,6 +120,7 @@ bool FormatTokenTakesArgument(FormatToken token)
         case FormatToken::Comma2dp32:
         case FormatToken::Comma16:
         case FormatToken::UInt16:
+        case FormatToken::UInt32:
```
duncanspumpkin, line 123: remove

src/openrct2/localisation/FormatCodes.h
```diff
@@ -33,6 +33,7 @@ enum class FormatToken
     Comma2dp32,
     Comma16,
     UInt16,
+    UInt32,
```
duncanspumpkin, line 36: remove

src/openrct2/localisation/Formatting.cpp
```diff
@@ -483,6 +483,7 @@ namespace OpenRCT2
         switch (token)
         {
             case FormatToken::UInt16:
+            case FormatToken::UInt32:
```
duncanspumpkin, line 486: remove

test/tests/FormattingTests.cpp
```diff
@@ -347,6 +347,14 @@ TEST_F(FormattingTests, format_number_basic)
     ASSERT_STREQ("123", ss.data());
 }
 
+TEST_F(FormattingTests, format_number_large)
```
duncanspumpkin, line 350: remove

src/openrct2/scenario/Scenario.cpp
```diff
@@ -856,35 +887,15 @@ ObjectiveStatus Objective::CheckMonthlyRideIncome() const
...
+    if (Get10RollerCoastersLengthProgress() >= 10)
     {
         return ObjectiveStatus::Success;
     }
 
     return ObjectiveStatus::Undecided;
 }
 
-ObjectiveStatus Objective::CheckFinish5RollerCoasters() const
+int Objective::GetFinish5RollerCoastersProgress() const
```
duncanspumpkin, line 898: don't use int use int32_t

duncanspumpkin:
Remove your formatting changes and use the already available formatting codes.

JonasKlose:
> Remove your formatting changes and use the already available formatting codes.

This would imply a downcast of the gNumGuestsInPark variable that is of type **uint32_t**. While I doubt that there will be OpenRCT2 with so many guests, shouldn't the variable converted to **uint16_t** too, then?

duncanspumpkin:
You would never output a raw 32bit value. You would always use the comma version. Users would expect the thousands comma and we already have the 32bit comma formatter.

JonasKlose:
> You would never output a raw 32bit value. You would always use the comma version. Users would expect the thousands comma and we already have the 32bit comma formatter.

Ah, ok. I'll try to do it that way. The comma formatter looked like it was supposed to only work with 16bit values.

JonasKlose:
Ok. I got a bit confused because my compiler didn't want to let go of one of the files.
However, I've successfully reverted all the changes to the Formatter and the connected files, merged in the current dev branch, reverted the changes to the german localization and fixed the code formatting issue as required.

I did test all of the possible changes of the park main window with my personal save games.
All of my save games opened correctly and the park window showed the progress of the scenario.

Sorry for the check fail on the commit message test. I need to learn more about proper commit messages, it seems. :$

JonasKlose:
Forgot the required switch from int to int32_t variables.

duncanspumpkin:
Okay this is looking much better. Does progress as a unit make more sense than as a percentage? I think it makes sense here but just want to see others opinions.

tupaschoal:
> Okay this is looking much better. Does progress as a unit make more sense than as a percentage? I think it makes sense here but just want to see others opinions.

I think it does, you have a direct answer of what you still need to do. As a percentage you might have to do some calculations

JonasKlose:
> Okay this is looking much better. Does progress as a unit make more sense than as a percentage? I think it makes sense here but just want to see others opinions.

For the amount of money and guests, progress bars could be a viable alternative.

For smaller progress like for roller coasters, there could be a hint that a certain number of coasters is close to completion. If the length requirement is 1200, this could count the coasters that are between 1000 and 1200, for example.

JonasKlose:
https://github.com/OpenRCT2/OpenRCT2/blob/c4c6baff36ccef5e903e981f90ca4d888fde40c8/src/openrct2-ui/windows/Park.cpp#L1393-L1401

I've expanded the window of the Park objectives window to fit the text. However I think it's a bit too large now.
Is there a way to calculate the proper window size, ideally in code? I think the size of the window will depend on the font size, the font itself, the length and width of the texts and the language of the texts.

duncanspumpkin:
Size of windows are just educated guesses usually based on en-gb size. If other languages need more space we do a follow up pr to increase the size. Some windows dynamically resize for the text but there isn't much point wasting performance on that for every window.

ocalhoun6:
> there could be a hint that a certain number of coasters is close to completion. If the length requirement is 1200, this could count the coasters that are between 1000 and 1200, for example.

That sounds like it would be getting a bit overly complicated.

IntelOrca:
Looks good so far, I think it should have percent or progress bar though. Perhaps something like:
```
Combined income:
    £400 [#####                        ] £1,000
```
using the same progress bars that are drawn on tab 2 of the guest window.

data/language/en-GB.txt
```diff
@@ -3661,6 +3661,15 @@ STR_6406    :GOG RollerCoaster Tycoon 2 Installer
 STR_6407    :This may take a few minutes.
 STR_6408    :Please install “innoextract” to extract GOG Installer, then restart OpenRCT2.
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
+STR_6410    :{WINDOW_COLOUR_2}Your Progress:
+STR_6411    :{BLACK}-
+STR_6412    :{BLACK}Guests: {COMMA32} of {COMMA16}
+STR_6413    :{BLACK}Rating: {COMMA16} of {INT32}
```
IntelOrca, line 3667: Different format types used here, should both be `COMMA16`

data/language/en-GB.txt
```diff
@@ -3661,6 +3661,15 @@ STR_6406    :GOG RollerCoaster Tycoon 2 Installer
 STR_6407    :This may take a few minutes.
 STR_6408    :Please install “innoextract” to extract GOG Installer, then restart OpenRCT2.
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
+STR_6410    :{WINDOW_COLOUR_2}Your Progress:
+STR_6411    :{BLACK}-
+STR_6412    :{BLACK}Guests: {COMMA32} of {COMMA16}
```
IntelOrca, line 3666: Different format types used here, should both be `COMMA32`

JonasKlose:
> Size of windows are just educated guesses usually based on en-gb size. If other languages need more space we do a follow up pr to increase the size. Some windows dynamically resize for the text but there isn't much point wasting performance on that for every window.

Ok. Though it is a bit strange to avoid size calculations on first draw. Even this webpage does it and no one complains about the performance. The calculation would only needed to be done on init or when there are changes in the type of rendered text.

JonasKlose:
> Looks good so far, I think it should have percent or progress bar though. Perhaps something like:
> 
> ```
> Combined income:
>     £400 [#####                        ] £1,000
> ```
> 
> using the same progress bars that are drawn on tab 2 of the guest window.
> 
> I also think it should omit the progress section altogether if there is no objective or you have completed / failed it.
> 
> There are some functions that measure string size allowing you update the window size accordingly.

I'll see what I can do about the progress bars. :)
About disabling the progress on a won/failed objective, I've thought about that, too. I'll try to implement it.

data/language/en-GB.txt
JonasKlose: The gParkRating variable is a uint16_t, while the target rating are the hard-coded numbers 600 and 700 which should be represented as int32_t. Or should I cast them to uint16_t?

data/language/en-GB.txt
JonasKlose: This is correct, according to the code. The gNumGuestsInPark variable is a uint32_t and the Objective NumGuests variable is a uint16_t. I was confused about that, too.

IntelOrca: Okay, the objective num guests will at some point become a `uint32_t`, so maybe you can cast it to int32 for now.

data/language/en-GB.txt
IntelOrca: They should both be comma separated. As for size, I'd make them both 32-bit and cast as necessary.

JonasKlose: OK, I'll do that. :)

data/language/en-GB.txt
JonasKlose: I've converted to uint32_t for now. I doubt that there'll ever be a negative number of guests and converting it to int32 would add information that isn't present in the base type.

JonasKlose:
> Looks good so far, I think it should have percent or progress bar though. Perhaps something like:
> 
> ```
> Combined income:
>     £400 [#####                        ] £1,000
> ```
> 
> using the same progress bars that are drawn on tab 2 of the guest window.
> 
> I also think it should omit the progress section altogether if there is no objective or you have completed / failed it.
> 
> There are some functions that measure string size allowing you update the window size accordingly.

The new commit adds a horizontal bar for the guest number and the park rating beneath the text. Check it out and tell me what you think about that. 😸

tupaschoal (commented):
@JonasKlose can you do us 2 favours, please?

1. Resolve every conversation where you have already changed the code according to the reviewers comments. This makes the PR look cleaner and it tells us that you've tended to them.
2. Add some screenshots showing how it's currently looking for different objectives.

Thanks!

data/language/en-GB.txt
JonasKlose: @IntelOrca 
Is this ok?

JonasKlose:
> @JonasKlose can you do us 2 favours, please?
> 
> 1. Resolve every conversation where you have already changed the code according to the reviewers comments. This makes the PR look cleaner and it tells us that you've tended to them.
> 2. Add some screenshots showing how it's currently looking for different objectives.
> 
> Thanks!

I've resolved all conversations that I think have been resolved. 😃 
Attached at this post are the screenshots of three different scenarios to show you what this branch looks like in game.

Vertigo Views Test
Ghost Town Test
Victoria Falls Test


.

James103:
Should the "Your Progress" section stay even after scenario completion (in addition to the victory/failure text if any)? Or should it be removed and replaced with the ending text when the scenario is completed/failed?

JonasKlose:
> Should the "Your Progress" section stay even after scenario completion (in addition to the victory/failure text if any)? Or should it be removed and replaced with the ending text when the scenario is completed/failed?

This was suggested earlier and I think you're right. The progress is now only visible when the scenario objective is not completed.

Umdlye:
This looks great!

The 'p' in 'progress' should probably be lowercase (like the 'o' in 'objective').

JonasKlose:
> This looks great!
> 
> The 'p' in 'progress' should probably be lowercase (like the 'o' in 'objective').

Thanks for the hint! I've updated the PR.

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1446,6 +1446,22 @@ static void window_park_objective_invalidate(rct_window* w)
...
+static void window_park_objective_bars_paint(
+    float percentage, int32_t x, int32_t y, rct_window* w, rct_drawpixelinfo* dpi, int32_t colour)
+{
+    if (percentage > 1)
+    {
+        percentage = 1;
+    }
+    int16_t barLength = static_cast<int16_t>((221 - 4) * percentage);
+
+    gfx_fill_rect_inset(dpi, x + 1, y + 1, x + 221 - 3, y + 9, w->colours[1], INSET_RECT_F_30);
```
tupaschoal, line 1458: Please use the `gfx_fill_rect_inset` that receives a `ScreenRect` instead.

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1446,6 +1446,22 @@ static void window_park_objective_invalidate(rct_window* w)
...
+    if (barLength > 2)
+    {
+        gfx_fill_rect_inset(dpi, x + 3, y + 2, x + barLength, y + 8, colour, 0);
```
tupaschoal, line 1461: Same as above.

src/openrct2-ui/windows/Park.cpp
tupaschoal, line 1458: What is the `221` magic number?

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1446,6 +1446,22 @@ static void window_park_objective_invalidate(rct_window* w)
     window_park_anchor_border_widgets(w);
 }
 
+static void window_park_objective_bars_paint(
+    float percentage, int32_t x, int32_t y, rct_window* w, rct_drawpixelinfo* dpi, int32_t colour)
```
tupaschoal, line 1450: Make it receive a `const ScreenCoordsXY&` object instead of the raw `x` and `y`

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1495,6 +1511,199 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                window_park_objective_bars_paint(guestsInParkPercentage, screenCoords.x, screenCoords.y, w, dpi, barColour);
+                screenCoords.y += LIST_ROW_HEIGHT;
+
+                ft = Formatter();
+                ft.Add<uint32_t>(static_cast<uint32_t>(gParkRating));
+
+                if (gScenarioObjective.Type == OBJECTIVE_GUESTS_BY)
+                {
+                    ft.Add<uint32_t>(static_cast<uint32_t>(600));
+                    parkRatingPercentage = (static_cast<float>(gParkRating) / static_cast<float>(600));
```
tupaschoal, line 1556: What's the `600` magic number? Maybe extract it to a `constexpr` for better documentation

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1495,6 +1511,199 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                }
+                else
+                {
+                    ft.Add<uint32_t>(static_cast<uint32_t>(700));
+                    parkRatingPercentage = (static_cast<float>(gParkRating) / static_cast<float>(700));
```
tupaschoal, line 1561: What's the `600` magic number? Maybe extract it to a `constexpr` for better documentation

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1495,6 +1511,199 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
         dpi, ft.Data(), screenCoords, 221, ObjectiveNames[gScenarioObjective.Type], COLOUR_BLACK);
     screenCoords.y += 5;
 
+    // Your progress
+    // if (gScenarioObjective.Check() == ObjectiveStatus::Undecided)
+    if (gScenarioCompletedCompanyValue == MONEY32_UNDEFINED)
+    {
+        gfx_draw_string_left(dpi, STR_OBJECTIVE_PROGRESS_LABEL, nullptr, COLOUR_BLACK, screenCoords);
```
tupaschoal, line 1518: I think the whole content of this `if` should be extracted to another function, something like `PaintProgressBar`. This is currently >300 lines added to a function that was already big.

JonasKlose:
@tupaschoal 
Thank you for all the feedback! 😃

src/openrct2-ui/windows/Park.cpp
JonasKlose: Thank you for the hint. I'll update the code accordingly.

JonasKlose: It just worked out the best for the size of the rectangle to fill the width of the park window.
I don't think the other code that used the function had a proper explanation for its size, too.

src/openrct2-ui/windows/Park.cpp
JonasKlose: The 600 and 700 magic number comes from Scenario.cpp. It's used in the functions CheckGuestsBy and CheckGuestsAndRating. I agree a constant would be better. I thought that the whole code of Scenario.cpp could be overhauled, but I guess a step between won't hurt anyone.

src/openrct2-ui/windows/Park.cpp
JonasKlose: You're right. I'll split the window_park_objective_paint function in four separate functions. It'll be much better to read.

tupaschoal:
You're welcome. Let us know when you've tended to the comments and also remember to mark the conversations as resolved when you fix them, thanks!

src/openrct2-ui/windows/Park.cpp
JonasKlose: > I think the whole content of this `if` should be extracted to another function, something like `PaintProgressBar`. This is currently >300 lines added to a function that was already big.

Phew... managed to split the function into four separate functions. I hope this worked out ok.

JonasKlose: Additionally, I've moved the check of the gScenarioCompletedCompanyValue into another function called GetCurrentObjectiveStatus(). This was a magical way of checking the objective status that also worked after the scenario was completet. Now the code is implemented in Scenario.cpp.

src/openrct2-ui/windows/Park.cpp
JonasKlose: Replaced 600 by the constexpr OBJECTIVE_GUESTS_BY_PARK_RATING.

src/openrct2-ui/windows/Park.cpp
JonasKlose: Replaced 700 by the constexpr OBJECTIVE_GUESTS_AND_RATING_PARK_RATING.

src/openrct2-ui/windows/Park.cpp
JonasKlose: I've switched to using ScreenRects now. :)

src/openrct2-ui/windows/Park.cpp
JonasKlose: I've switched to using ScreenRects now. :)

src/openrct2-ui/windows/Park.cpp
JonasKlose: I've looked at the function window_guest_stats_bars_paint in Guest.cpp. There are magic numbers in that code, too. It is used to paint the guest bars and that's the code that I used as example. I used my 221 like the 121 was used there.

Maybe, there's a solution for both functions. But I don't see it. Do you have an idea?

tupaschoal: I think it's alright to keep the magic numbers in this case.

src/openrct2-ui/windows/Park.cpp
tupaschoal: Nice!

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1446,31 +1446,43 @@ static void window_park_objective_invalidate(rct_window* w)
...
-static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
+static void window_park_objective_bars_paint(
+    float percentage, ScreenCoordsXY* screenCoords, rct_window* w, rct_drawpixelinfo* dpi, int32_t colour)
 {
-    WindowDrawWidgets(w, dpi);
-    window_park_draw_tab_images(dpi, w);
+    if (percentage > 1)
+    {
+        percentage = 1;
+    }
```
tupaschoal, line 1455: I'm wondering if this wouldn't be more elegant with a `clamp`, to guarantee that it is within boundaries:
```suggestion
    percentage = std::clamp(percentage, 0, 1);
```

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1446,31 +1446,43 @@ static void window_park_objective_invalidate(rct_window* w)
...
+    int16_t barLength = static_cast<int16_t>((221 - 4) * percentage);
+    ScreenRect* rect = new ScreenRect(screenCoords->x + 1, screenCoords->y + 1, screenCoords->x + 221 - 3, screenCoords->y + 9);
```
tupaschoal, line 1457: Don't use pointers, prefer stack allocations:
```suggestion
    auto rect = ScreenRect(screenCoords->x + 1, screenCoords->y + 1, screenCoords->x + 221 - 3, screenCoords->y + 9);
```

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1446,31 +1446,43 @@ static void window_park_objective_invalidate(rct_window* w)
...
-    screenCoords.y += gfx_draw_string_left_wrapped(dpi, ft.Data(), screenCoords, 222, STR_BLACK_STRING, COLOUR_BLACK);
-    screenCoords.y += 5;
+    gfx_fill_rect_inset(dpi, *rect, w->colours[1], INSET_RECT_F_30);
 
-    // Your objective:
-    gfx_draw_string_left(dpi, STR_OBJECTIVE_LABEL, nullptr, COLOUR_BLACK, screenCoords);
-    screenCoords.y += LIST_ROW_HEIGHT;
+    rect = new ScreenRect(screenCoords->x + 3, screenCoords->y + 2, barLength, screenCoords->y + 8);
+    if (barLength > 2)
+    {
```
tupaschoal, line 1463: Maybe add a comment as to why you're drawing the bar only when length is `> 2`

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1479,40 +1491,263 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                window_park_objective_bars_paint(guestsInParkPercentage, screenCoords, w, dpi, barColour);
+                screenCoords->y += LIST_ROW_HEIGHT;
+
+                *ft = Formatter();
+                ft->Add<uint32_t>(static_cast<uint32_t>(gParkRating));
+
+                if (gScenarioObjective.Type == OBJECTIVE_GUESTS_BY)
+                {
+                    ft->Add<uint32_t>(static_cast<uint32_t>(OBJECTIVE_GUESTS_BY_PARK_RATING));
+                    parkRatingPercentage = (static_cast<float>(gParkRating) / static_cast<float>(600));
```
tupaschoal, line 1554:
```suggestion
                    parkRatingPercentage = (static_cast<float>(gParkRating) / static_cast<float>(OBJECTIVE_GUESTS_BY_PARK_RATING));
```

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1479,40 +1491,263 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+                }
+                else
+                {
+                    ft->Add<uint32_t>(static_cast<uint32_t>(OBJECTIVE_GUESTS_AND_RATING_PARK_RATING));
+                    parkRatingPercentage = (static_cast<float>(gParkRating) / static_cast<float>(700));
```
tupaschoal, line 1559:
```suggestion
                    parkRatingPercentage = (static_cast<float>(gParkRating) / static_cast<float>(OBJECTIVE_GUESTS_AND_RATING_PARK_RATING));
```

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1479,40 +1491,263 @@ static void window_park_objective_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+            case OBJECTIVE_BUILD_THE_BEST:
+                screenCoords->y += gfx_draw_string_left_wrapped(
+                    dpi, nullptr, *screenCoords, 221, STR_OBJECTIVE_PROGRESS_NONE, COLOUR_BLACK);
+                break;
+            case OBJECTIVE_GUESTS_BY:
+            case OBJECTIVE_GUESTS_AND_RATING:
+            {
+                float guestsInParkPercentage
+                    = (static_cast<float>(gNumGuestsInPark) / static_cast<float>(gScenarioObjective.NumGuests));
+                float parkRatingPercentage;
```
tupaschoal, line 1535: Move this line to just before the `if (gScenarioObjective.Type == OBJECTIVE_GUESTS_BY)`

You can even initialize it to `static_cast<float>(gParkRating)` and then just divide it inside the `if` or `else`

tupaschoal (commented):
I've just done a partial review, but in general, try to avoid dynamic allocations like the `new ScreenRect` or `new ScreenCoordsXY`.

Also, it would be great to have most of the coordinate parameters as well as the formatter be const references, like it is in most of the code.

Gymnasiast:
Closing this due to inactivity. If you ever decide to pick it up again, you can open a new one. Do make sure you address the outstanding line notes if you do so.

PR #13687, opened 2021-01-03
Refactor code to dynamically add and remove viewports

ZehMatt:
Whenever a sprite is invalidated it iterated 64 times over the viewport array regardless if they are used or not, a viewport was considered valid when ```width > 0```. This PR uses std::list and now only iterates the actual amount of active viewports, I used std::list as std::vector can invalidate the pointers when it grows, std::list uses nodes. I also cleaned up some code around the variable and added appropriate functions instead. This PR removes the technical restriction of a maximum of 64 viewports but I left the check in as it would probably kill the framerate, this is now an arbitrary limit.

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -33,6 +33,7 @@
 
 #include <algorithm>
 #include <cstring>
+#include <list>
 
 using namespace OpenRCT2;
 
@@ -41,7 +42,7 @@ uint8_t gShowGridLinesRefCount;
 uint8_t gShowLandRightsRefCount;
 uint8_t gShowConstuctionRightsRefCount;
 
-rct_viewport g_viewport_list[MAX_VIEWPORT_COUNT];
+static std::list<rct_viewport> _viewports;
 rct_viewport* g_music_tracking_viewport;
 
 static std::unique_ptr<JobPool> _paintJobs;
@@ -75,12 +76,6 @@ void viewport_init_all()
     }
 
     window_init_all();
-
-    // Setting up viewports
-    for (int32_t i = 0; i < MAX_VIEWPORT_COUNT; i++)
-    {
-        g_viewport_list[i].width = 0;
-    }
 
     // ?
     input_reset_flags();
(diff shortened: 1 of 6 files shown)


src/openrct2/world/Sprite.cpp
```diff
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
```
Gymnasiast, line 1: BOM

Gymnasiast:
What is the advantage of doing it like this? More legible code? Better performance?

src/openrct2/interface/Viewport.cpp
```diff
@@ -202,6 +192,28 @@ void viewport_create(
     viewport->viewPos = *centreLoc;
 }
 
+void viewport_remove(rct_viewport* viewport)
```
duncanspumpkin, line 195: TitleCase all the new functions

src/openrct2/world/Sprite.cpp
ZehMatt: It appears to be on the left side, this reverts it apparently.

src/openrct2/interface/Viewport.cpp
ZehMatt: I want to refactor all this code into a namespace after this PR, its a big file.

ZehMatt:
> What is the advantage of doing it like this? More legible code? Better performance?

Slightly better performance and it killed some duplicate code.

src/openrct2/world/Sprite.cpp
Gymnasiast: Looks like you're right, it was introduced in efa1db826e2789e3d842995d5ced1ffb5e0a5a9c.

src/openrct2/interface/Viewport.cpp
Gymnasiast: Is there a reason why this cannot be put in the member function Viewport::Remove()?

ZehMatt: Rather a habbit but I guess its trivial to add, I'll see to do that.

IntelOrca: There is a reason you wouldn't do this... you typically tell the manager to remove a child, not the child to remove itself. So it would be more acceptable to do ViewportManager::Remove(viewport). Otherwise you typically have to set a flag on the viewport to self-delete at the end of the update or something since you wouldn't want to `delete this` within a member.

ZehMatt:
develop:
```
openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.070893s, 14 FPS
Zoom[1] average: 0.030573s, 33 FPS
Zoom[2] average: 0.014841s, 67 FPS
Total average: 0.038769s, 26 FPS
Time: 2.32614s
```

PR:
```
openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.065001s, 15 FPS
Zoom[1] average: 0.029304s, 34 FPS
Zoom[2] average: 0.014311s, 70 FPS
Total average: 0.036205s, 28 FPS
Time: 2.17233s
```
Its not huge but its something

Gymnasiast (approved):
I get slightly better performance as well.

duncanspumpkin (approved):

PR #13696, opened 2021-01-04
Rename dword_F64EB0 to marker

Gymnasiast:


duncanspumpkin:
What is marker?

Gymnasiast:
It should be evident if you read the code, that's why I picked this name.

But anyway: if the scenery should be drawn as a ghost, or as greyed out (when selecting scenery for a track design), this variable gets set.

duncanspumpkin:
Oh right it's the ghost or shadow modifier

PR #13697, opened 2021-01-04
Put the 0.3.2 changelog into debian/changelog

Gymnasiast:
The Ubuntu updater has gained the ability to show changelogs in recent years, so time to replace the dummy with an actual changelog.


tupaschoal (approved):
Just so we don't forget: I guess we'll need to update the release checklist to also mention touching this file

PR #13698, opened 2021-01-04
Improve performance of tweening entities between ticks

ZehMatt:
```
Renderer: Software, 720p, bpb.sv6
	develop: 125~ FPS
	PR: 160~ FPS
	
Renderer: Software (HW), 720p, bpb.sv6
	develop: 125~ FPS
	PR: 160~ FPS
	
Renderer: OpenGL, 720p, bpb.sv6
	develop: 385~ FPS
	PR: 465~ FPS
```
I think the numbers speak for themselves. Given that we only tween peeps and vehicles we can use the linked list in which each object is stored accordingly rather than messing around with every sprite slot there is.

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -46,9 +46,6 @@ const rct_string_id litterNames[12] = { STR_LITTER_VOMIT,
                                         STR_SHOP_ITEM_SINGULAR_EMPTY_JUICE_CUP,
                                         STR_SHOP_ITEM_SINGULAR_EMPTY_BOWL_BLUE };
 
-static CoordsXYZ _spritelocations1[MAX_SPRITES];
-static CoordsXYZ _spritelocations2[MAX_SPRITES];
-
 static size_t GetSpatialIndexOffset(int32_t x, int32_t y);
 static void move_sprite_to_list(SpriteBase* sprite, EntityListId newListIndex);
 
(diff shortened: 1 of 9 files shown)


src/openrct2/world/Sprite.h
```diff
@@ -378,4 +373,23 @@ template<typename T = SpriteBase> class EntityList
     }
 };
 
+class EntityTweener
+{
+    std::vector<SpriteBase*> _ents;
+    std::vector<CoordsXYZ> _prePos;
+    std::vector<CoordsXYZ> _postPos;
```
tupaschoal, line 380:
```suggestion
    std::vector<SpriteBase*> Entities;
    std::vector<CoordsXYZ> PrePos;
    std::vector<CoordsXYZ> PostPos;
```
According to our coding guidelines, `_bla` is for local static variables, where as member variables are title case.

src/openrct2/world/Sprite.cpp
```diff
@@ -928,106 +925,106 @@ uint16_t remove_floating_sprites()
...
-        case SpriteIdentifier::Null:
-            return false;
-    }
-    return false;
+    if (ent->sprite_identifier == SpriteIdentifier::Null)
+        return false;
+    if (ent->sprite_index == SPRITE_INDEX_NULL)
+        return false;
+    return true;
 }
```
tupaschoal, line 935: > Given that we only tween peeps and vehicles

Doesn't this change that claim?

src/openrct2/world/Sprite.cpp
```diff
@@ -928,106 +925,106 @@ uint16_t remove_floating_sprites()
...
+    for (auto ent : EntityList(id))
     {
-        // skip going through `get_sprite` to not get stalled on assert,
-        // this can get very expensive for busy parks with uncap FPS option on
-        const rct_sprite* sprite = &_spriteList[i];
-        sprite_locations[i].x = sprite->generic.x;
-        sprite_locations[i].y = sprite->generic.y;
-        sprite_locations[i].z = sprite->generic.z;
+        _ents.push_back(&(*ent));
+        _prePos.push_back({ ent->x, ent->y, ent->z });
```
tupaschoal, line 942:
```suggestion
        _prePos.emplace_back(ent->x, ent->y, ent->z);
```

src/openrct2/world/Sprite.h
ZehMatt: Thats news to me, the underscore is everywhere.

src/openrct2/world/Sprite.cpp
ZehMatt: I don't understand the question

src/openrct2/world/Sprite.h
tupaschoal: Well, I guess the underscores predates us having a style? Dunno, but that's what is documented here:
https://github.com/OpenRCT2/OpenRCT2/wiki/Coding-Style
```c++
/// Structure members are title case.
/**
 * Represents a currently running animation.
 */
struct PeepAnimation
{
    int32_t AnimationIndex;
    int32_t AnimationLength;
};
```

src/openrct2/world/Sprite.cpp
tupaschoal: This method is now returning true for `SpriteIdentifier::Misc` and `SpriteIdentifier::Litter`, unless those also have `ent->sprite_index == SPRITE_INDEX_NULL`, so isn't it now tweening those two?

src/openrct2/world/Sprite.h
ZehMatt: I mean it sort of makes sense for public members, none of those class members are public. Most of the code base has the underscore for private ones.

src/openrct2/world/Sprite.cpp
ZehMatt: Those sprites are never in the list.

ZehMatt: Hm actually now that I think about it, the sprite may be removed but re-used as something else, so yeah it probably makes sense to be specific here.

ZehMatt: Refactored that part, it takes the element out of the list and replaces it will null to on removal to ensure it doesn't consider re-allocated sprites still as valid.

src/openrct2/world/Sprite.h
IntelOrca: The style was underscore for private members or static global fields (although this should be kept to minimum)

duncanspumpkin:
All looks good to me all my points from discord have been taken into account.

src/openrct2/world/Sprite.h
ZehMatt: So should I leave it as is?

tupaschoal: Well, Ted did say *was*, so I guess it's what I said (?).
In any case I would really like the change to the ents variable name, at least

IntelOrca: Change to title case, public members are title case.

ZehMatt: They are not public tho

IntelOrca: Leave them with underscore prefixes then.

tupaschoal (approved):
LGTM if CI is clean

src/openrct2/world/Sprite.h
ZehMatt: Too late, we need to have an agreement on that at some point. Also should update the Wiki page with more samples and cases like this.

PR #13699, opened 2021-01-04
Collapse function template in PaintSessionArrange

ZehMatt:
Removes a bunch of branches and moves the selection to the top of the hierarchy, based on the rotation it now picks inlined code specific to the rotation rather than picking the rotation over and over.

### DIFF
--- a/src/openrct2/paint/Paint.cpp
+++ b/src/openrct2/paint/Paint.cpp
@@ -11,6 +11,7 @@
 
 #include "../Context.h"
 #include "../config/Config.h"
+#include "../core/Guard.hpp"
 #include "../drawing/Drawing.h"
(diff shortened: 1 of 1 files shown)


tupaschoal (commented):
Do you have some benchmark numbers? IIRC @duncanspumpkin recently touched this and I remember him doing something specifically to restore performance

ZehMatt:
He modified PaintSessionGenerate to also collapse direction/rotation into the function. The improvement on my machine is only 2 FPS but this may be better on older CPUs.

IntelOrca (approved):

duncanspumpkin:
Yes this makes sense. I just did this function in loco. May as well port over this change as well.

PR #13704, opened 2021-01-05
Move window update out of variable frame updates

ZehMatt:
1. The drawing engine shouldn't do windows update calls.
2. The vanilla logic is to always update the windows at 40hz, this did it every frame but is locked to the real time ticks variable so the updates are not doing anything between updates.

### DIFF
--- a/src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
+++ b/src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
@@ -338,11 +338,6 @@ public:
         window_draw_all(&_bitsDPI, 0, 0, _width, _height);
     }
 
-    void UpdateWindows() override
-    {
-        window_update_all();
-    }
-
     void PaintWeather() override
     {
         _drawingContext->SetDPI(&_bitsDPI);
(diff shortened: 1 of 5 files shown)


ZehMatt:
Would be nice if this could get some testing, but in theory this should be identical to running the game always at 40hz.

Gymnasiast:
It appears to all work correctly, but I don't really understand what you changed and why. Could you elaborate on it?

ZehMatt:
On develop if you have Uncap FPS enabled it would call the windows update function every frame so when you have 500 FPS it would do 500 calls to it which is a waste of cpu cycles. What it does now is only call it every 40Hz as if you were to not have Uncap FPS enabled, this is how vanilla works also. Regarding responsiveness there is no difference as the input is collected every 40Hz as well so dragging and things like that will not change if we have it updated at variable frame rate. Also it makes no sense to have the drawing engine do the update call, a drawing engines job is to draw things and not manage them.

Gymnasiast:
Could this potentially lead to better performance?

Gymnasiast (approved):

ZehMatt:
Potentially, but this heavily depends on whats going on windows wise.

IntelOrca:
Question is, do we want the interface to update equally to draw fps? This would give you smoother window dragging and animation.

ZehMatt:
It needs a lot of changes to do that, the mouse input needs to be gathered every frame as well then.

tupaschoal (approved):
Looks fine to me (and thanks for the detailed explanation!).

Seems mergeable if @IntelOrca has nothing against it.

IntelOrca (approved):

Issue #13693, opened 2021-01-04
Launched Free Fall with 93km/h goes higher than 96km/h


telk5093:
**OS:** Windows 10
**Version:** 0.3.2
**Commit/Build:** 920dd75

Launched free fall with 93km/h of launch speed goes higher and arrives faster than one with 96km/h.

**Steps to reproduce:**
1. Build two launched free falls
2. Set one's launch speed as 93km/h
3. Set the other's launch speed as 96km/h
4. Pause the game, test them and then unpause game to run them with synchronized

**Screenshots / Video:**
**In OpenRCT2:** (Left: 96km/h / Right: 93km/h)
![1](https://user-images.githubusercontent.com/10726524/103506226-8d296900-4e9f-11eb-8f61-372df1ebd29a.gif)

**In Vanilla:** (Left: 93km/h / Right: 96km/h)
![2](https://user-images.githubusercontent.com/10726524/103506230-8ef32c80-4e9f-11eb-8241-9b3fc550dce7.gif)

rottier:
Found the issue, I suspect something went wrong in a merge/regex. I'll open a PR to fix this one.

PR #13709, opened 2021-01-06
Fix OpenRCT2#13693: superfluous case in Upward Launch

rottier:
The updward launch mode contained a case for the Air Powered Vertical Coaster that doesn't make any sense because this coaster doesn't have this mode by default. The bitshift value used there was correct, however. This was possibly the result of a faulty regex or merge.

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -3270,17 +3270,8 @@ void Vehicle::UpdateDeparting()
         case RideMode::PoweredLaunchBlockSectioned:
         case RideMode::LimPoweredLaunch:
         case RideMode::UpwardLaunch:
-            if (curRide->type == RIDE_TYPE_AIR_POWERED_VERTICAL_COASTER)
-            {
-                if ((curRide->launch_speed << 16) > velocity)
-                {
-                    acceleration = curRide->launch_speed << 13;
-                }
-                break;
-            }
-
             if ((curRide->launch_speed << 16) > velocity)
-                acceleration = curRide->launch_speed << 12;
+                acceleration = curRide->launch_speed << 13;
             break;
         case RideMode::DownwardLaunch:
             if (var_CE >= 1)


duncanspumpkin:
Is this an original bug in the vanilla game?

rottier:
> Is this an original bug in the vanilla game?

Nope, it's fine in vanilla (see #13693 )

duncanspumpkin:
Then this is not the solution as this matches the assembly.

duncanspumpkin:
In the assembly it branches at 0x6d862e on ride type 0x4b. RIDE_TYPE_AIR_POWERED_VERTICAL_COASTER

The two branches are as per the current code one for left shift 16-4 the other for 16-3.

duncanspumpkin:
Also why would you change the behaviour of the air powered vertical coaster? This (RideMode::PoweredLaunchPasstrough)  is the default mode for that coaster so it is definitely in use.

rottier:
> Also why would you change the behaviour of the air powered vertical coaster? This (RideMode::PoweredLaunchPasstrough) is the default mode for that coaster so it is definitely in use.

I'm probably not understanding it correctly, but what I am seeing is that there is a rule for the Air Powered Vertical Coaster in the Upward Launch (so not the Powered Launch Passthrough) mode case, which doesn't make sense to me because that coaster doesn't have the Upward Launch mode by default. Could it be that the Powered Launch Passthrough code is in the wrong case, or should I look for solutions outside of the code that is based on the assembly (possibly because it's generated)?

rottier:
> Also why would you change the behaviour of the air powered vertical coaster? This (RideMode::PoweredLaunchPasstrough) is the default mode for that coaster so it is definitely in use.
> 
> Perhaps you are not familiar with c++. In c++ you can have multiple cases in a switch statement all go to the same bit of code. So as there are multiple modes stacked ontop of this case it is valid for all of the cases not just upward launch.

That explains a lot, thanks for the explanation! I'll close this one, it doesn't make sense and the problem lies elsewhere.

duncanspumpkin:
It will probably be something similar to this. It was a valid idea. Just not correct this time.

PR #13710, opened 2021-01-06
Refactor/string view

skdltmxn:
Refactored string_view usages through codebase

- Modified `const std::string_view&` to `std::string_view`
- Modified a number of potentially problematic usages of string_view such as printing out to exception messages

### DIFF
--- a/src/openrct2/core/File.cpp
+++ b/src/openrct2/core/File.cpp
@@ -43,21 +43,20 @@ namespace File
         return platform_file_move(srcPath.c_str(), dstPath.c_str());
     }
 
-    std::vector<uint8_t> ReadAllBytes(const std::string_view& path)
+    std::vector<uint8_t> ReadAllBytes(std::string_view path)
     {
-        std::vector<uint8_t> result;
-
 #if defined(_WIN32) && !defined(__MINGW32__)
-        auto pathW = String::ToWideChar(std::string(path));
+        auto pathW = String::ToWideChar(path);
         std::ifstream fs(pathW, std::ios::in | std::ios::binary);
 #else
         std::ifstream fs(std::string(path), std::ios::in | std::ios::binary);
 #endif
         if (!fs.is_open())
         {
-            throw IOException("Unable to open " + std::string(path.data()));
+            throw IOException("Unable to open " + std::string(path));
         }
 
+        std::vector<uint8_t> result;
         fs.seekg(0, std::ios::end);
         auto fsize = static_cast<size_t>(fs.tellg());
         if (fsize > SIZE_MAX)
@@ -76,7 +75,7 @@ namespace File
         return result;
     }
 
-    std::string ReadAllText(const std::string_view& path)
+    std::string ReadAllText(std::string_view path)
     {
         auto bytes = ReadAllBytes(path);
         // TODO skip BOM
@@ -85,13 +84,7 @@ namespace File
         return result;
     }
 
-    void WriteAllBytes(const std::string& path, const void* buffer, size_t length)
-    {
-        auto fs = OpenRCT2::FileStream(path, OpenRCT2::FILE_MODE_WRITE);
-        fs.Write(buffer, length);
-    }
-
-    std::vector<std::string> ReadAllLines(const std::string& path)
+    std::vector<std::string> ReadAllLines(std::string_view path)
     {
         std::vector<std::string> lines;
         auto data = ReadAllBytes(path);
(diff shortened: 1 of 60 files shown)


src/openrct2/core/Zip.cpp
```diff
@@ -31,7 +31,7 @@ class ZipArchive final : public IZipArchive
         }
 
         int32_t error;
-        _zip = zip_open(path.data(), zipOpenMode, &error);
+        _zip = zip_open(std::string(path).data(), zipOpenMode, &error);
```
ZehMatt, line 34: You should use c_str() to if you want null terminated string, I know you can argue that most implementations do same as data() but for correctness it should be c_str()

This should be changed in all cases.

src/openrct2/localisation/Convert.cpp
```diff
@@ -85,7 +85,7 @@ static std::string DecodeToMultiByte(const std::string_view& src)
 /**
  * Encodes a UTF-8 string as an RCT2 string.
  */
-static std::string Encode(const std::string_view& src)
+static std::string Encode(const std::string& src)
```
ZehMatt, line 88: Whats the reason behind the change here?

src/openrct2/scripting/Plugin.cpp
```diff
@@ -155,10 +155,10 @@ PluginType Plugin::ParsePluginType(const std::string_view& type)
     throw std::invalid_argument("Unknown plugin type.");
 }
 
-void Plugin::CheckForLicence(const DukValue& dukLicence, const std::string_view& pluginName)
+void Plugin::CheckForLicence(const DukValue& dukLicence, std::string_view pluginName)
 {
     if (dukLicence.type() != DukValue::Type::STRING || dukLicence.as_string().empty())
-        log_error("Plugin %s does not specify a licence", pluginName.data());
+        log_error("Plugin %s does not specify a licence", std::string(pluginName).data());
```
ZehMatt, line 161:
```suggestion
        log_error("Plugin %s does not specify a licence", std::string(pluginName).c_str());
```

test/tests/ImageImporterTests.cpp
```diff
@@ -19,7 +19,7 @@ using namespace OpenRCT2::Drawing;
 class ImageImporterTests : public testing::Test
 {
 public:
-    static std::string GetImagePath(const std::string_view& name)
+    static std::string GetImagePath(const std::string& name)
     {
         return Path::Combine(TestData::GetBasePath(), "images", name.data());
```
ZehMatt, line 24:
```suggestion
        return Path::Combine(TestData::GetBasePath(), "images", name.c_str());
```

src/openrct2/object/ObjectFactory.cpp
```diff
@@ -52,14 +52,14 @@ class FileSystemDataRetriever : public IFileDataRetriever
...
+    FileSystemDataRetriever(std::string_view basePath)
         : _basePath(basePath)
     {
     }
 
-    std::vector<uint8_t> GetData(const std::string_view& path) const override
+    std::vector<uint8_t> GetData(std::string_view path) const override
     {
-        auto absolutePath = Path::Combine(_basePath, path.data());
+        auto absolutePath = Path::Combine(_basePath, std::string(path).data());
```
ZehMatt, line 62:
```suggestion
        auto absolutePath = Path::Combine(_basePath, std::string(path).c_str());
```

ZehMatt (changes requested):
Overall pretty good changes, just a few minor things here and there.

src/openrct2/core/Zip.cpp
skdltmxn: Actually as of C++11, c_str() and data() are exactly the same by standard ;)

src/openrct2/localisation/Convert.cpp
skdltmxn: Because `src` is treated as null-terminating string inside of `Encode` function.

src/openrct2/core/Zip.cpp
ZehMatt: I still don't like the mixed usage, c_str() is preferred to clarify intent, when data() is used its usually assumed that its buffer is used as raw data.

src/openrct2/localisation/Convert.cpp
ZehMatt: This encode function is not pretty but yeah I see the point, ideally it should not seek for the null terminator and rather use the range specified by string_view/string but thats for another PR.

src/openrct2/core/Zip.cpp
skdltmxn: Fair point there 👍 
I will fix them

src/openrct2/core/File.cpp
```diff
@@ -120,11 +113,17 @@ namespace File
         return lines;
     }
 
+    void WriteAllBytes(const std::string& path, const void* buffer, size_t length)
```
Gymnasiast, line 116: Why did you move this function? Makes the diff noisier than necessary.

ZehMatt (approved):
Looks good, I did no testing on this but I see no issues here.

src/openrct2/core/Zip.cpp
IntelOrca: When dealing with `string_view`, it deliberately does not have c_str as it might not be null terminated, you never know with `data()`.

src/openrct2/core/File.cpp
skdltmxn: Because I find it more readable to have all Read* APIs in a row and then Write API.

tupaschoal (approved):
Seems fine to me, looks pretty safe

PR #13713, opened 2021-01-06
Cleanup variables usage and initialize them

ZehMatt:
I've been looking into some issues and one of them had this strange value used as a pointer so I went digging and found that when the operating mode is changed the code in UpdateMovingToEndOfStation could potentially use an uninitialized variable for the station index. While I was doing that I looked at what other variables are potentially uninitialized and fixed up a bunch of them, we all know thats never good. Also did some improvements here and there code readability wise.


Gymnasiast (approved):
LGTM, but would like Ted or Duncan to take a look as well.

duncanspumpkin (approved):

PR #13720, opened 2021-01-07
[Plugin] Re-add deprecated 'edgesAndCorners' support for footpaths

Basssiiie:
Pull request #13536 separated the `edgesAndCorners` property for footpaths into `edges` and `corners`. This is an improvement for the Plugin API, but it broke a few existing plugins in the process.

Some of the plugins affected are my [ProxyPather](http://openrct2plugins.org/plugin/MDEwOlJlcG9zaXRvcnkzMjU4NjIwMDA=/OpenRCT2-ProxyPather) plugin and Oli414's [PathExpander](http://openrct2plugins.org/plugin/MDEwOlJlcG9zaXRvcnkyNjE3MTA1Njg=/PathExpander).

This PR re-adds the property `edgesAndCorners` as a deprecated property. The original C++ methods are re-added, but the `openrct2.d.ts` definition is left out. This means that existing plugins will keep working while new and updated plugins can use `edges` and `corners` separately without knowing that `edgesAndCorners` exists.

### DIFF
--- a/src/openrct2/scripting/ScTile.hpp
+++ b/src/openrct2/scripting/ScTile.hpp
@@ -1121,6 +1121,26 @@ namespace OpenRCT2::Scripting
             }
         }
 
+        // Deprecated in favor of seperate 'edges' and 'corners' properties,
+        // left here to maintain compatibility with older plugins.
+        /** @deprecated */
+        uint8_t edgesAndCorners_get() const
+        {
+            auto el = _element->AsPath();
+            return el != nullptr ? el->GetEdgesAndCorners() : 0;
+        }
+        /** @deprecated */
+        void edgesAndCorners_set(uint8_t value)
+        {
+            ThrowIfGameStateNotMutable();
+            auto el = _element->AsPath();
+            if (el != nullptr)
+            {
+                el->SetEdgesAndCorners(value);
+                Invalidate();
+            }
+        }
+
         DukValue edges_get() const
         {
             auto ctx = GetContext()->GetScriptEngine().GetContext();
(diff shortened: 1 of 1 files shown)


IntelOrca (approved):

duncanspumpkin:
Is this the best way to mark them as depreciated in the ts file?

Basssiiie:
@duncanspumpkin I'm not sure what you mean. If you are talking about the `/** @deprecated */`, I could not find a consistent way in the OpenRCT2 source code, so I picked the variant that matched closest what IntelOrca mentioned on the OpenRCT2 Discord in the Plugin channel.

duncanspumpkin:
I don't know much about typescript but I was meaning is there a way to make it more obvious in the ts file that that is depreciated rather than just hidden.

Basssiiie:
Ah so. In Typescript/other linting programs it would just show up as a property that does not exist on `FootpathElement` at compile/linting time. Javascript would just run fine since it ignores any typing.

Another option is to add it back to `openrct2.d.ts` but use [JSDoc's @deprecated](https://stackoverflow.com/questions/60755711/is-it-possible-to-mark-something-as-deprecated-in-typescript). Actually looking back at Intel's comment, he might have meant that instead of the C++ comment, but I'm not sure.

If the JSDoc comment is preferred, please let me know and I'll add them in. :)

duncanspumpkin:
Okay. Well Ted's approved it so think this will do.

IntelOrca:
Oh, yes I did mean `@deprecated` should be in the openrct2.d.ts file. Never mind, we can remove it again after all the existing plugins have changed over to using the new API.

Basssiiie:
Oops, my mistake for misunderstanding! I'll update my plugin later this week but I can't speak for the other plugin(s). And there's also the question how quickly all users will update their installed plugins.

PR #13728, opened 2021-01-08
Minor code cleanups

gibmat:
Some minor cleanups identified while working on Debian packaging:

* Add some keywords to .desktop file
* Fix typos
* Fix groff warnings in manpage
* If cmake variables contain just empty strings, don't add them to the compile definitions, as this can result in incorrect version strings in the compiled program


duncanspumpkin:
Try to keep separate PR's for these things realise they are all small but different team members review different parts of this.

gibmat:
OK, I'll submit four separate pull requests

Issue #2876, opened 2016-02-03
Increase camera boundaries


Phyxion:
Video: https://vid.me/vBqK

Visible in 0-1.5s, you cannot see the entire roller coaster unless you zoom out. If the camera could go further away from the map you would be able to see it in the normal zoom level as well.

[Extreme Heights.sv6.txt](https://github.com/OpenRCT2/OpenRCT2/files/116167/Extreme.Heights.sv6.txt)

Gymnasiast:
(Never mind the label dance. To anyone: make sure you look at the video, the description is/was a bit ambiguous.)

Nubbie:
Same video as in issue #2875 , this is explains the same problem, maybe combine the two issues into one? 
This still happens on 0.0.5.0 build a937a65

PR #13734, opened 2021-01-09
Implement element offset scroll limits

RobertWHurst:
Adds some tracking of the highest tiles so that maps with really tall tiles are not cut from view by the map edge limits. Here's a [Screencap after changes](https://streamable.com/eepxco).

See #2876 for more details about the original issue.


src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1540,54 @@ void map_update_tiles()
     }
 }
 
+void map_calc_highest_tile_height()
+{
+    gMapHighestTileHeight = 0;
+    for (int32_t i = 0; i < MAX_TILE_TILE_ELEMENT_POINTERS; i++)
+    {
+        TileElement* tile = &gTileElements[i];
```
IntelOrca, line 1548: It is better to have a double for loop for y axis and x axis and then use `map_get_surface_element`. The surface element is not always going to be the first element on a tile and some tiles do not have any elements at all so you need to check for `nullptr`.

IntelOrca (changes requested):
Looks good.

Is your initial implementation only to take land height into account? What about water height, and any other elements like track or scenery?

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1540,54 @@ void map_update_tiles()
...
+MapRange map_get_edge_limits()
+{
+    uint8_t rotation = get_current_rotation();
+
+    int32_t x1 = MAP_MINIMUM_X_Y;
+    int32_t y1 = MAP_MINIMUM_X_Y;
+    int32_t x2 = gMapSizeMinus2;
+    int32_t y2 = gMapSizeMinus2;
+
+    switch (rotation)
```
duncanspumpkin, line 1566: We already have lots of rotation functions. Don't add another use the already existing one

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1540,54 @@ void map_update_tiles()
     }
 }
 
+void map_calc_highest_tile_height()
+{
+    gMapHighestTileHeight = 0;
+    for (int32_t i = 0; i < MAX_TILE_TILE_ELEMENT_POINTERS; i++)
+    {
+        TileElement* tile = &gTileElements[i];
+        int32_t base_z = tile->GetBaseZ();
```
duncanspumpkin, line 1549: The highest height should be the clearance height not the base height

src/openrct2/world/Map.cpp
RobertWHurst: Is there a function you have in mind that I can use for this? I don't see one that does what I'm doing here with this range.

RobertWHurst:
@IntelOrca Thanks! Indeed it is. I thought I'd start with that and account for other elements provided what I had implemented was sound to build upon. I'll add the water height next.

ZehMatt:
I feel like calculating the height of the entire map whenever land is raised/lowered seems a bit expensive. This is probably fine for smaller maps but the bigger it gets the slower this function will become.

IntelOrca:
It doesn't need to iterate throughout the whole map after a land edit action, it can just iterate the affected area.

robert-hurst-cmd:
@ZehMatt @IntelOrca But what if the effected area is lowered below another tile in a completely different part of the park? There would be no way to know when to stop decrementing the calculated height.

ZehMatt:
> It doesn't need to iterate throughout the whole map after a land edit action, it can just iterate the affected area.

I'm not sure how that would work, if the highest point is cached how do you lower it without iterating all tiles? 
Example:
I raise land to 128 at the bottom left corner, sets highest to 128
I raise land to 64 at the top right corner, highest stays at 128
I lower land to 32 at the bottom left corner, highest stays at 128 but 64 is now the highest on map.

Because I lowered the first corner back to 32 that would be the only area affected.

I think the most efficient way to do this is to divide the map into multiple areas, so when a single tile changes it just has to look at the area to determine the max height of that area, when thats done it has to only look at the known heights from the other areas. Having the areas sorted allows binary search speed and getting the highest point is always the last area in the list.

robert-hurst-cmd:
@ZehMatt what about computing the highest by checking a limited number of tiles in the background each tick. This was a suggestion of @IntelOrca and I probably should have did it that way initially, lol.

ZehMatt:
> @ZehMatt what about computing the highest by checking a limited number of tiles in the background each tick. This was a suggestion of @IntelOrca and I probably should have did it that way initially, lol.

I prefer to not have background tasks/threads when possible, this just adds a bunch of complexity and potential race conditions.

Heres a quick mockup of what I had in mind: https://gist.github.com/ZehMatt/c278e52fb060b678e1b8cc3d30b13033

RobertWHurst:
@ZehMatt Interesting, I can give that a shot 👍 . The advantage here too is that the viewport won't seemly randomly snap back after dropping tile heights.

IntelOrca:
> I prefer to not have background tasks/threads when possible, this just adds a bunch of complexity and potential race conditions.

No, not that kind of background. The same way it does grass growth and ride stat caclulations.

RobertWHurst:
Indeed. Sorry, I should have clarified that no threading was involved in the "background" solution.

duncanspumpkin:
As this is only checking surface tiles how will it work when other tile elements are much higher than the surface

RobertWHurst:
@duncanspumpkin Rides and other elements that are higher than 178m in-game get cut off still, but at least now that's true if the ride/scenery is built at the highest or lowest point on the map. The opportunity to build things on varied terrain is the same at all tile heights. Having said that I am going to try and account for other element types as well.

Here's a screenshot with an observation tower, and my viewport pushed right to the top scroll limits at the standard zoom level using my branch. Using a develop build of the game at the standard zoom level you can't even see the ground tiles the observation tower is built on let alone the ride itself.
image.

src/openrct2/Game.cpp
```diff
@@ -573,6 +573,7 @@ void game_load_init()
     reset_sprite_spatial_index();
     reset_all_sprite_quadrant_placements();
     scenery_set_default_placement_configuration();
+    map_calc_highest_tile_height(true);
```
RobertWHurst, line 576: On game load, we check the whole map to make sure the viewport doesn't have forced down for tall parks where the game was saved panned all the way to the top.

src/openrct2/GameState.cpp
```diff
@@ -281,6 +281,7 @@ void GameState::UpdateLogic()
     scenario_update();
     climate_update();
     map_update_tiles();
+    map_calc_highest_tile_height(false);
```
RobertWHurst, line 284: Each tick of the game loop we check a small set of map tiles for the tallest element.

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1542,92 @@ void map_update_tiles()
...
+        auto tileCoords = CoordsXY(x, y);
+        auto tileElement = map_get_first_element_at(tileCoords);
+
+        // Loop through elements on the current tile and find the tallest
+        do
+        {
+            if (tileElement == nullptr)
+                break;
+
+            if (tileElement->GetType() == TILE_ELEMENT_TYPE_CORRUPT)
```
RobertWHurst, line 1585: I'm only ignoring corrupted map elements. I had a look at the defined constants for element types, and it seemed like a good idea to include everything else in the check. Is there anything else that should be excluded?

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1542,92 @@ void map_update_tiles()
...
+    // position to start at tile zero, checking all tiles before returning.
+    // Otherwise check a limited number of tiles per each time the
+    // function is called.
+    if (fullCheck)
+    {
+        gMapHighestTileHeightLoopPosition = 0;
+    }
+    else
+    {
+        tilesToCheck /= 128;
```
RobertWHurst, line 1560: Dividing the check across 128 cycles of the game loop when not performing a full check. This seemed performant but might need some thought on what the right number here is.

RobertWHurst:
Right, so I've pushed another commit. This time I'm checking all element types with the exception of corrupt elements, and I'm doing it incrementally on the game loop as suggested by @IntelOrca. Now you can look right at the needle on the tallest observation tower built on the tallest edge of a map 😁

src/openrct2/world/Map.cpp
ZehMatt: The current time it needs to complete a full cycle: (256 * 256) / 128 * 40 = 20480 ms. Also I'm not sure if it should always  run, when nothing changes then it doesn't have to do anything.

duncanspumpkin: You could have a flag that gets invalidated from the game actions to retrigger the checks.

src/openrct2/world/Map.cpp
duncanspumpkin: This all makes sense.

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1542,92 @@ void map_update_tiles()
     }
 }
 
+void map_calc_highest_tile_height(bool fullCheck)
```
duncanspumpkin, line 1545: All new functions are to be `TitleCase`

RobertWHurst: Sounds good, I'll make the change 👍🏻

src/openrct2/world/Map.cpp
RobertWHurst: That makes sense to me, I'll add the flag, and a function to invalidate. I'll then place the invalidate call in actions that modify elements.

src/openrct2/world/Map.cpp
ocalhoun6: > Is there anything else that should be excluded?

Do free-floating balloons count as map elements? Because they might get pretty high when floating away, and the scroll limits probably shouldn't change just because of balloons.

RobertWHurst: Lol, I'm glad you brought that up. I'll check for balloons and ducks and make sure they're not included.

src/openrct2/world/Map.cpp
```diff
@@ -941,27 +945,8 @@ bool map_is_location_at_edge(const CoordsXY& loc)
...
-    // Mark the latest element with the last element flag.
-    (tileElement - 1)->SetLastForTile(true);
-    tileElement->base_height = MAX_ELEMENT_HEIGHT;
-
-    if ((tileElement + 1) == gNextFreeTileElement)
-    {
-        gNextFreeTileElement--;
-    }
+void tile_element_remove(TileElement* tileElement) {
+    tileElement->Remove();
```
RobertWHurst, line 949: I moved this logic into `TileElement::Remove()` as it seemed to make more sense to live in there rather than this decompiled function. Does this make sense?

src/openrct2/world/Map.cpp
RobertWHurst: A balloon isn't a tile element as far as I can tell. It looks like it's a different type of entity, so it shouldn't affect this logic at all.

src/openrct2/world/TileElement.cpp
```diff
@@ -168,7 +168,29 @@ void TileElement::ClearAs(uint8_t newType)
 
 void TileElementBase::Remove()
 {
-    tile_element_remove(static_cast<TileElement*>(this));
+    auto tileElement = this;
```
RobertWHurst, line 171: This is the logic that use to live in `tile_element_remove`. It seems to make more sense to have it here. Let me know if I'm off base here.

src/openrct2/world/TileElement.cpp
```diff
@@ -168,7 +168,29 @@ void TileElement::ClearAs(uint8_t newType)
...
+    // Mark the latest element with the last element flag.
+    (tileElement - 1)->SetLastForTile(true);
+    tileElement->base_height = MAX_ELEMENT_HEIGHT;
+
+    if ((tileElement + 1) == gNextFreeTileElement)
+    {
+        gNextFreeTileElement--;
+    }
+
+    mapScheduleCalcHighestTileHeight();
```
RobertWHurst, line 193: When an element is removed, we schedule a highest title calc

src/openrct2/world/TileElement.cpp
```diff
@@ -242,5 +264,10 @@ int32_t TileElementBase::GetClearanceZ() const
 
 void TileElementBase::SetClearanceZ(int32_t newZ)
 {
-    clearance_height = (newZ / COORDS_Z_STEP);
+    uint8_t new_clearance_height = (newZ / COORDS_Z_STEP);
+    if (clearance_height != new_clearance_height)
+    {
+        mapScheduleCalcHighestTileHeight();
```
RobertWHurst, line 270: If the clearance height of an element is changed then a highest tile calc is scheduled.

RobertWHurst:
Alright, I've now added scheduling so the calculation is only started if a tile element's clearance height is changed, or if the element is removed.

robert-hurst-cmd:
Ok there we go. I reverted the remove function change as requested by @duncanspumpkin. This should be ready for review again, let me know if it needs further tweaks, I'm happy to polish further if need be :)

tupaschoal (commented):
@RobertWHurst please make sure you mark the conversations that are dealt with as resolved, so that reviewers can know what is still pending comments from them :)

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1545,103 @@ void map_update_tiles()
...
+    {
+        tilesToCheck /= 64;
+    }
+
+    for (int32_t i = 0; i < tilesToCheck; i++)
+    {
+        gMapHighestTileHeightLoopPosition++;
+        gMapHighestTileHeightLoopPosition %= maxTileIndex;
+
+        if (gMapHighestTileHeightLoopPosition == 0)
```
ZehMatt, line 1581: In order for it to finish it would have to wrap at least once. It would be better to check if ```gMapHighestTileHeightLoopPosition == maxTileIndex```, this loop should only go from 0 to maxTileIndex.
This also makes the use of modulo redundant.

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1545,103 @@ void map_update_tiles()
...
 }
 
+void mapScheduleCalcHighestTileHeight()
+{
+    gCalcHighestTileHeightScheduled = true;
+}
+
+void mapTryCalcHighestTileHeight(bool fullCheck)
+{
+    if (!fullCheck && !gCalcHighestTileHeightScheduled)
```
ZehMatt, line 1555: You dont need check ```gCalcHighestTileHeightScheduled``` if you can check for ```gMapHighestTileHeightLoopPosition < maxTileIndex```, simply setting ```gMapHighestTileHeightLoopPosition``` to 0 will start re-start the loop.

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1545,103 @@ void map_update_tiles()
...
+    {
+        return;
+    }
+
+    int32_t maxTileIndex = MAXIMUM_MAP_SIZE_TECHNICAL * MAXIMUM_MAP_SIZE_TECHNICAL;
```
ZehMatt, line 1560:
```suggestion
    constexpr int32_t maxTileIndex = MAXIMUM_MAP_SIZE_TECHNICAL * MAXIMUM_MAP_SIZE_TECHNICAL;
```

ZehMatt (changes requested):

src/openrct2/world/Map.cpp
duncanspumpkin: Balloons are entities. Not relevant.

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1545,103 @@ void map_update_tiles()
     }
 }
 
+void mapScheduleCalcHighestTileHeight()
```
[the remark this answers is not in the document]
RobertWHurst: Indeed, that is a better name 👍

RobertWHurst:
I've added an update to address some suggestions, but also to fix some bugs I found.  I ended up placing calls to MapInvalidateHeightCache most places an element's clearance height or water height is modified. I excluded a few irrelevant spots and ignored ghosts as they probably shouldn't trigger this code.

src/openrct2/actions/WallPlaceAction.cpp
```diff
@@ -431,6 +431,10 @@ GameActions::Result::Ptr WallPlaceAction::Execute() const
     {
         wallElement->SetGhost(true);
     }
+    else
+    {
+        MapInvalidateHeightCache();
```
duncanspumpkin, line 436: why don't you just invalidate when calling `tile_element_insert`. It doesn't really matter that much about ghosts this isn't the be all end all if it gets invalidated slightly more than required but it would be unfortunate to miss an invalidation.

robert-hurst-cmd: I did it this way because the ghosts blink in and out of existence and cause the height check to restart over and over. It ends up being pretty wasteful and on large parks prevents a height calculation from completing while they flash about.

RobertWHurst: I found I needed to exclude ghosts because they blink in and out of existance quick enough that they prevent the height check from completing on larger parks and it just seemed wasteful. Idealy everything would have been hooked inside of the TileElement, but that didn't work either as I found a lot of code simply edits the clearence height directly via the property rather than the setter method.

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1548,106 @@ void map_update_tiles()
...
+    // If we're doing a full check of the map then zero out the loop
+    // position to start at tile zero, checking all tiles before returning.
+    // Otherwise check a limited number of tiles per each time the
+    // function is called.
+    if (fullCheck)
+        _mapHighestTileHeightLoopPosition = 0;
+    else
+        tilesToCheck /= 64;
+
+    for (int32_t i = 0; i <= tilesToCheck; i++)
```
ZehMatt, line 1574: I'm not sure the loop condition is correct. Total iterations would be (64 * 1024) + 64.
See https://godbolt.org/z/jsPvdo

ZehMatt (changes requested):

src/openrct2/world/Map.cpp
RobertWHurst: Your right. That's a bit embarrassing. I confused myself by naming the tile count maxTileIndex, which is wrong. Will update, sorry about that.

ZehMatt: No worries, happens to the best. Thats what reviews are good for 😉

RobertWHurst:
Alright, fixed the loop. Anything else I need to address?

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1548,104 @@ void map_update_tiles()
...
+        } while (!(tileElement++)->IsLastForTile());
+
+        if (_mapHighestTileHeightLoopPosition == mapTileCount - 1)
+        {
+            _mapHighestTileHeight = _mapHighestTileHeightPending;
+            _mapHighestTileHeightPending = 0;
+            break;
+        }
+        else
+            _mapHighestTileHeightLoopPosition++;
```
ZehMatt, line 1611: If you move this above the ```if``` then you can remove ```- 1``` in the test and at the top of the function.

RobertWHurst: Sounds good, done 👍

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1548,104 @@ void map_update_tiles()
...
+}
+
+void MapUpdateHeightCache(bool fullCheck)
+{
+    constexpr int32_t mapTileCount = MAXIMUM_MAP_SIZE_TECHNICAL * MAXIMUM_MAP_SIZE_TECHNICAL;
+
+    if (fullCheck)
+        _mapHighestTileHeightLoopPosition = 0;
+
+    if (_mapHighestTileHeightLoopPosition == mapTileCount - 1)
```
ZehMatt, line 1563: You missed this one

RobertWHurst:
Anything else you guys would like me to refine here?

ZehMatt:
I think the code is fine now, this needs to be tested @OpenRCT2/developers

src/openrct2/world/Map.cpp
```diff
@@ -1539,6 +1548,104 @@ void map_update_tiles()
...
+        _mapHighestTileHeightLoopPosition = 0;
+
+    if (_mapHighestTileHeightLoopPosition == mapTileCount)
+        return;
+
+    // If we're doing a full check of the map then zero out the loop
+    // position to start at tile zero, checking all tiles before returning.
+    // Otherwise check a limited number of tiles per each time the
+    // function is called.
+    int32_t tilesToCheck = fullCheck ? mapTileCount : mapTileCount / 64;
```
ZehMatt, line 1570:
```suggestion
    int32_t tilesToCheck = fullCheck ? mapTileCount : mapTileCount / 8;
```
I've done some testing and 64 seems to be too large, the update takes way too long. The effect is first apparent after 2,5 seconds.

8 seems to be a good value which would mean also 8 * 40ms to update.
```
INFO[F:\C++\OpenRCT2\src\openrct2\world\Map.cpp:1626 (MapUpdateHeightCache)]: MapUpdateHeightCache ran in 0.00004600 secs
INFO[F:\C++\OpenRCT2\src\openrct2\world\Map.cpp:1626 (MapUpdateHeightCache)]: MapUpdateHeightCache ran in 0.00005200 secs
INFO[F:\C++\OpenRCT2\src\openrct2\world\Map.cpp:1626 (MapUpdateHeightCache)]: MapUpdateHeightCache ran in 0.00004600 secs
INFO[F:\C++\OpenRCT2\src\openrct2\world\Map.cpp:1626 (MapUpdateHeightCache)]: MapUpdateHeightCache ran in 0.00004400 secs
INFO[F:\C++\OpenRCT2\src\openrct2\world\Map.cpp:1626 (MapUpdateHeightCache)]: MapUpdateHeightCache ran in 0.00004200 secs
INFO[F:\C++\OpenRCT2\src\openrct2\world\Map.cpp:1626 (MapUpdateHeightCache)]: MapUpdateHeightCache ran in 0.00003700 secs
INFO[F:\C++\OpenRCT2\src\openrct2\world\Map.cpp:1626 (MapUpdateHeightCache)]: MapUpdateHeightCache ran in 0.00002800 secs
```
Its fast enough not to worry about performance on my end. But the map size may or may not increase in the future.

ZehMatt: Otherwise this seems to be working well, after this change its ready to be merged as far I'm concerned.

ZehMatt (changes requested):

src/openrct2/world/Map.cpp
RobertWHurst: Sounds good. Thanks for giving it a good kick of the tires. I'll make the change this evening :)

ZehMatt (approved):

tupaschoal (approved):

RobertWHurst:
@IntelOrca are you happy with the implementation?

IntelOrca (approved):

ZehMatt:
This still needs a changelog entry.

src/openrct2/actions/LargeSceneryPlaceAction.cpp
```diff
@@ -275,6 +275,7 @@ GameActions::Result::Ptr LargeSceneryPlaceAction::Execute() const
             {
                 wall_remove_at({ curTile, zLow, zHigh });
             }
+            MapInvalidateHeightCache();
```
Gymnasiast, line 278: Should this check for `if (!(GetFlags() & GAME_COMMAND_FLAG_GHOST))` ?

RobertWHurst: It does; It's wrapped with an if block checking that 7 lines above.

RobertWHurst:
@Gymnasiast Thanks for rebasing the branch

Gymnasiast (commented):
Code looks good, could do with testing.

Gymnasiast:
Could you also amend the changelog? `Fix: [#2876] High roller coasters at the map edge are cut off from view (original bug).`

RobertWHurst:
Sure thing! I'll do that in the next hour.

[Robert Hurst - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=w7egd)	[w7egd]

On February 8, 2021 at 9:40 GMT, Michael Steenbeek <notifications@github.com> wrote:

Could you also amend the changelog? Fix: [#2876] High roller coasters at the map edge are cut off from view (original bug).

RobertWHurst:
Sorry guys, I got a bit busy today. I'll update it soon.

[Robert Hurst - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=w7nes)	[w7nes]

On February 8, 2021 at 18:30 GMT, Robert Hurst <robertwhurst@gmail.com> wrote:

Sure thing! I'll do that in the next hour.

[Robert Hurst - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=w7egd)	[w7egd]

On February 8, 2021 at 9:40 GMT, Michael Steenbeek <notifications@github.com> wrote:

Could you also amend the changelog? Fix: [#2876] High roller coasters at the map edge are cut off from view (original bug).

RobertWHurst:
@Gymnasiast Do I need to do the testing? I've been playing with the change for about a week with about five different parks. No issues that I've seen so far.

Gymnasiast:
@RobertWHurst You should always test your changes if you create a PR. The `testing required` label is applied when _we_ have to do some testing of our own. I plan to do that tonight and we'll get this in if everything works correctly.

Gymnasiast (changes requested):
With this change, I can now scroll way too far into the void, even to the point nothing is on screen any more:

Fourth Park 2021-02-15 18-36-49

I can upload the park I tested with if desired, although it isn't anything special.

RobertWHurst:
I'm aware that this can happen. My justification was that it's still fairly easy to find the map again as you can't go too far from the edge and that the possibility of going off the map a little bit was worth the ability to see and use parts of the map that were previously inaccessible.

That all said, I'll add some logic to take the tile coordinates into account when calculating the max height. That way only the tiles that would normally be cut off increase the scroll limits. Note that even with such a change, a single tile, like one containing an observation deck, can cause a similar situation. Say I built a tower in the top corner, then the left or right. The camera can still go off the screen. The only way to fix that would be to have nonuniform scroll limits, and that sounds like a mess.

Is it not worse to have elements stuck outside of view rather than the map a little bit beyond the view?

RobertWHurst:
Alright, so assuming that the map must always be in view, I need to also take the x and y position of the tile into account when determining the highest tile offset. Since the rotation of the view will also affect this calculation, I'll probably need four values instead of one; one max tile height per screen rotation. I'll make the changes over the next week, and see how it turns out. It might mean I can simplify some of the logic in MapGetEdgeLimits, but we'll see. Note that this will not cover the case where I have a rather tall tile or ride in one corner of the screen. Say I have an observation tower in the top corner, then scroll to the left or right, I will still be able to scroll off of the map because the scroll limit is uniform.

RobertWHurst:
I think I'm just going to close this for now. If you guys think it's worth working on then let me know, I have some ideas but I don't want to work on this anymore if there's no interest in it.

In any case, those who helped me out working on this answering my questions and looking at my code, thanks for that. It's been neat getting a look at the innards and getting a better understanding of what you're all up to. Definitely impressive stuff.

Gymnasiast:
It's not that there is no interest per se, it's that there was a bug in the implementation. We unfortunately lack the time to take over all abandoned PRs and will generally only do so when the issue is pressing or the team member fancies it.

PR #13736, opened 2021-01-09
Refactor to avoid unnecessary copies

skdltmxn:
A bit of optimization by avoiding unnecessary copies

### DIFF
--- a/src/openrct2-ui/scripting/ScUi.hpp
+++ b/src/openrct2-ui/scripting/ScUi.hpp
@@ -191,8 +191,8 @@ namespace OpenRCT2::Scripting
             try
             {
                 auto plugin = _scriptEngine.GetExecInfo().GetCurrentPlugin();
-                const auto& title = desc["title"].as_string();
-                const auto& description = desc["description"].as_string();
+                auto title = desc["title"].as_string();
+                auto description = desc["description"].as_string();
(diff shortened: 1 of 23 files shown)


janisozaur (approved):

src/openrct2-ui/scripting/ScUi.hpp
```diff
@@ -191,8 +191,8 @@ namespace OpenRCT2::Scripting
             try
             {
                 auto plugin = _scriptEngine.GetExecInfo().GetCurrentPlugin();
-                auto title = desc["title"].as_string();
-                auto description = desc["description"].as_string();
+                const auto& title = desc["title"].as_string();
+                const auto& description = desc["description"].as_string();
```
tupaschoal, line 195: Does it make sense to have these as `const &`? I'm assuming that the `as_string()` method does a convesion and returns that object and (likely) wouldn't have an actual reference to a string to return. Is that true?

skdltmxn:
```c++
inline const std::string& as_string() const {
    if (mType != STRING)
	throw DukException() << "Expected string, got " << type_name();
    return mString;
}
```

`as_string` just returns const ref to its member variable... but, now I found that `operator []` returns a new copy of DukValue!
I will revert those two lines. Thanks :)

tupaschoal (approved):

PR #13737, opened 2021-01-09
Start window refactor framework

IntelOrca:
This prepares a new framework for the in-game windows.

`Window` is a new class to replace `rct_window`, however the window and UI code is still heavily split between libopenrct2 and libopenrct2ui. In order to ease the transition, `Window` is in libopenrct2ui and will contain all new fields, methods. `rct_window` remains as the name to use within libopenrct2 where necessary.

Each window should have a new class which inherits `Window`. Each event should use a virtual function prefixed with `On` and member methods on Window should be used over old functions that pass around `rct_window*`. Make new members on `Window` as necessary.

The Cheats window has been moved to the new framework. When this PR is merged, other windows can commence.

There are no helper routines for doing tabs right now, so events will have to check what page the window is currently showing and then call the relevant sub-tab event for non-common widgets. See how the Cheats window does it. This is because you can't just swap to a different event handler table anymore.

Any static fields declared in the old window source should be moved to private fields within the new class. These can be allocated when the class is instantiated, and destroyed when the class is deleted. It might also be a good idea to stop using some of the rct_window fields and create new dedicated ones within the sub class.

### DIFF
--- a/src/openrct2-ui/interface/Window.cpp
+++ b/src/openrct2-ui/interface/Window.cpp
@@ -87,8 +87,8 @@ static bool WindowFitsOnScreen(const ScreenCoordsXY& loc, int32_t width, int32_t
 }
 
 rct_window* WindowCreate(
-    const ScreenCoordsXY& screenCoords, int32_t width, int32_t height, rct_window_event_list* event_handlers,
-    rct_windowclass cls, uint16_t flags)
+    std::unique_ptr<rct_window>&& wp, rct_windowclass cls, const ScreenCoordsXY& pos, int32_t width, int32_t height,
+    uint16_t flags)
 {
     // Check if there are any window slots left
     // include WINDOW_LIMIT_RESERVED for items such as the main viewport and toolbars to not appear to be counted.
@@ -129,7 +129,7 @@ rct_window* WindowCreate(
         }
     }
 
-    auto itNew = g_window_list.insert(itDestPos, std::make_unique<rct_window>());
+    auto itNew = g_window_list.insert(itDestPos, std::move(wp));
     auto w = itNew->get();
 
     // Setup window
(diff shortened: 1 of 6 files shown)


src/openrct2/interface/Window.h
```diff
@@ -173,15 +173,15 @@ struct rct_viewport
...
-    uint16_t flags;          // 0x00
-    uint16_t h_left;         // 0x02
-    uint16_t h_right;        // 0x04
-    uint16_t h_thumb_left;   // 0x06
-    uint16_t h_thumb_right;  // 0x08
-    uint16_t v_top;          // 0x0A
-    uint16_t v_bottom;       // 0x0C
-    uint16_t v_thumb_top;    // 0x0E
-    uint16_t v_thumb_bottom; // 0x10
+    uint16_t flags{};          // 0x00
```
Broxzier, line 176: This and all similar changes are there to ensure they are zero initialized?

Broxzier (approved):
The changes seem alright and everything worked as expected.

src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -509,757 +442,781 @@ static rct_string_id window_cheats_page_titles[] = {
...
-        case WIDX_SET_MONEY:
-            CheatsSet(CheatType::SetMoney, _moneySpinnerValue);
-            break;
-        case WIDX_CLEAR_LOAN:
-            CheatsSet(CheatType::ClearLoan, CHEATS_MONEY_DEFAULT);
-            break;
-    }
-}
+        // Set correct active tab
+        for (int32_t i = 0; i < 7; i++)
```
Gymnasiast, line 536: Naked number.

src/openrct2/interface/Window.h
IntelOrca: Yeah, although I could revert it now as this was when I wanted to try and do stuff in the sub class constructor before realising that was still too early and needed to make OnOpen. The unions make it difficult to fully zero the struct with initialisers.

src/openrct2-ui/interface/Window.cpp
```diff
@@ -87,8 +87,8 @@ static bool WindowFitsOnScreen(const ScreenCoordsXY& loc, int32_t width, int32_t
 }
 
 rct_window* WindowCreate(
-    const ScreenCoordsXY& screenCoords, int32_t width, int32_t height, rct_window_event_list* event_handlers,
-    rct_windowclass cls, uint16_t flags)
+    std::unique_ptr<rct_window> wp, rct_windowclass cls, const ScreenCoordsXY& pos, int32_t width, int32_t height,
```
ZehMatt, line 90: Shouldnt this be 
```suggestion
    std::unique_ptr<rct_window>&& wp, rct_windowclass cls, const ScreenCoordsXY& pos, int32_t width, int32_t height,
```

IntelOrca: I have done it, although not sure what difference it has made. It must have been moving it since it is a unique_ptr which can't be copied.

ZehMatt: It probably used the move constructor, but for clarity it should be there I think.

hjk321:
Hi, this issue has a very large number of open related issues. Did this merge resolve all of those or are they still in need of doing?

duncanspumpkin:
If you read the related issues you would see that this pull developed the framework and the individual issues are to use the framework on each of the individual Windows.

PR #13738, opened 2021-01-09
Implement ride music objects and refactor

IntelOrca:
Read and use the JSON music objects for ride music implementation. Supports both vanilla and custom music objects. More music objects can be added via object manager if you compile-in the hidden tabs.

Refactors a lot of the ride music logic into a new source file: RideAudio.


src/openrct2/ride/RideAudio.h
```diff
@@ -0,0 +1,22 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2020 OpenRCT2 developers
```
Gymnasiast, line 2: 2021

src/openrct2/object/MusicObject.cpp
```diff
@@ -0,0 +1,169 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2019 OpenRCT2 developers
```
Gymnasiast, line 2: 2021

src/openrct2/core/Path.hpp
```diff
@@ -16,9 +16,9 @@
 namespace Path
 {
     utf8* Append(utf8* buffer, size_t bufferSize, const utf8* src);
-    std::string Combine(const std::string& a, const std::string& b);
+    std::string Combine(const std::string_view& a, const std::string_view& b);
```
Gymnasiast, line 19: It shouldn't be necessary to do `const std::string_view&` IIRC.

```suggestion
    std::string Combine(std::string_view a, std::string_view b);
```

tupaschoal (commented):
Needs some rebasing to bring it up to speed.

IntelOrca:
Already? I only created it last night and there are now 5 conflicts...

src/openrct2/core/Path.cpp
```diff
@@ -30,12 +30,45 @@ namespace Path
         return safe_strcat_path(buffer, src, bufferSize);
     }
 
-    std::string Combine(const std::string& a, const std::string& b)
+    static constexpr bool IsPathSeperator(char c)
```
Gymnasiast, line 33:
```suggestion
    static constexpr bool IsPathSeparator(char c)
```

src/openrct2-ui/audio/FileAudioSource.cpp
```diff
@@ -202,4 +202,47 @@ namespace OpenRCT2::Audio
         }
         return source;
     }
+
+    IAudioSource* AudioSource::CreateStreamFromWAV(std::unique_ptr<std::istream> stream)
+    {
+        using streamptr = std::unique_ptr<std::istream>*;
```
tupaschoal, line 208: I'm a bit confused with the usage of a pointer to a `unique_ptr`

src/openrct2-ui/audio/FileAudioSource.cpp
```diff
@@ -202,4 +202,47 @@ namespace OpenRCT2::Audio
...
+        auto data = new std::unique_ptr<std::istream>(std::move(stream));
```
tupaschoal, line 209: Isn't the `new` here unnecessary? Sounds like you could just move directly

src/openrct2/core/Zip.cpp
```diff
@@ -106,66 +167,183 @@ class ZipArchive final : public IZipArchive
...
+            throw std::runtime_error("File does not exist.");
+        }
     }
 
 private:
-    /**
-     * Normalises both the given path and the stored paths and finds the first match.
-     */
-    zip_int64_t GetIndexFromPath(std::string_view path) const
+    class ifilestream final : public std::istream
```
ZehMatt, line 207: Can't you adapt it from the existing FileStream?

src/openrct2/core/ZipAndroid.cpp
```diff
@@ -127,6 +133,34 @@ class ZipArchive final : public IZipArchive
     {
         STUB();
     }
+
+private:
+    class memstream final : public std::istream
```
ZehMatt, line 138: We also have MemoryStream, is a new class really necessary?

src/openrct2/ride/RideAudio.h
```diff
@@ -0,0 +1,22 @@
...
+ *****************************************************************************/
+
+#pragma once
+
+#include <cstdint>
+
+struct CoordsXYZ;
+struct Ride;
+
+constexpr uint8_t TUNE_ID_NULL = 0xFF;
```
ZehMatt, line 17: Might be worth having a namespace such as 'RideAudio', all the prefixes are basically that.

src/openrct2/core/Zip.cpp
IntelOrca: When and when should we not try to use the C++ classes where possible? Shouldn't we start replacing FileStream with fstream?

ZehMatt: Yes, what actually meant is we should stick to one thing.

IntelOrca: Okay, let this be the start of using built-in C++ streams.

ZehMatt: We should first check how well the performance would be, there are some optimizations in MemoryStream to better handle the desync debugging its still pretty slow so we have to see how that holds up at the end.

src/openrct2-ui/audio/FileAudioSource.cpp
ZehMatt: See line 214.

ZehMatt (changes requested):
I have tested it and it seems to be working, there is some code that should be changed.

src/openrct2/core/Path.cpp
```diff
@@ -30,12 +30,45 @@ namespace Path
...
+            if (IsPathSeparator(bBegin))
+            {
+                return std::string(a) + std::string(b);
+            }
+            else
+            {
+                return std::string(a) + PATH_SEPARATOR + std::string(b);
+            }
+        }
     }
```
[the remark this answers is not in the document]
IntelOrca: Your new code does not the case of both paths having a separator so you will end up with two separators in a row.

src/openrct2/object/MusicObject.h
```diff
@@ -0,0 +1,64 @@
...
+class MusicObjectTrack
+{
+public:
+    std::string Name;
+    ObjectAsset Asset;
+
+    /**
+     * The number of PCM bytes to seek per game tick when the music is playing offscreen.
+     */
+    size_t BytesPerTick;
```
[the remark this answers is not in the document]
IntelOrca: The original code used the same bytes per tick for every music track, but this is not ideal. I therefore want to keep this variable as a reminder that we should at some point calculate it.

IntelOrca:
@ZehMatt I have changed to using `IStream`, `FileStream` and `MemoryStream`.

src/openrct2/core/Path.cpp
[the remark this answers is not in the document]
ZehMatt: Right, lets leave it for some other time.

ZehMatt (approved):

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -105,10 +105,10 @@ static constexpr const ObjectPageDesc ObjectSelectionPages[] = {
     { STR_OBJECT_SELECTION_WATER,                       SPR_TAB_WATER,              false },
 
     // Currently hidden until new save format arrives:
-    // { STR_OBJECT_SELECTION_TERRAIN_SURFACES,            SPR_G2_TAB_LAND,            false },
-    // { STR_OBJECT_SELECTION_TERRAIN_EDGES,               SPR_G2_TAB_LAND,            false },
-    // { STR_OBJECT_SELECTION_STATIONS,                    SPR_TAB_PARK,               false },
-    // { STR_OBJECT_SELECTION_MUSIC,                       SPR_TAB_MUSIC_0,            false },
+    { STR_OBJECT_SELECTION_TERRAIN_SURFACES,            SPR_G2_TAB_LAND,            false },
```
Gymnasiast, line 108: Enabling these tabs would be fine in my book, as long as we don't allow changes to these tabs. It means that people are aware of the possibilities.

Gymnasiast (changes requested):
The music styles are incorrectly imported. For example, “Avalanche” in Blackpool Pleasure Beach has the “Medieval Style” selected, but this is imported as “Urban style”. This applies to both SC4 and SV6 saves of BPB.

src/openrct2-ui/windows/EditorObjectSelection.cpp
IntelOrca: I didn't mean for this to be included in the PR, I just wanted to test it.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -4938,15 +4941,20 @@ static void window_ride_colour_scrollpaint(rct_window* w, rct_drawpixelinfo* dpi
 #pragma region Music
 
 static constexpr const uint8_t MusicStyleOrder[] = {
```
Gymnasiast, line 4943: When the new save format is here, the constants no longer mean anything. I think we should get rid of this hardcoded order and just sort the dropdown alphabetically.

Gymnasiast (approved):

PR #13740, opened 2021-01-09
Import RCT1 tile elements by x/y

Gymnasiast:
This makes the import cleaner and is a prerequisite to split up path tile elements and path addition tile elements.

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -1974,24 +1974,8 @@ private:
         gMapBaseZ = 7;
 
         // Build tile pointer cache (needed to get the first element at a certain location)
-        constexpr const uint16_t MaxTileElementPointers = RCT1_MAX_MAP_SIZE * RCT1_MAX_MAP_SIZE;
-        RCT12TileElement* tilePointers[MaxTileElementPointers];
-        for (size_t i = 0; i < MaxTileElementPointers; i++)
-        {
-            tilePointers[i] = TILE_UNDEFINED_TILE_ELEMENT;
-        }
-
-        RCT12TileElement* tileElement = _s4.tile_elements;
-        RCT12TileElement** tile = tilePointers;
-        for (size_t y = 0; y < RCT1_MAX_MAP_SIZE; y++)
-        {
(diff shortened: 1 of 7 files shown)


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1973,26 +1973,62 @@ class S4Importer final : public IParkImporter
...
+        // Build tile pointer cache (needed to get the first element at a certain location)
+        constexpr const uint16_t MaxTileElementPointers = RCT1_MAX_MAP_SIZE * RCT1_MAX_MAP_SIZE;
+        RCT12TileElement* tilePointers[MaxTileElementPointers];
+        for (size_t i = 0; i < MaxTileElementPointers; i++)
         {
-            auto src = &_s4.tile_elements[index];
-            auto dst = &gTileElements[index + dstOffset];
-            if (src->base_height == RCT12_MAX_ELEMENT_HEIGHT)
+            tilePointers[i] = TILE_UNDEFINED_TILE_ELEMENT;
+        }
```
IntelOrca, line 1982: You can just do this:
```
RCT12TileElement* tilePointers[MaxTileElementPointers]{};
```

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1973,26 +1973,62 @@ class S4Importer final : public IParkImporter
     {
         gMapBaseZ = 7;
 
-        for (uint32_t index = 0, dstOffset = 0; index < RCT1_MAX_TILE_ELEMENTS; index++)
+        // Build tile pointer cache (needed to get the first element at a certain location)
+        constexpr const uint16_t MaxTileElementPointers = RCT1_MAX_MAP_SIZE * RCT1_MAX_MAP_SIZE;
+        RCT12TileElement* tilePointers[MaxTileElementPointers];
```
duncanspumpkin, line 1978: You should break this up into a few functions save this function from getting crazy big. One for setup, one for access. That sort of thing.

Gymnasiast:
Pushed a commit that should address both concerns. I made it a template because I will need to do the same for S6 import.

IntelOrca (approved):

duncanspumpkin (approved):

PR #13751, opened 2021-01-10
Make entities a singly linked list

duncanspumpkin:
I can't see there being much point there being a doubly linked list for entities as the previous is only used when deleting/creating entities which is something that happens infrequently. Might be an idea to profile this one.

### DIFF
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -958,6 +958,30 @@ void S6Exporter::ExportSprites()
         _s6.sprite_lists_head[i] = gSpriteListHead[i];
         _s6.sprite_lists_count[i] = gSpriteListCount[i];
     }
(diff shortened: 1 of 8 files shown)


duncanspumpkin:
Note still need to rebuild the previous value for exporting.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -992,7 +992,6 @@ void S6Exporter::ExportSpriteCommonProperties(RCT12SpriteBase* dst, const Sprite
     dst->sprite_identifier = src->sprite_identifier;
     dst->next_in_quadrant = src->next_in_quadrant;
     dst->next = src->next;
-    dst->previous = src->previous;
```
duncanspumpkin, line 995: still needed

duncanspumpkin:
I'm going to close this in favour of just removing both next and previous in one go.

PR #13754, opened 2021-01-10
Remove next_in_quadrant

duncanspumpkin:
Another field that doesn't need to be located in the Entity base. We rebuild the spatial index on load so this field was never read from the SV6/4 file.

### DIFF
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -981,6 +981,27 @@ void S6Exporter::RebuildEntityLinks()
             }
         }
     }
+
+    // Rebuild next_in_quadrant linked list entity indexs
+    for (auto x = 0; x < 255; ++x)
+    {
+        for (auto y = 0; y < 255; ++y)
+        {
+            uint16_t previous = SPRITE_INDEX_NULL;
+            for (auto* entity : EntityTileList(TileCoordsXY{ x, y }.ToCoordsXY()))
+            {
+                if (previous != SPRITE_INDEX_NULL)
+                {
+                    _s6.sprites[previous].unknown.next_in_quadrant = entity->sprite_index;
+                }
+                previous = entity->sprite_index;
+            }
+            if (previous != SPRITE_INDEX_NULL)
+            {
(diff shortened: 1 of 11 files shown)


src/openrct2/world/Sprite.cpp
```diff
@@ -32,6 +33,7 @@ static rct_sprite _spriteList[MAX_SPRITES];
 static bool _spriteFlashingList[MAX_SPRITES];
 
 uint16_t gSpriteSpatialIndex[SPATIAL_INDEX_SIZE];
+std::array<std::vector<uint16_t>, SPATIAL_INDEX_SIZE> gSpriteSpatialIndex2;
```
duncanspumpkin, line 36: can delete the 2 and remove the old spatial index.

src/openrct2/world/Sprite.h
```diff
@@ -332,25 +332,75 @@ template<typename T, uint16_t SpriteBase::*NextList> class EntityIterator
     using iterator_category = std::forward_iterator_tag;
 };
 
+template<typename T> class EntityTileIterator
+{
+private:
+    const std::vector<uint16_t>& vec;
```
duncanspumpkin, line 338: only need the end iterator so might as well take two const iterators

duncanspumpkin:
Could do with doing some performance checking just to make sure it doesn't change anything.

duncanspumpkin:
Note still need to rebuild the next_in_quadrant value for exporting.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -990,7 +990,6 @@ void S6Exporter::ExportSprite(RCT2Sprite* dst, const rct_sprite* src)
 void S6Exporter::ExportSpriteCommonProperties(RCT12SpriteBase* dst, const SpriteBase* src)
 {
     dst->sprite_identifier = src->sprite_identifier;
-    dst->next_in_quadrant = src->next_in_quadrant;
```
duncanspumpkin, line 993: still needed

IntelOrca:
I notice your iterators are enumerating `Peep*`, should they be enumerating `Peep&` instead? It shouldn't ever return a null peep.

I think it would be good to make everything to do with the spatial index a class of its own. But that could be done in another PR.

PR #13756, opened 2021-01-10
Refactor guest list window to a class

IntelOrca:


Umdlye:
The page selection dropdown appears when there are 0 guests in the park, clicking it crashes the game.

IntelOrca:
@Umdlye thanks, now fixed!

Gymnasiast:
This is causing quite noticeable lag on my system, to the point of the entire screen freezing for a fraction of a second every few seconds.

Additionally, if I open Blackpool Pleasure Beach, the first tab has 3 pages, but the 3rd page is empty.

These two things do not happen on `develop`.

Gymnasiast (approved):

PR #13820, opened 2021-01-12
Import SV6 tile elements by x/y

Gymnasiast:


src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1031,27 +1031,49 @@ class S6Importer final : public IParkImporter
...
+                if (coords.x >= RCT2_MAXIMUM_MAP_SIZE_TECHNICAL || coords.y >= RCT2_MAXIMUM_MAP_SIZE_TECHNICAL)
+                {
+                    dstElement->ClearAs(TILE_ELEMENT_TYPE_SURFACE);
+                    dstElement->SetLastForTile(true);
+                    dstElement++;
+                    continue;
+                }
+
+                RCT12TileElement* srcElement = tilePointerIndex.GetFirstElementAt(coords);
+                do
```
Gymnasiast, line 1051: Needs null check.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1031,27 +1031,58 @@ class S6Importer final : public IParkImporter
 
     void ImportTileElements()
     {
-        for (uint32_t index = 0; index < RCT2_MAX_TILE_ELEMENTS; index++)
+        // Build tile pointer cache (needed to get the first element at a certain location)
+        auto tilePointerIndex = TilePointerIndex<RCT12TileElement>(RCT2_MAXIMUM_MAP_SIZE_TECHNICAL, _s6.tile_elements);
+
+        TileElement* dstElement = gTileElements;
```
duncanspumpkin, line 1037: What happens if we walk off the end of this array. This would be possible if the sv6 has deleted surface tiles to save a small amount of space.

Gymnasiast: I don't think that will happen. Even on SV6 files that have deleted surface tiles, there is always at least one other tile element - removing all tile element would just shift the rest of the tiles (and that case _is_ covered).

duncanspumpkin: Okay sounds fine.

Gymnasiast:
Tested this with some parks from NE (Thorpe Park, which has used up the tile element) and Wit's end (which uses deliberately hacked tile elements). Both parks still open correctly.

PR #13822, opened 2021-01-13
Avoid some allocations by using a fixed size vector

ZehMatt:
Also makes the code a bit easier to read. There are more places where we can use this, just one thing to note that its not just a pure flat array, it still keeps track of the count so this can not be replaced with raw array or std::array when the structure fields must remain POD.

### DIFF
--- a/src/openrct2/core/FixedVector.h
+++ b/src/openrct2/core/FixedVector.h
@@ -1,0 +1,176 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2021 OpenRCT2 developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include "Guard.hpp"
+
+#include <algorithm>
+#include <array>
+#include <utility>
+
+template<typename T, size_t MAX> class FixedVector
+{
+public:
+    using container = std::array<T, MAX>;
+    using iterator = typename container::iterator;
+    using const_iterator = typename container::const_iterator;
+    using reverse_iterator = typename container::reverse_iterator;
+    using const_reverse_iterator = typename container::const_reverse_iterator;
+    using value_type = typename container::value_type;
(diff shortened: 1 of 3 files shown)


src/openrct2/ride/Vehicle.h
```diff
@@ -155,6 +156,8 @@ struct rct_vehicle_info
 
 struct SoundIdVolume;
 
+using VehicleSoundParamList = FixedVector<OpenRCT2::Audio::VehicleSoundParams, OpenRCT2::Audio::MaxVehicleSounds>;
```
duncanspumpkin, line 159: Do we need to use the using here could we forward declare the class then can avoid adding the header

src/openrct2/ride/Vehicle.cpp
```diff
@@ -914,18 +914,20 @@ template<typename T> int32_t Train<T>::Mass()
     return totalMass;
 }
 
-bool Vehicle::SoundCanPlay() const
+using VehicleSoundParamList = FixedVector<OpenRCT2::Audio::VehicleSoundParams, OpenRCT2::Audio::MaxVehicleSounds>;
+
+static bool VehicleSoundCanPlay(Vehicle* vehicle)
```
duncanspumpkin, line 919: Will we do a follow on PR that moves all the sound updating into a separate class?

ZehMatt: Sounds like a good idea

IntelOrca: Perhaps wait until the music PR is merged first as that already moves the ride music out. Unless there should be RideAudio and VehicleAudio.

ZehMatt: I see, then I'll revert the vehicle part.

Gymnasiast:
I guess this is also a performance thing? Could you include benchmarks?

ZehMatt:
Initially it replaced the vector for the ride audio parameters but @IntelOrca is working on #13738 so I dropped the commit, there is no performance improvement right now. The vector is not just for performance but to also make the code a lot easier to grasp, there are a lot of areas where some sort of "list" is managed in fixed space, this is where the FixedVector can come in handy.

duncanspumpkin (approved):

Issue #8424, opened 2018-12-12
use boosters to change powered car speed


spacek531:
It would be useful if I could change the speed of powered cars at certain points in the ride, and allow for more creative ride design. Rides like Test Track and the Carsland ride come to mind as users of this feature.

To keep implementation costs at a minimum, here is my thought on implementation: when a powered vehicle rolls over a Booster piece, its `speed` property is changed to the speed value of the booster.

Additional implementation can include adding a special piece to powered rides:

- monorail
- suspended monorail
- miniature railway
- car ride
- monster trucks
- ghost train
- mine ride

This means that cars that are powered and spinning will have unusual behavior, but I am willing to live with that.

bdupree5:
Me and my team would like to work on this issue. We will post updates as we continue our work.

PR #13825, opened 2021-01-13
Powered speed controller

spacek531:
This modifies booster behavior to affect powered vehicle target speed when passing onto the piece. Boosters take on a new name when built as part of certain powered rides. The pieces inherit the graphics of the flat piece and are all hidden behind the 'all drawable track pieces' cheat.

https://user-images.githubusercontent.com/3454156/104439231-bdc76d80-5545-11eb-848f-2b2200f25be1.mp4

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2252,9 +2252,33 @@ static void window_ride_construction_invalidate(rct_window* w)
     if (_currentTrackCurve & RideConstructionSpecialPieceSelected)
     {
         stringId = RideConfigurationStringIds[_currentTrackCurve & ~RideConstructionSpecialPieceSelected];
-        if (stringId == STR_RAPIDS && (ride->type == RIDE_TYPE_MONSTER_TRUCKS || ride->type == RIDE_TYPE_CAR_RIDE))
+        if (stringId == STR_RAPIDS)
         {
-            stringId = STR_LOG_BUMPS;
+            switch (RideTypeDescriptors[ride->type].TrackBehaviours.RapidsBehaviour)
+            {
+                case RideRapidsBehaviour::Rapids:
+                    stringId = STR_RAPIDS;
+                    break;
+                case RideRapidsBehaviour::LogBumps:
+                    stringId = STR_LOG_BUMPS;
+                    break;
+            }
+        }
+        else if (stringId == STR_BOOSTER)
+        {
+            switch (RideTypeDescriptors[ride->type].TrackBehaviours.BoosterBehaviour)
+            {
+                case RideBoosterBehaviour::Booster:
+                    stringId = STR_BOOSTER;
+                    break;
+                case RideBoosterBehaviour::SpeedController:
+                    stringId = STR_SPEED_CONTROL;
+                    break;
+                default:
+                {
+                    // to make the build systems happy
+                }
+            }
         }
     }
     auto ft = Formatter::Common();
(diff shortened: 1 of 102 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -3355,7 +3378,16 @@ static void window_ride_construction_show_special_track_dropdown(rct_window* w,
         {
             auto ride = get_ride(_currentRideIndex);
             if (ride != nullptr && ride->type != RIDE_TYPE_SPINNING_WILD_MOUSE && ride->type != RIDE_TYPE_STEEL_WILD_MOUSE)
-                trackPieceStringId = STR_BOOSTER;
+            {
+                if (ride != nullptr
+                    && (ride->type == RIDE_TYPE_MINIATURE_RAILWAY || ride->type == RIDE_TYPE_CAR_RIDE
```
duncanspumpkin, line 3383: Don't add massive if checks like this. We have the ride type descriptor and it's associated flags for this sort of thing.

spacek531:
I don't know if all fields need to be filled out in the rtd. I added the booster field to all tracked rides and the rapids field to the rapids and car rides. Let me know if all the rtds need these fields populated.

spacek531:
Is it preferable to have a single struct in the rtd for the two things I added, or is it fine how it is now?

spacek531:
I'm happy with where this PR is at. Aside from any requested changes and utility stuff like the changelog, I think I'm done.

Generalcamo:
In the future, if we can get some art for this I wouldn't mind seeing this un-hidden. RCT3 did allow the setting of powered ride speeds, albeit that was globally and not per-track. If anything this is a more flexible alternative.

spacek531:
Considering re-implementing or rebasing this onto new-save-format. Marking as draft for now.

spacek531:
I have re-opened this pull request because I branched from this PR to extend the available track pieces, starting with separating the booster track piece into its own TrackTypeElem id. Originally I started it as a branch of `new-save-format` but it's possible to do the exact same thing without the new save format. This branch may or may not be a dependency for the other branch.

spacek531:
This PR will solve this issue https://github.com/OpenRCT2/OpenRCT2/issues/8424

spacek531:
Updating this history has become untenable, I am going to branch from develop and force-push onto this branch when I'm satisfied.

spacek531:
I think this branch is done, less any unforseen issues.

src/openrct2/ride/tracktypes/TrackDataVerify.cpp
```diff
@@ -0,0 +1,1335 @@
+#include "../Track.h"
```
duncanspumpkin, line 1: did you mean to include the track data files?

spacek531: For some reason these files got included in this branch, they are supposed to be for the track data refactor only.

duncanspumpkin:
Is it ever likely that someone would want a speed controller and a booster? I'm guessing no as speed controllers are for self powered vehicles.

spacek531:
I'm not sure what you are asking for. Speed controllers are boosters with a new name, like log bumps are to rapids.

spacek531:
Can someone look at this please? I'd like to get it merged soon.

ZehMatt:
This introduces quite a good amount of chances, I think it would be better to wait with new things after NSF.

spacek531:
It only appears to add a lot because it adds one field to the RTD. SV6 is irrelevant.

ZehMatt:
> It only appears to add a lot because it adds one field to the RTD. Save format is irrelevant: the vehicle's target speed is serialized in the vehicle already.

You are still tinkering with game logic, its not a matter of what is being saved.

spacek531:
Why does the save file matter? OpenRCT2 has made a number of logic changes w/o save file.

duncanspumpkin:
@spacek531 we get no end of abuse for absolutely every change that is made especially when it changes vehicle logic. There are so many hacks that rely on unexpected interactions it is very easy to accidentally break one of those. Don't get annoyed when we are hesitant over a change we are just trying to prevent regressions.
https://xkcd.com/1172/

src/openrct2/ride/RideData.h
```diff
@@ -348,7 +362,8 @@ constexpr const RideTypeDescriptor DummyRTD =
     SET_FIELD(BonusValue, 0),
     SET_FIELD(ColourPresets, DEFAULT_FLAT_RIDE_COLOUR_PRESET),
     SET_FIELD(ColourPreview, { static_cast<uint32_t>(SPR_NONE), static_cast<uint32_t>(SPR_NONE) }),
-    SET_FIELD(ColourKey, RideColourKey::Ride)
+    SET_FIELD(ColourKey, RideColourKey::Ride),
+    SET_FIELD(TrackBehaviours,{TrackElemTypeAlternateName::Default,TrackElemTypeAlternateName::LogBumps})
```
duncanspumpkin, line 366: all of the ridedata headers do not have clang format enabled so you need to space them out yourself. Could you go through and make it so that it follows the rest of the codebase. I know its a pain doing that but its also a pain looking at it.
e.g.:
```
SET_FIELD(TrackBehaviours, { TrackElemTypeAlternateName::Default, TrackElemTypeAlternateName::LogBumps }),
```

duncanspumpkin: This default is used for a lot of RTD's. Its a bit of an odd default as it refers to LogBumps i think it would be better if we had a:
```
constexpr RideTrackBehaviours _defaultTrackBehaviour = { TrackElemTypeAlternateName::Default, TrackElemTypeAlternateName::LogBumps };
```
in this header and refer to that for all of the RTD's that use the default. That makes it all easy to understand why we are talking about logbumps.

duncanspumpkin:
The changes are fine in my opinon. I suggest that we modify bpb to add a powered speed controller to one of its rides and create a new replay for it so that the behaviour is recorded and can never unconsciously be regressed. TBH we should probably create a bpb equivalent that is just filled with every known hack but thats a future task.

duncanspumpkin (approved):
But see comments suggest modifying bpb and breaking replays.

spacek531:
> @spacek531 we get no end of abuse for absolutely every change that is made especially when it changes vehicle logic. There are so many hacks that rely on unexpected interactions it is very easy to accidentally break one of those. Don't get annoyed when we are hesitant over a change we are just trying to prevent regressions.
> https://xkcd.com/1172/

You're right, I know there is the chance it could break someone's park. I got angry because I think the risk vs. reward is worth it and that my work to make this SV6 compatible was being dismissed out of hand.

src/openrct2/ride/RideData.h
spacek531: I chose Rapids as the default value for that field because that's what the TrackElemType is named. My thinking with the defaults was that any powered ride got the SpeedController and any water ride got the Rapids default.

ZehMatt:
> > @spacek531 we get no end of abuse for absolutely every change that is made especially when it changes vehicle logic. There are so many hacks that rely on unexpected interactions it is very easy to accidentally break one of those. Don't get annoyed when we are hesitant over a change we are just trying to prevent regressions.
> > https://xkcd.com/1172/
> 
> You're right, I know there is the chance it could break someone's park. I got angry because I think the risk vs. reward is worth it and that my work to make this SV6 compatible was being dismissed out of hand.

Nothing was dismissed, I just explained on why we are currently hesitant on changes in general, adding new features is not on our list of priorities currently, simple as that.

rik-smeets:
I'm closing this PR due to inactivity, feel free to reopen if you resume work on it.

PR #13831, opened 2021-01-14
Use specific type for peep list access

duncanspumpkin:
As mentioned on discord. The vast majority of accesses to these lists specifies the type so might as well split up the lists on type reducing type checking eventually. 

Wip. Will break replays. Need to test the mechanic changes.

### DIFF
--- a/src/openrct2/management/Award.cpp
+++ b/src/openrct2/management/Award.cpp
@@ -263,21 +263,15 @@ static bool award_is_deserved_best_staff(int32_t activeAwardTypes)
     if (activeAwardTypes & EnumToFlag(ParkAward::MostUntidy))
         return false;
 
-    auto peepCount = 0;
     auto staffCount = 0;
     auto staffTypeFlags = 0;
-    for (auto peep : EntityList<Peep>(EntityListId::Peep))
+    for (auto staff : EntityList<Staff>(EntityListId::Peep))
     {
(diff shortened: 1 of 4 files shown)


src/openrct2/management/Award.cpp
```diff
@@ -263,21 +263,15 @@ static bool award_is_deserved_best_staff(int32_t activeAwardTypes)
...
-        {
-            staffCount++;
-            staffTypeFlags |= (1 << static_cast<uint8_t>(peep->AssignedStaffType));
-        }
-        else
-        {
-            peepCount++;
-        }
+        staffCount++;
+        staffTypeFlags |= (1 << static_cast<uint8_t>(staff->AssignedStaffType));
```
ZehMatt, line 271:
```suggestion
        staffTypeFlags |= (1 << EnumValue(staff->AssignedStaffType));
```

tupaschoal (approved):
Seems ok

Gymnasiast:
> As mentioned on discord

I don't think I have seen that discussion, could you give me a quick summary?

duncanspumpkin:
>I'm going to split up the peep list into guest and staff. This will reduce the chance of mistaking one for the other and it's pretty much how we use the list anyway.

src/openrct2/management/Award.cpp
duncanspumpkin: Actually can simplify this further and remove that variable.

PR #13833, opened 2021-01-14
Fix #13832: Game action position is invalid on opening/closing a ride

### LINKED ISSUE #13832: Opening or closing a ride shows the wrong action location in multiplayer
When closing or opening a ride the last player action is incorrect, it looks like this:

openrct2_2021-01-14_10-21-497eaf76c7db18d27e7

Instead of this:

openrct2_2021-01-14_10-29-34cc5c2d5848645218e

Not sure how old this bug is, probably been a while.

ZehMatt:
Wrong variable has been used.

Closes #13832

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -34,6 +34,7 @@
 - Fix: [#13489] Mechanics continue heading to inspect broken down rides.
 - Fix: [#13510] [Plugin] list view scroll resets when items is set.
 - Fix: [#13574] Crash when a JSON object does not set `originalId`.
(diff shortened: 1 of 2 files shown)


Gymnasiast (approved):
Looks all right to me. I cannot test it right now, but I assume you did that yourself.

What was passed before as height, by the way? Just random garbage? If so, might have been a source of desyncs.

I guess this also needs a network version bump.

ZehMatt:
I should have mentioned, theres no need to bump up the network. It just used the wrong variable to calculate the height which was always x == LOCATION_NULL and always resulted a height of 0, its just relevant for the player window.

PR #13834, opened 2021-01-14
Fix replays not failing when last tick is the cause

ZehMatt:
With the current logic there is an edge case if the last tick has the faulty checksum. After the last tick the _mode is no longer set to Playing and IsPlaybackStateMismatching would always result false.

I changed it to wait for the playback to complete and then test if the state is mismatching. This might cause the CI to fail now with out of date replays.

Later me: Yes the CI did fail as expected, also updating the replay version in this PR to have them all working again.

### DIFF
--- a/test/tests/ReplayTests.cpp
+++ b/test/tests/ReplayTests.cpp
@@ -98,8 +98,9 @@ TEST_P(ReplayTests, RunReplay)
     while (replayManager->IsReplaying())
     {
         gs->UpdateLogic();
-        ASSERT_TRUE(replayManager->IsPlaybackStateMismatching() == false);
     }
+    ASSERT_FALSE(replayManager->IsReplaying());
+    ASSERT_FALSE(replayManager->IsPlaybackStateMismatching());
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


test/tests/ReplayTests.cpp
```diff
@@ -98,8 +98,9 @@ TEST_P(ReplayTests, RunReplay)
     while (replayManager->IsReplaying())
     {
         gs->UpdateLogic();
-        ASSERT_TRUE(replayManager->IsPlaybackStateMismatching() == false);
     }
+    ASSERT_FALSE(replayManager->IsReplaying());
+    ASSERT_FALSE(replayManager->IsPlaybackStateMismatching());
```
duncanspumpkin, line 103: Is there any point continuing the replay if there has been a mismatch

ZehMatt: I thought about that as well, I can stop the playback in CheckState I guess.

duncanspumpkin: Also is it possible for the checksum to be wrong at the start but then become correct in the subsequent tick? Not sure if that is possible.

ZehMatt: I don't think that's possible, they are checked after the update.

PR #13835, opened 2021-01-14
Reset Entity to 0 on removal

duncanspumpkin:
This is to force any use after remove bugs to the surface (hopefully.)


tupaschoal (approved):
Looks fine I guess.

PR #13841, opened 2021-01-15
Fix #13236: New ride type appears as new vehicle type in research

### LINKED ISSUE #13236: Researching new ride type results in "New vehicle now available" message instead

Gymnasiast:


tupaschoal (approved):
Seems ok

PR #13848, opened 2021-01-16
Introduce new shortcut engine

IntelOrca:
* Add tabs to shortcut keys window to improve grouping.
* Add plugin support for registering new shortcuts
* Shortcuts are now saved to `shortcuts.json`, plugin shortcuts and unused JSON keys are preserved.
* For now shortcuts are imported from `hotkeys.dat` if `shortcuts.json` does not exist.
* Multiple bindings now possible for each shortcut, although may require editing JSON file to do.
* Other input devices now supported for bindings: all buttons from mouse and gamepads.
* Toggle debug console can now be rebound.


src/openrct2-ui/input/InputManager.cpp
```diff
@@ -0,0 +1,385 @@
...
+    {
+        const auto& e = _events.front();
+        Process(e);
+        _events.pop();
+    }
+}
+
+void InputManager::Process(const InputEvent& e)
+{
+    auto& shortcutManager = GetShortcutManager();
```
ZehMatt, line 179: In my opinion the InputManager should only gather the state and let the other systems such as Chat and Console decide how to handle the input state, not the input manager. I'm thinking of either events or polling, so either
```
auto mouseDown = inputSystem->IsMouseDown(MOUSE_MIDDLE);
```
or
```
inputSystem->addHandler(inputState, []() { /* Callback */ });
```
Systems should be connected and not part of each other. If something else needs input handling it would just grow in code and complexity here.

IntelOrca: You still need this Process method for your second example then to call the callbacks. Right now, something needs to call the keyboard shortcut handler so that is why I have this. I am happy for you to change it to however you think it should be done afterwards.

ZehMatt: Yes, all systems require an Update/Process method. Also fair enough.

src/openrct2-ui/input/InputManager.h
```diff
@@ -0,0 +1,83 @@
...
+        bool GetState(const RegisteredShortcut& shortcut) const;
+        bool GetState(const ShortcutInput& shortcut) const;
+
+    public:
+        void QueueInputEvent(const SDL_Event& e);
+        void QueueInputEvent(InputEvent&& e);
+        void Process();
+    };
+
+    InputManager& GetInputManager();
```
ZehMatt, line 81: Shouldn't this be in the Context?

IntelOrca: Context or UiContext?

ZehMatt: Context

IntelOrca: InputManager is currently in libopenrct2ui, so I can't move it to Context. It doesn't make sense to either as any core code shouldn't be calling anything to do with input.

ZehMatt: Technically the interactive console requires input but I see that we use Linenoise so in this case it doesn't matter I guess, so UiContext it is.

IntelOrca: ok done

src/openrct2-ui/input/ShortcutInput.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+#include <SDL.h>
+#include <cstring>
+#include <openrct2/core/String.hpp>
+
+using namespace OpenRCT2::Ui;
+
+constexpr uint32_t UsefulModifiers = KMOD_SHIFT | KMOD_CTRL | KMOD_ALT | KMOD_GUI;
+
+static uint32_t ParseModifier(const std::string_view& text)
```
ZehMatt, line 20:
```suggestion
static uint32_t ParseModifier(const std::string_view text)
```

src/openrct2-ui/input/ShortcutInput.cpp
```diff
@@ -0,0 +1,407 @@
...
+    {
+        return KMOD_RGUI;
+    }
+    else
+    {
+        return 0;
+    }
+}
+
+static uint32_t ParseKey(const std::string_view& text)
```
ZehMatt, line 76:
```suggestion
static uint32_t ParseKey(const std::string_view text)
```

src/openrct2-ui/input/ShortcutInput.cpp
```diff
@@ -0,0 +1,407 @@
...
+        }
+        else
+        {
+            break;
+        }
+    }
+    return index;
+}
+
+ShortcutInput::ShortcutInput(const std::string_view& value)
```
ZehMatt, line 106:
```suggestion
ShortcutInput::ShortcutInput(const std::string_view value)
```

src/openrct2-ui/input/ShortcutIds.h
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+namespace OpenRCT2::Ui::ShortcutId
+{
+    // Interface / general
+    constexpr const char* InterfaceCloseTop = "interface.general.close_top";
```
ZehMatt, line 15: Might be better to use std::string_view for them, there is no reason not to.

src/openrct2-ui/input/ShortcutInput.cpp
```diff
@@ -0,0 +1,407 @@
...
+    if (keyCode != SDLK_UNKNOWN)
+    {
+        return keyCode;
+    }
+    return 0;
+}
+
+static size_t FindPlus(std::string_view s, size_t index)
+{
+    while (true)
```
ZehMatt, line 90: I think you can use find_last_of instead.

src/openrct2-ui/input/ShortcutInput.cpp
```diff
@@ -0,0 +1,407 @@
...
+            result += "JOY UP";
+        else if (Button & SDL_HAT_DOWN)
+            result += "JOY DOWN";
+        else
+            result += "JOY ?";
+    }
+    return result;
+}
+
+bool ShortcutInput::AppendModifier(std::string& s, const std::string_view& text, uint32_t left, uint32_t right) const
```
ZehMatt, line 325:
```suggestion
bool ShortcutInput::AppendModifier(std::string& s, const std::string_view text, uint32_t left, uint32_t right) const
```

src/openrct2-ui/input/ShortcutManager.cpp
```diff
@@ -0,0 +1,443 @@
...
+            }
+        }
+    }
+
+    Json::WriteToFile(path, root);
+}
+
+std::string_view ShortcutManager::GetLegacyShortcutId(size_t index)
+{
+    static constexpr const char* LegacyMap[] = {
```
ZehMatt, line 354:
```suggestion
    static constexpr std::string_view LegacyMap[] = {
```

src/openrct2-ui/input/ShortcutManager.h
```diff
@@ -0,0 +1,144 @@
...
+    public:
+        std::string Id;
+        rct_string_id LocalisedName = STR_NONE;
+        std::string CustomName;
+        std::vector<ShortcutInput> Default;
+        std::vector<ShortcutInput> Current;
+        std::function<void()> Action;
+
+        RegisteredShortcut() = default;
+        RegisteredShortcut(const std::string_view& id, std::string_view name, const std::function<void()>& action)
```
ZehMatt, line 61:
```suggestion
        RegisteredShortcut(const std::string_view id, std::string_view name, const std::function<void()>& action)
```

src/openrct2-ui/input/ShortcutManager.h
```diff
@@ -0,0 +1,144 @@
...
+            : Id(id)
+            , CustomName(name)
+            , Action(action)
+        {
+        }
+
+        RegisteredShortcut(const std::string_view& id, rct_string_id localisedName, const std::function<void()>& action)
```
ZehMatt, line 68:
```suggestion
        RegisteredShortcut(const std::string_view id, rct_string_id localisedName, const std::function<void()>& action)
```

src/openrct2-ui/input/ShortcutManager.h
```diff
@@ -0,0 +1,144 @@
...
+            : Id(id)
+            , LocalisedName(localisedName)
+            , Action(action)
+        {
+        }
+
+        RegisteredShortcut(
+            const std::string_view& id, rct_string_id localisedName, const std::string_view& defaultChord,
```
ZehMatt, line 76:
```suggestion
            const std::string_view id, rct_string_id localisedName, const std::string_view defaultChord,
```

src/openrct2-ui/input/ShortcutManager.h
```diff
@@ -0,0 +1,144 @@
...
+            , LocalisedName(localisedName)
+            , Default({ defaultChord })
+            , Current(Default)
+            , Action(action)
+        {
+        }
+
+        RegisteredShortcut(
+            const std::string_view& id, rct_string_id localisedName, const std::string_view& defaultChordA,
+            const std::string_view& defaultChordB, const std::function<void()>& action)
```
ZehMatt, line 88:
```suggestion
            const std::string_view id, rct_string_id localisedName, const std::string_view defaultChordA,
            const std::string_view defaultChordB, const std::function<void()>& action)
```

ZehMatt (dismissed):
Looking overall alright, few minor things. Also I tested it and it seems to be working as far as I can tell.

src/openrct2-ui/input/ShortcutInput.cpp
IntelOrca: This function is to find the next plus (excluding ones we want to ignore) from index, not the last plus.

AaronVanGeffen:
I tested this very briefly. The tabs are a nice touch; this makes things a little more organised. However, I noticed the key bindings are no longer localisable. Looking at the code, this seems to have been left as a todo. However, this can be considered a regression from #10147.

Compare the state after #10147:
Screenshot_20191026_204543

To this PR:
Diamond Heights 2021-02-07 14-26-58_

IntelOrca:
@AaronVanGeffen thanks, localisation should now be improved

AaronVanGeffen:
@IntelOrca Thanks! I haven't yet tested it, but the diff looks promising.

I notice there's no string yet for the `CMD` modifier on Macs. Is this key still handled?

IntelOrca:
@AaronVanGeffen not sure, I can't really test that but it might be `GUI`?

AaronVanGeffen:
@IntelOrca I'll gladly take it for a spin when a macOS artefact is available here. Please remind me if I forget.

AaronVanGeffen:
It does indeed show up as "GUI":

Screenshot 2021-02-08 at 22 07 49

IntelOrca:
@AaronVanGeffen Okay, I have renamed GUI to Cmd and made the rest of the names title case.

AaronVanGeffen:
The changes look good to me. I can no longer formally review your PR, though. :slightly_smiling_face:

Gymnasiast:
Appears to work correctly.

Some notes:
- The default shortcut no longer opens the console if you use deadkeys, but this is acceptable, as I can still set it to use "dead backtick".
- However, nothing appears in the list if you set it to that. It's an edge case and not a showstopper, but should be fixed at some point.
- I would move the four "move view up/down/left/right" to the second tab (eye).
- I'd move "rotate object" to the third tab.

data/language/en-GB.txt
```diff
@@ -3665,6 +3665,27 @@ STR_6408    :Please install “innoextract” to extract GOG Installer, then res
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
 STR_6410    :Zoom in/out
 STR_6411    :Show buttons for zooming in and out in the toolbar
+STR_6412    :NumPad Return
```
Gymnasiast, line 3668:
```suggestion
STR_6412    :NumPad Enter
```

The key on the numpad is always called Enter. (The Return key is above the right Shift key and also often called “Enter”.)

Gymnasiast:
To do:
- [ ] Also bind mouse buttons 4 and 5 to rotation (see https://github.com/OpenRCT2/OpenRCT2/pull/14110)
- [ ] Move the four "move view up/down/left/right" to the second tab (eye)
- [ ] Move "rotate object" to the third tab

src/openrct2-ui/UiContext.cpp
```diff
@@ -963,6 +942,32 @@ class UiContext final : public IUiContext
             return;
         }
     }
+
+    InputEvent GetInputEventFromSDLEvent(SDL_Event& e)
```
Gymnasiast, line 946: Would it make sense to do `const SDL_Event&` here?

src/openrct2-ui/UiContext.cpp
```diff
@@ -963,6 +942,32 @@ class UiContext final : public IUiContext
...
+    {
+        InputEvent ie;
+        ie.DeviceKind = InputDeviceKind::Keyboard;
+        ie.Modifiers = e.key.keysym.mod;
+        ie.Button = e.key.keysym.sym;
+
+        // Handle dead keys
+        if (ie.Button == SDLK_SCANCODE_MASK)
```
Gymnasiast, line 954: This could alternatively be written as `if (ie.Button == (SDLK_SCANCODE_MASK | 0))`, which is a little bit clearer.

IntelOrca:
I haven't moved "rotate object" to the third tab because it would be the only item in its own group and the intention of that tab was for window specific commands.

src/openrct2-ui/UiContext.cpp
IntelOrca: Yes, it should be `const`.

IntelOrca:
```suggestion
    InputEvent GetInputEventFromSDLEvent(const SDL_Event& e)
```

src/openrct2-ui/UiContext.cpp
IntelOrca:
```suggestion
        if (ie.Button == (SDLK_SCANCODE_MASK | 0))
```

Gymnasiast: @AaronVanGeffen It's because the `0` part is the most important bit.

Gymnasiast (approved):

Issue #4605, opened 2016-10-12
Custom Water not displaying correctly


RealRexen:
OS: Win10
Version 0.0.5
Build 6cb9297
This image here, http://i.imgur.com/wd9f1SX.png shows the water in the steam version of the game. This image is shows it in OpenRCT2 http://i.imgur.com/PBcIa5E.png Link to said object https://1drv.ms/u/s!AnIDdSY_cTSAvzBxiOLfz4dJfHGO for testing/debugging.
The odd thing is when the  object management window gets reopened the water then displays correctly.

Nubbie:
Cowl avatar, but also, recommend zipping up the files to .zip/.rar and post them in here for easy access :+1:

RealRexen:
Sorry this was a re-post from the subreddit .

kingcailan:
Have you tried changing the renderer? If you're using OpenGL, it changes the way water renders. I've noticed this in the vanilla water, I haven't tried any custom water yet though. I'll try it when I ge home and update you then.

RealRexen:
I use only software or software with hardware never OpenGL.

autofire372:
This is still an issue, and occurs with the pink water DAT that Chris Sawyer released a while back.

AaronVanGeffen:
Has this been resolved in the mean time?

RealRexen:
Still broken 
![Forest Frontiers 2020-06-09 13-52-55](https://user-images.githubusercontent.com/18408903/84182745-dffd8800-aa58-11ea-9513-853b6cd1000d.png)

AaronVanGeffen:
The object linked on OneDrive in the original post has since expired. Could you attach it here? (GitHub will accept it if you zip it first.)

RealRexen:
Here you go
[rexen_s_water_pack.zip](https://github.com/OpenRCT2/OpenRCT2/files/4754008/rexen_s_water_pack.zip)

Gymnasiast:
Do I understand correctly that it gets glitchy after you changed the water type?

PR #13855, opened 2021-01-17
Fix #11912: Update active palette on Object Selection

### LINKED ISSUE #11912: Switching between palettes requires to reopen object selection in order to load the new palette
**OS:** Windows 10
**Version:** 0.2.6-798
**Commit/Build:** g22d099d6b

Switching between two palettes requires you to reopen the object selection window and go to the water tab in order to correctly load the new palette you've chosen. In between loading the new palette it switches back to natural water.

Basssiiie:
Hey all,

This change fixes issue #11912 where the palette does not reload when selected or when the Object Selection window is closed. For better user experience I chose to make it update instantly when selected, as opposed to only update after the window is closed (which is the case for all other selected objects).

When no palette is selected, it will load the default cyan water palette.

A video of the new fix in action:
https://user-images.githubusercontent.com/20048660/104855359-a35a0f00-590c-11eb-9cff-96bea10fd2b4.mp4

I'd love to hear any feedback on the code and whether this fix suits the project. 😃

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -19,6 +19,7 @@
 - Feature: [#13614] Add terrain surfaces from RollerCoaster Tycoon 1.
 - Feature: [#13675] [Plugin] Add context.setInterval and context.setTimeout.
 - Change: [#13346] [Plugin] Renamed FootpathScenery to FootpathAddition, fix typos.
+- Fix: [#4605, #11912] Water palettes are not updated properly when selected in Object Selection.
 - Fix: [#9631, #10716] Banners drawing glitches when there are more than 32 on the screen at once.
 - Fix: [#12895] Mechanics are called to repair rides that have already been fixed.
 - Fix: [#13102] Underflow on height chart (Ride measurements).
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -21,6 +21,7 @@
 #include <openrct2/audio/audio.h>
 #include <openrct2/config/Config.h>
 #include <openrct2/core/String.hpp>
+#include <openrct2/drawing/Drawing.h>
 #include <openrct2/localisation/Localisation.h>
 #include <openrct2/object/ObjectList.h>
 #include <openrct2/object/ObjectManager.h>
@@ -1327,6 +1328,11 @@ static void editor_load_selected_objects()
                 }
             }
         }
+    }
+    if (_numSelectedObjectsForType[EnumValue(ObjectType::Water)] == 0)
+    {
+        // Reloads the default cyan water palette if no palette was selected.
+        load_palette();
     }
 }
 
--- a/src/openrct2/EditorObjectSelectionSession.cpp
+++ b/src/openrct2/EditorObjectSelectionSession.cpp
@@ -13,6 +13,7 @@
 #include "Editor.h"
 #include "Game.h"
 #include "OpenRCT2.h"
+#include "drawing/Drawing.h"
 #include "localisation/Localisation.h"
 #include "management/Research.h"
 #include "object/DefaultObjects.h"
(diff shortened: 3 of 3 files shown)


Gymnasiast:
Might possibly also fix https://github.com/OpenRCT2/OpenRCT2/issues/4605.

Basssiiie:
@Gymnasiast Most likely yes, that glitching happens when the palette has been replaced but `load_palette` has not been called yet (which pushes the palette to the renderer if I'm not mistaken). For the sparkles it already uses the new palette but the rest of the game still uses the old one. This pull request should fix that. :)

Gymnasiast:
Very good, could you amend the changelog with both issues, then?

Basssiiie:
Changelog updated! 👍

IntelOrca:
Could we go further and make it change the palette on highlight?

Umdlye:
A lot of custom palettes also change non-water colors including UI, that would make hovering over a diverse list of palettes an interesting (flashy) experience. If the hover preview would only apply the water colors for the preview image that would be cool though.

IntelOrca:
@Umdlye you could, but I don't think it would be that bad would it? I remember firefox and possibly other browsers allowing you to just hover over themes for it to preview them.

ocalhoun6:
The preview issue seems like something that should be left to a different PR to solve. Ideally, it would effect _only_ the preview image and nothing else. And maybe the preview image should have more than just water in it, to fully show what custom palettes are doing?

Anyway, it seems to me that improvements like that could be made into their own thing instead of trying to tack them onto this, which is mostly a bugfix.

IntelOrca:
> Ideally, it would effect only the preview image and nothing else.

That is not practical right now. Palette effects every pixel on the screen.

ocalhoun6:
> That is not practical right now. Palette effects every pixel on the screen.

Exactly ... which is why it's kind of a big issue on its own.

Basssiiie:
I'm not sure if updating on hover is a good idea. It would be fine if you only have base game templates, but with custom palettes there may be a risk of [epilepsy issues](http://gameaccessibilityguidelines.com/avoid-flickering-images-and-repetitive-patterns/), especially if you (accidentally) move the mouse from top to bottom quickly. Especially since palettes can come from downloaded savegames or multiplayer servers as well, so it's hard to keep track which palettes you have installed.

I checked Firefox as well and they currently use a "change theme on select" instead of hover as well.

For the preview image; for the base game templates water would be good enough, but it's kind of pointless for most custom palettes since they often change water not as drastically as the base game templates. One idea I can throw in here is to maybe show the color palette like this?
afbeelding
But I think the preview image might be something for another PR. :)

Gymnasiast:
I think risk of seizures is a good reason not to implement change-on-hover.

Perhaps we should just make it possible to add screenshots in JSON water objects, so the palette can be shown off.

IntelOrca:
> Perhaps we should just make it possible to add screenshots in JSON water objects, so the palette can be shown off.

This doesn't get round the issue of a fixed palette for the entire game screen.

Personally I would preview on hover, but I seem to be in the minority.

distribution/changelog.txt
```diff
@@ -19,6 +19,7 @@
 - Feature: [#13614] Add terrain surfaces from RollerCoaster Tycoon 1.
 - Feature: [#13675] [Plugin] Add context.setInterval and context.setTimeout.
 - Change: [#13346] [Plugin] Renamed FootpathScenery to FootpathAddition, fix typos.
+- Fix: [#4605, #11912] Water palettes not updating properly when selected in Object Selection.
```
Gymnasiast, line 22: Nitpic, but we don't use the present continuous.

```suggestion
- Fix: [#4605, #11912] Water palettes are not updated properly when selected in Object Selection.
```

On the subject of nitpicking: could you omit the dot at the end of the commit message? (But keep the one in the changelog!)

Basssiiie: Nitpicks fixed. 😉

Gymnasiast (approved):

Issue #13845, opened 2021-01-16
Separating Booster and SpinningControlToggle OpenRCT2-side


spacek531:
Right now, boosters and spinning control toggles exist as a single piece that has different behavior based on ride type. With the recent change to make TrackElement.trackType a 16 bit integer, it is possible to start adding new pieces OpenRCT2-side.

A flag can be added to the RTD to import spinning control toggle pieces and booster pieces as separate trackTypes and export them as the same trackType. This allows the myriad of checks made by OpenRCT2 to separate the behaviors of the same piece to be simplified.

The only potential issue is when using cheats users may assume they can use spinning control toggles and boosters on the same ride since they will no longer be the same piece with different behavior (after importing from SV6), although I have never heard of anyone asking for both to be present on one ride.

Gymnasiast:
This is something that I planned, but it takes a bit more work. You need to make sure that a track_type_t is used throughout. Currently there are still lots of places that use uint8_t.

spacek531:
I started a branch off #13825 to see what it's like and a bit more work is a bit of an understatement, I think.

spacek531:
I just discovered the vehicle track_type/track_direction union. I tried to separate this union but failed miserably. Until it's separated or the value is changed from 16 bit to 32 bit, track types will be limited to 14 bits (for a total of 16384 unique track types)

spacek531:
The PR was forked from another branch with an open PR, which was a mistake, and grew considerably in lines changed due to a lot of formatting changes which were unnecessary. I'm going to start over and force push over the branch once I'm satisfied with the new changes.

spacek531:
Should be obvious - this PR is in reference to this issue https://github.com/OpenRCT2/OpenRCT2/pull/13857

PR #13857, opened 2021-01-18
Separate booster track elem type

spacek531:
Extending the limits on track pieces OpenRCT2-side in preparation for the new save format. Currently the only piece to change is spinning control toggle, which becomes track piece number 256

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -269,7 +269,7 @@ static constexpr const rct_string_id RideConfigurationStringIds[] = {
     0,                                      // 97
     0,                                      // 98
     STR_BRAKES,                             // 99
-    STR_SPINNING_CONTROL_TOGGLE_TRACK,      // 100
+    STR_BOOSTER,                            // 100
     0,                                      // 101
     STR_HELIX_UP_LARGE,                     // 102
     STR_HELIX_UP_LARGE,                     // 103
@@ -424,7 +424,8 @@ static constexpr const rct_string_id RideConfigurationStringIds[] = {
     0,                                      // 252
     STR_QUARTER_LOOP,                       // 253
     STR_QUARTER_LOOP,                       // 254
-    STR_QUARTER_LOOP                        // 255
+    STR_QUARTER_LOOP,                       // 255
+    STR_SPINNING_CONTROL_TOGGLE_TRACK,      // 256
 };
 // clang-format on
 
@@ -440,7 +441,6 @@ static money32 _trackPlaceCost;
 static bool _autoOpeningShop;
 static bool _autoRotatingShop;
 static uint8_t _currentlyShowingBrakeOrBoosterSpeed;
-static bool _boosterTrackSelected;
 
 static uint32_t _currentDisabledSpecialTrackPieces;
 
@@ -2249,12 +2249,6 @@ static void window_ride_construction_invalidate(rct_window* w)
         {
             stringId = STR_LOG_BUMPS;
         }
-        else if (
-            stringId == STR_SPINNING_CONTROL_TOGGLE_TRACK && ride->type != RIDE_TYPE_SPINNING_WILD_MOUSE
-            && ride->type != RIDE_TYPE_STEEL_WILD_MOUSE)
-        {
-            stringId = STR_BOOSTER;
-        }
     }
     auto ft = Formatter::Common();
     ft.Add<uint16_t>(stringId);
@@ -2262,7 +2256,8 @@ static void window_ride_construction_invalidate(rct_window* w)
     if (_currentlyShowingBrakeOrBoosterSpeed)
     {
         uint16_t brakeSpeed2 = ((_currentBrakeSpeed2 * 9) >> 2) & 0xFFFF;
-        if (_boosterTrackSelected)
+        if (_selectedTrackType == TrackElemType::Booster
+            || _currentTrackCurve == (RideConstructionSpecialPieceSelected | TrackElemType::Booster))
         {
             brakeSpeed2 = get_booster_speed(ride->type, brakeSpeed2);
         }
@@ -2492,7 +2487,7 @@ void window_ride_construction_update_active_elements_impl()
 
     window_ride_construction_update_map_selection();
 
-    _selectedTrackType = 255;
+    _selectedTrackType = TrackElemType::None;
     if (_rideConstructionState == RIDE_CONSTRUCTION_STATE_SELECTED)
     {
         if (sub_6C683D(
@@ -3062,11 +3057,10 @@ static void window_ride_construction_update_widgets(rct_window* w)
 
     bool brakesSelected = _selectedTrackType == TrackElemType::Brakes
         || _currentTrackCurve == (RideConstructionSpecialPieceSelected | TrackElemType::Brakes);
-    _boosterTrackSelected = TrackTypeIsBooster(ride->type, _selectedTrackType)
-        || (ride->type != RIDE_TYPE_SPINNING_WILD_MOUSE && ride->type != RIDE_TYPE_STEEL_WILD_MOUSE
-            && _currentTrackCurve == (RideConstructionSpecialPieceSelected | TrackElemType::Booster));
+    bool boosterTrackSelected = _selectedTrackType == TrackElemType::Booster
+        || _currentTrackCurve == (RideConstructionSpecialPieceSelected | TrackElemType::Booster);
 
-    if (!brakesSelected && !_boosterTrackSelected)
+    if (!brakesSelected && !boosterTrackSelected)
     {
         if (is_track_enabled(TRACK_FLAT_ROLL_BANKING))
         {
@@ -3350,12 +3344,6 @@ static void window_ride_construction_show_special_track_dropdown(rct_window* w, 
             auto ride = get_ride(_currentRideIndex);
             if (ride != nullptr && (ride->type == RIDE_TYPE_MONSTER_TRUCKS || ride->type == RIDE_TYPE_CAR_RIDE))
                 trackPieceStringId = STR_LOG_BUMPS;
-        }
-        if (trackPieceStringId == STR_SPINNING_CONTROL_TOGGLE_TRACK)
-        {
-            auto ride = get_ride(_currentRideIndex);
-            if (ride != nullptr && ride->type != RIDE_TYPE_SPINNING_WILD_MOUSE && ride->type != RIDE_TYPE_STEEL_WILD_MOUSE)
-                trackPieceStringId = STR_BOOSTER;
         }
         gDropdownItemsFormat[i] = trackPieceStringId;
         if ((trackPiece | RideConstructionSpecialPieceSelected) == _currentTrackCurve)
@@ -3842,7 +3830,7 @@ void ride_construction_tooldown_construct(const ScreenCoordsXY& screenCoords)
                 || z < 0)
             {
                 int32_t saveTrackDirection = _currentTrackPieceDirection;
-                int32_t saveCurrentTrackCurve = _currentTrackCurve;
+                auto saveCurrentTrackCurve = _currentTrackCurve;
                 int32_t savePreviousTrackSlopeEnd = _previousTrackSlopeEnd;
                 int32_t saveCurrentTrackSlopeEnd = _currentTrackSlopeEnd;
                 int32_t savePreviousTrackBankEnd = _previousTrackBankEnd;
--- a/src/openrct2/rct12/RCT12.cpp
+++ b/src/openrct2/rct12/RCT12.cpp
@@ -752,7 +752,7 @@ void RCT12PathElement::SetIsBlockedByVehicle(bool isBlocked)
     }
 }
 
-bool RCT12TrackTypeIsBooster(uint8_t rideType, uint16_t trackType)
+bool RCT2TrackTypeIsBooster(uint8_t rideType, uint16_t trackType)
 {
     // Boosters share their ID with the Spinning Control track.
     return rideType != RIDE_TYPE_SPINNING_WILD_MOUSE && rideType != RIDE_TYPE_STEEL_WILD_MOUSE
--- a/src/openrct2/rct12/RCT12.cpp
+++ b/src/openrct2/rct12/RCT12.cpp
@@ -759,7 +759,6 @@ bool RCT12TrackTypeIsBooster(uint8_t rideType, uint16_t trackType)
         && trackType == TrackElemType::Booster;
 }
 
-
 void RCT12TrackElement::SetTrackType(uint8_t newType)
 {
     trackType = newType;
--- a/src/openrct2/rct12/RCT12.cpp
+++ b/src/openrct2/rct12/RCT12.cpp
@@ -752,13 +752,6 @@ void RCT12PathElement::SetIsBlockedByVehicle(bool isBlocked)
     }
 }
 
-bool RCT2TrackTypeIsBooster(uint8_t rideType, uint16_t trackType)
-{
-    // Boosters share their ID with the Spinning Control track.
-    return rideType != RIDE_TYPE_SPINNING_WILD_MOUSE && rideType != RIDE_TYPE_STEEL_WILD_MOUSE
-        && trackType == TrackElemType::Booster;
-}
-
 void RCT12TrackElement::SetTrackType(uint8_t newType)
 {
     trackType = newType;
--- a/src/openrct2/rct12/RCT12.cpp
+++ b/src/openrct2/rct12/RCT12.cpp
@@ -752,6 +752,14 @@ void RCT12PathElement::SetIsBlockedByVehicle(bool isBlocked)
     }
 }
 
+bool RCT12TrackTypeIsBooster(uint8_t rideType, uint16_t trackType)
+{
+    // Boosters share their ID with the Spinning Control track.
+    return rideType != RIDE_TYPE_SPINNING_WILD_MOUSE && rideType != RIDE_TYPE_STEEL_WILD_MOUSE
+        && trackType == TrackElemType::Booster;
+}
+
+
 void RCT12TrackElement::SetTrackType(uint8_t newType)
 {
     trackType = newType;
--- a/src/openrct2/rct2/RCT2.h
+++ b/src/openrct2/rct2/RCT2.h
@@ -571,6 +571,26 @@ struct RCT2SpriteVehicle : RCT12SpriteBase
     uint8_t colours_extended;     // 0xD7
     uint8_t seat_rotation;        // 0xD8
     uint8_t target_seat_rotation; // 0xD9
+    uint16_t GetTrackType() const
+    {
+        return track_type >> 2;
+    }
+    uint8_t GetTrackDirection() const
+    {
+        return track_direction & 3;
+    }
+    void SetTrackType(uint16_t trackType)
+    {
+        // set the upper 14 bits to 0
+        track_type &= 3;
+        track_type |= trackType << 2;
+    }
+    void SetTrackDirection(uint8_t trackDirection)
+    {
+        // set the lower 2 bits only
+        track_direction &= ~3;
+        track_direction |= trackDirection & 3;
+    }
 };
 assert_struct_size(RCT2SpriteVehicle, 0xDA);
 
--- a/src/openrct2/rct2/RCT2.h
+++ b/src/openrct2/rct2/RCT2.h
@@ -571,6 +571,7 @@ struct RCT2SpriteVehicle : RCT12SpriteBase
     uint8_t colours_extended;     // 0xD7
     uint8_t seat_rotation;        // 0xD8
     uint8_t target_seat_rotation; // 0xD9
+
     uint16_t GetTrackType() const
     {
         return track_type >> 2;
--- a/src/openrct2/rct2/RCT2.h
+++ b/src/openrct2/rct2/RCT2.h
@@ -800,6 +800,7 @@ assert_struct_size(RCT2RideRatingCalculationData, 76);
 
 std::vector<uint8_t> DecryptSea(const fs::path& path);
 ObjectEntryIndex RCT2RideTypeToOpenRCT2RideType(uint8_t rct2RideType, const rct_ride_entry* rideEntry);
+bool RCT2TrackTypeIsBooster(uint8_t rideType, uint16_t trackType);
 bool RCT2RideTypeNeedsConversion(uint8_t rct2RideType);
 uint8_t OpenRCT2RideTypeToRCT2RideType(ObjectEntryIndex openrct2Type);
 
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -1118,7 +1118,8 @@ void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
         if (trackType == TrackElemType::RotationControlToggle)
             trackType = TrackElemType::RotationControlToggleAlias;
         // Track direction and type are in the same field
-        dst->track_type = (trackType << 2) | (src->track_direction & 3);
+        dst->SetTrackType(trackType);
+        dst->SetTrackDirection(src->GetTrackDirection());
     }
     dst->track_x = src->TrackLocation.x;
     dst->track_y = src->TrackLocation.y;
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -1113,9 +1113,12 @@ void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
     }
     else
     {
+        // export SpinningControlToggle (256) as SpinningControlToggleAlias (100) for backwards compatability with OpenRCT2
+        auto trackType = src->GetTrackType();
+        if (trackType == TrackElemType::RotationControlToggle)
+            trackType = TrackElemType::RotationControlToggleAlias;
         // Track direction and type are in the same field
-        dst->track_direction = src->track_direction;
-        // dst->track_type = src->track_type;
+        dst->track_type = (trackType << 2) | (src->track_direction & 3);
     }
     dst->track_x = src->TrackLocation.x;
     dst->track_y = src->TrackLocation.y;
@@ -1573,7 +1576,11 @@ void S6Exporter::ExportTileElement(RCT12TileElement* dst, TileElement* src)
             auto dst2 = dst->AsTrack();
             auto src2 = src->AsTrack();
 
-            dst2->SetTrackType(src2->GetTrackType());
+            auto trackType = src2->GetTrackType();
+            if (trackType == TrackElemType::RotationControlToggle)
+                trackType = TrackElemType::RotationControlToggleAlias;
+            // SV6 track type is uint8_t; we have aliased Rotation Control (256) to Booster (100) to allow this
+            dst2->SetTrackType(static_cast<uint8_t>(trackType));
             dst2->SetSequenceIndex(src2->GetSequenceIndex());
             dst2->SetRideIndex(src2->GetRideIndex());
             dst2->SetColourScheme(src2->GetColourScheme());
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -1137,7 +1137,12 @@ public:
                 auto dst2 = dst->AsTrack();
                 auto src2 = src->AsTrack();
 
-                dst2->SetTrackType(src2->GetTrackType());
+                auto rideType = _s6.rides[src2->GetRideIndex()].type;
+                track_type_t trackType = static_cast<track_type_t>(src2->GetTrackType());
+
+                if (trackType == TrackElemType::RotationControlToggleAlias && !TrackTypeIsBooster(rideType, trackType))
+                    trackType = TrackElemType::RotationControlToggle;
+                dst2->SetTrackType(trackType);
                 dst2->SetSequenceIndex(src2->GetSequenceIndex());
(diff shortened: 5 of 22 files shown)


spacek531:
This PR will solve https://github.com/OpenRCT2/OpenRCT2/issues/13845

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -463,11 +464,11 @@ static void window_ride_construction_select_map_tiles(
     Ride* ride, int32_t trackType, int32_t trackDirection, const CoordsXY& tileCoords);
 static void window_ride_construction_show_special_track_dropdown(rct_window* w, rct_widget* widget);
 static void ride_selected_track_set_seat_rotation(int32_t seatRotation);
-static void loc_6C7502(int32_t al);
+static void ride_construction_set_slope_steepness(int32_t al);
```
Gymnasiast, line 467: This is not related to this PR, so could you make this a PR of its own?

data/language/en-GB.txt
```diff
@@ -3663,6 +3663,9 @@ STR_6406    :GOG RollerCoaster Tycoon 2 Installer
 STR_6407    :This may take a few minutes.
 STR_6408    :Please install “innoextract” to extract GOG Installer, then restart OpenRCT2.
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
+STR_6410    :Speed Controller
```
Gymnasiast, line 3666: Please remove all "Speed controller" bits from this PR. It makes it harder to review. A PR should be about one thing ideally,

spacek531:
bpb1000ticks fails because srand changed between versions, not sure what the other tests failed for.

spacek531:
notes for a later branch:

9cc72fff706b788f4ea3c633ccc62d0933a6d109

src/openrct2/peep/Staff.cpp
```diff
@@ -2164,7 +2164,7 @@ bool Staff::UpdateFixingMoveToBrokenDownVehicle(bool firstRun, const Ride* ride)
                 break;
             }
 
-            uint8_t trackType = vehicle->GetTrackType();
+            uint16_t trackType = vehicle->GetTrackType();
```
Gymnasiast, line 2167: Use `track_type_t`. Or even better: use `auto`.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1411,7 +1411,7 @@ class S6Importer final : public IParkImporter
             TileElement* tileElement2 = map_get_track_element_at_of_type_seq(dst->TrackLocation, TrackElemType::Booster, 0);
 
             if (tileElement2 != nullptr)
-                dst->track_type = (TrackElemType::Booster << 2) | (src->track_type & 3);
```
Gymnasiast: ?

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1039,10 +1039,9 @@ void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
     else
     {
         uint16_t trackType = src->GetTrackType();
-        int8_t trackDirection = src->GetTrackDirection();
         if (trackType == TrackElemType::Booster)
             trackType = TrackElemType::BoosterAlias;
-        dst->track_type = (trackType << 2) | (trackDirection & 3);
+        dst->track_type = (trackType << 2) | (src->track_direction & 3);
```
Gymnasiast, line 1044: Why?

src/openrct2/ride/Vehicle.h
```diff
@@ -343,20 +343,6 @@ struct Vehicle : SpriteBase
...
-    }
-    int8_t GetTrackDirection() const
-    {
-        return track_direction & 3;
-    }
-    void SetTrackDirection(int8_t direction)
-    {
-        track_direction &= ~(uint16_t)3;
-        track_direction |= direction & 3;
-    }
```
Gymnasiast: Why did you remove this?

spacek531:
Maintaining this history has become untenable in light of the random seed being affected. I am going to start from develop and add changes in a way that is hopefully easier to follow and debug, then force-push onto this branch.

spacek531:
Force push is complete, and all checks pass which is nice.

This branch touches some of the same code as https://github.com/OpenRCT2/OpenRCT2/pull/13825, so whichever is merged first, the other will require some changes. (this branch will have to change the other branch's booster name cases and `speedControlSelected` evaluation)

src/openrct2/ride/Ride.h
```diff
@@ -1149,6 +1149,13 @@ int32_t ride_get_random_colour_preset_index(uint8_t ride_type);
 money32 ride_get_common_price(Ride* forRide);
 RideNaming get_ride_naming(const uint8_t rideType, rct_ride_entry* rideEntry);
 
+money32 place_provisional_track_piece(
```
duncanspumpkin, line 1152: Why are you moving this code.

src/openrct2/ride/Ride.h
```diff
@@ -1149,6 +1149,13 @@ int32_t ride_get_random_colour_preset_index(uint8_t ride_type);
...
+    ride_id_t rideIndex, int32_t trackType, int32_t trackDirection, int32_t liftHillAndAlternativeState,
+    const CoordsXYZ& trackPos);
+bool window_ride_construction_update_state(
```
duncanspumpkin, line 1155: This function should not be moved here.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -467,7 +468,7 @@ static void loc_6C7502(int32_t al);
 static void ride_construction_set_brakes_speed(int32_t brakesSpeed);
 static void ride_construction_tooldown_entrance_exit(const ScreenCoordsXY& screenCoords);
 
-static uint8_t _currentPossibleRideConfigurations[32];
+static track_type_t _currentPossibleRideConfigurations[32];
```
duncanspumpkin, line 471: These type changes don't need to be part of this PR they can be standalone. Do them in a separate PR. Make this purely for splitting up the booster.

src/openrct2/ride/Ride.cpp
```diff
@@ -1304,6 +1313,480 @@ std::optional<CoordsXYZ> sub_6C683D(
     return retCoordsXYZ;
 }
 
+/**
+ *
+ *  rct2: 0x006CA162
+ */
+money32 place_provisional_track_piece(
```
duncanspumpkin, line 1320: These functions do not belong here. They are used only by the ui project and therefore if they are going to move they should move to the ui project. Its hard to see what you have changed as you have moved and I think modified after the move. Just do one thing.

src/openrct2-ui/windows/RideConstruction.cpp
Gymnasiast: They need to be done before the booster _can_ be split from the spinning control. But yeah, should still be a separate PR.

spacek531:
Contemplating whether or not it would be a good idea to leave booster as track elem 100 and change Spinning Control Toggle to track elem 256. Booster is the more common one of the two, and will have the effect of simplifying SV4 importing.

spacek531:
I've gone ahead and flipped Booster to stay at 100 and move RotationControlToggle to 256. I did this because it simplified this branch somewhat - S4/T4 import require no modifications whatsoever, and S6Import/S6Export, T6Import/T6Export require no more support than the other way around (since Booster behavior was already the default, it allowed me to simplify things from the reverse).

src/openrct2/rct12/RCT12.cpp
```diff
@@ -199,7 +199,7 @@ uint8_t RCT12PathElement::GetRCT1SupportType() const
     return (flags & 0b01100000) >> 5;
 }
 
-uint8_t RCT12TrackElement::GetTrackType() const
+uint16_t RCT12TrackElement::GetTrackType() const
```
duncanspumpkin, line 202: this is still returning a `uint8_t` you're function that up converts it to a `uint16_t` is in the import code. You don't need to make this change.

src/openrct2/rct2/S6Exporter.cpp
duncanspumpkin: This is correct. vehicle->track_type contains the track_direction as well. It might be an idea to keep the comment:
```
 // Track direction and type are in the same field
```

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1498,7 +1500,10 @@ void S6Exporter::ExportTileElement(RCT12TileElement* dst, TileElement* src)
             auto dst2 = dst->AsTrack();
             auto src2 = src->AsTrack();
 
-            dst2->SetTrackType(src2->GetTrackType());
+            uint16_t trackType = src2->GetTrackType();
+            if (trackType == TrackElemType::RotationControlToggle)
+                trackType = TrackElemType::RotationControlToggleAlias;
+            dst2->SetTrackType(trackType);
```
duncanspumpkin, line 1506: how is this not causing a warning. You need to convert to a uint8_t here. I would make it explicit add:
```
// SV6 track type is uint8_t we have aliased rotation control to booster to allow this
dst2->SetTrackType(static_cast<uint8_t>(trackType));
```
Also can you make it:
```
auto trackType = src2->GetTrackType();
```
Then when we make that function return a `track_type_t` no changes will be required.

src/openrct2/rct2/T6Exporter.cpp
```diff
@@ -99,7 +99,10 @@ bool T6Exporter::SaveTrack(OpenRCT2::IStream* stream)
     {
         for (const auto& trackElement : _trackDesign->track_elements)
         {
-            tempStream.WriteValue<uint8_t>(trackElement.type);
+            int8_t trackType = trackElement.type;
+            if (trackElement.type == TrackElemType::RotationControlToggle)
```
duncanspumpkin, line 103: There is already a alias inside of TD6 format for vertical elements. We should be doing the conversion code for both all in one location.

duncanspumpkin: See line 214 of TrackDesign.cpp

src/openrct2/rct2/T6Importer.cpp
```diff
@@ -169,7 +169,12 @@ class TD6Importer final : public ITrackImporter
                 _stream.SetPosition(_stream.GetPosition() - 1);
                 _stream.Read(&t6TrackElement, sizeof(rct_td46_track_element));
                 TrackDesignTrackElement trackElement{};
-                trackElement.type = t6TrackElement.type;
+
+                track_type_t trackType = t6TrackElement.type;
+                if (trackType == TrackElemType::RotationControlToggleAlias && !TrackTypeIsBooster(td->type, trackType))
```
duncanspumpkin, line 174: See also conversion at L1534 of TrackDesign.cpp. That conversion should move into this file.

src/openrct2/rct2/T6Exporter.cpp
```diff
@@ -99,7 +99,10 @@ bool T6Exporter::SaveTrack(OpenRCT2::IStream* stream)
     {
         for (const auto& trackElement : _trackDesign->track_elements)
         {
-            tempStream.WriteValue<uint8_t>(trackElement.type);
+            int8_t trackType = trackElement.type;
```
duncanspumpkin, line 102: `uint8_t` this is doing a conversion be explicit. `static_cast<uint8_t>(trackElement.type)`

src/openrct2/windows/_legacy.cpp
```diff
@@ -142,8 +142,8 @@ static std::tuple<bool, uint8_t> window_ride_construction_update_state_get_track
         endBank = _previousTrackBankEnd;
     }
 
-    uint16_t curve = _currentTrackCurve;
-    if (curve == 0xFFFF)
+    int32_t curve = _currentTrackCurve;
```
duncanspumpkin, line 145: auto

src/openrct2/ride/Ride.cpp
```diff
@@ -92,7 +92,7 @@ money32 _currentTrackPrice;
 uint16_t _numCurrentPossibleRideConfigurations;
 uint16_t _numCurrentPossibleSpecialTrackPieces;
 
-uint16_t _currentTrackCurve;
+int32_t _currentTrackCurve;
```
duncanspumpkin, line 95: `uint32_t`

src/openrct2/ride/Ride.h
```diff
@@ -1070,7 +1070,7 @@ extern money32 _currentTrackPrice;
 extern uint16_t _numCurrentPossibleRideConfigurations;
 extern uint16_t _numCurrentPossibleSpecialTrackPieces;
 
-extern uint16_t _currentTrackCurve;
+extern int32_t _currentTrackCurve;
```
duncanspumpkin, line 1073: `uint32_t`

src/openrct2/windows/_legacy.cpp
duncanspumpkin: These changes to `_currentTrackCurve` do not look correct. There is a lot of special code related to `_currentTrackCurve` and this needs some very very careful work to bring it up to `track_type_t`

spacek531:
From what I can tell, `_currentTrackCurve` holds data for two purposes:
- stores the direction and flat curvature as `0` through `8`
- stores the current special track piece as `TrackElemType bitwise-and with a special track type flag '0x1000` (` 0x100` formerly)

duncanspumpkin:
Okay then that sounds fine. I'm okay with this code as is.

duncanspumpkin:
Could you possibly look to squashing this down to a singular commit and then rebaseing. (Do it in that order otherwise you will have a bunch of rebase issues).

src/openrct2/ride/Vehicle.cpp
```diff
@@ -746,7 +746,7 @@ static bool vehicle_move_info_valid(VehicleTrackSubposition trackSubposition, in
     switch (trackSubposition)
     {
         case VehicleTrackSubposition::Default:
-            size = 1024;
+            size = 1028;
```
Gymnasiast, line 749: TrackElemType::Count * 4.

Don't hardcode the number of track pieces.

src/openrct2/rct2/S6Exporter.cpp
Gymnasiast: I'd much rather have a function for this.

src/openrct2/ride/TrackData.cpp
```diff
@@ -3813,6 +3822,7 @@ const uint8_t TrackPieceLengths[TrackElemType::Count] = {
     80,     //
     80,     //
     80,     //
+    32,     //
```
Gymnasiast, line 3825: Also annotate the TrackElemType here.

src/openrct2/ride/TrackData.cpp
```diff
@@ -5255,6 +5269,10 @@ const money32 FlatRideTrackPricing[] = {
     151552,
     151552,
     151552,
+    360448, // TrackElemType::MultiDimUp90ToInvertedFlatQuarterLoop
+    360448, // TrackElemType::MultiDimFlatToDown90QuarterLoop
+    360448, //
+    77824,  // TrackElemType::RotationControlToggle
```
Gymnasiast, line 5275: Set the size of this array to `FlatRideTrackPricing`, so we'll know if it contains too few entries. Same for the other ones.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -213,10 +213,6 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
         TrackDesignTrackElement track{};
         track.type = trackElement.element->AsTrack()->GetTrackType();
         // TODO move to RCT2 limit
```
Gymnasiast, line 215: You moved this, so you can remove this comment as well.

src/openrct2/windows/_legacy.cpp
```diff
@@ -308,7 +308,7 @@ bool window_ride_construction_update_state(
         auto alternativeType = AlternativeTrackTypes[trackType];
         if (alternativeType != -1 && availablePieces & (1ULL << trackType))
         {
-            trackType = static_cast<uint8_t>(alternativeType);
+            trackType = static_cast<track_type_t>(alternativeType);
```
Gymnasiast, line 311: Perhaps refactor this so that there is no need for a cast.

src/openrct2/ride/Track.h
```diff
@@ -24,13 +24,13 @@ using pitch_type_t = uint8_t;
...
 {
-    uint8_t type;
+    track_type_t type;
     pitch_type_t vangle_end;
     pitch_type_t vangle_start;
     roll_type_t bank_end;
     roll_type_t bank_start;
     int8_t preview_z_offset;
-    uint8_t pad[2] = {};
+    uint8_t pad[1] = {};
```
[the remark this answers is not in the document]
duncanspumpkin: It should be fine removing the packing and such. It's possible test paint night complain.

spacek531:
note: cannot refactor `src/openrct2/ride/Track.cpp` `TrackTypeIsBooster()` any further: S6/T6Import/Export requires this check.

spacek531:
For some reason the rotation control toggle is not buildable. Whoops! Will sort it out after I sleep.

spacek531:
Turns out the above is a false alarm, I forgot that OpenRCT2 *didn't* merge Steel Wild Mouse and Spinning Wild Mouse. Rotation Control Toggle works fine.

spacek531:
I'm happy where this branch is at. I've squashed and bumped network version. Is there anything else this branch needs to do?

src/openrct2/windows/_legacy.cpp
```diff
@@ -124,7 +124,7 @@ money32 place_provisional_track_piece(
     }
 }
 
-static std::tuple<bool, uint8_t> window_ride_construction_update_state_get_track_element()
+static std::tuple<bool, track_type_t> window_ride_construction_update_state_get_track_element()
```
ZehMatt, line 127:
```suggestion
static std::optional<track_type_t> window_ride_construction_update_state_get_track_element()
```
Makes the usage of this function a bit more clear.

src/openrct2/windows/_legacy.cpp
```diff
@@ -306,9 +306,9 @@ bool window_ride_construction_update_state(
     {
         auto availablePieces = rtd.CoveredTrackPieces;
         auto alternativeType = AlternativeTrackTypes[trackType];
-        if (alternativeType != -1 && availablePieces & (1ULL << trackType))
+        if (alternativeType != TrackElemType::None && availablePieces & (1ULL << trackType))
```
ZehMatt, line 309:
```suggestion
        if (alternativeType != TrackElemType::None && (availablePieces & (1ULL << trackType)))
```

ZehMatt (changes requested):

src/openrct2/windows/_legacy.cpp
duncanspumpkin: I would suggest we leave that to a different PR. Keep this one purely focused on splitting the booster.

spacek531: I made it return only track_type_t to use TrackElemType::None as the 'false' value.

spacek531: this fails the CI so I dropped the commit

src/openrct2/ride/VehicleSubpositionData.cpp
```diff
@@ -24060,7 +24060,8 @@ static constexpr const rct_vehicle_info_list *TrackVehicleInfoListDefault[] = {
     &TrackVehicleInfo_96C062, &TrackVehicleInfo_96C3C4, &TrackVehicleInfo_96AC16, &TrackVehicleInfo_96AF78, &TrackVehicleInfo_96B2DA, &TrackVehicleInfo_96B63C,
     &TrackVehicleInfo_96C726, &TrackVehicleInfo_96CA88, &TrackVehicleInfo_96CDEA, &TrackVehicleInfo_96D14C, &TrackVehicleInfo_9761F2, &TrackVehicleInfo_9766C5,
     &TrackVehicleInfo_976B98, &TrackVehicleInfo_97706B, &TrackVehicleInfo_97753E, &TrackVehicleInfo_977A11, &TrackVehicleInfo_977EE4, &TrackVehicleInfo_9783B7,
-    &TrackVehicleInfo_97888A, &TrackVehicleInfo_978D5D, &TrackVehicleInfo_979230, &TrackVehicleInfo_979703,
+    &TrackVehicleInfo_97888A, &TrackVehicleInfo_978D5D, &TrackVehicleInfo_979230, &TrackVehicleInfo_979703, &TrackVehicleInfo_8BE57A, &TrackVehicleInfo_8BE69C,
```
Gymnasiast, line 24063: Where did you get the missing data from? And why did you (have to) add four entries?

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1382,6 +1385,17 @@ class S6Importer final : public IParkImporter
...
+        if (dst->track_type && (src->track_type >> 2) == TrackElemType::RotationControlToggleAlias)
+        {
+            // Merging hacks mean the track type that's appropriate for the ride type is not necessarily the track type the ride
+            // is on. It's possible to create unwanted behavior if a user layers spinning control track on top of booster track
+            // but this is unlikely since only two rides have spinning control track - by default they load as booster
+            TileElement* tileElement2 = map_get_track_element_at_of_type_seq(
+                dst->TrackLocation, TrackElemType::RotationControlToggle, 0);
+
+            if (tileElement2 != nullptr)
+                dst->track_type = (TrackElemType::RotationControlToggle << 2) | (src->track_direction & 3);
```
Gymnasiast, line 1397: I'd like you to add getters and setters for track type and direction to both our tile element and to the RCT12 ones, so that we don't have to hardcode this.

src/openrct2/ride/VehicleSubpositionData.cpp
spacek531: This is the subposition data. Each entry represents one angle of vehicle positions of a track piece, in order of TrackElemType then direction. So TrackVehicleInfoListDefault[0] is TrackElemType::Flat subposition data direction 0, TrackVehicleInfoListDefault[1] is TrackElemType::Flat subposition data direction 1. The four entries I added are, in order, TrackElemType::Flat subposition data direction 0, TrackElemType::Flat subposition data direction 1, TrackElemType::Flat subposition data direction 2, TrackElemType::Flat subposition data direction 3.

spacek531: I had to add four entries otherwise track element type 256 has no subposition data

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1137,7 +1137,12 @@ class S6Importer final : public IParkImporter
                 auto dst2 = dst->AsTrack();
                 auto src2 = src->AsTrack();
 
-                dst2->SetTrackType(src2->GetTrackType());
+                auto rideType = _s6.rides[src2->GetRideIndex()].type;
+                track_type_t trackType = static_cast<track_type_t>(src2->GetTrackType());
+
+                if (trackType == TrackElemType::RotationControlToggleAlias && !TrackTypeIsBooster(rideType, trackType))
```
Gymnasiast, line 1143: `TrackTypeIsBooster` is now only used for RCT2 import/export. Please rename it RCT2TrackTypeIsBooster and move it to RCT2.cpp/h.

src/openrct2/rct2/S6Importer.cpp
spacek531: This is vehicle import code?

src/openrct2/ride/VehicleSubpositionData.cpp
Gymnasiast: I see. Did you copy the subposition from an existing piece?

spacek531: yes, I copied the existing TrackElemType::Flat subposition data for directions 0 thru 3. If you search the file for the variable names they appear in order at the very beginning of the array as I explained.

src/openrct2/rct2/S6Importer.cpp
Gymnasiast: Yeah, I didn't say it correctly. But the request still stands (cross out tile element and substitute `Ride` and `RCT2Ride`).

src/openrct2/ride/VehicleSubpositionData.cpp
spacek531: Interestingly enough, the value stored in track direction references this value without any other modifications.

src/openrct2/ride/Vehicle.h
```diff
@@ -221,7 +221,7 @@ struct Vehicle : SpriteBase
     union
     {
         int16_t track_direction; // (0000 0000 0000 0011)
-        int16_t track_type;      // (0000 0011 1111 1100)
+        int16_t track_type;      // (1111 1111 1111 1100)
```
duncanspumpkin, line 224: for a follow on PR we should remove this union make it so everything access's via the getters.

duncanspumpkin (approved):

src/openrct2/rct12/RCT12.cpp
```diff
@@ -752,6 +752,13 @@ void RCT12PathElement::SetIsBlockedByVehicle(bool isBlocked)
     }
 }
 
+bool RCT12TrackTypeIsBooster(uint8_t rideType, uint16_t trackType)
```
Gymnasiast, line 755: `RCT2TrackTypeIsBooster`. Not "RCT12...". RCT1 always has a booster at index 100, so does not need conversion.

src/openrct2/rct2/RCT2.h
```diff
@@ -571,6 +571,26 @@ struct RCT2SpriteVehicle : RCT12SpriteBase
     uint8_t colours_extended;     // 0xD7
     uint8_t seat_rotation;        // 0xD8
     uint8_t target_seat_rotation; // 0xD9
+    uint16_t GetTrackType() const
```
Gymnasiast, line 574: Put a blank line above this.

src/openrct2/rct12/RCT12.cpp
```diff
@@ -752,6 +752,13 @@ void RCT12PathElement::SetIsBlockedByVehicle(bool isBlocked)
     }
 }
 
+bool RCT2TrackTypeIsBooster(uint8_t rideType, uint16_t trackType)
```
Gymnasiast, line 755: Should be in `RCT2.cpp`/h, as I said [here](https://github.com/OpenRCT2/OpenRCT2/pull/13857#discussion_r565632018)

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1382,6 +1386,17 @@ class S6Importer final : public IParkImporter
         }
 
         dst->TrackLocation = { src->track_x, src->track_y, src->track_z };
+        if (dst->track_type && src->GetTrackType() == TrackElemType::RotationControlToggleAlias)
```
Gymnasiast, line 1389: What are you checking `dst->track_type` for?

duncanspumpkin: It's due to the boat code above.

Gymnasiast: Well, that would need a comment at least, and an explicit `!= 0` check.

spacek531: Making small diffs, since it seemed to be an issue earlier. It should belong in the true case of the above `if` statement.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1372,31 +1372,37 @@ class S6Importer final : public IParkImporter
...
+            if (src->GetTrackType() == TrackElemType::RotationControlToggleAlias)
+            {
+                // Merging hacks mean the track type that's appropriate for the ride type is not necessarily the track type the
+                // ride is on. It's possible to create unwanted behavior if a user layers spinning control track on top of
+                // booster track but this is unlikely since only two rides have spinning control track - by default they load as
+                // booster
+                TileElement* tileElement2 = map_get_track_element_at_of_type_seq(
+                    dst->TrackLocation, TrackElemType::RotationControlToggle, 0);
+
+                if (tileElement2 != nullptr)
```
duncanspumpkin, line 1391: You're not accounting for a bunch of conditions now. You want it to fall back to a normal Set(Get()) if this fails. Also you are no longer setting the direction

Gymnasiast (commented):
Code looks good, but I want to test it.

src/openrct2/rct2/S6Importer.cpp
spacek531: I'm not sure what you mean by "a bunch of conditions." There's only 3 possible cases:
- vehicle's boat location is set
- vehicle is on RotationControlToggleAlias
- vehicle is not on RotationControlToggleAlias

Regardless, the assignment at 1374 should be changed to the getter/setter

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1394,7 +1394,6 @@ class S6Importer final : public IParkImporter
             else
```
duncanspumpkin, line 1394: But what happens when line 1391 is nullptr. There is no SetTrackType now for it. Thats what I was trying to say before but didn't account for L1374.

spacek531: I've changed this now, it should no longer have a floating TrackType

Gymnasiast (approved):
Seems to work correctly as well. I take it you tested it with loading/saving parks and track designs?

spacek531:
Yes, I've tested T6 and S6 import and export of rotation control toggle, booster, and the 4D quarter loop.

Gymnasiast:
Great work, and thanks for picking it up! :+1:

PR #13863, opened 2021-01-18
Fix #13629: Refactor RCT1_PEEP_SPRITE_TYPE to use scoped enum

### LINKED ISSUE #13629: Refactor RCT1_PEEP_SPRITE_TYPE to use strong enum

JoelLarson:
This PR is an implementation for #13629

The intent of the modified code is a little confusing though. RCT1::PeepSpriteType is coupled to the index of the array (map) in RCT1::GetPeepSpriteType. I think it would be a good idea to better communicate the intent explicitly when mapping the RCT1 to OpenRCT2.

### DIFF
--- a/src/openrct2/rct1/Tables.cpp
+++ b/src/openrct2/rct1/Tables.cpp
@@ -68,7 +68,7 @@ namespace RCT1
         return map[colour];
     }
 
-    ::PeepSpriteType GetPeepSpriteType(PeepSpriteType rct1SpriteType)
+    ::PeepSpriteType GetPeepSpriteType(::RCT1::PeepSpriteType rct1SpriteType)
     {
         static constexpr const ::PeepSpriteType map[] =
         {
--- a/src/openrct2/rct1/Tables.cpp
+++ b/src/openrct2/rct1/Tables.cpp
@@ -68,52 +68,52 @@ namespace RCT1
         return map[colour];
     }
 
-    PeepSpriteType GetPeepSpriteType(uint8_t rct1SpriteType)
+    ::PeepSpriteType GetPeepSpriteType(PeepSpriteType rct1SpriteType)
     {
-        static constexpr const PeepSpriteType map[] =
+        static constexpr const ::PeepSpriteType map[] =
         {
-            PeepSpriteType::Normal, // 0x00
-            PeepSpriteType::Handyman, // 0x01
-            PeepSpriteType::Mechanic, // 0x02
-            PeepSpriteType::Security, // 0x03
-            PeepSpriteType::EntertainerPanda, // 0x04
-            PeepSpriteType::EntertainerTiger, // 0x05
-            PeepSpriteType::EntertainerElephant, // 0x06
-            PeepSpriteType::EntertainerRoman, // 0x07
-            PeepSpriteType::EntertainerGorilla, // 0x08
-            PeepSpriteType::EntertainerSnowman, // 0x09
-            PeepSpriteType::EntertainerKnight, // 0x0A
-            PeepSpriteType::EntertainerAstronaut, // 0x0B
-            PeepSpriteType::IceCream, // 0x0C
-            PeepSpriteType::Chips, // 0x0D
-            PeepSpriteType::Burger, // 0x0E
-            PeepSpriteType::Drink, // 0x0F
-            PeepSpriteType::Balloon, // 0x10
-            PeepSpriteType::Candyfloss, // 0x11
-            PeepSpriteType::Umbrella, // 0x12
-            PeepSpriteType::Pizza, // 0x13
-            PeepSpriteType::SecurityAlt, // 0x14
-            PeepSpriteType::Popcorn, // 0x15
-            PeepSpriteType::ArmsCrossed, // 0x16
-            PeepSpriteType::HeadDown, // 0x17
-            PeepSpriteType::Nauseous, // 0x18
-            PeepSpriteType::VeryNauseous, // 0x19
-            PeepSpriteType::RequireToilet, // 0x1A
-            PeepSpriteType::Hat, // 0x1B
-            PeepSpriteType::HotDog, // 0x1C
-            PeepSpriteType::Tentacle, // 0x1D
-            PeepSpriteType::ToffeeApple, // 0x1E
-            PeepSpriteType::Doughnut, // 0x1F
-            PeepSpriteType::Coffee, // 0x20
-            PeepSpriteType::Chicken, // 0x21
-            PeepSpriteType::Lemonade, // 0x22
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
It feels like the namespace is getting in the way of making this clear.

JoelLarson:
It sure does. I think it is also redundant with the `GetPeepSpriteType(uint8_t rct1SpriteType)` having an implicit mapping based on index of the enum. If it's ideal to allow the RCT1 related code to be separate to allow other projects to parse them into a data structure, then I'd fully separate them and have the mapper convert one to the other. I don't think that is the case here though and a lot of the RCT1 code could be simplified by merging the table code with the RCT1.h code.

The RCT1 code would just be responsible for mapping the data structure to the OpenRCT code format and would remove the RCT1 namespace.

duncanspumpkin:
Nothing actually uses this structure only the mapping inside of `GetPeepSpriteType`. `rct1SpriteType` is implied to be this type but its a uint8_t as we have to validate that it is within the bounds of the enum type before use. A lot of these enums are for reference more than anything else.

JoelLarson:
I used the scope resolution operator because of the collision within the RCT1 namespace. I think whether it's worth using depends on whether the main project should be in an OpenRCT namespace or stay global. The importer could also be in the OpenRCT (global) scope and reference `RCT1::PeepSpriteType` without this issue.

That aside, I think there is value in keeping the information about how RCT1 data is represented from its domain. I would be willing to work on writing something that unmarshal the save files into a data structure that OpenRCT can use within the importer code.

duncanspumpkin:
Oh right. Yes these types should be within the OpenRCT2 namespace but thats a big job we have not got into. Really as these are peep specific they would likely end up in `OpenRCT2::Peep::SpriteType`.

JoelLarson:
The amount of change in this sucks to look at when whitespace is in. I recommend adding `?w=1` to the end of the URL to see the preview without whitespace differences

JoelLarson:
The overall idea with these changes are to move the Importer into the OpenRCT2::RCT1 namespace and move the RCT1 specific code into its own RCT1 namespace to avoid collisions. After trying this out, I liked how it simplified some concepts and can see it working well for the remaining importer code as I refactor enums to scoped enums.

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -62,3040 +62,3050 @@
...
+                dst->vehicle_colours[0].Body = COLOUR_YELLOW;
+                dst->vehicle_colours[0].Trim = COLOUR_BRIGHT_RED;
+            }
+            else
+            {
+                for (int i = 0; i < RCT1_MAX_TRAINS_PER_RIDE; i++)
                 {
-                    dst->vehicle_colours[i].Ternary = RCT1::GetColour(src->vehicle_colours[i].trim);
+                    // RCT1 had no third colour
+                    auto colourSchemeCopyDescriptor = RCT1::GetColourSchemeCopyDescriptor(src->vehicle_type);
```
JoelLarson, line 1023: Changed to auto to make the line short enough to pass validation

JoelLarson:
I've squashed the commit down with this latest commit.

tupaschoal (approved):
Without the whitespace changes, this looks very ok.

Gymnasiast:
Do we really need to put everything in an `OpenRCT2` namespace? It causes massive diffs and more indentation (which is never a good thing for legibility). I understand namespaces for RCT1, RCT2, etc. but do we need one for OpenRCT2?

duncanspumpkin:
I think it's something we should be doing. You can have non white space diffs.

JoelLarson:
I added the OpenRCT2 namespace because the RCT1 structs collide with the global ones. If whitespace is an issue, we could adopt [Google's coding standard](https://google.github.io/styleguide/cppguide.html#Namespaces) for them which would be:

```
namespace OpenRCT2 {

// code goes here without indent

}
```

src/openrct2/rct1/T4Importer.cpp
```diff
@@ -25,271 +25,275 @@
 /**
  * Class to import RollerCoaster Tycoon 1 track designs (*.TD4).
  */
-class TD4Importer final : public ITrackImporter
+namespace OpenRCT2
```
Gymnasiast, line 28: Shouldn't this be in the namespace of `OpenRCT2::RCT1`, then?

JoelLarson: I could see it being in either namespace. With the file currently being in the rct1 directory, I'll try changing the namespace and see how it changes the rest of the code it imports.

JoelLarson:
@duncanspumpkin I updated the change to just be the enum change. Let me know if I misunderstood or there's any other changes I need to make.

duncanspumpkin:
You'll need to rebase this. If you are not comfortable doing that just say and i can take over the pr.

JoelLarson:
Feel free to take over the PR. I have the parent of the current commit off of `upstream/master` instead of develop which was an accident.

duncanspumpkin (approved):

duncanspumpkin:
Seems there were some more fundamental changes since this was originally written. Will rework later today.

src/openrct2/rct1/Tables.cpp
```diff
@@ -68,52 +68,52 @@ namespace RCT1
         return map[colour];
     }
 
-    PeepSpriteType GetPeepSpriteType(uint8_t rct1SpriteType)
+    ::PeepSpriteType GetPeepSpriteType(PeepSpriteType rct1SpriteType)
```
Broxzier, line 71: Why are all these `::PeepSpriteType`? It's defined in the `RCT1` namespace, which we are in.

duncanspumpkin: This is a conversion function from rct1 to openrct2. As the openrct2 fields are not in a namespace yet you get this horrible stuff

Gymnasiast (approved):

duncanspumpkin:
It might have to come prior to this PR as CI's dont appear to like my usage of `::` :(

src/openrct2/rct1/Tables.h
```diff
@@ -24,7 +24,7 @@ namespace RCT1
     enum class RideType : uint8_t;
 
     colour_t GetColour(colour_t colour);
-    PeepSpriteType GetPeepSpriteType(uint8_t rct1SpriteType);
+    ::PeepSpriteType GetPeepSpriteType(PeepSpriteType rct1SpriteType);
```
ZehMatt, line 27:
```suggestion
    ::PeepSpriteType GetPeepSpriteType(::RCT1::PeepSpriteType rct1SpriteType);
```
Should solve make the CI happy

PR #13865, opened 2021-01-19
Name MapFlashingFlags and split up guest/staff loop

duncanspumpkin:


IntelOrca:
Why does `DrawMapPeepPixel` need a callback to get the colour when the functions that get passed looked like they can be constants passed in in-advance?

duncanspumpkin:
That's a good point. Not sure what I was thinking. That would be much simpler

src/openrct2-ui/windows/Map.cpp
```diff
@@ -34,15 +34,21 @@
 #include <vector>
 
 #define MAP_COLOUR_2(colourA, colourB) (((colourA) << 8) | (colourB))
-#define MAP_COLOUR(colour) MAP_COLOUR_2(colour, colour)
-#define MAP_COLOUR_UNOWNED(colour) (PALETTE_INDEX_10 | ((colour)&0xFF00))
+#define MAP_COLOUR(flashColour) MAP_COLOUR_2(flashColour, flashColour)
+#define MAP_COLOUR_UNOWNED(flashColour) (PALETTE_INDEX_10 | ((flashColour)&0xFF00))
```
ZehMatt, line 38: Those are good candidates to be converted to constexpr functions I think.

duncanspumpkin: urgh didn't mean to rename these

src/openrct2-ui/windows/Map.cpp
```diff
@@ -33,16 +33,31 @@
...
+    return (colourA << 8) | colourB;
+}
+constexpr uint16_t MapColour(uint8_t colour)
+{
+    return MapColour2(colour, colour);
+}
+constexpr uint16_t MapColourUnowned(uint16_t colour)
+{
+    return MapColour2((colour & 0xFF00) >> 8, PALETTE_INDEX_10);
+}
```
ZehMatt, line 47: Missing ```static```

PR #13869, opened 2021-01-19
Allow the main game loop to scale the delta time

ZehMatt:
Which results in this:

https://www.youtube.com/watch?v=dNNmagd1nOs

The final goal is to get rid of the inner loop that drives the game speed and have input/windows update at fixed time, I already have scrolling fixed locally but I want to do smaller changes over time.

### DIFF
--- a/src/openrct2/Context.cpp
+++ b/src/openrct2/Context.cpp
@@ -118,7 +118,8 @@ namespace OpenRCT2
         bool _initialised = false;
         bool _isWindowMinimised = false;
         uint32_t _lastTick = 0;
-        uint32_t _accumulator = 0;
+        float _accumulator = 0.0f;
+        float _timeScale = 1.0f;
         uint32_t _lastUpdateTime = 0;
         bool _variableFrame = false;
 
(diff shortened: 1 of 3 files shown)


janisozaur:
Oh my, floats in the core, no like.

ZehMatt:
> Oh my, floats in the core, no like.

It just accumulates time in the one float and tweens with a float. Nothing of the game will change by this.

janisozaur:
ah, carry on then.

src/openrct2/Context.cpp
```diff
@@ -1220,6 +1221,16 @@ namespace OpenRCT2
         {
             return &_newVersionInfo;
         }
+
+        void SetTimeScale(float newScale) override
+        {
+            _timeScale = std::clamp(newScale, 0.0f, 5.0f);
```
janisozaur, line 1227: Seems the two values are related?
```suggestion
            _timeScale = std::clamp(newScale, 0.0f, GAME_MAX_UPDATES);
```
Also does it make sense to allow 0 scale?

ZehMatt: Good point, I think 0.1 should be small enough.

ZehMatt: GAME_MAX_UPDATES is used for something else, I accidentally changed that.

tupaschoal (approved):
LGTM, though I wonder what the motivation was (testing?)

ZehMatt:
There are a couple of things why I did this.
1. Testing the tweening when opening/closing rides in slow motion.
2. Preparing for future code clean to get rid of the nested loop to speed up the game.
3. Its cool to control the time scale as I wish.

PR #13870, opened 2021-01-19
Split up misc list access to be type specific

duncanspumpkin:
Took the opertunity to use a fold expression to simplify the code

Might break replays.
Note this decreases performance for now.

### DIFF
--- a/src/openrct2/network/NetworkBase.cpp
+++ b/src/openrct2/network/NetworkBase.cpp
@@ -34,7 +34,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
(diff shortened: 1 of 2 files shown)


tupaschoal (approved):
First I had to understand fold expressions (which I was unaware of and if anyone wants to, here's a [link](https://www.youtube.com/watch?v=nhk8pF_SlTk)).

After that, I think it looks fine :D

Broxzier:
Sure looks better, but now it iterates the list for each type if I understand this correctly, right?

duncanspumpkin:
It does for now but I'll be adding extra type specific lists in a couple PRs time. Do note also that there is not a lot of misc entries at any time.

Broxzier:
Just saw your message on Discord, splitting them up into separate lists will make this a lot better indeed. :)

PR #13880, opened 2021-01-20
Split closest peep calculation into Staff and Guest

duncanspumpkin:


src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -697,6 +697,34 @@ static void ViewportInteractionRemoveLargeScenery(TileElement* tileElement, cons
     }
 }
 
+struct PeepDistance
+{
+    Peep* peep = nullptr;
+    int32_t distance = std::numeric_limits<int32_t>::max();
```
Broxzier, line 703: Seems like a good place to use `decltype` instead of having the type there twice:

```suggestion
    int32_t distance = std::numeric_limits<decltype(distance)>::max();
```

src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -697,6 +697,34 @@ static void ViewportInteractionRemoveLargeScenery(TileElement* tileElement, cons
...
+template<typename T>
+PeepDistance GetClosestPeep(const ScreenCoordsXY& viewportCoords, const int32_t maxDistance, PeepDistance goal)
+{
+    for (auto peep : EntityList<T>(EntityListId::Peep))
+    {
+        if (peep->sprite_left == LOCATION_NULL)
+            continue;
+
+        auto distance = abs(((peep->sprite_left + peep->sprite_right) / 2) - viewportCoords.x)
+            + abs(((peep->sprite_top + peep->sprite_bottom) / 2) - viewportCoords.y);
```
ZehMatt, line 715:
```suggestion
    auto peepCenter = ScreenCoordsXY{ (peep->sprite_left + peep->sprite_right) / 2,
                                      (peep->sprite_top + peep->sprite_bottom) / 2 };
    auto offset = peepCenter - viewportCoords;
    auto distance = std::abs(offset.x) + std::abs(offset.y);
```
Is a bit easier to read I think. We could also at some point implement the Distance function into all the Location structs, there is probably more like this scattered around.

PR #13883, opened 2021-01-20
Refactor track data

spacek531:
Making this PR now because apparently there's some PR rules I don't understand about making tiny incremental changes.

So let me know before this PR gets "too much for one PR."

### DIFF
--- a/src/openrct2/libopenrct2.vcxproj
+++ b/src/openrct2/libopenrct2.vcxproj
@@ -381,7 +381,8 @@
     <ClInclude Include="ride\TrackDesign.h" />
     <ClInclude Include="ride\TrackDesignRepository.h" />
     <ClInclude Include="ride\TrackPaint.h" />
-    <ClInclude Include="ride\tracktypes\Flat.h" />
+    <ClInclude Include="ride\tracktypes\FlatTrack.h" />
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
That's a novel way of doing this. Meaning you don't need to switch everything at once. Good work.

spacek531:
Is there a way a test can be written for all this data, to verify it's all correct or at least monitor what's changed? I'm sure I can get at least 80% of the data correct but having a test to verify would make it easier to fix any errors.

duncanspumpkin:
If you hadn't removed it from the track data.cpp you could do a constexpr function that checks it all at compile time.

Gymnasiast:
Since this was made before the combination of flat and regular track element tables and https://github.com/OpenRCT2/OpenRCT2/pull/15261 has already been merged and takes a different approach, there is no way we can merge this. 

I do eventually want TEDs to look like you did [here](https://github.com/OpenRCT2/OpenRCT2/blob/00c59d9b9e0ab05a7c4e163365c4232de8d1b028/src/openrct2/ride/tracktypes/FlatBankedTrack.h), but that will have to wait until all access to the existing tables has been refactored away.

PR #13838, opened 2021-01-15
Block brake improvements

spacek531:
I started on this branch because at the time I didn't know about the old PR. @X123M3-256 told me about the old PRs https://github.com/OpenRCT2/OpenRCT2/pull/7737 and https://github.com/OpenRCT2/OpenRCT2/issues/11649 and that there were some larger technical issues, but he didn't remember what they were and the old PR does not say what they are.

OpenRCT2 side, everything is unified to the block brake flag indicating that brakes are active and the absence meaning they are open. 
SV6 side, the polarity is inverted for brakes, since they normally serialize with the flag as false.
TD6 side, block brake speed is not exported nor imported. It is set to the minimum on loading.
NSF side, a version bump is required and behavior to normalize old save format brake flags.

PR #13893, opened 2021-01-22
update ChecksumIndex when checksum mismatch

spacek531:
This appears to solve the issues outlined in the comments section of https://github.com/OpenRCT2/OpenRCT2/pull/13838

When cherry-picking a7a0441 onto this commit, the desync is only 3kb compared to the 8000kb desync without this commit.

### DIFF
--- a/src/openrct2/ReplayManager.cpp
+++ b/src/openrct2/ReplayManager.cpp
@@ -816,6 +816,7 @@ namespace OpenRCT2
 
                     _faultyChecksumIndex = checksumIndex;
 
+                    _currentReplay->checksumIndex++;
                     return false;
                 }
                 else


spacek531:
Thanks to @joellarson for suggesting this fix

duncanspumpkin:
Yes that looks like an oversight.

ZehMatt:
This doesn't fix the actual issue. The assumption of the snapshot attached to the replay is that the playback must complete, I changed the logic recently that would stop the replay the moment it detects a desync so the snapshot is compared against the wrong tick. I'm going to close this and will see to fix it.

ZehMatt:
Fixed by #13901

PR #13897, opened 2021-01-22
Use track_type_t in more places

Gymnasiast:
This expands the usage of `track_type_t` and also replaces a few hardcoded occurrences of 256 for the number of track pieces.

This should help @spacek531 's refactors.


duncanspumpkin:
Track type t is 16bit? This will need a network version bump for the action changes. Might break one of the replays as well.

spacek531:
Should I rebase my branch onto this? It seems like it's a prerequisite for things to work right generally; I seem to have lucked out with e.g. the wall flags that 256 rolls over to 0 which is a flat track. There's going to be some merge conflicts that I can sort out sooner than later waiting for this branch to get merged.

spacek531:
that answers that

PR #13899, opened 2021-01-22
Refactor guestlist

duncanspumpkin:
Was investigating `SPRITE_FLAGS_PEEP_VISIBLE` and noticed this could be simplified.
Also noticed a bug in the previous calculations that would cause groups to not be fully populated.

### DIFF
--- a/src/openrct2-ui/windows/GuestList.cpp
+++ b/src/openrct2-ui/windows/GuestList.cpp
@@ -110,7 +110,6 @@ private:
     struct GuestGroup
     {
         size_t NumGuests{};
-        uint8_t Index{};
         FilterArguments Arguments;
         uint8_t Faces[58]{};
     };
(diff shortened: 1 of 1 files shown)


src/openrct2-ui/windows/GuestList.cpp
```diff
@@ -809,73 +808,54 @@ class GuestListWindow final : public Window
         return true;
     }
 
+    GuestGroup& FindOrGetGroup(FilterArguments&& arguments)
+    {
+        auto foundGroup = std::find_if(
+            std::begin(_groups), std::end(_groups), [arguments](GuestGroup& group) { return group.Arguments == arguments; });
```
duncanspumpkin, line 814: Pass by ref the arguments

ZehMatt (approved):

src/openrct2-ui/windows/GuestList.cpp
```diff
@@ -809,73 +808,54 @@ class GuestListWindow final : public Window
         return true;
     }
 
+    GuestGroup& FindOrGetGroup(FilterArguments&& arguments)
```
IntelOrca, line 811: `GetOrAddGroup` or `FindOrAddGroup`

PR #13924, opened 2021-01-23
Split free id's into separate vector

duncanspumpkin:
This makes it so that free id's will eventually only need to store memory for the uint16_t index (future pr). There is potential future scope to make this fancier so that it doesn't need to store all of the free id's and just stores ranges but for now this will do.

Order is reversed so that back and pop_back can be used to get the next free element and save some memory shuffling.

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -28,6 +28,7 @@
 
 static rct_sprite _spriteList[MAX_SPRITES];
 static std::array<std::list<uint16_t>, EnumValue(EntityListId::Count)> gEntityLists;
+static std::vector<uint16_t> _freeIdList;
 
 static bool _spriteFlashingList[MAX_SPRITES];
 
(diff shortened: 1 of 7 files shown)


src/openrct2/world/Sprite.cpp
```diff
@@ -28,6 +28,7 @@
 
 static rct_sprite _spriteList[MAX_SPRITES];
 static std::array<std::list<uint16_t>, EnumValue(EntityListId::Count)> gEntityLists;
+static std::vector<uint16_t> _freeIdList;
```
tupaschoal, line 31: I took a very quick look and I don't understand why this has to be a vector to count the number of free entities 🤔

ZehMatt: Its not the count, it holds the index of a free entity.

tupaschoal: Hmm, I see. Thanks!

PR #13927, opened 2021-01-23
[Plugin] Add isVisible and text box widget

IntelOrca:
* Add isVisible property to widgets so that you can easily show / hide widgets.
* Add new text box widget for text input
* Allow plugin windows to be transparent
* Add `textAlign` property to label widget

### DIFF
--- a/src/openrct2-ui/interface/Widget.cpp
+++ b/src/openrct2-ui/interface/Widget.cpp
@@ -829,12 +829,19 @@ static void WidgetDrawImage(rct_drawpixelinfo* dpi, rct_window* w, rct_widgetind
 
 bool WidgetIsEnabled(rct_window* w, rct_widgetindex widgetIndex)
 {
+    if (!WidgetIsVisible(w, widgetIndex))
+        return false;
     return (w->enabled_widgets & (1LL << widgetIndex)) != 0;
 }
 
 bool WidgetIsDisabled(rct_window* w, rct_widgetindex widgetIndex)
 {
     return (w->disabled_widgets & (1LL << widgetIndex)) != 0;
+}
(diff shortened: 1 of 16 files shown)


src/openrct2-ui/interface/Widget.cpp
```diff
@@ -1026,6 +1033,18 @@ void WidgetSetDisabled(rct_window* w, rct_widgetindex widgetIndex, bool value)
     }
 }
 
+void WidgetSetVisible(rct_window* w, rct_widgetindex widgetIndex, bool value)
+{
+    if (value)
```
tupaschoal, line 1038: Maybe `show` instead of `value` ?

src/openrct2-ui/interface/Window.cpp
```diff
@@ -628,11 +628,13 @@ void WindowDrawWidgets(rct_window* w, rct_drawpixelinfo* dpi)
     widgetIndex = 0;
     for (widget = w->widgets; widget->type != WindowWidgetType::Last; widget++)
     {
-        // Check if widget is outside the draw region
-        if (w->windowPos.x + widget->left < dpi->x + dpi->width && w->windowPos.x + widget->right >= dpi->x)
-            if (w->windowPos.y + widget->top < dpi->y + dpi->height && w->windowPos.y + widget->bottom >= dpi->y)
-                WidgetDraw(dpi, w, widgetIndex);
-
+        if (!(widget->flags & WIDGET_FLAGS::IS_HIDDEN))
```
tupaschoal, line 631: Shouldn't we try to use `WidgetIsVisible` everywhere to hide the implementation detail?

PR #13937, opened 2021-01-23
Fix entrance or exit stacked on top considered a duplicate

ZehMatt:
It only checked the x and y axis and ignored the height. After the PR this setup works:

Edit: I just realized the mechanic was also unable to ever get to the ride when it broke down, this fix helps with that also.


Gymnasiast:
LGTM. Perhaps something to put in the changelog.

duncanspumpkin:
This will need a network version bump.

Gymnasiast (approved):

PR #13956, opened 2021-01-26
Implement train view and remove linked_list_index

duncanspumpkin:
I completely reworked this so that we no longer have a vehicle head list.
We now just have the ability to make a train view of the vehicle list.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -63,6 +63,7 @@
 #include "Track.h"
 #include "TrackData.h"
 #include "TrackDesign.h"
+#include "TrainManager.h"
 #include "Vehicle.h"
 
 #include <algorithm>
@@ -4092,12 +4093,10 @@ static Vehicle* vehicle_create_car(
         return nullptr;
 
     auto vehicleEntry = &rideEntry->vehicles[vehicleEntryIndex];
-    auto vehicle = &create_sprite(SpriteIdentifier::Vehicle, carIndex == 0 ? EntityListId::TrainHead : EntityListId::Vehicle)
-                        ->vehicle;
+    auto vehicle = &create_sprite(SpriteIdentifier::Vehicle)->vehicle;
     if (vehicle == nullptr)
         return nullptr;
 
-    vehicle->sprite_identifier = SpriteIdentifier::Vehicle;
     vehicle->ride = rideIndex;
     vehicle->ride_subtype = ride->subtype;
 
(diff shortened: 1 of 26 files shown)


duncanspumpkin:
Would have helped git adding the files. Doh!

src/openrct2/ride/Ride.cpp
```diff
@@ -1008,6 +1009,11 @@ static void ride_remove_cable_lift(Ride* ride)
     {
         ride->lifecycle_flags &= ~RIDE_LIFECYCLE_CABLE_LIFT;
         uint16_t spriteIndex = ride->cable_lift;
+        auto train = GetEntity<Vehicle>(ride->cable_lift);
```
duncanspumpkin, line 1012: remove

src/openrct2/ride/Ride.cpp
```diff
@@ -1036,6 +1042,7 @@ static void ride_remove_vehicles(Ride* ride)
         for (size_t i = 0; i <= MAX_VEHICLES_PER_RIDE; i++)
         {
             uint16_t spriteIndex = ride->vehicles[i];
+            ride->vehicles[i] = SPRITE_INDEX_NULL;
```
duncanspumpkin, line 1045: revert

duncanspumpkin:
Replays are expected to fail. This is due to removal of the field. I've checked them before removing the field and the only difference was the linked_list_index being incorrect for newly created vehicles as expected.

src/openrct2/ride/TrainManager.h
```diff
@@ -0,0 +1,76 @@
...
+ *****************************************************************************/
+#pragma once
+#include <cstdint>
+#include <list>
+
+struct Vehicle;
+
+namespace TrainManager
+{
+    class TrainView
```
ZehMatt, line 17:
```suggestion
    class View
```
Given its in TrainManager its clear that the view is about trains, but not really important.

ZehMatt, line 17: It would be probably also good to comment that the View is only returning vehicle heads, this might confuse someone trying to iterate all vehicles.

src/openrct2/ride/TrainManager.h
```diff
@@ -0,0 +1,76 @@
...
+    {
+    private:
+        const std::list<uint16_t>* vec;
+
+        class TrainViewIterator
```
ZehMatt, line 22:
```suggestion
        class Iterator
```

PR #13963, opened 2021-01-28
Fix #13958: Queue corruption due to intersecting entrances

### LINKED ISSUE #13958: Assertion failed: Peep Ridesubstate should be LeaveEntrance
**OS:** [e.g. Windows 10]
**Version:** [0.3.2]
**Commit/Build:** [8c5cd61]

My server keeps crashing (usually after several hours) with the following message:
image

Here's a save file where it will happen a minute or so after you open it (speed up the game a bit to make it happen faster). Removing all guests from the park will temporarily stop it from happening.
[autosave_2021-01-26_16-43-13.zip](https://github.com/OpenRCT2/OpenRCT2/files/5874470/autosave_2021-01-26_16-43-13.zip)

duncanspumpkin:
There was a whole host of bugs that this issue has exposed which were caused by intersecting the queue of a ride with an entrance/exit or shop. When this was done the peep would be removed from the queue incorrectly and could end up at the wrong station or a variety of other unexpected behaviour

### DIFF
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -2300,11 +2300,21 @@ static void peep_return_to_centre_of_tile(Peep* peep)
  *
  *  rct2: 0x00693f2C
  */
-static void peep_interact_with_entrance(Peep* peep, const CoordsXYE& coords, uint8_t& pathing_result)
+static bool peep_interact_with_entrance(Peep* peep, const CoordsXYE& coords, uint8_t& pathing_result)
 {
     auto tile_element = coords.element;
     uint8_t entranceType = tile_element->AsEntrance()->GetEntranceType();
+    auto rideIndex = tile_element->AsEntrance()->GetRideIndex();
 
+    if ((entranceType == ENTRANCE_TYPE_RIDE_ENTRANCE) || (entranceType == ENTRANCE_TYPE_RIDE_EXIT))
+    {
+        // If an entrance or exit that doesn't belong to the ride we are queuing for ignore the entrance/exit
+        // This can happen when paths clip through entrance/exits
+        if (peep->State == PeepState::Queuing && peep->CurrentRide != rideIndex)
+        {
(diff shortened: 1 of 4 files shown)


src/openrct2/peep/Peep.cpp
```diff
@@ -2880,6 +2890,13 @@ static bool peep_interact_with_shop(Peep* peep, const CoordsXYE& coords)
         return true;
     }
 
+    // If we are queuing ignore the 'shop'
+    // This can happen when paths clip through track
+    if (peep->State == PeepState::Queuing)
```
tupaschoal, line 2895: Shouldn't this be
```c++
 if (peep->State == PeepState::Queuing && peep->CurrentRide != rideIndex)
 ```

Like the other one above?

File: src/openrct2/peep/Peep.h [hunk 2]
```diff
@@ -264,7 +264,8 @@ enum class PeepRideSubState : uint8_t
     LeaveVehicle = 7,
     ApproachExit = 8,
     InExit = 9,
-    // 10, 11 not used
+    InQueue = 10,
+    AtQueueFront = 11,
```
tupaschoal, line 268: Nice find!

src/openrct2/peep/Peep.cpp
duncanspumpkin: So I thought about that for a while. I don't think so. It can only get to here if a track is intersecting a path. So in theory we want it to skip it even if the track is from the ride we are queueing for as we are going to the entrance only. Counter argument to this is that the ride type has already been checked to see if it's a shop. And shops don't have track or queues so it wouldn't matter having the extra check as it could never get to this point.

tupaschoal: Ok, sold!

tupaschoal (approved):
Looks ok, needs the network bump

Gymnasiast (approved):

PR #13969, opened 2021-01-30
[Plugin] Add APIs for editing title sequences

IntelOrca:
So that the title sequence editor can move into a plugin.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -19,6 +19,7 @@
 - Feature: [#13614] Add terrain surfaces from RollerCoaster Tycoon 1.
 - Feature: [#13675] [Plugin] Add context.setInterval and context.setTimeout.
 - Feature: [#13927] [Plugin] Add isVisible and text box widget.
+- Feature: [#13969] [Plugin] Add APIs for editing title sequences.
 - Feature: [#14002] [Plugin] Feature: Use allowed_hosts when checking the binding IP for listening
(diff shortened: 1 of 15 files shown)


spacek531:
The new editor plugin will be available from the main menu? and will it close when loading title sequence parks?

IntelOrca:
@spacek531 no, it will only be available in-game from the map menu due to limitations with plugin framework

Gymnasiast:
> @spacek531 no, it will only be available in-game from the map menu due to limitations with plugin framework. Once in-game, it has the same behaviour as before.

I would like that to be possible, though, before the built-in title sequence editor gets removed.

But since this PR merely adds the necessary APIs, I merged it anyway.

IntelOrca:
Is there a specific reason why it is important that this still be available on the title screen?

spacek531:
The title sequence editor opening from the options menu puts the title sequence selector and title sequence editor in the same place. Also means when I start up the game I can go straight into editing without having to decide on an arbitrary park to open just so I can close it again.

IntelOrca:
> The title sequence editor opening from the options menu puts the title sequence selector and title sequence editor in the same place.

Can you elaborate on this? Is this something that can be done when not on the title screen?

> Also means when I start up the game I can go straight into editing without having to decide on an arbitrary park to open just so I can close it again.

Is it okay if it is just a menu item under the toolbox and takes you to a blank map, with the editor open?

spacek531:
image
This is what I mean by "the same place" - all the title sequence stuff is in the same groupbox, a central area for all title sequence-related actions.

PR #13970, opened 2021-01-30
Use FixedVector for paint_structs

ZehMatt:
Using the FixedVector makes this a little easier to read and also got rid of two variables. I also re-ordered the members, MSVC generates 20 x86-64 instructions less now.


duncanspumpkin:
All seems sensible. Assume you have tested this comprehensively?

ZehMatt:
Yeah, the usual parks.

duncanspumpkin (approved):

PR #13972, opened 2021-01-30
Remove unused variable gCurrentFontFlags

ZehMatt:
It only carries the value zero so might as well remove it.

### DIFF
--- a/src/openrct2/drawing/Drawing.String.cpp
+++ b/src/openrct2/drawing/Drawing.String.cpp
@@ -337,7 +337,6 @@ void draw_string_centred_raw(rct_drawpixelinfo* dpi, const ScreenCoordsXY& coord
     gCurrentFontSpriteBase = FONT_SPRITE_BASE_MEDIUM;
     gfx_draw_string(dpi, "", COLOUR_BLACK, screenCoords);
     screenCoords = coords;
-    gCurrentFontFlags = 0;
 
     for (int32_t i = 0; i <= numLines; i++)
     {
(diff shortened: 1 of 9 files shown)


src/openrct2/drawing/Drawing.String.cpp
```diff
@@ -1028,7 +1025,6 @@ void gfx_draw_string_with_y_offsets(
     std::memcpy(text_palette, info.palette, sizeof(info.palette));
 
     gCurrentFontSpriteBase = info.font_sprite_base;
-    gCurrentFontFlags = info.flags;
```
tupaschoal, line 1031: Wouldn't it get new values here? Or is `info.flags` never modified? Because if it's the latter, then we can also delete the `flags`

ZehMatt: Earlier on it would do ```info.flags = gCurrentFontFlags``` and info.flags is used in some cases.

duncanspumpkin:
Do we know what it used to be used for?

src/openrct2/drawing/Drawing.String.cpp
Gymnasiast: To me, it seems like it basically saves whatever info.flags used to be before. But since the only thing that actually seems to vary in `info.flags` is whether or not `TEXT_DRAW_FLAG_TTF` is set, and we don't want that to stick anyway, it seems like nothing useful is written to `gCurrentFontFlags`.

Did you make sure to test with English/German (for the sprite font) and a TTF language (like Korean) to see if it did not break anything?

IntelOrca:
I thought it was for things like measuring text where it needed to know the size a mock draw text call.

src/openrct2/drawing/Drawing.String.cpp
ZehMatt: Everything seems to be working as expected. Given I eliminated all references and the only write was info.flags or 0 I already expected that nothing changes.

ZehMatt:
I think a lot of refactor work made this global variable simply obsolete without really realizing it. At some point the draw_text_info struct was introduced eliminating some of its use already, I haven't really tracked down the previous use as that seems a bit trivial to do now.

Issue #13815, opened 2021-01-12
Refactor window to class: Water


IntelOrca:
See #13737 for more details.

ankushduacodes:
Hi @IntelOrca, I am new to the repo and new to C++. I know basic OOP constructs like classes, Could you guide me through this issue. I am very enthusiastic to be part of OpenRCT2 and to learn intermediate C++ along with these type of contributions.

Thank you!..

IntelOrca:
Have a look at the changes in the following pull requests: #13756 and #13819
These are good examples to copy when doing the water window.

sstream17:
I just started browsing the good first issue label and came across this issue. It looks like the work has been finished in #13974

IntelOrca:
@sstream17 you're right, thanks for notifying us

PR #13974, opened 2021-01-30
Refactor water window into Window class #13815

Gecko05:
This PR addresses #13815 

Refactored the code from water window into a new class called WaterWindow which inherits from Window class and overrides the On methods.
Added a new generic method for WindowFocusOrCreate that additionally takes position as an argument to create a window.

### DIFF
--- a/src/openrct2-ui/windows/Water.cpp
+++ b/src/openrct2-ui/windows/Water.cpp
@@ -37,182 +37,148 @@ static rct_widget window_water_widgets[] = {
     MakeRemapWidget({43, 32}, {16, 16}, WindowWidgetType::TrnBtn, WindowColour::Tertiary, SPR_LAND_TOOL_INCREASE, STR_ADJUST_LARGER_WATER_TIP),  // increment size
     { WIDGETS_END },
 };
-
-static void window_water_close(rct_window *w);
-static void window_water_mouseup(rct_window *w, rct_widgetindex widgetIndex);
-static void window_water_mousedown(rct_window *w, rct_widgetindex widgetIndex, rct_widget *widget);
-static void window_water_update(rct_window *w);
-static void window_water_invalidate(rct_window *w);
-static void window_water_paint(rct_window *w, rct_drawpixelinfo *dpi);
-static void window_water_textinput(rct_window *w, rct_widgetindex widgetIndex, char *text);
-static void window_water_inputsize(rct_window *w);
-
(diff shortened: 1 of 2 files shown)


Gymnasiast (commented):
Tested it. Seems to work correctly. I'll leave checking the code to @IntelOrca .

src/openrct2-ui/windows/Water.cpp
```diff
@@ -37,182 +37,148 @@ static rct_widget window_water_widgets[] = {
...
-    // Draw number for tool sizes bigger than 7
-    if (gLandToolSize > MAX_TOOL_SIZE_WITH_SPRITE)
+    void OnDraw(rct_drawpixelinfo& dpi) override
     {
-        gfx_draw_string_centred(
-            dpi, STR_LAND_TOOL_SIZE_VALUE, screenCoords - ScreenCoordsXY{ 0, 2 }, COLOUR_BLACK, &gLandToolSize);
+        auto screenCoords = ScreenCoordsXY{ windowPos.x + window_water_widgets[WIDX_PREVIEW].midX(),
+                                            windowPos.y + window_water_widgets[WIDX_PREVIEW].midY() };
+
+        DrawWidgets(dpi);
```
JoelLarson, line 145: I believe this should be `DrawWidgets(&dpi);` with the change from *dpi to &dpi

src/openrct2-ui/windows/Water.cpp
```diff
@@ -37,182 +37,148 @@ static rct_widget window_water_widgets[] = {
...
+        {
+            Close();
+        }
+    }
 
-/**
- *
- *  rct2: 0x006E6AB8
- */
-static void window_water_invalidate(rct_window* w)
```
JoelLarson, line 179: Is `window_water_invalidate()` mapped to `OnPrepareDraw()` ? I wasn't sure what to do with this in my scenery window PR.

src/openrct2-ui/windows/Water.cpp
JoelLarson: I just noticed this was changed from `WindowDrawWidgets()`. My comment is invalid.

src/openrct2-ui/windows/Water.cpp
duncanspumpkin: Yes we renamed invalidate to prepare draw to better reflect what it is being used for

IntelOrca (approved):

PR #13975, opened 2021-01-31
Introduce TileElementsView

ZehMatt:
This greatly simplifies the way we loop over a single tile to obtain its elements. This should eliminate a bunch of boilerplate code such as checking for nullptr, checking for the type.

Examples:

Generic loop
```
// Obtain all tile elements.
for(auto* tileElement : TileElementsView(pos))
{
    // tileElement has type "TileElement"
}
```

Type based loop
```
// Obtains all tile elements which are 'SurfaceElement'.
for(auto* tileElement : TileElementsView<SurfaceElement>(pos))
{
    // tileElement has type "SurfaceElement"
}
```

Another thing a lot of those loops have in common is to skip ghost tiles, I was wondering I should perhaps add a second template parameter to skip ghosts, but this is something that can be improved upon later.

Another change is how the conversation worked, you can now simply do:
```
auto* surface = tileElement->as<SurfaceElement>();
```
It is no longer needed to pass TileElementType


src/openrct2/world/Map.h
```diff
@@ -297,4 +297,120 @@ uint16_t check_max_allowable_land_rights_for_tile(const CoordsXYZ& tileMapPos);
 void FixLandOwnershipTiles(std::initializer_list<TileCoordsXY> tiles);
 void FixLandOwnershipTilesWithOwnership(std::initializer_list<TileCoordsXY> tiles, uint8_t ownership);
 
+template<typename T = TileElement> class TileElementsView
```
duncanspumpkin, line 300: Can we put this in a separate header file

IntelOrca:
In OpenLoco, we do this, but it is effectively a Tile class where you also have access to the position and various other helpers.

ZehMatt:
I did not refactor all the things that still do the manual loop, there is a lot of refactor potential so it should be smaller PRs doing that. We can also create issues so people can pickup on that.

duncanspumpkin (approved):

PR #13977, opened 2021-01-31
Fix issues with spinning control toggle

spacek531:
Everything except the changes to import/export I know to be necessary. I tested with the uin16_t change but without the import/export after I thought I snuck the track element into it, but ratings did not seem to be affected in any way. I added the import/export stuff just to be on the safe side.

### DIFF
--- a/src/openrct2/ride/RideRatings.h
+++ b/src/openrct2/ride/RideRatings.h
@@ -14,6 +14,7 @@
 #include "RideTypes.h"
 
 using ride_rating = fixed16_2dp;
+using track_type_t = uint16_t;
 
 // Convenience function for writing ride ratings. The result is a 16 bit signed
 // integer. To create the ride rating 3.65 type RIDE_RATING(3,65)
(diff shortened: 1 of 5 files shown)


spacek531:
Todo: test this with exporting track elements 255 and 256 as TD6

src/openrct2/ride/RideRatings.h
```diff
@@ -44,7 +44,7 @@ struct RideRatingCalculationData
     CoordsXYZ ProximityStart;
     ride_id_t CurrentRide;
     uint8_t State;
-    uint8_t ProximityTrackType;
+    uint16_t ProximityTrackType;
```
Gymnasiast, line 47: Why not track_type_t ?

Gymnasiast (commented):
Appears to work correctly with all three default designs, whereas Dizzymouse could not be built before.

I also tried saving a track designs with a spinning control toggle, which appears to have worked.

Gymnasiast (approved):

Issue #13799, opened 2021-01-12
Refactor window to class: Scenery


IntelOrca:
See #13737 for more details.

JoelLarson:
Starting on this

PR #13980, opened 2021-02-01
#13799 - Refactor scenery window

JoelLarson:
I've started a first pass at refactoring the scenery window for #13799. There are a couple of spots that I'm not sure on and will need to figure out before I can call this done.

They are marked with "TODO:" in the comments throughout the file.

### DIFF
--- a/src/openrct2-ui/windows/Scenery.cpp
+++ b/src/openrct2-ui/windows/Scenery.cpp
@@ -180,322 +180,30 @@ void window_scenery_update_scroll(rct_window* w);
 // rct2: 0x00F64F2C
 static ScenerySelection window_scenery_tab_entries[SCENERY_WINDOW_TABS][SCENERY_ENTRIES_PER_TAB + 1];
 
-/**
- * Was part of 0x006DFA00
- * The same code repeated five times for every scenery entry type
- */
-static void init_scenery_entry(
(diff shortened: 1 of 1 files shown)


JoelLarson:
This PR also sucks to review. I may try to split it up by method moves/refactor so it's easy to follow the diffs but it wont be this week

src/openrct2-ui/windows/Scenery.cpp
```diff
@@ -180,779 +180,1127 @@ void window_scenery_update_scroll(rct_window* w);
...
+                }
+            }
+        }
+    }
+
+    /**
+     *
+     *  rct2: 0x006E1A91
+     */
+    ScreenSize OnScrollGetSize(int32_t scrollIndex /*, int32_t* width, int32_t* height */) override // TODO: width/height
```
JoelLarson, line 625: I've figured out that width and height are overridable with the new Window implementation. It seems like I only care about height here though and I can't have it ignore width by doing something like:

```
int32_t rows = window_scenery_count_rows();
auto height = window_scenery_rows_height(rows);

return { nullptr, height };
```

Any suggestions?

IntelOrca: yes, just return 0

JoelLarson:
Been busy the past few weeks. I'm planning to complete this by March 1

PR #13983, opened 2021-02-02
Check for correct peep state when removing mass

duncanspumpkin:
This has exposed a bunch of things related to peep references that I want to explore. It looks like vehicles (that aren't ferriswheels) do not clean up after themselves when removing a peep from their vehicle. This should be changed as this will leave potentially invalid references in the vehicle.

### DIFF
--- a/src/openrct2/actions/SetCheatAction.cpp
+++ b/src/openrct2/actions/SetCheatAction.cpp
@@ -644,7 +644,11 @@ void SetCheatAction::RemoveAllGuests() const
                     auto peep = TryGetEntity<Guest>(peepInTrainIndex);
                     if (peep != nullptr)
                     {
-                        vehicle->mass -= peep->Mass;
+                        if ((peep->State == PeepState::OnRide && peep->RideSubState == PeepRideSubState::OnRide)
+                            || (peep->State == PeepState::LeavingRide && peep->RideSubState == PeepRideSubState::LeaveVehicle))
+                        {
+                            vehicle->mass -= peep->Mass;
+                        }
                     }
                     peepInTrainIndex = SPRITE_INDEX_NULL;
                 }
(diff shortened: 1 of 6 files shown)


src/openrct2/actions/SetCheatAction.cpp
```diff
@@ -644,7 +644,11 @@ void SetCheatAction::RemoveAllGuests() const
                     auto peep = TryGetEntity<Guest>(peepInTrainIndex);
                     if (peep != nullptr)
                     {
-                        vehicle->mass -= peep->Mass;
+                        if ((peep->State == PeepState::OnRide && peep->RideSubState == PeepRideSubState::OnRide)
+                            || (peep->State == PeepState::LeavingRide && peep->RideSubState == PeepRideSubState::LeaveVehicle))
+                        {
+                            vehicle->mass -= peep->Mass;
```
IntelOrca, line 650:
```suggestion
if (vehicle->mass > peep->Mass)
    vehicle->mass -= peep->Mass;
else
    vehicle->mass = 0;
```

IntelOrca: Or
```suggestion
                            vehicle->ApplyMass(-peep->Mass);
```

duncanspumpkin: Why though? We don't verify in any other location adding/removing.

IntelOrca: Yeah we should do it though. Ideally we prevent all possible underflows / overflows. What reason is there not to?

duncanspumpkin: Okay I've added a function for it.

IntelOrca (approved):

PR #13990, opened 2021-02-03
Reset peep index to null

### LINKED ISSUE #13989: Refactor of Vehicle->Peep[]
Whilst investigating #13234 it showed that when a peep leaves a vehicle its peep reference within the vehicle (`vehicle->peep[]`) is not reset back to `SPRITE_INDEX_NULL`. Apart from being one of the reasons why #13234 happened its also not a very sensible way of operating. Plugins might incorrectly assume that the reference is valid and suffer from the same mistake as #13234. Therefore I suggest we refactor so that when a peeps mass is removed from a vehicle we also reset its index to `SPRITE_INDEX_NULL` in the respective vehicle struct. This may also allow us to refactor the `peep[]` array to simplify working out which seats are free. At present cars always load from front to back but we could allow for some randomisation if we know which seats are occupied. 

This issue is to refactor the peep leave vehicle method to set the `vehicle->peep[]` to `SPRITE_INDEX_NULL`. Will need rerecording of replays.

duncanspumpkin:
Fixes #13989 and applies even more checks to the remove all guests cheat.
Will break replays.

### DIFF
--- a/src/openrct2/actions/SetCheatAction.cpp
+++ b/src/openrct2/actions/SetCheatAction.cpp
@@ -644,10 +644,15 @@ void SetCheatAction::RemoveAllGuests() const
                     auto peep = TryGetEntity<Guest>(peepInTrainIndex);
                     if (peep != nullptr)
                     {
-                        if ((peep->State == PeepState::OnRide && peep->RideSubState == PeepRideSubState::OnRide)
-                            || (peep->State == PeepState::LeavingRide && peep->RideSubState == PeepRideSubState::LeaveVehicle))
+                        // Due to mistake in vanilla saves may have peep indexs that are invalid
+                        if (peep->CurrentRide == vehicle->ride)
                         {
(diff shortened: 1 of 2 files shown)


src/openrct2/actions/SetCheatAction.cpp
```diff
@@ -644,10 +644,15 @@ void SetCheatAction::RemoveAllGuests() const
                     auto peep = TryGetEntity<Guest>(peepInTrainIndex);
                     if (peep != nullptr)
                     {
-                        if ((peep->State == PeepState::OnRide && peep->RideSubState == PeepRideSubState::OnRide)
-                            || (peep->State == PeepState::LeavingRide && peep->RideSubState == PeepRideSubState::LeaveVehicle))
+                        // Due to mistake in vanilla saves may have peep indexs that are invalid
```
ZehMatt, line 647:
```suggestion
                        // Due to mistake in vanilla saves may have peep indices that are invalid
```

PR #13994, opened 2021-02-04
Make tile element insertion type explicit

ZehMatt:
Tile elements should not be inserted without a type, this also fixes one issue where it attempts to fix some surface elements but never and calls AsSurface but the type was not set so it would result in nullptr and crash the game.

This is based on #13975 so I will rebase this branch later.

### DIFF
--- a/src/openrct2/Game.cpp
+++ b/src/openrct2/Game.cpp
@@ -481,14 +481,12 @@ void game_fix_save_vars()
             if (surfaceElement == nullptr)
             {
                 log_error("Null map element at x = %d and y = %d. Fixing...", x, y);
-                auto tileElement = tile_element_insert(
-                    TileCoordsXYZ{ x, y, 14 }.ToCoordsXYZ(), 0b0000, TileElementType::Surface);
-                if (tileElement == nullptr)
+                surfaceElement = TileElementInsert<SurfaceElement>(TileCoordsXYZ{ x, y, 14 }.ToCoordsXYZ(), 0b0000);
+                if (surfaceElement == nullptr)
                 {
                     log_error("Unable to fix: Map element limit reached.");
                     return;
                 }
-                surfaceElement = tileElement->AsSurface();
             }
 
             // Fix the invisible border tiles.
--- a/src/openrct2/Game.cpp
+++ b/src/openrct2/Game.cpp
@@ -481,7 +481,8 @@ void game_fix_save_vars()
             if (surfaceElement == nullptr)
             {
                 log_error("Null map element at x = %d and y = %d. Fixing...", x, y);
-                auto tileElement = tile_element_insert(TileCoordsXYZ{ x, y, 14 }.ToCoordsXYZ(), 0b0000);
+                auto tileElement = tile_element_insert(
+                    TileCoordsXYZ{ x, y, 14 }.ToCoordsXYZ(), 0b0000, TileElementType::Surface);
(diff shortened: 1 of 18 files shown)


src/openrct2/world/Map.cpp
```diff
@@ -1184,7 +1153,7 @@ TileElement* tile_element_insert(const CoordsXYZ& loc, int32_t occupiedQuadrants
 
     // Insert new map element
     insertedElement = newTileElement;
-    newTileElement->type = 0;
+    newTileElement->SetType(static_cast<uint8_t>(type));
```
Gymnasiast, line 1156: This is not entirely the same, I think. The `type` field contains more information besides just the type and now not everything is 0-initialised.

src/openrct2/world/TileInspector.cpp
```diff
@@ -382,7 +382,7 @@ GameActionResultPtr tile_inspector_paste_element_at(const CoordsXY& loc, TileEle
 
         // The occupiedQuadrants will be automatically set when the element is copied over, so it's not necessary to set them
         // correctly _here_.
-        TileElement* const pastedElement = tile_element_insert({ loc, element.GetBaseZ() }, 0b0000);
+        TileElement* const pastedElement = tile_element_insert({ loc, element.GetBaseZ() }, 0b0000, TileElementType::Corrupt);
```
duncanspumpkin, line 385: I don't think you should use Corrupt here. Use anything but corrupt as we will eventually remove corrupt. Eitherway its overwritten on line 388

src/openrct2/world/Map.cpp
ZehMatt: Good point, I'll  initialize type to 0 and call SetType afterwards.

src/openrct2/Game.cpp
```diff
@@ -481,13 +481,13 @@ void game_fix_save_vars()
             if (surfaceElement == nullptr)
             {
                 log_error("Null map element at x = %d and y = %d. Fixing...", x, y);
-                auto tileElement = tile_element_insert(TileCoordsXYZ{ x, y, 14 }.ToCoordsXYZ(), 0b0000);
+                auto tileElement = TileElementInsert<SurfaceElement>(TileCoordsXYZ{ x, y, 14 }.ToCoordsXYZ(), 0b0000);
```
duncanspumpkin, line 484: No need for this temporary variable

duncanspumpkin (approved):

src/openrct2/actions/WallPlaceAction.cpp
```diff
@@ -427,12 +419,11 @@ GameActions::Result::Ptr WallPlaceAction::Execute() const
...
-    if (GetFlags() & GAME_COMMAND_FLAG_GHOST)
-    {
-        wallElement->SetGhost(true);
-    }
+    wallElement->SetGhost(GetFlags() & GAME_COMMAND_FLAG_GHOST);
 
-    res->tileElement = tileElement;
+    res->tileElement = wallElement->as<TileElement>();
+
+    map_animation_create(MAP_ANIMATION_TYPE_WALL, targetLoc);
```
Gymnasiast, line 426: Any reason you moved this line?

ZehMatt: It looked a bit out of place between the creation and initialization of the tile element.

Gymnasiast (approved):

PR #13998, opened 2021-02-05
Use TileElementsView and minor cleanups

ZehMatt:
I've tested all commits and the replays are still fine, so this should be good.

### DIFF
--- a/src/openrct2/actions/FootpathPlaceAction.cpp
+++ b/src/openrct2/actions/FootpathPlaceAction.cpp
@@ -20,7 +20,10 @@
 #include "../world/Park.h"
 #include "../world/Scenery.h"
 #include "../world/Surface.h"
+#include "../world/TileElementsView.h"
 #include "../world/Wall.h"
+
+using namespace OpenRCT2;
 
 FootpathPlaceAction::FootpathPlaceAction(const CoordsXYZ& loc, uint8_t slope, ObjectEntryIndex type, Direction direction)
     : _loc(loc)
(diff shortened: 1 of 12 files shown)


src/openrct2/actions/FootpathPlaceAction.cpp
```diff
@@ -435,20 +438,19 @@ void FootpathPlaceAction::RemoveIntersectingWalls(PathElement* pathElement) cons
...
-        if (tileElement->GetType() == TILE_ELEMENT_TYPE_PATH && tileElement->GetBaseZ() == footpathPos.z
-            && (tileElement->AsPath()->IsSloped() == isSloped)
-            && (tileElement->AsPath()->GetSlopeDirection() == (slope & FOOTPATH_PROPERTIES_SLOPE_DIRECTION_MASK)))
-        {
-            return tileElement->AsPath();
-        }
-    } while (!(tileElement++)->IsLastForTile());
+        if (pathElement->GetBaseZ() != footpathPos.z)
+            continue;
+        if (pathElement->IsSloped() == isSloped)
```
Gymnasiast, line 448:
```suggestion
        if (pathElement->IsSloped() != isSloped)
```

src/openrct2/actions/SmallSceneryRemoveAction.cpp
```diff
@@ -132,26 +135,18 @@ GameActions::Result::Ptr SmallSceneryRemoveAction::Execute() const
...
-        if ((tileElement->AsSmallScenery()->GetSceneryQuadrant()) != _quadrant)
+        if (sceneryElement->GetSceneryQuadrant() != _quadrant)
             continue;
-        if (tileElement->GetBaseZ() != _loc.z)
+        if (sceneryElement->GetBaseZ() != _loc.z)
             continue;
-        if (tileElement->AsSmallScenery()->GetEntryIndex() != _sceneryType)
+        if (sceneryElement->GetEntryIndex() != _sceneryType)
             continue;
-        if ((GetFlags() & GAME_COMMAND_FLAG_GHOST) && tileElement->IsGhost() == false)
```
Gymnasiast, line 149: This is subtly different. Previously, it would accept this action being called with the `GAME_COMMAND_FLAG_GHOST` *not* set and the tile element being a ghost (but not the other way round).

I'm not sure if that was correct or not.

src/openrct2/actions/FootpathPlaceAction.cpp
ZehMatt: Good catch

src/openrct2/actions/SmallSceneryRemoveAction.cpp
ZehMatt: You are right, however I believe the change is now correct. If ghost flag is set in the action should only interact with ghost elements. I will do some testing.

ZehMatt: I think the reason it previously worked was because you only ever have one ghost, this would have probably caused issue if you had more than one.

ZehMatt: I can't get any difference using the old logic, so it seems fine.

Gymnasiast (commented):
Code looks good, still needs testing.

Gymnasiast (approved):

PR #14001, opened 2021-02-05
Fix #13933: Clients do not run server plugins

### LINKED ISSUE #13933: Clients do not run server plugins
**OS:** Manjaro Linux
**Version:** 0.3.2 (stable) and 0.3.2-437 (latest)
**Commit/Build:** cea5fab (stable) and eb52391 (latest)

Maybe I'm just doing something wrong, but based on https://github.com/OpenRCT2/OpenRCT2/blob/develop/distribution/scripting.md, remote plugins should automatically be uploaded to all connecting users so that every user is running the same code on the same ticks (and thus avoiding network desyncs). I'm also assuming that would include the plugins config as well. However it seems that this plugin uploading process doesn't work?

For example: https://github.com/OpenRCT2/plugin-samples/blob/master/subscriptions.js. Although for this example the network desync error does not occur, the cash on the server and the cash on connected clients day by day gets further and further out of sync. Other plugins (such as the ride-price-manager) do more modifications which eventually leads to network desyncs.

I assumed that since the `park.cash += 100;` was within a day subscription, all clients would get the plugin and run this code on each day. Even running OpenRCT2 from the terminal and adding some console.logs to the script doesn't show the plugin even running on the clients, only the server.

I also tried this with a headless OpenRCT2 server, still doesn't send the remote scripts to the clients. The headless server does load all of the local plugins for some reason though; that might be another bug, or I might not understand what local means.

I love the idea of plugins, and I have some ideas for ones I want to write. I just would like to have them work when I host parks for friends :)

**Steps to reproduce:**
1. Put the https://github.com/OpenRCT2/plugin-samples/blob/master/subscriptions.js file in the OpenRCT2 plugin folder
2. (optional) change `park.cash += 100;` to something larger, so the difference is more noticeable.
3. (optional) remove `daySubscription.dispose();`, so that we always add cash no matter what for testing.
4. Run a server, notice your cash go up every in game day.
5. Have a client connect, notice that their cash isn't going up, presumably since they aren't running the plugin.

**Save game:**
None, just start a new scenario. Not save specific.

IntelOrca:


tupaschoal:
Should this have had a changelog entry?

Issue #13967, opened 2021-01-29
Display where a track design is located inside the track selector


Ryder17z:
As my collection of tracks has grown substantially, I've moved custom tracks into different subfolders of OpenRCT2/track/ to make it easier to sort them.

I would like to have the path added at the bottom of the window, similar to how the scenario selector lists the paths.

![image](https://user-images.githubusercontent.com/2000703/106329499-d0e67700-6281-11eb-90a1-362edd80c7d0.png)

kimjcalvin:
@Ryder17z by scenario selector do you mean 'Load Game' window from the title screen? For 'Start a new game' -> 'Select scenario for new game' , I don't see the path list.

Ryder17z:
@kimjcalvin 
![https://user-images.githubusercontent.com/10726524/39252015-a9ef89a2-48df-11e8-82d0-9e14652f6c12.png](https://user-images.githubusercontent.com/10726524/39252015-a9ef89a2-48df-11e8-82d0-9e14652f6c12.png)

I don't know why it's removed currently and i'm not a fan of that change

CorySanin:
It's not removed. It's at the bottom of the window. It's partially covered up by the park information window.

Ryder17z:
> It's not removed. It's at the bottom of the window. It's partially covered up by the park information window.

you misunderstood me. I mean I don't see it in version d26590c (latest found on flathub)

IntelOrca:
You need to have debug options enabled in settings for it to show.

kimjcalvin:
Ok will start working on this issue thanks guys

Ryder17z:
I wasn't aware of it being only visible when debug tools are enabled.

PR #14019, opened 2021-02-08
Display where a track design is located inside the track selector

kimjcalvin:
Issue: [ #13967 ]

[Code is similar to lines 470-480 in `ScenarioSelect.cpp` ](https://github.com/OpenRCT2/OpenRCT2/blob/e5b412fb5fa8f5b640cbe12a2affc17ddd6f2997/src/openrct2-ui/windows/ScenarioSelect.cpp#L469)

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,5 +1,6 @@
 0.3.3+ (in development)
 ------------------------------------------------------------------------
+- Feature: [#13967] Track List window now displays the path to the design when debugging tools are on.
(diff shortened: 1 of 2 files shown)


Gymnasiast:
Thanks for your PR!

The CIs are very strict about code formatting. You can see what needs changing here: https://github.com/OpenRCT2/OpenRCT2/pull/14019/checks?check_run_id=1857375899

You can change above things manually, but you can also use clang-format (or git clang-format, if you set it up correctly) to autoformat before committing.

Gymnasiast (approved):

PR #14021, opened 2021-02-08
Refactor flat ride track pieces

Gymnasiast:


src/openrct2/world/TileInspector.cpp
```diff
@@ -29,6 +29,8 @@
 #include "Scenery.h"
 #include "Surface.h"
 
+#include <src/openrct2/ride/TrackData.h>
```
Gymnasiast, line 32: My IDE is a dick.

src/openrct2/actions/TrackPlaceAction.cpp
```diff
@@ -341,15 +313,8 @@ GameActions::Result::Ptr TrackPlaceAction::Query() const
...
-        if (ride_type_has_flag(ride->type, RIDE_TYPE_FLAG_FLAT_RIDE))
-        {
-            entranceDirections = FlatRideTrackSequenceProperties[_trackType][0];
-        }
-        else
-        {
-            entranceDirections = TrackSequenceProperties[_trackType][0];
-        }
+        int32_t entranceDirections = TrackSequenceProperties[_trackType][0];
+        ;
```
say25, line 317:
```suggestion
```

duncanspumpkin:
Replays are only different due to the change in track id. All okay to rerecord.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -802,10 +803,7 @@ class S6Importer final : public IParkImporter
         dst.ProximityStart = { src.proximity_start_x, src.proximity_start_y, src.proximity_start_z };
         dst.CurrentRide = src.current_ride;
         dst.State = src.state;
-        dst.ProximityTrackType = src.proximity_track_type;
-        if (src.proximity_track_type == TrackElemType::RotationControlToggleAlias
-            && !RCT2TrackTypeIsBooster(_s6.rides[src.current_ride].type, src.proximity_track_type))
-            dst.ProximityTrackType = TrackElemType::RotationControlToggle;
+        dst.ProximityTrackType = RCT2TrackTypeToOpenRCT2(src.proximity_track_type, _s6.rides[src.current_ride].type);
```
duncanspumpkin, line 806: replays are crashing on this if the current ride is null then this will throw. Also `proximity_track_type` could be 0xFF to imply null this is a subtle bug from https://github.com/OpenRCT2/OpenRCT2/pull/13977.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -781,9 +781,7 @@ void S6Exporter::ExportRideRatingsCalcData()
     dst.proximity_start_z = src.ProximityStart.z;
     dst.current_ride = src.CurrentRide;
     dst.state = src.State;
-    dst.proximity_track_type = static_cast<uint8_t>(src.ProximityTrackType);
-    if (src.ProximityTrackType == TrackElemType::RotationControlToggle)
-        dst.proximity_track_type = static_cast<uint8_t>(TrackElemType::RotationControlToggleAlias);
+    dst.proximity_track_type = OpenRCT2TrackTypeToRCT2(src.ProximityTrackType);
```
duncanspumpkin, line 784: Need to think about the 0xFF / none issue here as well.

src/openrct2/rct2/S6Importer.cpp
Gymnasiast: Hm, but 255 was also the quarter loop of the multi-dim. What if the proximity_track_type was that piece?

duncanspumpkin: Yeah it's a bit of a mess. I think though you can get away with just checking for null rides and just pretend that the track null issue doesn't exist.

Gymnasiast: Hm, before I saw this I already pushed a commit that assumed that 255 was null (considering the special checks in RideRating, that seems safest to me). Is it okay to keep that?

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -803,7 +803,12 @@ class S6Importer final : public IParkImporter
         dst.ProximityStart = { src.proximity_start_x, src.proximity_start_y, src.proximity_start_z };
         dst.CurrentRide = src.current_ride;
         dst.State = src.state;
-        dst.ProximityTrackType = RCT2TrackTypeToOpenRCT2(src.proximity_track_type, _s6.rides[src.current_ride].type);
+        if (src.proximity_track_type == 0xFF)
```
duncanspumpkin, line 806: I'm not sure this will work. I think the first thing to check would be if the current ride is invalid.

Gymnasiast: Should I set it to null if `src.proximity_track_type == 0xFF || src.current_ride >= RCT12_MAX_RIDES_IN_PARK`?

duncanspumpkin: So I've looked into it and its impossible to get the information we need. We have lost too much information (really we never had the information)  due to the change in track types! We are just going to have to do something hacky and modify the ride rating code to get out of this.

Change the if to the following:
```
if (src.current_ride < RCT12_MAX_RIDES_IN_PARK && _s6.rides[src.current_ride].type != RIDE_TYPE_NULL)
            dst.ProximityTrackType = RCT2TrackTypeToOpenRCT2(src.proximity_track_type, _s6.rides[src.current_ride].type);
        else
            dst.ProximityTrackType = 0xFF;
```
We will then need to modify the ride rating checks L335 and L229 to add an additional hack:
```
// TODO: Hack to be removed with new save format trackType 0xFF should not be here
if (trackType == 0xFF || trackType == TrackElemType::None || ...
```

Gymnasiast: I just thought of something: do we even need to import these at all? Can't we just reset the ratings calc data upon import? 

The only downside I see is that it may take a bit longer for the data to come up if you save and reload during calculation, since it would have to start from scratch.

duncanspumpkin: Yes we do need it. Peep decisions are changed depending on rating. Changes to when the rating is available will break the simulation

duncanspumpkin (approved):

PR #14023, opened 2021-02-08
Move peep pickup test into Peep::CanBePickedUp

ZehMatt:


IntelOrca (approved):

tupaschoal:
We merged a switch over an array, yay! :D

ZehMatt:
> We merged a switch over an array, yay! :D

Worse, it was std::map

PR #14024, opened 2021-02-08
Use CoordsXY for for some types

ZehMatt:
The next refactor should be probably converting Peep x and y into CoordsXY. 

NOTE: Replays require an update, CoordsXY uses 32 bit instead of 16, so the checksums will mismatch.

### DIFF
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -167,8 +167,7 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
             newTile.y += offset;
         }
     }
-    peep->SetDestination(newTile);
-    peep->DestinationTolerance = 2;
+    peep->SetDestination(newTile, 2);
 
     return 0;
 }
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -132,9 +132,6 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
     }
 
     peep->PeepDirection = direction;
-    peep->DestinationX = newTile.x;
-    peep->DestinationY = newTile.y;
-    peep->DestinationTolerance = 2;
     if (peep->State != PeepState::Queuing)
     {
         // When peeps are walking along a path, we would like them to be spread out across the width of the path,
@@ -154,24 +151,25 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
         // coordinate constant, but instead clamp it to an acceptable range. This brings in 'outlier' guests from
         // the edges of the path, while allowing guests who are already in an acceptable position to stay there.
 
-        int8_t offset = (scenario_rand() & 7) - 3;
+        const int8_t offset = (scenario_rand() & 7) - 3;
         if (direction == 0 || direction == 2)
         {
             // Peep is moving along X, so apply the offset to the X position of the destination and clamp their current Y
-            peep->DestinationX += offset;
-            const uint16_t centerLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
-            peep->DestinationY = std::clamp(
-                peep->y, static_cast<int16_t>(centerLine - 3), static_cast<int16_t>(centerLine + 3));
+            const int32_t centerLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
+            newTile.x += offset;
+            newTile.y = std::clamp<int32_t>(peep->y, centerLine - 3, centerLine + 3);
         }
         else
         {
             // Peep is moving along Y, so apply the offset to the Y position of the destination and clamp their current X
-            const uint16_t centerLine = (peep->x & 0xFFE0) + COORDS_XY_HALF_TILE;
-            peep->DestinationX = std::clamp(
-                peep->x, static_cast<int16_t>(centerLine - 3), static_cast<int16_t>(centerLine + 3));
-            peep->DestinationY += offset;
+            const int32_t centerLine = (peep->x & 0xFFE0) + COORDS_XY_HALF_TILE;
+            newTile.x = std::clamp<int32_t>(peep->x, centerLine - 3, centerLine + 3);
+            newTile.y += offset;
         }
     }
+    peep->SetDestination(newTile);
+    peep->DestinationTolerance = 2;
+
     return 0;
 }
 
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -155,15 +155,15 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
         if (direction == 0 || direction == 2)
         {
             // Peep is moving along X, so apply the offset to the X position of the destination and clamp their current Y
-            const int32_t centerLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
+            const int32_t centreLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
             newTile.x += offset;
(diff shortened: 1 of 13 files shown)


src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -158,18 +157,18 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
         if (direction == 0 || direction == 2)
         {
             // Peep is moving along X, so apply the offset to the X position of the destination and clamp their current Y
-            peep->DestinationX += offset;
+            peep->Destination.x += offset;
             const uint16_t centerLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
-            peep->DestinationY = std::clamp(
+            peep->Destination.y = std::clamp(
                 peep->y, static_cast<int16_t>(centerLine - 3), static_cast<int16_t>(centerLine + 3));
```
Gymnasiast, line 163: I think you can make everything int32_t here and do away with the casting.

ZehMatt:
I was thinking I should also convert more of {x, y} to CoordsXY otherwise I have to update the replays over and over.

src/openrct2/peep/GuestPathfinding.cpp
tupaschoal: Some of these places change how the game work if we change them to 32 bits, gotta be careful (I don't remember which ones though, just faced this once when I was migrating to CoordsXY(ZD))

ZehMatt:
I don't like the way I did it. I should have a getter and setter and leave the type alone for the time being.

ZehMatt:
I managed to avoid breaking replays by first implementing getters and setters, this makes the code a bit more readable. This is a good starting point for more refactors.

src/openrct2/peep/Staff.cpp
```diff
@@ -653,9 +653,7 @@ bool Staff::DoHandymanPathFinding()
     }
 
     PeepDirection = newDirection;
-    DestinationX = chosenTile.x + 16;
-    DestinationY = chosenTile.y + 16;
-    DestinationTolerance = 3;
+    SetDestination(chosenTile + CoordsXY{ 16, 16 }, 3);
```
duncanspumpkin, line 656: Can we make these toTileCentre?

src/openrct2/peep/Staff.cpp
```diff
@@ -1173,8 +1169,8 @@ void Staff::UpdateMowing()
             return;
         }
 
-        DestinationX = _MowingWaypoints[Var37].x + NextLoc.x;
-        DestinationY = _MowingWaypoints[Var37].y + NextLoc.y;
+        auto destination = _MowingWaypoints[Var37] + NextLoc;
+        SetDestination(destination);
```
duncanspumpkin, line 1173: Assume this will be tolerance 3 as that's what it was too get to the first waypoint. Maybe do a follow on pr that forces setting the tolerance.

src/openrct2/peep/Staff.cpp
Gymnasiast: It's not guaranteed to be the same, as NextLoc is a CoordsXY as well, IIRC.

duncanspumpkin: We should go through at some point and either enforce using only to tile centre or clearly mark why it's not correct in this situation. Guess we can leave for later though.

src/openrct2/peep/Staff.cpp
ZehMatt: I initially did that where default was zero however that broke the replays as it changes the hash so for testing this was better.

duncanspumpkin (approved):

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -154,24 +151,24 @@ static int32_t peep_move_one_tile(Direction direction, Peep* peep)
...
-        int8_t offset = (scenario_rand() & 7) - 3;
+        const int8_t offset = (scenario_rand() & 7) - 3;
         if (direction == 0 || direction == 2)
         {
             // Peep is moving along X, so apply the offset to the X position of the destination and clamp their current Y
-            peep->DestinationX += offset;
-            const uint16_t centerLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
-            peep->DestinationY = std::clamp(
-                peep->y, static_cast<int16_t>(centerLine - 3), static_cast<int16_t>(centerLine + 3));
+            const int32_t centerLine = (peep->y & 0xFFE0) + COORDS_XY_HALF_TILE;
```
Gymnasiast, line 158: `centreLine`

(We use British English throughout.)

src/openrct2/peep/Peep.cpp
```diff
@@ -606,7 +606,8 @@ std::optional<CoordsXY> Peep::UpdateAction(int16_t& xy_distance)
         Action = PeepActionType::None2;
     }
 
-    CoordsXY diffrenceLoc = { x - DestinationX, y - DestinationY };
+    CoordsXY diffrenceLoc = GetLocation();
```
Gymnasiast, line 609: `differenceLoc`

src/openrct2/peep/Peep.cpp
```diff
@@ -3381,3 +3375,20 @@ bool Peep::HasItem(ShopItem peepItem) const
...
+
+void Peep::SetDestination(const CoordsXY& coords, int32_t tolerance)
+{
+    SetDestination(coords);
+    DestinationTolerance = tolerance;
+}
+
+CoordsXY Peep::GetDestination() const
+{
+    return CoordsXY{ static_cast<int16_t>(DestinationX), static_cast<int16_t>(DestinationY) };
```
Gymnasiast, line 3393: Why did you add these casts?

src/openrct2/peep/Staff.cpp
```diff
@@ -1449,17 +1445,15 @@ void Staff::UpdateHeadingToInspect()
...
+        SetDestination({ destX, destY }, 2);
         sprite_direction = PeepDirection << 3;
 
         z = rideEntranceExitElement->base_height * 4;
         SubState = 4;
         // Falls through into SubState 4
     }
 
-    int16_t delta_y = abs(y - DestinationY);
+    int16_t delta_y = abs(y - GetDestination().y);
```
Gymnasiast, line 1456: Looks like you're using y directly here instead of through `GetLocation()`.

src/openrct2/peep/Staff.cpp
```diff
@@ -1449,17 +1445,15 @@ void Staff::UpdateHeadingToInspect()
         int32_t destX = NextLoc.x + 16 + DirectionOffsets[PeepDirection].x * 53;
```
Gymnasiast, line 1445: Should also be changed to use CoordsXY, but may be left for another PR (in which case I'd like you to create an issue for it.)

src/openrct2/world/Park.cpp
```diff
@@ -722,10 +722,8 @@ Peep* Park::GenerateGuest()
             peep->sprite_direction = direction << 3;
 
             // Get the centre point of the tile the peep is on
```
Gymnasiast, line 724: This comment is now stating the obvious, so should be removed.

src/openrct2/peep/Peep.cpp
ZehMatt: DestinationX and DestinationY are uint16_t, so explicit cast. Those fields will be eventually superseded by CoordsXY.

Gymnasiast: But uint16_t fits into int32_t, so why does it need to be cast?

ZehMatt: Signed vs unsigned. But I think DestinationX and DestinationY never use negative numbers in that regard, I'll eliminate the casts.

Gymnasiast (approved):

PR #14028, opened 2021-02-09
Add ccache action to GitHub Actions

janisozaur:
We already have ccache discovery in our CMake setup, so it should work out of the box.

I'm not sure which of the builds will use it properly, as there are some that are being ran in docker container. To verify this, the action is said to report ccache stats at the end.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1
       - name: Build OpenRCT2
         run: |
           sudo su
@@ -100,6 +102,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1
       - name: Build OpenRCT2
         run: |
           sudo su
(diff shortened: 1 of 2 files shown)


janisozaur:
Seems to work fine, the ccache stats reports successful hits. The odd thing is not all jobs are affected equally, but overall it's a win. We've seen such behaviour with cache in the past on travis as well.

tupaschoal (commented):
I'd try hitting "rerun all jobs" a few times to see how it fares. We've seen some intermittent failures on jobs every now and then that are tied to ccache not starting properly. It'd be great if those went away

janisozaur:
Makes sense.

Initial run: pass
Re-run 1: ongoing

janisozaur:
This has now been ran 5 times and none of those resulted in any kind of error. While doing the re-runs I've also figured out the reason for caches not getting hit: it was all gathered in one big cache that was probably overwritten by whichever job finished last. This is now fixed by use of individual keys for each of jobs and times are reduced as expected:

|                      | Time before | Time cached |
|----------------------|-------------|-------------|
| Linux AMD64 portable | 10m 56s     | 2m 27s      |
| Linux i686 portable  | 8m 9s       | 1m 47s      |
| Linux AMD64 AppImage | 7m 37s      | 2m 18s      |
| Linux clang          | 11m 7s      | 1m 40s      |
| Android              | 18m 46s     | 3m 56s      |

tupaschoal (approved):
The results look amazing, and very nice find about the keys. CMake code looks ok too

Issue #12069, opened 2020-06-28
Remove_unused_objects removes used objects


Liampie:
I've never encountered this before, but now I'm worried this bug has existed for a while without me noticing. When I use the remove_unused_objects command, it actually removes one object that is in use. There may be more, but I'm just noticing this one. Note the support object before and after:

![before](https://user-images.githubusercontent.com/19242088/85942853-5a207e80-b92c-11ea-95a1-998feb0b022d.png)
![after](https://user-images.githubusercontent.com/19242088/85942898-b5eb0780-b92c-11ea-9091-976698dfc6db.png)

The disappeared object is QTQPC_DP.dat. I tried placing a few more instances of the object before applying the command again, but it doesn't change anything.

edit: Can also deselect the object manually in the object selection window. Will try that in vanilla in a minute.
edti2: Can't get the savegame to open in the landscape editor in vanilla, so can't compare.
edit3: Removing unused scenery in parkdat leaves this object in!

Savegame:
[Tycoonman - Live The Life.zip](https://github.com/OpenRCT2/OpenRCT2/files/4842243/Tycoonman.-.Live.The.Life.zip)

Using "v0.2.6-1002-gbdbf12892 (bfdbf128 on develop)" on Windows 8.
Same problem when I go back to release.

Gymnasiast:
https://github.com/OpenRCT2/OpenRCT2/commit/6c3c857dfa5cd0d267b89a9d70930fbacdfbaea4 is a release commit, not a dev one. From your text, it seems you also tried a develop commit - which one?

Also: why did you delete the template?

IntelOrca:
I think this can happen if there are two objects with the same identifier in your park. Note that object files can have a different identifier from their actual filename.

Liampie:
> [6c3c857](https://github.com/OpenRCT2/OpenRCT2/commit/6c3c857dfa5cd0d267b89a9d70930fbacdfbaea4) is a release commit, not a dev one. From your text, it seems you also tried a develop commit - which one?

My bad, I only added the version after I was done testing between release/develop. Adding it to the first post - with some new findings and the savegame as well.

 > Also: why did you delete the template?

What do you mean?

Liampie:
Some new findings in the opening post. Parkdat removed unused scenery correctly!

Broxzier:
> What do you mean?

[...]

PR #14056, opened 2021-02-10
Fix #7772. Check for vehicles using an object when setting inuse

### LINKED ISSUE #7772: Vehicles hacked on different type of train become invisible when vehicle type disabled
**OS:** Windows 10
**Version:** 0.2.0
**Commit/Build:** 3f8c3c3

A minor bug i discussed with @Broxzier , he recommended me to make an issue so we don't forget about this later.

Anyway, i spawned some covered cars on the middle 2 vehicles of a train using Visual Studio.
When I submitted the park to NE, a helpful mod disabled all unused objects, including the covered cars, because they were not marked as being used, while technically they were in the game.

image

The result was that the middle cars of the train turned invisible.

duncanspumpkin:
I think https://github.com/OpenRCT2/OpenRCT2/issues/12069 is also fixed by this. Hard to fully tell.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -24,6 +24,7 @@
 - Change: [#13346] [Plugin] Renamed FootpathScenery to FootpathAddition, fix typos.
 - Change: [#13857] Change Rotation Control Toggle to track element number 256
 - Fix: [#4605, #11912] Water palettes are not updated properly when selected in Object Selection.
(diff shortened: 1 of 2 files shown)


Gymnasiast:
> I think #12069 is also fixed by this. Hard to fully tell.

Doubt it. #12069 is about a piece of scenery getting deselected.

duncanspumpkin:
Oh yeah my bad. Didn't fully read the issue.

IntelOrca (approved):

PR #14059, opened 2021-02-10
[Plugin] Add optional filter to custom tools

IntelOrca:
* Invalidate viewport when changing the location (without scrolling)
* Add filter to custom tools so that plugins can adjust the hit testing logic.
* Improve filter logic, reverse the logic (mask is now what you want to be enabled for hit testing).

I found the way the hit testing flags worked quite confusing, so I have changed it.

### DIFF
--- a/src/openrct2-ui/scripting/CustomMenu.cpp
+++ b/src/openrct2-ui/scripting/CustomMenu.cpp
@@ -27,6 +27,21 @@ namespace OpenRCT2::Scripting
         "volcano_down",  "walk_down",  "paint_down",    "entrance_down", "hand_open",  "hand_closed",
     };
 
+    static const DukEnumMap<ViewportInteractionItem> ToolFilterMap({
+        { "terrain", ViewportInteractionItem::Terrain },
+        { "entity", ViewportInteractionItem::Entity },
+        { "ride", ViewportInteractionItem::Ride },
+        { "water", ViewportInteractionItem::Water },
+        { "scenery", ViewportInteractionItem::Scenery },
+        { "footpath", ViewportInteractionItem::Footpath },
+        { "footpath_item", ViewportInteractionItem::FootpathItem },
+        { "park_entrance", ViewportInteractionItem::ParkEntrance },
+        { "wall", ViewportInteractionItem::Wall },
+        { "large_scenery", ViewportInteractionItem::LargeScenery },
+        { "label", ViewportInteractionItem::Label },
+        { "banner", ViewportInteractionItem::Banner },
+    });
+
     template<> DukValue ToDuk(duk_context* ctx, const CursorID& cursorId)
     {
         auto value = EnumValue(cursorId);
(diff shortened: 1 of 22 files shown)


src/openrct2/interface/Viewport.h
```diff
@@ -68,18 +68,18 @@ enum class ViewportInteractionItem : uint8_t
...
-    VIEWPORT_INTERACTION_MASK_WATER = ~(1 << (EnumValue(ViewportInteractionItem::Water) - 1)),
-    VIEWPORT_INTERACTION_MASK_SCENERY = ~(1 << (EnumValue(ViewportInteractionItem::Scenery) - 1)),
-    VIEWPORT_INTERACTION_MASK_FOOTPATH = ~(1 << (EnumValue(ViewportInteractionItem::Footpath) - 1)),
-    VIEWPORT_INTERACTION_MASK_FOOTPATH_ITEM = ~(1 << (EnumValue(ViewportInteractionItem::FootpathItem) - 1)),
-    VIEWPORT_INTERACTION_MASK_PARK_ENTRANCE = ~(1 << (EnumValue(ViewportInteractionItem::ParkEntrance) - 1)),
-    VIEWPORT_INTERACTION_MASK_WALL = ~(1 << (EnumValue(ViewportInteractionItem::Wall) - 1)),
-    VIEWPORT_INTERACTION_MASK_LARGE_SCENERY = ~(1 << (EnumValue(ViewportInteractionItem::LargeScenery) - 1)),
-    VIEWPORT_INTERACTION_MASK_BANNER = ~(1 << (EnumValue(ViewportInteractionItem::Banner) - 2)), // Note the -2 for BANNER
+    VIEWPORT_INTERACTION_MASK_ALL = 0xFFFF,
+    VIEWPORT_INTERACTION_MASK_TERRAIN = 1 << EnumValue(ViewportInteractionItem::Terrain),
```
duncanspumpkin, line 72: If we are making this change why don't we just remove the whole thing and use EnumsToFlags

src/openrct2/interface/Viewport.cpp
```diff
@@ -1291,23 +1296,16 @@ void viewport_set_visibility(uint8_t mode)
...
-        mask = 1 << (EnumValue(ps->sprite_type) - 3);
-    else
-        mask = 1 << (EnumValue(ps->sprite_type) - 1);
-
-    if (filter & mask)
+    if (ps->sprite_type != ViewportInteractionItem::None && ps->sprite_type != ViewportInteractionItem::Label
+        && ps->sprite_type <= ViewportInteractionItem::Banner)
     {
-        return false;
+        auto mask = 1 << EnumValue(ps->sprite_type);
```
duncanspumpkin, line 1302: You can use enum flag here as well

src/openrct2-ui/scripting/CustomMenu.cpp
```diff
@@ -172,6 +185,26 @@ namespace OpenRCT2::Scripting
...
+                if (dukFilter.is_array())
+                {
+                    customTool.Filter = 0;
+                    auto dukItems = dukFilter.as_array();
+                    for (const auto& dukItem : dukItems)
+                    {
+                        if (dukItem.type() == DukValue::Type::STRING)
+                        {
+                            auto value = ToolFilterMap[dukItem.as_string()];
+                            customTool.Filter |= 1 << static_cast<uint32_t>(value);
```
duncanspumpkin, line 199: Can use enum flag here

duncanspumpkin (approved):

PR #14062, opened 2021-02-10
Implement SRV dns resolving

runelaenen:
Just like minecraft, OpenRCT2 should resolve SRV records to allow servers to run on non-standard ports.

I'm not a C/C++ developer, but have been able to create a POC that works on my machine (Macbook Pro 2019, Catalina, Intel i7) but I will need some help to get this to the code standards OpenRCT2 needs.
Any help is appreciated 🙏

Some context about srv records: https://kb.porkbun.com/article/148-how-to-connect-your-domain-to-minecraft-using-srv-records
The SRV-resolving code has been based on https://oliver-kaestner.de/english-c-query-srv-dns-record-with-example/

### DIFF
--- a/OpenRCT2.xcodeproj/project.pbxproj
+++ b/OpenRCT2.xcodeproj/project.pbxproj
@@ -1332,8 +1332,6 @@
 		9308D9FD209908090079EE96 /* Surface.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = Surface.h; sourceTree = "<group>"; };
 		930EEA6924FC00940070314E /* ScenarioSelect.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = ScenarioSelect.cpp; sourceTree = "<group>"; };
 		9329D51F240C17C60054301C /* BenchUpdate.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = BenchUpdate.cpp; sourceTree = "<group>"; };
-		932A20CD22D73CEE00C57EDB /* GuestSetNameAction.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = GuestSetNameAction.hpp; sourceTree = "<group>"; };
-		932A20CE22D73CEE00C57EDB /* RideSetVehiclesAction.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = RideSetVehiclesAction.hpp; sourceTree = "<group>"; };
 		932A20CF22D73CEE00C57EDB /* GameActionCompat.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = GameActionCompat.cpp; sourceTree = "<group>"; };
 		932A20D322D73CEF00C57EDB /* GameActionRegistration.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = GameActionRegistration.cpp; sourceTree = "<group>"; };
 		932A20F522D73CF300C57EDB /* GameAction.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = GameAction.h; sourceTree = "<group>"; };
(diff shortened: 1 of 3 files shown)


OpenRCT2.xcodeproj/project.pbxproj
```diff
@@ -1332,8 +1332,6 @@
 		9308D9FD209908090079EE96 /* Surface.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = Surface.h; sourceTree = "<group>"; };
 		930EEA6924FC00940070314E /* ScenarioSelect.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = ScenarioSelect.cpp; sourceTree = "<group>"; };
 		9329D51F240C17C60054301C /* BenchUpdate.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = BenchUpdate.cpp; sourceTree = "<group>"; };
-		932A20CD22D73CEE00C57EDB /* GuestSetNameAction.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = GuestSetNameAction.hpp; sourceTree = "<group>"; };
-		932A20CE22D73CEE00C57EDB /* RideSetVehiclesAction.hpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.h; path = RideSetVehiclesAction.hpp; sourceTree = "<group>"; };
```
tupaschoal, line 1336: Was this intended? Maybe you need to rebase somehow?

src/openrct2/network/Socket.h
```diff
@@ -33,6 +33,13 @@ enum class NetworkReadPacket : int32_t
     Disconnected
 };
 
+typedef struct SRVRecord {
+    u_int16_t priority;
+    u_int16_t weight;
+    u_int16_t port;
+    char dname[255];
+} SRVRecord;
```
tupaschoal, line 41: This seems C-like, we can probably use something a little more C++-zy.
We also use `uint16_y` instead of `u_int16_t`
```suggestion
struct SRVRecord {
    uint16_t priority;
    uint16_t weight;
    uint16_t port;
    char dname[255];
};
```

src/openrct2/network/Socket.cpp
```diff
@@ -245,6 +255,45 @@ class Socket
             return true;
         }
     }
+    
+    /*
+     * https://oliver-kaestner.de/english-c-query-srv-dns-record-with-example/
+     */
+    static bool ResolveSRV(const char* host, SRVRecord* resolved) {
+        struct __res_state res;
```
tupaschoal, line 263: I guess we would only need the `struct` if this was C code.
```suggestion
        __res_state res;
```

src/openrct2/network/Socket.cpp
```diff
@@ -245,6 +255,45 @@ class Socket
...
+        if (len < 0) {
+            return false;
+        }
+     
+        ns_msg handle;
+        ns_rr rr;
+     
+        ns_initparse(answer, len, &handle);
+             
+        for (int i = 0; i < ns_msg_count(handle, ns_s_an); i++) {
```
tupaschoal, line 279:
```suggestion
        for (int i = 0; i < ns_msg_count(handle, ns_s_an); ++i) {
```

tupaschoal (commented):
Seems like the coding style check is failing, you need to get it sorted out so that the other checks run. You can just click there and see the log (and sort out manually) or you can run the formatter itself inside the code to get it all done.

I know nothing about SRV, so I mostly did C++ suggestions. Other than that, you'll need to add the `lresolv` to other places for the windows and unix builds to work too.

src/openrct2/network/Socket.cpp
Gymnasiast: Why? We use i++ everywhere.

tupaschoal: It can be potentially faster, as it avoids one copy

Gymnasiast: I very much doubt that. It strikes me as something that compilers will take care of.

tupaschoal: https://stackoverflow.com/questions/24901/is-there-a-performance-difference-between-i-and-i-in-c

Gymnasiast: Let's go with your suggestion for now, then. But I would like to see it raised in the chat or in an issue of its own and applied throughout the codebase.

duncanspumpkin: I don't think any modern compiler would make a difference for this

ZehMatt: It sort of depends on a few things but yeah most compilers will figure this out if the type is a primitive

ZehMatt:
The code you are using only works on Linux I'm afraid. Windows 10 provides windns.h but I'm not sure there is anything prior to that. In order for us to accept your PR this has to work on all platforms that have networking enabled.

runelaenen:
I'm working on MacOS myself, but I'll try to find a working Windows computer to test these changes :+1:

Gymnasiast:
@runelaenen Are you still working on this?

PR #14064, opened 2021-02-10
Fix #14058: Ride entry or exit can not be built in multiplayer

### LINKED ISSUE #14058: Ride entrances and exits can't be placed in a multiplayer game
**OS:** Windows 10
**Version:** 0.3.2 develop
**Commit/Build:** f99c35d

In Multiplayer it is not possible to place ride entrances and exits. 
No matter if you're the host of the server or joined the server.
Instead an empty red alert box appears and nothing happens.
If a client tries to place entrances/exits, then the host can see them.
But the client has to reconnect in order to see them.
If the host tries to place them, only the error appears and nothing else.

- [ x] Specific to multiplayer?

**Steps to reproduce:**
1.  Host a network game (doesn't matter if people join or not)
2.  Build any attraction and try to place an entrance or an exit

ZehMatt:
Not entirely sure how only multiplayer was affected. I messed up the logic, it can only check for ghost if the flag is present, for entrances it expects it to just remove whatever entrance/exit was there.

### DIFF
--- a/src/openrct2/actions/LargeSceneryRemoveAction.cpp
+++ b/src/openrct2/actions/LargeSceneryRemoveAction.cpp
@@ -191,10 +191,11 @@ GameActions::Result::Ptr LargeSceneryRemoveAction::Execute() const
 
 TileElement* LargeSceneryRemoveAction::FindLargeSceneryElement(const CoordsXYZ& pos, int32_t sequenceIndex) const
 {
+    const bool isGhost = GetFlags() & GAME_COMMAND_FLAG_GHOST;
     for (auto* sceneryElement : TileElementsView<LargeSceneryElement>(pos))
     {
         // If we are removing ghost elements
-        if ((GetFlags() & GAME_COMMAND_FLAG_GHOST) && sceneryElement->IsGhost() == false)
+        if (isGhost && sceneryElement->IsGhost() == false)
             continue;
 
         if (sceneryElement->GetDirection() != _loc.direction)
--- a/src/openrct2/actions/RideEntranceExitRemoveAction.cpp
+++ b/src/openrct2/actions/RideEntranceExitRemoveAction.cpp
@@ -46,11 +46,13 @@ void RideEntranceExitRemoveAction::Serialise(DataSerialiser& stream)
 }
 
 static TileElement* FindEntranceElement(
-    const CoordsXY& loc, ride_id_t rideIndex, int32_t stationNum, int32_t entranceType, bool ghost)
+    const CoordsXY& loc, ride_id_t rideIndex, int32_t stationNum, int32_t entranceType, uint32_t flags)
 {
+    const bool isGhost = flags & GAME_COMMAND_FLAG_GHOST;
     for (auto* entranceElement : TileElementsView<EntranceElement>(loc))
     {
-        if (entranceElement->IsGhost() != ghost)
+        // If we are removing ghost elements
+        if (isGhost && entranceElement->IsGhost() == false)
             continue;
 
         if (entranceElement->GetRideIndex() != rideIndex)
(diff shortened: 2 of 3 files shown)


src/openrct2/actions/LargeSceneryRemoveAction.cpp
```diff
@@ -191,21 +191,18 @@ GameActions::Result::Ptr LargeSceneryRemoveAction::Execute() const
...
 TileElement* LargeSceneryRemoveAction::FindLargeSceneryElement(const CoordsXYZ& pos, int32_t sequenceIndex) const
 {
+    const bool isGhost = GetFlags() & GAME_COMMAND_FLAG_GHOST;
     for (auto* sceneryElement : TileElementsView<LargeSceneryElement>(pos))
     {
         // If we are removing ghost elements
-        if ((GetFlags() & GAME_COMMAND_FLAG_GHOST) && sceneryElement->IsGhost() == false)
+        if (isGhost && sceneryElement->IsGhost() == false)
             continue;
-
```
Gymnasiast, line 199: Please keep this blank lines, they help readability.

src/openrct2/actions/RideEntranceExitRemoveAction.cpp
```diff
@@ -46,22 +46,20 @@ void RideEntranceExitRemoveAction::Serialise(DataSerialiser& stream)
 }
 
 static TileElement* FindEntranceElement(
-    const CoordsXY& loc, ride_id_t rideIndex, int32_t stationNum, int32_t entranceType, bool ghost)
+    const CoordsXY& loc, ride_id_t rideIndex, int32_t stationNum, int32_t entranceType, uint32_t flags)
```
Gymnasiast, line 49: Is this change necessary?

Gymnasiast:
> Not entirely sure how only multiplayer was affected

Ghosts are not synced in multiplayer, so I expect it has to do with that.

src/openrct2/actions/RideEntranceExitRemoveAction.cpp
ZehMatt: I wanted to get this a bit more consistent with the others so I opted to use the flags.

src/openrct2/actions/LargeSceneryRemoveAction.cpp
ZehMatt: Its less readable the other way imho.

duncanspumpkin: I agree with @Gymnasiast and it's the general style used in the rest of the codebase for ifs without curly brackets

ZehMatt: The consistency part is actually not true, see:
https://github.com/OpenRCT2/OpenRCT2/blob/14cca95fecdb48f88e51e2f372eed5bf545ec022/src/openrct2/actions/BannerRemoveAction.cpp#L128
https://github.com/OpenRCT2/OpenRCT2/blob/14cca95fecdb48f88e51e2f372eed5bf545ec022/src/openrct2/actions/FootpathPlaceAction.cpp#L439
https://github.com/OpenRCT2/OpenRCT2/blob/14cca95fecdb48f88e51e2f372eed5bf545ec022/src/openrct2/actions/LandSetHeightAction.cpp#L254
etc..

Its inconsistent, thats the whole reason why I removed those empty lines, but I added them back.

duncanspumpkin (approved):

src/openrct2/actions/RideEntranceExitRemoveAction.cpp
Gymnasiast: Which others? 

If it does not need to know about flags, then I don't think it should receive them. If anything, the others should be modified if they never use any other flags besides `GAME_COMMAND_FLAG_GHOST`.

But I don't want to keep this PR from being merged just because of this, so this is fine for now.

Gymnasiast:
Please keep in mind to put any merged PRs on the milestone.

PR #14081, opened 2021-02-12
Refactor IStream::ReadArray to return unique_ptr

ZehMatt:
Fixes also a memory leak, small but no one released it.


duncanspumpkin (approved):

duncanspumpkin:
Would be good if @IntelOrca could have a quick look at this.

IntelOrca (approved):

PR #14091, opened 2021-02-13
Change data passed to UpdateTrackMotionMiniGolfVehicle

janisozaur:
The function now is passed the only flag it uses

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -8725,7 +8725,7 @@ loc_6DBA33:;
  *
  *
  */
-void Vehicle::UpdateTrackMotionMiniGolfVehicle(Ride* curRide, rct_ride_entry* rideEntry, rct_ride_entry_vehicle* vehicleEntry)
+void Vehicle::UpdateTrackMotionMiniGolfVehicle(Ride* curRide, bool vehicleHasFlag25, rct_ride_entry_vehicle* vehicleEntry)
 {
     uint16_t otherVehicleIndex = SPRITE_INDEX_NULL;
     TileElement* tileElement = nullptr;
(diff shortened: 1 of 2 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -9276,7 +9276,8 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
 
     for (Vehicle* vehicle = _vehicleFrontVehicle; vehicle != nullptr;)
     {
-        vehicle->UpdateTrackMotionMiniGolfVehicle(curRide, rideEntry, vehicleEntry);
+        bool vehicleHasFlag25 = rideEntry->vehicles[0].flags & VEHICLE_ENTRY_FLAG_25;
+        vehicle->UpdateTrackMotionMiniGolfVehicle(curRide, vehicleHasFlag25, vehicleEntry);
```
janisozaur, line 9280:
```suggestion
        bool vehicleHasBoatHireCollisionDetection = vehicleEntry->flags & VEHICLE_ENTRY_FLAG_BOAT_HIRE_COLLISION_DETECTION;
        vehicle->UpdateTrackMotionMiniGolfVehicle(curRide, vehicleHasFlag25, vehicleHasBoatHireCollisionDetection);
```

tupaschoal (commented):
This seems ok. Can we take the opportunity to rename flag 25 to something more meaningful?

janisozaur:
Something like `RESET_SWING_STATE`? `RESET_GOLF_CLUB_SWING_STATE`?

tupaschoal:
I like the latter!

Gymnasiast:
@IntelOrca I didn’t review it because Tulio and Jani were still discussing changes. But I brought up this PR in \#team some time ago, and I remember someone (perhaps @duncanspumpkin ?) being opposed to it. I’m neutral towards it myself, by the way.

IntelOrca:
Well it probably makes more sense, just to get the ride entry again inside the function, rather than pass it as a second parameter when `ride` is already passed.

duncanspumpkin:
Yeah I'm not a fan of the change as is. We ideally want the golf and non golf code to merge as much as possible and this will just make that harder to achieve.

Gymnasiast:
In that case, I’ll close it.

PR #14103, opened 2021-02-15
Randomize ride inspections

manatails:
Currently regular inspection for multi-station ride is done only on Station 1.
This isn't very practical or realistic, so I propose randomizing the station chosen for inspection every time so that each station gets more uniform number of inspections.

Following is a monorail with 4 stations where each station is assigned to a mechanic.

2021-02-16 02_44_41-OpenRCT2

Before the patch, station 1 got 12 inspections while other stations got zero inspections, mechanic 2 hasn't done anything at all

2021-02-16 02_48_21-OpenRCT2

After the patch, each station gets more inspections and the mechanics assigned to them are actually doing their job
 

2021-02-16 02_55_51-OpenRCT2

I also added a label that is enabled when there is more than 1 station which notifies the user about the station that the mechanic is set to visit during inspection/repair. This can also help players proactively position mechanics for minimizing downtime.

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3665,6 +3665,7 @@ STR_6408    :Please install “innoextract” to extract GOG Installer, then res
 STR_6409    :The selected file is not the offline GOG Installer for RollerCoaster Tycoon 2. You may have downloaded the GOG Galaxy downloader stub or selected the wrong file.
 STR_6410    :Zoom in/out
 STR_6411    :Show buttons for zooming in and out in the toolbar
+STR_6412    :{WINDOW_COLOUR_2}Station due for inspection: {BLACK}Station {COMMA16}
 
 #############
 # Scenarios #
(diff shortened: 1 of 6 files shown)


duncanspumpkin:
Its a nice idea but it would break a lot of peoples parks. They would suddenly find their optimised pathing is no longer valid.

Gymnasiast:
> Its a nice idea but it would break a lot of peoples parks. They would suddenly find their optimised pathing is no longer valid.

That, and pathfinding is already a can of worms.

Because of this, I'm afraid I won't merge it. Perhaps if we ever isolate the pathfinder and allow swapping them out in the way that OpenTTD does, we could be more experimental.

IntelOrca:
> Its a nice idea but it would break a lot of peoples parks.

Break how? Each mechanic still requires some access to every station as breakdowns are station specific aren't they?

ocalhoun6:
I think some multi-station rides don't have any station-specific breakdown modes.

The log flume and miniature railroad come to mind. The only station-specific breakdown modes are restraints stuck open/closed, right? But since those rides don't have those breakdown modes, you might have built a park where mechanics have access to station 1 of your railroad, but not to any of the other stations.

PR #14104, opened 2021-02-15
Add stubbed Vehicle test

janisozaur:
Allows for testing `Vehicle.cpp` code in isolation

### DIFF
--- a/test/tests/CMakeLists.txt
+++ b/test/tests/CMakeLists.txt
@@ -266,3 +266,55 @@ SET_CHECK_CXX_FLAGS(test_s6importexporttests)
 target_link_libraries(test_s6importexporttests ${GTEST_LIBRARIES} libopenrct2 ${LDL} z)
 target_link_platform_libraries(test_s6importexporttests)
 add_test(NAME s6importexporttests COMMAND test_s6importexporttests)
+
+# Vehicle test
+set(VEHICLE_TEST_SOURCES
+        "${CMAKE_CURRENT_LIST_DIR}/VehicleTest.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/Vehicle.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/VehicleSubpositionData.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/VehicleData.cpp"
+        "${ROOT_DIR}/src/openrct2/world/Sprite.cpp"
+        "${ROOT_DIR}/src/openrct2/world/TileElementBase.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/CableLift.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/RideData.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/TrackData.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/Track.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/Station.cpp"
+        "${ROOT_DIR}/src/openrct2/windows/Intent.cpp"
+        "${ROOT_DIR}/src/openrct2/core/Crypt.OpenSSL.cpp"
+        #"${ROOT_DIR}/src/openrct2/audio/Audio.cpp"
+        #"${ROOT_DIR}/src/openrct2/interface/Viewport.cpp"
+        #"${ROOT_DIR}/src/openrct2/ride/Ride.cpp"
+        #"${ROOT_DIR}/src/openrct2/ride/RideData.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Balloon.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Banner.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Climate.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Duck.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Entrance.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Footpath.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Fountain.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/LargeScenery.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/Map.cpp"
+        #"${ROOT_DIR}/src/openrct2/world/MapAnimation.cpp"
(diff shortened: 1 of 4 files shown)


test/tests/CMakeLists.txt
```diff
@@ -266,3 +266,55 @@ SET_CHECK_CXX_FLAGS(test_s6importexporttests)
...
+        "${ROOT_DIR}/src/openrct2/world/Sprite.cpp"
+        "${ROOT_DIR}/src/openrct2/world/TileElementBase.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/CableLift.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/RideData.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/TrackData.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/Track.cpp"
+        "${ROOT_DIR}/src/openrct2/ride/Station.cpp"
+        "${ROOT_DIR}/src/openrct2/windows/Intent.cpp"
+        "${ROOT_DIR}/src/openrct2/core/Crypt.OpenSSL.cpp"
+        #"${ROOT_DIR}/src/openrct2/audio/Audio.cpp"
```
janisozaur, line 285: These were temporaries I tried using initially, to figure out how far can I reduce the necessary data

test/tests/VehicleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+    { 16, 0 },
+    { 16, 16 },
+    { 64, 64 },
+    { 64, -32 },
+    { -32, -32 },
+    { -32, 64 },
+};
+// clang-format on
+
+static Vehicle* vehicle_create_car(
```
janisozaur, line 981: Copied verbatim from `Ride.cpp`, but including that file in the new target pulls a whole lot of other dependencies. I haven't really paid much attention to what it does, might be simplified

test/tests/VehicleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+
+int32_t Guest::GetEasterEggNameId() const
+{
+    // TODO(janisozaur): do something here
+    return -1;
+}
+
+TEST(Vehicle, Create)
+{
+    [[maybe_unused]] Vehicle v{};
```
janisozaur, line 938: Dummy test to show a Vehicle can be created

test/tests/VehicleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+}
+
+TEST(Vehicle, SimpleUpdate)
+{
+    Vehicle v{};
+    v.Update();
```
janisozaur, line 944: ... and updated

test/tests/VehicleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+}
+
+TEST(Vehicle, CableLiftUpdate)
+{
+    reset_sprite_list();
+    Ride& r = _rideList[0];
+    r = Ride{};
+    r.id = 0;
+    Vehicle* v = cable_lift_segment_create(r, 0, 0, 0, 0, 0, 0, true);
```
janisozaur, line 953: Semi-random cable lift test, as it required pulling in a new file into the target

test/tests/VehicleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+TEST(Vehicle, VehicleCreateSimple)
+{
+    reset_sprite_list();
+    Ride& r = _rideList[0];
+    r = Ride{};
+    r.id = 0;
+    r.type = RIDE_TYPE_MONORAIL;
+    int32_t rd{};
+    TrackElement te{};
+    vehicle_create_car(0, 0, 0, 0, {}, &rd, &te);
```
janisozaur, line 1187: There is some minimal configuration necessary to call this function, otherwise it could segfault

IntelOrca (changes requested):
If you want to add tests for vehicle, please find a way that doesn't involve adding `#ifdef` to the main sources.

The tests should just link libopenrct2, I can't see any value in trying to compile sources stand alone from the rest of the project.

test/tests/VehicleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+    r.lifecycle_flags = RIDE_LIFECYCLE_PASS_STATION_NO_STOPPING;
+    int32_t rd{};
+    TrackElement te{};
+    Vehicle* v = vehicle_create_car(0, 0, 0, 0, {}, &rd, &te);
+    v->status = Vehicle::Status::Arriving;
+    v->Update();
+    EXPECT_EQ(_goto_callsites.size(), 1);
+    std::vector<std::pair<std::string, int>> expected_results;
+    expected_results.push_back(std::make_pair("loc_6D8E36", 1));
+    EXPECT_EQ(_goto_callsites, expected_results);
```
janisozaur, line 1217: I find this the most exciting part of this whole PR. I've added a temporary call to a function `hit_goto_callsite` provided just above this test, which I can use to ensure all the callsites I'm interested in are hit. Once I'm happy with the quality of the tests, the next step would be gathering checksums of all affected objects after the test is done and removing the calls to `hit_goto_callsite`

janisozaur:
@IntelOrca the reason to not bother with libopenrct2 is the hardcoding of loading of various files that I don't want to bother with:

https://github.com/OpenRCT2/OpenRCT2/blob/c98746c910eea2c159837e939ccee20be6db1a90/src/openrct2/ride/Ride.cpp#L217-L229

I just want to test various parts of `Vehicle::Update`

IntelOrca:
Source files are not expected to be compiled / linked in isolation from the other source files. That is what the directories themselves are for; libsawyer, libopenrct2, libopenrct2g, libopenrct2ui, openrct2 etc. Each one of those would have a separate unit test folder linking that lib.

janisozaur:
If we had those, that would be preferred solution, yes. But the only actual split we have so far is `libopenrct2` and its clients; pretty much an all-or-nothing type of deal.

IntelOrca:
@janisozaur I am currently looking at libsawyer. libopenrct2g is something I would like but haven't got round to yet. The unit tests should just link with libopenrct2 for the time being, once they are split up, the unit test can be changed to link with the smaller lib.

janisozaur:
Sounds nice. Do you have any ETA on that or progress track/design somewhere?

IntelOrca:
@janisozaur at the moment it is just a private repo, still trying to figure out how to get cmake find_package working.

rik-smeets:
I'm closing this PR due to inactivity, feel free to reopen if you like to continue working on it.

PR #14107, opened 2021-02-16
Add plugin hook for guest generation

### LINKED ISSUE #14105: Add hook for new guest generation

duncanspumpkin:
Close #14105

### DIFF
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -31,6 +31,8 @@
 #include "../ride/Station.h"
 #include "../ride/Track.h"
 #include "../scenario/Scenario.h"
+#include "../scripting/HookEngine.h"
+#include "../scripting/ScriptEngine.h"
 #include "../sprites.h"
 #include "../util/Util.h"
 #include "../windows/Intent.h"
@@ -1766,6 +1768,22 @@ Peep* Peep::Generate(const CoordsXYZ& coords)
     peep->EnergyTarget = energy;
 
     increment_guests_heading_for_park();
+
+#ifdef ENABLE_SCRIPTING
+    auto& hookEngine = OpenRCT2::GetContext()->GetScriptEngine().GetHookEngine();
+    if (hookEngine.HasSubscriptions(OpenRCT2::Scripting::HOOK_TYPE::GUEST_GENERATION))
+    {
+        auto ctx = OpenRCT2::GetContext()->GetScriptEngine().GetContext();
+
+        // Create event args object
+        auto obj = OpenRCT2::Scripting::DukObject(ctx);
+        obj.Set("id", peep->sprite_index);
+
+        // Call the subscriptions
+        auto e = obj.Take();
+        hookEngine.Call(OpenRCT2::Scripting::HOOK_TYPE::GUEST_GENERATION, e, true);
+    }
+#endif
 
     return peep;
 }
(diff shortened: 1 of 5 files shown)


src/openrct2/peep/Peep.cpp
```diff
@@ -1785,6 +1787,22 @@ Peep* Peep::Generate(const CoordsXYZ& coords)
...
+    {
+        auto ctx = OpenRCT2::GetContext()->GetScriptEngine().GetContext();
+
+        // Create event args object
+        auto obj = OpenRCT2::Scripting::DukObject(ctx);
+        obj.Set("id", peep->sprite_index);
+
+        // Call the subscriptions
+        auto e = obj.Take();
+        hookEngine.Call(OpenRCT2::Scripting::HOOK_TYPE::GUEST_GENERATION, e, true);
```
duncanspumpkin, line 1802: guess this can be left as false as you don't need to modify the guest

IntelOrca: Game state is mutable because this is a non-client specific event.

src/openrct2/peep/Peep.cpp
```diff
@@ -1785,6 +1787,22 @@ Peep* Peep::Generate(const CoordsXYZ& coords)
...
 
+#ifdef ENABLE_SCRIPTING
+    auto& hookEngine = OpenRCT2::GetContext()->GetScriptEngine().GetHookEngine();
+    if (hookEngine.HasSubscriptions(OpenRCT2::Scripting::HOOK_TYPE::GUEST_GENERATION))
+    {
+        auto ctx = OpenRCT2::GetContext()->GetScriptEngine().GetContext();
+
+        // Create event args object
+        auto obj = OpenRCT2::Scripting::DukObject(ctx);
+        obj.Set("id", peep->sprite_index);
```
IntelOrca, line 1798: Should we add a `handled` bool so that plugins can set this to true if they do not want the guest to spawn?

ZehMatt: The time it calls hook is a bit late here but I agree that the hook should control the spawn.

duncanspumpkin: I don't think we should allow control at this point. A hook earlier when deciding to generate would be better for that. Then they can up the spawn rate or adjust it to suit. I don't want to move the hook to the start of the function to prevent consumers changes being overwritten.

ZehMatt: So what exactly is the point of this hook then? You could in theory just subscribe to tick and remember the last guest count if you wanna know how many new guests entered. Perhaps name some example on how this hook could be used?

duncanspumpkin: It gives you the entity id of the new guest. You can then do with that as you wish. There is currently a plug-in that iterates all entities every tick looking for a new entity so that it can assign a name based on twitch chat users. With this they just use the hook to assign the name.

IntelOrca: If there is another way to control guest generation then that's fine.

duncanspumpkin: You can already use the cheat game action to force spawn new guests. We don't at present expose anything else. ZehMatt suggested exposing a model that plugins can modify. That would improve performance over a hook as deciding to spawn a guest would be a function that is called relatively often compared to this hook where generation has already been decided.

IntelOrca (approved):

IntelOrca (changes requested):
Oh, you should probably update openrct2.d.ts and the API version.

Gymnasiast:
@duncanspumpkin Could you apply @IntelOrca 's requests so this can be merged?

tupaschoal (approved):

PR #14110, opened 2021-02-16
Add support for 5-button mice

robopilot99:
Allows rotating the camera with buttons 4 and 5 on 5-button-mice

Adds basic support for 5-button mice. Some high-end gaming and productivity mice have an additional 2 buttons, usually positioned underneath the thumb. These are usually used to provide easy access to additional actions like navigating forward and backward in a web browser, or undo and redo in a text editor. I have programed them so that they rotate the camera, both during normal input and during a right-click drag.

### DIFF
--- a/src/openrct2-ui/UiContext.cpp
+++ b/src/openrct2-ui/UiContext.cpp
@@ -413,6 +413,12 @@ public:
                             _cursorState.right = CURSOR_PRESSED;
                             _cursorState.old = 2;
                             break;
+                        case SDL_BUTTON_X1:
+                            StoreMouseInput(MouseState::FourPress, mousePos);
+                            break;
+                        case SDL_BUTTON_X2:
+                            StoreMouseInput(MouseState::FivePress, mousePos);
+                            break;
                     }
                     _cursorState.touch = false;
                     break;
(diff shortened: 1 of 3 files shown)


IntelOrca:
@robopilot99 thanks for your pull request. Unfortunately I am almost about to merge a new shortcut engine which supports additional mouse buttons thus allowing you bind them to anything you like.

It would be great if you could test my branch to see if it achieves the same behaviour as you have here:
https://github.com/OpenRCT2/OpenRCT2/pull/13848

robopilot99:
Briefly checked out your shortcut engine. I agree that it does completely replicate the functionality of this pull request with the slight caveats that it is not possible to use both the keyboard shortcuts and mouse buttons at the same time, and that the mappings aren't configured by default.
I see little reason not to go ahead and close this pull request as redundant

IntelOrca:
> with the slight caveats that it is not possible to use both the keyboard shortcuts and mouse buttons at the same time

You can, but not via the UI. You need to edit the JSON file directly.

> mappings aren't configured by default

I can change that, although I wasn't sure whether it should be zoom or rotation.

ocalhoun6:
> I can change that, although I wasn't sure whether it should be zoom or rotation.

Mouse wheel already does zoom, so I don't think you need mouse buttons to also control zoom.

Gymnasiast:
Closing this PR since https://github.com/OpenRCT2/OpenRCT2/pull/13848 already handles most of this. I'll make a note in that PR to add mouse buttons 4 and 5 for rotation.

IntelOrca:
> I'll make a note in that PR to add mouse buttons 4 and 5 for rotation.

I had already done that from the start.

Gymnasiast:
> > I'll make a note in that PR to add mouse buttons 4 and 5 for rotation.
> 
> I had already done that from the start.

Are they bound *by default*?

IntelOrca:
@Gymnasiast yes

jackson-cognosos:
They weren't already mapped in the version I tested

IntelOrca:
@jackson-cognosos It is probably because it imported your old hotkeys.dat file which wouldn't of had the binding.

jackson-cognosos:
@IntelOrca Yeah, I had an old config floating around. That is the likely cause

PR #14121, opened 2021-02-19
Refactor Tile Inspector

ZehMatt:


src/openrct2/world/TileInspector.cpp
```diff
@@ -30,201 +30,201 @@
...
 
-        tile_element_remove(tileElement);
-        map_invalidate_tile_full(loc);
+            tile_element_remove(tileElement);
+            map_invalidate_tile_full(loc);
 
-        // Update the window
-        rct_window* const tileInspectorWindow = window_find_by_class(WC_TILE_INSPECTOR);
-        if (tileInspectorWindow != nullptr && loc == windowTileInspectorTile.ToCoordsXY())
-        {
```
Broxzier, line 227: It looks like that now this check is removed, the code bellow will be called for anyone, even if they have a different tile selected.

src/openrct2/world/TileInspector.cpp
```diff
@@ -236,164 +236,154 @@ GameActionResultPtr tile_inspector_remove_element_at(const CoordsXY& loc, int16_
...
+            if (!SwapTileElements(loc, first, second))
+            {
+                return std::make_unique<GameActions::Result>(GameActions::Status::Unknown, STR_NONE);
+            }
+            map_invalidate_tile_full(loc);
 
-        // Update the window
-        rct_window* const tileInspectorWindow = window_find_by_class(WC_TILE_INSPECTOR);
-        if (tileInspectorWindow != nullptr && loc == windowTileInspectorTile.ToCoordsXY())
-        {
```
Broxzier, line 259: Same here

src/openrct2/world/TileInspector.cpp
```diff
@@ -30,1162 +30,1104 @@
...
 
-        // Set the base height to be the same as the selected element
-        TileElement* const selectedElement = map_get_nth_element_at(loc, elementIndex + 1);
-        if (selectedElement == nullptr)
+        return true;
+    }
+
+    static rct_window* GetTileInspectorWithPos(const CoordsXY& loc)
+    {
+        // Update the tile inspector's list for everyone who has the tile selected
```
Broxzier, line 78:
```suggestion
        // Return the tile inspector window for everyone who has the tile selected
```

src/openrct2/world/TileInspector.cpp
```diff
@@ -30,1162 +30,1104 @@
...
-        const auto rotatedCurrentTile = CoordsXYZ{ CoordsXY{ tiles[i].x_offset, tiles[i].y_offset }.Rotate(direction),
-                                                   tiles[i].z_offset };
+        const rct_scenery_entry* const largeEntry = largeScenery->GetEntry();
+        const auto* const tiles = largeEntry->large_scenery.tiles;
+        const auto direction = largeScenery->GetDirection();
+
+        const auto rotatedFirstTile = CoordsXYZ{ CoordsXY{ tiles[largeScenery->GetSequenceIndex()].x_offset,
+                                                           tiles[largeScenery->GetSequenceIndex()].y_offset }
+                                                     .Rotate(direction),
+                                                 tiles[largeScenery->GetSequenceIndex()].z_offset };
```
Broxzier, line 159: A trailing comma will likely make this readable:

```suggestion
        const auto rotatedFirstTile = CoordsXYZ{
            CoordsXY{ tiles[largeScenery->GetSequenceIndex()].x_offset, tiles[largeScenery->GetSequenceIndex()].y_offset }
                .Rotate(direction),
            tiles[largeScenery->GetSequenceIndex()].z_offset,
        };
```

Broxzier (changes requested):
The changes look good. I've tested it with a server and client.

src/openrct2/world/TileInspector.cpp
ZehMatt: I can probably do a little better by having sequence index being a separate variable also.

ZehMatt: Still pretty hefty but a bit better now.

Broxzier (approved):

PR #14128, opened 2021-02-20
Simplify velocity acceleration

ZehMatt:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -9667,20 +9667,9 @@ int32_t Vehicle::UpdateTrackMotion(int32_t* outStation)
...
-        accelerationDecrease1 = -accelerationDecrease1;
-    }
-    else
-    {
-        accelerationDecrease1 >>= 12;
-    }
 
-    curAcceleration -= accelerationDecrease1;
+    int32_t curAcceleration = newAcceleration;
+    curAcceleration -= vehicle->velocity / 4096;
```
duncanspumpkin, line 9672: is there the same calculation in the mini golf code?

ZehMatt: Can't say, thats the only place I've found so far that did this.

duncanspumpkin (approved):

PR #14129, opened 2021-02-20
Refactor UpdateSwingingCar function

ZehMatt:
This no longer hurts reading

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -6943,6 +6943,41 @@ int32_t Vehicle::GetSwingAmount() const
     return 0;
 }
 
+static uint8_t GetSwingSprite(int16_t swingPosition)
+{
+    if (swingPosition < -10012)
+        return 11;
+    else if (swingPosition > 10012)
+        return 12;
+
+    if (swingPosition < -8191)
+        return 9;
+    else if (swingPosition > 8191)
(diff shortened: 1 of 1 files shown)


ZehMatt:
Looks like I broke the replays, I'll look into that later.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6943,6 +6943,41 @@ int32_t Vehicle::GetSwingAmount() const
...
+        return 3;
+    else if (swingPosition > 2730)
+        return 4;
+
+    if (swingPosition < -910)
+        return 1;
+    else if (swingPosition > 910)
+        return 0;
+
+    return 11;
```
duncanspumpkin, line 6978: return 0

src/openrct2/ride/Vehicle.cpp
```diff
@@ -6943,6 +6943,41 @@ int32_t Vehicle::GetSwingAmount() const
...
+
+    if (swingPosition < -2730)
+        return 3;
+    else if (swingPosition > 2730)
+        return 4;
+
+    if (swingPosition < -910)
+        return 1;
+    else if (swingPosition > 910)
+        return 0;
```
duncanspumpkin, line 6976: return 2

duncanspumpkin (approved):

PR #14131, opened 2021-02-21
Fix regression in CreateSocket

IntelOrca:
`_socket` to `sock`


Gymnasiast:
What was the regression?

IntelOrca:
I used the wrong object for a parameter.

PR #14134, opened 2021-02-21
Rework how entity export works to allow for easier changes in future

duncanspumpkin:
I kind of wanted to use a fold expression in here to make it simpler but can't do that just yet unfortunately as the loops all have a few slight differences.

The new version first sets everything as a free sprite.
It then loops through all of a single type exporting them.
It then repeats this for each of the types.
When all of the loops are type specific this will be much simpler to iterate.

### DIFF
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -188,7 +188,7 @@ void S6Exporter::Export()
     // Map elements must be reorganised prior to saving otherwise save may be invalid
     map_reorganise_elements();
     ExportTileElements();
-    ExportEntitys();
+    ExportEntities();
     ExportParkName();
 
     _s6.initial_cash = gInitialCash;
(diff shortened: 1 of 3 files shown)


src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -188,7 +188,7 @@ void S6Exporter::Export()
     // Map elements must be reorganised prior to saving otherwise save may be invalid
     map_reorganise_elements();
     ExportTileElements();
-    ExportSprites();
+    ExportEntitys();
```
Gymnasiast, line 191: “Entities”, not “Entitys”.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1082,97 +1040,107 @@ void S6Exporter::ExportSpriteCommonProperties(RCT12SpriteBase* dst, const Sprite
     dst->sprite_direction = src->sprite_direction;
 }
 
-void S6Exporter::ExportSpriteVehicle(RCT2SpriteVehicle* dst, const Vehicle* src)
+template<> void S6Exporter::ExportEntity(RCT12SpriteBase* dst, const Vehicle* src)
 {
+    auto* cdst = static_cast<RCT2SpriteVehicle*>(dst);
```
Gymnasiast, line 1045: The change from dst to cdst makes the diff quite a bit larger. Could we perhaps name the `RCT12SpriteBase` object `origDst` or `baseDst` or something like this, so that the `RCT2SpriteVehicle` can be renamed back to `dst`, avoiding a large diff?

duncanspumpkin: I've cleaned it up and removed the whole base and specific so it should be easier to follow now.

PR #14141, opened 2021-02-21
Fix #14140: Some shortcut keys are not working

### LINKED ISSUE #14140: Some shortcut keys are not working
**OS:** Windows 10
**Version:** 0.3.2
**Commit/Build:** c9ca066

- Example 1: Pressing the shortcut key that toggles hiding vertical land (in this case, V) actually hides the base land instead, thus forcing the user to manually toggle "Hide Vertical Faces" on and off via the view options dropdown menu.
- Example 2: Pressing I to highlight path issues actually toggles on height markers for footpaths instead.

**Steps to reproduce:**
1. Map V as the "Hide vertical land toggle" shortcut key
2. Press it a few times
3. You will notice that it hides base land instead of vertical land

**Screenshots:**
Gaia's Genesis 2021-02-21 10-18-56
Before pressing V
Gaia's Genesis 2021-02-21 10-18-55
After pressing V
Gaia's Genesis 2021-02-21 10-26-41
As you can see here, I have pressed I to highlight path issues, but what actually happens instead is that the height markers for footpaths are toggled on and off.

IntelOrca:


src/openrct2-ui/input/InputManager.cpp
```diff
@@ -343,7 +343,8 @@ bool InputManager::GetState(const ShortcutInput& shortcut) const
             }
             case InputDeviceKind::Keyboard:
             {
-                if (_keyboardState[shortcut.Button & 0xFF])
+                auto scanCode = SDL_GetScancodeFromKey(shortcut.Button);
+                if (scanCode < _keyboardState.size() && _keyboardState[scanCode])
```
Gymnasiast, line 347: MSVC doesn't like this (says it's a signed/unsigned mismatch).

Gymnasiast (approved):
May be merged once CIs are green.

Gymnasiast:
Remember to set the milestone.

PR #14142, opened 2021-02-22
Feature: Add option for taking transparent screenshots

CorySanin:


src/openrct2/interface/Screenshot.h
```diff
@@ -49,6 +49,7 @@ struct CaptureOptions
     std::optional<CaptureView> View;
     ZoomLevel Zoom;
     uint8_t Rotation{};
+    bool Transparent;
```
IntelOrca, line 52:
```suggestion
    bool Transparent{};
```

distribution/openrct2.d.ts
```diff
@@ -314,6 +314,11 @@ declare global {
          * Rotation of the camera from 0 to 3.
          */
         rotation: number;
+
+        /**
+         * Whether to enable transparency in the screenshot
```
IntelOrca, line 319:
```suggestion
         * Whether to enable transparency in the screenshot.
```

IntelOrca (changes requested):
Plugin API version should be incremented.

IntelOrca (approved):

CorySanin:
Can someone re-run the Docker check? Looks like some kind of fluke. My Docker image builds fine locally.

PR #14163, opened 2021-02-24
Use GetRideTypeDescriptor() in more places

Gymnasiast:
Also guarantee that there will always be an RTD to return, preventing crashes from hacked rides or ride types only present in newer versions of OpenRCT2.

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -2107,7 +2107,7 @@ static void populate_vehicle_type_dropdown(Ride* ride)
 
     for (; rideTypeIterator <= rideTypeIteratorMax; rideTypeIterator++)
     {
-        if (selectionShouldBeExpanded && ride_type_has_flag(rideTypeIterator, RIDE_TYPE_FLAG_FLAT_RIDE))
+        if (selectionShouldBeExpanded && GetRideTypeDescriptor(rideTypeIterator).HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
             continue;
         if (selectionShouldBeExpanded && (rideTypeIterator == RIDE_TYPE_MAZE || rideTypeIterator == RIDE_TYPE_MINI_GOLF))
             continue;
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -1142,16 +1142,15 @@ static void window_ride_disable_tabs(rct_window* w)
     if (ride == nullptr)
         return;
 
-    uint8_t ride_type = ride->type; // ecx
     const auto& rtd = ride->GetRideTypeDescriptor();
 
     if (!rtd.HasFlag(RIDE_TYPE_FLAG_HAS_DATA_LOGGING))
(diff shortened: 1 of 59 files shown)


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2536,7 +2535,7 @@ static rct_string_id window_ride_get_status_vehicle(rct_window* w, Formatter& ft
         return STR_EMPTY;
 
     auto stringId = VehicleStatusNames[static_cast<size_t>(vehicle->status)];
-    if ((RideTypeDescriptors[ride->type].Flags & RIDE_TYPE_FLAG_SINGLE_SESSION)
+    if ((ride->GetRideTypeDescriptor().Flags & RIDE_TYPE_FLAG_SINGLE_SESSION)
```
IntelOrca, line 2538:
```suggestion
    if ((ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_SINGLE_SESSION))
```

IntelOrca:
There over 95 occurances of `ride->GetRideTypeDescriptor().HasFlag`. Do you think we should have a helper on Ride?
`ride->HasRideTypeDescriptorFlag`?

Gymnasiast:
I'm not entirely sure, I expect many functions to use the RTD more than once. In cases like those, doing
```C++
const auto& rtd = ride->GetRideTypeDescriptor();
if (rtd.HasFlag(RIDE_TYPE_FLAG_FOO)) 
{
    // etc.
}
```
would make more sense.

IntelOrca:
@Gymnasiast certainly in those cases. I was talking about for any case where you had a single method with only a call to check a flag. I don't know many of those there are.

Gymnasiast:
I'll look into tomorrow, then.

src/openrct2/ride/Ride.cpp
```diff
@@ -397,7 +397,7 @@ money32 Ride::CalculateIncomePerHour() const
         priceMinusCost -= GetShopItemDescriptor(currentShopItem).Cost;
     }
 
-    currentShopItem = (lifecycle_flags & RIDE_LIFECYCLE_ON_RIDE_PHOTO) ? RideTypeDescriptors[type].PhotoItem
+    currentShopItem = (lifecycle_flags & RIDE_LIFECYCLE_ON_RIDE_PHOTO) ? ::GetRideTypeDescriptor(type).PhotoItem
```
duncanspumpkin, line 400: You can just use the ride version here. Type is the member variable.

src/openrct2/ride/Ride.cpp
```diff
@@ -7333,10 +7320,7 @@ uint64_t Ride::GetAvailableModes() const
 
 const RideTypeDescriptor& Ride::GetRideTypeDescriptor() const
 {
-    if (type >= std::size(RideTypeDescriptors))
-        return DummyRTD;
-
-    return RideTypeDescriptors[type];
+    return ::GetRideTypeDescriptor(type);
 }
```
duncanspumpkin, line 7324: This I the only function in this file that needs to use `::` the ride members can call this version and not need to pass the type

duncanspumpkin (approved):

Issue #14095, opened 2021-02-14
Holding down [-][+] buttons does not decrease/increase num. of circuit


telk5093:
**OS:** Windows 10
**Version:** 0.3.2
**Commit/Build:** 6e20f41

![](https://user-images.githubusercontent.com/10726524/107879290-45284800-6f1b-11eb-8912-1ce3f2c3759a.gif)

There is [-][+] buttons in the ride configuration tab which can adjust integer values, such as Launch speed, chain speed, number of circuit.
Hold to click the left mouse button to the [-][+] widget buttons decreases/increases the values for Launch speed and chain speed, but not for **Number of circuit**.

IntelOrca:
Needs correct flag set in `hold_down_widgets`.

Gymnasiast:
In the longer run, I think that every spinner/stepper should automatically have this. This does require a widget refactor though, so we'll get to that in 2029, then.

PR #14171, opened 2021-02-24
[Plugin] Add custom widget and other improvements

IntelOrca:
* Add API for getting network traffic statistics.
* Add API for creating custom widgets.
* Add API for drawing graphics for custom widgets.
* Add click event to spinners and allow them to be held down.

image
image

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -71,6 +71,14 @@ declare global {
     }
 
     /**
+     * Represents the width and height in pixels.
+     */
+    interface ScreenSize {
+        width: number;
+        height: number;
+    }
+
+    /**
      * A coordinate within the game.
      * Each in-game tile is a size of 32x32.
      */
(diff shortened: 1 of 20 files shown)


distribution/openrct2.d.ts
```diff
@@ -2125,8 +2125,16 @@ declare global {
     interface SpinnerWidget extends WidgetBase {
         type: 'spinner';
         text?: string;
+
+        /**
+         * If true, the user can click and hold the decrement and increment
+         * button for repeated calls.
+         */
+        isHoldable?: boolean;
```
Gymnasiast, line 2133: Why would we want to turn this off? Native OS spinners/steppers always allow holding down, AFAICT. The fact that OpenRCT2 doesn't is a misfeature IMO.

IntelOrca: You mentioned this in https://github.com/OpenRCT2/OpenRCT2/issues/14095#issuecomment-779663385
I was thinking about making the default, but if you think the option should just be removed entirely, I can do that.

Gymnasiast: I think removing it entirely and always allowing holding them down is indeed best. It improves usability and consistency (including with native OS widgets).

IntelOrca: ok, done

src/openrct2-ui/interface/Window.cpp
```diff
@@ -681,6 +688,21 @@ void InvalidateAllWindowsAfterInput()
     });
 }
 
+bool Window::IsLegacy()
```
Gymnasiast, line 691: What does "legacy" mean in this context?

duncanspumpkin: If it has been converted to a class or if it's the old style rct_window

Gymnasiast (approved):

PR #14189, opened 2021-02-27
Refactor text drawing

Gymnasiast:
This PR:
- Removes gCurrentFontSpriteBase
- Unifies several text drawing calls with more consistent syntax
- Makes calling parameters easier to follow by using enum classes


src/openrct2-ui/interface/InGameConsole.cpp
```diff
@@ -26,10 +26,14 @@ using namespace OpenRCT2::Ui;
 
 static InGameConsole _inGameConsole;
 
+static FontSpriteBase InGameconsoleGetFontSpriteBase()
```
ZehMatt, line 29:
```suggestion
static FontSpriteBase InGameConsoleGetFontSpriteBase()
```

PR #14191, opened 2021-02-28
Fix #14190: Game crash likely related to plug-in hotkeys

### LINKED ISSUE #14190: Game crash likely related to plug-in hotkeys.
The issue seems to be with hotkey usage in plug-ins. This is what I could observe:
No hotkeys registered: No problem.
With hotkeys: The game crashes at seemingly random occasions (relatively soon after loading), e.g. when moving the camera, interacting with a window, ...
I do not see a relation between the actions that I can make to let the game crash and anything plug-in related. I.e. the plug-in is loaded but I do not interact with it and I do not use any hotkeys.

[2020e8c0-cab0-4bf8-bbc1-306882ba526d(314faa6_x86-64).dmp.gz](https://github.com/OpenRCT2/OpenRCT2/files/6055906/2020e8c0-cab0-4bf8-bbc1-306882ba526d.314faa6_x86-64.dmp.gz)
[2020e8c0-cab0-4bf8-bbc1-306882ba526d.sv6.gz](https://github.com/OpenRCT2/OpenRCT2/files/6055907/2020e8c0-cab0-4bf8-bbc1-306882ba526d.sv6.gz)

IntelOrca:
* Do not cache references to RegisteredShortcut as they can be invalidated when new shortcuts are registered / removed.
* Use a map to improve query performance of shortcut by ID.
* Store a separate list of strings for the map to use as a key.

@ZehMatt I tried using `std::list<std::string>` for `_ids`, but for some reason the string data for elements after the removed element were getting invalidated / shuffled. I am surprised by this as I thought `std::list` was supposed to ensure the data and address for each element is unchanged when the list is modified. I had to resort to `std::vector<std::unique_ptr<std::string>>` in the end which is a big ugly. The whole reason for the separate list of IDs is so I can use `std::string_view` for the map. That way you don't need to create `std::string` when calling `GetShortcut`.

### DIFF
--- a/src/openrct2-ui/input/InputManager.cpp
+++ b/src/openrct2-ui/input/InputManager.cpp
@@ -288,13 +288,16 @@ void InputManager::ProcessHoldEvents()
     _viewScroll.x = 0;
     _viewScroll.y = 0;
 
-    auto& shortcutManager = GetShortcutManager();
-    if (!shortcutManager.IsPendingShortcutChange())
+    if (!HasTextInputFocus())
     {
-        ProcessViewScrollEvent(ShortcutId::ViewScrollUp, { 0, -1 });
-        ProcessViewScrollEvent(ShortcutId::ViewScrollDown, { 0, 1 });
-        ProcessViewScrollEvent(ShortcutId::ViewScrollLeft, { -1, 0 });
-        ProcessViewScrollEvent(ShortcutId::ViewScrollRight, { 1, 0 });
+        auto& shortcutManager = GetShortcutManager();
+        if (!shortcutManager.IsPendingShortcutChange())
+        {
+            ProcessViewScrollEvent(ShortcutId::ViewScrollUp, { 0, -1 });
+            ProcessViewScrollEvent(ShortcutId::ViewScrollDown, { 0, 1 });
+            ProcessViewScrollEvent(ShortcutId::ViewScrollLeft, { -1, 0 });
+            ProcessViewScrollEvent(ShortcutId::ViewScrollRight, { 1, 0 });
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


ZehMatt:
> * Do not cache references to RegisteredShortcut as they can be invalidated when new shortcuts are registered / removed.
> * Use a map to improve query performance of shortcut by ID.
> * Store a separate list of strings for the map to use as a key.
> 
> @ZehMatt I tried using `std::list<std::string>` for `_ids`, but for some reason the string data for elements after the removed element were getting invalidated / shuffled. I am surprised by this as I thought `std::list` was supposed to ensure the data and address for each element is unchanged when the list is modified. I had to resort to `std::vector<std::unique_ptr<std::string>>` in the end which is a big ugly. The whole reason for the separate list of IDs is so I can use `std::string_view` for the map. That way you don't need to create `std::string` when calling `GetShortcut`.

I'm not sure how you managed to do that but std::list<std::string> should be fine. I've crafted this little test https://godbolt.org/z/ob7vjs to make sure I'm not talking nonsense.

Do you have perhaps some example on how things went side ways?

IntelOrca:
@ZehMatt I will have to create a commit that has it the way it was and then show you what was happening. From what I saw, when it erased an element in a list, it was moving the memory of the next element to the location of the removed element (much like a vector would). This causes issues when the string is using its own struct for the char data rather than a separately allocated block.

ZehMatt:
I mean std::string will invalidate its data when modified/moved but the actual issue is probably storing a pointer rather than passing the object if I had to guess.

ZehMatt (approved):
This fill fix the issue, me and @IntelOrca discussed potential refactoring but not in this PR.

PR #14193, opened 2021-02-28
various openrct2.d.ts improvements

Sadret:
Adds the exact type to each TileElement, overwriting the more general type of BaseTileElement.

Edit:
- add TileElement union type
- add the exact type to each TileElement, overwriting the more general type of BaseTileElement
- use TileElement type in Tile interface (instead of BaseTileElement)
- change single quotes to double quotes
- add type to WidgetBase

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -59,6 +59,9 @@
 - Improved: [#6022] Allow up to 128 ride objects to be selected in track designer.
 - Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 - Improved: [#13386] A GUI error message is now displayed if the language files are missing.
+- Improved: [#14193] [Plugin] Add TileElement union type and use it in Tile interface instead of BaseTileElement.
+- Improved: [#14193] [Plugin] Add exact type field to each TileElement, add type field to WidgetBase.
+- Improved: [#14193] [Plugin] Change single quotes to double quotes in openrct2.d.ts.
 - Removed: [#13423] Built-in explode guests cheat (replaced by plug-in).
 
 0.3.2 (2020-11-01)
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -555,6 +555,10 @@ declare global {
 
     type Direction = 0 | 1 | 2 | 3;
 
+    type TileElement =
+        SurfaceElement | FootpathElement | TrackElement | SmallSceneryElement | WallElement | EntranceElement
+        | LargeSceneryElement | BannerElement | CorruptElement;
+
     interface BaseTileElement {
         type: TileElementType;
         baseHeight: number;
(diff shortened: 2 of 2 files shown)


IntelOrca (approved):

IntelOrca (changes requested):
I think `type` needs to be removed from `BaseTileElement`. I remember having to do that for the Widget types.

Sadret:
Why do you think so? My local IDE does not report any problems.
Also, any BaseTileElement should have a type, so why remove it?

IntelOrca:
I am not quite sure. I just remember needing to do it for Widget, otherwise I couldn't define an array of widgets where the type field was enough to say what widget type it was.

Sadret:
But we don't have a discriminated union type for TileElements, so that cannot be a problem.

Sadret:
Especially, I think we should reintroduce `WidgetBase.type`. This code compiles fine with the `type` field included:
```
function typeTest(widget: Widget) {
    if (widget.type === "button")
        console.log(widget.border);
}
```

Sadret:
Maybe you tested against `WidgetBase` instead of `Widget`?

IntelOrca:
No I think was when I had an array of widgets defined:
```
widgets: [
    { type: 'button', etc. },
    { type: 'label', etc. },
]
```

Sadret:
Maybe the type inference just failed and assumed `WidgetBase` instead of `Widget`?

Sadret:
I have two other things that I think should be changed:

- add a `TileElement` union type and use it where it fits (similiar to `Widget` vs `WidgetBase`), e.g. in the `Tile` interface.
```
    type TileElement =
        SurfaceElement | FootpathElement | TrackElement | SmallSceneryElement | WallElement | EntranceElement
        | LargeSceneryElement | BannerElement | CorruptElement;
```

- use " (double quotes) consistently instead of ' (single quotes) (as double quotes is the majority right now)

If that's ok, I will modify the PR to include this and `WidgetBase.type`.

IntelOrca:
> If that's ok, I will modify the PR to include this and WidgetBase.type.

I will need to find out why I couldn't do this. In the meantime I am happy for you to do the other changes.

distribution/changelog.txt
```diff
@@ -59,6 +59,7 @@
 - Improved: [#6022] Allow up to 128 ride objects to be selected in track designer.
 - Improved: [#12917] Changed peep movement so that they stay more spread out over the full width of single tile paths.
 - Improved: [#13386] A GUI error message is now displayed if the language files are missing.
+- Improved: [#14193] Various openrct2.d.ts improvements.
```
tupaschoal, line 62: I'd rather have multiple lines with the same PR number than having a vague changelog entry.

IntelOrca (approved):

distribution/changelog.txt
Sadret: I grouped the changes into three entries.

PR #14196, opened 2021-02-28
Remove near-duplicates of map_get_highest_z

mbugert:
There were several near-duplicates of `map_get_highest_z` (finding the lowest z height on a given tile above slopes/water). This PR replaces those duplicates with calls to `map_get_highest_z`.

Large scenery can be placed underwater with cheats, so I added a boolean flag to `map_get_highest_z` to cover this use case.

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -1778,7 +1778,7 @@ static void clear_elements_at(const CoordsXY& loc)
     clear_element_at(loc, &tileElement);
 }
 
-int32_t map_get_highest_z(const CoordsXY& loc, bool strictlyAboveWater)
+int32_t map_get_highest_z(const CoordsXY& loc)
 {
     auto surfaceElement = map_get_surface_element_at(loc);
     if (surfaceElement == nullptr)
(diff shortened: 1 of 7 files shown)


duncanspumpkin:
Using boolean flags like this is pretty horrible to read it doesn't convey what it is there for. Perhaps two functions would make more sense.

mbugert:
You're right. WDYT now?

duncanspumpkin:
Could you rebase this?

src/openrct2/world/Map.cpp
```diff
@@ -1786,16 +1786,27 @@ int32_t map_get_highest_z(const CoordsXY& loc)
...
     if ((surfaceElement->GetSlope() & TILE_ELEMENT_SLOPE_ALL_CORNERS_UP) != TILE_ELEMENT_SLOPE_FLAT)
         z += LAND_HEIGHT_STEP;
     if ((surfaceElement->GetSlope() & TILE_ELEMENT_SLOPE_DOUBLE_HEIGHT) != 0)
         z += LAND_HEIGHT_STEP;
 
-    z = std::max(z, surfaceElement->GetWaterHeight());
     return z;
 }
 
+int32_t map_get_highest_z_above_water_height(const CoordsXY& loc)
```
duncanspumpkin, line 1798: All new functions should be UpperCamelCase

src/openrct2/world/Map.cpp
```diff
@@ -1786,16 +1786,27 @@ int32_t map_get_highest_z(const CoordsXY& loc)
 
     auto z = surfaceElement->GetBaseZ();
 
-    // Raise z so that is above highest point of land and water on tile
+    // Raise z so that is above highest point of land
```
duncanspumpkin, line 1789: I would suggest renaming this function to add land into the name. Make it more clear that this is now dedicated to land and the other function is dedicated to water.

duncanspumpkin: The other benefit of renaming is that it will show every user of the function and make it easier to confirm that all users are correctly modified.

tupaschoal:
@mbugert I'm closing this for lack of activity, if  you feel like fixing the issues pointed out, please reopen the PR!

Issue #14093, opened 2021-02-14
User defined literals for Money32


duncanspumpkin:
We currently use a macro for writing down a money amount as a literal. We should move to a more modern C++ representation with a user defined literal like the below:

```
constexpr money32 operator"" _GBP(long double money)
{
    return money * 10;
}

static_assert(MONEY(2, 40) == 2.40_GBP);
```

Go through the codebase and replace all MONEY macro uses with the equivalent version. You might need to also make a helper function that can convert from a runtime value to money as the macro is currently used in one or two places.

geoff-B:
I'm interested in taking on this issue, but one quick question: Why does the money have to be multiplied by 10? Doesn't 2.40_GBP simply represent £2.40?

mwyzhao:
Hi, is this issue still open? I see a ton of commits from geoff-B from a year ago but no merge request.

tupaschoal:
There is a PR #14293, though it seems to be stalled for quite some time, not sure if Geoff will be able to pick up on it or not (there are a lot of merge conflicts to solve)

I asked there anyway and can update here if so, in the meantime, I advise you to look for other good first issues.

PR #14207, opened 2021-03-01
#14093 User defined literals for Money32

geoff-B:
* Replace MONEY macro with user defined literal `_GBP` when converting literals
  * `MONEY(2, 40)` becomes `2.40_GBP`
* Replace MONEY macro with `contexpr auto MONEY()` function when converting runtime values 
 (not sure if this is the right way to do it)


src/openrct2/actions/FootpathAdditionRemoveAction.cpp
```diff
@@ -85,7 +85,7 @@ GameActions::Result::Ptr FootpathAdditionRemoveAction::Query() const
     }
     auto res = MakeResult();
     res->Position = _loc;
-    res->Cost = MONEY(0, 0);
+    res->Cost = 0.0_GBP;
```
Gymnasiast, line 88:
```suggestion
    res->Cost = 0.00_GBP;
```

Same for line 113.

src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -27,8 +27,8 @@
 #include <openrct2/world/Park.h>
 #include <openrct2/world/Surface.h>
 
-#define CHEATS_MONEY_DEFAULT MONEY(10000, 00)
-#define CHEATS_MONEY_INCREMENT_DIV MONEY(5000, 00)
+#define CHEATS_MONEY_DEFAULT 10000.00_GBP
```
Gymnasiast, line 30: Can you turn all defines for money into `constexpr`s?

src/openrct2-ui/windows/Cheats.cpp
```diff
@@ -559,7 +559,7 @@ class CheatsWindow final : public Window
             case WINDOW_CHEATS_PAGE_GUESTS:
             {
                 auto ft = Formatter::Common();
-                ft.Add<int32_t>(MONEY(1000, 00));
+                ft.Add<int32_t>(1000.00_GBP);
```
Gymnasiast, line 562:
```suggestion
                ft.Add<money32>(1000.00_GBP);
```

src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -458,24 +458,24 @@ static void window_editor_objective_options_arg_1_increase(rct_window* w)
...
         case OBJECTIVE_MONTHLY_FOOD_INCOME:
-            if (gScenarioObjective.Currency >= MONEY(2000000, 00))
+            if (gScenarioObjective.Currency >= 2000000.00_GBP)
             {
                 context_show_error(STR_CANT_INCREASE_FURTHER, STR_NONE, {});
             }
             else
             {
-                gScenarioObjective.Currency += MONEY(100, 0);
+                gScenarioObjective.Currency += 100.0_GBP;
```
Gymnasiast, line 478:
```suggestion
                gScenarioObjective.Currency += 100.00_GBP;
```

src/openrct2-ui/windows/EditorObjectiveOptions.cpp
```diff
@@ -458,24 +458,24 @@ static void window_editor_objective_options_arg_1_increase(rct_window* w)
...
         case OBJECTIVE_REPAY_LOAN_AND_PARK_VALUE:
-            if (gScenarioObjective.Currency >= MONEY(2000000, 00))
+            if (gScenarioObjective.Currency >= 2000000.00_GBP)
             {
                 context_show_error(STR_CANT_INCREASE_FURTHER, STR_NONE, {});
             }
             else
             {
-                gScenarioObjective.Currency += MONEY(1000, 0);
+                gScenarioObjective.Currency += 1000.0_GBP;
```
Gymnasiast, line 467:
```suggestion
                gScenarioObjective.Currency += 1000.00_GBP;
```

Please check the whole diff and make sure there are no more occurrences of `.0_GBP`.

src/openrct2/common.h
```diff
@@ -135,10 +135,18 @@ using money64 = fixed64_1dp;
...
+    return money * 10;
+}
+// Helper function to convert runtime values to money
+constexpr auto MONEY(int32_t whole, int32_t fraction)
+{
+    return (whole)*10 + ((fraction) / 10);
+}
 
-#define MONEY_FREE MONEY(0, 00)
+#define MONEY_FREE 0.00_GBP
```
Gymnasiast, line 149: Please remove this define entirely - it does not add anything of value (pun not intended).

src/openrct2-ui/windows/Cheats.cpp
geoff-B: done in 47469c6

src/openrct2/world/Park.h
```diff
@@ -16,7 +16,7 @@
 #define DECRYPT_MONEY(money) (static_cast<money32>(rol32((money) ^ 0xF4EC9621, 13)))
 #define ENCRYPT_MONEY(money) (static_cast<money32>(ror32((money), 13) ^ 0xF4EC9621))
 
-#define MAX_ENTRANCE_FEE MONEY(200, 00)
+#define MAX_ENTRANCE_FEE 200.00_GBP
```
Gymnasiast, line 19: Also make this a constexpr

src/openrct2/common.h
```diff
@@ -135,10 +135,17 @@ using money64 = fixed64_1dp;
...
 
-// Construct a money value in the format MONEY(10,70) to represent 10.70. Fractional part must be two digits.
-#define MONEY(whole, fraction) ((whole)*10 + ((fraction) / 10))
+// Convert money literal to money32
+constexpr money32 operator"" _GBP(long double money)
+{
+    return money * 10;
+}
+// Helper function to convert runtime values to money
+constexpr auto MONEY(int32_t whole, int32_t fraction)
```
Gymnasiast, line 144: Is this function still used?

src/openrct2/localisation/Localisation.cpp
```diff
@@ -502,7 +502,7 @@ money32 string_to_money(const char* string_to_monetise)
     auto whole = static_cast<int32_t>(number);
     auto fraction = static_cast<uint8_t>(ceil((number - whole) * 100.0));
 
-    money32 result = MONEY(whole, fraction);
+    money32 result = (whole)*10 + ((fraction) / 10);
```
Gymnasiast, line 505: I guess there are two approaches here:
- Reinstate the MONEY define/constexpr
- Just multiply/divide `number` by the conversion rate and feed the result into `result`, skipping the splitting out of `whole` and `fraction`.

Personally, I'm gravitating towards the latter. @duncanspumpkin what do you think?

duncanspumpkin: Yes scrap the whole, fraction just multiply by the factor 10.

src/openrct2/common.h
duncanspumpkin: I would like to see it gone

src/openrct2/localisation/Localisation.cpp
geoff-B: @duncanspumpkin How would we want to check for MONEY overflow then, as a float?

src/openrct2/common.h
geoff-B: @duncanspumpkin It is still used in InteractiveConsole.cpp and LandSetHeightAction.cpp
image

geoff-B:
Sorry for the mess, tried to change some commit messages but ended up duplicating these commits. Everything should be good now, except the MONEY conexpr and the issue in Localisation.cpp, which I need input on.

src/openrct2/localisation/Localisation.cpp
duncanspumpkin: You've two options. Either do `uint64_t amount = std::min<uint64_t>(amountFloat *10.0, std::limits<uint32>::max()` or do `floatAmount = std::min<double>(floatMount*10, std::limits<uint32>::max())`.
In general we try to minimise the amount of time anything is a floating point so I'd go with the first option. Then of course cast the final result back to uint32.

src/openrct2/localisation/Localisation.cpp
```diff
@@ -499,17 +499,11 @@ money32 string_to_money(const char* string_to_monetise)
...
-    money32 result = (whole)*10 + ((fraction) / 10);
     // Check if MONEY resulted in overflow
-    if ((whole > 0 && result < 0) || result / 10 < whole)
-    {
-        result = INT_MAX;
-    }
+    uint64_t result = std::min<uint64_t>(number * 10.0, (std::numeric_limits<uint32_t>::max)());
     result *= sign;
-    return result;
+    return (uint32_t)result;
```
duncanspumpkin, line 506: Don't use c style casts. Use a static cast

src/openrct2/common.h
duncanspumpkin: And can they be modified to not require it?

geoff-B: I've gone ahead and done that, it wasn't as difficult as I thought.

src/openrct2/actions/LandSetHeightAction.cpp
```diff
@@ -245,7 +245,7 @@ money32 LandSetHeightAction::GetSmallSceneryRemovalCost() const
         if (sceneryEntry == nullptr)
             continue;
 
-        cost += MONEY(sceneryEntry->small_scenery.removal_price, 0);
+        cost += sceneryEntry->small_scenery.removal_price;
```
duncanspumpkin, line 248: Shouldn't this be multiplied by 10

src/openrct2/actions/LandSetHeightAction.cpp
```diff
@@ -359,7 +359,7 @@ money32 LandSetHeightAction::GetSurfaceHeightChangeCost(SurfaceElement* surfaceE
     {
         int32_t cornerHeight = tile_element_get_corner_height(surfaceElement, i);
         cornerHeight -= map_get_corner_height(_height, _style & TILE_ELEMENT_SURFACE_SLOPE_MASK, i);
-        cost += MONEY(abs(cornerHeight) * 5 / 2, 0);
+        cost += abs(cornerHeight) * 5 / 2;
```
duncanspumpkin, line 362: And this

tupaschoal:
Hey @geoff-B , sorry for noticing after so long, but would you be willing to rebase this for another round of review?

Gymnasiast:
@tupaschoal I would like to review this.

src/openrct2/ride/ShopItem.cpp
```diff
@@ -25,61 +25,61 @@ uint64_t gSamePriceThroughoutPark;
...
+    // Item,                            Cost, Base value,  Hot value, Cold value, Default price, Image,                               Price label,                                      Singular,                                   Plural,                                     Indefinite,                                     Display (in guest inventory),                Shop Item Flag,                                       Litter type,                    Consumption time, Discard Container,          Peep thought price too much,            Peep thought price good value,
+    /* ShopItem::Balloon */          {  3,    14,          14,        14,         0.90_GBP,      SPR_SHOP_ITEM_BALLOON,             { STR_SHOP_ITEM_PRICE_LABEL_BALLOON,                STR_SHOP_ITEM_SINGULAR_BALLOON,             STR_SHOP_ITEM_PLURAL_BALLOON,               STR_SHOP_ITEM_INDEFINITE_BALLOON,               STR_SHOP_ITEM_DISPLAY_BALLOON             }, SHOP_ITEM_FLAG_IS_SOUVENIR,                           Litter::Type::Rubbish,          0,                ShopItem::None,             PeepThoughtType::BalloonMuch,           PeepThoughtType::Balloon            },
```
Gymnasiast, line 29: Please use the literals for the other columns too.

src/openrct2/actions/LandSetHeightAction.cpp
```diff
@@ -246,7 +246,7 @@ money32 LandSetHeightAction::GetSmallSceneryRemovalCost() const
         if (sceneryEntry == nullptr)
             continue;
 
-        cost += MONEY(sceneryEntry->removal_price, 0);
```
Gymnasiast, line 249: For cases like this, I would still use the macro. It is clearer.

tupaschoal:
> @tupaschoal I would like to review this.

Sure, I'm still getting it to compile again first

src/openrct2/ride/ShopItem.cpp
tupaschoal: This was how it was in the PR, I haven't tried to understand why yet, or of there is any difference.

First I'm just making sure the merge conflicts are solved and it is compiling

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: I think the macro is removed (?)

Not sure though, will check later

Broxzier: Yes, it's removed.

Gymnasiast: Then I think it should be reintroduced, for cases like these.

src/openrct2/ride/ShopItem.cpp
Gymnasiast: Sure, no worries. I don’t mind making the changes myself either.

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: What about some `constexpr` function like:

```c++
constexpr money32 toGBP(long double money) noexcept
{
    return money * 10;
}
```

At least we dodge from macros :P. I noticed most of the places that are not using literals are either using the macro with a fraction equal to 0, or a double/float already.


PS: I think while searching for this I also found some places where the conversion was not right after this patch, see the latest commit.

tupaschoal (commented):
Well, it now compiles (or used to before my two latest patches). I've also tended to the open issues and folded all of the previous patches into one, as they were most loose commits trying to fix formatting and code.

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: I went ahead and added it anyway to speed things up.

duncanspumpkin: Code should only ever use floating point for ui. Can you restructure this so that it follows that convention.

src/openrct2/actions/LandSetHeightAction.cpp
```diff
@@ -246,7 +246,7 @@ money32 LandSetHeightAction::GetSmallSceneryRemovalCost() const
         if (sceneryEntry == nullptr)
             continue;
 
-        cost += (sceneryEntry->removal_price) * 10;
+        cost += toGBP(static_cast<long double>(sceneryEntry->removal_price));
```
AaronVanGeffen, line 249: Not sure how this is easier to read than simply having `(sceneryEntry->removal_price) * 10` here like before. @Gymnasiast, were you thinking of another solution?

Broxzier: I'm confused why this is necessary at all. Why doesn't `sceneryEntry->removal_price` already hold the money value is represents?

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: Not sure what you meant, some of the parameters are `double`, like: 
```c++
            money32 money = toGBP(double_val[0]);
```

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: > Not sure how this is easier to read than simply having `(sceneryEntry->removal_price) * 10` here like before. @Gymnasiast, were you thinking of another solution?

To be fair, I'm not 100% sure the `static_cast` is needed, it is converting from signed int to double, so it should be alright to have it without cast, which would make it a lot easier on the eyes.

> I'm confused why this is necessary at all. Why doesn't sceneryEntry->removal_price already hold the money value is represents?

I don't know myself, it's a fair question, the code was already doing this `* 10` multiplication, the patch just kept it 🤔

tupaschoal: I now understand it, I think. The objects store only the whole part of it, so if it costs 30 to remove a given object, the objects will store it as:
```
        "removalPrice": -52,
```

I guess it's just easier on maintaining the objects to have the wholes as they really are than to have them all multiplied by `10` and then explaining people that the rightmost digit is the decimal.

duncanspumpkin: we could *10 on load then change the type to money32 on the object. That would simplify all of this.

src/openrct2/actions/LandSetHeightAction.cpp
duncanspumpkin: yeah but double_val[0] is from the interactive console which is ui. I know it would never cause issues but I like the convention of no floating point makes things easy to reason about and reduces the chance of difficult to think about issues.

src/openrct2/actions/LandSetHeightAction.cpp
Broxzier: That seems like a much better solution.

tupaschoal: I agree @duncanspumpkin , but it seems like too much for this PR, specially as it will break replays, won't it? There are at least four fields to change, `prince` and `removal_price` of small and large scenery.

I can happily tackle that in a different, follow-up PR.

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: Sorry, still not sure about what you're proposing.

src/openrct2/actions/LandSetHeightAction.cpp
Broxzier: Of course a follow-up PR is better, this PR is just about getting the user defined literals in. `toGBP` is a nice substitute for now, and makes it easy to find which places need refactoring.

One nitpicky remark: functions should use TitleCase, `ToGBP`.

tupaschoal: > toGBP is a nice substitute for now, and makes it easy to find which places need refactoring.

Yep, I even searched the code for `* 10` and found more places that can be potentially simplified later on.

> One nitpicky remark: functions should use TitleCase, `ToGBP`.

Definitely, nice catch.

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: I think that's the only missing piece from this PR now @duncanspumpkin

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -944,7 +944,7 @@ static void WindowNewRidePaintRideInformation(
         money64 price = GetRideTypeDescriptor(item.Type).BuildCosts.TrackPrice;
         const auto& ted = GetTrackElementDescriptor(startPieceId);
         price *= ted.Price;
-        price = (price >> 17) * 10 * GetRideTypeDescriptor(item.Type).BuildCosts.PriceEstimateMultiplier;
+        price = ToGBP(price >> 17) * GetRideTypeDescriptor(item.Type).BuildCosts.PriceEstimateMultiplier;
```
duncanspumpkin, line 947: its debatable if this is really correct. The *10 here doesn't really mean convert from gbp to money32 its more of a fancy floor function okay it is kind of related but not really.

src/openrct2/actions/LandSetHeightAction.cpp
```diff
@@ -360,7 +360,7 @@ money32 LandSetHeightAction::GetSurfaceHeightChangeCost(SurfaceElement* surfaceE
     {
         int32_t cornerHeight = tile_element_get_corner_height(surfaceElement, i);
         cornerHeight -= map_get_corner_height(_height, _style & TILE_ELEMENT_SURFACE_SLOPE_MASK, i);
-        cost += MONEY(abs(cornerHeight) * 5 / 2, 0);
+        cost += ToGBP(static_cast<long double>(abs(cornerHeight) * 5 / 2));
```
duncanspumpkin, line 363: the 10 here really represents the cost of modifying the height by a unit. I think `2.5_GBP * abs(cornerHeight)` would be just as good and much more descriptive.

src/openrct2/actions/MazePlaceTrackAction.cpp
```diff
@@ -119,7 +119,7 @@ GameActions::Result MazePlaceTrackAction::Query() const
 
     const auto& ted = GetTrackElementDescriptor(TrackElemType::Maze);
     money32 price = (((ride->GetRideTypeDescriptor().BuildCosts.TrackPrice * ted.Price) >> 16));
-    res.Cost = ((canBuild.Cost + price) / 2) * 10;
+    res.Cost = ToGBP(((canBuild.Cost + price) / 2));
```
duncanspumpkin, line 122: `price` and `canBuild.Cost` are both in terms of `money32` so this doesn't make sense to go through `ToGBP`. Either it should stay as *10 or they are not `money32`.

src/openrct2/actions/MazeSetTrackAction.cpp
```diff
@@ -138,7 +138,7 @@ GameActions::Result MazeSetTrackAction::Query() const
 
         const auto& ted = GetTrackElementDescriptor(TrackElemType::Maze);
         money32 price = (((ride->GetRideTypeDescriptor().BuildCosts.TrackPrice * ted.Price) >> 16));
-        res.Cost = price / 2 * 10;
+        res.Cost = ToGBP(price / 2);
```
duncanspumpkin, line 141: as above its either money32 and * 10 or its some unknown unit and `ToGBP`

src/openrct2/common.h
```diff
@@ -92,10 +92,17 @@ using money64 = fixed64_1dp;
...
 
-// Construct a money value in the format MONEY(10,70) to represent 10.70. Fractional part must be two digits.
-#define MONEY(whole, fraction) ((whole)*10 + ((fraction) / 10))
+// User defined literal to convert money literal to money32
+constexpr money32 operator"" _GBP(long double money) noexcept
+{
+    return money * 10;
+}
+
+constexpr money32 ToGBP(long double money) noexcept
```
duncanspumpkin, line 101: This is named wrong. It should be ToMoney32FromGBP or something shorter. The output is money32 not gbp.

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: Nice suggestion!

tupaschoal:
I reverted most of the `ToMoney32FromGBP` calls, but there are a couple not on InteractiveConsole that I think should remain. Let me know what you think @duncanspumpkin

duncanspumpkin:
@tupaschoal  Can you rebase and just take one more look at the actions. After last week's tweaks it should be a bit easier to follow now.

tupaschoal:
I think it was already rebased to include last week's changes, but done it again anyway @duncanspumpkin .

Note that there was no change wrt to the actions, as due to previous discussions I had already reverted the changes on places that were not literals, such as these two:
https://github.com/OpenRCT2/OpenRCT2/blob/4e8ea47e1b0f6c8564c295a8836c7e8c2a06694f/src/openrct2/actions/MazePlaceTrackAction.cpp#L122

https://github.com/OpenRCT2/OpenRCT2/blob/4e8ea47e1b0f6c8564c295a8836c7e8c2a06694f/src/openrct2/actions/TrackPlaceAction.cpp#L404

duncanspumpkin:
> I think it was already rebased to include last week's changes, but done it again anyway @duncanspumpkin .
> 
> Note that there was no change wrt to the actions, as due to previous discussions I had already reverted the changes on places that were not literals, such as these two:
> 
> https://github.com/OpenRCT2/OpenRCT2/blob/4e8ea47e1b0f6c8564c295a8836c7e8c2a06694f/src/openrct2/actions/MazePlaceTrackAction.cpp#L122
> 
> https://github.com/OpenRCT2/OpenRCT2/blob/4e8ea47e1b0f6c8564c295a8836c7e8c2a06694f/src/openrct2/actions/TrackPlaceAction.cpp#L404

Okay great. Those we can attack in a follow on to multiply out the 5.

duncanspumpkin (approved):

src/openrct2/actions/LandSetHeightAction.cpp
```diff
@@ -360,7 +360,7 @@ money32 LandSetHeightAction::GetSurfaceHeightChangeCost(SurfaceElement* surfaceE
     {
         int32_t cornerHeight = tile_element_get_corner_height(surfaceElement, i);
         cornerHeight -= map_get_corner_height(_height, _style & TILE_ELEMENT_SURFACE_SLOPE_MASK, i);
-        cost += MONEY(abs(cornerHeight) * 5 / 2, 0);
+        cost += 2.5_GBP * cornerHeight;
```
Gymnasiast, line 363: This looks wrong. What if cornerHeight is negative? Seems that make for a negative cost.

Gymnasiast:
@tupaschoal Have you tested the console commands? Especially with decimal values like 7.50?

src/openrct2/actions/LandSetHeightAction.cpp
tupaschoal: Nice catch, didn't mean to remove the `abs`.

src/openrct2/common.h
```diff
@@ -92,10 +92,17 @@ using money64 = fixed64_1dp;
...
 
-// Construct a money value in the format MONEY(10,70) to represent 10.70. Fractional part must be two digits.
-#define MONEY(whole, fraction) ((whole)*10 + ((fraction) / 10))
+// User defined literal to convert money literal to money32
+constexpr money32 operator"" _GBP(long double money) noexcept
+{
+    return money * 10;
+}
+
+constexpr money32 ToMoney32FromGBP(long double money) noexcept
```
duncanspumpkin, line 101: if you provided an int32_t overload for this then that would avoid using the float in CalculateRideValue

tupaschoal: Added it.

tupaschoal:
Can we wrap this up @Gymnasiast and @duncanspumpkin ?

duncanspumpkin:
> Can we wrap this up @Gymnasiast and @duncanspumpkin ?

I'm happy with it hence my approval

PR #14223, opened 2021-03-03
revert Xcode to v26 dependencies

adam-bloom:
Reverts the Xcode project file to its state at f3ed018

In Dependencies v28, we allowed freetype to use brotli. brotli does not appear to fully support rpath at this time, so we'll likely need to revert the inclusion of brotli. I'll investigate when I have some more time. We'll then need a new dependencies release, and can update both Xcode and cmake. FYI @LRFLEW.

The macos-cmake build is not yet "released" anywhere, so I'll leave that pointed at v28.


Gymnasiast:
“... works.” _Also sprach Zarathustra._

LRFLEW:
>  brotli does not appear to fully support rpath at this time

Oops. I can take a look into this and submit a fix upstream to vcpkg. I'll make a message here when that gets done.

PR #14244, opened 2021-03-06
Fix #14225: Desync when “allow early scenario completion” is enabled

### LINKED ISSUE #14225: Desync when "Enable early scenario completion" is checked on client and not on server.
Local:
**OS:** Windows 10
**Version:** 0.3.2
**Commit/Build:** `0b73781`

Server:
**OS:** Ubuntu Server
**Version:** 0.3.2
**Commit/Build:** `0.3.2-develop-0~2350-202103022302-g9b43ace64~ubuntu20.04.1`

Running into a desync where when our guest count reaches the scenario requirement, my local copy goes "wow congrats you've concluded your goal!" (even though we're not at the year requirement yet), but no one else's machine says so. (neither does the server) So I get a desync warning.

**Steps to reproduce:**
1. In options, assure that Enable early scenario completion is checked
image
2. Ensure server is not set to allow early completion.
3. Play a multiplayer game and reach early completion.

**Screenshots / Video:**
image

**Solution / Cause**
scenario.cpp
Any object check uses this boolean to determine whether a scenario is completed:
`gConfigGeneral.allow_early_completion`

The most reasonable solution will likely be to synchronize and override local settings while connected to a multiplayer server.

Gymnasiast:


duncanspumpkin:
So this is fixing it by disabling it in multiplayer. Is there any way we can make it a multiplayer configured option?

Gymnasiast:
I don't know enough about multiplayer to tell.

ZehMatt:
If this variable can't be changed while playing then theres already code that synchronies certain configurations in ```NetworkBase::SaveMap```, adding the variable there and in ```NetworkBase::LoadMap``` would be all thats needed.

duncanspumpkin:
Hmm think it is just a normal option so on second thoughts its probably best to disable in multiplayer.

PR #14252, opened 2021-03-06
Add plugin API for vehicle g-forces

Phelicks:
I'm writing a plugin which displays live ride information and this API would be really helpful.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -115,6 +115,14 @@ declare global {
     }
 
     /**
+     * Represents lateral and vertical g-forces.
+     */
+    interface GForces {
+        lateralG: number;
+        verticalG: number;
+    }
+
+    /**
      * Represents information about the plugin such as type, name, author and version.
      * It also includes the entry point.
      */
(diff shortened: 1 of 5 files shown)


distribution/openrct2.d.ts
```diff
@@ -1153,6 +1161,11 @@ declare global {
          */
         trackLocation: CoordsXYZD;
 
+        /**
+         * The current g-forces of this car.
+         */
+        gForces: GForces;
```
IntelOrca, line 1167: This should be `readonly` as you have not provided a setter.

IntelOrca:
Thanks, finally can you increment the API version in Scripting.cpp and include a line in the changelog referencing the pull request number. We usually do this for each plugin API change.

Phelicks:
done :)

Gymnasiast:
I'm a bit apprehensive about merging this so close to the release, especially since we have had crashes after adding plug-in APIs before.

IntelOrca (approved):

IntelOrca:
I think this is safe enough

duncanspumpkin (approved):

Gymnasiast (approved):

PR #14274, opened 2021-03-08
Update objects to v1.0.21

Gymnasiast:
This update mostly concerns two things: translations, and using the new JSON ids for scenery group instead of the legacy ones (so `rct2.scgwalls` instead of `SCGWALLS`). This had to be accompanied by some changes in the code to be able to handle both types.

@tellovishous @rzhao271 It looks like Esperanto is updated correctly: 
afbeelding

### DIFF
--- a/src/openrct2/object/SceneryObject.h
+++ b/src/openrct2/object/SceneryObject.h
@@ -25,7 +25,7 @@ public:
     }
     virtual ~SceneryObject() = default;
 
-    const ObjectEntryDescriptor& GetPrimarySceneryGroup()
+    const ObjectEntryDescriptor& GetPrimarySceneryGroup() const
     {
         return _primarySceneryGroupEntry;
     }
--- a/src/openrct2/object/SceneryObject.h
+++ b/src/openrct2/object/SceneryObject.h
@@ -16,7 +16,7 @@
 class SceneryObject : public Object
 {
 private:
-    rct_object_entry _primarySceneryGroupEntry = {};
+    ObjectEntryDescriptor _primarySceneryGroupEntry = {};
 
 public:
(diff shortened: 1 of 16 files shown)


src/openrct2/object/SceneryObject.h
```diff
@@ -25,15 +25,14 @@ class SceneryObject : public Object
     }
     virtual ~SceneryObject() = default;
 
-    const rct_object_entry* GetPrimarySceneryGroup()
+    const ObjectEntryDescriptor& GetPrimarySceneryGroup()
```
tupaschoal, line 28:
```suggestion
    const ObjectEntryDescriptor& GetPrimarySceneryGroup() const
```

Method should probably be `const` as well

src/openrct2/object/SmallSceneryObject.cpp
```diff
@@ -193,34 +193,34 @@ void SmallSceneryObject::PerformFixes()
...
 {
-    return Object::CreateHeader("SCGMINE ", 207140231, 3638141733);
+    return ObjectEntryDescriptor("rct2.scgpirat");
 }
 
-rct_object_entry SmallSceneryObject::GetScgAbstrHeader()
+ObjectEntryDescriptor SmallSceneryObject::GetScgAbstrHeader()
 {
-    return Object::CreateHeader("SCGABSTR", 207140231, 932253451);
+    return ObjectEntryDescriptor("rct2.scgabstr");
```
tupaschoal, line 223: These getters should probably be `const` functions

(Dunno why but GitHub didn't allow me to suggest changes on each individually)

tupaschoal (approved):
This looks ok from a review standpoint. I haven't tested it manually though

PR #14277, opened 2021-03-08
Part of #12100: Use ScreenLine on gfx_draw_line_software

frutiemax:


src/openrct2/drawing/Line.cpp
```diff
@@ -159,3 +159,8 @@ void gfx_draw_line_software(rct_drawpixelinfo* dpi, int32_t x1, int32_t y1, int3
         }
     }
 }
+
+void gfx_draw_line_software(rct_drawpixelinfo* dpi, const ScreenLine& line, int32_t colour)
+{
+    gfx_draw_line_software(dpi, line.GetX1(), line.GetY1(), line.GetX2(), line.GetY2(), colour);
```
ZehMatt, line 165: This should be the other way around, the new function should not call the deprecated variant.

ZehMatt (changes requested):

tupaschoal (approved):

ZehMatt (approved):

Issue #14253, opened 2021-03-06
Replace font size formatting tokens with adjusted DrawText calls


Gymnasiast:
These were mostly made obsolete by https://github.com/OpenRCT2/OpenRCT2/pull/14189 .

This includes the tokens in the translation files and the support for these tokens in our 

However, some of the places that call these strings need to be adjusted to instead pass a FontSpriteBase parameter.

geoff-B:
I will give this one a go.

Gymnasiast:
Great!

You _might_ run into problems with guest thoughts, though. But let me know if that's the case.

geoff-B:
@Gymnasiast You were right, removing SMALLFONT in the localization files makes all the thoughts become medium font size.

Gymnasiast:
That's expected, you will need to modify some calls:

> However, some of the places that call these strings need to be adjusted to instead pass a FontSpriteBase parameter.

geoff-B:
TO DO:

- [ ] Graph labels (STR_RIDE_STATS, STR_FINANCES_FINANCIAL_GRAPH)
- [ ] STR_STRING_TOOLTIP
- [ ] STR_SMALL_WINDOW_COLOUR_2_STRINGID (5138)

geoff-B:
Pretty much done with all of them, except for some unused strings listed above that I don't know what to do with, and some non-standard cases:

Non-universal {SMALLFONT} calls (i.e. not used for all languages):

- [ ] STR_FIRE_STAFF_ID = 1711 (Russian) - **Crashes when {smallfont} is removed**
  ![image](https://user-images.githubusercontent.com/38046772/110274567-54376e80-7f8c-11eb-8794-0b6fbb9ff3be.png)

geoff-B:
@Gymnasiast What should be done with those two Russian strings and the strings that are not used in the code (1926, 2681 and 5301)?

Gymnasiast:
The Russian strings needed to use {WINDOW_COLOUR_1} instead of {SMALLFONT}. I just corrected that in the Localisation repository.

For the other strings: strings 1926 and 2681 can be removed, but I'd like to restore 5301. However, you can safely take off the {SMALLFONT}{BLACK} parts of that.

PR #14283, opened 2021-03-11
Part of #14253: Remove {SMALLFONT}, {MEDIUMFONT}, {BIGFONT} and {TINYFONT}

geoff-B:
As a part of #14253, this PR:
- Removes all references to {SMALLFONT}, {MEDIUMFONT}, {BIGFONT} and {TINYFONT} in the code.
- Adds appropriate FontSpriteBase parameter where needed to replace the use of these tokens.


duncanspumpkin:
Please update en-gb in this pr. The other localisations are correctly split into the localisation repo but en-gb should be part of this.

src/openrct2/localisation/StringIds.h
```diff
@@ -3055,7 +3055,7 @@ enum
     STR_NOTIFICATION_GUEST_DIED = 5606,
     STR_TILE_INSPECTOR_BASE_HEIGHT_SHORT = 5608,
     STR_TILE_INSPECTOR_CLEARANGE_HEIGHT_SHORT = 5609,
-    // STR_5610    :{SMALLFONT}{BLACK}Remove the currently selected map element. This will forcefully remove it, so no cash will
```
Gymnasiast, line 3058: Remove this comment altogether.

Gymnasiast:
I have tested it. It works as intended, but something popped up I didn't take into account when creating that issue and it's the text on top of this window (`Guests thinking “I need to go to the toilet”`):

afbeelding

afbeelding

Also, the format tokens might be useful to plugins.

Could you restore the format tokens for {SMALLFONT}, {MEDIUMFONT} and {TINYFONT}, and include a {SMALLFONT} token in STR_1821 (just before the STRING_ID)?

Gymnasiast (approved):

PR #14297, opened 2021-03-12
Allow early scenario completion in network play

### LINKED ISSUE #14296: Allow using early scenario completion in multiplayer
**OS:** Windows 10
**Version:** 0.3.2 develop
**Commit/Build:** f0dc28e

When playing a game in multiplayer, the option "allow early scenario completion",
doesn't work as expected.
It's turned on for me and all of my friends playing with me and despite that it won't trigger the scenario completion.

- [x ] Specific to multiplayer?

**Steps to reproduce:**
1. Launch OpenRCT2 on two different machines
2. Activate "allow early scenario completion" in the Options-Menu on both of these machines
3. Start a server on one of the machines (Forest Frontiers would be a prime candidate) and let the other one join
4. Reach the goal of the scenario before the deadline and wait for the day to advance to the next one

MaximilianBottin:
Fixes #14296

This Commit fixes the issues by sending the current config for the allow early scenario completion and having the clients use this sent value to compare against instead of their own configuration.

This approach has two issues though. 
First of all if the host changes the setting mid-game the clients won't recognize the change and thus a desync will occur.
   (Maybe it could be possible to always send the current setting through the network when the host changes it)
Second, the Clients are still able to change their allow early scenario completion setting, but it won't have any effect in multiplayer which can be confusing so maybe the option to change it during multiplayer should be disabled (would also fix the issue I mentioned above)


duncanspumpkin:
You need to add a game action / modify an existing game action to handle changes that happen mid-game.

MaximilianBottin:
Thanks, I'll look into that and add it to my commit

src/openrct2/scenario/Scenario.cpp
```diff
@@ -944,6 +942,19 @@ ObjectiveStatus Objective::CheckMonthlyFoodIncome() const
     return ObjectiveStatus::Undecided;
 }
 
+bool Objective::AllowEarlyCompletion() const
```
Gymnasiast, line 945: Doesn't belong in the objective struct. Please turn this into a regular function.

src/openrct2/scenario/Scenario.h
```diff
@@ -452,6 +452,7 @@ void load_from_sc6(const char* path);
 void scenario_begin();
 void scenario_update();
 bool scenario_create_ducks();
+bool allowEarlyCompletion();
```
Gymnasiast, line 455:
```suggestion
bool AllowEarlyCompletion();
```

Gymnasiast (changes requested):
Looks good to me, please address the line note and bump the network version and this can be merged AFAIC.

Gymnasiast (approved):

PR #14305, opened 2021-03-13
Part of 12100 : Use ScreenCoordsXY for gfx_draw_line_on_buffer

frutiemax:


src/openrct2/drawing/Line.cpp
```diff
@@ -16,44 +16,43 @@
...
-        no_pixels -= x + no_pixels - dpi->width;
+        no_pixels -= offset.x + no_pixels - dpi->width;
         if (no_pixels <= 0)
             return;
     }
 
     // Get the buffer we are drawing to and move to the first coordinate.
-    uint8_t* bits_pointer = dpi->bits + y * (dpi->pitch + dpi->width) + x;
+    uint8_t* bits_pointer = dpi->bits
+        + offset.y * (static_cast<int64_t>(static_cast<int64_t>(dpi->pitch) + static_cast<int64_t>(dpi->width))) + offset.x;
```
tupaschoal, line 55: Why int64_t?

src/openrct2/drawing/Line.cpp
```diff
@@ -141,14 +140,14 @@ void gfx_draw_line_software(rct_drawpixelinfo* dpi, const ScreenLine& line, int3
     {
         // Vertical lines are drawn 1 pixel at a time
         if (steep)
-            gfx_draw_line_on_buffer(dpi, colour, x, y, 1);
+            gfx_draw_line_on_buffer(dpi, colour, { y, x }, 1);
```
tupaschoal, line 143: Very weird that X and y were swapped, maybe needs further investigation in the future

src/openrct2/drawing/Line.cpp
frutiemax: image

src/openrct2/drawing/Line.cpp
frutiemax: I'm not too sure why it was swapped in the function definition...

src/openrct2/drawing/Line.cpp
tupaschoal:
```suggestion
        + offset.y * (static_cast<int64_t>(dpi->pitch + dpi->width)) + offset.x;
```
I think this should be enough

src/openrct2/drawing/Line.cpp
tupaschoal: Yeah, but this was the only place calling it swapped, which looks like an error, maybe just an oversight.

src/openrct2/drawing/Line.cpp
frutiemax: image

tupaschoal: Ok we can mark this one resolved as is.

src/openrct2/drawing/Line.cpp
tupaschoal: Can you try checking what the function is doing and the impact of swapping it to the supposedly correct state, please?

frutiemax: y,x
image

x,y
image

frutiemax: It's only used for the plots I believe.

tupaschoal (approved):

PR #14308, opened 2021-03-13
Part of 12113 : Create RideTypeDescriptors table sorted by view order

frutiemax:
- Create a RideTypeDescriptors table sorted by view order
- Use RideTypeDescriptors in the ObjectManager code
- Add a ID field to the RideTypeDescriptor. Right now, it is the same as the RideType value in the enumeration to keep compatibility with the old code. In the future, we could change the ID type to a unique string.
- More use of the RideTypeDescriptor in some other places


duncanspumpkin:
There is quite a lot happening in this pr can you make it just adding the ride type to the descriptor. Make the other changes in follow on prs. Id is not a great name for ride type. Couldn't you just use type?

Gymnasiast:
> 
> 
> There is quite a lot happening in this pr can you make it just adding the ride type to the descriptor. Make the other changes in follow on prs. Id is not a great name for ride type. Couldn't you just use type?

All the RIDE_TYPE_FOO constants are essentially IDs, so I can see why you'd name them that.

frutiemax:
I named it ID as there way no easy way to make an equality operator, and RideTypeDescriptor already describes a type of ride. I'm not sure if naming a type field inside a RideTypeDescriptor struct would be a good idea even if the IDs takes its values from the legacy ride type enum.

src/openrct2/ride/Ride.cpp
```diff
@@ -5340,11 +5340,11 @@ void Ride::SetNameToDefault()
 /**
  * This will return the name of the ride, as seen in the New Ride window.
  */
-RideNaming get_ride_naming(const uint8_t rideType, rct_ride_entry* rideEntry)
+RideNaming get_ride_naming(const RideTypeDescriptor& rideType, rct_ride_entry* rideEntry)
```
Gymnasiast, line 5343: This is a good refactor by itself, but it would be better to split off and merge first, to get a clearer diff.

frutiemax:
How do you split a big commit into smaller commits?

Broxzier:
Soft reset the branch to before the commit that you want to split (soft reset keeps all changes intact), then only stage the lines you want to commit and create a the new commits.

frutiemax:
Split the pull request into 2 commits.

tupaschoal (commented):
Can you reword each commit to add a # before the issue/pr number? Then it will create the proper links (this is very minor, but...)

frutiemax:
image
Still have some problem with the transport tab...

frutiemax:
I'm not even sure if adding a ID field is a good idea anymore since a ride object is not the same as a ride type, there is a need of discussion around this.

PR #14326, opened 2021-03-15
Split Entity Lists by Entity type

duncanspumpkin:
Finally after lots of work we can split the entity lists to be type specific.
This should not break replays.
After this is merged I will then break replays by removing the `spriteIdentifier` field and the `MiscEntity::SubType` fields replacing with the `EntityType` field.


src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1242,19 +1242,12 @@ static int32_t cc_show_limits(InteractiveConsole& console, [[maybe_unused]] cons
 
     int32_t rideCount = ride_get_count();
     int32_t spriteCount = 0;
-    for (int32_t i = 1; i < static_cast<uint8_t>(EntityListId::Count); ++i)
+    for (int32_t i = 1; i < static_cast<uint8_t>(EntityType::Count); ++i)
```
duncanspumpkin, line 1245: should be from 0

PR #14328, opened 2021-03-15
Fix #14315: Crash when renaming APVC in Korean

### LINKED ISSUE #14315: Crash if try to rename Air Powered Vertical Coaster in CJK
**OS:** Windows 10
**Version:** 0.3.3
**Commit/Build:** 72b6160

When I try to rename Air Powered Vertical Coaster(``에어 파워드 버티컬 코스터``) in CJK language, it crashes.
Maybe @AaronVanGeffen can reproduce it.

It is possible if I (1) change language into English, (2) rename it (3) and re-change into Korean again

**Steps to reproduce:**
1. Change language into CJK, in my case, Korean.
2. Build an Air Powered Vertical Coaster as default name
3. Check that its name is default one (eg. ``에어 파워드 버티컬 코스터 1`` in Korean)
3. Try to rename it

Gymnasiast:


tupaschoal (commented):
Seems a bit wasteful to allocate and deallocate dynamic memory every time a text is set (?)

How do we do in other places that copy text?

Gymnasiast:
> Seems a bit wasteful to allocate and deallocate dynamic memory every time a text is set (?)

It only happens when opening the Text Input window.

duncanspumpkin:
This looks like very ugly code either way. It's allocating twice for temp buffers. Do we not have a safe resize function instead of this.

Gymnasiast:
> Do we not have a safe resize function instead of this.

I don't think so. We don't even have a safe_strcpy() for `std::string`. A lot of our string handling code still seems to assume that 1 character = 1 byte, which isn't even true for English.

IntelOrca:
> Do we not have a safe resize function instead of this.

We should make one, I didn't realise `safe_strcpy` handled clever truncation of codepoints. This is fine for now though if we want to quickly fix the bug.

Gymnasiast:
> > Do we not have a safe resize function instead of this.
> 
> We should make one, I didn't realise `safe_strcpy` handled clever truncation of codepoints. This is fine for now though if we want to quickly fix the bug.

You added that: https://github.com/OpenRCT2/OpenRCT2/commit/1daa6576575f9a1fbef321236c0368396b5ff3df

duncanspumpkin (approved):

PR #14331, opened 2021-03-15
Fix #14330: join_server uses default_port from config

### LINKED ISSUE #14330: join_server uses default_port from config
**OS:** Windows 10 / Manjaro Linux
**Version:** 0.3.3
**Commit/Build:** 4ac3fd4

The config property `default_port` does exactly what the name implies and changes the default port to use when no port is specified when connecting. It's also the default port used when starting up a multiplayer server. I don't believe the former is what the user would expect. So while I admit that this is probably a case of being a feature not a bug, I would still like to propose a change to its use.

I discovered this functionality not by looking through the config file or the codebase, but by failing to connect to a local server that I added to my server list. I was perplexed as to why I could connect to it using the local discovery entry in the browser, but not the server I entered manually. Then I opened the Start Server window and realized that the last server I ran was on a port other than 11753. Only then did I finally realize that the server port is the same port used for connecting to hosts that don't have a port specified.

Joining a server without a port specified should always default to 11753. `default_port` should only be used when spinning up a server.

**Steps to reproduce:**
1. Start a server on the standard port 11753
2. On a separate client, change the default_port to something else (this can be done through the Start Server screen)
3. On this second client, add the server to the server browser without specifying a port
4. Try to connect

CorySanin:
closes #14330

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -5,6 +5,7 @@
 - Fix: [#11829] Visual glitches and crashes when using RCT1 assets from mismatched or corrupt CSG1.DAT and CSG1i.DAT files.
 - Fix: [#13894] Block brakes do not animate.
 - Fix: [#14315] Crash when trying to rename Air Powered Vertical Coaster in Korean.
+- Fix: [#14330] join_server uses default_port from config.
 
 0.3.3 (2021-03-13)
 ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


distribution/changelog.txt
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#14071] “Vandals stopped” statistic for security guards.
 - Fix: [#13894] Block brakes do not animate.
+- Fix: [#14330] join_server uses default_port from config
```
Gymnasiast, line 5:
```suggestion
- Fix: [#14330] join_server uses default_port from config.
```

Gymnasiast:
This needs a rebase.

CorySanin:
@Gymnasiast rebased! 👍

IntelOrca (approved):

PR #14343, opened 2021-03-17
Use entity list size rather than distance

duncanspumpkin:
After the refactors with the entity lists we no longer require to walk the whole entity list to get the number of entries.


janisozaur:
Nice! Any more such spots?

duncanspumpkin:
I searched for all instances of std distance looking for this specific type of thing. Didn't find any. But there may be more that didn't use std distance

PR #14344, opened 2021-03-17
Entity type4: Remove MiscEntityType and add FountainType

duncanspumpkin:
Final few bits of reorg of the fundamental setup of entities. Removal of SubType as we can just use Type but for jumping fountains we needed a way to distinguish between snow and water so had to add a field for that.

### DIFF
--- a/src/openrct2/paint/sprite/Paint.Misc.cpp
+++ b/src/openrct2/paint/sprite/Paint.Misc.cpp
@@ -113,7 +113,7 @@ template<> void PaintEntity(paint_session* session, const JumpingFountain* jumpi
         isAntiClockwise = !isAntiClockwise;
     }
 
-    uint32_t baseImageId = (jumpingFountain->SubType == MiscEntityType::JumpingFountainSnow) ? 23037 : 22973;
+    uint32_t baseImageId = (jumpingFountain->FountainType == JumpingFountainType::Snow) ? 23037 : 22973;
     uint32_t imageId = baseImageId + ebx * 16 + jumpingFountain->frame;
     constexpr std::array<CoordsXY, 2> antiClockWiseBoundingBoxes = { CoordsXY{ -COORDS_XY_STEP, -3 }, CoordsXY{ 0, -3 } };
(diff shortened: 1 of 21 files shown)


src/openrct2/world/Balloon.cpp
```diff
@@ -81,11 +81,7 @@ void Balloon::Pop()
...
 void create_balloon(const CoordsXYZ& balloonPos, int32_t colour, bool isPopped)
 {
-    auto* sprite = CreateEntity<Balloon>();
-    if (sprite == nullptr)
-        return;
-    sprite->SubType = MiscEntityType::Balloon;
-    auto balloon = sprite->As<Balloon>();
+    auto* balloon = CreateEntity<Balloon>();
     if (balloon == nullptr)
         return; // can never happen
```
Gymnasiast, line 86: I think this comment is now no longer correct, right?

src/openrct2/paint/sprite/Paint.Misc.cpp
```diff
@@ -113,7 +113,7 @@ template<> void PaintEntity(paint_session* session, const JumpingFountain* jumpi
         isAntiClockwise = !isAntiClockwise;
     }
 
-    uint32_t baseImageId = (jumpingFountain->SubType == MiscEntityType::JumpingFountainSnow) ? 23037 : 22973;
+    uint32_t baseImageId = (jumpingFountain->FountainType == JumpingFountainType::Snow) ? 23037 : 22973;
```
Gymnasiast, line 116: Would like to have constants for these images.

duncanspumpkin:
`JUMPING_FOUNTAIN_TYPE_SNOW` no longer exists it was the basis for the new enum class.

duncanspumpkin:
Requires another network version bump prior to merge

Gymnasiast (commented):
Code looks good, would like to do a little test.

Gymnasiast (approved):
Appears to work correctly.

Gymnasiast:
Mind the milestone.

PR #14351, opened 2021-03-19
Fix #13581: Opening the Options menu kills performance

### LINKED ISSUE #13581: Opening the Options Menu kills performance
**OS:** Arch Linux
**Version:** 0.3.2-198
**Commit/Build:** 9ef493198

Opening the Options Menu, regardless of Rendering option, cause frame rates to drop to nearly 1/10th of expected performance.

**Steps to reproduce:**
1. Observe FPS
2. Open Options Menu
3. Observe again

Gymnasiast:


src/openrct2-ui/UiContext.Linux.cpp
```diff
@@ -251,10 +253,15 @@ namespace OpenRCT2::Ui
             return result;
         }
 
-        bool HasFilePicker() const override
+        bool HasFilePicker() override
```
duncanspumpkin, line 256: Think you would normally mark the cache as mutable rather than change the const of the function

Gymnasiast: Sounds better, how can I do that?

duncanspumpkin: https://en.cppreference.com/w/cpp/language/cv 
Just add mutable before the `std::optional<bool> _hasFilePicker = std::nullopt;`

src/openrct2-ui/windows/Options.cpp
```diff
@@ -2280,6 +2281,7 @@ rct_window* window_options_open()
     w->page = WINDOW_OPTIONS_PAGE_DISPLAY;
     w->frame_no = 0;
     WindowInitScrollWidgets(w);
+    _hasFilePicker = OpenRCT2::GetContext()->GetUiContext()->HasFilePicker();
```
IntelOrca, line 2284: You can probably revert the changes to this file now as you cache in the context instead.

Gymnasiast: Wouldn't that increase the overhead again?

duncanspumpkin: The overhead is in the call that you have now cached. So this is going to be pretty cheap to call

src/openrct2-ui/UiContext.Linux.cpp
```diff
@@ -35,6 +35,8 @@ namespace OpenRCT2::Ui
     class LinuxContext final : public IPlatformUiContext
     {
     private:
+        mutable std::optional<bool> _hasFilePicker = std::nullopt;
```
AaronVanGeffen, line 38: If no other methods are using this, you could make this a static variable in the HasFilePicker method instead.

IntelOrca: That would break a few principals.

AaronVanGeffen: How so? What am I missing?

IntelOrca: Lets say you are writing a test for this class, you would get state shared between one test and the next.

AaronVanGeffen: Right, and that wouldn't happen for properties. Gotcha. Thanks.

PR #14355, opened 2021-03-19
Fix #14353. Provide a buffer for text input

### LINKED ISSUE #14315: Crash if try to rename Air Powered Vertical Coaster in CJK

### LINKED ISSUE #14353: Crash when typing too long ride/stall name
**OS:** Windows 10
**Version:** 0.3.3
**Commit/Build:** a018787

When trying to change the name of a ride/stall the game will crash if the name gets too long.
The crash occurs the moment the letter a letter is typed.
For example when naming a ride "Lorem Ipsum Lorem Ipsum " the game will crash the moment
I try to moment I type the 'L' of the next Lorem.

However, it won't happen with some rides or stalls like the classic mini coaster (sorry I had to improvie the translation I only know the german name "Klassiche Miniachterbahn")

**Steps to reproduce:**
1.  Start a new game in the electric fields scenario
2. Change the name of the Balloon Stall if enough letters are typed in the game will crash

**Dump file**
[556a97fd-da55-4991-976f-eb5faaaff7b4(a018787_x86-64).zip](https://github.com/OpenRCT2/OpenRCT2/files/6172243/556a97fd-da55-4991-976f-eb5faaaff7b4.a018787_x86-64.zip)

duncanspumpkin:
After the text input window fix #14315 the buffer that was being provided for text input was only the size of the initial text.

### DIFF
--- a/src/openrct2-ui/windows/TextInput.cpp
+++ b/src/openrct2-ui/windows/TextInput.cpp
@@ -250,7 +250,7 @@ public:
                 cursorY = screenCoords.y;
 
                 int32_t textWidth = 6;
-                if (gTextInput->SelectionStart < strnlen_s(_buffer.data(), _buffer.size()))
+                if (gTextInput->SelectionStart < strlen(_buffer.data()))
                 {
                     // Make a 1 utf8-character wide string for measuring the width
                     // of the currently selected character.
--- a/src/openrct2-ui/windows/TextInput.cpp
+++ b/src/openrct2-ui/windows/TextInput.cpp
@@ -53,7 +53,7 @@ private:
 
     int32_t _cursorBlink{};
     size_t _maxInputLength{};
-    std::string _buffer;
+    std::vector<utf8> _buffer;
 
 public:
     void OnOpen() override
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
I've switched this to use a std::vector for the buffer this is for two reasons. A std::string does not at first glance look like a buffer. Adding characters to a std::string without using the std::string interface would be undefined. You would end up with the `length()` function possibly not returning the length of the string.

duncanspumpkin:
Its debatable if this line has any reason to exist. `if (gTextInput->SelectionStart < strlen(_buffer.data()))` the gTextInput will ensure that the selection start is at the right position gTextInput also has a Size which would avoid the need to do strlen. I've not made that change in this pr though.

Gymnasiast (commented):
Can't get it to crash with either Dutch or Korean.

PR #14384, opened 2021-03-24
Use entity list rather than MAX_ENTITIES loop

duncanspumpkin:
Small little cleanup reducing the number of whole entity list iterators.


src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -1580,17 +1580,11 @@ static int32_t cc_mp_desync(InteractiveConsole& console, const arguments_t& argv
         desyncType = atoi(argv[0].c_str());
     }
 
-    std::vector<Peep*> peeps;
+    std::vector<Guest*> peeps;
```
Gymnasiast, line 1583: Let's rename the variable to `guests` as well.

PR #14391, opened 2021-03-26
Close #13407: Allow building covered lift hill chain when cheat on

### LINKED ISSUE #13407: Cannot put chain lift on covered dinghy slide track.
**OS:** Windows 10 
**Version:**  0.3.2
**Commit/Build:** d61e81b

**Explanation of the issue**
When constructing a ride using dinghy slide track, selecting either covered track or chain lift deselects the other option, even with the "Allow chain lifts on all track pieces" cheat enabled. It was previously possible to create such a piece (covered track with chain lift) in vanilla using 8cars, and is also possible to achieve this through the tile inspector, but not through the ride creation menu

**Steps to reproduce:**
1. Activate "Allow chain lifts on all track pieces" cheat
2. Build a dinghy slide
3. Select "covered" track
4. Select chain lift

DKarandikar:
This seems to work, however I'm a little concerned touching these two files, there's a lot going on here. 

In RideConstruction don't make lift hill or O_track interact when they're clicked and the cheat is enabled. I.e. if the cheat is on, and you click lift chain don't switch to U_track and if you click O_track don't turn chain off. 

In _legacy.cpp ensure that this then actually gets built: if the cheat is on and we're building the alternativeType then don't turn off lift hill.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,6 +1,7 @@
 0.3.4.1+ (in development)
 ------------------------------------------------------------------------
 - Feature: [#3868] Initial support for using TTF in OpenGL mode (still contains bugs).
+- Feature: [#13407] Allow building chain lifts on enclosed dinghy slide pieces when cheats are on.
 - Feature: [#15084] [Plugin] Add "vehicle.crash" hook.
 - Feature: [#15143] Added a shortcut key for Giant Screenshot.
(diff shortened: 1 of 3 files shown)


Gymnasiast (commented):
The code looks reasonable, but the construction window being what it is (i.e. a complex labyrinth of code having unpredictable behaviour), I'd like to test it as well.

Gymnasiast (approved):

PR #14396, opened 2021-03-27
Apply changed strings from the new-save-format branch

Gymnasiast:
Discussed in the new-save-format meeting today. Avoids having to rebase the new-save-format branch too often.


tupaschoal:
Should this become an issue in Localisation?

Gymnasiast:
Good point. Opened https://github.com/OpenRCT2/Localisation/issues/2165.

PR #14397, opened 2021-03-27
Remove hardcoded terrain style constants from RCT1 and MapGen

Gymnasiast:


src/openrct2-ui/windows/MapGen.cpp
```diff
@@ -421,8 +421,8 @@ static void window_mapgen_draw_tab_images(rct_drawpixelinfo* dpi, rct_window* w)
 static int32_t _mapSize = 150;
 static int32_t _baseHeight = 12;
 static int32_t _waterLevel = 6;
-static int32_t _floorTexture = TERRAIN_GRASS;
-static int32_t _wallTexture = TERRAIN_EDGE_ROCK;
+static int32_t _floorTexture = 0;
+static int32_t _wallTexture = 0;
```
tupaschoal, line 425: Sounds a bit weird to drop the define in this case (?)

src/openrct2/object/Object.h
```diff
@@ -180,6 +181,8 @@ struct ObjectEntryDescriptor
         Generation = ObjectGeneration::JSON;
         Identifier = std::string(newIdentifier);
     }
+
+    explicit ObjectEntryDescriptor(const ObjectRepositoryItem& ori);
```
tupaschoal, line 185: All the other constructors are declared and defined in the header, I'd move their definition to the source file in the future.

src/openrct2/object/TerrainEdgeObject.cpp
```diff
@@ -58,3 +60,15 @@ void TerrainEdgeObject::ReadJson(IReadObjectContext* context, json_t& root)
...
+TerrainEdgeObject* TerrainEdgeObject::GetById(ObjectEntryIndex entryIndex)
+{
+    TerrainEdgeObject* result = nullptr;
+    auto& objMgr = OpenRCT2::GetContext()->GetObjectManager();
+    auto obj = objMgr.GetLoadedObject(ObjectType::TerrainSurface, entryIndex);
+    if (obj != nullptr)
+    {
+        result = static_cast<TerrainEdgeObject*>(obj);
+    }
+    return result;
```
tupaschoal, line 73: You can make this:
```suggestion
    return obj != nullptr ? static_cast<TerrainEdgeObject*>(obj) : nullptr;
```

Then you wouldn't need the `result` variable.

src/openrct2/object/TerrainSurfaceObject.cpp
```diff
@@ -145,3 +147,15 @@ uint32_t TerrainSurfaceObject::GetImageId(
...
+TerrainSurfaceObject* TerrainSurfaceObject::GetById(ObjectEntryIndex entryIndex)
+{
+    TerrainSurfaceObject* result = nullptr;
+    auto& objMgr = OpenRCT2::GetContext()->GetObjectManager();
+    auto obj = objMgr.GetLoadedObject(ObjectType::TerrainSurface, entryIndex);
+    if (obj != nullptr)
+    {
+        result = static_cast<TerrainSurfaceObject*>(obj);
+    }
+    return result;
```
tupaschoal, line 160: Same comment.

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -1279,8 +1279,8 @@ void surface_paint(paint_session* session, uint8_t direction, uint16_t height, c
 
     if (!(session->ViewFlags & VIEWPORT_FLAG_HIDE_VERTICAL))
     {
-        const uint32_t edgeStyle = tileElement->AsSurface()->GetEdgeStyle();
-        if (edgeStyle >= TERRAIN_EDGE_COUNT)
+        const auto edgeStyle = tileElement->AsSurface()->GetEdgeStyle();
+        if (static_cast<int32_t>(edgeStyle) >= object_entry_group_counts[EnumValue(ObjectType::TerrainEdge)])
```
tupaschoal, line 1283: Should we use `EnumValue` for `edgeStyle` instead?

src/openrct2/world/MapGen.cpp
```diff
@@ -82,7 +85,8 @@ static constexpr const char* SnowTrees[] = {
 #pragma endregion
 
 // Randomly chosen base terrains. We rarely want a whole map made out of chequerboard or rock.
-static constexpr const uint8_t BaseTerrain[] = { TERRAIN_GRASS, TERRAIN_SAND, TERRAIN_SAND_LIGHT, TERRAIN_DIRT, TERRAIN_ICE };
+static constexpr const std::string_view BaseTerrain[] = { "rct2.surface.grass", "rct2.surface.sand", "rct2.surface.sandbrown",
```
tupaschoal, line 88: Any reason for making this one a `string_view` array? All the others were `char*` array. Just a comment for consistency, I actually don't know which would be better.

src/openrct2/world/MapGen.cpp
```diff
@@ -134,46 +138,43 @@ void mapgen_generate_blank(mapgen_settings* settings)
...
-    mapSize = settings->mapSize;
-    floorTexture = settings->floor;
-    wallTexture = settings->wall;
-    waterLevel = settings->water_level;
-
-    if (floorTexture == -1)
+    auto mapSize = settings->mapSize;
+    auto waterLevel = settings->water_level;
+    const auto selectedFloor = TerrainSurfaceObject::GetById(settings->floor);
+    std::string floorTexture = selectedFloor != nullptr ? std::string(selectedFloor->GetIdentifier()) : "";
```
tupaschoal, line 144:
```suggestion
    std::string floorTexture = selectedFloor != nullptr ? std::string(selectedFloor->GetIdentifier()) : std::string();
```
I'd personally prefer doing this, but not everyone likes it

src/openrct2/world/MapGen.cpp
```diff
@@ -134,46 +138,43 @@ void mapgen_generate_blank(mapgen_settings* settings)
...
+    const auto selectedEdge = TerrainEdgeObject::GetById(settings->wall);
+    std::string edgeTexture = selectedFloor != nullptr ? std::string(selectedEdge->GetIdentifier()) : "";
+
+    if (floorTexture == "")
```
tupaschoal, line 148:
```suggestion
    if (floorTexture.empty())
```

src/openrct2/world/MapGen.cpp
```diff
@@ -134,46 +138,43 @@ void mapgen_generate_blank(mapgen_settings* settings)
...
         floorTexture = BaseTerrain[util_rand() % std::size(BaseTerrain)];
 
-    if (wallTexture == -1)
+    if (edgeTexture == "")
```
tupaschoal, line 151:
```suggestion
    if (edgeTexture.empty())
```

src/openrct2/object/Object.cpp
```diff
@@ -301,6 +302,20 @@ std::unique_ptr<IStream> ObjectAsset::GetStream() const
     return {};
 }
 
+ObjectEntryDescriptor::ObjectEntryDescriptor(const ObjectRepositoryItem& ori)
+{
+    if (ori.Identifier != "")
```
tupaschoal, line 307: If `Identifier` is a string, this should be preferably:
```suggestion
    if (!ori.Identifier.empty())
```

src/openrct2/paint/tile_element/Paint.Surface.cpp
Gymnasiast: ???

tupaschoal: Yeah I mean instead of the static_cast using our utility function called `EnumValue` or something like it

Gymnasiast: edgeStyle is not an enum value (not anymore), so there is no reason to do that.

src/openrct2-ui/windows/MapGen.cpp
Gymnasiast: No. There is no guarantee at all that index 0 will be grass and rock. It would be like assuming that ride entry index 0 is always a haunted house.

src/openrct2/world/MapGen.cpp
Gymnasiast: ... why? It's probably a matter of taste, but I find `== ""` to be much more recognisable.

duncanspumpkin: Although i doubt it makes any difference .empty() will likely be faster to compute.

src/openrct2/world/MapGen.cpp
Gymnasiast: I'll make everything string_view.

src/openrct2/object/TerrainEdgeObject.cpp
```diff
@@ -58,3 +60,10 @@ void TerrainEdgeObject::ReadJson(IReadObjectContext* context, json_t& root)
     PopulateTablesFromJson(context, root);
     NumImagesLoaded = GetImageTable().GetCount();
 }
+
+TerrainEdgeObject* TerrainEdgeObject::GetById(ObjectEntryIndex entryIndex)
+{
+    auto& objMgr = OpenRCT2::GetContext()->GetObjectManager();
+    auto obj = objMgr.GetLoadedObject(ObjectType::TerrainSurface, entryIndex);
```
Broxzier, line 67: This should've used `ObjectType::TerrainEdge`

Gymnasiast: Right, I guess this is (partly) responsible for the mapgen crashes?

Broxzier: Partly indeed, the main issue was that the map generator would use surfaces and edges that are not necessarily available. The fallback to the first available edge in the list would now select a surface for the edge instead.

PR #14398, opened 2021-03-27
Reorganise EntityImport for S4 to follow S6/park style

duncanspumpkin:
Will need some testing. I've stripped out a lot of fixing code so this should be a lot quicker to perform.


Gymnasiast (commented):
So far it seems to be holding up quite nicely. I do want to test a little bit more though.

Gymnasiast (approved):

PR #14410, opened 2021-03-30
Add missing call to gfx_unload_csg()

Gymnasiast:


duncanspumpkin:
Is there an associated issue or just something you noticed

Gymnasiast:
Just something I noticed. Technically, it's a memory leak, but according to Ted, Context is only destroyed when the game is exiting.

duncanspumpkin (approved):

PR #14411, opened 2021-03-30
Use 16-bit ride_id_t thoughout and raise RIDE_ID_NULL

Gymnasiast:
Needed to raise the maximum amount of rides in a park.

For reviewing, it is advised to look at the commits one by one. The first is a simple find-and-replace rename.

### DIFF
--- a/src/openrct2/ride/RideTypes.h
+++ b/src/openrct2/ride/RideTypes.h
@@ -17,8 +17,7 @@
 using ride_id_t = uint16_t;
 struct Ride;
 
-constexpr const ride_id_t RIDE_ID_NULL = 0xFF; // std::numeric_limits<ride_id_t>::max();
-constexpr const ride_id_t RideIdNewNull = std::numeric_limits<ride_id_t>::max();
+constexpr const ride_id_t RIDE_ID_NULL = std::numeric_limits<ride_id_t>::max();
 
 /**
  * Couples a ride type and subtype together.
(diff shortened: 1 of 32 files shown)


duncanspumpkin:
I take it the failure in the replays is due to different values for null? All looks fine. Do you want me to investigate the replay and rerecord them?

Gymnasiast:
> I take it the failure in the replays is due to different values for null? 

That's indeed what I expect.

> Do you want me to investigate the replay and rerecord them?

Yes, please. :+1:

src/openrct2/ride/RideTypes.h
```diff
@@ -17,8 +17,7 @@
 using ride_id_t = uint16_t;
 struct Ride;
 
-constexpr const ride_id_t RIDE_ID_NULL = 0xFF; // std::numeric_limits<ride_id_t>::max();
-constexpr const ride_id_t RideIdNewNull = std::numeric_limits<ride_id_t>::max();
+constexpr const ride_id_t RIDE_ID_NULL = std::numeric_limits<ride_id_t>::max();
```
duncanspumpkin, line 20: Would it be possible to change this to `RideIdNull` or something not in all caps.

Gymnasiast: We seem to use a mishmash of TitleCase and SCREAMING_SNAKE_CASE for constants. Would be good to have an agreed-upon standard. (My preference happens to go towards the latter to make clear it's a constant.)

PR #14413, opened 2021-03-31
Change more rct_object_entry to ObjectEntryDescriptor

IntelOrca:
Cherry-picked mostly from `new-save-format`

### DIFF
--- a/src/openrct2-ui/windows/ObjectLoadError.cpp
+++ b/src/openrct2-ui/windows/ObjectLoadError.cpp
@@ -20,7 +20,9 @@
 #include <openrct2/object/ObjectManager.h>
 #include <openrct2/object/ObjectRepository.h>
 #include <openrct2/platform/platform.h>
+#include <openrct2/ui/UiContext.h>
 #include <openrct2/windows/Intent.h>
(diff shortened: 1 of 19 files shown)


src/openrct2-ui/windows/ObjectLoadError.cpp
```diff
@@ -199,7 +201,7 @@ class ObjectDownloader
         }
 
         auto& entry = _entries[_currentDownloadIndex];
-        auto name = String::Trim(std::string(entry.name, sizeof(entry.name)));
+        auto name = String::Trim(std::string(entry.GetName()));
```
Gymnasiast, line 204: This assumes that the object download server recognises json-style object names. We need to make sure this is the case before the new save format is merged.

IntelOrca: Yes, I plan for it to use a different API for JSON object names.

Gymnasiast:
Why did you close this?

IntelOrca:
It is too out of date now with new-save-format branch. I have decided it is not worth the effort to backport.

PR #14421, opened 2021-04-02
Use correct null value

duncanspumpkin:
A few little errors from #14411


duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/blob/7344b7fbfdb6ec477461ca3a22df05945f640996/src/openrct2/world/Footpath.cpp#L57

PR #14425, opened 2021-04-02
Update rating skipping unused ride ids

duncanspumpkin:
After the change to the ride null value we know iterate from 0 -> 65535 when updating ratings. This is slightly wasteful when we only have 0-254 ride id's in use. I've changed the code so that gRideRatingsCalcData.CurrentRide isn't actually the current ride but the distance from ridemanager.begin. I don't quite know what to do about names and this all looks super hacky but it does resolve the issue.

### DIFF
--- a/src/openrct2/ride/RideRatings.cpp
+++ b/src/openrct2/ride/RideRatings.cpp
@@ -178,8 +178,8 @@ static RideId GetNextRideToUpdate(RideId currentRide)
 static void ride_ratings_update_state_0(RideRatingUpdateState& state)
 {
     // It is possible that the current ride being calculated has
-    // been removed or due to import is invalid. For both reset
-    // ratings check at the start
+    // been removed or due to import invalid. For both, reset
+    // ratings and start check at the start
     if (get_ride(state.CurrentRide) == nullptr)
     {
         state.CurrentRide = {};
--- a/src/openrct2/ride/RideRatings.cpp
+++ b/src/openrct2/ride/RideRatings.cpp
@@ -185,13 +185,14 @@ static void ride_ratings_update_state_0(RideRatingUpdateState& state)
         state.CurrentRide = {};
     }
 
-    auto nextRide = get_ride(GetNextRideToUpdate(state.CurrentRide));
+    auto nextRideId = GetNextRideToUpdate(state.CurrentRide);
+    auto nextRide = get_ride(nextRideId);
     if (nextRide != nullptr && nextRide->status != RideStatus::Closed
         && !(nextRide->lifecycle_flags & RIDE_LIFECYCLE_FIXED_RATINGS))
     {
         state.State = RIDE_RATINGS_STATE_INITIALISE;
     }
-    state.CurrentRide = nextRide->id;
+    state.CurrentRide = nextRideId;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


duncanspumpkin:
Might need recalculation on import as well. Which would also make for a good time to rename the variable.

Gymnasiast:
Why not just check for >= MAX_RIDES?

duncanspumpkin:
That's not a great solution post new save format and the ride ids could become more fragmented

Gymnasiast (commented):
Code looks good to me (as far as I can tell). Would like to test.

Gymnasiast (approved):

src/openrct2/ride/RideRatings.cpp
```diff
@@ -158,13 +158,28 @@ static void ride_ratings_update_state(RideRatingUpdateState& state)
 static void ride_ratings_update_state_0(RideRatingUpdateState& state)
 {
     ride_id_t currentRide = state.CurrentRide;
-
-    currentRide++;
-    if (currentRide >= MAX_RIDES)
+    // It is possible that the existing ride being calculated has
+    // been removed or due to import is invalid. For both reset
+    // ratings check at the start
```
tupaschoal, line 163:
```suggestion
    // been removed or due to invalid import. For both, reset
    // ratings check at the start
```

src/openrct2/ride/RideRatings.cpp
```diff
@@ -158,13 +158,28 @@ static void ride_ratings_update_state(RideRatingUpdateState& state)
 static void ride_ratings_update_state_0(RideRatingUpdateState& state)
 {
     ride_id_t currentRide = state.CurrentRide;
-
-    currentRide++;
-    if (currentRide >= MAX_RIDES)
+    // It is possible that the existing ride being calculated has
+    // been removed or due to invalid import. For both, reset
+    // ratings check at the start
+    if (get_ride(currentRide) == nullptr)
```
Gymnasiast, line 164: This could potentially cause problems if the user has a lot of rides, as the ones with low IDs will be updated more frequently if the counter resets. Would it be possible to just change this to go to the next available ID instead?

duncanspumpkin: In this hypothetical the user has to be deleting specific rides at specific timings during their update. That's not really realistic. You can't even spam create and delete as you would need to wait for ride measurements. I just can't see how this could ever be an issue. I can look see how the iterator handles the null but it seems overkill.

IntelOrca: You only lose one tick per unoccupied ride slot, although I guess that adds up. Might be good to move this to a function called, `GetNextRideToCalculate`

IntelOrca:
@duncanspumpkin For this PR, my suggestion was to extract the code into a new method called `GetNextRideToCalculate` or something. Otherwise, it just needs rebase and network update so that it can be merged.

IntelOrca:
@duncanspumpkin looks good, just need to fix the build errors. `GetNextRideToUpdate` could also be made `static`.

Gymnasiast:
@duncanspumpkin Could you get this into a mergeable state?

rik-smeets:
> @duncanspumpkin Could you get this into a mergeable state?

Could you, @duncanspumpkin ?

duncanspumpkin:
This should really break replays but as bpb is really a static park the ratings don't change and we don't check where we are in the ratings calc as part of the replay checks. Either way I've bumped the network version for safety.

PR #14430, opened 2021-04-02
Fix logging of replay info to stdout

IntelOrca:
Use Console::WriteLine and replace \n with \r\n.

### DIFF
--- a/src/openrct2/interface/StdInOutConsole.cpp
+++ b/src/openrct2/interface/StdInOutConsole.cpp
@@ -134,7 +134,23 @@ void StdInOutConsole::WriteLine(const std::string& s, FormatToken colourFormat)
     {
         if (_isPromptShowing)
         {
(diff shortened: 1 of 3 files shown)


src/openrct2/interface/StdInOutConsole.cpp
```diff
@@ -134,7 +134,23 @@ void StdInOutConsole::WriteLine(const std::string& s, FormatToken colourFormat)
...
+            if (s.find('\n') != std::string::npos)
+            {
+                for (auto ch : s)
+                {
+                    if (ch == '\n')
+                        newString += "\r\n";
+                    else
+                        newString += ch;
+                }
+                mainString = newString.c_str();
```
tupaschoal, line 150: It seems **very** weird, to replace an `std::string` internal `char*`. Does it need to? It created the `newString` already

IntelOrca: It would only be a `std::string*` otherwise, and would mean you just need a c_str in the WriteLine instead. I don't see any advantage over that way or this way.

PR #14434, opened 2021-04-03
Name num_sheltered_sections flags

Gymnasiast:


src/openrct2/ride/Ride.h
```diff
@@ -190,6 +190,13 @@ enum class RideClassification
     KioskOrFacility
 };
 
+enum ShelteredSectionsBits : uint8_t
```
duncanspumpkin, line 193: I think this would make more sense as a namespace with constexprs.

PR #14442, opened 2021-04-05
Fix ARM build in general and particularly MacOS cases.

devnexen:
Forcing keeping arch/target value as it s lost most of the time (empty
values).
Forcing no x86 build if this is arm thus not building mistakenly x86
instrinsics.


devnexen:
Note better to build with MACOS_USE_DEPENDENCIES as off since these are x86 binaries.

CMakeLists.txt
```diff
@@ -299,7 +306,9 @@ else ()
     ADD_CHECK_CXX_COMPILER_FLAG(CMAKE_CXX_FLAGS CXX_WARN_LOGICAL_OP -Wlogical-op)
     ADD_CHECK_CXX_COMPILER_FLAG(CMAKE_CXX_FLAGS CXX_WARN_RANGE_LOOP_ANALYSIS -Wrange-loop-analysis)
     ADD_CHECK_CXX_COMPILER_FLAG(CMAKE_CXX_FLAGS CXX_WARN_TAUTOLOGICAL_ZERO_COMPARE -Wtautological-unsigned-zero-compare)
-    ADD_CHECK_CXX_COMPILER_FLAG(CMAKE_CXX_FLAGS CXX_WARN_OLD_STYLE_CAST -Wold-style-cast)
+    if (NOT APPLE)
+        ADD_CHECK_CXX_COMPILER_FLAG(CMAKE_CXX_FLAGS CXX_WARN_OLD_STYLE_CAST -Wold-style-cast)
```
duncanspumpkin, line 310: Why this change? We shouldn't need to use old style casts on any platform

devnexen: I got this because of openssl and its old C cast style from its headers :-)

tupaschoal: Then we need to make sure SSL is identified as a system dependency and not the actual code

janisozaur: @devnexen can you address that, please?

tupaschoal: Thing is that we don't want to disable this warning. We need to find out where in the MacOS build the openSSL and other library are being included and tell the build system that it is a system include

src/openrct2/CMakeLists.txt
```diff
@@ -169,6 +170,12 @@ if (UNIX AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "BSD")
 endif ()
 
 set(THREADS_PREFER_PTHREAD_FLAG ON)
+if (APPLE)
+	set(CMAKE_THREAD_LIBS_INIT "-lpthread")
+	set(CMAKE_HAVE_THREADS_LIBRARY 1)
+	set(CMAKE_USE_WIN32_THREADS_INIT 1)
```
adam-bloom, line 176: why is a WIN32 cmake variable being set for macOS builds?

CMakeLists.txt
```diff
@@ -20,8 +20,15 @@ if (CCache_FOUND)
 endif (CCache_FOUND)
 
 if (APPLE)
-    set(CMAKE_OSX_ARCHITECTURES "x86_64")
-    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.14")
+    execute_process(COMMAND /usr/bin/uname -m OUTPUT_VARIABLE UNAME)
+    # oddily, cmake_system_processor can be empty
```
tupaschoal, line 24:
```suggestion
    # oddly, cmake_system_processor can be empty
```

Gymnasiast:
@adam-bloom Has everything you wanted to port now been ported?

adam-bloom:
> @adam-bloom Has everything you wanted to port now been ported?

Yes, I think we're good now.

PR #14462, opened 2021-04-11
Split guest and staff peep variables

duncanspumpkin:
This splits a huge amount of code so that it uses either Guest or Staff for the variables that are tied to guest or staff. (I've a follow on branch https://github.com/duncanspumpkin/OpenRCT2/commit/ddbc994769bb4c01234f15be60c890ce762fc731 that does the actual field moving so thats how I know I've caught all the instances)
Replays are expected to pass on this and the follow on will fail them.

### DIFF
--- a/src/openrct2-ui/windows/GameBottomToolbar.cpp
+++ b/src/openrct2-ui/windows/GameBottomToolbar.cpp
@@ -601,8 +601,9 @@ static void window_game_bottom_toolbar_draw_news_item(rct_drawpixelinfo* dpi, rc
                 return;
 
             auto clipCoords = ScreenCoordsXY{ 10, 19 };
-
-            if (peep->Is<Staff>() && peep->AssignedStaffType == StaffType::Entertainer)
+            auto* staff = peep->As<Staff>();
+            auto* guest = peep->As<Guest>();
+            if (staff != nullptr && staff->AssignedStaffType == StaffType::Entertainer)
             {
                 clipCoords.y += 3;
             }
@@ -614,17 +615,20 @@ static void window_game_bottom_toolbar_draw_news_item(rct_drawpixelinfo* dpi, rc
             auto image_id = ImageId(image_id_base, peep->TshirtColour, peep->TrousersColour);
             gfx_draw_sprite(&cliped_dpi, image_id, clipCoords);
 
-            if (image_id_base >= 0x2A1D && image_id_base < 0x2A3D)
+            if (guest != nullptr)
             {
-                gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, peep->BalloonColour), clipCoords);
-            }
-            else if (image_id_base >= 0x2BBD && image_id_base < 0x2BDD)
-            {
-                gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, peep->UmbrellaColour), clipCoords);
-            }
-            else if (image_id_base >= 0x29DD && image_id_base < 0x29FD)
-            {
-                gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, peep->HatColour), clipCoords);
+                if (image_id_base >= 0x2A1D && image_id_base < 0x2A3D)
+                {
+                    gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, guest->BalloonColour), clipCoords);
+                }
+                else if (image_id_base >= 0x2BBD && image_id_base < 0x2BDD)
+                {
+                    gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, guest->UmbrellaColour), clipCoords);
+                }
+                else if (image_id_base >= 0x29DD && image_id_base < 0x29FD)
+                {
+                    gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, guest->HatColour), clipCoords);
+                }
             }
             break;
         }
--- a/src/openrct2-ui/windows/GameBottomToolbar.cpp
+++ b/src/openrct2-ui/windows/GameBottomToolbar.cpp
@@ -602,7 +602,6 @@ static void window_game_bottom_toolbar_draw_news_item(rct_drawpixelinfo* dpi, rc
 
             auto clipCoords = ScreenCoordsXY{ 10, 19 };
             auto* staff = peep->As<Staff>();
-            auto* guest = peep->As<Guest>();
             if (staff != nullptr && staff->AssignedStaffType == StaffType::Entertainer)
             {
                 clipCoords.y += 3;
@@ -615,6 +614,7 @@ static void window_game_bottom_toolbar_draw_news_item(rct_drawpixelinfo* dpi, rc
             auto image_id = ImageId(image_id_base, peep->TshirtColour, peep->TrousersColour);
             gfx_draw_sprite(&cliped_dpi, image_id, clipCoords);
 
+            auto* guest = peep->As<Guest>();
             if (guest != nullptr)
             {
                 if (image_id_base >= 0x2A1D && image_id_base < 0x2A3D)
--- a/src/openrct2-ui/windows/Guest.cpp
+++ b/src/openrct2-ui/windows/Guest.cpp
@@ -759,7 +759,6 @@ static void window_guest_overview_tab_paint(rct_window* w, rct_drawpixelinfo* dp
     }
 
     auto* staff = peep->As<Staff>();
-    auto* guest = peep->As<Guest>();
     if (staff != nullptr && staff->AssignedStaffType == StaffType::Entertainer)
         screenCoords.y++;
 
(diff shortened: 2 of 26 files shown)


src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1173,10 +1173,89 @@ template<> void S6Exporter::ExportEntity(RCT2SpriteVehicle* dst, const Vehicle*
...
+    dst->happiness_target = src->HappinessTarget;
+    dst->nausea = src->Nausea;
+    dst->nausea_target = src->NauseaTarget;
+    dst->hunger = src->Hunger;
+    dst->thirst = src->Thirst;
+    dst->toilet = src->Toilet;
+    dst->time_to_consume = src->TimeToConsume;
+    dst->intensity = static_cast<uint8_t>(src->Intensity);
+    dst->nausea_tolerance = EnumValue(src->NauseaTolerance);
+    dst->window_invalidate_flags = src->WindowInvalidateFlags;
```
duncanspumpkin, line 1188: Incorrectly moved

duncanspumpkin:
Interesting that fails the replays must have made a mistake.

src/openrct2/world/SpriteBase.h
```diff
@@ -26,14 +26,14 @@ enum class EntityType : uint8_t
...
-    // Height from centre of sprite to bottom
-    uint8_t sprite_height_negative;
     uint16_t sprite_index;
     int16_t x;
     int16_t y;
     int16_t z;
     // Width from centre of sprite to edge
     uint8_t sprite_width;
+    // Height from centre of sprite to bottom
+    uint8_t sprite_height_negative;
```
duncanspumpkin, line 36: Revert leave for follow on

duncanspumpkin:
So I've investigated the differences and its for two reasons.

1. Staff no longer have a `FavouriteRide` which is always set to null
2. bpb1000 is weird and has a load of fields in use for staff that I have no idea why. I assume this is due to left over from back ages ago when we didn't clear the entity struct fully.

src/openrct2/GameStateSnapshots.cpp
```diff
@@ -263,58 +242,102 @@ struct GameStateSnapshots final : public IGameStateSnapshots
...
+        COMPARE_FIELD(Guest, ParkEntryTime);
+        COMPARE_FIELD(Guest, RejoinQueueTimeout);
+        COMPARE_FIELD(Guest, PreviousRide);
+        COMPARE_FIELD(Guest, PreviousRideTimeOut);
+        for (int i = 0; i < PEEP_MAX_THOUGHTS; i++)
+        {
+            COMPARE_FIELD(Guest, Thoughts[i]);
+        }
+        COMPARE_FIELD(Guest, GuestHeadingToRideId);
+        COMPARE_FIELD(Guest, GuestIsLostCountdown);
```
tupaschoal, line 316: This didn't use to be on the comparisons :O. I'm assuming it was an oversight from the past.

src/openrct2-ui/windows/GameBottomToolbar.cpp
```diff
@@ -601,8 +601,9 @@ static void window_game_bottom_toolbar_draw_news_item(rct_drawpixelinfo* dpi, rc
                 return;
 
             auto clipCoords = ScreenCoordsXY{ 10, 19 };
-
-            if (peep->Is<Staff>() && peep->AssignedStaffType == StaffType::Entertainer)
+            auto* staff = peep->As<Staff>();
+            auto* guest = peep->As<Guest>();
```
tupaschoal, line 605: I'd move this statement 10 lines further, to be just before the `if` that needs it.

src/openrct2-ui/windows/Guest.cpp
```diff
@@ -758,7 +758,9 @@ static void window_guest_overview_tab_paint(rct_window* w, rct_drawpixelinfo* dp
         return;
     }
 
-    if (peep->Is<Staff>() && peep->AssignedStaffType == StaffType::Entertainer)
+    auto* staff = peep->As<Staff>();
+    auto* guest = peep->As<Guest>();
+    if (staff != nullptr && staff->AssignedStaffType == StaffType::Entertainer)
```
tupaschoal, line 763: I'd move this statement ~10 lines further, to be just before the `if` that needs it.

tupaschoal (commented):
This is a great (exciting) step forward

src/openrct2/GameStateSnapshots.cpp
duncanspumpkin: It was one of the few union fields that had the staff field first so it was there just under the staff name

src/openrct2-ui/windows/GameBottomToolbar.cpp
duncanspumpkin: I thought it made sense to have them together. I don't really mind where they go though

src/openrct2-ui/windows/GameBottomToolbar.cpp
```diff
@@ -601,8 +601,8 @@ static void window_game_bottom_toolbar_draw_news_item(rct_drawpixelinfo* dpi, rc
                 return;
 
             auto clipCoords = ScreenCoordsXY{ 10, 19 };
-
-            if (peep->Is<Staff>() && peep->AssignedStaffType == StaffType::Entertainer)
+            auto* staff = peep->As<Staff>();
+            if (staff != nullptr && staff->AssignedStaffType == StaffType::Entertainer)
```
Broxzier, line 605: Since `staff` isn't used after this if, its scope can be reduces by writing it like this:
```suggestion
            if (auto* staff = peep->As<Staff>(); staff != nullptr && staff->AssignedStaffType == StaffType::Entertainer)
```

Though since it does more than just checking for nullptr, it might not be the best use case.

src/openrct2-ui/windows/GameBottomToolbar.cpp
```diff
@@ -614,17 +614,21 @@ static void window_game_bottom_toolbar_draw_news_item(rct_drawpixelinfo* dpi, rc
...
-            {
-                gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, peep->BalloonColour), clipCoords);
-            }
-            else if (image_id_base >= 0x2BBD && image_id_base < 0x2BDD)
-            {
-                gfx_draw_sprite(&cliped_dpi, ImageId(image_id_base + 32, peep->UmbrellaColour), clipCoords);
-            }
-            else if (image_id_base >= 0x29DD && image_id_base < 0x29FD)
+            auto* guest = peep->As<Guest>();
+            if (guest != nullptr)
```
Broxzier, line 618: This would be a better use case:

```suggestion
            if (auto* guest = peep->As<Guest>(); guest != nullptr)
```

src/openrct2/actions/StaffHireNewAction.cpp
```diff
@@ -128,7 +128,7 @@ GameActions::Result::Ptr StaffHireNewAction::QueryExecute(bool execute) const
         return MakeResult(GameActions::Status::NoFreeElements, STR_TOO_MANY_STAFF_IN_GAME);
     }
 
-    Peep* newPeep = CreateEntity<Staff>();
+    Staff* newPeep = CreateEntity<Staff>();
     if (newPeep == nullptr)
```
Broxzier, line 132:
```suggestion
    if (Staff* newPeep = CreateEntity<Staff>(); newPeep == nullptr)
```

src/openrct2/paint/sprite/Paint.Peep.cpp
```diff
@@ -82,25 +82,28 @@ template<> void PaintEntity(paint_session* session, const Peep* peep, int32_t im
     uint32_t imageId = baseImageId | peep->TshirtColour << 19 | peep->TrousersColour << 24 | IMAGE_TYPE_REMAP
         | IMAGE_TYPE_REMAP_2_PLUS;
     PaintAddImageAsParent(session, imageId, 0, 0, 1, 1, 11, peep->z, 0, 0, peep->z + 5);
-
-    if (baseImageId >= 10717 && baseImageId < 10749)
+    auto* guest = peep->As<Guest>();
+    if (guest != nullptr)
```
Broxzier, line 86:
```suggestion
    if (auto* guest = peep->As<Guest>(); guest != nullptr)
```

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1201,15 +1279,16 @@ void S6Exporter::ExportEntityPeep(RCT2SpritePeep* dst, const Peep* src)
     }
     if (generateName)
     {
-        if (src->Is<Staff>())
+        auto* staff = src->As<Staff>();
+        if (staff != nullptr)
```
Broxzier, line 1283:
```suggestion
        if (auto* staff = src->As<Staff>(); staff != nullptr)
```

src/openrct2-ui/windows/GameBottomToolbar.cpp
Broxzier: We could even remove the `guest != nullptr` part, though I think our warning settings don't like implicit nullptr checks.

src/openrct2-ui/windows/GameBottomToolbar.cpp
duncanspumpkin: I've previously used that style for if's but it was agreed that as its not used in the rest of the codebase it would be best to be consistent and not use it.

src/openrct2/rct2/S6Exporter.cpp
Gymnasiast: I have to disagree - the old code is much easier to read.

PR #14472, opened 2021-04-14
Clean up some footpath globals

Gymnasiast:


IntelOrca (approved):

src/openrct2/actions/PlaceParkEntranceAction.cpp
```diff
@@ -21,8 +21,9 @@
 #include "../world/Sprite.h"
 #include "../world/Surface.h"
 
-PlaceParkEntranceAction::PlaceParkEntranceAction(const CoordsXYZD& location)
+PlaceParkEntranceAction::PlaceParkEntranceAction(const CoordsXYZD& location, ObjectEntryIndex pathType)
     : _loc(location)
+    , _pathType(pathType)
```
duncanspumpkin, line 26: You've broken multiplayer park entrance placement as the path type is not being passed over the network in the serialise function

Gymnasiast: I’ll fix that.

By the way, I don’t think it worked properly before - it used an unsynced global. Wouldn’t that have caused desyncs?

Broxzier: Nope, they would only appear different, it doesn't affect any sprites or behaviour.

duncanspumpkin: Yeah before it wasn't synced but now it could be anything so might cause a crash.

IntelOrca: What exactly is causing the crash?

duncanspumpkin: Not a crash just a potential crash / non determinism. This field isn't being serialised. It wasn't being serialised before either but before it would just end up being whatever the clients last path type was.

PR #14479, opened 2021-04-14
workflow yml improvements

Margen67:
```
ci.yml:
 Use job-wide shell: bash.
 Always use checkout v2.
 Upgrade commitlint-github-action to v3.
 Remove brackets around check-code-formatting since it's only one item.
 Replace MinSizeRel with Release.
 upload-artifact:
  Always use v2.
  name:
   Remove unneeded quotation marks around name.
   Use runner.os when possible.
  if-no-files-found: error
 Make Windows mingw win32/NT5.1 into matrix.
 Add HOMEBREW_NO_ANALYTICS=1 to brew install.
 Make linux-portable(-32) into matrix.
 linux-docker:
  Use checkout action instead of git clone.
localisation.yml:
 Remove erroneous space after cron:
 Use --depth 1 to speed up cloning Localisation.
 Commit and push:
  Replace pushd with working-directory.
  Remove unneeded popd.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,8 @@
 name: CI
 on: [push, pull_request]
+defaults:
+  run:
+    shell: bash
 env:
   OPENRCT2_BUILD_SERVER: GitHub
   OPENRCT2_ORG_TOKEN: ${{ secrets.OPENRCT2_ORG_TOKEN }}
@@ -47,10 +50,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v1
       - name: Build OpenRCT2
-        shell: bash
         run: . scripts/setenv && build
       - name: Build artifacts
-        shell: bash
         run: |
           . scripts/setenv -q
           build-portable
@@ -62,10 +63,8 @@ jobs:
           name: "OpenRCT2-Windows-${{ matrix.platform }}"
           path: artifacts/*
       - name: Run Tests
-        shell: bash
         run: . scripts/setenv -q && run-tests
       - name: Upload artifacts (openrct2.org)
-        shell: bash
         run: |
           . scripts/setenv -q
           if [[ "$OPENRCT2_PUSH" == "true" ]]; then
@@ -164,10 +163,8 @@ jobs:
           brew install ninja
           . scripts/setenv -q && build -DWITH_TESTS=on -DCMAKE_BUILD_TYPE=MinSizeRel -DBUILD_SHARED_LIBS=on
       - name: Run Tests
-        shell: bash
         run: . scripts/setenv -q && run-tests
       - name: Build artifacts
-        shell: bash
         run: |
           . scripts/setenv
           mkdir -p artifacts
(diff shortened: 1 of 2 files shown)


tupaschoal (commented):
Changes like this are very welcome, thank you!

If you can make smaller, more self-contained commits though, it would be great. Reviewing it all at once, even though you helpfully describe all changes in the commit message, sometimes becomes confusing. Note it still could all be in one PR, but with commits like:

- Upgrade checkout action to v2
- Use default bash by default
- ...

Margen67:
> 
> 
> Changes like this are very welcome, thank you!
> 
> If you can make smaller, more self-contained commits though, it would be great. Reviewing it all at once, even though you helpfully describe all changes in the commit message, sometimes becomes confusing. Note it still could all be in one PR, but with commits like:
> 
>     * Upgrade checkout action to v2
> 
>     * Use default bash by default
> 
>     * ...

@tupaschoal Done.

.github/workflows/ci.yml
```diff
@@ -77,47 +77,42 @@ jobs:
...
         uses: hendrikmuhs/ccache-action@v1
         with:
-          key: windows-mingw-nt51
+          key: ${{ matrix.cache_key }}
       - name: Build OpenRCT2
         run: |
           sudo su
           mkdir bin && cd bin
-          cmake .. -G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DCMAKE_BUILD_TYPE=MinSizeRel -DDISABLE_IPO=on -DDISABLE_HTTP=Off -DFORCE32=on -DENABLE_SCRIPTING=ON -DCMAKE_CXX_FLAGS="-Wno-error=cast-function-type -Wno-error=unused-function" -DSTATIC=on -DMINGW_TARGET_NT5_1=ON
+          cmake .. -G Ninja -DCMAKE_TOOLCHAIN_FILE=../CMakeLists_mingw.txt -DCMAKE_BUILD_TYPE=Release -DDISABLE_IPO=on -DFORCE32=on ${{ matrix.build_flags }}
```
tupaschoal, line 109: Why the change from `MinSizeRel` to `Release`? (this is on commit https://github.com/OpenRCT2/OpenRCT2/pull/14479/commits/b513c0f297bdd394d65cdc728bda7a29128f2ca4)

.github/workflows/ci.yml
```diff
@@ -158,7 +158,7 @@ jobs:
           key: macos
       - name: Build OpenRCT2
         run: |
-          brew install ninja
+          HOMEBREW_NO_ANALYTICS=1 brew install ninja
```
tupaschoal, line 161: Why? What does it bring to the table? Maybe add the reasoning to the commit body

tupaschoal (commented):
I have just a few comments. It seems fine, the commit introducing the matrix is the one that concerns me most, though they look fine.

.github/workflows/ci.yml
Margen67: I thought `MinSizeRel` optimized for size instead of speed.

tupaschoal: From what I searched it seems to try to do both, but people do recommend `Release` indeed, so I'm ok with this.

.github/workflows/ci.yml
tupaschoal: What about this one?

tupaschoal (approved):
This looks fine to me, anyone else on @OpenRCT2/developers  wanna take a look?

janisozaur (approved):
Looks about right, the only thing in weary of is renaming the upload artifacts, that needs to be monitored post-merge

PR #14487, opened 2021-04-17
Fix compile error when DEBUG_LEVEL=1

ceeac:
This fixes the following compile errors:
```
/home/ceeac/Projects/code/OpenRCT2/src/openrct2/peep/GuestPathfinding.cpp:37:6: error: no previous extern declaration for non-static variable 'gPathFindDebug' [-Werror,-Wmissing-variable-declarations]
bool gPathFindDebug = false;
     ^
/home/ceeac/Projects/code/OpenRCT2/src/openrct2/peep/GuestPathfinding.cpp:37:1: note: declare 'static' if the variable is not intended to be used outside of this translation unit
bool gPathFindDebug = false;
^
/home/ceeac/Projects/code/OpenRCT2/src/openrct2/peep/GuestPathfinding.cpp:39:6: error: no previous extern declaration for non-static variable 'gPathFindDebugPeepName' [-Werror,-Wmissing-variable-declarations]
utf8 gPathFindDebugPeepName[256];
     ^
/home/ceeac/Projects/code/OpenRCT2/src/openrct2/peep/GuestPathfinding.cpp:39:1: note: declare 'static' if the variable is not intended to be used outside of this translation unit
utf8 gPathFindDebugPeepName[256];
^
```

### DIFF
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -34,9 +34,9 @@ ride_id_t gPeepPathFindQueueRideIndex;
 
 #if defined(DEBUG_LEVEL_1) && DEBUG_LEVEL_1
 // Use to guard calls to log messages
-bool gPathFindDebug = false;
+static bool gPathFindDebug = false;
 // Use to put the peep name in the log message
-utf8 gPathFindDebugPeepName[256];
+static utf8 gPathFindDebugPeepName[256];
 #endif // defined(DEBUG_LEVEL_1) && DEBUG_LEVEL_1
 
 static int32_t guest_surface_path_finding(Peep* peep);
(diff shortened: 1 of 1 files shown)


Gymnasiast (dismissed):

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -34,9 +34,9 @@ ride_id_t gPeepPathFindQueueRideIndex;
 
 #if defined(DEBUG_LEVEL_1) && DEBUG_LEVEL_1
 // Use to guard calls to log messages
-bool gPathFindDebug = false;
+static bool gPathFindDebug = false;
```
Gymnasiast, line 37: If it is confined to this file alone (which the `static` implies), it should not have the `g` prefix, which denotes globals.

Gymnasiast (changes requested):

src/openrct2/peep/GuestPathfinding.cpp
ceeac: Okay, I have changed it now.

PR #14490, opened 2021-04-18
Fix #14415: Entrances/exits are removed when built on top of each other

### LINKED ISSUE #14415: Lift(elevator)'s entrances/exits are disappeared if they are at the same tiles.
**OS:** Windows 10
**Version:** 0.3.3
**Commit/Build:** 465bcf7

Sorry and please close if this is duplicated one, but I couldn't find a similiar issues except for #14156
Elevator(Lift)'s entrance/exits are disappeared if they are at the same tiles with different height.

**Steps to reproduce:**
1. Build a Lift
2. Build a second station
3. Build entrance and exit to the first floor
4. Build entrance and exit to the second floor, with at the same tile of first floor.
5. Entrances/Exits disappears.

**Screenshots / Video:**
lift exit bug

It does not disappear only when it is not at the same tile.

adam-bloom:
Note that #14415 says Elevator/lift, but this bug applied to all ride types. There just probably isn't a use case to build a station for any other ride type on top of another station, but this can be seen with a simple roller coaster with two stations as well.

This loop was originally iterating over unique (x,y,z), casting each location to (x,y), and then handling each element on the tile. When the duplicate (x,y) iteration occurred (different z), the rest of this function logic deletes both the original entrance/exit and the one now being iterated over. Since all elements on the tile are iterated through, iterating over unique (x,y) elements instead resolves the issue. I validated with both the silly coaster and a lift - two entrances/exits can now be on the same tile.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -14,6 +14,7 @@
 - Fix: [#13986] OpenGL: Track preview window, flip/rotate button do not update the thumbnail.
 - Fix: [#14315] Crash when trying to rename Air Powered Vertical Coaster in Korean.
 - Fix: [#14330] join_server uses default_port from config.
+- Fix: [#14415] Entrances/exits are removed when built on top of each other.
 - Fix: [#14449] Surface smoothing at extra zoom levels not working.
 - Fix: [#14468] Cannot close Options window on Android.
 - Fix: [#14493] [Plugin] isHidden only works for tile elements up to the first element with a base height of over 32.
--- a/src/openrct2/network/NetworkBase.cpp
+++ b/src/openrct2/network/NetworkBase.cpp
@@ -36,7 +36,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "14"
+#define NETWORK_STREAM_VERSION "15"
 #define NETWORK_STREAM_ID OPENRCT2_VERSION "-" NETWORK_STREAM_VERSION
 
 static Peep* _pickup_peep = nullptr;
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -6671,6 +6671,8 @@ void sub_6CB945(Ride* ride)
         {
             if (tileElement->GetType() != TILE_ELEMENT_TYPE_ENTRANCE)
                 continue;
+            if (tileElement->base_height != locationCoords.z)
+                continue;
             if (tileElement->AsEntrance()->GetRideIndex() != ride->id)
                 continue;
             if (tileElement->AsEntrance()->GetEntranceType() > ENTRANCE_TYPE_RIDE_EXIT)


adam-bloom:
@ZehMatt looks like you modified this in 109a066. Is this same function used for both pathfinding and entrance/exit placement? I don't see a reference to this function from any pathfinding functions - just track place/remove actions (and similar). Perhaps I'm missing something?

distribution/changelog.txt
```diff
@@ -8,6 +8,7 @@
 - Fix: [#13894] Block brakes do not animate.
 - Fix: [#14315] Crash when trying to rename Air Powered Vertical Coaster in Korean.
 - Fix: [#14330] join_server uses default_port from config.
+- Fix: [#14415] Entrances/Exits removed if built on top of each other
```
Gymnasiast, line 11:
```suggestion
- Fix: [#14415] Entrances/exits are removed when built on top of each other.
```

ZehMatt:
> @ZehMatt looks like you modified this in [109a066](https://github.com/OpenRCT2/OpenRCT2/commit/109a06663d8ea01b3e9439c927baac76ee94a6c2) (this is effectively reverting that commit). Is this same function used for both pathfinding and entrance/exit placement? I don't see a reference to this function from any pathfinding functions - just track place/remove actions (and similar). Perhaps I'm missing something?

This will revert the fix that the exit/entrance station are connected to the wrong path. https://github.com/OpenRCT2/OpenRCT2/pull/13937 shows a demonstration. This whole thing probably needs some reconsideration on how its done.

adam-bloom:
> > @ZehMatt looks like you modified this in [109a066](https://github.com/OpenRCT2/OpenRCT2/commit/109a06663d8ea01b3e9439c927baac76ee94a6c2) (this is effectively reverting that commit). Is this same function used for both pathfinding and entrance/exit placement? I don't see a reference to this function from any pathfinding functions - just track place/remove actions (and similar). Perhaps I'm missing something?
> 
> This will revert the fix that the exit/entrance station are connected to the wrong path. #13937 shows a demonstration. This should be fixed correctly in my opinion.

Ah, because the entrances/exits are removed and re-added incorrectly? I wasn’t entirely sure why this function does the removal + addition in the first place. Do you think that needs to even be preserved?

adam-bloom:
> #13937 shows a demonstration. This should be fixed correctly in my opinion.

@ZehMatt I just tried to build a ride like that example, using the develop build - the entrances and exits disappeared.
about to place last exit:
Screen Shot 2021-04-18 at 5 55 29 PM

after clicking (when this issue is triggered):
Screen Shot 2021-04-18 at 5 55 48 PM

So yes, sounds like we need a more correct fix indeed :) Are you able to reproduce this as well? I can look through this all again to see if I can come up with something that works well for both cases.

adam-bloom:
I've made a change to instead check the z coordinate of the tileElement while iterating over unique (x,y,z) - this seems to do the trick!

@ZehMatt can you retest the path connections now? I tested them with a simple car ride and saw the correct behavior.

ZehMatt:
It seems to be working fine. You need to raise the network version since it changes the logic otherwise looks good to me.

Broxzier (approved):
Tested it, and the changes look make sense and look good to me.

PR #14495, opened 2021-04-18
Peep Guest/Staff Split: Actually move fields from split

duncanspumpkin:


duncanspumpkin:
This will break replays that is expected as it changes the layout of the entities in memory.

tupaschoal (approved):
Seems ok

duncanspumpkin:
Now that my network serialiser is in this no longer requires a change to the network version. I'm expecting all replays to pass.

duncanspumpkin:
Woo replays finally pass correctly.

PR #14535, opened 2021-04-27
Fix OpenRA logo url in readme

Basssiiie:
Fixes the url for the OpenRA logo in the readme, under '[Similar Projects](https://github.com/OpenRCT2/OpenRCT2#similar-projects)'. It seems the original image was [moved to a different folder](https://github.com/OpenRA/OpenRA/commit/c1f79b348ae189a17bead893e7f139b8ad6240f7) somewhere at the end of last year and it was probably broken in OpenRCT2's readme since then.


tupaschoal (approved):

tupaschoal:
Thanks for catching it!

PR #14536, opened 2021-04-27
Plugin API: Rename 'ListView' to 'ListViewWidget' in Typescript declaration for consistency

Basssiiie:
As previously discussed in the `#plugin` channel on the OpenRCT2 Discord, here's a pull request to rename `ListView` to `ListViewWidget` to be more consistent with other widgets.

Similar to #14496 this change has no effect on plugin users and would not require a plugin version increment. For plugin developers it is merely a Typescript name that has changed.


Basssiiie:
Tagging @Sadret and @IntelOrca :)

IntelOrca (approved):

PR #14541, opened 2021-04-28
Network serialiser for entities

duncanspumpkin:
will be used only for checksums and replay diffs

Similar but not the same as the .park entity serialiser as it discards some items (names, sprite dimensions, variable length items)

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -302,7 +302,7 @@ template<typename _Ty, size_t _Size> struct DataSerializerTraitsPODArray
         uint16_t swapped = ByteSwapBE(len);
         stream->Write(&swapped);
 
-        DataSerializerTraits<uint8_t> s;
+        DataSerializerTraits<_Ty> s;
         for (auto&& sub : val)
         {
             s.encode(stream, sub);
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -584,6 +584,34 @@ template<> struct DataSerializerTraits_t<CoordsXYZD>
         stream->Write(msg, strlen(msg));
     }
 };
+template<> struct DataSerializerTraits_t<rct12_xyzd8>
+{
+    static void encode(OpenRCT2::IStream* stream, const rct12_xyzd8& coord)
+    {
+        stream->WriteValue(ByteSwapBE(coord.x));
+        stream->WriteValue(ByteSwapBE(coord.y));
+        stream->WriteValue(ByteSwapBE(coord.z));
+        stream->WriteValue(ByteSwapBE(coord.direction));
+    }
+
+    static void decode(OpenRCT2::IStream* stream, rct12_xyzd8& coord)
+    {
+        auto x = ByteSwapBE(stream->ReadValue<uint8_t>());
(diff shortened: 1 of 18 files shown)


src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -584,6 +584,34 @@ template<> struct DataSerializerTraits_t<CoordsXYZD>
...
+
+    static void log(OpenRCT2::IStream* stream, const rct12_xyzd8& coord)
+    {
+        char msg[128] = {};
+        snprintf(
+            msg, sizeof(msg), "rct12_xyzd8(x = %d, y = %d, z = %d, direction = %d)", coord.x, coord.y, coord.z,
+            coord.direction);
+        stream->Write(msg, strlen(msg));
+    }
+};
```
IntelOrca, line 614: `rct12_xyzd8` is no longer used in new-save-format. It has been changed to standard tilecoords

IntelOrca:
Why is called network serialiser? The serialised data isn't actually sent over the network.

duncanspumpkin:
Because it's only going to be used for network checksums and replay entity snapshots

src/openrct2/core/DataSerialiserTraits.h
duncanspumpkin: That shouldn't be an issue.

duncanspumpkin:
I'm expecting that the replays will fail saying bad checksum and cannot read snapshot. I'll have to rerecord them.

src/openrct2/world/Sprite.cpp
```diff
@@ -323,9 +308,11 @@ rct_sprite_checksum sprite_checksum()
         }
 
         _spriteHashAlg->Clear();
+        OpenRCT2::MemoryStream ms;
+        DataSerialiser ds(true, ms);
 
-        ComputeChecksumForEntityTypes<Guest, Staff, Vehicle, Litter>(_spriteHashAlg.get());
-
+        NetworkSerialiseEntityTypes<Guest, Staff, Vehicle, Litter>(ds);
+        _spriteHashAlg->Update(ms.GetData(), ms.GetLength());
```
ZehMatt, line 315: This introduces a lot of overhead, this is almost as expensive as saving the park now minus the file i/o. 
1. MemoryStream uses dynamic memory allocation
2. All data is copied first into the buffer.
3. Has to run over the copied data again to compute the checksum

Theres probably a trick behind this by introducing a new stream that inherhits from IStream, lets call it ChecksumStream.

The serializer will feed all data via write so the checksum can be done on the fly.

duncanspumpkin:
This is causing a bit of a slowdown in time to execute the replay tests. Could I get a few testers to see if this is a noticble slowdown in general in multiplayer. The new hash function would ideally be used for this but thats only in the new save branch for now.

tupaschoal:
> This is causing a bit of a slowdown in time to execute the replay tests. Could I get a few testers to see if this is a noticble slowdown in general in multiplayer. The new hash function would ideally be used for this but thats only in the new save branch for now.

What are the tests you require? Just load multiplayer? I can try but I don't get what I should be looking at.

duncanspumpkin:
@tupaschoal if changed again how it works so that there shouldn't be any slowdown. Should be all good to go now.

IntelOrca (approved):

PR #14546, opened 2021-04-29
Improve performance on saving (new save format)

ZehMatt:
With the new limits the save time is pretty tremendous. I'm not a fan that my game comes to almost an halt when auto save kicks in, this makes it almost instant so the user shouldn't even notice it.

Holding the hotkey for save
Before:

After:

### DIFF
--- a/src/openrct2/ParkFile.cpp
+++ b/src/openrct2/ParkFile.cpp
@@ -47,6 +47,7 @@
 
 #include <cstdint>
 #include <ctime>
+#include <future>
 #include <numeric>
 #include <string_view>
 #include <vector>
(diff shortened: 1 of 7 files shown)


IntelOrca:
There are many changes here that have not been discussed like changing the hashing. It isn't obvious which of these changes is actually the main improvement on the performance... is it just that it is now asynchronous? If is async. that is great except, users may close the game before the file has fully saved to disc which would be tragic. Can we perhaps only use the async code when autosaving?

ZehMatt:
The main improvement is the sha1 replacement. 
Before:
sleepy_2021-04-29_12-50-39

After:
sleepy_2021-04-29_14-16-0917c1f0386a24c7253

Another part that is actually relatively slow is the gzip compression but that was fairly easy to do by moving it into async lambda.

duncanspumpkin:
I'm fine with a simpler hashing function just as long as its better than the old DAT checksums that can have collisions.

IntelOrca:
The changes to use a new hash function are a lot larger than they need to be. I would expect just change Crypt::SHA1 to Crypt::FNV1a.

IntelOrca:
I have applied just the FNV1a to the branch with some improvements.

PR #13718, opened 2021-01-07
Eliminate copying of paint_struct, initialize data in-place

ZehMatt:
Before:
sleepy_2021-01-07_10-13-44

After:
sleepy_2021-01-07_10-10-03

Small performance boost

PR #14561, opened 2021-05-03
Dynamically allocate paint structs and remove 4000 per column limit

IntelOrca:
Some parks have so many tile elements on a single tile, and vertical screen column of tiles, that it exceeds the number of available paint structs for that paint session. This results in large areas of the screen undrawn.

All paint sessions now share a common pool of paint structs. I have used an unrolled linked list which allows each session to grab 512 paint structs at a time and only grab another set of 512 when they have been used up. This reduces the amount of thread locking involved when multithreaded drawing is enabled.

The amount of memory used will now be based on the complexity of the scene being drawn. Less memory in some cases, more in the cases that would have exceeded the original limit.

### DIFF
--- a/src/openrct2/cmdline/BenchSpriteSort.cpp
+++ b/src/openrct2/cmdline/BenchSpriteSort.cpp
@@ -34,47 +34,50 @@
 #    include <iterator>
 #    include <vector>
 
-static void fixup_pointers(paint_session* s, size_t paint_session_entries, size_t paint_struct_entries, size_t quadrant_entries)
+static void fixup_pointers(std::vector<RecordedPaintSession>& s)
 {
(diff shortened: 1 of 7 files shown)


duncanspumpkin:
What's the performance impact of this? It looks like it reverts the revert of the paint struct allocation that was done previously due to performance concerns.

IntelOrca:
> paint struct allocation that was done previously due to performance concerns

Can you remind me what the changes were or PR?

> What's the performance impact of this?

Low, possibly some impact, but not sure which part could be optimised yet.

duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/pull/13718 hmm it's a bit different to what was reverted miss remembered it a bit

ZehMatt (approved):

PR #14570, opened 2021-05-04
Add ARM SIMD routines in SW renderer on Linux

HenkKalkwater:
This commit introduces an mask_neon function, which renders draws the
mask using the ARM NEON SIMD instruction set. This is based on the AVX
mask drawing function.

I do not know how to test for NEON support in non-Linux OSs, therefore
this commit only enables it on arm devices running Linux.

This commit has only been tested on an ARM 32-bit processor running
Linux, compiled with GCC. On my Sony Xperia X running SailfishOS
(32-bit ARM-based Linux distro) this boosts the performance on the title
screen from 8-12 FPS to 20-40 FPS (depending on how complex the scene
displayed screen actually is).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,6 +238,10 @@ elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "amd64.*|x86_64.*|AMD64.*")
     set(X86_64 1)
 elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "i686.*|i386.*|x86.*|amd64.*|AMD64.*")
     set(X86 1)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm.*|ARM.*)")
+  set(ARM 1)
+elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*)")
+  set(AARCH64 1)
 endif()
 
 # Check if a flag exists and add it to the list of compiler (so, not linker) options
--- a/src/openrct2/CMakeLists.txt
+++ b/src/openrct2/CMakeLists.txt
@@ -250,6 +250,10 @@ if((X86 OR X86_64) AND NOT MSVC)
     set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}/drawing/AVX2Drawing.cpp PROPERTIES COMPILE_FLAGS -mavx2)
 endif()
 
+if((ARM OR AARCH64) AND NOT MSVC)
+    set_source_files_properties(${CMAKE_CURRENT_LIST_DIR}/drawing/NEONDrawing.cpp PROPERTIES COMPILE_FLAGS -mfpu=neon)
+endif()
+
 # Add headers check to verify all headers carry their dependencies.
 # Only valid for Clang for now:
 # - GCC 8 does not support -Wno-pragma-once-outside-header
--- a/src/openrct2/common.h
+++ b/src/openrct2/common.h
@@ -179,11 +179,24 @@ constexpr uint16_t SPRITE_INDEX_NULL = 0xFFFF;
 #    define PLATFORM_X86
 #endif
 
+#if defined(__GNUC__) && defined(__arm__)
+#   define OPENRCT2_ARM
+#elif defined(_MSC_VER) && (defined(_M_ARM))
+#   define OPENRCT2_ARM
+#endif
+
+#if defined(__GNUC__) && defined(__aarch64__)
+#   define OPENRCT2_AARCH64
+#elif defined(_MSC_VER) && (defined(_M_ARM64))
+#   define OPENRCT2_AARCH64
+#endif
+
 #if defined(__LP64__) || defined(_WIN64)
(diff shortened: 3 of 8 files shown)


HenkKalkwater:
Seems like I forgot to translate a few SSE instructions to NEON ones. I wonder why it even compiles. I'm fixing it right now.

HenkKalkwater:
Fixed the problems which most likely were caused by a stale build. It seems this merge request didn't improve the speeds as I claimed, those can be explained by switching from the software to the software_hwd renderer. Anyway, this now seems to work and it still seems to give a performance boost of a few frames, but not in the tens of frames I was talking about.

HenkKalkwater:
Honestly, after some more testing, I don't know where the mask function is actually used, since intentionally corrupting it does not seem to generate any visual artefacts.. And I am not even sure if the performance difference is noticeable.

janisozaur:
Results from `benchgfx` using bpb.sv6 for 100 iterations on my aarch64 Rpi4B:

|               | `develop`        | this PR          |
|---------------|------------------|------------------|
| Zoom 0        | 0.464732s, 2 FPS | 0.509639s, 2 FPS |
| Zoom 1        | 0.184381s, 5 FPS | 0.198282s, 5 FPS |
| Zoom 2        | 0.106313s, 9 FPS | 0.110885s, 9 FPS |
| Total average | 0.251809s, 4 FPS | 0.272935s, 4 FPS |
| Total         | 302.17052s       | 327.52248s       |

Overall it is slower for me. Additionally, I had to remove `-mfpu=neon` option as it was not recognized - probably due to it being standard in ARMv8. This would need to have some detection added. I did confirm it was using the new masking function.

@HenkKalkwater can you run `benchgfx` on your system: 
`./openrct2-cli benchgfx ../test/tests/testdata/parks/bpb.sv6 100`
and report results?

You will also need to modify https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2-android/app/src/main/CMakeLists.txt

janisozaur:
My build was done with `g++ (Debian 8.3.0-6) 8.3.0` on raspberry pi os, and `RelWithDebInfo` build type

HenkKalkwater:
After a clean rebuild, I noticed I'm getting some very noticeable graphical glitches (on the `SOFTWARE_HWD` renderer). 

image


 When I have some more time, I'll try to figure out what's going on.

HenkKalkwater:
Results on gcc-8.3.0 with the flags `-O2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -march=armv7-a -mfloat-abi=hard -mfpu=neon -mthumb -Wno-psabi -fstrict-overflow -fstrict-aliasing -fPIC`, running on a Sony Xperia X with Sailfish OS 4.0.1.48 (32-bit, even though it has an AArch64 Processor):

| | `c3bfd22` | `c3bfd22` but by forcing it to use mask_scalar |
|---|---|---|
| Zoom 0 avg | 0.530221s, 2 FPS | 0.543725s, 2 FPS |
| Zoom 1 avg | 0.217304s, 5 FPS | 0.222952s, 4 FPS |
| Zoom 2 avg | 0.122383s, 8 FPS | 0.127128s, 8 FPS |
| Total avg     | 0.289969s, 3 FPS  | 0.297935s, 3 FPS |
| Time            |  347.96307s | 357.52204s |

I have to admit, I'm a bit disappointed. I honestly hoped for a higher performance increase. [It also seems like GCC won't automatically generate vector assembly functions for `mask_scalar`](https://godbolt.org/z/KsbqKTv3j). I wonder if there is an oversight in my code that slows it down or if it's just the case that the ARM NEON instructions aren't that useful for this case.

janisozaur:
Is there any chance for 64 bit build for your device?

Contrary to your description, this is more akin to SSE rather than AVX - perhaps a more unified (processing it in a single go rather than two halves) approach can be obtained?

HenkKalkwater:
This is indeed more like SSE than AVX. Sadly, the vector sizes have a maximum size of 128 bits it seems within NEON. So it does not seem to be possible to do the whole loop in 1 go.

When using SVE, the SIMD extension introduced somewhere in ARMv8, vector sizes larger than 128 bits should be possible. I'm able to produce AARCh64 builds with the toolchain I'm using, but since the whole operating system is 32-bits and as far as I know a 64-bit build is not available, I don't think it will actually run on my device.

janisozaur:
I did check actual parks now in OpenRCT2, it seems that `mask_scalar` went down from taking 5-7% of execution time to `mask_neon` taking 1%. That's a bit odd given `benchgfx` results. I will report back with actual perceived FPS later on.

@HenkKalkwater can you please check again what kind of results you see on your Xperia, when running actual parks? Ideally `printf` the FPS and take the average from 60 first results of running the default title screen.

janisozaur:
There are still some artifacts, though

HenkKalkwater:
> There are still some artifacts, though

Could you give an example? Besides some objects "fighting" over which should be in front, which also happens on an x64  PC for me, I cannot find any myself. 

I'll get to the testing somewhat later, I managed to break my Linux installation somehow.

PR #14583, opened 2021-05-07
FIX: Vehicle::GetCar function causing server crash

### LINKED ISSUE #14582: Vehicle::GetCar function issue
**OS:** Debian GNU/Linux 10.9
**Version:** 0.3.3

Vehicle::GetCar function is called 2 times in peep/Guest.cpp without checking for a null pointer before dereferencing while Vehicle::GetCar can return a null pointer. This causes servers to crash occasionally and occurs spontaneously without any currently known reproducibility triggers. Fixing this on the server-side keeps the server running but unpatched clients connected can still crash. 

The fix:
In peep/Guest.cpp right under line 3791 in master branch (3983 in develop branch), the line that reads 'vehicle = vehicle->GetCar(CurrentCar);', add:
if(!vehicle) return;

Also in peep/Guest.cpp under line 3474 (3666 in develop branch), the line that reads 'vehicle = vehicle->GetCar(CurrentCar);', add:
if(!vehicle) return;

ZxBiohazardZx:
https://github.com/OpenRCT2/OpenRCT2/issues/14582
Actual fix by @RoyGherbel 
Fixes #14582


duncanspumpkin:
We prefer to be explicit in this codebase when comparing to nullptr.

ZxBiohazardZx:
> We prefer to be explicit in this codebase when comparing to nullptr.

can you elaborate as i did not alter the nullptr comparison

duncanspumpkin (approved):

PR #14602, opened 2021-05-09
Sort cache

ZehMatt:
This is a proof of concept currently and more for testing purposes. This currently renders tweening useless as it builds the render list per tick rather per frame now. The tweener needs to be changed to update the render list instead and not the entities. The way this is currently implemented is also not very practical.

It has however some promising results.
```
---- bpb.sv6 @ 1080p, zoom/position from save, avg. FPS ----
develop: 130 (mt off)
sort-cache: 270 (mt off)

develop: 260 (mt on)
sort-cache: 330 (mt on)
```


ZehMatt:
Also gonna close this one, probably much easier to do once the paint stuff is properly refactored.

PR #14606, opened 2021-05-09
Test CI builtin

ZehMatt:


tupaschoal (commented):
Based on the Discord conversation, I'm assuming this is not to be merged as is (if at all), so I'm moving to draft just for tracking. Let me know if I misunderstood

ZehMatt:
I just wanted to see what would fail using this, we would need to crank up the clang version to min 9.

PR #14607, opened 2021-05-09
Fix some actions for plugins

rctdude2:
If an action doesn't have an AcceptParameters method, the parameters passed in executeAction will be ignored and the action will fail.

I only fixed the ones I found that were trivial, there are still some actions that involve coordinates that don't have AcceptParameters that I didn't modify because I'm not sure how they work in plugins. Also, I'm guessing this might need a network version increment, but I didn't do that because I wanted to ask first.

### DIFF
--- a/src/openrct2/actions/ParkMarketingAction.cpp
+++ b/src/openrct2/actions/ParkMarketingAction.cpp
@@ -32,7 +32,7 @@ void ParkMarketingAction::AcceptParameters(GameActionParameterVisitor& visitor)
 {
     visitor.Visit("type", _type);
     visitor.Visit("item", _item);
-    visitor.Visit("length", _numWeeks);
+    visitor.Visit("duration", _numWeeks);
 }
 
 uint16_t ParkMarketingAction::GetActionFlags() const
--- a/src/openrct2/actions/ParkMarketingAction.cpp
+++ b/src/openrct2/actions/ParkMarketingAction.cpp
@@ -28,6 +28,13 @@ ParkMarketingAction::ParkMarketingAction(int32_t type, int32_t item, int32_t num
 {
 }
 
+void ParkMarketingAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("type", _type);
+    visitor.Visit("item", _item);
+    visitor.Visit("length", _numWeeks);
+}
+
 uint16_t ParkMarketingAction::GetActionFlags() const
 {
     return GameAction::GetActionFlags() | GameActions::Flags::AllowWhilePaused;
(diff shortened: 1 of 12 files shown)


duncanspumpkin:
You'll need to increment the plug-in api version but no need to touch network version. The only other thing to think about is the naming of the variables. Are these consistent with the rest of the plug-in api I've not checked myself

src/openrct2/actions/SignSetNameAction.cpp
```diff
@@ -26,6 +26,12 @@ SignSetNameAction::SignSetNameAction(BannerIndex bannerIndex, const std::string&
 {
 }
 
+void SignSetNameAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("banner", _bannerIndex);
```
IntelOrca, line 31:
```suggestion
    visitor.Visit("id", _bannerIndex);
```

src/openrct2/actions/SetParkEntranceFeeAction.cpp
```diff
@@ -20,6 +20,11 @@ SetParkEntranceFeeAction::SetParkEntranceFeeAction(money16 fee)
 {
 }
 
+void SetParkEntranceFeeAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("fee", _fee);
```
IntelOrca, line 25:
```suggestion
    visitor.Visit("value", _fee);
```

src/openrct2/actions/ParkSetLoanAction.cpp
```diff
@@ -22,6 +22,11 @@ ParkSetLoanAction::ParkSetLoanAction(money32 value)
 {
 }
 
+void ParkSetLoanAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("loanAmount", _value);
```
IntelOrca, line 27:
```suggestion
    visitor.Visit("value", _value);
```

src/openrct2/actions/ParkMarketingAction.cpp
```diff
@@ -28,6 +28,13 @@ ParkMarketingAction::ParkMarketingAction(int32_t type, int32_t item, int32_t num
 {
 }
 
+void ParkMarketingAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("type", _type);
+    visitor.Visit("item", _item);
+    visitor.Visit("length", _numWeeks);
```
IntelOrca, line 35:
```suggestion
    visitor.Visit("duration", _numWeeks);
```

IntelOrca (commented):
I have made some naming suggestions which I think fit better or are more generic - intuitive amongst other actions.

IntelOrca (approved):

PR #14616, opened 2021-05-10
Change Vehicle Rotation namings in accordance with OpenRCT2 Codestyle

ZxBiohazardZx:
technically a followup of #11956 and upon a discussion with Duncanspumpkin on vehicle namings. Originated from the Screaming Issue: #11293

### DIFF
--- a/src/openrct2/rct1/RCT1.h
+++ b/src/openrct2/rct1/RCT1.h
@@ -224,8 +224,8 @@ struct rct1_unk_sprite : RCT12SpriteBase
 
 struct rct1_vehicle : RCT12SpriteBase
 {
-    uint8_t Pitch; // 0x1F
-    uint8_t bank_rotation;       // 0x20
+    uint8_t Pitch;         // 0x1F
+    uint8_t bank_rotation; // 0x20
     uint8_t pad_21[3];
     int32_t remaining_distance; // 0x24
     int32_t velocity;           // 0x28
--- a/src/openrct2/rct1/RCT1.h
+++ b/src/openrct2/rct1/RCT1.h
@@ -224,7 +224,7 @@ struct rct1_unk_sprite : RCT12SpriteBase
 
 struct rct1_vehicle : RCT12SpriteBase
 {
-    uint8_t vehicle_sprite_type; // 0x1F
+    uint8_t Pitch; // 0x1F
     uint8_t bank_rotation;       // 0x20
     uint8_t pad_21[3];
     int32_t remaining_distance; // 0x24
--- a/src/openrct2/rct2/RCT2.h
+++ b/src/openrct2/rct2/RCT2.h
@@ -453,7 +453,7 @@ assert_struct_size(rct_scores_entry, 0x02B0);
 
 struct RCT2SpriteVehicle : RCT12SpriteBase
 {
-    uint8_t vehicle_sprite_type; // 0x1F
+    uint8_t Pitch; // 0x1F
     uint8_t bank_rotation;       // 0x20
     uint8_t pad_21[3];
     int32_t remaining_distance; // 0x24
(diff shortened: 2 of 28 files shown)


ZxBiohazardZx:
> How does this differ from #14585? Generally it is better to get things fixed in a PR without reopening one that does roughly the same, so we can keep track of comments and stuff

branch got removed in a cleanup of my working dirs of github. essentially this is the same fix, but i couldnt restore the old

ZxBiohazardZx:
> > branch got removed in a cleanup of my working dirs of github. essentially this is the same fix, but i couldnt restore the old
> 
> I see, thanks for the answer. A tip for the future: create new branches to do your work instead of doing it on `development`, it typically prevents those headaches.
> 
> Other than that, please see the results of the code formatting CI (you can [click here](https://github.com/OpenRCT2/OpenRCT2/pull/14616/checks?check_run_id=2544041935) or on the statuses below) and fix with the suggestions made, so it can run the other tests. Let us know if you need help

already asked Duncan to fix the issue with CI. no idea why it fails on the whitespace. i used a simple refactor to replace all ocurrances

src/openrct2/ride/Vehicle.h
```diff
@@ -43,7 +43,7 @@ struct rct_vehicle_info
     int16_t y;                   // 0x02
     int16_t z;                   // 0x04
     uint8_t direction;           // 0x06
-    uint8_t vehicle_sprite_type; // 0x07
+    uint8_t Pitch; // 0x07
     uint8_t bank_rotation;       // 0x08
```
tupaschoal, line 47: This block should become:
```c++
    int16_t x;             // 0x00
    int16_t y;             // 0x02
    int16_t z;             // 0x04
    uint8_t direction;     // 0x06
    uint8_t Pitch;         // 0x07
    uint8_t bank_rotation; // 0x08
```

I can't suggest the change like the other places because GitHub only allows it within 3 lines of the changed code.

src/openrct2/rct1/RCT1.h
```diff
@@ -224,7 +224,7 @@ struct rct1_unk_sprite : RCT12SpriteBase
 
 struct rct1_vehicle : RCT12SpriteBase
 {
-    uint8_t vehicle_sprite_type; // 0x1F
+    uint8_t Pitch; // 0x1F
     uint8_t bank_rotation;       // 0x20
```
tupaschoal, line 228:
```suggestion
    uint8_t Pitch;         // 0x1F
    uint8_t bank_rotation; // 0x20
```

src/openrct2/rct2/RCT2.h
```diff
@@ -453,7 +453,7 @@ assert_struct_size(rct_scores_entry, 0x02B0);
 
 struct RCT2SpriteVehicle : RCT12SpriteBase
 {
-    uint8_t vehicle_sprite_type; // 0x1F
+    uint8_t Pitch; // 0x1F
     uint8_t bank_rotation;       // 0x20
```
tupaschoal, line 457:
```suggestion
    uint8_t Pitch;         // 0x1F
    uint8_t bank_rotation; // 0x20
```

tupaschoal (commented):
I understand you only renamed variable, but the formatting enforces the comments to be aligned and the change did break some of these alignments. I have commented on the places that need to be updated, it shouldn't be difficult.

duncanspumpkin:
Also in future always make a branch off of develop otherwise it will be a pain to rebase.

duncanspumpkin (approved):
This is fine. I know pitch isn't correct for some rides like the Ferris wheel but we will introduce a union for a better name for those situations. Corkscrews also use pitch in a special way but again i think that is acceptable. The next stage after this will be to create the union. Eventually I'd like this to be an enum class for the pitch variable.

PR #14620, opened 2021-05-10
Add more properties to Park API

rctdude2:
Adds more properties to the plugin Park API that are useful for certain kinds of plugins (for example, automatic price management). I added the guestInitialX properties here since the flags e.g. difficult guest generation that are also set in the scenario editor are also here.


duncanspumpkin:
Aren't these all just scenario options? You should only have get methods for them. Changing should happen through the action interface. Some of these when changed are meant to modify ui elements and such.

rctdude2:
These are only get methods, the intention is that plugins can query them and display/perform other actions based on their values (for example, guestInitialCash and totalRideValueForMoney could be used to automatically change the park entrance fee).

tupaschoal (approved):
LGTM, what do you think @IntelOrca ?

PR #14625, opened 2021-05-11
Rename PeepActionType None1/None2

ZehMatt:
I think this makes a bit more sense to read now.

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -223,178 +223,178 @@ static struct
 } PeepThoughtToActionMap[] = {
     { PeepActionType::ShakeHead, PEEP_THOUGHT_ACTION_FLAG_RIDE },
     { PeepActionType::EmptyPockets, PEEP_THOUGHT_ACTION_NO_FLAGS },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_NO_FLAGS },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_NO_FLAGS },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_FLAG_RIDE },
+    { PeepActionType::Walking, PEEP_THOUGHT_ACTION_NO_FLAGS },
+    { PeepActionType::Walking, PEEP_THOUGHT_ACTION_NO_FLAGS },
+    { PeepActionType::Walking, PEEP_THOUGHT_ACTION_FLAG_RIDE },
     { PeepActionType::Wow, PEEP_THOUGHT_ACTION_FLAG_RIDE },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_FLAG_SHOP_ITEM_SINGULAR },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_FLAG_RIDE },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_FLAG_RIDE },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_NO_FLAGS },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_FLAG_RIDE },
-    { PeepActionType::None2, PEEP_THOUGHT_ACTION_FLAG_SHOP_ITEM_INDEFINITE },
+    { PeepActionType::Walking, PEEP_THOUGHT_ACTION_FLAG_SHOP_ITEM_SINGULAR },
+    { PeepActionType::Walking, PEEP_THOUGHT_ACTION_FLAG_RIDE },
+    { PeepActionType::Walking, PEEP_THOUGHT_ACTION_FLAG_RIDE },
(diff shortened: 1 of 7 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -1247,13 +1247,13 @@ void Guest::UpdateSitting()
     }
     else if (SittingSubState == PeepSittingSubState::SatDown)
     {
-        if (Action < PeepActionType::None1)
+        if (Action < PeepActionType::Idle)
```
duncanspumpkin, line 1250: we should go through and make these explicit that its idle or walking

duncanspumpkin, line 1250: if not (idle or walking)

src/openrct2/peep/Guest.cpp
```diff
@@ -4781,7 +4781,7 @@ void Guest::UpdateRideMazePathfinding()
         return;
     }
 
-    if (Action >= PeepActionType::None1)
+    if (Action >= PeepActionType::Idle)
```
duncanspumpkin, line 4784: idle or walking

src/openrct2/peep/Peep.cpp
```diff
@@ -359,7 +359,7 @@ void Peep::SwitchToSpecialSprite(uint8_t special_sprite_id)
     SpecialSprite = special_sprite_id;
 
     // If NONE_1 or NONE_2
```
tupaschoal, line 361:
```suggestion
    // If Idle or Walking
```
We can either fix the comment or delete it altogether, your call

src/openrct2/peep/Peep.cpp
```diff
@@ -405,7 +405,7 @@ std::optional<CoordsXY> Peep::UpdateAction(int16_t& xy_distance)
 
     xy_distance = x_delta + y_delta;
 
-    if (Action == PeepActionType::None1 || Action == PeepActionType::None2)
+    if (IsWalking())
```
tupaschoal, line 408: This used to consider both, so shouldn't it be:
```suggestion
    if (IsActionInterruptable())
```
?

src/openrct2/world/MapAnimation.cpp
```diff
@@ -199,8 +199,6 @@ static bool map_animation_invalidate_small_scenery(const CoordsXYZ& loc)
                         continue;
                     if (peep->z != loc.z)
                         continue;
-                    if (peep->Action < PeepActionType::None1)
-                        continue;
```
tupaschoal, line 203: Should this one be removed? It feels a bit disconnected from the rest of the PR.

ZehMatt: A few lines above it already checks for a specific action

src/openrct2/peep/Peep.cpp
ZehMatt: Its an redundant condition, few lines above it will always convert the action from Idle into Walking, having Idle is impossible in that statement.

tupaschoal (approved):
Seems ok, I'm just wondering whether it should be Interruptable or Interruptible (I usually favour the second), but no big deal.

src/openrct2/world/MapAnimation.cpp
duncanspumpkin: I don't think that's exactly the same it checks for a specific state but not a specific action.

src/openrct2/peep/Peep.cpp
```diff
@@ -1528,6 +1527,21 @@ bool Peep::SetName(std::string_view value)
     return false;
 }
 
+bool Peep::IsWalking() const
```
duncanspumpkin, line 1530: I think this will cause confusion with the walking state. Agree that this should be renamed to add action into the name IsActionWalking same for idle to be consistent

src/openrct2/peep/Staff.cpp
```diff
@@ -1791,8 +1791,8 @@ void Staff::Tick128UpdateStaff()
     SpriteType = newSpriteType;
     ActionSpriteImageOffset = 0;
     WalkingFrameNum = 0;
-    if (Action < PeepActionType::None1)
-        Action = PeepActionType::None2;
+    if (Action < PeepActionType::Idle)
+        Action = PeepActionType::Walking;
```
duncanspumpkin, line 1795: How odd. Do you want me to check that in the assembly

duncanspumpkin (changes requested):
See comments

src/openrct2/peep/Staff.cpp
ZehMatt: Would be great

src/openrct2/world/MapAnimation.cpp
ZehMatt: Yeah removing this is actually wrong, got confused over state.

src/openrct2/peep/Staff.cpp
duncanspumpkin: image
looks like it is correct

duncanspumpkin (approved):

PR #14629, opened 2021-05-11
Split off a number of entity files into separate headers

duncanspumpkin:
Eventually this will hopefully reduce the amount of includes we have absolutely everywhere and reduce the amount of times we need to recompile everything.


src/openrct2/world/Sprite.h
```diff
@@ -13,97 +13,17 @@
...
 #include "../peep/Peep.h"
 #include "../ride/Vehicle.h"
+#include "Balloon.h"
+#include "Duck.h"
 #include "Entity.h"
 #include "Fountain.h"
+#include "Litter.h"
+#include "MoneyEffect.h"
 #include "SpriteBase.h"
+#include "Particle.h"
```
tupaschoal, line 23:
```suggestion
#include "Particle.h"
#include "SpriteBase.h"
```
Just to fix formatting and get the CI going

src/openrct2/world/Balloon.h
```diff
@@ -0,0 +1,30 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2020 OpenRCT2 developers
```
tupaschoal, line 2:
```suggestion
 * Copyright (c) 2014-2021 OpenRCT2 developers
```
If we are to be picky hahahaha (I won't open issues for this on the whole PR)

tupaschoal (approved):
Looks ok

PR #14633, opened 2021-05-11
Rename previously unknown vehicle flags

Basssiiie:
This PR gives sensible names to various previously unnamed flags, to make the code base a bit more readable. I used the OpenRCT2 source code, a custom plugin I wrote to read out the flags ingame and [OpenRCT2/objects](https://github.com/OpenRCT2/objects) repository to come to these concluded names. But I am still human, so please correct them if I interpreted them wrong!

Some notes:
- `VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS`
  This is what the related function that is called with this flag seems to be doing, but I could not find an example where the bounds were actually different after the function was called. This flag is used on a lot of rides, including many flatrides, stalls, minigolf, the maze, the reverser rollercoaster
- `VEHICLE_ENTRY_FLAG_SUSPENDED_SWING`, `VEHICLE_ENTRY_FLAG_SLIDE_SWING` and other swinging flags do not always map one to one to rides. The flags help determine the amount of horizontal (swinging?) sprites there are, but there are duplicate results here too. The table is as follows:
  Ride | Flags | Horizontal frames
  -- | -- | --
  Wooden Wild Mouse Coaster  | `SWINGING` + `WOODEN_WILD_MOUSE_SWING` | 3
  Mini Suspended Coaster | `SWINGING` | 5
  Dinghy Slide | `SWINGING` + `SLIDE_SWING` | 7
  Suspended Swinging Coaster | `SWINGING` + `SUSPENDED_SWING` | 7
  Bobsleigh Coaster | `SWINGING` + `SUSPENDED_SWING` + `SLIDE_SWING` | 13

  The flags apply to all vehicle variations of these coasters. As you can see both the Mini Suspended Coaster and the Bobsleigh Coaster do not have their own flags but use either no additional flags or multiple flags from other rides. (PS: While writing this I also noticed the horizontal frame values also kinda look like bit flags as well? Not sure if it is though.)
- For the JSON object mappings I've added new entries for the new flag names and moved the old ones to the bottom. They are still in the table because [OpenRCT2/objects](https://github.com/OpenRCT2/objects) still uses the old flags.
- I saw some people added themselves to the `contributers.md` so I added myself too. 😁

If there are any more questions or issues, please let me know!

### DIFF
--- a/src/openrct2/object/RideObject.cpp
+++ b/src/openrct2/object/RideObject.cpp
@@ -216,9 +216,9 @@ void RideObject::Load()
             {
                 int32_t b = vehicleEntry->base_num_frames * 32;
 
-                if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_11)
+                if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES)
                     b /= 2;
-                if (vehicleEntry->sprite_flags & VEHICLE_SPRITE_FLAG_15)
+                if (vehicleEntry->sprite_flags & VEHICLE_SPRITE_FLAG_USE_4_ROTATION_FRAMES)
                     b /= 8;
 
                 image_index += b;
@@ -338,10 +338,10 @@ void RideObject::Load()
             cur_vehicle_images_offset = image_index + vehicleEntry->no_seating_rows * vehicleEntry->no_vehicle_images;
             // 0x6DEB0D
 
-            if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_10))
+            if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS))
             {
                 int32_t num_images = cur_vehicle_images_offset - vehicleEntry->base_image_id;
-                if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_13)
+                if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SPRITE_BOUNDS_INCLUDE_INVERTED_SET)
                 {
                     num_images *= 2;
                 }
@@ -497,9 +497,10 @@ uint8_t RideObject::CalculateNumHorizontalFrames(const rct_ride_entry_vehicle* v
     uint8_t numHorizontalFrames;
     if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SWINGING)
     {
-        if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_21) && !(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SLIDE_SWING))
+        if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SUSPENDED_SWING)
+            && !(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SLIDE_SWING))
         {
-            if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_25)
+            if (vehicleEntry->flags & VEHICLE_ENTRY_FLAG_WOODEN_WILD_MOUSE_SWING)
             {
                 numHorizontalFrames = 3;
             }
@@ -508,7 +509,9 @@ uint8_t RideObject::CalculateNumHorizontalFrames(const rct_ride_entry_vehicle* v
                 numHorizontalFrames = 5;
             }
         }
-        else if (!(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_21) || !(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SLIDE_SWING))
+        else if (
+            !(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SUSPENDED_SWING)
+            || !(vehicleEntry->flags & VEHICLE_ENTRY_FLAG_SLIDE_SWING))
         {
             numHorizontalFrames = 7;
         }
@@ -805,27 +808,58 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
                 { "corkscrews", VEHICLE_SPRITE_FLAG_CORKSCREWS },
                 { "restraintAnimation", VEHICLE_SPRITE_FLAG_RESTRAINT_ANIMATION },
                 { "curvedLiftHill", VEHICLE_SPRITE_FLAG_CURVED_LIFT_HILL },
-                { "VEHICLE_SPRITE_FLAG_15", VEHICLE_SPRITE_FLAG_15 },
+                { "VEHICLE_SPRITE_FLAG_15", VEHICLE_SPRITE_FLAG_USE_4_ROTATION_FRAMES },
             });
     }
 
     car.flags |= Json::GetFlags<uint32_t>(
         jCar,
         {
+            { "isPoweredRideWithUnrestrictedGravity", VEHICLE_ENTRY_FLAG_POWERED_RIDE_UNRESTRICTED_GRAVITY },
+            { "hasNoUpstopWheels", VEHICLE_ENTRY_FLAG_NO_UPSTOP_WHEELS },
+            { "hasNoUpstopWheelsBobsleigh", VEHICLE_ENTRY_FLAG_NO_UPSTOP_BOBSLEIGH },
+            { "isMiniGolf", VEHICLE_ENTRY_FLAG_MINI_GOLF },
+            { "isReverserBogie", VEHICLE_ENTRY_FLAG_REVERSER_BOGIE },
+            { "isReverserPassengerCar", VEHICLE_ENTRY_FLAG_REVERSER_PASSENGER_CAR },
+            { "hasInvertedSpriteSet", VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET },
+            { "hasDodgemInUseLights", VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS },
+            { "hasAdditionalColour2", VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2 },
+            { "recalculateSpriteBounds", VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS },
+            { "VEHICLE_ENTRY_FLAG_11", VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES },
+            { "overrideNumberOfVerticalFrames", VEHICLE_ENTRY_FLAG_OVERRIDE_NUM_VERTICAL_FRAMES },
+            { "spriteBoundsIncludeInvertedSet", VEHICLE_ENTRY_FLAG_SPRITE_BOUNDS_INCLUDE_INVERTED_SET },
+            { "hasAdditionalSpinningFrames", VEHICLE_ENTRY_FLAG_SPINNING_ADDITIONAL_FRAMES },
+            { "isLift", VEHICLE_ENTRY_FLAG_LIFT },
+            { "hasAdditionalColour1", VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_1 },
+            { "hasSwinging", VEHICLE_ENTRY_FLAG_SWINGING },
+            { "hasSpinning", VEHICLE_ENTRY_FLAG_SPINNING },
+            { "isPowered", VEHICLE_ENTRY_FLAG_POWERED },
+            { "hasScreamingRiders", VEHICLE_ENTRY_FLAG_RIDERS_SCREAM },
+            { "useSuspendedSwing", VEHICLE_ENTRY_FLAG_SUSPENDED_SWING },
+            { "useBoatHireCollisionDetection", VEHICLE_ENTRY_FLAG_BOAT_HIRE_COLLISION_DETECTION },
+            { "hasVehicleAnimation", VEHICLE_ENTRY_FLAG_VEHICLE_ANIMATION },
+            { "hasRiderAnimation", VEHICLE_ENTRY_FLAG_RIDER_ANIMATION },
+            { "useWoodenWildMouseSwing", VEHICLE_ENTRY_FLAG_WOODEN_WILD_MOUSE_SWING },
+            { "useSlideSwing", VEHICLE_ENTRY_FLAG_SLIDE_SWING },
+            { "isChairlift", VEHICLE_ENTRY_FLAG_CHAIRLIFT },
+            { "isWaterRide", VEHICLE_ENTRY_FLAG_WATER_RIDE },
+            { "isGoKart", VEHICLE_ENTRY_FLAG_GO_KART },
+            { "useDodgemCarPlacement", VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT },
+
+            // Obsolete flags that have been replaced with named and camelCased values, but kept for now for backward
+            // compatibility.
             { "VEHICLE_ENTRY_FLAG_POWERED_RIDE_UNRESTRICTED_GRAVITY", VEHICLE_ENTRY_FLAG_POWERED_RIDE_UNRESTRICTED_GRAVITY },
             { "VEHICLE_ENTRY_FLAG_NO_UPSTOP_WHEELS", VEHICLE_ENTRY_FLAG_NO_UPSTOP_WHEELS },
             { "VEHICLE_ENTRY_FLAG_NO_UPSTOP_BOBSLEIGH", VEHICLE_ENTRY_FLAG_NO_UPSTOP_BOBSLEIGH },
             { "VEHICLE_ENTRY_FLAG_MINI_GOLF", VEHICLE_ENTRY_FLAG_MINI_GOLF },
-            { "VEHICLE_ENTRY_FLAG_4", VEHICLE_ENTRY_FLAG_4 },
-            { "VEHICLE_ENTRY_FLAG_5", VEHICLE_ENTRY_FLAG_5 },
+            { "VEHICLE_ENTRY_FLAG_4", VEHICLE_ENTRY_FLAG_REVERSER_BOGIE },
+            { "VEHICLE_ENTRY_FLAG_5", VEHICLE_ENTRY_FLAG_REVERSER_PASSENGER_CAR },
             { "VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET", VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET },
             { "VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS", VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS },
-            { "VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED", VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED },
             { "VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2", VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2 },
-            { "VEHICLE_ENTRY_FLAG_10", VEHICLE_ENTRY_FLAG_10 },
-            { "VEHICLE_ENTRY_FLAG_11", VEHICLE_ENTRY_FLAG_11 },
+            { "VEHICLE_ENTRY_FLAG_10", VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS },
             { "VEHICLE_ENTRY_FLAG_OVERRIDE_NUM_VERTICAL_FRAMES", VEHICLE_ENTRY_FLAG_OVERRIDE_NUM_VERTICAL_FRAMES },
-            { "VEHICLE_ENTRY_FLAG_13", VEHICLE_ENTRY_FLAG_13 },
+            { "VEHICLE_ENTRY_FLAG_13", VEHICLE_ENTRY_FLAG_SPRITE_BOUNDS_INCLUDE_INVERTED_SET },
             { "VEHICLE_ENTRY_FLAG_SPINNING_ADDITIONAL_FRAMES", VEHICLE_ENTRY_FLAG_SPINNING_ADDITIONAL_FRAMES },
             { "VEHICLE_ENTRY_FLAG_LIFT", VEHICLE_ENTRY_FLAG_LIFT },
             { "VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_1", VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_1 },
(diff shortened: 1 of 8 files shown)


src/openrct2/ride/Vehicle.h
```diff
@@ -379,20 +379,24 @@ enum : uint32_t
...
     VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET = 1 << 6, // Set on vehicles that support running inverted for extended periods
                                                          // of time, i.e. the Flying, Lay-down and Multi-dimension RCs.
     VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS = 1
         << 7, // When set the vehicle has an additional frame for when in use. Used only by dodgems.
     VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED = 1 << 8, // Not used any more - every vehicle will now work with doors
     VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2 = 1 << 9,
-    VEHICLE_ENTRY_FLAG_10 = 1 << 10,
-    VEHICLE_ENTRY_FLAG_11 = 1 << 11,
+    VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS = 1 << 10, // Only used during loading of the objects
+    VEHICLE_ENTRY_FLAG_HALF_ROTATION_SPRITE_COUNT = 1
```
duncanspumpkin, line 391: This doesn't make it clear that it only effects flat sprites. I would rather it was called 16_FLAT_ROTATION_FRAMES

src/openrct2/ride/Vehicle.h
```diff
@@ -470,7 +475,8 @@ enum : uint32_t
     VEHICLE_SPRITE_FLAG_CORKSCREWS = (1 << 12),
     VEHICLE_SPRITE_FLAG_RESTRAINT_ANIMATION = (1 << 13),
     VEHICLE_SPRITE_FLAG_CURVED_LIFT_HILL = (1 << 14),
-    VEHICLE_SPRITE_FLAG_15 = (1 << 15),
+    // Used only on lifts (the transport ride), to only use 4 rotation sprites instead of 32.
+    VEHICLE_SPRITE_FLAG_ROTATION_COUNT_DIVIDE_BY_EIGHT = (1 << 15),
```
duncanspumpkin, line 479: Again i would call it 4_FLAT_ROTATION_FRAMES

duncanspumpkin:
The divide by amount is just what you need to do to convert a higher precision rotation to a lower one. The flags really just say how many frames there are not that you need to divide.

src/openrct2/ride/Vehicle.h
Gymnasiast: > This doesn't make it clear that it only effects flat sprites.

Yes, but then, I think it should affect all places where it currently expects 32 rotations. Still though, using the number 16 explicitly is a good idea.

I thus propose `VEHICLE_ENTRY_FLAG_16_ROTATION_FRAMES`.

src/openrct2/ride/Vehicle.h
Gymnasiast: `VEHICLE_SPRITE_FLAG_4_ROTATION_FRAMES`

src/openrct2/object/RideObject.cpp
```diff
@@ -816,33 +822,42 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
-            { "VEHICLE_ENTRY_FLAG_25", VEHICLE_ENTRY_FLAG_25 },
+            { "VEHICLE_ENTRY_FLAG_WOODEN_WILD_MOUSE_SWING", VEHICLE_ENTRY_FLAG_WOODEN_WILD_MOUSE_SWING },
             { "VEHICLE_ENTRY_FLAG_SLIDE_SWING", VEHICLE_ENTRY_FLAG_SLIDE_SWING },
             { "VEHICLE_ENTRY_FLAG_CHAIRLIFT", VEHICLE_ENTRY_FLAG_CHAIRLIFT },
             { "VEHICLE_ENTRY_FLAG_WATER_RIDE", VEHICLE_ENTRY_FLAG_WATER_RIDE },
             { "VEHICLE_ENTRY_FLAG_GO_KART", VEHICLE_ENTRY_FLAG_GO_KART },
             { "VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT", VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT },
+
+            // Obsolete flags that have been replaced with names values, but kept for backward compatibility.
+            { "VEHICLE_ENTRY_FLAG_4", VEHICLE_ENTRY_FLAG_REVERSER_BOGIE },
```
Gymnasiast, line 854: When this is merged, could you create a pull request against `objects` that changes this for all existing JSON files and for the script that converts .DAT files (just a find-and-replace on the whole repo will probably do).

src/openrct2/ride/Vehicle.h
duncanspumpkin: Yes but it's specifically for flat sprites. In locomotion CS had separate fields for how many flat rotation sprites and how many sloped rotation sprites. This allows for a smooth rotation in the vehicle preview page whilst also not needing to generate all the sloped rotation frames.

Gymnasiast: > Yes but it's specifically for flat sprites

One can argue either way, since it is only set on vehicles that only have flat sprites to begin with.

duncanspumpkin: I guess the proper question to ask is what do object makers want? Would they want the ability to specify how many rotation sprites there are for flat and specify a different amount for sloped?

Basssiiie: Ideally this flag would be replaced in the future by something more alike to Locomotion.

My thoughts with the name of flag and `VEHICLE_SPRITE_FLAG_ROTATION_COUNT_DIVIDE_BY_EIGHT` is that it seems that they can be used together to only use 2 rotation sprites, instead of 4 or 16. 

But if that is something you guys don't wanna support (no vanilla rides use them both together), then my proposal would be to change them to `VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES` and `VEHICLE_SPRITE_FLAG_USE_4_ROTATION_FRAMES` (similar to Gymnasiasts proposal but including the word 'use' as well for easier readability).

src/openrct2/object/RideObject.cpp
Basssiiie: Sure!

Gymnasiast:
For JSON objects, we want to replace the flag with a field. Can you keep the old names for JSON objects only?

src/openrct2/ride/Vehicle.h
Gymnasiast: Hm, I prefer keeping to the slightly shorter one.

What do you think @duncanspumpkin ?

duncanspumpkin: Okay we are going to go with the suggestion of not including FLAT in the name. I dont mind too much about USE_ i guess it makes sense. We will modify the codebase so that this flag effects sloped sprites as well but in a seperate PR. Yes we do not support both flags being set.

Basssiiie:
> For JSON objects, we want to replace the flag with a field. Can you keep the old names for JSON objects only?

You mean revert the changed strings in the JSON maps in `RideObject.cpp`? All of them or only the rotation flags? And if all strings need to be reverted, then a second PR to OpenRCT2/objects repository would not be necessary I suppose?

Or do you mean something else?

src/openrct2/ride/Vehicle.h
Basssiiie: Alright, I've updated them accordingly. :)

src/openrct2/object/RideObject.cpp
```diff
@@ -816,33 +822,42 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
+            { "VEHICLE_ENTRY_FLAG_REVERSER_BOGIE", VEHICLE_ENTRY_FLAG_REVERSER_BOGIE },
+            { "VEHICLE_ENTRY_FLAG_REVERSER_PASSENGER_CAR", VEHICLE_ENTRY_FLAG_REVERSER_PASSENGER_CAR },
             { "VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET", VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET },
             { "VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS", VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS },
             { "VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED", VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED },
             { "VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2", VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2 },
-            { "VEHICLE_ENTRY_FLAG_10", VEHICLE_ENTRY_FLAG_10 },
-            { "VEHICLE_ENTRY_FLAG_11", VEHICLE_ENTRY_FLAG_11 },
+            { "VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS", VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS },
+            { "VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES", VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES },
```
Gymnasiast, line 832:
```suggestion
            { "VEHICLE_ENTRY_FLAG_11", VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES },
```

src/openrct2/object/RideObject.cpp
```diff
@@ -805,7 +808,10 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
                 { "corkscrews", VEHICLE_SPRITE_FLAG_CORKSCREWS },
                 { "restraintAnimation", VEHICLE_SPRITE_FLAG_RESTRAINT_ANIMATION },
                 { "curvedLiftHill", VEHICLE_SPRITE_FLAG_CURVED_LIFT_HILL },
-                { "VEHICLE_SPRITE_FLAG_15", VEHICLE_SPRITE_FLAG_15 },
+                { "use4RotationFrames", VEHICLE_SPRITE_FLAG_USE_4_ROTATION_FRAMES },
```
Gymnasiast, line 811:
```suggestion
                { "VEHICLE_SPRITE_FLAG_15", VEHICLE_SPRITE_FLAG_USE_4_ROTATION_FRAMES },
```

src/openrct2/object/RideObject.cpp
```diff
@@ -805,7 +808,10 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
+
+                // Obsolete, but kept for backward compatibility
+                { "VEHICLE_SPRITE_FLAG_15", VEHICLE_SPRITE_FLAG_USE_4_ROTATION_FRAMES },
```
Gymnasiast, line 814: revert

src/openrct2/object/RideObject.cpp
```diff
@@ -816,33 +822,42 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
             { "VEHICLE_ENTRY_FLAG_CHAIRLIFT", VEHICLE_ENTRY_FLAG_CHAIRLIFT },
             { "VEHICLE_ENTRY_FLAG_WATER_RIDE", VEHICLE_ENTRY_FLAG_WATER_RIDE },
             { "VEHICLE_ENTRY_FLAG_GO_KART", VEHICLE_ENTRY_FLAG_GO_KART },
             { "VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT", VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT },
+
+            // Obsolete flags that have been replaced with named values, but kept for backward compatibility.
+            { "VEHICLE_ENTRY_FLAG_4", VEHICLE_ENTRY_FLAG_REVERSER_BOGIE },
+            { "VEHICLE_ENTRY_FLAG_5", VEHICLE_ENTRY_FLAG_REVERSER_PASSENGER_CAR },
+            { "VEHICLE_ENTRY_FLAG_10", VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS },
+            { "VEHICLE_ENTRY_FLAG_11", VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES },
```
Gymnasiast, line 857: revert (only this line)

Gymnasiast (commented):
I have marked all the changes I meant.

Basssiiie:
Got it. Just wanted to make sure I don't walk into a misunderstanding again. :)

src/openrct2/object/RideObject.cpp
```diff
@@ -816,33 +819,41 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
             { "VEHICLE_ENTRY_FLAG_BOAT_HIRE_COLLISION_DETECTION", VEHICLE_ENTRY_FLAG_BOAT_HIRE_COLLISION_DETECTION },
             { "VEHICLE_ENTRY_FLAG_VEHICLE_ANIMATION", VEHICLE_ENTRY_FLAG_VEHICLE_ANIMATION },
             { "VEHICLE_ENTRY_FLAG_RIDER_ANIMATION", VEHICLE_ENTRY_FLAG_RIDER_ANIMATION },
-            { "VEHICLE_ENTRY_FLAG_25", VEHICLE_ENTRY_FLAG_25 },
+            { "VEHICLE_ENTRY_FLAG_WOODEN_WILD_MOUSE_SWING", VEHICLE_ENTRY_FLAG_WOODEN_WILD_MOUSE_SWING },
             { "VEHICLE_ENTRY_FLAG_SLIDE_SWING", VEHICLE_ENTRY_FLAG_SLIDE_SWING },
             { "VEHICLE_ENTRY_FLAG_CHAIRLIFT", VEHICLE_ENTRY_FLAG_CHAIRLIFT },
             { "VEHICLE_ENTRY_FLAG_WATER_RIDE", VEHICLE_ENTRY_FLAG_WATER_RIDE },
             { "VEHICLE_ENTRY_FLAG_GO_KART", VEHICLE_ENTRY_FLAG_GO_KART },
             { "VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT", VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT },
```
Gymnasiast, line 848: Come to think of it, we should probably rename all these flags in the JSON files as well to something use `camelCase`. What do you think @IntelOrca ?

IntelOrca: I don't mind

Basssiiie: I can rename them to use camelCase yes. All of them or just the ones I renamed? Should I camelCase the rotation flags as well then or keep them as is?

Gymnasiast: Keep the names of `VEHICLE_SPRITE_FLAG_15` and `VEHICLE_ENTRY_FLAG_11` as is (as we're going to replace them anyway, change everything else.

For the names, I was thinking along these lines:

VEHICLE_ENTRY_FLAG_POWERED -> isPowered
VEHICLE_ENTRY_FLAG_CHAIRLIFT -> isChairlift
VEHICLE_ENTRY_FLAG_DODGEM_CAR_PLACEMENT -> useDodgemCarPlacement
VEHICLE_ENTRY_FLAG_NO_UPSTOP_WHEELS -> hasNoUpstopWheels

Basssiiie: Alright, I'll update them like that somewhere today/tonight when possible! :)

Basssiiie: They're updated! But if there are still any nitpicks for the names, please let me know. :)

src/openrct2/object/RideObject.cpp
```diff
@@ -805,39 +808,72 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
+            { "isLift", VEHICLE_ENTRY_FLAG_LIFT },
+            { "hasAdditionalColour1", VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_1 },
+            { "hasSwinging", VEHICLE_ENTRY_FLAG_SWINGING },
+            { "hasSpinning", VEHICLE_ENTRY_FLAG_SPINNING },
+            { "isPowered", VEHICLE_ENTRY_FLAG_POWERED },
+            { "hasScreamingRiders", VEHICLE_ENTRY_FLAG_RIDERS_SCREAM },
+            { "useSuspendedSwing", VEHICLE_ENTRY_FLAG_SUSPENDED_SWING },
+            { "useBoatHireCollisionDetection", VEHICLE_ENTRY_FLAG_BOAT_HIRE_COLLISION_DETECTION },
+            { "useVehicleAnimation", VEHICLE_ENTRY_FLAG_VEHICLE_ANIMATION },
+            { "useRiderAnimation", VEHICLE_ENTRY_FLAG_RIDER_ANIMATION },
```
Gymnasiast, line 842:
```suggestion
            { "hasVehicleAnimation", VEHICLE_ENTRY_FLAG_VEHICLE_ANIMATION },
            { "hasRiderAnimation", VEHICLE_ENTRY_FLAG_RIDER_ANIMATION },
```

src/openrct2/object/RideObject.cpp
```diff
@@ -805,39 +808,72 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
+            { "useWoodenWildMouseSwing", VEHICLE_ENTRY_FLAG_WOODEN_WILD_MOUSE_SWING },
+            { "useSlideSwing", VEHICLE_ENTRY_FLAG_SLIDE_SWING },
+            { "isChairLift", VEHICLE_ENTRY_FLAG_CHAIRLIFT },
```
Gymnasiast, line 845:
```suggestion
            { "isChairlift", VEHICLE_ENTRY_FLAG_CHAIRLIFT },
```

src/openrct2/object/RideObject.cpp
```diff
@@ -805,39 +808,72 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
         {
+            { "isPoweredRideWithUnrestrictedGravity", VEHICLE_ENTRY_FLAG_POWERED_RIDE_UNRESTRICTED_GRAVITY },
+            { "hasNoUpstopWheels", VEHICLE_ENTRY_FLAG_NO_UPSTOP_WHEELS },
+            { "hasNoUpstopBobsleigh", VEHICLE_ENTRY_FLAG_NO_UPSTOP_BOBSLEIGH },
+            { "isMiniGolf", VEHICLE_ENTRY_FLAG_MINI_GOLF },
+            { "isReverserBogie", VEHICLE_ENTRY_FLAG_REVERSER_BOGIE },
+            { "isReverserPassengerCar", VEHICLE_ENTRY_FLAG_REVERSER_PASSENGER_CAR },
+            { "hasInvertedSpriteSet", VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET },
+            { "hasDodgemInUseLights", VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS },
+            { "allowDoorsDeprecated", VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED },
```
Gymnasiast, line 826: Get rid of this line - we don't use this flag any more. Also get rid of the "old line".

Basssiiie: The old one is still in [OpenRCT2/objects](https://github.com/OpenRCT2/objects/search?q=VEHICLE_ENTRY_FLAG_ALLOW_DOORS_DEPRECATED) though. Will the game be find if the flag cannot be found?

Basssiiie: Update: the game doesn't show any errors and Haunted Harbor still looks good, so it seems fine to remove the old one too. :)

src/openrct2/object/RideObject.cpp
```diff
@@ -805,39 +808,70 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
+                { "VEHICLE_SPRITE_FLAG_15", VEHICLE_SPRITE_FLAG_USE_4_ROTATION_FRAMES },
             });
     }
 
     car.flags |= Json::GetFlags<uint32_t>(
         jCar,
         {
+            { "isPoweredRideWithUnrestrictedGravity", VEHICLE_ENTRY_FLAG_POWERED_RIDE_UNRESTRICTED_GRAVITY },
+            { "hasNoUpstopWheels", VEHICLE_ENTRY_FLAG_NO_UPSTOP_WHEELS },
+            { "hasNoUpstopBobsleigh", VEHICLE_ENTRY_FLAG_NO_UPSTOP_BOBSLEIGH },
```
Gymnasiast, line 820:
```suggestion
            { "hasNoUpstopWheelsBobsleigh", VEHICLE_ENTRY_FLAG_NO_UPSTOP_BOBSLEIGH },
```

src/openrct2/object/RideObject.cpp
```diff
@@ -805,39 +808,70 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
         jCar,
         {
+            { "isPoweredRideWithUnrestrictedGravity", VEHICLE_ENTRY_FLAG_POWERED_RIDE_UNRESTRICTED_GRAVITY },
+            { "hasNoUpstopWheels", VEHICLE_ENTRY_FLAG_NO_UPSTOP_WHEELS },
+            { "hasNoUpstopWheelsBobsleigh", VEHICLE_ENTRY_FLAG_NO_UPSTOP_BOBSLEIGH },
+            { "isMiniGolf", VEHICLE_ENTRY_FLAG_MINI_GOLF },
+            { "isReverserBogie", VEHICLE_ENTRY_FLAG_REVERSER_BOGIE },
+            { "isReverserPassengerCar", VEHICLE_ENTRY_FLAG_REVERSER_PASSENGER_CAR },
+            { "hasInvertedSpriteSet", VEHICLE_ENTRY_FLAG_HAS_INVERTED_SPRITE_SET },
+            { "hasDodgemInUseLights", VEHICLE_ENTRY_FLAG_DODGEM_INUSE_LIGHTS },
```
Gymnasiast, line 825: (check)

src/openrct2/object/RideObject.cpp
```diff
@@ -805,39 +808,70 @@ rct_ride_entry_vehicle RideObject::ReadJsonCar(json_t& jCar)
...
+            { "hasAdditionalColour2", VEHICLE_ENTRY_FLAG_ENABLE_ADDITIONAL_COLOUR_2 },
+            { "recalculateSpriteBounds", VEHICLE_ENTRY_FLAG_RECALCULATE_SPRITE_BOUNDS },
+            { "VEHICLE_ENTRY_FLAG_11", VEHICLE_ENTRY_FLAG_USE_16_ROTATION_FRAMES },
+            { "overrideNumberOfVerticalFrames", VEHICLE_ENTRY_FLAG_OVERRIDE_NUM_VERTICAL_FRAMES },
+            { "spriteBoundsIncludeInvertedSet", VEHICLE_ENTRY_FLAG_SPRITE_BOUNDS_INCLUDE_INVERTED_SET },
```
Gymnasiast, line 830: (check)

PR #14636, opened 2021-05-12
Add climate information to Park API

rctdude2:
Adds fields to the Park API to get the climate and weather. I needed methods to convert the types to strings for the API, and I put them in world/Climate since that made the most sense but I can move them if they're better in a different location. Also, I used camelCase for the strings, but I wasn't sure what to use because the API doesn't seem to be consistent on casing (some things use camelCase, some use snake_case, and the actions are all lowercase), please let me know if they should be a different case.

### DIFF
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -37,6 +37,8 @@ declare global {
     var park: Park;
     /** APIs for the current scenario. */
     var scenario: Scenario;
+    /** APIs for the climate and weather. */
+    var climate: Climate;
     /**
      * APIs for creating and editing title sequences.
      * These will only be available to clients that are not running headless mode.
@@ -1601,23 +1603,6 @@ declare global {
         "scenarioCompleteNameInput" |
         "unlockAllPrices";
 
-    type ClimateType =
-        "coolAndWet" |
-        "warm" |
-        "hotAndDry" |
-        "cold";
-
-    type WeatherType =
-        "sunny" |
-        "partiallyCloudy" |
-        "cloudy" |
-        "rain" |
-        "heavyRain" |
-        "thunder" |
-        "snow" |
-        "heavySnow" |
-        "blizzard";
-
     interface Park {
         cash: number;
         rating: number;
@@ -1712,26 +1697,6 @@ declare global {
          * Updated every 4096 ticks.
          */
         readonly parkSize: number;
-
-        /**
-         * The climate of the park.
-         */
-        readonly climate: ClimateType;
-
-        /**
-         * The current weather in the park.
-         */
-        readonly currentWeather: WeatherType;
-
-        /**
-         * The current temperature in the park.
-         */
-        readonly currentTemperature: number;
-
-        /**
-         * The next weather the park will experience.
-         */
-        readonly futureWeather: WeatherType;
 
         name: string;
         messages: ParkMessage[];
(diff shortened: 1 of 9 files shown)


rctdude2:
Looks like other platforms want the Count value explicitly specified, should I add that and just return ""?

IntelOrca:
I am not sure climate properties belong to Park. I think it should be its own thing. E.g. `climate.current.temperature`.

rctdude2:
If I go ahead and move this into an ScClimate.hpp file, is there anything else I need to do to make the new file work?

IntelOrca:
@rctdude2 You'll need to register Climate as a new global, search for where this is done for things like Park, Map etc. and repeat changes.

IntelOrca:
current and future should be sub objects as they contain the same fields.

src/openrct2/scripting/ScClimate.hpp
```diff
@@ -0,0 +1,109 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2020 OpenRCT2 developers
```
tupaschoal, line 2:
```suggestion
 * Copyright (c) 2014-2021 OpenRCT2 developers
```

src/openrct2/scripting/ScClimate.hpp
```diff
@@ -0,0 +1,109 @@
...
+                case ClimateType::CoolAndWet:
+                    return "coolAndWet";
+                case ClimateType::Warm:
+                    return "warm";
+                case ClimateType::HotAndDry:
+                    return "hotAndDry";
+                case ClimateType::Cold:
+                    return "cold";
+                case ClimateType::Count:
+                    return "";
```
tupaschoal, line 39:
```suggestion
                    return std::string();
```

tupaschoal, line 39:
```suggestion
                    return std::string();
```

src/openrct2/scripting/ScClimate.hpp
```diff
@@ -0,0 +1,109 @@
...
+            }
+            return "";
```
tupaschoal, line 41:
```suggestion
            return std::string();
```

src/openrct2/scripting/ScClimate.hpp
```diff
@@ -0,0 +1,109 @@
...
+                case WeatherType::Thunder:
+                    return "thunder";
+                case WeatherType::Snow:
+                    return "snow";
+                case WeatherType::HeavySnow:
+                    return "heavySnow";
+                case WeatherType::Blizzard:
+                    return "blizzard";
+                case WeatherType::Count:
+                    return "";
```
tupaschoal, line 67:
```suggestion
                    return std::string();
```

src/openrct2/scripting/ScClimate.hpp
```diff
@@ -0,0 +1,109 @@
...
+            }
+            return "";
```
tupaschoal, line 69:
```suggestion
            return std::string();
```

tupaschoal (commented):
I'd rather use `std::string()` than returning `""`, but I know some people don't, so take it with a grain of salt.

rctdude2:
I left them as "" for now since other things under scripting/ use that.

tupaschoal (approved):

distribution/openrct2.d.ts
```diff
@@ -1819,6 +1821,45 @@ declare global {
...
+    interface Climate {
+        /**
+         * The climate of the park.
+         */
+        readonly climate: ClimateType;
+
+        /**
+         * The current weather in the park.
+         */
+        readonly currentWeather: ClimateState;
```
IntelOrca, line 1855: I think this should just be `current`, otherwise you end up writing `currentWeather.weather`.

distribution/openrct2.d.ts
```diff
@@ -1819,6 +1821,45 @@ declare global {
...
+    interface ClimateState {
+        readonly weather: WeatherType;
+        readonly temperature: number;
+    }
+
+    interface Climate {
+        /**
+         * The climate of the park.
+         */
+        readonly climate: ClimateType;
```
IntelOrca, line 1850: This should just be type, otherwise you end up writing `Climate.climate`.

IntelOrca (approved):

Issue #4896, opened 2016-12-20
Sprite interpolation (from PNG) may use the dynamic water colors


LRFLEW:
**OS:** macOS (but should apply to all)
**Version:** 0.0.5
**Commit/Build:** 2fca2ad

I discovered this while experimenting with using [this](https://www.reddit.com/r/rct/comments/5iwzyf/openrct2_heres_a_logo_for_the_game_that_i_made/) with the project. I started with [this](https://cloud.githubusercontent.com/assets/607715/21334203/2b66b544-c61b-11e6-9658-a7eeec6e3cdf.png) (which was a scaled-down version of [this full size version](https://cloud.githubusercontent.com/assets/607715/21334208/37733b46-c61b-11e6-8aa0-d556d9a53234.png)), and used the `sprite append` with `--mode dithering` and `sprite export` commands to get [this](https://cloud.githubusercontent.com/assets/607715/21334246/7a5de762-c61b-11e6-8990-fa3aa4836bac.png). When I tried replacing the 0.png file in `resources/g2`, I noticed that some of the dithered colors towards the bottom of the logo were flashing. I realized that this was happening in-sync with the water's sparkling effect, indicating that it's using colors from that palette. The interpolation options (`closest` and `dithering`) should probably be made to exclude the color indices used for water.

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Multiplayer?

**Steps to reproduce:**
1. Replace `resources/g2/0.png` with [this](https://cloud.githubusercontent.com/assets/607715/21334246/7a5de762-c61b-11e6-8990-fa3aa4836bac.png)
    * To generate this png, start with [this](https://cloud.githubusercontent.com/assets/607715/21334203/2b66b544-c61b-11e6-9658-a7eeec6e3cdf.png) and do the following (make sure to use appropriate file paths):
    1. `./openrct2 sprite append data/g2.dat above_image.png --mode dithering`
    2. `./openrct2 sprite export data/g2.dat 80 resources/g2/0.png`
2. Re-build g2.dat and run the game
3. Look at the logo on the title screen

**Screenshots / Video:**
![sprite_error](https://cloud.githubusercontent.com/assets/607715/21334368/86e28b18-c61c-11e6-86f0-4ece36f8db64.gif)

PR #14640, opened 2021-05-13
Correct palette indexes of remap colors in image importer

zrowny:
The code that finds the closest entry that isn't used for remapping has the indices of remappable colors slightly off, and would make some colors in the imported image remappable even though they shouldn't have been.

The header file also has a comment that had the wrong number for the remappable color indices.

### DIFF
--- a/src/openrct2/drawing/ImageImporter.h
+++ b/src/openrct2/drawing/ImageImporter.h
@@ -273,7 +273,7 @@ constexpr const GamePalette StandardPalette = { {
     { 231, 231, 171, 255 },
     { 255, 255, 207, 255 },
 
-    // 203 - 214 (Secondary remap)
+    // 202 - 214 (Secondary remap)
     { 27, 0, 63, 255 },
     { 51, 0, 103, 255 },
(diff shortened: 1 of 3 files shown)


src/openrct2/drawing/ImageImporter.h
```diff
@@ -273,7 +273,7 @@ constexpr const GamePalette StandardPalette = { {
     { 231, 231, 171, 255 },
     { 255, 255, 207, 255 },
 
-    // 203 - 214 (Secondary remap)
+    // 202 - 214 (Secondary remap)
```
Gymnasiast, line 276:
```suggestion
    // 202 - 213 (Secondary remap)
```

tupaschoal (commented):
Nice catch :O

Gymnasiast:
I wonder if this would solve https://github.com/OpenRCT2/OpenRCT2/issues/4896 .

zrowny:
> I wonder if this would solve #4896 .

This made me realize there are actually different problems with the way this works. See #14649

PR #14642, opened 2021-05-13
Clean up shop item prices

Gymnasiast:


tupaschoal (approved):
I checked them all and it looks fine. Android build had failed on download so I retriggered the CI, if all is green it looks good to go.

PR #14644, opened 2021-05-13
Simplify code for Peep easter eggs

ZehMatt:
This cleans up the easter egg code a bit, the functions are now passed via template argument rather than function pointer, this allows the compiler to inline. I also moved out the location check out of the easter egg functions, this was previously done each iteration of nearby peeps. I know this is a micro optimization but with soon lifted limits small things accumulate quickly.

Develop:
devenv_2021-05-13_18-36-24
It had to load the pointer and does a call later on, not optimal.

PR:
devenv_2021-05-13_18-44-17

All the easter egg functions inlined now.

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -47,6 +47,7 @@
 #include "Staff.h"
 
 #include <algorithm>
+#include <functional>
 #include <iterator>
 
 using namespace OpenRCT2;
@@ -431,11 +432,11 @@ template<> bool SpriteBase::Is<Guest>() const
     return Type == EntityType::Guest;
 }
 
-bool Guest::GuestHasValidXY() const
+static bool IsValidLocation(const CoordsXYZ& coords)
 {
-    if (x != LOCATION_NULL)
+    if (coords.x != LOCATION_NULL)
     {
-        if (map_is_location_valid({ x, y }))
+        if (map_is_location_valid(coords))
         {
             return true;
         }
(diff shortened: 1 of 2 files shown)


janisozaur:
any `benchsimulate` results?

ZehMatt:
I didn't bother, this cleans up the code a bit so its a win win best case.

ZehMatt:
I ran benchsimulate but I get different results each run, in the tolerance of a few ns up to 3000. It seems to be identical? I can't tell.

src/openrct2/peep/Guest.cpp
```diff
@@ -441,17 +441,23 @@ bool Guest::GuestHasValidXY() const
...
+        if (otherGuest == guest)
+        {
+            // Can not apply effect on self.
+            continue;
+        }
+        auto zDiff = std::abs(otherGuest->z - guestLoc.z);
         if (zDiff <= 32)
         {
-            (*this.*easter_egg)(otherGuest);
+            (*guest.*F)(otherGuest);
```
CookiePLMonster, line 460: If I'm not mistaken*, with `std::invoke` you should be able to simplify that awkward notation and `F` could potentially become a `typename Pred`, much like in many std functions.

*I haven't tested that.

ZehMatt: I used a strict type so its clear from the beginning what this function expects. std::invoke is a good point.

src/openrct2/peep/Guest.cpp
```diff
@@ -441,17 +442,23 @@ bool Guest::GuestHasValidXY() const
     return false;
 }
 
-void Guest::ApplyEasterEggToNearbyGuests(easter_egg_function easter_egg)
+template<void (Guest::*F)(Guest*)> static void ApplyEasterEggToNearbyGuests(Guest* guest)
```
duncanspumpkin, line 445: Could you use a more descriptive identifier rather than F `EasterEggFunc`?

duncanspumpkin (approved):

PR #14651, opened 2021-05-14
Fix issues with Image Importer

### LINKED ISSUE #14649: ImageImporter incorrectly  remaps colors outside the RCT2 palette
I already addressed one issue with the ImageImporter in #14640 but there's actually more issues with how ImageImporter imports images with colors outside the RCT2 palette, other than the palette indices just being off by one in a couple places.

First, the third remap colors are wrong anyway (it should be the entire first set of yellows, those extra three at the end are just extra colors and aren't used for remapping). And more importantly to #4896 it doesn't distinguish any of the other special colors used in the game, like the special water colors or the special colors at the start of the palette.

I actually think it's fine for the Importer to re-palettize colors into remap 2&3 (yellows and pinks). If your image isn't already in the right palette, you probably aren't using that image for something where those remap colors matter. But those other special colors should only be used if you actually intend to use them, and I think that's what happened there.

If no one has any problems with this I can probably work on it and get a PR to fix this later today.

zrowny:
Closes #14649: ImageImporter incorrectly remaps colors outside the RCT2 palette

When converting to the internal palette, if a color is not in the palette, the ImageImporter now re-maps it only to those non "special" color indices ("special," meaning those for water effects, font, chain lift and so on, as well as the primary remappable color).

Additionally, if the Importer is told to dither the image, it now does not dither colors that are already in the OpenRCT2 palette, or dither between remappable colors and non-remappable colors.

### DIFF
--- a/src/openrct2/drawing/ImageImporter.h
+++ b/src/openrct2/drawing/ImageImporter.h
@@ -65,10 +65,10 @@ namespace OpenRCT2::Drawing
 } // namespace OpenRCT2::Drawing
 
 constexpr const GamePalette StandardPalette = { {
-    // 0 (unused)
+    // 0 (Unused/Transparent)
     { 0, 0, 0, 255 },
 
-    // 1 - 9 (misc. e.g. font and water)
+    // 1 - 9 (Misc. e.g. font, water, chain lift)
     { 1, 1, 1, 255 },
     { 2, 2, 2, 255 },
     { 3, 3, 3, 255 },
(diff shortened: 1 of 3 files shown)


src/openrct2/drawing/ImageImporter.h
```diff
@@ -301,27 +342,29 @@ constexpr const GamePalette StandardPalette = { {
...
+
+    // 235 - 249 (Water sparkles)
     { 151, 155, 55, 255 },
     { 151, 155, 55, 255 },
     { 227, 227, 155, 255 },
     { 203, 203, 115, 255 },
     { 151, 155, 55, 255 },
 
-    // 240 - 242 (chain lift)
+    // 240 - 242 (Exra grey)
```
Gymnasiast, line 367:
```suggestion
    // 240 - 242 (Extra grey)
```

Gymnasiast (approved):

PR #14656, opened 2021-05-14
[NSF] Fix various implicit type casting warnings

Broxzier:
These warnings caused compilation to fail on VS2019 Version 16.10.0 Preview 2.1, because warnings are being treated as errors.

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -1338,7 +1338,7 @@ private:
         //                                          index in the array ----^     ^--- bit position in the 8-bit value
         // We do the opposite in this function to recover the x and y values.
 
-        int32_t peepOffset = staffId * RCT12_PATROL_AREA_SIZE;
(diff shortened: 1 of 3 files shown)


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1338,7 +1338,7 @@ class S4Importer final : public IParkImporter
         //                                          index in the array ----^     ^--- bit position in the 8-bit value
         // We do the opposite in this function to recover the x and y values.
 
-        int32_t peepOffset = staffId * RCT12_PATROL_AREA_SIZE;
+        int32_t peepOffset = (int32_t)staffId * RCT12_PATROL_AREA_SIZE;
```
Gymnasiast, line 1341: We don't use C-style cast. Please do either int32_t(staffId) or a static cast.

IntelOrca (approved):

tupaschoal (approved):

Issue #14468, opened 2021-04-13
[Android] Unable to close the options window


adammw:
**OS:** Android 11, Samsung OneUI 3.1
**Version:** 0.3.3
**Commit/Build:** Unable to find

Unable to close the options window. Other windows close fine. Tried increasing UI scale in case it was a small hit target, but the hover for 'close window' shows so I'm pretty sure I'm getting the right hit target.

**Steps to reproduce:**
1. Open the options window on Android
2. Attempt to close the window

**Screenshots / Video:**
![Screenshot_20210413-180601_OpenRCT2](https://user-images.githubusercontent.com/153219/114518969-40fe7980-9c83-11eb-816f-9ca062277823.png)

James103:
Have you tried attaching the output for `openrct2 --version` for the build/commit that you reproduced this on?

SilkyPantsDan:
I'm having the same issue using the latest available build - it doesn't appear to print the commit or build

![Screenshot_20210509-161750_OpenRCT2](https://user-images.githubusercontent.com/738172/117563046-6182ed80-b0e2-11eb-90d6-387c15021a15.jpg)

PR #14662, opened 2021-05-15
Fix Options window not closing on Android #14468

SilkyPantsDan:
Issue was caused by the WIDX_ALWAYS_NATIVE_LOADSAVE flag not being enabled/disabled correctly

Fixes issue #14468


Gymnasiast:
Good find! And ironically, I added that to _help_ Android...

SilkyPantsDan:
Can do - will push that update now

PR #14670, opened 2021-05-15
Improve corrupt element handling

Broxzier:
The way corrupt elements are replaced by the invisibility flag is not 100% accurate to how they behave in pre-NSF OpenRCT2.

Before NSF (middle tile of observation tower selected, there are 2 corrupt elements below the sloped path too):
Before

How it currently gets handled:
After Before

How it gets handled with this fix:
After After

Notice how the only difference now is the observation tower top. AFAIK this cannot be solved currently, until we add a new standard "nothing" element or separate the tower parts and tower tops into separate track pieces. The first is more like a hack, so we should probably avoid that, and is out of scope for this PR.


IntelOrca (approved):

Gymnasiast:
Please add a comment, as having two corrupt elements cancel each other out was a bug.

PR #14693, opened 2021-05-18
Enhancement: Improve Peep Falling and Drowning

hjk321:
This pull addresses odd behavior showcased in [this MarcelVos video](https://www.youtube.com/watch?v=0hHCv-KNLh4).

Previously, tired peeps would fall slower than energized peeps. This commit modifies the walkspeed logic to make gravity affect everyone the same while falling, regardless of their energy level. This has the side effect of normalizing the drowning animation speed as well since drowning is a subset of falling.

Since the drowning animation speed is now the same across the board, the amount of time a peep swims before drowning is now based on energy. (People with more energy can swim longer, after all!) This is achieved by dynamically setting the starting frame when first drowning. Maximum energy plays the animation in full, scaling all the way down to minimum energy levels, which starts on frame 48, giving the peep only about a second to flail helplessly before succumbing to fatigue.

The formula to determine the starting frame is `(MAX_ENERGY - Energy) / 2`.

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -92,6 +92,7 @@ The following people are not part of the development team, but have been contrib
 * Helio Batimarqui (batimarqui) - Misc.
 * Keith Stellyes (keithstellyes) - Misc.
 * Bas Cantrijn (Basssiiie) - Misc.
+* Hiram Anderson (hjk321) - Misc.
 
 ## Bug fixes
 * (halfbro)
--- a/src/openrct2/peep/Peep.cpp
+++ b/src/openrct2/peep/Peep.cpp
@@ -753,8 +753,10 @@ void Peep::UpdateFalling()
                             guest->InsertNewThought(PeepThoughtType::Drowning, PEEP_THOUGHT_ITEM_NONE);
                         }
 
+                        // Determine start frame for drowning based on peep energy.
+                        // The minimum peep energy starts the animation on frame 48 causing very quick death.
+                        ActionFrame = (PEEP_MAX_ENERGY - Energy) / 2;
                         Action = PeepActionType::Drowning;
-                        ActionFrame = 0;
                         ActionSpriteImageOffset = 0;
 
                         UpdateCurrentActionSpriteType();
@@ -935,6 +937,9 @@ void Peep::Update()
         if (State == PeepState::Queuing)
             stepsToTake += stepsToTake / 2;
     }
+    // Falling should be at maximum speed.
+    if (State == PeepState::Falling)
+        stepsToTake = PEEP_MAX_ENERGY;
 
     uint32_t carryCheck = StepProgress + stepsToTake;
     StepProgress = carryCheck;


duncanspumpkin:
Is that even something we want to change?

hjk321:
I think an case could be made for it. The fix doesn't really affect gameplay, but having your immersion broken by a tired person floating down through the air like a feather or taking triple the normal amount of time to drown may very well depending on the person. I guess my major argument is one of "why not."

Feel free to close it though, if you must.

ZehMatt:
I personally have no objections to this, it does seem a bit silly that their gravity is affected by their personal well being.

IntelOrca:
Perhaps using energy was just a way of randomising the time it takes for a peep to drown. It would make more sense if it was the other way round, tired guests should drown quicker.

hjk321:
Fixed whitespace formatting issues.

Broxzier:
Random gravity doesn't really make much sense so I'm in favour of this change. I haven't tried this yet, but I imagine that when removing a path from underneath the guests, they'll all hit the ground/water simultaneously now. Some random initial downwards velocity from the moment they start falling would look more realistic.

Like @IntelOrca mentions, 'randomness' for drowning should be the other way around as it currently is.

hjk321:
@Broxzier 
They do indeed hit the water simultaneously! I've tested this commit against the main branch and the difference is definitely there - they all fall at the same rate. I like the idea of changing the actual fall rate, but since I'd imagine that to be a much more controversial change, I'll leave that discussion for a later date.

@IntelOrca 
That is indeed how the current implementation is set up; tired peeps will drown significantly faster. I believe the total frames for the drowning animation is 68 or so; keep in mind that a **higher** starting frame actually results in **less** time in this animation before drowning, because even though the frames are being played at the same rate, a tired peep has less frames to go through and therefore drowns quicker. I've tested this already and a peep with 32 (min) energy drowns about three times quicker than a peep at 128 (max) energy

tupaschoal:
Would you mind recording a video or gif showing the before and after for a few cases?

duncanspumpkin:
I feel the drowning time should have some random element in it and not be purely based on the energy. I quite like the silly gravity based on energy but I'm not overly attached. This pr will break replays and will obviously need a network bump.

ZehMatt:
> I feel the drowning time should have some random element in it and not be purely based on the energy. I quite like the silly gravity based on energy but I'm not overly attached. This pr will break replays and will obviously need a network bump.

I don't think we have any replays that get peeps to fall, do we?

hjk321:
So, I was a little mistaken when I said everyone hits the water at the same time. Due to some unrelated behavior, guests start falling at different times when their platform is deleted. (If you know why this happens let me know!) This means they don't hit the water at the same time but they DO fall with the same speed which was the original goal anyway.

The falling example I recorded here is a pretty mild one since I couldn't modify each guest's energy one by one, but I'm sure most people have seen the symptoms. The drowning demonstration I recorded is pretty good though I think.

https://user-images.githubusercontent.com/37224753/118682896-79cfc800-b7c6-11eb-8982-9b58f451039d.mp4

duncanspumpkin:
> > I feel the drowning time should have some random element in it and not be purely based on the energy. I quite like the silly gravity based on energy but I'm not overly attached. This pr will break replays and will obviously need a network bump.
> 
> I don't think we have any replays that get peeps to fall, do we?

Falling isn't used only for falling. It's used for entering parks and leaving rides

IntelOrca:
It doesn't look very good, maybe we should complete changing falling to be fast and also accelerate with a maximum terminal velocity. When you compare them to e.g. crash particles, they don't fall quickly enough.

ocalhoun6:
> I like the idea of changing the actual fall rate

Each peep has a different weight, right?

Currently, that only affects things when the peep is on a ride, but it could work for falling, too. It would make sense for heavier peeps to fall slightly faster than lighter peeps.

hjk321:
I'm definitely open to adding more stuff to this PR if a maintainer approves any of the above ideas. I'm a little wary about adding actual physics stuff because sometimes simplicity is key in RCT2. If I were to toss my hat in the ring with an idea, I would just say double the fall rate. But we're starting to step out of enhancement territory and into feature suggestions that considerably affect the look and feel compared to the vanilla game.

@ocalhoun6 
> It would make sense for heavier peeps to fall slightly faster than lighter peeps.

Actually, it wouldn't. Remember [that one experiment](https://en.wikipedia.org/wiki/Galileo%27s_Leaning_Tower_of_Pisa_experiment) you studied in high school where they dropped the two cannonballs from the Tower of Pisa?

ZehMatt:
Where do we stand with this one?

rik-smeets:
@hjk32 Are you still up for continuing to work on this PR? I'll take another look at it and help with getting it merged :).

hjk321:
Hm, I forgot I made this PR! 

I'm not in a position to continue working on this right now, unfortunately. This silly physics tweak was made in less than a day and I haven't actually played RCT2 in over a year.

The changes don't look very hard to rebase but I'm not heartbroken if we just close this in favor of the existing behavior. I'll make sure that whole "maintainers can commit to fork" checkbox is on though if someone wants to pick this up.

rik-smeets:
I'll close this one for now then; it can always be opened if someone wants to continue working on this. Thanks for your response!

PR #14700, opened 2021-05-19
Raise water height limit

Broxzier:
Since the water height is not limited to the bits it was stored in anymore, it can be increased much higher now. This commits sets it to same max height as for surfaces.

The second commit could be moved to the develop branch instead if desired, since it isn't directly related to the limit itself.


IntelOrca (approved):

tupaschoal (commented):
Nice! I think this warrants a changelog entry, does it not?

EDIT: My bad, I didn't notice it was A PR for the NSF

Broxzier:
What's the best way to merge this? I'd go for a rebase, to keep the two commits separate and to keep the NSF branch linear, but that's disabled in this repository.

IntelOrca:
@Broxzier yeah, you can just cherry-pick your commits directly to NSF

Broxzier:
Done :+1:

PR #14702, opened 2021-05-19
Windows console changes

CookiePLMonster:
This PR changes the way Windows version of the game handles standard input/output and removes the need for a duplicate `OpenRCT2.com` file. Now only one executable is needed, with those remarks:
* Running the game normally will **not** spawn a console window.
* To show a console window, run the game with a newly added `--console` commandline argument. When the parent process has its own console (e.g. when running from `cmd.exe` or a batch file), OpenRCT2 will attach to it and use it for its own standard input/output. Otherwise, when the parent process doesn't have a console (e.g. when running from a shortcut file), a new console will be spawned.

image

I removed all mentions of `OpenRCT2.com` from readme files, build scripts and the installer. I did **not** add a "replacement" `OpenRCT2.bat` file, but if people are in favour of shipping such an one-liner helper batch file (running the game with a `--console` argument) I can add it.

### DIFF
--- a/src/openrct2-win/openrct2-win.cpp
+++ b/src/openrct2-win/openrct2-win.cpp
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2014-2020 OpenRCT2 developers
+ * Copyright (c) 2014-2021 OpenRCT2 developers
  *
  * For a complete list of all authors, please refer to contributors.md
  * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
@@ -20,37 +20,78 @@
 #include <iterator>
 #include <openrct2-ui/Ui.h>
 #include <openrct2/core/String.hpp>
+#include <shellapi.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string>
 #include <vector>
 
-static std::vector<std::string> GetCommandLineArgs(int argc, wchar_t** argvW);
+static std::vector<std::string> GetCommandLineArgs(const wchar_t* cmdLine);
+static void CreateConsole();
 
 /**
- * Windows entry point to OpenRCT2 with a console window using a traditional C main function.
+ * Windows entry point to OpenRCT2 for a GUI application, calling to the traditional C main function.
  */
-int wmain(int argc, wchar_t** argvW, [[maybe_unused]] wchar_t* envp)
+int WINAPI WinMain(HINSTANCE /*hInstance*/, HINSTANCE /*hPrevInstance*/, LPSTR /*lpCmdLine*/, int /*nShowCmd*/)
 {
-    auto argvStrings = GetCommandLineArgs(argc, argvW);
+    auto argvStrings = GetCommandLineArgs(GetCommandLineW());
 
-    SetConsoleCP(CODE_PAGE::CP_UTF8);
-    SetConsoleOutputCP(CODE_PAGE::CP_UTF8);
+    // Handle a special Windows-only commandline argument "--console". If present, create our own console and point stdin/stdout
+    // at that.
+#ifdef DEBUG
+    constexpr bool alwaysCreateConsole = true;
+#else
+    constexpr bool alwaysCreateConsole = false;
+#endif
+
+    auto consoleArg = std::find(argvStrings.begin(), argvStrings.end(), "--console");
+    if (consoleArg != argvStrings.end())
+    {
+        CreateConsole();
+
+        // Remove this argument so the game doesn't throw an unknown argument error.
+        argvStrings.erase(consoleArg);
+    }
+    else if constexpr (alwaysCreateConsole)
+    {
+        CreateConsole();
+    }
 
     std::vector<const char*> argv;
     std::transform(
         argvStrings.begin(), argvStrings.end(), std::back_inserter(argv), [](const auto& string) { return string.c_str(); });
-    auto exitCode = NormalisedMain(argc, argv.data());
+    argv.push_back(nullptr);
+    auto exitCode = main(static_cast<int>(argvStrings.size()), argv.data());
     return exitCode;
 }
 
-static std::vector<std::string> GetCommandLineArgs(int argc, wchar_t** argvW)
+static std::vector<std::string> GetCommandLineArgs(const wchar_t* cmdLine)
 {
     // Allocate UTF-8 strings
     std::vector<std::string> argv;
-    for (int i = 0; i < argc; i++)
+    int numArgs;
+    LPWSTR* args = CommandLineToArgvW(cmdLine, &numArgs);
+    if (args != nullptr)
     {
-        argv.push_back(String::ToUtf8(argvW[i]));
+        for (int i = 0; i < numArgs; i++)
+        {
+            argv.push_back(String::ToUtf8(args[i]));
+        }
     }
+    LocalFree(args);
     return argv;
+}
+
+static void CreateConsole()
+{
+    if (AllocConsole())
+    {
+        FILE* dummyFile;
+        freopen_s(&dummyFile, "CONIN$", "r", stdin);
+        freopen_s(&dummyFile, "CONOUT$", "w", stderr);
+        freopen_s(&dummyFile, "CONOUT$", "w", stdout);
+
+        SetConsoleCP(CODE_PAGE::CP_UTF8);
+        SetConsoleOutputCP(CODE_PAGE::CP_UTF8);
+    }
 }
--- a/distribution/scripting.md
+++ b/distribution/scripting.md
@@ -46,7 +46,7 @@ registerPlugin({
 });
 ```
 
-This will log a message to the terminal screen (`stdout`) when you open any park. If you are on Windows, make sure to run `openrct2.com` instead of `openrct2.exe` so you can interact with the `stdin` / `stdout` console. The console is a JavaScript interpreter (REPL), this means you can write and test expressions similar to the console found in web browsers when you press `F12`. When you make changes to your script, you must exit your current game and open it again for the script to reload... unless you use the hot reload feature.
+This will log a message to the terminal screen (`stdout`) when you open any park. If you are on Windows, make sure to run `openrct2.exe` with a `--console` argument so you can interact with the `stdin` / `stdout` console. The console is a JavaScript interpreter (REPL), this means you can write and test expressions similar to the console found in web browsers when you press `F12`. When you make changes to your script, you must exit your current game and open it again for the script to reload... unless you use the hot reload feature.
 
 The hot reload feature can be enabled by editing your `config.ini` file and setting `enable_hot_reloading` to `true` under `[plugin]`. When this is enabled, the game will auto-reload the script in real-time whenever you save your JavaScript file. This allows rapid development of plug-ins as you can write code and quickly preview your changes, such as closing and opening a specific custom window on startup. A demonstration of this can be found on YouTube: [OpenRCT2 plugin hot-reload demo](https://www.youtube.com/watch?v=jmjWzEhmDjk)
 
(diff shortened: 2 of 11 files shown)


Gymnasiast:
Yes, please add a .bat file. We want to make troubleshooting as easy as possible, and just double-clicking a .bat should be easy to do even for novice users.

CookiePLMonster:
> Yes, please add a .bat file.

How do I go about adding it to artifacts? Should it be created at build time or placed somewhere in the repository or pulled to artifacts, like readmes and other text files?

duncanspumpkin:
Think we will need to adjust the readme to add about setting this as a debug option when first setting up.

Gymnasiast:
@duncanspumpkin Could you clarify? Is this something that @CookiePLMonster needs to do?

IntelOrca:
I still need to review this, but I think debug builds should use console subsystem still.

duncanspumpkin:
Pretty much the same as Intel orca, in debug we would either want to keep it in console mode or add the switch when debugging (which would need adjusting the readme)

CookiePLMonster:
What for, though? You can easily debug with a commandline argument specified.

image

Although if needed I can make the game spawn the console unconditionally when compiled in Debug, without the need to use a console subsystem.

duncanspumpkin:
Yes so I'm saying that should be added to the readme as a specific step.

CookiePLMonster:
I guess spawning the console unconditionally in Debug may be more user/developer friendly then? One less step to take care about.

ZehMatt:
> I guess spawning the console unconditionally in Debug (or when a debugger is attached) may be more user/developer friendly then? One less step to take care about.

It should be probably unconditionally for debug builds. Also as suggested before this should be documented.

CookiePLMonster:
Updated with:
* Made the console show unconditionally in Debug builds.
* Fixed the commit message to satisfy the linter.
* Fixed an exit crash caused by freeing the console prematurely - it was sending an empty input to a now freed `StdInOutConsole`, and since the process is shutting down anyway there is no need to free the console early.
* Updated copyright years properly.
* Did my best to update the readme with a proper note of the new command line argument.

src/openrct2-win/openrct2-win.cpp
```diff
@@ -20,37 +20,86 @@
...
+        for (int i = 0; i < numArgs; i++)
+        {
+            argv.push_back(String::ToUtf8(args[i]));
+        }
     }
+    LocalFree(args);
     return argv;
 }
+
+static bool CreateConsole()
```
ZehMatt, line 89: Function results a value but is not used. What's the intent?

src/openrct2-win/openrct2-win.cpp
```diff
@@ -20,37 +20,86 @@
...
+static bool CreateConsole();
 
 /**
- * Windows entry point to OpenRCT2 with a console window using a traditional C main function.
+ * Windows entry point to OpenRCT2 for a GUI application, calling to the traditional C main function.
  */
-int wmain(int argc, wchar_t** argvW, [[maybe_unused]] wchar_t* envp)
+int WINAPI WinMain(
+    [[maybe_unused]] HINSTANCE hInstance, [[maybe_unused]] HINSTANCE hPrevInstance, [[maybe_unused]] LPSTR lpCmdLine,
+    [[maybe_unused]] int nShowCmd)
```
ZehMatt, line 37:
```suggestion
int WINAPI WinMain(HINSTANCE, HINSTANCE, LPSTR, int)
```
They are unconditionally unused so just removing the variable makes more sense.

ZehMatt (changes requested):

src/openrct2-win/openrct2-win.cpp
```diff
@@ -20,37 +20,86 @@
...
+        // Remove this argument so the game doesn't throw an unknown argument error.
+        argvStrings.erase(consoleArg);
+    }
+    else if constexpr (alwaysCreateConsole)
+    {
+        CreateConsole();
+    }
 
     std::vector<const char*> argv;
+    argv.reserve(argvStrings.size() + 1);
```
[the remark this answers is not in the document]
CookiePLMonster: It's not a big difference, but given the size is known up front, why not?

src/openrct2-win/openrct2-win.cpp
CookiePLMonster: It was used previously to free the console, but that caused an exit crash in `StdInOutConsole` so I removed it - the process is tearing down anyway. I personally don't see much harm in keeping this return value, but I can remove it if needed.

ZehMatt: Discarding the result of a function feels bad

src/openrct2-win/openrct2-win.cpp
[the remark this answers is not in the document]
ZehMatt: Extra line for no benefit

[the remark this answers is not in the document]
Gymnasiast: Does it make the startup quicker?

src/openrct2-win/openrct2-win.cpp
CookiePLMonster: I removed `[[maybe_unused]]` and kept the parameter names, but commented out. I feel this is the most readable, warning-free way to handle it.

ZehMatt: Fair enough

ZehMatt (approved):
I have no objections and code looks good. @IntelOrca are you okay with this?

IntelOrca:
@ZehMatt I haven't reviewed it yet, will get to it soon.

IntelOrca (changes requested):
There are are a few issues that mean this does not work or provide the experience we had before.

1. openrct2.com exists and uses the .com extension because that has higher precedence than .exe when not specifying the extension. E.g.

```> .\openrct2```

Will launch `openrct2.com` rather than `openrct2.exe`, which is what you want when already in a console.

.com files have to be binaries, they can not be batch scripts.

2. Launching `openrct2.exe --console`, will have a broken REPL in cmd or powershell. The prompt of the shell is still enabled before the game has completed... resulting in the screenshot below:

image

As you can see, powershell core just seems to outright crash afterwards. cmd is fine though.

Using a batch script wrapper does work around this, but results in an annoying "Terminate (Y/N)" message after exiting:

image

There was a specific reason I went with the double openrct2.exe openrct2.com route, and that was that it gave the best experience around the problems with the Windows console subsystem. We originally just had two very small executable wrappers (one .com, one .exe) which then called openrct2.dll (the main game). But as mentioned, this resulted in false positives from virus checkers.

CookiePLMonster:
> There was a specific reason I went with the double openrct2.exe openrct2.com route, and that was that it gave the best experience around the problems with the Windows console subsystem. We originally just had two very small executable wrappers (one .com, one .exe) which then called openrct2.dll (the main game). But as mentioned, this resulted in false positives from virus checkers.

If this is the reason, then `openrct2.com` could be a tiny console application that starts `openrct2.exe` with redirected IO.

IntelOrca:
Or the other way round. openrct2.exe is a Windows subsystem app that starts openrct2.com with hidden TTY.

CookiePLMonster:
Majority of the users want to just run the game and not worry about logging, so I would not optimize the startup flow for the debugging case.

CookiePLMonster:
> 1. Launching `openrct2.exe --console`, will have a broken REPL in cmd or powershell.

I just realized that since this is an issue (I agree it's annoying), perhaps attaching to the parent console is not a good idea after all - if a new console is spawned every time, this won't be an issue.

ZehMatt:
I dislike the .com thing a lot. I also dislike that .exe and .com statically link the code, making 24 MB in total. It would be better to have openrct2.dll and openrct2.exe and openrct2-cli.exe where both exes just link the dynamic library.

The com files are just exes with console subsystem, I always accidentally launch the .com rather than the .exe when typing "openrct2 join/host ..." which leads to being unable to debug the process with symbols since almost nothing supports this ancient binary anymore.

Backtrace suffers from this as well, all Backtrace records submitted from the .com are useless.

janisozaur:
> Backtrace suffers from this as well, all Backtrace records submitted from the .com are useless.

Really? That's odd, they're PE32+ I think I've even seen some proper reports using those. What's the problem with them?

ZehMatt:
I don't think it ever had symbols for reports that ran it from the .com binary. The same effect is noticable with VS, if you launch the com and attach VS it won't load symbols.

CookiePLMonster:
> I dislike the .com thing a lot. I also dislike that .exe and .com statically link the code, making 24 MB in total. It would be better to have openrct2.dll and openrct2.exe and openrct2-cli.exe where both exes just link the dynamic library.

If that's a requirement, then I think it may be cleaner to have `openrct2-cli.exe` launch `openrct2.exe` as a child process with redirected input and output. I **think** it would be equivalent, just without a DLL file.

ZehMatt:
> > I dislike the .com thing a lot. I also dislike that .exe and .com statically link the code, making 24 MB in total. It would be better to have openrct2.dll and openrct2.exe and openrct2-cli.exe where both exes just link the dynamic library.
> 
> If that's a requirement, then I think it may be cleaner to have `openrct2-cli.exe` launch `openrct2.exe` as a child process with redirected input and output. I **think** it would be equivalent, just without a DLL file.

I guess that works too.

CookiePLMonster:
> I guess that works too.

Alternately, what about this?
> I just realized that since this is an issue (I agree it's annoying), perhaps attaching to the parent console is not a good idea after all - if a new console is spawned every time, this won't be an issue.

Seems like the main issue @IntelOrca had with the current state of this PR is its weird behaviour when outputting to an already spawned cmd.exe instance - and always spawning a new instance will resolve that.

ZehMatt:
I was wrong about the .com files having bad/no symbols, that just seems to be an VS issue (on my end?). Backtrace however does miss them in a lot of reports.

CookiePLMonster:
What's the consensus on these changes? If it was up to me, I'd update this solution by making it spawn a **new** console no matter whether the game is launched from commandline or the usual way. This will also avoid the issue with cmd.exe mentioned in the command.

ZehMatt:
I'll leave it to @IntelOrca

IntelOrca:
If you are able to resolve all the issues I mentioned, then I will re-review it.

CookiePLMonster:
> If you are able to resolve all the issues I mentioned, then I will re-review it.

I pushed a rebased version with a change where a new debug console is always spawned. This resolves REPL semi-breaking when running OpenRCT2 from cmd.exe.

IntelOrca (changes requested):
Running `openrct2.exe --console --version` in an existing cmd window, does not terminate until RETURN is pressed.

Piping to a file does not work. E.g. `openrct2.exe --console --version > dog`. It does work if I remove `--console`, however if we are relying on a .bat file always passing `--console`, this will be confusing as the .bat version will not be able to pipe.

When running from an existing cmd window, the interactive console during the game is unusable as it keeps displaying the CMD prompt.

Running from PowerShell still ends up with exceptions when exiting like my screenshots from earlier.

On the whole, I am afraid that these issues just don't make the workarounds worthwhile. I think the only way you are going to ever get a good console experience, is by having a binary with the subsystem set to CONSOLE.

CookiePLMonster:
Yeah, with those issues in sight I can see that. I'll look into the other idea then, with `openrct2.com` being a lightweight app that launches `openrct2.exe` with stdout redirected. If that ends up not being viable, I think this idea might have to go after all.

CookiePLMonster:
@IntelOrca Please retest when you get the chance, I just noticed that I made a mistake when rebasing and I did **not** actually push my latest changes to the console attach behaviour 🤦‍♂️

IntelOrca:
@CookiePLMonster I just tested the latest version. It always creates a new console window rather than attaching to the existing one you are running from. Piping with `--console` does not work either.

CookiePLMonster:
In hindsight and having more knowledge since, this was not a great idea. Having the game executable duplicated as .exe and .com is still not a great solution, but this isn't the way to fix it.

If I come up with a better solution for this, I'll open a new PR.

IntelOrca:
@CookiePLMonster no problem, thank you for attempting and testing alternative options.

PR #14712, opened 2021-05-21
Startup speed optimizations

CookiePLMonster:
This PR consists of 3 changes to the file IO, resulting in a significant speedup to the startup times:

1. `ObjectAsset::GetSize()` obtained the size of the file by reading it all to memory and counting bytes read, then discarding the contents. This was especially wasteful when used on music tracks and took over 400ms during startup on my setup. To remedy this, I introduced a `File::GetSize()` function that queries the size without reading the file contents. Also made `ObjectAsset::GetSize()` return a 64-bit value, because there is no reason why should the file size be truncated in 32-bit builds.
2. Because of these changes, a "seek to end, tell, seek to begin" idiom can be replaced with a proper size check.
3. After the above two changes, `ftell` showed up as the second hottest function when profiling startup, The culprit turned out to be `FileStream::Read` repeatedly checking for EOF before attempting to read. Since this is an exceptional case that should never/rarely happen, I now let the game try to read from the file and throw an exception if it doesn't read enough - hitting EOF while reading is one of the cases where this will happen. This effectively "optimizes" the function for a success case, while previously it was closer to being "optimized" for a failure case.

I also took the liberty of adding this change to a changelog, since halving the startup time is something easily noticeable for the end user. On my machine, startup times **went down from ~830ms to ~380ms**:

Before: `TitleScreen::Load() took 835ms.`
After: `TitleScreen::Load() took 376ms.`

### DIFF
--- a/src/openrct2/core/FileStream.cpp
+++ b/src/openrct2/core/FileStream.cpp
@@ -165,13 +165,9 @@ namespace OpenRCT2
 
     void FileStream::Read(void* buffer, uint64_t length)
     {
-        uint64_t remainingBytes = GetLength() - GetPosition();
-        if (length <= remainingBytes)
+        if (fread(buffer, 1, static_cast<size_t>(length), _file) == length)
         {
-            if (fread(buffer, static_cast<size_t>(length), 1, _file) == 1)
-            {
-                return;
-            }
+            return;
         }
         throw IOException("Attempted to read past end of file.");
     }
--- a/src/openrct2/core/FileStream.cpp
+++ b/src/openrct2/core/FileStream.cpp
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2014-2020 OpenRCT2 developers
+ * Copyright (c) 2014-2021 OpenRCT2 developers
  *
  * For a complete list of all authors, please refer to contributors.md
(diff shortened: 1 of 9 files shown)


ZehMatt (approved):
Nice, feels almost instant on my end without the console

src/openrct2/core/FileStream.cpp
```diff
@@ -98,9 +100,12 @@ namespace OpenRCT2
...
 
-        Seek(0, STREAM_SEEK_END);
-        _fileSize = GetPosition();
-        Seek(0, STREAM_SEEK_BEGIN);
+#ifdef _WIN32
+        _fileSize = _filelengthi64(_fileno(_file));
+#else
+        std::error_code ec;
+        _fileSize = fs::file_size(fs::u8path(path), ec);
+#endif
```
tupaschoal, line 108: Why not use `Platform::GetSize` here? Maybe I missed something.

In any case, I'm (slowly) trying to keep all OS ifdefs on the platform files, for an easier to read code overall, so I'd be very interested in making this a bit friendlier.

CookiePLMonster: I could use that, but in this case on Windows I could get the size directly from a `FILE*` instead of a path.

PR #14716, opened 2021-05-22
More startup speed optimizations

CookiePLMonster:
A follow up to #14712, optimizing startup times further. When investigating `SawyerChunkReader`, I spotted that every chunk buffer followed an `allocate 16MB -> shrink down to real size` pattern when reading. However, **all** current instances of `SawyerChunk` are just short lived local variables that get freed soon after reading, so reallocation is a pointless memory optimization.

In this PR I made `SawyerChunkReader` not shrink allocations by default and instead introduced a (currently never used) `persistentChunks` parameter that acts as a hint to the reader that the `SawyerChunk` will be persisted and thus it makes sense to shrink that memory.

On my machine, those changes shorten the startup time by 80-100ms, and shorten the object index building by around 200ms.
image

**NOTE:** I'm not happy how `SawyerChunk` has to call to `SawyerChunkReader` to free memory without an allocator mismatch. In my opinion, this would be best refactored into a small, static-only `SawyerChunkAllocator` class that would essentially only house what is currently `AllocateLargeTempBuffer`, `FinaliseLargeTempBuffer` and `FreeLargeTempBuffer`.
EDIT: Alternately, those three functions could just use `Memory::Allocate`, `Memory::Reallocate` and `Memory::Free` internally - this would remove the need to modify `SawyerChunk` completely, but would remove the `__USE_HEAP_ALLOC__` Debug optimization.

### DIFF
--- a/src/openrct2/rct12/SawyerChunkReader.h
+++ b/src/openrct2/rct12/SawyerChunkReader.h
@@ -36,15 +36,17 @@ namespace OpenRCT2
 
 /**
  * Reads sawyer encoding chunks from a data stream. This can be used to read
- * SC6, SV6 and RCT2 objects.
+ * SC6, SV6 and RCT2 objects. persistentChunks is a hint to the reader that the chunk will be preserved,
+ * and thus the chunk memory should be shrunk.
  */
 class SawyerChunkReader final
 {
 private:
     OpenRCT2::IStream* const _stream = nullptr;
+    const bool _createsPersistentChunks = false;
 
 public:
-    explicit SawyerChunkReader(OpenRCT2::IStream* stream);
+    explicit SawyerChunkReader(OpenRCT2::IStream* stream, bool persistentChunks = false);
 
     /**
      * Skips the next chunk in the stream without decoding or reading its data
--- a/src/openrct2/rct12/SawyerChunkReader.h
+++ b/src/openrct2/rct12/SawyerChunkReader.h
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * Copyright (c) 2014-2020 OpenRCT2 developers
+ * Copyright (c) 2014-2021 OpenRCT2 developers
  *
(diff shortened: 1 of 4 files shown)


Gymnasiast (approved):
Nice! Seems to knock off a few seconds from my startup time (from ~14 to ~11 seconds, including building indices, using a debug build on Ubuntu 20.04)

CookiePLMonster:
Rebased to fix the merge conflict in the changelog.

src/openrct2/rct12/SawyerChunkReader.h
```diff
@@ -36,15 +36,17 @@ namespace OpenRCT2
...
  */
 class SawyerChunkReader final
 {
 private:
     OpenRCT2::IStream* const _stream = nullptr;
+    const bool _createsPersistentChunks = false;
 
 public:
-    explicit SawyerChunkReader(OpenRCT2::IStream* stream);
+    explicit SawyerChunkReader(OpenRCT2::IStream* stream, bool persistentChunks = false);
```
Broxzier, line 49: This argument is never set to true. This essentially remove sthe persistent chunks entirely. Doesn't this make loading of parks slower every time now?

CookiePLMonster: No, the opposite - unless I overlooked something, the codebase has no instances where a loaded `SawyerChunk` is persisted, they are always kept as short lived local variables. Therefore, with `persistentChunks` set to `false`, the code won't bother shrinking allocations as they are freed microseconds later anyway.

ZehMatt: If this is not used then I'm for removing it

CookiePLMonster: Both are fine for me - if you think it's better this way, I can remove those reallocate functions completely.

IntelOrca (approved):

PR #14726, opened 2021-05-23
Fix #14724: Money was ignored for ride creation

### LINKED ISSUE #14724: construction works despite being in "red numbers"
**OS:** MacOS 10.15.7
**Commit/Build:** 9a76224

I got into negative balance (hired too many staff), but i was still able to build new rides.
This obviously shouldn't be the case.
Active Plugins:
Ride Price manager
Soft guest gap calculator
(i did a second test with the plugins removed, the bug still occured)

- [ ] Reproducible in RCT2 (vanilla)? -> can't test, but probably not
- [ ] Specific to multiplayer? -> no

**Steps to reproduce:**
1. get your balance to be negative
2. build a ride/shop

**Video:**

https://youtu.be/uPD8gGnVATg

**Save game:**

[Evergreen Gardens.sv6.zip](https://github.com/OpenRCT2/OpenRCT2/files/6528437/Evergreen.Gardens.sv6.zip)

ZehMatt:
This is the right way to do it

### DIFF
--- a/src/openrct2/management/Finance.cpp
+++ b/src/openrct2/management/Finance.cpp
@@ -75,7 +75,7 @@ bool finance_check_money_required(uint32_t flags)
  */
 bool finance_check_affordability(money32 cost, uint32_t flags)
 {
-    return cost <= 0 || !finance_check_money_required(flags) || cost <= gCash;
+    return !finance_check_money_required(flags) || cost <= 0 || cost <= gCash;
 }
 
 /**
--- a/src/openrct2/actions/GameAction.cpp
+++ b/src/openrct2/actions/GameAction.cpp
@@ -310,8 +310,7 @@ namespace GameActions
 
         if (result->Error == GameActions::Status::Ok)
         {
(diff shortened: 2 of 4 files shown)


src/openrct2/management/Finance.cpp
```diff
@@ -75,7 +75,7 @@ bool finance_check_money_required(uint32_t flags)
  */
 bool finance_check_affordability(money32 cost, uint32_t flags)
 {
-    return cost <= 0 || !finance_check_money_required(flags) || cost <= gCash;
+    return !finance_check_money_required(flags) || cost <= 0 || cost <= gCash;
```
ZehMatt, line 78: I moved this condition since cost can be below zero but when its a ghost it shouldn't trigger a money error.

James103:
Does this also fix you being able to terraform while your bank balance is negative? That's how I reproduced #14724.

ZehMatt:
> Does this also fix you being able to terraform while your bank balance is negative? That's how I reproduced #14724.

It should, it reverts the bad condition from my previous commit.

Issue #613, opened 2014-12-02
Re-intruduce seeing through water


AaronVanGeffen:
Something I'm really missing compared to the original RollerCoaster series, is the ability to see through water. Especially in scenarios that consist mostly of water (e.g. Barony Bridge), it's just very hard to see the border of the park without such a feature.

I have no idea how much work it'd entail to re-introduce this, but I'm hoping it can be.

Superjustinbros:
I'd say keep the water the way it is; though perhaps add in an option for more transparent water.

AaronVanGeffen:
The many problems related to water (e.g. invisible scenario borders, staff patrol areas) have been discussed more in depth in #682. Closing this one.

PR #14731, opened 2021-05-24
Add option to make water opaque

Gymnasiast:


src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -3626,38 +3629,47 @@ static void top_toolbar_init_view_menu(rct_window* w, rct_widget* widget)
         ToggleOption(DDIDX_VIEW_CLIPPING, STR_VIEW_CLIPPING_MENU),
         ToggleOption(DDIDX_HIGHLIGHT_PATH_ISSUES, STR_HIGHLIGHT_PATH_ISSUES_MENU),
     };
+
     static_assert(ItemIDsMatchIndices(items));
 
     SetItems(items);
 
+    // Opaque water relies on RCT1 sprites.
```
Gymnasiast, line 3637: This block needs to be moved down because it won't work otherwise.

AaronVanGeffen:
Nice work! This takes me back to #613.

I get why you're calling the option 'Opaque water'. However, from a player perspective, I think it would be nicer to instead offer the option 'Transparent water' or 'Underwater view', either being checked by default. This would also be more in line with the option offered by the original RCT games, particularly if it is called 'Underwater view'.

Gymnasiast:
OK, did that. Also added a shortcut for it (defaulting to 2, like in RCT1).

distribution/changelog.txt
```diff
@@ -6,6 +6,7 @@
 - Feature: [#14538] [Plugin] Add property for getting current plugin api version.
 - Feature: [#14620] [Plugin] Add properties related to guest generation.
 - Feature: [#14636] [Plugin] Add properties related to climate and weather.
+- Feature: [#14731] Opaque water (like in RCT1).
```
tupaschoal, line 9: Should this have reflected the "Transparent water" string?

Gymnasiast: Nah, don’t think so. Transparent water is what we already had, opaque is what we added.

PR #14762, opened 2021-05-27
Close #14511: Unlock all music with 'Unlock operating limits' cheat

### LINKED ISSUE #14511: Allow selecting all music when "Unlock operating limits" is on
Currently, the merry-go-round can only select the fairground organ style, while other rides can select everything but the fairground organ style. We should extend the "Unlock operating limits" cheat to allow selecting all music.

Gymnasiast:


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -5031,7 +5031,7 @@ static void window_ride_music_mousedown(rct_window* w, rct_widgetindex widgetInd
                 }
             }
 
-            if (musicObj->SupportsRideType(ride->type))
+            if (gCheatsFastLiftHill || musicObj->SupportsRideType(ride->type))
```
duncanspumpkin, line 5034: This should be renamed the name no longer reflects its use.

PR #14764, opened 2021-05-28
Split up Sprite header

duncanspumpkin:
Hopefully this should help reduce the amount of full recompiles required when making small changes to headers.


tupaschoal (approved):
Took a glance and it seems alright.

PR #14766, opened 2021-05-28
Turn vehicle sprite flags into a 32-bit integer

Gymnasiast:
At the request of X7123M3-256.


duncanspumpkin:
RideObject.cpp needs modified for the json object reader

duncanspumpkin (changes requested):
RideObject.cpp L792 requires modifiying

duncanspumpkin (approved):

PR #14782, opened 2021-05-29
Refactor explicit passing of Ride type in OnEnterRide/OnExitRide

ZehMatt:
Those functions should be made private eventually. Also all the null checks only exist because the Peep can have a potentially invalid 'CurrentRide' since its nowhere reset with two exceptions, but changing that will break the replays.

I will do more smaller PRs, the goal is to pass Ride by reference and properly handle graceful demolishing/deletions so peeps will never run into a nullptr making all the checks obsolete.

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1695,6 +1695,7 @@ void Guest::OnEnterRide(Ride* ride)
 {
     if (ride == nullptr)
         return;
+
     // Calculate how satisfying the ride is for the peep. Can range from -140 to +105.
     int16_t satisfaction = peep_calculate_ride_satisfaction(this, ride);
 
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -1691,12 +1691,10 @@ bool Guest::DecideAndBuyItem(Ride* ride, ShopItem shopItem, money32 price)
  * ride's satisfaction value.
  *  rct2: 0x0069545B
  */
-void Guest::OnEnterRide(ride_id_t rideIndex)
+void Guest::OnEnterRide(Ride* ride)
 {
-    auto ride = get_ride(rideIndex);
     if (ride == nullptr)
         return;
-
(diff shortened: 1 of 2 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -1691,12 +1691,10 @@ bool Guest::DecideAndBuyItem(Ride* ride, ShopItem shopItem, money32 price)
  * ride's satisfaction value.
  *  rct2: 0x0069545B
  */
-void Guest::OnEnterRide(ride_id_t rideIndex)
+void Guest::OnEnterRide(Ride* ride)
 {
-    auto ride = get_ride(rideIndex);
     if (ride == nullptr)
         return;
-
```
Gymnasiast, line 1699: Little nitpick, but could you restore this line for legibility?

src/openrct2/peep/Guest.cpp
```diff
@@ -3492,7 +3489,7 @@ static void peep_update_ride_leave_entrance_maze(Guest* peep, Ride* ride, Coords
     peep->SetDestination(entrance_loc, 3);
 
     ride->cur_num_customers++;
-    peep->OnEnterRide(peep->CurrentRide);
+    peep->OnEnterRide(ride);
```
Gymnasiast, line 3492: Are we sure that this is the same ride in 100% of cases?

src/openrct2/peep/Guest.cpp
```diff
@@ -1691,12 +1691,10 @@ bool Guest::DecideAndBuyItem(Ride* ride, ShopItem shopItem, money32 price)
  * ride's satisfaction value.
  *  rct2: 0x0069545B
  */
-void Guest::OnEnterRide(ride_id_t rideIndex)
+void Guest::OnEnterRide(Ride* ride)
 {
-    auto ride = get_ride(rideIndex);
     if (ride == nullptr)
         return;
```
Gymnasiast, line 1697:
```suggestion
        return;

```

Gymnasiast (approved):

PR #14788, opened 2021-05-30
Banner Refactor from NSF

duncanspumpkin:
Refactor how banners work as per cherry-pick from NSF.

Previously the game actions sent out what banner id they were going to use. This changes it so that the banner id is always just the next available one. This means that the banner id must be kept in sync during network games. This change is only possible due to changes made to ghost placement.

### DIFF
--- a/src/openrct2/actions/BannerPlaceAction.cpp
+++ b/src/openrct2/actions/BannerPlaceAction.cpp
@@ -18,10 +18,29 @@
 
 using namespace OpenRCT2;
 
-BannerPlaceAction::BannerPlaceAction(const CoordsXYZD& loc, uint8_t bannerType, BannerIndex bannerIndex, uint8_t primaryColour)
+BannerPlaceActionResult::BannerPlaceActionResult()
+    : GameActions::Result(GameActions::Status::Ok, STR_CANT_POSITION_THIS_HERE)
+{
+}
+
+BannerPlaceActionResult::BannerPlaceActionResult(GameActions::Status err)
+    : GameActions::Result(err, STR_CANT_POSITION_THIS_HERE)
+{
+}
+
+BannerPlaceActionResult::BannerPlaceActionResult(GameActions::Status err, rct_string_id msg)
+    : GameActions::Result(err, STR_CANT_POSITION_THIS_HERE, msg)
+{
+}
+
+BannerPlaceActionResult::BannerPlaceActionResult(GameActions::Status err, rct_string_id title, rct_string_id message)
+    : GameActions::Result(err, title, message)
+{
+}
+
+BannerPlaceAction::BannerPlaceAction(const CoordsXYZD& loc, uint8_t bannerType, uint8_t primaryColour)
     : _loc(loc)
     , _bannerType(bannerType)
-    , _bannerIndex(bannerIndex)
     , _primaryColour(primaryColour)
 {
 }
@@ -31,7 +50,6 @@ void BannerPlaceAction::AcceptParameters(GameActionParameterVisitor& visitor)
     visitor.Visit(_loc);
     visitor.Visit("object", _bannerType);
     visitor.Visit("primaryColour", _primaryColour);
-    _bannerIndex = create_new_banner(0);
 }
 
 uint16_t BannerPlaceAction::GetActionFlags() const
@@ -43,7 +61,7 @@ void BannerPlaceAction::Serialise(DataSerialiser& stream)
 {
     GameAction::Serialise(stream);
 
-    stream << DS_TAG(_loc) << DS_TAG(_bannerType) << DS_TAG(_bannerIndex) << DS_TAG(_primaryColour);
+    stream << DS_TAG(_loc) << DS_TAG(_bannerType) << DS_TAG(_primaryColour);
 }
 
 GameActions::Result::Ptr BannerPlaceAction::Query() const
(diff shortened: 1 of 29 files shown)


src/openrct2/actions/BannerPlaceAction.cpp
```diff
@@ -86,17 +104,10 @@ GameActions::Result::Ptr BannerPlaceAction::Query() const
...
-        return MakeResult(GameActions::Status::InvalidParameters, STR_CANT_POSITION_THIS_HERE);
-    }
-
-    auto banner = GetBanner(_bannerIndex);
-    if (!banner->IsNull())
-    {
-        log_error("Banner index in use, bannerIndex = %u", _bannerIndex);
-        return MakeResult(GameActions::Status::InvalidParameters, STR_CANT_POSITION_THIS_HERE);
+        log_error("No free banners available");
+        return MakeResult(GameActions::Status::InvalidParameters, STR_TOO_MANY_BANNERS_IN_GAME);
```
Gymnasiast, line 110: I think this will result in a half-message. (I will check it.)

```suggestion
        return MakeResult(GameActions::Status::InvalidParameters, STR_CANT_POSITION_THIS_HERE, STR_TOO_MANY_BANNERS_IN_GAME);
```

src/openrct2/actions/BannerPlaceAction.cpp
```diff
@@ -124,24 +135,18 @@ GameActions::Result::Ptr BannerPlaceAction::Execute() const
...
 
-    auto banner = GetBanner(_bannerIndex);
-    if (!banner->IsNull())
+    auto banner = CreateBanner();
+    if (banner == nullptr)
     {
-        log_error("Banner index in use, bannerIndex = %u", _bannerIndex);
-        return MakeResult(GameActions::Status::InvalidParameters, STR_CANT_POSITION_THIS_HERE);
+        log_error("No free banners available");
+        return MakeResult(GameActions::Status::InvalidParameters, STR_TOO_MANY_BANNERS_IN_GAME);
```
Gymnasiast, line 149: See previous.

src/openrct2/actions/BannerPlaceAction.h
```diff
@@ -11,17 +11,27 @@
...
 private:
     CoordsXYZD _loc;
     ObjectEntryIndex _bannerType{ BANNER_NULL };
-    BannerIndex _bannerIndex{ BANNER_INDEX_NULL };
     uint8_t _primaryColour{};
 
 public:
     BannerPlaceAction() = default;
-    BannerPlaceAction(const CoordsXYZD& loc, uint8_t bannerType, BannerIndex bannerIndex, uint8_t primaryColour);
+    BannerPlaceAction(const CoordsXYZD& loc, uint8_t bannerType, uint8_t primaryColour);
```
Gymnasiast, line 34:
```suggestion
    BannerPlaceAction(const CoordsXYZD& loc, ObjectEntryIndex bannerType, colour_t primaryColour);
```

**Always** use ObjectEntryIndex for object indices. Especially now they can be higher than 254.

src/openrct2/actions/BannerSetNameAction.cpp
```diff
@@ -55,6 +56,12 @@ GameActions::Result::Ptr BannerSetNameAction::Query() const
 GameActions::Result::Ptr BannerSetNameAction::Execute() const
 {
     auto banner = GetBanner(_bannerIndex);
+    if (banner == nullptr)
+    {
+        log_warning("Invalid banner id, banner id = %d", _bannerIndex);
+        return MakeResult(GameActions::Status::InvalidParameters, STR_NONE);
```
Gymnasiast, line 62: I know this is just copied, but we should have a proper error message and not a blank one.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1518,7 +1518,14 @@ void S6Exporter::ExportBanners()
...
         auto src = GetBanner(i);
         auto dst = &_s6.banners[i];
-        ExportBanner(*dst, *src);
+        if (src != nullptr)
+        {
+            ExportBanner(*dst, *src);
+        }
+        else
+        {
+            ExportBanner(*dst, {});
```
Gymnasiast, line 1527: This is not correct. If there is no source banner, it should create a banner that RCT2 recognises as a "null banner". Right now, it will just create a banner with everything set to 0, which is not the same.

duncanspumpkin: The default constructor for the banner sets it to banner null not 0 so this should be setting it correctly

src/openrct2/actions/BannerPlaceAction.cpp
duncanspumpkin: Uses the `BannerPlaceActionResult` constructor so will insert the STR_CANT_POSITION_THIS_HERE.

src/openrct2/actions/BannerPlaceAction.h
duncanspumpkin: I would prefer that was addressed as a follow on PR. Keep this one purely for the banner indexs.

Gymnasiast: Hm, are you going to create one right after this one, then?

src/openrct2/actions/BannerSetNameAction.cpp
duncanspumpkin: Again will be addressed in follow on PRs. This PR is purely just pulling the commits from the NSF.

Gymnasiast (changes requested):
I tested the following:
- Spam banners until you hit the limit
- Remove one
- Try building a new one.

This _should_ allow you to build one extra banner, but it doesn’t.

src/openrct2/world/Banner.cpp
```diff
@@ -379,11 +360,111 @@ void BannerElement::ResetAllowedEdges()
...
+    }
+    return banner;
+}
+
+void DeleteBanner(BannerIndex id)
+{
+    auto banner = GetBanner(id);
+    if (banner != nullptr)
+    {
+        banner = {};
```
duncanspumpkin, line 435: This does not destroy a banner it just resets the pointer to nullptr.

duncanspumpkin:
Okay thats fixed the banner allocation issue that meant you couldn't place after using up all the slots.

Gymnasiast (changes requested):
When placing a banner, the window has a blank viewport. None of the buttons in it work either.

duncanspumpkin:
ah yes it would help if i saved the banner id. Doh! too tired.

duncanspumpkin:
And that has fixed the banner/sign windows

src/openrct2/world/Banner.cpp
```diff
@@ -379,11 +360,111 @@ void BannerElement::ResetAllowedEdges()
...
+{
+    auto bannerIndex = BannerGetNewIndex();
+    auto banner = GetOrCreateBanner(bannerIndex);
+    if (banner != nullptr)
+    {
+        banner->id = bannerIndex;
+        banner->flags = 0;
+        banner->type = 0;
+        banner->text = {};
+        banner->colour = 2;
```
Gymnasiast, line 424: Needs constants.

May be picked in a later PR (but please create an issue if you decide to do so).

Gymnasiast (approved):

PR #14791, opened 2021-05-30
Map refactor from NSF

duncanspumpkin:
Cherry picked out of the nsf to reduce the size of the PR.

### DIFF
--- a/src/openrct2/actions/PlacePeepSpawnAction.cpp
+++ b/src/openrct2/actions/PlacePeepSpawnAction.cpp
@@ -47,12 +47,6 @@ GameActions::Result::Ptr PlacePeepSpawnAction::Query() const
     res->Expenditure = ExpenditureType::LandPurchase;
     res->Position = _location;
 
-    if (!map_check_free_elements_and_reorganise(3))
-    {
-        return std::make_unique<GameActions::Result>(
-            GameActions::Status::NoFreeElements, STR_ERR_CANT_PLACE_PEEP_SPAWN_HERE, STR_NONE);
-    }
-
     if (!LocationValid(_location) || _location.x <= 16 || _location.y <= 16 || _location.x >= (gMapSizeUnits - 16)
         || _location.y >= (gMapSizeUnits - 16))
     {
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -194,8 +194,7 @@ void S6Exporter::Export()
     _s6.scenario_srand_0 = state.s0;
     _s6.scenario_srand_1 = state.s1;
 
-    // Map elements must be reorganised prior to saving otherwise save may be invalid
-    map_reorganise_elements();
+    ReorganiseTileElements();
     ExportTileElements();
     ExportEntities();
     ExportParkName();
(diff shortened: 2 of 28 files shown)


src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -194,8 +194,7 @@ void S6Exporter::Export()
     _s6.scenario_srand_0 = state.s0;
     _s6.scenario_srand_1 = state.s1;
 
-    // Map elements must be reorganised prior to saving otherwise save may be invalid
```
Gymnasiast, line 197: Could you restore this comment?

src/openrct2/world/Map.cpp
```diff
@@ -1500,12 +1522,19 @@ void map_update_tiles()
...
 
-        auto mapPos = TileCoordsXY{ x, y }.ToCoordsXY();
-        auto* surfaceElement = map_get_surface_element_at(mapPos);
-        if (surfaceElement != nullptr)
+        // Repeat for each 256x256 block on the map
+        for (int32_t blockY = 0; blockY < gMapSize; blockY += 256)
         {
-            surfaceElement->UpdateGrassLength(mapPos);
-            scenery_update_tile(mapPos);
+            for (int32_t blockX = 0; blockX < gMapSize; blockX += 256)
```
Gymnasiast, line 1528: This looks wrong. BlockX seems to contain big coords - you check against gMapSize and add 256 on every iteration -, yet the value is used to create a TIleCoordsXY.

duncanspumpkin: it isn't used until the nsf increases the map size. In normal map size it only ever does the one iteration.

src/openrct2/world/Map.cpp
```diff
@@ -1111,21 +1139,16 @@ bool map_check_free_elements_and_reorganise(int32_t numElements)
...
-    {
-        log_error("Cannot insert new element");
-        return nullptr;
-    }
 
-    newTileElement = gNextFreeTileElement;
-    originalTileElement = gTileElementTilePointers[tileLoc.y * MAXIMUM_MAP_SIZE_TECHNICAL + tileLoc.x];
+    auto numElementsOnTileOld = CountElementsOnTile(loc);
+    auto numElementsOnTileNew = numElementsOnTileOld + 1;
+    auto* newTileElement = AllocateTileElements(numElementsOnTileNew);
```
duncanspumpkin, line 1145: Allocate can fail. Must return nullptr if fail.

src/openrct2/actions/PlacePeepSpawnAction.cpp
```diff
@@ -47,12 +47,6 @@ GameActions::Result::Ptr PlacePeepSpawnAction::Query() const
     res->Expenditure = ExpenditureType::LandPurchase;
     res->Position = _location;
 
-    if (!map_check_free_elements_and_reorganise(3))
```
duncanspumpkin, line 50: peep spawns dont use tile elements so this check was pointless. Likely copied from the park entrance code.

Gymnasiast:
You merged this without any approval. That’s bad enough, but you also caused this problem: https://github.com/OpenRCT2/OpenRCT2/issues/14848

src/openrct2/world/TileInspector.cpp
```diff
@@ -94,7 +94,7 @@ namespace OpenRCT2::TileInspector
     GameActionResultPtr InsertCorruptElementAt(const CoordsXY& loc, int16_t elementIndex, bool isExecuting)
     {
         // Make sure there is enough space for the new element
-        if (!map_check_free_elements_and_reorganise(1))
+        if (!!MapCheckCapacityAndReorganise(loc))
```
Basssiiie, line 97: Double exclamation mark is probably not what you intended?

src/openrct2/world/TileInspector.cpp
```diff
@@ -359,7 +359,7 @@ namespace OpenRCT2::TileInspector
     GameActionResultPtr PasteElementAt(const CoordsXY& loc, TileElement element, bool isExecuting)
     {
         // Make sure there is enough space for the new element
-        if (!map_check_free_elements_and_reorganise(1))
+        if (!!MapCheckCapacityAndReorganise(loc))
```
Basssiiie, line 362: Same here.

Issue #14804, opened 2021-06-02
isBanner scenery wall flag should be renamed to isDoubleSided


ZeeMajora:
This flag makes a scenery wall double sided, It does not make it a banner by itself.
So it should be renamed as it's current name is confusing and inaccurate.
Example of a double sided wall that uses isBanner: https://github.com/OpenRCT2/objects/blob/master/objects/rct2/scenery_wall/rct2.scenery_wall.walltxgt.json

duncanspumpkin:
Suspect it will need to be split into two flags most likely.

KirilAngelov:
Hello, can I try this one? It will be my first issue and some information where to change this will be great.

duncanspumpkin:
@KirilAngelov this will need a bit of investigation to explore all users of the flag in the code to determine if it is being used for two separate features before being renamed.

duncanspumpkin:
I've confirmed that it is only being used incorrectly. Rename it in src/openrct2/world/scenery.h WALL_SCENERY_IS_BANNER.
Its mistakenly used in tileinspector.cpp just remove all of:
```
                const auto wallEntry = tileElement->AsWall()->GetEntry();
                if (wallEntry != nullptr && wallEntry->wall.flags & WALL_SCENERY_IS_BANNER)
                {
```
You will also need to adjust WallObject.cpp to add isDoubleSided as an additional flag. Will look a bit of a mess as you need to keep isBanner for backwards compatibility for now.

KirilAngelov:
I can't seem to find the code snippet in TileInspector.cpp even using search.

duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/blob/fd4a798cfb508c667d1b52c0c5852b80f17c890d/src/openrct2-ui/windows/TileInspector.cpp#L2099

PR #14807, opened 2021-06-02
Renamed banner for fix issue number #14804.

KirilAngelov:


src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -2094,26 +2094,6 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
                 int16_t wallType = tileElement->AsWall()->GetEntryIndex();
                 DrawTextBasic(dpi, screenCoords, STR_TILE_INSPECTOR_WALL_TYPE, &wallType, { COLOUR_WHITE });
 
-                // Banner info
-                const auto wallEntry = tileElement->AsWall()->GetEntry();
-                if (wallEntry != nullptr && wallEntry->wall.flags & WALL_SCENERY_IS_BANNER)
```
duncanspumpkin, line 2099: I meant remove this specific if check. Walls can have banners but the correct way to check if they exist or not is using `auto banner = tileElement->AsWall()->GetBanner();` and checking if the result is non null.

duncanspumpkin:
There will be a few other locations in the codebase that you will need to replace with the new name.

src/openrct2/world/Scenery.h
```diff
@@ -116,7 +116,7 @@ enum WALL_SCENERY_FLAGS
     WALL_SCENERY_HAS_PRIMARY_COLOUR = (1 << 0),   // 0x1
     WALL_SCENERY_HAS_GLASS = (1 << 1),            // 0x2
     WALL_SCENERY_CANT_BUILD_ON_SLOPE = (1 << 2),  // 0x4
-    WALL_SCENERY_IS_BANNER = (1 << 3),            // 0x8  // Probably indicates translucency
+    WALL_SCENERY_IS_DOUBLE_SIDED = (1 << 3),            // 0x8  // Probably indicates translucency
```
duncanspumpkin, line 119: You'll need to align the comment for the ci to pass. I would remove the "probably..." comment it is no longer relevant.

duncanspumpkin:
You are still missing a few renames. Do a find and replace of `WALL_SCENERY_IS_BANNER` replacing all instances with `WALL_SCENERY_IS_DOUBLE_SIDED`

src/openrct2/object/WallObject.cpp
```diff
@@ -120,6 +120,7 @@ void WallObject::ReadJson(IReadObjectContext* context, json_t& root)
                 { "hasTernaryColour",       WALL_SCENERY_HAS_TERNARY_COLOUR,    Json::FlagType::Normal },
                 { "hasGlass",               WALL_SCENERY_HAS_GLASS,             Json::FlagType::Normal },
                 { "isBanner",               WALL_SCENERY_IS_BANNER,             Json::FlagType::Normal },
```
duncanspumpkin, line 122: although counterintuitive this line needs the WALL_SCENERY_IS_BANNER replace with DOUBLE_SIDED version

KirilAngelov: Thank you for your help.

PR #14808, opened 2021-06-02
Rearrange WallSceneryEntry to enforce types

duncanspumpkin:
I encoutered a very minor mistake when i was correcting a bug and it got me thinking the way that Objects are handled is a little unsafe. So this looks to correct that by changing the structure and enforcing the type. I've done this like this so that we can do one object type at a time. Eventually I'd like to get rid of the legacy type but I feel we are too early for that at the moment.

### DIFF
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -1635,12 +1635,11 @@ static void window_tile_inspector_invalidate(rct_window* w)
         {
             bool canBeSloped = false;
             bool hasAnimation = false;
-            const auto sceneryEntry = tileElement->AsWall()->GetEntry();
-            if (sceneryEntry != nullptr)
+            const auto wallEntry = tileElement->AsWall()->GetEntry();
+            if (wallEntry != nullptr)
             {
-                const rct_wall_scenery_entry wallEntry = sceneryEntry->wall;
-                canBeSloped = !(wallEntry.flags & WALL_SCENERY_CANT_BUILD_ON_SLOPE);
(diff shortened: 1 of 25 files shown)


Gymnasiast (approved):

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -2094,6 +2093,7 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
                 int16_t wallType = tileElement->AsWall()->GetEntryIndex();
                 DrawTextBasic(dpi, screenCoords, STR_TILE_INSPECTOR_WALL_TYPE, &wallType, { COLOUR_WHITE });
 
+                // Banner info
```
Gymnasiast, line 2096: Why did you add this comment stating the bleeding obvious?

duncanspumpkin: When i rebased i noticed that line has disappeared in the previous pr so readded it

Gymnasiast: Yes, and? Why would you readd a comment that states the bleeding obvious?

ZehMatt: Just remove it?

duncanspumpkin: If you look at the rest of the file every single section of the drawing on the tile inspector has a little comment to say what that section is about. If i remove it again it will look a bit odd that this one section isn't commented. The tile inspector has a lot of special drawing for all of the different situations it has to handle so i feel these excessive comments in this case makes sense.

PR #14809, opened 2021-06-02
PathBitsEntry refactor

duncanspumpkin:
Follow on from #14808. Same changes but this time for the path additions/bits.

Found one potential bug whilst doing this. In the object validation code it accessed the wrong member variable. This is undefined behaviour but more importantly would cause valid objects to fail validation. I've not seen any reports of that though.


src/openrct2/object/FootpathItemObject.cpp
```diff
@@ -38,7 +38,7 @@ void FootpathItemObject::ReadLegacy(IReadObjectContext* context, OpenRCT2::IStre
     GetImageTable().Read(context, stream);
 
     // Validate properties
-    if (_legacyType.large_scenery.price <= 0)
+    if (_legacyType.price <= 0)
```
duncanspumpkin, line 41: This was a mistake that this pr corrected

PR #14811, opened 2021-06-03
Move gLastStringPos into rct_drawpixelinfo

ZehMatt:


duncanspumpkin:
It's interesting that only x is used

ZehMatt:
Might have been used before we updated some UI code, can eventually nuke this entirely.

duncanspumpkin (approved):

PR #14821, opened 2021-06-05
Fix #14751: Height limitation also affects trees

### LINKED ISSUE #14751: ironic error: Trees may be considered too tall to be built
image

Version: 0.3.3 Linux x64

Related: #14747

Gymnasiast:


duncanspumpkin:
Although it's a fix this is really a change.

PR #14828, opened 2021-06-05
Prevent exceeding limits of tiles

duncanspumpkin:
An NE park had >512 tiles at one location this caused memory corruption when inserting new tiles. This fix prevents the corruption and it will output an error but the game will still crash as callers are not checking for failing to insert.


tupaschoal (approved):

duncanspumpkin:
Im going to hold off on this as its also fixed by the map.cpp refactor

Issue #12626, opened 2020-08-10
Allow using saves from vanilla and RCTC to mark parks as finished


Gymnasiast:
Currently, if you load up a WW/TT save from vanilla, or a save from RCTC and win the scenario, it will not be marked as completed. This is because the scenario name embedded in the file does not match ours.

It would be a nice feature if OpenRCT2 had a map to convert one to the other, so that it _will_ mark the scenario as completed.

(I got the idea from https://www.reddit.com/r/openrct2/comments/i6p60g/transfer_rct_classic_scenario_progression_to/ )

PR #14832, opened 2021-06-06
Adding check to match RCTC save games with RCT2 scenarios

xtruan:
RCTC save games which were originally created from scenarios with the extension `.sea` are now correctly matched with their RCT2 `.sc6` equivalents when trying to add high score/flag as complete.

See #12626

### DIFF
--- a/contributors.md
+++ b/contributors.md
@@ -95,7 +95,6 @@ The following people are not part of the development team, but have been contrib
 * Adrian Zdanowicz (CookiePLMonster) - Misc.
 
 ## Bug fixes
-* Struan Clark (xtruan)
 * (KirilAngelov)
 * (halfbro)
 * (Myrtle)
@@ -172,6 +171,7 @@ The following people are not part of the development team, but have been contrib
 * (zrowny)
 * Emre Aydin (aemreaydin)
 * Daniel Karandikar (DKarandikar)
+* Struan Clark (xtruan)
 
 ## Toolchain
 * (Balletie) - macOS
--- a/contributors.md
+++ b/contributors.md
@@ -95,6 +95,7 @@ The following people are not part of the development team, but have been contrib
 * Adrian Zdanowicz (CookiePLMonster) - Misc.
 
 ## Bug fixes
+* Struan Clark (xtruan)
 * (KirilAngelov)
 * (halfbro)
 * (Myrtle)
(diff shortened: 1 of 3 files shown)


xtruan:
I've attached a Diamond Heights save file that is about to complete for testing. This was created/played through on RCTC (Android).

[Diamond Heights X.zip](https://github.com/OpenRCT2/OpenRCT2/files/6611203/Diamond.Heights.X.zip)

Gymnasiast:
Looks like a pretty nice solution! I’ll try to test it soon.

Perhaps we should also add code that does the opposite: allowing the user to finish scenarios using RCT2 saves when their scenarios are from RCTC.

contributors.md
```diff
@@ -95,6 +95,7 @@ The following people are not part of the development team, but have been contrib
 * Adrian Zdanowicz (CookiePLMonster) - Misc.
 
 ## Bug fixes
+* Struan Clark (xtruan)
```
AaronVanGeffen, line 98: This list is in chronological order. Please append your name at the end of the list instead.

xtruan: Moved to bottom, thanks!

xtruan:
> Looks like a pretty nice solution! I’ll try to test it soon.

Thanks!

> Perhaps we should also add code that does the opposite: allowing the user to finish scenarios using RCT2 saves when their scenarios are from RCTC.

I've updated my fix with code that should accomplish this.

Gymnasiast (approved):

src/openrct2/scenario/ScenarioRepository.cpp
```diff
@@ -423,6 +423,25 @@ class ScenarioRepository final : public IScenarioRepository
...
+
+            if (String::Equals(scenarioExtension, ".sea", true))
+            {
+                // Get scenario using RCT2 style name of RCTC scenario
+                scenario = GetByFilename((scenarioBaseName + ".sc6").c_str());
+            }
+            else if (String::Equals(scenarioExtension, ".sc6", true))
+            {
+                // Get scenario using RCTC style name of RCT2 scenario
+                scenario = GetByFilename((scenarioBaseName + ".sea").c_str());
```
AaronVanGeffen, line 441: These potentially recurse indefinitely if no matching scenario can be found.

Gymnasiast: Hm, how exactly?

AaronVanGeffen: Oh, my mistake. For some reason, I thought this was recursing from `GetByFilename`. Clearly, that is not the case. Sorry!

PR #14840, opened 2021-06-07
Refactor RIDE_STATUS to use strong enum

Sciencentistguy:


src/openrct2/actions/RideSetStatusAction.cpp
```diff
@@ -66,30 +66,21 @@ GameActions::Result::Ptr RideSetStatusAction::Query() const
         return res;
     }
 
-    if (_status >= RIDE_STATUS_COUNT)
```
duncanspumpkin, line 69: You will still need this code. We need to verify that the information over the network is valid.

Sciencentistguy: I've added this back, using a constexpr variable instead of an extra enum variant.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2342,25 +2342,25 @@ static void window_ride_main_invalidate(rct_window* w)
         SPR_TESTING,
         SPR_G2_SIMULATE,
     };
-    window_ride_main_widgets[WIDX_OPEN].image = spriteIds[ride->status];
+    window_ride_main_widgets[WIDX_OPEN].image = spriteIds[static_cast<uint8_t>(ride->status)];
```
ZehMatt, line 2345:
```suggestion
    window_ride_main_widgets[WIDX_OPEN].image = spriteIds[EnumValue(ride->status)];
```

src/openrct2/actions/RideSetStatusAction.cpp
```diff
@@ -66,7 +66,7 @@ GameActions::Result::Ptr RideSetStatusAction::Query() const
...
+    if (static_cast<uint8_t>(_status) >= RideStatusMax)
```
ZehMatt, line 69:
```suggestion
    if (EnumValue(_status) >= RideStatusMax)
```

src/openrct2/actions/RideSetStatusAction.cpp
```diff
@@ -75,21 +75,21 @@ GameActions::Result::Ptr RideSetStatusAction::Query() const
         return res;
     }
 
-    res->ErrorTitle = _StatusErrorTitles[_status];
+    res->ErrorTitle = _StatusErrorTitles[static_cast<uint8_t>(_status)];
```
ZehMatt, line 78:
```suggestion
    res->ErrorTitle = _StatusErrorTitles[EnumValue(_status)];
```

ZehMatt (changes requested):
I left a few comments, otherwise looking good.

duncanspumpkin:
In general we usually add another entry to the enum called Count so that you can do these bounds checks. I'm not opposed to using a separate constexpr though.

Sciencentistguy:
I think that using a constexpr variable is better than an extra enum variant for two main reasons:

- It is more obvious what it is for, upon a cursory reading of the code. I, for one, didn't realise that `RIDE_STATUS_COUNT` was not a status that rides could have, named `Count`, until I encountered the usage of this variant while writing this patch.
- If an enum variant is used for this, then `-Wswitch` will go off when a switch statement doesn't handle the (meaningless) case of `RideStatus::Count`.

Sciencentistguy:
Also; I think that the reason the CI is red on the latest commit is to do with the Fastly outage that happened, and not that the code itself is broken. I am not that familiar with Github Actions, can I request that it reruns?

src/openrct2/rct1/RCT1.h
```diff
@@ -64,7 +64,7 @@ struct rct1_ride
     colour_t track_primary_colour;                            // 0x01E
     colour_t track_secondary_colour;                          // 0x01F
     colour_t track_support_colour;                            // 0x020
-    uint8_t status;                                           // 0x021
+    RideStatus status;                                        // 0x021
```
duncanspumpkin, line 67: In general the header files in rct1 and rct2 are left without the enum classes. This is because the field is meant to be verified that it is within bounds before inporting into the game.

ZehMatt:
> I think that using a constexpr variable is better than an extra enum variant for two main reasons:
> 
> * It is more obvious what it is for, upon a cursory reading of the code. I, for one, didn't realise that `RIDE_STATUS_COUNT` was not a status that rides could have, named `Count`, until I encountered the usage of this variant while writing this patch.
> * If an enum variant is used for this, then `-Wswitch` will go off when a switch statement doesn't handle the (meaningless) case of `RideStatus::Count`.

I partially agree, it has a disadvantage that you require to update two things for when the enum gets changed. I think for consistency reasons we should continue to use Count in the enum.

Sciencentistguy:
I think that making this case worse for consistency is a bad idea. I will make the change here, but @ZehMatt what would your opinion be on a PR to move all enum bounds checking like this from an extra variant to a constexpr variable.

ZehMatt:
> I think that making a worse decision for consistency's sake is a bad idea. I will make the change here, but @ZehMatt what would your opinion be on a PR to move all enum bounds checking like this from an extra variant to a constexpr variable?

We had discussions about this before on how to treat enums/flags/indices etc. We ended up with what we have in the code base right now. I'm not opposed to changing it but I feel like we should think about perhaps a better solution in general that maybe involves some template uses and also perhaps look at existing libraries maybe. However this is also not up to me, we can discuss this in a new issue or in Discord perhaps.

ZehMatt (approved):
Looks good to me

PR #14841, opened 2021-06-07
Redesign About window

Broxzier:
Crazy Castle 2021-06-07 20-32-55

The button is only visible when a build hash is actually set.

### DIFF
--- a/src/openrct2-ui/windows/About.cpp
+++ b/src/openrct2-ui/windows/About.cpp
@@ -218,8 +218,9 @@ static void window_about_openrct2_paint(rct_window* w, rct_drawpixelinfo* dpi)
 
     auto const& versionPlaceholder = w->widgets[WIDX_VERSION];
     auto width = versionPlaceholder.right - versionPlaceholder.left;
-    auto height = versionPlaceholder.bottom - versionPlaceholder.top;
-    auto centerPos = w->windowPos + ScreenCoordsXY(versionPlaceholder.left + width / 2, versionPlaceholder.top + height / 2);
+    auto centerX = versionPlaceholder.left + width / 2;
+    auto centerY = (versionPlaceholder.top + versionPlaceholder.bottom - font_get_line_height(FontSpriteBase::MEDIUM)) / 2;
+    auto centerPos = w->windowPos + ScreenCoordsXY(centerX, centerY);
     DrawTextWrapped(dpi, centerPos, width, STR_STRING, ft, { w->colours[1], TextAlignment::CENTRE });
 }
 
--- a/src/openrct2-ui/windows/About.cpp
+++ b/src/openrct2-ui/windows/About.cpp
@@ -7,10 +7,12 @@
  * OpenRCT2 is licensed under the GNU General Public License version 3.
  *****************************************************************************/
 
+#include <SDL_clipboard.h>
 #include <openrct2-ui/interface/Widget.h>
 #include <openrct2-ui/windows/Window.h>
 #include <openrct2/Context.h>
 #include <openrct2/OpenRCT2.h>
+#include <openrct2/Version.h>
 #include <openrct2/drawing/Drawing.h>
 #include <openrct2/localisation/Localisation.h>
 #include <openrct2/localisation/LocalisationService.h>
@@ -18,9 +20,9 @@
 #include <openrct2/ui/UiContext.h>
 
 static constexpr const int32_t WW = 400;
-static constexpr const int32_t WH = 385;
+static constexpr const int32_t WH = 352;
 static constexpr const rct_string_id WINDOW_TITLE = STR_ABOUT;
-constexpr int32_t TABHEIGHT = 50;
+static constexpr const int32_t TABHEIGHT = 50;
 
 // clang-format off
 enum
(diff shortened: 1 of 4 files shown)


src/openrct2/Version.cpp
```diff
@@ -53,6 +53,24 @@ const char gVersionInfoFull[] = OPENRCT2_NAME ", "
 #endif
     ;
 
+bool has_build_hash()
```
ZehMatt, line 56: You could simplify this access with something like ```std::optional<std::string_view> GetBuildHash()```

src/openrct2/Version.h
```diff
@@ -78,6 +78,9 @@
 #    define OPENRCT2_CUSTOM_INFO ""
 #endif
 
+extern bool has_build_hash();
```
duncanspumpkin, line 81: Remove the extern.

All new functions should be HasBuildHash format

Gymnasiast:
The window is getting crowded. How about making the button smaller, putting the "copy" graphic on it and putting it next the build hash?

Broxzier:
@Gymnasiast The length of the version string can get quite long. "provided by GitHub" gets appended for official builds. Perhaps the version text could have a fixed width and wrap to the next line if necessary, like how the names in the tabs are done as well.

The About-window could do with some refactoring to make making changes easier. Subtracting the tab height from the bottom for the button placement makes no sense for instance.

Quick mockup with improved UI design:
image
Just like it currently is, the update button is only visible when there is an update.

pizza2004:
Would it be possible to simply make it so that clicking on the build hash itself copied it to the clipboard?  Then it could say as much here in the issues without adding an entire extra button.  You could even make it a button and have a little, "Copied to clipboard." message pop up when you click it if necessary, so there's not a totally unnecessary extra button.

AaronVanGeffen:
> Quick mockup with improved UI design:

While I understand it looks a little cluttered, I'm not a fan of the update button residing in the tab bar…

Perhaps the buttons could be shrunk horizontally and enlarged vertically instead, with the text wrapping? That way, they could be placed next to each other horizontally. (Widening the buttons if the update button is invisible, I suppose.)

Broxzier:
> While I understand it looks a little cluttered, I'm not a fan of the update button residing in the tab bar…

I based that on the Object Selection window's filter button. Redesigning this window is a bit out of scope for this PR, though. This PR is only intended to make it easier for users to get the build information.

Broxzier:
Made a new mockup anyway, thoughts?

image

I'd imagine the copy button then copying the entire version, rather than just the build hash.

tupaschoal:
> Made a new mockup anyway, thoughts?
> 
> image
> 
> I'd imagine the copy button then copying the entire version, rather than just the build hash.

I like it, though the copy doesn't look "clickable" (as it doesn't seek raised) and I'm assuming it has a meaningful tooltip

ZehMatt:
> Made a new mockup anyway, thoughts?
> 
> image
> 
> I'd imagine the copy button then copying the entire version, rather than just the build hash.

I like it, maybe you could center the contents a bit to make it look a bit less empty? I'm assuming its the height for the other tab.

Gymnasiast:
@ZehMatt The height of the other tab can be reduced, as we removed some text and buttons from there earlier.

AaronVanGeffen:
I like it! Can I suggest moving either the version text or the copy button slightly vertically? In the screenshot above, they're not aligned:

foo

Broxzier:
Turns out that calling `DrawTextWrapped` with `TextAlignment::CENTRE` in the text paint struct does not properly center the text vertically; half the line height has to be subtracted to actually draw it in the middle. I'm hesitant to changing `DrawTextWrapped` itself since that gets used quite a lot throughout the code base, so for this PR I chose to simply subtract half the line height from `centerY`.

Gymnasiast:
That’s to be expected, DrawTextWrapped aligns to the top and not the vertical centre.

src/openrct2-ui/windows/About.cpp
```diff
@@ -221,8 +216,12 @@ static void window_about_openrct2_paint(rct_window* w, rct_drawpixelinfo* dpi)
     auto ft = Formatter();
     ft.Add<const char*>(buffer);
 
-    aboutCoords.y = w->windowPos.y + WH - 25;
-    DrawTextWrapped(dpi, aboutCoords, width, STR_STRING, ft, { w->colours[1], TextAlignment::CENTRE });
+    auto const& versionPlaceholder = w->widgets[WIDX_VERSION];
+    auto width = versionPlaceholder.right - versionPlaceholder.left;
+    auto centerX = versionPlaceholder.left + width / 2;
+    auto centerY = (versionPlaceholder.top + versionPlaceholder.bottom - font_get_line_height(FontSpriteBase::MEDIUM)) / 2;
+    auto centerPos = w->windowPos + ScreenCoordsXY(centerX, centerY);
```
Gymnasiast, line 223: centre, not center. We are using the Queen’s English.

Broxzier:
> That’s to be expected, DrawTextWrapped aligns to the top and not the vertical centre.

But it _does_ vertically allign the text when providing the `TextAlignment::CENTRE` argument, only there's is a slight offset. I'd expect that it would move the text up with half the line height for every line, not for every line but the first. This illustration better shows why it's confusing, you'll need to provide the blue line as the Y coordinate instead of the Y position of the yellow line, while the yellow line is clearly in the middle.

image

Gymnasiast (approved):

PR #14857, opened 2021-06-08
Potentially Fix #14848. Rework check and reorganise

### LINKED ISSUE #14848: "Map elements limit has been reached" with an almost empty map
**OS:** Windows 10
**Version:** 0.3.3
**Commit/Build:** 072772a [Develop]

image

I am currently in a personal test park.
It appears to be a counting issue of some kind, probably something's forgetting to remove the count when it happens.
I recall having an error similar to this last month, where there were "too many banners" despite the fact that I didn't have more than 10 of them on my test park.

duncanspumpkin:
This version will defrag the current capacity before trying to increase the capacity

Not completely convinced if this will fix the problem with #14848 as i can't reproduce but it makes sense why this would fix it. At least it does in my head.

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -208,6 +208,15 @@ static bool map_check_free_elements_and_reorganise(size_t numElements)
     }
     else
     {
+        // First try reorganising the elements (as it removes fragmentation) and then recheck capacity.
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Okay ive confirmed that this fixes it after playing about with a very busy map.

PR #14862, opened 2021-06-09
Fix being able to exceed tile limit with a load near max

duncanspumpkin:
Due to the capacity of the vector when importing using its normal grow routine it could end up exceeding max elements. As we move the elements that over sized capacity would become the new limit.


duncanspumpkin:
Ive an alternative fix for this that will correct another issue

duncanspumpkin:
L1032 needs a decrement of the num in use tiles

Gymnasiast (approved):

PR #14864, opened 2021-06-09
Clean up create_new_banner()

Gymnasiast:
This function essentially did two things - looking up the next free banner ID and placing banners. Now it just does the latter. Also, setting the error message is not its responsibility, that should be set by the game action.


duncanspumpkin:
This function is about to be removed and completely reworked in both the nsf and the cherry pick of it https://github.com/OpenRCT2/OpenRCT2/pull/14788/files

PR #14868, opened 2021-06-09
Reduce usage of gGameCommandErrorText

Gymnasiast:


src/openrct2/Editor.cpp
```diff
@@ -453,60 +453,54 @@ namespace Editor
...
-        return ObjectType::None;
+        return { ObjectType::None, STR_NONE };
     }
 
     /**
      *
      *  rct2: 0x0066FEAC
      */
-    bool CheckPark()
+    std::pair<bool, rct_string_id> CheckPark()
```
tupaschoal, line 493: This one could be `std::optional<rct_string_id>` instead, as everytime it returns an actual error string, the `bool` is false

src/openrct2/actions/FootpathPlaceAction.cpp
```diff
@@ -313,14 +313,14 @@ GameActions::Result::Ptr FootpathPlaceAction::ElementInsertExecute(GameActions::
...
-            &res->Cost, crossingMode))
+    auto canBuild = MapCanConstructWithClearAt(
+        { _loc, zLow, zHigh }, &map_place_non_scenery_clear_func, quarterTile, GAME_COMMAND_FLAG_APPLY | GetFlags(),
+        crossingMode);
+    if (!entrancePath && canBuild->Error != GameActions::Status::Ok)
     {
-        return MakeResult(
-            GameActions::Status::NoClearance, STR_CANT_BUILD_FOOTPATH_HERE, gGameCommandErrorText, gCommonFormatArgs);
+        return canBuild;
     }
```
tupaschoal, line 322: Nice, seems like a lot more places can be cleaned up in this manner, maybe create a `good first issue` refactor issue? :)

src/openrct2/Editor.cpp
ZehMatt: Returning std::nullopt as "true" is a bit weird to me.

tupaschoal: I thought if it more like: "if it's empty, then there was no error"

src/openrct2/actions/FootpathPlaceAction.cpp
duncanspumpkin: I'm not sure it would make a good first issue you need to understand quite a bit of the codebase to know how to refactor this without causing issues.

PR #14880, opened 2021-06-10
Fix: Close button broken when failed to load version or changelog

Broxzier:
Found this little issue while working on #14841. When the version info or the changelog file could not be loaded, the close button would not be enabled. Only with hot-keys the window could be closed.


tupaschoal (approved):

tupaschoal:
I guess we should add a changelog

tupaschoal (commented):
Needs just to solve a merge conflict and looks good

PR #14884, opened 2021-06-11
Use std::vector for ride use memory

duncanspumpkin:
This is ready but I'll break replays and rerecord them when design is approved

### DIFF
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -2299,8 +2299,6 @@ void Guest::SetHasRidden(const Ride* ride)
 {
     OpenRCT2::RideUse::GetHistory().Add(sprite_index, ride->id);
 
-    RidesBeenOn[ride->id / 8] |= 1 << (ride->id % 8);
-
     SetHasRiddenRideType(ride->type);
 }
 
@@ -2315,8 +2313,6 @@ void Guest::SetHasRiddenRideType(int32_t rideType)
     rideType = OpenRCT2RideTypeToRCT2RideType(rideType);
 
     OpenRCT2::RideUse::GetTypeHistory().Add(sprite_index, rideType);
-
-    RideTypesBeenOn[rideType / 8] |= 1 << (rideType % 8);
 }
 
 bool Guest::HasRiddenRideType(int32_t rideType) const
@@ -7038,10 +7034,8 @@ Guest* Guest::Generate(const CoordsXYZ& coords)
 
     peep->Toilet = 0;
     peep->TimeToConsume = 0;
-    std::fill_n(peep->RidesBeenOn, 32, 0x00);
 
     peep->GuestNumRides = 0;
-    std::fill_n(peep->RideTypesBeenOn, 16, 0x00);
     peep->Id = gNextGuestNumber++;
     peep->Name = nullptr;
 
@@ -7381,11 +7375,6 @@ static bool IsThoughtShopItemRelated(const PeepThoughtType type)
 
 void Guest::RemoveRideFromMemory(ride_id_t rideId)
 {
-    uint8_t ride_id_bit = rideId % 8;
-    uint8_t ride_id_offset = rideId / 8;
-
-    // clear ride from potentially being in RidesBeenOn
-    RidesBeenOn[ride_id_offset] &= ~(1 << ride_id_bit);
     if (State == PeepState::Watching)
     {
         if (CurrentRide == rideId)
--- a/src/openrct2/peep/Guest.cpp
+++ b/src/openrct2/peep/Guest.cpp
@@ -45,6 +45,7 @@
 #include "../world/TileElementsView.h"
 #include "GuestPathfinding.h"
 #include "Peep.h"
+#include "RideUseSystem.h"
 #include "Staff.h"
 
 #include <algorithm>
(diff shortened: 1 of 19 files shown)


src/openrct2/peep/Guest.cpp
```diff
@@ -7356,3 +7376,27 @@ bool Guest::HasItem(ShopItem peepItem) const
...
+RideUseMemory& RideUseMemoryHandle::Create(uint16_t entityId)
+{
+    _rideUseMemory.insert(std::make_pair(entityId, RideUseMemory{}));
+    HasMemory = true;
+    return Get(entityId);
+}
+
+void RideUseMemoryHandle::Remove(uint16_t entityId)
+{
+    _rideUseMemory.erase(entityId);
```
duncanspumpkin, line 7401: Must set has memory to false

IntelOrca:
Why have you chosen a map as the way to do this? Doesn't that just incur a lookup cost every time it is accessed? What is the reason to do this over just having a pointer within the Entity struct?

ZehMatt:
> @duncanspumpkin We already have a pointer for name and patrol area, so this would only be another one. And you have still hooked the DestroyEntity method to clean up the memory. Even if you do use a separate table to avoid memory leaks, you could use an array of unique_ptr or something rather than an unordered_map. You could key on a new unique integer you assign to each entity if and only if they require the extra memory.

Please stop using more pointers. Memory should be stored in the corresponding system and only accessed via a unique key/handle in this case entity/peep id. 
Names should go in a name table/string resource System.
Patrol areas should go in the Patrol Areas System.
Rides been on should go into the Ride History System.
etc.

Adding more and more data to the entity will not benefit the performance over time. With systems you only ever access the data when it has to.

IntelOrca:
I wouldn't use pointers if Entity was allowed to have RAII types. I would for example use `std::string` instead of `char*` for name. And I much prefer that over a separate "system" as you call it.

ZehMatt:
> I wouldn't use pointers if Entity was allowed to have RAII types. I would for example use `std::string` instead of `char*` for name. And I much prefer that over a separate "system" as you call it.

The point of systems is that the owner of memory is explicit by that. RAII is fine but std::string still has a pointer internally and variable to hold the size, most implementations even have some extra space for small string optimizations. The reason why ECS is a lot faster is because you are not packing data into entities rather into components which are stored separately. The other benefit is that every system is also responsible for serialization, so for each system you would get a blob of data (could use this as a literal chunk for the save) because by avoiding pointers this makes loading and saving data incredibly easy and we can keep the data densely packed. Things like std::string and std::vector and anything that manages memory will fragment the memory access.

I highly recommend reading about ECS and how it generally benefits games, this also enforces a better structure by design.

IntelOrca:
ECS is not the answer to anything. I don't see much benefit of name being in a separate system. The ride list thing I do, but it would be nice if it didn't use `unordered_map`.

ZehMatt:
> ECS is not the answer to anything. I don't see much benefit of name being in a separate system. The ride list thing I do, but it would be nice if it didn't use `unordered_map`.

What should it use? I benchmarked also vector<pair<peep, ride>> and its a lot slower and consumes way more memory.

IntelOrca:
> > ECS is not the answer to anything. I don't see much benefit of name being in a separate system. The ride list thing I do, but it would be nice if it didn't use `unordered_map`.
> 
> What should it use? I benchmarked also vector<pair<peep, ride>> and its a lot slower and consumes way more memory.

Use a unique index to refer to that system, rather than peep ID.

src/openrct2/peep/RideUseSystem.cpp
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#include "RideUseSystem.h"
+
+namespace OpenRCT2::RideUse
+{
+    static RideHistory _history;
+    static RideTypeHistory _typeHistory;
```
IntelOrca, line 15: How do we go about avoiding new globals? Can this be part of a game state?

duncanspumpkin: yes game state can contain this.

ZehMatt: I think we should let this slip for now. Do also properly address this can also better organize this I think, we should create a new Milestone addressing globals and identify code and create corresponding issues. The GetContext() thing on its own is exactly the anti pattern we try to avoid really.

ZehMatt (approved):
Needs to rebased otherwise okay with me.

src/openrct2/peep/Guest.cpp
```diff
@@ -7373,11 +7375,6 @@ static bool IsThoughtShopItemRelated(const PeepThoughtType type)
 
 void Guest::RemoveRideFromMemory(ride_id_t rideId)
 {
-    uint8_t ride_id_bit = rideId % 8;
-    uint8_t ride_id_offset = rideId / 8;
-
-    // clear ride from potentially being in RidesBeenOn
-    RidesBeenOn[ride_id_offset] &= ~(1 << ride_id_bit);
```
Gymnasiast, line 7380: Shouldn't this have been replaced with a call to remove it from the ride use history?

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2832,14 +2833,24 @@ template<> void S4Importer::ImportEntity<Guest>(const RCT12SpriteBase& srcBase)
...
+        if (src->ride_types_been_on[i / 8] & (1 << (i % 8)))
+        {
+            typesBeenOn.push_back(i);
+        }
     }
-    for (size_t i = 0; i < 16; i++)
+    for (uint16_t i = 0; i < RCT12_MAX_RIDES_IN_PARK; i++)
     {
-        dst->RideTypesBeenOn[i] = src->ride_types_been_on[i];
+        if (src->rides_been_on[i / 8] & (1 << (i % 8)))
```
Gymnasiast, line 2847: This needs to be moved into a method, rather than just copypasting it into both of the importer.

src/openrct2/world/Sprite.cpp
```diff
@@ -537,17 +542,29 @@ void sprite_set_coordinates(const CoordsXYZ& spritePos, SpriteBase* sprite)
     sprite->z = spritePos.z;
 }
 
+static void DestroyEntity(SpriteBase* entity)
```
Gymnasiast, line 545: This is in essence the same name as `sprite_remove`. It’s also a very non-descriptive name, as it doesn’t actually destroy the entity or anything close to it.

src/openrct2/peep/Guest.cpp
duncanspumpkin: It's handled differently. At this point it has already been removed from the ride history.

src/openrct2/world/Sprite.cpp
duncanspumpkin: It's part of sprite remove and the naming is from the nsf branch where it also destroys the staff patrol area. I can rename it but will cause merge with nsf to be more annoying.

Gymnasiast: OK, can you create an issue to change it, then?

duncanspumpkin: Seems the nsf had a different name so ive swapped to that name `FreeEntity`

PR #14904, opened 2021-06-15
Close #14903. Use map_is_edge

### LINKED ISSUE #14903: Create function to check if coordinate is at map edge
See here https://github.com/OpenRCT2/OpenRCT2/pull/14894#discussion_r650490127

duncanspumpkin:


src/openrct2-ui/windows/Map.cpp
```diff
@@ -1578,7 +1578,7 @@ static void map_window_set_pixels(rct_window* w)
 
     for (int32_t i = 0; i < MAXIMUM_MAP_SIZE_TECHNICAL; i++)
     {
-        if (x > 0 && y > 0 && x < gMapSizeUnits && y < gMapSizeUnits)
+        if (!map_is_edge({ x, y }))
```
duncanspumpkin, line 1581: This is actually slightly different to before but i think it is correct.

src/openrct2/actions/PlaceParkEntranceAction.cpp
```diff
@@ -51,8 +51,7 @@ GameActions::Result::Ptr PlaceParkEntranceAction::Query() const
     res->Expenditure = ExpenditureType::LandPurchase;
     res->Position = { _loc.x, _loc.y, _loc.z };
 
-    if (!LocationValid(_loc) || _loc.x <= 32 || _loc.y <= 32 || _loc.x >= (gMapSizeUnits - 32)
-        || _loc.y >= (gMapSizeUnits - 32))
+    if (!LocationValid(_loc) || map_is_edge(_loc))
```
duncanspumpkin, line 54: This is also slightly different. <= 32 is now <32. Will need to check if that is correct

duncanspumpkin: No its not. Reverting.

Gymnasiast (approved):

PR #14905, opened 2021-06-15
Minor paint refactor

duncanspumpkin:
Names a few unknown flags.

I've no idea why there is both IsPassedSurface  and DidPassSurface. They are both set in the same situations. I would remove it but not sure if @OpenRCT2/developers have any better ideas as to why it exists.

### DIFF
--- a/src/openrct2/paint/tile_element/Paint.TileElement.cpp
+++ b/src/openrct2/paint/tile_element/Paint.TileElement.cpp
@@ -45,35 +45,14 @@ const int32_t SEGMENTS_ALL = SEGMENT_B4 | SEGMENT_B8 | SEGMENT_BC | SEGMENT_C0 |
  *
  *  rct2: 0x0068B35F
  */
-void tile_element_paint_setup(paint_session* session, int32_t x, int32_t y)
-{
-    if (x < gMapSizeUnits && y < gMapSizeUnits && x >= 32 && y >= 32)
-    {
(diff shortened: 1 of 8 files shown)


src/openrct2/paint/tile_element/Paint.TileElement.cpp
```diff
@@ -45,35 +45,14 @@ const int32_t SEGMENTS_ALL = SEGMENT_B4 | SEGMENT_B8 | SEGMENT_BC | SEGMENT_C0 |
...
- *  rct2: 0x0068B2B7
- */
-void sub_68B2B7(paint_session* session, const CoordsXY& mapCoords)
+void tile_element_paint_setup(paint_session* session, const CoordsXY& mapCoords, bool isTrackPiecePreview)
 {
     if (mapCoords.x < gMapSizeUnits && mapCoords.y < gMapSizeUnits && mapCoords.x >= 32 && mapCoords.y >= 32)
     {
         paint_util_set_segment_support_height(session, SEGMENTS_ALL, 0xFFFF, 0);
         paint_util_force_set_general_support_height(session, -1, 0);
+        session->Unk141E9DB = isTrackPiecePreview ? PaintSessionFlags::IsTrackPiecePreview : 0;
```
tupaschoal, line 54: Should the `0` be another named variable?

duncanspumpkin: No it's just to clear the flags

tupaschoal (approved):
Looks ok to me, but I can't answer for the flags, as I have no clue either.

PR #14908, opened 2021-06-15
Allow adding const/volatile types to the formatter

Broxzier:
Ran into this little issue when trying out some things with the formatter.

### DIFF
--- a/src/openrct2/localisation/Formatter.h
+++ b/src/openrct2/localisation/Formatter.h
@@ -77,16 +77,16 @@ public:
 
         // clang-format off
         static_assert(
-            std::is_same_v<TSpecified, char*> ||
-            std::is_same_v<TSpecified, const char*> ||
-            std::is_same_v<TSpecified, int16_t> ||
-            std::is_same_v<TSpecified, int32_t> ||
-            std::is_same_v<TSpecified, money32> ||
-            std::is_same_v<TSpecified, rct_string_id> ||
-            std::is_same_v<TSpecified, uint16_t> ||
-            std::is_same_v<TSpecified, uint32_t> ||
-            std::is_same_v<TSpecified, utf8*> ||
-            std::is_same_v<TSpecified, const utf8*>
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, char*> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, const char*> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, int16_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, int32_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, money32> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, rct_string_id> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, uint16_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, uint32_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, utf8*> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, const utf8*>
         );
         // clang-format on
 


src/openrct2/localisation/Formatter.h
```diff
@@ -77,16 +77,16 @@ class Formatter
...
-            std::is_same_v<TSpecified, const utf8*>
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, char*> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, const char*> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, int16_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, int32_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, money32> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, rct_string_id> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, uint16_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, uint32_t> ||
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, utf8*> ||
```
duncanspumpkin, line 88: Won't this fall on line 96/100 if the type is const?

Broxzier: Nope, removing `const` and `volatile` does not change the size nor does it affect `std::is_integral`.

Somewhat related note: It is actually possible to pass a static const to the function while specifying a non-const type, but since the function doesn't write to the argument, it doesn't matter.

duncanspumpkin: Do we still require all of the is_same_v's then for the const char* and const utf8* now that we remove cv i think they are not required.

src/openrct2/localisation/Formatter.h
```diff
@@ -77,16 +77,16 @@ class Formatter
...
-            std::is_same_v<TSpecified, const char*> ||
-            std::is_same_v<TSpecified, int16_t> ||
-            std::is_same_v<TSpecified, int32_t> ||
-            std::is_same_v<TSpecified, money32> ||
-            std::is_same_v<TSpecified, rct_string_id> ||
-            std::is_same_v<TSpecified, uint16_t> ||
-            std::is_same_v<TSpecified, uint32_t> ||
-            std::is_same_v<TSpecified, utf8*> ||
-            std::is_same_v<TSpecified, const utf8*>
+            std::is_same_v<typename std::remove_cv<TSpecified>::type, char*> ||
```
ZehMatt, line 80: You could make this a lot cleaner by using
```cpp
using TCVType = typename std::remove_cv_t<TSpecified>;
```
Above the assert.

Broxzier: Not sure if that's really cleaner. That removes the multiple calls to `std::remove_cv`, however, unless you go with a verbose identifier, it isn't obvious which types it actually is the base type of. `TSpecified` and `TDeduced` can be different. Since it's all computed at compile time, I don't see a reason to make it less obvious.

If you want to avoid the duplicated uses of `std::remove_cv`, perhaps a better approach to checking if the type is any of a given list. This could be achieved with a variadic template.

```cpp
template<class T, class... Rest>
inline constexpr bool is_any = (std::is_same_v<T, Rest> || ...);

...

static_assert(
    is_any<
        typename std::remove_cv<TSpecified>::type, char*, const char*, int16_t, int32_t, money32, rct_string_id, uint16_t, uint32_t, utf8*,
        const utf8*>,
    "Specified type is not supported");
```

And while this is compacter and has fewer duplicated lines, it is less readable.

src/openrct2/localisation/Formatter.h
Broxzier: Yes they are needed, since these are pointers, not types. std::remove_const/cv would make it a mutable pointer type.

```cpp
std::is_same_v<const char*, char*>; // False
std::is_same_v<char* const, char*>; // False
std::is_same_v<typename std::remove_cv<const char*>::type, char*>; // False
std::is_same_v<typename std::remove_cv<char* const>::type, char*>; // True
```

duncanspumpkin (approved):

PR #14919, opened 2021-06-18
Add maian_sos’ scenarios as the official OpenRCT2 campaign

Gymnasiast:


Gymnasiast:
I split off the window refactor. Closing this in favour of proper scenario list support (which would work the same as themes or title sequences).

PR #14926, opened 2021-06-19
A nice suggestion, the Malgev Track creation

TheClaudeQC:
So, i make this new type of transport ride, inspired by the Open TTD's train type. It will have same train as the rct2 trams, but whit less mass and higher max speed.


TheClaudeQC:
go check out https://github.com/OpenRCT2/OpenRCT2/pull/14927 for the track's sprites

duncanspumpkin:
This isn't how to do a pull request for a new track. Make one pull request for all of it. There is no way to test this without it all together. You haven't made any of the necessary code changes.

X123M3-256:
This isn't how you do this at all.

The contents of the `resources/g2/track/maglevtrack` folder should be your track sprites - not JSON files. If you want to create a vehicle type to go with your track, that should go in the object repository not this one. I told you which files to modify on Discord - you need to look at those files, see how existing rides were implemented, and then add your own code.

As I said, the easiest way to start is probably to copy the single rail code and s/single_rail/maglev. The first thing I would try to do is duplicate an existing ride type, then you have something that works and you can start replacing the sprites with yours.

Also, JSON objects do not have a "ride_id" property - the "type" property is what you're looking for. You will need to edit the loader code to map the ride type name to the ride id.

You should submit *one*  PR with all your changes once you have your track finished- not separate PRs for every file you want to change, and you should test your changes to make sure they work first (as you clearly didn't test this).

PR #14954, opened 2021-06-25
Minor cleanups

ZehMatt:
Some stuff found with static code analysis

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -253,7 +253,8 @@ InteractionInfo ViewportInteractionGetItemRight(const ScreenCoordsXY& screenCoor
     if ((gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER) && gS6Info.editor_step != EditorStep::RollercoasterDesigner)
         return info;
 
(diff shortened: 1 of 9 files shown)


src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -253,7 +253,8 @@ InteractionInfo ViewportInteractionGetItemRight(const ScreenCoordsXY& screenCoor
     if ((gScreenFlags & SCREEN_FLAGS_TRACK_DESIGNER) && gS6Info.editor_step != EditorStep::RollercoasterDesigner)
         return info;
 
-    auto flags = static_cast<int32_t>(~EnumsToFlags(ViewportInteractionItem::Terrain, ViewportInteractionItem::Water));
+    constexpr auto flags = static_cast<int32_t>(
```
tupaschoal, line 256: Which static checker flagged the constexpr? Sounds interesting

tupaschoal (approved):

src/openrct2-ui/interface/ViewportInteraction.cpp
ZehMatt: The one that comes with VS by default.

PR #14962, opened 2021-06-27
Fix 14945 Incorrect drop height penalty

duncanspumpkin:
Looks like a mistake from refactoring/implementation.


duncanspumpkin:
May need replays rerecorded.

ZehMatt (approved):

PR #14982, opened 2021-07-03
Add Malgun Gothic and change Nanum Gothic for Korean

telk5093:
# Add Malgun Gothic
I tested OpenRCT2 in Windows 11 today ~~and noticed that it seems Gulim is not exist in Windows 11.~~ (Maybe there is, I'm checking the facts)
Thus Korean users will run into mangled strings like ``ㅁㅁ`` as you can see below when Windows 11 is introduced at the late of this year, unless Windows 11 contains gulim.ttc again.

**Issue screenshot:** (Windows 11, not installed neither Gulim nor NanumGothic)
image

So adding ``Malgun Gothic`` as an alternative TTF font will solve this problem beforehand which will be occured in the future.
For notes, Malgun Gothic is introduced since Windows Vista and is used as a main font in modern Korean Windows.

**Applied screenshot:**
image

# Change Nanum Gothic
Nanum Gothic is introduced by [Naver](https://www.naver.com), which is the biggest site in Korean, and it is able to download via [its official webpage](https://hangeul.naver.com/font).
The official download file of NanumGothic is a zip file which contains four ``*.ttf`` files, not ``*.ttc``.
So OpenRCT2 does not detect (and maybe is not detecting currently too) NanumGothic in current state and causes mangled strings.
I have no idea why @AaronVanGeffen used *.ttc in here: (please comment if you had some thoughts at that moment 😃)
https://github.com/OpenRCT2/OpenRCT2/blob/a5a63f839aca539ad75abf09a765e590e5ba973b/src/openrct2/interface/Fonts.cpp#L76

I changed them into ``*.ttf`` and changes y_offset value and hinting options for NanumGothic as well.

**Applied screenshot:** (NanumGothic)
image

### DIFF
--- a/src/openrct2/interface/FontFamilies.cpp
+++ b/src/openrct2/interface/FontFamilies.cpp
@@ -34,6 +34,7 @@ TTFontFamily const TTFFamilyJapanese = {
 TTFontFamily const TTFFamilyKorean = {
     &TTFFontGulim,
     &TTFFontNanum,
+    &TTFFontMalgun,
 };
 
 TTFontFamily const TTFFamilySansSerif = {
--- a/src/openrct2/interface/Fonts.cpp
+++ b/src/openrct2/interface/Fonts.cpp
@@ -72,6 +72,12 @@ TTFFontSetDescriptor TTFFontGulim = { {
     { "gulim.ttc", "Gulim", 12, 1, 0, 15, HINTING_THRESHOLD_MEDIUM, nullptr },
 } };
 
+TTFFontSetDescriptor TTFFontMalgun = { {
+    { "malgun.ttf", "Malgun Gothic", 10, 1, -3, 10, HINTING_THRESHOLD_LOW, nullptr },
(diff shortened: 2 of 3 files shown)


AaronVanGeffen (approved):
Thanks for the patch. :) It's nice to see the game is being made future proof in this way!

As for why I used `.ttc` for Nanum Gothic… Most likely, I found it as a .ttc (**T**rue**T**ype font **C**ollection) somewhere, but I rely on others to test on Windows. (It appears the line was introduced in 2017, so that explains that. :slightly_smiling_face: )

If your Windows install is shipping Nanum Gothic as a `.ttf` file instead, let's go with that.

The patch in general looks good to me.

YJSoft:
NanumGothic was distributed with ttc format in past(when NanumGothic only have three variant). but when NAVER decided to add more fonts to Nanum Font family, they just seperated ttc with ttfs, and provided Installer for easy install.

Gymnasiast:
Oh, could you open a follow-up PR to also update the changelog? I forgot about that.

telk5093:
@Gymnasiast
Thanks to @Lastorder-DC, #15004 adds changelog instead of me ;)

Issue #15022, opened 2021-07-19
Remove usage of goto


Margen67:
- [x] https://github.com/OpenRCT2/OpenRCT2/blob/e0daac9442dec8bf1502690f6310af50c7bb5f25/src/openrct2/world/Map.cpp#L1353
- [x] https://github.com/OpenRCT2/OpenRCT2/blob/e0daac9442dec8bf1502690f6310af50c7bb5f25/src/openrct2/world/Map.cpp#L1364
- [x] https://github.com/OpenRCT2/OpenRCT2/blob/e0daac9442dec8bf1502690f6310af50c7bb5f25/src/openrct2/world/Map.cpp#L1478
- [x] https://github.com/OpenRCT2/OpenRCT2/blob/4f6d0a00fd0cd82288588544acbe42898bbfc74e/src/openrct2/ride/Vehicle.cpp#L3955
- [x] https://github.com/OpenRCT2/OpenRCT2/blob/4f6d0a00fd0cd82288588544acbe42898bbfc74e/src/openrct2/ride/Vehicle.cpp#L3961
- [ ] https://github.com/OpenRCT2/OpenRCT2/blob/193be924d46a8f8da2ea221460c8ce94ff1b4d35/src/openrct2/ride/Vehicle.cpp#L8381
- [ ] https://github.com/OpenRCT2/OpenRCT2/blob/193be924d46a8f8da2ea221460c8ce94ff1b4d35/src/openrct2/ride/Vehicle.cpp#L8711
- [ ] https://github.com/OpenRCT2/OpenRCT2/blob/193be924d46a8f8da2ea221460c8ce94ff1b4d35/src/openrct2/ride/Vehicle.cpp#L8719
- [ ] https://github.com/OpenRCT2/OpenRCT2/blob/193be924d46a8f8da2ea221460c8ce94ff1b4d35/src/openrct2/ride/Vehicle.cpp#L8724
- [ ] https://github.com/OpenRCT2/OpenRCT2/blob/193be924d46a8f8da2ea221460c8ce94ff1b4d35/src/openrct2/ride/Vehicle.cpp#L8727
- [ ] https://github.com/OpenRCT2/OpenRCT2/blob/193be924d46a8f8da2ea221460c8ce94ff1b4d35/src/openrct2/ride/Vehicle.cpp#L8811
[...]

PR #15024, opened 2021-07-19
Remove gotos from Map.cpp

Gymnasiast:
It’s easiest to view the commits one by one. They should be quite easy to follow.

Partial fix for https://github.com/OpenRCT2/OpenRCT2/issues/15022.

### DIFF
--- a/src/openrct2/world/Map.cpp
+++ b/src/openrct2/world/Map.cpp
@@ -1361,22 +1361,26 @@ std::unique_ptr<GameActions::ConstructClearResult> MapCanConstructWithClearAt(
             res->GroundFlags |= ELEMENT_IS_UNDERWATER;
             if (water_height < pos.clearanceZ)
             {
+                bool returnError = true;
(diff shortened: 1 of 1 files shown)


src/openrct2/world/Map.cpp
```diff
@@ -1361,22 +1361,26 @@ std::unique_ptr<GameActions::ConstructClearResult> MapCanConstructWithClearAt(
             res->GroundFlags |= ELEMENT_IS_UNDERWATER;
             if (water_height < pos.clearanceZ)
             {
+                bool returnError = true;
```
tupaschoal, line 1408: On commit `https://github.com/OpenRCT2/OpenRCT2/pull/15024/commits/41e564d16c73764fb8a567ed3765207e5f79255c`, seems like this would be better off as just `error` or something similar. Not a big deal though

tupaschoal (approved):
This seems great overall, I think it's alright and I have just one minor comment, nice job!

PR #15032, opened 2021-07-20
Fix #2987, #7417: Use game actions for most set commands

### LINKED ISSUE #2987: Set console commands can be used in multiplayer
**OS: Windows 7** 
**Version: 0.0.4.0 build caf8b45** 
- [ ] Original bug?
- [x] Multiplayer?

If you try to use **set** command in console , it cans affect the server even you don't have permission.

I have tried these:
- set park_open 0/1
- set money [money_value]
- set no_money 0/1

Margen67:
#2987 #7417

### DIFF
--- a/src/openrct2/interface/InteractiveConsole.cpp
+++ b/src/openrct2/interface/InteractiveConsole.cpp
@@ -17,8 +17,10 @@
 #include "../ReplayManager.h"
 #include "../Version.h"
 #include "../actions/ClimateSetAction.h"
+#include "../actions/ParkSetParameterAction.h"
 #include "../actions/RideSetPriceAction.h"
 #include "../actions/RideSetSettingAction.h"
+#include "../actions/ScenarioSetSettingAction.h"
 #include "../actions/SetCheatAction.h"
 #include "../actions/StaffSetCostumeAction.h"
 #include "../config/Config.h"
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Although I'm sure most of these errors would be permission denied they are not guaranteed to only be because of permissions. I wonder if the error should be more generic. @IntelOrca were you wanting to remove the console commands and make them into plugins instead?

ZehMatt:
I agree about the error handling, this is still a good change to avoid desyncs. There is also another thing to consider, the game action will also show an error window if it fails.

Margen67:
> 
> 
> I agree about the error handling, this is still a good change to avoid desyncs. There is also another thing to consider, the game action will also show an error window if it fails.

I just copied what other commands did:
https://github.com/OpenRCT2/OpenRCT2/blob/a371d1ff117e078a7e3a2e85417003a5cc3d9769/src/openrct2/interface/InteractiveConsole.cpp#L751-L769 https://github.com/OpenRCT2/OpenRCT2/blob/a371d1ff117e078a7e3a2e85417003a5cc3d9769/src/openrct2/interface/InteractiveConsole.cpp#L960-L977 https://github.com/OpenRCT2/OpenRCT2/blob/a371d1ff117e078a7e3a2e85417003a5cc3d9769/src/openrct2/interface/InteractiveConsole.cpp#L978-L990 https://github.com/OpenRCT2/OpenRCT2/blob/a371d1ff117e078a7e3a2e85417003a5cc3d9769/src/openrct2/interface/InteractiveConsole.cpp#L996-L1008

duncanspumpkin:
Just because the other commands do it doesn't mean that they are correct. I think a generic "set command failed" would be fine. As this is all going through the game actions you might not even be in a networked instance when it fails leading to a very cryptic error message.

ZehMatt:
We could add something like ```static bool CheckError(const GameActions::Result* res);``` and handle some of the error cases in there.

Margen67:
I fixed the error messages.

duncanspumpkin (approved):

PR #15035, opened 2021-07-20
Remove more gotos

Gymnasiast:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -3904,7 +3904,7 @@ void Vehicle::UpdateArriving()
     if (curRide == nullptr)
         return;
 
-    uint8_t unkF64E35 = 1;
+    bool stationBrakesWork = true;
```
duncanspumpkin: i think `hasBrakesFailure` and inverting would be a better name although you'll obviously have to work around the current interim bit that is currently called `hasBrakesFailure`

duncanspumpkin (approved):

PR #15043, opened 2021-07-21
Fix #15028, #15042. Crash when placing large scenery

### LINKED ISSUE #15028: Moving large scenery items through other scenery while having zero clearance and ctrl on crashes the game.
**OS:** [Windows 10]
**Version:** [0.3.4]
**Commit/Build:** [e0daac9]

If you move a large scenery object with zero clearance through a different object while holding ctrl it crashes the game. This doesn't happen always but it happens relatively often (had it happen 5 times within 10 minutes). It mostly happened when I also had ctrl on to keep the object on a certain height but it may have also happened without having ctrl on.

**Steps to reproduce:**
1. Take any large scenery object (For me it always happened with 2x2 objects but not sure if the size is relevant.
2. Click control
3. Drag it around for a bit through other scenery items, sometimes you may have to switch objects a couple of times. It may take up to 5 minutes for something to happen.
4. Sometimes the game will just randomly crash.

**Save game:**
I do not have a particular save file where this happened, but I was able to reproduce this bug in multiple different save files. Also different save files which were from different parks.

Hope this is enough information for you guys to fix this issue. Thanks in advance.

duncanspumpkin:
When the fragmented tile element limit is reached whilst placing a large scenery ghost the game will perform a reorg of the map elements invalidating tile element pointers. As large scenery are multi tiled this can happen mid action invalidating the pointer that was allocated for the first tile element. Large scenery actions pass back the first tile element pointer to the calling function for use with ghost removal. When this pointer is invalid it causes the crash. The ultimate fix for this would be to create an undo function for actions. As an interim the functions have been modified to instead return the first tile height.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,5 +1,6 @@
 0.3.4+ (in development)
 ------------------------------------------------------------------------
+- Fix: [#15028] Crash when placing large scenery.
 - Improved: [#12626] Allow using RCT2 saves to mark RCT Classic (.sea) parks as finished and vice versa. 
 
 0.3.4 (2021-07-19)
--- a/src/openrct2-ui/windows/TopToolbar.cpp
+++ b/src/openrct2-ui/windows/TopToolbar.cpp
@@ -2549,9 +2549,7 @@ static money32 try_place_ghost_large_scenery(
 
     gSceneryPlaceRotation = loc.direction;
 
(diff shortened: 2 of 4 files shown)


ZehMatt (approved):

Gymnasiast:
Looks good, needs changelog update.

PR #15054, opened 2021-07-22
Fix 15052

duncanspumpkin:
Another crash from holding bad tile element pointers.

### DIFF
--- a/src/openrct2-ui/windows/Banner.cpp
+++ b/src/openrct2-ui/windows/Banner.cpp
@@ -76,7 +76,6 @@ class BannerWindow final : public Window
 private:
     Banner* _banner;
     CoordsXYZ _bannerViewPos;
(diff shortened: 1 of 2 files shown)


src/openrct2-ui/windows/Banner.cpp
```diff
@@ -125,12 +133,11 @@ class BannerWindow final : public Window
         number = _number;
         _banner = GetBanner(number);
 
-        InitTileElement();
-        if (_bannerElement == nullptr)
+        auto* bannerElement = GetBannerElement();
+        if (bannerElement == nullptr)
             return;
 
-        frame_no = _bannerElement->GetBaseZ();
```
duncanspumpkin, line 132: This did not make sense and has been removed. It was not used by the window.

ZehMatt (approved):
LGTM

Issue #14476, opened 2021-04-14
Object Selection isn't synced in multiplayer


Margen67:
**Version:** 0.3.3
**Commit/Build:** https://github.com/OpenRCT2/OpenRCT2/commit/b24cdff0fb2ad1db4dffb304a3f6cd22976b29d2

Changing the Object Selection in multiplayer leads to a desync.

- [ ] Reproducible in RCT2 (vanilla)?
- [x] Specific to multiplayer?

**Steps to reproduce:**
1. Create a server.
2. Open Object Selection: `open object_selection`
3. Modify items.
4. Selections won't sync. (if you're the host players will have to rejoin)

duncanspumpkin:
At present there is no way to send objects over the network then there is the issues with adding vanilla objects that users might not have but can't be redistributed.

Issue #14475, opened 2021-04-14
Inventions list isn't synced in multiplayer


Margen67:
**Version:** 0.3.3
**Commit/Build:** https://github.com/OpenRCT2/OpenRCT2/commit/b24cdff0fb2ad1db4dffb304a3f6cd22976b29d2

Changing the invention list in multiplayer leads to a desync.

- [ ] Reproducible in RCT2 (vanilla)?
- [x] Specific to multiplayer?

**Steps to reproduce:**
1. Create a server.
2. Open Inventions list: `open inventions_list`
3. Move items.
4. Inventions won't sync. (if you're the host players will have to rejoin)

PR #15056, opened 2021-07-22
Disallow opening problematic windows in multiplayer

Margen67:
#14476 #14475


Gymnasiast:
I wonder how difficult it would be to make the object selection and inventions list sync properly. I guess the former is quite difficult, the latter might be feasible. What do you think @duncanspumpkin ?

duncanspumpkin:
Object list would be a complete pain to do, would need the ability to send objects over the action interface and then you have the issue of what happens if the clients don't have all of the same vanilla objects. 

Invention list could be done. Don't see any issues with that. Would need to move invention modifications to the game action interface though.

IntelOrca:
It seems sensible to disable them though until they are synced up. We could do the same with the console commands too.

PR #15058, opened 2021-07-22
Refactor lookup tables to use EnumMap

ZehMatt:
std::unordered_map is not the best type for having small lists of key/value mappings.

Performance comparison:
https://quick-bench.com/q/-JmgJYocMP1Qyo5BXRrIo5pIabM

I also moved out all the static initialization from functions into the global scope, having static initialization inside a function causes the compiler to emit code for thread-safe initialization. 
Illustration here: https://godbolt.org/z/48GEMvq3e

Also EnumMap uses less memory and is not as fragmented as the unordered_map


IntelOrca (approved):

src/openrct2/object/RideObject.cpp
```diff
@@ -969,155 +970,156 @@ bool RideObject::IsRideTypeShopOrFacility(uint8_t rideType)
...
+                                                   { "mini_rc", RIDE_TYPE_MINI_ROLLER_COASTER },
+                                                   { "mine_ride", RIDE_TYPE_MINE_RIDE },
+                                                   { "lim_launched_rc", RIDE_TYPE_LIM_LAUNCHED_ROLLER_COASTER },
+                                                   { "hypercoaster", RIDE_TYPE_HYPERCOASTER },
+                                                   { "hyper_twister", RIDE_TYPE_HYPER_TWISTER },
+                                                   { "monster_trucks", RIDE_TYPE_MONSTER_TRUCKS },
+                                                   { "spinning_wild_mouse", RIDE_TYPE_SPINNING_WILD_MOUSE },
+                                                   { "classic_mini_rc", RIDE_TYPE_CLASSIC_MINI_ROLLER_COASTER },
+                                                   { "hybrid_rc", RIDE_TYPE_HYBRID_COASTER },
+                                                   { "single_rail_rc", RIDE_TYPE_SINGLE_RAIL_ROLLER_COASTER } };
```
Gymnasiast, line 1058: This is way too far intended. Is there no way to just have it indented one extra level?

Broxzier: There is, just add a trailing comma, just like the other lists have.
```suggestion
                                                   { "single_rail_rc", RIDE_TYPE_SINGLE_RAIL_ROLLER_COASTER }, };
```

test/tests/EnumMapTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+    ASSERT_EQ(itF->second, TEnum::F);
+
+    auto itG = enumMap.find(TEnum::G);
+    ASSERT_NE(itG, enumMap.end());
+    ASSERT_EQ(itG->second, TEnum::G);
+
+    SUCCEED();
+}
+
+TEST(EnumMapTest, LookupContinousByKey)
```
Gymnasiast, line 128: Continuous

Gymnasiast (approved):

PR #15061, opened 2021-07-22
Fix #13465: Creating a scenario based on a won save game results in a scenario that's instantly won.

### LINKED ISSUE #13465: Creating a scenario from a won save game can only create scenarios that have already been won.
OS: Kubuntu Linux 20.04
Version: ??? Version info still not showing up on Ubuntu title screen (see #12909) updated as of this post
Commit/Build: ???

I'm creating a scenario based on a saved game, but the save file has already been won for its own scenario goals. The problem is that no matter what settings I try to use or what I try to modify in the save game file, the new scenario that's created is always already won from the start, even though the scenario goals clearly haven't been met.

I've asked for help elsewhere looking for a possible way to un-win the scenario as a workaround for this problem, but it seems like it shouldn't be a problem in the first place. Whatever flag is listing the saved game as 'already won', should be automatically removed as part of the process of making a new scenario based on it.

- [ no ] Specific to multiplayer?

**Steps to reproduce:**
1. Create a saved game.
2. In the cheats window, click 'Win Scenario'.
3. Save the game.
4: Exit to menu.
5: Create a scenario based on a saved game.
6: Create a new scenario based on the save from step 3.
7: Load the new scenario.
8: No matter what you set the new scenario's goals to, it's already won the moment you load it.

**Screenshots / Video:**

Purgatoria 2020-11-26 22-02-44
(Listing the insane goal in a separate issue.)

ocalhoun6:
Yeah ... I have _very_ little idea what I'm doing here. C++ is weird, and I'm not sure I'm doing this github stuff the right way, either.

Also very hazy on how setting these park flags works. I _think_ gParkFlags is some kind of array or vector full of binary true/false flags? But I'm not sure how it knows which flag to set based on these big variable names like "PARK_FLAGS_SCENARIO_COMPLETE_NAME_INPUT". And I'm new to C++, so these bitwise operators like &= and |= are kind of mind-boggling for me. Don't really understand why they're used this way. And _really_ not sure if I should be using &= or |= in this instance.

Also not sure if I'm doing everything that's necessary.
To fix this from the console takes two commands: 
`scenario.status = "inProgress" `
`park.setFlag("scenarioCompleteNameInput", false)`
But in the scenario.cpp code, I can't find _anything_ about scenario.status or the inProgress flag.

So, yeah ... this _might_ be an utter failure and only messing things up. But I think I've got it at least _close_ to where it needs to be. The extra line of code is in the right place, and if I got lucky enough to guess the correct operator _and_ there really aren't any other flags, this should work.

### DIFF
--- a/src/openrct2/Editor.cpp
+++ b/src/openrct2/Editor.cpp
@@ -140,6 +140,7 @@ namespace Editor
 
         // reset whether the scenario has been won, to make sure it hasn't:
         gParkFlags &= ~PARK_FLAGS_SCENARIO_COMPLETE_NAME_INPUT;
+        gScenarioCompletedCompanyValue = MONEY64_UNDEFINED;
 
         gScreenFlags = SCREEN_FLAGS_SCENARIO_EDITOR;
         gS6Info.editor_step = EditorStep::ObjectiveSelection;
--- a/src/openrct2/Editor.cpp
+++ b/src/openrct2/Editor.cpp
@@ -138,7 +138,7 @@ namespace Editor
         gS6Info.objective_arg_3 = gScenarioObjective.NumGuests;
         climate_reset(gClimate);
 
-        // reset whether the scenario has been won, to make sure it hasn't:
+        // Clear the scenario completion status
         gParkFlags &= ~PARK_FLAGS_SCENARIO_COMPLETE_NAME_INPUT;
         gScenarioCompletedCompanyValue = MONEY64_UNDEFINED;
 
--- a/src/openrct2/Editor.cpp
+++ b/src/openrct2/Editor.cpp
@@ -138,6 +138,9 @@ namespace Editor
         gS6Info.objective_arg_3 = gScenarioObjective.NumGuests;
         climate_reset(gClimate);
 
+        // reset whether the scenario has been won, to make sure it hasn't:
+        gParkFlags &= ~PARK_FLAGS_SCENARIO_COMPLETE_NAME_INPUT;
+
         gScreenFlags = SCREEN_FLAGS_SCENARIO_EDITOR;
         gS6Info.editor_step = EditorStep::ObjectiveSelection;
         gS6Info.category = SCENARIO_CATEGORY_OTHER;
--- a/contributors.md
+++ b/contributors.md
@@ -175,6 +175,7 @@ The following people are not part of the development team, but have been contrib
 * Struan Clark (xtruan)
 * Kane Shaw (seifer7)
 * Saad Rehman (SaadRehmanCS)
+* (ocalhoun6)
 
 ## Toolchain
 * (Balletie) - macOS
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#15084] [Plugin] Add "vehicle.crash" hook
 - Feature: [#15143] Added a shortcut key for Giant Screenshot.
+- Fix: [#13465] Creating a scenario based on a won save game results in a scenario that’s instantly won.
 - Fix: [#14316] Closing the Track Designs Manager window causes broken state.
 - Fix: [#15096] Crash when placing entrances in the scenario editor near the map corner.
 - Fix: [#15142] ToonTowner's mine roofs were moved into the pirate theme scenery group instead of the mine theme scenery group.


ocalhoun6:
Also ... I haven't figured out how to download and build my own branch yet ... especially how to do it without screwing up my main installation. Maybe I'll try to build it on a windows VM? Yeah ... I really wish I could test this to make sure it works.

For my own notes or in case anyone wants to try it, there are two things that need to be tested:
1: Open any scenario.
2: Win the scenario, either legitimately or by using the 'win scenario' cheat.
3: Save the scenario.
4: Convert the saved game to a scenario.
5: Open the new scenario as a new game.
6: Confirm that it doesn't show up as already completed.

1: Open any scenario.
2: _Don't_ win the scenario, either legitimately or by using the 'win scenario' cheat.
3: Save the scenario.
4: Convert the saved game to a scenario.
5: Open the new scenario as a new game.
6: Confirm that it doesn't show up as already completed. (To make sure that I didn't accidentally just invert the problem.)

ocalhoun6:
An alternate way to do this might be to edit scenario.cpp and and add this line at about line100.

There are pros and cons to both ways.
Doing it in editor.cpp means that newly created scenarios will be fixed and will work on other RCT versions.
Doing it in scenario.cpp means that it will be fixed when the game loads a faulty scenario -- which means that scenarios that were previously created with this fault will be able to be loaded properly.

If you want, maybe it could be done in _both_, to get the best of both worlds, redundantly fixing it when a new scenario is created _and_ when a faulty scenario that was built before this fix is loaded as a new game? It would be as simple as taking the line I added in editor.cpp and also adding it to the right place in scenario.cpp ... just let me know if you want it done that way.

As things are now, I wanted to minimize how intrusive I was being and limit the scope of my edits so that I couldn't screw up more than I was intending to work on.

ocalhoun6:
Oh, and I should also tag #13465 probably, so people reading that issue will know about this PR.

Broxzier:
Flags are used to save memory. Since there is not a single-bit structure in C++, the smallest type is `(unsigned) char` which is 8 bits. By giving each of its bit a different meaning, one `char` could be used to store 8 on/off values, saving you 7*8=56 bits in memory in case you would use a new `char` for each variable. Sawyer used flags a lot throughout the codebase.

The bitwise operators you use are correct.
`~` inverses all the bits. `00110011` becomes `11001100`.
`A &= B;` does two things, it `&` (AND)s A and B (leaving only `1`s where there are `1`s on both sides), then stores the result in A.

Imagine that `gParkFlags` is a 8-bit integer (holding a randomly chose value of `01100101`) and `PARK_FLAGS_SCENARIO_COMPLETE_NAME_INPUT` is 4 (binary `00000100`, B in the example).
```
gParkFlags = 01100101
B  = 00000100 // The flag
~B = 11111011

gParkFlags & ~B = 01100101 & 11111011

01100101 
11111011 &
----------
01100001
     ^
```

The result then gets stored in `gParkFlags`. This essentially removed the flag, since that is the only bit that is 0, as indicated by the `^`.

I hope this helps you understand flags and bitwise operators a _bit_ better.

ocalhoun6:
That _does_ help a lot! (Heh, guess I got lucky copying the right operator from scenario.cpp)

If I'm understanding this right, then `PARK_FLAGS_SCENARIO_COMPLETE_NAME_INPUT` is also an unsigned char with a value of 4 (in binary)? I suppose that's set in one of those header files, right?

Now I just need to figure out how to download, build, and run this branch so I can test it. That will probably have to wait until after the camping trip I'm going on. Thanks to your explanation, I'm now _much_ more confident that I've figured out how to set this flag correctly ... but I still need to test it and find out if that's really the only flag that needs to be changed.

Gymnasiast:
> Just saw the changelog tag ... should I also make an addition to the changelog within this PR?

Yep.

> And is this an original bug? I'm on linux, so I don't have a vanilla installation to test it on.

I just checked in vanilla, the bug is only in OpenRCT2. By the way, I’m on Linux too - RCT2 runs well in Wine.

Gymnasiast (approved):

Gymnasiast:
You probably also need to set `gScenarioCompletedCompanyValue` to MONEY32_UNDEFINED. (Or when https://github.com/OpenRCT2/OpenRCT2/pull/15117 is merged first, MONEY64_UNDEFINED).

ocalhoun6:
I'm pretty swamped this weekend, but I'll try to get this tested on Monday.
And thanks for pointing out that other variable! It might really save the day if this doesn't work properly the first time around.

ocalhoun6:
> You probably also need to set `gScenarioCompletedCompanyValue` to MONEY64_UNDEFINED.

Just finished testing it as-is ... and this doesn't seem to be the case.

Maybe `gScenarioCompletedCompanyValue` is still holding an incorrect value as you play the scenario, but it doesn't seem to mark the scenario as won, and I'm assuming it gets overwritten with the new correct value when you actually win the scenario.

Tested:
- Converting a new (unwon) forest frontiers save to a scenario -- worked correctly as usual
- Winning forest frontiers, saving the game, then converting the save to a scenario -- worked correctly; no more sign of the bug
- Winning dynamite dunes, saving the game, then converting the save to a scenario -- worked correctly; no more sign of the bug

I think I'm going to call this one fixed!

I think I'm supposed to add myself to the contributors file, right? I'll see if I can add that, and then change this from a draft PR to one ready to be merged.

ocalhoun6:
Okay, did the contributors thing ... and it already seems to no longer be a draft?
I guess as soon as the automated checks clear, this thing is ready to go through the process, then.

Gymnasiast:
For good measure, it’s better if you set `gScenarioCompletedCompanyValue` to MONEY64_UNDEFINED anyway.

ocalhoun6:
Okay, added that as well.
Once the checks go through, I'll test it again just to make sure that didn't screw anything up (even though I can't imagine that it would).

src/openrct2/Editor.cpp
```diff
@@ -138,6 +138,10 @@ namespace Editor
         gS6Info.objective_arg_3 = gScenarioObjective.NumGuests;
         climate_reset(gClimate);
 
+        // reset whether the scenario has been won, to make sure it hasn't:
```
Broxzier, line 141:
```suggestion
        // Clear the scenario completion status
```

Gymnasiast (approved):

PR #15072, opened 2021-07-23
Add constants for golf flags

duncanspumpkin:
This grew a few arms and legs along the way and its now a general refactor of golf related code.


src/openrct2/ride/Vehicle.cpp
```diff
@@ -8906,53 +8906,56 @@ void Vehicle::UpdateTrackMotionMiniGolfVehicle(Ride* curRide, rct_ride_entry* ri
                 }
                 track_progress++;
                 break;
-            case 1: // loc_6DC7ED
+            case MiniGolfState::Unk1: // loc_6DC7ED
```
duncanspumpkin, line 8909: This case is never set in the moveinfo so its safe to just assert false it. It also means the var_d3 is never set. It may be that RCT1 saves have a value in D3 though so removing it might cause some super minor difference.

src/openrct2/ride/Vehicle.h
```diff
@@ -376,6 +376,18 @@ struct train_ref
...
+namespace MiniGolfFlag
+{
+    constexpr uint8_t Flag0 = (1 << 0);
+    constexpr uint8_t Flag1 = (1 << 1);
+    constexpr uint8_t Flag2 = (1 << 2);
+    constexpr uint8_t Flag3 = (1 << 3);
+    constexpr uint8_t Flag4 = (1 << 4);
+    constexpr uint8_t Flag5 = (1 << 5); // transitioning between hole
+    constexpr uint8_t Flag6 = (1 << 6);
+    constexpr uint8_t Flag7 = (1 << 7);
```
tupaschoal, line 390: On https://github.com/OpenRCT2/OpenRCT2/pull/15072/commits/08632500183b3473dc3144c9eef77cb2543499b2 these two flags are unused

duncanspumpkin: yes realise that. Added it just in case can be removed now.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8906,51 +8906,54 @@ void Vehicle::UpdateTrackMotionMiniGolfVehicle(Ride* curRide, rct_ride_entry* ri
...
                     }
-                    if (trackPos.z == 6)
+                    if (animation == MiniGolfAnimation::PutLeft)
                     {
-                        auto* curPeep = GetEntity<Guest>(peep[0]);
-                        if (curPeep != nullptr && curPeep->Id & 7)
+                        if (curPeep->Id & 7)
                         {
-                            trackPos.z = 8;
+                            animation = MiniGolfAnimation::Put;
```
pizza2004, line 8943: I believe this should say putt?

src/openrct2/ride/VehicleSubpositionData.cpp
```diff
@@ -25824,7 +25833,7 @@ static constexpr const rct_vehicle_info_list *TrackVehicleInfoListReverserRCRear
 
 // rct2: 0x008B8F30
 constexpr const rct_vehicle_info_list * const * gTrackVehicleInfo[static_cast<uint8_t>(VehicleTrackSubposition::Count)] = {
-    TrackVehicleInfoListDefault,                  // VehicleTrackSubposition::Default
+    TrackVehicleInfoListDefault,                     // VehicleTrackSubposition::Default
```
Gymnasiast, line 25836: stray change

src/openrct2/ride/VehicleSubpositionData.cpp
```diff
@@ -18527,396 +18536,396 @@ CREATE_VEHICLE_INFO(TrackVehicleInfo_9395FF, {
 })
 
 CREATE_VEHICLE_INFO(TrackVehicleInfo_939691, {
-    { -32768,    0,    0,  0,  0,  0 }, {     31,   16,    0,  0,  0,  0 }, {     29,   16,    0,  0,  0,  0 }, {     27,   16,    0,  0,  0,  0 }, { -32768,    4,    1,  0,  0,  0 },
```
Gymnasiast, line 18530: What do the occurrences that were not replaced with either MINI_GOLF_STATE or MINI_GOLF_ANIMATION signify?

duncanspumpkin: They are normal sub position entries that move the vehicle the set distance

Gymnasiast:
Code looks good apart from the single line note. I do want to test it a little bit.

Gymnasiast (approved):

PR #15084, opened 2021-07-24
Add hook for vehicle crashes to plugin api

andrewpratt64:
- Add a new `vehicle.crash` hook to the api
  - Fires whenever a vehicle crashes, i.e. the moment the car explodes and it's in-game reads "Crashed!"
  - Callback function for the hook contains a single argument of type `VehicleCrashArgs`
- `VehicleCrashArgs` contains information about a vehicle crash
  - Property `id` is the numeric id of the car entity of type `Car`
  - Property `crashIntoType` is of type `VehicleCrashIntoType`
- `VehicleCrashIntoType` describes what the vehicle crashed into before exploding
  - A value of `"another_vehicle"` means the car collided with another car on the track
  - A value of `"land"` means the vehicle crashed into solid terrain
  - A value of `"water"` means the vehicle landed in water

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -24,6 +24,8 @@
 #include "../platform/platform.h"
 #include "../rct12/RCT12.h"
 #include "../scenario/Scenario.h"
+#include "../scripting/HookEngine.h"
+#include "../scripting/ScriptEngine.h"
 #include "../util/Util.h"
 #include "../windows/Intent.h"
 #include "../world/Map.h"
@@ -3598,6 +3600,23 @@ void Vehicle::UpdateCollisionSetup()
 
         train->sub_state = 2;
 
+#ifdef ENABLE_SCRIPTING
+        auto& hookEngine = OpenRCT2::GetContext()->GetScriptEngine().GetHookEngine();
+        if (hookEngine.HasSubscriptions(OpenRCT2::Scripting::HOOK_TYPE::VEHICLE_CRASH))
+        {
+            auto ctx = OpenRCT2::GetContext()->GetScriptEngine().GetContext();
+
+            // Create event args object
+            auto obj = OpenRCT2::Scripting::DukObject(ctx);
+            obj.Set("id", train->sprite_index);
+            obj.Set("crashIntoType", "another_vehicle");
+
+            // Call the subscriptions
+            auto e = obj.Take();
+            hookEngine.Call(OpenRCT2::Scripting::HOOK_TYPE::VEHICLE_CRASH, e, true);
+        }
+#endif
+
         OpenRCT2::Audio::Play3D(OpenRCT2::Audio::SoundId::Crash, { train->x, train->y, train->z });
 
         ExplosionCloud::Create({ train->x, train->y, train->z });
@@ -5334,6 +5353,23 @@ void Vehicle::CrashOnLand()
     }
     SetState(Vehicle::Status::Crashed, sub_state);
 
+#ifdef ENABLE_SCRIPTING
+    auto& hookEngine = OpenRCT2::GetContext()->GetScriptEngine().GetHookEngine();
+    if (hookEngine.HasSubscriptions(OpenRCT2::Scripting::HOOK_TYPE::VEHICLE_CRASH))
+    {
+        auto ctx = OpenRCT2::GetContext()->GetScriptEngine().GetContext();
+
+        // Create event args object
+        auto obj = OpenRCT2::Scripting::DukObject(ctx);
+        obj.Set("id", sprite_index);
+        obj.Set("crashIntoType", "land");
+
+        // Call the subscriptions
+        auto e = obj.Take();
+        hookEngine.Call(OpenRCT2::Scripting::HOOK_TYPE::VEHICLE_CRASH, e, true);
+    }
+#endif
+
     if (!(curRide->lifecycle_flags & RIDE_LIFECYCLE_CRASHED))
     {
         auto frontVehicle = GetHead();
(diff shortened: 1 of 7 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -3598,6 +3600,23 @@ void Vehicle::UpdateCollisionSetup()
 
         train->sub_state = 2;
 
+#ifdef ENABLE_SCRIPTING
```
duncanspumpkin, line 3603: all of the other crash hooks only do the train head. Why for this one do you want all of the train cars?

src/openrct2/scripting/HookEngine.cpp
```diff
@@ -31,6 +31,7 @@ HOOK_TYPE OpenRCT2::Scripting::GetHookType(const std::string& name)
         { "ride.ratings.calculate", HOOK_TYPE::RIDE_RATINGS_CALCULATE },
         { "action.location", HOOK_TYPE::ACTION_LOCATION },
         { "guest.generation", HOOK_TYPE::GUEST_GENERATION },
+        { "vehicle.crash", HOOK_TYPE::VEHICLE_CRASH},
```
IntelOrca, line 34:
```suggestion
        { "vehicle.crash", HOOK_TYPE::VEHICLE_CRASH },
```

src/openrct2/ride/Vehicle.cpp
andrewpratt64: Unless if I'm misunderstanding how the Vehicle class works, Vehicle::CrashOnLand() and Vehicle::CrashOnWater() is called for each individual car in a train, since the cars can hit the ground at different times. Vehicle::UpdateCollisionSetup() appears to be called for only the car that directly collided with a car from another train, hence the for loop that goes over every car. All three hook calls should fire for all cars in a train, not just the head.

IntelOrca (approved):

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: Oh yeah good point

duncanspumpkin (approved):
This all looks fine but couldn't we turn this into a function call so that you don't repeat code as much

andrewpratt64:
> This all looks fine but couldn't we turn this into a function call so that you don't repeat code as much

I could add a method to Vehicle that takes a sting or an enumerated type for the crashedInto property as an argument

duncanspumpkin:
Just make some static local function that takes the crash type and the vehicle id. That would be fine.
E.g.
```
static void VehicleCrashHook(uint16_t vehicleId, std::string_view crashType)
```

src/openrct2/ride/Vehicle.cpp
```diff
@@ -740,6 +740,34 @@ template<> bool SpriteBase::Is<Vehicle>() const
     return Type == EntityType::Vehicle;
 }
 
+/**
+ * Fires the "vehicle.crash" api hook
+ * @param vehicleId Entity id of the vehicle that just crashed
+ * @param crashId What the vehicle crashed into. Should be either "another_vehicle", "land", or "water"
+ */
+static void FireVehicleCrashHook(const uint16_t vehicleId, const std::string_view crashId)
```
ZehMatt, line 748:
```suggestion
static void InvokeVehicleCrashHook(const uint16_t vehicleId, const std::string_view crashId)
```
Sounds a bit better

src/openrct2/ride/Vehicle.cpp
```diff
@@ -738,6 +740,34 @@ template<> bool SpriteBase::Is<Vehicle>() const
...
+ * Fires the "vehicle.crash" api hook
+ * @param vehicleId Entity id of the vehicle that just crashed
+ * @param crashId What the vehicle crashed into. Should be either "another_vehicle", "land", or "water"
+ */
+static void InvokeVehicleCrashHook(const uint16_t vehicleId, const std::string_view crashId)
+{
+    // If ENABLE_SCRIPTING is not defined, the body of this function is empty.
+    // Calls to this function should probably still be wrapped in a #ifdef to avoid
+    //  unoptimized code from wasting time by calling an empty function
+#ifdef ENABLE_SCRIPTING
```
ZehMatt, line 753: You need to wrap the entire function or you'll get warnings about unused functions when scripting is disabled.

IntelOrca:
Could you increment the scripting API version and update the changelog?

andrewpratt64:
> Could you increment the scripting API version and update the changelog?

Done, prefixed the description in the changelog with [Plugin], I'm assuming it's used for plugin-related changes?

IntelOrca:
@andrewpratt64 plugins can check the API version to see if a feature or compatibility exists or not.

ZehMatt:
We normally rebase the branch instead of merging, could you please correct that?

andrewpratt64:
> We normally rebase the branch instead of merging, could you please correct that?

Okay, I can do that

andrewpratt64:
image

So I got it up to this point where it dosen't have the commits that were just for merging from the upstream, but I'm not really sure what the next step is. I haven't done much with rebasing before lol, I've only ever merged

tupaschoal:
We have a tutorial showing how to rebase and sync with upstream, including how to solve merges:
https://github.com/OpenRCT2/OpenRCT2/wiki/Rebase-and-Sync-fork-with-OpenRCT2

Do try it and let us know if you face difficulties!

IntelOrca:
@andrewpratt64 your rebase was successful, you don't need to do anything else.

IntelOrca:
@andrewpratt64 you need to rebase off the latest version of develop from the main upstream repository to deal with the changelog conflict.

andrewpratt64:
@tupaschoal @IntelOrca Okay, I think I got it now. The tutorial definitely made it easier

src/openrct2/scripting/HookEngine.cpp
```diff
@@ -30,6 +30,7 @@ static const EnumMap<HOOK_TYPE> HooksLookupTable({
     { "ride.ratings.calculate", HOOK_TYPE::RIDE_RATINGS_CALCULATE },
     { "action.location", HOOK_TYPE::ACTION_LOCATION },
     { "guest.generation", HOOK_TYPE::GUEST_GENERATION },
+    { "vehicle.crash", HOOK_TYPE::VEHICLE_CRASH},
```
tupaschoal, line 33:
```suggestion
    { "vehicle.crash", HOOK_TYPE::VEHICLE_CRASH },
```

tupaschoal:
I'm glad it helped! I just fixed a minor clang format issue and if all is well it should be good  to go :)

PR #15101, opened 2021-07-25
Dump crash dumps into a different directory and include version in reports

ZehMatt:


src/openrct2/platform/Crash.cpp
```diff
@@ -320,7 +322,11 @@ static std::wstring GetDumpDirectory()
 {
     char userDirectory[MAX_PATH];
     platform_get_user_directory(userDirectory, nullptr, sizeof(userDirectory));
-    auto result = String::ToWideChar(userDirectory);
+
+    auto path = Path::Combine(userDirectory, "crashes");
```
IntelOrca, line 326: Our directory names are usually singular.

ZehMatt: Not quite but I changed.

src/openrct2/platform/Crash.cpp
```diff
@@ -320,7 +322,11 @@ static std::wstring GetDumpDirectory()
 {
     char userDirectory[MAX_PATH];
     platform_get_user_directory(userDirectory, nullptr, sizeof(userDirectory));
-    auto result = String::ToWideChar(userDirectory);
+
+    auto path = Path::Combine(userDirectory, "crash");
```
Gymnasiast, line 326: We have enums for these types of directory.

Called like so: env->GetDirectoryPath(OpenRCT2::DIRBASE::OPENRCT2, OpenRCT2::DIRID::CRASH);

Gymnasiast (approved):
All my points have been addressed. :+1: 
Leaving it to @IntelOrca now.

Broxzier (approved):

src/openrct2/platform/Crash.cpp
```diff
@@ -318,9 +320,10 @@ static bool OnCrash(
 
 static std::wstring GetDumpDirectory()
 {
-    char userDirectory[MAX_PATH];
-    platform_get_user_directory(userDirectory, nullptr, sizeof(userDirectory));
-    auto result = String::ToWideChar(userDirectory);
+    auto env = GetContext()->GetPlatformEnvironment();
```
Gymnasiast, line 323: I think this should read

```suggestion
    auto env = OpenRCT2::GetContext()->GetPlatformEnvironment();
```

IntelOrca (approved):

src/openrct2/platform/Crash.cpp
ZehMatt: Oddly everything except MSVC built this, on my end it also did but yeah.

Issue #15006, opened 2021-07-13
Game crash scrolling through Misc scenery


rctcool0:
**OS:** Windows 10 (x86-64)
**Version:** V0.3.3
**Commit/Build:** 999daa9 on new-save-format

A crash occurs when scrolling through a long list of scenery. Specifically, it looks like the maximum image limit in Image.cpp does not match the new limits on the new save file format:
"ERROR[D:\a\OpenRCT2\OpenRCT2\src\openrct2\drawing\Image.cpp:200 (gfx_object_allocate_images)]: Reached maximum image limit."
[log.txt](https://github.com/OpenRCT2/OpenRCT2/files/6812496/log.txt)

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Specific to multiplayer?

**Steps to reproduce:**
1. Open the Gigabench.park file.
2. Open Scenery tab, click on the "?", and scroll down until the game crashes.

**Screenshots / Video:**
N/A, nothing abnormal visually.

**Save game:**
https://1drv.ms/u/s!AiQ_uynOpJhHhu4LauA1guhy1RlAMw?e=2vnAeD
Fair warning, park is 100mb and includes a MASSIVE amount of scenery.

Gymnasiast:
The error happens in this line in TextureCache.cpp:
```C++
    return _atlases.back().Allocate(imageWidth, imageHeight);
```
It calls Allocate, but there are no free slots available.

PR #15109, opened 2021-07-27
Map legacy paths in old .PARK saves

Gymnasiast:
This allows loading up older .PARK saves, created before splitting paths.
This includes saves like the one in https://github.com/OpenRCT2/OpenRCT2/issues/15006 and some of @IntelOrca ’s test saves.

### DIFF
--- a/OpenRCT2.xcodeproj/project.pbxproj
+++ b/OpenRCT2.xcodeproj/project.pbxproj
@@ -4341,7 +4341,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CLANG_ENABLE_OBJC_WEAK = YES;
-				MACOSX_DEPLOYMENT_TARGET = 10.13;
+				MACOSX_DEPLOYMENT_TARGET = 10.14;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 			};
 			name = Debug;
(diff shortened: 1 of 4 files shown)


ZehMatt:
Does this only fix loading older .park files? I would argue that we all knew those files would be quite volatile and given that we no longer export those old files its a bit unnecessary in my opinion. One could write a small tool to fix the save file maybe?

Gymnasiast:
> Does this only fix loading older .park files? I would argue that we all knew those files would be quite volatile and given that we no longer export those old files its a bit unnecessary in my opinion. One could write a small tool to fix the save file maybe?

This has been extensively discussed before, and the outcome was that we will at least support "development" saves temporarily, providing an upgrade path (if needed, through multiple version).

IntelOrca:
Would it be better to check if the park format version is older than (version number) and only run this mapping code if it is?

Gymnasiast:
@IntelOrca I don’t think it wouldn’t make a great deal of difference, as it would only match on old names for legacy paths. And it would also provide easy hooks for replacing more legacy paths in the future.

IntelOrca:
@Gymnasiast think about the performance, you are adding checks for every footpath tile element on the map

Gymnasiast:
Good point. I’ll add version checks.

Gymnasiast:
Oddly enough, the save in https://github.com/OpenRCT2/OpenRCT2/issues/15006 has version `2`, which is much higher than expected. Ah well, I’ll limit it to `<= 2`, then.

PR #15112, opened 2021-07-27
Map size refactors from NSF

duncanspumpkin:
Another round of cherry picks from the NSF.


src/openrct2/world/Map.h
```diff
@@ -101,11 +101,21 @@ extern const TileCoordsXY TileDirectionDelta[];
...
+extern int32_t gMapSize;
+extern int32_t gMapBaseZ;
+
+inline int32_t GetMapSizeUnits()
+{
+    return (gMapSize - 1) * COORDS_XY_STEP;
+}
+inline int32_t GetMapSizeMinus2()
+{
+    return (gMapSize * COORDS_XY_STEP) + MAXIMUM_MAP_SIZE_PRACTICAL;
```
Gymnasiast, line 113: `MAXIMUM_MAP_SIZE_PRACTICAL` is the maximum map size in small coordinates, e.g. 254 for the SV6 format. Adding that to a number that is in big coordinates makes no sense.

duncanspumpkin: Yeah, looks odd seems to be what the value was set to in the map window. Will need to investigate why before changing it to the more logical -2

duncanspumpkin: so i looked into it and the value looks to be correct (at least that is what the assembly had). Using that constant name is perhaps incorrect so i've changed it to `(8 * COORDS_XY_STEP - 2)`. I don't know how it works enough to know if that is sensible.

duncanspumpkin:
There seems to be an issue with drawing banners at certain locations.

src/openrct2/world/Map.h
Gymnasiast: Hm, do we have any idea _why_ it is that value?

duncanspumpkin: No, I've looked at it for a while and can't draw any sensible conclusion.

ZehMatt (approved):
Tested this and code looks good

duncanspumpkin:
Banner issue seems to be fixed. Possibly just a heisenbug from switching branchs.

PR #15113, opened 2021-07-27
Fix path finding for larger maps

ZehMatt:
The current structure is capped to uint8_t so the biggest tile position would be 0xFE, 0xFE before it considers it invalid. Probably breaks replays as the null value for TileCoords is 0xFFFF8000 if I remember right.


Gymnasiast (approved):
Would like another teammate to look at it as well.

duncanspumpkin (approved):

PR #15123, opened 2021-07-28
Fix #14316: Closing the Track Designs Manager window causes broken state

### LINKED ISSUE #14316: Track Design Manager window can be closed to allow for glitchy land modifications in the background map
**OS:** Windows 10
**Version:** 0.3.3
**Commit/Build:** 7793f4c

**Steps to reproduce:**
1. Open the Track Design Manager (not the Track Designer) under the Toolbox.
2. Exit out of the window.
3. Do weird things like land modifications on the background map.
4. There's also a button for paths, but it refuses to open any path window.

**Screenshots / Video:**
Unnamed park 2021-03-14 13-57-45
Unnamed park 2021-03-14 13-55-41

seifer7:
Fix #14316: Quiting track design manager broke editor
When exiting the Track Design Manager window the user was placed into
the scenario editor in a broken state.
User is now returned to the main menu when the window is closed.

PS. First contribution, hope it's right :)

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,8 +1,8 @@
 0.3.4.1+ (in development)
 ------------------------------------------------------------------------
+- Fix: [#14316] Closing the Track Designs Manager window causes broken state.
 - Fix: [#15096] Crash when placing entrances in the scenario editor near the map corner.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
-- Fix: [#14316] Closing the Track Designs Manager window causes broken state.
 
 0.3.4.1 (2021-07-25)
 ------------------------------------------------------------------------
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#15096] Crash when placing entrances in the scenario editor near the map corner.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
+- Fix: [#14316] Closing the Track Designs Manager window causes broken state.
 
 0.3.4.1 (2021-07-25)
 ------------------------------------------------------------------------
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -456,6 +456,11 @@ static void window_editor_object_selection_mouseup(rct_window* w, rct_widgetinde
             {
                 finish_object_selection();
             }
+            if (gScreenFlags & SCREEN_FLAGS_TRACK_MANAGER)
+            {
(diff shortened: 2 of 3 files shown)


src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -456,6 +456,11 @@ static void window_editor_object_selection_mouseup(rct_window* w, rct_widgetinde
             {
                 finish_object_selection();
             }
+            if (gScreenFlags & SCREEN_FLAGS_TRACK_MANAGER)
+            {
+                auto loadOrQuitAction = LoadOrQuitAction(LoadOrQuitModes::OpenSavePrompt, PromptMode::SaveBeforeQuit);
```
Gymnasiast, line 461: This is a purely single player affair, so there is no need to do it through a game action. Additionally, there is nothing to save, so it should just exit to the main menu.

Broxzier (changes requested):
Thanks for contributing! I tested this, and it works, but check out @Gymnasiast's remarks.

Please add a [changelog entry](https://github.com/OpenRCT2/OpenRCT2/blob/develop/distribution/changelog.txt) for this fix, and don't forget to add yourself to the [contributors.md](https://github.com/OpenRCT2/OpenRCT2/blob/develop/contributors.md) file.

src/openrct2-ui/windows/EditorObjectSelection.cpp
seifer7: I am very new to C++, but I will get better 🤞 
I think I've done it correctly now.

distribution/changelog.txt
```diff
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Fix: [#15096] Crash when placing entrances in the scenario editor near the map corner.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
+- Fix: [#14316] Closing the Track Designs Manager window causes broken state.
```
Gymnasiast, line 5: We group the changelog by type, then by issue number. Hence, your entry should be at the top.

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -456,6 +457,10 @@ static void window_editor_object_selection_mouseup(rct_window* w, rct_widgetinde
             {
                 finish_object_selection();
             }
+            if (gScreenFlags & SCREEN_FLAGS_TRACK_MANAGER)
+            {
+                title_load();
```
Gymnasiast, line 462: For good measure:

```suggestion
                game_unload_scripts();
                title_load();
```

Gymnasiast (approved):

PR #15124, opened 2021-07-29
Small network fixes

ZehMatt:


src/openrct2/network/NetworkBase.cpp
```diff
@@ -1698,15 +1698,23 @@ bool NetworkBase::ProcessConnection(NetworkConnection& connection)
 void NetworkBase::ProcessPacket(NetworkConnection& connection, NetworkPacket& packet)
 {
     const auto& handlerList = GetMode() == NETWORK_MODE_SERVER ? server_command_handlers : client_command_handlers;
-    auto it = handlerList.find(packet.GetCommand());
-    if (it != handlerList.end())
+
+    try
```
tupaschoal, line 1708: Sounds like the scope of the `try` block can be made smaller, as I'm assuming the `find` and the `if`s there wouldn't really throw.

src/openrct2/network/NetworkBase.cpp
```diff
@@ -49,6 +49,9 @@ static int32_t _pickup_peep_old_x = LOCATION_NULL;
 // with uint16_t and needs some spare room for other data in the packet.
 static constexpr uint32_t CHUNK_SIZE = 1024 * 63;
 
+// If data is sent fast enough it would halt the entire server, process only a maximum amount.
+static constexpr uint32_t MaxPacketsPerUpdate = 100;
```
tupaschoal, line 53: I'm assuming 100 is an educated guess, but maybe this could use more commenting on the magic number? Like how many packets does an average update have?

ZehMatt: The average update has maybe 3 to 10 packets at most, this is per connection. So 100 is a pretty high tolerance, I kept it high to not accidentally make the network game slower.

tupaschoal: Maybe it's worth adding those comments there to document why 100 was chosen.

One other suggestion is to maybe `log_verbose` or print a warning if the packets are reached. This would make the debugging easier in case anyone faithfully hits this limit in the future.

ZehMatt: There are two things that came out by fuzzing the game action network packet. The logging can cause a huge performance impact when being hammered. The number 100 was picked for when the server is getting hammered by lets say a fuzzer it was the threshold on my machine to keep the game somewhat stable and allow it to process enough packets so the fuzzer wouldn't start buffering output on the socket stream. So I'm so sure how to go about commenting this.

ZehMatt: Well I hope this will do.

tupaschoal: I think it works. What about the other part?

> One other suggestion is to maybe `log_verbose` or print a warning if the packets are reached. This would make the debugging easier in case anyone faithfully hits this limit in the future.

ZehMatt: I don't think it will matter much, its just shifted to the next update/tick.

tupaschoal (approved):

PR #15129, opened 2021-07-30
Minor cleanup

ZehMatt:


src/openrct2-ui/windows/Sign.cpp
```diff
@@ -156,6 +156,7 @@ class SignWindow final : public Window
                 if (tileElement == nullptr)
                 {
                     Close();
+                    return;
```
Broxzier, line 159: Nice catch. This would've always caused crashes before this change, since `tileElement` gets used below.

Broxzier (approved):

PR #15131, opened 2021-07-30
Close #10624: Display GUI message when changing Scale Factor by hotkey

### LINKED ISSUE #10624: Pixel scaler prompt
Just an idea for a small enhancement: now that we can address the pixel scaler via keyboard shortcuts, you no longer see what factor is set when using keys. Thus I was thinking it'd be nice to inform the user via a text prompt much like the fps counter, just one that immediately fades out in ~2 seconds.

"x1"

"x1.25"

"x1.5"

...in the bottom middle of the screen, or on top of the RCT1 style middle menu if people have that enabled.

seifer7:
- When changing the Window Scale Factor using a hotkey a GUI message is now displayed specifying the new scale factor using the Chat window.
- To ensure message is always visible, chat window is now drawn in single player games, title menu, and anywhere else.
- Chat window still cannot be manually toggle/brought up outside a mutiplayer game (intended.)

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3651,6 +3651,7 @@ STR_6450    :{BLACK}“{STRING}”
 STR_6451    :{BLACK}“{STRING}” - {STRING}
 STR_6452    :{WINDOW_COLOUR_2}Sells: {BLACK}{STRING}
 STR_6453    :Copy version info
+STR_6454    :Scale factor set to: {COMMA2DP32}
 
 #############
 # Scenarios #
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,5 +1,6 @@
 0.3.4.1+ (in development)
 ------------------------------------------------------------------------
+- Feature: [#10624] A GUI message is now displayed when changing Window Scale Factor by hotkey.
 - Fix: [#14316] Closing the Track Designs Manager window causes broken state.
 - Fix: [#15096] Crash when placing entrances in the scenario editor near the map corner.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
--- a/src/openrct2-ui/input/Shortcuts.cpp
+++ b/src/openrct2-ui/input/Shortcuts.cpp
@@ -470,6 +470,9 @@ static void ShortcutScaleUp()
     gfx_invalidate_screen();
     context_trigger_resize();
     context_update_cursor_scale();
+    int32_t scale = static_cast<int32_t>(gConfigGeneral.window_scale * 100);
+    std::string scalestring = format_string(STR_UI_SCALING_SET, &scale);
+    chat_history_add(scalestring.c_str());
 }
 
 static void ShortcutScaleDown()
@@ -480,6 +483,9 @@ static void ShortcutScaleDown()
     gfx_invalidate_screen();
     context_trigger_resize();
     context_update_cursor_scale();
+    int32_t scale = static_cast<int32_t>(gConfigGeneral.window_scale * 100);
+    std::string scalestring = format_string(STR_UI_SCALING_SET, &scale);
+    chat_history_add(scalestring.c_str());
 }
 
 // Tile inspector shortcuts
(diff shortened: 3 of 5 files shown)


AaronVanGeffen:
When #10624 was opened, the scaling feature was more limited than it is now. These days, it is possible to change scaling using keyboard shortcuts, for example.

I see these shortcuts used quite often by YouTubers streamers and such. These chat messages will get in the way of their videos. 

For these reasons, I am worried that these feedback messages will do more harm than good. It should be optional at least. But that begs the question, do we really want this to begin with?

seifer7:
> When #10624 was opened, the scaling feature was more limited than it is now. These days, it is possible to change scaling using keyboard shortcuts, for example.

Changing scaling using keyboard shortcuts is exactly what this change was targeting. If you're changing via the Options window you can already see the selected scale factor as you change it.

> I see these shortcuts used quite often by YouTubers streamers and such. These chat messages will get in the way of their videos.
> For these reasons, I am worried that these feedback messages will do more harm than good. It should be optional at least. 

I have just changed it so that the notification message doesn't appear if the Toolbars have been toggled off.
So, if they are in some kind of "cinematic" view, they'd likely want the Toolbars hidden, at which time the messages won't appear when they are changing scale factor by hotkey.
What do you think of that?

> But that begs the question, do we really want this to begin with?

It would make sense to me to give the user some feedback of what they've changed by hotkey.

Broxzier:
The interface scale up and down shortcut keys don't have any keys assigned by default, so the chance someone accidentally presses these without knowing about the feature is slim. The feedback they receive is seeing the game view getting larger and smaller already, so I don't think a message is really necessary.

seifer7:
> The interface scale up and down shortcut keys don't have any keys assigned by default, so the chance someone accidentally presses these without knowing about the feature is slim. The feedback they receive is seeing the game view getting larger and smaller already, so I don't think a message is really necessary.

Okay. I just saw the Issue still open and figured it was something I could tackle.
If it's no longer wanted, I guess you can close it.

Broxzier:
I may be in the minority with this opinion here. Perhaps the majority actually likes this. The original issue suggests some simple text like the FPS counter at the top of the screen that disappears quickly, which I think is nicer than having the chat pop up and partly obscure the view.

seifer7:
> It would make sense to me to give the user some feedback of what they've changed by hotkey.

In the same vein, I noticed that "Screenshot" provides no feedback at all, and "Giant Screenshot" provides feedback by the way of a red/error mouse tooltip.
I'm of the opinion that they could both use the same "Chat" system to give the user feedback that they have taken a screenshot and what the file is called.
_Also, side-note: You can't give Giant Screenshot a keybind._

IntelOrca:
> I'm of the opinion that they could both use the same "Chat" system to give the user feedback that they have taken a screenshot and what the file is called.

We / vanilla did show an "error" box to say that a screenshot had taken. But the issue was that if you take another screenshot in quick succession, that box would be in the screenshot. The same problem would apply if you displayed text in the chat window.

seifer7:
> > I'm of the opinion that they could both use the same "Chat" system to give the user feedback that they have taken a screenshot and what the file is called.
> 
> We / vanilla did show an "error" box to say that a screenshot had taken. But the issue was that if you take another screenshot in quick succession, that box would be in the screenshot. The same problem would apply if you displayed text in the chat window.

This should probably have its own thread, but, if we utilise the Chat box for system messages we could have the Chat hide during the screenshot process and then reappear. Unless the chat input is open in which case it stays visible for the screenshot.

IntelOrca:
> This should probably have its own thread, but, if we utilise the Chat box for system messages we could have the Chat hide during the screenshot process and then reappear. Unless the chat input is open in which case it stays visible for the screenshot.

Yes, that is a possibility, although it may still catch a user out if they want to quickly screenshot a recent chat message and don't realise they need to open the chat window.

seifer7:
> Yes, that is a possibility, although it may still catch a user out if they want to quickly screenshot a recent chat message and don't realise they need to open the chat window.

Yeah, you're right. A screenshot should represent what is on the screen or it'd be misleading and probably cause confusion.

Richard-L:
My initial concern was that it's important to know if you're at a full integer scale like x2 or x3 or not, as it's very easy to get lost when stepping in .25 increments. @seifer7 just a general thank you for looking into this, in any case. The toolbar on/off toggle for this seems perfect to me in regards to the Youtuber concern.

PR #15133, opened 2021-07-30
Fix #14918 refactor while loop and int variable size

### LINKED ISSUE #14918: Refactor use of overflow in Ride.cpp
This code relies on an overflow:
https://github.com/OpenRCT2/OpenRCT2/blob/9383fcef410aa6a94e3ad07a8da4a2eba41fbdb8/src/openrct2-ui/windows/Ride.cpp#L1603-L1612

There are multiple occurrences like this, all of them should be changed.

This is prone to errors when things like either type or max value change.

SaadRehmanCS:


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -1601,15 +1601,15 @@ static void window_ride_init_viewport(rct_window* w)
     else if (eax >= ride->num_vehicles && eax < (ride->num_vehicles + ride->num_stations))
     {
         StationIndex stationIndex = STATION_INDEX_NULL;
-        int32_t count = eax - ride->num_vehicles;
-        do
+        int count = eax - ride->num_vehicles;
```
duncanspumpkin, line 1604: we used the fixed width types in this codebase (or auto).

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -1601,15 +1601,15 @@ static void window_ride_init_viewport(rct_window* w)
...
+        while (count >= 0)
```
duncanspumpkin, line 1605: Rename eax to `viewSelectionIndex`. The overflow that we are trying to avoid is `StationIndex stationIndex = STATION_INDEX_NULL + 1` instead you should make it something like this:
```
for (const auto& station : ride->stations){
    if (station.Start.isNull()){
        continue;
    }
    count--;
    if (count == 0){
       // L1614-1618 auto location = station.GetStart(); ....
    }
}
```

duncanspumpkin:
Also note this issue was just one example of likely multiple in the codebase so need to remember to not close this issue at the end of this PR.

SaadRehmanCS:
I've just implemented the listed changes, let me know if any other changes need to be made!

duncanspumpkin (approved):
All looks good. Have a look and see if you can find more instances where a variable is set to STATION_INDEX_NULL and is then incremented there will probably be some other ones in the codebase. Feel free to open new PRs for them.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -1598,28 +1598,30 @@ static void window_ride_init_viewport(rct_window* w)
...
         }
     }
-    else if (eax >= ride->num_vehicles && eax < (ride->num_vehicles + ride->num_stations))
+    else if (viewSelectionIndex >= ride->num_vehicles && viewSelectionIndex < (ride->num_vehicles + ride->num_stations))
     {
-        StationIndex stationIndex = STATION_INDEX_NULL;
-        int32_t count = eax - ride->num_vehicles;
-        do
+        int32_t count = viewSelectionIndex - ride->num_vehicles;
+
```
duncanspumpkin, line 1604: clang format isn't happy with the whitespace here i think you have a number of spaces either delete the spaces or remove the line completely to satisfy clang format (or even better run clang format locally its built into Visual Studio if that is what you are using)

ZehMatt:
Since this is your first PR you might want to add yourself to ```contributors.md```

ZehMatt (approved):
Thanks!

PR #15135, opened 2021-07-30
Fix #8601

### LINKED ISSUE #8601: OpenRCT2 draws supports in places where vanilla wouldn't
So I noticed while looking at some older parks that a coaster's supports where showing, where they should be blocked. I don't think this has always been a problem with OpenRCT. I guess it has something to do with the drawing order on those tiles? 

I took two screens of the exact same thing in the exact (pixel perfect) location. A high quality gif alternating between the two would make the differences clear, but I don't know how to do that, so I'll just upload the two .pngs I have, and you have to do the alternating yourself if you download them...

Note the flying coaster supports, but also the tram supports.

scr1
varmland park - dimi 2008-2011 2019-01-18 16-25-13

The park in question: [Dimi - Värmland Park.zip](https://github.com/OpenRCT2/OpenRCT2/files/2773542/Dimi.-.Varmland.Park.zip)
https://www.nedesigns.com/park/1869/vaermland-park/

Using v0.2.1-600 (92412d4) on Windows 8

duncanspumpkin:


AaronVanGeffen:
Out of curiosity, why does applying that flag to those particular three objects affect the supports drawing?

Gymnasiast (changes requested):
This is not a bug. All vanilla base blocks allow supports on top, and so I fixed ToonTowner’s base blocks to also allow them. The bug is that TT forgot to set that flag on his objects in the first place.

duncanspumpkin:
The community clearly wants this fix reverted.

duncanspumpkin:
> Out of curiosity, why does applying that flag to those particular three objects affect the supports drawing?

People use those particular objects as support blockers. This is the flag that controls blocking supports.

Gymnasiast:
> The community clearly wants this fix reverted.

I know of exactly one user who has asked us for this. On the original issue, **nobody** else has chimed in. That’s a far cry from “the community” “clearly” wanting this fix reverted.

> People use those particular objects as support blockers. This is the flag that controls blocking supports.

Which is an ill-advised thing to do. You can’t use the vanilla base blocks as support blockers, so anyone who uses a custom base block as one is just asking for trouble.

src/openrct2/object/SmallSceneryObject.cpp
```diff
@@ -158,14 +158,12 @@ void SmallSceneryObject::PerformFixes()
     auto identifier = GetLegacyIdentifier();
     static const auto& scgWalls = Object::GetScgWallsHeader();
 
-    // ToonTowner's base blocks. Make them allow supports on top and put them in the Walls and Roofs group.
+    // ToonTowner's base blocks. Make put them in the Walls and Roofs group.
```
Gymnasiast, line 161:
```suggestion
    // ToonTowner's base blocks. Put them in the Walls and Roofs group.
```

PR #15144, opened 2021-07-31
Refactor/track design stuff

Gymnasiast:


src/openrct2/ride/TrackDesign.cpp
```diff
@@ -977,23 +976,24 @@ static bool TrackDesignPlaceSceneryElementGetPlaceZ(const TrackDesignSceneryElem
     return true;
 }
 
-static bool TrackDesignPlaceSceneryElement(
-    CoordsXY mapCoord, uint8_t mode, const TrackDesignSceneryElement& scenery, uint8_t rotation, int32_t originZ)
+static std::pair<bool, money32> TrackDesignPlaceSceneryElement(
```
ZehMatt, line 979: I think std::optional makes more sense

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -977,23 +976,30 @@ static bool TrackDesignPlaceSceneryElementGetPlaceZ(const TrackDesignSceneryElem
     return true;
 }
 
-static bool TrackDesignPlaceSceneryElement(
-    CoordsXY mapCoord, uint8_t mode, const TrackDesignSceneryElement& scenery, uint8_t rotation, int32_t originZ)
+static std::optional<money32> TrackDesignPlaceSceneryElement(
```
duncanspumpkin, line 979: I feel this is pretty odd way of doing these functions. It would be more logical that the std::optional<money32> is just the result of placing this singular piece of scenery. Move the actual addition of totalCost to the caller. Then we can just ignore everything to do with `MONEY32_UNDEFINED`. The other alternative is to not use std::optional and only return money32 and use MONEY32_UNDEFINED to indicate failure.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1234,40 +1240,40 @@ static bool TrackDesignPlaceSceneryElement(
...
+    return totalCost;
 }
 
 /**
  *
  *  rct2: 0x006D0964
  */
-static int32_t track_design_place_all_scenery(
-    const std::vector<TrackDesignSceneryElement>& sceneryList, const CoordsXYZ& origin)
+static std::optional<money32> track_design_place_all_scenery(
```
duncanspumpkin, line 1275: Same would go for this function. Remove the passing in of the track cost (You have inadvertently implemented as per my suggestion as L1305 is meant to be `+=`)

duncanspumpkin (changes requested):
We should be trying to remove MONEY32_UNDEFINED and this is a good start but I think the design isnt great. I would do all the additions in the callers of the functions.

duncanspumpkin (approved):
This looks much better. Not tested only reviewed the code.

PR #15147, opened 2021-08-01
Close #15143: Added a shortcut key for Giant Screenshot.

### LINKED ISSUE #15143: Can't set Shortcut Key for "Giant Screenshot"
**OS:** Windows 11
**Version:** 0.3.4.1
**Commit/Build:** dbdca55645632cbb6afb3ff6979e5534e808c738

The Shortcut Keys window does not have an option to set a key for "Giant Screenshot"

seifer7:
Adds the ability to set a shortcut key to take Giant Screenshots.


data/language/en-GB.txt
```diff
@@ -1861,6 +1861,7 @@ STR_2521    :Show staff list
 STR_2522    :Show recent messages
 STR_2523    :Show map
 STR_2524    :Screenshot
+STR_2525    :Giant Screenshot
```
Gymnasiast, line 1864: Always add new strings at the end. We don’t recycle IDs.

seifer7: Okay. Will push another commit.

Gymnasiast (approved):

Gymnasiast:
Since you added a string, please also open an issue over at https://github.com/OpenRCT2/Localisation

tupaschoal:
I wonder if we needed a new string for something that is exactly the same as STR_5260?

IntelOrca:
@tupaschoal sometimes it is better to have different IDs in case a language wants to use a different translation depending on the context.

tupaschoal:
Makes sense, thanks

PR #15150, opened 2021-08-01
Fix #15148: Track Designs Manager delete confirmation window doesn't display properly

### LINKED ISSUE #15148: Track Designs Manager delete confirmation window doesn't display properly
**OS:** [e.g. Windows 10]
**Version:** 0.3.4.1
**Commit/Build:** 5087e77

The "Delete Confirmation" window in the Track Designs Manager doesn't display properly.
The "Delete Confirmation" window title is the name of the Ride, whereas in the RCT2 the title is "Delete File"
The "Delete Confirmation" window title goes blank if you close the previous window asking if you want to Delete or Rename the ride. In ORCT2 the title stays as "Delete File"

I think the "Delete Confirmation" window title should be "Delete File" (like RCT2,) and when you close the previous window I think the "Delete Confirmation" window should close with it (unlike RCT2.)

**Steps to reproduce:**
1. Open Track Designs Manager
2. Select a ride that has pre-made tracks
3. Click the name of a track
4. Click the "Delete" button
5. (Notice the window does not render properly)
6. (Notice the window title is the Ride name, should be "Delete File")
7. Close the previous window that contains the "Rename" and "Delete" buttons
8. (Notice the "Delete Confirmation" window now has no title)

**RCT2 (vanilla) - Window asking to Rename or Delete, and Delete Confirmation window**
RCT2 (vanilla) - Window asking to Rename or Delete, and Delete Confirmation window

**OpenRCT2 - Window asking to Rename or Delete, and Delete Confirmation window**
Window asking to Rename or Delete, and Delete Confirmation window

**OpenRCT2 - Delete Confirmation window after closing previous window (notice title has gone blank)**
Delete Confirmation window after closing previous window (notice title has gone blank)

seifer7:
- Increased the height of the "Delete Confirmation" window in the Track Designs Manager to fit the content.
- Changed the title of the "Delete Confirmation" window to "Delete File" (as per RCT2 vanilla) instead of the track design name.
- "Delete Confirmation" window automatically closes when parent window is closed (Rename / Delete choice window.

**Before Fix**
Before Fix

**After Fix**
After Fix

PS. Please ignore unrelated issue of character corruption (lowercase "L" displays as "ğ")

### DIFF
--- a/src/openrct2-ui/windows/TrackDesignManage.cpp
+++ b/src/openrct2-ui/windows/TrackDesignManage.cpp
@@ -19,6 +19,8 @@
 static constexpr const rct_string_id WINDOW_TITLE = STR_STRING;
 static constexpr const int32_t WH = 44;
 static constexpr const int32_t WW = 250;
+static constexpr const int32_t WH2 = 74;
+static constexpr const int32_t WW2 = 250;
 
 #pragma region Widgets
 
@@ -42,7 +44,7 @@ static rct_widget window_track_manage_widgets[] = {
 };
 
 static rct_widget window_track_delete_prompt_widgets[] = {
-    WINDOW_SHIM(WINDOW_TITLE, WW, WH),
+    WINDOW_SHIM(WINDOW_TITLE, WW2, WH2),
     MakeWidget({ 10, 54}, {110, 12}, WindowWidgetType::Button, WindowColour::Primary, STR_TRACK_MANAGE_DELETE),
     MakeWidget({130, 54}, {110, 12}, WindowWidgetType::Button, WindowColour::Primary, STR_CANCEL             ),
     { WIDGETS_END }
@@ -131,6 +133,7 @@ static void window_track_manage_mouseup(rct_window* w, rct_widgetindex widgetInd
     switch (widgetIndex)
     {
         case WIDX_CLOSE:
+            window_close_by_class(WC_TRACK_DELETE_PROMPT);
             window_close(w);
             break;
         case WIDX_RENAME:
@@ -241,6 +244,10 @@ static void window_track_delete_prompt_mouseup(rct_window* w, rct_widgetindex wi
  */
 static void window_track_delete_prompt_paint(rct_window* w, rct_drawpixelinfo* dpi)
 {
+    char str_delete_file[256];
+    format_string(str_delete_file, 256, STR_DELETE_FILE, nullptr);
+    Formatter::Common().Add<char*>(str_delete_file);
+
     WindowDrawWidgets(w, dpi);
 
     DrawTextWrapped(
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
@seifer7 the ci is failing as your commit message is too long try to follow: https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

seifer7:
> @seifer7 the ci is failing as your commit message is too long try to follow: https://github.com/OpenRCT2/OpenRCT2/wiki/Commit-Messages

I will be more careful in future.

src/openrct2-ui/windows/TrackDesignManage.cpp
```diff
@@ -19,6 +19,8 @@
 static constexpr const rct_string_id WINDOW_TITLE = STR_STRING;
 static constexpr const int32_t WH = 44;
 static constexpr const int32_t WW = 250;
+static constexpr const int32_t WH2 = 74;
```
duncanspumpkin, line 22: This isn't a very descriptive name. It doesn't make it clear that this is for the track delete prompt. Also you should use these sizes for opening the window on L206. I'm somewhat surprised there isn't any drawing issues atm.

src/openrct2-ui/windows/TrackDesignManage.cpp
```diff
@@ -42,7 +44,7 @@ static rct_widget window_track_manage_widgets[] = {
 };
 
 static rct_widget window_track_delete_prompt_widgets[] = {
-    WINDOW_SHIM(WINDOW_TITLE, WW, WH),
+    WINDOW_SHIM(WINDOW_TITLE, WW2, WH2),
```
duncanspumpkin, line 47: By using STR_STRING as the Window Title it means you have to set the common format args in the paint function as you have done on L249. But as you are using a fixed string that doesn't change you dont need to do that. Instead just use STR_DELETE_FILE for the Window Title and it will be processed correctly and you don't need to mess about with setting the common format args.

duncanspumpkin:
Also I'm sure you realise this but if you want to correct your g2.dat render issue (incorrect L's) just regenerate your g2.dat file (or pinch it from a portable zip from the ci's) and copy it to your bin data directory.

duncanspumpkin (changes requested):
This all looks good but it can be improved and simplified see comments.

src/openrct2-ui/windows/TrackDesignManage.cpp
seifer7: Fixed. Also found existing code used in this file was not using the size originally defined constants, fixed that too.

src/openrct2-ui/windows/TrackDesignManage.cpp
seifer7: Thank you. I'll keep that in mind. I've fixed it now.

duncanspumpkin (approved):

PR #15154, opened 2021-08-01
Fix #14974: too many dodgems not handled correctly (from NSF)

### LINKED ISSUE #14974: [new-save-format] More than 31 Dodgems/Flying Saucers crashes the game.
Setting Dodgems, Flying Saucers or any Dodgems-like ride to have more than 31 cars causes the game to crash, probably because the game is unsure how and where to spawn them in.

Gymnasiast:
Properly check if all vehicles have been created, and clean them up properly if not.

### DIFF
--- a/src/openrct2/ride/RideConstruction.cpp
+++ b/src/openrct2/ride/RideConstruction.cpp
@@ -40,6 +40,7 @@
 #include "RideData.h"
 #include "Track.h"
 #include "TrackData.h"
+#include "TrainManager.h"
 #include "Vehicle.h"
 
 bool gGotoStartPlacementMode = false;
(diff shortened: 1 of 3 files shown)


src/openrct2/ride/RideConstruction.cpp
```diff
@@ -195,11 +196,21 @@ static void ride_remove_vehicles(Ride* ride)
...
-            ride->vehicles[i] = SPRITE_INDEX_NULL;
+            vehicles[i] = SPRITE_INDEX_NULL;
         }
 
         for (size_t i = 0; i < MAX_STATIONS; i++)
-            ride->stations[i].TrainAtStation = RideStation::NO_TRAIN;
+            stations[i].TrainAtStation = RideStation::NO_TRAIN;
+
+        // Also clean up orphaned vehicles for good measure.
+        for (auto* vehicle : TrainManager::View())
```
duncanspumpkin, line 206: I somewhat wonder if this may cause issues for heavily hacked park creation.

Gymnasiast: It’s only called when removing all vehicles from a ride, so I’d hazard a guess it’s _especially_ beneficial to such parks.

PR #15159, opened 2021-08-01
Fix #14918 refactor stationIndex and loop to avoid null in/decrement

### LINKED ISSUE #14918: Refactor use of overflow in Ride.cpp

SaadRehmanCS:


SaadRehmanCS:
I'm unable to solve the commit message error. Apologies for the mess created!

duncanspumpkin:
Ive rebased your work and compressed it to a single commit.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2528,35 +2528,42 @@ static rct_string_id window_ride_get_status_station(rct_window* w, Formatter& ft
         return STR_NONE;
 
     int32_t count = w->ride.view - ride->num_vehicles - 1;
-    StationIndex stationIndex = STATION_INDEX_NULL;
     rct_string_id stringId = STR_EMPTY;
+    int32_t stationIndex = 0;
+    auto lastStation = ride->stations[0];
 
-    do
+    for (const auto& station : ride->stations)
```
duncanspumpkin, line 2535: This is the exact same change as the previous PR suggesting that a function for this might be an idea.

This was my quick idea of what such a function might be. By making the return the index and an optional it makes it clear to callers that this can fail and that they must account for failure.
```cpp
std::optional<uint16_t> GetStationIndexFromViewSelection(const rct_window& w)
{
    const auto* ride = get_ride(w.number);
    if (ride == nullptr)
        return {};

    int32_t viewSelectionIndex = w.ride.view - 1 - ride->num_vehicles;
    if (viewSelectionIndex < 0)
    {
        return {};
    }

    for (uint16_t index = 0; index < sizeof(ride->stations); ++index)
    {
        const auto& station = ride->stations[index];
        if (!station.Start.isNull())
        {
            if (viewSelectionIndex-- == 0)
            {
                return { index };
            }
        }
    }
    return {};
}
```

duncanspumpkin:
Super seeded by #15163

PR #15160, opened 2021-08-01
Remove gS6Info (from NSF)

Gymnasiast:


src/openrct2-ui/windows/ObjectLoadError.cpp
```diff
@@ -21,6 +21,7 @@
 #include <openrct2/object/ObjectRepository.h>
 #include <openrct2/platform/platform.h>
 #include <openrct2/windows/Intent.h>
+#include <sstream>
```
duncanspumpkin, line 24: can be removed from this pr

Issue #7682, opened 2018-06-14
Feature request: lock camera to vehicle in main window


meatbee:
I think it would be a fun feature to lock the main camera (outside of the coaster's own window) to a ride vehicle. Might be fun for title sequences in future to enable a moving tour of the park instead of just static views (particularly if the ride is underground and the camera is in normal mode).

Maybe this could be through a double click of the already existing "locate this on main view" button.

meatbee:
Oops, my bad then. Still it would be nice to have the feature during gameplay.

SevenChords:
any update for if/when a more convenient way than the title sequence editor will be implemented?

duncanspumpkin:
@SevenChords rather than spam every issue related to this it would be more constructive if you implemented it yourself. We are all volunteers doing this in our spare time working on whatever takes our fancy.

SevenChords:
@duncanspumpkin you could've just said on the first issue where i asked this that it isn't planned to implement because other issues are at a higher priority, no need to get rude about it.
and about implementing it myself, i would do that, but unfortunately i am not good enough with c++ yet to do such a thing, which is why i asked wether this was already planned to implement on existing issues, and after the ones i asked it on got closed more or less directly afterwards instead of an answer, i asked again.

AaronVanGeffen:
We already implemented this in OpenLoco, which has a similar codebase. Perhaps some inspiration could be taken from there. cf. https://github.com/OpenLoco/OpenLoco/pull/760

SevenChords:
thanks aaron, will try to figure it out using that

seifer7:
I am working on this now. Will submit a pull request soon.

PR #15162, opened 2021-08-01
Feature #7682: Lock camera to vehicle in main window

seifer7:
Pressing the "Locate" button in the Ride window when selected on a "Vehicle/Train" will now have the main viewport track the vehicle.

**Still to complete:**
- [x] Drop-down selection for Locate or Follow
- [x] While following, automatically toggle underground view on/off as required
- [x] Implement tracking for Staff & Guests
- [x] More testing

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3655,6 +3655,7 @@ STR_6454    :Can’t rename banner…
 STR_6455    :Can’t rename sign…
 STR_6456    :Giant Screenshot
 STR_6457    :Report a bug on GitHub
+STR_6458    :Follow this on Main View
 
 #############
 # Scenarios #
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -1,6 +1,7 @@
 0.3.4.1+ (in development)
 ------------------------------------------------------------------------
 - Feature: [#3868] Initial support for using TTF in OpenGL mode (still contains bugs).
+- Feature: [#7682] Follow ride/guest/staff in main window viewport.
 - Feature: [#13407] Allow building chain lifts on enclosed dinghy slide pieces when cheats are on.
 - Feature: [#15084] [Plugin] Add "vehicle.crash" hook.
 - Feature: [#15143] Added a shortcut key for Giant Screenshot.
--- a/src/openrct2-ui/windows/Guest.cpp
+++ b/src/openrct2-ui/windows/Guest.cpp
@@ -7,6 +7,7 @@
  * OpenRCT2 is licensed under the GNU General Public License version 3.
  *****************************************************************************/
 
+#include <openrct2-ui/interface/Dropdown.h>
 #include <openrct2-ui/interface/Viewport.h>
 #include <openrct2-ui/interface/Widget.h>
 #include <openrct2-ui/windows/Window.h>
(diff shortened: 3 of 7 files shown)


Gymnasiast (dismissed):
> Pressing the "Locate" button in the Ride window when selected on a "Vehicle/Train" will now have the main viewport track the vehicle.

~~This is not a good idea. People are used to this button just scrolling and not following. Checking if CTRL is pressed would be a better idea.~~

seifer7:
> > Pressing the "Locate" button in the Ride window when selected on a "Vehicle/Train" will now have the main viewport track the vehicle.
> 
> This is not a good idea. People are used to this button just scrolling and not following. Checking if CTRL is pressed would be a better idea.

Maybe you missed the "Still to complete:...  _Drop-down selection for Locate or Track (currently replaces locate)_"

ocalhoun6:
Once this is done, it would be really great to see the same done for peep/staff windows.

ZehMatt:
I sort of feel like this can be probably achieved with a plugin, if not it would be good to implement the required API to do it.

ZehMatt:
> > I sort of feel like this can be probably achieved with a plugin, if not it would be good to implement the required API to do it.
> 
> I'll have a look at the APIs and see if it's possible. I did have an alternate implementation idea that I may be well suited for a plugin.
> 
> Where do you draw the line between something that should be implemented in the main game, and what should be a plugin?
> I thought things that are obscure would be for plugins, this feature however seems like it should have been included in the game to start with (imo.)

That depends highly on the feature and the benefits and how much code it would add that needs to be maintained, thats my personal opinion on that matter I also think there is no definite answer to your question.

Broxzier:
The moment the user attempts to move the camera, the sprite following should stop immediately. Rotating the view should not. This same feature could also be used to follow guests, though with them entering rides it can become a bit more complicated.

seifer7:
Video preview of the feature working so far:

[OpenRCT2 Development: Follow ride in main window](https://www.youtube.com/watch?v=KuXKicHo6d4)

IntelOrca:
> I sort of feel like this can be probably achieved with a plugin, if not it would be good to implement the required API to do it.

I think this is okay, particularly when we have no API yet for modifying existing windows. This seems similar in functionality to the isolate sound from window viewport mechanism.

seifer7:
Implemented the staff & guest following too now!

Unless someone has some suggestions, I think this is about ready for a review!

AaronVanGeffen:
What happens when the sprite id becomes invalid? For instance, when a vehicle crashes. I remember we had to handle that particular case for OpenLoco as well.

seifer7:
> What happens when the sprite id becomes invalid? For instance, when a vehicle crashes. I remember we had to handle that particular case for OpenLoco as well.

I did testing on all sorts of scenarios and couldn't find any problems.
As pointed out by @Broxzier, the `viewport_smart_follow_sprite` function is really... smart... and takes care of loads of different scenarios. It's the same function that is used to follow peeps etc in the Guest/Ride Window viewports.

seifer7:
Sorry about 3 commits to do the changelog entry... it's 3:45am here. I think I should go to bed.

Broxzier:
Since OP has not responded for a while, I'll squash and rebase this.

ZehMatt (approved):

Broxzier (approved):

PR #15164, opened 2021-08-01
Feature: Highlight selected Tile Inspector element

ZehMatt:
The tile inspector will highlight the selected element the same way its done with construction. Unfortunately the way tracks are drawn I did not implement this for tracks, this would add a huge amount of code into quite a few files.

### DIFF
--- a/src/openrct2/paint/tile_element/Paint.Entrance.cpp
+++ b/src/openrct2/paint/tile_element/Paint.Entrance.cpp
@@ -21,6 +21,7 @@
 #include "../../world/Entrance.h"
 #include "../../world/Footpath.h"
 #include "../../world/Park.h"
+#include "../../world/TileInspector.h"
 #include "../Paint.h"
 #include "../Supports.h"
 #include "Paint.TileElement.h"
@@ -95,6 +96,13 @@ static void ride_entrance_exit_paint(paint_session* session, uint8_t direction, 
     if (tile_element->IsGhost())
     {
         session->InteractionType = ViewportInteractionItem::None;
+        image_id = CONSTRUCTION_MARKER;
+        entranceImageId = image_id;
+        if (transparant_image_id)
+            transparant_image_id = image_id;
+    }
+    else if (OpenRCT2::TileInspector::IsElementSelected(tile_element))
+    {
         image_id = CONSTRUCTION_MARKER;
         entranceImageId = image_id;
         if (transparant_image_id)
@@ -225,6 +233,10 @@ static void park_entrance_paint(paint_session* session, uint8_t direction, int32
     if (tile_element->IsGhost())
     {
         session->InteractionType = ViewportInteractionItem::None;
+        ghost_id = CONSTRUCTION_MARKER;
+    }
+    else if (OpenRCT2::TileInspector::IsElementSelected(tile_element))
+    {
         ghost_id = CONSTRUCTION_MARKER;
     }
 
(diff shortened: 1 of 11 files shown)


Broxzier (commented):
As discussed on Discord, this introduces a new tile element pointer that you have to manually sync with the selection of the tile inspector. Because of this, it currently already lacks some syncing (Ctrl+clicking for instance). While that can be fixed by remembering to call the new setter everywhere where the tile or index changes, I'd rather see them being reused, or updating the getter to use the selection variable.

```cpp
    const TileElement* GetSelectedElement()
    {
        return map_get_nth_element_at(windowTileInspectorTile.ToCoordsXY(), windowTileInspectorSelectedIndex);
    }
```

This would guarantee that it'll be synced with what the Tile Inspector has selected, however this introduces a lot of extra calls to the map element lookups. Ideally, the pointer would be stored and updated between the game tick and rendering.

ZehMatt:
> As discussed on Discord, this introduces a new tile element pointer that you have to manually sync with the selection of the tile inspector. Because of this, it currently already lacks some syncing (Ctrl+clicking for instance). While that can be fixed by remembering to call the new setter everywhere where the tile or index changes, I'd rather see them being reused, or updating the getter to use the selection variable.
> 
> ```c++
>     const TileElement* GetSelectedElement()
>     {
>         return map_get_nth_element_at(windowTileInspectorTile.ToCoordsXY(), windowTileInspectorSelectedIndex);
>     }
> ```
> 
> This would guarantee that it'll be synced with what the Tile Inspector has selected, however this introduces a lot of extra calls to the map element lookups. Ideally, the pointer would be stored and updated between the game tick and rendering.

I fixed the remaining few cases, I kept it strictly pointer with get/set to avoid those lookups, it only cares if the element is the current one. The code can be surely improved but I think this is currently the best choice without sacrificing either performance or having too many changes which don't quite fall into this scope.

Broxzier:
Works nicely :+1: The whole getter/setter stuff should be cleaned up, but that's not in scope for this PR. This definitely deserves a changelog entry.

duncanspumpkin (changes requested):
This is going to cause all sorts of crashes. As I've said once or twice recently you cannot save tile element pointers. They will become invalid. You're best bet is to fetch the nth tile at the start of a paint session and stash the pointer into the paint session.

ZehMatt:
> This is going to cause all sorts of crashes. As I've said once or twice recently you cannot save tile element pointers. They will become invalid. You're best bet is to fetch the nth tile at the start of a paint session and stash the pointer into the paint session.

It does never read of that pointer, it acts as a symbol to match if the selected element is the currently drawn one, no data is accessed from that pointer.

Broxzier:
While how it's currently used there is no chance of crashes, it is possible that the pointer will get reused for another element in-between `SetSelectedElement` calls making it highlight the wrong one, and there is no way to prevent the pointer from being used in the future, for example for highlighting entire track blocks. I think the pointer itself is fine, but it has to be updated between every game update and paint session.

ZehMatt:
I could modify the code a bit and just ask the tile inspector if the element is highlighted this hides entirely how it works underneath and no one has to worry about pointers for the query.

src/openrct2/paint/tile_element/Paint.Entrance.cpp
```diff
@@ -227,6 +235,10 @@ static void park_entrance_paint(paint_session* session, uint8_t direction, int32
         session->InteractionType = ViewportInteractionItem::None;
         ghost_id = CONSTRUCTION_MARKER;
     }
+    else if (OpenRCT2::TileInspector::GetSelectedElement() == tile_element)
```
duncanspumpkin, line 238: this ones still the get

duncanspumpkin (approved):

PR #15165, opened 2021-08-02
Add createEntity API to scripting

ZehMatt:


distribution/openrct2.d.ts
```diff
@@ -1065,6 +1066,8 @@ declare global {
         "litter" |
         "money_effect" |
         "peep" |
+        "staff" |
+        "guest" |
```
IntelOrca, line 1070: This will confuse plugin writers, as this won't be the entity type returned from `entity.type`.

ZehMatt: Should I fix that up so it will rather return "staff" or "guest" instead of "peep"?

duncanspumpkin: Guests and staff are quite different I'd prefer that plugins had to know the difference

ZehMatt: > Guests and staff are quite different I'd prefer that plugins had to know the difference

Whats your suggestion? If I return "staff" or "guest" instead of "peep" on the entities then plugin may cease to function. Another way would be having distinct functions for entity creation maybe.

duncanspumpkin: I would prefer we break plugins, unsure when we should break them but they should be made to specify staff or guest.

IntelOrca: They aren't different, they are very much sub classes of peep. You have two options here:

* Specify a sub type in the initiliser object.
* Create a new type. You can avoid duplicates if you want: e.g.
```ts
type StrongEntityType = 'car' | 'litter'
type EntityType = 'peep' | StrongEntityType
type CreateEntityType = 'guest' | 'staff' | StrongEntityType
```

duncanspumpkin: `peep` isn't an entity type anymore though it hasn't been for a long time. The entity type is `guest` or `staff`. You'll notice we dont have `miscEntity` in this list either which was also split for the same reasons. It would make most sense that scripting API matched the game API as otherwise there will just be confusion at the bridge.

IntelOrca: @duncanspumpkin how would a plugin loop through all peeps, changing a common property?

duncanspumpkin: They would loop through all guests, then all staff same as what it would do under the hood anyway.

ZehMatt: > @duncanspumpkin how would a plugin loop through all peeps, changing a common property?

You could also do this with a single list asking if the entity is either "staff" or "guest", peep is something that should be entirely hidden.

Gymnasiast (approved):

duncanspumpkin (approved):

PR #15167, opened 2021-08-02
Convert ride_id_t to a strong type

ZehMatt:
Doing this revealed quite a few open issues, the changes are a bit bulky since this forces to use the correct type, but I think this is for the better. The same thing should be done to the BannerIndex type also, I'm sure it will reveal some things like in this one.

### DIFF
--- a/src/openrct2-ui/WindowManager.cpp
+++ b/src/openrct2-ui/WindowManager.cpp
@@ -396,7 +396,7 @@ public:
             case INTENT_ACTION_INVALIDATE_VEHICLE_WINDOW:
             {
                 auto vehicle = static_cast<Vehicle*>(intent.GetPointerExtra(INTENT_EXTRA_VEHICLE));
-                auto w = window_find_by_number(WC_RIDE, static_cast<int32_t>(vehicle->ride));
+                auto w = window_find_by_number(WC_RIDE, EnumValue(vehicle->ride));
                 if (w == nullptr)
                     return;
 
--- a/src/openrct2-ui/WindowManager.cpp
+++ b/src/openrct2-ui/WindowManager.cpp
@@ -184,9 +184,9 @@ public:
             case WD_BANNER:
                 return window_banner_open(id);
             case WD_DEMOLISH_RIDE:
-                return window_ride_demolish_prompt_open(get_ride(id));
+                return window_ride_demolish_prompt_open(get_ride(static_cast<ride_id_t>(id)));
             case WD_REFURBISH_RIDE:
-                return window_ride_refurbish_prompt_open(get_ride(id));
+                return window_ride_refurbish_prompt_open(get_ride(static_cast<ride_id_t>(id)));
             case WD_NEW_CAMPAIGN:
                 return window_new_campaign_open(id);
             case WD_SIGN:
@@ -264,7 +264,8 @@ public:
             }
             case WC_RIDE:
             {
-                auto ride = get_ride(intent->GetSIntExtra(INTENT_EXTRA_RIDE_ID));
+                const auto rideId = static_cast<ride_id_t>(intent->GetSIntExtra(INTENT_EXTRA_RIDE_ID));
+                auto ride = get_ride(rideId);
                 return ride == nullptr ? nullptr : window_ride_main_open(ride);
             }
             case WC_TRACK_DESIGN_PLACE:
@@ -343,13 +344,13 @@ public:
                 if (w == nullptr || w->number != rideIndex)
                 {
                     window_close_construction_windows();
-                    _currentRideIndex = rideIndex;
+                    _currentRideIndex = static_cast<ride_id_t>(rideIndex);
                     w = OpenWindow(WC_RIDE_CONSTRUCTION);
                 }
                 else
                 {
                     ride_construction_invalidate_current_track();
-                    _currentRideIndex = rideIndex;
+                    _currentRideIndex = static_cast<ride_id_t>(rideIndex);
                 }
                 break;
             }
@@ -395,7 +396,7 @@ public:
             case INTENT_ACTION_INVALIDATE_VEHICLE_WINDOW:
             {
                 auto vehicle = static_cast<Vehicle*>(intent.GetPointerExtra(INTENT_EXTRA_VEHICLE));
-                auto w = window_find_by_number(WC_RIDE, vehicle->ride);
+                auto w = window_find_by_number(WC_RIDE, static_cast<int32_t>(vehicle->ride));
                 if (w == nullptr)
                     return;
 
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -1235,7 +1235,7 @@ static rct_window* window_ride_open(Ride* ride)
     w->widgets = window_ride_page_widgets[WINDOW_RIDE_PAGE_MAIN];
     w->enabled_widgets = window_ride_page_enabled_widgets[WINDOW_RIDE_PAGE_MAIN];
     w->hold_down_widgets = window_ride_page_hold_down_widgets[WINDOW_RIDE_PAGE_MAIN];
-    w->number = static_cast<int32_t>(ride->id);
+    w->rideId = ride->id;
 
     w->page = WINDOW_RIDE_PAGE_MAIN;
     w->vehicleIndex = 0;
@@ -1265,7 +1265,7 @@ rct_window* window_ride_main_open(Ride* ride)
         return nullptr;
     }
 
-    rct_window* w = window_bring_to_front_by_number(WC_RIDE, static_cast<int32_t>(ride->id));
+    rct_window* w = window_bring_to_front_by_number(WC_RIDE, EnumValue(ride->id));
     if (w == nullptr)
     {
         w = window_ride_open(ride);
@@ -1306,7 +1306,7 @@ static rct_window* window_ride_open_station(Ride* ride, StationIndex stationInde
     if (ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_NO_VEHICLES))
         return window_ride_main_open(ride);
 
-    auto w = window_bring_to_front_by_number(WC_RIDE, static_cast<int32_t>(ride->id));
+    auto w = window_bring_to_front_by_number(WC_RIDE, EnumValue(ride->id));
     if (w == nullptr)
     {
         w = window_ride_open(ride);
@@ -1408,7 +1408,7 @@ rct_window* window_ride_open_vehicle(Vehicle* vehicle)
         view++;
     }
 
-    rct_window* w = window_find_by_number(WC_RIDE, static_cast<int32_t>(ride->id));
+    rct_window* w = window_find_by_number(WC_RIDE, EnumValue(ride->id));
     if (w != nullptr)
     {
         w->Invalidate();
@@ -1443,8 +1443,8 @@ rct_window* window_ride_open_vehicle(Vehicle* vehicle)
             }
         }
 
-        w = openedPeepWindow ? window_find_by_number(WC_RIDE, static_cast<int32_t>(ride->id))
-                             : window_bring_to_front_by_number(WC_RIDE, static_cast<int32_t>(ride->id));
+        w = openedPeepWindow ? window_find_by_number(WC_RIDE, EnumValue(ride->id))
+                             : window_bring_to_front_by_number(WC_RIDE, EnumValue(ride->id));
     }
 
     if (w == nullptr)
@@ -1756,7 +1756,7 @@ static void window_ride_main_mouseup(rct_window* w, rct_widgetindex widgetIndex)
             if (ride != nullptr)
             {
                 ride_construct(ride);
-                if (window_find_by_number(WC_RIDE_CONSTRUCTION, static_cast<int32_t>(ride->id)) != nullptr)
+                if (window_find_by_number(WC_RIDE_CONSTRUCTION, EnumValue(ride->id)) != nullptr)
                 {
                     window_close(w);
                 }
@@ -2252,8 +2252,7 @@ static void window_ride_main_dropdown(rct_window* w, rct_widgetindex widgetIndex
                 uint8_t rideType = RideDropdownData[rideLabelId].ride_type_id;
                 if (rideType < RIDE_TYPE_COUNT)
                 {
-                    const auto rideId = w->rideId;
-                    auto rideSetSetting = RideSetSettingAction(rideId, RideSetSetting::RideType, rideType);
+                    auto rideSetSetting = RideSetSettingAction(w->rideId, RideSetSetting::RideType, rideType);
                     rideSetSetting.SetCallback([](const GameAction* ga, const GameActions::Result* result) {
                         // Reset ghost track if ride construction window is open, prevents a crash
                         // Will get set to the correct Alternative variable during set_default_next_piece.
@@ -3093,8 +3092,7 @@ static void window_ride_vehicle_scrollpaint(rct_window* w, rct_drawpixelinfo* dp
  */
 static void window_ride_mode_tweak_increase(rct_window* w)
 {
-    const auto rideId = w->rideId;
-    auto ride = get_ride(rideId);
+    auto ride = get_ride(w->rideId);
     if (ride == nullptr)
         return;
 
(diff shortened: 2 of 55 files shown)


ZehMatt:
Replays are failing because of the updated peep thoughts

src/openrct2-ui/WindowManager.cpp
```diff
@@ -184,9 +184,9 @@ class WindowManager final : public IWindowManager
             case WD_BANNER:
                 return window_banner_open(id);
             case WD_DEMOLISH_RIDE:
-                return window_ride_demolish_prompt_open(get_ride(id));
+                return window_ride_demolish_prompt_open(get_ride(static_cast<ride_id_t>(id)));
```
duncanspumpkin, line 187: I feel we should just bubble up the change into the callers of OpenDetails

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1933,14 +1933,15 @@ static void window_ride_construction_mouseup_demolish(rct_window* w)
         _currentTrackPieceType, 0,
         { _currentTrackBegin.x, _currentTrackBegin.y, _currentTrackBegin.z, _currentTrackPieceDirection });
 
+    // FIXME: This can crash if this is multiplayer, the window might be closed before the callback fires.
```
duncanspumpkin, line 1936: Assume fix is for the game action result to specify the ride id?

src/openrct2/actions/RideDemolishAction.cpp
```diff
@@ -130,14 +130,15 @@ GameActions::Result::Ptr RideDemolishAction::DemolishRide(Ride* ride) const
...
+
+    News::DisableNewsItems(News::ItemType::Ride, static_cast<uint32_t>(_rideIndex));
 
     UnlinkAllBannersForRide(_rideIndex);
 
     for (auto peep : EntityList<Guest>())
     {
-        uint8_t ride_id_bit = _rideIndex % 8;
-        uint8_t ride_id_offset = _rideIndex / 8;
+        uint8_t ride_id_bit = static_cast<uint32_t>(_rideIndex) % 8;
```
duncanspumpkin, line 140: This is obviously broken for >255 rides but is fixed in my pr for the ride use memory. I'm only putting this note here for reference in case anyone else wonders.

src/openrct2/interface/InteractiveConsole.cpp
```diff
@@ -194,7 +194,7 @@ static int32_t cc_rides(InteractiveConsole& console, const arguments_t& argv)
                 }
                 else
                 {
-                    int32_t res = set_operating_setting(ride_index, RideSetSetting::RideType, type);
+                    int32_t res = set_operating_setting(static_cast<ride_id_t>(ride_index), RideSetSetting::RideType, type);
```
duncanspumpkin, line 197: you could do the cast on L185 to save a small amount of changes.

src/openrct2/interface/Window_internal.h
```diff
@@ -48,7 +48,7 @@ struct rct_window
     rct_windownumber number{};
     uint16_t flags{};
     rct_scroll scrolls[3];
-    uint8_t list_item_positions[1024]{};
+    uint32_t list_item_positions[1024]{};
```
duncanspumpkin, line 51: have you checked all users of this field to make sure they werent abusing its size?

src/openrct2-ui/windows/RideConstruction.cpp
ZehMatt: You could just capture the ride id before and find the window in the callback via id.

src/openrct2/interface/Window_internal.h
ZehMatt: The only actual use is the guest window, it has an assignment in multiplayer but its never used anywhere else really.

src/openrct2-ui/windows/RideConstruction.cpp
ZehMatt: I can get rid of the comment in this PR, I just didn't wanna loose track of it.

src/openrct2/ride/RideTypes.h
```diff
@@ -13,11 +13,12 @@
...
 #include <cstdint>
 #include <limits>
+#include <type_traits>
 
-using ride_id_t = uint16_t;
+enum class ride_id_t : uint16_t;
 struct Ride;
 
-constexpr const ride_id_t RIDE_ID_NULL = std::numeric_limits<ride_id_t>::max();
+constexpr const ride_id_t RIDE_ID_NULL = static_cast<ride_id_t>(std::numeric_limits<std::underlying_type_t<ride_id_t>>::max());
```
duncanspumpkin, line 21: As its an enum class couldn't this just be a defined value of the class:
```
ride_id_t::Null
```

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -323,18 +323,19 @@ namespace OpenRCT2::Scripting
             }
         }
 
-        ride_id_t ride_get() const
+        int32_t ride_get() const
         {
             auto vehicle = GetVehicle();
-            return vehicle != nullptr ? vehicle->ride : 0;
+            return static_cast<int32_t>(vehicle != nullptr ? vehicle->ride : RIDE_ID_NULL);
```
duncanspumpkin, line 329: should we increment plugin api for this change.

src/openrct2-ui/WindowManager.cpp
ZehMatt: If we provide strong types for the others we could simply add overloads for each and also use variants on the window but I think this sort of clean up would be the last step after having the types.

src/openrct2/ride/RideTypes.h
ZehMatt: It could, but RIDE_ID_NULL is probably heavily used everywhere, totally okay for a follow up PR.

src/openrct2/scripting/ScEntity.hpp
ZehMatt: I wasn't even sure if I should change that, but 0 is technically a valid ride in some cases.

IntelOrca: In JS you can return `undefined` or `null`, but you need to change the signature first. `std::optional` will work I think.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2250,7 +2252,8 @@ static void window_ride_main_dropdown(rct_window* w, rct_widgetindex widgetIndex
                 uint8_t rideType = RideDropdownData[rideLabelId].ride_type_id;
                 if (rideType < RIDE_TYPE_COUNT)
                 {
-                    auto rideSetSetting = RideSetSettingAction(w->number, RideSetSetting::RideType, rideType);
+                    const auto rideId = w->rideId;
```
duncanspumpkin, line 2255: this is already the correct type no need for the intermediate variable.

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -3090,7 +3093,8 @@ static void window_ride_vehicle_scrollpaint(rct_window* w, rct_drawpixelinfo* dp
  */
 static void window_ride_mode_tweak_increase(rct_window* w)
 {
-    auto ride = get_ride(w->number);
+    const auto rideId = w->rideId;
```
duncanspumpkin, line 3096: as is here.

src/openrct2/network/NetworkTypes.h
```diff
@@ -126,10 +126,11 @@ template<typename T, size_t _TypeID> struct NetworkObjectId_t
 };
 #pragma pack(pop)
 
+using NetworkRideId_t = ride_id_t;
```
duncanspumpkin, line 129: Won't this mean that there is a specialisation in the data serialiser for ride_id_t?

src/openrct2/interface/InteractiveConsole.cpp
ZehMatt: Theres no saving here, L191 needs changing then, its a local scope also.

src/openrct2/scripting/ScEntity.hpp
ZehMatt: I sort of think this should not be optional, you can set RIDE_ID_NULL but you would never be able to receive that value. I will do a version increment but return RIDE_ID_NULL instead, this is also a value that would be only returned on bad/unused vehicles I think.

src/openrct2/actions/RideSetAppearanceAction.cpp
```diff
@@ -139,7 +139,7 @@ GameActions::Result::Ptr RideSetAppearanceAction::Execute() const
             gfx_invalidate_screen();
             break;
     }
-    window_invalidate_by_number(WC_RIDE, _rideIndex);
+    window_invalidate_by_number(WC_RIDE, static_cast<int32_t>(_rideIndex));
```
Gymnasiast, line 142: This is used so much, I think we should create a function for it. E.g. `WindowRideInvalidateById(ride_id_t id)`. It would make the code much cleaner.

ZehMatt: I agree, theres a lot of UI code that needs some cleaning.

duncanspumpkin (approved):

src/openrct2/EditorObjectSelectionSession.cpp
```diff
@@ -192,14 +192,10 @@ void setup_in_use_selection_flags()
...
+    for (auto& ride : GetRideManager())
     {
-        auto ride = get_ride(ride_index);
-        if (ride != nullptr)
-        {
-            ObjectEntryIndex type = ride->subtype;
-            Editor::SetSelectedObject(ObjectType::Ride, type, OBJECT_SELECTION_FLAG_SELECTED);
-        }
+        ObjectEntryIndex type = ride.subtype;
+        Editor::SetSelectedObject(ObjectType::Ride, type, OBJECT_SELECTION_FLAG_SELECTED);
```
Gymnasiast, line 198: We should make a refactor item for iterating over all rides.

We should also make one for iterating over a tile.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -588,7 +588,7 @@ static void window_ride_construction_close(rct_window* w)
     gMapSelectFlags &= ~MAP_SELECT_FLAG_ENABLE_ARROW;
 
     // In order to cancel the yellow arrow correctly the
-    // selection tool should be cancelled. Don't do a tool cancel if
+    // selection tool should be canceled. Don't do a tool cancel if
```
Gymnasiast, line 591: cancelled is correct in ~~proper~~British English.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1930,7 +1932,8 @@ static bool track_design_place_preview(TrackDesign* td6, money32* cost, Ride** o
     int32_t mapSize = gMapSize << 4;
 
     _currentTrackPieceDirection = 0;
-    int32_t z = place_virtual_track(td6, PTD_OPERATION_GET_PLACE_Z, true, GetOrAllocateRide(0), { mapSize, mapSize, 16 });
+    int32_t z = place_virtual_track(
+        td6, PTD_OPERATION_GET_PLACE_Z, true, GetOrAllocateRide(static_cast<ride_id_t>(0)), { mapSize, mapSize, 16 });
```
Gymnasiast, line 1936:
```suggestion
        td6, PTD_OPERATION_GET_PLACE_Z, true, GetOrAllocateRide(ride_id_t(0)), { mapSize, mapSize, 16 });
```

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -437,7 +438,8 @@ rct_string_id TrackDesign::CreateTrackDesignMaze(const Ride& ride)
 
     // Save global vars as they are still used by scenery????
     int32_t startZ = _trackPreviewOrigin.z;
-    place_virtual_track(this, PTD_OPERATION_DRAW_OUTLINES, true, GetOrAllocateRide(0), { 4096, 4096, 0 });
+    place_virtual_track(
+        this, PTD_OPERATION_DRAW_OUTLINES, true, GetOrAllocateRide(static_cast<ride_id_t>(0)), { 4096, 4096, 0 });
```
Gymnasiast, line 442:
```suggestion
        this, PTD_OPERATION_DRAW_OUTLINES, true, GetOrAllocateRide(ride_id_t(0)), { 4096, 4096, 0 });
```

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2505,10 +2506,12 @@ namespace RCT1
         {
             if (_s4.scenario_slot_index == SC_URBAN_PARK && _isScenario)
             {
+                const auto firstRideId = static_cast<ride_id_t>(0);
```
Gymnasiast, line 2509: If we are naming this parameter anyway, `merryGoRoundId` is a better fit and also illustrates the purpose of the fix better.

src/openrct2/peep/Peep.h
```diff
@@ -774,6 +774,7 @@ struct Guest : Peep
     void UpdateEasterEggInteractions();
     void InsertNewThought(PeepThoughtType thought_type);
     void InsertNewThought(PeepThoughtType thought_type, ShopItem thought_arguments);
+    void InsertNewThought(PeepThoughtType thought_type, ride_id_t ride);
```
Gymnasiast, line 777:
```suggestion
    void InsertNewThought(PeepThoughtType thought_type, ride_id_t rideId);
```

If a variable is called ride, it should be a Ride, RCT2Ride or something like this, not just an ID, IMO.

src/openrct2-ui/windows/TrackDesignPlace.cpp
```diff
@@ -469,7 +472,9 @@ static int32_t window_track_place_get_base_z(const CoordsXY& loc)
     if (surfaceElement->GetWaterHeight() > 0)
         z = std::max(z, surfaceElement->GetWaterHeight());
 
-    return z + place_virtual_track(_trackDesign.get(), PTD_OPERATION_GET_PLACE_Z, true, GetOrAllocateRide(0), { loc, z });
+    return z
+        + place_virtual_track(
+               _trackDesign.get(), PTD_OPERATION_GET_PLACE_Z, true, GetOrAllocateRide(static_cast<ride_id_t>(0)), { loc, z });
```
Gymnasiast, line 477:
```suggestion
               _trackDesign.get(), PTD_OPERATION_GET_PLACE_Z, true, GetOrAllocateRide(ride_id_t(0)), { loc, z });
```

src/openrct2-ui/windows/RideConstruction.cpp
ZehMatt: Damn, that slipped it just annoyed me as VS kept outlining it.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -509,7 +509,7 @@ static void close_ride_window_for_construction(rct_windownumber number)
 rct_window* window_ride_construction_open()
 {
     ride_id_t rideIndex = _currentRideIndex;
-    close_ride_window_for_construction(rideIndex);
+    close_ride_window_for_construction(EnumValue(rideIndex));
```
Gymnasiast, line 512: Probably better to change the signature of `close_ride_window_for_construction()` to take `ride_id_t`?

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -85,6 +85,8 @@ static bool _trackDesignPlaceStateHasScenery = false;
 static bool _trackDesignPlaceStatePlaceScenery = true;
 static bool _trackDesignPlaceIsReplay = false;
 
+static constexpr ride_id_t PreviewRideId = static_cast<ride_id_t>(0);
```
Gymnasiast, line 88: I think we should add a comment:
```
// Used for the track preview in the Track Designs window, but not for the ghost shown when placing one.
```

src/openrct2/ride/TrackDesign.h
```diff
@@ -191,6 +191,9 @@ enum
     MAZE_ELEMENT_TYPE_EXIT = (1 << 7)
 };
 
+// Used for the track preview in the Track Designs window, but not for the ghost shown when placing one.
```
Gymnasiast, line 194: Can you remove this comment again? I know I asked for it, but on second glance, it seems like it’s not the case after all. (Which is also why I removed  the line note in which I asked for it.)

Gymnasiast (commented):
The comment is the last thing. After all, I’ll merge it.

Gymnasiast (approved):

PR #15174, opened 2021-08-04
Deprecate peep in scripting

ZehMatt:


src/openrct2/scripting/ScEntity.hpp
```diff
@@ -56,8 +56,9 @@ namespace OpenRCT2::Scripting
                     case EntityType::Vehicle:
                         return "car";
                     case EntityType::Guest:
+                        return "guest";
                     case EntityType::Staff:
-                        return "peep";
+                        return "staff";
```
IntelOrca, line 61: As discussed in discord#plugin, we will probably now have an `targetApiVersion` in the plugin header that controls the behaviour.

IntelOrca:
Looks good. I think just changelog.md and scripting.md need to be updated now.

distribution/openrct2.d.ts
```diff
@@ -1064,7 +1070,12 @@ declare global {
         "jumping_fountain_water" |
         "litter" |
         "money_effect" |
+        /**
+         * @deprecated since version 34, use guest or staff instead.
+         */
```
Broxzier, line 1075: Just to note, the `@deprecated` tag doesn't work for values. It's still good to mark them as such of course.

distribution/openrct2.d.ts
```diff
@@ -1214,7 +1225,7 @@ declare global {
         readonly remainingDistance: number;
 
         /**
-         * List of peep IDs ordered by seat.
+         * List of guest IDs ordered by seat.
          */
         peeps: Array<number | null>;
```
Broxzier, line 1230: This should be renamed to `guests` too. `peeps` should also still be around for now, but marked as deprecated.

src/openrct2/scripting/Plugin.cpp
```diff
@@ -161,4 +169,13 @@ void Plugin::CheckForLicence(const DukValue& dukLicence, std::string_view plugin
         log_error("Plugin %s does not specify a licence", std::string(pluginName).c_str());
 }
 
+int32_t Plugin::GetTargetAPIVersion() const
+{
+    if (_metadata.TargetApiVersion)
+        return *_metadata.TargetApiVersion;
+
+    // If not specified it defaults to 33.
```
Broxzier, line 177:
```suggestion
    // If not specified, default to 33 since that is the API version from before TargetAPIVersion was introduced.
```

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -56,8 +58,15 @@ namespace OpenRCT2::Scripting
                     case EntityType::Vehicle:
                         return "car";
                     case EntityType::Guest:
+                        if (targetApiVersion == 33)
```
Broxzier, line 61: Probably better to use `<=` in case someone (accidentally) specifies a lower API version.

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -56,8 +58,15 @@ namespace OpenRCT2::Scripting
...
+                            return "peep";
+                        else
+                            return "guest";
                     case EntityType::Staff:
-                        return "peep";
+                        if (targetApiVersion == 33)
```
Broxzier, line 66: Same as above.

Broxzier (changes requested):

src/openrct2/scripting/ScEntity.hpp
ZehMatt: But that would not specifically target a version then? If they mistarget it they will probably see errors anyway.

Broxzier: It used to be `peep` up until API > 33 anyway, so even if the specified version is < 33, it's technically correct. Targeting a version before 33 should definitely give the player an error.

Broxzier (approved):
Not tested, but changes look good.

ZehMatt:
I wrote this little script to validate the target api
```js
function main33() {
    console.log("main33");
    context.subscribe("interval.tick", () => {
        var peeps = map.getAllEntities("peep");
		for(var i = 0; i < peeps.length; i++)
		{
            var entType = peeps[i].type;
			if(entType == "staff" || entType == "guest")
			{
				console.error("This is not supposed to happen");
				break;
			}
		}
    });
}

function main34() {
    console.log("main34");
    context.subscribe("interval.tick", () => {
        var peeps = map.getAllEntities("staff");
		for(var i = 0; i < peeps.length; i++)
		{
            var entType = peeps[i].type;
			if(entType != "staff")
			{
				console.error("This is not supposed to happen");
				break;
			}
		}
    });
}

var apiVersion = 34;
registerPlugin({
    name: 'ApiTargetTest',
    version: '1.0',
    authors: ['ZehMatt'],
    type: 'remote',
    licence: 'MIT',
    targetApiVersion: apiVersion,
    main: apiVersion == 33 ? main33 : main34
});
```

distribution/openrct2.d.ts
Gymnasiast: Is it possible to move the `peep` entry to the bottom of the list? It would make the deprecation warning much less ambiguous.

src/openrct2/scripting/ScEntity.hpp
```diff
@@ -56,8 +58,15 @@ namespace OpenRCT2::Scripting
                     case EntityType::Vehicle:
                         return "car";
                     case EntityType::Guest:
+                        if (targetApiVersion <= 33)
```
Gymnasiast, line 61: Perhaps we should make a constant for 33, e.g. `API_VERSION_33_PEEP_DEPRECATION`.

Gymnasiast:
Could you also take a look at https://github.com/OpenRCT2/OpenRCT2/pull/15174#discussion_r684655473 ? For me, this is the last thing, I’ll merge it afterwards.

Gymnasiast (approved):
Will merge this once CIs are green.

PR #15188, opened 2021-08-09
Feature: Add ride index to TrackRemoveAction

CorySanin:
I only want "ride" to be an arg property for the GameActionEventArgs object in the plugin API. If there's a simpler and/or better way to accomplish that, I'm all ears 👂

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Feature: [#15084] [Plugin] Add "vehicle.crash" hook.
 - Feature: [#15143] Added a shortcut key for Giant Screenshot.
+- Feature: [#15188] Add ride index to TrackRemoveAction
 - Fix: [#13465] Creating a scenario based on a won save game results in a scenario that’s instantly won.
 - Fix: [#14316] Closing the Track Designs Manager window causes broken state.
 - Fix: [#14667] “Extreme Hawaiian Island” has unpurchaseable land tiles (original bug).
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -1930,7 +1930,7 @@ static void window_ride_construction_mouseup_demolish(rct_window* w)
     }
 
     auto trackRemoveAction = TrackRemoveAction(
-        _currentTrackPieceType, 0,
+        _currentRideIndex, _currentTrackPieceType, 0,
         { _currentTrackBegin.x, _currentTrackBegin.y, _currentTrackBegin.z, _currentTrackPieceDirection });
 
     trackRemoveAction.SetCallback([=](const GameAction* ga, const GameActions::Result* result) {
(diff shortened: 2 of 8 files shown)


duncanspumpkin:
What are you trying to achieve here? You're adding an extra parameter for no reason afaik

CorySanin:
Like I said, the goal is to be able to know the ride associated with a `trackremove` event in the API. For events like `trackplace` for instance, there's a `ride` property in the event's `args` object. The omission of this property is causing problems with one of my plugins. What would you recommend?

Gymnasiast:
The action _itself_ clearly shows you can fetch the ride id using `tileElement->AsTrack()->GetRideIndex()`. It would make far more sense to create a private method for getting the `TrackElement`, then using that for the visitor code, rather than requiring everyone to supply it for no good reason.

CorySanin:
I tried something like that but I was getting a seg fault 😕 I'll give it another shot though

duncanspumpkin:
Can you show us your plugin code you have and perhaps we can help.

IntelOrca:
One of the issues with this is that the caller is now assuming that it will only remove track pieces that have that ride ID - but the behaviour is still only to remove all pieces at the given location regardless of ride ID. Unless this behaviour were to change, I don't think the ride ID should be given to it. It would be much better to store the ride ID in the game action result instead.

CorySanin:
> 
> 
> Can you show us your plugin code you have and perhaps we can help.

[Here ya go](https://github.com/CorySanin/openrct2-ffa-plugin-suite/blob/master/src/ownership.ts#L14)

Some regulars on my server discovered that track pieces can be removed even if the ride doesn't belong to them. The ride just needs to be closed.

IntelOrca:
@CorySanin what specifically in your plugin / script was triggering a segfault?

CorySanin:
> 
> 
> @CorySanin what specifically in your plugin / script was triggering a segfault?

The segfault only occurred with the implementation @Gymnasiast suggested. In `AcceptParameters` (the "visitor" code), if I add the following:

```
    TileElement* tileElement = map_get_first_element_at(_origin);
    ride_id_t rideIndex = tileElement->AsTrack()->GetRideIndex();
    visitor.Visit("ride", rideIndex);
```

Calling `GetRideIndex()` causes a segfault.

Broxzier:
`map_get_first_element_at` gives back the first element on the tile. You need to iterate the map elements on that tile until you find the track piece. Since the first element is most likely a surface element, `tileElement->AsTrack()` will return `nullptr`.

Gymnasiast:
The code at https://github.com/OpenRCT2/OpenRCT2/pull/15188/files#diff-d00f029bb4c4cc8edd562c91a7bc8cf4b3206d82883204e0ddb83a92670b8947R264-R300 is responsible for getting the track element. You can spin that off into its own function, returning either the TileElement or nullptr if it cannot be found. Then you can call that function from both the existing location and from the place where you add the ride index to the visitor code.

duncanspumpkin:
I think we can close this then since you can just get the ride id from within the plug-in code (get track tile, get ride id)

PR #15194, opened 2021-08-10
Add new plugin api properties to entity, ride and park

Basssiiie:
Hey all,

On the side I've been experimenting with a new plugin that helps users figure out what impacts their park rating. The added properties are all related to calculating the park rating, though I've made a few settable as well (which might be useful for other plugin developers).

The changes:
1. Added `isInPark`, `isLost` and `lostCountdown` properties to `Guest`.
2. Added `downtime` property to `Ride`.
3. Added `casualtyPenalty` to `Park`.
4. Added overload to `getAllEntities` for `Car` and `Litter` since these interfaces exist too.
5. Auto-formatted `openrct2.d.ts` according to the rules in the project, which changed some incorrect whitespace too.
6. Updated my contribution description in `contributers.md` because someone mentioned `Misc.` is too plain and boring. 😛 
7. Incremented the plugin api version and added a changelog.

Love to hear if there's any feedback. 😄 If I should split the changes into multiple PR's, please let me know.

### DIFF
--- a/src/openrct2/scripting/ScRide.hpp
+++ b/src/openrct2/scripting/ScRide.hpp
@@ -642,6 +642,12 @@ namespace OpenRCT2::Scripting
             }
         }
 
+        uint8_t downtime_get() const
+        {
+            auto ride = GetRide();
+            return ride != nullptr ? ride->downtime : 0;
+        }
+
         Ride* GetRide() const
         {
             return get_ride(_rideId);
@@ -681,6 +687,7 @@ namespace OpenRCT2::Scripting
             dukglue_register_property(
                 ctx, &ScRide::inspectionInterval_get, &ScRide::inspectionInterval_set, "inspectionInterval");
             dukglue_register_property(ctx, &ScRide::value_get, &ScRide::value_set, "value");
+            dukglue_register_property(ctx, &ScRide::downtime_get, nullptr, "downtime");
         }
     };
 } // namespace OpenRCT2::Scripting
--- a/contributors.md
+++ b/contributors.md
@@ -91,7 +91,7 @@ The following people are not part of the development team, but have been contrib
 * Jim Verheijde (Jimver) - Make handymen less likely to get stuck in queue lines, misc.
 * Helio Batimarqui (batimarqui) - Misc.
 * Keith Stellyes (keithstellyes) - Misc.
-* Bas Cantrijn (Basssiiie) - Misc.
+* Bas Cantrijn (Basssiiie) - Various plugin additions, misc.
 * Adrian Zdanowicz (CookiePLMonster) - Misc.
 * Andrew Pratt (andrewpratt64) - Added api hook for vehicle crashes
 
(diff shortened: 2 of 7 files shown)


src/openrct2/scripting/ScRide.hpp
```diff
@@ -642,6 +641,21 @@ namespace OpenRCT2::Scripting
...
+            auto ride = GetRide();
+            return ride != nullptr ? ride->downtime : 0;
+        }
+        void downtime_set(uint8_t value)
+        {
+            ThrowIfGameStateNotMutable();
+            auto ride = GetRide();
+            if (ride != nullptr)
+            {
+                ride->downtime = std::min<uint8_t>(value, 100);
```
Broxzier, line 655: This value will be overwritten the next time `ride_breakdown_update` calculates its percentage based on the downtime history. Exposing the history to plugins and allowing modifications instead of setting a value that gets recalculated would be preferable.

IntelOrca: @Basssiiie not all properties need to have setters, or at least they could be added later if it isn't important right now.

Basssiiie: Ah didn't realize that it gets overridden. I'll remove this setter.

Broxzier:
> Auto-formatted openrct2.d.ts according to the rules in the project, which changed some incorrect whitespace too.

What did you use for this?

> If I should split the changes into multiple PR's, please let me know.

While it is nice to have each commit only make one change (one adding new fields, another fixing white space, etc.), it's fine to have them squashed for overseeable PRs like this one.

Basssiiie:
> What did you use for this?

@Broxzier Visual Studio's auto-format (Edit->Advanced->Format Document). If I'm not mistaken it takes the rules of the `.editorconfig` in the root of the project.

I always do this before I make a PR here, because I had CI failing on me before on code formatting in C++ files. :)

src/openrct2/scripting/ScRide.hpp
```diff
@@ -624,7 +624,6 @@ namespace OpenRCT2::Scripting
             }
             return ToDuk(ctx, nullptr);
         }
-
```
Gymnasiast, line 627: Restore this line.

Gymnasiast (approved):
I have _one_ little request (see the line note).

src/openrct2/scripting/ScRide.hpp
Basssiiie: @Gymnasiast Sure. The files are a bit inconsistent though, some getters and setters have no empty line separating them, some do. I removed this one because at first I though this one was the only inconsistent one, but I'm seeing more now.

But if you want it back I can add it back. :)

Broxzier: Other properties have no empty line between the getters and setters, probably done on purpose. This change makes the value property more consistent with the rest.

PR #15205, opened 2021-08-11
Fix #15184: Crash when hovering over water types in Object Selection

### LINKED ISSUE #15184: Crash when hovering over water types in scenario editor
**OS:** Kubuntu linux 20.04
**Version:** 0.3.4.1
**Commit/Build:** unknown, see #12909 

I can select any type of water I want, but if I hover the mouse over them too much, they'll crash.

Unnamed park 2021-08-07 15-15-56
This screen.

If I move the mouse over the water types, the game crashes.

**Steps to reproduce:**
1. Open the scenario editor to the object selection stage
2. Go to the water types tab
3. Repeatedly move the mouse over the list. I usually get a crash with less than 10 times moving the mouse up and down the list.

Gymnasiast:


src/openrct2/drawing/Image.cpp
```diff
@@ -164,8 +164,10 @@ static void FreeImageList(uint32_t baseImageId, uint32_t count)
     Guard::Assert(baseImageId >= BASE_IMAGE_ID, GUARD_LINE);
 
 #ifdef DEBUG
-    bool contains = AllocatedListRemove(baseImageId, count);
-    Guard::Assert(contains, GUARD_LINE);
+    if (!AllocatedListRemove(baseImageId, count))
+    {
+        log_error("Cannot unload %d items from offset %d", count, baseImageId);
```
janisozaur, line 169:
```suggestion
        log_error("Cannot unload %u items from offset %u", count, baseImageId);
```

janisozaur (changes requested):

src/openrct2/drawing/Image.cpp
```diff
@@ -164,8 +164,10 @@ static void FreeImageList(uint32_t baseImageId, uint32_t count)
     Guard::Assert(baseImageId >= BASE_IMAGE_ID, GUARD_LINE);
 
 #ifdef DEBUG
```
janisozaur, line 166: That's also silly to do, as it is often defined - check for the value instead

Broxzier: `DEBUG` and `NDEBUG` are often defined with no value, using `defined(...)` or `ifdef` is fine. The values are never checked anywhere else in our codebase.

Issue #13814, opened 2021-01-12
Refactor window to class: Viewport


IntelOrca:
See #13737 for more details.

seifer7:
Working on this.

PR #15217, opened 2021-08-14
Issue #13814: Refactor window to class: Viewport

seifer7:
Like the title says.

Issue #13814
Refactor window to class: Viewport

### DIFF
--- a/src/openrct2-ui/windows/Viewport.cpp
+++ b/src/openrct2-ui/windows/Viewport.cpp
@@ -55,22 +55,22 @@ static rct_widget window_viewport_widgets[] =
 class ViewportWindow final : public Window
 {
 private:
-    int32_t _viewportWindowNumber = 1;
-
-    void GetFreeViewportNumber(rct_window* w)
+    void GetFreeViewportNumber()
     {
-        if (w != nullptr && w->classification == WC_VIEWPORT)
-        {
-            if (w->number >= _viewportWindowNumber)
-                _viewportWindowNumber = w->number + 1;
-        }
+        number = 1;
+        window_visit_each([&](rct_window* w) {
+            if (w != nullptr && w != this && w->classification == WC_VIEWPORT)
+            {
+                if (w->number >= number)
+                    number = w->number + 1;
+            }
+        });
     }
 
 public:
     void OnOpen() override
     {
-        window_visit_each([&](rct_window* w) { GetFreeViewportNumber(w); });
-        number = _viewportWindowNumber;
+        GetFreeViewportNumber();
 
         widgets = window_viewport_widgets;
         enabled_widgets = (1ULL << WIDX_CLOSE) | (1ULL << WIDX_ZOOM_IN) | (1ULL << WIDX_ZOOM_OUT) | (1ULL << WIDX_LOCATE);
@@ -84,7 +84,7 @@ public:
             return;
         }
 
-        auto mainWindow = window_get_main();
+        auto* mainWindow = window_get_main();
         if (mainWindow != nullptr)
         {
             rct_viewport* mainViewport = mainWindow->viewport;
@@ -103,7 +103,7 @@ public:
 
     void OnUpdate() override
     {
-        auto mainWindow = window_get_main();
+        auto* mainWindow = window_get_main();
         if (mainWindow == nullptr)
             return;
 
@@ -139,7 +139,7 @@ public:
                 }
                 break;
             case WIDX_LOCATE:
-                auto mainWindow = window_get_main();
+                auto* mainWindow = window_get_main();
                 if (mainWindow != nullptr)
                 {
                     auto info = get_map_coordinates_from_pos(
--- a/src/openrct2-ui/windows/Viewport.cpp
+++ b/src/openrct2-ui/windows/Viewport.cpp
@@ -57,7 +57,7 @@ class ViewportWindow final : public Window
 private:
     int32_t _viewportWindowNumber = 1;
 
-    void get_free_viewport_number(rct_window* w)
+    void GetFreeViewportNumber(rct_window* w)
     {
         if (w != nullptr && w->classification == WC_VIEWPORT)
         {
@@ -69,7 +69,7 @@ private:
 public:
     void OnOpen() override
     {
-        window_visit_each([&](rct_window* w) { get_free_viewport_number(w); });
+        window_visit_each([&](rct_window* w) { GetFreeViewportNumber(w); });
         number = _viewportWindowNumber;
 
         widgets = window_viewport_widgets;
--- a/src/openrct2-ui/windows/Viewport.cpp
+++ b/src/openrct2-ui/windows/Viewport.cpp
@@ -10,192 +10,217 @@
 #include <openrct2-ui/interface/Viewport.h>
 #include <openrct2-ui/interface/Widget.h>
 #include <openrct2-ui/windows/Window.h>
+#include <openrct2/Context.h>
 #include <openrct2/Game.h>
 #include <openrct2/audio/audio.h>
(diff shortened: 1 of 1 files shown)


src/openrct2-ui/windows/Viewport.cpp
```diff
@@ -10,192 +10,217 @@
...
-    events.update = &window_viewport_update;
-    events.invalidate = &window_viewport_invalidate;
-    events.paint = &window_viewport_paint;
-});
-// clang-format on
+private:
+    int32_t _viewportWindowNumber = 1;
 
-static int32_t _viewportNumber = 1;
+    void get_free_viewport_number(rct_window* w)
```
duncanspumpkin, line 60: UpperCamelCase

src/openrct2-ui/windows/Viewport.cpp
```diff
@@ -10,192 +10,217 @@
...
-        int32_t y = mainViewport->viewPos.y + (mainViewport->view_height / 2);
-        w->savedViewPos = { x - (w->viewport->view_width / 2), y - (w->viewport->view_height / 2) };
+        window_visit_each([&](rct_window* w) { get_free_viewport_number(w); });
+        number = _viewportWindowNumber;
+
+        widgets = window_viewport_widgets;
+        enabled_widgets = (1ULL << WIDX_CLOSE) | (1ULL << WIDX_ZOOM_IN) | (1ULL << WIDX_ZOOM_OUT) | (1ULL << WIDX_LOCATE);
+
+        // Create viewport
+        viewport_create(this, windowPos, width, height, 0, TileCoordsXYZ(128, 128, 0).ToCoordsXYZ(), 1, SPRITE_INDEX_NULL);
```
IntelOrca, line 79: `viewport_create` can fail, in which case, `viewport` will be `nullptr`. You should there for check `viewport` is not `nullptr` before doing anything with it. This obviously improves on the original code which didn't check.

src/openrct2-ui/windows/Viewport.cpp
```diff
@@ -10,192 +10,217 @@
...
     }
 
-    w->viewport->flags |= VIEWPORT_FLAG_SOUND_ON;
+    void OnUpdate() override
+    {
+        rct_window* mainWindow;
 
-    return w;
-}
+        mainWindow = window_get_main();
```
IntelOrca, line 101: As you improve this code, you can start reducing the number of unnecessary lines which were reminiscent of when the codebase was in C. So in this example you can just write `auto mainWindow = window_get_main();`

seifer7: Which is better / preferred?
1. `rct_window* mainWindow = window_get_main();`
or
2. `auto mainWindow = window_get_main();`

seifer7: I went with 2 `auto mainWindow = window_get_main();`

src/openrct2-ui/windows/Viewport.cpp
```diff
@@ -10,192 +10,220 @@
...
-    w->widgets = window_viewport_widgets;
-    w->enabled_widgets = (1ULL << WIDX_CLOSE) | (1ULL << WIDX_ZOOM_IN) | (1ULL << WIDX_ZOOM_OUT) | (1ULL << WIDX_LOCATE);
-    w->number = _viewportNumber++;
-
-    // Create viewport
-    viewport_create(w, w->windowPos, w->width, w->height, 0, TileCoordsXYZ(128, 128, 0).ToCoordsXYZ(), 1, SPRITE_INDEX_NULL);
-    rct_window* mainWindow = window_get_main();
-    if (mainWindow != nullptr)
+private:
+    int32_t _viewportWindowNumber = 1;
```
duncanspumpkin, line 58: I realise this was in the original window but what is the point of this variable. It's identical to the window number afaik and therefore there is no point in it @IntelOrca Could you confirm what the idea with it was. All the vehicle, peep, banner windows just use the window number to be unique.

seifer7: I only created this var so I could increment it until I found the highest unused number. I can remove this var and instead work directly with the window->number var.

seifer7: Just seemed cleaner (to me) to only touch the window->number once, ie. after I'd found what to set it to.
Probably doesn't matter though.

ZehMatt: Theres is no difference between the two variables access wise

duncanspumpkin: Ahh you don't need it you can just pass by reference into the lambda.
```
window_visit_each([&](rct_window* w) { GetFreeViewportNumber(w, number); });
```
Then just modify get free viewport number to take the number by reference.

src/openrct2-ui/windows/Viewport.cpp
```diff
@@ -10,192 +10,220 @@
...
     }
 
-    w->viewport->flags |= VIEWPORT_FLAG_SOUND_ON;
-
-    return w;
-}
+public:
+    void OnOpen() override
+    {
+        window_visit_each([&](rct_window* w) { GetFreeViewportNumber(w); });
```
ZehMatt, line 72: This is somewhat confusing, this line should be replaced with ```GetFreeViewportNumber();``` where the function takes care of it.

ZehMatt (changes requested):

seifer7:
**Identified a bug that also exists before this pull request.**

When you rotate the main viewport, the extra viewport window(s) center on the wrong location.

src/openrct2-ui/windows/Viewport.cpp
IntelOrca: I think it was so that it displayed the number in the title of the window, and also so each viewport window had a unique window number as they all have identical class.

src/openrct2-ui/windows/Viewport.cpp
seifer7: Done

src/openrct2-ui/windows/Viewport.cpp
seifer7: I've removed the _viewportWindowNumber variable and just use the window->number to build the title of the window.

src/openrct2-ui/windows/Viewport.cpp
seifer7: Fixed.

src/openrct2-ui/windows/Viewport.cpp
seifer7: Added a null check and it will Close the window and display an error tooltip to say something went wrong.

IntelOrca (approved):

ZehMatt (approved):

seifer7:
Should this be added to the changelog? I noticed other Window refactors have not been.

ZehMatt:
Such changes are not visible to the end user so its not worth adding it, for the code history we have git.

src/openrct2-ui/windows/Viewport.cpp
```diff
@@ -10,192 +10,220 @@
...
+        if (mainWindow == nullptr)
+            return;
+
+        if (viewport != nullptr && viewport->flags != mainWindow->viewport->flags)
+        {
+            viewport->flags = mainWindow->viewport->flags;
+            Invalidate();
+        }
+
+        // Not sure how to invalidate part of the viewport that has changed, this will have to do for now
```
duncanspumpkin, line 116: What is this comment for? If it's about actual changes within the viewport you don't need to do anything that's already accounted for

seifer7: That was left from the original, before the refactor. I left it there in case the note is still relevant.

PR #15222, opened 2021-08-15
Use ObjectEntryIndex for entrance style

Gymnasiast:
Ported from NSF.

### DIFF
--- a/src/openrct2/actions/RideSetAppearanceAction.h
+++ b/src/openrct2/actions/RideSetAppearanceAction.h
@@ -29,12 +29,12 @@ DEFINE_GAME_ACTION(RideSetAppearanceAction, GameCommand::SetRideAppearance, Game
 private:
     NetworkRideId_t _rideIndex{ RIDE_ID_NULL };
     RideSetAppearanceType _type{};
-    uint8_t _value{};
+    uint16_t _value{};
     uint32_t _index{};
 
 public:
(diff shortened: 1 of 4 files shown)


src/openrct2/actions/RideSetAppearanceAction.h
```diff
@@ -29,7 +29,7 @@ DEFINE_GAME_ACTION(RideSetAppearanceAction, GameCommand::SetRideAppearance, Game
 private:
     NetworkRideId_t _rideIndex{ RIDE_ID_NULL };
     RideSetAppearanceType _type{};
-    uint8_t _value{};
+    uint16_t _value{};
```
ZehMatt, line 32: Is this ObjectEntryIndex? Also the type was not changed in the constructor.

Gymnasiast: _Sometimes_ it’s ObjectEntryIndex. `RideSetAppearanceAction` is used for several different settings.

ZehMatt: I see, either way you should probably changed uint8 to uint16 of the constructor parameter.

Gymnasiast: That was an oversight. I have now adjusted them.

ZehMatt (approved):

Issue #15225, opened 2021-08-15
Some senarios show the incorrect park name and description when loaded


RoelerCoaster:
### Operating System

Windows 10, 64 bit

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (767dbfe on develop) provided by GitHub

### Describe the issue

When loading some scenarios scenario selection window seems to change the park name to "Six Flags Magic Mountain" and when the park is loaded the description of the scenario "Six Flags Magic Mountain" is shown instead of the correct one. The objective remains correct. Also, the park name and the description is correct in the scenario selection window, the incorrect text is only shown once the scenario is loaded.

It doesn't seem to be happening on all of the parks. I haven't tested all of the scenarios, but some scenarios on which this seems to happen:
- Africa - Victoria falls
- Asia - Maharaja Palace
- South America - Rain Forest Plateau
- Rock 'n Roll - Flower Power
- Prehistoric - Stone Age
- Build your own Six Flags Magic mountain (only the description is incorrect)

Some examples where it doesn't seem to happen:
- As far as I can tell: any of the base RCT2 beginner, intermediate or expert scenario's 
- Asia - Great Wall of China Tourism Enhancement
- North America - Rollercoaster Heaven

I cannot seem to deduce a pattern from this which parks are affected and which are not.

### Steps to reproduce

1. Start a new game using one of the problematic scenarios, for example Africa - Victoria falls
2. Observe the text in the park-window that opens
3. Title/park name is incorrect
4. Description is incorrect
5. Objective is correct

### Attachments

![Six Flags Magic Mountain 2021-08-15 22-19-13](https://user-images.githubusercontent.com/5968886/129491628-42c351cf-c297-42c1-9f3d-f2145f7cb070.png)

seifer7:
I've identified the bug and will open a pull request shortly.

PR #15229, opened 2021-08-16
Issue #15225: Some scenarios show incorrect name/description in-game.

seifer7:
As some know, a number of the Scenarios created for the expansion packs by Frontier used the internal name "Six Flags Magic Mountain," why? I doubt God even knows. 

The S6Importer/SawyerChunkReader correctly decodes the scenario name and description and stores it under the `rct_s6_info` struct (`info.name` and `info.details`.)

The problem is that on lines 369-370 the name and description are replaced with the scenario internal name and description - which is sometimes incorrect - specifically a number of scenarios by Frontier in the expansion packs.

### DIFF
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -366,8 +366,6 @@ public:
         // pad_01358842
         ImportResearchList();
         gMapBaseZ = _s6.map_base_z;
-        gScenarioName = std::string_view(_s6.scenario_name, sizeof(_s6.scenario_name));
-        gScenarioDetails = std::string_view(_s6.scenario_description, sizeof(_s6.scenario_description));
         gBankLoanInterestRate = _s6.current_interest_rate;
         // pad_0135934B
         // Preserve compatibility with vanilla RCT2's save format.


seifer7:
Note:
`gScenarioName` and `gScenarioDetails` are already (correctly) set on lines 212-221 in the same function.

https://github.com/OpenRCT2/OpenRCT2/blob/0988496550ebfd9685e9a9845263efe5591209f0/src/openrct2/rct2/S6Importer.cpp#L212-L221

Broxzier:
Was this issue introduced since the last release? If not, a changelog entry is in order.

Gymnasiast:
I wasn’t quite sure, but since some refactoring went on in this area after v0.3.4.1 (most notably: removal of gS6Info), I figured it was introduced then.

seifer7:
The latest `release` version does not have this bug.

The bug was introduced in commit 5f853df1fde09aa0e98ce2895d945831784c6e47, but not in the way I expected.
The lines I removed in this pull request to correct the issue have been present since before that commit - so how were they the cause?

It turns out that while these lines have always(?) been present and storing the incorrect Scenario Name/Details...
https://github.com/OpenRCT2/OpenRCT2/blob/3503a433dc277f740f6dc1106b99b9df20fb6b1b/src/openrct2/rct2/S6Importer.cpp#L370-L371
... those variables where then incorrectly overwritten in another file (with the correct info):
https://github.com/OpenRCT2/OpenRCT2/blob/3503a433dc277f740f6dc1106b99b9df20fb6b1b/src/openrct2/scenario/Scenario.cpp#L109-L116


The `Scenario.cpp` file was corrected in 5f853df1fde09aa0e98ce2895d945831784c6e47...
```diff
    gCash = gInitialCash;

-    gScenarioDetails = std::string_view(gS6Info.details, 256);
-    gScenarioName = std::string_view(gS6Info.name, 64);

    {
        utf8 normalisedName[64];
-        ScenarioSources::NormaliseName(normalisedName, sizeof(normalisedName), gS6Info.name);
+        ScenarioSources::NormaliseName(normalisedName, sizeof(normalisedName), gScenarioName.c_str());

        rct_string_id localisedStringIds[3];
```

So with `Scenario.cpp` correctly using the already existing `gScenarioName` variable, it became evident that the value inside `gScenarioName` was wrong.

PR #15231, opened 2021-08-16
Refactor out capturing of window pointers with GA callbacks

### LINKED ISSUE #15169: Refactor game action callbacks to not capture pointers
There might be a couple of potential crashes dereferencing invalid memory when it comes to game action callbacks. In multiplayer mode those callbacks will fire after the server acknowledged the action so there is a delay, the user could have closed the window before the callback fires. 
Example here:
https://github.com/OpenRCT2/OpenRCT2/blob/3503a433dc277f740f6dc1106b99b9df20fb6b1b/src/openrct2-ui/windows/RideConstruction.cpp#L1943

The lambda captures ```w``` which is a windows pointer, to mitigate such problem the lambda can capture the ride id so that the window can be retrieved safely in the within the callback via the id. There are potentially more lines like this, all the game actions with callbacks should be reviewed and changed accordingly.

ZehMatt:
Closes #15169

### DIFF
--- a/src/openrct2-ui/windows/Player.cpp
+++ b/src/openrct2-ui/windows/Player.cpp
@@ -279,7 +279,7 @@ void window_player_overview_dropdown(rct_window* w, rct_widgetindex widgetIndex,
     const auto groupId = network_get_group_id(dropdownIndex);
     const auto windowHandle = std::make_pair(w->classification, w->number);
     auto playerSetGroupAction = PlayerSetGroupAction(playerId, groupId);
-    playerSetGroupAction.SetCallback([=](const GameAction* ga, const GameActions::Result* result) {
+    playerSetGroupAction.SetCallback([windowHandle](const GameAction* ga, const GameActions::Result* result) {
         if (result->Error == GameActions::Status::Ok)
         {
             window_invalidate_by_number(windowHandle.first, windowHandle.second);
--- a/src/openrct2-ui/windows/Player.cpp
+++ b/src/openrct2-ui/windows/Player.cpp
@@ -23,6 +23,7 @@
 #include <openrct2/network/network.h>
 #include <openrct2/sprites.h>
 #include <openrct2/util/Util.h>
+#include <utility>
 
 // clang-format off
 enum WINDOW_PLAYER_PAGE {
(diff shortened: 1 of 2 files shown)


src/openrct2-ui/windows/Player.cpp
```diff
@@ -265,21 +266,23 @@ void window_player_overview_mouse_down(rct_window* w, rct_widgetindex widgetInde
...
     if (dropdownIndex == -1)
     {
         return;
     }
-    int32_t group = network_get_group_id(dropdownIndex);
-    auto playerSetGroupAction = PlayerSetGroupAction(w->number, group);
+    const auto groupId = network_get_group_id(dropdownIndex);
+    const auto windowHandle = std::make_pair(w->classification, w->number);
+    auto playerSetGroupAction = PlayerSetGroupAction(playerId, groupId);
     playerSetGroupAction.SetCallback([=](const GameAction* ga, const GameActions::Result* result) {
```
tupaschoal, line 282: Maybe we should change the capture so that we explicitly only get `windowHandle`
```suggestion
    playerSetGroupAction.SetCallback([windowHandle](const GameAction* ga, const GameActions::Result* result) {
```

That would make it harder for us to sneak the usage of the pointer again.

tupaschoal (approved):

src/openrct2-ui/windows/Player.cpp
ZehMatt: True, ill change it.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1933,14 +1933,15 @@ static void window_ride_construction_mouseup_demolish(rct_window* w)
         _currentTrackPieceType, 0,
         { _currentTrackBegin.x, _currentTrackBegin.y, _currentTrackBegin.z, _currentTrackPieceDirection });
 
+    const auto rideId = w->number;
     trackRemoveAction.SetCallback([=](const GameAction* ga, const GameActions::Result* result) {
```
tupaschoal, line 1937: I think the previous issue also applies here, though it seems like this lambda has a lot more couplings.

ZehMatt: Quite a few arguments, so I think we'll leave that one. Normally with reviews this should sneak back in.

tupaschoal: I understand. Then I think this is good to go.

Broxzier (approved):

PR #15232, opened 2021-08-17
Move network instance to Context

ZehMatt:
This PR also introduces a System model that other classes can use, more refactor work is needed to get this right but this will also help eliminate global variables since all systems will always have a reference to IContext and IContext always has the full game state.

### DIFF
--- a/src/openrct2/Context.cpp
+++ b/src/openrct2/Context.cpp
@@ -654,12 +654,12 @@ namespace OpenRCT2
                 bool sendMap = false;
                 if (info.Type == FILE_TYPE::SAVED_GAME)
                 {
-                    if (network_get_mode() == NETWORK_MODE_CLIENT)
+                    if (_network.GetMode() == NETWORK_MODE_CLIENT)
                     {
-                        network_close();
+                        _network.Close();
                     }
                     game_load_init();
-                    if (network_get_mode() == NETWORK_MODE_SERVER)
+                    if (_network.GetMode() == NETWORK_MODE_SERVER)
                     {
                         sendMap = true;
                     }
@@ -667,13 +667,13 @@ namespace OpenRCT2
                 else
                 {
                     scenario_begin();
-                    if (network_get_mode() == NETWORK_MODE_SERVER)
+                    if (_network.GetMode() == NETWORK_MODE_SERVER)
                     {
                         sendMap = true;
                     }
-                    if (network_get_mode() == NETWORK_MODE_CLIENT)
+                    if (_network.GetMode() == NETWORK_MODE_CLIENT)
                     {
-                        network_close();
+                        _network.Close();
                     }
                 }
                 // This ensures that the newly loaded save reflects the user's
@@ -681,10 +681,10 @@ namespace OpenRCT2
                 peep_update_names(gConfigGeneral.show_real_names_of_guests);
                 if (sendMap)
                 {
-                    network_send_map();
+                    _network.Server_Send_MAP();
                 }
 #ifdef USE_BREAKPAD
-                if (network_get_mode() == NETWORK_MODE_NONE)
+                if (_network.GetMode() == NETWORK_MODE_NONE)
                 {
                     start_silent_record();
                 }
@@ -884,13 +884,13 @@ namespace OpenRCT2
 
                         if (String::IsNullOrEmpty(gCustomPassword))
                         {
-                            network_set_password(gConfigNetwork.default_password.c_str());
+                            _network.SetPassword(gConfigNetwork.default_password.c_str());
                         }
                         else
                         {
-                            network_set_password(gCustomPassword);
+                            _network.SetPassword(gCustomPassword);
                         }
-                        network_begin_server(gNetworkStartPort, gNetworkStartAddress);
+                        _network.BeginServer(gNetworkStartPort, gNetworkStartAddress);
                     }
                     else
 #endif // DISABLE_NETWORK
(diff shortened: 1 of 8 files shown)


src/openrct2/Context.h
```diff
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
```
Gymnasiast, line 1: BOM?

Broxzier: Yup, BOM has been removed here.

src/openrct2/Context.cpp
```diff
@@ -108,6 +109,9 @@ namespace OpenRCT2
 #ifdef ENABLE_SCRIPTING
         ScriptEngine _scriptEngine;
 #endif
+#ifndef DISABLE_NETWORK
+        NetworkBase _network;
+#endif
```
duncanspumpkin, line 114: Can't we just have some null network so this isn't required

ZehMatt: Its on my list for sure, I never got quite to the point of splitting up the three classes and adding a dummy.

src/openrct2/Context.cpp
```diff
@@ -164,7 +171,7 @@ namespace OpenRCT2
             //       If objects use GetContext() in their destructor things won't go well.
 
             GameActions::ClearQueue();
-            network_close();
+            _network.Close();
```
tupaschoal, line 175: I don't understand how this change is  tied to `https://github.com/OpenRCT2/OpenRCT2/pull/15232/commits/40313b1ea3182adbb4586e9c8ca13032a19a0350`

src/openrct2/network/NetworkBase.h
```diff
@@ -12,10 +12,17 @@
...
 #ifndef DISABLE_NETWORK
 
+namespace OpenRCT2
+{
+    struct IContext;
+}
+
 class NetworkBase
 {
+    OpenRCT2::IContext& _context;
```
tupaschoal: On https://github.com/OpenRCT2/OpenRCT2/pull/15232/commits/40313b1ea3182adbb4586e9c8ca13032a19a0350 shouldn't this be down there along with the other variable declarations?

src/openrct2/Context.cpp
```diff
@@ -654,37 +654,37 @@ namespace OpenRCT2
...
+                        _network.Close();
                     }
                 }
                 // This ensures that the newly loaded save reflects the user's
                 // 'show real names of guests' option, now that it's a global setting
                 peep_update_names(gConfigGeneral.show_real_names_of_guests);
                 if (sendMap)
                 {
-                    network_send_map();
+                    _network.Server_Send_MAP();
```
tupaschoal, line 695: PS: Something's off with whatever the naming standard was for this `Server_Send_MAP` :P

src/openrct2/System.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+        // Called after a tick.
+        virtual void PostTick()
+        {
+        }
+
+        // Called every frame.
+        virtual void Update()
+        {
+        }
```
tupaschoal, line 47: (On `https://github.com/OpenRCT2/OpenRCT2/pull/15232/commits/1cfc933a59f37355ed8f5f614daf7e9429cb65b7`)

Does it make sense for those to have empty implementations or should they be abstract?

src/openrct2/network/NetworkBase.cpp
```diff
@@ -1270,14 +1270,14 @@ void NetworkBase::Server_Send_OBJECTS_LIST(
     }
 }
 
-void NetworkBase::Server_Send_SCRIPTS(NetworkConnection& connection) const
+void NetworkBase::Server_Send_SCRIPTS(NetworkConnection& connection)
```
tupaschoal, line 1273: (On https://github.com/OpenRCT2/OpenRCT2/pull/15232/commits/b3bb2f5d80765bf383fd60d5bd5b2c79fd0dcbba)
Why was the constness droppped?

Maybe because it isn't called the `const` `GetContext()`?

src/openrct2/network/NetworkBase.h
```diff
@@ -32,7 +31,8 @@ class NetworkBase
     bool Init();
     void Close();
     uint32_t GetServerTick();
-    void Update();
+    // FIXME: This is currently the wrong function to override in System, will be refactored later.
```
tupaschoal, line 34: (On https://github.com/OpenRCT2/OpenRCT2/pull/15232/commits/b3bb2f5d80765bf383fd60d5bd5b2c79fd0dcbba)

Will this be fixed still in this PR or later? If later, it would help to detail why it's the wrong function

src/openrct2/Context.cpp
tupaschoal: Dummy would be great for not having all those guards, indeed :)

src/openrct2/System.hpp
ZehMatt: Not all systems will need all of those events  thats why they are empty, this is future stuff.

src/openrct2/network/NetworkBase.cpp
ZehMatt: Yeah I had issues getting this to const so I had to drop it for now, not happy about it but the other way would be much worse.

src/openrct2/network/NetworkBase.h
ZehMatt: Needs to be fixed later, the naming is inconsistent when it comes to Update/Tick. Update will be called every frame where Tick is fixed at 40hz. Thats why its wrong right now, its not an actual Update.

src/openrct2/network/NetworkBase.h
ZehMatt: This was there temporarily until I had the system base.

Gymnasiast (approved):

PR #15244, opened 2021-08-19
Use [[nodiscard]] for allocator and RAII functions

Broxzier:
This PR adds the `[[nodiscard]]`-attribute to a bunch of functions returning `std::vector`, `std::unique_ptr` and `std::shared_ptr`, as well as some neighbouring functions and many global functions, because their results should never be ignored.

I should mention that `-Wunused-result` exists for g++ and should already show warnings for all functions regardless of this attribute, but it doesn't seem to catch all cases and would only result in errors when using that compiler.


src/openrct2/Context.h
```diff
@@ -1,4 +1,4 @@
-﻿/*****************************************************************************
+/*****************************************************************************
```
Gymnasiast, line 1: BOM?

Broxzier: Most likely yeah, must've happened automatically. Probably a good idea to normalize them in another PR.

Gymnasiast: Indeed.

For this PR, can you revert it? The PR otherwise looks good to me.

Broxzier: Will do, fixing the tests right now.

Broxzier:
CI fails on a bunch of `error: attribute ignored [-Werror=attributes]` errors. From what I can find, this has to do with the GCC compiler being outdated? In the logs, I see that GCC 7.5.0 and 10.2.0, Clang 5.0.1, and AppleClang 12.0.0.12000032 are erroring on this attribute. I'm not sure how to go about updating those.

ZehMatt:
While I'm not a huge fan of of you could define a macro like NO_DISCARD or something which checks the compiler version, but upgrading the CI would be also a nice thing.

Gymnasiast:
Perhaps a stupid question, but will `[[nodiscard]] static` instead of `static [[nodiscard]]` work?

src/openrct2/Context.h
Broxzier: The last commit reintroduces this change again, done by Visual Studio. Will fix it after CI is done with this commit.

Broxzier:
Looks like CIs are passing now. Nice suggestions, @Gymnasiast! Odd that this is requires since there is no mention of its placement being before other modifiers on [the reference page](https://en.cppreference.com/w/cpp/language/attributes/nodiscard).

PR #15246, opened 2021-08-20
Fix refactor error in Paint.cpp assertions

Gymnasiast:
Before the refactor, the assertion read like this:
```
assert(static_cast<uint16_t>(bound_box_length_x) == static_cast<int16_t>(bound_box_length_x));
```
which meant that bound_box_length_x = 0 was valid. The refactor (likely accidentally) changed this, which causes assertions to get hit within 2 seconds after opening on my machine.

### DIFF
--- a/src/openrct2/paint/Paint.cpp
+++ b/src/openrct2/paint/Paint.cpp
@@ -801,8 +801,8 @@ paint_struct* PaintAddImageAsParent(
     int32_t bound_box_length_y, int32_t bound_box_length_z, int32_t z_offset, int32_t bound_box_offset_x,
     int32_t bound_box_offset_y, int32_t bound_box_offset_z)
 {
-    assert(bound_box_length_x > 0);
-    assert(bound_box_length_y > 0);
+    assert(bound_box_length_x >= 0);
+    assert(bound_box_length_y >= 0);
 
     session->LastPS = nullptr;
     session->LastAttachedPS = nullptr;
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
Are you able to find the call that is sending a 0 for the size?

Gymnasiast:
I have the backtrace [here](https://pastebin.com/aSkCGJkN), does that contain anything useful?

duncanspumpkin:
https://github.com/OpenRCT2/OpenRCT2/blob/39fcc5d0787dd5c1f68db2a467e87633b38ca01b/src/openrct2/ride/VehiclePaint.cpp#L3032
That looks a bit suspect. I'll try check against the assembly

duncanspumpkin:
Assembly has a bug for this bit of code and ends up writing a clobered register to the size. Size would end up being x = vehicleObj->effect_visual and y would be 0. Clearly that makes no sense so is wrong. I think what is meant to be passed to each of the splash effects is the vehicle bound boxes but that also doesn't quite make sense as these are pieces of water and they could be a completely different size to the vehicle.

src/openrct2/paint/Paint.cpp
```diff
@@ -863,8 +863,8 @@ paint_struct* PaintAddImageAsChild(
     int32_t bound_box_length_y, int32_t bound_box_length_z, int32_t z_offset, int32_t bound_box_offset_x,
     int32_t bound_box_offset_y, int32_t bound_box_offset_z)
 {
-    assert(bound_box_length_x > 0);
-    assert(bound_box_length_y > 0);
+    assert(bound_box_length_x >= 0);
+    assert(bound_box_length_y >= 0);
```
duncanspumpkin, line 867: The asserts are not required anymore. Can be removed.

Gymnasiast: Ok, I’ll remove them. Btw, do you happen to know why there were introduced?

duncanspumpkin (changes requested):
Had a bit of a look into this and asserts on this are not required. The original assert was because there was some calls that passed incorrect signed values but they no longer exist. Therefore all the asserts can be removed.

duncanspumpkin:
Also @ZehMatt pointed out that children don't actually use the bounding box size so it's especially pointless. But unfortunately if the child doesn't have a parent it becomes a parent so not too sure what to do.

ZehMatt:
> Also @ZehMatt pointed out that children don't actually use the bounding box size so it's especially pointless. But unfortunately if the child doesn't have a parent it becomes a parent so not too sure what to do.

Have u actually seen a case where an image is supposed to be a child but has no active parent?

duncanspumpkin (approved):

PR #15249, opened 2021-08-20
Refactor use of std::unique_ptr

Broxzier:
The first commit removes the unnecessary constructor calls and improves flexibility in case of typename changes.
All cases, except for the one in Context.cpp, are temporaries.

The second commit constructs arrays using `std::make_unique` instead of using new and passing it to the constructor.

The third commit refactors `LanguagePack` to return `std::unique_ptr` instead of raw pointers.

### DIFF
--- a/src/openrct2/CmdlineSprite.cpp
+++ b/src/openrct2/CmdlineSprite.cpp
@@ -185,7 +185,7 @@ bool SpriteFile::Save(const utf8* path)
 static bool SpriteImageExport(const rct_g1_element& spriteElement, const char* outPath)
 {
     const auto pixelBufferSize = spriteElement.width * spriteElement.height;
-    std::unique_ptr<uint8_t[]> pixelBuffer(new uint8_t[pixelBufferSize]);
+    auto pixelBuffer = std::make_unique<uint8_t[]>(pixelBufferSize);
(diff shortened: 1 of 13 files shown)


ZehMatt (approved):
Good stuff

src/openrct2/CmdlineSprite.cpp
```diff
@@ -185,7 +185,7 @@ bool SpriteFile::Save(const utf8* path)
 static bool SpriteImageExport(const rct_g1_element& spriteElement, const char* outPath)
 {
     const auto pixelBufferSize = spriteElement.width * spriteElement.height;
-    std::unique_ptr<uint8_t[]> pixelBuffer(new uint8_t[pixelBufferSize]);
+    std::unique_ptr<uint8_t[]> pixelBuffer = std::make_unique<uint8_t[]>(pixelBufferSize);
```
tupaschoal, line 188: Maybe use `auto` on the left hand side now? (No idea why GH is not allowing me to suggest it)

(Applies to all of https://github.com/OpenRCT2/OpenRCT2/pull/15249/commits/dee05d7edb25630c5fb69cfd4ad33d5f4ae3ed35)

tupaschoal (approved):

PR #15251, opened 2021-08-20
Update include guards

Broxzier:
- Update from C-style to pragma once
- Add missing include guard to ui.h


tupaschoal (approved):
Nice! Does this remove the remainder of the include guards that were not `#pragma once`?

Broxzier:
@tupaschoal I think so, but I'm not 100% sure that I got all of them.

PR #15252, opened 2021-08-21
Streamline animation vars

spacek531:
renames some vars to be more descriptive
refactors steam spawning logic
shouldn't require network bump

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -7276,10 +7276,10 @@ void Vehicle::UpdateSpinningCar()
 void Vehicle::UpdateAdditionalAnimation()
 {
     uint8_t targetFrame{};
-    uint8_t steamFrame{};
+    uint8_t curFrame{};
     uint32_t eax{};
 
-    uint32_t* curAnimation_state = reinterpret_cast<uint32_t*>(&animationState);
+    uint32_t* curAnimationState = reinterpret_cast<uint32_t*>(&animationState);
     auto vehicleEntry = Entry();
     if (vehicleEntry == nullptr)
     {
@@ -7288,15 +7288,15 @@ void Vehicle::UpdateAdditionalAnimation()
     switch (vehicleEntry->animation)
     {
         case VEHICLE_ENTRY_ANIMATION_MINITURE_RAILWAY_LOCOMOTIVE: // loc_6D652B
-            *curAnimation_state += _vehicleVelocityF64E08;
-            targetFrame = (*curAnimation_state >> 20) & 3;
+            *curAnimationState += _vehicleVelocityF64E08;
+            targetFrame = (*curAnimationState >> 20) & 3;
             if (animation_frame != targetFrame)
             {
-                steamFrame = animation_frame;
+                curFrame = animation_frame;
                 animation_frame = targetFrame;
                 targetFrame &= 0x02;
-                steamFrame &= 0x02;
-                if (targetFrame != steamFrame)
+                curFrame &= 0x02;
+                if (targetFrame != curFrame)
                 {
                     auto curRide = GetRide();
                     if (curRide != nullptr)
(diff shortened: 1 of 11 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7328,88 +7327,88 @@ void Vehicle::UpdateAdditionalAnimation()
...
                 animation_frame += 1;
                 animation_frame &= 7;
                 Invalidate();
             }
             break;
         case VEHICLE_ENTRY_ANIMATION_HELICARS: // loc_6D63F5
-            *curVar_C8 += _vehicleVelocityF64E08;
-            al = (*curVar_C8 >> 18) & 3;
-            if (animation_frame != al)
+            *curAnimation_state += _vehicleVelocityF64E08;
```
Gymnasiast, line 7381: `curAnimationState`.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7275,11 +7275,11 @@ void Vehicle::UpdateSpinningCar()
  */
 void Vehicle::UpdateAdditionalAnimation()
 {
-    uint8_t al{};
-    uint8_t ah{};
+    uint8_t target_frame{};
+    uint8_t steam_frame{};
```
Gymnasiast, line 7279:
```suggestion
    uint8_t targetFrame{};
    uint8_t steamFrame{};
```

src/openrct2/ride/Vehicle.h
```diff
@@ -181,7 +181,7 @@ struct Vehicle : SpriteBase
     };
     uint8_t animation_frame;
     uint8_t pad_C6[0x2];
-    uint16_t var_C8;
+    uint16_t animation_state;
```
Gymnasiast, line 184: We are in the process of renaming everything to `camelCase`. As such, the appropriate name would be `animationState`.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7288,16 +7288,15 @@ void Vehicle::UpdateAdditionalAnimation()
...
-                animation_frame = ah;
-                al &= 0x02;
-                ah &= 0x02;
-                if (al != ah)
+            *curAnimation_state += _vehicleVelocityF64E08;
+            target_frame = (*curAnimation_state >> 20) & 3;
+            if (animation_frame != target_frame)
+            {
+                steam_frame = animation_frame;
+                animation_frame = target_frame;
```
Gymnasiast, line 7296: Are these two lines the refactor you’re talking about?

spacek531: yes

duncanspumpkin:
This is based off of work from here? https://github.com/OpenRCT2/OpenRCT2/pull/11269

spacek531:
No, it's branched from develop. I want to rethink the animation speed modifier.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -7275,11 +7275,11 @@ void Vehicle::UpdateSpinningCar()
...
 void Vehicle::UpdateAdditionalAnimation()
 {
-    uint8_t al{};
-    uint8_t ah{};
+    uint8_t targetFrame{};
+    uint8_t curFrame{};
     uint32_t eax{};
 
-    uint32_t* curVar_C8 = reinterpret_cast<uint32_t*>(&var_C8);
+    uint32_t* curAnimationState = reinterpret_cast<uint32_t*>(&animationState);
```
ZehMatt, line 7282: This is horrible, its a cast from 16 bit to 32 bit. I know its been like this from before but this is surely wrong.

duncanspumpkin: Would need to introduce a union ideally

ZehMatt: I think this needs some further looking into, its probably one field and not two.

duncanspumpkin: I'm pretty sure it's just one more union situation. There were lots of them originally and it seems this one was missed.

ZehMatt: Some animation types have different shifts, I think making this 32 bit and then have the right masks is better than the union here.

Gymnasiast: Let’s do this in another PR.

PR #15256, opened 2021-08-21
Fix #15255: Wall banner index defaults to 0 instead of null type

### LINKED ISSUE #15255: New wall objects placed in 8de2965 contains "banner text" info
### Operating System

Windows 10

### OpenRCT2 build

v0.3.4.1 (8de2965 on develop)

### Describe the issue

Placing any size of wall objects in 8de2965 shows banner text info when the object placed is not a banner.
banner text wall object

The info for another wall object on the same tile placed before 8de2965 properly shows "not a banner"
banner text wall object2

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Place wall objects in 8de2965
2. Tile Inspector info shows "banner text" where there should not be any

### Attachments

Affected save can be provided via private message because the park is being built for an ongoing competition

ZehMatt:
No idea for how long this has been in, the side effect is only visible from the tile inspector as far as I can tell.

Closes #15255

Edit: Can reproduce this on 0.3.4.1 but not 0.3.3, so worth a changelog entry I guess.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -14,7 +14,7 @@
 - Fix: [#15170] Plugin: incorrect label text alignment.
 - Fix: [#15184] Crash when hovering over water types in Object Selection.
 - Fix: [#15193] Crash when rides/stalls are demolished.
-- Fix: [#15255] Normal wall scenery would have the banner index 0 at all times.
+- Fix: [#15255] Tile Inspector shows banner information on walls that do not contain one.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
 - Change: [#8601] Revert ToonTower base block fix to re-enable support blocking.
 - Change: [#15174] [Plugin] Deprecate the type "peep" and add support to target a specific scripting api version.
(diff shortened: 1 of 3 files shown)


distribution/changelog.txt
```diff
@@ -14,6 +14,7 @@
 - Fix: [#15170] Plugin: incorrect label text alignment.
 - Fix: [#15184] Crash when hovering over water types in Object Selection.
 - Fix: [#15193] Crash when rides/stalls are demolished.
+- Fix: [#15255] Normal wall scenery would have the banner index 0 at all times.
```
ZehMatt, line 17: I'm not sure about the wording here, this is not some bug you get to see without looking hard enough with the tile inspector.

Gymnasiast: I think it’s better to list the symptoms, which we almost always do anyway:

```suggestion
- Fix: [#15255] Tile Inspector shows banner information on walls that do not contain one.
```

Gymnasiast:
One thing we might need to think about is: what do we do with already affected saves? Is it something we’d want to fix on import (by looking the object and seeing if it has a banner)?

ZehMatt:
I think it should be fine without import fixes, if I'm not mistaken normal walls that are actual Banners will have a correct Banner Id and normal Walls would just point to the first Banner which would be only visible with the Tile Inspector.

duncanspumpkin:
Yes its fine for existing its only TileInspector that shows the potential issue.

Issue #13874, opened 2021-01-19
Refactoring track data


spacek531:
Currently, track data is hardcoded in a number of tables, which makes it a challenge to make new pieces. Something akin to the ride type descriptor may help congregate track data more succinctly, which can then be spread to the numerous tables OpenRCT2 uses currently to store track data.

frutiemax:
Is this being worked on? I am working on it.

PR #15261, opened 2021-08-22
Track Element Descriptor refactor (issue #13874)

frutiemax:
Instead of creating multiple files as the RideTypeDescriptor refactor, a loader class loads the legacy tables, and TrackElementDescriptor (TED) are created and stored in this class. To access the TED, you must use GetTrackElementDescriptor. Some points to adjust/to be done:

- When to load the tables, right now it is done at the start of the game
- Provide an interface to add new TEDs
- There is still the gTrackDescriptors table that is accessed in _legacy.cpp, I have no idea what this is for.

### DIFF
--- a/src/openrct2-ui/windows/NewRide.cpp
+++ b/src/openrct2-ui/windows/NewRide.cpp
@@ -946,7 +946,7 @@ static void window_new_ride_paint_ride_information(
         int32_t startPieceId = GetRideTypeDescriptor(item.Type).StartTrackPiece;
         money64 price = GetRideTypeDescriptor(item.Type).BuildCosts.TrackPrice;
         const auto& ted = GetTrackElementDescriptor(startPieceId);
-        price *= ted.Pricing;
+        price *= ted.Price;
         price = (price >> 17) * 10 * GetRideTypeDescriptor(item.Type).BuildCosts.PriceEstimateMultiplier;
 
         //
--- a/src/openrct2-ui/windows/NewRide.cpp
+++ b/src/openrct2-ui/windows/NewRide.cpp
@@ -33,6 +33,8 @@
 #include <openrct2/windows/Intent.h>
 #include <openrct2/world/Park.h>
 
+using namespace OpenRCT2::TrackMetaData;
+
 static constexpr const rct_string_id WINDOW_TITLE = STR_NONE;
 constexpr size_t AVAILABILITY_STRING_SIZE = 256;
 static constexpr const int32_t WH = 382;
@@ -943,7 +945,8 @@ static void window_new_ride_paint_ride_information(
         // Get price of ride
         int32_t unk2 = GetRideTypeDescriptor(item.Type).StartTrackPiece;
         money64 price = GetRideTypeDescriptor(item.Type).BuildCosts.TrackPrice;
-        price *= TrackPricing[unk2];
+        const auto& teDescriptor = GetTrackElementDescriptor(unk2);
+        price *= teDescriptor.Pricing;
         price = (price >> 17) * 10 * GetRideTypeDescriptor(item.Type).BuildCosts.PriceEstimateMultiplier;
 
         //
--- a/src/openrct2-ui/windows/NewRide.cpp
+++ b/src/openrct2-ui/windows/NewRide.cpp
@@ -943,10 +943,10 @@ static void window_new_ride_paint_ride_information(
     if (!(gParkFlags & PARK_FLAGS_NO_MONEY))
     {
         // Get price of ride
-        int32_t unk2 = GetRideTypeDescriptor(item.Type).StartTrackPiece;
+        int32_t startPieceId = GetRideTypeDescriptor(item.Type).StartTrackPiece;
         money64 price = GetRideTypeDescriptor(item.Type).BuildCosts.TrackPrice;
-        const auto& teDescriptor = GetTrackElementDescriptor(unk2);
-        price *= teDescriptor.Pricing;
+        const auto& ted = GetTrackElementDescriptor(startPieceId);
+        price *= ted.Pricing;
         price = (price >> 17) * 10 * GetRideTypeDescriptor(item.Type).BuildCosts.PriceEstimateMultiplier;
 
         //
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -2349,7 +2349,7 @@ static void window_ride_construction_draw_track_piece(
         return;
 
     const auto& teDescriptor = GetTrackElementDescriptor(trackType);
-    auto trackBlock = teDescriptor.Block;
+    const auto* trackBlock = teDescriptor.Block;
     while ((trackBlock + 1)->index != 0xFF)
         trackBlock++;
 
@@ -2404,7 +2404,7 @@ static void sub_6CBCE2(
     gMapSize = MAXIMUM_MAP_SIZE_TECHNICAL;
 
     const auto& teDescriptor = GetTrackElementDescriptor(trackType);
-    auto trackBlock = teDescriptor.Block;
+    const auto* trackBlock = teDescriptor.Block;
     while (trackBlock->index != 255)
     {
         auto quarterTile = trackBlock->var_08.Rotate(trackDirection);
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -42,6 +42,8 @@ static constexpr const int32_t WH = 394;
 static constexpr const int32_t WW = 166;
 static constexpr const uint16_t ARROW_PULSE_DURATION = 200;
 
+using namespace OpenRCT2::TrackMetaData;
+
 #pragma region Widgets
 
 // clang-format off
@@ -1924,7 +1926,8 @@ static void window_ride_construction_mouseup_demolish(rct_window* w)
             return;
         }
 
-        const rct_preview_track* trackBlock = TrackBlocks[tileElement->AsTrack()->GetTrackType()];
+        const auto& teDescriptor = GetTrackElementDescriptor(tileElement->AsTrack()->GetTrackType());
+        const rct_preview_track* trackBlock = teDescriptor.Block;
         newCoords->z = (tileElement->GetBaseZ()) - trackBlock->z;
         gGotoStartPlacementMode = true;
     }
@@ -2345,7 +2348,8 @@ static void window_ride_construction_draw_track_piece(
     if (ride == nullptr)
         return;
 
-    auto trackBlock = TrackBlocks[trackType];
+    const auto& teDescriptor = GetTrackElementDescriptor(trackType);
+    auto trackBlock = teDescriptor.Block;
     while ((trackBlock + 1)->index != 0xFF)
         trackBlock++;
 
@@ -2399,7 +2403,8 @@ static void sub_6CBCE2(
 
     gMapSize = MAXIMUM_MAP_SIZE_TECHNICAL;
 
-    auto trackBlock = TrackBlocks[trackType];
+    const auto& teDescriptor = GetTrackElementDescriptor(trackType);
+    auto trackBlock = teDescriptor.Block;
     while (trackBlock->index != 255)
     {
         auto quarterTile = trackBlock->var_08.Rotate(trackDirection);
@@ -3292,7 +3297,8 @@ static void window_ride_construction_select_map_tiles(
 
     const rct_preview_track* trackBlock;
 
-    trackBlock = TrackBlocks[trackType];
+    const auto& teDescriptor = GetTrackElementDescriptor(trackType);
+    trackBlock = teDescriptor.Block;
     trackDirection &= 3;
     gMapSelectionTiles.clear();
     while (trackBlock->index != 255)
@@ -3469,7 +3475,8 @@ void ride_construction_toolupdate_construct(const ScreenCoordsXY& screenCoords)
         // z = map_get_highest_z(x >> 5, y >> 5);
     }
     // loc_6CC91B:
-    trackBlock = TrackBlocks[trackType];
+    const auto& teDescriptor = GetTrackElementDescriptor(trackType);
+    trackBlock = teDescriptor.Block;
     int32_t bx = 0;
     do
     {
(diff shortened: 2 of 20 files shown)


ZehMatt (changes requested):
Its overall not bad but I think the interface design can be probably improved.

The first thing that sticks out is that you have statics but also an instance of it just for the member, you can move the member into the cpp file as a static and then you don't require an instance, it might better to move it entirely into a namespace instead of a class. Also the alternative to the current code would be to use singleton pattern instead of static functions.

Also is the data supposed to be loaded only once at startup or should the we allow modifications by loading/unloading the data at any point? This drives the decision if this needs to be in the context or can be a singleton over the entire process.

Also TEDLoader is not a very good name, something like TrackMetaData would be more sensible in my opinion.

frutiemax:
Something like this?

`namespace TrackMetaData
{
    static void Init();
    static const TrackElementDescriptor& GetTrackElementDescriptor(const uint32_t& type);
}`

For now, I assume this is modified only at startup.

ZehMatt:
> Something like this?
> 
> ```
> namespace TrackMetaData
> {
>     static void Init();
>     static const TrackElementDescriptor& GetTrackElementDescriptor(const uint32_t& type);
> }
> ```

Yeah this would work, dont forget the OpenRCT2 namespace. Also use only statics for functions you don't want to be global, the getter is not accessible if its static.

duncanspumpkin:
Will the idea to be to bring in the other track data as well like https://github.com/OpenRCT2/OpenRCT2/pull/13883?

src/openrct2/ride/RideConstruction.cpp
```diff
@@ -1467,7 +1474,9 @@ void sub_6CB945(Ride* ride)
                         continue;
                     if (tileElement->AsTrack()->GetSequenceIndex() != 0)
                         continue;
-                    if (!(TrackSequenceProperties[tileElement->AsTrack()->GetTrackType()][0] & TRACK_SEQUENCE_FLAG_ORIGIN))
+
+                    teDescriptor = GetTrackElementDescriptor(tileElement->AsTrack()->GetTrackType());
```
ZehMatt, line 1478:
```suggestion
                    const auto& teDescriptor = GetTrackElementDescriptor(tileElement->AsTrack()->GetTrackType());
```

src/openrct2/ride/RideConstruction.cpp
```diff
@@ -1494,7 +1503,8 @@ void sub_6CB945(Ride* ride)
                 continue;
             }
 
-            const rct_preview_track* trackBlock = TrackBlocks[tileElement->AsTrack()->GetTrackType()];
+            teDescriptor = GetTrackElementDescriptor(tileElement->AsTrack()->GetTrackType());
```
ZehMatt, line 1506:
```suggestion
            const auto& teDescriptor = GetTrackElementDescriptor(tileElement->AsTrack()->GetTrackType());
```

src/openrct2/ride/RideConstruction.cpp
```diff
@@ -1509,7 +1519,9 @@ void sub_6CB945(Ride* ride)
                         continue;
                     if (tileElement->GetType() != TILE_ELEMENT_TYPE_TRACK)
                         continue;
-                    if (!(TrackSequenceProperties[tileElement->AsTrack()->GetTrackType()][0] & TRACK_SEQUENCE_FLAG_ORIGIN))
+
+                    teDescriptor = GetTrackElementDescriptor(tileElement->AsTrack()->GetTrackType());
```
ZehMatt, line 1523:
```suggestion
                    const auto& teDescriptor = GetTrackElementDescriptor(tileElement->AsTrack()->GetTrackType());
```

src/openrct2/ride/RideConstruction.cpp
```diff
@@ -1604,7 +1616,8 @@ void sub_6CB945(Ride* ride)
 
                 Direction direction = (tileElement->GetDirection() - direction_reverse(trackElement->GetDirection())) & 3;
 
-                if (!(TrackSequenceProperties[trackType][trackSequence] & (1 << direction)))
+                teDescriptor = GetTrackElementDescriptor(trackType);
```
ZehMatt, line 1619:
```suggestion
                const auto& teDescriptor = GetTrackElementDescriptor(trackType);
```

src/openrct2/ride/Ride.cpp
```diff
@@ -2812,7 +2822,8 @@ static bool ride_check_start_and_end_is_station(CoordsXYE* input)
     // Check back of the track
     track_get_back(input, &trackBack);
     auto trackType = trackBack.element->AsTrack()->GetTrackType();
-    if (!(TrackSequenceProperties[trackType][0] & TRACK_SEQUENCE_FLAG_ORIGIN))
+    auto teDescriptor = GetTrackElementDescriptor(trackType);
```
ZehMatt, line 2825:
```suggestion
    const auto& teDescriptor = GetTrackElementDescriptor(trackType);
```

src/openrct2/ride/TrackData.h
```diff
@@ -67,10 +54,29 @@ constexpr const dodgems_track_size DodgemsTrackSize(track_type_t type)
...
+    std::array<uint8_t, MaxSequencesPerPiece> TrackSequenceProperties;
+};
 
-extern const uint16_t TrackFlags[TrackElemType::Count];
+namespace OpenRCT2
+{
+    namespace TrackMetaData
+    {
+        void Init();
+        const TrackElementDescriptor& GetTrackElementDescriptor(const uint32_t& type);
```
ZehMatt, line 80:
```suggestion
        const TrackElementDescriptor& GetTrackElementDescriptor(const uint32_t type);
```
Trivial types should be passed as value.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1646,7 +1650,7 @@ static std::optional<money32> track_design_place_ride(TrackDesign* td6, const Co
             }
         }
 
-        const rct_track_coordinates* track_coordinates = &TrackCoordinates[trackType];
+        const rct_track_coordinates* track_coordinates = &teDescriptor.Coordinates;
```
ZehMatt, line 1653:
```suggestion
        const rct_track_coordinates& track_coordinates = teDescriptor.Coordinates;
```

src/openrct2/world/TileInspector.cpp
```diff
@@ -854,7 +855,8 @@ namespace OpenRCT2::TileInspector
             if (ride == nullptr)
                 return std::make_unique<GameActions::Result>(GameActions::Status::Unknown, STR_NONE);
 
-            auto trackBlock = TrackBlocks[type];
+            const auto& teDescriptor = GetTrackElementDescriptor(type);
+            auto trackBlock = teDescriptor.Block;
```
ZehMatt, line 859:
```suggestion
            const auto* trackBlock = teDescriptor.Block;
```
Always specify pointer/reference type on auto.

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2345,7 +2348,8 @@ static void window_ride_construction_draw_track_piece(
     if (ride == nullptr)
         return;
 
-    auto trackBlock = TrackBlocks[trackType];
+    const auto& teDescriptor = GetTrackElementDescriptor(trackType);
+    auto trackBlock = teDescriptor.Block;
```
ZehMatt, line 2352:
```suggestion
    const auto* trackBlock = teDescriptor.Block;
```

src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -2399,7 +2403,8 @@ static void sub_6CBCE2(
 
     gMapSize = MAXIMUM_MAP_SIZE_TECHNICAL;
 
-    auto trackBlock = TrackBlocks[trackType];
+    const auto& teDescriptor = GetTrackElementDescriptor(trackType);
+    auto trackBlock = teDescriptor.Block;
```
ZehMatt, line 2407:
```suggestion
    const auto* trackBlock = teDescriptor.Block;
```

src/openrct2/ride/Ride.cpp
```diff
@@ -527,13 +531,13 @@ bool track_block_get_next(CoordsXYE* input, CoordsXYE* output, int32_t* z, int32
     if (ride == nullptr)
         return false;
 
-    auto trackBlock = TrackBlocks[inputElement->GetTrackType()];
+    auto trackBlock = teDescriptor.Block;
```
ZehMatt, line 534:
```suggestion
    const auto* trackBlock = teDescriptor.Block;
```

src/openrct2/ride/Ride.cpp
```diff
@@ -595,11 +599,11 @@ bool track_block_get_previous_from_zero(
         if (trackElement->GetRideIndex() != ride->id)
             continue;
 
-        auto nextTrackBlock = TrackBlocks[trackElement->GetTrackType()];
+        const auto& teDesc = GetTrackElementDescriptor(trackElement->GetTrackType());
+        auto nextTrackBlock = teDesc.Block;
```
ZehMatt, line 603:
```suggestion
        const auto* nextTrackBlock = teDesc.Block;
```

src/openrct2/ride/Ride.cpp
```diff
@@ -631,7 +635,8 @@ bool track_block_get_previous_from_zero(
 
         outTrackBeginEnd->begin_z = tileElement->GetBaseZ();
 
-        auto nextTrackBlock2 = TrackBlocks[trackElement->GetTrackType()];
+        const auto& teDescriptor = GetTrackElementDescriptor(trackElement->GetTrackType());
+        auto nextTrackBlock2 = teDescriptor.Block;
```
ZehMatt, line 639:
```suggestion
        const auto* nextTrackBlock2 = teDescriptor.Block;
```

src/openrct2/ride/Ride.cpp
```diff
@@ -671,13 +677,13 @@ bool track_block_get_previous(const CoordsXYE& trackPos, track_begin_end* outTra
     if (ride == nullptr)
         return false;
 
-    auto trackBlock = TrackBlocks[trackElement->GetTrackType()];
+    auto trackBlock = teDescriptor.Block;
```
ZehMatt, line 680:
```suggestion
    const auto* trackBlock = teDescriptor.Block;
```

src/openrct2/ride/RideConstruction.cpp
```diff
@@ -392,7 +394,8 @@ std::optional<CoordsXYZ> sub_6C683D(
     }
 
     // Possibly z should be & 0xF8
-    auto trackBlock = TrackBlocks[type];
+    const auto& teDescriptor = GetTrackElementDescriptor(type);
+    auto trackBlock = teDescriptor.Block;
```
ZehMatt, line 398:
```suggestion
    const auto* trackBlock = teDescriptor.Block;
```

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -182,8 +183,9 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
     trackElement.y = newCoords->y;
     z = newCoords->z;
 
-    const rct_track_coordinates* trackCoordinates = &TrackCoordinates[trackElement.element->AsTrack()->GetTrackType()];
-    auto trackBlock = TrackBlocks[trackType];
+    const auto& teDescriptor = GetTrackElementDescriptor(trackElement.element->AsTrack()->GetTrackType());
+    const rct_track_coordinates* trackCoordinates = &teDescriptor.Coordinates;
+    auto trackBlock = teDescriptor.Block;
```
ZehMatt, line 188:
```suggestion
    const auto* trackBlock = teDescriptor.Block;
```

ZehMatt (changes requested):
Looking better, few things left however.

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -943,7 +945,8 @@ static void window_new_ride_paint_ride_information(
         // Get price of ride
         int32_t unk2 = GetRideTypeDescriptor(item.Type).StartTrackPiece;
```
Gymnasiast, line 946: Rename `unk2` to `startPieceId`

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -943,7 +945,8 @@ static void window_new_ride_paint_ride_information(
...
         money64 price = GetRideTypeDescriptor(item.Type).BuildCosts.TrackPrice;
-        price *= TrackPricing[unk2];
+        const auto& teDescriptor = GetTrackElementDescriptor(unk2);
```
Gymnasiast, line 948: I would call the variable `ted`, shadowing the usage of `rtd`.

Gymnasiast:
I will do a proper review once `teDescriptor` is renamed, as that will cause all files to change again. Please also fix Matt’s line notes - I will leave memory/performance-related stuff to him, as that is not something that I’m particularly knowledgeable about.

In general, this looks like a good approach in making sure the code is converted to use TEDs sooner rather than later.

Gymnasiast:
You dropped the `Track` prefix in most cases, but not with `TrackSequenceProperties`. Probably best to also apply it to that.

ZehMatt:
Also please mark conversations as resolved if you changed the code.

src/openrct2/ride/RideConstruction.cpp
```diff
@@ -654,7 +659,8 @@ void ride_construction_set_default_next_piece()
                 }
             }
 
-            curve = gTrackCurveChain[trackType].next;
+            ted = &GetTrackElementDescriptor(trackType);
```
Gymnasiast, line 662: Any reason you opted for a pointer instead of a reference?

frutiemax: The variable gets reassigned.

ZehMatt (approved):
Code looks alright, did not test it.

Gymnasiast (approved):

Gymnasiast:
Would like to take it for a brief test.

src/openrct2/ride/TrackData.h
```diff
@@ -67,10 +54,29 @@ constexpr const dodgems_track_size DodgemsTrackSize(track_type_t type)
...
-extern const track_type_t TrackElementMirrorMap[TrackElemType::Count];
-
-extern const uint32_t TrackHeightMarkerPositions[TrackElemType::Count];
 
-extern const uint8_t TrackSequenceElementAllowedWallEdges[TrackElemType::Count][16];
+struct TrackElementDescriptor
+{
+    rct_track_coordinates Coordinates;
+
+    rct_preview_track* Block;
```
duncanspumpkin, line 62: For OpenLoco i modified this structure to return a `std::vector<rct_preview_track>` it makes things a lot simpler to handle but will require a lot of careful adjustment to the code to implement this.

PR #15267, opened 2021-08-23
Remove var ca

spacek531:
increment network version because animationState will desync

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -2722,7 +2722,6 @@ namespace RCT1
         dst->var_C4 = src->var_C4;
         dst->animation_frame = src->animation_frame;
         dst->animationState = src->animationState;
-        dst->var_CA = src->var_CA;
         dst->var_CE = src->var_CE;
         dst->var_D3 = src->var_D3;
         dst->scream_sound_id = OpenRCT2::Audio::SoundId::Null;
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -1180,7 +1180,6 @@ template<> void S6Exporter::ExportEntity(RCT2SpriteVehicle* dst, const Vehicle* 
     dst->dodgems_collision_direction = src->dodgems_collision_direction;
     dst->animation_frame = src->animation_frame;
     dst->animationState = src->animationState;
-    dst->var_CA = src->var_CA;
     dst->scream_sound_id = static_cast<uint8_t>(src->scream_sound_id);
     dst->TrackSubposition = static_cast<uint8_t>(src->TrackSubposition);
     dst->var_CE = src->var_CE;
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -1664,7 +1664,6 @@ template<> void S6Importer::ImportEntity<Vehicle>(const RCT12SpriteBase& baseSrc
     dst->dodgems_collision_direction = src->dodgems_collision_direction;
     dst->animation_frame = src->animation_frame;
     dst->animationState = src->animationState;
-    dst->var_CA = src->var_CA;
     dst->scream_sound_id = static_cast<OpenRCT2::Audio::SoundId>(src->scream_sound_id);
     dst->TrackSubposition = VehicleTrackSubposition{ src->TrackSubposition };
     dst->var_CE = src->var_CE;
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -3647,9 +3647,8 @@ void Vehicle::UpdateCollisionSetup()
 
         train->IsCrashedVehicle = true;
         train->animationState = scenario_rand();
-        train->var_CA = scenario_rand();
 
-        train->animation_frame = train->var_CA & 0x7;
+        train->animation_frame = (train->animationState >> 16) & 0x7;
         train->sprite_width = 13;
         train->sprite_height_negative = 45;
         train->sprite_height_positive = 5;
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -3646,9 +3646,9 @@ void Vehicle::UpdateCollisionSetup()
         }
 
         train->IsCrashedVehicle = true;
-        train->animationState = scenario_rand();
+        train->animationState = scenario_rand() & 0xFFFF;
 
-        train->animation_frame = (train->animationState >> 16) & 0x7;
+        train->animation_frame = scenario_rand() & 0x7;
         train->sprite_width = 13;
         train->sprite_height_negative = 45;
         train->sprite_height_positive = 5;
@@ -5512,9 +5512,9 @@ void Vehicle::UpdateCrash()
                     ExplosionCloud::Create({ curVehicle->x + xOffset, curVehicle->y + yOffset, curVehicle->z });
                 }
             }
-            if (curVehicle->animationState <= 58248)
+            if (curVehicle->animationState <= 0xe388)
             {
-                curVehicle->animationState += 7281;
+                curVehicle->animationState += 0x1c71;
             }
             else
             {
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -7467,7 +7467,7 @@ void Vehicle::UpdateAdditionalAnimation()
             UpdateAnimationAnimalFlying();
             // makes animation play faster with vehicle speed
             targetFrame = abs(_vehicleVelocityF64E08) >> 24;
-            animationState = std::max(animationState - targetFrame, 0);
+            animationState = std::max(animationState - targetFrame, 0u);
             break;
     }
 }
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -1522,16 +1522,16 @@ bool Vehicle::OpenRestraints()
         }
         if (vehicleEntry->animation == VEHICLE_ENTRY_ANIMATION_OBSERVATION_TOWER && vehicle->animation_frame != 0)
         {
-            if (vehicle->animationState + 0x3333 < 0xFFFF)
+            if (vehicle->animationState <= 0xCCCC)
             {
-                vehicle->animationState = vehicle->animationState + 0x3333 - 0xFFFF;
-                vehicle->animation_frame++;
-                vehicle->animation_frame &= 7;
-                vehicle->Invalidate();
+                vehicle->animationState += 0x3333;
             }
             else
             {
-                vehicle->animationState += 0x3333;
+                vehicle->animationState = 0;
+                vehicle->animation_frame++;
+                vehicle->animation_frame &= 7;
+                vehicle->Invalidate();
             }
             restraintsOpen = false;
             continue;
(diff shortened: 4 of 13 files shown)


src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1664,7 +1664,6 @@ template<> void S6Importer::ImportEntity<Vehicle>(const RCT12SpriteBase& baseSrc
     dst->dodgems_collision_direction = src->dodgems_collision_direction;
     dst->animation_frame = src->animation_frame;
     dst->animationState = src->animationState;
```
ZehMatt, line 1666: I think you forgot to combine var_CA and animationState here.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1180,7 +1180,6 @@ template<> void S6Exporter::ExportEntity(RCT2SpriteVehicle* dst, const Vehicle*
     dst->dodgems_collision_direction = src->dodgems_collision_direction;
     dst->animation_frame = src->animation_frame;
     dst->animationState = src->animationState;
-    dst->var_CA = src->var_CA;
```
ZehMatt, line 1183: This needs to be done correctly, you have to split them from animationState back into the two.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1664,7 +1664,6 @@ template<> void S6Importer::ImportEntity<Vehicle>(const RCT12SpriteBase& baseSrc
...
-    dst->var_CA = src->var_CA;
```
ZehMatt, line 1667: The import is not correct, it needs to combine var_CA and animationState into the single field now.

ZehMatt (changes requested):

src/openrct2/rct2/S6Exporter.cpp
duncanspumpkin: it can be done by just changing the definition of RCT2SpriteVehicle to make animationState uint32_t

src/openrct2/ride/Vehicle.cpp
```diff
@@ -3640,9 +3640,8 @@ void Vehicle::UpdateCollisionSetup()
 
         train->IsCrashedVehicle = true;
         train->animationState = scenario_rand();
-        train->var_CA = scenario_rand();
 
-        train->animation_frame = train->var_CA & 0x7;
+        train->animation_frame = (train->animationState >> 16) & 0x7;
```
duncanspumpkin, line 3644: This is a bit odd. It should be refactored. Will need to understand the rest of the users of this random information but id prefer if this was just `scenario_rand() & 0x7`

src/openrct2/ride/Vehicle.cpp
```diff
@@ -5506,14 +5505,18 @@ void Vehicle::UpdateCrash()
                     ExplosionCloud::Create({ curVehicle->x + xOffset, curVehicle->y + yOffset, curVehicle->z });
                 }
             }
-            if (curVehicle->animationState + 7281 > 0xFFFF)
+            if (curVehicle->animationState <= 58248)
```
duncanspumpkin, line 5508: Can you switch to using the hex values for this as it will be easier to understand. I'm not sure this logic is still the same due to the size of scenario_rand

src/openrct2/rct2/S6Exporter.cpp
ZehMatt: I didn't realize, yeah that should be fine.

spacek531 (commented):
I believe the only user of that information is the other function you've highlighted. It chooses a random frame of 8 frames of coaster debris while spawning the coaster debris, then in the other function it animates them.

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: > I believe the only user of that information is the other function you've highlighted. It chooses a random frame of 8 frames of coaster debris while spawning the coaster debris, then in the other function it animates them.

Yes my conclusion is the same. I think the best solution to this then is the following for this bit of code:
```
train->animationState = sceanrio_rand() & 0xFFFF;
train->animation_frame = sceanrio_rand() & 0x7;
```
That means that the animation frame will not immediately change on the next tick due to being >58248 in the vast majority of cases.

duncanspumpkin (approved):
This all looks correct. Have you tested it much?
I'll record the replays when i get a spare moment to get the ci's happy.

spacek531:
I tested it before the crash debris fix, and it seemed to work fine.

duncanspumpkin:
Close #10161 on merge.

duncanspumpkin:
Ive verified that the replays are only failing due to the layout of the memory:
The following code added to Entity.cpp will make them pass. Note I do not want this implemented and will rerecord the replay.
It did though bring up the issue of import and export. We are not the same as vanilla due to the byte order so we should swap this about. See comments on lines that require changed.
```cpp
    if (stream.IsSaving())
    {
        uint16_t value = (animationState >> 16);
        stream << value;
        value = animationState & 0xFFFF;
        stream << value;
    }
    else
    {
        uint16_t value = 0;
        stream << value;
        uint16_t value2 = 0;
        stream << value2;
        animationState = (value << 16) | value2;
    }
```

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2716,7 +2716,6 @@ template<> void S4Importer::ImportEntity<Vehicle>(const RCT12SpriteBase& srcBase
     dst->var_C4 = src->var_C4;
     dst->animation_frame = src->animation_frame;
     dst->animationState = src->animationState;
```
duncanspumpkin, line 2718:
```
    // Due to memory layout changes this field needs to be swapped around
    dst->animationState = ((src->animationState & 0xFFFF) << 16) | (src->animationState >> 16);
```

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1180,7 +1180,6 @@ template<> void S6Exporter::ExportEntity(RCT2SpriteVehicle* dst, const Vehicle*
     dst->dodgems_collision_direction = src->dodgems_collision_direction;
     dst->animation_frame = src->animation_frame;
     dst->animationState = src->animationState;
```
duncanspumpkin, line 1182:
```
    // Due to memory layout changes this field needs to be swapped around
    dst->animationState = ((src->animationState & 0xFFFF) << 16) | (src->animationState >> 16);
```

src/openrct2/rct2/S6Exporter.cpp
duncanspumpkin: Turns out there was some reorg required.

src/openrct2/rct2/S6Importer.cpp
duncanspumpkin, line 1666:
```
    // Due to memory layout changes this field needs to be swapped around
    dst->animationState = ((src->animationState & 0xFFFF) << 16) | (src->animationState >> 16);
```

spacek531:
It's not the proper order for little-endian?

spacek531:
> It's not the proper order for little-endian? I thought that writing the low 16 into var_C8 and the high 16 into var_CA is equivalent to writing var_C8 | (var_CA << 16) to the var_C8 memory location.

I think I am right about this. In 3.4 Release I made an observation tower and opened it. If the byte order is flipped, the read value should be > 0xFFFF or 0 because it would read var_C8 as the high 16. The value was 0x997E, which means the value is read correctly.

duncanspumpkin:
I'm just going by what the replays are showing me. For the nsf we wouldn't need to do this we would just save in the correct order. Import of legacy would keep the switching code though.

spacek531:
Could the replay desyncs not be caused by removing a variable? If what you're saying is true, that this breaks the sv6 standard, then the implications of my results mean that OpenRCT2 has broken the sv6 standard for longer than 3.4 Release.

duncanspumpkin:
It's easy enough to test your hypothesis. Apply my change to the entity.cpp file and if the replay for bpb passes then the import/export code does not need changed. The entity.cpp file at this moment is only used for the replays to calculate the hash and to store the entity.

duncanspumpkin:
Hmm perhaps i was wrong. I've done a variety of tests see below using replay bpb1000:
Before i was using a different replay but i had a thought, that replay doesn't actually have any vehicles in it on load! So you cant draw any conclusions from it!

This is without the import/export code 106eeee
```
tick left = 000147C5, tick right = 000147C5
srand0 left = 55D16A6B, srand0 right = 55D16A6B
Sprite modifications (Vehicle), index: 240
  Vehicle::animationState, len = 4, offset = 208, left = 0x000000007E70C853, right = 0x00000000C8537E70
Sprite modifications (Vehicle), index: 301
  Vehicle::animationState, len = 4, offset = 208, left = 0x0000000003AB6587, right = 0x00000000658703AB
So on and so on for all vehicles with none zero animation states...
Replay stops after 1 tick due to checksum mismatch.
Different sprite checksum at tick 83910 (Replay Tick: 1) ; Saved: 7593b2e3ca9ffeec000000000000000000000000, Current: 3e40e8d1dcfef327000000000000000000000000
```
without the import/export code 106eeee (and with my Entity.cpp change)
```
Sprite modifications (Vehicle), index: 240
  Vehicle::animationState, len = 4, offset = 208, left = 0x000000007E70C853, right = 0x00000000C8537E70
Sprite modifications (Vehicle), index: 301
  Vehicle::animationState, len = 4, offset = 208, left = 0x0000000003AB6587, right = 0x00000000658703AB
So on and so on for all vehicles with none zero animation states...
Replay stops after 1 tick due to checksum mismatch.
Different sprite checksum at tick 83910 (Replay Tick: 1) ; Saved: 7593b2e3ca9ffeec000000000000000000000000, Current: 535b2f87f6f92ca7000000000000000000000000
```

With bba86da (and with my Entity.cpp change)
```
Different sprite checksum at tick 83910 (Replay Tick: 1) ; Saved: 7593b2e3ca9ffeec000000000000000000000000, Current: c8de289428948efa000000000000000000000000
```
With bba86da
```
Different sprite checksum at tick 83910 (Replay Tick: 1) ; Saved: 7593b2e3ca9ffeec000000000000000000000000, Current: 7d56a98830aff1e0000000000000000000000000
```

ZehMatt (changes requested):
This PR requires a rebase because of the network version, it should be raised to 7. Also its best to have the network increment in a separate commit so its easier to change it.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1522,16 +1522,16 @@ bool Vehicle::OpenRestraints()
         }
         if (vehicleEntry->animation == VEHICLE_ENTRY_ANIMATION_OBSERVATION_TOWER && vehicle->animation_frame != 0)
         {
-            if (vehicle->animationState + 0x3333 < 0xFFFF)
+            if (vehicle->animationState <= 0xCCCC)
```
Gymnasiast, line 1525: To me, it looks like the old code checked for `< 0xCCCC`, not `<=`.

duncanspumpkin: The previous code relied on overflow to reset the count. This now does not. During the conversion the two branches of the if were flipped. If you look at the other animation modes in this function you'll see how they now operate in the same way. Basically on the frame where it resets to zero we want the animation to do something (change frame) all other times just count up.

spacek531: The change from `< 0xCCCC` to `<= 0xCCCC` goes hand-in-hand with resetting the timer to 0. Since `0x1000 % 0x3333 != 0`, there state is constantly 'slipping' backwards. This normally isn't a problem with the highlighted code unless the number of frames the animation has cycled through is a multiple of 0x3333, in which case the above code will update the car after 5 frames of waiting instead of 4. Since the new behavior of resetting animationState to 0 will always trigger that behavior, changing it to include 0xCCCC is necessary.

Gymnasiast (approved):
Fuck if I understand what this black magic is doing and why the change is equivalent, but since both Duncan and Kelson are convinced it is, I’m approving it.

PR #15268, opened 2021-08-23
Make all location types constexpr

ZehMatt:


src/openrct2/world/Location.hpp
```diff
@@ -244,32 +244,32 @@ struct CoordsXYZ : public CoordsXY
...
     }
 
-    CoordsXYZ ToTileCentre() const
+    constexpr CoordsXYZ ToTileCentre() const
     {
         return ToTileStart() + CoordsXYZ{ COORDS_XY_HALF_TILE, COORDS_XY_HALF_TILE, 0 };
     }
 
-    void setNull()
+    constexpr void setNull()
```
Broxzier, line 272: Since this function doesn't modify anything other than itself and doesn't return anything, what is the use of `constexpr` here? Is it required so that other `constexpr` functions can call this?

duncanspumpkin (approved):

src/openrct2/world/Location.hpp
```diff
@@ -380,22 +380,22 @@ struct TileCoordsXY
...
     }
 
-    bool isNull() const
+    constexpr bool isNull() const
     {
         return x == COORDS_NULL;
     };
 
-    void setNull()
+    constexpr void setNull()
```
Broxzier, line 398: See above

src/openrct2/world/Location.hpp
```diff
@@ -466,7 +466,7 @@ struct TileCoordsXYZ : public TileCoordsXY
         return { x * COORDS_XY_STEP, y * COORDS_XY_STEP, z * COORDS_Z_STEP };
     }
 
-    void setNull()
+    constexpr void setNull()
```
Broxzier, line 469: See above

src/openrct2/world/Location.hpp
```diff
@@ -668,7 +668,7 @@ struct TileCoordsXYZD : public TileCoordsXYZ
         return { x * COORDS_XY_STEP, y * COORDS_XY_STEP, z * COORDS_Z_STEP, direction };
     }
 
-    void setNull()
+    constexpr void setNull()
```
Broxzier, line 671: See above

Broxzier:
Looks like this got accepted and merged while I was reviewing this.  I'm not opposing to this being merged, just curious what the use of the constexpr void functions are here.

src/openrct2/world/Location.hpp
ZehMatt: Yeah in a constexpr context you require them to be constexpr.

PR #15272, opened 2021-08-24
Fix #15271. Use formatter to pass description args to text input

### LINKED ISSUE #15271: "Enter selection size" prompt displays wrong values
### Operating System

Windows 10, 64-bit

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (e48dd2d on develop) provided by GitHub

### Describe the issue

The prompt that allows input between 1 to 64 displays incorrect values.

selection size

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Open any window that has size selection option (landscaping, water tool, clear tool, scatter tool, land purchase mode, and map size)
2. Click on the size selection to open prompt that allows inputting values between 1 to 64
3. Notice underflow/overflow values

### Attachments

_No response_

duncanspumpkin:
~~Originally passed the variables via global vars which were not updated to 32bit during recent refactors.~~ This removes the global and makes the interface cleaner and corrects the type

Edit. Looks like the actual issue that was causing the issue was passing a pointer to the arguments to a formatter instead of passing the arguments to the formatter. Either way this PR makes things much simpler and removes the global.

### DIFF
--- a/src/openrct2-ui/windows/Banner.cpp
+++ b/src/openrct2-ui/windows/Banner.cpp
@@ -201,7 +201,7 @@ public:
             }
             case WIDX_BANNER_TEXT:
                 window_text_input_raw_open(
-                    this, WIDX_BANNER_TEXT, STR_BANNER_TEXT, STR_ENTER_BANNER_TEXT, banner->GetText().c_str(), 32);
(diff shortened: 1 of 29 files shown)


ZehMatt (approved):
Code looks fine, did not test this.

duncanspumpkin:
I've tested this a bit and resolves the issues

Gymnasiast:
Nice fix! That’s one fewer global to worry about.

PR #15273, opened 2021-08-24
Implement DirectionFlipXAxis

ZehMatt:


ZehMatt:
This produces better assembly it seems: https://godbolt.org/z/aj8PPjcan

duncanspumpkin (approved):

PR #15275, opened 2021-08-24
Use inline for direction functions in Location.hpp

ZehMatt:
Functions that are inside headers are best declared inline. This eliminates the need of [[maybe_unused]] since you are no longer forcing copies of the function on each include, a different library such as libopenrct2ui might not call any of the functions but they are still compiled because of the static keyword and causes unreferenced function warnings by that. This is one of the files where I noticed this and there might be probably more.

### DIFF
--- a/src/openrct2/world/Location.hpp
+++ b/src/openrct2/world/Location.hpp
@@ -496,12 +496,12 @@ constexpr Direction ALL_DIRECTIONS[] = { 0, 1, 2, 3 };
  * Given a direction, return the direction that points the other way,
  * on the same axis.
  */
-[[maybe_unused]] static constexpr Direction direction_reverse(Direction dir)
+inline constexpr Direction direction_reverse(Direction dir)
 {
     return dir ^ 2;
 }
 
-[[maybe_unused]] static constexpr bool direction_valid(Direction dir)
+inline constexpr bool direction_valid(Direction dir)
 {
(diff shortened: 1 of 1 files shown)


tupaschoal (commented):
> Functions that are inside headers are best declared inline.

I mean, at least on clang, using the `inline` keyword used to just bump the threshold that the compiler decided whether to actually inline or not, it was never meant as a guarantee of inlining.

> This eliminates the need of [[maybe_unused]] since you are no longer forcing copies of the function on each include, a different library such as libopenrct2ui might not call any of the functions but they are still compiled because of the static keyword and causes unreferenced function warnings by that. This is one of the files where I noticed this and there might be probably more.

Would it be fine if we just left it as `constexpr`?

ZehMatt:
> > Functions that are inside headers are best declared inline.
> 
> I mean, at least on clang, using the `inline` keyword used to just bump the threshold that the compiler decided whether to actually inline or not, it was never meant as a guarantee of inlining.
I know that it might not inline it but it will also suppress the one definition rule which is the point, unlike static the compiler is not doing anything if its not used.

> Would it be fine if we just left it as `constexpr`?
Nope

tupaschoal (approved):
LGTM, not sure if you want to wait for a few to leave room for others to review

Gymnasiast (approved):
It does look a bit cleaner this way.

PR #15280, opened 2021-08-25
Cleanup DrawSpriteArgs

ZehMatt:
Most of the data is read only so having DPI in there didn't allow it to be const, this also makes more sense composition wise.


Gymnasiast (approved):
Code looks ok.

In practical terms, what is the benefit of doing this?

duncanspumpkin:
Think it's so that more things can be const are const and making it clear what is not const

PR #15284, opened 2021-08-26
Strip RCT1 out of RCT1 struct names in namespace

duncanspumpkin:
As the OpenRCT2 structs are not in a namespace yet had to add a good few `::` and fully qualified the RCT1's as well to make things less ambiguous.


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2025,7 +2024,7 @@ namespace RCT1
             }
             else
             {
-                ResearchItem researchItem = {};
+                ::ResearchItem researchItem = {};
```
Broxzier, line 2027: (More a note for a future PR)
To avoid global `::...`, we should move these in an OpenRCT2 namespace and change these types to `OpenRCT2::...`.

Gymnasiast (approved):

PR #15285, opened 2021-08-26
Move RCT2 structs into RCT2 header

duncanspumpkin:
This is some cleanup to help reduce the number of places where the legacy structures are.


src/openrct2/rct2/RCT2.h
```diff
@@ -786,6 +789,263 @@ struct RCT2RideRatingCalculationData
...
+    uint16_t map_size_minus_2;
+    uint16_t map_size;
+    uint16_t map_max_xy;
+    uint32_t same_price_throughout;
+    uint16_t suggested_max_guests;
+    uint16_t park_rating_warning_days;
+    uint8_t last_entrance_style;
+    uint8_t rct1_water_colour;
+    uint8_t pad_01358842[2];
+    RCT12ResearchItem research_items[RCT2_MAX_RESEARCH_ITEMS];
```
Broxzier, line 981: This variable name suggests it's the max for both RCT1 and RCT2. Should the constant not reflect that?

duncanspumpkin: The structure is the same for both rct1 and 2 but the limits are different.

duncanspumpkin:
ScenarioSources missing an include.

Gymnasiast (approved):

PR #15294, opened 2021-08-27
Bird animation

spacek531:
For the NEDesigns semi-finals going live on 2021-8-29. Should be agnostic with the var_CA PR.

### DIFF
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -7285,7 +7285,6 @@ void Vehicle::UpdateAnimationAnimalFlying()
     }
     else
     {
-        constexpr std::array<uint8_t, 4> frameWaitTimes = { 5, 3, 5, 3 };
 
         if (animation_frame == 0)
         {
@@ -7306,6 +7305,7 @@ void Vehicle::UpdateAnimationAnimalFlying()
             Invalidate();
         }
         // number of frames to skip before updating again
+        constexpr std::array<uint8_t, 4> frameWaitTimes = { 5, 3, 5, 3 };
         animationState = frameWaitTimes[animation_frame];
     }
 }
--- a/src/openrct2/ride/Vehicle.cpp
+++ b/src/openrct2/ride/Vehicle.cpp
@@ -1536,6 +1536,13 @@ bool Vehicle::OpenRestraints()
             restraintsOpen = false;
             continue;
         }
+        if (vehicleEntry->animation == VEHICLE_ENTRY_ANIMATION_ANIMAL_FLYING
+            && (vehicle->animation_frame != 0 || vehicle->animationState > 0))
+        {
+            vehicle->UpdateAnimationAnimalFlying();
+            restraintsOpen = false;
+            continue;
+        }
 
         if (vehicle->HasUpdateFlag(VEHICLE_UPDATE_FLAG_BROKEN_CAR) && vehicle->restraints_position != 0xFF
             && (curRide->breakdown_reason_pending == BREAKDOWN_RESTRAINTS_STUCK_CLOSED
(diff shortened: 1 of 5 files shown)


src/openrct2/ride/Vehicle.cpp
```diff
@@ -7269,6 +7276,50 @@ void Vehicle::UpdateSpinningCar()
...
+        return;
+    }
+    else
+    {
+        auto trackType = GetTrackType();
+        TileElement* trackElement = map_get_track_element_at_of_type_seq(TrackLocation, trackType, 0);
+        // default waiting time is 3 ticks
+        // the timer determines the rest period for the next animation frame
+        // so case 3 setting animationState to 3 makes frame 0 play for 3 ticks
+        animationState = 3;
```
duncanspumpkin, line 7293: I was thinking about this last night and wouldn't a table be even simpler:

```
constexpr std::array<uint16_t, 4> flyingStateWaitTime = { 3, 5, 3, 5 };
if (animationFrame == 0){
 // Insert animation start check, return if false...
animationState = flyingStateWaitTime[animationFrame];
} else {
animationState = flyingStateWaitTime[animationFrame];
//Move frame to next state code...
}
Invalidate();

**Comment by duncanspumpkin** [comment]
Will need a network version increment as well.

**Comment by spacek531** [comment]
> Will need a network version increment as well.

Could this be pushed at the same time as https://github.com/OpenRCT2/OpenRCT2/pull/15267? It has network bump.

File: src/openrct2/ride/Vehicle.cpp (hunk 1 above)
Comment (spacek531) [reply src/openrct2/ride/Vehicle.cpp:7293]: It will have to change to uint32_t as soon as the var_ca PR is merged. I don't think it adds anything to use the table other than making it look like paint code.

Comment (spacek531) [reply src/openrct2/ride/Vehicle.cpp:7293]: One more reason, the current behavior catches out of bounds frames, something that the other animation types also do. A table doesn't do this so gracefully, and for only 4 valid states there's not much benefit.

**Comment by spacek531** [comment]
There's not much time before the match goes live, and it would be very disappointing if this isn't merged before then. What can I do to expedite this?

File: src/openrct2/ride/Vehicle.cpp [hunk 2]
```diff
@@ -7421,6 +7472,15 @@ void Vehicle::UpdateAdditionalAnimation()
                 }
             }
             break;
+        case VEHICLE_ENTRY_ANIMATION_ANIMAL_FLYING:
+            UpdateAnimationAnimalFlying();
+            // prevents underflow of animationState timer
+            targetFrame = abs(_vehicleVelocityF64E08) >> 24;
+            if (targetFrame > animationState)
```
duncanspumpkin, line 7479: Change this to `animationState = std::max(animationState - targetFrame, 0);` that is how we normally show bounded mathematics. The comment is useless everyone can see it's stopping underflows. It should state why we are doing something like `// make animation speed affected by vehicle speed`.

src/openrct2/ride/Vehicle.cpp
duncanspumpkin: I basically don't want to see that fallthrough. It's bad design in general. Your wasting time computing the track element despite only using it for one state (0). You can get around any out of bounds issue if you modulo the lookup but it's not an issue as it's already bounded due to the &3. The size doesn't make any difference it can just as easily be a uint8 the types expand just fine to 32bit. Tbh my example it's a bit back to front slightly better would be:
```
if (animationFrame == 0){
 // Insert animation start check, return if false...
auto trackType = GetTrackType();
        TileElement* trackElement = map_get_track_element_at_of_type_seq(TrackLocation, trackType, 0);
      if (trackElement != nullptr && trackElement->AsTrack()->HasChain())
                {
                    // start flapping, bird
                    animation_frame = 1;
}
else{ return; }
} else {
// Next frame
animation_frame = (animation_frame + 1) % 4;
}
constexpr std::array<uint8_t, 4> flyingStateWaitTime = { 3, 5, 3, 5 };
animationState = flyingStateWaitTime[animationFrame];
Invalidate();
```
Sorry if it looks horrible doing this on mobile.

spacek531:
This is the last commit I can make, I have to sleep now. I am desperately hoping it satisfies everything and gets merged soon.

Broxzier:
The title confused me, I thought this was about the flying ducks but then vehicle code was altered. 😄

Do you have a vehicle that we can test this with?

spacek531:
There will be a vehicle available in the Tile Inspectors Semifinals H2H9 park in the Tile Inspectors vs. Logan's Run match.

duncanspumpkin (approved):

PR #15302, opened 2021-08-27
Refactor peep thoughts for bigger limits

ZehMatt:
Cherry picked this from #15167 (bf31c01724412b6b06926b908d91a8b0881df1ca). It will probably break replays.


src/openrct2/peep/Peep.h
```diff
@@ -485,12 +484,19 @@ enum PeepRideDecision
...
-    uint8_t item;          // 1
-    uint8_t freshness;     // 2 larger is less fresh
-    uint8_t fresh_timeout; // 3 updates every tick
+    PeepThoughtType type;
+    union
+    {
+        ride_id_t rideId;
+        ShopItem shopItem;
+        uint16_t item;
+    };
```
Broxzier, line 497: Does this have to be a union? If size of the struct is not an issue, perhaps split them up or use `std::variant<ride_id_t, ShopItem, uint16_t>`. Import code likely needs to be changed then too.

duncanspumpkin (approved):

src/openrct2/peep/Peep.h
ZehMatt: I thought about it but until ride_id_t is not a unique type this is error prone, at some point maybe.

duncanspumpkin: It would be a little hard to work out what each type is for the import.

PR #15305, opened 2021-08-29
Fix #15289: Duplicated banner correction using the wrong index

### LINKED ISSUE #15289: Fix Banner Fix
The fix duplicate banners is broken and produces corrupted banners. 
https://github.com/OpenRCT2/OpenRCT2/blob/465d1c541446569fec7c627653ff5ea9543eea63/src/openrct2/world/Banner.cpp#L296
This bit of code overwrites the new banner id!
Modify the code to save the ID prior to copying over the other banner and then actually set the ID!

ZehMatt:
I also took the liberty of cleaning up the code a bit.

Closes #15289

### DIFF
--- a/src/openrct2/world/Banner.cpp
+++ b/src/openrct2/world/Banner.cpp
@@ -21,6 +21,7 @@
 #include "../ride/RideData.h"
 #include "../ride/Track.h"
 #include "../windows/Intent.h"
+#include "../world/TileElementsView.h"
 #include "Map.h"
 #include "MapAnimation.h"
 #include "Park.h"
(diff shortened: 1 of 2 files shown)


src/openrct2/world/Banner.cpp
```diff
@@ -253,57 +254,50 @@ void fix_duplicated_banners()
...
+                    log_info(
+                        "Duplicated banner with index %d found at x = %d, y = %d and z = %d.", bannerIndex, x, y,
+                        bannerElement->base_height);
+
+                    // Banner index is already in use by another banner, so duplicate it
+                    auto newBanner = CreateBanner();
+                    if (newBanner == nullptr)
+                    {
+                        log_error("Failed to create new banner.");
+                        continue;
```
ZehMatt, line 281: I think if it comes down to this it should probably just return, any further attempt will likely return nullptr.

Gymnasiast (approved):
Looks good, have you been able to test it?

ZehMatt:
> Looks good, have you been able to test it?

I did, I whacked the tile inspector to allow duplications and saved the park, it did fix them correctly.

Issue #14734, opened 2021-05-24
[.park format] Guests' time in park stats imported incorrectly from SC6 or SC4


Fred-1044276:
**OS:** Windows 10
**Version:** 0.3.3
**Commit/Build:** 0e13808 on refs/pull/10664/merge

If you start a new game in any scenario that has guests already in the park (e.g. Diamond Heights, Animatronic Antics 2), then click on any guest that's currently in the park, and check their stats (click the tab that shows their face), their "Time in park" stat will end up as something like 1,085hrs:23mins, which is very much incorrect because of an error made when importing SC6 or SC4 files. As a side note, if we go with how the game calculates guests' time spent in the park, 1,085 hours would obviously not be in one in-game month. On the other hand, loading up these same scenarios in a develop or release build will show the same guests with their correct time in park stat. In addition, even in the .park format test builds, guests that enter the park after the scenario is loaded will still correctly start with a "Time in park" stat of 0mins.

I loaded up a few SV6 files that already have guests in the park (e.g. matsku84's Extreme Heights) and none of them have this issue.

**Steps to reproduce:**
1. Start up a new game in any scenario that already has guests in the park (e.g. either Diamond Heights or the ones attached in the .zip file below)
2. Click on any one guest that's currently in the park
3. Check their stats by clicking on the face tab and you will see their "Time in park" stat has been massively inflated

**Screenshots / Video:**
![Luminous Landscapes 2021-05-21 10-23-47](https://user-images.githubusercontent.com/61885939/119378691-6368b700-bc73-11eb-8122-31404acfb0e5.png)
![Luminous Landscapes 2021-05-21 10-30-33](https://user-images.githubusercontent.com/61885939/119378704-66fc3e00-bc73-11eb-8931-8d12527408a9.png)
![Animatronic Antics 2 2021-05-21 10-35-45](https://user-images.githubusercontent.com/61885939/119378720-6d8ab580-bc73-11eb-8578-b4939fd44af0.png)
![Animatronic Antics 2 2021-05-21 10-34-59](https://user-images.githubusercontent.com/61885939/119378729-6ebbe280-bc73-11eb-8718-50915bbd5805.png)
As you can see, in the first and third pictures, the guests I have selected (except Guest 188 in Luminous Landscapes) have a "Time in park" stat that is significantly greater than it should be. In the second and fourth pictures, I have the same guests selected, but this time the screenshots are taken on the current develop and release builds, which show their correct "Time in park" stats.

[...]

PR #15308, opened 2021-08-29
Rename gScenarioTicks to gCurrentTicks

Gymnasiast:
Ported from the NSF. Will reduce the diff quite a bit.

Do **not** squash merge.

### DIFF
--- a/src/openrct2/GameState.cpp
+++ b/src/openrct2/GameState.cpp
@@ -367,7 +367,7 @@ void GameState::UpdateLogic(LogicTimings* timings)
     report_time(LogicTimePart::NetworkFlush);
 
     gCurrentTicks++;
-    gScenarioTicks++;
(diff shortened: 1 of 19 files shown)


src/openrct2/GameState.cpp
```diff
@@ -367,7 +367,7 @@ void GameState::UpdateLogic(LogicTimings* timings)
     report_time(LogicTimePart::NetworkFlush);
 
     gCurrentTicks++;
-    gScenarioTicks++;
+    gCurrentTicks++;
```
ZehMatt, line 370: That doesn't look right, should advance only one tick.

ZehMatt (changes requested):

ZehMatt:
Also please check if https://github.com/OpenRCT2/OpenRCT2/issues/14734 is alright.

PR #15309, opened 2021-08-29
Refactor/paths from nsf

Gymnasiast:
This PR mostly adds object loading code. The plan to add more in follow-up PRs - most of the stuff can be backported.


src/openrct2/object/FootpathRailingsObject.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    PopulateTablesFromJson(context, root);
+}
+
+RailingEntrySupportType FootpathRailingsObject::ParseSupportType(std::string_view s)
+{
+    if (s == "pole")
+        return RailingEntrySupportType::Pole;
+    else /* if (s == "box") */
```
ZehMatt, line 92: I think having this explicit would be a good idea, the else should be a Guard::Assert so its impossible to have invalid inputs.

Gymnasiast: I don’t think it’s a good idea to have the game _crash_ on an invalid object. It should just print a warning, reject the object and move on.

ZehMatt: Fair enough, just wanted to highlight a potential issue here that goes easily unnoticed like all the bad objects we already have with missing images and what not.

Gymnasiast: I think that is mostly with .DAT objects, but you’re right, we have the opportunity to be stricter with JSON objects.

Is it okay if I create an issue for that and pick it up later? I’d need to make some changes to how the code works to properly handle this kind of thing (rejecting the object from here).

ZehMatt (approved):
LGTM

src/openrct2/object/FootpathRailingsObject.cpp
ZehMatt: Yeah sure, I would normally request changes if I feel its necessary.

Broxzier:
This PR seems to have caused the Object Selection window to take a long time to open: https://github.com/OpenRCT2/OpenRCT2/issues/15319.

Gymnasiast:
I’m surprised to hear that. I have no idea which of the changes could even cause such a thing.

Broxzier:
Me neither. I've been placing breakpoints here and there, but don't get any wiser. Strangest thing is that in debug builds, it still opens instantly.

src/openrct2/object/Object.h
```diff
@@ -42,6 +42,8 @@ enum class ObjectType : uint8_t
     TerrainEdge,
     Station,
     Music,
+    FootpathSurface,
+    FootpathRailings,
```
Broxzier, line 46: Cause found, this adds two new types, but they are not present in `object_entry_group_counts`, so it reads out of bounds. In release builds that's garbage, possibly very high numbers, hence why it takes so long.

Gymnasiast:
Are you absolutely certain it is this commit and not one before/after?

src/openrct2/object/Object.h
Gymnasiast: Nice, shall I create a PR to fix?

PR #15312, opened 2021-08-29
Fix #14474: Map size can be synced in multiplayer

### LINKED ISSUE #14474: Map size isn't synced in multiplayer
**Version:** 0.3.3
**Commit/Build:** https://github.com/OpenRCT2/OpenRCT2/commit/b24cdff0fb2ad1db4dffb304a3f6cd22976b29d2

Setting the map size while sandbox mode is on in multiplayer leads to a desync.

**Steps to reproduce:**
1. Create a server.
2. Enable Sandbox mode.
3. Change map size.
4. Map size won't sync. (if you're the host players will have to rejoin)

**Screenshots / Video:**
image

frutiemax:
This feature seems to work perfectly fine when increasing the map size. However, the game seems to stall when decreasing the map size as it tries to remove elements from the park. I'm putting this as draft as it is too much game breaking.

### DIFF
--- a/src/openrct2-ui/windows/Map.cpp
+++ b/src/openrct2-ui/windows/Map.cpp
@@ -18,6 +18,7 @@
 #include <openrct2/Game.h>
 #include <openrct2/Input.h>
 #include <openrct2/OpenRCT2.h>
+#include <openrct2/actions/ChangeMapSizeAction.h>
 #include <openrct2/actions/LandSetRightsAction.h>
 #include <openrct2/actions/PlaceParkEntranceAction.h>
 #include <openrct2/actions/PlacePeepSpawnAction.h>
@@ -655,17 +656,8 @@ static void window_map_textinput(rct_window* w, rct_widgetindex widgetIndex, cha
                 size += 2;
                 size = std::clamp(size, MINIMUM_MAP_SIZE_TECHNICAL, MAXIMUM_MAP_SIZE_TECHNICAL);
 
-                int32_t currentSize = gMapSize;
-                while (size < currentSize)
-                {
-                    map_window_decrease_map_size();
-                    currentSize--;
-                }
-                while (size > currentSize)
-                {
-                    map_window_increase_map_size();
-                    currentSize++;
-                }
+                auto changeMapSizeAction = ChangeMapSizeAction(size);
+                GameActions::Execute(&changeMapSizeAction);
                 w->Invalidate();
             }
             break;
@@ -1370,17 +1362,8 @@ static void window_map_set_peep_spawn_tool_down(const ScreenCoordsXY& screenCoor
  */
 static void map_window_increase_map_size()
 {
-    if (gMapSize >= MAXIMUM_MAP_SIZE_TECHNICAL)
-    {
-        context_show_error(STR_CANT_INCREASE_MAP_SIZE_ANY_FURTHER, STR_NONE, {});
-        return;
-    }
-
-    gMapSize++;
-    map_extend_boundary_surface();
-    window_map_init_map();
-    window_map_centre_on_view_point();
-    gfx_invalidate_screen();
+    auto increaseMapSizeAction = ChangeMapSizeAction(gMapSize + 1);
+    GameActions::Execute(&increaseMapSizeAction);
 }
 
 /**
@@ -1389,17 +1372,8 @@ static void map_window_increase_map_size()
  */
 static void map_window_decrease_map_size()
 {
-    if (gMapSize < 16)
-    {
-        context_show_error(STR_CANT_DECREASE_MAP_SIZE_ANY_FURTHER, STR_NONE, {});
-        return;
-    }
-
-    gMapSize--;
-    map_remove_out_of_range_elements();
-    window_map_init_map();
-    window_map_centre_on_view_point();
-    gfx_invalidate_screen();
+    auto decreaseMapSizeAction = ChangeMapSizeAction(gMapSize - 1);
+    GameActions::Execute(&decreaseMapSizeAction);
 }
 
 static constexpr const uint16_t WaterColour = MapColour(PALETTE_INDEX_195);
--- a/src/openrct2/actions/ChangeMapSizeAction.cpp
+++ b/src/openrct2/actions/ChangeMapSizeAction.cpp
@@ -1,0 +1,77 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2021 OpenRCT2 developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#include "ChangeMapSizeAction.h"
+
+#include "../Context.h"
+#include "../drawing/IDrawingEngine.h"
+#include "../ui/UiContext.h"
+#include "../ui/WindowManager.h"
+#include "../windows/Intent.h"
+
+ChangeMapSizeAction::ChangeMapSizeAction(const int32_t targetSize)
+    : _targetSize(targetSize)
+{
+}
+
+uint16_t ChangeMapSizeAction::GetActionFlags() const
+{
+    return GameAction::GetActionFlags() | GameActions::Flags::AllowWhilePaused;
+}
+
+void ChangeMapSizeAction::Serialise(DataSerialiser& stream)
+{
+    GameAction::Serialise(stream);
+    stream << DS_TAG(_targetSize);
+}
+
+GameActions::Result::Ptr ChangeMapSizeAction::Query() const
+{
+    if (_targetSize >= MAXIMUM_MAP_SIZE_TECHNICAL)
+    {
+        return std::make_unique<GameActions::Result>(
+            GameActions::Status::InvalidParameters, STR_CANT_INCREASE_MAP_SIZE_ANY_FURTHER, STR_NONE);
+    }
+    if (_targetSize < 16)
+    {
+        return std::make_unique<GameActions::Result>(
+            GameActions::Status::InvalidParameters, STR_CANT_DECREASE_MAP_SIZE_ANY_FURTHER, STR_NONE);
+    }
+    return std::make_unique<GameActions::Result>();
+}
+
+GameActions::Result::Ptr ChangeMapSizeAction::Execute() const
+{
+    while (gMapSize != _targetSize)
+    {
+        if (_targetSize < gMapSize)
+        {
+            gMapSize--;
+            map_remove_out_of_range_elements();
+        }
+        else
+        {
+            gMapSize++;
+            map_extend_boundary_surface();
+        }
+    }
+
+    auto* ctx = OpenRCT2::GetContext();
+    auto uiContext = ctx->GetUiContext();
+    auto* windowManager = uiContext->GetWindowManager();
+
+    windowManager->BroadcastIntent(Intent(INTENT_ACTION_MAP));
+    gfx_invalidate_screen();
+    return std::make_unique<GameActions::Result>();
+}
+
+void ChangeMapSizeAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("targetSize", _targetSize);
+}
--- a/src/openrct2/actions/ChangeMapSizeAction.h
+++ b/src/openrct2/actions/ChangeMapSizeAction.h
@@ -1,0 +1,30 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2021 OpenRCT2 developers
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include "../world/Map.h"
+#include "GameAction.h"
+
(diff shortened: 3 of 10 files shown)


ocalhoun6:
Maybe that's good enough for the interim? I have to imagine that most of the time the feature is used, it's because people want to increase the map size anyway. Trying to increase the map size during multiplayer would probably be a lot more common than trying to decrease the map size during multiplayer, right?

Could replace the map size decreasing behavior with an error message explaining that the map size can't be decreased during multiplayer play (yet) and then leave it at that for now.

And then people would at least have _half_ the feature they wanted ... and probably the more important half.

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1389,13 +1389,12 @@ static void map_window_increase_map_size()
...
-        context_show_error(STR_CANT_DECREASE_MAP_SIZE_ANY_FURTHER, STR_NONE, {});
+    auto decreaseMapSizeAction = ChangeMapSizeAction(-1);
+    auto query = GameActions::Query(&decreaseMapSizeAction);
+    if (query->Error != GameActions::Status::Ok)
         return;
-    }
 
-    gMapSize--;
+    GameActions::Execute(&decreaseMapSizeAction);
     map_remove_out_of_range_elements();
```
duncanspumpkin, line 1398: All of this code should be in the game action. You can't just decrease/increase the map size global and not actually move the boundary.

duncanspumpkin:
Xcode is failing as you haven't added the new files to the project. Use ZehMatts editor to add the files to the pbxproj file https://github.com/ZehMatt/PBXEdit

Broxzier (commented):
This action should be added to the `NetworkActions::Actions` array in NetworkAction.cpp. Perhaps it could be added to "PERMISSION_PARK_PROPERTIES", or add an entirely new group.

src/openrct2/network/NetworkAction.cpp
```diff
@@ -235,10 +235,7 @@ const std::array<NetworkAction, static_cast<size_t>(NetworkPermission::Count)> N
     NetworkAction{
         STR_ACTION_CHEAT,
         "PERMISSION_CHEAT",
-        {
-            GameCommand::Cheat,
-            GameCommand::SetDate,
-        },
+        { GameCommand::Cheat, GameCommand::SetDate, GameCommand::ChangeMapSize },
```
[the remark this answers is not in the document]
frutiemax: I've put it in the Cheats section

frutiemax:
I've tried doing as @duncanspumpkin indicated, but it still stalls the server when decreasing the map size.

```
GameActions::Result::Ptr ChangeMapSizeAction::Execute() const
{
    if (_sizeChange > 0)
    {
        gMapSize++;
        map_extend_boundary_surface();
    }
    else
    {
        gMapSize--;
        map_remove_out_of_range_elements();
    }
    return std::make_unique<GameActions::Result>();
}
```

It really has to do with the existing elements in the park. For example, I load in the Create you own Six Flags scenario, increase the map size to 222, I can decrease it as long as I don't delete the peep spawn point it seems.

duncanspumpkin:
Your still not calling all the relevant code in the action. All map size changes must invalidate the screen of every client, reset every clients map window (if open),...

src/openrct2/actions/ChangeMapSizeAction.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "ChangeMapSizeAction.h"
+
+ChangeMapSizeAction::ChangeMapSizeAction(const int32_t sizeChange)
+    : _sizeChange(sizeChange)
+{
+}
+
+uint16_t ChangeMapSizeAction::GetActionFlags() const
+{
+    return GameAction::GetActionFlags() | GameActions::Flags::AllowWhilePaused;
```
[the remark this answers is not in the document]
duncanspumpkin: For cheats like this that aren't accessible except through sandbox mode I'm not too fussed if it's enabled with pause

duncanspumpkin:
As for peep spawns have you checked to see if it worked correctly before this pr. I don't see any code that moves them but i could be wrong

Broxzier:
I don't think that 'cheats' is the right group for this. As a server owner I'd give the players the ability to add money, prevent rides from breaking y, but wouldn't want to have them resize the map. I think that "park properties" would be a more fit group for this, or even better a whole new group for sandbox actions.

ZehMatt:
Also this PR is not adding a new feature as we can already resize the map. Your PR rather fixes an issue that the resize is not synchronized in multiplayer.

src/openrct2/actions/ChangeMapSizeAction.cpp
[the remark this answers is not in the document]
ZehMatt: The current behavior doesn't care if the game is paused or not, I would keep it as is.

src/openrct2/actions/ChangeMapSizeAction.cpp
```diff
@@ -0,0 +1,67 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2020 OpenRCT2 developers
```
AaronVanGeffen, line 2: 2021

src/openrct2/actions/ChangeMapSizeAction.h
```diff
@@ -0,0 +1,29 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2020 OpenRCT2 developers
```
AaronVanGeffen, line 2: 2021

frutiemax:
The game stalls in the current develop version when decreasing the map size.

ZehMatt:
> The game stalls in the current develop version when decreasing the map size.

Just tried, not happening here. Can you provide the map where I can investigate this? Also are you testing this on multiplayer?

frutiemax:
Here are the steps:

1. Start server on Six Flags Belgium
2. Enable Sandbox mode
3. Decrease map size. It stalls when you reach 117x117

frutiemax:
OK, I fixed the bug: I've modified the GameActions::Execute to GameActions::ExecuteNested when clearing elements since the actions were piling up.

src/openrct2/world/Map.cpp
```diff
@@ -1741,7 +1741,7 @@ static void clear_element_at(const CoordsXY& loc, TileElement** elementPtr)
                     break;
             }
             auto parkEntranceRemoveAction = ParkEntranceRemoveAction(CoordsXYZ{ seqLoc, element->GetBaseZ() });
-            auto result = GameActions::Execute(&parkEntranceRemoveAction);
+            auto result = GameActions::ExecuteNested(&parkEntranceRemoveAction);
```
ZehMatt, line 1744: Have you made sure that clear_element_at is not called from outside a game action? Otherwise this will break current code.

frutiemax:
It's used in the map generation code, but it doesn't seem to be affected by these changes.

src/openrct2-ui/windows/Map.cpp
```diff
@@ -1389,17 +1384,11 @@ static void map_window_increase_map_size()
  */
 static void map_window_decrease_map_size()
 {
-    if (gMapSize < 16)
-    {
-        context_show_error(STR_CANT_DECREASE_MAP_SIZE_ANY_FURTHER, STR_NONE, {});
+    auto decreaseMapSizeAction = ChangeMapSizeAction(-1);
+    auto queryResult = GameActions::Query(&decreaseMapSizeAction);
+    if (queryResult->Error != GameActions::Status::Ok)
```
duncanspumpkin, line 1389: Ummm what? Calling GameActions::Execute calls a Query already and only if the query fails then it does the execute. What you have written now performs the query twice then the execute.

src/openrct2/world/Map.cpp
duncanspumpkin: Its used by `map_init`, `decrease_map_size` (this PR), `mapgen_generate_blank`, `mapgen_generate`. All of which are single player (apart from obviously this PR). Its safe to perform a nested from singleplayer instances although it isn't pretty. Alternative would be to add a new mode to the `ClearAction` that performs a full wipe of a tile including park entrances and peep spawns. This can be left for a follow on PR though. An issue is to be created for this though so its not forgotten otherwise some person in the future might call this action at the wrong time.

src/openrct2/actions/ChangeMapSizeAction.h
```diff
@@ -0,0 +1,30 @@
...
+#include "GameAction.h"
+
+DEFINE_GAME_ACTION(ChangeMapSizeAction, GameCommand::ChangeMapSize, GameActions::Result)
+{
+private:
+    int32_t _sizeChange;
+
+public:
+    ChangeMapSizeAction() = default;
+    ChangeMapSizeAction(const int32_t sizeChange);
```
ZehMatt, line 22: I'm not a fan of this parameter, this should be enum class, there are only two possible inputs which is Increase, Decrease, passing -1 or 1 is just confusing, need to also consider that plugins may get access.

ZehMatt (changes requested):

src/openrct2/actions/ChangeMapSizeAction.h
duncanspumpkin: For plugins you would probably just want to specify the end size you want to get to. In general that is how the rest of the game actions work (perhaps not all of them though).

ZehMatt: I also thought about just specifying the size instead, you can work out from current size if its decreasing or increasing but we can do that afterwards I guess.

src/openrct2/network/NetworkAction.cpp
```diff
@@ -191,16 +191,9 @@ const std::array<NetworkAction, static_cast<size_t>(NetworkPermission::Count)> N
...
-            GameCommand::SetParkEntranceFee,
-            GameCommand::SetLandOwnership,
-            GameCommand::BuyLandRights,
-            GameCommand::PlaceParkEntrance,
-            GameCommand::RemoveParkEntrance,
-            GameCommand::PlacePeepSpawn,
-        },
+        { GameCommand::SetParkName, GameCommand::SetParkOpen, GameCommand::SetParkEntranceFee, GameCommand::SetLandOwnership,
+          GameCommand::BuyLandRights, GameCommand::PlaceParkEntrance, GameCommand::RemoveParkEntrance,
+          GameCommand::PlacePeepSpawn, GameCommand::ChangeMapSize },
```
ZehMatt, line 196:
```suggestion
          GameCommand::PlacePeepSpawn, GameCommand::ChangeMapSize, },
```
This should make clang-format add them on each line again

src/openrct2/network/NetworkAction.cpp
```diff
@@ -235,10 +228,7 @@ const std::array<NetworkAction, static_cast<size_t>(NetworkPermission::Count)> N
     NetworkAction{
         STR_ACTION_CHEAT,
         "PERMISSION_CHEAT",
-        {
-            GameCommand::Cheat,
-            GameCommand::SetDate,
-        },
+        { GameCommand::Cheat, GameCommand::SetDate },
```
ZehMatt, line 231:
```suggestion
        { GameCommand::Cheat, GameCommand::SetDate, },
```

ZehMatt (changes requested):
Just a couple minor things otherwise I think it looks good

src/openrct2/actions/ChangeMapSizeAction.cpp
```diff
@@ -0,0 +1,80 @@
...
+    auto* windowManager = uiContext->GetWindowManager();
+
+    windowManager->BroadcastIntent(Intent(INTENT_ACTION_MAP));
+    gfx_invalidate_screen();
+    return std::make_unique<GameActions::Result>();
+}
+
+void ChangeMapSizeAction::AcceptParameters(GameActionParameterVisitor& visitor)
+{
+    visitor.Visit("sizeChange", _sizeChange);
```
IntelOrca, line 79: How does this get serialised? Does it become, 0 or 1?

src/openrct2/actions/ChangeMapSizeAction.h
IntelOrca: I like the idea of specifying the size. It is more future proof, seems like a simpler API for the action and might help resolve conflicts where two players change the map size at the same time.

src/openrct2/actions/ChangeMapSizeAction.cpp
frutiemax: It is, these are the default values for an enum. I can change it to -1, 1.

src/openrct2/actions/ChangeMapSizeAction.h
duncanspumpkin: so what is the conclusion change the API as part of this PR?

ZehMatt: Well if @frutiemax is willing to refactor this PR I would be for the target size parameter instead of delta changes.

ZehMatt:
The code looks alright but I wanna test this, I'll do it later today.

ZehMatt (changes requested):
The game action its self seems to be working however you need to fix the the code for the map size text input 
See https://github.com/OpenRCT2/OpenRCT2/blob/e6aadbba9f18cc776dc65649386477bb73f66249/src/openrct2-ui/windows/Map.cpp#L659-L669
The entire while loop is no longer correct, it just needs to use the game action with the new size.
Its best you turn the increase/decrease into a single function with the new size argument, then you can use the same function for all 3 cases.

ZehMatt (approved):
Sorry I sort of forgot about this PR, I rebased it and raised the network version. I did test the changes and everything works as expected without desyncs 👍

PR #15315, opened 2021-08-30
Unlock setting seat rotation on brakes

spacek531:
The limitation of not being able to set seat rotation on brake track is due to TD6. Since NSF is sunsetting TD6, this functionality can be added.


duncanspumpkin:
Td6 should be replaced prior to performing this.

src/openrct2/world/TileElement.h
```diff
@@ -350,7 +350,11 @@ struct TrackElement : TileElementBase
         struct
         {
             uint8_t Sequence;
-            uint8_t ColourScheme;
+            union
+            {
+                uint8_t ColourScheme;
+                uint8_t SeatRotation;
+            };
```
Gymnasiast, line 357: This change is unnecessary and hides the fact that these share bits within a field. You can add a comment detailing which bits are what if you like.

Gymnasiast:
This PR is opened against `new-save-format`, has quite a lot of merge conflicts and cannot be merged before the NTDF is here. For this reason, I’m closing it, but feel free to create a new one once the NTDF is here.

PR #15317, opened 2021-08-31
even fewer platforms

spacek531:
Some flat ride platforms aren't quite no platform.
image


Gymnasiast:
The changes are welcome, but should be done against `develop`. `develop` already uses station objects internally, it’s just that there is no way to change the selection yet. (You can test by unhiding the Stations tab in EditorObjectSelection.cpp and selecting the style.)

The diff can most likely be applied to `develop` with no trouble.

Gymnasiast (approved):

PR #15321, opened 2021-09-01
Use music objects instead of IDs

Gymnasiast:
Backported from the NSF.

(This one could cut quite a lot of files from the NSF diff!)

Do not squash merge!

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -859,25 +859,33 @@ namespace RCT1
             // RCT1 used 5mph / 8 km/h for every lift hill
             dst->lift_hill_speed = 5;
 
-            if (_gameVersion == FILE_VERSION_RCT1)
+            dst->music = OBJECT_ENTRY_INDEX_NULL;
+            if (GetRideTypeDescriptor(dst->type).HasFlag(RIDE_TYPE_FLAG_ALLOW_MUSIC))
             {
-                // Original RCT had no music settings, take default style
-                dst->music = GetRideTypeDescriptor(dst->type).DefaultMusic;
-
-                // Only merry-go-round and dodgems had music and used
-                // the same flag as synchronise stations for the option to enable it
-                if (src->type == RCT1_RIDE_TYPE_MERRY_GO_ROUND || src->type == RCT1_RIDE_TYPE_DODGEMS)
+                if (_gameVersion == FILE_VERSION_RCT1)
                 {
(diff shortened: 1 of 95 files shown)


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1014,8 +1022,20 @@ namespace RCT1
                     dst->track_colour[i].additional = RCT1::GetColour(src->track_colour_additional[i]);
                     dst->track_colour[i].supports = RCT1::GetColour(src->track_colour_supports[i]);
                 }
+            }
+
+            dst->entrance_style = OBJECT_ENTRY_INDEX_NULL;
```
Gymnasiast, line 1027: revert this bit

src/openrct2/ride/gentle/meta/Circus.h
```diff
@@ -43,7 +43,7 @@ constexpr const RideTypeDescriptor CircusRTD =
     SET_FIELD(UpkeepCosts, { 50, 1, 0, 0, 0, 0 }),
     SET_FIELD(BuildCosts, { 125, 2, 1, }),
     SET_FIELD(DefaultPrices, { 15, 0 }),
-    SET_FIELD(DefaultMusic, MUSIC_STYLE_CIRCUS_SHOW),
+    SET_FIELD(DefaultMusic, {}),
```
Gymnasiast, line 46: As I expected, this breaks the circus sounds.

Broxzier (commented):
Haven't checked every single line, but the changes look good to me. One thing I did find while testing is that Custom 1 and Custom 2 did show up in the dropdown menu after adding the, but nothing was played when selecting them. I had to restart the game before they worked. This was probably already an issue before, not introduced by this PR.

Broxzier (approved):
Just confirmed that the above issue was already present before this PR.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -800,7 +800,12 @@ class S6Importer final : public IParkImporter
             }
         }
 
-        dst->music = src->music;
+        auto musicStyle = OBJECT_ENTRY_INDEX_NULL;
+        if (GetRideTypeDescriptor(dst->type).HasFlag(RIDE_TYPE_FLAG_ALLOW_MUSIC))
+        {
+            musicStyle = src->music;
+        }
+        dst->music = musicStyle;
```
duncanspumpkin, line 808: I dont see the reverse in S6Exporter. I assume its not required as it truncates fine?

PR #15324, opened 2021-09-01
Enable parallel builds for debug configuration

ZehMatt:
Visual Studio 2022 doesn't build in parallel because of this


Broxzier:
The title says debug configuration, but the change looks like it's also done for release?

ZehMatt:
> The title says debug configuration, but the change looks like it's also done for release?

It was building with MP without it but I added it for consistency.

Gymnasiast (approved):

PR #7449, opened 2018-04-25
Fix #7436: Only the first 32 vehicles of a train can be painted

Gymnasiast:
This also raises MAX_CARS_PER_TRAIN to 255, which it already was, in practical terms. When exporting to S6, only the colours of the first 32 vehicles are saved (this PR does not touch that).

PR #15329, opened 2021-09-02
Introduce Limits headers.

duncanspumpkin:
Idea is to have Limits headers one for each of RCT1, RCT2, RCT12, OpenRCT2, (Classic?).
The RCT1 and RCT2 headers would mostly inherit RCT12.
Outside of the Limits headers RCT1/RCT2 should only be referenced within the RCT1/RCT2 folders. (RCT12 likewise for RCT12 folder).
The OpenRCT2 limits header should be used for everything else.
OpenRCT2 would inherit mostly from RCT2 and would be where all of the park file limits would be changed at.

Follows the same design as I've introduced in OpenLoco https://github.com/OpenLoco/OpenLoco/pull/1129


duncanspumpkin:
That is the next step. I didn't have time this morning.

Broxzier:
This may be a good opportunity to use [inline namespaces](https://www.stroustrup.com/C++11FAQ.html#inline-namespace), so that we have one `Limits` namespace containing sub-namespaces for RCT1/2/OpenRCT2/Classic, and then make OpenRCT2's the default. Using them would change `RCT12::Limits::...` to `Limits::RCT12::...`.

With the `OpenRCT2` namespace being inlined, `Limits::MaxStationsPerRide` would refer to OpenRCT2's limit, and `Limits::RCT1::MaxStationsPerRide` would be RCT1's for instance.

ZehMatt:
Last time this idea was rejected, I'm all for it.

duncanspumpkin:
> This may be a good opportunity to use [inline namespaces](https://www.stroustrup.com/C++11FAQ.html#inline-namespace), so that we have one `Limits` namespace containing sub-namespaces for RCT1/2/OpenRCT2/Classic, and then make OpenRCT2's the default. Using them would change `RCT12::Limits::...` to `Limits::RCT12::...`.
> 
> With the `OpenRCT2` namespace being inlined, `Limits::MaxStationsPerRide` would refer to OpenRCT2's limit, and `Limits::RCT1::MaxStationsPerRide` would be RCT1's for instance.
> 
> Edit: This would also be more future-proof in case we change the limits later on in the project. Then a copy of the OpenRCT2 namespace would be made with some version number (`Limits::OpenRCT2v1` or whatever), and where you want to use the outdated size, you have to explicitly state that, while all other places would use the updated inline OpenRCT2 namespace.

Yeah thats basically the same idea I have but I was thinking it would be `OpenRCT2::Limits` `RCT1::Limits` so on so on.

ZehMatt:
> > This may be a good opportunity to use [inline namespaces](https://www.stroustrup.com/C++11FAQ.html#inline-namespace), so that we have one `Limits` namespace containing sub-namespaces for RCT1/2/OpenRCT2/Classic, and then make OpenRCT2's the default. Using them would change `RCT12::Limits::...` to `Limits::RCT12::...`.
> > With the `OpenRCT2` namespace being inlined, `Limits::MaxStationsPerRide` would refer to OpenRCT2's limit, and `Limits::RCT1::MaxStationsPerRide` would be RCT1's for instance.
> > Edit: This would also be more future-proof in case we change the limits later on in the project. Then a copy of the OpenRCT2 namespace would be made with some version number (`Limits::OpenRCT2v1` or whatever), and where you want to use the outdated size, you have to explicitly state that, while all other places would use the updated inline OpenRCT2 namespace.
> 
> Yeah thats basically the same idea I have but I was thinking it would be `OpenRCT2::Limits` `RCT1::Limits` so on so on.

I think that makes sense considering we are also moving the specific structs into the same namespace.

duncanspumpkin:
For reference #7449 (introduced desync hack). #10504 (lowered vehicle limit)

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2717,7 +2717,7 @@ static void window_ride_vehicle_mousedown(rct_window* w, rct_widgetindex widgetI
             window_ride_show_vehicle_type_dropdown(w, &w->widgets[widgetIndex]);
             break;
         case WIDX_VEHICLE_TRAINS_INCREASE:
-            if (ride->num_vehicles < MAX_VEHICLES_PER_RIDE)
+            if (ride->num_vehicles < OpenRCT2::Limits::MaxTrainsPerRide)
```
Gymnasiast, line 2720:
```suggestion
            if (ride->num_vehicles < OpenRCT2::Limits::MaxVehiclesPerRide)
```

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2717,7 +2717,7 @@ static void window_ride_vehicle_mousedown(rct_window* w, rct_widgetindex widgetI
...
                 ride->SetNumVehicles(ride->num_vehicles + 1);
```
Gymnasiast, line 2721: This illustrates that `vehicle` is very ambiguous, sometimes referring to cars and sometimes to trains. We should make an issue to rename methods and fields to be unambiguous.

src/openrct2/config/Config.cpp
```diff
@@ -1001,7 +1001,8 @@ bool RCT1DataPresentAtLocation(const utf8* path)
 
 bool CsgIsUsable(const rct_gx& csg)
 {
-    return csg.header.total_size == RCT1::RCT1_LL_CSG1_DAT_FILE_SIZE && csg.header.num_entries == RCT1::RCT1_NUM_LL_CSG_ENTRIES;
+    return csg.header.total_size == RCT1::Limits::LL_CSG1_DAT_FileSize
+        && csg.header.num_entries == RCT1::Limits::Num_LL_CSG_Entries;
```
Gymnasiast, line 1005: Is this a limit, per se?

src/openrct2/management/Research.h
```diff
@@ -11,6 +11,7 @@
 
 #include "../common.h"
 #include "../object/ObjectLimits.h"
+#include "../rct2/RCT2.h"
```
Gymnasiast, line 14: Stray change

src/openrct2/object/RideObject.cpp
```diff
@@ -124,7 +124,7 @@ void RideObject::ReadLegacy(IReadObjectContext* context, IStream* stream)
     }
 
     // Read peep loading positions
-    for (int32_t i = 0; i < RCT2_MAX_VEHICLES_PER_RIDE_ENTRY; i++)
+    for (int32_t i = 0; i < RCT2::Limits::MaxVehiclesPerRideEntry; i++)
```
Gymnasiast, line 127: I wonder if we should call these "RCT2 limits" or "DAT limits". I think the latter might be clearer.

src/openrct2/config/Config.cpp
duncanspumpkin: Yeah i was unsure what to do with these constants. I'm somewhat contemplating another header just for special constants.

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -829,11 +829,11 @@ namespace RCT1
             dst->num_stations = src->num_stations;
 
             // Vehicle links (indexes converted later)
-            for (int32_t i = 0; i < RCT1_MAX_TRAINS_PER_RIDE; i++)
+            for (int32_t i = 0; i < Limits::MaxTrainsPerRide; i++)
             {
                 dst->vehicles[i] = src->vehicles[i];
             }
-            for (int32_t i = RCT1_MAX_TRAINS_PER_RIDE; i <= MAX_VEHICLES_PER_RIDE; i++)
+            for (int32_t i = Limits::MaxTrainsPerRide; i < OpenRCT2::Limits::MaxTrainsPerRide; i++)
```
Gymnasiast, line 836: Don’t change the `<=`, it is there for a reason - to properly mark all vehicles as NULL.

If you absolutely must change this line, use std::size instead.

src/openrct2/management/Research.h
duncanspumpkin: It's actually required. But really should not be. The research should do the conversion in the rct2 files.

src/openrct2/rct1/S4Importer.cpp
duncanspumpkin: The change is correct. See my comments in ride.h

src/openrct2/ride/Ride.h
```diff
@@ -223,7 +216,7 @@ struct Ride
     std::string custom_name;
     uint16_t default_name_number;
     CoordsXY overall_view;
-    uint16_t vehicles[MAX_VEHICLES_PER_RIDE + 1]; // Points to the first car in the train
+    uint16_t vehicles[OpenRCT2::Limits::MaxTrainsPerRide]; // Points to the first car in the train
```
Gymnasiast, line 219: Why did you make this lower?

Even if it is correct, it’s unrelated to this PR.

duncanspumpkin: As stated in my comment the last entry was never used and always null. It made for a bad design as you had to have <= instead of < in many places. It now becomes clear that the size is the size and not a bit smaller than the size and we can use all the normal std lib functions on the array and not have to worry about the size not being the size

Gymnasiast: I’d prefer if this was done in a separate PR - it’s not related to this one.

src/openrct2/object/RideObject.cpp
duncanspumpkin: Yeah I'm of a similar mind. It will be easy enough to move things about after this pr.

src/openrct2/ride/Ride.h
duncanspumpkin: Hmm okay it just looks so hacky though.

duncanspumpkin: Issue for this has been created will revert changes for now.

src/openrct2/actions/TrackDesignAction.cpp
```diff
@@ -252,14 +252,14 @@ GameActions::Result::Ptr TrackDesignAction::Execute() const
...
-    for (int32_t i = 0; i < RCT12_NUM_COLOUR_SCHEMES; i++)
+    for (int32_t i = 0; i < RCT12::Limits::NumColourSchemes; i++)
     {
         ride->track_colour[i].main = _td.track_spine_colour[i];
         ride->track_colour[i].additional = _td.track_rail_colour[i];
         ride->track_colour[i].supports = _td.track_support_colour[i];
     }
 
-    for (size_t i = 0; i <= MAX_VEHICLES_PER_RIDE; i++)
+    for (size_t i = 0; i < RCT2::Limits::MaxVehicleColours; i++)
```
duncanspumpkin, line 262: This was just the wrong limit.

duncanspumpkin:
This can now be resurrected.

duncanspumpkin:
Okay i've reverted my work on introducing an OpenRCT2 limits header. That will now be a new PR.

ZehMatt (approved):

PR #15335, opened 2021-09-03
Fix ownership of loaded object data

### LINKED ISSUE #13912: Dome park no longer renders dome correctly
**Version:** 0.3.2
**Commit/Build:** develop

Broken in 0.3.2+:
openrct2_2021-01-23_00-12-4553a959da67e3725e5

Working in 0.3.1:
openrct2_2021-03-13_09-57-30b157fb33745139980

Park
[dome-export.zip](https://github.com/OpenRCT2/OpenRCT2/files/6134362/dome-export.zip)

ZehMatt:
This also fixes #13912, gonna append some minor cleanups.

### DIFF
--- a/src/openrct2/object/ObjectManager.cpp
+++ b/src/openrct2/object/ObjectManager.cpp
@@ -469,6 +469,18 @@ private:
         log_verbose("%u / %u objects unloaded", numObjectsUnloaded, totalObjectsLoaded);
     }
 
+    template<typename T> static void UpdateSceneryGroupIndexes(Object* object)
+    {
+        auto* sceneryEntry = static_cast<T*>(object->GetLegacyData());
+        sceneryEntry->scenery_tab_id = GetPrimarySceneryGroupEntryIndex(loadedObject);
+    }
+
+    template<> static void UpdateSceneryGroupIndexes<SceneryGroupObject>(Object* object)
+    {
+        auto sgObject = dynamic_cast<SceneryGroupObject*>(object);
+        sgObject->UpdateEntryIndexes();
+    }
+
     void UpdateSceneryGroupIndexes()
     {
         for (auto& loadedObject : _loadedObjects)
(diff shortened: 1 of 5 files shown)


src/openrct2/object/ObjectManager.cpp
```diff
@@ -396,16 +398,15 @@ class ObjectManager final : public IObjectManager
         Guard::ArgumentNotNull(object, GUARD_LINE);
 
         auto result = std::numeric_limits<size_t>().max();
-        auto it = std::find_if(
-            _loadedObjects.begin(), _loadedObjects.end(), [object](auto& obj) { return obj.get() == object; });
+        auto it = std::find_if(_loadedObjects.begin(), _loadedObjects.end(), [object](auto& obj) { return obj == object; });
```
Broxzier, line 401: I think this can be replaced by a simple `std::find` call, since all it does it compare pointers.

ZehMatt: True

src/openrct2/object/ObjectManager.cpp
```diff
@@ -120,14 +119,17 @@ class ObjectManager final : public IObjectManager
         auto requiredObjects = GetRequiredObjects(entries, count);
 
         // Load the required objects
-        size_t numNewLoadedObjects = 0;
-        auto loadedObjects = LoadObjects(requiredObjects, &numNewLoadedObjects);
+        {
+            auto loadedObjects = LoadObjects(requiredObjects);
```
duncanspumpkin, line 123: why even have the intermediate? saves having to write std::move

duncanspumpkin:
Its a super minor thing but could you go through and make sure you are using `auto*` when the parameter is a pointer. I think after replacing a few of those unique ptrs with ptrs it needs modification.

src/openrct2/object/ObjectManager.cpp
ZehMatt: I removed also the setter, don't see a reason to have it.

duncanspumpkin:
Appears to work fine from my testing. Loaded up a few different maps and deleted/added some objects.

duncanspumpkin (approved):

PR #15338, opened 2021-09-03
Use real values for gGuestsInParkHistory (from NSF)

Gymnasiast:
This one is nice and simple for a change :D

You know the drill: do not squash merge.

### DIFF
--- a/src/openrct2-ui/windows/Park.cpp
+++ b/src/openrct2-ui/windows/Park.cpp
@@ -1065,7 +1065,20 @@ static void window_park_guests_paint(rct_window* w, rct_drawpixelinfo* dpi)
     // Graph
     screenPos = w->windowPos + ScreenCoordsXY{ widget->left + 47, widget->top + 26 };
 
-    Graph::Draw(dpi, gGuestsInParkHistory, 32, screenPos);
+    uint8_t cappedHistory[32];
+    for (size_t i = 0; i < std::size(cappedHistory); i++)
+    {
+        auto value = gGuestsInParkHistory[i];
+        if (value != std::numeric_limits<uint32_t>::max())
+        {
+            cappedHistory[i] = static_cast<uint8_t>(std::min<uint32_t>(value, 5000) / 20);
+        }
+        else
+        {
+            cappedHistory[i] = std::numeric_limits<uint8_t>::max();
+        }
+    }
+    Graph::Draw(dpi, cappedHistory, static_cast<int32_t>(std::size(cappedHistory)), screenPos);
 }
 
 #pragma endregion
--- a/src/openrct2/rct2/S6Exporter.cpp
+++ b/src/openrct2/rct2/S6Exporter.cpp
@@ -351,7 +351,7 @@ void S6Exporter::Export()
 
     _s6.park_rating = gParkRating;
 
-    std::memcpy(_s6.park_rating_history, gParkRatingHistory, sizeof(_s6.park_rating_history));
+    std::copy(std::begin(gParkRatingHistory), std::end(gParkRatingHistory), _s6.park_rating_history);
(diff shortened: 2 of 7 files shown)


src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -352,6 +352,15 @@ void S6Exporter::Export()
     _s6.park_rating = gParkRating;
 
     std::memcpy(_s6.park_rating_history, gParkRatingHistory, sizeof(_s6.park_rating_history));
```
ZehMatt, line 354: Also this line would be better as
```suggestion
        std::copy(std::begin(gParkRatingHistory), std::end(gParkRatingHistory), _s6.park_rating_history);
```
If someone changes with the type this will no longer compile making it safe.

ZehMatt (changes requested):

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1065,7 +1065,20 @@ static void window_park_guests_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
     screenPos = w->windowPos + ScreenCoordsXY{ widget->left + 47, widget->top + 26 };
 
-    Graph::Draw(dpi, gGuestsInParkHistory, 32, screenPos);
+    uint8_t cappedHistory[32];
+    for (size_t i = 0; i < std::size(cappedHistory); i++)
+    {
+        auto value = gGuestsInParkHistory[i];
+        if (value != std::numeric_limits<uint32_t>::max())
+        {
+            cappedHistory[i] = static_cast<uint8_t>(std::min<uint32_t>(value, 5000) / 20);
```
ZehMatt, line 1074: I'm not sure I fully understand why 5000 was picked but this whole forth and back is a remap.
0-5000 is 0 to 250 and vice versa. 

So shouldn't 5000 be the maximum amount of guests possible?

src/openrct2-ui/windows/Park.cpp
```diff
@@ -1065,7 +1065,20 @@ static void window_park_guests_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
+        }
+        else
+        {
+            cappedHistory[i] = std::numeric_limits<uint8_t>::max();
+        }
+    }
+    Graph::Draw(dpi, cappedHistory, static_cast<int32_t>(std::size(cappedHistory)), screenPos);
```
ZehMatt, line 1081: Also we could skip remapping the value if we tell the graph what the max value is, so all it would need is the actual value and the maximum.

ZehMatt (approved):
The code looks good, I left few questions/remarks.

src/openrct2-ui/windows/Park.cpp
Gymnasiast: Not really, even in vanilla you can have more than 5000 guests (~9400 is the maximum). It’s just a nice round value that (divided by 20) just fitted in an uint8_t.

The graph drawing code should be modified to be able to take higher limits. I’ll create an issue for it.

src/openrct2-ui/windows/Park.cpp
Gymnasiast: Certainly. I’ll make sure to put that into the issue I’ll be creating.

src/openrct2-ui/windows/Park.cpp
ZehMatt: Well the graph will be off quite a bit it only does the first 5000, with NSF we can have a huge amount of peeps so this should be probably some sort of constant. 
We should also have a remap function that does remap(value, min1, max1, min1, max1), which would work for any range.

Gymnasiast: https://github.com/OpenRCT2/OpenRCT2/issues/15350

PR #15340, opened 2021-09-03
 Addendum to #15340: guard size of object_entry_group_counts

### LINKED ISSUE #15319: Clicking on Object Selection Cheat Freezes the Game
### Operating System

Windows 8

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (0be5850 on develop) provided by GitHub

### Describe the issue

When I click on the cheat "Object selection", it simply makes the game freeze. The game doesn't respond anymore.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Open any park or scenario.
2. Click on the cheat icon.
3. Select "Object selection"

### Attachments

image

Gymnasiast:


src/openrct2/object/ObjectList.cpp
```diff
@@ -38,6 +38,8 @@ int32_t object_entry_group_counts[] = {
     MAX_TERRAIN_EDGE_OBJECTS,
     MAX_STATION_OBJECTS,
     MAX_MUSIC_OBJECTS,
+    MAX_FOOTPATH_SURFACE_OBJECTS,
+    MAX_FOOTPATH_RAILINGS_OBJECTS,
```
ZehMatt, line 42: Please include all the limits even if they are zero, this should reflect all limits combined.

ZehMatt (changes requested):

src/openrct2/object/ObjectList.cpp
Gymnasiast: Uhm, this _does_ include all the limits...

ZehMatt: Sorry I got confused about the two tables, the other one was lacking a few.

duncanspumpkin (approved):

src/openrct2/object/ObjectList.cpp
```diff
@@ -22,7 +22,7 @@
 
 // 98DA00
 // clang-format off
-int32_t object_entry_group_counts[] = {
+int32_t object_entry_group_counts[EnumValue(ObjectType::Count)] = {
```
Broxzier, line 25: This will zero-initialize new entries as `Count` increases, and would only error in case this array exceeds `Count`. It's best to leave the brackets empty, and instead use a static_assert to verify the size.

```cpp
static_assert(std::size(object_entry_group_counts) == EnumValue(ObjectType::Count));
```

test/testpaint/Compat.cpp
```diff
@@ -71,18 +71,24 @@ uint8_t get_current_rotation()
...
-    128, // large scenery
-    128, // walls
-    32,  // banners
-    16,  // paths
-    15,  // path bits
-    19,  // scenery sets
-    1,   // park entrance
-    1,   // water
-    1    // scenario text
+int object_entry_group_counts[ObjectType::Count] = {
```
Broxzier, line 74: Same here.

Broxzier (changes requested):

src/openrct2/object/ObjectList.cpp
Broxzier: You'll need to include `<array>` or another header that defined `std::size`.

Broxzier (approved):

PR #15345, opened 2021-09-05
Add ability to block supports on individual track tiles

duncanspumpkin:
Note this does not save its state and will be gated behind the NSF in a future commit
This is a bit of proof of concept and to generate some discussion. I'm not saying we should/shouldn't add this or even do it like this just an idea really. See below video of it in use (ignore the interface it was before I had changed the string)

https://user-images.githubusercontent.com/1277401/132120197-172bc065-7a87-42bf-99aa-32ed651ebe1a.mp4


duncanspumpkin:
After some discussion on NE about it. It was identified that RCT1 tries to draw supports on every tile. Therefore it would probably make sense to just have a flag to indicate if supports should try be drawn. I'll have a look at feasibility of this. Would require moving a whole bunch of checks out of the drawing code and into import (which in general I'm in favour of).

Gymnasiast:
Nice idea :+1:

Broxzier:
Looks promising. I _support_ the idea of every track being able to have supports. When we get there, instead of it being a flag `TRACK_ELEMENT_FLAGS2_BLOCK_SUPPORTS` it'd be nicer to have it inversed like `TRACK_ELEMENT_FLAGS2_HAS_SUPPORTS`.

Gymnasiast:
> Looks promising. I _support_ the idea of every track being able to have supports. When we get there, instead of it being a flag `TRACK_ELEMENT_FLAGS2_BLOCK_SUPPORTS` it'd be nicer to have it inversed like `TRACK_ELEMENT_FLAGS2_HAS_SUPPORTS`.

Not too fussed either way, but an advantage of having 0 = supports and 1 = no supports is that it defaults to the most common occurrence.

spacek531:
I prefer the 1 = supports 0 = no supports notation because it slots well with `track_paint_util_should_track_have_supports` and it slots well with a checkbox labelled "Has supports" in the tile inspector.

ZehMatt:
> I prefer the 1 = supports 0 = no supports notation because it slots well with `track_paint_util_should_track_have_supports` and it slots well with a checkbox labelled "Has supports" in the tile inspector. I'd also say that having supports is not actually the most common occurrence, or if it is only by a slim margin.

The default value for the bit is currently/likely to be zero, if it imports from current saves you would invert all the supports. It should be "Supports disabled/blocked" and not "Supports enabled/visible", having no supports is rather the exception than the norm.

Broxzier:
There is no default value as of now, and to me it makes more sense to have a value of 0 indicating something is not present and a value of 1 indicate a value is present. For wooden supports, it's definitely more, but for beam support it's less than half of the track pieces that have supports. On SV4 import, the flag would just always be set, and for SV6 import it would be determined by the support type (always on for wooden supports) and position on the map ((tileX + tileY) % 2).

ZehMatt:
> There is no default value as of now, and to me it makes more sense to have a value of 0 indicating something is not present and a value of 1 indicate that something is present. For wooden supports, it's definitely more, but for beam support it's around half of the track pieces that have supports. On SV4 import, the flag would just always be set, and for SV6 import it would be determined by the support type (always on for wooden supports) and position on the map ((tileX + tileY) % 2).

There is an implicit default value of zero because all tile elements are being zeroed on import, the specific bit of Flags2 is just unused, so the default is 0. Also all the unused bits/paddings in sv4/sv6 is 0 so if you want to have s6 support you need to rely on that.

spacek531:
> There is an implicit default value of zero because all tile elements are being zeroed on import, the specific bit of Flags2 is just unused, so the default is 0. Also all the unused bits/paddings in sv4/sv6 is 0 so if you want to have s6 support you need to rely on that.

I don't expect this to serialize in SV6 or SV4 and I don't think Duncan does either. I'd rather wait for NSF to serialize the support flag and compute the flag based on a heuristic on SV6/SV4 import.

ZehMatt:
> > There is an implicit default value of zero because all tile elements are being zeroed on import, the specific bit of Flags2 is just unused, so the default is 0. Also all the unused bits/paddings in sv4/sv6 is 0 so if you want to have s6 support you need to rely on that.
> 
> I don't expect this to serialize in SV6 or SV4 and I don't think Duncan does either. I'd rather wait for NSF to serialize the support flag and compute the flag based on a heuristic on SV6/SV4 import.

What kind of heuristic would you use? Having a flag for blocking supports is the most optimal way of not breaking anything, that flag would make a feature, having it the other way is a potential park breaking thing. I don't see why this needs to be made more complicated than it has to be, if you want a function for convenience you can just negate the blocked state.

spacek531:
> What kind of heuristic would you use? Having a flag for blocking supports is the most optimal way of not breaking anything, that flag would make a feature, having it the other way is a potential park breaking thing. I don't see why this needs to be made more complicated than it has to be, if you want a function for convenience you can just negate the blocked state.

I'd add data to the TED, some behavior to the RTD, and some knowledge about what I'm importing from.

duncanspumpkin:
I really do not see why there is any need for discussion on the state of the flag. The import and export code already uses the convention of using get rct12 flag state, set openrct2 state, we would just continue that convention and then it is super super easy to switch between states. But as this flag is completely new we are going to be using some fancy code to decide the support state so again none of this matters. And anyway export doesn't matter as that will only be required for nsf. In conclusion whether it's 1 means supports or 0 means supports we need not discuss further at this time.

spacek531:
What prompted closing this? It has modifications to the tile inspector that are useful if nothing else

Issue #15195, opened 2021-08-10
in-game bug report button


Broxzier:
### Describe the feature

To make reporting bugs more accessible to the players and to help them provide the information that we need, we could introduce a way of opening the issue report page easily from within the game. That way they don't have to go looking for how to report bugs, and we're able to automatically fill the OS and build information. This can be done by constructing a URL like this:
[https:\/\/github.com/OpenRCT2/OpenRCT2/issues/new?labels=bug&template=bug_report.yaml&**operating_system=Windows+10+x64**&**openrct2_build=automatically+filled+version+info**](https://github.com/OpenRCT2/OpenRCT2/issues/new?labels=bug&template=bug_report.yaml&operating_system=Windows+10+x64&openrct2_build=automatically+filled+version+info)

The new button could be located in the about window, or as an item in the dropdown menu from the top toolbar.

### Feature type

Quality of life

Broxzier:
Creating a fully-fledged form in-game that supports all the markup and file attachments that you can use here will be way out of scope for this. Players will need a GitHub account to report things already, and if they want to see something fixed they'll probably sign up.

While we don't have any anonymous issue submission system in place, immediately opening the link to a pre-filled form is still better than letting them enter a form in-game that redirects them here when done, only for them to see they need to create an account or log in like some software like GitExtensions do, is a better design choice in my opinion.

ocalhoun6:
> Creating a fully-fledged form in-game that supports all the markup and file attachments that you can use here will be way out of scope for this.

lol, yeah. We don't even have support for the full compliment of keyboard symbols in the in-game fonts, do we? That's part of what's preventing #13471 from being done.

PR #15349, opened 2021-09-05
Add a bug-report item in file dropdown menu

Broxzier:
This implements https://github.com/OpenRCT2/OpenRCT2/issues/15195 as an item in the file dropdown menu. It automatically fills in the version information in the issue form.

bug report dropdown item

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -6,7 +6,8 @@
 - Feature: [#15164] Highlight elements selected by the Tile Inspector, tracks are currently not supported.
 - Feature: [#15165] [Plugin] Add the ability to create entities using "map.createEntity".
 - Feature: [#15194] [Plugin] Add guest properties, ride downtime and park casualty penalty.
-- Feature: [#15294] New vehicle animation type: flying animal
+- Feature: [#15195] Added a bug-report item in file dropdown menu.
+- Feature: [#15294] New vehicle animation type: flying animal.
 - Fix: [#13465] Creating a scenario based on a won save game results in a scenario that’s instantly won.
 - Fix: [#14316] Closing the Track Designs Manager window causes broken state.
 - Fix: [#14667] “Extreme Hawaiian Island” has unpurchaseable land tiles (original bug).
--- a/src/openrct2/core/String.cpp
+++ b/src/openrct2/core/String.cpp
@@ -14,7 +14,10 @@
 #endif // __MINGW32__
 
 #include <algorithm>
+#include <cctype>
 #include <cwctype>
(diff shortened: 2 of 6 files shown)


distribution/changelog.txt
```diff
@@ -6,7 +6,8 @@
 - Feature: [#15164] Highlight elements selected by the Tile Inspector, tracks are currently not supported.
 - Feature: [#15165] [Plugin] Add the ability to create entities using "map.createEntity".
 - Feature: [#15194] [Plugin] Add guest properties, ride downtime and park casualty penalty.
-- Feature: [#15294] New vehicle animation type: flying animal
+- Feature: [#15294] New vehicle animation type: flying animal.
+- Feature: [#15195] Added a bug-report item in file dropdown menu.
```
Gymnasiast, line 10: Incorrect order.

src/openrct2/core/String.cpp
```diff
@@ -823,6 +826,30 @@ namespace String
...
+            // Keep alphanumeric and other accepted characters intact
+            if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~')
+            {
+                escaped << c;
+                continue;
+            }
+
+            // Any other characters are percent-escaped
+            escaped << std::uppercase;
+            escaped << '%' << std::setw(2) << int32_t((unsigned char)c);
```
Gymnasiast, line 847: C-style cast.

src/openrct2/core/String.cpp
```diff
@@ -823,6 +826,30 @@ namespace String
...
+            if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~')
+            {
+                escaped << c;
+                continue;
+            }
+
+            // Any other characters are percent-escaped
+            escaped << std::uppercase;
+            escaped << '%' << std::setw(2) << int32_t(static_cast<unsigned char>(c));
+            escaped << std::nouppercase;
```
tupaschoal, line 848: Maybe
```suggestion
            if (std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~')
            {
                escaped << c;
            }
            else
            {
                // Any other characters are percent-escaped
                escaped << std::uppercase;
                escaped << '%' << std::setw(2) << int32_t(static_cast<unsigned char>(c));
                escaped << std::nouppercase;
            }
```
I see the benefits of using continue when the rest of the function is big, but in this case seems like we can go with `if...else` to be clearer.

tupaschoal (approved):

Gymnasiast (approved):

PR #15351, opened 2021-09-05
Fix enum serialization not using byte swap

ZehMatt:
Found this bug while changing ride_id_t to strong enum, the bytes were not ordered the same as the underlying type.

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -38,19 +38,23 @@ template<typename T> struct DataSerializerTraits_t
 
 template<typename T> struct DataSerializerTraits_enum
 {
+    using TUnderlying = std::underlying_type_t<T>;
+
     static void encode(OpenRCT2::IStream* stream, const T& val)
     {
-        stream->Write(&val);
+        TUnderlying temp = ByteSwapBE(static_cast<TUnderlying>(val));
+        stream->Write(&temp);
     }
     static void decode(OpenRCT2::IStream* stream, T& val)
     {
-        stream->Read(&val);
+        TUnderlying temp;
+        stream->Read(&temp);
+        val = static_cast<T>(ByteSwapBE(temp));
     }
     static void log(OpenRCT2::IStream* stream, const T& val)
     {
-        using underlying = std::underlying_type_t<T>;
         std::stringstream ss;
-        ss << std::hex << std::setw(sizeof(underlying) * 2) << std::setfill('0') << static_cast<underlying>(val);
+        ss << std::hex << std::setw(sizeof(TUnderlying) * 2) << std::setfill('0') << static_cast<TUnderlying>(val);
 
         std::string str = ss.str();
         stream->Write(str.c_str(), str.size());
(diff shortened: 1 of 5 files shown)


src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -38,19 +38,23 @@ template<typename T> struct DataSerializerTraits_t
...
     {
-        stream->Write(&val);
+        TUnderlying temp = ByteSwapBE(static_cast<TUnderlying>(val));
+        stream->Write(&temp);
     }
     static void decode(OpenRCT2::IStream* stream, T& val)
     {
-        stream->Read(&val);
+        TUnderlying temp;
+        stream->Read(&temp);
```
duncanspumpkin, line 51: Why is there no byte swap on import

ZehMatt: Good question

duncanspumpkin (approved):

IntelOrca:
Can we ditch the byte swapping? I can't see why it was ever done. Even if we did support big endian architectures, they'd be a minority - so not worth the bias.

ZehMatt:
I rather not remove the possibility honestly, the unfortunate side of it is rather that not more code uses the same serialization code so it wouldn't be big deal to support it.

IntelOrca:
@ZehMatt I don't understand your message

ZehMatt:
> @ZehMatt I don't understand your message

What is unclear?

IntelOrca:
@ZehMatt most of it, can you re-write it more clearly and in more depth.

ZehMatt:
> @ZehMatt most of it, can you re-write it more clearly and in more depth.

> Can we ditch the byte swapping?
No

>  I can't see why it was ever done.
Cross platform is a good reason.

> Even if we did support big endian architectures, they'd be a minority - so not worth the bias.
What kind of bias? Support would be great

IntelOrca:
If most people are on little endian devices, then you are sacrificing performance (even if small) for an architecture we don't support and if we did, would be a minority. I don't see the justification.

Why can we not remove the byte swapping?

ZehMatt:
There is no performance impact at all, it is also quite typical to use "Network Byte Order" I don't see why we should deviate from that.

IntelOrca:
What is the benefit of using network byte order?

The problem is also that this serialiser is now being used for data files like the cache which are being exported as big endian.

ZehMatt:
> What is the benefit of using network byte order?

Agreement on bit ordering.

> The problem is also that this serialiser is now being used for data files like the cache which are being exported as big endian.

Define how this is a problem when you know the data is using big endian.

IntelOrca:
Because the data files were using little endian beforehand and the change to use the serialiser inadvertently changed the format. I don't see any benefit for network or files to use big endian when we are dealing with little endian most of if not all the time. Little endian has benefits like being able to view the same data correctly at different sizes, without manipulation, which I assume is why it is more popular.

ZehMatt:
I'm not really against changing the layout to little endian as the default, I don't want to rip out the functions that would convert it to the right ones for lets say big endian systems. The ByteSwapBE function is bullshit anyway, it should be a function Host to network and Network to host so no one gets confused about which way is what. You can refactor it to be simply nop on x86 systems.

IntelOrca:
> I'm not really against changing the layout to little endian as the default, I don't want to rip out the functions that would convert it to the right ones for lets say big endian systems.

The conversion function is fine, I just would prefer little endian being the default for all our file formats and network.

ZehMatt:
Well feel free to change it

PR #15336, opened 2021-09-03
Prepare footpath handling for surfaces and railings (from NSF)

Gymnasiast:
Quite a complicated one, but I think I got it.

Again, do not squash merge.

PR #15352, opened 2021-09-05
Rework path surface and railings descriptor handling

Gymnasiast:
https://github.com/OpenRCT2/OpenRCT2/pull/15336 revealed a few problems, especially with the amount of places where it has to be aware of which path type (legacy footpath or separate path surface and railings objects) the tile element uses.

PathSurfaceEntry and PathRailingsEntry were semi-objects. The original idea was that when the new save format came, we would load any legacy footpath object as two path surface objects and one path railings object. This proved to be hard to pull off. However, the idea of passing the same structure to the drawing functions, regardless of what type of object was the origin, is still sound. 

This PR reworks the semi-objects of PathSurfaceEntry and PathRailingsEntry into path surface descriptors and path railings descriptors. These are used to do almost everything, from logic to drawing. They are generated by FootpathObject, FootpathSurfaceObject and FootpathRailingsObject. This allows us to abstract the difference between footpath object types.

### DIFF
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -1599,36 +1599,13 @@ namespace RCT1
                     dst2->SetIsBroken(false);
                     dst2->SetIsBlockedByVehicle(false);
 
-                    dst2->SetSurfaceEntryIndex(entryIndex);
+                    dst2->SetLegacyPathEntryIndex(entryIndex);
                     dst2->SetShouldDrawPathOverSupports(true);
                     if (RCT1::PathIsQueue(pathType))
                     {
                         dst2->SetIsQueue(true);
                     }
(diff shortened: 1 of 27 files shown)


src/openrct2/rct1/S4Importer.cpp
```diff
@@ -1599,36 +1599,13 @@ namespace RCT1
...
-                            case RCT1_PATH_SUPPORT_TYPE_BAMBOO:
-                                railingsEntryIndex = 5;
-                                break;
-                            case RCT1_PATH_SUPPORT_TYPE_TRUSS:
-                            default:
-                                railingsEntryIndex = 0;
-                        }
-                        dst2->SetRailingEntryIndex(railingsEntryIndex);
-                    }
+                    // TODO: Set railings type
```
Gymnasiast, line 1608: Above code never worked, it was tentatively put there for the NSF. But the NSF takes another approach here anyway.

Issue #15353, opened 2021-09-05
Add VerticalFactor and LateralFactor fields in TrackElementDescriptor to be used in Vehicle::GetGForces


frutiemax:
Right now, in Vehicle::GetGForces(), a huge switch case on the track element type assigns a vertical and lateral factor that is used in the calculation of g-forces. I suggest to add a VerticalFactor and a LateralFactor fields in the TrackElementDescriptor to eliminate that switch-case.

PR #15355, opened 2021-09-05
#15353 : Add VerticalFactor and LateralFactor fields in TrackElementDescriptor

frutiemax:
Vehicle::GetGForces(), a huge switch case on the track element type assigns a vertical and lateral factor that is used in the calculation of g-forces. This PR removes this by adding a VerticalFactor and LateralFactor fields, which are functions of the current progress on the track element.


src/openrct2/ride/TrackData.h
```diff
@@ -72,6 +72,13 @@ constexpr const dodgems_track_size DodgemsTrackSize(track_type_t type)
     return { 0, 0, 0, 0 };
 }
 
+class TrackElementFunction
```
ZehMatt, line 75: I'm against this type of polymorphism, this will introduce unwanted overhead. What you should use is a function prototype ```int32_t (*TrackEvaluatorFunction)(const uint16_t progress)``` which doesn't require vtables.

An even better way would be to have this type based such as ```template<TrackElemType TType> struct Evaluator { ... }``` which would also allow way better compiler optimizations also.

ZehMatt (changes requested):

src/openrct2/ride/TrackData.h
frutiemax: The problem is that there are multiple places where we need to return a constant. I would need to declare a constant function for each value.

frutiemax: Never mind, I think I know how to do it.

ZehMatt: > The problem is that there are multiple places where we need to return a constant. I would need to declare a constant function for each value.

You can also do this with the template system, something like ```Evaluator<Type, Constant>```

ZehMatt: Example https://godbolt.org/z/7ba5E9dbb

Issue #15346, opened 2021-09-05
Flag to draw supports


spacek531:
### Describe the feature

@duncanspumpkin has a working implementation to hide support columns using a TrackElement flag.

Liampie suggested using the flag to control if the supports are drawn (different than hiding) to improve SV4 import.

The function `track_paint_util_should_paint_supports` in `TrackPaint.cpp` has to be run for every TrackElement to be rendered. It's a small optimization if this function is only run when the TrackElement is built to determine the supports flag state and the support is drawn or not based on the flag.

Some challenges for this:
- Have to find a place to serialize the flag in SV6 or set the flag on import and do not export the flag
- RTDs have to be updated

### Feature type

Quality of life

duncanspumpkin:
Annoyingly not every track piece goes through track_paint_util_should_paint_supports so we would likely need a very complex import function to correctly reassign supports from an sv6.

spacek531:
You're talking about curves, loops, corkscrews, helices, and the likes?

The simplest approach is to set all of their sequence with the flag, and let the paint code figure it out.

The more complicated approach is to add supports to all the sequences in the paint code, then add a field in the TED for all the track sequences to determine if the supports are default present, default vacant, or default `track_paint_uitl_should_paint_supports`, then add a field to the RTD to determine if the track pieces should use the TED value or set all sequence supports true.

spacek531:
I did some testing today on RCT2 supports to see if every track piece is drawn the same way. There's a few discrepancies:

1. three loop support classes. Most track types have loop supports on the center 4 tiles. Inverted track and LIM-launched have only 2 supports on the lowest tiles. Looping Coaster has supports only on the highest tiles.

![image](https://user-images.githubusercontent.com/3454156/132139372-5a23ab8a-ed20-4746-be4a-9565007a6617.png)

2. Some ride types have a support every tile on straight track (monorail cycles not pictured).
3. The two rides with covered track elements have different supports.

![image](https://user-images.githubusercontent.com/3454156/132139430-7946d2f9-dd8e-43af-86ce-9706523fe3bf.png)

4. Inline twists are different on inverted coaster than LIM-Launched.

![image](https://user-images.githubusercontent.com/3454156/132139559-545ca1df-dfee-4c8f-9f33-81130eb958a6.png)

5. Differences between RCT2's vertical drop coaster supports and OpenRCT2's junior coaster supports

[...]

PR #15356, opened 2021-09-06
Supports field in rtd

spacek531:
Regarding https://github.com/OpenRCT2/OpenRCT2/issues/15346 and inspired by https://github.com/OpenRCT2/OpenRCT2/pull/15345

Bit 7 in the Flags2 field decides definitively if supports are drawn.

This feature is not intended to serialize in SV6. It's just not. It won't go into SV6 because there's no versioning SV6 to know if the park was saved before or after this feature was implemented. This feature is intended for NSF.

Gymnasiast says it's fine if RCT1 supports don't serialize.

Known discrepancies with RCT2 behaviour:
- Monster Trucks supports always draw on Up25ToUp60, Up60, Up60ToUp25, Down25ToDown60, Down60, Down60ToDown25 instead of ShouldHave
- Submarine ride supports always draw on LeftQuarterTurn1Tile, RightQuarterTurn1Tile, StationStart, StationEnd, StationMiddle instead of never
- Inverted Hairpin coaster supports ShouldHave draw on Up25ToUp60, Down60ToDown25 instead of never
- Inverted Hairpin coaster supports ShouldHave draw on Up60ToUp25, Down25ToDown60 instead of always

### DIFF
--- a/src/openrct2/actions/RideSetSettingAction.cpp
+++ b/src/openrct2/actions/RideSetSettingAction.cpp
@@ -223,6 +223,7 @@ GameActions::Result::Ptr RideSetSettingAction::Execute() const
             break;
         case RideSetSetting::RideType:
             ride->type = _value;
+            UpdateSupports(false);
             gfx_invalidate_screen();
             break;
     }
@@ -296,4 +297,31 @@ rct_string_id RideSetSettingAction::GetOperationErrorMessage(Ride* ride) const
                 return STR_CANT_CHANGE_LAUNCH_SPEED;
             }
     }
+}
+
+
+void RideSetSettingAction::UpdateSupports(bool RCT1SupportsOverride) const
+{
+
+    uint8_t oldpaused = gGamePaused;
+    gGamePaused = 0;
+
+    tile_element_iterator it;
+
+    tile_element_iterator_begin(&it);
+    while (tile_element_iterator_next(&it))
+    {
+        if (it.element->GetType() != TILE_ELEMENT_TYPE_TRACK)
+            continue;
+
+        if (it.element->AsTrack()->GetRideIndex() != static_cast<ride_id_t>(_rideIndex))
+            continue;
+
+        it.element->AsTrack()->SetHasSupport(
+            it.element->AsTrack()->DetermineSupportState({ it.x, it.y }, RCT1SupportsOverride));
+
+        tile_element_iterator_restart_for_tile(&it);
+    }
+
+    gGamePaused = oldpaused;
 }
--- a/src/openrct2/actions/RideSetSettingAction.cpp
+++ b/src/openrct2/actions/RideSetSettingAction.cpp
@@ -299,10 +299,8 @@ rct_string_id RideSetSettingAction::GetOperationErrorMessage(Ride* ride) const
     }
 }
 
-
 void RideSetSettingAction::UpdateSupports(bool RCT1SupportsOverride) const
 {
-
     uint8_t oldpaused = gGamePaused;
     gGamePaused = 0;
 
--- a/src/openrct2/ride/RideConstruction.cpp
+++ b/src/openrct2/ride/RideConstruction.cpp
@@ -454,6 +454,18 @@ std::optional<CoordsXYZ> GetTrackElementOriginAndApplyChanges(
         {
             trackElement->SetHasCableLift(false);
         }
+        if (flags & TRACK_ELEMENT_SET_HAS_SUPPORT_TRUE)
+        {
+            trackElement->SetHasSupport(true);
+        }
+        if (flags & TRACK_ELEMENT_SET_HAS_SUPPORT_FALSE)
+        {
+            trackElement->SetHasSupport(false);
+        }
+        if (flags & TRACK_ELEMENT_SET_HAS_SUPPORT_DEFAULT)
+        {
+            trackElement->SetHasSupport(trackElement->DetermineSupportState(cur,  false));
+        }
     }
     return retCoordsXYZ;
 }
--- a/src/openrct2/ride/RideConstruction.cpp
+++ b/src/openrct2/ride/RideConstruction.cpp
@@ -464,7 +464,7 @@ std::optional<CoordsXYZ> GetTrackElementOriginAndApplyChanges(
         }
         if (flags & TRACK_ELEMENT_SET_HAS_SUPPORT_DEFAULT)
         {
-            trackElement->SetHasSupport(trackElement->DetermineSupportState(cur,  false));
+            trackElement->SetHasSupport(trackElement->DetermineSupportState(cur, false));
         }
     }
     return retCoordsXYZ;
--- a/src/openrct2/ride/RideData.h
+++ b/src/openrct2/ride/RideData.h
@@ -175,6 +175,7 @@ struct RideTypeDescriptor
     track_colour_preset_list ColourPresets;
     RideColourPreview ColourPreview;
     RideColourKey ColourKey;
+    uint8_t SupportsBehaviour;
 
     bool HasFlag(uint64_t flag) const;
     uint64_t GetAvailableTrackPieces() const;
@@ -331,6 +332,22 @@ extern const uint16_t RideFilmLength[3];
 
 extern const rct_string_id RideModeNames[static_cast<uint8_t>(RideMode::Count)];
 
+namespace RideSupportsBehaviour
+{
+    constexpr uint8_t AlwaysHasSupports = 1 << 0;
+    constexpr uint8_t ChairliftSupports = 1 << 1;
+    constexpr uint8_t ShouldHaveSupports = 1 << 2;
+    constexpr uint8_t RCT1ExtraSupports = 1 << 3;
+    constexpr uint8_t CoveredExtraSupports = 1 << 4;
+    constexpr uint8_t InversionExtraSupports = 1 << 5;
+    // LoopASupports is implied
+    constexpr uint8_t LoopBSupports = 1 << 6;
+    constexpr uint8_t LoopCSupports = 1 << 7;
+
+}; // namespace RideSupportsBehaviour
+
+const uint8_t _defaultCoasterSupports = RideSupportsBehaviour::ShouldHaveSupports;
+
 // clang-format off
 constexpr const RideTypeDescriptor DummyRTD =
 {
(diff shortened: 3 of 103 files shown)


src/openrct2/ride/RideConstruction.cpp
```diff
@@ -454,6 +454,18 @@ std::optional<CoordsXYZ> GetTrackElementOriginAndApplyChanges(
         {
             trackElement->SetHasCableLift(false);
         }
+        if (flags & TRACK_ELEMENT_SET_HAS_SUPPORT_TRUE)
```
spacek531, line 457: I don't know if this would be used anywhere.

src/openrct2/ride/Track.cpp
```diff
@@ -899,3 +899,104 @@ void TrackElement::SetHighlight(bool on)
...
+        return true;
+    }
+    else
+    {
+        Guard::Assert(sequence < MaxSequencesPerPiece);
+        auto ted = GetTrackElementDescriptor(trackType);
+        // if it is only a chairlift, only give chairlift supports
+        if (rtd.SupportsBehaviour == RideSupportsBehaviour::ChairliftSupports)
+        {
+            return TrackSupportsHasFlag(TRACK_SUPPORT_CHAIRLIFT_SUPPORT, sequence);
```
spacek531, line 938: Special chairlift behavior in the off chance that one day users will have the option to create a chairlift support on flat, up25, and down 25 tracks.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -1563,7 +1563,7 @@ static std::optional<money32> track_design_place_ride(TrackDesign* td6, const Co
                 uint32_t brakeSpeed = (track.flags & 0x0F) * 2;
                 uint32_t seatRotation = track.flags & 0x0F;
 
-                int32_t liftHillAndAlternativeState = 0;
+                int32_t liftHillAndAlternativeState = CONSTRUCTION_BUILD_SUPPORTS;
```
spacek531, line 1566: This should be loaded from a cheat. Maybe another PR.

src/openrct2/ride/Track.cpp
```diff
@@ -899,3 +899,104 @@ void TrackElement::SetHighlight(bool on)
...
+        {
+            if (rtd.SupportsBehaviour & RideSupportsBehaviour::ShouldHaveSupports)
+                shouldDraw |= ((position.x & (1 << 5)) == (position.y & (1 << 5)))
+                    || ((!(position.x & (1 << 5))) && (!(position.y & (1 << 5))));
+            else
+                shouldDraw = true;
+        }
+
+        if (rtd.SupportsBehaviour & RideSupportsBehaviour::ChairliftSupports)
+            shouldDraw |= TrackSupportsHasFlag(TRACK_SUPPORT_CHAIRLIFT_SUPPORT, ted.SequenceSupportBehaviour[sequence]);
```
spacek531, line 951: Inverted Impulse Coaster always has a support on FlatToUp25, Up25ToFlat, FlatToDown25, Down25ToFlat

src/openrct2/ride/Track.cpp
```diff
@@ -899,3 +899,104 @@ void TrackElement::SetHighlight(bool on)
...
+
+        if (rtd.SupportsBehaviour & RideSupportsBehaviour::RCT1ExtraSupports)
+            shouldDraw |= TrackSupportsHasFlag(TRACK_SUPPORT_RCT1_EXTRA_SUPPORT, ted.SequenceSupportBehaviour[sequence]);
+
+        if (rtd.SupportsBehaviour & RideSupportsBehaviour::CoveredExtraSupports)
+            shouldDraw |= TrackSupportsHasFlag(TRACK_SUPPORT_COVERED_EXTRA_SUPPORT, ted.SequenceSupportBehaviour[sequence]);
```
spacek531, line 957: - B&M type always has a support on FlatToUp60, Up60ToFlat, FlatToDown60, Down60ToFlat
- wooden supports for coaster boats alt style
- FlatTrack1x4A, FlatTrack1x4B, FlatTrack1x4C have this set on the ends for future compatibility
- FlatTrack2x4 has this set on the sides for future compatibility
- FlatTrack1x5 has this set on the ends for future compatibility

src/openrct2/ride/Track.cpp
```diff
@@ -899,3 +899,104 @@ void TrackElement::SetHighlight(bool on)
...
+
+        if (rtd.SupportsBehaviour & RideSupportsBehaviour::InversionExtraSupports)
+            shouldDraw |= TrackSupportsHasFlag(TRACK_SUPPORT_INVERSION_EXTRA_SUPPORT, ted.SequenceSupportBehaviour[sequence]);
```
spacek531, line 960: Inverting coasters tend to have 2 supports on inline twists/barrel rolls, while non-inverted coasters tend to have only 1 support

spacek531:
This feature won't serialize in SV6.

src/openrct2/ride/TrackData.cpp
```diff
@@ -5427,6 +5427,278 @@ constexpr static uint8_t TrackTypeToSpinFunction[TrackElemType::Count] = {
     NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN
 };
 
+// clang-format off
+constexpr uint8_t TrackSupportsBehaviour[TrackElemType::Count][MaxSequencesPerPiece] = {
```
duncanspumpkin, line 5431: How did you generate this table?

duncanspumpkin:
Perhaps hold your horses for a bit let's have a little think about how this will work so repeated work doesn't need to be thrown out. All these additional changes to the draw code can i think be streamlined/simplified by moving the support checking into the support code would need to pass the track element though to the support code.

src/openrct2/ride/TrackData.cpp
spacek531: By hand

spacek531:
> Perhaps hold your horses for a bit let's have a little think about how this will work so repeated work doesn't need to be thrown out. All these additional changes to the draw code can i think be streamlined/simplified by moving the support checking into the support code would need to pass the track element though to the support code.

Good idea with the paint session.

src/openrct2/ride/RideData.h
```diff
@@ -331,6 +332,22 @@ extern const uint16_t RideFilmLength[3];
 
 extern const rct_string_id RideModeNames[static_cast<uint8_t>(RideMode::Count)];
 
+namespace RideSupportsBehaviour
+{
+    constexpr uint8_t AlwaysHasSupports = 1 << 0;
+    constexpr uint8_t ChairliftSupports = 1 << 1;
+    constexpr uint8_t ShouldHaveSupports = 1 << 2;
+    constexpr uint8_t RCT1ExtraSupports = 1 << 3;
```
spacek531, line 340: This may be unnecessary. Investigation into RCT1 supports pending.

spacek531:
I don't know how to solve the -WNoDeclarations warnings. Obviously I could declare the functions in Track.h, but at least one doesn't need to be there. I know it's possible because I've seen it in other files.

src/openrct2/ride/Track.cpp
```diff
@@ -899,3 +899,114 @@ void TrackElement::SetHighlight(bool on)
     if (on)
         Flags2 |= TRACK_ELEMENT_FLAGS2_HIGHLIGHT;
 }
+
+bool TrackSupportsHasFlag(uint8_t queryFlag, uint8_t sequenceFlags)
```
duncanspumpkin, line 903: if a function is purely only used in the singular .cpp file and has no declaration in a header then it is `static`. So just add `static` before the bool. Same goes to the other function in this file that the compiler is complaining about.

duncanspumpkin:
A couple of thoughts i was having about this over the weekend:

1. Using cheats to switch track types. This does not at present change supports state as it is defined from when you place the track piece. Is this an issue. In my opinion no. You can only switch track types with cheats and people using cheats will not mind / they tend to not want supports. Perhaps an additional cheat could be made to reset all supports to default.
2. This is going to be a really fundamental change for supports. Chance of bugs is high. Can we verify that there has been no regressions during this. Test paint could have verified this but the way supports are done it would never have been great at confirming it. Perhaps a new test paint like test could be written. This would not be easy.
3. Are there future support things we need to consider. I can't really recall much in terms of feature requests related to supports.
4. Paths and scenery also have supports should they have a flag. I think perhaps they should but should be left for a future pr. The interaction with the path supports changes for nsf may make this not an issue for paths.
5. Paths have a support object for the nsf. It is presumed that path supports and track supports are so vastly different that there is no overlap or similarities.

spacek531:
> A couple of thoughts i was having about this over the weekend:
> 
> 1. Using cheats to switch track types. This does not at present change supports state as it is defined from when you place the track piece. Is this an issue. In my opinion no. You can only switch track types with cheats and people using cheats will not mind / they tend to not want supports. Perhaps an additional cheat could be made to reset all supports to default.
> 2. This is going to be a really fundamental change for supports. Chance of bugs is high. Can we verify that there has been no regressions during this. Test paint could have verified this but the way supports are done it would never have been great at confirming it. Perhaps a new test paint like test could be written. This would not be easy.
> 3. Are there future support things we need to consider. I can't really recall much in terms of feature requests related to supports.
> 4. Paths and scenery also have supports should they have a flag. I think perhaps they should but should be left for a future pr. The interaction with the path supports changes for nsf may make this not an issue for paths.
> 5. Paths have a support object for the nsf. It is presumed that path supports and track supports are so vastly different that there is no overlap or similarities.

1. I didn't think about changing track types. The simplest option right now is to force the supports to the default state when changing track types. When NSF goes live a tickbox can be added next to the drop-down to allow users to choose the behavior.
image
2. I did think about this but I don't know how to write a test. Right now there are some changes (see the first post of this PR) but in my mind those two specific examples are probably fine. Any future changes might not be though.
3. There was talk about being able to change track paint function per-TrackElement in NSF. I don't know how that would affect this PR. I did add some future-proofing for flat ride bases and the RCT1ExtraSupport flag/behaviour for when the RCT1 tests come in.
4. I agree that paths should have a HasSupports flag and that can be another PR.
5. If rides get their paint code broken out to JSON or YML then I think this will be more relevant.

src/openrct2/actions/RideSetSettingAction.cpp
```diff
@@ -297,3 +298,28 @@ rct_string_id RideSetSettingAction::GetOperationErrorMessage(Ride* ride) const
             }
     }
 }
+
+void RideSetSettingAction::UpdateSupports(bool RCT1SupportsOverride) const
+{
+    uint8_t oldpaused = gGamePaused;
```
spacek531, line 304: I just copied this from RideDeleteAction. Is pausing necessary?

spacek531:
[TrackZoo.zip](https://github.com/OpenRCT2/OpenRCT2/files/8101829/TrackZoo.zip)

I have created a flat world with one ride that has every track element (2x of every element to test the every-other support). I imagine it could be used to test this code, although I do not know how to do that yet.

rik-smeets:
I'll close this PR for now due to inactivity, feel free to reopen if you continue working on it.

PR #15361, opened 2021-09-06
Do not turn off chain when building downhill if chain cheat is on

spacek531:


duncanspumpkin:
Could you provide some info as to why you are making this change. From first look at this your just expanding the code duplicating common items.

spacek531:
I don't like turning on the chain lift when it turns itself off automatically when building forward

spacek531:
does anyone know why going backwards with the cheat enabled explicitly does not pick up the chain state?

https://github.com/OpenRCT2/OpenRCT2/blob/6c3a417af35572ad4fcbf77ffd43bb99192f3824/src/openrct2/ride/RideConstruction.cpp#L732-L735

tupaschoal (commented):
@spacek531 it seems like this PR doesn't really match what it was opened as (at least in terms of title).

Can you update the title? Is there any open issue you're trying to fix?

Also is the reason for it still being in draft the question you left there? (I don't know the answer btw)

spacek531:
It was pointed out I have a lot of open PR about 2 months ago, so to try and hurry things along I marked the ones I wanted to merge first as ready and the other ones as draft.

tupaschoal:
Ah, I see, thanks for explaining.

Yeah, we plan on getting there.

Gymnasiast (approved):
~~Code looks good, please rebase.~~

PR #15363, opened 2021-09-06
Add support for surface and railings objects to PathElement

Gymnasiast:
Half backport of IntelOrca’s changes in the NSF branch, half reworked.


ZehMatt (approved):
Code looks alright, did not test it.

Issue #7317, opened 2018-03-19
Removing a track piece on a ride with a rotated entrance removes this entrance


Gymnasiast:
**OS:** macOS Sierra
**Version:** 0.1.3
**Commit/Build:** 2d1d60294c1ded1c510bffbd555cad9feae26198

- [ ] Reproducible in RCT2 (vanilla)?
- [ ] Specific to multiplayer?

**Steps to reproduce:**
1. Remove a track piece from a ride with a rotated entrance.
2. Entrance gets removed.

![dynamite dunes 2018-03-19 14-36-22](https://user-images.githubusercontent.com/1478678/37598771-65f1b2d0-2b83-11e8-946d-94cac2b9de31.png)
![dynamite dunes 2018-03-19 14-36-26](https://user-images.githubusercontent.com/1478678/37598784-6eb26f40-2b83-11e8-9858-3c0ab79cc40d.png)

PR #15364, opened 2021-09-06
Cheat for disconnected entrances

spacek531:
The unintended consequence is that ride entrances can live longer than their associated station platforms. Because ValidateStations is called a bunch of times per frame, and not all of them have non-null stationIndex Start locations, it's not possible to use them to determine if an entrance should be disconnected. This is fine though, because the cheat warns about instability when you activate it.


spacek531:
As a builder, when I have disconnected entrances, even as simple as a rotated entrance, entering build mode removes the entrances, which require replacement. If the entrances were not removed when the ride is edited, a lot of time could be saved.

https://user-images.githubusercontent.com/3454156/154032443-4bf8104c-dfff-4497-8e95-3e3d8fab8363.mp4

duncanspumpkin:
This is a bad idea it's encouraging putting the game into an invalid state.

spacek531:
What kind of invalid state?

Broxzier:
Related issue: https://github.com/OpenRCT2/OpenRCT2/issues/7317

I don't think this should be a cheat. Entrances that have been rotated should stay by default. It only makes sense to remove entrances when you remove the station piece that it's connected to, or when an entire station is removed.

Gymnasiast:
Closing due to negative feedback from other team members, and this lying dormant for half a year.

I think we would accept a PR that just fixes the game autoremoving rotated entrances/exits.

PR #15366, opened 2021-09-07
Fix hash bucketing of paint entries

ZehMatt:
Using clamp makes no sense for hash buckets.
develop:
```
>openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.075071s, 13 FPS
Zoom[1] average: 0.032724s, 31 FPS
Zoom[2] average: 0.015505s, 64 FPS
Total average: 0.041100s, 24 FPS
Time: 2.46601s
```

PR:
```
>openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.073520s, 14 FPS
Zoom[1] average: 0.032551s, 31 FPS
Zoom[2] average: 0.014496s, 69 FPS
Total average: 0.040189s, 25 FPS
Time: 2.41136s
```
Some free FPS, this reflects better in-game, in bpb for the max zoom I was on 94 (develop), now its 108~.

### DIFF
--- a/src/openrct2/paint/Paint.h
+++ b/src/openrct2/paint/Paint.h
@@ -113,7 +113,10 @@ struct tunnel_entry
     uint8_t type;
 };
 
-#define MAX_PAINT_QUADRANTS 512
+// NOTE: This should be preferably a prime number. This is the amount of
+// buckets used with the position hash, so the bucket is hash % max.
+#define MAX_PAINT_QUADRANTS 521
+
 #define TUNNEL_MAX_COUNT 65
 
 /**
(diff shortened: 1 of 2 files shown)


src/openrct2/paint/Paint.h
```diff
@@ -113,7 +113,10 @@ struct tunnel_entry
     uint8_t type;
 };
 
-#define MAX_PAINT_QUADRANTS 512
+// NOTE: This should be preferably a prime number. This is the amount of
```
duncanspumpkin, line 116: prime or power of 2? cause it will be much slower if it needs to modulo everything i would presume.

ZehMatt: Well the benchmark proves otherwise, hash maps typically use primes and not power of two for a reason, better distribution out weights the modulo performance.

duncanspumpkin: Yes but did you compare without the clamp and modulo 512?

ZehMatt: Yes its slightly better without changing the constant, but modifying the constant squeezes another 5-6 FPS on bpb

ZehMatt: It would be even best if every zoom level had a different constant to get the best out of it. But still better

duncanspumpkin: Ahh yeah im forgetting the bigger picture here in that this is just for assigning the bucket and with a prime we get better bucket distribution which ultimately would reduce other comparisons.

duncanspumpkin (approved):

PR #15368, opened 2021-09-07
Use prebuilt table for object entry indices

ZehMatt:
Minor performance improvement. This is used for every GetLoadedObject call

### DIFF
--- a/src/openrct2/object/ObjectManager.cpp
+++ b/src/openrct2/object/ObjectManager.cpp
@@ -32,6 +32,50 @@
 #include <thread>
 #include <unordered_set>
 
+enum class ObjectTypeEntryIndex : int32_t
+{
+    Rides = 0,
+    SmallScenery = Rides + MAX_RIDE_OBJECTS,
+    LargeScenery = SmallScenery + MAX_SMALL_SCENERY_OBJECTS,
+    WallScenery = LargeScenery + MAX_LARGE_SCENERY_OBJECTS,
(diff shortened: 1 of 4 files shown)


src/openrct2/object/ObjectManager.cpp
```diff
@@ -32,6 +32,50 @@
...
+    PathAdditions = Paths + MAX_PATH_OBJECTS,
+    SceneryGroups = PathAdditions + MAX_PATH_ADDITION_OBJECTS,
+    ParkEntrances = SceneryGroups + MAX_SCENERY_GROUP_OBJECTS,
+    Water = ParkEntrances + MAX_PARK_ENTRANCE_OBJECTS,
+    ScenarioTexts = Water + MAX_WATER_OBJECTS,
+    TerrainSurfaces = ScenarioTexts + MAX_SCENARIO_TEXT_OBJECTS,
+    TerrainEdges = TerrainSurfaces + MAX_TERRAIN_SURFACE_OBJECTS,
+    Stations = TerrainEdges + MAX_TERRAIN_EDGE_OBJECTS,
+    Music = Stations + MAX_STATION_OBJECTS,
+    FootpathSurfaces = Music + MAX_FOOTPATH_SURFACE_OBJECTS,
```
tupaschoal, line 52:
```suggestion
    FootpathSurfaces = Music + MAX_MUSIC_OBJECTS,
```
This will amount in the same number for now, but if the defines change in the future...

ZehMatt:
I don't think its worth having this temporarily, I'm almost done refactoring the way this works.

PR #15369, opened 2021-09-07
Refactor overload map_get_first_element_at

ZehMatt:
Going from TileCoordsXY to CoordsXY and then from CoordsXY back to TileCoordsXY doesn't seem very practical so I added an overload that accepts TileCoordsXY which is the right coordinate system to access the tiles. There is a lot of code that can be improved as there is a lot of forth and back since more functions have this the wrong way.

### DIFF
--- a/src/openrct2-ui/windows/Banner.cpp
+++ b/src/openrct2-ui/windows/Banner.cpp
@@ -96,7 +96,7 @@ private:
             return nullptr;
         }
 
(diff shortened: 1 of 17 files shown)


duncanspumpkin:
This was something that used to exist but was subsequently removed. I do like the idea of just having the two coordinates screen and coordsxy but it is a performance penalty.

Gymnasiast (approved):

Gymnasiast:
Mind the milestone!

PR #15370, opened 2021-09-07
Rename unknown variables in vehicle motion code

spacek531:
Another housekeeping thing. Fingers crossed that replays pass.

The todos will create monster diffs because of the indentation changes.

### DIFF
--- a/src/openrct2/ride/CableLift.cpp
+++ b/src/openrct2/ride/CableLift.cpp
@@ -266,21 +266,21 @@ bool Vehicle::CableLiftUpdateTrackMotionForwards()
 
         track_progress = trackProgress;
         const auto moveInfo = GetMoveInfo();
-        auto unk = CoordsXYZ{ moveInfo->x, moveInfo->y, moveInfo->z } + TrackLocation;
+        auto nextVehiclePosition = CoordsXYZ{ moveInfo->x, moveInfo->y, moveInfo->z } + TrackLocation;
 
-        uint8_t bx = 0;
-        unk.z += GetRideTypeDescriptor(curRide->type).Heights.VehicleZOffset;
-        if (unk.x != unk_F64E20.x)
-            bx |= (1 << 0);
-        if (unk.y != unk_F64E20.y)
-            bx |= (1 << 1);
-        if (unk.z != unk_F64E20.z)
-            bx |= (1 << 2);
+        uint8_t remainingDistanceFlags = 0;
+        nextVehiclePosition.z += GetRideTypeDescriptor(curRide->type).Heights.VehicleZOffset;
+        if (nextVehiclePosition.x != _vehicleCurPosition.x)
+            remainingDistanceFlags |= (1 << 0);
+        if (nextVehiclePosition.y != _vehicleCurPosition.y)
+            remainingDistanceFlags |= (1 << 1);
+        if (nextVehiclePosition.z != _vehicleCurPosition.z)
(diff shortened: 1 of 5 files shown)


spacek531:
For some reason this seems to get hung up on the tests. I'm not sure why.

src/openrct2/ride/Vehicle.h
```diff
@@ -583,6 +583,14 @@ enum
     SOUND_RANGE_NONE = 255
 };
 
+enum
+{
+    VEHICLE_MOVEMENT_TRANSLATE_X = 1 << 0,
```
duncanspumpkin, line 588: Follow the coding style for flags:
```cpp
namespace VehicleMovementTranslate
{
    constexpr uint8_t X = (1 << 0);
    ...
}
```

duncanspumpkin (commented):
There is a lot of common code between these. Do you think there is much scope to extract this out into a function.

spacek531:
I can probably move some of the common code into functions

src/openrct2/ride/CableLift.cpp
```diff
@@ -230,7 +230,7 @@ bool Vehicle::CableLiftUpdateTrackMotionForwards()
     auto curRide = GetRide();
     if (curRide == nullptr)
         return false;
-
+    // why 13962? This is the SubpositionMovementDistances XYZ translation value plus 1
```
duncanspumpkin, line 233: Suspect its the smallest distance that actually can move the vehicle in pixels.

spacek531:
I marked this PR as ready so that the diff doesn't get too big. There's more I want to do, but in another PR because it will generate huge diffs.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8206,6 +8241,7 @@ bool Vehicle::UpdateTrackMotionForwardsGetNewTrack(uint16_t trackType, Ride* cur
 bool Vehicle::UpdateTrackMotionForwards(rct_ride_entry_vehicle* vehicleEntry, Ride* curRide, rct_ride_entry* rideEntry)
 {
     uint16_t otherVehicleIndex = SPRITE_INDEX_NULL;
+    // TODO: replace this with a while loop or a for (;;_vehicleUnkF64E10++) loop
```
spacek531, line 8244: Will change indentation which will create a huge diff

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8318,45 +8352,32 @@ bool Vehicle::UpdateTrackMotionForwards(rct_ride_entry_vehicle* vehicleEntry, Ri
     // loc_6DB706
     const auto moveInfo = GetMoveInfo();
     trackType = GetTrackType();
+    // TODO: remove this redundant variable and remove the reduntant brackets
```
spacek531, line 8355: Will change indentation which will create a huge diff

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8647,29 +8669,14 @@ bool Vehicle::UpdateTrackMotionBackwards(rct_ride_entry_vehicle* vehicleEntry, R
 
         // loc_6DBD42
         track_progress = newTrackProgress;
+        // TODO: remove this redundant variable and remove the reduntant brackets
```
spacek531, line 8672: Will change indentation which will create a huge diff

duncanspumpkin (approved):

PR #15378, opened 2021-09-08
Refactor object storage in s6

ZehMatt:
The first step towards simplifying the overall object manager stuff. Separating the single list into multiple makes it a lot easier to refactor the object manager to also have individual object lists, all object access is usually type + local index so this would only make sense to also store it like that.

### DIFF
--- a/src/openrct2/rct2/RCT2.h
+++ b/src/openrct2/rct2/RCT2.h
@@ -838,20 +838,20 @@ struct rct_s6_data
     // SC6[3]
     union
     {
-        rct_object_entry objects[RCT2_OBJECT_ENTRY_COUNT];
+        rct_object_entry Objects[RCT2_OBJECT_ENTRY_COUNT];
         struct
         {
-            rct_object_entry rideObjects[RCT12_MAX_RIDE_OBJECTS];
-            rct_object_entry sceneryObjects[RCT2_MAX_SMALL_SCENERY_OBJECTS];
-            rct_object_entry largeSceneryObjects[RCT2_MAX_LARGE_SCENERY_OBJECTS];
-            rct_object_entry wallSceneryObjects[RCT2_MAX_WALL_SCENERY_OBJECTS];
-            rct_object_entry bannerObjects[RCT2_MAX_BANNER_OBJECTS];
-            rct_object_entry pathObjects[RCT2_MAX_PATH_OBJECTS];
-            rct_object_entry pathAdditionObjects[RCT2_MAX_PATH_ADDITION_OBJECTS];
-            rct_object_entry sceneryGroupObjects[RCT2_MAX_SCENERY_GROUP_OBJECTS];
-            rct_object_entry parkEntranceObjects[RCT2_MAX_PARK_ENTRANCE_OBJECTS];
-            rct_object_entry waterObjects[RCT2_MAX_WATER_OBJECTS];
-            rct_object_entry scenarioTextObjects[RCT2_MAX_SCENARIO_TEXT_OBJECTS];
+            rct_object_entry RideObjects[RCT12_MAX_RIDE_OBJECTS];
+            rct_object_entry SceneryObjects[RCT2_MAX_SMALL_SCENERY_OBJECTS];
+            rct_object_entry LargeSceneryObjects[RCT2_MAX_LARGE_SCENERY_OBJECTS];
+            rct_object_entry WallSceneryObjects[RCT2_MAX_WALL_SCENERY_OBJECTS];
+            rct_object_entry BannerObjects[RCT2_MAX_BANNER_OBJECTS];
+            rct_object_entry PathObjects[RCT2_MAX_PATH_OBJECTS];
+            rct_object_entry PathAdditionObjects[RCT2_MAX_PATH_ADDITION_OBJECTS];
+            rct_object_entry SceneryGroupObjects[RCT2_MAX_SCENERY_GROUP_OBJECTS];
+            rct_object_entry ParkEntranceObjects[RCT2_MAX_PARK_ENTRANCE_OBJECTS];
+            rct_object_entry WaterObjects[RCT2_MAX_WATER_OBJECTS];
+            rct_object_entry ScenarioTextObjects[RCT2_MAX_SCENARIO_TEXT_OBJECTS];
         };
     };
 
--- a/src/openrct2/rct2/RCT2.h
+++ b/src/openrct2/rct2/RCT2.h
@@ -836,7 +836,24 @@ struct rct_s6_data
     // packed objects
 
     // SC6[3]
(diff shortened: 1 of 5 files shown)


src/openrct2/rct2/RCT2.h
```diff
@@ -836,7 +836,24 @@ struct rct_s6_data
     // packed objects
 
     // SC6[3]
-    rct_object_entry objects[RCT2_OBJECT_ENTRY_COUNT];
+    union
+    {
+        rct_object_entry objects[RCT2_OBJECT_ENTRY_COUNT];
+        struct
+        {
+            rct_object_entry rideObjects[RCT12_MAX_RIDE_OBJECTS];
```
Gymnasiast, line 844: This should use TitleCase.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -278,20 +296,18 @@ void S6Exporter::Export()
...
+    ExportObjectList(objectMgr, _s6.SceneryObjects, ObjectType::SmallScenery, RCT2_MAX_SMALL_SCENERY_OBJECTS);
+    ExportObjectList(objectMgr, _s6.LargeSceneryObjects, ObjectType::LargeScenery, RCT2_MAX_LARGE_SCENERY_OBJECTS);
+    ExportObjectList(objectMgr, _s6.WallSceneryObjects, ObjectType::Walls, RCT2_MAX_WALL_SCENERY_OBJECTS);
+    ExportObjectList(objectMgr, _s6.BannerObjects, ObjectType::Banners, RCT2_MAX_BANNER_OBJECTS);
+    ExportObjectList(objectMgr, _s6.PathObjects, ObjectType::Paths, RCT2_MAX_PATH_OBJECTS);
+    ExportObjectList(objectMgr, _s6.PathAdditionObjects, ObjectType::PathBits, RCT2_MAX_PATH_ADDITION_OBJECTS);
+    ExportObjectList(objectMgr, _s6.SceneryGroupObjects, ObjectType::SceneryGroup, RCT2_MAX_SCENERY_GROUP_OBJECTS);
+    ExportObjectList(objectMgr, _s6.ParkEntranceObjects, ObjectType::ParkEntrance, RCT2_MAX_PARK_ENTRANCE_OBJECTS);
+    ExportObjectList(objectMgr, _s6.WaterObjects, ObjectType::Water, RCT2_MAX_WATER_OBJECTS);
+    ExportObjectList(objectMgr, _s6.ScenarioTextObjects, ObjectType::ScenarioText, RCT2_MAX_SCENARIO_TEXT_OBJECTS);
```
AaronVanGeffen, line 310: Do parks exported in this manner still load correctly in vanilla? I imagine vanilla treats it as once list, and might stop after the first empty entry it encounters.

ZehMatt: I don't think this is the game as all objects are mapped towards {type, index}

Gymnasiast: Vanilla simply expects a list of 721 items (in total), with any empty slots marked by setting the object header to -1.

ZehMatt: Well it uses an array that has indices of the entry by a specific object type which we don't really need if we just separate them simplifying a lot of code, I have it almost done in a local branch of mine.

ZehMatt: To be clear, its still exported/imported as expected, I can't break that for obvious reasons. So gonna mark this resolved.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -255,6 +255,24 @@ static void scenario_fix_ghosts(rct_s6_data* s6)
     }
 }
 
+template<typename T> static void ExportObjectList(IObjectManager& objMgr, T& objects, ObjectType type, size_t maxEntries)
```
duncanspumpkin, line 258: you can const type and maxEntries

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -255,6 +255,24 @@ static void scenario_fix_ghosts(rct_s6_data* s6)
...
+{
+    for (size_t i = 0; i < maxEntries; i++)
+    {
+        auto& dst = objects[i];
+
+        auto* object = objMgr.GetLoadedObject(type, i);
```
duncanspumpkin, line 264: `const auto*` i think

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -255,6 +255,24 @@ static void scenario_fix_ghosts(rct_s6_data* s6)
...
+        if (object == nullptr || object->GetObjectEntry() == nullptr)
+        {
+            std::memset(&dst, 0xFF, sizeof(dst));
```
duncanspumpkin, line 267: add a comment that this is for save compatibility that it has an entry of 0xFF, ...

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -278,20 +296,18 @@ void S6Exporter::Export()
...
-        if (entry == nullptr || entryData == nullptr || entryData == reinterpret_cast<void*>(-1))
-        {
-            std::memset(&_s6.objects[i], 0xFF, sizeof(rct_object_entry));
-        }
-        else
-        {
-            _s6.objects[i] = *entry;
-        }
-    }
+    auto& objectMgr = OpenRCT2::GetContext()->GetObjectManager();
```
duncanspumpkin, line 299: const?

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1561,24 +1561,40 @@ class S6Importer final : public IParkImporter
         return justText.data();
     }
 
-    std::vector<rct_object_entry> GetRequiredObjects()
+    template<typename T>
+    static void AddRequiredObjects(std::vector<rct_object_entry>& required, const T& list, size_t internalLimit)
```
duncanspumpkin, line 1565: const internalLimit

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1561,24 +1561,40 @@ class S6Importer final : public IParkImporter
...
     {
-        std::vector<rct_object_entry> result;
         rct_object_entry nullEntry = {};
```
duncanspumpkin, line 1567: This is used in a number of locations perhaps we should have a constexpr defined for it.

duncanspumpkin (approved):
Code looks good.

src/openrct2/rct2/S6Importer.cpp
ZehMatt: There is unfortunately no easy way to constexpr initialize this to have all bytes as 0xFF, it uses union.

PR #15383, opened 2021-09-08
Make smart follow functions easier to follow

Broxzier:
- Opted for a switch-case instead of if-else if-else if-else
- Reuse entity - is already checked for null and doesn't change
- Remove unnecessary else blocks after if-block that returns
- Make `peep` argument const - following should not change the peep in any way

Turn whitespace changes off in the diff, since most of it are just indention changes.

If you want to test this, you can do so by using the "follow sprite" option in the sequence editor, then playing the sequence.

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -683,15 +683,12 @@ void viewport_update_smart_sprite_follow(rct_window* window)
             break;
 
         case EntityType::Guest:
-        case EntityType::Staff:
-        {
-            Peep* peep = static_cast<Peep*>(entity);
-            if (peep->Is<Guest>())
-                viewport_update_smart_guest_follow(window, peep);
-            else if (peep->Is<Staff>())
-                viewport_update_smart_staff_follow(window, peep);
+            viewport_update_smart_guest_follow(window, entity->As<Guest>());
             break;
-        }
+
+        case EntityType::Staff:
+            viewport_update_smart_staff_follow(window, entity->As<Staff>());
+            break;
 
         default: // All other types don't need any "smart" following; steam particle, duck, money effect, etc.
             window->viewport_focus_sprite.sprite_id = window->viewport_smart_follow_sprite;
@@ -700,7 +697,7 @@ void viewport_update_smart_sprite_follow(rct_window* window)
     }
 }
 
-viewport_focus viewport_update_smart_guest_follow(rct_window* window, const Peep* peep)
+viewport_focus viewport_update_smart_guest_follow(rct_window* window, const Guest* peep)
 {
     viewport_focus focus{};
     focus.type = VIEWPORT_FOCUS_TYPE_SPRITE;
@@ -759,7 +756,7 @@ viewport_focus viewport_update_smart_guest_follow(rct_window* window, const Peep
     return focus;
 }
 
-void viewport_update_smart_staff_follow(rct_window* window, const Peep* peep)
+void viewport_update_smart_staff_follow(rct_window* window, const Staff* peep)
 {
     sprite_focus focus = {};
 
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -700,7 +700,7 @@ void viewport_update_smart_sprite_follow(rct_window* window)
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/openrct2/interface/Viewport.cpp
```diff
@@ -719,80 +713,74 @@ viewport_focus viewport_update_smart_guest_follow(rct_window* window, Peep* peep
...
+            focus.coordinate.y = xy.y;
+            focus.coordinate.z = tile_element_height(xy) + (4 * COORDS_Z_STEP);
+            focus.sprite.type |= VIEWPORT_FOCUS_TYPE_COORDINATE;
         }
-        focus.coordinate.rotation = get_current_rotation();
     }
+    else
+    {
+        focus.sprite.type |= VIEWPORT_FOCUS_TYPE_SPRITE | VIEWPORT_FOCUS_TYPE_COORDINATE;
+        focus.sprite.pad_486 &= 0xFFFF;
```
Broxzier, line 753: Seems to be unused

src/openrct2/interface/Viewport.cpp
```diff
@@ -669,44 +669,38 @@ void viewport_update_sprite_follow(rct_window* window)
...
         {
-            // will never happen
-            window->viewport_smart_follow_sprite = SPRITE_INDEX_NULL;
-            window->viewport_target_sprite = SPRITE_INDEX_NULL;
-            return;
+            Peep* peep = static_cast<Peep*>(entity);
+            if (peep->Is<Guest>())
+                viewport_update_smart_guest_follow(window, peep);
+            else if (peep->Is<Staff>())
+                viewport_update_smart_staff_follow(window, peep);
```
Broxzier, line 692: Can probably be split, no local Peep needed by passing `entity->As<Guest/Staff>()` to the functions.

src/openrct2/interface/Viewport.cpp
```diff
@@ -669,44 +669,38 @@ void viewport_update_sprite_follow(rct_window* window)
...
+            break;
+
+        case EntityType::Guest:
+        case EntityType::Staff:
         {
-            // will never happen
-            window->viewport_smart_follow_sprite = SPRITE_INDEX_NULL;
-            window->viewport_target_sprite = SPRITE_INDEX_NULL;
-            return;
+            Peep* peep = static_cast<Peep*>(entity);
```
ZehMatt, line 688:
```suggestion
            if (auto* peep = entity->As<Guest>())
                viewport_update_smart_guest_follow(window, peep);
            else if (auto* staff = entity->As<Staff>())
                viewport_update_smart_staff_follow(window, staff);
```
Could also do it this way

src/openrct2/interface/Viewport.h
```diff
@@ -111,8 +111,8 @@ void viewports_invalidate(int32_t left, int32_t top, int32_t right, int32_t bott
 void viewport_update_position(rct_window* window);
 void viewport_update_sprite_follow(rct_window* window);
 void viewport_update_smart_sprite_follow(rct_window* window);
-viewport_focus viewport_update_smart_guest_follow(rct_window* window, Peep* peep);
-void viewport_update_smart_staff_follow(rct_window* window, Peep* peep);
+viewport_focus viewport_update_smart_guest_follow(rct_window* window, const Peep* peep);
+void viewport_update_smart_staff_follow(rct_window* window, const Peep* peep);
```
ZehMatt, line 115:
```suggestion
void viewport_update_smart_staff_follow(rct_window* window, const Staff* staff);
```
Makes a bit more sense I think

src/openrct2/interface/Viewport.cpp
Broxzier: Splitting them into their own `EntityType::Guest` and `EntityType::Staff` case, the check for `Is<Guest/Staff>` won't even be necessary anymore.

src/openrct2/interface/Viewport.cpp
duncanspumpkin: It's very very easy to mess this up the union used for this has a ton of gotchas.

Broxzier: Hmm, I've noticed no differences with this line removed, also haven't seen any change to `focus.location` which shares its memory.

tupaschoal (approved):
Seems alright

Issue #15371, opened 2021-09-08
Console spam after running bpb1000ticks


spacek531:
### Operating System

Windows 10

### OpenRCT2 build

85fcc8325

### Describe the issue

After bpb1000ticks finishes comparing, after a minute or two this spam appears:

![image](https://user-images.githubusercontent.com/3454156/132463463-cc4dbb2b-9b27-4358-adc1-faf1116f36f2.png)

I ran Evergreen Gardens for a few minutes and the spam didn't appear, but I may not have run it long enough.

### Steps to reproduce

1. 
2. 
3.

duncanspumpkin:
That would be the music object. Your parent commit is 454eabe. @Gymnasiast could you have a look.

spacek531:
[Six Flags circus.zip](https://github.com/OpenRCT2/OpenRCT2/files/7157640/Six.Flags.circus.zip)

I think it's related to the circus because this park with just a circus loads with the spam

Gymnasiast:
This should have been fixed as of https://github.com/OpenRCT2/OpenRCT2/commit/2f2e8b4c483c0dd6d7a74269830bcb6b93d549ed . Could you check?

spacek531:
Tested using 5087e77 using bpb1000ticks 0.3.4.1-0. I did not see any of the above pictured console spam.

Gymnasiast:
Of course it won’t happen on 5087e77, that’s a commit from before the issue was introduced. I need you to test with a _develop_ build.

spacek531:
My launcher was set to release on accident. OpenRCT2, v0.3.4.1 (427bb7b on develop) provided by GitHub doesn't create the spam during or after bpb1000ticks 0.3.4.1-8.

PR #15385, opened 2021-09-09
Fix #15322: Circus Music Not Playing

### LINKED ISSUE #15322: Circus does not have any sound
### Operating System

Ubuntu 20.04

### OpenRCT2 build

b2d0b54d135e1b82899e0e96affbf265c781179a

### Describe the issue

The sound effects from the Circus are completely gone.

It’s extremely likely this has been caused by the introduction of music objects, since the circus show was not included as one (it’s technically music, but behaves as a sound effect).

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Let a Circus start its show
2. Listen for any sound

### Attachments

_No response_

seanmajorpayne:
Added check to ride_update_music to see if ride type is circus. Separated circus music processing from the rest of the rides.

Previously, Circus would not play music.
Now, Circus will play music as expected.

Issue was just an 'if' statement that was returning immediately if the music flag for a ride was not set. As the circus is a sound effect, not music, it was continually returning.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -1805,6 +1805,7 @@ static void ride_music_update(Ride* ride)
             }
         }
     }
+
     // Select random tune from available tunes for a music style (of course only merry-go-rounds have more than one tune)
     if (ride->music_tune_id == 255)
     {
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -1754,22 +1754,25 @@ Staff* ride_get_assigned_mechanic(Ride* ride)
  */
 static void ride_music_update(Ride* ride)
 {
-    const auto& rtd = ride->GetRideTypeDescriptor();
-    if (!rtd.HasFlag(RIDE_TYPE_FLAG_MUSIC_ON_DEFAULT) && !rtd.HasFlag(RIDE_TYPE_FLAG_ALLOW_MUSIC))
-    {
-        return;
-    }
-
-    if (ride->status != RideStatus::Open || !(ride->lifecycle_flags & RIDE_LIFECYCLE_MUSIC))
-    {
-        ride->music_tune_id = 255;
-        return;
-    }
-
+    // Circus does not have "music." Music is a sound effect
     if (ride->type == RIDE_TYPE_CIRCUS)
     {
         Vehicle* vehicle = GetEntity<Vehicle>(ride->vehicles[0]);
-        if (vehicle != nullptr && vehicle->status != Vehicle::Status::DoingCircusShow)
+        if (vehicle == nullptr || vehicle->status != Vehicle::Status::DoingCircusShow)
+        {
+            ride->music_tune_id = 255;
+            return;
+        }
+    }
+    else
+    {
+        const auto& rtd = ride->GetRideTypeDescriptor();
+        if (!rtd.HasFlag(RIDE_TYPE_FLAG_MUSIC_ON_DEFAULT) && !rtd.HasFlag(RIDE_TYPE_FLAG_ALLOW_MUSIC))
+        {
+            return;
+        }
+
+        if (ride->status != RideStatus::Open || !(ride->lifecycle_flags & RIDE_LIFECYCLE_MUSIC))
         {
             ride->music_tune_id = 255;
             return;
@@ -1803,9 +1806,9 @@ static void ride_music_update(Ride* ride)
         }
     }
 
-    // Select random tune from available tunes for a music style (of course only merry-go-rounds have more than one tune)
     if (ride->music_tune_id == 255)
     {
+        // Select random tune from available tunes for a music style (of course only merry-go-rounds have more than one tune)
         auto& objManager = GetContext()->GetObjectManager();
         auto musicObj = static_cast<MusicObject*>(objManager.GetLoadedObject(ObjectType::Music, ride->music));
         if (musicObj != nullptr)
@@ -1814,7 +1817,6 @@ static void ride_music_update(Ride* ride)
             ride->music_tune_id = static_cast<uint8_t>(util_rand() % numTracks);
             ride->music_position = 0;
         }
-        return;
     }
 
     CoordsXYZ rideCoords = ride->stations[0].GetStart().ToTileCentre();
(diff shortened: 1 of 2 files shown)


duncanspumpkin:
When was this mistake introduced. We don't add to the changelog unless the issue was in the previous release.

seanmajorpayne:
I'm not sure. I'll double check the previous release.

duncanspumpkin:
Although i really doubt there will be an issue with it we should increment the network version just in case. The music is ultimately used for some peep interaction i think.

Gymnasiast:
@duncanspumpkin The bug has been introduced with the introduction of music objects. It is already present in 0.3.3. So it should be in the changelog.

seanmajorpayne:
@duncanspumpkin Do you need me to update something in the code for the network? Or is that a deployment issue?

duncanspumpkin:
See network base.cpp just increment the number in there. But leave that until review is completed as we may have modified it subsequently.

seanmajorpayne:
Okay no problem.

ocalhoun6:
> Although i really doubt there will be an issue with it we should increment the network version just in case. The music is ultimately used for some peep interaction i think.

Only the merry-go-round music and the fairground organ music, I think. So it shouldn't affect the circus.

Broxzier:
Dodgems and one other soundtrack affect guest thoughts. Better safe than sorry.

src/openrct2/ride/Ride.cpp
```diff
@@ -1814,7 +1817,6 @@ static void ride_music_update(Ride* ride)
             ride->music_tune_id = static_cast<uint8_t>(util_rand() % numTracks);
             ride->music_position = 0;
         }
-        return;
```
Gymnasiast, line 1817: Why did you remove this return?

src/openrct2/ride/Ride.cpp
```diff
@@ -1803,9 +1806,9 @@ static void ride_music_update(Ride* ride)
         }
     }
 
-    // Select random tune from available tunes for a music style (of course only merry-go-rounds have more than one tune)
     if (ride->music_tune_id == 255)
     {
+        // Select random tune from available tunes for a music style (of course only merry-go-rounds have more than one tune)
```
Gymnasiast, line 1811: Stray change, could you revert this?

src/openrct2/ride/Ride.cpp
seanmajorpayne: Oh, that's strange. I didn't see this in Github Desktop when I reviewed.

I'll revert.

src/openrct2/ride/Ride.cpp
seanmajorpayne: Yes will do

src/openrct2/ride/Ride.cpp
```diff
@@ -1754,22 +1754,25 @@ Staff* ride_get_assigned_mechanic(Ride* ride)
  */
 static void ride_music_update(Ride* ride)
 {
-    const auto& rtd = ride->GetRideTypeDescriptor();
```
Gymnasiast, line 1757: I think we could just remove this if altogether.

Gymnasiast (commented):
Don’t change anything yet. I want to test some things.

seanmajorpayne:
Okay no problem.

Gymnasiast:
OK, I checked what I wanted. You can now proceed with the line notes.
You can also add yourself to contributors.md if you haven’t done so already.

seanmajorpayne:
I've fixed those small issues. Should be ready for review/testing.

I added to contributors.md last PR, so all good there.

Gymnasiast (changes requested):
Tested it. It doesn’t seem to work for existing circuses.

For example: open European Cultural Festival (WW), go to Main Circus Tent and listen. There is no circus music, even after the visitors exit and new ones enter.

seanmajorpayne:
This does not appear to be a bug with the circus music directly, but something separate.

[See this line in RideAudio.cpp](https://github.com/OpenRCT2/OpenRCT2/blob/a5a63f839aca539ad75abf09a765e590e5ba973b/src/openrct2/ride/RideAudio.cpp#L194)

It is causing this log_warning.
```
object\ObjectManager.cpp:75 (ObjectManager::GetLoadedObject)]: Object index 65535 exceeds maximum for type 14.
```

I'm not sure what this means or why this would occur. Do you have any intuition of what is happening here?

Gymnasiast:
That would be this issue: https://github.com/OpenRCT2/OpenRCT2/issues/15371

It’s caused by music being set to OBJECT_ENTRY_INDEX_NULL (=65535). I don’t know if it’s related to this, though.

seanmajorpayne:
I believe the two are related. I am not able to replicate the circus sound issue you experienced on European Cultural Festival outside of that map. If you watch the console output and disable the circus, it should stop reporting that error. If you put it back on test or open, as soon as the music would start playing that error starts popping up again.

spacek531:
I could not recreate the spam by building a new circus. Are circuses being imported incorrectly?

seanmajorpayne:
What do you mean you could not recreate the spam? If you are talking about
console I only saw that for the one map for the pre-spawned circus.

src/openrct2/ride/Ride.cpp
```diff
@@ -1802,7 +1805,6 @@ static void ride_music_update(Ride* ride)
             }
         }
     }
```
Gymnasiast, line 1807:
```suggestion
    }

```

src/openrct2/ride/Ride.cpp
```diff
@@ -1754,22 +1754,25 @@ Staff* ride_get_assigned_mechanic(Ride* ride)
  */
 static void ride_music_update(Ride* ride)
 {
-    const auto& rtd = ride->GetRideTypeDescriptor();
-    if (!rtd.HasFlag(RIDE_TYPE_FLAG_MUSIC_ON_DEFAULT) && !rtd.HasFlag(RIDE_TYPE_FLAG_ALLOW_MUSIC))
+    // Circus does not have "music." Music is a sound effect
```
Gymnasiast, line 1757:
```suggestion
    // The circus does not have music in the normal sense - its “music” is a sound effect.
```

Gymnasiast:
Since this does not fully solve the issue, but does improve on it, I decided to revert the changelog and merge this, leaving #15322 open until it is fully fixed. Up until then, at least _some_ circuses will play their music again.

Gymnasiast (approved):

seanmajorpayne:
Sounds good. I will try to find the source of the other problem.

PR #15391, opened 2021-09-10
Fix #15390: Revert quadrant distribution change

### LINKED ISSUE #15390: Visual corruption on a horizontal line when looking south to north
### Operating System

Windows 10, 64bit

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (e284147 on develop) provided by GitHub

### Describe the issue

There's a horizontal line of visual corruption through the entire map. On that line land height marks are not present and just above that line track height marks are not displayed. It happens about 256 tiles down from the top.
Originally reported on the [forum](https://forums.openrct2.org/topic/5677-drawing-glitch-in-one-area-of-park/)

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Open any map that is at least 128x128.
2. Set rotation to look from south to north.
3. Look around, it might be easier to find with (track or land) height marks on, or on maps with large bodies of water (like alcatraz)

### Attachments

Happy park 2021-09-10 09-14-03
Hook Panorama 2021-09-10 09-12-38

ZehMatt:
This reverts the changes I made recently.

There appears to be some visual glitches, I need to look further into why this happens. The original code just clamps position between 0 and max which probably has some corner cases when two positions end up in the same bucket but are not part of the same region, this has also probably to do with the coordinate type of int32 which was int16 originally.

### DIFF
--- a/src/openrct2/paint/Paint.cpp
+++ b/src/openrct2/paint/Paint.cpp
@@ -56,27 +56,39 @@ static void PaintPSImageWithBoundingBoxes(rct_drawpixelinfo* dpi, paint_struct* 
 static void PaintPSImage(rct_drawpixelinfo* dpi, paint_struct* ps, uint32_t imageId, int32_t x, int32_t y);
 static uint32_t PaintPSColourifyImage(uint32_t imageId, ViewportInteractionItem spriteType, uint32_t viewFlags);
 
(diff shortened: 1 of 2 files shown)


ZehMatt:
I cleaned up the code a bit and documented how this actually works, I tricked myself into thinking this is some sort of hash but its really not.

duncanspumpkin (approved):

PR #15398, opened 2021-09-12
Shift vehicles at speed

spacek531:
For the NEDesigns grand finals going live on 2021-9-19. It's not 2 days until the deadline this time 😃


duncanspumpkin:
Are these park objs? Can't you write this in such a way that it is more generic. We don't want 1000 small variations of animations ideally.

duncanspumpkin:
Also your going to have to provide an object for this feel free to dm it on discord if you don't want it in public yet

spacek531:
This is for a plug-in object yes. I thought about a different implementation before arriving at the current one:
Two animation types, one for 8 frame animation and one for 6 frame animation. The log flume alternate vehicle field in the vehicle entry would be split into two fields: speed occupying the upper 5 bits and car index occupying the lower 3 bits. Using 3 bits for car index allows park objects to double the number of vehicle entries to 8 in the future. The speed bits would indicate the maximum speed of the vehicle; if the speed is above the maximum, it attempts to switch to the car indicated in the lower 3 bits. If the speed is below the maximum speed of the vehicle, it checks the maximum speed of the vehicle one index below and if the maximum speed of the lower car is above the vehicle's current speed, it switches to this car. Overall I felt that switching vehicles this way would help keep the animation types as simple as possible and allow the animation to change at different speeds by switching the car, but was limited by trying to reuse a single byte.

As far as how flexible the feature is in the first commit, I think it's the most flexible animation type to date: it has the ability to animate a vehicle with 1 standing frame, 1 standing frame and 8 walking frames, 1 standing frame, 8 walking frames, and 8 trotting frames, 1 standing frame, 8 walking frames, 8 trotting frames, and 6 galloping frames, and 1 standing frame, 8 walking frames, 8 trotting frames, and 8 galloping frames. Although this is flexible, it's not very simple.

While writing this, I had the idea to use a car_visual function as a wrapper to the default car_visual function but add vehicle switching code. This doesn't solve the issue of what speeds the should switch the vehicle, but it does mean that the animation type itself does not decide if the vehicle should change.

spacek531:
If vehicle switching is implemented, the canoes animation can be re-used for the intended object because it has 6 frames of animation as long as its animation speed can be changed. I'm going to take a look at https://github.com/OpenRCT2/OpenRCT2/pull/11269 and see what can be done to refactor it to be simpler and bring it up to date.

spacek531:
Completely rethinking this branch to pair with #11269 to get the desired results. Instead of one animation type that has several different cycles, a ride cycles between cars with different animation modes.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -1486,6 +1486,10 @@ bool Vehicle::OpenRestraints()
         if (curRide == nullptr)
             continue;
 
+        int currentVehicleIndex = vehicle->vehicle_type;
+        while (vehicle->ForceShiftDown() && vehicle->vehicle_type != currentVehicleIndex)
```
spacek531, line 1490: This could alternatively be tracking how many shifts it performs and stopping after a certain number, because it is possible to create an infinite loop.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9409,6 +9414,49 @@ int32_t Vehicle::UpdateTrackMotionMiniGolf(int32_t* outStation)
...
+    if (vehicleEntry == nullptr)
+        return true;
+    if (vehicleEntry->SpeedShift.LowerVehicle != vehicle_type)
+    {
+        vehicle_type = vehicleEntry->SpeedShift.LowerVehicle;
+        vehicleEntry = Entry();
+        if (vehicleEntry == nullptr)
+        {
+            // if (animation_frame >= vehicleEntry->AnimationFrames)
+            //    animation_frame = 0;
```
spacek531, line 9453: This is necessary already but there's no good way to tell how many frames a ride has except with https://github.com/OpenRCT2/OpenRCT2/pull/11269

spacek531:
Right now the shifting is defined by the vehicle but this has the danger of creating an infinite loop. At some point I'll change this to be defined from the ride entry instead of the vehicle entry, which will not be in danger of creating an infinite loop.

Gymnasiast:
Seems to have stalled, and requires a bit more justification about its usefulness to objects in general.

PR #15399, opened 2021-09-12
Remove Nullable and use std::optional instead

ZehMatt:


IntelOrca (approved):

Margen67:
What about this? https://github.com/OpenRCT2/OpenRCT2/blob/603be1c425cee35eb0c1fc2ba3b6e44d382b127a/src/openrct2/libopenrct2.vcxproj#L178

src/openrct2/network/NetworkBase.cpp
```diff
@@ -912,9 +911,9 @@ uint8_t NetworkBase::GetGroupIDByHash(const std::string& keyhash)
     const NetworkUser* networkUser = _userManager.GetUserByHash(keyhash);
 
     uint8_t groupId = GetDefaultGroup();
-    if (networkUser != nullptr && networkUser->GroupId.HasValue())
+    if (networkUser != nullptr && networkUser->GroupId)
```
Gymnasiast, line 914: We generally use explicit comparisons for anything that is not a boolean. E.g.
```suggestion
    if (networkUser != nullptr && networkUser->GroupId.has_value())
```

Gymnasiast (approved):

Gymnasiast:
Looks like a nice cleanup!

src/openrct2/network/NetworkBase.cpp
```diff
@@ -2058,7 +2057,7 @@ NetworkPlayer* NetworkBase::AddPlayer(const std::string& name, const std::string
             }
             else
             {
-                player->Group = networkUser->GroupId.GetValueOrDefault(GetDefaultGroup());
+                player->Group = networkUser->GroupId.has_value() ? *networkUser->GroupId : GetDefaultGroup();
```
tupaschoal, line 2060: Btw, this could have been simplified even further as we're now using optional:
```c++
player->Group = networkUser->GroupId.value_or(GetDefaultGroup());
```

Broxzier: @tupaschoal Wouldn't that always call `GetDefaultGroup()` while with the ternary operator it would only call it if `has_value()` returns false?

src/openrct2/network/NetworkUser.cpp
```diff
@@ -52,9 +52,9 @@ json_t NetworkUser::ToJson() const
     jsonData["name"] = Name;
 
     json_t jsonGroupId;
-    if (GroupId.HasValue())
+    if (GroupId.has_value())
     {
-        jsonGroupId = GroupId.GetValue();
+        jsonGroupId = *GroupId;
```
Broxzier, line 57:
```suggestion
        jsonGroupId = GroupId.value();
```

src/openrct2/network/NetworkUser.cpp
```diff
@@ -172,9 +172,9 @@ void NetworkUserManager::UnsetUsersOfGroup(uint8_t groupId)
     for (const auto& kvp : _usersByHash)
     {
         NetworkUser* networkUser = kvp.second;
-        if (networkUser->GroupId.HasValue() && networkUser->GroupId.GetValue() == groupId)
+        if (networkUser->GroupId.has_value() && *networkUser->GroupId == groupId)
```
Broxzier, line 175:
```suggestion
        if (networkUser->GroupId.has_value() && networkUser->value() == groupId)
```

src/openrct2/network/NetworkBase.cpp
```diff
@@ -912,9 +911,9 @@ uint8_t NetworkBase::GetGroupIDByHash(const std::string& keyhash)
     const NetworkUser* networkUser = _userManager.GetUserByHash(keyhash);
 
     uint8_t groupId = GetDefaultGroup();
-    if (networkUser != nullptr && networkUser->GroupId.HasValue())
+    if (networkUser != nullptr && networkUser->GroupId.has_value())
     {
-        const uint8_t assignedGroup = networkUser->GroupId.GetValue();
+        const uint8_t assignedGroup = *networkUser->GroupId;
```
Broxzier, line 916:
```suggestion
         const uint8_t assignedGroup = networkUser->GroupId.value();
```

Broxzier (commented):
Left some line notes for readability. Using the `value` member function makes it clear that it's not an actual dereference operation.

src/openrct2/network/NetworkBase.cpp
Broxzier:
```suggestion
                player->Group = networkUser->GroupId.has_value() ? networkUser->GroupId.value() : GetDefaultGroup(); 
```

Gymnasiast:
I don’t think it would improve readability. Quite the opposite, in fact.

src/openrct2/network/NetworkBase.cpp
tupaschoal: > @tupaschoal Wouldn't that always call `GetDefaultGroup()` while with the ternary operator it would only call it if `has_value()` returns false?

Good point, I think so

PR #15401, opened 2021-09-12
Close #15353: Add VerticalFactor and LateralFactor fields to TrackElementDescriptor

### LINKED ISSUE #15353: Add VerticalFactor and LateralFactor fields in TrackElementDescriptor to be used in Vehicle::GetGForces

frutiemax:


duncanspumpkin:
Don't close old PRs and open new ones it just loses context. Always work from the one PR.

frutiemax:
Can anyone review this PR?

src/openrct2/ride/TrackData.cpp
```diff
@@ -5427,6 +5427,765 @@ constexpr static uint8_t TrackTypeToSpinFunction[TrackElemType::Count] = {
     NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN, NO_SPIN
 };
 
+template<const int32_t TConstant> static int32_t EvaluatorConst(const int16_t)
```
ZehMatt, line 5430:
```suggestion
template<int32_t TConstant> static int32_t EvaluatorConst(const int16_t)
```
Template arguments are not mutable.

src/openrct2/ride/TrackData.h
```diff
@@ -90,6 +90,9 @@ struct TrackElementDescriptor
 
     rct_trackdefinition Definition;
     uint8_t SpinFunction;
+
+    std::function<int32_t(const int16_t)> VerticalFactor;
```
ZehMatt, line 94: I would prefer we have a well defined function signature here. e.x.:
```
using TrackComputeFunction = int32_t(*)(const int16_t);
```
instead of std::function.

ZehMatt (changes requested):
One minor thing otherwise ok with me.

ZehMatt (approved):

src/openrct2/ride/Vehicle.cpp
```diff
@@ -5788,500 +5788,10 @@ GForces Vehicle::GetGForces() const
...
-            break;
-        case TrackElemType::SBendRight:
-        case TrackElemType::SBendRightCovered:
-            lateralFactor = (track_progress < 48) ? -98 : 98;
-            // 6d7608
-            break;
-        case TrackElemType::LeftVerticalLoop:
-        case TrackElemType::RightVerticalLoop:
-            vertFactor = (abs(track_progress - 155) / 2) + 28;
-            // 6d7690
```
Broxzier, line 5982: @duncanspumpkin with this PR these address comments are removed. Is that alright?

duncanspumpkin: They will have no use so sure thats fine.

src/openrct2/ride/TrackData.cpp
```diff
@@ -5427,6 +5427,765 @@ constexpr static uint8_t TrackTypeToSpinFunction[TrackElemType::Count] = {
...
+        return vertFactor;
+    vertFactor = 0;
+    if (progress < 96)
+        return vertFactor;
+    vertFactor = 103;
+    if (progress < 128)
+        return vertFactor;
+    vertFactor = -103;
+    // 6d74A0
+    return vertFactor;
```
Broxzier, line 5499: Same here:
```suggestion
    if (progress < 32)
        return 103;
    if (progress < 64)
        return -103;
    if (progress < 96)
        return 0;
    if (progress < 128)
        return 103;
    return -103;
```

src/openrct2/ride/TrackData.cpp
```diff
@@ -5427,6 +5427,765 @@ constexpr static uint8_t TrackTypeToSpinFunction[TrackElemType::Count] = {
...
+        return vertFactor;
+    vertFactor = 0;
+    if (progress < 96)
+        return vertFactor;
+    vertFactor = -103;
+    if (progress < 128)
+        return vertFactor;
+    vertFactor = 103;
+    // 6D74CA
+    return vertFactor;
```
Broxzier, line 5518:
```suggestion
    if (progress < 32)
        return -103;
    if (progress < 64)
        return 103;
    if (progress < 96)
        return 0;
    if (progress < 128)
        return -103;
    return 103;
```

src/openrct2/ride/TrackData.cpp
```diff
@@ -5427,6 +5427,765 @@ constexpr static uint8_t TrackTypeToSpinFunction[TrackElemType::Count] = {
...
+            return EvaluatorConst<0>;
+        case TrackElemType::Up60ToUp90:
+        case TrackElemType::Down90ToDown60:
+            return EvaluatorConst<0>;
+        case TrackElemType::Up90ToUp60:
+        case TrackElemType::Down60ToDown90:
+            return EvaluatorConst<0>;
+        case TrackElemType::LeftEighthToDiag:
+        case TrackElemType::LeftEighthToOrthogonal:
+            return EvaluatorConst<0>;
```
Broxzier, line 5778: This is not right. Previously `lateralFactor` was assigned 137.
```suggestion
            return EvaluatorConst<137>;
```

Broxzier (changes requested):
I've carefully went over the changes, and spotted only one mistake. I've also left a few suggestions for code improvements.

src/openrct2/ride/TrackData.cpp
```diff
@@ -5428,6 +5428,745 @@ constexpr static uint8_t TrackTypeToSpinFunction[TrackElemType::Count] = {
...
+}
+
+static int32_t EvaluatorVerticalLoop(const int16_t progress)
+{
+    return (abs(progress - 155) / 2) + 28;
+}
+
+static int32_t EvaluatorHalfLoopUp(const int16_t progress)
+{
+    return ((static_cast<uint16_t>(-(progress - 155))) / 2) + 28;
```
tupaschoal, line 5443: What if the result is negative? The cast will return a crazy value.

src/openrct2/ride/TrackData.cpp
```diff
@@ -5428,6 +5428,745 @@ constexpr static uint8_t TrackTypeToSpinFunction[TrackElemType::Count] = {
...
+{
+    if (progress < 32)
+        return -150;
+    if (progress < 64)
+        return 150;
+    if (progress < 96)
+        return 0;
+    if (progress < 128)
+        return 150;
+    return -150;
```
tupaschoal, line 5461:
```suggestion
    if (progress < 32)
        return -150;
    else if (progress < 64)
        return 150;
    else if (progress < 96)
        return 0;
    else if (progress < 128)
        return 150;
    else 
        return -150;
```

I wondered if this makes it any more efficient for the compiler, but it seems like it doesn't. Maybe it makes things a bit clearer?

Broxzier: These will both compile to the same instructions

src/openrct2/ride/TrackData.cpp
Broxzier: It's the exact same as how it was before this PR: https://github.com/OpenRCT2/OpenRCT2/pull/15401/files#diff-e80b0c445c9e2b2302d2ddd41d9feded6f6b18f9d68c38d49d02fc52b5061cd7L5863

frutiemax:
Do I need to make some changes before you merge this PR?

tupaschoal (approved):

tupaschoal (approved):

PR #15403, opened 2021-09-12
Modify object loading for JSON-only objects

Gymnasiast:


src/openrct2/object/ObjectManager.cpp
```diff
@@ -209,7 +215,7 @@ class ObjectManager final : public IObjectManager
         {
             const ObjectRepositoryItem* item = &_objectRepository.GetObjects()[i];
             if (item->LoadedObject != nullptr && IsObjectCustom(item) && item->LoadedObject->GetLegacyData() != nullptr
-                && !item->LoadedObject->IsJsonObject())
+                && item->LoadedObject->GetGeneration() == ObjectGeneration::DAT)
```
Gymnasiast, line 218: When the NSF is there, this bit should be removed.

src/openrct2/object/ObjectManager.cpp
```diff
@@ -352,42 +358,42 @@ class ObjectManager final : public IObjectManager
 
     Object* RepositoryItemToObject(const ObjectRepositoryItem* ori, std::optional<int32_t> slot = {})
     {
-        if (ori == nullptr)
-            return nullptr;
-
-        Object* loadedObject = ori->LoadedObject.get();
-        if (loadedObject != nullptr)
-            return loadedObject;
```
ZehMatt, line 360: Would be nice if you could revert it back to this, its a lot easier to understand imho.

Gymnasiast: Could you comment on this, @IntelOrca ? I don’t have any strong feelings either way.

src/openrct2/object/ObjectManager.cpp
Broxzier: Sounds like a TODO comment will be useful here, to make them easier to find.

src/openrct2/object/ObjectManager.cpp
IntelOrca: You can revert it if you want. This is subjective though.

Gymnasiast: The reason I asked is because you wrote it, and I’m quite sure you must have had a good reason to change it. I’m just porting your object changes from the NSF to develop.

ZehMatt: The code is still doing the same, it just bails early which is easier to understand if you read a function, they are basically the pre-conditions before even bothering with anything.

Gymnasiast: Ah, I see. I’ll restore that bit.

src/openrct2/object/ObjectManager.cpp
```diff
@@ -374,20 +381,19 @@ class ObjectManager final : public IObjectManager
...
+            auto object = GetOrLoadObject(ori);
             if (object != nullptr)
             {
                 if (_loadedObjects.size() <= static_cast<size_t>(*slot))
                 {
                     _loadedObjects.resize(slot.value() + 1);
                 }
                 loadedObject = object;
-                _loadedObjects[slot.value()] = object;
+                _loadedObjects[slot.value()] = std::move(object);
```
Gymnasiast, line 392: Is this change correct, @Zehmatt ? I know some stuff has changed in develop since this was put in, so it may be wrong/outdated.

ZehMatt: Yeah its probably not right, GetOrLoadObject returns a pointer value now.

src/openrct2/object/ObjectManager.cpp
```diff
@@ -374,20 +381,19 @@ class ObjectManager final : public IObjectManager
         }
         if (slot.has_value())
         {
-            auto* object = GetOrLoadObject(ori);
+            auto object = GetOrLoadObject(ori);
```
ZehMatt, line 384:
```suggestion
            auto* object = GetOrLoadObject(ori);
```

src/openrct2/Context.cpp
```diff
@@ -717,7 +717,7 @@ namespace OpenRCT2
                 // which the window function doesn't like
                 auto intent = Intent(WC_OBJECT_LOAD_ERROR);
                 intent.putExtra(INTENT_EXTRA_PATH, path);
-                intent.putExtra(INTENT_EXTRA_LIST, const_cast<rct_object_entry*>(e.MissingObjects.data()));
+                intent.putExtra(INTENT_EXTRA_LIST, const_cast<ObjectEntryDescriptor*>(e.MissingObjects.data()));
```
ZehMatt, line 720: I just realized that this is actually a pretty bad idea. Its passing a pointer of a temporary object declared in the catch statement. This is not for this PR but always good re-visiting lines, should raise an issue for this.

ZehMatt (changes requested):
I think it looks overall good, just two lines as far as I can tell.

ZehMatt:
I did some testing
I get a lot of
```
(ObjectManager::GetLoadedObject)]: Object index 65535 exceeds maximum for type 15.```

This is called from PathElement::GetSurfaceEntry, GetSurfaceEntryIndex() returning OBJECT_ENTRY_INDEX_NULL. We should probably not call any object manager things with a bad index.

File: src/openrct2/object/ObjectManager.cpp [hunk 6]
```diff
@@ -69,6 +69,12 @@ class ObjectManager final : public IObjectManager
 
     Object* GetLoadedObject(ObjectType objectType, size_t index) override
     {
+        // This is sometimes done deliberately (to avoid boilerplate), so no need to log_warn for this.
+        if (index == OBJECT_ENTRY_INDEX_NULL)
```
ZehMatt, line 73: Wouldn't it be better to solve this a bit higher up? We don't need to call this if we know that this will result null anyway.

Gymnasiast: It would mean extra boilerplate in a few dozen places, harming readability. We already get this kind of spam when working with music objects. That’s why I preferred this solution.

ZehMatt: Fair enough.

ZehMatt (approved):
I tested this with the infamous dome-export park and it all looks good. Opened a couple saves no issues. Overall good stuff

src/openrct2/object/ObjectManager.cpp
IntelOrca: This is fine, it means anything can freely pass an object ID into GetLoadObject without worrying about the range as it then only has to check `nullptr` afterwards. Not doing this means you have to make everything that calls this check for invalid range and you risk some callers not doing that.

ZehMatt: Well it is already protected against values out of the valid range, I just think it would make sense to check this in lets say GetSurfaceObject before calling GetContext(), GetObjectManager(), GetObject etc but only for the sake of avoiding those calls but it doesn't matter too much.

IntelOrca: That makes senses, although there will still be some things that will call GetObject directly and not be in a wrapper like GetSurfaceObject, so you'd still need both.

PR #15407, opened 2021-09-13
Make use of std::optional strict

ZehMatt:


src/openrct2/interface/Screenshot.cpp
```diff
@@ -161,14 +161,14 @@ std::string screenshot_dump_png(rct_drawpixelinfo* dpi)
...
     auto path = screenshot_get_next_path();
 
-    if (path == std::nullopt)
+    if (!path.has_value())
     {
         return "";
     }
 
-    if (WriteDpiToFile(path->c_str(), dpi, gPalette))
+    if (WriteDpiToFile(path.value(), dpi, gPalette))
```
tupaschoal, line 169: Shouldn't this be:
```suggestion
    if (WriteDpiToFile(path.value().c_str(), dpi, gPalette))
```

src/openrct2/interface/Screenshot.cpp
```diff
@@ -196,8 +196,8 @@ std::string screenshot_dump_png_32bpp(int32_t width, int32_t height, const void*
         image.Depth = 32;
         image.Stride = width * 4;
         image.Pixels = std::vector<uint8_t>(pixels8, pixels8 + pixelsLen);
-        Imaging::WriteToFile(path->c_str(), image, IMAGE_FORMAT::PNG_32);
-        return *path;
+        Imaging::WriteToFile(path.value(), image, IMAGE_FORMAT::PNG_32);
```
tupaschoal, line 199:
```suggestion
        Imaging::WriteToFile(path.value().c_str(), image, IMAGE_FORMAT::PNG_32);
```

src/openrct2/interface/Screenshot.cpp
```diff
@@ -412,7 +412,7 @@ void screenshot_giant()
         dpi = CreateDPI(viewport);
 
         RenderViewport(nullptr, viewport, dpi);
-        WriteDpiToFile(path->c_str(), &dpi, gPalette);
+        WriteDpiToFile(path.value(), &dpi, gPalette);
```
tupaschoal, line 415:
```suggestion
        WriteDpiToFile(path.value().c_str(), &dpi, gPalette);
```

src/openrct2/ride/Ride.cpp
```diff
@@ -2868,8 +2868,8 @@ static void ride_set_boat_hire_return_point(Ride* ride, CoordsXYE* startElement)
         trackType = trackBeginEnd.begin_element->AsTrack()->GetTrackType();
         auto newCoords = GetTrackElementOriginAndApplyChanges(
             { trackCoords, static_cast<Direction>(direction) }, trackType, 0, &returnPos.element, 0);
-        returnPos = newCoords == std::nullopt ? CoordsXYE{ trackCoords, returnPos.element }
-                                              : CoordsXYE{ *newCoords, returnPos.element };
+        returnPos = !newCoords.has_value() ? CoordsXYE{ trackCoords, returnPos.element }
+                                           : CoordsXYE{ *newCoords, returnPos.element };
```
tupaschoal, line 2872: I think it becomes clearer if we remove the negation and swap:

```suggestion
        returnPos = newCoords.has_value() ? CoordsXYE{ *newCoords, returnPos.element }
                                          : CoordsXYE{ trackCoords, returnPos.element };
```

tupaschoal (commented):
This is nice, I was hoping you had found a way to force this to be required, but seems like there isn't? (Or did I skip over it?)

Why is the Xcode dump necessary?

src/openrct2/interface/Screenshot.cpp
ZehMatt: The argument type is std::string_view so I don't think c_str() should be used here.

src/openrct2/interface/Screenshot.cpp
ZehMatt: Same as above

ZehMatt:
> This is nice, I was hoping you had found a way to force this to be required, but seems like there isn't? (Or did I skip over it?)
> 
> Why is the Xcode dump necessary?

There is as far as I can tell no way to enforce it other than clang-tidy maybe which we don't use, as for why the xcode change, see https://github.com/OpenRCT2/OpenRCT2/runs/3589851749

The version was already in raised in NSF also, this enables a few other missing features.

src/openrct2/interface/Screenshot.cpp
ZehMatt: See above

src/openrct2/interface/Screenshot.cpp
tupaschoal: Ah, I thought that might have been the case, wanted to make sure the change was intentional.

tupaschoal (approved):

src/openrct2/interface/Screenshot.cpp
ZehMatt: I see, I think it would be a compiler error to convert std::string to const char* but good thinking.

src/openrct2/ride/Ride.cpp
```diff
@@ -2868,8 +2868,8 @@ static void ride_set_boat_hire_return_point(Ride* ride, CoordsXYE* startElement)
         trackType = trackBeginEnd.begin_element->AsTrack()->GetTrackType();
         auto newCoords = GetTrackElementOriginAndApplyChanges(
             { trackCoords, static_cast<Direction>(direction) }, trackType, 0, &returnPos.element, 0);
-        returnPos = newCoords == std::nullopt ? CoordsXYE{ trackCoords, returnPos.element }
-                                              : CoordsXYE{ *newCoords, returnPos.element };
+        returnPos = newCoords.has_value() ? CoordsXYE{ *newCoords, returnPos.element }
+                                          : CoordsXYE{ trackCoords, returnPos.element };
```
Broxzier, line 2872:
```suggestion
        returnPos = newCoords.has_value() ? CoordsXYE{ newCoords.value(), returnPos.element }
                                          : CoordsXYE{ trackCoords, returnPos.element };
```

Broxzier (changes requested):
Changes look good, I've only found one case where a dereference operator is still being used.

Broxzier (approved):

PR #15408, opened 2021-09-13
Replace sprite_left,t,r,b with SpriteRect

### LINKED ISSUE #15405: Refactor EntityBase sprite_l,t,r,b with ScreenRect
As part of the NSF we need to increase the size of any Screen coordinates this includes the EntityBase::sprite_left, top, right, bottom. It should be refactored to use a ScreenRect. In the process a new null value should be used as 0x8000 could be a valid screen x position in the nsf due to the increased map sizes.

duncanspumpkin:
This is required for the NSF. I had to add a silly constructor to rct_sprite for now. Will eventually be removed one day...
Replays will likely fail.

Close #15405

### DIFF
--- a/src/openrct2-ui/interface/ViewportInteraction.cpp
+++ b/src/openrct2-ui/interface/ViewportInteraction.cpp
@@ -713,11 +713,11 @@ PeepDistance GetClosestPeep(const ScreenCoordsXY& viewportCoords, const int32_t 
 {
     for (auto peep : EntityList<T>())
     {
-        if (peep->sprite_left == LOCATION_NULL)
+        if (peep->x == LOCATION_NULL)
             continue;
 
-        auto distance = abs(((peep->sprite_left + peep->sprite_right) / 2) - viewportCoords.x)
-            + abs(((peep->sprite_top + peep->sprite_bottom) / 2) - viewportCoords.y);
+        auto distance = abs(((peep->SpriteRect.GetLeft() + peep->SpriteRect.GetRight()) / 2) - viewportCoords.x)
+            + abs(((peep->SpriteRect.GetTop() + peep->SpriteRect.GetBottom()) / 2) - viewportCoords.y);
         if (distance > maxDistance)
             continue;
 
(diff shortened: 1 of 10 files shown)


src/openrct2-ui/interface/ViewportInteraction.cpp
```diff
@@ -713,11 +713,11 @@ PeepDistance GetClosestPeep(const ScreenCoordsXY& viewportCoords, const int32_t
 {
     for (auto peep : EntityList<T>())
     {
-        if (peep->sprite_left == LOCATION_NULL)
+        if (peep->x == LOCATION_NULL)
```
duncanspumpkin, line 716: This is changed to x as both sprite_left and x are set to LOCATION_NULL under the same situations so no need to dirty both of them with the null state.

src/openrct2/world/Sprite.cpp
```diff
@@ -533,10 +532,9 @@ void sprite_set_coordinates(const CoordsXYZ& spritePos, EntityBase* sprite)
 {
     auto screenCoords = translate_3d_to_2d_with_z(get_current_rotation(), spritePos);
 
-    sprite->sprite_left = screenCoords.x - sprite->sprite_width;
-    sprite->sprite_right = screenCoords.x + sprite->sprite_width;
-    sprite->sprite_top = screenCoords.y - sprite->sprite_height_negative;
-    sprite->sprite_bottom = screenCoords.y + sprite->sprite_height_positive;
+    sprite->SpriteRect = ScreenRect(
+        screenCoords.x - sprite->sprite_width, screenCoords.y - sprite->sprite_height_negative,
+        screenCoords.x + sprite->sprite_width, screenCoords.y + sprite->sprite_height_positive);
```
tupaschoal, line 537:
```suggestion
        screenCoords - ScreenCordsXY{ sprite->sprite_width, sprite->sprite_height_negative },
        screenCoords + ScreenCordsXY{ sprite->sprite_width, sprite->sprite_height_positive });
```
Footprint is slightly smaller, and I think that there is a `ScreenRect` ctor that takes a pair of `ScreenCoordsXY`

src/openrct2/world/Sprite.h
```diff
@@ -23,6 +23,10 @@ union rct_sprite
 {
     uint8_t pad_00[0x200];
     EntityBase base;
+    rct_sprite()
```
tupaschoal, line 26: Maybe add a comment as to why this is needed and what is required for it to be removed?

src/openrct2/world/Sprite.cpp
duncanspumpkin: Hmm, i don't really like that constructor. It's not normal for a rectangle constructor to be defined as two points normally it would be a point and a pair of width height. I would think that a two screen coord constructor would be xy widthheight

PR #15410, opened 2021-09-13
Use std::nullopt

Broxzier:


tupaschoal (approved):
Looks ok, though one CI seems to have gone awry

Broxzier:
@tupaschoal I saw no reds, but one job (mac) that was interrupted? Anyway, I reran them and they're all green now.

tupaschoal:
> @tupaschoal I saw no reds, but one job (mac) that was interrupted? Anyway, I reran them and they're all green now.

Sometimes it fails with a timeout inexplicably, but if it's all green now, please proceed

PR #15411, opened 2021-09-13
Fix #13524: macOS Universal app

### LINKED ISSUE #13524: macOS arm64 support
Opening to track future efforts to build OpenRCT2 natively for Apple Silicon

adam-bloom:
This PR adds support for building universal macOS apps. Due to the continued desire to support different minimum macOS versions (and cmake only exposing a single flag), this builds each arch separately and then combines them into a universal app bundle. Note that the minimum is bumped to 10.14, since cmake does not support libopenrct2 having 10.14 but the enclosing project using 10.13 - we'd need to run two completely separate cmake invocations/builds to do that, and our project is not set up for it.

We really should make a decision about using cmake vs Xcode as the official macOS build tool. Our Xcode project is currently a bit of a mess (it seems files are arbitrarily included in libopenrct2 vs the final executable....cmake at least cleanly organizes this by source directory as we all expect), but it is possible to clean up and get the same functionality.

There's a mildly ugly workaround in here for sdl2 cross-compilation issues - I have yet to get a chance to open an issue on their repo to address. It was a fun head scratcher for a few hours as to why an Intel only header was being pulled into arm64 builds.

This PR should remain in draft until there is a new dependencies release (and this is pointed back at the main Dependencies artifacts instead of my alpha releases) and we get some user acceptance testing.


adam-bloom:
I've created libsdl-org/SDL#4753 for the SDL header issue. I don't think it's worth working around it with a vcpkg patch - probably best to wait for a proper fix upstream. The workaround here is simply setting a flag if building for arm64. Not ideal, but nowhere near ugly enough to care too much either. We can hopefully remove it eventually.

majchrzamemil:
HI, OpenRCT2-macOS-universal artifact has been tested on Apple M1 and run both from native terminal and terminal running on rosetta. While running from native terminal OpenRCT2 runs using ARM without rosetta
image

CMakeLists.txt
```diff
@@ -110,11 +125,16 @@ if (MACOS_USE_DEPENDENCIES)
...
+    # set(MACOS_DYLIBS_VERSION "28")
+    # set(MACOS_DYLIBS_ZIPFILE "openrct2-libs-v${MACOS_DYLIBS_VERSION}-x64-macos-dylibs.zip")
+    # set(MACOS_DYLIBS_SHA1 "29e5480376cf4ac5943f114387e32685204c8b78")
+    set(MACOS_DYLIBS_VERSION "29-alpha2")
+    set(MACOS_DYLIBS_ZIPFILE "openrct2-libs-v${MACOS_DYLIBS_VERSION}-universal-macos-dylibs.zip")
+    set(MACOS_DYLIBS_SHA1 "bb8899c4120f1add232af7f9ac774ea622e01ad3")
     set(MACOS_DYLIBS_DIR "${ROOT_DIR}/lib/macos")
-    set(MACOS_DYLIBS_URL "https://github.com/OpenRCT2/Dependencies/releases/download/v${MACOS_DYLIBS_VERSION}/${MACOS_DYLIBS_ZIPFILE}")
+    # set(MACOS_DYLIBS_URL "https://github.com/OpenRCT2/Dependencies/releases/download/v${MACOS_DYLIBS_VERSION}/${MACOS_DYLIBS_ZIPFILE}")
+    set(MACOS_DYLIBS_URL "https://github.com/adam-bloom/Dependencies/releases/download/v${MACOS_DYLIBS_VERSION}/${MACOS_DYLIBS_ZIPFILE}")
```
janisozaur, line 137: Does https://github.com/OpenRCT2/Dependencies/pull/28 address that already? Can you change it to upstream link?

src/openrct2-ui/CMakeLists.txt
```diff
@@ -35,6 +35,11 @@ file(GLOB_RECURSE OPENRCT2_UI_HEADERS
 if (APPLE)
     file(GLOB_RECURSE OPENRCT2_UI_MM_SOURCES "${CMAKE_CURRENT_LIST_DIR}/*.mm")
     set_source_files_properties(${OPENRCT2_UI_MM_SOURCES} PROPERTIES COMPILE_FLAGS "-x objective-c++ -fmodules")
+    if ("${CMAKE_OSX_ARCHITECTURES}" MATCHES "arm64")
+        # cross-compilation workaround for SDL2
```
janisozaur, line 39: Can you provide upstream ticket link here?

CMakeLists.txt
adam-bloom: OpenRCT2/Dependencies#29 addresses this. I just marked that as ready for review.

src/openrct2-ui/CMakeLists.txt
adam-bloom: Yes - let me update the comment. It's been resolved upstream, so we just need to wait for a new release, and then bump the vcpkg SDL2 version. We could add a vcpkg patch with the same change (until there is a new release), but this workaround is quite clean so I'm not sure it's worth creating a patch for it.

adam-bloom:
> Needs updating to v29 of dependencies, right?

Yes, and done!

I think the only other item that needs to be settled is whether we want to keep updating the Xcode project or not, as I have not updated that for the universal build. We should likely either update it, or move the official release upload to the new universal app.

adam-bloom:
> I can honestly say I'm the only person I've seen contribute that likes Xcode, just because there's nothing I despise more than using command line to build stuff, so yeah dropping Xcode is probably fine.

@pizza2004 have you used vscode yet? There is a cmake extension available - pretty sure you can build without touching a CLI if you don’t want to. I haven’t used the extension for much besides syntax highlighting myself, but I think you might be able to do what you seek.

janisozaur:
@adam-bloom can you rebase to resolve the conflicts? Seems it can be merged then.

janisozaur:
Seems I can do that with github UI now :)

PR #15412, opened 2021-09-14
Allow game actions to store custom data instead of using inheritance

ZehMatt:
As this PR illustrates this will reduce quite some code when game actions want to supply custom data with the result which currently requires to implement a new class based from GameAction::Result which also requires memory allocations. I only did BannerPlaceAction for now as everything remains compatible, once all uses are refactored game actions can be even further simplified in a lot of areas.

### DIFF
--- a/src/openrct2-ui/windows/TopToolbar.cpp
+++ b/src/openrct2-ui/windows/TopToolbar.cpp
@@ -1983,11 +1983,12 @@ static void window_top_toolbar_scenery_tool_down(const ScreenCoordsXY& windowPos
             CoordsXYZD loc{ gridPos, z, direction };
             auto primaryColour = gWindowSceneryPrimaryColour;
             auto bannerPlaceAction = BannerPlaceAction(loc, selectedScenery, primaryColour);
-            bannerPlaceAction.SetCallback([=](const GameAction* ga, const BannerPlaceActionResult* result) {
+            bannerPlaceAction.SetCallback([=](const GameAction* ga, const GameActions::Result* result) {
                 if (result->Error == GameActions::Status::Ok)
                 {
+                    auto data = result->GetData<BannerPlaceActionResult>();
                     OpenRCT2::Audio::Play3D(OpenRCT2::Audio::SoundId::PlaceItem, result->Position);
-                    context_open_detail_window(WD_BANNER, result->bannerId);
+                    context_open_detail_window(WD_BANNER, data.bannerId);
                 }
             });
             GameActions::Execute(&bannerPlaceAction);
--- a/src/openrct2/actions/BannerPlaceAction.h
+++ b/src/openrct2/actions/BannerPlaceAction.h
@@ -11,18 +11,12 @@
 
 #include "GameAction.h"
 
-class BannerPlaceActionResult final : public GameActions::Result
+struct BannerPlaceActionResult
 {
-public:
-    BannerPlaceActionResult();
-    BannerPlaceActionResult(GameActions::Status err);
-    BannerPlaceActionResult(GameActions::Status err, rct_string_id msg);
-    BannerPlaceActionResult(GameActions::Status err, rct_string_id title, rct_string_id message);
-
     BannerIndex bannerId = BANNER_INDEX_NULL;
 };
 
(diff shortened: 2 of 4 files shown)


tupaschoal (commented):
This looks very promising, I'm just a bit confused with all of the `BannerPlaceActionResult` ctors that were deleted. Were they just copy+paste bloat?

ZehMatt:
> This looks very promising, I'm just a bit confused with all of the `BannerPlaceActionResult` ctors that were deleted. Were they just copy+paste bloat?

Yeah to forward it to GameAction::Result

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -1983,11 +1983,12 @@ static void window_top_toolbar_scenery_tool_down(const ScreenCoordsXY& windowPos
             CoordsXYZD loc{ gridPos, z, direction };
             auto primaryColour = gWindowSceneryPrimaryColour;
             auto bannerPlaceAction = BannerPlaceAction(loc, selectedScenery, primaryColour);
-            bannerPlaceAction.SetCallback([=](const GameAction* ga, const BannerPlaceActionResult* result) {
+            bannerPlaceAction.SetCallback([=](const GameAction* ga, const GameActions::Result* result) {
                 if (result->Error == GameActions::Status::Ok)
                 {
+                    auto data = result->GetData<BannerPlaceActionResult>();
```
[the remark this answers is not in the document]
ZehMatt: Well this shouldn't throw when the result is Ok I also thought about it, but in 99.9% of the case this shouldn't throw, and if it throws it only means the type is wrong and should be fixed. I mean I can wrap it in try/catch but I don't see a big benefit.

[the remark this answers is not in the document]
Broxzier: Figured that is the case, though I wasn't sure since actions can be sent over the network.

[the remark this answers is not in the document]
ZehMatt: In case it doesn't properly network it then it won't even come to the callback, if the data is bad it will fail in query so I think its safe.

src/openrct2/actions/BannerPlaceAction.h
```diff
@@ -11,18 +11,12 @@
 
 #include "GameAction.h"
 
-class BannerPlaceActionResult final : public GameActions::Result
+struct BannerPlaceActionResult
```
Broxzier, line 14: This can be moved out of the header file.

src/openrct2/actions/BannerPlaceAction.h
```diff
@@ -11,18 +11,12 @@
...
-    BannerPlaceActionResult();
-    BannerPlaceActionResult(GameActions::Status err);
-    BannerPlaceActionResult(GameActions::Status err, rct_string_id msg);
-    BannerPlaceActionResult(GameActions::Status err, rct_string_id title, rct_string_id message);
-
     BannerIndex bannerId = BANNER_INDEX_NULL;
 };
 
-DEFINE_GAME_ACTION(BannerPlaceAction, GameCommand::PlaceBanner, BannerPlaceActionResult)
+DEFINE_GAME_ACTION(BannerPlaceAction, GameCommand::PlaceBanner, GameActions::Result)
```
Broxzier, line 19: Is `GameActions::Result` still necessary? Perhaps a new macro to remove this (or make it the default?) would be nicer.

src/openrct2/actions/BannerPlaceAction.h
ZehMatt: Move it where? Its needed in BannerPlaceAction.cpp to assign the data, and callers need to know the type as well.

src/openrct2/actions/BannerPlaceAction.h
ZehMatt: This is something I want to do once they are all refactored.

src/openrct2/actions/BannerPlaceAction.h
Broxzier: Nevermind, I thought this was only used in the .cpp file.

IntelOrca (approved):

Issue #15367, opened 2021-09-07
Encode Track Type in Track Element


duncanspumpkin:
### Describe the feature

Currently the rendering of a track element is governed by its associated ride "`type`" but most of the community likes to mix and match across different ride types. This is currently achevied by merging of two different rides. This though is a waste of ride indexs and means for complex building. I propose that instead the track elements rendering is governed by its track type. This would be acheived by using up the free space in the track element to add a new field `TrackType` (eventually renamed to `TrackObjectType` if we ever manage to move track drawing into object files).

https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/world/TileElement.h#L378

At present this would only use one byte of space but if we ever want to move to objects for tracks we might need to allocate two bytes. 

This is obviously only possible on the NSF and also it would be gated behind some cheats. Would require a new track format as well for best compatability.

### Feature type

Quality of life

spacek531:
When track type objects are added, the remove_unused_object command will have to look at all track elements.

spacek531:
Some people have said to me they would love to be able to change the appearance of just the brakes on e.g. giga coaster or looping coaster, to mimic more modern coaster styles like Intamin LSM and Vekoma loopers. Having a simple one bit 'this appears like this ride type' field is elegant, but to me it seems like a waste to create a new track type just to change one element appearance.

frutiemax:
```
        TRACK_PAINT_FUNCTION_GETTER paintFunctionGetter = ride->GetRideTypeDescriptor().TrackPaintFunction;
        if (paintFunctionGetter != nullptr)
        {
            TRACK_PAINT_FUNCTION paintFunction = paintFunctionGetter(trackType);
            if (paintFunction != nullptr)
            {
                paintFunction(session, ride, trackSequence, direction, height, trackElement);
            }
        }
```

If I understand, you think it would be better to first get the trackElement, then based on the new TrackType field, you would decide if you draw a giga track or a corkscrew track for example? I think this would multiply the number of TrackElementDescriptors as they are constants.

spacek531:
Aren't RTDs available by ride type with `GetRideTypeDescriptor(rideType)`? If so, then it can just get the paint function there.

frutiemax:
Right now, it's a TrackType field that got added to the Ride struct. You need an additional RideType variable that would be not the same as the one in the ride struct to do it like this.

spacek531:
[...]

PR #15414, opened 2021-09-14
#15367: Encode RideType in TrackElement

frutiemax:


src/openrct2/world/TileElement.h
```diff
@@ -380,13 +380,18 @@ struct TrackElement : TileElementBase
     ride_id_t RideIndex;
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunused-private-field"
-    uint8_t pad[2];
+    // uint8_t pad[2];
+    uint8_t RideType;
```
Gymnasiast, line 384: Put this above the pragma - it’s no longer an unused field. Only leave the `uint8_t pad;` in it.

Gymnasiast:
I think we should be looking at introducing `ride_type_t` instead of using a plain `uint8_t`, to make later refactors easier, if something like it doesn’t exist already.

IntelOrca:
If the plan is for ride type to be another object type, then it will be an `ObjectEntryIndex`, which is two bytes.

Gymnasiast:
That is certainly the plan _eventually_, I wonder if it makes sense to already make it into an ObjectEntryIndex, or just create a `ride_type_t` that is a `uint16_t` until rides are actually made into objects.

Gymnasiast:
Let’s do the latter, changing it into ObjectEntryIndex later on is trivial, while that doesn’t apply the other round.

ZehMatt:
> That is certainly the plan _eventually_, I wonder if it makes sense to already make it into an ObjectEntryIndex, or just create a `ride_type_t` that is a `uint16_t` until rides are actually made into objects.

I started with something like that here https://github.com/OpenRCT2/OpenRCT2/pull/15320

Gymnasiast (approved):
Would appreciate a review from another team member.

src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1928,28 +1928,31 @@ static void window_tile_inspector_paint(rct_window* w, rct_drawpixelinfo* dpi)
...
                 {
                     ft = Formatter();
                     ride->FormatNameTo(ft);
                     DrawTextBasic(
-                        dpi, screenCoords + ScreenCoordsXY{ 0, 22 }, STR_TILE_INSPECTOR_TRACK_RIDE_NAME, ft, { COLOUR_WHITE });
+                        dpi, screenCoords + ScreenCoordsXY{ 0, 11 }, STR_TILE_INSPECTOR_TRACK_RIDE_NAME, ft, { COLOUR_WHITE });
                 }
+
+                // Ride type. Individual pieces may be of a different ride type from the ride it belongs to.
+                auto rtd = GetRideTypeDescriptor(trackElement->GetRideType());
```
frutiemax, line 1950: Use constant reference here.

src/openrct2/ride/Track.cpp
```diff
@@ -244,6 +244,7 @@ bool track_add_station_element(CoordsXYZD loc, ride_id_t rideIndex, int32_t flag
                     targetTrackType = TrackElemType::MiddleStation;
                 }
                 stationElement->AsTrack()->SetTrackType(targetTrackType);
+                stationElement->AsTrack()->SetRideType(ride->type);
```
duncanspumpkin, line 247: I'm not sure this would be what the user expects. Imagine this scenario: 

You build a station and change its track type (ride type). Then you decide you want to expand the station but with a different track type (ride type). This 'helpfully' resets the track type of the original station end to the rides default.

Gymnasiast:
@frutiemax Please never merge the develop branch into a PR, always rebase.

duncanspumpkin:
This is causing issues with the arbitrary ride type change cheat.

Gymnasiast:
@duncanspumpkin I’m working on a fix.

PR #15416, opened 2021-09-15
Remove unnecessary else blocks

Broxzier:
This PR removes unnecessary `else` bocks that are placed directly after an `if` block that ends with `return`, `continue` or `break`. This does not change the flow in any way, but cleans up the unnecessary lines and reduces the indention quite a lot in some cases.

### DIFF
--- a/src/openrct2-ui/WindowManager.cpp
+++ b/src/openrct2-ui/WindowManager.cpp
@@ -156,10 +156,7 @@ public:
                 {
                     return this->OpenWindow(WC_RESEARCH);
                 }
-                else
-                {
-                    return window_new_ride_open_research();
-                }
+                return window_new_ride_open_research();
             case WV_MAZE_CONSTRUCTION:
                 return window_maze_construction_open();
             case WV_NETWORK_PASSWORD:
--- a/src/openrct2-ui/windows/TopToolbar.cpp
+++ b/src/openrct2-ui/windows/TopToolbar.cpp
@@ -1706,7 +1706,8 @@ static void window_top_toolbar_scenery_tool_down(const ScreenCoordsXY& windowPos
         repaint_scenery_tool_down(windowPos, widgetIndex);
         return;
     }
-    else if (gWindowSceneryEyedropperEnabled)
+
+    if (gWindowSceneryEyedropperEnabled)
     {
         scenery_eyedropper_tool_down(windowPos, widgetIndex);
         return;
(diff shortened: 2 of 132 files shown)


src/openrct2-ui/WindowManager.cpp
```diff
@@ -156,10 +156,7 @@ class WindowManager final : public IWindowManager
                 {
                     return this->OpenWindow(WC_RESEARCH);
                 }
-                else
-                {
-                    return window_new_ride_open_research();
-                }
+                return window_new_ride_open_research();
```
Gymnasiast, line 159: I prefer a space after an if block. This is, of course, a matter of preference, so feel free to decide yourself.

Broxzier: It's pretty mixed here. I decided to add spaces when subsequence if-statements check for different things, while keeping it as is when they check the same values. In this case it's inside a switch-case where no other cases have if-statements or empty lines, so I went without adding one.

Gymnasiast:
Checked 40 out of 132 files so far. Looking good so far, it’s just quite a lot to check ;-)

src/openrct2-ui/windows/TopToolbar.cpp
```diff
@@ -3041,16 +3040,14 @@ static money64 selection_lower_land(uint8_t flags)
...
-        auto landLowerAction = LandLowerAction(
-            { centreX, centreY },
-            { gMapSelectPositionA.x, gMapSelectPositionA.y, gMapSelectPositionB.x, gMapSelectPositionB.y }, gMapSelectType);
-        auto res = (flags & GAME_COMMAND_FLAG_APPLY) ? GameActions::Execute(&landLowerAction)
-                                                     : GameActions::Query(&landLowerAction);
 
-        return res->Error == GameActions::Status::Ok ? res->Cost : MONEY64_UNDEFINED;
-    }
+    auto landLowerAction = LandLowerAction(
+        { centreX, centreY }, { gMapSelectPositionA.x, gMapSelectPositionA.y, gMapSelectPositionB.x, gMapSelectPositionB.y },
```
Gymnasiast, line 3045: (making a note for a possible refactor here, will remove later.)

Broxzier:
Merged using GitHub's editor, will rebase later.

src/openrct2/drawing/Drawing.String.cpp
```diff
@@ -546,163 +548,160 @@ static void ttf_draw_string_raw_ttf(rct_drawpixelinfo* dpi, std::string_view tex
...
-        uint8_t colour = info->palette[1];
-        TTFSurface* surface = ttf_surface_cache_get_or_add(fontDesc->font, text);
-        if (surface == nullptr)
-            return;
 
-        int32_t drawX = info->x + fontDesc->offset_x;
-        int32_t drawY = info->y + fontDesc->offset_y;
-        int32_t width = surface->w;
-        int32_t height = surface->h;
+    uint8_t colour = info->palette[1];
```
Gymnasiast, line 552: colour_t (note for another PR)

src/openrct2/platform/Linux.cpp
```diff
@@ -83,11 +83,11 @@ uint16_t platform_get_locale_language()
         {
             return LANGUAGE_ENGLISH_US;
         }
-        else if (!fnmatch(pattern, "zh_CN", 0))
+        if (!fnmatch(pattern, "zh_CN", 0))
```
Gymnasiast, line 86: (note)

src/openrct2/util/Util.cpp
```diff
@@ -351,17 +351,14 @@ int32_t strlogicalcmp(const char* s1, const char* s2)
...
+        if (!(isdigit(*s1) && isdigit(*s2)))
         {
             if (toupper(*s1) != toupper(*s2))
                 return toupper(*s1) - toupper(*s2);
-            else
-            {
-                ++s1;
-                ++s2;
-            }
+            ++s1;
```
Gymnasiast, line 360: Can you put a newline above? Makes it easier to read, especially since the if does not have braces.

Broxzier: Sure thing. Funnily I had added one here but removed it later for consistency. I'll also add them to line 372 then.

Gymnasiast (approved):

PR #15419, opened 2021-09-16
Fix sv6 export not adding all objects

ZehMatt:
Exporting and importing crashes because the objects are not loaded.

Regression from #15403.


Broxzier (approved):
Change makes sense. I cannot test it right now, but assuming you have tested this fix yourself already. Ignore Android build status, it's not related to this PR.

duncanspumpkin:
I used admin privileges to merge as this is game breaking without it.

PR #15426, opened 2021-09-17
Remove focus union and replace with typed focus

duncanspumpkin:
This if for the NSF to allow for CoordsXYZ.

Names are pretty much placeholders in this as Focus2 is a rubbish name.

### DIFF
--- a/src/openrct2/interface/Viewport.cpp
+++ b/src/openrct2/interface/Viewport.cpp
@@ -119,27 +119,26 @@ std::optional<ScreenCoordsXY> centre_2d_coordinates(const CoordsXYZ& loc, rct_vi
 
 CoordsXYZ Focus2::GetPos() const
 {
-    CoordsXYZ result;
-    std::visit(
-        [&result](auto&& arg) {
+    return std::visit(
+        [](auto&& arg) {
             using T = std::decay_t<decltype(arg)>;
             if constexpr (std::is_same_v<T, Focus2::CoordinateFocus>)
-                result = arg;
+                return arg;
             else if constexpr (std::is_same_v<T, Focus2::EntityFocus>)
             {
                 auto* centreEntity = GetEntity(arg);
                 if (centreEntity != nullptr)
                 {
-                    result = { centreEntity->x, centreEntity->y, centreEntity->z };
+                    return CoordsXYZ{ centreEntity->x, centreEntity->y, centreEntity->z };
                 }
                 else
                 {
                     log_error("Invalid entity for focus.");
+                    return CoordsXYZ{};
                 }
             }
         },
         data);
-    return result;
 }
 
 /**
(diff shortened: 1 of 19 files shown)


src/openrct2/interface/Viewport.cpp
```diff
@@ -166,24 +185,16 @@ void viewport_create(
...
+    CoordsXYZ centrePos = focus.GetPos();
+    std::visit(
+        [&w](auto&& arg) {
+            using T = std::decay_t<decltype(arg)>;
+            if constexpr (std::is_same_v<T, Focus2::CoordinateFocus>)
+                w->viewport_target_sprite = SPRITE_INDEX_NULL;
+            else if constexpr (std::is_same_v<T, Focus2::EntityFocus>)
+                w->viewport_target_sprite = arg;
+        },
+        focus.data);
```
ZehMatt, line 197:
```suggestion
    w->viewport_target_sprite = std::visit(
        [](auto&& arg) {
            using T = std::decay_t<decltype(arg)>;
            if constexpr (std::is_same_v<T, Focus2::CoordinateFocus>)
                return SPRITE_INDEX_NULL;
            else if constexpr (std::is_same_v<T, Focus2::EntityFocus>)
                return arg;
        },
        focus.data);
```

duncanspumpkin: Oh didn't know you could return from a visit like that.

src/openrct2/interface/Window_internal.cpp
```diff
@@ -12,28 +12,10 @@ void rct_window::SetLocation(const CoordsXYZ& coords)
...
+    if (viewport == nullptr || !focus2.has_value())
         return;
 
-    CoordsXYZ newCoords = {};
-    if (viewport_focus_sprite.type & VIEWPORT_FOCUS_TYPE_SPRITE)
-    {
-        auto* sprite = GetEntity(viewport_focus_sprite.sprite_id);
-        if (sprite == nullptr)
-        {
-            return;
```
ZehMatt, line 25: Does anyone actually clear the focus when the entity vanishes? GetPos introduced log_error I'm not sure if this would continue to spam now if for reasons the entity vanishes.

duncanspumpkin: Yeah the smart functions keep the focus entity updated.

src/openrct2/interface/Viewport.cpp
```diff
@@ -117,6 +117,31 @@ std::optional<ScreenCoordsXY> centre_2d_coordinates(const CoordsXYZ& loc, rct_vi
     return { screenCoord };
 }
 
+CoordsXYZ Focus2::GetPos() const
+{
+    CoordsXYZ result;
+    std::visit(
```
ZehMatt, line 123:
```suggestion
    CoordsXYZ result = std::visit(
```

ZehMatt (approved):

Gymnasiast:
Is it possible that this PR is the cause of https://github.com/OpenRCT2/OpenRCT2/issues/15439 ?

duncanspumpkin:
Hmm possible. I'll investigate tomorrow. I completely forgot to rename the struct from its temporary name anyway so was going to be doing a subsequent pr

PR #15429, opened 2021-09-17
Remove scenario ticks and adjust export/import

ZehMatt:
Another NSF backport, this one might break replays.

### DIFF
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -225,7 +225,7 @@ public:
 
         gDateMonthsElapsed = static_cast<int32_t>(_s6.elapsed_months);
         gDateMonthTicks = _s6.current_day;
-        gScenarioTicks = _s6.scenario_ticks;
+        gCurrentTicks = _s6.game_ticks_1;
 
         scenario_rand_seed(_s6.scenario_srand_0, _s6.scenario_srand_1);
 
@@ -401,7 +401,6 @@ public:
             String::Set(gScenarioFileName, sizeof(gScenarioFileName), _s6.scenario_filename);
         }
         std::memcpy(gScenarioExpansionPacks, _s6.saved_expansion_pack_names, sizeof(_s6.saved_expansion_pack_names));
-        gCurrentTicks = _s6.game_ticks_1;
         gCurrentRealTimeTicks = 0;
 
         ImportRides();
@@ -480,6 +479,7 @@ public:
 
         research_determine_first_of_type();
     }
+
     void FixLandOwnership() const
     {
         if (String::Equals(_s6.scenario_filename, "Europe - European Cultural Festival.SC6"))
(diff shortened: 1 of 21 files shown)


src/openrct2/rct2/S6Importer.cpp
```diff
@@ -480,6 +479,11 @@ class S6Importer final : public IParkImporter
 
         research_determine_first_of_type();
     }
+
+    void ConvertScenarioTicks()
```
ZehMatt, line 483: Stray change

duncanspumpkin (approved):
All looks fine to me. Await @Gymnasiast  review prior to merge.

Gymnasiast:
Would like to test a few things, e.g. stuff like https://github.com/OpenRCT2/OpenRCT2/issues/14734.

ZehMatt:
> Would like to test a few things, e.g. stuff like #14734.

There are three things to watch out for with this PR
1. Guest time in park
2. Staff hire time
3. Litter creation time, this influences the park rating and is somewhat hard to test I guess.

But given the guest and staff times seemed fine at my end I guess litter will be also.

Gymnasiast:
I don’t see code to properly import litter creation time in S6Importer, though. Is that intentional?

ZehMatt:
> I don’t see code to properly import litter creation time in S6Importer, though. Is that intentional?

Nope it seems I forgot that one even tho I know I must do it, nice catch.

Gymnasiast (changes requested):
Does not work correctly. The mechanics in this save all have a hire date of June, yr 528 instead of March, yr 1.

[Pakhuis - Derde Alter Ego - fixedMR.sv6.zip](https://github.com/OpenRCT2/OpenRCT2/files/7190657/Pakhuis.-.Derde.Alter.Ego.-.fixedMR.sv6.zip)

ZehMatt:
HireDate is used as an absolute time for  {MONTHYEAR} unlike guests which use a relative time, it should be good now.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1727,6 +1727,15 @@ template<> void S6Importer::ImportEntity<Vehicle>(const RCT12SpriteBase& baseSrc
     dst->IsCrashedVehicle = src->flags & RCT12_SPRITE_FLAGS_IS_CRASHED_VEHICLE_SPRITE;
 }
 
+static uint32_t AdjustScenarioToCurrentTicks(const rct_s6_data& s6, uint32_t tick)
+{
+    // Previously gScenarioTicks was used as a time point now its gCurrentTicks.
```
Gymnasiast, line 1732:
```suggestion
    // Previously gScenarioTicks was used as a time point, now it’s gCurrentTicks.
```

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1727,6 +1727,15 @@ template<> void S6Importer::ImportEntity<Vehicle>(const RCT12SpriteBase& baseSrc
...
+    // gCurrentTicks and gScenarioTicks are now exported as the same, older saves that have a different
+    // scenario tick must add account for the distance between the two.
```
Gymnasiast, line 1734:
```suggestion
    // scenario tick must account for the difference between the two.
```

Gymnasiast (approved):
Still a bit apprehensive (if this breaks, it will be in subtle and hard-to-detect ways), but I haven’t found any more bugs.

Issue #14877, opened 2021-06-10
Staff Patrol Areas Carry Over Between Scenarios/Saves


Derpercoaster:
OS: Windows 10
Version: v0.3.3 (new save format)
Build: 16c10f6

Issue: Patrol areas of staff seem to carry over between scenarios. It seems to happen if the following criteria is met, in order:
1. Both scenarios have staff.
2. (At least) The staff in the first scenario have a patrol area assigned.

Reproduceable in vanilla: No
Specific to multiplayer: No

To reproduce:
1. Simply open a scenario. It can be a brand new one or an existing save.
2. Have staff. Place them if there are already any, and give them a patrol area.
3. Open a different save that already has staff in it (they do not need patrol areas when you open the save). If done corectly, they will have patrol areas at the same coordinates on the map as the staff in the previous scenario.

Issue #14824, opened 2021-06-05
[New Save - Public Beta Build 2] Clicking in the general overview of Staff Patrol Area brings you to the wrong staff member


karst:
**OS:** Win10 64bit
**Version:** 0.3.3 New Save Build
**Commit/Build:** 425f968

When I am clicking the area effected with the select tool it grabs a different area than I am selecting.

- [Unknown] Reproducible in RCT2 (vanilla)?
- [Unknown] Specific to multiplayer?

**Steps to reproduce:**
1. Go to staff
2. Go click on patrol areas
3. Click a patrol area from a staff member when you have several laid out.
4. It will bring you to the wrong staff member

**Screenshots / Video:**
https://www.youtube.com/watch?v=Q5sgzDy5bIQ

**Save game:**
https://drive.google.com/file/d/15q5l0ZVEbjEG2-Lq07kijmKfKopiWrS_/view?usp=sharing

karst:
Confirmed for snapshot 3 of NSF (4f6af6f).

PR #15436, opened 2021-09-19
Small Staff Refactor from NSF

duncanspumpkin:
Just trying to make the actual change that modified the patrol areas easier to follow. So to start I've introduced all of the new methods added in the NSF then after this can modify the functionality. Also noticed that `peep->ClearPatrolArea();` is not network safe. 
I think we will need to modify StaffSetPatrolAreaAction to have three states: Set, Unset, Clear. That will be a follow on PR.

### DIFF
--- a/src/openrct2-ui/windows/Staff.cpp
+++ b/src/openrct2-ui/windows/Staff.cpp
@@ -530,7 +530,7 @@ void window_staff_overview_mousedown(rct_window* w, rct_widgetindex widgetIndex,
             }
 
             // Disable clear patrol area if no area is set.
-            if (gStaffModes[peep->StaffId] != StaffMode::Patrol)
+            if (!peep->HasPatrolArea())
             {
                 Dropdown::SetDisabled(1, true);
             }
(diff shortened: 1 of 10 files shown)


pizza2004:
Would this make it easier to implement #681?

duncanspumpkin:
No it hasnt.

duncanspumpkin:
I found a bug whilst performing this refactor. The merged staff patrol areas was not updated when you fired staff this meant that it would display the wrong area until you hired a new staff member which has the same allocated staff id as the one that was fired. Its really easy to test just set a patrol area then delete the staff. Load up the staff list and show patrol areas.

Gymnasiast:
There are a few issues in our bug tracker related to staff and the NSF. E.g.:

- https://github.com/OpenRCT2/OpenRCT2/issues/14877
- https://github.com/OpenRCT2/OpenRCT2/issues/14824

We need to check if this refactor has not brought these over by accident.

duncanspumpkin:
2nd one definitely not related. 1st one seriously doubt it either. They are both to do with the second phase of this.

Gymnasiast (commented):
Code looks good, want to take it for a test.

Gymnasiast (approved):
Testing revealed no bugs.

PR #15437, opened 2021-09-19
Use min/max water height defines in mapgen (from NSF)

Gymnasiast:
Some explanation can be found here: https://github.com/OpenRCT2/OpenRCT2/pull/10664/files#r691862953

Do not squash merge.


Broxzier:
Duplicate of https://github.com/OpenRCT2/OpenRCT2/pull/15435

PR #15444, opened 2021-09-20
Fix #15439. Ride viewport not updating on size change.

### LINKED ISSUE #15439: Ride window viewport is partially grey
### Operating System

Ubuntu 20.04

### OpenRCT2 build

v0.3.4.1+ - d3f7660a46453af38c255c01d58d14edf94f1ee2

### Describe the issue

When clicking a ride, the viewport initially shows as partially grey. When selecting another view, this goes away.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Open a park with rides, e.g. Electric Fields
2. Click any ride, for example the station of the Boat Hire.
3. Observe part of the viewport is missing

### Attachments

afbeelding

duncanspumpkin:
The ride viewport does not work the same as other viewports due to the selection of views that it has. After refactoring the focus system to use a more streamlined approach the ride viewport lost its invalidation on resizing. If the ride window was to use the same viewport update code as say the guest window then the viewport focus still ends up incorrect due to it no longer centering the focus. Therefore the best approach was to lose the focus on resize and force a recalculation of it.

Also renamed Focus2 to Focus as Focus2 was meant to just be fill in whilst removing the original focus structs

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -1796,7 +1796,7 @@ static void window_ride_main_resize(rct_window* w)
 
     w->flags |= WF_RESIZABLE;
     window_set_resize(w, 316, minHeight, 500, 450);
-    // Unlike other windows the focus needs to be recentered so its best to just reset it.
+    // Unlike with other windows, the focus needs to be recentred so it’s best to just reset it.
     w->focus = std::nullopt;
     window_ride_init_viewport(w);
 }
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -1589,7 +1589,7 @@ static void window_ride_init_viewport(rct_window* w)
 
     int32_t viewSelectionIndex = w->ride.view - 1;
 
-    std::optional<Focus2> focus;
+    std::optional<Focus> focus;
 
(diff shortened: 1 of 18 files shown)


src/openrct2-ui/windows/Ride.cpp
```diff
@@ -1796,6 +1796,8 @@ static void window_ride_main_resize(rct_window* w)
 
     w->flags |= WF_RESIZABLE;
     window_set_resize(w, 316, minHeight, 500, 450);
+    // Unlike other windows the focus needs to be recentered so its best to just reset it.
```
Gymnasiast, line 1799:
```suggestion
    // Unlike with other windows, the focus needs to be recentred so it’s best to just reset it.
```

src/openrct2-ui/windows/Sign.cpp
```diff
@@ -127,7 +127,7 @@ class SignWindow final : public Window
 
         // Create viewport
         rct_widget& viewportWidget = window_sign_widgets[WIDX_VIEWPORT];
-        const auto focus = Focus2(CoordsXYZ{ signViewPosition, viewZ });
+        const auto focus = Focus(CoordsXYZ{ signViewPosition, viewZ });
```
Broxzier, line 130: This one and other locals named `focus` need updating, as they now hide the class member with the same name.

Gymnasiast (approved):

PR #15446, opened 2021-09-20
Stream changes from NSF

duncanspumpkin:
All of the changes that were made to streams as part of the NSF are contained in this PR. I've also added the OrcaStream which is standalone from the NSF.

### DIFF
--- a/src/openrct2/core/IStream.cpp
+++ b/src/openrct2/core/IStream.cpp
@@ -12,9 +12,6 @@
 #include "../object/Object.h"
 #include "Memory.hpp"
 #include "String.hpp"
-
-#include <vector>
-
 namespace OpenRCT2
 {
     utf8* IStream::ReadString()
@@ -55,6 +52,17 @@ namespace OpenRCT2
             size_t numBytes = String::SizeOf(str) + 1;
             Write(str, numBytes);
         }
+    }
+
+    void IStream::WriteString(std::string_view str)
+    {
+        for (auto c : str)
+        {
+            if (c == '\0')
+                break;
+            WriteValue<uint8_t>(c);
+        }
+        WriteValue<uint8_t>(0);
     }
 
     void IStream::WriteString(const std::string& str)
--- a/src/openrct2/core/IStream.cpp
+++ b/src/openrct2/core/IStream.cpp
@@ -12,6 +12,9 @@
 #include "../object/Object.h"
 #include "Memory.hpp"
 #include "String.hpp"
+
+#include <vector>
+
 namespace OpenRCT2
 {
     utf8* IStream::ReadString()
@@ -54,9 +57,9 @@ namespace OpenRCT2
         }
     }
 
-    void IStream::WriteString(std::string_view str)
+    void IStream::WriteString(const std::string_view str)
     {
-        for (auto c : str)
+        for (const auto c : str)
         {
             if (c == '\0')
                 break;
--- a/src/openrct2/core/OrcaStream.hpp
+++ b/src/openrct2/core/OrcaStream.hpp
@@ -506,7 +506,7 @@ namespace OpenRCT2
             {
                 if constexpr (sizeof(T) > 4)
                 {
-                    if (std::is_signed<T>())
+                    if constexpr (std::is_signed<T>())
                     {
                         int64_t raw{};
                         Read(&raw, sizeof(raw));
@@ -578,9 +578,9 @@ namespace OpenRCT2
                 buffer.reserve(64);
                 while (true)
                 {
-                    char c;
+                    char c{};
                     ReadBuffer(&c, sizeof(c));
-                    if (c == 0)
+                    if (c == '\0')
                     {
                         break;
                     }
--- a/src/openrct2/core/OrcaStream.hpp
+++ b/src/openrct2/core/OrcaStream.hpp
@@ -69,7 +69,7 @@ namespace OpenRCT2
         ChunkEntry _currentChunk;
 
     public:
-        OrcaStream(IStream& stream, Mode mode)
+        OrcaStream(IStream& stream, const Mode mode)
         {
             _stream = &stream;
             _mode = mode;
@@ -121,10 +121,7 @@ namespace OpenRCT2
 
         ~OrcaStream()
         {
-            if (_mode == Mode::READING)
-            {
-            }
-            else
+            if (_mode == Mode::WRITING)
             {
                 const void* uncompressedData = _buffer.GetData();
                 const uint64_t uncompressedSize = _buffer.GetLength();
@@ -184,7 +181,7 @@ namespace OpenRCT2
             return _header;
         }
 
-        template<typename TFunc> bool ReadWriteChunk(uint32_t chunkId, TFunc f)
+        template<typename TFunc> bool ReadWriteChunk(const uint32_t chunkId, TFunc f)
         {
             if (_mode == Mode::READING)
             {
@@ -209,12 +206,12 @@ namespace OpenRCT2
         }
 
     private:
-        bool SeekChunk(uint32_t id)
+        bool SeekChunk(const uint32_t id)
         {
-            auto result = std::find_if(_chunks.begin(), _chunks.end(), [id](const ChunkEntry& e) { return e.Id == id; });
+            const auto result = std::find_if(_chunks.begin(), _chunks.end(), [id](const ChunkEntry& e) { return e.Id == id; });
             if (result != _chunks.end())
             {
-                auto offset = result->Offset;
+                const auto offset = result->Offset;
                 _buffer.SetPosition(offset);
                 return true;
             }
@@ -238,7 +235,7 @@ namespace OpenRCT2
             std::stack<ArrayState> _arrayStack;
 
         public:
-            ChunkStream(MemoryStream& buffer, Mode mode)
+            ChunkStream(MemoryStream& buffer, const Mode mode)
                 : _buffer(buffer)
                 , _mode(mode)
             {
@@ -254,7 +251,7 @@ namespace OpenRCT2
                 return _buffer;
             }
 
-            void ReadWrite(void* addr, size_t len)
+            void ReadWrite(void* addr, const size_t len)
             {
                 if (_mode == Mode::READING)
                 {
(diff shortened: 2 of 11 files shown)


duncanspumpkin:
Hmm, gzip and ungzip for non network are causing a crash

duncanspumpkin:
Need to port over util changes to get the gzip/ungzip

src/openrct2/core/IStream.cpp
```diff
@@ -12,9 +12,6 @@
 #include "../object/Object.h"
 #include "Memory.hpp"
 #include "String.hpp"
-
-#include <vector>
```
Gymnasiast, line 16: This file does use vector (e.g. on line 22/19), so I don’t think this should be removed.

duncanspumpkin: Yeah but it's included in the associated header IStream.hpp

Gymnasiast: Shouldn’t files carry their own dependencies, rather than relying on other files to do it for them?

duncanspumpkin: I would say so expect when there is a directly associated header. This is the implementation of the istream header and therefore i don't really see the point in having it in both it's just increasing compile times for no reason.

Broxzier: The include in the .hpp file seems to be unused. It'd be better to move it to the source file.

duncanspumpkin: It's used on L236 and afaik you aren't meant to forward declare std classes

Broxzier: Ah yup, you're right. I'd still argue that the #include in the source file is nice to have, since std::vector gets used as an implementation detail outside of function declarations and data members. I don't expect the extra inclusion of system headers to have a real impact on build time.

ZehMatt: Its best to always include the headers you are using, if you rely on the include in another header and that gets refactored you need to change unrelated things.

duncanspumpkin:
I feel there is a couple of cases where more const could be applied across the orca stream and util.

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,685 @@
...
+            _currentChunk.Length = 0;
+            ChunkStream stream(_buffer, _mode);
+            f(stream);
+            _currentChunk.Length = static_cast<uint64_t>(_buffer.GetPosition()) - _currentChunk.Offset;
+            _chunks.push_back(_currentChunk);
+            return true;
+        }
+
+    private:
+        bool SeekChunk(uint32_t id)
```
Broxzier, line 212: Can probably be made const.

```suggestion
        bool SeekChunk(uint32_t id) const
```

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,685 @@
...
+                        return static_cast<T>(raw);
+                    }
+
+                    uint64_t raw{};
+                    Read(&raw, sizeof(raw));
+                    return static_cast<T>(raw);
+                }
+                else
+                {
+                    if constexpr (std::is_signed<T>())
```
Broxzier, line 525: Can be combined to reduce indention by one level.

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,685 @@
...
+                    }
+                    else
+                    {
+                        auto raw = static_cast<uint64_t>(value);
+                        Write(&raw, sizeof(raw));
+                    }
+                }
+                else
+                {
+                    if (std::is_signed<T>())
```
Broxzier, line 565: Same here

Broxzier, line 565:
```suggestion
                    if constexpr (std::is_signed<T>())
```

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,685 @@
...
+                return true;
+            }
+
+            void EndArray()
+            {
+                auto& arrayState = _arrayStack.top();
+                if (_mode == Mode::READING)
+                {
+                }
+                else
```
Broxzier, line 669:
```suggestion
                if (_mode != Mode::READING)
```

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,685 @@
...
+        }
+
+        OrcaStream(const OrcaStream&) = delete;
+
+        ~OrcaStream()
+        {
+            if (_mode == Mode::READING)
+            {
+            }
+            else
```
Broxzier, line 127:
```suggestion
            if (_mode != Mode::READING)
```

Broxzier: Or add `throw std::runtime_error("Incorrect mode");` if you want to error.

ZehMatt: I don't think that a container should typically have such complexity in a destructor other than releasing resources. The write should be made explicit, this container can not be used in parallel properly for this reason.

src/openrct2/core/OrcaStream.hpp
ZehMatt: It does mutate the container, I don't think const applies here.

Broxzier: Right, it modifies `_buffer`. Wasn't quite sure when looking this over the first time, hence 'probably'.

src/openrct2/util/Util.cpp
```diff
@@ -693,6 +693,96 @@ bool util_gzip_compress(FILE* source, FILE* dest)
...
+            if (ret == Z_STREAM_ERROR)
+            {
+                throw std::runtime_error("deflate failed with error " + std::to_string(ret));
+            }
+            output.resize(output.size() - strm.avail_out);
+        } while (strm.avail_out == 0);
+
+        src += nextBlockSize;
+    } while (flush != Z_FINISH);
+    deflateEnd(&strm);
```
duncanspumpkin, line 782: should be inflateEnd

src/openrct2/core/OrcaStream.hpp
duncanspumpkin: You mean combine the >4 and the signed check? That would lead to harder to follow code.

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,679 @@
...
+            void WriteString(const std::string_view s)
+            {
+                const char nullt = '\0';
+                auto len = s.find('\0');
+                if (len == std::string_view::npos)
+                {
+                    len = s.size();
+                }
+                _buffer.Write(s.data(), len);
+                _buffer.Write(&nullt, sizeof(nullt));
```
ZehMatt, line 602:
```suggestion
                const char nullt = '\0';
                auto len = s.size();
                while (len > 0 && s[len - 1] == nullt)
                {
                    len--;
                }
```
I think in 99.9% of the time this will simply skip 1 byte, not sure if this is cleaner to read but we don't have to search for null terminators at the beginning.

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,679 @@
...
+            }
+
+            void WriteString(const std::string_view s)
+            {
+                const char nullt = '\0';
+                auto len = s.find('\0');
+                if (len == std::string_view::npos)
+                {
+                    len = s.size();
+                }
```
ZehMatt, line 600:
```suggestion
                const char nullt = '\0';
                auto len = s.size();
                while (len > 0 && s[len - 1] == nullt)
                {
                    len--;
                }
```
I think in 99.9% of the time this will simply skip 1 byte, not sure if this is cleaner to read but we don't have to search for null terminators at the beginning.

src/openrct2/core/OrcaStream.hpp
```diff
@@ -0,0 +1,679 @@
...
+            }
+
+            std::string ReadString()
+            {
+                std::string buffer;
+                buffer.reserve(64);
+                while (true)
+                {
+                    char c;
```
ZehMatt, line 581:
```suggestion
                    char c{};
```

src/openrct2/core/OrcaStream.hpp
Broxzier: No, I mean instead of `else { if (...) { ... } }` you can do `else if (...) { ... }`.

src/openrct2/core/OrcaStream.hpp
IntelOrca: If you ever pass a `std::string_value` containing null terminators (which could even be the first byte)... it will break reading the string when opening the file. It has to ensure no more than one null terminator byte is written.

ZehMatt: I don't think we have such strings, if we do it would be perhaps best to encode the size as well and not treat them as strings

Gymnasiast: I’m a bit reluctant to just go on an _assumption_ we don’t have such strings. Much prefer having more robust code.

ZehMatt: I also prefer robust code, thats why I think meddling with null termination on string_views is a terrible idea, string_views give you zero guarantees about null termination. The point of string_view is to use the size provided and not rely on termination.

IntelOrca: The `string_view` is just an overload. You also have `char*` and `std::string` overloads. Regardless of the overload, they all need to comply with the file format.

ZehMatt: Well my point is somewhat that you should not pass data with string_view/string/char* or by any other means that is not an actual string or includes multiple null terminators, this only means the API is used incorrectly. Anyway its just a nitpick and would reduce the overall lookup time when it comes to string encoding.

src/openrct2/core/OrcaStream.hpp
duncanspumpkin: yeah thats what i mean it would be harder to follow

PR #15449, opened 2021-09-20
Fix #15442: Use after free in object manager

### LINKED ISSUE #15442: [NSF] Crashing when selecting Scenario Editor
### Operating System

Windows 64 bits

### OpenRCT2 build

New save format (75d1b9402cec5a45afa7bdbcd51fc0e065367fb2)

### Describe the issue

Crash when selecting Scenario Editor from start screen.

### Area(s) with issue?

New Save Format (NSF)

### Steps to reproduce

Select scenario editor from start screen.

### Attachments

_No response_

ZehMatt:
This fixes #15442 but its also in develop, the lines changed in NSF a bit.

### DIFF
--- a/src/openrct2/Editor.cpp
+++ b/src/openrct2/Editor.cpp
@@ -63,15 +63,18 @@ namespace Editor
 
     static void object_list_load()
     {
+        auto* context = GetContext();
+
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Tested it on both develop and NSF and it seems to solve the crash.

PR #15458, opened 2021-09-23
Check for nullptr explicitly

Broxzier:


src/openrct2/network/NetworkBase.cpp
```diff
@@ -3701,7 +3701,7 @@ GameActions::Result::Ptr network_kick_player(NetworkPlayerId_t playerId, bool is
         return std::make_unique<GameActions::Result>(GameActions::Status::Unknown, STR_NONE);
     }
 
-    if (player && player->Flags & NETWORK_PLAYER_FLAG_ISSERVER)
+    if (player->Flags & NETWORK_PLAYER_FLAG_ISSERVER)
```
Broxzier, line 3704: Note: This one could simply be removed since `player` is already checked for `nullptr` in the `if`-statement before this one, which then exits the function.

tupaschoal (approved):
Lgtm

src/openrct2/network/NetworkBase.cpp
Gymnasiast: Let’s remove it then.

Gymnasiast (approved):

Broxzier:
This is a draft because I plan on doing the same for the openrct2-ui project.

duncanspumpkin:
Just make another pr for that then. This is already reviewed no point rereviewing

PR #8427, opened 2018-12-13
Fix #8385: Ghost objects rendered in minimap

Mikroscopic:
Add a check for the ghost flag before changing map pixel color (First PR, please alert me to any mistakes)

PR #15462, opened 2021-09-25
Avoid drawing over elements that are higher

Broxzier:
Ghost elements would always draw on top of everything in the map window, even when built under other elements.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -26,6 +26,7 @@
 - Fix: [#15257] Chat icon shows in scenario/track editor. Other icons don't disable when deactivated in options menu.
 - Fix: [#15289] Unexpected behavior with duplicated banners which also caused desyncs in multiplayer.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
+- Improved: [#15462] Ghost elements do no longer overrule higher elements' colours in the map window.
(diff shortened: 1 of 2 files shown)


Broxzier:
Before and after:
Diamond Heights 2021-09-25 16-05-49 Diamond Heights 2021-09-25 16-05-12

IntelOrca:
Isn't that intentional?

Broxzier:
@IntelOrca Nothing about it is mentioned in the original implementation PR: https://github.com/OpenRCT2/OpenRCT2/pull/8427

With this change, ghost pieces are still visible in the map window when beneath something, they just don't overrule the entire tile, so I don't see any advantage of leaving it as it currently is.

duncanspumpkin:
I don't think it should really be changed. It's something you would be looking for on the map

Gymnasiast:
I have to agree with Duncan here, I don’t see much advantage to changing it, while it makes it much harder to see on the map, especially in dense areas.

Broxzier:
I guess I'm in the minority, then. I just found this odd when playing and considered it more a bug than a feature.

IntelOrca:
I don't mind this change, fine by me.

Gymnasiast:
Great, a stalemate then...

I’ll be the scapegoat then, and break it. Hope you don’t hate me too much for this.

PR #15463, opened 2021-09-25
Make WIDGETS_END constexpr

Broxzier:


Gymnasiast:
Perhaps we should rename it to `WidgetsEnd` to match `MakeWidget`?

Broxzier:
@Gymnasiast The difference is that `MakeWidget` is a constexpr function and this is a constexpr variable. All other constexpr variables are all caps, and I don't see the advantage of turning this into a function.

Gymnasiast:
> All other constexpr variables are all caps, and I don't see the advantage of turning this into a function.

I don’t think that’s true, there are lots of constexpr variables in our codebase that are not all caps.

Anyway, it was a suggestion and not necessarily a “change request”. I’ll review the PR as-is.

Gymnasiast (approved):

Broxzier:
Yes, correction: all others that I checked when renaming this. As you can see, I ended up renaming it back to be in line with the majority.

duncanspumpkin:
At some point we should refactor all this to use a vector and remove the widget end then the name of it won't matter

PR #15464, opened 2021-09-26
Fix arbitrary ride type cheat not working

Gymnasiast:
Bug introduced with https://github.com/OpenRCT2/OpenRCT2/pull/15414.

For the NSF, we should revisit this. It’s very well possible that people will want to change the "base" ride type while keeping the actual pieces the same.


duncanspumpkin:
Yeah I'm of the opinion that we should not do this in nsf but do this on develop

PR #15470, opened 2021-09-27
Fix #15213: Freeze when reading a partially cut unicode character 

### LINKED ISSUE #15213: Freeze when hovering over Reverse Freefall Coaster in Russian
### Operating System

Ubuntu 20.04

### OpenRCT2 build

v0.3.4.1, d6ad9b67e7a6a716dc7ffd705d26ac191f76397d

### Describe the issue

Originally reported here: https://forums.openrct2.org/topic/5642-freeze-crash-when-i-move-mouse-under-this-object/?tab=comments#comment-26170

The Cyrillic alphabet takes up twice the space per letter in UTF-8. This makes it likely it overruns a buffer somewhere.

### Area(s) with issue?

_No response_

### Steps to reproduce

1.  Load up a park that has the Reverse Freefall Coaster loaded. E.g. Six Flags Magic Mountain
2. Open the New Ride window
3. Hover over the Reverse Freefall Coaster - it will freeze.

Broxzier:
This is the main change:
```diff
-    utf8 buffer[512];
-    format_string(buffer, sizeof(buffer), format, args);
-
-    StaticLayout layout(buffer, textPaint, width);
+    // TODO: Refactor StaticLayout to take a std::string_view instead. It shouldn't have to write to the buffer.
+    const std::string buffer = format_string(format, args);
+    StaticLayout layout(const_cast<char*>(buffer.c_str()), textPaint, width);
```
The main problem was that it is possible for unicode characters that take up multiple bytes to only get partly written to the buffer. The codepoint iterator then breaks on this.

The `const_cast` is safe, since the `std::string` cannot be optimized away to static data, thought it should really be refactored out. I've made a start with this, but it turns into a way bigger refactor than what is needed to solve this crash.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -22,6 +22,7 @@
 - Fix: [#15184] Crash when hovering over water types in Object Selection.
 - Fix: [#15193] Crash when rides/stalls are demolished.
 - Fix: [#15199] Construction window is not closed when a ride gets demolished.
+- Fix: [#15213] Freeze when hovering over Reverse Freefall Coaster in Russian.
 - Fix: [#15255] Tile Inspector shows banner information on walls that do not contain one.
 - Fix: [#15257] Chat icon shows in scenario/track editor. Other icons don't disable when deactivated in options menu.
 - Fix: [#15289] Unexpected behavior with duplicated banners which also caused desyncs in multiplayer.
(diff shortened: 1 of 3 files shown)


Gymnasiast:
Nice, I want to take it for a quick spin.

Gymnasiast (approved):

PR #15471, opened 2021-09-27
Compile with C++20 standard

ZehMatt:
I'm curious to see how far our CI supports it and switching to C++20 reveals a ton of new warnings/errors since its even more strict. I might use this PR as simply fixing some of the things and omit the first commit which switches the standard.


ZehMatt:
Since C++20 the u8 string literal now uses char8_t which does not mix with char, ideally we have to refactor std::string to std::u8string, std::string_view to std::u8string_view and all const char* to const char8_t*. We do have a typedef utf8 which should also become char8_t. The filesystem API also deprecated string_view/string and expects u8string_view/u8string.

PR #15473, opened 2021-09-27
Use int32_t for sprite_paint_setup

ZehMatt:
NSF Backport.

Two more files less

### DIFF
--- a/src/openrct2/paint/sprite/Paint.Sprite.cpp
+++ b/src/openrct2/paint/sprite/Paint.Sprite.cpp
@@ -32,13 +32,8 @@
  * Paint Quadrant
  *  rct2: 0x0069E8B0
  */
-void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t y)
+void sprite_paint_setup(paint_session* session, const CoordsXY& pos)
 {
-    if ((x & 0xe000) | (y & 0xe000))
-    {
-        return;
-    }
-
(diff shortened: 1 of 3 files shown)


src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -32,13 +32,8 @@
  * Paint Quadrant
  *  rct2: 0x0069E8B0
  */
-void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t y)
+void sprite_paint_setup(paint_session* session, int32_t x, int32_t y)
```
duncanspumpkin, line 35: Use a `const CoordsXY&`

duncanspumpkin (approved):

src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -32,13 +32,8 @@
  * Paint Quadrant
  *  rct2: 0x0069E8B0
  */
-void sprite_paint_setup(paint_session* session, const uint16_t x, const uint16_t y)
+void sprite_paint_setup(paint_session* session, const CoordsXY& pos)
 {
-    if ((x & 0xe000) | (y & 0xe000))
```
duncanspumpkin, line 37: Thinking about this some more and this should have been replaced with a 
```
    if (!map_is_location_valid(pos))
    {
        return;
    }```
As the setup function is expected to give invalid tiles when getting near the edges which are not required to be processed.

PR #15474, opened 2021-09-27
Fix overflow in GetSpatialIndexOffset

ZehMatt:
(999 * 32) / 32 does not fit uint8_t.

Only relevant for NSF but no downside having it in develop.

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -48,21 +48,18 @@ static void FreeEntity(EntityBase& entity);
 
 constexpr size_t GetSpatialIndexOffset(int32_t x, int32_t y)
 {
-    size_t index = SPATIAL_INDEX_LOCATION_NULL;
-    if (x != LOCATION_NULL)
-    {
-        x = std::clamp(x, 0, 0xFFFF);
-        y = std::clamp(y, 0, 0xFFFF);
+    if (x == LOCATION_NULL)
+        return SPATIAL_INDEX_LOCATION_NULL;
 
-        int16_t flooredX = floor2(x, 32);
-        uint8_t tileY = y >> 5;
-        index = (flooredX << 3) | tileY;
-    }
+    const auto tileX = std::clamp<size_t>(x / COORDS_XY_STEP, 0, MAXIMUM_MAP_SIZE_TECHNICAL);
+    const auto tileY = std::clamp<size_t>(y / COORDS_XY_STEP, 0, MAXIMUM_MAP_SIZE_TECHNICAL);
+    const auto index = tileX * MAXIMUM_MAP_SIZE_TECHNICAL + tileY;
 
     if (index >= sizeof(gSpriteSpatialIndex))
     {
         return SPATIAL_INDEX_LOCATION_NULL;
     }
+
     return index;
 }
 


src/openrct2/world/Sprite.cpp
```diff
@@ -54,8 +54,8 @@ constexpr size_t GetSpatialIndexOffset(int32_t x, int32_t y)
         x = std::clamp(x, 0, 0xFFFF);
         y = std::clamp(y, 0, 0xFFFF);
 
-        int16_t flooredX = floor2(x, 32);
-        uint8_t tileY = y >> 5;
+        int32_t flooredX = floor2(x, 32);
```
duncanspumpkin, line 57: const auto?

Gymnasiast (approved):

ZehMatt:
Actually don't merge this, I think its still broken on NSF.

src/openrct2/world/Sprite.cpp
```diff
@@ -54,8 +54,8 @@ constexpr size_t GetSpatialIndexOffset(int32_t x, int32_t y)
         x = std::clamp(x, 0, 0xFFFF);
         y = std::clamp(y, 0, 0xFFFF);
 
-        int16_t flooredX = floor2(x, 32);
-        uint8_t tileY = y >> 5;
+        const auto flooredX = floor2(x, 32);
+        const auto tileY = y >> 5;
         index = (flooredX << 3) | tileY;
```
Broxzier, line 59: With the current code, for every tile X increases the index by 256, and Y increases it by 1. Just adding their tile coords and multiply X by the max size is so much simpler. It would be much clearer what this is trying to accomplish by simply adding (x * max map size + y) together, where x and y are tile coords.

Also, when the max map size is not a power of 2 anymore (NSF), this would skip certain parts in the array. With the multiplication + addition you don't have that problem.

Broxzier:
```suggestion
        const auto tileX = x / 32;
        const auto tileY = y / 32;
        index = x * MAXIMUM_MAP_SIZE_TECHNICAL + tileY;
```

ZehMatt:
Heres a simplified form of what the current code is actually doing
```cpp
constexpr size_t GetSpatialIndexOffset2(int32_t x, int32_t y)
{    
    x = std::clamp(x, 0, 0xFFFF);
    y = std::clamp(y, 0, 0xFFFF);

    return ((x / 32) << 8) | (y / 32);
}```
Demo: https://godbolt.org/z/W459qj8fK

But we can probably just map this to tile x/y in a much simpler fashion.

File: src/openrct2/world/Sprite.cpp (hunk 2 above)
Comment (Gymnasiast) [reply src/openrct2/world/Sprite.cpp:59]: *whacks Hielke with the keyboard*
We have spent a lot of effort to get rid of naked division by 32 and such things. Please don’t reintroduce them. Use TileCoords(CoordsXY(x, y)).

Comment (Broxzier) [reply src/openrct2/world/Sprite.cpp:59]: The divisions were already there, just not as visible. `floor2(x, 32)` then shifting left by 3 and `y >> 5` are essentially divisions by 32 in disguise!

**Comment by ZehMatt** [comment]
I think this is the best possible solution.

**Review by Broxzier — APPROVED** [review]

File: src/openrct2/world/Sprite.cpp [hunk 3]
```diff
@@ -48,21 +48,18 @@ static void FreeEntity(EntityBase& entity);
...
-        x = std::clamp(x, 0, 0xFFFF);
-        y = std::clamp(y, 0, 0xFFFF);
+    if (x == LOCATION_NULL)
+        return SPATIAL_INDEX_LOCATION_NULL;
 
-        int16_t flooredX = floor2(x, 32);
-        uint8_t tileY = y >> 5;
-        index = (flooredX << 3) | tileY;
-    }
+    const auto tileX = std::clamp<size_t>(x / COORDS_XY_STEP, 0, MAXIMUM_MAP_SIZE_TECHNICAL);
```
Gymnasiast, line 54: Shouldn’t this have been `MAXIMUM_MAP_SIZE_TECHNICAL - 1`?

Broxzier: Good catch, clamp is not inclusive.

PR #15477, opened 2021-09-28
Update entities xyz to int32_t

ZehMatt:
Needs replay update. I would use CoordsXYZ but this will be a huge list of changes then, once its int32_t it will be storage wise compatible for when we place CoordsXYZ there.


duncanspumpkin:
It's too fundamental for a beginner to change that

duncanspumpkin:
What about location null?

ZehMatt:
> What about location null?

Good point

Gymnasiast (approved):

PR #15479, opened 2021-09-28
Refactor GetSpatialIndexOffset

ZehMatt:
Further simplified

### DIFF
--- a/src/openrct2/world/Sprite.cpp
+++ b/src/openrct2/world/Sprite.cpp
@@ -46,21 +46,19 @@ static std::array<std::vector<uint16_t>, SPATIAL_INDEX_SIZE> gSpriteSpatialIndex
 
 static void FreeEntity(EntityBase& entity);
 
-constexpr size_t GetSpatialIndexOffset(int32_t x, int32_t y)
+static constexpr size_t GetSpatialIndexOffset(const CoordsXY& loc)
 {
-    if (x == LOCATION_NULL)
+    if (loc.IsNull())
         return SPATIAL_INDEX_LOCATION_NULL;
 
-    const auto tileX = std::clamp<size_t>(x / COORDS_XY_STEP, 0, MAXIMUM_MAP_SIZE_TECHNICAL);
-    const auto tileY = std::clamp<size_t>(y / COORDS_XY_STEP, 0, MAXIMUM_MAP_SIZE_TECHNICAL);
-    const auto index = tileX * MAXIMUM_MAP_SIZE_TECHNICAL + tileY;
+    // NOTE: The input coordinate is rotated and can have negative components.
+    const auto tileX = std::abs(loc.x) / COORDS_XY_STEP;
+    const auto tileY = std::abs(loc.y) / COORDS_XY_STEP;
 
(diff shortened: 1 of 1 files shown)


Gymnasiast (commented):
Code looks good and also gets rid of a refactor error that sneaked into the previous PR.

Perhaps a good idea if someone tests it.

src/openrct2/world/Sprite.cpp
```diff
@@ -46,21 +46,13 @@ static std::array<std::vector<uint16_t>, SPATIAL_INDEX_SIZE> gSpriteSpatialIndex
...
-    if (x == LOCATION_NULL)
-        return SPATIAL_INDEX_LOCATION_NULL;
-
-    const auto tileX = std::clamp<size_t>(x / COORDS_XY_STEP, 0, MAXIMUM_MAP_SIZE_TECHNICAL);
-    const auto tileY = std::clamp<size_t>(y / COORDS_XY_STEP, 0, MAXIMUM_MAP_SIZE_TECHNICAL);
-    const auto index = tileX * MAXIMUM_MAP_SIZE_TECHNICAL + tileY;
-
-    if (index >= std::size(gSpriteSpatialIndex))
-    {
+    if (loc.IsNull() || !map_is_location_valid(loc))
```
duncanspumpkin, line 51: Do we really need the additional `map_is_location_valid`? This function is going to be called super super often.

ZehMatt: It no longer clamps so it needs to check that.

ZehMatt: It turns out that this new form is also the fastest, see https://quick-bench.com/q/fHtaplRpsvlpiWH3dvulMwgR90A

Broxzier (approved):

Broxzier:
Weren't negative coordinates already filtered out by calling `map_is_location_valid`? Why is this call not necessary anymore?

ZehMatt:
> Weren't negative coordinates already filtered out by calling `map_is_location_valid`? Why is this call not necessary anymore?

They should not be filtered, they are valid coordinates. Based on the rotation it may turn a component negative its not really out of bounds.

Gymnasiast (approved):
Testing revealed no bugs.

Would like @duncanspumpkin to confirm this also solves the performance issue.

duncanspumpkin:
I've locally applied this ontop of another branch and it resolved my perfomance issues. I think though we should also add a 
```cpp
    if (!map_is_location_valid(pos))
    {
        return;
    }
```
to the sprite_paint_setup function to avoid further pointless work.

ZehMatt:
> I've locally applied this ontop of another branch and it resolved my perfomance issues. I think though we should also add a
> 
> ```c++
>     if (!map_is_location_valid(pos))
>     {
>         return;
>     }
> ```
> 
> to the sprite_paint_setup function to avoid further pointless work.

I'm pretty sure this will break the rendering.

duncanspumpkin:
There should be no entities outside of `map_is_location_valid` that require rendering. Lets discuss it outside of this as this fixes one bit either way.

Issue #15136, opened 2021-07-30
SV6 export broken with Vanilla


duncanspumpkin:
**OS:** Windows 10
**Version:** 0.3.4.1
**Commit/Build:** 426e106

SV6 export looks to have broken sometime after 0.3.2. When saving in OpenRCT2 and then opening the file in vanilla the game will hang and fail to load. We have made a large number of changes to the exporter since then so it could be one of a number of changes. 

**Save game:**
[3.3.2save.zip](https://github.com/OpenRCT2/OpenRCT2/files/6909274/3.3.2save.zip)
[3.4.1save.zip](https://github.com/OpenRCT2/OpenRCT2/files/6909275/3.4.1save.zip)

PR #15482, opened 2021-09-29
Fix of sv6 export

duncanspumpkin:
Still testing this.
#15136

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -16,7 +16,7 @@
 - Fix: [#14649] ImageImporter incorrectly remaps colours outside the RCT2 palette.
 - Fix: [#14667] “Extreme Hawaiian Island” has unpurchaseable land tiles (original bug).
 - Fix: [#15096] Crash when placing entrances in the scenario editor near the map corner.
-- Fix: [#15136] SV6 export causing vanilla RCT2 to crash under certain situations.
+- Fix: [#15136] Exported SV6 files cause vanilla RCT2 to hang.
 - Fix: [#15142] ToonTowner's mine roofs were moved into the pirate theme scenery group instead of the mine theme scenery group.
 - Fix: [#15148] Track Designs Manager delete confirmation window doesn't display properly.
 - Fix: [#15170] Plugin: incorrect label text alignment.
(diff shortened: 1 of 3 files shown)


duncanspumpkin:
doesnt appear to resolve it quite correctly. Will keep investigating.

src/openrct2/rct2/S6Exporter.cpp
```diff
@@ -1123,21 +1143,13 @@ void S6Exporter::RebuildEntityLinks()
...
-                previous = entity->sprite_index;
-            }
-            if (previous != SPRITE_INDEX_NULL)
-            {
-                _s6.sprites[previous].unknown.next_in_quadrant = SPRITE_INDEX_NULL;
-            }
+            RebuildEntitySpatialLocation(TileCoordsXY{ x, y });
         }
     }
+    // Revuild next_in_quadrant linked list for LOCATION_NULL
```
ZehMatt, line 1149:
```suggestion
    // Rebuild next_in_quadrant linked list for LOCATION_NULL
```

distribution/changelog.txt
```diff
@@ -16,6 +16,7 @@
 - Fix: [#14649] ImageImporter incorrectly remaps colours outside the RCT2 palette.
 - Fix: [#14667] “Extreme Hawaiian Island” has unpurchaseable land tiles (original bug).
 - Fix: [#15096] Crash when placing entrances in the scenario editor near the map corner.
+- Fix: [#15136] SV6 export causing vanilla RCT2 to crash under certain situations.
```
Gymnasiast, line 19: It was a freeze and not a crash. Also, I think it was in most or even all situations.

```suggestion
- Fix: [#15136] Exported SV6 files cause vanilla RCT2 to hang.
```

Gymnasiast (commented):
I have a minor thingy about the changelog.

Apart from that, I have tested this PR and it works for me :+1: . There is another (unrelated) issue with map animations, but that’s something for its own issue/PR.

distribution/changelog.txt
duncanspumpkin: Hmm well that means i fixed two bugs then as it was crashing this morning after only the e16a4a0 commit.

Gymnasiast (approved):

ZehMatt (approved):
👍

PR #15488, opened 2021-09-29
Fix #14798: Overflow on sound position

### LINKED ISSUE #14798: [.park format] No in-game sound if view is in certain rotations on huge maps
**OS:** Windows 10
**Version:** 0.3.3
**Commit/Build:** fe939ec on new-save-format

If the view is rotated to face either east or west (denoted by the map icon's compass pointing right or left), then no in-game sound aside from the UI can be heard. Currently not sure if this is caused by an antivirus, but I don't think an antivirus should cause this issue to occur. I have only managed to trigger this issue on maps with a park size larger than 254x254 because I tried to replicate this issue on an SC6 like [European Extravaganza 2](https://drive.google.com/file/d/1NccE15n1hovmgKgXLePH4QaOLHpB40Q4/view?usp=sharing) and the issue did not occur whatsoever.

**Steps to reproduce:**
1. Open any scenario or park with a map size larger than 254x254
2. Build a few rides and let them test
3. Rotate the view to face either east or west

**Video:**
[Video showcasing the bug](https://youtu.be/p2_IZSbwGEI)
In the video, you can see that the sound works perfectly fine (we can hear the hyper-twister trains rolling through the track), but as soon as the view is rotated to either west or east, the sound disappears.

**Save game:**
[Fred's Workbench 2021 (999x999)](https://drive.google.com/file/d/1SbCehqL9O86CKsranQSFd2-Iyhq5CMZb/view?usp=sharing)
The .park file is where I got the issue to occur. I started noticing it while messing around with the RCT2 junior coaster default pre-built track designs.

ZehMatt:


src/openrct2/ride/Vehicle.cpp
```diff
@@ -973,10 +973,10 @@ bool Vehicle::SoundCanPlay() const
     if (g_music_tracking_viewport == nullptr)
         return false;
 
-    int16_t left = g_music_tracking_viewport->viewPos.x;
-    int16_t bottom = g_music_tracking_viewport->viewPos.y;
-    int16_t quarter_w = g_music_tracking_viewport->view_width / 4;
-    int16_t quarter_h = g_music_tracking_viewport->view_height / 4;
+    auto left = g_music_tracking_viewport->viewPos.x;
+    auto bottom = g_music_tracking_viewport->viewPos.y;
+    auto quarter_w = g_music_tracking_viewport->view_width / 4;
```
duncanspumpkin, line 978: const

ZehMatt: Thats the best I can do

duncanspumpkin (approved):

Gymnasiast (approved):

PR #15489, opened 2021-09-29
Use ZoomLevel more strictly

Broxzier:
This PR makes the use of `ZoomLevel` more explicit and `constexpr` qualified.

### DIFF
--- a/src/openrct2-ui/title/TitleSequencePlayer.cpp
+++ b/src/openrct2-ui/title/TitleSequencePlayer.cpp
@@ -267,7 +267,7 @@ private:
                 RotateView(command.Rotations);
                 break;
             case TitleScript::Zoom:
-                SetViewZoom(command.Zoom);
+                SetViewZoom(ZoomLevel{ static_cast<int8_t>(command.Zoom) });
                 break;
             case TitleScript::Speed:
                 gGameSpeed = std::clamp<uint8_t>(command.Speed, 1, 4);
@@ -317,7 +317,7 @@ private:
         return true;
     }
 
-    void SetViewZoom(const uint32_t& zoom)
+    void SetViewZoom(ZoomLevel zoom)
     {
         rct_window* w = window_get_main();
         if (w != nullptr && w->viewport != nullptr)
--- a/src/openrct2/drawing/Drawing.Sprite.cpp
+++ b/src/openrct2/drawing/Drawing.Sprite.cpp
@@ -458,8 +458,8 @@ void FASTCALL gfx_draw_sprite_palette_set_software(
     }
 
     // Its used super often so we will define it to a separate variable.
-    auto zoom_level = dpi->zoom_level;
-    int32_t zoom_mask = zoom_level > ZoomLevel{ 0 } ? 0xFFFFFFFF * zoom_level : 0xFFFFFFFF;
+    const auto zoom_level = dpi->zoom_level;
+    const int32_t zoom_mask = zoom_level > ZoomLevel{ 0 } ? 0xFFFFFFFF * zoom_level : 0xFFFFFFFF;
 
     if (zoom_level > ZoomLevel{ 0 } && g1->flags & G1_FLAG_RLE_COMPRESSION)
     {
--- a/src/openrct2/drawing/Drawing.Sprite.cpp
+++ b/src/openrct2/drawing/Drawing.Sprite.cpp
@@ -435,7 +435,7 @@ void FASTCALL gfx_draw_sprite_palette_set_software(
         return;
     }
 
-    if (dpi->zoom_level > 0 && (g1->flags & G1_FLAG_HAS_ZOOM_SPRITE))
+    if (dpi->zoom_level > ZoomLevel{ 0 } && (g1->flags & G1_FLAG_HAS_ZOOM_SPRITE))
     {
         rct_drawpixelinfo zoomed_dpi = *dpi;
         zoomed_dpi.bits = dpi->bits;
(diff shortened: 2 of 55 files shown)


src/openrct2-ui/title/TitleSequencePlayer.cpp
```diff
@@ -317,12 +317,12 @@ class TitleSequencePlayer final : public ITitleSequencePlayer
         return true;
     }
 
-    void SetViewZoom(const uint32_t& zoom)
+    void SetViewZoom(int8_t zoom)
```
Gymnasiast, line 320: Why not typehint ZoomLevel here?

Broxzier: I was a bit hesitant with this one since it uses the value from `TitleCommand` directly, and I'm not sure how that gets loaded exactly. Looks like there is just one call for it, so just updating this function should be easy, though I'll leave `TitleCommand::Zoom` untouched for now.

src/openrct2/interface/Screenshot.cpp
```diff
@@ -387,9 +387,8 @@ void screenshot_giant()
             throw std::runtime_error("Giant screenshot failed, unable to find a suitable destination path.");
         }
 
-        int32_t rotation = get_current_rotation();
-        ZoomLevel zoom = 0;
-
+        auto rotation = get_current_rotation();
```
ZehMatt, line 390:
```suggestion
        const auto rotation = get_current_rotation();
```

src/openrct2/interface/Viewport.cpp
```diff
@@ -952,7 +952,7 @@ void viewport_paint(
     uint32_t viewFlags = viewport->flags;
     uint32_t width = right - left;
     uint32_t height = bottom - top;
-    uint32_t bitmask = viewport->zoom >= 0 ? 0xFFFFFFFF & (0xFFFFFFFF * viewport->zoom) : 0xFFFFFFFF;
+    uint32_t bitmask = viewport->zoom >= ZoomLevel{ 0 } ? 0xFFFFFFFF & (0xFFFFFFFF * viewport->zoom) : 0xFFFFFFFF;
```
ZehMatt, line 955:
```suggestion
    const uint32_t bitmask = viewport->zoom >= ZoomLevel{ 0 } ? 0xFFFFFFFF & (0xFFFFFFFF * viewport->zoom) : 0xFFFFFFFF;
```
I know we have this in at least two places, should move this at some point into a function.

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -452,16 +452,16 @@ void FASTCALL gfx_draw_sprite_palette_set_software(
...
-    if (dpi->zoom_level > 0 && (g1->flags & G1_FLAG_NO_ZOOM_DRAW))
+    if (dpi->zoom_level > ZoomLevel{ 0 } && (g1->flags & G1_FLAG_NO_ZOOM_DRAW))
     {
         return;
     }
 
     // Its used super often so we will define it to a separate variable.
     auto zoom_level = dpi->zoom_level;
-    int32_t zoom_mask = zoom_level > 0 ? 0xFFFFFFFF * zoom_level : 0xFFFFFFFF;
+    int32_t zoom_mask = zoom_level > ZoomLevel{ 0 } ? 0xFFFFFFFF * zoom_level : 0xFFFFFFFF;
```
ZehMatt, line 462:
```suggestion
    const int32_t zoom_mask = zoom_level > ZoomLevel{ 0 } ? 0xFFFFFFFF * zoom_level : 0xFFFFFFFF;
```

IntelOrca:
This is a subjective change. I originally made it implicit because I personally preferred it, but if no one else does, then fair enough.

Gymnasiast:
> This is a subjective change. I originally made it implicit because I personally preferred it, but if no one else does, then fair enough.

I prefer the implicit way too, if I’m honest.

ZehMatt:
I prefer that its explicit with all types.

Broxzier:
I also prefer explicit, especially because `ZoomLevel` overloads the `*` and `/` operators.

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -2020,25 +2020,24 @@ void track_design_draw_preview(TrackDesign* td6, uint8_t* pixels)
...
 
     if (size_x > 1600 || size_z > 1000)
     {
-        zoom_level = 3;
+        zoom_level = ZoomLevel{ 3 };
     }
 
-    size_x = 370 << zoom_level;
-    size_y = 217 << zoom_level;
+    size_x = 370 * zoom_level;
```
Gymnasiast, line 2039: Doesn’t look right to me. If the zoomlevel is 2, then << 2 = 4.

Gymnasiast (changes requested):
Some stuff is not correct. See the line notes.

Broxzier:
@Gymnasiast this PR is still awaiting your re-review. See comments as to why no further changes were made after your last review.

Gymnasiast:
When discussing this in the team chat, I made it clear that I disagree with the stated reasons. This has not changed, so I am not approving it. 

A ZoomLevel is a number and not a complex structure, so multiplication should be multiplication and not a left shift disguising as multiplication.

ZehMatt:
99% of the changes are unrelated to how we apply the zoom on coordinates. I suggest we merge this as is since its functionally equivalent to develop and create a new issue to do a follow up PR eliminating the shifts/multiply and instead use a function to be explicit about applying a zoom.

Broxzier:
A refactoring issue for that has already been created here: https://github.com/OpenRCT2/OpenRCT2/issues/15896

This PR did not introduce the operators, it merely applied it to a few cases where bitshifts were hardcoded, this making them more explicit.

As for the remark about renaming a variable from "max" to "num" (since that is what the value represents), that also remains to be accepted by you. I don't think there has been any discussion about that on Discord?

src/openrct2/interface/Screenshot.cpp
```diff
@@ -448,31 +447,32 @@ static void benchgfx_render_screenshots(const char* inputPath, std::unique_ptr<I
 
     // Create Viewport and DPI for every rotation and zoom.
     constexpr int32_t MAX_ROTATIONS = 4;
-    constexpr int32_t MAX_ZOOM_LEVEL = 3;
-    std::array<rct_drawpixelinfo, MAX_ROTATIONS * MAX_ZOOM_LEVEL> dpis;
-    std::array<rct_viewport, MAX_ROTATIONS * MAX_ZOOM_LEVEL> viewports;
+    constexpr auto NUM_ZOOM_LEVELS = static_cast<int8_t>(ZoomLevel::max());
```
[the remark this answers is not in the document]
Gymnasiast: I see, can you put that in a comment above, then?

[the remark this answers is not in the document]
Broxzier: Sure thing.

Gymnasiast (approved):
Since the `*` is going to be replaced in another PR, approving this.

PR #15490, opened 2021-09-29
Fix: Surface slopes edited by TI have invalid clearance

Broxzier:
Something I noticed a while ago: when making changes with the land tool, the clearance height of surfaces is not changed - it always matches the base height. The Tile Inspector does update the clearance height, however. This PR removes that behaviour.

I'm not sure if this has any influence on other game mechanics, I did not find any. Perhaps collision on crashing vehicles is different.


IntelOrca:
This should increment the network version. Does it also need a changelog entry?

PR #15493, opened 2021-09-29
Minor code cleanups

Broxzier:


src/openrct2-ui/WindowManager.cpp
```diff
@@ -342,7 +342,6 @@ class WindowManager final : public IWindowManager
                 {
                     window_close_construction_windows();
                     _currentRideIndex = static_cast<ride_id_t>(rideIndex);
-                    w = OpenWindow(WC_RIDE_CONSTRUCTION);
```
duncanspumpkin, line 345: Are you sure about this change. You are now no longer opening the construction window if it isn't already opened.

src/openrct2-ui/interface/Widget.cpp
```diff
@@ -491,8 +489,8 @@ static void WidgetGroupboxDraw(rct_drawpixelinfo* dpi, rct_window* w, rct_widget
     // Resolve the absolute ltrb
     l = w->windowPos.x + widget->left;
     t = w->windowPos.y + widget->top + 4;
-    r = w->windowPos.x + widget->right;
-    b = w->windowPos.y + widget->bottom;
+    int32_t r = w->windowPos.x + widget->right;
```
duncanspumpkin, line 492: `const auto`

src/openrct2-ui/WindowManager.cpp
Broxzier: Nope, that slipped in. Only the assignment to `w` is useless.

src/openrct2/Game.cpp
```diff
@@ -530,11 +528,10 @@ void game_load_init()
     {
         viewport_init_all();
         game_create_windows();
-        mainWindow = window_get_main();
     }
     else
     {
-        mainWindow = window_get_main();
+        const auto mainWindow = window_get_main();
```
Gymnasiast, line 534: Shouldn’t this be

```suggestion
        const auto* mainWindow = window_get_main();
```
?

src/openrct2/ride/TrackDesign.cpp
```diff
@@ -247,7 +247,6 @@ rct_string_id TrackDesign::CreateTrackDesignTrack(const Ride& ride)
         }
         trackElement.x = newCoords->x;
         trackElement.y = newCoords->y;
-        z = newCoords->z;
```
Gymnasiast, line 250: `z` is used on line 290 - removing the statement changes the value of it.

src/openrct2/Game.cpp
Broxzier: With and without explicit `*` seems to be a bit mixed.

```
src/openrct2-ui/WindowManager.cpp:527:        auto mainWindow = window_get_main();
src/openrct2-ui/input/InputManager.cpp:115:    auto mainWindow = window_get_main();
src/openrct2-ui/input/Shortcuts.cpp:60:        auto window = window_get_main();
src/openrct2-ui/input/Shortcuts.cpp:72:        auto window = window_get_main();
src/openrct2-ui/windows/RideConstruction.cpp:2141:            auto* mainWnd = window_get_main();
src/openrct2-ui/windows/TopToolbar.cpp:837:    const auto* mainWindow = window_get_main();
src/openrct2-ui/windows/TopToolbar.cpp:1217:                auto* mainWnd = window_get_main();
src/openrct2-ui/windows/Viewport.cpp:87:        auto* mainWindow = window_get_main();
src/openrct2-ui/windows/Viewport.cpp:106:        auto* mainWindow = window_get_main();
src/openrct2-ui/windows/Viewport.cpp:142:                auto* mainWindow = window_get_main();
src/openrct2/ReplayManager.cpp:862:                    auto* mainWindow = window_get_main();
src/openrct2/drawing/LightFX.cpp:305:                auto* w = window_get_main();
src/openrct2/interface/Screenshot.cpp:393:        auto mainWindow = window_get_main();
src/openrct2/interface/Window.cpp:1800:    auto mainWindow = window_get_main();
src/openrct2/interface/Window.cpp:1830:    auto mainWindow = window_get_main();
src/openrct2/interface/Window.cpp:1860:    auto mainWindow = window_get_main();
src/openrct2/interface/Window.cpp:1890:    auto mainWindow = window_get_main();
src/openrct2/interface/Window_internal.cpp:20:    auto mainWindow = window_get_main();
```

Gymnasiast: I think there is a general movement towards always using `auto*` if the variable in question is a pointer in our codebase. (Would be lovely if we could enforce that automatically using clang-format or something like that.)

src/openrct2/ride/TrackDesign.cpp
Broxzier: `z` gets assigned again before that on line 263. The value here is completely discarded.

src/openrct2/Game.cpp
Broxzier: Sounds like a good idea, that also helps finding instances that could be moved over to std::shared_ptr more easily.

src/openrct2/Game.cpp
```diff
@@ -530,11 +528,10 @@ void game_load_init()
     {
         viewport_init_all();
         game_create_windows();
-        mainWindow = window_get_main();
     }
     else
     {
-        mainWindow = window_get_main();
+        auto* mainWindow = window_get_main();
```
Gymnasiast, line 534:
```suggestion
        const auto* mainWindow = window_get_main();
```
Or just directly feed it into `window_unfollow_sprite()`.

Gymnasiast (approved):

ZehMatt (approved):
🎡

PR #15494, opened 2021-09-30
Modify StaffSetPatrolAreaAction to have three modes

duncanspumpkin:
This is to fix the potential desync if using the clear all patrol area button in the staff window.


duncanspumpkin:
No idea how the replays could have passed as there is one that sets the patrol area and we arent providing it all of the parameters?!?!?

src/openrct2-ui/windows/Staff.cpp
```diff
@@ -567,11 +567,10 @@ void window_staff_overview_dropdown(rct_window* w, rct_widgetindex widgetIndex,
...
                     return;
                 }
-                // TODO: THIS SHOULD BE NETWORKED
-                peep->ClearPatrolArea();
 
-                gfx_invalidate_screen();
-                staff_update_greyed_patrol_areas();
+                auto staffSetPatrolAreaAction = StaffSetPatrolAreaAction(
+                    peep->sprite_index, {}, StaffSetPatrolAreaMode::ClearAll);
+                GameActions::Execute(&staffSetPatrolAreaAction);
```
Gymnasiast, line 573: Why did you remove the `gfx_invalidate_screen()` call?

src/openrct2/actions/StaffSetPatrolAreaAction.cpp
```diff
@@ -54,19 +55,10 @@ GameActions::Result::Ptr StaffSetPatrolAreaAction::Query() const
     return MakeResult();
 }
 
-GameActions::Result::Ptr StaffSetPatrolAreaAction::Execute() const
+static void UpdateStaffMode(Staff& staff)
```
Gymnasiast, line 58:
```suggestion
static void UpdateStaffMode(const Staff& staff)
```

src/openrct2-ui/windows/Staff.cpp
```diff
@@ -1245,7 +1244,9 @@ void window_staff_overview_tool_down(rct_window* w, rct_widgetindex widgetIndex,
         {
             _staffPatrolAreaPaintValue = PatrolAreaValue::SET;
         }
-        auto staffSetPatrolAreaAction = StaffSetPatrolAreaAction(w->number, destCoords);
+        auto staffSetPatrolAreaAction = StaffSetPatrolAreaAction(
+            w->number, destCoords,
+            _staffPatrolAreaPaintValue == PatrolAreaValue::SET ? StaffSetPatrolAreaMode::Set : StaffSetPatrolAreaMode::Unset);
```
Gymnasiast, line 1249: Wouldn’t it be cleaner to change the type of `_staffPatrolAreaPaintValue` to `StaffSetPatrolAreaMode`? There seems to be little reason to keep two almost the same enums.

duncanspumpkin: No cant really do that without making the window code look a bit rubbish. Window code has 3 states: set, unset, None. Action has 3 states: set, unset, clear. I'd either need to change it so that _staffPatrolAreaPaintValue was a std::optional or I'd have to make it that clear meant none.

Gymnasiast (approved):

src/openrct2/actions/StaffSetPatrolAreaAction.cpp
```diff
@@ -78,20 +70,52 @@ GameActions::Result::Ptr StaffSetPatrolAreaAction::Execute() const
...
+}
 
+static void InvalidatePatrolTile(const CoordsXY& loc)
+{
     for (int32_t y = 0; y < 4 * COORDS_XY_STEP; y += COORDS_XY_STEP)
     {
         for (int32_t x = 0; x < 4 * COORDS_XY_STEP; x += COORDS_XY_STEP)
         {
-            map_invalidate_tile_full({ (_loc.x & 0x1F80) + x, (_loc.y & 0x1F80) + y });
+            map_invalidate_tile_full({ (loc.x & 0x1F80) + x, (loc.y & 0x1F80) + y });
```
ZehMatt, line 87:
```suggestion
            map_invalidate_tile_full(loc + CoordsXY{ x, y });
```
The location needs to be aligned only once, best to make a copy before the loop.

PR #15495, opened 2021-09-30
Use GetLocation in more places

ZehMatt:


src/openrct2/paint/sprite/Paint.Sprite.cpp
```diff
@@ -65,21 +65,23 @@ void sprite_paint_setup(paint_session* session, const CoordsXY& pos)
             }
         }
 
+        const auto entityPos = spr->GetLocation();
```
Gymnasiast, line 68: Just checking: not `const auto&`?

src/openrct2/peep/Guest.cpp
```diff
@@ -5027,11 +5027,13 @@ void Guest::UpdateRideShopLeave()
 {
     if (auto loc = UpdateAction(); loc.has_value())
     {
-        MoveTo({ loc.value(), z });
+        const auto curLoc = GetLocation();
+
+        MoveTo({ loc.value(), curLoc.z });
 
-        if ((x & 0xFFE0) != NextLoc.x)
+        if ((curLoc.x & 0xFFE0) != NextLoc.x)
```
ZehMatt, line 5034: This needs to use the new location

Gymnasiast, line 5034: This looks like a hand-rolled `ToTileStart()`

src/openrct2/paint/sprite/Paint.Sprite.cpp
ZehMatt: GetLocation currently returns a copy as we are not storing that just yet as CoordsXYZ

src/openrct2/peep/Peep.cpp
```diff
@@ -474,13 +474,14 @@ std::optional<CoordsXY> Peep::UpdateAction(int16_t& xy_distance)
     WindowInvalidateFlags |= PEEP_INVALIDATE_PEEP_2;
 
     // Create sick at location
```
Gymnasiast, line 476:
```suggestion
```

It’s clear from the code anyway, and it’s a colloquial usage of the word that won’t be familiar to everyone.

Gymnasiast (approved):

PR #15498, opened 2021-09-30
Backport scenery window from NSF

duncanspumpkin:
This was just a big checkout losing all history. Still needs testing and such.


duncanspumpkin:
Doesnt appear to be respecting research atm.

duncanspumpkin:
Could do with going on a `const` spree.

src/openrct2/world/Scenery.cpp
```diff
@@ -281,3 +284,149 @@ int32_t wall_entry_get_door_sound(const WallSceneryEntry* wallEntry)
...
+        case SCENERY_TYPE_LARGE:
+            return ObjectType::LargeScenery;
+        case SCENERY_TYPE_BANNER:
+            return ObjectType::Banners;
+        default:
+            throw std::runtime_error("Invalid scenery type");
+    }
+}
+
+uint8_t GetSceneryTypeFromObjectType(ObjectType type)
```
Gymnasiast, line 415: The same code also resides in the SceneryGroupObject. That code would need to be modified to use this as well.

duncanspumpkin: I dont really see the point in the SceneryType object. I'm going to look see if we can just remove it.

duncanspumpkin: Eitherway these two functions are unused so I'm removing them.

duncanspumpkin:
Possibly need to bring over the changes to the top toolbar window as well.

Gymnasiast:
> Possibly need to bring over the changes to the top toolbar window as well.

Have you decided yet if this is needed? (Would like to know when I can start testing.)

duncanspumpkin:
> > Possibly need to bring over the changes to the top toolbar window as well.
> 
> Have you decided yet if this is needed? (Would like to know when I can start testing.)

I've been away all weekend so forgot i had said this. I'll investigate later today

duncanspumpkin:
Decided that they were required and have brought them over.

duncanspumpkin:
That isnt a bug that's by design in the base game so that it resets every periodic update to prevent it displaying incorrectly when placing.

PR #15499, opened 2021-09-30
Fix incorrect staff costumes being available

duncanspumpkin:
Noticed this when trying to fix an NSF scenery window issue. The previous code meant that if any single scenery item in a set was researched then the costume was available but I'm pretty sure its meant to be tied to if the whole scenery set is invented.

### DIFF
--- a/src/openrct2/management/Research.cpp
+++ b/src/openrct2/management/Research.cpp
@@ -591,28 +591,20 @@ void scenery_set_not_invented(const ScenerySelection& sceneryItem)
 
 bool scenery_group_is_invented(int32_t sgIndex)
 {
-    auto invented = false;
     const auto sgEntry = get_scenery_group_entry(sgIndex);
     if (sgEntry != nullptr && sgEntry->entry_count > 0)
     {
         if (gCheatsIgnoreResearchStatus)
         {
-            invented = true;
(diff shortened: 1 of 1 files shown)


src/openrct2/management/Research.cpp
```diff
@@ -591,28 +591,20 @@ void scenery_set_not_invented(const ScenerySelection& sceneryItem)
 
 bool scenery_group_is_invented(int32_t sgIndex)
 {
-    auto invented = false;
     const auto sgEntry = get_scenery_group_entry(sgIndex);
     if (sgEntry != nullptr && sgEntry->entry_count > 0)
```
Broxzier, line 595: Is this check still useful now? I'd guess that this is made obsolete by the check `item.entryIndex == sgIndex` later on.

duncanspumpkin: I originally simplified it to just the none_of but then when you look at the caller its going to pass in completely invalid sgIndex's a lot of the time so might as well save some searching time by doing a quick nullptr check.

ZehMatt: I would prefer this returns early if its null, reduces a scope.

Broxzier (approved):

PR #15504, opened 2021-10-01
Fix #15503: Freeze when doing specific merges

### LINKED ISSUE #15503: Coaster merge that causes game to crash
### Operating System

Windows 10, 64 bit

### OpenRCT2 build

any version

### Describe the issue

Merging a dead end into a block section mode ride crashes the game and it seems like it also crashes a server.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Build a coaster in that works in block section mode (with at least two trains).
2. Merge a dead end into the coaster that the train would use.
3. Test the coaster or if it's already running reset it (double click red) and then test it.

### Attachments

game_and_server_crashing_coaster_merge

Gymnasiast:


tupaschoal (commented):
Will `car->remaining_distance` be at a sane value when this stops (and does it matter)?

Gymnasiast:
I don’t know, to be honest. Of the two saves in the issue, one roller coaster now works as it should, while the other places a train on the dead end, which then promptly comes crashing down. I figured it was on the user to build proper track and our most important concern was not crashing or freezing the game if they don’t.

I don’t know if there is a set of sane values we could apply. I guess I could set the value of `car->remaining_distance` to something - but what?

ZehMatt:
While it does alleviate the issue of entirely freezing the game this however introduces quite a heavy loop so depending on how often this is called this may be some invisible performance eater the end user might be not aware of to why this is happening now. It might be worthwhile to implement an news alert to the user when this loop breaks out hitting the limit, not asking to be implemented right now its just an idea.

duncanspumpkin:
Have you checked this on parks with very large block break track sections? I'm not a huge fan of this solution. Ideally we would have a tortoise hare algorithm but it's not at all easy to do that with motion update atm.

Gymnasiast:
@duncanspumpkin Is it plausible they would hits 10,000 iterations? I could make it 100,000 or 1,000,000 if that feels safer - a two-second freeze is far better than an infinite one, after all.

duncanspumpkin:
I think the largest track piece comprises of ~200 sub positions and most are ~32. That would mean 50 of the longest track pieces 
 or ~300 regular pieces would be enough to exceed this limit. That seems completely possible. 1Million would be a safe limit to go for but its still not great.

duncanspumpkin:
No i don't really have a solution. Just leave a todo stating that ideally this would be done differently.

src/openrct2/ride/Ride.cpp
```diff
@@ -3492,8 +3492,15 @@ void Ride::MoveTrainsToBlockBrakes(TrackElement* firstBlock)
             continue;
         }
 
+        size_t numIterations = 0;
         do
         {
+            // Fixes both freezing issues in #15503.
```
Gymnasiast, line 3498:
```suggestion
            // Fixes both freezing issues in #15503.
            // TODO: refactor the code so a tortoise-and-hare algorithm can be used.
```

Issue #15506, opened 2021-10-01
Change scenery_small_entry_has_flag() to a method


Gymnasiast:
Currently, this is a function that takes a SmallSceneryEntry. This needs to be changed to be a method of SmallSceneryEntry.

PR #15509, opened 2021-10-01
Make scenery_small_entry_has_flag a member function

Broxzier:
This implements #15506


tupaschoal (approved):
A few places are dereferencing the `sceneryEntry` without checking first, but it was already doing so before too, so I'm not opening issues in here for this.

PR #15511, opened 2021-10-01
Refactor RideCreateAction result data passing

ZehMatt:


ZehMatt:
When doing refactors like this one has to be careful about the hidden title passing in the custom result class. Once all the inheritance is gone I wanna look into making the title/message thing explicit.

duncanspumpkin (approved):
All looks sensible

PR #15512, opened 2021-10-01
Use smart pointers in OpenGLDrawingEngine.cpp

frutiemax:


src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -188,15 +188,15 @@ class OpenGLDrawingEngine final : public IDrawingEngine
     uint32_t _height = 0;
     uint32_t _pitch = 0;
     size_t _bitsSize = 0;
-    uint8_t* _bits = nullptr;
+    std::unique_ptr<uint8_t> _bits;
```
CookiePLMonster, line 191: I'm not convinced this works as expected. You replaced a pointer to an arbitrary number of `uint8_t` with a pointer to a single value. You probably intended this to be `std::unique_ptr<uint8_t[]>`.

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -431,17 +425,16 @@ class OpenGLDrawingEngine final : public IDrawingEngine
                     dst += pitch;
                 }
             }
-            delete[] _bits;
         }
 
-        _bits = newBits;
+        _bits.reset(newBits);
```
CookiePLMonster, line 430: I feel like `newBits` should also be an `unique_ptr<uint8_t[]>` that is moved into `_bits` instead.

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -206,8 +206,8 @@ class OpenGLDrawingEngine final : public IDrawingEngine
 
     explicit OpenGLDrawingEngine(const std::shared_ptr<IUiContext>& uiContext)
         : _uiContext(uiContext)
-        , _drawingContext(new OpenGLDrawingContext(this))
-        , _weatherDrawer(_drawingContext)
+        , _drawingContext(new OpenGLDrawingContext(*this))
```
ZehMatt, line 209: Use make_unique

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -452,22 +445,10 @@ class OpenGLDrawingEngine final : public IDrawingEngine
...
-        {
-            delete _scaleFramebuffer;
-            _scaleFramebuffer = nullptr;
-        }
-        if (_smoothScaleFramebuffer != nullptr)
-        {
-            delete _smoothScaleFramebuffer;
-            _smoothScaleFramebuffer = nullptr;
-        }
+        _screenFramebuffer.reset(new OpenGLFramebuffer(_window));
```
ZehMatt, line 448: Use make_unique

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -452,22 +445,10 @@ class OpenGLDrawingEngine final : public IDrawingEngine
...
         if (GetContext()->GetUiContext()->GetScaleQuality() != ScaleQuality::NearestNeighbour)
         {
-            _scaleFramebuffer = new OpenGLFramebuffer(_width, _height, false, false);
+            _scaleFramebuffer.reset(new OpenGLFramebuffer(_width, _height, false, false));
```
ZehMatt, line 451: make_unique

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
```diff
@@ -526,8 +501,7 @@ void OpenGLDrawingContext::Resize(int32_t width, int32_t height)
     _drawLineShader->SetScreenSize(width, height);
 
     // Re-create canvas framebuffer
-    delete _swapFramebuffer;
-    _swapFramebuffer = new SwapFramebuffer(width, height);
+    _swapFramebuffer.reset(new SwapFramebuffer(width, height));
```
ZehMatt, line 504: make_unique

ZehMatt (changes requested):

src/openrct2-ui/drawing/engines/opengl/OpenGLDrawingEngine.cpp
duncanspumpkin: In general I would replace it with a std::vector instead of a unique ptr.

CookiePLMonster: IMO use of `std::vector` implies resizability, if this array stays at a constant size then it shouldn't be needed.

ZehMatt: I think it only ever gets resized when the window size changes, either way works, the benefit of vector is that you will probably not re-allocate when you change the window size often but that doesn't matter as much here.

CookiePLMonster:
Also this is not related to the refactored lines specifically, but it seems like a good opportunity to refactor constructors like `OpenGLDrawingEngine` to take a `std::shared_ptr` instead of `const std::shared_ptr&`, this way `std::move` can be used to move pointers into the class instead of copying them.

PR #15517, opened 2021-10-02
Fix #14010: Refactor Velocity Literals to use _mph

### LINKED ISSUE #14010: Refactor Velocity Literals to use `_mph`
In our sister project I introduced the `_mph` literal for speeds https://github.com/OpenLoco/OpenLoco/pull/736
I'm almost definitely the underlying units are the same in OpenRCT2. Therefore we should bring in the same changes. First add the string literal code. Then find all of the velocity literals in the vehicle code.

Its quite easy to do just convert the literal to hex so `393216` = `0x60000` in hex then you drop the 4 `0`'s to get `0x6` which is the speed in `6.0_mph`

https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/ride/Vehicle.cpp#L5651
```
        if (velocity > -0x2C000)
            return OpenRCT2::Audio::SoundId::Null;
```
becomes
```
        if (velocity > -2.75_mph)
            return OpenRCT2::Audio::SoundId::Null;
```

You can use static_asserts to confirm that the values are identical `static_assert(2.75_mph == 0x2C000);`

sohamroy19:
`_mph` literal for speeds was introduced in [OpenLoco/OpenLoco](https://github.com/OpenLoco/OpenLoco)
Changed most speeds and velocities to `SpeedTemplate` & Literal `_mph` types

Ended up having to change significantly more code than I expected to, assuming this is what was supposed to be done.

### DIFF
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -29,6 +29,7 @@
 #include <openrct2/actions/RideSetPriceAction.h>
 #include <openrct2/actions/RideSetSettingAction.h>
 #include <openrct2/audio/audio.h>
+#include <openrct2/common.h>
 #include <openrct2/config/Config.h>
 #include <openrct2/core/String.hpp>
 #include <openrct2/localisation/Date.h>
@@ -2529,7 +2530,7 @@ static rct_string_id window_ride_get_status_vehicle(rct_window* w, Formatter& ft
             || trackType == TrackElemType::Up25ToFlat || trackType == TrackElemType::Up60ToFlat
             || trackType == TrackElemType::DiagUp25ToFlat || trackType == TrackElemType::DiagUp60ToFlat)
         {
-            if (ride->GetRideTypeDescriptor().SupportsTrackPiece(TRACK_BLOCK_BRAKES) && vehicle->velocity == 0)
+            if (ride->GetRideTypeDescriptor().SupportsTrackPiece(TRACK_BLOCK_BRAKES) && vehicle->velocity == 0.0_mph32)
             {
                 ft.Add<rct_string_id>(STR_STOPPED_BY_BLOCK_BRAKES);
                 return STR_BLACK_STRING;
--- a/src/openrct2/common.h
+++ b/src/openrct2/common.h
@@ -125,6 +125,29 @@ constexpr money16 ToMoney16(money64 value)
     return value == MONEY64_UNDEFINED ? MONEY16_UNDEFINED : static_cast<money16>(value);
 }
 
+constexpr int32_t operator""_mph32(long double speedMph)
+{
+    return static_cast<int32_t>(speedMph * 65536); // 65536 = 2^16, might truncate to cast to int32_t
+}
+
+constexpr int16_t operator""_mph(unsigned long long int speedMph)
+{
+    return static_cast<int16_t>(speedMph);
+}
+
+static_assert(439800 == 6.710816_mph32);
+static_assert(58640 == 0.894776_mph32);
+static_assert(0x20364 == 2.013245_mph32);
+static_assert(0x8000 == 0.5_mph32);
+static_assert(0xB0000 == 11.0_mph32);
+static_assert(131940 == 2.013245_mph32);
+static_assert(98955 == 1.509934_mph32);
+static_assert(1572864 == 24.0_mph32);
+static_assert(439800 == 6.710816_mph32);
+static_assert(0x2C000 == 2.75_mph32);
+static_assert(131072 == 2.0_mph32);
+static_assert(524288 == 8.0_mph32);
+
 using EMPTY_ARGS_VOID_POINTER = void();
 using rct_string_id = uint16_t;
 
--- a/src/openrct2/rct1/S4Importer.cpp
+++ b/src/openrct2/rct1/S4Importer.cpp
@@ -15,6 +15,7 @@
 #include "../ParkImporter.h"
 #include "../actions/WallPlaceAction.h"
 #include "../audio/audio.h"
+#include "../common.h"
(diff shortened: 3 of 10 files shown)


src/openrct2/world/Entity.cpp
```diff
@@ -239,7 +239,7 @@ void Vehicle::Serialise(DataSerialiser& stream)
     stream << Pitch;
     stream << bank_rotation;
     stream << remaining_distance;
-    stream << velocity;
+    stream << velocity.getRaw();
```
duncanspumpkin, line 242: This is broken you can't do this. You need to define a new serialising trait for the Speed16 and Speed32

src/openrct2-ui/windows/Ride.cpp
```diff
@@ -2530,7 +2530,7 @@ static rct_string_id window_ride_get_status_vehicle(rct_window* w, Formatter& ft
             || trackType == TrackElemType::Up25ToFlat || trackType == TrackElemType::Up60ToFlat
             || trackType == TrackElemType::DiagUp25ToFlat || trackType == TrackElemType::DiagUp60ToFlat)
         {
-            if (ride->GetRideTypeDescriptor().SupportsTrackPiece(TRACK_BLOCK_BRAKES) && vehicle->velocity == 0)
+            if (ride->GetRideTypeDescriptor().SupportsTrackPiece(TRACK_BLOCK_BRAKES) && vehicle->velocity.getRaw() == 0)
```
duncanspumpkin, line 2533: You should only use GetRaw if you absolutely have to. In this situation the simpler way is to change the rhs to 0_mph

src/openrct2/Speed.hpp
```diff
@@ -0,0 +1,136 @@
...
+    };
+
+    using Speed32 = SpeedTemplate<int32_t>;
+    using Speed16 = SpeedTemplate<int16_t>;
+
+    constexpr auto speed16Null = Speed16(-1);
+    // Truncates only use if safe to lose information
+    constexpr Speed16 toSpeed16(Speed32 speed)
+    {
+        return Speed16(speed.getRaw() / 65535);
```
Broxzier, line 104: Very slightly off error I think, but shouldn't this be 65536?
Pinging @duncanspumpkin since you probably know best.

src/openrct2/actions/RideCreateAction.cpp
```diff
@@ -187,7 +187,7 @@ GameActions::Result::Ptr RideCreateAction::Execute() const
     const auto& operatingSettings = rtd.OperatingSettings;
     ride->operation_option = (operatingSettings.MinValue * 3 + operatingSettings.MaxValue) / 4;
 
-    ride->lift_hill_speed = rtd.LiftData.minimum_speed;
+    ride->lift_hill_speed = OpenRCT2::SpeedTemplate<uint8_t>(rtd.LiftData.minimum_speed);
```
duncanspumpkin, line 190: You are using 8bit in tons of places just create a Speed8 like the Speed16

src/openrct2/Speed.hpp
duncanspumpkin: Hmm that's a good point. Not sure how I haven't had issues with that in loco

duncanspumpkin:
I'll review this properly next week but there are a few major issues i hadn't thought about when I wrote the original issue. I think the template needs a rework to handle 8bit. I've a few ideas on how best to do that.

src/openrct2/Speed.hpp
Broxzier: It's only a 0,0015% difference in speed, so hard to notice, but it would be best to be exact. Any `static_assert`s checking for `Speed16` may not be high enough to catch this.

sohamroy19:
The side effects are getting a bit too convoluted for me, perhaps it would help if you laid down multiple smaller steps (next week) or someone more experienced worked on this!

duncanspumpkin:
I had a think about this and the best approach will be to split up the Speed16/32 template stuff from the literals. So just remove all of `template<typename Value> struct SpeedTemplate`. And make the two literals return int32_t or int16_t and revert any instances where you have changed the type in a structure. It might be an idea to also encode the size into the mph literal so that its `_mph32` and `_mph`. This will hopefully reduce the chance that someone might accidentally compare a 32 with a 16. 

You will also need to rework the static casts as they will need adjustment as comparisons between types will not be possible.

sohamroy19:
> I had a think about this and the best approach will be to split up the Speed16/32 template stuff from the literals. So just remove all of `template<typename Value> struct SpeedTemplate`. And make the two literals return int32_t or int16_t and revert any instances where you have changed the type in a structure. It might be an idea to also encode the size into the mph literal so that its `_mph32` and `_mph`. This will hopefully reduce the chance that someone might accidentally compare a 32 with a 16.
> 
> You will also need to rework the static casts as they will need adjustment as comparisons between types will not be possible.

Why not this, simply, and similar for `int16_t`? What would be the issue with comparisons between `int32_t` and `int16_t`?
```c++
constexpr int32_t operator""_mph32(long double speedMph)
{
    return static_cast<int32_t>(speedMph * 65536); // 65536 = 2^16
}
```

sohamroy19:
Having to use the more specific `_mph32` and `_mph8u` and the absence of a default `_mph` would prevent accidental type intermixing. Also, there is a lot of intermixing of `abs()` and `std::abs()` in the code base.

duncanspumpkin:
bpb is still failing. Will have to go over it with a fine toothcomb.

sohamroy19:
This should be it! Took a long route to implement something relatively simple, but learnt a fair bit, thank you :)

tupaschoal:
Since the repo for this PR was deleted, I'll close this one and work on the one opened up with this commit there.

PR #15519, opened 2021-10-03
Fix #15322: Circus Music is Not Playing (Full Fix)

### LINKED ISSUE #15322: Circus does not have any sound

seanmajorpayne:
@Gymnasiast Updated the import files for RCT 1 & 2. Circus music is now set.

Tests for both games:
RCT2
1) Load the European Cultural Festival (WW)
2) Scroll to the Circus, wait a few seconds for the show to officially start (music will play)
3) Delete the circus and try building a new one
4) Set the ride to test (music will play) or open (when ride starts, music will play)

RCT1
1) Load the Three Monkeys Park (CF)
2) Create a circus & test ride (music will play) or open (when ride starts, music will play)

**Question**:
I did not see where the music IDs are defined for a particular ride, so is there an enum that can be used rather than using an integer?


Gymnasiast (changes requested):
This set the music to index 0 - MUSIC_STYLE_DODGEMS_BEAT. (Defined in Ride.h from line 753 onwards.) Btw, remember that in the new save format, this enum is no longer valid - the user can choose the desired music objects there.

This might _work_, but doesn’t really solve the problem. It should be able to do circus music even if the style is set to OBJECT_ENTRY_INDEX_NULL.

seanmajorpayne:
Okay I took a closer look.

Here is what is happening:

Prior to this PR, spawning a circus works because RideCreateAction.cpp sets the music_tune_id to 255. When ride_update_music() in Ride.cpp is called, it checks to see if a music_tune_id is set to 255. If it is time to start the music, it runs the part of the function which assigns a randomized music_tune_id, which ends up assigning the Circus a music_tune_id of 0.

So when a player creates a circus, this is essentially assigning it the MUSIC_STYLE_DODGEMS_BEAT music. That means even what's currently working in develop is not correct as you've described it (and can be attributed to my last PR).

**Question:**
Do you want the circus to be considered separately from all other rides for the sound effect playing? Meaning, it will probably need it's own set of functions for managing the music.

seanmajorpayne:
Also, the S6 importer appears to assign the circus a music_tune_id of 21, which doesn't make since as that's equivalent to the "Ice" music.

IntelOrca:
@seanmajorpayne Circus was odd in that it had ride music, which was always track 21, and could not be changed. NSF does away with the hard coded tracks, since they are now music object indexes. So yeah, I think now, the ride music logic should check if the ride type is Circus, and do its own thing.

seanmajorpayne:
@IntelOrca Thank you. I believe the changes pushed now have resolved this issue. I've left some redundancy in, but want to know if this is the right idea before further optimizing. If this is not correct, I'll need to review the NSF code.

src/openrct2/ride/Ride.cpp
```diff
@@ -1753,35 +1761,79 @@ Staff* ride_get_assigned_mechanic(Ride* ride)
...
-            ride->music_tune_id = 255;
-            return;
-        }
+        ride->music_position = 0;
+        ride->music_tune_id = 255;
+        return;
     }
-    else
+
+    // Oscillate parameters for a power cut effect when breaking down
```
Gymnasiast, line 1776: From here down, it looks like the code is exactly the same as `ride_music_update()`. As such, it should be put into a function, rather than be duplicated.

Gymnasiast:
The general idea of splitting it off looks all right to me, but I would put the if check at the start of ride_music_update() - the fact that circus is a special case is not relevant to the callers of ride_music_update().

E.g.
```
void ride_music_update(Ride* ride)
{
    if (ride->type == RIDE_TYPE_CIRCUS)
    {
        circus_music_update();
        return;
    }

    // Code for regular rides.
}

**Comment by seanmajorpayne** [comment]
Okay sounds good. I will update tomorrow.

**Comment by seanmajorpayne** [comment]
Okay, relocated that function call and created separate functions for the ride breakdown effect and the sample rate calculation.

Tested on the above scenarios and also verified breakdown effect is still working for both the circus and other rides.

File: src/openrct2/ride/Ride.cpp [hunk 2]
```diff
@@ -1753,35 +1756,30 @@ Staff* ride_get_assigned_mechanic(Ride* ride)
...
-        }
-    }
+    return sampleRate;
+}
 
+/**
+ *
+ *  Ride music slows down upon breaking. If it's completely broken, no music should play.
+ */
+static bool ride_music_breakdown_effect(Ride* ride)
```
duncanspumpkin, line 1781: CamelCase

src/openrct2/ride/Ride.cpp
```diff
@@ -1825,16 +1880,7 @@ static void ride_music_update(Ride* ride)
...
-
-    // Alter sample rate for a power cut effect
-    if (ride->lifecycle_flags & (RIDE_LIFECYCLE_BREAKDOWN_PENDING | RIDE_LIFECYCLE_BROKEN_DOWN))
-    {
-        sampleRate = ride->breakdown_sound_modifier * 70;
-        if (ride->breakdown_reason_pending != BREAKDOWN_CONTROL_FAILURE)
-            sampleRate *= -1;
-        sampleRate += 22050;
-    }
+    int32_t sampleRate = ride_music_sample_rate(ride);
```
duncanspumpkin, line 1883: `const auto`

PR #15520, opened 2021-10-03
Fix #15451: Guest List Name Filter Remains After Group Selection

### LINKED ISSUE #15451: Guest search filter still applies to "Guests thinking"
### Operating System

macOS Big Sur 11.4

### OpenRCT2 build

OpenRCT2,  (5087e77 on HEAD)

### Describe the issue

In the Guests window, the guest name search filter still applies when you access a grouped list of guests from the summary tab. However, this isn't very obvious because the search button is no longer there to indicate an active search filter. For the same reason, the filter can also not be turned off.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Open the Guests window in a park that has a fair number of guests.
2. Click the Search button, and enter any value that filters a decent portion of guests out.
3. Go to the Summary tab
4. Click on any action/thought group to view the list of guests currently doing/thinking that - a smaller group near the bottom of the list is more likely to illustrate the issue better
5. The search filter from earlier will still apply to the list of guests, meaning that fewer than expected guests will be shown

### Attachments

Image | Caption
-|-
Screenshot 2021-09-21 at 02 07 45 | Filtering the guest list with "Chris"
Screenshot 2021-09-21 at 02 09 25 | Yields 3 matching guests; 1 guest is thinking "I want to go home"
Screenshot 2021-09-21 at 02 08 21 | On the Summary tab, 4 guests in the park overall are thinking "I want to go home"
Screenshot 2021-09-21 at 02 08 27 | Clicking on it shows only 1 guest instead of 4.<br/>Note that the Search button has vanished, so it's not obvious that the filter still applies - nor is there any way to remove it

seanmajorpayne:
Fixes #15451

Expected Behavior: 
1) Open the guest list
2) Apply a search filter by name (ex. Chris)
3) Click the summarized tab
4) Select one of the groups
5) The search filter will still be active (but now there is a pressed icon which can be deactivated to remove the filter)

Previously, the search filter would still be applied, but no hourglass icon would appear so the user would be unaware the filter was still applied.

Alternatively, we could make it so selecting a group deactivates the filter, but this seems like a more robust way for players to work with the guest list.

### DIFF
--- a/src/openrct2-ui/windows/GuestList.cpp
+++ b/src/openrct2-ui/windows/GuestList.cpp
@@ -321,8 +321,11 @@ public:
                 _selectedPage = 0;
                 _numPages = 1;
                 widgets[WIDX_TRACKING].type = WindowWidgetType::Empty;
-                widgets[WIDX_FILTER_BY_NAME].type = WindowWidgetType::Empty;
-                if (_selectedTab == TabId::Individual)
+                if (_selectedTab == TabId::Summarised)
+                {
+                    widgets[WIDX_FILTER_BY_NAME].type = WindowWidgetType::Empty;   
+                }
+                else if (_selectedTab == TabId::Individual)
                 {
                     widgets[WIDX_TRACKING].type = WindowWidgetType::FlatBtn;
                     widgets[WIDX_FILTER_BY_NAME].type = WindowWidgetType::FlatBtn;
@@ -576,6 +579,10 @@ public:
                     _selectedTab = TabId::Individual;
                     widgets[WIDX_TRACKING].type = WindowWidgetType::FlatBtn;
                     Invalidate();
+                    if (!_filterName.empty())
+                    {
+                        widgets[WIDX_FILTER_BY_NAME].type = WindowWidgetType::FlatBtn;
+                    }
                     scrolls[0].v_top = 0;
                     RefreshList();
                 }


groenroos:
Having a button to disable the filter on the guest list is certainly an improvement (and should make it clearer that the filter is still applying, which is what confused me at first).

However, you still have the slight mismatch of clicking on something saying e.g. "4 guests" on the summarized group, but then getting a list that has fewer guests than that. I think a robust solution would still include the full guest name filtering button and functionality on the summarized tab, and thus hide the summarized groups that have 0 matching guests, and display the filtered number of guests on each visible summarized group.

On the other hand, I can't think of a good use case of why you'd want to search by name on the summarized tab - so I think to me the better solution would be to turn off the filter when you click on a summarized group.

seanmajorpayne:
I can update this tomorrow to remove the filter when the summarized tab is clicked. I also can't think of a good use case where you'd need to filter by guest before getting the groups.

seanmajorpayne:
The filter is now removed when the summarized tab is clicked. If a group is selected from the summarized tab, the search widget is now available for additional filtering within the group. The group will remain active even when a search is performed. This is more obvious for the player, since there is a clear message in the box that a particular group filter is active. Clicking back on the summarized tab will remove the group filter (original behavior).

Thoughts?

PR #15523, opened 2021-10-03
Fix #15377: Station Hover doesn't show entrance/exit ghost

### LINKED ISSUE #15377: Ride entrance placement seems to remember last hovered station
### Operating System

Windows 10 x64

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (a399108 on develop) provided by GitHub

### Describe the issue

When placing ride entrances and exits, it seems like the game remembers the last station you hovered over, and only tries to place the entrance for that one. Until you hover over another station, it simply won't place the ghost.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Build a tracked ride with two entrances
2. Try placing an entrance at one, then move the mouse to the other, like in the video
3. Notice how the game won't let you place one until you hover over the other entrance

### Attachments

https://user-images.githubusercontent.com/9705046/132563839-ab02ea03-4610-4860-84ca-40474d377fde.mp4

seanmajorpayne:
Fixes #15377 

Station ghosts were not appearing on track rides unless you touched the current station first before hovering over adjacent tiles. Now ghosts show on any valid ride station anytime the adjacent tiles to a station are hovered over. Removed code which differentiated entrance/exit placements for flat rides vs. track rides and set the station to always update to the nearest adjacent station to a tile.

Works even if stations are perpendicular and touching. Tried multiple ride types and do not see any issues.

https://user-images.githubusercontent.com/4562746/135750593-277bd82d-7074-4b12-8969-2579c776f0ba.mp4


src/openrct2/ride/RideConstruction.cpp
```diff
@@ -1302,132 +1265,67 @@ CoordsXYZD ride_get_entrance_or_exit_position_from_screen_position(const ScreenC
...
-                break;
-            bool goToNextTile = false;
-
+                continue;
             do
             {
                 if (tileElement->GetType() != TILE_ELEMENT_TYPE_TRACK)
                     continue;
-                if (tileElement->AsTrack()->GetRideIndex() != gRideEntranceExitPlaceRideIndex)
+                if (tileElement->GetBaseZ() != stationBaseZ)
```
ZehMatt, line 1297: I think it would worth introducing ```const auto* trackElement = tileElement->AsTrack()``` here instead of calling AsTrack all over the place.

ZehMatt (changes requested):

src/openrct2/ride/RideConstruction.cpp
seanmajorpayne: Sounds good. Updated

seanmajorpayne: Actually, is there a reason this needs to be const auto* and not TrackElement*? There are also redundant calls in the first block of code in that function that use ->AsTrack() multiple times, so it makes sense to define a trackElement as a nullptr at the top of the function and use it for all of these instances. See the updated changes as an example.

Broxzier: There is no need to specify the exact type. By using `AsTrack`, you already specify that whatever is returned has an interface to a track element. Using `auto` avoids having to change the type here when things like iterators or child classes are to be returned later on.

seanmajorpayne:
I'll try putting it back to auto, but it gave me an error before.

src/openrct2/ride/RideConstruction.cpp
seanmajorpayne: Thank you for explaining. I've replaced this with auto. const doesn't seem to work here because the tileElement will change within a loop.

ZehMatt: > Thank you for explaining. I've replaced this with auto. const doesn't seem to work here because the tileElement will change within a loop.

My bad for the const, didn't pay enough attention. But its well advised to always use const when possible.

seanmajorpayne: Good to know. I am relatively new to C++. It's quite a bit more complicated than Python, so I appreciate the feedback.

ZehMatt (changes requested):
I've tested the functionality and it works, I think its a great little improvement. Code wise, see comments.

tupaschoal:
@seanmajorpayne seems to be a small conflict in the changelog, can you please rebase and solve it?

Thanks!

ZehMatt:
We usually like to see it rebased and not have develop merged in, could you correct that?

PR #15515, opened 2021-10-02
Fix #15514: Two 'quit to menu' menu items available in track designer

Broxzier:

PR #15526, opened 2021-10-03
Show "Update available" in all game-options dropdowns

Broxzier:
Something I noticed when working on #15515: the "Update available" item in the game-options dropdown would only appear in scenario play - it would not appear in the scenario editor, track designer and track manager.


Gymnasiast:
I would personally place it above `Quit to menu/Exit`.

sta5is:
Would it be better if it was placed in the main menu below the 4 buttons? It would probably be better if the user gets the notification upon starting instead of going into a map first.

Gymnasiast:
> 
> 
> Would it be better if it was placed in the main menu below the 4 buttons? It would probably be better if the user gets the notification upon starting instead of going into a map first.

There is already a notification there.

Gymnasiast (approved):

PR #15530, opened 2021-10-04
Refactor action result passing for WallPlaceAction

ZehMatt:


src/openrct2/actions/WallPlaceAction.cpp
```diff
@@ -361,12 +348,13 @@ GameActions::Result::Ptr WallPlaceAction::Execute() const
             banner->ride_index = rideIndex;
             banner->flags |= BANNER_FLAG_LINKED_TO_RIDE;
         }
-
-        res->bannerId = banner->id;
     }
 
     auto* wallElement = TileElementInsert<WallElement>(targetLoc, 0b0000);
-    Guard::Assert(wallElement != nullptr);
```
ZehMatt, line 369: I also added a more sensible error here.

src/openrct2/actions/WallPlaceAction.cpp
```diff
@@ -21,25 +21,8 @@
...
-WallPlaceActionResult::WallPlaceActionResult(GameActions::Status err, rct_string_id msg)
-    : GameActions::Result(err, STR_CANT_BUILD_THIS_HERE, msg)
-{
-}
-
-WallPlaceActionResult::WallPlaceActionResult(GameActions::Status error, rct_string_id msg, uint8_t* args)
-    : GameActions::Result(error, STR_CANT_BUILD_THIS_HERE, msg, args)
-{
-}
+// STR_CANT_BUILD_THIS_HERE
```
duncanspumpkin, line 25: Remove comment

src/openrct2/actions/WallPlaceAction.h
```diff
@@ -15,19 +15,13 @@
...
 {
-public:
-    WallPlaceActionResult();
-    WallPlaceActionResult(GameActions::Status err);
-    WallPlaceActionResult(GameActions::Status err, rct_string_id msg);
-    WallPlaceActionResult(GameActions::Status error, rct_string_id msg, uint8_t* args);
-
-    TileElement* tileElement = nullptr;
-    BannerIndex bannerId = BANNER_INDEX_NULL;
+    TileElement* Element = nullptr;
```
duncanspumpkin, line 20: Same as the other action I don't think we should be passing about a pointer. Instead pass back the height.

duncanspumpkin (changes requested):
See comments

src/openrct2/actions/WallPlaceAction.cpp
```diff
@@ -386,12 +372,15 @@ GameActions::Result::Ptr WallPlaceAction::Execute() const
...
     wallElement->SetGhost(GetFlags() & GAME_COMMAND_FLAG_GHOST);
 
-    res->tileElement = wallElement->as<TileElement>();
-
     map_animation_create(MAP_ANIMATION_TYPE_WALL, targetLoc);
     map_invalidate_tile_zoom1({ _loc, wallElement->GetBaseZ(), wallElement->GetBaseZ() + 72 });
 
     res->Cost = wallEntry->price;
+
+    auto* tileElement = wallElement->as<TileElement>();
```
duncanspumpkin, line 380: No longer required you can just use wallElement

duncanspumpkin (approved):

PR #15536, opened 2021-10-06
Fix #15197: Cannot place flat ride after removing it in construction window

### LINKED ISSUE #15197: Right-clicking flatride and bulldozing while having hut placement selected causes ride construction window to 'freeze'
### Operating System

Windows 10, 64-bit

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (5087e77) provided by GitHub

### Describe the issue

After having placed a flatride the cursor automatically switches to the hut placement of the ride. When I now right-click the ride again and delete it (e.g. to relocate it) the cursor is still in the hut placement mode and you cannot place the ride anymore making the ride construction window not working anymore and you have to reload by selecting the flat in the ride selection again.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. place flatride
2. right click flatride
3. press demolish button
4. cannot place ride again

### Attachments

grafik

grafik

seanmajorpayne:
Fixes #15197 

Demolishing a flat ride (using the bulldozer icon in the construction menu) will now reopen the construction window to it's default state.

As a side note, when the ride menu is open, if you right click on the ride, the construction window opens. If you click the bulldozer icon, the ride will delete, but the ride window is still open. This happens on the develop version, not just this branch. Should that have a separate issue or should I try to roll a fix into this issue?

### DIFF
--- a/src/openrct2-ui/windows/RideConstruction.cpp
+++ b/src/openrct2-ui/windows/RideConstruction.cpp
@@ -1931,6 +1931,14 @@ static void window_ride_construction_mouseup_demolish(rct_window* w)
         const rct_preview_track* trackBlock = ted.Block;
         newCoords->z = (tileElement->GetBaseZ()) - trackBlock->z;
         gGotoStartPlacementMode = true;
+
+        // When flat rides are deleted, the window should be reset so the ride can be placed again.
+        const auto rideId = w->rideId;
+        auto ride = get_ride(rideId);
+        if (ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE))
+        {
+            ride_initialise_construction_window(ride);
+        }
     }
 
     auto trackRemoveAction = TrackRemoveAction(
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -22,6 +22,7 @@
 - Fix: [#15170] Plugin: incorrect label text alignment.
 - Fix: [#15184] Crash when hovering over water types in Object Selection.
 - Fix: [#15193] Crash when rides/stalls are demolished.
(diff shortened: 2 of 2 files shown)


src/openrct2-ui/windows/RideConstruction.cpp
```diff
@@ -1931,6 +1931,15 @@ static void window_ride_construction_mouseup_demolish(rct_window* w)
         const rct_preview_track* trackBlock = ted.Block;
         newCoords->z = (tileElement->GetBaseZ()) - trackBlock->z;
         gGotoStartPlacementMode = true;
+
+        // When flat rides are deleted window should be reset
+        const auto rideId = w->rideId;
+        auto ride = get_ride(rideId);
+        if (ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_FLAT_RIDE)
+            && !ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_IS_SHOP))
```
Gymnasiast, line 1939: I think you can drop the `!ride->GetRideTypeDescriptor().HasFlag(RIDE_TYPE_FLAG_IS_SHOP)` check here - even if shops would somehow end up in place entrance/exit mode, you _want_ them to snap out of it.

Gymnasiast: Very good. I have already tested your changes, they seem to solve the problem.

Gymnasiast:
> As a side note, when the ride menu is open, if you right click on the ride, the construction window opens. If you click the bulldozer icon, the ride will delete, but the ride window is still open. This happens on the develop version, not just this branch. Should that have a separate issue or should I try to roll a fix into this issue?

Please make a separate PR for that. It doesn’t necessarily need its own _issue_, though.

PR #15538, opened 2021-10-06
Implement multithreaded drawing

ZehMatt:
There are a couple of issues that needs fixing before this is ready.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -32,6 +32,7 @@
 - Fix: [#15496] Crash in paint_swinging_inverter_ship_structure().
 - Fix: [#15514] Two different “quit to menu” menu items are available in track designer and track design manager.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
+- Improved: [#15538] Software rendering can now draw in parallel when Multithreading is enabled.
 - Change: [#8601] Revert ToonTower base block fix to re-enable support blocking.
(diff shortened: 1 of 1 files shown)


IntelOrca:
Temporary solution, we should remove the globals altogether in the future and pass them as parameters.

I am curious though, how many extra FPS do you get with this?

ZehMatt:
I have no idea if I broke benchgfx somehow since everything works fine in-game, the reason I say it is because of:
```develop:
> openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.072873s, 14 FPS
Zoom[1] average: 0.031728s, 32 FPS
Zoom[2] average: 0.014515s, 69 FPS
Total average: 0.039705s, 25 FPS
Time: 2.38233s

PR:
> openrct2 benchgfx parks/bpb.sv6
Engine: Software
Render Count: 60
Zoom[0] average: 0.015528s, 64 FPS
Zoom[1] average: 0.007817s, 128 FPS
Zoom[2] average: 0.004416s, 226 FPS
Total average: 0.009254s, 108 FPS
Time: 0.55522s```

ZehMatt:
> Command line giant screenshot also works and is quicker I think when multithreading is on.

I think we could force it on for some things like that, the benchgfx also currently relies on the config.ini which should be addressed.

Gymnasiast:
(Both on your branch)
With multithreading off: 32
With it on: 67.

PR #15549, opened 2021-10-08
Use std::stol over C strtol on ClearScenery.cpp

janisozaur:


janisozaur:
@Broxzier I disagree on your review. I think you may have read the change as trying to fix the `std::string`, but it is fixing the `end` value instead.

> The function `strtol` sets the pointer pointed to by `str_end` to point to the character past the last character interpreted. If str_end is a null pointer, it is ignored.

The problem here it would've been set to the end of a temporary and when used one line below, it would have been a wild pointer already.

tupaschoal:
The lifetime issue had already been solved on #16559 20d ago, I reused this PR to tend to the suggestions to use the C++ version of the function

Broxzier (approved):

PR #15555, opened 2021-10-10
Port over research == operator from NSF

duncanspumpkin:
Another little standalone refactor from the NSF.


Gymnasiast:
Nice, I wanted to port this too.

PR #15561, opened 2021-10-11
Refactor VehiclePaint.cpp

frutiemax:


ZehMatt:
Since you are touching the functions could you also please use PascalCase for the names.

frutiemax:
I'm not too sure where this is going actually...

PR #15566, opened 2021-10-12
Fix #15567: Litter age underflow

### LINKED ISSUE #15567: Litter age is calculated wrong for park ratings
### Operating System

*

### OpenRCT2 build

*

### Describe the issue

The age for the litter is calculated incorrectly.
https://github.com/OpenRCT2/OpenRCT2/blob/b4cd8e462882ca495f602532e1cf8fd5dbf93518/src/openrct2/world/Park.cpp#L470
gCurrentTicks used to be gScenarioTicks but in both cases the sides are switched of the subtraction.

This is also a vanilla bug:
```
.text:0066A048                 mov     edx, [esi+24h]
.text:0066A04B                 sub     edx, gScenarioTicks
.text:0066A051                 cmp     edx, 1E00h
```
gCurrentTicks and gScenarioTicks are always higher than creationTick.

### Area(s) with issue?

The bug also exists in RCT2 (vanilla)

### Steps to reproduce

None

### Attachments

_No response_

ZehMatt:
Given the way it worked it never accounted litter in the rating, thats also a vanilla bug. This needs replays to be updated.

Closes #15567

### DIFF
--- a/src/openrct2/world/Park.cpp
+++ b/src/openrct2/world/Park.cpp
@@ -463,17 +463,11 @@ int32_t Park::CalculateParkRating() const
 
     // Litter
     {
-        int32_t litterCount = 0;
-        for (auto* litter : EntityList<Litter>())
-        {
-            if (litter->GetAge() < 7680)
-            {
-                // Ignore recently dropped litter.
-                continue;
-            }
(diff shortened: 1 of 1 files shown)


Broxzier:
Network version should be increased, since this can change the park rating, which influences peep generation I believe.

PR #15578, opened 2021-10-14
Automatically deduce std::array types

Broxzier:
I thought automatic type deduction for `std::array` was a C++20 feature, but apparently it was [introduced in C++17](https://en.cppreference.com/w/cpp/container/array/deduction_guides) already.

This also improves formatting of some arrays with weird or no indention.


Gymnasiast:
Mind the milestone!

PR #15582, opened 2021-10-15
Plugin API: Fix being unable to read litter information

Basssiiie:
Hey all,

After the change in #15566 I tried to update my plugin but found out the plugin api for litter entities seems to be a bit incomplete. It seems the PR that introduced litter (#15165) was only partially tested. This PR aims to fix the problems.

If you'd like to test it, you can use this script to see the difference (in a park with at least 1 litter entity):
```js
var litter = map.getAllEntities("litter")[0];
console.log(`Entity type: ${litter.type}\nLitter type: ${litter.litterType}\nCreation tick: ${litter.creationTick}`);
```
The script prints this on the most recent develop:
```
Entity type: litter
Litter type: undefined
Creation tick: undefined
```
And this on my branch:
```
Entity type: litter
Litter type: vomit_alt
Creation tick: 640
```

Full list of changes:
- Updated `creationTime` to `creationTick` to match names in `ScLitter.cpp` (it'd try to read a not existing property otherwise).
- Register `ScLitter` class in the Scripting Engine, so its getters and setters can actually be used.
- Add a nullptr check to `litterType_get()` to fix a crash there if the entity does not exist.
- Allow `getAllEntities("litter")` to return instances of litter objects instead of base entities.

Thank you for your time! :)

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -40,6 +40,7 @@
 - Fix: [#15514] Two different “quit to menu” menu items are available in track designer and track design manager.
 - Fix: [#15560] Memory leak due to OpenGL Renderer not releasing a texture.
 - Fix: [#15567] Litter not being counted correctly during Park rating calculation (original bug).
+- Fix: [#15582] [Plugin] Litter properties return incorrect values.
 - Improved: [#3417] Crash dumps are now placed in their own folder.
 - Improved: [#13524] macOS arm64 native (universal) app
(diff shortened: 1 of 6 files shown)


ZehMatt:
This requires a new plugin api version

Basssiiie:
Plugin api version is now incremented. :)

ZehMatt:
Also I forgot, the .ts file needs to be updated since you changed creationTime to creationTick

PR #15588, opened 2021-10-16
Move GameActionResult into its own unit

ZehMatt:
Some things need to be changed in order to refactor MapCanConstructWithClearAt, I don't want to include all of the GameActions in Map.hpp and once Result is no longer passed via unique_ptr it can't be forward declared, this is long overdue anyway.


IntelOrca:
.hpp is for a header file that has no corresponding source file because it is itself a source file. So it should just be GameActionResult.h.

tupaschoal:
> Thats not what hpp means, it just means header with c++.

To be fair I think there's no consensus on this. Isocpp vouches for `.h` and boost for `.hpp`

Issue #12098, opened 2020-07-02
Use ScreenRect on gfx_filter_rect


tupaschoal:
- Create an overload of `gfx_filter_rect` with the same name, but that takes a `const ScreenRect&` object instead of `left`, `top`, `right`, `bottom`
- Make the old function call the new one to avoid code duplication
- Replace the calls to the one using the 4 variables by the one using `ScreenRect`.
- If no calls to the old one are left, remove it.

*Note*: Don't try to update all calls at once, unless it doesn't occur a lot, as this might lead to a huge PR.

PR #15605, opened 2021-10-18
Part of #12098: Change several gfx_filter_rect calls to use ScreenRect

vijfhoek:
Another few "refactors" of gfx_filter_rect calls!

I've extracted the initialisation to a separate variable depending on whether clang-format would add a newline after the opening parentheses, since I find a separate variable to be much easier to read

On another note, while testing I've found some off-by-one errors in the LoadSave and RideList files, where adding or subtracting 1 from y values makes them more consistent with other scrollpaints. Do you feel it's okay to include these in this PR, or should I start a new one for them? We're talking 3 lines of codes here (LoadSave.cpp:762, RideList.cpp:591 and RideList.cpp:743).


tupaschoal:
Given that #15603 was merged, I'll close this PR.

PR #15623, opened 2021-10-20
Fix #14482: Rides with Crooked House hack sometimes misbehave

### LINKED ISSUE #14482: Ride that has been changed to Crooked House to make it invisible crashes after reloading save
**OS:** [Windows 10]
**Version:** [0.3.3]
**Commit/Build:** [3a51145]

I made a Twister Coaster track with large horizontal to steep transitions to use as a track for a custom ferris wheel. using the Allow arbitrary ride type changes cheat I changed it to a Crooked House to make it invisible. After reloading the save file, the large transitions no longer function correctly, they get track piece IDs that have a value around 265 and the ride vehicles fall down and crash once the ride starts running again.

The track that was affected in the save file is called _Ferris Wheel Track, and can be found on the ferris wheel on top of the mountain: 
[Mount Deurklink.zip](https://github.com/OpenRCT2/OpenRCT2/files/6320607/Mount.Deurklink.zip)

Gymnasiast:


Gymnasiast:
For me, it didn’t, but for good measure I now also update those as well.

PR #15624, opened 2021-10-20
Remove unused server_list_get_item_button

vijfhoek:
When working on updating `server_list_get_item_button` to use `ScreenCoordsXY` (as @tupaschoal suggested in their [comment](https://github.com/OpenRCT2/OpenRCT2/pull/15605#pullrequestreview-782675300) on #15605), I noticed the result of the function's only reference wasn't used anywhere.

When removing the code, I did not notice a difference in behaviour, and the git history of the file doesn't seem to explain the point of the code either.

While I was in there I also cleaned up the `clang-format off` part of the code by realigning it, and applying clang-format's rules manually where they didn't break anything.

### DIFF
--- a/src/openrct2-ui/windows/ServerList.cpp
+++ b/src/openrct2-ui/windows/ServerList.cpp
@@ -109,7 +109,7 @@ enum
     DDIDX_FAVOURITE
 };
 
-static bool _showVersionTooltip = false;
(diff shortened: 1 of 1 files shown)


vijfhoek:
Force-pushed to get rid of the formatting changes

PR #15627, opened 2021-10-21
Close #15358: Refactor name set park

### LINKED ISSUE #15358: Refactor Park Set Name
Park set name appears to be capped at 32 characters there doesn't appear to be any reason for that. Remove the limit. (note will be truncated on sv6 export as per all other user strings until nsf)

Bepso:
Changed the limit for setting the name of the park to be INT_MAX instead of the 32 char it was originally. As the comments in the issue display, there was a bug found when doing this


Gymnasiast:
Superseded by https://github.com/OpenRCT2/OpenRCT2/pull/17617.

PR #15812, opened 2021-10-28
Fix undefined behaviour in cctype functions

Broxzier:
These calls relied on undefined behaviour. Note from [cppreference](https://en.cppreference.com/w/cpp/string/byte/isalnum):

> Like all other functions from `<cctype>`, the behavior of `std::isalnum` is undefined if the argument's value is neither representable as `unsigned char` nor equal to EOF. To use these functions safely with plain `char`s (or `signed char`s), the argument should first be converted to `unsigned char`.

Multiple commits because changes are made via GitHub, squash on merge.

### DIFF
--- a/src/openrct2/scripting/bindings/network/ScPlayerGroup.cpp
+++ b/src/openrct2/scripting/bindings/network/ScPlayerGroup.cpp
@@ -56,7 +56,7 @@ namespace OpenRCT2::Scripting
         auto result = s.substr(sizeof("PERMISSION_") - 1);
         for (auto& c : result)
         {
-            c = std::tolower(c);
+            c = std::tolower(static_cast<unsigned char>(c));
         }
         return result;
     }
--- a/src/openrct2/scripting/bindings/network/ScPlayerGroup.cpp
+++ b/src/openrct2/scripting/bindings/network/ScPlayerGroup.cpp
@@ -14,6 +14,7 @@
 #    include "../../../Context.h"
 #    include "../../../actions/NetworkModifyGroupAction.h"
 #    include "../../../actions/PlayerSetGroupAction.h"
+#    include "../../../core/String.hpp"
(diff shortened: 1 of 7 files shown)


Broxzier:
`String::ToUpper` seems to already exists, and has some more logic to it than just calling `std::toupper` on each char, but it also returns a `std::string` instead.

Broxzier:
String filtering, where most of the calls were made, has been refactored. The remaining `static_cast`s are acceptable, IMO. Can you please take a look again, @duncanspumpkin and @ZehMatt?

src/openrct2/scripting/bindings/network/ScPlayerGroup.cpp
```diff
@@ -56,7 +56,7 @@ namespace OpenRCT2::Scripting
         auto result = s.substr(sizeof("PERMISSION_") - 1);
         for (auto& c : result)
         {
-            c = std::tolower(c);
+            c = std::tolower(static_cast<unsigned char>(c));
         }
         return result;
```
ZehMatt, line 61: Could just return ```String::ToLower(result)``` maybe?

ZehMatt (approved):
Looks reasonable as for the changes, left one comment where it could be improved.

duncanspumpkin:
Yup looks fine. I checked the String::ToUpper and it correctly uses ICU or Windows functions to handle utf8 upper casing. So in some ways this is probably a bug fix.

duncanspumpkin (approved):

src/openrct2/scripting/bindings/network/ScPlayerGroup.cpp
Broxzier: There is no `String::ToLower` currently, but the getter function here could make use of `String::ToUpper` instead of the loop. Changed that one anyway, left the rest as-is for a future refactor PR.

PR #15813, opened 2021-10-28
Ensure all fields of ImportResult are initialised

janisozaur:


tupaschoal:
> indeed, my bad. The field I wanted to fix was that of `outElement.offset` - would you accept default inits for
> 
> https://github.com/OpenRCT2/OpenRCT2/blob/b15dc66acb38d9002e5dea14fac116548823de4f/src/openrct2/drawing/Drawing.h#L77-L86
> 
> ?

@janisozaur can we close this PR and maybe open a "good first issue" for people to do this? Sounds reasonable

Gymnasiast:
Created https://github.com/OpenRCT2/OpenRCT2/pull/17501 to default-initialise rct_g1_element.

Issue #15620, opened 2021-10-20
Placing track designs at locations blocked by anything results in wrong error message


ZehMatt:
### Operating System

*

### OpenRCT2 build

OpenRCT v0.3.4.1

### Describe the issue

Instead of saying "Can't construct here, blocked by xyz" it always says "Land not owned".

Looks like it was introduced in 0.3.4(.1)

### Steps to reproduce

1. Pick any prebuilt coaster.
2. Place somewhere where it would normally say its blocked.

### Attachments

![openrct2_2021-10-20_21-57-579c7740a3c554466fa](https://user-images.githubusercontent.com/5415177/138155039-58bb644a-fbe1-4ca8-af61-3532e49570cc.png)

PR #15816, opened 2021-10-28
Close #15630: Refactor TrackDesign

### LINKED ISSUE #15630: Refactor TrackDesign code
The track design code is currently passing the error results by global variables. Considering how complex the code is it should be cleaned up first.

The track design placement has multiple operating modes and based on that it returns different things. Currently the results are either passed via return or static variables.

Following modes currently exist:
```
    PTD_OPERATION_DRAW_OUTLINES,
    PTD_OPERATION_PLACE_QUERY,
    PTD_OPERATION_PLACE,
    PTD_OPERATION_GET_PLACE_Z,
    PTD_OPERATION_PLACE_GHOST,
    PTD_OPERATION_PLACE_TRACK_PREVIEW,
    PTD_OPERATION_REMOVE_GHOST,
```

As an example, if the mode is using PTD_OPERATION_GET_PLACE_Z then place_virtual_track would result the z coordinate in the money32 return value, which makes things really hard to understand.

A better way to solve this is to give each operating mode its own function, so PTD_OPERATION_GET_PLACE_Z  would be something like TrackDesignGetPlacementZ, PTD_OPERATION_REMOVE_GHOST would become TrackDesignRemoveGhosts and so on. Those functions should return the specific data for the query/action, for TrackDesignGetPlacementZ the result type could be:
```cpp
struct TrackDesignPlacementZ
{
   rct_string_id ErrorMessage = STR_NONE;
   Formatter ErrorArgs;
   int32_t z{};
};
```
Since those functions call game actions, the error message and its arguments should be copied when something is wrong. When ErrorMessage is STR_NONE we can use z.

ZehMatt:
Initially I just wanted to get rid of the statics but I ended up fixing it all.

Closes #15630, #15620

### DIFF
--- a/src/openrct2-ui/windows/InstallTrack.cpp
+++ b/src/openrct2-ui/windows/InstallTrack.cpp
@@ -81,7 +81,7 @@ static void window_install_track_design(rct_window* w);
  */
 rct_window* window_install_track_open(const utf8* path)
 {
-    _trackDesign = track_design_open(path);
+    _trackDesign = TrackDesignImport(path);
     if (_trackDesign == nullptr)
     {
         context_show_error(STR_UNABLE_TO_LOAD_FILE, STR_NONE, {});
@@ -422,7 +422,7 @@ static void window_install_track_text_input(rct_window* w, rct_widgetindex widge
 
 static void window_install_track_update_preview()
 {
-    track_design_draw_preview(_trackDesign.get(), _trackDesignPreviewPixels.data());
+    TrackDesignDrawPreview(_trackDesign.get(), _trackDesignPreviewPixels.data());
 }
 
 static void window_install_track_design(rct_window* w)
--- a/src/openrct2-ui/windows/Ride.cpp
+++ b/src/openrct2-ui/windows/Ride.cpp
@@ -5325,7 +5325,8 @@ static void window_ride_measurements_design_save(rct_window* w)
 
     if (gTrackDesignSaveMode)
     {
-        auto errMessage = _trackDesign->CreateTrackDesignScenery();
+        TrackDesignState tds{};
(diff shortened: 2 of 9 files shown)


Gymnasiast:
Notes for testing:
- SV4 and SV6
- Placing on ground with only construction rights (e.g. Barony Bridge)
- Placing in parks with height limitations
- Placing in parks where trees may not be removed
- Placing when intersecting with large scenery
- Placing when intersecting with park entrance
- Placing when intersecting with other ride

ZehMatt:
> Notes for testing:
> 
> * SV4 and SV6
> * Placing on ground with only construction rights (e.g. Barony Bridge)
> * Placing in parks with height limitations
> * Placing in parks where trees may not be removed
> * Placing when intersecting with large scenery
> * Placing when intersecting with park entrance
> * Placing when intersecting with other ride

Having the track design collide with the park entrance results in "Land not owned" because its technically not inside the Park, given this is checked prior to tile elements this is something to keep in mind, I don't think that we need a specific error for this case, it would make the code a bit more complicated. As for the other tests I do get all the expected error messages and behavior now, I haven't tested any SV4 things.

Gymnasiast:
What happens if you make the land under the park entrance owned? This is what is often done by cheats.

ZehMatt:
image

Broxzier (approved):
Tested and works as expected.

Gymnasiast:
I will review this and merge it into v0.3.5 if I find no bugs.

Gymnasiast (approved):
Will merge it once you added a changelog entry.

Edit: added one myself: https://github.com/OpenRCT2/OpenRCT2/commit/b78908caa20310452340769a9ec050ae70756e66

Issue #15844, opened 2021-11-01
Tile Inspector inconsistent text colour


Broxzier:
### Operating System

Windows 10 x64

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (af78bbb on develop) provided by GitHub

### Describe the issue

This bug is mostly noticeable using a custom theme where the secondary colour is set to something dark.

### Steps to reproduce

1. Open the tile inspector
2. Select a couple different element
3. Observe how the text colour of some of the details and properties is coloured differently

### Attachments

![image](https://user-images.githubusercontent.com/9705046/139748436-a67d3c1f-2a68-4675-b702-7fb7b93a6b08.png)

chrhaase:
Hi, I'd like to look at this bug!

tupaschoal:
Alright

Issue #15257, opened 2021-08-21
Improper toolbar buttons in Scenario Editor


sta5is:
### Operating System

Windows 10, 64-bit

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (9511104 on develop) provided by GitHub

### Describe the issue

In the Scenario Editor, the chat toolbar button is present even though it is not a multiplayer session. Furthermore, the zoom toolbar buttons are present even if unchecked in the options.

![scenario editor chat zoom](https://user-images.githubusercontent.com/26368692/130334731-e7d94f9b-2b48-45a5-8453-2def03db23d3.PNG)

### Steps to reproduce

1. Open the scenario editor
2. If the chat toolbar button is checked in the options, the button for it will be present even though multiplayer cannot be used in this environment. If the zoom toolbar buttons are unchecked in the options, they will be present regardless

sta5is:
Same issues also in the Track Designer

![image](https://user-images.githubusercontent.com/26368692/130334883-fc4d7fff-6660-4dd9-8b32-e4ce8d733da8.png)

seanmajorpayne:
I will work on this if no one else has started :)

seanmajorpayne:
PR is up with a fix.

Issue #14310, opened 2021-03-13
Network window draws scroll bar incorrectly


frutiemax:
Windows 10
Version: 0.3.3
Commit cd9873a78de5191d9440608b70448f04f344464d

The network window which shows the players list draws incorrectly.

![openrct2_bugWindow](https://user-images.githubusercontent.com/12629817/111051628-8af50500-8422-11eb-88dc-29acb40744cf.PNG)
![openrct2_bugWindow2](https://user-images.githubusercontent.com/12629817/111051630-8cbec880-8422-11eb-92d8-32ce91418e0c.PNG)

sjoerddebruin:
This seems to be fixed as a result of #23621, can't reproduce on develop.

PR #15845, opened 2021-11-02
Add "User interface" as option in bug report form

Broxzier:


Gymnasiast:
I’m not sure that the average user will understand this. Heck, I’m not even why you’re adding it in the first place.

Broxzier:
@Gymnasiast I'm sure the average user that reports issues understands this. It could be written as "There is a problem with the user interface" or something to make it easier to understand?

I'm adding this because there are quite some reports being made that have to do with things that are broken in the user interface. Invalid text, weird artefacts, missing or misplaced widgets. Just to link a few to give you an impression:

 - https://github.com/OpenRCT2/OpenRCT2/issues/15439
 - https://github.com/OpenRCT2/OpenRCT2/issues/15844
 - https://github.com/OpenRCT2/OpenRCT2/issues/15257
 - https://github.com/OpenRCT2/OpenRCT2/issues/14310
 - https://github.com/OpenRCT2/OpenRCT2/issues/14135
 - https://github.com/OpenRCT2/OpenRCT2/issues/13758

ZehMatt:
I think it makes sense since we also try to keep UI away from core logic.

Gymnasiast:
I don’t see what it would tell us. It should be obvious from the description anyway, unlike "also present in vanilla" and "multiplayer-specific". We have dozens of labels we can apply to bug reports to class them, what makes this one so special that the user has to do it? It’s just another thing they will fill in incorrectly.

Broxzier:
What's "special" about this, is that it is an area that is often getting reported, and it's exactly an answer to what the question is about ("Area(s) with issue?"). Of course, reports should be obvious, but as you by no doubt are aware, that is not always the case, so this field may even help clarify what a report is about exactly.

I'm hoping that at some point, GitHub implements a feature that allows automatic labelling depending on the given answers, but unfortunately that is not yet supported.

Gymnasiast:
How will the user know whether pressing a button and the game not doing what they want is caused by the interface, or by (for example) the game actions? Hard for the user to determine, but easy for us.

The field in question is for things that are easy for the user to determine yet hard for us. Whether a problem is present on vanilla or only happening on multiplayer is much easier to determine for the user, but also a detail that they would probably leave out if not asked.

So basically, I think what you want to do is not going to provide us with anything useful, is a burden on the user, and is most likely only going to confuse everyone (bad information is worse than no information at all).

PR #15859, opened 2021-11-04
Fix #15858: Purple and Pizza cheats do not affect the named guest

### LINKED ISSUE #15858: Joanne Barton and Emma Garrell name cheats only work partially
### Operating System

Windows 10 64 bit

### OpenRCT2 build

efa40b1

### Describe the issue

A guest named Joanne Barton is supposed to get pizza and give guests are her pizza and a guest named Emma Garrell is supposed to get purple clothes and give guests are her purple clothes.

In both cases they do give the effect to guests around them, but they don't get the effect themselves.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. 
2. 
3. 

### Attachments

_No response_

duncanspumpkin:


Broxzier:
This way, they'll constantly give themselves pizza and purple shirts, instead of only when they are renamed. Is that the same behaviour as in vanilla RCT2?

Perhaps it's better to apply this once to themselves when they are renamed (in `Guest::HandleEasterEggName`), like the Melanie Warn (full happiness) Easter egg.

duncanspumpkin:
This is exactly how it was in the assembly. But yes perhaps that would be a better idea.

Broxzier:
Alright, then it's fine like this. For pizza slices, I guess this implementation makes sense since the guest hands out unlimited amount of slices, he should have one too. The shirt colouring seems unnecessary to repeat.

Broxzier (approved):

Gymnasiast (approved):

PR #15860, opened 2021-11-05
 Fix #15843; Tile Inspector resizes too small.

### LINKED ISSUE #15843: Tile Inspector can be resized too small
### Operating System

Windows 10 x64

### OpenRCT2 build

OpenRCT2, v0.3.4.1 (af78bbb on develop) provided by GitHub

### Describe the issue

The tile inspector's height can be resized smaller than it should. This used to be capped, but this seems to be broken now. Making it too small results in widgets overlapping and weird artefacts around the scrollbar.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Open the tile inspector
2. Select any element
3. Drag the bottom-right corner up, beyond the point where it should stop

### Attachments

Glitched tile inspector

rendello:
Fix for #15843.

Tile inspector could be resized too small when something was selected. Turns out, `window_tile_inspector_resize()` was needlessly replacing `w->min_height` on every call. Removing this reset (and the `w->min_width` reset for good measure) allows the window's properly computed size to be used.

This replacement had been here since before the file was moved to the current directory in 2017. It's possible I'm being naive here, but it seems to be the way.


Broxzier (changes requested):
Tested and works as expected. Please add a changelog entry for this fix.

rendello:
Botched that, please ignore.

Broxzier (approved):

distribution/changelog.txt
```diff
@@ -1,5 +1,6 @@
 0.3.5+ (in development)
 ------------------------------------------------------------------------
+- Fix [#15843] Tile Inspector can be resized too small.
```
tupaschoal, line 3:
```suggestion
- Fix: [#15843] Tile Inspector can be resized too small.
```

PR #15866, opened 2021-11-06
Fix: Possible out-of-bounds read on group encoding

Broxzier:
This fills the array and ensures that the size matches the object type count. This array is used in the [`SaveObject` function](https://github.com/OpenRCT2/OpenRCT2/blob/77141f57b0a91aa4321bc6908027168fe883a784/src/openrct2/object/ObjectRepository.cpp#L520) without any checks beforehand on the object type.
https://github.com/OpenRCT2/OpenRCT2/blob/77141f57b0a91aa4321bc6908027168fe883a784/src/openrct2/object/ObjectRepository.cpp#L520

I'm not 100% sure how this is supposed to work for new object types (are Sawyer chunks even used for new object types?), hence this PR rather than a direct commit on the NSF branch. The only current call stack comes from `S6Importer`, so I guess that's why it never caused any issues.

### DIFF
--- a/src/openrct2/object/ObjectList.cpp
+++ b/src/openrct2/object/ObjectList.cpp
@@ -44,21 +44,26 @@ int32_t object_entry_group_counts[] = {
 static_assert(std::size(object_entry_group_counts) == EnumValue(ObjectType::Count));
 
 // 98DA2C
-// clang-format off
 int32_t object_entry_group_encoding[] = {
-    CHUNK_ENCODING_RLE,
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
This will be unused. It's for dat file import export. The new object types aren't dats

Broxzier:
That makes sense for how it works now. This only gets called for saving embedded objects from opening S6 saves. How are object files that are embedded in .park files saved?

duncanspumpkin:
Afaik they are zipped.

PR #15869, opened 2021-11-07
Introduce templated identifier type

ZehMatt:
We are basically re-inventing the wheel for each time we want to have a strong type for different indices based things.

Here is an overview to what this type promises:
```cpp
using EntityId = TIdentifier<
    uint16_t, 
    std::numeric_limits<uint16_t>::max(), 
    struct EntityIdTag
    >;

using PatrolAreaId = TIdentifier<
    uint32_t, 
    std::numeric_limits<uint32_t>::max(), 
    struct PatrolAreaIdTag
    >;

int main()
{
    // Won't compile.
    EntityId id0 = EntityId(0);

    // Construction from underlying value, explicit
    EntityId id = EntityId::FromUnderlying(0);

    // Nulled type
    EntityId idNull = EntityId::GetNull();

    // Check against null value.
    auto isNull2 = idNull.IsNull();

    // Copy
    EntityId id2 = id;

    // Get underlying value
    printf("%u\n", id2.ToUnderlying());
    
    // Won't compile.
    PatrolAreaId id3 = id2;
}
```
Since this type is basically an enum class with members it helps avoid using static_cast all over the place by using FromUnderlying/ToUnderlying. Another great benefit is that the type holds the Null value, we don't need to define this separately anymore. This type also has a member "IsNull" to reduce even more boilerplate code.

I've picked ParkEntranceIndex as its not heavily used for a smooth introduction of the type.

### DIFF
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -1738,7 +1738,7 @@ static int32_t guest_path_find_park_entrance(Peep* peep, uint8_t edges)
     // If entrance no longer exists, choose a new one
     if ((peep->PeepFlags & PEEP_FLAGS_PARK_ENTRANCE_CHOSEN) && peep->ChosenParkEntrance.ToUnderlying() >= gParkEntrances.size())
     {
-        peep->ChosenParkEntrance = ParkEntranceIndex::Null;
+        peep->ChosenParkEntrance = ParkEntranceIndex::GetNull();
         peep->PeepFlags &= ~(PEEP_FLAGS_PARK_ENTRANCE_CHOSEN);
     }
 
@@ -1758,7 +1758,7 @@ static int32_t guest_path_find_park_entrance(Peep* peep, uint8_t edges)
             entranceNum++;
         }
 
-        if (chosenEntrance == ParkEntranceIndex::Null)
+        if (chosenEntrance.IsNull())
             return guest_path_find_aimless(peep, edges);
 
         peep->ChosenParkEntrance = chosenEntrance;
--- a/src/openrct2/peep/GuestPathfinding.cpp
+++ b/src/openrct2/peep/GuestPathfinding.cpp
@@ -1736,15 +1736,15 @@ static int32_t guest_path_find_leaving_park(Peep* peep, uint8_t edges)
 static int32_t guest_path_find_park_entrance(Peep* peep, uint8_t edges)
 {
     // If entrance no longer exists, choose a new one
-    if ((peep->PeepFlags & PEEP_FLAGS_PARK_ENTRANCE_CHOSEN) && peep->ChosenParkEntrance >= gParkEntrances.size())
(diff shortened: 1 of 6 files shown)


src/openrct2/peep/Peep.h
```diff
@@ -32,7 +33,7 @@ namespace GameActions
 {
     class Result;
 }
-using ParkEntranceIndex = uint8_t;
+using ParkEntranceIndex = TIdentifier<uint8_t, std::numeric_limits<uint8_t>::max(), struct ParkEntranceIndexTag>;
```
Broxzier, line 36: It's a bit odd that this is defined in two places. I think it's better to just include Entrance.h in this file.

src/openrct2/world/Entrance.h
```diff
@@ -44,8 +45,7 @@ extern CoordsXYZD gParkEntranceGhostPosition;
 
 constexpr int32_t MaxRideEntranceOrExitHeight = 244 * COORDS_Z_STEP;
 
-using ParkEntranceIndex = uint8_t;
-constexpr const ParkEntranceIndex PARK_ENTRANCE_INDEX_NULL = 255;
+using ParkEntranceIndex = TIdentifier<uint8_t, std::numeric_limits<uint8_t>::max(), struct ParkEntranceIndexTag>;
```
Broxzier, line 48: The other definition is here.

src/openrct2/peep/GuestPathfinding.cpp
```diff
@@ -1736,15 +1736,15 @@ static int32_t guest_path_find_leaving_park(Peep* peep, uint8_t edges)
...
     {
-        peep->ChosenParkEntrance = PARK_ENTRANCE_INDEX_NULL;
+        peep->ChosenParkEntrance = ParkEntranceIndex::Null;
         peep->PeepFlags &= ~(PEEP_FLAGS_PARK_ENTRANCE_CHOSEN);
     }
 
     if (!(peep->PeepFlags & PEEP_FLAGS_PARK_ENTRANCE_CHOSEN))
     {
-        uint8_t chosenEntrance = PARK_ENTRANCE_INDEX_NULL;
+        ParkEntranceIndex chosenEntrance = ParkEntranceIndex::Null;
```
Broxzier, line 1747: This is a good opportunity to use `auto`.

ZehMatt: The type has to be ParkEntranceIndex and not ParkEntranceIndex::ValueType to be able to use FromUnderlying in the loop.

src/openrct2/peep/Peep.h
ZehMatt: I avoided including it, Entrance.h is heavy due to Map.h

Broxzier: Include with "../ride/Station.h", and forward declare `struct CoordsXYE`, and you can remove Map.h.

duncanspumpkin: perhaps we should have a header that is just all the identifier types.

ZehMatt: Makes sense, ill revert the commit

duncanspumpkin (approved):
Idea seems sensible to me

PR #15872, opened 2021-11-07
Fix changelog order and quotation style

Broxzier:
The order is now from most to least interesting for the players: Feature - Improve - Change - Fix - Removed - Technical

This also makes the usage of the quotation marks consistent. Single quotes seem to be the way to go for British spelling:
https://grammar.yourdictionary.com/punctuation/rules-for-using-single-quotation-marks.html

- Right single-quotation mark (’) for possession (was ')
- Single quotation marks (‘...’) for names, strings, code, etc. (was '...', "...", and “...”)

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -250,6 +250,7 @@
 - Improved: [#12806] Add Esperanto diacritics to the sprite font.
 - Improved: [#12837] Arabic text is now drawn and shaped correctly on Windows.
 - Improved: [#12890] Add stroke to lowercase ‘L’ to differentiate from capital ‘I’.
+- Change: [#12749] The required version of macOS has been lowered to 10.13 (High Sierra).
 - Fix: [#400] Unable to place some saved tracks flush to the ground (original bug).
 - Fix: [#5753] Entertainers make themselves happy instead of the guests.
 - Fix: [#7037] Unable to save tracks starting with a sloped turn or helix.
@@ -268,7 +269,6 @@
 - Fix: [#12912] Plugin: selectedCell of CustomListView is being ignored on creation.
 - Fix: [#12918] Cannot place vanilla TD6 tracks of the Hypercoaster, Monster Trucks, Classic Mini Roller Coaster, Spinning Wild Mouse and Hyper-Twister types.
 - Fix: Incomplete loop collision box allowed overlapping track (original bug).
-- Technical: [#12749] The required version of macOS has been lowered to 10.13 (High Sierra).
 
 0.3.0 (2020-08-15)
 ------------------------------------------------------------------------
@@ -289,14 +289,17 @@
 - Feature: [#12347] Periodically check for new releases on GitHub, and show a notification on the title screen.
 - Feature: [#12347] The ‘About OpenRCT2’ window now has a link to the OpenRCT2 Discord Server.
 - Feature: [#12591] Show authors of an object on the object selection dialog.
-- Improved: [#6530] Allow water and land height changes on park borders.
-- Improved: [#11390] Build hash written to screenshot metadata.
 - Improved: [#3205] Make handymen less likely to get stuck in ride queues.
+- Improved: [#6530] Allow water and land height changes on park borders.
+- Improved: [#8110] OpenRCT2 now uses a single directory name for title sequences instead of three.
+- Improved: [#11390] Build hash written to screenshot metadata.
 - Change: [#11209] Warn when user is running OpenRCT2 through Wine.
 - Change: [#11358] Switch copy and paste button positions in tile inspector.
 - Change: [#11449] Remove complete circuit requirement from Air Powered Vertical Coaster (for RCT1 parity).
+- Change: [#11517] Windows Vista is supported again (libzip regression in the previous release).
 - Change: [#11898] The ‘openrct-data-path’ command-line argument has been renamed to ‘openrct2-data-path’.
 - Change: [#11944] The ride list sort mode is now remembered for the duration of the game.
+- Change: The required version of macOS has been increased to 10.14 (Mojave) for plugin support.
 - Fix: [#1013] Negative length displayed in Ride window.
 - Fix: [#1148] Research funding dropdown not shown in finances window.
 - Fix: [#5451] Guests scream on every descent, no matter how small.
@@ -348,9 +351,6 @@
 - Fix: Brakes keep working during "Brakes failure".
 - Fix: Guests maze pathfinding prefers a specific direction (original bug).
 - Removed: [#11820] Twitch support (relied on a server that has been down for a few years).
-- Technical: [#8110] OpenRCT2 now uses a single directory name for title sequences instead of three.
-- Technical: [#11517] Windows Vista is supported again (libzip regression in the previous release).
-- Technical: The required version of macOS has been increased to 10.14 (Mojave) for plugin support.
 
 0.2.6 (2020-04-17)
 ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


distribution/changelog.txt
```diff
@@ -542,10 +544,8 @@
 - Fix: [#9411] Ad campaigns end too soon.
 - Fix: [#9476] Running `simulate` command on park yields `Completed: (null)`.
```
Broxzier, line 545: There seem to be some cases using backticks.

AaronVanGeffen:
I would argue the 'removed' lines should appear before bug fixes as well. They are relatively small in number, but important to highlight.

distribution/changelog.txt
```diff
@@ -1187,6 +1182,11 @@ This is the first fully implemented version of OpenRCT2. RCT2.EXE is no longer r
 - Fix: [#3063] Search exe directory for SSL bundle as well as CWD.
 - Fix: [#3120] Negative cash in finance window is not red.
 - Fix: [#4134] Can’t enable park-wide photo price for log flume and river rapids.
+- Technical: lodepng dropped in return for libpng.
+- Technical: SDL2 upgraded from 2.0.3 to 2.0.4.
+- Technical: argparse dropped in return for bespoke command line parsing implementation.
+- Technical: Integrated breakpad for (manual) crash reporting.
+- Removed: Anti-cheat code that detected money hack attempts.
```
pizza2004, line 1189: Order here is still wrong.

IntelOrca:
It might be good to rename it to `changelog.md` now?

Broxzier:
@AaronVanGeffen I'm fine with either order, would like to hear what others think. The reason for putting 'fix' before 'removed' is that fixed issues are things that users could have noticed before and now they're fixed enjoy them better, while removed things are usually features that barely anyone used.

@IntelOrca Sounds like a plan.

ZehMatt:
> @AaronVanGeffen I'm fine with either order, would like to hear what others think. The reason for putting 'fix' before 'removed' is that fixed issues are things that users could have noticed before and now they're fixed enjoy them better, while removed things are usually features that barely anyone used.
> 
> @IntelOrca Sounds like a plan.

I think having "remove" before "fix" looks better in my opinion.

Gymnasiast:
I’m very much in favour of having a consistent style of quotes, but I think you drew the incorrect conclusion about the usage of single quotes. The page’s examples all have to do with quoting someone.

Also, the same site also has [this page about double quotes](https://grammar.yourdictionary.com/punctuation/using-quotation-marks-when-citing-information.html), which does tell you to use double quotes when citing titles and things like that. As such, it seems to me that _double quotes_ are correct here.

I still find that site a little bit vague about some usages. If someone could provide a better source, that would be appreciated.

Broxzier:
@Gymnasiast I can't find much information about the use of single- or double quotation marks when it's not for quoting someone. I did find one now that seems closer to what we do:
5.13.1 Names and titles: http://www.eng-lang.co.uk/ogs.htm

Although, there is no mention of this being the America or British grammar.
Also, apparently double quotes seem to be getting more standardized in British spelling for quotes.

Gymnasiast:
Ah, that is a clearer resource. I’m fine with single quotation marks, then.

Gymnasiast:
Also, what do we do with "technical"? It only occurs in a few places. Shall we change that to "change"?

tupaschoal (commented):
- I'd also vouch for remove before Fix.
- Renaming it would be great to use the backticks every (few) times we mention a command

> Also, what do we do with "technical"? It only occurs in a few places, but it’s not like the others - it’s more a category than a type of change. Shall we change that to "change"?

Taking this one as reference:
```
- Technical: lodepng dropped in return for libpng.
- Technical: SDL2 upgraded from 2.0.3 to 2.0.4.
- Technical: argparse dropped in return for bespoke command line parsing implementation.
- Technical: Integrated breakpad for (manual) crash reporting.
```

We could make it:

```
Technical:
- lodepng dropped in return for libpng.
- SDL2 upgraded from 2.0.3 to 2.0.4.
- argparse dropped in return for bespoke command line parsing implementation.
- Integrated breakpad for (manual) crash reporting.
```
It also poses the question of whether we want to have this effort or not. We almost never do Technical ones, maybe because it's generally "Implementation-detail"

distribution/changelog.txt
```diff
@@ -1234,31 +1234,31 @@ This is the first fully implemented version of OpenRCT2. RCT2.EXE is no longer r
 - Feature: Option to display all scrolling text banners as upper case (RCT1 style).
 - Feature: Option to mute audio when game is not focused.
 - Feature: Option to automatically place staff after hire.
-- Feature: Option to enable 'mow grass' by default for handymen (RCT1 style).
+- Feature: Option to enable ‘mow grass’ by default for handymen (RCT1 style).
 - Feature: Option to ignore invalid checksums on loaded parks.
 - Feature: Option to scale game display for better compatibility with high DPI screens.
-- Alteration: Autosave is now measured in real-time rather than in-game date.
-- Alteration: Hacked rides no longer have their reliability set to 0.
```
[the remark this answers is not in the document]
Broxzier: They're not lost, 'Alteration' was changes into 'Change' here. I agree that 'Fix' may be better.

distribution/changelog.txt
```diff
@@ -423,23 +427,22 @@
...
 - Fix: [#10963] Light effects are drawn off-centre in some rotations.
 - Fix: [#10993] Bottom toolbar not refreshing when a guest leaves the park.
 - Fix: [#11001] Rides list does not use natural sorting.
 - Fix: [objects#54] Stage Coach cars are not considered covered by the game.
 - Fix: [objects#56] Handymen cut grass incorrectly.
-- Improved: [#682] The staff patrol area is now drawn on the water, instead of on the surface under water.
-- Improved: [#10858] Added horizontal grid lines to finance charts.
-- Improved: [#10884] Added y-axes and labels to park window charts.
-- Improved: [#10970] Introduced optional light effects for vehicles at night.
 - Removed: [#6898] LOADMM and LOADRCT1 title sequence commands (use LOADSC instead).
```
Gymnasiast, line 437: Should be above the fixes.

distribution/changelog.txt
```diff
@@ -631,14 +639,6 @@
...
 - Fix: [#8824] Invalid read in footpath_chain_ride_queue.
-- Improved: [#2940] Allow mouse-dragging to set patrol area (Singleplayer only).
-- Improved: [#7730] Draw extreme vertical and lateral Gs red in the ride window's graph tab.
-- Improved: [#7930] Automatically create folders for custom content.
-- Improved: [#7980] Show the full path of the scenario in the scenario select window.
-- Improved: [#7993] Allow assigning a keyboard shortcut for opening the tile inspector.
-- Improved: [#8107] Support Discord release of RCT2.
-- Improved: [#8491] Highlight entrance and exit with different colours in track design previews.
-- Improved: Almost completely new Hungarian translation.
 - Removed: [#7929] Support for scenario text objects.
```
Gymnasiast, line 642: Idem

distribution/changelog.txt
Gymnasiast: Still incorrect.

distribution/changelog.txt
```diff
@@ -1234,31 +1234,31 @@ This is the first fully implemented version of OpenRCT2. RCT2.EXE is no longer r
...
 - Fix: TD4 River Rapids will now have the correct seat colour (original bug).
 - Fix: Message sound will no longer play in the editor modes (original bug).
 - Fix: Scenarios created with the Scenario Editor will now have the correct initial temperature for their climate (original bug).
-- Fix: Financial information can no longer be accessed from the rides/attractions menu in parks that don't use money (original bug).
+- Fix: Financial information can no longer be accessed from the rides/attractions menu in parks that don’t use money (original bug).
 - Fix: The path tool and tracked-ride construction tool no longer interfere with one another in certain situations (original bug).
 - Fix: Building a flat ride partially out of park boundaries will no longer trigger a misleading "too high for supports" message (original bug).
-- Fix: On-ride photos are now factored into the calculations of a ride's income and profit per hour (original bug).
+- Fix: On-ride photos are now factored into the calculations of a ride’s income and profit per hour (original bug).
+- Removed: Six Flags branding and limitations.
```
Gymnasiast, line 1259: Needs to be above fixes

Gymnasiast (changes requested):

distribution/changelog.txt
Broxzier: I closed it too early because I have the change ready, just not pushed yet because I want to clear up the history and rebase it.

tupaschoal (approved):

PR #15883, opened 2021-11-10
Close OpenRCT2#13792: Refactor window to class: Park

Makke94:
Refactor of ParkWindow to make use of the new Window class.

### DIFF
--- a/src/openrct2-ui/windows/Window.h
+++ b/src/openrct2-ui/windows/Window.h
@@ -80,7 +80,6 @@ rct_window* window_finances_open();
 rct_window* window_finances_research_open();
 
 // WC_PARK_INFORMATION
-rct_window* window_park_open();
 rct_window* window_park_awards_open();
 rct_window* window_park_entrance_open();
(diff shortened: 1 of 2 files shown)


src/openrct2-ui/windows/Window.h
```diff
@@ -80,6 +80,7 @@ rct_window* window_finances_open();
 rct_window* window_finances_research_open();
 
 // WC_PARK_INFORMATION
+rct_window* window_park_open();
```
duncanspumpkin, line 83: Why is there two of these L85 as well? This isn't even called so is pointless.

Makke94: Removed

Gymnasiast:
@Makke94 Could you rebase this, and drop the `this->` part of the member access/method calls? (the latter is for consistency with the rest of the codebase)

duncanspumpkin:
I'm closing this as it has gone stale and there have been a few changes to the window that would make rebasing to be hard to achieve.

PR #15886, opened 2021-11-10
Refactor out DEFINE_GAME_ACTION

ZehMatt:
The registration is also simplified, now Intellisense is actually capable of exploring those classes.


src/openrct2/actions/GameActionRegistry.cpp
```diff
@@ -0,0 +1,399 @@
...
+            case GameCommand::PlaceFootpathAddition:
+                return "PlaceFootpathAddition";
+            case GameCommand::RemoveFootpathAddition:
+                return "RemoveFootpathAddition";
+            case GameCommand::GuestSetFlags:
+                return "GuestSetFlags";
+            case GameCommand::SetDate:
+                return "SetDate";
+            case GameCommand::Custom:
+                return "Custom";
```
Broxzier, line 360: Shouldn't this be wrapped in `#ifdef ENABLE_SCRIPTING`?

May be nicer to refactor this entire function out, and have the names be part of the action classes, perhaps supplied where registered or as a template parameter. Probably better for another PR, though.

ZehMatt: The id exists regardless, so I dunno.

src/openrct2/actions/GameActionRegistry.cpp
```diff
@@ -0,0 +1,248 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2020 OpenRCT2 developers
```
Broxzier, line 2:
```suggestion
 * Copyright (c) 2014-2021 OpenRCT2 developers
```

Broxzier (approved):

Broxzier:
All changes look good, nice to see the macro gone. Haven't tested the build though.

PR #15901, opened 2021-11-12
Add function to get all entities on a tile to plugin api

andrewpratt64:
Add function map.getAllEntitiesOnTile() to api

- This was made to allow getting entities on a tile without the need to manually iterate over all entities of a given type, such as the value returned by getAllEntities()
- getAllEntitiesOnTile() was made to closely resemble the already existing getAllEntities()
- In general tems, getAllEntitiesOnTile() is to EntityTileList<T> as getAllEntities() is to EntityList<T>
- Currently, the tile position is taken in as two integer parameters that are then used to construct a CoordsXY object. This could be changed to take a single CoordsXY parameter instead, if that would be better
- I appended `, api function to get entities on a tile` to my name in contributors.md, I can change how it's worded or remove that change entirely if it's not necessary

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -11,6 +11,7 @@
 - Feature: [#13858] Flatride bases can be made invisible.
 - Feature: [#14676] [Plugin] Allow plugins to store data in .park files.
 - Feature: [#15367] Individual track elements can now be drawn as another ride type.
+- Feature: [#15901] [Plugin] Add map.getAllEntitiesOnTile to API.
 - Feature: [#16029] [Plugin] Add TrackElement.rideType to API.
 - Feature: [#16097] The Looping Roller Coaster can now draw all elements from the LIM Launched Roller Coaster.
 - Feature: [#16132, #16389] The Corkscrew, Twister and Vertical Drop Roller Coasters can now draw inline twists.
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -604,6 +604,15 @@ declare global {
         getAllEntities(type: "staff"): Staff[];
         getAllEntities(type: "car"): Car[];
         getAllEntities(type: "litter"): Litter[];
+        getAllEntitiesOnTile(type: EntityType, x: number, y: number): Entity[];
+        /**
+         * @deprecated since version 34, use guest or staff instead.
+         */
+        getAllEntitiesOnTile(type: "peep", x: number, y: number): Peep[];
+        getAllEntitiesOnTile(type: "guest", x: number, y: number): Guest[];
+        getAllEntitiesOnTile(type: "staff", x: number, y: number): Staff[];
+        getAllEntitiesOnTile(type: "car", x: number, y: number): Car[];
+        getAllEntitiesOnTile(type: "litter", x: number, y: number): Litter[];
         createEntity(type: EntityType, initializer: object): Entity;
     }
 
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -604,15 +604,11 @@ declare global {
         getAllEntities(type: "staff"): Staff[];
         getAllEntities(type: "car"): Car[];
         getAllEntities(type: "litter"): Litter[];
-        getAllEntitiesOnTile(type: EntityType, x: number, y: number): Entity[];
-        /**
-         * @deprecated since version 34, use guest or staff instead.
-         */
-        getAllEntitiesOnTile(type: "peep", x: number, y: number): Peep[];
-        getAllEntitiesOnTile(type: "guest", x: number, y: number): Guest[];
-        getAllEntitiesOnTile(type: "staff", x: number, y: number): Staff[];
-        getAllEntitiesOnTile(type: "car", x: number, y: number): Car[];
-        getAllEntitiesOnTile(type: "litter", x: number, y: number): Litter[];
+        getAllEntitiesOnTile(type: EntityType, tile: Tile): Entity[];
+        getAllEntitiesOnTile(type: "guest", tile: Tile): Guest[];
+        getAllEntitiesOnTile(type: "staff", tile: Tile): Staff[];
+        getAllEntitiesOnTile(type: "car", tile: Tile): Car[];
+        getAllEntitiesOnTile(type: "litter", tile: Tile): Litter[];
         createEntity(type: EntityType, initializer: object): Entity;
     }
 
--- a/distribution/openrct2.d.ts
+++ b/distribution/openrct2.d.ts
@@ -604,11 +604,11 @@ declare global {
         getAllEntities(type: "staff"): Staff[];
         getAllEntities(type: "car"): Car[];
         getAllEntities(type: "litter"): Litter[];
-        getAllEntitiesOnTile(type: EntityType, tile: Tile): Entity[];
-        getAllEntitiesOnTile(type: "guest", tile: Tile): Guest[];
-        getAllEntitiesOnTile(type: "staff", tile: Tile): Staff[];
-        getAllEntitiesOnTile(type: "car", tile: Tile): Car[];
-        getAllEntitiesOnTile(type: "litter", tile: Tile): Litter[];
+        getAllEntitiesOnTile(type: EntityType, tilePos: CoordsXY): Entity[];
+        getAllEntitiesOnTile(type: "guest", tilePos: CoordsXY): Guest[];
+        getAllEntitiesOnTile(type: "staff", tilePos: CoordsXY): Staff[];
+        getAllEntitiesOnTile(type: "car", tilePos: CoordsXY): Car[];
+        getAllEntitiesOnTile(type: "litter", tilePos: CoordsXY): Litter[];
         createEntity(type: EntityType, initializer: object): Entity;
     }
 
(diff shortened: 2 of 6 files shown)


distribution/openrct2.d.ts
```diff
@@ -579,6 +579,15 @@ declare global {
         getAllEntities(type: "staff"): Staff[];
         getAllEntities(type: "car"): Car[];
         getAllEntities(type: "litter"): Litter[];
+        getAllEntitiesOnTile(type: EntityType, x: number, y: number): Entity[];
+        /**
+         * @deprecated since version 34, use guest or staff instead.
+         */
+        getAllEntitiesOnTile(type: "peep", x: number, y: number): Peep[];
```
Broxzier, line 586: Since this is a new function, it doesn't make sense to immediately deprecate it. Just remove the "peep" support.

src/openrct2/scripting/bindings/world/ScMap.cpp
```diff
@@ -167,6 +167,72 @@ namespace OpenRCT2::Scripting
...
+            }
+        }
+        else if (type == "duck")
+        {
+            for (auto sprite : EntityTileList<Duck>(pos))
+            {
+                result.push_back(GetObjectAsDukValue(_context, std::make_shared<ScEntity>(sprite->sprite_index)));
+            }
+        }
+        else if (type == "peep")
```
duncanspumpkin, line 203: Remove

distribution/openrct2.d.ts
andrewpratt64: Done: https://github.com/OpenRCT2/OpenRCT2/pull/15901/commits/e8735b161066c57ab4085323925f0e92cfdceaa4

src/openrct2/scripting/bindings/world/ScMap.cpp
andrewpratt64: Done: https://github.com/OpenRCT2/OpenRCT2/pull/15901/commits/e8735b161066c57ab4085323925f0e92cfdceaa4

distribution/openrct2.d.ts
```diff
@@ -579,6 +579,15 @@ declare global {
         getAllEntities(type: "staff"): Staff[];
         getAllEntities(type: "car"): Car[];
         getAllEntities(type: "litter"): Litter[];
+        getAllEntitiesOnTile(type: EntityType, x: number, y: number): Entity[];
```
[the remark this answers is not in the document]
andrewpratt64: Done: https://github.com/OpenRCT2/OpenRCT2/pull/15901/commits/e8735b161066c57ab4085323925f0e92cfdceaa4

src/openrct2/scripting/bindings/world/ScMap.cpp
```diff
@@ -167,6 +167,74 @@ namespace OpenRCT2::Scripting
         return result;
     }
 
+    std::vector<DukValue> OpenRCT2::Scripting::ScMap::getAllEntitiesOnTile(const std::string& type, const DukValue& tile) const
```
Broxzier, line 170: Not sure if this works, but can you take a pointer to the tile instead of a generic `DukValue` object?

```suggestion
    std::vector<DukValue> OpenRCT2::Scripting::ScMap::getAllEntitiesOnTile(const std::string& type, std::shared_ptr<ScTile> tile) const
```

andrewpratt64: I could try it, I'll have to test what happens. If it does work, would it be better to take it as a ```const ScTile&```? Since tile is never written to and the function dosen't need ownership of the tile (The C++ Core Guidelines explains what I mean better than I can: [link](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f7-for-general-use-take-t-or-t-arguments-rather-than-smart-pointers))

andrewpratt64: > Not sure if this works, but can you take a pointer to the tile instead of a generic `DukValue` object?

Looked through ScTile's source code just now, both the member variables and the accessors for it's x/y coords are private. Having getAllEntitiesOnTile take a ScTile would require changing the visibility of ScTile's coords and/or or adding a new public method to ScTile

IntelOrca: Given you only require a coordinate, wouldn't the API be better as just: `getAllEntitiesOnTile(type: string, location: CoordsXY)`?

andrewpratt64: That's true, I'll change it to that

andrewpratt64: > Given you only require a coordinate, wouldn't the API be better as just: `getAllEntitiesOnTile(type: string, location: CoordsXY)`?

Playing with it now, I tried to see if it would take an actual CoordsXY in c++ but it dosen't, so I'll leave it as a DukValue there. In Typescript, I think technically both a CoordsXY and a Tile object would work, since both have "x" and "y" values. There could potentially be a function signature in Typescript for taking a CoordsXY and another for taking a Tile that both correspond to the same function in c++ so that users don't need to construct a whole new object, but I'm not sure if it really matters.

andrewpratt64 (commented):
Added it, but other commits somehow got duplicated in the process, so ignore them for now while I try to fix that

andrewpratt64:
> Added it, but other commits somehow got duplicated in the process, so ignore them for now while I try to fix that

Fixed

src/openrct2/scripting/bindings/world/ScMap.cpp
IntelOrca: You have to use `DukValue`, but you can now do `auto coords = FromDuk<CoordsXY>(dukValue)`

andrewpratt64: Oh okay, I didn't realize there was a ```FromDuk``` function lol

andrewpratt64: Looking at how FromDuk is implemented, I'm guessing it has defaults for missing properties instead of throwing a DukException so I could remove the outer try-catch as well

IntelOrca (approved):

distribution/changelog.txt
```diff
@@ -1,5 +1,6 @@
 0.3.5+ (in development)
 ------------------------------------------------------------------------
+- Feature: [#15901] [Plugin] Add function "map.getAllEntitiesOnTile"
```
Broxzier, line 3: Since there is a new release, this has to be rebased and moved up to the current version.

src/openrct2/scripting/bindings/world/ScMap.cpp
```diff
@@ -167,6 +167,67 @@ namespace OpenRCT2::Scripting
         return result;
     }
 
+    std::vector<DukValue> OpenRCT2::Scripting::ScMap::getAllEntitiesOnTile(
+        const std::string& type, const DukValue& tilePos) const
+    {
+        // Get the tile position
+        const auto pos = FromDuk<CoordsXY>(tilePos);
```
Broxzier, line 174: I think it would be better to take TileCoords instead of CoordsXY, so each tile increments X and Y by only 1 instead of 32. To get all entities on tile `{x: 70, y: 75}` for example, you now have to supply `{x: 70 * 32, y: 75 * 32}`.

Broxzier (changes requested):
Tested, and works as expected, however I think it would be better to take tile coordinates instead of big coordinates. This has to be rebased, and the changelog entry should be moved up to the current development version.

src/openrct2/scripting/bindings/world/ScMap.cpp
Broxzier: https://github.com/Broxzier/OpenRCT2/commit/6ee6ae3906ac77037d67a778804d2c78b517ba41

ZehMatt: I agree, tiles should use tile coordinates.

tupaschoal:
@andrewpratt64 can you rebase to fix the merge conflict on `distribution.txt`, please?

src/openrct2/scripting/bindings/world/ScMap.cpp
Broxzier: The use of big coords and tile coords seems to be rather mixed in the API. I've seen a couple plugins use game-actions where they multiply every value with 32 to iterate over some tiles, and the getTile function on the map takes two arguments for x and y where they are tile coords instead.

tupaschoal: Is this solved (and thus PR mergeable?)

andrewpratt64: Well it still takes big coords instead of tile coords. I tried changing it so that the pos was multiplied by 32, but my build environment get a little messed up after the new save format update was released. Is there anything special I'm supposed to do to update it? Running from the game launcher works fine but Visual Studio logs a thousand "Object not found." errors and I *think* it tries to load the older format

duncanspumpkin: you need to do a full rebuild (from the msbuild window) to make it download the updated objects.

andrewpratt64: I think I tried that, but I'll try again. If that dosen't work I'll just delete the local repository and clone it again

Broxzier: It seems that the use of TileCoords and Coords are a bit mixed in the API. This is better left for another issue.

Broxzier (approved):

PR #15920, opened 2021-11-16
Fix #15919. Ignore research status for scenery when in scenario editor

### LINKED ISSUE #15919: Scenery gets moved to Question Mark tab in scenario editor
### Operating System

Windows 10, 64 bit

### OpenRCT2 build

0.3.5 f84c561 (Also on release)

### Describe the issue

In the scenario editor, if you go through object editor and then to the next step, a lot of the scenery will be in the Question mark tab, and several tabs will not be shown.

This probably has to do with the invention status, because when you change all objects to pre-invented at the start of the game and then go back, all objects are in their correct tabs again.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Create a new scenario
2. Skip the object selection
3. Check the scenery window: image
4. Go to next step and set all objects to invented at start of game, and then go back
5. Check the scenery window: image

### Attachments

_No response_

duncanspumpkin:


src/openrct2/management/Research.cpp
```diff
@@ -597,6 +597,12 @@ bool scenery_group_is_invented(int32_t sgIndex)
         return false;
     }
 
+    // All scenery is temporarily invented when in the scenario editor
+    if (gScreenFlags & SCREEN_FLAGS_SCENARIO_EDITOR)
```
Gymnasiast, line 601:
```suggestion
    if (gScreenFlags & SCREEN_FLAGS_EDITOR)
```

Scenery is also available in the Track Designer.

src/openrct2/world/Scenery.cpp
```diff
@@ -291,6 +292,12 @@ int32_t wall_entry_get_door_sound(const WallSceneryEntry* wallEntry)
 
 bool IsSceneryAvailableToBuild(const ScenerySelection& item)
 {
+    // All scenery can be built when in the scenario editor
+    if (gScreenFlags & SCREEN_FLAGS_SCENARIO_EDITOR)
```
Gymnasiast, line 296:
```suggestion
    if (gScreenFlags & SCREEN_FLAGS_EDITOR)
```

Gymnasiast (approved):

PR #15922, opened 2021-11-16
Fix #15138: Sometimes small scenery stays when building through it

### LINKED ISSUE #15138: Quarter tile scenery not being removed consistently when building a track or path through it if there are multiple scenery items on the tile.
**OS:** Kubuntu linux 20.04
**Version:** 0.3.4.1
**Commit/Build:** Unknown develop. See #12909 

Lately, I've been seeing that quarter-tile scenery items are sometimes not removed when I build a path or track through them. _Sometimes_ it's removed, but some get missed. It happens pretty consistently (see screenshot below). There doesn't seem to be any pattern to which scenery items get missed.

- [ no ] Reproducible in RCT2 (vanilla)?
- [ no ] Specific to multiplayer?

**Steps to reproduce:**
1. Build quarter-tile scenery, preferably a lot of it -- Edit: there must be at least two scenery items per tile for this to happen.
2. Build a path or ride track through the scenery
3. Some will be removed, but you'll probably see that some of the scenery that should be removed stays where it is.

**Screenshots / Video:**
Forest Frontiers 2021-07-30 11-27-28

Broxzier:
This happened because the tile element pointer that was passed to the `map_place_clear_func` function was a copy. The function decreases the pointer after deletion, however, since the pointer was passed as a copy, the iterating loop skipped over the next element on the tile.

This can be tested by stacking small blocks or building quarter-tile scenery pieces on the same tile, then building a path through it.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -2,6 +2,7 @@
 ------------------------------------------------------------------------
 - Improved: [#12869] The Tile Inspector window’s layout has been tweaked slightly.
 - Change: [#15899] Weird bonus for path 0 - likely intended as a queue bonus (original bug).
+- Fix: [#15138] Sometimes small scenery stays when building through it.
 - Fix: [#15620] Placing track designs at locations blocked by anything results in wrong error message.
 - Fix: [#15843] Tile Inspector can be resized too small.
 - Fix: [#15844] Tile Inspector has inconsistent text colours.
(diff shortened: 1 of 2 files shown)


Broxzier:
Backtraced this to find this PR: https://github.com/OpenRCT2/OpenRCT2/pull/15024

ZehMatt (approved):
LGTM, needs a changelog entry.

Broxzier:
> needs a changelog entry.

There is one already 😄

ZehMatt:
Did not realize it was in the same commit. 👍

PR #15929, opened 2021-11-17
Fix #15928: Tile inspector hotkeys mapped incorrectly

### LINKED ISSUE #15928: Tile inspector hotkeys mapped incorrectly
### Operating System

Windows 10, 64-bit

### OpenRCT2 build

OpenRCT2, v0.3.5 (4e8916c on develop) provided by GitHub

### Describe the issue

I use hotkeys for some functions in the tile inspector, Specifically, I use `R` for "Increase element height", and `F` for "Decrease element height".

With this version of the game, the `F` key will raise an object while `R` does nothing. I have confirmed in the settings menu that my hotkeys have not changed.

I believe this behavior was caused by commit 072597f08027624d1ea1ca0d91c95640f8ebd1c9; installing the build from the previous commit seemed to revert the hotkey behavior to normal.

#### Invalid behavior for "increase"

1. Set a hotkey for "Increase element height" to `R`
2. Open the tile inspector and select an element.
3. Press `R`.
4. Nothing happens.

#### Invalid behavior for "decrease"

1. Set a hotkey for "Decrease element height" to `R`
2. Open the tile inspector and select an element.
3. Press `R`.
4. The element moves up instead of down.

Broxzier:


duncanspumpkin (commented):
Can you make sure nsf is also updated

Broxzier:
@duncanspumpkin What would be the best approach? Applying the same changes in another commit to the NSF branch?

duncanspumpkin:
After this is merged rebase nsf and make sure it is also updated to ensure it doesn't have the same bug

PR #15935, opened 2021-11-19
Replicate vanilla RCT logic when guests specifically heading for ride

reticulatingsplines:
As discussed in #15903. In vanilla RCT, guests performing the equivalent of the ShouldGoOnRide check who are specifically "Heading For" a ride will skip the weather and stats checks and instead only check that the intensity is below 10. The code comparison looks messy but all that's being modified is the addition of an `else {` after the the `if (ride->id == GuestHeadingToRideId)` block inside `if (ride_has_ratings(ride))`, so the rest of the `if (ride_has_ratings(ride))` is only performed if the guest wasn't specifically heading for that ride.


duncanspumpkin:
How have you validated this. Did you check the assembly?

reticulatingsplines:
No, but check the discussion thread. I've observed the behavior shown in the screenshots and described below in both RCT Classic and RCT1 (RCT Deluxe, specifically) by watching dozens and dozens of guests trying to enter the queue of rides in different parks, both g-force and non-g-force, sheltered and unsheltered, stats calculated and not calculated, intensity > 10, ticket price $20, etc.

If a guest is "Heading towards" Ride X, upon reaching the queue line entrance, they will:
- Never care that it's raining nor have the related thought
- Never care that the intensity is too high/low, nor have the related thoughts, as long as it's below 10
- Never care that the nausea stat is inappropriate for their nausea tolerance, nor have the related thought
- WILL very often bounce off the queue if the ride stats aren't yet calculated, per usual
- WILL bounce off and have the related thought if it's too expensive vs. the ride value, per usual

You can watch an unsheltered ride while it's raining; guests heading towards that ride specifically (equivalent of ride->id == GuestHeadingToRideId in OpenRCT2) will enter the queue if there's space. You can forcibly create populations of guests heading towards a specific ride by starting a free ride voucher campaign, which creates guests with GuestHeadingToRideId already set. As shown in the thread, these guests don't care about anything but if forced towards a ride with ludicrous intensity via voucher campaign, they WILL still bounce off in RCT Classic.

The current code doesn't really make any sense, and I think the comment makes it clear that the current behavior in OpenRCT2 wasn't intended. Right now, if a guest is specifically heading for a ride, they are actually _less_ likely to ride it, because the function returns false due to an intensity above 10 without even factoring in happiness.

duncanspumpkin:
The assembly is jumping all over the place in this function. Will be hard to verify 0x696263 is the GuestHeadingToRideId check. Looks like after this check it has to jump and skip other checks. But there are a few things different in this code. Will need careful unpicking.

Gymnasiast (changes requested):
I think we’d at least need a good assembly check. Just changing peep AI logic based on what RCT Classic **appears** to do is not a solid foundation.

reticulatingsplines:
Obviously understand the skepticism, but I do want to point out that this change would be reflective of OpenRCT2 prior to commit 2b66253d285e38b7c3aafbc5b033e74b4fa771a9, where it seems like this logic may have been inadvertently lost when replacing the gotos.

duncanspumpkin:
I think you are correct btw it's just we want to make sure it's correct. Little busy this weekend to verify it against the assembly though.

Gymnasiast:
@reticulatingsplines Did you close this by accident?

reticulatingsplines:
Yeah, sorry, my VS github extension went absolutely haywire. I'll remake a pr.

PR #15955, opened 2021-11-22
Use correct alignment for data used in PacketHeader

janisozaur:


ZehMatt:
Whats the point doing this?

ZehMatt (changes requested):
This will actually break the master server and I don't see why we need to include padding fields when packing was specified.

PR #15965, opened 2021-11-22
Fix BannerSetStyleAction having wrong type in constructor

ZehMatt:
Would bug out if the banner index is above 0xFF


duncanspumpkin (approved):
I guess this doesn't need a changelog entry as it will only be required post nsf.

PR #15969, opened 2021-11-22
Replicate vanilla RCT logic when guests heading for ride

reticulatingsplines:
Replacement for pr #15935, which I accidentally deleted while VS was going nuts. See discussion at #15903. Ensures guests deciding whether or not to ride a ride only perform standard weather, intensity, and nausea checks if they weren't already specifically heading for a ride.


duncanspumpkin:
Do try to keep to just the one PR. It can almost always be recovered through using rebase.

duncanspumpkin:
This looks correct but i see some other odd code in here. It looks like in vanilla that peeps ignore their ride rating preferences when it is raining and the ride has > 25% covering.

rik-smeets:
This PR has been replaced with #18361.

PR #15971, opened 2021-11-22
New API features re: guest ride, guest inventory, queue time/length

reticulatingsplines:
Hey, opening a draft PR for some new API features that would help me (and hopefully others!) with some plugins I've been designing.

Here's a modified version of the devtools plugin that can be used to check out what these API calls return, if you want to review: https://github.com/reticulatingsplines/plugin-devtools/releases/tag/v0.1


distribution/openrct2.d.ts
```diff
@@ -1068,7 +1068,9 @@ declare global {
         start: CoordsXYZ;
         length: number;
         entrance: CoordsXYZD;
-        exit: CoordsXYZD;
+        exit: CoordsXYZD
```
Broxzier, line 1071: Revert this accidental line change

reticulatingsplines: oops! done.

duncanspumpkin:
I'm not sure if being able to set the headingToRideId is safe to do. Think there might be a requirement to set a few more fields when that is modified.

Broxzier:
There are three locations where headingToRideId is set:

https://github.com/OpenRCT2/OpenRCT2/blob/3f19d6f7d7c148737793f81c7f57d203cb7a7ce7/src/openrct2/management/Marketing.cpp#L146-L147

https://github.com/OpenRCT2/OpenRCT2/blob/3f19d6f7d7c148737793f81c7f57d203cb7a7ce7/test/tests/Pathfinding.cpp#L82

https://github.com/OpenRCT2/OpenRCT2/blob/cbf922f33c0ae439aa87bb5a49f97306546f4c01/src/openrct2/peep/Guest.cpp#L1818-L1820

Only `GuestIsLostCountdown` seems to be also set when changed, and `ResetPathfindGoal` gets called. May be enough doing that too, though maybe it's safer to make it a readonly property, at least for now.

duncanspumpkin:
I think doing the following would be sufficient:
```
 GuestHeadingToRideId = ride->id; 
 GuestIsLostCountdown = 200; 
 ResetPathfindGoal(); 
```

reticulatingsplines:
Thanks for the input - I went ahead and added those two additional lines to ScGuest::headingToRideId_set.

distribution/openrct2.d.ts
```diff
@@ -1468,6 +1470,73 @@ declare global {
...
+
+        /**
+         * Removes all items from a guest's inventory.
+         */
+        removeAllItems(): void;
+
+        /**
+         * ID of voucher type; returns 255 if none held.
+         */
+        voucherType: number
```
Basssiiie, line 1534: Not sure if it's alright for me to comment but I'd prefer if this would return `null` instead of a magic number if no voucher is held. It'd make more sense for Typescript/Javascript, and it's more consistent with other API properties.

oli414:
Remember to increase the API version

rik-smeets:
Is this something you want to continue working on @reticulatingsplines ?

rik-smeets:
Closed due to inactivity. Feel free to reopen if you / someone wants to continue working on this.

PR #15975, opened 2021-11-23
Fix #15974 Cannot download object from server

### LINKED ISSUE #15974: Cannot connect multiplay server if objects are missing
### Operating System

Windows 11 64bit(Host) / Ubuntu 20.04 64bit on WSL(client)

### OpenRCT2 build

OpenRCT2, v0.3.5.1-41 (75073634f on develop)

### Describe the issue

If user connecting to server does not have all custom object, connection will fail.
According to log, looks like object name sent got corrupted somehow.

### Area(s) with issue?

New Save Format (NSF), This bug is specific to multiplayer

### Steps to reproduce

1. Open multiplay server with custom object
2. Try to connect server without custom object
3. Object download will fail, resulting immediate disconnect

### Attachments

Server side log : [orct2-server-error.txt](https://github.com/OpenRCT2/OpenRCT2/files/7585693/orct2-server-error.txt)
Client side log : [orct2-client-error.txt](https://github.com/OpenRCT2/OpenRCT2/files/7585695/orct2-client-error.txt)

YJSoft:
Due to some leftover code from NSF rebase, server will parse MAPREQUEST incorrectly. so client will fail to download objects.

### DIFF
--- a/src/openrct2/network/NetworkBase.cpp
+++ b/src/openrct2/network/NetworkBase.cpp
@@ -40,7 +40,7 @@
 // This string specifies which version of network stream current build uses.
 // It is used for making sure only compatible builds get connected, even within
 // single OpenRCT2 version.
-#define NETWORK_STREAM_VERSION "1"
+#define NETWORK_STREAM_VERSION "2"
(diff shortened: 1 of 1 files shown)


duncanspumpkin (approved):
Looks sensible

PR #15990, opened 2021-11-24
Mark legacy json objects to not be loaded

duncanspumpkin:
~~This is a pretty horrible little hack that works based on identifier name length. I didn't have time to finish testing it. Will maybe need further changes~~
This uses the conversion functions to mark objects that are legacy so that they are not loaded. It slows loading by a negligble amount but with the sheer amount of people that seem to have mixed installs I think it is sensible.


src/openrct2/object/ObjectRepository.cpp
```diff
@@ -429,6 +436,28 @@ class ObjectRepository final : public IObjectRepository
         {
             conflict = FindObject(item.Identifier);
         }
+
+        // Hack to override old json objects when mixed installs happen.
+        if (conflict != nullptr)
+        {
+            if (conflict->Identifier.size() < item.Identifier.size())
+            {
+                Console::Error::WriteLine("Errasing Object conflict: '%s'", conflict->Identifier.c_str());
```
AaronVanGeffen, line 445:
```suggestion
                Console::Error::WriteLine("Erasing Object conflict: '%s'", conflict->Identifier.c_str());
```

Broxzier:
Just speculation, but won't this cause the game to rescan the objects on every boot, since it finds object files that are not in the repository? I can test this tonight.

duncanspumpkin:
Yeah it's a horrible hack just getting fed up with the number of people that have the objects installed in the wrong location causing black screen at start. I don't think it does cause a rescan each time though.

duncanspumpkin:
I think I can do this a cleaner way but I need to rework some of how the parkfile legacy import works.

Gymnasiast:
Can we not just bump the version of all the objects to 1.1 or 2.0 and refuse to load 1.0 objects? It seems cleaner to me.

duncanspumpkin:
Yes could do that would probably still want this code though for if we want legacy plugins to remap correctly

Gymnasiast:
The diff looks quite wild. It thinks you renamed ParkFile.cpp to Legacy.cpp and then removed a lot, and then created a new ParkFile.cpp. Would you be able to correct that? It’s quite important that the history of ParkFile.cpp is correct in order to fix any bugs in it.

duncanspumpkin:
Unfortunately the legacy section is larger than the main section so git calls it a transition to the larger file. No way to change that.

Gymnasiast:
Can we perhaps create a commit that only adds the new files, followed by a commit that contain the changes to existing files? That should fix it _and_ be simpler.

duncanspumpkin:
Then we lose the history on the larger section (the legacy) all of these options are quite complex and i run the risk of breaking things.

IntelOrca:
Do this in the first commit: `src/openrct2/ParkFile.h → src/openrct2/park/ParkFile.h`
Then in the second commit do the extraction to Legacy.

Surely that will work, because ParkFile will retain its history?

duncanspumpkin:
It would interpret that as a rename i think. I basically need to do the octopus merge technique if i want the correct history for both files.

duncanspumpkin:
Okay as long as the commits are not squashed this will keep both the files histories intact. Bit annoying having to have merge commits but thats the only way to do this

IntelOrca (approved):

duncanspumpkin:
Discussed with @Gymnasiast will merge (not squash merge) on green ci's.

PR #15991, opened 2021-11-24
Use TIdentifier for BannerIndex

ZehMatt:


src/openrct2-ui/windows/Sign.cpp
```diff
@@ -61,7 +61,7 @@ class SignWindow final : public Window
 
     void ShowTextInput()
     {
-        auto* banner = GetBanner(number);
+        auto* banner = GetBanner(BannerIndex::FromUnderlying(number));
```
Gymnasiast, line 64: Could you add a helper function, just like you did in Banner.cpp?

src/openrct2/ParkFile.cpp
```diff
@@ -971,9 +971,9 @@ namespace OpenRCT2
                     cs.Write(static_cast<uint32_t>(numBanners));
 
                     [[maybe_unused]] size_t numWritten = 0;
-                    for (BannerIndex i = 0; i < MAX_BANNERS; i++)
+                    for (uint16_t i = 0; i < MAX_BANNERS; i++)
```
Gymnasiast, line 974: I guess there is no way to do a `for` loop of actual `BannerIndex`es?

ZehMatt: It could be implemented but I don't think one should do arithmetic with those handles, I could probably expose the underlying type so when the type changes no code needs updating.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1382,23 +1382,23 @@ namespace RCT2
...
                             if (dstBanner == nullptr)
                             {
-                                dst2->SetBannerIndex(BANNER_INDEX_NULL);
+                                dst2->SetBannerIndex(BannerIndex::GetNull());
                             }
                             else
                             {
                                 ImportBanner(dstBanner, srcBanner);
-                                dst2->SetBannerIndex(src2->GetBannerIndex());
+                                dst2->SetBannerIndex(BannerIndex::FromUnderlying(src2->GetBannerIndex()));
```
Gymnasiast, line 1401: See previous.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1349,23 +1349,23 @@ namespace RCT2
...
```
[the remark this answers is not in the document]
ZehMatt: I can even do better and remove the entire line, ImportBanner sets the Id and the creation of the banner sets the id.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1349,23 +1349,22 @@ namespace RCT2
...
+                            auto dstBanner = GetOrCreateBanner(BannerIndex::FromUnderlying(bannerIndex));
                             if (dstBanner == nullptr)
                             {
-                                dst2->SetBannerIndex(BANNER_INDEX_NULL);
+                                dst2->SetBannerIndex(BannerIndex::GetNull());
                             }
                             else
                             {
                                 ImportBanner(dstBanner, srcBanner);
-                                dst2->SetBannerIndex(src2->GetBannerIndex());
```
duncanspumpkin, line 1368: Afaik you shouldn't delete this

ZehMatt: ImportBanner assigns the id, and CreateBanner assigns the id which was also specified as the argument, I think its safe to say this banner ends up at the specified id.

ZehMatt: The tests seem to fail specifically on Windows now, I wonder if this is related. However forcing them to have a different id/index than they are actually in seems like a bug to me.

ZehMatt: I confused this for the banner its self, this has to be this way.

duncanspumpkin (approved):

PR #15997, opened 2021-11-25
Use precompiled headers for libopenrct2 with MSVC

janisozaur:
Reduces compilation with msvc by several minutes.

There's more where this came from - in future PRs


janisozaur:
Would like to hear from @duncanspumpkin how it works for you

tupaschoal:
If we're doing something for `msvc` only, I suspect modules might also be a good option, I think it's the first compiler to fully support it and there were claims of impressive speedups shared by Microsoft at CppCon2021

IntelOrca:
I don't like precompiled headers... can we instead just work on reducing the use of common.h?

ZehMatt:
> I don't like precompiled headers... can we instead just work on reducing the use of common.h?

Whats wrong with precompiled headers? Also common.h is almost empty now, won't improve a lot.

janisozaur:
I was under impression using MSVC requires the mindset to yearn for PCHs. Like Matt said, it's not the common.h that is expensive, but ride.h and others. This gives significant speed boost in CI and I'm sure on your machine as well - so why not use it? I would also implement that for Linux, but we already have much better solution in form of ccache - I played with clcache and it hit limitations with how the project is set up (`/Zi`), so this is next best thing. I have already spent a lot of time reducing both header inclusion and their heft - while this work is never 'done', we're at a point where it offers limited gains compared to method like this one.

Time samples from before this PR (https://github.com/OpenRCT2/OpenRCT2/runs/4323822849?check_suite_focus=true):
x86: 13m 49s
amd64: 15m 15s

With this PR (https://github.com/OpenRCT2/OpenRCT2/runs/4324965912?check_suite_focus=true):
x86: 10m 13s (3m 36s or 26% shorter)
amd64: 11m 18s (3m 57s or 26% shorter)

IntelOrca:
> I was under impression using MSVC requires the mindset to yearn for PCHs.

I generally find them a nuisance. They require some odd requirements sometimes like needing it to be the first include in a source file.

ZehMatt:
> > I was under impression using MSVC requires the mindset to yearn for PCHs.
> 
> I generally find them a nuisance. They require some odd requirements sometimes like needing it to be the first include in a source file.

Yeah the include requirement is a bit annoying but that is covered by adding it to the forced includes so you wouldn't actually need to include it manually. We should be a bit careful about what we put in there still.

Broxzier:
By using a precompiled header and forcing all source files to include it, you also force a full recompilation of the project when modifying any of the header files that get included. How does it compare to build times when making an adjustment to one of the big headers?

YJSoft:
First build / After cleanup build result
Compile order is `nopch -> pch` / `pch -> nopch`
Tested env:
```
CPU : Ryzen 7 5800X
Mem : 32GB
with NVMe SSD
```

<hr />

After clone
```
Without precompiled header : 2m25s / 1m29s
With precompiled header : 1m44s / 1m16s
```

After small code change(`src/openrct2/interface/Fonts.cpp` file)
```
Without precompiled header : 38s / 38s
With precompiled header : 29s / 30s
```

After small(but affect lots of file) code change(`src/openrct2/ride/Ride.h` file)
```
Without precompiled header : 1m38s / 1m38s
With precompiled header : 1m9s / 1m7s
```

Conclusion : build time is faster with precompiled header.

janisozaur:
> Yes. It recompiles almost every code. but It's same without precompiled header.

It's probably a bit different - build systems track inclusion of files and by modifying one of the files in PCH you force rebuild of all things that include it - so possibly a superset of what you would need to rebuild otherwise. That said Ride.h is included in 402 TUs, FileSystem.hpp in 433, Object.h in 491, so virtually in all of the project

duncanspumpkin:
I've started breaking up ride.h so it's not included as much. Presumably that will reduce the benefits of this?

janisozaur:
We shall see. As it won't affect your work and provides tangible improvement now, can we meet this? Also @duncanspumpkin is there improvement in your side as well?

duncanspumpkin:
> We shall see. As it won't affect your work and provides tangible improvement now, can we meet this? Also @duncanspumpkin is there improvement in your side as well?

I'll give it a whirl tomorrow.

tupaschoal:
Have you tried modules (with c++20)? Seems to give very significant boosts for msvc, though not yet available elsewhere.

I might play with the c++20 branch to get this closer

janisozaur:
No, I haven't. Didn't you say it would limit us to MSVC only? Also got a link to that cppcon talk?

tupaschoal:
> No, I haven't. Didn't you say it would limit us to MSVC only? 

Yeah, a few ifdefs here and there, but once modules were widespread, it would be an automatic cleanup.


> Also got a link to that cppcon talk?

https://youtu.be/9OWGgkuyFV8

janisozaur:
|       | [`develop`](https://github.com/OpenRCT2/OpenRCT2/runs/5699725818?check_suite_focus=true) | [this PR](https://github.com/OpenRCT2/OpenRCT2/actions/runs/2045713374) | delta   |
|-------|-----------|---------|---------|
| x86   | 19m 38s   | 13m 29s | -6m 9s  |
| amd64 | 17m 39s   | 11m 51s | -5m 48s |

janisozaur:
This was put on hold while we try getting ccache working: see https://github.com/OpenRCT2/OpenRCT2/issues/16895

tupaschoal:
Does the update here mean we got it to work with ccache? :D

rik-smeets:
What's the status of this, @janisozaur ?

janisozaur:
I've seen @tupaschoal try ccache for Windows. This PR does not do that. This is pretty much still gated by https://github.com/OpenRCT2/OpenRCT2/pull/15997#issuecomment-979500495, but would love to get this in for the massive gains [it actually provides](https://github.com/OpenRCT2/OpenRCT2/pull/15997#issuecomment-1079777890).

ZehMatt:
What should we do about this one?

IntelOrca:
> What should we do about this one?

Look at a header include graph and reduce large ones, or split large ones into smaller ones.

ZehMatt:
> > What should we do about this one?
> 
> Look at a header include graph and reduce large ones, or split large ones into smaller ones.

This is not the same as precompiled headers and you won't really get close to that by just reducing includes. Also it would be interesting to see how much it helps with LTCG off now where it was previously always enabled.

IntelOrca:
@ZehMatt well if you say so. I am not a fan of precompiled headers, I have only had bad experiences with them.

duncanspumpkin:
I'm not a fan of them either but my issues with them are more a problem if every platform was using them.

ZehMatt:
I'm fine without them, I think we can close this PR, no need for it to collect dust.

janisozaur:
@IntelOrca what kind of bad experiences do you mean?

733737:
windows builds take 2 or 3 times as long as linux builds. I think there would be much less people complaining about a broken launcher if they weren't waiting for windows builds to compile for 20 minutes or longer

janisozaur:
Personally I'd like to have it merged as well, the windows ci times are ridiculous and this offers significant improvement

733737:
also, sometimes they time out and/or fail

IntelOrca:
> what kind of bad experiences do you mean?

In the past, I thought you had to include the precompiled header name (usually stdafx.h) at the top of every source file. It always had to be the first header too. In other words, header order and inclusion becomes ever more stricter. This doesn't appear to be required based on this PR. But I worry that, it will now allow you to use definitions that are in the precompiled header without actually including the actual required header in your source file. So as you develop code, you have more inconsistent compilation with the other platforms. It is possible developers on other platforms may also hit errors with Windows due to the PCH requirements.

From a CI perspective it looks like a win-win. But from a developer's perspective, there could be unintended consequences.

Where do you go from here if this does get merged? It reduces the time slightly, but what is your next idea for reducing times further?

janisozaur:
Next items remain unchanged: reducing amount of code compiled, reducing recompilation by caching, replacing msbuild with ninja.

How about we use PCH in CI only then?

IntelOrca:
> How about we use PCH in CI only then?

Sure

If you are going to move to ninja (I assume via cmake), why not do that first before this?

janisozaur:
@IntelOrca PCHs are now guarded with a parameter, can this still be merged for 0.4.5?

ZehMatt (approved):
I think having it only on the CI should be okay, if this is going to get in the way somehow we have the option to revert the change so from where I see it no harm is done trying it out.

PR #16004, opened 2021-11-25
Fix truncation of coordinates

ZehMatt:


duncanspumpkin:
Have you run a few tests on large maps

ZehMatt:
I did not run into any issues, just something I found while moving code.

PR #16011, opened 2021-11-26
Start splitting up ride.h

duncanspumpkin:
As ride.h is one of the slowest to compile headers it makes sense that we split it up.
This PR is removing the ride entry from the header.
Eventually I'd like to see the ride entry be defined in RideObject.h but that will be a future refactor.


src/openrct2/ride/RideColour.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include "../common.h"
```
IntelOrca, line 12:
```suggestion
#include <cstdint>
```

src/openrct2/ride/CableLift.h
```diff
@@ -10,9 +10,9 @@
 #pragma once
 
 #include "../common.h"
```
ZehMatt, line 12:
```suggestion
#include <cstdint>
```

src/openrct2/ride/RideEntry.h
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+
+#include "../rct2/RCT2.h"
```
ZehMatt, line 12: Missing cstdint include

ZehMatt (approved):

PR #16030, opened 2021-11-27
Implement #16029: Add rideType to plugin API

Sadret:


src/openrct2/scripting/bindings/world/ScTileElement.cpp
```diff
@@ -351,6 +351,27 @@ namespace OpenRCT2::Scripting
...
+            duk_push_null(ctx);
+        return DukValue::take_from_stack(ctx);
+    }
+    void ScTileElement::rideType_set(uint16_t value)
+    {
+        ThrowIfGameStateNotMutable();
+        auto el = _element->AsTrack();
+        if (el != nullptr)
+        {
+            el->SetRideType(value);
```
duncanspumpkin, line 370: Should we perhaps be doing the validation of the value. Ensure its less than type count, ensure that allow arbitrary ride type cheat enabled.

Sadret: Then we should do the same in trackType_set, right?

Sadret: In general, there isn't much validation going on in ScTileElement.
IMO, whoever is fiddling with the data directly instead of using GameActions should know what he is doing.

duncanspumpkin: As changing the RideType to a value > the type count will cause an instant crash i think we have to at least do that.

Sadret: Actually there are tons of API calls that will cause an instant crash when done with the wrong parameters.
I don't see why this one is any different.

ZehMatt: Well then its good to start adding validation code. Scripts shouldn't be able to crash the game, if there are ways to do that currently it should be fixed.

IntelOrca: > As changing the RideType to a value > the type count will cause an instant crash i think we have to at least do that.

We should fix all the instances of this.

Sadret: Added `value < RIDE_TYPE_COUNT` check.

IntelOrca:
You have two new files that should not have been committed:
* AsTrack()
* SetRideType(value)

Sadret:
Don't know where they come from. Maybe VS Code created them?
They are removed now.

duncanspumpkin (approved):

tupaschoal (commented):
@Sadret can you please rebase to solve the merge conflict with `changelog.txt`?

Sadret:
Done.
My IDE deleted a trailing whitespace in three lines of the changelog file but I guess thats ok.

tupaschoal:
I think so. @IntelOrca and @ZehMatt wanna give your blessing before merging?

IntelOrca (approved):

ZehMatt (approved):

PR #16031, opened 2021-11-27
Refactor entity painting code

ZehMatt:
Also moved some remnants to EntityBase


src/openrct2/entity/Balloon.cpp
```diff
@@ -106,3 +107,15 @@ void Balloon::Serialise(DataSerialiser& stream)
     stream << time_to_move;
     stream << colour;
 }
+
+void Balloon::Paint(paint_session* session, int32_t imageDirection) const
+{
+    uint32_t imageId = 22651 + (frame & 7);
```
duncanspumpkin, line 113: since you are refactoring. Could you add the define to the sprites.h

ZehMatt: I forgot to add a TODO comment in there.

duncanspumpkin (approved):

PR #16034, opened 2021-11-27
Introduce DATLimits.h for RCT2 Dat file object limits

duncanspumpkin:


src/openrct2/rct2/DATLimits.h
```diff
@@ -0,0 +1,21 @@
...
+ * For a complete list of all authors, please refer to contributors.md
+ * Interested in contributing? Visit https://github.com/OpenRCT2/OpenRCT2
+ *
+ * OpenRCT2 is licensed under the GNU General Public License version 3.
+ *****************************************************************************/
+
+#pragma once
+#include <cstdint>
+
+namespace RCT2::ObjectLimits
```
Broxzier, line 13: Wouldn't `DATLimits` or `DATObjectLimits` make more sense for the naming?

duncanspumpkin: Hmm perhaps. I was imagining that OpenRCT2::ObjectLimits would be the other limits file so thats what the naming was designed around.

Broxzier: Ah well, I'm fine with this too.

Broxzier (approved):

PR #16040, opened 2021-11-27
Added missing #include for limits

Alib234:


src/openrct2/drawing/Font.cpp
```diff
@@ -16,6 +16,7 @@
 #include "Drawing.h"
 #include "TTF.h"
 
+#include <limits>
```
Gymnasiast, line 19: Our code style expects include lists to be alphabetically ordered. If you move it in-between, it should be all right.

PR #16051, opened 2021-11-28
Rework rct_vehicle_colour

duncanspumpkin:
This will break replays.

### DIFF
--- a/src/openrct2/core/DataSerialiserTraits.h
+++ b/src/openrct2/core/DataSerialiserTraits.h
@@ -773,22 +773,24 @@ template<> struct DataSerializerTraits_t<TrackDesignSceneryElement>
     }
 };
 
-template<> struct DataSerializerTraits_t<rct_vehicle_colour>
+template<> struct DataSerializerTraits_t<VehicleColour>
 {
-    static void encode(OpenRCT2::IStream* stream, const rct_vehicle_colour& val)
+    static void encode(OpenRCT2::IStream* stream, const VehicleColour& val)
     {
-        stream->Write(&val.body_colour);
-        stream->Write(&val.trim_colour);
+        stream->Write(&val.Body);
+        stream->Write(&val.Trim);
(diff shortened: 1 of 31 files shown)


src/openrct2/core/DataSerialiserTraits.h
```diff
@@ -749,22 +749,24 @@ template<> struct DataSerializerTraits_t<TrackDesignSceneryElement>
...
+        stream->Read(&val.Body);
+        stream->Read(&val.Trim);
+        stream->Read(&val.Ternary);
     }
-    static void log(OpenRCT2::IStream* stream, const rct_vehicle_colour& val)
+    static void log(OpenRCT2::IStream* stream, const VehicleColour& val)
     {
         char msg[128] = {};
-        snprintf(msg, sizeof(msg), "rct_vehicle_colour(body_colour = %d, trim_colour = %d)", val.body_colour, val.trim_colour);
+        snprintf(msg, sizeof(msg), "VehicleColour(body = %d, trim = %d, tertiary = %d)", val.Body, val.Trim, val.Ternary);
```
tupaschoal, line 769: Ternary or tertiary?

Broxzier: > Ternary: Made up of three things; treble, triadic, triple, triplex
> Tertiary: Of third rank or order; subsequent. 

Tertiary seems better.

duncanspumpkin: Okay i've swapped it to tertiary. Annoyingly the plugin api uses ternary but its not the first nore the last where the plugin api uses an old name.

tupaschoal (commented):
Seems fine, though I am a bit confused as to why it breaks replays (less fields being saved/read?)

duncanspumpkin:
Yeah reorg of elements will break replays. To be expected. Will get that sorted later

duncanspumpkin:
I've remade this PR could someone review then I will record new replays.

PR #16064, opened 2021-11-29
Fix #16063: broken Object Selection preview for glass

### LINKED ISSUE #16063: Object Selection preview for objects with glass is broken
### Operating System

Ubuntu 20.04

### OpenRCT2 build

OpenRCT2, v0.3.5.1-78 (bcb527d33 on develop)

### Describe the issue

OpenRCT2:

afbeelding

Vanilla:

afbeelding

### Area(s) with issue?

_No response_

### Steps to reproduce

Open the Small Scenery section and look up any object with glass.

### Attachments

_No response_

Gymnasiast:
Looked at the Scenery window code for the fix.

### DIFF
--- a/src/openrct2/object/SmallSceneryObject.cpp
+++ b/src/openrct2/object/SmallSceneryObject.cpp
@@ -119,11 +119,7 @@ void SmallSceneryObject::DrawPreview(rct_drawpixelinfo* dpi, int32_t width, int3
 
     if (_legacyType.HasFlag(SMALL_SCENERY_FLAG_HAS_GLASS))
     {
-        imageId = _legacyType.image + 0x44500004;
-        if (_legacyType.HasFlag(SMALL_SCENERY_FLAG_HAS_SECONDARY_COLOUR))
-        {
-            imageId |= 0x92000000;
-        }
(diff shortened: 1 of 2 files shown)


IntelOrca (approved):

src/openrct2/object/SmallSceneryObject.cpp
```diff
@@ -119,11 +119,7 @@ void SmallSceneryObject::DrawPreview(rct_drawpixelinfo* dpi, int32_t width, int3
 
     if (_legacyType.HasFlag(SMALL_SCENERY_FLAG_HAS_GLASS))
     {
-        imageId = _legacyType.image + 0x44500004;
-        if (_legacyType.HasFlag(SMALL_SCENERY_FLAG_HAS_SECONDARY_COLOUR))
-        {
-            imageId |= 0x92000000;
-        }
```
duncanspumpkin, line 126: Why do we no longer require the branch

Gymnasiast: I think it might be an implementation error. The scenery window does not do this. Additionally, the frame with the secondary colour is already drawn in the step before, and it seems glass always uses the primary colour.

PR #16070, opened 2021-11-29
Fix crash after alt-tab

janisozaur:
While this should solve the problematic part, I would like to implement some kind of test - even if for this PR only - to ensure it doesn't destroy current behaviour


janisozaur:
To test disable https://github.com/OpenRCT2/OpenRCT2/blob/5e82e9048bb860f9c65becf5d13f4b23e7ebde91/src/openrct2-ui/UiContext.cpp#L742 and resize window, while lightfx is in play

IntelOrca (approved):

PR #16072, opened 2021-11-29
Replace visibility button with checkboxes per element

Broxzier:
Instead of having to select the element, players can now tick the checkbox in front of an element to toggle its visibility:

gif

### DIFF
--- a/src/openrct2-ui/windows/TileInspector.cpp
+++ b/src/openrct2-ui/windows/TileInspector.cpp
@@ -68,7 +68,6 @@ enum WindowTileInspectorWidgetIdx
     WIDX_SPINNER_Y,
     WIDX_SPINNER_Y_INCREASE,
     WIDX_SPINNER_Y_DECREASE,
-    WIDX_BUTTON_INVISIBLE,
     WIDX_BUTTON_REMOVE,
     WIDX_BUTTON_MOVE_UP,
     WIDX_BUTTON_MOVE_DOWN,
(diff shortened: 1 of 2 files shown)


src/openrct2-ui/windows/TileInspector.cpp
```diff
@@ -1343,7 +1338,14 @@ static void WindowTileInspectorScrollmousedown(rct_window* w, int32_t scrollInde
 
     // Because the list items are displayed in reverse order, subtract the calculated index from the amount of elements
     const int16_t index = windowTileInspectorElementCount - (screenCoords.y - 1) / SCROLLABLE_ROW_HEIGHT - 1;
-    WindowTileInspectorSelectElementFromList(w, index);
+    if (screenCoords.x >= 2 && screenCoords.x <= 11)
+    { // Checkbox was clicked
+        WindowTileInspectorToggleInvisibility(index);
+    }
```
Gymnasiast, line 1344: This looks a bit... scary. Is there no “cleaner” way?

Broxzier: You mean the hardcoded numbers? They're pretty much the left and right side of the checkbox, and the checkbox has 2 pixels of padding to the left. I suppose storing them in named constants would make it more clear.

Gymnasiast: I mostly meant the fact we’re now checking screen coordinates instead of actual widgets, like before. But I guess there is no other way. Adding constants would be an improvement, though.

Broxzier: These `screenCoords` are local to the list itself, starting at {0,0} in the top-left corner. All interaction with window lists use this to determine where was clicked.

Gymnasiast: What I meant was that we used to check if `WIDX_BUTTON_INVISIBLE` was clicked. Now we’re checking coordinates.

Anyway, just add a few constants and it should be good.

IntelOrca: You can't have widgets inside a scroll view. This is very similar to the edit theme window, you can't use the colour widget, so it has to check coordinates and show a colour dropdown if clicked.

Gymnasiast (approved):

sta5is:
The "Invisible" column needs to be renamed to "Rendered" or something similar since now every element that is checked is displayed as visible rather than invisible.

ZehMatt:
> The "Invisible" column needs to be renamed to "Rendered" or something similar since now every element that is checked is displayed as visible rather than invisible.

That makes sense, I would prefer "Visible" however.

Broxzier:
The tooltip needs updating too. I was thinking of replacing the `I` in the header with an eye glyph: 👁, if that works.

PR #7010, opened 2018-01-12
Add RCT1 edge styles

Gymnasiast:
SV6 reserves 4 bits for the terrain edge. This means that it can actually save 16 different land types.

This PR uses the remaining 12 slots to add the RCT1 terrain edge styles.

Things to note:
- It only adds edge styles, not surface styles
- For people who don't have RCT1, there is no visual difference, it will just draw rock.

PR #16077, opened 2021-11-30
Only add RCT1 land objects to imported SV6 if they are used

duncanspumpkin:
We were mistakenly loading rct1 surfaces and sides when loading rct2 parks leading to a confusing ui.

### DIFF
--- a/src/openrct2/rct2/S6Importer.cpp
+++ b/src/openrct2/rct2/S6Importer.cpp
@@ -1205,11 +1205,8 @@ namespace RCT2
 
                     dst2->SetSlope(src2->GetSlope());
 
-                    const auto surfaceStyle = _terrainSurfaceEntries.GetOrAddEntry(
-                        RCT2::GetTerrainSurfaceObject(src2->GetSurfaceStyle()));
-                    dst2->SetSurfaceStyle(surfaceStyle);
-                    const auto edgeStyle = _terrainEdgeEntries.GetOrAddEntry(RCT2::GetTerrainEdgeObject(src2->GetEdgeStyle()));
-                    dst2->SetEdgeStyle(edgeStyle);
+                    dst2->SetSurfaceStyle(src2->GetSurfaceStyle());
+                    dst2->SetEdgeStyle(src2->GetEdgeStyle());
 
                     dst2->SetGrassLength(src2->GetGrassLength());
                     dst2->SetOwnership(src2->GetOwnership());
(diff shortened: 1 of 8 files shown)


Gymnasiast (changes requested):
No, this is NOT a mistake. We added the RCT1 sides to SV6 files years ago, and the surfaces one year ago. So they _have_ to be loaded. Otherwise this is going to break thousands of SV6 files created with OpenRCT2 in the past years.

Gymnasiast:
Shooting the messenger now, are we? Try loading this park on develop and on this PR: [Pickle Park-savetest.sv6.zip](https://github.com/OpenRCT2/OpenRCT2/files/7625905/Pickle.Park-savetest.sv6.zip)

How it looks on develop:


How it looks with this PR:
afbeelding

pizza2004:
> We added the RCT1 sides to SV6 files years ago (#7010), and the surfaces one year ago. So they _have_ to be loaded.

Couldn't you just add a check to see if they're saved in the SV6 file and if not just not load them as defaults?

Gymnasiast:
> > We added the RCT1 sides to SV6 files years ago (#7010), and the surfaces one year ago. So they _have_ to be loaded.
> 
> Couldn't you just add a check to see if they're saved in the SV6 file and if not just not load them as defaults?

That would work, but it’s not what this PR currently does.

Gymnasiast (changes requested):
The modified version does not work correctly:

afbeelding

Oddly, it seems to add the red roof tiles (a terrain surface) as an edge style. Perhaps it’s going past the end of an array?

duncanspumpkin:
Okay think thats it fixed 
image
obviously doesn't look correct on mine since i don't have RCT1 linked but its showing as the brick edge.

duncanspumpkin:
Hmm looks like there are still a few places with hardcoded offsets:
```
static uint32_t get_surface_image(
    const paint_session* session, ObjectEntryIndex index, int32_t offset, uint8_t rotation, int32_t grassLength, bool grid,
    bool underground)
{
    // Provide fallback for RCT1 surfaces if the user does have RCT1 linked.
    if (!is_csg_loaded() && index >= TERRAIN_RCT2_COUNT)
    {
        if (index == TERRAIN_ROOF_GREY)
            index = TERRAIN_ROCK;
        else
            index = TERRAIN_DIRT;
    }
```

Gymnasiast:
> Hmm looks like there are still a few places with hardcoded offsets:

We should be able to get of those once I implement the "no csg" fallback.

Gymnasiast:
Ok, I just tested with a few more SV4 files and hybrid SV6 files.

It looks a bit messy that the selection of land types varies for every hybrid save. The order can also be odd:

afbeelding

What I would propose as a solution is this: check if the file uses any RCT1 land texture. If it does, load all of them, and in the same order as we do currently. If none of used, only load the RCT2 textures.

duncanspumpkin:
> Ok, I just tested with a few more SV4 files and hybrid SV6 files.
> 
> It looks a bit messy that the selection of land types varies for every hybrid save. The order can also be odd:
> 
> afbeelding
> 
> What I would propose as a solution is this: check if the file uses any RCT1 land texture. If it does, load all of them, and in the same order as we do currently. If none of used, only load the RCT2 textures.

Can you try the latest version it now looks at all the tiles if there are any rct1 land or edges it will load all rct1 land and edges into the file. Saves having to recompute the index as well.

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1732,7 +1746,37 @@ namespace RCT2
                 }
             }
 
-            SetDefaultRCT2TerrainObjects(objectList);
+            // Add default rct2 terrain surfaces and edges
+            AddDefaultEntries();
+
+            // Find and if any rct1 terrain surfaces or edges have been used
```
pizza2004, line 1752:
```suggestion
            // Find if any rct1 terrain surfaces or edges have been used
```

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -1732,7 +1746,37 @@ namespace RCT2
...
+                        return true;
+                    }
+                    return false;
+                });
+
+            // If an rct1 surface or edge then load all the Hybrid surfaces and edges
+            if (hasRCT1Terrain)
+            {
+                _terrainSurfaceEntries.AddRange(OpenRCT2HybridTerrainSurfaces);
+                _terrainSurfaceEntries.AddRange(OpenRCT2HybridTerrainEdges);
```
Gymnasiast, line 1775:
```suggestion
                _terrainEdgeEntries.AddRange(OpenRCT2HybridTerrainEdges);
```

dannystrayer:
After this fix, how are we to obtain the RCT1 sides? The castle and urban sides as well as others don't show up in object selection now. I can't figure out how to add those textures to my park

Gymnasiast:
> After this fix, how are we to obtain the RCT1 sides? The castle and urban sides as well as others don't show up in object selection now. I can't figure out how to add those textures to my park

If they don’t show up in your Object Selection window, then you should probably check if your filters are set correctly. There is a button for those in the top right of the Object Selection window.

dannystrayer:
> > After this fix, how are we to obtain the RCT1 sides? The castle and urban sides as well as others don't show up in object selection now. I can't figure out how to add those textures to my park
> 
> If they don’t show up in your Object Selection window, then you should probably check if your filters are set correctly. There is a button for those in the top right of the Object Selection window.

Silly mistake! You are correct. Apparently I had LL and CF checked off. Thanks!

PR #16080, opened 2021-11-30
Fix #16007: Scenario Editor "Entry Price" appears to the right of the value field

### LINKED ISSUE #16007: Scenario Editor "Entry Price" appears to the right of the value field
### Operating System

Windows 10, 64-bit

### OpenRCT2 build

OpenRCT2, v0.3.5.1 (56e78d5 on develop) provided by GitHub

### Describe the issue

I was playing with the latest dev build, setting up a scenario on a 256x256 map with every object added.  I got to the Park Options section, and noticed "Entry Price" is on the right side of the value field, outside of the "Options - Park" window.  A second test  on the same version with a more regular map confirms the issue for me.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Launch the Scenario Editor
2. Progress to the "Options Selection" screen
3. Select the "Show Park Options" tab

### Attachments

image

Ian-Polito:
Entry Price label was being drawn to the right of the entryPriceWidget which is outside of the window. Changed to draw it to the right of the payForParkOrRidesWidget widget instead.

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -8,10 +8,12 @@
 - Feature: [#7660] Custom music objects that are distributed with the save.
 - Feature: [#8407] Ride platforms can be made invisible.
 - Feature: [#13858] Flatride bases can be made invisible.
+- Feature: [#16029] [Plugin] Add TrackElement.rideType to API.
 - Improved: [#3517] Cheats are now saved with the park.
 - Improved: [#10150] Ride stations are now properly checked if they’re sheltered.
-- Improved: [#10664] Invisibility status is now shown in the Tile Inspector.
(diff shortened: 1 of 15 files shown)


duncanspumpkin:

distribution/changelog.txt
```diff
@@ -16,6 +16,7 @@
 - Fix: [#15998] Cannot set map size to the actual maximum.
 - Fix: [#16008] Tile Inspector can select elements from last tile without reselecting it.
 - Fix: [#16063] Object Selection preview for objects with glass is broken.
+- Fix: [#16007] Scenario Editor "Entry Price" appears to the right of the value field
```
duncanspumpkin, line 19: sorry but could you move it so that it is in order of issue number.

PR #16081, opened 2021-11-30
Fix #15998: Map action only accepts 998×998 instead of 999×999

### LINKED ISSUE #15998: Park size maximum is 999 but maximum input accepted is 998
### OpenRCT2 build

OpenRCT2, v0.3.5.1 (2b4eaab on develop)

### Describe the issue

I noticed a lot of people saying that the new maximum park size is 999x999, but in game you can only make the spinner in the map view window go up to 998.

janisozaur pointed me to the source code to correct me saying that the size was 998 instead of 999, but it looks like it is intended to be 999:
https://github.com/OpenRCT2/OpenRCT2/blob/develop/src/openrct2/world/Map.h#L25-L27

### Steps to reproduce

1. Open a park in sandbox mode
2. Increase the park size to 999
3. The maximum value of the spinner is 998 instead of 999

Gymnasiast:


tupaschoal (approved):
Does this warrant a changelog entry?

Gymnasiast:
Hm, maybe, would have to check if v0.3.5.1 allows for 254×254 or only 253×253.

tupaschoal (commented):
LGTM then

PR #16085, opened 2021-12-01
Fix #16083: Tile Inspector shortcuts mapped incorrectly

### LINKED ISSUE #16083: Tile inspector hotkeys mapped incorrectly
### Operating System

Windows 10, 64-bit

### OpenRCT2 build

OpenRCT2, v0.3.5.1 (fb9d02c on develop) provided by GitHub

### Describe the issue

Basically #15928 happened again. Probably due to reorganizing the Tile Inspector widgets when the "invisible" button was removed.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Assign hotkeys to the tile inspector.
2. Open the TI and attempt to use the hotkeys.
3. Hotkeys don't do anything or perform the wrong action.

### Attachments

_No response_

Broxzier:
This really needs to be refactored so that this does not need manual updating every time the order of widgets changes.


IntelOrca:
I think other windows have static asserts to ensure it at least matches during build.

fidwell:
I tested this version (from [here](https://github.com/Broxzier/OpenRCT2/actions/runs/1524927879)).

* Raising and lowering an object with hotkeys now work as expected.
* The hotkey I picked for toggling invisibility deletes the object.
* Copying, pasting, and deleting do nothing
* Rotating (default Z) instead deselects the selected object (although it remains "ghosted" in the game world)

Broxzier:
@fidwell thanks for testing. This PR was still in draft because it was not finished - all top-toolbar buttons were broken, and have now been fixed.

> Rotating (default Z) instead deselects the selected object (although it remains "ghosted" in the game world)

I noticed that this is only partially implemented. The code for it is there, but there are no language strings for it and the hotkey is not registered. This could easily be added, but in another PR. Let's get this fix in first.

fidwell:
Awesome! Thanks again for fixing this!

PR #16087, opened 2021-12-01
Use g1 sprite for Looping RC booster

Gymnasiast:


duncanspumpkin:
This will change the graphics of existing parks. It should be performed in a different way.

duncanspumpkin (approved):

PR #16088, opened 2021-12-01
OpenRCT2 Limits

duncanspumpkin:
Adding the OpenRCT2 Limits file. WiP

Items for future:
```
    // constexpr const uint8_t PatrolAreaSize = 128;
    // constexpr const uint8_t StaffTypeCount = 4;

    // constexpr const uint8_t ExpenditureTableMonthCount = 16;
    // constexpr const uint8_t ExpenditureTypeCount = 14;
    // constexpr const uint8_t FinanceGraphSize = 128;

    // constexpr const uint16_t MaxUserStrings = 1024;
    // constexpr const uint8_t MaxUserStringLength = 32;

    // constexpr const uint8_t MaxPeepThoughts = 5;

    // constexpr const uint8_t MaxElementHeight = 255;

    // constexpr const uint8_t MaxBanners = 250;

    // constexpr const uint16_t MaxEntities = 10000;
    // constexpr const uint32_t MaxTileElements = 0x30000;
    // constexpr const uint16_t MaxAnimatedObjects = 2000;
    // constexpr const uint8_t MaxResearchedRideTypeQuads = 8;  // With 32 bits per uint32_t, this means there is room for
    //                                                         // 256 types.
    // constexpr const uint8_t MaxResearchedRideEntryQuads = 8; // With 32 bits per uint32_t, this means there is room for
    //                                                         // 256 entries.
    // constexpr const uint8_t MaxResearchedSceneryItemQuads = 56;
    // constexpr const uint16_t MaxResearchedSceneryItems = (MaxResearchedSceneryItemQuads * 32); // There are 32
    //                                                                                           // bits per
    //                                                                                           // quad.
    // constexpr const uint16_t MaxResearchItems = 500;

    // constexpr uint16_t TD6MaxTrackElements = 8192;
    // constexpr const uint16_t MaxMapSize = 256;
    // constexpr const uint8_t RideTypeCount = 91;
```


src/openrct2/Limits.h
```diff
@@ -0,0 +1,80 @@
...
+
+namespace OpenRCT2::Limits
+{
+    constexpr const uint16_t MaxRidesInPark = 1000;
+    constexpr const uint16_t MaxStationsPerRide = 255;
+    // constexpr const uint16_t RideMeasurementMaxItems = 4800; // Is part of Measurement struct
+    using RCT12::Limits::CustomerHistorySize;
+    using RCT12::Limits::MaxGolfHoles;
+    using RCT12::Limits::MaxHelices;
+    using RCT12::Limits::MaxInversions;
```
ZehMatt, line 21: I'm not a fan doing that, that still ties our internal limits to the import of rct1/rct2, they should be separate constants.

Legotown55:
Pleas will you fix this pull

src/openrct2/Limits.h
duncanspumpkin: I want it to be clear if a constant has changed from RCT12 or not. If its changed then you would have constexpr const ... but otherwise its just a using.

ZehMatt: I know but that assumes we use the same internal limit which I want to be explicit about rather. The RCT12 stuff is primarily useful for the legacy importer/exporters. Changing that constant will will either break our structures or the importer so I don't know if its actually sensible.

Broxzier: This look fine to me. It shows that these values are the same, and OpenRCT2::Limits is not used for importing SV6 or structure sizes afaik. I'd take this over magic numbers that are the same as RCT2's. What alternative do you propose?

ZehMatt: I would just declare it explicitly like 
```cpp
constexpr const uint16_t CustomerHistorySize = RCT12::Limits::CustomerHistorySize;
...
```
If we ever get to change numbers we just have to modify the assignment, either way I'm don't want to be in the way over this, can be changed if needed at any point.

src/openrct2/Limits.h
```diff
@@ -0,0 +1,30 @@
+/*****************************************************************************
+ * Copyright (c) 2014-2021 OpenRCT2 developers
```
tupaschoal, line 2:
```suggestion
 * Copyright (c) 2014-2022 OpenRCT2 developers
```

tupaschoal (approved):

src/openrct2/Limits.h
```diff
@@ -0,0 +1,30 @@
...
+{
+    constexpr const uint16_t MaxRidesInPark = 1000;
+    constexpr const uint16_t MaxStationsPerRide = 255;
+    constexpr const uint8_t CustomerHistorySize = RCT12::Limits::CustomerHistorySize;
+    constexpr const uint16_t MaxGolfHoles = RCT12::Limits::MaxGolfHoles;
+    constexpr const uint16_t MaxHelices = RCT12::Limits::MaxHelices;
+    constexpr const uint16_t MaxInversions = RCT12::Limits::MaxInversions;
+    constexpr const uint16_t MaxTrainsPerRide = 255;
+    constexpr const uint16_t MaxCarsPerTrain = 255;
+    constexpr const uint16_t MaxVehicleColours = MaxTrainsPerRide; // this should really be MaxTrainsPerRide * MaxCarsPerTrain
```
Broxzier, line 23: This one seems to be unused. Is that because `MaxTrainsPerRide` in used where it should be this `MaxVehicleColours` now?

duncanspumpkin: Yes, it would be split up again. I did have it split but that was reverted in the previous iteration of this pr

ZehMatt (approved):

PR #16092, opened 2021-12-02
Refactor update/time related things

ZehMatt:


oli414:
I think this change might have broken right click camera panning on Windows

src/openrct2/core/Timer.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+        Timepoint _tp = Clock::now();
+
+    public:
+        void Restart() noexcept
+        {
+            _tp = Clock::now();
+        }
+
+        float GetElapsed() const noexcept
```
Broxzier, line 28: It's not clear what time units `Timer` uses or what this returns.

```suggestion
        float GetElapsedSeconds() const noexcept
```

src/openrct2/core/Timer.hpp
```diff
@@ -0,0 +1,35 @@
...
+        {
+            auto elapsed = std::chrono::duration_cast<std::chrono::microseconds>(Clock::now() - _tp);
+            return static_cast<float>(static_cast<double>(elapsed.count()) / 1'000'000.0);
```
Broxzier, line 31: `std::chrono::duration` can take a float as its type, which is seconds by default (no ratio needed). This simplified the code a bit:

```suggestion
        auto elapsed = std::chrono::duration<float>(Clock::now() - _tp);
        return elapsed.count();
```

src/openrct2/network/DiscordService.cpp
```diff
@@ -24,7 +24,7 @@
 
 constexpr const char* APPLICATION_ID = "378612438200877056";
 constexpr const char* STEAM_APP_ID = nullptr;
-constexpr const uint32_t REFRESH_INTERVAL = 5 * GAME_UPDATE_FPS; // 5 seconds
+constexpr const float REFRESH_INTERVAL = 5.0f; // seconds
```
Broxzier, line 27: Have you tried changing this from float to `std::chrono::duration`? We could use the standard defined literals for duration values, then. [`5s` instead of `5.0f`](https://en.cppreference.com/w/cpp/chrono/operator%22%22s).

src/openrct2/Context.cpp
```diff
@@ -947,16 +948,23 @@ namespace OpenRCT2
             RunGameLoop();
         }
 
+        bool ShouldDraw()
+        {
+            if (gOpenRCT2Headless)
+                return false;
+            if (_uiContext->IsMinimised())
+                return false;
```
Margen67, line 956:
```suggestion
            if (gOpenRCT2Headless || _uiContext->IsMinimised())
                return false;
```

Broxzier: Running headless and minimized are not related, so I don't see the added value of combining the two on one line.

src/openrct2/Context.cpp
```diff
@@ -1099,12 +1106,11 @@ namespace OpenRCT2
...
         }
 
-        void Update()
+        void Tick()
         {
-            uint32_t currentUpdateTime = platform_get_ticks();
-
-            gCurrentDeltaTime = std::min<uint32_t>(currentUpdateTime - _lastUpdateTime, 500);
-            _lastUpdateTime = currentUpdateTime;
+            // TODO: This variable has been never "variable" in time, some code expects
```
Broxzier, line 1111:
```suggestion
            // TODO: This variable has never been "variable" in time, some code expects
```

Broxzier (commented):
Tested, and everything seems to work as expected.

src/openrct2/Context.cpp
```diff
@@ -987,11 +994,13 @@ namespace OpenRCT2
 
         void RunFrame()
         {
+            const auto deltaTime = _timer.GetElapsed();
+            _timer.Restart();
```
tupaschoal: Should there be a convenience function that does both? Gets the elapsed and then restarts?

src/openrct2/Context.cpp
```diff
@@ -1003,32 +1012,25 @@ namespace OpenRCT2
...
-            _lastTick = currentTick;
-            _accumulator = std::min(_accumulator + elapsed, static_cast<float>(GAME_UPDATE_MAX_THRESHOLD));
+            float scaledDeltaTime = deltaTime * _timeScale;
+            _accumulator = std::min(_accumulator + scaledDeltaTime, GAME_UPDATE_MAX_THRESHOLD);
 
             _uiContext->ProcessMessages();
 
             if (_accumulator < GAME_UPDATE_TIME_MS)
             {
-                platform_sleep(GAME_UPDATE_TIME_MS - _accumulator - 1);
```
tupaschoal, line 1031: Why is the `-1` gone?

ZehMatt: Avoiding an underflow, now with the more precise timing this was actually happening

src/openrct2/Context.cpp
ZehMatt: What would you call it? I was unable to come up with a sensible name that isn't too long and yet describes what it does.

ZehMatt: What would you call it? I was unable to come up with a sensible name that isn't too long and yet describes what it does.

tupaschoal: I haven't been able to come up with anything that looks great too, could only think of `getElapsedAndReset()`

Broxzier: Why not have `Reset` return the value before the reset? Many timer implementations do that.

ZehMatt: Not sure if its actually worth the trouble from turning two lines to one by adding a new function to be honest.

tupaschoal: I think it boils down to whether you'll want to read `GetElapsed` two times in a row without restarting the timer. If this seems to be a good use case, then we should be fine with two functions, but if every time we'll have to read and reset, then a single function seems less error prone.

tupaschoal (approved):

PR #16097, opened 2021-12-03
Make Looping Coaster draw missing sprites from LIM Launched Coaster

karst:
Makes the Looping Coaster draw missing track pieces from the LIM Launch Coaster.

I have been told replacing return nullptr; to return get_track_paint_function_lim_launched_rc would have also done the job, but this seems pretty clear to me and it works!


duncanspumpkin:
can you address the formatting issues.

Gymnasiast (changes requested):
> I have been told replacing return nullptr; to return get_track_paint_function_lim_launched_rc would have also done the job, but this seems pretty clear to me and it works!

I still think that’s the best way to go, because it means that, if we added another element to the LIM coaster, the Looping RC will also be able to draw it without having to update LoopingRollerCoaster.cpp. This is why I asked you to do it this way.

karst:
I for now added the lines like this. Formatting should be correct now (I used tabs instead of spacebars).
This code is consistent to what I found in other 
Can't seem to get the cpp files (like the LIM Launched Coaster itself)

Could not get the adviced line of code to work, as I got the following error. Otherwise I had done it like such.
image

For reference:
image

src/openrct2/ride/coaster/LoopingRollerCoaster.cpp
```diff
@@ -9463,6 +9463,29 @@ TRACK_PAINT_FUNCTION get_track_paint_function_looping_rc(int32_t trackType)
...
+        case TrackElemType::Down90ToDown60:
+        case TrackElemType::Up90ToUp60:
+        case TrackElemType::Down60ToDown90:
+        case TrackElemType::Up90ToInvertedFlatQuarterLoop:
+        case TrackElemType::InvertedFlatToDown90QuarterLoop:
+        case TrackElemType::LeftQuarterTurn1TileUp90:
+        case TrackElemType::RightQuarterTurn1TileUp90:
+        case TrackElemType::LeftQuarterTurn1TileDown90:
+        case TrackElemType::RightQuarterTurn1TileDown90:
+            return get_track_paint_function_lim_launched_rc(trackType);
```
ZehMatt, line 9488: Don't invoke the function, this is supposed to return the function address.

karst:
> > I have been told replacing return nullptr; to return get_track_paint_function_lim_launched_rc would have also done the job, but this seems pretty clear to me and it works!
> 
> I still think that’s the best way to go, because it means that, if we added another element to the LIM coaster, the Looping RC will also be able to draw it without having to update LoopingRollerCoaster.cpp. This is why I asked you to do it this way.

@Gymnasiast Got it to work in the singular line. :)

Gymnasiast:
Good, can you also amend the `ExtraTrackPieces` in `LoopingRollerCoaster.h`? That would, people who enable cheats can easily build the elements you added?

karst:
There we go!

src/openrct2/ride/coaster/meta/LoopingRollerCoaster.h
```diff
@@ -20,7 +20,7 @@ constexpr const RideTypeDescriptor LoopingRollerCoasterRTD =
     SET_FIELD(AlternateType, RIDE_TYPE_NULL),
     SET_FIELD(Category, RIDE_CATEGORY_ROLLERCOASTER),
     SET_FIELD(EnabledTrackPieces, (1ULL << TRACK_STRAIGHT) | (1ULL << TRACK_STATION_END) | (1ULL << TRACK_LIFT_HILL) | (1ULL << TRACK_FLAT_ROLL_BANKING) | (1ULL << TRACK_VERTICAL_LOOP) | (1ULL << TRACK_SLOPE) | (1ULL << TRACK_SLOPE_STEEP) | (1ULL << TRACK_SLOPE_CURVE) | (1ULL << TRACK_SLOPE_CURVE_STEEP) | (1ULL << TRACK_S_BEND) | (1ULL << TRACK_CURVE_SMALL) | (1ULL << TRACK_CURVE) | (1ULL << TRACK_HELIX_SMALL) | (1ULL << TRACK_BRAKES) | (1ULL << TRACK_ON_RIDE_PHOTO) | (1ULL << TRACK_BLOCK_BRAKES) | (1ULL << TRACK_SLOPE_ROLL_BANKING) | (1ULL << TRACK_BOOSTER)),
-    SET_FIELD(ExtraTrackPieces, 0),
+    SET_FIELD(ExtraTrackPieces, (1ULL << TRACK_SLOPE_VERTICAL) | (1ULL << TRACK_TWIST) | (1ULL << TRACK_CORKSCREW) | (1ULL << TRACK_CURVE_VERTICAL) | (1ULL << TRACK_QUARTER_LOOP) ),
```
Gymnasiast, line 23:
```suggestion
    SET_FIELD(ExtraTrackPieces, (1ULL << TRACK_SLOPE_VERTICAL) | (1ULL << TRACK_TWIST) | (1ULL << TRACK_CORKSCREW) | (1ULL << TRACK_CURVE_VERTICAL) | (1ULL << TRACK_QUARTER_LOOP)),
```

PR #16098, opened 2021-12-03
Refactor ImageId and use in more functions

IntelOrca:
* Extract `ImageId` into its own header.
* Introduce `ImageIndex` to refer to just the image number (no colour, palette or flags).
* Use ImageIndex and ImageId in OpenGL and `IDrawingContext` interface.
* Add a few more paint overloads, prep. for more ImageId changes.

### DIFF
--- a/src/openrct2-ui/windows/EditorObjectSelection.cpp
+++ b/src/openrct2-ui/windows/EditorObjectSelection.cpp
@@ -957,10 +957,8 @@ public:
         }
 
         const int32_t ride_tabs[] = {
-            SPR_TAB_RIDE_16,        IMAGE_TYPE_REMAP | SPR_TAB_RIDES_TRANSPORT_0,
-            SPR_TAB_RIDES_GENTLE_0, IMAGE_TYPE_REMAP | SPR_TAB_RIDES_ROLLER_COASTERS_0,
-            SPR_TAB_RIDES_THRILL_0, SPR_TAB_RIDES_WATER_0,
-            SPR_TAB_RIDES_SHOP_0,   SPR_TAB_FINANCES_RESEARCH_0,
+            SPR_TAB_RIDE_16,        SPR_TAB_RIDES_TRANSPORT_0, SPR_TAB_RIDES_GENTLE_0, SPR_TAB_RIDES_ROLLER_COASTERS_0,
+            SPR_TAB_RIDES_THRILL_0, SPR_TAB_RIDES_WATER_0,     SPR_TAB_RIDES_SHOP_0,   SPR_TAB_FINANCES_RESEARCH_0,
         };
         const int32_t ThrillRidesTabAnimationSequence[] = {
             5, 6, 5, 4, 3, 2, 1, 0, 0, 0, 0, 0, 0, 0, 1, 2, 3, 4, 0, 0, 0,
--- a/src/openrct2-ui/windows/NewRide.cpp
+++ b/src/openrct2-ui/windows/NewRide.cpp
@@ -629,9 +629,9 @@ static void WindowNewRideDrawTabImage(rct_drawpixelinfo* dpi, rct_window* w, int
 
 static void WindowNewRideDrawTabImages(rct_drawpixelinfo* dpi, rct_window* w)
 {
-    WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_TRANSPORT, IMAGE_TYPE_REMAP | SPR_TAB_RIDES_TRANSPORT_0);
+    WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_TRANSPORT, SPR_TAB_RIDES_TRANSPORT_0);
     WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_GENTLE, SPR_TAB_RIDES_GENTLE_0);
-    WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_ROLLER_COASTER, IMAGE_TYPE_REMAP | SPR_TAB_RIDES_ROLLER_COASTERS_0);
+    WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_ROLLER_COASTER, SPR_TAB_RIDES_ROLLER_COASTERS_0);
     WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_THRILL, SPR_TAB_RIDES_THRILL_0);
     WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_WATER, SPR_TAB_RIDES_WATER_0);
     WindowNewRideDrawTabImage(dpi, w, WINDOW_NEW_RIDE_PAGE_SHOP, SPR_TAB_RIDES_SHOP_0);
@@ -793,6 +793,13 @@ static void WindowNewRidePaint(rct_window* w, rct_drawpixelinfo* dpi)
     }
 }
 
+static ImageIndex GetRideImage(RideSelection rideSelection)
+{
+    auto& objMgr = OpenRCT2::GetContext()->GetObjectManager();
+    auto obj = static_cast<RideObject*>(objMgr.GetLoadedObject(ObjectType::Ride, rideSelection.EntryIndex));
+    return obj == nullptr ? ImageIndexUndefined : obj->GetPreviewImage(rideSelection.Type);
+}
+
 /**
  *
  *  rct2: 0x006B6ABF
@@ -808,7 +815,6 @@ static void WindowNewRideScrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int3
     RideSelection* listItem = _windowNewRideListItems;
     while (listItem->Type != RIDE_TYPE_NULL || listItem->EntryIndex != OBJECT_ENTRY_INDEX_NULL)
     {
-        rct_ride_entry* rideEntry;
         // Draw flat button rectangle
         int32_t flags = 0;
         if (w->new_ride.SelectedRide == *listItem)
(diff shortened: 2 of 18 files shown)


src/openrct2/drawing/NewDrawing.cpp
```diff
@@ -233,18 +233,18 @@ void gfx_draw_dashed_line(
     }
 }
 
-void FASTCALL gfx_draw_sprite(rct_drawpixelinfo* dpi, ImageId image_id, const ScreenCoordsXY& coords)
+void FASTCALL gfx_draw_sprite(rct_drawpixelinfo* dpi, int32_t image, const ScreenCoordsXY& coords, uint32_t tertiary_colour)
```
Gymnasiast, line 236: I guess we still need this overload?

src/openrct2-ui/windows/EditorObjectSelection.cpp
```diff
@@ -957,10 +957,8 @@ class EditorObjectSelectionWindow final : public Window
         }
 
         const int32_t ride_tabs[] = {
-            SPR_TAB_RIDE_16,        IMAGE_TYPE_REMAP | SPR_TAB_RIDES_TRANSPORT_0,
-            SPR_TAB_RIDES_GENTLE_0, IMAGE_TYPE_REMAP | SPR_TAB_RIDES_ROLLER_COASTERS_0,
-            SPR_TAB_RIDES_THRILL_0, SPR_TAB_RIDES_WATER_0,
-            SPR_TAB_RIDES_SHOP_0,   SPR_TAB_FINANCES_RESEARCH_0,
+            SPR_TAB_RIDE_16,        SPR_TAB_RIDES_TRANSPORT_0, SPR_TAB_RIDES_GENTLE_0, SPR_TAB_RIDES_ROLLER_COASTERS_0,
+            SPR_TAB_RIDES_THRILL_0, SPR_TAB_RIDES_WATER_0,     SPR_TAB_RIDES_SHOP_0,   SPR_TAB_FINANCES_RESEARCH_0,
```
Gymnasiast, line 961: Test note: do these tab images still work correctly?

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -818,18 +833,9 @@ static void WindowNewRideScrollpaint(rct_window* w, rct_drawpixelinfo* dpi, int3
...
-        {
-            if (rideEntry->ride_type[i] == listItem->Type)
-                break;
-
-            imageId++;
-        }
-
-        gfx_draw_sprite_raw_masked(dpi, coords + ScreenCoordsXY{ 2, 2 }, SPR_NEW_RIDE_MASK, imageId);
+        auto mask = ImageId(SPR_NEW_RIDE_MASK);
+        auto colour = GetRideImage(*listItem);
```
Gymnasiast, line 837: `colour` seems a bit of an odd name for this.

```suggestion
        auto rideImage = GetRideImage(*listItem);
```

src/openrct2/paint/Paint.cpp
```diff
@@ -549,19 +550,20 @@ static void PaintAttachedPS(rct_drawpixelinfo* dpi, paint_struct* ps, uint32_t v
     {
         auto screenCoords = ScreenCoordsXY{ attached_ps->x + ps->x, attached_ps->y + ps->y };
 
-        uint32_t imageId = PaintPSColourifyImage(attached_ps->image_id, ps->sprite_type, viewFlags);
+        auto imageId = PaintPSColourifyImage(
+            ImageId::FromUInt32(attached_ps->image_id, attached_ps->tertiary_colour), ps->sprite_type, viewFlags);
```
Gymnasiast, line 554: It previously used `ps->tertiary_colour` to set the tertiary colour. So shouldn’t this line read like this:

```suggestion
            ImageId::FromUInt32(attached_ps->image_id, ps->tertiary_colour), ps->sprite_type, viewFlags);
```

src/openrct2/paint/Supports.cpp
```diff
@@ -595,8 +595,8 @@ bool wooden_a_supports_paint_setup(
             {
                 hasSupports = true;
                 paint_struct* ps = PaintAddImageAsOrphan(
-                    session, imageId, 0, 0, bBox.length.x, bBox.length.y, bBox.length.z, z, bBox.offset.x, bBox.offset.y,
-                    bBox.offset.z + z);
+                    session, ImageId::FromUInt32(imageId), { 0, 0, z }, { bBox.length.x, bBox.length.y, bBox.length.z },
```
Gymnasiast, line 598: bBox.length is already a CoordsXYZ(D), right? So we could do this:

```suggestion
                    session, ImageId::FromUInt32(imageId), { 0, 0, z }, bBox.length,
```

src/openrct2/paint/Supports.cpp
```diff
@@ -770,8 +770,9 @@ bool wooden_b_supports_paint_setup(
             else
             {
                 paint_struct* paintStruct = PaintAddImageAsOrphan(
-                    session, imageId | imageColourFlags, 0, 0, boundBox.length.x, boundBox.length.y, boundBox.length.z,
-                    baseHeight, boundBox.offset.x, boundBox.offset.y, boundBox.offset.z + baseHeight);
+                    session, ImageId::FromUInt32(imageId | imageColourFlags), { 0, 0, baseHeight },
+                    { boundBox.length.x, boundBox.length.y, boundBox.length.z },
```
Gymnasiast, line 774:
```suggestion
                    boundBox.length,
```

src/openrct2/paint/Supports.cpp
```diff
@@ -1292,8 +1293,9 @@ bool path_a_supports_paint_setup(
         else
         {
             paint_struct* paintStruct = PaintAddImageAsOrphan(
-                session, imageId | imageColourFlags, 0, 0, boundBox.length.y, boundBox.length.x, boundBox.length.z, baseHeight,
-                boundBox.offset.x, boundBox.offset.y, baseHeight + boundBox.offset.z);
+                session, ImageId::FromUInt32(imageId | imageColourFlags), { 0, 0, baseHeight },
+                { boundBox.length.y, boundBox.length.x, boundBox.length.z },
```
Gymnasiast, line 1297:
```suggestion
                boundBox,
```

src/openrct2/paint/Paint.cpp
```diff
@@ -647,36 +649,31 @@ static void PaintPSImageWithBoundingBoxes(rct_drawpixelinfo* dpi, paint_struct*
...
-static void PaintPSImage(rct_drawpixelinfo* dpi, paint_struct* ps, uint32_t imageId, int32_t x, int32_t y)
+static void PaintPSImage(rct_drawpixelinfo* dpi, paint_struct* ps, ImageId imageId, int32_t x, int32_t y)
 {
     if (ps->flags & PAINT_STRUCT_FLAG_IS_MASKED)
     {
-        return gfx_draw_sprite_raw_masked(dpi, { x, y }, imageId, ps->colour_image_id);
+        return gfx_draw_sprite_raw_masked(dpi, { x, y }, imageId, ImageId::FromUInt32(ps->colour_image_id));
     }
 
-    gfx_draw_sprite(dpi, imageId, { x, y }, ps->tertiary_colour);
```
Gymnasiast, line 657: It seems as if calls to PaintPSImage() have not been modified to put `ps->tertiary_colour` into the ImageId.

src/openrct2-ui/windows/EditorObjectSelection.cpp
IntelOrca: My changes to ImageId stopped them from working. Because it was using the constructor for ImageId which only expects the index, not an original uint32 of the image. If you use the constructor where you pass the colour, that implies remap. So this fixes them.

src/openrct2/paint/Paint.cpp
IntelOrca: There are only 3 calls to `PaintPSImage`, one is for painting bound boxes which doesn't need to. The other two are directly after calls to PaintPSColourifyImage.

src/openrct2/drawing/NewDrawing.cpp
IntelOrca: There are hundreds of calls to this, will take a long time to go through them all. Most important ones are just any that draw an image from an object file. UI drawing isn't so important as those IDs will never be above the old image limit.

IntelOrca: Actually it seems many have already been changed, there are still a few left though. Can be removed once they are all changed.

src/openrct2-ui/windows/NewRide.cpp
```diff
@@ -793,6 +793,22 @@ static void WindowNewRidePaint(rct_window* w, rct_drawpixelinfo* dpi)
...
+        for (uint32_t i = 0; i < RCT2::ObjectLimits::MaxRideTypesPerRideEntry; i++)
+        {
+            if (rideEntry->ride_type[i] == rideSelection.Type)
+            {
+                return ImageId(rideEntry->images_offset + i);
+            }
+        }
+    }
+    return ImageId();
+}
```
ZehMatt, line 810:
```suggestion
static ImageId GetRideImage(RideSelection rideSelection)
{
    const auto* rideEntry = get_ride_entry(rideSelection.EntryIndex);
    if (rideEntry == nullptr)
        return ImageId{};

    const auto it = std::find(std::begin(rideEntry->ride_type), std::end(rideEntry->ride_type), rideSelection.Type);
    if (it == std::end(rideEntry->ride_type))
        return ImageId{};

    const auto i = std::distance(it, std::end(rideEntry->ride_type));
    return ImageId(rideEntry->images_offset + i);
}
```
In my personal opinion having early exits everywhere makes code a lot easier to read as it reduces scopes which reduces overall line length. Also always use standard algorithms if possible. The code may not compile as I wrote this here on GitHub.

src/openrct2/drawing/Drawing.Sprite.cpp
```diff
@@ -654,12 +654,12 @@ const rct_g1_element* gfx_get_g1_element(ImageId imageId)
...
 }
 
-const rct_g1_element* gfx_get_g1_element(int32_t image_id)
+const rct_g1_element* gfx_get_g1_element(ImageIndex image_id)
 {
     openrct2_assert(!gOpenRCT2NoGraphics, "gfx_get_g1_element called on headless instance");
 
     auto offset = static_cast<size_t>(image_id);
-    if (offset == 0x7FFFF)
+    if (offset == 0x7FFFF || offset == ImageIndexUndefined)
```
ZehMatt, line 662: This seems a bit wrong, shouldn't it use the function to get the index?

ZehMatt: Nevermind, I confused ImageIndex with ImageId

ZehMatt:
Code looks fine but I would like to give it a test run first.

ZehMatt (approved):

Gymnasiast (approved):

PR #16101, opened 2021-12-03
Refactor some code into UI for Staff

ZehMatt:


src/openrct2-ui/windows/StaffList.cpp
```diff
@@ -498,6 +499,49 @@ class StaffListWindow final : public Window
     }
 
 private:
+    /**
+     * Hires a new staff member of the given type.
+     */
+    bool HireNewMember(StaffType staffType, EntertainerCostume entertainerType)
```
Broxzier, line 505: The return value seems to be ignored, so it can be removed.

duncanspumpkin (approved):

PR #16102, opened 2021-12-03
Use templated function for the Twister and Vertical Drop

Gymnasiast:
Would like to know how this is expected to work out compilation-wise. I would hope it would be marginally quicker, since previously all Twister and Vertical RC functions only called their BM counterpart, while it now calls all these functions directly. But this is something out of my expertise.


ZehMatt (approved):
Performance wise this is probably identical but code wise its a lot better as it happens to nuke 3000 lines. If we want to actually gain performance we could look into making the function in the field rather an array of functions so we have a function per rotation.

Gymnasiast:
It nukes almost _4_000 lines ;-). Anyway, nice to hear this is an improvement overall. :+1:

PR #16104, opened 2021-12-03
Icons for new object selection tabs

oli414:
This PR adds 5 new icons for the object selection window

image

- Terrain edge
- Path surface
- Path support/railings
- Legacy paths
- Ride entrance


Gymnasiast:
Please rename path_supports.png to path_railings.png - we call it "path railings" everywhere in the code and it’s better to be consistent.

Gymnasiast:
The icon for path railings looks shifted to the left. It looks like it needs adjustment by one or two pixels.

oli414:
> The icon for path railings looks shifted to the left. It looks like it needs adjustment by one or two pixels.

It is intentionally not centered so that it is more clear that what you see underneath are supports.

Gymnasiast:
It looks a bit off to me, but well...

Gymnasiast (approved):

PR #16091, opened 2021-12-02
Fix #16075: Track designs save scenery in wrong location

ZehMatt:
Should have used the state from the ride instead of creating a new one.

PR #16116, opened 2021-12-04
Make track designs NSF map size safe

duncanspumpkin:
This really needs #16091 before it can be tested properly.
After this should do the same for entrances. Entrances are NSF map size safe I think but they could do with using the standard CoordsXYZ type instead of yet another xyz type.

### DIFF
--- a/src/openrct2/ride/TrackDesignSave.cpp
+++ b/src/openrct2/ride/TrackDesignSave.cpp
@@ -217,12 +217,9 @@ static bool track_design_is_supported_object(const Object* obj)
 static void track_design_save_push_tile_element_desc(
     const rct_object_entry& entry, const CoordsXYZ& loc, uint8_t flags, uint8_t primaryColour, uint8_t secondaryColour)
 {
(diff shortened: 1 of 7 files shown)


src/openrct2/ride/TrackDesignSave.cpp
```diff
@@ -446,11 +443,7 @@ static void track_design_save_pop_tile_element_desc(const ObjectEntryDescriptor&
     for (size_t i = 0; i < _trackSavedTileElementsDesc.size(); i++)
     {
         TrackDesignSceneryElement* item = &_trackSavedTileElementsDesc[i];
-        if (item->x != tileLoc.x)
```
Gymnasiast, line 449: You also need to remove the `tileLoc` variable entirely (it causes the CI to fail).

duncanspumpkin:
Tested this a bit now that #16091 is in and it seems to work fine. I created a ride at 260,260 and had scenery at 200,200, 260,200, 200,260 and 270,270 and they all save correctly now.

PR #16118, opened 2021-12-04
Refactor Custom Currency window to class

### LINKED ISSUE #13767: Refactor window to class: CustomCurrency
See #13737 for more details.

GalBr:
Closes #13767

I opened and played around with the Custom Currency window before and after my change and made sure it looks and functions the same.

### DIFF
--- a/src/openrct2-ui/windows/CustomCurrency.cpp
+++ b/src/openrct2-ui/windows/CustomCurrency.cpp
@@ -42,204 +42,186 @@ static rct_widget window_custom_currency_widgets[] = {
     WIDGETS_END,
 };
 
-
-static void CustomCurrencyWindowMousedown(rct_window *w, rct_widgetindex widgetIndex, rct_widget *widget);
-static void CustomCurrencyWindowMouseup(rct_window *w, rct_widgetindex widgetIndex);
-static void CustomCurrencyWindowDropdown(rct_window *w, rct_widgetindex widgetIndex, int32_t dropdownIndex);
-static void CustomCurrencyWindowTextInput(struct rct_window *w, rct_widgetindex widgetIndex, char *text);
-static void CustomCurrencyWindowPaint(rct_window *w, rct_drawpixelinfo *dpi);
-
-
-static rct_window_event_list _windowCustomCurrencyEvents([](auto& events)
-{
-    events.mouse_up = &CustomCurrencyWindowMouseup;
-    events.mouse_down = &CustomCurrencyWindowMousedown;
-    events.dropdown = &CustomCurrencyWindowDropdown;
-    events.text_input = &CustomCurrencyWindowTextInput;
-    events.paint = &CustomCurrencyWindowPaint;
-});
 // clang-format on
+
+class CustomCurrencyWindow final : public Window
+{
+public:
+    void OnOpen() override
+    {
+        widgets = window_custom_currency_widgets;
+        enabled_widgets = (1ULL << WIDX_CLOSE) | (1ULL << WIDX_RATE) | (1ULL << WIDX_RATE_UP) | (1ULL << WIDX_RATE_DOWN)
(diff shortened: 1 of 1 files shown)


src/openrct2-ui/windows/CustomCurrency.cpp
```diff
@@ -42,204 +42,182 @@ static rct_widget window_custom_currency_widgets[] = {
...
-                CurrencyDescriptors[EnumValue(CurrencyType::Custom)].rate = rate;
-                gConfigGeneral.custom_currency_rate = CurrencyDescriptors[EnumValue(CurrencyType::Custom)].rate;
                 config_save_default();
                 window_invalidate_all();
-            }
-            break;
+                break;
+
+            case WIDX_RATE:
+                rate = strtol(std::string(text).c_str(), &end, 10);
```
duncanspumpkin, line 172: Can  you use `String::Parse` in our core/string.hpp file instead of this.

src/openrct2-ui/windows/CustomCurrency.cpp
```diff
@@ -42,204 +42,182 @@ static rct_widget window_custom_currency_widgets[] = {
...
-    auto drawPos = w->windowPos
-        + ScreenCoordsXY{ window_custom_currency_widgets[WIDX_AFFIX_DROPDOWN].left + 1,
-                          window_custom_currency_widgets[WIDX_AFFIX_DROPDOWN].top };
-    rct_string_id stringId = (CurrencyDescriptors[EnumValue(CurrencyType::Custom)].affix_unicode == CurrencyAffix::Prefix)
-        ? STR_PREFIX
-        : STR_SUFFIX;
-    DrawTextBasic(dpi, drawPos, stringId, {}, { w->colours[1] });
+rct_window* CustomCurrencyWindowOpen()
+{
+    return WindowFocusOrCreate<CustomCurrencyWindow>(WC_CUSTOM_CURRENCY_CONFIG, 400, 100, 0);
```
duncanspumpkin, line 222: use WW and WH macros for the width and height.
Also this is not creating the window centered. You need to add the flag `WF_CENTRE_SCREEN`

duncanspumpkin (commented):
Looks good couple little points to address. Also can you add a space between each function it looks a bit neater.

src/openrct2-ui/windows/CustomCurrency.cpp
GalBr: That was missing in the original code and I wondered about it, thanks!

src/openrct2-ui/windows/CustomCurrency.cpp
GalBr: I'm having trouble using it, I would love some help.
It expects T to be a number type, but strtol expects char*, so I can't use it to cast text to char*. Am I missing something?

duncanspumpkin: you would use it like this: 
```cpp
auto res = String::Parse<int32_t>(text);
if (res.has_value()){
    rate = res.value();
    ...
}```

GalBr: Done.

GalBr:
Just an aside, I made sure it is also on develop (459c0d67dd5357b872d40850615390cf23f2e2fd) and not related to this branch, but suddenly everywhere in the game's text, the letter l (lowercase L) got replaced with different letter.
I built the develop branch to make sure it happens there as well and it does, but just in the off chance that I broke something, I'm letting you know about it.
Here's how it looks for me:
image

duncanspumpkin:
That happens when your g2.dat is out of date. Just Rerun the g2 generator.

GalBr:
Good to know, thanks!
So in that case I think this PR is ready.

duncanspumpkin (approved):

PR #16127, opened 2021-12-05
Create fallback system for csg1 images

Gymnasiast:
- [x] Fix terrain surfaces
- [x] Include object release 1.2.3 (or 1.2.4, if it turns out I made a mistake in the objects)

### DIFF
--- a/src/openrct2-ui/interface/LandTool.cpp
+++ b/src/openrct2-ui/interface/LandTool.cpp
@@ -64,8 +64,8 @@ void LandTool::ShowSurfaceStyleDropdown(rct_window* w, rct_widget* widget, Objec
     for (size_t i = 0; i < MAX_TERRAIN_SURFACE_OBJECTS; i++)
     {
         const auto surfaceObj = static_cast<TerrainSurfaceObject*>(objManager.GetLoadedObject(ObjectType::TerrainSurface, i));
-        // NumImagesLoaded can be 1 for RCT1 surfaces if the user does not have RCT1 linked.
-        if (surfaceObj != nullptr && surfaceObj->NumImagesLoaded > 1)
+        // If fallback images are loaded, the RCT1 styles will just look like copies of already existing styles, so hide them.
+        if (surfaceObj != nullptr && !surfaceObj->UsesFallbackImages())
         {
             gDropdownItemsFormat[itemIndex] = Dropdown::FormatLandPicker;
             gDropdownItemsArgs[itemIndex] = surfaceObj->IconImageId;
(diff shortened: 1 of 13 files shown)


src/openrct2-ui/interface/LandTool.cpp
```diff
@@ -64,8 +64,8 @@ void LandTool::ShowSurfaceStyleDropdown(rct_window* w, rct_widget* widget, Objec
     for (size_t i = 0; i < MAX_TERRAIN_SURFACE_OBJECTS; i++)
     {
         const auto surfaceObj = static_cast<TerrainSurfaceObject*>(objManager.GetLoadedObject(ObjectType::TerrainSurface, i));
-        // NumImagesLoaded can be 1 for RCT1 surfaces if the user does not have RCT1 linked.
-        if (surfaceObj != nullptr && surfaceObj->NumImagesLoaded > 1)
+        // If fallback images are loaded, the RCT1 styles will just look like copies of already existing styles, so hide them.
```
duncanspumpkin, line 67: We don't need to do this. The RCT1 files will still have the preview image so we can just load that anyway. Yeah when actually placing the object it will look the same but not at any other time.

src/openrct2/object/ImageTable.cpp
```diff
@@ -443,15 +443,23 @@ std::vector<std::pair<std::string, Image>> ImageTable::GetImageSources(IReadObje
     return result;
 }
 
-void ImageTable::ReadJson(IReadObjectContext* context, json_t& root)
+bool ImageTable::ReadJson(IReadObjectContext* context, json_t& root)
```
duncanspumpkin, line 446: Instead of returning the state this way you could just set a member variable of the image table to say its using fallbackSprites.

src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -416,7 +407,7 @@ static uint32_t get_tunnel_image(ObjectEntryIndex index, uint8_t type)
     if (obj != nullptr)
     {
         auto tobj = static_cast<TerrainEdgeObject*>(obj);
-        hasDoors = tobj->HasDoors;
+        hasDoors = tobj->HasDoors && !tobj->UsesFallbackImages();
```
duncanspumpkin, line 410: why this change?

duncanspumpkin:
Surface draw issue is caused by mistake in objects repo the fallback objects selected had different smoothing properties these smoothing properties require additional images which ends up knocking the whole image table off by 4. Just delete the `"$G1[28989..28994]",` lines and equivalent and it will work: 
image

src/openrct2/paint/tile_element/Paint.Surface.cpp
Gymnasiast: Because the fallback sprites lack doors. If I don’t include this, it will glitch heavily.

src/openrct2-ui/interface/LandTool.cpp
duncanspumpkin: I think it's fine to have the different styles shown as long as every time you open the park it has a warning that some images have been replaced due to rct1 not linked. This just gets really annoying if you want to draw things for rct1

Gymnasiast: Surely if you want to draw things for RCT1, you would link your RCT1 install so you can actually see what it will look like?

Nevertheless, if you feel strongly about this, we can discuss it in the team chat and see what the others think.

src/openrct2/paint/tile_element/Paint.Surface.cpp
duncanspumpkin: Should this become a flag of the object then?

Gymnasiast: It is - `UsesFallbackImages()` is simply a getter for a flag. Or is that not what you meant?

Gymnasiast:
(Merged after asking in team chat if it was okay. Line notes can be addressed in follow-up PR if you’d like me to.)

PR #16134, opened 2021-12-05
Make awards a vector

duncanspumpkin:
Just wanted to get rid of all the time == 0 comparisons. Also makes it easier to understand.

### DIFF
--- a/src/openrct2/management/Award.cpp
+++ b/src/openrct2/management/Award.cpp
@@ -65,7 +65,12 @@ static constexpr const rct_string_id AwardNewsStrings[] = {
     STR_NEWS_ITEM_BEST_GENTLE_RIDES,
 };
 
-Award gCurrentAwards[MAX_AWARDS];
+std::vector<Award> gCurrentAwards;
+
+std::vector<Award>& GetAwards()
+{
+    return gCurrentAwards;
+}
 
 bool award_is_positive(int32_t type)
 {
@@ -588,11 +593,7 @@ static bool award_is_deserved(int32_t awardType, int32_t activeAwardTypes)
 
 void award_reset()
 {
-    for (auto& award : gCurrentAwards)
-    {
-        award.Time = 0;
-        award.Type = 0;
-    }
+    gCurrentAwards.clear();
 }
 
 /**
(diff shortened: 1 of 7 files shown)


src/openrct2/management/Award.cpp
```diff
@@ -65,7 +65,12 @@ static constexpr const rct_string_id AwardNewsStrings[] = {
     STR_NEWS_ITEM_BEST_GENTLE_RIDES,
 };
 
-Award gCurrentAwards[MAX_AWARDS];
+std::vector<Award> gCurrentAwards;
```
ZehMatt, line 68:
```suggestion
static std::vector<Award> _currentAwards;
```
Doesn't seem to be a global anymore.

src/openrct2/management/Award.cpp
```diff
@@ -606,20 +607,16 @@ void award_update_all()
...
+            activeAwardTypes |= (1 << award.Type);
         }
 
         // Check if there was a free award entry
-        if (freeAwardEntryIndex != -1)
+        if (_currentAwards.size() < MAX_AWARDS)
         {
             // Get a random award type not already active
-            int32_t awardType;
+            uint16_t awardType;
```
tupaschoal, line 619: Why the type change?

src/openrct2/rct1/S4Importer.cpp
```diff
@@ -2130,12 +2130,13 @@ namespace RCT1
...
+            auto& awards = GetAwards();
+            for (auto& src : _s4.awards)
             {
-                rct12_award* src = &_s4.awards[i];
-                Award* dst = &gCurrentAwards[i];
-                dst->Time = src->time;
-                dst->Type = src->type;
+                if (src.time != 0)
+                {
+                    awards.push_back(Award{ src.type, src.time });
```
tupaschoal, line 2138:
```suggestion
                    awards.emplace_back(src.type, src.time);
```

src/openrct2/rct2/S6Importer.cpp
```diff
@@ -355,12 +355,13 @@ namespace RCT2
...
+            auto& awards = GetAwards();
+            for (auto& src : _s6.awards)
             {
-                rct12_award* src = &_s6.awards[i];
-                Award* dst = &gCurrentAwards[i];
-                dst->Time = src->time;
-                dst->Type = src->type;
+                if (src.time != 0)
+                {
+                    awards.push_back(Award{ src.type, src.time });
```
tupaschoal, line 363:
```suggestion
                    awards.emplace_back(src.type, src.time);
```

src/openrct2/management/Award.cpp
duncanspumpkin: So that I could use it in a braced initialiser. I'm going to make good first issue to replace all of these uint16_t awardTypes with the enum so thought it would also be good if they were the same type.

src/openrct2/rct1/S4Importer.cpp
duncanspumpkin: urgh. C++20 feature unfortunately. https://stackoverflow.com/questions/43394841/emplace-an-aggregate-in-stdvector

src/openrct2/management/Award.cpp
tupaschoal: Ok

src/openrct2/rct1/S4Importer.cpp
tupaschoal: Argh, that's a bummer.

tupaschoal (approved):

PR #16137, opened 2021-12-05
Refactor ride painting to use ImageId

IntelOrca:


Gymnasiast (approved):
Testing revealed no bugs (using bpb.sv6 to check all ride types touched). Haven’t been able to do a proper look through the code since it contains to many changes, but I expect it to be fine.

Issue #15485, opened 2021-09-29
Normalise tile element type


Gymnasiast:
Tile element type is stored in the middle four bits of the `type` variable (first byte). Rather than shifting it right by 2 and returning that (as we did in all other cases), we have been using unshifted values everywhere, only shifting them when they are uses to index arrays.

It would be much cleaner to use shifted values everywhere. This would require defining a new enum (preferably a strong enum), a new method to get and set them and then replacing all checks with the new stuff. After everything has been done, the old methods can be removed and the new ones renamed.

PR #16148, opened 2021-12-07
Replace many GetType() calls with GetTypeN()

Gymnasiast:
Part of #15485.

### DIFF
--- a/src/openrct2/ride/RideRatings.cpp
+++ b/src/openrct2/ride/RideRatings.cpp
@@ -218,7 +218,7 @@ static void ride_ratings_update_state_2(RideRatingUpdateState& state)
     {
         if (tileElement->IsGhost())
             continue;
-        if (tileElement->GetType() != TILE_ELEMENT_TYPE_TRACK)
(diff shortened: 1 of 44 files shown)


src/openrct2/ride/RideRatings.cpp
```diff
@@ -1469,8 +1469,8 @@ static int32_t ride_ratings_get_scenery_score(Ride* ride)
                 if (tileElement->IsGhost())
                     continue;
 
-                int32_t type = tileElement->GetType();
-                if (type == TILE_ELEMENT_TYPE_SMALL_SCENERY || type == TILE_ELEMENT_TYPE_LARGE_SCENERY)
+                auto type = tileElement->GetTypeN();
```
Gymnasiast, line 1472:
```suggestion
                const auto type = tileElement->GetTypeN();
```

ZehMatt:
Is there a reason to add the N there?

Gymnasiast:
@ZehMatt It’s all explained in #15485. The postfix will be dropped if all calls have been changed over.

PR #16150, opened 2021-12-07
Refactor footpath, scenery, terrain and mapgen windows to use ImageId

IntelOrca:


Gymnasiast (changes requested):
When zooming in fully, the zoom-in button disappears instead of getting greyed out. The same happens to the zoom-out button when fully zoomed out.

afbeelding

Gymnasiast (approved):
Since the button issue is in `develop` and will be addressed in a separate PR, merging this.

PR #16151, opened 2021-12-07
Refactor most map paint routines to use ImageId

IntelOrca:


src/openrct2/paint/tile_element/Paint.Surface.cpp
```diff
@@ -1073,20 +1072,25 @@ void PaintSurface(paint_session* session, uint8_t direction, uint16_t height, co
...
+        {
+            imageId = get_surface_image(session, terrain_type, image_offset, rotation, grassLength, showGridlines, false);
         }
         if (session->ViewFlags & (VIEWPORT_FLAG_UNDERGROUND_INSIDE | VIEWPORT_FLAG_HIDE_BASE))
         {
-            imageId &= 0xDC07FFFF; // remove colour
-            imageId |= 0x41880000;
+            imageId = imageId.WithRemap(FilterPaletteID::PaletteDarken1).WithBlended(true);
+            // imageId &= 0xDC07FFFF; // remove colour
+            // imageId |= 0x41880000;
```
Gymnasiast, line 1088:
```suggestion
```

Gymnasiast (changes requested):
Path supports with remap colour do not render correctly (these should be grey):

afbeelding

PR #16163, opened 2021-12-09
Fix Go Kart Random Speeds

deurklink:
Go karts in OpenRCT2 only get 2 different random speeds, very fast (around 37 kmh) and very slow (around 25 kmh)
In vanilla, Go Karts can also get a speed in between. 

The Randomvalue & 16 probably is a mistake, changing it to Randomvalue & 15 makes the behavior same as vanilla again.


duncanspumpkin:
Replays are expected to fail. I'll take care of that. Will need a network version increment as well

Issue #16162, opened 2021-12-09
Random Go Kart speeds are not calculated the same as in vanilla RCT2


deurklink:
### Operating System

Windows 10, 64 bit

### OpenRCT2 build

v0.3.4.1 a396289

### Describe the issue

In vanilla RCT2, Go Karts usually separate into three groups with different speeds: one with a speed of around 37 kmh, one with a speed of around 33 kmh, and one with a speed of around 27 kmh.

In OpenRCT2, Go Karts only get 2 speeds, half of the karts will travel at 37 kmh, and half of the karts travel at 25 kmh.

```
static void ride_race_init_vehicle_speeds(Ride* ride)
{
    for (size_t i = 0; i < ride->num_vehicles; i++)
    {
        Vehicle* vehicle = GetEntity<Vehicle>(ride->vehicles[i]);
        if (vehicle == nullptr)
            continue;

        vehicle->ClearUpdateFlag(VEHICLE_UPDATE_FLAG_6);

        rct_ride_entry* rideEntry = vehicle->GetRideEntry();

        vehicle->speed = (scenario_rand() & 15) - 8 + rideEntry->vehicles[vehicle->vehicle_type].powered_max_speed;
```

### Steps to reproduce

1. 
2. 
3.

PR #16165, opened 2021-12-09
Fix: #16162 Go Kart speeds not correctly randomised

deurklink:
In OpenRCT2, at the start of a race Go karts will only get the maximum possible speed or the minimum possible speed. In vanilla, Go karts could also get a speed in between. 
This PR will make the behavior same as in vanilla again.
Issue: #16162

### DIFF
--- a/distribution/changelog.txt
+++ b/distribution/changelog.txt
@@ -22,6 +22,7 @@
 - Fix: [#16063] Object Selection preview for objects with glass is broken.
 - Fix: [#16075] Exporting track designs saves scenery in incorrect locations.
 - Fix: [#16087] The Looping Roller Coaster booster is now always drawn correctly.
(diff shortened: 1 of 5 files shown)


distribution/changelog.txt
```diff
@@ -22,6 +22,7 @@
 - Fix: [#16063] Object Selection preview for objects with glass is broken.
 - Fix: [#16075] Exporting track designs saves scenery in incorrect locations.
 - Fix: [#16087] The Looping Roller Coaster booster is now always drawn correctly.
+- Fix: [#16162] Go Karts speeds not correctly randomized, they only go very fast or very slow.
```
Gymnasiast, line 25:
```suggestion
- Fix: [#16162] Go Karts speeds are not correctly randomised, they only go very fast or very slow.
```

duncanspumpkin:
image
confirmed this was a mistake during implementation.

duncanspumpkin (approved):

PR #16172, opened 2021-12-10
Refactor track groups

spacek531:
refactor track groups to allow for greater than 64 groups


spacek531:
There's a bug right now where all track elements are available and chain can be placed on any track. I don't know if this is a regression or not but it probably is.

src/openrct2/ride/RideData.h
```diff
@@ -142,16 +142,26 @@ struct UpkeepCostsDescriptor
     uint8_t CostPerStation;
 };
 
+using track_group_t = std::array<bool, TRACK_GROUP_COUNT>;
```
ZehMatt, line 145: I'm not so sure this is the best approach of doing this. This field currently acts as a bitfield to define the group, it would be practical to actually have a table that defines the actual groups and then we can reference the group by index which only requires a single byte for 255 entries.

ZehMatt (changes requested):

src/openrct2/ride/RideData.h
spacek531: That would require hardcoding or otherwise registering every desired combination of available and extra track groups/elements. I think that will make it more complicated.

ZehMatt: Also please in the future if you want to do such big refactors create an issue first where we get the chance to discuss it with the team.

spacek531: I created a discussion about it, is that not sufficient?

The ride_track_group_t is not used dynamically unless I'm mistaken about what 'using' does. The array of bools uses a bitfield internally so the memory footprint should be the same before and after this PR.

ZehMatt: You are confusing specialized std::vector<bool> with std::array<bool, N>. The std::array does not do that. As for the discussion, no one ever greenlit anything. The discussions are there to "discuss" the proposed thing and get to a conclusion before starting the work.

src/openrct2/ride/RideData.h
```diff
@@ -142,16 +142,46 @@ struct UpkeepCostsDescriptor
...
+    {
+        data[trackFlagIndex / trackGroupDigits] &= ~(1 << (trackFlagIndex & trackGroupMask));
+    };
+    constexpr ride_track_group_t(std::initializer_list<uint16_t> il)
+    {
+        for (uint16_t i : il)
+            EnableTrack(i);
+    };
+};
+
```
ZehMatt, line 174: I was thinking a bit about how this simplify this, you can basically remove the entire wrapper.
I think a more elegant solution would be doing
```
using RideTrackGroup = std::bitset<64>;
```
And as for the SetField problem
```
constexpr RideTrackGroup MakeTrackGroup(const std::initializer_list<uint16_t>& vals)
{
 // Construct from vals...
}
...
SET_FIELD(EnabledTrackPieces, MakeTrackGroup({})),
```
By doing this we can continue to use all bit operations such as  or, xor, and etc.

ZehMatt (changes requested):

src/openrct2/ride/RideData.h
spacek531: Duncan and I discussed bitset in the discussion regarding this pull request.

duncanspumpkin: My concern with bitset was serialisation and initialising. @ZehMatt is suggesting a solution to the initialisation issue and the serialisation part of it is not an issue (for now).

spacek531: Is calling the function in the SET_FIELD function and using the return value substantially different from https://github.com/OpenRCT2/OpenRCT2/pull/16172/commits/422f55796b50475d912abe0bc2c8ac2009a62d21 with regards to intellisense whining about using non-constexpr method 'set'?

ZehMatt: Apparently std::bitset is pretty shit, I looked up the documentation and only a single constructor has the constexpr keyword. We should probably build our own bitset in this case, fortunately that is rather easy to do.

src/openrct2/ride/RideData.h
```diff
@@ -142,16 +144,18 @@ struct UpkeepCostsDescriptor
     uint8_t CostPerStation;
 };
 
+using ride_track_group_t = OpenRCT2::BitSet<TRACK_GROUP_COUNT>;
```
ZehMatt, line 147:
```suggestion
using RideTrackGroup = OpenRCT2::BitSet<TRACK_GROUP_COUNT>;
```

src/openrct2/ride/RideData.cpp
```diff
@@ -320,18 +320,17 @@ bool RideTypeDescriptor::HasFlag(uint64_t flag) const
     return Flags & flag;
 }
 
-uint64_t RideTypeDescriptor::GetAvailableTrackPieces() const
+void RideTypeDescriptor::GetAvailableTrackPieces(ride_track_group_t* a) const
```
ZehMatt, line 323:
```suggestion
void RideTypeDescriptor::GetAvailableTrackPieces(RideTrackGroup& res) const
```

src/openrct2/ride/RideData.cpp
```diff
@@ -320,18 +320,17 @@ bool RideTypeDescriptor::HasFlag(uint64_t flag) const
...
 {
-    if (gCheatsEnableAllDrawableTrackPieces)
+    for (int i = 0; i < TRACK_GROUP_COUNT; i++)
```
ZehMatt, line 325: This is in-efficient, do it the old way.
```cpp
res |= EnabledTrackPieces;
if (gCheatsEnableAllDrawableTrackPieces)
    res |= ExtraTrackPieces;
```

src/openrct2/ride/RideData.cpp
```diff
@@ -356,3 +355,16 @@ ResearchCategory RideTypeDescriptor::GetResearchCategory() const
     log_error("Cannot get Research Category of invalid RideCategory");
     return ResearchCategory::Transport;
 }
+
+// used by _legacy.cpp and RideConstruction.cpp
```
ZehMatt, line 359:
```suggestion
// used by _legacy.cpp and RideConstruction.cpp
```
static can be only used in a single TU, drop comment.

ZehMatt (changes requested):

spacek531:
I've made the recommended changes but I can't compile it any more for reasons I don't know. I'm using VS 19 community.

The linker really does not like this, and fails with C1001 "internal compiler error" on `AirPoweredVerticalCoaster.cpp`, the first #include in `RideData.cpp`.

ZehMatt:
Try updating VS 2019 to the latest if you haven't. The CI seems fine

src/openrct2/ride/RideData.cpp
```diff
@@ -356,3 +354,17 @@ ResearchCategory RideTypeDescriptor::GetResearchCategory() const
     log_error("Cannot get Research Category of invalid RideCategory");
     return ResearchCategory::Transport;
 }
+
+static RideTrackGroup _enabledRidePieces = {};
+
+// used by RideConstruction.cpp and _legacy.cpp
```
ZehMatt, line 360: Not sure this comment is helping anyone really you can remove that as well, no one is gonna track where and how certain functions are used.

spacek531:
It exists there only because it's used by _legacy.cpp. If/when the _legacy.cpp code is rolled into RideConstruction.cpp, all the code can be put into RideConstruction.cpp

ZehMatt:
Either way the comments are pointless, comments should only exist when they make sense like briefly describing the functionality so one can skip reading the entire code.

spacek531:
If the wording of the comment is the issue, that can be fixed. Now it briefly explains why the code is there and saves a future developer from wondering. I might start a wiki page for it too, one can always appreciate documenting.

spacek531:
The comment is removed, I hope this can be looked at soon

ZehMatt (approved):
Seems alright to me, I did not test this I'm currently busy with other projects.

src/openrct2/ride/RideData.cpp
```diff
@@ -320,18 +320,16 @@ bool RideTypeDescriptor::HasFlag(uint64_t flag) const
     return Flags & flag;
 }
 
-uint64_t RideTypeDescriptor::GetAvailableTrackPieces() const
+void RideTypeDescriptor::GetAvailableTrackPieces(RideTrackGroup& res) const
```
Gymnasiast, line 323: Why are passing an output parameter instead of just returning something?

src/openrct2/ride/RideData.cpp
```diff
@@ -320,18 +320,16 @@ bool RideTypeDescriptor::HasFlag(uint64_t flag) const
...
-        return EnabledTrackPieces | ExtraTrackPieces;
-    }
-    return EnabledTrackPieces;
+        res |= ExtraTrackPieces;
 }
 
 bool RideTypeDescriptor::SupportsTrackPiece(const uint64_t trackPiece) const
 {
-    return GetAvailableTrackPieces() & (1ULL << trackPiece);
+    return EnabledTrackPieces.get(trackPiece) || (gCheatsEnableAllDrawableTrackPieces && ExtraTrackPieces.get(trackPiece));
```
Gymnasiast, line 332: What is the point of this, considering `GetAvailableTrackPieces()` already takes care of checking `gCheatsEnableAllDrawableTrackPieces`?

PR #16190, opened 2021-12-11
Part of #15485: normalise all OpenRCT2 tile element types

Gymnasiast:
This just leaves the RCT12TileElementType to be refactored, which will be in a PR of its own.


ZehMatt (approved):
Looks good

PR #16191, opened 2021-12-11
Close #16002 : Increase the buffer size to 256 and add an assert

### LINKED ISSUE #16002: Segfault when opening ride statistics for huge ride in NSF
### Operating System

Linux (x86-64)

### OpenRCT2 build

OpenRCT2, v0.3.5.1-54 (2b4eaab5d on develop, DEBUG)

### Describe the issue

When opening the statistics tab in the monorail ride in the attached park the game crashes with a segmentation fault. Presumably it's because the ride is humongous with 36 stations and 108 trains. Statistics for normal-sized rides work fine.

### Area(s) with issue?

_No response_

### Steps to reproduce

1. Open the attached park
2. Click on the monorail ride
3. Click on the statistics (measurements and test results) tab
4. The game crashes

### Attachments

[Big Monorail Stats Segfault.park.zip](https://github.com/OpenRCT2/OpenRCT2/files/7605076/Big.Monorail.Stats.Segfault.park.zip)

frutiemax:


duncanspumpkin:
I don't think this is a sensible solution. You are just making the performance worse for everyone for an edge case.

IntelOrca:
I don't think 256 is too big for a buffer, although I would just write 256, not numeric_limits. If we want to allow larger buffers, then we it might be best sharing a larger thread_safe std::string.

ZehMatt:
One of the buffers used to be 256, dont remember which one so its not too far fetched. However as suggested above using numeric limits for this is not practical.

PR #16195, opened 2021-12-11
Forward declare Formatter

frutiemax:
When modifying Formatter.h, it took too long to rebuild the code. With proper forward declaring, this should be faster.

### DIFF
--- a/src/openrct2/drawing/Text.cpp
+++ b/src/openrct2/drawing/Text.cpp
@@ -9,6 +9,7 @@
 
 #include "Text.h"
 
+#include "../localisation/Formatter.h"
 #include "../localisation/Localisation.h"
 #include "Drawing.h"
 
@@ -123,10 +124,24 @@ static void DrawText(
     DrawText(dpi, coords, paint, buffer);
 }
 
+void DrawTextBasic(rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, rct_string_id format)
+{
+    Formatter ft;
+    TextPaint textPaint;
+    DrawTextBasic(dpi, coords, format, ft, textPaint);
+}
+
 void DrawTextBasic(
     rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, rct_string_id format, const Formatter& ft, TextPaint textPaint)
 {
     DrawText(dpi, coords, textPaint, format, ft.Data());
+}
+
+void DrawTextEllipsised(rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, int32_t width, rct_string_id format)
+{
+    Formatter ft;
+    TextPaint textPaint;
+    DrawTextEllipsised(dpi, coords, width, format, ft, textPaint);
 }
 
 void DrawTextEllipsised(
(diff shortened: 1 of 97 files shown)


src/openrct2/core/ChecksumStream.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include "Endianness.h"
 
+#include <array>
```
Broxzier, line 14: This is unnecessary. std::array is already included in the header, and here it's only used for a function definition that comes from the same header.

src/openrct2/object/WaterObject.cpp
```diff
@@ -53,6 +57,7 @@ void WaterObject::Unload()
 void WaterObject::DrawPreview(rct_drawpixelinfo* dpi, int32_t width, int32_t height) const
 {
     // Write (no image)
+
```
tupaschoal, line 60: Unrelated change, please revert.

tupaschoal (approved):
Looks alright to me, I'm just not very happy with the duplication of the three `Draw` functions, but I guess it's either that or some opaque pointer implementation

src/openrct2/drawing/Text.cpp
```diff
@@ -123,12 +124,26 @@ static void DrawText(
     DrawText(dpi, coords, paint, buffer);
 }
 
+void DrawTextBasic(rct_drawpixelinfo* dpi, const ScreenCoordsXY& coords, rct_string_id format)
+{
+    Formatter ft;
```
Gymnasiast, line 129: I think this needs initialisation. (Also applies to the other two functions)

```suggestion
    Formatter ft{};
```

Gymnasiast:
Why is this duplication/overload in the functions necessary?

tupaschoal:
> Why is this duplication/overload in the functions necessary?

My guess is to avoid the default parameter of `Formatter = {}` which might have required the full definition to be included.

src/openrct2/drawing/Text.cpp
frutiemax: > Why is this duplication/overload in the functions necessary?

You cannot use the constructor in the header when it is forward declared. Some function calls didn't used the same number of parameters so the duplication is needed.

Gymnasiast:
I am reasonably satisfied with this, but I’d like the opinion of @IntelOrca , @duncanspumpkin or @ZehMatt on whether this is a good change.

Obviously, you’d also have to fix compilation for macOS and Android.

ZehMatt (approved):
It's actually a good change. As of the new overloads I think its fine, personally I dislike default parameter values especially when used on complex types so I'm in favor for that, there are too many stories where passing default values leads to users using the functions wrong.

frutiemax:
I am not sure how to fix this error for the macos build:
```
/Users/runner/work/OpenRCT2/OpenRCT2/src/openrct2/platform/Shared.cpp:241:33: error: no viable constructor or deduction guide for deduction of template arguments of 'array'
    static constexpr std::array prohibited = { '/' };
```

tupaschoal:
> I am not sure how to fix this error for the macos build:
> 
> ```
> /Users/runner/work/OpenRCT2/OpenRCT2/src/openrct2/platform/Shared.cpp:241:33: error: no viable constructor or deduction guide for deduction of template arguments of 'array'
>     static constexpr std::array prohibited = { '/' };
> ```

Try doing:
```c++
static constexpr std::array<std::string> prohibited = { '/' };
```

Or `char`

PR #16199, opened 2021-12-12
Fix #16197. Park award time and type accidentally swapped

### LINKED ISSUE #16197: All awards show up as the "worst value park" award
### Operating System

Windows 10 64bit

### OpenRCT2 build

OpenRCT2, v0.3.5.1 (b83f7c3 on develop) provided by GitHub

### Describe the issue

I got 3 awards, but in the awards screen they all show up as the worst value park.

### Area(s) with issue?

New Save Format (NSF)

### Steps to reproduce

1. Get awards
2. Check awards screen

### Attachments

Coaster Dreams 2021-12-12 01-47-47

duncanspumpkin:
Mistake caused by incorrect types so I went and also made it type safe.

### DIFF
--- a/src/openrct2/management/Award.h
+++ b/src/openrct2/management/Award.h
@@ -13,13 +13,7 @@
 
 #include <vector>
 
-struct Award
-{
-    uint16_t Time;
-    uint16_t Type;
-};
-
-enum class ParkAward : uint8_t
+enum class ParkAward : uint16_t
 {
     MostUntidy,
     MostTidy,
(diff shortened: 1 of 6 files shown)


src/openrct2/management/Award.h
```diff
@@ -38,13 +32,19 @@ enum class ParkAward : uint8_t
     MostDazzlingRideColours,
     MostConfusingLayout,
     BestGentleRides,
-    Count
+    Count // Count must be less than 32
```
Gymnasiast, line 35: If it must be less than 32, let’s add a static assert.

src/openrct2/management/Award.h
```diff
@@ -13,13 +13,7 @@
...
 #include <vector>
 
-struct Award
-{
-    uint16_t Time;
-    uint16_t Type;
-};
-
-enum class ParkAward : uint8_t
+enum class ParkAward : uint16_t
```
Gymnasiast, line 16: Having `ParkAward` and `Award` is quite confusing. Would be better to rename to `AwardType`.

May be done in a later PR, though.

Gymnasiast (approved):
While I would like to see the line notes addressed, I think it’s important to merge this quickly.

PR #16207, opened 2021-12-12
#16198: Implement custom BitSet and replace std::bitset

### LINKED ISSUE #16198: make custom bitset
Regarding expanding the track groups field past 64 bits:

Apparently std::bitset is pretty shit, I looked up the documentation and only a single constructor has the constexpr keyword. We should probably build our own bitset in this case, fortunately that is rather easy to do.

_Originally posted by @ZehMatt in https://github.com/OpenRCT2/OpenRCT2/pull/16172#discussion_r767187978_

ZehMatt:
A few key differences to std::bitset
```cpp
sizeof(BitSet<1>) == 1
sizeof(BitSet<15>) == 2
sizeof(BitSet<24>) == 4
sizeof(BitSet<44>) == 8
```
Where std::bitset
```cpp
sizeof(std::bitset<1>) == sizeof(size_t)
sizeof(std::bitset<15>) == sizeof(size_t)
sizeof(std::bitset<24>) == sizeof(size_t)
sizeof(std::bitset<44>) == sizeof(size_t)
```

Also supports initialization for indices:
```cpp
BitSet<12> bits({1, 5, 9});
// Has bit 1, 5, 9 set.
```

Construction for biggest possible value:
```cpp
BitSet<64> bits(0xDEADBEEFF00ULL);
// Assigns the 64 bit value.
```
In case of
```cpp
BitSet<128> bits(0xDEADBEEFF00ULL);
```
Would only be able to assign the first 64 bits.

Also to note: Currently shift and rotate is not implemented, we don't really use it and can be done when needed as they can be quite complex for when its bigger than the pointer size.

Closes #16198

### DIFF
--- a/src/openrct2/core/BitSet.hpp
+++ b/src/openrct2/core/BitSet.hpp
@@ -22,41 +22,41 @@ namespace OpenRCT2
     {
         namespace BitSet
         {
-            static constexpr size_t byte_bits = std::numeric_limits<std::underlying_type_t<std::byte>>::digits;
+            static constexpr size_t BitsPerByte = std::numeric_limits<std::underlying_type_t<std::byte>>::digits;
 
-            template<size_t TNumBits> static constexpr size_t round_bits()
+            template<size_t TNumBits> static constexpr size_t ByteAlignBits()
             {
-                const auto reminder = TNumBits % byte_bits;
+                const auto reminder = TNumBits % BitsPerByte;
                 if constexpr (reminder == 0u)
                 {
                     return TNumBits;
                 }
                 else
                 {
-                    return TNumBits + (byte_bits - (TNumBits % byte_bits));
+                    return TNumBits + (BitsPerByte - (TNumBits % BitsPerByte));
                 }
             }
 
-            static_assert(round_bits<1>() == 8);
-            static_assert(round_bits<4>() == 8);
-            static_assert(round_bits<8>() == 8);
-            static_assert(round_bits<9>() == 16);
-            static_assert(round_bits<9>() == 16);
-            static_assert(round_bits<17>() == 24);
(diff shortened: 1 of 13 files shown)


duncanspumpkin:
Why bother replacing the existing bit set users when they work fine with the std one. This is only required in some situations.

ZehMatt:
> Why bother replacing the existing bit set users when they work fine with the std one. This is only required in some situations.

Consistency I guess. Why mix the two if we have a better version

src/openrct2/entity/Guest.h
```diff
@@ -33,6 +32,8 @@
 #define PEEP_MAX_NAUSEA 255
 #define PEEP_MAX_THIRST 255
 
+using namespace OpenRCT2;
```
duncanspumpkin, line 35: I'd prefer that no `using namespaces` exist in headers.

src/openrct2/core/BitSet.hpp
```diff
@@ -0,0 +1,545 @@
...
+#include <limits>
+#include <string>
+
+namespace OpenRCT2
+{
+    namespace Detail
+    {
+        namespace BitSet
+        {
+            static constexpr size_t byte_bits = std::numeric_limits<std::underlying_type_t<std::byte>>::digits;
```
duncanspumpkin, line 25: i know we said that nothing at present conforms to much of a style for these but pretty much nothing should be snake_case.

ZehMatt: I wrote this code in STL style, I can change it all.

ZehMatt: I renamed everything to the point of not breaking STL compatibility

duncanspumpkin (approved):
Looks good.

PR #16208, opened 2021-12-13
Close #13768: Refactor window to class: DemolishRidePrompt 

### LINKED ISSUE #13768: Refactor window to class: DemolishRidePrompt
See #13737 for more details.

bkolstad1:
Refactors the DemolishRidePrompt window to use the new window class framework

Resolves #13768

### DIFF
--- a/src/openrct2-ui/windows/DemolishRidePrompt.cpp
+++ b/src/openrct2-ui/windows/DemolishRidePrompt.cpp
@@ -19,9 +19,12 @@
 static constexpr const int32_t WW = 200;
 static constexpr const int32_t WH = 100;
 
-static money32 _demolishRideCost;
-
 // clang-format off
+enum WindowRideDemolishPage {
+    WINDOW_RIDE_DEMOLISH,
+    WINDOW_RIDE_REFURBISH,
+};
+
 enum WindowRideDemolishWidgetIdx {
     WIDX_BACKGROUND,
     WIDX_TITLE,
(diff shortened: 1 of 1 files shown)


duncanspumpkin:
This looks like two different windows masquerading as one. Could you have a think about if they can be split into two windows you could even have a common base possibly but tbh it doesn't look like they share much information between them.

bkolstad1:
Yeah I agree it makes sense to split them up, didn't know if there was a reason they were together or something. I'll do that real quick

duncanspumpkin:
Looks good. I need to have a little think about if it should share window class numbers. I think it shouldn't but I'll check one or two things

duncanspumpkin:
Okay I think it needs a separate Window class number but lets do that in a follow on PR as it needs changes in a few locations.

duncanspumpkin (approved):

tupaschoal:
> Okay I think it needs a separate Window class number but lets do that in a follow on PR as it needs changes in a few locations.

Will you open an issue to track that? Or are you doing it? Just making sure it doesn't get lost :D

Gymnasiast:
This PR changed the permissions of two files for no good reason. That should have been fixed before merging.

duncanspumpkin:
> This PR set execute permissions on all three files. That should have been fixed before merging.

How do you even notice that?

Gymnasiast:
It’s easily visible under the Files Changed tab, at least for the existing files:

afbeelding

(Note the 644 changing to 755, indicating that the user, group and rest of the world were granted execute permission.)

PR #16215, opened 2021-12-13
Rename VAR_CE and its unions

spacek531:


duncanspumpkin:
NumLaps isn't a great name when its used for rotations / laps / duration...

spacek531:
I renamed it that because that's what the one named element in the union is called. Would you like me to add more descriptive names using a union?

duncanspumpkin:
First of all can we think of a name that would be better that would encompass all of the current users.

spacek531:
Like you said, it has 3 purposes: number of rotations of flatride, number of laps of tracked ride, and elapsed time in dodgems.

The associated parameter in the ride struct is named multiple names for the different circumstances but it may be that it is broken down by bit, I have not checked the object methods.
```c++
union
{
    uint8_t operation_option; // 0x0D0
    uint8_t time_limit;       // 0x0D0
    uint8_t num_laps;         // 0x0D0
    uint8_t launch_speed;     // 0x0D0
    uint8_t speed;            // 0x0D0
    uint8_t rotations;        // 0x0D0
}
 ```
 
It has all 3 purposes as a union.

RideCycleProgress is sort of like it but I think it's not specific enough. NumLapsRotationsElapsedTime does not roll off the tongue to me.

ZehMatt (approved):
This seems sensible to me, please re-word your commit message to better reflect the changes otherwise I'm fine with it.

spacek531:
I don't know how to re-word the commit messages.

I tried `git commit --amend` which did nothing, and `git rebase -i HEAD~1` and set the rebase type to `reword` and changed the message of the commit on its line. Both times `git log` showed the old commit messages.

ZehMatt:
Try ```git rebase -if HEAD~1```

spacek531:
I have rebased to the current develop. Is there anything else that needs to be done before this can be merged?

duncanspumpkin (approved):

PR #16216, opened 2021-12-13
rename VAR_CF

spacek531:


duncanspumpkin (changes requested):
Again doesn't make sense what does seat rotation have to do with brake speed

spacek531:
I have no idea but that's the minigolf shadow realm code. Whatever it's trying to read, it's probably not really rotation.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -8940,7 +8940,7 @@ void Vehicle::UpdateTrackMotionMiniGolfVehicle(Ride* curRide, rct_ride_entry* ri
 
     SetTrackType(tileElement->AsTrack()->GetTrackType());
     SetTrackDirection(direction);
-    var_CF = tileElement->AsTrack()->GetSeatRotation() << 1;
+    brake_speed = tileElement->AsTrack()->GetSeatRotation() << 1;
```
duncanspumpkin, line 8943: image
Looks like this was wrong it should have been:
```
brake_speed = tileElement->AsTrack()->GetBrakeBoosterSpeed();
```

duncanspumpkin:
Make my change and then that is all users of CF being correct and related to brake booster speed. Also you will need to bump the network version when you make my change.

src/openrct2/ParkFile.cpp
```diff
@@ -1823,7 +1823,7 @@ namespace OpenRCT2
         cs.ReadWrite(entity.scream_sound_id);
         cs.ReadWrite(entity.TrackSubposition);
         cs.ReadWrite(entity.var_CE);
-        cs.ReadWrite(entity.var_CF);
+        cs.ReadWrite(entity.brake_speed);
```
ZehMatt, line 1826: In the future use
```PascalCase``` for public variable names

PR #16217, opened 2021-12-13
Rename VAR_C4

spacek531:


duncanspumpkin (changes requested):
This one clearly does not make sense. What on earth does direction < 200 mean

spacek531:
Looking at it further, it appears it has one alternate use for the following specific cases:
- unsticking go karts colliding while changing lanes while MovingToEndOfStation
- unsticking boats that have become stuck in each other while TravellingBoat

So I would say it is also CollisionDetectionTimer

duncanspumpkin:
If its used for two purposes then I'd prefer a union for those two purposes.

spacek531:
I have rebased to the current develop. Is there anything else that needs to be done before this can be merged?

src/openrct2/rct2/RCT2.h
```diff
@@ -502,14 +502,10 @@ namespace RCT2
...
-        uint8_t powered_acceleration; // 0xC3
-        union
-        {
-            uint8_t dodgems_collision_direction; // 0xC4
-            uint8_t var_C4;
-        };
-        uint8_t animation_frame; // 0xC5
+        uint8_t speed;                       // 0xC2
+        uint8_t powered_acceleration;        // 0xC3
+        uint8_t CollisionDetectionDirection; // 0xC4
```
duncanspumpkin, line 507: Readd the union here

src/openrct2/ride/Vehicle.h
```diff
@@ -180,8 +180,8 @@ struct Vehicle : EntityBase
     uint8_t powered_acceleration;
     union
     {
-        uint8_t dodgems_collision_direction;
-        uint8_t var_C4;
+        uint8_t CollisionDetectionDirection;
```
duncanspumpkin, line 183: Rename back to DodgemsCollisionDirection so we know its only used for dodgems.

src/openrct2/ride/Vehicle.cpp
```diff
@@ -9838,7 +9838,7 @@ void Vehicle::Serialise(DataSerialiser& stream)
     stream << var_C0;
     stream << speed;
     stream << powered_acceleration;
-    stream << dodgems_collision_direction;
+    stream << CollisionDetectionDirection;
```
duncanspumpkin, line 9841: As the majority of users will be the timer switch all the things not dodgem related to use the CollisionDetectionTimer

duncanspumpkin (approved):

PR #16218, opened 2021-12-13
use field for rotation mode subtype

spacek531:
prerequisite for https://github.com/OpenRCT2/OpenRCT2/discussions/16211

I deliberately put merry-go-round in the 0 index because the bits line up nicely for it.

Tested to work on as many .DAT rotation mode flat rides I could put in a debug build.


spacek531:
I've marked this as ready for review.

src/openrct2/ride/Ride.h
```diff
@@ -464,9 +464,11 @@ enum
     RIDE_ENTRY_FLAG_PLAY_DEPART_SOUND = 1 << 3,
     RIDE_ENTRY_FLAG_ALTERNATIVE_SWING_MODE_1 = 1 << 4,
     // Twist type rotation ride
-    RIDE_ENTRY_FLAG_ALTERNATIVE_ROTATION_MODE_1 = 1 << 5,
+    RIDE_ENTRY_FLAG_ALTERNATIVE_ROTATION_MODE_1_DEPRECATED = 1 << 5,
     // Lifting arm rotation ride (enterprise)
-    RIDE_ENTRY_FLAG_ALTERNATIVE_ROTATION_MODE_2 = 1 << 6,
+    RIDE_ENTRY_FLAG_ALTERNATIVE_ROTATION_MODE_2_DEPRECATED = 1 << 6,
```
tupaschoal, line 469: Do we have a reason for keeping those deprecated enumerators?

I know we've done it before, looking a few lines below, but I don't remember if it was discussed or if there was any reasoning behind it, given they're not used anymore.

src/openrct2/ride/VehicleData.h
```diff
@@ -11,9 +11,23 @@
...
+        const uint8_t* accelerationAnimation, const uint8_t* continuousAnimation, const uint8_t* decelerationAnimation)
+    {
+        Animations[0] = accelerationAnimation;
+        Animations[1] = continuousAnimation;
+        Animations[2] = decelerationAnimation;
+    }
+    const uint8_t* GetAnimation(uint8_t sub_state) const
+    {
+        return Animations[sub_state];
+    }
```
tupaschoal, line 27: I think this can be made more C++zy by using `std::array` and constructor initializer lists, maybe even adding an `assert` to make sure we're accessing a valid index.

```suggestion
struct RotationSubtypeSpriteMap
{
    const std::array<const uint8_t*, 3> Animations;
    constexpr RotationSubtypeSpriteMap(
        const uint8_t* accelerationAnimation, const uint8_t* continuousAnimation, const uint8_t* decelerationAnimation)
        : Animations({ accelerationAnimation, continuousAnimation, decelerationAnimation })
    {
    }
    const uint8_t* GetAnimation(uint8_t sub_state) const
    {
        assert(sub_state < Animations.size());
        return Animations.at(sub_state);
    }
};
```

Another idea would be to defined a `constexpr` of the number of valid sub_states. Use that to choose the size of the `std::array` and to write `static_asserts` that the `enum`s controlling the substates have the same number of elements as this `constexpr` is, so compilation would fail if that constant ever changed without changing the enums and vice versa.

spacek531:
This PR is currently focused only on refactoring the existing system to use getters which is better but I think it could be focused better for a "final picture" of development that better integrates with other parts of flat ride code.


The following rides have a time-to-sprite map that writes to pitch: Enterprise, Merry-Go-Round, Twist, Swinging Ship, Flying Carpet, Swinging Inverter Ship, Space Rings, Motion Simulator.

Top spin has a time-to-sprite map that writes to pitch and bank.

Since top spin's two-axis sprite map can be seen as an extension of the single-axis sprite map, it makes some sense to apply the top spin system to all applicable flat rides - they can all use the same sprite updating code and can all use the same underlying data structure for the maps. Also, the data for these can't decide if it wants to be in VehicleData.cpp or Vehicle.cpp and if it should be a uint8_t or an int8_t (they should use uint8_t), so there is more room for standardizing things.

Going further from there, implementing new flat rides in OpenRCT2 would perhaps be made easier if the maps were defined in the ride object instead of the source code. So that's something that will have to be looked into.

sub_state is used in a few different ways by the different flat rides. Motion simulator and space rings make no use of sub state because their animations are always the same.

Top spin's sub_state is determined by the operating mode (the same as 3D Cinema, which it uses to determine ride cycle length).

Rotating operating mode rides (Enterprise, Merry-Go-Round, Twist) have exactly 3 sub_states. Sub_state 0 is the animation to ramp up to full motion. Sub_state 1 is the animation loop for one rotation. Sub_state 2 is the animation to slow down to a stop. Enterprise has an offset applied since it takes 4.5 rotations to accelerate to full or decelerate to stop - it applies the 9 rotation offset confusingly in Ride.cpp line 4893: 
```cpp
       bool shouldStop = true;
        if (curRide->status != RideStatus::Closed)
        {
            sprite = NumRotations + 1;
            if (curRide->type == RIDE_TYPE_ENTERPRISE)
                sprite += 9;

            if (sprite < curRide->rotations)
                shouldStop = false;
        }
```
This should be refactored to be a property of the ride object.

Swinging rides (Swinging Ship, Swinging Inverting Ship, Flying Carpet) use the same principle but have 4 sub_states instead of 3. The swing data SwingingTimeToSpriteMaps is mushed together instead of separating things by ride type like the rotation map - the swinging ship maps are 0-3, swinging inverting ship maps are 4-7, and flying carpet maps are 8-11.

Unlike rotating rides, the number of sub_states for swinging rides could be variable, since the animations are played in order until it reaches steady state, where it then plays the animations in reverse order (but not reverse) until it reaches the end of the ride cycle. Currently it is hardcoded with magic numbers:
```cpp
        // It takes 3 swings to get into full swing
        // ride->rotations already takes this into account
        if (NumSwings + 3 < curRide->rotations)
        {
            // Go to the next swing state until we
            // are at full swing.
            if (sub_state != 3)
            {
                sub_state++;
            }
            UpdateSwinging();
            return;
        }
```
Note: animations are not played in reverse, so the swing height should be the same for forward swing and backward swing.

spacek531:
This implementation is fine but there's also the ability to load these as vectors in the RideEntry, this would means new spinning rides could define their animations in the json instead of having to put them in the game. Thoughts?

spacek531:
Please keep open.

karst:
Can always reopen whenever this is continued with

Issue #12774, opened 2020-08-25
Plugin API: Scripts will not be re-initialised when a new scenario is loaded from within a running scenario.


Sadret:
Issue: Scripts will not be re-initialised when a new scenario is loaded from within a running scenario.
Possible solutions:
- Reload scripts when a new scenario is loaded.
- Add a scenario.loaded hook. (I personally favor this solution, as it also solves other problems / improves other use cases).

IntelOrca:
I will probably choose option two: adding a hook as some plugins may want to change the map such as one for editing title sequences or managing an automatic server.

Sadret:
As decided in a recent Discord discussion, this is now considered a bug and will be fixed by the first of the suggested solutions.

PR #16238, opened 2021-12-14
Fix: [#12774] Plug-ins now reload correctly

Sadret:
Note that plugins are automatically unloaded before they are loaded again.


Sadret:
Where do you suggest? Logically, plugins are loaded only after parks are opened (e.g. `LoadParkFromFile`), but not every time that a park is opened, the plugins have to be loaded.
One could bind it to the `asScenario` flag in `Context->LoadParkFromFile`, but I am not sure if that's enough for networked games. Also that parameter seems to be false if a park is loaded via `Game->game_load_or_quit_no_save_prompt`and I am not sure why.

ZehMatt:
Plugins should be always loaded when a park loads so I would try to find a function that is called in all instances. There might be exceptions for hot-reload but otherwise I can't think of any other place this has to happen. @IntelOrca can probably tell when plugins needs loading.

distribution/changelog.txt
```diff
@@ -15,6 +15,7 @@
 - Improved: [#10150] Ride stations are now properly checked if they’re sheltered.
 - Improved: [#10664, #16072] Visibility status can be modified directly in the Tile Inspector's list.
 - Change: [#16077] When importing SV6 files, the RCT1 land types are only added when they were actually used.
+- Fix: [#12774] [Plugin] Plug-ins are now reloaded when New Game option is used in a running game.
```
Gymnasiast, line 18: You need to describe the bug, not the fix. This also applies to the commit message.

PR #16241, opened 2021-12-14
Close #12733 : Sorting feature in TrackList window

### LINKED ISSUE #12733: [Enhancement/Feature?] Sorting options for track designs
In the same spirit as #7763 and #10675, it would be nice to be able to sort the list of track designs by things like excitement rating, cost, size, etc.

hjort96:
Been working on this for some time, and I would really appreciate it if someone wants to take a look at this.

This PR would implement sorting for the TrackList window as suggested in issue #12733.
Players would be able to sort the list of TrackDesigns on different types of values.

I'm unsure if it is implemented well or if there's a better way to do it.
Especially the different dropdowns and the function called `SelectDropDown`.

I've tried to test it thoroughly but there might still be some bugs and unwanted behavior.

Also, the reason for only having two commits is that I somehow broke (literally) everything on the previous branch and had to move code to this new one.

### DIFF
--- a/data/language/en-GB.txt
+++ b/data/language/en-GB.txt
@@ -3644,6 +3644,25 @@ STR_6538    :Shows regular paths in the queues dropdown of the Footpaths window.
 STR_6539    :Brake Closed
 STR_6540    :{WINDOW_COLOUR_2}Special thanks to the following companies for allowing their likeness:
 STR_6541    :{WINDOW_COLOUR_2}Rocky Mountain Construction Group, Josef Wiegand GmbH & Co. KG
+STR_6542    :Sort ▲
+STR_6543    :Sort ▼
+STR_6544    :Name
+STR_6545    :Cost
+STR_6546    :Excitement
+STR_6547    :Intensity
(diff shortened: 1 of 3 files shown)


karst:
I noticed this one has been open for extremely long without anyone responding or noticing things on it. I have tested the build and this does seem like a welcome addition to me. I noticed in my dev build it takes quite a bit of time for the items to re-sort items. I don't know if it's possible to optimize this any. In the actual release version this is not a problem. I'll make sure to poke the devs about this to look at it.

Gymnasiast:
> I'm unsure if it is implemented well or if there's a better way to do it.
> Especially the different dropdowns and the function called SelectDropDown.

 Yeah, this indeed needs a lot of work in that area. You also hardcoded a lot of magic numbers for the ride IDs. Instead, you should get the RideTypeDescriptor and determine which items to add to the dropdown based on the flags.

Gymnasiast:
Keep this open please.

ZehMatt:
@hjort96 do you plan to continue working on this?

Issue #16251, opened 2021-12-16
[Plugin] API specification and implementation do not match


Sadret:
This issue spans various parts of the API, mostly where a writable (non-readonly) field is specified as nullable, but writing null to it throws an error.
Examples include most nullable `TileElement` properties.

PR #16252, opened 2021-12-16
Improved: [#16251] Plugin API handles null values better.

Sadret:
The plugin API throws a lot of errors in situations where users do not expect it. This PR tries to improve this for all `TileElement`s.
Goals of this PR:
- `element.prop = element.prop` should not throw an error
- `element.prop = element.prop` should not change any data
- `element.prop = [value]` should do nothing if `prop` is not in use.
- `element.prop = null` should do nothing if `prop` is in use and `null` is no meaningful value in this context.
- If `element.prop = [value];` changed the property's value, `element.prop` should then return `[value]`.
- `element.prop` should return `null` if `prop` is not in use.

The third and fourth are important because e.g. `footpath.additionStatus` shares memory with `footpath.rideIndex`.

Changes in the API specification:
- Removed `CorruptElement` (deprecated).
- Made `footpath.object`, `footpath.surfaceObject`, `footpath.railingsObject`, `entrance.footpathObject` and `entrance.footpathSurfaceObject` nullable.
- (Added a missing space.)

Changes in the API implementation:
- Increased `OPENRCT2_PLUGIN_API_VERSION`.
- Fixed `ScTileElement::trackType_set` to accept `uint16_t` value.
- For each nullable property, setters can now handle null values without throwing an error. For details, see `ScTileElement.cpp` or `ScTileElement.hpp`.

### DIFF
--- a/src/openrct2/scripting/bindings/world/ScTileElement.cpp
+++ b/src/openrct2/scripting/bindings/world/ScTileElement.cpp
@@ -337,7 +337,7 @@ namespace OpenRCT2::Scripting
             duk_push_null(ctx);
         return DukValue::take_from_stack(ctx);
     }
-    void ScTileElement::trackType_set(uint8_t value)
+    void ScTileElement::trackType_set(uint16_t value)
     {
         ThrowIfGameStateNotMutable();
         auto el = _element->AsTrack();
@@ -406,37 +406,40 @@ namespace OpenRCT2::Scripting
         }
         return DukValue::take_from_stack(ctx);
     }
-    void ScTileElement::sequence_set(uint8_t value)
+    void ScTileElement::sequence_set(const DukValue& value)
     {
-        ThrowIfGameStateNotMutable();
-        switch (_element->GetType())
+        if (value.type() == DukValue::Type::NUMBER)
         {
-            case TileElementType::LargeScenery:
+            ThrowIfGameStateNotMutable();
+            switch (_element->GetType())
             {
-                auto el = _element->AsLargeScenery();
-                el->SetSequenceIndex(value);
-                Invalidate();
-                break;
-            }
-            case TileElementType::Track:
-            {
-                auto el = _element->AsTrack();
-                if (get_ride(el->GetRideIndex())->type != RIDE_TYPE_MAZE)
+                case TileElementType::LargeScenery:
                 {
-                    el->SetSequenceIndex(value);
+                    auto el = _element->AsLargeScenery();
+                    el->SetSequenceIndex(value.as_uint());
                     Invalidate();
+                    break;
                 }
-                break;
+                case TileElementType::Track:
+                {
+                    auto el = _element->AsTrack();
+                    if (get_ride(el->GetRideIndex())->type != RIDE_TYPE_MAZE)
+                    {
+                        el->SetSequenceIndex(value.as_uint());
+                        Invalidate();
+                    }
+                    break;
+                }
+                case TileElementType::Entrance:
+                {
+                    auto el = _element->AsEntrance();
+                    el->SetSequenceIndex(value.as_uint());
+                    Invalidate();
+                    break;
+                }
+                default:
+                    break;
             }
-            case TileElementType::Entrance:
-            {
-                auto el = _element->AsEntrance();
-                el->SetSequenceIndex(value);
-                Invalidate();
-                break;
-            }
-            default:
-                break;
         }
     }
 
@@ -474,7 +477,7 @@ namespace OpenRCT2::Scripting
         }
         return DukValue::take_from_stack(ctx);
     }
-    void ScTileElement::ride_set(int32_t value)
+    void ScTileElement::ride_set(const DukValue& value)
     {
         ThrowIfGameStateNotMutable();
         switch (_element->GetType())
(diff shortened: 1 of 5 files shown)


src/openrct2/scripting/bindings/world/ScTileElement.cpp
```diff
@@ -406,37 +406,40 @@ namespace OpenRCT2::Scripting
         }
         return DukValue::take_from_stack(ctx);
     }
-    void ScTileElement::sequence_set(uint8_t value)
+    void ScTileElement::sequence_set(const DukValue& value)
```
IntelOrca, line 409: Why not change this just to `uint16_t`? We don't handle any other type of value.

Sadret: For `TrackElement`s that are Mazes, the `sequence` is `null`, which means that `element.sequence = element.sequence` would throw an error if the parameter is of type `uint16_t`.

src/openrct2/scripting/bindings/world/ScTileElement.cpp
```diff
@@ -607,16 +628,19 @@ namespace OpenRCT2::Scripting
...
-            if (get_ride(el->GetRideIndex())->type == RIDE_TYPE_MAZE)
-            {
-                el->SetMazeEntry(value);
-                Invalidate();
-            }
+        if (value.type() == DukValue::Type::NUMBER)
+        {
+            ThrowIfGameStateNotMutable();
+            auto el = _element->AsTrack();
+            if (el != nullptr)
```
IntelOrca, line 637: We generally always use braces if the nested block has more than one line.

src/openrct2/scripting/bindings/world/ScTileElement.cpp
```diff
@@ -629,16 +653,19 @@ namespace OpenRCT2::Scripting
...
-            if (get_ride(el->GetRideIndex())->type != RIDE_TYPE_MAZE)
-            {
-                el->SetColourScheme(value);
-                Invalidate();
-            }
+        if (value.type() == DukValue::Type::NUMBER)
+        {
+            ThrowIfGameStateNotMutable();
+            auto el = _element->AsTrack();
+            if (el != nullptr)
```
IntelOrca, line 662: Same here

src/openrct2/scripting/bindings/world/ScTileElement.cpp
Sadret: As you can see, I just left it how it was.

src/openrct2/scripting/bindings/world/ScTileElement.cpp
```diff
@@ -673,16 +703,19 @@ namespace OpenRCT2::Scripting
...
-            if (TrackTypeHasSpeedSetting(el->GetTrackType()))
-            {
-                el->SetBrakeBoosterSpeed(value);
-                Invalidate();
-            }
+        if (value.type() == DukValue::Type::NUMBER)
+        {
+            ThrowIfGameStateNotMutable();
+            auto el = _element->AsTrack();
+            if (el != nullptr)
```
IntelOrca, line 712: and here

src/openrct2/scripting/bindings/world/ScTileElement.cpp
```diff
@@ -651,16 +678,19 @@ namespace OpenRCT2::Scripting
...
-            if (get_ride(el->GetRideIndex())->type != RIDE_TYPE_MAZE)
-            {
-                el->SetSeatRotation(value);
-                Invalidate();
-            }
+        if (value.type() == DukValue::Type::NUMBER)
+        {
+            ThrowIfGameStateNotMutable();
+            auto el = _element->AsTrack();
+            if (el != nullptr)
```
IntelOrca, line 687: and here

src/openrct2/scripting/bindings/world/ScTileElement.cpp
Sadret: Same here

src/openrct2/scripting/bindings/world/ScTileElement.cpp
Sadret: and here

src/openrct2/scripting/bindings/world/ScTileElement.cpp
Sadret: amd so on

src/openrct2/scripting/bindings/world/ScTileElement.cpp
```diff
@@ -1049,35 +1089,37 @@ namespace OpenRCT2::Scripting
             duk_push_int(ctx, idx.ToUnderlying());
         return DukValue::take_from_stack(ctx);
     }
-    void ScTileElement::bannerIndex_set(uint16_t value)
+    void ScTileElement::bannerIndex_set(const DukValue& value)
     {
-        ThrowIfGameStateNotMutable();
-
-        switch (_element->GetType())
+        if (value.type() == DukValue::Type::NUMBER)
```
IntelOrca, line 1094: Perhaps allow banner index to be set to null if the `null` is passed in?

Sadret: Yes that makes sense.

src/openrct2/scripting/bindings/world/ScTileElement.cpp
duncanspumpkin: well can you change it.

IntelOrca (approved):

Broxzier:
> `element.prop = [value]` should do nothing if `prop` is not in use.

At the very least, warn the plug-in writer that they're doing something wrong. It's up to them to ignore the warning. Failing silently is terrible design. Remember that properties exposed to JS are not standard JS properties, so setting any (invalid) value to property doesn't store it as with JS objects.

PR #16256, opened 2021-12-16
Close #16231: Do not load image table when creating object index

### LINKED ISSUE #16231: Do not load image table when creating object index
> @IntelOrca : ReadObjectContext has a flag to say whether it loads the images or not. But currently this is not set to false when creating the object cache. It constructs that context in CreateObjectFromJson, by which point you don't know whether you are creating the cache or not - so it needs to pass that down somehow.

Gymnasiast:


IntelOrca (approved):
Did you notice any speed up with building the index on game start?

Gymnasiast:
It seems to go from about 9-10 seconds to 7-8 seconds for building the cache. Would be nice if someone with Windows and a hard drive could test it as well.

PR #16258, opened 2021-12-17
Refactor paint structs to use ImageId and extend size of image list

IntelOrca:


src/openrct2/sprites.h
```diff
@@ -1119,5 +1119,5 @@ enum
     SPR_SCROLLING_TEXT_END = SPR_SCROLLING_TEXT_START + OpenRCT2::MaxScrollingTextEntries,
 
     SPR_IMAGE_LIST_BEGIN = SPR_SCROLLING_TEXT_END,
-    SPR_IMAGE_LIST_END = 0x7FFFE
+    SPR_IMAGE_LIST_END = SPR_IMAGE_LIST_BEGIN + 1000000,
```
Gymnasiast, line 1122: Could you create a constant for the `1000000`?

src/openrct2-ui/drawing/engines/opengl/TextureCache.h
```diff
@@ -202,7 +202,7 @@ class TextureCache final
     std::vector<Atlas> _atlases;
     std::unordered_map<GlyphId, AtlasTextureInfo, GlyphId::Hash, GlyphId::Equal> _glyphTextureMap;
     std::vector<AtlasTextureInfo> _textureCache;
-    std::array<uint32_t, 0x7FFFF> _indexMap;
+    std::array<uint32_t, 0x7FFFFF> _indexMap;
```
Gymnasiast, line 205: Could you create a constant for this?

src/openrct2/paint/Paint.cpp
```diff
@@ -551,11 +549,10 @@ static void PaintAttachedPS(rct_drawpixelinfo* dpi, paint_struct* ps, uint32_t v
     {
         auto screenCoords = ScreenCoordsXY{ attached_ps->x + ps->x, attached_ps->y + ps->y };
 
-        auto imageId = PaintPSColourifyImage(
-            ImageId::FromUInt32(attached_ps->image_id, ps->tertiary_colour), ps->sprite_type, viewFlags);
```
Gymnasiast, line 555: This is subtly different, as it used to set the tertiary colour from `ps` instead of `attached_ps`. Might not be a problem, but it’s something to watch.

src/openrct2/paint/Paint.h
```diff
@@ -82,12 +72,28 @@ struct paint_string_struct
     uint8_t* y_offsets;
 };
 
-union paint_entry
+struct paint_entry
```
Gymnasiast, line 75: What is the idea behind the changes to `paint_entry`?

duncanspumpkin: You can't use a complex type in a union

src/openrct2/paint/Paint.cpp
IntelOrca: Can you think why it would be using the tertiary colour from the parent paint node, rather than the tertiary colour of the node that was added?

src/openrct2/paint/Paint.h
IntelOrca: ImageId has a default constructor, you can not define structs / classes with default constructors inside a union.

src/openrct2-ui/drawing/engines/opengl/TextureCache.h
```diff
@@ -202,7 +203,7 @@ class TextureCache final
     std::vector<Atlas> _atlases;
     std::unordered_map<GlyphId, AtlasTextureInfo, GlyphId::Hash, GlyphId::Equal> _glyphTextureMap;
     std::vector<AtlasTextureInfo> _textureCache;
-    std::array<uint32_t, 0x7FFFF> _indexMap;
+    std::array<uint32_t, SPR_IMAGE_LIST_END> _indexMap;
```
Gymnasiast, line 206: Previously, `SPR_IMAGE_LIST_END` was `0x7FFFE` and this was `0x7FFFF`, so shouldn’t now be `SPR_IMAGE_LIST_END + 1`?

src/openrct2/paint/Paint.cpp
duncanspumpkin: We will still need to verify with some parks that there are no unexpected changes

src/openrct2-ui/drawing/engines/opengl/TextureCache.h
IntelOrca: I don't think so. `SPR_IMAGE_LIST_END` is now not part of the inclusive range of available image IDs. So it is already now implicity `lastid + 1`.

Gymnasiast (approved):
Tested it with several parks (including a few NE ones) and found no bugs.

src/openrct2/paint/Paint.h
```diff
@@ -82,12 +72,28 @@ struct paint_string_struct
...
 {
+private:
     paint_struct basic;
-    attached_paint_struct attached;
-    paint_string_struct string;
+
+public:
+    paint_struct* AsBasic()
+    {
+        return reinterpret_cast<paint_struct*>(this);
```
ZehMatt, line 83: I'm not a fan as I mentioned on Discord before. you can however make this actually safe by being explicit about it. Consider using ```    std::array<uint8_t, std::max(
        {
            sizeof(paint_struct), 
            sizeof(attached_paint_struct), 
            sizeof(paint_string_struct)
        })> data;``` as the data.

As for the function then.
```suggestion
        auto* res = reinterpret_cast<paint_struct*>(buffer.data());
        ::new ((void*)res) paint_struct();
        return res;
```
This would ensure you have properly initialized members, this approach would be the closest thing to variant without the need of index.

ZehMatt (changes requested):

src/openrct2/paint/Paint.h
IntelOrca: ok done. I am surprised this works, or doesn't break bench sprite sort though as these things might use `AsBasic` just to cast and have no intention on re-initialising it.

ZehMatt: Well glad it works, in the future we should look into splitting this into different pools maybe.

ZehMatt (approved):

PR #16266, opened 2021-12-18
More Ride.h cleanup

Gymnasiast:


ZehMatt (approved):
LGTM

src/openrct2/ride/Ride.cpp
```diff
@@ -5528,10 +5516,10 @@ int32_t ride_get_entry_index(int32_t rideType, int32_t rideSubType)
     return subType;
 }
 
-StationObject* ride_get_station_object(const Ride* ride)
+StationObject* Ride::GetStationObject() const
```
ZehMatt, line 5519:
```suggestion
const StationObject* Ride::GetStationObject() const
```
Makes probably more sense, no one should modify the result.

PR #16268, opened 2021-12-18
Pass Ride reference to track paint functions

Gymnasiast:
`ride` was always guaranteed to be set before attempting to call the paint function anyway, so turn it into a reference so we have one fewer thing to consider.

### DIFF
--- a/src/openrct2/ride/Ride.cpp
+++ b/src/openrct2/ride/Ride.cpp
@@ -5399,7 +5399,7 @@ bool ride_has_adjacent_station(Ride* ride)
 
 bool ride_has_station_shelter(Ride* ride)
 {
-    auto stationObj = ride->GetStationObject();
+    const auto stationObj = ride->GetStationObject();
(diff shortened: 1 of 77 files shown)


src/openrct2/ride/Ride.cpp
```diff
@@ -5399,7 +5399,7 @@ bool ride_has_adjacent_station(Ride* ride)
 
 bool ride_has_station_shelter(Ride* ride)
 {
-    auto stationObj = ride->GetStationObject();
+    const auto stationObj = ride->GetStationObject();
```
ZehMatt, line 5402:
```suggestion
    const auto* stationObj = ride->GetStationObject();
```
I've probably not seen that in the last PR, now is a good opportunity :)

ZehMatt (approved):
Nice, we should probably open an issue to do the same for paint_session* argument.

Gymnasiast:
@ZehMatt https://github.com/OpenRCT2/OpenRCT2/issues/16270

PR #16272, opened 2021-12-18
Pass paint_session as reference

Gymnasiast:
I got carried away, it seems. ;-)

### DIFF
--- a/src/openrct2/cmdline/BenchSpriteSort.cpp
+++ b/src/openrct2/cmdline/BenchSpriteSort.cpp
@@ -150,7 +150,7 @@ static void BM_paint_session_arrange(benchmark::State& state, const std::vector<
         state.PauseTiming();
         std::copy_n(local_s, std::size(sessions), sessions.begin());
         state.ResumeTiming();
-        PaintSessionArrange(&sessions[0].Session);
+        PaintSessionArrange(sessions[0].Session);
         benchmark::DoNotOptimize(sessions);
     }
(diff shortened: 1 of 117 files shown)


src/openrct2/cmdline/BenchSpriteSort.cpp
```diff
@@ -150,7 +150,7 @@ static void BM_paint_session_arrange(benchmark::State& state, const std::vector<
         state.PauseTiming();
         std::copy_n(local_s, std::size(sessions), sessions.begin());
         state.ResumeTiming();
-        PaintSessionArrange(&sessions[0].Session);
+        PaintSessionArrange(*(&sessions[0].Session));
```
Gymnasiast, line 153: @ZehMatt Is this correct, or should it be 
```suggestion
        PaintSessionArrange(sessions[0].Session);
```
?

ZehMatt: You are correct about that.

ZehMatt (approved):
Looks good to me except that one case, I don't expect any breakage with a change like that.

src/openrct2/cmdline/BenchSpriteSort.cpp
Gymnasiast: Which one is correct? The code in my suggestion?

ZehMatt: Yes, the first one is doing Dereference(PointerOf(Sessions[0].Session)) which makes no sense really.

